PR #8065, opened 2022-03-24
UCP/AM: Added implementation of rendez-vous protocol.

rakhmets:
## What
Added AM rendez-vous protocol for new infrastructure.

## Why ?
The changes are part of UCP transition to new infrastructure.

### DIFF
--- a/src/ucp/am/eager_multi.c
+++ b/src/ucp/am/eager_multi.c
@@ -8,6 +8,7 @@
 #endif
 
 #include "eager.inl"
+#include "ucp_am.inl"
 
 #include <ucp/proto/proto.h>
 #include <ucp/proto/proto_common.inl>
@@ -21,13 +22,6 @@
 #define UCP_AM_MID_FRAG_META_LEN \
     (sizeof(ucp_am_hdr_t) + sizeof(ucp_am_mid_ftr_t))
 
-
-static UCS_F_ALWAYS_INLINE int
-ucp_am_eager_multi_supported(const ucp_proto_init_params_t *init_params)
-{
-    return (init_params->select_param->op_id == UCP_OP_ID_AM_SEND) ||
-           (init_params->select_param->op_id == UCP_OP_ID_AM_SEND_REPLY);
-}
 
 static ucs_status_t
 ucp_am_eager_multi_bcopy_proto_init(const ucp_proto_init_params_t *init_params)
@@ -52,10 +46,11 @@ ucp_am_eager_multi_bcopy_proto_init(const ucp_proto_init_params_t *init_params)
         .first.tl_cap_flags  = UCT_IFACE_FLAG_AM_BCOPY,
         .middle.lane_type    = UCP_LANE_TYPE_AM_BW,
         .middle.tl_cap_flags = UCT_IFACE_FLAG_AM_BCOPY,
-        .max_lanes = init_params->worker->context->config.ext.max_eager_lanes
+        .max_lanes           = context->config.ext.max_eager_lanes
     };
 
-    if (!ucp_am_eager_multi_supported(init_params)) {
+    if (!ucp_am_check_init_params(init_params, UCP_AM_OP_ID_MASK_ALL,
+                                  UCP_PROTO_SELECT_OP_FLAG_AM_RNDV)) {
         return UCS_ERR_UNSUPPORTED;
     }
 
@@ -175,10 +170,11 @@ ucp_am_eager_multi_zcopy_proto_init(const ucp_proto_init_params_t *init_params)
         .first.tl_cap_flags  = UCT_IFACE_FLAG_AM_ZCOPY,
         .middle.lane_type    = UCP_LANE_TYPE_AM_BW,
         .middle.tl_cap_flags = UCT_IFACE_FLAG_AM_ZCOPY,
-        .max_lanes = init_params->worker->context->config.ext.max_eager_lanes
+        .max_lanes           = context->config.ext.max_eager_lanes
     };
 
-    if (!ucp_am_eager_multi_supported(init_params)) {
+    if (!ucp_am_check_init_params(init_params, UCP_AM_OP_ID_MASK_ALL,
+                                  UCP_PROTO_SELECT_OP_FLAG_AM_RNDV)) {
         return UCS_ERR_UNSUPPORTED;
     }
 
--- a/src/ucp/am/rndv.c
+++ b/src/ucp/am/rndv.c
@@ -1,0 +1,71 @@
+/**
+ * Copyright (C) 2022, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "ucp_am.inl"
+
+#include <ucp/proto/proto_single.inl>
+#include <ucp/rndv/proto_rndv.inl>
+
+
+static size_t ucp_am_rndv_rts_pack(void *dest, void *arg)
+{
+    ucp_request_t *req          = arg;
+    ucp_rndv_rts_hdr_t *rts_hdr = dest;
+    size_t rts_size;
+
+    ucp_am_fill_header(ucp_am_hdr_from_rts(rts_hdr), req);
+    rts_hdr->opcode = UCP_RNDV_RTS_AM;
+    rts_size        = ucp_proto_rndv_rts_pack(req, rts_hdr, sizeof(*rts_hdr));
+    ucp_am_pack_user_header(UCS_PTR_BYTE_OFFSET(rts_hdr, rts_size), req);
+
+    return rts_size + req->send.msg_proto.am.header_length;
+}
+
+UCS_PROFILE_FUNC(ucs_status_t, ucp_am_rndv_proto_progress, (self),
+                 uct_pending_req_t *self)
+{
+    ucp_request_t *req = ucs_container_of(self, ucp_request_t, send.uct);
+    const ucp_proto_rndv_ctrl_priv_t *rpriv;
+    ucs_status_t status;
+    size_t max_rts_size;
+
+    rpriv  = req->send.proto_config->priv;
+    status = UCS_PROFILE_CALL(ucp_proto_rndv_rts_request_init, req);
+    if (status != UCS_OK) {
+        ucp_proto_request_abort(req, status);
+        return UCS_OK;
+    }
+
+    max_rts_size = sizeof(ucp_rndv_rts_hdr_t) + rpriv->packed_rkey_size +
+                   req->send.msg_proto.am.header_length;
+    return UCS_PROFILE_CALL(ucp_proto_am_bcopy_single_progress, req,
+                            UCP_AM_ID_RNDV_RTS, rpriv->lane,
+                            ucp_am_rndv_rts_pack, req, max_rts_size, NULL);
+}
+
+ucs_status_t ucp_am_rndv_rts_init(const ucp_proto_init_params_t *init_params)
+{
+    if (!ucp_am_check_init_params(init_params, UCP_AM_OP_ID_MASK_ALL,
+                                  UCP_PROTO_SELECT_OP_FLAG_AM_EAGER) ||
+        (init_params->select_param->dt_class != UCP_DATATYPE_CONTIG)) {
+        return UCS_ERR_UNSUPPORTED;
+    }
+
+    return ucp_proto_rndv_rts_init(init_params);
+}
+
+ucp_proto_t ucp_am_rndv_proto = {
+    .name     = "am/rndv",
+    .desc     = NULL,
+    .flags    = 0,
+    .init     = ucp_am_rndv_rts_init,
+    .query    = ucp_proto_rndv_rts_query,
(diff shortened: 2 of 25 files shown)


src/ucp/am/ucp_am.inl
```diff
@@ -0,0 +1,56 @@
...
+static UCS_F_ALWAYS_INLINE int
+ucp_am_check_op_flag(const ucp_proto_init_params_t *init_params,
+                     uint16_t op_flags)
+{
+    return (init_params->select_param->op_flags & op_flags);
+}
+
+static UCS_F_ALWAYS_INLINE int
+ucp_am_check_init_params(const ucp_proto_init_params_t *init_params,
+                         uint16_t op_flags)
```
brminich, line 49: `exclude_flags`

src/ucp/proto/proto_select.h
```diff
@@ -31,6 +31,11 @@
 #define UCP_PROTO_SELECT_OP_FLAG_INTERNAL (UCP_PROTO_SELECT_OP_FLAGS_BASE << 1)
 
 
+/* Select eager/rendez-vous protocol for Active Message sends */
```
brminich, line 34: minor: plz use `rendezvous` (like in all other places in the code)

src/ucp/rndv/proto_rndv.c
```diff
@@ -311,13 +312,35 @@ ucs_status_t ucp_proto_rndv_rts_init(const ucp_proto_init_params_t *init_params)
         .mem_info.sys_dev    = init_params->select_param->sys_dev
     };
 
-    UCP_RMA_PROTO_INIT_CHECK(init_params, UCP_OP_ID_TAG_SEND);
+    return params;
+}
 
-    if (init_params->select_param->dt_class != UCP_DATATYPE_CONTIG) {
```
brminich, line 316: can it be removed already?

rakhmets: Do you mean remove the check for datatype inside initialization of RTS protocols?
If so, then the answer is no. Because it is not implemented yet, there are other places with asserts to check the datatype.

brminich: no, i meant we can't remove it now, but I see it is moved to another place

src/ucp/rndv/rndv_get.c
```diff
@@ -340,12 +340,6 @@ ucp_proto_rndv_ats_init(const ucp_proto_init_params_t *params)
         return UCS_ERR_UNSUPPORTED;
     }
 
-    if (params->rkey_config_key != NULL) {
```
dmitrygx, line 343: 👍

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1504,6 +1525,11 @@ class test_ucp_am_nbx_rndv_dts : public test_ucp_am_nbx_rndv {
 
         test_ucp_am_nbx::cleanup();
     }
+
+    virtual unsigned enable_proto()
+    {
+        return 0;
```
dmitrygx, line 1531: shall we test both new and old protocols for datatypes too?

src/ucp/proto/proto_select.h
```diff
@@ -31,6 +31,11 @@
 #define UCP_PROTO_SELECT_OP_FLAG_INTERNAL (UCP_PROTO_SELECT_OP_FLAGS_BASE << 1)
 
 
+/* Select eager/rendezvous protocol for Active Message sends */
+#define UCP_PROTO_SELECT_OP_FLAG_AM_EAGER (UCP_PROTO_SELECT_OP_FLAGS_BASE << 2)
+#define UCP_PROTO_SELECT_OP_FLAG_AM_RNDV  (UCP_PROTO_SELECT_OP_FLAGS_BASE << 3)
```
dmitrygx, line 36: can this operation flag be re-used for TAG protocol then?
if yes - wdyt to drop `AM` suffix from their names?

test/gtest/ucp/test_ucp_am.cc
rakhmets: It hasn't been implemented yet. I'll enable tests for new protocol within adding support of iov and generic datatypes.

src/ucp/proto/proto_select.h
rakhmets: guess these flags can be used in such way.
however, we can force eager/rndv selection for AM in current protocol implementation, but not fot TAG

test/gtest/ucp/test_ucp_am.cc
dmitrygx: I see, thanks

src/ucp/proto/proto_select.h
dmitrygx: > however, we can force eager/rndv selection for AM in current protocol implementation, but not fot TAG

I see, thanks

dmitrygx (approved):

src/ucp/am/rndv.c
```diff
@@ -0,0 +1,85 @@
...
+                 uct_pending_req_t *self)
+{
+    ucp_request_t *req = ucs_container_of(self, ucp_request_t, send.uct);
+    const ucp_proto_rndv_ctrl_priv_t *rpriv;
+    ucs_status_t status;
+    size_t max_rts_size;
+
+    rpriv          = req->send.proto_config->priv;
+    status         = UCS_PROFILE_CALL(ucp_proto_rndv_rts_request_init, req);
+    req->send.lane = rpriv->lane;
```
brminich, line 50: minor: is there a reason to initialize lane after calling init? It looks a bit weird that `status` is not checked immediately

src/ucp/am/ucp_am.inl
```diff
@@ -0,0 +1,56 @@
...
+
+static UCS_F_ALWAYS_INLINE int
+ucp_am_check_op_id(const ucp_proto_init_params_t *init_params,
+                   ucp_operation_id_t op_id)
+{
+    return (init_params->select_param->op_id == op_id);
+}
+
+static UCS_F_ALWAYS_INLINE int
+ucp_am_check_op_flag(const ucp_proto_init_params_t *init_params,
```
brminich, line 41: minor: plz rename to `ucp_am_check_op_flags` to be consistent with the parameter

brminich, line 41: minor: plz rename to `ucp_am_check_op_flags` to be consistent with the parameter

brminich (dismissed):

src/ucp/am/ucp_am.inl
```diff
@@ -0,0 +1,56 @@
...
+{
+    return (init_params->select_param->op_id == op_id);
+}
+
+static UCS_F_ALWAYS_INLINE int
+ucp_am_check_op_flags(const ucp_proto_init_params_t *init_params,
+                      uint16_t op_flags)
+{
+    return (init_params->select_param->op_flags & op_flags);
+}
```
yosefe, line 45: IMO these one-liners do not deserve a special function; in any case, they should be in proto.h or proto_common.h and can use them for all protocols - since they are not really am-specufic

src/ucp/am/rndv.c
```diff
@@ -0,0 +1,85 @@
...
+    rts_size        = ucp_proto_rndv_rts_pack(req, rts_hdr, sizeof(*rts_hdr));
+    if (req->send.msg_proto.am.header_length == 0) {
+        return rts_size;
+    }
+
+    total_size = rts_size + req->send.msg_proto.am.header_length;
+    max_bcopy  = ucp_ep_get_max_bcopy(req->send.ep, req->send.lane);
+    if (ucs_unlikely(total_size > max_bcopy)) {
+        ucs_fatal("RTS is too big %lu, max %lu", total_size, max_bcopy);
+    }
```
yosefe, line 33: not sure we need it: the transport that calls this function would already check if size exceeds limits. and in any case, better check it in ucp_am_send_nbx, compare to the max user header size that ucp_worker_query() would return, and only if ENABLE_PARAMS_CHECK is 1
so can also remove the if in line 25

src/ucp/am/rndv.c
```diff
@@ -0,0 +1,85 @@
...
+    size_t max_rts_size;
+
+    rpriv  = req->send.proto_config->priv;
+    status = UCS_PROFILE_CALL(ucp_proto_rndv_rts_request_init, req);
+    if (status != UCS_OK) {
+        ucp_proto_request_abort(req, status);
+        return UCS_OK;
+    }
+
+    req->send.lane = rpriv->lane;
```
yosefe, line 55: this is already done by ucp_proto_single_status_handle, as part of ucp_proto_am_bcopy_single_progress

src/ucp/am/ucp_am.inl
```diff
@@ -0,0 +1,56 @@
...
+
+static UCS_F_ALWAYS_INLINE int
+ucp_am_check_init_params(const ucp_proto_init_params_t *init_params,
+                         uint16_t exclude_flags)
+{
+    return (ucp_am_check_op_id(init_params, UCP_OP_ID_AM_SEND) ||
+            ucp_am_check_op_id(init_params, UCP_OP_ID_AM_SEND_REPLY)) &&
```
yosefe, line 52: maybe unite with ucp_am_eager_single_supported, by passing op_mask (which could be either SEND, SEND_REPLY, or both)

src/ucp/am/eager_multi.c
```diff
@@ -25,8 +26,8 @@
 static UCS_F_ALWAYS_INLINE int
 ucp_am_eager_multi_supported(const ucp_proto_init_params_t *init_params)
 {
-    return (init_params->select_param->op_id == UCP_OP_ID_AM_SEND) ||
-           (init_params->select_param->op_id == UCP_OP_ID_AM_SEND_REPLY);
+    return ucp_am_check_init_params(init_params,
```
yosefe, line 29: IMO we can remove the wrapper ucp_am_eager_multi_supported

src/ucp/rndv/proto_rndv.c
```diff
@@ -285,7 +285,8 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params)
     return UCS_OK;
 }
 
-ucs_status_t ucp_proto_rndv_rts_init(const ucp_proto_init_params_t *init_params)
+ucp_proto_rndv_ctrl_init_params_t
+ucp_proto_rndv_rts_init_params(const ucp_proto_init_params_t *init_params)
```
yosefe, line 289: why needed to change this function to return params, and call ucp_proto_rndv_ctrl_init outside?
we could check the AM conditions before calling it

src/ucp/tag/tag_rndv.c
```diff
@@ -23,12 +23,8 @@ void ucp_tag_rndv_matched(ucp_worker_h worker, ucp_request_t *rreq,
     rreq->recv.tag.info.sender_tag = ucp_tag_hdr_from_rts(rts_hdr)->tag;
     rreq->recv.tag.info.length     = rts_hdr->size;
 
-    if (worker->context->config.ext.proto_enable) {
-        ucp_proto_rndv_receive_start(worker, rreq, rts_hdr, rts_hdr + 1,
-                                     hdr_length - sizeof(*rts_hdr));
-    } else {
-        ucp_rndv_receive(worker, rreq, rts_hdr, rts_hdr + 1);
-    }
```
yosefe, line 31: IMO better check proto_enable before calling ucp_rndv_receive, to avoid extra func call overhead.
maybe put this if/else in a common inline function for both am and tag? in rndv.h

src/ucp/rndv/proto_rndv.inl
```diff
@@ -318,4 +318,17 @@ ucp_proto_rndv_recv_complete(ucp_request_t *req)
...
+static UCS_F_ALWAYS_INLINE void
+ucp_proto_rndv_receive(ucp_worker_h worker, ucp_request_t *rreq,
+                       const ucp_rndv_rts_hdr_t *rndv_rts_hdr,
+                       const void *rkey_buf, size_t rkey_length)
+{
+    if (worker->context->config.ext.proto_enable) {
+        ucp_proto_rndv_receive_start(worker, rreq, rndv_rts_hdr, rkey_buf,
+                                     rkey_length);
+    } else {
+        ucp_rndv_receive(worker, rreq, rndv_rts_hdr, rkey_buf);
```
yosefe, line 330: protov2 code should not call protov1 code (only vice versa) so move this function to rndv.h and rename to smth without the word proto

src/ucp/rndv/rndv.inl
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright (C) 2022, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#ifndef UCP_RNDV_RNDV_INL_
+#define UCP_RNDV_RNDV_INL_
```
yosefe, line 7: rndv.h already has some inline function, so either move them here, or move ucp_rndv_receive_start to rndv.h

rakhmets: Moved inline functions to rndv.inl, because it is required to include more files to .h file, otherwise.

yosefe (dismissed):
pls squash

yosefe (approved):

PR #8068, opened 2022-03-25
UCP: Do not report lane limit error for rma and bw lanes

brminich:
## What
Do not print error if UCP is not able to establish all rma/bw lanes. 

## Why ?
Currently UCX can create up to 6 lanes. When many rma|am bw lanes are requested they may not fit the requirement and UCX prints an error, like:
`ERROR   cannot add rma_bw lane - reached limit (6)`

But in this case it is not really an error, because ep is created with as many additional lanes as fit ucp ep limits

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -584,11 +584,12 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_add_lane_desc(
         const ucp_wireup_select_info_t *select_info,
         ucp_md_index_t dst_md_index, ucs_sys_device_t dst_sys_dev,
         ucp_lane_type_t lane_type, unsigned seg_size,
-        ucp_wireup_select_context_t *select_ctx)
+        ucp_wireup_select_context_t *select_ctx, int show_error)
 {
     ucp_wireup_lane_desc_t *lane_desc;
     ucp_lane_type_t lane_type_iter;
     ucp_lane_index_t lane;
+    ucs_log_level_t log_level;
 
     /* Add a new lane, but try to reuse already added lanes which are selected
      * on the same transport resources.
@@ -623,9 +624,10 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_add_lane_desc(
     }
 
     if (select_ctx->num_lanes >= UCP_MAX_LANES) {
-        ucs_error("cannot add %s lane - reached limit (%d)",
-                  ucp_lane_type_info[lane_type].short_name,
-                  select_ctx->num_lanes);
+        log_level = show_error ? UCS_LOG_LEVEL_ERROR : UCS_LOG_LEVEL_DEBUG;
+        ucs_log(log_level, "cannot add %s lane - reached limit (%d)",
+                ucp_lane_type_info[lane_type].short_name,
+                select_ctx->num_lanes);
         return UCS_ERR_EXCEEDS_LIMIT;
     }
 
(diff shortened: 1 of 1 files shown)


src/ucp/wireup/select.c
```diff
@@ -1229,8 +1234,10 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
             addr_index       = sinfo.addr_index;
             dev_index        = context->tl_rscs[rsc_index].dev_index;
             sinfo.path_index = local_dev_count[dev_index];
+            show_error       = (num_lanes == 0);
```
yosefe, line 1237: IMO no need this helper var

src/ucp/wireup/select.c
```diff
@@ -1161,7 +1165,7 @@ ucp_wireup_add_am_lane(const ucp_wireup_select_params_t *select_params,
             continue;
         }
 
-        return ucp_wireup_add_lane(select_params, am_info, UCP_LANE_TYPE_AM,
+        return ucp_wireup_add_lane(select_params, am_info, UCP_LANE_TYPE_AM, 1,
                                    select_ctx);
```
ivankochin, line 1169:
```suggestion
        return ucp_wireup_add_lane(select_params, am_info, UCP_LANE_TYPE_AM,
                                   /*show_error*/ 1, select_ctx);
```

[minor] Don't know if this is consistent with other code or not. So feel free to ignore this comment if you don't think it's a good idea.

src/ucp/wireup/select.c
```diff
@@ -655,16 +657,17 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_add_lane_desc(
 static UCS_F_NOINLINE ucs_status_t
 ucp_wireup_add_lane(const ucp_wireup_select_params_t *select_params,
                     const ucp_wireup_select_info_t *select_info,
-                    ucp_lane_type_t lane_type,
+                    ucp_lane_type_t lane_type, int is_show_error,
                     ucp_wireup_select_context_t *select_ctx)
 {
     ucp_address_entry_t *addr_list = select_params->address->address_list;
     unsigned addr_index            = select_info->addr_index;
+    int show_error                 = select_params->show_error && is_show_error;
```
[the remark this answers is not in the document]
brminich: we have to respect both values `select_params->show_error` and local `show_error`.
The flag in `select_params` is supposed to hide any error on transport selection. It is disabled when trying to select transports the first time. Then, when trying scalable tl bitmap, the flag is enabled.
The local flags is used to show whether inability to select a transport critical error or not for both tl bitmaps.

src/ucp/wireup/select.c
```diff
@@ -928,7 +931,8 @@ ucp_wireup_add_cm_lane(const ucp_wireup_select_params_t *select_params,
     /* server is not a proxy because it can create all lanes connected */
     return ucp_wireup_add_lane_desc(&select_info, UCP_NULL_RESOURCE,
                                     UCS_SYS_DEVICE_ID_UNKNOWN, UCP_LANE_TYPE_CM,
-                                    UINT_MAX, select_ctx);
+                                    UINT_MAX, select_ctx,
+                                    select_params->show_error);
```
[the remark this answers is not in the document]
brminich: ok

src/ucp/wireup/select.c
brminich: ok, can add this

src/ucp/wireup/select.c
brminich: i'd leave it, added this as a documentation instead of comments etc

yosefe (dismissed):

dmitrygx (approved):

PR #8070, opened 2022-03-28
UCS/DATASTRUCT: Check num regions before call to calloc.

rakhmets:
## What
Check that number of regions of page table is not 0 before starting to purge the table.

## Why ?
The issue was caught by memory checker tool ElectricFence.

### DIFF
--- a/src/ucs/datastruct/pgtable.c
+++ b/src/ucs/datastruct/pgtable.c
@@ -614,6 +614,11 @@ void ucs_pgtable_purge(ucs_pgtable_t *pgtable, ucs_pgt_search_callback_t cb,
     ucs_status_t status;
     unsigned i;
 
+    if (num_regions == 0) {
+        ucs_debug("purge empty page table");
+        goto out;
+    }
+
     all_regions = ucs_calloc(num_regions, sizeof(*all_regions),
                              "pgt_purge_regions");
     if (all_regions == NULL) {
@@ -642,6 +647,7 @@ void ucs_pgtable_purge(ucs_pgtable_t *pgtable, ucs_pgt_search_callback_t cb,
 
     ucs_free(all_regions);
 
+out:
     /* Page table should be totally empty */
     ucs_assert(!ucs_pgt_entry_present(&pgtable->root));
     ucs_assertv(pgtable->shift       == UCS_PGT_ADDR_SHIFT, "shift=%u", pgtable->shift);


src/ucs/datastruct/pgtable.c
```diff
@@ -614,6 +614,11 @@ void ucs_pgtable_purge(ucs_pgtable_t *pgtable, ucs_pgt_search_callback_t cb,
     ucs_status_t status;
     unsigned i;
 
+    if (num_regions == 0) {
+        ucs_diag("purge empty page table");
```
brminich, line 618: i'd better use `debug` here

dmitrygx (approved):

src/ucs/datastruct/pgtable.c
```diff
@@ -614,6 +614,11 @@ void ucs_pgtable_purge(ucs_pgtable_t *pgtable, ucs_pgt_search_callback_t cb,
     ucs_status_t status;
     unsigned i;
 
+    if (num_regions == 0) {
```
yosefe, line 617: why is this considered a problem? because of zero-size allocation?

rakhmets: Yes. 
Memory debugger (Electric Fence) reports issue on calling `calloc` with number of elements equal to zero.

dmitrygx: btw, it also could print a `ucs_warn` in case of `num_regions == 0` which is not expected

yosefe: why not? we can have empty rcache
btw, would the memory debugger report an issue if we used malloc instead of calloc?

rakhmets: I think that warning level is too high in this case.
Purging a page table with zero number of regions may occur in case of an issue at the external level. E.g. an object creates table, and immediately purges the table before adding a region, because smth went wrong. Thus, the warning/error should be reported at the external level. And here debug is enough, IMO.

rakhmets: >> btw, would the memory debugger report an issue if we used malloc instead of calloc?
Yes, it would.

yosefe (approved):

dmitrygx (approved):

yosefe:
tests passed but status not updated

PR #8061, opened 2022-03-23
UCT/IB/DEVX: invalidate MR using indirect key

evgeny-leksikov:
## What
Invalidate MR using indirect key. Another approach of #8002

## Why ?
To minizimize rkey collisions.

PR #8072, opened 2022-03-29
UCT/API: uct_md_mkey_pack_v2

evgeny-leksikov:
## What
Introduce forward compatible `uct_md_mkey_pack_v2`

## Why ?
UCT/IB/DEVX: invalidate MR using indirect key #8061 requires extra argument `flags` for `uct_md_mkey_pack`

### DIFF
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -202,6 +202,15 @@ typedef enum {
 
 
 /**
+ * @ingroup UCT_MD
+ * @brief MD memory key pack parameters field mask.
+ */
+typedef enum {
+    UCT_MD_MKEY_PACK_FIELD_FLAGS = UCS_BIT(0)  /**< flags field */
+} uct_md_mkey_pack_field_mask_t;
+
+
+/**
  * @ingroup UCT_RESOURCE
  * @brief UCT endpoint attributes field mask.
  *
@@ -238,6 +247,23 @@ typedef enum {
      */
     UCT_MD_MEM_DEREG_FLAG_INVALIDATE = UCS_BIT(0)
 } uct_md_mem_dereg_flags_t;
+
+
+typedef enum {
+    /**
+     * The flag is used indicate that remote access to a memory region
+     * associated with the remote key must fail once the memory region is
+     * deregister using @ref uct_md_mem_dereg_v2 with
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set. Using
+     * @ref uct_md_mem_dereg_v2 deregistration routine with
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set on an rkey that was
+     * generated without @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag will
+     * not function correctly, and may result in data corruption. In other words
+     * in order for @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag to function
+     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag must be set.
+     */
+    UCT_MD_MKEY_PACK_FLAG_INVALIDATE = UCS_BIT(0)
+} uct_md_mkey_pack_flags_t;
 
 
 /**
@@ -311,6 +337,22 @@ typedef struct uct_md_mem_dereg_params {
      */
     uct_completion_t             *comp;
 } uct_md_mem_dereg_params_t;
+
+
+typedef struct uct_md_mkey_pack_params {
+    /**
+     * Mask of valid fields in this structure, using bits from
+     * @ref uct_md_mkey_pack_field_mask_t. Fields not specified in this mask
+     * will be ignored. Provides ABI compatibility with respect to adding new
+     * fields.
+     */
+    uint64_t field_mask;
+
+    /**
+     * Remote key packing flags, using bits from @ref uct_md_mkey_pack_flags_t.
+     */
+    unsigned flags;
+} uct_md_mkey_pack_params_t;
 
 
 /**
@@ -390,6 +432,26 @@ uct_iface_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr);
  */
 ucs_status_t uct_md_mem_dereg_v2(uct_md_h md,
                                  const uct_md_mem_dereg_params_t *params);
+
+
+/**
+ * @ingroup UCT_MD
+ *
+ * @brief Pack a remote key.
+ *
+ * @param [in]  md          Handle to memory domain.
+ * @param [in]  memh        Pack a remote key for this memory handle.
+ * @param [in]  params      Operation parameters, see @ref
+ *                          uct_md_mkey_pack_params_t.
+ * @param [out] rkey_buffer Pointer to a buffer to hold the packed remote key.
+ *                          The size of this buffer has should be at least
+ *                          @ref uct_md_attr_t::rkey_packed_size, as returned by
+ *                          @ref uct_md_query.
+ * @return                  Error code.
+ */
+ucs_status_t uct_md_mkey_pack_v2(uct_md_h md, uct_mem_h memh,
+                                 const uct_md_mkey_pack_params_t *params,
+                                 void *rkey_buffer);
 
 
 /**
--- a/src/uct/base/uct_md.c
+++ b/src/uct/base/uct_md.c
@@ -319,9 +319,34 @@ ucs_status_t uct_config_modify(void *config, const char *name, const char *value
     return ucs_config_parser_set_value(bundle->data, bundle->table, name, value);
 }
 
+static ucs_status_t
+uct_md_mkey_pack_params_check(uct_md_h md, uct_mem_h memh, void *rkey_buffer)
+{
+    if (ENABLE_PARAMS_CHECK) {
+        return ((md != NULL) && (memh != NULL) && (rkey_buffer != NULL)) ?
+               UCS_OK : UCS_ERR_INVALID_PARAM;
+    } else {
+        return UCS_OK;
+    }
+}
+
+ucs_status_t uct_md_mkey_pack_v2(uct_md_h md, uct_mem_h memh,
+                                 const uct_md_mkey_pack_params_t *params,
+                                 void *rkey_buffer)
+{
+    ucs_status_t status = uct_md_mkey_pack_params_check(md, memh, rkey_buffer);
+
+    return (status == UCS_OK) ?
+           md->ops->mkey_pack(md, memh, params, rkey_buffer) : status;
+}
+
 ucs_status_t uct_md_mkey_pack(uct_md_h md, uct_mem_h memh, void *rkey_buffer)
 {
-    return md->ops->mkey_pack(md, memh, rkey_buffer);
+    uct_md_mkey_pack_params_t params = {
+        .field_mask = 0
+    };
+
+    return uct_md_mkey_pack_v2(md, memh, &params, rkey_buffer);
 }
 
 ucs_status_t uct_rkey_unpack(uct_component_h component, const void *rkey_buffer,
--- a/src/uct/base/uct_md.h
+++ b/src/uct/base/uct_md.h
@@ -105,8 +105,9 @@ typedef ucs_status_t (*uct_md_mem_query_func_t)(uct_md_h md,
                                                 size_t length,
                                                 uct_md_mem_attr_t *mem_attr);
 
-typedef ucs_status_t (*uct_md_mkey_pack_func_t)(uct_md_h md, uct_mem_h memh,
-                                                void *rkey_buffer);
+typedef ucs_status_t (*uct_md_mkey_pack_func_t)(
+        uct_md_h md, uct_mem_h memh, const uct_md_mkey_pack_params_t *params,
+        void *rkey_buffer);
 
 typedef int (*uct_md_is_sockaddr_accessible_func_t)(uct_md_h md,
                                                     const ucs_sock_addr_t *sockaddr,
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.c
@@ -63,8 +63,10 @@ static ucs_status_t uct_cuda_copy_md_query(uct_md_h md, uct_md_attr_t *md_attr)
     return UCS_OK;
 }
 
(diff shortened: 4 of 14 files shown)


src/uct/api/v2/uct_v2.h
```diff
@@ -240,6 +251,18 @@ typedef enum {
 } uct_md_mem_dereg_flags_t;
 
 
+typedef enum {
+    /**
+     * Should be used in case if memory handle is supposed to be invalidated in
+     * any time, for example in the meddle of RMA operation to interrupt it.
+     * Such packed key is stronger from remote access perspective.
+     * @note The memory domain must support @ref UCT_MD_FLAG_INVALIDATE
```
yosefe, line 259: Indicates that remote access to the returned remote key should fail if uct_md_mem_dereg_v2 is called with the flag UCT_MD_MEM_DEREG_FLAG_INVALIDATE on the memory handle.
If this flag is not set, using the remote key after invalidating the memory region has an undefined behavior.

src/uct/api/v2/uct_v2.h
```diff
@@ -313,6 +336,34 @@ typedef struct uct_md_mem_dereg_params {
...
 
 
+typedef struct uct_md_mkey_pack_params {
+    /**
+     * Mask of valid fields in this structure and operation flags, using
+     * bits from @ref uct_md_mkey_pack_field_mask_t. Fields not specified
+     * in this mask will be ignored. Provides ABI compatibility with respect
+     * to adding new fields.
+     */
+    uint64_t                     field_mask;
```
yosefe, line 346: too many spaces; use clang-format

src/uct/api/v2/uct_v2.h
```diff
@@ -313,6 +336,34 @@ typedef struct uct_md_mem_dereg_params {
 } uct_md_mem_dereg_params_t;
 
 
+typedef struct uct_md_mkey_pack_params {
+    /**
+     * Mask of valid fields in this structure and operation flags, using
```
yosefe, line 341: "operation flags" is irrelevant

src/uct/api/v2/uct_v2.h
```diff
@@ -313,6 +336,34 @@ typedef struct uct_md_mem_dereg_params {
...
+     * bits from @ref uct_md_mkey_pack_field_mask_t. Fields not specified
+     * in this mask will be ignored. Provides ABI compatibility with respect
+     * to adding new fields.
+     */
+    uint64_t                     field_mask;
+
+    /**
+     * Local key, whose remote key should be packed.
```
yosefe, line 349: Pack a remote key for this memory handle

src/uct/api/v2/uct_v2.h
```diff
@@ -313,6 +336,34 @@ typedef struct uct_md_mem_dereg_params {
...
+     */
+    uct_mem_h                    memh;
+
+    /**
+     * Filled with packed remote key.
```
yosefe, line 354: Pointer to a buffer to hold the packed remote key.

src/uct/api/v2/uct_v2.h
```diff
@@ -313,6 +336,34 @@ typedef struct uct_md_mem_dereg_params {
...
+     * @note The size of this buffer has to be according to
+     *       @ref uct_md_attr_t::rkey_packed_size.
+     */
+    void                         *rkey_buffer;
+
+    /**
+     * Operation specific flags, using bits from @ref uct_md_mkey_pack_flags_t.
```
yosefe, line 361: "Operation specific flags" -> "Remote key packing flags"

src/uct/base/uct_md.c
```diff
@@ -319,9 +319,36 @@ ucs_status_t uct_config_modify(void *config, const char *name, const char *value
     return ucs_config_parser_set_value(bundle->data, bundle->table, name, value);
 }
 
+static ucs_status_t
+uct_md_mkey_pack_params_check(const uct_md_mkey_pack_params_t *params)
+{
+#if ENABLE_PARAMS_CHECK
+    return ((uct_md_mkey_pack_param_memh(params) != NULL) &&
```
yosefe, line 326:
```c
return (ENABLE_PARAMS_CHECK && 
        (uct_md_mkey_pack_param_memh(params) != NULL) &&
        (uct_md_mkey_pack_param_rkey_buf(params) != NULL)) ?
       UCS_OK : UCS_ERR_INVALID_PARAM;
```

src/uct/base/uct_md.h
```diff
@@ -105,8 +105,8 @@ typedef ucs_status_t (*uct_md_mem_query_func_t)(uct_md_h md,
                                                 size_t length,
                                                 uct_md_mem_attr_t *mem_attr);
 
-typedef ucs_status_t (*uct_md_mkey_pack_func_t)(uct_md_h md, uct_mem_h memh,
-                                                void *rkey_buffer);
+typedef ucs_status_t (*uct_md_mkey_pack_func_t)(uct_md_h md,
```
brminich, line 108: should not we define a new func, like `pack_func_v2`?

yosefe: no need because v2 is only on api level. this function definition is internal.

src/uct/base/uct_md.c
evgeny-leksikov: this returns `UCS_ERR_INVALID_PARAM`  if `ENABLE_PARAMS_CHECK  == 0`
correct 1-line-condition IMO looks overcomplicated

yosefe: ok, but can we use ENABLE_PARAMS_CHECK as a runtime check instead of compile time check? with 'if' condition?
IMO code is simpler to read without ifdef

yosefe (dismissed):

yosefe:

src/uct/api/v2/uct_v2.h
```diff
@@ -293,26 +316,55 @@ typedef struct uct_md_mem_dereg_params {
...
+     * @ref uct_md_mkey_pack_field_mask_t. Fields not specified in this mask
+     * will be ignored. Provides ABI compatibility with respect to adding new
+     * fields.
+     */
+    uint64_t  field_mask;
+
+    /**
+     * Pack a remote key for this memory handle.
+     */
+    uct_mem_h memh;
```
shamisp, line 351: Based on the code the `memh` and `rkey_buffer` are not optional parameters and there for belong to the argument list and not the `uct_md_mkey_pack_params`

shamisp (changes requested):

src/uct/api/v2/uct_v2.h
```diff
@@ -240,6 +251,18 @@ typedef enum {
...
+    /**
+     * Indicates that remote access to the returned remote key should fail if
+     * @ref uct_md_mem_dereg_v2 is called with the flag
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE on the memory handle. If this flag
+     * is not set, using the remote key after invalidating the memory region has
+     * an undefined behavior.
+     */
+    UCT_MD_MKEY_PACK_FLAG_INVALIDATE = UCS_BIT(0)
+} uct_md_mkey_pack_flags_t;
+
```
shamisp, line 264: Also, I don't see UCT_MD_MKEY_PACK_FLAG_INVALIDATE used anywhere in the code so this is a bit hard to decipher.

src/uct/api/v2/uct_v2.h
```diff
@@ -240,6 +251,18 @@ typedef enum {
...
 
 
+typedef enum {
+    /**
+     * Indicates that remote access to the returned remote key should fail if
+     * @ref uct_md_mem_dereg_v2 is called with the flag
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE on the memory handle. If this flag
+     * is not set, using the remote key after invalidating the memory region has
+     * an undefined behavior.
+     */
```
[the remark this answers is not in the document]
yosefe: yes

src/uct/api/v2/uct_v2.h
yosefe: the implementation is in #8002 ; it will create an indirect key (similar concept to memory window) and destroy it to invalidate

src/uct/api/v2/uct_v2.h
evgeny-leksikov: In theory, `memh` can be replaced with something more light-weight in future for optimization purposes, as well  as `rkey_buffer` can be replaced with "in-place buffer", for example.. So, since there are options I would suggest to have it more flexible.

shamisp: `memh` is an opaque object and `key_buffer` is void *. Both provide quite a lot of flexibility in terms of what can reside behind the objects. Since we redesign the interface I suggest to bring all the proposal now.

evgeny-leksikov: ok, fixed according to your suggestion

shamisp: @evgeny-leksikov thanks. I will do another pass over the implementation patch and I will try to improve documentation.

src/uct/api/v2/uct_v2.h
```diff
@@ -316,23 +314,23 @@ typedef struct uct_md_mem_dereg_params {
      * in this mask will be ignored. Provides ABI compatibility with respect
      * to adding new fields.
      */
-    uint64_t         field_mask;
+    uint64_t                     field_mask;
```
yosefe, line 317: why changed?

evgeny-leksikov: reverted unrelated change

yosefe (dismissed):

yosefe:

src/uct/api/v2/uct_v2.h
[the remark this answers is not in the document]
shamisp: The flag is used indicate that remote access to a memory region associated with the remote key must fail once the memory region is deregister using @ref uct_md_mem_dereg_v2 and UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set. Using @ref uct_md_mem_dereg_v2  deregistration routine  with UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set while the rkey was generated without UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag set may result in data corruption. In other words in order for UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag to function the UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag must be set.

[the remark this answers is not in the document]
shamisp: @yosefe On a bit different topic. While I was reviewing `uct_md_mem_dereg_v2` I have noticed that `memh` that is non-optional parameter is passed through param/flag. I think for consistency reason it would be useful to move it to arg list.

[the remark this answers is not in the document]
evgeny-leksikov: > The flag is used indicate that remote access to a memory region associated with the remote key must fail once the memory region is deregister using @ref uct_md_mem_dereg_v2 and UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set. Using @ref uct_md_mem_dereg_v2 deregistration routine with UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set while the rkey was generated without UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag set may result in data corruption. In other words in order for UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag to function the UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag must be set.

applied

src/uct/api/v2/uct_v2.h
```diff
@@ -251,11 +251,16 @@ typedef enum {
...
 typedef enum {
     /**
-     * Indicates that remote access to the returned remote key should fail if
-     * @ref uct_md_mem_dereg_v2 is called with the flag
-     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE on the memory handle. If this flag
-     * is not set, using the remote key after invalidating the memory region has
-     * an undefined behavior.
+     * The flag is used indicate that remote access to a memory region
+     * associated with the remote key must fail once the memory region is
+     * deregister using @ref uct_md_mem_dereg_v2 and
```
yosefe, line 256: `deregister using @ref uct_md_mem_dereg_v2 with`

src/uct/api/v2/uct_v2.h
```diff
@@ -251,11 +251,16 @@ typedef enum {
...
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set. Using
+     * @ref uct_md_mem_dereg_v2 deregistration routine with
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set while the rkey was
+     * generated without @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag set may
+     * result in data corruption. In other words in order for
```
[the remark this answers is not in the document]
shamisp: @yosefe I thought access to outdate translation table will result in silent data corruption. I don't like "undefined behavior" because one the interpretation of this that your code will be "fine", which is dangerous in this case.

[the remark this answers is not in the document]
yosefe: maybe 
```
     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set. Using
     * @ref uct_md_mem_dereg_v2 deregistration routine with
     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set on an rkey that was
     * generated without @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag will
     * not function correctly, and may result in data corruption.
```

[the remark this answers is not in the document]
shamisp: Agreed - I think it is better. 

Can we cache the UCT_MD_MKEY_PACK_FLAG_INVALIDATE status on memh and assert if UCT_MD_MEM_DEREG_FLAG_INVALIDATE is set but UCT_MD_MKEY_PACK_FLAG_INVALIDATE  was not used during pack ? I guess there are some cases when the rkey is never packet but you can capture all states and raise the assert only if key was packet.

yosefe (approved):

src/uct/api/v2/uct_v2.h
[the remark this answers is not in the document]
yosefe: yes, @evgeny-leksikov will add it as part of the implementation of IB (only MD that supports invalidation for now).
other MDs do not support invalidation and will fail with UNSUPPORTED anwyay.

evgeny-leksikov:
> @evgeny-leksikov can you pls squash?

done

yosefe (approved):

PR #8075, opened 2022-03-29
UCT/IB/UD: Fix connection matching

dmitrygx:
## What

Fix connection matching in UD transport.

## Why ?

If user destroyed UCT EP, we shouldn't disconenct it, because it could be used for RX operations.

## How ?

1. Introduce TX and RX flags.
2. If both TX and RX flags are set, endpoint shouldn't be added to CEP.
3. If EP has TX-only - add to expected connection matching queue, if has RX-only - add to unexpected one.
4. Send `FIN` packet if closing won't use TX capability anymore on EP.
5. Handle `FIN` packet to remove RX capability on a EP-receiver.


dmitrygx:

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -596,9 +603,14 @@ ucs_status_t uct_ud_ep_create_connected_common(const uct_ep_params_t *ep_params,
         goto out;
     }
 
+    ep->flags &= ~UCT_UD_EP_FLAG_CONNECT_TO_EP; /* by default, the endpoint is
```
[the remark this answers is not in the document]
dmitrygx: fixed

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -724,15 +739,18 @@ static uct_ud_ep_t *uct_ud_ep_create_passive(uct_ud_iface_t *iface,
         return NULL;
     }
 
+    ep->flags &= ~UCT_UD_EP_FLAG_CONNECT_TO_EP; /* by default, the endpoint is
```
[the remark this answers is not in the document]
dmitrygx: fixed

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -137,30 +162,50 @@ uct_ud_ep_t *uct_ud_iface_cep_get_ep(uct_ud_iface_t *iface,
...
+        ucs_assertv(ep->flags & (UCT_UD_EP_FLAG_TX | UCT_UD_EP_FLAG_RX),
+                    "ep %p: flags 0x%x", ep, ep->flags);
     }
 
     return ep;
 }
 
+int uct_ud_iface_cep_has_ep(uct_ud_ep_t *ep)
+{
+    int res = !!(ep->flags & (UCT_UD_EP_FLAG_TX | UCT_UD_EP_FLAG_RX)) &&
```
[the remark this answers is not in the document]
dmitrygx: to have `0` or `1` instead of the result of `ep->flags & (UCT_UD_EP_FLAG_TX | UCT_UD_EP_FLAG_RX)` which could be `UCT_UD_EP_FLAG_TX ` or `UCT_UD_EP_FLAG_RX`  or both.

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1753,14 +1789,34 @@ void uct_ud_ep_disconnect(uct_ep_h tl_ep)
...
+         * exists there */
+        uct_ud_iface_cep_remove_ep(iface, ep);
+        ep->flags &= ~UCT_UD_EP_FLAG_TX;
+
+        uct_ud_ep_ctl_op_add(iface, ep, UCT_UD_EP_OP_FIN);
+
+        if (ep->flags & UCT_UD_EP_FLAG_RX) {
+            uct_ud_iface_cep_insert_ready_ep(iface, ep);
+            /* if the endpoint is marked to be able receive data from a peer,
+             * don't schedule a timer to destroy the endpoint */
```
brminich, line 1804: maybe: `Don't destroy ep if it has receive capability`

test/gtest/uct/ib/test_ud_ds.cc
```diff
@@ -27,6 +40,15 @@ class test_ud_ds : public uct_test {
         m_e2 = create_entity(0);
         m_entities.push_back(m_e2);
 
+        // overwrite purge_cb to remove and destroy UD endpoints from
```
brminich, line 43: plz start with upper case

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -137,30 +162,50 @@ uct_ud_ep_t *uct_ud_iface_cep_get_ep(uct_ud_iface_t *iface,
...
     if (conn_match == NULL) {
         return NULL;
     }
 
     ep = ucs_container_of(conn_match, uct_ud_ep_t, conn_match);
     ucs_assert(ep->flags & UCT_UD_EP_FLAG_ON_CEP);
 
-    if (is_private) {
+    if (is_rx_only) {
+        ucs_assertv((ep->flags & (UCT_UD_EP_FLAG_TX | UCT_UD_EP_FLAG_RX)) ==
```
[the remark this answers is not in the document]
dmitrygx: done

src/uct/ib/ud/base/ud_def.h
```diff
@@ -58,6 +58,7 @@ enum {
     UCT_UD_PACKET_FLAG_NACK    = UCS_BIT(27),
     UCT_UD_PACKET_FLAG_PUT     = UCS_BIT(28),
     UCT_UD_PACKET_FLAG_CTL     = UCS_BIT(29),
+    UCT_UD_PACKET_FLAG_FIN     = UCS_BIT(30),
```
yosefe, line 61: maybe send FIN packet by "control packet extended header", to not consume a reserved bit from the header? i.e add UCT_UD_PACKET_FIN and use uct_ud_ctl_hdr_t.
otherwise, pls update the ascii art below to represent the new FIN packet

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -257,8 +259,10 @@ static unsigned uct_ud_ep_deferred_timeout_handler(void *arg)
         return 0;
     }
 
-    if (ep->flags & UCT_UD_EP_FLAG_PRIVATE) {
+    if (!(ep->flags & UCT_UD_EP_FLAG_TX)) {
         ucs_assert(ucs_queue_is_empty(&ep->tx.window));
+        uct_ud_iface_cep_remove_ep(iface, ep);
+        ep->flags &= ~UCT_UD_EP_FLAG_RX;
```
yosefe, line 265: don't we need to do it even if ep->flags & UCT_UD_EP_FLAG_TX is 0?
so if we have timeout error no need to send FIN

src/uct/ib/ud/base/ud_ep.h
```diff
@@ -191,30 +192,35 @@ enum {
...
+    UCT_UD_EP_FLAG_IN_PENDING        = UCS_BIT(12), /* Endpoint is currently executing the
+                                                       pending queue */
+    UCT_UD_EP_FLAG_ON_CEP            = UCS_BIT(13),
 #else
-    UCT_UD_EP_FLAG_IN_PENDING        = 0
+    UCT_UD_EP_FLAG_IN_PENDING        = 0,
+    UCT_UD_EP_FLAG_ON_CEP            = 0,
 #endif
+
+    UCT_UD_EP_FLAG_CAPS              = UCT_UD_EP_FLAG_TX | UCT_UD_EP_FLAG_RX
```
yosefe, line 220: maybe UCT_UD_EP_FLAG_TX_RX

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -596,9 +603,13 @@ ucs_status_t uct_ud_ep_create_connected_common(const uct_ep_params_t *ep_params,
         goto out;
     }
 
+    /* Remove RX and CONNECT_TO_EP flags, since the endpoint is created as full
+     * duplex and p2p by default, but this endpoint should be TX-only */
+    ep->flags &= ~(UCT_UD_EP_FLAG_CONNECT_TO_EP | UCT_UD_EP_FLAG_RX);
```
yosefe, line 608: maybe pass the right flags to uct_ud_ep_create, instead of adjusting them later?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -633,10 +644,13 @@ ucs_status_t uct_ud_ep_create_connected_common(const uct_ep_params_t *ep_params,
...
     return status;
 
+err_cep_remove_ep:
+    uct_ud_iface_cep_remove_ep(iface, ep);
 err_ep_disconnect:
     uct_ud_ep_disconnect_from_iface(ep);
-err_ep_destroy:
-    uct_ep_destroy(&ep->super.super);
+err_ep_free:
+    uct_iface_invoke_ops_func(&iface->super, uct_ud_iface_ops_t,
```
yosefe, line 652: maybe introduce uct_ud_ep_free(ep) - it's used from multiple places

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -977,6 +995,20 @@ void uct_ud_ep_process_rx(uct_ud_iface_t *iface, uct_ud_neth_t *neth, unsigned b
     }
 
     if (ucs_unlikely(!is_am)) {
+        if ((neth->packet_type & UCT_UD_PACKET_FLAG_FIN) &&
+            (ep->flags & UCT_UD_EP_FLAG_RX)) {
+            uct_ud_iface_cep_remove_ep(iface, ep);
```
yosefe, line 1000: pls add comment what this code block is doing

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -977,6 +995,20 @@ void uct_ud_ep_process_rx(uct_ud_iface_t *iface, uct_ud_neth_t *neth, unsigned b
...
+
+            if (!(ep->flags & UCT_UD_EP_FLAG_TX)) {
+                ep->flags &= ~UCT_UD_EP_FLAG_RX;
```
yosefe, line 1003: can we do it before if/else? dup line 1006

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1753,14 +1789,33 @@ void uct_ud_ep_disconnect(uct_ep_h tl_ep)
      * to not invoke user callbacks for the endpoint which was destroyed */
     uct_ud_ep_purge(ep, UCS_ERR_CANCELED);
 
+    if (!(ep->flags & UCT_UD_EP_FLAG_CONNECT_TO_EP) &&
+        (ep->flags & UCT_UD_EP_FLAG_TX)) {
+        /* remove the endpoint from the connection matching context if it
```
yosefe, line 1794: start comment with capital case

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1753,14 +1789,33 @@ void uct_ud_ep_disconnect(uct_ep_h tl_ep)
...
+        uct_ud_ep_ctl_op_add(iface, ep, UCT_UD_EP_OP_FIN);
+
+        if (ep->flags & UCT_UD_EP_FLAG_RX) {
+            uct_ud_iface_cep_insert_ready_ep(iface, ep);
+            /* Don't destroy ep if it has RX capability */
+            goto out;
+        }
+    }
+
+    ucs_assert(!uct_ud_iface_cep_has_ep(ep));
```
yosefe, line 1808: isn't this already checked by ep destructor?

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -137,30 +161,49 @@ uct_ud_ep_t *uct_ud_iface_cep_get_ep(uct_ud_iface_t *iface,
...
         return NULL;
     }
 
     ep = ucs_container_of(conn_match, uct_ud_ep_t, conn_match);
     ucs_assert(ep->flags & UCT_UD_EP_FLAG_ON_CEP);
 
-    if (is_private) {
+    if (is_rx_only) {
+        ucs_assertv((ep->flags & UCT_UD_EP_FLAG_CAPS) == UCT_UD_EP_FLAG_RX,
+                    "ep %p: flags 0x%x", ep, ep->flags);
```
yosefe, line 174: ep=%p flags=0x%x

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -137,30 +161,49 @@ uct_ud_ep_t *uct_ud_iface_cep_get_ep(uct_ud_iface_t *iface,
...
         ep->flags &= ~UCT_UD_EP_FLAG_ON_CEP;
+    } else {
+        ucs_assertv(ep->flags & UCT_UD_EP_FLAG_CAPS, "ep %p: flags 0x%x", ep,
```
yosefe, line 177: ep=%p flags=0x%x

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -137,30 +161,49 @@ uct_ud_ep_t *uct_ud_iface_cep_get_ep(uct_ud_iface_t *iface,
...
     return ep;
 }
 
+int uct_ud_iface_cep_has_ep(uct_ud_ep_t *ep)
+{
+    int res = (ep->flags & UCT_UD_EP_FLAG_CAPS) &&
+              !(ep->flags & UCT_UD_EP_FLAG_CONNECT_TO_EP);
+
+    ucs_assertv(res == !!(ep->flags & UCT_UD_EP_FLAG_ON_CEP),
+                "res=%d on_cep=%d flags=0x%x", res,
```
yosefe, line 190: ep=%p flags=0x%x res=%d

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -619,7 +663,7 @@ static void uct_ud_iface_delete_eps(uct_ud_iface_t *iface)
     int i;
 
     ucs_ptr_array_for_each(ep, i, &iface->eps) {
-        ucs_assert(!(ep->flags & UCT_UD_EP_FLAG_ON_CEP));
+        ucs_assert(!uct_ud_iface_cep_has_ep(ep));
```
yosefe, line 666: why changed?

test/gtest/ucs/test_conn_match.cc
```diff
@@ -104,8 +104,9 @@ class test_conn_match : public ucs::test {
 
     void insert(const void *dest_address, ucs_conn_sn_t conn_sn,
                 ucs_conn_match_queue_type_t queue_type, conn_elem_t &elem) {
-        ucs_conn_match_insert(&m_conn_match_ctx, dest_address, conn_sn,
-                              &elem.elem, queue_type);
+        int ret = ucs_conn_match_insert(&m_conn_match_ctx, dest_address, conn_sn,
+                                        &elem.elem, queue_type);
+        ASSERT_EQ(1, ret);
```
yosefe, line 109: maybe ASSERT_TRUE?

test/gtest/uct/ib/test_ud_ds.cc
```diff
@@ -27,6 +40,15 @@ class test_ud_ds : public uct_test {
         m_e2 = create_entity(0);
         m_entities.push_back(m_e2);
 
+        // Overwrite purge_cb to remove and destroy UD endpoints from
+        // connection matching without checking flags set on the endpoint
+        uct_ud_iface_t *ud_iface1              =
```
yosefe, line 45: auto

test/gtest/uct/ib/test_ud_ds.cc
```diff
@@ -145,48 +175,56 @@ void test_ud_ds::insert_ep(entity *e, uct_ib_address_t *ib_addr,
...
+
+    ep(e, ep_index)->flags &= ~UCT_UD_EP_FLAG_CONNECT_TO_EP;
+    if (is_ep_internal) {
+        ep(e, ep_index)->flags &= ~UCT_UD_EP_FLAG_TX;
+    } else {
+        ep(e, ep_index)->flags &= ~UCT_UD_EP_FLAG_RX;
     }
 
     insert_ep(e, ib_addr, if_addr, conn_sn, ep(e, ep_index));
+
```
yosefe, line 208: minor: remove space line

test/gtest/uct/ib/test_ud_ds.cc
```diff
@@ -145,48 +175,56 @@ void test_ud_ds::insert_ep(entity *e, uct_ib_address_t *ib_addr,
...
     check_ep = uct_ud_iface_cep_get_ep(iface(e), ib_addr, if_addr, 0,
-                                       conn_sn, ep_private);
+                                       conn_sn, is_ep_internal);
     EXPECT_TRUE(check_ep != NULL);
     EXPECT_EQ(conn_sn, check_ep->conn_sn);
-    if (!ep_private) {
-        EXPECT_TRUE(check_ep->flags & UCT_UD_EP_FLAG_ON_CEP);
-        EXPECT_TRUE(!(check_ep->flags & UCT_UD_EP_FLAG_PRIVATE));
+    if (is_ep_internal) {
+        EXPECT_TRUE(!(check_ep->flags & UCT_UD_EP_FLAG_TX));
```
yosefe, line 224: maybe use EXPECT_FALSE

src/uct/ib/ud/base/ud_def.h
dmitrygx: done

src/uct/ib/ud/base/ud_ep.c
dmitrygx: done

src/uct/ib/ud/base/ud_ep.c
dmitrygx: done

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -262,11 +262,18 @@ static unsigned uct_ud_ep_deferred_timeout_handler(void *arg)
...
         ucs_assert(ucs_queue_is_empty(&ep->tx.window));
         uct_ud_iface_cep_remove_ep(iface, ep);
         ep->flags &= ~UCT_UD_EP_FLAG_RX;
         uct_ep_destroy(&ep->super.super);
         goto out;
+    } else if ((ep->flags & UCT_UD_EP_FLAG_RX) &&
+               !(ep->flags & UCT_UD_EP_FLAG_CONNECT_TO_EP)) {
+        uct_ud_iface_cep_remove_ep(iface, ep);
+        ep->flags &= ~UCT_UD_EP_FLAG_RX;
+        uct_ud_iface_cep_insert_ready_ep(iface, ep);
```
yosefe, line 275: can we unite with the if in lines 265-268? there is some duplication

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -561,9 +569,19 @@ static ucs_status_t uct_ud_ep_create(uct_ud_iface_t *iface, int path_index,
     *new_ep_p            = ucs_derived_of(new_ep_h, uct_ud_ep_t);
     (*new_ep_p)->conn_sn = conn_sn;
 
+    /* Adjust flags, since the endpoint is created as full duplex and p2p by
```
yosefe, line 572: maybe the endpoint should be created without any flags by default?
or use a lower-level function here (not uct_ep_create), so no need to add and then remove flags

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -875,28 +884,52 @@ static void uct_ud_ep_rx_ctl(uct_ud_iface_t *iface, uct_ud_ep_t *ep,
...
-    if (uct_ud_ep_is_connected(ep)) {
-        ucs_assertv_always(ep->dest_ep_id == ctl->conn_rep.src_ep_id,
-                           "ep=%p [id=%d dest_ep_id=%d flags=0x%x] "
-                           "crep [neth->dest=%d dst_ep_id=%d src_ep_id=%d]",
-                           ep, ep->ep_id, ep->dest_ep_id, ep->path_index, ep->flags,
-                           uct_ud_neth_get_dest_id(neth), ctl->conn_rep.src_ep_id);
-    }
+    switch (ctl->type) {
+    case UCT_UD_PACKET_FIN:
+        if (ep->flags & UCT_UD_EP_FLAG_RX) {
```
yosefe, line 890: can we put handling of each case to separate helper function?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -875,28 +884,52 @@ static void uct_ud_ep_rx_ctl(uct_ud_iface_t *iface, uct_ud_ep_t *ep,
...
+            /* Got FIN packet which means that the endpoint won't be used
+             * to receive data */
+            uct_ud_iface_cep_remove_ep(iface, ep);
+            ep->flags &= ~UCT_UD_EP_FLAG_RX;
```
yosefe, line 894: this code seems similar to timeout handler, can we unite?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1281,13 +1301,47 @@ static uct_ud_send_skb_t *uct_ud_ep_prepare_crep(uct_ud_ep_t *ep)
...
+    neth = skb->neth;
+    uct_ud_neth_init_data(ep, neth);
+
+    neth->packet_type       = ep->dest_ep_id | UCT_UD_PACKET_FLAG_CTL;
+    fin                     = (uct_ud_ctl_hdr_t*)(neth + 1);
+    fin->type               = UCT_UD_PACKET_FIN;
+
+    uct_ud_peer_name(ucs_unaligned_ptr(&fin->peer));
+
+    skb->len = sizeof(*neth) + sizeof(*fin);
```
yosefe, line 1330: can we unite common code with uct_ud_ep_prepare_crep?
the common function will initialize common neth fields and return pointer the the control packet (fin/crep). then, we fill the control packet fields in the separate functions, such as fin->type, crep->conn_rep, etc.

src/uct/ib/ud/base/ud_ep.c
yosefe: ok

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -170,11 +170,11 @@ uct_ud_ep_t *uct_ud_iface_cep_get_ep(uct_ud_iface_t *iface,
     ucs_assert(ep->flags & UCT_UD_EP_FLAG_ON_CEP);
 
     if (is_rx_only) {
-        ucs_assertv((ep->flags & UCT_UD_EP_FLAG_CAPS) == UCT_UD_EP_FLAG_RX,
-                    "ep %p: flags 0x%x", ep, ep->flags);
+        ucs_assertv((ep->flags & UCT_UD_EP_FLAG_TX_RX) == UCT_UD_EP_FLAG_RX,
+                    "ep %p flags 0x%x", ep, ep->flags);
```
yosefe, line 174: "ep=%p flags=0x%x"
currently the common style with "assertv" is use equal sign

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -170,11 +170,11 @@ uct_ud_ep_t *uct_ud_iface_cep_get_ep(uct_ud_iface_t *iface,
...
         ep->flags &= ~UCT_UD_EP_FLAG_ON_CEP;
     } else {
-        ucs_assertv(ep->flags & UCT_UD_EP_FLAG_CAPS, "ep %p: flags 0x%x", ep,
+        ucs_assertv(ep->flags & UCT_UD_EP_FLAG_TX_RX, "ep %p flags 0x%x", ep,
```
yosefe, line 177: "ep=%p flags=0x%x"
currently the common style with "assertv" is use equal sign

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -248,6 +248,26 @@ static void uct_ud_ep_purge(uct_ud_ep_t *ep, ucs_status_t status)
     ucs_assert(ucs_queue_is_empty(&ep->tx.window));
 }
 
+static int uct_ud_ep_handle_peer_disconnect(uct_ud_ep_t *ep)
```
yosefe, line 251: why not return void?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -547,10 +556,89 @@ static ucs_status_t uct_ud_ep_disconnect_from_iface(uct_ud_ep_t *ep)
...
+
+    neth->packet_type = ep->dest_ep_id | extra_packet_type |
+                        UCT_UD_PACKET_FLAG_CTL;
+    ctl               = (uct_ud_ctl_hdr_t*)(neth + 1);
+    ctl->type         = ctl_type;
+
+    uct_ud_peer_name(ucs_unaligned_ptr(&ctl->peer));
+
+    *ctl_p = ctl;
+
```
yosefe, line 588: minor: remove spaceline

src/uct/ib/ud/base/ud_ep.c
dmitrygx: added comment

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -248,6 +248,26 @@ static void uct_ud_ep_purge(uct_ud_ep_t *ep, ucs_status_t status)
     ucs_assert(ucs_queue_is_empty(&ep->tx.window));
 }
 
+static int uct_ud_ep_handle_peer_disconnect(uct_ud_ep_t *ep)
+{
+    uct_ud_iface_t *iface = ucs_derived_of(ep->super.super.iface,
+                                           uct_ud_iface_t);
+
+    if (ep->flags & UCT_UD_EP_FLAG_RX) {
```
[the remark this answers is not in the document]
dmitrygx: done

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -248,6 +248,26 @@ static void uct_ud_ep_purge(uct_ud_ep_t *ep, ucs_status_t status)
...
+        uct_ud_iface_cep_remove_ep(iface, ep);
+        ep->flags &= ~UCT_UD_EP_FLAG_RX;
+
+        if (!(ep->flags & UCT_UD_EP_FLAG_TX)) {
+            uct_ep_destroy(&ep->super.super);
+            return 1;
+        }
+
+        uct_ud_iface_cep_insert_ready_ep(iface, ep);
```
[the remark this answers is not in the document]
dmitrygx: need to return `int`

src/uct/ib/ud/base/ud_ep.c
dmitrygx: done

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -248,23 +248,27 @@ static void uct_ud_ep_purge(uct_ud_ep_t *ep, ucs_status_t status)
...
+    ep->flags &= ~UCT_UD_EP_FLAG_RX;
 
-        uct_ud_iface_cep_insert_ready_ep(iface, ep);
+    if (!(ep->flags & UCT_UD_EP_FLAG_TX)) {
+        uct_ep_destroy(&ep->super.super);
+        return 1;
     }
 
+    uct_ud_iface_cep_insert_ready_ep(iface, ep);
+
```
yosefe, line 271: minor: remove spaceline

src/uct/ib/ud/base/ud_def.h
```diff
@@ -132,12 +132,12 @@ enum {
     UCT_UD_SEND_SKB_FLAG_RESENDING  = UCS_BIT(3), /* An active control skb refers to this skb */
 
 #if UCS_ENABLE_ASSERT
-    UCT_UD_SEND_SKB_FLAG_CTL_ACK    = UCS_BIT(5), /* This is a control-ack skb */
-    UCT_UD_SEND_SKB_FLAG_CTL_RESEND = UCS_BIT(6), /* This is a control-resend rsb */
-    UCT_UD_SEND_SKB_FLAG_INVALID    = UCS_BIT(7)  /* skb is released */
+    UCT_UD_SEND_SKB_FLAG_CTL        = UCS_BIT(5), /* This is a control-ack skb */
+    UCT_UD_SEND_SKB_FLAG_CTL_RESEND = UCS_BIT(6), /* This is a control-resend skb */
+    UCT_UD_SEND_SKB_FLAG_INVALID    = UCS_BIT(8)  /* skb is released */
```
dmitrygx, line 137: need to roll back

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1309,38 +1305,20 @@ static uct_ud_send_skb_t *uct_ud_ep_prepare_crep(uct_ud_ep_t *ep)
...
-        return;
-    }
-
-    uct_ud_ep_set_state(ep, state);
-    uct_ud_ep_ctl_op_del(ep, op);
-    uct_ud_iface_send_ctl(iface, ep, skb, NULL, 0,
-                          UCT_UD_IFACE_SEND_CTL_FLAG_SOLICITED, 1);
-    uct_ud_iface_complete_tx_skb(iface, ep, skb);
-}
+}\
```
dmitrygx, line 1321: remove

dmitrygx:

dmitrygx:

src/uct/ib/ud/base/ud_ep.c
dmitrygx: done

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -830,41 +956,38 @@ static void uct_ud_ep_rx_creq(uct_ud_iface_t *iface, uct_ud_neth_t *neth)
...
 
     /* CREQ must have same psn */
-    ucs_assertv_always(ep->rx.ooo_pkts.head_sn == neth->psn,
-                       "iface=%p ep=%p conn_sn=%d ep_id=%d, dest_ep_id=%d"
-                       " rx_psn=%u neth_psn=%u ep_flags=0x%x ctl_ops=0x%x"
-                       " rx_creq_count=%d",
-                       iface, ep, ep->conn_sn, ep->ep_id, ep->dest_ep_id,
-                       ep->rx.ooo_pkts.head_sn, neth->psn, ep->flags,
-                       ep->tx.pending.ops, ep->rx_creq_count);
+    if (!uct_ud_ep_is_connected(ep)) {
```
yosefe, line 963: why would it be connected?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -830,41 +956,38 @@ static void uct_ud_ep_rx_creq(uct_ud_iface_t *iface, uct_ud_neth_t *neth)
...
+                        ep->rx.ooo_pkts.head_sn, neth->psn, ep->flags,
+                         ep->tx.pending.ops, ep->rx_creq_count);
+    }
 
     /* Schedule connection reply op */
-    UCT_UD_EP_HOOK_CALL_RX(ep, neth, sizeof(*neth) + sizeof(*ctl));
-    if (uct_ud_ep_ctl_op_check(ep, UCT_UD_EP_OP_CREQ)) {
-        uct_ud_ep_set_state(ep, UCT_UD_EP_FLAG_CREQ_NOTSENT);
-    }
-    uct_ud_ep_ctl_op_del(ep, UCT_UD_EP_OP_CREQ);
```
yosefe, line 850: why removed?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -989,12 +1083,19 @@ void uct_ud_ep_process_rx(uct_ud_iface_t *iface, uct_ud_neth_t *neth, unsigned b
...
         }
+
         if (neth->packet_type & UCT_UD_PACKET_FLAG_CTL) {
-            uct_ud_ep_rx_ctl(iface, ep, neth, skb);
-            goto out;
+            uct_ud_ep_rx_ctl(iface, ep, neth);
+        } else if (neth->packet_type & UCT_UD_PACKET_FLAG_PUT) {
+            /* TODO: remove once ucp implements put */
+            uct_ud_ep_rx_put(neth, byte_len);
+        } else {
```
yosefe, line 1092: why needed to refactor this?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1213,52 +1312,32 @@ ucs_status_t uct_ud_ep_check(uct_ep_h tl_ep, unsigned flags, uct_completion_t *c
...
     uct_ud_send_skb_t *skb;
-    uct_ud_neth_t *neth;
     uct_ud_ctl_hdr_t *crep;
-    uct_ud_iface_t *iface = ucs_derived_of(ep->super.super.iface, uct_ud_iface_t);
 
     ucs_assert_always(ep->dest_ep_id != UCT_UD_EP_NULL_ID);
     ucs_assert_always(ep->ep_id != UCT_UD_EP_NULL_ID);
 
-    /* Check that CREQ is neither scheduled nor waiting for CREP ack */
-    ucs_assertv_always(!uct_ud_ep_ctl_op_check(ep, UCT_UD_EP_OP_CREQ) &&
```
yosefe, line 1225: seems this assert is deleted?

yosefe (commented):
in general, maybe we don't need these fixes for keepalive?
I would expect that when using CM, ud ep will be connected in p2p mode, so it will not use conn_match flow

dmitrygx:
> in general, maybe we don't need these fixes for keepalive?
> I would expect that when using CM, ud ep will be connected in p2p mode, so it will not use conn_match flow

we need it for CONNECT_TO_WORKER mode anyway.
btw, it was bug in Keepalive mode selection (fixed by #8113), we used AM-based keepalive - so we created reply EP to send `EP_REMOVED` with `CONNECT_TO_WORKER` mode. I'm testing #8113 to check whether it is enough to fix "ACKed and TX PSNs mismatch".

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -241,11 +241,36 @@ static void uct_ud_ep_purge(uct_ud_ep_t *ep, ucs_status_t status)
...
+
+    uct_ud_iface_cep_remove_ep(iface, ep);
+    ep->flags &= ~UCT_UD_EP_FLAG_RX;
+
+    if (!(ep->flags & UCT_UD_EP_FLAG_TX)) {
+        uct_ep_destroy(&ep->super.super);
+        return 1;
+    }
+
+    uct_ud_iface_cep_insert_ready_ep(iface, ep);
```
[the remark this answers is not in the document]
dmitrygx: after detecting error or getting FIN, we still have TX capability and the endpoint should be in CEP

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -585,24 +698,25 @@ ucs_status_t uct_ud_ep_create_connected_common(const uct_ep_params_t *ep_params,
...
-                                                path_index, conn_sn, ep);
+        uct_ud_ep_set_state(ep, UCT_UD_EP_FLAG_CREQ_NOTSENT |
+                                UCT_UD_EP_FLAG_TX);
+        status = uct_ud_iface_cep_insert_ep(iface, ib_addr, if_addr,
+                                            path_index, conn_sn, ep);
         if (status != UCS_OK) {
             goto err_ep_disconnect;
         }
-        goto out_set_ep;
+        goto out_send_creq_and_set_ep;
```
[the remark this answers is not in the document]
dmitrygx: peer's endpoint should have RX capability, sending CREQ ensures it

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -781,15 +891,31 @@ static void uct_ud_ep_rx_creq(uct_ud_iface_t *iface, uct_ud_neth_t *neth)
                                  ctl->conn_req.path_index,
                                  ctl->conn_req.conn_sn, 0);
     if (ep == NULL) {
-        ep = uct_ud_ep_create_passive(iface, ctl);
-        if (ep == NULL) {
-            ucs_diag("iface %p: failed to create the passive endpoint", iface);
-            return;
-        }
+        if ((ctl->conn_req.dest_ep_id != UCT_UD_EP_NULL_ID) &&
+            ucs_ptr_array_lookup(&iface->eps, ctl->conn_req.dest_ep_id, ep)) {
```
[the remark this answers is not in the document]
dmitrygx: to check whether we have corresponding endpoint in PTR array or not.
it may happen that endpoint was destroy by use prior receiving CREQ from the peer - so checking this endpoint in PTR array helps to find such endpoits if they are not in CEP anymore.

src/uct/ib/ud/base/ud_ep.c
dmitrygx: they may be connected if we have alredy sent CREQ to a peer.
but receiving new CREQ means that peer wants update us to have RX capability.

src/uct/ib/ud/base/ud_ep.c
dmitrygx: we shouldn't delete sending CREQ - peer should know that we will send him

src/uct/ib/ud/base/ud_ep.c
dmitrygx: moved handling PUT back.
it was refactored to put CREQ/CREP/FIN to frag list too - to allow sending CREQ/FIN using previous CREQ/CREP code at the middle of the endpoint's life .

src/uct/ib/ud/base/ud_ep.c
dmitrygx: right, it is not needed anymore - we could send CREP even if CREQ was sent

src/uct/ib/ud/base/ud_iface.h
```diff
@@ -225,6 +227,7 @@ struct uct_ud_ctl_hdr {
             uct_ud_ep_addr_t    ep_addr;
             uct_ud_ep_conn_sn_t conn_sn;
             uint8_t             path_index;
+            uint32_t            dest_ep_id;
```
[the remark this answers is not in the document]
dmitrygx: this dest_ep_id is needed to be able to use it for handling CREQ in case if UD endpoint couldn't be found in CEP, but exists in PTR array.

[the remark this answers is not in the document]
yosefe: we already have dest id in neth, why need here too?

src/uct/ib/ud/base/ud_ep.c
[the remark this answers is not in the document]
yosefe: why it should be in cep?
I would expect it to be destroyed directly by user calling uct_ep_destroy

src/uct/ib/ud/base/ud_ep.c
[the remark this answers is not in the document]
yosefe: so now both sides send creq and crep?

src/uct/ib/ud/base/ud_ep.c
yosefe: ok

src/uct/ib/ud/base/ud_ep.c
yosefe: ok

src/uct/ib/ud/base/ud_ep.c
yosefe: ok
PS, will we send a CREP if already sent a CREQ?

yosefe:
> > in general, maybe we don't need these fixes for keepalive?
> > I would expect that when using CM, ud ep will be connected in p2p mode, so it will not use conn_match flow
> 
> we need it for CONNECT_TO_WORKER mode anyway. btw, it was bug in Keepalive mode selection (fixed by #8113), we used AM-based keepalive - so we created reply EP to send `EP_REMOVED` with `CONNECT_TO_WORKER` mode. I'm testing #8113 to check whether it is enough to fix "ACKed and TX PSNs mismatch".

Can we fix the p2p keepalive prior to fixing CONNECT_TO_WORKER mode, to make io_demo pass?

src/uct/ib/ud/base/ud_ep.c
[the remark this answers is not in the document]
dmitrygx: if endpoint has TX (i.e. `uct_ep_destroy` wasn't called by a user), it should be in CEP

src/uct/ib/ud/base/ud_iface.h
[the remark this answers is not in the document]
dmitrygx: to not have another branch which handle CREQ when `neth->packet_type` has real ep_id.
anyway we break backward compatibility by adding new packet type - we have `ucs_fatal` if received unknown packet.
btw, shall we remove this `ucs_fatal` for future?

dmitrygx:
> > > in general, maybe we don't need these fixes for keepalive?
> > > I would expect that when using CM, ud ep will be connected in p2p mode, so it will not use conn_match flow
> > 
> > 
> > we need it for CONNECT_TO_WORKER mode anyway. btw, it was bug in Keepalive mode selection (fixed by #8113), we used AM-based keepalive - so we created reply EP to send `EP_REMOVED` with `CONNECT_TO_WORKER` mode. I'm testing #8113 to check whether it is enough to fix "ACKed and TX PSNs mismatch".
> 
> Can we fix the p2p keepalive prior to fixing CONNECT_TO_WORKER mode, to make io_demo pass?

@yosefe @brminich as discussed offline, we need to add dropping packets if endpoint was disconnected, but we should fix connection matching to implement it

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1006,9 +1107,16 @@ void uct_ud_ep_process_rx(uct_ud_iface_t *iface, uct_ud_neth_t *neth, unsigned b
         goto out;
     }
 
-    if (ucs_unlikely(!is_am && (neth->packet_type & UCT_UD_PACKET_FLAG_PUT))) {
-        /* TODO: remove once ucp implements put */
-        uct_ud_ep_rx_put(neth, byte_len);
+    if (ucs_unlikely(!is_am)) {
+        if (neth->packet_type & UCT_UD_PACKET_FLAG_PUT) {
+            /* TODO: remove once ucp implements put */
```
yosefe, line 1112: remove the comment

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -881,92 +1001,76 @@ static void uct_ud_ep_rx_ctl(uct_ud_iface_t *iface, uct_ud_ep_t *ep,
...
+uct_ud_ep_process_unexp_packet(uct_ud_iface_t *iface, uct_ud_ep_t *ep,
+                               uct_ud_neth_t *neth, unsigned length)
+{
+    char buf[256] = {0};
 
-    skb->len = sizeof(*neth) + sizeof(*creq) + iface->super.addr_size;
-    return skb;
+    uct_ud_dump_packet(&iface->super.super, UCT_AM_TRACE_TYPE_RECV, neth,
+                       length, length, buf, sizeof(buf) - 1);
+    ucs_fatal("ep=%p: unexpected non-AM packet: [%s]", ep, buf);
```
yosefe, line 1036: should be DIAG that dropping the packet

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -881,92 +1001,76 @@ static void uct_ud_ep_rx_ctl(uct_ud_iface_t *iface, uct_ud_ep_t *ep,
...
+        ucs_fatal("unexpected control packet type - %d", ctl->type);
     }
+}
 
-    uct_ud_peer_name(ucs_unaligned_ptr(&creq->peer));
+static UCS_F_NOINLINE void
+uct_ud_ep_process_unexp_packet(uct_ud_iface_t *iface, uct_ud_ep_t *ep,
+                               uct_ud_neth_t *neth, unsigned length)
+{
+    char buf[256] = {0};
```
yosefe, line 1032: why need to initialize? i would expect uct_ud_dump_packet to make sure it ends by \0

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -881,92 +1001,76 @@ static void uct_ud_ep_rx_ctl(uct_ud_iface_t *iface, uct_ud_ep_t *ep,
...
-        /* must be connection request packet */
-        uct_ud_ep_rx_creq(iface, neth);
-        goto out;
+        /* Must be connection request packet */
+        ep = uct_ud_ep_rx_creq(iface, neth);
+        if (ep == NULL) {
+            goto out;
+        }
+
+        goto frag_list_insert;
```
yosefe, line 1060: why needed? can't just goto out?
i don't think that when we failed to create the ep we want to add CREQ to frag_list

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -881,92 +1001,76 @@ static void uct_ud_ep_rx_ctl(uct_ud_iface_t *iface, uct_ud_ep_t *ep,
...
-        /* Drop the packet because it is
-         * allowed to do disconnect without flush/barrier. So it
-         * is possible to get packet for the ep that has been destroyed
+                            (ep->flags & UCT_UD_EP_FLAG_DISCONNECTED))) {
+        /* Drop the packet because it is allowed to do disconnect without
+         * flush/barrier. So it is possible to get packet for the ep that has
+         * already been destroyed
          */
-        ucs_trace("RX: failed to find ep %d, dropping packet", dest_id);
+        ucs_trace("RX: dropping packet dest_id=%u, ep=%p", dest_id, ep);
```
yosefe, line 1067: IMO better print different error message for each case (ep not found, and ep disconnected)
then, can avoid initializing ep to NULL

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1006,9 +1107,16 @@ void uct_ud_ep_process_rx(uct_ud_iface_t *iface, uct_ud_neth_t *neth, unsigned b
...
-        uct_ud_ep_rx_put(neth, byte_len);
+    if (ucs_unlikely(!is_am)) {
+        if (neth->packet_type & UCT_UD_PACKET_FLAG_PUT) {
+            /* TODO: remove once ucp implements put */
+            uct_ud_ep_rx_put(neth, byte_len);
+        } else if (neth->packet_type & UCT_UD_PACKET_FLAG_CTL) {
+            uct_ud_ep_rx_ctl(iface, ep, neth);
+        } else {
+            uct_ud_ep_process_unexp_packet(iface, ep, neth, byte_len);
+        }
```
yosefe, line 1118: IMO better return control messages handling to line 992

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1757,14 +1836,36 @@ void uct_ud_ep_disconnect(uct_ep_h tl_ep)
      * to not invoke user callbacks for the endpoint which was destroyed */
     uct_ud_ep_purge(ep, UCS_ERR_CANCELED);
 
+    if (!(ep->flags & UCT_UD_EP_FLAG_CONNECT_TO_EP) &&
+        (ep->flags & UCT_UD_EP_FLAG_TX)) {
```
yosefe, line 1840: uct_ud_ep_disconnect is called for API endpoint, so how could it be without TX capability?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1757,14 +1836,36 @@ void uct_ud_ep_disconnect(uct_ep_h tl_ep)
...
+        uct_ud_iface_cep_remove_ep(iface, ep);
+        ep->flags &= ~UCT_UD_EP_FLAG_TX;
+        uct_ud_ep_ctl_op_del(ep, UCT_UD_EP_OP_CREQ);
+
+        if (uct_ud_ep_is_connected(ep)) {
+            uct_ud_ep_ctl_op_add(iface, ep, UCT_UD_EP_OP_FIN);
+        }
+
+        if (ep->flags & UCT_UD_EP_FLAG_RX) {
+            uct_ud_iface_cep_insert_ready_ep(iface, ep);
```
yosefe, line 1852: also here, if we destroyed ep form API, how could it be ready for matching with new ep from remote side?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1757,14 +1836,36 @@ void uct_ud_ep_disconnect(uct_ep_h tl_ep)
...
+            /* Don't destroy ep if it has RX capability */
+            goto out;
```
yosefe, line 1854: so no need to set UCT_UD_EP_FLAG_DISCONNECTED flag?

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -137,30 +161,48 @@ uct_ud_ep_t *uct_ud_iface_cep_get_ep(uct_ud_iface_t *iface,
...
     ucs_assert(ep->flags & UCT_UD_EP_FLAG_ON_CEP);
 
-    if (is_private) {
+    if (is_rx_only) {
+        ucs_assertv((ep->flags & UCT_UD_EP_FLAG_TX_RX) == UCT_UD_EP_FLAG_RX,
+                    "ep=%p flags=0x%x", ep, ep->flags);
         ep->flags &= ~UCT_UD_EP_FLAG_ON_CEP;
+    } else {
+        ucs_assertv(ep->flags & UCT_UD_EP_FLAG_TX_RX, "ep=%p flags=0x%x", ep,
+                    ep->flags);
```
yosefe, line 178: why needed?

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -137,30 +161,48 @@ uct_ud_ep_t *uct_ud_iface_cep_get_ep(uct_ud_iface_t *iface,
...
+int uct_ud_iface_cep_has_ep(uct_ud_ep_t *ep)
+{
+    int res = (ep->flags & UCT_UD_EP_FLAG_TX_RX) &&
+              !(ep->flags & UCT_UD_EP_FLAG_CONNECT_TO_EP);
+
+    ucs_assertv(res == !!(ep->flags & UCT_UD_EP_FLAG_ON_CEP),
+                "ep=%p flags=0x%x res=%d", ep, ep->flags, res);
+
+    return res;
+}
```
yosefe, line 193: seems redundant, could just check UCT_UD_EP_FLAG_ON_CEP

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -314,7 +356,8 @@ uct_ud_iface_conn_match_purge_cb(ucs_conn_match_ctx_t *conn_match_ctx,
     uct_ud_ep_t *ep       = ucs_container_of(elem, uct_ud_ep_t,
                                              conn_match);
 
-    ep->flags &= ~UCT_UD_EP_FLAG_ON_CEP;
+    ucs_assert(ep->flags & UCT_UD_EP_FLAG_ON_CEP);
```
yosefe, line 359: seems redundant

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -314,7 +356,8 @@ uct_ud_iface_conn_match_purge_cb(ucs_conn_match_ctx_t *conn_match_ctx,
...
+    ep->flags &= ~(UCT_UD_EP_FLAG_ON_CEP | UCT_UD_EP_FLAG_TX_RX);
```
yosefe, line 360: why need to remove TXRX?

src/uct/ib/ud/base/ud_log.c
```diff
@@ -26,10 +26,9 @@ void uct_ud_dump_packet(uct_base_iface_t *iface, uct_am_trace_type_t type,
     p    = buffer;
     endp = buffer + max;
 
-    snprintf(p, endp - p, " dst %d psn %u apsn %u %c%c",
+    snprintf(p, endp - p, " dst %d psn %u apsn %u %c",
              uct_ud_neth_get_dest_id(neth), neth->psn, neth->ack_psn,
-             (neth->packet_type & UCT_UD_PACKET_FLAG_ACK_REQ) ? 'r' : '-',
-             (neth->packet_type & UCT_UD_PACKET_FLAG_ECN)     ? 'e' : '-');
```
yosefe, line 32: this cleanup is not needed in a bugfix PR

yosefe (changes requested):
In general, conn_match seems not really neeeded for keepalive with CM. We can force packet drop on a disconnected ud endpoint only on p2p-connected endpoints, so it will not affect ep-to-iface scenario.

src/uct/ib/ud/base/ud_ep.c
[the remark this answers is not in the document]
dmitrygx: it will wait destroying from 5-min timer

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -533,9 +562,88 @@ static ucs_status_t uct_ud_ep_disconnect_from_iface(uct_ud_ep_t *ep)
...
+    }
+
+    uct_ud_ep_set_state(ep, state);
+    uct_ud_ep_ctl_op_del(ep, op);
+    uct_ud_iface_send_ctl(iface, ep, skb, NULL, 0,
+                          UCT_UD_IFACE_SEND_CTL_FLAG_SOLICITED, 1);
+    uct_ud_iface_complete_tx_skb(iface, ep, skb);
+}
+
+static uct_ud_send_skb_t *uct_ud_ep_prepare_creq(uct_ud_ep_t *ep)
```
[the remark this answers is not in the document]
dmitrygx: done

src/uct/ib/ud/base/ud_ep.c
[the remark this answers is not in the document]
dmitrygx: CREP could be sent even for internal ep (i.e. user didn't call `uct_ep_create`).
so, anyway we will need to send CREQ later.
and sending CREQ handles the following case:
1. A creates UCT EP, it sends CREQ.
2. B creates internal UCT EP, it sends CREP.
3. A sends some data.
4. A closes UCT EP and B creates UCT EP from internal one and sends CREQ.
5. A receives CREQ, can't find EP in CEP, checks ptr_array - will find UCT EP, stops disconenctign for it.

But CREP in this case will be just dropped, since DISCONNECTED flag set.

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -781,30 +896,38 @@ static void uct_ud_ep_rx_creq(uct_ud_iface_t *iface, uct_ud_neth_t *neth)
...
                                  ctl->conn_req.conn_sn, 0);
     if (ep == NULL) {
-        ep = uct_ud_ep_create_passive(iface, ctl);
-        if (ep == NULL) {
-            ucs_diag("iface %p: failed to create the passive endpoint", iface);
-            return;
-        }
+        if ((ctl->conn_req.dest_ep_id != UCT_UD_EP_NULL_ID) &&
+            ucs_ptr_array_lookup(&iface->eps, ctl->conn_req.dest_ep_id, ep)) {
+            ep->flags &= ~UCT_UD_EP_FLAG_DISCONNECTED;
```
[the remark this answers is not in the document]
dmitrygx: see example above

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -781,30 +896,38 @@ static void uct_ud_ep_rx_creq(uct_ud_iface_t *iface, uct_ud_neth_t *neth)
...
+            uct_ud_ep_set_state(ep, UCT_UD_EP_FLAG_RX);
+
+            status = uct_ud_iface_cep_insert_ready_ep(iface, ep);
```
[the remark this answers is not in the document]
dmitrygx: it means that EP has RX-only at this point

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -781,30 +896,38 @@ static void uct_ud_ep_rx_creq(uct_ud_iface_t *iface, uct_ud_neth_t *neth)
...
+                return NULL;
+            }
 
-        ep->rx.ooo_pkts.head_sn = neth->psn;
-        uct_ud_peer_copy(&ep->peer, ucs_unaligned_ptr(&ctl->peer));
-        uct_ud_ep_ctl_op_add(iface, ep, UCT_UD_EP_OP_CREP);
+            /* Stop disconnecting timer */
+            ucs_wtimer_remove(&iface->tx.timer, &ep->timer);
+        } else {
+            ep = uct_ud_ep_create_passive(iface, ctl);
```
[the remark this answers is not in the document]
dmitrygx: but we create new Endpoint, since we didn't find Endpoint which could be re-used

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -781,30 +896,38 @@ static void uct_ud_ep_rx_creq(uct_ud_iface_t *iface, uct_ud_neth_t *neth)
...
     } else if (ep->dest_ep_id == UCT_UD_EP_NULL_ID) {
         /* simultaneous CREQ */
         uct_ud_ep_set_dest_ep_id(ep, uct_ib_unpack_uint24(ctl->conn_req.ep_addr.ep_id));
-        ep->rx.ooo_pkts.head_sn = neth->psn;
         uct_ud_peer_copy(&ep->peer, ucs_unaligned_ptr(&ctl->peer));
         ucs_debug("simultaneous CREQ ep=%p"
                   "(iface=%p conn_sn=%d ep_id=%d, dest_ep_id=%d rx_psn=%u)",
                   ep, iface, ep->conn_sn, ep->ep_id,
                   ep->dest_ep_id, ep->rx.ooo_pkts.head_sn);
-        if (UCT_UD_PSN_COMPARE(ep->tx.psn, >, UCT_UD_INITIAL_PSN)) {
```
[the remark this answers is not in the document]
dmitrygx: if we have sending of CREQ scheduled, handling CREP as ACK - will stop resending of CREQ, because it will be considered as successfully sent

src/uct/ib/ud/base/ud_ep.c
dmitrygx: we initiailize it in other places too when using `uct_ud_dump_packet`.
it doesn't add trailing `\0`.

src/uct/ib/ud/base/ud_ep.c
dmitrygx: fixed

src/uct/ib/ud/base/ud_ep.c
dmitrygx: at this point EP was created successfully and we should go to frag_list.
because this CREQ could be sent in the middle of the communication flow (not as a first packet).

src/uct/ib/ud/base/ud_ep.c
dmitrygx: done

src/uct/ib/ud/base/ud_ep.c
dmitrygx: done

src/uct/ib/ud/base/ud_ep.c
dmitrygx: but we should add them to frag_list too

src/uct/ib/ud/base/ud_ep.c
dmitrygx: it is called from `uct_ud_ep_handle_peer_disconnect` though `uct_ep_destroy`

src/uct/ib/ud/base/ud_ep.c
dmitrygx: if there is RX flag, then it means that it is able to receive data from a peer even if local user closed this EP.
but keeping this EP in CEP is needed for consistency for the following rule: "UD EP is in connection matching unexpected queue if it has only RX capability, expected queue - if it has only TX capability or both RX/TX capabilities"

src/uct/ib/ud/base/ud_ep.c
dmitrygx: yes, we don't need to set `UCT_UD_EP_FLAG_DISCONNECTED` flag here

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -104,12 +107,33 @@ uct_ud_iface_cep_insert_ep(uct_ud_iface_t *iface,
...
     ret = ucs_conn_match_insert(&iface->conn_match_ctx, peer_address,
                                 conn_sn, &ep->conn_match, queue_type);
     ucs_assert_always(ret == 1);
 
     ep->flags |= UCT_UD_EP_FLAG_ON_CEP;
+
+    return UCS_OK;
+}
+
+ucs_status_t
```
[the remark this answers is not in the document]
dmitrygx: fixed

src/uct/ib/ud/base/ud_iface.c
dmitrygx: if it is found, it should have TX/RX ot both.

src/uct/ib/ud/base/ud_iface.c
dmitrygx: I'm going to remove `UCT_UD_EP_FLAG_ON_CEP` flag in favour of TX/RX flags which mean that endpoint is in CEP.
I kept it for debugging reason and check that all connection matching cases are covered.

src/uct/ib/ud/base/ud_iface.c
dmitrygx: i'll remove this flag

src/uct/ib/ud/base/ud_iface.c
dmitrygx: to not remove EP twice.
added the comment

src/uct/ib/ud/base/ud_iface.c
```diff
@@ -619,9 +662,8 @@ static void uct_ud_iface_delete_eps(uct_ud_iface_t *iface)
     int i;
 
     ucs_ptr_array_for_each(ep, i, &iface->eps) {
-        ucs_assert(!(ep->flags & UCT_UD_EP_FLAG_ON_CEP));
-        uct_iface_invoke_ops_func(&iface->super, uct_ud_iface_ops_t,
-                                  ep_free, &ep->super.super);
+        ucs_assert(!uct_ud_iface_cep_has_ep(ep));
+        uct_ud_ep_free(iface, ep);
```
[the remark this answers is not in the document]
dmitrygx: done

src/uct/ib/ud/base/ud_iface.h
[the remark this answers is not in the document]
dmitrygx: this branch will be not in CREQ handling code, but in `uct_ud_ep_process_rx`.
we will need to check whether it is CREQ or not in case of EP can't be fund in PTR array

dmitrygx:
> In general, conn_match seems not really neeeded for keepalive with CM. We can force packet drop on a disconnected ud endpoint only on p2p-connected endpoints, so it will not affect ep-to-iface scenario.

do you mean checking DISCONNECTED flag for CONNECT_TO_EP endpoints only in 1.13? it will have extra branch in `uct_ud_ep_process_rx`

yosefe:
> do you mean checking DISCONNECTED flag for CONNECT_TO_EP endpoints only in 1.13? it will have extra branch in `uct_ud_ep_process_rx`

Yes:
Let's start with a fix to check DISCONNECTED only for connect-to-ep. Maybe we can unite it with the dest_id check, by changing ep->dest_id of a disconnected p2p endpoint to an invalid value (instead of setting a DISCONNECTED flag). Make sure it solves the timeouts issue and port it to v1.13.x
I think the full connecting matching it's a complex change in a "sensitive" place that may break other use cases, so better postpone it for now. For example maybe need to move this logic to another layer, make sure that we don't send more wireup packets now, etc.

dmitrygx:
> Let's start with a fix to check DISCONNECTED only for connect-to-ep. Maybe we can unite it with the dest_id check, by changing ep->dest_id of a disconnected p2p endpoint to an invalid value (instead of setting a DISCONNECTED flag). Make sure it solves the timeouts issue and port it to v1.13.x

submitted #8151 which checks `CONNECT_TO_EP | DISCONNECT` flags to make a decision about dropping RX packet.
currently, testing it with IO-demo reproducer of "timeout waiting for N replies".

PR #8080, opened 2022-03-30
UCT/IB/UD: Fix multi-path handling for passive endpoint

dmitrygx:
## What

Fix multi-path handling for passive endpoint.

## Why ?

If RoCE LAG feature is used to spread traffic across two paths, address held in passive endpoint is incorrect, because it always use `path_index = 0`.
The issue is `uct_ud_ep_create_passive` function should set `path_index` prior invoking `uct_ep_connect_to_ep` function which uses `path_index` to unpack UD EP address.

## How ?

1. Introduce `uct_ud_ep_create` function which fills `uct_ep_params`, invokes `uct_ep_create` function, sets `conn_sn`.
2. Use `uct_ud_ep_create` in `uct_ud_ep_create_passive` and `uct_ud_ep_create_connected_common` to reduce duplication and fix bug.

### DIFF
--- a/src/uct/ib/ud/base/ud_ep.c
+++ b/src/uct/ib/ud/base/ud_ep.c
@@ -529,6 +529,30 @@ static ucs_status_t uct_ud_ep_disconnect_from_iface(uct_ud_ep_t *ep)
     return UCS_OK;
 }
 
+static ucs_status_t uct_ud_ep_create(uct_ud_iface_t *iface, int path_index,
+                                     uct_ud_ep_conn_sn_t conn_sn,
+                                     uct_ud_ep_t **new_ep_p)
+{
+    uct_ep_params_t params = {
+        .field_mask = UCT_EP_PARAM_FIELD_IFACE |
+                      UCT_EP_PARAM_FIELD_PATH_INDEX,
+        .iface      = &iface->super.super.super,
+        .path_index = path_index
+    };
+    uct_ep_h new_ep_h;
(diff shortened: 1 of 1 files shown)


dmitrygx:
the fix is setting `path_index` before invoking `uct_ep_connect_to_ep` during creating a passive endpoint

dmitrygx:

brminich (dismissed):

dmitrygx:
@yosefe could you review pls?

hoopoepg (approved):

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -569,18 +591,12 @@ ucs_status_t uct_ud_ep_create_connected_common(const uct_ep_params_t *ep_params,
...
-
-    status = uct_ep_create(&params, &new_ep_h);
+    status = uct_ud_ep_create(iface, path_index, conn_sn, &ep);
     if (status != UCS_OK) {
         goto out;
     }
 
-    ep          = ucs_derived_of(new_ep_h, uct_ud_ep_t);
-    ep->conn_sn = conn_sn;
+    *new_ep_p = &ep->super.super;
```
yosefe, line 599: already in line 633

dmitrygx: removed, and force-pushed (PR is small enough)

yosefe (approved):

PR #8085, opened 2022-04-01
UCT/IB: Make the memlock limit threshold configurable

ivankochin:
## What
Introducing the `UCX_MEMLOCK_LIMIT` user control that allows to set custom required memory locked limit level.

## Why ?
Now if system memory locked limit is less then 512M some parts of UCX (that depends on this limit, e.g. IB transports) are disabled. But in some cases the proper work of dependent UCX parts may be guaranteed by more or less strict limit. This patch introduces the control that allows to configure this threshold.

### DIFF
--- a/src/ucs/config/global_opts.c
+++ b/src/ucs/config/global_opts.c
@@ -44,6 +44,7 @@ ucs_global_opts_t ucs_global_opts = {
     .tuning_path           = "",
     .memtrack_dest         = "",
     .memtrack_limit        = UCS_MEMUNITS_INF,
+    .memlock_limit         = 500 * UCS_MBYTE,
     .stats_trigger         = "exit",
     .profile_mode          = 0,
     .profile_file          = "",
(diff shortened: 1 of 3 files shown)


src/ucs/config/global_opts.c
```diff
@@ -44,6 +44,7 @@ ucs_global_opts_t ucs_global_opts = {
     .tuning_path           = "",
     .memtrack_dest         = "",
     .memtrack_limit        = UCS_MEMUNITS_INF,
+    .memlock_limit         = 500 * UCS_MBYTE,
```
dmitrygx, line 47: `512`?

dmitrygx: BTW, it seems that it could be even `0`, because it will be set from configuration anyway, wdyt?

src/ucs/config/global_opts.c
```diff
@@ -165,6 +166,15 @@ static ucs_config_field_t ucs_global_opts_table[] = {
...
   ucs_offsetof(ucs_global_opts_t, memtrack_limit), UCS_CONFIG_TYPE_MEMUNITS},
 
+ {"MEMLOCK_LIMIT", "512m",
+  "Locked memory limit threshold required for proper UCX work.\n"
+  "The default value guarantees the proper work of all dependent UCX parts\n"
+  "in most cases. But in some cases the proper work can be guaranteed\n"
+  "by more or less strict limit threshold.\n"
+  "Please change the limit threshold only if you cannot change \n"
+  "system locked memory limit (ulimit -S -l unlimited).",
+  ucs_offsetof(ucs_global_opts_t, memlock_limit), UCS_CONFIG_TYPE_MEMUNITS},
```
brminich, line 176: I'd rephrase it somehow. For instance:
`The system limit for maximum amount of locked memory must not be smaller than this value.
Otherwise some UCX transports may not function properly. 
Please change this threshold only if you cannot change system locked memory limit (ulimit -S -l unlimited). `

yosefe: also, pls wrap lines (they should be longer)

src/ucs/config/global_opts.c
ivankochin: Good catch, thanks!
But I prefer to leave it as 512 to save the consistency between the variables that are already set to their default values at this point.

dmitrygx: > But I prefer to leave it as 512 to save the consistency between the variables that are already set to their default values at this point.

yes, makes sense

src/ucs/config/global_opts.c
ivankochin: Done.

But shouldn't we mention that this limit value is valid for the most, but not for all the systems. Are we sure that there are no systems when this limit isn't enough for proper work? WDYT about something like:
```
The system limit for maximum amount of locked memory must not be smaller than this value to guarantee the propers UCX work on most systems. Please change this threshold only if you cannot change system locked memory limit (ulimit -S -l unlimited).
```

brminich (approved):

ivankochin:
It was decided to close this request since warning reporting in this case was considered as a bad user experience.

PR #8090, opened 2022-04-05
UCT/IB: Remove RDMA transport disabling logic in case of too low memlock limit value

ivankochin:
## What
Now the RDMA transport is automatically disabled when the memlock limit is lower than some predefined value. This patch removes this logic.

## Why ?
It was decided that hard error during the further operation with IB transport is better customer experience than disabling of transport + warning. Plus we cannot define some common limit that will guarantee the proper RDMA transport work in all cases.

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -30,9 +30,7 @@
 #include <sys/resource.h>
 
 
-#define UCT_IB_MD_RCACHE_DEFAULT_ALIGN    16
-#define UCT_IB_MD_REQUIRED_MEMLOCK_LIMIIT (512 * UCS_MBYTE)
-
+#define UCT_IB_MD_RCACHE_DEFAULT_ALIGN 16
 
 static UCS_CONFIG_DEFINE_ARRAY(pci_bw,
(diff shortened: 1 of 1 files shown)


brminich:
btw, maybe print the limit value with `ucs_debug`, if it is not unlimited? Can be helpful for debugging

ivankochin:
> btw, maybe print the limit value with `ucs_debug`, if it is not unlimited? Can be helpful for debugging

The current limit value is printed [in the corresponding error message](https://github.com/ivankochin/ucx/blob/master/src/uct/ib/base/ib_log.c#L248). Do you mean that we need some additional debug message during components quering?

brminich:
> > btw, maybe print the limit value with `ucs_debug`, if it is not unlimited? Can be helpful for debugging
> 
> The current limit value is printed [in the corresponding error message](https://github.com/ivankochin/ucx/blob/master/src/uct/ib/base/ib_log.c#L248). Do you mean that we need some additional debug message during components quering?

No, I meant always print it in `debug` if it is not unlimited

brminich (approved):

yosefe (approved):

PR #8091, opened 2022-04-05
UCT/IB: Set 2 paths by default for CX7

brminich:
## What
Set default value of `IB_NUM_PATH` to `2` for CX-7

## Why ?
Better performance


src/uct/ib/base/ib_iface.c
```diff
@@ -1180,6 +1180,11 @@ static void uct_ib_iface_set_num_paths(uct_ib_iface_t *iface,
             /* IB - number of paths is LMC level */
             ucs_assert(iface->path_bits_count > 0);
             iface->num_paths = iface->path_bits_count;
+            if ((iface->num_paths == 1) &&
```
yosefe, line 1183: IMO it can be for both IB and RoCE

yosefe (approved):

PR #8095, opened 2022-04-06
UCT/IB: Improve DC full handshake logic

gleon99:
## What
1. Use address version flag.
2. Check whether HW reports better performance when no RDMA_WRITE

## Why ?
1. Currently device flag is used, which is a bug (incorrect value).
2. Only set `rdma_wr_disabled` bit when it has an impact.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.c
+++ b/src/uct/ib/dc/dc_mlx5.c
@@ -324,13 +324,16 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
         attr.super.max_inl_cqe[UCT_IB_DIR_RX] = 0;
         attr.uidx             = htonl(dci_index) >> UCT_IB_UIDX_SHIFT;
         attr.full_handshake   = full_handshake;
-        attr.rdma_wr_disabled = iface->flags & UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT;
+        attr.rdma_wr_disabled = (iface->flags & UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT) &&
+                                (md->flags & UCT_IB_MLX5_MD_FLAG_NO_RDMA_WR_OPTIMIZED);
         status = uct_ib_mlx5_devx_create_qp(ib_iface, &dci->txwq.super,
                                             &dci->txwq, &attr);
         if (status != UCS_OK) {
             return status;
         }
 
+        ucs_debug("created DevX DCI 0x%x, rdma_wr_disabled=%d", dci->txwq.super.qp_num,
+                  attr.rdma_wr_disabled);
         goto init_qp;
     }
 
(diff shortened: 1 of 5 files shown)


yosefe (dismissed):

src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
```diff
@@ -272,7 +272,8 @@ struct uct_ib_mlx5_cmd_hca_cap_bits {
     uint8_t    rc[0x1];
 
     uint8_t    uar_4k[0x1];
-    uint8_t    reserved_at_241[0x9];
+    uint8_t    dci_no_rdma_wr_optimized_performance[0x1];
+    uint8_t    reserved_at_241[0x8];
```
yosefe, line 276: should we rename it to reserved_at_242?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -324,7 +324,8 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
         attr.super.max_inl_cqe[UCT_IB_DIR_RX] = 0;
         attr.uidx             = htonl(dci_index) >> UCT_IB_UIDX_SHIFT;
         attr.full_handshake   = full_handshake;
-        attr.rdma_wr_disabled = iface->flags & UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT;
+        attr.rdma_wr_disabled = (iface->flags & UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT) &
+                                (md->flags & UCT_IB_MLX5_MD_FLAG_NO_RDMA_WR_OPTIMIZED);
```
yosefe, line 328: maybe add ucs_debug message to uct_ib_mlx5_devx_create_qp, that indicates which value we passed to rdma_wr_disabled

src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
gleon99: Nice!

src/uct/ib/dc/dc_mlx5.c
gleon99: Are you sure? In general I don't see such messages for other flags and fields in this file or in `ib_mlx5_dv.c`

src/uct/ib/mlx5/dv/ib_mlx5_dv.c
```diff
@@ -157,6 +157,8 @@ ucs_status_t uct_ib_mlx5_devx_create_qp(uct_ib_iface_t *iface,
 
     qp->devx.obj = mlx5dv_devx_obj_create(dev->ibv_context, in, sizeof(in),
                                           out, sizeof(out));
+    ucs_debug("created DevX DCI 0x%x, rdma_wr_disabled=%d", qp->devx.obj,
```
yosefe, line 160: it should be after creating the object successfully, and "DCI 0x%x" should print the qp_num
Also sorry but it should be in uct_dc_mlx5_iface_create_dci, not uct_ib_mlx5_devx_create_qp

yosefe (approved):

yosefe (approved):

PR #8099, opened 2022-04-06
UCP/WIREUP: Try fallback creating KA lane from exisiting lanes

dmitrygx:
## What

Add another fallback type by creating KA lane from already selected resources.

## Why ?

To not open additional resources in case of lack of space in CM private data to pack all addresses.

## How ?

1. Add `tl_bitmap` to UCP wireup select context and update it by adding `rsc_index` which were selected.
2. Introduce new EP init flag - `UCP_EP_INIT_KA_FROM_EXIST_LANES` to select transport fo keepalive lane from already selected lanes.
3. Add `UCP_EP_INIT_KA_FROM_EXIST_LANES` flag to fallback priority list.

### DIFF
--- a/src/ucp/core/ucp_ep.h
+++ b/src/ucp/core/ucp_ep.h
@@ -160,8 +160,8 @@ enum {
                                                            support CONNECT_TO_IFACE
                                                            mode only */
     UCP_EP_INIT_CREATE_AM_LANE_ONLY    = UCS_BIT(8),  /**< Endpoint requires an AM lane only */
-    UCP_EP_INIT_ALLOW_KA_AUX_TL        = UCS_BIT(9),  /**< Endpoint allows selecting of auxiliary
-                                                           transports for keepalive lane */
+    UCP_EP_INIT_KA_FROM_EXIST_LANES    = UCS_BIT(9),  /**< Use only existing lanes to create
+                                                           keepalive lane */
     UCP_EP_INIT_ALLOW_AM_AUX_TL        = UCS_BIT(10)  /**< Endpoint allows selecting of auxiliary
                                                            transports for AM lane */
 };
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -88,6 +88,7 @@ typedef struct {
                                                           * found during selection */
     ucp_lane_index_t          num_lanes;                 /* Number of active lanes */
     unsigned                  ucp_ep_init_flags;         /* Endpoint init extra flags */
+    ucp_tl_bitmap_t           tl_bitmap;                 /* TL bitmap of selected resources */
 } ucp_wireup_select_context_t;
 
 static const char *ucp_wireup_md_flags[] = {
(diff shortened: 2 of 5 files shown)


src/ucp/wireup/select.c
```diff
@@ -1691,7 +1710,7 @@ ucp_wireup_search_lanes(const ucp_wireup_select_params_t *select_params,
     ucp_wireup_select_info_t am_info;
     ucs_status_t status;
 
-    memset(select_ctx, 0, sizeof(*select_ctx));
+    ucp_wireup_select_context_init(select_ctx);
```
brminich, line 1713: why need to introduce a func? Could just use memset as before

src/ucp/core/ucp_ep.h
```diff
@@ -162,8 +162,10 @@ enum {
     UCP_EP_INIT_CREATE_AM_LANE_ONLY    = UCS_BIT(8),  /**< Endpoint requires an AM lane only */
     UCP_EP_INIT_ALLOW_KA_AUX_TL        = UCS_BIT(9),  /**< Endpoint allows selecting of auxiliary
                                                            transports for keepalive lane */
-    UCP_EP_INIT_ALLOW_AM_AUX_TL        = UCS_BIT(10)  /**< Endpoint allows selecting of auxiliary
+    UCP_EP_INIT_ALLOW_AM_AUX_TL        = UCS_BIT(10), /**< Endpoint allows selecting of auxiliary
                                                            transports for AM lane */
+    UCP_EP_INIT_KA_FROM_EXIST_LANES    = UCS_BIT(11)  /**< Use only existing lanes to create
```
brminich, line 167: now it conflicts with `UCP_EP_INIT_ALLOW_KA_AUX_TL`.
Is it possible to use `tl_bitmap`  from `ucp_wireup_select_context_t`  instead?

dmitrygx: > now it conflicts with UCP_EP_INIT_ALLOW_KA_AUX_TL.

yes, this is a problem

> now it conflicts with UCP_EP_INIT_ALLOW_KA_AUX_TL.
Is it possible to use tl_bitmap from ucp_wireup_select_context_t instead?

how it should be done? we should have a control to define whether we should transports from existing lanes to create KA lane or not

dmitrygx: discussed offline with @brminich that we coudl remove `UCP_EP_INIT_ALLOW_KA_AUX_TL` flag - thanks, good catch, fixed.

dmitrygx:

src/ucp/wireup/select.c
```diff
@@ -1647,16 +1652,21 @@ ucp_wireup_add_keepalive_lane(const ucp_wireup_select_params_t *select_params,
...
         !ucp_worker_keepalive_is_enabled(worker) ||
-        (select_params->ep_init_flags & (UCP_EP_INIT_CREATE_AM_LANE_ONLY |
-                                         UCP_EP_INIT_FLAG_INTERNAL))) {
+        (ep_init_flags & UCP_EP_INIT_FLAG_INTERNAL)) {
         return UCS_OK;
     }
 
+    if (ep_init_flags & (UCP_EP_INIT_CREATE_AM_LANE_ONLY |
+                         UCP_EP_INIT_KA_FROM_EXIST_LANES)) {
+        UCS_BITMAP_AND_INPLACE(&tl_bitmap, select_ctx->tl_bitmap);
```
yosefe, line 1667: maybe use if/else: 
const ucp_tl_bitmap_t *tl_bitmap;
if (...) tl_bitmap=&select_params->tl_bitmap;
else tl_bitmap = &select_ctx->tl_bitmap

dmitrygx: done

brminich (approved):

dmitrygx:
@yosefe could you review pls?

yosefe (dismissed):

brminich (approved):

PR #8103, opened 2022-04-06
UCT/IB: Fix defining elements of IB MDs array

dmitrygx:
## What

Fix defining elements of IB MDs array.

## Why ?

Fixes:
```
[1649063561.762911] [swx-ucx02:25944:0]         dc_mlx5.c:473  UCX  ERROR ibv_modify_qp(DCI, RTS) failed : Invalid argument
(gdb) p iface->super.super.config.max_rd_atomic
$4 = 80 'P'
```
`max_rd_atomic` has incorrect value, since it should be took from HCA's `max_rd_atomic_dc` capability, but Verbs IB MD which doesn't have this value at all.
the problem is still persists if Verbs IB MD is selected (e.g. setting `UCX_IB_MLX5_DEVX=no`) - DC_MLX5, RC_MLX5 and UD_MLX5 still use IB_DEVX.

## How ?

1. Fix adding DEVX IB MD entry by changing `HAVE_MLX5_DV` to `HAVE_MLX5_HW`.
2. Fix experimental Verbs IB MD entry by changing `HAVE_MLX5_HW && HAVE_VERBS_EXP_H` to `HAVE_VERBS_EXP_H`.

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -227,13 +227,13 @@ extern uct_ib_md_ops_entry_t UCT_IB_MD_OPS_NAME(exp);
 static uct_ib_md_ops_entry_t UCT_IB_MD_OPS_NAME(verbs);
 
 static uct_ib_md_ops_entry_t *uct_ib_ops[] = {
-#if defined (HAVE_MLX5_DV) && defined (HAVE_DEVX)
+#if defined (HAVE_MLX5_HW) && defined (HAVE_DEVX)
     &UCT_IB_MD_OPS_NAME(devx),
 #endif
 #if defined (HAVE_MLX5_DV)
     &UCT_IB_MD_OPS_NAME(dv),
 #endif
-#if defined (HAVE_MLX5_HW) && defined (HAVE_VERBS_EXP_H)
+#if defined(HAVE_VERBS_EXP_H)
     &UCT_IB_MD_OPS_NAME(exp),
 #endif
     &UCT_IB_MD_OPS_NAME(verbs)


brminich:
what is the test which failed this way?

dmitrygx:
> what is the test which failed this way?

```
./test/gtest/gtest --gtest_filter="ud/test_ucp_loopback.envelope/0"
```
on swx-ucx node

> Looks like we need to init max_rd_atomic_dc for dv case as well

yes, and EXP Verbs and normal Verbs IB MDs should be updated to.
also, we should extend Verbs IB MD to `uct_ib_mlx5_md_t`, because it Verbs IB MD could be used to initialize DC, RC_MLX5 and UD_MLX5.

brminich:
> > what is the test which failed this way?
> 
> ```
> ./test/gtest/gtest --gtest_filter="ud/test_ucp_loopback.envelope/0"
> ```
> 
> on swx-ucx node
> 
> > Looks like we need to init max_rd_atomic_dc for dv case as well
> 
> yes, and EXP Verbs and normal Verbs IB MDs should be updated to. also, we should extend Verbs IB MD to `uct_ib_mlx5_md_t`, because it Verbs IB MD could be used to initialize DC, RC_MLX5 and UD_MLX5.

what do you mean? We do not support dc with verbs and that parameter is only relevant for dc

dmitrygx:
> what do you mean? We do not support dc with verbs and that parameter is only relevant for dc

then we should disable selecting Verbs IB MD for DC/RC_MLX5/UD_MLX5

dmitrygx:
replace by #8107

PR #8104, opened 2022-04-07
UCT/IB/MLX5: CQE zipping feature testing

ivankochin:
## What
Testing of the CQE zipping feature.

## Why ?
No we don't have a special test case that covers CQE zipping.

## How ?
This test performs a lot of similar send operation using active messages UCT API (bcopy/zcopy) and then checks the CQE zipping counters after that.  If at least one CQE zipping event was noticed, the test was passed successfully.

### DIFF
--- a/test/gtest/uct/ib/test_cqe_zipping.cc
+++ b/test/gtest/uct/ib/test_cqe_zipping.cc
@@ -1,0 +1,157 @@
+/**
+* Copyright (C) NVIDIA Corporation. 2022.  ALL RIGHTS RESERVED.
+*/
+
+#include <common/test.h>
+#include <common/test_helpers.h>
+#include <uct/ib/test_ib.h>
+
+extern "C" {
+#include <uct/api/uct.h>
+#include <uct/ib/base/ib_iface.h>
+}
+
+#define UCT_INSTANTIATE_MLX5_TEST_CASE(_test_case) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, dc_mlx5) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, ud_mlx5)
+
+class test_cqe_zipping : public test_uct_ib_with_specific_port {
+public:
+    bool is_cqe_zipping_expected()
+    {
+#ifdef ENABLE_STATS
+        return !RUNNING_ON_VALGRIND &&
+               !IBV_PORT_IS_LINK_LAYER_ETHERNET(&m_port_attr);
+#else
+        return false;
+#endif
+    }
+
+    void send_while_possible()
+    {
+        while (am_zcopy() != UCS_ERR_NO_RESOURCE) {
+            ++m_send_cnt;
+        }
+    }
+
+    void wait_for_completion() const
+    {
+        /*
+         * Local deadline prevents hanging inside receive loop in case of
+         * connectivity issues
+         */
+        const ucs_time_t deadline = ucs::get_deadline(60);
+        while ((m_send_cnt != m_recv_cnt) && (ucs_get_time() < deadline)) {
+            progress();
+        }
+        ASSERT_EQ(m_send_cnt, m_recv_cnt);
+    }
+
+    size_t get_total_zipped_count() const
+    {
+        size_t zipped_cqes_count  = 0;
+#ifdef ENABLE_STATS
+        constexpr int counters[2] = {UCT_IB_IFACE_STAT_TX_COMPLETION_ZIPPED,
+                                     UCT_IB_IFACE_STAT_RX_COMPLETION_ZIPPED};
+
+        for (auto &entity : m_entities) {
+            auto stats = ucs_derived_of(entity->iface(), uct_ib_iface_t)->stats;
+            for (auto &counter : counters) {
+                zipped_cqes_count += UCS_STATS_GET_COUNTER(stats, counter);
+            }
+        }
+#endif
+        return zipped_cqes_count;
+    }
+
+    virtual void init()
+    {
+#ifdef ENABLE_STATS
+        stats_activate();
+#endif
+        modify_config("IB_CQE_ZIPPING_ENABLE", "y");
+
+        test_uct_ib_with_specific_port::init();
+        test_uct_ib::init();
+
+        uct_iface_set_am_handler(receiver()->iface(), 0, am_cb, &m_recv_cnt, 0);
+        m_send_buf = new mapped_buffer(m_buf_size, 0, *sender());
+    }
+
+    void flush_and_reset()
+    {
+        uct_test::flush();
+        m_send_cnt = m_recv_cnt = 0;
+    }
+
+private:
+    entity *sender() const
+    {
+        return m_e1;
+    }
+
+    entity *receiver() const
+    {
+        return m_e2;
+    }
+
+    virtual void cleanup()
+    {
+        delete m_send_buf;
+        uct_iface_set_am_handler(receiver()->iface(), 0, NULL, NULL, 0);
+
+        test_uct_ib::cleanup();
+        test_uct_ib_with_specific_port::cleanup();
+
+#ifdef ENABLE_STATS
+        stats_restore();
+#endif
+    }
+
+    ucs_status_t am_zcopy()
+    {
+        auto size = sender()->iface_attr().cap.am.max_zcopy;
+        size      = ucs_min(size, m_buf_size);
+        uct_iov_t iov{m_send_buf->ptr(), size, m_send_buf->memh(), 0, 1};
+
+        return uct_ep_am_zcopy(sender()->ep(0), 0, m_send_buf->ptr(), 0, &iov,
+                               1, 0, nullptr);
+    }
+
+    static ucs_status_t
+    am_cb(void *arg, void *data, size_t length, unsigned flags)
+    {
+        size_t &recv_cnt = *(static_cast<size_t*>(arg));
+        ++recv_cnt;
+        return UCS_OK;
+    }
+
+    static constexpr size_t m_buf_size = 2048;
+    size_t                  m_send_cnt = 0, m_recv_cnt = 0;
+    mapped_buffer           *m_send_buf{nullptr};
+};
+
+/* We test only ZCOPY with 4K size messages due to good PCI load and stably
+ * CQE zipping reproducing on all platforms.
+ */
+UCS_TEST_P(test_cqe_zipping, zcopy)
+{
+    int deadline_seconds      = is_cqe_zipping_expected() ? 90 : 5;
+    const ucs_time_t deadline = ucs::get_deadline(deadline_seconds);
+
+    while ((ucs_get_time() < deadline) && (get_total_zipped_count() == 0)) {
+        /* Flush resets the transort resources which were acquired during the
+         * previous iteration.
+         */
+        flush_and_reset();
+        send_while_possible();
+        wait_for_completion();
+    }
+
+    if (is_cqe_zipping_expected()) {
+        EXPECT_GT(get_total_zipped_count(), 0);
+    }
+}
+
+UCT_INSTANTIATE_MLX5_TEST_CASE(test_cqe_zipping)
--- a/test/gtest/Makefile.am
+++ b/test/gtest/Makefile.am
@@ -224,6 +224,10 @@ if HAVE_TL_DC
 gtest_SOURCES += \
 	uct/ib/test_dc.cc
 endif
+if HAVE_MLX5_DV
+gtest_SOURCES += \
+	uct/ib/test_cqe_zipping.cc
+endif
 if HAVE_RDMACM
 gtest_SOURCES += \
 	uct/ib/test_sockaddr.cc
--- a/test/gtest/uct/ib/test_ib.h
+++ b/test/gtest/uct/ib/test_ib.h
@@ -38,7 +38,7 @@ public:
     test_uct_ib_with_specific_port();
     void init();
     void cleanup();
-    virtual void check_port_attr() = 0;
+    virtual void check_port_attr() {};
 
 protected:
     std::string m_dev_name;


gleon99:
Pipeline failure seems relevant

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+#include <common/test.h>
+#include <common/test_helpers.h>
+#include <uct/uct_test.h>
+
+extern "C" {
+#include <uct/api/uct.h>
+#include <uct/ib/base/ib_iface.h>
+}
+
+#include <utility>
```
gleon99, line 16: What's used from there?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+#include <memory>
+#include <array>
+
+#if defined(ENABLE_STATS)
+#define UCT_INSTANTIATE_MLX5_TEST_CASE(_test_case) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5) \
```
gleon99, line 22: The transports list is essentially repeated here and in `check_skip_test_tl`. Do you think there is a feasible way to avoid this repetition?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+            m_test(test),
+            m_entity(test.create_entity(0)),
+            m_buf(new mapped_buffer(buf_size, 0, *m_entity))
+        {
+            m_test.m_entities.push_back(m_entity);
+            uct_iface_set_am_handler(m_entity->iface(), 0, am_cb, &m_test, 0);
+        }
+
+        test_cqe_zipping               &m_test;
+        entity                         *m_entity;
```
gleon99, line 40: Can replace ptr with ref/obj instead?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+        peers = {nullptr, nullptr};
+
+        uct_test::cleanup();
+        stats_restore();
+    }
+
+    static ucs_status_t
+    am_cb(void *arg, void *data, size_t length, unsigned flags)
+    {
+        test_cqe_zipping *test = reinterpret_cast<test_cqe_zipping*>(arg);
```
gleon99, line 143: `static_cast`?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+#include <memory>
+#include <array>
+
+#if defined(ENABLE_STATS)
+#define UCT_INSTANTIATE_MLX5_TEST_CASE(_test_case) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, dc_mlx5) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, ud_mlx5)
+
+struct test_cqe_zipping : public uct_test {
```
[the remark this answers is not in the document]
ivankochin: Just to get rid of `public` access specifier on the next line.

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+
+        test_cqe_zipping               &m_test;
+        entity                         *m_entity;
+        std::unique_ptr<mapped_buffer> m_buf;
+    };
+
+    void send_while_possible()
+    {
+        ucs_time_t deadline = ucs::get_deadline(30);
+        for (auto &peer : peers) {
```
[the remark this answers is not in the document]
ivankochin: As I understand the logic of the test, making a round-robin mae even reduce the probability of CQE zipping since during the round robin we will send messages to different QPs one after another and therefore the time between two messages send to one QP will increase. IRIC HW counts some deadline before placing CQE without zipping, so I don't think it is a good idea.

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin: The only way to get entity that I know is `test::create_entity` function that returns a pointer. And the entity have no copy/move semantic interfaces, so there is no way to use object. 

We can use reference instead of pointer, but what is the purpose to do that?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -63,80 +46,77 @@ struct test_cqe_zipping : public uct_test {
...
-            send_cnt = recv_cnt = 0;
-            flush();
-            send_while_possible();
-            wait_for_completion();
-
-            if (collect_zipped_counters() > 0) {
-                return;
+        std::vector<int> counters = {UCT_IB_IFACE_STAT_RX_COMPLETION_ZIPPED,
+                                     UCT_IB_IFACE_STAT_RX_COMPLETION_ZIPPED};
+        size_t zipped_cqes_cnt{};
```
gleon99, line 51: `{}` ?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -150,24 +130,33 @@ struct test_cqe_zipping : public uct_test {
...
-    {
-        if ((GetParam()->tl_name != "dc_mlx5") &&
-            (GetParam()->tl_name != "rc_mlx5") &&
-            (GetParam()->tl_name != "ud_mlx5")) {
-            UCS_TEST_SKIP_R("not supported yet");
-        }
-
-        check_skip_test();
-    }
+    constexpr static size_t buf_size = 4096;
```
gleon99, line 133: Minoir: usually `static constexpr` is used.

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -150,24 +130,33 @@ struct test_cqe_zipping : public uct_test {
...
 
 /* We test only ZCOPY with 4K size messages due to good PCI load and stably
  * CQE zipping reproducing on all platforms.
  */
 UCS_TEST_SKIP_COND_P(test_cqe_zipping, zcopy, RUNNING_ON_VALGRIND)
 {
-    do_test();
+    const ucs_time_t deadline = ucs::get_deadline(60);
+
+    /* Some transports (e.g. UD) dynamically extend the amout of sends,
```
gleon99, line 145: `amout` -> `amount`

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin: Didn't understand the question. Just a default initialization, what is the problem?

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin:
```suggestion
    static constexpr size_t buf_size = 4096;
```

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin:
```suggestion
    /* Some transports (e.g. UD) dynamically extend the amount of sends,
```

test/gtest/uct/ib/test_cqe_zipping.cc
gleon99: `{}` is usually used for classes/structs, etc.
Why not `= 0`?

ivankochin: > {} is usually used for classes/structs, etc.

Why do you think so?
There is no special reason, I just used one of possible ways to init the variable. I can change it if you like.

gleon99: You are right that it is possible, but AFAIK initializer lists are rarely used for vars. I can't find even a single example in our repo, so I'd say to avoid it even just for the sake of consistency :)

ivankochin:
```suggestion
        size_t zipped_cqes_cnt = 0;
```

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,170 @@
+/**
+* Copyright (C) NVIDIA 2022.  ALL RIGHTS RESERVED.
+*
+* See file LICENSE for terms.
+*/
+
+#include "common/test.h"
```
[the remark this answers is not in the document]
ivankochin: As I know the community defined standard is using `""` when including user-defined headers and using `<>` when using system-defined headers. The only practical difference here is the order of directories in which preprocessor will search this files.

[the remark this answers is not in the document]
gleon99: I'd go with `<` at least for the sake of consistency with other test files.

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
+/**
+* Copyright (C) NVIDIA Corporation. 2022.  ALL RIGHTS RESERVED.
+*/
+
+#include "common/test.h"
+#include "common/test_helpers.h"
+#include "test_ib.h"
```
ivankochin, line 7:
```suggestion
#include <common/test.h>
#include <common/test_helpers.h>
#include <test_ib.h>
```

gleon99:
👍 
@yosefe  please look

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+#include <test_ib.h>
+
+extern "C" {
+#include <uct/api/uct.h>
+#include <uct/ib/base/ib_iface.h>
+}
+
+#include <vector>
+
+#if defined(ENABLE_STATS)
```
yosefe, line 16: maybe run the test also without stats, but just don't check the stats counters?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+private:
+    struct peer {
+        peer(test_cqe_zipping &test) :
+            m_test(test),
+            m_entity(test.create_entity(0)),
+            m_buf(new mapped_buffer(buf_size, 0, *m_entity))
+        {
+            m_test.m_entities.push_back(m_entity);
```
yosefe, line 94: why need to save m_test?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+            uct_iface_set_am_handler(m_entity->iface(), 0, am_cb, &m_test, 0);
+        }
+
+        test_cqe_zipping               &m_test;
+        entity                         *m_entity;
+        std::unique_ptr<mapped_buffer> m_buf;
```
yosefe, line 100: it's not so much a class; i'd drop m_ prefix

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+    };
+
+    virtual void cleanup()
+    {
+        flush();
```
yosefe, line 105: why needed?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+        peers.clear();
+
+        uct_test::cleanup();
+        stats_restore();
+    }
+
+    ucs_status_t am_zcopy(peer &input_peer)
```
yosefe, line 112: .. or maybe make "peer" a real class and make am_zcopy its method

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+
+    static constexpr size_t buf_size = 2048;
+    std::vector<peer>       peers = {};
+    size_t                  send_cnt = 0, recv_cnt = 0;
+};
+
+/* We test only ZCOPY with 4K size messages due to good PCI load and stably
+ * CQE zipping reproducing on all platforms.
+ */
+UCS_TEST_SKIP_COND_P(test_cqe_zipping, zcopy, RUNNING_ON_VALGRIND)
```
yosefe, line 152: why skip on valgrind? because too slow to cause zipping?
maybe run the test but without validating counters?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+{
+    const ucs_time_t deadline = ucs::get_deadline(60);
+
+    /* Some transports (e.g. UD) dynamically extend the amount of sends,
+     * so we must do several send/recv iterations to make CQE zipping
+     * sensable from the FW side.
+     */
+    while (ucs_get_time() < deadline) {
+        flush();
```
yosefe, line 161: why need to flush?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+/* We test only ZCOPY with 4K size messages due to good PCI load and stably
+ * CQE zipping reproducing on all platforms.
+ */
+UCS_TEST_SKIP_COND_P(test_cqe_zipping, zcopy, RUNNING_ON_VALGRIND)
+{
+    const ucs_time_t deadline = ucs::get_deadline(60);
+
+    /* Some transports (e.g. UD) dynamically extend the amount of sends,
+     * so we must do several send/recv iterations to make CQE zipping
+     * sensable from the FW side.
```
yosefe, line 158: commend unclear

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+    void check_port_attr() {
+        if (IBV_PORT_IS_LINK_LAYER_ETHERNET(&m_port_attr)) {
+            /* Hard to get zipped CQEs on Ethernet link */
+            UCS_TEST_SKIP_R("skip CQE zipping test for Ethernet link type");
+        }
+    }
+
+    static constexpr size_t buf_size = 2048;
+    std::vector<peer>       peers = {};
+    size_t                  send_cnt = 0, recv_cnt = 0;
```
yosefe, line 146: use m_

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+         * connectivity issues
+         */
+        const ucs_time_t deadline = ucs::get_deadline(15);
+        while ((send_cnt != recv_cnt) && (ucs_get_time() < deadline)) {
+            progress();
+        }
+        ASSERT_EQ(send_cnt, recv_cnt);
+    }
+
+    size_t collect_zipped_counters()
```
yosefe, line 51: 1. can it be const method?
2. rename to something like get_total_zipped_count()

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+    {
+        std::vector<int> counters = {UCT_IB_IFACE_STAT_TX_COMPLETION_ZIPPED,
+                                     UCT_IB_IFACE_STAT_RX_COMPLETION_ZIPPED};
+        size_t zipped_cqes_cnt = 0;
```
yosefe, line 55: zipped_cqes_count

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin: Done

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin: > why skip on valgrind? because too slow to cause zipping?

There were several sporadic errors and the common trait for all these failures was testing with valgrind. So, i think yes, it is too slow for zipping.

> maybe run the test but without validating counters?

Yes, we can do that. Please consider my question above about the time which test should be executed in case of that.

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin: Removed

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin: Without that flush DC and RC transports successfully execute send command several times only during the first `send_while_possible` call and then always return `UCS_ERR_NO_RESOURCE` at the first iteration of send loop.

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+
+#include <vector>
+
+#if defined(ENABLE_STATS)
+#define UCT_INSTANTIATE_MLX5_TEST_CASE(_test_case) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, dc_mlx5) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, ud_mlx5)
+
+struct test_cqe_zipping : public test_uct_ib_with_specific_port {
```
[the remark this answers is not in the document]
yosefe: for consistency

test/gtest/uct/ib/test_cqe_zipping.cc
yosefe: maybe just once

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+    void send_while_possible()
+    {
+        /*
+         * The local deadline prevents hanging inside send loop in case of
+         * connectivity issues
```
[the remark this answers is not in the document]
yosefe: if there is an internal error it will be fatal (abort) or return other error code from the zcopy

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    void wait_for_completion()
+    {
+        /*
+         * The local deadline prevents hanging inside receive loop in case of
+         * connectivity issues
+         */
+        const ucs_time_t deadline = ucs::get_deadline(15);
```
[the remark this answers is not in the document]
yosefe: 60sec
because if all is fine, test will finish fast anyway. and in case of machine load, we don't want random failures.

test/gtest/uct/ib/test_cqe_zipping.cc
yosefe: run once, or up tom 10 sec

test/gtest/uct/ib/test_cqe_zipping.cc
yosefe: ok, pls comment it in the code

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,174 @@
...
+        ASSERT_EQ(send_cnt, recv_cnt);
+    }
+
+    size_t collect_zipped_counters()
+    {
+        std::vector<int> counters = {UCT_IB_IFACE_STAT_TX_COMPLETION_ZIPPED,
+                                     UCT_IB_IFACE_STAT_RX_COMPLETION_ZIPPED};
+        size_t zipped_cqes_cnt = 0;
+
+        for (const auto &peer: peers) {
```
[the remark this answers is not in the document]
yosefe: did it really make a difference or is it an assumption?

test/gtest/uct/ib/test_cqe_zipping.cc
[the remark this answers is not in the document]
ivankochin: So do i understand correctly that you propose to remove timeout inside `send_while_possible` function?

test/gtest/uct/ib/test_cqe_zipping.cc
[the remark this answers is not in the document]
ivankochin: Just an assumption, I haven't tried variant when send operations are performed only on one side.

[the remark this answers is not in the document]
yosefe: if only one side sends, potentially the test could be simpler

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -4,171 +4,158 @@
...
     _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5) \
     _UCT_INSTANTIATE_TEST_CASE(_test_case, dc_mlx5) \
     _UCT_INSTANTIATE_TEST_CASE(_test_case, ud_mlx5)
 
-struct test_cqe_zipping : public test_uct_ib_with_specific_port {
+class test_cqe_zipping : public test_uct_ib_with_specific_port {
+public:
+    static bool check_counters()
+    {
+#if defined(ENABLE_STATS)
```
yosefe, line 23: ifdef ?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+        const ucs_time_t deadline = ucs::get_deadline(60);
+        while ((m_send_cnt != m_recv_cnt) && (ucs_get_time() < deadline)) {
+            progress();
+        }
+        ASSERT_EQ(m_send_cnt, m_recv_cnt);
+    }
+
+    size_t get_total_zipped_count() const
+    {
+#if defined(ENABLE_STATS)
```
yosefe, line 52: ifdef?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+                                     UCT_IB_IFACE_STAT_RX_COMPLETION_ZIPPED};
+        size_t zipped_cqes_count  = 0;
+
+        for (auto &entity : m_entities) {
+            auto stats = ucs_derived_of(entity->iface(), uct_ib_iface_t)->stats;
+            for (auto &counter : counters) {
+                zipped_cqes_count += UCS_STATS_GET_COUNTER(stats, counter);
+            }
+        }
+        return zipped_cqes_count;
```
yosefe, line 63: minor: we could move `size_t zipped_cqes_count = 0;` and `return zipped_cqes_count;` outside the ifdef

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+#else
+        return 0;
+#endif
+    }
+
+    virtual void init()
+    {
+#if defined(ENABLE_STATS)
```
yosefe, line 71: is this if really needed?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+        stats_activate();
+#endif
+        modify_config("IB_CQE_ZIPPING_ENABLE", "y");
+
+        test_uct_ib_with_specific_port::init();
+
+        create_connected_entities();
```
yosefe, line 78: space line after

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+        uct_iface_set_am_handler(receiver()->iface(), 0, am_cb, &m_recv_cnt, 0);
```
yosefe, line 79: where do we remove this am handler?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+        m_send_buf = new mapped_buffer(m_buf_size, 0, *sender());
+    }
+
+    void flush()
```
yosefe, line 83: maybe rename the function to flush_and_reset?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+    {
+        return m_e2;
+    }
+
+    virtual void cleanup()
+    {
+        delete m_send_buf;
+        uct_test::cleanup();
+
+#if defined(ENABLE_STATS)
```
yosefe, line 105: can we move this if inside stats_restore?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+        size_t &recv_cnt = *(static_cast<size_t*>(arg));
+        ++recv_cnt;
+        return UCS_OK;
+    }
+
+    void check_port_attr()
+    {
+        if (IBV_PORT_IS_LINK_LAYER_ETHERNET(&m_port_attr)) {
+            /* Hard to get zipped CQEs on Ethernet link */
+            UCS_TEST_SKIP_R("skip CQE zipping test for Ethernet link type");
```
yosefe, line 132: maybe run the test but don't fail if no zipped cqe?

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+}
+
+#define UCT_INSTANTIATE_MLX5_TEST_CASE(_test_case) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, dc_mlx5) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, ud_mlx5)
+
+class test_cqe_zipping : public test_uct_ib_with_specific_port {
+public:
+    static bool check_counters()
```
yosefe, line 21: is_cqe_zipping_expected

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,161 @@
...
+    while ((ucs_get_time() < deadline) && (get_total_zipped_count() == 0)) {
+        /* Flush resets the transort resources which were acquired during the
+         * previous iteration.
+         */
+        flush();
+        send_while_possible();
+        wait_for_completion();
+    }
+
+    ASSERT_TRUE((!check_counters()) || (get_total_zipped_count() > 0));
```
yosefe, line 158:
```
if (is_cqe_zipping_expected()) {
  EXPECT_GE(get_total_zipped_count(), 0);
}
```

ivankochin: The overall proposal is good, but I think we should use `EXPECT_GT` here, don't we?

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin: Yes, it is. Without this `ifdef` the test will fail with the following error:

```
Invalid UCS configuration for STATS_DEST : file:/dev/null, error message: No such element(-12)
```

I can prevent this error by moving `ifdef` inside `stats_activate` and do the same thing with the `stats_restore` function with regard to [the comment below](https://github.com/openucx/ucx/pull/8104#discussion_r862645562). But I prefer to leave both this calls external and implicit `ifdef` since it make the code more clear and straightforward from my point of view. And in sake of consistency with other use cases of this API of cource.

yosefe: can we move this inside stats_activate/restore functions?

test/gtest/uct/ib/test_cqe_zipping.cc
yosefe: yes, need EXPECT_GT

yosefe (commented):
- move ifdef inside stats_activate/restore
- EXPECT_GE -> EXPECT_GT

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin: Decided to merge this request as is and cleanup this stuff in a separate patch.

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,157 @@
...
+        /* Flush resets the transort resources which were acquired during the
+         * previous iteration.
+         */
+        flush_and_reset();
+        send_while_possible();
+        wait_for_completion();
+    }
+
+    if (is_cqe_zipping_expected()) {
+        EXPECT_GE(get_total_zipped_count(), 0);
```
yosefe, line 153: EXPECT_GT

yosefe (approved):

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -0,0 +1,157 @@
...
+        const ucs_time_t deadline = ucs::get_deadline(60);
+        while ((m_send_cnt != m_recv_cnt) && (ucs_get_time() < deadline)) {
+            progress();
+        }
+        ASSERT_EQ(m_send_cnt, m_recv_cnt);
+    }
+
+    size_t get_total_zipped_count() const
+    {
+        size_t zipped_cqes_count  = 0;
```
gleon99, line 53: `size_t zipped_cqes_count = 0;` (remove space)

ivankochin:
```suggestion
        size_t zipped_cqes_count = 0;
```

ivankochin: Sorry for the applying this, but I saw on this code one more time and found that this double space here is for aligning with assignment on the next line, so it shouldn't be removed.

gleon99:
@yosefe please re-approve.

yosefe (approved):

PR #8106, opened 2022-04-07
DEVX/BUILD: fixed DEVX detection

hoopoepg:
- fixed DEVX detection: DEVX requires dv or devx infrastructure to be
  enabled. in case if no one is found - disable DEVX at all

### DIFF
--- a/src/uct/ib/configure.m4
+++ b/src/uct/ib/configure.m4
@@ -75,7 +75,10 @@ AC_ARG_WITH([dm],
 #
 # DEVX Support
 #
-AC_ARG_WITH([devx], [], [], [with_devx=check])
+AC_ARG_WITH([devx],
+            [AC_HELP_STRING([--with-devx], [Compile with DEVX support])],
+            [],
+            [with_devx=check])
 
 #
 # Check basic IB support: User wanted at least one IB transport, and we found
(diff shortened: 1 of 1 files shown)


src/uct/ib/configure.m4
```diff
@@ -453,3 +454,7 @@ m4_include([src/uct/ib/rdmacm/configure.m4])
 AC_DEFINE_UNQUOTED([uct_ib_MODULES], ["${uct_ib_modules}"], [IB loadable modules])
 AC_CONFIG_FILES([src/uct/ib/Makefile
                  src/uct/ib/ucx-ib.pc])
+AS_IF([test -n "$have_devx" -a \( "x$with_mlx5_dv" = xyes -o "x$with_mlx5_hw" = xyes \) ],
```
yosefe, line 457: we should not have set have_devx to yes when with_mlx5_dv is not yes
IMO it should have been 
`AM_CONDITIONAL([HAVE_DEVX], [test "x$have_devx" = xyes])`

hoopoepg: there are 3 variables tested independently, merging it into single flow will make code really complex: have to add ```have_devx=no``` code into all branches where with_mlx5_dv failed to detect or skipped

yosefe: ok, but not clear why with_mlx5_hw is needed?

yosefe: mlx5_hw comes from exp verbs, and does not include devx

PR #8109, opened 2022-04-07
Add the option to disable rcache invalidation on fork

SeyedMir:
## What
Add the option to disable rcache invalidation on fork.

## Why ?
rcache invalidation on fork can have performance overheads due to extra registration/deregistration. For IB in particular, and in presence of with ibv_fork_init, if the user knows that all buffers that are passed to UCX are page-aligned, there is no need to purge rcache on fork. This PR provides `UCX_IB_RCACHE_PURGE_ON_FORK` to disable IB rcache purge on fork.

### DIFF
--- a/src/uct/base/uct_md.c
+++ b/src/uct/base/uct_md.c
@@ -57,6 +57,10 @@ ucs_config_field_t uct_md_config_rcache_table[] = {
      ucs_offsetof(uct_md_rcache_config_t, max_unreleased),
      UCS_CONFIG_TYPE_MEMUNITS},
 
+    {"RCACHE_PURGE_ON_FORK", "y",
+     "Purge registration cache upon fork",
+     ucs_offsetof(uct_md_rcache_config_t, purge_on_fork), UCS_CONFIG_TYPE_BOOL},
+
     {NULL}
 };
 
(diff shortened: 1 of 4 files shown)


src/uct/ib/base/ib_md.c
```diff
@@ -1292,7 +1292,9 @@ uct_ib_md_parse_reg_methods(uct_ib_md_t *md,
             }
             rcache_params.context            = md;
             rcache_params.ops                = &uct_ib_rcache_ops;
-            rcache_params.flags              = UCS_RCACHE_FLAG_PURGE_ON_FORK;
+            if (md_config->rcache.purge_on_fork) {
+                rcache_params.flags          = UCS_RCACHE_FLAG_PURGE_ON_FORK;
```
yosefe, line 1296: move this code inside uct_md_set_rcache_params
also for knem

src/uct/base/uct_md.c
```diff
@@ -494,6 +498,10 @@ void uct_md_set_rcache_params(ucs_rcache_params_t *rcache_params,
     rcache_params->max_regions        = rcache_config->max_regions;
     rcache_params->max_size           = rcache_config->max_size;
     rcache_params->max_unreleased     = rcache_config->max_unreleased;
+    if (rcache_config->purge_on_fork) {
+        rcache_params->flags          = UCS_RCACHE_FLAG_PURGE_ON_FORK;
```
yosefe, line 502: 1. should be set to 0 if not purge on fork
2. does it mean all registration caches will now have purge-on-fork by default?

SeyedMir: This will enable purge-on-fork for all md rcaches by default. Though, I should also update gdr_copy too https://github.com/openucx/ucx/blob/7d8e496e9b397bfb7a5a4cd79f2497cf8bdd94f6/src/uct/cuda/gdr_copy/gdr_copy_md.c#L417

Alternatively, I can add the option only to the IB md by defining an extended rcache config type for the IB md.
https://github.com/openucx/ucx/blob/7d8e496e9b397bfb7a5a4cd79f2497cf8bdd94f6/src/uct/ib/base/ib_md.h#L151

yosefe (dismissed):

src/uct/base/uct_md.c
yosefe: ok, looks ok as it is now. can u pls squash to one commit?

yosefe (approved):

PR #6557, opened 2021-03-25
UCP/API: Added name to worker parameters.

rakhmets:
## What
Added worker name to UCP worker parameters to identify the worker.
Added worker name to UCP worker attributes to get the actual worker name.

## Why ?
The parameter is required for more convenient run time analysis of UCX-based application. E.g. ucx_vfs tool.

PR #8110, opened 2022-04-09
UCP/RNDV: Define min chunk size for multi-rail

brminich:
## What
Define minimal size of fragment when splitting message over multiple lanes

## Why ?
To not split the message into small chunks (which degrades performance)

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -157,6 +157,14 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "protocol. Lanes slower than the specified ratio will not be used.",
    ucs_offsetof(ucp_context_config_t, multi_lane_max_ratio), UCS_CONFIG_TYPE_DOUBLE},
 
+  {"MULTI_PATH_RATIO", "auto",
+   "Bandwidth efficiency ratio when more than one path per device is used.\n"
+   "This value represents the fraction of bandwidth taken by each connection\n"
+   "on the same device. A value of 'auto' means that fraction is calculated\n"
+   "based on the maximal number of paths supported by the device.",
+   ucs_offsetof(ucp_context_config_t, multi_path_ratio),
+   UCS_CONFIG_TYPE_POS_DOUBLE},
+
   {"MAX_EAGER_LANES", NULL, "",
    ucs_offsetof(ucp_context_config_t, max_eager_lanes), UCS_CONFIG_TYPE_UINT},
 
@@ -170,6 +178,11 @@ static ucs_config_field_t ucp_context_config_table[] = {
   {"MAX_RNDV_RAILS", "2",
    "Maximal number of devices on which a rendezvous operation may be executed in parallel",
    ucs_offsetof(ucp_context_config_t, max_rndv_lanes), UCS_CONFIG_TYPE_UINT},
+
+  {"MIN_RNDV_CHUNK_SIZE", "16k",
+   "Minimum chunk size to split the message sent with rendezvous protocol on\n"
+   "multiple rails. Must be greater than 0.",
+   ucs_offsetof(ucp_context_config_t, min_rndv_chunk_size), UCS_CONFIG_TYPE_MEMUNITS},
 
   {"RNDV_SCHEME", "auto",
    "Communication scheme in RNDV protocol.\n"
@@ -1648,7 +1661,7 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
     ucs_debug("estimated number of endpoints per node is %d",
               context->config.est_num_ppn);
 
-    if (UCS_CONFIG_BW_IS_AUTO(context->config.ext.bcopy_bw)) {
+    if (UCS_CONFIG_DBL_IS_AUTO(context->config.ext.bcopy_bw)) {
         /* bcopy_bw wasn't set via the env variable. Calculate the value */
         context->config.ext.bcopy_bw = ucs_cpu_get_memcpy_bw();
     }
@@ -1668,6 +1681,12 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
                 context->proto_bitmap |= UCS_BIT(proto_id);
             }
         }
+    }
+
+    if (context->config.ext.min_rndv_chunk_size == 0) {
+        ucs_error("minimum chunk size for rendezvous protocol must be greater"
+                  " than 0");
+        return UCS_ERR_INVALID_PARAM;
     }
 
     /* Save environment prefix to later notify user for unused variables */
--- a/src/ucp/rndv/rndv.c
+++ b/src/ucp/rndv/rndv.c
@@ -666,8 +666,9 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
     ucp_md_index_t dst_md_index;
     ucp_rsc_index_t rsc_index;
     uct_iface_attr_t *iface_attr;
-    double max_lane_bw, lane_bw;
-    int i;
+    double max_lane_bw, lane_bw, max_ratio;
+    size_t chunk_count;
+    int i, lanes_count;
 
     ucs_assert((proto == UCP_REQUEST_SEND_PROTO_RNDV_GET) ||
                (proto == UCP_REQUEST_SEND_PROTO_RNDV_PUT));
@@ -723,23 +724,41 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
         }
     }
 
-    if (ucs_popcount(lane_map) > 1) {
-        /* remove lanes if bandwidth is too less compare to best lane */
-        ucs_for_each_bit(lane_idx, lane_map) {
-            ucs_assert(lane_idx < UCP_MAX_LANES);
-            lane       = lanes[lane_idx];
-            rsc_index  = ep_config->key.lanes[lane].rsc_index;
-            iface_attr = ucp_worker_iface_get_attr(ep->worker, rsc_index);
-            lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+    if (ucs_is_pow2_or_zero(lane_map)) {
+        goto out; /* No multi-rail */
+    }
 
-            if ((lane_bw / max_lane_bw) <
-                (1. / context->config.ext.multi_lane_max_ratio)) {
-                lane_map                                &= ~UCS_BIT(lane_idx);
-                rndv_req->send.rndv.rkey_index[lane_idx] = UCP_NULL_RESOURCE;
-            }
+    chunk_count = ucs_max(rndv_req->send.length /
+                              context->config.ext.min_rndv_chunk_size,
+                          1);
+    lanes_count = 0;
+    max_ratio   = 1. / context->config.ext.multi_lane_max_ratio;
+
+    /* 1. Remove lanes if bandwidth is too small compared to the best lane.
+     * 2. If there are more lanes than chunks of the minimum size, we have to
+     * exclude some lanes, so that there will be no more than 1 lane per
+     * chunk.
+     */
+    ucs_for_each_bit(lane_idx, lane_map) {
+        ucs_assert(lane_idx < UCP_MAX_LANES);
+        lane       = lanes[lane_idx];
+        rsc_index  = ep_config->key.lanes[lane].rsc_index;
+        iface_attr = ucp_worker_iface_get_attr(ep->worker, rsc_index);
+        lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+        ++lanes_count;
+
+        if (((lane_bw / max_lane_bw) < max_ratio) ||
+            (lanes_count > chunk_count)) {
+            lane_map                                &= ~UCS_BIT(lane_idx);
+            rndv_req->send.rndv.rkey_index[lane_idx] = UCP_NULL_RESOURCE;
         }
     }
 
+    ucs_assertv(ucs_popcount(lane_map) <= chunk_count,
+                "lanes_count=%u chunk_count=%zu", ucs_popcount(lane_map),
+                chunk_count);
+
+out:
     ucp_rndv_req_init_lanes(rndv_req, lane_map, ucs_popcount(lane_map));
 }
 
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -65,6 +65,12 @@ typedef struct {
 } ucp_wireup_select_bw_info_t;
 
 
+typedef struct {
+    unsigned local[UCP_MAX_RESOURCES];
+    unsigned remote[UCP_MAX_RESOURCES];
+} ucp_wireup_dev_usage_count;
+
+
 /**
  * Global parameters for lanes selection during UCP wireup procedure
  */
@@ -328,6 +334,7 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
     const ucp_address_entry_t *ae;
     ucp_worker_iface_t *wiface;
     ucp_rsc_index_t rsc_index;
+    ucp_rsc_index_t dev_index;
     ucp_lane_index_t lane;
     char tls_info[256];
     char *p, *endp;
@@ -403,6 +410,7 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
     UCS_BITMAP_FOR_EACH_BIT(tl_bitmap, rsc_index) {
         local_md_flags = criteria->local_md_flags;
         resource       = &context->tl_rscs[rsc_index].tl_rsc;
+        dev_index      = context->tl_rscs[rsc_index].dev_index;
         wiface         = ucp_worker_iface(worker, rsc_index);
         iface_attr     = ucp_worker_iface_get_attr(worker, rsc_index);
         md_attr        =
@@ -471,8 +479,7 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
                      UCT_TL_RESOURCE_DESC_ARG(resource), criteria->title);
             p += strlen(p);
             continue;
-        } else if (!(local_dev_bitmap &
-                     UCS_BIT(context->tl_rscs[rsc_index].dev_index))) {
+        } else if (!(local_dev_bitmap & UCS_BIT(dev_index))) {
             ucs_trace(UCT_TL_RESOURCE_DESC_FMT " : disabled by device bitmap",
                       UCT_TL_RESOURCE_DESC_ARG(resource));
             snprintf(p, endp - p, UCT_TL_RESOURCE_DESC_FMT" - disabled for %s, ",
@@ -502,6 +509,7 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
         }
 
         is_reachable = 0;
+
         UCS_BITMAP_FOR_EACH_BIT(rsc_addr_index_map, addr_index) {
             ae = &address->address_list[addr_index];
             if (!ucp_wireup_is_reachable(ep, select_params->ep_init_flags,
@@ -510,14 +518,15 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
                 continue;
             }
 
-            score        = criteria->calc_score(wiface, md_attr,
-                                                &ae->iface_attr);
+            score        = criteria->calc_score(wiface, md_attr, ae,
+                                                criteria->arg);
             priority     = iface_attr->priority + ae->iface_attr.priority;
+            is_reachable = 1;
 
-            ucs_trace(UCT_TL_RESOURCE_DESC_FMT "->addr[%u] : %s score %.2f priority %d",
+            ucs_trace(UCT_TL_RESOURCE_DESC_FMT
+                      "->addr[%u] : %s score %.2f priority %d",
                       UCT_TL_RESOURCE_DESC_ARG(resource),
                       addr_index, criteria->title, score, priority);
-            is_reachable = 1;
 
             if (!found || (ucp_score_prio_cmp(score, priority, sinfo.score,
                                               sinfo.priority) > 0)) {
@@ -715,7 +724,7 @@ static int ucp_wireup_compare_lane_rma_score(const void *elem1, const void *elem
 }
 
 static int ucp_wireup_compare_lane_rma_bw_score(const void *elem1, const void *elem2,
-                                             void *arg)
+                                                void *arg)
 {
     return ucp_wireup_compare_score(elem1, elem2, arg, UCP_LANE_TYPE_RMA_BW);
 }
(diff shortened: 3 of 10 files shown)


brminich:
@rakhmets, @hoopoepg, @dmitrygx can you please review?

src/ucp/core/ucp_context.c
```diff
@@ -243,6 +243,11 @@ static ucs_config_field_t ucp_config_table[] = {
    "Maximal number of devices on which a rendezvous operation may be executed in parallel",
    ucs_offsetof(ucp_config_t, ctx.max_rndv_lanes), UCS_CONFIG_TYPE_UINT},
 
+  {"MIN_RNDV_CHUNK_SIZE", "16kb",
```
hoopoepg, line 246: minor: in other sizes in kb used "16k" format

src/ucp/core/ucp_context.c
```diff
@@ -243,6 +243,11 @@ static ucs_config_field_t ucp_config_table[] = {
...
+   "Minimal chunk size to split the message sent with rendezvous protocol on\n"
+   " multiple rails. Must be bigger than 0.",
```
hoopoepg, line 248: space on left of ```multiple``` is not needed

src/ucp/rndv/rndv.c
```diff
@@ -721,23 +724,70 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
+        } else {
+            md_map |= UCS_BIT(ep_config->md_index[lane]);
+        }
+    }
+
+    chunk_count = ucs_max(rndv_req->send.length /
+                              context->config.ext.min_rndv_chunk_size,
+                          1);
+
+    /* If there are more lanes than chunks of the mimimal size, we have to
```
hoopoepg, line 757: s/mimimal/minimal

src/ucp/core/ucp_context.c
brminich: ok

src/ucp/core/ucp_context.c
brminich: ok

src/ucp/rndv/rndv.c
```diff
@@ -721,23 +724,70 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
+        lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+
+        if ((lane_bw / max_lane_bw) <
+            (1. / context->config.ext.multi_lane_max_ratio)) {
+            lane_map                                &= ~UCS_BIT(lane_idx);
+            rndv_req->send.rndv.rkey_index[lane_idx] = UCP_NULL_RESOURCE;
+            continue;
+        }
+
+        if (UCS_BIT(ep_config->md_index[lane]) & md_map) {
```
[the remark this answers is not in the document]
brminich: ok

src/ucp/rndv/rndv.c
```diff
@@ -721,23 +724,70 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
+            md_map |= UCS_BIT(ep_config->md_index[lane]);
+        }
+    }
+
+    chunk_count = ucs_max(rndv_req->send.length /
+                              context->config.ext.min_rndv_chunk_size,
+                          1);
+
+    /* If there are more lanes than chunks of the mimimal size, we have to
+     * exclude some lanes, so that there will be no more that 1 lane per chunk.
```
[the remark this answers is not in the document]
brminich: i meant it like here https://en.wiktionary.org/wiki/so_that

src/ucp/rndv/rndv.c
brminich: ok

src/ucp/rndv/rndv.c
```diff
@@ -721,23 +724,70 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
-
-            if ((lane_bw / max_lane_bw) <
-                (1. / context->config.ext.multi_lane_max_ratio)) {
-                lane_map                                &= ~UCS_BIT(lane_idx);
-                rndv_req->send.rndv.rkey_index[lane_idx] = UCP_NULL_RESOURCE;
+            if (lanes_count++ == chunk_count) {
+                break;
             }
+            lane_map |= UCS_BIT(lane_idx);
         }
```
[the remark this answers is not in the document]
brminich: ok

src/ucp/core/ucp_context.c
```diff
@@ -243,6 +243,11 @@ static ucs_config_field_t ucp_config_table[] = {
    "Maximal number of devices on which a rendezvous operation may be executed in parallel",
    ucs_offsetof(ucp_config_t, ctx.max_rndv_lanes), UCS_CONFIG_TYPE_UINT},
 
+  {"MIN_RNDV_CHUNK_SIZE", "16kb",
+   "Minimal chunk size to split the message sent with rendezvous protocol on\n"
```
rakhmets, line 247: I'd prefer to use `minimum` instead `minimal`. I know that there are many `minimal` in the documentation and comments in the code. However, once I got the following comment https://github.com/openucx/ucx/pull/6557#discussion_r604469311, I use `maximum` instead `maximal`. ( But I'm not 100% sure that minimum/minimal is the same case).

src/ucp/core/ucp_context.c
rakhmets, line 248: bigger -> greater

src/ucp/core/ucp_context.c
```diff
@@ -1644,6 +1649,12 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
         }
     }
 
+    if (context->config.ext.min_rndv_chunk_size == 0) {
+        ucs_error("Minimal chunk size for rendezvous protocol must be bigger"
+                  " than 0");
```
rakhmets, line 1654: The same as above.

src/ucp/rndv/rndv.c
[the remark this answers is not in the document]
dmitrygx: sorry I wasn't clear - didn't noticed usage of `that` twice here:
```suggestion
     * exclude some lanes, so that there will be no more than 1 lane per chunk.
```

src/ucp/rndv/rndv.c
```diff
@@ -721,23 +724,70 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
+
+    /* Remove lanes if bandwidth is too small compared to the best lane */
+    ucs_for_each_bit(lane_idx, lane_map) {
+        ucs_assert(lane_idx < UCP_MAX_LANES);
+        lane       = lanes[lane_idx];
+        rsc_index  = ep_config->key.lanes[lane].rsc_index;
+        iface_attr = ucp_worker_iface_get_attr(ep->worker, rsc_index);
+        lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+
+        if ((lane_bw / max_lane_bw) <
```
rakhmets, line 739: If lane indexes are ordered according to their bandwidth, why do we go further after facing first low-bw lane?

src/ucp/rndv/rndv.c
```diff
@@ -721,23 +724,70 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
+    }
+
+    chunk_count = ucs_max(rndv_req->send.length /
+                              context->config.ext.min_rndv_chunk_size,
+                          1);
+
+    /* If there are more lanes than chunks of the mimimal size, we have to
+     * exclude some lanes, so that there will be no more that 1 lane per chunk.
+     */
+    if (ucs_popcount(lane_map) > chunk_count) {
```
rakhmets, line 760: Actually, we know maximum number of bits in lane_map (even before executing the previous loop).

And we know how many bits we should zero at this line of the code. Wouldn't that be a simplification?

brminich: can you please elaborate?

src/ucp/rndv/rndv.c
brminich: ok, it's possible to mask out all following lanes using `lane_map & = UCS_MASK(lane_idx)` instead of iterating, but we have to fill `rndv_req->send.rndv.rkey_index` with `UCP_NULL_RESOURCE` anyway.
Also please note that the number of lanes is supposed to be quite low number (currently up to 4)

dmitrygx: if I'm not mistaken, lanes are not sorted by their BW characteristics

dmitrygx (approved):

src/ucp/rndv/rndv.c
brminich: they are taken from rma_bw_lanes, which are sorted

src/ucp/rndv/rndv.c
```diff
@@ -723,23 +724,41 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
-            lane       = lanes[lane_idx];
-            rsc_index  = ep_config->key.lanes[lane].rsc_index;
-            iface_attr = ucp_worker_iface_get_attr(ep->worker, rsc_index);
-            lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+    if (ucs_popcount(lane_map) <= 1) {
+        goto out;
+    }
+
+    chunk_count = ucs_max(rndv_req->send.length /
+                              context->config.ext.min_rndv_chunk_size,
```
rakhmets, line 732: Please add 4 additional spaces.

Please review `format code` output. There are a couple more suggestions.

rakhmets (dismissed):

src/ucp/rndv/rndv.c
```diff
@@ -722,23 +723,41 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
 
-    if (ucs_popcount(lane_map) > 1) {
-        /* remove lanes if bandwidth is too less compare to best lane */
-        ucs_for_each_bit(lane_idx, lane_map) {
-            ucs_assert(lane_idx < UCP_MAX_LANES);
-            lane       = lanes[lane_idx];
-            rsc_index  = ep_config->key.lanes[lane].rsc_index;
-            iface_attr = ucp_worker_iface_get_attr(ep->worker, rsc_index);
-            lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+    if (ucs_popcount(lane_map) <= 1) {
```
yosefe, line 726: maybe check ucs_is_pow2 ?
for perf

src/ucp/rndv/rndv.c
```diff
@@ -722,23 +723,41 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
-                (1. / context->config.ext.multi_lane_max_ratio)) {
-                lane_map                                &= ~UCS_BIT(lane_idx);
-                rndv_req->send.rndv.rkey_index[lane_idx] = UCP_NULL_RESOURCE;
-            }
+    /* 1. Remove lanes if bandwidth is too small compared to the best lane.
+     * 2. If there are more lanes than chunks of the minimum size, we have to
+     * exclude some lanes, so that there will be no more than 1 lane per
+     * chunk.
+     */
+    ucs_for_each_bit(lane_idx, lane_map) {
```
yosefe, line 741: BTW this seems similar to waht @karasevb is doing for minimal size for alignment split
maybe makes sense to unite the config?

src/ucp/rndv/rndv.c
```diff
@@ -722,23 +723,41 @@ static void ucp_rndv_req_init_zcopy_lane_map(ucp_request_t *rndv_req,
...
+        ucs_assert(lane_idx < UCP_MAX_LANES);
+        lane       = lanes[lane_idx];
+        rsc_index  = ep_config->key.lanes[lane].rsc_index;
+        iface_attr = ucp_worker_iface_get_attr(ep->worker, rsc_index);
+        lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+        ++lanes_count;
+
+        if (((lane_bw / max_lane_bw) < max_ratio) ||
+            (lanes_count > chunk_count)) {
```
yosefe, line 750: can we exit loop when lanes_count == chunk_count ?

src/ucp/rndv/rndv.c
brminich: 1 would also pass this check. Also `ucs_is_pow2` has extra branch

src/ucp/rndv/rndv.c
brminich: the thresholds have a bit different meaning and need to have different default values

src/ucp/wireup/select.c
```diff
@@ -510,14 +519,19 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
                 continue;
             }
 
-            score        = criteria->calc_score(wiface, md_attr,
-                                                &ae->iface_attr);
+            score_factor = local_factor *
```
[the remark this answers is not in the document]
brminich: but we still set it

src/ucp/rndv/rndv.c
brminich: we could, but here we also do
`rndv_req->send.rndv.rkey_index[lane_idx] = UCP_NULL_RESOURCE`
for the unused lanes
I'd expect very low perf overhead, because of small number of lanes

yosefe: ok

src/ucp/rndv/rndv.c
yosefe: popcount is expensive

src/ucp/wireup/select.c
```diff
@@ -510,14 +537,19 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
...
+                           ucp_wireup_score_factor(wiface, remote_dev_counts,
```
yosefe, line 541: can we make ucp_wireup_score_factor part of calc_score callback so the common function ucp_wireup_select_transport would not need to change?

src/uct/api/v2/uct_v2.h
```diff
@@ -187,6 +190,9 @@ typedef struct {
      * This field is set by the UCT layer.
      */
     size_t              max_inflight_eps;
+
+    /* Bandwidth efficiency ratio when more than one path per device is used. */
```
yosefe, line 194: need to elaborate that it's an "exponential" value

src/uct/ib/base/ib_iface.c
```diff
@@ -1050,7 +1052,8 @@ static void uct_ib_iface_set_num_paths(uct_ib_iface_t *iface,
 
         if ((iface->num_paths == 1) &&
             (uct_ib_iface_port_attr(iface)->active_speed == UCT_IB_SPEED_NDR)) {
-            iface->num_paths = 2;
+            iface->num_paths     = 2;
+            iface->bw_path_ratio = 0.9;
```
yosefe, line 1056: can we calculate it in uct_ib_iface_estimate_perf instead of saving it on the iface?

src/ucp/core/ucp_context.c
```diff
@@ -157,6 +157,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "protocol. Lanes slower than the specified ratio will not be used.",
    ucs_offsetof(ucp_context_config_t, multi_lane_max_ratio), UCS_CONFIG_TYPE_DOUBLE},
 
+  {"MULTI_PATH_RATIO", "1",
```
yosefe, line 160: maybe make it simpler, just define "by-device" or "by-path"?
it would allow simplifying transport select logic

src/ucp/wireup/select.c
```diff
@@ -308,8 +316,8 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
         const ucp_wireup_select_params_t *select_params,
         const ucp_wireup_criteria_t *criteria, ucp_tl_bitmap_t tl_bitmap,
         uint64_t remote_md_map, uint64_t local_dev_bitmap,
-        uint64_t remote_dev_bitmap, int show_error,
-        ucp_wireup_select_info_t *select_info)
+        uint64_t remote_dev_bitmap, int show_error, unsigned *local_dev_counts,
+        unsigned *remote_dev_counts, ucp_wireup_select_info_t *select_info)
```
yosefe, line 320: this function is already complicated. i think device counts logic should be outside of it.

src/ucp/core/ucp_context.c
brminich: Imo, this config is more flexible.  What if we have CX5 + CX7 devices on the system? Or IB + low-speed Eth?

yosefe: eventually we want this to be exposed from UCT layer, so not sure having this in UCP is so useful
IMO UCP better have a more "strict"/"forceful" config that easy to understand how they affect the end result, and UCT could expose fine tunings of its performance estimations (similar to shared memory BW, PCIe BW , etc)

src/ucp/core/ucp_context.c
```diff
@@ -1670,6 +1683,12 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
         }
     }
 
+    if (context->config.ext.min_rndv_chunk_size == 0) {
+        ucs_error("Minimum chunk size for rendezvous protocol must be greater"
```
yosefe, line 1687: smallcase

src/ucs/sys/string.h
```diff
@@ -35,6 +35,7 @@ BEGIN_C_DECLS
 #define UCS_MEMUNITS_AUTO   ((size_t)-2)
 #define UCS_ULUNITS_AUTO    ((unsigned long)-2)
 #define UCS_HEXUNITS_AUTO   ((uint16_t)-2)
+#define UCS_DOUBLE_AUTO     DBL_MAX
```
yosefe, line 38: seems unused

src/ucp/wireup/select.c
```diff
@@ -502,6 +505,7 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
         }
 
         is_reachable = 0;
+
```
yosefe, line 508: can we revert the changes in this function? seems cosmetic/leftover

src/ucp/wireup/select.c
```diff
@@ -827,19 +831,20 @@ static uint64_t ucp_ep_get_context_features(const ucp_ep_h ep)
     return ep->worker->context->config.features;
 }
 
-static double ucp_wireup_rma_score_func(const ucp_worker_iface_t *wiface,
+static double ucp_wireup_rma_score_func(const ucp_wireup_criteria_t *self,
```
yosefe, line 834: maybe pass 'void*' user-defined argument instead of self (also, as last param) ?

src/ucp/wireup/select.c
```diff
@@ -1039,30 +1046,74 @@ ucp_wireup_add_amo_lanes(const ucp_wireup_select_params_t *select_params,
...
-                   wiface->attr.overhead + remote_iface_attr->overhead);
+                                               &remote_addr->iface_attr) +
+                   wiface->attr.overhead + remote_addr->iface_attr.overhead);
+}
+
+static double ucp_tl_iface_bandwidth_ratio(ucp_context_h context,
+                                           unsigned dev_count,
+                                           unsigned num_paths)
+{
+    double ratio = context->config.ext.multi_path_ratio * dev_count;
```
yosefe, line 1065: why need to initialize?

src/ucp/wireup/select.c
```diff
@@ -1039,30 +1046,74 @@ ucp_wireup_add_amo_lanes(const ucp_wireup_select_params_t *select_params,
...
+
+    if (UCS_CONFIG_DBL_IS_AUTO(context->config.ext.multi_path_ratio)) {
+        ratio = (double)dev_count / (double)num_paths;
+    } else {
+        ratio = context->config.ext.multi_path_ratio * dev_count;
+    }
+
+    return ucs_max(0.1, 1.0 - ratio);
```
yosefe, line 1073: why 0.1 and not 0 ?

src/ucp/wireup/select.c
```diff
@@ -1039,30 +1046,74 @@ ucp_wireup_add_amo_lanes(const ucp_wireup_select_params_t *select_params,
...
+}
+
+static double ucp_tl_iface_bandwidth_ratio(ucp_context_h context,
+                                           unsigned dev_count,
+                                           unsigned num_paths)
+{
+    double ratio = context->config.ext.multi_path_ratio * dev_count;
+
+    if (UCS_CONFIG_DBL_IS_AUTO(context->config.ext.multi_path_ratio)) {
+        ratio = (double)dev_count / (double)num_paths;
```
yosefe, line 1068: minor: do not need to cast dev_count, it's enough to case num_paths

src/ucp/wireup/select.c
brminich: it makes divider to be 0 for score calculations. I'd use some small value, indicating that some bw does exist for extra connections

src/ucp/wireup/select.c
brminich: not all of them, i modify `calc_score` params. others are quite trivial, remove anyway?

src/ucp/wireup/select.c
brminich: leftover

src/ucp/wireup/select.c
yosefe: maybe convert to multiplier so score would be 0 in this case?
if cant set something like 1e-5

brminich: ok

src/ucp/wireup/select.c
yosefe: ok

src/ucp/wireup/select.c
brminich: btw, shouldn't it be a part of criteria? Otherwise we will need to define new structure controlling lanes selection and only bw-related functions will pass it to the general selection routine  `ucp_wireup_select_transport`  (others will have to pass `NULL`)

yosefe: anyway the BW stuff should not be part of the general criteria so need to define a new struct

src/ucp/wireup/select.c
```diff
@@ -1258,9 +1255,11 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
                         ucp_tl_bitmap_t tl_bitmap, ucp_lane_index_t excl_lane,
                         ucp_wireup_select_context_t *select_ctx)
 {
-    ucp_ep_h ep                    = select_params->ep;
-    ucp_context_h context          = ep->worker->context;
-    ucp_wireup_select_info_t sinfo = {0};
+    ucp_ep_h ep                               = select_params->ep;
+    ucp_context_h context                     = ep->worker->context;
+    ucp_wireup_select_info_t sinfo            = {0};
+    unsigned dev_count[UCP_MAX_RESOURCES * 2] = {0};
```
yosefe, line 1261: IMO it would be better/obvious to have a struct with 2 arrays and not split same array to 2 parts

src/ucp/wireup/select.c
```diff
@@ -65,6 +65,12 @@ typedef struct {
 } ucp_wireup_select_bw_info_t;
 
 
+typedef struct {
+    unsigned     local[UCP_MAX_RESOURCES];
```
yosefe, line 69: can use less spaces

yosefe: pls squash after fix

brminich: actually I thought to use more (like in all other structures in this file).
leave just 1?

yosefe (approved):

PR #8113, opened 2022-04-10
UCP/CORE: Fix keepalive selection logic

dmitrygx:
## What

Fix keepalive selection logic.

## Why ?

UCP EP should select EP_CHECK only for lanes from `p2p_lanes` map, otherwise - use AM-based keepalive.

## How ?

1. Update `ucp_ep_is_am_keepalive` and `ucp_ep_do_uct_ep_keepalive` to accept lane on which keepalive should be done.
2. Update `ucp_ep_is_am_keepalive` to check whether `lane` exists in `p2p_lanes` or not. If doesn't exist there, use AM-based keepalive.

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -3241,27 +3241,21 @@ void ucp_ep_invoke_err_cb(ucp_ep_h ep, ucs_status_t status)
     ucp_ep_ext_control(ep)->err_cb(ucp_ep_ext_gen(ep)->user_data, ep, status);
 }
 
-static UCS_F_ALWAYS_INLINE int
-ucp_ep_is_am_keepalive(ucp_ep_h ep, ucp_rsc_index_t rsc_idx)
+int ucp_ep_is_am_keepalive(ucp_ep_h ep, ucp_rsc_index_t rsc_index, int is_p2p)
 {
-    ucp_worker_iface_t *wiface;
-
-    if (!(ep->flags & UCP_EP_FLAG_REMOTE_ID) ||
-        (rsc_idx == UCP_NULL_RESOURCE)) {
-        /* if remote ID isn't defined or rsc index is NULL (i.e. it is CM lane),
-         * don't do AM keepalive */
-        return 0;
-    }
-
-    wiface = ucp_worker_iface(ep->worker, rsc_idx);
-    return ucs_test_all_flags(wiface->attr.cap.flags,
-                              UCT_IFACE_FLAG_CONNECT_TO_IFACE |
-                              UCT_IFACE_FLAG_AM_BCOPY);
+    return /* Not a CM lane */
+            (rsc_index != UCP_NULL_RESOURCE) &&
+            /* Have a remote endpoint ID to send in the keepalive active message */
+            (ep->flags & UCP_EP_FLAG_REMOTE_ID) &&
+            /* Transport is not connected as point-to-point */
+            !is_p2p &&
+            /* Transport supports active messages */
+            (ucp_worker_iface(ep->worker, rsc_index)->flags &
+             UCT_IFACE_FLAG_AM_BCOPY);
 }
 
-ucs_status_t ucp_ep_do_uct_ep_keepalive(ucp_ep_h ucp_ep, uct_ep_h uct_ep,
-                                        ucp_rsc_index_t rsc_idx, unsigned flags,
-                                        uct_completion_t *comp)
+ucs_status_t ucp_ep_do_uct_ep_am_keepalive(ucp_ep_h ucp_ep, uct_ep_h uct_ep,
+                                           ucp_rsc_index_t rsc_idx)
 {
     ucp_tl_bitmap_t tl_bitmap = UCS_BITMAP_ZERO;
     ucs_status_t status;
@@ -3270,10 +3264,6 @@ ucs_status_t ucp_ep_do_uct_ep_keepalive(ucp_ep_h ucp_ep, uct_ep_h uct_ep,
     ucp_wireup_msg_t wireup_msg;
 
     ucs_assert(!(ucp_ep->flags & UCP_EP_FLAG_FAILED));
-
-    if (!ucp_ep_is_am_keepalive(ucp_ep, rsc_idx)) {
-        return uct_ep_check(uct_ep, flags, comp);
-    }
 
     UCS_BITMAP_SET(tl_bitmap, rsc_idx);
 
--- a/src/ucp/core/ucp_worker.c
+++ b/src/ucp/core/ucp_worker.c
@@ -3175,13 +3175,17 @@ static int ucp_worker_do_ep_keepalive(ucp_worker_h worker, ucs_time_t now)
     ucs_trace("ep %p: do keepalive on lane[%d]=%p ep->flags=0x%x", ep, lane,
               ep->uct_eps[lane], ep->flags);
 
-    status = ucp_ep_do_uct_ep_keepalive(ep, ep->uct_eps[lane], rsc_index, 0,
-                                        NULL);
-    if (status == UCS_ERR_NO_RESOURCE) {
-        return 0;
+    if (ucp_ep_is_am_keepalive(ep, rsc_index,
+                               ucp_ep_config(ep)->p2p_lanes & UCS_BIT(lane))) {
+        status = ucp_ep_do_uct_ep_am_keepalive(ep, ep->uct_eps[lane],
+                                               rsc_index);
+    } else {
+        status = uct_ep_check(ep->uct_eps[lane], 0, NULL);
     }
 
-    if (status != UCS_OK) {
+    if (status == UCS_ERR_NO_RESOURCE) {
+        return 0;
+    } else if (status != UCS_OK) {
         ucs_diag("worker %p: keepalive failed on ep %p lane[%d]=%p: %s",
                  worker, ep, lane, ep->uct_eps[lane],
                  ucs_status_string(status));
--- a/src/ucp/wireup/wireup_ep.c
+++ b/src/ucp/wireup/wireup_ep.c
@@ -260,14 +260,18 @@ UCS_CLASS_DEFINE_NAMED_NEW_FUNC(ucp_wireup_ep_create, ucp_wireup_ep_t, uct_ep_t,
                                 ucp_ep_h);
 
 void ucp_wireup_ep_set_aux(ucp_wireup_ep_t *wireup_ep, uct_ep_h uct_ep,
-                           ucp_rsc_index_t rsc_index)
+                           ucp_rsc_index_t rsc_index, int is_p2p)
 {
     ucs_assert(!ucp_wireup_ep_test(uct_ep));
     wireup_ep->aux_ep        = uct_ep;
     wireup_ep->aux_rsc_index = rsc_index;
+
+    if (is_p2p) {
+        wireup_ep->flags |= UCP_WIREUP_EP_FLAG_AUX_P2P;
+    }
 }
 
-ucs_status_t
+static ucs_status_t
 ucp_wireup_ep_connect_aux(ucp_wireup_ep_t *wireup_ep, unsigned ep_init_flags,
                           const ucp_unpacked_address_t *remote_address)
 {
@@ -306,7 +310,7 @@ ucp_wireup_ep_connect_aux(ucp_wireup_ep_t *wireup_ep, unsigned ep_init_flags,
         return status;
     }
 
-    ucp_wireup_ep_set_aux(wireup_ep, uct_ep, select_info.rsc_index);
+    ucp_wireup_ep_set_aux(wireup_ep, uct_ep, select_info.rsc_index, 0);
 
     ucp_worker_iface_progress_ep(wiface);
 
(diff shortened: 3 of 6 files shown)


src/ucp/core/ucp_worker.c
```diff
@@ -3175,8 +3175,8 @@ static int ucp_worker_do_ep_keepalive(ucp_worker_h worker, ucs_time_t now)
     ucs_trace("ep %p: do keepalive on lane[%d]=%p ep->flags=0x%x", ep, lane,
               ep->uct_eps[lane], ep->flags);
 
-    status = ucp_ep_do_uct_ep_keepalive(ep, ep->uct_eps[lane], rsc_index, 0,
-                                        NULL);
+    status = ucp_ep_do_uct_ep_keepalive(ep, ep->uct_eps[lane], lane, rsc_index,
```
yosefe, line 3178: no need to pass uct_ep, lane and rsc_index to all functions now
we can get them from ep (lane is always keepalive_lane)

src/ucp/core/ucp_ep.c
```diff
@@ -3253,14 +3254,17 @@ ucp_ep_is_am_keepalive(ucp_ep_h ep, ucp_rsc_index_t rsc_idx)
     }
 
     wiface = ucp_worker_iface(ep->worker, rsc_idx);
-    return ucs_test_all_flags(wiface->attr.cap.flags,
+    return (lane != UCP_NULL_LANE) &&
+           !(ucp_ep_config(ep)->p2p_lanes & UCS_BIT(lane)) &&
```
yosefe, line 3258: 1. can we add ucp_ep_config(ep)->is_am_keepalive and set it during ucp_ep_config_init?
2. do we still need to check UCT_IFACE_FLAG_CONNECT_TO_IFACE below?

src/ucp/wireup/wireup_ep.c
```diff
@@ -363,8 +363,8 @@ ucp_wireup_ep_do_check(ucp_ep_h ucp_ep, uct_ep_h uct_ep,
 
     wiface = ucp_worker_iface(worker, rsc_idx);
     if (wiface->attr.cap.flags & UCT_IFACE_FLAG_EP_CHECK) {
-        return ucp_ep_do_uct_ep_keepalive(ucp_ep, uct_ep, rsc_idx, flags,
-                                          comp);
+        return ucp_ep_do_uct_ep_keepalive(ucp_ep, uct_ep, UCP_NULL_LANE,
```
yosefe, line 366: can we make it so during the initial configuration - before CM is connected, and am_lane points to CM - keepalive_lane will also point to cm lane
then, no need to pass anything special here, ucp_ep_do_uct_ep_keepalive could take keepalive_lane

src/ucp/core/ucp_worker.c
dmitrygx: done

src/ucp/wireup/wireup_ep.c
dmitrygx: re-implemented to get lane/rsc_idx/uct_ep inside `ucp_ep_do_uct_ep_keepalive`

src/ucp/core/ucp_ep.c
dmitrygx: > can we add ucp_ep_config(ep)->is_am_keepalive and set it during ucp_ep_config_init?

we couldn't set `is_am_keepalive` during `ucp_ep_config_init`, since we should use ep_check in case of REMOTE_ID hasn't been resolved yet.

> do we still need to check UCT_IFACE_FLAG_CONNECT_TO_IFACE below?

yes, we could avoid checking it

dmitrygx:

src/ucp/core/ucp_ep.c
```diff
@@ -3253,24 +3254,78 @@ ucp_ep_is_am_keepalive(ucp_ep_h ep, ucp_rsc_index_t rsc_idx)
...
 }
 
-ucs_status_t ucp_ep_do_uct_ep_keepalive(ucp_ep_h ucp_ep, uct_ep_h uct_ep,
-                                        ucp_rsc_index_t rsc_idx, unsigned flags,
+static UCS_F_ALWAYS_INLINE uct_ep_h
+ucp_ep_select_keepalive(ucp_ep_h ucp_ep, ucp_rsc_index_t *rsc_idx_p,
+                        ucp_lane_index_t *lane_p)
+{
+    ucp_lane_index_t lane      = ucp_ep_config(ucp_ep)->key.keepalive_lane;
+    uct_ep_h uct_ep            = ucp_ep->uct_eps[lane];
```
brminich, line 3267: can init it under the `else` below, it is reinitialized for wireup

brminich: also can do:
```
if (wireup_ep == NULL) {
    ...
    goto out;
}

if (wireup_ep->flags & UCP_WIREUP_EP_FLAG_REMOTE_CONNECTED) {
...
```

dmitrygx: I thought about it, but then we should initialize wireup_ep as `ucp_wireup_ep(ucp_ep->uct_eps[ucp_ep_config(ucp_ep)->key.keepalive_lane])` which is pretty long and better to save `lane`/`uct_ep` and then reinitialize them

brminich (dismissed):

dmitrygx:
@yosefe could you review pls?

src/ucp/core/ucp_ep.c
```diff
@@ -3253,24 +3254,78 @@ ucp_ep_is_am_keepalive(ucp_ep_h ep, ucp_rsc_index_t rsc_idx)
...
+    ucp_wireup_ep_t *wireup_ep = ucp_wireup_ep(uct_ep);
+    ucp_rsc_index_t rsc_idx;
+
+    if (wireup_ep != NULL) {
```
yosefe, line 3271: i think it looks weird that we use this function twice, once from wireup ep and another time from do_keepalive, and then have a branch here to handle each case differently.
also not clear why need this refactoring, shouldn't we change only ucp_ep_is_am_keepalive() to fix the logic?

dmitrygx: I fixed the `ucp_ep_is_am_keepalive`, but I was asked to not pass `lane` which requires to move the uct_ep/rsc_index initialization logic to `ucp_ep_do_uct_ep_keepalive`

dmitrygx: https://github.com/openucx/ucx/pull/8113/commits/b26180772cec01e8699f4ad2291b1208046be2ce - original fix

dmitrygx:
failure is unrelated

src/ucp/wireup/wireup_ep.c
```diff
@@ -362,14 +366,15 @@ ucp_wireup_ep_do_check(ucp_ep_h ucp_ep, uct_ep_h uct_ep,
...
+    if (ucp_ep_is_am_keepalive(ucp_ep, rsc_idx, !is_p2p)) {
+        return ucp_ep_do_uct_ep_am_keepalive(ucp_ep, uct_ep, rsc_idx);
+    } else if (wiface->attr.cap.flags & UCT_IFACE_FLAG_EP_CHECK) {
+        return uct_ep_check(uct_ep, flags, comp);
     }
 
-    /* if EP_CHECK is not supported by UCT transport, it has to support a
-     * built-in keepalive mechanism to be able to detect peer failure during
-     * wireup
+    /* if EP_CHECK is not supported by UCT transport and AM-based keepalive is
```
brminich, line 375: plz start with upper case

dmitrygx: fixed

src/ucp/wireup/wireup_ep.c
```diff
@@ -362,14 +366,15 @@ ucp_wireup_ep_do_check(ucp_ep_h ucp_ep, uct_ep_h uct_ep,
     ucs_assert(rsc_idx != UCP_NULL_RESOURCE);
 
     wiface = ucp_worker_iface(worker, rsc_idx);
-    if (wiface->attr.cap.flags & UCT_IFACE_FLAG_EP_CHECK) {
-        return ucp_ep_do_uct_ep_keepalive(ucp_ep, uct_ep, rsc_idx, flags,
-                                          comp);
+    if (ucp_ep_is_am_keepalive(ucp_ep, rsc_idx, !is_p2p)) {
```
yosefe, line 369: why !is_p2p?

src/ucp/core/ucp_ep.c
```diff
@@ -3241,27 +3241,23 @@ void ucp_ep_invoke_err_cb(ucp_ep_h ep, ucs_status_t status)
...
 }
 
-static UCS_F_ALWAYS_INLINE int
-ucp_ep_is_am_keepalive(ucp_ep_h ep, ucp_rsc_index_t rsc_idx)
+int ucp_ep_is_am_keepalive(ucp_ep_h ep, ucp_rsc_index_t rsc_idx, int is_p2p)
 {
     ucp_worker_iface_t *wiface;
 
-    if (!(ep->flags & UCP_EP_FLAG_REMOTE_ID) ||
+    if (is_p2p || !(ep->flags & UCP_EP_FLAG_REMOTE_ID) ||
```
yosefe, line 3248: 1. rename rsc_idx ro rsc_indes
2. since we already pass rsc_index, i think is_p2p can be calculated from it: 
     `(rsc_index != NULL_RESOURCE) && ucp_ep_config_connect_p2p(worker, &ucp_ep_config(ep)->key, rsc_index)`

```c
int ucp_ep_is_am_keepalive(ucp_ep_h ep, ucp_rsc_index_t rsc_index)
{
    ucp_worker_h worker = ep->worker;

    return /* Not a CM lane */
           (rsc_index != NULL_RESOURCE) && 
           /* Have a remote endpoint ID to send in in the keepalive active message */
           (ep->flags & UCP_EP_FLAG_REMOTE_ID) &&
           /* Transport supports active messages */
           (ucp_worker_iface(worker, rsc_index)->flags & UCT_IFACE_FLAG_AM_BCOPY) &&
           /* Transport is not connected as point-to-point */
           !ucp_ep_config_connect_p2p(worker, &ucp_ep_config(ep)->key, rsc_index);
}
```

src/ucp/wireup/wireup_ep.c
dmitrygx: fixed

src/ucp/core/ucp_ep.c
dmitrygx: done

> since we already pass rsc_index, i think is_p2p can be calculated from it:
(rsc_index != NULL_RESOURCE) && ucp_ep_config_connect_p2p(worker, &ucp_ep_config(ep)->key, rsc_index)

we can't use such logic for AUX_EP

yosefe: > we can't use such logic for AUX_EP

because it could be created as to-iface inside wireup_ep and not set from existing lane?

src/ucp/core/ucp_worker.c
```diff
@@ -3147,6 +3147,7 @@ static int ucp_worker_do_ep_keepalive(ucp_worker_h worker, ucs_time_t now)
 {
     ucp_lane_index_t lane;
     ucp_rsc_index_t rsc_index;
+    int is_p2p;
```
yosefe, line 3150: maybe no need this helper var

dmitrygx: removed

src/ucp/core/ucp_ep.c
dmitrygx: >>we can't use such logic for AUX_EP

>because it could be created as to-iface inside wireup_ep and not set from existing lane?

it could be:
- created as to-IFACE inside WIREUP_EP if next_ep is CONNECT_TO_EP.
- set as to-IFACE/to-EP during connection inersecting during CM connection establishment.

src/ucp/core/ucp_worker.c
```diff
@@ -3177,7 +3175,8 @@ static int ucp_worker_do_ep_keepalive(ucp_worker_h worker, ucs_time_t now)
     ucs_trace("ep %p: do keepalive on lane[%d]=%p ep->flags=0x%x", ep, lane,
               ep->uct_eps[lane], ep->flags);
 
-    if (ucp_ep_is_am_keepalive(ep, rsc_index, is_p2p)) {
+    if (ucp_ep_is_am_keepalive(
+                ep, rsc_index, ucp_ep_config(ep)->p2p_lanes & UCS_BIT(lane))) {
```
yosefe, line 3179: this wrapping seems weird, is it from clang-format?

dmitrygx: fixed

yosefe (dismissed):

yosefe (dismissed):

src/ucp/core/ucp_ep.c
```diff
@@ -3241,27 +3241,21 @@ void ucp_ep_invoke_err_cb(ucp_ep_h ep, ucs_status_t status)
...
-        return 0;
-    }
-
-    wiface = ucp_worker_iface(ep->worker, rsc_idx);
-    return ucs_test_all_flags(wiface->attr.cap.flags,
-                              UCT_IFACE_FLAG_CONNECT_TO_IFACE |
-                              UCT_IFACE_FLAG_AM_BCOPY);
+    return /* Not a CM lane */
+            (rsc_index != UCP_NULL_RESOURCE) &&
+            /* Have a remote endpoint ID to send in in the keepalive active message */
```
brminich, line 3248: extra `in`

brminich (approved):

yosefe (approved):

PR #8118, opened 2022-04-12
UCT/IB/DC: Fix resending FC_PURE_GRANT in case of failure

dmitrygx:
## What

Fix resending `FC_PURE_GRANT` in case of failure of DCI.

## Why ?

Purging of pending queue of FC endpoints during error handling leads to removing `FC_PURE_GRANT` requests which were rescheduled to pending queue due to completing them with `WC_FLUSH_ERR` status.

## How ?

Don't release FC requests to mpool, need to collect them in a queue of FC pure grants and resend them after recovering DCI .

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.c
+++ b/src/uct/ib/dc/dc_mlx5.c
@@ -1647,13 +1647,11 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
     uct_ib_iface_t *ib_iface = &iface->super.super.super;
     ucs_status_t status;
     ucs_log_level_t log_lvl;
-    ucs_arbiter_t *waitq;
-    ucs_arbiter_group_t *group;
-    uint8_t pool_index;
 
-    uct_dc_mlx5_get_arbiter_params(iface, ep, &waitq, &group, &pool_index);
-    ucs_arbiter_group_purge(waitq, group,
-                            uct_dc_mlx5_ep_arbiter_purge_internal_cb, ep);
+    /* We don't purge an endpoint's pending queue, because only a FC endpoint
+     * could have internal TX operations scheduled there which shouldn't be
+     * purged - they need to be rescheduled when DCI will be recovered after an
+     * error */
 
     if (ep->flags & (UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED |
                      UCT_DC_MLX5_EP_FLAG_FLUSH_CANCEL)) {
--- a/src/uct/ib/dc/dc_mlx5_ep.c
+++ b/src/uct/ib/dc/dc_mlx5_ep.c
@@ -1310,9 +1310,10 @@ uct_dc_mlx5_iface_dci_do_common_pending_tx(uct_dc_mlx5_ep_t *ep,
      * arbiter group for which flush(CANCEL) was done */
     ucs_assert(!(ep->flags & UCT_DC_MLX5_EP_FLAG_FLUSH_CANCEL));
 
-    ucs_assertv(!uct_dc_mlx5_iface_dci_ep_can_send(ep),
-                "pending callback returned error, but send resources are"
-                " available");
+    ucs_assertv(!uct_dc_mlx5_iface_dci_ep_can_send(ep) ||
+                (ep == iface->tx.fc_ep),
+                "ep=%p: pending callback returned error, but send resources"
+                " are available and it is not fc_ep=%p", ep, iface->tx.fc_ep);
     return UCS_ARBITER_CB_RESULT_DESCHED_GROUP;
 }
 


dmitrygx:
failure is unrelated

brminich:
@hoopoepg, plz review as well

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1149,6 +1172,21 @@ ucs_status_t uct_dc_mlx5_iface_fc_grant(uct_pending_req_t *self)
...
+    ucs_status_t status;
+
+    status = uct_dc_mlx5_iface_fc_grant(&dc_req->super.super);
+    if (status == UCS_ERR_NO_RESOURCE){
+        uct_dc_mlx5_ep_do_pending_fc(ucs_derived_of(dc_req->super.ep,
+                                                    uct_dc_mlx5_ep_t),
+                                     dc_req);
+    } else {
+        ucs_assertv_always(status == UCS_OK, "Failed to send FC grant msg: %s",
+                           ucs_status_string(status));
```
hoopoepg, line 1186: assert_always will terminate process without any chance for error handling. is it ok?
@brminich what do you think? may be just report error and allow error handling to make it's work?

dmitrygx: we could just remove the assertion, but we don't have a chance to invoke error callabkac, since we don't know the endpoint on which we should invoke it

brminich: i think it is ok to assert on local failures, as we can consider it as a software error

hoopoepg (approved):

brminich:
what problem does it fix?

brminich:
is it possible to avoid using seperate queue, but do not purge pending queue for fc ep instead?

dmitrygx:
> is it possible to avoid using seperate queue, but do not purge pending queue for fc ep instead?

unfortunately, it is not possible.
I checked that we have the following assertion:
https://github.com/openucx/ucx/blob/e9a0f509a0bbf34d4dd815cf5ef37773f6bf5d27/src/uct/ib/dc/dc_mlx5_ep.c#L1313
which assumes that UCT pending request should be sent in case of there are resources on DCI.
but we return `UCS_ERR_NO_RESOURCE` in case of FC EP is failed (we own the DCI in this case, but it is under resetting)
https://github.com/openucx/ucx/blob/e9a0f509a0bbf34d4dd815cf5ef37773f6bf5d27/src/uct/ib/dc/dc_mlx5_ep.c#L961

I think we should keep the pending queue empty for FC EP too as we have for user's EPs.

brminich:
> > is it possible to avoid using seperate queue, but do not purge pending queue for fc ep instead?
> 
> unfortunately, it is not possible. I checked that we have the following assertion:
> 
> https://github.com/openucx/ucx/blob/e9a0f509a0bbf34d4dd815cf5ef37773f6bf5d27/src/uct/ib/dc/dc_mlx5_ep.c#L1313
> 
> 
> which assumes that UCT pending request should be sent in case of there are resources on DCI.
> but we return `UCS_ERR_NO_RESOURCE` in case of FC EP is failed (we own the DCI in this case, but it is under resetting)
> https://github.com/openucx/ucx/blob/e9a0f509a0bbf34d4dd815cf5ef37773f6bf5d27/src/uct/ib/dc/dc_mlx5_ep.c#L961
> 
> I think we should keep the pending queue empty for FC EP too as we have for user's EPs.

We can modify that assert to also check it is not an FC ep with failed DCI. 
Would the solution be shorter/simpler if we use just pending queue ?

brminich:
> > We can modify that assert to also check it is not an FC ep with failed DCI.
> > Would the solution be shorter/simpler if we use just pending queue ?
> 
> yes, we can modify assertion, but I think it will make FC EP inconsistent with other EPs

ok, but will it allow us to remove extra queue and simplify the change?

dmitrygx:
@yosefe @brminich unfortunately, we couldn't set `txwq->available` to `INT16_MIN` in case of DCI failure, because we won't be able to send tiny NO_OP packet in `flush(CANCEL)`

yosefe:
> @yosefe @brminich unfortunately, we couldn't set `txwq->available` to `INT16_MIN` in case of DCI failure, because we won't be able to send tiny NO_OP packet in `flush(CANCEL)`

so what's the alternative?

yosefe (dismissed):

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1647,13 +1647,11 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
...
-    ucs_arbiter_t *waitq;
-    ucs_arbiter_group_t *group;
-    uint8_t pool_index;
 
-    uct_dc_mlx5_get_arbiter_params(iface, ep, &waitq, &group, &pool_index);
-    ucs_arbiter_group_purge(waitq, group,
-                            uct_dc_mlx5_ep_arbiter_purge_internal_cb, ep);
+    /* We don't purge an endpoint's pending queue, because only a FC endpoint
+     * could have internal TX operations scheduled there which shouldn't be
+     * purged - they need to be rescheduled when DCI will be recover after an
```
brminich, line 1653: `will recover` or `will be recovered`

dmitrygx: fixed

yosefe (approved):

PR #8119, opened 2022-04-12
UCT/ROCM: increase max. number of hsa agents

edgargabriel:
## What

Increase the max. number of hsa agents to 63 (from 16).  Furthermore, minor reorganization of the uct_rocm_base_agents structure to reduce the number of cache misses.

## Why ?
The max. number of hsa agents includes both CPUs + GPUs and was too low for today's systems.

### DIFF
--- a/src/uct/rocm/base/rocm_base.c
+++ b/src/uct/rocm/base/rocm_base.c
@@ -14,12 +14,12 @@
 #include <pthread.h>
 
 
-#define MAX_AGENTS 16
+#define MAX_AGENTS 63
 static struct agents {
-    hsa_agent_t agents[MAX_AGENTS];
     int num;
-    hsa_agent_t gpu_agents[MAX_AGENTS];
+    hsa_agent_t agents[MAX_AGENTS];
     int num_gpu;
+    hsa_agent_t gpu_agents[MAX_AGENTS];
 } uct_rocm_base_agents;
 
 int uct_rocm_base_get_gpu_agents(hsa_agent_t **agents)


shamisp:
Edgar - can you switch this value from a compile time to runtime ? I guess
the optimal number of  hsa agents is dependent on the revison of hardware ?

edgargabriel:
@shamisp note, this is not the optimal number of hsa_agents, its the maximum number of hsa_agents that the rocm components can detect. The only reason we need to increase the value from 16 to something larger is because its the sum of CPUs+GPUs that a node can have.

shamisp:
@edgargabriel got it. I guess you can count number of GPUs (lspci) and number of CPUs (/proc/cpuinfo) and avoid hardcoded value. Just a thought.

edgargabriel:
@shamisp You are right. The function that we invoke returns the number of hsa_agents itself on the system (an assert failed because of that, which triggered this commit), and I can of course dynamically allocate the array. The original code didn't do that, but I can make the change.

edgargabriel:
On the other hand, I would have to worry about freeing this array to avoid leaking memory, I am not sure there is right now a function that we invoke at shutdown time.

shamisp:
@edgargabriel  it is up to you what works best for your use-case / hardware. If you feel like hardcoded value works better for now, I'm okay with this. This is just one of the things that will get broken again in a year or two :)

edgargabriel:
I would prefer to keep it as it is for now, can always revisit at a later stage. The value has been on purpose chosen such that it should be good for a few years (going from 16 to 63).

shamisp (approved):

PR #8120, opened 2022-04-13
UCP: Add API to query datatype attributes

rdietric:
## What
This PR adds a new public API function to query attributes for a UCP datatype. Currently, the packed size is the only attribute that can be queried.

## Why ?
The size of a message or data transfer is a performance-relevant information. Performance tools should be able to query this information. MPI provides `MPI_Type_size` to query the size of any MPI datatype.

## How ?
A follow-up PR will implement `ucp_dt_query` in `src/ucp/dt/dt.c`. Unfortunately, the size of a UCP generic datatype is not available before it has been used in a UCP communication operation without more significant API changes.

### DIFF
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -514,6 +514,17 @@ enum ucp_dt_type {
 
 
 /**
+ * @ingroup UCP_DATATYPE
+ * @brief UCP datatype attributes field mask.
+ *
+ * The enumeration allows querying a specific field in @ref ucp_datatype_attr_t.
+ */
+enum ucp_datatype_attr_field {
+    UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE = UCS_BIT(0) /**< packed datatype size */
+};
+
+
+/**
  * @ingroup UCP_MEM
  * @brief UCP memory mapping flags.
  *
@@ -958,6 +969,27 @@ typedef struct ucp_generic_dt_ops {
      */
     void (*finish)(void *state);
 } ucp_generic_dt_ops_t;
+
+
+/**
+ * @ingroup UCP_DATATYPE
+ * @brief UCP datatype attributes
+ *
+ * This structure provides attributes that can be queried for a UCP datatype.
+ */
+typedef struct ucp_datatype_attr {
+    /**
+     * Mask of valid fields in this structure, using bits from @ref
+     * ucp_datatype_attr_field. Fields not specified in this mask will be
+     * ignored. Provides ABI compatibility with respect to adding new fields.
+     */
+    uint64_t field_mask;
+
+    /**
+     * Packed size of a single element of the given datatype.
+     */
+    size_t   packed_size;
+} ucp_datatype_attr_t;
 
 
 /**
@@ -4496,6 +4528,29 @@ ucs_status_t ucp_dt_create_generic(const ucp_generic_dt_ops_t *ops, void *contex
  * @param [in]  datatype     Datatype object to destroy.
  */
 void ucp_dt_destroy(ucp_datatype_t datatype);
+
+
+/**
+ * @ingroup UCP_DATATYPE
+ * @brief Query attributes of a datatype.
+ *
+ * This routine fetches information about the attributes of a datatype.
+ * When @ref UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE is set in @a field_mask of @a attr,
+ * the field @a packed_size is set to the packed size (bytes) of the datatype.
+ *
+ * @note The size of a generic datatype is available only after it has been used
+ *       in a communication operation. If it is queried before,
+ *       @ref UCS_ERR_UNSUPPORTED will be returned.
+ *
+ * @note The size of IOV data cannot be determined from the UCP datatype. If an
+ *       IOV @a datatype is passed, @ref UCS_ERR_INVALID_PARAM will be returned.
+ *
+ * @param [in]  datatype     Datatype object to query.
+ * @param [out] attr         Filled with attributes of the datatype.
+ *
+ * @return Error code as defined by @ref ucs_status_t
+ */
+ucs_status_t ucp_dt_query(ucp_datatype_t datatype, ucp_datatype_attr_t *attr);
 
 
 /**


shamisp:
Hi @rdietric. Thanks for sharing your API proposal. Do you have signed CLA with the project ? Corporate and private CLAs can be found here: https://openucx.org/license/

Guidance to contributors is here: https://github.com/openucx/ucx/wiki/Guidance-for-contributors

yosefe:
@shamisp , @rdietric is with NVIDIA.

shamisp:
> @shamisp , @rdietric is with NVIDIA.

I just noticed the email signature. Typically I check the GitHub account affiliation first, which is not set in this case.

shamisp:
> I used my existing Github account and just added NVIDIA to my profile. Sorry for the inconveniences!

@rdietric not a problem, you did better than many others ;)

src/ucp/api/ucp.h
```diff
@@ -4497,6 +4527,17 @@ ucs_status_t ucp_dt_create_generic(const ucp_generic_dt_ops_t *ops, void *contex
  */
 void ucp_dt_destroy(ucp_datatype_t datatype);
 
+/**
+ * @ingroup UCP_DATATYPE
+ * @brief Query attributes of a datatype.
+ *
+ * When UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE is set in field_mask of @a attr, the
```
brminich, line 4534: please remove extra `the`

src/ucp/api/ucp.h
```diff
@@ -4497,6 +4527,17 @@ ucs_status_t ucp_dt_create_generic(const ucp_generic_dt_ops_t *ops, void *contex
...
+ * the field packed_size is set to the packed size (bytes) of the datatype.
+ * If the size could not be determined, 0 is returned.
+ * @note The size of a generic datatype is not available before it has been used
```
brminich, line 4537: also what value should we return for iov?

rdietric: That's right, the API does not work for IOV data and since there is no public API that creates an IOV datatype, we cannot know the size of the vector. Thus, it falls into the category "size could not be determined", but it should be noted in the description separately.

src/ucp/api/ucp.h
rdietric: Oops, will remove it.

src/ucp/api/ucp.h
```diff
@@ -513,6 +513,17 @@ enum ucp_dt_type {
...
 
 
+/**
+ * @ingroup UCP_DATATYPE
+ * @brief UCP datatype attributes field mask.
+ *
+ * The enumeration allows querying a specific field in @ref ucp_datatype_attr_t.
+ */
+enum ucp_dt_attr_field {
+    UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE = UCS_BIT(0) /**< packed datatype size */
```
brminich, line 523: maybe rename it to `PACKED_CONTIG_SIZE`, as we can't use this API for anything other than contig types?

brminich:
@yosefe, @shamisp can you please review?

shamisp (changes requested):

shamisp:
@tonycurtis can you please review documentation.

tonycurtis (approved):

src/ucp/api/ucp.h
shamisp: Depends. If there is no plan to support this query for any other datatypes, probably we can rename it. If there is a longer term plans to support this type of query for other datatypes we shell use generic name.

src/ucp/api/ucp.h
```diff
@@ -4497,6 +4527,21 @@ ucs_status_t ucp_dt_create_generic(const ucp_generic_dt_ops_t *ops, void *contex
  */
 void ucp_dt_destroy(ucp_datatype_t datatype);
 
+/**
+ * @ingroup UCP_DATATYPE
+ * @brief Query attributes of a datatype.
+ *
+ * When @ref UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE is set in @a field_mask of @a attr,
```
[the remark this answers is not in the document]
rdietric: Right, there is only the brief description. I'll add the description you are suggesting.

src/ucp/api/ucp.h
rdietric: So it might be safer to keep it as is. `enum ucp_dt_type` lists _contig_, _strided_, _iov_ and _generic_. iov is not supported and this seems to make sense to me with regard to the UCP API. It looks like _strided_ is not really used, yet. _contig_ and _generic_ are supported.

shamisp: I'm a bit concerned that 0 is used as an "unsupported". Imho, using -1 makes more sense.

rdietric: I expected these concerns, but for simplicity's sake I chose `0` for unsupported, which fits well with `size_t`. And a datatype with size 0 does not make sense, right? If we want `-1` for unsupported, we should also add a `#define UCP_DATATYPE_UNSUPPORTED ((size_t)-1)` or `#define UCP_DATATYPE_SIZE_UNKNOWN ((size_t)-1)`. So I am also okay with changing unsupported from `0` to `-1`.

shamisp: Looking at MPI API as a reference https://www.open-mpi.org/doc/v4.0/man3/MPI_Type_size.3.php
If MPI cannot figure out what to do, it returns MPI_UNDEFINED in size which is mapped to https://github.com/open-mpi/ompi/blob/eb87378cd8c55c08fde666c03a04aa2811da395a/ompi/include/mpi.h.in#L545

I think MPI solution is not elegant either :)

Looking at our other "query" APIs, I think the function should return `ucs_status_t`. It is consistent approach across our query API and provides clean path to return errors.

shamisp: It is between `UCS_ERR_INVALID_PARAM` and `UCS_ERR_UNSUPPORTED`.

rdietric: I leave the choice to you. What about using `UCS_ERR_INVALID_PARAM` in case an IOV datatype (created by `ucp_dt_make_iov()`) is passed and `UCS_ERR_UNSUPPORTED` for the other case where the size of the generic datatype is not available yet? Or should we add something like `UCS_ERR_NOT_AVAILABLE` for the latter?

For my use case I would only check for `UCS_OK`.

shamisp: > I leave the choice to you. What about using `UCS_ERR_INVALID_PARAM` in case an IOV datatype (created by `ucp_dt_make_iov()`) is passed and `UCS_ERR_UNSUPPORTED` for the other case where the size of the generic datatype is not available yet? Or should we add something like `UCS_ERR_NOT_AVAILABLE` for the latter?
> 
> For my use case I would only check for `UCS_OK`.

I agree with above proposal.

rdietric: I added `ucs_status_t` as return type and changed the description accordingly.

src/ucp/api/ucp.h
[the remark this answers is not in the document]
rdietric: I added the description.

shamisp:
Looks good

shamisp (dismissed):

rdietric:
I added PR #8150 with the implementation.

rdietric:
Is there anything else I can do to get the PR merged?

src/ucp/api/ucp.h
```diff
@@ -4497,6 +4527,27 @@ ucs_status_t ucp_dt_create_generic(const ucp_generic_dt_ops_t *ops, void *contex
...
+ *
+ * @note The size of IOV data cannot be determined from the UCP datatype. If an
+ *       IOV @a datatype is passed, @ref UCS_ERR_INVALID_PARAM will be returned.
+ *
+ * @param [in]  datatype     Datatype object to query.
+ * @param [out] attr         Filled with attributes of the datatype.
+ *
+ * @return Error code as defined by @ref ucs_status_t
+ */
+ucs_status_t ucp_dt_query(ucp_datatype_t datatype, ucp_datatype_attr_t *attr);
```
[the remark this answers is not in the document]
rdietric: Fixed.

src/ucp/api/ucp.h
```diff
@@ -4497,6 +4527,27 @@ ucs_status_t ucp_dt_create_generic(const ucp_generic_dt_ops_t *ops, void *contex
  */
 void ucp_dt_destroy(ucp_datatype_t datatype);
 
+/**
```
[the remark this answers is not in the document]
rdietric: Fixed.

src/ucp/api/ucp.h
```diff
@@ -513,6 +513,17 @@ enum ucp_dt_type {
 };
 
 
+/**
+ * @ingroup UCP_DATATYPE
+ * @brief UCP datatype attributes field mask.
+ *
+ * The enumeration allows querying a specific field in @ref ucp_datatype_attr_t.
+ */
+enum ucp_dt_attr_field {
```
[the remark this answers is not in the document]
rdietric: Changed.

src/ucp/api/ucp.h
```diff
@@ -960,6 +971,25 @@ typedef struct ucp_generic_dt_ops {
...
+typedef struct {
+    /**
+     * Mask of valid fields in this structure, using bits from @ref ucp_dt_attr_field.
+     * Fields not specified in this mask will be ignored.
+     * Provides ABI compatibility with respect to adding new fields.
+     */
+    uint64_t field_mask;
+
+     /* Packed size of a single element of the given datatype. */
+     size_t packed_size;
```
yosefe, line 989: wrong alignment

yosefe (approved):

yosefe:
@rdietric pls check the build failures

rdietric:
> @rdietric pls check the build failures

I fixed the build failures, which were caused by a wrong @ref in a field comment.

src/ucp/api/ucp.h
rdietric: Fixed. (Unfortunately, the clang-format from UCX's clang fork does not work on my system. It complains about many lines in _.clang-format_.)

yosefe (approved):

shamisp (dismissed):

yosefe (dismissed):

yosefe:
@tonycurtis can you pls take a look?

tonycurtis (dismissed):

rdietric:
I squashed the commits. The failing tests seem unrelated to changes in this PR.

PR #8121, opened 2022-04-13
UCT/MEM: fixed potential crash on MD mem alloc

hoopoepg:
- in case if in uct_iface_mp_chunk_alloc function used non-MD method
  without registration support it crashed
- fix:  force use MD allocation method in such cases

### DIFF
--- a/src/uct/base/uct_mem.c
+++ b/src/uct/base/uct_mem.c
@@ -334,14 +334,52 @@ ucs_status_t uct_mem_free(const uct_allocated_memory_t *mem)
     }
 }
 
+static int uct_iface_is_allowed_alloc_method(uct_base_iface_t *iface,
+                                             uct_alloc_method_t method)
+{
+    unsigned i;
+
+    for (i = 0; i < iface->config.num_alloc_methods; i++) {
+        if (iface->config.alloc_methods[i] == method) {
+            return 1;
+        }
+    }
+
+    return 0;
+}
+
 ucs_status_t uct_iface_mem_alloc(uct_iface_h tl_iface, size_t length, unsigned flags,
                                  const char *name, uct_allocated_memory_t *mem)
 {
-    uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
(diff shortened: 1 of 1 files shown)


src/uct/base/uct_mem.c
```diff
@@ -334,14 +334,54 @@ ucs_status_t uct_mem_free(const uct_allocated_memory_t *mem)
...
+    if (!(md_attr.cap.flags & UCT_MD_FLAG_REG) &&
+        uct_iface_is_allowed_alloc_method(iface, UCT_ALLOC_METHOD_MD)) {
+        /* If MD does not support registration, allow only the MD method */
+        alloc_methods     = &method_md;
+        num_alloc_methods = 1;
+    } else if (!(md_attr.cap.flags & UCT_MD_FLAG_REG)) {
+        /* If MD does not support registration and MD method is not in list
+         * return error */
+        ucs_assert(!uct_iface_is_allowed_alloc_method(iface,
+                                                      UCT_ALLOC_METHOD_MD));
```
brminich, line 378: imo this is excessive as it is guaranteed by the if conditions

src/uct/base/uct_mem.c
```diff
@@ -355,21 +395,15 @@ ucs_status_t uct_iface_mem_alloc(uct_iface_h tl_iface, size_t length, unsigned f
...
     /* If the memory was not allocated using MD, register it */
     if (mem->method != UCT_ALLOC_METHOD_MD) {
-
-        status = uct_md_query(iface->md, &md_attr);
-        if (status != UCS_OK) {
-            goto err_free;
-        }
-
-        /* If MD does not support registration, allow only the MD method */
+        /* If method is not MD - register memory */
```
brminich, line 406: it repeats the comment on the line 404

src/uct/base/uct_mem.c
hoopoepg: ok, removed

brminich (dismissed):

yosefe (approved):

PR #8122, opened 2022-04-13
UCT/IB/DC/RC: Add common update TX function for RC/DC

dmitrygx:
## What

Add common update TX function for RC/DC.

## Why ?

To have a common code to update TX resources.

## How ?

1. Introduce `uct_rc_mlx5_common_iface_txwq_update_tx_res` function which includes coe from `uct_rc_mlx5_iface_update_tx_res`.
2. Remove `uct_dc_mlx5_update_tx_res` and use `uct_rc_mlx5_common_iface_txwq_update_tx_res` instead of it.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.c
+++ b/src/uct/ib/dc/dc_mlx5.c
@@ -252,7 +252,8 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface)
                    iface, dci_index, txqp, hw_ci);
 
     uct_rc_mlx5_txqp_process_tx_cqe(txqp, cqe, hw_ci);
-    uct_dc_mlx5_update_tx_res(iface, txwq, txqp, hw_ci);
+    uct_rc_mlx5_common_iface_txwq_update_tx_res(&iface->super.super, txwq,
+                                                txqp, hw_ci);
 
     /**
      * Note: DCI is released after handling completion callbacks,
--- a/src/uct/ib/dc/dc_mlx5_ep.c
+++ b/src/uct/ib/dc/dc_mlx5_ep.c
@@ -1589,7 +1589,8 @@ void uct_dc_mlx5_ep_handle_failure(uct_dc_mlx5_ep_t *ep, void *arg,
 
     ucs_assert(!uct_dc_mlx5_iface_is_dci_rand(iface));
 
-    uct_dc_mlx5_update_tx_res(iface, txwq, txqp, pi);
+    uct_rc_mlx5_common_iface_txwq_update_tx_res(&iface->super.super, txwq,
+                                                txqp, pi);
     uct_rc_txqp_purge_outstanding(&iface->super.super, txqp, ep_status, pi, 0);
 
     /* Invoke a user's error callback and release TX/FC resources before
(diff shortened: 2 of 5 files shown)


dmitrygx:

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1589,7 +1589,8 @@ void uct_dc_mlx5_ep_handle_failure(uct_dc_mlx5_ep_t *ep, void *arg,
 
     ucs_assert(!uct_dc_mlx5_iface_is_dci_rand(iface));
 
-    uct_dc_mlx5_update_tx_res(iface, txwq, txqp, pi);
+    uct_rc_mlx5_common_iface_txwq_update_tx_res(&iface->super.super, txwq,
```
brminich, line 1592: func name is too long

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -252,7 +252,8 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface)
                    iface, dci_index, txqp, hw_ci);
 
     uct_rc_mlx5_txqp_process_tx_cqe(txqp, cqe, hw_ci);
-    uct_dc_mlx5_update_tx_res(iface, txwq, txqp, hw_ci);
+    uct_rc_mlx5_common_iface_txwq_update_tx_res(&iface->super.super, txwq,
```
yosefe, line 255: uct_rc_mlx5_common_update_tx_res

src/uct/ib/rc/accel/rc_mlx5.inl
```diff
@@ -919,6 +919,30 @@ static UCS_F_ALWAYS_INLINE void uct_rc_mlx5_common_txqp_bcopy_post(
...
 }
 
+static UCS_F_ALWAYS_INLINE uint16_t
+uct_rc_mlx5_common_iface_txwq_update_tx_res(uct_rc_iface_t *rc_iface,
+                                            uct_ib_mlx5_txwq_t *txwq,
+                                            uct_rc_txqp_t *txqp,
+                                            uint16_t hw_ci)
+{
+    uint16_t bb_num = uct_ib_mlx5_txwq_update_bb(txwq, hw_ci) -
+                      uct_rc_txqp_available(txqp);
```
yosefe, line 929: seems like there is extra math for dc case now

src/uct/ib/rc/accel/rc_mlx5_iface.c
```diff
@@ -115,24 +115,10 @@ static UCS_F_ALWAYS_INLINE void
...
-     * and all the rest report the amount of BBs the previous WQE has consumed.
-     */
-    ucs_assertv(bb_num > 0, "hw_ci=%d prev_sw_pi=%d available=%d bb_num=%d",
-                hw_ci, txwq->prev_sw_pi, txqp->available, bb_num);
-
-    uct_rc_txqp_available_add(txqp, bb_num);
-    ucs_assert(uct_rc_txqp_available(txqp) <= txwq->bb_max);
-
-    uct_rc_iface_update_reads(rc_iface);
+    uint16_t bb_num =
```
yosefe, line 118: don't init inline - this function has side effects

src/uct/ib/rc/accel/rc_mlx5.inl
dmitrygx: yes, then it means that they can't be moved to common code

PR #8123, opened 2022-04-13
UCT/IB/DC: Code simplifications

dmitrygx:
## What

DC code improvements.

## Why ?

To reduce code duplications and decrease amount of code.

## How ?

1. Remove separate `if` block for randmo DCI policy to do arbiter group purge in `uct_dc_mlx5_ep_pending_purge`.
2. Remvoe branch `uct_dc_mlx5_iface_is_dci_rand` from `uct_dc_mlx5_iface_dci_detach`.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5_ep.c
+++ b/src/uct/ib/dc/dc_mlx5_ep.c
@@ -1376,7 +1376,6 @@ uct_dc_mlx5_iface_dci_do_dcs_pending_tx(ucs_arbiter_t *arbiter,
                                                 uct_dc_mlx5_iface_t);
     int is_only                = ucs_arbiter_elem_is_only(elem);
     ucs_arbiter_cb_result_t res;
-    uint8_t dci;
 
     res = uct_dc_mlx5_iface_dci_do_common_pending_tx(ep, elem);
     if ((res != UCS_ARBITER_CB_RESULT_REMOVE_ELEM) || !is_only) {
@@ -1387,12 +1386,7 @@ uct_dc_mlx5_iface_dci_do_dcs_pending_tx(ucs_arbiter_t *arbiter,
      * and the dci has no outstanding operations. For example pending
      * callback did not send anything. (uct_ep_flush or just return ok)
      */
-    dci = ep->dci;
-    if (!uct_dc_mlx5_iface_dci_detach(iface, ep)) {
-        return res;
-    }
-
-    uct_dc_mlx5_iface_dci_schedule_release(iface, dci);
+    uct_dc_mlx5_iface_dci_detach(iface, ep);
     return res;
 }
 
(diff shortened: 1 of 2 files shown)


src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1481,30 +1481,20 @@ void uct_dc_mlx5_ep_pending_purge(uct_ep_h tl_ep,
 {
     uct_dc_mlx5_iface_t *iface = ucs_derived_of(tl_ep->iface, uct_dc_mlx5_iface_t);
     uct_dc_mlx5_ep_t *ep       = ucs_derived_of(tl_ep, uct_dc_mlx5_ep_t);
+    uint8_t dci                = ep->dci;
```
yosefe, line 1484: maybe remove this var?

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1481,30 +1481,20 @@ void uct_dc_mlx5_ep_pending_purge(uct_ep_h tl_ep,
...
     ucs_arbiter_group_purge(waitq, group, uct_dc_mlx5_ep_arbiter_purge_cb,
                             &args);
 
-    if (ep->dci == UCT_DC_MLX5_EP_NO_DCI) {
-        return;
-    }
-
-    dci = ep->dci;
-    if (!uct_dc_mlx5_iface_dci_detach(iface, ep)) {
+    if ((dci == UCT_DC_MLX5_EP_NO_DCI) ||
```
yosefe, line 1495: maybe
if (dci != NO_DCI && !is_rand && detach) { schedule_release }
and similar refactor (+remove dci helper var) in uct_dc_mlx5_iface_dci_do_dcs_pending_tx
BTW maybe we can move schedule_release inside detach ?

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1481,30 +1481,20 @@ void uct_dc_mlx5_ep_pending_purge(uct_ep_h tl_ep,
...
     ucs_arbiter_group_t *group;
     uint8_t pool_index;
-    uint8_t dci;
-
-    if (uct_dc_mlx5_iface_is_dci_rand(iface)) {
-        ucs_arbiter_group_purge(uct_dc_mlx5_iface_tx_waitq(iface),
-                                uct_dc_mlx5_ep_rand_arb_group(iface, ep),
-                                uct_dc_mlx5_ep_arbiter_purge_cb, &args);
-        return;
-    }
```
brminich, line 1496: in this case we may deschedule group for rand scheme on line 1511, which is not allowed

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: we do `uct_dc_mlx5_iface_dci_detach`, so we should save `dci`

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: we have a check for `uct_dc_mlx5_iface_is_dci_rand` below to return from the function

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: good idea, done

yosefe (dismissed):

brminich (approved):

yosefe (approved):

dmitrygx:
bot:pipe:retest

PR #8125, opened 2022-04-13
UCT/IB/DC: Remove assertion for sending FC grant

dmitrygx:
## What

Remove assertion which expects that sending FC grant will be completed with either `UCS_ERR_NO_RESOURCE` or `UCS_OK`.

## Why ?

This assertion could fail in case of `ibv_create_ah` fails during discovering some errors.

## How ?

1. Implement `uct_dc_mlx5_iface_fc_pure_grant_str` function to use for printing FC_PURE_GRANT request where it is needed.
2. Replace `ucs_assert_always` by `ucs_diag` for checking the result of `uct_dc_mlx5_iface_fc_grant`.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.c
+++ b/src/uct/ib/dc/dc_mlx5.c
@@ -1149,9 +1149,25 @@ ucs_status_t uct_dc_mlx5_iface_fc_grant(uct_pending_req_t *self)
     return status;
 }
 
-ucs_status_t uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
-                                          uct_rc_hdr_t *hdr, unsigned length,
-                                          uint32_t imm_data, uint16_t lid, unsigned flags)
+const char *
+uct_dc_mlx5_fc_req_str(uct_dc_fc_request_t *fc_req, char *buf, size_t max)
+{
+    char gid_str[32];
+
+    ucs_snprintf_zero(buf, max,
+                      "FC_PURE_GRANT seq %" PRIu64 " dct_num 0x%x"
+                      " lid %d gid %s",
+                      fc_req->sender.payload.seq, fc_req->dct_num, fc_req->lid,
+                      uct_ib_gid_str(ucs_unaligned_ptr(
+                                             &fc_req->sender.payload.gid),
+                                     gid_str, sizeof(gid_str)));
+    return buf;
+}
+
+static ucs_status_t
+uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
+                             uct_rc_hdr_t *hdr, unsigned length,
+                             uint32_t imm_data, uint16_t lid, unsigned flags)
 {
     uct_dc_mlx5_iface_t *iface = ucs_derived_of(rc_iface, uct_dc_mlx5_iface_t);
     uint8_t fc_hdr             = uct_rc_fc_get_fc_hdr(hdr->am_id);
@@ -1164,6 +1180,7 @@ ucs_status_t uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_
     ucs_arbiter_t *waitq;
     ucs_arbiter_group_t *group;
     uint8_t pool_index;
+    char buf[128];
 
     ucs_assert(rc_iface->config.fc_enabled);
 
@@ -1173,7 +1190,7 @@ ucs_status_t uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_
 
         dc_req = ucs_mpool_get(&iface->super.super.tx.pending_mp);
         if (ucs_unlikely(dc_req == NULL)) {
-            ucs_error("Failed to allocate FC request");
+            ucs_error("fc_ep=%p: failed to allocate FC request", ep);
             return UCS_ERR_NO_MEMORY;
         }
 
@@ -1186,10 +1203,10 @@ ucs_status_t uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_
         status = uct_dc_mlx5_iface_fc_grant(&dc_req->super.super);
         if (status == UCS_ERR_NO_RESOURCE){
             uct_dc_mlx5_ep_do_pending_fc(ep, dc_req);
-        } else {
-            ucs_assertv_always(status == UCS_OK,
-                               "Failed to send FC grant msg: %s",
-                               ucs_status_string(status));
+        } else if (status != UCS_OK) {
+            ucs_diag("fc_ep %p: failed to send %s: %s", ep,
+                     uct_dc_mlx5_fc_req_str(dc_req, buf, sizeof(buf)),
+                     ucs_status_string(status));
         }
     } else if (fc_hdr == UCT_RC_EP_FC_PURE_GRANT) {
         sender = (uct_dc_fc_sender_data_t *)(hdr + 1);
--- a/src/uct/ib/dc/dc_mlx5_ep.c
+++ b/src/uct/ib/dc/dc_mlx5_ep.c
@@ -1040,19 +1040,15 @@ uct_dc_mlx5_ep_fc_pure_grant_send_completion(uct_rc_iface_send_op_t *send_op,
     uct_dc_fc_request_t *fc_req = (uct_dc_fc_request_t*)send_op->buffer;
     uct_dc_mlx5_ep_t *fc_ep     = ucs_derived_of(fc_req->super.ep,
                                                  uct_dc_mlx5_ep_t);
-    char gid_str[32];
+    char buf[128];
 
     if (ucs_likely(!(send_op->flags & UCT_RC_IFACE_SEND_OP_STATUS) ||
                    (send_op->status != UCS_ERR_CANCELED))) {
         /* Pure grant sent - release it */
         ucs_mpool_put(fc_req);
     } else {
-        ucs_trace("fc_ep %p: re-sending FC_PURE_GRANT (seq:%" PRIu64 ")"
-                  " to dct_num:0x%x, lid:%d, gid:%s",
-                  fc_ep,  fc_req->sender.payload.seq, fc_req->dct_num,
-                  fc_req->lid,
-                  uct_ib_gid_str(ucs_unaligned_ptr(&fc_req->sender.payload.gid),
-                                 gid_str, sizeof(gid_str)));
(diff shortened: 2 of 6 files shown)


dmitrygx:

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1186,10 +1202,12 @@ ucs_status_t uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_
...
             uct_dc_mlx5_ep_do_pending_fc(ep, dc_req);
-        } else {
-            ucs_assertv_always(status == UCS_OK,
-                               "Failed to send FC grant msg: %s",
-                               ucs_status_string(status));
+            status = UCS_OK;
+        } else if (status != UCS_OK) {
+            ucs_diag("fc_ep %p: failed to send %s: %s", ep,
+                     uct_dc_mlx5_iface_fc_pure_grant_str(
+                             dc_req, buf, ucs_static_array_size(buf)),
```
yosefe, line 1209: sizeof(buf) for consistency with other code

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1149,21 +1149,37 @@ ucs_status_t uct_dc_mlx5_iface_fc_grant(uct_pending_req_t *self)
...
+    return buf;
+}
+
 ucs_status_t uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
                                           uct_rc_hdr_t *hdr, unsigned length,
                                           uint32_t imm_data, uint16_t lid, unsigned flags)
 {
     uct_dc_mlx5_iface_t *iface = ucs_derived_of(rc_iface, uct_dc_mlx5_iface_t);
     uint8_t fc_hdr             = uct_rc_fc_get_fc_hdr(hdr->am_id);
+    ucs_status_t status        = UCS_OK;
```
yosefe, line 1173: better avoid initializing status

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1040,19 +1040,16 @@ uct_dc_mlx5_ep_fc_pure_grant_send_completion(uct_rc_iface_send_op_t *send_op,
...
     } else {
-        ucs_trace("fc_ep %p: re-sending FC_PURE_GRANT (seq:%" PRIu64 ")"
-                  " to dct_num:0x%x, lid:%d, gid:%s",
-                  fc_ep,  fc_req->sender.payload.seq, fc_req->dct_num,
-                  fc_req->lid,
-                  uct_ib_gid_str(ucs_unaligned_ptr(&fc_req->sender.payload.gid),
-                                 gid_str, sizeof(gid_str)));
+        ucs_trace("fc_ep %p: re-sending %s", fc_ep,
+                  uct_dc_mlx5_iface_fc_pure_grant_str(
+                          fc_req, buf, ucs_static_array_size(buf)));
```
yosefe, line 1052: sizeof(buf) for consistency with other code

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1186,10 +1202,12 @@ ucs_status_t uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_
...
         if (status == UCS_ERR_NO_RESOURCE){
             uct_dc_mlx5_ep_do_pending_fc(ep, dc_req);
-        } else {
-            ucs_assertv_always(status == UCS_OK,
-                               "Failed to send FC grant msg: %s",
-                               ucs_status_string(status));
+            status = UCS_OK;
+        } else if (status != UCS_OK) {
+            ucs_diag("fc_ep %p: failed to send %s: %s", ep,
+                     uct_dc_mlx5_iface_fc_pure_grant_str(
```
yosefe, line 1208: uct_dc_mlx5_fc_req_str

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1149,21 +1149,37 @@ ucs_status_t uct_dc_mlx5_iface_fc_grant(uct_pending_req_t *self)
     return status;
 }
 
+const char* uct_dc_mlx5_iface_fc_pure_grant_str(uct_dc_fc_request_t *fc_req,
+                                                char *buf, size_t max)
+{
+    char gid_str[32];
+
+    ucs_snprintf_zero(buf, max, "FC_PURE_GRANT (seq:%" PRIu64 ") to"
```
yosefe, line 1157: `"FC_PURE_GRANT seq %" PRIu64 " dct_num 0x%x lid %d gid %s"`

dmitrygx: done

src/uct/ib/dc/dc_mlx5.c
dmitrygx: fixed

src/uct/ib/dc/dc_mlx5.c
dmitrygx: done

src/uct/ib/dc/dc_mlx5.c
dmitrygx: done

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: done

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1149,13 +1149,13 @@ ucs_status_t uct_dc_mlx5_iface_fc_grant(uct_pending_req_t *self)
...
-                                                char *buf, size_t max)
+const char* uct_dc_mlx5_fc_req_str(uct_dc_fc_request_t *fc_req, char *buf,
+                                   size_t max)
 {
     char gid_str[32];
 
-    ucs_snprintf_zero(buf, max, "FC_PURE_GRANT (seq:%" PRIu64 ") to"
-                      " dct_num:0x%x, lid:%d, gid:%s",
+    ucs_snprintf_zero(buf, max, "FC_PURE_GRANT seq %" PRIu64 " dct_num 0x%x"
+                      " lid %d  gid %s",
```
yosefe, line 1158: one spaceline before "gid"

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1202,12 +1202,11 @@ ucs_status_t uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_
...
         if (status == UCS_ERR_NO_RESOURCE){
             uct_dc_mlx5_ep_do_pending_fc(ep, dc_req);
-            status = UCS_OK;
         } else if (status != UCS_OK) {
             ucs_diag("fc_ep %p: failed to send %s: %s", ep,
-                     uct_dc_mlx5_iface_fc_pure_grant_str(
-                             dc_req, buf, ucs_static_array_size(buf)),
+                     uct_dc_mlx5_fc_req_str(dc_req, buf, sizeof(buf)),
                      ucs_status_string(status));
+            return status;
```
yosefe, line 1209: when this function returns error instead of OK, seems it's causing this:
https://github.com/openucx/ucx/blob/ab13538811ffe55eab4cf7058c50ded1fe1c596b/src/uct/ib/rc/accel/rc_mlx5.inl#L120
is this really what needs to happen?
maybe make fc_handler return void?

dmitrygx: nice catch, didn't notice it.
shall we fix RC and DC handlers to always return UCS_OK, except the case when RC handles AM inside it

yosefe: yes

yosefe: and maybe add comment inside the handlers to not miss it

dmitrygx: > and maybe add comment inside the handlers to not miss it

yes, makes sense it

src/uct/ib/rc/base/rc_iface.h
```diff
@@ -198,13 +198,13 @@ typedef void (*uct_rc_iface_cleanup_rx_func_t)(uct_rc_iface_t *iface);
...
 
-typedef ucs_status_t (*uct_rc_iface_fc_handler_func_t)(uct_rc_iface_t *iface,
-                                                       unsigned qp_num,
-                                                       uct_rc_hdr_t *hdr,
-                                                       unsigned length,
-                                                       uint32_t imm_data,
-                                                       uint16_t lid,
-                                                       unsigned flags);
+/* Returns 1 if AM data packed along with FC information, otherwise - 0 */
+typedef int (*uct_rc_iface_fc_handler_func_t)(uct_rc_iface_t *iface,
```
yosefe, line 202: IMO better keep return status as ucs_status_t to align with AM handlers

dmitrygx: but then it will be always `UCS_OK`, since I moved AM handling from the function

yosefe: if it always returns OK, why can't it be void?

dmitrygx: > if it always returns OK, why can't it be void?

we return whether we should handle AM or not.
but we could decide it after handling FC - see fc63c25

src/uct/ib/rc/accel/rc_mlx5.inl
```diff
@@ -413,15 +413,22 @@ uct_rc_mlx5_iface_common_am_handler(uct_rc_mlx5_iface_common_t *iface,
...
-                                    cqe->imm_inval_pkey, cqe->slid, flags);
-    } else {
-        status = uct_iface_invoke_am(&iface->super.super.super, hdr->rc_hdr.am_id,
-                                     hdr + 1, byte_len - sizeof(*hdr),
-                                     flags);
+        rc_ops->fc_handler(&iface->super, qp_num, &hdr->rc_hdr,
+                           byte_len - sizeof(*hdr), cqe->imm_inval_pkey,
+                           cqe->slid, flags);
+        if (uct_rc_fc_get_fc_hdr(hdr->rc_hdr.am_id) ==
+                    UCT_RC_EP_FC_PURE_GRANT) {
```
yosefe, line 420: the reason to put it inside fc_handler was to reduce code footprint of the fast-path polling function

yosefe:
LGTM, but pls check code style issues

brminich (approved):

dmitrygx:
> LGTM, but pls check code style issues

fixed

yosefe (approved):

brminich (approved):

PR #8127, opened 2022-04-15
UCP: count pending connection request

changchengx:
Signed-off-by: Changcheng Liu <jerrliu@nvidia.com>

## What
count the pending connection requests in ucp_listener

## Why ?
audit the connection requests that are dealed by the user application i.e. "After parsing the connection request parameters, it doesn't establish connection or reject the connection request"

## How ?
Add ucp_listener::conn_reqs member filed and audit the number.

### DIFF
--- a/src/ucp/core/ucp_listener.c
+++ b/src/ucp/core/ucp_listener.c
@@ -368,6 +368,12 @@ void ucp_listener_destroy(ucp_listener_h listener)
                             listener);
     UCS_ASYNC_UNBLOCK(&listener->worker->async);
 
+    if (listener->conn_reqs != 0) {
+        ucs_warn("destroying listener %p with "
+                 "%d unprocessed connection requests",
+                 listener, listener->conn_reqs);
+    }
+
     ucp_listener_free_uct_listeners(listener);
     ucs_free(listener);
 }
@@ -382,6 +388,7 @@ ucs_status_t ucp_listener_reject(ucp_listener_h listener,
     UCS_ASYNC_BLOCK(&worker->async);
     uct_listener_reject(conn_request->uct_listener, conn_request->uct_req);
     ucs_free(conn_request->remote_dev_addr);
+    listener->conn_reqs--;
     UCS_ASYNC_UNBLOCK(&worker->async);
 
     ucs_free(conn_request);
--- a/src/ucp/core/ucp_listener.h
+++ b/src/ucp/core/ucp_listener.h
@@ -34,6 +34,8 @@ typedef struct ucp_listener {
                                                  creates a handle to
                                                  connection request to the
                                                  remote endpoint */
+    int                            conn_reqs; /* count unprocessed connection
+                                                 requests */
     void                           *arg;      /* User's arg for the accept
                                                  callback */
     uct_worker_cb_id_t             prog_id;   /* Slow-path callback */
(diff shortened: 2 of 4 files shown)


changchengx:
No need to place ```ucp_listener::conn_reqs``` in the tail of the structure. User only uses the object pointer instead of object self.

src/ucp/core/ucp_listener.c
```diff
@@ -369,6 +369,10 @@ void ucp_listener_destroy(ucp_listener_h listener)
     UCS_ASYNC_UNBLOCK(&listener->worker->async);
 
     ucp_listener_free_uct_listeners(listener);
+
+    if (listener->conn_reqs != 0) {
+        ucs_warn("pending %d connection requests", listener->conn_reqs);
```
yosefe, line 374: 1. `ucs_warn("destroying listeneer %p with %u unprocessed connection requests"`
2. do it before line 371
3. space line after the "if" block

src/ucp/core/ucp_listener.h
```diff
@@ -34,6 +34,7 @@ typedef struct ucp_listener {
                                                  creates a handle to
                                                  connection request to the
                                                  remote endpoint */
+    int32_t                        conn_reqs; /* count connection requests */
```
yosefe, line 37: unsigned

src/ucp/wireup/wireup_cm.c
```diff
@@ -1201,11 +1203,13 @@ ucp_ep_cm_server_create_connected(ucp_worker_h worker, unsigned ep_init_flags,
 {
     ucp_tl_bitmap_t tl_bitmap;
     ucp_ep_h ep;
+    ucp_listener_h listener;
     ucs_status_t status;
     char client_addr_str[UCS_SOCKADDR_STRING_LEN];
     ucp_wireup_sockaddr_data_base_t *sa_data;
     ucp_object_version_t sa_data_version;
 
+    listener       = conn_request->listener;
```
yosefe, line 1212: init inline

src/ucp/core/ucp_listener.h
changchengx: @yosefe 
The unprocessed connection requests may not be more than ```2^31``` (qp number is 24 bits, socket fd number should be limited by system configuration and memory resources).
Could it keep using ```int32_t``` here to give obvious error info if there's something wrong in the code implmentation when the ```conn_reqs``` is lesss than 0?
If there's no need to do so, I'll change it to be ```unsigned```.

src/ucp/core/ucp_listener.c
changchengx: Done

src/ucp/wireup/wireup_cm.c
changchengx: Done

src/ucp/core/ucp_listener.h
yosefe: then, use int

changchengx: Done. Use ```int``` type here.

changchengx:
@yosefe It has passed all the CI TCs after changing the code according to the comments.

yosefe (approved):
pls squash

changchengx:
@yosefe Done. PR has been squashed

PR #8131, opened 2022-04-16
UCP/RNDV: Enabled non-contig datatypes in AM rendezvous.

rakhmets:
## What
Enabled non-contig datatypes in AM rendezvous.

## Why ?
The changes are part of UCP transition to new infrastructure.

### DIFF
--- a/src/ucp/am/rndv.c
+++ b/src/ucp/am/rndv.c
@@ -52,8 +52,7 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_am_rndv_proto_progress, (self),
 ucs_status_t ucp_am_rndv_rts_init(const ucp_proto_init_params_t *init_params)
 {
     if (!ucp_am_check_init_params(init_params, UCP_AM_OP_ID_MASK_ALL,
-                                  UCP_PROTO_SELECT_OP_FLAG_AM_EAGER) ||
-        (init_params->select_param->dt_class != UCP_DATATYPE_CONTIG)) {
+                                  UCP_PROTO_SELECT_OP_FLAG_AM_EAGER)) {
         return UCS_ERR_UNSUPPORTED;
     }
 
--- a/src/ucp/dt/datatype_iter.inl
+++ b/src/ucp/dt/datatype_iter.inl
@@ -503,7 +503,11 @@ ucp_datatype_iter_mem_reg(ucp_context_h context, ucp_datatype_iter_t *dt_iter,
                             (ucs_memory_type_t)dt_iter->mem_info.type, md_map,
                             uct_flags, &dt_iter->type.contig.memh);
     } else if (ucp_datatype_iter_is_class(dt_iter, UCP_DATATYPE_IOV, dt_mask)) {
-        return ucp_datatype_iter_iov_mem_reg(context, dt_iter, md_map, uct_flags);
+        return ucp_datatype_iter_iov_mem_reg(context, dt_iter, md_map,
+                                             uct_flags);
+    } else if (ucp_datatype_iter_is_class(dt_iter, UCP_DATATYPE_GENERIC,
+                                          dt_mask)) {
+        return UCS_OK;
     } else {
         ucs_error("datatype %s does not support registration",
                   ucp_datatype_class_names[dt_iter->dt_class]);
--- a/src/ucp/rndv/proto_rndv.inl
+++ b/src/ucp/rndv/proto_rndv.inl
@@ -48,8 +48,8 @@ ucp_proto_rndv_rts_request_init(ucp_request_t *req)
     status = ucp_datatype_iter_mem_reg(ep->worker->context,
                                        &req->send.state.dt_iter, rpriv->md_map,
                                        UCT_MD_MEM_ACCESS_RMA |
-                                       UCT_MD_MEM_FLAG_HIDE_ERRORS,
-                                       UCS_BIT(UCP_DATATYPE_CONTIG));
+                                               UCT_MD_MEM_FLAG_HIDE_ERRORS,
+                                       UCP_DT_MASK_ALL);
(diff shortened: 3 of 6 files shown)


src/ucp/rndv/proto_rndv.inl
```diff
@@ -45,13 +45,16 @@ ucp_proto_rndv_rts_request_init(ucp_request_t *req)
...
     }
 
-    status = ucp_datatype_iter_mem_reg(ep->worker->context,
-                                       &req->send.state.dt_iter, rpriv->md_map,
-                                       UCT_MD_MEM_ACCESS_RMA |
-                                       UCT_MD_MEM_FLAG_HIDE_ERRORS,
-                                       UCS_BIT(UCP_DATATYPE_CONTIG));
-    if (status != UCS_OK) {
-        return status;
+    if (!UCP_DT_IS_GENERIC(req->send.state.dt_iter.dt_class)) {
```
yosefe, line 48: since ucp_datatype_iter_mem_reg already checks the dt_class, maybe we can pass some parameter to it whether to print error or not? or just not fail if UCT_MD_MEM_FLAG_HIDE_ERRORS is set?

src/ucp/am/rndv.c
```diff
@@ -36,7 +36,8 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_am_rndv_proto_progress, (self),
     size_t max_rts_size;
 
     rpriv  = req->send.proto_config->priv;
-    status = UCS_PROFILE_CALL(ucp_proto_rndv_rts_request_init, req);
+    status = UCS_PROFILE_CALL(ucp_proto_rndv_rts_request_init, req,
+                              UCP_DT_MASK_CONTIG_IOV);
```
yosefe, line 40: why not UCP_DT_MASK_ALL?

src/ucp/tag/tag_rndv.c
```diff
@@ -134,7 +134,8 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_tag_rndv_rts_progress, (self),
     rpriv        = req->send.proto_config->priv;
     max_rts_size = sizeof(ucp_rndv_rts_hdr_t) + rpriv->packed_rkey_size;
 
-    status = UCS_PROFILE_CALL(ucp_proto_rndv_rts_request_init, req);
+    status = UCS_PROFILE_CALL(ucp_proto_rndv_rts_request_init, req,
+                              UCS_BIT(UCP_DATATYPE_CONTIG));
```
yosefe, line 138: why not UCP_DT_MASK_ALL?

src/ucp/dt/datatype_iter.inl
```diff
@@ -504,6 +504,8 @@ ucp_datatype_iter_mem_reg(ucp_context_h context, ucp_datatype_iter_t *dt_iter,
                             uct_flags, &dt_iter->type.contig.memh);
     } else if (ucp_datatype_iter_is_class(dt_iter, UCP_DATATYPE_IOV, dt_mask)) {
         return ucp_datatype_iter_iov_mem_reg(context, dt_iter, md_map, uct_flags);
+    } else if (uct_flags & UCT_MD_MEM_FLAG_HIDE_ERRORS) {
+        return UCS_OK;
```
brminich, line 508: what test is failed? I guess this flags is intended to hide possible registration errors on the specific MDs, while using incorrect datatype looks like a SW bug which needs to be reported

src/ucp/rndv/rndv_am.c
```diff
@@ -77,9 +77,6 @@ ucp_proto_rndv_am_bcopy_complete(ucp_request_t *req)
     if (req->send.rndv.rkey != NULL) {
         ucp_proto_rndv_rkey_destroy(req);
     }
-    ucp_datatype_iter_mem_dereg(req->send.ep->worker->context,
-                                &req->send.state.dt_iter,
-                                UCS_BIT(UCP_DATATYPE_CONTIG));
```
brminich, line 82: is it unrelated bug fix?

src/ucp/dt/datatype_iter.inl
rakhmets: We perform memory registration inside `ucp_proto_rndv_rts_request_init`.
Previously I added check for `GENERIC` datatype, and replaced it by check for the flag according to the https://github.com/openucx/ucx/pull/8131#discussion_r851996601

src/ucp/rndv/rndv_am.c
rakhmets: This is a redundant memory deregistration since `bcopy` doesn't perform registration.
This didn't lead to an error for `CONTIG` datatype, but for `IOV`.

yosefe (approved):

src/ucp/dt/datatype_iter.inl
brminich: i think checking for `GENERIC` datatype was cleaner. Using `UCT_MD_MEM_FLAG_HIDE_ERRORS` for checking proper datatype looks error prone to me. We could just remove the error from this function

yosefe: when this function is called without HIDE_ERRORS it means the caller expects registration to really succeed, IMO if we just remove the error it can hide real issues

brminich: but this `HIDE` flag is intended for other type of issues. What if we pass invalid datatype, due to a SW bug together with `HIDE` flag? Imo, hiding such errors is not a good approach

yosefe: so maybe add another wrapper inline function , so sometimes we would fail if datatype is invalid, and sometimes don't?

brminich (approved):

yosefe (dismissed):

PR #8133, opened 2022-04-18
UCT/IB: Backport device fatal handling

huaxiangfan:
## What
Backport https://github.com/Mellanox/ucx/pull/25

## Why ?
To prevent coredump in case of ifdown/ifup

### DIFF
--- a/src/uct/ib/base/ib_device.c
+++ b/src/uct/ib/base/ib_device.c
@@ -282,21 +282,41 @@ uct_ib_device_async_event_schedule_callback(uct_ib_device_t *dev,
 }
 
 static void
+uct_ib_device_async_event_dispatch_nolock(uct_ib_device_t *dev,
+                                          const uct_ib_async_event_t *event)
+{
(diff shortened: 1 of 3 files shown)


src/uct/ib/base/ib_device.c
```diff
@@ -506,9 +526,10 @@ void uct_ib_handle_async_event(uct_ib_device_t *dev, uct_ib_async_event_t *event
         level = UCS_LOG_LEVEL_DEBUG;
         break;
     case IBV_EVENT_DEVICE_FATAL:
+        uct_ib_device_async_event_dispatch_fatal(dev);
         snprintf(event_info, sizeof(event_info), "%s on port %d",
                  ibv_event_type_str(event->event_type), event->port_num);
-        level = UCS_LOG_LEVEL_ERROR;
+        level = UCS_LOG_LEVEL_DIAG;
```
evgeny-leksikov, line 532: I think it should be `UCS_LOG_LEVEL_FATAL` because actually this is not recoverable error.

evgeny-leksikov (approved):

yosefe (approved):

PR #8136, opened 2022-04-19
UCP/WORKER: cleanup dt_invalidate CBs

evgeny-leksikov:
## What
cleanup dt_invalidate CBs

## Why ?
resource cleanup

## How ?
invoke `ucp_request_dt_invalidate_progress` to complete scheduled requests

### DIFF
--- a/src/ucp/core/ucp_worker.c
+++ b/src/ucp/core/ucp_worker.c
@@ -2636,6 +2636,20 @@ static void ucp_worker_destroy_eps(ucp_worker_h worker,
     }
 }
 
+static int
+ucp_worker_dt_invalidate_remove_filter(const ucs_callbackq_elem_t *elem,
+                                       void *arg)
+{
+    ucs_assert(arg == NULL);
+
+    if (elem->cb == ucp_request_dt_invalidate_progress) {
+        elem->cb(elem->arg);
+        return 1;
+    }
+
+    return 0;
+}
+
 void ucp_worker_destroy(ucp_worker_h worker)
 {
     ucs_debug("destroy worker %p", worker);
@@ -2646,6 +2660,9 @@ void ucp_worker_destroy(ucp_worker_h worker)
     ucp_worker_destroy_eps(worker, &worker->all_eps, "all");
     ucp_worker_destroy_eps(worker, &worker->internal_eps, "internal");
     ucp_am_cleanup(worker);
+    ucs_callbackq_remove_if(&worker->uct->progress_q,
+                            ucp_worker_dt_invalidate_remove_filter, NULL);
+
     /* Put ucp_worker_remove_am_handlers after ucp_worker_discard_uct_ep_cleanup
      * to make sure iface->am[] always cleared.
      * ucp_worker_discard_uct_ep_cleanup might trigger ucp_worker_iface_deactivate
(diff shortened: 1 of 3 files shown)


hoopoepg (approved):

dmitrygx (approved):

src/ucp/core/ucp_worker.c
```diff
@@ -2646,6 +2660,9 @@ void ucp_worker_destroy(ucp_worker_h worker)
     ucp_worker_destroy_eps(worker, &worker->all_eps, "all");
     ucp_worker_destroy_eps(worker, &worker->internal_eps, "internal");
     ucp_am_cleanup(worker);
+    ucs_callbackq_remove_if(&worker->uct->progress_q,
+                            ucp_worker_dt_invalidate_remove_filter, NULL);
```
yosefe, line 2664: does it mean we destroyed a worker before all requests on it were completed?

evgeny-leksikov: yes, AFAIU it's possible in case of `TERMINATE_SIGNALED`

yosefe: i think it should be considered as bug in io_demo
in UCX, we can warn if destroying non-empty callbackq

evgeny-leksikov: do you think that's ok to block on waiting all requests after `TERMINATE_SIGNALED`? in common case it may take long time

yosefe: It seems like we already do it:
https://github.com/openucx/ucx/blob/c2b0874b1dac76a6d02e5142a8eb0d61a7544df5/test/apps/iodemo/io_demo.cc#L2323
so maybe there is a deeper issue..

evgeny-leksikov: > It seems like we already do it

It seems like we don't
https://github.com/openucx/ucx/blob/c2b0874b1dac76a6d02e5142a8eb0d61a7544df5/test/apps/iodemo/io_demo.cc#L2002

yosefe: hmm i think we need to still wait for requests; if we close endpoints with mode FORCE, all requests should complete

evgeny-leksikov: Ok, I'll open another PR with waiting all requests in io-demo, but I think this change has to be tested separately.
Also I'm going to rework this PR to complete the request in-place without scheduling to cb-queue when nothing to invalidate

yosefe: they may be completed later, but it's guaranteed the requests will complete without waiting for the remote peer

evgeny-leksikov: @yosefe you are right, this PR is useless from functionality perspective but I think It may be reworked to assert  on resources cleanup. If you think that it's redundant, free to close.

yosefe: i'm pretty sure ucs_callbackq_remove_if asserts if it's not empty; if it doesn't let's add it there

evgeny-leksikov: it prints `ucs_warn`, so closing this

PR #8138, opened 2022-04-19
[WIP] UCP/API: Enforce strongly typed enums

gleon99:
## What
Enforce types on enums, to the extent possible by the C standard and the compilers we use.

## Why ?
Prevent human errors caused by incorrect usage of enum values, therefore make our code more robust and idiot-proof.

## How ?
1.	Anonymous / named enums are now types (using `typedef`).
2.	Use typed enums wherever possible (vars, func params). Can’t use them in struct members, would be bad practice.
3.	Add compiler flags to enforce correct usage.
4.	Fix docs accordingly.

**The PR is not to be merged. Its purpose is to be a “demo” on a small part of the code of how the real change will look like – so that we can see whether it is appropriate, before proceeding to do more.**

### DIFF
--- a/src/ucp/core/ucp_ep.h
+++ b/src/ucp/core/ucp_ep.h
@@ -132,18 +132,18 @@ enum {
 /**
  * UCP endpoint statistics counters
  */
-enum {
+typedef enum {
     UCP_EP_STAT_TAG_TX_EAGER,
     UCP_EP_STAT_TAG_TX_EAGER_SYNC,
     UCP_EP_STAT_TAG_TX_RNDV,
     UCP_EP_STAT_LAST
-};
+} ucp_ep_stat_coutner_t;
 
 
 /**
  * Endpoint init flags
  */
-enum {
+typedef enum {
     UCP_EP_INIT_FLAG_MEM_TYPE          = UCS_BIT(0),  /**< Endpoint for local mem type transfers */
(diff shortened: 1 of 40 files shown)


gleon99:
@yosefe @brminich  WDYT? - See PR description.

shamisp:
This makes code better. I would start with internal changes first before changing externally exposed structures. 
I don't think it will break API/ABI but I think it may potentially generate warning in codes using UCX.

src/ucp/core/ucp_ep.h
```diff
@@ -132,18 +132,18 @@ enum {
...
  * UCP endpoint statistics counters
  */
-enum {
+typedef enum {
     UCP_EP_STAT_TAG_TX_EAGER,
     UCP_EP_STAT_TAG_TX_EAGER_SYNC,
     UCP_EP_STAT_TAG_TX_RNDV,
     UCP_EP_STAT_LAST
-};
+} ucp_ep_stat_coutner_t;
```
brminich, line 140: `counter`

brminich (commented):
the idea sounds good, but i'm afraid of that with such better error protection we will lose some flexibility.

gleon99:
Hi @gleon99 - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #8141, opened 2022-04-21
DOCS/FAQ: re-worded static build section

hoopoepg:
- re-worded some text in static build section

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -117,19 +117,19 @@ The file can be created using `ucx_info -Cf`.
 
 #### Build user application with UCX
 
-UCX supports metainformation subsystem based on pkg-config tool. In order to
-build the application with UCX development libraries.  For example, this is how
+In order to build the application with UCX development libraries, UCX supports a
+metainformation subsystem based on the pkg-config tool. For example, this is how
 pkg-config can be incorporated in a Makefile-based build:
 ```
 program: program.c
-        cc program.c $(pkg-config --cflags --libs ucx)
+        $(CC) program.c $(shell pkg-config --cflags --libs ucx)
 ```
-When linking with static UCX libraries, need to list explicitly all required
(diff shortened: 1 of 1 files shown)


yosefe (approved):

yosefe:
@tonycurtis @shamisp can you pls take a look?

docs/source/faq.md
```diff
@@ -117,19 +117,19 @@ The file can be created using `ucx_info -Cf`.
...
 pkg-config can be incorporated in a Makefile-based build:
 ```
 program: program.c
-        cc program.c $(pkg-config --cflags --libs ucx)
+        $(CC) program.c $(shell pkg-config --cflags --libs ucx)
 ```
-When linking with static UCX libraries, need to list explicitly all required
-transport modules.  For example, in order to support only cma and knem
-transports, can use this method:
+When linking with static UCX libraries, the user needs to list all required
```
shamisp, line 127: needs -> must

docs/source/faq.md
```diff
@@ -117,19 +117,19 @@ The file can be created using `ucx_info -Cf`.
...
+transport modules explicitly.  For example, in order to support only cma and
+knem transports, the user can use:
```
shamisp, line 129: have to use:

shamisp (changes requested):
@tonycurtis  ?

tonycurtis (dismissed):

docs/source/faq.md
hoopoepg: updated

docs/source/faq.md
hoopoepg: updated

shamisp (approved):

hoopoepg:
@yosefe ok to merge?

shamisp:
@hoopoepg it is good to go.

PR #8142, opened 2022-04-21
UCP/CORE/GTEST: Disable resolving remote EP ID when creating local EP by default

dmitrygx:
## What

Disable resolving remote EP ID when creating local EP by default.

## Why ?

To not create remote endpoints which were not closed after the local one is closed.

## How ?

1. Add `UCX_RESOLVE_REMOTE_EP_ID = { on, off, y, n, auto }` configuration parameter in UCP.
2. When creating "connect_to_worker" EPs, check this parameter and decide whether we need to resolve remote ID or not.
3. Resolve remote ID during creating a local EP only if configuration parameter is set to "y"/"on" or "auto" and error-handlign and keepalive are enabled.

### DIFF
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -801,6 +801,11 @@ UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx)
 
 class test_ucp_am_nbx_closed_ep : public test_ucp_am_nbx {
 public:
+    test_ucp_am_nbx_closed_ep()
+    {
+        modify_config("RESOLVE_REMOTE_EP_ID", "auto");
+    }
+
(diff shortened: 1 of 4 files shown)


test/gtest/ucp/test_ucp_am.cc
```diff
@@ -801,6 +801,11 @@ UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx)
 
 class test_ucp_am_nbx_closed_ep : public test_ucp_am_nbx {
 public:
+    test_ucp_am_nbx_closed_ep()
+    {
+        modify_config("RESOLVE_REMOTE_EP_ID", "on");
```
yosefe, line 806: "auto"

dmitrygx: yep, it should be `auto`
fixed

PR #8143, opened 2022-04-21
UCT/RDMACM: Remove locking from event loop

Artemy-Mellanox:
## What ?
Remove long lock over all pending RDMA_CM events.

## Why ?
We don't want to block main thread for long time.

## How ?
Release async lock after each event processed.
Also cache ibv_query_port() requests.
Keep event reading and processing under same lock to avoid collision with RDMA_CM id destruction.

### DIFF
--- a/src/uct/ib/rdmacm/rdmacm_cm.c
+++ b/src/uct/ib/rdmacm/rdmacm_cm.c
@@ -805,9 +805,7 @@ static void uct_rdmacm_cm_event_handler(int fd, ucs_event_set_types_t events,
             return;
         }
 
-        UCS_ASYNC_BLOCK(uct_rdmacm_cm_get_async(cm));
         uct_rdmacm_cm_process_event(cm, event);
-        UCS_ASYNC_UNBLOCK(uct_rdmacm_cm_get_async(cm));
     }
 }
 
@@ -941,7 +939,7 @@ UCS_CLASS_INIT_FUNC(uct_rdmacm_cm_t, uct_component_h component,
     status = ucs_async_set_event_handler(worker_priv->async->mode,
                                          self->ev_ch->fd, UCS_EVENT_SET_EVREAD,
                                          uct_rdmacm_cm_event_handler, self,
-                                         worker_priv->async);
+                                         NULL);
     if (status != UCS_OK) {
         goto err_destroy_ev_ch;
     }
--- a/src/uct/ib/rdmacm/rdmacm_cm.c
+++ b/src/uct/ib/rdmacm/rdmacm_cm.c
@@ -813,7 +813,7 @@ static void uct_rdmacm_cm_event_handler(int fd, ucs_event_set_types_t events,
     int                  ret;
 
     do {
-        UCS_ASYNC_BLOCK(uct_rdmacm_cm_get_async(cm));
+        UCS_ASYNC_TRY_BLOCK(uct_rdmacm_cm_get_async(cm), return);
 
         /* Fetch an event */
         ret = rdma_get_cm_event(cm->ev_ch, &event);
--- a/src/uct/ib/rdmacm/rdmacm_cm.c
+++ b/src/uct/ib/rdmacm/rdmacm_cm.c
@@ -736,7 +736,9 @@ uct_rdmacm_cm_process_event(uct_rdmacm_cm_t *cm, struct rdma_cm_event *event)
         break;
     case RDMA_CM_EVENT_ROUTE_RESOLVED:
         /* Client side event */
+        UCS_ASYNC_BLOCK(uct_rdmacm_cm_get_async(cm));
         uct_rdmacm_cm_handle_event_route_resolved(event);
+        UCS_ASYNC_UNBLOCK(uct_rdmacm_cm_get_async(cm));
         break;
     case RDMA_CM_EVENT_CONNECT_REQUEST:
         /* Server side event */
@@ -773,7 +775,9 @@ uct_rdmacm_cm_process_event(uct_rdmacm_cm_t *cm, struct rdma_cm_event *event)
         /* client and server error events */
     case RDMA_CM_EVENT_REJECTED:
     case RDMA_CM_EVENT_CONNECT_ERROR:
+        UCS_ASYNC_BLOCK(uct_rdmacm_cm_get_async(cm));
         uct_rdmacm_cm_handle_error_event(event);
+        UCS_ASYNC_UNBLOCK(uct_rdmacm_cm_get_async(cm));
         break;
     default:
         ucs_warn("unexpected RDMACM event: %s", rdma_event_str(event->event));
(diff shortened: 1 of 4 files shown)


yosefe (dismissed):

yosefe:
@Artemy-Mellanox pls add Why/How to PR desc

yosefe:
```
[swx-rdmz-ucx-gpu-01:32128:a:2401]   wireup_cm.c:549  Assertion `ucs_async_is_blocked(&(ep)->worker->async)' failed

/scrap/azure/agent-03/AZP_WORKSPACE/1/s/contrib/../src/ucp/wireup/wireup_cm.c: [ ucp_cm_client_resolve_cb() ]
      ...
      541     ucp_wireup_ep_t *cm_wireup_ep;
      542     char addr_str[UCS_SOCKADDR_STRING_LEN];
      543 
==>   544     UCP_EP_CM_CALLBACK_ENTER(ep, ucp_ep_get_cm_uct_ep(ep),
      545                              {
      546                                  ucs_assert(ep->flags & UCP_EP_FLAG_CLOSED);
      547                                  status = UCS_ERR_CANCELED;
```
failure in GO is relevant, seems need to fix asserts or add UCS_ASYNC_BLOCK in UCP CM callbacks

src/uct/ib/rdmacm/rdmacm_cm.c
```diff
@@ -83,11 +83,33 @@ uct_rdmacm_cm_device_context_init(uct_rdmacm_cm_device_context_t *ctx,
...
+    ctx->is_eth = 0;
+    for (i = 0; i < dev_attr.phys_port_count; ++i) {
+        ret = ibv_query_port(verbs, i + 1, &port_attr);
+        if (ret != 0) {
+            ucs_error("ibv_query_port (%s) failed: %m", dev_name);
+            return UCS_ERR_IO_ERROR;
+        }
+
+        if (IBV_PORT_IS_LINK_LAYER_ETHERNET(&port_attr)) {
+            ctx->is_eth |= UCS_BIT(i);
```
evgeny-leksikov, line 109: doing this in the loop seems weird to me, if different ports of the same device can be configured differently then this is a bug.
maybe initialize by first port and assert that others have same config?

evgeny-leksikov (commented):
So, locking scheme stays the same, right? pls don't forget update PR & commit tittle before merge

src/uct/ib/rdmacm/rdmacm_cm.c
Artemy-Mellanox: is_eth is bitset

Artemy-Mellanox:
> So, locking scheme stays the same, right? pls don't forget update PR & commit tittle before merge

No, we remove locking from event loop, see line 944

evgeny-leksikov:
> > So, locking scheme stays the same, right? pls don't forget update PR & commit tittle before merge
> 
> No, we remove locking from event loop, see line 944

And it works because we always take lock in UCP callbacks?

Artemy-Mellanox:
> > > So, locking scheme stays the same, right? pls don't forget update PR & commit tittle before merge
> > 
> > 
> > No, we remove locking from event loop, see line 944
> 
> And it works because we always take lock in UCP callbacks?

We remove long lock over all pending events, but still taking lock on each event in uct_rdmacm_cm_event_handler()

evgeny-leksikov:
I see now, thanks!

evgeny-leksikov (approved):

src/uct/ib/rdmacm/rdmacm_cm.c
```diff
@@ -792,23 +812,21 @@ static void uct_rdmacm_cm_event_handler(int fd, ucs_event_set_types_t events,
     struct rdma_cm_event *event;
     int                  ret;
 
-    for (;;) {
+    do {
+        UCS_ASYNC_BLOCK(uct_rdmacm_cm_get_async(cm));
```
yosefe, line 816: while (trylock == FAIL) yield()

src/uct/ib/rdmacm/rdmacm_cm.c
```diff
@@ -941,7 +959,7 @@ UCS_CLASS_INIT_FUNC(uct_rdmacm_cm_t, uct_component_h component,
     status = ucs_async_set_event_handler(worker_priv->async->mode,
                                          self->ev_ch->fd, UCS_EVENT_SET_EVREAD,
                                          uct_rdmacm_cm_event_handler, self,
-                                         worker_priv->async);
+                                         NULL);
```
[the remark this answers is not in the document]
Artemy-Mellanox: nope, events will be handled in next event_handler invocation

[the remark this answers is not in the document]
evgeny-leksikov: but then it will work great if progress thread do nothing related to UCX (or yields), but if it calls empty `ucp_worker_progress` in busy waiting (some recv, for example), then the lock might never (or too rare) be acquired by async thread.
maybe add some counter for try_block and then fallback to block?

src/uct/ib/rdmacm/rdmacm_cm.h
```diff
@@ -77,6 +77,7 @@ typedef struct uct_rdmacm_cm_device_context {
     uint32_t        log_reserved_qpn_granularity;
     uint32_t        num_dummy_qps;
     struct ibv_cq   *cq;
+    uint8_t         is_eth;
```
gleon99, line 80: Maybe add `bitmap` to the var name, and/or use `ucs_bitmap_t`, to emphasize that it is a bitmap and not a number.

src/uct/ib/rdmacm/rdmacm_cm.c
```diff
@@ -106,7 +106,7 @@ uct_rdmacm_cm_device_context_init(uct_rdmacm_cm_device_context_t *ctx,
         }
 
         if (IBV_PORT_IS_LINK_LAYER_ETHERNET(&port_attr)) {
-            ctx->is_eth |= UCS_BIT(i);
```
gleon99: Why `1` and not `i`? Same regarding `UCS_BITMAP_GET`.

src/uct/ib/rdmacm/rdmacm_cm.c
```diff
@@ -792,23 +812,21 @@ static void uct_rdmacm_cm_event_handler(int fd, ucs_event_set_types_t events,
     struct rdma_cm_event *event;
     int                  ret;
 
-    for (;;) {
+    do {
+        UCS_ASYNC_TRY_BLOCK(uct_rdmacm_cm_get_async(cm), return);
```
yosefe, line 816: maybe limit the number of events to process in one handler call (try 4/8/16) instead of setting async_context=NULL and locking here

yosefe:
@Artemy-Mellanox is this still needed?

PR #8144, opened 2022-04-21
UCP/WIREUP: Change CM wireup fallback log level

dmitrygx:
## What

Change CM wireup fallback log level.

## Why ?

To reduce amount of traces in case of `UCX_LOG_LEVEL=diag`

## How ?

1. Change logging of fallback as follows:
- debug: has another variant of endpoint initialization flags
- diag: has another CM
- error: otherwise
2. Trace CM fallback as `DIAG`.

### DIFF
--- a/src/ucp/wireup/wireup_cm.c
+++ b/src/ucp/wireup/wireup_cm.c
@@ -86,10 +86,9 @@ unsigned ucp_cm_client_try_next_cm_progress(void *arg)
         ucp_wireup_ep_destroy_next_ep(cm_wireup_ep);
     }
 
-    ucs_debug("client switching from %s to %s in attempt to connect to the"
-              " server",
-              ucp_context_cm_name(context, cm_idx - 1),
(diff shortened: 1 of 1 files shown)


src/ucp/wireup/wireup_cm.c
```diff
@@ -86,10 +86,9 @@ unsigned ucp_cm_client_try_next_cm_progress(void *arg)
         ucp_wireup_ep_destroy_next_ep(cm_wireup_ep);
     }
 
-    ucs_debug("client switching from %s to %s in attempt to connect to the"
-              " server",
-              ucp_context_cm_name(context, cm_idx - 1),
-              ucp_context_cm_name(context, cm_idx));
+    ucs_diag("ep=%p: client switching from %s to %s in attempt to connect to"
```
yosefe, line 89: "ep %p

dmitrygx: fixed

yosefe (approved):

PR #8148, opened 2022-04-22
UCT/IB/UD: Remove unneeded ECN flag

dmitrygx:
## What

Remove unneeded ECN flag.

## Why ?

It is not used.

## How ?

1. Remove `UCT_UD_PACKET_FLAG_ECN` flag.
2. Remove all mentions of ECN in UD code.

### DIFF
--- a/src/uct/ib/ud/base/ud_def.h
+++ b/src/uct/ib/ud/base/ud_def.h
@@ -54,7 +54,7 @@ enum {
 enum {
     UCT_UD_PACKET_FLAG_AM      = UCS_BIT(24),
     UCT_UD_PACKET_FLAG_ACK_REQ = UCS_BIT(25),
-    UCT_UD_PACKET_FLAG_ECN     = UCS_BIT(26),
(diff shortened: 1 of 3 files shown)


yosefe (commented):
IMO for compatibility, better keep this flag in the wire protocol, even if currently unused

dmitrygx:
> IMO for compatibility, better keep this flag in the wire protocol, even if currently unused

do you mean for future usage?
I saw this wasn't used even in UCX 1.2

yosefe:
> do you mean for future usage?

yes, for future usage

dmitrygx:
> > do you mean for future usage?
> 
> yes, for future usage

btw, we don't have a special handling for it. we will ignore if smth unknown was received.
but we print `E` bit - so, it worse keeping it as is

yosefe:
if we just remove, it could get reused it for smth else. IMO the cleanup benefit is not worth it, in case of something that affects wore format.

PR #8150, opened 2022-04-22
UCP: Implementation of routine to query datatype attributes

rdietric:
## What
Implementation of the the routine `ucp_dt_query` according to PR #8120. Currently, the only datatype attribute to query is the packed size.

## Why ?
The size of a message or data transfer is a performance-relevant information. Performance tools should be able to query this information similar to `MPI_Type_size` to query the size of any MPI datatype

## How ?
For _contig_ data types, the return value of `ucp_contig_dt_elem_size()` provides the element size.
For generic data types, a field `packed_size` has been added to `ucp_dt_generic_t`, which is set with the user-defined pack routine. When the size of the generic data type is queried, the value of this field is used. All other data types are not supported and `UCS_ERR_INVALID_PARAM` is returned as `ucs_status_t`.

A unit tests verifies the expected behavior for _contig_, _iov_ and _generic_ data types. It is also checked that `UCS_ERR_UNSUPPORTED` is returned, if the generic datatype is queried before packing.

### DIFF
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -517,10 +517,18 @@ enum ucp_dt_type {
  * @ingroup UCP_DATATYPE
  * @brief UCP datatype attributes field mask.
  *
- * The enumeration allows querying a specific field in @ref ucp_datatype_attr_t.
+ * The enumeration allows specifying which fields in @ref ucp_datatype_attr_t
+ * are present and which datatype attributes are queried.
  */
 enum ucp_datatype_attr_field {
-    UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE = UCS_BIT(0) /**< packed datatype size */
+    /** @ref ucp_datatype_attr_t::packed_size field is queried. */
+    UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE = UCS_BIT(0),
+
+    /** @ref ucp_datatype_attr_t::buffer field is set. */
+    UCP_DATATYPE_ATTR_FIELD_BUFFER      = UCS_BIT(1),
+
+    /** @ref ucp_datatype_attr_t::count field is set. */
+    UCP_DATATYPE_ATTR_FIELD_COUNT       = UCS_BIT(2)
 };
 
 
@@ -939,7 +947,7 @@ typedef struct ucp_generic_dt_ops {
  * @ingroup UCP_DATATYPE
  * @brief UCP datatype attributes
  *
- * This structure provides attributes that can be queried for a UCP datatype.
+ * This structure provides attributes of a UCP datatype.
  */
 typedef struct ucp_datatype_attr {
     /**
@@ -947,12 +955,28 @@ typedef struct ucp_datatype_attr {
      * ucp_datatype_attr_field. Fields not specified in this mask will be
      * ignored. Provides ABI compatibility with respect to adding new fields.
      */
-    uint64_t field_mask;
+    uint64_t   field_mask;
 
     /**
-     * Packed size of a single element of the given datatype.
+     * Packed size of the given datatype. (output parameter)
      */
-    size_t   packed_size;
+    size_t     packed_size;
+
+    /**
+     * Pointer to a data buffer of the associated data type.
+     * This field is optional.
+     * If @ref UCP_DATATYPE_ATTR_FIELD_BUFFER is not set in @ref field_mask,
+     * this field defaults to @e NULL.
+     */
+    const void *buffer;
+
+    /**
+     * Number of elements in @a buffer.
+     * This value is optional.
+     * If @ref UCP_DATATYPE_ATTR_FIELD_COUNT is not set in @ref field_mask, the
+     * value of this field defaults to 1.
+     */
+    size_t     count;
 } ucp_datatype_attr_t;
 
 
@@ -3768,15 +3792,8 @@ void ucp_dt_destroy(ucp_datatype_t datatype);
  * When @ref UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE is set in @a field_mask of @a attr,
  * the field @a packed_size is set to the packed size (bytes) of the datatype.
  *
- * @note The size of a generic datatype is available only after it has been used
- *       in a communication operation. If it is queried before,
- *       @ref UCS_ERR_UNSUPPORTED will be returned.
- *
- * @note The size of IOV data cannot be determined from the UCP datatype. If an
- *       IOV @a datatype is passed, @ref UCS_ERR_INVALID_PARAM will be returned.
- *
- * @param [in]  datatype     Datatype object to query.
- * @param [out] attr         Filled with attributes of the datatype.
+ * @param [in]    datatype   Datatype object to query.
+ * @param [inout] attr       Filled with attributes of the datatype.
  *
  * @return Error code as defined by @ref ucs_status_t
  */
--- a/src/ucp/dt/dt.c
+++ b/src/ucp/dt/dt.c
@@ -142,3 +142,49 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
     state->offset += result_len;
     return result_len;
 }
+
+ucs_status_t ucp_dt_query(ucp_datatype_t datatype, ucp_datatype_attr_t *attr)
+{
+    ucp_dt_generic_t *dt_gen;
+    void *state_gen;
+    size_t count;
+
+    /* Currently, the only datatype attribute to query is the packed size. */
+    if (!(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE)) {
+        return UCS_OK;
+    }
+
+    count = UCP_ATTR_VALUE(DATATYPE, attr, count, COUNT, 1);
+
+    switch (datatype & UCP_DATATYPE_CLASS_MASK) {
+    case UCP_DATATYPE_CONTIG:
+        attr->packed_size = ucp_contig_dt_elem_size(datatype) * count;
+        return UCS_OK;
+    case UCP_DATATYPE_IOV:
+        if (!(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_BUFFER) ||
+            (attr->buffer == NULL)) {
+            return UCS_ERR_INVALID_PARAM;
+        }
+
+        attr->packed_size = ucp_dt_iov_length(attr->buffer, count);
+        return UCS_OK;
+    case UCP_DATATYPE_GENERIC:
+        if (!(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_BUFFER) ||
+            (attr->buffer == NULL)) {
+            return UCS_ERR_INVALID_PARAM;
+        }
+
+        dt_gen = ucp_dt_to_generic(datatype);
+        if (dt_gen == NULL) {
+            return UCS_ERR_INVALID_PARAM;
+        }
+
+        state_gen = dt_gen->ops.start_pack(dt_gen->context, attr->buffer,
+                                           count);
+        attr->packed_size = dt_gen->ops.packed_size(state_gen);
+        dt_gen->ops.finish(state_gen);
+        return UCS_OK;
+    default:
+        return UCS_ERR_INVALID_PARAM;
+    }
+}
--- a/test/gtest/ucp/test_ucp_dt.cc
+++ b/test/gtest/ucp/test_ucp_dt.cc
@@ -181,6 +181,53 @@ private:
 
 ucp_dt_generic_t* test_ucp_dt_iter::dt_gen = 0;
 
+UCS_TEST_P(test_ucp_dt_iter, datatype_query) {
+    ucs_status_t status;
+    ucp_datatype_t datatype           = GetParam();
+    ucp_datatype_attr_t datatype_attr = {};
+    datatype_attr.field_mask          = UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE;
+
+    switch (datatype & UCP_DATATYPE_CLASS_MASK) {
+    case UCP_DATATYPE_CONTIG:
+        status = ucp_dt_query(datatype, &datatype_attr);
+        ASSERT_EQ(UCS_OK, status);
+        EXPECT_GT(datatype_attr.packed_size, 0);
+        break;
+    case UCP_DATATYPE_GENERIC:
+    {
+        const size_t count        = 2;
+        datatype_attr.count       = count;
+        datatype_attr.field_mask |= UCP_DATATYPE_ATTR_FIELD_COUNT;
+
+        int32_t buf[count]        = {1, 2};
+        datatype_attr.buffer      = buf;
+        datatype_attr.field_mask |= UCP_DATATYPE_ATTR_FIELD_BUFFER;
+
+        status = ucp_dt_query(datatype, &datatype_attr);
+        ASSERT_EQ(UCS_OK, status);
+        EXPECT_GT(datatype_attr.packed_size, 0);
+        break;
+    }
+    case UCP_DATATYPE_IOV:
+    {
+        const size_t count        = 2;
+        datatype_attr.count       = count;
+        datatype_attr.field_mask |= UCP_DATATYPE_ATTR_FIELD_COUNT;
+
+        ucp_dt_iov_t buf[count]   = {{NULL, 8}, {NULL, 4}};
+        datatype_attr.buffer      = buf;
+        datatype_attr.field_mask |= UCP_DATATYPE_ATTR_FIELD_BUFFER;
+
+        status = ucp_dt_query(datatype, &datatype_attr);
+        ASSERT_EQ(UCS_OK, status);
+        EXPECT_EQ(12, datatype_attr.packed_size);
+        break;
+    }
+    default:
+        break;
+    }
+}
+
 UCS_TEST_P(test_ucp_dt_iter, pack_100b) {
     test_pack(100);
 }


rdietric:
Since nobody reviewed yet, I rebased to have the new _ucp.h_ with `ucp_dt_query` in the branch. I also fixed a bug in the implementation and another issue in the unit test. The current pipeline failures seem unrelated to this PR.

brminich:
@rakhmets, can you please review?

src/ucp/dt/dt.c
```diff
@@ -142,3 +143,35 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
+
+    switch (datatype & UCP_DATATYPE_CLASS_MASK) {
+    case UCP_DATATYPE_CONTIG:
+        attr->packed_size = ucp_contig_dt_elem_size(datatype);
+        return UCS_OK;
+    case UCP_DATATYPE_GENERIC:
+        dt_gen = ucp_dt_to_generic(datatype);
+        ucs_assert(NULL != dt_gen);
+        attr->packed_size = dt_gen->packed_size;
+        if (attr->packed_size) {
```
brminich, line 168: `(attr->packed_size != 0)`

brminich (commented):
only minor comments

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,39 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
 
 ucp_dt_generic_t* test_ucp_dt_iter::dt_gen = 0;
 
+/* The following tests reuse the generic datatype and its packed size will be
+ * set after the first packing. */
+UCS_TEST_P(test_ucp_dt_iter, datatype_query) {
```
rakhmets, line 186:
```suggestion
UCS_TEST_P(test_ucp_dt_iter, datatype_query) 
{
```

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,39 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
...
+    ucs_status_t status;
+    ucp_datatype_t datatype = GetParam();
+    ucp_datatype_attr_t datatype_attr = {};
+    datatype_attr.field_mask = UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE;
```
rakhmets, line 190:
```suggestion
    ucp_datatype_t datatype           = GetParam();
    ucp_datatype_attr_t datatype_attr = {};
    datatype_attr.field_mask          = UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE;
```

test/gtest/ucp/test_ucp_dt.cc
rdietric: The existing `UCS_TEST_P` have the curly block opening bracket on the same line. I am okay with changing, but I wanted to be consistent with the existing code.

src/ucp/dt/dt.c
```diff
@@ -142,3 +143,35 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
     return result_len;
 }
+
+ucs_status_t ucp_dt_query(ucp_datatype_t datatype, ucp_datatype_attr_t *attr)
+{
+    ucp_dt_generic_t *dt_gen;
+
+    if (!attr) {
+        return UCS_ERR_INVALID_PARAM;
+    }
```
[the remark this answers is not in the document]
rdietric: Removed the if block.

src/ucp/dt/datatype_iter.inl
```diff
@@ -80,7 +80,9 @@ ucp_datatype_generic_iter_init(ucp_context_h context, void *buffer,
         state = dt_gen->ops.start_unpack(dt_gen->context, buffer, count);
     }
 
-    dt_iter->length              = dt_gen->ops.packed_size(state);
+    dt_gen->packed_size = dt_gen->ops.packed_size(state);
+
+    dt_iter->length              = dt_gen->packed_size;
```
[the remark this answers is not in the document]
rdietric: Done.

src/ucp/dt/dt.c
rdietric: Changed.

src/ucp/dt/dt_generic.c
```diff
@@ -27,9 +27,12 @@ ucs_status_t ucp_dt_create_generic(const ucp_generic_dt_ops_t *ops, void *contex
...
     }
 
-    dt_gen->ops     = *ops;
-    dt_gen->context = context;
-    *datatype_p     = ucp_dt_from_generic(dt_gen);
+    dt_gen->ops         = *ops;
+    dt_gen->context     = context;
+    dt_gen->packed_size = 0;
+
+    *datatype_p = ucp_dt_from_generic(dt_gen);
```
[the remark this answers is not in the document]
rdietric: Done.

test/gtest/ucp/test_ucp_dt.cc
rdietric: Changed according to your suggestion.

src/ucp/dt/dt_generic.c
```diff
@@ -27,9 +27,11 @@ ucs_status_t ucp_dt_create_generic(const ucp_generic_dt_ops_t *ops, void *contex
         return UCS_ERR_NO_MEMORY;
     }
 
-    dt_gen->ops     = *ops;
-    dt_gen->context = context;
-    *datatype_p     = ucp_dt_from_generic(dt_gen);
+    dt_gen->ops         = *ops;
+    dt_gen->context     = context;
+    dt_gen->packed_size = 0;
```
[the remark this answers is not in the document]
rdietric: I actually used 0 to allow checks like `if (attr->packed_size)`, but I changed it according to your suggestion, so we can also use `SIZE_MAX` now. I think that a datatype of size 0 does not make sense. If you still think it's better to initialize with `SIZE_MAX`, I am okay doing so, especially since I probably don't know all the data descriptions/types that are possible with UCX.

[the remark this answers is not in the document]
brminich: i think having 0-sized type is more likely than SIZE_MAX, so I'd change it.
@yosefe, wdyt?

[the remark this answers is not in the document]
yosefe: I don't think we can assume that packed size of bufferA is same as packed size of bufferB. It depends on the contents.
Perhaps dt_query routine should also accept buffer pointer and count.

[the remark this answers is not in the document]
rdietric: You are right, if we consider compression, the packed size of a generic datatype can be different depending on the data. I thought of packing more like C struct packing/padding.
This basically means that the packed size can be different for each call to a communication function. Hence, the query would always return the same value for contiguous datatypes and could return different values after each communication operation for the same generic datatype. The function then still serves its purpose and allows me to find out how much data is transferred.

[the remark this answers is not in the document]
rdietric: If we add the buffer pointer and count to `dt_query`, we could also support IOV. Should we then return the packed message/buffer size instead of the packed datatype size? Alternatively, we would have to divide by count.
I assume that `start_pack` and `packed_size` are short calls, so that is doesn't hurt calling them twice (a second time for the size query).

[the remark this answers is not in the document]
rdietric: Has a decision been made here? I agree with @yosefe that passing buffer and count and return the buffer size fits better to the existing UCP datatypes. The user of `ucp_dt_query` can still pass "1" as count to return the size of one element.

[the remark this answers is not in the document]
yosefe: @shamisp ping on this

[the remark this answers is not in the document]
shamisp: @yosefe This is good question actually. In our recent research work the size also depended on the target EP in addition to buffer and count - think about DT conversion across different architectures. Do we use this for any of the MPI call ? I don't think MPI passes buffer (I have to double check)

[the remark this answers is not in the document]
rdietric: I did not come across the EP during implementation. The current implementation basically returns the last datatype size, which has been determined during the regular message/transfer preparation. This should include the EP.

src/ucp/api/ucp.h
```diff
@@ -3768,19 +3768,17 @@ void ucp_dt_destroy(ucp_datatype_t datatype);
  * When @ref UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE is set in @a field_mask of @a attr,
  * the field @a packed_size is set to the packed size (bytes) of the datatype.
  *
- * @note The size of a generic datatype is available only after it has been used
- *       in a communication operation. If it is queried before,
- *       @ref UCS_ERR_UNSUPPORTED will be returned.
- *
- * @note The size of IOV data cannot be determined from the UCP datatype. If an
- *       IOV @a datatype is passed, @ref UCS_ERR_INVALID_PARAM will be returned.
+ * @note The parameters @a buffer and @a count are ignored for contiguous datatypes.
```
yosefe, line 3771: IMO this note is redundant

src/ucp/api/ucp.h
```diff
@@ -3768,19 +3768,17 @@ void ucp_dt_destroy(ucp_datatype_t datatype);
...
  *
  * @param [in]  datatype     Datatype object to query.
+ * @param [in]  buffer       Buffer for generic or IOV datatypes (@ref ucp_dt_iov_t).
```
yosefe, line 3774: No need to document the implementation as it may change.
should be something like:
`@param [in] buffer Pointer to a data buffer of the type @a datatype.`

src/ucp/dt/dt.c
```diff
@@ -142,3 +142,38 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
     state->offset += result_len;
     return result_len;
 }
+
+ucs_status_t ucp_dt_query(ucp_datatype_t datatype, const void *buffer, size_t count, ucp_datatype_attr_t *attr)
```
yosefe, line 146: pls check code format
https://github.com/openucx/ucx/wiki/Code-style-checking

src/ucp/api/ucp.h
```diff
@@ -3768,19 +3768,17 @@ void ucp_dt_destroy(ucp_datatype_t datatype);
...
+ * @param [in]  count        Number of elements in @a buffer.
  * @param [out] attr         Filled with attributes of the datatype.
  *
  * @return Error code as defined by @ref ucs_status_t
  */
-ucs_status_t ucp_dt_query(ucp_datatype_t datatype, ucp_datatype_attr_t *attr);
+ucs_status_t ucp_dt_query(ucp_datatype_t datatype, const void *buffer,
```
yosefe, line 3780: @shamisp do you think it would be better to make buffer,count parameters optional using a struct?
PS. This API was not released yet

src/ucp/dt/dt.c
```diff
@@ -142,3 +142,38 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
+{
+    ucp_dt_generic_t *dt_gen;
+    void             *state_gen;
+
+    /* Currently, the only datatype attribute to query is the packed size. */
+    if (!(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE)) {
+        return UCS_ERR_UNSUPPORTED;
```
yosefe, line 153: return UCS_OK

src/ucp/dt/dt.c
```diff
@@ -142,3 +142,38 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
+    }
+
+    switch (datatype & UCP_DATATYPE_CLASS_MASK) {
+    case UCP_DATATYPE_CONTIG:
+        attr->packed_size = ucp_contig_dt_elem_size(datatype);
+        return UCS_OK;
+    case UCP_DATATYPE_IOV:
+        if (buffer) {
```
yosefe, line 161: if (buffer == NULL) return UCS_ERR_INVALID_PARAM

src/ucp/dt/dt.c
```diff
@@ -142,3 +142,38 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
+            attr->packed_size = ucp_dt_iov_length(buffer, count);
+            return UCS_OK;
+        }
+        return UCS_ERR_INVALID_PARAM;
+    case UCP_DATATYPE_GENERIC:
+        dt_gen = ucp_dt_to_generic(datatype);
+        ucs_assert(NULL != dt_gen);
```
yosefe, line 168: return invalid param if NULL

src/ucp/dt/dt.c
```diff
@@ -142,3 +142,38 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
+        if (buffer) {
+            state_gen = dt_gen->ops.start_pack(dt_gen->context, buffer, count);
+            attr->packed_size = dt_gen->ops.packed_size(state_gen);
+            return UCS_OK;
+        }
+        return UCS_ERR_INVALID_PARAM;
+    default:
+        attr->packed_size = 0;
```
yosefe, line 176: not needed

src/ucp/dt/dt_generic.c
```diff
@@ -27,9 +27,10 @@ ucs_status_t ucp_dt_create_generic(const ucp_generic_dt_ops_t *ops, void *contex
         return UCS_ERR_NO_MEMORY;
     }
 
-    dt_gen->ops     = *ops;
-    dt_gen->context = context;
-    *datatype_p     = ucp_dt_from_generic(dt_gen);
+    dt_gen->ops         = *ops;
+    dt_gen->context     = context;
+    *datatype_p         = ucp_dt_from_generic(dt_gen);
```
yosefe, line 32: revert

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,44 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
 
 ucp_dt_generic_t* test_ucp_dt_iter::dt_gen = 0;
 
+/* The following tests reuse the generic datatype and its packed size will be
+ * set after the first packing. */
```
yosefe, line 185: is it still relevant after adding buffer,count?

src/ucp/dt/dt.c
rdietric: Moved `count` and `attr` parameters to next line. (Unfortunately, git-clang-format from the UCX LLVM fork fails to work properly on my local system.)

src/ucp/api/ucp.h
rdietric: Redundant with the description of the `buffer` parameter? Should that go into another PR, together with the API change?

src/ucp/dt/dt.c
rdietric: Changed.

src/ucp/dt/dt.c
rdietric: Changed.

src/ucp/dt/dt.c
rdietric: Changed.

src/ucp/dt/dt.c
rdietric: Removed the line.

src/ucp/dt/dt_generic.c
rdietric: Done.

test/gtest/ucp/test_ucp_dt.cc
rdietric: No, it is not. I removed the comment.

src/ucp/api/ucp.h
shamisp: @yosefe - I think it is a good idea.

src/ucp/api/ucp.h
yosefe: We should remove this note because it describes an implementation-defined behavior.
On API layer, we should document only the guarantees, like a specification.

src/ucp/dt/dt.c
```diff
@@ -143,37 +143,37 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
         dt_gen = ucp_dt_to_generic(datatype);
         ucs_assert(NULL != dt_gen);
-        if (buffer) {
-            state_gen = dt_gen->ops.start_pack(dt_gen->context, buffer, count);
-            attr->packed_size = dt_gen->ops.packed_size(state_gen);
-            return UCS_OK;
-        }
-        return UCS_ERR_INVALID_PARAM;
+        state_gen = dt_gen->ops.start_pack(dt_gen->context, buffer, count);
+        attr->packed_size = dt_gen->ops.packed_size(state_gen);
```
yosefe, line 174: need to also call dt_gen->ops.finish, or something may leak

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,41 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
...
+        ASSERT_EQ(UCS_OK, status);
+        EXPECT_TRUE(datatype_attr.packed_size > 0);
+        break;
+    case UCP_DATATYPE_GENERIC:
+    {
+        const size_t count = 2;
+        int32_t buf[count] = {1, 2};
+        status = ucp_dt_query(datatype, buf, count, &datatype_attr);
+        ASSERT_EQ(UCS_OK, status);
+        EXPECT_TRUE(datatype_attr.packed_size > 0);
```
yosefe, line 202: EXPECT_GT(datatype_attr.packed_size, 0)

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,41 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
...
+        break;
+    }
+    case UCP_DATATYPE_IOV:
+    {
+        const size_t count = 2;
+        ucp_dt_iov_t buf[count] = {{NULL, 8}, {NULL, 4}};
+        status = ucp_dt_query(datatype, buf, count, &datatype_attr);
+        ASSERT_EQ(UCS_OK, status);
+        EXPECT_TRUE(datatype_attr.packed_size == 12);
```
yosefe, line 211: EXPECT_EQ(12, datatype_attr.packed_size)

src/ucp/api/ucp.h
yosefe: @rdietric let's move buffer,count into ucp_datatype_attr_t (with corresponding flags)

src/ucp/dt/dt.c
rdietric: Added `dt_gen->ops.finish(state_gen);`

test/gtest/ucp/test_ucp_dt.cc
rdietric: Changed.

test/gtest/ucp/test_ucp_dt.cc
rdietric: Changed.

src/ucp/api/ucp.h
```diff
@@ -517,10 +517,18 @@ enum ucp_dt_type {
  * @ingroup UCP_DATATYPE
  * @brief UCP datatype attributes field mask.
  *
- * The enumeration allows querying a specific field in @ref ucp_datatype_attr_t.
+ * The enumeration allows specifying which fields in @ref ucp_datatype_attr_t
+ * are present and which datatype attributes are queried.
  */
 enum ucp_datatype_attr_field {
-    UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE = UCS_BIT(0) /**< packed datatype size */
+    /** Query the packed datatype size. */
```
yosefe, line 524: `/** @ref ucp_datatype_attr_t::packed_size field is queried. */`

src/ucp/api/ucp.h
```diff
@@ -939,20 +947,36 @@ typedef struct ucp_generic_dt_ops {
  * @ingroup UCP_DATATYPE
  * @brief UCP datatype attributes
  *
- * This structure provides attributes that can be queried for a UCP datatype.
+ * This structure provides attributes of a UCP datatype.
```
yosefe, line 950: This structure provides attributes for querying a UCP datatype
@tonycurtis WDYT?

src/ucp/api/ucp.h
```diff
@@ -939,20 +947,36 @@ typedef struct ucp_generic_dt_ops {
...
+     * this field defaults to @e NULL.
      */
-    size_t   packed_size;
+    const void *buffer;
+
+    /**
+     * Number of elements in @a buffer.
+     * This value is optional.
+     * If @ref UCP_DATATYPE_ATTR_FIELD_COUNT is not set in @ref field_mask, the
+     * value of this field defaults to 0.
```
yosefe, line 977: maybe the default should be 1?

src/ucp/api/ucp.h
```diff
@@ -3768,15 +3792,10 @@ void ucp_dt_destroy(ucp_datatype_t datatype);
  * When @ref UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE is set in @a field_mask of @a attr,
  * the field @a packed_size is set to the packed size (bytes) of the datatype.
  *
- * @note The size of a generic datatype is available only after it has been used
- *       in a communication operation. If it is queried before,
- *       @ref UCS_ERR_UNSUPPORTED will be returned.
- *
- * @note The size of IOV data cannot be determined from the UCP datatype. If an
- *       IOV @a datatype is passed, @ref UCS_ERR_INVALID_PARAM will be returned.
+ * @note The parameters @a buffer and @a count are ignored for contiguous datatypes.
```
yosefe, line 3795: pls remove this comment

src/ucp/dt/dt.c
```diff
@@ -142,3 +142,43 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
     state->offset += result_len;
     return result_len;
 }
+
+ucs_status_t ucp_dt_query(ucp_datatype_t datatype, ucp_datatype_attr_t *attr)
+{
+    ucp_dt_generic_t *dt_gen;
+    void             *state_gen;
```
yosefe, line 149: no need to vertical-align - remove spaces

src/ucp/dt/dt.c
```diff
@@ -142,3 +142,43 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
+    if (!(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE)) {
+        return UCS_OK;
+    }
+
+    switch (datatype & UCP_DATATYPE_CLASS_MASK) {
+    case UCP_DATATYPE_CONTIG:
+        attr->packed_size = ucp_contig_dt_elem_size(datatype);
+        return UCS_OK;
+    case UCP_DATATYPE_IOV:
+        if (attr->buffer == NULL ||
```
yosefe, line 161: 1. ( ) around == expression
2. when accessing ANY field in "attr", it should be protected by field_mask check
see for example: https://github.com/openucx/ucx/blob/6fe27d4b7a2d3dbc7ea68aee1a8f491c30a3cf57/src/ucp/core/ucp_mm.c#L655

src/ucp/dt/dt.c
```diff
@@ -142,3 +142,43 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
+        }
+        attr->packed_size = ucp_dt_iov_length(attr->buffer, attr->count);
+        return UCS_OK;
+    case UCP_DATATYPE_GENERIC:
+        if (attr->buffer == NULL ||
+            !(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_BUFFER) ||
+            !(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_COUNT)) {
+            return UCS_ERR_INVALID_PARAM;
+        }
+        dt_gen = ucp_dt_to_generic(datatype);
```
yosefe, line 174: space line before

src/ucp/dt/dt.c
```diff
@@ -142,3 +142,43 @@ size_t ucp_dt_pack(ucp_worker_h worker, ucp_datatype_t datatype,
...
+    case UCP_DATATYPE_CONTIG:
+        attr->packed_size = ucp_contig_dt_elem_size(datatype);
+        return UCS_OK;
+    case UCP_DATATYPE_IOV:
+        if (attr->buffer == NULL ||
+            !(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_BUFFER) ||
+            !(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_COUNT)) {
+            return UCS_ERR_INVALID_PARAM;
+        }
+        attr->packed_size = ucp_dt_iov_length(attr->buffer, attr->count);
```
yosefe, line 166: space line before

src/ucp/dt/dt.c
yosefe: return invalid param if NULL

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,51 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
...
+    ucs_status_t status;
+    ucp_datatype_t datatype           = GetParam();
+    ucp_datatype_attr_t datatype_attr = {};
+    datatype_attr.field_mask          = UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE;
+
+    switch (datatype & UCP_DATATYPE_CLASS_MASK) {
+    case UCP_DATATYPE_CONTIG:
+        status = ucp_dt_query(datatype, &datatype_attr);
+        ASSERT_EQ(UCS_OK, status);
+        EXPECT_TRUE(datatype_attr.packed_size > 0);
```
yosefe, line 194: EXPECT_GT(datatype_attr.packed_size, 0);

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,51 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
...
+        break;
+    case UCP_DATATYPE_GENERIC:
+    {
+        datatype_attr.count = 2;
```
yosefe, line 198: align = on column - in multiple places
pls check clang-format output and fix relevant code style issues

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,51 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
...
+        datatype_attr.field_mask |= UCP_DATATYPE_ATTR_FIELD_COUNT;
+
+        ucp_dt_iov_t buf[datatype_attr.count] = {{NULL, 8}, {NULL, 4}};
+        datatype_attr.buffer = buf;
+        datatype_attr.field_mask |= UCP_DATATYPE_ATTR_FIELD_BUFFER;
+
+        status = ucp_dt_query(datatype, &datatype_attr);
+        ASSERT_EQ(UCS_OK, status);
+        EXPECT_EQ(datatype_attr.packed_size, 12);
```
yosefe, line 221: as i commented before, it should be `EXPECT_EQ(12, datatype_attr.packed_size)`
the order of parameters is important - first param is expected, second param is actual value

src/ucp/api/ucp.h
rdietric: I am not sure. 0 should basically signal that this value is not set (invalid). 0 would also enable zero-initialization of all fields after `packed_size`.
Since you bring this up, what do we expect the `packed_size` to be for a contiguous datatype, when `count` >1 is passed? count * _sizeOfDatatype_?

src/ucp/api/ucp.h
rdietric: Ok, I'll remove the comment. It is certainly implicit for the `buffer` parameter. As mentioned in https://github.com/openucx/ucx/pull/8150#discussion_r951315768 I don't think it is implicit for `count`.

src/ucp/api/ucp.h
yosefe: IMO the default should be "1" and for contig type we should multiply it by sizeOfDatatype
@shamisp WDYT?

yosefe:
LGTM besides : 
- API doc review by @tonycurtis 
- Set default value of count to 1 and use it for contig types
- Fix CI failures

src/ucp/dt/dt.c
```diff
@@ -156,6 +156,11 @@ ucs_status_t ucp_dt_query(ucp_datatype_t datatype, ucp_datatype_attr_t *attr)
     switch (datatype & UCP_DATATYPE_CLASS_MASK) {
     case UCP_DATATYPE_CONTIG:
         attr->packed_size = ucp_contig_dt_elem_size(datatype);
+
+        if (attr->field_mask & UCP_DATATYPE_ATTR_FIELD_COUNT) {
```
yosefe, line 160: we can treat count as default 1 also for generic and iov

rdietric: Makes sense and then also fits to the API documentation.

src/ucp/dt/dt.c
```diff
@@ -147,33 +147,31 @@ ucs_status_t ucp_dt_query(ucp_datatype_t datatype, ucp_datatype_attr_t *attr)
...
+    size_t count = 1;
 
     /* Currently, the only datatype attribute to query is the packed size. */
     if (!(attr->field_mask & UCP_DATATYPE_ATTR_FIELD_PACKED_SIZE)) {
         return UCS_OK;
     }
 
+    if (attr->field_mask & UCP_DATATYPE_ATTR_FIELD_COUNT) {
+        count = attr->count;
+    }
```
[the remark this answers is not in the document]
rdietric: Ok, `UCP_ATTR_VALUE` would work with `enum ucp_datatype_attr_field`. Or should I change _attr_ to _param_ in the enum? It seems to be used almost interchangeable.

yosefe (approved):

yosefe:
@tonycurtis @shamisp can you pls review?

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,53 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
...
+        break;
+    case UCP_DATATYPE_GENERIC:
+    {
+        const size_t count        = 2;
+        datatype_attr.count       = count;
+        datatype_attr.field_mask |= UCP_DATATYPE_ATTR_FIELD_COUNT;
+
+        int32_t buf[count]  = {1, 2};
+        datatype_attr.buffer              = buf;
+        datatype_attr.field_mask         |= UCP_DATATYPE_ATTR_FIELD_BUFFER;
```
brminich, line 204: indentation

test/gtest/ucp/test_ucp_dt.cc
```diff
@@ -181,6 +181,53 @@ class test_ucp_dt_iter : public ucs::test_with_param<ucp_datatype_t> {
...
+    }
+    case UCP_DATATYPE_IOV:
+    {
+        const size_t count        = 2;
+        datatype_attr.count       = count;
+        datatype_attr.field_mask |= UCP_DATATYPE_ATTR_FIELD_COUNT;
+
+        ucp_dt_iov_t buf[count] = {{NULL, 8}, {NULL, 4}};
+        datatype_attr.buffer                  = buf;
+        datatype_attr.field_mask             |= UCP_DATATYPE_ATTR_FIELD_BUFFER;
```
brminich, line 219: indentation

brminich (approved):

brminich (dismissed):

yosefe (approved):
pls squash

rdietric:
I squashed the commits and all checks have passed.

test/gtest/ucp/test_ucp_dt.cc
rdietric: fixed

test/gtest/ucp/test_ucp_dt.cc
rdietric: fixed

yosefe:
👍 
@tonycurtis @shamisp can you pls review the API change?

src/ucp/api/ucp.h
tonycurtis: agree

src/ucp/api/ucp.h
```diff
@@ -939,20 +947,36 @@ typedef struct ucp_generic_dt_ops {
...
+     * this field defaults to @e NULL.
      */
-    size_t   packed_size;
+    const void *buffer;
+
+    /**
+     * Number of elements in @a buffer.
+     * This value is optional.
+     * If @ref UCP_DATATYPE_ATTR_FIELD_COUNT is not set in @ref field_mask, the
+     * value of this field defaults to 1.
```
tonycurtis, line 977: would 0 be a better default size for a NULL buffer?

tonycurtis (changes requested):

src/ucp/api/ucp.h
shamisp: Agree with @tonycurtis comment - 1 is odd

rdietric: We've had this topic before (https://github.com/openucx/ucx/pull/8150#discussion_r951272653). Intuitively, I would also have taken 0, but there are also arguments for 1. I do not have a strong opinion on this.

yosefe: @tonycurtis @shamisp if the default is 0, it means that if this parameter is not supplied the function will always return 0. and that is essentially as if we're making this parameter mandatory.
IMO it should be 1 by default since the neutral value w.r.t. multiplication is 1.

tonycurtis:
> On Sep 4, 2022, at 11:29 AM, Yossi Itigin ***@***.***> wrote:
> 
> 
> @yosefe commented on this pull request.
> 
> In src/ucp/api/ucp.h <https://github.com/openucx/ucx/pull/8150#discussion_r962332275>:
> 
> >       */
> -    size_t   packed_size;
> +    const void *buffer;
> +
> +    /**
> +     * Number of elements in @a buffer.
> +     * This value is optional.
> +     * If @ref UCP_DATATYPE_ATTR_FIELD_COUNT is not set in @ref field_mask, the
> +     * value of this field defaults to 1.
> @tonycurtis <https://github.com/tonycurtis> @shamisp <https://github.com/shamisp> if the default is 0, it means that if this parameter is not supplied the function will always return 0. and that is essentially as if we're making this parameter mandatory.
> IMO it should be 1 by default since the neutral value w.r.t. multiplication is 1.
> 

A NULL pointer encapsulating data of length 1 seems weird.   Maybe the semantics of this routine need to be revisited?

tony

yosefe:
> A NULL pointer encapsulating data of length 1 seems weird. Maybe the semantics of this routine need to be revisited? tony

The pointer itself is not mandatory for calculating the size. The single mandatory parameter is the datatype itself, and count=1 by default would mean to calculate the size of a single element.

yosefe:
> > A NULL pointer encapsulating data of length 1 seems weird. Maybe the semantics of this routine need to be revisited? tony
> 
> The pointer itself is not mandatory for calculating the size. The single mandatory parameter is the datatype itself, and count=1 by default would mean to calculate the size of a single element.

@shamisp @tonycurtis WDYT?

src/ucp/api/ucp.h
shamisp: `If @ref UCP_DATATYPE_ATTR_FIELD_COUNT is not set in @ref field_mask, the function executes the query for a single ucp_datatype_t` . I think this sounds better. @tonycurtis  ?

tonycurtis:
Yeah, that sounds good

src/ucp/api/ucp.h
shamisp: @yosefe , @tonycurtis approved the above and I think it sounds better while does not change the meaning

PR #8151, opened 2022-04-22
UCT/IB/UD: Drop packets if EP is p2p and disconnected

dmitrygx:
## What

Drop packets if EP is p2p and disconnected.

## Why ?

Drop packets on EP which p2p and disconnected to not answer with ACK packets to a peer.
So, a peer will detect an error due to not receiving ACKs.

## How ?

1. Add UD verbs/mlx5 `ep_create` operation to use it for creating EPs and pass initial `flags` for EP (e.g. CONNECT_TO_EP).
2. Updated `uct_ud_ep_process_rx` to drop packets if `UCT_UD_EP_FLAG_DISCONNECTED` and `UCT_UD_EP_FLAG_CONNECT_TO_EP` flags are set on EP.

### DIFF
--- a/src/uct/ib/ud/base/ud_ep.c
+++ b/src/uct/ib/ud/base/ud_ep.c
@@ -241,6 +241,8 @@ static void uct_ud_ep_purge(uct_ud_ep_t *ep, ucs_status_t status)
      * operation has to return UCS_OK */
     uct_ud_ep_reset_max_psn(ep);
     uct_ud_ep_purge_outstanding(ep);
+    uct_ud_ep_ctl_op_del(ep, ep->tx.pending.ops);
+    ucs_arbiter_group_desched(&iface->tx.pending_q, &ep->tx.pending.group);
     ep->tx.acked_psn = (uct_ud_psn_t)(ep->tx.psn - 1);
     uct_ud_ep_window_release(ep, status, 0);
     ucs_assert(ucs_queue_is_empty(&ep->tx.window));
@@ -533,9 +535,34 @@ static ucs_status_t uct_ud_ep_disconnect_from_iface(uct_ud_ep_t *ep)
     return UCS_OK;
 }
 
-static ucs_status_t uct_ud_ep_create(uct_ud_iface_t *iface, int path_index,
-                                     uct_ud_ep_conn_sn_t conn_sn,
-                                     uct_ud_ep_t **new_ep_p)
+ucs_status_t uct_ud_ep_create(const uct_ep_params_t *params, uct_ep_h *ep_p)
+{
+    uct_ud_iface_t *iface;
+    ucs_status_t status;
+
+    ucs_assert(params->field_mask & UCT_EP_PARAM_FIELD_IFACE);
+
+    if (ucs_test_all_flags(params->field_mask, UCT_EP_PARAM_FIELD_DEV_ADDR |
+                                               UCT_EP_PARAM_FIELD_IFACE_ADDR)) {
+        return uct_ud_ep_create_connected_common(params, ep_p);
+    }
+
+    iface  = ucs_derived_of(params->iface, uct_ud_iface_t);
+    status = uct_iface_invoke_ops_func(&iface->super, uct_ud_iface_ops_t,
+                                       ep_new, params, ep_p);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    uct_ud_ep_set_state(ucs_derived_of(*ep_p, uct_ud_ep_t),
+                        UCT_UD_EP_FLAG_CONNECT_TO_EP);
+    return UCS_OK;
+}
+
+static ucs_status_t uct_ud_ep_create_to_iface(uct_ud_iface_t *iface,
+                                              int path_index,
+                                              uct_ud_ep_conn_sn_t conn_sn,
+                                              uct_ud_ep_t **new_ep_p)
 {
     uct_ep_params_t params = {
         .field_mask = UCT_EP_PARAM_FIELD_IFACE |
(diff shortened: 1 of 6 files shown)


src/uct/ib/ud/base/ud_ep.c
```diff
@@ -954,14 +960,14 @@ void uct_ud_ep_process_rx(uct_ud_iface_t *iface, uct_ud_neth_t *neth, unsigned b
         /* must be connection request packet */
         uct_ud_ep_rx_creq(iface, neth);
         goto out;
-    } else if (ucs_unlikely(!ucs_ptr_array_lookup(&iface->eps, dest_id, ep) ||
-                            (ep->ep_id != dest_id)))
```
yosefe, line 958: this is not needed really, because if we found it in ptr_array it means dest_id is same?
maybe assert it?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -139,16 +139,14 @@ static void uct_ud_ep_reset(uct_ud_ep_t *ep)
...
+    ucs_time_t diff;
 
     diff = ucs_twheel_get_time(&iface->tx.timer) - ep->close_time;
     if (diff > iface->config.linger_timeout) {
         ucs_debug("ud_ep %p is destroyed after %fs with timeout %fs\n",
                   ep, ucs_time_to_sec(diff),
                   ucs_time_to_sec(iface->config.linger_timeout));
-        ops = ucs_derived_of(iface->super.ops, uct_ud_iface_ops_t);
-        ops->ep_free(&ep->super.super);
+        uct_ud_ep_free(iface, ep);
```
yosefe, line 149: it's cleanup - in another PR

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -546,7 +545,8 @@ static ucs_status_t uct_ud_ep_create(uct_ud_iface_t *iface, int path_index,
     uct_ep_h new_ep_h;
     ucs_status_t status;
 
-    status = uct_ep_create(&params, &new_ep_h);
+    status = uct_iface_invoke_ops_func(&iface->super, uct_ud_iface_ops_t,
+                                       ep_create, &params, 0, &new_ep_h);
```
yosefe, line 549: this change, and passing flags, seems not needed: we could check params and set CONNECT_TO_EP flag according to params content, after creating the ep

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -397,6 +395,7 @@ UCS_CLASS_INIT_FUNC(uct_ud_ep_t, uct_ud_iface_t *iface,
     ucs_wtimer_init(&self->timer, uct_ud_ep_timer);
     ucs_arbiter_group_init(&self->tx.pending.group);
     ucs_arbiter_elem_init(&self->tx.pending.elem);
+    uct_ud_ep_set_state(self, flags);
```
yosefe, line 398: check here:
if (!(params->flags & IFACE_ADDR)) {
 set_state(CONNECT_TO_EP);
}

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -954,14 +960,14 @@ void uct_ud_ep_process_rx(uct_ud_iface_t *iface, uct_ud_neth_t *neth, unsigned b
...
-         */
-        ucs_trace("RX: failed to find ep %d, dropping packet", dest_id);
+    } else if (ucs_unlikely(!ucs_ptr_array_lookup(&iface->eps, dest_id, ep))) {
+        ucs_trace("iface=%p: dropping packet dest_id=%u", iface, dest_id);
+        goto out;
+    } else if (ucs_unlikely(ucs_test_all_flags(
+                                    ep->flags,
+                                    UCT_UD_EP_FLAG_DISCONNECTED |
+                                    UCT_UD_EP_FLAG_CONNECT_TO_EP))) {
+        ucs_trace("ep=%p: dropping packet, ep was disconencted", ep);
```
yosefe, line 970: `"ep=%p: dropping packet, point-to-point ep was disconencted", ep`

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -954,18 +958,19 @@ void uct_ud_ep_process_rx(uct_ud_iface_t *iface, uct_ud_neth_t *neth, unsigned b
...
-    } else if (ucs_unlikely(!ucs_ptr_array_lookup(&iface->eps, dest_id, ep) ||
-                            (ep->ep_id != dest_id)))
-    {
-        /* Drop the packet because it is
-         * allowed to do disconnect without flush/barrier. So it
-         * is possible to get packet for the ep that has been destroyed
-         */
-        ucs_trace("RX: failed to find ep %d, dropping packet", dest_id);
+    } else if (ucs_unlikely(!ucs_ptr_array_lookup(&iface->eps, dest_id, ep))) {
+        ucs_trace("iface=%p: dropping packet dest_id=%u", iface, dest_id);
```
yosefe, line 962: minor: "iface %p: dropping packet with dest_id %u"

dmitrygx: fixed

yosefe (approved):

yosefe (dismissed):

test/gtest/uct/ib/test_ud.cc
```diff
@@ -277,6 +278,7 @@ UCS_TEST_SKIP_COND_P(test_ud, flush_ep,
 UCS_TEST_SKIP_COND_P(test_ud, flush_iface,
                      !check_caps(UCT_IFACE_FLAG_AM_SHORT)) {
     connect();
+
```
yosefe, line 281: pls revert the unrelated whitespace changes

dmitrygx:
@yosefe could you review pls?

yosefe (dismissed):

yosefe (approved):

PR #8154, opened 2022-04-22
UCT/IB/UD: Drop packets if EP is p2p and disconnected [v1.13.x]

dmitrygx:
## What

Drop packets if EP is p2p and disconnected.

## Why ?

Backportinf of #8151.

## How ?

See #8151's description.


yosefe (approved):

yosefe (approved):

yosefe:
not yet, will merge when possible

PR #8155, opened 2022-04-22
UCT/IB: correct reclaim resources at failure cases

changchengx:
In case of failing to create QP to avoid segment fault

Signed-off-by: Changcheng Liu <jerrliu@nvidia.com>

## What
check QP existence before destroying it.

## Why ?
It's possible that it failed to create QP and there's no need to destory the un-existed QP.

## How ?
Verify whethe the QP is created or not before destroying it.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.c
+++ b/src/uct/ib/dc/dc_mlx5.c
@@ -339,11 +339,12 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
         goto init_qp;
     }
 
-    status = uct_ib_mlx5_iface_fill_attr(ib_iface, &dci->txwq.super, &attr);
+    status = uct_ib_mlx5_iface_get_res_domain(ib_iface, &dci->txwq.super);
     if (status != UCS_OK) {
         return status;
     }
 
+    uct_ib_mlx5_iface_fill_attr(ib_iface, &dci->txwq.super, &attr);
     uct_ib_iface_fill_attr(ib_iface, &attr.super);
     attr.super.ibv.cap.max_recv_sge     = 0;
 
@@ -357,7 +358,8 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
     if (qp == NULL) {
         ucs_error("mlx5dv_create_qp("UCT_IB_IFACE_FMT", DCI): failed: %m",
                   UCT_IB_IFACE_ARG(ib_iface));
-        return UCS_ERR_IO_ERROR;
+        status = UCS_ERR_IO_ERROR;
+        goto err_put_res_domain;
     }
 
     dci->txwq.super.verbs.qp = qp;
@@ -411,6 +413,12 @@ err:
     uct_rc_txqp_cleanup(&iface->super.super, &dci->txqp);
 err_qp:
     uct_ib_mlx5_destroy_qp(md, &dci->txwq.super);
+#if HAVE_DC_DV
+err_put_res_domain:
+    if (!(md->flags & UCT_IB_MLX5_MD_FLAG_DEVX_DCI)) {
+        uct_ib_mlx5_iface_put_res_domain(&dci->txwq.super);
+    }
+#endif
     return status;
 }
 
--- a/src/uct/ib/mlx5/ib_mlx5.c
+++ b/src/uct/ib/mlx5/ib_mlx5.c
@@ -298,18 +298,25 @@ ucs_status_t uct_ib_mlx5_iface_create_qp(uct_ib_iface_t *iface,
 {
     ucs_status_t status;
 
-    status = uct_ib_mlx5_iface_fill_attr(iface, qp, attr);
+    status = uct_ib_mlx5_iface_get_res_domain(iface, qp);
     if (status != UCS_OK) {
-        return status;
+        goto err;
     }
+
+    uct_ib_mlx5_iface_fill_attr(iface, qp, attr);
 
     status = uct_ib_iface_create_qp(iface, &attr->super, &qp->verbs.qp);
     if (status != UCS_OK) {
-        return status;
+        goto err_put_res_domain;
     }
 
     qp->qp_num = qp->verbs.qp->qp_num;
     return UCS_OK;
+
+err_put_res_domain:
+    uct_ib_mlx5_iface_put_res_domain(qp);
+err:
+    return status;
 }
 
 #if !HAVE_DEVX
--- a/src/uct/ib/ud/accel/ud_mlx5.c
+++ b/src/uct/ib/ud/accel/ud_mlx5.c
@@ -726,14 +726,14 @@ static ucs_status_t uct_ud_mlx5_iface_create_qp(uct_ib_iface_t *ib_iface,
 
     status = uct_ib_mlx5_iface_create_qp(ib_iface, qp, &attr);
     if (status != UCS_OK) {
-        goto err_destroy_qp;
+        return status;
     }
 
     status = uct_ib_mlx5_txwq_init(iface->super.super.super.worker,
                                    iface->tx.mmio_mode, &iface->tx.wq,
                                    qp->verbs.qp);
     if (status != UCS_OK) {
-        return status;
+        goto err_destroy_qp;
     }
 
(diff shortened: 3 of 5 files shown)


changchengx:
There's below error when running io_demo in some cases in code ```openux/ucx/master/f02a47d0``
```
#[swx-ucx05:54105:0:54105] Caught signal 11 (Segmentation fault: address not mapped to object at address (nil))
==== backtrace (tid:  54105) ====
 0 0x0000000000017760 ibv_modify_qp()  ???:0
 1 0x0000000000027365 uct_ib_destroy_qp()  /.autodirect/swgwork/jerrliu/ucx/src/uct/ib/base/ib_verbs.h:296
 2 0x0000000000027365 uct_ib_mlx5_destroy_qp()  /.autodirect/swgwork/jerrliu/ucx/src/uct/ib/mlx5/ib_mlx5.c:973
 3 0x000000000005f71b uct_ud_mlx5_iface_create_qp()  /.autodirect/swgwork/jerrliu/ucx/src/uct/ib/ud/accel/ud_mlx5.c:747
 4 0x0000000000055195 uct_ud_iface_create_qp()  /.autodirect/swgwork/jerrliu/ucx/src/uct/ib/ud/base/ud_iface.c:200
 5 0x000000000005584d uct_ud_iface_t_init()  /.autodirect/swgwork/jerrliu/ucx/src/uct/ib/ud/base/ud_iface.c:550
 6 0x0000000000063018 uct_ud_mlx5_iface_t_init()  /.autodirect/swgwork/jerrliu/ucx/src/uct/ib/ud/accel/ud_mlx5.c:849
 7 0x0000000000063018 uct_ud_mlx5_iface_t_new()  /.autodirect/swgwork/jerrliu/ucx/src/uct/ib/ud/accel/ud_mlx5.c:910
 8 0x0000000000012758 uct_iface_open()  /.autodirect/swgwork/jerrliu/ucx/src/uct/base/uct_md.c:271
 9 0x000000000003ed60 ucp_worker_iface_open()  /.autodirect/swgwork/jerrliu/ucx/src/ucp/core/ucp_worker.c:1281
10 0x000000000003fdae ucp_worker_add_resource_ifaces()  /.autodirect/swgwork/jerrliu/ucx/src/ucp/core/ucp_worker.c:1079
11 0x00000000000419dd ucp_worker_create()  /.autodirect/swgwork/jerrliu/ucx/src/ucp/core/ucp_worker.c:2358
12 0x0000000000406d47 UcxContext::init()  /.autodirect/swgwork/jerrliu/ucx/test/apps/iodemo/ucx_wrapper.cc:171
13 0x0000000000404a68 do_client()  /.autodirect/swgwork/jerrliu/ucx/test/apps/iodemo/io_demo.cc:2919
14 0x0000000000404a68 main()  /.autodirect/swgwork/jerrliu/ucx/test/apps/iodemo/io_demo.cc:2966
```

src/uct/ib/ud/accel/ud_mlx5.c
```diff
@@ -744,7 +744,9 @@ static ucs_status_t uct_ud_mlx5_iface_create_qp(uct_ib_iface_t *ib_iface,
     return status;
 
 err_destroy_qp:
-    uct_ib_mlx5_destroy_qp(ib_md, qp);
+    if (qp->verbs.qp) {
```
dmitrygx, line 747: do not go to `err_destroy_qp` in case of `uct_ib_mlx5_iface_create_qp` failed?
and we should go here if `uct_ib_mlx5_txwq_init` failed.

yosefe: that's right, seems wrong goto

changchengx: @dmitrygx 
Besides correcting the ```wrong goto```,
should it release ```resource domain``` when failing at [uct_ib_iface_create_qp](https://github.com/openucx/ucx/blob/v1.12.1/src/uct/ib/mlx5/ib_mlx5.c#L304)? the resource domain is got at calling [uct_ib_mlx5_iface_fill_attr](https://github.com/openucx/ucx/blob/v1.12.1/src/uct/ib/mlx5/ib_mlx5.c#L296)

yosefe: if it's created inside uct_ib_mlx5_iface_create_qp, it should already be released inside uct_ib_mlx5_iface_create_qp

changchengx: @yosefe 
The current code doesn't release it https://github.com/openucx/ucx/blob/v1.12.1/src/uct/ib/mlx5/ib_mlx5.c#L290-L309
I also checked other places which use ```uct_ib_mlx5_iface_fill_attr``` to get ```resource domain```, they also doesn't release the ```resource domain``` at failure case.
1. https://github.com/openucx/ucx/blob/v1.12.1/src/uct/ib/dc/dc_mlx5.c#L330
2. https://github.com/openucx/ucx/blob/v1.12.1/src/uct/ib/rc/accel/rc_mlx5_iface.c#L313

yosefe: seems like a bug, @Artemy-Mellanox do you agree?

Artemy-Mellanox: Absolutely. Missing uct_ib_mlx5_iface_put_res_domain() call in failure flow in all three cases.

changchengx: I'll add them in this PR.

Artemy-Mellanox (dismissed):

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -409,6 +410,12 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
     uct_rc_txqp_cleanup(&iface->super.super, &dci->txqp);
 err_qp:
     uct_ib_mlx5_destroy_qp(md, &dci->txwq.super);
+#if HAVE_DC_DV
```
yosefe, line 413: I think better remove uct_ib_mlx5_iface_get_res_domain from uct_ib_mlx5_iface_fill_attr and call it separately. also, make uct_ib_mlx5_iface_fill_attr return void.
reason: all "fill_attr" functions should not have side effects. it's better the caller function would use uct_ib_mlx5_iface_get_res_domain explicitly so it would be clear what needs to be cleaned-up in case of error.

changchengx: @yosefe I've refined the implementation of ```uct_ib_mlx5_iface_fill_attr``` and release ```resource domain``` in the proper way. PR has passed CI test cases.

changchengx: @yosefe @Artemy-Mellanox 
What's the next step for this PR?

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -298,18 +298,24 @@ ucs_status_t uct_ib_mlx5_iface_create_qp(uct_ib_iface_t *iface,
...
-        return status;
+        goto err_rd;
     }
 
     qp->qp_num = qp->verbs.qp->qp_num;
     return UCS_OK;
+
+err_rd:
+    uct_ib_mlx5_iface_put_res_domain(qp);
+    return status;
```
yosefe, line 318:
```
err_put_res_domain:
    uct_ib_mlx5_iface_put_res_domain(qp);
err:
    return status;
```
and "goto err" in line 303

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -411,6 +413,12 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
...
+err_rd:
```
yosefe, line 417: err_put_res_domain:

yosefe (commented):
@Artemy-Mellanox can you pls review as well?
do you think (as part of mlx5 cleanup) we can get rid of HAVE_DC_DV and HAVE_DECL_MLX5DV_CREATE_QP macros (and maybe others), and define that mlx5 support already requires such basic capabilities?

yosefe (approved):

changchengx:
@yosefe This squashed PR has passed CI tests.

PR #8160, opened 2022-04-25
DT/MEMPACK: changed error flow

hoopoepg:
- due to no error processing in mem_pack/mem_unpack functions
  ucs_error macro is replaced by ucs_fatal

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -587,8 +587,7 @@ ucs_status_t ucp_mem_type_reg_buffers(ucp_worker_h worker, void *remote_addr,
     cmpt   = context->tl_cmpts[tl_md->cmpt_index].cmpt;
 
     status = ucp_mem_rereg_mds(context, UCS_BIT(md_index), remote_addr, length,
-                               UCT_MD_MEM_ACCESS_ALL |
-                               UCT_MD_MEM_FLAG_HIDE_ERRORS,
+                               UCT_MD_MEM_ACCESS_ALL,
                                NULL, mem_type, NULL, memh, md_map);
     if (status != UCS_OK) {
         goto out;
--- a/src/ucp/dt/dt.c
+++ b/src/ucp/dt/dt.c
@@ -25,10 +25,10 @@ const char * ucp_datatype_class_names[] = {
     [UCP_DATATYPE_GENERIC]  = "generic"
 };
 
-UCS_PROFILE_FUNC(ucs_status_t, ucp_mem_type_unpack,
-                 (worker, buffer, recv_data, recv_length, mem_type),
-                 ucp_worker_h worker, void *buffer, const void *recv_data,
-                 size_t recv_length, ucs_memory_type_t mem_type)
+UCS_PROFILE_FUNC_VOID(ucp_mem_type_unpack,
+                      (worker, buffer, recv_data, recv_length, mem_type),
+                      ucp_worker_h worker, void *buffer, const void *recv_data,
+                      size_t recv_length, ucs_memory_type_t mem_type)
 {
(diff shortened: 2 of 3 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -594,6 +594,11 @@ ucs_status_t ucp_mem_type_reg_buffers(ucp_worker_h worker, void *remote_addr,
         goto out;
     }
 
+    if (*md_map == 0) {
+        status = UCS_ERR_NO_MEMORY;
```
yosefe, line 598: maybe we should not pass HIDE_ERRORS flag? at least in the context of dt_pack/unpack?

src/ucp/dt/dt.c
```diff
@@ -49,26 +49,24 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_mem_type_unpack,
...
+        ucs_fatal("failed to register buffer with mem type domain %s",
                   ucs_memory_type_names[mem_type]);
-        return status;
     }
 
     status = uct_ep_put_short(ep->uct_eps[lane], recv_data, recv_length,
                               (uint64_t)buffer, rkey_bundle.rkey);
     if (status != UCS_OK) {
-        ucs_error("uct_ep_put_short() failed %s", ucs_status_string(status));
+        ucs_fatal("uct_ep_put_short() failed %s", ucs_status_string(status));
```
yosefe, line 59: need to print it's related to mem type unpack

src/ucp/dt/dt.c
```diff
@@ -88,20 +86,18 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_mem_type_pack,
...
+        ucs_fatal("failed to register buffer with mem type domain %s",
                   ucs_memory_type_names[mem_type]);
-        return status;
     }
 
     status = uct_ep_get_short(ep->uct_eps[lane], dest, length,
                               (uint64_t)src, rkey_bundle.rkey);
     if (status != UCS_OK) {
-        ucs_error("uct_ep_get_short() failed %s", ucs_status_string(status));
+        ucs_fatal("uct_ep_get_short() failed %s", ucs_status_string(status));
```
yosefe, line 96: need to print it's related to mem type pack

src/ucp/core/ucp_mm.c
hoopoepg: yep, will work, and then ucp_mem_type_reg_buffers may be left unchanged

src/ucp/dt/dt.c
hoopoepg: as I can see ucs_fatal macro adds `__FILE__`,` __LINE__` and `__FUNCTION__` to output. do we need additional note about mem type unpack?

src/ucp/dt/dt.c
hoopoepg: same as for unpack - do we need additional note to built-in macro file/line/function?

src/ucp/dt/dt.c
yosefe: yes, let's make it even more clear

hoopoepg: ok, updated

yosefe:
cc @Sergei-Lebedev

yosefe:
@hoopoepg pls squash

yosefe (approved):

hoopoepg:
@yosefe are we need this in 1.13?

yosefe:
> @yosefe are we need this in 1.13?

@Sergei-Lebedev ?

Sergei-Lebedev:
> > @yosefe are we need this in 1.13?
> 
> @Sergei-Lebedev ?

not really, it just useful for debugging, thanks for the fix!

PR #8161, opened 2022-04-25
UCT/EP_FLUSH: added flag for remote complete

hoopoepg:
## What
Extend uct_ep_flush API to allow remote completion

## Why ?
For implementing multi-rail RMA fence functionality

## How ?
- added flag UCT_FLUSH_FLAG_REMOTE to notify UCT to update
  remote buffers

### DIFF
--- a/src/uct/api/uct.h
+++ b/src/uct/api/uct.h
@@ -503,8 +503,8 @@ enum uct_flush_flags {
     UCT_FLUSH_FLAG_LOCAL    = 0,            /**< Guarantees that the data
                                                  transfer is completed but the
                                                  target buffer may not be
-                                                 updated yet.*/
-    UCT_FLUSH_FLAG_CANCEL   = UCS_BIT(0)    /**< The library will make a best
+                                                 updated yet. */
+    UCT_FLUSH_FLAG_CANCEL   = UCS_BIT(0),   /**< The library will make a best
                                                  effort attempt to cancel all
                                                  uncompleted operations.
                                                  However, there is a chance that
@@ -519,6 +519,15 @@ enum uct_flush_flags {
                                                  error state, and it becomes
                                                  unusable for send operations
                                                  and should be destroyed. */
+    UCT_FLUSH_FLAG_REMOTE   = UCS_BIT(1)    /**< Guarantees that all previous
+                                                 UCP memory update operations
+                                                 (put, atomics, etc.) are
+                                                 completed, the target memory
+                                                 of these operation was updated,
+                                                 and the updated memory is
+                                                 globally visible for all
+                                                 processing elements in the
+                                                 system. */
 };
 
 


yosefe:
cc @manjugv

src/uct/api/uct.h
```diff
@@ -519,6 +519,10 @@ enum uct_flush_flags {
                                                  error state, and it becomes
                                                  unusable for send operations
                                                  and should be destroyed. */
+    UCT_FLUSH_FLAG_REMOTE   = UCS_BIT(1)    /**< Guarantees that the target
+                                                 buffer is updated. Effective
+                                                 for RDMA write operations
```
yosefe, line 524: 1. What about atomic write?
2. The API flag should be documented in UCT terms (not RDMA)
3. Will it be supported on all transports? if not, is there cap flag to query?
 For example, UCT_FLUSH_FLAG_CANCEL requires PEER_FAILURE capability.

hoopoepg: > * What about atomic write?

added note about atomics

> * The API flag should be documented in UCT terms (not RDMA)

rephrased, RDMA write here is just example to better explanation of flag purpose

> * Will it be supported on all transports? if not, is there cap flag to query?
>   For example, UCT_FLUSH_FLAG_CANCEL requires PEER_FAILURE capability.

this flag make any effect when remote write operation was scheduled on EP, else it is just ignored (nothing to remote complete). using this logic flag should be ignore for transports where remote completion is not actual, no need for additional capabilities

yosefe: will it be implemented for TCP and put_zcopy?

src/uct/api/uct.h
```diff
@@ -519,6 +519,11 @@ enum uct_flush_flags {
                                                  error state, and it becomes
                                                  unusable for send operations
                                                  and should be destroyed. */
+    UCT_FLUSH_FLAG_REMOTE   = UCS_BIT(1)    /**< Guarantees that the target
+                                                 buffer is updated. Effective
+                                                 for remote write operations
+                                                 such as RDMA write or
```
yosefe, line 525: Effective for remote memory write and remote atomic operations

src/uct/api/uct.h
hoopoepg: TCP may just ignore this flag because transport itself guarantee delivery, not sure about put_zcopy - is it SW emulation over TCP or RDMA?

src/uct/api/uct.h
hoopoepg: updated

yosefe: per UCX conf call: "flush to point of visibility"

yosefe:
@shamisp @tonycurtis can you pls take a look?

src/uct/api/uct.h
```diff
@@ -519,6 +519,10 @@ enum uct_flush_flags {
                                                  error state, and it becomes
                                                  unusable for send operations
                                                  and should be destroyed. */
+    UCT_FLUSH_FLAG_REMOTE   = UCS_BIT(1)    /**< Guarantees that the target
```
shamisp, line 522: I think something along the lines: `Guarantees that all previous UCP memory update operations (put and  atomics) are completed - the target memory was updated  and updated memory is visible for all processing elements in the system`. @tonycurtis Any thoughts ?

hoopoepg: @tonycurtis any comments?

shamisp: Slightly updating it - `Guarantees that all previous UCP memory update operations (put, atomics, etc.) are completed, the target memory of these operation was updated, and the updated memory is visible for all processing elements in the system.`

shamisp: @brminich It is actually very generic term. Arm specification, x86, and many other use the term to describe compute elements in the system. We cannot use "endpoints" because the intent here to make the memory visible outside of scope of UCX library. Anyways, I'm open for any other suggestions.

shamisp: @tonycurtis , can you please review ?

tonycurtis: Sorry, gtihub invalidated my email address until I did something, for some reason.

You could use "ranks" instead of PEs and I think everyone would understand what is meant (Cray/HPE still do this conflation).  Pasha's orig comment looks ok to me

shamisp: We don't have ranks definition anywhere.  UCX is used a lot outside of MPI/SHMEM so I would avoid using ranks.

yosefe: How about
```
Guarantees that all previous remote memory update operations on the endpoint @a ep (put and atomics) 
are completed and the updated memory contents are visible to the remote process the endpoint is connected to.
```

shamisp: @yosefe  the way I read `the remote process the endpoint is connected` that it is visible for other EP on the network, which is relatively weak guarantee, since does not require flush to the memory. 
I was about to suggest the following:
`Guarantees that all previous remote memory update operations on the endpoint @a ep (put and atomics) 
are completed and the updated memory contents are visible to all processes (remote and local) on the target @a ep.`
But `process` is a software construct and for example the update may not be visible if in a memory if it is not shared with `all` processes. `Processing elements` is a hardware construct which actually means that any CPU core or and accelerator that has connection to memory bus (like GPU) can actually access the memory. Sorry for been stickler on this.

yosefe: I'm a bit confused by "all processes (remote and local) on the target ep". Maybe "... and the updated memory contents are globally visible" which implies any hardware that can access the memory?

shamisp: `global visible for all processing elements in the system` ?

yosefe: sounds good to me

shamisp: @hoopoepg I think we have a consensus.

yosefe: @hoopoepg pls update according to the latest comments and let's also get Tony's approval

hoopoepg: done

hoopoepg:
@tonycurtis could you review, please
thank you

tonycurtis (changes requested):

src/uct/api/uct.h
```diff
@@ -519,6 +519,15 @@ enum uct_flush_flags {
                                                  error state, and it becomes
                                                  unusable for send operations
                                                  and should be destroyed. */
+    UCT_FLUSH_FLAG_REMOTE   = UCS_BIT(1)    /**< Guarantees that all previous
+                                                 UCP memory update operations
+                                                 (put, atomics, etc.) are
+                                                 completed, the target memory
+                                                 of these operation was updated,
+                                                 and the updated memory is
+                                                 global visible for all
```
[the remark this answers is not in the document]
hoopoepg: fixed
thank you

hoopoepg:
@yosefe ok to merge?

yosefe:
@shamisp @tonycurtis can you pls take a look?

shamisp (approved):
Look good. thanks

PR #8162, opened 2022-04-26
UCP/AM: Support reuse/free of the provided user header after ucp_am_s…

ofirfarjun7:
API changes description

## What
Support non persistence user header buffer during sending AM.

## Why ?
Current implementation dictate that "header must be valid until the operation completes", 
the new API enhancement free the user from this obligation, he can reuse/free the header after 
ucp_am_send_nbx returns.

## How ?
Copy the provided user header to UCX internal buffer when it's needed (bcopy/RNDV) until operation completes.


src/ucp/api/ucp.h
```diff
@@ -591,6 +591,13 @@ enum ucp_send_am_flags {
     UCP_AM_SEND_FLAG_RNDV  = UCS_BIT(2),             /**< Force UCP to use only
                                                           rendezvous protocol for
                                                           AM sends. */
+    UCP_AM_SEND_FLAG_COPY_HEADER  = UCS_BIT(3),     /**< Valid user header during send
```
gleon99, line 594: Please align `=` and comments.

src/ucp/api/ucp.h
```diff
@@ -591,6 +591,13 @@ enum ucp_send_am_flags {
     UCP_AM_SEND_FLAG_RNDV  = UCS_BIT(2),             /**< Force UCP to use only
                                                           rendezvous protocol for
                                                           AM sends. */
+    UCP_AM_SEND_FLAG_COPY_HEADER  = UCS_BIT(3),     /**< Valid user header during send 
+                                                         operation is not guaranteed, 
+                                                         Signal UCP to copy it when needed*/
+    UCP_AM_SEND_PRIV_FLAG_FREE_HEADER  = UCS_BIT(4),     /**< Internal UCP Flag,
```
gleon99, line 597: Use another enum for `PRIV` flags (see `UCP_AM_CB_PRIV_FIRST_FLAG` for example) - these should not be exposed to the user.

src/ucp/core/ucp_am.c
```diff
@@ -359,6 +359,11 @@ ucp_am_fill_first_footer(ucp_am_first_ftr_t *ftr, ucp_request_t *req)
     ftr->total_size = req->send.length;
 }
 
+//Release user header buf back to UCP mpool
```
gleon99, line 362: This file changes are not part of the API request.
Specifically, the "API PR" usually contains only headers under `uc*/api/`

src/ucp/core/ucp_request.c
```diff
@@ -289,6 +289,11 @@ ucs_mpool_ops_t ucp_rndv_get_mpool_ops = {
     .obj_str       = NULL
 };
 
+//Copy user header to UCP mpool buf
```
gleon99, line 292: This file changes are not part of the API request.

src/ucp/proto/proto_am.c
```diff
@@ -70,7 +70,8 @@ ucp_do_am_single(uct_pending_req_t *self, uint8_t am_id,
         ucs_assertv((packed_len >= 0) && (packed_len <= max_packed_size),
                     "packed_len=%zd max_packed_size=%zu", packed_len,
                     max_packed_size);
-
+        //turn off copy header flag - not needed in short
```
gleon99, line 73: This file changes are not part of the API request.

src/ucp/core/ucp_am.h
```diff
@@ -28,6 +28,14 @@ enum {
 };
 
 
+enum {
```
gleon99, line 31: That's not part of the API

src/ucp/core/ucp_am.c
```diff
@@ -971,6 +971,13 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
     UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);
 
     flags     = ucp_request_param_flags(param);
+    if (flags >= UCP_AM_SEND_PRIV_FLAG_FIRST_FLAG) {
```
gleon99, line 974: That's not part of the API

src/ucp/api/ucp.h
```diff
@@ -591,6 +591,9 @@ enum ucp_send_am_flags {
     UCP_AM_SEND_FLAG_RNDV  = UCS_BIT(2),             /**< Force UCP to use only
                                                           rendezvous protocol for
                                                           AM sends. */
+    UCP_AM_SEND_FLAG_COPY_HEADER  = UCS_BIT(3),      /**< Valid user header during send
```
gleon99, line 594: Please align `=` on column

gleon99: I think `COPY` is not a good word here, because UCX doesn't really always copy it.
Maybe something like: `TEMP_USER_HEADER`, or something else. WDYT?

ofirfarjun7: You are right but the user don't know it... 
It's maybe make us look "bad" from the user perspective, and this is a good reason to pick a better name if this is what you meant

ofirfarjun7: But maybe this is a good reason to find a better name, we always copy it so it's not really a good name...
It's not describe the functionality correctly.

src/ucp/api/ucp.h
```diff
@@ -582,16 +582,19 @@ enum ucp_am_cb_flags {
...
+    UCP_AM_SEND_FLAG_REPLY       = UCS_BIT(0),             /**< Force relevant reply
+                                                                endpoint to be passed to
+                                                                the data callback on the
+                                                                receiver. */
+    UCP_AM_SEND_FLAG_EAGER       = UCS_BIT(1),             /**< Force UCP to use only
+                                                                eager protocol for AM sends. */
+    UCP_AM_SEND_FLAG_RNDV        = UCS_BIT(2),             /**< Force UCP to use only
+                                                                rendezvous protocol for
+                                                                AM sends. */
+    UCP_AM_SEND_FLAG_COPY_HEADER = UCS_BIT(3),             /**< Valid user header during send
```
gleon99, line 594: Maybe rephrase:
```
UCX can only use the user header buffer until returning from the send function.
If it is needed afterwards, UCX must copy it.
```

src/ucp/api/ucp.h
ofirfarjun7: Maybe UCP_AM_SEND_FLAG_TEMP_HEADER
This signal UCX that this header is temporary and valid only for the duration of the send function?

src/ucp/api/ucp.h
ofirfarjun7: What do you think about this comment and new name?

gleon99: Maybe `until the send function returns` to avoid user confusion between "send func" and "send op" duration.

gleon99:
👍 
@yosefe  please look

src/ucp/api/ucp.h
```diff
@@ -582,16 +582,19 @@ enum ucp_am_cb_flags {
...
+    UCP_AM_SEND_FLAG_REPLY       = UCS_BIT(0),             /**< Force relevant reply
+                                                                endpoint to be passed to
+                                                                the data callback on the
+                                                                receiver. */
+    UCP_AM_SEND_FLAG_EAGER       = UCS_BIT(1),             /**< Force UCP to use only
+                                                                eager protocol for AM sends. */
+    UCP_AM_SEND_FLAG_RNDV        = UCS_BIT(2),             /**< Force UCP to use only
+                                                                rendezvous protocol for
+                                                                AM sends. */
+    UCP_AM_SEND_FLAG_TEMP_HEADER = UCS_BIT(3),             /**< Signal UCX that this header
```
yosefe, line 594: IMO need to find a better name for it..

ofirfarjun7: Story of my life...
Do you have any idea?

yosefe: UCP_AM_SEND_FLAG_COPY_HEADER
The user header buffer is copied-out during the active message send routine, and can be released or invalidated when the send routing returns, even if the send request is not completed yet.
@shamisp WDYT?

gleon99: IMO shouldn't put "COPY" there, Because UCX doesn't **always** copy the header (but only when needed).
Maybe the flag name should reflect that the user can release it / UCX doesn't need it, rather than misleading to think that it is "always copied".

yosefe: actually, it **is** always copied - either by UCP or UCT layer. Even when put immediately on the QP, we copy it to UCT bounce buffer.
It it wasn't copied to another location, we had to wait at least until the data is sent to the remote peer - which we don't.

shamisp: @ofirfarjun7 Do you have reference implementation that you can share ? It will help me to understand what you are trying to say here.

ofirfarjun7: Hi.
I submitted a PR of the implementation, currently it's marked as draft but you can check it (#8173).
I've also added a flowchart describing the send operation in UCP layer with the changes I added, look again (just added it) in the PR description for the link, I hope you have access to this directory.

ofirfarjun7: 👍
So I remove the link...

ofirfarjun7: Hi.
As Leonid mentioned I don't think you have access to our cloud so I removed the link.

In short - ucp_am_send_nbx can return success or in progress.
In case it's not succeed immediately it goes to the pending queue.
In this case the provided user header is still not part of the future srq, and therefore if the user release/corrupt it before progressing the workers, and retry sending the message bad things can happen...

ucp_send_nbx -> success -> user can reuse/release buffer
       |
      \/
in progress -> send request not posted to HW, user must keep header persistent until progressing the worker and complete the send operation.

We want to change it, in case the send operation didn't complete we will store the header in UCX internal buffer.
This is relevant for short/bcopy/RNDV.
In zcopy we already copy the buffer to registered UCP buffer.

shamisp: @ofirfarjun7 great job on diagram.

shamisp: UCP_AM_SEND_FLAG_COPY_HEADER ->UCP_AM_SEND_FLAG_CACHE_HEADER

The flag indicates that the header must be cached by UCP library. Once the ucp_send_am_* functions returns the header buffer can be reused or released.

ofirfarjun7: I think this name is better than all we come with until now

shamisp: BTW, I think this should be default behavior to start with. It is too late to change the behavior now so the flag is really welcome. The implementation looks surprisingly complex...

ofirfarjun7: Yes. it's surprise me too...
The difference between short and the bcopy complicated the implementation...
In short the pack CB is called before we know if we have enough resources because we use iov API.
In bcopy the pack CB is called after we know we have enough resources.

If you have any idea how to simplify it I will be happy to hear it.

src/ucp/api/ucp.h
```diff
@@ -582,16 +582,22 @@ enum ucp_am_cb_flags {
...
+    UCP_AM_SEND_FLAG_REPLY       = UCS_BIT(0),             /**< Force relevant reply
+                                                                endpoint to be passed to
+                                                                the data callback on the
+                                                                receiver. */
+    UCP_AM_SEND_FLAG_EAGER       = UCS_BIT(1),             /**< Force UCP to use only
+                                                                eager protocol for AM sends. */
+    UCP_AM_SEND_FLAG_RNDV        = UCS_BIT(2),             /**< Force UCP to use only
+                                                                rendezvous protocol for
+                                                                AM sends. */
+    UCP_AM_SEND_FLAG_COPY_HEADER = UCS_BIT(3),             /**< The user header buffer is temporary
```
yosefe, line 594: @ofirfarjun7 pls also document the lifetime of the header parameter in ucp_am_send_nbx

src/ucp/api/ucp.h
```diff
@@ -3170,6 +3170,9 @@ ucs_status_ptr_t ucp_am_send_nb(ucp_ep_h ep, uint16_t id,
  * @param [in]  header        User defined Active Message header. NULL value is
  *                            allowed if no header needed. In this case
  *                            @a header_length should be set to 0.
+ *                            By default the header is assumed to be valid 
+ *                            until the send operation completes, see also 
+ *                            @ref UCP_AM_SEND_FLAG_COPY_HEADER.
```
yosefe, line 3175:
```
By default the header is assumed to be valid 
until the send operation completes; if the flag
@ref UCP_AM_SEND_FLAG_COPY_HEADER is specified, 
the header assumed to be valid only during the call to this function.
```
@shamisp @tonycurtis WDYT?

brminich (dismissed):

src/ucp/api/ucp.h
yosefe: @tonycurtis ping on this

src/ucp/api/ucp.h
```diff
@@ -3164,6 +3170,11 @@ ucs_status_ptr_t ucp_am_send_nb(ucp_ep_h ep, uint16_t id,
  * @param [in]  header        User defined Active Message header. NULL value is
  *                            allowed if no header needed. In this case
  *                            @a header_length should be set to 0.
+ *                            By default the header must be valid until
+ *                            the active message send operation completed.
```
tonycurtis, line 3174: completes

src/ucp/api/ucp.h
```diff
@@ -3164,6 +3170,11 @@ ucs_status_ptr_t ucp_am_send_nb(ucp_ep_h ep, uint16_t id,
...
+ *                            If the flag @ref UCP_AM_SEND_FLAG_COPY_HEADER
+ *                            is specified, the header must be valid only 
+ *                            until return for ucp_am_send_nbx function call.
```
tonycurtis, line 3177: until the u_a_s_nbx function call returns

tonycurtis (changes requested):

src/ucp/api/ucp.h
```diff
@@ -582,16 +582,22 @@ enum ucp_am_cb_flags {
...
+                                                                receiver. */
+    UCP_AM_SEND_FLAG_EAGER       = UCS_BIT(1),             /**< Force UCP to use only
+                                                                eager protocol for AM sends. */
+    UCP_AM_SEND_FLAG_RNDV        = UCS_BIT(2),             /**< Force UCP to use only
+                                                                rendezvous protocol for
+                                                                AM sends. */
+    UCP_AM_SEND_FLAG_COPY_HEADER = UCS_BIT(3),             /**< The user header buffer is temporary
+                                                                and is valid only within the scope
+                                                                of this function call.
+                                                                If needed by UCX for later use,
```
shamisp, line 597: @tonycurtis, I suggest:
`The flag indicates that the UCX has to copy the header in order to access it after function call return. Once the function call returns, user car reuse or release the header.` What do you prefer.

tonycurtis: Sounds ok to me (car -> can)

gleon99: `the UCX` -> `UCX`

src/ucp/api/ucp.h
```diff
@@ -3164,6 +3168,11 @@ ucs_status_ptr_t ucp_am_send_nb(ucp_ep_h ep, uint16_t id,
  * @param [in]  header        User defined Active Message header. NULL value is
  *                            allowed if no header needed. In this case
  *                            @a header_length should be set to 0.
+ *                            By default the header must be valid until
+ *                            the active message send operation completes.
+ *                            If the flag @ref UCP_AM_SEND_FLAG_COPY_HEADER
+ *                            is specified, the header must be valid only 
+ *                            until the ucp_am_send_nbx function call returns.
```
yosefe, line 3175: `the ucp_am_send_nbx function` -> `this function`

src/ucp/api/ucp.h
```diff
@@ -582,20 +582,33 @@ enum ucp_am_cb_flags {
...
+
+    /**
+     * Force UCP to use only rendezvous protocol for AM sends.
+     */
+    UCP_AM_SEND_FLAG_RNDV        = UCS_BIT(2),
+
+    /** 
+     * The flag indicates that the header should be copied to an internal buffer
+     * in case it's needed after the send function returns. If this flag is
+     * specified, the header can be released immediately after the send
```
yosefe, line 603: minor: add trailing comma

yosefe (dismissed):

yosefe (approved):

yosefe (approved):

yosefe:
@tonycurtis @shamisp can you pls review as well?

tonycurtis (changes requested):

gleon99:
@tonycurtis  please review again.

tonycurtis (approved):

gleon99:
@ofirfarjun7  please squash

shamisp (approved):

yosefe (approved):

PR #8165, opened 2022-04-26
TEST/GTEST: Fix connect_and_send_recv in test_ucp_sockaddr

gleon99:
## What
Fix arguments passing to `client_ep_connect`.

## Why ?
`specify_src_addr` is passed instead of `ep_index`, which is a bug.


hoopoepg (approved):
good catch :)

PR #8166, opened 2022-04-26
UCT/IB/DC: Schedule EP for waiting DCI allocation even if no FC resources

dmitrygx:
## What

Schedule an endpoint for waiting DCI allocation even if no FC resources.

## Why ?

To resend `FC_HARD_REQ` packets in case of sending `FC_PURE_GRANT` packet was failed on a remote FC endpoint.

## How ?

Don't check whether an endpoint has FC resources or not when scheduling an endpoint's FC group for DCI allocation.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5_ep.h
+++ b/src/uct/ib/dc/dc_mlx5_ep.h
@@ -376,16 +376,13 @@ static inline int uct_dc_mlx5_iface_dci_ep_can_send(uct_dc_mlx5_ep_t *ep)
 }
 
 static UCS_F_ALWAYS_INLINE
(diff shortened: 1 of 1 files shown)


yosefe:
so we keep every ep that waits for grant on the arbiter? it will add overhead to progress.

yosefe:
we cant afford iterating over endpoints that most of the time don't have sending resource (fc)

dmitrygx:
> we cant afford iterating over endpoints that most of the time don't have sending resource (fc)

ok, if we have at least one DCI and it will be allocated for EP with empty FC window, it will be then progress in TX waitq, but moved back to wait DCI resources due to lack of FC resources.
I agree it is inefficient. wdyt about not allocating DCI

PR #8167, opened 2022-04-26
UCP/PROTO: Fixed max payload calculation for AM in new infra.

rakhmets:
## What
Fixed max payload calculation for AM in new infra to take into account AM user header size.

### DIFF
--- a/src/ucp/am/eager_multi.c
+++ b/src/ucp/am/eager_multi.c
@@ -77,17 +77,15 @@ static size_t ucp_am_eager_multi_bcopy_pack_args_first(void *dest, void *arg)
     ucp_am_hdr_t *hdr                    = dest;
     ucp_proto_multi_pack_ctx_t *pack_ctx = arg;
     ucp_request_t *req                   = pack_ctx->req;
-    size_t length, max_length;
+    size_t length;
 
     ucs_assertv(req->send.state.dt_iter.offset == 0, "offset %zu",
                 req->send.state.dt_iter.offset);
 
     ucp_am_fill_header(hdr, req);
 
-    max_length = ucs_min(ucp_am_send_req_total_size(req),
-                         pack_ctx->max_payload);
-    length     = ucp_am_eager_bcopy_pack_data(hdr + 1, req, max_length,
-                                              pack_ctx->next_iter);
+    length = ucp_am_eager_bcopy_pack(hdr + 1, req, pack_ctx->max_payload,
+                                     pack_ctx->next_iter);
     ucp_am_eager_fill_first_footer(UCS_PTR_BYTE_OFFSET(hdr + 1, length), req);
 
     return UCP_AM_FIRST_FRAG_META_LEN + length;
@@ -117,10 +115,12 @@ static UCS_F_ALWAYS_INLINE ucs_status_t ucp_am_eager_multi_bcopy_send_func(
         ucp_request_t *req, const ucp_proto_multi_lane_priv_t *lpriv,
         ucp_datatype_iter_t *next_iter)
 {
+    size_t user_hdr_size = req->send.msg_proto.am.header_length;
+
     return ucp_proto_eager_bcopy_multi_common_send_func(
             req, lpriv, next_iter, UCP_AM_ID_AM_FIRST,
             ucp_am_eager_multi_bcopy_pack_args_first,
-            UCP_AM_FIRST_FRAG_META_LEN, UCP_AM_ID_AM_MIDDLE,
+            UCP_AM_FIRST_FRAG_META_LEN + user_hdr_size, UCP_AM_ID_AM_MIDDLE,
             ucp_am_eager_multi_bcopy_pack_args_mid, UCP_AM_MID_FRAG_META_LEN);
 }
 
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -1144,6 +1144,13 @@ UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx_seg_size)
 
 class test_ucp_am_nbx_dts : public test_ucp_am_nbx_reply {
 public:
+    test_ucp_am_nbx_dts()
+    {
+        if (is_proto_enabled()) {
+            modify_config("PROTO_ENABLE", "y");
+        }
+    }
+
     static const uint64_t dts_bitmap = UCS_BIT(UCP_DATATYPE_CONTIG) |
                                        UCS_BIT(UCP_DATATYPE_IOV) |
                                        UCS_BIT(UCP_DATATYPE_GENERIC);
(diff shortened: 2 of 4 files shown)


yosefe:
@rakhmets which test failure is it fixing - or do we need to add gtest for it?

rakhmets:
> @rakhmets which test failure is it fixing - or do we need to add gtest for it?

test_ucp_am_nbx_dts.short_bcopy_send with enabled new protocols.

yosefe:
> test_ucp_am_nbx_dts.short_bcopy_send with enabled new protocols.

Can we enable new protocols in that gtest as part of this PR? To demonstrate test pass now?

rakhmets:
> > test_ucp_am_nbx_dts.short_bcopy_send with enabled new protocols.
> 
> Can we enable new protocols in that gtest as part of this PR? To demonstrate test pass now?

Done

yosefe:
@rakhmets the test failures seems relevant

src/ucp/am/eager_multi.c
```diff
@@ -117,11 +117,14 @@ static UCS_F_ALWAYS_INLINE ucs_status_t ucp_am_eager_multi_bcopy_send_func(
...
         ucp_datatype_iter_t *next_iter)
 {
+    size_t user_hdr_size = req->send.msg_proto.am.header_length;
+
     return ucp_proto_eager_bcopy_multi_common_send_func(
             req, lpriv, next_iter, UCP_AM_ID_AM_FIRST,
             ucp_am_eager_multi_bcopy_pack_args_first,
-            UCP_AM_FIRST_FRAG_META_LEN, UCP_AM_ID_AM_MIDDLE,
-            ucp_am_eager_multi_bcopy_pack_args_mid, UCP_AM_MID_FRAG_META_LEN);
+            UCP_AM_FIRST_FRAG_META_LEN + user_hdr_size, UCP_AM_ID_AM_MIDDLE,
```
yosefe, line 125: it seems weird that we are passing user_hdr_size to ucp_proto_eager_bcopy_multi_common_send_func, but actually this function does not really pack the header; then we add the header size inside ucp_am_eager_multi_bcopy_pack_args_first to compensate to subtracting it..
firstly, ucp_am_eager_bcopy_pack_data should take max_length without the header. maybe even rename it to ucp_am_eager_bcopy_pack

src/ucp/proto/proto_multi.inl
```diff
@@ -268,7 +267,8 @@ ucp_proto_eager_bcopy_multi_common_send_func(
         pack_cb  = pack_cb_middle;
         hdr_size = hdr_size_middle;
     }
-    pack_ctx.max_payload = ucp_proto_multi_max_payload(req, lpriv, hdr_size);
+    pack_ctx.max_payload = ucp_proto_multi_max_payload(req, lpriv, hdr_size,
```
yosefe, line 270: maybe we can just add hdr_size to req->send.state.dt_iter.length ?

src/ucp/proto/proto_multi.inl
```diff
@@ -44,9 +44,8 @@ ucp_proto_multi_scaled_length(uint32_t weight, size_t length)
 static UCS_F_ALWAYS_INLINE size_t
 ucp_proto_multi_max_payload(ucp_request_t *req,
                             const ucp_proto_multi_lane_priv_t *lpriv,
-                            size_t hdr_size)
+                            size_t hdr_size, size_t length)
```
yosefe, line 47: maybe pass hdr_size_first, hdr_size_middle?

src/ucp/proto/proto_multi.inl
rakhmets: From existing implementation perspective it is more consistent to keep hrd_size out of req->send.state.dt_iter.length. E.g. user header is sent within first send of multi fragment send.

src/ucp/proto/proto_multi.inl
rakhmets: Reverted changes in `ucp_proto_multi_max_payload`.

yosefe (dismissed):

src/ucp/am/eager_multi.c
```diff
@@ -77,17 +77,15 @@ static size_t ucp_am_eager_multi_bcopy_pack_args_first(void *dest, void *arg)
...
                 req->send.state.dt_iter.offset);
 
     ucp_am_fill_header(hdr, req);
 
-    max_length = ucs_min(ucp_am_send_req_total_size(req),
-                         pack_ctx->max_payload);
-    length     = ucp_am_eager_bcopy_pack_data(hdr + 1, req, max_length,
-                                              pack_ctx->next_iter);
+    length = ucp_am_eager_bcopy_pack(hdr + 1, req, pack_ctx->max_payload,
+                                     pack_ctx->next_iter);
```
brminich, line 88: is it because the first fragment is always supposed to consume the whole fragment?

rakhmets: We use `ucp_proto_multi_max_payload` to calculate `pack_ctx->max_payload`.
`ucp_proto_multi_max_payload` takes into account fragment size, length, and header size.
Thus, this line is redundant.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1144,6 +1144,13 @@ UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx_seg_size)
 
 class test_ucp_am_nbx_dts : public test_ucp_am_nbx_reply {
 public:
+    test_ucp_am_nbx_dts()
+    {
+        if (is_proto_enabled()) {
+            modify_config("PROTO_ENABLE", "y");
+        }
```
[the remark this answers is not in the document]
rakhmets: `test_ucp_am_nbx_seg_size` will be fixed in separate PR.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1210,9 +1223,14 @@ class test_ucp_am_nbx_dts : public test_ucp_am_nbx_reply {
...
     }
+
+    bool is_proto_enabled() const
+    {
+        return get_variant_value(5);
+    }
 };
 
-UCS_TEST_P(test_ucp_am_nbx_dts, short_send)
+UCS_TEST_P(test_ucp_am_nbx_dts, short_send, "ZCOPY_THRESH=-1", "RNDV_THRESH=-1")
```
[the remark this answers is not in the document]
rakhmets: to be sure that we use short, because `test_am` calls `test_am_send_recv` three times. the third one passes user header with maximum AM header size.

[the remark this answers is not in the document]
brminich: ok, but i think bcopy (not short) will be used for that case anyway, so this setting is not really relevant

src/ucp/am/eager_multi.c
brminich: ok

test/gtest/ucp/test_ucp_am.cc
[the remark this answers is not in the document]
rakhmets: yes, bcopy should be used. however there is an issue for shm_ib transport: rendezvous is selected instead.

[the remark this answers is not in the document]
brminich: why is it selected for such a small message?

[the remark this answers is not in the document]
rakhmets: As I mentioned, there are three sends in this test. The third one send a message with the size equals to maximum size of AM header plus 1. Thus the message size can be almost the same as rendezvous threshold.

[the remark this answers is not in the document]
brminich: ok

brminich (approved):

PR #8169, opened 2022-04-26
UCT/IB/DC: Schedule EP for resending FC_HARD_REQ

dmitrygx:
## What

Schedule EP for resending `FC_HARD_REQ` packet.

## Why ?

To resend `FC_HARD_REQ` packets in case of sending `FC_PURE_GRANT` packet was failed on a remote FC endpoint.

## How ?

1. Implement `uct_dc_mlx5_ep_fc_hard_req_progress` which goes over FC hash and schedules EP to DCI waitq or TX waiq (if DCI was allocated).
2. Register `uct_dc_mlx5_ep_fc_hard_req_progress` callback in CBQ if FC hash isn't empty.
3. Deregister `uct_dc_mlx5_ep_fc_hard_req_progress` callback if FC hash becomes empty.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5_ep.c
+++ b/src/uct/ib/dc/dc_mlx5_ep.c
@@ -1153,7 +1153,7 @@ static void uct_dc_mlx5_ep_fc_cleanup(uct_dc_mlx5_ep_t *ep)
 
     it = kh_get(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash, (uint64_t)ep);
     if (it != kh_end(&iface->tx.fc_hash)) {
-        kh_del(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash, it);
+        uct_dc_mlx5_fc_entry_iter_del(iface, it);
     }
 }
 
@@ -1471,6 +1471,28 @@ void uct_dc_mlx5_ep_pending_purge(uct_ep_h tl_ep,
     }
 }
 
+unsigned uct_dc_mlx5_ep_fc_hard_req_progress(void *arg)
+{
+    uct_dc_mlx5_iface_t *iface = arg;
+    ucs_time_t now             = ucs_get_time();
+    uint64_t ep_key;
+    uct_dc_mlx5_ep_t *ep;
+
+    if (ucs_likely(now < iface->tx.fc_hard_req_resend_time)) {
+        return 0;
+    }
+
+    /* Go over all endpoints that are waiting for FC window being restored and
+     * resend FC_HARD_REQ packet to make sure a peer will resend FC_PURE_GRANT
+     * packet in case of failure on the remote FC endpoint */
+    kh_foreach_key(&iface->tx.fc_hash, ep_key, {
+        ep = (uct_dc_mlx5_ep_t*)ep_key;
+        uct_dc_mlx5_ep_schedule(iface, ep);
+    })
+
+    return 1;
+}
+
 ucs_status_t uct_dc_mlx5_ep_check_fc(uct_dc_mlx5_iface_t *iface,
                                      uct_dc_mlx5_ep_t *ep)
 {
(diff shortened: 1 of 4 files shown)


src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1471,6 +1471,25 @@ void uct_dc_mlx5_ep_pending_purge(uct_ep_h tl_ep,
...
 }
 
+unsigned uct_dc_mlx5_ep_fc_hard_req_progress(void *arg)
+{
+    uct_dc_mlx5_iface_t *iface = arg;
+    ucs_time_t now             = ucs_get_time();
+    uint64_t ep_key;
+    uct_dc_mlx5_ep_t *ep;
+
+    if (ucs_likely(now - iface->tx.fc_hard_req_resend_time)) {
```
yosefe, line 1481: Should be less-than and not minus?

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1471,6 +1471,25 @@ void uct_dc_mlx5_ep_pending_purge(uct_ep_h tl_ep,
...
+        return 0;
+    }
+
+    kh_foreach_key(&iface->tx.fc_hash, ep_key, {
+        ep = (uct_dc_mlx5_ep_t*)(uintptr_t)ep_key;
```
yosefe, line 1486: IMO (uintptr_t) is not needed here

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1471,6 +1471,25 @@ void uct_dc_mlx5_ep_pending_purge(uct_ep_h tl_ep,
...
+        uct_dc_mlx5_ep_schedule(iface, ep);
```
yosefe, line 1487: Pls add comment to explain why we do it

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1522,6 +1541,12 @@ ucs_status_t uct_dc_mlx5_ep_check_fc(uct_dc_mlx5_iface_t *iface,
         }
 
         goto out;
+    } else {
```
yosefe, line 1544: Do we really need else here, given the goto ?

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: fixed

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: removed

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: removed

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: done

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1478,12 +1478,15 @@ unsigned uct_dc_mlx5_ep_fc_hard_req_progress(void *arg)
     uint64_t ep_key;
     uct_dc_mlx5_ep_t *ep;
 
-    if (ucs_likely(now - iface->tx.fc_hard_req_resend_time)) {
+    if (ucs_likely(now < iface->tx.fc_hard_req_resend_time)) {
         return 0;
     }
 
+    /* Go over all endpoints which are waiting for FC window being restored and
```
yosefe, line 1485: which -> that

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1478,12 +1478,15 @@ unsigned uct_dc_mlx5_ep_fc_hard_req_progress(void *arg)
...
+     * and resend FC_HARD_REQ packet to make sure a peer will resend
+     * FC_PURE_GRANT packet in case of failure */
```
yosefe, line 1487: FC_PURE_GRANT packet in case of failure on the remote FC endpoint

yosefe (commented):
can squash after fix

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: fixed

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: fixed

yosefe (approved):

dmitrygx:
@yosefe can confirm that #8169 and #8151 allow IO-demo passing successfully (30 iterations) for both RC and DC

yosefe:
@dmitrygx can you pls backport this fix to v1.13.x as well?

PR #8116, opened 2022-04-11
UCP/ENDPOINT: Implementation of API changes to query transport/endpoint pairs associated with an endpoint

drwootton:
## What
Implement a change to the API for the ucp_ep_query function so the caller can query the transport/device name pairs associated with an active endpoint. This implements the API header changes in pull request #7990.

## Why ?
OpenMPI users have requested the ability to query the transport and device name pairs used by an MPI task when the transport used by OpenMPI is UCX. This information is useful to OpenMPI users so they can verify that application tasks are actually using the transports that the user requested. 

## How ?
The ucp_ep_attr structure is modified by adding a new bitmask bit, UCP_EP_ATTR_FIELD_TRANSPORTS that the caller sets to request endpoint transport/device name pairs. The caller fills in the transports structure in ucp_ep_attrs to specify the attributes of the ucp_transport_entry_t array the caller allocated as well as setting the UCP_EP_ATTR_FIELD_TRANSPORTS bit and calls ucp_ep_query.

 The new ucp_ep_query_transport function is called by ucp_ep_query to iterate thru the transport structures to fill in the transport and device names.

The ucp_ep_query_transport function indexes the ucp_transport_entry_t array by multiplying its loop index by the size of one entry in this structure. This is done for forward/backward compatibility if the caller is using a different version of UCX with a different ucp_transport_entry_t definition.

Before this function updates a field in this structure, it ensures that the field's ending offset is not past the end of the ucp_transport_entry_t it was passed to avoid storage overlay problems.

Signed-off-by: David Wootton <dwootton@us.ibm.com>

PR #8171, opened 2022-04-27
UCT/GTEST: Fix test w/o mlx5

Artemy-Mellanox:
Fix for https://github.com/openucx/ucx/pull/8116


yosefe:
@Artemy-Mellanox how is it fixing the problem? shouldn't we fix ib_mlx5.h?
is is still needed after removing exp verbs support?

PR #8175, opened 2022-04-27
UCP/CORE: Remove assertion for checking WIREUP_MSGs when purging

dmitrygx:
## What

Remove assertion for checking `WIREUP_MSG`s when purging.

## Why ?

Only `WIREUP_MSG`/`PRE-REQ` packet is allowed to be scheduled on normal UCT EP. But no need to check it.

## How ?

Update `ucp_request_send_state_ff` to not check `WIREUP_MSG`s being scheduled on normal UCT EP.

### DIFF
--- a/src/ucp/core/ucp_request.c
+++ b/src/ucp/core/ucp_request.c
@@ -676,16 +676,6 @@ void ucp_request_send_state_ff(ucp_request_t *req, ucs_status_t status)
     if (req->send.uct.func == ucp_proto_progress_am_single) {
         req->send.proto.comp_cb(req);
     } else if (req->send.uct.func == ucp_wireup_msg_progress) {
-        /* Sending EP_REMOVED/EP_CHECK/ACK WIREUP_MSGs could be scheduled on
-         * UCT endpoint which is not a WIREUP_EP. Other WIREUP MSGs should not
-         * be returned from 'uct_ep_pending_purge()', since they are released
-         * by WIREUP endpoint's purge function
-         */
-        ucs_assertv((req->send.wireup.type == UCP_WIREUP_MSG_EP_REMOVED) ||
-                    (req->send.wireup.type == UCP_WIREUP_MSG_EP_CHECK) ||
-                    (req->send.wireup.type == UCP_WIREUP_MSG_ACK),
-                    "req %p ep %p: got %s message", req, req->send.ep,
-                    ucp_wireup_msg_str(req->send.wireup.type));
         ucs_free(req->send.buffer);
         ucp_request_mem_free(req);
     } else if (req->send.state.uct_comp.func == ucp_ep_flush_completion) {


dmitrygx:
@hoopoepg @evgeny-leksikov could you review pls?

hoopoepg (approved):

src/ucp/core/ucp_request.c
```diff
@@ -676,16 +676,7 @@ void ucp_request_send_state_ff(ucp_request_t *req, ucs_status_t status)
...
-         * UCT endpoint which is not a WIREUP_EP. Other WIREUP MSGs should not
-         * be returned from 'uct_ep_pending_purge()', since they are released
-         * by WIREUP endpoint's purge function
-         */
-        ucs_assertv((req->send.wireup.type == UCP_WIREUP_MSG_EP_REMOVED) ||
-                    (req->send.wireup.type == UCP_WIREUP_MSG_EP_CHECK) ||
-                    (req->send.wireup.type == UCP_WIREUP_MSG_ACK),
-                    "req %p ep %p: got %s message", req, req->send.ep,
-                    ucp_wireup_msg_str(req->send.wireup.type));
+        ucs_assert(req->send.wireup.type != UCP_WIREUP_MSG_PRE_REQUEST);
```
evgeny-leksikov, line 679: pls revert to `ucs_assertv`, sometime it's useful to know request or ep ptr to grep its history in log

src/ucp/wireup/wireup.c
```diff
@@ -124,6 +125,11 @@ ucs_status_t ucp_wireup_msg_progress(uct_pending_req_t *self)
 
     /* send the active message */
     req->send.lane = ucp_wireup_get_msg_lane(ep, req->send.wireup.type);
+    uct_ep         = ep->uct_eps[req->send.lane];
+
+    if (!ucp_wireup_ep_test(uct_ep)) {
+        ucs_assert(req->send.wireup.type != UCP_WIREUP_MSG_PRE_REQUEST);
+    }
```
evgeny-leksikov, line 132: maybe move to separate function and assert all allowed types there?
This condition hardly to read and might  be broken by new WIREUP_MSG type. Also probably it could be reused above.

dmitrygx: > maybe move to separate function and assert all allowed types there?
> This condition hardly to read and might be broken by new WIREUP_MSG type.

we disallow sending `PRE_REQUEST` only.
adding new packet type will broke it too, if it will be allowed to be sent.

src/ucp/core/ucp_request.c
dmitrygx: done

src/ucp/wireup/wireup.c
dmitrygx: removed this assertion, because it is not needed

dmitrygx:
@yosefe could you review pls?
I catched this assertion failed one time and we don't need this assertion (we don't have any logic which relies on the checked fact there) - so, it worth just deleting it.

yosefe (approved):

evgeny-leksikov (approved):

PR #8176, opened 2022-04-28
UCT/RDMACM: Cache ibv_query_port

Artemy-Mellanox:


src/ucs/datastruct/bitmap.h
```diff
@@ -563,6 +563,7 @@ ucs_bitmap_ffs(const ucs_bitmap_word_t *bitmap_words, size_t num_words,
 }
 
 
+_UCS_BITMAP_DECLARE_TYPE(2)
```
yosefe, line 566: do we really need type for such small bitmap? could just use uint8/32/64?

src/uct/ib/rdmacm/rdmacm_cm.c
```diff
@@ -83,11 +83,33 @@ uct_rdmacm_cm_device_context_init(uct_rdmacm_cm_device_context_t *ctx,
...
 
+    ret = ibv_query_device(verbs, &dev_attr);
+    if (ret != 0) {
+        ucs_error("ibv_query_device(%s) failed: %m", dev_name);
+        return UCS_ERR_IO_ERROR;
+    }
+
+    UCS_BITMAP_CLEAR(&ctx->is_eth);
+    for (i = 0; i < dev_attr.phys_port_cnt; ++i) {
+        ret = ibv_query_port(verbs, i + 1, &port_attr);
```
yosefe, line 102: can we define first port macro as 1, and use it also in https://github.com/openucx/ucx/blob/master/src/uct/ib/base/ib_device.c#L760

src/uct/ib/rdmacm/rdmacm_cm.h
```diff
@@ -71,12 +72,13 @@ typedef struct uct_rdmacm_cm_reserved_qpn_blk {
...
-    uint32_t        log_reserved_qpn_granularity;
-    uint32_t        num_dummy_qps;
-    struct ibv_cq   *cq;
+    int                                use_reserved_qpn;
+    ucs_spinlock_t                     lock;                         /** Avoid competed condition on the qpn resource for multi-threads */
+    ucs_list_link_t                    blk_list;
+    uint32_t                           log_reserved_qpn_granularity;
+    uint32_t                           num_dummy_qps;
+    struct ibv_cq                      *cq;
+    ucs_bitmap_t(UCT_IB_DEV_MAX_PORTS) is_eth;
```
yosefe, line 81: eth_ports

src/ucs/datastruct/bitmap.h
```diff
@@ -21,6 +21,11 @@ BEGIN_C_DECLS
 typedef uint64_t ucs_bitmap_word_t;
 
 
+typedef struct {
+    uint8_t bits[1];
+} ucs_bitmap8_t;
```
yosefe, line 26: i mean, can't we use just uint8_t in rdma_cm?

Artemy-Mellanox: I use ucs_bitmap_t to emphasize that it is a bitmap and not a number

yosefe (dismissed):
pls squash

yosefe (approved):

PR #8188, opened 2022-05-01
UCT/IB: Fix log_ack_req_freq field initialization

gleon99:
## What
Fix initialization of the `log_ack_req_freq` field in RC MLX5 ifaces.

## Why ?
There is a bug - the field appears in 2 config types and gets overridden with 0 from the second (unused) config.


src/uct/ib/rc/accel/rc_mlx5_iface.c
```diff
@@ -876,8 +876,6 @@ UCS_CLASS_INIT_FUNC(uct_rc_mlx5_iface_t,
 
     self->super.super.config.tx_moderation = ucs_min(config->super.tx_cq_moderation,
                                                      self->super.tx.bb_max / 4);
-    self->super.config.log_ack_req_freq    = ucs_min(config->super.log_ack_req_freq,
```
yosefe, line 879: do we need to move this to base class?
otherwise, seems UCT_RC_MLX5_MAX_LOG_ACK_REQ_FREQ is not used except for config help string

yosefe (dismissed):

PR #8190, opened 2022-05-01
UCT/SELF: Don't add device index if have only one device

yosefe:
## Why
Without this fix, the self transport uses "memory0" device name, but it should use "memory" to show that shared memory transports such as cma/knem/sysv/posix should not be instantiated alongside with it.


gleon99 (approved):
👍

PR #8191, opened 2022-05-02
CONTRIB/PR_MERGE: Support dismissed review

yosefe:
## Why
Pushing (or force-pushing) commits would change the state of reviews from APPROVED to DISMISSED

### DIFF
--- a/contrib/pr_merge_check.py
+++ b/contrib/pr_merge_check.py
@@ -71,14 +71,15 @@ class PRChecker(object):
         """
         Set self.approved_commit to the hash of the latest approved PR commit.
         """
+        if self.approved_commit:
+            return;
         page = 1
-        self.approved_commit = None
         while True:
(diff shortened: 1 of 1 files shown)


shamisp:
@yosefe what this will do to squashed commits ?

yosefe:
the original goal of the script was to validate squash+force push after approval did not introduce any changes; this PR also considers that force-push (including squash) would dismiss existing approved reviews.

shamisp:
Is it possible to avoid invalidation of approved reviews if `git diff` of
the squashed request didn't change compared to previous diff ?

yosefe:
> Is it possible to avoid invalidation of approved reviews if `git diff` of the squashed request didn't change compared to previous diff ?

Not that i know of.. otherwise would not need this script..

yosefe:
@shamisp is this ok to merge?

PR #8195, opened 2022-05-02
UCT/IB/rlimit_mem_lock: disabled warning message for privileged user

shasson5:
## What
Do not pop a warning message when a limit on memory lock size exists, and the user is privileged.

## Why ?
for privileged user, RLimit is not enforced, so it never fails an attempt to lock memory.


shamisp:
@shasson5 you probably want to sign this in with a more appropriate email

```
From 119ba3263624616bf9425782569c87809d8a2b12 Mon Sep 17 00:00:00 2001
From: root <root@jazz20.swx.labs.mlnx>
Date: Mon, 2 May 2022 17:09:48 +0300
Subject: [PATCH] UCT/ib/rlimit_mem_lock: disabled warning message when user is
```

src/ucs/sys/sys.c
```diff
@@ -1565,6 +1584,30 @@ ucs_status_t ucs_sys_get_memlock_rlimit(size_t *rlimit_value)
...
+        return status;
+    }
+
+    status = ucs_sys_get_memlock_privilege(&lock_priv);
+    if(status != UCS_OK) {
+        return status;
+    }
+
+    /* overwrite system wide value for priviledged users */
+    if(lock_priv) {
```
gleon99, line 1607: Can simplify the logic: first check if has CAP and return `SIZE_MAX`, otherwise continue.

src/ucs/sys/sys.c
```diff
@@ -1565,6 +1584,30 @@ ucs_status_t ucs_sys_get_memlock_rlimit(size_t *rlimit_value)
...
+    int lock_priv;
+
+    /* get system wide rlimit value */
+    ucs_status_t status = ucs_sys_get_memlock_rlimit(rlimit_value);
+    if(status != UCS_OK) {
+        return status;
+    }
+
+    status = ucs_sys_get_memlock_privilege(&lock_priv);
+    if(status != UCS_OK) {
```
gleon99, line 1602: `if (...`

gleon99, line 1602: `if (...`

src/ucs/sys/sys.c
```diff
@@ -767,23 +767,42 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
...
     uint32_t ecap;
+    ucs_status_t status;
 
     UCS_STATIC_ASSERT(CAP_IPC_LOCK < 32);
     status = ucs_sys_get_proc_cap(&ecap);
-    if ((status == UCS_OK) && !(ecap & UCS_BIT(CAP_IPC_LOCK))) {
-        /* detected missing CAP_IPC_LOCK */
-        snprintf(buf, max, ", CAP_IPC_LOCK privilege is needed for SHM_HUGETLB");
-        return;
+    if(status != UCS_OK) {
```
gleon99, line 778: In other places in UCX, it is ignored. Can simply return "false" (0) in such case and make the func signature simpler (no in-out ptr arg).

src/ucs/sys/sys.c
```diff
@@ -1565,6 +1584,30 @@ ucs_status_t ucs_sys_get_memlock_rlimit(size_t *rlimit_value)
...
+        return status;
+    }
+
+    status = ucs_sys_get_memlock_privilege(&lock_priv);
+    if(status != UCS_OK) {
+        return status;
+    }
+
+    /* overwrite system wide value for priviledged users */
```
gleon99, line 1606: `Overwrite`

`privileged`

src/ucs/sys/sys.c
gleon99: After refactoring according to this suggestion, we'll have the following pattern:
```
  status = ucs_sys_get_proc_cap(&ecap);
  if ((status == UCS_OK) && (ecap & SOME_CAP)) {
```

in 3 places in our code (your PR + 2 existing places).
Maybe make in an inline func?

gleon99, line 778: Can ignore return value (see 2 other examples)

shasson5: @gleon99 can you point the other 2 places? I only found one more

gleon99: `ucs_sysv_shmget_error_check_EPERM`, `uct_cma_test_ptrace_scope`

shasson5: notice that I already refactored `ucs_sysv_shmget_error_check_EPERM`, so that it won't call directly to `ucs_sys_get_proc_cap`.

shasson5: not sure I understand, in `ucs_sys_get_memlock_rlimit` they indeed return the status if it failed

gleon99: I meant, refactor the call to `ucs_sys_get_proc_cap` + status/cap check to a single func (see https://github.com/openucx/ucx/pull/8195/files#r867890526).

Then, instead of writing:
```
status = ucs_sys_get_proc_cap(&ecap);
if(status != UCS_OK) {
    ....
```

we can do:
```
if (!ucs_sys_has_proc_cap(CAP_IPC_LOCK)) {
  ....
```

gleon99:
Pipeline failure seems relevant.

src/ucs/sys/sys.c
```diff
@@ -776,9 +780,9 @@ static ucs_status_t ucs_sys_get_memlock_privilege(int *lock_priv)
     UCS_STATIC_ASSERT(CAP_IPC_LOCK < 32);
     status = ucs_sys_get_proc_cap(&ecap);
 
-    *lock_priv = 0;
-    if((status == UCS_OK) && (ecap & UCS_BIT(CAP_IPC_LOCK))) {
-        *lock_priv = 1;
+    *mlock_cap = 0;
```
gleon99, line 783: Use `else`

src/ucs/sys/sys.c
```diff
@@ -789,14 +793,15 @@ static ucs_status_t ucs_sys_get_memlock_privilege(int *lock_priv)
...
+    int mlock_cap = 0;
+
+    if (ucs_sys_get_mlock_cap(&mlock_cap) != UCS_OK) {
         snprintf(buf, max,
                 ", please check for CAP_IPC_LOCK privilege for using SHM_HUGETLB");
         return;
     }
 
-    if(lock_priv == 0) {
+    if (mlock_cap == 0) {
```
gleon99, line 804: `if (!mlock_cap) {`

src/ucs/sys/sys.c
shasson5: why? isn't it simpler to read without `else`?

gleon99: Without `else` you init even if it is not needed..

src/ucs/sys/sys.c
```diff
@@ -769,8 +769,11 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
 
 /**
  * Checks whether the process have memory lock capability.
+ *
  * @param [out] mlock_cap: Flag that indicates whether we have the required capability.
+ * @return UCS_OK or error in case of failure.
```
gleon99, line 774: Add a blank line between "param" and "return"

src/ucs/sys/sys.c
```diff
@@ -1565,9 +1590,29 @@ ucs_status_t ucs_sys_get_memlock_rlimit(size_t *rlimit_value)
...
                             SIZE_MAX :
                             limit_info.rlim_cur;
+
     return UCS_OK;
 }
 
+ucs_status_t ucs_sys_get_effective_memlock_rlimit(size_t *rlimit_value)
+{
+    int mlock_cap = 0;
+
```
gleon99, line 1600: Move blank line below (separate vars decl from other code).

src/ucs/sys/sys.c
```diff
@@ -1565,9 +1590,29 @@ ucs_status_t ucs_sys_get_memlock_rlimit(size_t *rlimit_value)
...
+    ucs_status_t status = ucs_sys_get_mlock_cap(&mlock_cap);
+    if (status != UCS_OK) {
+        return status;
```
gleon99, line 1603: Do we really want to return "unsupported" from `ucs_sys_get_effective_memlock_rlimit`, just because the "capability API" isn't there?

src/ucs/sys/sys.h
```diff
@@ -635,6 +635,12 @@ unsigned long ucs_sys_get_proc_create_time(pid_t pid);
  */
 ucs_status_t ucs_sys_get_memlock_rlimit(size_t *rlimit_value);
 
+/*
+ * Get effective max locked memory limit (unlimited for privileged user).
+ *
+ * @param [out] rlimit_value If successful, set to the current limit value.
```
gleon99, line 641: Add "return"

src/ucs/sys/sys.c
shasson5: `mlock_cap` is already separated. 
do you mean also separate `ucs_status_t status`?

gleon99:
```
int mlock_cap = 0;
ucs_status_t status = ucs_sys_get_mlock_cap(&mlock_cap);

if (status != UCS_OK) {
...
```

src/ucs/sys/sys.c
shasson5: well, there are 3 options I see we can consider instead:
1) replace it with some other, more informative error code.
2) fallback to checking uid = 0.
3) fallback to using raw kernel API to query caps.

I think first option is best because:
- option 2 is not exactly what this function should do.
- option 3 is not portable between kernel versions and may break in the future.

gleon99: Why not simply ignore it and take the value from `ucs_sys_get_memlock_rlimit`?

src/ucs/sys/sys.c
```diff
@@ -767,23 +767,49 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
 #endif
 }
 
-static void ucs_sysv_shmget_error_check_EPERM(int flags, char *buf, size_t max)
+/**
+ * Checks whether the process have memory lock capability.
+ *
+ * @param [out] mlock_cap: Flag that indicates whether we have the required capability.
```
gleon99, line 773: Remove `:` after `mlock_cap`

src/ucs/sys/sys.c
```diff
@@ -1565,9 +1591,26 @@ ucs_status_t ucs_sys_get_memlock_rlimit(size_t *rlimit_value)
     *rlimit_value = (limit_info.rlim_cur == RLIM_INFINITY) ?
                             SIZE_MAX :
                             limit_info.rlim_cur;
+
```
gleon99, line 1594: Why?

gleon99:
@shasson5 please address `clang-format` comments (long lines).

src/ucs/sys/sys.c
```diff
@@ -767,23 +767,51 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
 #endif
 }
 
-static void ucs_sysv_shmget_error_check_EPERM(int flags, char *buf, size_t max)
+/**
+ * Checks whether the process have memory lock capability.
+ *
+ * @param [out] mlock_cap Flag that indicates whether we have the required capability.
```
gleon99, line 773: 1. "Pointer to..."
2. There's no "we", it is a process.

src/ucs/sys/sys.c
```diff
@@ -770,7 +770,7 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
 /**
  * Checks whether the process have memory lock capability.
  *
- * @param [out] mlock_cap Flag that indicates whether we have the required capability.
+ * @param [out] mlock_cap Pointer to a flag that indicates whether the process have the required capability.
```
gleon99, line 773: Line too long

src/ucs/sys/sys.h
```diff
@@ -635,6 +635,16 @@ unsigned long ucs_sys_get_proc_create_time(pid_t pid);
  */
 ucs_status_t ucs_sys_get_memlock_rlimit(size_t *rlimit_value);
 
+/*
+ * Get effective max locked memory limit (unlimited for privileged user).
+ * In case we can't query the system for capabilities, we fallback to
+ * 'ucs_sys_get_memlock_rlimit'
```
gleon99, line 641: Use @ref for func name.

gleon99:
@yosefe  please look

src/ucs/sys/sys.c
```diff
@@ -767,23 +767,52 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
...
-static void ucs_sysv_shmget_error_check_EPERM(int flags, char *buf, size_t max)
+/**
+ * Checks whether the process have memory lock capability.
+ *
+ * @param [out] mlock_cap Pointer to a flag that indicates whether the
+ *              process have the required capability.
+ *
+ * @return UCS_OK or error in case of failure.
+ */
+
```
yosefe, line 778: remove space line

src/ucs/sys/sys.c
```diff
@@ -767,23 +767,52 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
...
+static ucs_status_t ucs_sys_get_mlock_cap(int *mlock_cap)
```
yosefe, line 779: why not return 0/1 ?
seems in all places a failure is treated same way as mlock_cap==0

src/ucs/sys/sys.c
```diff
@@ -767,23 +767,52 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
...
+        snprintf(buf, max,
+                 ", please check for CAP_IPC_LOCK privilege for using "
+                 "SHM_HUGETLB");
+        return;
+    }
+
+    if (!mlock_cap) {
+        /* detected missing CAP_IPC_LOCK */
+        snprintf(buf, max,
+                 ", CAP_IPC_LOCK privilege is needed for SHM_HUGETLB");
```
yosefe, line 814: these error messages seem the same

src/ucs/sys/sys.h
```diff
@@ -635,6 +635,16 @@ unsigned long ucs_sys_get_proc_create_time(pid_t pid);
  */
 ucs_status_t ucs_sys_get_memlock_rlimit(size_t *rlimit_value);
 
+/*
```
yosefe, line 638: 2 space lines in H file

src/ucs/sys/sys.h
```diff
@@ -635,6 +635,16 @@ unsigned long ucs_sys_get_proc_create_time(pid_t pid);
...
+ * Get effective max locked memory limit (unlimited for privileged user).
+ * In case we can't query the system for capabilities, we fallback to
+ * 'ucs_sys_get_memlock_rlimit'
+ *
+ * @param [out] rlimit_value If successful, set to the current limit value.
+ *
+ * @return UCS_OK if successful, or error status if failed.
+ */
+ucs_status_t ucs_sys_get_effective_memlock_rlimit(size_t *rlimit_value);
```
yosefe, line 647: 2 space lines in H file

src/uct/ib/base/ib_log.c
```diff
@@ -241,7 +241,9 @@ void uct_ib_mem_lock_limit_msg(const char *message, int sys_errno,
     ucs_status_t status;
 
     if (sys_errno == ENOMEM) {
-        status = ucs_sys_get_memlock_rlimit(&memlock_limit);
+        /* only if the user is unprivileged, print a warning */
+        status = ucs_sys_get_effective_memlock_rlimit(&memlock_limit);
+
```
yosefe, line 246: remove spaceline

src/ucs/sys/sys.c
shasson5: you mean we should combine them to a single message?

src/ucs/sys/sys.c
gleon99: Status is used in `ucs_sysv_shmget_error_check_EPERM`.

src/ucs/sys/sys.c
shasson5: I think there's indeed a difference:
the first message tells the user to check the privilege himself because we didn't succeed.
the second message tells him that the required privilege is missing.
@yosefe

yosefe: i think "please check for CAP_IPC_LOCK privilege for using SHM_HUGETLB" message can cover both cases

src/ucs/sys/sys.c
yosefe: i think the slightly different message does not justify the extra "confusion"
ucs_sys_check_mlock_cap() returns 1 if have memlock capability, or 0 if do not have the capability or it could not be detected.

src/ucs/sys/sys.c
```diff
@@ -770,48 +770,30 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
...
     status = ucs_sys_get_proc_cap(&ecap);
-
-    if ((status == UCS_OK) && (ecap & UCS_BIT(CAP_IPC_LOCK))) {
-        *mlock_cap = 1;
-    } else {
-        *mlock_cap = 0;
-    }
-
-    return status;
+    return ((status == UCS_OK) && (ecap & UCS_BIT(CAP_IPC_LOCK)));
```
gleon99, line 784: Redundant `()`

gleon99 (dismissed):
👍

src/ucs/sys/sys.c
```diff
@@ -767,23 +767,34 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
...
+    ucs_status_t status;
 
     UCS_STATIC_ASSERT(CAP_IPC_LOCK < 32);
     status = ucs_sys_get_proc_cap(&ecap);
-    if ((status == UCS_OK) && !(ecap & UCS_BIT(CAP_IPC_LOCK))) {
-        /* detected missing CAP_IPC_LOCK */
-        snprintf(buf, max, ", CAP_IPC_LOCK privilege is needed for SHM_HUGETLB");
-        return;
-    }
+    return (status == UCS_OK) && (ecap & UCS_BIT(CAP_IPC_LOCK));
```
yosefe, line 784: 👍

src/ucs/sys/sys.c
```diff
@@ -767,23 +767,34 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
...
+ *
+ * @return A boolean that indicates whether the
+ *         process have the required capability.
+ */
+static int ucs_sys_get_mlock_cap()
 {
 #if HAVE_SYS_CAPABILITY_H
-    ucs_status_t status;
     uint32_t ecap;
+    ucs_status_t status;
```
yosefe, line 780: move this before line 779

src/ucs/sys/sys.h
```diff
@@ -636,6 +636,18 @@ unsigned long ucs_sys_get_proc_create_time(pid_t pid);
...
+/*
+ * Get effective max locked memory limit (unlimited for privileged user).
+ * In case we can't query the system for capabilities, we fallback to
+ * @ref ucs_sys_get_memlock_rlimit
+ *
+ * @param [out] rlimit_value If successful, set to the current limit value.
+ *
+ * @return UCS_OK if successful, or error status if failed.
+ */
+ucs_status_t ucs_sys_get_effective_memlock_rlimit(size_t *rlimit_value);
```
yosefe, line 648: i think we better replace ucs_sys_get_memlock_rlimit (or rename it to ucs_sys_get_memlock_limit)
since the bottom line that UCS users need to know is the effective limit.
ucs_sys_get_memlock_rlimit could be removed altogether (inlined to caller function)

src/uct/ib/base/ib_log.c
```diff
@@ -241,7 +241,8 @@ void uct_ib_mem_lock_limit_msg(const char *message, int sys_errno,
     ucs_status_t status;
 
     if (sys_errno == ENOMEM) {
-        status = ucs_sys_get_memlock_rlimit(&memlock_limit);
+        /* only if the user is unprivileged, print a warning */
```
yosefe, line 244: this comment can be removed

yosefe (dismissed):

yosefe (dismissed):

src/ucs/sys/sys.c
```diff
@@ -767,23 +767,34 @@ ucs_status_t ucs_sys_get_proc_cap(uint32_t *effective)
 #endif
 }
 
-static void ucs_sysv_shmget_error_check_EPERM(int flags, char *buf, size_t max)
+/**
+ * Checks whether the process have memory lock capability.
```
gleon99, line 771: `have` -> `has`. Also in `return`.

yosefe (approved):

gleon99:
@yosefe  please re-approve

PR #8197, opened 2022-05-03
UCT/ROCM: fix memory type detection

edgargabriel:
## What
Fix the approach used to identify ROCm memory type. 

## Why ?
The current solution marks certain memory types as ROCm memory, that can however not be used for typical operations on ROCm memory, such as opening an IPC handle.

## How ?
This commit identifies ROCm memory as of type HSA_EXT_POINTER_TYPE_HSA with the owner agent being a GPU.


src/uct/rocm/base/rocm_base.c
```diff
@@ -183,10 +183,18 @@ ucs_status_t uct_rocm_base_detect_memory_type(uct_md_h md, const void *addr,
     }
 
     info.size = sizeof(hsa_amd_pointer_info_t);
-    status = hsa_amd_pointer_info((void*)addr, &info, NULL, NULL, NULL);
+    status    = hsa_amd_pointer_info((void*)addr, &info, NULL, NULL, NULL);
     if ((status == HSA_STATUS_SUCCESS) &&
-        (info.type != HSA_EXT_POINTER_TYPE_UNKNOWN)) {
-        *mem_type_p = UCS_MEMORY_TYPE_ROCM;
+        (info.type == HSA_EXT_POINTER_TYPE_HSA)) {
+        hsa_device_type_t dev_type;
```
yosefe, line 189: better move this to the beginning of the function

edgargabriel: done

yosefe (approved):

PR #8201, opened 2022-05-06
UCP/UCT/UD: Reduce error detection time

dmitrygx:
## What

1. Reduce error detection time by setting `UCX_KEEPALIVE_INTERVAL=10s` and `UCX_UD_PEER_TIMEOUT=20s`.
2. Compare current time and time of last unacked SKB in TX window to check whether peer is unreachable or not.

## Why ?

1. To meet requirement of detecting an error wihtin 1 minute time limit.
It fixes the following case:
`1651857404` - switch port off
`1651857406` - [client] IO-demo traffic
`1651857406` - [client] did EP_CHECK on UD
`1651857424` - switch port on
<EP_CHECK was acked>
`1651857426` - [client] did EP_CHECK on UD
<EP_CHECK was acked>
`1651857427` - [server] detected error on RC (Retry count exceeded), server EP was destroyed
`1651857446` - [client] did EP_CHECK on UD
`1651857446` - [client] did EP_CHECK on UD
`1651857466` - [client] timeout waiting for replies after 1 minute
2. Using EP's `send_time` isn't correct, since the error could be not detected at all, if doing send operations every `[0..peer_timeout)` seconds.

## How ?

1. Set `UCX_KEEPALIVE_INTERVAL=10s` (instead of `20s`) and `UCX_UD_PEER_TIMEOUT=20s` (instead of `30s`, it aligns with RC/DC which detect an error in `~21s`).
2. To use SKB's `send_time` instead of EP's `send_time` do the following things:
- Remove unneeded code.
- Add `send_time` field to SKB.
- Set `ucs_get_time()` to SKB's `send_time` every time operation is done.
- Compare SKB's `send_time` and current time in `uct_ud_ep_timer` to check whether peer is unreachable or not.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -397,7 +397,7 @@ static ucs_config_field_t ucp_config_table[] = {
    "Experimental: enable new protocol selection logic",
    ucs_offsetof(ucp_config_t, ctx.proto_enable), UCS_CONFIG_TYPE_BOOL},
 
-  {"KEEPALIVE_INTERVAL", "20s",
+  {"KEEPALIVE_INTERVAL", "10s",
    "Time interval between keepalive rounds. Must be non-zero value.",
    ucs_offsetof(ucp_config_t, ctx.keepalive_interval),
    UCS_CONFIG_TYPE_TIME_UNITS},
--- a/src/uct/ib/ud/base/ud_def.h
+++ b/src/uct/ib/ud/base/ud_def.h
@@ -149,6 +149,7 @@ enum {
  */
 typedef struct uct_ud_send_skb {
     ucs_queue_elem_t        queue;      /* in send window */
+    ucs_time_t              send_time;  /* time of sending */
     uint32_t                lkey;
     uint16_t                len;        /* data size */
     uint16_t                flags;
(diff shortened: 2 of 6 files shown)


yosefe:
This PR is pretty intrusive: it increases the KA frequency and increases UD skb size.
Seems there are 2 issues here:
1. After server destroys ucp endpoint (1651857427), there is not enough time for UD to detect the issue. Can this be solved by increasing the timeout for replies in MTT from 60 to 120 sec?
2. "Using EP's send_time isn't correct, since the error could be not detected at all" - maybe instead of adding send_time to each sub, which increases its size, we don't update ep->send_time when window is non empty - so it will be the sending time of the "oldest" skb and not the "newest" sub?

dmitrygx:
> 1. After server destroys ucp endpoint (1651857427), there is not enough time for UD to detect the issue. Can this be solved by increasing the timeout for replies in MTT from 60 to 120 sec?

yes, it would fix the problem. I thought that we have a 1-minute requirement for error detection.

> 2\. maybe instead of adding send_time to each sub, which increases its size, we don't update ep->send_time when window is non empty - so it will be the sending time of the "oldest" skb and not the "newest" sub?

yes, it makes sense. I'll rework the PR.

dmitrygx:
the PR will be replaced by the new solution which is less intrusive.

PR #8203, opened 2022-05-08
UCP/UCT/UCS: Add define for zero linear function

yosefe:
## Why
Will be used in several places in protocol v2 performance calculations


yosefe:
@dmitrygx can you pls take a look?

dmitrygx (approved):

PR #8204, opened 2022-05-09
UCP/CORE: added atomic caps comparison to 'ucp_worker_iface_find_better'

shasson5:
## What
Same or better atomics capabilities are now a condition when choosing interfaces for a specific worker

## Why ?
For each worker we need the minimal set of interfaces to use, such that all capabilities are at least as good as all the rest.
(including atomics)

### DIFF
--- a/src/ucp/core/ucp_worker.c
+++ b/src/ucp/core/ucp_worker.c
@@ -887,6 +887,22 @@ static int ucp_worker_iface_support_keepalive(ucp_worker_iface_t *wiface)
                               UCT_IFACE_FLAG_EP_KEEPALIVE);
 }
 
+/* Compare 'compare_attr' atomic capabilities to 'test_attr',
+ * and decide if it contains all of 'test_attr' capabilities */
+static int
+ucp_worker_iface_compare_atomic_caps(const uct_iface_attr_t *compare_attr,
+                                                const uct_iface_attr_t *test_attr)
+{
+    return ucs_test_all_flags(compare_attr->cap.atomic32.op_flags,
+                              test_attr->cap.atomic32.op_flags) &&
+           ucs_test_all_flags(compare_attr->cap.atomic32.fop_flags,
+                              test_attr->cap.atomic32.fop_flags) &&
+           ucs_test_all_flags(compare_attr->cap.atomic64.op_flags,
+                              test_attr->cap.atomic64.op_flags) &&
+           ucs_test_all_flags(compare_attr->cap.atomic64.fop_flags,
+                              test_attr->cap.atomic64.fop_flags);
+}
+
 static int ucp_worker_iface_find_better(ucp_worker_h worker,
                                         ucp_worker_iface_t *wiface,
                                         ucp_rsc_index_t *better_index)
@@ -924,17 +940,20 @@ static int ucp_worker_iface_find_better(ucp_worker_h worker,
         if (/* 1. Supports all capabilities of the target iface (at least),
              *    except ...CONNECT_TO... and KEEPALIVE-related caps. */
             ucs_test_all_flags(if_iter->attr.cap.flags, test_flags) &&
-            /* 2. Has the same or better performance characteristics */
+            /* 2. Supports all atomic capabilities of the target iface (at least) */
+            ucp_worker_iface_compare_atomic_caps(&if_iter->attr,
+                                                 &wiface->attr) &&
+            /* 3. Has the same or better performance characteristics */
             (if_iter->attr.overhead <= wiface->attr.overhead) &&
             (ucp_tl_iface_bandwidth(ctx, &if_iter->attr.bandwidth) >= bw_cur) &&
             /* swap latencies in args list since less is better */
-            (ucp_score_prio_cmp(latency_cur,  if_iter->attr.priority,
+            (ucp_score_prio_cmp(latency_cur, if_iter->attr.priority,
                                 latency_iter, wiface->attr.priority) >= 0) &&
-            /* 3. The found transport is scalable enough or both
+            /* 4. The found transport is scalable enough or both
              *    transports are unscalable */
             (ucp_is_scalable_transport(ctx, if_iter->attr.max_num_eps) ||
              !ucp_is_scalable_transport(ctx, wiface->attr.max_num_eps)) &&
-            /* 4. The found transport supports at least one keepalive mechanism
+            /* 5. The found transport supports at least one keepalive mechanism
              *    or both transports don't support them
              */
             (ucp_worker_iface_support_keepalive(if_iter) ||


gleon99:
Please fix the PR description:
1. Start sentences with capital letters.
2. Technically we don't necessarily look for "better" caps, but "at least as good as..."

src/ucp/core/ucp_worker.c
```diff
@@ -887,6 +887,16 @@ static int ucp_worker_iface_support_keepalive(ucp_worker_iface_t *wiface)
                               UCT_IFACE_FLAG_EP_KEEPALIVE);
 }
 
+/* Compare target iface atomic capabilities to candidate's, and decide if candidate is better */
+static int ucp_worker_iface_is_better_atomic_caps(uct_iface_attr_t *cand_attr,
```
gleon99, line 891: Don't abbreviate (`cand`).
Possible alternatives: target - compare, attr - test_attr.

src/ucp/core/ucp_worker.c
```diff
@@ -887,6 +887,16 @@ static int ucp_worker_iface_support_keepalive(ucp_worker_iface_t *wiface)
...
+                                                  uct_iface_attr_t *target_attr)
+{
+    return (ucs_test_all_flags(cand_attr->cap.atomic32.op_flags, target_attr->cap.atomic32.op_flags)) &&
+           (ucs_test_all_flags(cand_attr->cap.atomic32.fop_flags, target_attr->cap.atomic32.fop_flags)) &&
+           (ucs_test_all_flags(cand_attr->cap.atomic64.op_flags, target_attr->cap.atomic64.op_flags)) &&
+           (ucs_test_all_flags(cand_attr->cap.atomic64.fop_flags, target_attr->cap.atomic64.fop_flags));
```
gleon99, line 897: Did you test it, at least manually (runtime injection)?

src/ucp/core/ucp_worker.c
gleon99, line 891: I would drop the "better" semantics from this func name (and comment).
It does something very specific - checks whether atomic caps are contained (subset).

shasson5: how about `ucp_worker_iface_atomic_caps_is_subset`?

src/ucp/core/ucp_worker.c
shasson5: I ran ucx_info and verified I that this code runs properly (returns true for subset and false for no subset)

src/ucp/core/ucp_worker.c
shasson5: not sure I understand, you mean switch the names?
can you write the full function signature?

gleon99: I'd go with something like `test_atomic_caps` (like `ucs_test_all_flags`)

gleon99: 1. Abbreviations are bad (`cand`).
2. Let's replace the names. Instead of `cand_attr` - `target_attr`, can have some combination of `test_attr` / `compare_attr` / `target_attr`.

gleon99, line 891: Why pass ptr to attr and no to the caps?

shasson5: notice that the second argument is already called `target_attr` (cause I had `candidate` which was compared to the `target`).
so if we rename the first argument to `target`, we should also rename the second argument.
maybe we should just call them `first` and `second`?

shasson5: `cap` is an implicit struct inside `uct_iface_attr_t` (it has no type name).
is it possible to reference this kind of struct as a pointer?

gleon99: `first` and `second` doesn't show the difference in their meaning.
How about `compare_attr` - `test_attr`?
To resemble something similar to https://github.com/openucx/ucx/blob/master/src/ucp/core/ucp_worker.c#L926

gleon99: Right. Then let's leave it with `attr`.

src/ucp/core/ucp_worker.c
```diff
@@ -887,6 +887,20 @@ static int ucp_worker_iface_support_keepalive(ucp_worker_iface_t *wiface)
                               UCT_IFACE_FLAG_EP_KEEPALIVE);
 }
 
+/* Compare target iface atomic capabilities to candidate's, and decide if candidate contains all target's capabilities */
```
gleon99, line 890: `candidate` doesn't exist in func sig anymore, so please adjust the comment text accordingly.

gleon99, line 890: Line too long

gleon99:
👍

gleon99:
@yosefe  please look

gleon99 (approved):

src/ucp/core/ucp_worker.c
```diff
@@ -887,6 +887,21 @@ static int ucp_worker_iface_support_keepalive(ucp_worker_iface_t *wiface)
                               UCT_IFACE_FLAG_EP_KEEPALIVE);
 }
 
+/* Compare 'compare_attr' atomic capabilities to 'test_attr',
+ * and decide if it contains all of 'test_attr' capabilities */
+static int ucp_worker_iface_test_atomic_caps(uct_iface_attr_t *compare_attr,
```
yosefe, line 892: 1. maybe ucp_worker_iface_compare_atomic_caps
2. make the params const*

src/ucp/core/ucp_worker.c
```diff
@@ -887,6 +887,21 @@ static int ucp_worker_iface_support_keepalive(ucp_worker_iface_t *wiface)
...
+                                             uct_iface_attr_t *test_attr)
+{
+    return (ucs_test_all_flags(compare_attr->cap.atomic32.op_flags,
```
yosefe, line 895: extra ( ) - in every call to ucs_test_all_flags

shasson5: they are already wrapped with extra ().
maybe another () around the whole `return`?

gleon99: I think Yossi meant that these () are not needed there and can be removed.

src/ucp/core/ucp_worker.c
```diff
@@ -889,17 +889,17 @@ static int ucp_worker_iface_support_keepalive(ucp_worker_iface_t *wiface)
...
-                               test_attr->cap.atomic64.op_flags)) &&
-           (ucs_test_all_flags(compare_attr->cap.atomic64.fop_flags,
-                               test_attr->cap.atomic64.fop_flags));
+                               test_attr->cap.atomic32.op_flags) &&
+            ucs_test_all_flags(compare_attr->cap.atomic32.fop_flags,
+                               test_attr->cap.atomic32.fop_flags) &&
+            ucs_test_all_flags(compare_attr->cap.atomic64.op_flags,
+                               test_attr->cap.atomic64.op_flags) &&
+            ucs_test_all_flags(compare_attr->cap.atomic64.fop_flags,
+                                test_attr->cap.atomic64.fop_flags));
```
yosefe, line 902: we can also remove the ( ) around the whole expression

yosefe (approved):
pls squash

yosefe (approved):

yosefe (approved):

PR #8205, opened 2022-05-09
UCT/ROCM/IPC: use remote_agent if available

edgargabriel:
## What
use the correct remote_agent handle if information is available. Using the local_agent for both src and destination is just the backup plan in case the remote_agent info is not available, e.g. in case the user limited the visibility of devices for a/some/all processes.

## Why ?
using the local_agent for both src and destination triggers a different code both in the runtime layer than using two different agents, limiting the availability of the CUs for compute operations.

### DIFF
--- a/src/uct/rocm/ipc/rocm_ipc_ep.c
+++ b/src/uct/rocm/ipc/rocm_ipc_ep.c
@@ -58,7 +58,8 @@ ucs_status_t uct_rocm_ipc_ep_zcopy(uct_ep_h tl_ep,
 {
     uct_rocm_ipc_ep_t *ep = ucs_derived_of(tl_ep, uct_rocm_ipc_ep_t);
     hsa_status_t status;
-    hsa_agent_t local_agent;
+    hsa_agent_t local_agent, remote_agent;
+    hsa_agent_t dst_agent, src_agent;
     size_t size = uct_iov_get_length(iov);
     ucs_status_t ret = UCS_OK;
     void *base_addr, *local_addr = iov->buffer;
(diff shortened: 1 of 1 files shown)


src/uct/rocm/ipc/rocm_ipc_ep.c
```diff
@@ -94,20 +99,46 @@ ucs_status_t uct_rocm_ipc_ep_zcopy(uct_ep_h tl_ep,
...
+    memset(&remote_agent, 0, sizeof(hsa_agent_t));
+    status = uct_rocm_base_get_ptr_info(remote_copy_addr, size, &tmp_base_ptr,
+					&tmp_base_size,	&remote_agent);
+    if (status != HSA_STATUS_SUCCESS) {
+	return UCS_ERR_INVALID_ADDR;
+    }
+
+    if (remote_agent.handle == 0) {
+	//No access to remote agent, e.g. because of limited visability of devices to
+	//this process. Using local_agent as a backup plan.
```
yosefe, line 112: pls use C-style comments

src/uct/rocm/ipc/rocm_ipc_ep.c
```diff
@@ -94,20 +99,46 @@ ucs_status_t uct_rocm_ipc_ep_zcopy(uct_ep_h tl_ep,
...
+	remote_agent = local_agent;
+    }
+    else {
```
yosefe, line 115: code style
pls check the output of "Code format" stage in the CI

src/uct/rocm/ipc/rocm_ipc_ep.c
edgargabriel: fixed.

src/uct/rocm/ipc/rocm_ipc_ep.c
edgargabriel: if the issue here is the location of the 'else' statement, I moved it up to follow the rest of the ucx code. I also saw that there are some tabs/spaces issues, I fixed them. Beyond it, I am working on compiling the llvm tool to run the clang-format manually (which takes however a while), since the output of the CI stage does not actually tell what is wrong with the formatting, it just says that it 'passed with issues'.

yosefe: @edgargabriel  pls see https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=43598&view=logs&j=cc064a77-22b5-56bf-ecc0-70b5fe764261&t=aedfd754-44a6-53e6-6843-8659d800fee2

yosefe (approved):

PR #8211, opened 2022-05-12
UCT/IB/RC: Fix always having STATUS flag in iface send op

dmitrygx:
## What

Fix always having STATUS flag in IFACE send ops.

## Why ?

Send ops which belong to IFACE must have only `UCT_RC_IFACE_SEND_OP_FLAG_IFACE` flag set when kept in the free list.

## How ?

1. Move setting status and `UCT_RC_IFACE_SEND_OP_STATUS` flag to the place where we're calling `op`'s `handler` (only there the status could be read).
2. Add `assert`s to make sure only `UCT_RC_IFACE_SEND_OP_FLAG_IFACE` is set when doing put/get for IFACE send ops.

### DIFF
--- a/src/uct/ib/rc/base/rc_ep.c
+++ b/src/uct/ib/rc/base/rc_ep.c
@@ -470,9 +470,6 @@ void uct_rc_txqp_purge_outstanding(uct_rc_iface_t *iface, uct_rc_txqp_t *txqp,
 
     ucs_queue_for_each_extract(op, &txqp->outstanding, queue,
                                UCS_CIRCULAR_COMPARE16(op->sn, <=, sn)) {
-        op->status = status;
-        op->flags |= UCT_RC_IFACE_SEND_OP_STATUS;
-
         if (op->handler != (uct_rc_send_handler_t)ucs_mpool_put) {
             /* Allow clean flush cancel op from destroy flow */
             if (warn &&
@@ -520,6 +517,8 @@ void uct_rc_txqp_purge_outstanding(uct_rc_iface_t *iface, uct_rc_txqp_t *txqp,
             desc = ucs_derived_of(op, uct_rc_iface_send_desc_t);
             ucs_mpool_put(desc);
         } else {
+            op->status = status;
+            op->flags |= UCT_RC_IFACE_SEND_OP_STATUS;
             op->handler(op, NULL);
         }
     }
--- a/src/uct/ib/rc/base/rc_iface.h
+++ b/src/uct/ib/rc/base/rc_iface.h
@@ -494,8 +494,8 @@ uct_rc_iface_add_cq_credits(uct_rc_iface_t *iface, uint16_t cq_credits)
 static UCS_F_ALWAYS_INLINE uct_rc_iface_send_op_t*
 uct_rc_iface_get_send_op(uct_rc_iface_t *iface)
 {
-    uct_rc_iface_send_op_t *op;
-    op = iface->tx.free_ops;
+    uct_rc_iface_send_op_t *op = iface->tx.free_ops;
+
     iface->tx.free_ops = op->next;
     return op;
 }
@@ -505,7 +505,8 @@ uct_rc_iface_put_send_op(uct_rc_iface_send_op_t *op)
 {
     uct_rc_iface_t *iface = op->iface;
 
-    ucs_assert(op->flags & UCT_RC_IFACE_SEND_OP_FLAG_IFACE);
+    ucs_assertv(op->flags == UCT_RC_IFACE_SEND_OP_FLAG_IFACE,
+                "op %p flags 0x%x", op, op->flags);
 
     op->next = iface->tx.free_ops;
     iface->tx.free_ops = op;


src/uct/ib/rc/base/rc_iface.h
```diff
@@ -505,7 +510,8 @@ uct_rc_iface_put_send_op(uct_rc_iface_send_op_t *op)
 {
     uct_rc_iface_t *iface = op->iface;
 
-    ucs_assert(op->flags & UCT_RC_IFACE_SEND_OP_FLAG_IFACE);
+    ucs_assertv(op->flags == UCT_RC_IFACE_SEND_OP_FLAG_IFACE,
```
yosefe, line 513: IMO no need to have assert both when allocating and when releasing. one of them is enough

src/uct/ib/rc/base/rc_ep.c
```diff
@@ -520,6 +517,9 @@ void uct_rc_txqp_purge_outstanding(uct_rc_iface_t *iface, uct_rc_txqp_t *txqp,
             desc = ucs_derived_of(op, uct_rc_iface_send_desc_t);
             ucs_mpool_put(desc);
         } else {
+            ucs_assert(!(op->flags & UCT_RC_IFACE_SEND_OP_FLAG_IFACE));
+            op->status = status;
+            op->flags |= UCT_RC_IFACE_SEND_OP_STATUS;
```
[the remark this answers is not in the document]
dmitrygx: no, IFACE's `tx.ops_buffer` expect to be initialized only by `UCT_RC_IFACE_SEND_OP_FLAG_IFACE` flag. others should be removed upon returning op to the `tx.free_ops` list.

src/uct/ib/rc/base/rc_iface.h
dmitrygx: removed from `uct_rc_iface_get_send_op`

dmitrygx:
@yosefe could you review pls?
the PR is small enough, new changes were force-pushed.

src/uct/ib/rc/base/rc_iface.h
```diff
@@ -494,9 +494,12 @@ uct_rc_iface_add_cq_credits(uct_rc_iface_t *iface, uint16_t cq_credits)
 static UCS_F_ALWAYS_INLINE uct_rc_iface_send_op_t*
 uct_rc_iface_get_send_op(uct_rc_iface_t *iface)
 {
-    uct_rc_iface_send_op_t *op;
-    op = iface->tx.free_ops;
+    uct_rc_iface_send_op_t *op = iface->tx.free_ops;
+
+    ucs_assert(op != NULL);
```
yosefe, line 499: remove this assert

src/uct/ib/rc/base/rc_iface.h
```diff
@@ -494,9 +494,12 @@ uct_rc_iface_add_cq_credits(uct_rc_iface_t *iface, uint16_t cq_credits)
...
+
     iface->tx.free_ops = op->next;
+
```
yosefe, line 502: remove space line

src/uct/ib/rc/base/rc_ep.c
```diff
@@ -520,6 +517,9 @@ void uct_rc_txqp_purge_outstanding(uct_rc_iface_t *iface, uct_rc_txqp_t *txqp,
             desc = ucs_derived_of(op, uct_rc_iface_send_desc_t);
             ucs_mpool_put(desc);
         } else {
+            ucs_assert(!(op->flags & UCT_RC_IFACE_SEND_OP_FLAG_IFACE));
```
yosefe, line 520: why need this assert

dmitrygx: we shouldn't set `UCT_RC_IFACE_SEND_OP_STATUS` on IFACE-owned ops, because `UCT_RC_IFACE_SEND_OP_STATUS` isn't removed then

src/uct/ib/rc/base/rc_iface.h
dmitrygx: it would be helpful to find errors not by Valgrind, but by assertion

src/uct/ib/rc/base/rc_iface.h
dmitrygx: done

src/uct/ib/rc/base/rc_iface.h
yosefe: wouldn't the next line segfault anyway?

src/uct/ib/rc/base/rc_ep.c
yosefe: but we would anyway fail when returning the iface op to the memory pool, no?

dmitrygx: but we fail if we use `uct_rc_iface_put_send_op` function which checks the flag

src/uct/ib/rc/base/rc_iface.h
dmitrygx: ok, removed

src/uct/ib/rc/base/rc_ep.c
dmitrygx: here we just call `op->handler`

yosefe: i would expect the handler to release the op - and fail on the assert in uct_rc_iface_get_send_op later

dmitrygx: ok, removed

yosefe (approved):

brminich (approved):

PR #8214, opened 2022-05-12
UCP/CONTEXT/CONFIG: Use ucs_config_parser_clone_opts for copying config

yosefe:
## Why
While adding new string configuration to ucp_context_config_t (proto_info_dir), it turned out we need to strdup() it manually. But a better method is to use ucs_config_parser_clone_opts().

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -113,6 +113,264 @@ const char *ucp_object_versions[] = {
 static UCS_CONFIG_DEFINE_ARRAY(memunit_sizes, sizeof(size_t),
                                UCS_CONFIG_TYPE_MEMUNITS);
 
+static ucs_config_field_t ucp_context_config_table[] = {
+  {"SELECT_DISTANCE_MD", "cuda_cpy",
+   "MD whose distance is queried when evaluating transport selection score",
+   ucs_offsetof(ucp_context_config_t, select_distance_md), UCS_CONFIG_TYPE_STRING},
+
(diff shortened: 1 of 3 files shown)


src/ucp/core/ucp_context.c
```diff
@@ -1653,24 +1665,11 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
...
 
-    /* always init MT lock in context even though it is disabled by user,
-     * because we need to use context lock to protect ucp_mem_ and ucp_rkey_
-     * routines */
-    UCP_THREAD_LOCK_INIT(&context->mt_lock);
-
-    /* save comparison MD for iface_attr adjustment */
-    context->config.selection_cmp = ucs_strdup(config->selection_cmp,
-                                               "selection cmp");
-    if (context->config.selection_cmp == NULL) {
```
yosefe, line 1664: this cleanup is possible now that using clone_opts

src/ucp/core/ucp_context.c
```diff
@@ -1878,7 +1878,7 @@ ucs_status_t ucp_init_version(unsigned api_major_version, unsigned api_minor_ver
     context = ucs_calloc(1, sizeof(*context), "ucp context");
     if (context == NULL) {
         status = UCS_ERR_NO_MEMORY;
-        goto err_release_config;
+        goto err_release_dfl_config;
```
yosefe, line 1881: renamed label to avoid confusion

yosefe:
@dmitrygx can you pls take a look?

dmitrygx (approved):

PR #8219, opened 2022-05-13
UCT/IB/UD/GTEST: Fix error detection if new operations are scheduled

dmitrygx:
## What

Fix error detection if new operations are scheduled.

## Why ?

If peer was killed, and new operations are still scheduled on local side, it happens that `send_time` is always updated and we can't detect that `now - send_time > peer_timeout`.

## How ?

1. Reproduce the issue in the gtest by updating `test_uct_peer_failure_keepalive.killed_post_am` and do many AM operations with 2-second interval while an error isn't detected. Also, set `UCX_UD_TIMER_TICK=2s` to not try resending the data - otherwise, posting AM operations will be completed with `UCS_ERR_NOR_RESOURCE`.
2. Update `uct_ud_iface_complete_tx_skb` to setting `ep->tx.send_time` only if `ep->tx.window` is empty. So, `send_time` will be updated only when ACK will be received from a peer.


test/gtest/uct/test_peer_failure.cc
```diff
@@ -620,7 +625,8 @@ UCS_TEST_SKIP_COND_P(test_uct_peer_failure_keepalive, killed,
 
 UCS_TEST_SKIP_COND_P(test_uct_peer_failure_keepalive, killed_post_am,
                      !check_caps(UCT_IFACE_FLAG_EP_CHECK |
-                                 UCT_IFACE_FLAG_AM_SHORT))
+                                 UCT_IFACE_FLAG_AM_SHORT),
+                     "UD_TIMER_TICK?=2s")
```
yosefe, line 629: why needed?

test/gtest/uct/test_peer_failure.cc
```diff
@@ -594,16 +594,21 @@ class test_uct_peer_failure_keepalive : public test_uct_peer_failure
...
-            if (m_err_count == 0) {
-                ASSERT_UCS_OK(status);
+            const ucs_time_t loop_end_limit = ucs::get_deadline(10.0);
+            while ((m_err_count == 0) && (ucs_get_time() < loop_end_limit)) {
+                const uint64_t send_data = 0;
+                status = uct_ep_am_short(ep0(), 0, 0, &send_data,
+                                         sizeof(send_data));
+                if (m_err_count == 0) {
+                    ASSERT_UCS_OK(status);
+                    wait_for_flag(&m_err_count, 2.0);
```
yosefe, line 604: why 2.0 and not 2?

dmitrygx: fixed

test/gtest/uct/test_peer_failure.cc
dmitrygx: this is needed to start resending which would prevent us from doing AM operations (they return NO_RESOURCE)

yosefe: it seems weird, would the test fail without it?
and why do we expect m_err_count to be 2 ?

dmitrygx: > it seems weird, would the test fail without it?

yes, it will fail and wouldn’t test the original idea, because we need to do multiple AM sends without getting NO_RESOURCE

> and why do we expect m_err_count to be 2 ?

it is 2-second deadline

yosefe: > yes, it will fail and wouldn’t test the original idea, because we need to do multiple AM sends without getting NO_RESOURCE

Do you mean we need to avoid resending quickly?
What will be the failure?

> it is 2-second deadline

Ok

Can you please add a comment in the test to explain why need 2s timer tick?

dmitrygx: > > yes, it will fail and wouldn’t test the original idea, because we need to do multiple AM sends without getting NO_RESOURCE
> 
> Do you mean we need to avoid resending quickly? What will be the failure?

the failure will be returning `UCS_ERR_NO_RESOURCE` from `uct_ep_am_short`, but we want to test the case when AM operations are scheduled every `2` seconds to reproduce issue with setting `send_time` and not detecting an error


> > it is 2-second deadline
> 
> Ok
> 
> Can you please add a comment in the test to explain why need 2s timer tick?

done

yosefe: > the failure will be returning `UCS_ERR_NO_RESOURCE` from `uct_ep_am_short`, but we want to test the case when AM operations are scheduled every `2` seconds to reproduce issue with setting `send_time` and not detecting an error

Maybe we can test this in test_ud.cc instead?

dmitrygx: we don't have a test which continues sending when peer was killed, so it makes sense having it for all transports

dmitrygx: or you think that it worth moving it to move it to test_ud.cc, because it has dependency on UD behavior?

dmitrygx: unfortunately, we don't have peer failure tests in UD and we should introduce it first

yosefe: I'd say to remove this test for now. and then in next PR add UD specific test in test_ud.cc that includes packet drop by rx filter and making sure we start resending even if adding new packet every time.
in these tests we have better control over ud state.

dmitrygx: we don’t need resending, that’s why setting UD_TIMER_TICK to big enough value.
I’ll move this test to test_ud.cc

yosefe (approved):

PR #8224, opened 2022-05-15
DOC/FAQ: Use HTML markup for static build table

yosefe:
## Why
Fix static build section not rendered correctly in the read-the-docs site

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -132,13 +132,15 @@ program: program.c
         $(CC) -static program.c $(shell pkg-config --cflags --libs --static ucx-cma ucx-knem ucx)
 ```
 Currently, the following transport modules can be used with pkg-config:
-| Package name  |  Provided transport service |
-|-----------|-----------|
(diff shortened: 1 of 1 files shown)


shamisp:
It does not very aligned either https://openucx--8224.org.readthedocs.build/en/8224/faq.html

yosefe:
> It does not very aligned either https://openucx--8224.org.readthedocs.build/en/8224/faq.html

I want to convert the whole file to RST which is a more robust format. But first would like to fix the table for static build users, currently it's totally unreadable.

shamisp:
Maybe just render it as an PNG 🤷

yosefe:
> Maybe just render it as an PNG 🤷
> […](#)

Would like to support copy/paste

shamisp:
Good point. Also it is good for search.

yosefe:
@shamisp added space after the table

shamisp (approved):

PR #8228, opened 2022-05-16
UCP/PROTO: Refactor protocol initialization ranges

yosefe:
## Why
Preparation for adding protocols performance debug info.
The main goal is to get rid of `ucp_proto_select_range_t` and unify the performance envelope / stages calculation functions.
Each performance range will (in next PR) contain a performance "tree node" so better remove this extra hierarchy.

## How
- Flatten performance range heirarchy by moving cfg_thresh from `ucp_proto_select_range_t` to `ucp_proto_config_t`
- Remove perf_range name field; it will be replaced by a performance tree node with detailed information
- `ucp_proto_perf_envelope_make()` now just calculates the envelope of linear functions and is not aware of perf ranges (for better abstraction)
- `ucp_proto_init_parallel_stages()` combines the logic of `ucp_proto_perf_envelope_append()` and `ucp_proto_common_add_perf_ranges()`
- Split `ucp_proto_select_elem_init_thresh()` to 3 parts: generate candidates, calculate envelope, and add the selected protocols from the envelope.
- Add helper function `ucp_proto_caps_range_find()`
- Make `ucs_array_append()` more convenieit.

### DIFF
--- a/src/ucp/proto/proto.h
+++ b/src/ucp/proto/proto.h
@@ -119,9 +119,6 @@ typedef enum {
  * Performance estimation for a range of message sizes.
  */
 typedef struct {
-    /* Protocol name */
-    const char        *name;
-
     /* Maximal payload size for this range */
     size_t            max_length;
 
@@ -292,7 +289,7 @@ extern const char *ucp_operation_descs[];
 
 
 /* Performance types names */
-extern const char *ucp_proto_perf_types[];
+extern const char *ucp_proto_perf_type_names[];
 
 
 /* Get number of globally registered protocols */
@@ -303,5 +300,21 @@ unsigned ucp_protocols_count(void);
    description from proto->desc, and set config to an empty string. */
 void ucp_proto_default_query(const ucp_proto_query_params_t *params,
                              ucp_proto_query_attr_t *attr);
+
+static inline void
+ucp_proto_perf_copy(ucs_linear_func_t dest[UCP_PROTO_PERF_TYPE_LAST],
+                    const ucs_linear_func_t src[UCP_PROTO_PERF_TYPE_LAST])
+{
+    dest[UCP_PROTO_PERF_TYPE_SINGLE] = src[UCP_PROTO_PERF_TYPE_SINGLE];
+    dest[UCP_PROTO_PERF_TYPE_MULTI]  = src[UCP_PROTO_PERF_TYPE_MULTI];
+}
+
+static inline void
+ucp_proto_perf_add(ucs_linear_func_t perf[UCP_PROTO_PERF_TYPE_LAST],
+                   ucs_linear_func_t func)
+{
+    ucs_linear_func_add_inplace(&perf[UCP_PROTO_PERF_TYPE_SINGLE], func);
+    ucs_linear_func_add_inplace(&perf[UCP_PROTO_PERF_TYPE_MULTI], func);
+}
 
 #endif
--- a/src/ucp/proto/proto_init.h
+++ b/src/ucp/proto/proto_init.h
@@ -16,18 +16,19 @@
 
 /* Element of list of trends to select next best one */
 typedef struct {
-    uint64_t                 id;
-    ucp_proto_select_range_t range;
+    size_t   max_length;  /* Maximal message size */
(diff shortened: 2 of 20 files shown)


src/ucs/datastruct/string_buffer.c
```diff
@@ -35,6 +35,10 @@ void ucs_string_buffer_init_fixed(ucs_string_buffer_t *strb, char *buffer,
                                   size_t capacity)
 {
     ucs_array_init_fixed(&strb->str, buffer, capacity);
+    if (capacity > 0) {
+        // TODO gtest this by initializing empty onstack and checking it's ""
```
dmitrygx, line 39: shouldn't we avoid C++-style comments in C code?

src/ucp/proto/proto_init.h
```diff
@@ -44,32 +45,42 @@ void ucp_proto_common_init_base_caps(
...
-/*
- * Accepts a list of performance functions for a given range
- * (min_length..max_length), and appends the convex (or concave)
- * envelope segments of these functions to the 'envelope_list'. The envelope
- * is the minimal (or maximal) function at each point in that range.
+/**
+ * Initialize the performance of a protocol that consists of several parallel
+ * stages. The performance estimations are added to params->caps.
+ *
+ * @param [in] params        Protocol initialization parameters
```
dmitrygx, line 70: add a trailing dot as for other fields?

src/ucp/proto/proto_init.h
```diff
@@ -44,32 +45,42 @@ void ucp_proto_common_init_base_caps(
 
 
 /*
- * Accepts list of operations that happen in parallel and create a ranges
- * on min..max which represents their overall performance.
+ * Accepts a list of performance functions for a given range and appends the
+ * convex or concave envelope of these functions to an output list.
+ *
+ * @param [in] perf_list       List of performance functions.
+ * @param [in] range_start     Range interval start
```
dmitrygx, line 52: add a trailing dot as for other fields?

src/ucp/proto/proto_init.h
```diff
@@ -44,32 +45,42 @@ void ucp_proto_common_init_base_caps(
...
+ * @param [in] range_end       Range interval end.
+ * @param [in] convex          Whether to select convex (maximal) or concave
+ *                             (minimal) function from 'perf_list'.
```
dmitrygx, line 55: maybe replace "function" by "trend", because we couldn't select "convex"/"concave" from linear functions (which are both convex and concave), but used points of the linear functions could give non-linear function.
wdyt?

src/ucp/proto/proto.h
```diff
@@ -304,4 +301,20 @@ unsigned ucp_protocols_count(void);
...
+ucp_proto_perf_copy(ucs_linear_func_t dest[UCP_PROTO_PERF_TYPE_LAST],
+                    const ucs_linear_func_t src[UCP_PROTO_PERF_TYPE_LAST])
+{
+    dest[UCP_PROTO_PERF_TYPE_SINGLE] = src[UCP_PROTO_PERF_TYPE_SINGLE];
+    dest[UCP_PROTO_PERF_TYPE_MULTI]  = src[UCP_PROTO_PERF_TYPE_MULTI];
+}
+
+static inline void
+ucp_proto_perf_add(ucs_linear_func_t perf[UCP_PROTO_PERF_TYPE_LAST],
+                   ucs_linear_func_t func)
```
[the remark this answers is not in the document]
yosefe: in most places we pass `ucs_linear_func_t` by value so it can be the result of a function call like `ucs_linear_func_make()`

src/ucp/proto/proto_init.h
yosefe: actually, we do select a function: for each range interval, the result is "index" of the function from "perf_list" that is min/max in this interval.

yosefe:
@dmitrygx can you pls review again?

dmitrygx (approved):

yosefe:
@brminich @Artemy-Mellanox can you pls take a look?

brminich (dismissed):

PR #8230, opened 2022-05-17
UCT/IB: Add hybrid SRQ topology

huaxiangfan:
## What
Add a new SRQ topology: hybrid

## Why ?
list topology is suboptimal while cyclic_emulated is stopping SRQ in case of error flows. we need a hybrid which behaves as cyclic_emulated in normal case and behaves as list in case of error flows.

## How ?
try to use segments in order. only use next noncontiguous segment if wait for a time limit.


gleon99:
Hi @huaxiangfan - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #8236, opened 2022-05-17
UCT/IB: Remove unneeded return statement

dmitrygx:
## What

Remove unneeded `return` statement.

## Why ?

We could simply remove unneeded code.

## How ?

Remove check for `IBV_QPT_UD` qp type and `return` statement..


brminich (approved):

dmitrygx:
@yosefe could you review pls?

yosefe (approved):

PR #8238, opened 2022-05-18
UCP/CORE: Refactor endpoint extension

gleon99:
## What
Unite all EP extensions to a single struct

## Why ?
Preparation for supporting many lanes

### DIFF
--- a/src/tools/info/type_info.c
+++ b/src/tools/info/type_info.c
@@ -261,8 +261,7 @@ void print_type_info(const char * tl_name)
     PRINT_SIZE(ucp_context_t);
     PRINT_SIZE(ucp_worker_t);
     PRINT_SIZE(ucp_ep_t);
-    PRINT_SIZE(ucp_ep_ext_gen_t);
-    PRINT_SIZE(ucp_ep_ext_proto_t);
+    PRINT_SIZE(ucp_ep_ext_t);
     PRINT_SIZE(ucp_ep_config_key_t);
     PRINT_SIZE(ucp_ep_config_t);
     PRINT_SIZE(ucp_datatype_iter_t);
--- a/src/ucp/core/ucp_ep.h
+++ b/src/ucp/core/ucp_ep.h
@@ -439,6 +439,60 @@ struct ucp_ep_config {
 
 
 /**
+ * Status of protocol-level remote completions
+ */
+typedef struct {
+    ucs_hlist_head_t reqs; /* Queue of flush requests which
+                              are waiting for remote completion */
+    uint32_t         send_sn; /* Sequence number of sent operations */
+    uint32_t         cmpl_sn; /* Sequence number of completions */
+} ucp_ep_flush_state_t;
+
+
+/**
+ * Endpoint extension
+ */
+typedef struct ucp_ep_ext {
+    ucp_ep_h                      ep;            /* Back pointer to endpoint */
+    void                          *user_data;    /* User data associated with ep */
+    ucs_list_link_t               ep_list;       /* List entry in worker's all eps list */
+    ucp_rsc_index_t               cm_idx;        /* CM index */
+    ucs_ptr_map_key_t             local_ep_id;   /* Local EP ID */
+    ucs_ptr_map_key_t             remote_ep_id;  /* Remote EP ID */
+    ucp_err_handler_cb_t          err_cb;        /* Error handler */
+    ucp_request_t                 *close_req;    /* Close protocol request */
+    khash_t(ucp_ep_peer_mem_hash) *peer_mem;     /* Hash of remote memory segments
+                                                    used by 2-stage ppln rndv proto */
+    /* List of requests which are waiting for remote completion */
+    ucs_hlist_head_t              proto_reqs;
+#if UCS_ENABLE_ASSERT
+    ucs_time_t                    ka_last_round; /* Time of last KA round done */
+#endif
+
+    /* Endpoint match context and remote completion status are mutually exclusive,
+     * since remote completions are counted only after the endpoint is already
+     * matched to a remote peer.
+     */
+    union {
+        ucp_ep_match_elem_t       ep_match;      /* Matching with remote endpoints */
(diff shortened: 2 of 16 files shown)


src/ucp/core/ucp_ep.h
```diff
@@ -506,43 +511,37 @@ typedef struct {
...
      * matched to a remote peer.
      */
     union {
-        ucp_ep_match_elem_t       ep_match;      /* Matching with remote endpoints */
-        ucp_ep_flush_state_t      flush_state;   /* Remote completion status */
+        ucp_ep_match_elem_t  ep_match;      /* Matching with remote endpoints */
+        ucp_ep_flush_state_t flush_state;   /* Remote completion status */
     };
-    ucp_ep_ext_control_t          *control_ext;  /* Control data path extension */
+    ucp_ep_ext_control_t     control_ext;  /* Control data path extension */
```
gleon99, line 529: Maybe flatten the struct and get rid of `ucp_ep_ext_control_t`?

src/ucp/core/ucp_ep.h
```diff
@@ -438,6 +438,10 @@ struct ucp_ep_config {
 };
 
 
+/* Forward declaration */
+typedef struct ucp_ep_ext ucp_ep_ext_t;
```
yosefe, line 442: can we define ucp_ep_ext struct before ucp_ep to avoid this?

src/tools/info/type_info.c
```diff
@@ -261,8 +261,7 @@ void print_type_info(const char * tl_name)
     PRINT_SIZE(ucp_context_t);
     PRINT_SIZE(ucp_worker_t);
     PRINT_SIZE(ucp_ep_t);
-    PRINT_SIZE(ucp_ep_ext_gen_t);
-    PRINT_SIZE(ucp_ep_ext_proto_t);
+    PRINT_SIZE(ucp_ep_ext_t);
```
yosefe, line 264: what is the size?

src/ucp/core/ucp_ep.h
yosefe: yes, it would be better

src/ucp/core/ucp_request.h
```diff
@@ -447,7 +447,7 @@ struct ucp_request {
             ucp_send_nbx_callback_t cb;         /* Completion callback */
             uct_worker_cb_id_t      prog_id;    /* Progress callback ID */
             int                     comp_count; /* Countdown to request completion */
-            ucp_ep_ext_gen_t        *next_ep;   /* Next endpoint to flush */
+            ucp_ep_ext_t            *next_ep;   /* Next endpoint to flush */
```
yosefe, line 450: maybe point to ep

src/ucp/core/ucp_ep.h
gleon99: Then we'll have to forward-declare `ucp_ep`, because the `ext` has a pointer to it.

yosefe: https://github.com/openucx/ucx/blob/3d4ccf50de87c2e1652da440e3fb27b6cfa71490/src/ucp/api/ucp_def.h#L92

src/tools/info/type_info.c
gleon99: 184

src/ucp/core/ucp_request.h
gleon99: We can't do it because of how flushing flow works.

Currently, in `ucp_worker_flush_req_set_next_ep` it might set `worker->all_eps` as the "next EP ext" to flush, then [this](https://github.com/openucx/ucx/blob/master/src/ucp/rma/flush.c#L491) ot [this](https://github.com/openucx/ucx/blob/master/src/ucp/rma/flush.c#L461) condition will be false and `set_next_ep` is skipped which is OK.

If we store EP instead of EP ext, need to [dereference](https://github.com/openucx/ucx/blob/master/src/ucp/rma/flush.c#L419) from an invalid `ep_ext` back to `ep`, which can't work, because then need to deref back from that `ep` to `ep_ext`. Then, the above-mentioned `if` conditions won't check correctly -> `set_next_ep` will be reached pointing to invalid memory as "EP".

src/ucp/core/ucp_ep.h
```diff
@@ -438,6 +438,59 @@ struct ucp_ep_config {
...
+        ucs_list_link_t           ready_list;     /* List entry in worker's EP list */
+        ucs_queue_head_t          match_q;        /* Queue of receive data or requests,
+                                                     depends on UCP_EP_FLAG_STREAM_HAS_DATA */
+    } stream;
+
+    struct {
+        ucs_list_link_t           started_ams;
+        ucs_queue_head_t          mid_rdesc_q;    /* Queue of middle fragments, which
+                                                     arrived before the first one */
+    } am;
```
brminich, line 490: maybe we can store this as a pointer and allocate when `UCP_FEATURE_STREAM` or/and `UCP_FEATURE_AM`  is requested only. This way we will save 48 bytes per ep for MPI

gleon99: As discussed on the team meeting, we'll keep it flat for now.

brminich (dismissed):

gleon99:
@yosefe please review

src/ucp/core/ucp_ep.h
```diff
@@ -456,27 +456,27 @@ typedef struct ucp_ep_ext {
...
+#endif
 
+    /* Endpoint match context and remote completion status are mutually exclusive,
+     * since remote completions are counted only after the endpoint is already
+     * matched to a remote peer.
+     */
+    union {
+        ucp_ep_match_elem_t       ep_match;      /* Matching with remote endpoints */
+        ucp_ep_flush_state_t      flush_state;   /* Remote completion status */
+    };
```
yosefe, line 479: space line after

yosefe (approved):

brminich (approved):

PR #8243, opened 2022-05-18
UCT/CM: Fix destroying listener in parallel with conn_req_cb

dmitrygx:
## What

Fix case when destroying listener in parallel with conn_req_cb invocation.

## Why ?

Reproduces and fixes the case when UCT listener is destroyed in parallel with conn_req_cb invocation.

## How ?

1. Add gtest which emulates destroying of UCT listener and conn_req_cb invocation in different threads simultaneously.
2. Add `UCS_ASYNC_BLOCK`/`UCS_ASYNC_UNBLOCK` in RDMACM's and TCP_SOCKM's listener destroy functions.

### DIFF
--- a/src/uct/ib/rdmacm/rdmacm_listener.c
+++ b/src/uct/ib/rdmacm/rdmacm_listener.c
@@ -11,6 +11,7 @@
 #include "rdmacm_listener.h"
 
 #include <ucs/sys/sock.h>
+#include <ucs/async/async.h>
 
 
 #define UCS_RDMACM_MAX_BACKLOG_PATH        "/proc/sys/net/rdma_ucm/max_backlog"
@@ -128,8 +129,12 @@ ucs_status_t uct_rdmacm_listener_reject(uct_listener_h listener,
 
 UCS_CLASS_CLEANUP_FUNC(uct_rdmacm_listener_t)
 {
+    ucs_async_context_t *async = self->super.cm->iface.worker->async;
+
+    UCS_ASYNC_BLOCK(async);
     ucs_debug("listener %p: destroying rdma_cm_id %p", self, self->id);
     uct_rdmacm_cm_destroy_id(self->id);
+    UCS_ASYNC_UNBLOCK(async);
 }
 
 ucs_status_t uct_rdmacm_listener_query(uct_listener_h listener,
--- a/src/uct/tcp/tcp_listener.c
+++ b/src/uct/tcp/tcp_listener.c
@@ -145,7 +145,10 @@ err:
 
 UCS_CLASS_CLEANUP_FUNC(uct_tcp_listener_t)
 {
+    ucs_async_context_t *async = self->super.cm->iface.worker->async;
     ucs_status_t status;
+
+    UCS_ASYNC_BLOCK(async);
 
     status = ucs_async_remove_handler(self->listen_fd, 1);
     if (status != UCS_OK) {
@@ -154,6 +157,8 @@ UCS_CLASS_CLEANUP_FUNC(uct_tcp_listener_t)
     }
 
     ucs_close_fd(&self->listen_fd);
+
+    UCS_ASYNC_UNBLOCK(async);
 }
 
 ucs_status_t uct_tcp_listener_reject(uct_listener_h listener,
--- a/test/gtest/Makefile.am
+++ b/test/gtest/Makefile.am
@@ -119,6 +119,7 @@ gtest_SOURCES = \
 	ucs/test_event_set.cc \
 	ucs/test_stats_filter.cc \
 	uct/test_peer_failure.cc \
+	uct/test_sockaddr.cc \
 	uct/test_tag.cc \
 	uct/tcp/test_tcp.cc \
 	\
@@ -227,10 +228,6 @@ endif
 if HAVE_MLX5_DV
 gtest_SOURCES += \
 	uct/ib/test_cqe_zipping.cc
-endif
-if HAVE_RDMACM
-gtest_SOURCES += \
-	uct/ib/test_sockaddr.cc
 endif
 endif # HAVE_IB
 
--- a/test/gtest/uct/test_sockaddr.cc
+++ b/test/gtest/uct/test_sockaddr.cc
@@ -17,6 +17,8 @@ extern "C" {
 }
 
 #include <queue>
+#include <thread>
+#include <atomic>
 
 
 class test_uct_sockaddr : public uct_test {
@@ -35,7 +37,9 @@ protected:
         TEST_STATE_SERVER_REJECTED               = UCS_BIT(5),
         TEST_STATE_CLIENT_GOT_REJECT             = UCS_BIT(6),
         TEST_STATE_CLIENT_GOT_ERROR              = UCS_BIT(7),
-        TEST_STATE_CLIENT_GOT_SERVER_UNAVAILABLE = UCS_BIT(8)
+        TEST_STATE_CLIENT_GOT_SERVER_UNAVAILABLE = UCS_BIT(8),
+        TEST_STATE_SERVER_LISTENER_READY_CLOSE   = UCS_BIT(9),
+        TEST_STATE_SERVER_LISTENER_CLOSED        = UCS_BIT(10)
     };
 
     enum {
@@ -380,14 +384,15 @@ protected:
         }
     }
 
-    virtual void server_accept(entity *server, uct_conn_request_h conn_request,
-                               uct_cm_ep_server_conn_notify_callback_t notify_cb,
-                               uct_ep_disconnect_cb_t disconnect_cb,
-                               void *user_data, bool can_fail)
+    virtual void
+    server_accept(uct_conn_request_h conn_request,
+                  uct_cm_ep_server_conn_notify_callback_t notify_cb,
+                  uct_ep_disconnect_cb_t disconnect_cb, void *user_data,
+                  bool can_fail)
     {
         ucs::scoped_async_lock listen_lock(m_server->async());
-        ucs::scoped_async_lock accept_lock(server->async());
-        accept(server->cm(), conn_request, notify_cb, disconnect_cb,
+        ucs::scoped_async_lock accept_lock(m_server->async());
+        accept(m_server->cm(), conn_request, notify_cb, disconnect_cb,
                user_data, can_fail);
     }
 
@@ -413,12 +418,12 @@ protected:
      * return true if the server should accept the connection request and
      * false if not.
      */
-    static bool common_conn_request(uct_listener_h listener, void *arg,
-                                    const uct_cm_listener_conn_request_args_t
-                                    *conn_req_args) {
-        test_uct_sockaddr *self = reinterpret_cast<test_uct_sockaddr *>(arg);
+    bool
+    common_conn_request(uct_listener_h listener,
+                        const uct_cm_listener_conn_request_args_t *conn_req_args)
+    {
         ucs_sock_addr_t m_connect_addr_sock_addr =
-                        self->m_connect_addr.to_ucs_sock_addr();
+                m_connect_addr.to_ucs_sock_addr();
         uct_conn_request_h conn_request;
         const uct_cm_remote_data_t *remote_data;
         uint16_t client_port;
@@ -440,16 +445,17 @@ protected:
         ASSERT_UCS_OK(status);
         EXPECT_GT(client_port, 0);
 
-        self->verify_remote_data(remote_data->conn_priv_data, remote_data->conn_priv_data_length);
+        verify_remote_data(remote_data->conn_priv_data,
+                           remote_data->conn_priv_data_length);
 
-        self->m_state |= TEST_STATE_CONNECT_REQUESTED;
+        m_state |= TEST_STATE_CONNECT_REQUESTED;
 
-        if (self->m_delay_conn_reply) {
-            self->m_delayed_conn_reqs.push(conn_request);
-        } else if (self->m_reject_conn_request) {
+        if (m_delay_conn_reply) {
+            m_delayed_conn_reqs.push(conn_request);
+        } else if (m_reject_conn_request) {
             status = uct_listener_reject(listener, conn_request);
             ASSERT_UCS_OK(status);
-            self->m_state |= TEST_STATE_SERVER_REJECTED;
+            m_state |= TEST_STATE_SERVER_REJECTED;
         } else {
             /* do regular server accept */
             return true;
@@ -458,23 +464,56 @@ protected:
         return false;
     }
 
-    static void
-    conn_request_cb(uct_listener_h listener, void *arg,
-                    const uct_cm_listener_conn_request_args_t *conn_req_args) {
-        test_uct_sockaddr *self = reinterpret_cast<test_uct_sockaddr *>(arg);
-
-        if (self->common_conn_request(listener, arg, conn_req_args)) {
+    void
+    handle_conn_request(uct_listener_h listener,
+                        const uct_cm_listener_conn_request_args_t *conn_req_args,
+                        uct_cm_ep_server_conn_notify_callback_t notify_cb,
+                        uct_ep_disconnect_cb_t disconnect_cb)
+    {
+        if (common_conn_request(listener, conn_req_args)) {
             EXPECT_TRUE(conn_req_args->field_mask &
                         UCT_CM_LISTENER_CONN_REQUEST_ARGS_FIELD_CONN_REQUEST);
-            self->server_accept(self->m_server, conn_req_args->conn_request,
-                                server_connect_cb, server_disconnect_cb, self,
-                                false);
+            server_accept(conn_req_args->conn_request, notify_cb,
+                          disconnect_cb, this, false);
         }
 
-        ucs_memory_cpu_store_fence();
-        self->m_server_recv_req_cnt++;
+        m_server_recv_req_cnt++;
     }
 
+    static void
+    conn_request_cb(uct_listener_h listener, void *arg,
+                    const uct_cm_listener_conn_request_args_t *conn_req_args)
+    {
+        test_uct_sockaddr *self = static_cast<test_uct_sockaddr*>(arg);
+        self->handle_conn_request(listener, conn_req_args,
+                                  test_uct_sockaddr::server_connect_cb,
+                                  test_uct_sockaddr::server_disconnect_cb);
+    }
+
+    static void concurrent_conn_request_and_listener_close_cb(
+            uct_listener_h listener, void *arg,
+            const uct_cm_listener_conn_request_args_t *conn_req_args)
+    {
+        test_uct_sockaddr *self = static_cast<test_uct_sockaddr*>(arg);
+
+        ASSERT_FALSE(self->m_state & TEST_STATE_SERVER_LISTENER_READY_CLOSE);
+        ASSERT_FALSE(self->m_state & TEST_STATE_SERVER_LISTENER_CLOSED);
+
+        self->handle_conn_request(listener, conn_req_args,
+                                  test_uct_sockaddr::server_connect_cb,
+                                  test_uct_sockaddr::server_disconnect_cb);
+
+        // Wait until listener will be set for being destroyed
+        while (!(self->m_state & TEST_STATE_SERVER_LISTENER_READY_CLOSE)) {
+            std::this_thread::yield();
+        }
+
+        // Wait to increase the probability of trying to lock mutex during
+        // destroying while it is already acquired by this thread
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+        ASSERT_FALSE(self->m_state & TEST_STATE_SERVER_LISTENER_CLOSED);
+    }
 
     static void
     server_connect_cb(uct_ep_h ep, void *arg,
@@ -588,7 +627,6 @@ protected:
     void test_delayed_server_response(bool reject, bool early_destroy)
     {
         ucs_status_t status;
-        ucs_time_t deadline;
 
         m_delay_conn_reply = true;
 
@@ -610,14 +648,8 @@ protected:
             m_client->destroy_eps();
         }
 
-        deadline = ucs_get_time() + ucs_time_from_sec(DEFAULT_TIMEOUT_SEC) *
-                                    ucs::test_time_multiplier();
-
-        while ((m_server_recv_req_cnt == 0) && (ucs_get_time() < deadline)) {
-            progress();
-        }
+        wait_for_flag(&m_server_recv_req_cnt);
         ASSERT_EQ(1, m_server_recv_req_cnt);
-        ucs_memory_cpu_load_fence();
 
         if (reject) {
             /* wrap errors since a reject is expected */
@@ -631,9 +663,8 @@ protected:
                 EXPECT_TRUE(m_state & TEST_STATE_CLIENT_GOT_REJECT);
             }
         } else {
-            server_accept(m_server, m_delayed_conn_reqs.front(),
-                          server_connect_cb, server_disconnect_cb, this,
-                          early_destroy);
+            server_accept(m_delayed_conn_reqs.front(), server_connect_cb,
+                          server_disconnect_cb, this, early_destroy);
             if (!early_destroy) {
                 wait_for_bits(&m_state, TEST_STATE_SERVER_CONNECTED |
                                         TEST_STATE_CLIENT_CONNECTED);
@@ -740,10 +771,11 @@ private:
 
 protected:
     ucs::sock_addr_storage m_listen_addr, m_connect_addr;
-    uint64_t               m_state;
+    std::atomic<uint64_t>  m_state;
     entity                 *m_server;
     entity                 *m_client;
-    volatile int           m_server_recv_req_cnt, m_client_connect_cb_cnt,
+    std::atomic<int>       m_server_recv_req_cnt;
+    volatile int           m_client_connect_cb_cnt,
                            m_server_connect_cb_cnt;
     volatile int           m_server_disconnect_cnt, m_client_disconnect_cnt;
     bool                   m_reject_conn_request;
@@ -1040,6 +1072,34 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
     }
 }
 
+
+// The test calls uct_listener_destroy() from the main thread at the same time
+// conn_request_cb is invoked from the async thread, to make sure this is
+// properly handled
+UCS_TEST_P(test_uct_sockaddr, concurrent_handle_conn_req_and_close_listener)
+{
+    m_reject_conn_request = true;
+
+    start_listen(concurrent_conn_request_and_listener_close_cb);
+    connect(*m_client, 0, client_disconnect_cb, nullptr);
+
+    while (!m_server_recv_req_cnt) {
+        std::this_thread::yield();
+    }
+
+    m_state |= TEST_STATE_SERVER_LISTENER_READY_CLOSE;
+    uct_listener_destroy(m_server->revoke_listener());
+    m_state |= TEST_STATE_SERVER_LISTENER_CLOSED;
+
+    wait_for_bits(&m_state, TEST_STATE_SERVER_REJECTED |
+                            TEST_STATE_CLIENT_GOT_REJECT);
+    EXPECT_TRUE(m_state & TEST_STATE_SERVER_REJECTED);
+    EXPECT_TRUE(m_state & TEST_STATE_CLIENT_GOT_REJECT);
+
+    EXPECT_FALSE(m_state & TEST_STATE_SERVER_CONNECTED);
+    EXPECT_FALSE(m_state & TEST_STATE_CLIENT_CONNECTED);
+}
+
 UCT_INSTANTIATE_SOCKADDR_TEST_CASE(test_uct_sockaddr)
 
 
@@ -1171,33 +1231,26 @@ public:
         self->del_user_data(sa_user_data);
     }
 
-    void server_accept(entity *server, uct_conn_request_h conn_request,
+    void server_accept(uct_conn_request_h conn_request,
                        uct_cm_ep_server_conn_notify_callback_t notify_cb,
                        uct_ep_disconnect_cb_t disconnect_cb,
                        void *user_data, bool can_fail) {
         ucs::scoped_async_lock listen_lock(m_server->async());
-        ucs::scoped_async_lock accept_lock(server->async());
-        test_uct_sockaddr::accept(server->cm(), conn_request, notify_cb,
+        ucs::scoped_async_lock accept_lock(m_server->async());
+        test_uct_sockaddr::accept(m_server->cm(), conn_request, notify_cb,
                                   disconnect_cb, user_data, can_fail);
     }
 
     static void
     conn_request_cb(uct_listener_h listener, void *arg,
-                    const uct_cm_listener_conn_request_args_t *conn_req_args) {
+                    const uct_cm_listener_conn_request_args_t *conn_req_args)
+    {
         test_uct_sockaddr_stress *self =
-                        reinterpret_cast<test_uct_sockaddr_stress *>(arg);
-
-        if (test_uct_sockaddr::common_conn_request(listener, arg,
-                                                   conn_req_args)) {
-            EXPECT_TRUE(conn_req_args->field_mask &
-                        UCT_CM_LISTENER_CONN_REQUEST_ARGS_FIELD_CONN_REQUEST);
-            self->server_accept(self->m_server, conn_req_args->conn_request,
-                                server_connect_cb, server_disconnect_cb, self,
-                                false);
-        }
-
-        ucs_memory_cpu_store_fence();
-        self->m_server_recv_req_cnt++;
+                static_cast<test_uct_sockaddr_stress *>(arg);
+        self->handle_conn_request(
+                listener, conn_req_args,
+                test_uct_sockaddr_stress::server_connect_cb,
+                test_uct_sockaddr_stress::server_disconnect_cb);
     }
 
     static void
@@ -1363,7 +1416,7 @@ public:
         test_uct_sockaddr::cleanup();
     }
 
-    void server_accept(entity *server, uct_conn_request_h conn_request,
+    void server_accept(uct_conn_request_h conn_request,
                        uct_cm_ep_server_conn_notify_callback_t notify_cb,
                        uct_ep_disconnect_cb_t disconnect_cb,
                        void *user_data, bool can_fail)
--- a/test/gtest/uct/uct_test.cc
+++ b/test/gtest/uct/uct_test.cc
@@ -1099,6 +1099,12 @@ uct_listener_h uct_test::entity::listener() const {
     return m_listener;
 }
 
+uct_listener_h uct_test::entity::revoke_listener() const {
+    uct_listener_h uct_listener = listener();
+    m_listener.revoke();
+    return uct_listener;
+}
+
 uct_iface_h uct_test::entity::iface() const {
     return m_iface;
 }
--- a/test/gtest/uct/uct_test.h
+++ b/test/gtest/uct/uct_test.h
@@ -20,7 +20,7 @@
 #include <common/mem_buffer.h>
 #include <common/test.h>
 #include <vector>
-
+#include <atomic>
 
 
 #define DEFAULT_DELAY_MS           1.0
@@ -167,6 +167,8 @@ protected:
         const uct_cm_attr_t& cm_attr() const;
 
         uct_listener_h listener() const;
+
+        uct_listener_h revoke_listener() const;
 
         uct_iface_h iface() const;
 
@@ -319,7 +321,8 @@ protected:
         }
     }
 
-    void wait_for_bits(volatile uint64_t *flag, uint64_t mask,
+    template <typename FlagType, typename MaskType>
+    void wait_for_bits(FlagType *flag, MaskType mask,
                        double timeout = DEFAULT_TIMEOUT_SEC) const
     {
         ucs_time_t deadline = ucs_get_time() +


test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1060,90 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+        ASSERT_TRUE(m_server_uct_listener == NULL);
+
+        // Notify listener_close_thread to destroy the listener
+        {
+            std::unique_lock<std::mutex> l(m_notify_mutex);
+            m_notify_cv.notify_one();
+        }
+
+        // Wait until the listener is set from listener_close_thread and check
+        // whether the server's listener is in good shapre
```
ivankochin, line 1092:
```suggestion
        // whether the server's listener is in good shape
```
Typo?

dmitrygx: thanks, rephrased

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1059,95 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+                test_uct_sockaddr::server_disconnect_cb);
+
+        ASSERT_TRUE(m_server_uct_listener == NULL);
+
+        // Notify listener_close_thread to destroy the listener
+        m_close_listener_notify = true;
+
+        // Wait until the listener is set from listener_close_thread and check
+        // whether the server's listener is set or not
+        self->wait_for_flag(&m_server_uct_listener);
```
ivankochin, line 1090: Consider making `m_server_uct_listener` an atomic since formally it is a race condition.

dmitrygx:
@brminich @evgeny-leksikov could you review pls?

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1061,100 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+                test_uct_sockaddr::server_disconnect_cb);
+
+        ASSERT_EQ(nullptr, m_server_uct_listener);
+
+        // Notify listener_close_thread to destroy the listener
+        m_close_listener_notify = true;
+
+        // Wait until uct_listener_destroy is being called from
+        // listener_close_thread and dereference the server's listener to check
+        // whether the
```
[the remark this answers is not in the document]
dmitrygx: fixed

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1061,100 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+        // Wait until uct_listener_destroy is being called from
+        // listener_close_thread and dereference the server's listener to check
+        // whether the 
+        while (!m_server_uct_listener) {
+            std::this_thread::yield();
+        }
+
+        uct_listener_h uct_listener = m_server_uct_listener;
+        ASSERT_TRUE(uct_listener != nullptr);
+        ASSERT_TRUE(uct_listener->cm != nullptr);
```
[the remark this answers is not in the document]
dmitrygx: it is leftover from the previous solution, fixed

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1183,21 +1295,15 @@ class test_uct_sockaddr_stress : public test_uct_sockaddr {
...
-            EXPECT_TRUE(conn_req_args->field_mask &
-                        UCT_CM_LISTENER_CONN_REQUEST_ARGS_FIELD_CONN_REQUEST);
-            self->server_accept(self->m_server, conn_req_args->conn_request,
-                                server_connect_cb, server_disconnect_cb, self,
-                                false);
-        }
-
-        ucs_memory_cpu_store_fence();
-        self->m_server_recv_req_cnt++;
+                reinterpret_cast<test_uct_sockaddr_stress *>(arg);
```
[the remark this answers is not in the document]
dmitrygx: done

dmitrygx:
@ivankochin is it ok now?

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1061,98 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+    static std::atomic<bool>           m_close_listener_notify;
+    static std::atomic<uct_listener_h> m_server_uct_listener;
+};
+
+
+std::atomic<bool> test_uct_sockaddr_listener::m_close_listener_notify(false);
+std::atomic<uct_listener_h> test_uct_sockaddr_listener::m_server_uct_listener;
+
+
+UCS_TEST_P(test_uct_sockaddr_listener, handle_conn_req_and_close_listener_in_parallel)
```
evgeny-leksikov, line 1122:
```suggestion
UCS_TEST_P(test_uct_sockaddr_listener, concurrent_handle_conn_req_and_close_listener)
```

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1059,95 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+};
+
+
+std::atomic<bool> test_uct_sockaddr_listener::m_close_listener_notify(false);
+uct_listener_h test_uct_sockaddr_listener::m_server_uct_listener = NULL;
+
+
+UCS_TEST_P(test_uct_sockaddr_listener, handle_conn_req_and_close_listener_in_parallel)
+{
+    m_reject_conn_request = true;
```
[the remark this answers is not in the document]
evgeny-leksikov: +1 to move it to ctor untill we don't have such test

evgeny-leksikov (commented):
aslo seems like need to increase `UCS_ASYNC_BLOCK` scope in `ucp_listener_destroy`

test/gtest/uct/test_sockaddr.cc
```diff
@@ -458,28 +462,42 @@ class test_uct_sockaddr : public uct_test {
         return false;
     }
 
-    static void
-    conn_request_cb(uct_listener_h listener, void *arg,
-                    const uct_cm_listener_conn_request_args_t *conn_req_args) {
-        test_uct_sockaddr *self = reinterpret_cast<test_uct_sockaddr *>(arg);
-
-        if (self->common_conn_request(listener, arg, conn_req_args)) {
+    void server_handle_conn_request(
```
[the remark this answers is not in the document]
dmitrygx: removed it from the name

test/gtest/uct/test_sockaddr.cc
```diff
@@ -458,28 +462,42 @@ class test_uct_sockaddr : public uct_test {
...
+            uct_cm_ep_server_conn_notify_callback_t notify_cb,
+            uct_ep_disconnect_cb_t disconnect_cb)
+    {
+        if (common_conn_request(listener, conn_req_args)) {
             EXPECT_TRUE(conn_req_args->field_mask &
                         UCT_CM_LISTENER_CONN_REQUEST_ARGS_FIELD_CONN_REQUEST);
-            self->server_accept(self->m_server, conn_req_args->conn_request,
-                                server_connect_cb, server_disconnect_cb, self,
-                                false);
+            server_accept(m_server, conn_req_args->conn_request, notify_cb,
```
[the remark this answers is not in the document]
dmitrygx: we use `m_` prefix in other tests too

test/gtest/uct/test_sockaddr.cc
```diff
@@ -458,28 +462,42 @@ class test_uct_sockaddr : public uct_test {
...
+                listener, conn_req_args,
+                test_uct_sockaddr::server_connect_cb,
+                test_uct_sockaddr::server_disconnect_cb);
+        self->server_conn_req_recv();
+    }
 
     static void
     server_connect_cb(uct_ep_h ep, void *arg,
                       const uct_cm_ep_server_conn_notify_args_t *notify_args) {
-        test_uct_sockaddr *self = reinterpret_cast<test_uct_sockaddr *>(arg);
```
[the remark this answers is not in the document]
dmitrygx: removed

test/gtest/uct/test_sockaddr.cc
```diff
@@ -458,28 +462,42 @@ class test_uct_sockaddr : public uct_test {
...
                         UCT_CM_LISTENER_CONN_REQUEST_ARGS_FIELD_CONN_REQUEST);
-            self->server_accept(self->m_server, conn_req_args->conn_request,
-                                server_connect_cb, server_disconnect_cb, self,
-                                false);
+            server_accept(m_server, conn_req_args->conn_request, notify_cb,
+                          disconnect_cb, this, false);
         }
+    }
 
+    void server_conn_req_recv()
```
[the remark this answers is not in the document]
dmitrygx: 1. incrementing of `m_server_recv_req_cnt` must be done a bit later in new `conn_req_cb`,
2. renamed

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1061,98 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+class test_uct_sockaddr_listener : public test_uct_sockaddr
+{
+protected:
+    test_uct_sockaddr_listener()
+    {
+        m_close_listener_notify = false;
+        m_server_uct_listener   = nullptr;
+    }
+
+    static void conn_request_in_parallel_with_listener_close_cb(
```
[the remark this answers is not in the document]
dmitrygx: could be just `conn_request_cb`

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1061,98 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+
+        self->server_handle_conn_request(
+                listener, conn_req_args,
+                test_uct_sockaddr::server_connect_cb,
+                test_uct_sockaddr::server_disconnect_cb);
+
+        ASSERT_EQ(nullptr, m_server_uct_listener);
+
+        // Notify listener_close_thread to destroy the listener
+        m_close_listener_notify = true;
```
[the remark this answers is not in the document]
dmitrygx: not, it is a good time

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1061,98 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+
+        // Wait until uct_listener_destroy is being called from
+        // listener_close_thread and dereference the server's listener to check
+        // whether the object is valid or not
+        while (m_server_uct_listener == nullptr) {
+            std::this_thread::yield();
+        }
+
+        ASSERT_TRUE(m_server_uct_listener.load()->cm != nullptr);
```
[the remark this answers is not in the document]
dmitrygx: why is it incorrect?

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1061,98 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+            std::this_thread::yield();
+        }
+
+        self->m_server_uct_listener = self->m_server->revoke_listener();
+        uct_listener_destroy(self->m_server_uct_listener);
+    }
+
+protected:
+    static std::atomic<bool>           m_close_listener_notify;
+    static std::atomic<uct_listener_h> m_server_uct_listener;
```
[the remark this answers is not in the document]
dmitrygx: why needed?

test/gtest/uct/test_sockaddr.cc
dmitrygx: done

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1061,98 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+    connect(*m_client, 0, client_disconnect_cb, nullptr);
+
+    ucs_time_t deadline = ucs_get_time() +
+                          ucs_time_from_sec(DEFAULT_TIMEOUT_SEC) *
+                          ucs::test_time_multiplier();
+
+    while ((m_server_recv_req_cnt == 0) && (ucs_get_time() < deadline)) {
+        progress();
+    }
+    ASSERT_EQ(1, m_server_recv_req_cnt);
```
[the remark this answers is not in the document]
dmitrygx: done

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1061,98 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+
+    // Wait listener_close_thread to be completed
+    close_thread.join();
+
+    wait_for_bits(&m_state, TEST_STATE_SERVER_REJECTED |
+                            TEST_STATE_CLIENT_GOT_REJECT);
+    EXPECT_TRUE(ucs_test_all_flags(m_state, (TEST_STATE_SERVER_REJECTED |
+                                             TEST_STATE_CLIENT_GOT_REJECT)));
+
+    EXPECT_FALSE((m_state &
```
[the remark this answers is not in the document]
dmitrygx: removed

dmitrygx:
> aslo seems like need to increase UCS_ASYNC_BLOCK scope in ucp_listener_destroy

let's do it in another PR

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
evgeny-leksikov: sorry for missleading, I ment no need to pass class member to its non-static method

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
evgeny-leksikov: can we avoid adding new class and add the test case simpler?

evgeny-leksikov:
> > aslo seems like need to increase UCS_ASYNC_BLOCK scope in ucp_listener_destroy
> 
> @evgeny-leksikov good catch, let's do it in another PR

👍 it would be great if it possible to reproduce the race with UCP test case too

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
dmitrygx: done

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
evgeny-leksikov: maybe I don't understant the test's logic completely. but at this point the connection request has been already handled in `handle_conn_request` above, so the is no guarantee what is exact value of internal fields
this comment is linked to https://github.com/openucx/ucx/pull/8243/files#r879772858

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
dmitrygx: > can we avoid adding new class and add the test case simpler?

done

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
dmitrygx: the idea of the test that UCT listner shouldn't be closed when user's `conn_request_cb` was invoked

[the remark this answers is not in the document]
dmitrygx: if the UCT listener was destroyed, `m_server_uct_listener.load()->cm != nullptr` check will cause invalid pointer dereference which could lead to UB or will be detected by Valgrind.

test/gtest/uct/test_sockaddr.cc
```diff
@@ -476,7 +476,7 @@ class test_uct_sockaddr : public uct_test {
         }
     }
 
-    void server_conn_req_recv()
+    void conn_req_recv_cnt_inc()
```
evgeny-leksikov, line 481: IIUC at this point `con_req` is not only recvd but also handeled, I would rename to `m_server_recv_req_cnt` too since its logic was changed

test/gtest/uct/test_sockaddr.cc
```diff
@@ -755,9 +810,16 @@ class test_uct_sockaddr : public uct_test {
     std::vector<char>      m_long_priv_data;
     pthread_mutex_t        m_ep_client_data_lock;
     ep_client_data_map_t   m_ep_client_data;
+
+    static std::atomic<bool>           m_close_listener_notify;
+    static std::atomic<uct_listener_h> m_server_uct_listener;
```
evgeny-leksikov, line 815: but these are still here... My point was if we can do the same with existing members, but probaly convert counters to atomics

test/gtest/uct/test_sockaddr.cc
dmitrygx: but it is currently named as `m_server_recv_req_cnt`

test/gtest/uct/test_sockaddr.cc
dmitrygx: unfortunately we can't, because these static members are used to synchronize new callback and thread

test/gtest/uct/test_sockaddr.cc
dmitrygx: do you mean to rename it to `m_server_conn_req_handle_cnt`?

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
evgeny-leksikov: updated code does not add new class but it's not simpler, it adds new members and methods to class where it's not needed , if it's impossible to avoid new members, beter to have them in separate class as well as corresponding methods

test/gtest/uct/test_sockaddr.cc
evgeny-leksikov: new name depends on what this counter means, receiving and handling were in the same cb/function before your change

dmitrygx: but now they are always in the same function.
the called at the end of the `conn_req_cb`

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
dmitrygx: yes, it is impossible to not have them.
moved to separate class as it was

test/gtest/uct/test_sockaddr.cc
```diff
@@ -729,6 +742,43 @@ class test_uct_sockaddr : public uct_test {
...
+        // Wait until uct_listener_destroy is being called from
+        // listener_close_thread and dereference the server's listener to check
+        // whether the object is valid or not
+        while (m_server_uct_listener == nullptr) {
+            std::this_thread::yield();
+        }
+
+        ASSERT_TRUE(m_server_uct_listener.load()->cm != nullptr);
+
+        self->conn_req_recv_cnt_inc();
```
evgeny-leksikov: why need to delay this call? I dont thisk this counter is critical for this test

dmitrygx: yes, you're right - we don't need to delay incrementing this counter.
so, returned to the way as it was

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1049,87 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+        while (!m_close_listener_notify) {
+            std::this_thread::yield();
+        }
+
+        self->m_server_uct_listener = self->m_server->revoke_listener();
+        uct_listener_destroy(self->m_server_uct_listener);
+    }
+protected:
+    static std::atomic<bool>           m_close_listener_notify;
+    static std::atomic<uct_listener_h> m_server_uct_listener;
```
ivankochin, line 1097: WDYTA making this variables non-static?

dmitrygx: done, accessed through `self`

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1049,85 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+            std::this_thread::yield();
+        }
+
+        self->m_server_uct_listener = self->m_server->revoke_listener();
+        uct_listener_destroy(self->m_server_uct_listener);
+    }
+
+protected:
+    std::atomic<bool>           m_close_listener_notify;
+    std::atomic<uct_listener_h> m_server_uct_listener;
```
ivankochin, line 1098:
```suggestion
    std::atomic<bool>           m_close_listener_notify{false};
    std::atomic<uct_listener_h> m_server_uct_listener{nullptr};
```

If both these values are value-initialized, then compiler-generated default ctor will have the same behavior as the ctor defined above. So manual-defined ctor can be removed in this case.

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
dmitrygx: not needed anymore

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1049,85 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+    {
+        while (!self->m_close_listener_notify) {
+            std::this_thread::yield();
+        }
+
+        self->m_server_uct_listener = self->m_server->revoke_listener();
+        uct_listener_destroy(self->m_server_uct_listener);
+    }
+
+protected:
```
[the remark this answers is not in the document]
dmitrygx: removed

test/gtest/uct/test_sockaddr.cc
dmitrygx: if not initialize `std::atomic`'s, they contain garbage

dmitrygx:
@ivankochin @evgeny-leksikov could you review pls?

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1048,76 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+                                  test_uct_sockaddr::server_connect_cb,
+                                  test_uct_sockaddr::server_disconnect_cb);
+
+        // Wait until listener will be set for being destroyed
+        while (self->m_listener_to_close == nullptr) {
+            std::this_thread::yield();
+        }
+
+        // Wait for 0.5s to give a chance for destroying the listener
+        std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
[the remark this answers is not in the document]
dmitrygx: Helps only allocating large piece of memory, which takes >= than 0.5s.
Just doing yield several times doesn’t help.
So, I decided to use sleep(), because it has predictable time limit and 0.5s gives good reproducibility. It may be not reproduced, but in my experiments, it was reproduced every time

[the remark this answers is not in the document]
evgeny-leksikov: It seems weird to me, uct_listener_destroy can't take 0.5s. maybe do smth like below?
```
for (int i = 0; i < 10; ++i) {
    std::vector<uint8_t> v(i*page_size, 0);
    std::this_thread::yield();
    std::fill(v.begin(), v.end(). 1);
}
```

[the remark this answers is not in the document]
dmitrygx: > It seems weird to me, uct_listener_destroy can't take 0.5s

It is approximately.

> maybe do smth like below?

I think it is better to do `sleep()`. It is more obvious

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1048,76 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+            std::this_thread::yield();
+        }
+
+        self->m_listener_to_close = self->m_server->revoke_listener();
+        uct_listener_destroy(self->m_listener_to_close);
+        self->m_listener_closed = true;
+    }
+
+    std::atomic<bool>           m_listener_closed{false};
+    std::atomic<uct_listener_h> m_listener_to_close{nullptr};
```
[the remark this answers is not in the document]
dmitrygx: done

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
evgeny-leksikov: I'm afraid sleep for 0.5s will significantly increase regular CI testing time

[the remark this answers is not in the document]
dmitrygx: Running 27 tests took 13.6 seconds on `swx-ucx`: `27 tests from tcp/test_uct_sockaddr_listener (13638 ms total)`
I don't think that it is critical.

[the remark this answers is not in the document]
dmitrygx: Decreased time from 500ms to 100ms and it also works fine, so now we get:
`27 tests from tcp/test_uct_sockaddr_listener (2833 ms total)`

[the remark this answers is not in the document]
evgeny-leksikov: I still disagree with sleep in tests but let's go with it for now and ask other reviewers.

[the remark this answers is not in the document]
dmitrygx: > I still disagree with sleep in tests but let's go with it for now and ask other reviewers.

ok, let's see what other reviewers will propose

test/gtest/uct/uct_test.cc
```diff
@@ -1099,6 +1099,12 @@ uct_listener_h uct_test::entity::listener() const {
     return m_listener;
 }
 
+uct_listener_h uct_test::entity::revoke_listener() const {
+    uct_listener_h uct_listener = listener();
+    m_listener.revoke();
+    return uct_listener;
+}
```
evgeny-leksikov, line 1106: no need any more

test/gtest/uct/uct_test.h
```diff
@@ -319,7 +321,14 @@ class uct_test : public testing::TestWithParam<const resource*>,
         }
     }
 
-    void wait_for_bits(volatile uint64_t *flag, uint64_t mask,
+    void wait_for_bits(uint64_t *flag, uint64_t mask,
+                       double timeout = DEFAULT_TIMEOUT_SEC) const
+    {
+        std::atomic<uint64_t> val(*flag);
+        wait_for_bits(&val, mask, timeout);
```
evgeny-leksikov, line 328: how does it work with copy?

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
dmitrygx: no, we should reject

test/gtest/uct/uct_test.cc
dmitrygx: it is needed

test/gtest/uct/uct_test.h
dmitrygx: can be removed, it is not used at all

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
evgeny-leksikov: ok

test/gtest/uct/uct_test.cc
dmitrygx: it can't be done, we don't have an access to `m_listener` of `entity` which is defined as `private`.

evgeny-leksikov: ic

test/gtest/uct/test_sockaddr.cc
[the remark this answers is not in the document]
ivankochin: Noone like sleeps in test but I cannot imagine way to make this test deterministic except patching mutexes implementation.
But I think this way doesn't worth implementation now. Agree to leave it as is for now.

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1050,68 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+        self->handle_conn_request(listener, conn_req_args,
+                                  test_uct_sockaddr::server_connect_cb,
+                                  test_uct_sockaddr::server_disconnect_cb);
+
+        // Wait until listener will be set for being destroyed
+        while (!(self->m_state & TEST_STATE_SERVER_LISTENER_READY_CLOSE)) {
+            std::this_thread::yield();
+        }
+
+        // Wait for 100ms to give a chance for destroying the listener
```
ivankochin, line 1076:
```suggestion
        // Wait to increase the probability of trying to lock mutex during destroying while it is already acquired by this thread
```
Minor, just a proposal that sounds more informative for me

ivankochin (approved):

test/gtest/uct/test_sockaddr.cc
dmitrygx: done

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1050,69 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
 UCT_INSTANTIATE_SOCKADDR_TEST_CASE(test_uct_sockaddr)
 
 
+class test_uct_sockaddr_listener : public test_uct_sockaddr {
```
evgeny-leksikov, line 1053: IMO no need new class

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1043,6 +1050,69 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+                                  test_uct_sockaddr::server_disconnect_cb);
+
+        // Wait until listener will be set for being destroyed
+        while (!(self->m_state & TEST_STATE_SERVER_LISTENER_READY_CLOSE)) {
+            std::this_thread::yield();
+        }
+
+        // Wait to increase the probability of trying to lock mutex during
+        // destroying while it is already acquired by this thread
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
evgeny-leksikov, line 1078: I hope for smth better

test/gtest/uct/test_sockaddr.cc
```diff
@@ -380,14 +384,14 @@ class test_uct_sockaddr : public uct_test {
...
-    virtual void server_accept(entity *server, uct_conn_request_h conn_request,
+    virtual void server_accept(uct_conn_request_h conn_request,
                                uct_cm_ep_server_conn_notify_callback_t notify_cb,
                                uct_ep_disconnect_cb_t disconnect_cb,
                                void *user_data, bool can_fail)
     {
         ucs::scoped_async_lock listen_lock(m_server->async());
-        ucs::scoped_async_lock accept_lock(server->async());
-        accept(server->cm(), conn_request, notify_cb, disconnect_cb,
+        ucs::scoped_async_lock accept_lock(m_server->async());
```
evgeny-leksikov, line 393: something wrong here, AFAIR double locking was added because listen and accept servers were different for ep migration test

dmitrygx: I'll not investigate the problem of double locking in this PR

test/gtest/uct/test_sockaddr.cc
dmitrygx: removed

test/gtest/uct/test_sockaddr.cc
dmitrygx: currently, we don't have a good solution to guarantee 100% reproducibility

test/gtest/uct/test_sockaddr.cc
evgeny-leksikov: changing of virtual method semantic should go thru all the hierarchy, otherwise this is not what was initially expected

dmitrygx: I changed `server_accept` for all the hierarchy. It is initially defined in `test_uct_sockaddr`.

evgeny-leksikov: I see change in 1 from 3 implementations

evgeny-leksikov: I missed 2 others, so need to refactor in separate PR

dmitrygx:

brminich (approved):

src/uct/tcp/tcp_listener.c
```diff
@@ -147,13 +147,17 @@ UCS_CLASS_CLEANUP_FUNC(uct_tcp_listener_t)
 {
     ucs_status_t status;
 
+    UCS_ASYNC_BLOCK(self->super.cm->iface.worker->async);
```
yosefe, line 150: maybe use helper var for worker or async context?

src/uct/ib/rdmacm/rdmacm_listener.c
```diff
@@ -128,8 +129,10 @@ ucs_status_t uct_rdmacm_listener_reject(uct_listener_h listener,
 
 UCS_CLASS_CLEANUP_FUNC(uct_rdmacm_listener_t)
 {
+    UCS_ASYNC_BLOCK(self->super.cm->iface.worker->async);
```
yosefe, line 132: maybe use helper var for worker or async context?

test/gtest/Makefile.am
```diff
@@ -228,10 +229,6 @@ if HAVE_MLX5_DV
 gtest_SOURCES += \
 	uct/ib/test_cqe_zipping.cc
 endif
-if HAVE_RDMACM
-gtest_SOURCES += \
-	uct/ib/test_sockaddr.cc
-endif
```
yosefe, line 234: wow...

test/gtest/uct/test_sockaddr.cc
```diff
@@ -413,12 +417,13 @@ class test_uct_sockaddr : public uct_test {
...
      * false if not.
      */
-    static bool common_conn_request(uct_listener_h listener, void *arg,
-                                    const uct_cm_listener_conn_request_args_t
-                                    *conn_req_args) {
-        test_uct_sockaddr *self = reinterpret_cast<test_uct_sockaddr *>(arg);
+    bool
+    common_conn_request(
+            uct_listener_h listener,
+            const uct_cm_listener_conn_request_args_t *conn_req_args)
```
yosefe, line 423: did this pass clang-format?

test/gtest/uct/test_sockaddr.cc
```diff
@@ -458,28 +464,63 @@ class test_uct_sockaddr : public uct_test {
...
 
+    static void concurrent_conn_request_and_listener_close_cb(
+            uct_listener_h listener, void *arg,
+            const uct_cm_listener_conn_request_args_t *conn_req_args)
+    {
+        test_uct_sockaddr *self = static_cast<test_uct_sockaddr*>(arg);
+
+        ASSERT_FALSE(ucs_test_all_flags(
+                self->m_state,
+                (TEST_STATE_SERVER_LISTENER_READY_CLOSE |
```
yosefe, line 501: extra ( )

test/gtest/uct/test_sockaddr.cc
```diff
@@ -740,10 +773,11 @@ class test_uct_sockaddr : public uct_test {
 
 protected:
     ucs::sock_addr_storage m_listen_addr, m_connect_addr;
-    uint64_t               m_state;
+    std::atomic<uint64_t>  m_state;
     entity                 *m_server;
     entity                 *m_client;
-    volatile int           m_server_recv_req_cnt, m_client_connect_cb_cnt,
+    std::atomic<int>       m_server_recv_req_cnt;
```
yosefe, line 779: cool

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1040,6 +1074,34 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
     }
 }
 
+
+// The idea of the test is to check that execution of a user's conn_request_cb
+// could be done simultaneously with uct_listener_destroy() calling from
+// another thread
```
yosefe, line 1080: The test calls uct_listener_destroy() from the main thread at the same time conn_request_cb is invoked from the async thread, to make sure this is properly handled

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1040,6 +1074,34 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+        std::this_thread::yield();
+    }
+
+    m_state |= TEST_STATE_SERVER_LISTENER_READY_CLOSE;
+    uct_listener_destroy(m_server->revoke_listener());
+    m_state |= TEST_STATE_SERVER_LISTENER_CLOSED;
+
+    wait_for_bits(&m_state, TEST_STATE_SERVER_REJECTED |
+                            TEST_STATE_CLIENT_GOT_REJECT);
+    EXPECT_TRUE(ucs_test_all_flags(m_state, (TEST_STATE_SERVER_REJECTED |
```
yosefe, line 1098: 1. extra ()
2. maybe when we use EXPECT_TRUE, could test each flag separately, instead of test_all_flags . it would help pinpoint the problem (in multiple places). same for EXPECT_FALSE.

test/gtest/uct/uct_test.h
```diff
@@ -319,7 +321,7 @@ class uct_test : public testing::TestWithParam<const resource*>,
         }
     }
 
-    void wait_for_bits(volatile uint64_t *flag, uint64_t mask,
+    void wait_for_bits(std::atomic<uint64_t> *flag, uint64_t mask,
```
yosefe, line 324: maybe make it templated so it could be any integral type? it's inline anyway..

test/gtest/uct/test_sockaddr.cc
```diff
@@ -1040,6 +1072,34 @@ UCS_TEST_P(test_uct_sockaddr, ep_disconnect_err_codes)
...
+    m_state |= TEST_STATE_SERVER_LISTENER_READY_CLOSE;
+    uct_listener_destroy(m_server->revoke_listener());
+    m_state |= TEST_STATE_SERVER_LISTENER_CLOSED;
+
+    wait_for_bits(&m_state, TEST_STATE_SERVER_REJECTED |
+                            TEST_STATE_CLIENT_GOT_REJECT);
+    EXPECT_TRUE(m_state & TEST_STATE_SERVER_REJECTED);
+    EXPECT_TRUE(m_state & TEST_STATE_CLIENT_GOT_REJECT);
+
+    EXPECT_FALSE(m_state & (TEST_STATE_SERVER_CONNECTED |
```
yosefe, line 1099: break it down to 2 EXPECT_FALSE statements

yosefe (dismissed):

yosefe (approved):

PR #8246, opened 2022-05-19
UCX: Adding support for RISC-V architecture; RV64G

jleidel:
## What
Adding support for the RISC-V architecture to OpenUCX.  This includes support for generic RV64G (IMAFDC) micro architectures.  Core development by @ct-clmsn and @jleidel

## Why ?
Adds support for RISC-V architecture across all subcomponents.  This includes support for basic build infrastructure, tooling, ucg, ucm, ucm/bistro, ucp, ucs and uct.  Tested using the in-situ test suite on actual RISC-V RV64G hardware in a cluster configuration (see below for configuration details). 

## How ?
All development and testing was performed using the following parameters: 
- SiFive HiFive Unmatched boards
- Ubuntu 21.04
- GCC 10.3.0


shamisp:
@yosefe - @jleidel 's company has signed CLA.

src/ucm/bistro/bistro.c
```diff
@@ -53,7 +54,13 @@ ucs_status_t ucm_bistro_apply_patch(void *dst, void *patch, size_t len)
         return status;
     }
 
+#if defined(__riscv)
+    __sync_synchronize();
```
yosefe, line 58: maybe sync on all platforms? or use memory barriers macros?
(to avoid if defined here)

src/ucm/util/log.c
```diff
@@ -168,7 +170,18 @@ static void ucm_log_vsnprintf(char *buf, size_t max, const char *fmt, va_list ap
                 if (!value.s) {
                     value.s = "(null)";
                 }
+#if defined(__riscv)    
+               trace_idx = strstr(value.s, "TRACE") - value.s;
+               if(trace_idx > 0) {
+                    pad -= strnlen(value.s, trace_idx+6);
+               }
+               else if (trace_idx == 0) {
```
yosefe, line 178: there are few code style issues, pls see https://github.com/openucx/ucx/wiki/Code-style-checking

src/ucs/arch/riscv/atomic.h
```diff
@@ -0,0 +1,67 @@
...
+* Copyright (C) Tactical Computing Labs, LLC. 2022. ALL RIGHTS RESERVED.
+*
+* See file LICENSE for terms.
+*/
+
+#ifndef UCS_RISCV64_ATOMIC_H_
+#define UCS_RISCV64_ATOMIC_H_
+
+#define UCS_DEFINE_ATOMIC_ADD(wordsize, suffix) \
+    static inline void ucs_atomic_add##wordsize(volatile uint##wordsize##_t *ptr, \
```
yosefe, line 11: do we really want to define these, or could rely on compiler __builtin_XX constructs?

src/ucs/arch/riscv/bitops.h
```diff
@@ -0,0 +1,48 @@
+/**
+* Copyright (C) Tactical Computing Labs, LLC. 2022. ALL RIGHTS RESERVED.
+*
+* See file LICENSE for terms.
+*/
+
+#ifndef UCS_RV64_BITOPS_H_
```
yosefe, line 7: UCS_RISCV_BITOPS_
let's use RISCV name throughout the code and not RV64

src/ucs/arch/riscv/cpu.c
```diff
@@ -0,0 +1,72 @@
...
+    char value[256] = {0};
+    FILE* f;
+
+    f = fopen("/proc/cpuinfo","r");
+    if (!f) {
+        return;
+    }
+
+    while (fgets(buf, sizeof(buf), f)) {
+        if (sscanf(buf, "uarch           : %s", value) == 1) {
```
yosefe, line 29: no need so many spaces before :

src/ucs/arch/riscv/cpu.c
```diff
@@ -0,0 +1,72 @@
...
+                memcpy(cpuid->uarch, value, 256); //printf("%s\n", value);
```
yosefe, line 30: 1. remove comments
2. use strcpy and sizeof(cpuid->uarch) instead of 256

src/ucs/arch/riscv/cpu.c
```diff
@@ -0,0 +1,72 @@
...
+
+        if (count == 3) {
+            break;
+        }
+    }
+
+    fclose(f);
+}
+
+void ucs_rv64_cpuid(ucs_rv64_cpuid_t *cpuid)
```
yosefe, line 49: replace rv64 by riscv in all places

src/ucs/arch/riscv/cpu.h
```diff
@@ -0,0 +1,105 @@
...
+
+static inline double ucs_arch_get_clocks_per_sec()
+{
+    uint64_t freq;
+    asm volatile("rdcycle %0" : "=r" (freq));
+    return (double) freq;
+}
+
+
+//double ucs_rv64_init_tsc_freq();
```
yosefe, line 46: remove

src/ucs/arch/riscv/cpu.h
```diff
@@ -0,0 +1,105 @@
...
+#define ucs_memory_bus_fence()        asm volatile ("fence rw,rw"::: "memory")
+#define ucs_memory_bus_store_fence()  asm volatile ("fence w,w" ::: "memory")
+#define ucs_memory_bus_load_fence()   asm volatile ("fence r,r" ::: "memory")
+#define ucs_memory_bus_cacheline_wc_flush()
+#define ucs_memory_cpu_fence()        ucs_compiler_fence()
+#define ucs_memory_cpu_store_fence()  ucs_compiler_fence()
+#define ucs_memory_cpu_load_fence()   ucs_compiler_fence()
+#define ucs_memory_cpu_wc_fence()     asm volatile ("fence w,w" ::: "memory")
+
+static inline double ucs_arch_get_clocks_per_sec()
```
yosefe, line 38: UCS_F_ALWAYS_INLINE

src/ucs/arch/riscv/cpu.h
```diff
@@ -0,0 +1,105 @@
...
+    return (double) freq;
+}
+
+
+//double ucs_rv64_init_tsc_freq();
+
+typedef struct ucs_rv64_cpuid {
+	 char isa[256];
+	 char mmu[256];
+	 char uarch[256];
```
yosefe, line 51: how is the cpuid used?

src/ucs/arch/riscv/global_opts.c
```diff
@@ -0,0 +1,44 @@
...
+  {"BUILTIN_MEMCPY_MIN", "auto",
+   "Minimal threshold of buffer length for using built-in memcpy.",
+   ucs_offsetof(ucs_arch_global_opts_t, builtin_memcpy_min), UCS_CONFIG_TYPE_MEMUNITS},
+
+  {"BUILTIN_MEMCPY_MAX", "auto",
+   "Maximal threshold of buffer length for using built-in memcpy.",
+   ucs_offsetof(ucs_arch_global_opts_t, builtin_memcpy_max), UCS_CONFIG_TYPE_MEMUNITS},
+#endif
+  {NULL}
+};
```
yosefe, line 27: are we going to have special memcpy for riscv?

yosefe (commented):
as general comment need to update ucx docx (under docs folder) to show riscv support and also see how can we add riscv to CI

src/ucs/arch/riscv/global_opts.c
ct-clmsn: A run of `ucx_info`, `ucx_perf`, etc without this call causes illegal instruction segmentation faults. I'd like to retain this code.

src/ucs/arch/riscv/cpu.h
ct-clmsn: it's used to load data from the operating system as required by ucs.

src/ucs/arch/riscv/atomic.h
ct-clmsn: Did a little quick grep work and saw that these files have dependencies on the functions generated by the macros:

* src/ucp/wireup/wireup_ep.c
* src/ucs/memory/rcache.c
* src/ucs/async/async.c
* src/ucs/async/thread.c
* src/uct/sm/base/sm_ep.c

not sure what the right answer looks like; this implementation attempts to maintain consistency with the other architecture implementations.

src/ucm/util/log.c
ct-clmsn: found and fixed indentation issues. we had errors running the `.clang-format` script.

src/ucs/arch/riscv/atomic.h
yosefe: see https://github.com/openucx/ucx/blob/1a9873aa6475b434a43cf36978cd725c53c3cc61/src/ucs/arch/atomic.h#L17, maybe we can use generic/atomic.h for riscv as well

src/ucs/arch/riscv/cpu.h
yosefe: i don't see any debug message using "uarch" or "mmu" for example, can you pls point to it?

src/ucs/arch/riscv/global_opts.c
yosefe: aarch64 for example does not have it

src/ucs/arch/riscv/bitops.h
shamisp: @jleidel @yosefe looking at linux source code I saw that in some places they use RISCV and in other RV64.
It seems line RV64 is an equivalent of aarch64, which actually describes the arch better.
In some other places I see people using riscv64. Anyway, if the code only works for RV64 ISA then maybe it is better to use RV64

src/ucm/bistro/bistro.c
ct-clmsn: the riscv processor has a `[fence.i](https://five-embeddev.com/riscv-isa-manual/latest/zifencei.html)` instruction synchronizing both the instruction and data caches; it exists for self modifying code (JITs). would you and @shamisp be open to each architecture providing a macro that is selectively implemented to optionally support this processor feature?

src/ucs/arch/riscv/bitops.h
ct-clmsn: fixed these issues. @shamisp; I chatted with @jleidel about the naming convention rv64 or riscv. we landed on riscv; we have a similar set of concerns and observations. if ya'll want or prefer rv64 that's an easy enough modification.

src/ucs/arch/riscv/cpu.c
ct-clmsn: fixed these issues.

src/ucs/arch/riscv/cpu.c
ct-clmsn: @shamisp 's comment about riscv or rv64 applies here. will need to sort out naming convention.

src/ucs/arch/riscv/cpu.h
ct-clmsn: This was implemented because the ARM support had `ucs_aarch64_cpuid_t`. `ucs_aarch64_cpuid_t` exists because of an `#ifdef __aarch64__` in `./uct/ib/base/ib_device.c`; an assumption was made that riscv64 would require similar work.

The type and the code that initializes the type are used in the riscv64 version of `ucs_arch_get_cpu_vendor`. `ucs_arch_get_cpu_vendor` is called in several places but the information stored in the type is never actually consumed or used.

This structure and that code has been removed.

src/ucm/bistro/bistro.c
yosefe: yes; for now it can be empty for all other architectures

src/ucs/arch/riscv/global_opts.c
ct-clmsn: Quick note, there is still a tremendous amount of experimental code and/or lacking features in portions of the RISC-V system software and tool chains. Went back through the patch, updated some `fence.i` locations, and was able to remove this successfully.

src/ucs/arch/riscv/cpu.h
ct-clmsn: fixed this one.

src/ucm/bistro/bistro.c
shamisp: I have it implemented for arm, but we had no reason to expose it outside of arch:
https://github.com/openucx/ucx/blob/57615271cb77440ccb131d581e4f108a2b234fbc/src/ucs/arch/aarch64/cpu.h#L39

We have `#define ucs_memory_cpu_fence` and I think this one can follow similar naming  convention `#define ucs_instruction_cpu_fence`

shamisp: I'm quire a bit confused why we need I-berrier here. Eventually, when patched applied we invoke `ucs_clear_cache` which flushes the I-cache.

src/ucs/arch/cpu.h
```diff
@@ -98,6 +101,11 @@ typedef struct ucs_cpu_builtin_memcpy {
 #  include "ppc64/cpu.h"
 #elif defined(__aarch64__)
 #  include "aarch64/cpu.h"
+#elif defined(__riscv)
+#  ifdef HAVE___CLEAR_CACHE
```
shamisp, line 105: What is the problem with built-in clear cache ?

src/ucs/arch/riscv/cpu.h
```diff
@@ -0,0 +1,105 @@
...
+}
+
+ucs_cpu_vendor_t ucs_arch_get_cpu_vendor();
+
+static inline ucs_status_t ucs_arch_get_cache_size(size_t *cache_sizes)
+{
+    return UCS_ERR_UNSUPPORTED;
+}
+
+#if !HAVE___CLEAR_CACHE
```
shamisp, line 76: In the above code you undefined HAVE___CLEAR_CACHE , so it is never defined.

src/ucs/arch/riscv/cpu.h
```diff
@@ -0,0 +1,105 @@
...
+static inline void ucs_arch_clear_cache(void *start, void *end)
+{
+    __riscv_flush_icache(start, end, 0);
```
shamisp, line 79: I suspect clear cache maps to the same call. IMHO  something really broken here and this is why you had to introduce fence.i in multiple places.

src/ucs/arch/cpu.h
jleidel: It doesn't necessarily clear the cache in the manner expected.  The particular implementation of RISC-V that we're testing on predates the ratification of the CMOBase set of extensions (Zicbom, Zicbop, Zicboz) that implement explicit cache flush operations.  Given that the community was well aware of the forthcoming CMOBase infrastructure (ratified in Nov 2021), the toolchain and system libs have somewhat lagged their support of particular builtins.  They're "implemented", but not fully implemented.  This is quite literally a situation where we're running on rather early hardware/software.

src/ucs/arch/riscv/bitops.h
ct-clmsn: @shamisp which is your preference? `rv64` or `riscv`?

src/ucs/arch/riscv/cpu.h
ct-clmsn: Thank you for pointing this out - this issue has been fixed.

src/ucs/arch/riscv/cpu.h
ct-clmsn: fix made.

src/ucs/arch/riscv/cpu.h
ct-clmsn: @jleidel's response clarifies things a bit. gcc has 2 intrinsics for barriers on both caches. `__sync_synchronize` for data cache and `__riscv_flush_icache` for instruction cache. It was a little unclear which cache is flushed by this function.

Additionally, I've done some runtime gdb disassembly of both (`__sync_synchronize `, `__riscv_flush_icache `) intrinsics; the disassembly for both intrinsics have no reference to the `fence.i` instruction. The disassembly  looks like an attempt to perform both memory barriers using software.

src/ucs/arch/riscv/bitops.h
shamisp: @ct-clmsn my approach similar to Yossis' - as long as it is consistent it is all good. if it only works on 64b platforms IMHO rv64 is better name, but this is really up to you.

src/ucs/arch/riscv/cpu.h
shamisp: This is the reference I mentioned on the call: https://github.com/openucx/ucx/blob/57615271cb77440ccb131d581e4f108a2b234fbc/src/ucs/arch/aarch64/cpu.h#L157

ct-clmsn:
@shamisp @yosefe quick update: several modifications have been made and pushed into this PR.

yosefe:
@ct-clmsn can you pls fix the commit title according to code style https://github.com/openucx/ucx/wiki/Guidance-for-contributors item 3?
(would need to force-push to resolve)

ct-clmsn:
@yosefe change the automated commit message for the merge (commit # `[a3e2624](https://github.com/openucx/ucx/pull/8246/commits/a3e2624f54054b5bd280ff45d27fcddec0e4311a)`)?

yosefe:
Not the merge commit; the other ones.

src/ucm/bistro/bistro.c
shamisp: @ct-clmsn Do you still need the `if defined`, even after fixing cache flush ?

ct-clmsn:
@yosefe apologies for all the rebase work - was able to get the commit titles/messages updated.

src/ucs/arch/rv64/global_opts.h
```diff
@@ -0,0 +1,29 @@
...
+#define UCS_ARCH_GLOBAL_OPTS_INITALIZER {   \
+    .builtin_memcpy_min = UCS_MEMUNITS_AUTO, \
+    .builtin_memcpy_max = UCS_MEMUNITS_AUTO  \
+}
+
+/* built-in memcpy config */
+typedef struct ucs_arch_global_opts {
+    size_t builtin_memcpy_min;
+    size_t builtin_memcpy_max;
+} ucs_arch_global_opts_t;
```
yosefe, line 25: can remove builtin memcpy options for rv64, it's not used anyway
for example aarch64 does not have it

test/gtest/uct/ib/test_ud.cc
```diff
@@ -918,6 +918,91 @@ UCS_TEST_SKIP_COND_P(test_ud, ctls_loss,
 
 UCT_INSTANTIATE_UD_TEST_CASE(test_ud)
 
+
+class test_ud_peer_failure : public ud_base_test {
```
yosefe, line 922: is this a merge artifact? if not why needed to change UD tests?

src/ucm/mmap/install.c
```diff
@@ -31,10 +31,27 @@
...
 
 #define UCM_HOOK_STR \
     ((ucm_mmap_hook_mode() == UCM_MMAP_HOOK_RELOC) ?  "reloc" : "bistro")
 
+#if defined(__riscv)
+#define UCM_FIRE_EVENT(_event, _mask, _data, _call)                            \
+     do {                                                                      \
+         int exp_events = (_event) & (_mask);                                  \
+         (_data)->fired_events = 0;                                            \
+         ucs_rv64_isb();                                                       \
```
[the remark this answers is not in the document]
shamisp: well, ISB is actually and an instruction on Arm.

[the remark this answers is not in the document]
shamisp: Why do you think instruction barrier is needed here - from the code is not as obvious.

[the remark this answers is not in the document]
ct-clmsn: The RV64 manual provides mappings for fence instructions that map back into other architectures (ARM, PPC, x86, etc). I implemented the ARM flavors of fences to provide something that looks familiar to ya'll. RV64's data cache flush support is not implemented in the available hardware. Segmentation faults happen when fences and the icache flush are not placed before the first time a bistro'd function is invoked.

src/ucm/mmap/install.c
```diff
@@ -181,7 +199,11 @@ ucm_fire_mmap_events_internal(int events, ucm_mmap_test_events_data_t *data,
     }
 
     if (events & (UCM_EVENT_SHMAT|UCM_EVENT_SHMDT|UCM_EVENT_VM_MAPPED|UCM_EVENT_VM_UNMAPPED)) {
+#if defined(__riscv)
+        ucs_arch_clear_cache(p, p+ucm_get_page_size());
+#endif
```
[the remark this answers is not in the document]
ct-clmsn: Segmentation faults happen when fences and the icache flush are not placed before the first time a bistro'd function is invoked.

src/ucm/util/replace.h
```diff
@@ -27,6 +27,23 @@ extern pthread_t volatile ucm_reloc_get_orig_thread;
 #define UCM_DEFINE_REPLACE_FUNC(_name, _rettype, _fail_val, ...) \
     _UCM_DEFINE_REPLACE_FUNC(ucm_override_##_name, ucm_##_name, _rettype, _fail_val, __VA_ARGS__)
 
+#if defined(__riscv)
```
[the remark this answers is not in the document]
ct-clmsn: I found an issue in ucm_trace on RV64, there is a `memcpy` in `ucm_trace` that was causing segmentation faults. Removing the `ucm_trace` fixed the issue.

src/ucs/arch/rv64/global_opts.h
ct-clmsn: thought this was removed previously, looks like the last push didn't remove it.

test/gtest/uct/ib/test_ud.cc
ct-clmsn: merge artifact; will fix.

src/ucm/mmap/install.c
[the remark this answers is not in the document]
yosefe: keep the cache flush. but make it unconditional, i.e do it for all architectures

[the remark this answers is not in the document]
ct-clmsn: would a macro like the following suffice?

`UCM_INVOKE(mmap, ...);`

On RV64, cache flushes could be added by the macro prior to mmap(...) getting invoked. On the non-RV64 platform, the cacheflushes would not occur.

src/ucm/util/replace.h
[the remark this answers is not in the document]
ct-clmsn: apologies, the issue wasn't `memcpy`, the segmentation fault happens in [ucm_log_vsnprintf](https://github.com/tactcomplabs/ucx/blob/b84d18df554f6b2499bcfdbe50e01133fdfb322f/src/ucm/util/log.c#L171); which is called by `ucm_trace` under the hood. There is a situation on the RV64 platform where the `strlen` on that line segfaults. I have a custom RV64 patch for `ucm_log_vsnprintf` to fix the strlen issue, not sure if ya'll would want to provide a custom architecture fix for that function.

src/ucm/mmap/install.c
```diff
@@ -181,7 +199,11 @@ ucm_fire_mmap_events_internal(int events, ucm_mmap_test_events_data_t *data,
...
         shmid = shmget(IPC_PRIVATE, ucm_get_page_size(), IPC_CREAT | SHM_R | SHM_W);
+
```
yosefe, line 206: remove the spaceline

src/ucm/mmap/install.c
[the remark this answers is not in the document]
yosefe: @ct-clmsn should be something like:
```c
if (events & (UCM_EVENT_SHMAT|UCM_EVENT_SHMDT|UCM_EVENT_VM_MAPPED|UCM_EVENT_VM_UNMAPPED)) {
    ucs_arch_sync_inst_cache(p, p + ucm_get_page_size());
    shmid = shmget(IPC_PRIVATE, ucm_get_page_size(), IPC_CREAT | SHM_R | SHM_W);
   ...
}
```

And ucs_arch_sync_inst_cache() is defined to ucs_arch_clear_cache on rv4, empty on other archs.

However, looking at the code more, why is this needed only before shmget and not in other places?

src/ucm/util/replace.h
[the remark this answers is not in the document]
yosefe: @ct-clmsn what is the reason strlen fails on rv64 and not on other archs?
what is the fix?
IMO better try to fix the underlying issue rather and working around it

src/ucm/util/replace.h
```diff
@@ -27,6 +27,23 @@ extern pthread_t volatile ucm_reloc_get_orig_thread;
...
+#define _UCM_DEFINE_REPLACE_FUNC(_over_name, _ucm_name, _rettype, _fail_val, ...) \
+    \
+    /* Define a symbol which goes to the replacement - in case we are loaded first */ \
+    _rettype _over_name(UCM_FUNC_DEFINE_ARGS(__VA_ARGS__)) \
```
shamisp, line 34: it seems like only the difference here is ucm_trace call ? What is the problem with the trace call ?

shamisp: Sorry I missed the above discussion

src/ucm/util/replace.h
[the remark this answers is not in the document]
shamisp: I agree with @yosefe suggestion. Let's fix the actual problem and not hack around it.

src/ucm/mmap/install.c
shamisp, line 46: Hmm, why ISB is needed here ? I don't see any obvious reason.
If there is good reason probably it is required on all architectures.

shamisp: More I read it  and more I don't understand it. The ISB if followed by system call that switches to privilege mode. In theory it should  include ISB and memory barrier.  Out curiosity, have you tried to replace it with memory write barrier ?

ct-clmsn:
@shamisp @yosefe did some testing today. It turns out riscv64's implementation of `brk` is not equivalent to `syscall(SYS_brk, ...)`. This is the source of several issues. I'm in the process of implementing a work-around. This should resolve all of the issues above, with exception to any outstanding character-spacing issues.

ct-clmsn:
@yosefe @shamisp `ucm_log_vsnprintf` (in the file `src/ucm/util/log.c`) accepts an argument `va_list ap`. A line of code `value.s = va_arg(ap, char *);` expects `va_arg` to return `NULL` or `0` when it currently does not behave like this on RV64. Working on a fix for this issue.

jleidel:
At this point, I believe we have exhausted all possible options for removing architecture-specific I$ flushes in the RV64 implementation.  Given that this is currently the "most" functional platform we have available for UCX prototyping, we limited in our options with respect to removing architecture specific code from non-architecture specific sections.  Our current platform lacks certain support for hardware-driven cache flushes (RV-CMOBASE extension), compat headers (for libfabrics support; kernel 5.11.0-1028) and certain ELF extensions (DT_JMPREL is missing in binutils).  We've tried all combinations of architecture-agnostic user and system calls to arrive at a solution.  Many of which would potentially induce performance issues in future platform variants.  Given the "experimental" nature of RISC-V platforms in HPC, we would like to request that we allow this PR to merge as is.  This would allow everyone in the RISC-V community to make forward progress on other HPC runtime/comm layers.  However, we will also sign up to amend the code once we have a more functional/stable baseline platform to use for tests and experiments.  The RISC-V "platforms" and "profiles" specs will define specific extension requirements that will ease the pain of porting complex runtimes such as OpenUCX to future RISC-V variants.

shamisp:
> At this point, I believe we have exhausted all possible options for removing architecture-specific I$ flushes in the RV64 implementation. Given that this is currently the "most" functional platform we have available for UCX prototyping, we limited in our options with respect to removing architecture specific code from non-architecture specific sections. Our current platform lacks certain support for hardware-driven cache flushes (RV-CMOBASE extension), compat headers (for libfabrics support; kernel 5.11.0-1028) and certain ELF extensions (DT_JMPREL is missing in binutils). We've tried all combinations of architecture-agnostic user and system calls to arrive at a solution. Many of which would potentially induce performance issues in future platform variants. Given the "experimental" nature of RISC-V platforms in HPC, we would like to request that we allow this PR to merge as is. This would allow everyone in the RISC-V community to make forward progress on other HPC runtime/comm layers. However, we will also sign up to amend the code once we have a more functional/stable baseline platform to use for tests and experiments. The RISC-V "platforms" and "profiles" specs will define specific extension requirements that will ease the pain of porting complex runtimes such as OpenUCX to future RISC-V variants.

@yosefe @ct-clmsn @jleidel Since the u-arch/arch issues cannot be resolved what about moving parts of the abstraction that got `ifdef-ed` (`UCM_FIRE_EVENT, ucs_arch_clear_cache, _UCM_DEFINE_REPLACE_FUNC`) to arch files in bistro. This way we can have one generic implementation and RV64 will have an option to overwrite those without ifdefs in the main code. @yosefe  ?

ct-clmsn:
@shamisp @yosefe I've made some updates to the bistro support and placed compile guards around some of the openmp code. if ya'll need me to rebase against master say the word.

yosefe:
@ct-clmsn can you pls fix the commit title? would need force-push

ct-clmsn:
@yosefe apologies, this commit? [06a1c3f](https://github.com/openucx/ucx/pull/8246/commits/06a1c3fa7c09e2900beb8552c1335b661f16874e)

src/ucm/bistro/bistro_rv64.c
```diff
@@ -0,0 +1,105 @@
...
+  * Register X0 is always zero
+  */
+ #define X31 31
+ #define X30 30
+ #define X0  0
+
+ /**
+  * @brief JALR - Add 12 bit immediate to source register, save to destination register, jump and link from destination register
+  *
+  * @param[in] _reg  register number (0-31), @param[out] _reg register number (0-31), @param[imm] 12 bit immmediate value
```
rbradford, line 43: I fixed this in https://github.com/openucx/ucx/pull/9168/commits/f9083eb7874e0f0dc6f187fac1c54f61e4f97d7a

These are not out variables in the sense that doxygen would consider it. Also there is no "imm" type.

src/ucs/arch/rv64/cpu.h
```diff
@@ -0,0 +1,121 @@
...
+
+ucs_cpu_vendor_t ucs_arch_get_cpu_vendor();
+
+static inline ucs_status_t ucs_arch_get_cache_size(size_t *cache_sizes)
+{
+    return UCS_ERR_UNSUPPORTED;
+}
+
+static inline uint64_t ucs_arch_read_hres_clock()
+{
```
rbradford, line 94: Our solution - https://github.com/openucx/ucx/pull/9168/files#diff-9769d56a24d09635863e55444aa766f0773c064395acefc2aaddd8761b5d0b0eR54-R58

I've seen you've changed it from rdcycle which was wrong (gave absolute number of cycles since some time)

src/ucs/arch/rv64/cpu.h
```diff
@@ -0,0 +1,121 @@
...
+
+#define ucs_arch_wait_mem ucs_arch_generic_wait_mem
+
+static inline void ucs_arch_clear_cache(void * start, void * end) {
+    ucs_rv64_dcache_flush(start, end);
+    ucs_rv64_dmb();
+    ucs_rv64_icache_flush(start, end);
+}
+
+static inline void *ucs_memcpy_relaxed(void *dst, const void *src, size_t len)
```
rbradford, line 108: @yosefe Suggested we could remove the memcpy implementations.

rbradford (commented):
Did you test with `reloc` mode as well as bistro? Did you just run the UCX tests or also the UCC tests? We found some issues when we also ran the UCC tests.

Please review the improvements that have been made in this PR - https://github.com/openucx/ucx/pull/9168

yosefe:
> @yosefe apologies, to clarify this commit? [06a1c3f](https://github.com/openucx/ucx/pull/8246/commits/06a1c3fa7c09e2900beb8552c1335b661f16874e)

@ct-clmsn pls see https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=66476&view=logs&j=a3e7afe8-b282-5869-712a-9857fbd4b4a9&t=fd231581-c084-51fb-f10e-187ae5f1c153&l=30

ct-clmsn:
@yosefe @jleidel @shamisp UCX has changed significantly since this PR was proposed. This PR is OBE (overcome by events) and I'd encourage merging #9168 as it's a more up to date proposal.

PR #8249, opened 2022-05-20
UCP/PROTO: error flows

evgeny-leksikov:
## What
- add ucp_eager_bcopy_single_proto.abort
- fix pending perge for rndv get with ats stage]

## Why ?
fixes io_demo with new protocols

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -1171,8 +1171,8 @@ ucs_status_ptr_t ucp_ep_modify_nb(ucp_ep_h ep, const ucp_ep_params_t *params)
 
 void ucp_ep_err_pending_purge(uct_pending_req_t *self, void *arg)
 {
-    ucp_request_t *req      = ucs_container_of(self, ucp_request_t, send.uct);
-    ucs_status_t  status    = UCS_PTR_STATUS(arg);
+    ucp_request_t *req   = ucs_container_of(self, ucp_request_t, send.uct);
+    ucs_status_t  status = UCS_PTR_STATUS(arg);
 
     /* TODO: check for context->config.ext.proto_enable when all protocols are
      *       implemented, such as flush, AM/RNDV, etc */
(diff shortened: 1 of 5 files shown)


src/ucp/core/ucp_ep.c
```diff
@@ -1176,7 +1176,8 @@ void ucp_ep_err_pending_purge(uct_pending_req_t *self, void *arg)
 
     /* TODO: check for context->config.ext.proto_enable when all protocols are
      *       implemented, such as flush, AM/RNDV, etc */
-    if (req->flags & UCP_REQUEST_FLAG_PROTO_SEND) {
+    if (req->flags & (UCP_REQUEST_FLAG_PROTO_SEND |
+                      UCP_REQUEST_FLAG_PROTO_INITIALIZED)) {
```
yosefe, line 1180: Not all protocols have to use UCP_REQUEST_FLAG_PROTO_INITIALIZED flag (for example bcopy or single)
so abort function of each protocol should check this flag individually, or maybe some of them will use common helper function

evgeny-leksikov: Actually I used this flag for RNDV GET proto on receiver side to avoid adding new flag.
Please note that condition based on `req->flags` is a temporary and should be replaced with `context->config.ext.proto_enable` in short term and completely removed in long term.

yosefe: even when `config.ext.proto_enable` is removed, it's still not OK to use UCP_REQUEST_FLAG_PROTO_INITIALIZED flag in the **common** code. it's only OK to use it for specific protocols (such as RNDV GET)

evgeny-leksikov: then what do think about to set `UCP_REQUEST_FLAG_PROTO_SEND` for rndv_get internal request? (I see that this flag is set only from API functions) or need to add some new flag?

yosefe: > then what do think about to set `UCP_REQUEST_FLAG_PROTO_SEND` for rndv_get internal request? (I see that this flag is set only from API functions) or need to add some new flag?

Yes, this is the right way (not setting UCP_REQUEST_FLAG_PROTO_SEND for rndv_get internal request is a bug)

yosefe (approved):

PR #8250, opened 2022-05-22
UCT/API: Introduce uct_mem_reg_v2

dmitrygx:
## What

Introduced a new APIv2 function - `uct_mem_reg_v2`.

## Why ?

The `uct_mem_reg_v2` will close demand on extending memory registration routines. And it will extend already existing `uct_mem_dereg_v2` API function.

## How ?

1. Define `uct_md_mem_reg_field_mask_t` which could be extended in the future by new fields.
2. Define `uct_md_mem_reg_params_t` which could be extended in the future by new fields.
3. Update `uct_md_mem_dereg_v2` to mention `uct_md_mem_reg_v2` along with `uct_md_mem_reg` and describe return value for consistency.
4. Declare `uct_md_mem_reg_v2` and add its implementation.
5. Move implementation of `uct_md_mem_reg` to `uct_md_mem_reg_v2`. And implement `uct_md_mem_reg` on top of `uct_md_mem_reg_v2`.
6. Define `UCT_MD_MEM_REG_FIELD_VALUE` auxiliary macro for getting fields from registration parameters.
7. Update `mem_reg` MD ops in all transports to support `uct_md_mem_reg_v2`'s `params` and its further extensions.
8. Update gtests to use `uct_md_mem_reg_v2` instead of `uct_md_mem_reg` where `uct_md_mem_dereg_v2` is used.
9. Add gtest to check bad parameters passed to `uct_md_mem_reg_v2`.

### DIFF
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -192,6 +192,15 @@ typedef struct {
 
 /**
  * @ingroup UCT_MD
+ * @brief MD memory registration operation flags.
+ */
+typedef enum {
+    UCT_MD_MEM_REG_FIELD_FLAGS = UCS_BIT(0)
+} uct_md_mem_reg_field_mask_t;
+
+
+/**
+ * @ingroup UCT_MD
  * @brief MD memory de-registration operation flags.
  */
 typedef enum {
@@ -310,6 +319,26 @@ typedef void (*uct_md_mem_invalidate_cb_t)(void *arg);
 
 /**
  * @ingroup UCT_MD
+ * @brief Operation parameters passed to @ref uct_md_mem_reg_v2.
+ */
+typedef struct uct_md_mem_reg_params {
+    /**
+     * Mask of valid fields in this structure and operation flags, using
+     * bits from @ref uct_md_mem_reg_field_mask_t. Fields not specified
+     * in this mask will be ignored. Provides ABI compatibility with respect
+     * to adding new fields.
+     */
+    uint64_t                     field_mask;
+
+    /**
+     * Operation specific flags, using bits from @ref uct_md_mem_flags_t.
+     */
+    uint64_t                     flags;
+} uct_md_mem_reg_params_t;
+
+
+/**
+ * @ingroup UCT_MD
  * @brief Operation parameters passed to @ref uct_md_mem_dereg_v2.
  */
 typedef struct uct_md_mem_dereg_params {
@@ -419,16 +448,39 @@ uct_iface_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr);
 
 /**
  * @ingroup UCT_MD
- * @brief Undo the operation of @ref uct_md_mem_reg() and invalidate memory
- *        region.
+ * @brief Register memory for zero-copy sends and remote access.
+ *
+ * Register memory on the memory domain. In order to use this function, @a md
+ * must support the @ref UCT_MD_FLAG_REG flag.
+ *
+ * @param [in]  md          Memory domain that was used to register the memory.
+ * @param [in]  address     Memory to register.
+ * @param [in]  length      Size of memory to register. Must be > 0.
+ * @param [in]  params      Operation parameters, see @ref
+ *                          uct_md_mem_reg_params_t.
+ * @param [out] memh_p      Filled with handle for allocated region.
+ *
+ * @return Error code.
+ */
+ucs_status_t uct_md_mem_reg_v2(uct_md_h md, void *address, size_t length,
+                               const uct_md_mem_reg_params_t *params,
+                               uct_mem_h *memh_p);
+
+
+/**
+ * @ingroup UCT_MD
+ * @brief Undo the operation of @ref uct_md_mem_reg() or
+ *        @ref uct_md_mem_reg_v2() and invalidate memory region.
  *
  * This routine deregisters the memory region registered by @ref uct_md_mem_reg
- * and allow the memory region to be invalidated with callback called when the
+ * and allows the memory region to be invalidated with callback called when the
  * memory region is unregistered.
  *
  * @param [in]  md          Memory domain that was used to register the memory.
  * @param [in]  params      Operation parameters, see @ref
  *                          uct_md_mem_dereg_params_t.
+ *
+ * @return Error code.
  */
 ucs_status_t uct_md_mem_dereg_v2(uct_md_h md,
                                  const uct_md_mem_dereg_params_t *params);
--- a/src/uct/base/uct_md.c
+++ b/src/uct/base/uct_md.c
@@ -465,6 +465,19 @@ uct_md_mem_advise(uct_md_h md, uct_mem_h memh, void *addr, size_t length,
 ucs_status_t uct_md_mem_reg(uct_md_h md, void *address, size_t length,
                             unsigned flags, uct_mem_h *memh_p)
 {
+    uct_md_mem_reg_params_t params = {
+        .field_mask = UCT_MD_MEM_REG_FIELD_FLAGS,
+        .flags      = flags
+    };
+
+    return uct_md_mem_reg_v2(md, address, length, &params, memh_p);
+}
+
+ucs_status_t uct_md_mem_reg_v2(uct_md_h md, void *address, size_t length,
+                               const uct_md_mem_reg_params_t *params,
+                               uct_mem_h *memh_p)
+{
+    uint64_t flags = UCT_MD_MEM_REG_FIELD_VALUE(params, flags, FIELD_FLAGS, 0);
     ucs_status_t status;
 
     if ((length == 0) || (address == NULL)) {
@@ -477,12 +490,12 @@ ucs_status_t uct_md_mem_reg(uct_md_h md, void *address, size_t length,
     status = uct_mem_check_flags(flags);
     if (status != UCS_OK) {
         uct_md_log_mem_reg_error(flags,
-                                 "uct_md_mem_reg(flags=0x%x): invalid flags",
-                                 flags);
+                                 "uct_md_mem_reg_v2(flags=0x%lx): invalid"
+                                 " flags", flags);
         return status;
     }
 
-    return md->ops->mem_reg(md, address, length, flags, memh_p);
+    return md->ops->mem_reg(md, address, length, params, memh_p);
 }
 
 ucs_status_t uct_md_mem_dereg(uct_md_h md, uct_mem_h memh)
@@ -517,6 +530,25 @@ ucs_status_t uct_md_detect_memory_type(uct_md_h md, const void *addr, size_t len
                                        ucs_memory_type_t *mem_type_p)
 {
     return md->ops->detect_memory_type(md, addr, length, mem_type_p);
+}
+
+ucs_status_t uct_md_dummy_mem_reg(uct_md_h md, void *address, size_t length,
+                                  const uct_md_mem_reg_params_t *params,
+                                  uct_mem_h *memh_p)
+{
+    /* We have to emulate memory registration. Return dummy pointer */
+    *memh_p = (void*)0xdeadbeef;
+    return UCS_OK;
+}
+
+ucs_status_t uct_md_dummy_mem_dereg(uct_md_h uct_md,
+                                    const uct_md_mem_dereg_params_t *params)
+{
+    UCT_MD_MEM_DEREG_CHECK_PARAMS(params, 0);
+
+    ucs_assert(params->memh == (void*)0xdeadbeef);
+
+    return UCS_OK;
 }
 
 void uct_md_set_rcache_params(ucs_rcache_params_t *rcache_params,
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.c
@@ -88,19 +88,15 @@ static ucs_status_t uct_cuda_copy_rkey_release(uct_component_t *component,
 }
 
 UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_reg,
-                 (md, address, length, flags, memh_p),
+                 (md, address, length, params, memh_p),
                  uct_md_h md, void *address, size_t length,
-                 unsigned flags, uct_mem_h *memh_p)
+                 const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
+    uint64_t flags = UCT_MD_MEM_REG_FIELD_VALUE(params, flags, FIELD_FLAGS, 0);
     ucs_log_level_t log_level;
     CUmemorytype memType;
     CUresult result;
     ucs_status_t status;
-
-    if (address == NULL) {
-        *memh_p = address;
-        return UCS_OK;
-    }
 
     result = cuPointerGetAttribute(&memType, CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
                                    (CUdeviceptr)(address));
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -752,9 +752,12 @@ err:
     return status;
 }
 
-static ucs_status_t uct_ib_mem_reg(uct_md_h uct_md, void *address, size_t length,
-                                   unsigned flags, uct_mem_h *memh_p)
+static ucs_status_t
+uct_ib_mem_reg(uct_md_h uct_md, void *address, size_t length,
+               const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
+    uint64_t flags  = UCT_MD_MEM_REG_FIELD_VALUE(params, flags, FIELD_FLAGS,
+                                                 0);
     uct_ib_md_t *md = ucs_derived_of(uct_md, uct_ib_md_t);
     ucs_status_t status;
     uct_ib_mem_t *memh;
@@ -958,10 +961,12 @@ static inline uct_ib_rcache_region_t* uct_ib_rcache_region_from_memh(uct_mem_h m
     return ucs_container_of(memh, uct_ib_rcache_region_t, memh);
 }
 
-static ucs_status_t uct_ib_mem_rcache_reg(uct_md_h uct_md, void *address,
-                                          size_t length, unsigned flags,
-                                          uct_mem_h *memh_p)
+static ucs_status_t
+uct_ib_mem_rcache_reg(uct_md_h uct_md, void *address, size_t length,
+                      const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
+    uint64_t flags  = UCT_MD_MEM_REG_FIELD_VALUE(params, flags, FIELD_FLAGS,
+                                                 0);
     uct_ib_md_t *md = ucs_derived_of(uct_md, uct_ib_md_t);
     ucs_rcache_region_t *rregion;
     ucs_status_t status;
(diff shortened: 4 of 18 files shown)


src/uct/api/v2/uct_v2.h
```diff
@@ -308,6 +319,36 @@ struct uct_ep_attr {
...
+     */
+    unsigned                     flags;
+
+    /**
+     * Memory to register.
+     */
+    void                         *address;
+
+    /**
+     * Size of memory to register. Must be > 0.
```
shamisp, line 346: >=0 ?

dmitrygx: it should be > 0, see https://github.com/openucx/ucx/blob/e9abd8ab7b1146b47a326954fc0d09d91992c943/src/uct/base/uct_md.c#L470
[ibv_reg_mr](https://www.rdmamojo.com/2012/09/07/ibv_reg_mr/) requires the following:
`Size of the memory block to register, in bytes. This value must be at least 1 and less than dev_cap.max_mr_size`

dmitrygx:

src/uct/api/v2/uct_v2.h
yosefe: @dmitrygx since length cannot be 0, and address cannot be NULL, does it mean address and length are mandatory fields? so they should be passed as regular parameters

dmitrygx:
@brminich is it ok?

brminich (approved):

dmitrygx:
@yosefe could you pls review?

src/uct/api/v2/uct_v2.h
shamisp: Do we treat the length zero as a special case because of a decade old bug in verbs registration code ?

dmitrygx: > Do we treat the length zero as a special case because of a decade old bug in verbs registration code ?

actually, this behavior came from `uct_mem_reg` API.
btw, registering zero-length doesn't sound as a thing that really makes sense.

dmitrygx: @yosefe @shamisp making memh to be some `NULL_HANDLE` stub or just `NULL` requires changing many places to support `NULL_HANDLE`/`NULL`.
Do you think it is worth it? IMO, it complicates the code.

dmitrygx:
@shamisp during 0-length support implementation for uct_mem_reg(), we found that some transport don’t support registering 0-length, e.g. CUDA (and probably, ROCm).
Assigning dummy memh couldn’t be done in a common code, because we should create valid transport-specific memh. In its turn, it require checking for dummy memh in mkey_pack/rkey_unpack and mem_dereg. Updating pack/unpack is needed to create a valid rkey to pass it to operations and don’t add more branches.
In addition, if we are updating uct_mem_reg() to support 0-length, we should do the same for uct_mem_alloc().
Also, we don’t have requests from UCT-based users.
So, taking into account all these points, I think it doesn’t make sense to add support for 0-length registration. And it could be done in the future without breaking backward compatibility.
wdyt?

src/uct/api/v2/uct_v2.h
shamisp: if this overcomplicates the code, I'm okay to stay with the requirement `(length!=0)`

dmitrygx: @shamisp yes, it over complicated the functionality.
see https://github.com/openucx/ucx/pull/8250#issuecomment-1139459009 for more details

shamisp: @dmitrygx I saw this comment. As I mentioned, I'm okay with not supporting length 0

src/uct/api/v2/uct_v2.h
```diff
@@ -190,6 +190,17 @@ typedef struct {
...
 
 
+/**
+ * @ingroup UCT_MD
+ * @brief MD memory registration operation flags.
+ */
+typedef enum {
+    UCT_MD_MEM_REG_FIELD_FLAGS   = UCS_BIT(0), /**< flags field */
+    UCT_MD_MEM_REG_FIELD_ADDRESS = UCS_BIT(1), /**< address field */
+    UCT_MD_MEM_REG_FIELD_LENGTH  = UCS_BIT(2)  /**< length field */
```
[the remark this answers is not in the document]
brminich: i'd not do this a requirement based on that

src/uct/base/uct_md.c
```diff
@@ -465,9 +465,32 @@ uct_md_mem_advise(uct_md_h md, uct_mem_h memh, void *addr, size_t length,
...
+                                                            FIELD_ADDRESS,
+                                                            NULL);
+    size_t UCS_V_UNUSED length = UCT_MD_MEM_REG_FIELD_VALUE(params, length,
+                                                            FIELD_LENGTH, 0);
+    uint64_t flags             = UCT_MD_MEM_REG_FIELD_VALUE(params, flags,
+                                                            FIELD_FLAGS, 0);
     ucs_status_t status;
 
-    if ((length == 0) || (address == NULL)) {
+    if (ENABLE_PARAMS_CHECK && ((length == 0) || (address == NULL))) {
```
brminich, line 493: why do we add `PARAMS_CHECK`? Can lead to a weird errors

src/uct/rocm/copy/rocm_copy_md.c
```diff
@@ -150,9 +150,18 @@ static ucs_status_t uct_rocm_copy_mem_reg_internal(
...
+                      uct_mem_h *memh_p)
 {
+    void *address                 = UCT_MD_MEM_REG_FIELD_VALUE(params, address,
+                                                               FIELD_ADDRESS,
+                                                               NULL);
+    size_t length                 = UCT_MD_MEM_REG_FIELD_VALUE(params, length,
+                                                               FIELD_LENGTH,
+                                                               0);
+    uint64_t flags                = UCT_MD_MEM_REG_FIELD_VALUE(params, flags,
+                                                               FIELD_FLAGS, 0);
```
brminich, line 164: not needed here

test/gtest/uct/test_md.cc
```diff
@@ -731,10 +756,17 @@ UCS_TEST_SKIP_COND_P(test_md, dereg_bad_arg,
...
-    ptr    = malloc(size);
-    status = uct_md_mem_reg(md(), ptr, size, md_flags, &memh);
+    ptr                   = malloc(size);
+    reg_params.field_mask = UCT_MD_MEM_REG_FIELD_FLAGS |
+                            UCT_MD_MEM_REG_FIELD_ADDRESS |
+                            UCT_MD_MEM_REG_FIELD_LENGTH;
+    reg_params.flags      = md_flags;
+    reg_params.address    = ptr;
+    reg_params.length     = size;
+    status                = uct_md_mem_reg_v2(md(), &reg_params, &memh);
```
brminich, line 769: i'd create a wrapper, accepting `addr`, `length` and `flags` and returning status of `uct_md_mem_reg`, as this part of code is used 3 times here

dmitrygx: done

src/uct/api/v2/uct_v2.h
```diff
@@ -308,6 +319,36 @@ struct uct_ep_attr {
...
+
+    /**
+     * Memory to register.
+     */
+    void                         *address;
+
+    /**
+     * Size of memory to register. Must be > 0.
+     */
+    size_t                       length;
```
shamisp, line 348: Based on the above discussion it seems like length is not optional anymore ?

brminich (approved):

src/uct/api/v2/uct_v2.h
dmitrygx: currently, it is mandatory parameter.
but we are thinking about adding Cross-GVMI support in UCX which would allow us passing rkey/rkey_buffer instead of buffer:length pair.

dmitrygx:
@shamisp @yosefe @tonycurtis could you review pls?

src/uct/api/v2/uct_v2.h
```diff
@@ -419,8 +460,27 @@ uct_iface_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr);
 
 /**
  * @ingroup UCT_MD
- * @brief Undo the operation of @ref uct_md_mem_reg() and invalidate memory
- *        region.
+ * @brief Register memory for zero-copy sends and remote access.
+ *
+ * Register memory on the memory domain. In order to use this function, MD
+ * must support @ref UCT_MD_FLAG_REG flag.
```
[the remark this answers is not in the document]
dmitrygx: fixed

src/uct/api/v2/uct_v2.h
```diff
@@ -419,8 +460,27 @@ uct_iface_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr);
...
+                               uct_mem_h *memh_p);
+
+
+/**
+ * @ingroup UCT_MD
+ * @brief Undo the operation of @ref uct_md_mem_reg() or
+ *        @ref uct_md_mem_reg_v2() and invalidate memory region.
  *
  * This routine deregisters the memory region registered by @ref uct_md_mem_reg
  * and allow the memory region to be invalidated with callback called when the
```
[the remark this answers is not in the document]
dmitrygx: fixed

src/uct/api/v2/uct_v2.h
```diff
@@ -190,6 +190,17 @@ typedef struct {
...
 
 
+/**
+ * @ingroup UCT_MD
+ * @brief MD memory registration operation flags.
+ */
+typedef enum {
+    UCT_MD_MEM_REG_FIELD_FLAGS   = UCS_BIT(0),
+    UCT_MD_MEM_REG_FIELD_ADDRESS = UCS_BIT(1),
+    UCT_MD_MEM_REG_FIELD_LENGTH  = UCS_BIT(2)
```
yosefe, line 200: maybe first address, then length, then flags?

src/uct/api/v2/uct_v2.h
```diff
@@ -419,16 +460,37 @@ uct_iface_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr);
 
 /**
  * @ingroup UCT_MD
- * @brief Undo the operation of @ref uct_md_mem_reg() and invalidate memory
- *        region.
+ * @brief Register memory for zero-copy sends and remote access.
+ *
+ * Register memory on the memory domain. In order to use this function, MD
```
yosefe, line 465: MD -> `@a md`

src/uct/api/v2/uct_v2.h
```diff
@@ -419,16 +460,37 @@ uct_iface_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr);
...
+ * must support the @ref UCT_MD_FLAG_REG flag.
+ *
+ * @param [in]  md          Memory domain that was used to register the memory.
+ * @param [in]  params      Operation parameters, see @ref
+ *                          uct_md_mem_reg_params_t.
+ * @param [out] memh_p      Filled with handle for allocated region.
+ *
+ * @return Error code.
+ */
+ucs_status_t uct_md_mem_reg_v2(uct_md_h md,
```
yosefe, line 475: did it pass clang-format?

src/uct/base/uct_md.c
```diff
@@ -465,6 +465,29 @@ uct_md_mem_advise(uct_md_h md, uct_mem_h memh, void *addr, size_t length,
...
+}
+
+ucs_status_t uct_md_mem_reg_v2(uct_md_h md,
+                               const uct_md_mem_reg_params_t *params,
+                               uct_mem_h *memh_p)
+{
+    void UCS_V_UNUSED *address = UCT_MD_MEM_REG_FIELD_VALUE(params, address,
+                                                            FIELD_ADDRESS,
+                                                            NULL);
+    size_t UCS_V_UNUSED length = UCT_MD_MEM_REG_FIELD_VALUE(params, length,
```
yosefe, line 487: why need UCS_V_UNUSED?

test/gtest/uct/test_md.cc
```diff
@@ -731,10 +762,12 @@ UCS_TEST_SKIP_COND_P(test_md, dereg_bad_arg,
...
     void *ptr;
     ucs_status_t status;
+    uct_md_mem_reg_params_t reg_params;
     uct_md_mem_dereg_params_t params;
 
-    ptr    = malloc(size);
-    status = uct_md_mem_reg(md(), ptr, size, md_flags, &memh);
+    ptr = malloc(size);
+    set_reg_params(&reg_params, md_flags, ptr, size);
+    status = uct_md_mem_reg_v2(md(), &reg_params, &memh);
```
yosefe, line 770: maybe wrap both functions?
reg_mem(ptr, size, md_flags) that also calls uct_md_mem_reg_v2()

yosefe (approved):

dmitrygx:
@shamisp @tonycurtis could you review pls?

src/uct/api/v2/uct_v2.h
shamisp: @dmitrygx Can you please clarify the above. I don't understand why address and length moved to optional parameters. I will need a bit more details on this.

dmitrygx: > @dmitrygx Can you please clarify the above. I don't understand why address and length moved to optional parameters. I will need a bit more details on this.

@shamisp sorry missed your question...
Cross GVMI memory key is used to allow cross GVMI memory access using indirect memory registration which crosses vHCA context. (ref https://docs.nvidia.com/networking/display/ConnectX6DxFirmwarev22322004/Changes+and+New+Feature+History)
In other words, we register user-allocated virtual memory on HOST, share the key and address:length pair with another process which is running on BlueField's ARM. So, this process is able to use this memory for doing send/recv/RMA/AMO.

shamisp: @dmitrygx But I would expect that address:length pair  is still required for this registration ? Is there verbs interface for this ?

dmitrygx: @shamisp 

> But I would expect that address:length pair is still required for this registration ?

it is not required, since we are going to pass `rkey_buffer` which contains Cross-GVMI rkey.

> Is there verbs interface for this ?

From my knowledge, currently there are no Verbs interface for this

shamisp: @dmitrygx can provide some examples how it supposed to work ? How user will know that this is a special type of a rkey that you have to pass to another registration function ? I'm not sure I understand the workflow.

tonycurtis (dismissed):
text looks good

src/uct/api/v2/uct_v2.h
dmitrygx: @shamisp 

> How will user know that this is a special type of a rkey that you have to pass to another registration function ?

this is preliminary flow how Cross-GVMI feature will be implemented/used.
1. user's buffer will be registered by a process on HOST using ucp_mem_map(`SHARED`) and packing rkey_buffer.
2. sending rkey_buffer oob to a process on DPU.
3. the process on DPIU does ucp_mem_map(`rkey_buffer`).
4. the process on DPU does `ucp_am_send_nbx` providing memory handle from 3rd step.

btw, having address/length as optional parameters makes API more flexible for future updates, even if Cross-GVMI will be implemented another way,

shamisp: @dmitrygx `void *` even more flexible interface :) Our philosophy for an API - it should do one thing and do it right. 
In this specific case (remapping rkeys) I don't think this is the right interface. I think `ucp_ep_rkey_unpack` https://github.com/openucx/ucx/blob/4e1aa5e94cd56d93d6301051a82c881174aad29f/src/ucp/api/ucp.h#L2925 is probably more appropriate interface for remapping keys. Adding @yosefe to the discussion.

dmitrygx: > I think ucp_ep_rkey_unpack
 is probably more appropriate interface for remapping keys

yes, but we started to discuss UCP API.

sorry, I was wrong, `uct_reg_mem_reg_v2` will accept `uct_rkey_h`.

shamisp: @dmitrygx is not the unpack function directly goes to https://github.com/openucx/ucx/blob/16b610c558b0d57aeb0414c584432883749ea2fc/src/uct/ib/base/ib_md.c#L1869 

I don't think for this we need to break semantics of the `ucp_mem_map()`.

dmitrygx: > @dmitrygx is not the unpack function directly goes to .rkey_unpack        = uct_ib_rkey_unpack, 

we don't update neither UCP nor UCT rkey API

> I don't think for this we need to break semantics of the ucp_mem_map().

we don't do it

dmitrygx: I think passing `rkey_buffer`/`rkey` to `uct_mem_reg_v2()` doesn't break the idea of the registration, because:
- the result of the operation is memory handle - `uct_mem_h`
- using `rkey_buffer`/`rkey` we register memory exported by HOST to use it transprately on SmartNIC

shamisp: @dmitrygx Since you were advocating for making length/address optional I was trying to understand better the use case. At least based on the above discussion I don't think `uct_mem_reg_v2()` is the right interface for this. My preference it to move length and address back to argument list.

If you convinced rkey sharing use-case will not work with UCT interface, let's have deep dive discussion on this to understand the requirement, how it will work and how it will be used. I do believe this an interesting use case but I don't believe it will impact this specific interface.

dmitrygx: >  At least based on the above discussion I don't think uct_mem_reg_v2() is the right interface for this. 

Internally, it is still memory registration. IMO, from user perspective, it should look like memory registration, because we return local memory handle which could be passed to Zcopy operation.

Also, my additional points for having optional address/length (even if we will decide to not use `uct_mem_reg_v2` for passing reky_buffer):
- as an example, `uct_mem_dereg_v2` accept `memh` as optional one: https://github.com/openucx/ucx/blob/98e89780b0216e70e8c5c69efee8506152a7f238/src/uct/api/v2/uct_v2.h#L198
so, `uct_mem_dereg_v2` and `uct_mem_reg_v2` could be alligned.
- `uct_mem_reg_v2` could be extended by accepting an array of IOVs (e.g. to register under a single memory handle) instead of address/length.
- we could continue discussion of the new UCT API, while `uct_mem_reg_v2` is ready for different options.

dmitrygx:
@shamisp @brminich @yosefe @tonycurtis as discussed today on bi-weekly UCX call, update the PR to have address/length as mandatory parameters

src/uct/base/uct_md.c
```diff
@@ -465,6 +465,19 @@ uct_md_mem_advise(uct_md_h md, uct_mem_h memh, void *addr, size_t length,
...
+    };
+
+    return uct_md_mem_reg_v2(md, address, length, &params, memh_p);
+}
+
+ucs_status_t uct_md_mem_reg_v2(uct_md_h md, void *address, size_t length,
+                               const uct_md_mem_reg_params_t *params,
+                               uct_mem_h *memh_p)
+{
+    uint64_t flags = UCT_MD_MEM_REG_FIELD_VALUE(params, flags, FIELD_FLAGS, 0);
```
yosefe, line 480: should be unsigned (or convert flags to uint64_t)

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -88,20 +88,16 @@ static ucs_status_t uct_cuda_copy_rkey_release(uct_component_t *component,
 }
 
 UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_reg,
-                 (md, address, length, flags, memh_p),
+                 (md, address, length, params, memh_p),
                  uct_md_h md, void *address, size_t length,
-                 unsigned flags, uct_mem_h *memh_p)
+                 const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
+    uint64_t flags = UCT_MD_MEM_REG_FIELD_VALUE(params, flags, FIELD_FLAGS, 0);
```
yosefe, line 95: unsigned

src/uct/ib/base/ib_md.c
```diff
@@ -752,9 +752,12 @@ static ucs_status_t uct_ib_mem_reg_internal(uct_md_h uct_md, void *address,
     return status;
 }
 
-static ucs_status_t uct_ib_mem_reg(uct_md_h uct_md, void *address, size_t length,
-                                   unsigned flags, uct_mem_h *memh_p)
+static ucs_status_t
+uct_ib_mem_reg(uct_md_h uct_md, void *address, size_t length,
+               const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
+    uint64_t flags  = UCT_MD_MEM_REG_FIELD_VALUE(params, flags, FIELD_FLAGS,
```
yosefe, line 759: unsigned (multiple places)

shamisp:
> @shamisp @brminich @yosefe @tonycurtis as discussed today on bi-weekly UCX call, update the PR to have address/length as mandatory parameters

API changes looks good. thanks.

src/uct/ib/base/ib_md.c
dmitrygx: I checked all placed, it looks aligned

src/uct/base/uct_md.c
dmitrygx: converted to `uint64_t`

src/uct/cuda/cuda_copy/cuda_copy_md.c
dmitrygx: converted to `uint64_t`

yosefe (dismissed):

test/gtest/uct/test_md.h
```diff
@@ -10,6 +10,8 @@
 
 #include "uct_test.h"
 
+#include <uct/api/v2/uct_v2.h>
+
```
brminich, line 14: no need, it is already included in `test_md.cc`

dmitrygx: removed and force-pushed

yosefe (approved):

brminich (approved):

PR #8254, opened 2022-05-24
TOOLS/INFO: Add runtime library version to ucx_info -v output

ivankochin:
## What
Adding runtime linked libraries version to the `ucx_info -v` output.

## Why ?
Now `ucx_info -v` prints only version of the library headers that were used to compile this tool. But printing runtime libraries version may be useful too.

For example if customer has ucx shared objects but old `ucx_info` binary, then it still will report only the old version.

### DIFF
--- a/src/tools/info/build_info.c
+++ b/src/tools/info/build_info.c
@@ -13,13 +13,6 @@
 #include <ucs/sys/preprocessor.h>
 
 
-void print_version()
-{
-    printf("# Version %s\n", UCT_VERNO_STRING);
-    printf("# Git branch '%s', revision %s\n", UCT_SCM_BRANCH, UCT_SCM_VERSION);
-    printf("# Configured with: %s\n", UCX_CONFIGURE_FLAGS);
-}
-
 void print_build_config()
 {
     typedef struct {
--- a/src/tools/info/version_info.c
+++ b/src/tools/info/version_info.c
@@ -1,0 +1,24 @@
+/*
+ * Copyright (C) 2022, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "ucx_info.h"
+
+#include <ucs/sys/lib.h>
+#include <ucs/sys/string.h>
+
+void print_version()
+{
+    const char *path = ucs_sys_get_lib_path();
+
+    printf("# Library version: %s\n", ucp_get_version_string());
+    printf("# Library path: %s\n", (path != NULL) ? path : "<unknown>");
+    printf("# API headers version: %s\n", UCT_VERNO_STRING);
+    printf("# Git branch '%s', revision %s\n", UCT_SCM_BRANCH, UCT_SCM_VERSION);
+    printf("# Configured with: %s\n", UCX_CONFIGURE_FLAGS);
+}
--- a/src/tools/info/Makefile.am
+++ b/src/tools/info/Makefile.am
@@ -1,6 +1,7 @@
 #
 # Copyright (C) Mellanox Technologies Ltd. 2001-2014.  ALL RIGHTS RESERVED.
 # Copyright (C) The University of Tennessee and the University of Tennessee Research Foundation. 2016. ALL RIGHTS RESERVED.
+# Copyright (C) 2022, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
 #
 # See file LICENSE for terms.
 #
@@ -31,6 +32,7 @@ ucx_info_SOURCES  = \
 	sys_info.c \
 	tl_info.c \
 	type_info.c \
+	version_info.c \
 	ucx_info.c
 
 noinst_HEADERS = \


ivankochin:
There are two more questions:
- Should we report headers version and in what cases this information may be useful for customer?
- UCX contains several shared objects. Do we support the case when they have different version? If no, I think it may be senseful to move version related API to the UCS module to get rid of UCT/UCP version specific.

src/tools/info/build_info.c
```diff
@@ -15,7 +15,8 @@
 
 void print_version()
 {
-    printf("# Version %s\n", UCT_VERNO_STRING);
+    printf("# UCX version %s\n", ucp_get_version_string());
+    printf("# ucx_info UCX headers version %s\n", UCT_VERNO_STRING);
```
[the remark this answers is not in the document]
yosefe: is it to find issues of wrong loaded library?
maybe print the lib path?

src/tools/info/version_info.c
```diff
@@ -0,0 +1,33 @@
...
+
+
+    printf("# UCX libraries version: %s\n", ucp_get_version_string());
+    if (dladdr(ucp_get_version_string, &ucp_info)) {
+        lib_path = strdup(ucp_info.dli_fname);
+        printf("# UCX libraries path: %s\n", dirname(lib_path));
+        free(lib_path);
+    } else {
+        printf("# Cannot get path to UCX libraries\n");
+    }
```
yosefe, line 29: maybe use ucs_sys_get_lib_path instead? then no need -fPIC

src/tools/info/version_info.c
```diff
@@ -0,0 +1,33 @@
...
+#include <string.h>
+
+
+void print_version()
+{
+    Dl_info ucp_info;
+    char* lib_path;
+
+
+    printf("# UCX libraries version: %s\n", ucp_get_version_string());
```
yosefe, line 22: maybe just "Library version: %s"

src/tools/info/version_info.c
```diff
@@ -0,0 +1,33 @@
...
+    if (dladdr(ucp_get_version_string, &ucp_info)) {
+        lib_path = strdup(ucp_info.dli_fname);
+        printf("# UCX libraries path: %s\n", dirname(lib_path));
+        free(lib_path);
+    } else {
+        printf("# Cannot get path to UCX libraries\n");
+    }
+    printf("# UCX headers version: %s\n", UCT_VERNO_STRING);
```
yosefe, line 30: maybe just "API headers version: %s"

src/tools/info/version_info.c
```diff
@@ -0,0 +1,32 @@
...
+    const char *path_p;
+    char path[PATH_MAX];
+
+    printf("# Library version: %s\n", ucp_get_version_string());
+    path_p = ucs_sys_get_lib_path();
+    if (path_p != NULL) {
+        ucs_strncpy_safe(path, path_p, PATH_MAX);
+        printf("# Library path: %s\n", dirname(path));
+    } else {
+        printf("# Cannot get path to library\n");
```
yosefe, line 27: 1. better print full path, no need to remove file name, in case multiple SO files/symlink/etc
2. better keep "Library path" title in both cases to make the output more consistent
```c
path = ucs_sys_get_lib_path();
printf("# Library path: %s\n", (path != NULL) ? path : "<unknown>");
```

src/tools/info/version_info.c
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (C) 2022, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#ifdef HAVE_CONFIG_H
+#  include "config.h"
+#endif
+#include "ucx_info.h"
```
yosefe, line 9: space line before

src/tools/info/version_info.c
```diff
@@ -0,0 +1,32 @@
...
+
+#include <ucs/sys/lib.h>
+#include <ucs/sys/string.h>
+
+#include <libgen.h>
```
yosefe, line 14: are these includes still needed?

ivankochin: No, but I always thought that all courses shouldn't depend on the internal inclusions inside the headers. It may cause a problem if `ucs/sys/string.h` will stop using `libgen` for some reason. So I will better add `stdio.h` inclusion here then remove this one. WDYT?

yosefe: in UCX each header should be self-sufficient

ivankochin: Yes, but we use `dirname` call directly which is not required to be provided from any of UCX header and that's we I didn't remove this include.

yosefe: see another comment: "better print full path, no need to remove file name"

yosefe (dismissed):

src/tools/info/version_info.c
ivankochin: Done, but doesn't it feels strange that we prints only path to UCS library in this case. WDYTA printing path to UCT/UCP too (requires introducing additional APIs)?

yosefe: maybe add helper func to ucs like `const char *ucs_get_dl_path(void *func)` to get library path by symbol name?
use it from ucs_sys_get_lib_path and also from ucp_init_version, and ucx_info tool (with UCP symbols)

yosefe:
@ivankochin pls squash

ivankochin:
@yosefe @gleon99 pls approve

yosefe (approved):

PR #8255, opened 2022-05-25
UCT/SHM/KNEM: mark knem BW as dedicated

hoopoepg:
- due to last investigations KNEM transport has good
  multithreading support which provides good
  performance on runs with high PPN
- mark KNEM BW as dedicated


brminich (approved):

yosefe (approved):

hoopoepg:
@yosefe ok to merge?

PR #8257, opened 2022-05-26
UCT/IB/DEVX: Create CQ using DEVX

ivankochin:
## What
CQ creation via DEVX API + using DEVX event channel for handling completion events.

## Why ?
This way of creation CQ allows to support more FW features (e.g. enhanced CQE zipping)

## How ?

Before moving this PR out of draft state, several "preparation" requests were merged:
- #8297
- #8299
- #8300
- #8302

At this moment union with both verbs and DEVX cq representation was created inside `uct_ib_mlx5_cq_t`. There is no any practical sense of keeping verbs part there at this moment but it's there to save the consistency of storing corresponding objects on the same abstraction level. The question of removing verbs representation from that union can be discussed separately.


yosefe:
superseded by #8376

PR #8258, opened 2022-05-26
UCP/MM: Imporve user memh refcount handling

Artemy-Mellanox:


src/ucp/core/ucp_request.inl
```diff
@@ -395,13 +395,15 @@ ucp_request_send_state_init(ucp_request_t *req, ucp_datatype_t datatype,
 
     switch (datatype & UCP_DATATYPE_CLASS_MASK) {
     case UCP_DATATYPE_CONTIG:
+        req->send.state.dt.dt.contig.memh       = NULL;
```
yosefe, line 398: do we still need to keep md_map field in dt.contig?
what is the relation between dt.contig.md_map and dt.contig.memh->md_map?

Artemy-Mellanox: ucp_rma_basic_progress_get/put still using it, not sure if it's right

Artemy-Mellanox: ahh - ucp_rndv_rts_pack()

yosefe: maybe move it under "rndv" field in ucp_request_t::send?
I see this md_map is being used/initialized in non-rndv places as well, probably it's not needed

Artemy-Mellanox: md_map is not used out of rndv flow 
but ucp_request_t::send::state initialization is common code 
same is about ucp_request_t::recv::state - for ucp_rndv_rtr_pack() we may add md_map under ucp_request_t::recv::am, but ::recv::state is being initialized from ucp_rndv_receive() which is common for tag and am

yosefe: what if we add md_map to https://github.com/openucx/ucx/blob/a1492f96ba2721b766e8a4739294a263beb50244/src/ucp/core/ucp_request.h#L242 for rndv RTS, and https://github.com/openucx/ucx/blob/a1492f96ba2721b766e8a4739294a263beb50244/src/ucp/core/ucp_request.h#L298 in case of RTR (so it will be in "rndv_req" and not "rreq")
I assume we need md map for rndv to know which remote keys to pack?

src/ucp/core/ucp_mm.c
```diff
@@ -443,6 +445,25 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
+{
+    ucs_status_t status;
+
+    if (ucs_test_all_flags(memh->md_map, md_map) ||
+        ucp_memh_is_zero_length(memh)) {
+        return UCS_OK;
+    }
+
+    UCP_THREAD_CS_ENTER(&context->mt_lock);
+    status = ucp_memh_register(context, memh, ~memh->md_map & md_map,
```
yosefe, line 459: can we add assert to ucp_memh_register that context is locked?

src/ucp/core/ucp_mm.c
```diff
@@ -892,7 +914,8 @@ ucp_mm_get_alloc_md_map(ucp_context_h context, ucp_md_map_t *md_map_p)
 
         context->alloc_md_map_initialized = 1;
         context->alloc_md_map             = memh->md_map;
-        ucp_memh_put(context, memh, 1);
+        ucp_memh_invalidate(context, memh,
```
yosefe, line 917: can use  here

src/ucp/core/ucp_mm.c
```diff
@@ -996,6 +1019,11 @@ static void ucp_mem_rcache_dump_region_cb(void *rcontext, ucs_rcache_t *rcache,
     ucp_context_h context = rcontext;
     unsigned md_index;
 
+    if (memh->md_map == 0) {
+        ucs_string_buffer_appendf(&strb, "empty");
```
yosefe, line 1023: maybe "(not registered)"

src/ucp/core/ucp_request.c
```diff
@@ -330,101 +330,37 @@ static void ucp_request_mem_invalidate_completion(uct_completion_t *comp)
...
-                                          &params);
-        if (status != UCS_OK) {
-            ucs_warn("failed to dereg from md[%d]=%s: %s", md_index,
-                     context->tl_mds[md_index].rsc.md_name,
-                     ucs_status_string(status));
-            req->send.state.uct_comp.count--;
-        }
-        memh_index++;
+    if (req->send.state.dt.dt.contig.memh != NULL) {
+        ucp_memh_invalidate(context, req->send.state.dt.dt.contig.memh,
```
yosefe, line 348: where would we pass UCT_MD_MEM_DEREG_FLAG_INVALIDATE to the UCT-level deregistration?

src/ucp/core/ucp_request.c
```diff
@@ -330,101 +330,37 @@ static void ucp_request_mem_invalidate_completion(uct_completion_t *comp)
...
-        status      = uct_md_mem_dereg_v2(context->tl_mds[md_index].md,
-                                          &params);
-        if (status != UCS_OK) {
-            ucs_warn("failed to dereg from md[%d]=%s: %s", md_index,
-                     context->tl_mds[md_index].rsc.md_name,
-                     ucs_status_string(status));
-            req->send.state.uct_comp.count--;
-        }
-        memh_index++;
+    if (req->send.state.dt.dt.contig.memh != NULL) {
```
yosefe, line 347: why needed? seems it wasn't checked before this PR

src/ucp/core/ucp_request.c
```diff
@@ -461,51 +400,73 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_request_memory_reg,
          * Forget about what user provided and register what we need.
          */
         ucp_trace_req(req, "mds 0x%" PRIx64 " not registered - drop user memh",
-                      md_map & ~state->dt.contig.md_map);
+                      md_map & ~state->dt.contig.memh->md_map);
         req->flags             &= ~UCP_REQUEST_FLAG_USER_MEMH;
-        state->dt.contig.md_map = 0;
+        state->dt.contig.memh   = NULL;
```
yosefe, line 405: seems we can remove lines 388-406 now. we **can** mix user-provided memh with internal registration - by calling ucp_memh_update

src/ucp/core/ucp_request.c
```diff
@@ -461,51 +400,73 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_request_memory_reg,
...
         iovcnt = state->dt.iov.iovcnt;
         iov    = buffer;
-        dt_reg = ((state->dt.iov.dt_reg == NULL) ?
-                  ucs_calloc(iovcnt, sizeof(*dt_reg), "iov_dt_reg") :
-                  state->dt.iov.dt_reg);
-        if (NULL == dt_reg) {
-            status = UCS_ERR_NO_MEMORY;
-            goto err;
+        if (ucs_unlikely(state->dt.iov.memhs != NULL)) {
+            memhs = state->dt.iov.memhs;
```
yosefe, line 436: align =

src/ucp/core/ucp_request.c
```diff
@@ -461,51 +400,73 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_request_memory_reg,
...
+                ucs_free(memhs);
+                goto err;
+            }
+
+            ucp_trace_req(req, "mem reg iov %ld/%ld md_map 0x%" PRIx64
+                          "/0x%" PRIx64 " rereg %d", iov_it, iovcnt,
+                          memhs[iov_it]->md_map, reg_md_map, update);
         }
-        state->dt.iov.dt_reg = dt_reg;
+        state->dt.iov.memhs  = memhs;
```
yosefe, line 469: 1. maybe just assign it in line 439?
2. maybe ucp_memh_update could include checking ig memh is NULL and if yes call ucp_memh_get. it means it would accept ucp_mem_h* instead of ucp_mem_h

src/ucp/core/ucp_request.c
```diff
@@ -537,14 +498,17 @@ UCS_PROFILE_FUNC_VOID(ucp_request_memory_dereg, (context, datatype, state, req),
 
     switch (datatype & UCP_DATATYPE_CLASS_MASK) {
     case UCP_DATATYPE_CONTIG:
-        ucp_request_dt_dereg(context, &state->dt.contig, 1, req);
+        if (state->dt.contig.memh != NULL) {
+            ucp_request_dt_dereg(context, &state->dt.contig.memh, 1, req);
+            state->dt.contig.memh   = NULL;
```
yosefe, line 503: extra spaces before =

src/ucp/core/ucp_request.c
```diff
@@ -537,14 +498,17 @@ UCS_PROFILE_FUNC_VOID(ucp_request_memory_dereg, (context, datatype, state, req),
...
     case UCP_DATATYPE_IOV:
-        if (state->dt.iov.dt_reg != NULL) {
-            ucp_request_dt_dereg(context, state->dt.iov.dt_reg,
+        if (state->dt.iov.memhs != NULL) {
+            ucp_request_dt_dereg(context, state->dt.iov.memhs,
                                  state->dt.iov.iovcnt, req);
-            ucs_free(state->dt.iov.dt_reg);
-            state->dt.iov.dt_reg = NULL;
+            ucs_free(state->dt.iov.memhs);
+            state->dt.iov.memhs  = NULL;
```
yosefe, line 511: extra spaces before =

src/ucp/core/ucp_request.inl
```diff
@@ -672,10 +611,10 @@ ucp_send_request_set_user_memh(ucp_request_t *req, ucp_md_map_t md_map,
 
     /* req->send.state.dt should not be used with protov2 */
     ucs_assert(!req->send.ep->worker->context->config.ext.proto_enable);
-
     ucs_assert(!(req->flags & UCP_REQUEST_FLAG_USER_MEMH));
-    ucp_request_init_dt_reg_from_memh(req, md_map, param->memh,
-                                      &req->send.state.dt.dt.contig);
+
+    req->send.state.dt.dt.contig.memh = param->memh;
+    req->flags |= UCP_REQUEST_FLAG_USER_MEMH;
```
yosefe, line 617: align =

src/ucp/core/ucp_rkey.c
```diff
@@ -116,16 +116,17 @@ ucp_rkey_unpack_distance(const ucp_rkey_packed_distance_t *packed_distance,
...
 }
 
-static ssize_t
-ucp_rkey_pack_common(ucp_context_h context, ucp_md_map_t md_map,
-                     const uct_mem_h *memh, const ucp_memory_info_t *mem_info,
-                     ucp_sys_dev_map_t sys_dev_map,
-                     const ucs_sys_dev_distance_t *sys_distance, void *buffer,
-                     int sparse_memh, unsigned uct_flags)
+UCS_PROFILE_FUNC(ssize_t, ucp_rkey_pack_memh,
+                 (context, md_map, memh, mem_info, sys_dev_map, uct_flags,
```
yosefe, line 120: IMO sys_dev_map should be right before sys_dev_map

src/ucp/rndv/rndv.c
```diff
@@ -1343,41 +1329,25 @@ static void ucp_rndv_send_frag_rtr(ucp_worker_h worker, ucp_request_t *rndv_req,
...
 
-        alloc_md_map = 0;
-        /* TODO: Check if can avoid for inter-node eps */
-        ucs_for_each_bit(md_index, mdesc->memh->md_map) {
-            if (md_index == mdesc->memh->alloc_md_index) {
-                md_attr = &context->tl_mds[md_index].attr;
-                if (md_attr->cap.flags & UCT_MD_FLAG_RKEY_PTR) {
-                    alloc_md_map |= UCS_BIT(md_index);
-                }
-                break;
```
yosefe, line 1361: seems this logic was not preserved because it checks for UCT_MD_FLAG_RKEY_PTR and now code isn't

src/ucp/rndv/rndv.c
```diff
@@ -2026,7 +1998,7 @@ UCS_PROFILE_FUNC_VOID(ucp_rndv_put_pipeline_frag_get_completion, (self),
 
     freq->send.uct.func                  = ucp_rndv_progress_rma_put_zcopy;
     freq->send.lane                      = fsreq->send.lane;
-    freq->send.state.dt.dt.contig.md_map = 0;
+    freq->flags                         |= freq_flags;
```
yosefe, line 2001: seems weird, why need USER_MEMH in fragment?

src/ucp/core/ucp_request.c
evgeny-leksikov: > Maybe we still have a chance to obsolete UCT_MD_MEM_DEREG_FLAG_INVALIDATE/UCT_MD_MEM_DEREG_FIELD_COMPLETION and leave only UCT_MD_MKEY_PACK_FLAG_INVALIDATE to control invalidation behavior? @evgeny-leksikov?

in this case we must limit UCT reg implemetnation to guarantee that every call returns unique key, i.e. transport cannot cache it.

src/ucp/core/ucp_mm.c
```diff
@@ -361,22 +361,24 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
                     context->tl_mds[md_index].rsc.md_name,
                     ucs_status_string(status));
 
-            ucp_memh_dereg(context, memh, md_map_registered);
+            if (!(uct_flags & UCT_MD_MEM_FLAG_HIDE_ERRORS)) {
```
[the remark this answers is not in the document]
Artemy-Mellanox: hide errors in this case means best-effort
memory leak?

src/ucp/core/ucp_request.c
Artemy-Mellanox: don't we want to remove transport caches and leave only UCP level cache?

yosefe: Evgeny probably meant "transport cannot recycle the remote key value"

src/ucp/rndv/rndv.c
yosefe: it's something GPU specific. let's try to preserve the logic.
@bureddy and @Akshay-Venkatesh  may be able to provide more details

src/ucp/core/ucp_request.c
evgeny-leksikov: > don't we want to remove transport caches and leave only UCP level cache?

AFAIU we remove rcache from IB implementation only but in theory we cannot control all transports

> Evgeny probably meant "transport cannot recycle the remote key value"

I dont think we can avoid recycling all together, but removing existing flags progibites return the same rkey value for several "reg" calls in row similar like rcache behavior.

yosefe: since INVALIDATE is v2 API, we can change it to by synchronic, and remove completion for it altogether - but only after removing rcache from IB MD

src/ucp/core/ucp_request.c
```diff
@@ -444,29 +413,26 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_request_memory_reg,
...
+                                     flags, &memhs[iov_it]);
             if (status != UCS_OK) {
                 /* unregister previously registered memory */
-                ucp_request_dt_dereg(context, memhs, iov_it, req);
+                if (state->dt.iov.memhs == NULL) {
+                    ucp_request_dt_dereg(context, memhs, iov_it, req);
+                } else {
+                    ucp_request_dt_dereg(context, memhs, iovcnt, req);
+                    state->dt.iov.memhs = NULL;
+                }
```
evgeny-leksikov, line 426: is this condition for proto_v2? then better to check for `context->config.ext.proto_enable` or `req->flags & UCP_REQUEST_FLAG_PROTO_SEND`

src/ucp/core/ucp_mm.c
```diff
@@ -289,6 +289,16 @@ static inline int ucp_mem_map_is_allocate(const ucp_mem_map_params_t *params)
 static void ucp_memh_dereg(ucp_context_h context, ucp_mem_h memh,
                            ucp_md_map_t md_map)
 {
+    uct_completion_t comp = {
+        .count = 2
```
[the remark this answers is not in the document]
yosefe: yeah, if we don't expect uct_md_mem_dereg_v2 to really activate comp, it can be 1 and callback point to empty function that calls assert

src/ucp/core/ucp_mm.c
```diff
@@ -300,8 +310,15 @@ static void ucp_memh_dereg(ucp_context_h context, ucp_mem_h memh,
...
-        status = uct_md_mem_dereg(context->tl_mds[md_index].md,
-                                  memh->uct[md_index]);
+        if ((memh->flags & UCP_MEMH_FLAG_INVALIDATE) &&
+            (context->tl_mds[md_index].attr.cap.flags &
+             UCT_MD_FLAG_INVALIDATE)) {
+            params.memh = memh->uct[md_index];
+            status = uct_md_mem_dereg_v2(context->tl_mds[md_index].md, &params);
+        } else {
+            status = uct_md_mem_dereg(context->tl_mds[md_index].md,
+                                      memh->uct[md_index]);
```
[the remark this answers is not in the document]
yosefe: yeah, just make UCT_MD_MEM_DEREG_FLAG_INVALIDATE depend on that if condition

src/ucp/core/ucp_mm.c
```diff
@@ -340,15 +357,19 @@ void ucp_memh_unmap(ucp_context_h context, ucp_mem_h memh, ucp_md_map_t md_map)
...
     ucp_md_map_t md_map_registered = 0;
     ucs_log_level_t log_level;
     ucp_md_index_t md_index;
     ucs_status_t status;
+    void *address;
+    size_t length;
 
-    ucs_for_each_bit(md_index, md_map) {
+    address = (void*)memh->super.super.start;
+    length  = memh->super.super.end - memh->super.super.start;
```
yosefe, line 370: can we init inline?

src/ucp/core/ucp_mm.c
```diff
@@ -443,6 +461,29 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
     return status;
 }
 
+ucs_status_t ucp_memh_update(ucp_context_h context, void *address,
```
yosefe, line 464: IMO it should be inline function

src/ucp/core/ucp_mm.c
yosefe: can we use ucp_mem_unmap here?

src/ucp/core/ucp_mm.h
```diff
@@ -31,6 +36,7 @@ typedef struct ucp_mem {
     ucs_memory_type_t   mem_type;       /* Type of allocated or registered memory */
     ucp_md_index_t      alloc_md_index; /* Index of MD used to allocated the memory */
     ucp_md_map_t        md_map;         /* Which MDs have valid memory handles */
+    uint32_t            flags;
```
yosefe, line 39: move after alloc_md_index
+ pls add ucp_mem to object sizes gtest and output of `ucx_info -y` if not there already

src/ucp/core/ucp_mm.inl
```diff
@@ -73,13 +74,19 @@ ucp_memh_put(ucp_context_h context, ucp_mem_h memh, int invalidate)
     }
 
     UCP_THREAD_CS_ENTER(&context->mt_lock);
-    if (invalidate) {
-        ucs_rcache_region_invalidate(context->rcache, &memh->super,
-                (ucs_rcache_invalidate_comp_func_t)ucs_empty_function, NULL);
+    if (cb != NULL) {
+        memh->flags &= UCP_MEMH_FLAG_INVALIDATE;
```
yosefe, line 78: looks weird
did you want to remove this flag or add it? I would not expect we need to remove it

src/ucp/core/ucp_request.c
yosefe: how about:
```c
if (datatype-is-not-contig ||
    (req->send.state.dt.dt.contig.memh != NULL)) {
    /* Nothing to invalidate */
    ucp_request_complete_send(req, status);
    return;
}

ucp_memh_invalidate(...);
```

src/ucp/core/ucp_request.c
```diff
@@ -447,65 +384,56 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_request_memory_reg,
...
 
-    if (req->flags & UCP_REQUEST_FLAG_USER_MEMH) {
-        ucs_assert(UCP_DT_IS_CONTIG(datatype));
-
-        /* All memory domains that we need were provided by user memh */
-        if (ucs_likely(ucs_test_all_flags(state->dt.contig.md_map, md_map))) {
-            ucp_trace_req(req, "memh already registered");
-            return UCS_OK;
-        }
+    status     = UCS_OK;
```
yosefe, line 387: can we remove this?

src/ucp/core/ucp_request.c
```diff
@@ -447,65 +384,56 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_request_memory_reg,
...
+    flags      = UCT_MD_MEM_ACCESS_RMA | uct_flags;
+    reg_md_map = md_map & context->reg_md_map[mem_type];
```
yosefe, line 389: maybe init inline

src/ucp/core/ucp_request.inl
```diff
@@ -673,9 +612,8 @@ ucp_send_request_set_user_memh(ucp_request_t *req, ucp_md_map_t md_map,
     /* req->send.state.dt should not be used with protov2 */
     ucs_assert(!req->send.ep->worker->context->config.ext.proto_enable);
 
-    ucs_assert(!(req->flags & UCP_REQUEST_FLAG_USER_MEMH));
-    ucp_request_init_dt_reg_from_memh(req, md_map, param->memh,
-                                      &req->send.state.dt.dt.contig);
+    req->send.state.dt.dt.contig.memh = param->memh;
+    req->send.state.dt.dt.contig.memh->super.refcount++;
```
yosefe, line 616: 1. why needed - ucp_memh_get should add refcount?
2. can we use rcache API for this?

src/ucp/rndv/rndv.c
```diff
@@ -896,15 +894,8 @@ static ucs_status_t ucp_rndv_req_send_rma_get(ucp_request_t *rndv_req,
...
-        ucs_for_each_bit(lane, rndv_req->send.rndv.lanes_map_all) {
-            ucs_assert(lane < UCP_MAX_LANES);
-            md_map |= ucp_ep_md_index(ep, lane);
-        }
-
-        ucp_request_init_dt_reg_from_memh(rndv_req, md_map,
-                                          rreq->recv.user_memh,
-                                          &rndv_req->send.state.dt.dt.contig);
+        rndv_req->send.state.dt.dt.contig.memh = rreq->recv.user_memh;
+        rndv_req->send.state.dt.dt.contig.memh->super.refcount++;
```
yosefe, line 898: pls use rcache api

src/ucp/rndv/rndv.c
```diff
@@ -1001,13 +991,12 @@ ucp_rndv_init_mem_type_frag_req(ucp_worker_h worker, ucp_request_t *freq, int rn
...
         ucs_assert(mem_type_rma_lane != UCP_NULL_LANE);
 
-        freq->send.lane                       = mem_type_rma_lane;
-        freq->send.ep                         = mem_type_ep;
-        freq->send.state.dt.dt.contig.memh[0] = mdesc->memh->uct[md_index];
-        freq->send.state.dt.dt.contig.md_map  = UCS_BIT(md_index);
+        freq->send.lane                    = mem_type_rma_lane;
+        freq->send.ep                      = mem_type_ep;
+        freq->send.state.dt.dt.contig.memh = mdesc->memh;
+        freq->send.state.dt.dt.contig.memh->super.refcount++;
```
yosefe, line 999: pls use rcache api

src/ucp/rndv/rndv.c
```diff
@@ -1343,41 +1328,27 @@ static void ucp_rndv_send_frag_rtr(ucp_worker_h worker, ucp_request_t *rndv_req,
...
-        }
-        ucs_assert(memh_index <= UCP_MAX_OP_MDS);
-
         frndv_req->flags             = 0;
         frndv_req->send.ep           = rndv_req->send.ep;
         frndv_req->send.rndv.mdesc   = mdesc;
         frndv_req->send.pending_lane = UCP_NULL_LANE;
 
+        md_map = ucp_ep_config(rndv_req->send.ep)->key.rma_bw_md_map & mdesc->memh->md_map;
+        if ((mdesc->memh->alloc_md_index != UCP_NULL_RESOURCE) &&
```
yosefe, line 1344: maybe user helper var for mdesc->memh->alloc_md_index

src/ucp/rndv/rndv.c
```diff
@@ -1735,10 +1707,8 @@ UCS_PROFILE_FUNC_VOID(ucp_rndv_receive, (worker, rreq, rndv_rts_hdr, rkey_buf),
         if (!is_get_zcopy_failed || !UCP_MEM_IS_HOST(src_mem_type)) {
             if (rreq->flags & UCP_REQUEST_FLAG_USER_MEMH) {
                 /* At this point we know the datatype is contig */
-                ucp_request_init_dt_reg_from_memh(rreq,
-                                                  ep_config->key.rma_bw_md_map,
-                                                  rreq->recv.user_memh,
-                                                  &rreq->recv.state.dt.contig);
+                rreq->recv.state.dt.contig.memh = rreq->recv.user_memh;
+                rreq->recv.state.dt.contig.memh->super.refcount++;
```
yosefe, line 1711: same here, use rcache api

src/ucp/rndv/rndv.c
```diff
@@ -1747,6 +1717,9 @@ UCS_PROFILE_FUNC_VOID(ucp_rndv_receive, (worker, rreq, rndv_rts_hdr, rkey_buf),
             ucs_assert(rndv_rts_hdr->size <= rreq->recv.length);
             ucp_request_recv_buffer_reg(rreq, ep_config->key.rma_bw_md_map,
                                         rndv_rts_hdr->size);
+
+            rtr_md_map = ep_config->key.rma_bw_md_map &
+                rreq->recv.state.dt.contig.memh->md_map;
```
yosefe, line 1722: maybe ucp_rndv_req_send_rtr should always use
`(md_map & rreq->recv.state.dt.contig.memh->md_map)`
as the map of sending RTR.
in case of line 1718, we can set `rtr_md_map = ep_config->key.rma_bw_md_map`
and in case of line 1343, we can remove the 2nd `&&` argument

src/ucp/rndv/rndv.c
```diff
@@ -2268,6 +2240,7 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_rndv_atp_handler,
         ucp_rndv_recv_frag_put_mem_type(ucp_request_get_super(req), req, mdesc,
                                         req->recv.length,
                                         req->recv.frag.offset);
+        mdesc->memh->super.refcount--;
```
yosefe, line 2243: use rcache api

src/ucp/tag/offload.c
```diff
@@ -299,15 +299,20 @@ ucp_tag_offload_do_post(ucp_request_t *req)
                                         req->recv.length, req->recv.datatype,
                                         &req->recv.state, req->recv.mem_type,
                                         req, UCT_MD_MEM_FLAG_HIDE_ERRORS);
-        if ((status != UCS_OK) || !req->recv.state.dt.contig.md_map) {
+        if (status != UCS_OK) {
```
yosefe, line 302: could we return error in case of UCT_MD_MEM_FLAG_HIDE_ERRORS ?
change to ucs_assertv(status == UCS_OK)
then no need to use goto since block_mem_reg will be used only once

src/ucs/memory/rcache.inl
```diff
@@ -82,11 +82,11 @@ ucs_rcache_get_unsafe(ucs_rcache_t *rcache, void *address, size_t length,
 static UCS_F_ALWAYS_INLINE void
 ucs_rcache_region_put_unsafe(ucs_rcache_t *rcache, ucs_rcache_region_t *region)
 {
-    ucs_rcache_region_lru_add(rcache, region);
-
     ucs_assert(region->refcount > 0);
     if (ucs_unlikely(--region->refcount == 0)) {
         ucs_mem_region_destroy_internal(rcache, region);
+    } else {
+        ucs_rcache_region_lru_add(rcache, region);
```
yosefe, line 89: rcache bugfix?

src/ucp/core/ucp_mm.inl
```diff
@@ -73,39 +73,62 @@ not_found:
 }
 
 static UCS_F_ALWAYS_INLINE void
-ucp_memh_put(ucp_context_h context, ucp_mem_h memh)
+ucp_memh_invalidate(ucp_context_h context, ucp_mem_h memh,
```
yosefe, line 76: 1. I would revert to previous name (ucp_memh_put)
2. can we move some slow-path code to C file (e.g when no action is required, user memory handle, or rcache was disabled)

src/ucp/core/ucp_mm.inl
```diff
@@ -73,39 +73,62 @@ not_found:
...
         return;
     }
 
+    memh->inv_md_map |= inv_md_map;
+
     /* User memory handle, or rcache was disabled */
     if (ucs_unlikely(memh->parent != NULL)) {
-        ucp_memh_cleanup(context, memh);
-        ucs_free(memh);
+        if (!--memh->super.refcount) {
```
yosefe, line 93: is it ok to do this without context lock?

src/ucp/core/ucp_mm.inl
```diff
@@ -73,39 +73,62 @@ not_found:
...
+
+            if (cb != NULL) {
+                cb(arg);
+            }
+        }
         return;
     }
 
     UCP_THREAD_CS_ENTER(&context->mt_lock);
+    if (cb != NULL) {
```
yosefe, line 105: maybe check inv_md_map?

src/ucp/core/ucp_mm.inl
```diff
@@ -149,4 +172,50 @@ ucp_memh_is_iov_buffer_in_range(const ucp_mem_h memh, const void *buffer,
...
+    return status;
+}
+
+
+static UCS_F_ALWAYS_INLINE ucp_mem_h ucp_memh_hold(ucp_context_h context,
+                                                   ucp_mem_h memh)
+{
+    ucs_assert((memh->parent != NULL) || (context->rcache == NULL) ||
+               ucp_memh_is_zero_length(memh));
+    memh->super.refcount++;
```
yosefe, line 217: lock?

src/ucp/core/ucp_request.c
```diff
@@ -469,77 +435,67 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_request_memory_reg,
...
-                    ucp_request_dt_dereg(context, dt_reg, iov_it, req);
-                    ucs_free(dt_reg);
-                    goto err;
+            status = ucp_memh_update(context, iov[iov_it].buffer,
+                                     iov[iov_it].length, mem_type, reg_md_map,
+                                     flags, &memhs[iov_it]);
+            if (status != UCS_OK) {
+                /* unregister previously registered memory */
+                /* coverity[check_after_deref] */
+                if (state->dt.iov.memhs == NULL) {
```
yosefe, line 482: why would it be NULL?

src/ucp/core/ucp_request.c
```diff
@@ -564,20 +520,19 @@ UCS_PROFILE_FUNC_VOID(ucp_request_memory_dereg, (context, datatype, state, req),
...
 
-    if (req->flags & UCP_REQUEST_FLAG_USER_MEMH) {
-        return;
-    }
-
     switch (datatype & UCP_DATATYPE_CLASS_MASK) {
     case UCP_DATATYPE_CONTIG:
-        ucp_request_dt_dereg(context, &state->dt.contig, 1, req);
+        if (state->dt.contig.memh != NULL) {
+            ucp_request_dt_dereg(context, &state->dt.contig.memh, 1, req);
```
yosefe, line 526: maybe ucp_request_dt_dereg can handle NULL memh (ignore) ?

src/ucp/core/ucp_request.inl
```diff
@@ -693,14 +631,24 @@ ucp_recv_request_set_user_memh(ucp_request_t *req,
...
     req->recv.user_memh = param->memh;
     /* dt_reg will be updated later if the protocol needs it */
 
     return UCS_OK;
 }
 
+static UCS_F_ALWAYS_INLINE void
+ucp_request_use_user_memh(ucp_dt_state_t *state, ucp_request_t *rreq)
+{
+    if (rreq->flags & UCP_REQUEST_FLAG_USER_MEMH) {
```
yosefe, line 644: if (!...) return

test/gtest/uct/test_md.cc
```diff
@@ -104,29 +104,34 @@ void test_md::test_reg_mem(unsigned access_mask,
         params.field_mask = UCT_MD_MEM_DEREG_FIELD_COMPLETION |
                             UCT_MD_MEM_DEREG_FIELD_FLAGS |
                             UCT_MD_MEM_DEREG_FIELD_MEMH;
+        /* coverity[freed_arg:FALSE] */
```
yosefe, line 107: still needed?

src/ucp/core/ucp_mm.c
```diff
@@ -315,6 +315,16 @@ ucp_mem_map_params2uct_flags(const ucp_context_h context,
...
 {
+    uct_completion_t comp            = {
+        .count = 1,
+        .func  = (uct_completion_callback_t)ucs_empty_function_do_assert_void
+    };
+    uct_md_mem_dereg_params_t params = {
+        .field_mask = UCT_MD_MEM_DEREG_FIELD_MEMH |
+                      UCT_MD_MEM_DEREG_FIELD_FLAGS |
+                      UCT_MD_MEM_DEREG_FIELD_COMPLETION,
+        .comp       = &comp
```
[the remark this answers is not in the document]
Artemy-Mellanox: why? we overwrite it before use anyway

src/ucp/core/ucp_mm.c
```diff
@@ -315,6 +315,16 @@ ucp_mem_map_params2uct_flags(const ucp_context_h context,
 static void ucp_memh_dereg(ucp_context_h context, ucp_mem_h memh,
                            ucp_md_map_t md_map)
 {
+    uct_completion_t comp            = {
+        .count = 1,
+        .func  = (uct_completion_callback_t)ucs_empty_function_do_assert_void
```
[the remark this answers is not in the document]
Artemy-Mellanox: UCP rcache effectively makes any UCT entry uniquely referenced so callback should be called synchronously
we going to remove rcache from UCT anyway

src/ucp/core/ucp_mm.c
```diff
@@ -1267,7 +1280,6 @@ ucp_mem_rcache_mem_reg_cb(void *ctx, ucs_rcache_t *rcache, void *arg,
 
     ucp_memory_detect(context, address, length, &info);
     memh->context        = context;
-    memh->md_map         = 0;
```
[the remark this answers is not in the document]
Artemy-Mellanox: first I wanted to add inv_md_map=0, then noticed ucs_rcache_create_region does memset(0)

src/ucp/core/ucp_request.c
Artemy-Mellanox: first call

src/ucp/core/ucp_request.c
Artemy-Mellanox: it's a double pointer, so need to check for NULL for both memhs and all memhs[i], kinda awkward?

src/ucp/core/ucp_mm.c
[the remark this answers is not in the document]
yosefe: so could we remove invalidate from from **UCS** rcache after we remove the UCT rcaches?

src/ucp/core/ucp_request.c
yosefe: yes, just in ucp_request_dt_dereg

src/ucp/core/ucp_mm.inl
Artemy-Mellanox: we may want invalidate with inv_md_map 0

src/ucp/core/ucp_request.c
Artemy-Mellanox: we call ucp_request_dt_dereg four times and only in one case we check memh for NULL, so I think it's better to do it outside in this specific case

src/ucp/core/ucp_mm.inl
Artemy-Mellanox: 1. it's not rename, ucp_memh_put is next function
2. we need to identify those cases before calling

src/ucp/core/ucp_mm.c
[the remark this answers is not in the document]
Artemy-Mellanox: but we still use invalidate in UCP rcache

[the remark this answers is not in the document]
yosefe: but I think this time will not need callback API for ucs layer?

src/ucp/core/ucp_request.c
yosefe: ok

src/ucp/core/ucp_mm.c
```diff
@@ -372,12 +390,12 @@ void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
     }
 }
 
-static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
-                                      ucp_md_map_t md_map, void *address,
-                                      size_t length, ucs_memory_type_t mem_type,
-                                      unsigned uct_flags)
+ucs_status_t ucp_memh_register_slow(ucp_context_h context, ucp_mem_h memh,
```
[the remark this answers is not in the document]
Artemy-Mellanox: in get and update context we lock, in map context we don't

src/ucp/core/ucp_mm.inl
```diff
@@ -149,4 +177,54 @@ ucp_memh_is_iov_buffer_in_range(const ucp_mem_h memh, const void *buffer,
...
+                unsigned uct_flags, ucp_mem_h *memh_p)
+{
+    ucs_status_t status;
+
+    if (*memh_p == NULL) {
+        return ucp_memh_get(context, address, length, mem_type, md_map,
+                            uct_flags, memh_p);
+    }
+
+    if (ucs_test_all_flags((*memh_p)->md_map, md_map) ||
```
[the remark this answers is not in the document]
Artemy-Mellanox: it is but this is the way to avoid unnecessary lock

src/ucp/core/ucp_request.h
```diff
@@ -260,6 +260,9 @@ struct ucp_request {
 
                             /* Actual lanes count */
                             uint8_t        lanes_count;
+
+                            /* Memory domains to send remote keys */
+                            ucp_md_map_t   md_map;
```
Artemy-Mellanox, line 265: bad placement, we use `offset` from "new" union part in the same flow, see ucp_rndv_req_send_rtr

Artemy-Mellanox:
replaced by https://github.com/openucx/ucx/pull/9038

PR #8261, opened 2022-05-27
TEST/COMMENT: fixed typo in comment

hoopoepg:


test/gtest/uct/test_flush.cc
```diff
@@ -689,7 +689,7 @@ class uct_cancel_test : public uct_test {
 
         m_s0->connect();
 
-        /* there is a chance that one side getting disconect error before
```
brminich, line 692: plz also start with upper case

hoopoepg: done

yosefe (approved):

brminich (approved):

PR #8266, opened 2022-05-30
UCP/PROTO: Fix max fragment size calculation.

rakhmets:
## What
Fixed max fragment size calculation taking into account segment size.
Enabled new protocols for the appropriate test.

### DIFF
--- a/src/ucp/am/eager_multi.c
+++ b/src/ucp/am/eager_multi.c
@@ -41,7 +41,7 @@ ucp_am_eager_multi_bcopy_proto_init(const ucp_proto_init_params_t *init_params)
         .super.hdr_size      = sizeof(ucp_am_hdr_t),
         .super.send_op       = UCT_EP_OP_AM_BCOPY,
         .super.memtype_op    = UCT_EP_OP_GET_SHORT,
-        .super.flags         = 0,
+        .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_CAP_SEG_SIZE,
         .max_lanes           = context->config.ext.max_eager_lanes,
         .initial_reg_md_map  = 0,
         .first.lane_type     = UCP_LANE_TYPE_AM,
@@ -166,7 +166,8 @@ ucp_am_eager_multi_zcopy_proto_init(const ucp_proto_init_params_t *init_params)
         .super.hdr_size      = sizeof(ucp_am_hdr_t),
         .super.send_op       = UCT_EP_OP_AM_ZCOPY,
         .super.memtype_op    = UCT_EP_OP_LAST,
-        .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY,
+        .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY |
+                               UCP_PROTO_COMMON_INIT_FLAG_CAP_SEG_SIZE,
         .max_lanes           = context->config.ext.max_eager_lanes,
         .initial_reg_md_map  = 0,
         .first.lane_type     = UCP_LANE_TYPE_AM,
--- a/src/ucp/proto/proto_common.c
+++ b/src/ucp/proto/proto_common.c
@@ -22,6 +22,14 @@ ucp_proto_common_get_rsc_index(const ucp_proto_init_params_t *params,
     return params->ep_config_key->lanes[lane].rsc_index;
 }
 
+static size_t
+ucp_proto_common_get_seg_size(const ucp_proto_common_init_params_t *params,
+                              ucp_lane_index_t lane)
+{
+    ucs_assert(lane < UCP_MAX_LANES);
+    return params->super.ep_config_key->lanes[lane].seg_size;
+}
+
 void ucp_proto_common_lane_priv_init(const ucp_proto_common_init_params_t *params,
                                      ucp_md_map_t md_map, ucp_lane_index_t lane,
                                      ucp_proto_common_lane_priv_t *lane_priv)
@@ -150,7 +158,8 @@ size_t ucp_proto_common_get_iface_attr_field(const uct_iface_attr_t *iface_attr,
 static void
 ucp_proto_common_get_frag_size(const ucp_proto_common_init_params_t *params,
                                const uct_iface_attr_t *iface_attr,
-                               size_t *min_frag_p, size_t *max_frag_p)
+                               ucp_lane_index_t lane, size_t *min_frag_p,
+                               size_t *max_frag_p)
 {
     *min_frag_p = ucp_proto_common_get_iface_attr_field(iface_attr,
                                                         params->min_frag_offs,
@@ -158,6 +167,13 @@ ucp_proto_common_get_frag_size(const ucp_proto_common_init_params_t *params,
     *max_frag_p = ucp_proto_common_get_iface_attr_field(iface_attr,
                                                         params->max_frag_offs,
                                                         SIZE_MAX);
+
+    /* Adjust maximum fragment size taking into account segment size to prevent
+       sending more than the remote side supports. */
+    if (params->flags & UCP_PROTO_COMMON_INIT_FLAG_CAP_SEG_SIZE) {
+        *max_frag_p = ucs_min(ucp_proto_common_get_seg_size(params, lane),
+                              *max_frag_p);
+    }
 }
 
 static void ucp_proto_common_update_lane_perf_by_distance(
@@ -208,7 +224,7 @@ ucp_proto_common_get_lane_perf(const ucp_proto_common_init_params_t *params,
     uct_perf_attr_t perf_attr;
     ucs_status_t status;
 
-    ucp_proto_common_get_frag_size(params, &wiface->attr, &tl_min_frag,
+    ucp_proto_common_get_frag_size(params, &wiface->attr, lane, &tl_min_frag,
                                    &tl_max_frag);
 
     status = ucp_proto_common_lane_perf_attr(&params->super, lane,
@@ -332,7 +348,6 @@ static ucp_lane_index_t ucp_proto_common_find_lanes_internal(
                  UCT_TL_RESOURCE_DESC_ARG(&context->tl_rscs[rsc_index].tl_rsc));
 
         /* Check if lane type matches */
-        ucs_assert(lane < UCP_MAX_LANES);
         if (!(ep_config_key->lanes[lane].lane_types & UCS_BIT(lane_type))) {
             ucs_trace("%s: no %s in name types", lane_desc,
                       ucp_lane_type_info[lane_type].short_name);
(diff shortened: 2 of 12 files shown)


src/ucp/proto/proto_common.c
```diff
@@ -147,17 +147,35 @@ size_t ucp_proto_common_get_iface_attr_field(const uct_iface_attr_t *iface_attr,
...
+    const ucp_ep_config_key_lane_t *lane;
+
     *min_frag_p = ucp_proto_common_get_iface_attr_field(iface_attr,
                                                         params->min_frag_offs,
                                                         0);
     *max_frag_p = ucp_proto_common_get_iface_attr_field(iface_attr,
                                                         params->max_frag_offs,
                                                         SIZE_MAX);
+
+    ucs_assert(lane_idx < UCP_MAX_LANES);
```
yosefe, line 172: 1. pls add comment that we do it to not send more than what the remote side supports
2. i think we should decide about using seg_size according to the protocol logic of using this lane, and not according to the lane type. for example, pass uct opcode to ucp_proto_common_get_frag_size, deduce it from "params", or pass boolean parameter.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1093,6 +1093,17 @@ class test_ucp_am_nbx_seg_size : public test_ucp_am_nbx_reply {
...
+            modify_config("PROTO_ENABLE", "y");
+        }
+    }
+
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_values(variants, test_ucp_am_nbx_reply::get_test_variants,
+                           0);
+        add_variant_values(variants, test_ucp_am_nbx_reply::get_test_variants,
+                           1, "proto");
```
yosefe, line 1106: can we do this in test_ucp_am_nbx_reply (the base class)? or in test_ucp_am_nbx ?

src/ucp/proto/proto_common.c
```diff
@@ -150,14 +158,23 @@ size_t ucp_proto_common_get_iface_attr_field(const uct_iface_attr_t *iface_attr,
...
                                                         params->min_frag_offs,
                                                         0);
     *max_frag_p = ucp_proto_common_get_iface_attr_field(iface_attr,
                                                         params->max_frag_offs,
                                                         SIZE_MAX);
+
+    /* Adjust maximum fragment size taking into account segment size to prevent
+       sending more than the remote side supports. */
+    if ((params->send_op == UCT_EP_OP_AM_BCOPY) ||
+        (params->send_op == UCT_EP_OP_AM_ZCOPY)) {
```
brminich, line 174: it is also needed for `UCT_EP_OP_AM_SHORT` and `UCT_EP_OP_EAGER_(S|B|Z)*`

src/ucp/proto/proto_common.c
```diff
@@ -147,17 +155,36 @@ size_t ucp_proto_common_get_iface_attr_field(const uct_iface_attr_t *iface_attr,
     return *(const size_t*)UCS_PTR_BYTE_OFFSET(iface_attr, field_offset);
 }
 
+static int
+ucp_proto_common_adjust_max_frag(const ucp_proto_common_init_params_t *params)
+{
+    return (params->send_op == UCT_EP_OP_AM_SHORT) ||
+        (params->send_op == UCT_EP_OP_AM_BCOPY) ||
```
brminich, line 162: looking at this, maybe better define new flag in `ucp_proto_single_init_params_t.super.flags`, which would indicate whether max_frag_size adjustment is needed? Then we can set this flag in `proto_init` functions and will have to check for just 1 flag

brminich (approved):

src/ucp/proto/proto_common.h
```diff
@@ -50,6 +50,10 @@ typedef enum {
 
     /* Supports non-zero minimal fragment size */
     UCP_PROTO_COMMON_INIT_FLAG_MIN_FRAG      = UCS_BIT(7),
+
+    /* Adjust maximum fragment size to prevent sending more than the remote side
+       supports */
+    UCP_PROTO_COMMON_INIT_FLAG_ADJ_MAX_FRAG  = UCS_BIT(8)
```
yosefe, line 56: UCP_PROTO_COMMON_INIT_FLAG_CAP_SEG_SIZE or UCP_PROTO_COMMON_INIT_FLAG_MAX_SEG_SIZE,
or something else with "seg_size"

src/ucp/proto/proto_common.c
```diff
@@ -22,6 +22,14 @@ ucp_proto_common_get_rsc_index(const ucp_proto_init_params_t *params,
     return params->ep_config_key->lanes[lane].rsc_index;
 }
 
+size_t
+ucp_proto_common_get_seg_size(const ucp_proto_common_init_params_t *params,
+                              ucp_lane_index_t lane)
```
yosefe, line 27: 1. did it pass clang-format?
2. can it be static

src/ucp/proto/proto_common.c
```diff
@@ -150,14 +158,22 @@ size_t ucp_proto_common_get_iface_attr_field(const uct_iface_attr_t *iface_attr,
...
     *min_frag_p = ucp_proto_common_get_iface_attr_field(iface_attr,
                                                         params->min_frag_offs,
                                                         0);
     *max_frag_p = ucp_proto_common_get_iface_attr_field(iface_attr,
                                                         params->max_frag_offs,
                                                         SIZE_MAX);
+
+    /* Adjust maximum fragment size taking into account segment size to prevent
+       sending more than the remote side supports. */
+    if (params->flags & UCP_PROTO_COMMON_INIT_FLAG_ADJ_MAX_FRAG) {
```
yosefe, line 173: @brminich are you sure this is better than checking send_op? 
seems its equivalent to send_op == am_xx || eager_xx

src/ucp/rndv/rndv_rtr.c
```diff
@@ -53,7 +53,8 @@ ucp_proto_rndv_rtr_common_init(const ucp_proto_init_params_t *init_params,
         .super.hdr_size      = sizeof(ucp_rndv_rtr_hdr_t),
         .super.send_op       = UCT_EP_OP_AM_BCOPY,
         .super.memtype_op    = UCT_EP_OP_LAST,
-        .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_RESPONSE,
+        .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_RESPONSE |
+                               UCP_PROTO_COMMON_INIT_FLAG_ADJ_MAX_FRAG,
```
yosefe, line 57: looking at this, i think it can limit the size of RTR.size field and this is not what we want.
if it really happens, need to keep using a flag (and not send_op) and just remove this flag from here.

src/ucp/proto/proto_common.c
brminich: Imo, it is better than using `uct_ep_operation_t`, like it was before (otherwise need to have 6 branches).
But do you mean using `UCP_OP_ID_TAG_SEND` and `UCP_OP_ID_AM_SEND`? What shall we do with amo/rma emulation and rndv replies then?

src/ucp/rndv/rndv_rtr.c
brminich: did not get this comment, can you please elaborate?

src/ucp/proto/proto_common.c
rakhmets: yes (to both questions)

src/ucp/rndv/rndv_rtr.c
yosefe: rtr protocol is about fetching data from remote. so need to make sure this will not cause the size of the fetched data to be reduced.

src/ucp/am/eager_multi.c
```diff
@@ -167,7 +167,7 @@ ucp_am_eager_multi_zcopy_proto_init(const ucp_proto_init_params_t *init_params)
         .super.send_op       = UCT_EP_OP_AM_ZCOPY,
         .super.memtype_op    = UCT_EP_OP_LAST,
         .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY |
-                               UCP_PROTO_COMMON_INIT_FLAG_ADJ_MAX_FRAG,
+                                      UCP_PROTO_COMMON_INIT_FLAG_CAP_SEG_SIZE,
```
brminich: why need to add offset (here and in other places below)?

src/ucp/proto/proto_common.c
yosefe: branches should not be important here, it's slow-path. i means using UCT_EP_OP_xx

src/ucp/am/eager_multi.c
yosefe: in this case , need to ignore clang format

src/ucp/rndv/rndv_rtr.c
rakhmets: It means that we cannot use `send_op` for the purpose. `ucp_rndv_rtr_proto` has `send_op` equals to `UCT_EP_OP_AM_BCOPY`. But we do not want to limit RTR size. Than the flag can be used. And it should be removed from flags of protocol  `ucp_rndv_rtr_proto`.

src/ucp/am/eager_multi.c
rakhmets: fixed

src/ucp/rndv/rndv_rtr.c
rakhmets: Removed.

yosefe: "It means that we cannot use send_op for the purpose. "
can you pls check it's really the case (i am not sure). Also can you pls check it for rndv/rts protocol as well and remove from there if needed.

brminich: but we are sending rts and rtr as AM, why is seg_size limitation not relevant?

rakhmets: > but we are sending rts and rtr as AM, why is seg_size limitation not relevant?

Currently, fragment size affects protocol selection thresholds. In this PR we take into account segment size while retrieving fragment size. We choose protocol depending on message size. However, we don't take into account fragment size when we send RTS/RTR. Therefore, segment size is not counted.
The only place was in https://github.com/openucx/ucx/blob/db3606d6826c6c69692df22fdbfaa9dd23a1f4f1/src/ucp/core/ucp_am.c#L763

At least we can add an assert. E.g. here https://github.com/openucx/ucx/blob/db3606d6826c6c69692df22fdbfaa9dd23a1f4f1/src/ucp/proto/proto_single.inl#L41

yosefe: let's not handle ctrl message size limit for now, we'll do it when adding rndv/iov support

rakhmets: ok, 
@brminich, @yosefe do you have any other comments to this PR?

yosefe (approved):

PR #8267, opened 2022-05-30
UCT/IB/DEVX: Set QP ts_format to default

ivankochin:
## What
Set QP `ts_format` to default to inherit timestamp format from the system configuration.

## Why ?
Some HCAs supports only "free running" or only "real time" ts format, and if `ts_format` configured to the "free running" value (0), but system supports only "real time" mode, QP creation fails. To avoid it we should set QP `ts_format` to default mode.

## Testing
- Tested manually on ConnectX-6 Dx

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
+++ b/src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
@@ -1245,6 +1245,12 @@ static inline unsigned uct_ib_mlx5_qpc_cs_res(unsigned size, int dc)
                          UCT_IB_MLX5_QPC_CS_RES_DISABLE;
 }
 
+enum {
+    UCT_IB_MLX5_QPC_TS_FORMAT_FREE_RUNNING = 0x0,
+    UCT_IB_MLX5_QPC_TS_FORMAT_DEFAULT      = 0x1,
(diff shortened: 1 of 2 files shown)


yosefe (approved):

src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
```diff
@@ -1245,6 +1245,12 @@ static inline unsigned uct_ib_mlx5_qpc_cs_res(unsigned size, int dc)
                          UCT_IB_MLX5_QPC_CS_RES_DISABLE;
 }
 
+enum {
+    UCT_IB_MLX5_QPC_TIMESTAMP_FORMAT_FREE_RUNNING = 0x0,
```
gleon99, line 1249: Maybe `TIMESTAMP` -> `TS`? Both to be shorter, and for consistency with the fieldname.

brminich (approved):

yosefe (approved):

gleon99:
@ivankochin please squash

ivankochin:
@gleon99 @yosefe @brminich pls approve

brminich (approved):

PR #8270, opened 2022-06-01
UCS/MPOOL: add max_chunk_size check

ybedbug:
## What
add max_chunk_size check in ucs_mpool_init()

## Why ?
mpool may allocate a chunk not enough for even one element and code may get segv at corner case such as element size is less than alignment.

## How ?
Calculate min_chunk_size and compare to max_chunk_size in ucs_mpool_init(), return fail creating the memory pool for invalid input max_chunk_size .

### DIFF
--- a/src/ucs/datastruct/mpool.c
+++ b/src/ucs/datastruct/mpool.c
@@ -72,8 +72,17 @@ void ucs_mpool_params_reset(ucs_mpool_params_t *params)
     params->name            = "";
 }
 
+static size_t ucs_mpool_chunk_size(ucs_mpool_t *mp, unsigned num_elems)
+{
+    return sizeof(ucs_mpool_chunk_t) + mp->data->alignment +
+           (num_elems * ucs_mpool_elem_total_size(mp->data));
+}
+
 ucs_status_t ucs_mpool_init(const ucs_mpool_params_t *params, ucs_mpool_t *mp)
 {
+    size_t min_chunk_size;
+    ucs_status_t status;
+
     /* Check input values */
     if ((params->elem_size == 0) ||
         (params->align_offset > params->elem_size) ||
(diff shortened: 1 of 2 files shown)


src/ucs/datastruct/mpool.c
```diff
@@ -113,6 +121,15 @@ ucs_status_t ucs_mpool_init(const ucs_mpool_params_t *params, ucs_mpool_t *mp)
         goto err_strdup;
     }
 
+    min_chunk_size = ucs_mpool_chunk_size(mp, 1);
+    if (params->max_chunk_size < min_chunk_size) {
+        ucs_free(mp->data->name);
+        ucs_free(mp->data);
+        mp->data = NULL;
+        ucs_error("Invalid memory pool parameter: max_chunk_size");
+        return UCS_ERR_INVALID_PARAM;
```
yosefe, line 130: need to use goto error flow. something like:

```
    ucs_error("invalid memory pool parameter: chunk size is too small (%zu)", 
              params->max_chunk_size);
    status = UCS_ERR_INVALID_PARAM;
    goto err_free_name;
}

...
err_free_name:
     ucs_free(mp->data->name);
err_strdup:
    ucs_free(mp->data);
    mp->data = NULL;
    return status;
```

test/gtest/ucs/test_mpool.cc
```diff
@@ -126,6 +126,25 @@ UCS_TEST_F(test_mpool, wrong_ops) {
...
+
+    ucs_mpool_params_reset(&mp_params);
+    mp_params.elem_size       = header_size + data_size;
+    mp_params.align_offset    = header_size;
+    mp_params.alignment       = align;
+    mp_params.elems_per_chunk = 1;
+    mp_params.max_chunk_size  = mp_params.elems_per_chunk * mp_params.elem_size;
+    mp_params.ops             = &ops;
+    mp_params.name            = "tests";
+    status = ucs_mpool_init(&mp_params, &mp);
```
yosefe, line 144: `ucs_status_t status = ucs_mpool_init(&mp_params, &mp);`

test/gtest/ucs/test_mpool.cc
```diff
@@ -126,6 +126,25 @@ UCS_TEST_F(test_mpool, wrong_ops) {
...
+    EXPECT_TRUE(status == UCS_ERR_INVALID_PARAM);
```
yosefe, line 145: `EXPECT_EQ(UCS_ERR_INVALID_PARAM, status)`

src/ucs/datastruct/mpool.c
ybedbug: fixed accordingly

test/gtest/ucs/test_mpool.cc
ybedbug: fixed accordingly

yosefe (dismissed):

yosefe:
@ybedbug can you pls squash?

PR #8279, opened 2022-06-03
UCT/IB/DC: Fix managing FC_HARD_REQ resend time [v1.13.x]

dmitrygx:
## What

Fix managing `FC_HARD_REQ` resend time.

## Why ?

`iface->tx.fc_hard_req_resend_time` is initialized only, but then never updated. So, it always leads to invoking of `uct_dc_mlx5_ep_fc_hard_req_progress`.

## How ?

Backport of https://github.com/openucx/ucx/pull/8253


yosefe (approved):

yosefe:
@dmitrygx seems like need to port KSM/atomic PRs to v1.13.x to make testing pass

PR #8264, opened 2022-05-28
UCP/PROTO: Infra to print protocols performance

yosefe:
## Why
Add the infrastructure to print protocol performance visualization as .dot file. The output is demonstrated in #8233

PR #8233, opened 2022-05-17
UCP/PROTO: Add performance tree data structure

yosefe:
## Why
Start an infrastructure for debugging protocol selection. The below diagram will be produced (rendered with graphviz) and it shows the estimated performance of each protocol stage and how it's based on low-level UCT performance and system topology.
If a wrong protocol is selected, it means one (or more) of the performance estimation stages is wrong.

graphviz
[source](https://gist.github.com/yosefe/4a73f973fab024f978b3aa59f878b558#file-proto_node_demo-dot)

PR #8282, opened 2022-06-04
UCP/PROTO: Add performance tree data to non-rndv protocols

yosefe:
## Why
Follow up on https://github.com/openucx/ucx/pull/8233 and https://github.com/openucx/ucx/pull/8264, populate the performance information tree for non-rendezvous protocols (split to reduce the PR size).

### DIFF
--- a/src/ucp/proto/proto_common.c
+++ b/src/ucp/proto/proto_common.c
@@ -160,84 +160,159 @@ ucp_proto_common_get_frag_size(const ucp_proto_common_init_params_t *params,
                                                         SIZE_MAX);
 }
 
+/* Update 'perf' with the distance */
 static void ucp_proto_common_update_lane_perf_by_distance(
-        ucp_proto_common_tl_perf_t *perf,
-        const ucs_sys_dev_distance_t *distance)
+        ucp_proto_common_tl_perf_t *perf, ucp_proto_perf_node_t *perf_node,
+        const ucs_sys_dev_distance_t *distance, const char *perf_name,
+        const char *perf_fmt, ...)
 {
-    perf->bandwidth    = ucs_min(perf->bandwidth, distance->bandwidth);
-    perf->sys_latency += distance->latency;
-}
+    ucp_proto_perf_node_t *sys_perf_node;
+    ucs_linear_func_t distance_func;
+    char perf_node_desc[128];
+    va_list ap;
 
-ucs_status_t
-ucp_proto_common_lane_perf_attr(const ucp_proto_init_params_t *params,
-                                ucp_lane_index_t lane, uct_ep_operation_t op,
-                                uint64_t uct_field_mask,
-                                uct_perf_attr_t* perf_attr)
-{
-    ucp_worker_h worker        = params->worker;
-    ucp_rsc_index_t rsc_index  = ucp_proto_common_get_rsc_index(params, lane);
-    ucp_worker_iface_t *wiface = ucp_worker_iface(worker, rsc_index);
-    ucs_status_t status;
+    distance_func.c = distance->latency;
+    distance_func.m = 1.0 / distance->bandwidth;
 
-    /* Use the v2 API to query overhead and BW */
-    perf_attr->field_mask = UCT_PERF_ATTR_FIELD_OPERATION | uct_field_mask;
-    perf_attr->operation  = op;
-
-    status = uct_iface_estimate_perf(wiface->iface, perf_attr);
-    if (status != UCS_OK) {
-        ucs_error("failed to get iface %p performance: %s", wiface->iface,
-                  ucs_status_string(status));
+    if (ucs_linear_func_is_zero(distance_func, UCP_PROTO_PERF_EPSILON)) {
+        return;
     }
 
-    return status;
+    perf->bandwidth    = ucs_min(perf->bandwidth, distance->bandwidth);
+    perf->sys_latency += distance->latency;
+
+    va_start(ap, perf_fmt);
+    ucs_vsnprintf_safe(perf_node_desc, sizeof(perf_node_desc), perf_fmt, ap);
+    va_end(ap);
+
+    sys_perf_node = ucp_proto_perf_node_new_data(perf_name, "%s",
+                                                 perf_node_desc);
+    ucp_proto_perf_node_add_data(sys_perf_node, "", distance_func);
+    ucp_proto_perf_node_own_child(perf_node, &sys_perf_node);
+}
+
+static void ucp_proto_common_lane_perf_node(ucp_context_h context,
+                                            ucp_rsc_index_t rsc_index,
+                                            const uct_perf_attr_t *perf_attr,
+                                            ucp_proto_perf_node_t **perf_node_p)
+{
+    const uct_tl_resource_desc_t *tl_rsc = &context->tl_rscs[rsc_index].tl_rsc;
+    ucp_proto_perf_node_t *perf_node;
+
+    perf_node = ucp_proto_perf_node_new_data(
+            uct_ep_operation_names[perf_attr->operation],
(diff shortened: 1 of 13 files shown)


yosefe:
@brminich @rakhmets @dmitrygx can you pls take a look?

src/ucp/proto/proto_init.c
```diff
@@ -33,22 +33,25 @@ void ucp_proto_common_add_ppln_range(const ucp_proto_init_params_t *init_params,
...
-                                  frag_range->max_length) -
+                                  frag_size) -
             ucs_linear_func_apply(frag_range->perf[UCP_PROTO_PERF_TYPE_MULTI],
-                                  frag_range->max_length);
+                                  frag_size);
 
-    ppln_range->max_length = max_length;
-    ppln_range->node       = NULL;
+    ucs_trace("frag-size: %zd" UCP_PROTO_TIME_FMT(frag_overhead),
+              frag_range->max_length, UCP_PROTO_TIME_ARG(frag_overhead));
```
rakhmets, line 54:
```suggestion
              frag_size, UCP_PROTO_TIME_ARG(frag_overhead));
```

src/ucp/proto/proto_init.c
```diff
@@ -248,14 +295,16 @@ ucp_proto_init_parallel_stages(const ucp_proto_init_params_t *params,
 
 ucs_status_t
 ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
-                           const ucp_proto_common_tl_perf_t *perf,
+                           const ucp_proto_common_tl_perf_t *tl_perf,
+                           ucp_proto_perf_node_t *const tl_perf_node,
```
rakhmets, line 299: The method seems overloaded.
Is it possible to re-factor code a little? E.g. put calculation of sender overhead, transport time, receiver overhead into separate functions.

It is not obvious that the method does what it does from the signature. Guess the comment at the method declaration will help.

src/ucp/proto/proto_reconfig.c
```diff
@@ -69,24 +69,25 @@ static ucs_status_t ucp_proto_reconfig_progress(uct_pending_req_t *self)
...
-    init_params->caps->cfg_priority         = 0;
-    init_params->caps->min_length           = 0;
-    init_params->caps->num_ranges           = 1;
-    init_params->caps->ranges[0].max_length = SIZE_MAX;
-    init_params->caps->ranges[0].node       = NULL;
+    *init_params->priv_size         = 0;
+    init_params->caps->cfg_thresh   = UCS_MEMUNITS_INF;
+    init_params->caps->cfg_priority = 0;
+    init_params->caps->min_length   = 0;
+    init_params->caps->num_ranges   = 1;
```
dmitrygx, line 83: can we unite initializing `caps` here with `ucp_proto_select_caps_cleanup`? the only difference between them is `num_ranges` - can be passed as parameter to common function

src/ucp/proto/proto_common.c
```diff
@@ -521,45 +622,68 @@ ucp_proto_common_find_am_bcopy_hdr_lane(const ucp_proto_init_params_t *params)
...
+        md_name = context->tl_mds[md_index].rsc.md_name;
+        ucs_linear_func_add_inplace(memreg_time, md_attr->reg_cost);
+        ucs_trace("md %s" UCP_PROTO_PERF_FUNC_FMT(reg_cost), md_name,
                   UCP_PROTO_PERF_FUNC_ARG(&md_attr->reg_cost));
+
+        ucp_proto_perf_node_add_data(perf_node, md_name, md_attr->reg_cost);
     }
 
-    return reg_cost;
+    if (ucs_popcount(reg_md_map) > 1) {
```
brminich, line 656: minor: maybe use `is_power_of2` instead?

src/ucp/proto/proto_common.c
```diff
@@ -187,57 +210,117 @@ ucp_proto_common_lane_perf_attr(const ucp_proto_init_params_t *params,
...
+        ucp_proto_perf_node_add_scalar(perf_node, "send-pre",
+                                       perf_attr->send_pre_overhead);
+    }
+
+    if (uct_field_mask & UCT_PERF_ATTR_FIELD_SEND_POST_OVERHEAD) {
+        ucp_proto_perf_node_add_scalar(perf_node, "send-post",
+                                       perf_attr->send_post_overhead);
+    }
+
+    *perf_node_p = perf_node;
```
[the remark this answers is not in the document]
yosefe: this method is returning `perf_attr` along with `perf_node_p` that is the performance tree node associated with these attributes. So, I think those functionalities are tightly coupled. If we move the performance node initialization to a separate function, then any places that calls `ucp_proto_common_lane_perf_attr` would have to call that function as well.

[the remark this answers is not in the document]
yosefe: On the other hand, after refactoring as suggested in https://github.com/openucx/ucx/pull/8282#discussion_r890111234, it makes sense.

src/ucp/proto/proto_init.c
```diff
@@ -293,70 +293,68 @@ ucp_proto_init_parallel_stages(const ucp_proto_init_params_t *params,
...
     /* Add constant CPU overhead */
-    send_overhead.c                             += tl_perf->send_pre_overhead;
-    send_perf.perf[UCP_PROTO_PERF_TYPE_SINGLE]   = send_overhead;
-    send_perf.perf[UCP_PROTO_PERF_TYPE_MULTI]    = send_overhead;
-    send_perf.perf[UCP_PROTO_PERF_TYPE_MULTI].c += tl_perf->send_post_overhead;
-    ucp_proto_perf_range_add_data(&send_perf);
+    send_overhead.c                            += tl_perf->send_pre_overhead;
+    send_perf->perf[UCP_PROTO_PERF_TYPE_SINGLE] = send_overhead;
+    send_perf->perf[UCP_PROTO_PERF_TYPE_MULTI]  = send_overhead;
+    send_perf->perf[UCP_PROTO_PERF_TYPE_MULTI].c += tl_perf->send_post_overhead;
```
dmitrygx, line 338: [minor] unaligned by `=`
it seems it was applied by clang

dmitrygx (approved):

dmitrygx (approved):

src/ucp/proto/proto_common.c
brminich: do you think it's not worth it, because it is not a fast path?

yosefe: I think I fixed it:
https://github.com/openucx/ucx/pull/8282/commits/ea3c84fa2960fa9cf0a0ab53eb45ee9e484f303b#diff-9402784f6eb886015832f7c7ffd1ed06df79272660219f51ef42c6b7eac39ef3R648

yosefe: oh it's another place

rakhmets (commented):
Honestly, I can't say that I fully understood these changes. Could you please share a little more information.
What do we get when this PR will be merged? What should be done next to have the performance tree as debug tool of protocol selection?
You also mentioned that you split PR to reduce size. What is the other part? Will it enable rendezvous protocols?

yosefe:
> Honestly, I can't say that I fully understood these changes.
> What do we get when this PR will be merged? What should be done next to have the performance tree as debug tool of protocol selection?
> You also mentioned that you split PR to reduce size. What is the other part? Will it enable rendezvous protocols?

This PR populates the "performance tree" that shown in https://github.com/openucx/ucx/pull/8233 description, this tree is generated as .dot file by passing `UCX_PROTO_INFO_DIR=y` as env var.
This PR add the performance data generation to common functions (such as staging/pipeline) and eager protocols, and next PR add information for RNDV protocols.

rakhmets (approved):

brminich (dismissed):

PR #7155, opened 2021-07-26
UCP/PROTO: Preparations for adding rdnv/put/zcopy protocol

yosefe:
## Why
Adding rdnv/put/zcopy new protocol is a large change to it's split to 2 PRs.
This PR adds some infra changes, next PR will add the protocol itself

PR #8287, opened 2022-06-04
UCP/PROTOv2: Add rndv offload protos

brminich:
## What
Add tag offload rendezvous and offload via sw rndv protocols for protov2 infrastructure

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -304,9 +304,11 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, tm_force_thresh), UCS_CONFIG_TYPE_MEMUNITS},
 
   {"TM_SW_RNDV", "n",
-   "Use software rendezvous protocol with tag offload. If enabled, tag offload\n"
-   "mode will be used for messages sent with eager protocol only.",
-   ucs_offsetof(ucp_context_config_t, tm_sw_rndv), UCS_CONFIG_TYPE_BOOL},
+   "Use software rendezvous protocol even when tag matching offload is enabled.\n"
+   "In this case tag matching offload will be used for messages sent with eager\n"
+   "protocol only. If the value is set to \"try\", the rendezvous protocol is\n"
+   "selected automatically according to the performance characteristics.",
+   ucs_offsetof(ucp_context_config_t, tm_sw_rndv), UCS_CONFIG_TYPE_TERNARY},
 
   {"NUM_EPS", "auto",
    "An optimization hint of how many endpoints would be created on this context.\n"
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -2133,9 +2133,12 @@ static void ucp_ep_config_adjust_max_short(ssize_t *max_short,
  * user buffer when matched. Thus, minimum message size allowed to be sent with
  * RNDV protocol should be bigger than maximal possible SW RNDV request
  * (i.e. header plus packed keys size). */
-size_t ucp_ep_tag_offload_min_rndv_thresh(ucp_ep_config_t *config)
+size_t ucp_ep_tag_offload_min_rndv_thresh(ucp_context_h context,
+                                          const ucp_ep_config_key_t *key)
 {
-    return sizeof(ucp_rndv_rts_hdr_t) + config->rndv.rkey_size;
+    return sizeof(ucp_rndv_rts_hdr_t) +
+           ucp_rkey_packed_size(context, key->rma_bw_md_map,
+                                UCS_SYS_DEVICE_ID_UNKNOWN, 0);
 }
 
 static void ucp_ep_config_init_short_thresh(ucp_memtype_thresh_t *thresh)
@@ -2667,7 +2670,8 @@ ucs_status_t ucp_ep_config_init(ucp_worker_h worker, ucp_ep_config_t *config,
             config->tag.lane                   = lane;
             max_rndv_thresh                    = iface_attr->cap.tag.eager.max_zcopy;
             max_am_rndv_thresh                 = iface_attr->cap.tag.eager.max_bcopy;
-            min_rndv_thresh                    = ucp_ep_tag_offload_min_rndv_thresh(config);
+            min_rndv_thresh                    = ucp_ep_tag_offload_min_rndv_thresh(
+                                                     context, &config->key);
             min_am_rndv_thresh                 = min_rndv_thresh;
 
             ucs_assertv_always(iface_attr->cap.tag.rndv.max_hdr >=
--- a/src/ucp/rndv/proto_rndv.c
+++ b/src/ucp/rndv/proto_rndv.c
@@ -223,7 +223,8 @@ ucp_proto_rndv_ctrl_perf(const ucp_proto_init_params_t *params,
 }
 
 static ucs_status_t
-ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params)
+ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params,
+                              ucp_lane_index_t lane)
 {
     ucp_context_h context             = params->super.super.worker->context;
     ucp_proto_rndv_ctrl_priv_t *rpriv = params->super.super.priv;
@@ -234,12 +235,7 @@ ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params)
 
     select_param                   = params->super.super.select_param;
     *params->super.super.priv_size = sizeof(ucp_proto_rndv_ctrl_priv_t);
-
-    /* Find lane to send the initial message */
-    rpriv->lane = ucp_proto_common_find_am_bcopy_hdr_lane(&params->super.super);
-    if (rpriv->lane == UCP_NULL_LANE) {
-        return UCS_ERR_NO_ELEM;
-    }
+    rpriv->lane                    = lane;
 
     op_attr_mask = ucp_proto_select_op_attr_unpack(select_param->op_attr) &
                    UCP_OP_ATTR_FLAG_MULTI_SEND;
@@ -276,7 +272,8 @@ ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params)
 }
 
 ucs_status_t
-ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params)
+ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
+                         ucp_lane_index_t lane)
 {
     ucp_proto_rndv_ctrl_priv_t *rpriv = params->super.super.priv;
     const char *rndv_op_name          = ucp_operation_names[params->remote_op_id];
@@ -298,7 +295,7 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params)
     }
 
     /* Initialize 'rpriv' structure */
-    status = ucp_proto_rndv_ctrl_init_priv(params);
+    status = ucp_proto_rndv_ctrl_init_priv(params, lane);
     if (status != UCS_OK) {
         goto out;
     }
@@ -312,7 +309,6 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params)
     max_length     = ucs_min(params->super.max_length, max_length);
     ctrl_perf.node = ucp_proto_perf_node_new_data(params->ctrl_msg_name, "");
 
-    ucs_assert(params->super.send_op == UCT_EP_OP_AM_BCOPY);
     /* Set send_overheads to the time to send and receive RTS message */
     status = ucp_proto_rndv_ctrl_perf(&params->super.super, rpriv->lane,
                                       &send_time, &receive_time);
@@ -396,6 +392,22 @@ static size_t ucp_proto_rndv_thresh(const ucp_proto_init_params_t *init_params)
     return cfg->rndv_thresh;
 }
 
+ucs_status_t
+ucp_proto_rndv_ctrl_am_init(const ucp_proto_rndv_ctrl_init_params_t *params)
+{
+    ucp_lane_index_t lane;
+
+    /* Find lane to send the initial message */
+    lane = ucp_proto_common_find_am_bcopy_hdr_lane(&params->super.super);
+    if (lane == UCP_NULL_LANE) {
+        return UCS_ERR_NO_ELEM;
+    }
+
+    ucs_assert(params->super.send_op == UCT_EP_OP_AM_BCOPY);
+
+    return ucp_proto_rndv_ctrl_init(params, lane);
+}
+
 ucs_status_t ucp_proto_rndv_rts_init(const ucp_proto_init_params_t *init_params)
 {
     ucp_context_h context                    = init_params->worker->context;
@@ -426,7 +438,7 @@ ucs_status_t ucp_proto_rndv_rts_init(const ucp_proto_init_params_t *init_params)
         .md_map              = 0
     };
 
-    return ucp_proto_rndv_ctrl_init(&params);
+    return ucp_proto_rndv_ctrl_am_init(&params);
 }
 
 void ucp_proto_rndv_rts_query(const ucp_proto_query_params_t *params,
@@ -843,6 +855,11 @@ ucp_proto_rndv_handle_rtr(void *arg, void *data, size_t length, unsigned flags)
 
     ucp_trace_req(req, "RTR offset %zu length %zu/%zu req %p", rtr->offset,
                   rtr->size, req->send.state.dt_iter.length, req);
+
+    if (req->flags & UCP_REQUEST_FLAG_OFFLOADED) {
+        ucp_tag_offload_cancel_rndv(req);
+        ucs_assert(!ucp_ep_use_indirect_id(req->send.ep));
+    }
 
     /* RTR covers the whole send request - use the send request directly */
     ucs_assert(req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED);
--- a/src/ucp/rndv/proto_rndv.h
+++ b/src/ucp/rndv/proto_rndv.h
@@ -106,11 +106,22 @@ typedef struct {
 
     /* Map of mandatory mds which keys should be packed to the rkey */
     ucp_md_map_t                   md_map;
+
 } ucp_proto_rndv_ctrl_init_params_t;
 
 
+/* Initializes protocol which sends rendezvous control message using AM lane
+ * (e.g. RTS and ATS). */
 ucs_status_t
-ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params);
+ucp_proto_rndv_ctrl_am_init(const ucp_proto_rndv_ctrl_init_params_t *params);
+
+
+/* Initializes protocol which sends rendezvous control message using specified
+ * lane. Can be used by tag matching offload rendezvous protocols, which use
+ * tag lane for sending control messages. */
+ucs_status_t
+ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
+                         ucp_lane_index_t lane);
 
 
 ucs_status_t
--- a/src/ucp/tag/eager.h
+++ b/src/ucp/tag/eager.h
@@ -88,8 +88,8 @@ void ucp_tag_eager_sync_zcopy_req_complete(ucp_request_t *req, ucs_status_t stat
 void ucp_tag_eager_sync_zcopy_completion(uct_completion_t *self);
 
 static UCS_F_ALWAYS_INLINE int
-ucp_proto_eager_check_op_id(const ucp_proto_init_params_t *init_params,
-                            ucp_operation_id_t op_id, int offload_enabled)
+ucp_tag_eager_check_op_id(const ucp_proto_init_params_t *init_params,
+                          ucp_operation_id_t op_id, int offload_enabled)
 {
     return ucp_proto_init_check_op(init_params, UCS_BIT(op_id)) &&
            (offload_enabled ==
--- a/src/ucp/tag/offload.c
+++ b/src/ucp/tag/offload.c
@@ -633,7 +633,10 @@ void ucp_tag_offload_cancel_rndv(ucp_request_t *req)
         ucs_error("Failed to cancel tag rndv op %s", ucs_status_string(status));
     }
 
-    req->flags &= ~UCP_REQUEST_FLAG_OFFLOADED;
+    /* Rndv will be completed by SW, reinitialize the completed size needed for
+     * RTR, rkey_ptr, etc. This field is relevant for proto v2 only. */
+    req->send.state.completed_size = 0;
+    req->flags                    &= ~UCP_REQUEST_FLAG_OFFLOADED;
 }
 
 ucs_status_t ucp_tag_offload_start_rndv(ucp_request_t *sreq,
--- a/src/ucp/tag/offload/rndv.c
+++ b/src/ucp/tag/offload/rndv.c
@@ -1,0 +1,237 @@
+/**
+ * Copyright (C) 2023, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+ * See file LICENSE for terms.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <ucp/proto/proto_single.inl>
+#include <ucp/rndv/proto_rndv.h>
+#include <ucp/rndv/proto_rndv.inl>
+#include <ucp/tag/offload.h>
+#include <ucp/tag/tag_rndv.h>
+
+
+/* Init HW rendezvous as a single protocol which does not require operation on
+ * remote side (which is the case when tag is matched - everyting is done by the
+ * HW/FW).
+ */
+static ucs_status_t
+ucp_tag_rndv_offload_proto_init(const ucp_proto_init_params_t *init_params)
+{
+    ucp_worker_h worker                   = init_params->worker;
+    ucp_context_h context                 = worker->context;
+    ucp_proto_single_init_params_t params = {
+       .super.super         = *init_params,
+       .super.latency       = 0,
+       .super.overhead      = 40e-9,
+       .super.cfg_thresh    = context->config.ext.rndv_thresh,
+       .super.cfg_priority  = 60,
+       .super.min_length    = ucp_ep_tag_offload_min_rndv_thresh(
+                                  context, init_params->ep_config_key),
+       .super.max_length    = SIZE_MAX,
+       .super.min_iov       = 0,
+       .super.min_frag_offs = UCP_PROTO_COMMON_OFFSET_INVALID,
+       .super.max_frag_offs = ucs_offsetof(uct_iface_attr_t,
+                                           cap.tag.rndv.max_zcopy),
+       .super.max_iov_offs  = ucs_offsetof(uct_iface_attr_t,
+                                           cap.tag.rndv.max_iov),
+       .super.hdr_size      = 0,
+       .super.send_op       = UCT_EP_OP_RNDV_ZCOPY,
+       .super.memtype_op    = UCT_EP_OP_LAST,
+       .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY |
+                              UCP_PROTO_COMMON_INIT_FLAG_RECV_ZCOPY |
+                              UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG,
+       .super.exclude_map   = 0,
+       .lane_type           = UCP_LANE_TYPE_TAG,
+       .tl_cap_flags        = UCT_IFACE_FLAG_TAG_RNDV_ZCOPY
+    };
+
+    if (!ucp_tag_rndv_check_op_id(init_params) ||
+        (init_params->select_param->dt_class != UCP_DATATYPE_CONTIG)) {
+        return UCS_ERR_UNSUPPORTED;
+    }
+
+    return ucp_proto_single_init(&params);
+}
+
+static void
+ucp_tag_rndv_offload_completion(uct_completion_t *uct_comp)
+{
+    ucp_request_t *req = ucs_container_of(uct_comp, ucp_request_t,
+                                          send.state.uct_comp);
+
+    ucp_send_request_id_release(req);
+    ucp_proto_request_zcopy_complete(req, uct_comp->status);
+}
+
+static ucs_status_t
+ucp_tag_rndv_offload_request_init(ucp_request_t *req, ucp_md_map_t md_map,
+                                  uct_completion_callback_t comp_func,
+                                  unsigned uct_reg_flags, unsigned dt_mask)
+{
+    const ucp_proto_single_priv_t *spriv = req->send.proto_config->priv;
+    ucs_status_t status;
+
+    status = ucp_ep_resolve_remote_id(req->send.ep, spriv->super.lane);
+    if (status != UCS_OK) {
+        return status;
+    }
+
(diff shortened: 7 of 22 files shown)


brminich:
@hoopoepg, @Artemy-Mellanox, can you please review?

src/ucp/proto/proto_common.h
```diff
@@ -278,4 +278,12 @@ ucp_proto_common_buffer_copy_time(ucp_worker_h worker, const char *title,
                                   ucs_linear_func_t *copy_time,
                                   ucp_proto_perf_node_t **perf_node_p);
 
+
+ucp_lane_index_t ucp_proto_common_find_lanes_internal(
```
rakhmets, line 282: Why do we need this method in header file?

src/ucp/rndv/proto_rndv.c
```diff
@@ -202,12 +203,7 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params)
 
     select_param                   = params->super.super.select_param;
     *params->super.super.priv_size = sizeof(ucp_proto_rndv_ctrl_priv_t);
-
-    /* Find lane to send the initial message */
-    rpriv->lane = ucp_proto_common_find_am_bcopy_hdr_lane(&params->super.super);
-    if (rpriv->lane == UCP_NULL_LANE) {
-        return UCS_ERR_NO_ELEM;
-    }
+    rpriv->lane                     = lane;
```
rakhmets, line 206:
```suggestion
    rpriv->lane                    = lane;
```

src/ucp/rndv/proto_rndv.h
```diff
@@ -93,6 +93,18 @@ ucs_status_t
 ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params);
 
 
+ucs_status_t
+ucp_proto_rndv_ctrl_init_common(const ucp_proto_rndv_ctrl_init_params_t *params,
```
rakhmets, line 97: Please add brief function description, showing difference between using `ucp_proto_rndv_ctrl_init` and `ucp_proto_rndv_ctrl_init_common`, also to explain meaning of `common` postfix.

src/ucp/tag/eager.h
```diff
@@ -87,8 +87,8 @@ void ucp_tag_eager_sync_zcopy_req_complete(ucp_request_t *req, ucs_status_t stat
 void ucp_tag_eager_sync_zcopy_completion(uct_completion_t *self);
 
 static UCS_F_ALWAYS_INLINE int
-ucp_proto_eager_check_op_id(const ucp_proto_init_params_t *init_params,
-                            ucp_operation_id_t op_id, int offload_enabled)
+ucp_proto_tag_eager_check_op_id(const ucp_proto_init_params_t *init_params,
```
rakhmets, line 90: Could it be  `ucp_tag_eager_check_op_id` ?

brminich: why would it better? I'd keep `proto` in the name to highlight it is used by proto v2 only

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,292 @@
...
+
+static UCS_F_ALWAYS_INLINE int
+ucp_proto_tag_check_op_id(const ucp_proto_init_params_t *init_params)
+{
+    return (init_params->select_param->op_id == UCP_OP_ID_TAG_SEND) ||
+           (init_params->select_param->op_id == UCP_OP_ID_TAG_SEND_SYNC);
+}
+
+static UCS_F_ALWAYS_INLINE int
+ucp_proto_tag_offload_rndv_supported(const ucp_proto_init_params_t *init_params,
```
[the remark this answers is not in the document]
brminich: why?

src/ucp/tag/eager.h
brminich: ok, but this is the existing convention, so we have to either follow it or consider renaming existing names
@yosefe, wdyt?

yosefe: In general, I agree with @rakhmets ; we should use filename-consistent naming convention.
However in the time being IMO it's more important to distinguish new and old protocols. So far we used "ucp_proto_XX" prefix for everything related to new protocols. After we finish the transition, we would be free to rename according to file/directory hierarchy.

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,309 @@
...
+}
+
+UCS_PROFILE_FUNC(ucs_status_t, ucp_proto_tag_rndv_offload_progress, (self),
+                 uct_pending_req_t *self)
+{
+    ucp_request_t *req = ucs_container_of(self, ucp_request_t, send.uct);
+    ucp_ep_h ep        = req->send.ep;
+    size_t max_iov     = ucp_ep_config(ep)->tag.offload.max_rndv_iov;
+    uct_iov_t *iov     = ucs_alloca(max_iov * sizeof(uct_iov_t));
+    const ucp_proto_rndv_ctrl_priv_t *rpriv;
```
rakhmets, line 145: we only use `lane` of `rpriv`.
```suggestion
    ucp_lane_index_t lane;
...
    lane = req->send.proto_config->priv->lane;
...
    rndv_hdr.md_index = ucp_ep_md_index(ep, lane);
...
    rndv_op = uct_ep_tag_rndv_zcopy(ep->uct_eps[lane],
...
    return ucp_proto_single_status_handle(req, 1, NULL, lane,
```

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,309 @@
...
+{
+    return (init_params->select_param->op_id == UCP_OP_ID_TAG_SEND) ||
+           (init_params->select_param->op_id == UCP_OP_ID_TAG_SEND_SYNC);
+}
+
+static UCS_F_ALWAYS_INLINE int
+ucp_proto_tag_offload_rndv_supported(const ucp_proto_init_params_t *init_params,
+                                     ucp_proto_rndv_ctrl_init_params_t *params)
+{
+    ucp_worker_h worker   = init_params->worker;
```
rakhmets, line 27: Could be replaced by
```suggestion
    ucp_context_h context = init_params->worker->context;
```

src/ucp/core/ucp_context.c
```diff
@@ -256,9 +256,10 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, tm_force_thresh), UCS_CONFIG_TYPE_MEMUNITS},
 
   {"TM_SW_RNDV", "n",
-   "Use software rendezvous protocol with tag offload. If enabled, tag offload\n"
-   "mode will be used for messages sent with eager protocol only.",
-   ucs_offsetof(ucp_context_config_t, tm_sw_rndv), UCS_CONFIG_TYPE_BOOL},
+   "Use software rendezvous protocol even when tag matching offload is enabled.\n"
+   "In this case tag matching offload will be used for messages sent with eager\n"
+   "protocol only.",
```
rakhmets, line 261: I am not sure I understand the behavior of `try` option from this description.
Maybe it is better to describe behavior of each variant: 
`yes` - tag matching offload will be used for messages sent with eager protocol only
`try` - ...
`no` - ...

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,309 @@
...
+        .super.send_op       = UCT_EP_OP_EAGER_SHORT,
+        .super.memtype_op    = UCT_EP_OP_LAST,
+        .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_RESPONSE,
+        .remote_op_id        = UCP_OP_ID_RNDV_RECV,
+        .unpack_time         = UCS_LINEAR_FUNC_ZERO,
+        .perf_bias           = context->config.ext.rndv_perf_diff / 100.0,
+        .mem_info.type       = init_params->select_param->mem_type,
+        .mem_info.sys_dev    = init_params->select_param->sys_dev
+    };
+    ucp_proto_rndv_ctrl_priv_t *rpriv = init_params->priv;
```
rakhmets, line 244: Please fix alignment

src/ucp/tag/eager.h
rakhmets: I'd rather use separate files.
E.g. all new AM protocols are in `src/ucp/am `.
I think using `ucp_proto_` fits for common functions, infrastructure, etc. But we can avoid using of it for protocol implementations. So there is no need to update names in the future.

src/ucp/tag/offload/rndv.c
brminich: `worker` is also used in this func

rakhmets: line 35:
```C
md_attr = &context->tl_mds[md_index].attr;
```

line 40:
```C
(context->config.ext.tm_sw_rndv == UCS_YES)) {
```

src/ucp/tag/offload/rndv.c
```diff
@@ -145,20 +138,22 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_proto_tag_rndv_offload_progress, (self),
...
-    rpriv  = req->send.proto_config->priv;
-    status = UCS_PROFILE_CALL(ucp_proto_tag_rndv_offload_rts_request_init, req);
+    status = UCS_PROFILE_CALL(ucp_tag_rndv_offload_request_init, req);
     if (status != UCS_OK) {
         ucp_proto_request_abort(req, status);
         return UCS_OK;
     }
 
+    rpriv             = req->send.proto_config->priv;
+    lane              = rpriv->lane;
```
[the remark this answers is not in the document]
brminich: it will work only if directly cast priv to rpriv type, which looks quite ugly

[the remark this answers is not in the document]
rakhmets: I see.

src/ucp/tag/eager.h
rakhmets, line 90: ucp_tag_eager_check_op_id?

rakhmets (dismissed):

rakhmets (dismissed):

src/ucp/proto/proto_common.h
```diff
@@ -278,4 +278,5 @@ ucp_proto_common_buffer_copy_time(ucp_worker_h worker, const char *title,
                                   ucs_linear_func_t *copy_time,
                                   ucp_proto_perf_node_t **perf_node_p);
 
+
```
yosefe, line 281: revert?

src/ucp/rndv/proto_rndv.c
```diff
@@ -15,7 +15,7 @@
 #include <ucp/proto/proto_common.inl>
 
 
-static void
+void
```
yosefe, line 18: pls check clang-format

src/ucp/rndv/proto_rndv.c
```diff
@@ -179,7 +179,8 @@ ucp_proto_rndv_ctrl_perf(const ucp_proto_init_params_t *params,
 }
 
 ucs_status_t
-ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params)
+ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
+                         ucp_lane_index_t lane)
```
yosefe, line 183: add lane to ucp_proto_rndv_ctrl_init_params_t

src/ucp/rndv/proto_rndv.c
```diff
@@ -233,9 +228,6 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params)
                                     UCP_DATATYPE_CONTIG, &mem_info, 1);
     }
 
-    /* Initialize estimated memory registration map */
-    ucp_proto_rndv_ctrl_get_md_map(params, &rpriv->md_map, &rpriv->sys_dev_map,
```
yosefe, line 237: why removed?
i think rndv/offload does not have to be derived from the common rndv protocol since in case of expected it will not require an operation on the remote side, and will be fully completed in HW.

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,303 @@
...
+    const uct_md_attr_t *md_attr;
+
+    /* Multirail is not supported with tag offload rndv, need to register
+     * memory on tag lane only. */
+    md_index = ucp_proto_common_get_md_index(init_params, lane);
+    md_attr  = &context->tl_mds[md_index].attr;
+
+    /* Not supported if can't register the buffer on tag lane MD (e.g. can't
+     * send GPU memory if no GPU_DIRECT_RDMA configured). */
+    if (!(md_attr->cap.reg_mem_types & UCS_BIT(params->mem_info.type)) ||
```
yosefe, line 32: this logic should be done by ucp_proto_common_find_lanes, when we derive the rndv/offload protocol from ucp_proto_single_init

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,303 @@
...
+
+    rpriv             = req->send.proto_config->priv;
+    lane              = rpriv->lane;
+    rndv_hdr.ep_id    = ucp_send_request_get_ep_remote_id(req);
+    rndv_hdr.req_id   = ucp_send_request_get_id(req);
+    rndv_hdr.md_index = ucp_ep_md_index(ep, lane);
+
+    ucs_assert(!ucp_ep_use_indirect_id(req->send.ep));
+
+    iov_cnt = ucp_datatype_iter_next_iov(&req->send.state.dt_iter, SIZE_MAX,
```
yosefe, line 160: can we reuse ucp_proto_zcopy_single_progress?

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,303 @@
...
+ucp_tag_rndv_offload_proto_query(const ucp_proto_query_params_t *params,
+                                 ucp_proto_query_attr_t *attr)
+{
+    UCS_STRING_BUFFER_FIXED(strb, attr->config, sizeof(attr->config));
+    const ucp_proto_rndv_ctrl_priv_t *rpriv = params->priv;
+    ucp_context_h context                   = params->worker->context;
+    const ucp_ep_config_key_lane_t *ep_lane_cfg;
+    const ucp_tl_resource_desc_t *rsc;
+
+    ucs_assert(rpriv->lane < UCP_MAX_LANES);
```
yosefe, line 190: seems redundant

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,303 @@
...
+    rsc         = &context->tl_rscs[ep_lane_cfg->rsc_index];
+
+    ucp_proto_default_query(params, attr);
+    ucs_string_buffer_appendf(&strb, UCT_TL_RESOURCE_DESC_FMT,
+                              UCT_TL_RESOURCE_DESC_ARG(&rsc->tl_rsc));
+}
+
+ucp_proto_t ucp_tag_rndv_offload_proto = {
+    .name     = "tag/rndv/offload",
+    .desc     = "rndv offload",
```
yosefe, line 201: "rendezvous tag offload"

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,303 @@
...
+        (init_params->select_param->dt_class != UCP_DATATYPE_CONTIG) ||
+        !ucp_ep_config_key_has_tag_lane(init_params->ep_config_key)) {
+        return UCS_ERR_UNSUPPORTED;
+    }
+
+    if (ucp_tag_rndv_offload_supported(init_params, &params) &&
+        (worker->context->config.ext.tm_sw_rndv == UCS_NO)) {
+        /* Do not completely disable SW rndv, because it is used when hw offload
+         * protocol can't be used. */
+        params.super.min_length = ep_config->tag.offload.max_rndv_zcopy + 1;
```
yosefe, line 251: we prefer to be independent from ep_config and rely only on init_params->ep_config_key

src/ucp/tag/offload/rndv.c
brminich: it's needed to avoid coverity error.
I think [here](https://github.com/openucx/ucx/blob/master/src/ucp/proto/proto_common.c#L84) it is done for the same reason

src/ucp/tag/offload/rndv.c
brminich: can you please elaborate?

src/ucp/tag/offload/rndv.c
yosefe: ok
maybe unite to helper func

src/ucp/tag/offload/rndv.c
yosefe: when new protocols take over, we plan to remote most of ep_config content

evgeny-leksikov:
@brminich can we remove [this](https://github.com/openucx/ucx/blob/b5d81f91bd4c1ffe4082d68ea5dbc7344653a464/test/gtest/ucp/test_ucp_tag_offload.cc#L720) in this PR? it can be tested manually with `UCX_PROTO_ENABLE=y gtest ...`

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,272 @@
...
+        ucp_tag_rndv_offload_supported(&hw_rndv_params, init_params)) {
+        /* SW rndv is suppressed by the configuration and HW offload rndv can be
+         * used with these selection parameters. Do not completely disable SW
+         * rndv, because it may be used for large messages which are not
+         * supported by the HW offload rndv protocol.
+         */
+        params.super.min_length = ep_config->tag.offload.max_rndv_zcopy + 1;
+    } else {
+        params.super.min_length = ucp_ep_tag_offload_min_rndv_thresh(ep_config);
+    }
```
brminich, line 230: remove this, always enable sw emulation

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+    ucp_ep_config_t *ep_config = &ucs_array_elem(&worker->ep_config,
+                                                 init_params->ep_cfg_index);
+    ucp_proto_single_init_params_t params = {
+       .super.super         = *init_params,
+       .super.latency       = 0,
+       .super.overhead      = 40e-9,
+       .super.cfg_thresh    = context->config.ext.rndv_thresh,
+       .super.cfg_priority  = 60,
+       .super.min_length    = ucp_ep_tag_offload_min_rndv_thresh(ep_config),
+       .super.max_length    = ep_config->tag.offload.max_rndv_zcopy,
```
yosefe, line 35: we should take this from user config
since ep_config is result of protov1 logic calculation

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+       .super.min_iov       = 0,
+       .super.min_frag_offs = UCP_PROTO_COMMON_OFFSET_INVALID,
+       .super.max_frag_offs = UCP_PROTO_COMMON_OFFSET_INVALID,
```
yosefe, line 38: is max_frag really undefined?

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+                                           cap.tag.rndv.max_iov),
+       .super.hdr_size      = 0,
+       .super.send_op       = UCT_EP_OP_RNDV_ZCOPY,
+       .super.memtype_op    = UCT_EP_OP_LAST,
+       .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY |
+                              UCP_PROTO_COMMON_INIT_FLAG_RECV_ZCOPY |
+                              UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG,
+       .super.exclude_map   = 0,
+       .lane_type           = UCP_LANE_TYPE_TAG,
+       .tl_cap_flags        = UCT_IFACE_FLAG_GET_ZCOPY |
```
yosefe, line 49: why need GET_ZCOPY?

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+    const ucp_proto_single_priv_t *spriv = req->send.proto_config->priv;
+    ucs_status_t status;
+
+    status = ucp_ep_resolve_remote_id(req->send.ep, spriv->super.lane);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    ucs_assert(md_map != 0);
+    ucs_assertv(dt_mask == UCS_BIT(UCP_DATATYPE_CONTIG), "dt_mask 0x%x",
```
yosefe, line 85: dt_mask=0x%x

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+{
+    const ucp_proto_single_priv_t *spriv = req->send.proto_config->priv;
+    ucs_status_t status;
+
+    status = ucp_ep_resolve_remote_id(req->send.ep, spriv->super.lane);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    ucs_assert(md_map != 0);
```
yosefe, line 84: why needed?

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+    ucs_assertv(dt_mask == UCS_BIT(UCP_DATATYPE_CONTIG), "dt_mask 0x%x",
+                dt_mask);
+
+    status = ucp_proto_request_zcopy_init(req, md_map, comp_func, uct_reg_flags,
+                                          dt_mask);
+    if (status == UCS_OK) {
```
yosefe, line 90: if (status!=OK) return

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+                                                         spriv->super.lane),
+                                    req->send.msg_proto.tag,
+                                    &rndv_hdr, sizeof(rndv_hdr), iov, 1, 0,
+                                    &req->send.state.uct_comp);
+
+    if (ucs_unlikely(UCS_PTR_IS_ERR(rndv_op))) {
+        return UCS_PTR_STATUS(rndv_op);
+    }
+
+    req->flags                   |= UCP_REQUEST_FLAG_OFFLOADED;
```
yosefe, line 119: maybe it should be part of ucp_tag_rndv_offload_request_init ?

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+    ucp_context_h context      = worker->context;
+    ucp_ep_config_t *ep_config = &ucs_array_elem(&worker->ep_config,
+                                                 init_params->ep_cfg_index);
+    ucp_proto_rndv_ctrl_init_params_t params = {
+        .super.super         = *init_params,
+        .super.latency       = 0,
+        .super.overhead      = 40e-9,
+        .super.cfg_thresh    = context->config.ext.rndv_thresh,
+        .super.cfg_priority  = 60,
+        .super.min_length    = ucp_ep_tag_offload_min_rndv_thresh(ep_config),
```
yosefe, line 165: let's try to not reuse protov1 logic

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+        .super.max_length    = SIZE_MAX,
+        .super.min_iov       = 0,
+        .super.min_frag_offs = UCP_PROTO_COMMON_OFFSET_INVALID,
+        .super.max_frag_offs = ucs_offsetof(uct_iface_attr_t,
+                                            cap.tag.eager.max_short),
```
yosefe, line 170: why max_short?

src/ucp/tag/offload/rndv.c
```diff
@@ -0,0 +1,232 @@
...
+        .super.max_iov_offs  = UCP_PROTO_COMMON_OFFSET_INVALID,
+        .super.hdr_size      = 0,
+        .super.send_op       = UCT_EP_OP_EAGER_SHORT,
```
yosefe, line 173: why only short? SW TM RTS can also use bcopy..

src/uct/ib/rc/accel/rc_mlx5.inl
```diff
@@ -1218,7 +1218,8 @@ uct_rc_mlx5_iface_handle_expected(uct_rc_mlx5_iface_common_t *iface, struct mlx5
     uct_rc_mlx5_iface_tag_del_from_hash(iface, priv->buffer);
 
     if (is_inline) {
-        ucs_assert(byte_len <= priv->length);
+        ucs_assertv(byte_len <= priv->length, "byte_len %u length %u",
```
yosefe, line 1221: "byte_len=%u length=%u",

src/ucp/tag/offload/rndv.c
brminich: this restriction is still relevant

src/ucp/tag/offload/rndv.c
brminich: i'd keep it here, because it means that rndv operation was sucesfully sent thru UCT TAG API and we need to cancel it there.

src/ucp/tag/offload/rndv.c
brminich: otherwise it means that rndv offload can't be used

src/ucp/tag/offload/rndv.c
brminich: yes, but we limit `super.max_length`

src/ucp/rndv/proto_rndv.c
```diff
@@ -396,6 +390,21 @@ static size_t ucp_proto_rndv_thresh(const ucp_proto_init_params_t *init_params)
...
+    /* Find lane to send the initial message */
+    params->lane = ucp_proto_common_find_am_bcopy_hdr_lane(
+                              &params->super.super);
+    if (params->lane == UCP_NULL_LANE) {
+        return UCS_ERR_NO_ELEM;
+    }
+
+    ucs_assert(params->super.send_op == UCT_EP_OP_AM_BCOPY);
+
+    return ucp_proto_rndv_ctrl_init(params);
```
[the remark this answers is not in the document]
brminich: it is consistent with other params. 
also would need to add the same code (finding am lane and checking for errors) to many places

src/ucp/tag/offload.c
```diff
@@ -633,7 +633,10 @@ void ucp_tag_offload_cancel_rndv(ucp_request_t *req)
         ucs_error("Failed to cancel tag rndv op %s", ucs_status_string(status));
     }
 
-    req->flags &= ~UCP_REQUEST_FLAG_OFFLOADED;
+    /* Rndv will be completed by SW, reinitialize the completed size needed for
+     * RTR, rkey_ptr, etc. */
+    req->send.state.completed_size = 0;
```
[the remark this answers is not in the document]
brminich: yes, this field was added for new protos in #7155

src/ucp/tag/offload/rndv.c
brminich: it is not explicitly stated in the API, but uct_ep_tag_rndv_request uses short protocol

src/ucp/tag/offload/rndv.c
yosefe: wouldn't it fail in ucp_proto_request_zcopy_init?

src/ucp/rndv/proto_rndv.c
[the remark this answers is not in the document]
yosefe: we still keep ucp_proto_rndv_ctrl_am_init function to avoid code duplication, like this
```c
ucp_proto_rndv_ctrl_am_init(const ucp_proto_rndv_ctrl_init_params_t *params)
{
     /* Find lane to send the initial message */
    ucp_lane_index_t lane;

    lane  = ucp_proto_common_find_am_bcopy_hdr_lane(
                               &params->super.super);
     if (lane == UCP_NULL_LANE) {
         return UCS_ERR_NO_ELEM;
     }

     ucs_assert(params->super.send_op == UCT_EP_OP_AM_BCOPY);

     return ucp_proto_rndv_ctrl_init(params, lane);
```

src/ucp/tag/offload.c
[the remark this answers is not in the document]
yosefe: pls add this to the comment

src/ucp/tag/offload/rndv.c
yosefe: but when we call ucp_tag_rndv_offload_request_init, we already know the request is going to be sent thru tag api only. and that function should initialize the fields that do not change during protocol progress

src/ucp/tag/offload/rndv.c
yosefe: do we have uct capability for max_rndv_request?

src/ucp/tag/offload/rndv.c
brminich: i do not see a similar assert there, so it seems to be useful.
remove?

yosefe: actually this assert should fail:
https://github.com/openucx/ucx/blob/d4bdc86aa18abbacaea3fa2220ba6b8b1ed39896/src/ucp/dt/datatype_iter.inl#L474
from ucp_proto_zcopy_single_progress->ucp_datatype_iter_next_iov

brminich: ic, though it does not seem really obvious and can be changed unrelatively to hwtm. So, i'd still keep it.
remove?

src/ucp/tag/offload/rndv.c
brminich: why? I can try to invoke `ucp_rkey_packed_size` directly, but this would be the same as we do when calculating `config->rndv.rkey_size

src/ucp/tag/offload/rndv.c
brminich: no we have just `max_hdr` which is not the same as it represents free space ater RVH/RAVH in offloaded rndv

src/ucp/tag/offload/rndv.c
brminich: will add comment

src/ucp/tag/offload/rndv.c
yosefe: yes. In theory there could be HWTM transport that does not need uct memh.

src/ucp/tag/offload/rndv.c
brminich: i'd want to keep it here, because this flag shows that `req->send.tag_offload.rndv_op` is initialized and therefore needs to be freed/canceled in some cleanup flow. Setting this flag earlier (before the actual operation is peformed) is error prone, because it may lead to incorect flows during cleaup/error scenarios.
UCT API does not define invalid rndv operation id, so we have to maintain this flag to check whether UCT rndv resource needs to be freed.

src/ucp/tag/offload/rndv.c
yosefe: when we remove protov1 we would want to remove most of config contents.

src/ucp/tag/offload/rndv.c
brminich: ok

src/ucp/tag/offload/rndv.c
yosefe: maybe we should add it?

brminich: we can add that, but probably in a separate PR?

yosefe: yes

yosefe: for now pls add comment

src/ucp/tag/offload/rndv.c
brminich: updated `ucp_ep_tag_offload_min_rndv_thresh`

src/ucp/core/ucp_ep.c
```diff
@@ -2133,9 +2133,12 @@ static void ucp_ep_config_adjust_max_short(ssize_t *max_short,
  * user buffer when matched. Thus, minimum message size allowed to be sent with
  * RNDV protocol should be bigger than maximal possible SW RNDV request
  * (i.e. header plus packed keys size). */
-size_t ucp_ep_tag_offload_min_rndv_thresh(ucp_ep_config_t *config)
+size_t ucp_ep_tag_offload_min_rndv_thresh(ucp_context_h context,
+                                          ucp_ep_config_t *config)
```
yosefe, line 2137: pass const ucp_ep_config_key_t*

yosefe: @brminich ucp_ep_config_key_t*, not ucp_ep_config_t*

yosefe (approved):

yosefe (approved):

PR #8291, opened 2022-06-07
UCT/DC/EP: ep datatype refactoring

hoopoepg:
- as first step for DC flush remote preparation
- optimized DC EP datatype to allow to add more data without
  EP size increase
- added DC-specific AV datatype to optimize data usage


hoopoepg:
https://github.com/openucx/ucx/pull/8318

Issue #8278, opened 2022-06-03
Build failure


haampie:
Using clang 13, ucx 1.13.0-rc1 and CUDA 11.7

```
$ grep 'implicit conversion' build.txt
cuda_copy/cuda_copy_ep.c:120:31: error: implicit conversion from enumeration type 'CUresult' (aka 'enum cudaError_enum') to different enumeration type 'cudaError_t' (aka 'enum cudaError') [-Werror,-Wenum-conversion]
cuda_ipc/cuda_ipc_ep.c:78:31: error: implicit conversion from enumeration type 'CUresult' (aka 'enum cudaError_enum') to different enumeration type 'cudaError_t' (aka 'enum cudaError') [-Werror,-Wenum-conversion]
cuda_ipc/cuda_ipc_iface.c:445:27: error: implicit conversion from enumeration type 'CUresult' (aka 'enum cudaError_enum') to different enumeration type 'cudaError_t' (aka 'enum cudaError') [-Werror,-Wenum-conversion]
cuda_ipc/cuda_ipc_iface.c:532:27: error: implicit conversion from enumeration type 'CUresult' (aka 'enum cudaError_enum') to different enumeration type 'cudaError_t' (aka 'enum cudaError') [-Werror,-Wenum-conversion]
cuda_copy/cuda_copy_iface.c:313:27: error: implicit conversion from enumeration type 'CUresult' (aka 'enum cudaError_enum') to different enumeration type 'cudaError_t' (aka 'enum cudaError') [-Werror,-Wenum-conversion]
cuda_copy/cuda_copy_iface.c:456:27: error: implicit conversion from enumeration type 'CUresult' (aka 'enum cudaError_enum') to different enumeration type 'cudaError_t' (aka 'enum cudaError') [-Werror,-Wenum-conversion]
```

Is there a convenient way to disable `-Werror`?

yosefe:
@evgeny-leksikov @Akshay-Venkatesh FYI

zzzoom:
Seconded, please add an option to disable the currently hardcoded `-Werror` to allow builds with bleeding edge compilers and weird compiler options.

PR #8293, opened 2022-06-07
UCT/CUDA: use right driver macro to log errors with drv fxns

Akshay-Venkatesh:
## Why ?
Fix issues reported by clang and cuda 11.7 reported here https://github.com/openucx/ucx/issues/8278


Akshay-Venkatesh:
cc @yosefe

yosefe:
@Akshay-Venkatesh can you pls port to v1.13.x as well?

PR #8297, opened 2022-06-08
GTEST/IB: Get CQ events info from mlx5 structure if possible

ivankochin:
## What
Get info about CQ events from the corresponding mlx5 related structure (`uct_ib_mlx5_cq_t`)  during the test if possible.

## Why ?
It's a part of bigger #8257 PR that adds ability to create CQ using DEVX. If CQ created using DEVX, there is no `ibv_cq` entity anymore, so any CQ information should be discovered via corresponding UCX cq structure if possible.

### DIFF
--- a/test/gtest/uct/ib/test_ib.cc
+++ b/test/gtest/uct/ib/test_ib.cc
@@ -8,6 +8,7 @@
 #ifdef HAVE_MLX5_DV
 extern "C" {
 #include <uct/ib/mlx5/ib_mlx5.h>
+#include <uct/ib/rc/accel/rc_mlx5_common.h>
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


test/gtest/uct/ib/test_ib.cc
```diff
@@ -866,29 +867,38 @@ class test_uct_event_ib : public test_uct_ib {
...
-            uint32_t completed_evt = recv_cq->comp_events_completed;
-            /* need this call to acknowledge the completion to prevent iface dtor hung*/
-            ibv_ack_cq_events(ib_iface->cq[UCT_IB_DIR_RX], 1);
-            UCS_TEST_ABORT("recv_cq->comp_events_completed have to be 1 but the value "
-                           << completed_evt);
+
+        ibv_ack_cq_events(ib_iface->cq[dir], 1);
+    }
+
+    void check_cq(uct_iface_t *iface, size_t dir, size_t val)
```
dmitrygx, line 892:
```suggestion
    void check_cq(uct_iface_t *iface, uct_ib_dir_t dir, size_t val)
```

dmitrygx (approved):

dmitrygx (dismissed):

test/gtest/uct/ib/test_ib.cc
```diff
@@ -866,29 +867,37 @@ class test_uct_event_ib : public test_uct_ib {
...
-            uint32_t completed_evt = recv_cq->comp_events_completed;
-            /* need this call to acknowledge the completion to prevent iface dtor hung*/
-            ibv_ack_cq_events(ib_iface->cq[UCT_IB_DIR_RX], 1);
-            UCS_TEST_ABORT("recv_cq->comp_events_completed have to be 1 but the value "
-                           << completed_evt);
+
+        ibv_ack_cq_events(ib_iface->cq[dir], 1);
+    }
+
+    void check_cq(uct_iface_t *iface, uct_ib_dir_t dir, size_t val)
```
gleon99, line 891: I know it's not "your code", but I'd replace `val` with `value` (abbreviations are bad and confusing. Is "val" a "value" or "valid"?).

If we could I'd also change `dir` (direction? directory? direct?) but it has the `uct_ib_dir_t` type so we'll leave that for another time :)

test/gtest/uct/ib/test_ib.cc
```diff
@@ -866,29 +867,38 @@ class test_uct_event_ib : public test_uct_ib {
...
-        if (val != send_cq->comp_events_completed) {
-            uint32_t completed_evt = send_cq->comp_events_completed;
-            /* need this call to acknowledge the completion to prevent iface dtor hung*/
-            ibv_ack_cq_events(ib_iface->cq[UCT_IB_DIR_TX], 1);
-            UCS_TEST_ABORT("send_cq->comp_events_completed have to be 1 but the value "
-                           << completed_evt);
+    size_t get_comp_events(const uct_iface_t *iface, const uct_ib_dir_t dir)
+    {
+        uct_ib_iface_t *ib_iface;
+#if IBV_HW_TM
```
brminich, line 873: why `HW_TM`?

ivankochin: It is just the only macro that fits my need that I've found. Do you have any suggestions for what it should be replaced?

brminich: i'd suggest `HAVE_MLX5_DV`. `HW_TM` controls just tag offload capability

brminich (dismissed):

hoopoepg (approved):

gleon99:
@ivankochin please squash.

hoopoepg (approved):

PR #8298, opened 2022-06-08
UCP/WIREUP: Symmetric latency calculation in transport selection

shasson5:
## What
 1) Added fp8 pack/unpack to local latency value when calculating transport score (address V2).
 2) Added num_eps to latency calculation (address V1).

## Why ?
1) FP8 is a lossy compression method, so it can cause modification to remote latency values, when packing/unpacking remote address. 
As a result, score calculation is effected, and may cause asymmetric transports selection in a pair of connected endpoints.
The solution is adding pf8 pack/unpack to local address, which will make the calculation symmetric.
2) In address V1, remote address latency now contains num_eps. This was done to match the existing local address latency calculation.

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -566,17 +566,30 @@ out:
     return UCS_OK;
 }
 
-static inline double ucp_wireup_tl_iface_latency(ucp_context_h context,
-                                                 const uct_iface_attr_t *iface_attr,
-                                                 const ucp_address_iface_attr_t *remote_iface_attr)
+static double ucp_wireup_fp8_pack_unpack_latency(double latency)
 {
+    ucs_fp8_t packed_lat = UCS_FP8_PACK(LATENCY, latency);
+    return UCS_FP8_UNPACK(LATENCY, packed_lat) / UCS_NSEC_PER_SEC;
+}
+
+static inline double
+ucp_wireup_tl_iface_latency(ucp_context_h context,
+                            const uct_iface_attr_t *iface_attr,
+                            const ucp_address_iface_attr_t *remote_iface_attr)
+{
+    double local_lat, lat_nsec;
+
     if (remote_iface_attr->addr_version == UCP_OBJECT_VERSION_V1) {
         /* Address v1 contains just latency overhead */
         return ucs_max(iface_attr->latency.c, remote_iface_attr->lat_ovh) +
                (iface_attr->latency.m * context->config.est_num_eps);
     } else {
-        return ucs_max(remote_iface_attr->lat_ovh,
-                       ucp_tl_iface_latency(context, &iface_attr->latency));
+        /* FP8 is a lossy compression method, so in order to create a symmetric
+         * calculation we pack/unpack the local latency as well */
+        lat_nsec  = ucp_tl_iface_latency(context, &iface_attr->latency) *
+                    UCS_NSEC_PER_SEC;
+        local_lat = ucp_wireup_fp8_pack_unpack_latency(lat_nsec);
+        return ucs_max(remote_iface_attr->lat_ovh, local_lat);
     }
 }
 
--- a/test/gtest/ucp/test_ucp_wireup.cc
+++ b/test/gtest/ucp/test_ucp_wireup.cc
@@ -1228,6 +1228,50 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
 UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, ud, "ud")
 UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc_dc, "rc,dc")
 
+class select_ep_transport : public test_ucp_wireup {
+public:
+    static void get_test_variants(std::vector<ucp_test_variant>& variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_RMA, TEST_RMA, "rma");
+    }
+
+    const char *rma_transport(const entity *e)
+    {
+        const ucp_ep_config_t *config = ucp_ep_config(e->ep());
+        ucp_lane_index_t lane_index   = config->key.rma_lanes[0];
+        return ucp_ep_get_tl_rsc(e->ep(), lane_index)->tl_name;
+    }
+
+    void verify_symmetric_tl_selection(const std::string &num_eps1,
+                                       const std::string &num_eps2)
+    {
+        /* First context initialized with num_eps above threshold */
+        modify_config("NUM_EPS", num_eps1);
+        entity *e1 = create_entity();
+
+        /* Second context initialized with num_eps below threshold */
+        modify_config("NUM_EPS", num_eps2);
+        entity *e2 = create_entity();
+
+        e1->connect(e2, get_ep_params());
+        e2->connect(e1, get_ep_params());
+
+        /* Verify that selection is the same for both eps */
+        ASSERT_STREQ(rma_transport(e1), rma_transport(e2));
+    }
+};
+
+UCS_TEST_P(select_ep_transport, fp8_modify_result_v2)
+{
+    /* num_eps threshold for selecting DC over RC is 60 */
+    modify_config("ADDRESS_VERSION", "v2");
+    verify_symmetric_tl_selection("65", "55");
+}
+
+UCP_INSTANTIATE_TEST_CASE_TLS(select_ep_transport, ib, "ib")
+UCP_INSTANTIATE_TEST_CASE_TLS(select_ep_transport, ib_shm, "ib,shm")
+UCP_INSTANTIATE_TEST_CASE_TLS(select_ep_transport, all, "all")
+
 class test_ucp_wireup_fallback_amo : public test_ucp_wireup {
 protected:
     void init() {


src/ucp/wireup/select.c
```diff
@@ -570,13 +570,18 @@ static inline double ucp_wireup_tl_iface_latency(ucp_context_h context,
...
         /* Address v1 contains just latency overhead */
         return ucs_max(iface_attr->latency.c, remote_iface_attr->lat_ovh) +
                (iface_attr->latency.m * context->config.est_num_eps);
     } else {
-        return ucs_max(remote_iface_attr->lat_ovh,
-                       ucp_tl_iface_latency(context, &iface_attr->latency));
+        /* FP8 is a lossy compression method, so in order to create a symmetric calculation we pack/unpack the local latency as well */
+        packed_lat = UCS_FP8_PACK(LATENCY, ucp_tl_iface_latency(context, &iface_attr->latency) * UCS_NSEC_PER_SEC);
+        unpacked_lat = (UCS_FP8_UNPACK(LATENCY, packed_lat)) / UCS_NSEC_PER_SEC;
+        return ucs_max(remote_iface_attr->lat_ovh, unpacked_lat);
```
brminich, line 584: please wrap the lines to be not longer than 80 syms

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,43 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
+
+UCS_TEST_P(select_ep_transport, fp8_round)
+{
+    modify_config("ADDRESS_VERSION", "v2");
+
+    modify_config("NUM_EPS", "65");
+    entity *e1 = create_entity();
+
+    modify_config("NUM_EPS", "55");
+    entity *e2 = create_entity();
```
brminich, line 1255: 👍

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,43 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
+    e1->connect(e2, get_ep_params());
+    char *selected1 = get_selected_transport(e1);
+
+    e2->connect(e1, get_ep_params());
+    char *selected2 = get_selected_transport(e2);
+
+    ASSERT_STREQ(selected1, selected2);
+}
+
+UCP_INSTANTIATE_TEST_CASE_TLS(select_ep_transport, rc_dc, "rc,dc")
```
brminich, line 1266: maybe we can run it for all ib transports

src/ucp/wireup/select.c
```diff
@@ -570,13 +570,18 @@ static inline double ucp_wireup_tl_iface_latency(ucp_context_h context,
...
     if (remote_iface_attr->addr_version == UCP_OBJECT_VERSION_V1) {
         /* Address v1 contains just latency overhead */
         return ucs_max(iface_attr->latency.c, remote_iface_attr->lat_ovh) +
                (iface_attr->latency.m * context->config.est_num_eps);
     } else {
-        return ucs_max(remote_iface_attr->lat_ovh,
-                       ucp_tl_iface_latency(context, &iface_attr->latency));
+        /* FP8 is a lossy compression method, so in order to create a symmetric calculation we pack/unpack the local latency as well */
+        packed_lat = UCS_FP8_PACK(LATENCY, ucp_tl_iface_latency(context, &iface_attr->latency) * UCS_NSEC_PER_SEC);
+        unpacked_lat = (UCS_FP8_UNPACK(LATENCY, packed_lat)) / UCS_NSEC_PER_SEC;
```
dmitrygx, line 583: - align by `=`
- pls drop the line to fit 80 symbols per line requirement
see clang code formatter suggestions - https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=45681&view=logs&jobId=a3e7afe8-b282-5869-712a-9857fbd4b4a9&j=cc064a77-22b5-56bf-ecc0-70b5fe764261&t=aedfd754-44a6-53e6-6843-8659d800fee2

dmitrygx: what about creating a helper function to do fp8 pack/unpack inside?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,43 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
 UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, ud, "ud")
 UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc_dc, "rc,dc")
 
+class select_ep_transport : public test_ucp_wireup {
```
dmitrygx, line 1231: I think we could use already existing `test_ucp_wireup` test suite and add the new test there

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,43 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
+    static void get_test_variants(std::vector<ucp_test_variant>& variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_RMA, TEST_RMA, "rma");
+    }
+
+    char *get_selected_transport(entity *e)
+    {
+        ucp_worker *worker = e->worker();
+        ucp_lane_index_t lane_index = worker->ep_config[0].key.rma_lanes[0];
+        ucp_rsc_index_t rsc_index = worker->ep_config[0].key.lanes[lane_index].rsc_index;
```
dmitrygx, line 1242: align by `=`

gleon99:
@shasson5 please fix commit title, to run the pipeline.

src/ucp/wireup/address.c
```diff
@@ -654,7 +654,7 @@ ucp_address_pack_iface_attr_v1(ucp_worker_h worker, void *ptr,
     packed->overhead       = iface_attr->overhead;
     packed->bandwidth      = ucp_tl_iface_bandwidth(worker->context,
                                                     &iface_attr->bandwidth);
-    packed->lat_ovh        = iface_attr->latency.c;
+    packed->lat_ovh        = ucp_tl_iface_latency(worker->context, &iface_attr->latency);
```
gleon99, line 657: Line too long

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,65 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
+public:
+    static void get_test_variants(std::vector<ucp_test_variant>& variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_RMA, TEST_RMA, "rma");
+    }
+
+    char *get_selected_transport(entity *e)
+    {
+        ucp_worker *worker = e->worker();
+        ucp_lane_index_t lane_index = worker->ep_config[0].key.rma_lanes[0];
```
gleon99, line 1241: Use `const` where possible

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,65 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
+{
+    /* num_eps threshold for selecting DC instead of RC is 60 */
+    modify_config("ADDRESS_VERSION", "v2");
+
+    /* First ep initialized with num_eps above threshold */
+    modify_config("NUM_EPS", "65");
+    entity *e1 = create_entity();
+
+    /* Second ep initialized with num_eps below threshold */
+    modify_config("NUM_EPS", "55");
```
gleon99, line 1257: There is a lot of common code in these 2 tests, let's try to avoid duplication.

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,65 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
+        return e->ucph()->tl_rscs[rsc_index].tl_rsc.tl_name;
+    }
+};
+
+UCS_TEST_P(select_ep_transport, fp8_modify_result_v2)
+{
+    /* num_eps threshold for selecting DC instead of RC is 60 */
+    modify_config("ADDRESS_VERSION", "v2");
+
+    /* First ep initialized with num_eps above threshold */
```
gleon99, line 1252: First **ep**? Or maybe context/worker?

src/ucp/wireup/select.c
gleon99: Extra `()`?

test/gtest/ucp/test_ucp_wireup.cc
shasson5: We only changed common code (nothing for a specific transport).
so we only need to check the general case where you have 2 transports with different priorities for each endpoint.

src/ucp/wireup/address.c
```diff
@@ -654,7 +654,8 @@ ucp_address_pack_iface_attr_v1(ucp_worker_h worker, void *ptr,
     packed->overhead       = iface_attr->overhead;
     packed->bandwidth      = ucp_tl_iface_bandwidth(worker->context,
                                                     &iface_attr->bandwidth);
-    packed->lat_ovh        = iface_attr->latency.c;
+    packed->lat_ovh        = ucp_tl_iface_latency(worker->context,
+                                                  &iface_attr->latency);
```
brminich, line 658: looks like it is not wire compatible

src/ucp/wireup/select.c
```diff
@@ -566,18 +566,33 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
...
+        unpacked_lat = ucp_tl_iface_latency(context, &iface_attr->latency);
     } else {
-        return ucs_max(remote_iface_attr->lat_ovh,
-                       ucp_tl_iface_latency(context, &iface_attr->latency));
+        /* FP8 is a lossy compression method, so in order to create a symmetric
+         * calculation we pack/unpack the local latency as well */
+        unpacked_lat = ucp_wireup_fp8_pack_unpack_latency(context,
+                                                          &iface_attr->latency);
     }
+    return ucs_max(remote_iface_attr->lat_ovh, unpacked_lat);
```
brminich, line 595: `remote_iface_attr->lat_ovh` can contain just `iface_attr->latency.c` portion

src/ucp/wireup/address.c
shasson5: can you explain?

gleon99: @brminich right. But there is a bug here. Do you think we can do any better than just keeping it there?

yosefe: @brminich why do you think it's not wire-compatible? from remote side perspective it always treats this value from address as latency number

test/gtest/ucp/test_ucp_wireup.cc
yosefe: we can do like this - instantiate on multiple transports only:
```
UCP_INSTANTIATE_TEST_CASE_TLS(select_ep_transport, ib, "ib")
UCP_INSTANTIATE_TEST_CASE_TLS(select_ep_transport, ib_shm, "ib,shm")
UCP_INSTANTIATE_TEST_CASE_TLS(select_ep_transport, all, "all")
```

src/ucp/wireup/select.c
brminich: but this change increases probability for ep reconfig for address v1, while the PR fixes an issue with address v2 (off by default)

yosefe: i think it also should fix ep reconfiguration with address v1 (and not increase the error probability)
since in v1 we take max of remote and num_eps-adjusted-local, we better also adjust before packing the value we provide to remove peer.

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,75 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
+        char *selected1 = get_selected_transport(e1);
+
+        e2->connect(e1, get_ep_params());
+        char *selected2 = get_selected_transport(e2);
+
+        /* Verify that selection is the same for both eps */
+        ASSERT_STREQ(selected1, selected2);
+
+        /* Verify correct transport was selected */
+        ASSERT_STREQ(selected1, expected_tl.c_str());
```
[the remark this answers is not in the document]
shasson5: but do we know for sure that the code will crash (and not only print a message and continue)?
Also, for `select_rc_for_32_eps_v2` and `select_rc_for_32_eps_v1` tests it is indeed relevant, because we want to verify RC

src/ucp/wireup/select.c
```diff
@@ -582,17 +577,18 @@ ucp_wireup_tl_iface_latency(ucp_context_h context,
...
 
     if (remote_iface_attr->addr_version == UCP_OBJECT_VERSION_V1) {
-        unpacked_lat = ucp_tl_iface_latency(context, &iface_attr->latency);
+        local_lat = ucp_tl_iface_latency(context, &iface_attr->latency);
     } else {
         /* FP8 is a lossy compression method, so in order to create a symmetric
          * calculation we pack/unpack the local latency as well */
-        unpacked_lat = ucp_wireup_fp8_pack_unpack_latency(context,
-                                                          &iface_attr->latency);
+        double lat_nsec = ucp_tl_iface_latency(context, &iface_attr->latency) *
```
yosefe, line 587: minor: this is C code, better define it in the beginning of the function

test/gtest/ucp/test_ucp_wireup.cc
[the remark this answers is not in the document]
yosefe: if there is an error message from the code it will cause gtest to fail

src/ucp/wireup/select.c
brminich: @yosefe, afair we had similar problem with BW, which we had to fix here:
https://github.com/openucx/ucx/blob/master/src/ucp/wireup/address.c#L818

yosefe: the fix was passing BW after considering PPN: https://github.com/openucx/ucx/blob/master/src/ucp/wireup/address.c#L655
Therefore IMO we should do the same with latency packing

src/ucp/wireup/select.c
```diff
@@ -582,17 +577,18 @@ ucp_wireup_tl_iface_latency(ucp_context_h context,
...
         /* FP8 is a lossy compression method, so in order to create a symmetric
          * calculation we pack/unpack the local latency as well */
-        unpacked_lat = ucp_wireup_fp8_pack_unpack_latency(context,
-                                                          &iface_attr->latency);
+        lat_nsec  = ucp_tl_iface_latency(context, &iface_attr->latency) *
+                    UCS_NSEC_PER_SEC;
+        local_lat = ucp_wireup_fp8_pack_unpack_latency(lat_nsec);
     }
-    return ucs_max(remote_iface_attr->lat_ovh, unpacked_lat);
+    return ucs_max(remote_iface_attr->lat_ovh, local_lat);
```
yosefe, line 591: we should revert v1 logic to the way it was before this PR

src/ucp/wireup/address.c
yosefe: pls revert this to preserve old v1 logic

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,61 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
+    {
+        add_variant_with_value(variants, UCP_FEATURE_RMA, TEST_RMA, "rma");
+    }
+
+    char *get_selected_transport(const entity *e)
+    {
+        const ucp_worker *worker    = e->worker();
+        ucp_lane_index_t lane_index = worker->ep_config[0].key.rma_lanes[0];
+        ucp_rsc_index_t rsc_index   =
+                worker->ep_config[0].key.lanes[lane_index].rsc_index;
```
yosefe, line 1243: we should use endpoint's config, rather than worker->ep_config[0]
`ucp_ep_config(e->ep())`

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,61 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
 UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc_dc, "rc,dc")
 
+class select_ep_transport : public test_ucp_wireup {
+public:
+    static void get_test_variants(std::vector<ucp_test_variant>& variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_RMA, TEST_RMA, "rma");
+    }
+
+    char *get_selected_transport(const entity *e)
```
yosefe, line 1238: return const char*

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1228,6 +1228,61 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_wireup_unified, rc, "rc")
...
+        entity *e2 = create_entity();
+
+        e1->connect(e2, get_ep_params());
+        char *selected1 = get_selected_transport(e1);
+
+        e2->connect(e1, get_ep_params());
+        char *selected2 = get_selected_transport(e2);
+
+        /* Verify that selection is the same for both eps */
+        ASSERT_STREQ(selected1, selected2);
```
yosefe, line 1265: maybe just
```
ASSERT_STREQ(rma_transport(e1), rma_transport(e2));
```

yosefe (dismissed):

brminich (dismissed):

PR #8299, opened 2022-06-08
UCT/IB/DEVX: Add CQ creation related DEVX structures

ivankochin:
## Why ?
It is a prework before #8257 merging. Creation CQ using DEVX patch is to big to be merged in one PR, so it going to be merged in parts. This part is the first one.


src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
```diff
@@ -1617,4 +1618,77 @@ enum {
     UCT_IB_MLX5_OBJ_TYPE_RESERVED_QPN = 0x002C,
 };
 
+struct uct_ib_mlx5_cqc_bits {
+    uint8_t status[0x4];
+    uint8_t as_notify[0x1];
```
yosefe, line 1623: since this is a "special" header file, probably better to match fields indentation to the other structs in this file

ivankochin: Thanks, good point!

yosefe: seems not fixed for all structs

ivankochin: Oh, I think that I've finally got what you mean.

ivankochin: Oh, I think that I've finally got what you mean.

yosefe (approved):

PR #8300, opened 2022-06-08
UCT/IB/DEVX: Using UCX CQ structure during DEVX calls

ivankochin:
## What
Getting all required CQ related information during DEVX calls from the UCX structure (instead of IBV strucuture).

## Why ?
It's a part of bigger https://github.com/openucx/ucx/pull/8257 PR that adds ability to create CQ using DEVX. This patch is too big to be merged in one PR so it's one of the parts.

If CQ created using DEVX then there is no IBV CQ at all, so all information about CQs should be discovered from the common `uct_ib_mlx5_cq_t` entity.

## How? 

- Move `uct_ib_mlx5_get_cq` right after the CQ creation (inside `create_cq` function pointer) to make all CQ related info accessible earlier. It is required at least for XRQ creation via DEVX.
- Pass UCT CQ structures to all related APIs and get all required information from them.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5_devx.c
+++ b/src/uct/ib/dc/dc_mlx5_devx.c
@@ -17,20 +17,18 @@
 
 ucs_status_t uct_dc_mlx5_iface_devx_create_dct(uct_dc_mlx5_iface_t *iface)
 {
-    uct_ib_device_t *dev  = uct_ib_iface_device(&iface->super.super.super);
-    struct mlx5dv_pd dvpd = {};
-    struct mlx5dv_cq dvcq = {};
-    struct mlx5dv_obj dv  = {};
+    uct_ib_device_t *dev        = uct_ib_iface_device(&iface->super.super.super);
+    const uct_ib_mlx5_cq_t *cq  = &iface->super.cq[UCT_IB_DIR_RX];
+    struct mlx5dv_pd dvpd       = {};
+    struct mlx5dv_obj dv        = {};
     char in[UCT_IB_MLX5DV_ST_SZ_BYTES(create_dct_in)]   = {};
     char out[UCT_IB_MLX5DV_ST_SZ_BYTES(create_dct_out)] = {};
     int dvflags;
     void *dctc;
 
-    dvflags   = MLX5DV_OBJ_PD | MLX5DV_OBJ_CQ;
+    dvflags   = MLX5DV_OBJ_PD;
     dv.pd.in  = uct_ib_iface_md(&iface->super.super.super)->pd;
     dv.pd.out = &dvpd;
-    dv.cq.in  = iface->super.super.super.cq[UCT_IB_DIR_RX];
-    dv.cq.out = &dvcq;
     mlx5dv_init_obj(&dv, dvflags);
 
     UCT_IB_MLX5DV_SET(create_dct_in, in, opcode, UCT_IB_MLX5_CMD_OP_CREATE_DCT);
@@ -41,7 +39,7 @@ ucs_status_t uct_dc_mlx5_iface_devx_create_dct(uct_dc_mlx5_iface_t *iface)
     if (UCT_RC_MLX5_TM_ENABLED(&iface->super)) {
         UCT_IB_MLX5DV_SET(dctc, dctc, offload_type, UCT_IB_MLX5_QPC_OFFLOAD_TYPE_RNDV);
     }
-    UCT_IB_MLX5DV_SET(dctc, dctc, cqn, dvcq.cqn);
+    UCT_IB_MLX5DV_SET(dctc, dctc, cqn, cq->cq_num);
     UCT_IB_MLX5DV_SET64(dctc, dctc, dc_access_key, UCT_IB_KEY);
 
     UCT_IB_MLX5DV_SET(dctc, dctc, rre, true);
--- a/src/uct/ib/mlx5/ib_mlx5.c
+++ b/src/uct/ib/mlx5/ib_mlx5.c
@@ -111,15 +111,15 @@ uct_ib_mlx5_create_cq(uct_ib_iface_t *iface, uct_ib_dir_t dir,
                       const uct_ib_mlx5_iface_config_t *mlx5_config,
                       const uct_ib_iface_config_t *ib_config,
                       const uct_ib_iface_init_attr_t *init_attr,
-                      int preferred_cpu, size_t inl)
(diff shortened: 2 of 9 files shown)


src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -719,6 +720,8 @@ void uct_ib_mlx5_txwq_validate_always(uct_ib_mlx5_txwq_t *wq, uint16_t num_bb,
 #if HAVE_DEVX
 
 ucs_status_t uct_ib_mlx5_devx_create_qp(uct_ib_iface_t *iface,
+                                        uct_ib_mlx5_cq_t *send_cq,
+                                        uct_ib_mlx5_cq_t *recv_cq,
```
brminich, line 724: maybe pass it as array?

ivankochin: Seems like we cannot do it by a such way because `uct_rc_mlx5_devx_create_cmd_qp` passes RX CQ to both send and receive CQs before QP creation (check out [here](https://github.com/openucx/ucx/pull/8300/files#diff-f12c727872eb471ada737731e26c261d093dc0ea19feb77e1643e6bdc7267b4aR284))

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -119,7 +120,7 @@ uct_ib_mlx5_create_cq(uct_ib_iface_t *iface, uct_ib_dir_t dir,
     struct mlx5dv_cq_init_attr dv_attr = {};
     uct_ib_mlx5_md_t *md               = ucs_derived_of(iface->super.md,
                                                         uct_ib_mlx5_md_t);
-    ucs_status_t status;
+    ucs_status_t status = UCS_OK;
```
brminich, line 123: why needed? It is initialized in the code below

src/uct/ib/mlx5/ib_mlx5.h
brminich: ok

src/uct/ib/rc/accel/rc_mlx5_devx.c
```diff
@@ -241,7 +238,7 @@ uct_rc_mlx5_devx_init_rx_tm(uct_rc_mlx5_iface_common_t *iface,
     UCT_IB_MLX5DV_SET(xrqc, xrqc, tag_matching_topology_context.log_matching_list_sz,
                                   ucs_ilog2(iface->tm.num_tags) + 1);
     UCT_IB_MLX5DV_SET(xrqc, xrqc, dc,       dc);
-    UCT_IB_MLX5DV_SET(xrqc, xrqc, cqn,      dvcq.cqn);
+    UCT_IB_MLX5DV_SET(xrqc, xrqc, cqn,      iface->cq[UCT_IB_DIR_RX].cq_num);
```
brminich, line 241: Minor: imo, this alignment looks weird (laso it is broken by the line 238). maybe have just one space here and above?

brminich (approved):

src/uct/ib/rc/accel/rc_mlx5_devx.c
ivankochin: It is aligned for the 4th argument and it's broken on 238 just because this row has too long 3rd argument. Anyway if you do not object I don't want to change the existing alignment at least as part of this PR.

brminich: ok

brminich (dismissed):

src/uct/ib/dc/dc_mlx5_devx.c
```diff
@@ -18,19 +18,17 @@
 ucs_status_t uct_dc_mlx5_iface_devx_create_dct(uct_dc_mlx5_iface_t *iface)
 {
     uct_ib_device_t *dev  = uct_ib_iface_device(&iface->super.super.super);
+    uct_ib_mlx5_cq_t *cq  = &iface->super.cq[UCT_IB_DIR_RX];
```
gleon99, line 21: `const`?

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -111,15 +111,16 @@ uct_ib_mlx5_create_cq(uct_ib_iface_t *iface, uct_ib_dir_t dir,
                       const uct_ib_mlx5_iface_config_t *mlx5_config,
                       const uct_ib_iface_config_t *ib_config,
                       const uct_ib_iface_init_attr_t *init_attr,
+                      uct_ib_mlx5_cq_t* mlx5_cq,
```
gleon99, line 114: `uct_ib_mlx5_cq_t *mlx5_cq`

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -111,15 +111,16 @@ uct_ib_mlx5_create_cq(uct_ib_iface_t *iface, uct_ib_dir_t dir,
...
                       int preferred_cpu, size_t inl)
 {
+    ucs_status_t status                = UCS_OK;
```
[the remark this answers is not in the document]
ivankochin: Because now status is used in both #if/else branches and even outside of this context.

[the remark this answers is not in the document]
gleon99: Yes, I understand why you moved it.
The question is, why do we initialize it? It is overridden in both (preprocessor) branches.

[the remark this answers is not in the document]
ivankochin: If we not initialize it, then we need to put it under all other initialized variables. And it would look like:
```c++
#if HAVE_DECL_MLX5DV_CQ_INIT_ATTR_MASK_CQE_SIZE
    uct_ib_device_t *dev               = uct_ib_iface_device(iface);
    struct ibv_cq_init_attr_ex cq_attr = {};
    struct mlx5dv_cq_init_attr dv_attr = {};
    uct_ib_mlx5_md_t *md               = ucs_derived_of(iface->super.md,
                                                        uct_ib_mlx5_md_t);
    ucs_status_t status;
    struct ibv_cq *cq;
    char message[128];
    int cq_errno;
#endif
    ucs_status_t status;
#if HAVE_DECL_MLX5DV_CQ_INIT_ATTR_MASK_CQE_SIZE
...
```

So I preferred just to initialize it to avoid this additional `#if`. Don't you think it is a lesser evil?

gleon99 (approved):

Artemy-Mellanox (approved):

PR #8307, opened 2022-06-10
OFI uct transport

MattBBaker:
## What
Adding support to run UCX on top of a libfabrics transport. This implementation is incomplete but I wanted to do a WIP PR. I've implemented many of the low level details that the rest of the implementation should be down to implementing the message injection routines. I'm hoping that someone with more libfabric knowledge than me can help with this part.

Known deficiencies right now includes the inability to select with fabric to use specifically. It defaults to the first available fabric, which seems to be the most common/correct choice, but should be configurable.

Second, I'm not happy with how the EP to libfabric address vector mapping works. It 'works for now' but only at small sizes and definitely needs more work.

## Why ?
Adds functionality to allow UCX to run on such systems as Frontier.

## How ?
Implements the usual UCT APIs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,6 @@ test/apps/profiling/ucx_profiling
 test/apps/uct_info/uct_info
 test/apps/uct_info/uct_info_static
 cmake/*.cmake
+*~
+\#*\#
+.\#*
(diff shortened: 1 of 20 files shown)


gleon99:
Hi @MattBBaker - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #8309, opened 2022-06-13
UCP/PROTO: Show protocol selection flags as part of info

yosefe:
## Why
Show select flags (in addition to op_attr_mask) in protocols information

### DIFF
--- a/src/ucp/proto/proto_debug.c
+++ b/src/ucp/proto/proto_debug.c
@@ -317,17 +317,33 @@ void ucp_proto_select_param_str(const ucp_proto_select_param_t *select_param,
         [ucs_ilog2(UCP_OP_ATTR_FLAG_FAST_CMPL)]  = "fast-completion",
         [ucs_ilog2(UCP_OP_ATTR_FLAG_MULTI_SEND)] = "multi",
     };
+    static const uint64_t op_flag_bits = UCP_PROTO_SELECT_OP_FLAG_PPLN_FRAG |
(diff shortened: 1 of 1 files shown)


src/ucp/proto/proto_debug.c
```diff
@@ -317,17 +317,33 @@ void ucp_proto_select_param_str(const ucp_proto_select_param_t *select_param,
         [ucs_ilog2(UCP_OP_ATTR_FLAG_FAST_CMPL)]  = "fast-completion",
         [ucs_ilog2(UCP_OP_ATTR_FLAG_MULTI_SEND)] = "multi",
     };
+    static const uint64_t op_flag_bits = UCP_PROTO_SELECT_OP_FLAG_PPLN_FRAG |
+                                         UCP_PROTO_SELECT_OP_FLAG_AM_EAGER |
+                                         UCP_PROTO_SELECT_OP_FLAG_AM_RNDV;
+    static const char *op_flag_names[] = {
+        [ucs_ilog2(UCP_PROTO_SELECT_OP_FLAG_PPLN_FRAG)] = "frag",
+        [ucs_ilog2(UCP_PROTO_SELECT_OP_FLAG_AM_EAGER)]  = "egr",
```
shamisp, line 325: `eager` is compact enough

brminich (approved):

src/ucp/proto/proto_debug.c
yosefe: all info is on same line so should be as short as possible

PR #8310, opened 2022-06-13
UCP/PROTOV2: Enable rndv for non-contig dt

brminich:
## What
Enable rendezvous protocol for non-contig datatypes

## Why ?
There are rendezvous protocols which support non-contig dts (e.g. am bcopy)


src/ucp/proto/proto_common.inl
```diff
@@ -80,7 +80,7 @@ ucp_proto_request_zcopy_cleanup(ucp_request_t *req, unsigned dt_mask)
 static UCS_F_ALWAYS_INLINE void
 ucp_proto_request_zcopy_complete(ucp_request_t *req, ucs_status_t status)
 {
-    ucp_proto_request_zcopy_cleanup(req, UCP_DT_MASK_CONTIG_IOV);
+    ucp_proto_request_zcopy_cleanup(req, UCP_DT_MASK_ALL);
```
yosefe, line 83: seems weird. zcopy should not be used for generic type.

brminich: I can change it to be a parameter. We have to handle a case when ATS is received for generic dt send, because of truncation (this func is called from `ucp_proto_rndv_ats_handler`).
BTW, [here](https://github.com/openucx/ucx/blob/master/src/ucp/proto/proto_common.c#L786) "all" mask is already used in zcopy completion routine

test/gtest/ucp/test_ucp_tag_xfer.cc
```diff
@@ -595,22 +595,22 @@ UCS_TEST_P(test_ucp_tag_xfer, iov_unexp) {
     test_xfer(&test_ucp_tag_xfer::test_xfer_iov, false, false, false);
 }
 
-UCS_TEST_P(test_ucp_tag_xfer, generic_err_exp) {
+UCS_TEST_P(test_ucp_tag_xfer, generic_err_exp, "PROTO_INDIRECT_ID=y") {
```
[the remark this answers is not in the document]
brminich: old protocols receive data in a different way. Once the error is detected by receiver it saves the status in the request and drops all consecutive chunks (just returns ok). Then the error is returned for the last chunk only.
https://github.com/openucx/ucx/blob/master/src/ucp/core/ucp_request.inl#L891

src/ucp/proto/proto_common.inl
yosefe: the other place seems like error as well. or, at least, wrong name for this function.
can we not call ucp_proto_request_zcopy_complete from ucp_proto_rndv_ats_handler? since now that we changed RTS selection logic, ATS can complete also a non-copy request.

test/gtest/ucp/test_ucp_tag_xfer.cc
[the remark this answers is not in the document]
yosefe: why need indirect id for new proto then?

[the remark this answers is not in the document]
brminich: different handler is used for proto v2:
https://github.com/openucx/ucx/blob/master/src/ucp/rndv/rndv_rtr.c#L435

src/ucp/rndv/proto_rndv.c
```diff
@@ -585,15 +591,21 @@ void ucp_proto_rndv_receive_start(ucp_worker_h worker, ucp_request_t *recv_req,
...
+    }
+
+    if (ucs_likely(size != 0)) {
+        UCS_PROFILE_CALL_VOID(ucp_datatype_iter_init_from_dt_state,
+                              worker->context, recv_req->recv.buffer, rts->size,
+                              recv_req->recv.datatype, &recv_req->recv.state,
+                              &req->send.state.dt_iter, &sg_count);
+    } else {
         ucp_request_recv_generic_dt_finish(recv_req);
         ucp_datatype_iter_init_null(&req->send.state.dt_iter, rts->size,
```
brminich, line 610: use `size` here

src/ucp/rndv/proto_rndv.inl
```diff
@@ -45,13 +45,16 @@ ucp_proto_rndv_rts_request_init(ucp_request_t *req)
...
-                                       UCP_DT_MASK_ALL);
-    if (status != UCS_OK) {
-        return status;
+    if (rpriv->md_map != 0) {
+        status = ucp_datatype_iter_mem_reg(ep->worker->context,
+                                           &req->send.state.dt_iter,
+                                           rpriv->md_map,
+                                           UCT_MD_MEM_ACCESS_RMA |
+                                                   UCT_MD_MEM_FLAG_HIDE_ERRORS,
+                                           UCS_BIT(UCP_DATATYPE_CONTIG));
```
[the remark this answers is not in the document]
brminich: currently we register memory only for contig datatypes when sending RTS (other dts do not support get_zcopy rndv protocol)

src/ucp/rndv/proto_rndv.c
```diff
@@ -697,6 +715,10 @@ ucp_proto_rndv_handle_rtr(void *arg, void *data, size_t length, unsigned flags)
             goto err_request_fail;
         }
     } else {
+        ucs_assertv(&req->send.state.dt_iter.dt_class == UCP_DATATYPE_CONTIG,
```
rakhmets, line 718:
```suggestion
        ucs_assertv(req->send.state.dt_iter.dt_class == UCP_DATATYPE_CONTIG,
```

src/ucp/rndv/proto_rndv.c
```diff
@@ -697,6 +715,10 @@ ucp_proto_rndv_handle_rtr(void *arg, void *data, size_t length, unsigned flags)
...
+                    "fragmented rndv is not supported with datatype %s",
```
rakhmets, line 719: `rendezvous` is more common in messages for users
```suggestion
                    "fragmented rendezvous is not supported with datatype %s",
```

src/ucp/rndv/proto_rndv.inl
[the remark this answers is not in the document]
rakhmets: Ok, I see.
contiguous datatype check is removed from RTS protocol initialization in this PR.
However, there is contiguous datatype check in remote protocol initialization.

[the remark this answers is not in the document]
brminich: No, RNDV can be used for non-contig data types, but we do not need to do registration for them

rakhmets (dismissed):

src/ucp/rndv/proto_rndv.c
```diff
@@ -585,18 +591,23 @@ void ucp_proto_rndv_receive_start(ucp_worker_h worker, ucp_request_t *recv_req,
...
+        size             = rts->size;
     } else {
         /* Short receive: complete with error, and send reply to sender */
         rkey_length      = 0; /* Override rkey length to disable data fetch */
         op_id            = UCP_OP_ID_RNDV_RECV_DROP;
         recv_req->status = UCS_ERR_MESSAGE_TRUNCATED;
+        size             = 0;
+    }
+
+    if (ucs_likely(size != 0)) {
```
yosefe, line 603: what is it fixing? some leak with rndv/ats? if yes, it's probably not the right way to fix since this code cannot make any assumptions on the protocol that will be selected

src/ucp/rndv/proto_rndv.c
```diff
@@ -356,10 +356,15 @@ void ucp_proto_rndv_rts_query(const ucp_proto_query_params_t *params,
 
 void ucp_proto_rndv_rts_abort(ucp_request_t *req, ucs_status_t status)
 {
+    const ucp_proto_rndv_ctrl_priv_t *rpriv = req->send.proto_config->priv;
+
     if (req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED) {
         ucp_send_request_id_release(req);
-        ucp_datatype_iter_mem_dereg(req->send.ep->worker->context,
-                                    &req->send.state.dt_iter, UCP_DT_MASK_ALL);
+        if (rpriv->md_map != 0) {
```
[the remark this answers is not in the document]
brminich: if we have unregistered IOV dt `ucp_datatype_iter_iov_mem_dereg` will unconditionally try to deregister it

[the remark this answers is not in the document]
brminich: maybe need to add check for `NULL` to `ucp_datatype_iter_iov_mem_dereg` instead

src/ucp/rndv/proto_rndv.c
yosefe: what is the problem with initializing iterator when RTS has no data? it should still work, just like user can send 0-length buffer in any operation.

brminich: the particular problem was that we do not call `dt_generic.ops.finish` if zero sized RTS arrives. I think it is more reasonable to  avoid initialization in such case, rather than doing init/finalize

src/ucp/dt/datatype_iter.c
```diff
@@ -72,6 +72,10 @@ void ucp_datatype_iter_iov_mem_dereg(ucp_context_h context,
     ucp_mem_h *memh = dt_iter->type.iov.memh;
     size_t iov_index, length;
 
+    if (memh == NULL) {
+        return;
+    }
+
     ucp_datatype_iter_iov_for_each(iov_index, length, dt_iter) {
         ucp_memh_put(context, memh[iov_index], 0);
```
rakhmets, line 80: Shouldn't  it be smth like this:

```C
void ucp_memh_dereg(ucp_context_h context, ucp_mem_h memh)
{
    if (memh != NULL) {
        ucp_memh_put(context, dt_iter->type.contig.memh, 0);
        memh = NULL;
    }
}

void ucp_datatype_iter_iov_mem_dereg(ucp_context_h context,
                                     ucp_datatype_iter_t *dt_iter)
{
    ucp_mem_h *memh = dt_iter->type.iov.memh;
    size_t iov_index, length;

    if (memh == NULL) {
        return;
    }

    ucp_datatype_iter_iov_for_each(iov_index, length, dt_iter) {
        ucp_memh_dereg(context, memh[iov_index]);
    }

    ucs_free(memh);
    dt_iter->type.iov.memh = NULL;
}

ucp_datatype_iter_mem_dereg(ucp_context_h context, ucp_datatype_iter_t *dt_iter,
                            unsigned dt_mask)
{
    if (ucp_datatype_iter_is_class(dt_iter, UCP_DATATYPE_CONTIG, dt_mask)) {
        ucp_memh_dereg(context, dt_iter->type.contig.memh);
    } else if (ucp_datatype_iter_is_class(dt_iter, UCP_DATATYPE_IOV, dt_mask)) {
        ucp_datatype_iter_iov_mem_dereg(context, dt_iter);
    }
}
```

src/ucp/dt/datatype_iter.c
```diff
@@ -72,6 +72,10 @@ void ucp_datatype_iter_iov_mem_dereg(ucp_context_h context,
     ucp_mem_h *memh = dt_iter->type.iov.memh;
     size_t iov_index, length;
 
+    if (memh == NULL) {
```
yosefe, line 75: how could the meme **array** be NULL?

src/ucp/rndv/proto_rndv.c
yosefe: rndv/ats protocol should call finish, just like other protocols for RNDV_RECV opcode call finish.

src/ucp/dt/datatype_iter.c
brminich: when it is not registered, plz check line 47 in this file

src/ucp/dt/datatype_iter.c
brminich: unification is nice, but do we really need to have a branch for checking every iov memh if dt is registered?

src/ucp/rndv/proto_rndv.c
brminich: ok, then the same needs to be applied to RECV_DROP opcode and dt cleanup should be removed from this func completely

src/ucp/dt/datatype_iter.c
brminich: no, this cleanup code is broken anyway, `dt_iter->memh` is set only on [line 65](https://github.com/openucx/ucx/blob/29fa51260b0012872582f8b6b3563994e324a082/src/ucp/dt/datatype_iter.c#L65)

brminich: i'll address this issue in this PR  as well

rakhmets (approved):

src/ucp/dt/datatype_iter.c
yosefe: ok

src/ucp/dt/datatype_iter.inl
```diff
@@ -488,6 +488,15 @@ ucp_datatype_iter_is_end(const ucp_datatype_iter_t *dt_iter)
     return ucp_datatype_iter_is_end_position(dt_iter, dt_iter);
 }
 
+static UCS_F_ALWAYS_INLINE void
+ucp_datatype_memh_dereg(ucp_context_h context, ucp_mem_h *memh_p)
```
yosefe, line 492: 👍

src/ucp/rndv/proto_rndv.c
```diff
@@ -761,3 +765,72 @@ void ucp_proto_rndv_bulk_request_init_lane_idx(
 
     req->send.multi_lane_idx = lane_idx - 1;
 }
+
+static ucs_status_t
+ucp_proto_rndv_ats_init(const ucp_proto_init_params_t *params)
```
yosefe, line 770: better move it to a new file - rndv_ats.c
proto_rndv should be only infra

src/ucp/rndv/proto_rndv.inl
```diff
@@ -45,13 +45,16 @@ ucp_proto_rndv_rts_request_init(ucp_request_t *req)
...
     }
 
-    status = ucp_datatype_iter_mem_reg(ep->worker->context,
-                                       &req->send.state.dt_iter, rpriv->md_map,
-                                       UCT_MD_MEM_ACCESS_RMA |
-                                               UCT_MD_MEM_FLAG_HIDE_ERRORS,
-                                       UCP_DT_MASK_ALL);
-    if (status != UCS_OK) {
-        return status;
+    if (rpriv->md_map != 0) {
```
yosefe, line 48: is this if really needed?

src/ucp/rndv/proto_rndv.inl
```diff
@@ -80,8 +83,12 @@ ucp_proto_rndv_ats_handler(void *arg, void *data, size_t length, unsigned flags)
         }
     }
 
+    ucp_datatype_iter_mem_dereg(req->send.ep->worker->context,
+                                &req->send.state.dt_iter, UCP_DT_MASK_ALL);
+    ucp_datatype_iter_cleanup(&req->send.state.dt_iter, UCP_DT_MASK_ALL);
```
yosefe, line 88: maybe use  ucp_proto_request_zcopy_cleanup() ?
or add dt_mask parameter to ucp_proto_request_zcopy_complete?

src/ucp/rndv/proto_rndv.c
```diff
@@ -761,3 +765,72 @@ void ucp_proto_rndv_bulk_request_init_lane_idx(
...
+
+    return ucp_proto_rndv_recv_complete(req);
+}
+
+static ucs_status_t
+ucp_proto_rndv_ats_proto_progress(uct_pending_req_t *uct_req)
+{
+    ucp_request_t *req = ucs_container_of(uct_req, ucp_request_t, send.uct);
+
+    return ucp_proto_rndv_ack_progress(req, req->send.proto_config->priv,
```
yosefe, line 822: if we want to make it more similar to other rndv protocol such as rndv_get, we could do something like
so ucp_proto_rndv_ats_complete will destroy rkey and dt->finish, and move to STAGE_ATS
```
.progress = {
    [STAGE_INIT] = ucp_proto_rndv_ats_complete,
    [STAGE_ATS] =  ucp_proto_rndv_ats_proto_progress,
}
```

src/ucp/rndv/proto_rndv.inl
brminich: yes, otherwise we will register mem for IOVs, even if it is not really needed.

yosefe: maybe it should be in ucp_datatype_iter_mem_reg?

src/ucp/rndv/proto_rndv.c
brminich: i made it as a part of request initialization. Imo no need for a separate stage which does not send (or wait) anything

brminich:
@yosefe, your comments fixed

src/ucp/dt/datatype_iter.inl
```diff
@@ -393,7 +393,7 @@ ucp_datatype_iter_next_slice(const ucp_datatype_iter_t *dt_iter,
 static UCS_F_ALWAYS_INLINE uct_mem_h
 ucp_datatype_iter_uct_memh(const ucp_mem_h memh, ucp_rsc_index_t memh_index)
 {
-    if (memh_index == UCP_NULL_RESOURCE) {
+    if ((memh_index == UCP_NULL_RESOURCE) || (memh == NULL)) {
```
yosefe, line 396: can memh be NULL only for IOV type?
if yes, better move memh==NULL check to caller function, to not use it in contig type

src/ucp/dt/datatype_iter.inl
```diff
@@ -505,6 +505,10 @@ ucp_datatype_iter_mem_reg(ucp_context_h context, ucp_datatype_iter_t *dt_iter,
                           ucp_md_map_t md_map, unsigned uct_flags,
                           unsigned dt_mask)
 {
+    if (md_map == 0) {
+        return UCS_OK;
+    }
```
yosefe, line 510: let's say we remove this branch.
would it cause something to fail, or just "waste" more time to create memh with md_map==0?

brminich: the latter

yosefe: it's not clear not me why we would register non-contig type. dt_iter already checks the datatype class before registering

brminich: here - https://github.com/openucx/ucx/blob/master/src/ucp/rndv/proto_rndv.inl#L52

yosefe: wouldn't it go here?
https://github.com/openucx/ucx/blob/master/src/ucp/dt/datatype_iter.inl#L508

brminich: yes for generic, but no for IOV

src/ucp/rndv/rndv_rtr.c
```diff
@@ -180,14 +180,16 @@ static ucs_status_t ucp_proto_rndv_rtr_progress(uct_pending_req_t *self)
...
     if (!(req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED)) {
-        status = ucp_datatype_iter_mem_reg(req->send.ep->worker->context,
-                                           &req->send.state.dt_iter,
-                                           rpriv->super.md_map,
-                                           UCT_MD_MEM_ACCESS_REMOTE_PUT,
-                                           UCS_BIT(UCP_DATATYPE_CONTIG));
-        if (status != UCS_OK) {
-            ucp_proto_request_abort(req, status);
-            return UCS_OK;
+        if (rpriv->super.md_map != 0) {
```
yosefe, line 183: why is it not covered by ucp_datatype_iter_mem_reg already?

src/ucp/rndv/proto_rndv.inl
```diff
@@ -80,8 +81,10 @@ ucp_proto_rndv_ats_handler(void *arg, void *data, size_t length, unsigned flags)
         }
     }
 
+    ucp_proto_request_zcopy_cleanup(req, UCP_DT_MASK_ALL);
```
yosefe, line 84: to maintain prev order, shall we move this after ucp_send_request_id_release ?

src/ucp/rndv/rndv_rtr.c
brminich: because I removed a branch from there according to the previous comment.
The failing flow is ud + contig datatype. If we get an empty memh here ,we will need to put it back later, which is not done in case of `receive_bcopy` (when we get messages by the AMs)

src/ucp/rndv/rndv_rtr.c
```diff
@@ -133,8 +136,6 @@ ucp_proto_rndv_rtr_data_received_zcopy(ucp_request_t *req, int in_buffer)
 {
     static const unsigned dt_mask = UCS_BIT(UCP_DATATYPE_CONTIG);
 
-    ucp_datatype_iter_mem_dereg(req->send.ep->worker->context,
```
yosefe, line 136: i think we should remove ucp_proto_rndv_rtr_data_received_bcopy/ucp_proto_rndv_rtr_data_received_zcopy now, and use ucp_proto_rndv_rtr_common_complete directly, with UCP_DT_MASK_ALL

src/ucp/rndv/proto_rndv.inl
yosefe: up

src/ucp/rndv/rndv_rtr.c
brminich: I'd leave it, this is extra protection that recv_zcopy is only relevant for contig dts

yosefe: why is it important to protect this? it's not like there would be a leak or some other error.. also we plan to extend zcopy to iov dt as well

brminich: to have an assert on an invalid flow.
ok, will unite

yosefe (approved):

yosefe (approved):

PR #8323, opened 2022-06-17
GTEST/UCP: Separate test for AM eager flag.

rakhmets:
## What
Added separate test checking UCP_AM_SEND_FLAG_EAGER to avoid multiple checks of the same functionality in different test class.
Removed overlapping tests from test_ucp_am_nbx_eager_data_release which were added to check the intermediate steps of enabling new protocol infrastructure for UCP AM.

## Why ?
The changes reduces the number of redundant tests and the testing time as well.

### DIFF
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -820,6 +820,27 @@ UCS_TEST_P(test_ucp_am_nbx, rx_am_mpools,
 UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx)
 
 
+class test_ucp_am_nbx_send_flag : public test_ucp_am_nbx {
+public:
+    virtual ucs_status_t
+    am_data_handler(const void *header, size_t header_length, void *data,
+                    size_t length, const ucp_am_recv_param_t *rx_param)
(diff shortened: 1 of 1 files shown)


test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1244,11 +1255,6 @@ UCS_TEST_P(test_ucp_am_nbx_dts, long_zcopy_send, "ZCOPY_THRESH=1",
     test_am(64 * UCS_KBYTE);
 }
 
-UCS_TEST_P(test_ucp_am_nbx_dts, send_eager_flag, "RNDV_THRESH=128")
```
brminich, line 1247: maybe reduce number of tests here even more? (as they are run for all combinations of dts and flags). Say we can test just `bcopy` and `zcopy` sends. Or maybe do just one test which sends messages of several  random sizes

yosefe: at least, we can unite 4K and 64K in single test instance
anyway I'd like to merge this PR since it helps other PRs to pass test faster, and we can optimize it more in next PR

yosefe (approved):

PR #8327, opened 2022-06-20
NEWS: 1.13.0 update for RC2

evgeny-leksikov:
## What
Update NEWS file for RC2 

## Why ?
Full list of changes since RC1:
 - ff64ffad7 AZP: remove HWI hosts from testing
 - 7e2df9171 UCT/CUDA: use right driver macro to log errors with drv fxns
 - 28b0f30b5 UCT/IB/DC: Always schedule DCI allocation during FC_HARD_REQ progress
 - f92b1727c TEST/UCT: Remove atomic access from invalidate tests
 - 5cec39013 UCT/IB: Fix indirect key registration by using correct MR type
 - a6c5a472f UCT/IB/DC: Fix managing FC_HARD_REQ resend time

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,7 +11,16 @@
 ### Features:
 ### Bugfixes:
 
-## 1.13.0 (May 27, 2022)
+## 1.13.0-rc2 (June 27, 2022)
+#### Bugfixes
+##### RDMA CORE (IB, ROCE, etc.)
+* Fixed indirect key registration
+* Fixed flow control protocol for DC transport
+##### GPU (CUDA, ROCM)
+* Fixed CUDA module compilation with clang 13
+* Fixes in ROCm memory detection and performance estimation
+
+## 1.13.0-rc1 (May 27, 2022)
 #### Features
 ##### Core
 * Added new objects to VFS: local and remote address of endpoint, statistics of ucp_ep_create success/failure, failed/destroyed endpoints


yosefe:
1. Need to update date to June 26 :)
2. pls add "Fixes in ROCm memory detection and performance estimation" @edgargabriel

evgeny-leksikov:
> 1. Need to update date to June 26 :)
> 2. pls add "Fixes in ROCm memory detection and performance estimation" @edgargabriel

done

NEWS
```diff
@@ -11,7 +11,7 @@
 ### Features:
 ### Bugfixes:
 
-## 1.13.0 (May 27, 2022)
+## 1.13.0 (June 26, 2022)
```
yosefe, line 14: it should be 1.13.0-rc2

evgeny-leksikov: > 1.13.0-rc2

should we then also update extra version in configure.ac?

yosefe: yes, let's try this, it would be first time we use it in a release.

configure.ac
```diff
@@ -8,10 +8,10 @@
...
 AC_PREREQ([2.63])
 
-define([ucx_ver_major], 1)  # Major version. Usually does not change.
-define([ucx_ver_minor], 13) # Minor version. Increased for each release.
-define([ucx_ver_patch], 0)  # Patch version. Increased for a bugfix release.
-define([ucx_ver_extra], )   # Extra version string. Empty for a general release.
+define([ucx_ver_major], 1)    # Major version. Usually does not change.
+define([ucx_ver_minor], 13)   # Minor version. Increased for each release.
+define([ucx_ver_patch], 0)    # Patch version. Increased for a bugfix release.
+define([ucx_ver_extra], .rc2) # Extra version string. Empty for a general release.
```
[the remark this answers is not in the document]
evgeny-leksikov: It does not work for RPM build on some systems:

The version Tag
The version tag defines the version of the software being packaged. The version specified should be as close as possible to the format of the original software's version. In most cases, there should be no problem specifying the version just as the software's original developer did. However, there is a restriction. There can be no dashes in the version. If you forget, RPM will remind you:
```
# rpmbuild -ba cdplayer-1.0.spec

* Package: cdplayer
Illegal '-' char in version: 1.0-a
```
http://rpm5.org/docs/max-rpm.html

this is why we decided to use `.` for int4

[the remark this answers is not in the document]
yosefe: ok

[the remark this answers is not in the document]
yosefe: @shamisp FYI

yosefe (approved):

edgargabriel (approved):

configure.ac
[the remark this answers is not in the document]
shamisp: What about deb and other package managers ?

NEWS
yosefe: we don't have it in the NEWS now, because it was replaced it to v1.12.0 (for example) before the GA

yosefe: this is how it was done so far, so the diff reflects the changes vs. the previous GA.
@shamisp WDYT?

shamisp: We do this to track in case somebody picks up one of the RC instead of releases - this way it is actually reflected in news.

NEWS
```diff
@@ -11,6 +11,15 @@
 ### Features:
 ### Bugfixes:
 
+## 1.13.0-rc2 (July 24)
```
yosefe, line 14: since it's 1.13 release need to update the 1.13.0 entry and set date to today

NEWS
evgeny-leksikov: OK, added section on Wiki page https://github.com/openucx/ucx/wiki/Release-process#release-candidates

NEWS
evgeny-leksikov: then maybe update 1.13.0 entry to 1.13.0-rc1 and add new one 1.13.0-rc2 (today)?

yosefe (approved):

PR #8329, opened 2022-06-20
GTEST/UCP: Reduced number of tests in AM datatype test class.

rakhmets:
## What
Combined `test_ucp_am_nbx_dts` tests with overlapping functionality.

## Why ?
The changes reduces the number of redundant tests and the testing time as well.
Addressed https://github.com/openucx/ucx/pull/8323#discussion_r901125577 .

### DIFF
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -1231,27 +1231,15 @@ UCS_TEST_P(test_ucp_am_nbx_dts, short_send, "ZCOPY_THRESH=-1", "RNDV_THRESH=-1")
     test_am(1);
 }
 
-UCS_TEST_P(test_ucp_am_nbx_dts, short_bcopy_send, "ZCOPY_THRESH=-1",
-                                                  "RNDV_THRESH=-1")
+UCS_TEST_P(test_ucp_am_nbx_dts, bcopy_send, "ZCOPY_THRESH=-1", "RNDV_THRESH=-1")
 {
     test_am(4 * UCS_KBYTE);
-}
-
-UCS_TEST_P(test_ucp_am_nbx_dts, long_bcopy_send, "ZCOPY_THRESH=-1",
-                                                 "RNDV_THRESH=-1")
-{
     test_am(64 * UCS_KBYTE);
 }
 
-UCS_TEST_P(test_ucp_am_nbx_dts, short_zcopy_send, "ZCOPY_THRESH=1",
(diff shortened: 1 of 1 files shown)


test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1231,26 +1231,12 @@ UCS_TEST_P(test_ucp_am_nbx_dts, short_send, "ZCOPY_THRESH=-1", "RNDV_THRESH=-1")
     test_am(1);
 }
 
-UCS_TEST_P(test_ucp_am_nbx_dts, short_bcopy_send, "ZCOPY_THRESH=-1",
-                                                  "RNDV_THRESH=-1")
+UCS_TEST_P(test_ucp_am_nbx_dts, bcopy_send, "ZCOPY_THRESH=-1", "RNDV_THRESH=-1")
```
yosefe, line 1234: how about setting ZCOPY_THRESH=16k RNDV_THRESH=32k, and the running 20 random sizes in the range 0..64K?

brminich: it may cover various corner cases, like sending IOV dt with length 0. That's why tag matching rndv tests found some errors with rndv and non-contig datatypes

rakhmets: then it is better to have specific test for the case. otherwise, it may not detect the exact moment when the error occurred

rakhmets: Returned back number of sending operations.
The change now is only that two different sizes are combined in a single test.
https://github.com/openucx/ucx/pull/8323#discussion_r901125577

brminich: how much time does it save? Maybe we can also unite bcopy and zcopy?

yosefe (dismissed):

test/gtest/ucp/test_ucp_am.cc
rakhmets: The time of executing _bcopy_ and _zcopy_ tests reduced from **1367 s** to **785 s**. I used the following command `gtest --gtest_filter="*test_ucp_am_nbx_dts*copy_send*"` to run the test on `dc_x, ud, ud_x, rc, rc_x, shm_ib, self, tcp` transports.

_bcopy_ and _zcopy_ tests have different `ZCOPY_THRESH` to be sure that tests use the right operation.
I think the tests can be left as separate tests for now.

brminich (approved):
plz, squash

PR #8335, opened 2022-06-23
UCP/MM: Track ucp_mem_map/unmap refcount

Artemy-Mellanox:


src/ucp/core/ucp_mm.c
```diff
@@ -781,7 +787,7 @@ ucp_mpool_free(ucp_worker_h worker, ucs_mpool_t *mp, void *chunk)
     ucp_mem_desc_t *chunk_hdr;
 
     chunk_hdr = (ucp_mem_desc_t*)chunk - 1;
-    ucp_memh_put(worker->context, chunk_hdr->memh, 0);
+    ucp_memh_put(worker->context, chunk_hdr->memh, 1);
```
yosefe, line 790: why need to invalidate?
invalidate is only for error handling flows

Artemy-Mellanox: IMHO this action should revert action done at line 768 i.e. release memory. Without invalidating it will leave rcache entry in pgtable and will not release memory.

yosefe: i would expect that when we do the actual memory release (like munmap or shmdt), it will trigger UCM hooks to invalidate the rcache entry.

yosefe: also, this change seems not related to ucp_mem_map/unmap flow

src/ucp/core/ucp_mm.c
```diff
@@ -479,6 +474,8 @@ ucp_memh_alloc(ucp_context_h context, void *address, size_t length,
         goto err;
     }
 
+    ucs_assert(memh->map_count == 1);
```
yosefe, line 477: 1. assertv
2. how does it work? seems ucp_mem_map increments map_count only after calling this function

src/ucp/core/ucp_mm.c
```diff
@@ -831,7 +832,7 @@ ucp_rndv_frag_free_mpools(ucs_mpool_t *mp, void *chunk)
     ucp_rndv_frag_mp_chunk_hdr_t *chunk_hdr;
 
     chunk_hdr = (ucp_rndv_frag_mp_chunk_hdr_t*)chunk - 1;
-    ucp_memh_put(mpriv->worker->context, chunk_hdr->memh, 0);
+    ucp_memh_put(mpriv->worker->context, chunk_hdr->memh, 1);
```
yosefe, line 835: same here, mpool flow seems irrelevant for ucp_mem_map/unmap

src/ucp/core/ucp_mm.c
```diff
@@ -583,13 +578,17 @@ ucs_status_t ucp_mem_map(ucp_context_h context, const ucp_mem_map_params_t *para
...
+    }
+
 out:
     return status;
 }
 
 ucs_status_t ucp_mem_unmap(ucp_context_h context, ucp_mem_h memh)
 {
-    ucp_memh_put(context, memh, 1);
+    ucp_memh_put(context, memh, !--memh->map_count);
```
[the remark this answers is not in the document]
yosefe: i would expect that even when UCP performs the allocation, it would still go through system memory allocator (eventually), so it will also trigger memory hooks for release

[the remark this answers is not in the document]
Artemy-Mellanox: without rcache it would.
but rcache will keep this memory referenced by pgtable even after it was seemingly put and this will prevent it's release.
it just won't reach system dealocator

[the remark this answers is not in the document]
yosefe: i see, is it because we release from system memory only when rcache completely removes the region?
so, can we invalidate only in UCP rcache, but without calling uct_mem_dereg_v2(INVALIDATE) ?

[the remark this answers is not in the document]
Artemy-Mellanox: Sure, rcache invalidation has no direct relation to UCT invalidation.

yosefe (dismissed):

yosefe:
@Artemy-Mellanox can u pls squash?

PR #8340, opened 2022-06-28
AZP: Kill own iodemo process in CI

avildema:
## Why ?
run_io_demo.sh can kill iodemo process from other test

## How ?
grep by IODEMO_ROLE (it's applicable only for process which launch via run_io_demo.sh

### DIFF
--- a/test/apps/iodemo/run_io_demo.sh
+++ b/test/apps/iodemo/run_io_demo.sh
@@ -538,7 +538,7 @@ make_scripts()
 			}
 
 			kill_iodemo() {
-			    pids="\$(list_pids)"
+			    pids="\$(list_pids_with_role all)"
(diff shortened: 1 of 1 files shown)


yosefe:
what about prep by path?

avildema:
> what about prep by path?

I think that it is not good idea 

Path is `247355 ?        Rl     0:10 drop_46778/install/bin/io_demo -d`

and if we grep by pattern we will depend on this path . we change this path (or azure change) and this approach will be fail

yosefe:
@avildema can we check "readlink -f /proc/$pid/exe" ?

avildema:
> @avildema can we check "readlink -f /proc/$pid/exe" ?

can, but why, if we have process we have `/proc/\${pid}/environ`

yosefe:
Ok

PR #8341, opened 2022-06-28
GTEST/UCP: Replaced variants by iterating datatypes in AM tests.

rakhmets:
## What
Replaced using of variants by iterating through datatypes inside a test in UCP AM gtests.

## Why ?
The objective of the changes is to reduce test execution time while keeping coverage at the same level.

### DIFF
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -350,6 +350,8 @@ public:
         m_rx_memh     = NULL;
     }
 
+    void test_datatypes(std::function<void()> test_f);
+
 protected:
     virtual ucs_memory_type_t tx_memtype() const
     {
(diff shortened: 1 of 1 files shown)


test/gtest/ucp/test_ucp_am.cc
```diff
@@ -350,6 +350,8 @@ class test_ucp_am_nbx : public test_ucp_am_base {
         m_rx_memh     = NULL;
     }
 
+    void test_datatypes(std::function<void()> send_f);
```
brminich, line 353: according to the usage it's better to call `test_f`

rakhmets:
> how much time does it save?

Execution time of UCP Active message datatype tests, including eager and rendezvous, went down from **20.5 minutes** to **6 minutes** on my setup (dc_x, ud, ud_x, rc, rc_x, shm_ib, self, tcp transports).

brminich (dismissed):

yosefe (approved):

brminich (approved):

PR #8343, opened 2022-06-28
UCP/WIREUP: Always pack TL resource indices into CM private data

dmitrygx:
## What

Always pack TL resource indices into CM private data.

## Why ?

TL resources must be packed into CM private data to make sure that `ucp_wireup_init_lanes()` uses correct destination resource indices to check config intersections.
1. `dst_rsc_indices` stored in CM wireup endpoint are by `UCP_NULL_RESOURCE`s when client's EP was created.
2. `dst_rsc_indices` are set to the proper values on a client after receiving CM private data from a server.
3. `dst_rsc_indices` will be used to check config intersections during WIREUP_MSG phase, if `UCX_CM_USE_ALL_DEVICES=y`

## How ?

1. Update `ucp_ep_create_server_accept()` to use `ucp_cm_address_pack_flags()` for unpacking remote worker address to make sure the same flags are used for packing/unpacking CM private data.
2. Update `UCP_ADDRESS_PACK_FLAGS_CM_DEFAULT` to include `UCP_ADDRESS_PACK_FLAG_TL_RSC_IDX`.
3. Update `ucp_wireup_check_config_intersect()` to check that packed `dst_rsc_index` isn't `UCP_NULL_RESOURCE` on worker address.

### DIFF
--- a/src/ucp/wireup/wireup.c
+++ b/src/ucp/wireup/wireup.c
@@ -935,7 +935,7 @@ ucp_wireup_connect_lane_to_iface(ucp_ep_h ep, ucp_lane_index_t lane,
              * is WIREUP MSG lane) until CM and WIREUP_MSG phases are done. The
              * lane is added during WIREUP_MSG exchange or created as an initial
              * configuration after a connection request on a server side */
-            status = ucp_wireup_ep_create(ep, &wireup_ep);
+            status = ucp_wireup_ep_create(ep, NULL, &wireup_ep);
             if (status != UCS_OK) {
                 /* coverity[leaked_storage] */
                 return status;
@@ -975,7 +975,7 @@ ucp_wireup_connect_lane_to_ep(ucp_ep_h ep, unsigned ep_init_flags,
     ucs_status_t status;
 
     if (ucp_ep_get_lane(ep, lane) == NULL) {
-        status = ucp_wireup_ep_create(ep, &uct_ep);
+        status = ucp_wireup_ep_create(ep, NULL, &uct_ep);
         if (status != UCS_OK) {
             /* coverity[leaked_storage] */
             return status;
@@ -1175,6 +1175,43 @@ ucp_wireup_get_reachable_mds(ucp_ep_h ep, unsigned ep_init_flags,
     key->reachable_md_map = dst_md_map;
 }
 
+void
+ucp_wireup_get_dst_rsc_indices(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
+                               const ucp_unpacked_address_t *remote_address,
+                               const unsigned *addr_indices,
+                               ucp_rsc_index_t *dst_rsc_indices)
+{
+    ucp_lane_index_t lane;
+    unsigned addr_index;
+    ucp_rsc_index_t dst_rsc_index;
+    ucp_address_entry_t *ae;
+
+    for (lane = 0; lane < new_key->num_lanes; ++lane) {
+        addr_index = addr_indices[lane];
+
+        if (lane == ucp_ep_get_cm_lane(ep)) {
+            ucs_assert(addr_index == UINT_MAX);
+            dst_rsc_index = UCP_NULL_RESOURCE;
+        } else {
+            ucs_assert(addr_index != UINT_MAX);
+            ae            = &remote_address->address_list[addr_index];
+            dst_rsc_index = ae->iface_attr.dst_rsc_index;
+            if (ep->worker->context->config.ext.cm_use_all_devices) {
+                ucs_assert(dst_rsc_index != UCP_NULL_RESOURCE);
+            }
+        }
+
+        /* Save destination resource index in the CM wireup EP for doing
+         * further intersections, if needed */
+        dst_rsc_indices[lane] = dst_rsc_index;
+    }
+
+    /* Fill remaining elements of 'dst_rsc_indices' by UCP_NULL_RESOURCE */
+    for (lane = new_key->num_lanes; lane < UCP_MAX_LANES; ++lane) {
+        dst_rsc_indices[lane] = UCP_NULL_RESOURCE;
+    }
+}
+
 static void
 ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
                                   const ucp_unpacked_address_t *remote_address,
@@ -1185,13 +1222,9 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
     uct_ep_h new_uct_eps[UCP_MAX_LANES]                = { NULL };
     ucp_lane_index_t reuse_lane_map[UCP_MAX_LANES]     = { UCP_NULL_LANE };
     ucp_rsc_index_t old_dst_rsc_indices[UCP_MAX_LANES] = { UCP_NULL_RESOURCE };
-    ucp_rsc_index_t new_dst_rsc_indices[UCP_MAX_LANES] = { UCP_NULL_RESOURCE };
     ucp_wireup_ep_t *cm_wireup_ep                      = NULL;
     ucp_ep_config_key_t *old_key;
     ucp_lane_index_t lane, reuse_lane;
-    ucp_address_entry_t *ae;
-    unsigned addr_index;
-    ucp_rsc_index_t dst_rsc_index;
     uct_ep_h uct_ep;
 
     *connect_lane_bitmap = UCS_MASK(new_key->num_lanes);
@@ -1214,28 +1247,14 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
 
     memcpy(old_dst_rsc_indices, cm_wireup_ep->dst_rsc_indices,
            sizeof(old_dst_rsc_indices));
-    for (lane = 0; lane < new_key->num_lanes; ++lane) {
-        addr_index = addr_indices[lane];
-
-        if (lane == ucp_ep_get_cm_lane(ep)) {
-            ucs_assert(addr_index == UINT_MAX);
-            dst_rsc_index = UCP_NULL_RESOURCE;
-        } else {
-            ucs_assert(addr_index != UINT_MAX);
-            ae            = &remote_address->address_list[addr_index];
-            dst_rsc_index = ae->iface_attr.dst_rsc_index;
-        }
-
-        /* save destination resource index in the CM wireup EP for doing
-         * further intersections, if needed */
-        cm_wireup_ep->dst_rsc_indices[lane] = dst_rsc_index;
-        new_dst_rsc_indices[lane]           = dst_rsc_index;
-    }
+    ucp_wireup_get_dst_rsc_indices(ep, new_key, remote_address, addr_indices,
+                                   cm_wireup_ep->dst_rsc_indices);
 
     old_key = &ucp_ep_config(ep)->key;
 
     ucp_ep_config_lanes_intersect(old_key, old_dst_rsc_indices, new_key,
-                                  new_dst_rsc_indices, reuse_lane_map);
+                                  cm_wireup_ep->dst_rsc_indices,
+                                  reuse_lane_map);
 
     /* CM lane has to be re-used by the new EP configuration */
     ucs_assert(reuse_lane_map[ucp_ep_get_cm_lane(ep)] != UCP_NULL_LANE);
@@ -1492,7 +1511,7 @@ ucs_status_t ucp_wireup_connect_remote(ucp_ep_h ep, ucp_lane_index_t lane)
 
     ucs_trace("ep %p: connect lane %d to remote peer with wireup ep", ep, lane);
 
-    status = ucp_wireup_ep_create(ep, &wireup_ep);
+    status = ucp_wireup_ep_create(ep, NULL, &wireup_ep);
     if (status != UCS_OK) {
         goto err;
     }
--- a/src/ucp/wireup/wireup_cm.c
+++ b/src/ucp/wireup/wireup_cm.c
@@ -251,10 +251,15 @@ ucp_cm_priv_data_length(size_t addr_size, ucp_object_version_t sa_data_version)
     return addr_size + ucp_cm_sa_data_length(sa_data_version);
 }
 
-static unsigned ucp_cm_address_pack_flags(ucp_worker_h worker)
+unsigned ucp_cm_address_pack_flags(ucp_worker_h worker)
 {
     return ucp_worker_common_address_pack_flags(worker) |
-           UCP_ADDRESS_PACK_FLAGS_CM_DEFAULT;
+           UCP_ADDRESS_PACK_FLAGS_CM_DEFAULT |
+           /* Pack TL resource indices into CM private data to use them inside
+            * ucp_wireup_init_lanes() for checking intersection of endpoint
+            * configurations */
+           (worker->context->config.ext.cm_use_all_devices ?
+            UCP_ADDRESS_PACK_FLAG_TL_RSC_IDX : 0);
 }
 
 static void*
@@ -396,7 +401,7 @@ static ucs_status_t ucp_cm_ep_init_lanes(ucp_ep_h ep,
             continue;
         }
 
-        status = ucp_wireup_ep_create(ep, &uct_ep);
+        status = ucp_wireup_ep_create(ep, NULL, &uct_ep);
         if (status != UCS_OK) {
             goto out;
         }
@@ -1211,6 +1216,7 @@ ucp_ep_cm_server_create_connected(ucp_worker_h worker, unsigned ep_init_flags,
     char client_addr_str[UCS_SOCKADDR_STRING_LEN];
     ucp_wireup_sockaddr_data_base_t *sa_data;
     ucp_object_version_t sa_data_version;
+    unsigned addr_indices[UCP_MAX_LANES];
 
     ep_init_flags |= UCP_EP_INIT_CM_WIREUP_SERVER | UCP_EP_INIT_CM_PHASE;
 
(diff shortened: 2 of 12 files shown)


yosefe:
1. Is this a bugfix? if yes which test fails?
2. maybe need to pack the RSC index only if "UCX_CM_USE_ALL_DEVICES=y" ?
3. does it increase address size?

dmitrygx:
> * Is this a bugfix? if yes which test fails?

Yes, it is bugfix.
I managed to reproduce the connectivity issue using `ucp_client_server` when running client with `UCX_NET_DEVICES=ip_iface1` and server with `UCX_NET_DEVICES=ip_iface2,ip_iface3` on another node. `ip_iface1` has connectivity with `ip_iface2` and `ip_iface3`. During CM phase server selected `ip_iface2` for EP configuration, but then selected `ip_iface3`. Client doesn't change the configuration in its turn, because client doesn't know destination TL resources.

> * maybe need to pack the RSC index only if "UCX_CM_USE_ALL_DEVICES=y" ?

Yes, we could do it to decrease CM private data in case of `UCX_CM_USE_ALL_DEVICES=n`.
Shall we?

> * does it increase address size?

Yes, it increases CM private data size by 1 byte per packed MD.

yosefe:
> Yes, we could do it to decrease CM private data in case of `UCX_CM_USE_ALL_DEVICES=n`.
> Shall we?

Yes, please.
Can we add unit test in UCX CI to reproduce it? or some special iodemo variant?

dmitrygx:
> > Can we add unit test in UCX CI to reproduce it? or some special iodemo or ucx_client_server variant?
> 
> I guess running io-demo for this issue is time-consuming. I'll try to add sockaddr gtest which restricts `UCX_NET_DEVICES` for client.

@yosefe done

dmitrygx:

src/ucp/wireup/wireup.c
```diff
@@ -1224,6 +1224,8 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
             ucs_assert(addr_index != UINT_MAX);
             ae            = &remote_address->address_list[addr_index];
             dst_rsc_index = ae->iface_attr.dst_rsc_index;
+            ucs_assert(!ep->worker->context->config.ext.cm_use_all_devices ||
```
yosefe, line 1227: pls use if condition to simplify

src/ucp/wireup/wireup_cm.c
```diff
@@ -251,10 +251,12 @@ ucp_cm_priv_data_length(size_t addr_size, ucp_object_version_t sa_data_version)
     return addr_size + ucp_cm_sa_data_length(sa_data_version);
 }
 
-static unsigned ucp_cm_address_pack_flags(ucp_worker_h worker)
+unsigned ucp_cm_address_pack_flags(ucp_worker_h worker)
 {
     return ucp_worker_common_address_pack_flags(worker) |
-           UCP_ADDRESS_PACK_FLAGS_CM_DEFAULT;
+           UCP_ADDRESS_PACK_FLAGS_CM_DEFAULT |
+           (worker->context->config.ext.cm_use_all_devices ?
```
yosefe, line 258: pls add commeny why

test/gtest/common/test_helpers.cc
```diff
@@ -534,24 +534,35 @@ static std::set<std::string> get_all_rdmacm_net_devices()
...
+                                ucs::to_string(port_num));
             }
         }
     }
 
     return devices;
 }
 
-bool is_rdmacm_netdev(const char *ifa_name) {
+bool is_rdmacm_netdev(const char *ifa_name, std::string *dev_name)
```
yosefe, line 547: maybe return std::string (and it could be empty)

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -2757,6 +2764,87 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_diff_config,
...
+
+        return net_device;
+    }
+
+    void create_entities_and_connect()
+    {
+        std::string net_device = find_net_device();
+
+        {
+            modify_config("NET_DEVICES", net_device, FAIL_IF_NOT_EXIST);
```
yosefe, line 2821: can you document what is this test doing ? maybe give example with specific device names

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -2757,6 +2764,87 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_diff_config,
...
+    std::string find_net_device()
+    {
+        auto e = create_entity(); // temporary entity
+
+        std::string net_device;
+        ucp_rsc_index_t tl_idx;
+        for (tl_idx = 0; tl_idx < sender().ucph()->num_tls; ++tl_idx) {
+            uct_tl_resource_desc_t *tl_rsc = &e->ucph()->tl_rscs[tl_idx].tl_rsc;
+            if (tl_rsc->dev_type == UCT_DEVICE_TYPE_NET) {
+                if (m_test_addr.netdev_name().compare(tl_rsc->dev_name) == 0) {
```
yosefe, line 2798: do we really have to add infra to cache netdev_name in m_test_addr?
can't just find the netdev in this specific test?

test/gtest/ucp/test_ucp_sockaddr.cc
dmitrygx: also, found bug on a server when restricting `UCX_NET_DEVICES` for server instead of client

src/ucp/wireup/wireup_cm.c
```diff
@@ -1266,6 +1272,12 @@ ucp_ep_cm_server_create_connected(ucp_worker_h worker, unsigned ep_init_flags,
         goto err_destroy_ep;
     }
 
+    cm_wireup_ep = ucp_ep_get_cm_wireup_ep(ep);
+    ucs_assert(cm_wireup_ep != NULL);
+    ucp_wireup_dst_rsc_indices_fill(ep, &ucp_ep_config(ep)->key, remote_addr,
```
yosefe, line 1277: why need to do it here?
ucp_ep_create_to_worker_addr() already calls ucp_wireup_init_lanes() which calls ucp_wireup_check_config_intersect()

src/ucp/wireup/wireup.c
```diff
@@ -1214,25 +1241,13 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
...
-        }
-
-        /* save destination resource index in the CM wireup EP for doing
-         * further intersections, if needed */
-        cm_wireup_ep->dst_rsc_indices[lane] = dst_rsc_index;
-        new_dst_rsc_indices[lane]           = dst_rsc_index;
-    }
+    /* Save destination resource index in the CM wireup EP for doing
+     * further intersections, if needed */
+    memcpy(cm_wireup_ep->dst_rsc_indices, new_dst_rsc_indices,
```
yosefe, line 1249: can we pass cm_wireup_ep->dst_rsc_indices directly to ucp_wireup_dst_rsc_indices_fill?

dmitrygx: done

src/ucp/wireup/wireup_cm.c
dmitrygx: added the comment

src/ucp/wireup/wireup_cm.c
```diff
@@ -1272,11 +1271,10 @@ ucp_ep_cm_server_create_connected(ucp_worker_h worker, unsigned ep_init_flags,
         goto err_destroy_ep;
     }
 
-    cm_wireup_ep = ucp_ep_get_cm_wireup_ep(ep);
-    ucs_assert(cm_wireup_ep != NULL);
-    ucp_wireup_dst_rsc_indices_fill(ep, &ucp_ep_config(ep)->key, remote_addr,
-                                    addr_indices,
-                                    cm_wireup_ep->dst_rsc_indices);
+    /* Since CM lane isn't available and no previous configuration available on
+     * the endpoint, save peer's resource indices after creating CM lane */
```
yosefe, line 1275: it means the previously saved `cm_wireup_ep->dst_rsc_indices` would not be valid prior to calling `ucp_wireup_check_config_intersect`?
why is this not needed also for client code?

test/gtest/common/test_helpers.cc
```diff
@@ -504,10 +504,18 @@ static std::set<std::string> get_all_rdmacm_net_devices()
...
+                                              "/device/infiniband";
+        std::vector<std::string> ib_devices = read_dir(infiniband_dir);
+
+        for (auto &ib_device : ib_devices) {
+            std::string ports_dir             = infiniband_dir + "/" +
+                                                ib_device + "/ports";
+            std::vector<std::string> ib_ports = read_dir(ports_dir);
+
+            for (auto &ib_port : ib_ports) {
+                devices.emplace(ndevs[i], ib_device + ":" + ib_port);
```
yosefe, line 517: seems only one ib dev will be saved eventually per ndev, so why need to loop over all devices and ports?

test/gtest/common/test_helpers.cc
```diff
@@ -534,24 +542,33 @@ static std::set<std::string> get_all_rdmacm_net_devices()
...
+                                ucs::to_string(port_num));
             }
         }
     }
 
     return devices;
 }
 
-bool is_rdmacm_netdev(const char *ifa_name) {
+std::string rdmacm_netdev(const char *ifa_name)
```
yosefe, line 555: get_rdmacm_netdev

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -176,6 +176,26 @@ class test_ucp_sockaddr : public ucp_test {
...
+        bool result           = false;
+
+        for (ucp_rsc_index_t cm_idx = 0; cm_idx < context->config.num_cm_cmpts;
+             ++cm_idx) {
+            std::string cm_name = ucp_context_cm_name(context, cm_idx);
+            if (cm_name.compare("rdmacm") == 0) {
+                result = true;
+                goto out;
+            }
+        }
```
yosefe, line 192: can we make this test less intrusive, without so much looking into UCP internals?

test/gtest/ucp/test_ucp_sockaddr.cc
yosefe: what is the point of returning net device?
seems just need to skip the test if m_test_addr.netdev_name() is not in the UCP resource list.
But could this really happen? we could skip if create_entity fails with wrong UCX_NET_DEVICES...

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -2757,6 +2785,119 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_diff_config,
...
+                            m_test_addr.netdev_name() + "\" CM iface");
+        }
+
+        return net_device;
+    }
+
+    void modify_net_devices(const std::string &net_device, bool only_cm_device)
+    {
+        if (only_cm_device) {
+            modify_config("NET_DEVICES", net_device, FAIL_IF_NOT_EXIST);
```
yosefe, line 2857: `modify_config("NET_DEVICES", only_net_device ? m_test_addr.netdev_name() : "all", FAIL_IF_NOT_EXIST);`

src/ucp/wireup/wireup_cm.c
dmitrygx: > it means the previously saved cm_wireup_ep->dst_rsc_indices would not be valid prior to calling ucp_wireup_check_config_intersect?

there are no previously saved `cm_wireup_ep->dst_rsc_indices` at this point, all indices are set to `UCP_NULL_RESOURCE`

> why is this not needed also for client code?

client creates CM lane, connects to the server and only after that it gets server's CM private data and `ucp_wireup_check_config_intersect` saves it to `cm_wireup_ep->dst_rsc_indices`

test/gtest/common/test_helpers.cc
dmitrygx: > seems only one ib dev will be saved eventually per ndev, so why need to loop over all devices and ports?

yes, it should be only one `device:port` pair.
maybe better to add assertion - to  make sure it is true?

test/gtest/common/test_helpers.cc
dmitrygx: done

test/gtest/ucp/test_ucp_sockaddr.cc
dmitrygx: we need to know whether RDMACM is supported or not to set `UCX_NET_DEVICES` to either IB device name (to create auxiliary IB transport) or to IP interface name (to create auxiliary TCP transport).

test/gtest/ucp/test_ucp_sockaddr.cc
dmitrygx: removed the function

test/gtest/ucp/test_ucp_sockaddr.cc
dmitrygx: done

src/ucp/wireup/wireup_cm.c
yosefe: for client side: seems ucp_wireup_check_config_intersect compares previous cm_wireup_ep->dst_rsc_indices to the new ones. in this case, AFAIK the previous ones will be all UCP_NULL_RESOURCE which is ok?

test/gtest/common/test_helpers.cc
yosefe: IMO no need, just take first one as previously

test/gtest/ucp/test_ucp_sockaddr.cc
yosefe: what if we set to both?

src/ucp/wireup/wireup_cm.c
dmitrygx: > for client side: seems ucp_wireup_check_config_intersect compares previous cm_wireup_ep->dst_rsc_indices to the new ones. in this case, AFAIK the previous ones will be all UCP_NULL_RESOURCE which is ok?

yes, it is ok for the first iteration, we don't expect reconfiguration.
but then we will have updated `cm_wireup_ep->dst_rsc_indices` which will be used during WIREUP_MSG phase to find best configuration

test/gtest/common/test_helpers.cc
dmitrygx: ok, done

src/ucp/wireup/wireup_cm.c
dmitrygx: > so maybe the problem is the server does not call ucp_wireup_check_config_intersect (like the client)?

it calls, but before CM lane is created.

dmitrygx: if we will create CM lane and save it to UCP EP's `uct_eps` before real configuration exists.
actually, client and server behaviors are aligned:
Client/Server creates UCP EP, does `ucp_wireup_init_lanes` whichs call `ucp_wireup_check_config_intersect ` - but it is just called, doesn't do any real work, since no CM lane and previously create configuration and then create CM lane

dmitrygx: the only difference is that when creating CM lane on sever, we already know destination resource indices and we set them after CM lane is created.
these indices will be used during next `ucp_wireup_init_lanes` and `ucp_wireup_check_config_intersect`

dmitrygx: moved `ucp_wireup_dst_rsc_indices_save` to `ucp_wireup_ep_create`

dmitrygx: @yosefe is it what you meant?

src/ucp/wireup/wireup_cm.c
```diff
@@ -1396,7 +1405,7 @@ ucs_status_t ucp_ep_cm_connect_server_lane(ucp_ep_h ep,
     ucs_assert(ucp_ep_get_lane(ep, lane) == NULL);
 
     /* TODO: split CM and wireup lanes */
-    status = ucp_wireup_ep_create(ep, &uct_ep);
+    status = ucp_wireup_ep_create(ep, remote_address, addr_indices, &uct_ep);
```
yosefe, line 1408: instead of passing both remote_address and addr_indices to ucp_wireup_ep_create, pass only "dst_addr_indices" - after calling ucp_wireup_get_dst_rsc_indices() to calculate them to a local array
this way we would pass minimal information to ucp_wireup_ep_create constructor

test/gtest/common/test_helpers.cc
```diff
@@ -504,10 +504,17 @@ static std::set<std::string> get_all_rdmacm_net_devices()
...
-        std::string infiniband_dir = sysfs_net_dir + "/" + ndevs[i] +
-                                     "/device/infiniband";
-        if (!read_dir(infiniband_dir).empty()) {
-            devices.insert(ndevs[i]);
+        std::string infiniband_dir          = sysfs_net_dir + "/" + ndevs[i] +
+                                              "/device/infiniband";
+        std::vector<std::string> ib_devices = read_dir(infiniband_dir);
+
+        if (!ib_devices.empty()) {
+            std::string ib_device = ib_devices[0];
```
yosefe, line 512: minor: maybe ib_devices.front()

src/ucp/wireup/wireup_cm.c
dmitrygx: to calculate `addr_indices` we should do `ucp_wireup_select_lanes` once again

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -192,9 +212,14 @@ class test_ucp_sockaddr : public ucp_test {
...
             ASSERT_UCS_OK(status);
-            saddrs.back().set_sock_addr(*ifa->ifa_addr, size,
-                                        ucs::is_rdmacm_netdev(ifa->ifa_name));
-            saddrs.back().set_port(0); /* listen on any port then update */
+
+            std::string rdmacm_netdev_name =
+                    ucs::get_rdmacm_netdev(ifa->ifa_name);
+            bool is_rdmacm_netdev          = !rdmacm_netdev_name.empty();
+            saddrs.back().set_sock_addr(*ifa->ifa_addr, size, is_rdmacm_netdev,
+                                        ifa->ifa_name, rdmacm_netdev_name);
```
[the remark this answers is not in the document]
dmitrygx: done

src/ucp/wireup/wireup_cm.c
dmitrygx: done as suggested by @yosefe above

yosefe (approved):

yosefe (approved):

PR #8346, opened 2022-06-30
UCP/CORE: Don't unset CMs from temporary CM bitmap

dmitrygx:
## What

Don't unset CMs from temporary CM bitmap.

## Why ?

Both `cm_cmpts_bitmap_safe` and `cm_cmpts_bitmap` bitmaps are temporary. So, unsetting bits from them doesn't have any effects.

## How ?

1. Remove `cm_cmpts_bitmap` and `cm_cmpts_bitmap_safe` in `ucp_fill_sockaddr_cms_prio_list`.
2. Remove doing `UCS_BITMAP_UNSET` for CM index which is not going to be used.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -1223,8 +1223,6 @@ static void ucp_fill_sockaddr_cms_prio_list(ucp_context_h context,
                                             const char **sockaddr_cm_names,
                                             ucp_rsc_index_t num_sockaddr_cms)
 {
-    ucp_tl_bitmap_t cm_cmpts_bitmap = context->config.cm_cmpts_bitmap;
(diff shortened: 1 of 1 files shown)


dmitrygx:
@evgeny-leksikov @gleon99 could you review pls?

gleon99:
@dmitrygx what about unsetting some CMs (line 1245), then in the next iteration of the outer `for` loop (line 1234) we don't iterate over those CMs (because of lines 1238-1239)? Seems this behavior is modified by your commit.

dmitrygx:
> @dmitrygx what about unsetting some CMs (line 1245), then in the next iteration of the outer `for` loop (line 1234) we don't iterate over those CMs (because of lines 1238-1239)?

@gleon99 yes, you're right - didn't notice it

PR #8289, opened 2022-06-05
UCT/IB/DC: Always schedule DCI allocation during FC_HARD_REQ progress

dmitrygx:
## What

 Always schedule DCI allocation during FC_HARD_REQ progress.

## Why ?

To fix not sending `FC_HARD_REQ` in case of no DCI is allocated for EP.

## How ?

1. Update `uct_dc_mlx5_ep_schedule` and `uct_dc_mlx5_iface_schedule_dci_alloc` to accept `force` parameter.
2. If `force` specified in `uct_dc_mlx5_iface_schedule_dci_alloc`, schedule EP on DCI wait queue.
3. Pass `force=1` to `uct_dc_mlx5_ep_schedule` in `uct_dc_mlx5_ep_fc_hard_req_progress`.

PR #8348, opened 2022-07-01
UCT/IB/DC: Resend FC_HARD_REQ instead of scheduling EP on waitq

dmitrygx:
## Why
Fix "timeout waiting for replies" in IO-demo (30 iterations passed successfully with the fix).
Happened because all DCIs were acquired by endpoints for sending FC hard requests, but did not actually send them, because no UCT-API operations were done on these endpoints.

## How 
1. Partially revert changes from #8169 and #8289 PRs.
2. Invoked `uct_dc_mlx5_ep_check_fc` from `uct_dc_mlx5_ep_fc_hard_req_progress`'s loop over all EPs added to FC hash.

## Details 
UCT user may not have pending operations scheduled, but FC_HARD_REQ could be posted and then resend.
If DC transport will acquire DCI, it couldn't be used at all if user doesn't post new operations and no pending operations scheduled.
The following scenario was observed:
`FC_HARD_LIMIT` was reached on DC endpoint and `FC_HARD_REQ` packet was sent. But it took some time to receive `FC_GRANT` from a peer, so the endpoint did resend of `FC_HARD_REQ`, but no pending operation were scheduled on the endpoint - so, DCI was acquired, but not released. And finally, `FC_GRANT` was received and FC window was restored on the endpoint. S, this DCI was acquired by the endpoint forever, because IO-demo did not schedule new operations.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5_ep.c
+++ b/src/uct/ib/dc/dc_mlx5_ep.c
@@ -1493,17 +1493,12 @@ static unsigned uct_dc_mlx5_ep_fc_hard_req_progress(void *arg)
      * resend FC_HARD_REQ packet to make sure a peer will resend FC_PURE_GRANT
      * packet in case of failure on the remote FC endpoint */
     kh_foreach_key(&iface->tx.fc_hash, ep_key, {
-        ep = (uct_dc_mlx5_ep_t*)ep_key;
-
-        /* Allocate DCI for the endpoint to schedule the endpoint to DCI wait
-         * queue if there is free DCI */
-        status = uct_dc_mlx5_iface_dci_get(iface, ep);
-        ucs_assertv((status == UCS_OK) || (status == UCS_ERR_NO_RESOURCE),
-                    "%s", ucs_status_string(status));
-
-        /* Force DCI scheduling, since FC resources may never become available
-         * unless we send FC_HARD_REQ packet */
-        uct_dc_mlx5_ep_schedule(iface, ep, 1);
+        ep     = (uct_dc_mlx5_ep_t*)ep_key;
+        status = uct_dc_mlx5_ep_check_fc(iface, ep);
+        if ((status == UCS_OK) || (status == UCS_ERR_NO_RESOURCE)) {
+            ucs_warn("ep %p: flow-control check failed: %s", ep,
+                     ucs_status_string(status));
+        }
     })
 
     return 1;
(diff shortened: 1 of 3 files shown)


dmitrygx:
It fixes "timeout waiting for replies" in IO-demo (30 iterations passed successfuly). See PR description where the scenario was explained.

yosefe:
@dmitrygx how "this DCI was acquired by the endpoint forever" caused "timeout waiting for replies"?
did you see all DCIs were allocated to such endpoints and never released, so other endpoints could not send anything?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -378,13 +378,13 @@ static inline int uct_dc_mlx5_iface_dci_ep_can_send(uct_dc_mlx5_ep_t *ep)
...
-                                          uct_dc_mlx5_ep_t *ep, int force)
+                                          uct_dc_mlx5_ep_t *ep)
 {
     ucs_arbiter_t *waitq;
 
-    /* If FC window is empty and force scheduling wasn't requested, the group
-     * will be scheduled when grant is received */
-    if (force || uct_rc_fc_has_resources(&iface->super.super, &ep->fc)) {
+    /* If FC window is empty the group will be scheduled when grant is received
+     */
```
yosefe, line 386: `*/` on prev line

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1493,17 +1493,10 @@ static unsigned uct_dc_mlx5_ep_fc_hard_req_progress(void *arg)
      * resend FC_HARD_REQ packet to make sure a peer will resend FC_PURE_GRANT
      * packet in case of failure on the remote FC endpoint */
     kh_foreach_key(&iface->tx.fc_hash, ep_key, {
-        ep = (uct_dc_mlx5_ep_t*)ep_key;
-
-        /* Allocate DCI for the endpoint to schedule the endpoint to DCI wait
-         * queue if there is free DCI */
-        status = uct_dc_mlx5_iface_dci_get(iface, ep);
+        ep     = (uct_dc_mlx5_ep_t*)ep_key;
```
yosefe, line 1496: ep helper var can be removed

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1493,17 +1493,10 @@ static unsigned uct_dc_mlx5_ep_fc_hard_req_progress(void *arg)
...
+        status = uct_dc_mlx5_ep_check_fc(iface, ep);
```
yosefe, line 1497: IMO line 1498 should be ucs_warn

dmitrygx:
> @dmitrygx how "this DCI was acquired by the endpoint forever" caused "timeout waiting for replies"?
did you see all DCIs were allocated to such endpoints and never released, so other endpoints could not send anything?

@yosefe yes, all DCIs are acquired by such endpoints and never released, because new operations are not scheduled.

yosefe (approved):

yosefe:
> yes, all DCIs are acquired by such endpoints and never released, because new operations are not scheduled.

Thanks, pls see updated PR desc

dmitrygx:
> >yes, all DCIs are acquired by such endpoints and never released, because new operations are not scheduled.

>Thanks, pls see updated PR desc


Thanks, these updates make sense to me.

PR #8356, opened 2022-07-06
UCT/CUDA: Improve performance estimation

yosefe:
## Why
- Remove the assumption that cuda_copy source is HOST and destination is DEVICE
- Set right CPU overhead/latency for sync/async operations


yosefe:
> @yosefe Changes look good for the time being. Probably need to follow up with a PR that decides bandwidth based on PCIe gen, lanes, and GPU architecture detection instead of values such as 8300, 7900, 11660, 9320 MB/s or 320 GB/s. Can you remind me if `bandwidth.shared` should report peak bandwidth or bandwidth for a specific message size?

It's the peak bandwidth so that if have multiple processes per node the bandwidth is shared/split between them.

PR #8357, opened 2022-07-06
UCP/PROTO/IB: Improve performance estimations for short/bcopy/zcopy

yosefe:
## Why
- Short protocol having lower latency should be implemented as UCT performance estimation and not assumptions on UCP layer
- Current IB latency estimation is actually the minimal value, so need to increase the cost of bcopy/zcopy rather than adding negative latency to short.

### DIFF
--- a/src/uct/base/uct_iface.h
+++ b/src/uct/base/uct_iface.h
@@ -975,11 +975,29 @@ void uct_tl_register(uct_component_t *component, uct_tl_t *tl);
 
 void uct_tl_unregister(uct_tl_t *tl);
 
+static UCS_F_ALWAYS_INLINE int uct_ep_op_is_short(uct_ep_operation_t op)
+{
+    return UCS_BIT(op) & (UCS_BIT(UCT_EP_OP_AM_SHORT) |
+                          UCS_BIT(UCT_EP_OP_PUT_SHORT) |
+                          UCS_BIT(UCT_EP_OP_GET_SHORT) |
+                          UCS_BIT(UCT_EP_OP_EAGER_SHORT) |
+                          UCS_BIT(UCT_EP_OP_ATOMIC_POST));
+}
+
+static UCS_F_ALWAYS_INLINE int uct_ep_op_is_bcopy(uct_ep_operation_t op)
+{
+    return UCS_BIT(op) & (UCS_BIT(UCT_EP_OP_AM_BCOPY) |
+                          UCS_BIT(UCT_EP_OP_PUT_BCOPY) |
+                          UCS_BIT(UCT_EP_OP_GET_BCOPY) |
+                          UCS_BIT(UCT_EP_OP_EAGER_BCOPY));
+}
+
 static UCS_F_ALWAYS_INLINE int uct_ep_op_is_zcopy(uct_ep_operation_t op)
 {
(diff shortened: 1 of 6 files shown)


src/uct/base/uct_iface.h
```diff
@@ -975,11 +975,28 @@ void uct_tl_register(uct_component_t *component, uct_tl_t *tl);
 
 void uct_tl_unregister(uct_tl_t *tl);
 
+static UCS_F_ALWAYS_INLINE int uct_ep_op_is_short(uct_ep_operation_t op)
+{
+    return UCS_BIT(op) & (UCS_BIT(UCT_EP_OP_AM_SHORT) |
+                          UCS_BIT(UCT_EP_OP_PUT_SHORT) |
+                          UCS_BIT(UCT_EP_OP_GET_SHORT) |
+                          UCS_BIT(UCT_EP_OP_ATOMIC_POST));
```
brminich, line 983: `UCT_EP_OP_EAGER_SHORT` is missing

src/uct/ib/base/ib_iface.c
```diff
@@ -1666,6 +1671,9 @@ uct_ib_iface_estimate_perf(uct_iface_h iface, uct_perf_attr_t *perf_attr)
 
     if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_LATENCY) {
         perf_attr->latency = iface_attr.latency;
+        if (uct_ep_op_is_bcopy(op) || uct_ep_op_is_zcopy(op)) {
+            perf_attr->latency.c += wqe_fetch;
```
brminich, line 1675: isn't it per sge in case of zcopy?

src/uct/ib/base/ib_iface.c
```diff
@@ -1641,11 +1642,15 @@ uct_ib_iface_estimate_perf(uct_iface_h iface, uct_perf_attr_t *perf_attr)
         break;
     default:
         send_pre_overhead        = iface_attr.overhead;
-        send_post_overhead_bcopy = send_post_overhead_zcopy = 0;
+        send_post_overhead_bcopy = 40e-9;
```
brminich, line 1645: is it only for calling packing function and obtaining bounce buffer?

src/uct/ib/base/ib_iface.c
yosefe: no, the whole wqe fetched at once

src/uct/ib/base/ib_iface.c
yosefe: it's for the cost of doorbell, actually "bcopy" also includes short case (see lines 1657-1661)

src/uct/base/uct_iface.h
yosefe: fixed by force push

src/uct/ib/base/ib_iface.c
brminich: maybe worth adding a comment that this is for doorbell?
also what does UCT report as a software overhead in iface_attr then?

yosefe: fixed (force push)
iface_attr.overhead is the preparation of posting the doorbell, and post_overhead would be the cost of writing to IO memory and its effect on instruction pipeline

brminich (approved):

PR #8361, opened 2022-07-07
UCP/PROTO/DEBUG: Extend performance tree infra

yosefe:
## Why
Support replacing node's children - will be used by rendezvous protocol to represent the remote system performance estimation

### DIFF
--- a/src/ucp/proto/proto_debug.c
+++ b/src/ucp/proto/proto_debug.c
@@ -544,8 +544,9 @@ void ucp_proto_perf_node_deref(ucp_proto_perf_node_t **perf_node_p)
     --perf_node->refcount;
     if (perf_node->refcount == 0) {
         ucp_proto_perf_node_free(perf_node);
-        *perf_node_p = NULL;
     }
+
+    *perf_node_p = NULL;
 }
 
 static void
@@ -580,6 +581,16 @@ void ucp_proto_perf_node_add_child(ucp_proto_perf_node_t *perf_node,
 
     ucp_proto_perf_node_append_child(perf_node, child_perf_node);
     ucp_proto_perf_node_ref(child_perf_node);
+}
+
+ucp_proto_perf_node_t *
+ucp_proto_perf_node_get_child(ucp_proto_perf_node_t *perf_node, unsigned n)
+{
+    if ((perf_node == NULL) || (n >= ucs_array_length(&perf_node->children))) {
+        return NULL;
+    }
+
+    return ucs_array_elem(&perf_node->children, n);
 }
 
 void ucp_proto_perf_node_add_data(ucp_proto_perf_node_t *perf_node,
@@ -626,6 +637,22 @@ const char *ucp_proto_perf_node_name(ucp_proto_perf_node_t *perf_node)
 const char *ucp_proto_perf_node_desc(ucp_proto_perf_node_t *perf_node)
 {
     return (perf_node == NULL) ? "(null)" : perf_node->desc;
+}
+
+void ucp_proto_perf_node_replace(ucp_proto_perf_node_t **old_perf_node_p,
+                                 ucp_proto_perf_node_t **new_perf_node_p)
+{
+    ucp_proto_perf_node_t **child_elem;
+
+    if (*old_perf_node_p != NULL) {
+        ucs_array_for_each(child_elem, &(*old_perf_node_p)->children) {
+            ucp_proto_perf_node_add_child(*new_perf_node_p, *child_elem);
+        }
+    }
+
+    ucp_proto_perf_node_deref(old_perf_node_p);
+    *old_perf_node_p = *new_perf_node_p;
+    *new_perf_node_p = NULL;
 }
 
 static void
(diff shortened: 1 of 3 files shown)


src/ucp/proto/proto_debug.c
```diff
@@ -544,8 +544,9 @@ void ucp_proto_perf_node_deref(ucp_proto_perf_node_t **perf_node_p)
     --perf_node->refcount;
     if (perf_node->refcount == 0) {
         ucp_proto_perf_node_free(perf_node);
-        *perf_node_p = NULL;
     }
+
+    *perf_node_p = NULL;
```
brminich, line 549: do we need to pass `**` as a parameter now, when we always set `*perf_node_p = NULL`?

yosefe: yes, in order to set it to NULL.
the caller can also set it to NULL, but IMO it would be safer to enforce it

yosefe: yes, it's different because perf tree nodes point to each other (unlike all other objects we have) so harder to track use-after-free issues

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -270,9 +270,77 @@ UCS_TEST_P(test_perf_node, basic)
...
+    ucp_proto_perf_node_t *child1 = ucp_proto_perf_node_new_data("child1", "");
+    ucp_proto_perf_node_t *child2 = ucp_proto_perf_node_new_data("child2", "");
+
+    ucp_proto_perf_node_add_child(parent1, child1);
+    ucp_proto_perf_node_add_child(parent1, child2);
+    EXPECT_EQ(child1, ucp_proto_perf_node_get_child(parent1, 0));
+    EXPECT_EQ(child2, ucp_proto_perf_node_get_child(parent1, 1));
+
+    ucp_proto_perf_node_t *parent2 = ucp_proto_perf_node_new_data("parent2", "");
+    ucp_proto_perf_node_t *parent2_ptr = parent2;
```
dmitrygx, line 292: minor: align by `=`

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -270,9 +270,77 @@ UCS_TEST_P(test_perf_node, basic)
...
 
     ucp_proto_perf_node_deref(&n1); /* Release n1,n2,n3 */
+    EXPECT_EQ(nullptr, n1);
+}
+
+UCS_TEST_P(test_perf_node, replace_node)
+{
+    ucp_proto_perf_node_t *parent1 = ucp_proto_perf_node_new_data("parent1", "");
+    ucp_proto_perf_node_t *child1 = ucp_proto_perf_node_new_data("child1", "");
+    ucp_proto_perf_node_t *child2 = ucp_proto_perf_node_new_data("child2", "");
```
dmitrygx, line 284: minor: align by `=`

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -270,9 +270,77 @@ UCS_TEST_P(test_perf_node, basic)
...
+        ucp_proto_perf_node_t *n1 = NULL;
+        ucp_proto_perf_node_t *n2 = NULL;
+        ucp_proto_perf_node_replace(&n1, &n2);
+    }
+
+    /* Replacing a node by NULL should release the node (without leaks) */
+    {
+        ucp_proto_perf_node_t *parent1 = ucp_proto_perf_node_new_data("parent1", "");
+        ucp_proto_perf_node_t *child1 = ucp_proto_perf_node_new_data("child1", "");
+        ucp_proto_perf_node_t *child2 = ucp_proto_perf_node_new_data("child2", "");
```
dmitrygx, line 326: minor: align by `=`

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -270,9 +270,77 @@ UCS_TEST_P(test_perf_node, basic)
...
+
+        ucp_proto_perf_node_t *null_node = NULL;
+        ucp_proto_perf_node_replace(&parent1, &null_node);
+        EXPECT_EQ(nullptr, parent1);
+    }
+
+    /* Replacing a NULL should swap variables */
+    {
+        ucp_proto_perf_node_t *node = ucp_proto_perf_node_new_data("node", "");
+        ucp_proto_perf_node_t *null_node = NULL;
```
dmitrygx, line 338: minor: align by `=`

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -270,9 +270,77 @@ UCS_TEST_P(test_perf_node, basic)
...
+    ucp_proto_perf_node_deref(&child2);
+}
+
+
+UCS_TEST_P(test_perf_node, replace_null)
+{
+    /* Replacing NULL with NULL is a no-op */
+    {
+        ucp_proto_perf_node_t *n1 = NULL;
+        ucp_proto_perf_node_t *n2 = NULL;
```
rakhmets, line 318: I think we can use `nullptr` here (and below: line 330 & 338) as well as it already done in `EXPECT_EQ`.

rakhmets (dismissed):

rakhmets (dismissed):

PR #8369, opened 2022-07-08
UCT/IB: Report correct BW for RoCE LAGs

brminich:
## What
Fixes BW reporting for RoCE LAGs

## Why ?
For UCT users to get correct BW characteristics for RoCE LAG tls

### DIFF
--- a/src/uct/ib/base/ib_iface.c
+++ b/src/uct/ib/base/ib_iface.c
@@ -1036,20 +1036,22 @@ void uct_ib_verbs_destroy_cq(uct_ib_iface_t *iface, uct_ib_dir_t dir)
     }
 }
 
-static void uct_ib_iface_set_num_paths(uct_ib_iface_t *iface,
-                                       const uct_ib_iface_config_t *config)
+static unsigned uct_ib_iface_roce_lag_level(uct_ib_iface_t *iface)
 {
     uct_ib_device_t *dev = uct_ib_iface_device(iface);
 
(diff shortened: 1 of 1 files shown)


yosefe (approved):

src/uct/ib/base/ib_iface.c
```diff
@@ -1594,8 +1597,10 @@ ucs_status_t uct_ib_iface_query(uct_ib_iface_t *iface, size_t xport_hdr_len,
     iface_attr->latency.c += numa_latency;
     iface_attr->latency.m  = 0;
 
-    /* Wire speed calculation: Width * SignalRate * Encoding */
-    wire_speed = (width * signal_rate * encoding) / 8.0;
+    /* Wire speed calculation: Width * SignalRate * Encoding * Num_paths */
+    roce_lag_level = uct_ib_iface_is_roce(iface) ?
```
yosefe, line 1601: minor: rename "roce_lag_level" to "lag_level"

brminich: isn't lag only ethernet specific? i'd keep `roce` part then

yosefe: it kinda weird that this var is used also for ib (set to 1) but it's relevant only for roce
maybe it should be num_paths?

brminich: ok

yosefe (approved):

PR #8375, opened 2022-07-10
Fix compile errors with GCC 13 snapshot

bertwesarg:
## What
fixes compile errors with GCC C13 snapshot.


shamisp:
@bertwesarg thanks for the patch. In order to accept the patch we need signed CLA https://openucx.org/license/. thanks !

yosefe (approved):

bertwesarg:
solved with 8d6032ec864190c9f079d96e731c5004a975e153

PR #8351, opened 2022-07-05
UCT/IB/DC: Fix handling of return value of checking FC

dmitrygx:
## What

Fix handling of return value of checking FC.

## Why ?

The if condition should be reverted.

## How ?

Replace `((status == UCS_OK) || (status == UCS_ERR_NO_RESOURCE))` by `((status != UCS_OK) && (status != UCS_ERR_NO_RESOURCE))`.

PR #8378, opened 2022-07-11
UCT/IB/DC: Resend FC_HARD_REQ instead of scheduling EP on waitq [v1.13.x]

dmitrygx:
## What

Resend FC_HARD_REQ instead of scheduling EP on waitq.

## Why ?

To fix "timeout waiting for replies" errors seen on DC when running io-demo on large scale.

## How ?

Include commits from https://github.com/openucx/ucx/pull/8348 and https://github.com/openucx/ucx/pull/8351.
No merge conflicts.


yosefe:
depends on #8381

yosefe (approved):

PR #8380, opened 2022-07-12
UCP/WIREUP/SELECT: Update wireup lane bandwidth calculation

SeyedMir:
## What
Update lane bandwidth calculation in wireup.

## Why ?
Currently, we use the minimum of local and remote iface bandwidths to calculate lane scores. 
This can result in the same score for all lanes if the local (remote) ifaces have a lower bandwidth
than all remote (local) ifaces. This can lead to undesirable lane selections because the NIC-device
affinity on the remote (local) side is ignored.

## How ?
Add weighted sum of the local and remote iface bandwidths when calculating lane bandwidth for lane scores.


SeyedMir:
@yosefe @Akshay-Venkatesh
I'm using a simple weighted sum, but we can think of other combinations too.
I think we also need similar changes for the latency function.

SeyedMir:
@yosefe

yosefe (approved):

PR #8385, opened 2022-07-14
UCP/AM: refactor ucp_request_t::send::msg_proto::am

evgeny-leksikov:
## What
refactor ucp_request_t::send::msg_proto::am

## Why ?
this is a preparation to re-initialization of protocol related part of request
from pending queue in case when ep is reconfigured during wireup.

## How?
replace union of user header and reg_desc with structure

### DIFF
--- a/src/ucp/am/eager.inl
+++ b/src/ucp/am/eager.inl
@@ -59,6 +59,7 @@ ucp_am_eager_zcopy_pack_user_header(ucp_request_t *req)
 {
     ucp_mem_desc_t *reg_desc;
 
+    /* reg_desc could already be allocated for a replayed request */
     if (req->send.msg_proto.am.header.reg_desc != NULL) {
         return UCS_OK;
     }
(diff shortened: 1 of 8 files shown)


src/ucp/core/ucp_request.h
```diff
@@ -183,15 +183,16 @@ struct ucp_request {
                         ucp_tag_t tag;
 
                         struct {
-                            union {
+                            struct {
                                 /* Can be union, because once header is packed to
                                  * reg_desc, it is not accessed anymore. */
```
evgeny-leksikov, line 188: Not relevant comment should be deleted

src/ucp/core/ucp_request.h
```diff
@@ -183,15 +183,14 @@ struct ucp_request {
                         ucp_tag_t tag;
 
                         struct {
-                            union {
-                                /* Can be union, because once header is packed to
-                                 * reg_desc, it is not accessed anymore. */
-                                void           *header;
+                            struct {
+                                void           *usr_ptr;
```
ofirfarjun7, line 187: Maybe it's better to use "user_ptr" instead of the "usr_ptr" so it will be more clear.
In my latest PR I was advised to do so when I used "usr" instead of "user"

src/ucp/am/eager.inl
```diff
@@ -59,26 +59,29 @@ ucp_am_eager_zcopy_pack_user_header(ucp_request_t *req)
 {
     ucp_mem_desc_t *reg_desc;
 
+    if (req->send.msg_proto.am.header.reg_desc != NULL) {
+        return UCS_OK;
```
yosefe, line 63: add comment
`/* reg_desc could already be allocated for a replayed request */`

yosefe (approved):

PR #8389, opened 2022-07-17
UCP/RMA/GTEST: Add support of user's memh for RMA operations

dmitrygx:
## What

Add support of user's memh for RMA operations.

## Why ?

Passing a pre-allocated memory handle could be used by some users to not rely on RCACHE or for XGVMI implementation in the future which always requires passing a memory handle to operations.

## How ?

1. Use `ucp_send_request_set_user_memh()` in `ucp_rma_request_init()` to set user's memory handle.
2. Introduce new variant in RMA gtests and update do_put/do_get send functions to accept `memh` to test passing user' memory handle.

### DIFF
--- a/src/ucp/rma/rma_send.c
+++ b/src/ucp/rma/rma_send.c
@@ -123,13 +123,18 @@ static void ucp_rma_request_zcopy_completion(uct_completion_t *self)
 static UCS_F_ALWAYS_INLINE ucs_status_t
 ucp_rma_request_init(ucp_request_t *req, ucp_ep_h ep, const void *buffer,
                      size_t length, uint64_t remote_addr, ucp_rkey_h rkey,
-                     uct_pending_callback_t cb, size_t zcopy_thresh)
+                     uct_pending_callback_t cb, size_t zcopy_thresh,
+                     const ucp_request_param_t *param)
 {
+    ucp_context_h context = ep->worker->context;
+    ucs_status_t status;
+
     req->flags                = 0;
     req->send.ep              = ep;
     req->send.buffer          = (void*)buffer;
     req->send.datatype        = ucp_dt_make_contig(1);
-    req->send.mem_type        = UCS_MEMORY_TYPE_HOST;
+    req->send.mem_type        = ucp_request_get_memory_type(context, buffer,
+                                                            length, param);
     req->send.length          = length;
     req->send.rma.remote_addr = remote_addr;
     req->send.rma.rkey        = rkey;
@@ -148,6 +153,13 @@ ucp_rma_request_init(ucp_request_t *req, ucp_ep_h ep, const void *buffer,
         return UCS_OK;
     }
 
+    status = ucp_send_request_set_user_memh(req,
+                                            ucp_ep_config(ep)->key.rma_md_map,
+                                            param);
+    if (status != UCS_OK) {
+        return status;
+    }
+
     return ucp_request_send_buffer_reg_lane(req, req->send.lane, 0);
 }
 
@@ -164,7 +176,7 @@ ucp_rma_nonblocking(ucp_ep_h ep, const void *buffer, size_t length,
                                 {return UCS_STATUS_PTR(UCS_ERR_NO_MEMORY);});
 
     status = ucp_rma_request_init(req, ep, buffer, length, remote_addr, rkey,
-                                  progress_cb, zcopy_thresh);
+                                  progress_cb, zcopy_thresh, param);
     if (ucs_unlikely(status != UCS_OK)) {
         return UCS_STATUS_PTR(status);
     }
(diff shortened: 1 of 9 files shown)


petro-rudenko (dismissed):

src/ucp/rma/rma_send.c
```diff
@@ -148,6 +153,13 @@ ucp_rma_request_init(ucp_request_t *req, ucp_ep_h ep, const void *buffer,
         return UCS_OK;
     }
 
+    status = ucp_send_request_set_user_memh(req,
+                                            ucp_ep_config(ep)->key.rma_md_map,
```
yosefe, line 157: configuration key should contain only direct result of wireup
so rma_md_map should be in ucp_ep_config_t - not the key - and set during ucp_ep_config_init

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -25,6 +25,8 @@ class test_ucp_rma : public test_ucp_memheap {
         add_variant_with_value(variants, UCP_FEATURE_RMA, FLUSH_EP, "flush_ep");
         add_variant_with_value(variants, UCP_FEATURE_RMA,
                                FLUSH_EP | ENABLE_PROTO, "flush_ep_proto");
+        add_variant_with_value(variants, UCP_FEATURE_RMA, USER_MEMH,
```
yosefe, line 28: by how much it increased test time?

test/gtest/ucp/test_ucp_memheap.cc
```diff
@@ -23,12 +23,12 @@ void test_ucp_memheap::test_xfer(send_func_t send_func, size_t size,
...
                                  ucs_memory_type_t target_mem_type,
                                  unsigned mem_map_flags,
-                                 bool is_ep_flush, void *arg)
+                                 bool is_ep_flush, bool user_memh, void *arg)
 {
     ucp_mem_map_params_t params;
     ucs_status_t status;
     ptrdiff_t padding;
-    ucp_mem_h memh;
+    ucp_mem_h recv_memh, send_memh = NULL;
```
yosefe, line 31: recv_memh -> memheap_memh

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -53,41 +55,41 @@ class test_ucp_rma : public test_ucp_memheap {
...
     void put_nbi_iov(size_t size, void *target_ptr, ucp_rkey_h rkey,
-                     void *expected_data, void *arg) {
+                     void *expected_data, ucp_mem_h memh, void *arg) {
         do_nbi_iov(&test_ucp_rma::do_put_iov, size, target_ptr, rkey,
                    expected_data, arg);
     }
 
     void get_b(size_t size, void *target_ptr, ucp_rkey_h rkey,
-               void *expected_data, void *arg) {
+               void *expected_data, ucp_mem_h memh, void *arg) {
```
yosefe, line 78: i think memh should be after size (local side parameters are first)

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -145,13 +153,25 @@ class test_ucp_rma : public test_ucp_memheap {
         iov[iov_count - 1].length += size % iov_count;
     }
 
+    void request_param_set_memh(ucp_request_param_t *param, ucp_mem_h memh)
```
yosefe, line 156: maybe move param.op_attr_mask = 0; into this function
and rename it to request_param_init()

src/ucp/rma/rma_send.c
dmitrygx: I followed the same idea as we have for `rma_bw_md_map`.
Shall I move both?

yosefe: ok, let's keep as-is, will probably be able to remove it soon anyway after using ucp rcache for proto v1 (since then md_map will not be needed for ucp_send_request_set_user_memh )

test/gtest/ucp/test_ucp_rma.cc
yosefe: can change the order in the callback too..

dmitrygx: done, but requires updating many places

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -177,9 +185,11 @@ class test_ucp_rma : public test_ucp_memheap {
                            (uintptr_t)target_ptr, rkey, &param);
     }
 
-    ucs_status_ptr_t do_put_iov(size_t size, void *target_ptr, ucp_rkey_h rkey,
-                                void *expected_data, ucp_request_param_t *param,
-                                ucp_dt_iov_t *iov, size_t iov_count, void *arg) {
+    ucs_status_ptr_t do_put_iov(size_t size,  void *expected_data,
```
yosefe, line 188: extra space before void

dmitrygx: fixed

yosefe (approved):

yosefe (approved):

PR #7811, opened 2021-12-20
UCP/MM: Add UCP rcache

Artemy-Mellanox:
## Why

To improve performance of zcopy protocols.

## How

In amortized case UCP fast-path flow will bypass UCT registration logic.

PR #8392, opened 2022-07-20
UCP/CORE: Fixed memory handle registration.

rakhmets:
## What
Fixed memory handle registration.

## Why ?
`UCT_MD_MEM_FLAG_HIDE_ERRORS` flag should prevent from failing memory handler registration process in case of memory registration fail on a memory domain.
`ucp_memh_get_slow` should perform memory handler registration process with `UCT_MD_MEM_FLAG_HIDE_ERRORS` flag.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -11,6 +11,7 @@
 #endif
 
 #include "ucp_context.h"
+#include "ucp_mm.inl"
 #include "ucp_request.h"
 
 #include <ucs/config/parser.h>
@@ -1398,6 +1399,25 @@ out:
     return status;
 }
 
+static ucs_status_t ucp_context_reg_md_map_alloc(ucp_context_h context)
+{
+    ucs_status_t status;
+    ucp_mem_h memh;
+
+    /* Allocate dummy 1-byte buffer to get the expected md_map */
+    status = ucp_memh_alloc(context, NULL, 1, UCS_MEMORY_TYPE_HOST,
+                            UCT_MD_MEM_ACCESS_ALL | UCT_MD_MEM_FLAG_HIDE_ERRORS,
+                            "get_alloc_md_map", &memh);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    context->reg_md_map[UCS_MEMORY_TYPE_HOST] &= memh->md_map;
+    ucp_memh_put(context, memh, 1);
+
+    return UCS_OK;
+}
+
 static ucs_status_t ucp_fill_resources(ucp_context_h context,
                                        const ucp_config_t *config)
 {
@@ -1415,8 +1435,6 @@ static ucs_status_t ucp_fill_resources(ucp_context_h context,
     context->num_cmpts                = 0;
     context->tl_mds                   = NULL;
     context->num_mds                  = 0;
-    context->alloc_md_map_initialized = 0;
-    context->alloc_md_map             = 0;
     context->tl_rscs                  = NULL;
     context->num_tls                  = 0;
     context->mem_type_mask            = 0;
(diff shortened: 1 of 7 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -355,6 +355,10 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
                     context->tl_mds[md_index].rsc.md_name,
                     ucs_status_string(status));
 
+            if (uct_flags & UCT_MD_MEM_FLAG_HIDE_ERRORS) {
```
yosefe, line 358: can we unite with the condition at line 350?

yosefe (dismissed):

yosefe:

src/ucp/core/ucp_mm.c
```diff
@@ -428,7 +437,7 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
     status         = ucp_memh_register(context, memh,
                                        ~memh->md_map & reg_md_map,
                                        reg_address, reg_length,
-                                       uct_flags);
+                                       uct_flags | UCT_MD_MEM_FLAG_HIDE_ERRORS);
```
Artemy-Mellanox, line 440: Can you explain what problem does it solve?
AFAIK ucp_memh_get() should hide errors only if uct_flags contains HIDE_ERRORS flag and not always.
At least that was ucp_datatype_iter_mem_reg_internal was doing.

rakhmets: This pull request fixes failure while calling `ucp_mem_map` with `UCP_MEM_MAP_ALLOCATE` parameter on some systems. E.g. simple command `ucx_info -m 4096 -u t` fails on Ubuntu 20.04 because of failing of memory registration on KNEM.
Previously (before https://github.com/openucx/ucx/pull/7811) we always call `ucp_mem_rereg_mds` with `UCT_MD_MEM_FLAG_HIDE_ERRORS`. That is why there was no error.

Artemy-Mellanox: Adding this flag here is too broad, new proto flow shouldn't hide errors.
Need to add it in `ucp_mm_get_alloc_md_map()`, `ucp_mem_map()`, `ucp_rndv_frag_malloc_mpools()` and `ucp_mpool_malloc()`.
I would remove `alloc_md_map`/`_initialized`, make `ucp_mm_get_alloc_md_map()` code unconditional part of context initialization and `&` it's outcome with `reg_md_map[UCS_MEMORY_TYPE_HOST]`
@yosefe WDYT?

Artemy-Mellanox (approved):

src/ucp/core/ucp_context.c
```diff
@@ -1939,6 +1957,14 @@ ucs_status_t ucp_init_version(unsigned api_major_version, unsigned api_minor_ver
         context->rcache = NULL;
     }
 
+    /* Update registration memory domain map for host memory type taking into
+     * account result of ucp_memh_alloc. */
+    status = ucp_context_reg_md_map_alloc(context);
```
yosefe, line 1962: maybe move this inside ucp_fill_resources, after the call to ucp_add_component_resources, which initializes reg_md_map.

yosefe (dismissed):

Artemy-Mellanox (approved):

src/ucp/wireup/select.c
```diff
@@ -306,6 +306,28 @@ ucp_wireup_init_select_info(double score, unsigned addr_index,
...
+/**
+ * Get bitmap of memory types that Memory Domain can be registered with taking
+ * into account context's maps of Memory Domains that provide registration for
+ * given memory type.
+ */
+static uint64_t ucp_wireup_select_reg_mem_types(uct_md_attr_t *md_attr,
+                                                ucp_context_h context,
+                                                ucp_rsc_index_t rsc_index)
+{
+    uint64_t reg_mem_types  = md_attr->cap.reg_mem_types;
```
yosefe, line 318: do we really need md_attr here?
maybe start with 0 and add bits from reg_md_map[mem_type] (instead of removing them)?

yosefe (dismissed):

yosefe (approved):

PR #8395, opened 2022-07-21
UCP/MM: Make ucp_mem_map return unique memh

Artemy-Mellanox:
## Why ?
This way ucp_mem_unmap will match corresponding ucp_mem_map and will be able to unroll operations that should be unrolled (i.e. allocations, registrations of MDs which shouldn't be cached)

## How ?
ucp_mem_map() won't return memh from rcache but will wrap it in unique temporary memh.

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -259,7 +259,7 @@ out:
     return status;
 }
 
-static inline unsigned
+static unsigned
 ucp_mem_map_params2uct_flags(const ucp_mem_map_params_t *params)
 {
     unsigned flags = 0;
@@ -306,8 +306,9 @@ static void ucp_memh_dereg(ucp_context_h context, ucp_mem_h memh,
     }
 }
 
-void ucp_memh_unmap(ucp_context_h context, ucp_mem_h memh, ucp_md_map_t md_map)
+void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
 {
+    ucp_md_map_t md_map = memh->md_map;
     uct_allocated_memory_t mem;
     ucs_status_t status;
 
@@ -322,7 +323,15 @@ void ucp_memh_unmap(ucp_context_h context, ucp_mem_h memh, ucp_md_map_t md_map)
         md_map  &= ~UCS_BIT(memh->alloc_md_index);
     }
 
-    ucp_memh_dereg(context, memh, md_map);
+    /* Have a parent memory handle from rcache */
+    if ((memh->parent != NULL) && (memh->parent != memh)) {
+        /* TODO remove invalidate, replace with ucp_memh_dereg weak MDs */
+        ucs_rcache_region_invalidate(context->rcache, &memh->parent->super,
+                (ucs_rcache_invalidate_comp_func_t)ucs_empty_function, NULL);
+        ucp_memh_put(context, memh->parent);
+    } else {
+        ucp_memh_dereg(context, memh, md_map);
+    }
 
     /* If the memory was also allocated, release it */
     if (memh->alloc_method != UCT_ALLOC_METHOD_LAST) {
@@ -365,13 +374,6 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
             ucp_memh_register_log_fail(UCS_LOG_LEVEL_ERROR, address, length,
                                        md_index, context, status);
             ucp_memh_dereg(context, memh, md_map_registered);
-
-            if (context->rcache != NULL) {
-                ucs_rcache_region_put_unsafe(context->rcache, &memh->super);
-            } else {
-                ucs_free(memh);
-            }
-
             return status;
         }
 
@@ -386,17 +388,133 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
     return UCS_OK;
 }
 
+static size_t ucp_memh_size(ucp_context_h context)
+{
+    return sizeof(ucp_mem_t) + (sizeof(uct_mem_h) * context->num_mds);
+}
+
+static void ucp_memh_set(ucp_mem_h memh, void* address, size_t length,
+                         ucs_memory_type_t mem_type, uct_alloc_method_t method)
+{
+    memh->super.super.start = (uintptr_t)address;
+    memh->super.super.end   = (uintptr_t)address + length;
+    memh->mem_type          = mem_type;
+    memh->alloc_method      = method;
+    memh->alloc_md_index    = UCP_NULL_RESOURCE;
+}
+
+static ucs_status_t
+ucp_memh_create(ucp_context_h context, void *address, size_t length,
+                ucs_memory_type_t mem_type, ucp_mem_h *memh_p)
+{
+    ucp_mem_h memh;
+
+    memh = ucs_calloc(1, ucp_memh_size(context), "ucp_rcache");
+    if (memh == NULL) {
+        return UCS_ERR_NO_MEMORY;
+    }
+
+    ucp_memh_set(memh, address, length, mem_type, UCT_ALLOC_METHOD_LAST);
+
+    if (context->rcache == NULL) {
+        /* Point to self */
+        memh->parent = memh;
+    }
+
+    *memh_p = memh;
+    return UCS_OK;
+}
+
+static ucs_status_t
+ucp_memh_rcache_get(ucp_context_h context, void *address, size_t length,
+                    ucs_memory_type_t mem_type, ucp_mem_h *memh_p)
+{
+    ucp_mem_attr_t attr = {
+        .mem_type = mem_type
+    };
+    ucs_rcache_region_t *rregion;
+    ucs_status_t status;
+
+    status = ucs_rcache_get(context->rcache, address, length,
+                            PROT_READ | PROT_WRITE, &attr, &rregion);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    *memh_p = ucs_derived_of(rregion, ucp_mem_t);
+    return UCS_OK;
+}
+
+static void
+ucp_memh_init_from_mem(ucp_context_h context, ucp_mem_h memh,
+                       uct_allocated_memory_t *mem)
+{
+    ucp_md_index_t md_index;
+
+    ucp_memh_set(memh, mem->address, mem->length, mem->mem_type, mem->method);
+
+    if (mem->method != UCT_ALLOC_METHOD_MD) {
+        return;
+    }
+
+    for (md_index = 0; md_index < context->num_mds; md_index++) {
+        if (mem->md != context->tl_mds[md_index].md) {
+            continue;
+        }
+
+        memh->alloc_md_index = md_index;
+        memh->uct[md_index]  = mem->memh;
+        memh->md_map        |= UCS_BIT(md_index);
+        ucs_trace("allocated address %p length %zu on md[%d]=%s %p",
+                  mem->address, mem->length, md_index,
+                  context->tl_mds[md_index].rsc.md_name, memh->uct[md_index]);
+        break;
+    }
+
+    ucs_assert(memh->alloc_md_index != UCP_NULL_RESOURCE);
+}
+
+static ucs_status_t
+ucp_memh_init_uct_reg(ucp_context_h context, ucp_mem_h memh,
+                      ucp_md_map_t reg_md_map, unsigned uct_flags)
+{
+    void *address = ucp_memh_address(memh);
+    size_t length = ucp_memh_length(memh);
+    ucp_md_index_t md_index;
+    ucs_status_t status;
+
+    if (context->rcache == NULL) {
+        status = ucp_memh_register(context, memh, reg_md_map, address, length,
+                                   uct_flags);
+        if (status != UCS_OK) {
+            return status;
+        }
+    } else {
+        status = ucp_memh_get(context, address, length, memh->mem_type,
+                              reg_md_map, uct_flags, &memh->parent);
+        if (status != UCS_OK) {
+            return status;
+        }
+
+        ucs_for_each_bit(md_index, reg_md_map) {
+            memh->uct[md_index] = memh->parent->uct[md_index];
+            memh->md_map       |= UCS_BIT(md_index);
+        }
+    }
+
+    return UCS_OK;
+}
+
 ucs_status_t
 ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
                   ucs_memory_type_t mem_type, ucp_md_map_t reg_md_map,
                   unsigned uct_flags, ucp_mem_h *memh_p)
 {
-    ucs_rcache_region_t *rregion;
+    ucp_mem_h memh = NULL; /* To suppress compiler warning */
     void *reg_address;
     size_t reg_length;
     ucs_status_t status;
     ucs_memory_info_t mem_info;
-    ucp_mem_h memh;
 
     if (context->config.ext.reg_whole_alloc_bitmap & UCS_BIT(mem_type)) {
         ucp_memory_detect_internal(context, address, length, &mem_info);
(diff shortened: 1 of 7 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -429,16 +444,65 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
+    ucp_md_index_t md_index;
+    ucs_status_t status;
+    ucp_mem_h memh;
+
+    status = ucp_memh_get_nocache(context, address, length, &memh);
+    if (status != UCS_OK) {
+        goto err;
+    }
+
+    memh->mem_type = mem_type;
```
dmitrygx, line 484:
```suggestion
    status = ucp_memh_get_nocache(context, address, length, mem_type, &memh);
    if (status != UCS_OK) {
        goto err;
    }
```

src/ucp/core/ucp_mm.c
```diff
@@ -429,16 +444,65 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
+    status         = ucp_memh_get(context, address, length, mem_type,
+                                  reg_md_map, uct_flags, &memh->parent);
+    if (status != UCS_OK) {
+        goto err_free_memh;
+    }
+
+    ucs_for_each_bit(md_index, memh->parent->md_map) {
+        memh->uct[md_index] = memh->parent->uct[md_index];
+        memh->md_map |= UCS_BIT(md_index);
```
dmitrygx, line 493:
```suggestion
        memh->uct[md_index] = memh->parent->uct[md_index];
        memh->md_map       |= UCS_BIT(md_index);
```

src/ucp/core/ucp_mm.h
```diff
@@ -31,7 +31,9 @@ typedef struct ucp_mem {
     ucs_memory_type_t   mem_type;       /* Type of allocated or registered memory */
     ucp_md_index_t      alloc_md_index; /* Index of MD used to allocated the memory */
     ucp_md_map_t        md_map;         /* Which MDs have valid memory handles */
-    uint32_t            map_count;      /* ucp_mem_map/unmap referrence count */
+    struct ucp_mem      *parent;        /* NULL if entry should be returned to rcache
+                                           pointer to self if rcache disabled
+                                           pointer to rcache memh if entry is weak memh */
```
dmitrygx, line 36:
```suggestion
    struct ucp_mem      *parent;        /* - NULL if entry should be returned to rcache
                                           - pointer to self if rcache disabled
                                           - pointer to rcache memh if entry is a weak memh */
```

src/ucp/core/ucp_mm.c
```diff
@@ -400,18 +423,10 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
         }
-
-        memh->super.super.start = (uintptr_t)reg_address;
-        memh->super.super.end   = (uintptr_t)reg_address + reg_length;
-        memh->alloc_md_index    = UCP_NULL_RESOURCE;
-        memh->alloc_method      = UCT_ALLOC_METHOD_LAST;
-        memh->map_count         = 0;
     } else {
         status = ucs_rcache_get_unsafe(context->rcache, reg_address, reg_length,
                                        PROT_READ | PROT_WRITE, NULL, &rregion);
```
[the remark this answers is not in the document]
Artemy-Mellanox: where `mem_type` set in  `ucp_memh_create_rcache` flow?

src/ucp/core/ucp_mm.c
```diff
@@ -423,27 +447,26 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
-    status         = ucp_memh_register(context, memh,
-                                       ~memh->md_map & reg_md_map,
-                                       reg_address, reg_length,
-                                       uct_flags);
+    /* Reinitialize address and length, because they could be greater in case
+     * a region which includes the searched region was found in the rcache */
+    reg_address = ucp_memh_address(memh);
+    reg_length  = ucp_memh_length(memh);
+
+    ucs_assert(memh->mem_type == mem_type);
```
dmitrygx, line 466: sorry, I guess it is not needed anymore, since we pass mem_type through rcache attributes

src/ucp/core/ucp_mm.c
```diff
@@ -393,24 +397,44 @@ ucs_status_t ucp_memh_get_nocache(ucp_context_h context, void *address,
...
     return UCS_OK;
 }
 
+static ucs_status_t
+ucp_memh_create_rcache(ucp_context_h context, void *address, size_t length,
+                       ucs_memory_type_t mem_type, ucp_mem_h *memh_p)
+{
+    ucp_mem_attr_t attr = {
+        .mem_type = mem_type
+    };
```
dmitrygx, line 414: 👍

dmitrygx (approved):

src/ucp/core/ucp_mm.c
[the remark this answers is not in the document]
dmitrygx: yes, missed it

dmitrygx (dismissed):

dmitrygx (approved):

src/ucs/memory/rcache.inl
```diff
@@ -74,8 +74,7 @@ ucs_rcache_get_unsafe(ucs_rcache_t *rcache, void *address, size_t length,
         }
     }
 
-    return UCS_PROFILE_CALL(ucs_rcache_create_region, rcache, address, length,
-                            prot, arg, region_p);
+    return UCS_ERR_NO_ELEM;
```
yosefe, line 77: 1. maybe rename this function to `ucs_rcache_`<b>`lookup`</b>`_unsafe`
since `ucs_rcache_`<b>`get`</b> still creates new region

2. return region ptr (which could be NULL if not found)

3. because of this change, can use "if .. return" pattern instead of nested if

src/ucp/core/ucp_mm.inl
```diff
@@ -40,7 +40,9 @@ ucp_memh_get(ucp_context_h context, void *address, size_t length,
         UCP_THREAD_CS_ENTER(&context->mt_lock);
         status = ucs_rcache_get_unsafe(context->rcache, address, length,
                                        PROT_READ | PROT_WRITE, NULL, &rregion);
-        if (status != UCS_OK) {
+        if (status == UCS_ERR_NO_ELEM) {
+            goto not_found;
+        } else if (status != UCS_OK) {
```
yosefe, line 45: can it really be another error here?

src/ucp/core/ucp_mm.inl
```diff
@@ -68,28 +71,25 @@ out_unlock:
...
+              memh, ucp_memh_address(memh), ucp_memh_length(memh),
+              memh->md_map);
 
     if (ucs_unlikely(ucp_memh_is_zero_length(memh))) {
         return;
     }
 
-    if (ucs_unlikely(context->rcache == NULL)) {
-        ucp_memh_unmap(context, memh, memh->md_map);
+    if (ucs_unlikely(memh->parent != NULL)) {
```
yosefe, line 84: add comment `/* User memory handle, or rcache was disabled */`

src/ucp/core/ucp_mm.c
```diff
@@ -240,7 +240,7 @@ ucp_mem_do_alloc(ucp_context_h context, void *address, size_t length,
                                  UCT_MEM_ALLOC_PARAM_FIELD_MDS      |
                                  UCT_MEM_ALLOC_PARAM_FIELD_NAME;
         params.flags           = uct_flags;
-        params.name            = name;
+        params.name            = "ucp_rcache";
```
yosefe, line 243: why not passing name from user?

src/ucp/core/ucp_mm.c
```diff
@@ -322,7 +302,16 @@ void ucp_memh_unmap(ucp_context_h context, ucp_mem_h memh, ucp_md_map_t md_map)
         md_map  &= ~UCS_BIT(memh->alloc_md_index);
     }
 
-    ucp_memh_dereg(context, memh, md_map);
+    if ((memh->parent != NULL) && (memh->parent != memh)) {
+        /* TODO remove invalidate, replace with ucp_memh_dereg weak MDs */
+        if (context->rcache != NULL) {
```
yosefe, line 307: isn't the check in line 305 already implies rcache is on (since we have "real" parent)?
so can convert to assert

src/ucp/core/ucp_mm.c
```diff
@@ -322,7 +302,16 @@ void ucp_memh_unmap(ucp_context_h context, ucp_mem_h memh, ucp_md_map_t md_map)
         md_map  &= ~UCS_BIT(memh->alloc_md_index);
     }
 
-    ucp_memh_dereg(context, memh, md_map);
+    if ((memh->parent != NULL) && (memh->parent != memh)) {
+        /* TODO remove invalidate, replace with ucp_memh_dereg weak MDs */
```
yosefe, line 306: add to the comment:
/* Have a parent memory handle from rcache */

src/ucp/core/ucp_mm.c
```diff
@@ -400,99 +425,152 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
+              reg_length, ucp_memh_length(memh), reg_md_map, memh->md_map);
+    *memh_p = memh;
+
 out:
     UCP_THREAD_CS_EXIT(&context->mt_lock);
     return status;
+
+err_free_memh:
+    if (context->rcache == NULL) {
+        ucs_free(memh);
```
yosefe, line 464: maybe add helper function ucp_memh_free that asserts md_map==0 and then does ucs_free(memh)
seems it would be useful to detect leaked registrations especially in error flow

src/ucp/core/ucp_mm.c
```diff
@@ -377,17 +359,60 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
...
 }
 
+static ucs_status_t
+ucp_memh_create_no_rcache(ucp_context_h context, void *address,
+                          size_t length, ucs_memory_type_t mem_type,
+                          ucp_mem_h *memh_p)
+{
+    ucp_mem_h memh;
+
+    memh = ucs_calloc(1, sizeof(*memh) +
```
yosefe, line 369: maybe use helper var for memh struct size

src/ucp/core/ucp_mm.c
```diff
@@ -400,99 +425,152 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
+static void
+ucp_memh_init_from_mem(ucp_context_h context, ucp_mem_h memh,
+                       uct_allocated_memory_t *mem)
+{
+    ucp_md_index_t md_index;
+
+    memh->super.super.start = (uintptr_t)mem->address;
+    memh->super.super.end   = (uintptr_t)mem->address + mem->length;
+    memh->mem_type          = mem->mem_type;
+    memh->alloc_method      = mem->method;
```
yosefe, line 481: maybe could reuse memh init code with ucp_memh_create_no_rcache and

src/ucp/core/ucp_mm.c
```diff
@@ -400,99 +425,152 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
+    if (mem->method != UCT_ALLOC_METHOD_MD) {
+        return;
+    }
+
+    for (md_index = 0; md_index < context->num_mds; md_index++) {
+        if (mem->md != context->tl_mds[md_index].md) {
+            continue;
+        }
+
+        memh->alloc_md_index = md_index;
```
yosefe, line 492: maybe set this to default value after line 481?

src/ucp/core/ucp_mm.c
```diff
@@ -400,99 +425,152 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
+    }
+
+    ucs_assert(memh->alloc_md_index != UCP_NULL_RESOURCE);
 }
 
 static ucs_status_t
-ucp_memh_alloc(ucp_context_h context, void *address, size_t length,
-               ucs_memory_type_t memory_type, unsigned uct_flags,
-               const char *alloc_name, ucp_mem_h *memh_p)
+ucp_memh_init(ucp_context_h context, ucp_mem_h memh,
```
yosefe, line 505: can we call it something like ucp_memh_init_uct_reg?

src/ucp/core/ucp_mm.c
```diff
@@ -400,99 +425,152 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
+        }
 
-    if (mem.method == UCT_ALLOC_METHOD_MD) {
-        for (md_index = 0; md_index < context->num_mds; md_index++) {
-            if (mem.md == context->tl_mds[md_index].md) {
-                alloc_md_index = md_index;
-                reg_md_map    &= ~UCS_BIT(md_index);
-                break;
-            }
+        memh->parent = memh;
```
yosefe, line 520: add comment /* Point to self */ to emphasize this is intentional

src/ucp/core/ucp_mm.c
```diff
@@ -377,17 +359,60 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
     return UCS_OK;
 }
 
+static ucs_status_t
+ucp_memh_create_no_rcache(ucp_context_h context, void *address,
```
yosefe, line 363: maybe
ucp_memh_create
since sometimes it's used to create user_memh, even if there is an rcache (from ucp_memh_alloc for example)

src/ucp/core/ucp_mm.c
```diff
@@ -400,99 +425,152 @@ ucp_memh_get_slow(ucp_context_h context, void *address, size_t length,
...
-                mem.address, mem.length, alloc_md_index,
-                context->tl_mds[alloc_md_index].rsc.md_name,
-                memh->uct[alloc_md_index]);
+    status = ucp_mem_do_alloc(context, NULL, length, UCT_MD_MEM_ACCESS_ALL,
+                              mem_type, &mem);
+    if (status != UCS_OK) {
+        goto err_free_memh;
+    }
+
+    ucp_memh_init_from_mem(context, memh, &mem);
```
yosefe, line 557: seems like ucp_memh_init_from_mem overrides memh fields that were set by ucp_memh_create_no_rcache ?
maybe we can first call ucp_mem_do_alloc and only then initialize the memh once with the right fields?

src/ucp/core/ucp_mm.c
```diff
@@ -554,44 +637,77 @@ ucs_status_t ucp_mem_map(ucp_context_h context, const ucp_mem_map_params_t *para
...
+
+    if (params->field_mask & UCP_MEM_MAP_PARAM_FIELD_FLAGS) {
+        if (params->flags & UCP_MEM_MAP_NONBLOCK) {
+            uct_flags |= UCT_MD_MEM_FLAG_NONBLOCK;
+        }
+
+        if (params->flags & UCP_MEM_MAP_FIXED) {
+            uct_flags |= UCT_MD_MEM_FLAG_FIXED;
+        }
+    }
```
yosefe, line 669: helper function for this?

src/ucp/core/ucp_mm.c
```diff
@@ -554,44 +637,77 @@ ucs_status_t ucp_mem_map(ucp_context_h context, const ucp_mem_map_params_t *para
...
-        status = ucp_memh_get(context, address, params->length, memory_type,
-                              context->reg_md_map[memory_type],
-                              ucp_mem_map_params2uct_flags(params), memh_p);
+        status = ucp_mem_do_alloc(context, address, length, uct_flags, mem_type,
+                                  &mem);
+        if (status != UCS_OK) {
+            goto err_free_memh;
+        }
+
+        ucp_memh_init_from_mem(context, memh, &mem);
```
yosefe, line 686: maybe move ucp_memh_init_from_mem into ucp_mem_do_alloc?

src/ucp/core/ucp_mm.c
```diff
@@ -554,44 +637,77 @@ ucs_status_t ucp_mem_map(ucp_context_h context, const ucp_mem_map_params_t *para
...
+        reg_md_map &= ~memh->md_map;
```
yosefe, line 687: i would expect reg_md_map to always become 0 here

Artemy-Mellanox: Why? should be all except alloc_md

yosefe: AFAIU, ucp_mem_do_alloc should already register all of `context->reg_md_map[mem_type]`

src/ucp/core/ucp_mm.c
```diff
@@ -259,6 +259,27 @@ ucp_mem_do_alloc(ucp_context_h context, void *address, size_t length,
     return status;
 }
 
+static inline unsigned
```
yosefe, line 262: minor: can remove inline

src/ucp/core/ucp_mm.c
```diff
@@ -877,6 +885,14 @@ ucp_mem_advise(ucp_context_h context, ucp_mem_h memh,
     return status;
 }
 
+static ucs_status_t
+ucp_memh_alloc(ucp_context_h context, size_t length,
```
yosefe, line 889: maybe remove this wrapper

src/ucp/core/ucp_mm.c
yosefe: missed

yosefe (dismissed):
pls squash

yosefe (approved):

PR #8396, opened 2022-07-21
TOOLS/PERF: Simplify Socket RTE initialization

dmitrygx:
## What

 Simplify Socket RTE initialization.

## Why ?

The code becomes clearer and it helps remove one `goto` label and reduce some code duplication.

## How ?

1. Set `sockfd` from value `connfd` in case of server in loop over all IP interfaces in `setup_sock_rte_p2p` after listening `sockfd` was clsoed.
2. Print `accept` error through `err_str` as it is done for other operations.

### DIFF
--- a/src/tools/perf/perftest.c
+++ b/src/tools/perf/perftest.c
@@ -319,12 +319,12 @@ static ucs_status_t setup_sock_rte_loobkack(struct perftest_context *ctx)
 
 static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
 {
-    int optval = 1;
+    int optval  = 1;
     int sockfd = -1;
+    int connfd = -1;
     char addr_str[UCS_SOCKADDR_STRING_LEN];
     struct sockaddr_storage client_addr;
     socklen_t client_addr_len;
-    int connfd;
     struct addrinfo hints, *res, *t;
     ucs_status_t status;
     int ret;
@@ -358,22 +358,26 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
 
         if (ctx->server_addr != NULL) {
             if (connect(sockfd, t->ai_addr, t->ai_addrlen) == 0) {
+                connfd = sockfd;
+                sockfd = -1;
                 break;
             }
+
             snprintf(err_str, 64, "connect() failed: %m");
         } else {
             status = ucs_socket_setopt(sockfd, SOL_SOCKET, SO_REUSEADDR,
                                        &optval, sizeof(optval));
             if (status != UCS_OK) {
-                status = UCS_ERR_IO_ERROR;
+                snprintf(err_str, 64, "ucs_socket_setopt() failed: %s",
+                         ucs_status_string(status));
                 goto err_close_sockfd;
             }
 
-            if (bind(sockfd, t->ai_addr, t->ai_addrlen) == 0) {
+            ret = bind(sockfd, t->ai_addr, t->ai_addrlen);
+            if (ret == 0) {
                 ret = listen(sockfd, 10);
                 if (ret < 0) {
-                    ucs_error("listen() failed: %m");
-                    status = UCS_ERR_IO_ERROR;
+                    snprintf(err_str, 64, "accept() failed: %m");
                     goto err_close_sockfd;
                 }
 
@@ -384,28 +388,26 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
                 connfd          = accept(sockfd, (struct sockaddr*)&client_addr,
                                          &client_addr_len);
                 if (connfd < 0) {
-                    ucs_error("accept() failed: %m");
-                    status = UCS_ERR_IO_ERROR;
+                    snprintf(err_str, 64, "accept() failed: %m");
                     goto err_close_sockfd;
                 }
 
                 ucs_sockaddr_str((struct sockaddr*)&client_addr, addr_str,
                                  sizeof(addr_str));
                 printf("Accepted connection from %s\n", addr_str);
-                close(sockfd);
+                ucs_close_fd(&sockfd);
                 break;
             }
+
             snprintf(err_str, 64, "bind() failed: %m");
         }
-        close(sockfd);
-        sockfd = -1;
+
+        ucs_close_fd(&sockfd);
     }
 
-    if (sockfd < 0) {
-        ucs_error("%s failed. %s",
-                  (ctx->server_addr != NULL) ? "client" : "server", err_str);
+    if (connfd < 0) {
         status = UCS_ERR_IO_ERROR;
-        goto out_free_res;
+        goto err_close_sockfd;
     }
 
     if (ctx->server_addr == NULL) {
@@ -431,7 +433,7 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
 
             ret = safe_recv(connfd, ctx->params.super.msg_size_list,
                             sizeof(*ctx->params.super.msg_size_list) *
-                            ctx->params.super.msg_size_cnt,
+                                    ctx->params.super.msg_size_cnt,
                             NULL, NULL);
             if (ret) {
                 status = UCS_ERR_IO_ERROR;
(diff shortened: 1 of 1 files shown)


src/tools/perf/perftest.c
```diff
@@ -383,20 +383,21 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
                 client_addr_len = sizeof(client_addr);
                 connfd          = accept(sockfd, (struct sockaddr*)&client_addr,
                                          &client_addr_len);
-                if (connfd < 0) {
-                    ucs_error("accept() failed: %m");
-                    status = UCS_ERR_IO_ERROR;
-                    goto err_close_sockfd;
+                if (connfd >= 0) {
+                    ucs_sockaddr_str((struct sockaddr*)&client_addr, addr_str,
+                                 sizeof(addr_str));
```
yosefe, line 388: pls check clang-format - indent

dmitrygx: done

src/tools/perf/perftest.c
```diff
@@ -452,13 +451,13 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
...
-        ctx->sock_rte_group.sendfd     = sockfd;
-        ctx->sock_rte_group.recvfd     = sockfd;
-        ctx->sock_rte_group.peer       = 0;
-        ctx->sock_rte_group.is_server  = 0;
+        ctx->sock_rte_group.peer      = 0;
+        ctx->sock_rte_group.is_server = 0;
     }
 
-    ctx->sock_rte_group.size = 2;
+    ctx->sock_rte_group.sendfd = sockfd;
```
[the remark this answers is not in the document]
yosefe: I agree there should be one variable for FD. It's just that need to switch roles of "sockfd" and "connfd" because now the usage does not match variable name.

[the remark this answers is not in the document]
dmitrygx: > I agree there should be one variable for FD. It's just that need to switch roles of "sockfd" and "connfd" because now the usage does not match variable name.

@yosefe not sure that I understand what we should do to fix matching variable names and roles

src/tools/perf/perftest.c
```diff
@@ -358,50 +358,55 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
...
-
-                ucs_sockaddr_str((struct sockaddr*)&client_addr, addr_str,
-                                 sizeof(addr_str));
-                printf("Accepted connection from %s\n", addr_str);
-                close(sockfd);
-                break;
+            ret = bind(sockfd, t->ai_addr, t->ai_addrlen);
+            if (ret < 0) {
+                snprintf(err_str, 64, "bind() failed: %m");
+                goto out_close_sockfd;
```
yosefe, line 378: in previous code, only this place continued to next iteration

src/tools/perf/perftest.c
```diff
@@ -358,22 +358,26 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
...
             }
 
-            if (bind(sockfd, t->ai_addr, t->ai_addrlen) == 0) {
+            ret = bind(sockfd, t->ai_addr, t->ai_addrlen);
+            if (ret == 0) {
                 ret = listen(sockfd, 10);
                 if (ret < 0) {
-                    ucs_error("listen() failed: %m");
-                    status = UCS_ERR_IO_ERROR;
+                    snprintf(err_str, 64, "accept() failed: %m");
```
yosefe, line 380: why removed "status = UCS_ERR_IO_ERROR;"?

src/tools/perf/perftest.c
```diff
@@ -358,22 +358,26 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
...
                 break;
             }
+
             snprintf(err_str, 64, "connect() failed: %m");
         } else {
             status = ucs_socket_setopt(sockfd, SOL_SOCKET, SO_REUSEADDR,
                                        &optval, sizeof(optval));
             if (status != UCS_OK) {
-                status = UCS_ERR_IO_ERROR;
+                snprintf(err_str, 64, "ucs_socket_setopt() failed: %s",
```
yosefe, line 371: why removed "status = UCS_ERR_IO_ERROR;"?

src/tools/perf/perftest.c
```diff
@@ -384,28 +388,26 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
                 connfd          = accept(sockfd, (struct sockaddr*)&client_addr,
                                          &client_addr_len);
                 if (connfd < 0) {
-                    ucs_error("accept() failed: %m");
-                    status = UCS_ERR_IO_ERROR;
+                    snprintf(err_str, 64, "accept() failed: %m");
```
yosefe, line 391: why removed "status = UCS_ERR_IO_ERROR;"?

src/tools/perf/perftest.c
```diff
@@ -384,28 +388,26 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
...
+        ucs_close_fd(&sockfd);
     }
 
-    if (sockfd < 0) {
-        ucs_error("%s failed. %s",
-                  (ctx->server_addr != NULL) ? "client" : "server", err_str);
+    if (connfd < 0) {
         status = UCS_ERR_IO_ERROR;
-        goto out_free_res;
+        goto err_close_sockfd;
```
yosefe, line 410: should we print some error here, for example, that could not find matching address to connect to XX ?

src/tools/perf/perftest.c
```diff
@@ -431,34 +433,32 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
 
             ret = safe_recv(connfd, ctx->params.super.msg_size_list,
                             sizeof(*ctx->params.super.msg_size_list) *
-                            ctx->params.super.msg_size_cnt,
+                                    ctx->params.super.msg_size_cnt,
```
yosefe, line 436: revert

src/tools/perf/perftest.c
```diff
@@ -471,8 +471,9 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
 
 err_close_connfd:
     ucs_close_fd(&connfd);
-    goto out_free_res;
 err_close_sockfd:
+    ucs_error("%s failed. %s",
```
yosefe, line 475: seems err_str could be un initialized when jumping to err_close_connfd

src/tools/perf/perftest.c
dmitrygx: leftover from previous implementation.
fixed

src/tools/perf/perftest.c
dmitrygx: fixed

src/tools/perf/perftest.c
dmitrygx: fixed

src/tools/perf/perftest.c
dmitrygx: done

src/tools/perf/perftest.c
dmitrygx: it is initialized

src/tools/perf/perftest.c
dmitrygx: `err_str` is set to one of the following string:
- `"bind() failed: %m"`
- `"connect() failed: %m"`
- `"getaddrinfo() returned empty list"`

src/tools/perf/perftest.c
yosefe: what about lines 425, 434, 443 for example?

src/tools/perf/perftest.c
```diff
@@ -384,28 +390,27 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
...
+        ucs_close_fd(&sockfd);
     }
 
-    if (sockfd < 0) {
-        ucs_error("%s failed. %s",
-                  (ctx->server_addr != NULL) ? "client" : "server", err_str);
+    if (connfd < 0) {
         status = UCS_ERR_IO_ERROR;
-        goto out_free_res;
+        goto err_close_sockfd;
```
yosefe, line 413: seems we can get here when "res" list is empty, for example, in which case err_str will not be initialized when jumping to err_close_sockfd and in line 478

gleon99:
Hi @dmitrygx - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

Issue #8398, opened 2022-07-22
UCP server callback is not invoked upon a client connection


rrgargeya:
### Describe the bug
We are trying to use UCP to implement clients and servers. 
This is replace our Verbs based implementation where clients and servers are connected via RC QPs.

I have an issue in that the server's callback is not getting called when the client tries to connect.

The error callback on the client is not getting called (properly perhaps as there was no error).

These are IB interfaces, and I see in the trace that a TCP listener is created, and not a RDMACM listener.

Is that the problem?

I see that 
UCX_SOCKADDR_TLS_PRIORITY=rdmacm,tcp,sockcm

So, I am assuming rdmacm must take precedence before tcp.

Server's listener creation code snippet:
   params.field_mask         = UCP_LISTENER_PARAM_FIELD_SOCK_ADDR |
                               UCP_LISTENER_PARAM_FIELD_CONN_HANDLER;
   params.sockaddr.addr      = (const struct sockaddr*)&server_addr;
   params.sockaddr.addrlen   = sizeof(struct sockaddr);
   params.conn_handler.cb    = server_conn_handle_cb;
   params.conn_handler.arg   = nullptr;
   status = ucp_listener_create(ucp_worker, &params, &ucp_listener_);
   
Client's endpoint creation code snippet:
   ep_params.field_mask       =  UCP_EP_PARAM_FIELD_FLAGS       |
                                 UCP_EP_PARAM_FIELD_SOCK_ADDR   |
                                 UCP_EP_PARAM_FIELD_USER_DATA   |
                                 UCP_EP_PARAM_FIELD_ERR_HANDLER |
                                 UCP_EP_PARAM_FIELD_ERR_HANDLING_MODE;
   ep_params.err_mode         =  UCP_ERR_HANDLING_MODE_PEER;
   ep_params.err_handler.cb   =  client_err_cb;
   ep_params.err_handler.arg  =  nullptr;
   ep_params.user_data        = &connect_user_data;
   ep_params.flags            =  UCP_EP_PARAMS_FLAGS_CLIENT_SERVER |
                                 UCP_EP_PARAMS_FLAGS_SEND_CLIENT_ID;
   ep_params.sockaddr.addr    = (struct sockaddr*) &connect_addr;
   ep_params.sockaddr.addrlen = sizeof(connect_addr);

   status = ucp_ep_create(ucp_worker, &ep_params, &ucp_client_ep);
 
Server connection handler prototype: 
[...]

PR #8403, opened 2022-07-26
README: add requirements for IB and RoCE transports

evgeny-leksikov:
## What
add requirements for IB and RoCE transports

## Why ?
https://github.com/openucx/ucx/issues/8398

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,7 @@ $ make -C test/gtest test
     * posix, sysv, [cma](https://dl.acm.org/citation.cfm?id=2616532), [knem](http://knem.gforge.inria.fr/), and [xpmem](https://github.com/hjelmn/xpmem)
 * TCP/IP
 
+  **NOTE:** [Infiniband](https://www.infinibandta.org/) and [RoCE](http://www.roceinitiative.org/) transports requires rdma-core >= 28.0 or MLNX_OFED >= 5.0
 <hr>
 
 ## Supported CPU Architectures


README.md
```diff
@@ -165,6 +165,7 @@ $ make -C test/gtest test
     * posix, sysv, [cma](https://dl.acm.org/citation.cfm?id=2616532), [knem](http://knem.gforge.inria.fr/), and [xpmem](https://github.com/hjelmn/xpmem)
 * TCP/IP
 
+  **NOTE:** [Infiniband](https://www.infinibandta.org/) and [RoCE](http://www.roceinitiative.org/) transports requires rdma-core >= 28.0 or MLNX_OFED >= 5.0
```
yosefe, line 168: requires --> require

yosefe:
@tonycurtis can you pls take a look?

README.md
tonycurtis: ditto

shamisp: Starting from what UCX version ?

evgeny-leksikov: > Starting from what UCX version ?

I think master version of README should be relevant to master version, otherwise we should describe it for all versions, what can be problematic to restore. @yosefe WDYT?

yosefe (approved):

evgeny-leksikov:
@tonycurtis @shamisp do you have other comments?

tonycurtis (approved):

README.md
shamisp: @evgeny-leksikov This is not accurate. The readme is used as a front page of the project. If you look at the prevision notes you will see that we explicitly call out UCX versions with problems and etc. We need to specify UCX version that requires latest MOFED and is prone to this specific problem.

evgeny-leksikov: @shamisp do you mean Known Issues section? In this PR I would like to highlight a requirement but not an issue.

shamisp: @evgeny-leksikov It is okay. Do we require this for all versions or starting from certain version ?

evgeny-leksikov: @shamisp basically every SW requires updates time to time, we cannot track every UCX version in past where any issue was fixed. especially if it's found in future. Please note, that actualy this is not an issue, issue was highlited by warning in configure output.
I agree that switching to new MOFED version was not highlited in time, this is an issue and probably should be handled separately.

shamisp: There are minimum software requirement for every SW version.  I think the statement "Use latest MOFED" by itself is not very useful.

evgeny-leksikov: Why? if you have an issue - try to upgrade and in many cases it's already fixed, If everything is stable and working as expected - juct don't touch it :) IMHO front page of the project is important firstly for newcomers.

shamisp: Well, sometime UCX is behind rdma-core on changes. So the advice "use latest" is not always true. 
My suggestion - just pick a version; UCX 1.12.0 and above require MOFED XX. If later you discover that 1.11.x requires later MOFED, we can update the message.

evgeny-leksikov: ok, added 1.12.0

shamisp (approved):

yosefe (dismissed):

evgeny-leksikov:
squashed

PR #8408, opened 2022-07-27
UCT/IB: Check verbs support relaxed order - v1.13.x

Artemy-Mellanox:


yosefe (approved):

src/uct/ib/base/ib_md.c
```diff
@@ -1584,7 +1584,7 @@ void uct_ib_md_parse_relaxed_order(uct_ib_md_t *md,
             ucs_warn("relaxed order memory access requested but not supported");
         }
     } else if (md_config->mr_relaxed_order == UCS_CONFIG_AUTO) {
-        if (ucs_cpu_prefer_relaxed_order()) {
+        if (IBV_ACCESS_RELAXED_ORDERING && ucs_cpu_prefer_relaxed_order()) {
```
shamisp, line 1587: Is not this just enum ? I don't understand this branch.

shamisp:
The `HAVE_*` is really more of ifdef than runtime branch. 
Is there rdma-core runtime query that you can use to check this ?

yosefe:
> IBV_ACCESS_RELAXED_ORDERING

Seems it's a best effort flag and no indication if it's actually doing anything, right @jgunthorpe ?
https://lore.kernel.org/linux-rdma/ceb93ff8-80f0-f321-531f-4c534da38e07@mellanox.com/t/

src/uct/ib/base/ib_md.c
```diff
@@ -1578,13 +1578,14 @@ void uct_ib_md_parse_relaxed_order(uct_ib_md_t *md,
...
     if (md_config->mr_relaxed_order == UCS_CONFIG_ON) {
-        if (IBV_ACCESS_RELAXED_ORDERING) {
+        if (HAVE_DECL_IBV_ACCESS_RELAXED_ORDERING) {
             md->relaxed_order = 1;
         } else {
             ucs_warn("relaxed order memory access requested but not supported");
         }
     } else if (md_config->mr_relaxed_order == UCS_CONFIG_AUTO) {
-        if (ucs_cpu_prefer_relaxed_order()) {
+        if (HAVE_DECL_IBV_ACCESS_RELAXED_ORDERING &&
```
yosefe, line 1587: can we unite this if with the if in line 1586, using && ?

yosefe (dismissed):

src/uct/ib/base/ib_md.c
shamisp: @yosefe I would rather use HAVE_DECL_IBV_ACCESS_RELAXED_ORDERING to ifdef passing the flag. It is fixed value so IMHO it cleaner to ifdef.

jgunthorpe:
Yes, best effort. So long as the C symbol exists users should set it in places where it is appropriate to use.

yosefe (approved):
pls squash

yosefe (approved):

yosefe:
@shamisp is that OK now?

PR #8386, opened 2022-07-15
UCP/PROTO: Add request reset for switching ep config during wireup

evgeny-leksikov:
## What
Add UCP request clean and reinit after wireup:
 - infrastructure and basic protocols

## Why ?
EP might be reconfigured during wireup, then initially selected protocols are not applicable for pending requests and should be re-selected, so the requests also should be re-inisialized.

## How ?
 - add `ucp_proto_t::clean`
 - restart requests from pending replay:
   - clean
   - init
   - send

PR #8410, opened 2022-07-28
UCP/WIREUP: connect lanes in reverse ordering

evgeny-leksikov:
## What
connect lanes in reverse ordering

## Why ?
 - to optimize multiple re-scheduling of requests with multi-fragment protocols
 - to avoid re-initializtion of started protocol according to #8386

### DIFF
--- a/src/ucp/wireup/wireup.c
+++ b/src/ucp/wireup/wireup.c
@@ -419,9 +419,12 @@ out:
 
 void ucp_wireup_remote_connect_lanes(ucp_ep_h ep, int ready)
 {
-    ucp_lane_index_t lane;
+    ucp_lane_index_t i, lane;
 
-    for (lane = 0; lane < ucp_ep_num_lanes(ep); ++lane) {
+    for (i = 0; i < ucp_ep_num_lanes(ep);) {
+        /* Connect lanes in reverse ordering to avoid replaying requests with
+         * multi fragment protocol from one to other wireup lane. */
+        lane = ucp_ep_num_lanes(ep) - (++i);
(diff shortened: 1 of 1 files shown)


src/ucp/wireup/wireup.c
```diff
@@ -419,9 +419,12 @@ ucp_wireup_connect_local(ucp_ep_h ep,
...
 void ucp_wireup_remote_connect_lanes(ucp_ep_h ep, int ready)
 {
-    ucp_lane_index_t lane;
+    ucp_lane_index_t i, lane;
 
-    for (lane = 0; lane < ucp_ep_num_lanes(ep); ++lane) {
+    for (i = 0; i < ucp_ep_num_lanes(ep);) {
+        /* Connect lanes in reverse ordering to avoid replaying requests with
+         * multi fragment protocol from one to other wireup lane. */
+        lane = ucp_ep_num_lanes(ep) - (++i);
```
yosefe, line 427: 1. why reverse order helps here?
2.  IMO this looks better:
```
int lane;
for (lane = ucp_ep_num_lanes(ep) - 1; lane >= 0; --lane) {
```

evgeny-leksikov: > why reverse order helps here?

no moving request from one wireup ep to another. If ordering of connecting lanes is the same as `request->send.lane` then the request will go via all wireup endpoints and their pending queues

evgeny-leksikov: > IMO this looks better

My loop is more positive :) and no type cast.. but need to deside ordering approach first

evgeny-leksikov:
replaced with https://github.com/openucx/ucx/pull/8467

PR #8411, opened 2022-07-28
UCP/PROTO/AM: impl ucp_proto_t::reset for eager protocols

evgeny-leksikov:
## What
Implementation of `ucp_proto_t::reset` for eager protocols
Depends on https://github.com/openucx/ucx/pull/8386

## Why ?
The method was intoduced in https://github.com/openucx/ucx/pull/8386 and should be implemented for all protocols

### DIFF
--- a/src/ucp/core/ucp_am.c
+++ b/src/ucp/core/ucp_am.c
@@ -1737,3 +1737,12 @@ const ucp_request_send_proto_t ucp_am_reply_proto = {
     .zcopy_completion       = ucp_am_zcopy_completion,
     .only_hdr_size          = sizeof(ucp_am_hdr_t) + sizeof(ucp_am_reply_ftr_t)
 };
+
+void ucp_am_proto_request_zcopy_reset(ucp_request_t *request)
+{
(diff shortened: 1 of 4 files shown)


src/ucp/core/ucp_am.c
```diff
@@ -1737,3 +1737,14 @@ const ucp_request_send_proto_t ucp_am_reply_proto = {
     .zcopy_completion       = ucp_am_zcopy_completion,
     .only_hdr_size          = sizeof(ucp_am_hdr_t) + sizeof(ucp_am_reply_ftr_t)
 };
+
+void ucp_am_proto_request_zcopy_reset(ucp_request_t *request)
+{
+    ucs_assert(request->send.msg_proto.am.header.reg_desc != NULL);
+    ucs_mpool_put_inline(request->send.msg_proto.am.header.reg_desc);
+#if ENABLE_ASSERT
```
yosefe, line 1745: maybe remove? this is slow path

yosefe (approved):

Issue #8415, opened 2022-07-28
Segfaults in malloc when compiled with LTO


omor1:
### Describe the bug
Compiling UCX 1.13.0 with AOCC and using `-flto`, I get segfaults. This occurs both with `ucx_info` (see #8402) and with other consumers e.g. Open MPI. Segfaults occur from inside libc, when allocating memory in  `mlx5_alloc_context` called from `uct_ib_mlx5_devx_open_device`.

### Steps to Reproduce
Compile UCX 1.13.0 with `CFLAGS="-flto" CXXFLAGS="-flto" LDFLAGS="-flto"`

### Setup and versions
- Rocky Linux 8.5
- AMD "Rome" x86_64
- MLNX_OFED_LINUX-4.9-4.1.7.0

### Additional information (depending on the issue)
- Seems to be an issue with the malloc hooks—the build with LTO shows that the config test for them fails, while the normal build is fine.

omor1:
Specifically, it seems that at `-O2` or `-O3` with `-flto`, the test program (below) does not work.

```C
#include <malloc.h>

static int rc = 1;
void *ptr;
void *myhook(size_t size, const void *caller) {
    rc = 0;
    return NULL;
}

int main(int argc, char** argv) {
    __malloc_hook = myhook;
    ptr = malloc(1);
    return rc;
}
```

yosefe:
@omor1 can you pls post the command line and the backtrace of the segfault?

omor1:
```console
$ gdb /path/to/ucx_info
GNU gdb (GDB) 10.1
/* ... */
(gdb) run -c
Starting program: /path/to/ucx_info -c
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
warning: Loadable section ".note.gnu.property" outside of ELF segments
warning: Loadable section ".note.gnu.property" outside of ELF segments
warning: Loadable section ".note.gnu.property" outside of ELF segments
warning: Loadable section ".note.gnu.property" outside of ELF segments
warning: Loadable section ".note.gnu.property" outside of ELF segments

Program received signal SIGSEGV, Segmentation fault.
0x0000155554563b96 in sysmalloc () from /lib64/libc.so.6
(gdb) bt
#0  0x0000155554563b96 in sysmalloc () from /lib64/libc.so.6
#1  0x0000155554564aa9 in _int_malloc () from /lib64/libc.so.6
[...]

PR #8417, opened 2022-07-31
UCM/SYS: Don't use inline assembly for brk syscall

yosefe:
## Why
Potential fix for https://github.com/openucx/ucx/issues/8415

cc @omor1

### DIFF
--- a/src/ucm/util/sys.c
+++ b/src/ucm/util/sys.c
@@ -366,19 +366,10 @@ char *ucm_concat_path(char *buffer, size_t max, const char *dir, const char *fil
 
 void *ucm_brk_syscall(void *addr)
 {
-    void *result;
+    /* Return type is equivalent to full pointer size */
+    UCS_STATIC_ASSERT(sizeof(syscall(0)) == sizeof(void*));
 
-#ifdef __x86_64__
-    asm volatile("mov %1, %%rdi\n\t"
-                 "mov $0xc, %%eax\n\t"
-                 "syscall\n\t"
-                 : "=a"(result)
-                 : "m"(addr));
-#else
-    /* TODO implement 64-bit syscall for aarch64, ppc64le */
-    result = (void*)syscall(SYS_brk, addr);
-#endif
-    return result;
+    return (void*)syscall(SYS_brk, addr);
 }
 
 pid_t ucm_get_tid()


hoopoepg (approved):

shamisp:
was this tested ?

hoopoepg:
it should be covered by CI - we have set of tests for Bistro infra

shamisp:
Well clearly it does not cover @omor1 case ...

yosefe:
@omor1 can you pls check if this PR fixes your issue? IMO it should because it eliminates the inline asm, so also the register clobbering issue.

omor1:
I'll double-check soon, but it should fix the issue—the glibc syscall wrapper should be correct.

omor1:
I can confirm that this fixes the segfault. Instead of calling the `syscall` instruction directly there are calls (and, in some leaf function cases, direct jumps) to the `syscall` function via the PLT. There still needs to be a fix for the `HAVE_MALLOC_HOOK` test, since that still fails. Your suggestion to print `ptr` works to fix that.

yosefe:
> I can confirm that this fixes the segfault. Instead of calling the `syscall` instruction directly there are calls (and, in some leaf function cases, direct jumps) to the `syscall` function via the PLT. There still needs to be a fix for the `HAVE_MALLOC_HOOK` test, since that still fails. Your suggestion to print `ptr` works to fix that.

I don't think we need to change it. If LTO is really eliminating calls to malloc, it means malloc hooks really do not work.
Anyway, bistro mode (patching mmap etc) is enough; malloc hooks are needed today only when running with valgrind

omor1:
> I don't think we need to change it. If LTO is really eliminating calls to malloc, it means malloc hooks really do not work. Anyway, bistro mode (patching mmap etc) is enough; malloc hooks are needed today only when running with valgrind

The call to `malloc` is eliminated because `ptr` itself is eliminated; the compiler is able to prove—in the simple test used for malloc hooks—that `ptr` isn't used by anything and, since it's local to the executable, that it can't be accessed by an external caller. In an actual application—and particularly in a shared library—these are both not true. I think it's still worthwhile to fix the test so that using LTO doesn't silently disable a capability that otherwise works correctly. Someday you may get a compiler that even without LTO is able to make similar optimizations...

omor1:
I'll also note that this is—potentially—a bug in both LLVM and GCC where specifically calls to `malloc` can be optimized away; similarly-declared functions do not get that luxury. Linking to a miniature shared library with a function `void *mymalloc(size_t size) __attribute__((nothrow, warn_unused_result, malloc))` that simply calls `malloc` prevents the allocation from being eliminated—it seems that LLVM and GCC both assume that `malloc` does not have observable side-effects when compiled with LTO.

PR #8418, opened 2022-08-01
UCT/TCP/CM: added support for IPv6 link-local

hoopoepg:
- added support for IPv6 link local connections
- added scope_id value into tcp device address and
  conn_request device address


gleon99:
Hi @hoopoepg - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #8423, opened 2022-08-02
CONTRIB/JENKINS: Move TCP port range to not collide with ip_local_port_range

yosefe:
## Why
When using ports from range 33000..60000, it can collide with TCP local (client) port numbers used by various processes on the system, for example NFS client. This can cause random CI failures.

### DIFF
--- a/src/uct/tcp/tcp_iface.c
+++ b/src/uct/tcp/tcp_iface.c
@@ -542,53 +542,6 @@ err:
     return status;
 }
 
-static ucs_status_t uct_tcp_iface_set_port_range(uct_tcp_iface_t *iface,
-                                                 uct_tcp_iface_config_t *config)
-{
-    ucs_range_spec_t system_port_range;
-    unsigned start_range, end_range;
-    ucs_status_t status;
-
-    if ((config->port_range.first == 0) && (config->port_range.last == 0)) {
-        /* using a random port */
-        goto out_set_config;
-    }
-
-    /* get the system configuration for usable ports range */
-    status = ucs_sockaddr_get_ip_local_port_range(&system_port_range);
-    if (status != UCS_OK) {
-        /* using the user's config */
-        goto out_set_config;
-    }
-
-    /* find a common range between the user's ports range and the one on the system */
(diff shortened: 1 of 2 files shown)


yosefe:
@hoopoepg @dmitrygx can you pls take a look?

src/uct/tcp/tcp_iface.c
```diff
@@ -694,17 +655,14 @@ static UCS_CLASS_INIT_FUNC(uct_tcp_iface_t, uct_md_h md, uct_worker_h worker,
...
-
     if (self->config.tx_seg_size > self->config.rx_seg_size) {
         ucs_error("RX segment size (%zu) must be >= TX segment size (%zu)",
                   self->config.rx_seg_size, self->config.tx_seg_size);
-        return UCS_ERR_INVALID_PARAM;
+        status = UCS_ERR_INVALID_PARAM;
+        goto err;
     }
 
+    uct_tcp_iface_set_port_range(self, config);
```
hoopoepg, line 665: minor: may be expand uct_tcp_iface_set_port_range here? there is only 2 lines of code + debug now

hoopoepg (approved):
just minor question

src/uct/tcp/tcp_iface.c
yosefe: @hoopoepg fixed, and also removed the debug print, since uct_tcp_iface_listener_init already prints the selected server port number.

hoopoepg (approved):

src/uct/tcp/tcp_iface.c
```diff
@@ -542,53 +542,6 @@ static ucs_status_t uct_tcp_iface_listener_init(uct_tcp_iface_t *iface)
     return status;
 }
 
-static ucs_status_t uct_tcp_iface_set_port_range(uct_tcp_iface_t *iface,
```
dmitrygx, line 545: `ucs_sockaddr_get_ip_local_port_range` is no longer used.
shall we remove it? or probably we can leave for future reuse somewhere?

PR #8424, opened 2022-08-02
CONFIG/BUILD: Fix malloc hook detection with LTO

yosefe:
## Why
When LTO is enabled, malloc hook test can fail because ptr is unused and eliminated.

Per https://github.com/openucx/ucx/pull/8417#issuecomment-1202151666

@cc omor1


hoopoepg (approved):

shamisp:
@yosefe do we see this problem when UCX is compiled "regularly" but let's say MPI or any other upper layer compiled with LTO ?

shamisp:
I'm asking about this because II suspect that configure level detection is not enough ...

yosefe:
it should not be related to upper layer; issue is only with the small test program in UCX configure stage

PR #8437, opened 2022-08-08
UCT/AM: Change UCT AM handlers API to support additional optional par…

ofirfarjun7:
…ameters

## What
Change UCT AM API to support additional optional parameters
 
## Why ?
Supporting user memory allocator and signature offload requires passing additional parameters to
AM handlers. 

## How ?
Defining new params struct that holds the additional required parameters and pass it to AM handler.

### DIFF
--- a/src/uct/api/uct_def.h
+++ b/src/uct/api/uct_def.h
@@ -488,7 +488,7 @@ enum uct_am_callback_params_field {
     /** Enables @ref uct_am_callback_params_t::payload
      *  Indicates that payload field in uct_am_callback_params_t is valid.
      */
-    UCT_AM_CALLBACK_PARAM_FIELD_PAYLOAD  = UCS_BIT(0),
+    UCT_AM_CALLBACK_PARAM_FIELD_PAYLOAD = UCS_BIT(0),
 };
 
 
@@ -541,7 +541,9 @@ typedef struct uct_am_callback_params {
  *                          released later by their owner.
  *
  */
-typedef ucs_status_t (*uct_am_callback_t)(void *arg, void *data, size_t length, unsigned flags, uct_am_callback_params_t *params);
+typedef ucs_status_t (*uct_am_callback_t)(void *arg, void *data, size_t length,
+                                          unsigned flags,
+                                          uct_am_callback_params_t *params);
 
 
 /**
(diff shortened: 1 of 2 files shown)


src/uct/api/uct_def.h
```diff
@@ -475,34 +475,75 @@ typedef struct uct_cm_ep_server_conn_notify_args {
...
- * If @a flags contain @ref UCT_CB_PARAM_FLAG_DESC value, it means @a data is part of
+ * When the callback is called, @a flags indicates how @a msg_header should be handled.
+ * If @a flags contain @ref UCT_CB_PARAM_FLAG_DESC value, it means @a msg_header is part of
  * a descriptor which must be released later by @ref uct_iface_release_desc by
  * the user if the callback returns @ref UCS_INPROGRESS.
+ * See @ref uct_am_callback_params_t for more information about @a param.
  *
  * @param [in]  arg      User-defined argument.
- * @param [in]  data     Points to the received data. This may be a part of
+ * @param [in]  msg_hdr  Points to the received message header. This may be a part of
```
yosefe, line 525: why changed?

src/uct/api/uct_def.h
```diff
@@ -475,34 +475,75 @@ typedef struct uct_cm_ep_server_conn_notify_args {
...
+
+
+typedef struct uct_am_callback_params {
+    /**
+     * Mask of valid fields in this structure, using bits from
+     * @ref uct_am_callback_params_field.
+     * Fields not specified by this mask should not be accessed by the callback.
+     */
+    uint64_t field_mask;
+    /**
```
yosefe, line 502: space line before

src/uct/api/uct_def.h
ofirfarjun7: After the change data actually point to the message header descriptor, IMO it's not a name that gives a good description, so I thought to change it.
Can reverse it if you not agree or you want to reduce the changes to the absolute minimum.

yosefe: but i see function arg name was not changed, only the doc

ofirfarjun7: This is a mistake... it's should be change as well to msg_hdr

yosefe: ok

PR #8445, opened 2022-08-11
DC/HW-DCS: enabled HW DCS

hoopoepg:
- enabled HW DCS support
- added runtime configuration to limit DCI streams to use
- updated *post_op routines to set DCI stream channel ID
- added HCA capabilities read to get max DCI stream number

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.c
+++ b/src/uct/ib/dc/dc_mlx5.c
@@ -103,6 +103,12 @@ ucs_config_field_t uct_dc_mlx5_iface_config_sub_table[] = {
      ucs_offsetof(uct_dc_mlx5_iface_config_t, fc_hard_req_timeout),
      UCS_CONFIG_TYPE_TIME_UNITS},
 
+    {"NUM_DCI_CHANNELS", "1",
+     "Number of stream channels per DCI to be used in DCI-rand mode. A value "
+     "of 1 disables DCI multi-channel support.",
+     ucs_offsetof(uct_dc_mlx5_iface_config_t, num_dci_channels),
+     UCS_CONFIG_TYPE_UINT},
+
     {NULL}
 };
 
@@ -342,10 +348,11 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
 
     if (md->flags & UCT_IB_MLX5_MD_FLAG_DEVX_DCI) {
         attr.super.max_inl_cqe[UCT_IB_DIR_RX] = 0;
-        attr.uidx             = htonl(dci_index) >> UCT_IB_UIDX_SHIFT;
-        attr.full_handshake   = full_handshake;
-        attr.rdma_wr_disabled = (iface->flags & UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT) &&
-                                (md->flags & UCT_IB_MLX5_MD_FLAG_NO_RDMA_WR_OPTIMIZED);
+        attr.uidx                        = htonl(dci_index) >> UCT_IB_UIDX_SHIFT;
+        attr.full_handshake              = full_handshake;
+        attr.rdma_wr_disabled            = (iface->flags & UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT) &&
+                                           (md->flags & UCT_IB_MLX5_MD_FLAG_NO_RDMA_WR_OPTIMIZED);
+        attr.log_num_dci_stream_channels = ucs_ilog2(iface->tx.num_dci_channels);
         status = uct_ib_mlx5_devx_create_qp(ib_iface,
                                             &iface->super.cq[UCT_IB_DIR_TX],
                                             &iface->super.cq[UCT_IB_DIR_RX],
@@ -1341,6 +1348,7 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
     size_t sq_length;
     ucs_status_t status;
     unsigned tx_cq_size;
+    unsigned num_dci_channels;
 
     ucs_trace_func("");
 
@@ -1406,6 +1414,12 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
     self->tx.pend_cb   = uct_dc_mlx5_iface_is_dci_rand(self) ?
                          uct_dc_mlx5_iface_dci_do_rand_pending_tx :
                          uct_dc_mlx5_iface_dci_do_dcs_pending_tx;
+
+    /* calculate num_dci_channels: select minimum from requested by runtime
+     * and supported by HCA, must be power of two */
+    num_dci_channels = ucs_roundup_pow2(ucs_max(config->num_dci_channels, 1));
+    self->tx.num_dci_channels = ucs_min(num_dci_channels,
+                                        UCS_BIT(md->log_max_dci_stream_channels));
 
     self->tx.dci_pool_release_bitmap = 0;
 
--- a/src/uct/ib/dc/dc_mlx5.h
+++ b/src/uct/ib/dc/dc_mlx5.h
@@ -170,6 +170,7 @@ typedef struct uct_dc_mlx5_iface_config {
     unsigned                            rand_seed;
     ucs_time_t                          fc_hard_req_timeout;
     uct_ud_mlx5_iface_common_config_t   mlx5_ud;
+    unsigned                            num_dci_channels;
 } uct_dc_mlx5_iface_config_t;
 
 
@@ -194,6 +195,8 @@ typedef struct uct_dc_dci {
     };
     uint8_t                       pool_index; /* DCI pool index. */
     uint8_t                       path_index; /* Path index */
+    uint8_t                       next_channel_index; /* next DCI channel index
+                                                         to be used by EP */
 } uct_dc_dci_t;
 
 
@@ -296,6 +299,8 @@ struct uct_dc_mlx5_iface {
         uint8_t                   dci_pool_release_bitmap;
 
         uint8_t                   av_fl_mlid;
+
+        uint8_t                   num_dci_channels;
     } tx;
 
     struct {
--- a/src/uct/ib/dc/dc_mlx5_ep.h
+++ b/src/uct/ib/dc/dc_mlx5_ep.h
@@ -76,6 +76,7 @@ struct uct_dc_mlx5_ep {
     uint16_t              flush_rkey_hi;
     uct_rc_fc_t           fc;
     uct_dc_mlx5_base_av_t av;
+    uint8_t               dci_channel_index;
 };
 
 typedef struct {
@@ -307,9 +308,12 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 
     if (uct_dc_mlx5_iface_is_dci_rand(iface)) {
         /* coverity[dont_call] */
-        ep->dci = rand_r(&iface->tx.rand_seed) % iface->tx.ndci;
+        ep->dci               = rand_r(&iface->tx.rand_seed) % iface->tx.ndci;
+        ep->dci_channel_index = iface->tx.dcis[ep->dci].next_channel_index++ %
+                                iface->tx.num_dci_channels;
     } else {
-        ep->dci = UCT_DC_MLX5_EP_NO_DCI;
+        ep->dci               = UCT_DC_MLX5_EP_NO_DCI;
+        ep->dci_channel_index = 0;
     }
 
     return uct_rc_fc_init(&ep->fc, &iface->super.super
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -14,6 +14,9 @@
 #include <ucs/arch/bitops.h>
 #include <ucs/profile/profile.h>
 
+/* max log value to store in uint8_t */
+#define UCT_IB_MLX5_MD_MAX_DCI_CHANNELS 8
+
 typedef struct {
     struct mlx5dv_devx_obj     *dvmr;
     int                        mr_num;
@@ -946,6 +949,12 @@ static ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
                 UCT_IB_MLX5DV_GET(query_hca_cap_out, out, syndrome));
         goto err_free;
     }
+
+    UCS_STATIC_ASSERT(UCS_MASK(UCT_IB_MLX5_MD_MAX_DCI_CHANNELS) <= UINT8_MAX);
+    md->log_max_dci_stream_channels = UCT_IB_MLX5DV_GET(cmd_hca_cap, cap,
+                                                        log_max_dci_stream_channels);
+    md->log_max_dci_stream_channels = ucs_min(md->log_max_dci_stream_channels,
+                                              UCT_IB_MLX5_MD_MAX_DCI_CHANNELS);
 
     if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, log_max_msg) !=
         UCT_IB_MLX5_LOG_MAX_MSG_SIZE) {
--- a/src/uct/ib/mlx5/ib_mlx5.inl
+++ b/src/uct/ib/mlx5/ib_mlx5.inl
@@ -6,6 +6,14 @@
 
 #include "ib_mlx5.h"
 
+typedef struct uct_ib_mlx5_wqe_ctrl_seg {
+    __be32      opmod_idx_opcode;
+    __be32      qpn_ds;
+    uint8_t     signature;
+    __be16      dci_stream_channel_id;
+    uint8_t     fm_ce_se;
+    __be32      imm;
+} UCS_S_PACKED uct_ib_mlx5_wqe_ctrl_seg_t;
 
 static UCS_F_ALWAYS_INLINE UCS_F_NON_NULL struct mlx5_cqe64*
 uct_ib_mlx5_get_cqe(uct_ib_mlx5_cq_t *cq,  unsigned cqe_index)
@@ -329,7 +337,8 @@ uct_ib_mlx5_set_ctrl_qpn_ds(struct mlx5_wqe_ctrl_seg *ctrl, uint32_t qp_num,
 static UCS_F_ALWAYS_INLINE void
 uct_ib_mlx5_set_ctrl_seg(struct mlx5_wqe_ctrl_seg* ctrl, uint16_t pi,
                          uint8_t opcode, uint8_t opmod, uint32_t qp_num,
-                         uint8_t fm_ce_se, unsigned wqe_size)
+                         uint8_t fm_ce_se, uint16_t dci_channel,
+                         unsigned wqe_size)
 {
     uint8_t ds = ucs_div_round_up(wqe_size, UCT_IB_MLX5_WQE_SEG_SIZE);
 #if defined(__ARM_NEON)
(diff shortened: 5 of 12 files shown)


src/uct/ib/dc/dc_mlx5.c
```diff
@@ -103,6 +103,11 @@ ucs_config_field_t uct_dc_mlx5_iface_config_sub_table[] = {
      ucs_offsetof(uct_dc_mlx5_iface_config_t, fc_hard_req_timeout),
      UCS_CONFIG_TYPE_TIME_UNITS},
 
+    {"MAX_DCI_STREAM_CHANNELS", "1",
```
yosefe, line 106: NUM_DCI_CHANNELS

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -103,6 +103,11 @@ ucs_config_field_t uct_dc_mlx5_iface_config_sub_table[] = {
...
+     "Maximum DCI stream channels per DCI to be used in DCI-rand mode",
```
yosefe, line 107: "Number of stream channels per DCI to be used in DCI-rand mode. A value of 1 disables"
"DCI multi-channel support." ,

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1407,6 +1413,10 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
                          uct_dc_mlx5_iface_dci_do_rand_pending_tx :
                          uct_dc_mlx5_iface_dci_do_dcs_pending_tx;
 
+    self->tx.log_num_dci_stream_channels =
+            ucs_min(ucs_log2(ucs_roundup_pow2(config->max_dci_stream_channels)),
```
yosefe, line 1417: should we make sure this value is at least 1?

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -170,6 +170,7 @@ typedef struct uct_dc_mlx5_iface_config {
     unsigned                            rand_seed;
     ucs_time_t                          fc_hard_req_timeout;
     uct_ud_mlx5_iface_common_config_t   mlx5_ud;
+    unsigned                            max_dci_stream_channels;
```
yosefe, line 173: num_dci_channels

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -194,6 +195,8 @@ typedef struct uct_dc_dci {
     };
     uint8_t                       pool_index; /* DCI pool index. */
     uint8_t                       path_index; /* Path index */
+    uint8_t                       dci_channel_index; /* DCI channel index to
```
yosefe, line 198: next_channel_index
and pls elaborate on the documentation

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -76,6 +76,7 @@ struct uct_dc_mlx5_ep {
     uint16_t              flush_rkey_hi;
     uct_rc_fc_t           fc;
     uct_dc_mlx5_base_av_t av;
+    uint8_t               dci_stream_channel_id;
```
yosefe, line 79: 1. dci_channel_index
2. do we really have to keep dci_channel_index? can't we take it from dci?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -307,9 +308,12 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 
     if (uct_dc_mlx5_iface_is_dci_rand(iface)) {
         /* coverity[dont_call] */
-        ep->dci = rand_r(&iface->tx.rand_seed) % iface->tx.ndci;
+        ep->dci                   = rand_r(&iface->tx.rand_seed) % iface->tx.ndci;
+        ep->dci_stream_channel_id = iface->tx.dcis[ep->dci].dci_channel_index++ %
```
yosefe, line 312: can we use & operation?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -947,6 +947,10 @@ static ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
         goto err_free;
     }
 
+    md->log_max_dci_stream_channels = UCT_IB_MLX5DV_GET(cmd_hca_cap, cap,
+                                                        log_max_dci_stream_channels);
+    ucs_assert(md->log_max_dci_stream_channels < 8);
```
yosefe, line 952: why is this assert needed?
i think need better handling for such case

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -296,6 +299,8 @@ struct uct_dc_mlx5_iface {
         uint8_t                   dci_pool_release_bitmap;
 
         uint8_t                   av_fl_mlid;
+
+        uint8_t                   log_num_dci_stream_channels;
```
yosefe, line 303: num_dci_channels

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -329,91 +337,103 @@ uct_ib_mlx5_set_ctrl_qpn_ds(struct mlx5_wqe_ctrl_seg *ctrl, uint32_t qp_num,
...
-                        16, 16,          /* reserved (set 0) */
+                        9, 8,            /* dci stream channel ID in BE */
                         0,               /* signal/fence_mode */
                         16, 16, 16, 16}; /* immediate (set to 0)*/
-    uint32x4_t data = {(opcode << 16) | (opmod << 8) | (uint32_t)fm_ce_se,
-                       pi, ds, qp_num};
+    uint32x4_t data = {(ds << 24) | (opcode << 16) | (opmod << 8) | fm_ce_se,
+                       pi, dci_channel, qp_num};
+#elif !defined (__SSE4_2__)
+    uct_ib_mlx5_wqe_ctrl_seg_t *uct_ctrl = (uct_ib_mlx5_wqe_ctrl_seg_t*)ctrl;
```
yosefe, line 357: we should use uct_ib_mlx5_wqe_ctrl_seg_t instead of mlx5_wqe_ctrl_seg everywhere; not just cast to it here

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -329,91 +337,103 @@ uct_ib_mlx5_set_ctrl_qpn_ds(struct mlx5_wqe_ctrl_seg *ctrl, uint32_t qp_num,
...
-                        8,               /* data size */
+                        4,               /* data size */
                         16,              /* signature (set 0) */
-                        16, 16,          /* reserved (set 0) */
+                        9, 8,            /* dci stream channel ID in BE */
                         0,               /* signal/fence_mode */
                         16, 16, 16, 16}; /* immediate (set to 0)*/
-    uint32x4_t data = {(opcode << 16) | (opmod << 8) | (uint32_t)fm_ce_se,
-                       pi, ds, qp_num};
+    uint32x4_t data = {(ds << 24) | (opcode << 16) | (opmod << 8) | fm_ce_se,
```
yosefe, line 354: why needed to move ds ?

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -329,91 +337,103 @@ uct_ib_mlx5_set_ctrl_qpn_ds(struct mlx5_wqe_ctrl_seg *ctrl, uint32_t qp_num,
...
-                                 12, 13, 14, /* QP num */
-                                 2,          /* opcode */
-                                 4, 5,       /* sw_pi in BE */
-                                 1           /* opmod */
+                    _mm_set_epi32(qp_num, dci_channel, pi,
+                                  (ds << 24) | (opcode << 16) | (opmod << 8) | fm_ce_se), /* OR of constants */
+                    _mm_set_epi8(0x80, 0x80, 0x80, 0x80, /* immediate (set 0) */
+                                 0,                      /* signal/fence_mode */
+                                 8, 9,                   /* reserved */
+                                 0x80,                   /* signature (set 0) */
```
yosefe, line 369: what 0x80 means? zero?

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -329,91 +337,103 @@ uct_ib_mlx5_set_ctrl_qpn_ds(struct mlx5_wqe_ctrl_seg *ctrl, uint32_t qp_num,
...
+                                 1                       /* opmod */
                                  ));
 #elif defined(__ARM_NEON)
     *(uint8x16_t *)ctrl = vqtbl1q_u8((uint8x16_t)data, table);
 #else
-    ctrl->opmod_idx_opcode = (opcode << 24) | (htons(pi) << 8) | opmod;
+    uct_ctrl->opmod_idx_opcode      = (opcode << 24) | (htons(pi) << 8) | opmod;
     uct_ib_mlx5_set_ctrl_qpn_ds(ctrl, qp_num, ds);
-    ctrl->fm_ce_se         = fm_ce_se;
+    uct_ctrl->dci_stream_channel_id = htons(dci_channel);
```
yosefe, line 381: can dci channel be more than uint8?
if not, endianess swap could be simpler

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -329,91 +337,103 @@ uct_ib_mlx5_set_ctrl_qpn_ds(struct mlx5_wqe_ctrl_seg *ctrl, uint32_t qp_num,
...
+                        4,               /* data size */
                         16,              /* signature (set 0) */
-                        16, 16,          /* reserved (set 0) */
+                        7, 6,            /* dci stream channel ID in BE */
                         0,               /* signal/fence_mode */
-                        8, 9, 10, 11}; /* immediate (set to 0)*/
-    uint32x4_t data = {(opcode << 16) | (opmod << 8) | (uint32_t)fm_ce_se,
-                       (ds << 16) | pi, imm,  qp_num};
+                        8, 9, 10, 11}; /* immediate */
+    uint32x4_t data = {(ds << 16) | (opcode << 16) | (opmod << 8) | fm_ce_se,
```
yosefe, line 404: how does it work that both ds and opcode are shifted by 16?

src/uct/ib/rc/accel/rc_mlx5.inl
```diff
@@ -1687,7 +1694,8 @@ static ucs_status_t UCS_F_ALWAYS_INLINE uct_rc_mlx5_common_ep_short_iov_dm(
                                        hdr_len + iov_length, rdma_raddr,
                                        rdma_rkey, av_size, fm_ce_se, 0, desc,
                                        buffer,
-                                       (log_sge.num_sge > 0) ? &log_sge : NULL);
+                                       (log_sge.num_sge > 0) ? &log_sge : NULL,
+                                       dci_channel);
```
yosefe, line 1698: IMO better add dci_channel after fm_ce_se in all functions since it's also part of the control seg

src/uct/ib/dc/dc_mlx5.c
yosefe: it's the same QP but can use multiple "engines" in the underlying implementation

shamisp: So you don't have to mange QP pull anymore ?

yosefe: Don't have to. But we still want to support older HW generations in the code.

shamisp: Very cool.

src/uct/ib/rc/accel/rc_mlx5.inl
hoopoepg: > IMO better add dci_channel after fm_ce_se in all functions since it's also part of the control seg

ok, but it make changes **much** more complicated for verification  :)

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -351,7 +352,7 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
         attr.full_handshake              = full_handshake;
         attr.rdma_wr_disabled            = (iface->flags & UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT) &&
                                            (md->flags & UCT_IB_MLX5_MD_FLAG_NO_RDMA_WR_OPTIMIZED);
-        attr.log_num_dci_stream_channels = iface->tx.log_num_dci_stream_channels;
+        attr.log_num_dci_stream_channels = iface->tx.num_dci_channels;
```
yosefe, line 355: should probably use ilog2 here
since left side has "log" and right side does not have "log" in the name

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1413,9 +1414,9 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
                          uct_dc_mlx5_iface_dci_do_rand_pending_tx :
                          uct_dc_mlx5_iface_dci_do_dcs_pending_tx;
 
-    self->tx.log_num_dci_stream_channels =
-            ucs_min(ucs_log2(ucs_roundup_pow2(config->max_dci_stream_channels)),
-                    md->log_max_dci_stream_channels);
+    self->tx.num_dci_channels = ucs_min(ucs_log2(ucs_roundup_pow2(
+            ucs_max(config->num_dci_channels, 1))),
```
yosefe, line 1418: maybe use helper vars, hard to follow

src/uct/ib/dc/dc_mlx5_ep.h
yosefe: ok, missed that in random scheme same dci used by multiple eps

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -308,12 +308,12 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 
     if (uct_dc_mlx5_iface_is_dci_rand(iface)) {
         /* coverity[dont_call] */
-        ep->dci                   = rand_r(&iface->tx.rand_seed) % iface->tx.ndci;
-        ep->dci_stream_channel_id = iface->tx.dcis[ep->dci].dci_channel_index++ %
-                                    UCS_BIT(iface->tx.log_num_dci_stream_channels);
+        ep->dci               = rand_r(&iface->tx.rand_seed) % iface->tx.ndci;
+        ep->dci_channel_index = iface->tx.dcis[ep->dci].next_channel_index++ &
+                                UCS_MASK(iface->tx.num_dci_channels);
```
yosefe, line 313: can we precompute `UCS_MASK(iface->tx.num_dci_channels)` and save on the iface? instead of saving tx.num_dci_channels

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -949,7 +949,8 @@ static ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
 
     md->log_max_dci_stream_channels = UCT_IB_MLX5DV_GET(cmd_hca_cap, cap,
                                                         log_max_dci_stream_channels);
-    ucs_assert(md->log_max_dci_stream_channels < 8);
+    md->log_max_dci_stream_channels = ucs_min(md->log_max_dci_stream_channels,
+                                              8);
```
yosefe, line 953: why 8? should we define some constant?

src/uct/ib/mlx5/ib_mlx5.inl
yosefe: ok

src/uct/ib/mlx5/ib_mlx5.inl
yosefe: so shall we change the parameter type of "dci_channel" from uint16_t to uint8_t?

src/uct/ib/dc/dc_mlx5.c
hoopoepg: > should probably use ilog2 here
> since left side has "log" and right side does not have "log" in the name

hmmm, `iface->tx.num_dci_channels` stores log-value of channels number... ok, it will store `real` value

src/uct/ib/dc/dc_mlx5.c
hoopoepg: > maybe use helper vars, hard to follow

added helper value, simplified

src/uct/ib/dc/dc_mlx5_ep.h
hoopoepg: > can we precompute `UCS_MASK(iface->tx.num_dci_channels)` and save on the iface? instead of saving tx.num_dci_channels

this code is not performance critical, do you think storing pre-calculated value is reasonable here?
not actual anymore: used non-log value and `%` operation

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
hoopoepg: > why 8? should we define some constant?

added `UCT_IB_MLX5_MD_MAX_DCI_CHANNELS` constant

src/uct/ib/mlx5/ib_mlx5.inl
hoopoepg: > so shall we change the parameter type of "dci_channel" from uint16_t to uint8_t?

ctrl segment accepts 16 bit value for dci channel, so it expects big endian value. we may calculate it by storing BE value in EP, but it will require one more byte to store

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1414,9 +1415,11 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
                          uct_dc_mlx5_iface_dci_do_rand_pending_tx :
                          uct_dc_mlx5_iface_dci_do_dcs_pending_tx;
 
-    self->tx.num_dci_channels = ucs_min(ucs_log2(ucs_roundup_pow2(
-            ucs_max(config->num_dci_channels, 1))),
-            md->log_max_dci_stream_channels);
+    /* calculate num_dci_channels: select minimum from requested by runtime
+     * and supported by HCA, must be power of two */
+    num_dci_channels = ucs_roundup_pow2(ucs_max(config->num_dci_channels, 1)),
```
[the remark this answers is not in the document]
hoopoepg: > ; at end of line

yep, fixed

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -950,7 +953,7 @@ static ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
     md->log_max_dci_stream_channels = UCT_IB_MLX5DV_GET(cmd_hca_cap, cap,
                                                         log_max_dci_stream_channels);
     md->log_max_dci_stream_channels = ucs_min(md->log_max_dci_stream_channels,
-                                              8);
+                                              UCT_IB_MLX5_MD_MAX_DCI_CHANNELS);
```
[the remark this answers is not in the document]
hoopoepg: > maybe static assert that UCS_MASK(UCT_IB_MLX5_MD_MAX_DCI_CHANNELS) <= UINT8_MAX ?

ok, added static assert

src/uct/ib/mlx5/ib_mlx5.inl
hoopoepg: > can you check if the compiler really eliminated the htons ?

it seems GCC has no ideas how to optimise this (called tons via plt section). added macro to convert byte into network datatype

btw, we may create own implementation for hton*** calls via compiler intrinsics

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -378,7 +378,7 @@ uct_ib_mlx5_set_ctrl_seg(struct mlx5_wqe_ctrl_seg* ctrl, uint16_t pi,
 #else
     uct_ctrl->opmod_idx_opcode      = (opcode << 24) | (htons(pi) << 8) | opmod;
     uct_ib_mlx5_set_ctrl_qpn_ds(ctrl, qp_num, ds);
-    uct_ctrl->dci_stream_channel_id = htons(dci_channel);
+    uct_ctrl->dci_stream_channel_id = ucs_btons(dci_channel);
```
yosefe, line 381: maybe just `htons((uint8_t)dci_channel)`

hoopoepg: > maybe just `htons((uint8_t)dci_channel)`

doesn't work:
objdump -S modules/libuct_ib.so
```
 uct_ctrl->dci_stream_channel_id = htons((uint8_t)dci_channel);              
5084a:   0f b7 85 24 fd ff ff    movzwl -0x2dc(%rbp),%eax                    
50851:   0f b6 c0                movzbl %al,%eax                             
50854:   89 c7                   mov    %eax,%edi                            
50856:   e8 65 7d fc ff          callq  185c0 <htons@plt>                    
```

yosefe: weird that htons is not even inlined.. maybe missing some optimizations?

hoopoepg: it seems it depends from optimisation level... I tried to build UCX with -O2 switch and compiler excluded all calls to htons at all

yosefe (dismissed):

yosefe:
@hoopoepg can you pls squash?

brminich (approved):

PR #8450, opened 2022-08-15
UCS/DEBUG: replace PTR with void *

hzhou:
The PTR macro is missing on the latest Arch linux.

## What
Replace the cast `(PTR)` with `(void *)` in `debug.c`.

## Why ?
The `PTR` macro is not defined in the latest Arch Linux bfd headers.

See reported error in the comment in https://aur.archlinux.org/packages/openucx

## How ?
`PTR` is probably a pointer to a specific type. Since we are casting here, casting to `void *` should work.

### DIFF
--- a/src/ucs/debug/debug.c
+++ b/src/ucs/debug/debug.c
@@ -272,10 +272,10 @@ static int load_file(struct backtrace_file *file)
         goto err_close;
     }
 
-    symcount = bfd_read_minisymbols(file->abfd, 0, (PTR)&file->syms, &size);
+    symcount = bfd_read_minisymbols(file->abfd, 0, (void *)&file->syms, &size);
     if (symcount == 0) {
         free(file->syms);
-        symcount = bfd_read_minisymbols(file->abfd, 1, (PTR)&file->syms, &size);
+        symcount = bfd_read_minisymbols(file->abfd, 1, (void *)&file->syms, &size);
     }
     if (symcount < 0) {
         goto err_close;


jedbrown:
This change comes from binutils-2.39. There were some commits this spring saying that `PTR` would be removed (https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=845cbaa9ffbfd6a1f7976a6c7f3e4461e4d41993) in imminent syncs with gcc. The commit that actually removed `PTR` from `ansidecl.h` came [here](https://sourceware.org/git/?p=binutils-gdb.git;a=blobdiff;f=include/ansidecl.h;h=46fe3ffabd9ec3093385027513cdcbcb331ad395;hp=efee5b6904b3593e36159cd4c09f6cb6e45b577e;hb=31b15688c414c7caf957be63d2914faafa1b9dda;hpb=845cbaa9ffbfd6a1f7976a6c7f3e4461e4d41993).

yosefe:
@hzhou thank you for your contribution! UCX requires signing a CLA in order to submit code to upstream (see https://github.com/openucx/ucx/wiki/Guidance-for-contributors).
cc @shamisp

yosefe (approved):

hzhou:
@yosefe I am a member of Argonne MPICH team. Is there a CLA for us in place already? I recall we have contributed in the past without needing to sign anything.

shamisp:
ANL has signed CLA

PR #8452, opened 2022-08-15
GTEST: fixed build with -O1 optimization

hoopoepg:
- there is build issue when gtest is built with -O1 optimization
  flag:
  make[3]: Entering directory `/auto/mtrswgwork/sergeyo/vs-projects/ucx/test/gtest'
  CXX      uct/gtest-test_tag.o
In file included from uct/test_tag.cc:8:0:
/labhome/sergeyo/ucx/src/uct/api/uct.h: In member function ‘virtual void test_tag_tag_hold_uct_desc_Test::test_body()’:
/labhome/sergeyo/ucx/src/uct/api/uct.h:2725:21: error: inlining failed in call to always_inline ‘void uct_iface_release_desc(void*)’: indirect function call with a yet undetermined callee
 UCT_INLINE_API void uct_iface_release_desc(void *desc)
                     ^
In file included from /usr/include/c++/4.8.2/algorithm:62:0,
                 from ./common/googletest/internal/gtest-port.h:264,
                 from ./common/googletest/internal/gtest-internal.h:40,
                 from ./common/googletest/gtest.h:62,
                 from ./common/test_helpers.h:11,
                 from ./common/test.h:23,
                 from uct/test_tag.cc:10:

- added proxy call to hide force inline attribute

### DIFF
--- a/test/gtest/uct/test_tag.cc
+++ b/test/gtest/uct/test_tag.cc
@@ -567,7 +567,7 @@ UCS_TEST_SKIP_COND_P(test_tag, tag_hold_uct_desc,
     }
 
     std::for_each(m_uct_descs.begin(), m_uct_descs.end(),
-                  uct_iface_release_desc);
+                  [](void *desc) -> void {uct_iface_release_desc(desc);});
 
 }
 
@@ -1242,7 +1242,7 @@ UCS_TEST_P(test_tag_mp_xrq, desc_release)
     }
 
     std::for_each(m_uct_descs.begin(), m_uct_descs.end(),
-                  uct_iface_release_desc);
+                  [](void *desc) -> void {uct_iface_release_desc(desc);});
 }
 
 UCS_TEST_P(test_tag_mp_xrq, am)


test/gtest/uct/test_tag.cc
```diff
@@ -14,6 +14,12 @@ extern "C" {
     _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5) \
     _UCT_INSTANTIATE_TEST_CASE(_test_case, dc_mlx5)
 
+
+static void iface_release_desc(void *desc) {
```
yosefe, line 18: add comment why

hoopoepg: > add comment why

ok, added comment

test/gtest/uct/test_tag.cc
```diff
@@ -14,6 +14,14 @@ extern "C" {
     _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5) \
     _UCT_INSTANTIATE_TEST_CASE(_test_case, dc_mlx5)
 
+
+/* proxy function for uct_iface_release_desc to eliminate
+ * errors related to failed of inlining uct_iface_release_desc */
```
yosefe, line 19: Wrapper for uct_iface_release_desc() to eliminate errors of failing to inline it when passed to std::for_each
BTW, can we use lambda expression in std::for_each instead of named wrapper?

hoopoepg: > BTW, can we use lambda expression in std::for_each instead of named wrapper?

as I remember we have minimal supported C++ standard C11, I believe we can use lambdas (not sure if GCC 4.4 will compile it)

yosefe (approved):

PR #8173, opened 2022-04-27
UCP/AM: Support sending AM with non persistent user header - Implementation

ofirfarjun7:
Implementation of the new API feature

## What
Support non persistence user header buffer during sending AM.

## Why ?
Current implementation dictate that "header must be valid until the operation completes",
the new API enhancement free the user from this obligation, he can reuse/free the header after
ucp_am_send_nbx returns.

## How ?
Copy the provided user header to UCX internal buffer when it's needed (short/bcopy/RNDV) until operation completes.

PR #8455, opened 2022-08-15
UCP/PROTOV2/EAGER/SHORT: Bug fix - iov array is initialized with a wrong length

ofirfarjun7:
## What
 - Fix a bug in proto V2 "short" send function that can cause segmentation fault 
 - Check why gtest missed it and make sure it won't happen again

## Why ?
- In "ucp_am_eager_short_proto_progress_common" iov array is initialized with size 3, but when sending AM
with user header and reply EP iov list contains 4 elements.
- @rakhmets found that gtest missed it because of the order we are running the tests. The test first sends message without header, and the message with the header is sent after, when wireup is completed, therefore the message with the header is sent using old proto, ucp_am_send_short, and there the iov list length is correct. 

## How ?
- Fix the iov length to be 4 instead of 3.
- Decide how to change the tests so we will not have another miss.

### DIFF
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -432,13 +432,13 @@ protected:
         EXPECT_EQ(check_pattern, m_hdr);
     }
 
-    ucs_status_ptr_t send_am(const ucp::data_type_desc_t &dt_desc,
-                             unsigned flags = 0, const void *hdr = NULL,
-                             unsigned hdr_length = 0,
-                             const ucp_mem_h memh = NULL)
+    ucs_status_ptr_t
+    send_am(const ucp::data_type_desc_t &dt_desc, unsigned flags = 0,
+            const void *hdr = NULL, unsigned hdr_length = 0,
+            const ucp_mem_h memh = NULL, uint32_t op_attr_mask = 0)
     {
         ucp_request_param_t param;
-        param.op_attr_mask      = UCP_OP_ATTR_FIELD_DATATYPE;
+        param.op_attr_mask      = op_attr_mask | UCP_OP_ATTR_FIELD_DATATYPE;
         param.datatype          = dt_desc.dt();
 
         if (flags != 0) {
@@ -458,13 +458,14 @@ protected:
     }
 
     void test_am_send_recv(size_t size, size_t header_size = 0ul,
-                           unsigned flags = 0, unsigned data_cb_flags = 0)
+                           unsigned flags = 0, unsigned data_cb_flags = 0,
+                           uint32_t op_attr_mask = 0)
     {
         mem_buffer sbuf(size, tx_memtype());
         sbuf.pattern_fill(SEED);
         m_hdr.resize(header_size);
         ucs::fill_random(m_hdr);
-        m_am_received = false;
+        m_am_received  = false;
         ucp_mem_h memh = NULL;
 
         set_am_data_handler(receiver(), TEST_AM_NBX_ID, am_data_cb, this,
@@ -477,7 +478,8 @@ protected:
         }
 
         ucs_status_ptr_t sptr = send_am(sdt_desc, get_send_flag() | flags,
-                                        m_hdr.data(), m_hdr.size(), memh);
+                                        m_hdr.data(), m_hdr.size(), memh,
+                                        op_attr_mask);
 
         wait_for_flag(&m_am_received);
         request_wait(sptr);
@@ -847,13 +849,8 @@ UCS_TEST_P(test_ucp_am_nbx, rx_am_mpools,
 UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx)
 
 
-class test_ucp_am_nbx_single_multi_reply : public test_ucp_am_nbx {
+class test_ucp_am_nbx_reply_always : public test_ucp_am_nbx {
 public:
-    test_ucp_am_nbx_single_multi_reply()
-    {
-        modify_config("RNDV_THRESH", "inf");
-    }
-
     bool am_lane_has_caps(uint64_t caps);
 
 protected:
@@ -863,7 +860,7 @@ protected:
     }
 };
 
-bool test_ucp_am_nbx_single_multi_reply::am_lane_has_caps(uint64_t caps)
+bool test_ucp_am_nbx_reply_always::am_lane_has_caps(uint64_t caps)
 {
     ucp_ep_config_key_t key  = ucp_ep_config(sender().ep())->key;
     ucp_lane_index_t am_lane = key.am_lane;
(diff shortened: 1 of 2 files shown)


rakhmets (approved):

brminich:
👍 ,
can you please make a gtest?

ofirfarjun7:
Added gtest

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -863,6 +863,55 @@ UCS_TEST_P(test_ucp_am_nbx_send_flag, eager, "RNDV_THRESH=128")
 UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx_send_flag)
 
 
+class test_ucp_am_nbx_short_slow_path : public test_ucp_am_nbx {
```
rakhmets, line 866: You can use this class instead: `test_ucp_am_nbx_reply`

ofirfarjun7: I thought to use it but in this case the test will run with more variants... I thought to save time but also can change .

rakhmets: It will add only `prereg` variant, which I will move from base class to inherited classes.

ofirfarjun7: Yes. will run 8 tests for each TL instead of 4.
I will take your advice unless someone else disagree.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -863,6 +863,55 @@ UCS_TEST_P(test_ucp_am_nbx_send_flag, eager, "RNDV_THRESH=128")
...
+    virtual unsigned get_send_flag() const
+    {
+        return get_variant_value(0);
+    }
+};
+
+UCS_TEST_P(test_ucp_am_nbx_short_slow_path, short_slow_path)
+{
+    size_t header_size = 8;
+    size_t data_size   = 8;
```
yosefe, line 891: const

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -863,6 +863,55 @@ UCS_TEST_P(test_ucp_am_nbx_send_flag, eager, "RNDV_THRESH=128")
...
+    mem_buffer sbuf(data_size, tx_memtype());
+    sbuf.pattern_fill(SEED);
+    m_hdr.resize(header_size);
+    ucs::fill_random(m_hdr);
+    m_am_received = false;
+    ucs_status_ptr_t sptr;
```
yosefe, line 897: better declare and initialize variables right before they are used
for example:
`ucs_status_ptr_t sptr = ...`

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -863,6 +863,55 @@ UCS_TEST_P(test_ucp_am_nbx_send_flag, eager, "RNDV_THRESH=128")
...
+
+    set_am_data_handler(receiver(), TEST_AM_NBX_ID, am_data_cb, this);
+
+    ucp::data_type_desc_t sdt_desc(m_dt, sbuf.ptr(), data_size);
+    /*
```
yosefe, line 902: space line before

rakhmets (dismissed):

ofirfarjun7:
I was asked to reduce the number of variants to the minimum.
This test always run with return flag and the only variant decide between ProtoV1/ProtoV2

rakhmets:
> I was asked to reduce the number of variants to the minimum. This test always run with return flag and the only variant decide between ProtoV1/ProtoV2

There is a PR: https://github.com/openucx/ucx/pull/8480. Once the PR will be merged, than the only variant in base class will be protoV1 vs protoV2.

There is a class `test_ucp_am_nbx_single_multi_reply` which may be used as test class for your case.

You can do the following:
1. rename `test_ucp_am_nbx_single_multi_reply` to `test_ucp_am_nbx_reply_always`. Because your name is more appropriate name for the test class.
2. test bcopy and zcopy shouldbe renamed to multi_bcopy, multi_zcopy then.
3. you don't need to introduce a new class. and just add your test `short_slow_path`

rakhmets:
> Sounds reasonable to me but it will require a change after merging with #8480: **get_send_flag should be change to return get_variant_value(1) instead of get_variant_value(2).** in test_ucp_am_nbx_reply_always.
> 
> Also this will make #8173 dependent in both PR's, this one and #8480. Will wait for approval before changing the code.

There is no need to change `get_send_flag` because:
```C
class test_ucp_am_nbx_single_multi_reply : public test_ucp_am_nbx {
...
protected:
    virtual unsigned get_send_flag() const
    {
        return UCP_AM_SEND_FLAG_REPLY;
    }
};
```

ofirfarjun7:
You are right.
The problem I see in this approach is that you are not adding reply to the variants, and the only thing that show you that this test is running with reply flag is the test name, but what will happen if another test will inherit from it?
In this case you can't tell from the test logs that it's running with reply unless you mention it in the new test name or look in the code...

For this reason I think it is better to add reply as a constant variant and use it in get_send_flag.

ofirfarjun7:
After the merge of #8480 tests variants will be reduced to the minimum required.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -900,11 +910,16 @@ UCS_TEST_P(test_ucp_am_nbx_single_multi_reply, zcopy, "ZCOPY_THRESH=1")
...
- * be less than (fragment size + reply footer size) to check if the switch
- * between single and multi protocols is correct. */
-UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx_single_multi_reply)
+UCS_TEST_P(test_ucp_am_nbx_reply_always, short_slow_path)
+{
+    /*
+     * This message is sent with UCP_OP_ATTR_FLAG_NO_IMM_CMPL so it will
+     * always go through AM short slowpath.
+     */
+    test_am_send_recv(8, 8, get_send_flag(), 0, UCP_OP_ATTR_FLAG_NO_IMM_CMPL);
```
rakhmets, line 919: There is no need to call to `get_send_flag` here. `get_send_flag` is called inside `test_am_send_recv` when `send_am` is called.

rakhmets:
> You are right. The problem I see in this approach is that you are not adding reply to the variants, and the only thing that show you that this test is running with reply flag is the test name, but what will happen if another test will inherit from it? In this case you can't tell from the test logs that it's running with reply unless you mention it in the new test name or look in the code...
> 
> Your approach is more safe and simple I think, but can provide less information. Edit: I can still add it to the variants and always return UCP_AM_SEND_FLAG_REPLY.

If someone wants to inherit the class, then it will be done to inherit the implementation of `get_send_flag`. Because this is the only thing that overridden in `test_ucp_am_nbx_reply_always`. Thus if someone use this class as base, the name of the inherited class will probably contain `reply`.

ofirfarjun7:
> Thus if someone use this class as base, the name of the inherited class will probably contain reply

It is likely but still it's an assumption, for example when you used this test before I change the name it didn't contain reply... I guess it's can happen again.
I don't mind removing it though.

rakhmets:
> > Thus if someone use this class as base, the name of the inherited class will probably contain reply
> 
> It is likely but still it's an assumption, for example when you used this test before I change the name it didn't contain reply... I guess it's can happen again. I don't mind removing it though.

Yes, class name did not, but test cases did.
And after your changes, the class name became clearer. 👍

rakhmets (approved):

yosefe (dismissed):

yosefe:
@ofirfarjun7 pls squash

yosefe (approved):

PR #8464, opened 2022-08-17
UCM/BISTRO: Implement cuda bistro hooks for aarch64

yosefe:


test/gtest/ucm/malloc_hook.cc
```diff
@@ -1046,12 +1046,13 @@ UCS_TEST_F(malloc_hook_cplusplus, remap_override_multi_threads) {
 typedef int (munmap_f_t)(void *addr, size_t len);
 
 UCS_TEST_SKIP_COND_F(malloc_hook, bistro_patch, RUNNING_ON_VALGRIND) {
+    const size_t max_patch_size = 32;
```
yosefe, line 1049: needed to make test pass.
before this PR, the first 64bit of the patch were affected by hook address
after this PR, 64bit of the patch are same for all hooks (ldr x15 ; br x15) - need to compare more bytes so `EXPECT_NE(patched, origin);` in line 1088 will not fail

src/ucm/bistro/bistro.c
```diff
@@ -154,4 +155,60 @@ void *ucm_bistro_allocate_code(size_t size)
     return result;
 }
 
+static const char *
+ucm_bistro_dump_code(const void *code, size_t length, char *str, size_t max)
```
yosefe, line 159: moved from x86 to common code

src/ucm/bistro/bistro.c
```diff
@@ -154,4 +155,60 @@ void *ucm_bistro_allocate_code(size_t size)
...
+        snprintf(p, endp - p, " %02X",
+                 *ucs_serialize_next(&code_p, const uint8_t));
+        p += strlen(p);
+    }
+
+    return str;
+}
+
+ucs_status_t
+ucm_bistro_relocate_code(void *dst, const void *src, size_t min_src_length,
```
yosefe, line 175: moved from x86 to common code and added custom context pointer 'ctx'

src/ucm/bistro/bistro_aarch64.c
```diff
@@ -50,30 +66,235 @@
...
                               ucm_bistro_restore_point_t **rp)
 {
-    ucm_bistro_patch_t patch = {
-        .reg3 = MOVZ(R15, 3, (uintptr_t)hook >> 48),
-        .reg2 = MOVK(R15, 2, (uintptr_t)hook >> 32),
-        .reg1 = MOVK(R15, 1, (uintptr_t)hook >> 16),
-        .reg0 = MOVK(R15, 0, (uintptr_t)hook),
-        .br   = BR(R15)
-    };
+    ucm_bistro_jmp_indirect_t patch;
```
yosefe, line 287: replaced by a shorted patch (4 instr instead of 5 inst)

hoopoepg (approved):

PR #8473, opened 2022-08-23
UCT/TCP: Fix status returned when disconnected

Artemy-Mellanox:
## What?

Add failed EP status tracking.
Fix status returned when TCP disconnected.

## Why?

For consistency.
To make value returned via UCP EP error callback and values returned from following operations be the same.
To make value returned to client when server closed be same with RDMA and TCP protocols.

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -65,9 +65,12 @@ typedef struct ucp_ep_set_failed_arg {
  * Argument for discarding UCP endpoint's lanes
  */
 typedef struct ucp_ep_discard_lanes_arg {
-    unsigned     counter; /* How many discarding operations on UCT lanes are
-                           * in-progress if purging of the UCP endpoint is
-                           * required */
+    uct_ep_t     failed_ep;
+    unsigned     discard_counter; /* How many discarding operations on UCT
+                                     lanes are in-progress if purging of
+                                     the UCP endpoint is required */
+    unsigned     destroy_counter; /* How many destroy operations on UCT
+                                     will be called */
     ucs_status_t status; /* Completion status of operations after discarding is
                           * done */
     ucp_ep_h     ucp_ep; /* UCP endpoint which should be discarded */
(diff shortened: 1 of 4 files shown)


yosefe (dismissed):

dmitrygx (approved):

yosefe:

src/ucp/core/ucp_ep.c
```diff
@@ -1287,6 +1306,31 @@ static void ucp_ep_discard_lanes_callback(void *request, ucs_status_t status,
...
 }
 
+static ucs_status_t ucp_ep_failed_op(uct_ep_h ep)
+{
+    ucp_ep_failed_arg_t *arg = ucs_derived_of(ep, ucp_ep_failed_arg_t);
+
+    return arg->status;
+}
+
+static ssize_t ucp_ep_failed_bc_op(uct_ep_h ep)
```
dmitrygx, line 1316: can we use `ucp_ep_failed_op` for Bcopy too? anyway, we do explicit casting.

src/ucp/core/ucp_ep.c
```diff
@@ -1307,6 +1352,16 @@ static void ucp_ep_discard_lanes(ucp_ep_h ep, ucs_status_t discard_status)
         return;
     }
 
+    failed_arg = ucs_malloc(sizeof(*failed_arg), "failed_arg");
+    if (failed_arg == NULL) {
+        failed_arg = &ucp_failed_tl_ep;
+    } else {
```
dmitrygx, line 1358: why need to handle it this way? if this memory allocation fails, then memory allocation below will probably fail too.
wdyt about just doing the same actions as we have when handling failed memory allocation for `discard_arg` below?

src/ucp/core/ucp_ep.c
```diff
@@ -74,6 +74,17 @@ typedef struct ucp_ep_discard_lanes_arg {
 } ucp_ep_discard_lanes_arg_t;
 
 
+/**
+ * Failed UCT EP placeholder
+ */
+typedef struct {
+    uct_ep_t     super;
+    ucs_status_t status;
```
evgeny-leksikov, line 82: it seems like shared status across many EPs also can be wrong

evgeny-leksikov (commented):
@Artemy-Mellanox pls fill PR description, from the patch it's not clear why it's needed and which problem it fixes

src/ucp/core/ucp_ep.c
```diff
@@ -74,6 +74,17 @@ typedef struct ucp_ep_discard_lanes_arg {
...
+    unsigned     counter;
+    int          free;
```
yosefe, line 84: compare w global var instead

src/ucp/core/ucp_ep.c
yosefe: no, since return type is different size, ucp_ep_failed_op would set only lower portion of the return value register rax on x86_64

src/ucp/core/ucp_ep.c
```diff
@@ -1287,6 +1306,31 @@ static void ucp_ep_discard_lanes_callback(void *request, ucs_status_t status,
...
+    ucp_ep_failed_arg_t *arg = ucs_derived_of(ep, ucp_ep_failed_arg_t);
+
+    return arg->status;
+}
+
+static void ucp_ep_failed_destroy(uct_ep_h ep)
+{
+    ucp_ep_failed_arg_t *arg = ucs_derived_of(ep, ucp_ep_failed_arg_t);
+
+    if (!arg->free || (--arg->counter > 0)) {
```
yosefe, line 1327: according to ucp_worker_discard_uct_ep_destroy_progress, we first destroy uct_ep and then release discard_arg, so should be able to place the failed ep on discard_arg and not use a separate counter

src/ucp/core/ucp_ep.c
yosefe: the first failed lane should determine the error code on the whole ucp ep

evgeny-leksikov: But as far as I can see it's static instance which is shared across all UCP EPs per process

yosefe: in case of discard, there is another instance that is dynamically-allocated

src/ucp/core/ucp_ep.c
```diff
@@ -74,6 +74,17 @@ typedef struct ucp_ep_discard_lanes_arg {
 } ucp_ep_discard_lanes_arg_t;
 
 
+/**
+ * Failed UCT EP placeholder
+ */
+typedef struct {
+    uct_ep_t     super;
+    ucs_status_t status;
+    unsigned     counter;
```
evgeny-leksikov, line 83: maybe use uct completion? it already has counter + status and has suitable API, static instance could has empty completion func, and `free(arg)` for EP allocated instance.

src/ucp/core/ucp_ep.c
```diff
@@ -65,9 +65,12 @@ typedef struct ucp_ep_set_failed_arg {
  * Argument for discarding UCP endpoint's lanes
  */
 typedef struct ucp_ep_discard_lanes_arg {
-    unsigned     counter; /* How many discarding operations on UCT lanes are
-                           * in-progress if purging of the UCP endpoint is
-                           * required */
+    uct_ep_t     super;
```
yosefe, line 68: rename to failed_ep

src/ucp/core/ucp_ep.c
```diff
@@ -91,49 +94,54 @@ static ucs_stats_class_t ucp_ep_stats_class = {
...
-        .ep_get_address      = (uct_ep_get_address_func_t)ucs_empty_function_return_ep_timeout
+        .ep_connect_to_ep    = (uct_ep_connect_to_ep_func_t)ucp_ep_failed_op,
+        .ep_destroy          = ucp_ep_failed_destroy,
+        .ep_get_address      = (uct_ep_get_address_func_t)ucp_ep_failed_op
     }
 };
 
-static uct_ep_t ucp_failed_tl_ep = {
-    .iface = &ucp_failed_tl_iface
+static ucp_ep_discard_lanes_arg_t ucp_failed_tl_ep = {
```
yosefe, line 136: ucp_failed_tl_ep_discard_arg

src/ucp/core/ucp_ep.c
yosefe: but seems there is no completion callback needed here?

evgeny-leksikov: we can invoke it manually like we do it in ucp_flush, for example.
this is not a critical to me, just an idea how to reuse existing code for similar functionality:
- it has a counter
- it stores a first happening error status
- it has a function pointer which can be defined depends on 2 different cases

src/ucp/core/ucp_ep.c
yosefe: I've meant to rename "super" to "failed_ep"..

yosefe (dismissed):
pls squash

yosefe (approved):

PR #8480, opened 2022-08-29
GTEST/UCP: Moved prereg variant to inherited classes.

rakhmets:
## What
Moved memory preregistration variant of test from base class to inherited classes.

## Why ?
To get rid of redundant test variants.

### DIFF
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -38,18 +38,10 @@ public:
         }
     }
 
-    static void
-    get_test_variants_prereg(std::vector<ucp_test_variant> &variants)
+    static void get_test_variants(variant_vec_t &variants)
     {
         add_variant_with_value(variants, UCP_FEATURE_AM, 0, "");
-        add_variant_with_value(variants, UCP_FEATURE_AM, TEST_FLAG_PREREG,
-                               "prereg");
-    }
-
-    static void get_test_variants(std::vector<ucp_test_variant> &variants)
-    {
-        add_variant_values(variants, get_test_variants_prereg, 0);
-        add_variant_values(variants, get_test_variants_prereg, 1, "proto");
+        add_variant_with_value(variants, UCP_FEATURE_AM, 1, "proto");
     }
 
     virtual void init() {
@@ -60,20 +52,10 @@ public:
         receiver().connect(&sender(), get_ep_params());
     }
 
-protected:
-    enum {
-        TEST_FLAG_PREREG = UCS_BIT(0)
-    };
-
-    bool prereg() const
-    {
-        return get_variant_value(0) & TEST_FLAG_PREREG;
-    }
-
 private:
     bool is_proto_enabled() const
     {
-        return get_variant_value(1);
+        return get_variant_value();
     }
 };
 
@@ -638,6 +620,11 @@ protected:
         self->am_recv_check_data(length);
     }
 
+    virtual bool prereg() const
+    {
+        return 0;
+    }
+
     static const uint16_t           TEST_AM_NBX_ID = 0;
     ucp_datatype_t                  m_dt;
     volatile bool                   m_am_received;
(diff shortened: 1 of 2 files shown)


brminich (approved):

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -30,6 +30,9 @@ extern "C" {
 #define UCP_RELEASE 1
 
 class test_ucp_am_base : public ucp_test {
+protected:
+    using variant_vec_t = std::vector<ucp_test_variant>;
```
yosefe, line 34: maybe move it to ucp_test? or in another PR?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1037,15 +1034,37 @@ UCS_TEST_P(test_ucp_am_nbx_closed_ep, rx_rts_am_on_closed_ep, "RNDV_THRESH=32K")
 UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx_closed_ep)
 
 
-class test_ucp_am_nbx_eager_memtype : public test_ucp_am_nbx {
+class test_ucp_am_nbx_prereg : public test_ucp_am_nbx {
+public:
+    static void get_test_variants(variant_vec_t &variants)
+    {
+        add_prereg_variants(variants, test_ucp_am_nbx::get_test_variants);
```
yosefe, line 1041: maybe move add_prereg_variants to "test_ucp_am_nbx_prereg" class 
and if need, inherit test_ucp_am_nbx_reply also from test_ucp_am_nbx_prereg
it's weird that we have some "prereg" logic leftover in test_ucp_am, that does not really use it

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -29,10 +29,10 @@ extern "C" {
 #define UCP_REPLY_ID 1
 #define UCP_RELEASE 1
 
-class test_ucp_am_base : public ucp_test {
-protected:
-    using variant_vec_t = std::vector<ucp_test_variant>;
+using variant_vec_t = std::vector<ucp_test_variant>;
```
yosefe, line 32: can we move it to ucp_test class?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -55,6 +55,11 @@ class test_ucp_am_base : public ucp_test {
         receiver().connect(&sender(), get_ep_params());
     }
 
+    virtual unsigned int get_num_variants() const
```
yosefe, line 58: virtual unsigned num_variants() const

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -925,22 +923,56 @@ UCS_TEST_P(test_ucp_am_nbx_send_flag, eager, "RNDV_THRESH=128")
...
+    {
+        return T::get_num_variants() + 1;
     }
 };
 
+using test_ucp_am_nbx_reply        = test_ucp_am_nbx_reply_var<test_ucp_am_nbx>;
+using test_ucp_am_nbx_prereg       =
+        test_ucp_am_nbx_prereg_var<test_ucp_am_nbx>;
+using test_ucp_am_nbx_prereg_reply =
+        test_ucp_am_nbx_prereg_var<test_ucp_am_nbx_reply>;
```
yosefe, line 971: IMO it's over complicated.. can't we use simple inheritance?

test/gtest/ucp/test_ucp_am.cc
rakhmets: Yes, we can.
Many files should be changed then.
I think it is better to do in separate review.

yosefe: Can we move it there but not change existing places yet?

test/gtest/ucp/test_ucp_am.cc
rakhmets: I'm trying to get rid of code duplication here.
Because in some classes we need only reply or prereg variant, but eager_datatype need both.

This can be replaced by three classes.

yosefe: maybe have one class but enumerate variants by flags, and not necessarily add all combinations, but only interesting ones?
for example:
add_variant(eager)
add_variant(eager|prereg)
add_variant(eager|reply)
...

rakhmets: and we will need to do this in every class that has variants.
it would be many duplicated lines, each class will have add_variant(proto).

I tried this approach, but many extra code was required.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -928,22 +920,53 @@ UCS_TEST_P(test_ucp_am_nbx_send_flag, eager, "RNDV_THRESH=128")
...
+        return T::get_variant_value(T::get_num_variants());
+    }
+
+    virtual unsigned get_num_variants() const
+    {
+        return T::get_num_variants() + 1;
     }
 };
 
+using test_ucp_am_nbx_reply  = test_ucp_am_nbx_reply_var<test_ucp_am_nbx>;
```
yosefe, line 965: nbx -> prereg (pull up tests from eager/rndv)-> {eager_memtype, rndv -> rndv/memtype}
nbc -> reply -> dts

brminich (approved):

yosefe (dismissed):

yosefe (approved):

PR #8488, opened 2022-08-31
UCP/AM: Removed check for min IOV in short protocols.

rakhmets:
## What
Removed check for minimum number of supported IOV in Active Message short protocols (non-reply and reply).

## Why ?
Active Message short reply protocol may use 4 IOV elements: AM header, payload, user header, reply footer.
RCX/DCX/UDX only support 3 IOVs. Thus, we cannot select 'short reply' protocol for small messages, and choose 'bcopy reply' instead. However, limitation of maximum number of IOVs makes sense for 'zcopy' protocols. But for 'short' protocols we can remove the check due to transports capabilities.

### DIFF
--- a/src/ucp/am/eager_single.c
+++ b/src/ucp/am/eager_single.c
@@ -87,7 +87,7 @@ ucp_am_eager_short_proto_progress_common(uct_pending_req_t *self, int is_reply)
 
 static ucs_status_t
 ucp_am_eager_short_proto_init_common(const ucp_proto_init_params_t *init_params,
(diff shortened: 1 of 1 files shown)


brminich (approved):
number of iovs is relevant for verbs transports, but they support up to 5 IOVs by default, which is enough for AM short.
👍 , because there seems to be no way to clearly distinguish verbs and PRM transports on UCP level

ofirfarjun7 (approved):
Good catch.

yosefe (approved):

PR #8491, opened 2022-09-01
UCT/ROCM: remove rocm/gdr component

edgargabriel:
## What
remove the rocm/gdr component

## Why ?
there is no known use case that  required gdrcopy for rocm memory at this time. In addition, the rocm/gdr component only works with an ancient version of gdrcopy (1.3)


yosefe (approved):

nope8:
@edgargabriel Hello,Can you help to answer my confused question: 
Since cuda gdrcopy provide support: After use gdr_pin_buffer, gdr_map operation to  device memory buffer allocated by cudaMalloc API ，App can directly do load/store operation to the device memory buffer bypass cuda runtime/driver to reduce access latency. 
If remove rocm_gdr， how rocm implement the respond function to reduce access latency?

edgargabriel:
@nope8 AMD GPUs require that large BAR support has been enabled on the system, and hence the entire GPU address space is mapped to the CPU address space. Consequently, you can do a regular memcpy() from the CPU side to any GPU address, which achieves basically the same effect as gdrcopy.

See in the source code how rocm_copy_ep_{put, get}_short are implemented https://github.com/openucx/ucx/blob/master/src/uct/rocm/copy/rocm_copy_ep.c#L281

PR #8492, opened 2022-09-02
UCT/IB/DC: Always remove EPs from FC hash when restoring FC window

dmitrygx:
## What

Always remove EPs from FC hash when restoring FC window.

## Why ?

1. If some endpoint was failed and its FC window was restored, it could still exist on FC hash to resend `FC_HARD_REQ`s, but it doesn't make sense anymore.
2. If some endpoint was `flush(CANCEL)`ed, it could still exist on FC hash to resend `FC_HARD_REQ`s with exhausted FC window (since we don't restore it), and consume new DCIs for trying to resend `FC_HARD_REQ` packet, but these attempts would fail in case of peer is unreachable. So, it could lead to "timeout waiting for replies" in IO-demo, since DCIs can't be reused by normal not failed endpoints to send their data.

## How ?

1. Introduce `uct_dc_mlx5_fc_restore_wnd` function which gets an iterator from FC hash, removes an element from FC hash updates FC window, and progress pending if the functions is called from `FC_PURE_GRANT` handler.
2. Invoke `uct_dc_mlx5_fc_restore_wnd` in `uct_dc_mlx5_iface_set_ep_failed` prior checking for flags `UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED` and `UCT_DC_MLX5_EP_FLAG_FLUSH_CANCEL` to restore FC window even if `flush(CANCEL)` was done for an endpoint.
3. Add assertion to check that FC window is full in `uct_dc_mlx5_ep_handle_failure` after `uct_dc_mlx5_iface_set_ep_failed`.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.c
+++ b/src/uct/ib/dc/dc_mlx5.c
@@ -1093,6 +1093,22 @@ void uct_dc_mlx5_fc_entry_iter_del(uct_dc_mlx5_iface_t *iface, khiter_t it)
     }
 }
 
+static UCS_F_ALWAYS_INLINE int
+uct_dc_mlx5_fc_remove_ep(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep,
+                         uint64_t seq)
+{
+    khiter_t it = kh_get(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash,
+                         (uint64_t)ep);
+    if ((it != kh_end(&iface->tx.fc_hash)) &&
+        ((seq == UINT64_MAX) ||
+         (kh_value(&iface->tx.fc_hash, it).seq == seq))) {
+        uct_dc_mlx5_fc_entry_iter_del(iface, it);
+        return 1;
+    }
+
+    return 0;
+}
+
 static ucs_status_t
 uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
                              uct_rc_hdr_t *hdr, unsigned length,
@@ -1105,7 +1121,6 @@ uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
     int16_t cur_wnd;
     ucs_status_t status;
     uct_dc_mlx5_ep_t *ep;
-    khiter_t it;
     ucs_arbiter_t *waitq;
     ucs_arbiter_group_t *group;
     uint8_t pool_index;
@@ -1138,31 +1153,23 @@ uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
                      ucs_status_string(status));
         }
     } else if (fc_hdr == UCT_RC_EP_FC_PURE_GRANT) {
-        sender = (uct_dc_fc_sender_data_t *)(hdr + 1);
-
-        it = kh_get(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash, sender->ep);
-        if ((it == kh_end(&iface->tx.fc_hash)) ||
-            (kh_value(&iface->tx.fc_hash, it).seq != sender->payload.seq)) {
-            /* Just ignore, ep was removed. We either not expecting grant on
-             * this EP or this is not the same grant sequence number we are
-             * expecting. */
+        sender = (uct_dc_fc_sender_data_t*)(hdr + 1);
+        ep     = (uct_dc_mlx5_ep_t*)sender->ep;
+        if (!uct_dc_mlx5_fc_remove_ep(iface, ep, sender->payload.seq)) {
+            /* Don't process FC_PURE_GRANT further, if an endpoint doesn't
+             * exist anymore, unexpected grant was received on an endpoint or
+             * this is not the same grant sequence number which was expected */
             return UCS_OK;
         }
 
-        ep      = (uct_dc_mlx5_ep_t *)sender->ep;
         cur_wnd = ep->fc.fc_wnd;
+        uct_rc_fc_restore_wnd(&iface->super.super, &ep->fc);
+        UCS_STATS_UPDATE_COUNTER(ep->fc.stats, UCT_RC_FC_STAT_RX_PURE_GRANT,
+                                 1);
 
-        /* Peer granted resources, so update wnd */
-        uct_rc_fc_restore_wnd(rc_iface, &ep->fc);
-
-        /* Remove FC entry to complete */
-        uct_dc_mlx5_fc_entry_iter_del(iface, it);
-
-        UCS_STATS_UPDATE_COUNTER(ep->fc.stats, UCT_RC_FC_STAT_RX_PURE_GRANT, 1);
-
-        /* To preserve ordering we have to dispatch all pending
-         * operations if current fc_wnd is <= 0 */
-        if (cur_wnd <= 0) {
+        /* To preserve ordering we have to dispatch all pending operations if
+         * current fc_wnd is <= 0 */
+        if (cur_wnd == 0) {
             uct_dc_mlx5_get_arbiter_params(iface, ep, &waitq, &group,
                                            &pool_index);
             ucs_arbiter_group_schedule(waitq, group);
@@ -1663,27 +1670,34 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
      * purged - they need to be rescheduled when DCI will be recovered after an
      * error */
 
-    if (ep->flags & (UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED |
-                     UCT_DC_MLX5_EP_FLAG_FLUSH_CANCEL)) {
+    if (ep == iface->tx.fc_ep) {
+        /* Do not report errors on flow control endpoint */
+        if (!(iface->flags & UCT_DC_MLX5_IFACE_FLAG_FC_EP_FAILED)) {
+            ucs_debug("got error on DC flow-control endpoint, iface %p: %s",
+                      iface, ucs_status_string(ep_status));
+        }
+
+        iface->flags |= UCT_DC_MLX5_IFACE_FLAG_FC_EP_FAILED;
         return;
     }
 
-    if (ep == iface->tx.fc_ep) {
-        iface->flags |= UCT_DC_MLX5_IFACE_FLAG_FC_EP_FAILED;
-
-        /* Do not report errors on flow control endpoint */
-        ucs_debug("got error on DC flow-control endpoint, iface %p: %s", iface,
-                  ucs_status_string(ep_status));
+    if (ep->flags & UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED) {
         return;
     }
 
     ep->flags |= UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED;
+
+    uct_dc_mlx5_fc_remove_ep(iface, ep, UINT64_MAX);
     uct_rc_fc_restore_wnd(&iface->super.super, &ep->fc);
 
-    status  = uct_iface_handle_ep_err(&ib_iface->super.super,
-                                      &ep->super.super, ep_status);
-    log_lvl = uct_base_iface_failure_log_level(&ib_iface->super, status,
-                                               ep_status);
+    if (ep->flags & UCT_DC_MLX5_EP_FLAG_FLUSH_CANCEL) {
+        return;
+    }
+
+    status     = uct_iface_handle_ep_err(&ib_iface->super.super,
+                                         &ep->super.super, ep_status);
+    log_lvl    = uct_base_iface_failure_log_level(&ib_iface->super, status,
+                                                  ep_status);
     uct_ib_mlx5_completion_with_err(ib_iface, (uct_ib_mlx5_err_cqe_t*)cqe, txwq,
                                     log_lvl);
 }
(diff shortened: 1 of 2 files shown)


dmitrygx:

Artemy-Mellanox (approved):

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1663,6 +1674,7 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
      * purged - they need to be rescheduled when DCI will be recovered after an
      * error */
 
+    uct_dc_mlx5_fc_restore_wnd(iface, ep, NULL);
```
yosefe, line 1677: do we really want to do it when UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED is true?
or when the ep is equal to iface->tx.fc_ep ?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1093,6 +1093,50 @@ void uct_dc_mlx5_fc_entry_iter_del(uct_dc_mlx5_iface_t *iface, khiter_t it)
...
+    int16_t cur_wnd;
+    khiter_t it;
+
+    it = kh_get(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash, (uint64_t)ep);
+    if ((fc_data != NULL) &&
+        ((it == kh_end(&iface->tx.fc_hash)) ||
+         (kh_value(&iface->tx.fc_hash, it).seq != fc_data->payload.seq))) {
+        /* Don't process FC_PURE_GRANT further, if an endpoint doesn't exist
+         * anymore, unexpected grant was received on an endpoint or this is not
+         * the same grant sequence number which was expected */
```
yosefe, line 1112: remove some spaces before `*/`

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1093,6 +1093,50 @@ void uct_dc_mlx5_fc_entry_iter_del(uct_dc_mlx5_iface_t *iface, khiter_t it)
...
+        return;
+    }
```
yosefe, line 1114: seems this code should be only in FC_PURE_GRANT handler, and not in the common function

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1093,6 +1093,50 @@ void uct_dc_mlx5_fc_entry_iter_del(uct_dc_mlx5_iface_t *iface, khiter_t it)
...
+
+    /* To preserve ordering we have to dispatch all pending operations if
+     * current fc_wnd is <= 0 */
+    if (cur_wnd <= 0) {
+        uct_dc_mlx5_get_arbiter_params(iface, ep, &waitq, &group,
+                                       &pool_index);
+        ucs_arbiter_group_schedule(waitq, group);
+        uct_dc_mlx5_iface_progress_pending(iface, pool_index);
+        uct_dc_mlx5_iface_check_tx(iface);
+    }
```
yosefe, line 1137: seems this code should also be only in FC_PURE_GRANT handler

src/uct/ib/dc/dc_mlx5.c
dmitrygx: it simplifies the code, because we should do it in case of `flush(CANCEL)` was done, but `UCT_DC_MLX5_EP_FLAG_FLUSH_CANCEL` isn't set here.
so, we can call it many times, but it wouldn't perform any actions.
but we can check for `fc_ep` first to not restore FC window in case of FC  EP - done

src/uct/ib/dc/dc_mlx5.c
dmitrygx: done

src/uct/ib/dc/dc_mlx5.c
dmitrygx: but there is only one space

src/uct/ib/dc/dc_mlx5.c
dmitrygx: refactored

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1677,13 +1681,19 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
...
 
-    ep->flags |= UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED;
+    uct_dc_mlx5_fc_retrieve_ep(iface, ep, NULL);
     uct_rc_fc_restore_wnd(&iface->super.super, &ep->fc);
 
-    status  = uct_iface_handle_ep_err(&ib_iface->super.super,
-                                      &ep->super.super, ep_status);
-    log_lvl = uct_base_iface_failure_log_level(&ib_iface->super, status,
-                                               ep_status);
+    if (ep->flags & (UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED |
```
yosefe, line 1687: if UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED, can we skip uct_dc_mlx5_fc_retrieve_ep and uct_rc_fc_restore_wnd ?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1093,6 +1093,22 @@ void uct_dc_mlx5_fc_entry_iter_del(uct_dc_mlx5_iface_t *iface, khiter_t it)
     }
 }
 
+static UCS_F_ALWAYS_INLINE uct_dc_mlx5_ep_t*
+uct_dc_mlx5_fc_retrieve_ep(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep,
+                           uct_dc_fc_sender_data_t *fc_data)
```
yosefe, line 1098: IMO this function should be like
```
int uct_dc_mlx5_fc_remove_ep(iface, ep, seq) 
{
   it = ...
   if (it != end && (seq == UINT64_MAX || seq==kh_value(it).seq) {
      iter_del(it)
      return 1;
   }
   return 0;
}
```

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1093,6 +1093,22 @@ void uct_dc_mlx5_fc_entry_iter_del(uct_dc_mlx5_iface_t *iface, khiter_t it)
...
+{
+    khiter_t it = kh_get(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash,
+                         (uint64_t)ep);
+    if ((it != kh_end(&iface->tx.fc_hash)) &&
+        ((fc_data == NULL) ||
```
brminich: when can it be NULL? On line 1157 you unconditionally access this struct to get ep

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1138,31 +1153,24 @@ uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
...
-
-        it = kh_get(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash, sender->ep);
-        if ((it == kh_end(&iface->tx.fc_hash)) ||
-            (kh_value(&iface->tx.fc_hash, it).seq != sender->payload.seq)) {
-            /* Just ignore, ep was removed. We either not expecting grant on
-             * this EP or this is not the same grant sequence number we are
-             * expecting. */
-            return UCS_OK;
+        sender = (uct_dc_fc_sender_data_t*)(hdr + 1);
+        ep     = (uct_dc_mlx5_ep_t*)sender->ep;
```
brminich, line 1158: `uct_dc_mlx5_fc_retrieve_ep` converts it back to `uint64`. Better do not pass ep to `uct_dc_mlx5_fc_retrieve_ep` and make all handling inside

src/uct/ib/dc/dc_mlx5.c
dmitrygx: refactored

src/uct/ib/dc/dc_mlx5.c
dmitrygx: refactored, now it is not needed

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1138,31 +1154,24 @@ uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
...
+        ret    = uct_dc_mlx5_fc_remove_ep(iface, ep, sender->payload.seq);
```
yosefe, line 1159: ret variable seems redundant

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1138,31 +1154,24 @@ uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
...
+        if (ret == 0) {
+            /* Don't process FC_PURE_GRANT further, if an endpoint doesn't
+             * exist anymore, unexpected grant was received on an endpoint or
+             * this is not the same grant sequence number which was expected */
+            goto out;
```
yosefe, line 1164: maybe just "return UCS_OK"?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1677,13 +1682,22 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
         return;
     }
 
-    ep->flags |= UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED;
+    if (ep->flags & UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED) {
+        return;
+    }
```
yosefe, line 1687: can we move this back before checking fc ep?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1677,13 +1682,22 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
...
 
-    status  = uct_iface_handle_ep_err(&ib_iface->super.super,
-                                      &ep->super.super, ep_status);
-    log_lvl = uct_base_iface_failure_log_level(&ib_iface->super, status,
-                                               ep_status);
+    if (ep->flags & UCT_DC_MLX5_EP_FLAG_FLUSH_CANCEL) {
+        return;
+    }
+
+    ep->flags |= UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED;
```
yosefe, line 1696: need to do it before checking CANCEL

src/uct/ib/dc/dc_mlx5.c
dmitrygx: removed

src/uct/ib/dc/dc_mlx5.c
dmitrygx: done

src/uct/ib/dc/dc_mlx5.c
dmitrygx: done

src/uct/ib/dc/dc_mlx5.c
dmitrygx: since we don't need to do flush(CANCEL) or invoke error handler for FC EP, I guess it worth checking FC EP first

dmitrygx:

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1669,8 +1669,9 @@ ucs_status_t uct_dc_mlx5_ep_check_fc(uct_dc_mlx5_iface_t *iface,
         fc_entry = &kh_value(&iface->tx.fc_hash, it);
     }
 
-    fc_entry->seq       = iface->tx.fc_seq++;
     fc_entry->send_time = now;
+    fc_entry->seq       = iface->tx.fc_seq++;
+    ucs_assert(iface->tx.fc_seq != UINT64_MAX);
```
brminich, line 1674: what corner case does it check?

dmitrygx: we pass `UINT64_MAX` to `uct_dc_mlx5_fc_remove_ep` function which indicates that no need to check `seq` when getting element from FC hash

brminich: so you checking that counter on iface does not reach `UINT64_MAX`?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1156,12 +1155,11 @@ uct_dc_mlx5_iface_fc_handler(uct_rc_iface_t *rc_iface, unsigned qp_num,
     } else if (fc_hdr == UCT_RC_EP_FC_PURE_GRANT) {
         sender = (uct_dc_fc_sender_data_t*)(hdr + 1);
         ep     = (uct_dc_mlx5_ep_t*)sender->ep;
-        ret    = uct_dc_mlx5_fc_remove_ep(iface, ep, sender->payload.seq);
-        if (ret == 0) {
+        if (uct_dc_mlx5_fc_remove_ep(iface, ep, sender->payload.seq) == 0) {
```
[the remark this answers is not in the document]
dmitrygx: done

src/uct/ib/dc/dc_mlx5.c
dmitrygx: FC endpoint can't enter the body of this `if` statement.
need to check `UCT_DC_MLX5_IFACE_FLAG_FC_EP_FAILED` before printing a debug message - done

src/uct/ib/dc/dc_mlx5_ep.c
dmitrygx: > so you checking that counter on iface does not reach UINT64_MAX?

yes, that's right to make sure that check `uct_dc_mlx5_fc_remove_ep` works correctly.
added the comment.

yosefe (approved):

brminich (approved):

PR #8493, opened 2022-09-02
UCP/EXAMPLE: Get rid of deprecated API usage inside ucp_hello_world

ivankochin:
## What
Switches deprecated API calls inside `ucp_hello_world` example to non-deprecated ones.

### DIFF
--- a/examples/ucp_hello_world.c
+++ b/examples/ucp_hello_world.c
@@ -88,10 +88,7 @@ static const ucp_tag_t tag_mask = UINT64_MAX;
 static const char *addr_msg_str = "UCX address message";
 static const char *data_msg_str = "UCX data message";
 static int print_config         = 0;
-static ucp_address_t *local_addr;
 static ucp_address_t *peer_addr;
-
-static size_t local_addr_len;
 static size_t peer_addr_len;
 
 static ucs_status_t parse_cmd(int argc, char * const argv[], char **server_name);
@@ -131,7 +128,7 @@ static void failure_handler(void *arg, ucp_ep_h ep, ucs_status_t status)
 }
 
 static void recv_handler(void *request, ucs_status_t status,
-                        ucp_tag_recv_info_t *info)
+                         const ucp_tag_recv_info_t *info, void *user_data)
 {
     struct ucx_context *context = (struct ucx_context *)request;
 
@@ -226,12 +223,13 @@ static void ep_close_err_mode(ucp_worker_h ucp_worker, ucp_ep_h ucp_ep)
     ep_close(ucp_worker, ucp_ep, ep_close_flags);
 }
 
-static int run_ucx_client(ucp_worker_h ucp_worker)
+static int run_ucx_client(ucp_worker_h ucp_worker, ucp_address_t *local_addr,
+                          size_t local_addr_len)
 {
     struct msg *msg = NULL;
     size_t msg_len  = 0;
     int ret         = -1;
-    ucp_request_param_t send_param;
+    ucp_request_param_t send_param, recv_param;
     ucp_tag_recv_info_t info_tag;
     ucp_tag_message_h msg_tag;
     ucs_status_t status;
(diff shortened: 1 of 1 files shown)


examples/ucp_hello_world.c
```diff
@@ -575,8 +583,13 @@ int main(int argc, char **argv)
...
 
-    status = ucp_worker_get_address(ucp_worker, &local_addr, &local_addr_len);
-    CHKERR_JUMP(status != UCS_OK, "ucp_worker_get_address\n", err_worker);
+    worker_attr.field_mask = UCP_WORKER_ATTR_FIELD_ADDRESS;
+
+    status = ucp_worker_query(ucp_worker, &worker_attr);
+    CHKERR_JUMP(status != UCS_OK, "ucp_worker_query\n", err_worker);
+
+    local_addr     = worker_attr.address;
+    local_addr_len = worker_attr.address_length;
```
yosefe, line 592: maybe the variables local_addr, local_addr_len could be removed now?

ivankochin: Yes, we can do this, but both this variable are also used inside `run_ucx_client` function. So I see three option how to deal with it:
1. Leave it as it done now
2. Make `worker_attr` a global variable. I don't think that this is a good option.
3. Pass `worker_attr` to `run_test` and therefore to `run_ucx_client` to obtain address information there. Maybe even remove the `run_test` wrapper at all.

I would prefer to implement the third option. Do you agree with me?

yosefe: option3 + remove run_test + pass local_addr+local_addr_len as **parameters** to run_ucx_client (instead of having them as global vars or passing ucp_worker_attr_t* parameter)

examples/ucp_hello_world.c
```diff
@@ -319,9 +317,15 @@ static int run_ucx_client(ucp_worker_h ucp_worker)
...
     CHKERR_JUMP(msg == NULL, "allocate memory\n", err_ep);
 
-    request = ucp_tag_msg_recv_nb(ucp_worker, msg, info_tag.length,
-                                  ucp_dt_make_contig(1), msg_tag,
-                                  recv_handler);
+    recv_param.op_attr_mask = UCP_OP_ATTR_FIELD_CALLBACK |
+                              UCP_OP_ATTR_FIELD_DATATYPE |
+                              UCP_OP_ATTR_FLAG_NO_IMM_CMPL;
+    recv_param.datatype     = ucp_dt_make_contig(1);
+    recv_param.cb.recv      = (ucp_tag_recv_nbx_callback_t)recv_handler;
```
evgeny-leksikov, line 324: let's add another parameter to `recv_handler` to avoid type cast

examples/ucp_hello_world.c
```diff
@@ -603,17 +600,22 @@ int main(int argc, char **argv)
...
-        addr_len = local_addr_len;
+        addr_len = worker_attr.address_length;
+
         ret = send(oob_sock, &addr_len, sizeof(addr_len), 0);
         CHKERR_JUMP_RETVAL(ret != (int)sizeof(addr_len),
                            "send address length\n", err_peer_addr, ret);
 
-        ret = send(oob_sock, local_addr, local_addr_len, 0);
-        CHKERR_JUMP_RETVAL(ret != (int)local_addr_len, "send address\n",
+        ret = send(oob_sock, worker_attr.address, addr_len, 0);
```
evgeny-leksikov, line 609: minor: `addr_len` seems redundant here - IMO `worker_attr.address, worker_attr.address_length` looks clearer

ivankochin: So the all this "OOB connection establishment" part would look like the following:
```c
        ret = send(oob_sock, &worker_attr.address_length,
                   sizeof(worker_attr.address_length), 0);
        CHKERR_JUMP_RETVAL(ret != (int)sizeof(worker_attr.address_length),
                           "send address length\n", err_peer_addr, ret);

        ret = send(oob_sock, worker_attr.address, worker_attr.address_length,
                   0);
        CHKERR_JUMP_RETVAL(ret != (int)worker_attr.address_length,
                           "send address\n", err_peer_addr, ret);
```
Instead of current:
```c
        addr_len = worker_attr.address_length;

        ret = send(oob_sock, &addr_len, sizeof(addr_len), 0);
        CHKERR_JUMP_RETVAL(ret != (int)sizeof(addr_len),
                           "send address length\n", err_peer_addr, ret);

        ret = send(oob_sock, worker_attr.address, addr_len, 0);
        CHKERR_JUMP_RETVAL(ret != (int)addr_len, "send address\n",
                           err_peer_addr, ret);
```
As you can see the first option contains many line breaks in the middle of the function and at my point of view it looks more complex that the second one. So I would prefer to keep the things in current shape.

ivankochin:
@evgeny-leksikov pls take a look one more time.

examples/ucp_hello_world.c
evgeny-leksikov: Then you can add another variable `addr` to save couple lines. My comment was about making func parameters aligned.

evgeny-leksikov (dismissed):

evgeny-leksikov (approved):

ivankochin:
@evgeny-leksikov pls re-approve

evgeny-leksikov (approved):

PR #8494, opened 2022-09-02
CONFIG: Fix bashisms in configure.ac, config/m4/go.m4

thesamesam:
configure scripts need to be runnable with a POSIX-compliant /bin/sh.

On many (but not all!) systems, /bin/sh is provided by Bash, so errors
like this aren't spotted. Notably Debian defaults to /bin/sh provided
by dash which doesn't tolerate such bashisms as '=='.

This retains compatibility with bash.

Fixes configure warnings/errors like:
```
checking for go... yes
./configure: 26781: test: xyes: unexpected operator
```

Signed-off-by: Sam James <sam@gentoo.org>

## What
Remove Bashisms from the build system which cause issues
when `/bin/sh` is not Bash, but is a POSIX-compliant shell.

## Why ?
The build may be misconfigured or fail with a POSIX-compliant-but-not-Bash
/bin/sh.

### DIFF
--- a/config/m4/go.m4
+++ b/config/m4/go.m4
@@ -21,7 +21,7 @@ AS_IF([test "x$with_go" != xno],
                   [AS_VERSION_COMPARE([1.16], [`go version | awk '{print substr($3, 3, length($3)-2)}'`],
                                       [go_happy="yes"], [go_happy="yes"], [go_happy=no])],
                   [go_happy=no])
-            AS_IF([test "x$go_happy" == xno],
+            AS_IF([test "x$go_happy" = xno],
                   [AS_IF([test "x$with_go" = "xguess"],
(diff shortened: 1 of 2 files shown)


thesamesam:
I'm not sure how I can satisfy the commit message style checker for this change.

shamisp:
@thesamesam, thanks for the bugfix !  

Here the guidance how to structure your contribution and commit messages:
https://github.com/openucx/ucx/wiki/Guidance-for-contributors

We also ask contributors to sign CLA: https://openucx.org/license/

Hopefully this helps !

thesamesam:
Thanks @shamisp. I emailed my CLA earlier today. Hopefully this looks OK now.

yosefe (approved):

shamisp:
We have received the CLA

shamisp (approved):

PR #8499, opened 2022-09-05
UCT/EP: basic implementation for ep_connect_to_ep_v2

hoopoepg:
- added basic implementation for ep_connect_to_ep
  routine
- added stubs for existing transports


test/gtest/uct/uct_test.cc
```diff
@@ -1241,23 +1249,15 @@ void uct_test::entity::connect_to_ep(unsigned index, entity& other,
...
         return; /* Already connected */
     }
 
     other.reserve_ep(other_index);
-
-    iface_addr = (uct_iface_addr_t*)ucs_alloca(other.iface_attr().iface_addr_len);
-
-    ep_params.field_mask = UCT_EP_PARAM_FIELD_IFACE |
-                           UCT_EP_PARAM_FIELD_IFACE_ADDR;
+    ep_params.field_mask = UCT_EP_PARAM_FIELD_IFACE;
```
brminich, line 1259: can we do both - create ep with iface addr and fo connect_to_ep_v2

src/uct/base/uct_iface.c
```diff
@@ -460,7 +460,8 @@ uct_iface_internal_ops_t uct_base_iface_internal_ops = {
     .iface_estimate_perf = uct_base_iface_estimate_perf,
     .iface_vfs_refresh   = (uct_iface_vfs_refresh_func_t)ucs_empty_function,
     .ep_query            = (uct_ep_query_func_t)ucs_empty_function_return_unsupported,
-    .ep_invalidate       = (uct_ep_invalidate_func_t)ucs_empty_function_return_unsupported
+    .ep_invalidate       = (uct_ep_invalidate_func_t)ucs_empty_function_return_unsupported,
+    .ep_connect_to_ep_v2 = ucs_empty_function_return_ep_timeout
```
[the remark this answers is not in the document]
hoopoepg: it is supported only on subset of TLs where ep_connect_to_ep is supported

test/gtest/uct/uct_test.cc
brminich: maybe in this case we can check whether it is different in the connect and return an error?

hoopoepg: ok, done

src/uct/ib/rc/accel/rc_mlx5_ep.c
```diff
@@ -1017,19 +1027,6 @@ UCS_CLASS_INIT_FUNC(uct_rc_mlx5_ep_t, const uct_ep_params_t *params)
...
-                                                               IFACE_ADDR,
-                                                               NULL);
-    /* if iface address is not provided or address doesn't contain
-     * FLUSH_RKEY flag - mark flush_rkey as invalid */
-    if ((addr != NULL) &&
-        (addr->super.flags & UCT_RC_MLX5_IFACE_ADDR_FLAG_FLUSH_RKEY)) {
-        self->super.flush_rkey = (uint32_t)addr->flush_rkey_hi << 16;
-    } else {
-        self->super.flush_rkey = UCT_IB_MD_INVALID_FLUSH_RKEY;
-    }
```
brminich, line 1031: why removed? seems needed for connect to iface transports

hoopoepg: not sure I got question... this is rc_mlx5 transport - connect to EP supported only, rkey is updated in connect_to_ep_v2 routine

brminich: oh, you are right

brminich (dismissed):

hoopoepg:
@yosefe ok to squash?

src/ucp/core/ucp_proxy_ep.h
```diff
@@ -24,16 +25,16 @@
...
  */
 typedef struct ucp_proxy_ep {
-    uct_ep_t        super;     /**< Derived from uct_ep */
-    uct_iface_t     iface;     /**< Embedded stub interface */
-    ucp_ep_h        ucp_ep;    /**< Pointer to UCP endpoint */
-    uct_ep_h        uct_ep;    /**< Underlying transport endpoint */
-    int             is_owner;  /**< Is uct_ep owned by this proxy ep */
-    ucp_rsc_index_t rsc_index; /**< Resource index of underlying transport endpoint */
+    uct_ep_t         super;     /**< Derived from uct_ep */
+    uct_base_iface_t iface;     /**< Embedded stub interface */
```
yosefe, line 29: we should not use an internal UCT header in UCP
instead, need to check if a uct_ep_t is proxy or regular ep every time before calling uct_ep_connect_to_ep_v2() from UCP

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -395,7 +395,8 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
     .iface_estimate_perf = uct_cuda_copy_estimate_perf,
     .iface_vfs_refresh   = (uct_iface_vfs_refresh_func_t)ucs_empty_function,
     .ep_query            = (uct_ep_query_func_t)ucs_empty_function_return_unsupported,
-    .ep_invalidate       = (uct_ep_invalidate_func_t)ucs_empty_function_return_unsupported
+    .ep_invalidate       = (uct_ep_invalidate_func_t)ucs_empty_function_return_unsupported,
+    .ep_connect_to_ep_v2 = ucs_empty_function_return_ep_timeout
```
yosefe, line 399: should be ucs_empty_function_return_unsupported on all 2iface transports

src/uct/tcp/tcp_iface.c
```diff
@@ -550,6 +550,14 @@ static ucs_mpool_ops_t uct_tcp_mpool_ops = {
     .obj_str       = NULL
 };
 
+uct_iface_internal_ops_t uct_tcp_iface_internal_ops = {
```
yosefe, line 553: static

src/uct/ugni/smsg/ugni_smsg_iface.c
```diff
@@ -267,6 +267,14 @@ static ucs_mpool_ops_t uct_ugni_smsg_mbox_mpool_ops = {
     .obj_str       = NULL
 };
 
+uct_iface_internal_ops_t uct_ugni_smsg_iface_internal_ops = {
```
yosefe, line 270: can it be static?

src/ucp/core/ucp_proxy_ep.h
hoopoepg: hmmm, but test for proxy is UCP call... is it ok?

yosefe: yes; since UCP is always calling uct_ep_connect_to_ep_v2, it should manually check if it's proxy ep or not, every time.

hoopoepg:
closed via https://github.com/openucx/ucx/pull/8565

PR #7683, opened 2021-11-21
UCP/SELECT: Add mem access lanes for all supported memory types

yosefe:
## Why
Make sure optimal memory-type transports are selected for intra-node endpoints, and not just host/ipc transports

PR #8503, opened 2022-09-06
UCP: Fix rndv lanes selections for memtypes

brminich:
## What
Fix rma_bw lanes selection for different memory types. 

## Why ?
No need to search rma_bw lanes for different memory types on the same memory domains to avoid asymmetrical lanes selection  (e.g. when one node has nv_pee_mem loaded, while the other does not) 

## How ?
Partially revert #7683

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1600,7 +1600,7 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
     ucs_memory_type_t mem_type;
     size_t added_lanes;
     uint64_t md_reg_flag;
-    ucp_tl_bitmap_t tl_bitmap;
+    ucp_tl_bitmap_t tl_bitmap, mem_type_tl_bitmap;
     uint8_t i;
     ucp_wireup_select_flags_t iface_rma_flags, peer_rma_flags;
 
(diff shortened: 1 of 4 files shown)


hoopoepg (approved):

src/ucp/wireup/select.c
```diff
@@ -1691,11 +1691,23 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
         /* Add lanes that can access the memory by short operations */
         added_lanes = 0;
+        UCS_BITMAP_CLEAR(&tl_bitmap);
+
         for (mem_type = 0; mem_type < UCS_MEMORY_TYPE_LAST; mem_type++) {
+            UCP_CONTEXT_MEM_CAP_TLS(context, mem_type, reg_mem_types,
+                                    mem_type_tl_bitmap);
+            if (UCS_BITMAP_IS_ZERO_INPLACE(&mem_type_tl_bitmap)) {
+                continue;
+            }
```
yosefe, line 1701: why is it needed, wouldn't ucp_wireup_add_bw_lanes return 0 lanes in this case anyway?

brminich: No, for instance if for HOST memory we selected one IB and one tcp lane, then on the next round it will add more IB lanes for CUDA. This leads to asymmetrical lanes selection if GDR is enabled on one of the peers only

yosefe: right, so I'd think that just adding the UCP_TL_BITMAP_AND_NOT and UCS_BITMAP_OR_INPLACE below would be needed, and the check 'if UCS_BITMAP_IS_ZERO_INPLACE' can be removed.

yosefe (approved):

yosefe (approved):

PR #8506, opened 2022-09-07
Am user header release/support protov2 reset

ofirfarjun7:
## What
Support re-initialization of protocols with non persistent user header

## Why ?
#8386 
Zcopy clean function need to handle the case when user header is not persistent and need to be copied.

## How ?
[Description can be found here](https://github.com/openucx/ucx/pull/8173#discussion_r928840502)

### DIFF
--- a/src/ucp/am/eager.inl
+++ b/src/ucp/am/eager.inl
@@ -70,8 +70,9 @@ ucp_am_eager_zcopy_pack_user_header(ucp_request_t *req)
     }
 
     if (req->send.msg_proto.am.header.length != 0) {
-        ucs_assert(req->send.msg_proto.am.header.user_ptr != NULL);
+        ucs_assert(req->send.msg_proto.am.header.ptr != NULL);
         ucp_am_pack_user_header(reg_desc + 1, req);
+        ucp_am_release_user_header(req);
     }
 
     req->send.msg_proto.am.header.reg_desc = reg_desc;
--- a/src/ucp/am/eager_single.c
+++ b/src/ucp/am/eager_single.c
@@ -51,7 +51,7 @@ ucp_am_eager_short_proto_progress_common(uct_pending_req_t *self, int is_reply)
                          &iov_cnt);
 
     if (header_length != 0) {
-        ucp_add_uct_iov_elem(iov,req->send.msg_proto.am.header.user_ptr,
+        ucp_add_uct_iov_elem(iov, req->send.msg_proto.am.header.ptr,
                              header_length, UCT_MEM_HANDLE_NULL, &iov_cnt);
     }
 
@@ -165,7 +165,7 @@ ucp_proto_t ucp_am_eager_short_reply_proto = {
     .init     = ucp_am_eager_short_reply_proto_init,
     .query    = ucp_proto_single_query,
     .progress = {ucp_am_eager_short_reply_proto_progress},
-    .abort    = ucp_proto_request_bcopy_abort,
+    .abort    = ucp_proto_am_request_bcopy_abort,
     .reset    = ucp_proto_request_bcopy_reset
 };
 
@@ -402,7 +402,7 @@ ucp_proto_t ucp_am_eager_single_zcopy_proto = {
     .init     = ucp_am_eager_single_zcopy_proto_init,
     .query    = ucp_proto_single_query,
     .progress = {ucp_am_eager_single_zcopy_proto_progress},
-    .abort    = ucp_proto_request_zcopy_abort,
+    .abort    = ucp_proto_am_request_zcopy_abort,
     .reset    = ucp_am_proto_request_zcopy_reset
 };
 
@@ -458,6 +458,6 @@ ucp_proto_t ucp_am_eager_single_zcopy_reply_proto = {
     .init     = ucp_am_eager_single_zcopy_reply_proto_init,
     .query    = ucp_proto_single_query,
     .progress = {ucp_am_eager_single_zcopy_reply_proto_progress},
-    .abort    = ucp_proto_request_zcopy_abort,
+    .abort    = ucp_proto_am_request_zcopy_abort,
     .reset    = ucp_am_proto_request_zcopy_reset
 };
--- a/src/ucp/core/ucp_am.c
+++ b/src/ucp/core/ucp_am.c
@@ -370,7 +370,7 @@ ucp_am_zcopy_pack_user_header(ucp_request_t *req)
     }
 
     if (req->send.msg_proto.am.header.length != 0) {
-        ucs_assert(req->send.msg_proto.am.header.user_ptr != NULL);
+        ucs_assert(req->send.msg_proto.am.header.ptr != NULL);
         ucp_am_pack_user_header(reg_desc + 1, req);
     }
 
@@ -576,7 +576,7 @@ static ucs_status_t ucp_am_contig_short(uct_pending_req_t *self)
     req->send.lane = ucp_ep_get_am_lane(ep);
     status         = ucp_am_send_short(ep, req->send.msg_proto.am.am_id,
                                        req->send.msg_proto.am.flags,
-                                       req->send.msg_proto.am.header.user_ptr,
+                                       req->send.msg_proto.am.header.ptr,
                                        req->send.msg_proto.am.header.length,
                                        req->send.buffer, req->send.length, 0);
     status         = ucp_am_handle_user_header_send_status(req, status);
@@ -592,7 +592,7 @@ static ucs_status_t ucp_am_contig_short_reply(uct_pending_req_t *self)
     req->send.lane = ucp_ep_get_am_lane(ep);
     status         = ucp_am_send_short(ep, req->send.msg_proto.am.am_id,
                                        req->send.msg_proto.am.flags,
-                                       req->send.msg_proto.am.header.user_ptr,
+                                       req->send.msg_proto.am.header.ptr,
                                        req->send.msg_proto.am.header.length,
                                        req->send.buffer, req->send.length, 1);
     status         = ucp_am_handle_user_header_send_status(req, status);
@@ -806,7 +806,7 @@ static void ucp_am_send_req_init(ucp_request_t *req, ucp_ep_h ep,
     req->send.ep                           = ep;
     req->send.msg_proto.am.am_id           = am_id;
     req->send.msg_proto.am.flags           = flags;
-    req->send.msg_proto.am.header.user_ptr = (void*)header;
+    req->send.msg_proto.am.header.ptr      = (void*)header;
     req->send.msg_proto.am.header.reg_desc = NULL;
     req->send.msg_proto.am.header.length   = header_length;
     req->send.buffer                       = (void*)buffer;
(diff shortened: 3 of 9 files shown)


ofirfarjun7:
TODO - Add tests: send a message with copy header flag, change header content after the send and before progressing.

src/ucp/core/ucp_am.c
```diff
@@ -1784,6 +1784,19 @@ const ucp_request_send_proto_t ucp_am_reply_proto = {
...
     ucs_assert(request->send.msg_proto.am.header.reg_desc != NULL);
+
+    if ((request->send.msg_proto.am.flags & UCP_AM_SEND_FLAG_COPY_HEADER) &&
+        !(request->flags & UCP_REQUEST_FLAG_USER_HEADER_COPIED) &&
+        (request->send.msg_proto.am.header.length != 0)) {
+        request->send.msg_proto.am.header.ptr = ucs_mpool_set_get_inline(
+                &request->send.ep->worker->am_mps,
+                request->send.msg_proto.am.header.length);
+        request->flags |= UCP_REQUEST_FLAG_USER_HEADER_COPIED;
+    }
```
evgeny-leksikov, line 1795: Reset of a request may be called when it's already on pending queue, so header must be copied before the request is enqueued since user header might be freed at this time.
IMO need to add an assertion here but copy header at very first trying to send the request.

ofirfarjun7: Correct me if I'm wrong, in Zcopy don't we copy the user header to UCX's registered buffer at the very first try of send?

evgeny-leksikov: Right, I missed that it's copied from `reg_desc`

evgeny-leksikov (approved):

src/ucp/am/eager.inl
```diff
@@ -50,6 +50,10 @@ static void ucp_am_eager_zcopy_completion(uct_completion_t *self)
                                           send.state.uct_comp);
 
     ucs_assert(req->send.msg_proto.am.header.reg_desc != NULL);
+    if (ucs_unlikely(req->flags & UCP_REQUEST_FLAG_USER_HEADER_COPIED)) {
+        ucp_am_release_user_header(req);
```
yosefe, line 54: can we do it in ucp_am_eager_zcopy_pack_user_header?
so we release the mpool element as soon as we copy it to reg_desc
IMO it will be more self-documenting

src/ucp/core/ucp_am.c
```diff
@@ -1784,6 +1784,19 @@ const ucp_request_send_proto_t ucp_am_reply_proto = {
 void ucp_am_proto_request_zcopy_reset(ucp_request_t *request)
 {
     ucs_assert(request->send.msg_proto.am.header.reg_desc != NULL);
+
+    if ((request->send.msg_proto.am.flags & UCP_AM_SEND_FLAG_COPY_HEADER) &&
```
yosefe, line 1788: pls add doc

src/ucp/core/ucp_request.h
```diff
@@ -186,7 +186,7 @@ struct ucp_request {
 
                         struct {
                             struct {
-                                void           *user_ptr;
+                                void           *ptr;
```
yosefe, line 189: pls document what this ptr could be in different cases (user ptr or am mpool element)

src/ucp/am/eager.inl
ofirfarjun7: It might require to set `UCP_REQUEST_FLAG_USER_HEADER_COPIED` flag back to zero because protocol might change more than once if I'm not mistaken. 
For example:
[Some initial protocol] - Allocate ucx buffer in ptr. copy hdr to ucx buffer in ptr.
Zcopy - pack and release buffer in ptr. zcopy reset will not allocate new buffer in ptr because `UCP_REQUEST_FLAG_USER_HEADER_COPIED` flag is on
[New protocol selection]

If I'm not mistaken I don't like the idea of setting the flag back to zero.

yosefe: yes, i meant setting UCP_REQUEST_FLAG_USER_HEADER_COPIED to 0 after we call ucp_am_release_user_header (which is after we copied the header to reg_desc).

ofirfarjun7: We can do that, although I don't like the idea of setting the flag back to zero...
We can also take a different approach, we basically need to keep an invariant, header resides in only one buffer, reg_desc when proto is zcopy (after pack) or .ptr when it's not (after reset).

- pack func will copy the header to reg_desc and free .ptr if needed ( `UCP_REQUEST_FLAG_USER_HEADER_COPIED` is set).
- reset func will ignore `UCP_REQUEST_FLAG_USER_HEADER_COPIED` flag because of the invariant, it will allocate buffer if `UCP_AM_SEND_FLAG_COPY_HEADER` is set and copy the header from reg_desc to .ptr
- abort func will be the old `ucp_proto_request_zcopy_abort` (no need to release the header because of the invariant)

WDYT?

src/ucp/am/eager.inl
```diff
@@ -70,8 +70,11 @@ ucp_am_eager_zcopy_pack_user_header(ucp_request_t *req)
     }
 
     if (req->send.msg_proto.am.header.length != 0) {
-        ucs_assert(req->send.msg_proto.am.header.user_ptr != NULL);
+        ucs_assert(req->send.msg_proto.am.header.ptr != NULL);
         ucp_am_pack_user_header(reg_desc + 1, req);
+        if ((req->flags & UCP_REQUEST_FLAG_USER_HEADER_COPIED)) {
```
ofirfarjun7, line 75: Will remove redundant parentheses

src/ucp/am/eager.inl
gleon99: As far as I understood @yosefe suggestion, it allows for a clearer and self-explanatory code, also indeed makes more sense to release immediately after copy in such case IMO.

src/ucp/am/eager_single.c
```diff
@@ -458,6 +458,6 @@ ucp_proto_t ucp_am_eager_single_zcopy_reply_proto = {
     .init     = ucp_am_eager_single_zcopy_reply_proto_init,
     .query    = ucp_proto_single_query,
     .progress = {ucp_am_eager_single_zcopy_reply_proto_progress},
-    .abort    = ucp_proto_am_request_zcopy_abort,
+    .abort    = ucp_proto_request_zcopy_abort,
```
yosefe, line 461: can we really do this (i.e remove ucp_proto_am_request_zcopy_abort)?
what if ucp_am_eager_zcopy_pack_user_header was not called yet when we abort the operation?

src/ucp/core/ucp_request.h
```diff
@@ -186,7 +186,9 @@ struct ucp_request {
 
                         struct {
                             struct {
-                                void           *ptr;
+                                void           *ptr; /* ptr can hold address of external buffer
+                                                        provided by the user or ucx mpool buffer
+                                                        when user's buffer is not guaranteed to be valid. */
```
yosefe, line 191: can you document which flags indicate what?

src/ucp/core/ucp_am.c
```diff
@@ -1784,6 +1784,22 @@ const ucp_request_send_proto_t ucp_am_reply_proto = {
...
+        (request->send.msg_proto.am.header.length != 0)) {
+        request->send.msg_proto.am.header.ptr = ucs_mpool_set_get_inline(
+                &request->send.ep->worker->am_mps,
+                request->send.msg_proto.am.header.length);
+        request->flags |= UCP_REQUEST_FLAG_USER_HEADER_COPIED;
+    }
+
+    memcpy(request->send.msg_proto.am.header.ptr,
+           request->send.msg_proto.am.header.reg_desc + 1,
+           request->send.msg_proto.am.header.length);
```
yosefe, line 1802: in case the if in line 1791 was false, what would be the value of request->send.msg_proto.am.header.ptr here?

src/ucp/core/ucp_am.c
```diff
@@ -1784,6 +1784,22 @@ const ucp_request_send_proto_t ucp_am_reply_proto = {
 void ucp_am_proto_request_zcopy_reset(ucp_request_t *request)
 {
     ucs_assert(request->send.msg_proto.am.header.reg_desc != NULL);
+
+    /* If user header is not guaranteed to be valid, 
+     * use mpool buffer for storing the user header.
+     */
+    if ((request->send.msg_proto.am.flags & UCP_AM_SEND_FLAG_COPY_HEADER) &&
+        !(request->flags & UCP_REQUEST_FLAG_USER_HEADER_COPIED) &&
```
yosefe, line 1792: we are in zcopy protocol here.
seems the only way UCP_REQUEST_FLAG_USER_HEADER_COPIED could be set is if we "switched" from zcopy to zcopy?
maybe need to release request->send.msg_proto.am.header.ptr in this case?

src/ucp/am/eager_single.c
ofirfarjun7: if it's indeed possible we can't

src/ucp/core/ucp_am.c
```diff
@@ -1784,22 +1784,27 @@ const ucp_request_send_proto_t ucp_am_reply_proto = {
 void ucp_am_proto_request_zcopy_reset(ucp_request_t *request)
 {
     ucs_assert(request->send.msg_proto.am.header.reg_desc != NULL);
+    /* Zcopy pack function releases header if it's
+     * stored in mpool buffer and set copied flag to zero.
+     * Zcopy pack function is always called
+     * before reset function.
```
yosefe, line 1790: wrap the comment text (longer lines)

src/ucp/core/ucp_request.h
```diff
@@ -186,12 +186,21 @@ struct ucp_request {
...
                             struct {
-                                void           *ptr; /* ptr can hold address of external buffer
-                                                        provided by the user or ucx mpool buffer
-                                                        when user's buffer is not guaranteed to be valid. */
-                                ucp_mem_desc_t *reg_desc; /* pointer to pre-registered buffer,
-                                                             used for sending header with
-                                                             zcopy protocol */
+                                /* Pointer to buffer used for sending header.
+                                 * - When UCP_AM_SEND_FLAG_COPY_HEADER is set
+                                 *   and reg_desc is not valid ptr holds
```
yosefe, line 191: what is the way to know if reg_desc is valid?

src/ucp/core/ucp_request.h
```diff
@@ -186,12 +186,21 @@ struct ucp_request {
...
+                                 *   address of ucx mpool buffer.
+                                 * - When UCP_AM_SEND_FLAG_COPY_HEADER is not set
+                                 *   ptr holds address of external buffer, provided
+                                 *   by the user, that is expected to be valid
+                                 *   during the send operation.
+                                 */
+                                void           *ptr;
+                                /* pointer to pre-registered buffer,
+                                 * used for sending header with
+                                 * zcopy protocol
```
yosefe, line 201: wrap lines

src/ucp/am/eager.inl
```diff
@@ -70,8 +70,11 @@ ucp_am_eager_zcopy_pack_user_header(ucp_request_t *req)
     }
 
     if (req->send.msg_proto.am.header.length != 0) {
-        ucs_assert(req->send.msg_proto.am.header.user_ptr != NULL);
+        ucs_assert(req->send.msg_proto.am.header.ptr != NULL);
         ucp_am_pack_user_header(reg_desc + 1, req);
+        if (req->flags & UCP_REQUEST_FLAG_USER_HEADER_COPIED) {
```
yosefe, line 75: seems ucp_am_release_user_header already checks COPIED flag

yosefe (dismissed):

yosefe:
@ofirfarjun7 pls squash

yosefe (approved):

PR #8507, opened 2022-09-07
UCT/IB/DC: Don't allocate new sequence number for resending existing FC_HARD_REQ

dmitrygx:
## What

Don't allocate a new sequence number for resending an existing `FC_HARD_REQ` packet.

## Why ?

Not reusing a sequence number which was allocated for FC entry existed in FC hash could lead to always dropping `FC_PURE_GRANT` packets with "old" sequence numbers if a peer is not fast enough to send `FC_PURE_GRANT` back.
This PR fixes IO-demo's "timeout waiting for replies", because server can't send replies, since FC window is exhausted and not renewed by receiving `FC_PURE_GRANT` with an expected sequence number.

## How ?

1. If `kh_put` completed with `UCS_KH_PUT_KEY_PRESENT`, don't assign new FC sequence number. just reuse the existing one from a FC entry returned from `kh_put`.
2. Add useful assertions to make sure FC hash is used correctly, i.e. a FC entry should exist in FC hash only if endpoint's `fc_wnd` < `fc_hard_thresh` and a FC entry shouldn't exist if endpoint's `fc_wnd` == `fc_hard_thresh`.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5_ep.c
+++ b/src/uct/ib/dc/dc_mlx5_ep.c
@@ -1657,8 +1657,15 @@ ucs_status_t uct_dc_mlx5_ep_check_fc(uct_dc_mlx5_iface_t *iface,
         ucs_error("failed to create hash entry for fc hard req");
         status = UCS_ERR_NO_MEMORY;
         goto out;
-    } else if (ret == UCS_KH_PUT_KEY_PRESENT) {
-        fc_entry = &kh_value(&iface->tx.fc_hash, it);
+    }
+
+    fc_entry = &kh_value(&iface->tx.fc_hash, it);
+    if (ret == UCS_KH_PUT_KEY_PRESENT) {
+        /* FC entry for a given endpoint should exist in FC hash only if
+         * endpoint's current FC window is lower than FC hard threshold */
+        ucs_assertv(ep->fc.fc_wnd < iface->super.super.config.fc_hard_thresh,
+                    "ep %p: fc_wnd=%d fc_hard_thresh=%d", ep, ep->fc.fc_wnd,
+                    iface->super.super.config.fc_hard_thresh);
 
         /* Do not resend FC request if timeout is not reached */
         if (ucs_likely((now - fc_entry->send_time) <
@@ -1666,17 +1673,15 @@ ucs_status_t uct_dc_mlx5_ep_check_fc(uct_dc_mlx5_iface_t *iface,
             goto out_set_status;
         }
     } else {
-        fc_entry = &kh_value(&iface->tx.fc_hash, it);
+        /* Expect that a FC sequence number doesn't reach UINT64_MAX value. So,
+         * uct_dc_mlx5_fc_remove_ep() can rely on "seq != UINT64_MAX" condition
+         * to check that a FC entry's sequnce number needs to be verified */
+        ucs_assert(iface->tx.fc_seq != UINT64_MAX);
+        fc_entry->seq = iface->tx.fc_seq++;
     }
 
-    /* Expect that a FC sequence number doesn't reach UINT64_MAX value. So,
-     * uct_dc_mlx5_fc_remove_ep() can rely on "seq != UINT64_MAX" condition to
-     * check that a FC entry's sequnce number needs to be verified */
-    ucs_assert(iface->tx.fc_seq != UINT64_MAX);
-    fc_entry->seq       = iface->tx.fc_seq++;
     fc_entry->send_time = now;
-
-    status = uct_dc_mlx5_ep_fc_hard_req_send(ep, fc_entry->seq);
+    status              = uct_dc_mlx5_ep_fc_hard_req_send(ep, fc_entry->seq);
     if (status != UCS_OK) {
         if (ret != UCS_KH_PUT_KEY_PRESENT) {
             kh_del(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash, it);
--- a/test/gtest/uct/ib/test_dc.cc
+++ b/test/gtest/uct/ib/test_dc.cc
@@ -496,9 +496,9 @@ public:
         }
     }
 
-    virtual void test_pending_grant(int wnd, uint64_t *wait_fc_seq = NULL)
+    virtual void test_pending_grant(int16_t wnd)
     {
-        test_rc_flow_control::test_pending_grant(wnd, wait_fc_seq);
+        test_rc_flow_control::test_pending_grant(wnd);
         flush();
     }
 };
@@ -519,14 +519,6 @@ UCS_TEST_P(test_dc_flow_control, general_disabled)
 UCS_TEST_P(test_dc_flow_control, pending_grant)
 {
     test_pending_grant(5);
-}
-
-UCS_TEST_P(test_dc_flow_control, pending_grant_and_resend_hard_req,
-           "DC_FC_HARD_REQ_TIMEOUT=0.1s")
-{
-    uct_dc_mlx5_iface_t *iface = ucs_derived_of(m_e1->iface(),
-                                                uct_dc_mlx5_iface_t);
-    test_pending_grant(5, &iface->tx.fc_seq);
 }
 
 UCS_TEST_P(test_dc_flow_control, fc_disabled_flush)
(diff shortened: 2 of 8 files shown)


yosefe (approved):

brminich (approved):

dmitrygx:
@yosefe @brminich thanks for the review!
could you review one more time pls? I noticed that an assertion checking `fc_wnd == fc_hard_req` in case of FC entry doesn't present might not be correct if sending FC_HARD_REQ was unsuccessful (e.g. no DCI), because we delete a FC entry from FC hash in this case.

brminich (dismissed):

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1691,6 +1684,9 @@ ucs_status_t uct_dc_mlx5_ep_check_fc(uct_dc_mlx5_iface_t *iface,
     status              = uct_dc_mlx5_ep_fc_hard_req_send(ep, fc_entry->seq);
     if (status != UCS_OK) {
         if (ret != UCS_KH_PUT_KEY_PRESENT) {
+            /* Decrement a FC sequence number to roll back changes done above,
+             * because FC_HARD_REQ wasn't sent with this sequence number */
+            iface->tx.fc_seq--;
```
yosefe, line 1689: is it really important to make sure we never "skip" a sequence number?
or maybe just ++ it after line 1700?

dmitrygx: > or maybe just ++ it after line 1700?

done

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1677,20 +1677,19 @@ ucs_status_t uct_dc_mlx5_ep_check_fc(uct_dc_mlx5_iface_t *iface,
...
         if (ret != UCS_KH_PUT_KEY_PRESENT) {
-            /* Decrement a FC sequence number to roll back changes done above,
-             * because FC_HARD_REQ wasn't sent with this sequence number */
-            iface->tx.fc_seq--;
             kh_del(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash, it);
         }
 
         goto out;
+    } else if (ret == UCS_KH_PUT_KEY_PRESENT) {
+            iface->tx.fc_seq++;
```
yosefe, line 1692: - indent
- why else if have goto?
- maybe ++ without checking ret?

dmitrygx: > indent
> why else if have goto?

done

> maybe ++ without checking ret?

I guess that better to increment only if FC seq was really used

dmitrygx: ok, removed

test/gtest/uct/uct_test.h
```diff
@@ -349,6 +349,24 @@ class uct_test : public testing::TestWithParam<const resource*>,
         }
     }
 
+    template <typename T>
+    void wait_for_value_change(volatile T *var, bool progress,
```
yosefe, line 353: maybe assume prrogress==true?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -496,9 +496,28 @@ class test_dc_flow_control : public test_rc_flow_control {
...
+                                                    uct_dc_mlx5_ep_t);
+
+        khiter_t it = kh_get(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash,
+                             (uint64_t)ep);
+        ASSERT_NE(it, kh_end(&iface->tx.fc_hash));
+
+        uct_dc_mlx5_ep_fc_entry_t *fc_entry = &kh_value(&iface->tx.fc_hash,
+                                                        it);
+        ucs_time_t fc_send_time = fc_entry->send_time;
+        wait_for_value_change(&fc_entry->send_time, true);
```
yosefe, line 513: could fc_entry be removed?

dmitrygx: no, it shouldn't be empty, since we shouldn't receive FC_PURE_GRANT from a peer.
but in order to make sure that FC_PURE_GRANT won't be received, I've added progressing only sender entity

test/gtest/uct/ib/test_rc.cc
```diff
@@ -666,7 +666,13 @@ void test_rc_flow_control::test_general(int wnd, int soft_thresh,
...
 }
 
-void test_rc_flow_control::test_pending_grant(int wnd, uint64_t *wait_fc_seq)
+void test_rc_flow_control::wait_fc_hard_resend(entity *e)
+{
+    UCS_TEST_ABORT("waiting FC_HARD_REQ resend isn't implemented in RC");
+}
+
+void test_rc_flow_control::test_pending_grant(int16_t wnd,
+                                              bool wait_fc_hard_req_resend)
```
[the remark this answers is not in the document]
dmitrygx: > BTW is it still needed? given that we ++ tx.seq unconditionally

yes, it fixes the test, because we don't increment FC entry anymore if it already exists in FC hash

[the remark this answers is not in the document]
dmitrygx: > why need bool? virt function could be empty..

not sure that I understand, functions should be identical and should have the same set of parameters, shouldn't it?

test/gtest/uct/uct_test.h
dmitrygx: done

test/gtest/uct/ib/test_rc.cc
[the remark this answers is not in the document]
yosefe: 1. why can't we wait for iface->tx.fc_seq to change? why must wait on hash entry?
2. instead of aborting the test with RC, can we just make the RC virtual function empty?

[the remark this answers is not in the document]
dmitrygx: > why can't we wait for iface->tx.fc_seq to change? why must wait on hash entry?

because `iface->tx.fc_seq` won't be changed in case of sending `FC_HARD_REQ` second time

> instead of aborting the test with RC, can we just make the RC virtual function empty?

it doesn't make sense to run this test for RC, so I decided to abort the test if it will be tested by mistake.

dmitrygx:
@yosefe fixed as discussed offline

dmitrygx:
@yosefe are these latest changes ok?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -658,6 +650,47 @@ UCS_TEST_P(test_dc_flow_control, dci_leak)
...
+class test_dc_flow_control_with_hard_req_resend : public test_dc_flow_control
+{
+protected:
+    virtual void init() {
+        ucs_status_t status = uct_config_modify(m_iface_config,
+                                                "DC_FC_HARD_REQ_TIMEOUT",
+                                                "0.1s");
+        if (status != UCS_OK) {
+            UCS_TEST_ABORT("cannot set DC_FC_HARD_REQ_TIMEOUT");
+        }
```
yosefe, line 663: maybe use uct_test::modify_config ?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -658,6 +650,47 @@ UCS_TEST_P(test_dc_flow_control, dci_leak)
...
+                                                    uct_dc_mlx5_ep_t);
+
+        khiter_t it = kh_get(uct_dc_mlx5_fc_hash, &iface->tx.fc_hash,
+                             (uint64_t)ep);
+        ASSERT_NE(it, kh_end(&iface->tx.fc_hash));
+
+        uct_dc_mlx5_ep_fc_entry_t *fc_entry = &kh_value(&iface->tx.fc_hash,
+                                                        it);
+        ucs_time_t fc_send_time = fc_entry->send_time;
+        wait_for_value_change(&fc_entry->send_time, e);
```
yosefe, line 681: do we need to avoid progress on target side?

dmitrygx: no, we don't need to progress a target side, otherwise - it will send FC_PURE_GRANT which is not needed

test/gtest/uct/ib/test_dc.cc
dmitrygx: done

yosefe (approved):

dmitrygx:
@yosefe squashed, could you review pls?

yosefe (approved):

PR #8508, opened 2022-09-07
UCT/MD/QUERY: Check that the size of dev/shm is not too small

ofirfarjun7:
## What
Check that the size of /dev/shm is not "too small" according to UCX perspective and disqualify its md to prevent the TL from opening.

## Why ?
Machine running UCX can limit the size of /dev/shm. This can result in insufficient memory for the TL resources, for example receive buffers pool. 

## How ?
In uct_posix_md_query check that the size of /dev/shm is bigger than the min value expected by UCX.
A configurable parameter (UCX_POSIX_SHM_MIN_SIZE) was added to the UCX configuration. it's enforce shm minimum size. Default value is set to 16MB.

### DIFF
--- a/src/uct/sm/mm/base/mm_iface.c
+++ b/src/uct/sm/mm/base/mm_iface.c
@@ -82,6 +82,28 @@ static ucs_status_t uct_mm_iface_get_address(uct_iface_t *tl_iface,
     return uct_mm_md_mapper_ops(md)->iface_addr_pack(md, iface_addr + 1);
 }
 
+ucs_status_t
+uct_mm_iface_query_tl_devices(uct_md_h md,
+                              uct_tl_device_resource_t **tl_devices_p,
+                              unsigned *num_tl_devices_p)
+{
+    uct_md_attr_t md_attr;
+    ucs_status_t status;
+
+    status = uct_md_query(md, &md_attr);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    if (!(md_attr.cap.flags & UCT_MD_FLAG_ALLOC)) {
+        *num_tl_devices_p = 0;
+        *tl_devices_p     = NULL;
+        return UCS_ERR_NO_DEVICE;
+    }
+
+    return uct_sm_base_query_tl_devices(md, tl_devices_p, num_tl_devices_p);
+}
+
 static int
 uct_mm_iface_is_reachable(const uct_iface_h tl_iface,
                           const uct_device_addr_t *dev_addr,
--- a/src/uct/sm/mm/posix/mm_posix.c
+++ b/src/uct/sm/mm/posix/mm_posix.c
@@ -51,9 +51,10 @@
 
 
 typedef struct uct_posix_md_config {
-    uct_mm_md_config_t        super;
-    char                      *dir;
-    int                       use_proc_link;
+    uct_mm_md_config_t super;
+    char               *dir;
+    int                use_proc_link;
+    size_t             shm_min_size;
 } uct_posix_md_config_t;
 
 typedef struct uct_posix_packed_rkey {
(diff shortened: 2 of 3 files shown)


test/gtest/uct/uct_test.cc
```diff
@@ -417,7 +417,9 @@ std::vector<const resource*> uct_test::enum_resources(const std::string& tl_name
 
             uct_md_attr_t md_attr;
             status = uct_md_query(md, &md_attr);
-            ASSERT_UCS_OK(status);
+            if (status != UCS_OK) {
```
yosefe, line 420: should cause posix to not appear in enum_md_resources output when it's disqualified

src/uct/sm/mm/base/mm_iface.c
```diff
@@ -850,6 +850,28 @@ static UCS_CLASS_CLEANUP_FUNC(uct_mm_iface_t)
     ucs_arbiter_cleanup(&self->arbiter);
 }
 
+ucs_status_t
+uct_mm_iface_base_query_tl_devices(uct_md_h md,
```
gleon99, line 854: Why put this func specifically between `CLEANUP` and `DEFINE`?
Maybe move above, where the other helper funcs are?

src/uct/sm/mm/posix/mm_posix.c
```diff
@@ -21,6 +21,7 @@
 #include <uct/api/v2/uct_v2.h>
 
 
+#define UCT_POSIX_SHM_MIN_SIZE          16*UCS_MBYTE
```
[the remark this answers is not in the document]
ofirfarjun7: Maybe @yosefe can explain why we chose this specific value. I don't remember that it was for any special reason besides "It's suppose to be enough".

Regarding why we need min size, I will add a comment.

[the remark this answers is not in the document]
yosefe: it should be a configurable value, 16MB should be enough to allocate a recv memory pool. pls make sure it is larger than the docker default.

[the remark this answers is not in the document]
ofirfarjun7: I guess you meant that it is smaller than the docker default (docker default is 8GB...).

src/uct/sm/mm/base/mm_iface.c
```diff
@@ -82,6 +82,28 @@ static ucs_status_t uct_mm_iface_get_address(uct_iface_t *tl_iface,
     return uct_mm_md_mapper_ops(md)->iface_addr_pack(md, iface_addr + 1);
 }
 
+ucs_status_t
+uct_mm_iface_base_query_tl_devices(uct_md_h md,
```
yosefe, line 86: uct_mm_iface_query_tl_devices

src/uct/sm/mm/posix/mm_posix.c
```diff
@@ -64,20 +65,28 @@ typedef struct uct_posix_packed_rkey {
...
-  {NULL}
+    {"MM_", "", NULL, ucs_offsetof(uct_posix_md_config_t, super),
+     UCS_CONFIG_TYPE_TABLE(uct_mm_md_config_table)},
+
+    {"DIR", UCT_POSIX_SHM_OPEN_DIR,
+     "The path to the backing file. If it's equal to " UCT_POSIX_SHM_OPEN_DIR
+     " then \n"
+     "shm_open() is used. Otherwise, open() is used.",
+     ucs_offsetof(uct_posix_md_config_t, dir), UCS_CONFIG_TYPE_STRING},
+    {"SHM_MIN_SIZE", "16mb",
```
yosefe, line 76: space line before

src/uct/sm/mm/posix/mm_posix.c
```diff
@@ -64,20 +65,28 @@ typedef struct uct_posix_packed_rkey {
...
+     "This is the minimum size of the backing file that enforced by UCX.\n"
+     "Default value ensures sufficient resources to most use cases.\n"
+     "Limiting memory space can lead to undefined behavior of UCX.",
```
yosefe, line 79: Minimal size of the shared memory file system. If the file system size is less than this value, the transport will be disabled since it may not be able to allocate sufficient resources for communications.

src/uct/sm/mm/posix/mm_posix.c
```diff
@@ -122,15 +131,19 @@ uct_posix_md_query(uct_md_h tl_md, uct_md_attr_v2_t *md_attr)
...
         return UCS_ERR_NO_DEVICE;
     }
 
-    uct_mm_md_query(&md->super, md_attr,
-                    shm_statvfs.f_bsize * shm_statvfs.f_bavail);
+    shm_size = shm_statvfs.f_bsize * shm_statvfs.f_bavail;
+    uct_mm_md_query(&md->super, md_attr, shm_size);
+    if (shm_size < posix_config->shm_min_size) {
+        md_attr->alloc_mem_types = 0;
+    }
```
yosefe, line 146: there is a better way to do it by passing appropriate parameter to uct_mm_md_query as "max_alloc"

src/uct/sm/mm/posix/mm_posix.c
```diff
@@ -73,10 +73,12 @@ static ucs_config_field_t uct_posix_md_config_table[] = {
      " then \n"
      "shm_open() is used. Otherwise, open() is used.",
      ucs_offsetof(uct_posix_md_config_t, dir), UCS_CONFIG_TYPE_STRING},
+
     {"SHM_MIN_SIZE", "16mb",
-     "This is the minimum size of the backing file that enforced by UCX.\n"
-     "Default value ensures sufficient resources to most use cases.\n"
-     "Limiting memory space can lead to undefined behavior of UCX.",
+     "Minimal size of the shared memory file system.\n"
+     "If the file system size is less than this value, the transport\n"
```
yosefe, line 79: pls use longer lines in the desc

src/uct/sm/mm/posix/mm_posix.c
```diff
@@ -140,10 +142,10 @@ uct_posix_md_query(uct_md_h tl_md, uct_md_attr_v2_t *md_attr)
     }
 
     shm_size = shm_statvfs.f_bsize * shm_statvfs.f_bavail;
-    uct_mm_md_query(&md->super, md_attr, shm_size);
     if (shm_size < posix_config->shm_min_size) {
-        md_attr->alloc_mem_types = 0;
+        shm_size = 0;
     }
+    uct_mm_md_query(&md->super, md_attr, shm_size);
```
yosefe, line 148:
```c
    uct_mm_md_query(&md->super, md_attr,
                    (shm_size < posix_config->shm_min_size) ? 0 : shm_size);
```

yosefe (dismissed):
pls squash

yosefe (approved):

PR #8509, opened 2022-09-07
AZP: add to CI gtest with proto enable

avildema:
add to CI gtest with proto enable eq yes

### DIFF
--- a/contrib/test_jenkins.sh
+++ b/contrib/test_jenkins.sh
@@ -1291,10 +1291,34 @@ run_tests() {
 	do_distributed_task 0 4 run_release_mode_tests
 }
 
+run_test_proto_enable() {
+	export UCX_HANDLE_ERRORS=bt
+	export UCX_ERROR_SIGNALS=SIGILL,SIGSEGV,SIGBUS,SIGFPE,SIGPIPE,SIGABRT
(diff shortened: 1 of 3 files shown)


contrib/test_jenkins.sh
```diff
@@ -1291,9 +1291,30 @@ run_tests() {
...
+
+	# all are running gtest
+	run_gtest "default"
+}
+
 prepare
 try_load_cuda_env
-if [ -n "$JENKINS_RUN_TESTS" ] || [ -n "$RUN_TESTS" ]
+if [[ "$PROTO_ENABLE" == "yes" ]]; then
+	run_test_proto_enable
```
yosefe, line 1316: we can still do check_machine

evgeny-leksikov (changes requested):
CI has been failed with error. Can we report it as a warning?

evgeny-leksikov (approved):

contrib/test_jenkins.sh
yosefe:
```sh
if [ -n "$JENKINS_RUN_TESTS" ] || [ -n "$RUN_TESTS" ]
    check_machine
    if [[ "$PROTO_ENABLE" == "yes" ]]; then
        run_test_proto_enable
    else
        run_tests
    fi
fi
```

yosefe: @avildema pls squash + force push due to small PR

yosefe (approved):

PR #8489, opened 2022-09-01
UCP/API: Introduce exported memory handle

dmitrygx:
## What

Introduce exported memory handle API.

## Why ?

External memory handle will be used to allow using UCP operation on a memory allocated by another process on same or another process, but registered on the same network device.

## How ?

1. Introduce new value `UCP_EP_PARAMS_FLAGS_EXPORTED_MEMH` in `ucp_ep_params_flags_field` enumeration.
2. Introduce new value `UCP_MEM_MAP_PARAM_FIELD_EXPORTED_MEMH_BUFFER` in `ucp_mem_map_params_field` enumeration.
3. Introduce new value `UCP_MEM_MAP_EXPORTED` in enumeration which are used then as flags passed to `ucp_mem_map()`.
4. Introduce new field `exported_memh_buffer` in `ucp_mem_map_params_t` structure.
5. Move comments for each field before the value in `enum ucp_ep_params_flags_field`, `enum ucp_mem_map_params_field`, `ucp_mem_map()` flags enumerations to fit 80 symbols per line requirement.

PR #8516, opened 2022-09-09
TOOLS/PERF: Introduce perftest with exported memh API

dmitrygx:
## What

Introduce perftest with exported memh.

## Why ?

After merging #8489 perftest with exported memh could be added to show how the new API could be used.

## How ?

1. Add `ucx_peftest_daemon` which offloads communications using exported memory handle mapped to a memory shared by `ucx_perftest`.
2. Update `ucx_peftest` to communicate with `ucx_peftest_daemon` by means of AM-based protocol and exporting memory through the new API.
3. Update `ucx_peftest` to accept `ucx_peftest_daemon` addresses (client's and server's ones) via `-Z` command line argument.


yosefe:
replaced by #8767

PR #8518, opened 2022-09-09
COMMENT/TYPOS: fixed set of typos in comments

hoopoepg:


src/ucp/wireup/wireup_ep.c
```diff
@@ -645,7 +645,7 @@ int ucp_wireup_aux_ep_is_owner(ucp_wireup_ep_t *wireup_ep, uct_ep_h owned_ep)
     ucp_lane_index_t cm_lane_idx = ucp_ep_get_cm_lane(ucp_ep);
 
     return (wireup_ep->aux_ep == owned_ep) ||
-           /* Auxilliary EP can be WIREUP EP in case of it is on CM lane */
+           /* Auxillary EP can be WIREUP EP in case of it is on CM lane */
```
brminich, line 648: `Auxiliary` seems to be the correct one

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -475,7 +475,7 @@ static UCS_CLASS_CLEANUP_FUNC(uct_ud_ep_t)
                             uct_ud_ep_pending_cancel_cb, 0);
 
     if (!ucs_queue_is_empty(&self->tx.window)) {
-        ucs_debug("ep=%p id=%d conn_sn=%d has %d unacked packets",
+        ucs_debug("ep=%p id=%d conn_sn=%d has %d unpacked packets",
```
brminich, line 478: this is incorrect, the packets are unacked (even though it is not a real world)

src/uct/tcp/tcp.h
```diff
@@ -250,7 +250,7 @@ typedef struct uct_tcp_ep_put_ack_hdr {
 typedef struct uct_tcp_ep_put_completion {
     uct_completion_t              *comp;           /* User's completion passed to
                                                     * uct_ep_flush */
-    uint32_t                      wait_put_sn;     /* Sequence number of the last unacked
+    uint32_t                      wait_put_sn;     /* Sequence number of the last unpacked
```
brminich, line 253: no,  `unacked` op is meant here

hoopoepg: ok, fixed

src/ucp/wireup/wireup_ep.c
hoopoepg: reverted

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -475,7 +475,7 @@ static UCS_CLASS_CLEANUP_FUNC(uct_ud_ep_t)
                             uct_ud_ep_pending_cancel_cb, 0);
 
     if (!ucs_queue_is_empty(&self->tx.window)) {
-        ucs_debug("ep=%p id=%d conn_sn=%d has %d unacked packets",
+        ucs_debug("ep=%p id=%d conn_sn=%d has %d non-acked packets",
```
brminich, line 478: minor: imo, unacked sounds better

brminich (dismissed):

src/uct/ib/ud/base/ud_ep.c
hoopoepg: ok, reverted

yosefe (approved):

PR #8522, opened 2022-09-09
GTEST: fix blocking UCP cases

evgeny-leksikov:
## What
Fix blocking UCP cases

## Why ?
SW protocols and RNDV requires progress of both sides

### DIFF
--- a/test/gtest/ucp/test_ucp_wireup.cc
+++ b/test/gtest/ucp/test_ucp_wireup.cc
@@ -632,7 +632,9 @@ UCS_TEST_P(test_ucp_wireup_1sided, disconnect_nonexistent) {
     sender().destroy_worker();
 }
 
-UCS_TEST_P(test_ucp_wireup_1sided, disconnect_reconnect) {
+UCS_TEST_P(test_ucp_wireup_1sided, disconnect_reconnect,
+           /* RNDV requires progress on both sides - sender and receiver */
+           "RNDV_THRESH=inf") {
     sender().connect(&receiver(), get_ep_params());
     send_b(sender().ep(), 1000, 1);
     disconnect(sender());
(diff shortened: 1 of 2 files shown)


hoopoepg (dismissed):

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -632,7 +632,7 @@ UCS_TEST_P(test_ucp_wireup_1sided, disconnect_nonexistent) {
     sender().destroy_worker();
 }
 
-UCS_TEST_P(test_ucp_wireup_1sided, disconnect_reconnect) {
+UCS_TEST_P(test_ucp_wireup_1sided, disconnect_reconnect, "RNDV_THRESH=inf") {
```
yosefe, line 635: 1. do we need this for all tests in this test suite?
2. pls add comment why it's needed

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1459,22 +1466,19 @@ UCS_TEST_P(test_ucp_wireup_amo, relese_key_after_flush) {
     ucp_rkey_h rkey = get_rkey(sender());
     add_rkey(rkey);
 
-    ucs_status_t status = ucp_atomic_post(sender().ep(), UCP_ATOMIC_POST_OP_ADD,
-                                          m_send_data[0], sizeof(elem_type),
-                                          (uint64_t)&m_recv_data[0], rkey);
-    ASSERT_UCS_OK(status);
-
-    ucp_request_param_t param;
+    param.op_attr_mask = UCP_OP_ATTR_FIELD_DATATYPE;
```
yosefe, line 1469: minor: per C++, define param, reqs, status here and not function beginning

test/gtest/ucp/test_ucp_wireup.cc
evgeny-leksikov: > 1. do we need this for all tests in this test suite?

Rest cases send small messages, e.g. 0 or 8 bytes. Do you propose to set "RNDV_THRESH=inf" in `test_ucp_wireup_1sided::init()`?
> 2. pls add comment why it's needed

ok

yosefe: > Rest cases send small messages, e.g. 0 or 8 bytes. Do you propose to set "RNDV_THRESH=inf" in `test_ucp_wireup_1sided::init()`?

yes

evgeny-leksikov: [here](https://github.com/openucx/ucx/blob/eed3d5668fed4020804124eb0fcb6a5e93f83534/test/gtest/ucp/test_ucp_wireup.cc#L559) is an exception which may lead to confusion, IMO we can create a new class to override `init()` there or continue `UCS_TEST_P` using to keep these tests aligned

yosefe (approved):

PR #8524, opened 2022-09-11
UCS/VFS/LOG: Fix fork() child pid handling when assigning multiple atfork handlers

ofirfarjun7:
## What
Fix UCS logger wrongly handle fork() and prints wrong pid.

## Why ?
To reduce calls to getpid() process pid is cached on a static var. 
After calling to fork() process pid is changed and global vars are duplicated so the cached pid values of the child and parent are equal. 
UCS Logger handle that by attaching atfork handler to the child process. this handler change the log_pid var value to the child pid, 
but if the application attach more atfork handlers before UCS they will be called before UCS atfork handler and can cause the logger to run with the wrong pid...

From the Man:
"The handlers for each phase are called in a specified order: 
       the prepare handlers are called in reverse order of registration; 
       **the parent and child handlers are called in the order of registration**"

## How ?
By creating "init stage", as long as the init stage didn't complete UCS will call getpid() instead of using the cached value. Each time we call fork both child and parent will enter the init stage.

### DIFF
--- a/src/ucs/debug/log.c
+++ b/src/ucs/debug/log.c
@@ -41,7 +41,7 @@
     ucs_log_level_names[_level], (ucs_log_current_indent * 2), ""
 
 #define UCS_LOG_PROC_DATA_ARG() \
-    ucs_log_hostname, ucs_log_pid, ucs_log_get_thread_name()
+    ucs_log_hostname, ucs_log_get_pid(), ucs_log_get_thread_name()
 
 #define UCS_LOG_COMPACT_ARG(_tv)\
     UCS_LOG_TIME_ARG(_tv), UCS_LOG_PROC_DATA_ARG()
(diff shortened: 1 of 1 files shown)


ofirfarjun7:
TODO - add new application that attach multiple atfork handlers, use UCS and call fork to test the logger prints the correct pid.

yosefe (commented):
LGTM besides adding unit test

src/ucs/debug/log.c
```diff
@@ -91,6 +91,14 @@ static ucs_spinlock_t ucs_log_global_filter_lock;
 static khash_t(ucs_log_filter) ucs_log_global_filter;
 
 
+static inline int ucs_log_get_pid() {
```
yosefe, line 94: code style

yosefe (approved):
pls squash

yosefe (approved):

PR #8352, opened 2022-07-05
UCT/IB/RC: flush remote

hoopoepg:
- added RC_V/RC_MLX5 implementation for flush_remote
- updated tests itialization to use iface address

PR #8529, opened 2022-09-12
UCT/API: fixup for uct_ep_connect_to_ep_v2

hoopoepg:
## What
Fixed backward compatibility issue

## How ?
- due to compatibility issue updated uct_ep_connect_to_ep_v2 routine parameters
- added uct_ep_connect_to_ep_param_field_t enum to add ep_address length
- iface address will not be used in ep_connect_to_ep routine

### DIFF
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -297,6 +297,23 @@ typedef enum {
 
 /**
  * @ingroup UCT_RESOURCE
+ * @brief uct_ep_connect_to_ep_v2 operation fields and flags
+ * 
+ * The enumeration allows specifying which fields in @ref
+ * uct_ep_connect_to_ep_params_t are present and operation flags are used. It is
+ * used to enable backward compatibility support.
+ */
+typedef enum {
+    /** Device address length */
+    UCT_EP_CONNECT_TO_EP_PARAM_FIELD_DEVICE_ADDR_LENGTH = UCS_BIT(0),
+
+    /** Endpoint address length */
+    UCT_EP_CONNECT_TO_EP_PARAM_FIELD_EP_ADDR_LENGTH     = UCS_BIT(1)
+} uct_ep_connect_to_ep_param_field_t;
+
+
+/**
+ * @ingroup UCT_RESOURCE
  * @brief Endpoint attributes, capabilities and limitations.
  */
 struct uct_ep_attr {
@@ -480,9 +497,24 @@ typedef struct uct_iface_is_reachable_params {
  */
 typedef struct uct_ep_connect_to_ep_params {
     /**
-     * Reserved, must be 0.
+     * Mask of valid fields in this structure and operation flags, using
+     * bits from @ref uct_ep_connect_to_ep_param_field_t. Fields not specified
+     * in this mask will be ignored. Provides ABI compatibility with respect to
+     * adding new fields.
      */
     uint64_t                      field_mask;
+
+    /**
+     * Device address length. If not provided, the transport will assume a
+     * default minimal length according to the address buffer contents.
+     */
+    size_t                        device_addr_length;
+
+    /**
+     * Endpoint address length. If not provided, the transport will assume a
+     * default minimal length according to the address buffer contents.
+     */
+    size_t                        ep_addr_length;
 } uct_ep_connect_to_ep_params_t;
 
 
@@ -784,7 +816,6 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
  */
 ucs_status_t uct_ep_connect_to_ep_v2(uct_ep_h ep,
                                      const uct_device_addr_t *device_addr,
-                                     const uct_iface_addr_t *iface_addr,
                                      const uct_ep_addr_t *ep_addr,
                                      const uct_ep_connect_to_ep_params_t *params);
 
--- a/src/uct/base/uct_iface.c
+++ b/src/uct/base/uct_iface.c
@@ -621,14 +621,13 @@ ucs_status_t uct_ep_connect_to_ep(uct_ep_h ep, const uct_device_addr_t *dev_addr
 
 ucs_status_t uct_ep_connect_to_ep_v2(uct_ep_h ep,
                                      const uct_device_addr_t *device_addr,
-                                     const uct_iface_addr_t *iface_addr,
                                      const uct_ep_addr_t *ep_addr,
                                      const uct_ep_connect_to_ep_params_t *params)
 {
     const uct_base_iface_t *iface = ucs_derived_of(ep->iface, uct_base_iface_t);
 
-    return iface->internal_ops->ep_connect_to_ep_v2(ep, device_addr, iface_addr,
-                                                    ep_addr, params);
+    return iface->internal_ops->ep_connect_to_ep_v2(ep, device_addr, ep_addr,
+                                                    params);
 }
 
 ucs_status_t uct_cm_client_ep_conn_notify(uct_ep_h ep)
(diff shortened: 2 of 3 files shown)


src/uct/api/v2/uct_v2.h
```diff
@@ -480,9 +493,17 @@ typedef struct uct_iface_is_reachable_params {
...
+     * bits from @ref uct_ep_connect_to_ep_param_field_t. Fields not specified
+     * in this mask will be ignored. Provides ABI compatibility with respect to
+     * adding new fields.
      */
     uint64_t                      field_mask;
+
+    /**
+     * Endpoint address length
+     */
+    uint8_t                       ep_addr_length;
```
brminich, line 506: maybe better `size_t` or `int`?

hoopoepg: ok, used size_t

src/uct/api/v2/uct_v2.h
```diff
@@ -480,9 +493,17 @@ typedef struct uct_iface_is_reachable_params {
...
+     * bits from @ref uct_ep_connect_to_ep_param_field_t. Fields not specified
+     * in this mask will be ignored. Provides ABI compatibility with respect to
+     * adding new fields.
      */
     uint64_t                      field_mask;
+
+    /**
+     * Endpoint address length
+     */
+    size_t                        ep_addr_length;
```
yosefe, line 506: let's add also device_addr_length?

src/uct/api/v2/uct_v2.h
```diff
@@ -480,9 +493,17 @@ typedef struct uct_iface_is_reachable_params {
...
-     * Reserved, must be 0.
+     * Mask of valid fields in this structure and operation flags, using
+     * bits from @ref uct_ep_connect_to_ep_param_field_t. Fields not specified
+     * in this mask will be ignored. Provides ABI compatibility with respect to
+     * adding new fields.
      */
     uint64_t                      field_mask;
+
+    /**
+     * Endpoint address length
```
yosefe, line 504: what happens if not specified?

hoopoepg: implementation-specific, TL will have to detect it another way

src/uct/api/v2/uct_v2.h
hoopoepg: ok, added device add len

brminich (approved):

src/uct/api/v2/uct_v2.h
```diff
@@ -480,9 +497,22 @@ typedef struct uct_iface_is_reachable_params {
...
+     */
+    size_t                        ep_addr_length;
+
+    /**
+     * Device address length
+     */
+    size_t                        dev_addr_length;
```
yosefe, line 515: device_addr_length

src/uct/api/v2/uct_v2.h
```diff
@@ -295,6 +295,23 @@ typedef enum {
...
+ * The enumeration allows specifying which fields in @ref
+ * uct_ep_connect_to_ep_params_t are present and operation flags are used. It is
+ * used to enable backward compatibility support.
+ */
+typedef enum {
+    /** EP address length */
+    UCT_EP_CONNECT_TO_EP_PARAM_FIELD_EP_ADDR_LENGTH  = UCS_BIT(0),
+
+    /** Device address length */
+    UCT_EP_CONNECT_TO_EP_PARAM_FIELD_DEV_ADDR_LENGTH = UCS_BIT(1)
```
yosefe, line 311: UCT_EP_CONNECT_TO_EP_PARAM_FIELD_DEVICE_ADDR_LENGTH

yosefe: device address should be first according to parameters order

src/uct/api/v2/uct_v2.h
hoopoepg: renamed & reordered values in structure

brminich (approved):

src/uct/api/v2/uct_v2.h
```diff
@@ -480,9 +497,22 @@ typedef struct uct_iface_is_reachable_params {
...
      */
     uint64_t                      field_mask;
+
+    /**
+     * Device address length
+     */
+    size_t                        device_addr_length;
+
+    /**
+     * Endpoint address length
```
yosefe, line 513: pls document what happens if length is not provided

hoopoepg: added comment

src/uct/api/v2/uct_v2.h
```diff
@@ -480,9 +497,24 @@ typedef struct uct_iface_is_reachable_params {
...
+     * Mask of valid fields in this structure and operation flags, using
+     * bits from @ref uct_ep_connect_to_ep_param_field_t. Fields not specified
+     * in this mask will be ignored. Provides ABI compatibility with respect to
+     * adding new fields.
      */
     uint64_t                      field_mask;
+
+    /**
+     * Device address length. If not provided transport should detect address
+     * length another way.
```
yosefe, line 509: If not provided, the transport will assume a default minimal length according to the address buffer contents.

same for ep.

yosefe (approved):

hoopoepg:
@yosefe ok to merge?

yosefe:
@shamisp @tonycurtis can you pls review?

tonycurtis (dismissed):

brminich (approved):
@shamisp, please take a look

shamisp:
Can you please provide additional details about backward compatibility issue ?

hoopoepg:
> Can you please provide additional details about backward compatibility issue ?

we were going to extend rc_x iface address flags by new bits, but we found that in older versions of UCX there is exact comparing of flags with some values and adding non-zero bits to address bytes breaks compatibility

hoopoepg:
https://github.com/openucx/ucx/pull/8352 - this is regression PR

brminich:
It appeared that we can't extend iface address, like it is done in #8352 without breaking wire compatibility. 
So the new approach is to introduce new API which also accepts address length, so that the transport can determine (in compatible way) whether the address contains an extensions

src/uct/api/v2/uct_v2.h
```diff
@@ -295,6 +295,23 @@ typedef enum {
...
+ * 
+ * The enumeration allows specifying which fields in @ref
+ * uct_ep_connect_to_ep_params_t are present and operation flags are used. It is
+ * used to enable backward compatibility support.
+ */
+typedef enum {
+    /** Device address length */
+    UCT_EP_CONNECT_TO_EP_PARAM_FIELD_DEVICE_ADDR_LENGTH = UCS_BIT(0),
+
+    /** EP address length */
```
shamisp, line 310: EP -> endpoint

src/uct/api/v2/uct_v2.h
```diff
@@ -480,9 +497,24 @@ typedef struct uct_iface_is_reachable_params {
...
+     * Mask of valid fields in this structure and operation flags, using
+     * bits from @ref uct_ep_connect_to_ep_param_field_t. Fields not specified
+     * in this mask will be ignored. Provides ABI compatibility with respect to
+     * adding new fields.
      */
     uint64_t                      field_mask;
+
+    /**
+     * Device address length. If not provided, the transport will assume a
+     * default minimal length according to the address buffer contents.
```
shamisp, line 509: can we suggest to developer how one can find current minimum

src/uct/api/v2/uct_v2.h
```diff
@@ -480,9 +497,24 @@ typedef struct uct_iface_is_reachable_params {
...
+     */
+    size_t                        device_addr_length;
+
+    /**
+     * Endpoint address length. If not provided, the transport will assume a
+     * default minimal length according to the address buffer contents.
```
shamisp, line 515: similar to the above

shamisp (changes requested):

shamisp (approved):

PR #8530, opened 2022-09-12
UCT/ROCM: make rocm components compile again

edgargabriel:
## What
the md_attr_v2 commit broke the rocm compilation. This commit fixes the compilation part.

Note: gtest generates even with this fix a segfault in rocm_copy/test_p2p_rma_madvise, I will investigate that later this week.


yosefe (approved):

yosefe:
@edgargabriel WDYT to add a container with ROCm build environment to the build so we'll catch these?
it could be added like this #8532 (as example)

edgargabriel:
@yosefe yes, that is definitely a good idea. I will look into this.

PR #8535, opened 2022-09-13
UCP/PROTO: add more ucp_proto_t::reset

evgeny-leksikov:
## What
add more `ucp_proto_t::reset`
 - RNDV
 - sync send

### DIFF
--- a/src/ucp/rma/get_am.c
+++ b/src/ucp/rma/get_am.c
@@ -108,5 +108,5 @@ ucp_proto_t ucp_get_am_bcopy_proto = {
     .query    = ucp_proto_single_query,
     .progress = {ucp_proto_get_am_bcopy_progress},
     .abort    = (ucp_request_abort_func_t)ucs_empty_function_fatal_not_implemented_void,
-    .reset    = ucp_proto_get_reset
+    .reset    = ucp_proto_request_bcopy_id_reset
 };
--- a/src/ucp/rndv/proto_rndv.c
+++ b/src/ucp/rndv/proto_rndv.c
@@ -413,13 +413,18 @@ void ucp_proto_rndv_rts_query(const ucp_proto_query_params_t *params,
 
 void ucp_proto_rndv_rts_abort(ucp_request_t *req, ucs_status_t status)
 {
-    if (req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED) {
-        ucp_send_request_id_release(req);
-        ucp_datatype_iter_mem_dereg(req->send.ep->worker->context,
-                                    &req->send.state.dt_iter, UCP_DT_MASK_ALL);
+    ucp_proto_rndv_rts_reset(req);
+    ucp_request_complete_send(req, status);
+}
+
+void ucp_proto_rndv_rts_reset(ucp_request_t *req)
+{
+    if (!(req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED)) {
+        return;
     }
 
-    ucp_request_complete_send(req, status);
+    ucp_send_request_id_release(req);
+    ucp_proto_request_zcopy_clean(req, UCP_DT_MASK_ALL);
 }
 
 static ucs_status_t
@@ -671,8 +676,10 @@ void ucp_proto_rndv_receive_start(ucp_worker_h worker, ucp_request_t *recv_req,
     uint8_t sg_count;
     ucp_ep_h ep;
 
-    UCP_WORKER_GET_VALID_EP_BY_ID(&ep, worker, rts->sreq.ep_id, return,
-                                  "RTS on non-existing endpoint");
+    UCP_WORKER_GET_VALID_EP_BY_ID(&ep, worker, rts->sreq.ep_id, {
+        ucp_proto_rndv_recv_super_complete_status(recv_req, UCS_ERR_CANCELED);
+        return;
+    }, "RTS on non-existing endpoint");
 
     req = ucp_request_get(worker);
     if (req == NULL) {
@@ -791,10 +798,8 @@ ucp_proto_rndv_handle_rtr(void *arg, void *data, size_t length, unsigned flags)
         /* RTR covers the whole send request - use the send request directly */
         ucs_assert(rtr->offset == 0);
 
-        ucp_datatype_iter_mem_dereg(worker->context, &req->send.state.dt_iter,
-                                    UCP_DT_MASK_ALL);
         ucp_send_request_id_release(req);
-        req->flags &= ~UCP_REQUEST_FLAG_PROTO_INITIALIZED;
+        ucp_proto_request_zcopy_clean(req, UCP_DT_MASK_ALL);
 
         sg_count = req->send.proto_config->select_param.sg_count;
         status   = ucp_proto_rndv_send_start(worker, req, 0, op_flags, rtr,
--- a/src/ucp/rndv/rndv_ats.c
+++ b/src/ucp/rndv/rndv_ats.c
@@ -52,29 +52,13 @@ ucp_proto_rndv_ats_init(const ucp_proto_init_params_t *init_params)
     return status;
 }
 
-static ucs_status_t
-ucp_proto_rndv_ats_proto_progress(uct_pending_req_t *uct_req)
-{
-    ucp_request_t *req = ucs_container_of(uct_req, ucp_request_t, send.uct);
-
-    if (!(req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED)) {
-        if (req->send.rndv.rkey != NULL) {
-            ucp_proto_rndv_rkey_destroy(req);
-        }
-
-        ucp_datatype_iter_cleanup(&req->send.state.dt_iter, UCP_DT_MASK_ALL);
-        req->flags |= UCP_REQUEST_FLAG_PROTO_INITIALIZED;
-    }
-
-    return ucp_proto_rndv_ats_progress(uct_req);
-}
-
 ucp_proto_t ucp_rndv_ats_proto = {
     .name     = "rndv/ats",
     .desc     = "no data fetch",
     .flags    = 0,
     .init     = ucp_proto_rndv_ats_init,
     .query    = ucp_proto_default_query,
-    .progress = {ucp_proto_rndv_ats_proto_progress},
-    .abort    = (ucp_request_abort_func_t)ucs_empty_function_do_assert_void
+    .progress = {ucp_proto_rndv_ats_progress},
+    .abort    = (ucp_request_abort_func_t)ucs_empty_function_fatal_not_implemented_void,
+    .reset    = (ucp_request_reset_func_t)ucs_empty_function
 };
--- a/src/ucp/rndv/rndv_get.c
+++ b/src/ucp/rndv/rndv_get.c
@@ -208,6 +208,28 @@ static void ucp_rndv_get_zcopy_proto_abort(ucp_request_t *request,
     }
 }
 
+static void ucp_rndv_get_zcopy_proto_reset(ucp_request_t *req)
+{
+    if (!(req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED)) {
+        return;
+    }
+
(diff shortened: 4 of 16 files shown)


src/ucp/rma/get_am.c
```diff
@@ -100,6 +100,14 @@ ucp_proto_get_am_bcopy_init(const ucp_proto_init_params_t *init_params)
     return ucp_proto_single_init(&params);
 }
 
+static void ucp_proto_get_am_bcopy_reset(ucp_request_t *request)
+{
+    if (request->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED) {
```
yosefe, line 105: why do we check UCP_REQUEST_FLAG_PROTO_INITIALIZED here, but not in ucp_proto_request_sync_[b/z]copy_reset?

src/ucp/rndv/proto_rndv.c
```diff
@@ -668,8 +689,10 @@ void ucp_proto_rndv_receive_start(ucp_worker_h worker, ucp_request_t *recv_req,
     uint8_t sg_count;
     ucp_ep_h ep;
 
-    UCP_WORKER_GET_VALID_EP_BY_ID(&ep, worker, rts->sreq.ep_id, return,
-                                  "RTS on non-existing endpoint");
+    UCP_WORKER_GET_VALID_EP_BY_ID(&ep, worker, rts->sreq.ep_id, {
+        ucp_rndv_recv_req_complete(recv_req, UCS_ERR_CANCELED);
```
yosefe, line 693: IMO we should call request abort here for consistency
and better not reuse protov1 functions from protov2

src/ucp/rndv/proto_rndv.c
```diff
@@ -558,14 +566,27 @@ static size_t ucp_proto_rndv_ats_pack_ack(void *dest, void *arg)
...
+
 ucs_status_t ucp_proto_rndv_ats_progress(uct_pending_req_t *uct_req)
 {
     ucp_request_t *req = ucs_container_of(uct_req, ucp_request_t, send.uct);
 
     return ucp_proto_rndv_ack_progress(req, req->send.proto_config->priv,
                                        UCP_AM_ID_RNDV_ATS,
                                        ucp_proto_rndv_ats_pack_ack,
-                                       ucp_proto_rndv_recv_complete);
+                                       ucp_proto_rndv_ats_complete);
```
yosefe, line 589: why need to change this? now there is extra branch (and call to ucp_datatype_iter_cleanup) in rndv/get protocols as well
we probably need to have special "ucp_proto_rndv_ats_progress" replacement only for rndv/ats protocol - that means, fully implement ucp_proto_rndv_ats_proto_progress function without calling ucp_proto_rndv_ats_progress

src/ucp/rndv/rndv_ats.c
```diff
@@ -76,5 +68,6 @@ ucp_proto_t ucp_rndv_ats_proto = {
     .init     = ucp_proto_rndv_ats_init,
     .query    = ucp_proto_default_query,
     .progress = {ucp_proto_rndv_ats_proto_progress},
-    .abort    = (ucp_request_abort_func_t)ucs_empty_function_do_assert_void
+    .abort    = (ucp_request_abort_func_t)ucs_empty_function_fatal_not_implemented_void,
```
yosefe, line 71: why not implemented?

src/ucp/rndv/rndv_get.c
```diff
@@ -207,6 +207,23 @@ static void ucp_rndv_get_zcopy_proto_abort(ucp_request_t *request,
...
 }
 
+static void ucp_rndv_get_zcopy_proto_reset(ucp_request_t *request)
+{
+    switch (request->send.proto_stage) {
+    case UCP_PROTO_RNDV_GET_STAGE_FETCH:
+        ucp_proto_request_zcopy_reset(request);
+        break;
+    case UCP_PROTO_RNDV_GET_STAGE_ATS:
+        ucp_proto_request_bcopy_reset(request);
```
yosefe, line 217: looks weird that we call "bcopy" reset from get_zcopy_proto_reset
maybe break to complements with "fall-through" ?

src/ucp/rndv/rndv_ppln.c
```diff
@@ -179,6 +179,13 @@ void ucp_proto_rndv_ppln_send_frag_complete(ucp_request_t *freq, int send_ack)
                                       "ppln_send");
 }
 
+void ucp_proto_rndv_ppln_recv_frag_clean(ucp_request_t *freq)
+{
+    ucp_proto_rndv_ppln_frag_complete(
```
yosefe, line 184: this function can send ACK to remote peer, is this ok?

src/ucp/rndv/rndv_rtr.c
```diff
@@ -235,6 +240,14 @@ static void ucp_proto_rndv_rtr_abort(ucp_request_t *req, ucs_status_t status)
     rpriv->data_received(req, 0);
 }
 
+static void
+ucp_proto_rndv_rtr_reset(ucp_request_t *req)
+{
+    ucp_proto_rndv_rtr_clean(req, UCP_DT_MASK_ALL);
+    ucp_send_request_id_release(req);
+    req->flags &= ~UCP_REQUEST_FLAG_PROTO_INITIALIZED;
```
yosefe, line 248: we assume UCP_REQUEST_FLAG_PROTO_INITIALIZED was set?

evgeny-leksikov: need to double check resource allocation

src/ucp/rndv/rndv_ppln.c
evgeny-leksikov: it can't because parameter `send_ack=0`

src/ucp/rndv/proto_rndv.c
evgeny-leksikov: Hmm.. gtest passes rndv tests with origin code, so just reverting for now

src/ucp/rma/get_am.c
evgeny-leksikov: probably because setting the flag was missed in `ucp_proto_eager_sync_tag_offload_bcopy_progress`, now it's aligned

src/ucp/rndv/rndv.c
```diff
@@ -336,8 +336,7 @@ void ucp_rndv_req_send_ack(ucp_request_t *ack_req, size_t ack_size,
     ucp_request_send(ack_req);
 }
 
-static UCS_F_ALWAYS_INLINE void
```
yosefe, line 339: still needed?

src/ucp/rndv/rndv.h
```diff
@@ -81,6 +81,8 @@ void ucp_rndv_req_send_ack(ucp_request_t *ack_req, size_t ack_size,
                            ucs_ptr_map_key_t remote_req_id, ucs_status_t status,
                            ucp_am_id_t am_id, const char *ack_str);
 
+void ucp_rndv_recv_req_complete(ucp_request_t *req, ucs_status_t status);
```
yosefe, line 84: still needed?

src/ucp/rndv/rndv_ats.c
```diff
@@ -57,15 +57,7 @@ ucp_proto_rndv_ats_proto_progress(uct_pending_req_t *uct_req)
...
-    if (!(req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED)) {
-        if (req->send.rndv.rkey != NULL) {
-            ucp_proto_rndv_rkey_destroy(req);
-        }
-
-        ucp_datatype_iter_cleanup(&req->send.state.dt_iter, UCP_DT_MASK_ALL);
-        req->flags |= UCP_REQUEST_FLAG_PROTO_INITIALIZED;
-    }
-
+    req->flags |= UCP_REQUEST_FLAG_PROTO_INITIALIZED;
```
yosefe, line 60: maybe don't need this flag for rndv/ats protocol since nothing is really initialized here?

yosefe (changes requested):
in general i think there is an issue that things initialized before UCP_REQUEST_FLAG_PROTO_INITIALIZED is set (such as dt_iter and unpacked rkey) are cleaned under the condition of UCP_REQUEST_FLAG_PROTO_INITIALIZED check
let's sync on this

src/ucp/rndv/rndv_get.c
```diff
@@ -208,6 +208,32 @@ static void ucp_rndv_get_zcopy_proto_abort(ucp_request_t *request,
...
 }
 
+static void ucp_rndv_get_zcopy_proto_reset(ucp_request_t *request)
+{
+    if (!(request->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED)) {
+        return;
+    }
+
+    request->flags &= ~UCP_REQUEST_FLAG_PROTO_INITIALIZED;
+    ucp_proto_request_check_reset_state(request);
```
yosefe, line 218: why are we using this function only from ucp_rndv_get_zcopy_proto_reset and not other protocols?

src/ucp/rndv/rndv_get.c
```diff
@@ -208,6 +208,32 @@ static void ucp_rndv_get_zcopy_proto_abort(ucp_request_t *request,
...
+    ucp_datatype_iter_cleanup(&request->send.state.dt_iter, UCP_DT_MASK_ALL);
```
yosefe, line 219: do we really need datatype iter cleanup here?

src/ucp/rndv/rndv_rtr.c
```diff
@@ -113,15 +113,20 @@ ucp_proto_rndv_rtr_hdr_pack(ucp_request_t *req, ucp_rndv_rtr_hdr_t *rtr,
 }
 
 static UCS_F_ALWAYS_INLINE void
-ucp_proto_rndv_rtr_common_complete(ucp_request_t *req, unsigned dt_mask)
+ucp_proto_rndv_rtr_clean(ucp_request_t *req, unsigned dt_mask)
```
yosefe, line 116: seems ucp_proto_rndv_rtr_clean used only once

src/ucp/tag/offload/eager.c
```diff
@@ -326,6 +327,8 @@ ucp_proto_eager_sync_tag_offload_zcopy_send_completion(uct_completion_t *self)
                                           send.state.uct_comp);
 
     ucp_proto_request_zcopy_clean(req, UCS_BIT(UCP_DATATYPE_CONTIG));
+    ucp_datatype_iter_cleanup(&req->send.state.dt_iter,
+                              UCS_BIT(UCP_DATATYPE_CONTIG));
```
yosefe, line 331: is it needed?

src/ucp/rndv/rndv_get.c
evgeny-leksikov: All basic protocols are checked in `ucp_proto_request_restart`, rndv get has several stages and super request, so I decided added extra check. @yosefe Do you think it's redundant here or need to move common check to every `*_reset` function?

src/ucp/tag/offload/eager.c
evgeny-leksikov: Basically, this is a common approach for all protocols to cleanup iterator on completion, the fact of `UCP_DATATYPE_CONTIG` iterator cleanup is noop, it's implementation details which can be changed, so I think it should be here.

src/ucp/rndv/rndv_get.c
yosefe: why the check in ucp_proto_request_restart() is not enough for rndv?

src/ucp/tag/offload/eager.c
yosefe: ok, missed that it's not reset

src/ucp/rndv/rndv_get.c
evgeny-leksikov: I believe it should be enough, added just because rndv looks more complex, will remove then .

yosefe (dismissed):

yosefe (approved):

PR #8537, opened 2022-09-14
UCT/WIREUP: enable RMA BW lanes on RMA_FEATURE

hoopoepg:
- enabled RMA BW lane selection in case if context was created with UCP_FEATURE_RMA enabled


yosefe (approved):

brminich:
do we need to add the same to `ucp_wireup_add_am_bw_lanes` for SW RMA emulation?

hoopoepg:
> do we need to add the same to `ucp_wireup_add_am_bw_lanes` for SW RMA emulation?

not needed: for now SW RMA is not used in multirail

brminich (approved):

hoopoepg:
@yosefe ok to merge?

PR #8541, opened 2022-09-19
UCT/BASE: Handle exported_mkey_packed_size and global_id field in uct_md_attr_v2_copy

dmitrygx:
## What

Handle `exported_mkey_packed_size` and `global_id` fields in `uct_md_attr_v2_copy`.

## Why ?

They should be handle to make sure copying of MD attributes from temporary attributes to user's ones work correctly in `uct_md_attr_v2`.

## How ?

Invoke `UCT_MD_ATTR_V2_FIELD_COPY` for `exported_mkey_packed_size` and `global_id` field in `uct_md_attr_v2_copy` auxiliary function.


dmitrygx:

brminich (approved):
approved, but is not really needed without implementation

dmitrygx:
> approved, but is not really needed without implementation

it is just for consistency to have copying of all fields of MD attributes v2

yosefe (approved):

PR #8543, opened 2022-09-20
GTEST/UCT: Use uct_md_query_v2() instead of uct_md_query()

dmitrygx:
## What

Use uct_md_query_v2() instead of uct_md_query().

## Why ?

To make sure that tests have an access to new fields introduced in `uct_md_query_v2`'s attributes.

## How ?

Find `uct_md_attr_t` and replace by `uct_md_attr_v2_t` and adopt its usage.


brminich (approved):

dmitrygx:
@yosefe could you review pls?

yosefe (approved):

PR #8544, opened 2022-09-20
UCT/IFACE/CAPS: added relaxed ordering capability

hoopoepg:
## What
Added UCT_IFACE_FLAG_RELAXED_ORDERING capability to mark ifaces where delivery ordering is not guaranteed

## Why ?
Some transports could not provide strong data delivery ordering and UCX has to use more robust algorithms
to provide data integrity

## How ?
Added new capability

### DIFF
--- a/src/uct/api/uct.h
+++ b/src/uct/api/uct.h
@@ -427,6 +427,10 @@ typedef enum uct_atomic_op {
 #define UCT_IFACE_FLAG_TAG_EAGER_BCOPY UCS_BIT(51) /**< Hardware tag matching bcopy eager support */
 #define UCT_IFACE_FLAG_TAG_EAGER_ZCOPY UCS_BIT(52) /**< Hardware tag matching zcopy eager support */
 #define UCT_IFACE_FLAG_TAG_RNDV_ZCOPY  UCS_BIT(53) /**< Hardware tag matching rendezvous zcopy support */
+
+        /* Ordering flags */
+#define UCT_IFACE_FLAG_RELAXED_ORDERING UCS_BIT(54) /**< Interface can't guarantee data
+                                                         delivery ordering */
 /**
  * @}
  */


src/uct/api/uct.h
```diff
@@ -427,6 +427,10 @@ typedef enum uct_atomic_op {
 #define UCT_IFACE_FLAG_TAG_EAGER_BCOPY UCS_BIT(51) /**< Hardware tag matching bcopy eager support */
 #define UCT_IFACE_FLAG_TAG_EAGER_ZCOPY UCS_BIT(52) /**< Hardware tag matching zcopy eager support */
 #define UCT_IFACE_FLAG_TAG_RNDV_ZCOPY  UCS_BIT(53) /**< Hardware tag matching rendezvous zcopy support */
+
+        /* Ordering flags */
+#define UCT_IFACE_FLAG_RELAXED_ORDERING UCS_BIT(54) /**< Interface can't guarantee data
```
shamisp, line 432: Do you think single flag is enough for all types of communication semantics ? For example, one transport can provide order for send/receive but no order for put/gets, etc.

shamisp: Also, what transport is going to be imitate user of this

hoopoepg: > Also, what transport is going to be imitate user of this

IB over adaptive routing

yosefe: > Do you think single flag is enough for all types of communication semantics ? For example, one transport can provide order for send/receive but no order for put/gets, etc.

It should be defined only for active messages. Put/get/atomic are already defined as weak order in UCT API.
Also, we may consider adding iface_open feature flag to "force" strict ordering (or to allow weak ordering if it improves BW)

shamisp: Can we go the other way around and assume that AM is relaxed by default and we will be adding strict order flag ? 
BTW, If this is just for AM, we have to make it AM specific.

PR #8547, opened 2022-09-20
UCP/PROTO: RMA and AMO restart

evgeny-leksikov:
## What
RMA and AMO proto_v2 request restart support


src/ucp/rma/amo_send.c
```diff
@@ -219,6 +219,15 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_atomic_op_nbx,
                 ep, &ucp_rkey_config(ep->worker, rkey)->proto_select,
                 rkey->cfg_index, req, op_id, buffer, 1, param->datatype,
                 op_size, param, 0, 0);
+        if (ucs_likely(req->flags & UCP_REQUEST_FLAG_PROTO_AMO_PACKED)) {
```
yosefe, line 222: is it possible to set AMO_PACKED flag only for t protocols, to avoid the branch on "op_id == UCP_OP_ID_AMO_POST" in line 264?

evgeny-leksikov: It depends on if `pack_arg` must be called for fetch (non-post) ops as well. As far as I can see, in current implementation, it does.

evgeny-leksikov: IMO it's confusing since buffered (packed) value actually does not complete operation. Also such flag will have different behavior for different ops.

yosefe: ok

yosefe (approved):

PR #8548, opened 2022-09-20
UCT/ROCM: send-to-self fixes

edgargabriel:
## What
1. Cleanup handling of remote_addr vs. iov->buffer in rocm_copy_ep_zcopy.
For put operations, src has to be iov->buffer and dst has to be remote_addr; for get operations exactly the other way around.

2. the rkey does not necessarily contain the registration information of the host buffer, it might be the rkey of the device buffer. Hence, we need to check for that and locally 'lock' the host buffer using hsa_amd_memory_lock() if the provided rkey  belongs to a device buffer.

3. Introduce the ability to perform host-to-host memcpy in the component, since the rocm/copy component  is also selected for host-to-host operations in send-to-self scenario. 

## Why ?
fixes a number of issues observed with send-to-self operations with Open MPI + UCX  with ROCm devices

### DIFF
--- a/src/uct/rocm/copy/rocm_copy_ep.c
+++ b/src/uct/rocm/copy/rocm_copy_ep.c
@@ -11,6 +11,7 @@
 #include "rocm_copy_ep.h"
 #include "rocm_copy_iface.h"
 #include "rocm_copy_md.h"
+#include "rocm_copy_cache.h"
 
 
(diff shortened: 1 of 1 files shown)


edgargabriel:
putting it on hold for a bit, I might have a better solution. Will reopen when done.

PR #8550, opened 2022-09-20
UCP/CORE: Check result md_map inside ucp_mem_map()

dmitrygx:
## What

Check result md_map inside `ucp_mem_map()` and report an error if it is empty.

## Why ?

Memory handle registration/allocation should be considered as invalid, if no Memory Domains could be used to register memory.

## How ?

1. Combine `ucp_memh_create()` and `ucp_memh_init_uct_reg()` to `ucp_memh_reg()` function.
2. After calling `ucp_mem_alloc()`/`ucp_memh_reg()` check `memh->md_map == 0`. If it is true, report an error and deallocate a created memory handle.

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -630,6 +630,34 @@ out:
     return status;
 }
 
+static ucs_status_t
+ucp_memh_reg(ucp_context_h context, void *address, size_t length,
+             ucs_memory_type_t mem_type, unsigned uct_flags, ucp_mem_h *memh_p)
+{
+    ucs_status_t status;
+    ucp_mem_h memh;
+
+    status = ucp_memh_create(context, address, length, mem_type,
+                             UCT_ALLOC_METHOD_LAST, &memh);
+    if (status != UCS_OK) {
+        goto out;
+    }
+
+    status = ucp_memh_init_uct_reg(context, memh, uct_flags);
+    if (status != UCS_OK) {
+        goto err_free_memh;
+    }
+
+    *memh_p = memh;
+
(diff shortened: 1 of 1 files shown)


dmitrygx:
@Artemy-Mellanox @brminich could you review pls?

src/ucp/core/ucp_mm.c
```diff
@@ -713,31 +741,30 @@ ucs_status_t ucp_mem_map(ucp_context_h context, const ucp_mem_map_params_t *para
...
 
-        status = ucp_memh_init_uct_reg(context, memh, uct_flags);
-        if (status != UCS_OK) {
-            goto err_free_memh;
-        }
+    if (status != UCS_OK) {
+        goto out;
+    }
+
+    if (memh->md_map == 0) {
```
Artemy-Mellanox, line 755: Can this happen if `context->reg_md_map[mem_type]` is not zero?
If not - can't we just fail while validating params?

src/ucp/core/ucp_mm.c
```diff
@@ -630,6 +630,34 @@ ucp_memh_alloc(ucp_context_h context, void *address, size_t length,
     return status;
 }
 
+static ucs_status_t
+ucp_memh_reg(ucp_context_h context, void *address, size_t length,
```
Artemy-Mellanox, line 634: Maybe pick some better name for this functons. It's hard to tell the difference between `ucp_memh_reg`, `ucp_memh_register` and `ucp_memh_init_uct_reg`

dmitrygx:
will be combined with #8552, because need to check md_map only for imported memhs

PR #8552, opened 2022-09-21
UCP/CORE: Implement exported/imported memh

dmitrygx:
## What

 Implement exported/imported memh.

## Why ?

To implement support of memory handles which could be used on DPU side to perform AM/TAG/RMA/AMO operations without involving HOST side into this process.

## How ?

1. Add implementation of API introduced in https://github.com/openucx/ucx/pull/8489.
2. Add RCACHE for exported/imported memory handles.


dmitrygx:
replaced by #8584

PR #8553, opened 2022-09-21
UCP/CORE/WIREUP: Replace err_mode by flags in EP config key and wireup_msg

dmitrygx:
## What

Replace `err_mode` by `flags` in EP config key and WIREUP messages.

## Why ?

It is needed to add new flags which should be packed and sent to a peer to control transport selection, e.g. EXPORTED_MEMH feature.

## How ?

1. Replace `err_mode` by `flags` in `ucp_ep_config_key` and `ucp_wireup_msg` structures.
2. Calculate `err_mode` based on EP configuration's flags where it is needed.

### DIFF
--- a/src/ucp/wireup/wireup.c
+++ b/src/ucp/wireup/wireup.c
@@ -195,8 +195,13 @@ ucp_wireup_msg_prepare(ucp_ep_h ep, uint8_t type,
                             UCP_ADDRESS_PACK_FLAG_TL_RSC_IDX;
     ucs_status_t status;
 
+    /* They must be equal to keep wire compatibility with previous UCP versions
+     * which pack error handling mode only */
+    UCS_STATIC_ASSERT((int)UCP_ERR_HANDLING_MODE_PEER ==
+                      (int)UCP_EP_CONFIG_KEY_FLAG_ERR_HANDLING_MODE_PEER);
+
     msg_hdr->type      = type;
-    msg_hdr->err_mode  = ucp_ep_config(ep)->key.err_mode;
+    msg_hdr->flags     = ucp_ep_config(ep)->key.flags;
     msg_hdr->conn_sn   = ep->conn_sn;
     msg_hdr->src_ep_id = ucp_ep_local_id(ep);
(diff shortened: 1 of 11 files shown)


src/ucp/wireup/wireup.h
```diff
@@ -110,8 +110,10 @@ typedef struct {
  */
 typedef struct ucp_wireup_msg {
     uint8_t                type; /* Message type */
-    uint8_t                err_mode; /* Peer error handling mode defined in
-                                        @ucp_err_handling_mode_t */
+    uint8_t                flags; /* Peer error handling mode defined in
```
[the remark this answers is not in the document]
dmitrygx: no, it shouldn't.
i've added static assertion to make sure that `UCP_ERR_HANDLING_MODE_PEER == UCP_EP_CONFIG_KEY_FLAG_ERR_HANDLING_MODE_PEER`

src/ucp/wireup/wireup.c
```diff
@@ -551,20 +556,27 @@ void ucp_wireup_remote_connected(ucp_ep_h ep)
...
 static UCS_F_NOINLINE void
 ucp_wireup_process_pre_request(ucp_worker_h worker, ucp_ep_h ep,
                                const ucp_wireup_msg_t *msg,
                                const ucp_unpacked_address_t *remote_address)
 {
-    unsigned ep_init_flags = UCP_EP_INIT_CREATE_AM_LANE |
-                             UCP_EP_INIT_CM_WIREUP_CLIENT |
-                             ucp_ep_err_mode_init_flags(msg->err_mode);
+    ucp_err_handling_mode_t err_mode =
+            ucp_ep_config_key_flags_err_handling_mode(msg->flags);
```
brminich, line 576: i think wire compat is broken here. Previously `msg->flags` was compated by value, not it is treated as a flag

dmitrygx: yes, this is a problem…
but I think we can avoid this change, I’ll check if it is possible

PR #8555, opened 2022-09-22
UCT/UCP/API: Add RKEY_PTR flag on component

brminich:
## What
Introduce `UCT_COMPONENT_FLAG_RKEY_PTR` and use it instead of `UCT_MD_FLAG_RKEY_PTR`

## Why ?
`uct_rkey_ptr()` function is invoked for component (not memory domain),  so the flag indicating support for this API needs to be checked/present on component rather than md

### DIFF
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -549,6 +549,14 @@ ucp_tl_iface_bandwidth(ucp_context_h context, const uct_ppn_bandwidth_t *bandwid
            (bandwidth->shared / context->config.est_num_ppn);
 }
 
+static UCS_F_ALWAYS_INLINE const uct_component_attr_t*
+ucp_cmpt_attr_by_md_index(ucp_context_h context, ucp_md_index_t md_index)
+{
+    ucp_tl_md_t *tl_md = &context->tl_mds[md_index];
+
+    return &context->tl_cmpts[tl_md->cmpt_index].attr;
+}
+
 static UCS_F_ALWAYS_INLINE void
 ucp_memory_info_set_host(ucp_memory_info_t *mem_info)
 {
--- a/src/uct/api/uct.h
+++ b/src/uct/api/uct.h
@@ -290,7 +290,13 @@ enum {
      * If set, the component supports @ref uct_cm_h functionality.
      * See @ref uct_cm_open for details.
      */
-    UCT_COMPONENT_FLAG_CM = UCS_BIT(0)
+    UCT_COMPONENT_FLAG_CM       = UCS_BIT(0),
+
+    /**
+     * If set, the component supports direct access to remote memory using a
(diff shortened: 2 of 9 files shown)


hoopoepg (approved):

src/uct/api/uct.h
```diff
@@ -290,7 +290,13 @@ enum {
      * If set, the component supports @ref uct_cm_h functionality.
      * See @ref uct_cm_open for details.
      */
-    UCT_COMPONENT_FLAG_CM = UCS_BIT(0)
+    UCT_COMPONENT_FLAG_CM       = UCS_BIT(0),
+
+    /**
+     * Component supports direct access to remote memory via a pointer that is
+     * returned by @ref uct_rkey_ptr
```
yosefe, line 297: `If set, the component supports direct access to remote memory using a local pointer returned from @ref uct_rkey_ptr function`

src/ucp/core/ucp_context.h
```diff
@@ -549,6 +549,14 @@ ucp_tl_iface_bandwidth(ucp_context_h context, const uct_ppn_bandwidth_t *bandwid
            (bandwidth->shared / context->config.est_num_ppn);
 }
 
+static UCS_F_ALWAYS_INLINE uct_component_attr_t*
```
yosefe, line 552: maybe return const pointer?

src/uct/api/uct.h
```diff
@@ -2678,6 +2686,8 @@ ucs_status_t uct_rkey_unpack(uct_component_h component, const void *rkey_buffer,
  * This routine returns a local pointer to the remote memory
  * described by the rkey bundle. The MD must support
  * @ref UCT_MD_FLAG_RKEY_PTR flag.
+ * @note @ref UCT_MD_FLAG_RKEY_PTR is deprecated and replaced by the
+ * @ref UCT_COMPONENT_FLAG_RKEY_PTR component flag.
```
yosefe, line 2690: let's replace lines 2687-2688 instead of adding a note

yosefe (commented):
shall we also update tl_info?

src/uct/api/uct.h
brminich: maybe better keep it  showing that backward compatibility is not broken?

yosefe: i think the documentation should aim for new application and show the recommended/main way to use the API.
older application, that need the backward compat, are already written and don't need the API doc anymore.
if really important, we can add note that the component flag replaced the older MD flag.

brminich: ok, fixed

yosefe (dismissed):

yosefe:
@shamisp @tonycurtis can you pls review?

yosefe (approved):

tonycurtis (approved):

shamisp (approved):

PR #8561, opened 2022-09-23
AZP: change path to docker images

avildema:
swx-infra was be deleted from rdmz side 
move all need images to ucx proj


yosefe (approved):

yosefe:
@avildema can you pls port to v1.13.x?

PR #8567, opened 2022-09-26
UCP/CORE: Refactor memory registration to support export_md_map

dmitrygx:
## What

Refactor memory registration to support export_md_map in the future.

## Why ?

New field `export_md_map` will be introduced in `ucp_mem_h` to save exported memory handles.
The main goal of this PR is to prepare memory registration code to have common code for normal `md_map` and `export_md_map` with minor changes.

## How ?

1. `ucp_memh_dereg` -> `ucp_memh_deregister` + `ucp_memh_uct_deregister` (`ucp_memh_uct_deregister` will be called for `export_md_map` in the future).
2. `ucp_memh_register` -> `ucp_memh_register` +  `ucp_memh_uct_register` (`ucp_memh_uct_register` will be called for `export_md_map` in the future).
3. Update `ucp_memh_rcache_get` to be able use another rcache, not only from `context->rcache` (`ucp_memh_rcache_get` will be used for rcache of imported memory handles).
4. Introduce new function `ucp_memh_update_from_parent_memh` to use inside `ucp_memh_init_uct_reg` (`ucp_memh_update_from_parent_memh` will be called for `export_md_map` in the future).
5. Introduce new function `ucp_memh_normal_reg` to use inside `ucp_mem_map` and simplify code.
6. `ucp_mem_advise` -> `ucp_mem_advise` + `ucp_memh_uct_mem_advise` (`ucp_memh_uct_mem_advise` will be called for `export_md_map` in the future).
7. Update `ucp_mem_reg_md_map_update` to use `ucp_memh_set` to initialize all required fields of memory handle and not miss ones by mistake.

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -286,9 +286,8 @@ static void ucp_memh_uct_deregister(ucp_context_h context, ucp_mem_h memh,
     ucs_status_t status;
 
     ucs_assert(*md_map_p & UCS_BIT(md_index));
-    ucs_assertv(md_index != memh->alloc_md_index,
-                "memh %p: md_index %u alloc_md_index %u", memh, md_index,
-                memh->alloc_md_index);
+    ucs_assertv(md_index != memh->alloc_md_index, "memh %p: alloc_md_index %u",
+                memh, memh->alloc_md_index);
 
     ucs_trace("de-registering memh[%d]=%p", md_index, memh->uct[md_index]);
     ucs_assert(context->tl_mds[md_index].attr.flags & UCT_MD_FLAG_REG);
@@ -924,11 +923,10 @@ static ucs_status_t ucp_advice2uct(unsigned ucp_advice, uct_mem_advice_t *uct_ad
 static ucs_status_t ucp_memh_uct_mem_advise(ucp_context_h context,
                                             ucp_mem_h memh,
                                             ucp_md_index_t md_index,
-                                            ucp_md_index_t uct_memh_offset,
                                             void *address, size_t length,
                                             uct_mem_advice_t uct_advice)
 {
-    uct_mem_h uct_memh = memh->uct[uct_memh_offset + md_index];
+    uct_mem_h uct_memh = memh->uct[md_index];
 
     if (!(context->tl_mds[md_index].attr.flags & UCT_MD_FLAG_ADVISE) ||
         (uct_memh == NULL)) {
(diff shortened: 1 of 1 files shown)


dmitrygx:

src/ucp/core/ucp_mm.c
```diff
@@ -484,6 +517,22 @@ static ucs_status_t ucp_memh_create_from_mem(ucp_context_h context,
     return UCS_OK;
 }
 
+static UCS_F_ALWAYS_INLINE void
+ucp_memh_update_from_parent_memh(ucp_mem_h memh,
+                                 ucp_md_map_t cache_md_map,
```
Artemy-Mellanox, line 522: IMHO it's hard to explain meaning of those 3 md_map arguments. Maybe leave only one and make bitwise ops outside?

dmitrygx: added description.
`ucp_memh_update_from_parent_memh` will be used twice for `md_map` and `export_md_map`

src/ucp/core/ucp_mm.c
```diff
@@ -279,29 +279,39 @@ ucp_mem_map_params2uct_flags(const ucp_mem_map_params_t *params)
...
+                                    ucp_md_index_t md_index,
+                                    ucp_md_map_t *md_map_p)
 {
-    ucp_md_index_t md_index;
     ucs_status_t status;
 
+    ucs_assert(*md_map_p & UCS_BIT(md_index));
+    ucs_assertv(md_index != memh->alloc_md_index,
+                "memh %p: md_index %u alloc_md_index %u", memh, md_index,
+                memh->alloc_md_index);
```
brminich, line 291: no need to print both indices because they are equal

src/ucp/core/ucp_mm.c
```diff
@@ -850,52 +921,61 @@ static ucs_status_t ucp_advice2uct(unsigned ucp_advice, uct_mem_advice_t *uct_ad
...
     uct_mem_advice_t uct_advice;
-    uct_mem_h uct_memh;
 
     if (!ucs_test_all_flags(params->field_mask,
                             UCP_MEM_ADVISE_PARAM_FIELD_ADDRESS|
                             UCP_MEM_ADVISE_PARAM_FIELD_LENGTH|
                             UCP_MEM_ADVISE_PARAM_FIELD_ADVICE)) {
         return UCS_ERR_INVALID_PARAM;
     }
-
```
brminich, line 868: no need to remove here and below in this func

src/ucp/core/ucp_mm.c
```diff
@@ -850,52 +921,61 @@ static ucs_status_t ucp_advice2uct(unsigned ucp_advice, uct_mem_advice_t *uct_ad
     return UCS_ERR_INVALID_PARAM;
 }
 
+static ucs_status_t ucp_memh_uct_mem_advise(ucp_context_h context,
+                                            ucp_mem_h memh,
+                                            ucp_md_index_t md_index,
+                                            ucp_md_index_t uct_memh_offset,
```
brminich, line 927: Why do you need offset here? do you expect to be called for exported keys?

src/ucp/core/ucp_mm.c
```diff
@@ -1177,22 +1257,22 @@ void ucp_mem_rcache_cleanup(ucp_context_h context)
 
 ucs_status_t ucp_mem_reg_md_map_update(ucp_context_h context)
 {
-    ucp_mem_dummy_handle_t memh = {
-        .memh.alloc_md_index = UCP_NULL_RESOURCE,
-    };
+    ucp_mem_dummy_handle_t memh = {};
```
brminich, line 1260: why removed initialization? `UCP_NULL_RESOURCE` is not `0`

src/ucp/core/ucp_mm.c
brminich: also would not be having a separate array for exported keys be more clear than having 2*num_mds array in memh?

dmitrygx: > removed, needed in the PR which adds EXPORT_MEMH

dmitrygx: > also would not be having a separate array for exported keys be more clear than having 2*num_mds array in memh?

it is necessary to discuss it within the team

src/ucp/core/ucp_mm.c
dmitrygx: it will be needed for future changes - see the PR which introduces import/export - #8552

src/ucp/core/ucp_mm.c
dmitrygx: oh, it seems it was added mistakenly in VS code
fixed

PR #8569, opened 2022-09-27
UCT/CUDA: address "provided PTX was compiled with an unsupported toolchain" error

Akshay-Venkatesh:
## What

Report ptx error even if error is hidden and add FAQ

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -351,6 +351,15 @@ In some cases, the internal memory type cache can misdetect GPU memory as host
 memory, also leading to invalid memory access. This cache can be disabled by
 setting `UCX_MEMTYPE_CACHE=n`.
 
+#### Why am I getting the error "provided PTX was compiled with an unsupported toolchain"?
+
+The application is loading a cuda binary that was compiled for newer version of
+cuda than installed, and the failure is detected asynchronously by a Cuda API
+call from UCX. In order to fix the issue, you would need to install a newer cuda
+version or compile the cuda binaries with the appropriate -arch option to nvcc.
+Refer to https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#virtual-architecture-feature-list
+for the appropriate `-arch` option to pass to nvcc.
+
 <br/>
 
 ### Performance considerations
--- a/src/uct/cuda/base/cuda_iface.h
+++ b/src/uct/cuda/base/cuda_iface.h
@@ -36,12 +36,31 @@
     UCT_CUDAR_CALL(UCS_LOG_LEVEL_ERROR, _func, __VA_ARGS__)
 
 
+#if CUDART_VERSION >= 11010
+#define UCT_CUDA_FUNC_PTX_ERR(_result, _func, _err_str)         \
+    do {                                                        \
+        if (_result == cudaErrorUnsupportedPtxVersion) {        \
+            ucs_error("%s() failed: %s",                        \
+                      UCS_PP_MAKE_STRING(_func), _err_str);     \
+        }                                                       \
+    } while (0);
+#else
+#define UCT_CUDA_FUNC_PTX_ERR(_result, _func, _err_str)         \
+    do {                                                        \
+    } while (0);
+#endif
+
+
 #define UCT_CUDA_FUNC(_func, _log_level)                        \
     ({                                                          \
         ucs_status_t _status = UCS_OK;                          \
         do {                                                    \
             cudaError_t _result = (_func);                      \
             if (cudaSuccess != _result) {                       \
+                if (_log_level != UCS_LOG_LEVEL_ERROR) {        \
+                    UCT_CUDA_FUNC_PTX_ERR(_result,  _func,      \
+                        cudaGetErrorString(_result));           \
+                }                                               \
                 ucs_log((_log_level), "%s() failed: %s",        \
                         UCS_PP_MAKE_STRING(_func),              \
                         cudaGetErrorString(_result));           \


Akshay-Venkatesh:
cc @yosefe

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -117,11 +117,13 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_reg,
 
     log_level = (flags & UCT_MD_MEM_FLAG_HIDE_ERRORS) ? UCS_LOG_LEVEL_DEBUG :
                 UCS_LOG_LEVEL_ERROR;
-    status    = UCT_CUDA_FUNC(cudaHostRegister(address, length,
-                                               cudaHostRegisterPortable),
-                              log_level);
-    if (status != UCS_OK) {
-        return status;
+    result = cuMemHostRegister(address, length, CU_MEMHOSTREGISTER_PORTABLE);
```
brminich, line 120: is there a reason to change `cudaHostRegister` to `cuMemHostRegister`?
Do we need to use [cuMemHostUnregister()](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g63f450c8125359be87b7623b1c0b2a14) in `uct_cuda_copy_mem_dereg` then?

docs/source/faq.md
```diff
@@ -351,6 +351,13 @@ In some cases, the internal memory type cache can misdetect GPU memory as host
 memory, also leading to invalid memory access. This cache can be disabled by
 setting `UCX_MEMTYPE_CACHE=n`.
 
+#### Seeing the error "provided PTX was compiled with an unsupported toolchain"?
```
yosefe, line 354: since it's FAQ, i'd phrase it as
`How to fix the error "provided PTX was compiled with an unsupported toolchain"?`
or
`Why am I getting the error "provided PTX was compiled with an unsupported toolchain"?`

docs/source/faq.md
```diff
@@ -351,6 +351,13 @@ In some cases, the internal memory type cache can misdetect GPU memory as host
...
+
+The name of the class of NVIDIA 'virtual' GPU architecture for which CUDA input
+files were compiled does not match with underlying GPU architecture. Refer to
+https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#virtual-architecture-feature-list
+for the appropriate `-arch` option to pass to nvcc.
```
yosefe, line 359: this is very unclear...
i'd write something like "The application is loading a cuda binary that was compiled for newer version of cuda than installed, and the failure is detected asynchronously by a Cuda API call from UCX. In order to fix the issue, need to install a newer cuda version or compile the cuda binaries with the appropriate -arch option to nvcc"

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -117,11 +117,13 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_reg,
...
+    if (result != CUDA_SUCCESS) {
+        log_level = (result == CUDA_ERROR_UNSUPPORTED_PTX_VERSION) ?
+                    UCS_LOG_LEVEL_ERROR : log_level;
+        cuGetErrorString(result, &cu_err_str);
+        ucs_log(log_level, "cuMemHostRegister failed: %s", cu_err_str);
```
yosefe, line 125: pls print address and length

src/uct/cuda/cuda_copy/cuda_copy_md.c
Akshay-Venkatesh: @brminich No particular reason other than trying to rely on only driver API. Will replace unregister call with driver API as well.

Akshay-Venkatesh:
@yosefe @brminich I incorporated suggested changes but also removed a few more instances of runtime API calls that I found in cuda transports. I hope it's ok to include these change this PR. If not, then I can create a separate PR for it. Let me know

Akshay-Venkatesh:
@yosefe Added https://github.com/openucx/ucx/pull/8590. Once that is merged I'll remove 

> @Akshay-Venkatesh i'd prefer runtime/driver in a separate PR besides, seems tests are failing..

@yosefe I added https://github.com/openucx/ucx/pull/8590 to address runtime API replacement. Once that is merged, I can make this PR more minimal.

yosefe:
@Akshay-Venkatesh what should this PR depend on removing runtime API?

bureddy:
@Akshay-Venkatesh are you sure that "provided PTX was compiled..." is because of cuda runtime in UCX. I think is is because of HCOLL cuda kernels compilation. should not see this error if there is no kernels.

yosefe:
> > @Akshay-Venkatesh what should this PR depend on removing runtime API?
> 
> Only because we want to handle any PTX unsupported error by catching it in driver [macro](https://github.com/openucx/ucx/pull/8569/files#diff-9bd95d0094ca852dd384041d01aeb696a1b40fb227ff5745c61f9cad2ac908baR64). Otherwise we would catch it RT macro and we would remove that at the end of these two PRs.

i'd prefer to avoid such dependency and merge the PTX fix first since it's fixing a user-facing problem. moving to driver API is a broader change.

Akshay-Venkatesh:
> @Akshay-Venkatesh what should this PR depend on removing runtime API?

Only because we want to handle any PTX unsupported error by catching it in driver [macro](https://github.com/openucx/ucx/pull/8569/files#diff-9bd95d0094ca852dd384041d01aeb696a1b40fb227ff5745c61f9cad2ac908baR64). Otherwise we would catch it RT macro and we would remove that at the end of these two PRs.

> @Akshay-Venkatesh are you sure that "provided PTX was compiled..." is because of cuda runtime in UCX. I think is is because of HCOLL cuda kernels compilation. should not see this error if there is no kernels.

@bureddy Yes. It does come from kernels (from user kernel if not hcoll kernels), but it happened that user didn't catch using cudaGetLastError() and UCX was reporting the error. The current change is to report this even if errors are hidden by UCX and add description in FAQs.

src/uct/cuda/base/cuda_iface.h
```diff
@@ -36,12 +36,31 @@
     UCT_CUDAR_CALL(UCS_LOG_LEVEL_ERROR, _func, __VA_ARGS__)
 
 
+#if CUDART_VERSION >= 11010
+#define UCT_CUDA_FUNC_PTX_ERR(_result, _func, _err_str)         \
+    do {                                                        \
+        if (_result == cudaErrorUnsupportedPtxVersion) {        \
+            ucs_log(UCS_LOG_LEVEL_ERROR, "%s() failed: %s",     \
```
yosefe, line 43: ucs_error

yosefe (dismissed):

yosefe:
@tonycurtis @shamisp can you pls review the documentation?

shamisp (dismissed):

docs/source/faq.md
```diff
@@ -351,6 +351,15 @@ In some cases, the internal memory type cache can misdetect GPU memory as host
 memory, also leading to invalid memory access. This cache can be disabled by
 setting `UCX_MEMTYPE_CACHE=n`.
 
+#### Why am I getting the error "provided PTX was compiled with an unsupported toolchain"?
+
+The application is loading a cuda binary that was compiled for newer version of
+cuda than installed, and the failure is detected asynchronously by a Cuda API
+call from UCX. In order to fix the issue, need to install a newer cuda version
```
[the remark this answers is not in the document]
yosefe: @Akshay-Venkatesh FYI

yosefe (dismissed):

brminich (approved):

yosefe:
@Akshay-Venkatesh can you pls squash?

yosefe (approved):

PR #8570, opened 2022-09-27
UCT/BASE: Use uct_md_query_v2 in uct_mem_alloc and check EXPORT

dmitrygx:
## What

Use `uct_md_query_v2` in `uct_mem_alloc` and check `UCT_MD_FLAG_EXPORTED_MKEY` capability if `UCT_MD_MEM_FLAG_EXPORT` was asked along with memory allocation.

## Why ?

To theoretically support MDs with memory allocation and exporting memory.

## How ?

1. Use `uct_md_query_v2` instead of `uct_md_query` in `uct_mem_alloc`.
2. Check `UCT_MD_FLAG_EXPORTED_MKEY` MD attribute if `UCT_MD_MEM_FLAG_EXPORT` flag has been specified.

### DIFF
--- a/src/uct/base/uct_mem.c
+++ b/src/uct/base/uct_mem.c
@@ -70,7 +70,7 @@ ucs_status_t uct_mem_alloc(size_t length, const uct_alloc_method_t *methods,
     const char *alloc_name;
     const uct_alloc_method_t *method;
     ucs_memory_type_t mem_type;
-    uct_md_attr_t md_attr;
+    uct_md_attr_v2_t md_attr;
     ucs_status_t status;
     unsigned flags;
     size_t alloc_length;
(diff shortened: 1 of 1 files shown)


dmitrygx:

src/uct/base/uct_mem.c
```diff
@@ -120,26 +120,33 @@ ucs_status_t uct_mem_alloc(size_t length, const uct_alloc_method_t *methods,
...
                 /* Check if MD supports allocation with fixed address
                  * if it's requested */
                 if ((flags & UCT_MD_MEM_FLAG_FIXED) &&
-                    !(md_attr.cap.flags & UCT_MD_FLAG_FIXED)) {
+                    !(md_attr.flags & UCT_MD_FLAG_FIXED)) {
+                    continue;
+                }
+
+                /* Check if MD supports allocation of memory is exported if
+                 * it's requested */
```
brminich, line 142: imo, this comment is less clear than the code below.  I'd remove it or did you mean
`Check if MD supports allocation of memory as exported if it's requested */`

src/uct/base/uct_mem.c
```diff
@@ -120,26 +120,33 @@ ucs_status_t uct_mem_alloc(size_t length, const uct_alloc_method_t *methods,
...
+                if ((flags & UCT_MD_MEM_FLAG_EXPORT) &&
+                    !(md_attr.flags & UCT_MD_FLAG_EXPORTED_MKEY)) {
```
brminich, line 144: do we really need to support this combination? What is the use-case?

src/uct/base/uct_mem.c
dmitrygx: yes, I meant using `as` instead of `is`
fixed

src/uct/base/uct_mem.c
dmitrygx: yes, we should support it for theoretical transports which could support ALLOCATE and EXPORT simultaneously.
now we don't have such transports, but users could pass `EXPORT` flag anyway and we should handle it, i.e. not allocate memory using MD and move to another method of memory allocation.

dmitrygx:
@brminich is it ok?

brminich (approved):

dmitrygx:
@yosefe could you review pls?

PR #8571, opened 2022-09-28
UCP/MM: lock free when deregistration memory region

changchengx:
drop pgt_lock while de-registrating memory region which maybe called frequently and raise serious lock contentions.

## What
Do not hold pgt_lock when dereg MR

## Why ?
In some cases, dereg MR maybe called frequently and hold pgt_lock for long time, then raise serious pgt_lock contention.

## How ?
release pgt_lock during degreg MR


yosefe:
@Artemy-Mellanox do you think such a change can cause trouble with ibv_dontfork_range/ibv_dofork order in libibverbs?

changchengx:
@Artemy-Mellanox @yosefe Any comment before fixing the coverity problem?

changchengx:
@Artemy-Mellanox what's about your opinion?

src/ucs/memory/rcache.c
```diff
@@ -364,10 +364,12 @@ void ucs_mem_region_destroy_internal(ucs_rcache_t *rcache,
 
     if (region->flags & UCS_RCACHE_REGION_FLAG_REGISTERED) {
         UCS_STATS_UPDATE_COUNTER(rcache->stats, UCS_RCACHE_DEREGS, 1);
+        pthread_rwlock_unlock(&rcache->pgt_lock);
```
yosefe, line 367: need to bass a flag to ucs_mem_region_destroy_internal function to indicate whether the lock should be dropped or not (according to whether the lock is being held by the called).
For example when called from ucs_rcache_region_put_internal, it can depend on "flags" parameter

yosefe: the flag for ucs_mem_region_destroy_internal should indicate whether to drop the lock or not. it should not include UCS_RCACHE_REGION_PUT_FLAG_ADD_TO_GC

changchengx: @yosefe 
For the flag ```UCS_RCACHE_REGION_PUT_FLAG_ADD_TO_GC```, it can use new enumerate or ```1``` or ```0``` directly.
What's about dropping the lock during GC process? The reported trace shows much time is spent on GC cleanup function.

changchengx: @yosefe As we've aligned offline, I'll update the squashed patch & do force push since the change is within 50 lines.

yosefe (approved):

yosefe:
@Artemy-Mellanox can you pls review as well?

Artemy-Mellanox (approved):

Issue #8523, opened 2022-09-11
shm_ib/test_ucp_am_nbx_send_copy_header.all_protos/3 <shm,ib/reply/proto> fails on swx-rdmz-ucx-gpu-01


brminich:
### Describe the bug
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=50521&view=logs&j=f97d6e39-20b4-5224-cb57-ac6707fa3b18&t=65630ad1-e2d8-5140-3757-25a4f549d6a1

```
2022-09-10T19:47:14.5269298Z [ RUN      ] shm_ib/test_ucp_am_nbx_send_copy_header.all_protos/3 <shm,ib/reply/proto>
2022-09-10T19:47:14.8133395Z [     INFO ] header length 8 data length 8
2022-09-10T19:47:14.8356245Z [     INFO ] header length 8 data length 8 RNDV
2022-09-10T19:47:14.8373897Z [     INFO ] header length 32 data length 32
2022-09-10T19:47:14.8374559Z [     INFO ] header length 32 data length 32 RNDV
2022-09-10T19:47:14.8383474Z [     INFO ] header length 64 data length 64
2022-09-10T19:47:14.8390461Z [     INFO ] header length 64 data length 64 RNDV
2022-09-10T19:47:14.8418378Z [     INFO ] header length 8 data length 4124
2022-09-10T19:47:14.8422923Z /__w/1/s/contrib/../test/gtest/common/mem_buffer.cc:245: Failure
2022-09-10T19:47:14.8439848Z Pattern check failed at 0x4f44aa0 offset 0: Expected: 0x1111111111111111 Actual: 0x8c93540
2022-09-10T19:47:14.8445101Z [swx-rdmz-ucx-gpu-01:11288:0:11288]  ucp_worker.c:2806 Assertion `worker->inprogress++ == 0' failed
2022-09-10T19:47:14.8446281Z ==== backtrace (tid:  11288) ====
2022-09-10T19:47:14.8447713Z  0  /__w/1/s/build-test/src/ucs/.libs/libucs.so.0(ucs_handle_error+0x12c) [0x7f316bfab95c]
2022-09-10T19:47:14.8449418Z  1  /__w/1/s/build-test/src/ucs/.libs/libucs.so.0(ucs_fatal_error_message+0x58) [0x7f316bfa8a18]
[...]

PR #8573, opened 2022-09-28
GTEST: AM copy header - Enforce receiving all the messages in a test set before moving to the next set

ofirfarjun7:
## What
Fix issue reported in [bug](https://github.com/openucx/ucx/issues/8523)

## Why ?
Receiving messages out of order caused the unit test pattern check to fail.
All AM unit tests use the same class field ("m_rx_buf" of test_ucp_am_nbx ) to store the pointer to the buffer used when calling "ucp_am_recv_data_nbx".
Due to false progressing scheme in the test, message was received out of order and therefore the pattern check was called with the wrong buffer.

Example:
//Msg 1 AM CB
m_rx_buf = allocate new buffer
call ucp_am_recv_data_nbx with

//Before msg 1 finished - Msg 2 AM CB
m_rx_buf = allocate new buffer
call ucp_am_recv_data_nbx with m_rx_buf 

//msg 2 received 
pattern check called with m_rx_buf 
release m_rx_buf
//msg 1 received 
pattern check called with m_rx_buf 
error...

## How ?
Fix the unit test progress scheme to enforce receiving all the messages belong to a test set before moving to the next test set.
This is done by advancing the received message counter only after pattern check completed.

### DIFF
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -25,9 +25,9 @@ extern "C" {
 #define NUM_MESSAGES 17
 
 #define UCP_REALLOC_ID 1000
-#define UCP_SEND_ID 0
-#define UCP_REPLY_ID 1
-#define UCP_RELEASE 1
+#define UCP_SEND_ID    0
+#define UCP_REPLY_ID   1
+#define UCP_RELEASE    1
 
 class test_ucp_am_base : public ucp_test {
 public:
@@ -44,7 +44,8 @@ public:
         add_variant_with_value(variants, UCP_FEATURE_AM, 1, "proto");
     }
 
-    virtual void init() {
+    virtual void init()
+    {
         modify_config("MAX_EAGER_LANES", "2");
 
         ucp_test::init();
@@ -211,7 +212,7 @@ void test_ucp_am::do_send_process_data_test(int test_release, uint16_t am_id,
     }
 
     if (test_release) {
-        for(int i = 0; i < recv_ams; i++) {
+        for (int i = 0; i < recv_ams; i++) {
             if (for_release[i] != NULL) {
                 ucp_am_data_release(receiver().worker(), for_release[i]);
             }
@@ -325,11 +326,11 @@ public:
 
     test_ucp_am_nbx()
     {
-        m_dt          = ucp_dt_make_contig(1);
-        m_am_received = false;
-        m_rx_dt       = ucp_dt_make_contig(1);
-        m_rx_buf      = NULL;
-        m_rx_memh     = NULL;
+        reset_counters();
+        m_dt      = ucp_dt_make_contig(1);
+        m_rx_dt   = ucp_dt_make_contig(1);
+        m_rx_buf  = NULL;
+        m_rx_memh = NULL;
     }
 
     void test_datatypes(std::function<void()> test_f);
@@ -343,6 +344,17 @@ protected:
     virtual ucs_memory_type_t rx_memtype() const
     {
         return UCS_MEMORY_TYPE_HOST;
+    }
+
+    void reset_counters()
+    {
+        m_send_counter = 0;
+        m_recv_counter = 0;
+    }
+
+    void wait_receives()
+    {
+        wait_for_value(&m_recv_counter, m_send_counter);
     }
 
     size_t max_am_hdr()
@@ -368,9 +380,9 @@ protected:
     ucp_datatype_t make_dt(int dt)
     {
         if (dt == UCP_DATATYPE_CONTIG) {
-           return ucp_dt_make_contig(1);
+            return ucp_dt_make_contig(1);
         } else if (dt == UCP_DATATYPE_IOV) {
-           return ucp_dt_make_iov();
+            return ucp_dt_make_iov();
         } else {
             ucs_assertv(UCP_DATATYPE_GENERIC == dt, "dt=%d", dt);
             ucp_datatype_t ucp_dt;
@@ -387,8 +399,9 @@ protected:
         }
     }
 
-    void set_am_data_handler(entity &e, uint16_t am_id, ucp_am_recv_callback_t cb,
-                             void *arg, unsigned flags = 0)
+    void set_am_data_handler(entity &e, uint16_t am_id,
+                             ucp_am_recv_callback_t cb, void *arg,
+                             unsigned flags = 0)
     {
         ucp_am_handler_param_t param;
 
@@ -415,13 +428,23 @@ protected:
     }
 
     ucs_status_ptr_t
+    update_counter_and_send_am(const void *header, size_t header_length,
+                               const void *buffer, size_t count,
+                               const ucp_request_param_t *param)
+    {
+        m_send_counter++;
+        return ucp_am_send_nbx(sender().ep(), TEST_AM_NBX_ID, header,
+                               header_length, buffer, count, param);
+    }
+
+    ucs_status_ptr_t
     send_am(const ucp::data_type_desc_t &dt_desc, unsigned flags = 0,
             const void *hdr = NULL, unsigned hdr_length = 0,
             const ucp_mem_h memh = NULL, uint32_t op_attr_mask = 0)
     {
         ucp_request_param_t param;
-        param.op_attr_mask      = op_attr_mask | UCP_OP_ATTR_FIELD_DATATYPE;
-        param.datatype          = dt_desc.dt();
+        param.op_attr_mask = op_attr_mask | UCP_OP_ATTR_FIELD_DATATYPE;
+        param.datatype     = dt_desc.dt();
 
         if (flags != 0) {
             param.op_attr_mask |= UCP_OP_ATTR_FIELD_FLAGS;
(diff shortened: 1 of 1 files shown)


ofirfarjun7:
Technically this fix is not enforcing message ordering but it's enforce order between different test sets (each set sending message with different header and data lengths). 
I run the test 6000 times on the same azure cluster that showed the problem and I didn't saw it after the fix.

Now when the sent and received counters are part of the test_ucp_am_nbx class and CB's maybe we can remove the m_am_received flags and replace it with some logic involve the counters. I left this to be decide by the reviewers.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -433,6 +433,7 @@ class test_ucp_am_nbx : public test_ucp_am_base {
             param.memh          = memh;
         }
 
+        m_am_sent_counter++;
```
gleon99, line 436: Move to after `send`

gleon99:
> Technically this fix is not enforcing message ordering but it's enforce order between messages of different test sets (in each set messages with specific header and data lengths are being sent). I run the test repeatedly for 6000 times on the same azure cluster that showed the problem and I didn't saw it after the fix. Before the fix it appeared quickly (after <100 runs).
> 
> Now when the sent and received counters are part of the test_ucp_am_nbx class and AM CBs advancing them, maybe we can remove the m_am_received flag and replace it with some logic involving the counters. I left this to be decide by the reviewers.

IMO we don't need both the flag and the counter, can get rid of the flag. @yosefe , WDYT?

test/gtest/ucp/test_ucp_am.cc
ofirfarjun7: I have fear of concurrencies issues for example AM CB is called before sent_counter is advanced. it can cause the received counter to be larger than sent counter, and I think the invariant should be that received counter is always <= sent counter.
Maybe it's not causing issues now, but it can cause issues in the future, for example if we choose to replace the m_am_received flag we will probably compare the values of sent and receive counters in the CB.

yosefe: better keep it before sending, but rename to m_am_send_counter

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -641,7 +644,9 @@ class test_ucp_am_nbx : public test_ucp_am_base {
         return 0;
     }
 
-    static const uint16_t           TEST_AM_NBX_ID = 0;
+    static const uint16_t           TEST_AM_NBX_ID        = 0;
+    volatile size_t                 m_am_sent_counter     = 0;
+    volatile size_t                 m_am_received_counter = 0;
```
yosefe, line 649: remove m_am_received and use m_am_received_counter instead in the base class as well

yosefe, line 649: m_send_counter
m_recv_counter

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -326,7 +326,7 @@ class test_ucp_am_nbx : public test_ucp_am_base {
     test_ucp_am_nbx()
     {
         m_dt          = ucp_dt_make_contig(1);
-        m_am_received = false;
+        init_send_recv_counters();
```
yosefe, line 329: 1. reset_counters()
2. move to 1st line in the function

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -345,6 +345,11 @@ class test_ucp_am_nbx : public test_ucp_am_base {
         return UCS_MEMORY_TYPE_HOST;
     }
 
+    void init_send_recv_counters() {
```
yosefe, line 348: { on next line

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -570,15 +576,15 @@ class test_ucp_am_nbx : public test_ucp_am_base {
                                          const ucp_am_recv_param_t *rx_param)
     {
 
-        EXPECT_FALSE(m_am_received);
+        EXPECT_LT(m_recv_counter, m_send_counter);
```
yosefe, line 579: maybe EXPECT_EQ(0, m_recv_counter) ?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -588,7 +594,7 @@ class test_ucp_am_nbx : public test_ucp_am_base {
 
     virtual void am_recv_check_data(size_t length)
     {
-        ASSERT_FALSE(m_am_received);
+        EXPECT_LT(m_recv_counter, m_send_counter);
```
yosefe, line 597: maybe ASSERT_EQ(0, m_recv_counter) ?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -987,13 +993,13 @@ class test_ucp_am_nbx_send_copy_header : public test_ucp_am_nbx {
 
         ucs::fill_random(m_hdr_copy);
         while (progress());
-        wait_for_value(&cb_counter, req_counter);
+        wait_for_value(&m_recv_counter, m_send_counter);
```
yosefe, line 996: maybe add common wrapper function wait_receives() for "wait_for_value(&m_recv_counter, m_send_counter);"

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1232,15 +1225,15 @@ class test_ucp_am_nbx_closed_ep : public test_ucp_am_nbx_reply {
...
             request_wait(sreq);
-            wait_for_flag(&m_am_received);
+            wait_for_value(&m_recv_counter, m_send_counter);
+            EXPECT_EQ(m_recv_counter, m_send_counter);
         } else {
             // Send request may complete with error
             // (rndv should complete with EP_TIMEOUT)
             scoped_log_handler wrap_err(wrap_errors_logger);
             request_wait(sreq);
+            EXPECT_LT(m_recv_counter, m_send_counter);
```
yosefe, line 1235: EXPECT_EQ(0, m_recv_counter)

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1333,14 +1326,14 @@ class test_ucp_am_nbx_eager_data_release : public test_ucp_am_nbx {
     am_data_handler(const void *header, size_t header_length, void *data,
                     size_t length, const ucp_am_recv_param_t *rx_param)
     {
-        EXPECT_FALSE(m_am_received);
+        EXPECT_LT(m_recv_counter, m_send_counter);
```
yosefe, line 1329: EXPECT_EQ(0, m_recv_counter)

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1333,14 +1326,14 @@ class test_ucp_am_nbx_eager_data_release : public test_ucp_am_nbx {
...
         EXPECT_TRUE(rx_param->recv_attr & UCP_AM_RECV_ATTR_FLAG_DATA);
 
-        m_am_received = true;
         m_data_ptr    = data;
 
         check_header(header, header_length);
         mem_buffer::pattern_check(data, length, SEED);
+        m_recv_counter++;
```
yosefe, line 1336: remove extra spaces in line 1332

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1572,8 +1565,8 @@ class test_ucp_am_nbx_rndv : public test_ucp_am_nbx_prereg {
     {
         test_ucp_am_nbx_rndv *self = reinterpret_cast<test_ucp_am_nbx_rndv*>(arg);
 
-        EXPECT_FALSE(self->m_am_received);
-        self->m_am_received = true;
+        EXPECT_LT(self->m_recv_counter, self->m_send_counter);
```
yosefe, line 1568: EXPECT_EQ(0, m_recv_counter)

test/gtest/ucp/test_ucp_am.cc
ofirfarjun7: That can cause an error because this CB handles a batch of messages sent by the AM header test, the m_recv_counter in this case can be larger than 0.
I think that in the tests that send only one message we can treat 
"EXPECT_LT(m_recv_counter, m_send_counter)" and "EXPECT_EQ(0, m_recv_counter)" the same because m_send_counter is equal to 1.

This explanation is true also for the other common CBs that uses the counters. I can change the EXPECT_LT to EXPECT_EQ only in tests that sending only one message and performing this check in non common code, but again in my perspective if you send only one message both checks can be considered equal.

test/gtest/ucp/test_ucp_am.cc
ofirfarjun7: See reply above

test/gtest/ucp/test_ucp_am.cc
ofirfarjun7: I can change it because it's non common code. Will be more clear, although the conditions has the same logic

test/gtest/ucp/test_ucp_am.cc
ofirfarjun7: I can change it because it's non common code. Will be more clear, although the conditions has the same logic

test/gtest/ucp/test_ucp_am.cc
ofirfarjun7: I can change it because it's non common code. Will be more clear, although the conditions has the same logic

test/gtest/ucp/test_ucp_am.cc
yosefe: how did it work before this PR then? the test header test that sends multiple messages already existed

ofirfarjun7: Different AM_CB's that uses common CB. Only one checked the flags.
Can use it again and check for two different conditions.

yosefe (dismissed):

yosefe (approved):

PR #8576, opened 2022-09-29
UCT/TCP: Consider PCIe BW for TCP iface BW calculation

shasson5:
## What
Consider PCIe BW for TCP iface BW calculation.

## Why ?
OFED query does not consider PCIe BW value, so we add it to get the full BW calculation for TCP iface.

### DIFF
--- a/src/ucs/sys/sys.c
+++ b/src/ucs/sys/sys.c
@@ -1615,3 +1615,24 @@ int ucs_sys_is_dynamic_lib(void)
 #endif
 }
 
+ucs_status_t ucs_sys_read_sysfs_file(const char *dev_name,
+                                     const char *sysfs_path,
+                                     const char *file_name, char *output_buffer,
+                                     size_t max, ucs_log_level_t err_level)
+{
+    ssize_t nread;
+
+    if (sysfs_path == NULL) {
+        return UCS_ERR_NO_ELEM;
+    }
+
+    nread = ucs_read_file_str(output_buffer, max, 1, "%s/%s", sysfs_path,
+                              file_name);
+    if (nread < 0) {
+        ucs_log(err_level, "%s: could not read from '%s/%s'", dev_name,
+                sysfs_path, file_name);
+        return UCS_ERR_NO_ELEM;
+    }
+
+    return UCS_OK;
+}
--- a/test/gtest/ucp/test_ucp_wireup.cc
+++ b/test/gtest/ucp/test_ucp_wireup.cc
@@ -83,6 +83,18 @@ protected:
     bool ep_iface_has_caps(const entity& e, const std::string& tl,
                            uint64_t caps);
 
+    bool
+    has_resource(const ucp_test_base::entity *e, const std::string &tl_name)
+    {
+        for (int i = 0; i < e->ucph()->num_tls; ++i) {
+            if (tl_name == e->ucph()->tl_rscs[i].tl_rsc.tl_name) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
 protected:
     vec_type                               m_send_data;
     vec_type                               m_recv_data;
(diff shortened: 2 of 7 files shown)


test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1324,6 +1324,31 @@ UCS_TEST_P(select_transport_rndv, select_dc, "NUM_EPS=300")
...
+                    (config->key.rma_bw_lanes[i] != UCP_NULL_LANE);
+         ++i) {
+        /* Verify RC is selected for rma_bw lanes */
+        ucp_lane_index_t lane = config->key.rma_bw_lanes[i];
+        ASSERT_STREQ("rc_mlx5",
+                     ucp_ep_get_tl_rsc(sender().ep(), lane)->tl_name);
+    }
+}
+
+UCP_INSTANTIATE_TEST_CASE_TLS(select_transport_rma_bw, tcp_ib, "tcp,ib")
```
yosefe, line 1350: does it work if there is no IB device on the machine?

src/ucs/sys/sys.c
```diff
@@ -1615,3 +1616,155 @@ int ucs_sys_is_dynamic_lib(void)
...
+    if (nread < 0) {
+        ucs_log(err_level, "%s: could not read from '%s/%s'", dev_name,
+                sysfs_path, file_name);
+        return UCS_ERR_NO_ELEM;
+    }
+
+    return UCS_OK;
+}
+
+double ucs_sys_get_pci_bw(const char *dev_name, const char *sysfs_path)
```
yosefe, line 1706: IMO this function (and related helper code) should go to sys/topo.h,c
but ucs_sys_read_sysfs_file should stay in sys/sys.h,c

test/gtest/ucp/test_ucp_wireup.cc
shasson5: how can I check that? is there any way to simulate it? or do I need a machine with no IB devices?

src/ucs/sys/sys.c
```diff
@@ -29,6 +29,7 @@
 #include <sys/resource.h>
 #include <net/if.h>
 #include <dirent.h>
+#include <float.h>
```
yosefe, line 32: still needed?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -83,6 +83,19 @@ class test_ucp_wireup : public ucp_test {
     bool ep_iface_has_caps(const entity& e, const std::string& tl,
                            uint64_t caps);
 
+    bool has_resource(const ucp_test_base::entity *e, const char *tl_name)
```
yosefe, line 86: maybe tl_name argument could be const std::string& ?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1324,6 +1324,35 @@ UCS_TEST_P(select_transport_rndv, select_dc, "NUM_EPS=300")
...
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_RMA, TEST_RMA, "rma");
+    }
+};
+
+UCS_TEST_P(select_transport_rma_bw, select_rc, "NUM_EPS=56", "NUM_PPN=28")
+{
+    if (!has_resource(&sender(), "rc_mlx5")) {
```
yosefe, line 1337: use UCS_TEST_SKIP_COND_P macro

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1324,6 +1324,35 @@ UCS_TEST_P(select_transport_rndv, select_dc, "NUM_EPS=300")
...
+        UCS_TEST_SKIP_R("rc_mlx5 transport not found");
+    }
+
+    sender().connect(&receiver(), get_ep_params());
+    const ucp_ep_config_t *config = ucp_ep_config(sender().ep());
```
yosefe, line 1342: space line before instead of after

yosefe (approved):

brminich (dismissed):

PR #8582, opened 2022-09-30
UCP/AMO/SW: fixed incorrect assert

hoopoepg:
- fixed incorrect assert

### DIFF
--- a/src/ucp/rma/amo_sw.c
+++ b/src/ucp/rma/amo_sw.c
@@ -65,8 +65,6 @@ ucp_amo_sw_progress(uct_pending_req_t *self, uct_pack_callback_t pack_cb,
     ucp_request_t *req = ucs_container_of(self, ucp_request_t, send.uct);
     ucs_status_t status;
 
-    ucs_assert(req->flags & UCP_REQUEST_FLAG_PROTO_AMO_PACKED);
-
     req->send.lane = ucp_ep_get_am_lane(req->send.ep);
     if (fetch) {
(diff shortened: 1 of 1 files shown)


evgeny-leksikov (approved):

src/ucp/rma/amo_sw.c
```diff
@@ -65,7 +65,10 @@ ucp_amo_sw_progress(uct_pending_req_t *self, uct_pack_callback_t pack_cb,
     ucp_request_t *req = ucs_container_of(self, ucp_request_t, send.uct);
     ucs_status_t status;
 
-    ucs_assert(req->flags & UCP_REQUEST_FLAG_PROTO_AMO_PACKED);
+    if (req->send.ep->worker->context->config.ext.proto_enable) {
+        /* assert is valid only when protocols are enabled */
+        ucs_assert(req->flags & UCP_REQUEST_FLAG_PROTO_AMO_PACKED);
```
shamisp, line 70: maybe just move the branch/check inside of the assert statement ?

hoopoepg: > maybe just move the branch/check inside of the assert statement ?

not needed - compiler will exclude whole `if` statement in case of empty action block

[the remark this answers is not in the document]
hoopoepg: > maybe move the assertion to ucp_proto_amo_sw_progress?

ok, moved

yosefe (approved):

PR #8587, opened 2022-10-02
UCT/API/IB: Add support for dmabuf registration

yosefe:
## Why
As part of dmabuf support, add dambuf registration (import) support to UCT API and implement it for IB transport

## How
- API: Add `UCT_MD_FLAG_REG_DMABUF` md capability and `dmabuf_fd`,`dmabuf_offset` params to memory registration.
- IB: Check dmabuf support on startup, and pass dmabuf fd and offset to internal registration functions

This PR complements #8581 
cc @Akshay-Venkatesh

### DIFF
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -195,7 +195,9 @@ typedef struct {
  * @brief MD memory registration operation flags.
  */
 typedef enum {
-    UCT_MD_MEM_REG_FIELD_FLAGS = UCS_BIT(0)
+    UCT_MD_MEM_REG_FIELD_FLAGS         = UCS_BIT(0),
+    UCT_MD_MEM_REG_FIELD_DMABUF_FD     = UCS_BIT(1),
+    UCT_MD_MEM_REG_FIELD_DMABUF_OFFSET = UCS_BIT(2)
 } uct_md_mem_reg_field_mask_t;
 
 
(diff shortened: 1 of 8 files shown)


yosefe:

Artemy-Mellanox (approved):

yosefe:
@shamisp @tonycurtis can you pls review

src/uct/api/v2/uct_v2.h
```diff
@@ -378,6 +380,41 @@ typedef struct uct_md_mem_reg_params {
      * Operation specific flags, using bits from @ref uct_md_mem_flags_t.
      */
     uint64_t                     flags;
+
+    /**
+     * dmabuf file descriptor of the memory region to register.
+     *
+     * If is set (along with its corresponding bit in the field_mask -
+     * @ref UCT_MD_MEM_REG_FIELD_DMABUF_FD), the memory region will be
+     * registered by using dmabuf mechanism. In this case, the address argument
```
shamisp, line 389: you can remove "by"

src/uct/api/v2/uct_v2.h
```diff
@@ -378,6 +380,41 @@ typedef struct uct_md_mem_reg_params {
...
+     * provided to the registration function is the remote access virtual
```
shamisp, line 390: remote access virtual - ? I think it is still local address, right ?

shamisp: This sentence is a bit difficult to parse

src/uct/api/v2/uct_v2.h
```diff
@@ -378,6 +380,41 @@ typedef struct uct_md_mem_reg_params {
...
+     * Can be used only if the memory domain supports dmabuf registration by
+     * returning @ref UCT_MD_FLAG_REG_DMABUF flag from @ref uct_md_query_v2.
+     *
+     * When registering memory by dmabuf mechanism, any memory type is supported
+     * (assuming a valid dmabuf file descriptor could be created).
+     * Therefore, @ref uct_md_attr_v2_t.reg_mem_types field is not relevant for
+     * such registrations.
+     *
+     * If not set, it's assumed to be @ref UCT_DMABUF_FD_INVALID, and the memory
+     * region will be registered directly.
```
shamisp, line 402: registered directly - what does directly means here ?

shamisp (changes requested):

yosefe:
@tonycurtis @shamisp  fixed

src/uct/api/v2/uct_v2.h
```diff
@@ -378,6 +380,42 @@ typedef struct uct_md_mem_reg_params {
...
+
+    /**
+     * dmabuf file descriptor of the memory region to register.
+     *
+     * If is set (along with its corresponding bit in the field_mask -
+     * @ref UCT_MD_MEM_REG_FIELD_DMABUF_FD), the memory region will be
+     * registered using dmabuf mechanism. In this case, the address argument
+     * provided to the registration function is the remote access virtual
+     * address of the memory region (for RMA or AMO operation from a peer
+     * endpoint), and the local address to register is deduced from
```
shamisp, line 392: @yosefe  maybe add ref to the function itself. 

I'm still puzzled why local function registration requires target/remote VA ? I was try to trace it down in the code and it seems like it is not used for dmabuf flow.

shamisp: As I read it here https://github.com/linux-rdma/rdma-core/blob/6c8b24fb64770449d76193f8dead3e13a356ebe6/libibverbs/man/ibv_reg_mr.3#L89

I don't see reference for remote addr

shamisp: Oh ... "remote access virtual" - I think you don't mean "remote" here. It is really local VA that is shared with remote peers within the range of this DMA buffer.

shamisp: "In this case, the address argument provided to the registration function is the remote access virtual address of the memory region (for RMA or AMO operation from a peer endpoint), and the local address to register is deduced from"
->
"In this case, the address argument provided to the registration function is the local virtual address of the memory region that will be used for an access (RMA or AMO) by peer endpoints"

shamisp: So, you just nuked the entire paragraph ?
```
In this case, the address argument
     * provided to the registration function is the remote access virtual
     * address of the memory region (for RMA or AMO operation from a peer
     * endpoint), and the local address to register is deduced from
     * @ref uct_md_mem_reg_params_t.dmabuf_offset field.
```

yosefe: @shamisp yes, it was replaced by comment in dmabuf_offset

shamisp (approved):

PR #8589, opened 2022-10-03
A separate gdrcopy binary Deb package

tzafrir-mellanox:
Move gdrcopy-dependent modules independent of the CUDA ones: it is  possible to have just CUDA without gdrcopy.

First commit slightly edits the descriptions of the other two binary deb packages. The second commit adds the new binary deb package ucx-gdrcopy and moves files to it from the original ucx-cuda package.


tzafrir-mellanox:
Note: this one is still untested

yosefe (approved):

PR #8590, opened 2022-10-03
UCT/CUDA: move rt api -> drv api

Akshay-Venkatesh:
## What/Why?
Replace remaining references to CUDA runtime API to reduce dependencies.

### DIFF
--- a/src/uct/cuda/base/cuda_md.c
+++ b/src/uct/cuda/base/cuda_md.c
@@ -71,26 +71,43 @@ uct_cuda_base_query_md_resources(uct_component_t *component,
     CUdevice cuda_device;
     ucs_status_t status;
     char device_name[10];
-    int num_gpus;
+    int i, num_gpus;
 
-    status = UCT_CUDA_CALL(UCS_LOG_LEVEL_DIAG, cudaGetDeviceCount, &num_gpus);
+    status = UCT_CUDADRV_FUNC(cuDeviceGetCount(&num_gpus), UCS_LOG_LEVEL_DIAG);
     if ((status != UCS_OK) || (num_gpus == 0)) {
         return uct_md_query_empty_md_resource(resources_p, num_resources_p);
     }
 
-    for (cuda_device = 0; cuda_device < num_gpus; ++cuda_device) {
-        uct_cuda_base_get_sys_dev(cuda_device, &sys_dev);
-        if (sys_dev != UCS_SYS_DEVICE_ID_UNKNOWN) {
-            ucs_snprintf_safe(device_name, sizeof(device_name), "GPU%d",
-                              cuda_device);
-            status = ucs_topo_sys_device_set_name(sys_dev, device_name,
-                                                  sys_device_priority);
-            ucs_assert_always(status == UCS_OK);
+    for (i = 0; i < num_gpus; ++i) {
+        status = UCT_CUDADRV_FUNC(cuDeviceGet(&cuda_device, i),
+                                  UCS_LOG_LEVEL_DIAG);
+        if (status != UCS_OK) {
+            continue;
         }
+
+        uct_cuda_base_get_sys_dev(cuda_device, &sys_dev);
+        if (sys_dev == UCS_SYS_DEVICE_ID_UNKNOWN) {
+            continue;
+        }
+
+        ucs_snprintf_safe(device_name, sizeof(device_name), "GPU%d",
+                          cuda_device);
+        status = ucs_topo_sys_device_set_name(sys_dev, device_name,
+                                              sys_device_priority);
+        ucs_assert_always(status == UCS_OK);
     }
 
     return uct_md_query_single_md_resource(component, resources_p,
                                            num_resources_p);
+}
+
+UCS_STATIC_INIT
+{
+    cuInit(0);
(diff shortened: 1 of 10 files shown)


yosefe (commented):
In general, LGTM
Should we also remove it from link (e.g do not link with libcudart)?
What about ucm and perftest?

Akshay-Venkatesh:
> In general, LGTM Should we also remove it from link (e.g do not link with libcudart)? What about ucm and perftest?

@yosefe The concern with removing RT calls from UCM is that we want to support fallback case when bistro hooks are disabled. Also, for UCM, we need to keep `-lcudart` in `$CUDA_FLAGS` under config. We could introduce `$CUDART_FLAGS` and explicitly link `-lcudart` only where needed (for eg. in ucm). 

I'll remove perftest and gtest references (but not all as one test check RT malloc hooks support) as well.

Akshay-Venkatesh:
@yosefe go perftests are failing because of unsupported memory type but I'm not sure if these are related errors. Are these tests passing in other PRs?

examples/hello_world_util.h
```diff
@@ -48,18 +47,68 @@
...
     } while(0)
 
 
 void print_common_help(void);
 
+#ifdef HAVE_CUDA
+CUcontext cuda_ctx;
+int cuda_ctx_set = 0;
+
+int init_cuda_context()
```
rakhmets, line 68: Does it mean that user should always initialize the device and create the context manually to use UCX with CUDA?
Otherwise all driver API calls will return `CUDA_ERROR_NOT_INITIALIZED` or `CUDA_ERROR_INVALID_CONTEXT`.

yosefe: Yes

rakhmets: Sorry, the comment is not relevant.
Noticed that `cuInit` is called inside static initialization in `cuda_md.c`.

yosefe:
failure seems relevant
```
[1691759077.058089] [swx-rdmz-ucx-gpu-02:12658:0]    cuda_copy_md.c:177  UCX  ERROR cuMemHostRegister_v2(address, length, 0x01)() failed: invalid device context
[1691759077.058121] [swx-rdmz-ucx-gpu-02:12658:0]          ucp_mm.c:63   UCX  ERROR failed to register address 0x7fa485695000 (host) length 1060864 on md[4]=cuda_cpy: Input/output error (md supports: host|cuda|cuda-managed)
[1691759077.058226] [swx-rdmz-ucx-gpu-02:12658:0]           mpool.c:270  UCX  ERROR Failed to allocate memory pool (name=ucp_reg_bufs) chunk: Input/output error
```

src/uct/cuda/base/cuda_md.c
```diff
@@ -71,28 +71,45 @@ uct_cuda_base_query_md_resources(uct_component_t *component,
...
-                              cuda_device);
-            status = ucs_topo_sys_device_set_name(sys_dev, device_name,
-                                                  sys_device_priority);
-            ucs_assert_always(status == UCS_OK);
+        if (sys_dev == UCS_SYS_DEVICE_ID_UNKNOWN) {
+            continue;
         }
+
+        ucs_snprintf_safe(device_name, sizeof(device_name), "GPU%d",
+                          cuda_device);
```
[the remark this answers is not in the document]
rakhmets: `i` is an ordinal number. It corresponds to the number in the list of visible devices. `cuDeviceGet` returns in `cuda_device` a device handle. The device handle will be the same for the same device regardless of device visibility.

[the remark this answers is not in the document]
rakhmets: E.g. there are two devices on the system. if we use `i` instead of `cuda_device`, then:
a. If `CUDA_VISIBLE_DEVICES=0,1`, GPU0 will correspond to device 0, and GPU1 - to device 1.
b. If `CUDA_VISIBLE_DEVICES=1,0`, GPU0 will correspond to device 1, and GPU1 - to device 0.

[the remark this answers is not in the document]
brminich: ok, seems it is more representative than ordinal number which was used before

brminich (approved):

rakhmets:
@Akshay-Venkatesh do you have any comments? If not so, I will squash the commits and merge the changes.

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -194,7 +199,7 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_dereg,
         return UCS_OK;
     }
 
-    status = UCT_CUDA_CALL_LOG_ERR(cudaHostUnregister, address);
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemHostUnregister(address));
```
Akshay-Venkatesh, line 202: @rakhmets Do we need to handle the case of attempting unregister when there is no active context?

Akshay-Venkatesh (commented):
@rakhmets Changes look good except for the one comment. If dummy registration is handled in mem_dereg implementation, then you can ignore my comment.

PR #8598, opened 2022-10-05
UCP/PROTO: fix RNDV/ATS completion

evgeny-leksikov:
## What
fix RNDV/ATS completion

## Why ?
finish callback was not invoked for generic DT


evgeny-leksikov:
> @evgeny-leksikov how was it found? do we need to add som gtest to catch it in CI?

just retested gtests locally after several merged PRs, CI with `PROTO_REQUEST_RESET=y UCX_PROTO_ENABLE=y` covers it, so once it's ready to report errors instead of warnings, it should be fine.

PR #8618, opened 2022-10-12
UCP/PROTO: fix remote MD map in RNDV protocols

evgeny-leksikov:
## What
fix remote MD map in RNDV protocols

## Why ?
CM lane does not have dedicated resource (`UCP_NULL_RESOURCE`)

### DIFF
--- a/src/ucp/rndv/proto_rndv.c
+++ b/src/ucp/rndv/proto_rndv.c
@@ -104,6 +104,10 @@ ucp_proto_rndv_md_map_to_remote(const ucp_proto_rndv_ctrl_init_params_t *params,
 
     ucs_carray_for_each(lane_cfg, ep_config->key.lanes,
                         ep_config->key.num_lanes) {
+        if (lane_cfg->rsc_index == UCP_NULL_RESOURCE) {
+            continue;
+        }
+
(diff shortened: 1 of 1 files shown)


src/ucp/rndv/proto_rndv.c
```diff
@@ -104,6 +104,12 @@ ucp_proto_rndv_md_map_to_remote(const ucp_proto_rndv_ctrl_init_params_t *params,
 
     ucs_carray_for_each(lane_cfg, ep_config->key.lanes,
                         ep_config->key.num_lanes) {
+        if (lane_cfg->rsc_index == UCP_NULL_RESOURCE) {
+            ucs_assert((lane_cfg - ep_config->key.lanes) ==
+                       ep_config->key.cm_lane);
```
yosefe, line 109: IMO this assert is not needed. can just skip the lane.

evgeny-leksikov: Ok, will remove after CI passes, I want to check if there are any other cases.

evgeny-leksikov:
@yosefe ok to merge?

yosefe (approved):

PR #8619, opened 2022-10-12
UCP/PROTO/AM: fix send status handler

evgeny-leksikov:
## What
fix proto/AM send status handler 

## Why ?
request can be completed and released

### DIFF
--- a/src/ucp/am/eager_multi.c
+++ b/src/ucp/am/eager_multi.c
@@ -145,8 +145,8 @@ static UCS_F_ALWAYS_INLINE ucs_status_t ucp_am_eager_multi_bcopy_send_func(
         packed_size = uct_ep_am_bcopy(uct_ep, UCP_AM_ID_AM_FIRST,
                                       ucp_am_eager_multi_bcopy_pack_args_first,
                                       &pack_ctx, 0);
-        status      = ucp_am_handle_user_header_send_status(
-                req, ucp_am_multi_bcopy_send_func_status(packed_size));
+        status      = ucp_am_multi_bcopy_send_func_status(packed_size);
+        status      = ucp_proto_am_handle_user_header_send_status(req, status);
     } else {
         pack_ctx.max_payload = ucp_proto_multi_max_payload(
                 req, lpriv, UCP_AM_MID_FRAG_META_LEN);
@@ -169,7 +169,14 @@ ucp_am_eager_multi_bcopy_proto_progress(uct_pending_req_t *uct_req)
     return ucp_proto_multi_bcopy_progress(
             req, req->send.proto_config->priv, ucp_proto_msg_multi_request_init,
             ucp_am_eager_multi_bcopy_send_func,
-            ucp_proto_request_bcopy_complete_success);
+            ucp_proto_request_am_bcopy_complete_success);
+}
+
+void ucp_am_eager_multi_bcopy_proto_abort(ucp_request_t *req,
+                                          ucs_status_t status)
+{
+    ucp_am_release_user_header(req);
+    ucp_request_complete_send(req, status);
 }
 
 ucp_proto_t ucp_am_eager_multi_bcopy_proto = {
@@ -179,7 +186,7 @@ ucp_proto_t ucp_am_eager_multi_bcopy_proto = {
     .init     = ucp_am_eager_multi_bcopy_proto_init,
     .query    = ucp_proto_multi_query,
     .progress = {ucp_am_eager_multi_bcopy_proto_progress},
-    .abort    = ucp_request_complete_send,
+    .abort    = ucp_proto_am_request_bcopy_abort,
     .reset    = ucp_proto_request_bcopy_reset
 };
 
--- a/src/ucp/core/ucp_am.c
+++ b/src/ucp/core/ucp_am.c
@@ -601,20 +601,22 @@ static ucs_status_t ucp_am_contig_short_reply(uct_pending_req_t *self)
 
 static ucs_status_t ucp_am_bcopy_single(uct_pending_req_t *self)
 {
-    ucp_request_t *req  = ucs_container_of(self, ucp_request_t, send.uct);
-    ucs_status_t status = ucp_do_am_bcopy_single(self, UCP_AM_ID_AM_SINGLE,
-                                                 ucp_am_bcopy_pack_args_single);
+    ucp_request_t *req = ucs_container_of(self, ucp_request_t, send.uct);
+    ucs_status_t status;
 
+    status = ucp_do_am_bcopy_single(self, UCP_AM_ID_AM_SINGLE,
+                                    ucp_am_bcopy_pack_args_single);
     status = ucp_am_handle_user_header_send_status(req, status);
     return ucp_am_bcopy_handle_status_from_pending(self, 0, 0, status);
 }
 
 static ucs_status_t ucp_am_bcopy_single_reply(uct_pending_req_t *self)
 {
-    ucp_request_t *req  = ucs_container_of(self, ucp_request_t, send.uct);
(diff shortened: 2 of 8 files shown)


src/ucp/proto/proto_am.inl
```diff
@@ -28,7 +28,9 @@ static UCS_F_ALWAYS_INLINE ucs_status_t ucp_am_handle_user_header_send_status(
 {
     ucs_status_t status;
 
-    if (ucs_unlikely(send_status == UCS_ERR_NO_RESOURCE) &&
+    if (ucs_likely(send_status == UCS_OK)) {
+        return UCS_OK;
+    } else if (ucs_unlikely(send_status == UCS_ERR_NO_RESOURCE) &&
```
hoopoepg, line 33: `else` not needed here

src/ucp/proto/proto_am.inl
```diff
@@ -28,7 +28,9 @@ static UCS_F_ALWAYS_INLINE ucs_status_t ucp_am_handle_user_header_send_status(
 {
     ucs_status_t status;
 
-    if (ucs_unlikely(send_status == UCS_ERR_NO_RESOURCE) &&
+    if (ucs_likely(send_status == UCS_OK)) {
```
ofirfarjun7, line 31: But we need to check if header was copied before returning no? for example we tried to send it and received "UCS_ERR_NO_RESOURCE", than try to send it again and received UCS_OK. After the first attempt we might copied the header.

This was the logic behind this branch.
If "UCS_ERR_NO_RESOURCE" copy if needed
Any other status means that the request was completed or failed -> release header and return status.

evgeny-leksikov: AFAIU if msg was sent (including the header), then no need to copy it at all.

ofirfarjun7: I think the question is if it was sent on the first attempt or in later attempts...

evgeny-leksikov: I see your point. Anyway, once request completed and freed we can't access it, so `UCP_REQUEST_FLAG_USER_HEADER_COPIED` must be checked before this place, I'll see how we can refactor it...

ofirfarjun7: Not sure I understand. 
Calling "ucp_am_handle_user_header_send_status" happens before calling "ucp_request_complete" or i'm missing something?

ofirfarjun7: I see... that is a problem...
Need to fix it.
Is this the only place "ucp_am_handle_user_header_send_status" is called after "ucp_request_complete"?

evgeny-leksikov: > Is this the only place "ucp_am_handle_user_header_send_status" is called after "ucp_request_complete"?

no

ofirfarjun7: ucp_am_eager_single_bcopy_reply_proto_progress?

If this are the only places (I didn't find more) maybe we need to call the header handler in "ucp_proto_am_bcopy_single_progress" after the send?

yosefe: Before this PR, send_status==UCS_OK would go to checking "req->flags & UCP_REQUEST_FLAG_USER_HEADER_COPIED" and release the mpool buffer if this check is true.
After this PR, send_status==UCS_OK would just return UCS_OK.
So where would we release the mpool buffer after this PR, if the header was copied to it?

evgeny-leksikov: yes, this is what we discussed with @ofirfarjun7 above, I'm looking how better to fix it.

evgeny-leksikov: I think we need to release `user_header` on completion. I found another path where we lose it, this is `abort`. So, it could be common place to do that in most cases @ofirfarjun7 @yosefe WDYT?

ofirfarjun7: you mean release it in ucp_request_complete?

evgeny-leksikov: > you mean release it in ucp_request_complete?

no, it depends per protocol

src/ucp/am/eager_multi.c
```diff
@@ -145,8 +145,11 @@ static UCS_F_ALWAYS_INLINE ucs_status_t ucp_am_eager_multi_bcopy_send_func(
         packed_size = uct_ep_am_bcopy(uct_ep, UCP_AM_ID_AM_FIRST,
                                       ucp_am_eager_multi_bcopy_pack_args_first,
                                       &pack_ctx, 0);
-        status      = ucp_am_handle_user_header_send_status(
-                req, ucp_am_multi_bcopy_send_func_status(packed_size));
+        status      = ucp_am_multi_bcopy_send_func_status(packed_size);
+        status      = ucp_am_handle_user_header_send_status(req, status);
+        if (ucs_likely(status != UCS_ERR_NO_RESOURCE)) {
+            ucp_am_release_user_header(req);
```
ofirfarjun7, line 151: What will happen if we send msg using bcopy multi, the first fragment is sent and then msg request is aborted? I think "ucp_am_release_user_header" will be called twice in this case.

Need to make sure "ucp_am_release_user_header" is not called twice for same msg request because we are not checking if we already released the user header. The alternative is to add some check to "ucp_am_release_user_header" before releasing the header (Check if it's NULL maybe?).

evgeny-leksikov: `bcopy_multi` protocol aborts request by `ucp_request_complete_send`, so it won't be called twice. In case of failure (or UCS_OK), the header will be released here.

evgeny-leksikov: but if `status == UCS_ERR_NO_RESOURCE` end then request will be aborted, there is a leak

yosefe: IMO better do ucp_am_release_user_header inside ucp_am_handle_user_header_send_status
we added this helper function so all relevant action would be inside it and no code duplication..

evgeny-leksikov: The problem here is that it's might be late - request is completed and released by completion callback.

evgeny-leksikov: per offline discussion - need to unify all protocols to release header desc from completion callbacks and abort to avoid extra check right after send

src/ucp/core/ucp_am.c
```diff
@@ -596,6 +600,10 @@ static ucs_status_t ucp_am_contig_short_reply(uct_pending_req_t *self)
                                        req->send.msg_proto.am.header.length,
                                        req->send.buffer, req->send.length, 1);
     status         = ucp_am_handle_user_header_send_status(req, status);
+    if (ucs_unlikely(status != UCS_ERR_NO_RESOURCE)) {
+        ucp_am_release_user_header(req);
+    }
```
yosefe, line 605: can we move this inside ucp_am_short_handle_status_from_pending?

src/ucp/core/ucp_am.c
```diff
@@ -616,6 +628,10 @@ static ucs_status_t ucp_am_bcopy_single_reply(uct_pending_req_t *self)
                                                  ucp_am_bcopy_pack_args_single_reply);
 
     status = ucp_am_handle_user_header_send_status(req, status);
+    if (ucs_unlikely(status != UCS_ERR_NO_RESOURCE)) {
+        ucp_am_release_user_header(req);
+    }
```
yosefe, line 633: can we move this inside ucp_am_bcopy_handle_status_from_pending?

yosefe (approved):

yosefe:
@ofirfarjun7 can you pls review as well?

yosefe (approved):

ofirfarjun7 (dismissed):

yosefe (approved):

PR #8621, opened 2022-10-12
GTEST: fix memory leak in test_md.exported_mkey

evgeny-leksikov:
## What
fix memory leak in test_md.exported_mkey

### DIFF
--- a/test/gtest/uct/test_md.cc
+++ b/test/gtest/uct/test_md.cc
@@ -821,11 +821,10 @@ UCS_TEST_SKIP_COND_P(test_md, exported_mkey,
     static const unsigned md_flags = UCT_MD_MEM_ACCESS_REMOTE_PUT |
                                      UCT_MD_MEM_ACCESS_REMOTE_GET;
     uct_mem_h export_memh;
-    void *ptr;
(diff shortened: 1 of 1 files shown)


hoopoepg (dismissed):

brminich (dismissed):

test/gtest/uct/test_md.cc
```diff
@@ -841,6 +841,7 @@ UCS_TEST_SKIP_COND_P(test_md, exported_mkey,
     dereg_params.memh       = export_memh;
     status                  = uct_md_mem_dereg_v2(md(), &dereg_params);
     ASSERT_UCS_OK(status);
+    free(ptr);
```
yosefe, line 844: IMO better make ptr an std::vector/string so it will be release also when test fails by ASSERT_UCS_OK

evgeny-leksikov: yes, better to use C++ allocation but for other reason - proper handling of memory allocation and code style alignment. Failed `ASSERT_UCS_OK` should kill the process, so no need to care about `ptr` release.

yosefe (approved):

PR #8622, opened 2022-10-12
UCT/DC: Fix max_rd_atom for cmd_qp connect

brminich:
## What
Do not take `max_rd_atomic` value from the `iface->config` when connecting RC QP. Always take `max_rd_atomic` from device attributes instead. 

## Why ?
When dc interface creates command QP for tag offload it creates it as RC, but its config contains `max_rd_atomic` value relevant for DC. And this value can be different for RC and DC.
Fixes the following issue when running UCX with tag offload enabled:
`ib_mlx5_dv.c:416  UCX  ERROR MBR mlx5dv_devx_obj_modify(opcode=0x503) failed, syndrome 0x6da61b: Remote I/O error`

### DIFF
--- a/src/uct/ib/rc/accel/rc_mlx5_devx.c
+++ b/src/uct/ib/rc/accel/rc_mlx5_devx.c
@@ -393,13 +393,10 @@ void uct_rc_mlx5_devx_cleanup_srq(uct_ib_mlx5_md_t *md, uct_ib_mlx5_srq_t *srq)
     uct_ib_mlx5_md_buf_free(md, srq->buf, &srq->devx.mem);
 }
 
-ucs_status_t
-uct_rc_mlx5_iface_common_devx_connect_qp(uct_rc_mlx5_iface_common_t *iface,
-                                         uct_ib_mlx5_qp_t *qp,
-                                         uint32_t dest_qp_num,
-                                         struct ibv_ah_attr *ah_attr,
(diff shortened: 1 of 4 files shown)


src/uct/ib/rc/accel/rc_mlx5_devx.c
```diff
@@ -471,9 +473,14 @@ uct_rc_mlx5_iface_common_devx_connect_qp(uct_rc_mlx5_iface_common_t *iface,
         }
     }
 
+    /* Do not take max_rd_atomic from iface config, because this function
```
yosefe, line 476: maybe fix iface->super.config.max_rd_atomic so it will be correct also for DC?

brminich: it is already correct for DC, that's why it fails here. Here RC QP is created - we always create command QP as RC QP even with DC transport. The problem is that we take DC value for creating RC QP

yosefe: Can we pass a parameter to this function so that for command QP it will be 0 or 1 for example?
because we don't want to ignore UCX_RC_MAX_RD_ATOMIC config

brminich: ic your point, will update

yosefe (approved):

PR #8623, opened 2022-10-12
UCM/CUDA: Add support for CUDA virtual memory with `cudaMallocAsync`

pentschev:
## What
Adding support for CUDA virtual/stream-ordered memory with `cudaMallocAsync`.

## Why ?
For proper CUDA virtual memory identification, add required hooking for `cudaMallocAsync`. Although this is currently adding support for `cudaMallocAsync`/`cudaMallocAsyncFromPool`, the stream-ordered memory allocator is part of the [Virtual Memory Management in CUDA](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VA.html), which is why its special case was named `UCS_MEMORY_TYPE_CUDA_VIRTUAL` for the time being.

This change was successfully tested with Dask+UCX-Py using bounce buffers setting `UCX_RNDV_FRAG_MEM_TYPE=cuda` and demonstrates to behave correctly so far, also providing similar performance to CUDA managed memory under the same conditions.

### DIFF
--- a/src/ucm/cuda/cudamem.c
+++ b/src/ucm/cuda/cudamem.c
@@ -45,15 +45,18 @@
     }
 
 /* Create a body of CUDA memory release replacement function */
-#define UCM_CUDA_FREE_FUNC(_name, _retval, _ptr_type, _mem_type) \
-    _retval ucm_##_name(_ptr_type ptr) \
+#define UCM_CUDA_FREE_FUNC(_name, _mem_type, _retval, _ptr_arg, _args_fmt, \
+                           ...) \
+    _retval ucm_##_name(UCM_FUNC_DEFINE_ARGS(__VA_ARGS__)) \
     { \
         _retval ret; \
         \
         ucm_event_enter(); \
-        ucm_trace("%s(ptr=%p)", __FUNCTION__, (void*)ptr); \
-        ucm_cuda_dispatch_mem_free((CUdeviceptr)ptr, _mem_type, #_name); \
-        ret = ucm_orig_##_name(ptr); \
+        ucm_trace("%s(" _args_fmt ")", __FUNCTION__, \
+                  UCM_FUNC_PASS_ARGS(__VA_ARGS__)); \
+        ucm_cuda_dispatch_mem_free((CUdeviceptr)(_ptr_arg), _mem_type, \
+                                   #_name); \
+        ret = ucm_orig_##_name(UCM_FUNC_PASS_ARGS(__VA_ARGS__)); \
         ucm_event_leave(); \
         return ret; \
     }
@@ -81,15 +84,35 @@ UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cuMemAllocPitch, CUresult, -1, CUdeviceptr*,
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cuMemAllocPitch_v2, CUresult, -1,
                                   CUdeviceptr*, size_t*, size_t, size_t,
                                   unsigned int)
+#if CUDA_VERSION >= 11020
+UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cuMemAllocAsync, CUresult, -1, CUdeviceptr*,
+                                  size_t, CUstream)
+UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cuMemAllocFromPoolAsync, CUresult, -1,
+                                  CUdeviceptr*, size_t, CUmemoryPool, CUstream)
+#endif
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cuMemFree, CUresult, -1, CUdeviceptr)
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cuMemFree_v2, CUresult, -1, CUdeviceptr)
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cuMemFreeHost, CUresult, -1, void*)
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cuMemFreeHost_v2, CUresult, -1, void*)
+#if CUDA_VERSION >= 11020
+UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cuMemFreeAsync, CUresult, -1, CUdeviceptr,
+                                  CUstream)
+#endif
 
 /* Runtime API */
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cudaFree, cudaError_t, -1, void*)
+#if CUDA_VERSION >= 11020
+UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cudaFreeAsync, cudaError_t, -1, void*,
+                                  cudaStream_t)
+#endif
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cudaFreeHost, cudaError_t, -1, void*)
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cudaMalloc, cudaError_t, -1, void**, size_t)
+#if CUDA_VERSION >= 11020
+UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cudaMallocAsync, cudaError_t, -1, void**,
+                                  size_t, cudaStream_t)
+UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cudaMallocFromPoolAsync, cudaError_t, -1,
+                                  void**, size_t, cudaMemPool_t, cudaStream_t)
+#endif
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cudaMallocManaged, cudaError_t, -1, void**,
                                   size_t, unsigned int)
 UCM_DEFINE_REPLACE_DLSYM_PTR_FUNC(cudaMallocPitch, cudaError_t, -1, void**,
@@ -154,10 +177,27 @@ UCM_CUDA_ALLOC_FUNC(cuMemAllocPitch_v2, UCS_MEMORY_TYPE_CUDA, CUresult,
                     CUDA_SUCCESS, (size_t)arg1 * arg2, CUdeviceptr,
                     "pitch=%p width=%zu height=%zu elem=%u", size_t*, size_t,
                     size_t, unsigned)
-UCM_CUDA_FREE_FUNC(cuMemFree, CUresult, CUdeviceptr, UCS_MEMORY_TYPE_CUDA)
-UCM_CUDA_FREE_FUNC(cuMemFree_v2, CUresult, CUdeviceptr, UCS_MEMORY_TYPE_CUDA)
-UCM_CUDA_FREE_FUNC(cuMemFreeHost, CUresult, void*, UCS_MEMORY_TYPE_HOST)
-UCM_CUDA_FREE_FUNC(cuMemFreeHost_v2, CUresult, void*, UCS_MEMORY_TYPE_HOST)
+#if CUDA_VERSION >= 11020
+UCM_CUDA_ALLOC_FUNC(cuMemAllocAsync, UCS_MEMORY_TYPE_CUDA_MANAGED, CUresult,
+                    CUDA_SUCCESS, arg0, CUdeviceptr, "size=%zu stream=%p",
+                    size_t, CUstream)
+UCM_CUDA_ALLOC_FUNC(cuMemAllocFromPoolAsync, UCS_MEMORY_TYPE_CUDA_MANAGED,
+                    CUresult, CUDA_SUCCESS, arg0, CUdeviceptr,
+                    "size=%zu pool=%p stream=%p", size_t, CUmemoryPool,
+                    CUstream)
+#endif
+UCM_CUDA_FREE_FUNC(cuMemFree, UCS_MEMORY_TYPE_CUDA, CUresult, arg0,
+                   "ptr=0x%llx", CUdeviceptr)
+UCM_CUDA_FREE_FUNC(cuMemFree_v2, UCS_MEMORY_TYPE_CUDA, CUresult, arg0,
+                   "ptr=0x%llx", CUdeviceptr)
+UCM_CUDA_FREE_FUNC(cuMemFreeHost, UCS_MEMORY_TYPE_HOST, CUresult, arg0,
+                   "ptr=%p", void*)
+UCM_CUDA_FREE_FUNC(cuMemFreeHost_v2, UCS_MEMORY_TYPE_HOST, CUresult, arg0,
+                   "ptr=%p", void*)
+#if CUDA_VERSION >= 11020
+UCM_CUDA_FREE_FUNC(cuMemFreeAsync, UCS_MEMORY_TYPE_CUDA_MANAGED, CUresult, arg0,
+                   "ptr=0x%llx, stream=%p", CUdeviceptr, CUstream)
+#endif
 
 static ucm_cuda_func_t ucm_cuda_driver_funcs[] = {
     UCM_CUDA_FUNC_ENTRY(cuMemAlloc),
@@ -165,10 +205,17 @@ static ucm_cuda_func_t ucm_cuda_driver_funcs[] = {
     UCM_CUDA_FUNC_ENTRY(cuMemAllocManaged),
     UCM_CUDA_FUNC_ENTRY(cuMemAllocPitch),
     UCM_CUDA_FUNC_ENTRY(cuMemAllocPitch_v2),
+#if CUDA_VERSION >= 11020
+    UCM_CUDA_FUNC_ENTRY(cuMemAllocAsync),
+    UCM_CUDA_FUNC_ENTRY(cuMemAllocFromPoolAsync),
+#endif
     UCM_CUDA_FUNC_ENTRY(cuMemFree),
     UCM_CUDA_FUNC_ENTRY(cuMemFree_v2),
     UCM_CUDA_FUNC_ENTRY(cuMemFreeHost),
     UCM_CUDA_FUNC_ENTRY(cuMemFreeHost_v2),
+#if CUDA_VERSION >= 11020
+    UCM_CUDA_FUNC_ENTRY(cuMemFreeAsync),
+#endif
     {{NULL}, NULL}
 };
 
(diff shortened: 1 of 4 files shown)


pentschev:
cc @Akshay-Venkatesh for awareness

Akshay-Venkatesh:
@pentschev Looks like `cuda_copy` transport has the added capability of detecting virtual memory and UCP happens to handle virtual pinned memory the same way as managed memory. When mallocasync supports managed memory the changes introduced through this PR would also need to distinguish types in UCS_MEMORY_TYPE_CUDA_VIRTUAL . 

The general question is - What are the performance expectations for mallocasync memory?
1. If the expectation is that staging through pinned memory for both intra-node and inter-node paths, then we can rework without introducing UCS_MEMORY_TYPE_CUDA_VIRTUAL but instead by adding mempool attribute to cuda memory (whether be it pinned malloc async or managed malloc async in the future). 
2. If the expectation is that mallocasync memory needs to use IPC capabilities introduced for mempools, then we would need to introduce new paths in `cuda_ipc` transport.

pentschev:
> @pentschev Looks like `cuda_copy` transport has the added capability of detecting virtual memory and UCP happens to handle virtual pinned memory the same way as managed memory. When mallocasync supports managed memory the changes introduced through this PR would also need to distinguish types in UCS_MEMORY_TYPE_CUDA_VIRTUAL .

This is a good point. Is there a mechanism for distinguishing "sub-types", or would we have to introduce a new type, e.g., `UCS_MEMORY_TYPE_CUDA_VIRTUAL_MANAGED`?

> The general question is - What are the performance expectations for mallocasync memory?
> 
> 1. If the expectation is that staging through pinned memory for both intra-node and inter-node paths provides enough performance, then we can rework without introducing UCS_MEMORY_TYPE_CUDA_VIRTUAL but instead by adding mempool attribute to cuda memory (whether be it pinned malloc async or managed malloc async in the future).
> 2. If the expectation is that mallocasync memory needs to use IPC capabilities introduced for mempools, then we would need to introduce new paths in `cuda_ipc` transport.

This is a very difficult question to answer. In a perfect world we would, of course, want to have the same performance as we do for CUDA "default" memory (PA-backed memory allocated via `cudaMalloc`), but that generally is unfeasible in practice. As for the tests we did so far, using the staging buffers provide very similar performance to PA-backed CUDA allocations, the penalty we observed was roughly 10%, which is definitely acceptable now. However, the tests currently involved single-node DGX-1 only, so I can't say whether this performance will scale.

In the future I would say it would be worth having `cuda_ipc` support as well, it feels that as we push towards newer systems with much higher bandwidth limits for `cuda_ipc` like the H100, staging allocations may impose a higher overall penalty particularly if we have to transfer large amounts of data at once, where the time spent to stage buffers may eventually be too high and would otherwise force us to dedicated more and more memory to the staging buffers to work around that issue. Please take this with a grain of salt, I do not have numbers at this time to back or falsify my theory, just trying to make an educated guess here.

yosefe:
@pentschev IMO adding another memory type is pretty intrusive. Why is it different than CUDA pinned memory?

pentschev:
> @pentschev IMO adding another memory type is pretty intrusive. How is it different from CUDA pinned memory?

A complete different API set is required for handling both [virtual](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VA.html) and [stream-ordered](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC). Virtual and stream-ordered memory are not PA-backed as regular pinned memory, so supporting `cuda_ipc` or verbs will have different requirements than for pinned memory, but not for `cuda_copy` or staging buffers, which will be supported with this PR already.

To be fair, I do not know internals of UCX at this level that well, so there's definitely a chance that adding a new memory type is indeed not required. My reasoning for this is currently that virtual/stream-ordered memory will have different requirements than pinned memory and will most likely require a different code-path for `cuda_ipc` and verbs. Plus, the way testing is organized in UCX requires having a different memory type, for example passing `ucx_perftest -m cuda-virtual` would not be possible without a different memory type and it would definitely be useful to be able to test/profile virtual/stream-ordered memory individually as well. Finally, as Akshay pointed out above, at one point we will have "virtual managed" memory, integrating both concepts, but it currently isn't clear to me whether the requirements will be different or not.

I'm certainly open to not having another memory type if we can manage the reasons I pointed out above in a different way.

pentschev:
I now pushed changes removing the new memory type.

Also replying @yosefe 's points:

> * It could be in another PR to expedite merging this PR for v1.14.x

Sounds reasonable to me. The only regression from the original PR is we now can't test `cudaMallocAsync` in: https://github.com/openucx/ucx/blob/c5d5b0ca9b77aace33bf90eec42743084e91941b/test/gtest/common/mem_buffer.cc#L163-L170

Are there any other existing ways we could test that there besides having separate memory type, or would that have to be done by adding a new "allocation method" as you're suggesting for perftest?

> * We can consider adding "allocation method" parameter to perftest, in addition to memory type. Even pinned memory can be allocated by more than one API. Not to mention host memory...

That would be great, it's very useful to be able to test and verify performance of the stream-ordered allocator, as it is very different under-the-hood than pinned/managed memory.

src/uct/cuda/base/cuda_md.c
```diff
@@ -127,6 +128,8 @@ uct_cuda_base_query_attributes(uct_cuda_copy_md_t *md, const void *address,
     attr_data[1] = &is_managed;
     attr_type[2] = CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL;
     attr_data[2] = &cuda_device;
+    attr_type[3] = CU_POINTER_ATTRIBUTE_CONTEXT;
+    attr_data[3] = &cuda_mem_ctx;
```
pentschev, line 132: @Akshay-Venkatesh `cudaMallocAsync` has first appeared in 11.2.0, but `CU_POINTER_ATTRIBUTE_MEMPOOL_HANDLE` only appeared in 11.3.0, therefore I changed the check here which should still be equivalent, based on offline conversations with the CUDA team, see the updated code comments just below. Also, `CU_POINTER_ATTRIBUTE_CONTEXT` has existed since at least CUDA 10.0, so I believe we don't need any extra preprocessor checks here, what do you think?

src/ucm/cuda/cudamem.c
```diff
@@ -154,23 +179,42 @@ UCM_CUDA_ALLOC_FUNC(cuMemAllocPitch_v2, UCS_MEMORY_TYPE_CUDA, CUresult,
...
+         UCM_CUDA_FUNC_ENTRY(cuMemAllocFromPoolAsync),
+#endif
+         UCM_CUDA_FUNC_ENTRY(cuMemFree),
+         UCM_CUDA_FUNC_ENTRY(cuMemFree_v2),
+         UCM_CUDA_FUNC_ENTRY(cuMemFreeHost),
+         UCM_CUDA_FUNC_ENTRY(cuMemFreeHost_v2),
+#if CUDA_VERSION >= 11020
+         UCM_CUDA_FUNC_ENTRY(cuMemFreeAsync),
+#endif
+         {{NULL}, NULL}};
```
bureddy, line 217: for completeness, can you add hooks below for runtime APIs as well? it might useful in when bistro hooks fails

src/ucm/cuda/cudamem.c
```diff
@@ -58,6 +58,21 @@
         return ret; \
     }
 
+/* Create a body of CUDA memory release replacement function */
+#define UCM_CUDA_FREE_ASYNC_FUNC(_name, _retval, _ptr_type, _stream_type, \
```
yosefe, line 62: can we extend UCM_CUDA_FREE_FUNC to have var-args instead of _stream_type?
see for example how UCM_CUDA_ALLOC_FUNC is using varargs

src/ucm/cuda/cudamem.c
```diff
@@ -154,23 +179,42 @@ UCM_CUDA_ALLOC_FUNC(cuMemAllocPitch_v2, UCS_MEMORY_TYPE_CUDA, CUresult,
...
-    UCM_CUDA_FUNC_ENTRY(cuMemAllocPitch),
-    UCM_CUDA_FUNC_ENTRY(cuMemAllocPitch_v2),
-    UCM_CUDA_FUNC_ENTRY(cuMemFree),
-    UCM_CUDA_FUNC_ENTRY(cuMemFree_v2),
-    UCM_CUDA_FUNC_ENTRY(cuMemFreeHost),
-    UCM_CUDA_FUNC_ENTRY(cuMemFreeHost_v2),
-    {{NULL}, NULL}
-};
+static ucm_cuda_func_t ucm_cuda_driver_funcs[] =
+        {UCM_CUDA_FUNC_ENTRY(cuMemAlloc),
```
yosefe, line 201: can we avoid changing the layout/indentation of the whole array and just add new items?

src/ucm/cuda/cudamem.h
```diff
@@ -20,10 +20,18 @@ CUresult ucm_cuMemAllocPitch(CUdeviceptr *dptr, size_t *pPitch,
 CUresult ucm_cuMemAllocPitch_v2(CUdeviceptr *dptr, size_t *pPitch,
                                 size_t WidthInBytes, size_t Height,
                                 unsigned int ElementSizeBytes);
+#if CUDA_VERSION >= 11020
```
yosefe, line 23: the `#if ` is probably not needed here since it's just func prototype, it's not calling any Cuda API here

src/ucm/cuda/cudamem.h
```diff
@@ -20,10 +20,18 @@ CUresult ucm_cuMemAllocPitch(CUdeviceptr *dptr, size_t *pPitch,
...
+CUresult ucm_cuMemAllocAsync(CUdeviceptr *dptr, size_t size, CUstream hStream);
+CUresult ucm_cuMemAllocFromPoolAsync(CUdeviceptr *dptr, size_t size,
+                                     CUmemoryPool pool, CUstream hStream);
+#endif
 CUresult ucm_cuMemFree(CUdeviceptr dptr);
 CUresult ucm_cuMemFree_v2(CUdeviceptr dptr);
 CUresult ucm_cuMemFreeHost(void *p);
 CUresult ucm_cuMemFreeHost_v2(void *p);
+#if CUDA_VERSION >= 11020
```
yosefe, line 32: same here

src/uct/cuda/base/cuda_md.c
```diff
@@ -140,9 +143,19 @@ uct_cuda_base_query_attributes(uct_cuda_copy_md_t *md, const void *address,
         return status;
     }
 
-    if (is_managed) {
-        /* cuMemGetAddress range does not support managed memory so use provided
-         * address and length as base address and alloc length respectively */
+    if (is_managed || !cuda_mem_ctx) {
```
yosefe, line 146: (cuda_mem_ctx == NULL)

src/uct/ib/rc/accel/rc_mlx5_iface.c
```diff
@@ -755,6 +755,12 @@ UCS_CLASS_INIT_FUNC(uct_rc_mlx5_iface_common_t, uct_iface_ops_t *tl_ops,
         return UCS_ERR_INVALID_PARAM;
     }
 
+    if (mlx5_config->tm.seg_size > UCT_IB_MLX5_MP_RQ_BYTE_CNT_MASK) {
+        ucs_error("TM segment size is too big %ld, it must not exceed %d",
+                  mlx5_config->tm.seg_size, UCT_IB_MLX5_MP_RQ_BYTE_CNT_MASK);
+        return UCS_ERR_INVALID_PARAM;
```
yosefe, line 761: merge artifact?

src/ucm/cuda/cudamem.c
pentschev: I've been trying to do but failed so far. I tried two different approaches, described below.

The first one would be just to add `_args_format` and `...`, but that would then require existing functions to also pass those empty arguments, and it seems problematic to handle empty var-args, not sure if there are any preprocessor macros currently to handle that case, plus it would pollute the existing `UCM_CUDA_FREE_FUNC` calls with empty parameters.

The second was to pass everything as var-args, including the pointer. We could then extract the first argument (pointer) with `UCS_PP_TUPLE_0`, and then make use of it like the explicit parameter is currently used. However, this would then require passing `__VA_ARGS__` to `ucm_trace` which would need type-casting of the individual variables.

I'm not very experience writing var-arg macros, so maybe you could suggest other approaches?

src/ucm/cuda/cudamem.c
pentschev: Done, as per our conversation offline this may be ignored.

src/ucm/cuda/cudamem.c
pentschev: Done.

src/ucm/cuda/cudamem.h
pentschev: Done.

src/ucm/cuda/cudamem.h
pentschev: Done.

src/uct/cuda/base/cuda_md.c
pentschev: Done.

src/uct/ib/rc/accel/rc_mlx5_iface.c
pentschev: Indeed, sorry for that, should be gone now.

src/ucm/cuda/cudamem.h
pentschev: The lack of `#if` fails for older CUDA because `cudaMemPool_t`/`CUmemoryPool` only appeared in CUDA 11.2. Seems like we need those directives anyway, or is there some better way of handling that case?

pentschev: For the time being I reinstated the `#if`s. I guess one alternative would be to define `cudaMemPool_t`/`CUmemoryPool` ourselves for older CUDA versions, but feels like that would pollute code more so not sure this is a great alternative.

test/gtest/ucm/cuda_hooks.cc
```diff
@@ -25,6 +25,18 @@ static void cuda_mem_free_callback(ucm_event_type_t event_type,
...
+static void check_cuda_malloc_async_capable() {
+    int supported_cuda_version;
+    CUresult cu_err;
+
+    cu_err = cuDriverGetVersion(&supported_cuda_version);
+    ASSERT_EQ(cu_err, CUDA_SUCCESS);
+
+    if (supported_cuda_version < 11020) {
+		GTEST_SKIP_("cuMemAllocAsync not available");
+    }
```
pentschev, line 37: Originally I tried to do this with `cuGetProcAddress`, but `cuGetProcAddress` only got introduced in CUDA 11.3. I guess we could skip tests for that version as well but that feels like introducing an artificial cap, so checking the latest CUDA version supported by the driver seems like a better alternative here.

yosefe:
@pentschev i've pushed another commit to this PR for https://github.com/openucx/ucx/pull/8623#discussion_r996029921

yosefe:
> Awesome, thanks so much @yosefe ! I got so close, but the magic I was missing was `arg0`, would you mind explaining/pointing me to how `arg0` gets expanded there?

arg0 is the first argument in the vararg list. it means that the released pointer (that we need to pass down to mem hooks) is the first argument to the cuda function. We have similar thing with allocation functions, where we provide the expression to calculate the allocated buffer size based to arguments.

> I'm not really sure how that happens, it seems to occur for all CUDA versions, including those that do support cudaMallocAsync. Do you have any clue as to why the symbols are not being found in CI? Locally they pass successfully for me with CUDA 11.5.

There was a missing definition for cudaMallocAsync that I added in my commit. Hopefully the tests would pass now.

yosefe (approved):

src/ucm/cuda/cudamem.c
```diff
@@ -154,21 +177,45 @@ UCM_CUDA_ALLOC_FUNC(cuMemAllocPitch_v2, UCS_MEMORY_TYPE_CUDA, CUresult,
...
+                    size_t, CUstream)
+UCM_CUDA_ALLOC_FUNC(cuMemAllocFromPoolAsync, UCS_MEMORY_TYPE_CUDA_MANAGED,
+                    CUresult, CUDA_SUCCESS, arg0, CUdeviceptr,
+                    "size=%zu pool=%p stream=%p", size_t, CUmemoryPool,
+                    CUstream)
+#endif
+UCM_CUDA_FREE_FUNC(cuMemFree, UCS_MEMORY_TYPE_CUDA, CUresult, arg0,
+                   "ptr=0x%llx", CUdeviceptr)
+UCM_CUDA_FREE_FUNC(cuMemFree_v2, UCS_MEMORY_TYPE_CUDA, CUresult, arg0,
+                   "ptr=0x%llx", CUdeviceptr)
```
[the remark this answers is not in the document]
yosefe: @Akshay-Venkatesh actually it doesn't matter which memory type is used for memory free event.
we can remove memtype from free altogether (pass UCS_MEM_TYPE_UNKNOWN to the event) but since that "issue" existed before this PR, IMO it should not block this PR for v1.14

src/ucm/cuda/cudamem.c
```diff
@@ -181,13 +228,34 @@ UCM_CUDA_ALLOC_FUNC(cudaMallocManaged, UCS_MEMORY_TYPE_CUDA_MANAGED,
...
+#if CUDA_VERSION >= 11020
+UCM_CUDA_ALLOC_FUNC(cudaMallocAsync, UCS_MEMORY_TYPE_CUDA_MANAGED, cudaError_t,
+                    cudaSuccess, arg0, void*, "size=%zu stream=%p", size_t,
+                    cudaStream_t)
+UCM_CUDA_ALLOC_FUNC(cudaMallocFromPoolAsync, UCS_MEMORY_TYPE_CUDA_MANAGED, cudaError_t,
+                    cudaSuccess, arg0, void*, "size=%zu pool=%p stream=%p", size_t,
+                    cudaMemPool_t, cudaStream_t)
+#endif
+UCM_CUDA_FREE_FUNC(cudaFree, UCS_MEMORY_TYPE_CUDA, cudaError_t, arg0,
+                   "devPtr=%p", void*)
```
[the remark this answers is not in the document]
yosefe: it does: ucs_memtype_cache_update_internal uses "mem_type" parameter only if action == UCS_MEMTYPE_CACHE_ACTION_SET_MEMTYPE.
When called from memory free hook, action is UCS_MEMTYPE_CACHE_ACTION_REMOVE

pentschev:
This passes now, thanks so much for all the help in getting everything fixed @yosefe !

yosefe:
@pentschev can you pls squash the commits to one commit?

yosefe (approved):

pentschev:
Thanks @yosefe for the help in getting things correct here!

PR #8625, opened 2022-10-13
AZP: Add ubuntu20 container with rocm software stack

edgargabriel:
## What
add a rocm docker container for testing. Build only at the moment. 

## Why ?
simplify code development and maintenance for rocm devices

### DIFF
--- a/buildlib/pr/main.yml
+++ b/buildlib/pr/main.yml
@@ -151,6 +151,9 @@ resources:
     - container: ubuntu20_cuda11
       image: rdmz-harbor.rdmz.labs.mlnx/ucx/ubuntu20.04-mofed5-cuda11:1
       options: $(DOCKER_OPT_VOLUMES) $(DOCKER_OPT_GPU)
+    - container: ubuntu2004_rocm
+      image: registry.hub.docker.com/rocm/dev-ubuntu-20.04:latest
+      options: $(DOCKER_OPT_VOLUMES)
 
 stages:
   - stage: Codestyle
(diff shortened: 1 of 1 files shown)


buildlib/pr/main.yml
```diff
@@ -151,6 +151,9 @@ resources:
     - container: ubuntu20_cuda11
       image: rdmz-harbor.rdmz.labs.mlnx/ucx/ubuntu20.04-mofed5-cuda11:1
       options: $(DOCKER_OPT_VOLUMES) $(DOCKER_OPT_GPU)
+    - container: ubuntu20_rocm
+      image: registry.hub.docker.com/rocm/dev-ubuntu-20.04
```
yosefe, line 155: seems image is missing tag
A container resource with name ubuntu2004_rocm could not be found. The container resource does not exist. If you intended to specify an image, use NAME:TAG or NAME@DIGEST. For example, ubuntu:latest

edgargabriel: I appended :latest, lets see whether this works

edgargabriel:
Not sure whether the generic docker-hub rocm image  will work, the error is because of missing modules environment. I had a brief look at the code in buildlibs/az-helper.sh, it would be a major hack to make these scripts work without modules

yosefe:
> Not sure whether the generic docker-hub rocm image will work, the error is because of missing modules environment. I had a brief look at the code in buildlibs/az-helper.sh, it would be a major hack to make these scripts work without modules

Pls see https://github.com/openucx/ucx/blob/master/buildlib/dockers/ubuntu-release.Dockerfile for a receipt to build container for build test

PR #8631, opened 2022-10-14
UCT/BASE/IB: Pack component_name to global_id in MD attributes

dmitrygx:
## What

Pack `component_name` to `global_id` in MD attributes.

## Why ?

To make sure that `global_id` is unique even for different components.

## How ?

1. Save `md_attr->global_id` to a local variable.
2. Copy `md->component->name` to `md_attr->global_id`.
3. And copy a data from a local variable of `global_id` to `md_attr->global_id + strlen(md->component->name)`.

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -271,8 +271,9 @@ typedef struct {
 
 static ucs_status_t uct_ib_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
 {
-    uct_ib_md_t *md = ucs_derived_of(uct_md, uct_ib_md_t);
-    uint64_t guid   = IBV_DEV_ATTR(&md->dev, sys_image_guid);
+    uct_ib_md_t *md              = ucs_derived_of(uct_md, uct_ib_md_t);
+    size_t component_name_length = strlen(md->super.component->name);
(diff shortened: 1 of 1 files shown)


src/uct/base/uct_md.c
```diff
@@ -458,9 +460,15 @@ static ucs_status_t uct_md_attr_v2_init(uct_md_h md, uct_md_attr_v2_t *md_attr)
         return status;
     }
 
-    /* Component name + data */
+    memcpy(global_id, md_attr->global_id,
+           sizeof(md_attr->global_id) - component_name_length);
+    memcpy(md_attr->global_id, md->component->name, component_name_length);
+    memcpy(UCS_PTR_BYTE_OFFSET(md_attr->global_id,
+                               component_name_length),
+           global_id, sizeof(md_attr->global_id) - component_name_length);
```
yosefe, line 468: IMO it's simpler to just do it in UCT IB.
Besides, it doesn't always have to be the component name in the prefix. We just decided it as convention. But it could be anything unique per component. So no need to have it in the "base" code.

dmitrygx: done

yosefe (dismissed):

brminich (approved):

dmitrygx:
@brminich @yosefe squashed, could you approve pls?

PR #8635, opened 2022-10-16
GTEST/UCP: handle proto_enable if it's set externally

evgeny-leksikov:
## What
handle proto_enable configuration if it's set externally

### DIFF
--- a/test/gtest/ucp/test_ucp_mmap.cc
+++ b/test/gtest/ucp/test_ucp_mmap.cc
@@ -199,7 +199,7 @@ void test_ucp_mmap::test_rkey_management(ucp_mem_h memh, bool is_dummy,
     EXPECT_TRUE(ucs_test_all_flags(memh->md_map, rkey->md_map));
 
     /* Test remote key protocols selection */
-    if (enable_proto()) {
+    if (m_ucp_config->ctx.proto_enable) {
         test_rkey_proto(memh);
     } else {
         bool have_rma              = resolve_rma(&receiver(), rkey);


yosefe:
@evgeny-leksikov this PR seems wierd
1. maybe check env var? because the output of test_ucp_mmap::enable_proto() should affect m_ucp_config so it seems like a circular dependency
2. aren't there similar functions in other test classes? need to change all of them?
3. if there is already a test variant with enabled protocols v2, why need to force other variants to also use protocols v2?

evgeny-leksikov:
> @evgeny-leksikov this PR seems wierd
> 
> 1. maybe check env var? because the output of test_ucp_mmap::enable_proto() should affect m_ucp_config so it seems like a circular dependency
> 2. aren't there similar functions in other test classes? need to change all of them?
> 3. if there is already a test variant with enabled protocols v2, why need to force other variants to also use protocols v2?

so, maybe check env var and skip the test if `UCX_PROTO_ENABLE==y` **and** `!enable_proto()`

yosefe:
> so, maybe check env var and skip the test if `UCX_PROTO_ENABLE==y` **and** `!enable_proto()`

It's possible, but why needed? to shorten the test time?

evgeny-leksikov:
> It's possible, but why needed? to shorten the test time?

there is an inconsistency in test and library
https://github.com/openucx/ucx/blob/a1ff474a72097469a6b0b7a614c19393275e4d73/test/gtest/ucp/test_ucp_mmap.cc#L202
which leads to assertion in UCP

yosefe:
> > It's possible, but why needed? to shorten the test time?
> 
> there is an inconsistency in test and library
> 
> https://github.com/openucx/ucx/blob/a1ff474a72097469a6b0b7a614c19393275e4d73/test/gtest/ucp/test_ucp_mmap.cc#L202
> 
> 
> which leads to assertion in UCP

Can we check content of m_ucp_config in this "if" ?

test/gtest/ucp/test_ucp_mmap.cc
```diff
@@ -199,7 +199,7 @@ void test_ucp_mmap::test_rkey_management(ucp_mem_h memh, bool is_dummy,
     EXPECT_TRUE(ucs_test_all_flags(memh->md_map, rkey->md_map));
 
     /* Test remote key protocols selection */
-    if (enable_proto()) {
+    if (enable_proto() || m_ucp_config->ctx.proto_enable) {
```
yosefe, line 202: do we really need to keep enable_proto() in this condition?

yosefe (approved):

PR #8636, opened 2022-10-17
UCT/IB/DEVX: Fix HAVE_DEVX macro

Artemy-Mellanox:
## Why
Fixing internal issue [3232874](https://redmine.mellanox.com/issues/3232874)

### DIFF
--- a/src/uct/ib/mlx5/ib_mlx5.h
+++ b/src/uct/ib/mlx5/ib_mlx5.h
@@ -767,6 +767,9 @@ void uct_ib_mlx5_devx_uar_cleanup(uct_ib_mlx5_devx_uar_t *uar);
 void uct_ib_mlx5_txwq_validate_always(uct_ib_mlx5_txwq_t *wq, uint16_t num_bb,
                                       int hw_ci_updated);
 
+void uct_ib_mlx5_parse_cqe_zipping(uct_ib_mlx5_md_t *md,
+                                   const uct_ib_mlx5_iface_config_t *mlx5_config,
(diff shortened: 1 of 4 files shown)


src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -939,6 +938,11 @@ uct_ib_mlx5_devx_query_qp_peer_info(uct_ib_iface_t *iface, uct_ib_mlx5_qp_t *qp,
 
 static inline void uct_ib_mlx5_devx_destroy_qp(uct_ib_mlx5_md_t *md, uct_ib_mlx5_qp_t *qp) { }
 
+static inline uint8_t
+uct_ib_mlx5_devx_md_get_counter_set_id(uct_ib_mlx5_md_t *md, uint8_t port_num) {
```
yosefe, line 942: maybe use HAVE_DEVX inside uct_ib_mlx5_iface_get_counter_set_id?

Artemy-Mellanox: may we deside on it? here I followed [kernel style](https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst#21-conditional-compilation)

yosefe: that seems a good and justified approach, though we never followed it yet in the code
so can keep it so, just fix code style { on next line

yosefe (approved):

yosefe (approved):

PR #8643, opened 2022-10-18
UCP/CORE: valgrind max lanes test fails

shasson5:
## What
test


yosefe:
@shasson5 seems this PR is generating too many logs which kills the test machines
need to print logs only from the failing test and not from all

PR #8645, opened 2022-10-18
TEST/CONFIG: Fixes #8593

### LINKED ISSUE #8593: UCX compilation fails with --with-mpi when shmem.h is not present in the system
### Describe the bug
When UCX compiled with --with-mpi, it automatically assumes the presence of shmem.h header file.

### Steps to Reproduce
```
../contrib/configure-release --prefix /home/ec2-user/ucx_prod --with-openmp --with-mpi
```

### Setup and versions
Master

```
make[2]: Entering directory `/home/ec2-user/ucx/build_prod/test/mpi'
  CC       test_memhooks_lib.lo
  CC       test_memhooks.o
  CC       shmem_pingpong.o
/home/ec2-user/ucx/contrib/../test/mpi/shmem_pingpong.c:13:10: fatal error: shmem.h: No such file or directory
 #include <shmem.h>
          ^~~~~~~~~
```

shamisp:
Test that shmemcc is present as the part of MPI install.

## What
Fixing #8593 

## Why ?
The makefile should not assume that every MPI install comes with SHMEM.

## How ?
Check that shmemcc is part of MPI install, if not disable the build.

### DIFF
--- a/config/m4/mpi.m4
+++ b/config/m4/mpi.m4
@@ -26,6 +26,8 @@ AS_IF([test "x$with_mpi" = xyes],
         [
         AC_ARG_VAR(MPICC,[MPI C compiler command])
         AC_PATH_PROGS(MPICC,mpicc mpiicc,"",$mpi_path)
+        AC_ARG_VAR(SHMEMCC,[SHMEM C compiler command])
+        AC_PATH_PROGS(SHMEMCC,shmemcc,"",$mpi_path)
         AC_ARG_VAR(MPIRUN,[MPI launch command])
         AC_PATH_PROGS(MPIRUN,mpirun mpiexec aprun orterun,"",$mpi_path)
         AS_IF([test -z "$MPIRUN"],
@@ -38,4 +40,5 @@ AS_IF([test -n "$MPICC"],
       [mpi_enable=disabled])
 AM_CONDITIONAL([HAVE_MPI],    [test -n "$MPIRUN"])
 AM_CONDITIONAL([HAVE_MPICC],  [test -n "$MPICC"])
+AM_CONDITIONAL([HAVE_SHMEMCC],[test -n "$SHMEMCC"])
 AM_CONDITIONAL([HAVE_MPIRUN], [test -n "$MPIRUN"])
--- a/test/mpi/Makefile.am
+++ b/test/mpi/Makefile.am
@@ -8,29 +8,39 @@
 CC = $(MPICC)
 LD = $(MPICC)
 
-if HAVE_MPICC
 
 # Test application for memory hooks when running with MPI
 # (some MPIs have hooks of their own and we make sure ours still work)
-noinst_PROGRAMS       = test_memhooks shmem_pingpong
+noinst_PROGRAMS       =
+
+if HAVE_MPICC
+noinst_PROGRAMS      += test_memhooks
+endif
+
+if HAVE_SHMEMCC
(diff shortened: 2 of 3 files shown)


shamisp:
@yosefe is there chance that some of these machines do not run autogen ?

```
configure: Documents only requested
checking that generated files are newer than configure... done
configure: error: conditional "HAVE_SHMEMCC" was never defined.
Usually this means the macro was only invoked conditionally.
##[error]Bash exited with code '1'.
```

config/m4/mpi.m4
```diff
@@ -26,6 +26,8 @@ AS_IF([test "x$with_mpi" = xyes],
         [
         AC_ARG_VAR(MPICC,[MPI C compiler command])
         AC_PATH_PROGS(MPICC,mpicc mpiicc,"",$mpi_path)
+        AC_ARG_VAR(SHMEMCC,[SHMEM C compiler command])
+        AC_PATH_PROGS(SHMEMCC,shmemcc,"",$mpi_path)
```
mike-dubman, line 30: isn`t it ```oshcc```?

config/m4/mpi.m4
```diff
@@ -38,4 +40,5 @@ AS_IF([test -n "$MPICC"],
       [mpi_enable=disabled])
 AM_CONDITIONAL([HAVE_MPI],    [test -n "$MPIRUN"])
 AM_CONDITIONAL([HAVE_MPICC],  [test -n "$MPICC"])
+AM_CONDITIONAL([HAVE_SHMEMCC],[test -n "$SHMEMCC"])
```
yosefe, line 43: need to define it also in https://github.com/openucx/ucx/blob/83a23dcecd117f7daef3e37136f7e730d4adba9f/configure.ac#L223
when UCX is built with `--with-docs-only`

config/m4/mpi.m4
yosefe: AFAIK `shmemcc` is the standard name

test/mpi/Makefile.am
```diff
@@ -12,7 +12,12 @@ if HAVE_MPICC
 
 # Test application for memory hooks when running with MPI
 # (some MPIs have hooks of their own and we make sure ours still work)
-noinst_PROGRAMS       = test_memhooks shmem_pingpong
+noinst_PROGRAMS       = test_memhooks
+
+if HAVE_SHMEMCC
```
yosefe, line 17: HAVE_SHMEMCC better be outside of HAVE_MPICC condition

config/m4/mpi.m4
shamisp: Ah..I will fix

test/mpi/Makefile.am
shamisp: Since this is MPI test in MPI folder IMO it makes much more sense to make MPI external. SHMEM test is a subset in this case

config/m4/mpi.m4
shamisp: shmemcc, this goes back to SGI implementation

shamisp:
@yosefe pls take a look

test/mpi/Makefile.am
yosefe: HAVE_MPICC and HAVE_SHMEMCC are tested independently so it makes sense to not nest them. It also makes the makefile easier to read by less nesting of conditions.

yosefe (approved):
pls squash

PR #8648, opened 2022-10-20
UCP/PROTO: Changed top node name in proto selection tree.

rakhmets:
## What
Changed top node name in proto selection tree.

## Why ?
Using of `frag` is confusing because `fragment` is already used in protocol terminology. E.g. single/multiple fragment bcopy/zcopy protocols.


rakhmets:
node_name_before
node_name_after

brminich (approved):

yosefe (approved):

PR #8652, opened 2022-10-21
BUILD/COMPILATION: Switch to C99 standard

ivankochin:
## What
Force compilation using C99 standard.

## Why ?
C99 (with GNU extensions) is compatible with the current codebase but allows to use [additional features](https://en.cppreference.com/w/c/99).

### DIFF
--- a/config/m4/compiler.m4
+++ b/config/m4/compiler.m4
@@ -530,6 +530,21 @@ BASE_CXXFLAGS="$BASE_CFLAGS"
 
 
 #
+# Check for C99 standard support
+#
+CHECK_COMPILER_FLAG([-std=gnu99], [-std=gnu99],
+                    [AC_LANG_SOURCE([[
+                          int main(int argc, char** argv){
+                            // check for for loop init, one-line comment and
+                            // implicit return in main function
+                            for (int i = 0; i < 10; ++i) {}
+                          }
+                    ]])],
+                    [BASE_CFLAGS="$BASE_CFLAGS -std=gnu99"],
+                    [AC_MSG_ERROR([Cannot continue. Please install C compiler that supports C99.])])
+
+
+#
 # Add strict flags supported by C compiler only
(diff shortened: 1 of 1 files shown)


config/m4/compiler.m4
```diff
@@ -529,6 +529,21 @@ ADD_COMPILER_FLAGS_IF_SUPPORTED([[-Wno-missing-field-initializers],
 BASE_CXXFLAGS="$BASE_CFLAGS"
 
 
+#
+# Check for C99 standard support
+#
+CHECK_COMPILER_FLAG([-std=gnu99], [-std=gnu99],
```
ivankochin, line 535: We can change `-std=gnu99` to `-std=c11` to be fully standard compliant. But it also requires changes all `asm` and `typeof` constructions to `__asm__` and `__typeof__` since constructions without underscores are GNU extension and not a part of the standard.

But if using gnu extensions is not a problem, then current solution was checked at GCC, Clang and Intel compilers and they all support this option. So if we want change nothing, then we can leave `-std=gnu99` option and it should work well on most compilers.

@yosefe @gleon99 what you think?

gleon99: @shamisp can we define what is the condition and when we'll be able to switch to C11?
That's a transition from a 23-years old standard to only an 11-years old one :)

shamisp: I would not go it in one go. Move to g99, which is already a progress for us. See how the release process goes.
If this goes well, prepare c11 asm patches and anything else that comes up and push it for the next release.

If you ask me, I would spend the resources on BSD support ;-)

ivankochin: @gleon99 it seems like some platforms that are important for us have no full C11 support by default. For example CentOS 7 has GCC 4.8.5 by default which [doesn't support some parts of C11 standard (atomics and thread local)](https://gcc.gnu.org/wiki/C11Status).

gleon99:

PR #8655, opened 2022-10-23
UCP/WIREUP_CM/BUG: Bug Fix - Wrong multiple indirection in ep cleanup

ofirfarjun7:
## What
Wrong multiple indirection of uct_pending_purge_callback_t argument

## Why ?
Will cause segmentation fault if ucp_request_purge_enqueue_cb is called with non empty pending queue.

## How ?
Remove wrong indirection.


shamisp:
This is the risk of using `void *`, otherwise compiler would catch this.

yosefe:
good catch @ofirfarjun7 ?
did you see any related test failure?

yosefe (approved):

ofirfarjun7:
Yes, but only after adding the connection establishment changes and replaying pending request of previous wireup EP.
I'm guessing that because we didn't do it before, new Wireup EP's doesn't have any pending requests so we don't enter the loop in "ucp_wireup_ep_pending_queue_purge" and the CB is not being called.

PR #8661, opened 2022-10-24
UCP/EP/WIREUP: Replay pending requests of wireup EP CM during connection establishment

ofirfarjun7:
## What
Replay pending requests of wireup EP CM during connection establishment.

## Why ?
To prevent potential ordering issues and wrong configuration.

## How ?
During connection establishment phases replay all pending requests including Wireup EP CM.

### DIFF
--- a/src/ucp/wireup/wireup.c
+++ b/src/ucp/wireup/wireup.c
@@ -454,7 +454,6 @@ unsigned ucp_wireup_eps_progress(void *arg)
     ucp_ep_h ucp_ep = arg;
     ucp_wireup_ep_t *wireup_ep;
     ucs_queue_head_t tmp_pending_queue;
-    int pending_count;
     ucp_lane_index_t lane;
 
     UCS_ASYNC_BLOCK(&ucp_ep->worker->async);
@@ -479,7 +478,7 @@ unsigned ucp_wireup_eps_progress(void *arg)
     /* Move wireup pending queue to temporary queue and remove references to
      * the wireup progress function
      */
-    ucs_queue_head_init(&tmp_pending_queue);
+    ucp_wireup_eps_pending_extract(ucp_ep, &tmp_pending_queue);
     for (lane = 0; lane < ucp_ep_num_lanes(ucp_ep); ++lane) {
         wireup_ep = ucp_wireup_ep(ucp_ep_get_lane(ucp_ep, lane));
         if (wireup_ep == NULL) {
@@ -491,9 +490,6 @@ unsigned ucp_wireup_eps_progress(void *arg)
 
         ucs_trace("ep %p: switching wireup_ep %p to ready state", ucp_ep,
                   wireup_ep);
-        pending_count = ucp_wireup_ep_pending_extract(wireup_ep,
-                                                      &tmp_pending_queue);
-        ucp_worker_flush_ops_count_add(ucp_ep->worker, -pending_count);
 
         /* Switch to real transport and destroy proxy endpoint (aux_ep as well) */
         ucp_proxy_ep_replace(&wireup_ep->super);
@@ -1366,7 +1362,6 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
     ucs_status_t status;
 
     *connect_lane_bitmap = UCS_MASK(new_key->num_lanes);
-    ucs_queue_head_init(replay_pending_queue);
 
     if (!ucp_ep_has_cm_lane(ep) ||
         (ep->cfg_index == UCP_WORKER_CFG_INDEX_NULL)) {
@@ -1419,8 +1414,6 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
                               old_key->lanes[reuse_lane].rsc_index,
                               ucp_ep_config(ep)->p2p_lanes &
                                       UCS_BIT(reuse_lane));
-        ucp_wireup_ep_pending_queue_purge(uct_ep, ucp_request_purge_enqueue_cb,
-                                          replay_pending_queue);
 
         /* reset the UCT EP from the previous WIREUP lane and destroy its WIREUP EP,
          * since it's not needed anymore in the new configuration, UCT EP will be
@@ -1440,8 +1433,10 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
                 ucs_assert(lane != ucp_ep_get_cm_lane(ep));
                 ucp_worker_discard_uct_ep(
                         ep, uct_ep, UCP_NULL_RESOURCE, UCT_FLUSH_FLAG_LOCAL,
-                        ucp_request_purge_enqueue_cb, replay_pending_queue,
-                        (ucp_send_nbx_callback_t)ucs_empty_function, NULL);
+                        (uct_pending_purge_callback_t)
+                                ucs_empty_function_do_assert_void,
+                        NULL, (ucp_send_nbx_callback_t)ucs_empty_function,
+                        NULL);
                 ucp_ep_set_lane(ep, lane, NULL);
             }
         } else if (uct_ep != NULL) {
@@ -1496,6 +1491,7 @@ ucs_status_t ucp_wireup_init_lanes(ucp_ep_h ep, unsigned ep_init_flags,
 
     ucp_ep_config_key_reset(&key);
     ucp_ep_config_key_set_err_mode(&key, ep_init_flags);
+    ucp_wireup_eps_pending_extract(ep, &replay_pending_queue);
 
     status = ucp_wireup_select_lanes(ep, ep_init_flags, tl_bitmap,
                                      remote_address, addr_indices, &key, 1);
@@ -1584,12 +1580,11 @@ ucs_status_t ucp_wireup_init_lanes(ucp_ep_h ep, unsigned ep_init_flags,
         ucp_ep_update_flags(ep, UCP_EP_FLAG_LOCAL_CONNECTED, 0);
     }
 
-    ucp_wireup_replay_pending_requests(ep, &replay_pending_queue);
-
     ucp_worker_keepalive_add_ep(ep);
     status = UCS_OK;
 
 out:
+    ucp_wireup_replay_pending_requests(ep, &replay_pending_queue);
     ucs_log_indent(-1);
     return status;
 }
--- a/src/ucp/wireup/wireup_cm.c
+++ b/src/ucp/wireup/wireup_cm.c
@@ -358,7 +358,7 @@ err:
 }
 
 static void
-ucp_wireup_cm_ep_cleanup(ucp_ep_t *ucp_ep, ucs_queue_head_t *queue)
+ucp_wireup_cm_ep_cleanup(ucp_ep_t *ucp_ep)
 {
     ucp_lane_index_t lane_idx;
 
@@ -367,15 +367,13 @@ ucp_wireup_cm_ep_cleanup(ucp_ep_t *ucp_ep, ucs_queue_head_t *queue)
             continue;
         }
 
-        /* During discarding transfer the pending queues content from the
-         * previously configured UCP EP to a temporary queue for further
-         * replaying */
-        ucp_worker_discard_uct_ep(ucp_ep, ucp_ep_get_lane(ucp_ep, lane_idx),
-                                  ucp_ep_get_rsc_index(ucp_ep, lane_idx),
-                                  UCT_FLUSH_FLAG_CANCEL,
-                                  ucp_request_purge_enqueue_cb, queue,
-                                  (ucp_send_nbx_callback_t)ucs_empty_function,
-                                  NULL);
+        /* During discarding, UCT EP's pending queues 
+         * are expected to be empty */
+        ucp_worker_discard_uct_ep(
+                ucp_ep, ucp_ep_get_lane(ucp_ep, lane_idx),
+                ucp_ep_get_rsc_index(ucp_ep, lane_idx), UCT_FLUSH_FLAG_CANCEL,
+                (uct_pending_purge_callback_t)ucs_empty_function_do_assert_void,
+                NULL, (ucp_send_nbx_callback_t)ucs_empty_function, NULL);
         ucp_ep_set_lane(ucp_ep, lane_idx, NULL);
     }
 }
@@ -450,17 +448,16 @@ static unsigned ucp_cm_client_uct_connect_progress(void *arg)
 
     UCS_ASYNC_BLOCK(&worker->async);
 
-    ucs_queue_head_init(&tmp_pending_queue);
-
     /* To silence Coverity warning */
     UCS_STATIC_ASSERT(ucs_static_array_size(ep_init_flags_prio) > 0);
 
+    ucp_wireup_eps_pending_extract(ep, &tmp_pending_queue);
     for (i = 0; i < ucs_static_array_size(ep_init_flags_prio); ++i) {
         ep_init_flags = ep_init_flags_prio[i];
 
         /* Cleanup the previously created UCP EP. The one that was created on
          * the previous call to this client's resolve_cb */
-        ucp_wireup_cm_ep_cleanup(ep, &tmp_pending_queue);
+        ucp_wireup_cm_ep_cleanup(ep);
 
         status = ucp_cm_ep_client_initial_config_get(ep, ep_init_flags,
                                         &cm_wireup_ep->cm_resolve_tl_bitmap,
@@ -483,9 +480,6 @@ static unsigned ucp_cm_client_uct_connect_progress(void *arg)
         if (status != UCS_OK) {
             goto err;
         }
-
-        /* Replay pending requests from the tmp_pending_queue */
-        ucp_wireup_replay_pending_requests(ep, &tmp_pending_queue);
 
         if (i < (ucs_static_array_size(ep_init_flags_prio) - 1)) {
             /* Can use another variant of endpoint's initialization flags for
@@ -523,6 +517,7 @@ static unsigned ucp_cm_client_uct_connect_progress(void *arg)
                                                 &params);
     ucs_free(priv_data);
 
+    ucp_wireup_replay_pending_requests(ep, &tmp_pending_queue);
     if (status != UCS_OK) {
         goto err;
     }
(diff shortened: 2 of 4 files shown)


gleon99:
@ofirfarjun7 notice `coverity-devel` errors. Skipped tests -> dead code.

src/ucp/wireup/wireup_cm.c
```diff
@@ -357,13 +357,33 @@ ucp_cm_ep_priv_data_pack(ucp_ep_h ep, const ucp_tl_bitmap_t *tl_bitmap,
     return status;
 }
 
+void ucp_wireup_cm_ep_pending_extract(ucp_ep_t *ucp_ep, ucs_queue_head_t *queue)
+{
+    const ucp_lane_index_t lane_idx = ucp_ep_get_cm_lane(ucp_ep);
```
gleon99, line 362: Can pass `lane` as an arg (it exists in the calling scope)

src/ucp/wireup/wireup_cm.c
```diff
@@ -357,13 +357,33 @@ ucp_cm_ep_priv_data_pack(ucp_ep_h ep, const ucp_tl_bitmap_t *tl_bitmap,
...
+    ucp_wireup_ep_t *wireup_ep;
+    int pending_count;
+
+    /* Old proto selection is not supporting protocol reset.
```
gleon99, line 366: `does not support`

src/ucp/wireup/wireup_cm.c
```diff
@@ -357,13 +357,33 @@ ucp_cm_ep_priv_data_pack(ucp_ep_h ep, const ucp_tl_bitmap_t *tl_bitmap,
...
+    ucp_wireup_ep_t *wireup_ep;
+    int pending_count;
+
+    /* Old proto selection is not supporting protocol reset. 
+     * With new EP config previous protocol selection might be invalid. */
+    if (!ucp_ep->worker->context->config.ext.proto_enable) {
+        return;
+    }
+
+    if (ucp_ep_get_lane(ucp_ep, lane_idx) != NULL) {
```
gleon99, line 372: 1. Can simplify: if NULL -> return.
2. Add comment for such case.

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -2655,40 +2659,52 @@ UCS_TEST_P(test_ucp_sockaddr_protocols, tag_rndv_unexp, "RNDV_THRESH=10k")
 
 UCS_TEST_P(test_ucp_sockaddr_protocols, stream_bcopy_4k_exp, "ZCOPY_THRESH=inf")
 {
-    UCS_TEST_SKIP_R("ProtoV2 is not implemented");
+    if (m_ucp_config->ctx.proto_enable) {
+        UCS_TEST_SKIP_R("ProtoV2 is not implemented");
+    }
```
gleon99, line 2664: Please add a blank line after `}`.

src/ucp/wireup/wireup_cm.c
ofirfarjun7: I thought to simplify the function definition and also make it more general, can be called when lane is not in the scope of calling scope (At the moment it's in the scope but maybe it will be used in the future in more places).

yosefe:

src/ucp/wireup/wireup_cm.c
```diff
@@ -357,13 +357,37 @@ ucp_cm_ep_priv_data_pack(ucp_ep_h ep, const ucp_tl_bitmap_t *tl_bitmap,
...
+
+    /* Old proto selection does not support protocol reset.
+     * With new EP config previous protocol selection might be invalid. */
+    if (!ucp_ep->worker->context->config.ext.proto_enable) {
+        return;
+    }
+
+    /* At early stages of the connection establishment wireup EP
+       can have only cm lanes with no transport lanes. */
+    if (ucp_ep_get_lane(ucp_ep, lane_idx) == NULL) {
```
gleon99, line 374: Can put the lane in a helper var (avoid calling `ucp_ep_get_lane` twice).

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -2653,34 +2661,58 @@ UCS_TEST_P(test_ucp_sockaddr_protocols, tag_rndv_unexp, "RNDV_THRESH=10k")
...
+
     test_stream_send_recv(4 * UCS_KBYTE, true);
 }
 
 UCS_TEST_P(test_ucp_sockaddr_protocols, stream_bcopy_4k_unexp,
            "ZCOPY_THRESH=inf")
 {
+    if (m_ucp_config->ctx.proto_enable) {
```
gleon99, line 2674: Maybe we can use `UCS_TEST_SKIP_COND_P` instead?

src/ucp/wireup/wireup_cm.c
```diff
@@ -357,13 +357,37 @@ ucp_cm_ep_priv_data_pack(ucp_ep_h ep, const ucp_tl_bitmap_t *tl_bitmap,
...
+    int pending_count;
+
+    /* Old proto selection does not support protocol reset.
+     * With new EP config previous protocol selection might be invalid. */
+    if (!ucp_ep->worker->context->config.ext.proto_enable) {
+        return;
+    }
+
+    /* At early stages of the connection establishment wireup EP
+       can have only cm lanes with no transport lanes. */
```
gleon99, line 373: `*`

src/ucp/wireup/wireup_cm.c
gleon99: IMO better to avoid a redundant call to `ucp_ep_get_cm_lane`, @evgeny-leksikov WDYT?

test/gtest/ucp/test_ucp_sockaddr.cc
ofirfarjun7: Inside `test_stream_send_recv` seems like the right place.

src/ucp/wireup/wireup_cm.c
ofirfarjun7: After another review and adding the helper var - I agree. I've change it, less calls and looks better.

src/ucp/wireup/wireup.c
```diff
@@ -1564,6 +1564,7 @@ ucs_status_t ucp_wireup_init_lanes(ucp_ep_h ep, unsigned ep_init_flags,
     /* establish connections on all underlying endpoints */
     for (lane = 0; lane < ucp_ep_num_lanes(ep); ++lane) {
         if (ucp_ep_get_cm_lane(ep) == lane) {
+            ucp_wireup_cm_ep_pending_extract(lane, ep, &replay_pending_queue);
```
evgeny-leksikov, line 1567: IMO need to extract `replay_pending_queue`  logic from `ucp_wireup_check_config_intersect` function, extract all pending requests from all lanes before it and replay them in the end.

src/ucp/wireup/wireup_cm.c
```diff
@@ -357,13 +357,39 @@ ucp_cm_ep_priv_data_pack(ucp_ep_h ep, const ucp_tl_bitmap_t *tl_bitmap,
...
+{
+    uct_ep_h uct_ep;
+    ucp_wireup_ep_t *wireup_ep;
+    int pending_count;
+
+    /* Old proto selection does not support protocol reset.
+     * With new EP config previous protocol selection might be invalid. */
+    if (!ucp_ep->worker->context->config.ext.proto_enable) {
+        return;
+    }
```
evgeny-leksikov, line 371: Can it be relevant for any wireup lane?

src/ucp/wireup/wireup_cm.c
```diff
@@ -357,13 +357,39 @@ ucp_cm_ep_priv_data_pack(ucp_ep_h ep, const ucp_tl_bitmap_t *tl_bitmap,
...
+
+    uct_ep = ucp_ep_get_lane(ucp_ep, lane_idx);
+    /* At early stages of the connection establishment wireup EP
```
ofirfarjun7, line 374: Need to double check that this is the reason or if it's a bug.

src/ucp/wireup/wireup.c
ofirfarjun7: Creating one function that pop all the elements from the pending queue seems like a good idea in theory. This way this logic will be in just one place instead of 3+ places in the code.
The downside is that it's will require changing/removing logic from existing code (`ucp_wireup_check_config_intersect`).
It's will require more time and it also can cause issues with "Old proto".

I can try to do it and see how it's works.
WDYT?

ofirfarjun7: @evgeny-leksikov i'm not sure it's possible to do the refactoring before ProtoV2 support for stream is added.
If we have pending requests with stream protocol in the pending queue a common function for extracting the requests will not work when ProtoV2 is enabled if i'm not mistaken.

Maybe it's better to postpone this, at least until ProtoV2 support all protocols?

evgeny-leksikov: maybe if it's hard to split paths inside of the function, we can implement different independent functions for protov1 and protov2?

src/ucp/wireup/wireup_cm.c
evgeny-leksikov, line 371: BTW according to function name `ucp_wireup_cm_ep_pending_extract` this if-statement should be done out side of the function since it does nothing if protov2 is enabled

src/ucp/wireup/wireup_cm.c
```diff
@@ -357,13 +357,39 @@ ucp_cm_ep_priv_data_pack(ucp_ep_h ep, const ucp_tl_bitmap_t *tl_bitmap,
...
+    if (!ucp_ep->worker->context->config.ext.proto_enable) {
+        return;
+    }
+
+    uct_ep = ucp_ep_get_lane(ucp_ep, lane_idx);
+    /* When creating EP with remote worker address
+     * EP is using transport lanes only, with no CM lane. */
+    if (uct_ep == NULL) {
+        return;
+    }
```
evgeny-leksikov, line 378: I think this test also should be done by a caller

src/ucp/wireup/wireup.c
ofirfarjun7: We can do that but it's will cause new code duplication and new branches in the code.
If the main goal of this suggested refactoring is to reduce code duplication i'm not sure if we want to go in this path.
Eventually we will be left only with the ProtoV2 functions, The questions is if we want to do it now in this PR or wait until Old proto is deprecated.

Let me know what you think.

gleon99: If it's over-complicating and the profit is unclear, maybe leave it as is for now? @evgeny-leksikov WDYT?

gleon99:
@ofirfarjun7  need to merge from master after https://github.com/openucx/ucx/pull/8692.

src/ucp/wireup/wireup_cm.c
```diff
@@ -370,12 +370,11 @@ ucp_wireup_cm_ep_cleanup(ucp_ep_t *ucp_ep, ucs_queue_head_t *queue)
...
-                                  ucp_ep_get_rsc_index(ucp_ep, lane_idx),
-                                  UCT_FLUSH_FLAG_CANCEL,
-                                  ucp_request_purge_enqueue_cb, queue,
-                                  (ucp_send_nbx_callback_t)ucs_empty_function,
-                                  NULL);
+        ucp_worker_discard_uct_ep(
+                ucp_ep, ucp_ep_get_lane(ucp_ep, lane_idx),
+                ucp_ep_get_rsc_index(ucp_ep, lane_idx), UCT_FLUSH_FLAG_CANCEL,
+                (uct_pending_purge_callback_t)ucs_empty_function, NULL,
+                (ucp_send_nbx_callback_t)ucs_empty_function, NULL);
```
evgeny-leksikov, line 377: do we expect that pending queues are empty? so can we pass `NULL` instead of `ucs_empty_function` in both places?

ofirfarjun7: In theory yes, you are right, but ucp_worker_discard_uct_ep has an assertion that checks if purge_cb != NULL

evgeny-leksikov: then could pass `ucs_empty_function_do_assert_void`, WDYT?

ofirfarjun7: Hi.
- I'm not sure why you prefer ucs_empty_function_do_assert_void over `ucs_empty_function`, can you please explain?
- We have more places with calls to `ucp_worker_discard_uct_ep` that we pass `ucs_empty_function` as purge_cb

evgeny-leksikov: > I'm not sure why you prefer `ucs_empty_function_do_assert_void` over `ucs_empty_function`, can you please explain?

to detect bugs earlier

> We have more places with calls to `ucp_worker_discard_uct_ep` that we pass `ucs_empty_function` as purge_cb, I think we need to be consistent.

We are discussing this particular case where expectation by design is to have queues empty. In other cases, probably, expectation is to do no-op on purging.

ofirfarjun7: I see. You are right.

evgeny-leksikov (dismissed):

src/ucp/wireup/wireup.c
```diff
@@ -479,25 +487,8 @@ unsigned ucp_wireup_eps_progress(void *arg)
...
-        ucs_trace("ep %p: switching wireup_ep %p to ready state", ucp_ep,
-                  wireup_ep);
-        pending_count = ucp_wireup_ep_pending_extract(wireup_ep,
-                                                      &tmp_pending_queue);
-        ucp_worker_flush_ops_count_add(ucp_ep->worker, -pending_count);
-
-        /* Switch to real transport and destroy proxy endpoint (aux_ep as well) */
-        ucp_proxy_ep_replace(&wireup_ep->super);
-    }
+    ucp_wireup_eps_pending_extract(ucp_ep, ucp_wireup_proxy_ep_replace,
```
yosefe, line 490: why need to pass ucp_wireup_proxy_ep_replace as a callback, can't we go over the lanes after calling ucp_wireup_eps_pending_extract and call ucp_wireup_proxy_ep_replace for each lane?
in other places the callback seems to be empty function anyway

src/ucp/wireup/wireup_cm.c
```diff
@@ -370,12 +370,11 @@ ucp_wireup_cm_ep_cleanup(ucp_ep_t *ucp_ep, ucs_queue_head_t *queue)
         /* During discarding transfer the pending queues content from the
          * previously configured UCP EP to a temporary queue for further
          * replaying */
-        ucp_worker_discard_uct_ep(ucp_ep, ucp_ep_get_lane(ucp_ep, lane_idx),
-                                  ucp_ep_get_rsc_index(ucp_ep, lane_idx),
-                                  UCT_FLUSH_FLAG_CANCEL,
-                                  ucp_request_purge_enqueue_cb, queue,
-                                  (ucp_send_nbx_callback_t)ucs_empty_function,
-                                  NULL);
+        ucp_worker_discard_uct_ep(
```
yosefe, line 373: 1. update the comment
2. remove "queue" parameter from ucp_wireup_cm_ep_cleanup - seems not needed anymore

src/ucp/wireup/wireup_cm.c
```diff
@@ -515,6 +512,7 @@ static unsigned ucp_cm_client_uct_connect_progress(void *arg)
         goto try_fallback;
     }
 
+    ucp_wireup_replay_pending_requests(ep, &tmp_pending_queue);
```
yosefe, line 515: shouldn't we do it after line 520?
@evgeny-leksikov WDYT?

src/ucp/wireup/wireup_ep.c
```diff
@@ -620,6 +620,38 @@ unsigned ucp_wireup_ep_pending_extract(ucp_wireup_ep_t *wireup_ep,
...
+        uct_ep = ucp_ep_get_lane(ucp_ep, lane_idx);
+        /* When creating EP with remote worker address
+         * EP is using transport lanes only, with no CM lane. */
+        if ((uct_ep == NULL) || (ucp_wireup_ep(uct_ep) == NULL)) {
+            continue;
+        }
+
+        wireup_ep     = ucp_wireup_ep(uct_ep);
+        pending_count = ucp_wireup_ep_pending_extract(wireup_ep, queue);
+        ucp_worker_flush_ops_count_add(ucp_ep->worker, -pending_count);
```
yosefe, line 650: we could sum up pending_count and call ucp_worker_flush_ops_count_add once after the loop

src/ucp/wireup/wireup.c
ofirfarjun7: We can do it. I thought to reduce iteration by using a CB.
In the rest of the calls it's indeed redundant at least for now.

src/ucp/wireup/wireup_cm.c
ofirfarjun7: 1. Will do
2. Will remove if it's indeed not needed.

src/ucp/wireup/wireup_ep.c
ofirfarjun7: Good point

src/ucp/wireup/wireup.c
yosefe: performance is not important here as much as making code simpler

src/ucp/wireup/wireup_cm.c
evgeny-leksikov: 👍 that would be better aligned with initial ep_create flow

src/ucp/wireup/wireup_ep.c
```diff
@@ -620,17 +620,14 @@ unsigned ucp_wireup_ep_pending_extract(ucp_wireup_ep_t *wireup_ep,
...
-void ucp_wireup_eps_pending_extract(ucp_ep_t *ucp_ep,
-                                    ucp_wireup_ep_extract_pending_cb extract_cb,
-                                    ucs_queue_head_t *queue)
+void ucp_wireup_eps_pending_extract(ucp_ep_t *ucp_ep, ucs_queue_head_t *queue)
 {
     ucp_lane_index_t lane_idx;
     ucp_wireup_ep_t *wireup_ep;
     uct_ep_h uct_ep;
-    int pending_count;
+    int pending_count = 0;
```
yosefe, line 628: initialized vars first

yosefe (dismissed):

yosefe:
@ofirfarjun7 pls squash

yosefe (approved):

PR #7902, opened 2022-02-04
UCP/API/INFO: Add API to define endpoint scope name for debug info

yosefe:
## Why
Show which kind of endpoint are we printing the information for - "intra-node", "inter-node", "self".
The scope name will be set from the upper layer, for example - an MPI library.

PR #8663, opened 2022-10-25
UCP/EP/PROTO: Print endpoint locality with protov2 select info

yosefe:
## Why
Improve protocol information output by printing endpoint locality information

Replaces #7902 

Example output:
```
+----------------+--------------------------------------------------------------------------------------------+
| mpi self cfg#0 | tagged message by ucp_tag_send*() from host memory                                         |
+----------------+-------------------------------------------+------------------------------------------------+
|        0..8184 | eager short                               | self/memory                                    |
|    8185..37142 | eager copy-in copy-out                    | self/memory                                    |
|     37143..inf | (?) rendezvous zero-copy read from remote | 68% on knem/memory and 32% on rc_mlx5/mlx5_0:1 |
+----------------+-------------------------------------------+------------------------------------------------+
+----------------------+--------------------------------------------------------------------------------------------+
| mpi intra-node cfg#1 | tagged message by ucp_tag_send*() from host memory                                         |
+----------------------+-------------------------------------------+------------------------------------------------+
|                0..92 | eager short                               | sysv/memory                                    |
|             93..8248 | eager copy-in copy-out                    | sysv/memory                                    |
|         8249..171045 | eager copy-in copy-out                    | sysv/memory                                    |
|          171046..inf | (?) rendezvous zero-copy read from remote | 68% on knem/memory and 32% on rc_mlx5/mlx5_0:1 |
+----------------------+-------------------------------------------+------------------------------------------------+
```

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -2043,7 +2043,72 @@ static unsigned ucp_wireup_default_path_index(unsigned path_index)
     return (path_index == UCP_WIREUP_PATH_INDEX_UNDEFINED) ? 0 : path_index;
 }
 
-static UCS_F_NOINLINE void
+static ucs_status_t ucp_wireup_select_set_locality_flags(
+        const ucp_wireup_select_params_t *select_params,
+        ucp_ep_config_key_t *key)
+{
+    ucp_worker_h worker = select_params->ep->worker;
+    ucp_worker_iface_t *wiface;
+    ucp_rsc_index_t rsc_index;
+    ucp_rsc_index_t iface_id;
+    ucp_address_entry_t *ae;
+    ucp_lane_index_t lane;
+    ucs_status_t status;
+    void *dev_addr;
+
+    if (select_params->address->uuid == worker->uuid) {
+        key->flags |= UCP_EP_CONFIG_KEY_FLAG_SELF;
+        return UCS_OK;
+    }
+
+    /* If selected at least one shared-memory transport, it's intra-node */
+    for (lane = 0; lane < key->num_lanes; ++lane) {
+        rsc_index = key->lanes[lane].rsc_index;
+        if ((rsc_index != UCP_NULL_RESOURCE) &&
+            (worker->context->tl_rscs[rsc_index].tl_rsc.dev_type ==
+             UCT_DEVICE_TYPE_SHM)) {
+            key->flags |= UCP_EP_CONFIG_KEY_FLAG_INTRA_NODE;
+            return UCS_OK;
+        }
+    }
+
+    /* If the local context matching at least one remote device address, it's
+       intra-node */
+    for (iface_id = 0; iface_id < worker->num_ifaces; ++iface_id) {
+        wiface = worker->ifaces[iface_id];
+        if (wiface->attr.device_addr_len == 0) {
+            continue;
+        }
+
+        dev_addr = ucs_malloc(wiface->attr.device_addr_len, "ucp_tmp_dev_addr");
+        if (dev_addr == NULL) {
+            ucs_error("failed to allocated device address of size %zu",
+                      wiface->attr.device_addr_len);
+            return UCS_ERR_NO_MEMORY;
+        }
+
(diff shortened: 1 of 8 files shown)


brminich (dismissed):

yosefe:
@brminich can you pls take a look?

src/ucp/wireup/select.c
```diff
@@ -2043,7 +2043,70 @@ static unsigned ucp_wireup_default_path_index(unsigned path_index)
...
+                key->flags |= UCP_EP_CONFIG_KEY_FLAG_INTRA_NODE;
+                ucs_free(dev_addr);
+                return UCS_OK;
+            }
+        }
+
+        ucs_free(dev_addr);
+    }
+
+    return UCS_OK;
```
brminich, line 2106:
```suggestion
        status = uct_iface_get_device_address(wiface->iface, dev_addr);
        if (status != UCS_OK) {
            goto out;
        }

        ucp_unpacked_address_for_each(ae, select_params->address) {
            if ((wiface->attr.device_addr_len == ae->dev_addr_len) &&
                !memcmp(ae->dev_addr, dev_addr, ae->dev_addr_len)) {
                key->flags |= UCP_EP_CONFIG_KEY_FLAG_INTRA_NODE;
                goto out;
            }
        }


    }
out:
    ucs_free(dev_addr);
    return status;
```

src/ucp/wireup/select.c
```diff
@@ -2174,6 +2237,9 @@ ucp_wireup_construct_lanes(const ucp_wireup_select_params_t *select_params,
     /* use AM lane first for eager AM transport to simplify processing single/middle
      * msg packets */
     key->am_bw_lanes[0] = key->am_lane;
+
+    /* Set locality flags */
```
brminich, line 2241: imo, this is excessive comment - the same info is provided by func name

src/ucp/wireup/select.c
```diff
@@ -2214,7 +2280,11 @@ ucp_wireup_select_lanes(ucp_ep_h ep, unsigned ep_init_flags,
     }
 
 out:
-    ucp_wireup_construct_lanes(&select_params, &select_ctx, addr_indices, key);
+    status = ucp_wireup_construct_lanes(&select_params, &select_ctx,
+                                        addr_indices, key);
+    if (status != UCS_OK) {
+        return status;
+    }
```
brminich, line 2287: is it unrelated bug fix?

src/uct/tcp/tcp_iface.c
```diff
@@ -128,6 +128,7 @@ static ucs_status_t uct_tcp_iface_get_device_address(uct_iface_h tl_iface,
 
     if (ucs_sockaddr_is_inaddr_loopback(saddr)) {
         dev_addr->flags |= UCT_TCP_DEVICE_ADDR_FLAG_LOOPBACK;
+        memset(pack_ptr, 0, sizeof(uct_iface_local_addr_ns_t));
```
brminich, line 131: is it some bug fix?

src/ucp/wireup/select.c
yosefe: so dev_addr can leak if the loop is run more than once.. would need separate err flow and good flow (since need to release dev_addr inside the loop)

src/ucp/wireup/select.c
yosefe: one of the gtests replaces get_device_address by a stub that returns error and expects wireup to fail if this stub is called at least once. so this change makes sure we fail whenever get_device_address fails from the context of the new locality code

src/uct/tcp/tcp_iface.c
yosefe: yes, it's an uninitialized memory issue found with valgrind, probably because we are calling get_device_address from another flow with this PR

src/ucp/wireup/select.c
brminich: yeah, missed it

brminich (approved):

brminich (approved):

Issue #8627, opened 2022-10-13
Problem with ucx master on large device-to-host transfers


edgargabriel:
### Describe the bug

We have been seeing an issue with UCX master recently with a testcase that used to work. This testcase performs a large data transfer from device to host memory (large =  4MB).  Using git bisect I identified this commit here as the starting point when we started to see the issue:

```
commit 6d2b6466ea5be8f59f710a1136852e6126f90f18
Date:   Fri Oct 7 14:28:41 2022 +0300

    UCP/CORE: 16 lanes support
```

Note, the output below is with the current HEAD (c5d5b0ca9b) 

### Steps to Reproduce
Output of ucx_info -v and ucx_info -d are in ucx_info.og (attached to this ticket)
[ucx_info.log](https://github.com/openucx/ucx/files/9780106/ucx_info.log)

### Setup and versions
- OS:  ubuntu 20.04
- AMD MI250x
- Config.log attached to this ticket:
[config.log.gz](https://github.com/openucx/ucx/files/9780109/config.log.gz)

### Additional inf
The actual error message comes down to:
```
[ZT-DH170-26:2426049:0:2426049] ucp_rkey.inl:48   Assertion `rkey_index < ((sizeof(rkey->md_map) <= 4) ? __builtin_popcount((uint32_t)(rkey->md_map)) :      __builtin_popcountl(rkey->md_map))' failed

/home/egabriel/ucx/build/../src/ucp/core/ucp_rkey.inl: [ ucp_rkey_get_tl_rkey() ]
      ...
       45         return UCT_INVALID_RKEY;
       46     }
       47
==>    48     ucs_assert(rkey_index < ucs_popcount(rkey->md_map));
       49     return rkey->tl_rkey[rkey_index].rkey.rkey;
       50 }
       51

==== backtrace (tid:2426049) ====
 0 0x00000000000ab998 ucp_rkey_get_tl_rkey()  /home/egabriel/ucx/build/../src/ucp/core/ucp_rkey.inl:48
 1 0x00000000000b85d3 ucp_rndv_send_start_put_pipeline()  /home/egabriel/ucx/build/src/ucp/../../../src/ucp/rndv/rndv.c:2184
 2 0x00000000000b9272 ucp_rndv_rtr_handler()  /home/egabriel/ucx/build/src/ucp/../../../src/ucp/rndv/rndv.c:2362
[...]

PR #8664, opened 2022-10-25
UCP/CORE: fix corner case for rkey_index

edgargabriel:
## What
fix a corner case for calculating the rkey_index in case the md_index is not part of the md_map.

## Why ?
Fixes issue https://github.com/openucx/ucx/issues/8627

### DIFF
--- a/src/ucp/rndv/rndv.c
+++ b/src/ucp/rndv/rndv.c
@@ -451,7 +451,8 @@ static uint8_t ucp_rndv_get_rkey_index(ucp_request_t *rndv_req, ucp_rkey_h rkey,
     uint8_t mem_type            = rndv_req->send.mem_type;
     uct_md_attr_v2_t *md_attr;
 
-    if ((md_index == UCP_NULL_RESOURCE) || (rkey == NULL)) {
+    if ((md_index == UCP_NULL_RESOURCE) || (rkey == NULL) ||
+        !(UCS_BIT(dst_md_index) & rkey->md_map)) {
         return UCP_NULL_RESOURCE;
     }
 


src/ucp/core/ucp_rkey.inl
```diff
@@ -45,7 +45,7 @@ ucp_rkey_get_tl_rkey(ucp_rkey_h rkey, ucp_md_index_t rkey_index)
         return UCT_INVALID_RKEY;
     }
 
-    ucs_assert(rkey_index < ucs_popcount(rkey->md_map));
+    ucs_assert(rkey_index < UCP_MAX_MDS);
```
brminich, line 48: looks incorrect, rkey_index is supposed to be `rkey->tl_rkey` array index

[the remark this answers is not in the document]
edgargabriel: you are right in that ucs_popcount(rkey->md_map) is the number of valid entries, but the valid index doesn't necessarily seem to be the first bits in the md_map

edgargabriel: rkey->md_map 0x10 dst_md_index 7 rkey_index 1

yosefe: then something is weird - it should not fail since 1 <= ucs_popcount(0x10)=1

yosefe: seems the issues is that dst_md_index is not in rkey->md_map

edgargabriel: I checkout out an earlier version of ucx that still seemed to work: what I found was that the logic in ucp_rndv_zcopy_get_lane() lead to an rkey_index of 255, and because of that the if statement in ucp_rkey_get_tl_rkey 
```
if (rkey_index == UCP_NULL_RESOURCE) {
        return UCT_INVALID_RKEY;
    }
```
was kicking in.

yosefe: can you try returning UCP_NULL_RESOURCE from ucp_rndv_get_rkey_index if (!(UCS_BIT(md_index) & rkey->md_map)) ?

edgargabriel: md_index or dst_md_index? I assume its the second one.  Tests are running right now, but looks good so far.

edgargabriel: @yosefe This solution seems to work, thank you! I have update the pr accordingly.

src/ucp/rndv/rndv.c
```diff
@@ -463,6 +463,10 @@ static uint8_t ucp_rndv_get_rkey_index(ucp_request_t *rndv_req, ucp_rkey_h rkey,
         }
     }
 
+    if (!(UCS_BIT(dst_md_index) & rkey->md_map)) {
```
yosefe, line 466: maybe add it to the check in line 454

edgargabriel: done, thanks!

yosefe (approved):

PR #8666, opened 2022-10-26
UCP/PROTO: RTR abort implementation

evgeny-leksikov:
## What
RTR abort implementation

## Why ?
```
Fatal: abort request 0x55a552290840 proto rndv/rtr/mtype, not implemented
```

### DIFF
--- a/src/ucp/rndv/rndv_ppln.c
+++ b/src/ucp/rndv/rndv_ppln.c
@@ -146,7 +146,7 @@ static void ucp_proto_rndv_ppln_query(const ucp_proto_query_params_t *params,
 }
 
 static void
-ucp_proto_rndv_ppln_frag_complete(ucp_request_t *freq, int send_ack,
+ucp_proto_rndv_ppln_frag_complete(ucp_request_t *freq, int send_ack, int abort,
                                   ucp_proto_complete_cb_t complete_func,
                                   const char *title)
 {
@@ -156,7 +156,7 @@ ucp_proto_rndv_ppln_frag_complete(ucp_request_t *freq, int send_ack,
         req->send.rndv.ppln.ack_data_size += freq->send.state.dt_iter.length;
     }
 
-    if (!ucp_proto_rndv_frag_complete(req, freq, title)) {
+    if (!ucp_proto_rndv_frag_complete(req, freq, title) && !abort) {
         return;
     }
 
@@ -164,7 +164,7 @@ ucp_proto_rndv_ppln_frag_complete(ucp_request_t *freq, int send_ack,
         ucp_proto_rndv_rkey_destroy(req);
     }
 
-    if (req->send.rndv.ppln.ack_data_size > 0) {
+    if ((req->send.rndv.ppln.ack_data_size > 0) && !abort) {
         ucp_proto_request_set_stage(req, UCP_PROTO_RNDV_PPLN_STAGE_ACK);
         ucp_request_send(req);
     } else {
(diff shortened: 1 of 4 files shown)


src/ucp/rndv/rndv_ppln.c
```diff
@@ -156,15 +156,15 @@ ucp_proto_rndv_ppln_frag_complete(ucp_request_t *freq, int send_ack,
         req->send.rndv.ppln.ack_data_size += freq->send.state.dt_iter.length;
     }
 
-    if (!ucp_proto_rndv_frag_complete(req, freq, title)) {
+    if (!ucp_proto_rndv_frag_complete(req, freq, title) && !abort) {
```
yosefe, line 159: is it possible we would complete more fragments of the same top-level request, but then their "super" would point to already released request?

evgeny-leksikov: I don't see the case for `rndv/rtr/mtype` but it's possible for other ppln or multi protocols like `rndv/get/mtype`, for example, which does not have `abort` implementation yet. In such case, I think `abort` flag should be set by completion status to handle outstanding fragments.

src/ucp/rndv/rndv_ppln.c
```diff
@@ -174,21 +174,22 @@ ucp_proto_rndv_ppln_frag_complete(ucp_request_t *freq, int send_ack,
...
     ucp_proto_rndv_ppln_frag_complete(
-            freq, 0, (ucp_proto_complete_cb_t)ucs_empty_function,
+            freq, 0, 0, (ucp_proto_complete_cb_t)ucs_empty_function,
             "ppln_recv_clean");
 }
 
 void ucp_proto_rndv_ppln_recv_frag_complete(ucp_request_t *freq, int send_ack)
 {
     ucp_proto_rndv_ppln_frag_complete(freq, send_ack,
+                                      UCS_STATUS_IS_ERR(freq->status),
```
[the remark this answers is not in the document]
evgeny-leksikov: I'm not sure if it would speed up somehow noticeable not inlined function, but OK it makes sense from perspective of design and existing code alignment.
BTW, CI failures happened because of `status` was not initialized. Is it intentionally or by mistake?

src/ucp/rndv/rndv_ppln.c
yosefe: I don't think it will help if we complete the request immediately; additional fragments may arrive and use an already-released request.
you are right it cannot happen for RTR/mtype because we search for request by its ID and now we'd remove it from the hash.

src/ucp/rndv/rndv_ppln.c
```diff
@@ -174,21 +174,22 @@ ucp_proto_rndv_ppln_frag_complete(ucp_request_t *freq, int send_ack,
...
 void ucp_proto_rndv_ppln_recv_frag_clean(ucp_request_t *freq)
 {
     ucp_proto_rndv_ppln_frag_complete(
-            freq, 0, (ucp_proto_complete_cb_t)ucs_empty_function,
+            freq, 0, 0, (ucp_proto_complete_cb_t)ucs_empty_function,
             "ppln_recv_clean");
 }
 
-void ucp_proto_rndv_ppln_recv_frag_complete(ucp_request_t *freq, int send_ack)
+void ucp_proto_rndv_ppln_recv_frag_complete(ucp_request_t *freq, int send_ack,
```
yosefe, line 189: maybe make this function inline?

evgeny-leksikov: It's used in `rndv_rtr.c` and `rndv_get.c`, also it's aligned with `ucp_proto_rndv_ppln_send_frag_complete`. Moving `ucp_proto_rndv_ppln_recv_frag_complete` to `rndv_ppln.inl` will require more refactoring since `ucp_proto_rndv_ppln_frag_complete` is defined as a `static` in `rndv_ppln.c`.
So far the functions look to me in their correct place.

src/ucp/rndv/rndv_ppln.c
evgeny-leksikov: So, do you agree with this implementation of RTR or request some change?

yosefe: yes, current rtr abort is ok because of request id hash.

src/ucp/rndv/rndv_ppln.c
yosefe: ok

yosefe (approved):

PR #8668, opened 2022-10-27
UCP/RNDV: Add 2-stage ppln protocol for protocols v2

brminich:
## What
Complete adding 2-stage ppln protocol for protocols v2 (#8526)

PR #8674, opened 2022-10-28
UCP/PROTOV2: Add exclude map to ucp_proto_common_init_params_t

brminich:
## What
Add exclude lane map to `ucp_proto_common_init_params_t`, so that `ucp_proto_common_find_lanes` will use it instead of just `0` (which is now passed in most of the places)

## Why ?
Will be used by 2-stage pipeline protocol, it will search for a lane with rkey_ptr capabilities, but not a real rkey_ptr lane (need to exclude it from the search)

ref #8668


yosefe:
Maybe add it to proto_common params ?

yosefe (commented):
pls update commit and PR title

brminich:
> pls update commit and PR title

done

brminich:
> > pls update commit and PR title
> 
> done

also will amend the commit title shortly

brminich:
@yosefe, pls review

yosefe (approved):

PR #8675, opened 2022-10-29
UCS/FUSE: Fallback from username to uid for VFS socket name

yosefe:


yosefe:
@brminich can you pls take a look?

src/ucs/vfs/fuse/vfs_fuse.c
```diff
@@ -421,7 +416,7 @@ static void *ucs_vfs_fuse_thread_func(void *arg)
                          un_addr.sun_path, ucs_status_string(status));
             }
         } else {
-            ucs_warn("failed to connect to vfs socket '%s': %m",
+            ucs_diag("failed to connect to vfs socket '%s': %m",
```
gleon99, line 419: Why downgrade this to `diag`? As `ucs_vfs_sock_get_address` always succeeds and returns a valid address now, this `else` branch represents, at least, a real warning?

yosefe: since VFS is not critical for UCX usage, in case there is another problem with connecting to VFS socket no need to create false alarm..

brminich (approved):

PR #8676, opened 2022-10-31
UCT/MLX5: CQE zipping user control per CQ direction

ivankochin:
## What
Allows controlling CQE zipping per CQ RX/TX direction.

## Why ?
This option would help to control the UCX configuration more accurately.

### DIFF
--- a/src/uct/ib/base/ib_iface.h
+++ b/src/uct/ib/base/ib_iface.h
@@ -72,7 +72,7 @@ typedef enum uct_ib_mtu {
 typedef enum {
     UCT_IB_DIR_RX,
     UCT_IB_DIR_TX,
-    UCT_IB_DIR_NUM
+    UCT_IB_DIR_LAST
 } uct_ib_dir_t;
 
 enum {
@@ -149,7 +149,7 @@ struct uct_ib_iface_config {
     } rx;
 
     /* Inline space to reserve in CQ */
-    size_t                  inl[UCT_IB_DIR_NUM];
+    size_t                  inl[UCT_IB_DIR_LAST];
 
     /* Change the address type */
     int                     addr_type;
@@ -204,16 +204,16 @@ enum {
 
 
 typedef struct uct_ib_iface_init_attr {
-    unsigned    rx_priv_len;            /* Length of transport private data to reserve */
-    unsigned    rx_hdr_len;             /* Length of transport network header */
-    unsigned    cq_len[UCT_IB_DIR_NUM]; /* CQ length */
-    size_t      seg_size;               /* Transport segment size */
-    unsigned    fc_req_size;            /* Flow control request size */
-    int         qp_type;                /* IB QP type */
-    int         flags;                  /* Various flags (see enum) */
+    unsigned    rx_priv_len;             /* Length of transport private data to reserve */
+    unsigned    rx_hdr_len;              /* Length of transport network header */
+    unsigned    cq_len[UCT_IB_DIR_LAST]; /* CQ length */
+    size_t      seg_size;                /* Transport segment size */
+    unsigned    fc_req_size;             /* Flow control request size */
+    int         qp_type;                 /* IB QP type */
+    int         flags;                   /* Various flags (see enum) */
     /* The maximum number of outstanding RDMA Read/Atomic operations per QP */
     unsigned    max_rd_atomic;
-    uint8_t     cqe_zip_sizes;
+    uint8_t     cqe_zip_sizes[UCT_IB_DIR_LAST];
 } uct_ib_iface_init_attr_t;
 
 
@@ -231,7 +231,7 @@ typedef struct uct_ib_qp_attr {
     struct ibv_srq              *srq;
     uint32_t                    srq_num;
     unsigned                    sq_sig_all;
-    unsigned                    max_inl_cqe[UCT_IB_DIR_NUM];
+    unsigned                    max_inl_cqe[UCT_IB_DIR_LAST];
     uct_ib_qp_init_attr_t       ibv;
 } uct_ib_qp_attr_t;
 
@@ -273,7 +273,7 @@ struct uct_ib_iface_ops {
 struct uct_ib_iface {
     uct_base_iface_t          super;
 
-    struct ibv_cq             *cq[UCT_IB_DIR_NUM];
+    struct ibv_cq             *cq[UCT_IB_DIR_LAST];
     struct ibv_comp_channel   *comp_channel;
     uct_recv_desc_t           release_desc;
 
@@ -295,7 +295,7 @@ struct uct_ib_iface {
         unsigned              tx_max_poll;
         unsigned              seg_size;
         unsigned              roce_path_factor;
-        uint8_t               max_inl_cqe[UCT_IB_DIR_NUM];
+        uint8_t               max_inl_cqe[UCT_IB_DIR_LAST];
         uint8_t               port_num;
         uint8_t               sl;
         uint8_t               traffic_class;
--- a/src/uct/ib/mlx5/ib_mlx5.c
+++ b/src/uct/ib/mlx5/ib_mlx5.c
@@ -68,10 +68,17 @@ ucs_config_field_t uct_ib_mlx5_iface_config_table[] = {
      "+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++\n",
      ucs_offsetof(uct_ib_mlx5_iface_config_t, ar_enable), UCS_CONFIG_TYPE_TERNARY_AUTO},
 
-    {"CQE_ZIPPING_ENABLE", "no",
-     "Enable CQE zipping feature. CQE zipping reduces PCI utilization by\n"
+    {"TX_CQE_ZIP_ENABLE", "no",
+     "Enable CQE zipping feature for sender side. CQE zipping reduces PCI utilization by\n"
      "merging several similar CQEs to a single CQE written by the device.",
-     ucs_offsetof(uct_ib_mlx5_iface_config_t, cqe_zipping_enable), UCS_CONFIG_TYPE_BOOL},
+     ucs_offsetof(uct_ib_mlx5_iface_config_t, cqe_zip_enable[UCT_IB_DIR_TX]),
+     UCS_CONFIG_TYPE_BOOL},
+
+    {"RX_CQE_ZIP_ENABLE", "no",
+     "Enable CQE zipping feature for receiver side. CQE zipping reduces PCI utilization by\n"
+     "merging several similar CQEs to a single CQE written by the device.",
+     ucs_offsetof(uct_ib_mlx5_iface_config_t, cqe_zip_enable[UCT_IB_DIR_RX]),
+     UCS_CONFIG_TYPE_BOOL},
 
     {NULL}
 };
@@ -80,16 +87,19 @@ void uct_ib_mlx5_parse_cqe_zipping(uct_ib_mlx5_md_t *md,
                                    const uct_ib_mlx5_iface_config_t *mlx5_config,
                                    uct_ib_iface_init_attr_t *init_attr)
 {
-    if (!mlx5_config->cqe_zipping_enable) {
-        return;
-    }
+    uct_ib_dir_t dir;
 
-    if (md->flags & UCT_IB_MLX5_MD_FLAG_CQE64_ZIP) {
-        init_attr->cqe_zip_sizes |= 64;
-    }
+    for (dir = 0; dir < UCT_IB_DIR_LAST; ++dir) {
+        if (!mlx5_config->cqe_zip_enable[dir]) {
+            continue;
+        }
 
-    if (md->flags & UCT_IB_MLX5_MD_FLAG_CQE128_ZIP) {
-        init_attr->cqe_zip_sizes |= 128;
+        if (md->flags & UCT_IB_MLX5_MD_FLAG_CQE64_ZIP) {
+            init_attr->cqe_zip_sizes[dir] |= 64;
+        }
+        if (md->flags & UCT_IB_MLX5_MD_FLAG_CQE128_ZIP) {
+            init_attr->cqe_zip_sizes[dir] |= 128;
+        }
     }
 }
 
--- a/src/uct/ib/mlx5/ib_mlx5.h
+++ b/src/uct/ib/mlx5/ib_mlx5.h
@@ -322,7 +322,7 @@ typedef struct uct_ib_mlx5_iface_config {
 #endif
     uct_ib_mlx5_mmio_mode_t  mmio_mode;
     ucs_ternary_auto_value_t ar_enable;
-    int                      cqe_zipping_enable;
+    int                      cqe_zip_enable[UCT_IB_DIR_LAST];
 } uct_ib_mlx5_iface_config_t;
 
 
--- a/test/gtest/uct/ib/test_cqe_zipping.cc
+++ b/test/gtest/uct/ib/test_cqe_zipping.cc
@@ -68,7 +68,8 @@ public:
     virtual void init()
     {
         stats_activate();
-        modify_config("IB_CQE_ZIPPING_ENABLE", "y");
+        modify_config("IB_TX_CQE_ZIP_ENABLE", "yes");
+        modify_config("IB_RX_CQE_ZIP_ENABLE", "yes");
 
         test_uct_ib_with_specific_port::init();
         test_uct_ib::init();
--- a/src/uct/ib/mlx5/dv/ib_mlx5_dv.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5_dv.c
@@ -630,7 +630,7 @@ uct_ib_mlx5_devx_create_cq(uct_ib_iface_t *iface, uct_ib_dir_t dir,
     UCT_IB_MLX5DV_SET(cqc, cqctx, log_cq_size, log_cq_size);
     UCT_IB_MLX5DV_SET(cqc, cqctx, cqe_sz, (cqe_size == 128) ? 1 : 0);
 
-    if (init_attr->cqe_zip_sizes & cqe_size) {
+    if (init_attr->cqe_zip_sizes[dir] & cqe_size) {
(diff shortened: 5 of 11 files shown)


ivankochin:
With regard to performed measurements patch shouldn't affect the performance if CQE zipping disabled.

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -80,16 +87,28 @@ void uct_ib_mlx5_parse_cqe_zipping(uct_ib_mlx5_md_t *md,
                                    const uct_ib_mlx5_iface_config_t *mlx5_config,
                                    uct_ib_iface_init_attr_t *init_attr)
 {
-    if (!mlx5_config->cqe_zipping_enable) {
-        return;
-    }
+    uct_ib_dir_t dirs[UCT_IB_DIR_NUM] = {};
```
gleon99, line 90: Why is it of type `uct_ib_dir_t`?
Seems that the type of each array member is "bool"

gleon99: WDYT of having a bitmap instead of an array here,
then using something like `UCS_BIT(dir_idx)` in the `for` loop below?

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -313,16 +313,22 @@ typedef enum {
 } uct_ib_mlx5_mmio_mode_t;
 
 
+typedef enum {
+    UCT_IB_MLX5_CQE_ZIPPING_RX,
+    UCT_IB_MLX5_CQE_ZIPPING_TX,
+    UCT_IB_MLX5_CQE_ZIPPING_LAST
+} uct_ib_mlx5_cqe_zipping_mode_t;
```
gleon99, line 320: Maybe we can remove it at all and use `uct_ib_dir_t` instead?
Then `uct_ib_mlx5_parse_cqe_zipping` will be simpler

src/uct/ib/rc/accel/rc_mlx5_iface.c
```diff
@@ -180,39 +180,34 @@ static UCS_F_ALWAYS_INLINE unsigned
 uct_rc_mlx5_iface_progress(void *arg, int flags)
 {
     uct_rc_mlx5_iface_common_t *iface = arg;
+    int zip_flag;
     unsigned count;
 
-    count = uct_rc_mlx5_iface_common_poll_rx(iface, flags);
+    zip_flag = iface->cq[UCT_IB_DIR_RX].zip ? UCT_IB_MLX5_POLL_FLAG_CQE_ZIP : 0;
```
gleon99, line 186: See chat message, regarding perf consideration.

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -313,22 +313,16 @@ typedef enum {
...
+        size_t               seg_len;
+        unsigned             count;
     } dm;
 #endif
-    uct_ib_mlx5_mmio_mode_t        mmio_mode;
-    ucs_ternary_auto_value_t       ar_enable;
-    uct_ib_mlx5_cqe_zipping_mode_t cqe_zipping_enable;
+    uct_ib_mlx5_mmio_mode_t  mmio_mode;
+    ucs_ternary_auto_value_t ar_enable;
+    int                      cqe_zipping_dirs_bitmap;
```
gleon99, line 325: `unsigned`?

ivankochin: It has `int` type in the master branch so what is the reason for change?

gleon99: In `master` it's `cqe_zipping_enable` flag.
Bitmaps from config are usually **unsigned** (see examples in other places).

ivankochin:
```suggestion
    unsigned                 cqe_zipping_dirs_bitmap;
```

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -322,7 +322,7 @@ typedef struct uct_ib_mlx5_iface_config {
 #endif
     uct_ib_mlx5_mmio_mode_t  mmio_mode;
     ucs_ternary_auto_value_t ar_enable;
-    int                      cqe_zipping_enable;
+    unsigned                 cqe_zipping_dirs_bitmap;
```
gleon99, line 325: Minor: usually adding the var type to its name is redundant and "bad practice".
e.g we don't usually name things `..._arr`, `..._struct` etc.
Let's remove the `_bitmap`.

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -31,6 +31,12 @@ static const char *uct_ib_mlx5_mmio_modes[] = {
     [UCT_IB_MLX5_MMIO_MODE_LAST]       = NULL
 };
 
+static const char *uct_ib_mlx5_cqe_zipping_modes[] = {
+    [UCT_IB_DIR_RX]  = "rx",
+    [UCT_IB_DIR_TX]  = "tx",
+    [UCT_IB_DIR_NUM] = NULL
```
gleon99, line 37: `NUM` -> `LAST` (consistency with similar code).

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -68,10 +74,11 @@ ucs_config_field_t uct_ib_mlx5_iface_config_table[] = {
      "+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++\n",
      ucs_offsetof(uct_ib_mlx5_iface_config_t, ar_enable), UCS_CONFIG_TYPE_TERNARY_AUTO},
 
-    {"CQE_ZIPPING_ENABLE", "no",
+    {"CQE_ZIPPING_ENABLE", "",
```
gleon99, line 77: 1. Maybe rename the param, as it is not a "flag" anymore. Consider how other `UCS_CONFIG_TYPE_BITMAP` params are named.
2. In any case, please make the explanation here more explicit, so that the user know exactly what are the possible values.

ivankochin: 1. What do you think about reverting the syntax? I mean introduce two `CQ_[TX/RX]_FEATURES` controllers. At this moment, the only possible values for them are "empty" or "cqe zipping". But if in the future we add controllable more features for CQ, it would be pretty extensible.

gleon99: Then it can be something like `CQ_[TX|RX]_MODE`. Not bad. @yosefe  WDYT?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -279,36 +279,34 @@ static UCS_F_ALWAYS_INLINE unsigned
...
 {
     uct_dc_mlx5_iface_t *iface = arg;
+    const int zip_rx           = iface->super.cq[UCT_IB_DIR_RX].zip;
+    const int zip_tx           = iface->super.cq[UCT_IB_DIR_TX].zip;
+    int zip_flag;
     unsigned count;
 
-    count = uct_rc_mlx5_iface_common_poll_rx(&iface->super, flags);
+    zip_flag = zip_rx ? UCT_IB_MLX5_POLL_FLAG_CQE_ZIP : 0;
+    count    = uct_rc_mlx5_iface_common_poll_rx(&iface->super, flags | zip_flag);
```
yosefe, line 288: seems it beats the purpose of resolving poll flag during compile time

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -31,6 +31,12 @@ static const char *uct_ib_mlx5_mmio_modes[] = {
     [UCT_IB_MLX5_MMIO_MODE_LAST]       = NULL
 };
 
+static const char *uct_ib_mlx5_cqe_zipping_modes[] = {
```
yosefe, line 34: uct_ib_mlx5_cq_dir_names

src/uct/ib/dc/dc_mlx5.c
ivankochin: Yes, but I checked the performance and there is no degradation.

We can separate CQE_ZIP flag for TX and RX directions, but then we will have the exponential growth of `iface_progress` methods with different `poll_flags` configurations.

yosefe: > Yes, but I checked the performance and there is no degradation.

How did you check?

ivankochin: Compared two OSU benchmarks execution results, one with the UCX built with this patch and other built without it.

src/uct/ib/mlx5/ib_mlx5.c
yosefe: IMO, `CQ_[TX/RX]_FEATURES` is too generic.
We can have `[TX/RX]_CQ_ZIPPING_ENABLE`

src/uct/ib/dc/dc_mlx5.c
gleon99: @ivankochin as discussed, please separate the funcs to "with" / "without" CQE zipping, aggregating all the CQE zipping "variants" in one place.

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -68,10 +68,17 @@ ucs_config_field_t uct_ib_mlx5_iface_config_table[] = {
      "+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++\n",
      ucs_offsetof(uct_ib_mlx5_iface_config_t, ar_enable), UCS_CONFIG_TYPE_TERNARY_AUTO},
 
-    {"CQE_ZIPPING_ENABLE", "no",
-     "Enable CQE zipping feature. CQE zipping reduces PCI utilization by\n"
+    {"TX_CQE_ZIPPING_ENABLE", "no",
```
yosefe, line 71: maybe if we're already renaming, use TX_CQE_ZIP_ENABLE - to make it shorter?

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -322,7 +322,7 @@ typedef struct uct_ib_mlx5_iface_config {
 #endif
     uct_ib_mlx5_mmio_mode_t  mmio_mode;
     ucs_ternary_auto_value_t ar_enable;
-    int                      cqe_zipping_enable;
+    unsigned                 cqe_zipping_dirs[UCT_IB_DIR_LAST];
```
yosefe, line 325: 1. should be int
2. keep using the name cqe_zipping_enable

src/uct/ib/rc/accel/rc_mlx5.inl
```diff
@@ -24,6 +24,19 @@ uct_rc_mlx5_ep_fence_put(uct_rc_mlx5_iface_common_t *iface, uct_ib_mlx5_txwq_t *
...
 }
 
+/**
+ * Pass the CQE zipping flag only if both
+ * runtime and compile time flags are set
+ */
+static UCS_F_ALWAYS_INLINE int
+uct_rc_mlx5_check_zip_flag(uct_rc_mlx5_iface_common_t *iface, int dir,
+                           int flags) {
+    if ((flags & UCT_IB_MLX5_POLL_FLAG_CQE_ZIP) && !iface->cq[dir].zip) {
```
yosefe, line 34: why need to check iface->cq[dir].zip?
we didn't do it before this PR

ivankochin: Because of what we discussed previously. Before this patch we had only two variants of `progress` function: with and without zipping. And this options could be simply expressed by passing compile-time poll flags only.

Now we have four options:
- no zipping
- tx zipping
- rx zipping
- tx and rx zipping

Expression all these options by compile-time `flags` only will cause exponential growth of different `progress` functions. So, as I remember, we decided to implement mixed approach: existing of zipping still expressed by compile time flags, but rx/tx zipping presence checked using per CQ values.

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -279,34 +279,39 @@ static UCS_F_ALWAYS_INLINE unsigned
...
     uct_dc_mlx5_iface_t *iface = arg;
-    const int zip_rx           = iface->super.cq[UCT_IB_DIR_RX].zip;
-    const int zip_tx           = iface->super.cq[UCT_IB_DIR_TX].zip;
-    int zip_flag;
+    int rx_flags, tx_flags;
     unsigned count;
 
-    zip_flag = zip_rx ? UCT_IB_MLX5_POLL_FLAG_CQE_ZIP : 0;
-    count    = uct_rc_mlx5_iface_common_poll_rx(&iface->super, flags | zip_flag);
+    rx_flags = uct_rc_mlx5_check_zip_flag(&iface->super, UCT_IB_DIR_RX, flags);
```
gleon99, line 285: Align `=`

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -684,7 +689,12 @@ uct_dc_mlx5_init_rx(uct_rc_iface_t *rc_iface,
     }
 
     if (iface->super.config.srq_topo == UCT_RC_MLX5_SRQ_TOPO_LIST) {
-        iface->super.super.progress = uct_dc_mlx5_iface_progress_ll;
+        if (iface->super.cq[UCT_IB_DIR_RX].zip ||
```
gleon99, line 692: So we check the same flags twice? First here, then in `uct_rc_mlx5_check_zip_flag`?

src/uct/ib/rc/accel/rc_mlx5_iface.c
```diff
@@ -180,34 +180,42 @@ static UCS_F_ALWAYS_INLINE unsigned
 uct_rc_mlx5_iface_progress(void *arg, int flags)
 {
     uct_rc_mlx5_iface_common_t *iface = arg;
-    int zip_flag;
+    int rx_flags, tx_flags;
     unsigned count;
 
-    zip_flag = iface->cq[UCT_IB_DIR_RX].zip ? UCT_IB_MLX5_POLL_FLAG_CQE_ZIP : 0;
-    count    = uct_rc_mlx5_iface_common_poll_rx(iface, flags | zip_flag);
+    rx_flags = uct_rc_mlx5_check_zip_flag(iface, UCT_IB_DIR_RX, flags);
```
gleon99, line 186: Align `=`

src/uct/ib/dc/dc_mlx5.c
ivankochin: If I understand you correctly - then the answer is here https://github.com/openucx/ucx/pull/8676#discussion_r1029999639

src/uct/ib/base/ib_iface.h
```diff
@@ -206,14 +206,14 @@ enum {
 typedef struct uct_ib_iface_init_attr {
     unsigned    rx_priv_len;            /* Length of transport private data to reserve */
     unsigned    rx_hdr_len;             /* Length of transport network header */
-    unsigned    cq_len[UCT_IB_DIR_NUM]; /* CQ length */
+    unsigned    cq_len[UCT_IB_DIR_LAST]; /* CQ length */
```
gleon99, line 209: Fix indentation

src/uct/ib/rc/accel/rc_mlx5.inl
yosefe: what if we just pass poll_flags as-is to both rx poll and tx poll?

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -322,7 +322,7 @@ typedef struct uct_ib_mlx5_iface_config {
 #endif
     uct_ib_mlx5_mmio_mode_t  mmio_mode;
     ucs_ternary_auto_value_t ar_enable;
-    unsigned                 cqe_zipping_dirs[UCT_IB_DIR_LAST];
+    int                      cqe_zipping_enable[UCT_IB_DIR_LAST];
```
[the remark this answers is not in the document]
ivankochin: Have you seen the latest version? I already [renamed it](https://github.com/openucx/ucx/pull/8676/commits/6561000d9d76235317513d3b0ac56a3f04c5ddab#diff-196b574d0c411393a8fa68c6ee6d152b3a318a6453d1f902c29a5d240bb468dfR325).

src/uct/ib/rc/accel/rc_mlx5.inl
ivankochin: Seems like we can, but want to recheck the perf.

src/uct/ib/rc/accel/rc_mlx5_iface.c
```diff
@@ -182,7 +182,7 @@ uct_rc_mlx5_iface_progress(void *arg, int flags)
     uct_rc_mlx5_iface_common_t *iface = arg;
     unsigned count;
 
-    count = uct_rc_mlx5_iface_common_poll_rx(iface, flags);
+    count    = uct_rc_mlx5_iface_common_poll_rx(iface, flags);
```
ivankochin, line 185:
```suggestion
    count = uct_rc_mlx5_iface_common_poll_rx(iface, flags);
```

yosefe (approved):

test/gtest/uct/ib/test_cqe_zipping.cc
```diff
@@ -68,7 +68,8 @@ class test_cqe_zipping : public test_uct_ib_with_specific_port {
     virtual void init()
     {
         stats_activate();
-        modify_config("IB_CQE_ZIPPING_ENABLE", "y");
+        modify_config("IB_TX_CQE_ZIP_ENABLE", "yes");
```
gleon99, line 71: @ivankochin 
Can we feasibly test the RX / TX zipping separately and independently?

yosefe (approved):

test/gtest/uct/ib/test_cqe_zipping.cc
ivankochin: I think we can, but let me please do it as a separate patch.

PR #8677, opened 2022-10-31
UCT/IB/MLX5: Set counter_set_id during modify 2INIT instead of create

yosefe:
## Why
Fixing #8601 - counter_set_id field was set in a wrong place for RC / DCI QP
Fix NVIDIA internal issue [3139906](https://redmine.mellanox.com/issues/3139906)


yosefe:

Artemy-Mellanox (approved):

PR #8678, opened 2022-11-01
UCP/PROTOV2: Check rkey_ptr cap on component

brminich:
## What
- Check rkey_ptr capability on the component if protocol sets `UCP_PROTO_COMMON_INIT_FLAG_RKEY_PTR` flag
- Allow to search any type of line (i.e. `UCP_LANE_TYPE_LAST`)

## Why ?
Will be used by 2-stage pipeline protocol to find any suitable lane with rkey_ptr property on its component

ref #8668

### DIFF
--- a/src/ucp/proto/proto_common.c
+++ b/src/ucp/proto/proto_common.c
@@ -386,6 +386,7 @@ static ucp_lane_index_t ucp_proto_common_find_lanes_internal(
     const uct_iface_attr_t *iface_attr;
     ucp_lane_index_t lane, num_lanes;
     const uct_md_attr_v2_t *md_attr;
+    const uct_component_attr_t *cmpt_attr;
     ucp_rsc_index_t rsc_index;
     ucp_md_index_t md_index;
     ucp_lane_map_t lane_map;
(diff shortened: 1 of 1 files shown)


src/ucp/proto/proto_common.c
```diff
@@ -442,7 +443,8 @@ static ucp_lane_index_t ucp_proto_common_find_lanes_internal(
                  UCT_TL_RESOURCE_DESC_ARG(&context->tl_rscs[rsc_index].tl_rsc));
 
         /* Check if lane type matches */
-        if (!(ep_config_key->lanes[lane].lane_types & UCS_BIT(lane_type))) {
+        if (!(ep_config_key->lanes[lane].lane_types & UCS_BIT(lane_type)) &&
+            (lane_type != UCP_LANE_TYPE_LAST)) {
```
yosefe, line 447: IMO better check !=UCP_LANE_TYPE_LAST before feeding it into UCS_BIT

src/ucp/proto/proto_common.c
```diff
@@ -490,6 +492,15 @@ static ucp_lane_index_t ucp_proto_common_find_lanes_internal(
                 goto out;
             }
 
+            if (flags & UCP_PROTO_COMMON_INIT_FLAG_RKEY_PTR) {
+                cmpt_attr = ucp_cmpt_attr_by_md_index(context, md_index);
```
yosefe, line 496: IMO can move this to line 462

yosefe: i meant just the line `cmpt_attr = ucp_cmpt_attr_by_md_index(context, md_index);`
and then use single if with &&

brminich: ok, done. Left the check outside `UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS` branch, because it does check anything in rkey_config

yosefe (approved):

PR #8683, opened 2022-11-03
UCP/RKEY: Remove double dt cleanup from rndv rkey proto

ivankochin:
## What
Now `dt_cleanup` can be called twice during `rndv/rkey` proto execution. This patch removes the redundant `dt_cleanup` call from `ucp_proto_rndv_progress_rkey_ptr` while other one from `ucp_proto_rndv_ats_progress` is still on the place.

Tested with xpmem manually.


gleon99:
Please add the place of the other call to the PR desc (`ucp_proto_rndv_ats_complete`...)

yosefe (approved):

yosefe:
@evgeny-leksikov can you pls take a look as well?

evgeny-leksikov (approved):

PR #8689, opened 2022-11-09
UCT/IB: Fix exported memh packing

brminich:
## What
Create umem with default page size only

## Why ?
A temporary fix for packing exported mkeys. Fixes the following problem:
```
mlx5dv_devx_obj_create(MKEY) failed on mlx5_bond_1, syndrome 0x317175: Remote I/O error
```
where 0x317175 is _page offset of the starting point in the umem must match the page offset indicated by the mkey.start_address_

Also driver bug is created - [Nvidia internal link](https://redmine.mellanox.com/issues/3260653)

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -1402,7 +1402,7 @@ uct_ib_mlx5_devx_reg_exported_key(uct_ib_md_t *ib_md, uct_ib_mem_t *ib_memh)
     struct mlx5dv_devx_umem_in umem_in;
     ucs_status_t status;
     void *access_key;
-    void *address;
+    void *address, *aligned_address;
     size_t length;
     void *mkc;
     int ret;
@@ -1414,7 +1414,8 @@ uct_ib_mlx5_devx_reg_exported_key(uct_ib_md_t *ib_md, uct_ib_mem_t *ib_memh)
     umem_in.addr        = address;
     umem_in.size        = length;
     umem_in.access      = UCT_IB_MLX5_MD_UMEM_ACCESS;
-    umem_in.pgsz_bitmap = UINT64_MAX & ~UCS_MASK(UCT_IB_MLX5_PAGE_SHIFT);
+    aligned_address     = ucs_align_down_pow2_ptr(address, ucs_get_page_size());
+    umem_in.pgsz_bitmap = UCS_MASK(ucs_ffs64((uint64_t)aligned_address) + 1);
     umem_in.comp_mask   = 0;
 
     ucs_assertv(memh->umem == NULL, "memh %p umem %p", memh, memh->umem);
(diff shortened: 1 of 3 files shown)


brminich:
@Artemy-Mellanox

test/gtest/uct/test_md.cc
```diff
@@ -817,14 +817,14 @@ UCS_TEST_SKIP_COND_P(test_md, dereg_bad_arg,
...
-    static const unsigned md_flags = UCT_MD_MEM_ACCESS_REMOTE_PUT |
-                                     UCT_MD_MEM_ACCESS_REMOTE_GET;
+    size_t size = ucs::rand() % UCS_MBYTE;
     uct_mem_h export_memh;
+    void *address;
     ucs_status_t status;
 
-    std::vector<uint8_t> mem_buffer(size);
-    status = reg_mem(md_flags, mem_buffer.data(), size, &export_memh);
+    status = ucs_mmap_alloc(&size, &address, 0, "test_md_exp_mkey");
```
yosefe, line 825: print the size with UCS_TEST_MESSAGE so if it fails we could repro

test/gtest/uct/test_md.cc
```diff
@@ -817,14 +817,14 @@ UCS_TEST_SKIP_COND_P(test_md, dereg_bad_arg,
...
+    ASSERT_UCS_OK(status);
+    status = reg_mem(UCT_MD_MEM_ACCESS_ALL, address, size, &export_memh);
```
yosefe, line 827: space line before

test/gtest/uct/test_md.cc
brminich: ok.
actually I did it initially, but it does not help) It fails sporadically with the same message size

yosefe (approved):

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1414,7 +1414,9 @@ uct_ib_mlx5_devx_reg_exported_key(uct_ib_md_t *ib_md, uct_ib_mem_t *ib_memh)
     umem_in.addr        = address;
     umem_in.size        = length;
     umem_in.access      = UCT_IB_MLX5_MD_UMEM_ACCESS;
-    umem_in.pgsz_bitmap = UINT64_MAX & ~UCS_MASK(UCT_IB_MLX5_PAGE_SHIFT);
+    /* TODO: Allow other page sizes as well when umem offset handling is
+     * fixed in the driver. */
+    umem_in.pgsz_bitmap = ucs_get_page_size();
```
Artemy-Mellanox, line 1419: AFAIR (need to check with Jason) since the driver doesn't know UMEM will be used for MKEY, it leaves possible page sizes selection to the application.
Maybe we could pass `UCS_MASK(ucs_ffs64(address))` so the driver still will be able to pick a larger page size aligned with `address`?

brminich:
@Artemy-Mellanox, pls review

yosefe (approved):

PR #8692, opened 2022-11-10
TEST/JENKINS: Disable osc/ucx component when using mpirun

yosefe:
## Why
Fix recent test failures like this:
```
+ mpirun --bind-to none -x UCX_ERROR_SIGNALS -x UCX_HANDLE_ERRORS -mca pml ob1 -mca btl tcp,self -mca btl_tcp_if_include lo -mca orte_allowed_exit_without_sync 1 -mca coll '^hcoll,ml' -np 1 ./test/mpi/test_memhooks -t flag_no_install -m reloc
==== Running memory hook (flag_no_install mode reloc) on MPI ====
flag_no_install: initialized
total_mapped=1048576
test_memhooks.c:432: Callback for mmap invoked, while hooks were not set
flag_no_install: FAIL
```

It happens because osc/ucx now calls ucp_init() during component_init, which triggers memory hooks install.

### DIFF
--- a/contrib/test_jenkins.sh
+++ b/contrib/test_jenkins.sh
@@ -855,6 +855,7 @@ run_mpi_tests() {
 					-x UCX_ERROR_SIGNALS \
 					-x UCX_HANDLE_ERRORS \
 					-mca pml ob1 \
+					-mca osc ^ucx \
 					-mca btl tcp,self \
 					-mca btl_tcp_if_include lo \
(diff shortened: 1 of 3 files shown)


brminich (dismissed):

yosefe:
@brminich seems MPI tests are passing now. Can you pls take a look again?

brminich (approved):

PR #8695, opened 2022-11-10
TEST/JENKINS: Disable osc/ucx component when using mpirun - v1.14.x

yosefe:
## Why
Clean backport of #8692

### DIFF
--- a/test/mpi/test_memhooks.c
+++ b/test/mpi/test_memhooks.c
@@ -45,16 +45,20 @@
 
 #define SHMAT_FAILED ((void*)-1)
 
-void *event_init(const char *path, ucm_mmap_hook_mode_t mmap_mode);
-void *ext_event_init(const char *path, ucm_mmap_hook_mode_t mmap_mode);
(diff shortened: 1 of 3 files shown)


test/mpi/test_memhooks.c
```diff
@@ -142,40 +146,46 @@ void* open_dyn_lib(const char *lib_path)
...
-    dl_ucm = open_dyn_lib(path);
+    dl_ucm = open_dyn_lib(ucm_path, 0);
     if (dl_ucm == NULL) {
         return NULL;
     }
 
     status = init_ucm_config(dl_ucm, 1, mmap_mode);
     CHKERR_JUMP(status != UCS_OK, "Failed to initialize UCM", fail);
 
+    open_dyn_lib(ucs_path, RTLD_GLOBAL);
```
gleon99, line 163: No need to check rc here? Same in other similar places

yosefe: the test would fail another way if failed to load UCS (also the test is intended to test UCM, not UCS, so if UCS loading fails but test passes it's still OK)

PR #8696, opened 2022-11-10
UCT/CUDA: draft of using managed memory preferred location

Akshay-Venkatesh:
## What
Use preferred location if set by user for managed memory region.

## Why ?
If the user sets preferred location, then it is quite likely that memory of pages reside close to preferred location. If this is different from static decision of bounce buffer location, then performance will be sub-optimal for pipeline protocols.


gleon99:
Hi @Akshay-Venkatesh - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #8700, opened 2022-11-14
UCP/CORE: fix corner case for rkey_index - v.14.x

edgargabriel:
## What
fix a corner case for calculating the rkey_index in case the md_index is not part of the md_map.

## Why ?
Fixes issue https://github.com/openucx/ucx/issues/8627

## How ?
(cherry picked from commit 2034fbc0b911f7ccb3601312f090d28e11f30c6a)


brminich (approved):

brminich:
@edgargabriel, there seems to be a relevant valgrind error  on `swx-rdmz-ucx-gpu-01.rdmz.labs.mlnx`

yosefe (approved):

PR #8701, opened 2022-11-14
UCP/PROTOV2: Pack rkey in rts/rtr when CMA is present

brminich:
## What
Pack rkey to RTR/RTS if CMA transport is present

## Why ?
Enable CMA for rendezvous transports when it is the only RMA capable transport on the system.
This way rndv pipeline protocol based on CMA can be selected when transferring data from device to host.
Gives ~10x speedup with `osu_bw -d cuda D H` and UCX_TLS=`sysv,cma,cuda_copy`
OR about 15% speedup with host memory

### DIFF
--- a/src/ucp/rndv/proto_rndv.inl
+++ b/src/ucp/rndv/proto_rndv.inl
@@ -101,7 +101,8 @@ static UCS_F_ALWAYS_INLINE size_t ucp_proto_rndv_rts_pack(
     rts->size        = req->send.state.dt_iter.length;
     rpriv            = req->send.proto_config->priv;
 
-    if ((rts->size == 0) || (rpriv->md_map == 0)) {
+    if ((rts->size == 0) ||
+        (req->send.state.dt_iter.dt_class != UCP_DATATYPE_CONTIG)) {
         rts->address = 0;
         rkey_size    = 0;
     } else {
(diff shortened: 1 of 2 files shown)


yosefe:
If I understand the problem correctly, then the side that unpacks rts/rtr rkey does not select CMA because its rkey is not found?
in this case, IMO the right fix should be that a rndv bulk lane selection process will select CMA if we have CMA lane, even if the CMA rkey was not found in the unpacked ucp_rkey

brminich:
> If I understand the problem correctly, then the side that unpacks rts/rtr rkey does not select CMA because its rkey is not found? in this case, IMO the right fix should be that a rndv bulk lane selection process will select CMA if we have CMA lane, even if the CMA rkey was not found in the unpacked ucp_rkey

How would we know the remote memory type in this case?

yosefe:
> How would we know the remote memory type in this case?

The remote memory type is also part of the packed UCP rkey

brminich:
> > How would we know the remote memory type in this case?
> 
> The remote memory type is also part of the packed UCP rkey

don't you propose to not pack a key?

yosefe:
> don't you propose to not pack a key?

Yes, i propose that TL Rkey will not be packed for CMA transport at all

brminich:
> > don't you propose to not pack a key?
> 
> Yes, i propose that TL Rkey will not be packed for CMA transport at all

it is not packed, we do not have key for CMA. Before this fix UCP did not sent rkey in the rtr. Now it sends empty rkey (9 bytes) with zero md map and memory type

brminich:
CI problem is #8719

brminich:
@yosefe, pls review. CI failure was unrelated

yosefe (approved):

PR #8703, opened 2022-11-15
UCP/PROTO: Enhanced multiple operations estimation.

rakhmets:
## What
Enhanced estimation of multiple operations performance, introducing a new parameter. The parameter defines the contribution of single operation performance into estimation of multiple operations performance.


yosefe:
IMO a better way is to declare slightly higher overhead for the multi protocol (I guess that's the reason we prefer single when possible, right?)

brminich:
> 

with "multi sends" option on scale performance is charachterized by bw part (which is identical for sinle and multi fragment protos)

yosefe:
> with "multi sends" option on scale performance is charachterized by bw part (which is identical for sinle and multi fragment protos)

I see.. Maybe multi-send should actually be a combination of (1/N)*single + (N-1)/N*multi where N is a configurable "window size" of estimation, for example, 64 or 128. So it would be enough to "tip the scale".
Also, maybe introduce protocol priority instead of relying on the declaration order?

rakhmets:
> > with "multi sends" option on scale performance is charachterized by bw part (which is identical for sinle and multi fragment protos)
> 
> I see.. Maybe multi-send should actually be a combination of (1/N)_single + (N-1)/N_multi where N is a configurable "window size" of estimation, for example, 64 or 128. So it would be enough to "tip the scale". Also, maybe introduce protocol priority instead of relying on the declaration order?

Actually, we have `cfg_priority` field in `ucp_proto_caps_t`. However, it is taken into account only if protocol threshold is set by user. Does it make sense to reconsider the purpose of the field.

rakhmets:
> > with "multi sends" option on scale performance is charachterized by bw part (which is identical for sinle and multi fragment protos)
> 
> I see.. Maybe multi-send should actually be a combination of (1/N)_single + (N-1)/N_multi where N is a configurable "window size" of estimation, for example, 64 or 128. So it would be enough to "tip the scale". Also, maybe introduce protocol priority instead of relying on the declaration order?

@yosefe @brminich I committed changes to update calculation of multiple operation estimation. Do we really need to also change the way we use protocol priority?

brminich:
> 

no, better avoid that

yosefe (commented):
shall we also the update performance tree, or is it in next PR?

yosefe (approved):

src/ucp/proto/proto_select.c
```diff
@@ -464,11 +486,11 @@ static ucs_status_t ucp_proto_select_elem_add_envelope(
...
 
 static ucs_status_t
-ucp_proto_select_elem_init_thresh(ucp_worker_h worker,
-                                  ucp_proto_select_elem_t *select_elem,
+ucp_proto_select_elem_init_thresh(ucp_proto_select_elem_t *select_elem,
                                   ucp_proto_select_init_protocols_t *proto_init,
                                   ucp_worker_cfg_index_t ep_cfg_index,
-                                  ucp_worker_cfg_index_t rkey_cfg_index)
+                                  ucp_worker_cfg_index_t rkey_cfg_index,
+                                  unsigned window_size)
```
brminich, line 493: do we really need this param? I guess it will always be taken from the config, so maybe just take it from there in `ucp_proto_thresholds_next_range`?

brminich (dismissed):

yosefe:
per offline discussion, need to update the performance tree as well

Issue #8669, opened 2022-10-27
Assertion `req->send.ep == ucp_ep' failed when endpoint/process terminates


pentschev:
### Describe the bug
An assertion failure occurs when an endpoint closes and `tcp` and `cuda_copy` (`UCX_TLS=tcp,cuda_copy`) are active and the remote endpoint terminates, the issue does not occur with `UCX_TLS=rc,cuda_copy`. In the simplest reproducer, a `ucx_perftest` listener is launched and then the client process is launched for 2-3 seconds, after which the user presses `CTRL+C` to terminate the client. The reverse (`CTRL+C` on the listener) does not seem to cause the same behavior.

This issue occurs on Dask/UCX-Py as well, but it isn't clear yet whether it happens when the endpoint is terminating or just during some transfer.

<details><summary>Complete output</summary>

```
$ UCX_TLS=tcp,cuda_copy CUDA_VISIBLE_DEVICES=0,1 ucx_perftest -t tag_bw -m cuda-managed -n 1000 -s 1000000000
[1666902559.828992] [dgx13:34156:0]        perftest.c:921  UCX  WARN  CPU affinity is not set (bound to 80 cpus). Performance may be impacted.
Waiting for connection...
Accepted connection from 127.0.0.1:58042
+----------------------------------------------------------------------------------------------------------+
| API:          protocol layer                                                                             |
| Test:         tag match bandwidth                                                                        |
| Data layout:  (automatic)                                                                                |
| Send memory:  cuda-managed                                                                               |
| Recv memory:  cuda-managed                                                                               |
| Message size: 1000000000                                                                                 |
+----------------------------------------------------------------------------------------------------------+
[dgx13:34156:0:34156]      ucp_ep.c:3410 Assertion `req->send.ep == ucp_ep' failed
==== backtrace (tid:  34156) ====
 0  /datasets/pentschev/miniconda3/envs/rn-221026/lib/libucs.so.0(ucs_handle_error+0x2d4) [0x7f12e32f6e14]
 1  /datasets/pentschev/miniconda3/envs/rn-221026/lib/libucs.so.0(ucs_fatal_error_message+0xb8) [0x7f12e32f3ce8]
 2  /datasets/pentschev/miniconda3/envs/rn-221026/lib/libucs.so.0(ucs_fatal_error_format+0xe1) [0x7f12e32f3dd1]
[...]

PR #8705, opened 2022-11-16
UCP/RNDV: fix rtr/ppl error flow

evgeny-leksikov:
## What
fix rtr/ppl error flow

## Why ?
 - `UCP_REQUEST_FLAG_RNDV_FRAG` is a recv request, so we can not access send part of it
 - leaked `req->send.rndv.mdesc` from rtr request

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -3451,7 +3451,8 @@ void ucp_ep_req_purge(ucp_ep_h ucp_ep, ucp_request_t *req,
 
     /* Only send operations could have request ID allocated */
     if (!(req->flags &
-          (UCP_REQUEST_FLAG_RECV_AM | UCP_REQUEST_FLAG_RECV_TAG))) {
+          (UCP_REQUEST_FLAG_RECV_AM | UCP_REQUEST_FLAG_RECV_TAG |
+           UCP_REQUEST_FLAG_RNDV_FRAG))) {
         ucp_send_request_id_release(req);
     }
 
@@ -3473,7 +3474,6 @@ void ucp_ep_req_purge(ucp_ep_h ucp_ep, ucp_request_t *req,
         ucp_request_complete_tag_recv(req, status);
     } else if (req->flags & UCP_REQUEST_FLAG_RNDV_FRAG) {
         ucs_assert(req->flags & UCP_REQUEST_FLAG_SUPER_VALID);
-        ucs_assert(req->send.ep == ucp_ep);
         ucs_assert(recursive); /* Mustn't be directly contained in an EP list
                                 * of tracking requests */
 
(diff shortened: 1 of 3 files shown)


evgeny-leksikov:
@pentschev FYI

evgeny-leksikov:
fixes https://github.com/openucx/ucx/issues/8669

src/ucp/core/ucp_ep.c
```diff
@@ -3496,14 +3496,21 @@ void ucp_ep_req_purge(ucp_ep_h ucp_ep, ucp_request_t *req,
...
-        /* SW RMA/PUT and AMO/Post operations don't allocate local request ID
-         * and don't need to be tracked, since they complete UCP request upon
-         * sending all data to a peer. Receiving RMA/CMPL and AMO/REP packets
-         * complete flush requests */
-        ucs_assert((req->send.uct.func != ucp_rma_sw_proto.progress_put) &&
-                   (req->send.uct.func != ucp_amo_sw_proto.progress_post));
         ucs_assert(req->send.ep == ucp_ep);
 
+        if (req->send.uct.func == ucp_proto_progress_rndv_rtr) {
+            if (req->send.rndv.mdesc != NULL) {
```
yosefe, line 3502: it seems that ucp_rndv_receive->ucp_rndv_req_send_rtr does not initialize desc to NULL

evgeny-leksikov: yeah, it wrongly set by a caller
[`        frndv_req->send.rndv.mdesc   = mdesc;`](https://github.com/openucx/ucx/blob/7ea28295471b18f691ec449d39dee97bd11e6553/src/ucp/rndv/rndv.c#L1405)
which uses other union part of send request `send.rndv` but `ucp_proto_progress_rndv_rtr` uses `send.rndv_rtr`

yosefe: maybe we should check RNDV_FRAG request flag?

evgeny-leksikov: RTR does not have this flag

src/ucp/core/ucp_ep.c
```diff
@@ -3473,7 +3474,6 @@ void ucp_ep_req_purge(ucp_ep_h ucp_ep, ucp_request_t *req,
         ucp_request_complete_tag_recv(req, status);
     } else if (req->flags & UCP_REQUEST_FLAG_RNDV_FRAG) {
         ucs_assert(req->flags & UCP_REQUEST_FLAG_SUPER_VALID);
-        ucs_assert(req->send.ep == ucp_ep);
```
yosefe, line 3476: why?

evgeny-leksikov: see PR description, this is `recv` request, we can not access its `send` part

yosefe: how can it be recv request if we removed it from the pending queue?

evgeny-leksikov: this is a super request of RTR, hierarchy of completions is 
```
rtr_req -> frag_req -> recv_req
(send)     (recv)      (recv)      
```

yosefe: i see
https://github.com/openucx/ucx/blob/dfdd4f4dd733d77c38a70f8de4e3366f8d4c060d/src/ucp/rndv/rndv.c#L1378

yosefe (approved):

PR #8707, opened 2022-11-16
AZP: report roce_proto_enable failures as errors

evgeny-leksikov:
## What
report `roce_proto_enable` failures as errors

## Why ?
to highlight regressions


yosefe (approved):
LGTM, but let's make sure CI passes few times before merging it

PR #8709, opened 2022-11-17
UCP/CORE: Don't disable tl if it used as aux tl in alias.

rakhmets:
## What
Updated the logic of processing the allow/deny list of transports.
Fixed disabling the transport used in the alias as an auxiliary transport.

## Why ?
E.g. `ud_mlx5` transport was disabled, if a user uses `UCX_TLS=^dc`.

Before:
image

After:
image

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -854,7 +854,7 @@ ucp_is_resource_in_transports_list(const char *tl_name,
                                    uint8_t *rsc_flags, uint64_t *tl_cfg_mask)
 {
     uint64_t dummy_mask, tmp_tl_cfg_mask;
-    uint8_t tmp_rsc_flags;
+    uint8_t tmp_rsc_flags, dummy_flags;
     ucp_tl_alias_t *alias;
     char info[32];
     unsigned alias_arr_count;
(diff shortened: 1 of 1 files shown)


src/ucp/core/ucp_context.c
```diff
@@ -883,14 +884,17 @@ ucp_is_resource_in_transports_list(const char *tl_name,
...
+            aux_tl = tmp_rsc_flags & UCP_TL_RSC_FLAG_AUX;
             if (ucp_config_is_tl_name_present(allow_list, alias->alias, 1,
                                               &tmp_rsc_flags,
                                               &tmp_tl_cfg_mask)) {
                 if (allow_list->mode == UCS_CONFIG_ALLOW_LIST_ALLOW) {
                     *rsc_flags   |= tmp_rsc_flags;
                     *tl_cfg_mask |= tmp_tl_cfg_mask;
                     return 1;
-                } else {
+                } else if (!aux_tl) {
```
brminich, line 895: why not just checking `tmp_rsc_flags & UCP_TL_RSC_FLAG_AUX` here?

brminich: from the other side `AUX` flags can also be added in `ucp_config_is_tl_name_present``on line 895 and we will miss it

rakhmets: Yes, but this is not what we want to check. We need only results from `ucp_tls_array_is_present`, because it checks presence of `tl_name` in transport layers belong to the current alias. And `ucp_config_is_tl_name_present` checks presence of alias name in allow/deny list.

brminich (approved):

yosefe (approved):
pls squash

yosefe (approved):

PR #8711, opened 2022-11-17
ROCM/COPY: add rcache parameters

edgargabriel:
## What
1. add parameters to allow control over the rcache used by the rocm_copy component.
2. Remove the short d2h and h2d parameters, there is no need for separate threshold parameters for short vs. zcopy functions.

## Why ?
Improve control of rcache options and simplify threshold parameter handling

CC @akolliasAMD

### DIFF
--- a/src/uct/rocm/copy/rocm_copy_md.c
+++ b/src/uct/rocm/copy/rocm_copy_md.c
@@ -31,6 +31,10 @@ static ucs_config_field_t uct_rocm_copy_md_config_table[] = {
      ucs_offsetof(uct_rocm_copy_md_config_t, enable_rcache),
      UCS_CONFIG_TYPE_TERNARY},
 
+    {"", "", NULL,
+     ucs_offsetof(uct_rocm_copy_md_config_t, rcache),
+     UCS_CONFIG_TYPE_TABLE(uct_md_config_rcache_table)},
+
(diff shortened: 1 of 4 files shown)


src/uct/rocm/copy/rocm_copy_md.c
```diff
@@ -22,17 +22,18 @@
 #include <uct/api/v2/uct_v2.h>
 #include <hsa_ext_amd.h>
 
-static ucs_config_field_t uct_rocm_copy_md_config_table[] = {
-    {"", "", NULL,
-     ucs_offsetof(uct_rocm_copy_md_config_t, super),
-     UCS_CONFIG_TYPE_TABLE(uct_md_config_table)},
+static ucs_config_field_t uct_rocm_copy_md_config_table[] =
+        {{"", "", NULL, ucs_offsetof(uct_rocm_copy_md_config_t, super),
```
yosefe, line 26: pls use the previous code alignment

edgargabriel: ok, I will revert the last change that introduced the format changes to this block, but just for clarification, the only reason I made this change was because the code formatting checker otherwise always reported 'passed with issues',

yosefe (approved):

PR #8714, opened 2022-11-18
Pass user-provided memory type to inline send check

SeyedMir:
## What
Pass user-provided memory type to the function that checks whether the buffer can be sent inline or not.

## Why ?
Inline send can be done for host memory only and we don't detect memory type in inline send path due to overheads. Therefore, on a GPU system, inline send is never used even for host memory buffers. However, if the memory type is already provided by the user and it is host memory, we can try inline send without any extra overhead.

## How ?
Pass the user-provided memory type to the function that checks whether the buffer can be sent inline or not.
Pass UCS_MEMORY_TYPE_UNKNOWN if user has not provided memory type.

### DIFF
--- a/src/ucp/core/ucp_am.c
+++ b/src/ucp/core/ucp_am.c
@@ -926,13 +926,15 @@ static UCS_F_ALWAYS_INLINE ucs_status_t
 ucp_am_try_send_short(ucp_ep_h ep, uint16_t id, uint32_t flags,
                       const void *header, size_t header_length,
                       const void *buffer, size_t length,
-                      const ucp_memtype_thresh_t *max_eager_short)
+                      const ucp_memtype_thresh_t *max_eager_short,
+                      const ucp_request_param_t *param)
 {
     if (ucs_unlikely(flags & UCP_AM_SEND_FLAG_RNDV)) {
         return UCS_ERR_NO_RESOURCE;
     }
 
-    if (ucp_proto_is_inline(ep, max_eager_short, header_length + length)) {
+    if (ucp_proto_is_inline(ep, max_eager_short,
+                            header_length + length, param)) {
         return ucp_am_send_short(ep, id, flags, header, header_length, buffer,
                                  length, flags & UCP_AM_SEND_FLAG_REPLY);
     }
@@ -1019,7 +1021,7 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
 
     if (ucs_likely(attr_mask == 0)) {
         status = ucp_am_try_send_short(ep, id, flags, header, header_length,
-                                       buffer, count, max_short);
+                                       buffer, count, max_short, param);
         ucp_request_send_check_status(status, ret, goto out);
         datatype      = ucp_dt_make_contig(1);
         contig_length = count;
@@ -1028,7 +1030,8 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
         if (ucs_likely(UCP_DT_IS_CONTIG(datatype))) {
             contig_length = ucp_contig_dt_length(datatype, count);
             status = ucp_am_try_send_short(ep, id, flags, header, header_length,
-                                           buffer, contig_length, max_short);
+                                           buffer, contig_length, max_short,
+                                           param);
             ucp_request_send_check_status(status, ret, goto out);
         } else {
             contig_length = 0ul;
--- a/src/ucp/proto/proto_am.inl
+++ b/src/ucp/proto/proto_am.inl
@@ -596,11 +596,13 @@ ucp_proto_get_short_max(const ucp_request_t *req,
 
 static UCS_F_ALWAYS_INLINE int
 ucp_proto_is_inline(ucp_ep_h ep, const ucp_memtype_thresh_t *max_eager_short,
-                    ssize_t length)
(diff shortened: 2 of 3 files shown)


src/ucp/proto/proto_am.inl
```diff
@@ -596,11 +596,15 @@ ucp_proto_get_short_max(const ucp_request_t *req,
 
 static UCS_F_ALWAYS_INLINE int
 ucp_proto_is_inline(ucp_ep_h ep, const ucp_memtype_thresh_t *max_eager_short,
-                    ssize_t length)
+                    ssize_t length, ucs_memory_type_t mem_type)
 {
+    if (ucs_likely(mem_type == UCS_MEMORY_TYPE_UNKNOWN) &&
```
yosefe, line 601: this PR adds at least 2 extra branches on the fast path: `(param->op_attr_mask & UCP_OP_ATTR_FIELD_MEMORY_TYPE)` and this one.
We should make sure these branches are executed only when `ucs_memtype_cache_is_empty()` is false (i.e. there is GPU in the system)

SeyedMir: What if there is GPU in the system and GPU buffers are passed for communications, but the user always provides the memory type? In that case, memtype cache will be empty even though we're using GPU buffers, isn't that so?

yosefe: no, the mem type cache will not be empty because UCM code adds all "unknown" regions to it during initialization.
and anyway, we would add any cuda memory to the memtype cache when it's allocated.

SeyedMir: @yosefe I updated the code and reduced it to one extra branch.

src/ucp/core/ucp_am.c
```diff
@@ -968,13 +970,21 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
...
                                     return UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM));
     UCP_REQUEST_CHECK_PARAM(param);
 
     UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);
 
+    if (ucs_likely(ucs_memtype_cache_is_empty())) {
+        mem_type = UCS_MEMORY_TYPE_HOST;
+    } else {
+        mem_type = (param->op_attr_mask & UCP_OP_ATTR_FIELD_MEMORY_TYPE) ?
+                    param->memory_type : UCS_MEMORY_TYPE_UNKNOWN;
```
yosefe, line 985: IMO we can still avoid the extra branch when memtype cache is empty. since after this PR there would be 2 branches on the ucs_memtype_cache_is_empty() check instead of one.

SeyedMir: How about I initialize mem_type to `UCS_MEMORY_TYPE_HOST` and update it `if (!ucs_memtype_cache_is_empty())` ?

yosefe: same effect. need to push checking mem_type bit field_mask after the existing check of ucs_memtype_cache_is_empty

yosefe: that means something like:
```diff
diff --git a/src/ucp/proto/proto_am.inl b/src/ucp/proto/proto_am.inl
index cdd867120b..2c1f5e93ce 100644
--- a/src/ucp/proto/proto_am.inl
+++ b/src/ucp/proto/proto_am.inl
@@ -596,11 +596,13 @@ ucp_proto_get_short_max(const ucp_request_t *req,
 
 static UCS_F_ALWAYS_INLINE int
 ucp_proto_is_inline(ucp_ep_h ep, const ucp_memtype_thresh_t *max_eager_short,
-                    ssize_t length)
+                    ssize_t length, const ucp_request_param_t *param)
 {
     return (ucs_likely(length <= max_eager_short->memtype_off) ||
             (length <= max_eager_short->memtype_on &&
-             ucs_memtype_cache_is_empty()));
+             ucs_memtype_cache_is_empty())) ||
+           ((param->op_attr_mask & UCP_OP_ATTR_FIELD_MEMORY_TYPE) &&
+            (param->memory_type == UCS_MEMORY_TYPE_HOST));
 }
```

yosefe (approved):

src/ucp/proto/proto_am.inl
```diff
@@ -596,11 +596,13 @@ ucp_proto_get_short_max(const ucp_request_t *req,
 
 static UCS_F_ALWAYS_INLINE int
 ucp_proto_is_inline(ucp_ep_h ep, const ucp_memtype_thresh_t *max_eager_short,
-                    ssize_t length)
+                    ssize_t length, const ucp_request_param_t *param)
 {
     return (ucs_likely(length <= max_eager_short->memtype_off) ||
             (length <= max_eager_short->memtype_on &&
-             ucs_memtype_cache_is_empty()));
+             (ucs_memtype_cache_is_empty() ||
```
brminich, line 603: it's not your code, but can you pelase fix coding style for line above
`length <= max_eager_short->memtype_on` -> `(length <= max_eager_short->memtype_on)`

SeyedMir: Done.

brminich (dismissed):

yosefe (dismissed):
@SeyedMir pls squash

yosefe (approved):

Issue #8684, opened 2022-11-07
Misleading error messages, should be handled by the library user


civodul:
### Describe the bug

Upon error, UCX emits messages along the lines of:

```
ib_iface.c:966  UCX  ERROR ibv_create_cq(cqe=4096) failed: Operation not supported
ib_iface.c:966  UCX  ERROR ibv_create_cq(cqe=4096) failed: Operation not supported
```

Those messages are misleading in the context of Open MPI: when Open MPI is compiled with support for several drivers [as discussed here](https://hpc.guix.info/blog/2019/12/optimized-and-portable-open-mpi-packaging/), it's entirely possible that UCX will not be able to handle the underlying hardware, and Open MPI knows how to deal with that, essentially trying the next driver on its priority list.

Yet the message suggests an error occurred, even though it's up to the caller (Open MPI in this case) to decide whether or not this is an error (this is generally the case with libraries).

I would recommend printing these messages if and only if the user asks for debugging-level logging.

### Setup and versions
- GNU Guix on x86_64, with UCX 1.9.0

### Additional information (depending on the issue)
- OpenMPI 4.1.4
- Output of `ucx_info -d` to show transports and devices recognized by UCX
Note: This is on machines with OmniPath devices:

```
#
# Memory domain: posix
#     Component: posix
#             allocate: unlimited
#           remote key: 24 bytes
#           rkey_ptr is supported
#
#   Transport: posix
#      Device: memory
#
#      capabilities:
#            bandwidth: 0.00/ppn + 12179.00 MB/sec
#              latency: 80 nsec
#             overhead: 10 nsec
#            put_short: <= 4294967295
#            put_bcopy: unlimited
#            get_bcopy: unlimited
#             am_short: <= 100
#             am_bcopy: <= 8256
#               domain: cpu
#           atomic_add: 32, 64 bit
#           atomic_and: 32, 64 bit
#            atomic_or: 32, 64 bit
#           atomic_xor: 32, 64 bit
#          atomic_fadd: 32, 64 bit
#          atomic_fand: 32, 64 bit
#           atomic_for: 32, 64 bit
#          atomic_fxor: 32, 64 bit
#          atomic_swap: 32, 64 bit
#         atomic_cswap: 32, 64 bit
#           connection: to iface
#      device priority: 0
#     device num paths: 1
#              max eps: inf
#       device address: 8 bytes
#        iface address: 8 bytes
#       error handling: none
#
#
# Memory domain: sysv
#     Component: sysv
[...]

PR #8715, opened 2022-11-21
UCT/IB: Disable RDMA transport in case of too low memlock limit value

ivankochin:
## What
Revert #8090

## Why ?
Due to this discussion: https://github.com/openucx/ucx/issues/8684

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -31,7 +31,9 @@
 #include <sys/resource.h>
 
 
-#define UCT_IB_MD_RCACHE_DEFAULT_ALIGN 16
+#define UCT_IB_MD_RCACHE_DEFAULT_ALIGN    16
+#define UCT_IB_MD_REQUIRED_MEMLOCK_LIMIIT (512 * UCS_MBYTE)
+
 
 #define UCT_IB_MD_MEM_DEREG_CHECK_PARAMS(_ib_md, _params) \
     do { \
@@ -1351,6 +1353,7 @@ static ucs_status_t uct_ib_query_md_resources(uct_component_t *component,
     struct ibv_device **device_list;
     ucs_status_t status;
     int i, num_devices;
+    size_t memlock_limit;
 
     UCS_MODULE_FRAMEWORK_LOAD(uct_ib, 0);
 
@@ -1392,6 +1395,26 @@ static ucs_status_t uct_ib_query_md_resources(uct_component_t *component,
         num_resources++;
     }
 
+    if (num_resources == 0) {
+        goto init_res;
+    }
+
+    status = ucs_sys_get_effective_memlock_rlimit(&memlock_limit);
+    if ((status == UCS_OK) &&
+        (memlock_limit <= UCT_IB_MD_REQUIRED_MEMLOCK_LIMIIT)) {
+        /* Disable the RDMA devices because of too strict locked memory limit*/
+        ucs_diag("RDMA transports are disabled because max locked memory "
+                 "limit (%llu kbytes) is too low. Please set max locked "
+                 "memory (ulimit -l) to 'unlimited'",
+                 (memlock_limit / UCS_KBYTE));
+
+        *resources_p     = NULL;
+        *num_resources_p = 0;
+        ucs_free(resources);
+        goto out_free_device_list;
+    }
+
+init_res:
     *resources_p     = resources;
     *num_resources_p = num_resources;
     status = UCS_OK;


src/uct/ib/base/ib_md.c
```diff
@@ -1392,6 +1395,23 @@ static ucs_status_t uct_ib_query_md_resources(uct_component_t *component,
         num_resources++;
     }
 
+    if (num_resources != 0) {
+        status = ucs_sys_get_memlock_rlimit(&memlock_limit);
+        if ((status == UCS_OK) &&
+            (memlock_limit <= UCT_IB_MD_REQUIRED_MEMLOCK_LIMIIT)) {
+            /* Disable the RDMA devices because of too strict locked memory limit*/
+            ucs_warn("RDMA transports are disabled because max locked memory "
```
ivankochin, line 1403:
```suggestion
            ucs_diag("RDMA transports are disabled because max locked memory "
```

src/uct/ib/base/ib_md.c
```diff
@@ -1392,6 +1395,23 @@ static ucs_status_t uct_ib_query_md_resources(uct_component_t *component,
         num_resources++;
     }
 
+    if (num_resources != 0) {
+        status = ucs_sys_get_memlock_rlimit(&memlock_limit);
```
ivankochin, line 1399:
```suggestion
        status = ucs_sys_get_effective_memlock_rlimit(&memlock_limit);
```

gleon99:
👍🏼 
@yosefe ?

src/uct/ib/base/ib_md.c
```diff
@@ -1392,6 +1395,23 @@ static ucs_status_t uct_ib_query_md_resources(uct_component_t *component,
...
+            (memlock_limit <= UCT_IB_MD_REQUIRED_MEMLOCK_LIMIIT)) {
+            /* Disable the RDMA devices because of too strict locked memory limit*/
+            ucs_diag("RDMA transports are disabled because max locked memory "
+                     "limit (%llu kbytes) is too low. Please set max locked "
+                     "memory (ulimit -l) to 'unlimited'",
+                     (memlock_limit / UCS_KBYTE));
+
+            *resources_p     = NULL;
+            *num_resources_p = 0;
+            ucs_free(resources);
```
brminich, line 1410: we changed that behavior because of several reports by different users, that UCX does not use IB transport even though it is available on the system. Maybe initialize the device, but print warning/error instead?
`ucs_diag` is disabled by default, users will not notice the reason for IB mailfunctioning

yosefe: other users complain that we print this error when they don't want to use RDMA.
IMO ucs_diag is a good enough solution for a case things still work but maybe with worse performance

src/uct/ib/base/ib_md.c
```diff
@@ -1392,6 +1395,23 @@ static ucs_status_t uct_ib_query_md_resources(uct_component_t *component,
         num_resources++;
     }
 
+    if (num_resources != 0) {
```
yosefe, line 1398: maybe use `if (num_resources == 0) { goto ... }` to make code look better

yosefe (commented):
pls also change the commit and PR title to better reflect what it's actually doing

src/uct/ib/base/ib_md.c
brminich: i was analysing two reports where users complained about missing ib transports (without any warning sign)

yosefe: we should probably make habit of using UCX_LOG_LEVEL=diag for any issue as 1st debug measure

brminich: ok, this makes sense. I though about making threshold configurable (so we could try 0 while analysing), but we also can just notice the error with diag trace

brminich (dismissed):

yosefe (approved):

src/uct/ib/base/ib_md.c
shamisp: I actually agree with @brminich - imho this patch will cause more harm than good. In this particular bug report, they used HFI NIC. I recently pushed changes to master that check that the device is IB or ROCE. If they have NIC that reports IB or ROCE but for whatever reason `ibv_create_cq` is not supported, this is a driver problem. Please ask them to check with the master and if they still have problem, I believe it is an issue in driver reporting invalid transport.

shamisp: https://github.com/openucx/ucx/commit/f848eea3fff3146ef1539b45d72a595136858399

shamisp (changes requested):
I don't believe this is an UCX issue.

yosefe:
> I don't believe this is an UCX issue.

Better fallback if RDMA is unsupported than to print a scary error message, especially when UCX RDMA stack is not really intended to be used

shamisp:
@yosefe Agree - let's check type of device and if it is not ROCE or IB (we already have this code) let's skip the error. If the device reports IB or ROCE I would keep the error. We probably can and an option to silence the error if the user insists.

src/uct/ib/base/ib_md.c
ivankochin: @shamisp this patch didn't helped at least one of customers. Also, I have reproduced it on the system with IB device even if it wasn't used. So do you still think that we shouldn't merge this patch?

shamisp: Is it real scenario or artificial ? In a single setup mode I get better BW when both IB and SM are used for process to process communication on the same node. I have recently pushed a patch to MD code that disables verbs device that is not ROCE or IB transport.  I think, this potentially can be expanded to check vendor_id and if the specific device does not require memory pinning, the warning can be disabled.

ivankochin: I just ran one of osu benchmark on the system with IB device and memlock limit set to 2 KB. The error was printed, but benchmark executed successfully. As I understand, one of customers in the issue complains about the similar scenario. 

Ok, will try to apply this logic to the patch.

ivankochin: @yosefe @shamisp It seems like little misunderstanding happens. In the issue we have two errors reported by two different people:

1. `ERROR ibv_create_cq(cqe=4096) failed: Operation not supported`, Seems to be unrelated to locked memory limit.
2. `ERROR mlx5_0: ibv_create_cq() failed: Cannot allocate memory. Please set max locked memory (ulimit -l) to 'unlimited' (current: 64 kbytes)`. Related to memory lock limit, but in that case it seems like customer has RoCE device with disabled port, so it is OK that https://github.com/openucx/ucx/commit/f848eea3fff3146ef1539b45d72a595136858399 didn't help. And also it seems OK that we report error after `ibv_create_cq` call because this device require memory pinning and reports about it.

I see the following ways of this problem solution:
1. Apply this patch to prevent errors during `ibv_create_cq` call.
1. Leave the logic as is, but provide programable way to disable current.
1. Try to understand whether device has active ports or not and do not use it if all ports are disabled.

What do you prefer?

shamisp: I agree that there is some confusion is going on.

If all ports are in down mode, we shell probably completely ignore the nic and not to try to create CQ.

ivankochin:
The issue was fixed by https://github.com/openucx/ucx/pull/8757

PR #8718, opened 2022-11-22
UCP/WIREUP: Filter slow lanes in selection process - v1.14.x

shasson5:
## What
Filter slow lanes in selection process

## Why ?
Avoid selecting TCP lanes if we have IB lanes.
The filtering is done by removing lanes with a small enough BW compared to the best lane BW.
By default, each lane with less than 0.25 of the maximum BW is filtered out.
We perform the filter process for both local lanes BW and remote lanes BW.

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1415,26 +1415,55 @@ ucp_wireup_am_bw_score_func(const ucp_worker_iface_t *wiface,
     return size / t * 1e-5;
 }
 
+static void ucp_wireup_get_lane_bw(ucp_wireup_select_info_t *sinfo,
+                                   ucp_worker_h worker,
+                                   const ucp_address_entry_t *address,
+                                   double *bw_local, double *bw_remote)
+{
+    const uct_iface_attr_t *iface_attr;
+    ucp_context_h context = worker->context;
+
+    iface_attr = ucp_worker_iface_get_attr(worker, sinfo->rsc_index);
+    *bw_local  = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+    *bw_remote = address[sinfo->addr_index].iface_attr.bandwidth;
+}
+
 static unsigned
 ucp_wireup_add_fast_lanes(const ucp_wireup_select_params_t *select_params,
-                          ucp_wireup_select_info_t *sinfo, unsigned num_sinfo,
-                          ucp_context_h context, ucp_worker_h worker,
-                          double max_bw, ucp_lane_type_t lane_type,
+                          ucp_wireup_select_info_t *sinfo_array,
+                          unsigned num_sinfo, ucp_worker_h worker,
+                          ucp_lane_type_t lane_type,
                           ucp_wireup_select_context_t *select_ctx)
 {
-    unsigned num_lanes = 0;
-    double max_ratio   = 1. / context->config.ext.multi_lane_max_ratio;
-    const uct_iface_attr_t *iface_attr;
+    unsigned num_lanes     = 0;
+    double max_bw_local    = 0;
+    double max_bw_remote   = 0;
+    ucp_context_h context  = worker->context;
+    const double max_ratio = 1. / context->config.ext.multi_lane_max_ratio;
     ucs_status_t status;
     int show_error;
     unsigned sinfo_index;
-    double lane_bw;
+    double bw_local, bw_remote;
 
+
+    /* Iterate over all elements and calculate max BW */
     for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
-        iface_attr = ucp_worker_iface_get_attr(worker,
-                                               sinfo[sinfo_index].rsc_index);
-        lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
-        if ((lane_bw / max_bw) < max_ratio) {
+        ucp_wireup_get_lane_bw(&sinfo_array[sinfo_index], worker,
+                               select_params->address->address_list, &bw_local,
+                               &bw_remote);
+
+        max_bw_local  = ucs_max(bw_local, max_bw_local);
+        max_bw_remote = ucs_max(bw_remote, max_bw_remote);
+    }
+
+    /* Compare each element to max BW and filter only fast lanes */
+    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
+        ucp_wireup_get_lane_bw(&sinfo_array[sinfo_index], worker,
+                               select_params->address->address_list, &bw_local,
+                               &bw_remote);
+
+        if (((bw_local / max_bw_local) < max_ratio) ||
+            ((bw_remote / max_bw_remote) < max_ratio)) {
             continue;
         }
 
@@ -1471,10 +1500,8 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
     ucp_md_map_t md_map;
     ucp_rsc_index_t rsc_index;
     unsigned addr_index;
-    double max_bw, lane_bw;
 
     num_sinfo             = 0;
-    max_bw                = 0;
     md_map                = bw_info->md_map;
     local_dev_bitmap      = bw_info->local_dev_bitmap;
     remote_dev_bitmap     = bw_info->remote_dev_bitmap;
@@ -1500,10 +1527,6 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
             dev_index                   = context->tl_rscs[rsc_index].dev_index;
             sinfo[num_sinfo].path_index = dev_count.local[dev_index];
             num_sinfo++;
-
-            iface_attr = ucp_worker_iface_get_attr(ep->worker, rsc_index);
-            lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
-            max_bw     = ucs_max(lane_bw, max_bw);
         } else {
             /* disqualify/count lane_desc_idx */
             addr_index      = select_ctx->lane_descs[excl_lane].addr_index;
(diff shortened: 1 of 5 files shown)


gleon99:
@shasson5 please elaborate in the PR description, how the filtering is done.

src/ucp/wireup/select.c
```diff
@@ -1415,29 +1415,66 @@ ucp_wireup_am_bw_score_func(const ucp_worker_iface_t *wiface,
...
 
+static unsigned
+ucp_wireup_add_fast_lanes(const ucp_wireup_select_params_t *select_params,
+                          ucp_wireup_select_info_t *sinfo, unsigned num_sinfo,
+                          ucp_context_h context, ucp_worker_h worker,
+                          double max_bw, ucp_lane_type_t lane_type,
+                          ucp_wireup_select_context_t *select_ctx)
+{
+    unsigned num_lanes = 0;
+    double max_ratio   = 1. / context->config.ext.multi_lane_max_ratio;
```
gleon99, line 1426: `const`

src/ucp/wireup/select.c
```diff
@@ -1498,8 +1532,9 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
     }
 
     bw_info->criteria.arg = NULL; /* To suppress compiler warning */
-
-    return num_lanes;
+    return ucp_wireup_add_fast_lanes(select_params, sinfo, num_sinfo, context,
```
gleon99, line 1535: Restore blank line

gleon99:
@shasson5  pipeline failures seem real
https://dev.azure.com/ucfconsort/0b36e3f0-8ab9-4a48-b68b-4b2350e02c88/_apis/build/builds/54446/logs/489

src/ucp/wireup/select.c
```diff
@@ -1446,30 +1483,27 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
-                                                   bw_info->criteria.lane_type,
-                                                   show_error, select_ctx);
-            if (status != UCS_OK) {
-                break;
-            }
+            rsc_index                   = sinfo[num_sinfo].rsc_index;
+            addr_index                  = sinfo[num_sinfo].addr_index;
+            dev_index                   = context->tl_rscs[rsc_index].dev_index;
+            sinfo[num_sinfo].path_index = dev_count.local[dev_index];
+            num_sinfo++;
```
gleon99, line 1502: Notice that `{rsc,addr,dev}_index` are used afterwards (still inside the loop), even though the lane is not actually added

src/ucp/wireup/select.c
```diff
@@ -1446,30 +1483,27 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
 
-            num_lanes++;
+            iface_attr = ucp_worker_iface_get_attr(ep->worker, rsc_index);
+            lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+            max_bw     = ucs_max(lane_bw, max_bw);
```
yosefe, line 1506: i would move this calculation inside ucp_wireup_add_fast_lanes

src/ucp/wireup/select.c
```diff
@@ -1415,26 +1415,55 @@ ucp_wireup_am_bw_score_func(const ucp_worker_iface_t *wiface,
...
     for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
-        iface_attr = ucp_worker_iface_get_attr(worker,
-                                               sinfo[sinfo_index].rsc_index);
-        lane_bw    = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
-        if ((lane_bw / max_bw) < max_ratio) {
+        ucp_wireup_get_lane_bw(&sinfo_array[sinfo_index], worker,
+                               select_params->address->address_list, &bw_local,
+                               &bw_remote);
+
+        max_bw_local  = ucs_max(bw_local, max_bw_local);
```
yosefe, line 1455: calc one max which is the min of {local, remote} for each lane

src/ucp/wireup/select.c
gleon99: Other variables are modified according to that, assuming the lane was added (which is not the case anymore, in your implementation).
It also might be related to the observed pipeline failures.

brminich:
what is the use case with tcp when local bw is good, but the remote bw is not sufficient enough?
Also we take into account local and remote perf charasteristics in `ucp_wireup_rma_bw_score_func`

src/ucp/wireup/select.c
```diff
@@ -1415,29 +1415,87 @@ ucp_wireup_am_bw_score_func(const ucp_worker_iface_t *wiface,
...
+
+    iface_attr = ucp_worker_iface_get_attr(worker, sinfo->rsc_index);
+    bw_local   = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+    bw_remote  = address[sinfo->addr_index].iface_attr.bandwidth;
+    return ucs_min(bw_local, bw_remote);
+}
+
+static unsigned
+ucp_wireup_add_fast_lanes(const ucp_wireup_select_params_t *select_params,
+                          ucp_wireup_select_info_t *sinfo_array,
```
yosefe, line 1434: could it be const?

shasson5:
@brminich 

> what is the use case with tcp when local bw is good, but the remote bw is not sufficient enough? Also we take into account local and remote perf charasteristics in `ucp_wireup_rma_bw_score_func`

If remote BW is not sufficient then it will affect lanes selection on the remote endpoint (which will cause endpoint reconf issue).

yosefe (dismissed):

brminich:
> ucp_wireup_rma_bw_score_func

Ok, then do we need to cleanup other places where the filtering is done currently? Like `ucp_ep_config_rndv_zcopy_set`

shasson5:
> > ucp_wireup_rma_bw_score_func
> 
> Ok, then do we need to cleanup other places where the filtering is done currently? Like `ucp_ep_config_rndv_zcopy_set`

I guess we should, but anyway I think it does not effects this PR, so we can do it on a separate PR.
@yosefe WDYT?

yosefe:
> > ucp_wireup_rma_bw_score_func
> 
> Ok, then do we need to cleanup other places where the filtering is done currently? Like `ucp_ep_config_rndv_zcopy_set`

That filtering considers only lanes that could be used for the operation when calculating max bw
Which can actually create a problem: cuda_ipc is very fast so it will effectively disable other shared memory transports.

@shasson5 let's change this PR so it will consider also memory type when filtering the lanes

shasson5:
> 

you mean to only filter network transports?

yosefe:
> you mean to only filter network transports?

No; calculate max bw per memory type. if a memory type is not supported the max bw is considered 0
then, we remove a lane if : for each memory type it supports, lane_bw is lower than max_bw for that memory type.
for example shared mem (cma/knem/..) will not be removed because they support HOST but max_bw[HOST] is will not include cuda_ipc bw.

yosefe (commented):
LGTM, need to double-check per-mem-type lane selection works as expected

gleon99:
@shasson5 there is no calculation per mem type..

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -211,7 +211,9 @@ class test_ucp_worker_discard : public ucp_test {
         flush_req = sender().flush_worker_nb(0);
         if ((flush_req == NULL) &&
             (get_variant_value() & TEST_DISCARD_DISABLED)) {
-            UCS_TEST_SKIP_R("all EPs returned UCS_OK in 'flush_worker_nb'");
+            ucs::detail::message_stream ms("INFO");
+            ms << "all EPs returned UCS_OK in 'flush_worker_nb'" << std::endl;
```
yosefe, line 215: `UCS_TEST_MESSAGE << "all EPs returned UCS_OK in 'flush_worker_nb'"`

yosefe (dismissed):

test/gtest/ucp/ucp_test.cc
```diff
@@ -113,6 +113,18 @@ bool ucp_test::has_transport(const std::string& tl_name) const {
     return check_transport(tl_name, GetParam().transports);
 }
 
+bool
```
gleon99, line 116: Try to keep the same order of methods in header and implementation filed.

test/gtest/ucp/ucp_test.cc
```diff
@@ -113,6 +113,18 @@ bool ucp_test::has_transport(const std::string& tl_name) const {
...
+ucp_test::has_resource(const ucp_test_base::entity *e, const std::string &tl_name) const
+{
+    for (int i = 0; i < e->ucph()->num_tls; ++i) {
```
gleon99, line 119: `auto`?

src/ucp/wireup/select.c
```diff
@@ -1415,29 +1415,87 @@ ucp_wireup_am_bw_score_func(const ucp_worker_iface_t *wiface,
...
+                                     const ucp_address_entry_t *address)
+{
+    ucp_context_h context = worker->context;
+    const uct_iface_attr_t *iface_attr;
+    double bw_local, bw_remote;
+
+    iface_attr = ucp_worker_iface_get_attr(worker, sinfo->rsc_index);
+    bw_local   = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+    bw_remote  = address[sinfo->addr_index].iface_attr.bandwidth;
+    return ucs_min(bw_local, bw_remote);
```
gleon99, line 1429: Add blank line

test/gtest/ucp/ucp_test.cc
shasson5: actually these changes can be reverted. (no need to refactor this function out of test_wireup)

test/gtest/ucp/ucp_test.cc
shasson5: actually these changes can be reverted. (no need to refactor this function out of test_wireup)

src/ucp/wireup/select.c
shasson5: you mean before "return"?

gleon99: Yes

yosefe (approved):

src/ucp/wireup/select.c
```diff
@@ -1415,29 +1415,88 @@ ucp_wireup_am_bw_score_func(const ucp_worker_iface_t *wiface,
...
+                                         select_params->address->address_list);
+        max_bw  = ucs_max(lane_bw, max_bw);
+    }
+
+    /* Compare each element to max BW and filter only fast lanes */
+    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
+        lane_bw = ucp_wireup_get_lane_bw(&sinfo_array[sinfo_index], worker,
+                                         select_params->address->address_list);
+
+        if ((max_bw > 0) && ((lane_bw / max_bw) < max_ratio)) {
```
gleon99, line 1461: If `max_bw == 0` for some reason, can skip this loop - no need to check in each iteration.

PR #8727, opened 2022-11-27
UCS: enable remove unused function by linker

changchengx:
"__attribute__((used))" prevents the compiler from generating warnings if the function is not referenced and avoids being removed by linker.
"__attribute__((unsed))" does not change the behavior of the unused function removal process .i.e. linker could remove the unused function.

## What
remove unused functions from the generated executable file.

### DIFF
--- a/src/ucm/util/reloc.h
+++ b/src/ucm/util/reloc.h
@@ -51,8 +51,8 @@ ucs_status_t ucm_reloc_modify(ucm_reloc_patch_t* patch);
  *
  * @return Original function pointer for 'symbol'.
  */
-static void* UCS_F_MAYBE_UNUSED
-ucm_reloc_get_orig(const char *symbol, void *replacement)
+static UCS_F_MAYBE_UNUSED
+void* ucm_reloc_get_orig(const char *symbol, void *replacement)
 {
     const char *error;
     void *func_ptr;
--- a/src/ucs/datastruct/ptr_array.c
+++ b/src/ucs/datastruct/ptr_array.c
@@ -80,7 +80,8 @@ ucs_ptr_array_freelist_element_set_free_ahead(ucs_ptr_array_elem_t *elem,
                                        ucs_ptr_array_freelist_get_next(*elem));
 }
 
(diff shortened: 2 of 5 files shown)


changchengx:
```static void* UCS_F_MAYBE_UNUSED function(...)``` may need to be changed as ```static UCS_F_MAYBE_UNUSED void* function(...)```, below experiment shows the reason.
```C
// Method A, has error : $ cat this_file | gcc -x c -Werror=unused-function -
static void* __attribute__((unused)) func(void) {return 0;} 
// Method B, no error : $ cat this_file | gcc -x c -Werror=unused-function -
// static __attribute__((unused)) void* func(void) {return 0;} 

int main(void)
{
    return 0;
}
```

src/ucm/util/reloc.h
```diff
@@ -51,8 +51,8 @@ ucs_status_t ucm_reloc_modify(ucm_reloc_patch_t* patch);
  *
  * @return Original function pointer for 'symbol'.
  */
-static void* UCS_F_MAYBE_UNUSED
-ucm_reloc_get_orig(const char *symbol, void *replacement)
+static UCS_F_MAYBE_UNUSED
+void* ucm_reloc_get_orig(const char *symbol, void *replacement)
```
yosefe, line 55: seems need to fix also these, similarly:
ucs_ptr_array_dump
uct_rc_mlx5_devx_create_cmd_qp
uct_rc_mlx5_verbs_create_cmd_qp

changchengx: Done. Force push with small(<50) LOC change

yosefe (dismissed):

src/ucm/util/reloc.h
gleon99: Also `uct_ib_mlx5_iface_fill_attr`

changchengx: @gleon99 Done. Sorry for missing it.

yosefe (approved):

PR #8731, opened 2022-11-29
AZP: proto_enable on gpu

evgeny-leksikov:
## What
proto_enable on gpu (warn)

## Why ?
test coverage

### DIFF
--- a/buildlib/pr/main.yml
+++ b/buildlib/pr/main.yml
@@ -265,6 +265,12 @@ stages:
         proto_enable: yes
     - template: tests.yml
       parameters:
+        name: gpu_proto_enable
+        demands: ucx_gpu -equals yes
+        test_perf: 0
+        proto_enable: yes
+    - template: tests.yml
+      parameters:
         name: BlueField
         demands: ucx_bf -equals yes
         test_perf: 0
--- a/buildlib/pr/tests.yml
+++ b/buildlib/pr/tests.yml
@@ -30,13 +30,7 @@ jobs:
           source ./buildlib/az-helpers.sh
           az_init_modules
           ./contrib/test_jenkins.sh
-          rc=$?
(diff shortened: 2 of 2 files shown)


buildlib/pr/tests.yml
```diff
@@ -30,7 +30,13 @@ jobs:
           source ./buildlib/az-helpers.sh
           az_init_modules
           ./contrib/test_jenkins.sh
-          exit $?
+          rc=$?
+          if [[ $rc != 0 && x'${{ parameters.proto_enable }}' == x'yes' && \
+                x'${{ parameters.ucx_gpu }}' == x'yes' ]]; then
```
evgeny-leksikov, line 35: x'${{ parameters.demands }}' == x'ucx_gpu -equals yes'

buildlib/pr/main.yml
```diff
@@ -263,6 +263,12 @@ stages:
         demands: ucx_roce -equals yes
         test_perf: 0
         proto_enable: yes
+    - template: tests.yml
+      parameters:
+        name: gpu_proto_enable
+        demands: ucx_gpu -equals yes
+        test_perf: 0
+        proto_enable: yes
```
yosefe, line 271: this will double the load on GPU machines and can lead to a longer time to get CI results

yosefe: IMO we should keep this PR open to fix the new CI failures

evgeny-leksikov: if we won't merge this PR then maybe report failures as error?
I think we could add new variants to tests failed in this PR to cover related code.

yosefe: > if we won't merge this PR then maybe report failures as error?

Yes, let's do that

artemry-nv:
Is it still actual?

evgeny-leksikov:
new protocols are enabled by default now, closing

PR #8733, opened 2022-11-30
UCT/IB: Add ports check during md_open

ivankochin:
## What
Additional ports availability check during `md_open`.

## Why ?
The ports availability should be checked as soon as possible to make customer experience better.

### DIFF
--- a/examples/uct_hello_world.c
+++ b/examples/uct_hello_world.c
@@ -346,11 +346,8 @@ static ucs_status_t dev_tl_lookup(const cmd_args_t *cmd_args,
         for (md_index = 0; md_index < component_attr.md_resource_count; ++md_index) {
             status = uct_md_config_read(components[cmpt_index], NULL, NULL,
                                         &md_config);
-            if (UCS_OK != status) {
-                fprintf(stderr, "Failed to read MD config: %s\n",
-                        component_attr.md_resources[md_index].md_name);
-                continue;
-            }
+            CHKERR_JUMP(UCS_OK != status, "read MD config",
+                        release_component_list);
 
             status = uct_md_open(components[cmpt_index],
                                  component_attr.md_resources[md_index].md_name,
--- a/examples/uct_hello_world.c
+++ b/examples/uct_hello_world.c
@@ -346,19 +346,24 @@ static ucs_status_t dev_tl_lookup(const cmd_args_t *cmd_args,
         for (md_index = 0; md_index < component_attr.md_resource_count; ++md_index) {
             status = uct_md_config_read(components[cmpt_index], NULL, NULL,
                                         &md_config);
-            CHKERR_JUMP(UCS_OK != status, "read MD config",
-                        release_component_list);
+            if (UCS_OK != status) {
+                fprintf(stderr, "Failed to read MD config: %s\n",
+                        component_attr.md_resources[md_index].md_name);
+                continue;
+            }
 
             status = uct_md_open(components[cmpt_index],
                                  component_attr.md_resources[md_index].md_name,
                                  md_config, &iface_p->md);
             uct_config_release(md_config);
-            CHKERR_JUMP(UCS_OK != status, "open memory domains",
-                        release_component_list);
+            if (UCS_OK != status) {
+                fprintf(stderr, "Failed to open MD: %s\n",
+                        component_attr.md_resources[md_index].md_name);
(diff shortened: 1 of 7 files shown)


yosefe:
@ivankochin where is the failing ibv_create_cq() coming from?
also need to align the code style

ivankochin:
From here: https://github.com/openucx/ucx/blob/9365d7dcaca2f4ca5bcbbf5d641c781078892e68/src/uct/ib/mlx5/dv/ib_mlx5_dv.c#L64

src/uct/ib/base/ib_device.c
```diff
@@ -1095,6 +1095,21 @@ ucs_status_t uct_ib_modify_qp(struct ibv_qp *qp, enum ibv_qp_state state)
...
+
+    for (port_num = dev->first_port; port_num < dev->first_port + dev->num_ports;
+         ++port_num)
+    {
+        status = uct_ib_device_port_check(dev, port_num, 0);
+        if (status == UCS_OK) {
+            return UCS_OK;
+        }
+    }
+    return UCS_ERR_UNREACHABLE;
```
gleon99, line 1110: Maybe `return status` instead?

examples/uct_hello_world.c
```diff
@@ -346,19 +346,24 @@ static ucs_status_t dev_tl_lookup(const cmd_args_t *cmd_args,
         for (md_index = 0; md_index < component_attr.md_resource_count; ++md_index) {
             status = uct_md_config_read(components[cmpt_index], NULL, NULL,
                                         &md_config);
-            CHKERR_JUMP(UCS_OK != status, "read MD config",
-                        release_component_list);
+            if (UCS_OK != status) {
```
gleon99, line 349: Are you sure we need to modify this flow and the code after `uct_md_query`?
Isn't it enough to only change the logic for `uct_md_open`?

examples/uct_hello_world.c
```diff
@@ -346,19 +346,24 @@ static ucs_status_t dev_tl_lookup(const cmd_args_t *cmd_args,
...
+                continue;
+            }
 
             status = uct_md_open(components[cmpt_index],
                                  component_attr.md_resources[md_index].md_name,
                                  md_config, &iface_p->md);
             uct_config_release(md_config);
-            CHKERR_JUMP(UCS_OK != status, "open memory domains",
-                        release_component_list);
+            if (UCS_OK != status) {
```
gleon99, line 359: Seems such changes are needed in several places, see also pipeline (e,g `test_md`).

test/gtest/uct/test_md.cc
```diff
@@ -97,13 +97,20 @@ test_md::test_md()
 
 void test_md::init()
 {
+    uct_md_h h;
```
gleon99, line 100: `h` -> `md`

src/uct/ib/base/ib_device.c
ivankochin, line 1110:
```suggestion
    return status;
```

examples/uct_hello_world.c
```diff
@@ -353,12 +353,14 @@ static ucs_status_t dev_tl_lookup(const cmd_args_t *cmd_args,
...
+                fprintf(stderr, "Failed to open MD: %s\n",
+                        component_attr.md_resources[md_index].md_name);
+                continue;
+            }
 
             status = uct_md_query(iface_p->md, &iface_p->md_attr);
-            CHKERR_JUMP(UCS_OK != status, "query iface",
-                        close_md);
+            CHKERR_JUMP(UCS_OK != status, "query iface", close_md);
```
gleon99, line 363: Minor: maybe fix `query iface` -> `query MD`?

src/uct/ib/base/ib_device.c
```diff
@@ -1095,6 +1095,19 @@ ucs_status_t uct_ib_modify_qp(struct ibv_qp *qp, enum ibv_qp_state state)
     return UCS_OK;
 }
 
+ucs_status_t uct_ib_device_check_ports(uct_ib_device_t *dev) {
+    ucs_status_t status = UCS_ERR_UNREACHABLE;
```
gleon99, line 1099: Is the initialization needed?

ivankochin:
It was decided that we cannot fail md_open call in case of absence of the available ports.

PR #8736, opened 2022-12-01
UCT/DC: Do not show cq_available in VFS

brminich:
## What
Do not show available CQ resources for DC transport in VFS

## Why ?
It is misleading value, because DC creates CQ of sufficient length and does not do CQ resources moderation


src/uct/ib/rc/base/rc_iface.c
```diff
@@ -1042,7 +1045,7 @@ void uct_rc_iface_vfs_refresh(uct_iface_h iface)
     uct_rc_ep_t *ep;
 
     /* Add iface resources */
-    uct_rc_iface_vfs_populate(rc_iface);
+    uct_rc_iface_vfs_populate(rc_iface, 1);
```
yosefe, line 1048: can we call ucs_vfs_obj_add_ro_file(..., "cq_available") here instead of passing a flag to uct_rc_iface_vfs_populate?

brminich: ok

yosefe (approved):

PR #8737, opened 2022-12-01
UCP/PROTO: reset rndv/recv/ppln and rndv/get/mtype

evgeny-leksikov:
## What
reset implementation for rndv/recv/ppln and rndv/get/mtype protocols

## Why ?
missed functionality in protov2

### DIFF
--- a/src/ucp/rndv/rndv_ppln.c
+++ b/src/ucp/rndv/rndv_ppln.c
@@ -179,11 +179,28 @@ void ucp_proto_rndv_ppln_send_frag_complete(ucp_request_t *freq, int send_ack)
                                       "ppln_send");
 }
 
+static ucs_status_t ucp_proto_rndv_recv_ppln_reset(ucp_request_t *req)
+{
+    ucs_assert(req->send.rndv.ppln.ack_data_size == 0);
+
+    if (!ucp_proto_common_multi_frag_is_completed(req)) {
+        return UCS_OK;
+    }
+
+    req->status                    = UCS_OK;
+    req->send.state.dt_iter.offset = 0;
+    ucp_proto_request_restart(req);
+    return UCS_OK;
+}
+
 void ucp_proto_rndv_ppln_recv_frag_clean(ucp_request_t *freq)
 {
-    ucp_proto_rndv_ppln_frag_complete(
-            freq, 0, 0, (ucp_proto_complete_cb_t)ucs_empty_function,
-            "ppln_recv_clean");
+    ucp_send_request_id_release(freq);
+
+    /* abort freq since super request may change protocol */
+    ucp_proto_rndv_ppln_frag_complete(freq, 0, 1,
+                                      ucp_proto_rndv_recv_ppln_reset,
+                                      "ppln_recv_clean");
 }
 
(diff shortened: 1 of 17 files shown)


src/ucp/rndv/rndv_ppln.c
```diff
@@ -179,11 +179,25 @@ void ucp_proto_rndv_ppln_send_frag_complete(ucp_request_t *freq, int send_ack)
                                       "ppln_send");
 }
 
+static ucs_status_t ucp_proto_rndv_ppln_recv_frag_reset(ucp_request_t *req)
+{
+    ucs_assert(req->send.rndv.ppln.ack_data_size == 0);
+
+    if (!ucp_proto_common_frag_is_completed(req)) {
```
yosefe, line 186: ucp_proto_common_frag_is_completed is actually expecting a top-level request (not a frag request) since it's checking completed_size field. So its name is misleading. It should be something like ucp_proto_common_multi_frag_is_completed
here also, ucp_proto_rndv_ppln_recv_frag_reset is expecting a top-level request, so its name should be ucp_proto_rndv_recv_ppln_reset or ucp_proto_rndv_recv_ppln_clean

src/ucp/rndv/rndv_ppln.c
```diff
@@ -239,7 +253,8 @@ static ucs_status_t ucp_proto_rndv_ppln_progress(uct_pending_req_t *uct_req)
         ucp_proto_request_select_proto(freq, &rpriv->frag_proto,
                                        freq->send.state.dt_iter.length);
 
-        ucp_trace_req(req, "send fragment request %p", freq);
+        ucp_trace_req(req, "send fragment req %p offset %zu size %zu", freq,
```
yosefe, line 256: req is already printed by ucp_trace_req

src/ucp/rndv/rndv_rtr.c
```diff
@@ -309,14 +309,14 @@ ucp_proto_rndv_rtr_mtype_abort(ucp_request_t *req, ucs_status_t status)
...
+static ucs_status_t ucp_proto_rndv_rtr_mtype_reset(ucp_request_t *req)
 {
     ucs_mpool_put_inline(req->send.rndv.mdesc);
     if (ucp_proto_rndv_request_is_ppln_frag(req)) {
         ucp_proto_rndv_ppln_recv_frag_clean(req);
-        ucp_proto_request_bcopy_id_reset(req);
+        return UCS_ERR_CANCELED;
     } else {
-        ucp_proto_request_zcopy_id_reset(req);
+        return ucp_proto_request_zcopy_id_reset(req);
```
yosefe, line 319: remove else on line 318

src/ucp/rndv/rndv_ppln.c
evgeny-leksikov: `req` and `freq` are different requests, IMO it's useful to see their relationship in debug traces

yosefe: ah, so need to print "freq" instead of "req" here

yosefe (dismissed):

evgeny-leksikov:
@yosefe squashed

yosefe (approved):

PR #8739, opened 2022-12-02
GTEST: Skip transports with sockaddr NULL

shamisp:
## What
On VPN virtual NICs the value is NULL and it causes segfault in gtest

Signed-off-by: Pasha (Pavel) Shamis <pshamis@nvidia.com>

### DIFF
--- a/test/gtest/common/test_helpers.cc
+++ b/test/gtest/common/test_helpers.cc
@@ -434,6 +434,10 @@ void safe_usleep(double usec) {
 }
 
 bool is_inet_addr(const struct sockaddr* ifa_addr) {
+    if (ifa_addr == NULL) {
+        return false;
+    }
+
     if (ifa_addr->sa_family == AF_INET6) {
         /* Skip IPv6 link-local and loopback address, that could not be used for
            connection establishment */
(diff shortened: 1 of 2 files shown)


shamisp:
@yosefe seems like CI failure is irrelevant

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -165,7 +165,7 @@ class test_ucp_sockaddr : public ucp_test {
              * isn't as such, we continue to the next one. */
             skip = 1;
         } else if ((has_transport("tcp") || has_transport("all")) &&
-                   (ifa->ifa_addr->sa_family == AF_INET6)) {
+                   (ifa->ifa_addr != 0 && ifa->ifa_addr->sa_family == AF_INET6)) {
```
yosefe, line 168: 0 -> NULL
use ( ) around inner expressions

test/gtest/common/test_helpers.cc
```diff
@@ -434,6 +434,10 @@ void safe_usleep(double usec) {
 }
 
 bool is_inet_addr(const struct sockaddr* ifa_addr) {
+    if (ifa_addr == NULL) {
+        return 0;
```
yosefe, line 438: return false

shamisp:
@yosefe it seems that the failures are irrelevant. Looks like some major issue with CI/CD

yosefe (approved):

yosefe (approved):

PR #8745, opened 2022-12-06
PERF: Remove support for librte

shamisp:
Why: Librte does not exits anymore and it is not maintained.


gleon99:
Can remove also mentions of RTE in:
- `glossaries.rst`
- `src/tools/perf/configure.m4` ("TODO")
- `src/tools/perf/{,lib/}Makefile.am` (flags)
- `libperf.h`

shamisp:
`libperf.h` - the rte struct is used for MPI
`glossaries.rst ` - the term is still relevant since we use it in the code.

yosefe (dismissed):

yosefe (dismissed):

src/tools/perf/Makefile.am
```diff
@@ -23,9 +23,9 @@ ucx_perftest_SOURCES = \
 	perftest_run.c \
 	perftest_params.c
 
-ucx_perftest_CPPFLAGS = $(BASE_CPPFLAGS) $(RTE_CPPFLAGS)
+ucx_perftest_CPPFLAGS = $(BASE_CPPFLAGS)
```
gleon99, line 26: Also the same under `lib/Makefile.am`?

shamisp: Good catch. I updated and squashed to safe the time.

shamisp:
the error looks irrelevant

yosefe (approved):

PR #8747, opened 2022-12-06
GTEST: Limit max gtest message size to BAR1 size

shamisp:
## Why
Current max message size does not take in consideration the limitations of BAR1 size. For ML GPUs the BAR1 is very large and it has never been a limitation. For consumer class of GPUs the BAR1 size can be in a range of ~ megabytes. As a result, gtest can fail. 

## What 
This patch adds query for the BAR1 size and limits the max message size.

## How ?
NVML has an API to query Bar1 size

### DIFF
--- a/test/gtest/common/mem_buffer.cc
+++ b/test/gtest/common/mem_buffer.cc
@@ -153,7 +153,7 @@ size_t mem_buffer::get_bar1_free_size()
 {
     /* All gtest CUDA tests explicitly assume that all memory allocations are
      * done on the device 0. The same assumption is followed here. */
-    size_t max_availible_size = SIZE_MAX;
+    size_t availible_size = SIZE_MAX;
 
 #if HAVE_NVML_H
     nvmlReturn_t ret;
@@ -165,19 +165,19 @@ size_t mem_buffer::get_bar1_free_size()
         /* For whatever reason we cannot open device handle.
          * As a result let's assume there is no limit on the size
          * and in the worse case scenario gtest will fail in runtime */
-        return max_availible_size;
+        return availible_size;
     }
 
     ret = nvmlDeviceGetBAR1MemoryInfo(device, &bar1mem);
     if (ret != NVML_SUCCESS) {
         /* Similarly let's assume there is no limit on the size */
-        return max_availible_size;
+        return availible_size;
     }
 
-    max_availible_size = (size_t)bar1mem.bar1Free;
+    availible_size = (size_t)bar1mem.bar1Free;
 #endif
 
-    return max_availible_size;
+    return availible_size;
 }
 
 void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type)
--- a/test/gtest/common/mem_buffer.cc
+++ b/test/gtest/common/mem_buffer.cc
@@ -149,7 +149,7 @@ void mem_buffer::set_device_context()
     device_set = true;
 }
 
-const size_t mem_buffer::get_bar1_free_size()
+size_t mem_buffer::get_bar1_free_size()
 {
     /* All gtest CUDA tests explicitly assume that all memory allocations are
      * done on the device 0. The same assumption is followed here. */
--- a/test/gtest/common/mem_buffer.cc
+++ b/test/gtest/common/mem_buffer.cc
@@ -32,6 +32,10 @@
 
 #endif
 
+#if HAVE_NVML_H
+#include <nvml.h>
+#endif
+
 #if HAVE_ROCM
 #  include <hip_runtime.h>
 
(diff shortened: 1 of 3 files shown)


test/gtest/common/mem_buffer.h
```diff
@@ -88,6 +88,10 @@ class mem_buffer {
     /* returns whether ROCM device supports managed memory */
     static bool is_rocm_managed_supported();
 
+    /* Return free memory on the BAR1 / GPU. If GPU is not used
+     * SIZE_MAX is returned */
+    static const size_t get_bar1_free_size();
```
gleon99, line 93: Perhaps `static size_t get_bar1_free_size() const`?

test/gtest/common/mem_buffer.cc
```diff
@@ -32,6 +32,10 @@
 
 #endif
 
+#if HAVE_NVML_H
```
gleon99, line 35: Need to add this definition to `configure`?

test/gtest/uct/uct_p2p_test.cc
```diff
@@ -182,6 +182,10 @@ void uct_p2p_test::test_xfer_multi_mem_type(send_func_t send, size_t min_length,
     /* Trim at 4.1 GB */
     max_length = ucs_min(max_length, (size_t)(4.1 * (double)UCS_GBYTE));
 
+    /* Trim by BAR1 size if relevant */
+    max_length = ucs_min(max_length,
+                         (size_t)(0.8 * mem_buffer::get_bar1_free_size()));
```
yosefe, line 187: can we do it only for GPU memory?

shamisp: Good point

test/gtest/common/mem_buffer.h
shamisp: I can remove const all together

test/gtest/common/mem_buffer.cc
shamisp: it is already defined and coming from config.h

Akshay-Venkatesh (dismissed):

test/gtest/uct/uct_p2p_test.cc
```diff
@@ -182,6 +182,13 @@ void uct_p2p_test::test_xfer_multi_mem_type(send_func_t send, size_t min_length,
     /* Trim at 4.1 GB */
     max_length = ucs_min(max_length, (size_t)(4.1 * (double)UCS_GBYTE));
 
+    /* Trim by BAR1 size if relevant */
+    if (mem_type == UCS_MEMORY_TYPE_CUDA ||
+        mem_type == UCS_MEMORY_TYPE_CUDA_MANAGED) {
```
Akshay-Venkatesh, line 187: @shamisp On most systems, we can't register cuda_managed memory, so maybe we should be returning 0 for get_bar1_free_size() for managed memory?

test/gtest/common/mem_buffer.cc
```diff
@@ -145,6 +149,37 @@ void mem_buffer::set_device_context()
...
+        return max_availible_size;
+    }
+
+    ret = nvmlDeviceGetBAR1MemoryInfo(device, &bar1mem);
+    if (ret != NVML_SUCCESS) {
+        /* Similarly let's assume there is no limit on the size */
+        return max_availible_size;
+    }
+
+    max_availible_size = (size_t)bar1mem.bar1Free;
```
yosefe, line 177: IMO because of this line need to rename max_availible_size to availible_size
also, can we cache this value in a static (global) variable?

test/gtest/uct/uct_p2p_test.cc
yosefe: pls use ( ) around inner == expressions

test/gtest/uct/uct_p2p_test.cc
```diff
@@ -182,6 +182,13 @@ void uct_p2p_test::test_xfer_multi_mem_type(send_func_t send, size_t min_length,
...
+        max_length = ucs_min(max_length,
+                (size_t)(0.8 * mem_buffer::get_bar1_free_size()));
```
yosefe, line 189: maybe we need to multiply BAR size by 0.4 (not 0.8) since xfer tests have both send and recv buffer?
I wonder if this really worked on a small BAR card?

test/gtest/common/mem_buffer.cc
shamisp: I have not seen failures on my machine with small bar1 but I do have to some other failures around asymetic connectivity.

test/gtest/uct/uct_p2p_test.cc
shamisp: I have not seen failures on my machine with small bar1 but I do have to some other failures around asymetic connectivity.

yosefe (approved):

brminich (approved):

PR #8748, opened 2022-12-06
CONFIG: Removing failt-injector option

shamisp:
## Why

The option has never been implemented. There is no reason to have it in configure. This can also mislead developers.


yosefe (approved):

yosefe:
@shamisp pls squash

shamisp:
The failure looks irrelevant

PR #8751, opened 2022-12-08
UCT/CUDA_IPC: Improve overhead estimation

yosefe:
## Why
Make gdrcopy path preferred over rndv/get/zcopy with cuda_ipc for intra-node small messages when using protocols v2

## How
Increase estimated overhead of cuda_ipc transfer - make it closer to actual performance.


yosefe:
cc @Akshay-Venkatesh

shamisp:
since this impacts performance, it would be helpful to attach a plot (ucx_perf) showing the impact.

Akshay-Venkatesh (approved):

Akshay-Venkatesh:
@yosefe I've approved code changes for now. I'm still getting perf results on Selene and will post differences if there are any.

yosefe:
> since this impacts performance, it would be helpful to attach a plot (ucx_perf) showing the impact.

@shamisp the impact is that gdrcopy (eager) is now selected for intra-node small message instead of cuda_ipc (rndv)

Issue #8746, opened 2022-12-06
Misleading ibv_create_cq error message on node without Infiniband


gkatev:
Originally posted as comment in #8684, split out to new issue as it was unrelated. FYI @ivankochin

---

Hi, I'm seeing encountering the following error message:
```
ib_log.c:246  UCX  ERROR mlx5_0: ibv_create_cq() failed: Cannot allocate memory. Please set max locked memory (ulimit -l) to 'unlimited' (current: 64 kbytes)
```

I say it is misleading because the node does not have Infiniband. Furthermore, it happens with OpenMPI (compiled with UCX support), but even when I'm not actually selecting UCX. I'm using OpenMPI 5.0.0rc8 and UCX 1.13.1.

```
$ mpirun -n 2 --mca pml ob1 --mca btl sm,self --mca smsc xpmem osu_bcast -m 1
[1668693503.473099] [ampere:1188443:0]          ib_log.c:246  UCX  ERROR mlx5_0: ibv_create_cq() failed: Cannot allocate memory. Please set max locked memory (ulimit -l) to 'unlimited' (current: 64 kbytes)
[1668693503.473424] [ampere:1188442:0]          ib_log.c:246  UCX  ERROR mlx5_0: ibv_create_cq() failed: Cannot allocate memory. Please set max locked memory (ulimit -l) to 'unlimited' (current: 64 kbytes)

# OSU MPI Broadcast Latency Test v5.8
# Size       Avg Latency(us)
1                       0.23
```
```
$ ibstat
CA 'mlx5_0'
	CA type: MT4115
	Number of ports: 1
	Firmware version: 12.28.2006
	Hardware version: 0
	Node GUID: 0x98039b03009283de
	System image GUID: 0x98039b03009283de
	Port 1:
		State: Down
		Physical state: Disabled
		Rate: 40
		Base lid: 0
		LMC: 0
		SM lid: 0
		Capability mask: 0x00010000
		Port GUID: 0x9a039bfffe9283de
		Link layer: Ethernet

$ ucx_info -d
#
# Memory domain: self
#     Component: self
#             register: unlimited, cost: 0 nsec
#           remote key: 0 bytes
#
#      Transport: self
#         Device: memory0
#           Type: loopback
#  System device: <unknown>
#
#      capabilities:
#            bandwidth: 0.00/ppn + 6911.00 MB/sec
#              latency: 0 nsec
#             overhead: 10 nsec
#            put_short: <= 4294967295
[...]

PR #8757, opened 2022-12-09
UCT/IB: Make memlock limit settings non-fatal for md_open

ivankochin:
## What
Removes hard errors in case of too strict memlock limit during md_open

## Why ?
Now if system has too strict memlock limit the md_open() will fail and error messages would be printed. That can confuse customers if they don't plan to use IB transport at all.

Should solve https://github.com/openucx/ucx/issues/8746

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5_dv.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5_dv.c
@@ -54,10 +54,12 @@ void uct_ib_mlx5dv_dct_qp_init_attr(uct_ib_qp_init_attr_t *qp_attr,
 
 ucs_status_t
 uct_ib_mlx5dv_qp_tmp_objs_create(uct_ib_device_t *dev, struct ibv_pd *pd,
-                                 uct_ib_mlx5dv_qp_tmp_objs_t *qp_tmp_objs)
+                                 uct_ib_mlx5dv_qp_tmp_objs_t *qp_tmp_objs,
+                                 int silent)
 {
     struct ibv_srq_init_attr srq_attr = {};
-    ucs_status_t status;
+    ucs_log_level_t level             = silent ? UCS_LOG_LEVEL_DEBUG :
+                                                 UCS_LOG_LEVEL_ERROR;
     int cq_errno;
     char message[128];
 
@@ -66,8 +68,7 @@ uct_ib_mlx5dv_qp_tmp_objs_create(uct_ib_device_t *dev, struct ibv_pd *pd,
         cq_errno = errno;
         ucs_snprintf_safe(message, sizeof(message), "%s: ibv_create_cq()",
                           uct_ib_device_name(dev));
-        uct_ib_mem_lock_limit_msg(message, cq_errno, UCS_LOG_LEVEL_ERROR);
-        status = UCS_ERR_IO_ERROR;
+        uct_ib_mem_lock_limit_msg(message, cq_errno, level);
         goto out;
     }
 
@@ -75,8 +76,8 @@ uct_ib_mlx5dv_qp_tmp_objs_create(uct_ib_device_t *dev, struct ibv_pd *pd,
     srq_attr.attr.max_wr  = 1;
     qp_tmp_objs->srq      = ibv_create_srq(pd, &srq_attr);
     if (qp_tmp_objs->srq == NULL) {
-        ucs_error("%s: ibv_create_srq() failed: %m", uct_ib_device_name(dev));
-        status = UCS_ERR_IO_ERROR;
+        ucs_log(level, "%s: ibv_create_srq() failed: %m",
+                uct_ib_device_name(dev));
         goto out_destroy_cq;
     }
 
@@ -85,7 +86,7 @@ uct_ib_mlx5dv_qp_tmp_objs_create(uct_ib_device_t *dev, struct ibv_pd *pd,
 out_destroy_cq:
     ibv_destroy_cq(qp_tmp_objs->cq);
 out:
-    return status;
+    return UCS_ERR_IO_ERROR;
 }
 
 void uct_ib_mlx5dv_qp_tmp_objs_destroy(uct_ib_mlx5dv_qp_tmp_objs_t *qp_tmp_objs)
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -1585,9 +1585,8 @@ UCT_IB_MD_DEFINE_ENTRY(devx, uct_ib_mlx5_devx_md_ops);
 
 #endif
 
-static ucs_status_t uct_ib_mlx5dv_check_dc(uct_ib_device_t *dev)
+static void uct_ib_mlx5dv_check_dc(uct_ib_device_t *dev)
 {
-    ucs_status_t status = UCS_OK;
 #if HAVE_DC_DV
     struct ibv_context *ctx            = dev->ibv_context;
     uct_ib_qp_init_attr_t qp_init_attr = {};
@@ -1596,16 +1595,16 @@ static ucs_status_t uct_ib_mlx5dv_check_dc(uct_ib_device_t *dev)
     uct_ib_mlx5dv_qp_tmp_objs_t qp_tmp_objs;
     struct ibv_pd *pd;
     struct ibv_qp *qp;
+    ucs_status_t status;
     int ret;
 
     pd = ibv_alloc_pd(ctx);
     if (pd == NULL) {
-        ucs_error("%s: ibv_alloc_pd() failed: %m", uct_ib_device_name(dev));
-        status = UCS_ERR_IO_ERROR;
+        ucs_debug("%s: ibv_alloc_pd() failed: %m", uct_ib_device_name(dev));
         goto out;
     }
 
-    status = uct_ib_mlx5dv_qp_tmp_objs_create(dev, pd, &qp_tmp_objs);
+    status = uct_ib_mlx5dv_qp_tmp_objs_create(dev, pd, &qp_tmp_objs, 1);
     if (status != UCS_OK) {
         goto out_dealloc_pd;
     }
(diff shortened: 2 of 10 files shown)


shamisp:
@ivankochin when it becomes fatal and the warning gets printed ?

ivankochin:
> @ivankochin when it becomes fatal and the warning gets printed ?

If I correctly understand your question, then the answer is: it never becomes fatal and warnings are never printed.

All the warnings now printed as a debug information. 

`mlx5dv_md_open()` now returns `UCS_ERR_UNSUPPORTED` which allows to roll back to `verbs_md_open()`.
`verbs_md_open()` creates CQ and QP to check ECE and now inability to create them means just absence of ECE support, so the `md_open()` will still return `UCS_OK`.

I can change some parts of this PR, for example:
- Print the warnings and fail `verbs_md_open()` when reason of error is different then memlock limit
- Fail `mlx5dv_md_open()` when SRQ creation failed 
- Fail `verbs_md_open()` when QP creation failed

I want to discuss all these points together with you and @yosefe.

gleon99:
@ivankochin "That can confuse customers if they don't plan to use IB transport at all." but we can't know that ahead of time.
Meaning, we will just silently ignore in any case?

ivankochin:
> @ivankochin "That can confuse customers if they don't plan to use IB transport at all." but we can't know that ahead of time. Meaning, we will just silently ignore in any case?

Yes, if we are talking about `md_open()`. Further operations with IB transport would still fail. But if we want to define some point of memlock limit checking, I think we can add it. For example as an additional step of `md_query()`. WDYT?

shamisp:
@ivankochin this looks like another variant of https://github.com/openucx/ucx/pull/8715 which I believe was trigger by case with non-Infiniband Nic with partially supported Verbs API ?

If you have mlx5 NIC (or any other verbs compliant NIC that requires memory pinning) with a port in active state, it is beneficial to print out the warning rather than silently disable the device. In 99.9% cases this is a real problem.

ivankochin:
> @ivankochin this looks like another variant of #8715 which I believe was trigger by case with non-Infiniband Nic with partially supported Verbs API ?

Yes, your are right, but this PR was triggered by the case with RoCE NIC withour active ports. The difference is that patch doesn't silently suppress all information about IB/RoCE transports. It just postpones the error until the first try to create some structure that requires locked memory.

So users who use IB will just see the error on the later stage of their program. Users who don't use IB won't see IB related warnings or errors at all.

> If you have mlx5 NIC (or any other verbs compliant NIC that requires memory pinning) with a port in active state, it is beneficial to print out the warning rather than silently disable the device. In 99.9% cases this is a real problem.

I agree with that.

yosefe:
@shamisp this PR does not suppress memlock limit error during transport initialization. It only suppress the error from creating temp objects during md_open, so that we could query ports state and ignore inactive ports.

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -865,7 +865,7 @@ uct_ib_mlx5_md_buf_alloc(uct_ib_mlx5_md_t *md, size_t size, int silent,
     mem->mem  = mlx5dv_devx_umem_reg(md->super.dev.ibv_context, buf, size,
                                      access_mode);
     if (mem->mem == NULL) {
-        ucs_log(level, "mlx5dv_devx_umem_reg() failed: %m");
+        uct_ib_mem_lock_limit_msg("mlx5dv_devx_umem_reg()", errno, level);
```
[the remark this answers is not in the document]
ivankochin:
```suggestion
        ucs_log(level, "mlx5dv_devx_umem_reg() failed: %m");
```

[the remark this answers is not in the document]
yosefe: yep, pls revert if it's not critical

src/uct/ib/mlx5/dv/ib_mlx5_dv.c
```diff
@@ -54,10 +54,14 @@ void uct_ib_mlx5dv_dct_qp_init_attr(uct_ib_qp_init_attr_t *qp_attr,
...
-                                 uct_ib_mlx5dv_qp_tmp_objs_t *qp_tmp_objs)
+                                 uct_ib_mlx5dv_qp_tmp_objs_t *qp_tmp_objs,
+                                 int silent)
 {
     struct ibv_srq_init_attr srq_attr = {};
-    ucs_status_t status;
+    ucs_log_level_t level             = silent ? UCS_LOG_LEVEL_DEBUG :
+                                                 UCS_LOG_LEVEL_ERROR;
+    ucs_status_t error_status         = silent ? UCS_ERR_UNSUPPORTED :
+                                                 UCS_ERR_IO_ERROR;
```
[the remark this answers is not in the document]
yosefe: can we do this logic outside this function?

shamisp:
When the memlimit error will be reported ? Can you please share the complete solution ? 

If device has any active ports we must report the warning. So, my question is when do we report it ? thanks !

ivankochin:
> When the memlimit error will be reported ? Can you please share the complete solution ?
> 
> If device has any active ports we must report the warning. So, my question is when do we report it ? thanks !

We already have several places that report this error during IB ifaces initialization. E.g. after [CQs](https://github.com/openucx/ucx/blob/master/src/uct/ib/mlx5/ib_mlx5.c#L131) or [SRQs](https://github.com/openucx/ucx/blob/master/src/uct/ib/rc/base/rc_iface.c#L508) creation.

But we still need to check all the IB related objects creation places that can be affected by memlock limit. I think it would be better to cover all these places in a separate PR.

Pls contact me if you still have questions.

yosefe:
> When the memlimit error will be reported ? Can you please share the complete solution ?
> 
> If device has any active ports we must report the warning. So, my question is when do we report it ? thanks !

@shamisp when have active ports, UCP will create uct_iface_t objects during ucp_worker_create(), and these will fail to create CQ and report the error.

shamisp:
I discussed offline with @ivankochin - looks good !

shamisp (dismissed):

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1605,8 +1605,10 @@ static ucs_status_t uct_ib_mlx5dv_check_dc(uct_ib_device_t *dev)
         goto out;
     }
 
-    status = uct_ib_mlx5dv_qp_tmp_objs_create(dev, pd, &qp_tmp_objs);
+    status = uct_ib_mlx5dv_qp_tmp_objs_create(dev, pd, &qp_tmp_objs, 1);
     if (status != UCS_OK) {
+        /* Change status to allow other md_open() versions*/
```
yosefe, line 1610: See other places in the function - we need to set status==UCS_OK.
just by jumping to out_dealloc_pd, we skip `dev->flags |= UCT_IB_DEVICE_FLAG_DC`

ivankochin: It's not clear for me, why do we want to set the `UCT_IB_DEVICE_FLAG_DC` flag when we have too low memlock limit.

yosefe: we could set status==UCS_OK (instead of UNSUPPORTED) and jump to out_dealloc_pd
so UCT_IB_DEVICE_FLAG_DC  will **not** be set

ivankochin: So the only place where we return error from this function is that:
```c
    pd = ibv_alloc_pd(ctx);
    if (pd == NULL) {
        ucs_error("%s: ibv_alloc_pd() failed: %m", uct_ib_device_name(dev));
        status = UCS_ERR_IO_ERROR;
        goto out;
    }
```
Can we just skip DC support in that case too and print the message on debug level? If so, then we can make `check_dc` function returning void.

yosefe: > Can we just skip DC support in that case too and print the message on debug level? If so, then we can make `check_dc` function returning void.

yes

test/gtest/uct/test_md.cc
```diff
@@ -912,3 +913,38 @@ UCS_TEST_SKIP_COND_P(test_md_fork, fork,
...
+    struct rlimit previous_limit;
+};
+
+UCS_TEST_P(test_md_memlock_limit, md_open) {
+    UCS_TEST_CREATE_HANDLE(uct_md_h, m_md, uct_md_close, uct_md_open,
+                           GetParam().component, GetParam().md_name.c_str(),
+                           m_md_config);
+}
+
+UCT_MD_INSTANTIATE_TEST_CASE(test_md_memlock_limit)
```
yosefe, line 950: add newline

test/gtest/uct/test_md.cc
```diff
@@ -912,3 +913,38 @@ UCS_TEST_SKIP_COND_P(test_md_fork, fork,
...
+            UCS_TEST_ABORT("Failed to restore memlock limit. "
+                           "Can cause other tests failures!");
+        }
+        test_md::cleanup();
+    }
+private:
+    struct rlimit previous_limit;
+};
+
+UCS_TEST_P(test_md_memlock_limit, md_open) {
```
yosefe, line 944: { on next line

test/gtest/uct/test_md.cc
```diff
@@ -912,3 +913,38 @@ UCS_TEST_SKIP_COND_P(test_md_fork, fork,
 
 UCT_MD_INSTANTIATE_TEST_CASE(test_md_fork)
 
+
+struct test_md_memlock_limit : public test_md {
```
yosefe, line 917: class

test/gtest/uct/test_md.cc
```diff
@@ -912,3 +913,38 @@ UCS_TEST_SKIP_COND_P(test_md_fork, fork,
...
+
+    void cleanup()
+    {
+        if (setrlimit(RLIMIT_MEMLOCK, &previous_limit) != 0) {
+            UCS_TEST_ABORT("Failed to restore memlock limit. "
+                           "Can cause other tests failures!");
+        }
+        test_md::cleanup();
+    }
+private:
```
yosefe, line 940: space line before

test/gtest/uct/test_md.cc
```diff
@@ -912,3 +913,38 @@ UCS_TEST_SKIP_COND_P(test_md_fork, fork,
...
+    struct rlimit previous_limit;
```
yosefe, line 941: m_previous_limit

yosefe (dismissed):

ivankochin:
squashed. @yosefe pls reapprove

gleon99 (approved):

PR #8760, opened 2022-12-09
UCT/CUDA_IPC: add a simplistic impl of estimate_perf

Akshay-Venkatesh:
## What/Why?
Use peak pcie or nvlink bandwidth for cuda_ipc perf estimates to be used by protov2. 

### TODO in a future PR
With this implemenation, UCP needs to rely on rkey_unpack(IpcOpenMemhandle) failing to fallback to non-cuda-ipc path. In the future we could improve this implementation further to potentially allow protov2 to make the decision to not use cuda-ipc rma_bw lane if cuda_ipc_iface_estimate_perf could process rkey and return bandwidth 0 if remote GPU is not peer-reachable.

### DIFF
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
@@ -473,12 +473,70 @@ ucs_status_t uct_cuda_ipc_iface_init_streams(uct_cuda_ipc_iface_t *iface)
     return UCS_OK;
 }
 
+static ucs_status_t
+uct_cuda_ipc_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr)
+{
+    static const double latency  = 1.8e-6;
+    static const double overhead = 4.0e-6;
+
+    perf_attr->bandwidth.dedicated = 0;
+
+    if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_BANDWIDTH) {
+
+        /* TODO:
+         *      1. differentiate read vs write perf
+         *      2. check if src,dst memory is cuda type;
+         *         (currently UCP does not pass mem types so cuda_ipc would
+         *          report zero bandwidth)
+         *      3. Check nvlinks and report bandwidth;
+         *         (for now not checking because we do not want to report zero
+         *          bandwidth for PCIe connected IPC-accessible devices)
+         */
+
+        perf_attr->bandwidth.shared = uct_cuda_ipc_iface_get_bw();
+    }
+
+    if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_SEND_PRE_OVERHEAD) {
+        perf_attr->send_pre_overhead = overhead;
+    }
+
+    if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_SEND_POST_OVERHEAD) {
+        /* In case of sync mem copy, the send operation CPU overhead includes
+           the latency of waiting for the copy to complete */
+        perf_attr->send_post_overhead = 0;
+    }
+
+    if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_RECV_OVERHEAD) {
+        perf_attr->recv_overhead = 0;
+    }
+
+    if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_LATENCY) {
+        /* In case of async mem copy, the latency is not part of the overhead
+           and it's a standalone property */
+        perf_attr->latency = ucs_linear_func_make(latency, 0.0);
+    }
+
+    if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_MAX_INFLIGHT_EPS) {
+        perf_attr->max_inflight_eps = SIZE_MAX;
+    }
+
+    return UCS_OK;
+}
+
 static ucs_mpool_ops_t uct_cuda_ipc_event_desc_mpool_ops = {
     .chunk_alloc   = ucs_mpool_chunk_malloc,
     .chunk_release = ucs_mpool_chunk_free,
     .obj_init      = uct_cuda_ipc_event_desc_init,
     .obj_cleanup   = uct_cuda_ipc_event_desc_cleanup,
     .obj_str       = NULL
+};
+
+static uct_iface_internal_ops_t uct_cuda_ipc_iface_internal_ops = {
+    .iface_estimate_perf = uct_cuda_ipc_estimate_perf,
+    .iface_vfs_refresh   = (uct_iface_vfs_refresh_func_t)ucs_empty_function,
+    .ep_query            = (uct_ep_query_func_t)ucs_empty_function_return_unsupported,
+    .ep_invalidate       = (uct_ep_invalidate_func_t)ucs_empty_function_return_unsupported,
+    .ep_connect_to_ep_v2 = ucs_empty_function_return_unsupported
 };
 
 static UCS_CLASS_INIT_FUNC(uct_cuda_ipc_iface_t, uct_md_h md, uct_worker_h worker,
@@ -491,7 +549,7 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_ipc_iface_t, uct_md_h md, uct_worker_h worke
 
     config = ucs_derived_of(tl_config, uct_cuda_ipc_iface_config_t);
     UCS_CLASS_CALL_SUPER_INIT(uct_base_iface_t, &uct_cuda_ipc_iface_ops,
-                              &uct_base_iface_internal_ops, md, worker, params,
+                              &uct_cuda_ipc_iface_internal_ops, md, worker, params,
                               tl_config UCS_STATS_ARG(params->stats_root)
                               UCS_STATS_ARG("cuda_ipc"));
 


Akshay-Venkatesh:
cc @yosefe

Akshay-Venkatesh:
@yosefe FYI this PR forces protov2 to view cuda_ipc bandwidth only as pcie bandwidth and not nvlink bandwidth based on local tests. It seems that get_lane_perf function doesn't populate src/dst memory type [here](https://github.com/openucx/ucx/blob/master/src/ucp/proto/proto_common.c#L302). Before a rendezvous control message is sent or after it is received, are protocols re-evaluated because we have both local and remote memory types? If yes, can you point to the code?

Akshay-Venkatesh:
> @Akshay-Venkatesh How do we know if the device in NVLINK or PCIe GDR ? thanks

This PR checks [here](https://github.com/openucx/ucx/pull/8760/files#diff-04769c9662cb2e769aff68bed88f18cef6d6bdaa369b178d4223d769cecb7d4fR506) but also [here](https://github.com/openucx/ucx/blob/master/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c#L139) to limit get_zcopy capability of cuda_ipc for specific architectures.

Akshay-Venkatesh:
I was unable to repro the errors:

```
$ UCX_PROTO_ENABLE=y taskset -c 10 ./examples/ucp_hello_world -m cuda -p 16519 &
[1] 15034
INFO: UCP_HELLO_WORLD mode = 0 server = (null) port = 16519, pid = 15034
[0x27f3000] local address length: 920
Waiting for connection...

$ UCX_PROTO_ENABLE=y taskset -c 11 ./examples/ucp_hello_world -m cuda -n localhost -p 16519
INFO: UCP_HELLO_WORLD mode = 0 server = localhost port = 16519, pid = 15075
[0x4d90d000] local address length: 920
finish to send UCX address message
[0x27f3000] receive handler called with status 0 (Success), length 928
finish to receive UCX address message
finish to send UCX data message
flush_ep completed with status 0 (Success)
[0x4d90d000] receive handler called with status 0 (Success), length 24
finish to receive UCX data message


----- UCP TEST SUCCESS ----

ABCDEFGHIJKLMNO

---------------------------
```

and I also couldn't find `install/share/ucx/perftest/test_types_short_ucp` to repro the other error.

src/uct/cuda/base/cuda_md.c
```diff
@@ -67,6 +67,42 @@ ucs_status_t uct_cuda_base_get_sys_dev(CUdevice cuda_device,
...
+double uct_cuda_base_get_pci_bw(void *address)
+{
+    char sysfs_path[PATH_MAX];
+    char device_name[10];
+    ucs_sys_device_t sys_dev;
+    CUdevice cuda_device;
+    CUresult cu_err;
+    ucs_status_t status;
+
+    if (address != NULL) {
```
yosefe, line 79: when address is not NULL?

src/uct/cuda/base/cuda_md.c
```diff
@@ -67,6 +67,42 @@ ucs_status_t uct_cuda_base_get_sys_dev(CUdevice cuda_device,
...
+
+    status = ucs_topo_get_sysfs_path(sys_dev, sysfs_path, PATH_MAX);
+    if (status != UCS_OK) {
+        return 0;
+    }
+
+    ucs_snprintf_safe(device_name, sizeof(device_name), "GPU%d",
+                      cuda_device);
+
+    return ucs_topo_get_pci_bw(device_name, sysfs_path);
```
yosefe, line 103: maybe expose function ucs_topo_get_pci_bw(device_name, sys_dev) ?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -467,6 +471,73 @@ ucs_status_t uct_cuda_ipc_iface_init_streams(uct_cuda_ipc_iface_t *iface)
...
+                                                    UCS_MEMORY_TYPE_UNKNOWN);
+    int zcopy                      = uct_ep_op_is_zcopy(op);
+    const double latency           = 1.8e-6;
+    const double overhead          = 4.0e-6;
+    int num_nvlinks;
+
+    if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_BANDWIDTH) {
+        if (uct_ep_op_is_fetch(op)) {
+            /* TODO: differentiate read vs write perf */
+        }
```
yosefe, line 495: remove

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -467,6 +471,73 @@ ucs_status_t uct_cuda_ipc_iface_init_streams(uct_cuda_ipc_iface_t *iface)
...
+
+        perf_attr->bandwidth.dedicated = 0;
+        perf_attr->bandwidth.shared    = uct_cuda_base_get_pci_bw(NULL);
+
+        if ((src_mem_type == UCS_MEMORY_TYPE_CUDA) &&
+            (dst_mem_type == UCS_MEMORY_TYPE_CUDA)) {
```
yosefe, line 501: could it be if/else? no need to get pci bw for cuda/cuda memory
also, can we cache the pci bw on cuda md?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -467,6 +471,73 @@ ucs_status_t uct_cuda_ipc_iface_init_streams(uct_cuda_ipc_iface_t *iface)
...
+                                                    OPERATION, UCT_EP_OP_LAST);
+    ucs_memory_type_t src_mem_type = UCT_ATTR_VALUE(PERF, perf_attr,
+                                                    local_memory_type,
+                                                    LOCAL_MEMORY_TYPE,
+                                                    UCS_MEMORY_TYPE_UNKNOWN);
+    ucs_memory_type_t dst_mem_type = UCT_ATTR_VALUE(PERF, perf_attr,
+                                                    remote_memory_type,
+                                                    REMOTE_MEMORY_TYPE,
+                                                    UCS_MEMORY_TYPE_UNKNOWN);
+    int zcopy                      = uct_ep_op_is_zcopy(op);
```
yosefe, line 487: zcopy should always be 1 for cuda_ipc

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -499,6 +578,7 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_ipc_iface_t, uct_md_h md, uct_worker_h worke
     self->config.max_streams         = config->max_streams;
     self->config.enable_cache        = config->enable_cache;
     self->config.enable_get_zcopy    = config->enable_get_zcopy;
+    self->config.bandwidth           = config->bandwidth;
```
yosefe, line 581: where is it used?

yosefe:
@Akshay-Venkatesh The failures are consistent, seems the change in perf estimation uncovered some issue with protov2

Akshay-Venkatesh:
@yosefe Tests are passing after the logic was simplified even more. Can you provide a review when possible?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -467,6 +467,58 @@ ucs_status_t uct_cuda_ipc_iface_init_streams(uct_cuda_ipc_iface_t *iface)
...
+    const double overhead = 4.0e-6;
+
+    perf_attr->bandwidth.dedicated = 0;
+
+    if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_BANDWIDTH) {
+
+        /* TODO:
+         *      1. differentiate read vs write perf
+         *      2. check if src,dst memory is cuda type;
+         *         (currently UCP does not pass mem types so cuda_ipc would
```
yosefe, line 483: it does in protov2:
https://github.com/openucx/ucx/blob/4f8240871d11bc53bbb609263d32d0127be0dc57/src/ucp/proto/proto_init.c#L424

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -467,6 +467,58 @@ ucs_status_t uct_cuda_ipc_iface_init_streams(uct_cuda_ipc_iface_t *iface)
...
+         *          report zero bandwidth)
+         *      3. Check nvlinks and report bandwidth;
+         *         (for now not checking because we do not want to report zero
+         *          bandwidth for PCIe connected IPC-accessible devices)
+         */
+
+        perf_attr->bandwidth.shared = uct_cuda_ipc_iface_get_bw();
+    }
+
+    ucs_debug("cuda_ipc perf_attr.shared b/w = %lf", perf_attr->bandwidth.shared);
```
yosefe, line 493: remove

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -467,6 +467,58 @@ ucs_status_t uct_cuda_ipc_iface_init_streams(uct_cuda_ipc_iface_t *iface)
     return UCS_OK;
 }
 
+static ucs_status_t
+uct_cuda_ipc_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr)
+{
+    const double latency  = 1.8e-6;
+    const double overhead = 4.0e-6;
```
yosefe, line 474: static?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
Akshay-Venkatesh: @yosefe This seems to be true only for proto_init_buffer_copy_time function. Is this function used for RMA transfers or just local memtype_ep transfers (like pack/unpack)?

 I noticed that cuda-ipc perf estimate (most likely from zcopy estimation_perf logic) didn't provide memory types and cuda_ipc_estimate_perf was returning default performance. Examples include:
1. https://github.com/openucx/ucx/blob/master/src/ucp/core/ucp_ep.c#L1976
2. https://github.com/openucx/ucx/blob/master/src/ucp/proto/proto_common.c#L302
3. https://github.com/openucx/ucx/blob/master/src/ucp/rndv/proto_rndv.c#L211

yosefe: right, it's missing for the actual transfer which is relevant for cuda-ipc - in ucp_proto_common_get_lane_perf
place (1) is protov1 and place (3) is rndv control message that is always host mem.

yosefe (dismissed):
pls squash

yosefe (approved):

PR #8763, opened 2022-12-12
CUDA: make cuGetProcAddress to work with cuda 12

shamisp:
## What
CUDA: Adding support for CUDA-12

### DIFF
--- a/src/uct/cuda/base/cuda_md.c
+++ b/src/uct/cuda/base/cuda_md.c
@@ -231,6 +231,17 @@ static int uct_cuda_copy_get_dmabuf_fd(const ucs_memory_info_t *addr_mem_info)
     /* Get fxn ptr for cuMemGetHandleForAddressRange in case installed libcuda
      * does not have the definition for it even though 11.7 header includes the
      * declaration and avoid link error */
+#if CUDA_VERSION >= 12000
+    CUdriverProcAddressQueryResult proc_addr_res;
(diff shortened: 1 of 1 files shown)


shamisp:
@Akshay-Venkatesh can you please take a look ? Thanks

Akshay-Venkatesh (dismissed):
@shamisp changes look fine but we probably ought to consolidate code which gets split based on version. There are other instances similar to these where fxn signature changes but we should probably have a common function so that caller doesn't see all the ugly bits.

shamisp:
@Akshay-Venkatesh In a longer term this makes sense but it is much more substantial overhaul of CUDA UCT. Majority of the ifdef flows were introduced in 11070 and I didn't want to change this.

Akshay-Venkatesh:
> @Akshay-Venkatesh In a longer term this makes sense but it is much more substantial overhaul of CUDA UCT. Majority of the ifdef flows were introduced in 11070 and I didn't want to change this.

Agreed that we should not make that overhaul in this PR.

Minor nitpick: the title of this PR is deceiving. Shouldn't it just be "make cuGetProcAddress work with cuda 12"?

brminich (dismissed):

yosefe (dismissed):

yosefe:
@shamisp can you pls squash?

shamisp:
@yosefe DONE. The failures are unrelated: `fatal error: error writing to /tmp/ccn6FDdP.s: No space left on device`

PR #8764, opened 2022-12-12
AZP: Add ubuntu20 container w/ rocm software stack

edgargabriel:
## What
add a rocm docker container for testing. Build only at the moment.

## Why ?
simplify code development and maintenance for rocm devices

### DIFF
--- a/buildlib/tools/builds.sh
+++ b/buildlib/tools/builds.sh
@@ -241,6 +241,19 @@ build_cuda() {
 }
 
 #
+# Build ROCm
+#
+build_rocm() {
+	if [ -f /opt/rocm/bin/rocminfo ]; then
+		echo "==== Build with enable rocm  ===="
+		${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst --with-rocm
+		$MAKEP
+	else
+		echo "==== Not building with rocm ===="
+	fi
+}
+
+#
 # Build with clang compiler
 #
 build_clang() {
@@ -419,6 +432,7 @@ do_task "${prog}" build_prof
 do_task "${prog}" build_ugni
 do_task "${prog}" build_disable_numa
 do_task "${prog}" build_cuda
+do_task "${prog}" build_rocm
 do_task "${prog}" build_no_verbs
 do_task "${prog}" build_release_pkg
 do_task "${prog}" build_cmake_examples
--- a/buildlib/pr/main.yml
+++ b/buildlib/pr/main.yml
@@ -151,6 +151,8 @@ resources:
     - container: ubuntu20_cuda11
       image: rdmz-harbor.rdmz.labs.mlnx/ucx/ubuntu20.04-mofed5-cuda11:1
       options: $(DOCKER_OPT_VOLUMES) $(DOCKER_OPT_GPU)
+    - container: ubuntu2004_rocm_5_4_0
+      image: registry.hub.docker.com/rocm/ucx:rocm-5.4.0
 
 stages:
   - stage: Codestyle
@@ -204,6 +206,8 @@ stages:
               long_test: yes
             centos7:
               CONTAINER: centos7_ib
+            ubuntu2004_rocm:
+              CONTAINER: ubuntu2004_rocm_5_4_0
         container: $[ variables['CONTAINER'] ]
         timeoutInMinutes: 340
 


edgargabriel:
The 'enable-debug' build actually finished correctly, it automatically picked up the rocm directory and the components were compiled correctly.

https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=55553&view=logs&j=b86ef40a-ec94-56c9-ded9-5d536bb3fe84&t=288623fd-d89b-56b5-3a57-ecb116883784&l=9508

The test failed however for the ugni build.  Trying to disect how that is supposed to work or whether we could disable the ugni builds for the rocm image.

edgargabriel:
Adding a side note here: it seems like ucx does not compile if both cuda and rocm are being found. Will investigate this potentially later, right now I just added a '--without-rocm' to the cuda builds, and a '--without-cuda' to the rocm build. The conflict stems from both rocm and cuda defining some datatypes with the same name (e.g. dim3), so the compiler complains about redefinition of these types.

buildlib/tools/builds.sh
```diff
@@ -191,21 +191,26 @@ build_prof() {
...
-		PKG_CONFIG_TOP_BUILD_DIR=${WORKSPACE}
-	$MAKEP
-
-	# make sure UGNI transport is enabled
-	grep '#define HAVE_TL_UGNI 1' config.h
-
-	$MAKEP distcheck
+        # Disable building ugni for the rocm docker image.
+        if [ -f /opt/rocm/bin/rocminfo ]; then
+                  azure_log_warning "Not building with cray-ugni"
```
yosefe, line 196: why?

buildlib/tools/builds.sh
```diff
@@ -236,10 +241,23 @@ build_cuda() {
 		echo "==== Running test_link_map with cuda build but no cuda module ===="
 		env UCX_HANDLE_ERRORS=bt ./test/apps/test_link_map
 	else
-		echo "==== Not building with cuda flags ===="
+                echo "==== Not building with cuda flags ===="
```
yosefe, line 244: why changed?

buildlib/tools/builds.sh
```diff
@@ -236,10 +241,23 @@ build_cuda() {
...
+#
+# Build ROCm
+#
+build_rocm() {
+        if [ -f /opt/rocm/bin/rocminfo ]; then
+                echo "==== Build with enable rocm  ===="
+                ${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst --with-rocm
+                $MAKEP
+        else
+                echo "==== Not building with rocm ===="
```
yosefe, line 257: use tabs instead of spaces

buildlib/tools/builds.sh
edgargabriel: that is a mistake, will fix. I had it briefly changed to an azure_log_warning for making debugging easier, since the cuda cases are now not compiled anymore in the rocm image. Will fix it.

buildlib/tools/builds.sh
edgargabriel: the ugni stuff doesn't compile on the rocm image (its missing xpmem), I wanted to find a way to exclude it from compiling it with the rocm image.

edgargabriel: is there a better way for detecting whether ugni can be compiled? Many of the other functions rely on a module load succeeding or not, but there is nothing comparable for ugni

yosefe: > the ugni stuff doesn't compile on the rocm image (its missing xpmem)

Can you pls post the error?

edgargabriel: yes, give me a min. I noticed that configure succeeds with the ugni option when trying to run it separately in the same docker image, so I will try to re-trigger the CI build to see whether it still fails (and remove at least temporarily the change in the script).

edgargabriel: well, it looks like its passing now in the CI as well, not entirely sure what change triggered the test to pass/fail. But I was able to undo the change to the build_ugni() function and it passes that test, so we should be good.

edgargabriel: actually, I know what the problem was. I misread the configure error early on, I though it was related to xpmem, it wasn't.  My first iterations of the docker image were missing pkg-config, and that was the reason the ugni configure step failed. I found the output file of a failed run:

```
configure: WARNING: KNEM requested but required file (knem_io.h) could not be found
configure: XPMEM - failed to open the requested location (guess), guessing ...
/__w/1/s/contrib/../configure: line 25057: pkg-config: command not found
checking cray-ugni... /__w/1/s/contrib/../configure: line 25191: pkg-config: command not found
```
I fixed the pkg-config issue when making the build_release_pgk() step work later

yosefe (approved):

PR #8769, opened 2022-12-14
UCT/IB/MLX5DV: separate testing of AMO and KSM caps

evgeny-leksikov:
## What
separate testing of AMO and KSM caps

## Why ?
they are not always needed together

### DIFF
--- a/src/uct/api/uct.h
+++ b/src/uct/api/uct.h
@@ -746,7 +746,10 @@ enum {
     UCT_MD_FLAG_SOCKADDR      = UCS_BIT(7),
 
     /**
-     * MD supports memory invalidation
+     * MD supports memory invalidation.
+     * @note This flag is equivalent to the combination of
+     *       UCT_MD_FLAG_INVALIDATE_RMA and UCT_MD_FLAG_INVALIDATE_AMO for
+     *       uct_md_attr_v2_t.flags
      */
     UCT_MD_FLAG_INVALIDATE    = UCS_BIT(8),
 
@@ -759,7 +762,13 @@ enum {
     /**
      * MD supports registering a dmabuf file descriptor.
      */
-    UCT_MD_FLAG_REG_DMABUF    = UCS_BIT(10)
+    UCT_MD_FLAG_REG_DMABUF    = UCS_BIT(10),
+
+    /**
+     * The enum must not be extended. Any additional flags must be defined in
+     * API v2 uct_md_flags_v2_t.
+     */
+    UCT_MD_FLAG_LAST          = UCS_BIT(11)
 };
 
 /**
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -274,23 +274,42 @@ typedef enum {
     /**
      * The flag is used indicate that remote access to a memory region
      * associated with the remote key must fail once the memory region is
-     * deregister using @ref uct_md_mem_dereg_v2 with
+     * deregistered using @ref uct_md_mem_dereg_v2 with
      * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set. Using
      * @ref uct_md_mem_dereg_v2 deregistration routine with
      * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set on an rkey that was
-     * generated without @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag will
-     * not function correctly, and may result in data corruption. In other words
-     * in order for @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag to function
-     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag must be set.
+     * generated without
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA and/or
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO flags will not function
+     * correctly, and may result in data corruption. In other words, in order
+     * for @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag to function
+     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA and/or
+     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO flag must be set.
      */
-    UCT_MD_MKEY_PACK_FLAG_INVALIDATE = UCS_BIT(0),
+    UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA = UCS_BIT(0),
+
+    /**
+     * The flag is used to indicate that the atomic operations must fail once
+     * the memory region is deregistered using @ref uct_md_mem_dereg_v2 with
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set.
+     * Using @ref uct_md_mem_dereg_v2 deregistration routine with
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set on an rkey that was
+     * generated without
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA and/or
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO flags will not function
+     * correctly, and may result in data corruption. In other words, in order
+     * for @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag to function
+     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA and/or
+     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO flag must be set.
+     */
+    UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO = UCS_BIT(1),
 
     /**
      * The flag is used to indicate that the memory region should be packed in
      * order to be accessed by another process using the same device to perform
      * UCT operations.
      */
-    UCT_MD_MKEY_PACK_FLAG_EXPORT     = UCS_BIT(1)
+    UCT_MD_MKEY_PACK_FLAG_EXPORT         = UCS_BIT(2)
 } uct_md_mkey_pack_flags_t;
 
 
@@ -377,7 +396,7 @@ typedef struct uct_md_mem_reg_params {
     uint64_t                     field_mask;
 
     /**
-     * Operation specific flags, using bits from @ref uct_md_mem_flags_t.
+     * Operation specific flags, using bits from @ref uct_md_mem_flags.
      */
     uint64_t                     flags;
 
@@ -773,6 +792,31 @@ typedef struct {
      */
     char              global_id[UCT_MD_GLOBAL_ID_MAX];
 } uct_md_attr_v2_t;
+
+
+/**
+ * @ingroup UCT_MD
+ * @brief  Memory domain capability flags.
+ */
+typedef enum {
+    UCT_MD_FLAG_V2_FIRST       = UCT_MD_FLAG_LAST,
+
+    /**
+     * Memory domain supports invalidation of memory handle registered by
+     * @ref uct_md_mem_reg_v2 with @ref UCT_MD_MEM_ACCESS_RMA flag and packed
+     * key by @ref uct_md_mkey_pack_v2 with
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag.
+     */
+    UCT_MD_FLAG_INVALIDATE_RMA = UCT_MD_FLAG_V2_FIRST,
+
+    /**
+     * Memory domain supports invalidation of memory handle registered by
+     * @ref uct_md_mem_reg_v2 with @ref UCT_MD_MEM_ACCESS_REMOTE_ATOMIC flag and
+     * packed key by @ref uct_md_mkey_pack_v2 with
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag.
+     */
+    UCT_MD_FLAG_INVALIDATE_AMO = UCS_BIT(12)
+} uct_md_flags_v2_t;
 
 
 /**
--- a/src/uct/base/uct_mem.c
+++ b/src/uct/base/uct_mem.c
@@ -386,7 +386,7 @@ ucs_status_t uct_iface_mem_alloc(uct_iface_h tl_iface, size_t length, unsigned f
                              UCT_MEM_ALLOC_PARAM_FIELD_MEM_TYPE |
                              UCT_MEM_ALLOC_PARAM_FIELD_MDS      |
                              UCT_MEM_ALLOC_PARAM_FIELD_NAME;
-    params.flags           = UCT_MD_MEM_ACCESS_ALL;
+    params.flags           = flags;
     params.name            = name;
     params.mem_type        = UCS_MEMORY_TYPE_HOST;
     params.address         = address;
@@ -452,7 +452,9 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_iface_mp_chunk_alloc,
 
     length = sizeof(*hdr) + *size_p;
     status = uct_iface_mem_alloc(&iface->super, length,
-                                 UCT_MD_MEM_ACCESS_ALL | UCT_MD_MEM_FLAG_LOCK,
+                                 UCT_MD_MEM_ACCESS_LOCAL_READ  |
+                                 UCT_MD_MEM_ACCESS_LOCAL_WRITE |
+                                 UCT_MD_MEM_FLAG_LOCK,
                                  ucs_mpool_name(mp), &mem);
     if (status != UCS_OK) {
         return status;
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -35,11 +35,8 @@
 
 #define UCT_IB_MD_MEM_DEREG_CHECK_PARAMS(_ib_md, _params) \
     do { \
-        ucs_status_t _status; \
-        \
-        UCT_MD_MEM_DEREG_CHECK_PARAMS( \
-                _params, (_ib_md)->cap_flags & UCT_MD_FLAG_INVALIDATE); \
-        _status = uct_ib_md_mem_dereg_params_invalidate_check(_params); \
+        ucs_status_t _status = \
+                uct_ib_md_mem_dereg_params_invalidate_check(_ib_md, _params); \
         if (_status != UCS_OK) { \
             return _status; \
         } \
@@ -759,6 +756,11 @@ uct_ib_mem_reg_internal(uct_md_h uct_md, void *address, size_t length,
         memh->flags |= UCT_IB_MEM_ACCESS_REMOTE_ATOMIC;
     }
 
+    if (params->flags & (UCT_MD_MEM_ACCESS_REMOTE_GET |
+                         UCT_MD_MEM_ACCESS_REMOTE_PUT)) {
+        memh->flags |= UCT_IB_MEM_ACCESS_REMOTE_RMA;
+    }
+
     status = uct_ib_md_reg_mr(md, address, length, access_flags,
                               params->dmabuf_fd, params->dmabuf_offset, silent,
                               memh, UCT_IB_MR_DEFAULT);
@@ -1003,6 +1005,10 @@ uct_ib_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
             ucs_trace("created atomic key 0x%x for 0x%x", memh->atomic_rkey,
                       memh->lkey);
         } else if (status == UCS_ERR_UNSUPPORTED) {
+            if (flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO) {
+                return UCS_ERR_IO_ERROR;
+            }
+
             /* ignore for atomic MR */
         } else {
             return status;
@@ -1016,7 +1022,8 @@ uct_ib_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
     }
 
     if (flags & UCT_MD_MKEY_PACK_FLAG_EXPORT) {
-        if (flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE) {
+        if (flags & (UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA |
+                     UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO)) {
             ucs_error("packing a memory key which should support invalidation"
                       "and exporting is unsupported");
             return UCS_ERR_UNSUPPORTED;
@@ -1030,7 +1037,7 @@ uct_ib_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
         }
 
         mkey = memh->exported_lkey;
-    } else if ((flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE) &&
+    } else if ((flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA) &&
                ((atomic_rkey != UCT_IB_INVALID_MKEY) ||
                 !(memh->flags & UCT_IB_MEM_ACCESS_REMOTE_ATOMIC))) {
         /**
--- a/src/uct/ib/base/ib_md.h
+++ b/src/uct/ib/base/ib_md.h
@@ -64,8 +64,14 @@ enum {
     UCT_IB_MEM_FLAG_RELAXED_ORDERING = UCS_BIT(4), /**< The memory region will issue
                                                         PCIe writes with relaxed order
                                                         attribute */
-    UCT_IB_MEM_FLAG_NO_RCACHE        = UCS_BIT(5)  /**< Memory handle wasn't stored
+    UCT_IB_MEM_FLAG_NO_RCACHE        = UCS_BIT(5), /**< Memory handle wasn't stored
                                                         in RCACHE */
+#if ENABLE_PARAMS_CHECK
+    UCT_IB_MEM_ACCESS_REMOTE_RMA     = UCS_BIT(6) /**< RMA access was requested
+                                                        for the memory region */
+#else
+    UCT_IB_MEM_ACCESS_REMOTE_RMA     = 0
+#endif
 };
 
 enum {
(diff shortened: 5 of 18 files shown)


src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -434,10 +434,10 @@ static void uct_ib_mlx5_devx_mr_lru_cleanup(uct_ib_mlx5_md_t *md)
 }
 
 static UCS_F_ALWAYS_INLINE int
-uct_ib_mlx5_devx_md_is_ksm_atomic_supported(uct_ib_mlx5_md_t *md)
+uct_ib_mlx5_devx_md_flags_test(uct_ib_mlx5_md_t *md, uint32_t flags)
 {
-    return ucs_test_all_flags(md->flags, UCT_IB_MLX5_MD_FLAG_KSM |
-                                         UCT_IB_MLX5_MD_FLAG_INDIRECT_ATOMICS);
+    UCS_STATIC_ASSERT(sizeof(md->flags) == sizeof(flags));
```
gleon99, line 439: This is basically a one-liner.
If we think it justifies a separate func, WDYT of sharing it between similar places which test MD flags, e.g in `dc_ml5x.c`, `rc_mlx5_iface.c`, etc? There's nothing particularly "devx" about it.

evgeny-leksikov: Let's see how it goes in CI then refactor, there are multiple options:
 - delete one-liner at all
 - reuse it everywhere to test `UCT_IB_MLX5_MD_FLAG_XXX`  flags
 - rename it to something related to KSM/INDIRECT_ATOMICS or wider set of capabilities for traces and/or debugging purposes on control path

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1254,7 +1258,7 @@ static ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
 
     memcpy(md->super.vhca_id, vhca_id, sizeof(vhca_id));
 
-    if (uct_ib_mlx5_devx_md_is_ksm_atomic_supported(md)) {
+    if (uct_ib_mlx5_devx_md_flags_test(md, UCT_IB_MLX5_MD_FLAG_KSM)) {
```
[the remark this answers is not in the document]
evgeny-leksikov: do you mean test AMO_FEATURE  from context?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
evgeny-leksikov: Initially I defined `flags` with wrong type. The idea was to avoid a bug with overflow since even inline function does typecast. Most likely `md->flags` type will be changed with adding new flags but it's easy to miss usage.
IMO it's minor but good to have.

yosefe:
go test failure seems related:
```
[swx-rdmz-ucx-new-01:16627:0:16627]       ib_md.c:1017 Assertion `md->cap_flags & UCT_MD_FLAG_INVALIDATE_AMO' failed

/scrap/azure/agent-01/AZP_WORKSPACE/1/s/contrib/../src/uct/ib/base/ib_md.c: [ uct_ib_mkey_pack() ]
      ...
     1014                       memh->lkey);
     1015         } else if (status == UCS_ERR_UNSUPPORTED) {
     1016             if (flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE) {
==>  1017                 ucs_assert(md->cap_flags & UCT_MD_FLAG_INVALIDATE_AMO);
```

src/uct/api/v2/uct_v2.h
```diff
@@ -775,6 +775,31 @@ typedef struct {
...
 
+/**
+ * @ingroup UCT_MD
+ * @brief  Memory domain capability flags.
+ */
+enum {
+    UCT_MD_FLAG_V2_FIRST       = UCT_MD_FLAG_LAST,
+
+    /**
+     * MD supports invalidation of memory handle registered with
```
yosefe, line 786: maybe we should phrase it like

MD supports invalidation of remote keys packed by ... for put/get operations

since we could register the key with several rights, but invalidate only for atomics and/or rma access

src/uct/base/uct_mem.c
```diff
@@ -386,7 +386,7 @@ ucs_status_t uct_iface_mem_alloc(uct_iface_h tl_iface, size_t length, unsigned f
                              UCT_MEM_ALLOC_PARAM_FIELD_MEM_TYPE |
                              UCT_MEM_ALLOC_PARAM_FIELD_MDS      |
                              UCT_MEM_ALLOC_PARAM_FIELD_NAME;
-    params.flags           = UCT_MD_MEM_ACCESS_ALL;
+    params.flags           = UCT_MD_MEM_ACCESS_ALL & flags;
```
yosefe, line 389: IMO it should be
params.flags = flags;
Maybe fix callers accordingly

src/uct/base/uct_mem.c
```diff
@@ -452,7 +452,7 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_iface_mp_chunk_alloc,
 
     length = sizeof(*hdr) + *size_p;
     status = uct_iface_mem_alloc(&iface->super, length,
-                                 UCT_MD_MEM_ACCESS_ALL | UCT_MD_MEM_FLAG_LOCK,
+                                 iface->mem_access_flags | UCT_MD_MEM_FLAG_LOCK,
```
yosefe, line 455: this is local memory pools.
I think it could be just LOCAL_READ|LOCAL_WRITE

src/uct/ib/base/ib_md.c
```diff
@@ -38,8 +38,11 @@
         ucs_status_t _status; \
         \
         UCT_MD_MEM_DEREG_CHECK_PARAMS( \
-                _params, (_ib_md)->cap_flags & UCT_MD_FLAG_INVALIDATE); \
-        _status = uct_ib_md_mem_dereg_params_invalidate_check(_params); \
+                _params, ((_ib_md)->cap_flags & (UCT_MD_FLAG_INVALIDATE | \
```
yosefe, line 41: 1. why need UCT_MD_FLAG_INVALIDATE flag?
2. maybe move this check inside uct_ib_md_mem_dereg_params_invalidate_check() function, to have less code in the macro?

src/uct/ib/base/ib_md.c
```diff
@@ -759,6 +762,13 @@ uct_ib_mem_reg_internal(uct_md_h uct_md, void *address, size_t length,
         memh->flags |= UCT_IB_MEM_ACCESS_REMOTE_ATOMIC;
     }
 
+#if ENABLE_PARAMS_CHECK
+    if (params->flags & (UCT_MD_MEM_ACCESS_REMOTE_GET |
+                         UCT_MD_MEM_ACCESS_REMOTE_PUT)) {
+        memh->flags |= UCT_IB_MEM_ACCESS_REMOTE_RMA;
```
yosefe, line 768: we could define UCT_IB_MEM_ACCESS_REMOTE_RMA to 0 if ENABLE_PARAMS_CHECK==0

src/uct/ib/base/ib_md.c
```diff
@@ -1003,7 +1013,12 @@ uct_ib_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
             ucs_trace("created atomic key 0x%x for 0x%x", memh->atomic_rkey,
                       memh->lkey);
         } else if (status == UCS_ERR_UNSUPPORTED) {
-            /* ignore for atomic MR */
+            if (flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE) {
+                ucs_assert(md->cap_flags & UCT_MD_FLAG_INVALIDATE_AMO);
```
yosefe, line 1017: what is this assert checking?
could it be done in an earlier stage? why only assert when we fail?

src/uct/ib/base/ib_md.c
```diff
@@ -1003,7 +1013,12 @@ uct_ib_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
...
+                return UCS_ERR_IO_ERROR;
+            } else {
+                /* ignore for atomic MR */
+            }
```
yosefe, line 1021: remove?

src/uct/ib/base/ib_md.h
```diff
@@ -558,26 +562,56 @@ static UCS_F_ALWAYS_INLINE uint32_t uct_ib_memh_get_lkey(uct_mem_h memh)
...
+    return UCS_OK;
+}
+
+
 static UCS_F_ALWAYS_INLINE ucs_status_t
 uct_ib_md_mem_dereg_params_invalidate_check(
-        const uct_md_mem_dereg_params_t *params)
+        const uct_md_mem_dereg_params_t *params, uint64_t cap_flags)
 {
+#if ENABLE_PARAMS_CHECK
```
yosefe, line 585: if (!ENABLE_PARAMS_CHECK) return UCS_OK;

src/uct/ib/base/ib_md.h
```diff
@@ -558,26 +562,56 @@ static UCS_F_ALWAYS_INLINE uint32_t uct_ib_memh_get_lkey(uct_mem_h memh)
...
-                                                              FIELD_MEMH, NULL);
-        flags   = UCT_MD_MEM_DEREG_FIELD_VALUE(params, flags, FIELD_FLAGS, 0);
-        if ((flags & UCT_MD_MEM_DEREG_FLAG_INVALIDATE) &&
-            (ib_memh->indirect_rkey == UCT_IB_INVALID_MKEY)) {
-            return UCS_ERR_INVALID_PARAM;
-        }
+    ib_memh = (uct_ib_mem_t*)UCT_MD_MEM_DEREG_FIELD_VALUE(params, memh,
+                                                          FIELD_MEMH, NULL);
+    flags   = UCT_MD_MEM_DEREG_FIELD_VALUE(params, flags, FIELD_FLAGS, 0);
+    if (!(flags & UCT_MD_MEM_DEREG_FLAG_INVALIDATE)) {
```
yosefe, line 593: seems wrong

src/uct/ib/base/ib_md.h
```diff
@@ -558,26 +562,56 @@ static UCS_F_ALWAYS_INLINE uint32_t uct_ib_memh_get_lkey(uct_mem_h memh)
...
+            ib_memh->flags & UCT_IB_MEM_ACCESS_REMOTE_RMA,
+            cap_flags & UCT_MD_FLAG_INVALIDATE_RMA);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    return uct_ib_md_rkey_mem_dereg_invalidate_check(
+            ib_memh->atomic_rkey,
+            ib_memh->flags & UCT_IB_MEM_ACCESS_REMOTE_ATOMIC,
+            cap_flags & UCT_MD_FLAG_INVALIDATE_AMO);
```
yosefe, line 608: these checks are redundant IMO
we want that even when only RMA invalidate is supported, we are still able to register uct_memh with both AMO|RMA and invalidate. in this case we expect remote RMA will really be invalidated, but remote AMO, if have such, could have undefined behavior.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: IMO this function should be removed

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -447,7 +447,7 @@ static ucs_status_t uct_ib_mlx5_devx_reg_indirect_key(uct_ib_md_t *ibmd,
     uct_ib_mlx5_mem_t *memh = ucs_derived_of(ib_memh, uct_ib_mlx5_mem_t);
     ucs_status_t status;
 
-    ucs_assert(uct_ib_mlx5_devx_md_is_ksm_atomic_supported(md));
+    ucs_assert(uct_ib_mlx5_devx_md_flags_test(md, UCT_IB_MLX5_MD_FLAG_KSM));
```
yosefe, line 450: ucs_assertv(md->flags & UCT_IB_MLX5_MD_FLAG_KSM, "md=%d", md_dev_name...)

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -532,7 +532,9 @@ static ucs_status_t uct_ib_mlx5_devx_reg_atomic_key(uct_ib_md_t *ibmd,
     ucs_status_t status;
     uint8_t mr_id;
 
-    if (!uct_ib_mlx5_devx_md_is_ksm_atomic_supported(md)) {
+    if (!uct_ib_mlx5_devx_md_flags_test(md,
```
yosefe, line 535: ucs_test_all_flags(...)

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -569,7 +571,9 @@ static ucs_status_t uct_ib_mlx5_devx_dereg_atomic_key(uct_ib_md_t *ibmd,
     uct_ib_mlx5_mem_t *memh = ucs_derived_of(ib_memh, uct_ib_mlx5_mem_t);
     uct_ib_mlx5_md_t *md    = ucs_derived_of(ibmd, uct_ib_mlx5_md_t);
 
-    if (!uct_ib_mlx5_devx_md_is_ksm_atomic_supported(md)) {
+    if (!uct_ib_mlx5_devx_md_flags_test(md,
```
yosefe, line 574: ucs_test_all_flags

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -587,13 +591,20 @@ static ucs_status_t uct_ib_mlx5_devx_reg_multithreaded(uct_ib_md_t *ibmd,
...
     ucs_status_t status;
     uint32_t mkey;
     int mr_num;
 
-    if (!uct_ib_mlx5_devx_md_is_ksm_atomic_supported(md)) {
+    if (atomic) {
+        test_flags |= UCT_IB_MLX5_MD_FLAG_INDIRECT_ATOMICS;
+    }
+
+    if (!uct_ib_mlx5_devx_md_flags_test(md, test_flags)) {
```
yosefe, line 607:
```c
if (!(md->flags & UCT_IB_MLX5_MD_FLAG_KSM) ||
    (is_atomic && !(md->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_ATOMICS))) {
    return UCS_ERR_UNSUPPORTED;
}
```

src/uct/sm/mm/base/mm_iface.c
```diff
@@ -775,8 +775,8 @@ static UCS_CLASS_INIT_FUNC(uct_mm_iface_t, uct_md_h md, uct_worker_h worker,
     /* Allocate the receive FIFO */
     status = uct_iface_mem_alloc(&self->super.super.super,
                                  UCT_MM_GET_FIFO_SIZE(self),
-                                 UCT_MD_MEM_ACCESS_ALL, "mm_recv_fifo",
-                                 &self->recv_fifo_mem);
+                                 self->super.super.mem_access_flags,
```
yosefe, line 778: IMO the field mem_access_flags should be removed at all, we're not going to invalidate access to the shared memory FIFO

test/gtest/uct/test_md.cc
```diff
@@ -96,6 +99,22 @@ test_md::test_md()
     /* coverity[uninit_member] */
 }
 
+bool test_md::is_bf_arm() const
+{
+#if HAVE_IB
+    uct_md_h md_h = md();
+    if ((ucs_arch_get_cpu_model() == UCS_CPU_MODEL_ARM_AARCH64) &&
```
yosefe, line 106: can we put less code inside the HAVE_IB?
only when we define and use ib_md

test/gtest/uct/test_md.cc
```diff
@@ -751,68 +770,87 @@ UCS_TEST_SKIP_COND_P(test_md, dereg_bad_arg,
...
+
+        if (is_unsupported_caps(md_flags)) {
+            params.field_mask = UCT_MD_MEM_DEREG_FIELD_COMPLETION |
+                                UCT_MD_MEM_DEREG_FIELD_FLAGS |
+                                UCT_MD_MEM_DEREG_FIELD_MEMH;
+            status            = uct_md_mem_dereg_v2(md(), &params);
+            ASSERT_UCS_STATUS_EQ(UCS_ERR_UNSUPPORTED, status);
+
+            params.field_mask = UCT_MD_MEM_DEREG_FIELD_MEMH;
+            status            = uct_md_mem_dereg_v2(md(), &params);
```
yosefe, line 819: isn't this code duplication of some cases in "else"?

test/gtest/uct/test_md.cc
```diff
@@ -751,68 +770,87 @@ UCS_TEST_SKIP_COND_P(test_md, dereg_bad_arg,
...
+        md_flags_remote_rma,
+        UCT_MD_MEM_ACCESS_REMOTE_ATOMIC
+    };
+    auto is_unsupported_caps                        =
+            [this] (unsigned md_flags) {
+        return (md_flags & md_flags_remote_rma) ?
+               !check_caps(UCT_MD_FLAG_INVALIDATE_RMA) :
+               (md_flags & UCT_MD_MEM_ACCESS_REMOTE_ATOMIC) ?
+               !check_caps(UCT_MD_FLAG_INVALIDATE_AMO) : true;
+    };
```
yosefe, line 786: seems over complicated
maybe put the checks in helper function and call this helper function with different flags parameters?

src/uct/api/uct.h
```diff
@@ -759,7 +761,12 @@ enum {
     /**
      * MD supports registering a dmabuf file descriptor.
      */
-    UCT_MD_FLAG_REG_DMABUF    = UCS_BIT(10)
+    UCT_MD_FLAG_REG_DMABUF    = UCS_BIT(10),
+
+    /**
+     * All further flags must be added as part of API v2.
```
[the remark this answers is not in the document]
shamisp: I think the comment is useful and will prevent other dev from mistake. I would make the statement much more strict. The enum must not be extended. Any additional flags must to to API v2 @ref

src/uct/api/v2/uct_v2.h
```diff
@@ -775,6 +775,31 @@ typedef struct {
 } uct_md_attr_v2_t;
 
 
+/**
+ * @ingroup UCT_MD
+ * @brief  Memory domain capability flags.
+ */
+enum {
```
shamisp, line 782: Please define the type of enum

src/uct/api/v2/uct_v2.h
shamisp: MD -> memory domain

src/uct/api/v2/uct_v2.h
```diff
@@ -775,6 +775,31 @@ typedef struct {
...
+    UCT_MD_FLAG_V2_FIRST       = UCT_MD_FLAG_LAST,
+
+    /**
+     * MD supports invalidation of memory handle registered with
+     * @ref UCT_MD_MEM_ACCESS_RMA rights by @ref uct_md_mem_reg_v2 and
```
shamisp, line 787: rights - > flag

src/uct/api/v2/uct_v2.h
```diff
@@ -775,6 +775,31 @@ typedef struct {
...
+     * packed key by @ref uct_md_mkey_pack_v2 with
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag.
+     */
+    UCT_MD_FLAG_INVALIDATE_RMA = UCT_MD_FLAG_V2_FIRST,
+
+    /**
+     * MD supports invalidation of memory handle registered with
```
shamisp, line 794: MD - Memory domain

shamisp (changes requested):

test/gtest/uct/test_md.cc
evgeny-leksikov: I don't see another place where only `params.field_mask = UCT_MD_MEM_DEREG_FIELD_MEMH;` is initialized

src/uct/ib/base/ib_md.h
yosefe: nvm

src/uct/ib/base/ib_md.h
yosefe: IMO it's API restriction that we don't need to add since it will make UCP code more complex

test/gtest/uct/test_md.cc
yosefe: ok

src/uct/ib/base/ib_md.h
```diff
@@ -558,26 +564,60 @@ static UCS_F_ALWAYS_INLINE uint32_t uct_ib_memh_get_lkey(uct_mem_h memh)
 }
 
 
+static UCS_F_ALWAYS_INLINE UCS_F_MAYBE_UNUSED ucs_status_t
+uct_ib_md_rkey_mem_dereg_invalidate_check(uint32_t rkey, uint32_t access_mask,
+                                          uint64_t cap_mask)
+{
+    if (access_mask) {
```
yosefe, line 571: if (!access_mask) { return UCS_OK; }

test/gtest/uct/test_md.cc
```diff
@@ -96,6 +99,29 @@ test_md::test_md()
...
+           check_caps(UCT_MD_FLAG_INVALIDATE_AMO) : false;
+};
+
+bool test_md::is_bf_arm() const
+{
+    if ((ucs_arch_get_cpu_model() == UCS_CPU_MODEL_ARM_AARCH64) &&
+        (std::string("ib") == md_attr().component_name)) {
+#if HAVE_IB
+        uct_ib_md_t *ib_md = (uct_ib_md_t*)md();
+        if (ib_md->dev.pci_id.device == 0xa2d6) {
```
yosefe, line 116: add comment it's bluefield2

test/gtest/uct/test_md.cc
```diff
@@ -751,68 +777,82 @@ UCS_TEST_SKIP_COND_P(test_md, dereg_bad_arg,
...
     uct_mem_h memh;
     void *ptr;
     ucs_status_t status;
     uct_md_mem_dereg_params_t params;
 
-    ptr    = malloc(size);
-    status = reg_mem(md_flags, ptr, size, &memh);
-    ASSERT_UCS_OK(status);
+    for (auto mem_flags : mem_flags_vec) {
+        if ((mem_flags.first & UCT_MD_MEM_ACCESS_REMOTE_ATOMIC) && is_bf_arm()) {
```
yosefe, line 796: how about putting the content of for loop in a function. and use it like this:
```
..
test_reg_mem(md_flags_remote_rma,
                       UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA);
test_reg_mem(UCT_MD_MEM_ACCESS_REMOTE_ATOMIC,
                       UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO);
...
```

test/gtest/uct/uct_test.cc
```diff
@@ -927,18 +927,29 @@ void uct_test::entity::mem_alloc_host(size_t length,
     void *address = NULL;
     ucs_status_t status;
     uct_mem_alloc_params_t params;
+    unsigned mem_access_flags;
+
+    if (!(iface_params().field_mask & UCT_IFACE_PARAM_FIELD_FEATURES)) {
+        mem_access_flags = UCT_MD_MEM_ACCESS_ALL;
```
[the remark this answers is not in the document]
yosefe: seems unresolved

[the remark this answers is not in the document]
evgeny-leksikov: ah.. this case is when features are not set at all, AFAIU it should be handled as any feature can be used, in other worlds "all" features are requested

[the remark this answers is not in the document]
yosefe: actually, it would be wrong usage to call ucp_init without features.
so i think this condition will never be true

yosefe:
@shamisp @tonycurtis can you pls review the UCT API changes?

shamisp:
Adding @gitrepoidoscar for review

src/uct/api/v2/uct_v2.h
```diff
@@ -274,23 +274,43 @@ typedef enum {
...
+     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO flag must be set.
      */
-    UCT_MD_MKEY_PACK_FLAG_INVALIDATE = UCS_BIT(0),
+    UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA = UCS_BIT(0),
+
+    /**
+     * The flag is used indicate that remote atomic operation to a memory region
+     * associated with the remote key must fail once the memory region is
+     * deregistered using @ref uct_md_mem_dereg_v2 with
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set. Using
```
shamisp, line 295: This does not read right. I suggest the following (@gitrepoidoscar  please review):

The flag is used indicate that access to a memory region using remote atomic operations must fail once the memory region is deregistered using @ref uct_md_mem_dereg_v2 with @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set,

test/gtest/uct/uct_test.cc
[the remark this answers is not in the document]
evgeny-leksikov: As far as I can see no-one UCT gtest doesn't set iface features, and UCT does not have any parameters check for this except `UCT_IFACE_FEATURE_PUT` in DC transport.
maybe we need to fix it in separate PR?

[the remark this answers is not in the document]
yosefe: - check features for RMA as well
- switch if/else

test/gtest/uct/uct_test.cc
```diff
@@ -924,21 +924,36 @@ uct_test::entity::entity(const resource& resource, uct_md_config_t *md_config,
...
-    void *address = NULL;
+    void *address             = NULL;
+    unsigned mem_access_flags = 0;
     ucs_status_t status;
     uct_mem_alloc_params_t params;
 
+    if (iface_params().field_mask & UCT_IFACE_PARAM_FIELD_FEATURES) {
+        if (iface_params().features & (UCT_IFACE_FEATURE_PUT |
+                                       UCT_IFACE_FEATURE_GET)) {
+            mem_access_flags = UCT_MD_MEM_ACCESS_RMA;
```
yosefe, line 935: |= for consistency

src/uct/api/v2/uct_v2.h
evgeny-leksikov: fixed

test/gtest/uct/uct_test.cc
evgeny-leksikov: done

yosefe (dismissed):

src/uct/api/v2/uct_v2.h
evgeny-leksikov: @shamisp could you please have another look?

yosefe:
@evgeny-leksikov pls squash

src/uct/api/v2/uct_v2.h
```diff
@@ -274,23 +274,43 @@ typedef enum {
...
+     * correctly, and may result in data corruption. In other words in order for
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag to function
+     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA and/or
+     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO flag must be set.
      */
-    UCT_MD_MKEY_PACK_FLAG_INVALIDATE = UCS_BIT(0),
+    UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA = UCS_BIT(0),
+
+    /**
+     * The flag is used indicate that remote atomic operation to a memory region
```
shamisp: to indicate

shamisp: that the automic

src/uct/api/v2/uct_v2.h
```diff
@@ -274,23 +274,43 @@ typedef enum {
...
+     * associated with the remote key must fail once the memory region is
+     * deregistered using @ref uct_md_mem_dereg_v2 with
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set. Using
+     * @ref uct_md_mem_dereg_v2 deregistration routine with
+     * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set on an rkey that was
+     * generated without
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA and/or
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO flags will not function
+     * correctly, and may result in data corruption. In other words in order for
```
shamisp, line 300: In other words,

src/uct/api/v2/uct_v2.h
```diff
@@ -274,23 +274,43 @@ typedef enum {
...
      * @ref uct_md_mem_dereg_v2 deregistration routine with
      * @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag set on an rkey that was
-     * generated without @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag will
-     * not function correctly, and may result in data corruption. In other words
-     * in order for @ref UCT_MD_MEM_DEREG_FLAG_INVALIDATE flag to function
-     * the @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE flag must be set.
+     * generated without
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA and/or
+     * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO flags will not function
+     * correctly, and may result in data corruption. In other words in order for
```
shamisp, line 284: In other words, in order for

shamisp (changes requested):

shamisp (approved):

evgeny-leksikov:
@yosefe commits are squashed after CRs

yosefe (approved):

PR #8771, opened 2022-12-16
GTEST: Don't run gtest on wireless interfaces

shamisp:
## What
Don't run gtest on wireless interfaces

## Why ?
Makes it difficult to develop on desktops and workstations.


shamisp:
@yosefe please take a look

yosefe (approved):

PR #8773, opened 2022-12-17
BUILD: Fix obsolete macro warnings in newer autoconf/libtool

nysal:
## What
Fix obsolete macro warnings when using newer versions of autoconf/libtool

## How ?
* AC_HELP_STRING -> AS_HELP_STRING
* AC_TRY_COMPILE -> AC_COMPILE_IFELSE
* AC_GNU_SOURCE -> AC_USE_SYSTEM_EXTENSIONS
* AC_PROG_LIBTOOL -> LT_INIT
* AC_HEADER_STDC
  - This macro is obsolescent, as current systems have conforming header files

Relevant autoconf and libtool documentation:
https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/Obsolete-Macros.html https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html


yosefe (approved):

shamisp:
I have similar pending patch but I was a bit concerned to push it because it may break some older systems.
Would not be better safer to insert (if present) `autoupdate` command in `autogen.sh` ?

nysal:
Since AC_PREREQ is 2.63 on master branch, all the macros in the PR should be supported. I did verify that the 2.63 documentation does at least list them.

shamisp:
Okay. Good to go then !

PR #8777, opened 2022-12-20
UCP/WIREUP: Filter slow lanes in selection process

shasson5:
## What
Filter slow lanes in selection process.

## Why ?
Avoid selecting TCP lanes if we have IB lanes.
The filtering is done by removing lanes with a large enough progress overhead compared to the best lane.
By default, each lane with more than 0.25 of the minimum overhead is filtered out.

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -14,6 +14,7 @@
 #include "address.h"
 
 #include <ucs/algorithm/qsort_r.h>
+#include <ucs/datastruct/array.inl>
 #include <ucs/datastruct/queue.h>
 #include <ucs/sys/sock.h>
 #include <ucp/core/ucp_ep.inl>
@@ -97,6 +98,8 @@ typedef struct {
     unsigned                  ucp_ep_init_flags;         /* Endpoint init extra flags */
     ucp_tl_bitmap_t           tl_bitmap;                 /* TL bitmap of selected resources */
 } ucp_wireup_select_context_t;
+
+UCS_ARRAY_DEFINE_INLINE(select_info, unsigned, ucp_wireup_select_info_t);
 
 static const char *ucp_wireup_cmpt_flags[] = {
     [ucs_ilog2(UCT_COMPONENT_FLAG_RKEY_PTR)]     = "obtain remote memory pointer",
@@ -1426,8 +1429,8 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
            (ucs_popcount(md_map) < UCP_MAX_OP_MDS);
 }
 
-static double ucp_wireup_get_lane_bw(const ucp_wireup_select_info_t *sinfo,
-                                     ucp_worker_h worker,
+static double ucp_wireup_get_lane_bw(ucp_worker_h worker,
+                                     const ucp_wireup_select_info_t *sinfo,
                                      const ucp_address_entry_t *address)
 {
     ucp_context_h context = worker->context;
@@ -1442,53 +1445,44 @@ static double ucp_wireup_get_lane_bw(const ucp_wireup_select_info_t *sinfo,
 }
 
 static unsigned
-ucp_wireup_add_fast_lanes(const ucp_wireup_select_params_t *select_params,
-                          const ucp_wireup_select_info_t *sinfo_array,
-                          unsigned num_sinfo, ucp_worker_h worker,
+ucp_wireup_add_fast_lanes(ucp_worker_h worker,
+                          const ucp_wireup_select_params_t *select_params,
+                          ucs_array_t(select_info) *sinfo_array,
                           ucp_lane_type_t lane_type,
                           ucp_wireup_select_context_t *select_ctx)
 {
-    unsigned num_lanes     = 0;
-    double max_bw          = 0;
-    ucp_context_h context  = worker->context;
-    const double max_ratio = 1. / context->config.ext.multi_lane_max_ratio;
+    ucp_lane_index_t num_lanes = 0;
+    double max_bw              = 0;
+    ucp_context_h context      = worker->context;
+    const double max_ratio     = 1. / context->config.ext.multi_lane_max_ratio;
     const ucp_address_entry_t *address_list;
     ucs_status_t status;
-    int show_error;
-    unsigned sinfo_index;
     double lane_bw;
     const ucp_wireup_select_info_t *sinfo;
-
-    if (num_sinfo == 0) {
-        return 0;
-    }
 
     address_list = select_params->address->address_list;
 
     /* Iterate over all elements and calculate max BW */
-    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
-        lane_bw = ucp_wireup_get_lane_bw(&sinfo_array[sinfo_index], worker,
-                                         address_list);
+    ucs_array_for_each(sinfo, sinfo_array) {
+        lane_bw = ucp_wireup_get_lane_bw(worker, sinfo, address_list);
         max_bw  = ucs_max(lane_bw, max_bw);
     }
 
     /* Compare each element to max BW and filter only fast lanes */
-    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
-        sinfo   = &sinfo_array[sinfo_index];
-        lane_bw = ucp_wireup_get_lane_bw(sinfo, worker, address_list);
+    ucs_array_for_each(sinfo, sinfo_array) {
+        lane_bw = ucp_wireup_get_lane_bw(worker, sinfo, address_list);
 
         if (lane_bw < (max_bw * max_ratio)) {
             ucs_trace(UCT_TL_RESOURCE_DESC_FMT
-                      " : BW lower than threshold (%f < %f), dropping lane",
+                      " : bandwidth %.2f lower than %.2f x %.2f, dropping lane",
                       UCT_TL_RESOURCE_DESC_ARG(
                               &context->tl_rscs[sinfo->rsc_index].tl_rsc),
-                      lane_bw / max_bw, max_ratio);
+                      lane_bw, max_ratio, max_bw);
             continue;
         }
 
-        show_error = (num_lanes == 0);
-        status     = ucp_wireup_add_lane(select_params, sinfo, lane_type,
-                                         show_error, select_ctx);
+        status = ucp_wireup_add_lane(select_params, sinfo, lane_type,
+                                     num_lanes == 0, select_ctx);
         if (status != UCS_OK) {
             break;
         }
@@ -1505,22 +1499,21 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
                         ucp_tl_bitmap_t tl_bitmap, ucp_lane_index_t excl_lane,
                         ucp_wireup_select_context_t *select_ctx)
 {
-    ucp_ep_h ep                                   = select_params->ep;
-    ucp_context_h context                         = ep->worker->context;
-    ucp_wireup_select_info_t sinfo[UCP_MAX_LANES] = {{0}};
-    ucp_wireup_dev_usage_count dev_count          = {};
+    ucp_ep_h ep                          = select_params->ep;
+    ucp_context_h context                = ep->worker->context;
+    ucp_wireup_dev_usage_count dev_count = {};
     const uct_iface_attr_t *iface_attr;
     const ucp_address_entry_t *ae;
     ucs_status_t status;
-    unsigned num_sinfo;
     uint64_t local_dev_bitmap;
     uint64_t remote_dev_bitmap;
     ucp_rsc_index_t dev_index;
     ucp_md_map_t md_map;
     ucp_rsc_index_t rsc_index;
     unsigned addr_index;
+    UCS_ARRAY_DEFINE_ONSTACK(sinfo_array, select_info, UCP_MAX_LANES);
+    ucp_wireup_select_info_t *sinfo;
 
-    num_sinfo             = 0;
     md_map                = bw_info->md_map;
     local_dev_bitmap      = bw_info->local_dev_bitmap;
     remote_dev_bitmap     = bw_info->remote_dev_bitmap;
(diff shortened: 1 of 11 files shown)


src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,94 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+    ucp_worker_iface_t *wiface = ucp_worker_iface(worker, rsc_index);
+    uct_perf_attr_t perf_attr;
+
+    perf_attr.field_mask = UCT_PERF_ATTR_FIELD_OPERATION |
+                           UCT_PERF_ATTR_FIELD_PROGRESS_OVERHEAD;
+
+    /* We use AM_SHORT as a dummy operation because we don't know which
+        protocol will be selected (anyway it has no effect on the value of
+        progress_overhead). */
+    perf_attr.operation = UCT_EP_OP_AM_SHORT;
```
gleon99, line 1442: Is it needed at all?

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,94 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+        overhead     = ucp_wireup_get_lane_progress_overhead(
+                worker, sinfo_array[sinfo_index].rsc_index);
+        min_overhead = ucs_min(overhead, min_overhead);
+    }
+
+    /* Compare each element to minimum progress overhead and filter only fast lanes */
+    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
+        overhead = ucp_wireup_get_lane_progress_overhead(
+                worker, sinfo_array[sinfo_index].rsc_index);
+        if ((min_overhead > 0) && ((min_overhead / overhead) < max_ratio)) {
```
gleon99, line 1479: Something seems wrong with the logic hear.
`min_overhead` will be `0` if there's at least 1 "fast" iface, so this `if` would never happen?
Also, I would care more about checking the denominator (`overhead`) here to prevent division by 0.

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,94 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+            continue;
+        }
+
+        show_error = (num_lanes == 0);
```
gleon99, line 1483: `Align `=`.

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,94 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+{
+    ucp_worker_iface_t *wiface = ucp_worker_iface(worker, rsc_index);
+    uct_perf_attr_t perf_attr;
+
+    perf_attr.field_mask = UCT_PERF_ATTR_FIELD_OPERATION |
+                           UCT_PERF_ATTR_FIELD_PROGRESS_OVERHEAD;
+
+    /* We use AM_SHORT as a dummy operation because we don't know which
+        protocol will be selected (anyway it has no effect on the value of
+        progress_overhead). */
```
brminich, line 1441: alignment

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,94 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+    perf_attr.operation = UCT_EP_OP_AM_SHORT;
+
+    if (uct_iface_estimate_perf(wiface->iface, &perf_attr) != UCS_OK) {
```
brminich, line 1444: why can't we rely on just BW? We already filtering lanes  according to the BW, but on a later stage, when RMA BW lanes already setup. Maybe we can move this filtering earlier (here), so slow lanes will not be initialized at all?

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,94 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+ucp_wireup_add_fast_lanes(const ucp_wireup_select_params_t *select_params,
+                          const ucp_wireup_select_info_t *sinfo_array,
+                          unsigned num_sinfo, ucp_worker_h worker,
+                          ucp_lane_type_t lane_type,
+                          ucp_wireup_select_context_t *select_ctx)
+{
+    unsigned num_lanes     = 0;
+    double min_overhead    = DBL_MAX;
+    ucp_context_h context  = worker->context;
+    const double max_ratio = 1. / context->config.ext.multi_lane_max_ratio;
```
brminich, line 1462: now the same parameter is used as ratio of BW and progress overhead, which is confusing

src/uct/api/v2/uct_v2.h
```diff
@@ -167,6 +170,12 @@ typedef struct {
      */
     double              recv_overhead;
 
+    /**
+     * This is the time spent progressing the request, in seconds.
```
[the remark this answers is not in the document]
shasson5: why not?

src/ucp/wireup/select.c
shasson5: BW filtering cause some issue, because it also filters IB transport from intra node endpoints (where you have SHM transports).

src/ucp/wireup/select.c
shasson5: maybe we should rename it to some more general purpose parameter?

src/ucp/wireup/select.c
shasson5: correct, it was some leftovers from prev code.

shasson5: min_overhead is always bigger than overhead, so min_overhead > 0 also implies overhead > 0.
if min_overhead is 0 than we don't want to filter anything, so the condition will never be true.

gleon99: "min_overhead is always bigger than overhead" - are you sure?
Line 1472:
`min_overhead = ucs_min(overhead, min_overhead);`

src/uct/api/v2/uct_v2.h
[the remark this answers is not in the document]
gleon99: The progress is for `iface`.

src/ucp/wireup/select.c
shasson5: that's an option but wouldn't it be a tailor made solution only for this specific issue?

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -209,6 +209,12 @@ class test_ucp_worker_discard : public ucp_test {
         }
 
         flush_req = sender().flush_worker_nb(0);
+        if ((flush_req == NULL) &&
```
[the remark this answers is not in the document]
shasson5: AFAIR, there was an implicit assumption in this test that there will always be some pending requests on the worker.
But when I filtered TCP it stopped being true

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,90 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+    /* Iterate over all elements and calculate minimum progress overhead */
+    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
+        overhead     = ucp_wireup_get_lane_progress_overhead(
+                worker, sinfo[sinfo_index].rsc_index);
+        min_overhead = ucs_min(overhead, min_overhead);
+    }
+
+    /* Compare each element to minimum progress overhead and filter only fast lanes */
+    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
+        overhead = ucp_wireup_get_lane_progress_overhead(
```
gleon99, line 1473: Assert `overhead` before dividing.

src/ucp/wireup/select.c
```diff
@@ -1457,30 +1519,23 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
-                                                   bw_info->criteria.lane_type,
-                                                   show_error, select_ctx);
+                                                 remote_dev_bitmap, 0,
+                                                 &sinfo[num_sinfo]);
             if (status != UCS_OK) {
                 break;
             }
 
-            num_lanes++;
+            rsc_index  = sinfo[num_sinfo].rsc_index;
```
gleon99, line 1534: Align `=`

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,90 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
            (ucs_popcount(md_map) < UCP_MAX_OP_MDS);
 }
 
+static double ucp_wireup_get_lane_progress_overhead(ucp_worker_h worker,
+                                                    ucp_rsc_index_t rsc_index)
+{
+    static const double min_valid_overhead = 10e-9;
```
gleon99, line 1433: Align `=`

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,90 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+    ucp_worker_iface_t *wiface = ucp_worker_iface(worker, rsc_index);
+    uct_perf_attr_t perf_attr;
+
+    perf_attr.field_mask = UCT_PERF_ATTR_FIELD_PROGRESS_OVERHEAD;
+    if (uct_iface_estimate_perf(wiface->iface, &perf_attr) != UCS_OK) {
+        /* Skip the overhead check if no value was retrieved */
+        return min_valid_overhead;
```
gleon99, line 1440: WDYT of simplifying the flow (pseudo-code):
```
overhead = 10e-9
if uct_iface_estimate_perf... = OK:
    overhead = max(overhead, perf_attr...)
return overhead
```

test/gtest/ucp/test_ucp_worker.cc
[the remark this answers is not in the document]
gleon99: As discussed, please add a comment here.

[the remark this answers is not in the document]
shasson5: already done in line 212

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1428,96 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+        return 0;
+    }
+
+    /* Iterate over all elements and calculate max BW */
+    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
+        lane_bw = ucp_wireup_get_lane_bw(&sinfo_array[sinfo_index], worker,
+                                         select_params->address->address_list);
+        max_bw  = ucs_max(lane_bw, max_bw);
+
+        printf("lane_bw %f\n", lane_bw);
```
rakhmets, line 1472: Please remove this line.

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1428,96 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+    }
+
+    ucs_assertv(max_bw > 0, "max_bw has illegal value: %f", max_bw);
+
+    /* Compare each element to max BW and filter only fast lanes */
+    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
+        lane_bw = ucp_wireup_get_lane_bw(&sinfo_array[sinfo_index], worker,
+                                         select_params->address->address_list);
+
+        if ((lane_bw / max_bw) < max_ratio) {
```
brminich, line 1482: maybe add ucs_trace here what lane (and on what device) was dropped? Can help debugging issues

src/uct/api/v2/uct_v2.h
```diff
@@ -167,6 +170,12 @@ typedef struct {
      */
     double              recv_overhead;
 
+    /**
+     * This is the time spent progressing the iface, in seconds.
+     * This field is set by the UCT layer.
+     */
+    double              progress_overhead;
+
```
brminich, line 178: are these changes needed now?

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1426,101 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+    }
+
+    /* Iterate over all elements and calculate max BW */
+    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
+        lane_bw = ucp_wireup_get_lane_bw(&sinfo_array[sinfo_index], worker,
+                                         select_params->address->address_list);
+        max_bw  = ucs_max(lane_bw, max_bw);
+    }
+
+    ucs_assertv(max_bw > 0, "max_bw has illegal value: %f", max_bw);
```
rakhmets, line 1472: Which problem does the assert reveal?
Maybe it is better replace line 1480 by `if (lane_bw < (max_bw * max_ratio)) {`, and get rid of assert, if the only purpose of the assert is to avoid division by zero.

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1427,94 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+        /* Skip the overhead check if no value was retrieved */
+        return 0;
+    }
+
+    return perf_attr.progress_overhead;
+}
+
+static unsigned
+ucp_wireup_add_fast_lanes(const ucp_wireup_select_params_t *select_params,
+                          const ucp_wireup_select_info_t *sinfo_array,
```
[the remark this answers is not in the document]
gleon99: @shasson5 unresolved

src/ucp/wireup/select.c
```diff
@@ -1426,29 +1426,101 @@ ucp_wireup_is_md_map_count_valid(ucp_context_h context, ucp_md_map_t md_map)
...
+    const ucp_wireup_select_info_t *sinfo;
+
+    if (num_sinfo == 0) {
+        return 0;
+    }
+
+    /* Iterate over all elements and calculate max BW */
+    for (sinfo_index = 0; sinfo_index < num_sinfo; ++sinfo_index) {
+        lane_bw = ucp_wireup_get_lane_bw(&sinfo_array[sinfo_index], worker,
+                                         select_params->address->address_list);
```
gleon99, line 1468: Can put `select_params->address->address_list` in a var (used in 2 loops).

src/ucp/wireup/select.c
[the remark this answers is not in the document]
shasson5: notice I added a local var with the same name (sinfo)

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -209,6 +209,14 @@ class test_ucp_worker_discard : public ucp_test {
         }
 
         flush_req = sender().flush_worker_nb(0);
+        /* In some cases, there will be nothing to flush
+         * so we need to skip the progress loop */
```
yosefe, line 213: pls word-wrap the text to fill the whole line up to 80 chars

src/ucp/wireup/select.c
```diff
@@ -1442,53 +1445,44 @@ static double ucp_wireup_get_lane_bw(const ucp_wireup_select_info_t *sinfo,
 }
 
 static unsigned
-ucp_wireup_add_fast_lanes(const ucp_wireup_select_params_t *select_params,
-                          const ucp_wireup_select_info_t *sinfo_array,
-                          unsigned num_sinfo, ucp_worker_h worker,
+ucp_wireup_add_fast_lanes(ucp_worker_h worker,
+                          const ucp_wireup_select_params_t *select_params,
+                          ucs_array_t(select_info) *sinfo_array,
```
yosefe, line 1450: const

src/ucp/wireup/select.c
```diff
@@ -1505,22 +1499,21 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
     const ucp_address_entry_t *ae;
     ucs_status_t status;
-    unsigned num_sinfo;
     uint64_t local_dev_bitmap;
     uint64_t remote_dev_bitmap;
     ucp_rsc_index_t dev_index;
     ucp_md_map_t md_map;
     ucp_rsc_index_t rsc_index;
     unsigned addr_index;
+    UCS_ARRAY_DEFINE_ONSTACK(sinfo_array, select_info, UCP_MAX_LANES);
```
yosefe, line 1514: this is initialized var; it should be 1st

yosefe (approved):

yosefe (approved):

PR #8778, opened 2022-12-20
UCP/CORE: Fixed processing of allow/deny list of transports.

rakhmets:
## What
Updated logic to handle allow/deny list of transports.

## Why ?
Previously, transport was completely disabled if it was mentioned in deny list. However, some transport may be used as an auxiliary transport, and disabling them in this way may implicitly break using of another transport.
E.g, transport `tl_name_1` can be used only with `tl_name_2` as an auxiliary transport. `UCX_TLS=^tl_name_2` implicitly break usage of  `tl_name_1`.
After these changes, such transports can be disabled only explicitly: `UCX_TLS=^tl_name_2,tl_name_2:aux`.
Note, that to completely disable other transports it is enough to mention them without 'aux' suffix.

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -190,6 +190,7 @@ communication and TCP sockets for inter-node communication.
   memory pointer type and copying to/from GPU memory.  
 
 It's possible to restrict the transports in use by setting `UCX_TLS=<tl1>,<tl2>,...`.
+`^` at the beginning turns the list into a deny list.
 The list of all transports supported by UCX on the current machine can be generated
 by `ucx_info -d` command.
 > **IMPORTANT NOTE**
@@ -221,8 +222,12 @@ In addition to the built-in transports it's possible to use aliases which specif
  
 For example:
 - `UCX_TLS=rc` will select RC, UD for bootstrap, and prefer accelerated transports
-- `UCX_TLS=rc,cuda` will select RC along with Cuda memory transports.
-
+- `UCX_TLS=rc,cuda` will select RC along with Cuda memory transports
+- `UCX_TLS=^rc` will select all available transports, except RC
+> **IMPORTANT NOTE**
+> `UCX_TLS=^ud` will select all available transports, except UD. However, UD
+will still be available for bootstrap. Only `UCX_TLS=^ud,ud:aux` will disable UD
+completely.
 
 <br/>
 
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -61,6 +61,16 @@
 
 #define UCP_CPU_EST_BCOPY_BW_DEFAULT (5800 * UCS_MBYTE)
 
+#define UCP_TL_AUX_SUFFIX    "aux"
+#define UCP_TL_AUX(_tl_name) _tl_name ":" UCP_TL_AUX_SUFFIX
+
+
+typedef enum ucp_transports_list_search_result {
+    UCP_TRANSPORTS_LIST_SEARCH_RESULT_PRIMARY      = UCS_BIT(0),
+    UCP_TRANSPORTS_LIST_SEARCH_RESULT_AUX_IN_MAIN  = UCS_BIT(1),
+    UCP_TRANSPORTS_LIST_SEARCH_RESULT_AUX_IN_ALIAS = UCS_BIT(2)
+} ucp_transports_list_search_result_t;
+
 
 /* Declare all am handlers */
 UCP_AM_HANDLER_FOREACH(UCP_AM_HANDLER_DECL)
@@ -546,12 +556,13 @@ static ucp_tl_alias_t ucp_tl_aliases[] = {
   { "ud_v",  { "ud_verbs", NULL } },
   { "ud_x",  { "ud_mlx5", NULL } },
   { "ud",    { "ud_mlx5", "ud_verbs", NULL } },
-  { "rc_v",  { "rc_verbs", "ud_verbs:aux", NULL } },
-  { "rc_x",  { "rc_mlx5", "ud_mlx5:aux", NULL } },
-  { "rc",    { "rc_mlx5", "ud_mlx5:aux", "rc_verbs", "ud_verbs:aux", NULL } },
-  { "dc",    { "dc_mlx5", "ud_mlx5:aux", NULL } },
-  { "dc_x",  { "dc_mlx5", "ud_mlx5:aux", NULL } },
-  { "ugni",  { "ugni_smsg", "ugni_udt:aux", "ugni_rdma", NULL } },
+  { "rc_v",  { "rc_verbs", UCP_TL_AUX("ud_verbs"), NULL } },
+  { "rc_x",  { "rc_mlx5", UCP_TL_AUX("ud_mlx5"), NULL } },
+  { "rc",    { "rc_mlx5", UCP_TL_AUX("ud_mlx5"), "rc_verbs",
+               UCP_TL_AUX("ud_verbs"), NULL } },
+  { "dc",    { "dc_mlx5", UCP_TL_AUX("ud_mlx5"), NULL } },
+  { "dc_x",  { "dc_mlx5", UCP_TL_AUX("ud_mlx5"), NULL } },
+  { "ugni",  { "ugni_smsg", UCP_TL_AUX("ugni_udt"), "ugni_rdma", NULL } },
   { "cuda",  { "cuda_copy", "cuda_ipc", "gdr_copy", NULL } },
   { "rocm",  { "rocm_copy", "rocm_ipc", "rocm_gdr", NULL } },
   { NULL }
@@ -792,31 +803,20 @@ static unsigned ucp_tl_alias_count(ucp_tl_alias_t *alias)
 }
 
 static int ucp_tls_array_is_present(const char **tls, unsigned count,
-                                    const char *tl_name, const char *info,
-                                    uint8_t *rsc_flags, uint64_t *tl_cfg_mask)
+                                    const char *tl_name, const char *str_suffix,
+                                    uint64_t *tl_cfg_mask)
 {
     uint64_t mask;
 
-    if ((mask = ucp_str_array_search(tls, count, tl_name, NULL)) != 0) {
-        *tl_cfg_mask |= mask;
-        ucs_trace("enabling tl '%s'%s", tl_name, info);
-        return 1;
-    } else if ((mask = ucp_str_array_search(tls, count, tl_name, "aux")) != 0) {
-        /* Search for tl names with 'aux' suffix, such tls can be
-         * used for auxiliary wireup purposes only */
-        *rsc_flags   |= UCP_TL_RSC_FLAG_AUX;
-        *tl_cfg_mask |= mask;
-        ucs_trace("enabling auxiliary tl '%s'%s", tl_name, info);
-        return 1;
-    } else {
-        return 0;
-    }
+    mask          = ucp_str_array_search(tls, count, tl_name, str_suffix);
+    *tl_cfg_mask |= mask;
+    return (mask != 0);
 }
 
 static int
-ucp_config_is_tl_name_present(const ucs_config_allow_list_t *allow_list,
+ucp_config_is_tl_name_present(const ucs_config_names_array_t *tl_array,
                               const char *tl_name, int is_alias,
-                              uint8_t *rsc_flags, uint64_t *tl_cfg_mask)
+                              const char *str_suffix, uint64_t *tl_cfg_mask)
 {
     char strict_name[UCT_TL_NAME_MAX + 1];
 
@@ -824,13 +824,13 @@ ucp_config_is_tl_name_present(const ucs_config_allow_list_t *allow_list,
 
     return /* strict name, with leading \\ */
             (!is_alias &&
-             (ucp_tls_array_is_present((const char**)allow_list->array.names,
-                                       allow_list->array.count, strict_name, "",
-                                       rsc_flags, tl_cfg_mask))) ||
+             (ucp_tls_array_is_present((const char**)tl_array->names,
+                                       tl_array->count, strict_name, str_suffix,
+                                       tl_cfg_mask))) ||
             /* plain transport name */
-            (ucp_tls_array_is_present((const char**)allow_list->array.names,
-                                      allow_list->array.count, tl_name, "",
-                                      rsc_flags, tl_cfg_mask));
+            (ucp_tls_array_is_present((const char**)tl_array->names,
+                                      tl_array->count, tl_name, str_suffix,
+                                      tl_cfg_mask));
 }
 
 static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource,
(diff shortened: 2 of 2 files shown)


shamisp:
Imho we have to add a section to readthedocs describing this.

src/ucp/core/ucp_context.c
```diff
@@ -559,6 +559,10 @@ static ucp_tl_alias_t ucp_tl_aliases[] = {
 };
 
 
+/* List of transports that can be used as an auxiliary transport. */
+const char *ucp_context_tl_aux[] = {"ud_verbs", "ud_mlx5", "ugni_udt"};
```
brminich, line 563: why do we need this? In theory we can declare any transport to be `aux`

src/ucp/core/ucp_context.c
```diff
@@ -865,58 +903,104 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
+                search_result |= UCP_TLS_ARRAY_SEARCH_RESULT_FOUND_AUX;
             }
         }
     }
 
-    return allow_list->mode == UCS_CONFIG_ALLOW_LIST_NEGATE;
+    if (ucp_enable_tl_aux(is_allow_mode, search_result) ||
+        ucp_tls_array_found(search_result)) {
+        /* Transport may be added as auxiliary in two cases:
+         * 1. Explicitely requested in allow list as auxiliary:
```
brminich, line 992: `explicitly`

src/ucp/core/ucp_context.c
```diff
@@ -865,58 +903,104 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
+         *     1.a. UCX_TLS=...tl_name:aux...
+         *     1.b. alias = { tl_name }, UCX_TLS=...alias:aux,...
+         * 2. Implicetely mentioned as auxiliary in alias:
```
brminich, line 995: `implicitly`?

src/ucp/core/ucp_context.c
rakhmets: We need this list to not disable the transport from this list as auxiliary transport if the transport was mentioned in the deny list. E.g.:
`UCX_TLS=^ud` should only disable `ud_verbs`, and `ud_mlx5` as a _real_ transport. The output of list of transport should be:
```

Comment (brminich) [reply src/ucp/core/ucp_context.c:563]: why is it specific to the transport name, rather than `aux` specificator?

Comment (rakhmets) [reply src/ucp/core/ucp_context.c:563]: We can enable any transport as auxiliary using _allow_ list, e.g. `UCX_TLS=any_transport:aux,any_alias:aux`.
However, we need the list to completely disable transports from this list only if they mentioned in _deny_ list in the following way: `UCX_TLS=^transport_from_the_list:aux,transport_from_the_list`, or `UCX_TLS=^alias_containing_transport_from_the_list:aux,alias_containing_transport_from_the_list`, or `UCX_TLS=^alias_containing_transport_from_the_list:aux,transport_from_the_list`, or `UCX_TLS=^transport_from_the_list:aux,alias_containing_transport_from_the_list`.

Comment (brminich) [reply src/ucp/core/ucp_context.c:563]: why not avoiding to disable any transport with `aux` if it is not specified explicitly in the deny list?
with the newly added array the code is error prone - if new aux transport is added to our alias names array it should also be reflected in this array (which is not really obvious)

**Review by yosefe — DISMISSED** [review]

File: docs/source/faq.md [hunk 4]
```diff
@@ -221,8 +222,12 @@ In addition to the built-in transports it's possible to use aliases which specif
  
 For example:
 - `UCX_TLS=rc` will select RC, UD for bootstrap, and prefer accelerated transports
-- `UCX_TLS=rc,cuda` will select RC along with Cuda memory transports.
-
+- `UCX_TLS=rc,cuda` will select RC along with Cuda memory transports
+- `UCX_TLS=^rc` will select all available transports, but RC
```
yosefe, line 226: but -> except

src/ucp/core/ucp_context.c
```diff
@@ -574,6 +579,38 @@ const ucp_tl_bitmap_t ucp_tl_bitmap_max = {{UINT64_MAX, UINT64_MAX}};
...
+        }
+    }
+
+    return 1;
+}
+
+static int ucp_tls_array_found(uint64_t search_result)
+{
+    return ucp_tls_array_check_search_result(search_result,
+                                             UCP_TLS_ARRAY_SEARCH_RESULT_FOUND);
```
yosefe, line 604: IMO this can be done much simpler , without ucp_tls_array_check_search_result, since we have only 2 possible bits

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +903,109 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
+                                      uint64_t *search_result,
+                                      uint64_t *tl_cfg_mask)
+{
+    *search_result |= ucp_config_tl_name_search(allow_list, tl_name, is_alias,
+                                                tl_cfg_mask);
+    if (ucp_tls_array_check_search_result(*search_result, search_stop_mask)) {
+        return 1;
+    }
+
+    return 0;
```
yosefe, line 918: seems over complicated

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +902,109 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
+                                                          tl_cfg_mask);
+            if (ucp_tls_array_found(search_result_tmp)) {
+                search_result |= UCP_TLS_ARRAY_SEARCH_RESULT_FOUND_AUX;
             }
         }
     }
 
-    return allow_list->mode == UCS_CONFIG_ALLOW_LIST_NEGATE;
+    if (ucp_enable_tl_aux(is_allow_mode, search_result) ||
+        ucp_tls_array_found(search_result)) {
```
yosefe, line 987: can we really reach this point if `ucp_tls_array_found(search_result)` is true?
I would expect we exited the function in line 956 because search_stop_mask always contains UCP_TLS_ARRAY_SEARCH_RESULT_FOUND

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +902,109 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
+         * considered in case of deny list. E.g:
+         * alias = { tl_name_1, tl_name_2:aux }, UCX_TLS=^...alias,... should
+         * not disable tl_name_2. */
+        if (ucp_tls_array_found(search_result_tmp) &&
+            ucp_check_resource_in_transports_list(allow_list, alias->alias, 1,
+                                                  search_stop_mask,
+                                                  &search_result,
+                                                  tl_cfg_mask)) {
+            return is_allow_mode;
+        } else if (ucp_enable_tl_aux(is_allow_mode, search_result_tmp)) {
```
yosefe, line 976: remove else

src/ucp/core/ucp_context.c
rakhmets: Yes, e.g: `UCX_TLS=^ud`. `ucp_tls_array_found(search_result)` is true. But we still need to enable UD as auxiliary transport.

src/ucp/core/ucp_context.c
rakhmets: Removed.

src/ucp/core/ucp_context.c
rakhmets: updated

src/ucp/core/ucp_context.c
```diff
@@ -865,62 +859,103 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
+    if (allow_list->mode == UCS_CONFIG_ALLOW_LIST_ALLOW_ALL) {
+        return 1;
+    }
+
+    ucs_assert(allow_list->array.count > 0);
+    enable = ucp_is_tl_in_transports_list(tl_name, &allow_list->array,
+                                          is_allow_mode, 0, rsc_flags,
+                                          tl_cfg_mask);
+
+    if (is_allow_mode || !ucs_string_set_contains(aux_tls, tl_name) || enable) {
```
yosefe, line 938: seems weird, I would expect something like
if (enable) return is_allow_mode;
can you pls document the logic?

src/ucp/core/ucp_context.c
```diff
@@ -865,62 +859,103 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
 }
 
 static int
-ucp_is_resource_in_transports_list(const char *tl_name,
-                                   const ucs_config_allow_list_t *allow_list,
-                                   uint8_t *rsc_flags, uint64_t *tl_cfg_mask)
+ucp_is_tl_in_transports_list(const char *tl_name,
```
yosefe, line 862: maybe this function should return combination of 3 bits - regardless of allow_mode:
1. transport is in the list as regular transport (no aux)
2. transport is in the "main" list as aux
3. transport is in the "alias" list as aux

Then, the caller function will combine the "allow_mode" logic with this result
this way we fully separate between "understanding the configuration" and "applying allow-mode logic"

rakhmets: The problem here is that there different interpretation of __transport is in the "alias" list as aux__ for allow and deny list.
E.g. `UCX_TLS=rc_x` should enable `rc_mlx5`, and `ud_mlx5` as auxiliary. `UCX_TLS=^rc_x` should only disable `rc_mlx5`, and do nothing with `ud_mlx5`. In the same time `UCX_TLS=ud_x` should disable `ud_mlx5`, but keep it available as auxiliary.
Taking into account this behavior we cannot make a simple check for three state inside the function. It can be done only using 4 states.

src/ucp/core/ucp_context.c
```diff
@@ -61,6 +61,9 @@
 
 #define UCP_CPU_EST_BCOPY_BW_DEFAULT (5800 * UCS_MBYTE)
 
+#define UCP_TL_AUX_SUFFIX    "aux"
```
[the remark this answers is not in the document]
rakhmets: What help message is meant?

src/ucp/core/ucp_context.c
```diff
@@ -792,45 +796,35 @@ static unsigned ucp_tl_alias_count(ucp_tl_alias_t *alias)
...
-ucp_config_is_tl_name_present(const ucs_config_allow_list_t *allow_list,
-                              const char *tl_name, int is_alias,
-                              uint8_t *rsc_flags, uint64_t *tl_cfg_mask)
+ucp_config_is_tl_name_present(const ucs_config_names_array_t *tl_array,
+                              const char *tl_name, int is_alias, int check_aux,
+                              uint64_t *tl_cfg_mask)
 {
     char strict_name[UCT_TL_NAME_MAX + 1];
 
     snprintf(strict_name, sizeof(strict_name), "\\%s", tl_name);
```
rakhmets, line 817: @yosefe 
Should we keep this feature (\\\\)?
If I understand right, this feature was introduced to resolve the issue when there are the same alias names and transport names.

yosefe: I prefer to keep it. any reason to remove?

src/ucp/core/ucp_context.c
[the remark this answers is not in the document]
yosefe: nvm, see it's not really documented in UCX_TLS help message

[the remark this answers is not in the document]
rakhmets: ok

src/ucp/core/ucp_context.c
rakhmets: Ok.
There are no serious reasons. Just thought about simplifying the code.

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +858,142 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
                                    const ucs_config_allow_list_t *allow_list,
+                                   const ucs_string_set_t *aux_tls,
                                    uint8_t *rsc_flags, uint64_t *tl_cfg_mask)
 {
-    uint64_t dummy_mask, tmp_tl_cfg_mask;
-    uint8_t tmp_rsc_flags, dummy_flags;
-    ucp_tl_alias_t *alias;
-    char info[32];
-    unsigned alias_arr_count;
+    uint8_t ret;
```
yosefe, line 933: rename ret -> search_result
or some other significant name

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +858,142 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
-        /* If the TL was found by its strict name - the result is known,
-           otherwise checking aliases is required */
-        return (allow_list->mode == UCS_CONFIG_ALLOW_LIST_ALLOW);
+    ret = ucp_is_tl_in_transports_list(tl_name, &allow_list->array,
+                                       tl_cfg_mask);
+
+    if (allow_list->mode == UCS_CONFIG_ALLOW_LIST_ALLOW) {
+        /* Enable the transport if UCX_TLS=tl_name, or alias={tl_name} and
+         * UCX_TLS=alias. */
+        if ((ret & UCP_TL_IN_TRANSPORTS_LIST_REGULAR) != 0) {
```
yosefe, line 946:
```
if (ret & UCP_TL_IN_TRANSPORTS_LIST_REGULAR) {
    return 1;
}
```

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +858,142 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
+        }
+
+        /* Enable the transport as an auxiliary if UCX_TLS=tl_name:aux, or
+         * alias={tl_name} and UCX_TLS=alias:aux, or alias={tl_name:aux} and
+         * UCX_TLS=alias. */
+        if (((ret & UCP_TL_IN_TRANSPORTS_LIST_AUX_IN_MAIN) != 0) ||
+            ((ret & UCP_TL_IN_TRANSPORTS_LIST_AUX_IN_ALIAS) != 0)) {
+            *rsc_flags |= UCP_TL_RSC_FLAG_AUX;
+            return 1;
+        }
```
yosefe, line 957:
```
if (ret & (UCP_TL_IN_TRANSPORTS_LIST_AUX_IN_MAIN |
           UCP_TL_IN_TRANSPORTS_LIST_AUX_IN_ALIAS)) {
    *rsc_flags |= UCP_TL_RSC_FLAG_AUX;
    return 1;
}
```

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +858,142 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
+    /* If alias1={tl_name1, tl_name2:aux} and alias2={tl_name2}, then
+     * UCX_TLS=^tl_name2 or UCX_TLS=^alias2 should not break the usage of
+     * tl_name1. */
+    if ((ret & UCP_TL_IN_TRANSPORTS_LIST_REGULAR) != 0) {
+        *rsc_flags |= UCP_TL_RSC_FLAG_AUX;
+        return 1;
     }
 
-    return allow_list->mode == UCS_CONFIG_ALLOW_LIST_NEGATE;
+    return 1;
```
yosefe, line 989: comment: transport was not found in the deny-list

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +858,142 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
+        ((ret & UCP_TL_IN_TRANSPORTS_LIST_AUX_IN_MAIN) != 0)) {
+        return 0;
+    }
+
+    /* If alias1={tl_name1, tl_name2:aux} and alias2={tl_name2}, then
+     * UCX_TLS=^tl_name2 or UCX_TLS=^alias2 should not break the usage of
+     * tl_name1. */
+    if ((ret & UCP_TL_IN_TRANSPORTS_LIST_REGULAR) != 0) {
+        *rsc_flags |= UCP_TL_RSC_FLAG_AUX;
+        return 1;
```
yosefe, line 986: shouldn't it be return 0?
since we return 1 in line 989 anyway

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +858,142 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
...
         }
+
+        return 1;
+    }
+
+    /* Only explicit indication in the deny list can completely disable
+     * transport which can be used as an auxiliary.
+     * E.g: UCX_TLS=^tl_name,tl_name:aux, or alias={tl_name} and
+     * UCX_TLS=^alias,alias:aux. */
+    if (((ret & UCP_TL_IN_TRANSPORTS_LIST_REGULAR) != 0) &&
```
yosefe, line 976: no need !=0 since it's bit check

src/ucp/core/ucp_context.c
```diff
@@ -864,63 +858,142 @@ static int ucp_is_resource_in_device_list(const uct_tl_resource_desc_t *resource
     return !!mask;
 }
 
+typedef enum ucp_tl_in_transports_list_ret {
+    UCP_TL_IN_TRANSPORTS_LIST_REGULAR      = UCS_BIT(0),
+    UCP_TL_IN_TRANSPORTS_LIST_AUX_IN_MAIN  = UCS_BIT(1),
+    UCP_TL_IN_TRANSPORTS_LIST_AUX_IN_ALIAS = UCS_BIT(2)
+} ucp_tl_in_transports_list_ret_t;
+
+static int ucp_tls_alias_is_present(ucp_tl_alias_t *alias, const char *tl_name,
```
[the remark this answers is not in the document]
rakhmets: The wrapper function hides a temporary variable and a dummy parameter passing to the target function. It is used three  times and simplifies the code.

src/ucp/core/ucp_context.c
rakhmets: We should return 1 here, to enable transport as an auxiliary.

src/ucp/core/ucp_context.c
```diff
@@ -973,19 +976,21 @@ ucp_is_resource_in_transports_list(const char *tl_name,
...
+                           UCP_TRANSPORTS_LIST_SEARCH_RESULT_AUX_IN_MAIN)) {
         return 0;
     }
 
     /* If alias1={tl_name1, tl_name2:aux} and alias2={tl_name2}, then
      * UCX_TLS=^tl_name2 or UCX_TLS=^alias2 should not break the usage of
-     * tl_name1. */
-    if ((ret & UCP_TL_IN_TRANSPORTS_LIST_REGULAR) != 0) {
+     * tl_name1 keeping tl_name2 enabled as auxiliary transport. */
+    if (search_result & UCP_TRANSPORTS_LIST_SEARCH_RESULT_PRIMARY) {
```
yosefe, line 988: why do we need to set *rsc_flags to UCP_TL_RSC_FLAG_AUX when we found a primary transport?
is it because it's an alias of another transport that was not disabled?
also what if in this example, `UCX_TLS=^alias1,tl_name2`?

src/ucp/core/ucp_context.c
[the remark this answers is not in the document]
yosefe: tl_alias_count and dummy_mask could be reused (declared once) in ucp_transports_list_search() function

src/ucp/core/ucp_context.c
rakhmets: E.g.: TLS_UCX=^ud -> UD transports are still enabled as auxiliary transports.

`UCX_TLS=^alias1,tl_name2` -> tl_name1 is disabled, tl_name2 is enabled as auxiliary. tl_name2 will be compeltely disabled, if: `UCX_TLS=^tl_name2,tl_name2:aux`,  `UCX_TLS=^alias2,tl_name2:aux`, `UCX_TLS=^alias2:aux,tl_name2`

src/ucp/core/ucp_context.c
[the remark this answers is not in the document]
rakhmets: Yes, they can. But they are only used for calling `ucp_tls_array_is_present`. And it makes `ucp_transports_list_search` a little easier to read, IMO.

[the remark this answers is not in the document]
yosefe: ok

src/ucp/core/ucp_context.c
```diff
@@ -1543,6 +1641,12 @@ static ucs_status_t ucp_fill_resources(ucp_context_h context,
         ucs_string_set_init(&avail_devices[dev_type]);
     }
 
+    ucs_string_set_init(&aux_tls);
+    status = ucp_fill_aux_tls(&aux_tls);
```
brminich, line 1645: maybe better init it later where it is needed? to avoid passing it down thru the stack

rakhmets: It is more convenient to have init/cleanup in a single method.
Otherwise, the code will be more complex.

yosefe: @brminich ucp_add_component_resources is called multiple times. Do we want to re-generate aux_tls every time?

brminich (dismissed):

SeyedMir:
So, `UCX_TLS=^ud` will completely disable UD? or it keeps it enabled for internal UCX usage?

rakhmets:
> So, `UCX_TLS=^ud` will completely disable UD? or it keeps it enabled for internal UCX usage?

`UCX_TLS=^ud` keeps UD transports enabled for internal UCX usage, e.g. RC.

`UCX_TLS=^ud,ud:aux` should be used to completely disable UD.

yosefe (approved):

PR #8779, opened 2022-12-20
GTEST: Adding explicit call to nvml init

shamisp:
There are gtest flows that do not call CUDA TL. As a result the nvml library does not get initialized. The side effect of this is that the trim function always returns the MAX size_t value.

### DIFF
--- a/test/gtest/common/mem_buffer.cc
+++ b/test/gtest/common/mem_buffer.cc
@@ -153,31 +153,48 @@ size_t mem_buffer::get_bar1_free_size()
 {
     /* All gtest CUDA tests explicitly assume that all memory allocations are
      * done on the device 0. The same assumption is followed here. */
-    size_t availible_size = SIZE_MAX;
+    size_t available_size = SIZE_MAX;
 
 #if HAVE_NVML_H
     nvmlReturn_t ret;
     nvmlDevice_t device;
     nvmlBAR1Memory_t bar1mem;
 
+    ret = nvmlInit_v2();
+    if (ret != NVML_SUCCESS) {
+        UCS_TEST_MESSAGE << "nvmlInit_v2 failed: " << nvmlErrorString(ret)
+                         << ", error code: " << ret;
+        return available_size;
+    }
+
     ret = nvmlDeviceGetHandleByIndex(0, &device);
     if (ret != NVML_SUCCESS) {
         /* For whatever reason we cannot open device handle.
          * As a result let's assume there is no limit on the size
          * and in the worse case scenario gtest will fail in runtime */
-        return availible_size;
+        UCS_TEST_MESSAGE << "nvmlDeviceGetHandleByIndex failed: "
+                         << nvmlErrorString(ret) << ", error code: " << ret;
+        return available_size;
     }
 
     ret = nvmlDeviceGetBAR1MemoryInfo(device, &bar1mem);
     if (ret != NVML_SUCCESS) {
         /* Similarly let's assume there is no limit on the size */
-        return availible_size;
+        UCS_TEST_MESSAGE << "nvmlDeviceGetBAR1MemoryInfo failed: "
+                         << nvmlErrorString(ret) << ", error code: " << ret;
+        return available_size;
     }
 
-    availible_size = (size_t)bar1mem.bar1Free;
+    available_size = (size_t)bar1mem.bar1Free;
+
+    ret = nvmlShutdown();
+    if (ret != NVML_SUCCESS) {
+        UCS_TEST_MESSAGE << "nvmlShutdown failed: " << nvmlErrorString(ret)
+                         << ", error code: " << ret;
+    }
 #endif
 
-    return availible_size;
+    return available_size;
 }
 
 void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type)
--- a/test/gtest/uct/uct_p2p_test.cc
+++ b/test/gtest/uct/uct_p2p_test.cc
@@ -185,7 +185,7 @@ void uct_p2p_test::test_xfer_multi_mem_type(send_func_t send, size_t min_length,
     /* Trim by BAR1 size if relevant */
     if ((mem_type == UCS_MEMORY_TYPE_CUDA) ||
         (mem_type == UCS_MEMORY_TYPE_CUDA_MANAGED)) {
-        /* Allocate 40% to accommodate send/receive buffers allocation */
+        /* Allocate 40% x 2 (RX/TX) to accommodate send/receive buffers allocation */
         max_length = ucs_min(max_length,
                 (size_t)(0.4 * mem_buffer::get_bar1_free_size()));
     }


test/gtest/common/mem_buffer.cc
```diff
@@ -160,21 +160,41 @@ size_t mem_buffer::get_bar1_free_size()
...
-        /* For whatever reason we cannot open device handle.
-         * As a result let's assume there is no limit on the size
-         * and in the worse case scenario gtest will fail in runtime */
+        UCS_TEST_MESSAGE << "nvmlInit_v2 failed: " << nvmlErrorString(ret) <<
+                            " ,error code: " << ret;
         return availible_size;
     }
 
-    ret = nvmlDeviceGetBAR1MemoryInfo(device, &bar1mem);
+    if (availible_size == SIZE_MAX) {
```
yosefe, line 170: 1. According to the code, availible_size would always be SIZE_MAX. Did you want to make it static?
2. can do "if (avaialble_size != SIZE_MAX) { return available_size; }" even before calling nvmlInit

test/gtest/common/mem_buffer.cc
```diff
@@ -160,21 +160,41 @@ size_t mem_buffer::get_bar1_free_size()
     nvmlDevice_t device;
     nvmlBAR1Memory_t bar1mem;
 
-    ret = nvmlDeviceGetHandleByIndex(0, &device);
+    ret = nvmlInit_v2();
```
yosefe, line 163: Is there nvmlInit() without _v2?

test/gtest/common/mem_buffer.cc
```diff
@@ -160,21 +160,41 @@ size_t mem_buffer::get_bar1_free_size()
...
     if (ret != NVML_SUCCESS) {
-        /* For whatever reason we cannot open device handle.
-         * As a result let's assume there is no limit on the size
-         * and in the worse case scenario gtest will fail in runtime */
+        UCS_TEST_MESSAGE << "nvmlInit_v2 failed: " << nvmlErrorString(ret) <<
+                            " ,error code: " << ret;
```
brminich, line 166: `, error code`?

test/gtest/common/mem_buffer.cc
```diff
@@ -160,21 +160,41 @@ size_t mem_buffer::get_bar1_free_size()
...
 
-    ret = nvmlDeviceGetBAR1MemoryInfo(device, &bar1mem);
+    if (availible_size == SIZE_MAX) {
+        ret = nvmlDeviceGetHandleByIndex(0, &device);
+        if (ret != NVML_SUCCESS) {
+            /* For whatever reason we cannot open device handle.
+             * As a result let's assume there is no limit on the size
+             * and in the worse case scenario gtest will fail in runtime */
+            UCS_TEST_MESSAGE << "nvmlDeviceGetHandleByIndex failed: " <<
+            nvmlErrorString(ret) << " ,error code: " << ret;
```
brminich, line 177: `, error code`?

test/gtest/common/mem_buffer.cc
```diff
@@ -160,21 +160,41 @@ size_t mem_buffer::get_bar1_free_size()
...
+            return availible_size;
+        }
+
+        ret = nvmlDeviceGetBAR1MemoryInfo(device, &bar1mem);
+        if (ret != NVML_SUCCESS) {
+            /* Similarly let's assume there is no limit on the size */
+            UCS_TEST_MESSAGE << "nvmlDeviceGetBAR1MemoryInfo failed: " <<
+            nvmlErrorString(ret) << " ,error code: " << ret;
```
brminich, line 185: `, error code`?

test/gtest/common/mem_buffer.cc
```diff
@@ -160,21 +160,41 @@ size_t mem_buffer::get_bar1_free_size()
...
+        }
+
+        availible_size = (size_t)bar1mem.bar1Free;
+    }
+
+    ret = nvmlShutdown();
     if (ret != NVML_SUCCESS) {
-        /* Similarly let's assume there is no limit on the size */
+        UCS_TEST_MESSAGE << "nvmlShutdown failed: " << nvmlErrorString(ret) <<
+                            " ,error code: " << ret;
```
brminich, line 195: `, error code`?

test/gtest/common/mem_buffer.cc
```diff
@@ -153,31 +153,54 @@ size_t mem_buffer::get_bar1_free_size()
...
+    }
+
     ret = nvmlDeviceGetHandleByIndex(0, &device);
     if (ret != NVML_SUCCESS) {
         /* For whatever reason we cannot open device handle.
          * As a result let's assume there is no limit on the size
          * and in the worse case scenario gtest will fail in runtime */
-        return availible_size;
+        UCS_TEST_MESSAGE << "nvmlDeviceGetHandleByIndex failed: " <<
+        nvmlErrorString(ret) << ", error code: " << ret;
```
yosefe, line 181: add indent
see https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=56897&view=logs&j=cc064a77-22b5-56bf-ecc0-70b5fe764261&t=aedfd754-44a6-53e6-6843-8659d800fee2 for code style issues

test/gtest/common/mem_buffer.cc
```diff
@@ -153,31 +153,54 @@ size_t mem_buffer::get_bar1_free_size()
...
     }
 
-    availible_size = (size_t)bar1mem.bar1Free;
+    available_size = (size_t)bar1mem.bar1Free;
+
+    ret = nvmlShutdown();
+    if (ret != NVML_SUCCESS) {
+        UCS_TEST_MESSAGE << "nvmlShutdown failed: " << nvmlErrorString(ret) <<
+                            ", error code: " << ret;
+        return available_size;
```
yosefe, line 199: seems this return is not needed

test/gtest/common/mem_buffer.cc
```diff
@@ -153,31 +153,48 @@ size_t mem_buffer::get_bar1_free_size()
 {
     /* All gtest CUDA tests explicitly assume that all memory allocations are
      * done on the device 0. The same assumption is followed here. */
-    size_t availible_size = SIZE_MAX;
+    size_t available_size = SIZE_MAX;
```
yosefe, line 156: why removed static?

test/gtest/ucp/test_ucp_mmap.cc
```diff
@@ -848,6 +848,11 @@ UCS_TEST_P(test_ucp_mmap, fixed) {
     ucs_status_t status;
     bool         is_dummy;
 
+    if (has_transport("cuda_ipc") ||
+        has_transport("cuda_copy")) {
+        UCS_TEST_SKIP_R("UCP_MEM_MAP_FIXED test is unsupported with CUDA transport");
+    }
+
```
yosefe, line 855: pls remove this, it should be fixed by different fixed address

test/gtest/uct/uct_p2p_test.cc
```diff
@@ -185,7 +185,7 @@ void uct_p2p_test::test_xfer_multi_mem_type(send_func_t send, size_t min_length,
     /* Trim by BAR1 size if relevant */
     if ((mem_type == UCS_MEMORY_TYPE_CUDA) ||
         (mem_type == UCS_MEMORY_TYPE_CUDA_MANAGED)) {
-        /* Allocate 40% to accommodate send/receive buffers allocation */
+        /* Allocate 40% x 2 (RX/TX) to accommodate send/receive buffers allocation */
```
yosefe, line 188: where is the x2 in the code?

test/gtest/common/mem_buffer.cc
shamisp: We discussed it on the call - depending what run on the machine, bar1 size can change. It is safer to query bar1 size on each allocation.

shamisp:
@yosefe can you please take a look ?

test/gtest/ucp/test_ucp_mmap.cc
yosefe: this worksaround another issue not related to BAR1 size

test/gtest/uct/uct_p2p_test.cc
yosefe: ok

test/gtest/ucp/test_ucp_mmap.cc
shamisp: Do you want this as a separate patch ?

test/gtest/common/mem_buffer.cc
shamisp: Yes, it is can and depends on machine where it runs. CUDA tl already calls init/cleanup and this call increases reference counter. Only if you run as a stand alone benchmark it may cause an actual init/cleanup calls.

test/gtest/ucp/test_ucp_mmap.cc
yosefe: yes

test/gtest/common/mem_buffer.cc
yosefe: ok

yosefe (approved):

PR #8783, opened 2023-01-01
UCP/PROTO: Unite select_param->op_flags and select_param->op_id

yosefe:
## Why
Free up more space in `ucp_proto_select_param_t` structure, so we could add reply_buffer.{mem_type, sys_dev} fields to it.
- We need to add reply_buffer mem info to fix atomic operations on GPU buffers, to take into account reply buffer memory type and not just send buffer memory type. UCP API supports any combination of memory types for send buffer and reply_buffer.
- Fix for atomic protocols v2 will be in next PR

## How
- Remove INTERNAL flag from select params - it could be a parameter for the protocol select functions, since it only affects debug tracing.
- The flags for different operations could overlap, so 4 upper bits in op_id field should suffice.
- Add helper functions to access flags and op_id
- Increase the reserved space (padding) to 2 bytes

### DIFF
--- a/src/ucp/proto/proto_debug.c
+++ b/src/ucp/proto/proto_debug.c
@@ -11,6 +11,8 @@
 #include "proto_init.h"
 #include "proto_select.inl"
 
+#include <ucp/am/ucp_am.inl>
+#include <ucp/rndv/proto_rndv.h>
 #include <ucs/arch/atomic.h>
 #include <ucs/datastruct/array.inl>
 #include <fnmatch.h>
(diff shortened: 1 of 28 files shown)


src/ucp/proto/proto_debug.c
```diff
@@ -317,54 +319,74 @@ void ucp_proto_select_dump_short(const ucp_proto_select_short_t *select_short,
...
-        if (op_flags & op_flag_bits) {
-            ucs_string_buffer_append_flags(strb, op_flags & op_flag_bits,
-                                           op_flag_names);
+        if (op_flags) {
+            if (ucp_proto_select_is_rndv_op(select_param)) {
+                ucs_string_buffer_append_flags(strb, op_flags, rndv_flag_names);
+            } else if (ucp_proto_select_is_am_op(select_param)) {
+                ucs_string_buffer_append_flags(strb, op_flags, am_flag_names);
+            }
             ucs_string_buffer_appendf(strb, ",");
```
rakhmets, line 383: This line can be removed.

src/ucp/proto/proto_init.h
```diff
@@ -106,4 +106,17 @@ ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
                            ucp_proto_perf_node_t *const tl_perf_node,
                            ucp_md_map_t reg_md_map);
 
+
+/**
+ * Check if prototocol initialization parameters contain one of the specified
```
rakhmets, line 111: Typo: `prototocol` -> `protocol`

rakhmets (dismissed):

rakhmets (approved):

PR #8791, opened 2023-01-03
UCP/AMO/SW: Fix CPU/device atomics selection with protov2

yosefe:
## Why
Fix `Unsupported: got software atomic request while device atomics are selected on worker nnn` error with protov2 atomics.

## How
Disable SW atomics if target worker uses device atomics


yosefe:
@evgeny-leksikov @brminich  can you pls take a look?

brminich (approved):

evgeny-leksikov (approved):

Issue #8762, opened 2022-12-12
Include arm64 in the UCX and JUCX published artifacts


abellina:
In the RAPIDS Accelerator for Apache Spark we use/point users to published artifacts for UCX here https://github.com/openucx/ucx/releases and the JUCX jar here https://search.maven.org/artifact/org.openucx/jucx.

We would like for arm64 versions of these artifacts to be published as well.

artemry-nv:
CC @Al3xR01

PR #8796, opened 2023-01-04
AZP/RELEASE: Add ARM release

Al3xR01:
## What
Build and release ARM-based artifacts.

## Why ?
https://github.com/openucx/ucx/issues/8762


petro-rudenko (commented):
With jucx_publish i think there's a problem: https://github.com/openucx/ucx/pull/8796/files#diff-62bc3edaaacedb94d2bd4b5ef1f01ca3bc99d1b6a78299e1ceb740e6c7440f5fL68

It'll first publish x86 jucx-1.15.0.jar jar and then try to publish arm jucx-1.15.0.jar - that will fail, because x86 is already published with the same version.

Now bigger question: do we want to publish (jucx-1.15.0-x86.jar and jucx-1.15.0-arm64.jar) or we want to publish only jucx-1.15.0.jar that would have jucx_x86.so and jucx_arm64.so inside it and in runtime will load correct shared library based on the platform it's running). @abellina @yosefe WDYT?

abellina:
@petro-rudenko I'd vote for a single jar since jucx's jar is fairly small. It would make it a lot more user friendly to not have to specify the architecture as the classifier, and should hopefully simplify the release and documentation.

petro-rudenko:
So here's what needs to change:
1. For arm64 bulild jucx but not publish. Copy libjucx.so to libjucx_aarch64.so
2. Build jucx on x86 but not publish. Copy libjucx.so to libjucx_amd64.so
3. Change https://github.com/openucx/ucx/blob/master/bindings/java/src/main/java/org/openucx/jucx/NativeLibs.java#L27 to have something:
```java
 loadLibrary(JUCX + "_" + System.getProperty("os.arch"));
```
4. Change https://github.com/openucx/ucx/blob/master/bindings/java/pom.xml.in#L235 to something like:
```<include>libjucx_*.so</include>```
5. Run publish job that will contain both libjucx_{amd64, aarch64}.so files

File: buildlib/az-distro-release.yml [hunk 1]
```diff
@@ -67,13 +84,18 @@ jobs:
...
           set -eEx
-          cd pkg-build
-          echo 10 > debian/compat   # https://www.debian.org/doc/manuals/maint-guide/dother.en.html#compat
-          sed -i 's|./contrib|../contrib|' debian/rules  # Hack
-          dpkg-buildpackage -us -uc
-          find .. -name '*.deb'
-          find .. -name '*.deb' -exec cp {} "../${AZ_ARTIFACT_NAME}" \;
-          dpkg-deb -I "../${AZ_ARTIFACT_NAME}"
+          ./autogen.sh
+          ./contrib/configure-release --with-cuda --with-java=no
```
artemry-nv, line 88: Is it OK `--with-java=no`?

buildlib/dockers-arm/centos-release.Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+        --skip-distro-check \
+        --without-ucx \
+        --without-hcoll \
+        --without-openmpi \
+        --without-sharp \
+        --skip-distro-check \
+        --distro ${MOFED_OS} \
+    && rm -rf ${MOFED_DIR} && rm -rf *.tgz
+
+ENV CPATH /usr/local/cuda/include:${CPATH}
```
artemry-nv, line 57: Aren't these paths already in the default env?

buildlib/dockers-arm/push-release-images.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -eEx
+
+# shellcheck disable=SC2086
+basedir=$(cd "$(dirname $0)" && pwd)
+
+registry=harbor.mellanox.com/ucx/aarch64
```
artemry-nv, line 6: Is it correct harbor url? Shouldn't be the RDMZ one?

buildlib/dockers-arm/ubuntu-release.Dockerfile
```diff
@@ -0,0 +1,50 @@
+ARG CUDA_VERSION=10.1
```
artemry-nv, line 1: Default looks obsolete, should be 11.x I suppose

buildlib/dockers-arm/ubuntu-release.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+        --without-openmpi \
+        --without-sharp && \
+    rm -rf ${MOFED_DIR} && rm -rf *.tgz
+
+ENV CPATH /usr/local/cuda/include:${CPATH}
+ENV LD_LIBRARY_PATH /usr/local/cuda/lib64:/usr/local/cuda/compat:${LD_LIBRARY_PATH}
+ENV LIBRARY_PATH /usr/local/cuda/lib64:/usr/local/cuda/compat:${LIBRARY_PATH}
+ENV PATH /usr/local/cuda/compat:${PATH}
+
+RUN ml_stub=$(find /usr -name libnvidia-ml.so) && ln -s $ml_stub /usr/lib/aarch64-linux-gnu/libnvidia-ml.so.1
```
artemry-nv, line 50: Do you know what is this for? Small comment would be helpful

artemry-nv (approved):

buildlib/az-distro-release.yml
Al3xR01: Java package gets built by a separate job.

buildlib/dockers-arm/push-release-images.sh
Al3xR01: RDMZ Harbor is a mirror of the lab one. Can't push directly to it.

buildlib/dockers-arm/ubuntu-release.Dockerfile
Al3xR01: The default values in the Dockerfile are overridden by the values from docker-compose.yml.

buildlib/dockers-arm/ubuntu-release.Dockerfile
Al3xR01: I got this legacy as it is ¯\\_(ツ)_/¯

buildlib/dockers-arm/centos-release.Dockerfile
Al3xR01: Not the CPATH.

Al3xR01:
@yosefe Can we please merge this one?

buildlib/az-GitHubDraft.yml
```diff
@@ -0,0 +1,47 @@
+# Create an empty draft to avoid race condition in distro release
```
yosefe, line 1: how is it related to ARM build?

buildlib/dockers-arm/centos-release.Dockerfile
```diff
@@ -0,0 +1,63 @@
+ARG CUDA_VERSION
```
yosefe, line 1: why need different compose file for ARM?

yosefe (changes requested):
IMO need to split it:
1. Add docker files for ARM, including rename to x86_64
2. Build DEB/RPM artifacts on ARM
3. Build JUCX on ARM

2 and 3 depend on 1

buildlib/dockers-arm/centos-release.Dockerfile
Al3xR01: OS and MOFED support matrix differs between hardware platforms.

buildlib/az-GitHubDraft.yml
Al3xR01: It's an existing job, just separated into 2 platforms.

Al3xR01:
Splitting into:
https://github.com/openucx/ucx/pull/8894
https://github.com/openucx/ucx/pull/8895
https://github.com/openucx/ucx/pull/8896

PR #8798, opened 2023-01-05
UCT/CUDA: Detect hopper architecture

Akshay-Venkatesh:
## Why
Without this detection, `ib` gen5 is preferred over `cuda_ipc` which leads to lower bandwidth between p2p accessible Hopper GPUs. 

## What?

- Detect Hopper architecture and provide a more accurate cuda_ipc bandwidth estimation
- Add env var to allow user to explicitly set cuda_ipc bandwidth

### DIFF
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
@@ -47,6 +47,10 @@ static ucs_config_field_t uct_cuda_ipc_iface_config_table[] = {
      "Max number of cuda events. -1 is infinite",
      ucs_offsetof(uct_cuda_ipc_iface_config_t, max_cuda_ipc_events), UCS_CONFIG_TYPE_UINT},
 
+    {"BW", "auto",
+     "Effective p2p memory bandwidth",
+     ucs_offsetof(uct_cuda_ipc_iface_config_t, bandwidth), UCS_CONFIG_TYPE_BW},
+
     {NULL}
 };
 
@@ -106,6 +110,8 @@ static double uct_cuda_ipc_iface_get_bw()
         return 250000.0 * UCS_MBYTE;
     case UCT_CUDA_BASE_GEN_A100:
         return 300000.0 * UCS_MBYTE;
+    case UCT_CUDA_BASE_GEN_H100:
+        return 400000.0 * UCS_MBYTE;
     default:
         return 6911.0  * UCS_MBYTE;
     }
@@ -218,7 +224,7 @@ static ucs_status_t uct_cuda_ipc_iface_query(uct_iface_h tl_iface,
 
     iface_attr->latency                 = ucs_linear_func_make(1e-6, 0);
     iface_attr->bandwidth.dedicated     = 0;
-    iface_attr->bandwidth.shared        = uct_cuda_ipc_iface_get_bw();
+    iface_attr->bandwidth.shared        = iface->config.bandwidth;
     iface_attr->overhead                = 7.0e-6;
     iface_attr->priority                = 0;
 
@@ -500,6 +506,8 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_ipc_iface_t, uct_md_h md, uct_worker_h worke
     self->config.enable_cache        = config->enable_cache;
     self->config.enable_get_zcopy    = config->enable_get_zcopy;
     self->config.max_cuda_ipc_events = config->max_cuda_ipc_events;
+    self->config.bandwidth           = UCS_CONFIG_DBL_IS_AUTO(config->bandwidth) ?
+                                       uct_cuda_ipc_iface_get_bw() : config->bandwidth;
 
     ucs_mpool_params_reset(&mp_params);
     mp_params.elem_size       = sizeof(uct_cuda_ipc_event_desc_t);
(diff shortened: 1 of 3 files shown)


Akshay-Venkatesh:
cc @bureddy @yosefe

bureddy (dismissed):

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -47,6 +47,15 @@ static ucs_config_field_t uct_cuda_ipc_iface_config_table[] = {
...
      ucs_offsetof(uct_cuda_ipc_iface_config_t, max_cuda_ipc_events), UCS_CONFIG_TYPE_UINT},
 
+    {"OVERRIDE_BW", "n",
+     "Override bandwidth detection in cuda_ipc. User is expected to set UCX_CUDA_IPC_BW",
+     ucs_offsetof(uct_cuda_ipc_iface_config_t, override_bw),
+     UCS_CONFIG_TYPE_BOOL},
+
+    {"BW", "20000MBs",
+     "Effective memory bandwidth",
+     ucs_offsetof(uct_cuda_ipc_iface_config_t, bandwidth), UCS_CONFIG_TYPE_BW},
```
bureddy, line 57: Can we simplify with a single variable i.e UCX_CUDA_IPC_BW with default 0. if the user sets it non-zero then override it.

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -47,6 +47,10 @@ static ucs_config_field_t uct_cuda_ipc_iface_config_table[] = {
      "Max number of cuda events. -1 is infinite",
      ucs_offsetof(uct_cuda_ipc_iface_config_t, max_cuda_ipc_events), UCS_CONFIG_TYPE_UINT},
 
+    {"BW", "0MBs",
```
brminich, line 50: pls use `auto` instead of `0`

Akshay-Venkatesh: @brminich if we use `auto` would we be able to check [this](https://github.com/openucx/ucx/pull/8798/files#diff-04769c9662cb2e769aff68bed88f18cef6d6bdaa369b178d4223d769cecb7d4fR227) condition?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -500,6 +507,7 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_ipc_iface_t, uct_md_h md, uct_worker_h worke
     self->config.enable_cache        = config->enable_cache;
     self->config.enable_get_zcopy    = config->enable_get_zcopy;
     self->config.max_cuda_ipc_events = config->max_cuda_ipc_events;
+    self->config.bandwidth           = config->bandwidth;
```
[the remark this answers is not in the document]
Akshay-Venkatesh: @brminich any particular benefit in moving bw initialization here? for most other transports, bandwidth values seem to be populated in iface_query function.

Akshay-Venkatesh:
@brminich somewhat unexpected that we're seeing [leak errors ](https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=56486&view=logs&j=f97d6e39-20b4-5224-cb57-ac6707fa3b18&t=65630ad1-e2d8-5140-3757-25a4f549d6a1&l=43390) from this PR as it shouldn't be touching memory. Would it be possible to rerun this test?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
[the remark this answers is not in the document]
yosefe: agree with @brminich , since we save bandwidth on the face already, better make it the "final" value

yosefe:
> @brminich somewhat unexpected that we're seeing [leak errors ](https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=56486&view=logs&j=f97d6e39-20b4-5224-cb57-ac6707fa3b18&t=65630ad1-e2d8-5140-3757-25a4f549d6a1&l=43390) from this PR as it shouldn't be touching memory. Would it be possible to rerun this test?

@Akshay-Venkatesh the failure is https://github.com/openucx/ucx/issues/8800
Pls force push the CR fixes since it's a small PR

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
brminich: yes, but need to compare with `UCS_CONFIG_DBL_AUTO` instead

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -500,6 +506,8 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_ipc_iface_t, uct_md_h md, uct_worker_h worke
     self->config.enable_cache        = config->enable_cache;
     self->config.enable_get_zcopy    = config->enable_get_zcopy;
     self->config.max_cuda_ipc_events = config->max_cuda_ipc_events;
+    self->config.bandwidth           = (config.bandwidth > UCS_CONFIG_DBL_AUTO) ?
```
yosefe, line 509: use UCS_CONFIG_DBL_IS_AUTO

yosefe (dismissed):

brminich (approved):

yosefe (approved):

yosefe:
@Akshay-Venkatesh we probably need to port it to v1.14.x

PR #8806, opened 2023-01-09
UCT/CUDA: Leave cuda base_addr unchanged if dmabuf is queried to avoid cuda_ipc errors

Akshay-Venkatesh:
## What/Why ?
If dmabuf fd query is supported, base_address and alloc_length is altered to mapping_base_addr and mapping_base_size returned by cuda. The latter two may correspond to a larger underlying allocation from which cuda sub-allocates but the base_addr of the underlying allocation would not be accessible to cuda_ipc even though all sub-allocations from that allocation may have the same dmabuf_fd. For this reason, base_addr memory attribute should not be altered and should continue to correspond to the base of the just the cudamalloc sub-allocation.

### DIFF
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.c
@@ -425,6 +425,7 @@ uct_cuda_copy_md_mem_query(uct_md_h tl_md, const void *address, size_t length,
     uct_cuda_copy_md_t *md = ucs_derived_of(tl_md, uct_cuda_copy_md_t);
     unsigned value         = 1;
     ucs_memory_info_t addr_mem_info;
+    ucs_memory_info_t dmabuf_addr_mem_info;
     const char *cu_err_str;
     ucs_status_t status;
     CUresult cu_err;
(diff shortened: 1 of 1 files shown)


Akshay-Venkatesh:
cc @yosefe

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -478,12 +479,15 @@ uct_cuda_copy_md_mem_query(uct_md_h tl_md, const void *address, size_t length,
...
 
     if (mem_attr->field_mask & UCT_MD_MEM_ATTR_FIELD_DMABUF_FD) {
-        mem_attr->dmabuf_fd = uct_cuda_copy_md_get_dmabuf_fd(&addr_mem_info);
-    }
-
-    if (mem_attr->field_mask & UCT_MD_MEM_ATTR_FIELD_DMABUF_OFFSET) {
-        mem_attr->dmabuf_offset = UCS_PTR_BYTE_DIFF(addr_mem_info.base_address,
-                                                    address);
+        dmabuf_addr_mem_info = addr_mem_info;
+        mem_attr->dmabuf_fd  = uct_cuda_copy_md_get_dmabuf_fd(&dmabuf_addr_mem_info);
```
yosefe, line 483: seems this change does not change the behavior because dmabuf_addr_mem_info is not modified by uct_cuda_copy_md_get_dmabuf_fd

PR #8812, opened 2023-01-12
UCP/CORE: reduce call times of `ucp_worker_discard_uct_ep_destroy_progress`

jeynmann:
## What
optimize algorithm of `ucp_worker_discard_uct_ep_destroy_progress` to reduce the counts of called.

## Why ?
Customers want to reduce the time cost of disconnection/connection.
`ucp_worker_discard_uct_ep_destroy_progress` has been added to progress queue 4k times for 1k connections with DC mode.
It takes nearly 50ms to progress it.

## How ?
Add a discard `uct_ep` request list in worker and call it only once.


yosefe:

src/ucp/core/ucp_worker.h
```diff
@@ -253,6 +253,12 @@ struct ucp_worker_cm {
 };
 
 
+typedef struct ucp_worker_discard_req {
+    ucs_list_link_t req_list;
+    void            *req;
+} ucp_worker_discard_req_t;
```
evgeny-leksikov, line 259: can we re-use some space in ucp_request to avoid this structure and extra allocation?

evgeny-leksikov (commented):
how much speedup we get with this approach?

jeynmann:
> how much speedup we get with this approach?

About 50ms for 1k connections. (~70ms -> ~20ms)

src/ucp/core/ucp_worker.c
```diff
@@ -2579,8 +2623,12 @@ static int
...
 {
     if ((elem->arg == arg) &&
-        ((elem->cb == ucp_worker_discard_uct_ep_destroy_progress) ||
-         (elem->cb == ucp_worker_discard_uct_ep_progress))) {
+        (elem->cb == ucp_worker_discard_uct_ep_destroy_progress)) {
+        return 1;
+    }
+
+    if ((elem->arg == arg) &&
+        (elem->cb == ucp_worker_discard_uct_ep_progress)) {
```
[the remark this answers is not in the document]
jeynmann: done.

src/ucp/core/ucp_worker.h
jeynmann: done.

jeynmann:
@evgeny-leksikov 
All comments have been fixed.
It has passed all CI-checks now.

src/ucp/core/ucp_worker.c
```diff
@@ -2629,6 +2666,11 @@ static void ucp_worker_discard_uct_ep_cleanup(ucp_worker_h worker)
          * could move a discard operation to the progress queue */
         ucs_callbackq_remove_if(&worker->uct->progress_q,
                                 ucp_worker_discard_remove_filter, req);
+        if (req->send.discard_uct_ep.state ==
+            UCP_WORKER_DISCARD_UCT_EP_FLUSHED) {
+            req->send.discard_uct_ep.state = UCP_WORKER_DISCARD_UCT_EP_FINISHED;
```
evgeny-leksikov, line 2671: do wee really need `req->send.discard_uct_ep.state` variable?
IMO at this point, we need just to release all resources gracefully but if `state != UCP_WORKER_DISCARD_UCT_EP_FLUSHED` then `req` is leaked

src/ucp/core/ucp_worker.c
```diff
@@ -2468,33 +2469,39 @@ static void ucp_worker_discard_uct_ep_complete(ucp_request_t *req)
...
+
+    req->send.discard_uct_ep.state = UCP_WORKER_DISCARD_UCT_EP_FINISHED;
     ucp_worker_discard_uct_ep_complete(req);
 
-    ucs_assert(kh_value(&worker->discard_uct_ep_hash, iter) == req);
-    kh_del(ucp_worker_discard_uct_ep_hash, &worker->discard_uct_ep_hash, iter);
+    return 1;
+}
+
+static unsigned ucp_worker_discard_uct_ep_destroy_progress(void *arg)
```
evgeny-leksikov, line 2490: `uct_ep` -> `uct_eps` since this function destroys all discard EPs

src/ucp/core/ucp_worker.c
```diff
@@ -2468,33 +2469,39 @@ static void ucp_worker_discard_uct_ep_complete(ucp_request_t *req)
...
+    UCS_ASYNC_BLOCK(&worker->async);
+    kh_foreach_value(&worker->discard_uct_ep_hash, req, {
+        if (req->send.discard_uct_ep.state ==
+            UCP_WORKER_DISCARD_UCT_EP_FLUSHED) {
+            kh_del(ucp_worker_discard_uct_ep_hash,
+                   &worker->discard_uct_ep_hash, __i);
+            ucp_worker_discard_uct_ep_destroy_once(req);
+        }
+    });
+    worker->discard_uct_ep_comp = 0;
```
evgeny-leksikov, line 2504: please do `worker->discard_uct_ep_comp--` in the loop then assert that it's `0` in the end

jeynmann: done

src/ucp/core/ucp_worker.c
jeynmann: done

src/ucp/core/ucp_worker.c
jeynmann: 1. I think we need this variable to separate the requests who don't finish `uct_ep_flush`, as there are at least 2 async process during discarding `uct_ep`.
image
2. The original code here is to call `ucp_worker_discard_uct_ep_complete` only if the callback is `ucp_worker_discard_uct_ep_destroy_progress` or `ucp_worker_discard_uct_ep_progress`. Here we mark the state as `UCP_WORKER_DISCARD_UCT_EP_FLUSHED`, then remove it in the loop. The algorithm is the same as before. I'm afraid of that it might cause double-free if the callback is not anyone of the 2 functions above. See [here](https://github.com/openucx/ucx/blob/master/src/ucp/core/ucp_worker.c#L2579)
``` c
static int
ucp_worker_discard_remove_filter(const ucs_callbackq_elem_t *elem, void *arg)
{
    if ((elem->arg == arg) &&
        ((elem->cb == ucp_worker_discard_uct_ep_destroy_progress) ||
         (elem->cb == ucp_worker_discard_uct_ep_progress))) {
        ucp_worker_discard_uct_ep_complete((ucp_request_t*)elem->arg);
        return 1;
    }

    return 0;
}
```

evgeny-leksikov: AFAIU `ucp_worker_destroy` should cleanup resources with blocked async context, if `discard_remove_filter` works properly, then it should not cause double free

jeynmann: Got it. Done.

evgeny-leksikov: > do wee really need `req->send.discard_uct_ep.state` variable?

the field looks redundant to me

jeynmann: Perhaps I didn't make it clear. 
For example, there are 5 eps and we don't use `req->send.discard_uct_ep.state`.
3 of them calls `ucp_worker_discard_uct_ep_progress` well, while the others are pending.
However, the first 3 eps already ___flushed uct eps___ and put `ucp_worker_discard_uct_eps_destroy_progress` into the progress queue.
When the worker processs `ucp_worker_discard_uct_eps_destroy_progress`, it will complete all eps' discard requests. Then ep4 and ep5 would be discard without flush.
image

evgeny-leksikov: ah.. I think I got the issue. Then logic around `discard_uct_ep_comp` looks wrong to me:
 - the counter should be incremented before putting into the hash
 - flush completion callback should decremented the the counter
 - if the counter equals to `0`, then  handle all requests in the hash since all of them are "flushed"

WDYT?

evgeny-leksikov: another option which looks even better could be just put into the hash only flushed EPs, then the counter is not needed as well. You can decide schedule the progress callback or not just by checking khash empty condition

jeynmann: The `discard_uct_ep_hash` doestn't save eps only, but also saves the requests.
There is another case. If all requests are pending, we need to clean up the worker according to the hash table.
Maybe I should rename `discard_uct_ep_comp` to `uct_ep_flush_comp`, right?
image

The counter is needed because it's used to reduce the pending functions. As the following figure shows, ep2+ep3's callback won't be added into the progress queue.

image

evgeny-leksikov: > The discard_uct_ep_hash doestn't save eps only, but also saves the requests.

in your case every request represents every single EP, so it's the same

> The counter is needed because it's used to reduce the pending functions. As the following figure shows, ep2+ep3's callback won't be added into the progress queue.

yes, all the logic is implemented in `ucp_worker_discard_uct_ep_pending_cb`, so, if use shared `uct_comp` (probably on worker) instead of `req->send.state.uct_comp`, then `ucp_worker_discard_uct_ep_flush_comp` will be invoked only once for all discarded and flushed EPs

src/ucp/core/ucp_worker.c
```diff
@@ -2468,33 +2469,41 @@ static void ucp_worker_discard_uct_ep_complete(ucp_request_t *req)
...
+        if (req->send.discard_uct_ep.state ==
+            UCP_WORKER_DISCARD_UCT_EP_FLUSHED) {
+            kh_del(ucp_worker_discard_uct_ep_hash, &worker->discard_uct_ep_hash,
+                   __i);
+            ucp_worker_discard_uct_ep_destroy_once(req);
+            --worker->discard_uct_ep_comp;
+        }
+    });
+    ucs_assertv(worker->discard_uct_ep_comp == 0,
+                "worker->discard_uct_ep_comp=%u", worker->discard_uct_ep_comp);
```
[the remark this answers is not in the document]
jeynmann: done

src/ucp/core/ucp_worker.c
jeynmann: I've tried to adjust the codes, but I met some issues.
1. Add counter when add ep to hash and reduce counter when ep is flushed.
    If flush is neither 'no resource' nor 'in progress', the counter is always changes as 0 -> 1 -> 0.
    As a result, we add a callback for every ep.
    That means it can not reduce the counts of the callback functions.
    And the optimization just falls back. 
2. Add ep into hash only after it's flushed.
    - If we destroy all eps in `ucp_worker_destroy`, it will result in double destroy for those eps which are destroyed in `ucp_worker_discard_uct_ep_destroy_progress` already.
    - If we destroy all eps in hash only in `ucp_worker_destroy`, it will leak those eps which are in progress of `uct_ep_flush`.

evgeny-leksikov: I meant to rework `static void ucp_worker_discard_uct_ep_flush_comp(uct_completion_t *self)`. If `self` would be `worker->uct_comp' shared across all discarding endpoints/requests then this callback is invoked only when all endpoints are flushed and then you can cleanup all EPs stored in the hash. I think it could significantly simplify logic with the same result.

jeynmann: Got it.
And I found another way: compare the hash size with the ep flush counter `uct_ep_flush_comp`.

yosefe:
@zizhao let's put this PR on hold. I want to replace it by refactoring of callbacks remove_if so all slowpath callbacks will be fast to add/remove, and not only the discard callback.

leibin2014:
> @zizhao let's put this PR on hold. I want to replace it by refactoring of callbacks remove_if so all slowpath callbacks will be fast to add/remove, and not only the discard callback.

@yosefe So can we close this one for now?

jeynmann:
This will be closed, because `9098` which improved the performance of callbacks has been merged.
Profile data for 1k connections:

| version | timecost|
|---------|---------|
| 1.13rc1 | 80662us |
| PR 9098 | 28823us |
| PR 8812 | 23270us |

PR #8814, opened 2023-01-12
UCT/TCP: Fixed sysfs_path expansion for TCP iface

shasson5:
## What
Fixed sysfs_path expansion for TCP iface.

## Why ?
In order to find sysfs path for a device, there are few stages:
1) Getting the unresolved path from sys/class. (for example: `/sys/class/net/enp3s0f0s0`).
2) Removing all symbolic links. for example:
    `/sys/devices/pci0000:00/0000:00:00.0/0000:01:00.0/0000:02:00.0/0000:03:00.0/mlx5_core.sf.2/net/enp3s0f0s0`
3) Find the sysfs root for the device:
   `/sys/devices/pci0000:00/0000:00:00.0/0000:01:00.0/0000:02:00.0/0000:03:00.0`

For TCP iface, we didn't used this process, which caused some architectures to fail (ARM Blufield in this case).
This PR fixes this issue.

### DIFF
--- a/src/ucs/sys/topo/base/topo.c
+++ b/src/ucs/sys/topo/base/topo.c
@@ -257,7 +257,7 @@ ucs_status_t ucs_topo_get_device_bus_id(ucs_sys_device_t sys_dev,
 }
 
 static ucs_status_t
-ucs_topo_get_sysfs_path(ucs_sys_device_t sys_dev, char *path, size_t max)
+ucs_topo_sys_dev_to_sysfs_path(ucs_sys_device_t sys_dev, char *path, size_t max)
 {
     const size_t prefix_length = strlen(UCS_TOPO_SYSFS_PCI_PREFIX);
     char link_path[PATH_MAX];
@@ -349,12 +349,12 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
         return UCS_OK;
     }
 
-    status = ucs_topo_get_sysfs_path(device1, path1, sizeof(path1));
+    status = ucs_topo_sys_dev_to_sysfs_path(device1, path1, sizeof(path1));
     if (status != UCS_OK) {
         return status;
     }
 
-    status = ucs_topo_get_sysfs_path(device2, path2, sizeof(path2));
+    status = ucs_topo_sys_dev_to_sysfs_path(device2, path2, sizeof(path2));
     if (status != UCS_OK) {
         return status;
     }
@@ -388,7 +388,7 @@ ucs_topo_get_memory_distance_sysfs(ucs_sys_device_t device,
         return ucs_topo_get_memory_distance_default(device, distance);
     }
 
-    status = ucs_topo_get_sysfs_path(device, path, sizeof(path));
+    status = ucs_topo_sys_dev_to_sysfs_path(device, path, sizeof(path));
     if (status != UCS_OK) {
         return ucs_topo_get_memory_distance_default(device, distance);
     }
@@ -741,4 +741,52 @@ double ucs_topo_get_pci_bw(const char *dev_name, const char *sysfs_path)
 out_max_bw:
     ucs_debug("%s: pci bandwidth undetected, using maximal value", dev_name);
     return DBL_MAX;
+}
+
+const char *ucs_topo_resolve_sysfs_path(const char *dev_path, char *path_buffer)
+{
+    const char *detected_type = NULL;
+    char device_file_path[PATH_MAX];
+    char *sysfs_realpath;
+    struct stat st_buf;
+    char *sysfs_path;
+
+    /* realpath name is expected to be like below:
+     * PF: /sys/devices/.../0000:03:00.0/<interface_type>/<dev_name>
+     * SF: /sys/devices/.../0000:03:00.0/<UUID>/<interface_type>/<dev_name>
+     */
+
+    sysfs_realpath = realpath(dev_path, path_buffer);
+    if (sysfs_realpath == NULL) {
+        goto out_undetected;
+    }
+
+    /* Try PF: strip 2 components */
+    sysfs_path = ucs_dirname(sysfs_realpath, 2);
+    ucs_snprintf_safe(device_file_path, sizeof(device_file_path), "%s/device",
+                      sysfs_path);
+
+    if (!stat(device_file_path, &st_buf)) {
+        detected_type = "PF";
+        goto out_detected;
+    }
+
+    /* Try SF: strip 3 components (one more) */
+    sysfs_path = ucs_dirname(sysfs_path, 1);
+    ucs_snprintf_safe(device_file_path, sizeof(device_file_path), "%s/device",
+                      sysfs_path);
+
+    if (!stat(device_file_path, &st_buf)) {
+        detected_type = "SF";
+        goto out_detected;
+    }
+
+out_undetected:
+    ucs_debug("%s: sysfs path undetected", dev_path);
+    return NULL;
+
+out_detected:
+    ucs_debug("%s: %s sysfs path is '%s'\n", dev_path, detected_type,
+              sysfs_path);
+    return sysfs_path;
 }
--- a/src/ucs/sys/topo/base/topo.h
+++ b/src/ucs/sys/topo/base/topo.h
@@ -202,6 +202,19 @@ double ucs_topo_get_pci_bw(const char *dev_name, const char *sysfs_path);
 
 
 /**
+ * Returns sysfs path of a given device. for example:
+ * input:  '/sys/class/infiniband/mlx5_1'
+ * output: '/sys/devices/pci0000:80/0000:80:01.1/0000:83:00.0'
+ *
+ * @param [in]  dev_path    Device file path.
+ * @param [out] path_buffer Filled with the result path.
+ *
+ * @return Pointer to sysfs path or NULL on error.
+ */
+const char *
+ucs_topo_resolve_sysfs_path(const char *dev_path, char *path_buffer);
+
+/**
  * Get the name of a given system device. If the name was never set, it defaults
  * to the BDF representation of the system device bus id.
  *
--- a/src/uct/tcp/tcp_iface.c
+++ b/src/uct/tcp/tcp_iface.c
@@ -218,6 +218,50 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
     return 1;
 }
 
+static ucs_status_t
+uct_tcp_iface_parse_virtual_dev(const struct dirent *entry, void *ctx)
+{
+    static const char *low_level_dev_prefix = "lower_";
+    ucs_string_buffer_t *dev_path           = ctx;
+
+    if (!strncmp(entry->d_name, low_level_dev_prefix,
+                 strlen(low_level_dev_prefix))) {
+        ucs_string_buffer_appendf(dev_path, "/%s", entry->d_name);
+        return UCS_ERR_CANCELED;
+    }
+
+    return UCS_OK;
+}
+
+static const char *
+uct_tcp_iface_get_sysfs_path(const char *dev_name, char *path_buffer)
+{
+    ucs_string_buffer_t dev_path = UCS_STRING_BUFFER_INITIALIZER;
+    const char *sysfs_path;
+    ucs_status_t status;
+
+    /* Find and return the correct device sysfs path:
+     * 1) For regular device, use regular sysfs form.
+     * 2) For virtual device (RoCE LAG/VLAN), search for symbolic link of the
+     *    form "lower_*" */
+    ucs_string_buffer_appendf(&dev_path, "%s/%s", UCT_TCP_IFACE_NETDEV_DIR,
+                              dev_name);
+
+    status = ucs_sys_readdir(ucs_string_buffer_cstr(&dev_path),
+                             uct_tcp_iface_parse_virtual_dev, &dev_path);
+    if (status != UCS_ERR_CANCELED) {
+        ucs_string_buffer_cleanup(&dev_path);
+        return NULL;
+    }
+
+    /* 'path_buffer' size is PATH_MAX */
+    sysfs_path = ucs_topo_resolve_sysfs_path(ucs_string_buffer_cstr(&dev_path),
+                                             path_buffer);
+
+    ucs_string_buffer_cleanup(&dev_path);
+    return sysfs_path;
+}
+
 static ucs_status_t uct_tcp_iface_query(uct_iface_h tl_iface,
                                         uct_iface_attr_t *attr)
 {
@@ -226,8 +270,9 @@ static ucs_status_t uct_tcp_iface_query(uct_iface_h tl_iface,
                              sizeof(uct_tcp_am_hdr_t);
     ucs_status_t status;
     int is_default;
-    char sysfs_path[PATH_MAX];
     double pci_bw, network_bw, calculated_bw;
+    char path_buffer[PATH_MAX];
+    const char *sysfs_path;
 
     uct_base_iface_query(&iface->super, attr);
 
@@ -236,8 +281,7 @@ static ucs_status_t uct_tcp_iface_query(uct_iface_h tl_iface,
         return status;
     }
 
-    ucs_snprintf_safe(sysfs_path, PATH_MAX, "%s/%s/device",
-                      UCT_TCP_IFACE_NETDEV_DIR, iface->if_name);
+    sysfs_path             = uct_tcp_iface_get_sysfs_path(iface->if_name, path_buffer);
     pci_bw                 = ucs_topo_get_pci_bw(iface->if_name, sysfs_path);
     calculated_bw          = ucs_min(pci_bw, network_bw);
 
@@ -843,7 +887,7 @@ ucs_status_t uct_tcp_query_devices(uct_md_h md,
     ucs_status_t status;
     DIR *dir;
     const char *sysfs_path;
-    char path_buffer[PATH_MAX], dev_path[PATH_MAX];
+    char path_buffer[PATH_MAX];
     ucs_sys_device_t sys_dev;
 
     dir = opendir(UCT_TCP_IFACE_NETDEV_DIR);
@@ -900,13 +944,9 @@ ucs_status_t uct_tcp_query_devices(uct_md_h md,
         }
         devices = tmp;
 
-        ucs_snprintf_safe(dev_path, PATH_MAX, "%s/%s", UCT_TCP_IFACE_NETDEV_DIR,
-                          entry->d_name);
-        sysfs_path = uct_iface_get_sysfs_path(dev_path, entry->d_name,
-                                              path_buffer);
-
-        sys_dev = ucs_topo_get_sysfs_dev(entry->d_name, sysfs_path,
-                                         sys_device_priority);
+        sysfs_path = uct_tcp_iface_get_sysfs_path(entry->d_name, path_buffer);
+        sys_dev    = ucs_topo_get_sysfs_dev(entry->d_name, sysfs_path,
+                                            sys_device_priority);
 
         ucs_snprintf_zero(devices[num_devices].name,
                           sizeof(devices[num_devices].name),
--- a/test/gtest/ucp/test_ucp_wireup.cc
+++ b/test/gtest/ucp/test_ucp_wireup.cc
@@ -8,6 +8,7 @@
 
 #include "ucp_test.h"
 #include "common/test.h"
+#include "common/test_helpers.h"
 #include "ucp/ucp_test.h"
 
 #include <algorithm>
@@ -83,16 +84,20 @@ protected:
     bool ep_iface_has_caps(const entity& e, const std::string& tl,
                            uint64_t caps);
 
-    bool
-    has_resource(const ucp_test_base::entity *e, const std::string &tl_name)
+    size_t count_resources(const ucp_test_base::entity &e,
+                           const std::string &tl_name) const
     {
-        for (int i = 0; i < e->ucph()->num_tls; ++i) {
-            if (tl_name == e->ucph()->tl_rscs[i].tl_rsc.tl_name) {
-                return true;
-            }
-        }
+        return std::count_if(e.ucph()->tl_rscs,
+                             e.ucph()->tl_rscs + e.ucph()->num_tls,
+                             [&](const ucp_tl_resource_desc_t &rsc) {
+                                 return tl_name == rsc.tl_rsc.tl_name;
+                             });
+    }
 
-        return false;
+    bool has_resource(const ucp_test_base::entity &e,
+                      const std::string &tl_name) const
+    {
+        return count_resources(e, tl_name) != 0;
     }
 
 protected:
@@ -1322,7 +1327,7 @@ UCS_TEST_P(select_transport_rndv, select_dc, "NUM_EPS=300")
     ucs::scoped_setenv max_num_eps("UCX_RC_MAX_NUM_EPS", "1024");
     ucp_test_base::entity *e1 = create_entity();
 
-    if (!has_resource(e1, "dc_mlx5")) {
+    if (!has_resource(*e1, "dc_mlx5")) {
         UCS_TEST_SKIP_R("dc_mlx5 transport is not present");
     }
 
@@ -1346,23 +1351,27 @@ public:
 
 UCS_TEST_P(select_transport_rma_bw, select_rc, "NUM_EPS=56", "NUM_PPN=28")
 {
-    UCS_TEST_SKIP_R("FIXME: Disabled due to unresolved failure");
+    const auto rc_count = count_resources(sender(), "rc_mlx5");
 
-    /* coverity[unreachable] */
-    if (!has_resource(&sender(), "rc_mlx5")) {
-        UCS_TEST_SKIP_R("no rc resources");
+    if (rc_count == 0) {
+        UCS_TEST_SKIP_R("no RC resources");
     }
 
     sender().connect(&receiver(), get_ep_params());
+    const auto config = ucp_ep_config(sender().ep());
 
-    const ucp_ep_config_t *config = ucp_ep_config(sender().ep());
-    for (int i = 0; (i < config->key.num_lanes) &&
-                    (config->key.rma_bw_lanes[i] != UCP_NULL_LANE);
-         ++i) {
-        /* Verify RC is selected for rma_bw lanes */
-        ucp_lane_index_t lane = config->key.rma_bw_lanes[i];
-        ASSERT_STREQ("rc_mlx5",
-                     ucp_ep_get_tl_rsc(sender().ep(), lane)->tl_name);
+    for (int i = 0; i < config->key.num_lanes; ++i) {
+        const auto lane = config->key.rma_bw_lanes[i];
+        /* Only the first rc_count lanes are relevant, because
+         * no more RC resources exist. */
+        if ((lane == UCP_NULL_LANE) || (i >= rc_count)) {
+            break;
+        }
+
+        /* Verify TCP is not selected for rma_bw lanes */
+        const std::string tl_name =
+                ucp_ep_get_tl_rsc(sender().ep(), lane)->tl_name;
+        ASSERT_TRUE((tl_name == "rc_mlx5") || (tl_name == "dc_mlx5"));
     }
 }
 
--- a/src/uct/base/uct_iface.c
+++ b/src/uct/base/uct_iface.c
@@ -858,56 +858,6 @@ void uct_iface_get_local_address(uct_iface_local_addr_ns_t *addr_ns,
     }
 }
 
-const char *uct_iface_get_sysfs_path(const char *dev_path, const char *dev_name,
-                                     char *path_buffer)
-{
-    const char *detected_type = NULL;
-    char device_file_path[PATH_MAX];
-    char *sysfs_realpath;
-    struct stat st_buf;
-    char *sysfs_path;
-    int ret;
-
-    /* realpath name is expected to be like below:
-     * PF: /sys/devices/.../0000:03:00.0/<interface_type>/<dev_name>
-     * SF: /sys/devices/.../0000:03:00.0/<UUID>/<interface_type>/<dev_name>
-     */
-
-    sysfs_realpath = realpath(dev_path, path_buffer);
-    if (sysfs_realpath == NULL) {
-        goto out_undetected;
-    }
-
-    /* Try PF: strip 2 components */
-    sysfs_path = ucs_dirname(sysfs_realpath, 2);
-    ucs_snprintf_safe(device_file_path, sizeof(device_file_path), "%s/device",
-                      sysfs_path);
-    ret = stat(device_file_path, &st_buf);
-    if (ret == 0) {
-        detected_type = "PF";
-        goto out_detected;
-    }
-
-    /* Try SF: strip 3 components (one more) */
-    sysfs_path = ucs_dirname(sysfs_path, 1);
-    ucs_snprintf_safe(device_file_path, sizeof(device_file_path), "%s/device",
-                      sysfs_path);
-    ret = stat(device_file_path, &st_buf);
-    if (ret == 0) {
-        detected_type = "SF";
-        goto out_detected;
-    }
-
-out_undetected:
-    ucs_debug("%s: sysfs path undetected", dev_name);
-    return NULL;
-
-out_detected:
-    ucs_debug("%s: %s sysfs path is '%s'\n", dev_name, detected_type,
-              sysfs_path);
-    return sysfs_path;
-}
-
 int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
                                  ucs_sys_namespace_type_t sys_ns_type)
 {
--- a/src/uct/base/uct_iface.h
+++ b/src/uct/base/uct_iface.h
@@ -856,19 +856,6 @@ void uct_iface_get_local_address(uct_iface_local_addr_ns_t *addr_ns,
 int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
                                  ucs_sys_namespace_type_t sys_ns_type);
 
-/**
- * Returns sysfs path for the required device.
- *
- * @param [in]  dev_path    Device file path.
- * @param [in]  dev_name    Device Name.
- * @param [in]  path_buffer Allocated buffer to hold the string to be
- *                          returned.
- *
- * @return Pointer to sysfs path or NULL on error.
- */
-const char *uct_iface_get_sysfs_path(const char *dev_path, const char *dev_name,
-                                     char *path_buffer);
-
 /*
  * Invoke active message handler.
  *
--- a/src/uct/ib/base/ib_device.c
+++ b/src/uct/ib/base/ib_device.c
@@ -542,7 +542,7 @@ ucs_status_t uct_ib_device_query(uct_ib_device_t *dev,
         }
     }
 
-    sysfs_path = uct_iface_get_sysfs_path(dev_path, dev_name, path_buffer);
+    sysfs_path   = ucs_topo_resolve_sysfs_path(dev_path, path_buffer);
     dev->sys_dev = ucs_topo_get_sysfs_dev(dev_name, sysfs_path,
                                           sys_device_priority);
     uct_ib_device_set_pci_id(dev, sysfs_path);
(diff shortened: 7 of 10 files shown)


src/uct/tcp/tcp_iface.c
```diff
@@ -221,7 +230,8 @@ static ucs_status_t uct_tcp_iface_query(uct_iface_h tl_iface,
                              sizeof(uct_tcp_am_hdr_t);
     ucs_status_t status;
     int is_default;
-    char sysfs_path[PATH_MAX];
+    char path_buffer[PATH_MAX];
```
gleon99, line 233: Can allocate it in `uct_tcp_iface_get_sysfs_path` instead?

gleon99:
@shasson5 please elaborate in the PR description the specific problem and motivation.

src/uct/tcp/tcp_iface.c
shasson5: you mean with malloc?

src/uct/tcp/tcp_iface.c
```diff
@@ -213,6 +213,15 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
     return 1;
 }
 
+static const char *
+uct_tcp_iface_get_sysfs_path(const char *dev_name, char *path_buffer)
```
yosefe, line 217: pass max size of path buffer

src/uct/tcp/tcp_iface.c
```diff
@@ -213,6 +213,15 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
...
+{
+    char dev_path[PATH_MAX];
+
+    ucs_snprintf_safe(dev_path, PATH_MAX, "%s/%s", UCT_TCP_IFACE_NETDEV_DIR,
+                      dev_name);
+    return uct_iface_get_sysfs_path(dev_path, dev_name, path_buffer);
+}
```
yosefe, line 224: space line after

src/uct/tcp/tcp_iface.c
```diff
@@ -213,6 +213,15 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
     return 1;
 }
 
+static const char *
+uct_tcp_iface_get_sysfs_path(const char *dev_name, char *path_buffer)
+{
+    char dev_path[PATH_MAX];
+
+    ucs_snprintf_safe(dev_path, PATH_MAX, "%s/%s", UCT_TCP_IFACE_NETDEV_DIR,
```
yosefe, line 221: sizeof(dev_path) better than PATH_MAX

yosefe:
> @shasson5 please elaborate in the PR description the specific problem and motivation.

Right - pls give example of how path is different, and how this PR fixes it

src/uct/tcp/tcp_iface.c
yosefe: IMO it's worse to malloc because then malloc and free won't be in the same function

src/uct/tcp/tcp_iface.c
```diff
@@ -231,8 +241,7 @@ static ucs_status_t uct_tcp_iface_query(uct_iface_h tl_iface,
         return status;
     }
 
-    ucs_snprintf_safe(sysfs_path, PATH_MAX, "%s/%s/device",
-                      UCT_TCP_IFACE_NETDEV_DIR, iface->if_name);
+    sysfs_path             = uct_tcp_iface_get_sysfs_path(iface->if_name, path_buffer);
```
yosefe, line 244: so the problem was that need to remove "/device" here?
won't it break anything? since seems this fix is not BF/arm specific

shasson5: Actually this was a bug, because we should not use /device at all.
All the logic for the 2 cases (x86/Arm) is covered in `uct_iface_get_sysfs_path`

yosefe: so it didn't work at all, even on x86?

shasson5: on x86 it worked by chance, because `device` file points exactly to the location of the required files (`current_link_width/speed`)

yosefe: seems that without device it wouldn't work on x86
```
13:35 <yosefe@jazz07:5 launch>$ ll /sys/class/net/ib0/current_link_speed
ls: cannot access /sys/class/net/ib0/current_link_speed: No such file or directory
13:35 <yosefe@jazz07:5 launch>$ ll /sys/class/net/ib0/device/current_link_speed
-r--r--r-- 1 root root 4096 Jan 15 13:35 /sys/class/net/ib0/device/current_link_speed
```

yosefe: please add some unit test that makes sure this process works

shasson5: we have the `select_transport_rma_bw` which failed before the fix and pass after it.
do we need another test?

yosefe: yeah, something that always fails if we cannot read the relevant sysfs file, regardless of lane selection
maybe check that sys_dev in uct_query_resources is not UNKNOWN

shasson5: There's no simple way to do this, because we need to check the value of pci_bw and compare it with something.
The 2 options I see are:
1) store sysfs_path as a new member field on tcp_iface, and then use it to cal pci_bw. (then we can check it's value from test).
2) call `uct_tcp_iface_get_sysfs_path` and `get_pci_bw` from test to get the expected value and compare it with the result of `uct_iface_query`)

yosefe: option 2: wrap uct_tcp_iface_get_sysfs_path+ get_pci_bw in some helper function in TCP transport, and call it explicitly from the test. this function can take tcp iface name and return its pci bw.

src/uct/tcp/tcp_iface.c
```diff
@@ -213,15 +213,27 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
...
 }
 
-static const char *
-uct_tcp_iface_get_sysfs_path(const char *dev_name, char *path_buffer)
+static const char *uct_tcp_iface_get_sysfs_path(const char *dev_name,
+                                                char *path_buffer,
+                                                size_t path_buffer_size)
 {
-    char dev_path[PATH_MAX];
+    char dev_path[path_buffer_size];
```
yosefe, line 220: ok, seems no point to pass path_buffer_size. because uct_iface_get_sysfs_path uses it for realpath() that already assumes its size is PATH_MAX. so maybe just need to document the size of path_buffer is PATH_MAX.

test/gtest/uct/tcp/test_tcp.cc
```diff
@@ -286,5 +286,20 @@ UCS_TEST_P(test_uct_tcp, check_addr_len)
...
 
+UCS_TEST_P(test_uct_tcp, get_pci_bw)
+{
+    double expected_pci_bw;
+    uct_iface_attr_t iface_attr;
+
+    expected_pci_bw = uct_tcp_iface_get_pci_bw(GetParam()->dev_name.c_str());
+    if (expected_pci_bw == DBL_MAX) {
+        UCS_TEST_SKIP_R("Device is not an IB device");
+    }
```
yosefe, line 297: space line after

test/gtest/uct/tcp/test_tcp.cc
```diff
@@ -286,5 +286,20 @@ UCS_TEST_P(test_uct_tcp, check_addr_len)
...
+    ucs_status_t status = uct_iface_query(m_ent->iface(), &iface_attr);
+    ASSERT_UCS_OK(status);
+
```
yosefe, line 300: extra space line

src/uct/tcp/tcp_iface.c
```diff
@@ -213,15 +213,27 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
...
 
-    ucs_snprintf_safe(dev_path, PATH_MAX, "%s/%s", UCT_TCP_IFACE_NETDEV_DIR,
-                      dev_name);
+    ucs_snprintf_safe(dev_path, sizeof(dev_path), "%s/%s",
+                      UCT_TCP_IFACE_NETDEV_DIR, dev_name);
     return uct_iface_get_sysfs_path(dev_path, dev_name, path_buffer);
 }
+
+double uct_tcp_iface_get_pci_bw(const char *dev_name)
```
yosefe, line 227: can we move uct_tcp_iface_get_sysfs_path+uct_tcp_iface_get_pci_bw+uct_iface_get_sysfs_path to ucs/topo?

test/gtest/uct/tcp/test_tcp.cc
```diff
@@ -286,5 +286,20 @@ UCS_TEST_P(test_uct_tcp, check_addr_len)
...
+
+    ASSERT_EQ(iface_attr.bandwidth.shared, expected_pci_bw);
```
yosefe, line 302: the test seems wrong. if we had bug in reading sysfs the test would be skipped (because get_pci_bw returns DBL_MAX) and not fail.
we should probably test this:
- for any iface, (iface_attr.bandwidth.shared+dedicated) <= pci_bw
- for iface that is rdma device (with rc/dc/ud/tcp transports), pci_bw is NOT max - which means we really detected it from sysfs

shasson5: The test is running for tcp transports only. We want to move it to some other place?

yosefe: yes, run it for all transports.

shasson5: how can I identify TCP/rdma device?

src/uct/tcp/tcp_iface.c
shasson5: I think `uct_tcp_iface_get_sysfs_path` better stay in tcp_iface otherwise topo will have too many identical interfaces, and it will be confusing.

tschuett:
Could you put all the sysfs handling into a struct with function pointers? LLVM has the notion of real and virtual filesystems. In production, you would use the real filesystem. In gtest, you can use the real and prepopulated virtual filesystem for testing.

yosefe:
pipeline failure is relevant:
```
2023-01-18T12:57:03.7060918Z [----------] 1 test from all/test_pci_bw
2023-01-18T12:57:03.7061539Z [ RUN      ] all/test_pci_bw.get_pci_bw/0 <all/tag>
2023-01-18T12:57:03.9798445Z /scrap/azure/agent-05/AZP_WORKSPACE/2/s/contrib/../test/gtest/ucp/test_ucp_wireup.cc:1894: Failure
2023-01-18T12:57:03.9799986Z Expected: (pci_bw) != (double(1.79769313486231570815e+308L)), actual: 1.79769e+308 vs 1.79769e+308
2023-01-18T12:57:04.0407435Z [  FAILED  ] all/test_pci_bw.get_pci_bw/0, where GetParam() = all/tag (335 ms)
2023-01-18T12:57:04.0409415Z [----------] 1 test from all/test_pci_bw (335 ms total)
```

yosefe:
[Failure in select_transport_rma_bw - test on Jazz mlx5_2](https://trello.com/c/KmVgxHkd/1494-failure-in-selecttransportrmabw-test-on-jazz-mlx52)

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1344,23 +1353,26 @@ class select_transport_rma_bw : public test_ucp_wireup {
 
 UCS_TEST_P(select_transport_rma_bw, select_rc, "NUM_EPS=56", "NUM_PPN=28")
 {
-    UCS_TEST_SKIP_R("FIXME: Disabled due to unresolved failure");
+    size_t ib_count = count_resource(&sender(), "rc_mlx5");
```
gleon99, line 1356: `const`

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1344,23 +1353,26 @@ class select_transport_rma_bw : public test_ucp_wireup {
...
     sender().connect(&receiver(), get_ep_params());
-
     const ucp_ep_config_t *config = ucp_ep_config(sender().ep());
-    for (int i = 0; (i < config->key.num_lanes) &&
-                    (config->key.rma_bw_lanes[i] != UCP_NULL_LANE);
-         ++i) {
-        /* Verify RC is selected for rma_bw lanes */
+
+    for (int i = 0; i < config->key.num_lanes; ++i) {
+        /* Only the first ib_count lanes are interesting, because
```
gleon99, line 1366: `Only the first ib_count lanes are relevant` /
`Count only the first ib_count lanes`

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1827,3 +1839,89 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+    std::string get_tcp_device(const std::string &dev_name)
+    {
+        std::string dev_path = "/sys/class/net/" + dev_name;
+
+        std::vector<std::string> sysfs_files = ucs::read_dir(dev_path);
+        if (sysfs_files.size() == 0) {
+            return "";
+        }
+
+        for (int i = 0; i < sysfs_files.size(); ++i) {
```
gleon99, line 1859:
```
for (const auto &it : sysfs_files) {
    ....
```

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1827,3 +1839,89 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
 UCP_INSTANTIATE_TEST_CASE(test_ucp_address_v2)
+
+class test_pci_bw : public test_ucp_wireup {
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_TAG, TEST_TAG, "tag");
+    }
+
+    std::string get_tcp_device(const std::string &dev_name)
```
gleon99, line 1850: `auto`?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1827,3 +1839,89 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+                break;
+            }
+        }
+
+        return dev_path;
+    }
+
+    std::string get_dev_path(const std::string &dev_name)
+    {
+        std::string device = get_tcp_device(dev_name);
```
gleon99, line 1871: `const auto`?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1827,3 +1839,89 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+
+        struct stat st;
+        if (stat(ib_dev_path.c_str(), &st) >= 0) {
+            return ib_dev_path;
+        }
+
+        return "";
+    }
+
+    bool is_ib_device(const std::string &dev_name)
```
gleon99, line 1895: `) const`

test/gtest/uct/tcp/test_tcp.cc
```diff
@@ -286,5 +286,4 @@ UCS_TEST_P(test_uct_tcp, check_addr_len)
     EXPECT_EQ(sizeof(uct_tcp_ep_addr_t), iface_attr.ep_addr_len);
 }
 
-
```
gleon99, line 289: Why?

test/gtest/ucp/test_ucp_wireup.cc
shasson5: replace std::string?

test/gtest/ucp/test_ucp_wireup.cc
shasson5: replace std::string?

test/gtest/ucp/test_ucp_wireup.cc
gleon99: If possible, why not?

test/gtest/ucp/test_ucp_wireup.cc
gleon99: See above

test/gtest/ucp/test_ucp_wireup.cc
shasson5: you mean for the function return type? seems it's not simple in c++ 11:
ucp/test_ucp_wireup.cc:1850:52: error: ‘get_tcp_device’ function uses ‘auto’ type specifier without trailing return type [-Werror]

test/gtest/ucp/test_ucp_wireup.cc
gleon99: Unresolved

src/uct/tcp/tcp_iface.c
```diff
@@ -213,32 +219,43 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
...
+{
+    static const char *low_level_dev_prefix = "lower_";
+    uct_tcp_iface_parse_virtual_dev_t *data = ctx;
+    size_t prefix_len;
+
+    prefix_len = ucs_string_common_prefix_len(entry->d_name,
+                                              low_level_dev_prefix);
+
+    if (prefix_len == strlen(low_level_dev_prefix)) {
+        if (!data->found) {
```
gleon99, line 233: Unite in a single `if`

src/uct/tcp/tcp_iface.c
```diff
@@ -213,32 +219,43 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
...
 
-    /* Check RoCE LAG */
-    status = ucs_sys_read_sysfs_file(dev_name, ucs_string_buffer_cstr(dev_path),
-                                     "bonding/slaves", slaves, sizeof(slaves),
-                                     UCS_LOG_LEVEL_DEBUG);
-    if (status != UCS_OK) {
-        /* Not RoCE LAG */
-        return;
-    }
+    data.dev_path = dev_path;
```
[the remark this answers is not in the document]
gleon99: @shasson5  Unresolved

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -83,16 +84,24 @@ class test_ucp_wireup : public ucp_test {
     bool ep_iface_has_caps(const entity& e, const std::string& tl,
                            uint64_t caps);
 
-    bool
-    has_resource(const ucp_test_base::entity *e, const std::string &tl_name)
+    size_t
+    count_resource(const ucp_test_base::entity *e, const std::string &tl_name)
```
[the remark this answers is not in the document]
gleon99: `count_resources`?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1841,89 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
 UCP_INSTANTIATE_TEST_CASE(test_ucp_address_v2)
+
+class test_pci_bw : public test_ucp_wireup {
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_TAG, TEST_TAG, "tag");
+    }
+
+    std::string get_tcp_device(const std::string &dev_name)
```
gleon99, line 1852: `const` method?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1841,89 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+    }
+};
+
+UCS_TEST_P(test_pci_bw, get_pci_bw)
+{
+    ucp_worker_h worker = sender().worker();
+    ucp_context_h ctx   = worker->context;
+    char path_buffer[PATH_MAX];
+
+    for (unsigned i = 0; i < worker->num_ifaces; ++i) {
```
gleon99, line 1909: `auto i`

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1841,89 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+class test_pci_bw : public test_ucp_wireup {
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_TAG, TEST_TAG, "tag");
+    }
+
+    std::string get_tcp_device(const std::string &dev_name)
+    {
+        std::string dev_path = "/sys/class/net/" + dev_name;
```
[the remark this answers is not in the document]
shasson5: can't, it's appended with another string later in this function.

src/uct/tcp/tcp_iface.c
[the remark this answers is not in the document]
shasson5: moved it to start of function. Isn't that what you meant?

test/gtest/ucp/test_ucp_wireup.cc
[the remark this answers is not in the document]
shasson5: not sure I understand, it is already changed.

src/uct/tcp/tcp_iface.c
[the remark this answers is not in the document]
gleon99: Why not `uct_tcp_iface_virtual_dev_path_t data = {...` ?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1842,93 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+};
+
+UCS_TEST_P(test_pci_bw, get_pci_bw)
+{
+    ucp_worker_h worker = sender().worker();
+    ucp_context_h ctx   = worker->context;
+    char path_buffer[PATH_MAX];
+
+    for (auto i = 0; i < worker->num_ifaces; ++i) {
+        ucp_worker_iface_t *wiface = worker->ifaces[i];
```
gleon99, line 1915: Align `=`

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1842,93 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
 }
 
 UCP_INSTANTIATE_TEST_CASE(test_ucp_address_v2)
+
+class test_pci_bw : public test_ucp_wireup {
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
```
gleon99, line 1848: Please apply `const` and/or `auto` wherever possible.

src/ucs/sys/topo/base/topo.c
```diff
@@ -625,3 +625,54 @@ double ucs_topo_get_pci_bw(const char *dev_name, const char *sysfs_path)
...
+    int ret;
+
+    /* realpath name is expected to be like below:
+     * PF: /sys/devices/.../0000:03:00.0/<interface_type>/<dev_name>
+     * SF: /sys/devices/.../0000:03:00.0/<UUID>/<interface_type>/<dev_name>
+     */
+
+    sysfs_realpath = realpath(dev_path, path_buffer);
+    if (sysfs_realpath == NULL) {
+        goto out_undetected;
```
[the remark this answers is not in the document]
shasson5: was part of the old code:
https://github.com/openucx/ucx/blob/a932dab947b255c8ac2d46f2f258f7b94fe6affc/src/ucs/sys/topo/base/topo.c#L203

src/uct/tcp/tcp_iface.c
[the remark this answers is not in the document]
shasson5: https://github.com/openucx/ucx/blob/66613c1ff9afa08d7777c04356d72f57e0e9ff35/src/uct/tcp/tcp_iface.c#L231

test/gtest/ucp/test_ucp_wireup.cc
shasson5: this function is inherited from a base class:
```
class test_ucp_context : public ucp_test {
public:
    static void get_test_variants(std::vector<ucp_test_variant> &variants);
}
```

src/ucs/sys/topo/base/topo.c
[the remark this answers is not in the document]
gleon99: Nevertheless, maybe remove it? To simplify.

src/uct/tcp/tcp_iface.c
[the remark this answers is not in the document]
gleon99: @shasson5 Unresolved

gleon99:
@yosefe please take a look.

src/uct/tcp/tcp_iface.c
gleon99: @shasson5  please check whether this one is still relevant.

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1365,14 +1365,14 @@ UCS_TEST_P(select_transport_rma_bw, select_rc, "NUM_EPS=56", "NUM_PPN=28")
     const auto config = ucp_ep_config(sender().ep());
 
     for (int i = 0; i < config->key.num_lanes; ++i) {
-        /* Only the first ib_count lanes are relevant, because
-         * no more IB resources exist. */
-        if ((config->key.rma_bw_lanes[i] == UCP_NULL_LANE) || (i >= rc_count)) {
+        const auto lane           = config->key.rma_bw_lanes[i];
```
gleon99, line 1368: Excessive spacing?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1842,98 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+    {
+        add_variant_with_value(variants, UCP_FEATURE_TAG, TEST_TAG, "tag");
+    }
+
+    std::string get_tcp_device(const std::string &dev_name) const
+    {
+        auto dev_path = "/sys/class/net/" + dev_name;
+
+        std::vector<std::string> sysfs_files = ucs::read_dir(dev_path);
+        if (sysfs_files.size() == 0) {
```
gleon99, line 1858: `.empty()`

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1842,98 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_TAG, TEST_TAG, "tag");
+    }
+
+    std::string get_tcp_device(const std::string &dev_name) const
+    {
+        auto dev_path = "/sys/class/net/" + dev_name;
+
+        std::vector<std::string> sysfs_files = ucs::read_dir(dev_path);
```
gleon99, line 1857: `auto`?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1842,98 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+            }
+        }
+
+        return dev_path;
+    }
+
+    std::string get_dev_path(const std::string &dev_name)
+    {
+        const auto device = get_tcp_device(dev_name);
+        if (!device.empty()) {
```
gleon99, line 1875: Maybe use `?:`

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1842,98 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+            return device;
+        }
+
+        return get_ib_device(dev_name);
+    }
+
+    std::string get_ib_device(const std::string &dev_name) const
+    {
+        /* Length of IB device name suffix (eg. ':1') */
+        static const size_t suffix_length = 2;
```
gleon99, line 1885: `constexpr`

src/ucs/sys/topo/base/topo.c
```diff
@@ -625,3 +625,49 @@ double ucs_topo_get_pci_bw(const char *dev_name, const char *sysfs_path)
     ucs_debug("%s: pci bandwidth undetected, using maximal value", dev_name);
     return DBL_MAX;
 }
+
+const char *ucs_topo_dev_path_to_sysfs_path(const char *dev_path,
+                                            const char *dev_name,
```
yosefe, line 630: seems dev_name is used only for debug purpose. can we avoid passing it? print the dev_path in case of failure

src/ucs/sys/topo/base/topo.c
```diff
@@ -625,3 +625,49 @@ double ucs_topo_get_pci_bw(const char *dev_name, const char *sysfs_path)
     ucs_debug("%s: pci bandwidth undetected, using maximal value", dev_name);
     return DBL_MAX;
 }
+
+const char *ucs_topo_dev_path_to_sysfs_path(const char *dev_path,
```
yosefe, line 629: maybe
ucs_topo_resolve_sysfs_path

src/ucs/sys/topo/base/topo.c
```diff
@@ -625,3 +625,49 @@ double ucs_topo_get_pci_bw(const char *dev_name, const char *sysfs_path)
...
+                                            const char *dev_name,
+                                            char *path_buffer)
+{
+    static const char *path_detected_log_format   = "%s: %s sysfs path is '%s'\n";
+    static const char *path_undetected_log_format = "%s: sysfs path undetected";
```
yosefe, line 634: IMO better unite debug prints by "goto" instead of reusing same format string

src/ucs/sys/topo/base/topo.h
```diff
@@ -202,6 +202,19 @@ ucs_status_t ucs_topo_sys_device_set_name(ucs_sys_device_t sys_dev,
  */
 double ucs_topo_get_pci_bw(const char *dev_name, const char *sysfs_path);
 
+/**
+ * Returns sysfs path of a given device.
```
yosefe, line 206: can you give example of input and output of this function? add it to the doc

src/uct/tcp/tcp_iface.c
```diff
@@ -218,6 +224,55 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
...
 }
 
+static ucs_status_t
+uct_tcp_iface_parse_virtual_dev(struct dirent *entry, void *ctx)
+{
+    static const char *low_level_dev_prefix = "lower_";
+    uct_tcp_iface_virtual_dev_path_t *data  = ctx;
+    size_t prefix_len;
+
+    prefix_len = ucs_string_common_prefix_len(entry->d_name,
```
yosefe, line 234: use strncmp instead?

src/uct/tcp/tcp_iface.c
```diff
@@ -218,6 +224,55 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
...
+                                              low_level_dev_prefix);
+
+    if ((prefix_len == strlen(low_level_dev_prefix)) && !data->found) {
+        ucs_string_buffer_appendf(data->dev_path, "/%s", entry->d_name);
+        data->found = 1;
```
yosefe, line 239: return UCS_ERR_CANCELED to stop iteration

src/uct/tcp/tcp_iface.c
```diff
@@ -218,6 +224,55 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
...
+    }
+
+    return UCS_OK;
+}
+
+static const char *
+uct_tcp_iface_get_sysfs_path(const char *dev_name, char *path_buffer)
+{
+    ucs_string_buffer_t dev_path          = UCS_STRING_BUFFER_INITIALIZER;
+    uct_tcp_iface_virtual_dev_path_t data = {0, &dev_path};
```
yosefe, line 249: uct_tcp_iface_virtual_dev_ctx_t ctx = ...

src/uct/tcp/tcp_iface.c
```diff
@@ -218,6 +224,55 @@ static int uct_tcp_iface_is_reachable(const uct_iface_h tl_iface,
...
+        data->found = 1;
+    }
+
+    return UCS_OK;
+}
+
+static const char *
+uct_tcp_iface_get_sysfs_path(const char *dev_name, char *path_buffer)
+{
+    ucs_string_buffer_t dev_path          = UCS_STRING_BUFFER_INITIALIZER;
```
yosefe, line 248: can use UCS_STRING_BUFFER_ONSTACK with PATH_MAX length

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -83,16 +84,24 @@ class test_ucp_wireup : public ucp_test {
...
+                             [&](const ucp_tl_resource_desc_t &rsc) {
+                                 return tl_name == rsc.tl_rsc.tl_name;
+                             });
+    }
 
-        return false;
+    bool has_resource(const ucp_test_base::entity &e,
+                      const std::string &tl_name) const
+    {
+        return std::any_of(e.ucph()->tl_rscs,
```
yosefe, line 100: maybe we can do count_resources(e, tl_name) != 0?

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1842,88 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
...
+        const char *sysfs_path = ucs_topo_dev_path_to_sysfs_path(
+                dev_path.c_str(), dev_name, path_buffer);
+        const double pci_bw    = ucs_topo_get_pci_bw(dev_name, sysfs_path);
+
+        uct_iface_attr_t attr;
+        ASSERT_UCS_OK(uct_iface_query(wiface->iface, &attr));
+        ASSERT_LE(attr.bandwidth.shared + attr.bandwidth.dedicated, pci_bw);
+
+        if (ucs::is_rdmacm_netdev(dev_name) || is_ib_device(dev_name)) {
+            ASSERT_NE(pci_bw, std::numeric_limits<double>::max());
```
yosefe, line 1924: `ASSERT_LT(pci_bw, 1e15)`

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -1829,3 +1842,88 @@ UCS_TEST_SKIP_COND_P(test_ucp_address_v2, diff_seg_sizes,
 }
 
 UCP_INSTANTIATE_TEST_CASE(test_ucp_address_v2)
+
+class test_pci_bw : public test_ucp_wireup {
```
yosefe, line 1846: maybe this test can be in test_ucp_context.cc or test_ucp_worker.cc?
sees it does not need endpoints and not related to wireup

src/uct/tcp/tcp_iface.c
shasson5: was removed after refactoring

src/uct/tcp/tcp_iface.c
shasson5: too big to fit in stack:
string_buff_elem * MAX_PATH > UCS_ALLOCA_MAX_SIZE (24 * 256 > 1200)

src/uct/tcp/tcp_iface.c
```diff
@@ -258,16 +248,15 @@ uct_tcp_iface_get_sysfs_path(const char *dev_name, char *path_buffer)
                               dev_name);
 
     status = ucs_sys_readdir(ucs_string_buffer_cstr(&dev_path),
-                             uct_tcp_iface_parse_virtual_dev, &data);
-    if (status != UCS_OK) {
+                             uct_tcp_iface_parse_virtual_dev, &dev_path);
+    if ((status != UCS_OK) && (status != UCS_ERR_CANCELED)) {
```
yosefe, line 252: IMO it should be `if ((status != UCS_ERR_CANCELED)`
status==UCS_OK means we did not find a corresponding entry
otherwise, need to check if dev_path is an empty string

yosefe (approved):

gleon99:
@shasson5  please squash.

yosefe (approved):

PR #8816, opened 2023-01-13
UCT/CUDA: Pass page-aligned address and length to dmabuf query

yosefe:
## Why
Fix dmabuf not being used when the buffer is not page-aligned (its size or address)

cc @Akshay-Venkatesh


yosefe:
@Akshay-Venkatesh can you pls review?

Akshay-Venkatesh (approved):
@yosefe I tested IMB, OMB, and perftest. Without this PR IMB/perftest were failing. All tests are passing now and changes look good. Thanks for the fixes.

PR #8822, opened 2023-01-18
UCM/TEST: Fix mremap() override - pass new_address

### LINKED ISSUE #8820: Incorrect `mremap` signature in ucm
Linking to `libucm` breaks `mremap` with `MREMAP_FIXED` because `libucm` hooks it with an incorrect signature.

The [signature in libucm](https://github.com/openucx/ucx/blob/5e8621f95002cf2ad7135987c2a7dc32d4fc72fb/src/ucm/mmap/mmap.h#L29) is:

```C
void *ucm_override_mremap(void *old_address, size_t old_size, size_t new_size, int flags);
```

but since glibc 2.4 (March 2006), the correct signature is:

```C
void *             mremap(void *old_address, size_t old_size, size_t new_size, int flags, void *new_address);
```

(the fifth argument is sort of optional, because it is only used with `MREMAP_FIXED`).

This means that, when the code is trying to call `mremap`, the call will get intercepted by libucm’s hook, it will lose the fifth argument, and then call the correct five-argument `mremap` from glibc passing garbage as the fifth argument. This probably does not have any visible effect unless `MREMAP_FIXED` was set in `flags`, but if it was, everything will break.

yosefe:
## Why
Fix #8820

### DIFF
--- a/src/ucm/api/ucm.h
+++ b/src/ucm/api/ucm.h
@@ -120,6 +120,7 @@ typedef union ucm_event {
         size_t             old_size;
         size_t             new_size;
         int                flags;
+        void               *new_address;
     } mremap;
 
     /*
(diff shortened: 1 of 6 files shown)


brminich (dismissed):

yosefe:
@brminich can u pls reapprove?

brminich (dismissed):

PR #8825, opened 2023-01-19
UCT/IB/DC: support bigger DCI_NUM (>127)

jeynmann:
## What
Support bigger DCI_NUM (> 127).
Add checks/asserts of DCI number to avoid invalid environment variable values.

## Why ?
Cannot establish 4k connections with DCI_NUM set to 128.
Because the DCI index is stored in `int8_t` and the DCI num is stored in `uint8_t`,
the UCX doesn't support more than 127 DCIs each interface or 255 DCIs total.

## How ?
Change the data type of the DCI index/num.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5_ep.h
+++ b/src/uct/ib/dc/dc_mlx5_ep.h
@@ -13,7 +13,8 @@
 
 #include "dc_mlx5.h"
 
-#define UCT_DC_MLX5_EP_NO_DCI ((uint8_t)-1)
+#define UCT_DC_MLX5_EP_NO_DCI ((uint16_t)-1)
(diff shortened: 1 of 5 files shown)


src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -70,7 +71,7 @@ typedef struct uct_dc_mlx5_base_av {
 struct uct_dc_mlx5_ep {
     uct_base_ep_t         super;
     ucs_arbiter_group_t   arb_group;
-    uint8_t               dci;
+    uint16_t              dci;
```
yosefe, line 74: we do not want to increase DC endpoint size

PR #8829, opened 2023-01-20
JUCX: Fix leakage of jucx_request global references

fruhland:
## What
JUCX allocates a global reference to each request, which is used as `user_data` for the request parameters. These global references are deleted inside the native callback. However, when requests are finished directly, the references are not deleted, causing a memory leak. This PR fixes the leakage of global references.

## Why ?
I ran a test application, which sends/receives small tagged messages in a ping-pong pattern using 12 connections (1 thread per connection) and monitored the allocated memory and GC activity:

leak

After 1:25 minutes, the JVM heap is around 10 GB large and grows further, while the GC activity spikes well over 30%, negatively impacting performance.

With the fix applied, the heap size is stable at around 2 GB and the GC activity stays under 1%:

fixed

## How ?
By deleting the global reference, when a request has finished directly. To have access to the reference inside `process_request`, I pass the global reference instead of `jucx_request`.

### DIFF
--- a/bindings/java/src/main/native/endpoint.cc
+++ b/bindings/java/src/main/native/endpoint.cc
@@ -129,7 +129,7 @@ Java_org_openucx_jucx_ucp_UcpEndpoint_closeNonBlockingNative(JNIEnv *env, jclass
     param.cb.send       = jucx_request_callback;
 
     ucs_status_ptr_t status = ucp_ep_close_nbx((ucp_ep_h)ep_ptr, &param);
-    process_request(env, jucx_request, status);
+    process_request(env, &param, status);
 
     return jucx_request;
 }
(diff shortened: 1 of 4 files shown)


bindings/java/src/main/native/endpoint.cc
```diff
@@ -129,7 +129,7 @@ Java_org_openucx_jucx_ucp_UcpEndpoint_closeNonBlockingNative(JNIEnv *env, jclass
     param.cb.send       = jucx_request_callback;
 
     ucs_status_ptr_t status = ucp_ep_close_nbx((ucp_ep_h)ep_ptr, &param);
-    process_request(env, jucx_request, status);
+    process_request(env, reinterpret_cast<jobject>(param.user_data), status);
```
yosefe, line 132: maybe pass `const ucp_request_param_t *param` to `process_request()`?
to make code more readable and ensure such leak would not be added any more

fruhland: Sounds good to me. I wasn't sure what is the best way to pass the reference to `process_request()`, but thinking of it again, avoiding the typecast seems better than my current solution.

yosefe (approved):

yosefe:
@petro-rudenko can you pls review as well?

petro-rudenko (approved):

PR #8835, opened 2023-01-25
UCP/PROTO: add err handling test on initialization

evgeny-leksikov:
## What
add err handling test on initialization

## Why ?
to disqualify protocols without abort implementation if error handling is requested


src/ucp/proto/proto.h
```diff
@@ -176,6 +176,7 @@ typedef struct {
                                                         protocol, for debugging */
 
     /* Output parameters */
+    const ucp_proto_t              *proto;      /* Pointer to protocol itself */
```
yosefe, line 179: can we avoid adding this?

evgeny-leksikov: there are 2 options:
- change init() semantic adding this parameter
- calculate the result as a constant manually for every protocol, but I don't like this approach because of human factor

yosefe: 2nd option means we may have to check errh requirement in every protocol, but IMO some protocols checks could be united (such as rma/amo/tag offload), and also some rndv protocols have to implement err handling anyway. So maybe it won't be in so many places in the code.

evgeny-leksikov: to make it in runtime, we need to a have some parameter with access to `abort` function pointer, otherwise the initialization path must be checked for every protocol manually, IMO it's a bug-prone way

yosefe: not sure that presence of "abort" is right indication of errh support.
maybe the protocol is such that abort is not even needed in order to handle errors?

evgeny-leksikov: then it would be `empty_func`, anyway we call `proto.abort` on error flow in many paces

src/ucp/proto/proto_multi.c
```diff
@@ -41,6 +41,10 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
     ucs_assert(params->max_lanes >= 1);
     ucs_assert(params->max_lanes <= UCP_PROTO_MAX_LANES);
 
+    if (!ucp_proto_init_check_err_handling(&params->super.super)) {
```
yosefe, line 44: IMO we should pass err handling support as part of https://github.com/openucx/ucx/blob/d7a1b4527dd05e37a1f4630ed99142c8c41eaa93/src/ucp/proto/proto_common.h#L113

instead of checking by function pointer.
we can assert by function pointer if needed

src/ucp/proto/proto_common.c
```diff
@@ -15,6 +15,15 @@
 #include <uct/api/v2/uct_v2.h>
 
 
+int
+ucp_proto_common_init_check_err_handling(
```
yosefe, line 19: can you pls document this function return value?

src/ucp/proto/proto_common.h
```diff
@@ -176,6 +179,18 @@ typedef void (*ucp_proto_init_cb_t)(ucp_request_t *req);
...
+/**
+ * Check if protocol can be used according to error handling requirements.
+ *
+ * @param [in] init_params      Protocol initialization parameters.
+ *
+ * @return Nonzero if protocol can be used.
+ */
+int
+ucp_proto_common_init_check_err_handling(
+        const ucp_proto_common_init_params_t *init_params);
```
yosefe, line 191: format looks weird, did it pass clang-format?

src/ucp/proto/proto_select.c
```diff
@@ -268,6 +267,15 @@ ucp_proto_select_init_protocols(ucp_worker_h worker,
             continue;
         }
 
+        if (init_params.ep_config_key->err_mode != UCP_ERR_HANDLING_MODE_NONE) {
+            ucs_assertv(ucp_protocols[proto_id]->abort !=
+                        ucp_proto_abort_fatal_not_implemented,
+                        "proto %s was selected for err handling mode %d with "
+                        "missed abort implementation",
```
yosefe, line 274: what if abort() is not really needed? maybe the protocol may never adds anything to pending queue

src/ucp/tag/eager_multi.c
```diff
@@ -55,7 +55,7 @@ ucp_proto_eager_multi_init_common(ucp_proto_multi_init_params_t *params,
 
 static ucs_status_t ucp_proto_eager_bcopy_multi_common_init(
         const ucp_proto_init_params_t *init_params, ucp_proto_id_t op_id,
-        size_t hdr_size)
+        size_t hdr_size, unsigned super_flags)
```
yosefe, line 58: super_flags seems unused

src/ucp/proto/proto_select.c
evgeny-leksikov: I think it should be initialized with empty function stub

yosefe: ok
so pls change it to
"selected protocol %s doesn't implement abort() function, but err handling mode is %d"

yosefe (dismissed):

yosefe (approved):

PR #8845, opened 2023-02-02
RELEASE: Update NEWS file for v1.14-rc2

brminich:
## What
Updating NEWS to release ucx v1.14-rc2

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,11 @@
 ### Features:
 ### Bugfixes:
 
+## 1.14.0 RC2 (February 2, 2023)
+### Bugfixes
+#### GPU (CUDA, ROCM)
+* Updated cuda_copy transport to use event fd instead of async callback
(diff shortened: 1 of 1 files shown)


brminich:
@yosefe

NEWS
```diff
@@ -11,6 +11,11 @@
 ### Features:
 ### Bugfixes:
 
+## 1.14.0 RC2 (February 2, 2023)
```
yosefe, line 14: RC2 should replace RC1 in the NEWS, and also add this to the full list of bugfixes

brminich: but we made separate sections with just relevant updates for v.12-rc* and v13-rc* series.
E.g. [here](https://github.com/openucx/ucx/releases/tag/v1.12.0-rc2)

yosefe (approved):

Issue #8854, opened 2023-02-07
UCX 1.14.0-rc1 and rc2 packages for ubuntu are empty


abellina:
I am trying to test UCX 1.14.0-rc2 for Ubuntu 22.04 and others, but I am finding that the .deb packages are single digit KB and don't include the binaries. So it seems since 1.14 the packaging has some issues for Ubuntu. 

I am looking at the packages published here: https://github.com/openucx/ucx/releases

abellina:
@Al3xR01 I see that the .deb packages are now published in ucx 1.14.0-rc3, but upon installing them I get they were built using `--without-cuda`, which is unexpected:

```
$ ucx_info -v
# Library version: 1.14.0
# Library path: /usr/lib/libucs.so.0
# API headers version: 1.14.0
# Git branch '', revision 6910a39
# Configured with: --disable-logging --disable-debug --disable-assertions --disable-params-check --enable-mt --prefix=/usr --enable-examples --with-java=no --without-cuda
```

abellina:
@Al3xR01 @yosefe I tried rc4 with the fixed cuda support, but I can't find the CUDA libraries in the .deb files:

```
$ dpkg -c ucx-v1.14.0-rc4-ubuntu18.04-mofed5-cuda11.deb|grep cuda|wc -l
0
```

However, it does look like UCX was built with CUDA support, just that the CUDA so's didn't get distributed:

```
$ ucx_info -v
# Library version: 1.14.0
# Library path: /usr/lib/libucs.so.0
# API headers version: 1.14.0
# Git branch '', revision 3ee6d20
# Configured with: --disable-logging --disable-debug --disable-assertions --disable-params-check --enable-mt --prefix=/usr --enable-examples --with-java=no --with-cuda=/usr/local/cuda
```

yosefe:
@Al3xR01 need to pack add DEB packages into the artifact:
```console
+ find . -name '*.deb'
./ucx-gdrcopy_1.14.3ee6d20_amd64.deb
./ucx-cuda_1.14.3ee6d20_amd64.deb
./ucx_1.14.3ee6d20_amd64.deb
```

@abellina since v1.14.0, the cuda packages will be a separate DEB file.

abellina:
ok that makes sense @yosefe thanks, will wait for those artifacts.

abellina:
@yosefe @Al3xR01 then I assume each debian artifact becomes a tarball, not unlike the centos ones?

[...]

PR #8855, opened 2023-02-08
AZP/RELEASE: Build in-tree, move back to the working directory

Al3xR01:
## What
Switch to in-tree Deb build.

## Why ?
https://github.com/openucx/ucx/issues/8854

### DIFF
--- a/buildlib/az-distro-release.yml
+++ b/buildlib/az-distro-release.yml
@@ -70,9 +70,10 @@ jobs:
           cd $(tar tf ${tarball} | head -1) # go to extracted tarball directory
           echo 10 > debian/compat   # https://www.debian.org/doc/manuals/maint-guide/dother.en.htmdpl#compat
           dpkg-buildpackage -us -uc
-          find .. -name '*.deb'
-          find .. -name '*.deb' -exec cp {} "../${AZ_ARTIFACT_NAME}" \;
-          dpkg-deb -I "../${AZ_ARTIFACT_NAME}"
+          cd ..                             # Move back to the working directory
+          find . -name '*.deb'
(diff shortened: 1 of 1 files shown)


yosefe (approved):

buildlib/az-distro-release.yml
```diff
@@ -67,9 +62,13 @@ jobs:
 
       - bash: |
           set -eEx
-          cd pkg-build
-          echo 10 > debian/compat   # https://www.debian.org/doc/manuals/maint-guide/dother.en.html#compat
-          sed -i 's|./contrib|../contrib|' debian/rules  # Hack
+          ./autogen.sh
+          ./contrib/configure-release --with-cuda --with-java=no
```
brminich, line 66: why is there no need for sed hack with contrib path anymore?

yosefe: because the build is no longer out-of-tree on ubuntu

yosefe (dismissed):

buildlib/az-distro-release.yml
```diff
@@ -47,14 +47,9 @@ jobs:
       - bash: |
           set -eEx
           ./autogen.sh
-          mkdir pkg-build
+          mkdir -p pkg-build
```
artemry-nv, line 50: Seems -p is redundant here

artemry-nv (approved):

buildlib/az-distro-release.yml
Al3xR01: Might be superficial here, but does not harm.
I've seen at least one CI failure, where this command was executed before the parent folder was created.

yosefe (approved):

PR #8858, opened 2023-02-09
UCP: ODPv2 control configuration

ivankochin:
## What
Add the user control that allow to force ODP registration for certain memory types.

## Why ?
At this moment it seems reasonable to force memory registration through ODP for certain memory types on some systems.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -444,6 +444,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "directory.",
    ucs_offsetof(ucp_context_config_t, proto_info_dir), UCS_CONFIG_TYPE_STRING},
 
+  {"REG_NONBLOCK_MEM_TYPES", "",
+   "Perform only non-blocking memory registration for these memory types:\n"
+   "page registration may be deferred until it is accessed by the CPU or a transport.",
+   ucs_offsetof(ucp_context_config_t, reg_nb_mem_types),
+   UCS_CONFIG_TYPE_BITMAP(ucs_memory_type_names)},
+
   {NULL}
 };
 
@@ -1507,6 +1513,11 @@ ucp_add_component_resources(ucp_context_h context, ucp_rsc_index_t cmpt_index,
             }
 
             ucs_memory_type_for_each(mem_type) {
+                if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
+                    !(md_attr->reg_nonblock_mem_types & UCS_BIT(mem_type))) {
+                    continue;
+                }
+
                 if (md_attr->flags & UCT_MD_FLAG_REG) {
                     if (md_attr->reg_mem_types & UCS_BIT(mem_type)) {
                         context->reg_md_map[mem_type] |= UCS_BIT(md_index);
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -162,6 +162,8 @@ typedef struct ucp_context_config {
     char                                   *select_distance_md;
     /** Directory to write protocol selection information */
     char                                   *proto_info_dir;
+    /** Memory types that perform non-blocking registration by default */
+    uint64_t                               reg_nb_mem_types;
 } ucp_context_config_t;
 
 
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -393,6 +393,10 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
     err_level = (uct_flags & UCT_MD_MEM_FLAG_HIDE_ERRORS) ? UCS_LOG_LEVEL_DIAG :
                                                             UCS_LOG_LEVEL_ERROR;
 
+    if (context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) {
+        uct_flags |= UCT_MD_MEM_FLAG_NONBLOCK;
+    }
+
     reg_params.flags         = uct_flags;
     reg_params.dmabuf_fd     = UCT_DMABUF_FD_INVALID;
     reg_params.dmabuf_offset = 0;
@@ -441,10 +445,12 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
             goto out_close_dmabuf_fd;
         }
 
-        ucs_trace("register address %p length %zu dmabuf-fd %d on md[%d]=%s %p",
+        ucs_trace("register address %p length %zu dmabuf-fd %d flags %ld "
+                  "on md[%d]=%s %p",
                   address, length,
                   (dmabuf_md_map & UCS_BIT(md_index)) ? reg_params.dmabuf_fd :
                                                         UCT_DMABUF_FD_INVALID,
+                  reg_params.flags,
                   md_index, context->tl_mds[md_index].rsc.md_name,
                   memh->uct[md_index]);
         md_map_registered |= UCS_BIT(md_index);
--- a/src/tools/info/tl_info.c
+++ b/src/tools/info/tl_info.c
@@ -494,6 +494,7 @@ static void print_md_info(uct_component_h component,
 
             PRINT_MD_MEM_TYPE(&strb, md_attr, mem_type, access);
             PRINT_MD_MEM_TYPE(&strb, md_attr, mem_type, alloc);
+            PRINT_MD_MEM_TYPE(&strb, md_attr, mem_type, reg_nonblock);
             PRINT_MD_MEM_TYPE(&strb, md_attr, mem_type, reg);
             PRINT_MD_MEM_TYPE(&strb, md_attr, mem_type, cache);
             PRINT_MD_MEM_TYPE(&strb, md_attr, mem_type, detect);


src/ucp/core/ucp_context.c
```diff
@@ -2044,6 +2060,19 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
             ((context->config.ext.fence_mode == UCP_FENCE_MODE_AUTO) &&
              (context->config.ext.max_rma_lanes > 1));
 
+    if (context->config.ext.odp_reg_mem_types & UCS_BIT(UCP_ODP_MEM_REG_TYPES_AUTO)) {
+        if (context->config.ext.odp_reg_mem_types != UCS_BIT(UCP_ODP_MEM_REG_TYPES_AUTO)) {
+            ucs_error("If UCP_ODP_MEM_REG_TYPES contains 'auto' then it "
```
rakhmets, line 2065: If this is an error, then the function should return `UCS_INVALID_PARAM`.

src/ucp/core/ucp_context.c
```diff
@@ -2044,6 +2060,19 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
...
+                      "shouldn't contain other memory types");
+        }
+    } else {
```
rakhmets, line 2068: Do we allow an empty configuration parameter to be used?

src/ucp/core/ucp_context.c
```diff
@@ -138,6 +138,13 @@ const size_t ucp_context_est_bcopy_bw[UCS_CPU_VENDOR_LAST] = {
     [UCS_CPU_VENDOR_ZHAOXIN]     = UCP_CPU_EST_BCOPY_BW_DEFAULT
 };
 
+static const char *ucp_odp_reg_mem_types[] = {
```
gleon99, line 141: Following consistency with other similar places, it's `.._type_names`.

src/ucp/core/ucp_context.c
```diff
@@ -444,6 +451,14 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "directory.",
    ucs_offsetof(ucp_context_config_t, proto_info_dir), UCS_CONFIG_TYPE_STRING},
 
+  {"ODP_MEM_REG_TYPES", "auto",
+   "Memory types that are registered as ODP regions by default"
```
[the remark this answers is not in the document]
brminich: lets use  `NONBLOCK` (or `NB` in case the name is too long) instead

src/ucp/core/ucp_mm.c
```diff
@@ -576,6 +576,16 @@ ucp_memh_init_from_parent(ucp_mem_h memh, ucp_md_map_t parent_md_map)
     }
 }
 
+static void
+ucp_memh_check_odp_memory_types(ucp_context_h context,
```
gleon99, line 580: Maybe rename it?
I'd expect a `check` function to be a boolean predicate or to `assert()` something, rather than apply config.

src/ucp/core/ucp_context.c
```diff
@@ -444,6 +451,14 @@ static ucs_config_field_t ucp_context_config_table[] = {
...
+   "Either 'auto' for default settings or a combination of:\n"
+    " - 'host'        : Default system memory\n"
```
gleon99, line 457: Excessive spacing?

src/ucp/core/ucp_context.c
```diff
@@ -138,6 +138,13 @@ const size_t ucp_context_est_bcopy_bw[UCS_CPU_VENDOR_LAST] = {
     [UCS_CPU_VENDOR_ZHAOXIN]     = UCP_CPU_EST_BCOPY_BW_DEFAULT
 };
 
+static const char *ucp_odp_reg_mem_types[] = {
+    [UCP_ODP_MEM_REG_TYPES_AUTO] = "auto",
+    [UCP_ODP_MEM_REG_TYPES_HOST] = "host",
+    [UCP_ODP_MEM_REG_TYPES_CUDA] = "cuda",
+    [UCP_ODP_MEM_REG_TYPES_LAST] = NULL
+};
+
```
brminich, line 147: imo better add `auto` handling to `ucs_config_sscanf_bitmap`

src/ucp/core/ucp_context.c
```diff
@@ -2044,6 +2060,19 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
             ((context->config.ext.fence_mode == UCP_FENCE_MODE_AUTO) &&
              (context->config.ext.max_rma_lanes > 1));
 
+    if (context->config.ext.odp_reg_mem_types & UCS_BIT(UCP_ODP_MEM_REG_TYPES_AUTO)) {
+        if (context->config.ext.odp_reg_mem_types != UCS_BIT(UCP_ODP_MEM_REG_TYPES_AUTO)) {
```
brminich, line 2064: this should be handled on parser level

src/ucp/core/ucp_context.h
```diff
@@ -162,6 +162,8 @@ typedef struct ucp_context_config {
     char                                   *select_distance_md;
     /** Directory to write protocol selection information */
     char                                   *proto_info_dir;
+    /** Memory types that are registered using ODP by default */
```
brminich, line 165: pls use `nonblock registration` terminology instead of `odp`

src/ucp/core/ucp_context.c
ivankochin: As far as I know, yes, we do. But this API isn't finalized yet, so maybe it will be changed in the future.

src/ucp/core/ucp_context.c
```diff
@@ -138,6 +138,13 @@ const size_t ucp_context_est_bcopy_bw[UCS_CPU_VENDOR_LAST] = {
     [UCS_CPU_VENDOR_ZHAOXIN]     = UCP_CPU_EST_BCOPY_BW_DEFAULT
 };
 
+static const char *ucp_reg_nb_mem_type_names[] = {
+    [UCP_MEMTYPE_REG_NB_TYPES_AUTO] = "auto",
+    [UCP_MEMTYPE_REG_NB_TYPES_HOST] = "host",
+    [UCP_MEMTYPE_REG_NB_TYPES_CUDA] = "cuda",
+    [UCP_MEMTYPE_REG_NB_TYPES_LAST] = NULL
+};
```
brminich, line 146: i think we can reuse `ucs_memory_type_names` (like `MEMTYPE_REG_WHOLE_ALLOC_TYPES` does) and use `UCS_MEMORY_TYPE_LAST` to indicate auto value.

src/ucp/core/ucp_context.c
rakhmets: Is it ok that `MEMTYPE_REG_NB_TYPES=auto` is the same as `MEMTYPE_REG_NB_TYPES=`?

ivankochin: At this moment, yes. It is going to be different in the one of future configurations.

src/ucp/core/ucp_context.c
yosefe: I think we can define special config value for UCS_CONFIG_TYPE_BITMAP case, "all", which would set the bitmap to 0xff..ff.
"all" would mean to use any supported (by UCT) memory type.

brminich: sounds good, but we would need to be able to handle `auto` as well, so looks like need a bit for this (maybe use `UNKNOWN`)

yosefe: what would be difference between "auto" and "all"?

brminich: we can apply different default values on different platforms. E.g. sometimes it will be 0

yosefe: sounds weird to have specific platform logic in UCP
anyway I think "auto" is not needed at this point. we may consider adding it later.

src/ucp/core/ucp_context.c
```diff
@@ -444,6 +444,13 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "directory.",
    ucs_offsetof(ucp_context_config_t, proto_info_dir), UCS_CONFIG_TYPE_STRING},
 
+  {"MEMTYPE_REG_NB_TYPES", "auto",
+   "Memory types that perform non-blocking registration by default:\n"
+   "page registration may be deferred until it is accessed by the CPU or a transport.\n"
+   "Allowed memory types: cuda, host.",
```
ivankochin, line 450: @yosefe @brminich are that "default memory types" correct?

brminich: no need to mention those, also `cuda` is not going to be supported

src/ucp/core/ucp_context.c
```diff
@@ -444,6 +444,13 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "directory.",
    ucs_offsetof(ucp_context_config_t, proto_info_dir), UCS_CONFIG_TYPE_STRING},
 
+  {"MEMTYPE_REG_NB_TYPES", "",
+   "Memory types that perform non-blocking registration by default:\n"
+   "page registration may be deferred until it is accessed by the CPU or a transport.\n"
+   "Allowed memory types: cuda, host.",
```
ivankochin, line 450:
```suggestion
   "page registration may be deferred until it is accessed by the CPU or a transport.",
```

src/ucp/core/ucp_context.h
```diff
@@ -364,6 +366,9 @@ typedef struct ucp_context {
            unsigned               count;
            size_t                 *sizes;
         } am_mpools;
+
+        /* Memory types that perform non-blocking registration by default */
+        uint64_t                  reg_nb_mem_types;
```
yosefe, line 371: why needed here?
use it through context->config.ext

src/ucp/core/ucp_mm.c
```diff
@@ -586,6 +595,8 @@ ucp_memh_init_uct_reg(ucp_context_h context, ucp_mem_h memh, unsigned uct_flags)
     size_t length              = ucp_memh_length(memh);
     ucs_status_t status;
 
+    ucp_memh_apply_reg_nb_flags(context, mem_type, &uct_flags);
```
yosefe, line 598: do it only when there is a miss in rcache - in ucp_memh_register

src/ucp/core/ucp_context.c
```diff
@@ -444,6 +444,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "directory.",
    ucs_offsetof(ucp_context_config_t, proto_info_dir), UCS_CONFIG_TYPE_STRING},
 
+  {"MEMTYPE_REG_NB_TYPES", "",
```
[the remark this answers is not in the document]
ivankochin: I did the name by this way to be consistent with `MEMTYPE_REG_WHOLE_ALLOC_TYPES` that has the similar interface. Wouldn't it be better to keep this consistency?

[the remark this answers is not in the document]
yosefe: MEMTYPE_REG_WHOLE_ALLOC_TYPES is also a bad name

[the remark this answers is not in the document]
ivankochin:
```suggestion
  {"REG_NONBLOCK_MEM_TYPES", "",
```

src/ucp/core/ucp_mm.c
```diff
@@ -366,6 +366,15 @@ void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
     }
 }
 
+static void
+ucp_memh_apply_reg_nb_flags(ucp_context_h context, ucs_memory_type_t mem_type,
```
yosefe, line 370: maybe no need a separate function?
@brminich WDYT?

brminich: yes, we can avoid that

src/ucp/core/ucp_context.c
```diff
@@ -444,6 +444,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "directory.",
    ucs_offsetof(ucp_context_config_t, proto_info_dir), UCS_CONFIG_TYPE_STRING},
 
+  {"REG_NONBLOCK_MEM_TYPES", "",
```
brminich, line 447: general question, are we going to warn/diag if user sets this value to some memory type and we do not have any MD which supports non-blocking registration for such type?

yosefe (commented):
Currently missing the part that avoids registering a memory type that was configured in NONBLOCK mode on MDs that do not support nonblocking mode.

src/ucp/core/ucp_mm.c
```diff
@@ -427,6 +426,13 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
                                      UCT_MD_MEM_REG_FIELD_DMABUF_OFFSET;
         }
 
+        if (context->tl_mds[md_index].attr.reg_nonblock_mem_types &
+            (context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type))) {
+            reg_params.flags = uct_flags | UCT_MD_MEM_FLAG_NONBLOCK;
+        } else {
+            reg_params.flags = uct_flags;
```
brminich, line 433: let's add some `ucs_trace` here, or better add ODP related info to the trace on line 450 below.

brminich (approved):

src/ucp/core/ucp_mm.c
```diff
@@ -427,6 +426,13 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
...
+        }
```
yosefe, line 434: need to register only on MD that supports nonblocking reg (on the memory type)
on MD that doesn't support - we don't register AT ALL

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -767,7 +767,9 @@ uct_ib_mlx5_devx_check_odp(uct_ib_mlx5_md_t *md,
         } else {
             md->super.config.odp.max_size = 1ul << 28;
         }
+    }
 
+    if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, umr_extended_translation_offset)) {
```
brminich, line 772: this needs to be a part of another pr in UCT

src/ucp/core/ucp_mm.c
```diff
@@ -419,21 +420,24 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
     }
 
     ucs_for_each_bit(md_index, md_map) {
+        tl_md = &context->tl_mds[md_index];
+
+        if (context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) {
```
brminich, line 425: minor: can introduce some aux var and initialize it outside the loop, e.g.
`reg_nb_mtype = context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)`
and then use it here and on line 468

src/ucp/core/ucp_mm.c
```diff
@@ -461,6 +465,17 @@ static ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
...
+    if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
+        (md_map_registered == 0)) {
+        ucs_log(err_level,
+                "Available MDs cannot perform non-blocking registration "
+                "for %s memory type", ucs_memory_type_names[mem_type]);
+
+        if (!(uct_flags & UCT_MD_MEM_FLAG_HIDE_ERRORS)) {
+            status = UCS_ERR_UNSUPPORTED;
+        }
+    }
```
yosefe, line 477: IMO don't need to consider this an error. it's ok we don't register at all

src/ucp/core/ucp_mm.c
yosefe: we could initialize the map of memory domains to register on during ucp context initialization according to nonblocking capabilities

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Yes, I just add it here in testing purpose. BTW I think that it won't bring any harm if we merge before this UCT PR with this change.

src/ucp/core/ucp_mm.c
brminich: but we need to distinguish the cases when `context->config.ext.reg_nb_mem_types` is not set and when ODP is not supported by MD

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
brminich: yes, but we need to remove it from this PR

src/ucp/core/ucp_mm.c
ivankochin: We have assertion that will fail if we consider this case not an error. https://github.com/openucx/ucx/blob/master/src/ucp/core/ucp_mm.c#L849

Should we remove this assertion too?

BTW what is the case when not-registering memory region is OK? As I understand it will just fail during the later calls, am I correct?

brminich: in theory it will fall-back to some (possibly non-optimal) transport

brminich (approved):

src/ucp/core/ucp_context.c
```diff
@@ -444,6 +444,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
...
+   "Perform non-blocking memory registration for these memory types:\n"
```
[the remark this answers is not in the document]
ivankochin:
```suggestion
   "Perform only non-blocking memory registration for these memory types:\n"
```

brminich (approved):

yosefe (approved):
pls squash

brminich (approved):

PR #8864, opened 2023-02-12
UCP/CORE: refine get ep_init_flags

changchengx:
ucp_cm_ep_init_flags is called in ucp_ep_init_flags.

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -854,8 +854,7 @@ static ucs_status_t ucp_ep_create_to_sock_addr(ucp_worker_h worker,
 
     /* allocate endpoint */
     ucs_sockaddr_str(params->sockaddr.addr, peer_name, sizeof(peer_name));
-    ep_init_flags = ucp_ep_init_flags(worker, params) |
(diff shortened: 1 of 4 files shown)


src/ucp/core/ucp_ep.c
```diff
@@ -854,8 +854,7 @@ static ucs_status_t ucp_ep_create_to_sock_addr(ucp_worker_h worker,
 
     /* allocate endpoint */
     ucs_sockaddr_str(params->sockaddr.addr, peer_name, sizeof(peer_name));
-    ep_init_flags = ucp_ep_init_flags(worker, params) |
-                    ucp_cm_ep_init_flags(params);
+    ep_init_flags = ucp_ep_init_flags(worker, params);
```
rakhmets, line 857: This is the only place where `ucp_cm_ep_init_flags` is used.
You can remove the declaration of the function from `src/ucp/wireup/wireup_cm.h`,
and move the implementation of the function to `src/ucp/wireup/wireup.c`.

rakhmets (approved):

yosefe (approved):

PR #8865, opened 2023-02-13
UCP/EP: Fix rendezvous threshold for multi-path

yosefe:
## Why
Fix performance degradation due to using 2 paths on NDR400 by default

## How
Divide the estimated bandwidth of each lane by the number of paths on the resource.

This fix would be ported to v1.14.x branch.

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -1927,8 +1927,9 @@ static ucs_status_t ucp_ep_config_calc_params(ucp_worker_h worker,
                                               ucp_ep_thresh_params_t *params,
                                               int eager)
 {
-    ucp_context_h context = worker->context;
-    ucp_md_map_t md_map   = 0;
+    ucp_context_h context                = worker->context;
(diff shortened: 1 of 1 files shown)


src/ucp/core/ucp_ep.c
```diff
@@ -1964,7 +1970,9 @@ static ucs_status_t ucp_ep_config_calc_params(ucp_worker_h worker,
             }
         }
 
-        bw = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+        bw = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth) /
```
brminich, line 1973: maybe use `ucp_tl_iface_bandwidth_ratio` or create some other common function which would take `multi_path_ratio` into account

yosefe: per offline discussion we cannot use ucp_tl_iface_bandwidth_ratio function in this flow.
@brminich can we keep the PR as is then?

brminich (approved):

PR #8866, opened 2023-02-13
UCP/WORKER: convert EP configs storage to dynamic array

evgeny-leksikov:
## What
convert EP configurations storage to dynamic array

## Why ?
in some cases 64 EPs are not enough

## How ?
convert storage from static to dynamic array

### DIFF
--- a/src/ucp/core/ucp_worker.h
+++ b/src/ucp/core/ucp_worker.h
@@ -23,6 +23,8 @@
 #include <ucs/datastruct/ptr_map.h>
 #include <ucs/arch/bitops.h>
 
+#include <ucs/datastruct/array.inl>
+
 
 /* The size of the private buffer in UCT descriptor headroom, which UCP may
  * use for its own needs. This size does not include ucp_recv_desc_t length,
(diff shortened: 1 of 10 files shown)


src/ucp/core/ucp_worker.h
```diff
@@ -221,6 +223,9 @@ typedef struct ucp_worker_mpool_key {
 KHASH_TYPE(ucp_worker_mpool_hash, ucp_worker_mpool_key_t, ucs_mpool_t);
 typedef khash_t(ucp_worker_mpool_hash) ucp_worker_mpool_hash_t;
 
+/* EP configurations storage */
+UCS_ARRAY_DEFINE_INLINE(ep_config_arr, unsigned, ucp_ep_config_t);
```
yosefe, line 227: maybe the inline function should be in ucp_worker.inl
here only do UCS_ARRAY_DECLARE_TYPE

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -134,8 +134,9 @@ UCS_TEST_P(test_ucp_proto, dump_protocols) {
     ucp_worker_cfg_index_t rkey_cfg_index = UCP_WORKER_CFG_INDEX_NULL;
 
     auto select_elem = ucp_proto_select_lookup(
-            worker, &worker->ep_config[ep_cfg_index].proto_select, ep_cfg_index,
-            rkey_cfg_index, &select_param, 0);
+            worker,
+            &ucs_array_elem(&worker->ep_config, ep_cfg_index).proto_select,
```
yosefe, line 138: maybe use helper var (with auto type)

yosefe (approved):

PR #8870, opened 2023-02-14
TEST/UCP: Expect single thread without MT support build

tvegas1:
## What
Fixes `test_ucp_worker_thread_mode` test when built without `--enable-mt` configuration.

## Why ?
Test fails when built without thread support.

## How ?
Skip MT variant when built without thread support.

### DIFF
--- a/test/gtest/ucp/test_ucp_worker.cc
+++ b/test/gtest/ucp/test_ucp_worker.cc
@@ -750,7 +750,14 @@ UCS_TEST_P(test_ucp_worker_thread_mode, query)
     worker_attr.field_mask = UCP_WORKER_ATTR_FIELD_THREAD_MODE;
     ucs_status_t status    = ucp_worker_query(sender().worker(), &worker_attr);
     ASSERT_EQ(UCS_OK, status);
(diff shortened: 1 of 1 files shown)


test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -750,7 +750,14 @@ UCS_TEST_P(test_ucp_worker_thread_mode, query)
     worker_attr.field_mask = UCP_WORKER_ATTR_FIELD_THREAD_MODE;
     ucs_status_t status    = ucp_worker_query(sender().worker(), &worker_attr);
     ASSERT_EQ(UCS_OK, status);
-    EXPECT_EQ(thread_mode(), worker_attr.thread_mode);
+
+    ucs_thread_mode_t mode = thread_mode();
+#if !ENABLE_MT
```
brminich, line 755: maybe better check in `get_test_variants` and do not add multi-thread variants if compiled without mt support?

tvegas1: fixed by skipping the add of multi-thread variant above, and reran tests with and without mt.

brminich (approved):
👍

yosefe (approved):

PR #8874, opened 2023-02-14
clean some unnecessary code

changchengx:


src/ucp/core/ucp_context.c
```diff
@@ -1676,7 +1676,6 @@ static ucs_status_t ucp_fill_resources(ucp_context_h context,
 
     max_mds = 0;
     for (i = 0; i < context->num_cmpts; ++i) {
-        memset(&context->tl_cmpts[i].attr, 0, sizeof(context->tl_cmpts[i].attr));
```
yosefe, line 1679: but "resources" field will remain uninitialized

changchengx: @yosefe When allocating the context->tl_cmpts space in below code, it uses ```calloc``` API which fills the space with 0.
https://github.com/openucx/ucx/blob/v1.14.0-rc4/src/ucp/core/ucp_context.c#L1536-L1537
So, it doesn't need to use ```memset``` API to clear the space with 0 again.

yosefe: ok, thanks

yosefe (approved):

PR #8885, opened 2023-02-17
TEST/UCP: Separate progress to avoid concurrent ucp_worker_progress

tvegas1:
## What
Fix assert failure below, on tests like `test_ucp_fence*`, when built without `--enable-mt`:
```
==>  2807     ucs_assert(worker->inprogress++ == 0);
```

## Why ?
The sender thread calls `request_wait()` or `flush_worker()` that eventually calls `check_events()`. That later one progresses all entities, so including the sender and the receiver workers. This is an issue because the main thread is already progressing the receiver worker.

## How ?
To fix, we make sure that the sender thread does not progress the receiver worker by adding one new parameter to restrict processing to a subset of the list of entities. In case of `self`, only the sender needs to progress.

### DIFF
--- a/test/gtest/ucp/test_ucp_fence.cc
+++ b/test/gtest/ucp/test_ucp_fence.cc
@@ -28,7 +28,7 @@ public:
                                                 initial_buf, 1,
                                                 (uintptr_t)memheap_addr, rkey,
                                                 &param);
-        ucs_status_t status = request_wait(request);
+        ucs_status_t status = request_wait(request, {e});
         ASSERT_UCS_OK(status);
     }
 
@@ -39,7 +39,7 @@ public:
         void *request = ucp_atomic_fetch_nb(e->ep(), UCP_ATOMIC_FETCH_OP_FADD,
                                             *initial_buf, (T*)result_buf, sizeof(T),
                                             (uintptr_t)memheap_addr, rkey, send_cb);
-        request_wait(request);
+        request_wait(request, {e});
     }
 
     template <typename T, typename F>
@@ -95,7 +95,7 @@ public:
                 (test->*m_send_2)(m_entity, &zero, &result,
                                   m_memheap, m_rkey);
 
-                test->flush_worker(*m_entity);
+                test->flush_worker(*m_entity, 0, {m_entity});
 
                 if (result != (uint64_t)(i+1))
                     (*error)++;
@@ -115,14 +115,14 @@ public:
                      ucp_rkey_h rkey, void *memheap_ptr,
                      uint64_t initial_value, uint32_t* error) {
         ucs::ptr_vector<worker> m_workers;
-        std::vector<entity*> entities;
         m_workers.clear();
         m_workers.push_back(new worker(this, send1, send2, sender, rkey,
                                        memheap_ptr, initial_value, error));
-        entities.push_back(&receiver());
-        while (m_workers.at(0).running) {
+        if (!is_loopback()) {
             /* allow receiver to progress incoming ops */
-            progress(entities, 0);
+            while (m_workers.front()->running) {
+                progress({&receiver()});
+            }
         }
         m_workers.at(0).join();
         m_workers.clear();
--- a/test/gtest/ucp/ucp_test.cc
+++ b/test/gtest/ucp/ucp_test.cc
@@ -185,13 +185,14 @@ void ucp_test::short_progress_loop(int worker_index) const {
 void ucp_test::flush_ep(const entity &e, int worker_index, int ep_index)
 {
     void *request = e.flush_ep_nb(worker_index, ep_index);
-    request_wait(request, worker_index);
+    request_wait(request, {}, worker_index);
 }
 
-void ucp_test::flush_worker(const entity &e, int worker_index)
+void ucp_test::flush_worker(const entity &e, int worker_index,
+                            const std::vector<entity*> &entities)
 {
     void *request = e.flush_worker_nb(worker_index);
-    request_wait(request, worker_index);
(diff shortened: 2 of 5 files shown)


test/gtest/ucp/ucp_test.h
```diff
@@ -242,6 +245,8 @@ class ucp_test : public ucp_test_base,
     void short_progress_loop(int worker_index = 0) const;
     void flush_ep(const entity &e, int worker_index = 0, int ep_index = 0);
     void flush_worker(const entity &e, int worker_index = 0);
+    void flush_worker(const entity &e, const std::vector<entity*> &entities,
```
yosefe, line 248: can we define
```cpp
void flush_worker(const entity &e, int worker_index = 0, 
                  const std::vector<entity*> &entities = entities());
```

and same for request_wait

test/gtest/ucp/test_ucp_fence.cc
```diff
@@ -39,7 +40,8 @@ class test_ucp_fence : public ucp_test {
         void *request = ucp_atomic_fetch_nb(e->ep(), UCP_ATOMIC_FETCH_OP_FADD,
                                             *initial_buf, (T*)result_buf, sizeof(T),
                                             (uintptr_t)memheap_addr, rkey, send_cb);
-        request_wait(request);
+        std::vector<entity*> entities = {e};
+        request_wait(request, entities);
```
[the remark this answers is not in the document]
tvegas1: It should not hang because the main thread executing `run_workers()` below is taking care of progressing the receiver side. 

Actually, I tried removing the `progress({&receiver()})` call in `run_workers()` and the tests properly hangs.

test/gtest/ucp/test_ucp_fence.cc
```diff
@@ -108,21 +111,23 @@ class test_ucp_fence : public ucp_test {
         void *m_memheap;
         send_func_t m_send_1, m_send_2;
         entity* m_entity;
+        const std::vector<entity*> m_local_entity;
```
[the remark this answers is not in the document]
tvegas1: fixed, removed completely.

test/gtest/ucp/test_ucp_fence.cc
```diff
@@ -108,21 +111,23 @@ class test_ucp_fence : public ucp_test {
...
         m_workers.clear();
         m_workers.push_back(new worker(this, send1, send2, sender, rkey,
                                        memheap_ptr, initial_value, error));
-        entities.push_back(&receiver());
-        while (m_workers.at(0).running) {
+        if (!is_loopback()) {
             /* allow receiver to progress incoming ops */
-            progress(entities, 0);
+            while (m_workers.front()->running) {
+                progress(entities);
```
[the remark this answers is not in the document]
tvegas1: fixed.

test/gtest/ucp/ucp_test.h
tvegas1: Fixed, cannot use non-static class members in default argument, but added `&entities = {}` instead. Now using `entities()` when the passed vector size is zero. (could not get rid of the two `check_events()` call as we are using a reference).

test/gtest/ucp/ucp_test.cc
```diff
@@ -287,7 +289,11 @@ ucs_status_t ucp_test::request_process(void *req, int worker_index, bool wait,
     while (wait &&
            (ucp_request_check_status(req) == UCS_INPROGRESS) &&
            (ucs_get_time() < deadline)) {
-        check_events(entities(), wakeup, worker_index);
+        if (entity_list.size()) {
```
yosefe, line 292: minor: entity_list.empty() (and inverse if/else)
or:
```c
        check_events(entity_list.empty() ? entities() : entity_list, wakeup, 
                     worker_index);
```

tvegas1: fixed

yosefe (dismissed):

test/gtest/ucp/ucp_test.h
```diff
@@ -241,15 +242,18 @@ class ucp_test : public ucp_test_base,
...
+                      const std::vector<entity*> &entities = {});
     void flush_workers();
     void disconnect(entity& entity);
     void check_events(const std::vector<entity*> &entities, bool wakeup,
                       int worker_index = 0);
     ucs_status_t
     request_progress(void *req, const std::vector<entity*> &entities,
                      double timeout = 10.0, int worker_index = 0);
-    ucs_status_t request_wait(void *req, int worker_index = 0, bool wakeup = false);
+    ucs_status_t request_wait(void *req, int worker_index = 0,
```
brminich, line 254: minor: maybe change the order of parameters in this method, so that we could omit specifying `worker_index` and `wakeup` in most of the cases?
So it would be 
`request_wait(request, {e});`   instead of
`request_wait(request, 0, false, {e});`

tvegas1: Actually the first PR pushed was like your suggestion, but to address review comment, I moved the `{e}` parameter at the end of the function. I am ok with the two options, but I guess we should keep it like this now?

brminich: i did not notice that, but looks like less code is needed with the original version
@yosefe, wdyt?

yosefe: I'm ok with it as long as we have default value {}

tvegas1: fixed it and kept default valud `{}`

brminich (approved):

yosefe (approved):
pls squash

brminich (approved):

PR #8887, opened 2023-02-18
UCT/RDMACM: show log_max_num_reserved_qpn info in debug log

changchengx:
## What
print log_max_num_reserved_qpn if it can use reserved qpn in rdmacm

## Why ?
debug purpose


src/uct/ib/rdmacm/rdmacm_cm.c
```diff
@@ -161,8 +164,9 @@ uct_rdmacm_cm_device_context_init(uct_rdmacm_cm_device_context_t *ctx,
 
     uct_rdmacm_cm_reserved_qpn_blk_release(blk);
 
-    ucs_debug("%s reserved qpn cap: log_reserved_qpn_granularity is 0x%x",
-              dev_name, ctx->log_reserved_qpn_granularity);
+    ucs_debug("%s with reserved qpn cap, log_max_num_reserved_qpn=%d, "
```
yosefe, line 167: pls remove all `,` from the log line

changchengx: Done

yosefe (approved):

PR #8888, opened 2023-02-18
support enable specify repeat connection to same server

changchengx:
## What
use new server info format ```<server-ip>:<port>[:<count>]``` to support repeat connect with same server.

## Why ?
process argument number is limited, e.g. getconf ARG_MAX. The current io_demo process can't accept too long arguments.

## How ?
support [server_ip:port:repeat_count] to specify the number for repeating connection with same server.

### DIFF
--- a/test/apps/iodemo/io_demo.cc
+++ b/test/apps/iodemo/io_demo.cc
@@ -32,8 +32,9 @@
 #include <cuda_runtime.h>
 #endif
 
-#define ALIGNMENT           4096
-#define BUSY_PROGRESS_COUNT 1000
+#define ALIGNMENT               4096
+#define BUSY_PROGRESS_COUNT     1000
+#define MAX_SERVER_REPEAT_COUNT (65536U - 1024U)
 
 /* IO operation type */
 typedef enum {
@@ -62,7 +63,7 @@ const bool do_assert = false;
 
 /* test options */
 typedef struct {
-    std::vector<const char*> servers;
+    std::vector<std::string> servers;
     int                      port_num;
     double                   connect_timeout;
     double                   client_timeout;
@@ -2068,7 +2069,7 @@ public:
 
     void connect(size_t server_index)
     {
-        const char *server = opts().servers[server_index];
+        const char *server = opts().servers[server_index].c_str();
         struct sockaddr_storage *src_addr_p = NULL;
         struct sockaddr_storage dst_addr, src_addr;
         uint32_t addr_index;
@@ -2682,6 +2683,40 @@ static int parse_window_size(const char *optarg, long &window_size,
     return 0;
 }
 
+static int add_servers(const char *server, std::vector<std::string> &servers)
+{
+    const char *repeat_separator;
+    const char *port_separator;
+    std::string server_repeated;
+    int repeat_count;
+
+    port_separator = strchr(server, ':');
+    if (port_separator == NULL) {
+        servers.push_back(server);
+        return 0;
+    }
+
+    repeat_separator = strchr(port_separator + 1, ':');
+    if (repeat_separator == NULL) {
+        servers.push_back(server);
+        return 0;
+    }
+
+    server_repeated = std::string(server, repeat_separator - server);
+    repeat_count    = atoi(repeat_separator + 1);
+    if ((repeat_count == 0) || (repeat_count > MAX_SERVER_REPEAT_COUNT)) {
+        std::cout << "Server repeat_count should be in the range "
+                  << "[1.. " << MAX_SERVER_REPEAT_COUNT << "]" << std::endl;
+        return -1;
+    }
+
+    for (int i = 0; i < repeat_count; i++) {
+        servers.push_back(server_repeated);
+    }
+
+    return 0;
+}
+
 static int parse_args(int argc, char **argv, options_t *test_opts)
 {
     static const char *optstring =
@@ -2881,8 +2916,9 @@ static int parse_args(int argc, char **argv, options_t *test_opts)
             break;
         case 'h':
         default:
-            std::cout << "Usage: io_demo [options] [server_address]" << std::endl;
-            std::cout << "       or io_demo [options] [server_address0:port0] [server_address1:port1]..." << std::endl;
+            std::cout << "Usage: io_demo [options] [server_address[:port:[repeat_count]]]" << std::endl;
+            std::cout << "Server repeat_count should be in the range "
+                      << "[1.. " << MAX_SERVER_REPEAT_COUNT << "]" << std::endl;
             std::cout << "" << std::endl;
             std::cout << "Supported options are:" << std::endl;
             std::cout << "  -p <port>                   TCP port number to use" << std::endl;
@@ -2929,7 +2965,9 @@ static int parse_args(int argc, char **argv, options_t *test_opts)
     }
 
     while (optind < argc) {
-        test_opts->servers.push_back(argv[optind++]);
+        if (add_servers(argv[optind++], test_opts->servers) != 0) {
+            return -1;
+        };
     }
 
     adjust_opts(test_opts);


yosefe:
@changchengx maybe instead of long command line better provide a file name with the list of servers?

changchengx:
> @changchengx maybe instead of long command line better provide a file name with the list of servers?
>
@yosefe
1) Currently, there're no lots of servers infrastructure in the lab.
2) While issuing burst connections from client to server, two server hosts(client role & server role) are enough until all the QPs are used out. It doesn't need to enumerate too many **different** server info in the command line. **Most of the enumerated server info are same** and this PR uses ```-x``` to indicate how many times the server info should be repeated.
3) If we really need to enumerate **tens thousands of different server info** in the command line, we can ```provide a file name with the list of servers``` because the process's arguments length are limited by linux system(e.g. ```getconf ARG_MAX``` ).
Could we use this PR solution to solve the current problem wrote in the **Why?** section? I haven't imaged the scenario the it needs tens thousands of different server info and mock the scenario in the lab infrastructure.

yosefe:
@changchengx so the problem this PR is trying to solve is connecting to same server multiple times without passing it in the command line multiple times.
WDYT to support a syntax of `<server-ip>:<port>[:<count>]`, for example 1.1.1.1:3245:100, which would mean to create 100 connections to 1.1.1.1 port 3245. if not provided, the count would be 1.

changchengx:
@yosefe Agree with the ```syntax of <server-ip>:<port>[:<count>]``` solution. I'll follow this idea to implement the code.

test/apps/iodemo/io_demo.cc
```diff
@@ -2664,6 +2669,38 @@ static void adjust_opts(options_t *test_opts) {
         test_opts->chunk_size = std::min(test_opts->chunk_size,
                                          test_opts->max_data_size);
     }
+
+    if (servers.empty()) {
+        return;
+    }
```
yosefe, line 2675: why needed?

test/apps/iodemo/io_demo.cc
```diff
@@ -2664,6 +2669,38 @@ static void adjust_opts(options_t *test_opts) {
...
+
+    for (auto server = servers.cbegin(); server != servers.cend(); ++server) {
+        repeat_separator = strchr(*server, ':');
+        if (repeat_separator == NULL) {
+            test_opts->servers.push_back(std::string(*server));
```
yosefe, line 2680: can remove std::string it will use the constructor automatically

test/apps/iodemo/io_demo.cc
```diff
@@ -2664,6 +2669,38 @@ static void adjust_opts(options_t *test_opts) {
         test_opts->chunk_size = std::min(test_opts->chunk_size,
                                          test_opts->max_data_size);
     }
+
+    if (servers.empty()) {
+        return;
+    }
+
+    for (auto server = servers.cbegin(); server != servers.cend(); ++server) {
```
yosefe, line 2677: can we use for (auto server: servers)

test/apps/iodemo/io_demo.cc
```diff
@@ -2664,6 +2669,38 @@ static void adjust_opts(options_t *test_opts) {
...
+        repeat_separator = strchr(*server, ':');
+        if (repeat_separator == NULL) {
```
yosefe, line 2679: isn't this port_separator?

test/apps/iodemo/io_demo.cc
```diff
@@ -2916,10 +2955,10 @@ static int parse_args(int argc, char **argv, options_t *test_opts)
     }
 
     while (optind < argc) {
-        test_opts->servers.push_back(argv[optind++]);
+        servers.push_back(argv[optind++]);
```
yosefe, line 2958: maybe instead of using local server array, we can call something like add_servers(argv[optind++]) for each argument. and add each server including repeat count if needed

test/apps/iodemo/io_demo.cc
```diff
@@ -2664,6 +2669,38 @@ static void adjust_opts(options_t *test_opts) {
...
+        for (auto i = 0; i < repeat_count; i++) {
+            test_opts->servers.push_back(server_repeated);
+
+            if (test_opts->src_addrs.empty()) {
+                continue;
+            }
+
+            /* align with round robin method to select the src address */
+            test_opts->src_addrs.push_back(test_opts->src_addrs.at(
+                        std::distance(servers.cbegin(), server)));
```
yosefe, line 2701: can't it exceed the size of src_addrs array?

changchengx: I'll fix the wrong logic here.

changchengx:
I'll refine this PR this today or next weekend until passing all the CI test cases.

test/apps/iodemo/io_demo.cc
```diff
@@ -2916,10 +2953,19 @@ static int parse_args(int argc, char **argv, options_t *test_opts)
...
 
-    adjust_opts(test_opts);
+    if (!test_opts->src_addrs.empty() &&
+        (test_opts->src_addrs.size() != 1) &&
+        (test_opts->src_addrs.size() != servers.size())) {
+        std::cout << "In client size, source addr number should be 1 "
+                  << "OR equal server addr number" << std::endl;
+        return -1;
+    }
+
```
yosefe, line 2966: why is this needed?

test/apps/iodemo/io_demo.cc
yosefe: can we round-robin the src addresses between all servers that we created (either directly or by repeat count)?
IMO it will make the code simpler and more generic - remove the dependency between test_opts->src_addrs.size() and servers.size()

test/apps/iodemo/io_demo.cc
```diff
@@ -2664,6 +2670,35 @@ static void adjust_opts(options_t *test_opts) {
...
+        repeat_separator = strchr(port_separator + 1, ':');
+        if (repeat_separator == NULL) {
+            test_opts->servers.push_back(servers[idx]);
+            continue;
+        }
+
+        server_repeated =
+            std::string(servers[idx], repeat_separator - servers[idx]);
+        repeat_count    = atoi(repeat_separator + 1);
+        for (int i = 0; i < repeat_count; i++) {
```
yosefe, line 2690: To solve Coverity error, probably need to set upper bound for repeat_count, for example
```cpp
#define MAX_SERVER_REPEAT_COUNT 100000ul
...
int parse_server(const char *server_arg) {
    ...
    repeat_count = atoi(repeat_separator + 1);
    if (repeat_count > MAX_SERVER_REPEAT_COUNT) {
        std::cout << "Exceeded maximal server repeat count (" << MAX_SERVER_REPEAT_COUNT << ")" << std::endl;
        return -1;
    }
```

test/apps/iodemo/io_demo.cc
changchengx: >
>can we round-robin the src addresses between all servers that we created (either directly or by repeat count)?
>
I didn't understand it well.
1. ```src_addrs``` is used one by one to connect with the server.
2. suppose there're two source addresses ```src_ip0, src_ip1``` and there're two servers ```server0:port0, server1:port1:2```.
    1) The current logic is to use ```src_ip0``` to connect with ```server0:port0``` and use ```src_ip1``` to connect with ```server1:port1``` for ```2``` times.
    2) Do you mean that it should ```src_ip0``` to connect with ```server0:port0``` and use ```src_ip1``` to connect with ```server1:port1``` for the first time, then use ```src_ip0``` to connect with ```server1:port1``` for the second time?

test/apps/iodemo/io_demo.cc
changchengx: It requires the src_addrs are used to connect with the server one by one
1) ```empty``` case: use default destination reachable src ip to connect with server.
2) ```1``` case: the only given src ip could connect with all servers.
3) ```src_addrs.size() = servers.size()``` case: the src addresses are used one by one to connect with the servers.

test/apps/iodemo/io_demo.cc
yosefe: in case (2), there are actually 3 servers: server0:port0, server1:port1, server1:port1
yes, i would suggest in this case we use src_addrs in the following order: 
server0:port0 - src_ip0
server1:port1 - src_ip1
server1:port1 - src_ip0

yosefe: it's up to the user to specify right combination of server and src_addr

changchengx: Got it. I'll simplify the code here and update this PR implementation on weekend time.

test/apps/iodemo/io_demo.cc
changchengx: Done

test/apps/iodemo/io_demo.cc
changchengx: Done

test/apps/iodemo/io_demo.cc
changchengx: Done

test/apps/iodemo/io_demo.cc
changchengx: Done

test/apps/iodemo/io_demo.cc
changchengx: Done. Logic has been removed after simplifying the design

test/apps/iodemo/io_demo.cc
changchengx: Removed the check here after simplifying the design

test/apps/iodemo/io_demo.cc
changchengx: Done. I also check whether the ```repeat_count``` is ```0``` here to 1) check wrong input 2) ```0``` should not be used by client.

test/apps/iodemo/io_demo.cc
changchengx: The code is simplified a lot after following this idea to make the design.

changchengx:
The failure in CI test case ```Tests BlueField on worker 0``` is not related with this PR. The error log is below:
```##[error]The job running on agent swx-rain03-bf1-01 ran longer than the maximum time of 360 minutes. For more information, see https://go.microsoft.com/fwlink/?linkid=2077134```

File: test/apps/iodemo/io_demo.cc [hunk 9]
```diff
@@ -2682,6 +2683,40 @@ static int parse_window_size(const char *optarg, long &window_size,
...
+    if (repeat_separator == NULL) {
+        servers.push_back(server);
+        return 0;
+    }
+
+    server_repeated = std::string(server, repeat_separator - server);
+    repeat_count    = atoi(repeat_separator + 1);
+    if ((repeat_count == 0) || (repeat_count > MAX_SERVER_REPEAT_COUNT)) {
+        std::cout << "repeat connection number should be in range "
+                  << "[1, " << MAX_SERVER_REPEAT_COUNT << "]" << std::endl;
```
yosefe, line 2709:
```cpp
    std::cout << "Server repeat_count should be in the range "
              << "[1.. " << MAX_SERVER_REPEAT_COUNT << "]" << std::endl;
```

yosefe: @changchengx unresolved

changchengx: @yosefe Done

yosefe (dismissed):
pls squash

yosefe (approved):

PR #8889, opened 2023-02-18
README: build from git or from tarball

bkmgit:
## What
Update installation instructions in the README 

## Why ?
There is no autogen script, use autoconf -fi instead.

## How ?
Update Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,16 @@ Please review our ["Membership Voluntary Consensus Standard"](https://ucfconsort
 ### Release Builds
 
 Building UCX is typically a combination of running "configure" and "make".
-Execute the following commands to install the UCX system from within the
-directory at the top of the tree:
+If using a release tarball execute the following commands to install the
+UCX system from within the directory at the top of the tree:
+
+```sh
+$ ./contrib/configure-release --prefix=/where/to/install
+$ make -j8
+$ make install
+```
+
+If directly cloning the git repository use:
 
 ```sh
 $ ./autogen.sh


bkmgit:
autogen.sh is available when cloning from git, but not in the release tarballs. The README is also distributed with the release tarballs.

yosefe:
> autogen.sh is available when cloning from git, but not in the release tarballs. The README is also distributed with the release tarballs.

Need to update the instructions that autogen should be run only when cloning from git. Otherwise could just use `./configure`, no need to re-run autoconf.

yosefe:
@bkmgit Thank you for your contribution, in order to merge code to UCX can you please sign the contributor agreement - see https://github.com/openucx/ucx/blob/master/.github/CONTRIBUTING.md

bkmgit:
Thanks. Related issue on CLA signing https://github.com/openucx/ucx/issues/8892

yosefe:
> > autogen.sh is available when cloning from git, but not in the release tarballs. The README is also distributed with the release tarballs.
> 
> Need to update the instructions that autogen should be run only when cloning from git. Otherwise could just use `./configure`, no need to re-run autoconf.

@bkmgit to clarify, we don't need to advise running autoconf directly in the README. Should advise not running it at all with source release tarball.

yosefe (approved):
pls squash to a single commit and force-push

bkmgit:
Done

yosefe (approved):

PR #8893, opened 2023-02-19
BUILDLIB/PR: Retry on test failure

Alexey-Rivkin:
## What
Add auto-retry count for flaky tests.

## Why ?
Some test failures are environmental, thus unrelated to the PR itself.

### DIFF
--- a/buildlib/pr/tests.yml
+++ b/buildlib/pr/tests.yml
@@ -32,6 +32,7 @@ jobs:
           ./contrib/test_jenkins.sh
           exit $?
         displayName: Run ./contrib/test_jenkins.sh
+        retryCountOnTaskFailure: 2
         env:
           nworkers: ${{ parameters.num_workers }}
           worker: $(worker_id)


buildlib/pr/tests.yml
```diff
@@ -32,6 +32,7 @@ jobs:
           ./contrib/test_jenkins.sh
           exit $?
         displayName: Run ./contrib/test_jenkins.sh
+        retryCountOnTaskFailure: 2
```
yosefe, line 35: which tests would it affect?

Al3xR01: This change should eventually affect every test that might fail due to network timeout etc.
Let's see how it works and evaluate the need to extend it to IO demo tests too.

yosefe: This may hide random, but real, issues 
I prefer to review failure before restarting.
we can add it only to build tests, which are deterministic.

Alexey-Rivkin: Good point!
For now I'm using this PR to test `/azp run perf` failures.
Please, unsubscribe.

PR #8894, opened 2023-02-19
AZP/RELEASE: Add ARM release - Part #1

Alexey-Rivkin:
## What
Build and release ARM-based artifacts.
Part 1: Dockerfiles.

1. Unified Dockerfiles and docker compose files for both platforms.
2. Update MOFED to the minimal 5.x version supported by both x86 and ARM platforms.
3. Address minor remarks from PR https://github.com/openucx/ucx/pull/8907

Depends on https://github.com/openucx/ucx/pull/8928

## Why ?
https://github.com/openucx/ucx/issues/8762


yosefe (commented):
can't we reuse docker files for arm and x86_64 to avoid duplication?
what is the difference between arm and x86_64?

Al3xR01:
> can't we reuse docker files for arm and x86_64 to avoid duplication? what is the difference between arm and x86_64?

We can create multi-platform Docker images.
I suggest that this be treated as a distinct project.

yosefe:
> > can't we reuse docker files for arm and x86_64 to avoid duplication? what is the difference between arm and x86_64?
> 
> We can create multi-platform Docker images. I suggest that this be treated as a distinct project.

Can we generate images for different arch from same source docker file?

Al3xR01:
> > > can't we reuse docker files for arm and x86_64 to avoid duplication? what is the difference between arm and x86_64?
> > 
> > 
> > We can create multi-platform Docker images. I suggest that this be treated as a distinct project.
> 
> Can we generate images for different arch from same source docker file?

Base images from nvidia/cuda vary slightly between platforms. Dockerfiles per platform contain adaptations.
It should be possible to create unified Dockerfiles, but it's a tedious and time-consuming task.

Al3xR01:
Done

buildlib/dockers/docker-compose-aarch64.yml
```diff
@@ -0,0 +1,55 @@
+version: "3.4"
+
+# How to:
+# export ARCH=$(uname -m)
+# docker compose -f docker-compose-${ARCH}.yml build
+
+services:
```
yosefe, line 7: docker-compose-aarch64.yml docker-compose-x86_64.yml still seem to have a lot of "code duplication", can we have common yaml file for them?
how do they differ from one another?

buildlib/dockers/docker-compose-x86_64.yml
```diff
@@ -0,0 +1,43 @@
...
+        OS_VERSION: 7
+        ARCH: ${ARCH}
+  centos8-mofed5-cuda11:
+    image: centos8-mofed5-cuda11:5
+    build:
+      context: .
+      network: host
+      dockerfile: centos-release.Dockerfile
+      args:
+        MOFED_VERSION: 5.0-1.0.0.0
```
yosefe, line 39: do we need CUDA_COMPAT here?

buildlib/dockers/docker-compose-aarch64.yml
Al3xR01: 1. MOFED and CUDA support matrixes differ between CPU architectures.
2. Not a single image spec is shared between the two yml files.
3. Docker compose does not support a conditional statement.
With the existing manual process of building images in place, I see no other choice but to keep the two yml files separated.

buildlib/dockers/docker-compose-x86_64.yml
Al3xR01: No more cuda-compat usage.

buildlib/dockers/docker-compose-aarch64.yml
Al3xR01: Your wish was granted. Please, review.

buildlib/dockers/docker-compose.yml
```diff
@@ -26,41 +35,49 @@ services:
...
   ubuntu18.04-mofed5-cuda11:
-    image: ubuntu18.04-mofed5-cuda11:3
+    image: ubuntu18.04-mofed5-cuda11:5
     build:
       context: .
       network: host
       dockerfile: ubuntu-release.Dockerfile
       args:
-        MOFED_VERSION: 5.0-1.0.0.0
+        MOFED_VERSION: 5.4-3.6.8.1
```
yosefe, line 60: why do we have to bump version? because of ARM?
I'm afraid it means we support MLNX_OFED >=5.4 now, instead of >= 5.0

Al3xR01: Because of ARM. 
The alternative is to keep separate docker-compose.yml files per platform, as suggested before.

buildlib/dockers/docker-compose-x86_64.yml
```diff
@@ -69,4 +89,4 @@ services:
         MOFED_VERSION: 5.4-3.6.8.1
         UBUNTU_VERSION: 22.04
         CUDA_VERSION: 11.7.0
-        NV_DRIVER_VERSION: 515
+        ARCH: ${ARCH}
```
yosefe, line 92: add new line

buildlib/dockers/docker-compose-aarch64.yml
yosefe: this file is docker-compose-**aarch64**.yml
Maybe it should just define `ARCH: aarch64` ?
same for x86_64

buildlib/dockers/docker-compose-aarch64.yml
```diff
@@ -0,0 +1,60 @@
+version: "3.4"
+# To build images, run:
+# export ARCH=$(uname -m) && docker compose build
+# docker compose -f docker-compose-${ARCH}.yml build
```
yosefe, line 4: can it be 
`docker compose -f docker-compose-$(uname -m).yml build`

buildlib/dockers/docker-compose-aarch64.yml
Al3xR01: Right, those are leftovers of attempts to select platforms dynamically,

buildlib/dockers/docker-compose-aarch64.yml
Al3xR01: Changed to fixed ARCH values.

yosefe (dismissed):

yosefe:
@Al3xR01 if this is done form your side pls squash

Al3xR01:
Squashed

yosefe (approved):

artemry-nv (approved):

PR #8899, opened 2023-02-22
UCP/WIREUP: Use distance info in lane score calculation with new protocols

SeyedMir:
## What
This PR adds support for using GPU-NIC distance information when calculating lane scores even if the new protocols are enabled.

## Why ?
During lane creation, GPU-NIC distance information and its impact on latency/bandwidth was used only if new protocols were NOT enabled. Otherwise, we could end up selecting sub-optimal NICs for GPU buffers.

## How ?
1. Add distance information to worker iface so we can use it for local iface.
2. Use distance-updated latency/bandwidth for remote iface attributes during address pack.

### DIFF
--- a/src/ucp/core/ucp_worker.c
+++ b/src/ucp/core/ucp_worker.c
@@ -1175,7 +1175,7 @@ static void ucp_worker_close_ifaces(ucp_worker_h worker)
     UCS_ASYNC_UNBLOCK(&worker->async);
 }
 
-static ucs_status_t
+static void
 ucp_worker_get_sys_device_distance(ucp_context_h context,
                                    ucp_rsc_index_t rsc_index,
                                    ucs_sys_dev_distance_t *distance)
@@ -1183,6 +1183,8 @@ ucp_worker_get_sys_device_distance(ucp_context_h context,
     ucs_sys_device_t device     = UCS_SYS_DEVICE_ID_UNKNOWN;
     ucs_sys_device_t cmp_device = UCS_SYS_DEVICE_ID_UNKNOWN;
     ucp_rsc_index_t md_index, i;
+
+    *distance = ucs_topo_default_distance;
 
     for (i = 0; i < context->num_tls; i++) {
         md_index = context->tl_rscs[i].md_index;
@@ -1194,10 +1196,8 @@ ucp_worker_get_sys_device_distance(ucp_context_h context,
         device     = context->tl_rscs[rsc_index].tl_rsc.sys_device;
         cmp_device = context->tl_rscs[i].tl_rsc.sys_device;
 
-        return ucs_topo_get_distance(device, cmp_device, distance);
+        ucs_topo_get_distance(device, cmp_device, distance);
     }
-
-    return UCS_ERR_NO_RESOURCE;
 }
 
 ucs_status_t ucp_worker_iface_open(ucp_worker_h worker, ucp_rsc_index_t tl_id,
@@ -1207,7 +1207,6 @@ ucs_status_t ucp_worker_iface_open(ucp_worker_h worker, ucp_rsc_index_t tl_id,
     ucp_context_h context            = worker->context;
     ucp_tl_resource_desc_t *resource = &context->tl_rscs[tl_id];
     uct_md_h md                      = context->tl_mds[resource->md_index].md;
-    ucs_sys_dev_distance_t distance  = {.latency = 0, .bandwidth = 0};
     uct_iface_config_t *iface_config;
     ucp_worker_iface_t *wiface;
     ucs_status_t status;
@@ -1304,16 +1303,14 @@ ucs_status_t ucp_worker_iface_open(ucp_worker_h worker, ucp_rsc_index_t tl_id,
         goto err_close_iface;
     }
 
+    ucp_worker_get_sys_device_distance(context, wiface->rsc_index,
+                                       &wiface->distance);
     if (!context->config.ext.proto_enable) {
-        status = ucp_worker_get_sys_device_distance(context, wiface->rsc_index,
-                                                    &distance);
-        if (status == UCS_OK) {
-            wiface->attr.latency.c          += distance.latency;
-            wiface->attr.bandwidth.shared    =
-                    ucs_min(wiface->attr.bandwidth.shared, distance.bandwidth);
-            wiface->attr.bandwidth.dedicated = ucs_min(
-                    wiface->attr.bandwidth.dedicated, distance.bandwidth);
-        }
+        wiface->attr.latency.c          += wiface->distance.latency;
+        wiface->attr.bandwidth.shared    = ucs_min(wiface->attr.bandwidth.shared,
+                                                   wiface->distance.bandwidth);
+        wiface->attr.bandwidth.dedicated = ucs_min(
+                wiface->attr.bandwidth.dedicated, wiface->distance.bandwidth);
     }
 
     ucs_debug("created interface[%d]=%p using "UCT_TL_RESOURCE_DESC_FMT" on worker %p",
--- a/src/ucp/core/ucp_worker.h
+++ b/src/ucp/core/ucp_worker.h
@@ -236,6 +236,7 @@ struct ucp_worker_iface {
     ucp_worker_h                  worker;        /* The parent worker */
     ucs_list_link_t               arm_list;      /* Element in arm_ifaces list */
     ucp_rsc_index_t               rsc_index;     /* Resource index */
+    ucs_sys_dev_distance_t        distance;      /* Distance from given MD */
     int                           event_fd;      /* Event FD, or -1 if undefined */
     unsigned                      activate_count;/* How many times this iface has
                                                     been activated */
--- a/src/ucp/wireup/address.c
+++ b/src/ucp/wireup/address.c
@@ -667,16 +667,15 @@ static uint64_t ucp_address_flags_from_iface_flags(uint64_t iface_cap_flags,
 }
 
 static unsigned
-ucp_address_pack_iface_attr_v1(ucp_worker_h worker, void *ptr,
-                               const uct_iface_attr_t *iface_attr,
+ucp_address_pack_iface_attr_v1(const ucp_worker_iface_t *wiface, void *ptr,
                                unsigned atomic_flags)
 {
+    const uct_iface_attr_t *iface_attr      = &wiface->attr;
     ucp_address_packed_iface_attr_t *packed = ptr;
 
-    packed->overhead       = iface_attr->overhead;
-    packed->bandwidth      = ucp_tl_iface_bandwidth(worker->context,
-                                                    &iface_attr->bandwidth);
-    packed->lat_ovh        = iface_attr->latency.c;
+    packed->lat_ovh   = ucp_wireup_iface_lat_distance_v1(wiface);
+    packed->bandwidth = ucp_wireup_iface_bw_distance(wiface);
+    packed->overhead  = iface_attr->overhead;
     /* Pack prio, capability and atomic flags */
     packed->prio_cap_flags = (uint8_t)iface_attr->priority |
                              ucp_address_pack_flags(iface_attr->cap.flags,
@@ -717,23 +716,24 @@ size_t ucp_address_iface_seg_size(const uct_iface_attr_t *iface_attr)
 }
 
 static unsigned
-ucp_address_pack_iface_attr_v2(ucp_worker_h worker, void *ptr,
-                               const uct_iface_attr_t *iface_attr,
+ucp_address_pack_iface_attr_v2(const ucp_worker_iface_t *wiface, void *ptr,
                                unsigned atomic_flags)
 {
+    const uct_iface_attr_t *iface_attr         = &wiface->attr;
     ucp_address_v2_packed_iface_attr_t *packed = ptr;
+
     uint64_t addr_iface_flags;
-    double latency_nsec, overhead_nsec;
+    double latency_nsec, overhead_nsec, latency, bandwidth;
     size_t seg_size;
 
-    latency_nsec  = ucp_tl_iface_latency(worker->context, &iface_attr->latency) *
-                    UCS_NSEC_PER_SEC;
+    latency   = ucp_wireup_iface_lat_distance_v2(wiface);
+    bandwidth = ucp_wireup_iface_bw_distance(wiface);
+
+    latency_nsec  = latency * UCS_NSEC_PER_SEC;
     overhead_nsec = iface_attr->overhead * UCS_NSEC_PER_SEC;
 
     packed->overhead  = UCS_FP8_PACK(OVERHEAD, overhead_nsec);
-    packed->bandwidth = UCS_FP8_PACK(BANDWIDTH,
-                                     ucp_tl_iface_bandwidth(worker->context,
-                                     &iface_attr->bandwidth));
+    packed->bandwidth = UCS_FP8_PACK(BANDWIDTH, bandwidth);
     packed->latency   = UCS_FP8_PACK(LATENCY, latency_nsec);
     packed->prio      = ucs_min(UINT8_MAX, iface_attr->priority);
     addr_iface_flags  = ucp_address_flags_from_iface_flags(
@@ -748,15 +748,17 @@ ucp_address_pack_iface_attr_v2(ucp_worker_h worker, void *ptr,
     return sizeof(*packed);
 }
 
-static int ucp_address_pack_iface_attr(ucp_worker_h worker, void *ptr,
-                                       ucp_rsc_index_t rsc_index,
-                                       const uct_iface_attr_t *iface_attr,
+static int ucp_address_pack_iface_attr(const ucp_worker_iface_t *wiface,
+                                       void *ptr, ucp_rsc_index_t rsc_index,
                                        unsigned pack_flags,
                                        ucp_object_version_t addr_version,
                                        int enable_atomics)
 {
-    unsigned atomic_flags = 0;
+    const uct_iface_attr_t *iface_attr = &wiface->attr;
+    ucp_worker_h worker                = wiface->worker;
+    unsigned atomic_flags              = 0;
     unsigned packed_len;
+    double lat_ovh;
     ucp_address_unified_iface_attr_t *unified;
 
     if (ucp_worker_is_unified_mode(worker)) {
@@ -766,8 +768,11 @@ static int ucp_address_pack_iface_attr(ucp_worker_h worker, void *ptr,
          * depends on device NUMA locality. */
         unified            = ptr;
         unified->rsc_index = rsc_index;
-        unified->lat_ovh   = enable_atomics ? -iface_attr->latency.c :
-                                               iface_attr->latency.c;
+        lat_ovh            = ucp_wireup_iface_lat_distance_v1(wiface);
+        if (enable_atomics) {
+            lat_ovh = -lat_ovh;
+        }
+        unified->lat_ovh = lat_ovh;
 
         return sizeof(*unified);
     }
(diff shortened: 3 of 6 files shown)


src/ucp/core/ucp_worker.c
```diff
@@ -1304,16 +1303,16 @@ ucs_status_t ucp_worker_iface_open(ucp_worker_h worker, ucp_rsc_index_t tl_id,
...
-        status = ucp_worker_get_sys_device_distance(context, wiface->rsc_index,
-                                                    &distance);
-        if (status == UCS_OK) {
-            wiface->attr.latency.c          += distance.latency;
-            wiface->attr.bandwidth.shared    =
-                    ucs_min(wiface->attr.bandwidth.shared, distance.bandwidth);
-            wiface->attr.bandwidth.dedicated = ucs_min(
-                    wiface->attr.bandwidth.dedicated, distance.bandwidth);
-        }
+    status = ucp_worker_get_sys_device_distance(context, wiface->rsc_index,
```
yosefe, line 1306: maybe ucp_worker_get_sys_device_distance should return void. and set the value to ucs_topo_default_distance if there is an error.

src/ucp/core/ucp_worker.h
```diff
@@ -236,6 +236,7 @@ struct ucp_worker_iface {
     ucp_worker_h                  worker;        /* The parent worker */
     ucs_list_link_t               arm_list;      /* Element in arm_ifaces list */
     ucp_rsc_index_t               rsc_index;     /* Resource index */
+    ucs_sys_dev_distance_t        distance; /* Distance from given MD */
```
yosefe, line 239: align on column

src/ucp/wireup/address.c
```diff
@@ -719,21 +727,37 @@ size_t ucp_address_iface_seg_size(const uct_iface_attr_t *iface_attr)
...
+    double latency_nsec, overhead_nsec, bandwidth;
     size_t seg_size;
 
-    latency_nsec  = ucp_tl_iface_latency(worker->context, &iface_attr->latency) *
-                    UCS_NSEC_PER_SEC;
+    if (worker->context->config.ext.proto_enable) {
+        latency_nsec = ucp_tl_iface_latency_distance(worker->context,
+                                                     &iface_attr->latency,
+                                                     distance) *
+                       UCS_NSEC_PER_SEC;
```
yosefe, line 743: maybe move the multiplication by UCS_NSEC_PER_SEC outside the if/else

src/ucp/wireup/address.c
```diff
@@ -766,8 +791,12 @@ static int ucp_address_pack_iface_attr(ucp_worker_h worker, void *ptr,
          * depends on device NUMA locality. */
         unified            = ptr;
         unified->rsc_index = rsc_index;
-        unified->lat_ovh   = enable_atomics ? -iface_attr->latency.c :
-                                               iface_attr->latency.c;
+        unified->lat_ovh   = worker->context->config.ext.proto_enable ?
+                                       iface_attr->latency.c + distance->latency :
```
yosefe, line 795: add ( )

src/ucp/wireup/address.c
```diff
@@ -766,8 +791,12 @@ static int ucp_address_pack_iface_attr(ucp_worker_h worker, void *ptr,
...
+                                       iface_attr->latency.c;
+        if (enable_atomics) {
+            unified->lat_ovh = -unified->lat_ovh;
```
yosefe, line 798: use helper var instead of updating unified->lat_ovh

src/ucp/wireup/select.c
```diff
@@ -653,22 +653,38 @@ static double ucp_wireup_fp8_pack_unpack_latency(double latency)
...
 {
-    double local_lat, lat_nsec;
+    ucp_context_h context = wiface->worker->context;
+    double local_lat;
 
     if (remote_iface_attr->addr_version == UCP_OBJECT_VERSION_V1) {
+        /* With the old protocols, iface latency/bw attributes are already
+         * updated based on topology distance. That is NOT the case with the
+         * new protocols for the local iface. Therefore, we need to add the
+         * distance latency for the local iface in here.
```
yosefe, line 666: this commend seems relevant for both if and else
maybe move it before the if/else?

src/ucp/wireup/select.c
```diff
@@ -653,22 +653,38 @@ static double ucp_wireup_fp8_pack_unpack_latency(double latency)
...
+         * Remote iface latency/bw always carry distance-updated values. */
+        local_lat = context->config.ext.proto_enable ?
+                            (wiface->attr.latency.c +
+                             wiface->distance.latency) :
+                            wiface->attr.latency.c;
```
yosefe, line 671: can we extract common logic with address packing code?
same of address v2 case

src/ucp/wireup/select.c
```diff
@@ -653,22 +653,38 @@ static double ucp_wireup_fp8_pack_unpack_latency(double latency)
...
-                (iface_attr->latency.m * context->config.est_num_eps);
+        return ((local_lat + remote_iface_attr->lat_ovh) / 2) +
+               (wiface->attr.latency.m * context->config.est_num_eps);
     } else {
         /* FP8 is a lossy compression method, so in order to create a symmetric
          * calculation we pack/unpack the local latency as well */
-        lat_nsec  = ucp_tl_iface_latency(context, &iface_attr->latency) *
-                    UCS_NSEC_PER_SEC;
-        local_lat = ucp_wireup_fp8_pack_unpack_latency(lat_nsec);
+        local_lat = context->config.ext.proto_enable ?
```
yosefe, line 678: better keep using another var - lat_nsec - than updating existing var

src/ucp/wireup/select.c
```diff
@@ -1218,6 +1229,12 @@ static double ucp_tl_iface_bandwidth_ratio(ucp_context_h context,
     return ucs_max(1e-5, 1.0 - ratio);
 }
 
+static double ucp_wireup_fp8_pack_unpack_bandwidth(double bandwidth)
```
yosefe, line 1232: pls reuse code with ucp_wireup_fp8_pack_unpack_bw and put this function after ucp_wireup_fp8_pack_unpack_latency

src/ucp/wireup/select.c
```diff
@@ -1229,9 +1246,21 @@ ucp_wireup_iface_avail_bandwidth(const ucp_worker_iface_t *wiface,
...
-                                            wiface->attr.dev_num_paths);
+    local_bw = context->config.ext.proto_enable ?
+                       ucp_tl_iface_bandwidth_distance(context,
+                                                       &wiface->attr.bandwidth,
+                                                       &wiface->distance) :
+                       ucp_tl_iface_bandwidth(context, &wiface->attr.bandwidth);
+
+    local_bw *= ucp_tl_iface_bandwidth_ratio(context,
+                                             local_dev_count[dev_index],
+                                             wiface->attr.dev_num_paths);
```
yosefe, line 1257: reuse code with address pack/unpack

src/ucp/wireup/wireup.c
```diff
@@ -1803,5 +1803,44 @@ unsigned ucp_ep_init_flags(const ucp_worker_h worker,
     return flags;
 }
 
+double ucp_wireup_iface_lat_distance_v1(ucp_context_h context,
+                                        const ucs_linear_func_t *latency,
+                                        const ucs_sys_dev_distance_t *distance)
+{
+    return (context->config.ext.proto_enable) ?
```
yosefe, line 1810: remove ( )

src/ucp/wireup/wireup.c
```diff
@@ -1803,5 +1803,44 @@ unsigned ucp_ep_init_flags(const ucp_worker_h worker,
...
+}
+
+double ucp_wireup_iface_lat_distance_v2(ucp_context_h context,
+                                        const ucs_linear_func_t *latency,
+                                        const ucs_sys_dev_distance_t *distance)
+{
+    ucs_linear_func_t lat;
+
+    if (context->config.ext.proto_enable)
+    {
```
yosefe, line 1821:
```c
double tl_latency = ucp_tl_iface_latency(context, latency);

return context->config.ext.proto_enable ? (tl_latency + distance->latency) : tl_latency ;
```

src/ucp/wireup/wireup.c
```diff
@@ -1803,5 +1803,44 @@ unsigned ucp_ep_init_flags(const ucp_worker_h worker,
     return flags;
 }
 
+double ucp_wireup_iface_lat_distance_v1(ucp_context_h context,
```
yosefe, line 1806: maybe pass worker instead of context to reduce some code when this function is called?

src/ucp/wireup/wireup.c
```diff
@@ -1803,5 +1803,44 @@ unsigned ucp_ep_init_flags(const ucp_worker_h worker,
...
+                                        const ucs_linear_func_t *latency,
+                                        const ucs_sys_dev_distance_t *distance)
+{
+    return (context->config.ext.proto_enable) ?
+        (latency->c + distance->latency) : latency->c;
+}
+
+double ucp_wireup_iface_lat_distance_v2(ucp_context_h context,
```
yosefe, line 1814: maybe pass worker instead of context to reduce some code when this function is called?

src/ucp/wireup/wireup.c
```diff
@@ -1803,5 +1803,44 @@ unsigned ucp_ep_init_flags(const ucp_worker_h worker,
...
+    {
+        lat    = *latency;
+        lat.c += distance->latency;
+        return ucp_tl_iface_latency(context, &lat);
+    } else {
+        return ucp_tl_iface_latency(context, latency);
+    }
+}
+
+double ucp_wireup_iface_bw_distance(ucp_context_h context,
```
yosefe, line 1830: maybe pass worker instead of context to reduce some code when this function is called?

src/ucp/wireup/wireup.c
```diff
@@ -1803,5 +1803,44 @@ unsigned ucp_ep_init_flags(const ucp_worker_h worker,
...
+                                    const uct_ppn_bandwidth_t *bandwidth,
+                                    const ucs_sys_dev_distance_t *distance)
+{
+    uct_ppn_bandwidth_t bw;
+
+    if (context->config.ext.proto_enable) {
+        bw.shared    = ucs_min(bandwidth->shared, distance->bandwidth);
+        bw.dedicated = ucs_min(bandwidth->dedicated, distance->bandwidth);
+        return ucp_tl_iface_bandwidth(context, &bw);
```
yosefe, line 1839:
```c
double tl_bandwidth = ucp_tl_iface_bandwidth(context, bandwidth);
return context->config.ext.proto_enable ? ucs_min(tl_bandwidth, distance->bandwidth) : tl_bandwidth;
```

src/ucp/core/ucp_worker.c
```diff
@@ -1304,16 +1303,14 @@ ucs_status_t ucp_worker_iface_open(ucp_worker_h worker, ucp_rsc_index_t tl_id,
...
-            wiface->attr.latency.c          += distance.latency;
-            wiface->attr.bandwidth.shared    =
-                    ucs_min(wiface->attr.bandwidth.shared, distance.bandwidth);
-            wiface->attr.bandwidth.dedicated = ucs_min(
-                    wiface->attr.bandwidth.dedicated, distance.bandwidth);
-        }
+        wiface->attr.latency.c       += wiface->distance.latency;
+        wiface->attr.bandwidth.shared = ucs_min(wiface->attr.bandwidth.shared,
+                                                wiface->distance.bandwidth);
+        wiface->attr.bandwidth.dedicated = ucs_min(
```
brminich, line 1312: minor: can align by `=`

src/ucp/wireup/wireup.h
```diff
@@ -208,6 +208,16 @@ uct_ep_h ucp_wireup_extract_lane(ucp_ep_h ep, ucp_lane_index_t lane);
...
+double ucp_wireup_iface_lat_distance_v1(ucp_context_h context,
+                                        const ucs_linear_func_t *latency,
+                                        const ucs_sys_dev_distance_t *distance);
+double ucp_wireup_iface_lat_distance_v2(ucp_context_h context,
+                                        const ucs_linear_func_t *latency,
+                                        const ucs_sys_dev_distance_t *distance);
+double ucp_wireup_iface_bw_distance(ucp_context_h context,
+                                    const uct_ppn_bandwidth_t *bandwidth,
+                                    const ucs_sys_dev_distance_t *distance);
+
```
brminich, line 220: minor: space line between funcs

src/ucp/wireup/wireup.c
```diff
@@ -1803,5 +1803,44 @@ unsigned ucp_ep_init_flags(const ucp_worker_h worker,
...
+
+double ucp_wireup_iface_bw_distance(ucp_context_h context,
+                                    const uct_ppn_bandwidth_t *bandwidth,
+                                    const ucs_sys_dev_distance_t *distance)
+{
+    uct_ppn_bandwidth_t bw;
+
+    if (context->config.ext.proto_enable) {
+        bw.shared    = ucs_min(bandwidth->shared, distance->bandwidth);
+        bw.dedicated = ucs_min(bandwidth->dedicated, distance->bandwidth);
```
brminich, line 1838: how is it different from what is done in `ucp_worker_iface_open` for proto v1? Why need to handle it differently there then?

src/ucp/wireup/select.c
```diff
@@ -652,24 +652,36 @@ static double ucp_wireup_fp8_pack_unpack_latency(double latency)
     return UCS_FP8_UNPACK(LATENCY, packed_lat) / UCS_NSEC_PER_SEC;
 }
 
+static double ucp_wireup_fp8_pack_unpack_bw(double bandwidth)
```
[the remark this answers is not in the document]
SeyedMir: Where exactly in proto_common.c is that?

src/ucp/wireup/wireup.c
SeyedMir: ucp_worker_iface_open overwrites the latency and bandwidth of the local iface. Here we use the distance info to set the latency and bandwidth for the packed address.

src/ucp/wireup/wireup.c
SeyedMir: Will this have the same result? ucp_tl_iface_latency() applies a linear function on the latency. With proto v1, we add distance latency to latency.c and pass it to that linear function. But the above change will still use the original latency in that function.

src/ucp/wireup/wireup.c
SeyedMir: Similar to my question for latency. ucp_tl_iface_bandwidth uses dedicated and shared parts. So, the above change could  result in a different value.

src/ucp/wireup/address.c
```diff
@@ -719,21 +723,27 @@ size_t ucp_address_iface_seg_size(const uct_iface_attr_t *iface_attr)
...
-    double latency_nsec, overhead_nsec;
+    double latency_nsec, overhead_nsec, latency, bandwidth;
     size_t seg_size;
 
-    latency_nsec  = ucp_tl_iface_latency(worker->context, &iface_attr->latency) *
-                    UCS_NSEC_PER_SEC;
+    latency   = ucp_wireup_iface_lat_distance_v2(worker->context,
+                                                 &iface_attr->latency,
+                                                 distance);
+    bandwidth = ucp_wireup_iface_bw_distance(worker->context,
```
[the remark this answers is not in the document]
SeyedMir: But ucp_wireup_iface_bw_distance does not depend on address version. I mean if I pass address version, it will not be used in the function at all.

src/ucp/wireup/address.c
```diff
@@ -766,8 +768,11 @@ static int ucp_address_pack_iface_attr(ucp_worker_h worker, void *ptr,
          * depends on device NUMA locality. */
         unified            = ptr;
         unified->rsc_index = rsc_index;
-        unified->lat_ovh   = enable_atomics ? -iface_attr->latency.c :
-                                               iface_attr->latency.c;
+        lat_ovh            = ucp_wireup_iface_lat_distance_v1(wiface);
+        if (enable_atomics) {
+            lat_ovh = -lat_ovh;
+        }
+        unified->lat_ovh = lat_ovh;
```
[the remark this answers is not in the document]
SeyedMir: I used `unified->lat_ovh` directly first, but @yosefe suggested to use a helper variable instead.
`ucp_wireup_iface_lat_distance_v1` is used in other places too where we don't negate latency if `enable_atomics` is true.

brminich (dismissed):

src/ucp/wireup/address.c
```diff
@@ -766,8 +768,11 @@ static int ucp_address_pack_iface_attr(ucp_worker_h worker, void *ptr,
          * depends on device NUMA locality. */
         unified            = ptr;
         unified->rsc_index = rsc_index;
-        unified->lat_ovh   = enable_atomics ? -iface_attr->latency.c :
-                                               iface_attr->latency.c;
+        lat_ovh            = ucp_wireup_iface_lat_distance_v1(wiface);
```
yosefe, line 771: shouldn't we check v1/v2 here?

src/ucp/wireup/select.c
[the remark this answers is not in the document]
yosefe: can disregard, I looked at non upstream code

src/ucp/wireup/select.c
```diff
@@ -652,24 +652,33 @@ static double ucp_wireup_fp8_pack_unpack_latency(double latency)
...
-                (iface_attr->latency.m * context->config.est_num_eps);
+        return ((local_lat + remote_iface_attr->lat_ovh) / 2) +
+               (wiface->attr.latency.m * context->config.est_num_eps);
     } else {
+        local_lat = ucp_wireup_iface_lat_distance_v2(wiface);
         /* FP8 is a lossy compression method, so in order to create a symmetric
          * calculation we pack/unpack the local latency as well */
-        lat_nsec  = ucp_tl_iface_latency(context, &iface_attr->latency) *
+        lat_nsec  = local_lat * UCS_NSEC_PER_SEC;
+        lat_lossy = ucp_wireup_fp8_pack_unpack_latency(lat_nsec) /
```
yosefe, line 678: does it make sense to move the multiplication and division by UCS_NSEC_PER_SEC inside ucp_wireup_fp8_pack_unpack_latency function?

src/ucp/wireup/wireup.c
yosefe: so maye
```c
ucs_linear_func_t lat = wiface->attr.latency;

if (context->config.ext.proto_enable) {
    lat.c += wiface->distance.latency;
}
return ucp_tl_iface_latency(context, &lat);
```

src/ucp/wireup/address.c
SeyedMir: Do you mean checking address version to decide whether we should use `ucp_wireup_iface_lat_distance_v1` or `ucp_wireup_iface_lat_distance_v2`?

src/ucp/wireup/wireup.c
```diff
@@ -1803,5 +1803,40 @@ unsigned ucp_ep_init_flags(const ucp_worker_h worker,
...
+        wiface->attr.latency.c;
+}
+
+double ucp_wireup_iface_lat_distance_v2(const ucp_worker_iface_t *wiface)
+{
+    ucp_context_h context = wiface->worker->context;
+    ucs_linear_func_t lat = wiface->attr.latency;
+
+    if (context->config.ext.proto_enable)
+    {
```
yosefe, line 1819: { on prev line

src/ucp/wireup/address.c
yosefe: ok

ivankochin:
Shouldn't this patch also affect such functions like `ucp_ep_config_calc_params` and `ucp_ep_evaluate_perf`? Both these functions use `iface_attr->latency` without taking distance into account.

SeyedMir:
@ivankochin with proto_v1, iface latency and bandwidth have already been updated based on distance. With proto_v2, my understanding is that we do NOT want distance-updated values for those functions. @yosefe is that right?

yosefe (dismissed):
pls squash

yosefe (approved):

brminich (approved):

PR #8908, opened 2023-02-27
UCT/IB/MLX5DV: fix not use defined var error under no DEVX

changchengx:
fix below error when DEVX is not support in build environment
```
error: ‘uct_ib_mkey_token’ defined but not used [-Werror=unused-const-variable=]
 static const char uct_ib_mkey_token[] = "uct_ib_mkey_token";
                    ^~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```


yosefe:
maybe we can add UCX build test without devx?

yosefe (approved):

PR #8917, opened 2023-03-08
TEST/IODEMO: Add option to send client id

yosefe:
## Why
Test connection establishment with client_id sending. It increases the connection meta-data size and exercises additional code flows such as wireup_cm fallback.


yosefe:

brminich (approved):

PR #8921, opened 2023-03-08
GTEST: Reduce test deadline

ofirfarjun7:
## What
Reduce test timeout used when polling/waiting for flag

## Why ?
When running with Valgrind test deadline is multiplied by 20 and exceeds watchdog timeout. This can lead to a wrong report if the test fails, watchdog timeout instead of test timeout.

## How ?
Set upper limit to test deadline based on watchdog timeout.

### DIFF
--- a/test/gtest/common/test_helpers.cc
+++ b/test/gtest/common/test_helpers.cc
@@ -335,8 +335,9 @@ int test_time_multiplier()
 
 ucs_time_t get_deadline(double timeout_in_sec)
 {
-    return ucs_get_time() + ucs_time_from_sec(timeout_in_sec *
-                                              test_time_multiplier());
+    return ucs_get_time() +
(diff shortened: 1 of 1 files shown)


test/gtest/common/test_helpers.cc
```diff
@@ -24,7 +24,7 @@ namespace ucs {
 
 typedef std::pair<std::string, ::testing::TimeInMillis> test_result_t;
 
-const double test_timeout_in_sec = 60.;
+const double test_timeout_in_sec = 20.;
```
yosefe, line 27: instead, update get_deadline() so it will have upper limit that is derived from watchdog_timeout.
for example 50% or 75% of watchdog_timeout

yosefe (dismissed):
pls squash

gleon99:
@ofirfarjun7  please fix grammar mistakes in the PR description (e.g. "exceeds", "causes", etc).

yosefe (approved):

PR #8923, opened 2023-03-08
UCT/IB: use cached device name in log

changchengx:
## What
use cached device name in log

### DIFF
--- a/src/uct/ib/base/ib_device.c
+++ b/src/uct/ib/base/ib_device.c
@@ -528,8 +528,7 @@ ucs_status_t uct_ib_device_query(uct_ib_device_t *dev,
 
     if (dev->num_ports > UCT_IB_DEV_MAX_PORTS) {
         ucs_debug("%s has %d ports, but only up to %d are supported",
-                  ibv_get_device_name(ibv_device), dev->num_ports,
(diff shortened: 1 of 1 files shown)


src/uct/ib/base/ib_device.c
```diff
@@ -496,12 +496,12 @@ uct_ib_device_set_pci_id(uct_ib_device_t *dev, const char *sysfs_path)
...
 
-ucs_status_t uct_ib_device_query(uct_ib_device_t *dev,
-                                 struct ibv_device *ibv_device)
+ucs_status_t uct_ib_device_query(uct_ib_device_t *dev)
 {
-    const char *dev_name               = uct_ib_device_name(dev);
-    const char *dev_path               = dev->ibv_context->device->ibdev_path;
-    const unsigned sys_device_priority = 20;
+    const char *dev_name                = uct_ib_device_name(dev);
+    const struct ibv_device* ibv_device = dev->ibv_context->device;
```
yosefe, line 502: seems ibv_device used only one. so no need for the helper var

changchengx:
Update this PR content with using the cached device name to be output in the log

yosefe (approved):

Issue #8916, opened 2023-03-07
CUDA IPC performance regression with endpoint error handling


pentschev:
### Describe the bug
A performance regression on CUDA IPC transfers when endpoint error handling is enabled (it doesn't happen when error handling is disabled) has surged after https://github.com/openucx/ucx/pull/8769 was merged in. We see ~10GB/s at https://github.com/openucx/ucx/commit/eadd74f9fe5b0edc081ba1ce589fb850d6809934, as below:

<details>

```
$ ucx_perftest -m cuda -t tag_bw -s 100000000 -n 1000 -e localhost
[1678185484.800004] [dgx13:44878:0]        perftest.c:782  UCX  WARN  CPU affinity is not set (bound to 80 cpus). Performance may be impacted.
Accepted connection from 127.0.0.1:34556
+--------------+--------------+------------------------------+---------------------+-----------------------+
|              |              |       overhead (usec)        |   bandwidth (MB/s)  |  message rate (msg/s) |
+--------------+--------------+----------+---------+---------+----------+----------+-----------+-----------+
|    Stage     | # iterations | 50.0%ile | average | overall |  average |  overall |  average  |  overall  |
+--------------+--------------+----------+---------+---------+----------+----------+-----------+-----------+
+----------------------------------------------------------------------------------------------------------+
| API:          protocol layer                                                                             |
| Test:         tag match bandwidth                                                                        |
| Data layout:  (automatic)                                                                                |
| Send memory:  cuda                                                                                       |
| Recv memory:  cuda                                                                                       |
| Message size: 100000000                                                                                  |
+----------------------------------------------------------------------------------------------------------+
[thread 0]               129      0.406  7973.471  7973.471    11960.59   11960.59         125         125
[thread 0]               241      0.414  9185.359  8536.672    10382.55   11171.50         109         117
[thread 0]               353      0.449  9225.892  8755.348    10336.93   10892.48         108         114
[thread 0]               465      0.455  9182.758  8858.294    10385.49   10765.89         109         113
[thread 0]               577      0.471  9228.385  8930.132    10334.14   10679.29         108         112
[thread 0]               689      0.488  9183.232  8971.274    10384.95   10630.31         109         111
[thread 0]               801      0.492  9228.589  9007.253    10333.91   10587.85         108         111
[thread 0]               913      0.489  9183.249  9028.843    10384.93   10562.53         109         111
Final:                  1000      0.490 10981.080  9198.688     8684.70   10367.50          91         109
```

[...]

PR #8926, opened 2023-03-10
UCT/MD: fix missed invalidate flags

evgeny-leksikov:
## What
fix missed invalidate flags

## Why ?
fixes https://github.com/openucx/ucx/issues/8916

@pentschev FYI


yosefe (approved):

pentschev:
I can confirm this fixes the perf regression with UCX-Py as well. Thanks @evgeny-leksikov !

yosefe (approved):

PR #8935, opened 2023-03-13
UCS/MPOOL: Fix memory corruption when using UCX_MPOOL_FIFO=y

tvegas1:
## What
Fixes memory corruption as below, when using `UCX_MPOOL_FIFO=y`:
```
#7  0x00007fda667b9e66 in ucs_error_signal_handler (signo=11, info=<optimized out>, context=<optimized out>) at /labhome/tvegas/share/srcs/ucx/contrib/../src/ucs/debug/debug.c:1075
#8  <signal handler called>
#9  0x00007fda67b27e53 in ucp_rndv_am_zcopy_send_req_complete (req=0x7ffd41c46e30, status=-128) at /labhome/tvegas/share/srcs/ucx/contrib/../src/ucp/rndv/rndv.c:1895
#10 0x00007fda67acbdc4 in ucp_proto_perf_envelope_make (perf_list=<optimized out>, range_start=<optimized out>, range_end=0, convex=<optimized out>, envelope_list=0x4d430000)
    at /labhome/tvegas/share/srcs/ucx/contrib/../src/ucp/proto/proto_init.c:171
#11 0x00007fda67b32ad8 in ucp_request_complete_am_recv (status=UCS_OK, req=0x7ffd41c47500) at /labhome/tvegas/share/srcs/ucx/contrib/../src/ucp/core/ucp_request.inl:872
#12 ucp_rndv_recv_req_complete (status=UCS_OK, req=0x7ffd41c47500) at /labhome/tvegas/share/srcs/ucx/contrib/../src/ucp/rndv/rndv.c:343
#13 ucp_rndv_recv_frag_put_completion_inner (self=0x0) at /labhome/tvegas/share/srcs/ucx/contrib/../src/ucp/rndv/rndv.c:1008
```

Command to reproduce:
```
mpirun -x UCX_MPOOL_FIFO=y -x UCX_TLS=xpmem -mca pml ucx -np 64 osu_alltoall -i 1
```

## Why ?
With mpool in FIFO mode, we fail to clear `mp->data->tail` when the free list is emptied. Later, an element is released which corrupts `tail->next`, leading to a crash when doing mpool put.

## How ?
Maintain `->tail` up to date, by making sure that `->tail` is eventually reset before adding a new chunk.

### DIFF
--- a/test/gtest/ucs/test_mpool.cc
+++ b/test/gtest/ucs/test_mpool.cc
@@ -77,6 +77,30 @@ protected:
     static const size_t data_size = 152;
     static const size_t align = 128;
     static size_t leak_count;
+
+    ucs_status_t setup_mpool(ucs_mpool_t *mp, size_t elem_size,
+                             unsigned elems_per_chunk, unsigned max_elems = 0)
+    {
+        static ucs_mpool_ops_t mpool_ops = {ucs_mpool_chunk_malloc,
+                                            ucs_mpool_chunk_free, NULL, NULL,
+                                            NULL};
+        if (max_elems == 0) {
+            max_elems = elems_per_chunk;
+        }
+
+        ucs_mpool_params_t mp_params;
+
+        ucs_mpool_params_reset(&mp_params);
+        mp_params.elem_size       = header_size + elem_size;
+        mp_params.align_offset    = header_size;
+        mp_params.alignment       = align;
+        mp_params.max_chunk_size  = 4 * UCS_GBYTE;
+        mp_params.elems_per_chunk = elems_per_chunk;
+        mp_params.max_elems       = max_elems;
+        mp_params.ops             = &mpool_ops;
+        mp_params.name            = "tests";
+        return ucs_mpool_init(&mp_params, mp);
+    }
 };
 
 size_t test_mpool::leak_count;
@@ -407,21 +431,9 @@ UCS_TEST_SKIP_COND_F(test_mpool, alloc_4g, RUNNING_ON_VALGRIND) {
     const size_t elem_size         = 32 * UCS_MBYTE;
     const unsigned elems_per_chunk = ucs::limit_buffer_size(4 * UCS_GBYTE) /
                                      elem_size;
-    ucs_mpool_ops_t mpool_ops = {ucs_mpool_chunk_malloc, ucs_mpool_chunk_free,
-                                 NULL, NULL, NULL};
-    ucs_mpool_t mp;
-    ucs_mpool_params_t mp_params;
 
-    ucs_mpool_params_reset(&mp_params);
-    mp_params.elem_size       = header_size + elem_size;
-    mp_params.align_offset    = header_size;
-    mp_params.alignment       = align;
-    mp_params.max_chunk_size  = 4 * UCS_GBYTE;
-    mp_params.elems_per_chunk = elems_per_chunk;
-    mp_params.max_elems       = elems_per_chunk;
-    mp_params.ops             = &mpool_ops;
-    mp_params.name            = "tests";
-    ucs_status_t status = ucs_mpool_init(&mp_params, &mp);
+    ucs_mpool_t mp;
+    ucs_status_t status = setup_mpool(&mp, elem_size, elems_per_chunk);
     ASSERT_UCS_OK(status);
 
     // Allocate objects per one chunk size
@@ -443,4 +455,43 @@ UCS_TEST_SKIP_COND_F(test_mpool, alloc_4g, RUNNING_ON_VALGRIND) {
     }
 
     ucs_mpool_cleanup(&mp, 1);
+}
+
+class test_mpool_fifo : public test_mpool {
+public:
+    void init()
+    {
+        modify_config("MPOOL_FIFO", ucs::to_string(1).c_str());
+        test_mpool::init();
+    }
+};
+
+UCS_TEST_SKIP_COND_F(test_mpool_fifo, alloc_release, !ENABLE_DEBUG_DATA)
+{
+    const size_t elem_size         = 16;
+    const unsigned elems_per_chunk = 32;
+    const unsigned max_elems       = 128;
+    ucs_mpool_t mp;
+    std::vector<void*> objs;
+
+    ucs_status_t status = setup_mpool(&mp, elem_size, elems_per_chunk,
+                                      max_elems);
+    ASSERT_UCS_OK(status);
+
+    for (unsigned i = 0; i < 2 * elems_per_chunk; ++i) {
+        objs.push_back(ucs_mpool_get(&mp));
+    }
+
+    void *obj = ucs_mpool_get(&mp);
+    EXPECT_NE(nullptr, obj);
+    ucs_mpool_put(obj);
+
+    for (auto obj : objs) {
+        ASSERT_NE(nullptr, obj);
+        ucs_mpool_elem_t *elem = (ucs_mpool_elem_t*)obj - 1;
+        VALGRIND_MAKE_MEM_DEFINED(elem, sizeof *elem);
+        ASSERT_EQ(&mp, elem->mpool);
+    }
+
+    ucs_mpool_cleanup(&mp, 0); // skip individual put as obj could be corrupted
 }
--- a/src/ucs/datastruct/mpool.c
+++ b/src/ucs/datastruct/mpool.c
@@ -287,11 +287,7 @@ void ucs_mpool_grow(ucs_mpool_t *mp, unsigned num_elems)
         if (data->ops->obj_init != NULL) {
             data->ops->obj_init(mp, elem + 1, chunk);
         }
-
-        ucs_mpool_add_to_freelist(mp, elem, 0);
-        if (data->tail == NULL) {
-            data->tail = elem;
-        }
+        ucs_mpool_add_to_freelist(mp, elem);
     }
 
     chunk->next  = data->chunks;
--- a/src/ucs/datastruct/mpool.inl
+++ b/src/ucs/datastruct/mpool.inl
@@ -35,12 +35,12 @@ static inline void *ucs_mpool_get_inline(ucs_mpool_t *mp)
     return obj;
 }
 
-static inline void ucs_mpool_add_to_freelist(ucs_mpool_t *mp, ucs_mpool_elem_t *elem,
-                                             int add_to_tail)
+static inline void
+ucs_mpool_add_to_freelist(ucs_mpool_t *mp, ucs_mpool_elem_t *elem)
 {
     ucs_mpool_elem_t *tail;
 
-    if (add_to_tail) {
+    if (ENABLE_DEBUG_DATA && ucs_global_opts.mpool_fifo) {
         elem->next = NULL;
         if (mp->freelist == NULL) {
             mp->freelist = elem;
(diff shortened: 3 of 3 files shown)


brminich (approved):

brminich:
btw, do you think it is possible to cover by gtest?

yosefe:
I think the right place is here : https://github.com/openucx/ucx/blob/fe89ac6de9ded33b1a38b12aacdd6dbcf0ab4a2f/src/ucs/datastruct/mpool.inl#L46

tvegas1:
> btw, do you think it is possible to cover by gtest?

Added corresponding tests and confirmed that without the fix, the test fails.

test/gtest/ucs/test_mpool.cc
```diff
@@ -444,3 +454,36 @@ UCS_TEST_SKIP_COND_F(test_mpool, alloc_4g, RUNNING_ON_VALGRIND) {
...
+    const unsigned elems_per_chunk = 32;
+    const unsigned max_elems       = 128;
+    ucs_mpool_t mp;
+    std::vector<void*> objs;
+
+    ucs_status_t status = setup_mpool(&mp, elem_size, elems_per_chunk,
+                                      max_elems);
+    ASSERT_UCS_OK(status);
+
+    int old                    = ucs_global_opts.mpool_fifo;
```
brminich, line 470: pls use `modify_config()` instead

brminich (commented):
do you kow why the problem is not seen with `test_mpool, basic`?

tvegas1:
> do you kow why the problem is not seen with `test_mpool, basic`?

This can only be seen when `ENABLE_DEBUG_DATA && ucs_global_opts.mpool_fifo` is true, the freelist is emptied, a new chunk is allocated so freelist is repopulated from the front, and an obj put is performed.

yosefe:
> But when we allocate a new chunk, we insert all of them at the front, and only update tail if it was null, but it was not. So we leave a non-empty freelist, with an invalid tail, which will be overwritten at the next object put.

So maybe remove https://github.com/openucx/ucx/blob/fe89ac6de9ded33b1a38b12aacdd6dbcf0ab4a2f/src/ucs/datastruct/mpool.c#L292-L294

In addition to doing `mp->data->tail = elem` in ucs_mpool_add_to_freelist?
Then, the call to ucs_mpool_add_to_freelist called from ucs_mpool_grow would be enough to update the tail

yosefe:
BTW, the problem with NOT fixing ucs_mpool_add_to_freelist could be that if we do "mpool_put" when freelist is exactly NULL (without triggering grow) the tail may not be updated as needed

tvegas1:
> In addition to doing mp->data->tail = elem in ucs_mpool_add_to_freelist?
> Then, the call to ucs_mpool_add_to_freelist called from ucs_mpool_grow would be enough to update the tail

From ucs_mpool_grow we have the call made with _non-fifo_ mode, which skips setting tail:
- `ucs_mpool_add_to_freelist(mp, elem, ->0<-);`

> BTW, the problem with NOT fixing ucs_mpool_add_to_freelist could be that if we do "mpool_put" when freelist is exactly NULL (without triggering grow) the tail may not be updated as needed

As per my understanding that's not exact, currently, when freelist is null, we can assume that `data->tail` is either null or invalid value. We enter `if (mp->freelist == NULL)` and later properly do overwrite `mp->data->tail = elem;`, so tail cannot be used:

```
static inline void ucs_mpool_add_to_freelist(ucs_mpool_t *mp, ucs_mpool_elem_t *elem,
                                             int add_to_tail)
{
    ucs_mpool_elem_t *tail;

    if (add_to_tail) {
        elem->next = NULL;
        if (mp->freelist == NULL) {
            mp->freelist = elem; <<=====
        } else {
            tail = mp->data->tail;
            VALGRIND_MAKE_MEM_DEFINED(tail, sizeof *tail);
            tail->next = elem;
            VALGRIND_MAKE_MEM_NOACCESS(tail, sizeof *tail);
        }
        mp->data->tail = elem; <<=====
    } else {
        elem->next = mp->freelist;
        mp->freelist = elem;
    }
}
```

Based on comment, if I got it right, if you think it's cleaner, I can update the PR to change the code like:

1- change add freelist call in grow to not use `0` but the macro and `mpool_fifo` as:
  - `ucs_mpool_add_to_freelist(mp, elem, ENABLE_DEBUG_DATA && ucs_global_opts.mpool_fifo);`

2- remove existing block in grow, in the for loop:
  - `if (data->tail == NULL) { data->tail = elem; }`

test/gtest/ucs/test_mpool.cc
tvegas1: added push/modify/pop config and tested with/without fix

yosefe:
> 1- change add freelist call in grow to not use `0` but the macro and `mpool_fifo` as:
> 
> * `ucs_mpool_add_to_freelist(mp, elem, ENABLE_DEBUG_DATA && ucs_global_opts.mpool_fifo);`

Maybe even remove "add_to_tail" param and let mpool_add_to_freelist check "ENABLE_DEBUG_DATA && ucs_global_opts.mpool_fifo" directly

tvegas1:
> > 1- change add freelist call in grow to not use `0` but the macro and `mpool_fifo` as:
> > 
> > * `ucs_mpool_add_to_freelist(mp, elem, ENABLE_DEBUG_DATA && ucs_global_opts.mpool_fifo);`
> 
> Maybe even remove "add_to_tail" param and let mpool_add_to_freelist check "ENABLE_DEBUG_DATA && ucs_global_opts.mpool_fifo" directly

Pushed it

tvegas1:
> do you kow why the problem is not seen with `test_mpool, basic`?

To besure, I analysed the test basic, and confirmed that because we always start the put on empty list, the issue does not repro. Changing the loop count from `18` to `17` triggers crash, as expected.

test/gtest/ucs/test_mpool.cc
```diff
@@ -444,3 +454,36 @@ UCS_TEST_SKIP_COND_F(test_mpool, alloc_4g, RUNNING_ON_VALGRIND) {
...
+
+    void *obj = ucs_mpool_get(&mp);
+    EXPECT_TRUE(obj != NULL);
+    ucs_mpool_put(obj);
+
+    for (auto obj : objs) {
+        EXPECT_TRUE(obj != NULL);
+        ucs_mpool_elem_t *elem = (ucs_mpool_elem_t*)obj - 1;
+        VALGRIND_MAKE_MEM_DEFINED(elem, sizeof *elem);
+        EXPECT_TRUE(&mp == elem->mpool);
```
yosefe, line 485: EXPECT_EQ

test/gtest/ucs/test_mpool.cc
```diff
@@ -444,3 +454,36 @@ UCS_TEST_SKIP_COND_F(test_mpool, alloc_4g, RUNNING_ON_VALGRIND) {
...
+    ucs_status_t status = setup_mpool(&mp, elem_size, elems_per_chunk,
+                                      max_elems);
+    ASSERT_UCS_OK(status);
+
+    for (unsigned i = 0; i < 2 * elems_per_chunk; ++i) {
+        objs.push_back(ucs_mpool_get(&mp));
+    }
+
+    void *obj = ucs_mpool_get(&mp);
+    EXPECT_TRUE(obj != NULL);
```
yosefe, line 478: EXPECT_NE

test/gtest/ucs/test_mpool.cc
```diff
@@ -444,3 +454,36 @@ UCS_TEST_SKIP_COND_F(test_mpool, alloc_4g, RUNNING_ON_VALGRIND) {
...
+    ucs_mpool_put(obj);
+
+    for (auto obj : objs) {
+        EXPECT_TRUE(obj != NULL);
+        ucs_mpool_elem_t *elem = (ucs_mpool_elem_t*)obj - 1;
+        VALGRIND_MAKE_MEM_DEFINED(elem, sizeof *elem);
+        EXPECT_TRUE(&mp == elem->mpool);
+    }
+    ucs_mpool_cleanup(&mp, 0); // skip individual put as obj could be corrupted
```
yosefe, line 487: space line before

test/gtest/ucs/test_mpool.cc
```diff
@@ -77,6 +77,28 @@ class test_mpool : public ucs::test {
...
+
+    ucs_status_t setup_mpool(ucs_mpool_t *mp, size_t elem_size,
+                             unsigned elems_per_chunk, unsigned max_elems = 0)
+    {
+        ucs_mpool_ops_t mpool_ops = {ucs_mpool_chunk_malloc,
+                                     ucs_mpool_chunk_free, NULL, NULL, NULL};
+        if (!max_elems) {
+            max_elems = elems_per_chunk;
+        }
+        ucs_mpool_params_t mp_params;
```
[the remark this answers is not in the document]
tvegas1: fixed

test/gtest/ucs/test_mpool.cc
```diff
@@ -444,3 +454,36 @@ UCS_TEST_SKIP_COND_F(test_mpool, alloc_4g, RUNNING_ON_VALGRIND) {
...
+    const unsigned max_elems       = 128;
+    ucs_mpool_t mp;
+    std::vector<void*> objs;
+
+    push_config();
+    modify_config("MPOOL_FIFO", ucs::to_string(1).c_str());
+
+    ucs_status_t status = setup_mpool(&mp, elem_size, elems_per_chunk,
+                                      max_elems);
+    ASSERT_UCS_OK(status);
```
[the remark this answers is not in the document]
tvegas1: added new test fixture

test/gtest/ucs/test_mpool.cc
tvegas1: fixed

test/gtest/ucs/test_mpool.cc
tvegas1: fixed

test/gtest/ucs/test_mpool.cc
```diff
@@ -77,6 +77,30 @@ class test_mpool : public ucs::test {
...
     static const size_t align = 128;
     static size_t leak_count;
+
+    ucs_status_t setup_mpool(ucs_mpool_t *mp, size_t elem_size,
+                             unsigned elems_per_chunk, unsigned max_elems = 0)
+    {
+        static ucs_mpool_ops_t mpool_ops = {ucs_mpool_chunk_malloc,
+                                            ucs_mpool_chunk_free, NULL, NULL,
+                                            NULL};
+        if (!max_elems) {
```
brminich, line 87: minor: coding style `if (max_elems != 0)`, as `max_elems` is a number not a boolean

brminich (dismissed):
👍, besides minor comment

yosefe (dismissed):

test/gtest/ucs/test_mpool.cc
tvegas1: fixed

brminich (approved):

test/gtest/ucs/test_mpool.cc
tvegas1: fixed

brminich:
@tvegas1, can you please squash the commits?

brminich (approved):

PR #8939, opened 2023-03-14
UCP/PROTO: Fixed operation attributes packing to select params.

rakhmets:
## What
Fixed operation attributes packing to select parameters.

## Why ?
Only flags in `UCP_PROTO_SELECT_OP_ATTR_MASK` should affect protocol selection.
Otherwise, we may have a cache miss, and print duplicating protocol information.

### DIFF
--- a/src/ucp/proto/proto_select.inl
+++ b/src/ucp/proto/proto_select.inl
@@ -62,14 +62,14 @@ ucp_proto_select_op_attr_pack(uint32_t op_attr_mask)
     UCS_STATIC_ASSERT(
             (UCP_PROTO_SELECT_OP_ATTR_MASK / UCP_PROTO_SELECT_OP_ATTR_BASE) <
             UCP_PROTO_SELECT_OP_FLAGS_BASE);
-    return op_attr_mask / UCP_PROTO_SELECT_OP_ATTR_BASE;
+    return (op_attr_mask & UCP_PROTO_SELECT_OP_ATTR_MASK) /
+           UCP_PROTO_SELECT_OP_ATTR_BASE;
 }
 
 static UCS_F_ALWAYS_INLINE uint32_t
 ucp_proto_select_op_attr_unpack(uint8_t op_attr)
 {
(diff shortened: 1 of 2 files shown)


src/ucp/proto/proto_select.inl
```diff
@@ -62,7 +62,8 @@ ucp_proto_select_op_attr_pack(uint32_t op_attr_mask)
     UCS_STATIC_ASSERT(
             (UCP_PROTO_SELECT_OP_ATTR_MASK / UCP_PROTO_SELECT_OP_ATTR_BASE) <
             UCP_PROTO_SELECT_OP_FLAGS_BASE);
-    return op_attr_mask / UCP_PROTO_SELECT_OP_ATTR_BASE;
+    return (op_attr_mask & UCP_PROTO_SELECT_OP_ATTR_MASK) /
```
yosefe, line 65: why needed - can't we just ignore unused flags?

rakhmets: This is a fix for fast proto selection: https://github.com/openucx/ucx/blob/fe89ac6de9ded33b1a38b12aacdd6dbcf0ab4a2f/src/ucp/proto/proto_select.inl#L109

yosefe: you mean we have cache miss because of irrelevant flags in op_attr_mask? so it's a performance issue?
can you pls update PR desc accordingly?

rakhmets: yes.
e.g. we send with `UCP_OP_ATTR_FLAG_NO_IMM_CMPL`, then `select_param` will reflect the flag. And protocol selection will go through the slow path.
Along with that we print the same configuration twice in case of `UCX_PROTO_INFO=y`.
Thus, it is a performance issue, and duplication it protocol information output.

rakhmets: P.s. With these changes there is no need to apply mask in `ucp_proto_select_op_attr_unpack`.
I will update the PR.

brminich (dismissed):

yosefe (approved):

yosefe (approved):

PR #7644, opened 2021-11-11
UCP/PROTO/AM: Add AM short and bcopy protocols v2

brminich:
## What
- Define UCP AM short and bcopy protocols for new proto infrastructure
- Add new proto test variant for just a couple of tests for now until zcopy and reply protocols are added

PR #5850, opened 2020-10-28
UCP/RMA/TAG: Support fast-path short with new protocols

yosefe:

PR #8941, opened 2023-03-14
UCP/PROTO: Fix short threshold calculation

ofirfarjun7:
## What
Fix short threshold calculation.

## Why ?
Short init threshold function didn't consider opt flags when calculating the short threshold, that cause send operation to enter short fast path, when the optimized protocol was different than short.

## How ?
Pass `UCP_OP_ATTR_FLAG_MULTI_SEND` optimization flag so the selection of the threshold will also consider short threshold of multi operation.

### DIFF
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -749,7 +749,9 @@ typedef enum {
     UCP_OP_ATTR_FLAG_MULTI_SEND     = UCS_BIT(19)  /**< optimize for bandwidth of
                                                         multiple in-flight operations,
                                                         rather than for the latency
-                                                        of a single operation */
+                                                        of a single operation.
+                                                        This flag and UCP_OP_ATTR_FLAG_FAST_CMPL
+                                                        are mutually exclusive. */
 } ucp_op_attr_t;
 
 
--- a/src/ucp/core/ucp_request.inl
+++ b/src/ucp/core/ucp_request.inl
@@ -177,10 +177,21 @@ UCS_PTR_MAP_IMPL(request, 0);
 
 
 #define UCP_REQUEST_CHECK_PARAM(_param) \
-    if (((_param)->op_attr_mask & UCP_OP_ATTR_FIELD_MEMORY_TYPE) && \
-        ((_param)->memory_type > UCS_MEMORY_TYPE_LAST)) { \
-        ucs_error("invalid memory type parameter: %d", (_param)->memory_type); \
-        return UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM); \
+    if (ENABLE_PARAMS_CHECK) { \
+        if (((_param)->op_attr_mask & UCP_OP_ATTR_FIELD_MEMORY_TYPE) && \
+            ((_param)->memory_type > UCS_MEMORY_TYPE_LAST)) { \
+            ucs_error("invalid memory type parameter: %d", \
+                      (_param)->memory_type); \
+            return UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM); \
+        } \
+        \
+        if (ucs_test_all_flags((_param)->op_attr_mask, \
+                               (UCP_OP_ATTR_FLAG_FAST_CMPL | \
+                                UCP_OP_ATTR_FLAG_MULTI_SEND))) { \
+            ucs_error("UCP_OP_ATTR_FLAG_FAST_CMPL and " \
+                      "UCP_OP_ATTR_FLAG_MULTI_SEND are mutually exclusive"); \
+            return UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM); \
+        } \
     }
 
 
--- a/src/ucp/core/ucp_worker.c
+++ b/src/ucp/core/ucp_worker.c
@@ -1938,13 +1938,13 @@ ucp_worker_ep_config_short_init(ucp_worker_h worker, ucp_ep_config_t *ep_config,
     if (worker->context->config.features & feature_flag) {
         ucp_proto_select_short_init(worker, &ep_config->proto_select,
                                     ep_cfg_index, UCP_WORKER_CFG_INDEX_NULL,
-                                    op_id, 0, proto_flags, &proto_short);
+                                    op_id, proto_flags, &proto_short);
 
-         /* Short protocol should be either disabled, or use expected lane */
-         ucs_assertv((proto_short.max_length_host_mem < 0) ||
-                     (proto_short.lane == exp_lane),
-                     "max_length_host_mem %ld, lane %d",
-                     proto_short.max_length_host_mem, proto_short.lane);
+        /* Short protocol should be either disabled, or use expected lane */
+        ucs_assertv((proto_short.max_length_host_mem < 0) ||
+                            (proto_short.lane == exp_lane),
+                    "max_length_host_mem %ld, lane %d",
+                    proto_short.max_length_host_mem, proto_short.lane);
     } else {
         ucp_proto_select_short_disable(&proto_short);
     }
@@ -2104,8 +2104,7 @@ ucp_worker_add_rkey_config(ucp_worker_h worker,
     if (worker->context->config.features & UCP_FEATURE_RMA) {
         ucp_proto_select_short_init(worker, &rkey_config->proto_select,
                                     key->ep_cfg_index, rkey_cfg_index,
-                                    UCP_OP_ID_PUT, UCP_OP_ATTR_FLAG_FAST_CMPL,
-                                    UCP_PROTO_FLAG_PUT_SHORT,
+                                    UCP_OP_ID_PUT, UCP_PROTO_FLAG_PUT_SHORT,
                                     &rkey_config->put_short);
     } else {
         ucp_proto_select_short_disable(&rkey_config->put_short);
--- a/src/ucp/proto/proto_select.c
+++ b/src/ucp/proto/proto_select.c
@@ -689,34 +689,35 @@ void ucp_proto_select_short_disable(ucp_proto_select_short_t *proto_short)
     proto_short->rkey_index             = UCP_NULL_RESOURCE;
 }
 
-void
-ucp_proto_select_short_init(ucp_worker_h worker, ucp_proto_select_t *proto_select,
-                            ucp_worker_cfg_index_t ep_cfg_index,
-                            ucp_worker_cfg_index_t rkey_cfg_index,
-                            ucp_operation_id_t op_id, uint32_t op_attr_mask,
-                            unsigned proto_flags,
-                            ucp_proto_select_short_t *proto_short)
+void ucp_proto_select_short_init(ucp_worker_h worker,
+                                 ucp_proto_select_t *proto_select,
+                                 ucp_worker_cfg_index_t ep_cfg_index,
+                                 ucp_worker_cfg_index_t rkey_cfg_index,
+                                 ucp_operation_id_t op_id, unsigned proto_flags,
+                                 ucp_proto_select_short_t *proto_short)
 {
-    ucp_context_h context    = worker->context;
-    const ucp_proto_t *proto = NULL;
+    static const uint32_t op_attributes[] = {0, UCP_OP_ATTR_FLAG_FAST_CMPL,
+                                             UCP_OP_ATTR_FLAG_MULTI_SEND};
+    ucp_context_h context                 = worker->context;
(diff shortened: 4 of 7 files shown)


yosefe (dismissed):

rakhmets (commented):
Should the flags in line 2108 also be updated?

gleon99:
@ofirfarjun7  following @rakhmets  remark,

can simply add `| UCP_OP_ATTR_FLAG_MULTI_SEND` inside `ucp_proto_select_short_init`.

src/ucp/core/ucp_worker.c
```diff
@@ -1938,13 +1938,14 @@ ucp_worker_ep_config_short_init(ucp_worker_h worker, ucp_ep_config_t *ep_config,
...
         ucp_proto_select_short_init(worker, &ep_config->proto_select,
                                     ep_cfg_index, UCP_WORKER_CFG_INDEX_NULL,
-                                    op_id, 0, proto_flags, &proto_short);
-
-         /* Short protocol should be either disabled, or use expected lane */
-         ucs_assertv((proto_short.max_length_host_mem < 0) ||
-                     (proto_short.lane == exp_lane),
-                     "max_length_host_mem %ld, lane %d",
-                     proto_short.max_length_host_mem, proto_short.lane);
+                                    op_id, UCP_OP_ATTR_FLAG_MULTI_SEND,
```
brminich, line 1941: do we also need `UCP_OP_ATTR_FLAG_FAST_CMPL` here?

src/ucp/proto/proto_select.c
```diff
@@ -689,34 +689,34 @@ void ucp_proto_select_short_disable(ucp_proto_select_short_t *proto_short)
...
-     * in both regular mode and UCP_OP_ATTR_FLAG_FAST_CMPL mode.
+     * Find the minimal threshold for operation 'op_id' among all protocols
+     * with attribute from 'op_attributes'. Fast-path short protocol
+     * can be used only if the message size fits this minimal threshold.
      */
     ucs_log_indent(1);
-    ucs_for_each_submask(op_attr, op_attr_mask) {
-        ucp_proto_select_param_init(&select_param, op_id, op_attr, 0,
+    for (i = 0; i < ucs_static_array_size(op_attributes); ++i) {
+        ucp_proto_select_param_init(&select_param, op_id, op_attributes[i], 0,
```
rakhmets, line 719: You check the threshold only when the flags are used individually.
Is there a use case when we don't want to check `UCP_OP_ATTR_FLAG_FAST_CMPL | UCP_OP_ATTR_FLAG_MULTI_SEND`?

And for me, it isn't still clear, why we previously (or originally) pass `UCP_OP_ATTR_FLAG_FAST_CMPL` only for put operation (in this PR https://github.com/openucx/ucx/pull/5850).

Maybe there was a reason for this. And it become clear, what combination of flags we should use during short protocol threshold calculation.

ofirfarjun7: I'm not sure if there is a use case. I think it's more related to meaning of each opt attribute, one prefers latency the other bandwidth. In that case I'm not sure why we need to check `UCP_OP_ATTR_FLAG_FAST_CMPL` if we check `UCP_OP_ATTR_FLAG_MULTI_SEND`. Seems to me that if you optimize bandwidth over latency short threshold will always be smaller, but it's just an assumption.

I'm not sure why we pass `UCP_OP_ATTR_FLAG_FAST_CMPL` only for put operation, but here are some basic facts:
- `UCP_OP_ATTR_FLAG_MULTI_SEND` was added after `UCP_OP_ATTR_FLAG_FAST_CMPL`
- The call to `ucp_proto_select_short_init` for AM/TAG was added after the call for PUT

So maybe it's just a bug, but lets wait for others to reply.

yosefe (dismissed):

src/ucp/proto/proto_select.c
rakhmets: `UCP_OP_ATTR_FLAG_MULTI_SEND` is for optimizing protocol selection in favor of bandwidth. However, I don't agree that `UCP_OP_ATTR_FLAG_FAST_CMPL` has the opposite meaning, and intendent for optimizing protocol selection in favor of latency. Otherwise, there is no need to have both flags if they have opposite effect. Thus, I don't see reasons, why we cannot use both flags in the same time for the protocol selection, and calculation of thresholds in particular.

Agree that `UCP_OP_ATTR_FLAG_MULTI_SEND` was added after `UCP_OP_ATTR_FLAG_FAST_CMPL`. But the next fact is not completely correct. The call to `ucp_proto_select_short_init` only for AM (https://github.com/openucx/ucx/pull/7644) was added after the call for TAG and PUT (https://github.com/openucx/ucx/pull/5850).

@yosefe do you remember, why did you pass `UCP_OP_ATTR_FLAG_FAST_CMPL` only in case of short put in this PR https://github.com/openucx/ucx/pull/5850?

```C
        /* Set threshold for short send */
        ucp_proto_select_short_init(worker, &ep_config->proto_select,
                                    ep_cfg_index, UCP_WORKER_CFG_INDEX_NULL,
                                    UCP_OP_ID_TAG_SEND, 0,
                                    UCP_PROTO_FLAG_AM_SHORT, &tag_short);
...
       /* Set threshold for short put */
        ucp_proto_select_short_init(worker, &rkey_config->proto_select,
                                    key->ep_cfg_index, rkey_cfg_index,
                                    UCP_OP_ID_PUT, UCP_OP_ATTR_FLAG_FAST_CMPL,
                                    UCP_PROTO_FLAG_PUT_SHORT,
                                    &rkey_config->put_short);
```

yosefe: FAST_CMPL usually implies blocking operation, which contradicts MULTI_SEND.

> @yosefe do you remember, why did you pass UCP_OP_ATTR_FLAG_FAST_CMPL only in case of short put in this PR https://github.com/openucx/ucx/pull/5850?

It was wrong, i didn't consider the fast-path threshold would be too high.
IMO the right thing is for fast-path threshold to be the minimum of all valid flag combinations.

gleon99:
@ofirfarjun7 please squash.

src/ucp/proto/proto_select.c
ofirfarjun7: Yossi what exactly do you mean by "valid flag"?
Is it a combination you consider as valid or any flag combination that user can pass to UCX?

yosefe: A combination that "makes sense". IMO we can document (and validate with ENABLE_PARAM_CHECK) that UCP_OP_ATTR_FLAG_FAST_CMPL is not used along with UCP_OP_ATTR_FLAG_MULTI_SEND

ofirfarjun7: I think we need to document it and validate it at least with ENABLE_PARAM_CHECK so it will be well defined.
I think it also bothered Raul that we treat this combination as not Valid but we are not validating it or document it.

WDIT @rakhmets ?

rakhmets: The changes are good for me in this case.

rakhmets (approved):

src/ucp/core/ucp_request.inl
```diff
@@ -176,12 +176,23 @@ UCS_PTR_MAP_IMPL(request, 0);
     }
 
 
+#define UCP_REQUEST_CHECK_OPT_FLAGS(_param) \
+    if (ENABLE_PARAMS_CHECK && \
+        ucs_test_all_flags((_param)->op_attr_mask, \
+                           (UCP_OP_ATTR_FLAG_FAST_CMPL | \
```
ofirfarjun7, line 182: We can use `UCP_PROTO_SELECT_OP_ATTR_MASK` for the combination of the flags, but I thought that maybe in the future this MACRO can hold another flag that can be combined with other opt flags.

src/ucp/core/ucp_request.inl
```diff
@@ -176,12 +176,23 @@ UCS_PTR_MAP_IMPL(request, 0);
...
+                            UCP_OP_ATTR_FLAG_MULTI_SEND))) { \
+        ucs_error("invalid optimization flags combination: %u", \
```
rakhmets, line 184: We should return `UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM)` in case of error. (if this was a warning, we would continue).

src/ucp/core/ucp_request.inl
```diff
@@ -176,12 +176,23 @@ UCS_PTR_MAP_IMPL(request, 0);
     }
 
 
+#define UCP_REQUEST_CHECK_OPT_FLAGS(_param) \
```
rakhmets, line 179:
```suggestion
#define UCP_REQUEST_CHECK_OP_ATTR_FLAGS(_param) \
```

src/ucp/api/ucp.h
```diff
@@ -741,15 +741,19 @@ typedef enum {
...
                                                         operation, fail if the
                                                         operation cannot be
                                                         completed immediately */
     UCP_OP_ATTR_FLAG_MULTI_SEND     = UCS_BIT(19)  /**< optimize for bandwidth of
                                                         multiple in-flight operations,
                                                         rather than for the latency
-                                                        of a single operation */
+                                                        of a single operation.
+                                                        This flag and UCP_OP_ATTR_FLAG_FAST_CMPL
+                                                        are mutually exclusive. */
```
ofirfarjun7, line 756: Can document it in the description of `ucp_op_attr_t` instead for each flag.
WDYT?

yosefe: add it to MULTI_SEND only. No need to duplicate.

src/ucp/core/ucp_request.inl
```diff
@@ -176,12 +176,24 @@ UCS_PTR_MAP_IMPL(request, 0);
     }
 
 
+#define UCP_REQUEST_CHECK_OP_ATTR_FLAGS(_param) \
```
yosefe, line 179: why need separate macro?

src/ucp/core/ucp_request.inl
```diff
@@ -176,12 +176,24 @@ UCS_PTR_MAP_IMPL(request, 0);
...
+    if (ENABLE_PARAMS_CHECK && \
+        ucs_test_all_flags((_param)->op_attr_mask, \
+                           (UCP_OP_ATTR_FLAG_FAST_CMPL | \
+                            UCP_OP_ATTR_FLAG_MULTI_SEND))) { \
+        ucs_error("invalid optimization flags combination: %u", \
+                  (_param)->op_attr_mask); \
```
yosefe, line 185: Why make the user guess?

`ucs_error("UCP_OP_ATTR_FLAG_FAST_CMPL and UCP_OP_ATTR_FLAG_MULTI_SEND are mutually exclusive");`

src/ucp/proto/proto_select.c
```diff
@@ -689,34 +689,34 @@ void ucp_proto_select_short_disable(ucp_proto_select_short_t *proto_short)
...
-     * use fast-path short protocol only if the message size fits short protocol
-     * in both regular mode and UCP_OP_ATTR_FLAG_FAST_CMPL mode.
+     * Find the minimal threshold for operation 'op_id' among all protocols
+     * with attribute from 'op_attributes'. Fast-path short protocol
+     * can be used only if the message size fits this minimal threshold.
      */
     ucs_log_indent(1);
-    ucs_for_each_submask(op_attr, op_attr_mask) {
-        ucp_proto_select_param_init(&select_param, op_id, op_attr, 0,
+    for (i = 0; i < ucs_static_array_size(op_attributes); ++i) {
```
yosefe, line 718: ucs_carray_for_each

src/ucp/core/ucp_request.inl
ofirfarjun7: The `UCP_REQUEST_CHECK_PARAM` macro is not used by put operation. i'm not sure if it's a mistake or not.
From what I see it uses `ucs_assert` to validate the memory type when calling `ucp_request_get_memory_type`.

You think we should change that?

yosefe: it's a mistake. we need to use UCP_REQUEST_CHECK_PARAM also for put/get/atomic operations

src/ucp/proto/proto_select.c
ofirfarjun7: Will do.

We have many candidates for refactoring in the code...
*search `for i < ucs_static_array_size(`

yosefe (dismissed):

gleon99 (approved):

PR #8945, opened 2023-03-15
UCT/BASE: Added base implementation to is_reachable_v2 API using intra/inter flag

shasson5:
## What
Added base implementation to is_reachable_v2 API using intra/inter flag.

## Why ?
Support identifying intra/inter node communications.

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -2110,8 +2110,7 @@ static ucs_status_t ucp_wireup_select_set_locality_flags(
     ucp_rsc_index_t rsc_index;
     ucp_address_entry_t *ae;
     ucp_lane_index_t lane;
-    ucs_status_t status;
-    void *dev_addr;
+    uct_iface_is_reachable_params_t params;
 
     if (select_params->address->uuid == worker->uuid) {
         key->flags |= UCP_EP_CONFIG_KEY_FLAG_SELF;
@@ -2142,29 +2141,25 @@ static ucs_status_t ucp_wireup_select_set_locality_flags(
             continue;
         }
 
-        dev_addr = ucs_malloc(wiface->attr.device_addr_len, "ucp_tmp_dev_addr");
-        if (dev_addr == NULL) {
-            ucs_error("failed to allocated device address of size %zu",
-                      wiface->attr.device_addr_len);
-            return UCS_ERR_NO_MEMORY;
-        }
-
-        status = uct_iface_get_device_address(wiface->iface, dev_addr);
-        if (status != UCS_OK) {
-            ucs_free(dev_addr);
-            return status;
-        }
-
         ucp_unpacked_address_for_each(ae, select_params->address) {
-            if ((wiface->attr.device_addr_len == ae->dev_addr_len) &&
-                !memcmp(ae->dev_addr, dev_addr, ae->dev_addr_len)) {
+            if ((wiface->attr.device_addr_len != ae->dev_addr_len) ||
+                (worker->context->tl_rscs[rsc_index].tl_name_csum !=
+                 ae->tl_name_csum)) {
+                continue;
+            }
+
+            params.field_mask  = UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR |
+                                 UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR  |
+                                 UCT_IFACE_IS_REACHABLE_FIELD_SCOPE;
+            params.device_addr = ae->dev_addr;
+            params.iface_addr  = ae->iface_addr;
+            params.scope       = UCT_IFACE_REACHABILITY_SCOPE_DEVICE;
+
+            if (uct_iface_is_reachable_v2(wiface->iface, &params)) {
                 key->flags |= UCP_EP_CONFIG_KEY_FLAG_INTRA_NODE;
-                ucs_free(dev_addr);
                 return UCS_OK;
             }
         }
-
-        ucs_free(dev_addr);
     }
 
     return UCS_OK;
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -254,7 +254,8 @@ typedef enum {
     UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR        = UCS_BIT(0), /**< device_addr field */
     UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR         = UCS_BIT(1), /**< iface_addr field */
     UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING        = UCS_BIT(2), /**< info_string field */
-    UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING_LENGTH = UCS_BIT(3)  /**< info_string_length field */
+    UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING_LENGTH = UCS_BIT(3), /**< info_string_length field */
+    UCT_IFACE_IS_REACHABLE_FIELD_SCOPE              = UCS_BIT(4) /**<  scope field */
 } uct_iface_is_reachable_field_mask_t;
 
 
@@ -517,6 +518,20 @@ extern const char *uct_ep_operation_names[];
 
 
 /**
+ * @ingroup UCS_RESOURCE
+ * @brief Reachability scope
+ *
+ * Reachability scope. Can be one of the following values:
+ * Device scope: Checks if addresses describe the same device.
+ * Network scope: Checks reachability between different devices.
+ */
+typedef enum {
+    UCT_IFACE_REACHABILITY_SCOPE_DEVICE, /**< Local device scope */
+    UCT_IFACE_REACHABILITY_SCOPE_NETWORK /**< Network scope */
+} uct_iface_reachability_scope_t;
+
+
+/**
  * @ingroup UCT_RESOURCE
  * @brief Operation parameters passed to @ref uct_iface_is_reachable_v2.
  */
@@ -527,19 +542,19 @@ typedef struct uct_iface_is_reachable_params {
      * mask will be ignored. Provides ABI compatibility with respect to adding
      * new fields.
      */
-    uint64_t                     field_mask;
+    uint64_t                       field_mask;
 
     /**
      * Device address to check for reachability.
      * This field must not be passed if iface_attr.dev_addr_len == 0.
      */
-    const uct_device_addr_t      *device_addr;
+    const uct_device_addr_t       *device_addr;
 
     /**
      * Interface address to check for reachability.
      * This field must not be passed if iface_attr.iface_addr_len == 0.
      */
-    const uct_iface_addr_t       *iface_addr;
+    const uct_iface_addr_t        *iface_addr;
 
     /**
      * User-provided pointer to a string buffer.
@@ -547,13 +562,18 @@ typedef struct uct_iface_is_reachable_params {
      * null-terminated information string explaining why the remote address is
      * not reachable if the return value is 0.
      */
-    char                         *info_string;
+    char                          *info_string;
 
     /**
      * The length of the @a info_string is provided in bytes.
      * This value must be specified in conjunction with @a info_string.
      */
-    size_t                        info_string_length;
+    size_t                         info_string_length;
+
+    /**
+     * Reachability scope.
+     */
+    uct_iface_reachability_scope_t scope;
 } uct_iface_is_reachable_params_t;
 
 
--- a/src/uct/base/uct_iface.c
+++ b/src/uct/base/uct_iface.c
@@ -219,11 +219,61 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
     return iface->ops.iface_is_reachable(iface, dev_addr, iface_addr);
 }
 
-int uct_iface_is_reachable_v2(const uct_iface_h iface,
+static int uct_iface_is_same_device(const uct_iface_h iface,
+                                    const uct_device_addr_t *device_addr)
+{
+    void *dev_addr;
+    uct_iface_attr_t attr;
+    ucs_status_t status;
+
+    status = uct_iface_query(iface, &attr);
+    if (status != UCS_OK) {
+        ucs_error("failed to query iface %p", iface);
+        return 0;
+    }
+
+    dev_addr = ucs_alloca(attr.device_addr_len);
+    status   = uct_iface_get_device_address(iface, dev_addr);
+    if (status != UCS_OK) {
+        ucs_error("failed to get device address from %p", iface);
+        return 0;
+    }
+
+    return !memcmp(device_addr, dev_addr, attr.device_addr_len);
+}
+
+int
+uct_base_iface_is_reachable_v2(const uct_iface_h iface,
+                               const uct_iface_is_reachable_params_t *params)
+{
+    uct_iface_reachability_scope_t scope;
+
+    if (!ucs_test_all_flags(params->field_mask,
+                            UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR |
+                            UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR)) {
+        ucs_error("missing params (field_mask: %lu), both device_addr and "
+                  "iface_addr should be supplied.", params->field_mask);
+        return 0;
+    }
+
+    if (params->field_mask & UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING) {
+        params->info_string[0] = '\0';
+    }
+
+    scope = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params, scope, SCOPE,
+                            UCT_IFACE_REACHABILITY_SCOPE_NETWORK);
+
+    return (scope == UCT_IFACE_REACHABILITY_SCOPE_NETWORK) ?
+                   uct_iface_is_reachable(iface, params->device_addr,
+                                          params->iface_addr) :
+                   uct_iface_is_same_device(iface, params->device_addr);
+}
+
+int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
                               const uct_iface_is_reachable_params_t *params)
 {
-    ucs_fatal("uct_iface_is_reachable_v2 not supported yet");
-    return 0;
+    const uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
+    return iface->internal_ops->iface_is_reachable_v2(tl_iface, params);
 }
 
 ucs_status_t uct_ep_check(const uct_ep_h ep, unsigned flags,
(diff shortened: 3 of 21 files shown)


src/uct/api/v2/uct_v2.h
```diff
@@ -554,6 +554,13 @@ typedef struct uct_iface_is_reachable_params {
      * This value must be specified in conjunction with @a info_string.
      */
     size_t                        info_string_length;
+
+    /**
+     * Output parameter; Indicates whether both addresses describe the same
+     * device.
+     */
+    int                           is_self;
```
brminich, line 562: do we need to update `uct_iface_is_reachable_field_mask_t`?

src/ucp/wireup/select.c
```diff
@@ -2142,29 +2141,19 @@ static ucs_status_t ucp_wireup_select_set_locality_flags(
...
-
         ucp_unpacked_address_for_each(ae, select_params->address) {
-            if ((wiface->attr.device_addr_len == ae->dev_addr_len) &&
-                !memcmp(ae->dev_addr, dev_addr, ae->dev_addr_len)) {
+            if (wiface->attr.device_addr_len != ae->dev_addr_len) {
+                continue;
+            }
+
+            params.device_addr = ae->dev_addr;
+            params.iface_addr  = ae->iface_addr;
```
brminich, line 2150: we need to set `field_mask`

src/uct/base/uct_iface.c
```diff
@@ -456,12 +456,51 @@ uct_base_iface_estimate_perf(uct_iface_h iface, uct_perf_attr_t *perf_attr)
...
+    void *dev_addr;
+    ucs_status_t status;
+    uct_iface_attr_t attr;
+
+    status = uct_iface_query(iface, &attr);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    dev_addr = ucs_malloc(attr.device_addr_len, "ucp_tmp_dev_addr");
```
brminich, line 471: `ucp` -> `uct`

src/uct/base/uct_iface.c
```diff
@@ -456,12 +456,51 @@ uct_base_iface_estimate_perf(uct_iface_h iface, uct_perf_attr_t *perf_attr)
...
+int uct_base_iface_is_reachable_v2(const uct_iface_h iface,
+                                   uct_iface_is_reachable_params_t *params)
+{
+    void *dev_addr;
+    ucs_status_t status;
+    uct_iface_attr_t attr;
+
+    status = uct_iface_query(iface, &attr);
+    if (status != UCS_OK) {
+        return status;
```
brminich, line 468: note this function returns int

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -388,11 +388,12 @@ static ucs_mpool_ops_t uct_cuda_copy_event_desc_mpool_ops = {
...
-    .iface_vfs_refresh   = (uct_iface_vfs_refresh_func_t)ucs_empty_function,
-    .ep_query            = (uct_ep_query_func_t)ucs_empty_function_return_unsupported,
-    .ep_invalidate       = (uct_ep_invalidate_func_t)ucs_empty_function_return_unsupported,
-    .ep_connect_to_ep_v2 = ucs_empty_function_return_unsupported
+    .iface_estimate_perf   = uct_cuda_copy_estimate_perf,
+    .iface_vfs_refresh     = (uct_iface_vfs_refresh_func_t)ucs_empty_function,
+    .ep_query              = (uct_ep_query_func_t)ucs_empty_function_return_unsupported,
+    .ep_invalidate         = (uct_ep_invalidate_func_t)ucs_empty_function_return_unsupported,
+    .ep_connect_to_ep_v2   = ucs_empty_function_return_unsupported,
+    .iface_is_reachable_v2 = uct_base_iface_is_reachable_v2,
```
brminich, line 396: can we just call `uct_base_iface_is_reachable_v2` from `uct_iface_is_reachable_v2` and do not update every transport?

src/uct/api/v2/uct_v2.h
shasson5: is it relevant for output params?

brminich: i'd expect so

src/uct/api/v2/uct_v2.h
```diff
@@ -554,6 +554,13 @@ typedef struct uct_iface_is_reachable_params {
...
+
```
brminich, line 563: extra line

src/ucp/wireup/select.c
```diff
@@ -2142,29 +2141,21 @@ static ucs_status_t ucp_wireup_select_set_locality_flags(
...
-        status = uct_iface_get_device_address(wiface->iface, dev_addr);
-        if (status != UCS_OK) {
-            ucs_free(dev_addr);
-            return status;
-        }
-
         ucp_unpacked_address_for_each(ae, select_params->address) {
-            if ((wiface->attr.device_addr_len == ae->dev_addr_len) &&
-                !memcmp(ae->dev_addr, dev_addr, ae->dev_addr_len)) {
+            if (wiface->attr.device_addr_len != ae->dev_addr_len) {
```
brminich, line 2145: maybe better to do this check in `uct_base_iface_is_reachable_v2` itself? Otherwise it may wrongly initialize `is_self` if two addresses have different length, but their content is partially the same

shasson5: Currently `is_reachable_v2` does not have addr_len argument, so we can't perform this comparison there.

src/uct/api/v2/uct_v2.h
```diff
@@ -254,7 +254,8 @@ typedef enum {
     UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR        = UCS_BIT(0), /**< device_addr field */
     UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR         = UCS_BIT(1), /**< iface_addr field */
     UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING        = UCS_BIT(2), /**< info_string field */
-    UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING_LENGTH = UCS_BIT(3)  /**< info_string_length field */
+    UCT_IFACE_IS_REACHABLE_FIELD_IS_SELF            = UCS_BIT(3), /**< is_self field */
+    UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING_LENGTH = UCS_BIT(4)  /**< info_string_length field */
```
gleon99, line 258: Any reason `STRING_LENGTH` moved below? Is that intentional?

src/uct/api/v2/uct_v2.h
gleon99: IMO using the word `self` is problematic, as it causes confusion with the `self` transport.
Maybe `is_same_device` instead?

src/uct/base/uct_iface.c
```diff
@@ -219,11 +219,59 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
     return iface->ops.iface_is_reachable(iface, dev_addr, iface_addr);
 }
 
-int uct_iface_is_reachable_v2(const uct_iface_h iface,
-                              const uct_iface_is_reachable_params_t *params)
+static int
+uct_base_iface_is_reachable_v2(const uct_iface_h iface,
```
gleon99, line 223: Why is that in an API PR?

src/uct/api/v2/uct_v2.h
```diff
@@ -554,6 +555,12 @@ typedef struct uct_iface_is_reachable_params {
      * This value must be specified in conjunction with @a info_string.
      */
     size_t                        info_string_length;
+
+    /**
+     * Output parameter; Indicates whether both addresses describe the same
+     * device.
+     */
+    int                           is_same_device;
```
gleon99, line 563: Please change as we discussed ("reachability scope").

src/uct/api/v2/uct_v2.h
```diff
@@ -516,6 +517,16 @@ typedef struct uct_md_mem_attach_params {
 extern const char *uct_ep_operation_names[];
 
 
+/**
+ * @ingroup UCS_RESOURCE
+ * @brief Reachability scope
+ */
+typedef enum uct_iface_reachability_scope {
+    UCT_IFACE_REACHABILITY_INTRA_NODE, /**< Intra node scope */
+    UCT_IFACE_REACHABILITY_INTER_NODE, /**< Inter node scope */
```
yosefe, line 526: what if it's same node but different device?

src/uct/api/v2/uct_v2.h
```diff
@@ -527,33 +538,41 @@ typedef struct uct_iface_is_reachable_params {
...
      * This value must be specified in conjunction with @a info_string.
      */
-    size_t                        info_string_length;
+    size_t                         info_string_length;
+
+    /**
+     * Reachability scope. Can be one of the following values:
+     * Intra-node: Checks if addresses describe the same device.
+     * Inter-node: Checks reachability between different nodes.
+     * Default is inter-node.
```
yosefe, line 573: 1. intra_node / inter_node is not a representative name, need something with the word "Device"
2. document these in uct_iface_reachability_scope_t enum

src/uct/api/v2/uct_v2.h
```diff
@@ -903,14 +922,14 @@ ucs_status_t uct_ep_query(uct_ep_h ep, uct_ep_attr_t *ep_attr);
  * check is a local operation it does not detect issues such as network
  * mis-configuration or lack of connectivity.
  *
- * @param [in]  iface       Local interface to check reachability from.
- * @param [in]  params      Operation parameters, see @ref
- *                          uct_iface_is_reachable_params_t.
+ * @param [in]    iface   Local interface to check reachability from.
+ * @param [inout] params  Operation parameters, see @ref
```
yosefe, line 926: why changed?

src/uct/api/v2/uct_v2.h
```diff
@@ -903,14 +922,14 @@ ucs_status_t uct_ep_query(uct_ep_h ep, uct_ep_attr_t *ep_attr);
...
+ *                        uct_iface_is_reachable_params_t.
  *
  * @return Nonzero if reachable, 0 if not.
  */
 int uct_iface_is_reachable_v2(uct_iface_h iface,
-                              const uct_iface_is_reachable_params_t *params);
+                              uct_iface_is_reachable_params_t *params);
```
yosefe, line 932: why needed?

src/uct/base/uct_iface.c
```diff
@@ -219,11 +219,67 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
     return iface->ops.iface_is_reachable(iface, dev_addr, iface_addr);
 }
 
-int uct_iface_is_reachable_v2(const uct_iface_h iface,
-                              const uct_iface_is_reachable_params_t *params)
+static int
+uct_iface_is_intra_node_reachable(const uct_iface_h iface,
```
yosefe, line 223: uct_iface_is_same_device

src/uct/base/uct_iface.c
```diff
@@ -219,11 +219,67 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
...
+    uct_iface_attr_t attr;
+    ucs_status_t status;
+    int reachable;
+
+    status = uct_iface_query(iface, &attr);
+    if (status != UCS_OK) {
+        return 0;
+    }
+
+    dev_addr = ucs_malloc(attr.device_addr_len, "uct_tmp_dev_addr");
```
yosefe, line 236: maybe use ucs_alloca?

src/uct/base/uct_iface.c
```diff
@@ -219,11 +219,67 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
...
+        return 0;
+    }
+
+    status = uct_iface_get_device_address(iface, dev_addr);
+    if (status != UCS_OK) {
+        ucs_free(dev_addr);
+        return 0;
+    }
+
+    reachable = !memcmp(params->device_addr, dev_addr, attr.device_addr_len);
```
yosefe, line 249: are we sure params->device_addr is valid? check field_mask

src/uct/base/uct_iface.c
```diff
@@ -219,11 +219,67 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
     return iface->ops.iface_is_reachable(iface, dev_addr, iface_addr);
 }
 
-int uct_iface_is_reachable_v2(const uct_iface_h iface,
-                              const uct_iface_is_reachable_params_t *params)
+static int
+uct_iface_is_intra_node_reachable(const uct_iface_h iface,
+                                  const uct_iface_is_reachable_params_t *params)
```
yosefe, line 224: pass const void *device_address instead of params

src/uct/base/uct_iface.c
```diff
@@ -219,11 +219,67 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
...
+                               uct_iface_is_reachable_params_t *params)
+{
+    params->info_string_length = 0;
+
+    if (params->field_mask & UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING) {
+        params->info_string[0] = '\0';
+    }
+
+    if ((params->field_mask & UCT_IFACE_IS_REACHABLE_FIELD_SCOPE) &&
+        (params->scope == UCT_IFACE_REACHABILITY_INTRA_NODE)) {
```
yosefe, line 265: maybe use UCS_PARAM_VALUE macro

src/uct/base/uct_iface.c
```diff
@@ -219,11 +219,67 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
...
+    return uct_iface_is_reachable(iface, params->device_addr,
+                                  params->iface_addr);
+}
+
+int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
+                              uct_iface_is_reachable_params_t *params)
+{
+    uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
+
+    if (iface->internal_ops->iface_is_reachable_v2 == NULL) {
```
yosefe, line 278: all transports should define uct_base_iface_is_reachable_v2 function pointer instead of NULL

src/uct/api/v2/uct_v2.h
```diff
@@ -516,6 +517,16 @@ typedef struct uct_md_mem_attach_params {
 extern const char *uct_ep_operation_names[];
 
 
+/**
+ * @ingroup UCS_RESOURCE
+ * @brief Reachability scope
+ */
+typedef enum uct_iface_reachability_scope {
```
gleon99, line 524: `typedef enum {`

src/uct/api/v2/uct_v2.h
gleon99, line 526: `REACHABILITY_SCOPE_` (like the `enum` name`).

gleon99: Remove trailing `,`

src/uct/api/v2/uct_v2.h
gleon99: @yosefe  `params` can't be const anymore -> the arg is `inout`.

yosefe: why should it be inout?

gleon99: Because `params` is not `const` anymore?

src/uct/api/v2/uct_v2.h
shasson5: As we already changed the API several times, maybe should discuss this issue @yosefe , @gleon99, a
and define API which is acceptible for all

src/uct/api/v2/uct_v2.h
shasson5: see above comment

src/uct/base/uct_iface.c
shasson5: can you explain why is it better?

src/uct/api/v2/uct_v2.h
yosefe: talked w @gleon99 , the param can remain const/in since the scope is input and reachable indication is function return value

src/uct/api/v2/uct_v2.h
```diff
@@ -520,10 +520,14 @@ extern const char *uct_ep_operation_names[];
 /**
  * @ingroup UCS_RESOURCE
  * @brief Reachability scope
+ *
+ * Reachability scope. Can be one of the following values:
+ * Device scope: Checks if addresses describe the same device.
+ * Network scope: Checks reachability between different nodes.
```
yosefe, line 526: different nodes -> different devices

src/uct/api/v2/uct_v2.h
yosefe: unresolved

src/uct/base/uct_iface.c
```diff
@@ -219,66 +219,60 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
...
-    dev_addr = ucs_malloc(attr.device_addr_len, "uct_tmp_dev_addr");
-    if (dev_addr == NULL) {
-        ucs_error("failed to allocated device address of size %zu",
-                  attr.device_addr_len);
-        return 0;
-    }
-
+    dev_addr = ucs_alloca(attr.device_addr_len);
     status = uct_iface_get_device_address(iface, dev_addr);
+
```
yosefe, line 237: remove

src/uct/base/uct_iface.c
```diff
@@ -219,66 +219,60 @@ int uct_iface_is_reachable(const uct_iface_h iface, const uct_device_addr_t *dev
...
-uct_base_iface_is_reachable_v2(const uct_iface_h iface,
-                               uct_iface_is_reachable_params_t *params)
+int uct_base_iface_is_reachable_v2(
+        const uct_iface_h iface, const uct_iface_is_reachable_params_t *params)
 {
-    params->info_string_length = 0;
+    uct_iface_reachability_scope_t scope;
+
+    if (!(params->field_mask & UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR) ||
+        !(params->field_mask & UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR)) {
```
yosefe, line 252: ucs_test_all_flags

src/uct/api/v2/uct_v2.h
shasson5:
```

    /**
     * User-provided pointer to a string buffer.
     * The function @ref uct_iface_is_reachable_v2 fills this buffer with a
     * null-terminated information string explaining why the remote address is
     * not reachable if the return value is 0.
     */
    char                          *info_string;
```

src/uct/api/v2/uct_v2.h
```diff
@@ -527,33 +542,38 @@ typedef struct uct_iface_is_reachable_params {
...
 
     /**
      * The length of the @a info_string is provided in bytes.
      * This value must be specified in conjunction with @a info_string.
      */
-    size_t                        info_string_length;
+    size_t                         info_string_length;
+
+    /**
+     * Reachability scope. Default is network scope.
```
gleon99, line 574: IMO it is incorrect to mention any "default" here.
The struct only defines how data fields are stored to form something meaningful, it doesn't have any notion of "default".

The "default" relates to the behavior of a (particular) function.

src/uct/base/uct_iface.h
```diff
@@ -255,13 +255,20 @@ typedef ucs_status_t (*uct_ep_connect_to_ep_v2_func_t)(
...
-    uct_iface_vfs_refresh_func_t   iface_vfs_refresh;
-    uct_ep_query_func_t            ep_query;
-    uct_ep_invalidate_func_t       ep_invalidate;
-    uct_ep_connect_to_ep_v2_func_t ep_connect_to_ep_v2;
+    uct_iface_estimate_perf_func_t   iface_estimate_perf;
+    uct_iface_vfs_refresh_func_t     iface_vfs_refresh;
+    uct_ep_query_func_t              ep_query;
+    uct_ep_invalidate_func_t         ep_invalidate;
+    uct_ep_connect_to_ep_v2_func_t   ep_connect_to_ep_v2;
+    uct_iface_is_reachable_v2_func_t iface_is_reachable_v2;
```
gleon99, line 271: Need to add `is_reachable_v2` to `uct_base_iface_internal_ops`?

yosefe (approved):

yosefe (approved):

PR #8955, opened 2023-03-19
AZP: Enable advance debugging in running Docker containers

ofirfarjun7:
## What
Enable advance debugging capabilities when running docker containers.

## Why ?
Debug issues in running containers.

## How ?
Allow tracing by passing SYS_PTRACE when running the containers.

### DIFF
--- a/buildlib/pr/main.yml
+++ b/buildlib/pr/main.yml
@@ -2,155 +2,156 @@ variables:
   DOCKER_OPT_VOLUMES: -v /hpc/local:/hpc/local -v /auto/sw_tools:/auto/sw_tools
   DOCKER_OPT_IB: --ulimit memlock=-1:-1 --device=/dev/infiniband/
   DOCKER_OPT_GPU: --gpus all $(DOCKER_OPT_IB)
+  DOCKER_OPT_ARGS: --cap-add=SYS_PTRACE
 
 resources:
   containers:
     - container: centos7
       image: rdmz-harbor.rdmz.labs.mlnx/ucx/centos7:4
-      options: $(DOCKER_OPT_VOLUMES)
+      options: $(DOCKER_OPT_ARGS) $(DOCKER_OPT_VOLUMES)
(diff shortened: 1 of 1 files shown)


yosefe (dismissed):

gleon99 (approved):

ofirfarjun7:
Need to stash last commit for now and edit test_jenkins so it will check if strace installed instead of if ptrace installed.

```
test_ucs_load() {
	if [ -z "${have_ptrace}" ] **<--- Change to have_ptrace**
	then
		log_warning "==== Not running UCS library loading test ===="
		return
	fi
```

contrib/test_jenkins.sh
```diff
@@ -73,6 +73,7 @@ MAKEP="make -j${parallel_jobs}"
 export AUTOMAKE_JOBS=$parallel_jobs
 
 have_ptrace=$(capsh --print | grep 'Bounding' | grep ptrace || true)
+have_strace=$(strace -V || true)
```
ofirfarjun7, line 76: My only concern is differences between versions of strace.
For example in the MAN the option --version exists but on strace running on Jazz it's not exists.
I used `-V ` because it's mentioned in the MAN and also available on Jazz.

We can do a different check if you have different idea, for example
`whereis strace | grep \/ || true` (will not check if strace running but if the path exists it's probably avaialble.)

gleon99: Do you have any specific reason to believe these will be some setup where `-V` is unavailable?

ofirfarjun7: I didn't encounter with such setup or MAN that not include the -V option. This is why I checked it this way.
I just mentioned that for you to consider.

gleon99: IMO the concern regarding the potential absence of the `-V` flag is not so reasonable, so let's stick with it for now.

yosefe (approved):

PR #8959, opened 2023-03-20
UCP/CORE: Wake up UCT iface on send completion event.

rakhmets:
## What
Added send completion event to list of events on which the UCT interface should wake up.

## Why ?
The example of the possible issue is in RC wire up procedure. RC uses UD to establish connection. However, the last message in wire up messaging is acknowledge reply, which is sent using RC. If RC interface's progress isn't enabled, and events are being checked, the RC interface can wake up only on send completion event.

### DIFF
--- a/src/ucp/core/ucp_worker.c
+++ b/src/ucp/core/ucp_worker.c
@@ -620,8 +620,13 @@ static ucs_status_t ucp_worker_iface_check_events_do(ucp_worker_iface_t *wiface,
         return UCS_OK;
     } else if (*progress_count == 0) {
         /* Arm the interface to wait for next event */
-        ucs_assert(wiface->attr.cap.event_flags & UCT_IFACE_FLAG_EVENT_RECV);
-        status = uct_iface_event_arm(wiface->iface, UCT_EVENT_RECV);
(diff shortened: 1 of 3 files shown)


src/ucp/core/ucp_worker.c
```diff
@@ -621,7 +621,10 @@ static ucs_status_t ucp_worker_iface_check_events_do(ucp_worker_iface_t *wiface,
     } else if (*progress_count == 0) {
         /* Arm the interface to wait for next event */
         ucs_assert(wiface->attr.cap.event_flags & UCT_IFACE_FLAG_EVENT_RECV);
-        status = uct_iface_event_arm(wiface->iface, UCT_EVENT_RECV);
+        ucs_assert(wiface->attr.cap.event_flags &
+                   UCT_IFACE_FLAG_EVENT_SEND_COMP);
```
brminich, line 625: maybe use a single assert**v** including both flags?

rakhmets: Updated assertv, along with check for iface flags.

brminich (dismissed):

yosefe (approved):

brminich (approved):

PR #8963, opened 2023-03-21
UCT/ROCM: move memh caches to iface

edgargabriel:
## What
move the memhandle caches from the endpoint to the iface structures. 

## Why ?
Improves scalability and reduces memory footprint

### DIFF
--- a/src/uct/rocm/copy/rocm_copy_iface.c
+++ b/src/uct/rocm/copy/rocm_copy_iface.c
@@ -235,6 +235,7 @@ static UCS_CLASS_INIT_FUNC(uct_rocm_copy_iface_t, uct_md_h md, uct_worker_h work
                                                           uct_rocm_copy_iface_config_t);
     ucs_status_t status;
     ucs_mpool_params_t mp_params;
+    char target_name[64];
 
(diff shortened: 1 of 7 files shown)


src/uct/rocm/copy/rocm_copy_iface.c
```diff
@@ -261,6 +262,13 @@ static UCS_CLASS_INIT_FUNC(uct_rocm_copy_iface_t, uct_md_h md, uct_worker_h work
 
     ucs_queue_head_init(&self->signal_queue);
 
+    snprintf(target_name, sizeof(target_name), "dest:%ld", self->id);
```
yosefe, line 265: ucs_snprintf_safe

src/uct/rocm/ipc/rocm_ipc_iface.c
```diff
@@ -180,6 +182,15 @@ static UCS_CLASS_INIT_FUNC(uct_rocm_ipc_iface_t, uct_md_h md, uct_worker_h worke
 
     ucs_queue_head_init(&self->signal_queue);
 
+    iface_pid = getpid();
+    snprintf(target_name, sizeof(target_name), "dest:%d", iface_pid);
```
yosefe, line 186: ucs_snprintf_safe
can remove iface_pid var

yosefe (approved):

PR #8966, opened 2023-03-22
UCT/IB/RC/ACCEL: Pre-posted receive buffers during initialization.

rakhmets:
## What
Pre-posted receive buffers during initialization of accelerated RC interface.

## Why ?
To be able to receive messages before enabling progress.

## How ?
Only limited number of receive buffers are posted during initialization to avoid growing of memory consumption during interface creation. The rest of buffers are created during progress enabling.


src/uct/ib/rc/accel/rc_mlx5_common.c
```diff
@@ -206,7 +206,7 @@ unsigned uct_rc_mlx5_iface_srq_post_recv_ll(uct_rc_mlx5_iface_common_t *iface)
     return count;
 }
 
-void uct_rc_mlx5_iface_common_prepost_recvs(uct_rc_mlx5_iface_common_t *iface)
+void uct_rc_mlx5_common_prepost_recvs(uct_rc_mlx5_iface_common_t *iface)
```
brminich, line 209: imo it is ok to leave the name as it is, because preposting itself happens on `iface`, not `ep`

rakhmets: Yes, but file name is usually a prefix, that is why I changed the name. And the involvement of the interface is reflected in the parameters.

rakhmets: There are the following options:
1. leave it as-is (`uct_rc_mlx5_iface_common_prepost_recvs(iface)`), because many function names started with `uct_rc_mlx5_iface` in this header file.
2. change name to `uct_rc_mlx5_common_prepost_recvs_on_iface(iface)`, or `uct_rc_mlx5_common_iface_prepost_recvs(iface)`, to have `iface` in the name and to use more suitable prefix.
3. use `uct_rc_mlx5_common_prepost_recvs(iface)`, because I think there is a word duplication in the second item.

I'd prefer the third. @brminich would you still vote for the first?

brminich: according to our naming convention, we use the object as a prefix which is  `uct_rc_mlx5_iface`, so i'd not change the name. 
the second option does not look consistent with the rest of the code.
3-d is better than the 2d, but the 1st one is still preferrable for me (also because it's better not to change just the name to avoid messing up commits history)

src/uct/ib/rc/accel/rc_mlx5_ep.c
```diff
@@ -1051,6 +1051,7 @@ UCS_CLASS_INIT_FUNC(uct_rc_mlx5_ep_t, const uct_ep_params_t *params)
     self->tx.wq.bb_max = ucs_min(self->tx.wq.bb_max, iface->tx.bb_max);
     self->mp.free      = 1;
     uct_rc_txqp_available_set(&self->super.txqp, self->tx.wq.bb_max);
+    uct_rc_mlx5_iface_common_prepost_recvs(iface);
```
brminich, line 1054: this should be relevant for verbs as well

src/uct/ib/rc/verbs/rc_verbs_ep.c
```diff
@@ -673,7 +673,10 @@ UCS_CLASS_INIT_FUNC(uct_rc_verbs_ep_t, const uct_ep_params_t *params)
     uct_rc_txqp_available_set(&self->super.txqp, iface->config.tx_max_wr);
     uct_rc_verbs_txcnt_init(&self->txcnt);
     uct_ib_fence_info_init(&self->fi);
-    uct_rc_verbs_iface_common_prepost_recvs(iface, UINT_MAX);
+    status = uct_rc_verbs_iface_common_prepost_recvs(iface);
+    if (status != UCS_OK) {
+        goto err_qp_cleanup;
+    }
```
brminich, line 679: better to do it on line 671, otherwise the cleanup flow is not fully correct - need also remove that qp from iface array of qps (added on line 672)

rakhmets: Fixed.

src/uct/ib/rc/verbs/rc_verbs_ep.c
```diff
@@ -669,6 +669,11 @@ UCS_CLASS_INIT_FUNC(uct_rc_verbs_ep_t, const uct_ep_params_t *params)
         goto err_qp_cleanup;
     }
 
+    status = uct_rc_verbs_iface_common_prepost_recvs(iface);
```
brminich, line 672: just noticed that even in this case we have to do some additional cleanup - `uct_ib_device_async_event_unregister`

brminich (dismissed):

src/uct/ib/rc/verbs/rc_verbs_ep.c
```diff
@@ -659,24 +659,40 @@ UCS_CLASS_INIT_FUNC(uct_rc_verbs_ep_t, const uct_ep_params_t *params)
...
 
-err_qp_cleanup:
+
+err_remove_qp:
+    uct_rc_iface_remove_qp(&iface->super, self->qp->qp_num);
+err_event_unreg:
+    uct_ib_device_async_event_unregister(&md->dev,
+                                         IBV_EVENT_QP_LAST_WQE_REACHED,
+                                         self->qp->qp_num);
+err_destroy_qp:
```
yosefe, line 695: 👍

yosefe (approved):

yosefe:
@rakhmets pls squash

brminich (approved):

PR #8895, opened 2023-02-19
AZP/RELEASE: Add ARM release - Part #2

Al3xR01:
What
Build and release ARM-based artifacts.
Part 2: Build DEB/RPM artifacts on ARM.

Why ?
https://github.com/openucx/ucx/issues/8762

PR #8970, opened 2023-03-27
UCT/UCP: Fix compilation errors with armcc-22.1

brminich:
## What
Fix compilation errors with armclang-22.1

### DIFF
--- a/src/ucp/wireup/ep_match.c
+++ b/src/ucp/wireup/ep_match.c
@@ -88,7 +88,7 @@ ucp_ep_h ucp_ep_match_retrieve(ucp_worker_h worker, uint64_t dest_uuid,
                                ucp_ep_match_conn_sn_t conn_sn,
                                ucs_conn_match_queue_type_t conn_queue_type)
 {
-    ucp_ep_flags_t exp_ep_flags = UCP_EP_FLAG_ON_MATCH_CTX;
+    ucp_ep_flags_t UCS_V_UNUSED exp_ep_flags = UCP_EP_FLAG_ON_MATCH_CTX;
     ucs_conn_match_elem_t *conn_match;
     ucp_ep_h ep;
 
--- a/src/uct/ib/base/ib_verbs.h
+++ b/src/uct/ib/base/ib_verbs.h
@@ -71,6 +71,9 @@ static inline ucs_status_t uct_ib_query_device(struct ibv_context *ctx,
 #  define IBV_ACCESS_RELAXED_ORDERING               0
 #endif
 
+#if !HAVE_DECL_IBV_ACCESS_ON_DEMAND
+#  define IBV_ACCESS_ON_DEMAND                      0
+#endif
 
 /*
  * DC support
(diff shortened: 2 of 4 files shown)


rakhmets (dismissed):

src/ucp/wireup/ep_match.c
```diff
@@ -88,7 +88,7 @@ ucp_ep_h ucp_ep_match_retrieve(ucp_worker_h worker, uint64_t dest_uuid,
                                ucp_ep_match_conn_sn_t conn_sn,
                                ucs_conn_match_queue_type_t conn_queue_type)
 {
-    ucp_ep_flags_t exp_ep_flags = UCP_EP_FLAG_ON_MATCH_CTX;
+    ucp_ep_flags_t UCS_V_UNUSED exp_ep_flags = UCP_EP_FLAG_ON_MATCH_CTX;
```
yosefe, line 91: replace the usage of exp_ep_flags by the flag

src/uct/ib/base/ib_verbs.h
```diff
@@ -71,6 +71,9 @@ static inline ucs_status_t uct_ib_query_device(struct ibv_context *ctx,
 #  define IBV_ACCESS_RELAXED_ORDERING               0
 #endif
 
+#if !HAVE_DECL_IBV_ACCESS_ON_DEMAND
+#  define IBV_ACCESS_ON_DEMAND                      0
+#endif
```
yosefe, line 76: why needed? can't we remove it from the code?

yosefe (commented):
can we add arm22 to ci?

src/uct/ib/base/ib_verbs.h
brminich: Can you pls elaborate?
Avoiding compilation failure with arm clang on specific config

src/ucp/wireup/ep_match.c
brminich: you mean rename to `flags`? (note we may add a couple of flags to this var, not just one)

src/uct/ib/base/ib_verbs.h
yosefe: AFAIR, we removed IBV_ACCESS_ON_DEMAND because it's supposed to be unused. Why do we need to return it?

yosefe: right..
so, maybe we need to expose from UCT that ODP is disabled if IBV_ACCESS_ON_DEMAND==0?
even if it's supported by FW

src/ucp/wireup/ep_match.c
yosefe: I mean remove this variable at all. and use UCP_EP_FLAG_ON_MATCH_CTX constant instead of it

brminich: we also add `UCP_EP_FLAG_REMOTE_ID` to the flags var below

yosefe: ok

artemry-nv:
> can we add arm22 to ci?

@Al3xR01 FYI

yosefe (dismissed):
pls squash

Al3xR01:
>

Al3xR01:
We can add arm22 to ci once we add arm platform support.

yosefe (approved):

Al3xR01:
arm22 in CI: addressed by PR #8895
(https://github.com/openucx/ucx/pull/8895/files#diff-940fa8ef89b586de941ddb47d8566f84686adbcfe592e0d2ab28ef96d7bf4f73)

PR #8982, opened 2023-03-29
AZP: retry module load when automount fails

Al3xR01:
## What
Workaround for automount availability issues while loading modules.

## Why ?
Multiple job failures due to the on and off availability of automount shares.

## How ?
Attempt to remount by accessing the NFS.

### DIFF
--- a/buildlib/az-helpers.sh
+++ b/buildlib/az-helpers.sh
@@ -105,21 +105,30 @@ function az_init_modules() {
 
 #
 # Test if an environment module exists and load it if yes.
+# Retry 5 times in case of automount failure.
 # Otherwise, return error code.
 #
 function az_module_load() {
     module=$1
+    retries=5
 
-    if module avail -t 2>&1 | grep -q "^$module\$"
-    then
-        module load $module
-        return 0
-    else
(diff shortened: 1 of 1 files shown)


buildlib/az-helpers.sh
```diff
@@ -105,21 +105,30 @@ function az_init_modules() {
...
-        azure_log_warning "Module $module cannot be loaded"
-        return 1
-    fi
+    retries=5
+
+    until module avail -t 2>&1 | grep -q "^$module\$"; do
+        if [ $retries -gt 1 ]; then
+            # Attempt to refresh automount
+            echo "Module $module not found, retrying..."
+            ls /hpc/local > /dev/null 2>&1
```
artemry-nv, line 119: As far as I understand it's to trigger autofs to re-read /hpc/local content, isn't it?
I think touching just /hpc/local isn't enough because it's an existing mount point.
But touching the content (e.g. find /hpc/local/...) may trigger autofs refresh the file tree.

artemry-nv (approved):

yosefe (approved):

PR #8990, opened 2023-04-02
GTEST: Skip test while debugging CI Hang bug

ofirfarjun7:
## What
Skip AM header copy test.

## Why ?
AM header copy test is causing CI to hang from time to time, need to find the root cause and solve it. 
We will skip the test until we release a fix for the problem.

## How ?
Add skip with reason to the test.

### DIFF
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -1069,7 +1069,8 @@ private:
  * except when running with Valgrind, because its very time consuming.
  */
 UCS_TEST_SKIP_COND_P(test_ucp_am_nbx_send_copy_header, all_protos,
-                     (has_transport("self") && RUNNING_ON_VALGRIND),
+                     /* FIXME: Disabled due to unresolved failure - CI Hang */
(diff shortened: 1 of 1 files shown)


gleon99:
@ofirfarjun7 please fix commit title (see pipeline).

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1078,6 +1078,8 @@ UCS_TEST_SKIP_COND_P(test_ucp_am_nbx_send_copy_header, all_protos,
     size_t header_length;
     size_t data_length;
 
+    UCS_TEST_SKIP_R("Skip while fixing CI hang bug.");
```
gleon99, line 1081: `FIXME: Disabled due to unresolved failure`

ofirfarjun7: `FIXME` is a convention we use in this situations?

gleon99: Kind of (search in repo).

yosefe (approved):

yosefe:
@ofirfarjun7 pls squash

PR #8993, opened 2023-04-03
UCT/MLX5: Only pack flush rkey when flush remote is enabled

tvegas1:
## What
Do not add the `flush_rkey` to the packed address when RMA and AMO32/64 are not there.

## Why ?
The flush rkey is making address too big for `rdmacm`.

## How ?
Use the existing flush remote flag to disable remote flush at interface instantiation. Handle it at `rc_mlx5` interface level as other transports are using the same memory domain.

### Tested
perftest `ucp_get` still contains the `flush_rkey`, io_demo skips it:
```
UCX_SOCKADDR_TLS_PRIORITY=rdmacm ./test/apps/iodemo/io_demo -C 12 -P 1000000 &
UCX_SOCKADDR_TLS_PRIORITY=rdmacm libtool --mode=execute  ./test/apps/iodemo/io_demo -C 12 -P 1000000 <ip address>
```

### DIFF
--- a/src/uct/ib/rc/accel/rc_mlx5_common.h
+++ b/src/uct/ib/rc/accel/rc_mlx5_common.h
@@ -721,4 +721,13 @@ uct_rc_mlx5_common_iface_init_rx(uct_rc_mlx5_iface_common_t *iface,
 
 void uct_rc_mlx5_destroy_srq(uct_ib_mlx5_md_t *md, uct_ib_mlx5_srq_t *srq);
 
+static UCS_F_ALWAYS_INLINE int
+uct_rc_mlx5_iface_flush_rkey_enabled(uct_rc_mlx5_iface_common_t *iface)
+{
+    uct_ib_md_t *md = uct_ib_iface_md(&iface->super.super);
+
+    return iface->super.config.flush_remote &&
+           uct_ib_md_is_flush_rkey_valid(md->flush_rkey);
+}
+
 #endif
--- a/test/gtest/uct/ib/test_rc.cc
+++ b/test/gtest/uct/ib/test_rc.cc
@@ -143,6 +143,74 @@ UCS_TEST_P(test_rc_max_wr, send_limit)
 
 UCT_INSTANTIATE_RC_TEST_CASE(test_rc_max_wr)
 
+
+class test_rc_iface_address : public uct_test {
+protected:
+    entity *m_entity;
+    entity *m_entity_flush_rkey;
+
+public:
+    static uct_iface_params_t iface_params()
+    {
+        uct_iface_params_t params = {};
+
+        params.field_mask |= UCT_IFACE_PARAM_FIELD_OPEN_MODE;
+        params.field_mask |= UCT_IFACE_PARAM_FIELD_FEATURES;
+
+        params.features  = UCT_IFACE_FEATURE_PUT;
+        params.open_mode = UCT_IFACE_OPEN_MODE_DEVICE;
+        return params;
+    }
+
+    void init()
+    {
+        uct_test::init();
+
+        uct_iface_params_t params = iface_params();
+        m_entity                  = uct_test::create_entity(params);
+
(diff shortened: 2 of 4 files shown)


src/uct/ib/rc/accel/rc_mlx5_common.h
```diff
@@ -721,4 +721,14 @@ uct_rc_mlx5_common_iface_init_rx(uct_rc_mlx5_iface_common_t *iface,
...
 
+static UCS_F_ALWAYS_INLINE int
+uct_rc_mlx5_iface_flush_rkey_enabled(uct_rc_mlx5_iface_common_t *iface)
+{
+    uct_rc_iface_t *rc_iface = &iface->super;
+    uct_ib_md_t *md          = uct_ib_iface_md(&rc_iface->super);
+
+    return rc_iface->config.flush_remote &&
+           uct_ib_md_is_flush_rkey_valid(md->flush_rkey);
+}
```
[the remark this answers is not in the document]
tvegas1: fixed

test/gtest/uct/ib/test_rc.cc
```diff
@@ -7,7 +7,6 @@
 
 #include "test_rc.h"
 #include <uct/ib/rc/verbs/rc_verbs.h>
-#include <uct/ib/rc/accel/rc_mlx5.h>
```
brminich: you can leave the `include`, pls check the way it is included here
https://github.com/openucx/ucx/blob/master/test/gtest/uct/ib/test_ib_event.cc#L17

test/gtest/uct/ib/test_rc.cc
```diff
@@ -143,6 +143,52 @@ UCS_TEST_P(test_rc_max_wr, send_limit)
...
+        m_entity_flush_rkey = uct_test::create_entity(params);
+
+        m_entities.push_back(m_entity);
+        m_entities.push_back(m_entity_flush_rkey);
+    }
+};
+
+UCS_TEST_SKIP_COND_P(test_rc_mlx5, iface_get_device_addr,
+                     !check_caps(UCT_IFACE_FLAG_GET_ZCOPY |
+                                 UCT_IFACE_FLAG_GET_BCOPY))
```
yosefe, line 182: why?

test/gtest/uct/uct_test.h
```diff
@@ -570,6 +570,9 @@ class test_uct_iface_attrs : public uct_test {
     _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_verbs) \
     _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5)
 
+#define UCT_INSTANTIATE_RC_MLX5_TEST_CASE(_test_case) \
+    _UCT_INSTANTIATE_TEST_CASE(_test_case, rc_mlx5)
```
yosefe, line 574: maybe test that also DC address length not increased by flush key?

test/gtest/uct/ib/test_rc.cc
tvegas1: removed

test/gtest/uct/ib/test_rc.cc
tvegas1: added rc/dc/rc_verbs numerical values for ep iface address size, as requested in comments to pin size, instead of reusing same structure, so include is not needed.

brminich: i think better to use sizeof of the relevant structs (like in the original commit) rather than defining size to ep name mapping
wdyt?
cc @yosefe

yosefe (approved):

test/gtest/uct/ib/test_rc.cc
tvegas1: squashing commit as @yosefe approved, will commit on top if needed

yosefe (approved):

PR #8995, opened 2023-04-04
UCP/PROTOV2: Add rcache lookup perf estimation

ivankochin:
## What
Now protov2 performance calculation for choosing the protocols use only memory registration cost estimation, while UCP use rcache lookup by default now. In approximate case it can be defined that lookup is successful, so the lookup cost should be taken into account instead of memory registration.

### DIFF
--- a/src/ucp/proto/proto_init.c
+++ b/src/ucp/proto/proto_init.c
@@ -326,8 +326,8 @@ void ucp_proto_init_memreg_time(const ucp_proto_common_init_params_t *params,
                                 ucs_linear_func_t *memreg_time,
                                 ucp_proto_perf_node_t **perf_node_p)
 {
-    ucp_context_h context = params->super.worker->context;
-    ucp_proto_perf_node_t *perf_node;
+    ucp_context_h context            = params->super.worker->context;
+    ucp_proto_perf_node_t *perf_node = NULL;
     const uct_md_attr_v2_t *md_attr;
     ucp_md_index_t md_index;
     const char *md_name;
@@ -335,8 +335,17 @@ void ucp_proto_init_memreg_time(const ucp_proto_common_init_params_t *params,
     *memreg_time = UCS_LINEAR_FUNC_ZERO;
 
     if (reg_md_map == 0) {
-        *perf_node_p = NULL;
-        return;
+        goto out;
+    }
+
+    if (context->rcache != NULL) {
+        perf_node = ucp_proto_perf_node_new_data("rcache lookup", "");
+
+        *memreg_time = ucs_linear_func_make(50.0e-9, 0);
+
+        ucp_proto_perf_node_add_data(perf_node, "lookup", *memreg_time);
+
+        goto out;
     }
 
     perf_node = ucp_proto_perf_node_new_data("mem reg", "");
@@ -357,6 +366,7 @@ void ucp_proto_init_memreg_time(const ucp_proto_common_init_params_t *params,
         ucp_proto_perf_node_add_data(perf_node, "total", *memreg_time);
     }
 
+out:
     *perf_node_p = perf_node;
 }
 


src/ucp/proto/proto_init.c
```diff
@@ -339,22 +339,30 @@ void ucp_proto_init_memreg_time(const ucp_proto_common_init_params_t *params,
...
-        md_attr = &context->tl_mds[md_index].attr;
-        md_name = context->tl_mds[md_index].rsc.md_name;
-        ucs_linear_func_add_inplace(memreg_time, md_attr->reg_cost);
-        ucs_trace("md %s" UCP_PROTO_PERF_FUNC_FMT(reg_cost), md_name,
-                  UCP_PROTO_PERF_FUNC_ARG(&md_attr->reg_cost));
+        *memreg_time = ucs_rcache_lookup_cost();
 
-        ucp_proto_perf_node_add_data(perf_node, md_name, md_attr->reg_cost);
-    }
+        ucp_proto_perf_node_add_data(perf_node, "lookup", *memreg_time);
```
rakhmets, line 347: Do we have a place where we define labels for protocol selection tree?
If not so, I think it may be a good thing to start defining all the labels in one place. (I mean "lookup", "mem reg" ...).

And could you please attach a screenshot of the protocol selection tree with the new node to better understand how the new names are used.

ivankochin: Here is the screenshot:
image


Will take a look for a place for labels little-bit later

brminich (approved):

src/ucp/proto/proto_init.c
```diff
@@ -339,22 +339,30 @@ void ucp_proto_init_memreg_time(const ucp_proto_common_init_params_t *params,
         return;
     }
 
-    perf_node = ucp_proto_perf_node_new_data("mem reg", "");
+    if (context->rcache != NULL) {
+        perf_node = ucp_proto_perf_node_new_data("lookup", "");
```
yosefe, line 343: "lookup" -> "rcache" or "rcache lookup"

src/ucp/proto/proto_init.c
```diff
@@ -339,22 +339,30 @@ void ucp_proto_init_memreg_time(const ucp_proto_common_init_params_t *params,
...
-        md_name = context->tl_mds[md_index].rsc.md_name;
-        ucs_linear_func_add_inplace(memreg_time, md_attr->reg_cost);
-        ucs_trace("md %s" UCP_PROTO_PERF_FUNC_FMT(reg_cost), md_name,
-                  UCP_PROTO_PERF_FUNC_ARG(&md_attr->reg_cost));
+        *memreg_time = ucs_rcache_lookup_cost();
 
-        ucp_proto_perf_node_add_data(perf_node, md_name, md_attr->reg_cost);
-    }
+        ucp_proto_perf_node_add_data(perf_node, "lookup", *memreg_time);
+    } else {
```
yosefe, line 348: maybe use "return" instead of "else" to avoid changing existing code?

src/ucp/proto/proto_init.c
ivankochin: The only place that I can suggest is the field inside `ucp_context_t` structure, but I don't think that this is a good idea.

brminich (approved):

src/ucs/memory/rcache.c
```diff
@@ -1158,6 +1159,10 @@ size_t ucs_rcache_distribution_get_num_bins()
     return ucs_ilog2(ucs_rcache_stat_max_pow2() / UCS_RCACHE_STAT_MIN_POW2) + 2;
 }
 
+ucs_linear_func_t ucs_rcache_lookup_cost() {
+    return ucs_linear_func_make(50.0e-9, 0);
+}
```
yosefe, line 1164: IMO this wrapper is redundant

src/ucp/proto/proto_init.c
rakhmets: Sorry, my comment was not very clear.
I meant that it is better to have a place where we define labels for the tree:
```C
/* proto_debug.h, or create new proto_debug_def.h */
#define UCP_PROTO_PERF_NODE_RCACHE_LOOKUP      "rcache lookup"
#define UCP_PROTO_PERF_NODE_DATA_RCACHE_LOOKUP "lookup"

/* proto_init.c */
perf_node    = ucp_proto_perf_node_new_data(UCP_PROTO_PERF_NODE_RCACHE_LOOKUP, "");
*memreg_time = ucs_rcache_lookup_cost();
ucp_proto_perf_node_add_data(perf_node, UCP_PROTO_PERF_NODE_DATA_RCACHE_LOOKUP, *memreg_time);
```

ivankochin: It was decided to implement it for all the cases in the separate PR.

rakhmets (dismissed):

src/ucp/proto/proto_init.c
```diff
@@ -326,17 +326,26 @@ void ucp_proto_init_memreg_time(const ucp_proto_common_init_params_t *params,
...
     if (reg_md_map == 0) {
-        *perf_node_p = NULL;
-        return;
+        goto out;
+    }
+
+    if (context->rcache != NULL) {
+        perf_node = ucp_proto_perf_node_new_data("rcache lookup", "");
+
+        *memreg_time = ucs_linear_func_make(50.0e-9, 0);
```
ivankochin, line 344: As we discussed, this patch change the performance estimation through changing several `mem_reg` overheads to one `lookup`. But we already have `double uct_md_rcache_overhead(const uct_md_rcache_config_t *rcache_config)` internal interface which is supposed to be defined for the same purpose.

Now, this function is used to set `md->reg_cost` for IB MD. Is that a good idea to reuse this function in that patch?

If so, then rcache lookup seems MD independent from my point of view, but this function take `rcache_config` to take the `RCACHE_OVERHEAD` user control into account. Shouldn't we move this user control to MD independent place?

@yosefe @brminich @rakhmets  WDYT?

yosefe: eventually we'll remove uct rcache md config, but since there is no equivalent configuration in UCP today I think we better keep this PR simple, as-is, to resolve the performance issue and don't do any better UCP rcache config for now.

yosefe (dismissed):

yosefe (approved):

brminich (approved):

PR #7847, opened 2022-01-13
UCT/API: add dmabuf to md_mem_query attributes

Akshay-Venkatesh:
## What
Add dmabuf fd field in md_mem attributes

## Why ?
Needed by UCT/IB to register device memory exposed as a dmabuf

PR #8997, opened 2023-04-05
UCT/ROCM: introduce dmabuf support

edgargabriel:
## What
add support for dmabuf for the rocm memory domain.

### DIFF
--- a/src/uct/rocm/base/rocm_base.c
+++ b/src/uct/rocm/base/rocm_base.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) Advanced Micro Devices, Inc. 2019. ALL RIGHTS RESERVED.
+ * Copyright (C) Advanced Micro Devices, Inc. 2019-2023. ALL RIGHTS RESERVED.
  * See file LICENSE for terms.
  */
 
@@ -287,10 +287,51 @@ ucs_status_t uct_rocm_base_detect_memory_type(uct_md_h md, const void *addr,
     return UCS_OK;
 }
 
+int uct_rocm_base_is_dmabuf_supported()
+{
+    int dmabuf_supported = 0;
+
+#if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF
+    dmabuf_supported = 1;
+#endif
+
+    return dmabuf_supported;
+}
+
+static void uct_rocm_base_dmabuf_export(const void *addr, const size_t length,
+                                        ucs_memory_type_t mem_type,
+                                        int *dmabuf_fd, size_t *dmabuf_offset)
+{
+    int fd          = UCT_DMABUF_FD_INVALID;
+    uint64_t offset = 0;
+#if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF
+    hsa_status_t status;
+
+    if (mem_type == UCS_MEMORY_TYPE_ROCM) {
+        status = hsa_amd_portable_export_dmabuf(addr, length, &fd, &offset);
+        if (status != HSA_STATUS_SUCCESS) {
+            fd     = UCT_DMABUF_FD_INVALID;
+            offset = 0;
+            ucs_warn("failed to export dmabuf handle for addr %p / %zu", addr,
+                     length);
+        }
+
+        ucs_trace("dmabuf export addr %p %lu to dmabuf fd %d offset %zu\n",
+                  addr, length, fd, offset);
+    }
+#endif
+    *dmabuf_fd     = fd;
+    *dmabuf_offset = (size_t)offset;
+}
+
 ucs_status_t uct_rocm_base_mem_query(uct_md_h md, const void *addr,
                                      const size_t length,
                                      uct_md_mem_attr_t *mem_attr_p)
 {
+    size_t dmabuf_offset       = 0;
+    int is_exported            = 0;
+    ucs_memory_type_t mem_type = UCS_MEMORY_TYPE_HOST;
+    int dmabuf_fd;
     hsa_status_t status;
     hsa_device_type_t dev_type;
     hsa_amd_pointer_type_t hsa_mem_type;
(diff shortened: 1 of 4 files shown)


src/uct/rocm/base/rocm_base.c
```diff
@@ -287,10 +287,49 @@ ucs_status_t uct_rocm_base_detect_memory_type(uct_md_h md, const void *addr,
...
+int uct_rocm_base_is_dmabuf_supported()
+{
+    int dmabuf_supported = 0;
+
+#if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF == 1
+    dmabuf_supported = 1;
+#endif
+
+    return dmabuf_supported;
+}
```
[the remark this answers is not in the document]
edgargabriel: @ivankochin I think this is a good idea, I will make the change

src/uct/rocm/base/rocm_base.c
```diff
@@ -287,10 +287,46 @@ ucs_status_t uct_rocm_base_detect_memory_type(uct_md_h md, const void *addr,
...
+
+    if (mem_type == UCS_MEMORY_TYPE_ROCM) {
+        status = hsa_amd_portable_export_dmabuf(addr, length, &fd, &offset);
+        if (status != HSA_STATUS_SUCCESS) {
+            fd     = UCT_DMABUF_FD_INVALID;
+            offset = 0;
+            ucs_warn("failed to export dmabuf handle for addr %p / %lu", addr,
+                     length);
+        }
+        ucs_trace("dmabuf export addr %p %lu to dmabuf fd %d offset %lu\n",
```
yosefe, line 315: minor: space line before

src/uct/rocm/base/rocm_base.c
```diff
@@ -287,10 +287,46 @@ ucs_status_t uct_rocm_base_detect_memory_type(uct_md_h md, const void *addr,
...
+    uint64_t offset = 0;
+#if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF == 1
+    hsa_status_t status;
+
+    if (mem_type == UCS_MEMORY_TYPE_ROCM) {
+        status = hsa_amd_portable_export_dmabuf(addr, length, &fd, &offset);
+        if (status != HSA_STATUS_SUCCESS) {
+            fd     = UCT_DMABUF_FD_INVALID;
+            offset = 0;
+            ucs_warn("failed to export dmabuf handle for addr %p / %lu", addr,
```
yosefe, line 312: addr %p length %zu

src/uct/rocm/base/rocm_base.c
```diff
@@ -332,11 +368,13 @@ ucs_status_t uct_rocm_base_mem_query(uct_md_h md, const void *addr,
...
     if (mem_attr_p->field_mask & UCT_MD_MEM_ATTR_FIELD_DMABUF_FD) {
-        mem_attr_p->dmabuf_fd = UCT_DMABUF_FD_INVALID;
+        uct_rocm_base_dmabuf_export(addr, length, mem_type, &dmabuf_fd,
+                                    &dmabuf_offset);
+        mem_attr_p->dmabuf_fd = dmabuf_fd;
     }
 
     if (mem_attr_p->field_mask & UCT_MD_MEM_ATTR_FIELD_DMABUF_OFFSET) {
-        mem_attr_p->dmabuf_offset = 0;
+        mem_attr_p->dmabuf_offset = dmabuf_offset;
```
yosefe, line 377: what if user asked for UCT_MD_MEM_ATTR_FIELD_DMABUF_OFFSET without UCT_MD_MEM_ATTR_FIELD_DMABUF_OFFSET?
seems offset will be always 0 in this case which is wrong..

edgargabriel: @yosefe So that is a good question, can this really occur? We had this discussion previously whether it is legal to ask for the offset without the handle and vice versa, see

https://github.com/openucx/ucx/pull/7847#discussion_r882048898

yosefe: Seems we didn't document or prohibit it explicitly. And I think it should be simple enough to support it.

edgargabriel: ok, I can add that, but basically we have to re-export the same buffer in that case, and we make the assumption that the offset obtained in the second export can be used with the dmabuf handle of the first export. I would expect it to work in most scenarios, but I am wondering whether this is truly correct in the most generic sense

edgargabriel:
@yosefe, actually the header file does mandate that both offset and dmabfd_fd are set at the same time:

UCT_MD_MEM_ATTR_FIELD_DMABUF_FD     = UCS_BIT(4), /**< Request a cross-device dmabuf file
                                                           descriptor that represents a memory
                                                           region, and can be used to register
                                                           the region with another memory domain.
                                                           When UCT_MD_MEM_ATTR_FIELD_DMABUF_FD is
                                                           set, UCT_MD_MEM_ATTR_FIELD_DMABUF_OFFSET
                                                           must also be set. */
    UCT_MD_MEM_ATTR_FIELD_DMABUF_OFFSET = UCS_BIT(5)  /**< Request a cross-device dmabuf file
                                                           descriptor that represents a memory
                                                           region, and can be used to register
                                                           the region with another memory domain.
                                                           When UCT_MD_MEM_ATTR_FIELD_DMABUF_OFFSET
                                                           is set, UCT_MD_MEM_ATTR_FIELD_DMABUF_FD
                                                           must also be set. */

edgargabriel:
I take it back, it looks like it was removed again, it was originally in the file

src/uct/rocm/base/rocm_base.c
```diff
@@ -287,10 +287,51 @@ ucs_status_t uct_rocm_base_detect_memory_type(uct_md_h md, const void *addr,
     return UCS_OK;
 }
 
+int uct_rocm_base_is_dmabuf_supported()
+{
+    int dmabuf_supported = 0;
+
+#if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF == 1
```
brminich, line 294: minor: for consistency with the other code:
```suggestion
#if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF
```

edgargabriel: done, thanks!

edgargabriel:
The failures are unrelated to this pr, I verified that the ROCm components are not even compiled in those cases (Tests new on worker 1 and 2)

--snip--
```
configure: =========================================================
configure: UCX build configuration:
configure:         Build prefix:   /scrap/azure/agent-08/AZP_WORKSPACE/2/s/install
configure:    Configuration dir:   ${prefix}/etc/ucx
configure:   Preprocessor flags:   -DCPU_FLAGS="" -I${abs_top_srcdir}/src -I${abs_top_builddir} -I${abs_top_builddir}/src
configure:           C compiler:   gcc -O3 -g -Wall -Werror -fno-omit-frame-pointer -funwind-tables -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-label -Wno-long-long -Wno-endif-labels -Wno-sign-compare -Wno-multichar -Wno-deprecated-declarations -Winvalid-pch -Wno-pointer-sign -Werror-implicit-function-declaration -Wno-format-zero-length -Wnested-externs -Wshadow -Werror=declaration-after-statement
configure:         C++ compiler:   g++ -O3 -g -Wall -Werror -fno-omit-frame-pointer -funwind-tables -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-label -Wno-long-long -Wno-endif-labels -Wno-sign-compare -Wno-multichar -Wno-deprecated-declarations -Winvalid-pch
configure:         Multi-thread:   enabled
configure:         NUMA support:   enabled
configure:            MPI tests:   disabled
configure:          VFS support:   no
configure:        Devel headers:   no
configure: io_demo CUDA support:   no
configure:             Bindings:   < >
configure:          UCS modules:   < >
configure:          UCT modules:   < ib rdmacm cma knem xpmem >
configure:         CUDA modules:   < >
configure:         ROCM modules:   < >
configure:           IB modules:   < >
configure:          UCM modules:   < >
configure:         Perf modules:   < >
configure: =========================================================
```

they both fail at a similar spot with test_ucp_perf.envelope
--snip--
```
[ RUN      ] dcx/test_ucp_perf.envelope/32 <dc_x/am_lat_b>
[     INFO ]                   am_lat_b : 4091.447 usec
/scrap/azure/agent-10/AZP_WORKSPACE/2/s/contrib/../test/gtest/common/test_helpers.cc:55: Failure
Failed
Connection timed out - abort testing
```

yosefe (approved):

PR #9004, opened 2023-04-10
UCP: conditional params check in fast-path

ofirfarjun7:
## What
In fast path check params if and only if `ENABLE_PARAMS_CHECK` is set.

## Why ?
Performance.

## How ?
Check params if `ENABLE_PARAMS_CHECK` is set.


src/ucs/type/param.h
```diff
@@ -31,6 +31,13 @@ BEGIN_C_DECLS
              (_default))
 
 
+/**
+ * Check params on fast path cond.
+ */
+#define UCS_PARAM_CHECK_FAST_PATH(_condition) \
```
brminich, line 37: - imo, we do not need  to have `FAST_PATH` in the name, because this macro just adds check for `ENABLE_PARAMS`, which is reflected by `PARAM` prefix.
- maybe we can also define two other macro for `likely` and `unlikely` check? E.g.
`UCS_PARAM_CHECK_LIKELY`
`UCS_PARAM_CHECK_UNLIKELY`

src/ucp/rma/amo_send.c
```diff
@@ -25,40 +25,40 @@
...
                                         _action); \
         \
-        if (ENABLE_PARAMS_CHECK && \
-            (ucs_unlikely((_opcode) >= (_last_opcode)))) { \
-            ucs_error("invalid atomic opcode %d ", _opcode); \
-            _action; \
-        } \
+        if ( \
+            UCS_PARAM_CHECK_FAST_PATH((ucs_unlikely((_opcode) >= (_last_opcode))))) \
+            { \
```
gleon99, line 49: `) {` ?

src/uct/base/uct_md.c
```diff
@@ -337,9 +337,9 @@ ucs_status_t uct_config_modify(void *config, const char *name, const char *value
 static ucs_status_t
 uct_md_mkey_pack_params_check(uct_md_h md, uct_mem_h memh, void *mkey_buffer)
 {
-    if (ENABLE_PARAMS_CHECK) {
-        return ((md != NULL) && (memh != NULL) && (mkey_buffer != NULL)) ?
-               UCS_OK : UCS_ERR_INVALID_PARAM;
+    if (UCS_PARAM_CHECK_FAST_PATH(
+                !((md != NULL) && (memh != NULL) && (mkey_buffer != NULL)))) {
```
gleon99, line 341: Maybe simplify to:
`(md == NULL) || (memh == NULL) || (mkey_buffer == NULL)`

src/uct/base/uct_iface.h
```diff
@@ -155,7 +155,7 @@ enum {
  * Check the condition and return status as a pointer if not true.
  */
 #define UCT_CHECK_PARAM_PTR(_condition, _err_message, ...) \
-    if (ENABLE_PARAMS_CHECK && !(_condition)) { \
+    if (UCS_PARAM_CHECK_FAST_PATH(!(_condition))) { \
```
yosefe, line 158: IMO better make the names consistent, for example
UCS_PARAM_CHECK_FAST_PATH --> UCS_CHECK_PARAM
or
UCT_CHECK_PARAM -> UCT_PARAM_CHECK

ofirfarjun7: Regarding "UCS_PARAM_CHECK_FAST_PATH --> UCS_CHECK_PARAM", I think it should be 
UCS_PARAM_CHECK_FAST_PATH --> UCS_PARAM_CHECK because the name of the file is `param.h` under UCS.

The `_FAST_PATH` is actually leftovers. I thought to move `UCT_CHECK_PARAM` to `param.h` and call it `UCS_PARAM_CHECK`, This macro can replace many if statements in our code. I also thought to create another macro similar to this one without error message. It can also replace many if statements in our code.

WDYT?

yosefe: let's not expand this PR, which is mainly intended to fix performance issue and not refactor
we can rename UCS_PARAM_CHECK_FAST_PATH -> UCS_PARAM_CHECK if it doesn't affect many places

yosefe (dismissed):

yosefe (approved):

PR #9005, opened 2023-04-10
Revert "UCT/ROCM: move memh caches to iface"

edgargabriel:
## What
This reverts commit a5246b8b497d6d4d2899d288bec82274ed2afafc.

## Why ?
Commit a5246b8b4 moved the memhandle caches from the endpoints to the iface in rocm_copy and rocm_ipc.  
Both memhandle caches operate based on lookup of the virtual address. Different processes can use the same virtual address. While the memhandle cache does perform a comparison of the ipc keys in rocm_ipc, this will lead to thrashing of ipc registered regions across different processes if the same virtual address is used. 

I *think* the rocm_copy component  (which was the initial driver for the commit originally) does not have this problem, and the bug is/was trying to apply the same method to rocm_ipc. For the sake of simplicity lets revert the commit first before deciding whether to introduce a patch for rocm_copy only.


edgargabriel:
Closed the pr, will create a new pr that only reverts the changes for rocm_ipc.  I am pretty sure that the new version for rocm_copy is correct and has some benefits.

PR #9008, opened 2023-04-11
UCP: Fix memory detection for iov dts

brminich:
## What
Fixes memory detection for ucp iov datatypes for proto v2

## Why ?
Fixes [bug](https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=4065349&cmtNo=) (Nvidia internal link)

### DIFF
--- a/src/ucp/core/ucp_request.inl
+++ b/src/ucp/core/ucp_request.inl
@@ -753,8 +753,11 @@ ucp_request_recv_data_unpack(ucp_request_t *req, const void *data,
     ucs_assertv(req->status == UCS_OK, "status: %s",
                 ucs_status_string(req->status));
 
-    ucp_trace_req(req, "unpack recv_data req_len %zu data_len %zu offset %zu last: %s",
-                  req->recv.length, length, offset, last ? "yes" : "no");
+    ucp_trace_req(req,
+                  "unpack recv_data req_len %zu data_len %zu offset %zu "
+                  "memtype %s last: %s", req->recv.length, length, offset,
+                  ucs_memory_type_names[req->recv.mem_type],
+                  last ? "yes" : "no");
 
     if (ucs_unlikely((length + offset) > req->recv.length)) {
         return ucp_request_recv_msg_truncated(req, length, offset);
@@ -1003,19 +1006,30 @@ ucp_request_param_user_data(const ucp_request_param_t *param)
 
 static UCS_F_ALWAYS_INLINE ucs_memory_type_t
 ucp_request_get_memory_type(ucp_context_h context, const void *address,
-                            size_t length, const ucp_request_param_t *param)
+                            size_t count, ucp_datatype_t datatype,
+                            size_t contig_length,
+                            const ucp_request_param_t *param)
 {
     ucp_memory_info_t mem_info;
+    uint8_t UCS_V_UNUSED dummy_sg_count;
 
-    if (!(param->op_attr_mask & UCP_OP_ATTR_FIELD_MEMORY_TYPE) ||
-        (param->memory_type == UCS_MEMORY_TYPE_UNKNOWN)) {
-        ucp_memory_detect(context, address, length, &mem_info);
-        ucs_assert(mem_info.type < UCS_MEMORY_TYPE_UNKNOWN);
-        return (ucs_memory_type_t)mem_info.type;
+    if ((param->op_attr_mask & UCP_OP_ATTR_FIELD_MEMORY_TYPE) &&
+        (param->memory_type != UCS_MEMORY_TYPE_UNKNOWN)) {
+        return param->memory_type;
     }
 
-    ucs_assert(param->memory_type < UCS_MEMORY_TYPE_UNKNOWN);
-    return param->memory_type;
+    if (UCP_DT_IS_CONTIG(datatype)) {
+        ucp_memory_detect_param(context, address, contig_length, param,
+                                &mem_info);
+    } else if (UCP_DT_IS_IOV(datatype)) {
+        ucp_dt_iov_memtype_detect(context, (ucp_dt_iov_t*)address, count,
+                                  param, &dummy_sg_count, &mem_info);
+    } else {
+        ucp_memory_info_set_host(&mem_info);
+    }
+
+    ucs_assert((ucs_memory_type_t)mem_info.type < UCS_MEMORY_TYPE_UNKNOWN);
+    return (ucs_memory_type_t)mem_info.type;
 }
 
 static UCS_F_ALWAYS_INLINE void
--- a/src/ucp/dt/datatype_iter.inl
+++ b/src/ucp/dt/datatype_iter.inl
@@ -36,16 +36,14 @@ ucp_datatype_contig_iter_init(ucp_context_h context, void *buffer,
                               size_t length, ucp_datatype_iter_t *dt_iter,
                               const ucp_request_param_t *param)
 {
-    if (!(param->op_attr_mask & UCP_OP_ATTR_FIELD_MEMH)) {
-        ucp_memory_detect(context, buffer, length, &dt_iter->mem_info);
-    }
+    ucp_memory_detect_param(context, buffer, length, param, &dt_iter->mem_info);
 
-    dt_iter->length                 = length;
-    dt_iter->type.contig.buffer     = buffer;
-    dt_iter->type.contig.memh       = NULL;
+    dt_iter->length             = length;
+    dt_iter->type.contig.buffer = buffer;
+    dt_iter->type.contig.memh   = NULL;
 }
 
-static UCS_F_ALWAYS_INLINE void
+static UCS_F_ALWAYS_INLINE ucs_status_t
 ucp_datatype_iov_iter_init(ucp_context_h context, void *buffer, size_t count,
                            size_t length, ucp_datatype_iter_t *dt_iter,
                            uint8_t *sg_count, const ucp_request_param_t *param)
@@ -61,16 +59,8 @@ ucp_datatype_iov_iter_init(ucp_context_h context, void *buffer, size_t count,
     dt_iter->type.iov.iov_offset = 0;
     dt_iter->type.iov.memh       = NULL;
 
-    if (ucs_likely(count > 0)) {
-        *sg_count = ucs_min(count, (size_t)UINT8_MAX);
-        if (!(param->op_attr_mask & UCP_OP_ATTR_FIELD_MEMH)) {
-            ucp_memory_detect(context, iov->buffer, iov->length,
-                              &dt_iter->mem_info);
-        }
-    } else {
-        *sg_count = 1;
-        ucp_memory_info_set_host(&dt_iter->mem_info);
-    }
+    return ucp_dt_iov_memtype_detect(context, iov, count, param, sg_count,
+                                     &dt_iter->mem_info);
 }
 
 static UCS_F_ALWAYS_INLINE void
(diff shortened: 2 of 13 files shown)


src/ucp/core/ucp_request.inl
```diff
@@ -1003,19 +1003,28 @@ ucp_request_param_user_data(const ucp_request_param_t *param)
...
+        return param->memory_type;
+    }
+
+    if ((dt & UCP_DATATYPE_CLASS_MASK) == UCP_DATATYPE_CONTIG) {
         ucp_memory_detect(context, address, length, &mem_info);
-        ucs_assert(mem_info.type < UCS_MEMORY_TYPE_UNKNOWN);
-        return (ucs_memory_type_t)mem_info.type;
+    } else if ((dt & UCP_DATATYPE_CLASS_MASK) == UCP_DATATYPE_IOV) {
+        iov = (const ucp_dt_iov_t*)address;
+        ucp_memory_detect(context, iov->buffer, iov->length, &mem_info);
```
Akshay-Venkatesh, line 1021: @brminich This assumes uniformity in iovs, right? If iov is composed of two buffers (one host, one device), then mem_info reporting mem_info.type = UCS_MEMORY_TYPE_HOST would not be accurate, right? If you agree, then mem_info will likely have to carry sys_dev, mem_type info for each vector to allow UCP to actually handle an iov with non-uniform memory types. If we don't plan to support that, shouldn't we fail here?

brminich: maybe we can add an assert to avoid hurting performance

Akshay-Venkatesh: If you mean, assert that all iovs have the same memory type, then agreed.

brminich: @Akshay-Venkatesh, sorry force-pushed the changes as the original code was modified a lot

src/ucp/core/ucp_request.inl
```diff
@@ -1003,19 +1006,29 @@ ucp_request_param_user_data(const ucp_request_param_t *param)
...
-        ucs_assert(mem_info.type < UCS_MEMORY_TYPE_UNKNOWN);
-        return (ucs_memory_type_t)mem_info.type;
+    if ((param->op_attr_mask & UCP_OP_ATTR_FIELD_MEMORY_TYPE) &&
+        (param->memory_type != UCS_MEMORY_TYPE_UNKNOWN)) {
+        return param->memory_type;
     }
 
-    ucs_assert(param->memory_type < UCS_MEMORY_TYPE_UNKNOWN);
-    return param->memory_type;
+    if ((dt & UCP_DATATYPE_CLASS_MASK) == UCP_DATATYPE_CONTIG) {
```
yosefe, line 1019: UCP_DT_IS_CONTIG

src/ucp/core/ucp_request.inl
```diff
@@ -1003,19 +1006,29 @@ ucp_request_param_user_data(const ucp_request_param_t *param)
 
 static UCS_F_ALWAYS_INLINE ucs_memory_type_t
 ucp_request_get_memory_type(ucp_context_h context, const void *address,
-                            size_t length, const ucp_request_param_t *param)
+                            size_t count, ucp_datatype_t dt,
```
yosefe, line 1009: dt -> datatype

src/ucp/core/ucp_request.inl
```diff
@@ -1003,19 +1006,29 @@ ucp_request_param_user_data(const ucp_request_param_t *param)
...
+        return param->memory_type;
     }
 
-    ucs_assert(param->memory_type < UCS_MEMORY_TYPE_UNKNOWN);
-    return param->memory_type;
+    if ((dt & UCP_DATATYPE_CLASS_MASK) == UCP_DATATYPE_CONTIG) {
+        ucp_memory_detect_param(context, address,
+                                ucp_contig_dt_length(dt, count), param,
+                                &mem_info);
+    } else if ((dt & UCP_DATATYPE_CLASS_MASK) == UCP_DATATYPE_IOV) {
```
yosefe, line 1023: UCP_DT_IS_IOV

src/ucp/dt/dt_iov.c
```diff
@@ -125,3 +126,31 @@ size_t ucp_dt_iov_count_nonempty(const ucp_dt_iov_t *iov, size_t iovcnt)
...
+                               size_t iovcnt, const ucp_request_param_t *param,
+                               ucp_memory_info_t *mem_info)
+{
+    ucp_memory_info_t UCS_V_UNUSED check_mem_info;
+    int UCS_V_UNUSED i;
+
+    if (ucs_likely(iovcnt > 0)) {
+        ucp_memory_detect_param(context, iov->buffer, iov->length, param,
+                                mem_info);
+#if ENABLE_ASSERT
```
yosefe, line 140: IMO it should be with ENABLE_PARAM_CHECK

src/ucp/dt/dt_iov.c
```diff
@@ -125,3 +126,31 @@ size_t ucp_dt_iov_count_nonempty(const ucp_dt_iov_t *iov, size_t iovcnt)
...
+        for (i = 1; i < iovcnt; ++i) {
+            ucp_memory_detect_param(context, iov[i].buffer, iov[i].length,
+                                    param, &check_mem_info);
+            if ((check_mem_info.type != mem_info->type) ||
+                (check_mem_info.sys_dev != mem_info->sys_dev)) {
+                ucs_fatal("inconsistent iov memtypes: iov[%d]=%u-%u "
+                          "iov[0]=%u-%u iovcnt=%zu", i, check_mem_info.type,
+                          check_mem_info.sys_dev, mem_info->type,
```
yosefe, line 148: print names type by name

src/ucp/dt/datatype_iter.inl
```diff
@@ -61,16 +59,9 @@ ucp_datatype_iov_iter_init(ucp_context_h context, void *buffer, size_t count,
...
-            ucp_memory_detect(context, iov->buffer, iov->length,
-                              &dt_iter->mem_info);
-        }
-    } else {
-        *sg_count = 1;
-        ucp_memory_info_set_host(&dt_iter->mem_info);
-    }
+    ucp_dt_iov_memtype_detect(context, iov, count, param, &dt_iter->mem_info);
+
+    *sg_count = ucs_likely(count > 0) ? ucs_min(count, (size_t)UINT8_MAX) : 1;
```
yosefe, line 64: extra branch here

src/ucp/dt/dt_iov.c
```diff
@@ -125,3 +126,31 @@ size_t ucp_dt_iov_count_nonempty(const ucp_dt_iov_t *iov, size_t iovcnt)
...
+    if (ucs_likely(iovcnt > 0)) {
+        ucp_memory_detect_param(context, iov->buffer, iov->length, param,
+                                mem_info);
+#if ENABLE_ASSERT
+        for (i = 1; i < iovcnt; ++i) {
+            ucp_memory_detect_param(context, iov[i].buffer, iov[i].length,
+                                    param, &check_mem_info);
+            if ((check_mem_info.type != mem_info->type) ||
+                (check_mem_info.sys_dev != mem_info->sys_dev)) {
+                ucs_fatal("inconsistent iov memtypes: iov[%d]=%u-%u "
```
yosefe, line 146: maybe we could fail just the one send op, and not fatal error?

src/ucp/rma/rma_send.c
```diff
@@ -133,8 +133,9 @@ ucp_rma_request_init(ucp_request_t *req, ucp_ep_h ep, const void *buffer,
     req->send.ep              = ep;
     req->send.buffer          = (void*)buffer;
     req->send.datatype        = ucp_dt_make_contig(1);
-    req->send.mem_type        = ucp_request_get_memory_type(context, buffer,
-                                                            length, param);
+    req->send.mem_type        = ucp_request_get_memory_type(
+                                    context, buffer, length, req->send.datatype,
```
yosefe, line 137: ucp_dt_make_contig(1) instead of req->send.datatype
or we could use local helper var for it..
let compiler optimize

src/ucp/core/ucp_request.inl
```diff
@@ -1003,19 +1006,29 @@ ucp_request_param_user_data(const ucp_request_param_t *param)
...
-        return (ucs_memory_type_t)mem_info.type;
+    if ((param->op_attr_mask & UCP_OP_ATTR_FIELD_MEMORY_TYPE) &&
+        (param->memory_type != UCS_MEMORY_TYPE_UNKNOWN)) {
+        return param->memory_type;
     }
 
-    ucs_assert(param->memory_type < UCS_MEMORY_TYPE_UNKNOWN);
-    return param->memory_type;
+    if ((dt & UCP_DATATYPE_CLASS_MASK) == UCP_DATATYPE_CONTIG) {
+        ucp_memory_detect_param(context, address,
```
yosefe, line 1020: IMO we should not introduce ucp_memory_detect_param(), but instead check for MEMH field in this function (ucp_request_get_memory_type)

src/ucp/core/ucp_request.inl
```diff
@@ -1003,19 +1006,29 @@ ucp_request_param_user_data(const ucp_request_param_t *param)
...
+                                ucp_contig_dt_length(dt, count), param,
```
yosefe, line 1021: extra multiplication operation here

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -497,6 +501,17 @@ class test_ucp_am_nbx : public test_ucp_am_base {
         EXPECT_EQ(m_recv_counter, m_send_counter);
     }
 
+    void test_am_send_recv_memtype(size_t size, size_t header_size = 8)
+    {
+        std::vector<int> dts = {UCP_DATATYPE_CONTIG};
+
+        if (is_proto_enabled()) {
```
yosefe, line 508: why only protov2?
seems like the fix you did in UCP is also for protov1

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -497,6 +501,17 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+            dts.push_back(UCP_DATATYPE_IOV);
+        }
+
+        test_datatypes([&]() { test_am_send_recv(size, header_size); }, dts);
```
yosefe, line 512: why not test generic dt?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -333,7 +333,11 @@ class test_ucp_am_nbx : public test_ucp_am_base {
         m_rx_memh = NULL;
     }
 
-    void test_datatypes(std::function<void()> test_f);
+    void test_datatypes(std::function<void()> test_f,
+                        const std::vector<int> &datatypes =
```
yosefe, line 337: std::vector<ucp_dt_type>

src/ucp/core/ucp_request.inl
brminich: ok

src/ucp/core/ucp_request.inl
yosefe: before this change, for config datatype, we calculated ucp_contig_dt_length() once
now, for protov1, we calculate it twice - both in req->send.length init and memtype detect

src/ucp/dt/dt_iov.c
yosefe: let's do this, all functions are inline so it could be done in case ENABLE_PARAM_CHECK is set

brminich: `ucp_dt_iov_memtype_detect` is not inlined,
also have to propogate status up to `ucp_tag_recv_common` and a couple of functions on the stack are not inlined. 

imo it does not worth the effort

yosefe (approved):

yosefe (approved):

PR #9009, opened 2023-04-12
UCP/WIREUP: replay pending AM short request after EP reconf

evgeny-leksikov:
## What
replay pending AM short request after EP reconfiguration

## Why ?
TCP has too large max_short value which is not supported by IB transport

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -561,11 +561,21 @@ void ucp_ep_release_id(ucp_ep_h ep)
     ep->ext->local_ep_id = UCS_PTR_MAP_KEY_INVALID;
 }
 
+/* TODO: err_mode field could be part of flags */
 void ucp_ep_config_key_set_err_mode(ucp_ep_config_key_t *key,
                                     unsigned ep_init_flags)
 {
     key->err_mode = (ep_init_flags & UCP_EP_INIT_ERR_MODE_PEER_FAILURE) ?
                     UCP_ERR_HANDLING_MODE_PEER : UCP_ERR_HANDLING_MODE_NONE;
+}
+
+void ucp_ep_config_key_init_flags(ucp_ep_config_key_t *key,
+                                  unsigned ep_init_flags)
+{
+    if (ucs_test_all_flags(ep_init_flags,
+                           UCP_EP_INIT_CREATE_AM_LANE | UCP_EP_INIT_CM_PHASE)) {
+        key->flags |= UCP_EP_CONFIG_KEY_FLAG_INTERMEDIATE;
+    }
 }
 
 ucs_status_t
@@ -761,6 +771,7 @@ ucs_status_t ucp_ep_init_create_wireup(ucp_ep_h ep, unsigned ep_init_flags,
 
     ucp_ep_config_key_reset(&key);
     ucp_ep_config_key_set_err_mode(&key, ep_init_flags);
+    ucp_ep_config_key_init_flags(&key, ep_init_flags);
 
     key.num_lanes = 1;
     /* all operations will use the first lane, which is a stub endpoint before
@@ -2470,6 +2481,7 @@ ucs_status_t ucp_ep_config_init(ucp_worker_h worker, ucp_ep_config_t *config,
     size_t min_rndv_thresh, min_am_rndv_thresh;
     size_t rma_zcopy_thresh;
     size_t am_max_eager_short;
+    uint64_t short_am_cap_flag, short_tag_cap_flag;
     double get_zcopy_max_bw[UCS_MEMORY_TYPE_LAST];
     double put_zcopy_max_bw[UCS_MEMORY_TYPE_LAST];
     ucs_status_t status;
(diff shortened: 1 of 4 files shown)


src/ucp/core/ucp_am.c
```diff
@@ -910,6 +910,10 @@ ucp_am_send_req(ucp_request_t *req, size_t count,
      * Otherwise, return the request.
      */
     ucp_request_send(req);
+    if (ucs_unlikely(param == NULL)) {
```
brminich, line 913: isn't it too late to check? Note that it previously used on lines 860 and 886

src/ucp/core/ucp_am.c
```diff
@@ -567,7 +567,7 @@ ucp_am_send_short(ucp_ep_h ep, uint16_t id, uint16_t flags, const void *header,
     return uct_ep_am_short_iov(ucp_ep_get_am_uct_ep(ep), am_id, iov, iov_cnt);
 }
 
-static ucs_status_t ucp_am_contig_short(uct_pending_req_t *self)
+ucs_status_t ucp_am_contig_short(uct_pending_req_t *self)
```
brminich, line 570: not needed

src/ucp/core/ucp_am.c
```diff
@@ -583,7 +583,7 @@ static ucs_status_t ucp_am_contig_short(uct_pending_req_t *self)
     return ucp_am_short_handle_status_from_pending(req, status);
 }
 
-static ucs_status_t ucp_am_contig_short_reply(uct_pending_req_t *self)
+ucs_status_t ucp_am_contig_short_reply(uct_pending_req_t *self)
```
brminich, line 586: not needed

src/ucp/wireup/wireup.c
```diff
@@ -973,11 +973,20 @@ ucp_wireup_replay_pending_request(uct_pending_req_t *self, ucp_ep_h ucp_ep)
...
+                      req->send.lane,
+                      ucs_debug_get_symbol_name(req->send.uct.func));
+        if (req->send.uct.func == ucp_am_contig_short) {
+            ucs_assert(!(req->flags & UCP_REQUEST_FLAG_PROTO_SEND));
+            ucp_am_short_request_replay(req);
+        } else if (req->send.uct.func == ucp_am_contig_short_reply) {
+            ucs_assert(!(req->flags & UCP_REQUEST_FLAG_PROTO_SEND));
+            ucp_am_short_reply_request_replay(req);
+        } else {
+            ucp_request_send(req);
```
brminich, line 988: why is this relevant for short sends only with AM API?

src/ucp/core/ucp_am.c
evgeny-leksikov: it's needed for `src/ucp/wireup/wireup.c`

src/ucp/core/ucp_am.c
evgeny-leksikov: it's needed for `src/ucp/wireup/wireup.c`

src/ucp/core/ucp_am.c
evgeny-leksikov: it's not used for short/EAGER flow. I agree that it looks hacky but on other hand it could code duplication. Any idea how to refactor this code is welcome

src/ucp/wireup/wireup.c
evgeny-leksikov: technically the issue is relevant for any protocol, AM/short is real use case, proto v2 solves the issue by design

src/ucp/core/ucp_am.c
brminich: for instance using the empty params structure instead of NULL

evgeny-leksikov: applied, pls take a look if it looks better for you

evgeny-leksikov: but it seems that the empty params  is valid use case for `ucp_am_send_nbx` and may cause unexpected behavior from perspective of this PR

src/ucp/core/ucp_ep.c
```diff
@@ -561,13 +561,25 @@ void ucp_ep_release_id(ucp_ep_h ep)
...
+void ucp_ep_config_key_init_flags(ucp_ep_config_key_t *key,
+                                  unsigned ep_init_flags)
+{
+    if (ucs_test_all_flags(ep_init_flags,
+                           UCP_EP_INIT_CREATE_AM_LANE_ONLY |
+                           UCP_EP_INIT_CM_PHASE)) {
+        key->flags |= UCP_EP_CONFIG_KEY_FLAG_INTERMEDIATE;
+    }
+}
+
```
yosefe, line 581: remove space line

src/ucp/core/ucp_ep.c
```diff
@@ -2697,7 +2710,9 @@ ucs_status_t ucp_ep_config_init(ucp_worker_h worker, ucp_ep_config_t *config,
             }
 
             config->tag.eager.max_short = ucp_ep_config_max_short(
-                    worker->context, iface_attr, UCT_IFACE_FLAG_TAG_EAGER_SHORT,
+                    worker->context, iface_attr,
+                    (config->key.flags & UCP_EP_CONFIG_KEY_FLAG_INTERMEDIATE) ?
+                    0 : UCT_IFACE_FLAG_TAG_EAGER_SHORT,
```
yosefe, line 2715: use helper var (in multiple places)

src/ucp/core/ucp_ep.h
```diff
@@ -173,8 +173,21 @@ enum {
...
+     */
+    UCP_EP_CONFIG_KEY_FLAG_SELF         = UCS_BIT(0),
+
+    /**
+     * Endpoint is within same node.
+     */
+    UCP_EP_CONFIG_KEY_FLAG_INTRA_NODE   = UCS_BIT(1),
+
+    /**
+     * Endpoint is in intermediate state, some capabilities may be limited or
```
yosefe, line 187: Endpoint is in an intermediate connection establishment phase

yosefe (dismissed):

yosefe (approved):

PR #9012, opened 2023-04-12
UCP/GTEST: Fix AM rndv memtypes

brminich:
## What
Fix mem types recognition for UCP AM rndv gtests. Currently all the tests are using host memory (even if cuda requested)

### DIFF
--- a/src/ucp/rndv/rndv.c
+++ b/src/ucp/rndv/rndv.c
@@ -122,7 +122,8 @@ ucp_rndv_is_put_pipeline_needed(uintptr_t remote_address, size_t length,
                                 const ucp_ep_rndv_zcopy_config_t *put_zcopy,
                                 int is_get_zcopy_failed)
 {
-    if (ucp_rkey_packed_mem_type(rkey_buf) == UCS_MEMORY_TYPE_HOST) {
+    if ((remote_address != 0) &&
+        (ucp_rkey_packed_mem_type(rkey_buf) == UCS_MEMORY_TYPE_HOST)) {
         return 0;
     }
 
@@ -1299,6 +1300,8 @@ ucp_rndv_recv_start_get_pipeline(ucp_worker_h worker, ucp_request_t *rndv_req,
         goto err;
     }
 
+    ucp_rndv_recv_data_init(rreq, size);
+
     offset = 0;
     while (offset != size) {
         length = ucp_rndv_adjust_zcopy_length(min_zcopy, max_frag_size, 0,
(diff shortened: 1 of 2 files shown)


test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1798,6 +1798,17 @@ class test_ucp_am_nbx_rndv_memtype : public test_ucp_am_nbx_rndv {
         modify_config("RNDV_THRESH", "128");
         test_ucp_am_nbx::init();
     }
+
+private:
+    virtual ucs_memory_type_t tx_memtype() const
```
gleon99, line 1803: Maybe add `override`?

gleon99: These functions are code duplication with `test_ucp_am_nbx_eager_memtype`.
WDYT of common class (something like `test_ucp_am_memtype` which would contain these 2, then `test_ucp_am_nbx_rndv_memtype` and `test_ucp_am_nbx_eager_memtype` can inherit it?

brminich: i'd want to avoid multiple inheritance and `test_ucp_am_nbx_rndv_memtype` needs to inherit rndv class `test_ucp_am_nbx_rndv`

gleon99: 👍🏼

src/ucp/rndv/rndv.c
```diff
@@ -1712,6 +1713,7 @@ UCS_PROFILE_FUNC_VOID(ucp_rndv_receive, (worker, rreq, rndv_rts_hdr, rkey_buf),
                         worker, rndv_req, rreq, rndv_rts_hdr->sreq.req_id,
                         rkey_buf, rndv_rts_hdr->address, rndv_rts_hdr->size, 0);
                 if (status == UCS_OK) {
+                    ucp_rndv_recv_data_init(rreq, rndv_rts_hdr->size);
```
yosefe, line 1716: shouldn't we do it before calling ucp_rndv_recv_start_get_pipeline?
what if something completes in-place inside ucp_rndv_recv_start_get_pipeline ?

src/ucp/rndv/rndv.c
```diff
@@ -1696,12 +1699,12 @@ UCS_PROFILE_FUNC_VOID(ucp_rndv_receive, (worker, rreq, rndv_rts_hdr, rkey_buf),
                                              rreq->recv.mem_type,
                                              is_get_zcopy_failed)) {
             put_zcopy = &ep_config->rndv.put_zcopy;
-            ucp_rndv_recv_data_init(rreq, rndv_rts_hdr->size);
             if (ucp_rndv_is_put_pipeline_needed(rndv_rts_hdr->address,
                                                 rndv_rts_hdr->size, rkey_buf,
                                                 get_zcopy, put_zcopy,
                                                 is_get_zcopy_failed)) {
                 /* send FRAG RTR for sender to PUT the fragment. */
+                ucp_rndv_recv_data_init(rreq, rndv_rts_hdr->size);
```
yosefe, line 1707: what about line 1736 and 1745?

yosefe (approved):

PR #9014, opened 2023-04-13
AZP: retry modules load when automount fails - v1.14.x

Al3xR01:
What
Workaround for automount availability issues while loading modules.

Why ?
Multiple job failures due to the on and off availability of automount shares.

How ?
Attempt to remount by accessing the NFS.

### DIFF
--- a/buildlib/az-helpers.sh
+++ b/buildlib/az-helpers.sh
@@ -104,21 +104,30 @@ function az_init_modules() {
 
 #
 # Test if an environment module exists and load it if yes.
+# Retry 5 times in case of automount failure.
 # Otherwise, return error code.
 #
 function az_module_load() {
     module=$1
+    retries=5
 
-    if module avail -t 2>&1 | grep -q "^$module\$"
-    then
-        module load $module
-        return 0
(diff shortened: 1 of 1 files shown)


artemry-nv (dismissed):

yosefe (approved):

yosefe:
@Al3xR01 failure seems relevant, maybe more fixes are missing?
```
dot         module-git  module-info modules     null        use.own
+ eval
+ azure_log_warning 'Module dev/go-latest cannot be loaded'
+ test xyes = xno
+ msg='Module dev/go-latest cannot be loaded'
+ set +x
##[warning]Module dev/go-latest cannot be loaded
```

Al3xR01:
It's missing https://github.com/openucx/ucx/pull/9015 which is now merged, so should be Ok.
```
+ source buildlib/az-helpers.sh
  (missing `az_init_modules` here)
+ az_module_load dev/go-lates

```

yosefe:
@Al3xR01 it still fails

yosefe (approved):

PR #9031, opened 2023-04-20
AZP: Kill zombie processes

Alexey-Rivkin:
## What
Clean up leftover zombie processes.

The script runs hourly on all CI hosts, scheduled via crontab.
The PR serves exclusively for visibility and documentation reasons.

### DIFF
--- a/buildlib/tools/zombie_killer.sh
+++ b/buildlib/tools/zombie_killer.sh
@@ -1,0 +1,20 @@
+#!/bin/bash -eEx
+
+# Search for zombie or defunct processes; kill them all
+out=$(ps -eo pid,stat | awk '$2 == "Z|defunct" {print $1}')
+if [ -z "$out" ]; then
+  echo "No zombies found"
+  exit 0
+else
+  for i in $out ; do
+    kill -9 "$i"
+  done
+fi
+
+# Search again; tell parents you killed their kids; kill parents
+out=$(ps -eo pid,stat | awk '$2 == "Z|defunct" {print $1}')
+for i in $out ; do
+  pid_parent=$(ps -o ppid=$i)
+  kill -s SIGCHLD $pid_parent
+  kill -9 $pid_parent
+done


buildlib/tools/zombie_killer.sh
```diff
@@ -0,0 +1,29 @@
...
+out=$(ps -eo pid,stat | awk '$2 == "Z|defunct" {print $1}')
+for i in $out ; do
+  pid_parent=$(ps -o ppid=$i)
+  kill -s SIGCHLD $pid_parent
+  kill -9 $pid_parent
+done
+
+# DISABLED as it interrupts jobs.
+# Search for processes running for more than 6 hours; kill them all
+# out=$(ps -eo pid,user,etimes | grep 'swx-azu' | awk '$3 > 21600 {print $1}')
```
yosefe, line 24: maybe some parameters (username, duration) should come as command line arguments for this script?

Al3xR01: 1. The script is not intended to run interactively.
2. For now, I see no reason to kill long-lasting processes. Keeping the code block commented out for potential future reconsideration.

yosefe: @Alexey-Rivkin maybe better remove this commented out code? can always keep it somewhere else
we prefer to not have commented out code in upstream

buildlib/tools/zombie_killer.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ -z "$out" ]; then
+  echo "No zombies found"
+  exit 0
+else
+  for i in $out ; do
+    kill -9 "$i"
+  done
+fi
+
+# Search again; tell parents you killed their kids; kill parents
```
artemry-nv, line 14: Why do we need 2nd iteration?

Alexey-Rivkin: The original repo doesn't provide an explanation ([internal link](https://gitlab-master.nvidia.com/nbu-swx/hpc/swx-infrastructure/-/blob/master/scripts/azure/zombie_killer.sh)).
My guess is that if a process is still up after the first round, we need to approach its parents.

artemry-nv (approved):

yosefe (approved):

PR #9032, opened 2023-04-20
UCT/IB: Added IB iface is_reachable_v2

shasson5:
## What
UCT/IB: Added IB iface is_reachable_v2

## Why ?
Improve IB iface locality test by only comparing specific address fields.

### DIFF
--- a/src/uct/base/uct_iface.c
+++ b/src/uct/base/uct_iface.c
@@ -248,6 +248,23 @@ uct_base_iface_is_reachable_v2(const uct_iface_h iface,
 {
     uct_iface_reachability_scope_t scope;
 
+    if (!uct_iface_is_reachable(iface, params->device_addr,
+                                params->iface_addr)) {
+        return 0;
+    }
+
+    scope = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params, scope, SCOPE,
+                            UCT_IFACE_REACHABILITY_SCOPE_NETWORK);
+
+    return (scope == UCT_IFACE_REACHABILITY_SCOPE_NETWORK) ||
+           uct_iface_is_same_device(iface, params->device_addr);
+}
+
+int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
+                              const uct_iface_is_reachable_params_t *params)
+{
+    const uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
+
     if (!ucs_test_all_flags(params->field_mask,
                             UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR |
                             UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR)) {
@@ -260,19 +277,6 @@ uct_base_iface_is_reachable_v2(const uct_iface_h iface,
         params->info_string[0] = '\0';
     }
 
-    scope = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params, scope, SCOPE,
-                            UCT_IFACE_REACHABILITY_SCOPE_NETWORK);
-
-    return (scope == UCT_IFACE_REACHABILITY_SCOPE_NETWORK) ?
-                   uct_iface_is_reachable(iface, params->device_addr,
-                                          params->iface_addr) :
-                   uct_iface_is_same_device(iface, params->device_addr);
-}
-
-int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
-                              const uct_iface_is_reachable_params_t *params)
-{
-    const uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
     return iface->internal_ops->iface_is_reachable_v2(tl_iface, params);
 }
 
--- a/src/uct/ib/base/ib_iface.c
+++ b/src/uct/ib/base/ib_iface.c
@@ -486,9 +486,7 @@ void uct_ib_address_unpack(const uct_ib_address_t *ib_addr,
     } else {
         /* Default prefix */
         params.gid.global.subnet_prefix = UCT_IB_LINK_LOCAL_PREFIX;
-        params.gid.global.interface_id  = 0;
-        params.flags                   |= UCT_IB_ADDRESS_PACK_FLAG_SUBNET_PREFIX |
-                                          UCT_IB_ADDRESS_PACK_FLAG_INTERFACE_ID;
+        params.flags                   |= UCT_IB_ADDRESS_PACK_FLAG_SUBNET_PREFIX;
 
         /* If the link layer is not ETHERNET, then it is IB and a lid
          * must be present */
@@ -497,6 +495,7 @@ void uct_ib_address_unpack(const uct_ib_address_t *ib_addr,
         if (ib_addr->flags & UCT_IB_ADDRESS_FLAG_IF_ID) {
             params.gid.global.interface_id =
                     *ucs_serialize_next(&ptr, const uint64_t);
+            params.flags |= UCT_IB_ADDRESS_PACK_FLAG_INTERFACE_ID;
         }
 
         if (ib_addr->flags & UCT_IB_ADDRESS_FLAG_SUBNET16) {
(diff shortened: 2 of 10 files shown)


src/uct/ib/base/ib_iface.c
```diff
@@ -667,6 +667,60 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
+    scope = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params, scope, SCOPE,
+                            UCT_IFACE_REACHABILITY_SCOPE_NETWORK);
+
+    if (scope == UCT_IFACE_REACHABILITY_SCOPE_DEVICE) {
+        uct_ib_address_unpack(ib_addr, &pack_params);
+        return uct_ib_iface_is_same_device(&pack_params, iface);
+    }
+
+    return uct_ib_iface_is_reachable(tl_iface, params->device_addr,
+                                     params->iface_addr);
```
yosefe, line 721: we can call uct_ib_iface_is_reachable anyway. if it fails - it's definitely not same device.
then we can skip some checks in uct_ib_iface_is_same_device

src/uct/ib/base/ib_iface.c
```diff
@@ -667,6 +667,60 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
+        }
+        if (params->flags & UCT_IB_ADDRESS_PACK_FLAG_SUBNET_PREFIX) {
+            res &= (params->gid.global.subnet_prefix ==
+                    gid->global.subnet_prefix);
+        }
+    }
+
+    if (params->flags & UCT_IB_ADDRESS_PACK_FLAG_PKEY) {
+        res &= (params->pkey == iface->pkey);
+    }
```
yosefe, line 699: why need to check pkey?

src/uct/ib/base/ib_iface.c
```diff
@@ -667,6 +667,60 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
+    const union ibv_gid *gid = &iface->gid_info.gid;
+    int res = 1;
+
+    if ((params->flags & UCT_IB_ADDRESS_PACK_FLAG_ETH) !=
+        uct_ib_iface_is_roce(iface)) {
+        return 0;
+    }
+
+    if (params->flags & UCT_IB_ADDRESS_PACK_FLAG_ETH) {
+        res &= !memcmp(gid->raw, params->gid.raw, sizeof(params->gid.raw));
```
yosefe, line 683: bitwise &= is wrong here

src/uct/ib/base/ib_iface.c
```diff
@@ -667,6 +667,60 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
+    } else {
+        res &= (uct_ib_iface_port_attr(iface)->lid == params->lid);
+        if (params->flags & UCT_IB_ADDRESS_PACK_FLAG_INTERFACE_ID &&
+            params->gid.global.interface_id != 0) {
```
yosefe, line 687: why need to check != 0 ?

yosefe (commented):
AFAIR there was a unit test that we could enable now after fixing this?

src/uct/ib/base/ib_iface.c
shasson5: can you explain why?

src/uct/ib/base/ib_iface.c
shasson5: in case we did not pack it in address:
https://github.com/openucx/ucx/blob/f02f1578a3225cdca97148e57048842373fcc9e1/src/uct/ib/base/ib_iface.c#L497

src/uct/ib/base/ib_iface.c
yosefe: it should be logical && with shorthand compute
something like 
if (memcmp(...)) return 0;

src/uct/ib/base/ib_iface.c
yosefe: isn't it checked by `params->flags & UCT_IB_ADDRESS_PACK_FLAG_INTERFACE_ID` ?

src/uct/base/uct_iface.c
```diff
@@ -273,6 +261,19 @@ int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
                               const uct_iface_is_reachable_params_t *params)
 {
     const uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
+
+    if (!ucs_test_all_flags(params->field_mask,
+                            UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR |
+                            UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR)) {
```
brminich, line 267: maybe redefine API so that `iface` and `device` addresses are provided as separate parameters?
they are mandatory anyway..

shasson5: should it be done it this PR or a separated one?

src/uct/ib/base/ib_iface.c
shasson5: seems that this flag is appended always, regardless to the value of `UCT_IB_ADDRESS_FLAG_IF_ID`:

https://github.com/openucx/ucx/blob/f02f1578a3225cdca97148e57048842373fcc9e1/src/uct/ib/base/ib_iface.c#L491

src/uct/base/uct_iface.c
yosefe: question is if we think they will always be mandatory in the future.

src/uct/ib/base/ib_iface.c
yosefe: so let's try to fix the unpack

src/uct/ib/base/ib_iface.c
```diff
@@ -667,58 +666,46 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
-        res &= (uct_ib_iface_port_attr(iface)->lid == params->lid);
-        if (params->flags & UCT_IB_ADDRESS_PACK_FLAG_INTERFACE_ID &&
-            params->gid.global.interface_id != 0) {
-            res &= (params->gid.global.interface_id ==
-                    gid->global.interface_id);
-        }
-        if (params->flags & UCT_IB_ADDRESS_PACK_FLAG_SUBNET_PREFIX) {
-            res &= (params->gid.global.subnet_prefix ==
-                    gid->global.subnet_prefix);
+        same_lid = (uct_ib_iface_port_attr(iface)->lid == params.lid);
```
yosefe, line 682: if (!=) return 0;

src/uct/ib/base/ib_iface.c
```diff
@@ -667,58 +666,46 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
+        if (params.flags & UCT_IB_ADDRESS_PACK_FLAG_INTERFACE_ID) {
+            iface_id_verified = (params.gid.global.interface_id ==
```
yosefe, line 684: if (!=) return 0;

src/uct/ib/base/ib_iface.c
```diff
@@ -670,23 +670,24 @@ static int uct_ib_iface_is_same_device(uct_ib_iface_t *iface,
...
     uct_ib_address_pack_params_t params;
-    int same_lid;
 
     uct_ib_address_unpack(ib_addr, &params);
 
     if (params.flags & UCT_IB_ADDRESS_PACK_FLAG_ETH) {
         return !memcmp(gid->raw, params.gid.raw, sizeof(params.gid.raw));
     } else {
-        same_lid = (uct_ib_iface_port_attr(iface)->lid == params.lid);
+        if (uct_ib_iface_port_attr(iface)->lid != params.lid) {
```
yosefe, line 680: can remove the "else" in line 679

yosefe (approved):
@gleon99 pls take a look as well

src/uct/base/uct_iface.c
gleon99: IMO current impl is good, I won't redefine.

gleon99 (approved):
@shasson5  please squash.

yosefe (approved):

PR #9037, opened 2023-04-26
UCP: update comment before detecting resources

changchengx:
It only checks users configurations simply here. It can't assert that there're existed resources.

## What
Remove comment since the log explains very well.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -1247,8 +1247,6 @@ static void ucp_free_resources(ucp_context_t *context)
 
 static ucs_status_t ucp_check_resource_config(const ucp_config_t *config)
 {
-     /* if we got here then num_resources > 0.
-      * if the user's device list is empty, there is no match */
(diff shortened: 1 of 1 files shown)


src/ucp/core/ucp_context.c
```diff
@@ -1247,21 +1247,19 @@ static void ucp_free_resources(ucp_context_t *context)
 
 static ucs_status_t ucp_check_resource_config(const ucp_config_t *config)
 {
-     /* if we got here then num_resources > 0.
-      * if the user's device list is empty, there is no match */
      if ((0 == config->devices[UCT_DEVICE_TYPE_NET].count) &&
          (0 == config->devices[UCT_DEVICE_TYPE_SHM].count) &&
          (0 == config->devices[UCT_DEVICE_TYPE_ACC].count) &&
          (0 == config->devices[UCT_DEVICE_TYPE_SELF].count)) {
+         /* user's device list is empty, there is no match */
```
yosefe, line 1254: i'd remove the comment altogether since the error message is already self-explanatory

PR #9043, opened 2023-04-27
UCP: update memory registration type info by MD

changchengx:
The current code outputs nothing for the unsupported memory type if there's no MD supports it.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -1592,6 +1592,12 @@ static void ucp_fill_resources_reg_md_map_update(ucp_context_h context)
             context->reg_md_map[mem_type] |= context->dmabuf_reg_md_map;
         }
 
+        if (context->reg_md_map[mem_type] == 0) {
+            ucs_debug("no memory domain supports registering %s memory",
+                      ucs_memory_type_names[mem_type]);
+            continue;
+        }
+
         ucs_string_buffer_reset(&strb);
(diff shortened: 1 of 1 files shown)


src/ucp/core/ucp_context.c
```diff
@@ -1599,6 +1599,10 @@ static void ucp_fill_resources_reg_md_map_update(ucp_context_h context)
             context->reg_md_map[mem_type] |= context->dmabuf_reg_md_map;
         }
 
+        if (context->reg_md_map[mem_type] == 0) {
+            continue;
```
yosefe, line 1603: maybe print that memory registration is not supported by any md?

changchengx: @yosefe I've changed the output info if there's no md supports the specified memory type registration. It has all CI test cases.

src/ucp/core/ucp_context.c
```diff
@@ -1592,6 +1592,12 @@ static void ucp_fill_resources_reg_md_map_update(ucp_context_h context)
             context->reg_md_map[mem_type] |= context->dmabuf_reg_md_map;
         }
 
+        if (context->reg_md_map[mem_type] == 0) {
+            ucs_debug("None md support register %s memory",
```
yosefe, line 1596: "no memory domain supports registering %s memory"

changchengx: Done

yosefe (approved):

PR #9045, opened 2023-04-28
UCT/ROCM: add parameter to control dmabuf usage

edgargabriel:
## What
add a parameter that allows to control whether to use dmabuf in rocm_copy md. The parameter is currently set by default to 'off', i.e. dmabuf will not be used. The plan is to turn it on by default starting from 1.16 release series.

### DIFF
--- a/src/uct/rocm/copy/rocm_copy_md.c
+++ b/src/uct/rocm/copy/rocm_copy_md.c
@@ -35,12 +35,19 @@ static ucs_config_field_t uct_rocm_copy_md_config_table[] = {
      ucs_offsetof(uct_rocm_copy_md_config_t, rcache),
      UCS_CONFIG_TYPE_TABLE(uct_md_config_rcache_table)},
 
+    {"DMABUF", "no",
+     "Enable using cross-device dmabuf file descriptor",
+     ucs_offsetof(uct_rocm_copy_md_config_t, enable_dmabuf),
+     UCS_CONFIG_TYPE_TERNARY},
+
     {NULL}
 };
 
 static ucs_status_t
-uct_rocm_copy_md_query(uct_md_h md, uct_md_attr_v2_t *md_attr)
+uct_rocm_copy_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
 {
+    uct_rocm_copy_md_t *md = ucs_derived_of(uct_md, uct_rocm_copy_md_t);
+
     md_attr->flags                  = UCT_MD_FLAG_REG | UCT_MD_FLAG_NEED_RKEY |
                                       UCT_MD_FLAG_ALLOC;
     md_attr->reg_mem_types          = UCS_BIT(UCS_MEMORY_TYPE_HOST) |
@@ -52,7 +59,7 @@ uct_rocm_copy_md_query(uct_md_h md, uct_md_attr_v2_t *md_attr)
     md_attr->access_mem_types       = UCS_BIT(UCS_MEMORY_TYPE_ROCM);
     md_attr->detect_mem_types       = UCS_BIT(UCS_MEMORY_TYPE_ROCM);
     md_attr->dmabuf_mem_types       = 0;
-    if (uct_rocm_base_is_dmabuf_supported()) {
+    if (md->have_dmabuf) {
         md_attr->dmabuf_mem_types |= UCS_BIT(UCS_MEMORY_TYPE_ROCM);
     }
     md_attr->max_alloc              = SIZE_MAX;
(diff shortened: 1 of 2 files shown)


src/uct/rocm/copy/rocm_copy_md.c
```diff
@@ -35,12 +35,19 @@ static ucs_config_field_t uct_rocm_copy_md_config_table[] = {
      ucs_offsetof(uct_rocm_copy_md_config_t, rcache),
      UCS_CONFIG_TYPE_TABLE(uct_md_config_rcache_table)},
 
+    {"ENABLE_DMABUF", "n",
+     "Enable using cross-device dmabuf file descriptor",
+     ucs_offsetof(uct_rocm_copy_md_config_t, enable_dmabuf),
+     UCS_CONFIG_TYPE_BOOL},
```
brminich, line 41: maybe better use `UCS_CONFIG_TYPE_TERNARY`, so you will set it to `try` in 1.16 (which means that it will be enabled iff it is supported)

edgargabriel: we can do that, I would like to have this pr merged in master first in order to bring it over to the 1.15 branch as well with the default being 'no'.. And we can change after that the default for 1.16

edgargabriel: oh, you mean change it to UCS_CONFIG_TYPE_TERNARY with the default set to 'no' for 1.15, and we change it to 'try' in 1.16?

edgargabriel: done, thanks!

src/uct/rocm/copy/rocm_copy_md.c
```diff
@@ -52,7 +59,8 @@ uct_rocm_copy_md_query(uct_md_h md, uct_md_attr_v2_t *md_attr)
     md_attr->access_mem_types       = UCS_BIT(UCS_MEMORY_TYPE_ROCM);
     md_attr->detect_mem_types       = UCS_BIT(UCS_MEMORY_TYPE_ROCM);
     md_attr->dmabuf_mem_types       = 0;
-    if (uct_rocm_base_is_dmabuf_supported()) {
+    if ((md->enable_dmabuf != UCS_NO) &&
+        (uct_rocm_base_is_dmabuf_supported())) {
```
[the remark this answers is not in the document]
edgargabriel: @brminich thank you, that is a good idea. I added the logic to the md creation to return an error if dmabuf is requested by the user but is not supported by the rocm library.

src/uct/rocm/copy/rocm_copy_md.c
```diff
@@ -35,12 +35,19 @@ static ucs_config_field_t uct_rocm_copy_md_config_table[] = {
      ucs_offsetof(uct_rocm_copy_md_config_t, rcache),
      UCS_CONFIG_TYPE_TABLE(uct_md_config_rcache_table)},
 
+    {"ENABLE_DMABUF", "no",
```
yosefe, line 38: maybe just "DMABUF"

src/uct/rocm/copy/rocm_copy_md.c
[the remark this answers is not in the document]
yosefe: 1. there is extra () here
2. maybe move the logic to md_open and set md->enable_dmabuf to true/false according to both user config and support level on the system.

src/uct/rocm/copy/rocm_copy_md.c
```diff
@@ -408,7 +416,14 @@ uct_rocm_copy_md_open(uct_component_h component, const char *md_name,
     md->super.component = &uct_rocm_copy_component;
     md->rcache          = NULL;
     md->reg_cost        = UCS_LINEAR_FUNC_ZERO;
+    md->enable_dmabuf   = md_config->enable_dmabuf;
 
+    if ((md_config->enable_dmabuf == UCS_YES) &&
+	!(uct_rocm_base_is_dmabuf_supported())) {
+        ucs_error("dmabuf support requested but not found");
```
yosefe, line 423: "ROCm dmabuf support requested but not found"

src/uct/rocm/copy/rocm_copy_md.h
```diff
@@ -32,6 +33,7 @@ typedef struct uct_rocm_copy_md_config {
     uct_md_rcache_config_t      rcache;       /**< Registration cache config */
     ucs_linear_func_t           uc_reg_cost;  /**< Memory registration cost estimation
                                                    without using the cache */
+    int                         enable_dmabuf; /**< Turn using dmabuf on/off */
```
yosefe, line 36: ucs_ternary_value_t

edgargabriel:
@yosefe thanks, incorporated all the change requests/suggestions.

yosefe (approved):

PR #9046, opened 2023-05-01
TOOLS/PERF/PERFTEST: Added a print for window size

roiedanino:
# What
Changed fetch & add, swap and cswap window size to 32

# Why
When running `ucx_perftest` without setting -O, fetch & add, swap and cswap tests are currently waiting for a response after each send, affecting performance heavily compared to the `ib_atomic_bw` benchmark.

### DIFF
--- a/src/tools/perf/perftest_run.c
+++ b/src/tools/perf/perftest_run.c
@@ -144,6 +144,8 @@ static void print_header(struct perftest_context *ctx)
         printf("| Send memory:  %-60s                               |\n", ucs_memory_type_names[ctx->params.super.send_mem_type]);
         printf("| Recv memory:  %-60s                               |\n", ucs_memory_type_names[ctx->params.super.recv_mem_type]);
         printf("| Message size: %-60zu                               |\n", ucx_perf_get_message_size(&ctx->params.super));
+        printf("| Window size:  %-60u                               |\n", ctx->params.super.max_outstanding);
+
         if ((test->api == UCX_PERF_API_UCP) &&
             (test->command == UCX_PERF_CMD_AM)) {
             printf("| AM header size: %-60zu             |\n",


gleon99:
@roiedanino please fix the PR description:
1. "What" & "Why" - according to our convention.
2. changed -> Changed

src/tools/perf/perftest.c
```diff
@@ -42,13 +42,13 @@ test_type_t tests[] = {
...
-     "atomic fetch-and-add latency / rate", "latency", 1},
+     "atomic fetch-and-add latency / rate", "latency", 32},
 
     {"swap", UCX_PERF_API_UCT, UCX_PERF_CMD_SWAP, UCX_PERF_TEST_TYPE_STREAM_UNI,
-     "atomic swap latency / rate", "latency", 1},
+     "atomic swap latency / rate", "latency", 32},
 
     {"cswap", UCX_PERF_API_UCT, UCX_PERF_CMD_CSWAP, UCX_PERF_TEST_TYPE_STREAM_UNI,
-     "atomic compare-and-swap latency / rate", "latency", 1},
+     "atomic compare-and-swap latency / rate", "latency", 32},
```
gleon99, line 51: There are also other atomic operations, not affected by this PR. Is that intentional? Yes / No / explain...

roiedanino: Intentional, those are the only atomic operations that are waiting for responses:
image
image

src/tools/perf/perftest.c
```diff
@@ -42,13 +42,13 @@ test_type_t tests[] = {
      "get latency / bandwidth / message rate", "latency", 1},
 
     {"fadd", UCX_PERF_API_UCT, UCX_PERF_CMD_FADD, UCX_PERF_TEST_TYPE_STREAM_UNI,
-     "atomic fetch-and-add latency / rate", "latency", 1},
+     "atomic fetch-and-add latency / rate", "latency", 32},
```
yosefe, line 45: i'm not sure we should do it. maybe just print the window size in the output?
we still need a test for atomic latency..

yosefe (approved):

PR #9048, opened 2023-05-02
UCP/PROTOV2: Fix wire-compat CI with proto v2

brminich:
## What
Enable proto v2 by default

### DIFF
--- a/buildlib/pr/distro.yml
+++ b/buildlib/pr/distro.yml
@@ -29,22 +29,6 @@ jobs:
           make install
         displayName: Build ucx artifact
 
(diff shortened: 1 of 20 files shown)


changchengx (commented):
@brminich Does UUX1.11 keep compatible with UCX1.16(after merging this PR)?

brminich:
> @brminich Does UCX1.11 keep compatible with UCX1.16(after merging this PR)?

no, it will be compatible with 1.12 and newer

brminich:
replaced by #9188

PR #9052, opened 2023-05-03
UCT/CUDA: Add config for dmabuf support

brminich:
## What
Add configuration to control usage of dmabuf feature for cuda memory transfers

### DIFF
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.c
@@ -50,6 +50,11 @@ static ucs_config_field_t uct_cuda_copy_md_config_table[] = {
      " is registered. Otherwise only the user specified region is registered.",
      ucs_offsetof(uct_cuda_copy_md_config_t, max_reg_ratio), UCS_CONFIG_TYPE_DOUBLE},
 
+    {"DMABUF", "try",
+     "Enable using cross-device dmabuf file descriptor",
+     ucs_offsetof(uct_cuda_copy_md_config_t, enable_dmabuf),
+                  UCS_CONFIG_TYPE_TERNARY},
+
     {NULL}
 };
 
@@ -78,8 +83,10 @@ static int uct_cuda_copy_md_is_dmabuf_supported()
 }
 
 static ucs_status_t
-uct_cuda_copy_md_query(uct_md_h md, uct_md_attr_v2_t *md_attr)
+uct_cuda_copy_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
 {
+    uct_cuda_copy_md_t *md = ucs_derived_of(uct_md, uct_cuda_copy_md_t);
+
     md_attr->flags                  = UCT_MD_FLAG_REG | UCT_MD_FLAG_ALLOC;
     md_attr->reg_mem_types          = UCS_BIT(UCS_MEMORY_TYPE_HOST) |
                                       UCS_BIT(UCS_MEMORY_TYPE_CUDA) |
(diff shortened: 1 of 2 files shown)


src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -50,6 +50,11 @@ static ucs_config_field_t uct_cuda_copy_md_config_table[] = {
      " is registered. Otherwise only the user specified region is registered.",
      ucs_offsetof(uct_cuda_copy_md_config_t, max_reg_ratio), UCS_CONFIG_TYPE_DOUBLE},
 
+    {"DMABUF_ENABLE", "try",
```
yosefe, line 53: "DMABUF"

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -543,13 +548,27 @@ uct_cuda_copy_md_open(uct_component_t *component, const char *md_name,
...
-    md->super.ops              = &md_ops;
-    md->super.component        = &uct_cuda_copy_component;
-    md->config.alloc_whole_reg = config->alloc_whole_reg;
-    md->config.max_reg_ratio   = config->max_reg_ratio;
-    *md_p                      = (uct_md_h)md;
+    md->super.ops               = &md_ops;
+    md->super.component         = &uct_cuda_copy_component;
+    md->config.alloc_whole_reg  = config->alloc_whole_reg;
+    md->config.max_reg_ratio    = config->max_reg_ratio;
+    md->config.dmabuf_supported = uct_cuda_copy_md_is_dmabuf_supported();
```
yosefe, line 555: don't override dmabuf_supported var
see how ROCm is doing it..

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -543,13 +548,27 @@ uct_cuda_copy_md_open(uct_component_t *component, const char *md_name,
...
+        ucs_debug("dmabuf support is disabled by configuration");
+    }
+
+    *md_p = (uct_md_h)md;
 
     return UCS_OK;
+
+err_not_supported:
+    ucs_free(md);
+    return UCS_ERR_UNSUPPORTED;
```
yosefe, line 571: jump here from line 548

src/uct/cuda/cuda_copy/cuda_copy_md.h
```diff
@@ -32,6 +33,7 @@ typedef struct uct_cuda_copy_md_config {
     uct_md_config_t             super;
     ucs_on_off_auto_value_t     alloc_whole_reg;
     double                      max_reg_ratio;
+    ucs_ternary_auto_value_t    enable_dmabuf;
```
yosefe, line 36: dmabuf_enable

src/uct/cuda/cuda_copy/cuda_copy_md.c
brminich: why is this better?

src/uct/cuda/cuda_copy/cuda_copy_md.h
brminich: this is like it is called in RoCM

src/uct/cuda/cuda_copy/cuda_copy_md.c
brminich: it is inconsistent with our other vars which enable something (hwtm, cqe zipping, etc)

src/uct/cuda/cuda_copy/cuda_copy_md.c
yosefe: easier to track the logic when md->config.dmabuf_supported is set to the right value once

src/uct/cuda/cuda_copy/cuda_copy_md.c
brminich: ok, but it will be more code:
- i'll have to introduce `status` variable
- i'll need to have 2 labels (just for exit and for freeing the md)

yosefe: it's more consistent with our style that the function will not have many exit points

yosefe (approved):

yosefe (approved):

PR #9054, opened 2023-05-05
UCT/TCP: output tcp iface device name for debug purpose

changchengx:
## What
output tcp iface device name

### DIFF
--- a/src/uct/tcp/tcp_iface.c
+++ b/src/uct/tcp/tcp_iface.c
@@ -591,9 +591,11 @@ static ucs_status_t uct_tcp_iface_listener_init(uct_tcp_iface_t *iface)
         goto err_close_sock;
     }
 
-    ucs_debug("tcp_iface %p: listening for connections (fd=%d) on %s",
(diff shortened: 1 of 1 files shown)


src/uct/tcp/tcp_iface.c
```diff
@@ -591,9 +591,10 @@ static ucs_status_t uct_tcp_iface_listener_init(uct_tcp_iface_t *iface)
         goto err_close_sock;
     }
 
-    ucs_debug("tcp_iface %p: listening for connections (fd=%d) on %s",
-              iface, iface->listen_fd, ucs_sockaddr_str((struct sockaddr *)&bind_addr,
-                                                       ip_port_str, sizeof(ip_port_str)));
+    ucs_debug("tcp_iface %p on dev %s: listening for connections (fd=%d) on %s",
```
yosefe, line 594: ucs_debug("tcp_iface %p: listening for connections (fd=%d) on %s netif %s",

changchengx: Done

yosefe (approved):

PR #9062, opened 2023-05-09
DOCS: bump up the Python ver - v1.15.x

Alexey-Rivkin:
## What
Bump up the Python version used to build docs and limit the urllib3 ver.

## Why ?
Failure due to package version mismatch.


yosefe:
Tests passed but status was not updated

PR #9064, opened 2023-05-10
UCT/ROCM: check kernel config for dmabuf support

edgargabriel:
## What
The kernel has to have certain options enabled in order for dmabuf support to be fully functional. This commit adds a check whether the kernel has the required options.

### DIFF
--- a/src/uct/rocm/base/rocm_base.c
+++ b/src/uct/rocm/base/rocm_base.c
@@ -11,7 +11,7 @@
 
 #include <ucs/sys/string.h>
 #include <ucs/sys/module.h>
-
+#include <sys/utsname.h>
 #include <pthread.h>
 
 #define MAX_AGENTS 63
@@ -292,9 +292,44 @@ int uct_rocm_base_is_dmabuf_supported()
     int dmabuf_supported = 0;
 
 #if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF
-    dmabuf_supported = 1;
-#endif
+    const char kernel_opt1[] = "CONFIG_DMABUF_MOVE_NOTIFY=y";
+    const char kernel_opt2[] = "CONFIG_PCI_P2PDMA=y";
+    int found_opt1           = 0;
+    int found_opt2           = 0;
+    FILE *fp;
+    struct utsname utsname;
+    char kernel_conf_file[128];
+    char buf[256];
 
+    if (uname(&utsname) == -1) {
+        ucs_trace("could not get kernel name");
+        goto out;
+    }
+
+    ucs_snprintf_safe(kernel_conf_file, sizeof(kernel_conf_file),
+                      "/boot/config-%s", utsname.release);
+    fp = fopen(kernel_conf_file, "r");
+    if (fp == NULL) {
+        ucs_trace("could not open kernel conf file %s error: %m",
+                  kernel_conf_file);
+        goto out;
+    }
+
+    while (fgets(buf, sizeof(buf), fp) != NULL) {
+        if (strstr(buf, kernel_opt1) != NULL) {
+            found_opt1 = 1;
+        }
+        if (strstr(buf, kernel_opt2) != NULL) {
+            found_opt2 = 1;
+        }
+        if (found_opt1 && found_opt2) {
+            dmabuf_supported = 1;
+            break;
+        }
+    }
+    fclose(fp);
+#endif
+out:
     return dmabuf_supported;
 }
 


src/uct/rocm/base/rocm_base.c
```diff
@@ -292,9 +292,43 @@ int uct_rocm_base_is_dmabuf_supported()
...
+    struct utsname utsname;
+    char kernel_conf_file[128];
+    char buf[256];
+
+    if (uname(&utsname) == -1) {
+        ucs_trace("could not get kernel name");
+        goto exit;
+    }
+
+    ucs_snprintf_safe(kernel_conf_file, 128, "/boot/config-%s",
```
yosefe, line 309: seems quite hacky.. isn't there a better way to check it? such as query rocm driver? or the running kernel?
not always the config file exists in that location..

edgargabriel: I chatted with somebody from our kernel team, he was not aware of an easier solution. Furthermore, I did check on a couple of platforms that I have access to, including ubuntu, rhel, and sles systems, and the file did exist in all of them. So unless we have a better idea, I would suggest that we go ahead with this version for now. We can also adjust it later if a new/better solution comes up.

yosefe: Can we try obtaining dmabuf fd from room driver for a dummy allocation and check if it works?

edgargabriel: Unfortunately not. Here is the actual problem: even without these two kernel config options, the runtime will return a valid dmabuf handle that works if you register it with ibv_reg_dmabuf_mr, and as long as the buffer would be used only for this, you would be good. The problem occurs once the same buffer is also registered and used for IPC communication. I am do not fully know what is happening on the runtime/kernel level, but once a dmabuf handle has been created the IPC mechanism doesn't work on that buffer anymore, unless these two kernel option have been enabled.

yosefe: Is this something we can test in this function? i.e creating dmabuf then using the buffer with IPC

edgargabriel: I have testcodes where this being done (although they are multi-node tests, I didn't try whether importing the ipc handle on the same process as it was created on triggers the same problem). I don't think it is however practical to do this here. The problem appears if you import an ipc key *after* calling ibv_reg_dmabuf_mr function for the same buffer. For this I would have to go through the entire chain of required ibv functions since ibv_reg_dmabuf_mr requires a protection domain as an argument  I don't think that is a practical solution, and would in my opinion be a worse abstraction violation than checking for the two kernel config parameters in the rocm component.

src/uct/rocm/base/rocm_base.c
```diff
@@ -292,9 +292,43 @@ int uct_rocm_base_is_dmabuf_supported()
     int dmabuf_supported = 0;
 
 #if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF
-    dmabuf_supported = 1;
-#endif
+    FILE *fp             = NULL;
+    char kernel_opt1[28] = "CONFIG_DMABUF_MOVE_NOTIFY=y";
+    char kernel_opt2[20] = "CONFIG_PCI_P2PDMA=y";
```
yosefe, line 297: these could be `const char*`

src/uct/rocm/base/rocm_base.c
yosefe, line 309: sizeof(kernel_conf_file)

yosefe: ok

edgargabriel:
@yosefe thank you for your review, I think I incorporated all the requested changes.

src/uct/rocm/base/rocm_base.c
```diff
@@ -292,9 +292,44 @@ int uct_rocm_base_is_dmabuf_supported()
     int dmabuf_supported = 0;
 
 #if HAVE_HSA_AMD_PORTABLE_EXPORT_DMABUF
-    dmabuf_supported = 1;
+    const char kernel_opt1[28] = "CONFIG_DMABUF_MOVE_NOTIFY=y";
```
yosefe, line 295: const char kernel_opt1[] = ..

src/uct/rocm/base/rocm_base.c
```diff
@@ -292,9 +292,44 @@ int uct_rocm_base_is_dmabuf_supported()
...
+    char buf[256];
+
+    if (uname(&utsname) == -1) {
+        ucs_trace("could not get kernel name");
+        goto exit;
+    }
+
+    ucs_snprintf_safe(kernel_conf_file, sizeof(kernel_conf_file),
+                      "/boot/config-%s", utsname.release);
+    errno = 0;
```
yosefe, line 311: why needed?

src/uct/rocm/base/rocm_base.c
```diff
@@ -292,9 +292,44 @@ int uct_rocm_base_is_dmabuf_supported()
...
+            found_opt1 = 1;
+        }
+        if (strstr(buf, kernel_opt2) != NULL) {
+            found_opt2 = 1;
+        }
+        if (found_opt1 && found_opt2) {
+            dmabuf_supported = 1;
+            break;
+        }
+    }
```
yosefe, line 330: add fclose() after the loop end

src/uct/rocm/base/rocm_base.c
```diff
@@ -292,9 +292,44 @@ int uct_rocm_base_is_dmabuf_supported()
...
 #endif
-
+exit:
```
yosefe, line 332: exit -> out

src/uct/rocm/base/rocm_base.c
edgargabriel: Some programming guides (e.g. https://www.ibm.com/docs/en/rdfi/9.6.0?topic=errors-checking-errno-value) recommend resetting errno to zero before calling a function that uses errno.

yosefe: This is if errno is used as indication if error happened or not.
In case we already know the function failed - then errno is set to the error value - see https://man7.org/linux/man-pages/man3/fopen.3.html section "ERRORS" for example

edgargabriel: @yosefe ok, sounds good, I removed the line.

yosefe (approved):

PR #9065, opened 2023-05-10
JUCX: Add methods to query local and remote address from endpoint

fruhland:
## What
This PR adds the methods `getLocalAddress()` and `getRemoteAddress()` to `UcpEndpoint`. The implementation is similar to `UcpListener.getAddress()` (query address only one time and cache it in Java space).

## Why ?
There is no other way to get the local/remote address from an endpoint.


bindings/java/src/main/java/org/openucx/jucx/ucp/UcpEndpoint.java
```diff
@@ -8,8 +8,13 @@
 
 import java.io.Closeable;
 import java.nio.ByteBuffer;
+import java.net.InetSocketAddress;
 
 public class UcpEndpoint extends UcxNativeStruct implements Closeable {
+
+    private InetSocketAddress localAddress;
+    private InetSocketAddress remoteAddress;
```
yosefe, line 16: why need to save them on the endpoint?

bindings/java/src/test/java/org/openucx/jucx/UcpEndpointTest.java
```diff
@@ -55,6 +56,27 @@ public void testConnectToListenerByWorkerAddr() {
...
+        UcpListener listener = worker.newListener(listenerParams);
+
+        UcpEndpointParams endpointParams = new UcpEndpointParams()
+                .setPeerErrorHandlingMode()
+                .setSocketAddress(listener.getAddress());
+        UcpEndpoint endpoint = worker.newEndpoint(endpointParams);
+
+        assertNotEquals(0, endpoint.getLocalAddress().getPort());
+        assertNotEquals(0, endpoint.getRemoteAddress().getPort());
+        assertEquals(listener.getAddress().getAddress(), endpoint.getRemoteAddress().getAddress());
```
yosefe, line 77: can we compare to "127.0.0.1" ?

bindings/java/src/main/java/org/openucx/jucx/ucp/UcpEndpoint.java
fruhland: It is not necessary to save them, but this way, they only need to be queried once and not every time `getLocalAddress()` or `getRemoteAddress()` are called. The same is done in `UcpListener` and I wanted to keep my implementation consistent with the rest of the code.

yosefe (approved):

fruhland:
I fixed my JUnit test, so that the CI pipeline passes the JUCX tests, but I don't know, why some of the other tests are still failing.

yosefe (approved):

PR #9069, opened 2023-05-11
UCS/HLIST: Add for_each_safe macro and length function

yosefe:
## Why
Preparation for callbackq scalability improvements
We need to use list because it can be stored as a value in khash table.
ucs_hlist_for_each_safe macro is needed for callback remove-by-filter operation

## How
- ucs_hlist_for_each_safe has to save next pointer and head pointer, because either one can be changed by removing the current list element.


yosefe:
@brminich @tvegas1 can you pls take a look?

src/ucs/datastruct/hlist.h
```diff
@@ -66,6 +66,21 @@ ucs_hlist_is_empty(const ucs_hlist_head_t *head)
 }
 
 
+/**
+ * Get the length of the list.
+ *
+ * @param [in] head   List head to get length of.
```
brminich, line 72: `hlist` ? or rename the param

src/ucs/datastruct/hlist.h
```diff
@@ -239,6 +254,38 @@ ucs_hlist_extract_head(ucs_hlist_head_t *head)
          )
 
 
+/**
+ * Iterate over detached-head list and allow removing the current element
+ * during the iteration.
+ *
+ * @param _elem     Variable to hold the current list element
```
brminich, line 261: missing `.` in the end

test/gtest/ucs/test_datatype.cc
```diff
@@ -261,6 +270,52 @@ UCS_TEST_F(test_datatype, hlist_for_each_extract_if) {
...
+
+        /* Add elements to the list */
+        ucs_hlist_head_t head;
+        ucs_hlist_head_init(&head);
+        for (elem_t &e : v_elems) {
+            ucs_hlist_add_tail(&head, &e.hlist);
+        }
+        EXPECT_EQ(length, ucs_hlist_length(&head));
+
+        /* Radomly remove elements from the list */
```
brminich, line 289:
```suggestion
        /* Randomly remove elements from the list */
```

test/gtest/ucs/test_datatype.cc
```diff
@@ -261,6 +270,52 @@ UCS_TEST_F(test_datatype, hlist_for_each_extract_if) {
     EXPECT_TRUE(v_elems.empty());
 }
 
+UCS_TEST_F(test_datatype, hlist_foreach_safe) {
+    for (int length : {0, 1, 1, 2, 3, 4, 5, 10, 100, 1000}) {
```
tvegas1, line 274: detail: may only one `1,`?

yosefe: because of random in line 300 wanted to test length=1 twice

tvegas1 (dismissed):

yosefe:
@brminich fixed

tvegas1 (approved):

PR #9077, opened 2023-05-16
UCT/IB: Don't create KSM with atomic access just for strict-order access

yosefe:


src/uct/ib/base/ib_md.c
```diff
@@ -559,21 +559,21 @@ static ucs_status_t uct_ib_memh_dereg(uct_ib_md_t *md, uct_ib_mem_t *memh)
 {
     ucs_status_t s, status = UCS_OK;
 
-    if (memh->flags & UCT_IB_MEM_FLAG_ATOMIC_MR) {
-        memh->flags &= ~UCT_IB_MEM_FLAG_ATOMIC_MR;
+    if (memh->atomic_rkey != UCT_IB_INVALID_MKEY) {
```
yosefe, line 562: since UCT_IB_MEM_FLAG_ATOMIC_MR could be set when failed to create KSM

src/uct/ib/base/ib_md.c
```diff
@@ -923,13 +891,53 @@ uct_ib_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
...
+                memh->flags |= UCT_IB_MEM_FLAG_ATOMIC_MR;
+            } else {
+                return status;
+            }
+        }
+
+        extra_rkey = memh->atomic_rkey;
+    } else if (memh->flags & UCT_IB_MEM_FLAG_RELAXED_ORDERING) {
+        /* TODO register strict order key on demand */
+        extra_rkey = memh->strict_order_rkey;
```
yosefe, line 928: relaxed ordering without atomic access

src/ucp/core/ucp_mm.c
```diff
@@ -1247,6 +1249,14 @@ void ucp_mem_print_info(const char *mem_spec, ucp_context_h context,
     fprintf(stream, "\n");
     fprintf(stream, "#\n");
 
+    status = ucp_rkey_pack(context, memh, &rkey_buffer, &rkey_size);
+    if (status != UCS_OK) {
+        printf("<Failed to pack rkey: %s>\n", ucs_status_string(status));
```
brminich, line 1254: should not we use `fprintf(stream` instead?

src/uct/ib/base/ib_md.c
```diff
@@ -559,21 +559,21 @@ static ucs_status_t uct_ib_memh_dereg(uct_ib_md_t *md, uct_ib_mem_t *memh)
...
+    memh->flags &= ~UCT_IB_MEM_FLAG_ATOMIC_MR;
 
     if (memh->flags & UCT_IB_MEM_FLAG_RELAXED_ORDERING) {
         s = uct_ib_memh_dereg_key(md, memh, UCT_IB_MR_STRICT_ORDER);
-        memh->flags &= ~UCT_IB_MEM_FLAG_RELAXED_ORDERING;
         if (s != UCS_OK) {
             status = s;
         }
     }
+    memh->flags &= ~UCT_IB_MEM_FLAG_RELAXED_ORDERING;
```
brminich, line 576: why need to move?

src/uct/ib/base/ib_md.c
```diff
@@ -923,13 +891,53 @@ uct_ib_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
...
+     */
+    if (memh->flags & UCT_IB_MEM_ACCESS_REMOTE_ATOMIC) {
+        if (!(memh->flags & UCT_IB_MEM_FLAG_ATOMIC_MR)) {
+            /* create UMR on-demand */
+            status = UCS_PROFILE_NAMED_CALL_ALWAYS("reg atomic key",
+                                                   md->ops->reg_atomic_key, md,
+                                                   memh);
+            if (status == UCS_OK) {
+                memh->flags |= UCT_IB_MEM_FLAG_ATOMIC_MR;
+                extra_rkey   = memh->atomic_rkey;
```
brminich, line 909: another initialization is on line 925 below

src/uct/ib/base/ib_md.c
```diff
@@ -923,13 +891,53 @@ uct_ib_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
...
+                ucs_trace("created atomic key 0x%x for 0x%x", memh->atomic_rkey,
+                          memh->lkey);
+            } else if (status == UCS_ERR_UNSUPPORTED) {
+                if (flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO) {
+                    /* Must have indirect atomic key to invalidate */
+                    return UCS_ERR_IO_ERROR;
+                }
+
```
brminich, line 917: maybe explicitly set `extra_rkey = UCT_IB_INVALID_MKEY` here for documentation purposes? (instead of line 925)

src/uct/ib/base/ib_md.c
yosefe: to make it consistent with the previous code block

src/uct/ib/base/ib_md.c
yosefe: right - it was also found by static check

yosefe:
@Artemy-Mellanox @brminich can you pls take a look?

Artemy-Mellanox (commented):
Currently we set strict_order mkey as atomic key if KSM is not supported, in https://github.com/openucx/ucx/blob/master/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c#L529.
Maybe would be easier to extend this logic and enter this if in case memh->flags has no UCT_IB_MEM_ACCESS_REMOTE_ATOMIC?

yosefe:
> Currently we set strict_order mkey as atomic key if KSM is not supported, in https://github.com/openucx/ucx/blob/master/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c#L529.
> Maybe would be easier to extend this logic and enter this if in case memh->flags has no UCT_IB_MEM_ACCESS_REMOTE_ATOMIC?

IMO the logic of not registering atomic key if not really needed better be in IB/base part since it's also (theoretically) relevant for Verbs

Artemy-Mellanox:
> > Currently we set strict_order mkey as atomic key if KSM is not supported, in https://github.com/openucx/ucx/blob/master/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c#L529.
> > Maybe would be easier to extend this logic and enter this if in case memh->flags has no UCT_IB_MEM_ACCESS_REMOTE_ATOMIC?
> 
> IMO the logic of not registering atomic key if not really needed better be in IB/base part since it's also (theoretically) relevant for Verbs

verbs already work this way, the only case we miss it is devx

yosefe:
> verbs already work this way, the only case we miss it is devx

If we do the logic in base code we are also avoiding extra function call (for rndv flow)

yosefe:
> > verbs already work this way, the only case we miss it is devx
> 
> If we do the logic in base code we are also avoiding extra function call (for rndv flow)

@Artemy-Mellanox refactored according to your comment

src/ucp/core/ucp_mm.c
```diff
@@ -1247,6 +1249,15 @@ void ucp_mem_print_info(const char *mem_spec, ucp_context_h context,
     fprintf(stream, "\n");
     fprintf(stream, "#\n");
 
+    status = ucp_rkey_pack(context, memh, &rkey_buffer, &rkey_size);
+    if (status != UCS_OK) {
+        fprintf(stream, "<Failed to pack rkey: %s>\n",
```
Artemy-Mellanox, line 1254: in the rest of the function, we print information to stream and errors to stdout - why vice versa here?

yosefe: @Artemy-Mellanox fixed

Issue #8912, opened 2023-03-01
CUDA 12 support for .deb and .rpm packages


abellina:
Currently .deb and .rpm distributed shared libs are linking dynamically against the CUDA 11 runtime, and these so's have the "11" in the name, this is for libuct_cuda.so:

```
	libcudart.so.11.0 => /usr/local/cuda/targets/x86_64-linux/lib/libcudart.so.11.0 (0x00007f8b1d602000)
```

In machines with CUDA 12 installed, this shared library _will not exist_, and so we will fail to link against CUDA. This ticket is to ask for .deb and .rpm packages for CUDA 12.

Note we would want both CUDA 11 and CUDA 12 at least for the time being, eventually dropping CUDA 11.

PR #9079, opened 2023-05-16
AZP/RELEASE: Add CUDA-12 release - v1.14.x

pentschev:
## What
Add CUDA-12 support to the release pipeline.

## Why ?
https://github.com/openucx/ucx/issues/8912


yosefe:
@pentschev can we close this, as replaced by #9080

PR #9082, opened 2023-05-17
UCP/API: Introduce ucp_rkey_build API and add mkey_index to ucp_mem_map parameters

tvegas1:
## What
Introduce new `ucp_rkey_build()` API and add `mkey_index` to `ucp_mem_map()` input parameters.

## Why ?
Needed to implement symmetric keys schemes, where as a consequence rkeys do not have to be exchanged or stored for each endpoint.

## How ?
Add routines and parameters to enable, on one side, the registering of memory with pre-shared parameters, and on the other side, the build of corresponding rkey, using same parameters.

### DIFF
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -392,7 +392,10 @@ enum ucp_mem_map_params_field {
     UCP_MEM_MAP_PARAM_FIELD_MEMORY_TYPE          = UCS_BIT(4),
 
     /** Exported memory handle buffer. */
-    UCP_MEM_MAP_PARAM_FIELD_EXPORTED_MEMH_BUFFER = UCS_BIT(5)
+    UCP_MEM_MAP_PARAM_FIELD_EXPORTED_MEMH_BUFFER = UCS_BIT(5),
+
+    /** Mkey index to use when doing memory registration. */
+    UCP_MEM_MAP_PARAM_FIELD_MKEY_INDEX           = UCS_BIT(6)
 };
 
 /**
@@ -1667,6 +1670,17 @@ typedef struct ucp_mem_map_params {
       * set to NULL by default.
       */
     const void              *exported_memh_buffer;
+
+    /**
+     * Mkey index to use when doing memory registration.
+     * This field is optional.
+     *
+     * This parameter is application defined and it is used to produce memory
+     * handles that will pack into predictable rkeys. With that capability, it
+     * becomes possible to avoid exchanging rkeys, as the remote endpoint can
+     * directly build them, provided that it uses the same mkey index.
+     */
+    uint32_t                mkey_index;
 } ucp_mem_map_params_t;
 
 
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -410,6 +410,22 @@ enum ucp_mem_advise_params_field {
 
 
 /**
+ * @ingroup UCP_MEM
+ * @brief UCP rkey build parameters field mask.
+ *
+ * The enumeration allows specifying which fields in @ref ucp_rkey_build_params_t
+ * are present. It enables backward compatibility support when adding fields.
+ */
+enum ucp_rkey_build_params_field {
+    /** Memory handle already registered locally as reference */
+    UCP_RKEY_BUILD_PARAM_FIELD_MEMH       = UCS_BIT(0),
+
+    /** Application defined mkey index used to build a given rkey */
+    UCP_RKEY_BUILD_PARAM_FIELD_MKEY_INDEX = UCS_BIT(1)
+};
+
+
+/**
  * @ingroup UCP_CONTEXT
  * @brief UCP library attributes field mask.
  *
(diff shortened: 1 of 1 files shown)


src/ucp/api/ucp.h
```diff
@@ -1651,9 +1670,56 @@ typedef struct ucp_mem_map_params {
...
+     * It is used to identify the memory domain where we need to perform rkey
+     * build.
+     */
+    const ucp_mem_h memh;
+
+    /**
+     * Mkey index to use when locally building an rkey.
+     * This field is mandatory and @ref ucp_rkey_build routine will return an
+     * error if it is missing.
+     * The remote doing the corresponding memory registration must use the same
```
brminich, line 1716: `The remote peer`

src/ucp/api/ucp.h
```diff
@@ -1651,9 +1670,56 @@ typedef struct ucp_mem_map_params {
...
+     * compatibility with respect to adding new fields.
+     */
+    uint64_t        field_mask;
+
+    /**
+     * Memory handle for memory already registered locally with same mkey index.
+     * This field is mandatory and ucp_rkey_build routine will return an error
+     * if it is missing.
+     * It is used to identify the memory domain where we need to perform rkey
+     * build.
```
brminich, line 1708: i would avoid this phrase as memory domain is not a UCP API term

brminich:
@yosefe, @shamisp, @artpol84 
pls take a look

src/ucp/api/ucp.h
```diff
@@ -1651,9 +1670,56 @@ typedef struct ucp_mem_map_params {
...
     const void              *exported_memh_buffer;
+
+    /**
+     * Mkey index to use when doing memory registration.
+     * This field is optional.
+     *
+     * This parameter is application defined and it is used to produce memory
+     * handles that will pack into predictable rkeys. With that capability, it
+     * becomes possible to avoid exchanging rkeys, as the remote endpoint can
+     * directly build them, provided that it uses the same mkey index.
```
artpol84, line 1681:
```suggestion
     * This parameter is application-defined and is used to produce memory
     * handles for which rkeys can be computed remotely without requiring 
     * any data exchange with the owner of the memory region.
     * Knowing the logical Mkey index, the remote endpoint will be able
     * to directly (locally) build the corresponding rkey.
```

artpol84 (commented):
I left a few suggestions for the content of the descriptions.
The code looks good.

src/ucp/api/ucp.h
yosefe: is it going to be supported on all system types?
what happens if the underlying layer does not support it?

src/ucp/api/ucp.h
```diff
@@ -1651,9 +1670,56 @@ typedef struct ucp_mem_map_params {
...
+    uint64_t        field_mask;
+
+    /**
+     * Memory handle for memory already registered locally with same mkey index.
+     * This field is mandatory and ucp_rkey_build routine will return an error
+     * if it is missing.
+     * It is used to identify the memory domain where we need to perform rkey
+     * build.
+     */
+    const ucp_mem_h memh;
```
yosefe, line 1710: why const?

src/ucp/api/ucp.h
```diff
@@ -1651,9 +1670,56 @@ typedef struct ucp_mem_map_params {
...
+
+    /**
+     * Mkey index to use when locally building an rkey.
+     * This field is mandatory and @ref ucp_rkey_build routine will return an
```
yosefe, line 1714: if it's mandatory, it should not be a part of the optional parameters structure

src/ucp/api/ucp.h
artpol84: Good point!
I think that the target system should get ‘ERR_NOT_SUPPORTED’ when doing ucp_mem_map.
The initiator side should probably negotiate with the target if the logical mkey indexing is supported.

artpol84: Also it is possible that while logical mkey indexing is supported, the app will run out of resource.

At the device level we have a way to query the amount of resources available. I think we may want to expose that on the target side in a form of the number of available logical mkey indexes. In UCP it’s complicated be the fact that multiple transports are available and some may support this and some not.
I’m not sure what’s the best way to communicate that. Maybe query worker for the inter-node-only capabilities (for which this feature is mainly useful) to see how many logical mkey indexes the networking transport provides.

shamisp: I don't like the "unsupported" approached because it is going against core design principles of UCX - write code once and it works everywhere.  This is where this API proposal breaks apart IMHO, because it changes the entire workflow of the application, like exchanging RKEYS, etc.

src/ucp/api/ucp.h
```diff
@@ -3106,6 +3158,29 @@ ucs_status_t ucp_ep_rkey_unpack(ucp_ep_h ep, const void *rkey_buffer,
...
 
+/**
+ * @ingroup UCP_MEM
+ * @brief Locally build a remote key using known applicative pre-shared
+ * parameters.
+ *
+ * This routine builds an rkey to access remote memory regions on endpoints that
+ * have registered them using the same pre-shared parameters.
+ *
+ * When some nodes are implementing such symmetric scheme where they
```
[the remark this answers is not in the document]
shamisp: it is too low-level for UCX. The entire workflow of application changes.

shamisp:
In current design we don't have a way to prevent rkey exchange. rkey always have to exchanged since we don't want to introduce branches in a main flow. In some cases it might be "empty" but it is always changed. This is why I think this proposal does not work well.

I suggest to present the feature at UCX call and have RFC discussion.

shamisp (changes requested):
more discussion required

src/ucp/api/ucp.h
artpol84: I don't think it changes that. At the rkey exchange step you'd exchange with other peers the fact that rkey's can be generated locally and no exchange is needed

artpol84:
I’d like to also emphasize that the old way of managing Rkeys is still available for the apps that do not benefit from the feature or not yet updated to use it.
The branch would only be required on the control path (during rkey exchange).
The rkeys that are “ucp_unpack”’ed or “usp_built” feel and work exactly the same, so no data path modifications required

shamisp:
> I’d like to also emphasize that the old way of managing Rkeys is still available for the apps that do not benefit from the feature or not yet updated to use it.
> The branch would only be required on the control path (during rkey exchange).
> The rkeys that are “ucp_unpack”’ed or “usp_built” feel and work exactly the same, so no data path modifications required

This is a precedent in UCX that we have not had before. We have APIs that works for all interconnects and does not require branches to check if feature is supported or not. I don't see how this behavior can be easily simulated in software.  Essentially we would have to develop the same infra as the one you have in SHMEM.

PR #9080, opened 2023-05-17
AZP/RELEASE: Add CUDA-12 release

Alexey-Rivkin:
## What
Porting https://github.com/openucx/ucx/pull/8996

PR #9087, opened 2023-05-17
V1.14.1 rc2 news update

shamisp:
Preparing for RC2

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,10 @@
 ### Features:
 ### Bugfixes:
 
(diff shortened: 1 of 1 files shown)


shamisp:
@pentschev please take a look

pentschev (approved):
LGTM, thanks @shamisp . I see https://github.com/openucx/ucx/pull/9080 is in, but I presume we still want to wait for https://github.com/openucx/ucx/pull/9081, is that right?

yosefe (approved):

PR #9094, opened 2023-05-22
TEST/IODEMO: do not use deprecated APIs

changchengx:
## What
Do not use deprecated APIs in io_demo.

### DIFF
--- a/test/apps/iodemo/ucx_wrapper.cc
+++ b/test/apps/iodemo/ucx_wrapper.cc
@@ -87,6 +87,8 @@ UcxLog::~UcxLog()
 
 #define UCX_LOG UcxLog("[UCX]", true)
 
+ucp_request_param_t UcxContext::recv_param;
+
 UcxContext::UcxAcceptCallback::UcxAcceptCallback(UcxContext &context,
                                                  UcxConnection &connection) :
     _context(context), _connection(connection)
@@ -191,6 +193,10 @@ bool UcxContext::init(const char *name)
     }
 
     UCX_LOG << "created worker " << _worker;
+
+    recv_param.op_attr_mask = UCP_OP_ATTR_FIELD_CALLBACK |
+                              UCP_OP_ATTR_FLAG_NO_IMM_CMPL;
+    recv_param.cb.recv      = (ucp_tag_recv_nbx_callback_t)iomsg_recv_callback;
 
     if (_use_am) {
         set_am_handler(am_recv_callback, this);
@@ -529,11 +535,10 @@ UcxContext::wait_completion(ucs_status_ptr_t status_ptr, const char *title,
 
 void UcxContext::recv_io_message()
 {
-    ucs_status_ptr_t status_ptr = ucp_tag_recv_nb(_worker, &_iomsg_buffer[0],
-                                                  _iomsg_buffer.size(),
-                                                  ucp_dt_make_contig(1),
-                                                  IOMSG_TAG, IOMSG_TAG,
-                                                  iomsg_recv_callback);
+    ucs_status_ptr_t status_ptr = ucp_tag_recv_nbx(_worker, &_iomsg_buffer[0],
+                                                   _iomsg_buffer.size(),
+                                                   IOMSG_TAG, IOMSG_TAG,
+                                                   &recv_param);
     assert(status_ptr != NULL);
     _iomsg_recv_request = reinterpret_cast<ucx_request*>(status_ptr);
 }
@@ -955,7 +960,7 @@ bool UcxConnection::send_am(const void *meta, size_t meta_length,
     ucp_request_param_t param;
     param.op_attr_mask = UCP_OP_ATTR_FIELD_CALLBACK  |
                          UCP_OP_ATTR_FIELD_FLAGS;
-    param.cb.send      = common_request_callback_nbx;
+    param.cb.send      = (ucp_send_nbx_callback_t)common_request_callback;
     param.flags        = UCP_AM_SEND_FLAG_REPLY;
     param.datatype     = 0; // make coverity happy
     if (memh) {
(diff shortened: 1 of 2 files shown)


test/apps/iodemo/ucx_wrapper.cc
```diff
@@ -529,11 +529,19 @@ UcxContext::wait_completion(ucs_status_ptr_t status_ptr, const char *title,
...
 void UcxContext::recv_io_message()
 {
-    ucs_status_ptr_t status_ptr = ucp_tag_recv_nb(_worker, &_iomsg_buffer[0],
-                                                  _iomsg_buffer.size(),
-                                                  ucp_dt_make_contig(1),
-                                                  IOMSG_TAG, IOMSG_TAG,
-                                                  iomsg_recv_callback);
+    ucp_request_param_t param;
+
+    param.op_attr_mask = UCP_OP_ATTR_FIELD_DATATYPE |
```
yosefe, line 534: no need to pass datatype for ucp_dt_make_contig(1)

test/apps/iodemo/ucx_wrapper.cc
```diff
@@ -1131,8 +1146,11 @@ void UcxConnection::connect_tag(UcxCallback *callback)
     }
 
     // send local connection id
-    void *sreq = ucp_stream_send_nb(_ep, &_conn_id, 1, dt_int,
-                                    common_request_callback, 0);
+    param.cb.send = common_request_callback_nbx;
```
yosefe, line 1149: can we remove the existing common_request_callback, and rename common_request_callback_nbx to common_request_callback?

changchengx:
I'll deal with the comments after June-09.

test/apps/iodemo/ucx_wrapper.cc
```diff
@@ -529,11 +529,17 @@ UcxContext::wait_completion(ucs_status_ptr_t status_ptr, const char *title,
 
 void UcxContext::recv_io_message()
 {
-    ucs_status_ptr_t status_ptr = ucp_tag_recv_nb(_worker, &_iomsg_buffer[0],
-                                                  _iomsg_buffer.size(),
-                                                  ucp_dt_make_contig(1),
-                                                  IOMSG_TAG, IOMSG_TAG,
-                                                  iomsg_recv_callback);
+    ucp_request_param_t param;
```
yosefe, line 532: can we make param a global var (static class member) so it will not consume stack space?

changchengx: Sounds good. I'll do it at next weekend.

test/apps/iodemo/ucx_wrapper.cc
```diff
@@ -1243,7 +1242,7 @@ bool UcxConnection::send_common(const void *buffer, size_t length,
 
     assert(_ucx_status == UCS_OK);
 
-    ucp_request_param_t param;
+    ucp_request_param_t &param = _context.param;
```
yosefe, line 1245: we should make a global param var only for params that could be initialized statically (once) and not changed later. and make a different param var for each case.
This way we can pre-initialize param struct and then just use it during send/recv.
It's not possible when some param fields depend on send function parameters - for example, when passing memh.
Also, it's not needed for slow-path case such as stream operation for exchanging conn_id.

changchengx: @yosefe Could we define class type and constructor to initialize ucp_request_param_t in c++ program? I find it must initialize all the structure members one by one explicitly under ```__cplusplus 199711L```. For example, it needs to initialize all the members until ```.cb.recv``` if we only want to initialize ```op_attr_mask``` and ```cb.recv``` explicitly.

yosefe: @changchengx can make these structs static member variables and initialize them like global variables.

changchengx: 1. For slow path(such as ```connect_tag``` which uses ```stream recv/send```), keep the original code there
2. For the ```ucp_request_param_t type object``` that depends on function parameter(such as ```memh```), keep the original code there.
3. Define ```static ucp_request_param_t UcxContext::param``` object and initialize the object in ```UcxContext::init``` function. Then, use this object in ```UcxContext::recv_io_message``` function without constructing it again.

changchengx: @yosefe Passed all CI test cases.

test/apps/iodemo/ucx_wrapper.cc
```diff
@@ -136,6 +136,7 @@ UcxContext::~UcxContext()
     }
 }
 
+ucp_request_param_t UcxContext::param;
```
yosefe, line 139: 1. pls rename it to recv_param
2. space line after
3. move it to just before line 90

changchengx: @yosefe Done. Waiting for CI result.

test/apps/iodemo/ucx_wrapper.cc
```diff
@@ -529,11 +535,11 @@ UcxContext::wait_completion(ucs_status_ptr_t status_ptr, const char *title,
...
-    ucs_status_ptr_t status_ptr = ucp_tag_recv_nb(_worker, &_iomsg_buffer[0],
-                                                  _iomsg_buffer.size(),
-                                                  ucp_dt_make_contig(1),
-                                                  IOMSG_TAG, IOMSG_TAG,
-                                                  iomsg_recv_callback);
+    ucs_status_ptr_t status_ptr = ucp_tag_recv_nbx(_worker, &_iomsg_buffer[0],
+                                                   _iomsg_buffer.size(),
+                                                   IOMSG_TAG, IOMSG_TAG,
+                                                   &recv_param);
+
```
yosefe, line 542: remove space line

changchengx: Done

yosefe (approved):

yosefe:
@changchengx pls squash

changchengx:
@yosefe squashed

yosefe (approved):

PR #9098, opened 2023-05-23
UCP/UCT: Use the new callbackq oneshot API

yosefe:
## Why
Improve scalability of connection establishment and error handling

## How
Use the new callbackq oneshot API, to remove callbacks by key instead of going over the whole callbacks list.

Replaces #8812

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -485,7 +485,9 @@ static int ucp_ep_remove_filter(const ucs_callbackq_elem_t *elem, void *arg)
 
 void ucp_ep_destroy_base(ucp_ep_h ep)
 {
+    ucp_worker_h worker = ep->worker;
     ucp_ep_peer_mem_data_t data;
+
     ucp_ep_refcount_field_assert(ep, refcount, ==, 0);
     ucp_ep_refcount_assert(ep, create, ==, 0);
     ucp_ep_refcount_assert(ep, flush, ==, 0);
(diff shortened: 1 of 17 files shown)


yosefe:
@evgeny-leksikov @brminich can you pls take a look?

brminich (commented):
1. Imo, it is worth to create UCT worker progress wrappers rather than explicitly using callbackq 
2. Maybe we need to set `UCP_WORKER_THREAD_CS_CHECK_IS_BLOCKED(worker);` in the functions where `register_safe` and `unregister_safe` were previously used

yosefe:
> * Imo, it is worth to create UCT worker progress wrappers rather than explicitly using callbackq
> * Maybe we need to set `UCP_WORKER_THREAD_CS_CHECK_IS_BLOCKED(worker);` in the functions where `register_safe` and `unregister_safe` were previously used

1. It means need to update UCT API, I'd prefer to avoid this, since we already expose callbackq directly. It will also avoid extra function call
2. Why needed? one-shot callback add/remove function are thread-safe

brminich:
> 
> 1. It means need to update UCT API, I'd prefer to avoid this, since we already expose callbackq directly. It will also avoid extra function call

This way we are managing the same entity with different APIs, which seems to be quite confusing

> 2. Why needed? one-shot callback add/remove function are thread-safe

I was wondering about the flow when `cbq` is being destroyed at the same time  (not sure there is such use-case though)

brminich (approved):

PR #9110, opened 2023-05-29
UCT/IB/MLX5: Improve ODP support output

yosefe:
## Why
Improve log message

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -720,28 +720,34 @@ static ucs_mpool_ops_t uct_ib_mlx5_dbrec_ops = {
     .obj_str       = NULL
 };
 
-static ucs_status_t
-uct_ib_mlx5_devx_check_odp(uct_ib_mlx5_md_t *md,
-                           const uct_ib_md_config_t *md_config, void *cap)
+static void uct_ib_mlx5_devx_check_odp(uct_ib_mlx5_md_t *md,
+                                       const uct_ib_md_config_t *md_config,
+                                       void *cap)
 {
     char out[UCT_IB_MLX5DV_ST_SZ_BYTES(query_hca_cap_out)] = {};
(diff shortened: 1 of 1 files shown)


src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -759,7 +759,7 @@ uct_ib_mlx5_devx_check_odp(uct_ib_mlx5_md_t *md,
                           capability.odp_cap.mem_page_fault)) {
         odp = UCT_IB_MLX5DV_ADDR_OF(query_hca_cap_out, out,
                 capability.odp_cap.memory_page_fault_scheme_cap);
-        scheme = "memory";
+        scheme = "remote-access";
```
Artemy-Mellanox, line 762: What is the meaning of the change?

yosefe: Before:
`mlx5_0: ODP transport is enabled` or `mlx5_0: ODP memory is enabled` 
These messages are confusing because there is no such thing as ODP transport type (there are RC/DC/UD transport types), and also we don't really enable ODP here (because it's enabled by setting UCP param of which memory types to use nonblock registration)
After:
`mlx5_0: remote-access ODP is supported` or `mlx5_0: full ODP is supported`

Artemy-Mellanox: Maybe ODPv1/ODPv2?

yosefe: we also enable full odp with odpv1 if qp is not created by devx

brminich (approved):

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
Artemy-Mellanox: yes, we don't have other options - full or disabled

yosefe: @Artemy-Mellanox fixed

PR #9117, opened 2023-06-04
AZP: Fix ROCM building and testing

Alexey-Rivkin:
## What

1. Add volume mapping for ROCM container.
2. Skip building CUDA with ROCM.

## Why ?
Failure to load env modules served as a canary, unearthing a bigger problem of dysfunctional ROCM job.

### DIFF
--- a/buildlib/tools/builds.sh
+++ b/buildlib/tools/builds.sh
@@ -201,6 +201,11 @@ build_ugni() {
 # Build CUDA
 #
 build_cuda() {
+	if [[ $CONTAINER == *"rocm"* ]]; then
+		echo "==== Not building with cuda flags ===="
+		return
+	fi
+
 	if az_module_load $CUDA_MODULE
 	then
 		if az_module_load $GDRCOPY_MODULE
--- a/buildlib/pr/main.yml
+++ b/buildlib/pr/main.yml
@@ -176,7 +176,7 @@ resources:
       options: $(DOCKER_OPT_ARGS) $(DOCKER_OPT_VOLUMES) $(DOCKER_OPT_GPU)
     - container: ubuntu2004_rocm_5_4_0
       image: rdmz-harbor.rdmz.labs.mlnx/ucx/x86_64/ubuntu2004:rocm_5_4_0
-      options: $(DOCKER_OPT_ARGS)
+      options: $(DOCKER_OPT_ARGS) $(DOCKER_OPT_VOLUMES)
 
 stages:
   - stage: Codestyle


buildlib/tools/builds.sh
```diff
@@ -201,31 +201,33 @@ build_ugni() {
...
-		$MAKEP
+			az_module_unload $CUDA_MODULE
 
-		az_module_unload $CUDA_MODULE
-
-		echo "==== Running test_link_map with cuda build but no cuda module ===="
-		env UCX_HANDLE_ERRORS=bt ./test/apps/test_link_map
-	else
-		echo "==== Not building with cuda flags ===="
+			echo "==== Running test_link_map with cuda build but no cuda module ===="
```
artemry-nv, line 226: Check formatting

artemry-nv (approved):

buildlib/tools/builds.sh
```diff
@@ -201,31 +201,33 @@ build_ugni() {
 # Build CUDA
 #
 build_cuda() {
-	if az_module_load $CUDA_MODULE
-	then
-		if az_module_load $GDRCOPY_MODULE
+	if [ ! -f /opt/rocm/bin/rocminfo ]; then
```
yosefe, line 204: maybe check the presence of /usr/lib/cuda or something that indicates we have cuda on the container?
alternatively, define something like "build_cuda: no" for ubuntu2004_rocm job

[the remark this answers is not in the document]
Alexey-Rivkin: Part of the process - a plugin does it on the fly.

yosefe: > I followed the existing logic:
> https://github.com/openucx/ucx/blob/master/buildlib/tools/builds.sh#L236
> Should we change there too?

No: 
build_rocm is checking for rocm files
build_cuda should should check for cuda files

buildlib/tools/builds.sh
```diff
@@ -201,31 +201,33 @@ build_ugni() {
 # Build CUDA
 #
 build_cuda() {
-	if az_module_load $CUDA_MODULE
-	then
-		if az_module_load $GDRCOPY_MODULE
+	if [[ ! $CONTAINER == *"rocm"* ]]; then
```
yosefe, line 204: 1. can we check something related to Cuda? it's better that each build func will check its own requirements and not that the "other" requirements are present.
2. better exit the function if the condition is not met, to reduce indentation level

Alexey-Rivkin: Sounds reasonable. However, there is no Cuda preinstalled on those images:
https://github.com/openucx/ucx/blob/master/buildlib/dockers/centos7.Dockerfile#L5-L31

yosefe (approved):
pls squash

yosefe (approved):

artemry-nv (approved):

PR #9119, opened 2023-06-06
README: Fix rdma-core note placement

yosefe:
## Why ?
Fix rdma-core note so it would not look like it's under "TCP" transport

Before:
image

After:
image


marinav123 (approved):

yosefe:

shamisp (approved):

PR #9120, opened 2023-06-07
TOOLS/PERF: Enhance multi-thread test output

yosefe:
## Why
Fix med output lines when running ucx_perftest with multiple threads

## How
Build the output line in a string buffer and then print it at once

### DIFF
--- a/src/tools/perf/perftest_run.c
+++ b/src/tools/perf/perftest_run.c
@@ -25,6 +25,7 @@ void print_progress(char **test_names, unsigned num_names,
                     unsigned flags, int final, int is_server,
                     int is_multi_thread)
 {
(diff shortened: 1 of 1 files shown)


src/tools/perf/perftest_run.c
```diff
@@ -25,6 +25,7 @@ void print_progress(char **test_names, unsigned num_names,
                     unsigned flags, int final, int is_server,
                     int is_multi_thread)
 {
+    UCS_STRING_BUFFER_ONSTACK(strb, 256);
```
brminich, line 28: will 256 be always sufficient? Maybe 1024?

yosefe: it's supposed to be one line of output, shouldn't be that long

brminich (approved):

PR #9122, opened 2023-06-07
UCP/MM: Remove context param from ucp_memh_put()

yosefe:
## Why
Optimization - no need to pass context parameter to ucp_memh_put() since it's already a field in memh


yosefe:

PR #9123, opened 2023-06-07
UCP/MM: Add mem type assertion and logging

yosefe:
## Why
Catch error of trying to register an unsupported memory type


yosefe:

Artemy-Mellanox (approved):

PR #9125, opened 2023-06-08
TEST/COMMON: Add env var to set log level just before test run

yosefe:
## Why
Show logs only during test run and not during initialization

### DIFF
--- a/test/gtest/common/main.cc
+++ b/test/gtest/common/main.cc
@@ -55,6 +55,14 @@ static void modify_config_for_valgrind(const char *name, const char *value)
     }
 }
 
+static void set_log_level()
+{
+    const char *log_level = getenv("GTEST_LOG_LEVEL");
+    if (log_level != NULL) {
+        ucs_global_opts_set_value("LOG_LEVEL", log_level);
+    }
+}
+
 int main(int argc, char **argv) {
     // coverity[fun_call_w_exception]: uncaught exceptions cause nonzero exit anyway, so don't warn.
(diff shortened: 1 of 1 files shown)


test/gtest/common/main.cc
```diff
@@ -55,6 +55,14 @@ static void modify_config_for_valgrind(const char *name, const char *value)
     }
 }
 
+static void set_log_level()
+{
+    const char *log_level = getenv("GTEST_LOG_LEVEL");
+    if (log_level != NULL) {
+        ucs_global_opts_set_value("LOG_LEVEL", log_level);
```
gleon99, line 62: Check return code? If no such log level...

yosefe: In such case, the parse will print an error message anyway:
```
Making all in test/mpi
[     INFO ] Using random seed of 5551
[1686246831.957229] [hpchead:25271:0]          parser.c:1126 UCX  ERROR Invalid value for LOG_LEVEL: 'dasdsadsa'. Expected: [FATAL|ERROR|WARN|DIAG|INFO|DEBUG|TRACE|REQ|DATA|ASYNC|FUNC|POLL]
Note: Google Test filter = *ucs*
Note: Randomizing tests' orders with a seed of 78149 .
[==========] Running 41 tests from 17 test suites.
[----------] 1 test from dc_mlx5/test_uct_pending
...
```

yosefe:
@gleon99 ok to merge?

gleon99 (approved):

PR #9128, opened 2023-06-08
UCT/CUDA/GDR_COPY: Estimated BW of get operation is dedicated.

rakhmets:
## What
Changed the performance estimation of gdr_copy UCT interface, switching the bandwidth of get operation from shared to dedicated.

## Why ?
Experiments showed that the bandwidth of gdr_copy get operation remained maximum when simultaneously running the number of processes equal to the number of GPUs on the node.
The change of the performance estimation of gdr_copy UCT interface enhanced the protocol selection logic, preventing switching from Eager to Rendezvous protocol too early.

### DIFF
--- a/src/uct/cuda/gdr_copy/gdr_copy_iface.c
+++ b/src/uct/cuda/gdr_copy/gdr_copy_iface.c
@@ -104,22 +104,26 @@ static ucs_status_t
 uct_gdr_copy_estimate_perf(uct_iface_h iface, uct_perf_attr_t *perf_attr)
 {
     if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_BANDWIDTH) {
-        perf_attr->bandwidth.dedicated = 0;
         if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_OPERATION) {
             switch (perf_attr->operation) {
             case UCT_EP_OP_GET_SHORT:
             case UCT_EP_OP_GET_ZCOPY:
-                perf_attr->bandwidth.shared = 250.0 * UCS_MBYTE;
+                perf_attr->bandwidth.dedicated = 250.0 * UCS_MBYTE;
+                perf_attr->bandwidth.shared    = 0;
                 break;
             case UCT_EP_OP_PUT_SHORT:
-                perf_attr->bandwidth.shared = 10200.0 * UCS_MBYTE;
+                perf_attr->bandwidth.dedicated = 0;
+                perf_attr->bandwidth.shared    = 10200.0 * UCS_MBYTE;
                 break;
             default:
-                perf_attr->bandwidth.shared =
+                perf_attr->bandwidth.dedicated = 0;
+                perf_attr->bandwidth.shared    =
                         UCT_GDR_COPY_IFACE_DEFAULT_BANDWIDTH;
             }
         } else {
-            perf_attr->bandwidth.shared = UCT_GDR_COPY_IFACE_DEFAULT_BANDWIDTH;
+            perf_attr->bandwidth.dedicated = 0;
+            perf_attr->bandwidth.shared    =
+                    UCT_GDR_COPY_IFACE_DEFAULT_BANDWIDTH;
         }
     }
 


src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -109,7 +109,8 @@ uct_gdr_copy_estimate_perf(uct_iface_h iface, uct_perf_attr_t *perf_attr)
             switch (perf_attr->operation) {
             case UCT_EP_OP_GET_SHORT:
             case UCT_EP_OP_GET_ZCOPY:
-                perf_attr->bandwidth.shared = 250.0 * UCS_MBYTE;
+                perf_attr->bandwidth.dedicated = 250.0 * UCS_MBYTE;
```
brminich, line 112: Assuming we do not have many processes per GPU, we may apply this change to the rest of ops here

yosefe: actually, put bw may be limited by PCIe bw that its shared. Since it's higher that get bw. need to test it..

rakhmets: `ucx_perftest -t put_bw` showed that bandwidth of operation decreases when using more than one process per node, even if the processes use different GPUs. That is why I changed `shared` to `dedicated` only for get operation.

yosefe: ok
maybe we should move line 107 into each of the cases for `switch (perf_attr->operation)`?

rakhmets: > ok maybe we should move line 107 into each of the cases for `switch (perf_attr->operation)`?

done

yosefe (approved):

brminich (dismissed):

src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -104,21 +104,24 @@ static ucs_status_t
...
+                perf_attr->bandwidth.shared    = 0;
                 break;
             case UCT_EP_OP_PUT_SHORT:
-                perf_attr->bandwidth.shared = 10200.0 * UCS_MBYTE;
+                perf_attr->bandwidth.dedicated = 0;
+                perf_attr->bandwidth.shared    = 10200.0 * UCS_MBYTE;
                 break;
             default:
+                perf_attr->bandwidth.dedicated = 0;
```
brminich, line 119: minor: align with the line below

src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -104,21 +104,24 @@ static ucs_status_t
...
                 perf_attr->bandwidth.shared =
                         UCT_GDR_COPY_IFACE_DEFAULT_BANDWIDTH;
             }
         } else {
+            perf_attr->bandwidth.dedicated = 0;
```
brminich, line 124: minor: align with the line below

src/uct/cuda/gdr_copy/gdr_copy_iface.c
rakhmets: This alignment was suggested by CI's 'format code'.

src/uct/cuda/gdr_copy/gdr_copy_iface.c
brminich: that's the problem with our CI code style check which is not solved yet
I'd change it acording to our style anyway

src/uct/cuda/gdr_copy/gdr_copy_iface.c
rakhmets: Reverted 1b9790d4d6e2e7044bdf231eb01ea14ab756aebf

src/uct/cuda/gdr_copy/gdr_copy_iface.c
rakhmets: Reverted 1b9790d4d6e2e7044bdf231eb01ea14ab756aebf

yosefe (approved):

brminich (approved):

PR #9132, opened 2023-06-09
UCS/RCACHE: Disable atfork hooks before forking and attaching a debugger

yosefe:
## Why
Fix deadlock when forked debugger is attached during an error in rcache operation

### DIFF
--- a/src/ucs/memory/rcache.c
+++ b/src/ucs/memory/rcache.c
@@ -1153,6 +1153,11 @@ static void ucs_rcache_global_list_remove(ucs_rcache_t *rcache)
     ucs_async_pipe_destroy(&pipe);
 }
 
+void ucs_rcache_atfork_disable()
+{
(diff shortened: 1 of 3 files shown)


src/ucs/memory/rcache.c
```diff
@@ -1153,6 +1153,11 @@ static void ucs_rcache_global_list_remove(ucs_rcache_t *rcache)
     ucs_async_pipe_destroy(&pipe);
 }
 
+void ucs_rcache_atfork_disable()
+{
+    ucs_list_head_init(&ucs_rcache_global_context.list);
```
Artemy-Mellanox, line 1158: need no lock?

yosefe: it could cause deadlock..

Artemy-Mellanox (approved):

PR #9133, opened 2023-06-09
UCP/DATATYPE: Refactor and optimize datatype iterator

yosefe:
## Why
- Preparation for using datatype_iterator for receive requests (instead of dt_state)
- Optimize user memh and memory detection
- Consolidate datatype code under datatype_iterator


src/ucp/dt/dt_iov.h
```diff
@@ -110,6 +110,22 @@ void ucp_dt_iov_seek(ucp_dt_iov_t *iov, size_t iovcnt, ptrdiff_t distance,
...
 
+/**
+ * Check that all iov entries match the given memory info
+ *
+ * @param [in]     context        Context for memory detection
+ * @param [in]     iov            @ref ucp_dt_iov_t buffer to check
+ * @param [in]     iovcnt         Number of entries in the @a iov buffer
+ * @param [in]     mem_info       Compare the iov entries to this memory info
+ *
+ * @return UCS_OK if all iov entries match  the given memory info, otherwise
```
brminich, line 121: extra space after `match`

src/ucp/core/ucp_request.inl
```diff
@@ -667,6 +667,34 @@ ucp_request_wait_uct_comp(ucp_request_t *req)
     }
 }
 
+static UCS_F_ALWAYS_INLINE ucs_status_t ucp_request_recv_data_unpack_only(
+        ucp_worker_h worker, void *buffer, size_t count, const void *data,
```
brminich, line 671: `request` is not passed to this func, worth to rename somehow and move to `dt`* file

brminich (dismissed):

src/ucp/dt/dt_contig.h
rakhmets, line 1: After the changes: in addition to the inline functions, there is only one method and structure definition (which is used by the method). This method is only used in src/ucp/rma/rma_basic.c.
I think it is better to do the following in the separate PR: move inlines to .inl, and move ucp_memcpy_pack_cb to the file where it is used.

src/ucp/dt/dt_iov.c
```diff
@@ -127,21 +122,16 @@ size_t ucp_dt_iov_count_nonempty(const ucp_dt_iov_t *iov, size_t iovcnt)
     return count;
 }
 
-static UCS_F_ALWAYS_INLINE ucs_status_t
-ucp_dt_iov_memtype_check(ucp_context_h context, const ucp_dt_iov_t *iov,
-                         size_t iovcnt, const ucp_request_param_t *param,
-                         ucp_memory_info_t *mem_info)
+ucs_status_t ucp_dt_iov_memtype_check(ucp_context_h context,
+                                      const ucp_dt_iov_t *iov, size_t iovcnt,
+                                      const ucp_memory_info_t *mem_info)
```
rakhmets, line 127: Guess `UCS_V_UNUSED` can be removed at lines 129-130.

src/ucp/dt/dt_iov.c
```diff
@@ -127,21 +122,16 @@ size_t ucp_dt_iov_count_nonempty(const ucp_dt_iov_t *iov, size_t iovcnt)
...
 
-    if (!ENABLE_PARAMS_CHECK) {
-        goto out;
-    }
-
     for (i = 0; i < iovcnt; ++i) {
-        ucp_memory_detect_param(context, iov[i].buffer, iov[i].length, param,
-                                &mem_info_iter);
+        ucp_memory_detect(context, iov[i].buffer, iov[i].length,
+                          &mem_info_iter);
```
rakhmets, line 134: Maybe it is better to update the error message at line 137 too. `iov[0]` to smth like `expected`.

src/ucp/dt/datatype_iter.h
```diff
@@ -68,17 +68,22 @@ typedef struct {
...
 
 ucs_status_t ucp_datatype_iter_iov_mem_reg(ucp_context_h context,
                                            ucp_datatype_iter_t *dt_iter,
                                            ucp_md_map_t md_map,
                                            unsigned uct_flags);
 
-
 void ucp_datatype_iter_iov_mem_dereg(ucp_datatype_iter_t *dt_iter);
 
+void ucp_datatype_iter_iov_seek_always(ucp_datatype_iter_t *dt_iter,
```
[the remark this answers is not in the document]
yosefe: right, but the function itself is not inline on purpose - it handles slow path case when the offset is different than expected (out of order) so we must declare it in the header

src/ucp/dt/datatype_iter.h
```diff
@@ -68,17 +68,22 @@ typedef struct {
...
+                                       size_t offset);
+
+void ucp_datatype_iter_iov_cleanup_check(ucp_datatype_iter_t *dt_iter);
```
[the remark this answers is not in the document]
yosefe: same answer - it's slowpath non inline function

src/ucp/dt/dt_iov.c
yosefe: imo better keep the message that indicates why we expect the memory type to be that value (because of iov[0])

src/ucp/dt/datatype_iter.inl
```diff
@@ -138,32 +135,67 @@ ucp_datatype_iter_init(ucp_context_h context, void *buffer, size_t count,
...
-        if (status != UCS_OK) {
-            return status;
-        }
-    } else {
-        ucs_assert(dt_iter->dt_class == UCP_DATATYPE_GENERIC);
+        ucp_datatype_iter_iov_set_sg_count(sg_count, count);
+        length = ucp_dt_iov_length((const ucp_dt_iov_t*)buffer, count);
+        return ucp_datatype_iov_iter_init(context, buffer, count, length,
+                                          dt_iter, param);
+    } else if (!ENABLE_PARAMS_CHECK ||
```
rakhmets, line 152: Why do we check `ENABLE_PARAMS_CHECK` here?
E.g. `ENABLE_PARAMS_CHECK = 0`, and `dt_iter->dt_class = 1 // strided`, `ucp_datatype_generic_iter_init` will be called.

src/ucp/dt/datatype_iter.inl
```diff
@@ -138,32 +135,67 @@ ucp_datatype_iter_init(ucp_context_h context, void *buffer, size_t count,
...
+
+    if (ucs_likely(dt_iter->dt_class == UCP_DATATYPE_CONTIG)) {
+        length = ucp_contig_dt_length(datatype, count);
+        return ucp_datatype_contig_iter_init(context, buffer, length, dt_iter,
+                                             param);
+    } else if (dt_iter->dt_class == UCP_DATATYPE_IOV) {
+        length = ucp_dt_iov_length((const ucp_dt_iov_t*)buffer, count);
+        return ucp_datatype_iov_iter_init(context, buffer, count, length,
+                                          dt_iter, param);
+    } else if (!ENABLE_PARAMS_CHECK ||
```
rakhmets, line 190: The same comment as for `ucp_datatype_iter_init`.

src/ucp/dt/datatype_iter.inl
yosefe: if ENABLE_PARAMS_CHECK==0 - we don't check for invalid dt_class and assume the 'else' case is generic
if ENABLE_PARAMS_CHECK==1 - we fail with UCS_ERR_INVALID_PARAM if dt_class is invalid

yosefe:

src/ucp/dt/datatype_iter.inl
```diff
@@ -235,18 +262,88 @@ ucp_datatype_iter_slice(const ucp_datatype_iter_t *dt_iter, size_t offset,
...
+    }
+
+    ucs_free(dt_iter->type.iov.memh);
+}
+
+static UCS_F_ALWAYS_INLINE void
+ucp_datatatype_iter_memh_cleanup_check(ucp_mem_h memh)
+{
+    ucs_assertv((memh == NULL) || ucp_memh_is_zero_length(memh) ||
+                        ucp_memh_is_user_memh(memh),
```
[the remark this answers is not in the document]
yosefe: IMO performance is not important here (it's assert) ; better hide the details of memh and parent management in the ucp_memh_XX functions and not assume/rely on them in datatype_iter code

src/ucp/dt/datatype_iter.inl
```diff
@@ -235,18 +262,88 @@ ucp_datatype_iter_slice(const ucp_datatype_iter_t *dt_iter, size_t offset,
...
-ucp_datatype_iter_cleanup(ucp_datatype_iter_t *dt_iter, unsigned dt_mask)
+ucp_datatype_iter_cleanup(ucp_datatype_iter_t *dt_iter, int dereg,
+                          unsigned dt_mask)
 {
     if (ucp_datatype_iter_is_class(dt_iter, UCP_DATATYPE_CONTIG, dt_mask)) {
-        dt_iter->type.contig.memh = NULL;
+        if (dereg) {
+            ucp_datatype_iter_mem_dereg_single(&dt_iter->type.contig.memh);
+        }
+        ucp_datatatype_iter_memh_cleanup_check(dt_iter->type.contig.memh);
```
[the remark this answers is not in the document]
yosefe: ucp_datatatype_iter_memh_cleanup_check is already inline so compiler could ignore the assert in release build

src/ucp/dt/datatype_iter.inl
```diff
@@ -235,18 +262,88 @@ ucp_datatype_iter_slice(const ucp_datatype_iter_t *dt_iter, size_t offset,
...
+}
+
+static UCS_F_ALWAYS_INLINE void
+ucp_datatype_iter_mem_dereg_single(ucp_mem_h *memh_p)
+{
+    if (*memh_p == NULL) {
+        return;
+    }
+
+    if (ucp_memh_put(*memh_p)) {
```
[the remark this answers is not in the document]
yosefe: Seems to me that ucp_memh_put() will not do anything with user-memh, regardless or rcache enabled/disabled. Even when rcache is disabled, memh->parent will be memh (non-NULL) and this function will return immediately.
https://github.com/openucx/ucx/blob/d5fb870584727917155154b266858d3f29e0add4/src/ucp/core/ucp_mm.inl#L85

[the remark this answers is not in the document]
Artemy-Mellanox: Right. It also will return 1 for zero memh, but it doesn't seem a problem. Maybe add some comment about meaning of `ucp_memh_put()` return valuse?

src/ucp/proto/proto_common.inl
```diff
@@ -86,8 +86,8 @@ ucp_proto_request_zcopy_clean(ucp_request_t *req, unsigned dt_mask)
 static UCS_F_ALWAYS_INLINE void
 ucp_proto_request_zcopy_complete(ucp_request_t *req, ucs_status_t status)
 {
-    ucp_proto_request_zcopy_clean(req, UCP_DT_MASK_CONTIG_IOV);
-    ucp_datatype_iter_cleanup(&req->send.state.dt_iter, UCP_DT_MASK_CONTIG_IOV);
+    ucp_datatype_iter_cleanup(&req->send.state.dt_iter, 1,
+                              UCP_DT_MASK_CONTIG_IOV);
```
rakhmets, line 90: `ucp_proto_request_zcopy_clean` includes `req->flags &= ~UCP_REQUEST_FLAG_PROTO_INITIALIZED;` and memory deregestration. Now memory deregestritaion is a part of cleanup. Isn't this an issue, that we no longer remove the flag in this method?

yosefe: it should not be an issue because ucp_proto_request_zcopy_complete() would release the request anyway

rakhmets (approved):

yosefe:
tests passed but the status was not updated

PR #9135, opened 2023-06-11
AZP: Check GPU presence for GPU tests

Alexey-Rivkin:
## What
Make sure the GPU is detected for all GPU-based tests.

## Why 
Reduce troubleshooting efforts.

### DIFF
--- a/buildlib/az-helpers.sh
+++ b/buildlib/az-helpers.sh
@@ -140,6 +140,16 @@ function az_module_unload() {
     module unload "${module}" || true
 }
 
+# Ensure that GPU is present
+check_gpu() {
+    name=$1
+    if [ "$name" == "gpu" ]; then
+        if ! nvidia-smi -L |& grep -q GPU; then
+            azure_log_error "No GPU device found on $(hostname -s)"
+            exit 1
+        fi
+    fi
+}
 
 #
 # try load cuda modules if nvidia driver is installed
--- a/buildlib/jucx/jucx-test.yml
+++ b/buildlib/jucx/jucx-test.yml
@@ -26,6 +26,7 @@ jobs:
       - bash: |
           set -xeE
           source buildlib/az-helpers.sh
+          check_gpu ${{ parameters.name }}
           az_init_modules
           az_module_load dev/mvn
           az_module_load dev/jdk-${JAVA_VERSION}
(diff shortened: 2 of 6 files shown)


buildlib/az-helpers.sh
```diff
@@ -140,6 +140,13 @@ function az_module_unload() {
     module unload "${module}" || true
 }
 
+# Ensure that GPU is present
+check_gpu() {
+    if [ "$(nvidia-smi -L | wc -l)" -eq 0 ]; then
```
yosefe, line 145: `if ! nvidia-smi -L |& grep -q GPU; then`

artemry-nv, line 145: I'd also add `command -v nvidia-smi` to make sure CUDA toolkit is installed

artemry-nv (approved):

buildlib/jucx/jucx-test.yml
```diff
@@ -26,6 +26,9 @@ jobs:
       - bash: |
           set -xeE
           source buildlib/az-helpers.sh
+          if [ "${{ parameters.name }}" == "gpu" ]; then
```
yosefe, line 29: maybe it should be part of try_load_cuda_env? 
or at least do it after loading cuda env

Alexey-Rivkin: 'try_load_cuda_env' is rather forgiving, as it's used by non-GPU tests too. 
A separate function is needed to only run for GPU tests.

Any reason to load the cuda env if there is no GPU?
Per the "fail fast" concept, a pipeline should halt whenever a crucial condition is detected. This should save resources and provide fast feedback.

buildlib/az-helpers.sh
Alexey-Rivkin: A function 'try_load_cuda_env' runs nvidia-smi already, so this is covered.

buildlib/pr/tests.yml
```diff
@@ -30,6 +30,9 @@ jobs:
       - bash: |
           source ./buildlib/az-helpers.sh
           az_init_modules
+          if [ "${{ parameters.name }}" == "gpu" ]; then
```
brminich, line 33: minor:
can have another wrapper which would accept the name, so that this `if` will be hidden there (you use this `if` twice already)

Alexey-Rivkin: Done, thanks!

yosefe (approved):

Alexey-Rivkin:
@yosefe Can we please merge this?

PR #9145, opened 2023-06-14
INFRA: Fix PPID cmd

Alexey-Rivkin:
## What
Fix cmd to find the parent process ID (PPID) for zombie processes.
Also improved zombie search. 

## Why
Wrong PPID was the reason that the script couldn't kill some zombies.

### DIFF
--- a/buildlib/tools/zombie_killer.sh
+++ b/buildlib/tools/zombie_killer.sh
@@ -1,9 +1,9 @@
-#!/bin/bash -eEx
+#!/bin/bash -eE
 
-# Search for zombie or defunct processes; kill them all
-out=$(ps -eo pid,stat | awk '$2 == "Z|defunct" {print $1}')
+# Search for zombies; kill them all
+out=$(ps -eo pid,stat,%cpu | grep Z | awk '{print $1}')
 if [ -z "$out" ]; then
-  echo "No zombies found"
+  echo "No zombies were found"
   exit 0
 else
   for i in $out ; do
@@ -11,10 +11,10 @@ else
   done
 fi
 
-# Search again; tell parents you killed their kids; kill parents
-out=$(ps -eo pid,stat | awk '$2 == "Z|defunct" {print $1}')
+# Search for survivors; tell parents you killed their kids; kill parents
+out=$(ps -eo pid,stat,%cpu | grep Z | awk '$3>85.0 {print $1}')
 for i in $out ; do
-  pid_parent=$(ps -o ppid=$i)
+  pid_parent=$(ps -p "$i" -o ppid=)
   kill -s SIGCHLD $pid_parent
   kill -9 $pid_parent
 done


Alexey-Rivkin:
Per Yossi's advice, only kill zombies with high (above 85%) CPU consumption.

buildlib/tools/zombie_killer.sh
```diff
@@ -1,20 +1,20 @@
-#!/bin/bash -eEx
+#!/bin/bash -eE
 
-# Search for zombie or defunct processes; kill them all
-out=$(ps -eo pid,stat | awk '$2 == "Z|defunct" {print $1}')
+# Search for zombie with high CPU usage; kill them all
+out=$(ps -eo pid,stat,%cpu | grep Z | awk '$3>85.0 {print $1}')
```
artemry-nv, line 4: Why don't we terminate all the zombie processes?

buildlib/tools/zombie_killer.sh
```diff
@@ -1,20 +1,20 @@
...
 
-# Search again; tell parents you killed their kids; kill parents
-out=$(ps -eo pid,stat | awk '$2 == "Z|defunct" {print $1}')
+# Search for survivors; tell parents you killed their kids; kill parents
+out=$(ps -eo pid,stat,%cpu | grep Z | awk '$3>85.0 {print $1}')
 for i in $out ; do
-  pid_parent=$(ps -o ppid=$i)
+  pid_parent=$(ps -p "$i" -o ppid=)
   kill -s SIGCHLD $pid_parent
   kill -9 $pid_parent
```
artemry-nv, line 19: Why do we need to terminate parent process with SIGKILL? As you mentioned parent process may be a docker process

buildlib/tools/zombie_killer.sh
Alexey-Rivkin: Most zombies are harmless, but we may kill them just out of spite.

buildlib/tools/zombie_killer.sh
Alexey-Rivkin: If a zombie couldn't be dealt with directly, approaching its parents is the only choice left.
For the high CPU-eating zombies it well worth the hustle.

yosefe (approved):

buildlib/tools/zombie_killer.sh
artemry-nv: But if we parent for example `init` or another important process?

buildlib/tools/zombie_killer.sh
artemry-nv: They may occupy resources and slow down the system. If we don't terminate all the zombie processes maybe at least print all the remaining ones in the end (for future debug purposes)?

Alexey-Rivkin: We do kill them all now, changed per your recommendation.

buildlib/tools/zombie_killer.sh
artemry-nv: It appears we cannot terminate init even with SIGKILL.

PR #9148, opened 2023-06-19
UCM/BISTRO: Add locking step while hooking library functions

tvegas1:
## What
Adds a locking step and grace period while patching the hooked functions to narrow existing race condition.

## Why ?
Race conditions could be seen when ucm bistro code is being invoked through a `dlopen()` call.

## How ?
After inserting a lock, we wait for existing parallel code flows to complete. The lock itself is eventually removed so that we don't grow the patch size. For instance, including the first instruction after the syscall instruction under the patched area could lead to corruption, as a system call could take longer to complete than the grace period.

Added a test program to exercise the busy loop locking/unlocking sequence.

### Tested
Built in release and tested:
- `while :; do ./test/apps/test_hooks || break; done`
  - test on x86_64 (short and long patch) and aarch64
- repro with and without fix on aarch64

### DIFF
--- a/contrib/test_jenkins.sh
+++ b/contrib/test_jenkins.sh
@@ -974,6 +974,25 @@ test_ucp_dlopen() {
 	fi
 }
 
+test_ucm_hooks() {
+    total=30
+    echo "==== Running UCM Bistro hook test ===="
+    for i in $(seq 1 $total); do
+        threads=$(((RANDOM % (2 * `nproc`)) + 1))
+
+        echo "iteration $i/$total: $threads threads"
+        timeout 10 ./test/apps/test_hooks -n $threads >test_hooks.log 2>&1 || \
+            { \
+                echo "ERROR running bistro hook test:"; \
+                cat test_hooks.log; \
+                exit 1; \
+            }
+    done
+
+    echo "SUCCESS running bistro hook test:"
+    cat test_hooks.log
+}
+
 test_init_mt() {
 	echo "==== Running multi-thread init ===="
 	# Each thread requires 5MB. Cap threads number by total available shared memory.
@@ -1267,6 +1286,7 @@ run_release_mode_tests() {
 	test_ucs_load
 	test_ucp_dlopen
 	run_gtest_release
+	test_ucm_hooks
 }
 
 #
--- a/src/ucm/bistro/bistro.c
+++ b/src/ucm/bistro/bistro.c
@@ -11,11 +11,14 @@
 #include <dlfcn.h>
 #include <stdlib.h>
 #include <pthread.h>
+#include <unistd.h>
 
 #include <ucm/bistro/bistro.h>
 #include <ucm/bistro/bistro_int.h>
 #include <ucs/type/serialize.h>
 #include <ucs/sys/math.h>
+#include <ucs/time/time.h>
+#include <ucs/arch/atomic.h>
 
 
 ucs_status_t ucm_bistro_remove_restore_point(ucm_bistro_restore_point_t *rp)
@@ -43,6 +46,52 @@ static ucs_status_t ucm_bistro_protect(void *addr, size_t len, int prot)
     }
 
     return UCS_OK;
+}
+
+void ucm_bistro_modify_code(void *dst, const ucm_bistro_lock_t *bytes)
+{
+    uint32_t value;
+
+    UCS_STATIC_ASSERT(sizeof(*bytes) <= sizeof(value));
+    memcpy(&value, dst, sizeof(value));
+    memcpy(&value, bytes, sizeof(*bytes));
+    (void)ucs_atomic_swap32(dst, value);
+}
+
+ucs_status_t
+ucm_bistro_apply_patch_atomic(void *dst, const void *patch, size_t len)
+{
+    size_t skip           = sizeof(ucm_bistro_lock_t);
+    double grace_duration = 5e-3;
+    double deadline;
+    ucs_status_t status;
+
+    status = ucm_bistro_protect(dst, len, UCM_PROT_READ_WRITE_EXEC);
+    if (UCS_STATUS_IS_ERR(status)) {
+        return status;
+    }
+
+    /* Lock the codepatch and wait for existing flows to complete */
+    ucm_bistro_patch_lock(dst);
+    ucs_clear_cache(dst, UCS_PTR_BYTE_OFFSET(dst, len));
+
+    deadline = ucm_get_time() + grace_duration;
+    while (ucm_get_time() < deadline) {
+        sched_yield();
+    }
+
+    /* Copy the payload behind the lock */
+    memcpy(UCS_PTR_BYTE_OFFSET(dst, skip), UCS_PTR_BYTE_OFFSET(patch, skip),
+           len - skip);
+    ucs_clear_cache(dst, UCS_PTR_BYTE_OFFSET(dst, len));
+
+    /* Unlock the codepath */
+    ucm_bistro_modify_code(dst, patch);
+
+    status = ucm_bistro_protect(dst, len, UCM_PROT_READ_EXEC);
+    ucs_clear_cache(dst, UCS_PTR_BYTE_OFFSET(dst, len));
+
+    return status;
 }
 
 ucs_status_t ucm_bistro_apply_patch(void *dst, void *patch, size_t len)
--- a/src/ucm/bistro/bistro_aarch64.h
+++ b/src/ucm/bistro/bistro_aarch64.h
@@ -35,4 +35,17 @@ ucs_status_t ucm_bistro_patch(void *func_ptr, void *hook, const char *symbol,
                               void **orig_func_p,
                               ucm_bistro_restore_point_t **rp);
 
+/* Instruction type */
+typedef uint32_t ucm_bistro_inst_t;
+
+/* Lock implementation */
+typedef struct {
+    ucm_bistro_inst_t b; /* either self branch, or no-op branch */
+} UCS_S_PACKED ucm_bistro_lock_t;
+
+/**
+ * Helper functions to improve atomicity of function patching
+ */
+void ucm_bistro_patch_lock(void *dst);
+
 #endif
--- a/src/ucm/util/sys.h
+++ b/src/ucm/util/sys.h
@@ -10,6 +10,7 @@
 
 #include <ucm/api/ucm.h>
 #include <ucs/sys/checker.h>
+#include <ucs/time/time.h>
 #include <sys/types.h>
 #include <stddef.h>
 
@@ -129,6 +130,20 @@ ucm_get_hook_mode(ucm_mmap_hook_mode_t config_mode)
 
     return config_mode;
 #endif
+}
+
+
+/**
+ * Get current time without depending on UCS library code
+ *
+ * @return The current time value in seconds
+ */
+static UCS_F_ALWAYS_INLINE double ucm_get_time()
+{
+    struct timeval tv;
+
+    gettimeofday(&tv, NULL);
+    return (double)tv.tv_sec + ((double)tv.tv_usec / UCS_USEC_PER_SEC);
 }
 
 #endif
--- a/src/ucs/time/time.h
+++ b/src/ucs/time/time.h
@@ -72,7 +72,6 @@ static inline double ucs_time_sec_value()
     return ucs_get_cpu_clocks_per_sec();
 }
 
-
 /**
  * Convert seconds to UCS time units.
  */
--- a/test/apps/Makefile.am
+++ b/test/apps/Makefile.am
@@ -21,9 +21,16 @@ noinst_PROGRAMS = \
 	test_link_map \
 	test_dlopen_cfg_print \
 	test_init_mt \
-	test_memtrack_limit
+	test_memtrack_limit \
+	test_hooks
 
 objdir = $(shell sed -n -e 's/^objdir=\(.*\)$$/\1/p' $(LIBTOOL))
+
+test_hooks_SOURCES  = test_hooks.c
+test_hooks_CPPFLAGS = $(BASE_CPPFLAGS) \
+                      -DLIB_PATH=$(abs_top_builddir)/src/ucs/$(objdir)/libucs.so
+test_hooks_CFLAGS   = $(BASE_CFLAGS)
+test_hooks_LDADD    = -ldl
 
 test_ucs_dlopen_SOURCES  = test_ucs_dlopen.c
 test_ucs_dlopen_CPPFLAGS = $(BASE_CPPFLAGS) \
--- a/test/apps/test_hooks.c
+++ b/test/apps/test_hooks.c
@@ -1,0 +1,205 @@
+/**
+ * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2023. ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <sys/mman.h>
+#include <sys/time.h>
+#include <dlfcn.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <pthread.h>
+
+#include <ucs/sys/preprocessor.h>
+#include <ucm/api/ucm.h>
+#include <ucm/util/sys.h>
+
+
+#define DEFAULT_THREAD_COUNT   16
+#define DEFAULT_THREAD_TIMEOUT 1
+
+
+typedef struct {
+    const char   *libname;
+    int          thread_count;
+    pthread_t    *threads;
+    volatile int stop;
+    double       thread_timeout;
+    void         *dl_handle;
+} context_t;
+
+static ucs_status_t ctx_init(context_t *ctx, const char *filename,
+                             int thread_count, int thread_timeout)
+{
+    pthread_t *threads;
+
+    threads = calloc(thread_count, sizeof(*threads));
+    if (threads == NULL) {
+        return UCS_ERR_NO_MEMORY;
+    }
+
+    ctx->dl_handle      = NULL;
+    ctx->libname        = filename;
+    ctx->thread_count   = thread_count;
+    ctx->threads        = threads;
+    ctx->stop           = 0;
+    ctx->thread_timeout = thread_timeout;
+
+    return UCS_OK;
+}
+
+static ucs_status_t ctx_load_library(context_t *ctx)
+{
+    printf("loading library to trigger hooks installation\n");
+
+    dlerror();
+    ctx->dl_handle = dlopen(ctx->libname, RTLD_NOW);
+    if (ctx->dl_handle == NULL) {
+        printf("dlopen(filename=\"%s\", RTLD_NOW) failed: %s\n", ctx->libname,
+               dlerror());
+        return UCS_ERR_INVALID_PARAM;
+    }
+
+    return UCS_OK;
+}
+
+static void *thread_handler(void *arg)
+{
+    context_t *ctx  = arg;
+    double deadline = ucm_get_time() + ctx->thread_timeout;
+
+    /*
+     * Many threads racing for read lock are preventing ucm to finish testing
+     * the installed hooks, so we make sure to stop the threads after
+     * reasonable time.
+     */
+    do {
+        if (mmap(NULL, 0, PROT_READ | PROT_WRITE, MAP_PRIVATE, -1, 0) !=
+            MAP_FAILED) {
+            printf("failed to return mmap_call() failure\n");
+            exit(EXIT_FAILURE);
+        }
+    } while (!ctx->stop &&
+             ((ctx->thread_timeout == 0) || (ucm_get_time() < deadline)));
+
+    return NULL;
+}
+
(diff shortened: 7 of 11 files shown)


brminich (dismissed):
👍

src/ucs/time/time.h
```diff
@@ -72,6 +74,17 @@ static inline double ucs_time_sec_value()
     return ucs_get_cpu_clocks_per_sec();
 }
 
+/**
+ * @return The current time, without dependency on UCM code
```
tvegas1, line 78: we want `dlopen(libucm.so)` to succeed without needing `libucs.so` (symbol: `ucs_arch_x86_enable_rdtsc`)

src/ucm/bistro/bistro.c
```diff
@@ -45,6 +48,59 @@ static ucs_status_t ucm_bistro_protect(void *addr, size_t len, int prot)
     return UCS_OK;
 }
 
+void ucm_bistro_apply_locking(void *dst, const ucm_bistro_lock_t *bytes)
+{
+    uint32_t value;
+
+    if (sizeof(value) > sizeof(*bytes)) {
```
yosefe, line 55: maybe remove this condition to simplify the code?

src/ucm/bistro/bistro.c
```diff
@@ -45,6 +48,59 @@ static ucs_status_t ucm_bistro_protect(void *addr, size_t len, int prot)
...
+        memcpy(&value, dst, sizeof(value));
+    }
+
+    memcpy(&value, bytes, sizeof(*bytes));
+    (void)ucs_atomic_swap32(dst, value);
+}
+
+static void ucm_bistro_patch_unlock(void *dst, const void *patch)
```
yosefe, line 63: maybe remove this wrapper?

src/ucm/bistro/bistro.c
```diff
@@ -45,6 +48,59 @@ static ucs_status_t ucm_bistro_protect(void *addr, size_t len, int prot)
...
+void ucm_bistro_apply_locking(void *dst, const ucm_bistro_lock_t *bytes)
+{
+    uint32_t value;
+
+    if (sizeof(value) > sizeof(*bytes)) {
+        memcpy(&value, dst, sizeof(value));
+    }
+
+    memcpy(&value, bytes, sizeof(*bytes));
+    (void)ucs_atomic_swap32(dst, value);
```
yosefe, line 60: does it mean sizeof(*bytes) must be <= sizeof(value) ? can add static assert for it..

src/ucm/bistro/bistro.c
```diff
@@ -45,6 +48,59 @@ static ucs_status_t ucm_bistro_protect(void *addr, size_t len, int prot)
     return UCS_OK;
 }
 
+void ucm_bistro_apply_locking(void *dst, const ucm_bistro_lock_t *bytes)
```
yosefe, line 51: maybe rename to ucm_bistro_modofy_code? and move clear_cache inside this function?

src/ucm/bistro/bistro_aarch64.h
```diff
@@ -35,4 +35,17 @@ ucs_status_t ucm_bistro_patch(void *func_ptr, void *hook, const char *symbol,
                               void **orig_func_p,
                               ucm_bistro_restore_point_t **rp);
 
+/* Instruction type */
+typedef uint32_t ucm_bistro_inst_t;
+
+/* Lock implementation */
+typedef struct {
+    ucm_bistro_inst_t b; /* either self branch, or no-op branch */
+} UCS_S_PACKED ucm_bistro_lock_t;
```
yosefe, line 44: do we really need a typedef for ucm_bistro_lock_t?

src/ucs/time/time.h
```diff
@@ -36,6 +36,8 @@ typedef uint32_t             ucs_short_time_t;
 #define UCS_MSEC_PER_SEC   1000ull
 /* Micro per sec */
 #define UCS_USEC_PER_SEC   1000000ul
+/* Micro per milli */
+#define UCS_USEC_PER_MSEC 1000ull
```
yosefe, line 40: IMO no need to define it, can use the other constants

src/ucs/time/time.h
yosefe: maybe just call gettimeofday from UCM?

test/apps/Makefile.am
```diff
@@ -21,10 +21,17 @@ noinst_PROGRAMS = \
...
+	test_memtrack_limit \
+	test_hooks
 
 objdir = $(shell sed -n -e 's/^objdir=\(.*\)$$/\1/p' $(LIBTOOL))
 
+test_hooks_SOURCES       = test_hooks.c
+test_hooks_CPPFLAGS      = $(BASE_CPPFLAGS) \
+                           -DLIB_PATH=$(abs_top_builddir)/src/ucs/$(objdir)/libucs.so
+test_hooks_CFLAGS        = $(BASE_CFLAGS)
+test_hooks_LDADD         = -ldl
```
yosefe, line 33: remove extra spaces before = : align test_hooks_XX defines on same column, but not necessarily align with other targets

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <pthread.h>
+
+#include <ucm/api/ucm.h>
+
+
+#define _QUOTE(x) #x
+#define QUOTE(x)  _QUOTE(x)
```
yosefe, line 23: can we use ucs_pp_quote?

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+
+#define DEFAULT_THREAD_COUNT   16
+#define DEFAULT_TIMEOUT        10
+#define DEFAULT_THREAD_TIMEOUT 1
+
+#define log(fmt, ...) printf(fmt "\n", ##__VA_ARGS__)
```
yosefe, line 29: IMO just use printf or ucs_log

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+
+
+static void *mmap_call(size_t size)
```
yosefe, line 32: IMO no need this wrapper - used only once

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+{
+    return mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE, -1, 0);
+}
+
+struct context {
```
yosefe, line 37: typedef struct { ... } context_t;

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+    const char   *libname;
+    int          thread_count;
+    pthread_t    *thread;
```
yosefe, line 40: rename to threads ? (in plural)

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+    if (ctx->dl_handle == NULL) {
+        log("dlopen(filename=\"%s\", RTLD_NOW) failed: %s", ctx->libname,
+            dlerror());
+        return UCS_ERR_INVALID_PARAM;
+    }
+
+    return UCS_OK;
+}
+
+static uint64_t get_time_msec(void)
```
yosefe, line 81: maybe use ucs_get_time()?

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+     * the installed hooks, so we make sure to stop the threads after
+     * reasonable time.
+     */
+    start   = get_time_msec();
+    current = start;
+
+    while (!ctx->stop && ((ctx->thread_timeout_msec == 0) ||
+                          (current < (start + ctx->thread_timeout_msec)))) {
+        ptr = mmap_call(0);
+        if (ptr != (void*)-1) {
```
yosefe, line 106: use MAP_FAILED constant

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+    void *ptr;
+    uint64_t start, current;
+
+    /*
+     * Many threads racing for read lock are preventing ucm to finish testing
+     * the installed hooks, so we make sure to stop the threads after
+     * reasonable time.
+     */
+    start   = get_time_msec();
+    current = start;
```
yosefe, line 101: can use do..while, then no need to declare current variable

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+
+    return NULL;
+}
+
+static ucs_status_t ctx_start_threads(struct context *ctx)
+{
+    int i, ret;
+
+    for (i = 0; i < ctx->thread_count; i++) {
+        ret = pthread_create(ctx->thread + i, NULL, thread_handler, ctx);
```
yosefe, line 122: &ctx->thread[I]

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+        if (ret < 0) {
+            log("pthread_create(thread #%d) failed: %m", i);
+            goto err;
+        }
+    }
+
+    log("started %d threads", ctx->thread_count);
+
```
yosefe, line 130: remove spaceline

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+    return UCS_OK;
+
+err:
+    for (; i > 0; i--) {
+        pthread_join(ctx->thread[i - 1], NULL);
+    }
+
```
yosefe, line 137: remove spaceline

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+{
+    int i;
+
+    ctx->stop = 1;
+
+    log("stopping %d threads", ctx->thread_count);
+
+    for (i = 0; i < ctx->thread_count; i++) {
+        pthread_join(ctx->thread[i], NULL);
+    }
```
yosefe, line 151: maybe reuse code with error flow

test/apps/test_hooks.c
```diff
@@ -0,0 +1,240 @@
...
+}
+
+static void ctx_free(struct context *ctx)
+{
+    if (ctx->dl_handle != NULL) {
+        dlclose(ctx->dl_handle);
+    }
+
```
yosefe, line 159: remove spaceline

test/apps/Makefile.am
```diff
@@ -21,10 +21,17 @@ noinst_PROGRAMS = \
 	test_link_map \
 	test_dlopen_cfg_print \
 	test_init_mt \
-	test_memtrack_limit
+	test_memtrack_limit \
+	test_hooks
 
 objdir = $(shell sed -n -e 's/^objdir=\(.*\)$$/\1/p' $(LIBTOOL))
 
+test_hooks_SOURCES       = test_hooks.c
```
yosefe, line 29: is this test added to CI?

test/apps/test_hooks.c
tvegas1: fixed, using `UCS_PP_MAKE_STRING`

test/apps/test_hooks.c
tvegas1: fixed, using printf

test/apps/test_hooks.c
tvegas1: Putting gdb breakpoints on general mmap() function is painful as calls are everywhere. Using that function helps to simply narrow the break so i'd like to keep it?

test/apps/test_hooks.c
tvegas1: fixed

test/apps/test_hooks.c
tvegas1: fixed

test/apps/test_hooks.c
tvegas1: cannot use it as it brings dependency on libucs code that we are dlopen'ing.

test/apps/test_hooks.c
tvegas1: fixed

test/apps/test_hooks.c
tvegas1: fixed

test/apps/test_hooks.c
tvegas1: fixed

test/apps/test_hooks.c
tvegas1: fixed

test/apps/test_hooks.c
tvegas1: fixed

test/apps/test_hooks.c
tvegas1: fixed

test/apps/Makefile.am
tvegas1: fixed

test/apps/Makefile.am
tvegas1: fixed, added to `test_jenkins.sh`

src/ucs/time/time.h
tvegas1: fixed

src/ucs/time/time.h
tvegas1: fixed

src/ucm/bistro/bistro.c
tvegas1: fixed

src/ucm/bistro/bistro_aarch64.h
tvegas1: I think so because its size depends on platform. (or are you suggesting to replace `ucm_bistro_lock_t` by `struct ucm_bistro_lock`?)

src/ucm/bistro/bistro.c
tvegas1: Fixed for the name, but for clarity I'd rather keep the three clear cache calls in the single function below. (Also the last change sequence is modify_code+protect_change+clear_cache, so adding clear cache to modify would make one more clear_cache call).

contrib/test_jenkins.sh
```diff
@@ -974,6 +974,22 @@ test_ucp_dlopen() {
 	fi
 }
 
+test_ucm_hooks() {
+    echo "==== Running UCM Bistro hook test ===="
+    for i in $(seq 1 30); do
+        echo -n "."
```
yosefe, line 980: maybe print every iteration index in new line? not sure it will work well with log output redirection in CI...

src/ucm/bistro/bistro.c
```diff
@@ -48,23 +48,28 @@ static ucs_status_t ucm_bistro_protect(void *addr, size_t len, int prot)
     return UCS_OK;
 }
 
-void ucm_bistro_apply_locking(void *dst, const ucm_bistro_lock_t *bytes)
+static double ucm_bistro_get_time_msec()
```
yosefe, line 51: maybe move it to ucm/util/sys.{c,h} as "ucm_get_time"

src/ucm/bistro/bistro.c
yosefe: I mean, always copy previous bytes to value without size check

src/ucm/bistro/bistro.c
yosefe: ik

src/ucm/bistro/bistro_aarch64.h
yosefe: ok, I see

test/apps/test_hooks.c
```diff
@@ -88,88 +84,78 @@ static uint64_t get_time_msec(void)
...
-
+    printf("started %d threads\n", ctx->thread_count);
     return UCS_OK;
 
 err:
-    for (; i > 0; i--) {
-        pthread_join(ctx->thread[i - 1], NULL);
-    }
-
+    ctx->thread_count = i;
```
yosefe, line 134: minor: IMO would be nicer to pass "count" parameter to ctx_stop_threads

src/ucm/bistro/bistro.c
```diff
@@ -45,6 +48,64 @@ static ucs_status_t ucm_bistro_protect(void *addr, size_t len, int prot)
...
+    status = ucm_bistro_protect(dst, len, UCM_PROT_READ_WRITE_EXEC);
+    if (UCS_STATUS_IS_ERR(status)) {
+        return status;
+    }
+
+    /* Lock the codepatch and wait for existing flows to complete */
+    ucm_bistro_patch_lock(dst);
+    ucs_clear_cache(dst, UCS_PTR_BYTE_OFFSET(dst, len));
+
+    deadline = ucm_bistro_get_time_msec() + grace_duration;
```
yosefe, line 90: IMO to void confusion, all time values in bistro and in the test should in "seconds" units and not msec

test/apps/test_hooks.c
```diff
@@ -0,0 +1,227 @@
...
+{
+    return mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE, -1, 0);
+}
+
+typedef struct {
+    const char   *libname;
+    int          thread_count;
+    pthread_t    *threads;
+    volatile int stop;
+    uint64_t     thread_timeout_msec;
```
yosefe, line 38: IMO better take it "double timeout" and in seconds units.

test/apps/test_hooks.c
```diff
@@ -0,0 +1,227 @@
...
+        case 'h':
+        default:
+            usage(argv[0]);
+            return -1;
+        }
+    }
+
+    if (timeout > 0) {
+        alarm(timeout); /* patching could corrupt threads */
+    }
```
yosefe, line 189: IMO better run the test with "timeout" command to kill it if it's stuck, rather than relying on internal timers - that could be corrupted. it also allows easier debug of the test when it hangs and simplifies the test by removing arguments parsing code.

contrib/test_jenkins.sh
tvegas1: fixed

src/ucm/bistro/bistro.c
tvegas1: i see, fixed

src/ucm/bistro/bistro.c
tvegas1: fixed in test_hooks and bistro.c

test/apps/test_hooks.c
tvegas1: fixed

test/apps/test_hooks.c
tvegas1: fixed

src/ucm/bistro/bistro.c
tvegas1: fixed, now using second units

src/ucm/util/sys.h
```diff
@@ -131,4 +132,17 @@ ucm_get_hook_mode(ucm_mmap_hook_mode_t config_mode)
...
+ *
+ * @return The current time value in seconds
+ */
+static UCS_F_ALWAYS_INLINE double ucm_get_time()
+{
+    struct timeval tv;
+
+    gettimeofday(&tv, NULL);
+    return (double)tv.tv_sec + ((double)tv.tv_usec / UCS_USEC_PER_SEC);
+}
```
yosefe, line 147: space line after

contrib/test_jenkins.sh
```diff
@@ -975,18 +975,19 @@ test_ucp_dlopen() {
...
 
 test_ucm_hooks() {
+    total=30
     echo "==== Running UCM Bistro hook test ===="
-    for i in $(seq 1 30); do
-        echo -n "."
-        ./test/apps/test_hooks -n $((`nproc`*4/5+1)) >test_hooks.log 2>&1 || \
+    for i in $(seq 1 $total); do
+        echo "iteration $i/$total"
+        timeout 10 ./test/apps/test_hooks -n $((`nproc`*4/5+1)) >test_hooks.log 2>&1 || \
```
yosefe, line 982: why *4/5+1?

tvegas1: take ~80% percent of the cores but at least one cpu.

src/ucm/util/sys.h
tvegas1: fixed

contrib/test_jenkins.sh
yosefe: why not all the cores?

yosefe: I would even run more threads than cores (or random number of threads every time) to make more stress

tvegas1: fixed

tvegas1: out of curiosity, i checked by starting 522 threads on a 28 cores machine that we can repro issue without fix, added delay, issue still visible, then using the lock and delay, issue cannot be seen. (with 522 that delay is hundreds of msec but it's bit extreme).

yosefe:
@tvegas1 can you pls squash?

cponder:
Can you tell us if this make it into the V1.15.0-rc4 or another upcoming tag?

yosefe:
Currently the fix is only on master branch. The next tag is v1.16.0 which is Nov-Dec timeframe.
@cponder do we need the fix in 1.15.0?

cponder:
I think it will only matter if we can also get #9282 into the same release.

PR #9152, opened 2023-06-21
UCP/LOG: Fix printing protov2 lanes information

yosefe:
## Why
Adjust information printing after enabling protov2 by default

## How
- Print lanes configuration with UCX_LOG_LEVEL=info
- Don't print protov1 thresholds
- Print protov2 info by `ucx_info -u t -e` command

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -3141,6 +3141,11 @@ static void ucp_ep_config_print(FILE *stream, ucp_worker_h worker,
         }
         fprintf(stream, "#                 %s\n", ucs_string_buffer_cstr(&strb));
     }
+
+    if (worker->context->config.ext.proto_enable) {
+        return;
+    }
+
     fprintf(stream, "#\n");
 
     if (context->config.features & UCP_FEATURE_TAG) {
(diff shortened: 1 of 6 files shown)


yosefe:
@rakhmets can you pls take a look?

src/ucp/proto/proto_debug.c
```diff
@@ -219,7 +219,8 @@ ucp_proto_select_elem_info(ucp_worker_h worker,
     ucp_proto_select_param_dump(worker, ep_cfg_index, rkey_cfg_index,
                                 select_param, ucp_operation_descs, &ep_cfg_strb,
                                 &sel_param_strb);
-    if (!ucp_proto_debug_is_info_enabled(
+    if (!show_all &&
```
brminich, line 222: should not it be `||`?

src/ucp/core/ucp_ep.c
```diff
@@ -3143,6 +3143,11 @@ static void ucp_ep_config_print(FILE *stream, ucp_worker_h worker,
     }
     fprintf(stream, "#\n");
 
+
+    if (worker->context->config.ext.proto_enable) {
+        return;
+    }
```
rakhmets, line 3149: Please move this block before the line `fprintf(stream, "#\n");` to remove extra empty line in the output of `ucx_info -u t -e`.
We adding an empty line in `ucp_ep_print_info_internal` after the `ucp_ep_config_print` call.

src/ucp/proto/proto_debug.c
rakhmets: It should be `&&`.
I think the name of the parameter may be misunderstood.
If `show_all` is true, we print information.
Otherwise, `show_all` is false, and if `PROTO_INFO=y`, we print the information.

src/ucp/core/ucp_ep.c
yosefe: fixed and force-pushed due to small pr

rakhmets (approved):

brminich (approved):

PR #9153, opened 2023-06-21
UCT/IB: Remove mem_prefetch operation

yosefe:
## Why
Preparation for MD ops refactoring

## How
ODP prefetch is not mlx5-specific. We can move it to the common code and remote the prefetch callback.

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -618,6 +618,35 @@ static uint64_t uct_ib_md_access_flags(uct_ib_md_t *md, unsigned flags,
     return access_flags;
 }
 
+static ucs_status_t uct_ib_mem_prefetch(uct_ib_md_t *md, uct_ib_mem_t *ib_memh,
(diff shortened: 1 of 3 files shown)


yosefe:

src/uct/ib/base/ib_md.c
```diff
@@ -669,7 +698,7 @@ uct_ib_mem_reg_internal(uct_md_h uct_md, void *address, size_t length,
               params->dmabuf_offset);
 
     if (md->config.odp.prefetch) {
-        md->ops->mem_prefetch(md, memh, address, length);
+        uct_ib_mem_prefetch(md, memh, address, length);
```
gleon99, line 701: No rc checking here - intentional?

yosefe: Yes, we don't want to fail registration just because prefetch failed. Also, it was the previous behavior

Artemy-Mellanox (approved):

PR #9155, opened 2023-06-22
UCP/RNDV: Fix "rndv through am" thresholds definition

ivankochin:
## What
Allows to define RNDV/AM scheme via RNDV_SCHEME control and applies this new option to protocols. 

## Why ?
Now RNDV/AM protos can be chosen even if RNDV_SCHEME is defined to different value.

### DIFF
--- a/src/ucp/dt/datatype_iter.c
+++ b/src/ucp/dt/datatype_iter.c
@@ -55,6 +55,10 @@ ucp_datatype_iter_set_iov_memh(ucp_datatype_iter_t *dt_iter, ucp_mem_h memh)
     size_t iov_index;
     ucs_status_t status;
 
+    if (iov_count == 0) {
+        return UCS_OK;
+    }
+
     status = ucp_datatype_iter_iov_allocate_memh(dt_iter, iov_count);
     if (status != UCS_OK) {
         return status;
@@ -78,7 +82,7 @@ ucs_status_t ucp_datatype_iter_iov_mem_reg(ucp_context_h context,
     ucs_status_t status;
     size_t iov_index;
 
-    if (md_map == 0) {
+    if ((md_map == 0) || (iov_count == 0)) {
         return UCS_OK;
     }
 
--- a/src/ucp/rndv/rndv_am.c
+++ b/src/ucp/rndv/rndv_am.c
@@ -11,6 +11,21 @@
 #include "proto_rndv.inl"
 
 
+static size_t
+ucp_rndv_am_cfg_thresh(ucp_context_t *context, size_t am_thresh)
+{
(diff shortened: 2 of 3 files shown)


yosefe:
Seems this PR reveals existing bug - "memtrack.c:329  UCX  WARN  allocated zero-size block 0x615a9d0 for dt_iov_memh"

src/ucp/core/ucp_context.c
```diff
@@ -223,6 +223,7 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Communication scheme in RNDV protocol.\n"
    " get_zcopy - use get_zcopy scheme in RNDV protocol.\n"
    " put_zcopy - use put_zcopy scheme in RNDV protocol.\n"
+   " am        - use am scheme in RNDV protocol.\n"
```
tvegas1, line 226: Maybe remove that line as #9154 is already taking of this?

ivankochin: Done, tnx!

yosefe (approved):

src/ucp/dt/datatype_iter.c
```diff
@@ -115,7 +119,7 @@ void ucp_datatype_iter_iov_mem_dereg(ucp_datatype_iter_t *dt_iter)
     ucp_mem_h *memh = dt_iter->type.iov.memh;
     size_t iov_index, length;
 
-    if ((memh == NULL) || ucp_memh_is_user_memh(*memh)) {
+    if ((memh == NULL) || (*memh == NULL) || ucp_memh_is_user_memh(*memh)) {
```
brminich, line 122: pls check when we can have `memh != NULL` but `*memh ==NULL`

src/ucp/dt/datatype_iter.c
```diff
@@ -78,7 +82,7 @@ ucs_status_t ucp_datatype_iter_iov_mem_reg(ucp_context_h context,
     ucs_status_t status;
     size_t iov_index;
 
-    if (md_map == 0) {
+    if (md_map == 0 || iov_count == 0) {
```
brminich, line 85:
```suggestion
    if ((md_map == 0) || (iov_count == 0)) {
```

src/ucp/dt/datatype_iter.c
ivankochin: With the new `iov_count` checks it seems like there is no such case. Thanks!

```suggestion
    if ((memh == NULL) || ucp_memh_is_user_memh(*memh)) {
```

brminich (approved):

yosefe (approved):

ivankochin:
Squashed, @brminich @yosefe pls re-approve

brminich (approved):

PR #9158, opened 2023-06-22
GTEST: use std=c++17 for compiling hip gtests

edgargabriel:
## What

hip_runtime.h contains in some rocm releases a C++17 feature. Prior to this commit, we forced using gnu++11 using HIP_CXXFLAGS in UCX, but combined with the `-Werror` flag of UCX this leads now to a compilation error in gtests.

## How

This commits adds the configure logic to check for `-std=c++17`,  and uses this flag instead of `-std=gnu++11` if available. Forcing using the `-std=c++17` flag revealed a second issue regarding ignoring error codes in a gtest routine. 
This commit fixes this issue as well.

### DIFF
--- a/config/m4/rocm.m4
+++ b/config/m4/rocm.m4
@@ -109,13 +109,38 @@ AS_IF([test "x$with_rocm" != "xno"],
     LDFLAGS="$HIP_LDFLAGS $LDFLAGS"
     LIBS="$HIP_LIBS $LIBS"
 
+    #
+    # hip_runtime.h contains a C++17 features
+    # in some ROCm releases. Only releveant for
+    # compiling gtests with UCX.
+    #
+    AC_MSG_CHECKING([c++17 support])
+    AC_LANG_PUSH([C++])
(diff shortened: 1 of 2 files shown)


yosefe:
This will build all gtest with c++17 when ROCm is used. Is this the intention?
A general question - is c++17 a minimal requirement for new ROCm versions?

edgargabriel:
@yosefe fundamentally yes. The small caveat is that this is occurring with an experimental ROCm build, and there is a non-zero chance that the feature will not make it to the mainline. We still need a fix for this issue however for some ongoing work.

An alternative could btw. to investigate why setting CXXFLAGS="-std=c++17" when configuring/compiling UCX did not do the trick, it seemed like the environment variable is ignored.

(as a side note, the change only impacts gtests, not the main ucx build. In addition, I *think* all distro's / releases supported by ROCm have a gcc version recent enough to support c++17, since c++17 is supported since gcc7 as far as I can see. So the actual impact might be minimal).

edgargabriel:
lets hold back with reviews and merging for a bit.

PR #9160, opened 2023-06-23
UCS/LOG: Store an allocated filename in log filter key hash

tvegas1:
## What
Make the global log filter key hash store allocated filenames.

## Why ?
At lookup time, comparing against a previously stored `__FILE__` leads to `SEGFAULT` if the corresponding library has already been `dlclose()`'d.

## How ?
Use variable size chunk allocation.

Gtest are missing, if approach is ok, we can move this to `ucs/datastruct`, and add corresponding tests.

### Crash
configure: `--enable-logging=trace_poll`
run: `mpirun -x UCX_LOG_LEVEL=trace -x UCX_PROTO_ENABLE=y -mca pml ucx -np 2 osu_bw -x 400 -m 4096:4096`

```
[jazz29:39743] *** Process received signal ***
[jazz29:39743] Signal: Segmentation fault (11)
[jazz29:39743] Signal code:  (-6)
[jazz29:39743] Failing at address: 0x182cd00009b3f
[jazz29:39743] [ 0] /lib64/libpthread.so.0(+0xf5d0)[0x7f13f847e5d0]
[jazz29:39743] [ 1] /lib64/libc.so.6(+0x13e016)[0x7f13f81e0016]
[jazz29:39743] [ 2] /labhome/tvegas/share/srcs/hpcx/modules/ucx/rfs/lib/libucs.so.0(ucs_log_default_handler+0xe9)[0x7f13e91331b9]
[jazz29:39743] [ 3] /labhome/tvegas/share/srcs/hpcx/modules/ucx/rfs/lib/libucs.so.0(ucs_log_dispatch+0xdf)[0x7f13e9133d0f]
[jazz29:39743] [ 4] /labhome/tvegas/share/srcs/hpcx/modules/ucx/rfs/lib/libucs.so.0(+0x4ba4b)[0x7f13e911ca4b]
[jazz29:39743] [ 5] /labhome/tvegas/share/srcs/hpcx/modules/ucx/rfs/lib/libucs.so.0(ucs_async_global_cleanup+0x33)[0x7f13e911add3]
[jazz29:39743] [ 6] /labhome/tvegas/share/srcs/hpcx/modules/ucx/rfs/lib/libucs.so.0(+0x44039)[0x7f13e9115039]
[jazz29:39743] [ 7] /lib64/ld-linux-x86-64.so.2(+0xffca)[0x7f13f8ccbfca]
[jazz29:39743] [ 8] /lib64/libc.so.6(+0x39b69)[0x7f13f80dbb69]
[jazz29:39743] [ 9] /lib64/libc.so.6(+0x39bb7)[0x7f13f80dbbb7]
[jazz29:39743] [10] /lib64/libc.so.6(__libc_start_main+0xfc)[0x7f13f80c43dc]
[jazz29:39743] [11] /hpc/local/benchmarks/daily/next/2023-06-19/hpcx-gcc-redhat7/ompi/tests/osu-micro-benchmarks-5.8/osu_bw[0x401df0]
[jazz29:39743] *** End of error message ***
==== backtrace (tid:  39744) ====
 0 0x000000000013e016 __strcmp_sse42()  :0
 1 0x00000000000621b9 kh_str_hash_equal()  /labhome/tvegas/share/srcs/hpcx/modules/ucx/src/ucs/datastruct/khash.h:456
 2 0x000000000004ba4b ucs_async_signal_global_cleanup()  /labhome/tvegas/share/srcs/hpcx/modules/ucx/src/ucs/async/thread.c:433
 3 0x0000000000049dd3 ucs_async_global_cleanup()  /labhome/tvegas/share/srcs/hpcx/modules/ucx/src/ucs/async/async.c:682
 4 0x0000000000044039 ucs_cleanup()  /labhome/tvegas/share/srcs/hpcx/modules/ucx/src/ucs/sys/init.c:130
 5 0x000000000000ffca _dl_fini()  :0
 6 0x0000000000039b69 __run_exit_handlers()  :0
 7 0x0000000000039bb7 __GI_exit()  :0
 8 0x00000000000223dc __libc_start_main()  ???:0
 9 0x0000000000401df0 _start()  ???:0
```

### DIFF
--- a/src/ucs/debug/log.c
+++ b/src/ucs/debug/log.c
@@ -90,7 +90,6 @@ static ucs_log_func_t ucs_log_handlers[UCS_MAX_LOG_HANDLERS];
 static ucs_spinlock_t ucs_log_global_filter_lock;
 static khash_t(ucs_log_filter) ucs_log_global_filter;
 
-
 static inline int ucs_log_get_pid()
 {
     if (ucs_unlikely(ucs_log_pid == 0)) {
@@ -284,6 +283,7 @@ ucs_log_default_handler(const char *file, unsigned line, const char *function,
     char match;
     int khret;
     char *buf;
+    const char *filename;
 
     if (!ucs_log_component_is_enabled(level, comp_conf) &&
         (level != UCS_LOG_LEVEL_PRINT)) {
@@ -296,7 +296,14 @@ ucs_log_default_handler(const char *file, unsigned line, const char *function,
         /* Add source file name to the hash */
         match  = fnmatch(ucs_global_opts.log_component.file_filter, file, 0) !=
                  FNM_NOMATCH;
-        khiter = kh_put(ucs_log_filter, &ucs_log_global_filter, file, &khret);
+
+        filename = ucs_strdup(file, "log filter filename");
+        if (filename == NULL) {
+            ucs_fatal("cannot allocate log filtering entry for '%s'", file);
+        }
+
+        khiter = kh_put(ucs_log_filter, &ucs_log_global_filter, filename,
+                        &khret);
         ucs_assert((khret == UCS_KH_PUT_BUCKET_EMPTY) ||
                    (khret == UCS_KH_PUT_BUCKET_CLEAR));
         kh_val(&ucs_log_global_filter, khiter) = match;
@@ -544,6 +551,8 @@ void ucs_log_init()
 
 void ucs_log_cleanup()
 {
+    const char *filename;
+
     ucs_assert(ucs_log_initialized);
 
     ucs_log_flush();
@@ -552,7 +561,13 @@ void ucs_log_cleanup()
     }
 
     ucs_spinlock_destroy(&ucs_log_global_filter_lock);
+
+    kh_foreach_key(&ucs_log_global_filter, filename,
+                   { ucs_free((void*)filename); })
+        ; /* code format script wants it there */
+
     kh_destroy_inplace(ucs_log_filter, &ucs_log_global_filter);
+
     ucs_free(ucs_log_file_base_name);
     ucs_log_file_base_name = NULL;
     ucs_log_file           = NULL;


src/ucs/debug/log.c
```diff
@@ -296,10 +355,17 @@ ucs_log_default_handler(const char *file, unsigned line, const char *function,
         /* Add source file name to the hash */
         match  = fnmatch(ucs_global_opts.log_component.file_filter, file, 0) !=
                  FNM_NOMATCH;
-        khiter = kh_put(ucs_log_filter, &ucs_log_global_filter, file, &khret);
-        ucs_assert((khret == UCS_KH_PUT_BUCKET_EMPTY) ||
-                   (khret == UCS_KH_PUT_BUCKET_CLEAR));
-        kh_val(&ucs_log_global_filter, khiter) = match;
+
+        filename = ucs_mem_chunk_alloc_str(file);
```
yosefe, line 359: why not just use ucs_malloc/ucs_strdup?

tvegas1: small allocations that don't need individual tracking, shall we use:
- create: `ucs_strdup()`
- destroy: `kh_foreach_key() { ucs_free() }`, then destroy kh itself

yosefe: Yes

tvegas1: fixed, retested final patch, kh foreach needs `const char *` buf `ucs_free` needs unconstified ptr.

brminich (approved):

src/ucs/debug/log.c
```diff
@@ -296,7 +296,14 @@ ucs_log_default_handler(const char *file, unsigned line, const char *function,
         /* Add source file name to the hash */
         match  = fnmatch(ucs_global_opts.log_component.file_filter, file, 0) !=
                  FNM_NOMATCH;
-        khiter = kh_put(ucs_log_filter, &ucs_log_global_filter, file, &khret);
+
+        filename = ucs_strdup(file, "log filter filename");
```
yosefe, line 300: maybe replace it by ucs_string_set_t?

tvegas1: 1- Currently `ucs_log_global_filter` is used to cache if a given file is a match or not. That boolean is accessed with `kh_val()`. But with `ucs_string_set_t` we can only track presence, so we would need to add one set to track caching state (already cached or not), and another set to track matching value (true or false).

2- If suggestion is to add `ucs_string_set_t` to handle memory allocation/tracking only, seems it does not expose string pointers, seems the function could be added though.

3- IMHO seems simpler like it currently is, let me know which option or any other.

yosefe: ok

yosefe (approved):
@tvegas1 can u pls squash?

yosefe (approved):

Issue #9099, opened 2023-05-23
Segfault if CUDA fails to get human-readable error


Arc676:
### Describe the bug
The `UCT_CUDADRV_FUNC` macro in `src/uct/cuda/base/cuda_iface.h` assumes that `cuGetErrorString` succeeds, but according to the [documentation](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__ERROR.html#group__CUDA__ERROR_1g72758fcaf05b5c7fac5c25ead9445ada), this conversion can fail, in which case the string gets set to the null pointer instead. In the macro, UCX then tries to print this string, causing a segmentation fault by attempting to read a non-existent string.

This happens consistently on program exit during clean-up. After modifying the macro to check if the error string was obtained successfully, UCX reports the following:

```
cuda_copy_iface.c:460  UCX  ERROR cuCtxGetCurrent(&cuda_context)() failed with no error string: 4
 cuda_ipc_iface.c:472  UCX  ERROR cuCtxGetCurrent(&cuda_context)() failed with no error string: 4
```

In the case that prompted this issue, the error turned out to be CUDA error 4, which is `CUDA_ERROR_DEINITIALIZED` according to the CUDA header file.

For context, the modification was just a simple check
```cpp
if (cuGetErrorString(_result, &cu_err_str) == CUDA_SUCCESS)    \
    ucs_log((_log_level), "%s() failed: %d (%s)",              \
            UCS_PP_MAKE_STRING(_func), _result, cu_err_str);   \
else                                                           \
    ucs_log((_log_level),                                      \
            "%s() failed with no error string: %d",            \
            UCS_PP_MAKE_STRING(_func), _result);               \
```

### Steps to Reproduce
Build script:
```sh
if [ ! -d ucx ]; then
    git clone https://github.com/openucx/ucx.git
fi
cd ucx
git fetch
git checkout v1.14.1
./autogen.sh
./configure --with-pic --enable-optimizations --without-cm --without-rocm --without-xpmem --without-ugni \
    --without-java --enable-cma --enable-mt --with-verbs --with-cuda=$CUDA_HOME \
    --with-knem --with-rdmacm \
    --prefix=[...]
```

### Setup and versions
- OS version + CPU architecture: Red Hat 7.9, x86_64

- For GPU related issues:
  - GPU type: GTX 1080
  - Cuda: 
      - Drivers version: 530.30.02
      - Toolkit: 12.1.1

### Additional information (depending on the issue)
[...]

PR #9161, opened 2023-06-23
UCT/CUDA: Don't pass null pointer to logging function

tvegas1:
## What
Make sure that the error string is always non-null.

## Why ?
Crash report #9099.

## How ?
Wrap the call and use similar default string in driver API as done in applicative API.

### DIFF
--- a/src/uct/cuda/base/cuda_iface.c
+++ b/src/uct/cuda/base/cuda_iface.c
@@ -10,6 +10,22 @@
 
 #include "cuda_iface.h"
 
+#include <ucs/sys/string.h>
+
+
+const char *uct_cuda_base_cu_get_error_string(CUresult result)
+{
+    static __thread char buf[64];
+    const char *error_str;
+
+    if (cuGetErrorString(result, &error_str) != CUDA_SUCCESS) {
+        ucs_snprintf_safe(buf, sizeof(buf), "unrecognized error code %d",
+                          result);
+        error_str = buf;
+    }
+
+    return error_str;
+}
 
 ucs_status_t
 uct_cuda_base_query_devices_common(
--- a/src/uct/cuda/base/cuda_iface.h
+++ b/src/uct/cuda/base/cuda_iface.h
@@ -15,6 +15,8 @@
 #include <nvml.h>
 
 
+const char *uct_cuda_base_cu_get_error_string(CUresult result);
+
 #define UCT_CUDA_DEV_NAME       "cuda"
 
 #define UCT_CUDAR_CALL(_log_level, _func, ...) \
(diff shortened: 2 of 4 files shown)


src/uct/cuda/base/cuda_iface.c
```diff
@@ -11,6 +11,13 @@
 #include "cuda_iface.h"
 
 
+void uct_cuda_base_cu_get_error_string(CUresult result, const char **error_str)
+{
+    if (cuGetErrorString(result, error_str) != CUDA_SUCCESS) {
+        *error_str = "unrecognized error code";
```
yosefe, line 17: shall we include the error code in the string?
could use a global/TLS string buffer

src/uct/cuda/base/cuda_iface.h
```diff
@@ -15,6 +15,8 @@
 #include <nvml.h>
 
 
+void uct_cuda_base_cu_get_error_string(CUresult result, const char **error_str);
```
yosefe, line 18: maybe return const char* ?

src/uct/cuda/base/cuda_iface.c
```diff
@@ -11,11 +11,17 @@
...
-void uct_cuda_base_cu_get_error_string(CUresult result, const char **error_str)
+const char *uct_cuda_base_cu_get_error_string(CUresult result)
 {
-    if (cuGetErrorString(result, error_str) != CUDA_SUCCESS) {
-        *error_str = "unrecognized error code";
+    static __thread char buf[64];
+    const char *error_str;
+
+    if (cuGetErrorString(result, &error_str) != CUDA_SUCCESS) {
+        snprintf(buf, sizeof(buf), "unrecognized error code %d", result);
```
yosefe, line 20: ucs_snprintf_safe

src/uct/cuda/base/cuda_iface.c
tvegas1: fixed

src/uct/cuda/base/cuda_iface.h
tvegas1: fixed

src/uct/cuda/base/cuda_iface.c
tvegas1: fixed

src/uct/cuda/base/cuda_iface.c
brminich: why need TLS if UCT is not thread safe?

yosefe: we should support using 2 uct workers from different threads, for example

yosefe (approved):

yosefe (approved):

PR #9168, opened 2023-06-28
UCX: Add support for RISC-V

rbradford:
## What
Adds support for compiling and running UCX on RISC-V (tested under RV64 QEMU.)

## Why ?
Architecture specific code was missing for RISC-V (this adds that) and this
also addresses non RISC-V specific issues (although architecture dependent)
around the ELF symbol address and patching the PLT which are needed for correct
functioning.

## How ?

Tested with UCX gtests and UCC gtests (patch required for UCC) on Linux under
RV64 QEMU. 

- UCS/ARCH/RV64: Add RISC-V support
- UCM: Fix RELOC patching method for shared libraries
- UCM/UTIL/RELOC: Don't rely on RTLD_NEXT for lookup


rbradford:
@yosefe The PR is passing the CI and is ready for review. Please take a look when you can. Thanks!

src/ucm/util/reloc.c
```diff
@@ -366,11 +365,23 @@ static ucs_status_t ucm_reloc_dl_info_get(const struct dl_phdr_info *phdr_info,
         goto out;
     }
 
+    if (symtab < (void*)dl_info->start) {
+        symtab = (char*)symtab + dlpi_addr;
+    }
```
yosefe, line 370: what is this doing?
maybe add a helper function to reduce code duplication?

src/ucm/util/reloc.h
```diff
@@ -55,12 +55,32 @@ static UCS_F_MAYBE_UNUSED
 void* ucm_reloc_get_orig(const char *symbol, void *replacement)
 {
     const char *error;
-    void *func_ptr;
+    void *func_ptr = NULL;
+    int ret;
+    void *dl;
+    Dl_info info;
+    int flags = RTLD_LOCAL | RTLD_NODELETE | RTLD_LAZY;
```
yosefe, line 62: 1. initialized vars should be first
2. can it be static const?

src/ucm/util/reloc.h
```diff
@@ -55,12 +55,32 @@ static UCS_F_MAYBE_UNUSED
...
+        func_ptr = dlsym(dl, symbol);
+        ucm_debug("(libucm) Found symbol %s at %p", symbol, func_ptr);
+        dlclose(dl);
+    }
 
-    func_ptr = dlsym(RTLD_NEXT, symbol);
     if (func_ptr == NULL) {
         (void)dlerror();
         func_ptr = dlsym(RTLD_DEFAULT, symbol);
+        ucm_debug("(Default) Found symbol %s at %p", symbol, func_ptr);
```
yosefe, line 83: remove or: convert to ucm_trace and make it small case

src/ucm/util/reloc.h
```diff
@@ -55,12 +55,32 @@ static UCS_F_MAYBE_UNUSED
...
     const char *error;
-    void *func_ptr;
+    void *func_ptr = NULL;
+    int ret;
+    void *dl;
+    Dl_info info;
+    int flags = RTLD_LOCAL | RTLD_NODELETE | RTLD_LAZY;
+
+    (void)dlerror();
+    ret = dladdr((void*)ucm_reloc_get_orig, &info);
```
yosefe, line 65: can we remove the cast to void*?

src/ucm/util/reloc.h
```diff
@@ -55,12 +55,32 @@ static UCS_F_MAYBE_UNUSED
...
+    if (ret == 0) {
+        ucm_warn("could not find address of current library: %s", dlerror());
+        return NULL;
+    }
+
+    (void)dlerror();
+    dl = dlopen(info.dli_fname, flags);
+    if (dl != NULL) {
+        (void)dlerror();
+        func_ptr = dlsym(dl, symbol);
```
yosefe, line 75: this changes the behavior: before the change the symbol was searched in the caller library (that could be libucm_cuda.so for example) and now it's always searched in libucm.so

src/ucs/arch/rv64/bitops.h
```diff
@@ -0,0 +1,53 @@
...
+*/
+
+#ifndef UCS_ARCH_RV64_BITOPS_H_
+#define UCS_ARCH_RV64_BITOPS_H_
+
+#include <ucs/sys/compiler_def.h>
+#include <stdint.h>
+
+/* __builtin_ctz(0) is undefined, guard against it. */
+#define ctz_safe(n) ((n) ? __builtin_ctz(n) : 32)
```
yosefe, line 14: ucs_ctz_safe

src/ucs/arch/rv64/cpu.h
```diff
@@ -0,0 +1,113 @@
...
+}
+
+static inline uint64_t ucs_arch_read_hres_clock()
+{
+    return ucs_arch_generic_read_hres_clock();
+}
+
+#define ucs_arch_wait_mem ucs_arch_generic_wait_mem
+
+#if !HAVE___CLEAR_CACHE
```
yosefe, line 87: can we remove this `#if` ?

src/ucs/arch/rv64/global_opts.c
```diff
@@ -0,0 +1,47 @@
...
+    {"BUILTIN_MEMCPY_MIN", "auto",
+     "Minimal threshold of buffer length for using built-in memcpy.",
+     ucs_offsetof(ucs_arch_global_opts_t, builtin_memcpy_min),
+     UCS_CONFIG_TYPE_MEMUNITS},
+
+    {"BUILTIN_MEMCPY_MAX", "auto",
+     "Maximal threshold of buffer length for using built-in memcpy.",
+     ucs_offsetof(ucs_arch_global_opts_t, builtin_memcpy_max),
+     UCS_CONFIG_TYPE_MEMUNITS},
+#endif
```
yosefe, line 27: maybe remove? seems there is no special memcpy implementation on rv64

src/ucm/bistro/bistro_rv64.c
```diff
@@ -0,0 +1,119 @@
...
+    if (orig_func_p != NULL) {
+        return UCS_ERR_UNSUPPORTED;
+    }
+
+    status = ucm_bistro_create_restore_point(func_ptr, sizeof(patch), rp);
+    if (UCS_STATUS_IS_ERR(status)) {
+        return status;
+    }
+
+    return ucm_bistro_apply_patch(func_ptr, &patch, sizeof(patch));
```
tvegas1, line 111: could you please check #9148 to add the function `ucm_bistro_patch_lock` for risc-v to enable and use `ucm_bistro_apply_patch_atomic` instead?

rbradford (commented):
@yosefe @tvegas1 Thank you for your responses. I've added additional patches to the PR to respond to your points (I understand you prefer patches only to be squashed when the review is concluded.)

src/ucm/util/reloc.c
evangreen: Right, parsing the headers to get the DT_* value doesn't necessarily get you the final VA of that entity if the library wasn't loaded at 0. So we sanity check to see "is this pointer at least beyond the lowest loaded address we found for this image?", and if not, we add the start to compensate for the offset the binary was actually loaded at. Basically a poor man's relocation.

rbradford:
> @rbradford please look at the prior PR for RISCV support. @yosefe thank you for bringing both efforts together!
> 
> #8246

Yes, this code is derived from that PR and fixed for coding style and bug fixes - I see you've updated it between when I created this PR and now - I thought your original PR was abandoned.

I would appreciate it if you would let me review the code before you finalise it as I think there might still be some minor issues.

rbradford:
@yosefe I've updated this PR based on your feedback - the CI issues seem unrelated to the changes included in it.

ct-clmsn:
@rbradford great work; UCX has changed a lot since #8246 was completed, so much so that I consider #8246 OBE (overcome by events). @yosefe I'd encourage merging this PR and closing out #8246.

This PR has been successfully tested on a sifive unmatched and a pine64/star64 (starfive chip) sbc.

@rbradford I've made a PR on your branch for these files, feel free to ignore. The PR was supposed to update these files with compile guards around OpenMP pragmas.

./src/tools/perf/perftest.c
./src/tools/perf/lib/libperf.c
./test/apps/test_init_mt.c

ct-clmsn:
@rbradford if possible, could you compare #8246's `ucm_bistro_patch` in [https://github.com/openucx/ucx/pull/8246/files#diff-716c0e7eef395234420ea9d54e916cc30b07a0d3080982bd3700faa712dbd759](here) between lines 85 and 92 for correctness against the implementation in this PR? We had issues with the code in this PR.

rbradford:
@ct-clmsn Thank you for testing - I have integrated your OpenMP related patch. Also to avoid the sign extension problem I have switched to a different method for making the jump. By using AUIPC we can store the absolute 64-bit address in the instruction stream as a constant and then load it and jump to it. This avoids all sorts of issues with the sign extension of immediates. If you can - I would appreciate you giving it another test.

ct-clmsn:
@rbradford all is clear on this end; successfully completed several loopback tests on a star64 over ethernet.

src/ucm/util/reloc.h
yosefe: is it fixing a failure that happens with RISC-V? what was that change needed?
I'm worried it may break other use cases..

src/ucs/arch/rv64/bitops.h
```diff
@@ -0,0 +1,41 @@
...
+static UCS_F_ALWAYS_INLINE unsigned __ucs_ilog2_u64(uint64_t n)
+{
+    uint32_t lower, upper;
+
+    upper = n >> 32;
+    lower = n;
+    if (upper == 0) {
+        return __ucs_ilog2_u32(lower);
+    } else {
+        return 32 + __ucs_ilog2_u32(upper);
```
yosefe, line 27: does RISC-V have CLZ for 64-bit?

src/ucs/arch/rv64/global_opts.c
```diff
@@ -0,0 +1,25 @@
...
+#  include "config.h"
+#endif
+
+#include <ucs/arch/global_opts.h>
+#include <ucs/config/parser.h>
+
+ucs_config_field_t ucs_arch_global_opts_table[] = {
+    {NULL}
+};
+
```
yosefe, line 19: remove one space line

src/ucm/util/reloc.c
yosefe: but on other architectures the libraries are also not loaded at 0, and this code works.. is there some different with RISC-V here?
In any case, can use UCS_PTR_BYTE_OFFSET macro to add pointer to offset

src/ucs/arch/rv64/cpu.h
yosefe: ok

src/ucs/arch/rv64/global_opts.c
rbradford: Fixes, thanks!

src/ucs/arch/rv64/bitops.h
rbradford: I discovered that there is a `__builtin_clzll` which always 8 bytes - I switched to that and I switched the ffs fuctions to builtin equivalents. This means that if the hardware + compiler supports the `Zbb` extension (which adds clz instructions) then it can used optimised versions - otherwise it will fallback to a software implementation.

src/ucm/util/reloc.h
rbradford: I understand - since this was specifically seen in with the UCC gtest (where RTLD_NEXT fell through to patching the PLT) i'm going to propose moving this to a separate PR/GitHub issue where we can discuss the behaviour (i'd also like to see if that fallthough to patching the PLT also occurs with the UCC gtest on x86-64.)

src/ucm/util/reloc.c
rbradford: Since this fix was to specifically address `=reloc` mode I would propose that this commit be dropped from this PR and I open another PR specifically about this incorporating some of the feedback and identifying the exact cause of the offset requirement (and why it isn't needed elsewhere.)

rbradford (commented):
Hi @yosefe - i'm proposing to drop the two parts of this PR that are not purely RISC-V enabling and put them in separate PRs for more focussed discussion.

yosefe:
@rbradford per https://github.com/openucx/ucx/wiki/Guidance-for-contributors, please do not force push code review changes since it's not possible to track the diff vs. the previous review round.
Please rebase the branch on the previously reviewed revision and force-push to fix it

rbradford:
> @rbradford per https://github.com/openucx/ucx/wiki/Guidance-for-contributors, please do not force push code review changes since it's not possible to track the diff vs. the previous review round. Please rebase the branch on the previously reviewed revision and force-push to fix it

Done!

src/ucm/util/reloc.c
rbradford: > but on other architectures the libraries are also not loaded at 0, and this code works.. is there some different with RISC-V here? In any case, can use UCS_PTR_BYTE_OFFSET macro to add pointer to offset

@yosefe I was able to identify the root cause of the address vs offset situation which I documented in #9283 and addressed in #9284 taking in the feedback to use `UCS_PTR_BYTE_OFFSET` and consolidate where the addition of the base address needs to happen.

src/ucm/util/reloc.h
yosefe: can we remove it from this PR then?

src/ucm/util/reloc.c
yosefe: can we remove it from this PR then?

yosefe:
@rbradford pls also note the merge conflict

rbradford:
> @rbradford pls also note the merge conflict

Apologies for the delay in response i've been away. @yosefe How would you like me to resolve this? Ordinarily I would rebase onto master since the problem comes from a conflict with a fix that has been landed into master, however you asked me not to rebase. Do you prefer a merge commit from master into this branch (I couldn't find anything in the UCX developer documentation about this.)

yosefe:
> > @rbradford pls also note the merge conflict
> 
> Apologies for the delay in response i've been away. @yosefe How would you like me to resolve this? Ordinarily I would rebase onto master since the problem comes from a conflict with a fix that has been landed into master, however you asked me not to rebase. Do you prefer a merge commit from master into this branch (I couldn't find anything in the UCX developer documentation about this.)

@rbradford please push a merge commit, thanks!

rbradford:
@yosefe I had to force push since the commit title check doesn't like "Revert" commits (but it is happy with "Merge" ones)

src/ucm/util/reloc.c
```diff
@@ -91,6 +91,19 @@ ucm_reloc_get_entry(ElfW(Addr) base, const ElfW(Phdr) *dphdr, ElfW(Sxword) tag)
     return 0;
 }
 
+static void *
+ucm_reloc_get_pointer(ElfW(Addr) base, const ElfW(Phdr) *dphdr, ElfW(Sxword) tag)
+{
+    uintptr_t entry;
```
yosefe, line 97: init entry inline, and add a space line after it

rbradford: Done!

yosefe (approved):
pls squash

yosefe (approved):

PR #9171, opened 2023-06-28
GTEST: Use RTE to yield CPU when configured with valgrind

tvegas1:
## What
Make sure to yield the CPU when running on Valgrind.

## Why ?
Valgrind's memcheck's VM uses actual threads, but serializes them anyways with one global code lock. With the added yield, the producer gets the opportunity to run more often.

For instance the time to run `ud_verbs/test_uct_perf.envelope/2` is:
- without the fix: between 32 seconds and 55 sec
- with the fix, the runtime is a constant 18sec.

There are many tests using this generic infra.

## How ?
Build with `--with-valgrind`, but also need to detect if actually running inside valgrind otherwise some tests might become too slow (`ud/test_ucp_perf.envelope/37`).

### Tested
```
valgrind --tool=memcheck --leak-check=full --track-origins=yes \
    --fair-sched=try --num-callers=25 --error-exitcode=1 --child-silent-after-fork=yes \
    ./test/gtest/gtest --gtest_filter=ud_verbs/test_uct_perf.envelope/2 --gtest_repeat=-1
...
[       OK ] ud_verbs/test_uct_perf.envelope/2 (18542 ms)
[----------] 1 test from ud_verbs/test_uct_perf (18543 ms total)
```

### DIFF
--- a/test/gtest/common/test_perf.cc
+++ b/test/gtest/common/test_perf.cc
@@ -37,7 +37,8 @@ void test_perf::rte_comm::push(const void *data, size_t size) {
 void test_perf::rte_comm::pop(void *data, size_t size,
                               void (*progress)(void *arg), void *arg) {
     bool done = false;
-    do {
+
+    for (;;) {
         pthread_mutex_lock(&m_mutex);
         if (m_queue.length() >= size) {
             memcpy(data, &m_queue[0], size);
@@ -45,10 +46,18 @@ void test_perf::rte_comm::pop(void *data, size_t size,
             done = true;
         }
         pthread_mutex_unlock(&m_mutex);
-        if (!done) {
-            progress(arg);
+
+        if (done) {
+            break;
         }
-    } while (!done);
+
+        progress(arg);
+
+        /* Valgrind runs threads, but serializes them with one global lock */
+        if (RUNNING_ON_VALGRIND) {
+            sched_yield();
+        }
+    }
 }
 
 


brminich (approved):

test/gtest/common/test_helpers.cc
```diff
@@ -895,4 +895,18 @@ void skip_on_address_sanitizer()
 #endif
 }
 
+bool running_inside_valgrind()
```
yosefe, line 898: why just RUNNING_ON_VALGRIND is not enough?

test/gtest/common/test_perf.cc
```diff
@@ -37,18 +37,27 @@ void test_perf::rte_comm::push(const void *data, size_t size) {
...
+
+        if (done) {
+            break;
+        }
+
+        progress(arg);
+
+        /* Valgrind runs threads, but serializes them with one global lock */
+        if (ucs::running_inside_valgrind()) {
+            sched_yield();
```
yosefe, line 58: maybe it would be better to yield every N iterations to reduce syscall overhead?

test/gtest/common/test_helpers.cc
tvegas1: We run tests configured for valgrind twice, once without and once with valgrind wrapper. For the first occurence, the yield is actually killing performance expectations. Saw it on CI `ud/test_ucp_perf.envelope/37` that seems to have started failing.

yosefe: I'd expect RUNNING_ON_VALGRIND to be 0 when the tests are configured with valgrind but running without valgrind

test/gtest/common/test_perf.cc
tvegas1: Double checked:
- current patch: 18.3sec 0.1 time variability
- yield 1/5: 18.9sec + time variability
- yield 1/10: 19.2sec + more time variability

We failed to pop, we progress locally (likely) as much as we can, so it seems better (under valgrind) to allow the other thread to progress in turn before trying to pop the result.

test/gtest/common/test_helpers.cc
tvegas1: actually you're right, reverted that part and pushed, let's see CI, most likely the failure was a coincidence.

yosefe (approved):

yosefe:
@tvegas1 can you pls squash?

tvegas1:
> @tvegas1 can you pls squash?

done, but just noticed other places in tests (uct/ucp sockaddr related, where we create threads), will start another PR for those

yosefe (approved):

PR #9172, opened 2023-06-29
GTEST/COMMON: Reduced performance tests iteration number when running on Valgrind.

rakhmets:
## What
Limited performance tests to one iteration when running on Valgrind.
Removed extra arguments required for performance tests, but for main gtest execution.

### DIFF
--- a/test/gtest/common/test_perf.cc
+++ b/test/gtest/common/test_perf.cc
@@ -210,9 +210,7 @@ void test_perf::test_params_init(const test_spec &test,
         params.max_iter    = test.iters;
     } else {
         params.warmup_iter = 0;
-        params.max_iter    = ucs_min(20u, test.iters /
-                                          ucs::test_time_multiplier() /
-                                          ucs::test_time_multiplier());
+        params.max_iter    = 1;
     }
 
     params.warmup_time     = 100e-3;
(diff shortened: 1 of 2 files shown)


tvegas1:
With this patch, we don't need to merge #9171 and we could simply close it, since it is only improving perf for ucp/uct perf based tests running under valgrind. @yosefe wdyt?

yosefe:
@tvegas1 we should keep running performance tests under valgrind, in order to find memory leaks and other problems in perftest infrastructure, for example.
We can reduce iterations etc to make it finish faster.

test/gtest/common/test_perf.cc
```diff
@@ -210,9 +210,7 @@ void test_perf::test_params_init(const test_spec &test,
         params.max_iter    = test.iters;
     } else {
         params.warmup_iter = 0;
-        params.max_iter    = ucs_min(20u, test.iters /
-                                          ucs::test_time_multiplier() /
-                                          ucs::test_time_multiplier());
+        params.max_iter    = 1;
```
yosefe, line 213: IMO 1 is not enough.. maybe 100?
e.g keep the previous calculation, and add `params.max_iter = ucs_max(params.max_iter, 100)`

rakhmets: Why do we need more than one iteration when running with Valgrind?

PR #9175, opened 2023-06-29
GTEST: Yield CPU during UCT sockaddr tests

tvegas1:
## What
In the UCT sockaddr code, make sure to yield the CPU when running on valgrind.

## Why ?
Valgrind serializes the OS threads anyways so we need to manually yield them to help in client/server senario involving multiple threads.

On the test subset evaluated, the duration is:
- 69 sec without fix
- 17 sec with the fix

## How ?
Make small independent PRs and run/test them individually through CI as they could cause other issues.

### Tested gain
Parameters: `jazz, SHARD=1/4, --gtest_filter=*tcp/test_uct_sockaddr*`
Without fix:
```
[2023-06-29 15:07:45] [==========] 95 tests from 5 test suites ran. (69270 ms total)
[2023-06-29 15:07:45] [  PASSED  ] 95 tests.
```

With fix:
```
[2023-06-29 15:57:34] [==========] 95 tests from 5 test suites ran. (17316 ms total)
[2023-06-29 15:57:34] [  PASSED  ] 95 tests.
```

### DIFF
--- a/test/gtest/uct/uct_test.cc
+++ b/test/gtest/uct/uct_test.cc
@@ -728,6 +728,15 @@ const uct_test::entity& uct_test::ent(unsigned index) const {
     return m_entities.at(index);
 }
 
+unsigned uct_test::progress_mt() const
+{
+    if (RUNNING_ON_VALGRIND) {
+        sched_yield();
+    }
+
+    return progress();
+}
+
 unsigned uct_test::progress() const {
     unsigned count = 0;
     FOR_EACH_ENTITY(iter) {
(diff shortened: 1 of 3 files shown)


test/gtest/uct/uct_test.cc
```diff
@@ -728,6 +728,17 @@ const uct_test::entity& uct_test::ent(unsigned index) const {
...
 
+unsigned uct_test::progress_mt() const
+{
+    unsigned count = progress();
+
+    if (RUNNING_ON_VALGRIND) {
+        sched_yield();
+    }
+
+    return count;
```
yosefe, line 739: maybe just
```
{
  if (RUNNING_ON_VALGRIND) {
   sched_yield();
  }
  return progress();
}
```

brminich (approved):

yosefe (approved):

PR #9176, opened 2023-07-02
AZP: Update tags for cuda11 images

Alexey-Rivkin:
## What
Update tags for cuda11 images, as older versions were removed from Dockerhub.
Also replaced non-existing Centos8 images with UBI8 alternative.


Alexey-Rivkin:
Should we port it to v1.15.x?

yosefe (approved):

yosefe:
> Should we port it to v1.15.x?

Yes please

PR #9182, opened 2023-07-03
TOOLS/PERF: removing allocators and using uct api instead

roiedanino:
## What
Memory allocation using `uct_mem_alloc` / `uct_iface_mem_alloc` in `ucx_perftest`, replacing the use of allocators which didn't use UCX APIs.

## Why ?
We want to use UCX APIs when measuring the performance of the different memory types such as CUDA and ROCM.

### DIFF
--- a/src/tools/perf/lib/uct_tests.cc
+++ b/src/tools/perf/lib/uct_tests.cc
@@ -19,6 +19,7 @@
 #include "libperf_int.h"
 
 #include <limits>
+#include <vector>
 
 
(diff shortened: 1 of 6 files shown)


src/tools/perf/lib/uct_tests.cc
```diff
@@ -274,11 +269,11 @@ class uct_perf_test_runner {
             if ((TYPE == UCX_PERF_TEST_TYPE_PINGPONG) ||
                 (TYPE == UCX_PERF_TEST_TYPE_PINGPONG_WAIT_MEM)) {
                 /* Put the control word at the latest byte of the IOV message */
+
```
roiedanino, line 272: delete

PR #9187, opened 2023-07-04
TOOLS/PERF: Fix perftest memory access flags

yosefe:
## Why
Without this fix, KSM-atomic can be created without atomic access enabled. 
As a result, `rc_mlx5/test_uct_loopback.envelope/0` fails on rock cluster when running atomic tests.

### DIFF
--- a/src/tools/perf/lib/libperf_memory.c
+++ b/src/tools/perf/lib/libperf_memory.c
@@ -221,9 +221,28 @@ ucs_status_t uct_perf_test_alloc_mem(ucx_perf_context_t *perf)
 
     /* TODO use params->alignment  */
 
-    flags = (params->flags & UCX_PERF_TEST_FLAG_MAP_NONBLOCK) ?
-             UCT_MD_MEM_FLAG_NONBLOCK : 0;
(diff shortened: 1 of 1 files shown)


src/tools/perf/lib/libperf_memory.c
```diff
@@ -223,7 +223,7 @@ ucs_status_t uct_perf_test_alloc_mem(ucx_perf_context_t *perf)
 
     flags = (params->flags & UCX_PERF_TEST_FLAG_MAP_NONBLOCK) ?
              UCT_MD_MEM_FLAG_NONBLOCK : 0;
-    flags |= UCT_MD_MEM_ACCESS_RMA;
+    flags |= UCT_MD_MEM_ACCESS_RMA | UCT_MD_MEM_ACCESS_REMOTE_ATOMIC;
```
evgeny-leksikov, line 226: maybe check `params->command`?

yosefe: fixed

evgeny-leksikov (approved):

PR #9188, opened 2023-07-05
UCP: Wire compatibility issues fixes

brminich:
## What
- Fix wire compatibility testing + extend the testing scope
- Enable sa_data v2 and worker address v2 by default
- Fix several wire_compatibility issues

## Why ?
To preserve wire_compatibility with older UCX versions

replaces #9048


brminich:
replaced by #9273

PR #9191, opened 2023-07-06
UCP: Improve ep config info

brminich:
## What
Improve ep config info, so that every unique lane config field is printed (seg size is currently missing)

## Why ?
Better debugging experience when ep reconfig error happens

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -3054,20 +3054,22 @@ void ucp_ep_config_lane_info_str(ucp_worker_h worker,
     uct_tl_resource_desc_t *rsc;
     ucp_rsc_index_t rsc_index;
     ucp_md_index_t dst_md_index;
+    ucp_md_index_t md_index;
     ucp_rsc_index_t cmpt_index;
     unsigned path_index;
     int prio;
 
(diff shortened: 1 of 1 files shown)


brminich:
@yosefe

src/ucp/core/ucp_ep.c
```diff
@@ -3054,30 +3054,33 @@ void ucp_ep_config_lane_info_str(ucp_worker_h worker,
...
         ucs_string_buffer_appendf(strbuf, "addr[%d].", addr_indices[lane]);
     }
 
     dst_md_index = key->lanes[lane].dst_md_index;
     cmpt_index   = ucp_ep_config_get_dst_md_cmpt(key, dst_md_index);
-    ucs_string_buffer_appendf(strbuf, "md[%d]/%s/sysdev[%d]", dst_md_index,
-                              context->tl_cmpts[cmpt_index].attr.name,
-                              key->lanes[lane].dst_sys_dev);
+    ucs_string_buffer_appendf(
+         strbuf, "md[%d]/%s/sysdev[%d] (seg %zu)", dst_md_index,
```
yosefe, line 3081: maybe remove ( ) around seg size?

yosefe (approved):

PR #9199, opened 2023-07-07
UCT/CUDA/BASE: Query GPUs using NVML.

rakhmets:
## What
Query number of GPUs using NVML interfaces to query all GPUs ignoring CUDA_VISIBLE_DEVICES variable.

## Why ?
To properly handle cases when different processes using different CUDA_VISIBLE_DEVICES on the same host.

### DIFF
--- a/src/uct/cuda/base/cuda_md.c
+++ b/src/uct/cuda/base/cuda_md.c
@@ -12,7 +12,6 @@
 
 #include <ucs/sys/module.h>
 #include <ucs/sys/string.h>
-#include <cuda.h>
 
 
 void uct_cuda_base_get_sys_dev(CUdevice cuda_device,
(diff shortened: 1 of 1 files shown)


src/ucp/proto/proto_multi.inl
```diff
@@ -66,6 +66,11 @@ ucp_proto_multi_max_payload(ucp_request_t *req,
     size_t max_frag = lpriv->max_frag - hdr_size;
     size_t max_payload;
 
+    ucs_assertv(lpriv->max_frag > hdr_size,
```
changchengx, line 69: Is this done in https://github.com/openucx/ucx/pull/9211?

rakhmets: Yes, it is.
This pull request mark as `WIP-DNM`. It will be ready to review after I'll remove the label.
This cherry-pick was added just in an attempt to catch the unstable reproducible issue.

PR #9204, opened 2023-07-09
INFRA: Log zombie kill events

Alexey-Rivkin:
## What
Log zombie-killing events.

## Why ?
Suspected job kills by zombie-killer script.


buildlib/tools/zombie_killer.sh
```diff
@@ -8,6 +13,7 @@ if [ -z "$out" ]; then
 else
   for i in $out ; do
     kill -9 "$i"
+    logger "Killed zombie" "$i"
```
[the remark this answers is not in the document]
Alexey-Rivkin: Nice cache, thank you Thomas!
Changed accordingly.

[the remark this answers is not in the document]
Alexey-Rivkin: image
Found a great book on the subject 😉

buildlib/tools/zombie_killer.sh
```diff
@@ -1,12 +1,18 @@
 #!/bin/bash -eE
 
+logger() {
```
yosefe, line 3: can we use the system "logger" command that will write to /var/log/messages and add timestamps etc?
https://man7.org/linux/man-pages/man1/logger.1.html

Alexey-Rivkin: Thanks Yossi, I'm learning something new every day. 
Modified accordingly.

tvegas1: I think the bash function `logger` cannot be the same as `/bin/logger`, as it will call itself.

```
$ cat a.sh
#!/bin/bash
logger() {
    logger -h
}
logger
$ time bash a.sh
<ctrl-c>
real    0m5.192s
```

Alexey-Rivkin: Definitely, was just about to change it myself, thanks!

buildlib/tools/zombie_killer.sh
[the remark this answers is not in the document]
tvegas1: bit extreme..

tvegas1 (approved):

yosefe (approved):

PR #9208, opened 2023-07-10
CONFIG/M4: Separated library flags for CUDA, CUDA RT, and NVML.

rakhmets:
## What
Separated library flags for CUDA, CUDA RT, and NVML.

## Why ?
To use only required libraries during building modules.
The changes facilitate switching to CUDA Driver API in UCT.


yosefe (approved):
pls squash

brminich (approved):

rakhmets:
> is it expected that UCX can be built on the system without `nvidia-ml` and have unresolved dependencies?
> 
> ```
> ldd <ucx_install>/lib/ucx/libuct_cuda.so
> ...
> libcuda.so.1 => not found
> ...
> libnvidia-ml.so.1 => not found
> ...
> ```

If  `nvidia-ml` is absent on the system:
1. If UCX is configured with default option `--with-cuda=guess`, then CUDA support is disabled.
2. If CUDA support is explicitly requested, then configure will fail with error `libnvidia-ml not found. Install appropriate nvidia-driver package`.

evgeny-leksikov:
> 2. is not true

It seems like environment magic related to container and/or platform.
all missed dependencies are found in 
```
/etc/alternatives/cuda/lib64/stubs/
```

yosefe:
> > 2. is not true
> 
> It seems like environment magic related to container and/or platform. all missed dependencies are found in
> 
> ```
> /etc/alternatives/cuda/lib64/stubs/
> ```

@evgeny-leksikov so are we good with this PR?

evgeny-leksikov (approved):

PR #9220, opened 2023-07-17
UCP/CORE: Check length of user header in ucp_am_send_nbx.

rakhmets:
## What
Added check for the length of user defined Active Message header.
`ucp_am_send_nbx` returns error if the length of the user header is greater than the length the maximum allowed header size, which is available via `ucp_worker_query`.

Added gtest.

Removed printing the return value of `ucp_am_max_header_size` in assert in method `ucp_proto_multi_max_payload`. Because the value is a feature of Active Message, but the method is not.

## Why ?
Follow up on https://github.com/openucx/ucx/pull/9211

### DIFF
--- a/src/ucp/core/ucp_am.c
+++ b/src/ucp/core/ucp_am.c
@@ -87,52 +87,6 @@ void ucp_am_ep_cleanup(ucp_ep_h ep)
                    " dropped on ep %p", ep->worker, count, ep);
 }
 
-size_t ucp_am_max_header_size(ucp_worker_h worker)
-{
-    ucp_context_h context = worker->context;
-    uct_iface_attr_t *if_attr;
-    ucp_rsc_index_t iface_id;
-    size_t max_am_header, max_uct_fragment;
-    size_t max_rts_size, max_ucp_header;
-
-    if (!(context->config.features & UCP_FEATURE_AM)) {
-        return 0ul;
-    }
-
-    max_am_header  = SIZE_MAX;
-    max_rts_size   = sizeof(ucp_rndv_rts_hdr_t) +
-                     ucp_rkey_packed_size(context, UCS_MASK(context->num_mds),
-                                          UCS_SYS_DEVICE_ID_UNKNOWN, 0);
-    max_ucp_header = ucs_max(max_rts_size, UCP_AM_FIRST_FRAG_META_LEN);
-
-    /* Make sure maximal AM header can fit into one bcopy fragment
-     * together with RTS or first eager header (whatever is bigger)
-     */
-    for (iface_id = 0; iface_id < worker->num_ifaces; ++iface_id) {
-        if_attr = &worker->ifaces[iface_id]->attr;
-
-        /* UCT_IFACE_FLAG_AM_BCOPY is required by UCP AM feature, therefore
-         * at least one interface should support it.
-         * Make sure that except user header single UCT fragment can fit
-         * first fragment header and footer and at least 1 byte of data. It is
-         * needed to correctly use generic AM based multi-fragment protocols,
-         * which expect some amount of payload to be packed to the first
-         * fragment.
-         * TODO: fix generic AM based multi-fragment protocols, so that this
-         * trick is not needed.
-         */
-        if (if_attr->cap.flags & UCT_IFACE_FLAG_AM_BCOPY) {
-            max_uct_fragment = ucs_max(if_attr->cap.am.max_bcopy,
-                                       max_ucp_header - 1) - max_ucp_header - 1;
-            max_am_header    = ucs_min(max_am_header, max_uct_fragment);
-        }
-    }
-
-    ucs_assert(max_am_header < SIZE_MAX);
-
-    return ucs_min(max_am_header, UINT32_MAX);
-}
-
 static void ucp_am_rndv_send_ats(ucp_worker_h worker, ucp_rndv_rts_hdr_t *rts,
                                  ucs_status_t status)
 {
@@ -969,6 +923,19 @@ ucp_am_params_check_memh(const ucp_request_param_t *param, uint32_t *flags_p)
     return UCS_OK;
 }
 
+static UCS_F_ALWAYS_INLINE ucs_status_t
+ucp_am_send_nbx_check_header_length(ucp_worker_h worker, size_t header_length)
+{
+    if (ENABLE_PARAMS_CHECK && (header_length > worker->max_am_header)) {
+        ucs_error("active message header length (%zi) is greater than maximum "
+                  "allowed header size (%zi)",
+                  header_length, worker->max_am_header);
+        return UCS_ERR_INVALID_PARAM;
+    }
+
+    return UCS_OK;
+}
+
 UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
                  (ep, id, header, header_length, buffer, count, param),
                  ucp_ep_h ep, unsigned id, const void *header,
@@ -992,6 +959,12 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
     UCP_REQUEST_CHECK_PARAM(param);
 
     UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);
+
+    status = ucp_am_send_nbx_check_header_length(worker, header_length);
+    if (status != UCS_OK) {
+        ret = UCS_STATUS_PTR(status);
+        goto out;
+    }
 
     flags     = ucp_request_param_flags(param);
     attr_mask = param->op_attr_mask &
--- a/src/ucp/core/ucp_worker.c
+++ b/src/ucp/core/ucp_worker.c
@@ -2324,6 +2324,52 @@ void ucp_worker_create_vfs(ucp_context_h context, ucp_worker_h worker)
                             "counters/ep_failures");
 }
 
+static void ucp_worker_set_max_am_header(ucp_worker_h worker)
+{
+    ucp_context_h context = worker->context;
+    uct_iface_attr_t *if_attr;
+    ucp_rsc_index_t iface_id;
+    size_t max_am_header, max_rts_size, max_ucp_header, max_uct_fragment;
+
+    if (!(context->config.features & UCP_FEATURE_AM)) {
+        worker->max_am_header = 0ul;
+        return;
+    }
+
+    max_am_header  = SIZE_MAX;
+    max_rts_size   = sizeof(ucp_rndv_rts_hdr_t) +
+                     ucp_rkey_packed_size(context, UCS_MASK(context->num_mds),
+                                          UCS_SYS_DEVICE_ID_UNKNOWN, 0);
+    max_ucp_header = ucs_max(max_rts_size, UCP_AM_FIRST_FRAG_META_LEN);
+
+    /* Make sure maximal AM header can fit into one bcopy fragment
+     * together with RTS or first eager header (whatever is bigger)
+     */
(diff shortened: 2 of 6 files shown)


src/ucp/core/ucp_am.c
```diff
@@ -969,6 +969,27 @@ ucp_am_params_check_memh(const ucp_request_param_t *param, uint32_t *flags_p)
     return UCS_OK;
 }
 
+static int
+ucp_am_send_nbx_check_header_length(ucp_worker_h worker, size_t header_length)
+{
+    ucp_worker_attr_t attr;
+
+    attr.field_mask = UCP_WORKER_ATTR_FIELD_MAX_AM_HEADER;
+    if (ucp_worker_query(worker, &attr) != UCS_OK) {
```
yosefe, line 978: IMO better use ucp_am_max_header_size. Or maybe even cache max header size on the worker?

rakhmets: I think this is a better way in terms of code support. `ucp_am_send_nbx_check_header_length` is less sensitive to changes in other methods in this case.
A user obtains the value using `ucp_worker_query`. Which is an interface. `ucp_am_max_header_size` is an internal method. The implementation of `ucp_worker_query` may change, `ucp_am_max_header_size` may change its name, or may be gone. These are more likely events than the change of the interface.

brminich: so, maybe cache then?

rakhmets: ok, I will cache the value on the worker.

rakhmets: Also moved `ucp_am_max_header_size` to worker.c, and renamed it to `ucp_worker_get_max_am_header`, because the method operates with a worker handler.

src/ucp/core/ucp_worker.c
```diff
@@ -2835,6 +2839,61 @@ static ucs_status_t ucp_worker_address_pack(ucp_worker_h worker,
...
+
+static size_t ucp_worker_get_max_am_header(ucp_worker_h worker)
+{
+    ucp_context_h context = worker->context;
+    uct_iface_attr_t *if_attr;
+    ucp_rsc_index_t iface_id;
+    size_t max_am_header, max_uct_fragment;
+    size_t max_rts_size, max_ucp_header;
+
+    if (worker->max_am_header != UCP_WORKER_MAX_AM_HEADER_UNINIT) {
```
yosefe, line 2851: maybe just calculate it during ucp worker initialization?

src/ucp/core/ucp_am.c
yosefe: IMO we should not call API functions from internal functions, it's violating SW layers

src/ucp/core/ucp_am.c
```diff
@@ -969,6 +923,27 @@ ucp_am_params_check_memh(const ucp_request_param_t *param, uint32_t *flags_p)
...
+        ucs_error("failed to query maximum allowed header size");
+        return 0;
+    }
+
+    if (header_length > attr.max_am_header) {
+        ucs_error("active message header length (%zi) is greater than maximum "
+                  "allowed header size (%zi)",
+                  header_length, attr.max_am_header);
+        return 0;
```
yosefe, line 941: since this function prints an error, it should also return status code (not 1/0)
we can make it inline, and do something like:
```c
if (ENABLE_PARAMS_CHECK && (header_length > worker->max_am_header)) {
    ucs_error(...);
    return UCS_ERR_INVALID_PARAM;
}

return UCS_OK;
```

```c
status = ucm_am_check_header_length(..);
if (status != UCS_OK) {
    ret = UCS_STATUS_PTR(status);
    goto out;
}
```

src/ucp/core/ucp_worker.c
```diff
@@ -2835,6 +2839,61 @@ static ucs_status_t ucp_worker_address_pack(ucp_worker_h worker,
                             address_length_p, (void**)address_p);
 }
 
+
+static size_t ucp_worker_get_max_am_header(ucp_worker_h worker)
```
brminich: minor: `ucp_worker_max_am_header` seems to be clear enough

src/ucp/core/ucp_worker.c
```diff
@@ -2835,6 +2839,61 @@ static ucs_status_t ucp_worker_address_pack(ucp_worker_h worker,
...
+         * trick is not needed.
+         */
+        if (if_attr->cap.flags & UCT_IFACE_FLAG_AM_BCOPY) {
+            max_uct_fragment = ucs_max(if_attr->cap.am.max_bcopy,
+                                       max_ucp_header - 1) - max_ucp_header - 1;
+            max_am_header    = ucs_min(max_am_header, max_uct_fragment);
+        }
+    }
+
+    ucs_assert(max_am_header < SIZE_MAX);
```
brminich: this is not consistent now, because we use `INT32_MAX` as uninitialized value, so we implicitly assume that the size is less than `INT32_MAX`, right? Maybe use `SIZE_MAX` as uninitialized value instead?

src/ucp/core/ucp_worker.c
rakhmets: done

src/ucp/core/ucp_am.c
rakhmets: fixed

src/ucp/core/ucp_am.c
rakhmets: using the field of UCP worker directly

src/ucp/core/ucp_am.c
```diff
@@ -968,9 +960,9 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
 
     UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);
 
-    if (ENABLE_PARAMS_CHECK &&
-        !ucp_am_send_nbx_check_header_length(worker, header_length)) {
-        ret = UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM);
+    status = ucp_am_send_nbx_check_header_length(worker, header_length);
```
brminich, line 963: maybe no need for a separate check function now? Seems to be a simple check and you will not need to check return status here

src/ucp/core/ucp_am.c
```diff
@@ -993,6 +960,12 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
 
     UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);
 
+    status = ucp_am_send_nbx_check_header_length(worker, header_length);
+    if (status != UCS_OK) {
+        ret = UCS_STATUS_PTR(status);
+        goto out;
+    }
```
rakhmets, line 967:
```suggestion
    if (ENABLE_PARAMS_CHECK && (header_length > worker->max_am_header)) {
        ucs_error("active message header length (%zi) is greater than maximum "
                  "allowed header size (%zi)",
                  header_length, worker->max_am_header);
        ret = UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM);
        goto out;
    }
```

@brminich the variant with additional method looks prettier for me.
And there is another check in this method, so it looks also more unified.

src/ucp/core/ucp_worker.c
```diff
@@ -2520,6 +2567,8 @@ ucs_status_t ucp_worker_create(ucp_context_h context,
 
     ucp_worker_create_vfs(context, worker);
 
+    ucp_worker_set_max_am_header(worker);
```
yosefe, line 2570: i would do it before line 2558
in case vfs would show max am header

src/ucp/core/ucp_worker.h
```diff
@@ -358,6 +358,8 @@ typedef struct ucp_worker {
         /* Number of failed endpoints */
         uint64_t                     ep_failures;
     } counters;
+
+    size_t max_am_header; /* Maximum allowed header size used by UCP AM API */
```
yosefe, line 362: maybe move it after line 319 or somewhere in that area

src/ucp/core/ucp_worker.c
rakhmets: fixed

src/ucp/core/ucp_worker.c
```diff
@@ -2854,7 +2904,8 @@ ucs_status_t ucp_worker_query(ucp_worker_h worker,
     }
 
     if (attr->field_mask & UCP_WORKER_ATTR_FIELD_MAX_AM_HEADER) {
-        attr->max_am_header = ucp_am_max_header_size(worker);
+        ucp_worker_set_max_am_header(worker);
```
brminich, line 2907: no need to set here, because query is not possible when worker is not created yet

src/ucp/core/ucp_worker.c
```diff
@@ -2324,6 +2324,54 @@ void ucp_worker_create_vfs(ucp_context_h context, ucp_worker_h worker)
...
+            max_uct_fragment = ucs_max(if_attr->cap.am.max_bcopy,
+                                       max_ucp_header - 1) -
+                               max_ucp_header - 1;
+            max_am_header    = ucs_min(max_am_header, max_uct_fragment);
+        }
+    }
+
+    worker->max_am_header = (max_am_header != SIZE_MAX) ?
+                                    ucs_min(max_am_header, UINT32_MAX) :
+                                    0;
```
brminich, line 2372: imo this style suggestion is ugly, i'd avoid this style. Something like below is better:

```suggestion
    worker->max_am_header = (max_am_header != SIZE_MAX) ?
                             ucs_min(max_am_header, UINT32_MAX) : 0;
```

src/ucp/core/ucp_worker.c
rakhmets: sorry, that was added for debug purpose.

src/ucp/core/ucp_worker.c
rakhmets: Reverted.

brminich (dismissed):

src/ucp/core/ucp_worker.c
```diff
@@ -2324,6 +2324,49 @@ void ucp_worker_create_vfs(ucp_context_h context, ucp_worker_h worker)
...
 
+static void ucp_worker_set_max_am_header(ucp_worker_h worker)
+{
+    ucp_context_h context = worker->context;
+    uct_iface_attr_t *if_attr;
+    ucp_rsc_index_t iface_id;
+    size_t max_rts_size, max_ucp_header, max_uct_fragment;
+
+    worker->max_am_header = SIZE_MAX;
+    if (!(context->config.features & UCP_FEATURE_AM)) {
```
brminich, line 2335: should be 0 here and in case no interface with AM_BCOPY found

rakhmets: Fixed.

src/ucp/core/ucp_worker.c
```diff
@@ -2366,9 +2366,8 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
         }
     }
 
-    ucs_assert(max_am_header < SIZE_MAX);
-
-    worker->max_am_header = ucs_min(max_am_header, UINT32_MAX);
+    worker->max_am_header = (max_am_header != SIZE_MAX) ?
```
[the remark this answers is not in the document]
rakhmets: keep the check to set 0, according to https://github.com/openucx/ucx/pull/9220/files#r1266911253

brminich (approved):

yosefe (approved):

yosefe (approved):

PR #9222, opened 2023-07-18
UCT: no need to specify default CPU mask field without mask CPU

changchengx:
## What
There's no need to specify CPU mask if using the default cpu mask.

### DIFF
--- a/src/tools/info/tl_info.c
+++ b/src/tools/info/tl_info.c
@@ -129,8 +129,7 @@ static void print_iface_info(uct_worker_h worker, uct_md_h md,
         .field_mask            = UCT_IFACE_PARAM_FIELD_OPEN_MODE   |
                                  UCT_IFACE_PARAM_FIELD_DEVICE      |
                                  UCT_IFACE_PARAM_FIELD_STATS_ROOT  |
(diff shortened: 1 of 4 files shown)


examples/uct_hello_world.c
```diff
@@ -253,15 +253,13 @@ static ucs_status_t init_iface(char *dev_name, char *tl_name,
     params.field_mask           = UCT_IFACE_PARAM_FIELD_OPEN_MODE   |
                                   UCT_IFACE_PARAM_FIELD_DEVICE      |
                                   UCT_IFACE_PARAM_FIELD_STATS_ROOT  |
-                                  UCT_IFACE_PARAM_FIELD_RX_HEADROOM |
-                                  UCT_IFACE_PARAM_FIELD_CPU_MASK;
+                                  UCT_IFACE_PARAM_FIELD_RX_HEADROOM;
```
yosefe, line 256: IMO better not remove it from hello world - since it's an example

changchengx: Done.

changchengx:
Checked all the failed test cases, they're not related with this PR.

yosefe (approved):

PR #9224, opened 2023-07-18
UCP: Fix fp8 packing

brminich:
## What
Fix fp8 packing function

## Why ?
Values with exponent equal to `min_exp - 1` are packed (and therefore unpacked) with 0 exponent, while they need to be rounded up to `min_exp` value
E.g.:
UCP packs BW as `UCS_FP8_DECLARE_TYPE(BANDWIDTH, 512 * UCS_MBYTE, 4 * UCS_TBYTE)`
So, without this fix, values from  `[256 * UCS_MBYTE, 512 * UCS_MBYTE)` range are packed as zeroes.

### DIFF
--- a/test/gtest/ucs/test_type.cc
+++ b/test/gtest/ucs/test_type.cc
@@ -118,9 +118,12 @@ UCS_TEST_F(test_type, pack_float) {
     EXPECT_TRUE(isnan(unpacked));
 
     /* Below min -> min */
-    EXPECT_EQ(UCS_FP8_UNPACK(TEST_LATENCY,
-                             UCS_FP8_PACK(TEST_LATENCY, UCS_BIT(7))),
-              UCS_FP8_UNPACK(TEST_LATENCY, UCS_FP8_PACK(TEST_LATENCY, 15)));
+    for (uint64_t min_val = UCS_BIT(0); min_val <= UCS_BIT(7); min_val <<= 1) {
+        UCS_TEST_MESSAGE << " Pack/unpack " << min_val;
+        EXPECT_EQ(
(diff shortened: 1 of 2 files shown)


test/gtest/ucs/test_type.cc
```diff
@@ -118,9 +118,12 @@ UCS_TEST_F(test_type, pack_float) {
     EXPECT_TRUE(isnan(unpacked));
 
     /* Below min -> min */
-    EXPECT_EQ(UCS_FP8_UNPACK(TEST_LATENCY,
-                             UCS_FP8_PACK(TEST_LATENCY, UCS_BIT(7))),
-              UCS_FP8_UNPACK(TEST_LATENCY, UCS_FP8_PACK(TEST_LATENCY, 15)));
+    for (uint64_t min_val = UCS_BIT(0); min_val <= UCS_BIT(7); min_val <<= 1) {
```
gleon99, line 121: Are you sure there's any real point in looping here, rather than just testing a single value?

test/gtest/ucs/test_type.cc
```diff
@@ -118,9 +118,12 @@ UCS_TEST_F(test_type, pack_float) {
...
+        UCS_TEST_MESSAGE << " Pack/unpack " << min_val;
```
gleon99, line 122: This test is mainly w/o messages. Any reason to add msg in this particular place?

brminich: otherwise particular min_val is not printed - only it's wrongly unpacked value (0)

yosefe (approved):

yosefe (approved):

PR #9227, opened 2023-07-20
UCT/CUDA: Add profiling to Cuda runtime API calls

yosefe:
## Why
Get more detailed information about Cuda function calls and their timings
[[Nvidia internal tracker](https://redmine.mellanox.com/issues/3519158)]

### DIFF
--- a/src/uct/cuda/base/cuda_md.c
+++ b/src/uct/cuda/base/cuda_md.c
@@ -69,13 +69,12 @@ uct_cuda_base_query_md_resources(uct_component_t *component,
     const unsigned sys_device_priority = 10;
     ucs_sys_device_t sys_dev;
     CUdevice cuda_device;
-    cudaError_t cudaErr;
     ucs_status_t status;
     char device_name[10];
     int num_gpus;
 
-    cudaErr = cudaGetDeviceCount(&num_gpus);
-    if ((cudaErr != cudaSuccess) || (num_gpus == 0)) {
+    status = UCT_CUDA_CALL(UCS_LOG_LEVEL_DIAG, cudaGetDeviceCount, &num_gpus);
+    if ((status != UCS_OK) || (num_gpus == 0)) {
         return uct_md_query_empty_md_resource(resources_p, num_resources_p);
     }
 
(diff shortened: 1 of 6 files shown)


yosefe:
@rakhmets can you pls review?

src/uct/cuda/base/cuda_md.c
```diff
@@ -74,7 +74,7 @@ uct_cuda_base_query_md_resources(uct_component_t *component,
     char device_name[10];
     int num_gpus;
 
-    cudaErr = cudaGetDeviceCount(&num_gpus);
+    cudaErr = UCS_PROFILE_CALL_ALWAYS(cudaGetDeviceCount, &num_gpus);
     if ((cudaErr != cudaSuccess) || (num_gpus == 0)) {
```
rakhmets, line 78: I think we should print error here in case of failed CUDA call.
```suggestion
     status = UCT_CUDA_CALL_LOG_ERR(cudaGetDeviceCount, &num_gpus);
    if ((status != UCS_OK) || (num_gpus == 0)) {
```

yosefe: maybe ucs_diag?

rakhmets: This is usually the first CUDA Runtime call from UCX.
If the call returns error, then the query will return empty MD, and CUDA transports will not be supported.
We should print an error here, IMO.

yosefe: So far the policy is to not print errors if something is not supported.
Do you know in which cases cudaGetDeviceCount would return error?
@Akshay-Venkatesh WDYT?

yosefe: @rakhmets fixed (force pushed due to small pr)

src/uct/cuda/base/cuda_md.c
```diff
@@ -74,7 +74,7 @@ uct_cuda_base_query_md_resources(uct_component_t *component,
     char device_name[10];
     int num_gpus;
 
-    cudaErr = cudaGetDeviceCount(&num_gpus);
+    cudaErr = UCT_CUDA_CALL(UCS_LOG_LEVEL_DIAG, cudaGetDeviceCount, &num_gpus);
```
rakhmets, line 77: Please use ucs_status_t here, and in the following check `status != UCS_OK`.

rakhmets (approved):

brminich (approved):

PR #9228, opened 2023-07-20
UCM/BISTRO: Fix conditional jump translation

yosefe:
## Why
Fix conditional jump patching: push instruction sign-extends the 32-bit operand to 64-bit, so "push, push, ret" does not really jump to the destination address. Replace by indirect absolute jump "jmp *(%rip), .long addr"

To be backported for v1.15.x


yosefe:
@brminich can you pls review?

PR #9229, opened 2023-07-20
UCT/IB: add flow_label info in ah_attr output string

changchengx:
## What
None-zero flow_label value is used to affect the source udp port for none rc devx qp transport.
output flow_label value for debug purpose.

### DIFF
--- a/src/uct/ib/base/ib_device.c
+++ b/src/uct/ib/base/ib_device.c
@@ -1304,7 +1304,8 @@ const char* uct_ib_ah_attr_str(char *buf, size_t max,
         p += strlen(p);
         uct_ib_gid_str(&ah_attr->grh.dgid, p, endp - p);
         p += strlen(p);
-        snprintf(p, endp - p, " sgid_index=%d traffic_class=%d",
+        snprintf(p, endp - p, " flow_label=0x%x sgid_index=%d "
+                 "traffic_class=%d", ah_attr->grh.flow_label,
                  ah_attr->grh.sgid_index, ah_attr->grh.traffic_class);
     }
 


src/uct/ib/base/ib_device.c
```diff
@@ -1294,17 +1294,17 @@ const char* uct_ib_ah_attr_str(char *buf, size_t max,
     char *p    = buf;
     char *endp = buf + max;
 
-    snprintf(p, endp - p, "dlid=%d sl=%d port=%d src_path_bits=%d",
-             ah_attr->dlid, ah_attr->sl,
-             ah_attr->port_num, ah_attr->src_path_bits);
+    snprintf(p, endp - p, "dlid=0x%x flow_label=0x%x sl=0x%x port=%d "
+             "src_path_bits=%d", ah_attr->dlid, ah_attr->grh.flow_label,
```
yosefe, line 1298: it should be in line 1307 because it's part of ah_attr->grr

src/uct/ib/base/ib_device.c
```diff
@@ -1294,17 +1294,19 @@ const char* uct_ib_ah_attr_str(char *buf, size_t max,
     char *p    = buf;
     char *endp = buf + max;
 
-    snprintf(p, endp - p, "dlid=%d sl=%d port=%d src_path_bits=%d",
+    snprintf(p, endp - p, "dlid=0x%x sl=0x%x port=%d src_path_bits=%d",
```
yosefe, line 1297: why need to print LID and SL as hex? in other places they are printed as decimal

src/uct/ib/base/ib_device.c
```diff
@@ -1294,17 +1294,19 @@ const char* uct_ib_ah_attr_str(char *buf, size_t max,
...
              ah_attr->dlid, ah_attr->sl,
              ah_attr->port_num, ah_attr->src_path_bits);
     p += strlen(p);
 
     if (ah_attr->is_global) {
+        snprintf(p, endp - p, " flow_label=0x%x", ah_attr->grh.flow_label);
```
yosefe, line 1303: add to line 1309

src/uct/ib/base/ib_device.c
```diff
@@ -1294,17 +1294,19 @@ const char* uct_ib_ah_attr_str(char *buf, size_t max,
...
+        p += strlen(p);
         snprintf(p, endp - p, " dgid=");
         p += strlen(p);
         uct_ib_gid_str(&ah_attr->grh.dgid, p, endp - p);
         p += strlen(p);
-        snprintf(p, endp - p, " sgid_index=%d traffic_class=%d",
+        snprintf(p, endp - p, " sgid_index=%d traffic_class=0x%x",
```
yosefe, line 1309: why need to print traffic class as hex?

src/uct/ib/base/ib_device.c
changchengx: I'll keep the original print format. The original idea is to keep use similar format as GID(hex) to print LID.

src/uct/ib/base/ib_device.c
```diff
@@ -1304,7 +1304,8 @@ const char* uct_ib_ah_attr_str(char *buf, size_t max,
         p += strlen(p);
         uct_ib_gid_str(&ah_attr->grh.dgid, p, endp - p);
         p += strlen(p);
-        snprintf(p, endp - p, " sgid_index=%d traffic_class=%d",
+        snprintf(p, endp - p, " flow_label = 0x%x sgid_index=%d "
```
yosefe, line 1307: remove space around =

yosefe (approved):

PR #9233, opened 2023-07-21
UCT/CUDA/GDR_COPY: Cleaned up configure and makefile.

rakhmets:
## What
Removed CUDA flags and libs from GDR_COPY Makefile.
Added GDR_COPY_LIBS to distinguish LD_FLAGS and LIBADD.

### DIFF
--- a/src/uct/cuda/gdr_copy/Makefile.am
+++ b/src/uct/cuda/gdr_copy/Makefile.am
@@ -6,13 +6,12 @@
 if HAVE_GDR_COPY
 
 module_LTLIBRARIES              = libuct_cuda_gdrcopy.la
-libuct_cuda_gdrcopy_la_CPPFLAGS = $(BASE_CPPFLAGS) $(CUDA_CPPFLAGS) $(GDR_COPY_CPPFLAGS)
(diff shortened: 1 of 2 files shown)


src/uct/cuda/gdr_copy/Makefile.am
```diff
@@ -6,13 +6,11 @@
 if HAVE_GDR_COPY
 
 module_LTLIBRARIES              = libuct_cuda_gdrcopy.la
-libuct_cuda_gdrcopy_la_CPPFLAGS = $(BASE_CPPFLAGS) $(CUDA_CPPFLAGS) $(GDR_COPY_CPPFLAGS)
-libuct_cuda_gdrcopy_la_CFLAGS   = $(BASE_CFLAGS)
```
yosefe, line 10: why removed libuct_cuda_gdrcopy_la_CFLAGS ?

rakhmets: Returned back, because there is no need to remove the flags.

yosefe (approved):

yosefe (approved):

PR #9242, opened 2023-07-26
AZP: Performance testing

Alexey-Rivkin:
## What
A pipeline for performance testing.
The pipeline should be started manually by adding the comment below under selected PRs:
`/azp run perf`

The test will fail if degradation of >5% will be detected (set by the `threshold` var).

## How
A shared FS is used as a working directory, to prevent compiling on each node.

### Sample build:
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=66519

### DIFF
--- a/buildlib/azure-pipelines-perf.yml
+++ b/buildlib/azure-pipelines-perf.yml
@@ -1,0 +1,144 @@
+trigger: none
+pr:
+  branches:
+    include:
+    - '*'
+  paths:
+    include:
+    - '*'
+
+variables:
+- name: WorkDir
+  value: /hpc/scrap/azure/$(Build.DefinitionName)/$(Build.BuildId)-$(Build.BuildNumber)
+- name: perf_nodes
+  value: vulcan03,vulcan04
+- name: threshold
+  value: 5
+
+resources:
+  repositories:
+  - repository: PerfX
+    type: github
+    name: Mellanox-lab/PerfX
+    endpoint: Al3xR01
+    ref: master
+
+stages:
+  - stage: Prepare
+    jobs:
+      - job: Prepare
+        pool:
+          name: MLNX
+          demands:
+          - ucx_perf
+
+        steps:
+          - checkout: self
+            clean: true
+            fetchDepth: 10
+
+          - checkout: PerfX
+            clean: true
+            fetchDepth: 1
+
+          - bash: |
+              set -xeE
+              mkdir -p $(WorkDir)
+              mv $(Build.SourcesDirectory)/* $(WorkDir)    
+            displayName: Prepare WorkDir with code
+
+
+  - stage: Performance
+    dependsOn: Prepare
+    jobs:
+      - job: Perf
+        displayName: Performance testing
+        pool:
+          name: MLNX
+          demands:
+          - ucx_perf
+
+        steps:
+          - checkout: none
+            clean: true
+
+          - bash: |
+              printf '\n%79s\n\n' | tr ' ' =
+              echo "Triggered by repo: $(Build.Repository.Name)"
+              echo "Triggering repo URL: $(Build.Repository.Uri)"
+              echo "The branch of the triggering repo the build was queued for: $(Build.SourceBranch)"
+              echo "PR's merge message: $(Build.SourceVersionMessage)"
+              
+              # Extract commit SHAs from the PR's merge message.
+              SHA_After=$(echo $(Build.SourceVersionMessage) | awk '{print $2}')
+              SHA_Before=$(echo $(Build.SourceVersionMessage) | awk '{print $4}')
+                            
+              printf '\n%79s\n\n' | tr ' ' =
+              echo "Base SHA from Git log:       $(cd $(WorkDir)/ucx && git rev-parse HEAD^)"
+              echo "Base SHA from merge message: $SHA_Before"
+
+              printf '\n%79s\n\n' | tr ' ' =
+              echo "Latest SHA from Git log:  $(cd $(WorkDir)/ucx && git rev-parse HEAD)"
+              echo "Latest SHA from Azure:    $(Build.SourceVersion)"
+              echo "Topic commit to merge:    $SHA_After"
+              echo -e "The merge commit will be used for testing. \n"
+            displayName: Debug info
+            workingDirectory: $(WorkDir)
+
+          - bash: |
+              set -eE
+              # Extract commit SHAs from the PR's merge message.
+              # SHA_After=$(Build.SourceVersion)
+              SHA_After=$(echo $(Build.SourceVersionMessage) | awk '{print $2}')
+              SHA_Before=$(echo $(Build.SourceVersionMessage) | awk '{print $4}')
+
+              python3 -m venv VENV
+              source VENV/bin/activate
+              cd $(WorkDir)/PerfX
+              pip3 install -q -r requirements.txt
+              module load /hpc/local/etc/modulefiles/hpcx-gcc
+
+              set -xeE -o pipefail
+              python3 -u perfx.py \
+                --config ucx-rdmz.yml \
+                --before "$SHA_Before" \
+                --after "$SHA_After" \
+                --module hpcx-gcc \
+                --filter \
+                --extra-ompi-opts " -H $(perf_nodes)" | tee $(WorkDir)/results.txt
+            displayName: Perf test
+            workingDirectory: $(WorkDir)
+
+          - script: $(WorkDir)/ucx/buildlib/tools/perf_results.py $(WorkDir)/results.txt $(threshold)
+            displayName: Results analyzer
+            workingDirectory: $(WorkDir)
+
+          - bash: |
+              set -x
+              cat perfx.log
+            displayName: Print raw log
+            workingDirectory: $(WorkDir)/PerfX
+
+          - task: PublishBuildArtifacts@1
+            inputs:
+              pathToPublish: '$(WorkDir)/PerfX/perfx.log'
+              artifactName: perfx.log
+            displayName: Publish raw log
+            condition: always()
+
+
+  - stage: Cleanup
+    dependsOn: Performance
+    condition: always()
+    jobs:
+      - job: Cleanup
+        displayName: Cleanup WorkDir
+        pool:
+          name: MLNX
+        steps:
+          - checkout: none
+            clean: true
+          - bash: |
+              set -x
+              rm -rf $(WorkDir)
+            displayName: Cleanup WorkDir
--- a/buildlib/tools/perf_results.py
+++ b/buildlib/tools/perf_results.py
@@ -1,0 +1,42 @@
+#!/usr/bin/python3
+
+"""
+Checks perf test results for regressions based on a specified threshold.
+Usage: ./script_name.py <filename> <threshold>
+"""
+
+import sys
+
+
+def print_usage():
+    print(f"Usage: {sys.argv[0]} <filename> <threshold>")
+    print("\t<filename>: The file containing performance test results.")
+    print("\t<threshold>: The percentage threshold to detect significant performance regression.")
+    exit(1)
+
+
+def check_perf_results(filename, threshold):
+    max_regression = 0
+    test_name_max = test_name = None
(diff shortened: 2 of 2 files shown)


Alexey-Rivkin:
Related PerfX PRs (merged):
https://github.com/Mellanox-lab/PerfX/pull/43 
https://github.com/Mellanox-lab/PerfX/pull/47   
https://github.com/Mellanox-lab/PerfX/pull/48

buildlib/azure-pipelines-perf.yml
```diff
@@ -0,0 +1,129 @@
...
+              SHA_After=$(echo $(Build.SourceVersionMessage) | awk '{print $2}')
+              SHA_Before=$(echo $(Build.SourceVersionMessage) | awk '{print $4}')
+
+              python3 -m venv VENV
+              source VENV/bin/activate
+              cd $(WorkDir)/PerfX
+              pip3 install -q -r requirements.txt
+              module load /hpc/local/etc/modulefiles/hpcx-gcc
+
+              set -o pipefail; set -x
```
MrBr-github, line 99: This can be single line
I'd also suggest `Ee` flags
```
set -xeE -o pipefail
```

buildlib/azure-pipelines-perf.yml
```diff
@@ -0,0 +1,129 @@
...
+                --extra-ompi-opts " -H $(perf_nodes)" | tee $(WorkDir)/results.txt
+            displayName: Perf test
+            workingDirectory: $(WorkDir)
+
+          - script: $(WorkDir)/ucx/buildlib/tools/perf_results.py $(WorkDir)/results.txt $(threshold)
+            displayName: Results analyzer
+            workingDirectory: $(WorkDir)
+
+
+  - stage: Cleanup
```
MrBr-github, line 115: What happens artefacts with jobs that never reach cleanup stage? `Always` not always executed

buildlib/azure-pipelines-perf.yml
```diff
@@ -0,0 +1,129 @@
...
+          - checkout: self
+            clean: true
+            fetchDepth: 10
+
+          - checkout: PerfX
+            clean: true
+            fetchDepth: 1
+
+          - bash: |
+              set -x
```
MrBr-github, line 43: Add `Ee`

MrBr-github (changes requested):

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: That's the purpose of `always`. Tested and it works.

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: Changing, thanks!
`-eE` are set a few lines above.

buildlib/azure-pipelines-perf.yml
MrBr-github: Not always
If pipeline fails as a pipeline you'll never reach this point and your artifacts will be there for ever

Alexey-Rivkin: How can we simulate such a situation?

MrBr-github:
Questions:
- Why manual trigger?
- Which code are you checking? the one that came from module or the one that comes from PR? 
- If from PR please show where you enforcing it

Alexey-Rivkin:
> 

The manual trigger was required by dev, so it would only be tested for selected PRs.
The test is per PR code, as compared to its merge base.
There is an info stage dedicated to this point exactly:
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=66532&view=logs&j=94be751c-9e4c-5c28-d13e-4c720246b05d&t=dceb8979-95f2-5ac1-086b-092e6f4a7bb5&l=29

buildlib/tools/perf_results.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/python3
+
+import sys
+
+def check_perf_results(filename, threshold):
```
amaslenn, line 5: Shouldn't that be a PerfX functionality? We can set a threshold in CLI or YAML config and depends on that it will return a non-zero code for a failure.

buildlib/azure-pipelines-perf.yml
```diff
@@ -0,0 +1,129 @@
...
+              source VENV/bin/activate
+              cd $(WorkDir)/PerfX
+              pip3 install -q -r requirements.txt
+              module load /hpc/local/etc/modulefiles/hpcx-gcc
+
+              set -xeE -o pipefail
+              python3 -u perfx.py \
+                --config ucx-rdmz.yml \
+                --before "$SHA_Before" \
+                --after "$SHA_After" \
```
amaslenn, line 103: Since PerfX already have before/after, do you think it make sense to introduce something like `--before master --after HEAD`? I assume it should even work our of the box.

buildlib/azure-pipelines-perf.yml
```diff
@@ -0,0 +1,129 @@
...
+resources:
+  repositories:
+  - repository: PerfX
+    type: github
+    name: Mellanox-lab/PerfX
+    endpoint: Al3xR01
+    ref: master
+
+stages:
+  - stage: Prepare
```
amaslenn, line 25: Why do you need so many stages here? They depend on one another, in this case you can even put it all into a single job with multiple step. Or I might be missing something here.

buildlib/tools/perf_results.py
Alexey-Rivkin: Absolutely! 
An external analyzer was a faster solution. If other projects may benefit from such functionality, let's make it a part of PerfX.

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: Discovering those flags was a lifesaver!
It's great as it is if you ask me.

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: Separating into stages improves visibility. It's easier to see which stage fails from the summary page, no other benefits.

buildlib/tools/perf_results.py
amaslenn: Should we do it right now?

buildlib/azure-pipelines-perf.yml
amaslenn: But preparation and cleanup are utility stages, real customers won't benefit from seeing them IMO.

buildlib/tools/perf_results.py
Alexey-Rivkin: Up to you. I need to align my priorities with the mgmt first.

buildlib/azure-pipelines-perf.yml
amaslenn: I just mean we don't need to define `SHA`s like that, we can use existing functionality already. YAML can become shorter :)

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: Agree, it was easier for me to debug.
Can dense it all into a single stage.
@yosefe WDYT?

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: You mean, just put master and HEAD into yaml? Can try that.
Please note, that I'm using the merge topic commit instead of HEAD. Not that it should be a difference.

yosefe:
@Alexey-Rivkin where can we see a successful perf testing result? seems some jobs failed and for others i can't see logs

Alexey-Rivkin:
> @Alexey-Rivkin where can we see a successful perf testing result? seems some jobs failed and for others i can't see logs

There is an outstanding issue with Azure bot being unstable. 

Sometimes [it works](https://github.com/openucx/ucx/pull/9238#issuecomment-1650215063):
`Azure Pipelines successfully started running 1 pipeline(s). `
Sometimes [it doesn't](https://github.com/openucx/ucx/pull/9238#issuecomment-1650538426):
`No commit pushedDate could be found for PR 9238 in repo openucx/ucx`

Errors might vary. Working on it right now.

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: The cleanup stage is set with the agent's pool and no demands. It means that it will pick literally any agent available and send it cleaning.

Alexey-Rivkin:
Regarding Azure bot malfunction, there is an open bug on Github:
https://github.com/microsoft/winget-pkgs/issues/111634#issuecomment-1636218300

buildlib/azure-pipelines-perf.yml
amaslenn: Yes, that is what I mean.

Why do you use merge topic and and just something latest stable? Aren't we supposed to measure PR effect?

buildlib/tools/perf_results.py
amaslenn: OK, let's see how it goes. Ideally we should not produce such scripts but put such functionality into the main tool. But it that is OK with UCX team, let's use it as a "temp solution" :)

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: Debatable. This merge commit represents the state of the code after the two branches were merged, so it should best reflect the "After" state. In fact, I see no diff between the merge commit and the last one.

Alexey-Rivkin: The merge base is not always 'master'. It might be v.*.*, so better to keep it dynamic instead of hardcoding in yaml.

amaslenn: OK, I see. For clarity I'd rely on some specific variables here instead of merge message. There should be something like `$targetBranch` and `$currentCommit`.

Alexey-Rivkin: Following an offline discussion with Yossi, agreed to rely on the merge commit, as I do now.
I found no predefined var in Azure for it, thus keeping the current implementation.

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: My personal preference is to keep stages separated for better visibility.

yosefe:
@Alexey-Rivkin can we print full run log in the perf pipeline result (or as an artifact)?

buildlib/azure-pipelines-perf.yml
```diff
@@ -0,0 +1,142 @@
...
+            displayName: Results analyzer
+            workingDirectory: $(WorkDir)
+
+          - script: cat perfx.log
+            displayName: Print debug log
+            workingDirectory: $(WorkDir)
+
+          - task: PublishBuildArtifacts@1
+            inputs:
+              pathToPublish: '$(Build.SourcesDirectory)'
```
amaslenn, line 122: Should it also include `perfx.log`? Otherwise the whole directory will be uploaded.

Alexey-Rivkin: Right, thanks!

Alexey-Rivkin:
> @Alexey-Rivkin can we print full run log in the perf pipeline result (or as an artifact)?

@yosefe Please, find the full raw log:
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=67249&view=logs&j=94be751c-9e4c-5c28-d13e-4c720246b05d&t=91332361-356c-5009-c13f-158bec505cbf&l=3871

Also uploaded as an artifact:
https://dev.azure.com/ucfconsort/_apis/resources/Containers/15545080/perfx.log?itemPath=perfx.log%2Fperfx.log

Alexey-Rivkin:
@yosefe 
Added CLI prints:
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=67334&view=logs&j=94be751c-9e4c-5c28-d13e-4c720246b05d&t=91332361-356c-5009-c13f-158bec505cbf&l=3897

yosefe (approved):

PR #9246, opened 2023-07-26
UCT/IB/RC: only use DEVx to connect RCQP created by DEVx

changchengx:
## What
Only use DEVx to connect RC/QP created by DEVx 

## Why ?
If ```rcqp``` isn't enumerated in ```UCX_IB_MLX5_DEVX_OBJECTS``` environment variable, it uses dv API to create the ```rcqp```. Under this case, it should use ```dv&verbs``` API to operate the ```rcqp``` instead of using DEVx.

## How ?
check ```md->flags``` against ```UCT_IB_MLX5_MD_FLAG_DEVX_RC_QP``` instead of ```UCT_IB_MLX5_MD_FLAG_DEVX```

### DIFF
--- a/src/uct/ib/rc/accel/rc_mlx5_ep.c
+++ b/src/uct/ib/rc/accel/rc_mlx5_ep.c
@@ -743,7 +743,7 @@ uct_rc_mlx5_ep_connect_qp(uct_rc_mlx5_iface_common_t *iface,
     uct_ib_mlx5_md_t *md = ucs_derived_of(iface->super.super.super.md, uct_ib_mlx5_md_t);
 
     ucs_assert(path_mtu != UCT_IB_ADDRESS_INVALID_PATH_MTU);
-    if (md->flags & UCT_IB_MLX5_MD_FLAG_DEVX) {
+    if (md->flags & UCT_IB_MLX5_MD_FLAG_DEVX_RC_QP) {
         return uct_rc_mlx5_iface_common_devx_connect_qp(
                 iface, qp, qp_num, ah_attr, path_mtu, path_index,
                 iface->super.config.max_rd_atomic);


changchengx:
It reports below two failed test cases in **UCX PR(Tests roce_proto_disable on worker3)**. I'll check it at weekend.
```
2023-07-27T00:13:14.2204077Z [  FAILED  ] 2 tests, listed below:
2023-07-27T00:13:14.2204426Z [  FAILED  ] ib/multi_rail_max.max_lanes/2, where GetParam() = ib/rndv_put_zcopy
2023-07-27T00:13:14.2204828Z [  FAILED  ] ib/multi_rail_max.max_lanes/6, where GetParam() = ib/send_nbr
```

changchengx:
It needs to trigger re-run the above failed test cases after analysis.

yosefe (approved):

yosefe (changes requested):
Actually, why is this needed?
We can create QP by verbs and connect with DevX

changchengx:
> Actually, why is this needed? We can create QP by verbs and connect with DevX
>
1. I agree with that ```We can create QP by verbs and connect with DevX```.
2. When some object types are not listed in ```UCX_IB_MLX5_DEVX_OBJECTS```, it means that it won't use DEVX to create the object. It maybe better to use the corresponding API to modify the created object to check how it's implemented with using verbs or direct verbs.

yosefe:
> > Actually, why is this needed? We can create QP by verbs and connect with DevX
> 
> 1. I agree with that `We can create QP by verbs and connect with DevX`.
> 2. When some object types are not listed in `UCX_IB_MLX5_DEVX_OBJECTS`, it means that it won't use DEVX to create the object. It maybe better to use the corresponding API to modify the created object to check how it's implemented with using verbs or direct verbs.

It's intentional (and supported by fw/driver) to allow creating qp by verbs and modifying using devx.

changchengx:
> It's intentional (and supported by fw/driver) to allow creating qp by verbs and modifying using devx.
>
I agree with this. Let's suppose one scenario:
1. Build UCX with DEVx
2. The running environment also supports DEVx
3. ```rcqp``` isn't listed in ```UCX_IB_MLX5_DEVX_OBJECTS```. Then, dv API is used to create rcqp. In current code, it use DEVx to modify QP into RTS.
4. However, if the user wants to use rdma-core dv API to check some qp info, without this PR, the user has to look for the method to do it with DEVx. With this PR, the user can use standard rdma-core dv API to check it directly.
5. One example: if the user want to check rc qp affinity, ```mlx5dv_query_qp_lag_port``` is enough to do it without the requirement to know how to use DEVx to do it. Use ```mlxdv_query_qp_lag_port``` in ```uct_rc_iface_qp_connect``` is easy to be understood.

yosefe:
@changchengx item (4) is not clear - user code do not have access to UCX QP.
And when DevX is enabled, UCX should use DevX to query QP properties (including QP affinity) and not DV API.
This way we can also support rdma-core versions that do not have mlx5dv_query_qp_lag_port API.

changchengx:
> when DevX is enabled, UCX should use DevX to query QP properties (including QP affinity) and not DV API.
>
I agree with you now. Let me close this PR.

PR #9262, opened 2023-08-01
UCM/BISTRO: Use appropriate atomic swap for size of update

rbradford:
On RISC-V for an atomic update the destination must be naturally aligned with the size of the type. Since the destinations in this case are addresses of library functions to patch we cannot control the addresses. Fortunately the instructions must be aligned to a 16-bit alignment which also matches the potential size of the patch on RISC-V (compressed J instruction.)

For other architectures the size of patch is either 4 bytes (aarch64) or 2 bytes (x86-64) which are both handled by this patch.

## What
Use atomic updates that match size of data to be updated

## Why ?
For RISC-V (and potentially for other architectures) the target address for atomic instructions should be naturally aligned with the size of the value.

## How ?
Check the size of the value and use the atomic update function that matches the size.

### DIFF
--- a/src/ucm/bistro/bistro.c
+++ b/src/ucm/bistro/bistro.c
@@ -50,12 +50,18 @@ static ucs_status_t ucm_bistro_protect(void *addr, size_t len, int prot)
 
 void ucm_bistro_modify_code(void *dst, const ucm_bistro_lock_t *bytes)
 {
-    uint32_t value;
+    uint16_t value16;
+    uint32_t value32;
 
-    UCS_STATIC_ASSERT(sizeof(*bytes) <= sizeof(value));
-    memcpy(&value, dst, sizeof(value));
-    memcpy(&value, bytes, sizeof(*bytes));
-    (void)ucs_atomic_swap32(dst, value);
+    UCS_STATIC_ASSERT((sizeof(*bytes) == 2) || (sizeof(*bytes) == 4));
+
+    if (sizeof(*bytes) == 2) {
+        memcpy(&value16, bytes, sizeof(*bytes));
+        (void)ucs_atomic_swap16(dst, value16);
+    } else {
+        memcpy(&value32, bytes, sizeof(*bytes));
+        (void)ucs_atomic_swap32(dst, value32);
+    }
 }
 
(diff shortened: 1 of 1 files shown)


shamisp:
@rbradford , Have you had a chance to sign CLA: https://openucx.org/license/
thanks

rbradford:
> @rbradford , Have you had a chance to sign CLA: https://openucx.org/license/ thanks

My employer, Rivos Inc, is covered by a corporate CLA.

src/ucm/bistro/bistro.c
```diff
@@ -50,12 +50,16 @@ static ucs_status_t ucm_bistro_protect(void *addr, size_t len, int prot)
 
 void ucm_bistro_modify_code(void *dst, const ucm_bistro_lock_t *bytes)
 {
-    uint32_t value;
-
-    UCS_STATIC_ASSERT(sizeof(*bytes) <= sizeof(value));
-    memcpy(&value, dst, sizeof(value));
-    memcpy(&value, bytes, sizeof(*bytes));
-    (void)ucs_atomic_swap32(dst, value);
+    UCS_STATIC_ASSERT(sizeof(*bytes) == 2 || sizeof(*bytes) == 4);
```
tvegas1, line 53: although not strictly needed, as per coding style, please add parentheses like `ASSERT((sizeof...) || (sizeof...))`, and blank line after.

tvegas1 (approved):
variables declarations need to be on top of function, as per coding guidelines.

src/ucm/bistro/bistro.c
rbradford: Updated, thank you!

rbradford:
> variables declarations need to be on top of function, as per coding guidelines.

So, like this:

```

```

> variables declarations need to be on top of function, as per coding guidelines.

Updated, thank you!

tvegas1:
Failure in cuda_hooks, test_cuMemAllocHost/test_cudaMallocManaged under valgrind should be unrelated.

yosefe (approved):

brminich (approved):

PR #9263, opened 2023-08-01
UCP/PROTOV2: Fix AM multi-eager protocols

brminich:
## What
Fix UCP AM multi-eager protocols for protoV2

## Why
When reporting maximal allowed AM header size to the user, we consider the minimal UCT fragment size among all enabled transports. However, we may decide to send less that transport's max fragment size, for example in case its bandwidth is lower than other lanes.
(to fill which assert fails)

## How ?
Need to take into account max AM user header size when calculating how much data UCP can send on a single lane. (because user header and all UCP headers must fit into the first fragment)


src/ucp/proto/proto_multi.c
```diff
@@ -150,12 +150,15 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
...
+        if (mpriv->num_lanes == 1) {
+            max_frag      = ucs_max(max_frag, params->super.hdr_size);
+            perf.max_frag = max_frag;
+        }
 
         /* Make sure fragment is not zero */
         ucs_assert(max_frag > 0);
 
         lpriv->max_frag = max_frag;
-        perf.max_frag   = ucs_min(perf.max_frag, lpriv->max_frag);
```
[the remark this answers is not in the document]
brminich: yes, i  suppose it will be ok, because we want to provide perf info about the first fragment to `ucp_proto_common_init_caps`.

rakhmets (dismissed):

src/ucp/am/eager_multi.c
```diff
@@ -15,10 +15,16 @@
 #include <ucp/proto/proto_multi.inl>
 
 
+static size_t ucp_am_eager_max_first_hdr_size(ucp_worker_h worker)
+{
+    return worker->max_am_header + UCP_AM_FIRST_FRAG_META_LEN + 1;
```
yosefe, line 20: why +1?

src/ucp/am/eager_multi.c
```diff
@@ -15,10 +15,16 @@
 #include <ucp/proto/proto_multi.inl>
 
 
+static size_t ucp_am_eager_max_first_hdr_size(ucp_worker_h worker)
```
yosefe, line 18: ucp_am_eager_multi_max_first_hdr_size

src/ucp/proto/proto_multi.c
```diff
@@ -150,12 +150,15 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
            Due to floating-point accuracy, max_frag may be too large. */
         max_frag = ucs_double_to_sizet(lane_perf->bandwidth / max_frag_ratio,
                                        lane_perf->max_frag);
+        if (mpriv->num_lanes == 1) {
```
yosefe, line 153: can you pls add comment?

brminich:
this PR affects performance estimations - now UCP AM multi-eager protocol always assumes that maximal user header is being sent (which is not always the true). The fix needs to be reconsidered

brminich:
> Should we revert [0e92c1d](https://github.com/openucx/ucx/commit/0e92c1d8fa749daa4c1b9ee4be6c807f4688eb75) in this PR to confirm the fix?

it's already done in the 1st commit

rakhmets (dismissed):

src/ucp/proto/proto_multi.c
yosefe: this would cause sending max_frag on first lane every time it's used, and not only for the first fragment in the message, and it's not a desired behavior

src/ucp/proto/proto_multi.h
```diff
@@ -87,6 +87,11 @@ typedef struct {
      * by this protocol */
     ptrdiff_t                      opt_align_offs;
 
+    /* Minimal size of the first fragment (e.g. with AM API user may send a
+     * custom header in the first fragment which may consume the whole
+     * fragment) */
+    size_t                         first_min_size;
```
yosefe, line 93: let's try to avoid protocol-specific field in a common structure

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1526,9 +1535,8 @@ class test_ucp_am_nbx_dts : public test_ucp_am_nbx_reply {
 
 /* Skip tests for ud_v and ud_x because of unstable reproducible failures during
  * roce on worker CI jobs. The test fails with invalid am_bcopy length. */
-UCS_TEST_SKIP_COND_P(test_ucp_am_nbx_dts, short_bcopy_send,
-                     is_proto_enabled() && has_any_transport({"ud_v", "ud_x"}),
-                     "ZCOPY_THRESH=-1", "RNDV_THRESH=-1")
+UCS_TEST_P(test_ucp_am_nbx_dts, short_bcopy_send, "ZCOPY_THRESH=-1",
```
yosefe, line 1538: fix comment above

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1485,7 +1485,16 @@ UCS_TEST_SKIP_COND_P(test_ucp_am_nbx_seg_size, multi, has_transport("self"))
...
+/* Different segments sizes are not supported with UCP AM, because:
+ * - max_am_header_size is fetched from worker attributes and is based
+ *   on local iface capabilities (smallest is taken)
+ * - when connecting peer with smaller segment size, ep adjusts its caps
+ *   (e.g. am.max_bcopy), but the user may already queired larger size
+ *   for max_am_hdr from the worker
+ *
+ * TODO: Enable these tests when the issue above is fixed
+ * UCP_INSTANTIATE_TEST_CASE(test_ucp_am_nbx_seg_size)
+ */
```
yosefe, line 1497: we may want to introduce querying max am header size per ucp endpoint

brminich:
fixed by #10263

Issue #9260, opened 2023-08-01
Build with cuda fails on master


bathmatt:
### Describe the bug
Starting at the commit listed below one cannot compile ucx with cuda.  I have uploaded a docker file and command lines.
```
make[4]: Entering directory '/var/tmp/ucx/src/uct/cuda/gdr_copy'
  CC       libuct_cuda_gdrcopy_la-gdr_copy_md.lo
  CC       libuct_cuda_gdrcopy_la-gdr_copy_iface.lo
In file included from gdr_copy_md.c:22:
/var/tmp/ucx/src/uct/cuda/base/cuda_iface.h:13:10: fatal error: cuda_runtime.h: No such file or directory
   13 | #include <cuda_runtime.h>
      |          ^~~~~~~~~~~~~~~~
```

### Steps to Reproduce

On this version it started to happen

```
commit 0697b7cdd484c035dd4b1f7d99c034c05310434d (HEAD)
Author: Raul Akhmetshin <rakhmetshin@nvidia.com>
Date:   Mon Jul 24 18:59:34 2023 +0300

    UCT/CUDA/GDR_COPY: Cleaned up configure and makefile.
```

- Command line

./autogen.sh ; ./configure --prefix=/usr/local/ucx --disable-assertions --disable-backtrace-detail --disable-debug --disable-doxygen-doc --disable-logging --disable-params-check --disable-static --enable-mt --with-cuda=/usr/local/cuda --with-gdrcopy=/usr/local/gdrcopy --with-knem=/usr/local/knem ; make -j 45

### Setup and versions
Linux with nvidia container.

[Dockerfile.txt](https://github.com/openucx/ucx/files/12226786/Dockerfile.txt)

### Additional information (depending on the issue)
- OpenMPI version
- Output of `ucx_info -d` to show transports and devices recognized by UCX
- Configure result - config.log
- Log file - configure UCX with "--enable-logging" - and run with "UCX_LOG_LEVEL=data"

rakhmets:
https://github.com/openucx/ucx/pull/9264 was merged.
Master branch should be fixed.

PR #9264, opened 2023-08-02
UCT/CUDA: Moved declaration of commonly used methods to the separate header file.

rakhmets:
## What
Created cuda_base.h to include the file in gdr_copy subcomponent to get rid of including of CUDA Runtime header.
Moved uct_cuda_base_query_devices to cuda_base.h.
Added uct_cuda_base_check_device_name to cuda_base.h, and moved macro UCT_CUDA_DEV_NAME from cuda_iface.h to cuda_iface.c to incapsulate the usage of the string constant.
Replaced `CUdeviceptr` by `unsigned long` to pass the type used by gdr copy API.
Replaced cuda_iface.h by cuda_md.h in gdr_copy_md.c because only a method from cuda_md.h is used.

## Why ?
The PR fixes build issues https://github.com/openucx/ucx/issues/9260 introduced by https://github.com/openucx/ucx/pull/9233, completing removing dependency on CUDA headers and libraries from gdr_copy subcomponent .

### DIFF
--- a/buildlib/tools/builds.sh
+++ b/buildlib/tools/builds.sh
@@ -211,13 +211,20 @@ build_cuda() {
 		if az_module_load $GDRCOPY_MODULE
 		then
 			echo "==== Build with enable cuda, gdr_copy ===="
-			${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst --with-cuda --with-gdrcopy
-			$MAKEP
-			make_clean distclean
-
 			${WORKSPACE}/contrib/configure-release --prefix=$ucx_inst --with-cuda --with-gdrcopy
 			$MAKEP
 			make_clean distclean
+
+			echo "==== Build with enable cuda, gdr_copy by ./configure parameter ===="
+			# Use path to CUDA instead of loading CUDA as module, to check
+			# GDRCopy subcomponent does not include CUDA headers
+			cuda_path=$(dirname $(dirname $(which nvcc)))
+			az_module_unload $CUDA_MODULE
+			${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst --with-cuda=$cuda_path --with-gdrcopy
+			$MAKEP
+			make_clean distclean
+			az_module_load $CUDA_MODULE
+
 			az_module_unload $GDRCOPY_MODULE
 		fi
 
--- a/src/tools/perf/lib/libperf.c
+++ b/src/tools/perf/lib/libperf.c
@@ -1486,7 +1486,8 @@ static ucs_status_t uct_perf_setup(ucx_perf_context_t *perf)
     uct_iface_params_t iface_params = {
         .field_mask           = UCT_IFACE_PARAM_FIELD_OPEN_MODE   |
                                 UCT_IFACE_PARAM_FIELD_STATS_ROOT  |
-                                UCT_IFACE_PARAM_FIELD_RX_HEADROOM,
+                                UCT_IFACE_PARAM_FIELD_RX_HEADROOM |
+                                UCT_IFACE_PARAM_FIELD_DEVICE,
         .open_mode            = UCT_IFACE_OPEN_MODE_DEVICE,
         .mode.device.tl_name  = params->uct.tl_name,
         .mode.device.dev_name = params->uct.dev_name,
--- a/src/uct/cuda/base/cuda_iface.c
+++ b/src/uct/cuda/base/cuda_iface.c
@@ -9,8 +9,12 @@
 #endif
 
 #include "cuda_iface.h"
+#include "cuda_md.h"
 
 #include <ucs/sys/string.h>
+
+
+#define UCT_CUDA_DEV_NAME "cuda"
 
 
 const char *uct_cuda_base_cu_get_error_string(CUresult result)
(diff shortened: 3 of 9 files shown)


src/uct/cuda/base/cuda_iface.c
```diff
@@ -81,6 +85,17 @@ ucs_status_t uct_cuda_base_iface_event_fd_get(uct_iface_h tl_iface, int *fd_p)
     return UCS_OK;
 }
 
+ucs_status_t uct_cuda_base_check_device_name(const uct_iface_params_t *params)
+{
+    if (strncmp(params->mode.device.dev_name, UCT_CUDA_DEV_NAME,
+                strlen(UCT_CUDA_DEV_NAME)) != 0) {
+        ucs_error("No device was found: %s", params->mode.device.dev_name);
```
tvegas1, line 92: lower case in message

src/uct/cuda/base/cuda_base.h
```diff
@@ -0,0 +1,39 @@
...
+ *         array of device resources.
+ */
+ucs_status_t
+uct_cuda_base_query_devices(uct_md_h md,
+                            uct_tl_device_resource_t **tl_devices_p,
+                            unsigned *num_tl_devices_p);
+
+
+/**
+ * Check the divece name of iface parameter.
```
tvegas1, line 29: typo `device`, maybe also `iface` -> `interface` for all

tvegas1 (commented):
small details

src/uct/cuda/base/cuda_iface.c
rakhmets: Fixed.

src/uct/cuda/base/cuda_base.h
rakhmets: Fixed.

tvegas1 (dismissed):

src/uct/cuda/base/cuda_iface.c
```diff
@@ -81,6 +85,17 @@ ucs_status_t uct_cuda_base_iface_event_fd_get(uct_iface_h tl_iface, int *fd_p)
     return UCS_OK;
 }
 
+ucs_status_t uct_cuda_base_check_device_name(const uct_iface_params_t *params)
+{
+    if (strncmp(params->mode.device.dev_name, UCT_CUDA_DEV_NAME,
```
brminich, line 90: i think we also need to check `params->field_mask & UCT_IFACE_PARAM_FIELD_DEVICE`

src/uct/cuda/base/cuda_base.h
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param [in]  md               Memory domain to run the query on.
+ * @param [out] tl_devices_p     List of available devices on the md.
+ * @param [out] num_tl_devices_p Number of available devices on the md.
+ *
+ * @return UCS_OK if successful, or UCS_ERR_NO_MEMORY if failed to allocate the
+ *         array of device resources.
+ */
+ucs_status_t
+uct_cuda_base_query_devices(uct_md_h md,
```
yosefe, line 23: maybe move these funcs to cuda_md.h, instead of creating cuda_base.h?
it seems gdrcopy is already including cuda_md.h

yosefe (commented):
can we add some test to catch such issue (that someone includes cuda H files from gdrcopy by mistake)?

yosefe (commented):
LGTM;
WDYT about adding build test?

rakhmets:
> LGTM; WDYT about adding build test?

Path to CUDA should be passed using --with-cuda configure option, and the path shouldn't be in environment variables.
I will add this configure & build scenario.

buildlib/tools/builds.sh
```diff
@@ -211,13 +211,19 @@ build_cuda() {
...
 
-			${WORKSPACE}/contrib/configure-release --prefix=$ucx_inst --with-cuda --with-gdrcopy
+			# Use path to CUDA instead of loading CUDA as module, to check
+			# GDRCopy subcomponent does not include CUDA headers
+			cuda_path=$(dirname $(dirname $(which nvcc)))
+			az_module_unload $CUDA_MODULE
+			${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst --with-cuda=$cuda_path --with-gdrcopy
 			$MAKEP
 			make_clean distclean
+			az_module_load $CUDA_MODULE
```
yosefe, line 225: IMO would be easier to move the new command lines after line 235 below, when cuda is already unloaded

yosefe: ok

src/tools/perf/lib/libperf.c
```diff
@@ -1486,7 +1486,8 @@ static ucs_status_t uct_perf_setup(ucx_perf_context_t *perf)
     uct_iface_params_t iface_params = {
         .field_mask           = UCT_IFACE_PARAM_FIELD_OPEN_MODE   |
                                 UCT_IFACE_PARAM_FIELD_STATS_ROOT  |
-                                UCT_IFACE_PARAM_FIELD_RX_HEADROOM,
+                                UCT_IFACE_PARAM_FIELD_RX_HEADROOM |
+                                UCT_IFACE_PARAM_FIELD_DEVICE,
```
yosefe, line 1490: good fix, was the issue found with this PR?

rakhmets: Yes, test `cuda_copy/test_uct_loopback_cuda.envelope/0` failed  
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=67174&view=logs&j=f6a00158-c162-5d05-e321-5b7f75412dc2&t=3041e05d-2adc-5ccc-5b20-437cf039c2a1&l=21430
, after I addressed the comment https://github.com/openucx/ucx/pull/9264#discussion_r1283333535.

yosefe (approved):

Issue #9269, opened 2023-08-05
Compilation error in bistro code re. memcpy()


tonycurtis:
### Describe the bug

As title.

### Setup and versions
- OS version (e.g Linux distro) + CPU architecture (x86_64/aarch64/ppc64le/...)
   - `cat /etc/issue` or `cat /etc/redhat-release` + `uname -a`

```
$ cat /etc/redhat-release
Rocky Linux release 8.4 (Green Obsidian)
$ uname -a
Linux login1 4.18.0-305.25.1.el8_4.aarch64 #1 SMP Tue Nov 2 21:37:36 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux
```
### Git version

master @ 5221a5c2cf4265b6a83da3cd540f5d4898fa6f70

### Build environment

gcc 8.5.0 from rocky 8.4.  Compiles with e.g. gcc 12, 13.

```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../../ucx-git/src/ucm -I../.. -DCPU_FLAGS= -I/lustre/projects/ucx-group/arcurtis/src/ucx/build/../ucx-git/src -I/lustre/projects/ucx-group/arcurtis/src/ucx/build -I/lustre/projects/ucx-group/arcurtis/src/ucx/build/src -DUCM_MALLOC_PREFIX=ucm_dl -fno-strict-aliasing -DUSE_LOCKS=1 -DMALLINFO_FIELD_TYPE=int -I/usr/include -O3 -g -Wall -Werror -mcpu=thunderx2t99 -march=armv8.1-a+lse -fno-omit-frame-pointer -funwind-tables -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-label -Wno-long-long -Wno-endif-labels -Wno-sign-compare -Wno-multichar -Wno-deprecated-declarations -Winvalid-pch -Wno-pointer-sign -Werror-implicit-function-declaration -Wno-format-zero-length -Wnested-externs -Wshadow -Werror=declaration-after-statement -Wno-deprecated-declarations -MT bistro/libucm_la-bistro.lo -MD -MP -MF bistro/.deps/libucm_la-bistro.Tpo -c ../../../ucx-git/src/ucm/bistro/bistro.c  -fPIC -DPIC -DUCX_SHARED_LIB -o bistro/.libs/libucm_la-bistro.o
[...]

PR #9271, opened 2023-08-06
UCM/BISTRO: Use constant size in memcpy for ucm_bistro_modify_code

rbradford:
By using the static constant value we can be sure that the compiler will not generate an out of bounds error even for code that is unreachable.

e.g If sizeof(ucm_bistro_lock_t) == 4 the compiler generates an out of bounds error for the copy of bytes into the uint16_t (as 4 > 2) even though that code is unreachable.

Fixes: #9269

## What
Using constant values for memcpy to avoid compiler error for unreachable code.

## Why ?
Fix for #9269

## How ?
Using a constant value will prevent the out-of-bounds since the number of bytes is the same as the size of the type.

### DIFF
--- a/src/ucm/bistro/bistro.c
+++ b/src/ucm/bistro/bistro.c
@@ -53,13 +53,14 @@ void ucm_bistro_modify_code(void *dst, const ucm_bistro_lock_t *bytes)
     uint16_t value16;
     uint32_t value32;
 
-    UCS_STATIC_ASSERT((sizeof(*bytes) == 2) || (sizeof(*bytes) == 4));
+    UCS_STATIC_ASSERT((sizeof(*bytes) == sizeof(value16)) ||
+                      (sizeof(*bytes) == sizeof(value32)));
 
-    if (sizeof(*bytes) == 2) {
-        memcpy(&value16, bytes, sizeof(*bytes));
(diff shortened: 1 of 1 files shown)


src/ucm/bistro/bistro.c
```diff
@@ -56,10 +56,10 @@ void ucm_bistro_modify_code(void *dst, const ucm_bistro_lock_t *bytes)
     UCS_STATIC_ASSERT((sizeof(*bytes) == 2) || (sizeof(*bytes) == 4));
 
     if (sizeof(*bytes) == 2) {
-        memcpy(&value16, bytes, sizeof(*bytes));
+        memcpy(&value16, bytes, 2);
```
yosefe, line 59:
```c
UCS_STATIC_ASSERT((sizeof(*bytes) == sizeof(value16)) ||
                  (sizeof(*bytes) == sizeof(value32)));

if (sizeof(*bytes) == sizeof(value16)) {
    memcpy(&value16, bytes, sizeof(value16));
} else {
    memcpy(&value32, bytes, sizeof(value32));
}
```

rbradford: That's a nice way to do it - updated!

rbradford:
@yosefe Java111 tests is still timing out - looks like it should only really take < 10 minutes :-/

PR #9272, opened 2023-08-07
UCP/WORKER: Destroy mem_type EPs in case of worker creation error

ivankochin:
## What
Call `ucp_worker_mem_type_eps_destroy` instead of `ucp_worker_mem_type_eps_create` during `ucp_worker_cleanup` cleanup.


brminich (approved):

yosefe (approved):
good catch :)

PR #9282, opened 2023-08-09
UCT/IB: Avoid using ibv_mr::addr field

ivankochin:
## What
Stores memory region address explicitly and separately of `ibv_mr` to avoid using `ibv_mr::addr` field.

## Why
For some reason `ibv_reg_dmabuf_mr` stores offset instead of address to the output structure (https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/verbs.c#L411). So we need to avoid using it, since it leads to incorrect KSM creation.

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -34,6 +34,8 @@ uct_ib_mlx5_reg_key(uct_ib_md_t *md, void *address, size_t length,
     uct_ib_mlx5_devx_mem_t *memh = ucs_derived_of(ib_memh,
                                                   uct_ib_mlx5_devx_mem_t);
 
+    memh->address = address;
+
     return uct_ib_reg_key_impl(md, address, length, access_flags, dmabuf_fd,
                                dmabuf_offset, ib_memh,
                                &memh->mrs[mr_type].super, mr_type, silent);
@@ -129,11 +131,12 @@ uct_ib_mlx5_devx_reg_ksm(uct_ib_mlx5_md_t *md, int atomic, uintptr_t address,
 }
 
 static ucs_status_t
-uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md, int atomic,
+uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md, int atomic, void *address,
                               uct_ib_mlx5_devx_ksm_data_t *ksm_data,
                               size_t length, off_t off, const char *reason,
                               struct mlx5dv_devx_obj **mr_p, uint32_t *mkey)
 {
+    void *mr_address = address;
     ucs_status_t status;
     char *in;
     void *klm;
@@ -146,15 +149,15 @@ uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md, int atomic,
 
     klm = UCT_IB_MLX5DV_ADDR_OF(create_mkey_in, in, klm_pas_mtt);
     for (i = 0; i < ksm_data->mr_num; i++) {
-        UCT_IB_MLX5DV_SET64(klm, klm, address,
-                            (uintptr_t)ksm_data->mrs[i]->addr);
+        UCT_IB_MLX5DV_SET64(klm, klm, address, (uintptr_t)mr_address);
         UCT_IB_MLX5DV_SET(klm, klm, byte_count, ksm_data->mrs[i]->length);
         UCT_IB_MLX5DV_SET(klm, klm, mkey, ksm_data->mrs[i]->lkey);
-        klm = UCS_PTR_BYTE_OFFSET(klm, UCT_IB_MLX5DV_ST_SZ_BYTES(klm));
+        klm        = UCS_PTR_BYTE_OFFSET(klm, UCT_IB_MLX5DV_ST_SZ_BYTES(klm));
+        mr_address = UCS_PTR_BYTE_OFFSET(mr_address, ksm_data->mrs[i]->length);
     }
 
     status = uct_ib_mlx5_devx_reg_ksm(md, atomic,
-                                      (uintptr_t)ksm_data->mrs[0]->addr + off,
+                                      (uintptr_t)address + off,
                                       length, ksm_data->mr_num,
                                       ksm_data->mrs[0]->length, in, reason,
                                       mr_p, mkey);
@@ -195,20 +198,22 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_addr(
 static ucs_status_t
 uct_ib_mlx5_devx_reg_ksm_data_contig(uct_ib_mlx5_md_t *md,
                                      uct_ib_mlx5_devx_mr_t *mr, off_t offset,
-                                     int atomic, const char *reason,
+                                     void *address, int atomic,
+                                     const char *reason,
                                      struct mlx5dv_devx_obj **mr_p,
                                      uint32_t *mkey)
 {
-    uintptr_t mr_address = (uintptr_t)mr->super.ib->addr;
+    size_t mr_length = mr->super.ib->length;
     uintptr_t ksm_address;
     size_t ksm_length;
     int list_size;
 
     /* FW requires indirect atomic MR address and length to be aligned
      * to max supported atomic argument size */
-    ksm_address = ucs_align_down_pow2(mr_address, UCT_IB_MD_MAX_MR_SIZE);
-    ksm_length  = ucs_align_up(mr->super.ib->length + mr_address - ksm_address,
-                               md->super.dev.atomic_align);
+    ksm_address = ucs_align_down_pow2((uintptr_t)address,
+                                      UCT_IB_MD_MAX_MR_SIZE);
+    ksm_length  = mr_length + (uintptr_t)address - ksm_address;
+    ksm_length  = ucs_align_up(ksm_length, md->super.dev.atomic_align);
 
     /* Add offset to workaround CREATE_MKEY range check issue */
     list_size = ucs_div_round_up(ksm_length + offset, UCT_IB_MD_MAX_MR_SIZE);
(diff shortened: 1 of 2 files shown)


src/uct/ib/base/ib_md.c
```diff
@@ -303,149 +292,6 @@ uct_ib_md_print_mem_reg_err_msg(const char *title, void *address, size_t length,
...
-    mr_idx = 0;
-    cpu_id = 0;
-    for (thread_idx = 0; thread_idx < thread_num; thread_idx++) {
-        /* calculate number of mrs for each thread so each one will
-         * get proportional amount */
-        thread_num_mrs    = ucs_div_round_up(mr_num - mr_idx,
-                                             thread_num - thread_idx);
-        ctx               = &ctxs[thread_idx];
-        ctx->md           = md;
-        ctx->address      = UCS_PTR_BYTE_OFFSET(address, mr_idx * chunk_size);
```
yosefe, line 397: can't we calculate the address of each mr[I] like this instead of saving all of them in the ksm_data array?

ivankochin: Good idea, thanks! Done.

BTW, I moved several functions (`uct_ib_md_handle_mr_list_mt` and `uct_ib_md_mem_handle_thread_func`) from `ib_md.c` to `ib_mlx5dv_md.c` since they were used only if DEVX is enabled. WDYT, should I move them back?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -19,13 +20,24 @@
 #define UCT_IB_MLX5_MD_UMEM_ACCESS \
     (IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_READ | IBV_ACCESS_REMOTE_WRITE)
```
ivankochin, line 22:
```suggestion


```

yosefe (commented):
do we still need to move multi-thread registration functions to ib_mlx5dv_md.c?

ivankochin:
> do we still need to move multi-thread registration functions to ib_mlx5dv_md.c?

Seems like no, but they are used only in case of DEVX, so shouldn't we leave them there?

yosefe:
> > do we still need to move multi-thread registration functions to ib_mlx5dv_md.c?
> 
> Seems like no, but they are used only in case of DEVX, so shouldn't we leave them there?

We can discuss such refactoring in the context of another PR; in this PR, let's try to change only what's needed to fix the original issue (mr->addr being NULL)

src/uct/ib/base/ib_md.c
```diff
@@ -544,8 +544,7 @@ ucs_status_t uct_ib_dereg_mr(struct ibv_mr *mr)
         return UCS_OK;
     }
 
-    ucs_trace("ibv_dereg_mr(mr=%p addr=%p length=%zu)", mr, mr->addr,
-              mr->length);
+    ucs_trace("ibv_dereg_mr(mr=%p length=%zu)", mr, mr->length);
```
yosefe, line 547: why removed addr? even if NULL it's ok to print.

src/uct/ib/base/ib_md.h
```diff
@@ -611,18 +611,17 @@ ucs_status_t uct_ib_dereg_mr(struct ibv_mr *mr);
...
-                            uint64_t access_flags, struct ibv_mr **mrs);
-
 ucs_status_t uct_ib_reg_key_impl(uct_ib_md_t *md, void *address, size_t length,
                                  uint64_t access_flags, int dmabuf_fd,
                                  size_t dmabuf_offset, uct_ib_mem_t *memh,
                                  uct_ib_mr_t *mr, uct_ib_mr_type_t mr_type,
                                  int silent);
 
+ucs_status_t
+uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
```
yosefe: pls revert to prev location

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -146,15 +149,16 @@ uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md, int atomic,
 
     klm = UCT_IB_MLX5DV_ADDR_OF(create_mkey_in, in, klm_pas_mtt);
     for (i = 0; i < ksm_data->mr_num; i++) {
-        UCT_IB_MLX5DV_SET64(klm, klm, address,
-                            (uintptr_t)ksm_data->mrs[i]->addr);
+        UCT_IB_MLX5DV_SET64(klm, klm, address, (uintptr_t)mr_address);
         UCT_IB_MLX5DV_SET(klm, klm, byte_count, ksm_data->mrs[i]->length);
         UCT_IB_MLX5DV_SET(klm, klm, mkey, ksm_data->mrs[i]->lkey);
         klm = UCS_PTR_BYTE_OFFSET(klm, UCT_IB_MLX5DV_ST_SZ_BYTES(klm));
+
```
yosefe, line 156: remove space line (and align = on column)

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -194,12 +198,15 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_addr(
 
 static ucs_status_t
 uct_ib_mlx5_devx_reg_ksm_data_contig(uct_ib_mlx5_md_t *md,
-                                     uct_ib_mlx5_devx_mr_t *mr, off_t offset,
+                                     uct_ib_mlx5_devx_mem_t *memh,
+                                     uct_ib_mr_type_t mr_type,
```
yosefe, line 202: maybe pass mr_address as extra param (instead of memh+mr_type) since uct_ib_mlx5_devx_reg_ksm_data_contig is low-level function

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1469,7 +1476,7 @@ static ucs_status_t uct_ib_mlx5_devx_xgvmi_umem_mr(uct_ib_mlx5_md_t *md,
     size_t length;
     void *mkc;
 
-    address = memh->mrs[UCT_IB_MR_DEFAULT].super.ib->addr;
+    address = memh->addr;
```
yosefe, line 1479: maybe remove the local var address ?

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -245,6 +245,7 @@ typedef union {
 
 typedef struct {
     uct_ib_mem_t            super;
+    void                    *addr;
```
yosefe, line 248: addr -> address

src/uct/ib/base/ib_md.c
ivankochin: Reverted.

src/uct/ib/base/ib_md.h
ivankochin: Done.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Done.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Done.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Done.

src/uct/ib/mlx5/ib_mlx5.h
ivankochin: Done.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -565,9 +569,9 @@ uct_ib_mlx5_devx_reg_mt(uct_ib_md_t *ibmd, void *address, size_t length,
         goto err_free;
     }
 
-    status = uct_ib_mlx5_devx_reg_ksm_data(md, is_atomic, ksm_data, length, 0,
-                                           "multi-thread key", &ksm_data->dvmr,
-                                           &mkey);
+    status = uct_ib_mlx5_devx_reg_ksm_data(md, is_atomic, address, ksm_data,
```
brminich, line 572: seems like number of parameters can be reduced for this func and for `uct_ib_mlx5_devx_reg_ksm_data_contig` if pass memh

brminich (approved):

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -198,23 +197,20 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_addr(
 
 static ucs_status_t
 uct_ib_mlx5_devx_reg_ksm_data_contig(uct_ib_mlx5_md_t *md,
-                                     uct_ib_mlx5_devx_mem_t *memh,
-                                     uct_ib_mr_type_t mr_type,
-                                     off_t offset,
-                                     int atomic, const char *reason,
+                                     uct_ib_mlx5_devx_mr_t *mr, off_t offset,
+                                     void *addr, int atomic, const char *reason,
```
yosefe, line 201: addr -> address or mr_address

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: it's the prev version, would need to pass mr_type

brminich (approved):

cponder:
Can you tell us if this make it into the V1.15.0-rc4 or another upcoming tag?

ivankochin:
> Can you tell us if this make it into the V1.15.0-rc4 or another upcoming tag?

Yes, it was included to RC4.

Issue #9283, opened 2023-08-09
UCX_MEM_MMAP_HOOK_MODE=reloc assumes dynamic table is relocated


rbradford:
When using `UCX_MEM_MMAP_HOOK_MODE=reloc` (vs `=bistro`) the code for parsing the relocation tables assumes that the loader has relocated the entries in the dynamic table (`PT_DYNAMIC`) at runtime. Unfortunately depending on the C library or the architecture this is not the case.

The following test program (as derived from the UCX source code)
```C
#define _GNU_SOURCE
#include <link.h>
#include <stdlib.h>
#include <stdio.h>

#define UCS_PTR_BYTE_OFFSET(_ptr, _offset) \
        ((void *)((intptr_t)(_ptr) + (intptr_t)(_offset)))

static uintptr_t ucm_reloc_get_entry(ElfW(Addr) base, const ElfW(Phdr) *dphdr,
                                     ElfW(Sxword) tag)
{
        ElfW(Dyn) *entry;
        for (entry = (void *)(base + dphdr->p_vaddr); entry->d_tag != 0; ++entry) {
                if (entry->d_tag == tag) {
                        return entry->d_un.d_val;
                }
        }
        return 0;
}

static int callback(struct dl_phdr_info *info, size_t size, void *data)
{
        void *symtab, *strtab;
        ElfW(Phdr) *phdr, *dphdr;
        int found_pt_load;

        dphdr = NULL;
        found_pt_load = 0;
        for (int i = 0; i < info->dlpi_phnum; ++i) {
                phdr = UCS_PTR_BYTE_OFFSET(info->dlpi_phdr, sizeof(ElfW(Phdr)) * i);
                if (phdr->p_type == PT_LOAD) {
                        found_pt_load = 1;
                } else if (phdr->p_type == PT_DYNAMIC) {
                        dphdr = phdr;
                }
        }

        if (found_pt_load != 1) {
                return 0;
        }

        symtab = (void *)ucm_reloc_get_entry(info->dlpi_addr, dphdr, DT_SYMTAB);
        strtab = (void *)ucm_reloc_get_entry(info->dlpi_addr, dphdr, DT_STRTAB);
        printf("name=%s dlpi_addr=%p symtab=%p strtab=%p\n", info->dlpi_name, 
               (void *)info->dlpi_addr, symtab, strtab);
        return 0;
}

int main(int argc, char *argv[])
{
        dl_iterate_phdr(callback, NULL);
[...]

PR #9284, opened 2023-08-09
UCM/UTIL/RELOC: Don't assume dynamic table values are pointers

rbradford:
On some combinations of C library implementation and architecture the entries in the PT_DYNAMIC table are not relocated and stored as pointers. Instead the values are offsets from the address that the library is loaded at.

This was discovered when using UCX_MEM_MMAP_HOOK_MODE=reloc on RISC-V however it is not limited to RISC-V only as some other C library implementations (like musl) will also not update the dynamic table with pointers

Unfortunately since the internal glibc headers that expose whether the value is relocated on not are not available the best solution is to use a heuristic to determine if the values needed from the table (strtab, symtab, etc) are offsets or pointers is by comparing against the library start address.

Since this should only be done for the values from the table that are treated like pointers create a new function ucm_reloc_get_pointer() that applies the heuristic.

Fixes: #9283

Co-developed-by: Rob Bradford <rbradford@rivosinc.com>

## What
Improves the `UCX_MEM_MMAP_HOOK_MODE=reloc` support to remove the assumption that the values in the dynamic (`PT_DYNAMIC`) table have been relocated and are pointers rather than offsets.

## Why?
Unfortunately whether the table values are pointers or offsets is architecture/C library specific and failure to do this means that this functionality does not work on systems that don't have an updated dynamic relocation table.

## How ?
A heuristic is applied to identify if the value from the table is and offset or address by comparing against the address of where the library is loaded. If it is an offset the load address is added.

### DIFF
--- a/src/ucm/util/reloc.c
+++ b/src/ucm/util/reloc.c
@@ -91,6 +91,23 @@ ucm_reloc_get_entry(ElfW(Addr) base, const ElfW(Phdr) *dphdr, ElfW(Sxword) tag)
     return 0;
 }
 
+static void *
+ucm_reloc_get_pointer(ElfW(Addr) base, const ElfW(Phdr) *dphdr, ElfW(Sxword) tag)
+{
+    uintptr_t entry;
+    entry = ucm_reloc_get_entry(base, dphdr, tag);
+    /*
+     * On some combinations of C library and architecture the entries in the
+     * dynamic table are not relocated. As a heuristic check if the value from
+     * the table looks like a pointer or an offset by checking if it is below
+     * the loaded address of the library.
+    */
+    if ((entry > 0) && (entry < base)) {
+        return UCS_PTR_BYTE_OFFSET(base, entry);
+    }
+    return (void *)entry;
+}
+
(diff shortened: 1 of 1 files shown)


src/ucm/util/reloc.c
```diff
@@ -91,6 +91,23 @@ ucm_reloc_get_entry(ElfW(Addr) base, const ElfW(Phdr) *dphdr, ElfW(Sxword) tag)
...
+{
+    uintptr_t entry;
+    entry = ucm_reloc_get_entry(base, dphdr, tag);
+    /*
+     * On some combinations of C library and architecture the entries in the
+     * dynamic table are not relocated. As a heuristic check if the value from
+     * the table looks like a pointer or an offset by checking if it is below
+     * the loaded address of the library.
+    */
+    if ((entry > 0) && (entry < base)) {
```
tvegas1, line 105: it might be possible to avoid heuristic, tried to detail on corresponding issue: https://github.com/openucx/ucx/issues/9283#issuecomment-1673644180

PR #9285, opened 2023-08-09
UCT/IB/MLX5: Remove ib_md operations and move logic to devx_md

yosefe:
## Why
Last part of ib_md cleanup - remove most operation callbacks and move specific logic to ib_mlx5_devx_md


yosefe:
@Artemy-Mellanox @brminich can you pls take a loot?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1580,8 +1688,98 @@ uct_ib_mlx5_devx_reg_exported_key(uct_ib_md_t *ib_md, uct_ib_mem_t *ib_memh)
...
+        }
+
+        uct_ib_md_pack_exported_mkey(&md->super, memh->exported_lkey,
+                                     mkey_buffer);
+        return UCS_OK;
+    }
+
+    if (ENABLE_PARAMS_CHECK &&
+        (flags & (UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA |
+                  UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO)) &&
```
brminich, line 1736: minor: this can be a short inlined func (this check is used tiwce in this func and also would look better)

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1580,8 +1688,98 @@ uct_ib_mlx5_devx_reg_exported_key(uct_ib_md_t *ib_md, uct_ib_mem_t *ib_memh)
...
+{
+    uct_ib_mlx5_md_t *md         = ucs_derived_of(uct_md, uct_ib_mlx5_md_t);
+    uct_ib_mlx5_devx_mem_t *memh = uct_memh;
+    ucs_status_t status;
+    unsigned flags;
+    uint32_t rkey;
+
+    flags = UCS_PARAM_VALUE(UCT_MD_MKEY_PACK_FIELD, params, flags, FLAGS, 0);
+    if (flags & UCT_MD_MKEY_PACK_FLAG_EXPORT) {
+        if (ENABLE_PARAMS_CHECK &&
```
brminich, line 1709: imo we could avoid checking `ENABLE_PARAMS_CHECK` inside this func, also it is under EXPORT check

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: done

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: IMO EXPORT is still fast path because we could be exporting key from rcache

src/uct/ib/base/ib_md.h
```diff
@@ -597,32 +350,25 @@ void uct_ib_md_free(uct_ib_md_t *md);
...
+                                 void *addr, size_t length);
 
 /**
  * Check if IB md device has ECE capability
  */
 void uct_ib_md_ece_check(uct_ib_md_t *md);
 
-
 ucs_status_t
 uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
```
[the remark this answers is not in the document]
yosefe: this is a big change, and the current PR is already big enough..
also we may want to reg_mt in mlx5dv as well with UMR

yosefe:
@Artemy-Mellanox @brminich can you pls review again?

Artemy-Mellanox (approved):

brminich (approved):

PR #9286, opened 2023-08-13
AZP/PERF: Pass GitHub PR as param

Alexey-Rivkin:
## What
Resolve the error in perf pipeline:
"fatal: reference is not a tree"

### DIFF
--- a/buildlib/azure-pipelines-perf.yml
+++ b/buildlib/azure-pipelines-perf.yml
@@ -100,6 +100,7 @@ stages:
 
               set -xeE -o pipefail
               python3 -u perfx.py \
+                --source-branch $(Build.SourceBranch) \
                 --config ucx-rdmz.yml \
                 --before "$SHA_Before" \
                 --after "$SHA_After" \


Alexey-Rivkin:
Working build:
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=67405&view=logs&j=94be751c-9e4c-5c28-d13e-4c720246b05d&t=af4fd65c-9e75-50b3-0a86-2048b64e0183&l=44
(Depends on pending changes in PerfX)

buildlib/azure-pipelines-perf.yml
```diff
@@ -99,6 +99,7 @@ stages:
               module load /hpc/local/etc/modulefiles/hpcx-gcc
 
               set -xeE -o pipefail
+              export SOURCE_BRANCH=$(Build.SourceBranch)
```
yosefe, line 102: why needed? who uses this var?

yosefe: if it's perfx.py - maube should pass it by cmdline param?

Alexey-Rivkin: Used by the PerfX code:
https://github.com/Mellanox-lab/PerfX/pull/53

Alexey-Rivkin: Passing as a param is an option, but would require bigger changes at PerfX.
Need to add args parsing, then path it all the way to the library that actually uses it.
I would prefer the current solution.

yosefe: but i see other things are passed as parameters: "--config", "--before", "--after"
why SOURCE_BRANCH is different?

Alexey-Rivkin: Good point, will do. 
I was expecting Azure built-in vars to be available in ENV, like it works with "AGENT_ID".
That's how I started with using it first and exporting later.

yosefe: IMO, PerfX should not expect any Azure env vars, so we could run it also from command line

Alexey-Rivkin: Done

yosefe (approved):
pls squash

yosefe (approved):

PR #9289, opened 2023-08-15
UCP/API: Add ucp_rkey_compare() API

tvegas1:
## What
Add new API for remote key comparisons, proto v2 support only.
Add flag to request ucp_mem_map() to perform memory registration on specific range.

## Why ?
We need to compare remote keys to deduplicate received remote keys, > 0, = 0 or < 0

## How ?

Usage:
```
ucp_rkey_compare_params_t params = {};
status = ucp_rkey_compare(context, rkey1, rkey2, &params, &result);
```

### DIFF
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -545,20 +545,29 @@ enum {
      * mapping up-front, and mapping them later when they are accessed by
      * communication routines.
      */
-    UCP_MEM_MAP_NONBLOCK = UCS_BIT(0),
+    UCP_MEM_MAP_NONBLOCK       = UCS_BIT(0),
 
     /**
      * Identify requirement for allocation, if passed address is not a
      * null-pointer, then it will be used as a hint or direct address for
      * allocation.
      */
-    UCP_MEM_MAP_ALLOCATE = UCS_BIT(1),
+    UCP_MEM_MAP_ALLOCATE       = UCS_BIT(1),
 
     /**
      * Don't interpret address as a hint: place the mapping at exactly that
      * address. The address must be a multiple of the page size.
      */
-    UCP_MEM_MAP_FIXED    = UCS_BIT(2)
+    UCP_MEM_MAP_FIXED          = UCS_BIT(2),
+
+    /**
+     * Register the memory region so its remote access key would likely be
+     * equal to remote access keys received from other peers, when compared with
+     * @ref ucp_rkey_compare. This flag is a hint. When remote access keys
+     * received from different peers are compared equal, they can be used
+     * interchangeably, avoiding the need to keep all of them in memory.
+     */
+    UCP_MEM_MAP_SYMMETRIC_RKEY = UCS_BIT(3)
 };
 
 
@@ -1254,6 +1263,24 @@ typedef struct ucp_worker_attr {
      */
     size_t                max_debug_string;
 } ucp_worker_attr_t;
+
+
+/**
+ * @ingroup UCP_MEM
+ * @brief Tuning parameters for the comparison function @ref ucp_rkey_compare
+ *
+ * The structure defines the parameters that can be used for UCP library remote
+ * keys comparison using @ref ucp_rkey_compare routine.
+ *
+ */
+typedef struct ucp_rkey_compare_params {
+    /**
+     * Mask of valid fields in this structure, must currently be zero. Fields
+     * not specified in this mask will be ignored.
+     * Provides ABI compatibility with respect to adding new fields.
+     */
+    uint64_t                field_mask;
+} ucp_rkey_compare_params_t;
 
 
 /**
@@ -2098,6 +2125,31 @@ void ucp_cleanup(ucp_context_h context_p);
  */
 ucs_status_t ucp_context_query(ucp_context_h context_p,
                                ucp_context_attr_t *attr);
+
+
+/**
+ * @ingroup UCP_MEM
+ * @brief Compare two remote keys
+ *
+ * This routine compares two remote keys.
+ *
+ * It sets the @a result argument to < 0 if rkey1 is lower than rkey2, 0 if they
+ * are equal or > 0 if rkey1 is greater than rkey2. The result value can be used
+ * for sorting remote keys.
+ *
+ * @param [in]  context     Handle to @ref ucp_context_h
+ * @param [in]  rkey1       First rkey to compare
+ * @param [in]  rkey2       Second rkey to compare
+ * @param [in]  params      Additional parameters to the comparison
+ * @param [out] result      Result of the comparison
+ *
+ * @return UCS_OK                - @a result contains the comparison result
+ * @return UCS_ERR_INVALID_PARAM - The routine arguments are invalid
+ * @return Other                 - Error code as defined by @ref ucs_status_t
+ */
+ucs_status_t
+ucp_rkey_compare(ucp_context_h context, ucp_rkey_h rkey1, ucp_rkey_h rkey2,
+                 const ucp_rkey_compare_params_t *params, int *result);
 
 
 /**
--- a/src/ucp/core/ucp_rkey.c
+++ b/src/ucp/core/ucp_rkey.c
@@ -1181,3 +1181,10 @@ void ucp_rkey_proto_select_dump(ucp_worker_h worker,
     ucp_proto_select_info(worker, rkey_config->key.ep_cfg_index, rkey_cfg_index,
                           &rkey_config->proto_select, 0, strb);
 }
+
+ucs_status_t
+ucp_rkey_compare(ucp_context_h context, ucp_rkey_h rkey1, ucp_rkey_h rkey2,
+                 const ucp_rkey_compare_params_t *params, int *result)
+{
+    return UCS_ERR_UNSUPPORTED;
+}


src/ucp/api/ucp.h
```diff
@@ -190,6 +190,15 @@ enum ucp_worker_params_field {
 };
 
 
+/**
+ * @ingroup UCP_MEM
+ * @brief UCP rkey comparison parameters field mask.
+ */
+enum ucp_rkey_compare_params_field {
+    UCP_RKEY_COMPARE_PARAM_FIELD_MODE = UCS_BIT(0) /**< Comparison mode */
```
yosefe, line 198: do we need mode in the initial implementation? since only one mode is supported now
maybe not define mode now, and if needed - extend it in the future with field_mask

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2137,35 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
                                ucp_context_attr_t *attr);
 
 
+/**
+ * @ingroup UCP_MEM
+ * @brief Compare two remote keys according to parameters
+ *
+ * This routine returns equal if all the transports can compare equal and
+ * configurations are equal. At least one compare mode is needed.
```
yosefe, line 2145: this sentence is not clear

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2137,35 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ *
+ * The comparison result is < 0 if rkey1 is lower than rkey2, 0 if they
+ * are equal or > 0 if rkey1 is greater than rkey2.
+ *
+ * If rkey2 contains one or more transports that are not comparable,
```
yosefe, line 2150: what does it mean "transport that is not comparable"?

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2137,35 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ * appropriate error is returned to allow the application to not try to later
+ * deduplicate against it. The api can be called with rkey1 equal to NULL, rkey2
+ * is considered greater than the NULL key.
+ *
+ * @param [in]  context     Handle to @ref ucp_context_h
+ * @param [in]  params      Used to tune the type of comparison requested
```
yosefe, line 2156: params Additional parameters to the comparison

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2137,35 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ * @param [in]  rkey1       First rkey to compare, can be null
```
yosefe, line 2157: why we say it can be null?

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2137,35 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ * @param [in]  rkey2       Second rkey to compare
+ * @param [out] result      Result of the comparison
+ *
+ * @return Error code or UCS_OK if result contains the comparison result. If
+ * rkey2 is not a comparable key, UCS_ERR_REJECTED is returned. If rkey2 is
+ * null, UCS_ERR_INVALID_PARAM is returned.
+ */
+ucs_status_t
+ucp_rkey_compare(ucp_context_h context, const ucp_rkey_compare_params_t *params,
```
yosefe, line 2166: order of args should be:
context, rkey1, rkey2, params, result

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2137,35 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+                 const ucp_rkey_h rkey1, const ucp_rkey_h rkey2, int *result);
```
yosefe, line 2167: remove const from ucp_rkey_h, for consistency with other APIs

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2137,35 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
                                ucp_context_attr_t *attr);
 
 
+/**
+ * @ingroup UCP_MEM
+ * @brief Compare two remote keys according to parameters
```
yosefe, line 2142: @brief Compare two remote keys

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2137,35 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ *
+ * This routine returns equal if all the transports can compare equal and
+ * configurations are equal. At least one compare mode is needed.
+ *
+ * The comparison result is < 0 if rkey1 is lower than rkey2, 0 if they
+ * are equal or > 0 if rkey1 is greater than rkey2.
```
yosefe, line 2148: The comparison result is < 0 if rkey1 is lower than rkey2, 0 if they
are equal or > 0 if rkey1 is greater than rkey2. The remote key greater-than
semantic is arbitrary and implementation dependent, nevertheless it's guaranteed
to be consistent (transitive and antisymmetric). It can be used, for example, to 
perform an efficient binary search in a sorted array of remote keys.

src/ucp/api/ucp.h
tvegas1: kept only `field_mask`, removed `enum ucp_rkey_compare_params_fields`

src/ucp/api/ucp.h
tvegas1: fixed, will use similar message in corresponding uct api header

src/ucp/api/ucp.h
tvegas1: fixed

src/ucp/api/ucp.h
tvegas1: removed. proposal was to identify ucp rkeys that would never be subject to compare as equal, so to avoid inserting them into sorted array, in order to avoid pointer-related memory consumption. after discussing with @brminich we now simplify the api making the compare function consistent and we treat all ucp rkey the same.

src/ucp/api/ucp.h
tvegas1: removed as per [comment](https://github.com/openucx/ucx/pull/9289#discussion_r1297110413)

src/ucp/api/ucp.h
tvegas1: fixed

src/ucp/api/ucp.h
tvegas1: removed, as per https://github.com/openucx/ucx/pull/9289#discussion_r1297110413

src/ucp/api/ucp.h
tvegas1: fixed

src/ucp/api/ucp.h
tvegas1: fixed

yosefe (approved):

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2124,33 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
                                ucp_context_attr_t *attr);
 
 
+/**
+ * @ingroup UCP_MEM
+ * @brief Compare two remote keys
+ *
+ * This routine returns equal if all the transports can compare equal and
+ * configurations are equal. At least one compare mode is needed.
```
brminich, line 2132: seems to be a leftover about modes

tvegas1: removed thanks

src/ucp/api/ucp.h
```diff
@@ -1256,6 +1265,21 @@ typedef struct ucp_worker_attr {
 } ucp_worker_attr_t;
 
 
+/**
+ * @ingroup UCP_MEM
+ * @brief Tuning parameters for the comparison function @ref ucp_rkey_compare
+ *
```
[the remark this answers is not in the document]
tvegas1: fixed

brminich (approved):

yosefe (dismissed):

yosefe:
@shamisp @tonycurtis can you pls take a look?

src/ucp/api/ucp.h
```diff
@@ -545,20 +545,29 @@ enum {
...
      */
-    UCP_MEM_MAP_FIXED    = UCS_BIT(2)
+    UCP_MEM_MAP_FIXED          = UCS_BIT(2),
+
+    /**
+     * Register the memory region so its remote access key would likely be
+     * equal to remote access keys received from other peers, when compared with
+     * @ref ucp_rkey_compare. This flag is best-effort. When remote access keys
+     * received from different peers are equal, they can be used
+     * interchangeably, avoiding the need to keep all of them in memory.
```
shamisp, line 568: This flag is a hint.  `ucp_rkey_compare` routine should be used to check if the two key are equal.

tvegas1: added hint and detailed that they can be used when they are compared equal

yosefe (dismissed):

src/ucp/api/ucp.h
brminich: @shamisp, pls re-review

brminich (approved):

src/ucp/api/ucp.h
shamisp: I think "best-effort hint" sound odd. I think " a hint" is cleaner description. Hint is always a best effort.

src/ucp/api/ucp.h
```diff
@@ -1256,6 +1265,20 @@ typedef struct ucp_worker_attr {
 } ucp_worker_attr_t;
 
 
+/**
+ * @ingroup UCP_MEM
+ * @brief Tuning parameters for the comparison function @ref ucp_rkey_compare
```
shamisp, line 1270: Please add full description in addition to `@brief`. We typically don't allow just brief.

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2123,33 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
                                ucp_context_attr_t *attr);
 
 
+/**
+ * @ingroup UCP_MEM
+ * @brief Compare two remote keys
+ *
+ * This routine returns equal if all the transports can compare equal and
```
shamisp, line 2130: ` This routine returns equal if all the transports can compare equal and configurations are equal.` - this is very puzzling.  What are you trying to state here ?
`This routine compares two remote keys`.

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2123,33 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ * configurations are equal.
+ *
+ * The comparison result is < 0 if rkey1 is lower than rkey2, 0 if they
```
shamisp, line 2133: It sets the `result` argument to ....`

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2123,33 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ * are equal or > 0 if rkey1 is greater than rkey2. The remote key greater-than
+ * semantic is arbitrary and implementation dependent, nevertheless it's
```
shamisp, line 2135: `greater-than semantics` ?

shamisp: `arbitrary and implementation dependent` ? arbitrary == random .  I don't think the semantics can be random and consistent in the same time.

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2123,33 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ * guaranteed to be consistent (transitive and antisymmetric). It can be used,
+ * for example, to perform an efficient binary search in a sorted array of
```
shamisp, line 2137: Maybe just state that it can be used for sorting ( obviously one can use it for search later on)

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2123,33 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ * remote keys.
+ *
+ * @param [in]  context     Handle to @ref ucp_context_h
+ * @param [in]  rkey1       First rkey to compare
+ * @param [in]  rkey2       Second rkey to compare
+ * @param [in]  params      Additional parameters to the comparison
+ * @param [out] result      Result of the comparison
+ *
+ * @return Error code or UCS_OK if result contains the comparison result
```
shamisp, line 2146: In what cases do we return an error ? Can one detect rkey corruption ?

shamisp (changes requested):
please see above comments

src/ucp/api/ucp.h
shamisp: Not really sure what error will mean here...

src/ucp/api/ucp.h
tvegas1: replaced sentence with `This routine compares two remote keys`.

src/ucp/api/ucp.h
tvegas1: fixed

src/ucp/api/ucp.h
tvegas1: based on comment above and this one I replaced the whole paragraph with:
`It can be used for sorting.`

src/ucp/api/ucp.h
tvegas1: added list of error cases

src/ucp/api/ucp.h
tvegas1: based on comments, replaced by `it can be used for sorting`.

src/ucp/api/ucp.h
tvegas1: added description

src/ucp/api/ucp.h
tvegas1: removed `best-effort`

yosefe (approved):

brminich (approved):

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2127,32 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
+ * are equal or > 0 if rkey1 is greater than rkey2. It can be used for sorting.
+ *
+ * @param [in]  context     Handle to @ref ucp_context_h
+ * @param [in]  rkey1       First rkey to compare
+ * @param [in]  rkey2       Second rkey to compare
+ * @param [in]  params      Additional parameters to the comparison
+ * @param [out] result      Result of the comparison
+ *
+ * @return UCS_OK                - @a result contains the comparison result
+ * @return UCS_ERR_UNSUPPORTED   - The operation parameters passed with
```
shamisp, line 2146: I really don't like the idea of non-supported. If it is non supported is really means keys are not equal. In this case we just should return consistent  `<` or `>' based on some other parameters. E.g u can use pointer values of the underlaying layer, etc.

shamisp (changes requested):

src/ucp/api/ucp.h
yosefe: I agree. If the keys are valid we should always be able to compare keys as if they are binary strings.
@tvegas1 in which case we cannot compare?

tvegas1: The proposal is to have using rkey compare with proto v1 return unsupported. As this ucp api seems there could be other future cases.

tvegas1: Alternative proposal:
- Have all UCT rkey compare implement proper compare and return UCS_OK (the current case). But keep passing them a params and return a ucs_status_t for future proof, add assert preventing not supported error code in UCP after UCT rkey compare.
- In case UCT rkey compare would return an error (future): have UCP rkey compare return it immediately.

For the protov1 check:
- return invalid params
For bad params->field_mask:
- return invalid params

yosefe: why can't we compare unpacked keys with protov1?

tvegas1: I would need to check in details, but rkey->cache content might not all be constant, for instance there seems to be rkey resolve at send time. I was under assumption that we did not need v1 support.

yosefe: per offline discussion, will remove ERR_UNSUPPORTED return value from doc

tvegas1: fixed, will add comparison for v1 proto too in implementation.

yosefe (approved):

src/ucp/api/ucp.h
shamisp: @yosefe @tvegas1 - to confirm both proto v1 and proto v2 will be supported ?

src/ucp/api/ucp.h
```diff
@@ -2100,6 +2127,30 @@ ucs_status_t ucp_context_query(ucp_context_h context_p,
...
 
 
+/**
+ * @ingroup UCP_MEM
+ * @brief Compare two remote keys
+ *
+ * This routine compares two remote keys.
+ *
+ * It sets the @a result argument to < 0 if rkey1 is lower than rkey2, 0 if they
+ * are equal or > 0 if rkey1 is greater than rkey2. It can be used for sorting.
```
shamisp, line 2137: `It can be used for sorting` -> `e.g. the result value can be used for sorting keys.`

tvegas1: fixed

src/ucp/api/ucp.h
yosefe: yes, both will be

shamisp (approved):

src/ucp/api/ucp.h
shamisp: OK

yosefe (approved):
pls squash

PR #9296, opened 2023-08-17
UCS/DATASTRUCT: Include ucs_status in array.inl

shasson5:
## What
Include ucs_status in array.inl

## Why ?
Array class uses ucs_status_t

### DIFF
--- a/src/ucs/datastruct/array.h
+++ b/src/ucs/datastruct/array.h
@@ -9,6 +9,7 @@
 
 #include <ucs/sys/compiler_def.h>
 #include <ucs/sys/preprocessor.h>
+#include <ucs/type/status.h>
 
 BEGIN_C_DECLS
 


src/ucs/datastruct/array.inl
```diff
@@ -8,7 +8,7 @@
 #define UCS_ARRAY_INL_
 
 #include "array.h"
-
+#include <ucs/type/status.h>
```
yosefe, line 11: 1. space line before and after
2. add it to array.h because it uses ucs_status_t

shasson5: array.h has status only inside macros, so it's not really needed threre.

gleon99: @shasson5 every file which `#include`s `array.h` must have `ucs_status_t`, therefore better it is included from there.

gleon99 (approved):

yosefe (approved):
pls squash

PR #9298, opened 2023-08-17
UCP/RNDV: Use only reachable MDs for protocol selection.

rakhmets:
## What
Enhanced remote protocol selection during rendezvous initialization, taking into account reachability of memory domains.
The reachability is checking during remote key unpacking.

## Why ?
Ignoring this information can lead to performance issues. E.g. RTR protocol can be mistakenly selected as a remote protocol, even if there is no valid MD for the protocol. Which leads to fallback to rndv/am protocol.

### DIFF
--- a/src/ucp/rndv/proto_rndv.c
+++ b/src/ucp/rndv/proto_rndv.c
@@ -154,6 +154,9 @@ static ucs_status_t ucp_proto_rndv_ctrl_select_remote_proto(
     rkey_config_key.ep_cfg_index = ep_cfg_index;
     rkey_config_key.sys_dev      = params->mem_info.sys_dev;
     rkey_config_key.mem_type     = params->mem_info.type;
+
+    rkey_config_key.unreachable_md_map = 0;
+
     for (lane = 0; lane < ep_config->key.num_lanes; ++lane) {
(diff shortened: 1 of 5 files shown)


src/ucp/rndv/proto_rndv.c
```diff
@@ -262,6 +262,9 @@ ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params,
     /* Initialize estimated memory registration map */
     ucp_proto_rndv_ctrl_get_md_map(params, &rpriv->md_map, &rpriv->sys_dev_map,
                                    rpriv->sys_dev_distance);
+    if (params->super.super.rkey_config_key != NULL) {
```
yosefe, line 265: Pls add code comment why we're doing this

rakhmets: Added

yosefe: or we can do it in the if/else in lines 244-260. Seems we can move the call to ucp_proto_rndv_ctrl_get_md_map before that if/else.

yosefe (approved):

brminich (approved):

PR #9299, opened 2023-08-17
UCT/IB: avoid using ibv mr addr and fix relaxed order memh size - v1.15

ivankochin:
## What
Backport of #9282. To make this PR work the #9200 was also backported as part of this PR.

## How ?
The `uct_ib_md_parse_reg_methods` was manually moved later because of deoendency on `memh_struct_size` which definition was moved in another place in #9200.

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -1258,7 +1258,7 @@ static void uct_ib_md_release_device_config(uct_ib_md_t *md)
     ucs_free(md->custom_devices.specs);
 }
 
-static ucs_status_t
+ucs_status_t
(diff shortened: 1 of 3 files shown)


src/uct/ib/base/ib_md.c
```diff
@@ -1524,9 +1524,21 @@ ucs_status_t uct_ib_md_open(uct_component_t *component, const char *md_name,
     return status;
 }
 
+
```
yosefe, line 1527: remove space line

brminich: fixed

yosefe (approved):
pls squash (or amend this fix into one of the 2 previous commits)

PR #9305, opened 2023-08-22
UCP: Fix ep reconfig because of fp8 inaccuracy

brminich:
## What
1. When deciding what rma bw lanes to drop, local bw value needs to be packed/unpacked in fp8 format to make sure both sides make the same decision
2. Device num_paths coefficient should be applied after fp8 pack/unpack. Otherwise the ratio effect maybe just hidden by fp8 inaccuracy

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1241,15 +1241,20 @@ ucp_wireup_iface_avail_bandwidth(const ucp_worker_iface_t *wiface,
     double eps                = 1e-3;
     double local_bw, remote_bw;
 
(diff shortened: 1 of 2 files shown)


yosefe (dismissed):
do we have a test to catch the issue?

brminich:
> 

i was getting this issues with wire-compat tests with addrV2

yosefe:
@brminich maybe worth adding some addrv2 variants to wire combat testing?

brminich:
@yosefe added gtest which reproduces the issue

yosefe (approved):

PR #9307, opened 2023-08-23
UCS/DISTANCE: Update Intel Ice Lake sys root bandwidth

ivankochin:
## What
Set new value for sys root Ice Lake bandwidth

## Why ?
Current value is 220 MB/s. Running `ib_write_bw` testing showed that actual value is significantly higher. String for testing:

```
./ib_write_bw --use_cuda=0 --use_cuda_dmabuf --size=262144 --inline_size=64 -D 60 127.0.0.1
```
Checked on Intel(R) Xeon(R) Platinum 8380 with NVIDIA A100.

CPUID related info was took from [wikichip](https://en.wikichip.org/wiki/intel/microarchitectures/ice_lake_(server)#CPUID).


yosefe (approved):

ivankochin:
It was decided to close this PR for now, to better investigate the correctnes of distance detection (pci vs sysroot)

PR #9309, opened 2023-08-24
Revert "UCS/DISTANCE: Update Intel Ice Lake sys root bandwidth"

ivankochin:
## What
This reverts commit f33c9aa514891a38a8d416960e6b2809a3bd7100.

## Why ?
This commit was merged by mistake. See [comment](https://github.com/openucx/ucx/pull/9307#issuecomment-1691472650).

### DIFF
--- a/src/ucs/sys/topo/base/topo.c
+++ b/src/ucs/sys/topo/base/topo.c
@@ -308,9 +308,6 @@ static void ucs_topo_sys_root_distance(ucs_sys_dev_distance_t *distance)
 {
     distance->latency = 500e-9;
     switch (ucs_arch_get_cpu_model()) {
-    case UCS_CPU_MODEL_INTEL_ICELAKE:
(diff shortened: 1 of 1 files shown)


src/ucs/sys/topo/base/topo.c
```diff
@@ -308,9 +308,6 @@ static void ucs_topo_sys_root_distance(ucs_sys_dev_distance_t *distance)
 {
     distance->latency = 500e-9;
     switch (ucs_arch_get_cpu_model()) {
-    case UCS_CPU_MODEL_INTEL_ICELAKE:
```
yosefe, line 311: i would revert only this part, and keep ICELake CPU detection

ivankochin: Done.

yosefe (approved):

PR #9311, opened 2023-08-24
UCT/API: Remove iface_is_reachable member from iface ops

tvegas1:
## What
Remove `->iface_is_reachable` member of iface ops structure and make API call directly invoke the v2 callback.

## Why ?
Moving to v2.

## How ?
UGNI and proxy ep needs careful review.


yosefe:
Cannot do this since it changes UCT API/ABI

tvegas1:
> Cannot do this since it changes UCT API/ABI

I was assuming ops itself was not part of supported API/ABI, closing then.

PR #9312, opened 2023-08-25
WIREUP: Restrict creation more BW lanes than previously

ivankochin:
## What
Restrict creation more BW lanes than during first lanes initialization.

## Why ?
To prevent endpoint reconfiguration error in case when some BW lanes were dropped due to low BW during first initialization, but there are available resources with higher BW that still can be chosen instead of dropped lanes during second initialization.

This is the short-term solution. The long-term solution is to support endpoint reconfiguration.

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1509,6 +1509,28 @@ ucp_wireup_add_fast_lanes(ucp_worker_h worker,
 }
 
 static unsigned
+ucp_wireup_get_current_num_lanes(ucp_ep_h ep, ucp_lane_type_t type)
+{
+    unsigned current_num_lanes = 0;
+    ucp_lane_index_t lane;
+
+    /* First initialization (current lanes weren't chosen yet) or
+     * CM is enabled (so we can reconfigure the endpoint).
+     */
+    if ((ep->cfg_index == UCP_WORKER_CFG_INDEX_NULL) ||
+        ucp_ep_has_cm_lane(ep)) {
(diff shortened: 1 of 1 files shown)


src/ucp/wireup/select.c
```diff
@@ -1508,6 +1508,30 @@ ucp_wireup_add_fast_lanes(ucp_worker_h worker,
     return num_lanes;
 }
 
+/* Restrict choosing more lanes that were chosen previously to prevent
+ * endpoint reconfiguration error in case of dropping some lanes due to low BW.
+ */
+static void
+ucp_wireup_bw_lanes_check_prev(const ucp_wireup_select_params_t *select_params,
```
yosefe, line 1515: `static unsigned ucp_wireup_get_num_lanes(const ucp_ep_config_key_t *ep_cfg_key, ucp_lane_type lane_type)`

ivankochin: As I understand, it is not guaranteed that EP would have assigned key in that point, so we cannot just use `ucp_ep_config(ep)` to pass it. Do you suggest to move `(ep->cfg_index != UCP_WORKER_CFG_INDEX_NULL)` outside of this function or maybe pass key to `ucp_wireup_search_lanes` as argument?

yosefe: yes, move it outside the function

ivankochin: As it was discussed, changed  the return type to unsigned + passed EP as input parameter.

yosefe (approved):

PR #9313, opened 2023-08-25
TEST/JENKINS: Move functions to buildlib/tools/common.sh

tvegas1:
## What
Move functions to common library.

## Why ?
Script clean up.

## How ?
Use copy/paste first approach for that move.

Note: test_jenkins.sh now uses V=1 on make argument.


yosefe (approved):
pls squash

Issue #9287, opened 2023-08-14
GPUDirect RDMA Performance issue


kzmymmt:
### Describe the bug
I measured bandwidth on GPUDirect RDMA with OSU Micro-Benchmarks osu_bw D D.
Bandwidth was lower with ucx 1.15.0rc3 than with 1.13.1.

```
# OSU MPI-CUDA Bandwidth Test v7.2
# Send Buffer on DEVICE (D) and Receive Buffer on DEVICE (D)
# Size      Bandwidth (MB/s)
# Datatype: MPI_CHAR.
1                       1.60
2                       3.21
4                       6.34
8                      10.81
16                     21.42
32                     42.94
64                     83.96
128                   166.28
256                   329.42
512                   629.98
1024                 1146.88
2048                 2149.59
4096                 3761.13
8192                 6421.73
16384                1484.76
32768                1868.04
65536                2345.00
131072               2879.16
262144               3526.87
524288               4084.24
1048576              4031.50
2097152              3895.92
4194304              3840.19
```

Previously ucx1.13.1 had 24756.14 MB/sec with Size 4194304.
This was close to the speed of the NIC (IB NDR200, 200Gbps) and was fine.

What logs should be taken to identify the problem?

### Steps to Reproduce
- Command line
```
mpirun -host host1,host2 -np 2 -npernode 1 ./osu_bw D D
```
- UCX version used (from github branch XX or release YY) + UCX configure flags (can be checked by `ucx_info -v`)
```
# Library version: 1.15.0
# Library path: /home/user/privateinstall/ucx-1.15.0-rc3/nvhpc22.11-cuda11.8/lib/libucs.so.0
# API headers version: 1.15.0
# Git branch '', revision 
# Configured with: --prefix=/home/user/privateinstall/ucx-1.15.0-rc3/nvhpc22.11-cuda11.8 --with-gdrcopy=/usr/local --with-cuda=/system/apps/ubuntu/20.04-202304/nvhpc/22.11/Linux_x86_64/22.11/cuda --disable-optimizations --enable-logging --disable-debug --disable-assertions --enable-mt --disable-params-check --without-java --enable-cma --with-verbs --without-knem --with-rdmacm --without-rocm --without-xpmem --without-ugni
```
```
$ ucx_info -d | grep -e cuda -e gdr
# Memory domain: cuda_cpy
#     Component: cuda_cpy
[...]

PR #9318, opened 2023-08-28
UCS/SYS/TOPO: Added bw estimation for Sapphire Rapids family - v1.15.x

rakhmets:
## What
Added estimation of bandwidth for Intel Sapphire Rapids family.
## Why ?
To avoid TCP lane selection instead of IB lanes.
https://github.com/openucx/ucx/issues/9287


brminich (approved):

src/ucs/sys/topo/base/topo.c
```diff
@@ -308,6 +308,9 @@ static void ucs_topo_sys_root_distance(ucs_sys_dev_distance_t *distance)
 {
     distance->latency = 500e-9;
     switch (ucs_arch_get_cpu_model()) {
+    case UCS_CPU_MODEL_INTEL_SAPPHIRERAPIDS:
+        distance->bandwidth = 23288 * UCS_MBYTE;
```
ivankochin, line 312: How did you get this number?

rakhmets: The value is copied from the issue. This is a value of bandwidth for message size 262144.

ivankochin (dismissed):

yosefe (dismissed):

ivankochin:
Minor: It seems like we need to update this table too https://github.com/openucx/ucx/blob/master/src/tools/info/sys_info.c#L25

rakhmets:
> Minor: It seems like we need to update this table too https://github.com/openucx/ucx/blob/master/src/tools/info/sys_info.c#L25

Done. Also squashed commits because of the changes are minor.

yosefe (approved):

brminich (approved):

PR #9322, opened 2023-08-28
UCS/MEMORY: Introduce auto memory type

Akshay-Venkatesh:
## What
Introduce special value `UCS_MEMORY_TYPE_AUTO` which when used with `UCX_REG_NONBLOCK_MEM_TYPES=auto` can result in automatically setting the right memory types that should be registered using on demand paging flag when using ibv_reg_mr with non-ucp_mem_map memory. For memory ranges that are passed to `ucp_mem_map`, ODP registrations are forced only when using the flag UCP_MEM_MAP_NONBLOCK.


gleon99:
Hi @Akshay-Venkatesh - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #9333, opened 2023-09-01
UCP/PROTO: Add bias information to performance tree

ivankochin:
## What
Adds information about non-zero bias to the performance tree. e.g.

image

## Why ?
To make performance calculation process more clear.

### DIFF
--- a/src/ucp/proto/proto_init.c
+++ b/src/ucp/proto/proto_init.c
@@ -267,8 +267,13 @@ ucp_proto_init_parallel_stages(const ucp_proto_common_init_params_t *params,
     ucs_array_for_each(elem, &concave) {
         range             = &caps->ranges[caps->num_ranges];
         range->max_length = elem->max_length;
-        range->node       = ucp_proto_perf_node_new_data(params->super.proto_name,
+        if (fabs(bias) > UCP_PROTO_PERF_EPSILON) {
+            range->node   = ucp_proto_perf_node_new_data(params->super.proto_name,
+                                                         "bias %f", bias);
+        } else {
+            range->node   = ucp_proto_perf_node_new_data(params->super.proto_name,
                                                          "");
+        }
 
         /* "single" performance estimation is sum of "stages" with the bias */
         range->perf[UCP_PROTO_PERF_TYPE_SINGLE] =


src/ucp/proto/proto_init.c
```diff
@@ -267,8 +267,13 @@ ucp_proto_init_parallel_stages(const ucp_proto_common_init_params_t *params,
     ucs_array_for_each(elem, &concave) {
         range             = &caps->ranges[caps->num_ranges];
         range->max_length = elem->max_length;
-        range->node       = ucp_proto_perf_node_new_data(params->super.proto_name,
+        if (bias) {
+            range->node   = ucp_proto_perf_node_new_data(params->proto_name,
+                                                         "bias %.2f", bias);
+        } else {
+            range->node   = ucp_proto_perf_node_new_data(params->proto_name,
```
ivankochin, line 274:
```suggestion
        if (bias) {
            range->node   = ucp_proto_perf_node_new_data(params->super.proto_name,
                                                         "bias %.2f", bias);
        } else {
            range->node   = ucp_proto_perf_node_new_data(params->super.proto_name,
```

src/ucp/proto/proto_init.c
```diff
@@ -267,8 +267,13 @@ ucp_proto_init_parallel_stages(const ucp_proto_common_init_params_t *params,
     ucs_array_for_each(elem, &concave) {
         range             = &caps->ranges[caps->num_ranges];
         range->max_length = elem->max_length;
-        range->node       = ucp_proto_perf_node_new_data(params->super.proto_name,
+        if (bias) {
```
yosefe, line 270: if (fabs(bias) > UCP_PROTO_PERF_EPSILON)

ivankochin: Done.

yosefe (approved):

brminich (approved):

src/ucp/proto/proto_init.c
```diff
@@ -267,8 +267,13 @@ ucp_proto_init_parallel_stages(const ucp_proto_common_init_params_t *params,
     ucs_array_for_each(elem, &concave) {
         range             = &caps->ranges[caps->num_ranges];
         range->max_length = elem->max_length;
-        range->node       = ucp_proto_perf_node_new_data(params->super.proto_name,
+        if (fabs(bias) > UCP_PROTO_PERF_EPSILON) {
+            range->node   = ucp_proto_perf_node_new_data(params->super.proto_name,
+                                                         "bias %f", bias);
```
rakhmets, line 272: Maybe to many digits after the dot, `%.3f` ?

ivankochin: There may be cases then when we have note about `0.000` bias since `UCP_PROTO_PERF_EPSILON` is `1e-15`.

PR #9335, opened 2023-09-01
UCT/CUDA/CUDA_IPC: Check memory registration availability.

rakhmets:
## What
Added device attributes check to identify memory registration capability of cuda_ipc transport.

## Why ?
Memory registration method of cuda_ipc memory domain uses cuIpcGetMemHandle. Users can test their device for IPC functionality by calling cuDeviceGetAttribute with CU_DEVICE_ATTRIBUTE_IPC_EVENT_SUPPORTED. However, the attribute was added in CUDA 12. Test memory registration for 1 byte allocated on device in case of previous CUDA versions.

### DIFF
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -35,7 +35,10 @@ static int uct_cuda_ipc_md_check_mem_reg()
     int value;
     int ret;
 
-    if (UCT_CUDADRV_FUNC_LOG_DEBUG(cuCtxGetDevice(&dev)) != UCS_OK) {
+    if ((UCT_CUDADRV_FUNC_LOG_DEBUG(cuCtxGetDevice(&dev)) != UCS_OK) &&
+        /* Context may be not set yet. Check attribute for 0 device, assuming
+         * support is uniform across all devices. */
+        (UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&dev, 0)) != UCS_OK)) {
         return 0;
     }
 
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -28,41 +28,7 @@ static ucs_config_field_t uct_cuda_ipc_md_config_table[] = {
     {NULL}
 };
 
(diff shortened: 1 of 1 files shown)


rakhmets:
> @rakhmets `CU_DEVICE_ATTRIBUTE_IPC_EVENT_SUPPORTED` only implies that Ipc events can be shared with remote process and event can be queried by remote process. I doesn't necessarily have an implication on whether IPC registrations are possible or not.

@Akshay-Venkatesh the following sentence is copied from the description of [cuIpcGetMemHandle](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g6f1b5be767b275f016523b2ac49ebec1) from CUDA driver documentation.

```
Users can test their device for IPC functionality by calling cuDeviceGetAttribute with CU_DEVICE_ATTRIBUTE_IPC_EVENT_SUPPORTED.
 ```
 
 I used the attribute because we use `cuIpcGetMemHandle` during registration method. Is it really incorrect usage for this case?

yosefe (dismissed):

yosefe:
@Akshay-Venkatesh @brminich can you pls review as well?

brminich (dismissed):

rakhmets:
> > @rakhmets `CU_DEVICE_ATTRIBUTE_IPC_EVENT_SUPPORTED` only implies that Ipc events can be shared with remote process and event can be queried by remote process. I doesn't necessarily have an implication on whether IPC registrations are possible or not.
> 
> @Akshay-Venkatesh the following sentence is copied from the description of [cuIpcGetMemHandle](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g6f1b5be767b275f016523b2ac49ebec1) from CUDA driver documentation.
> 
> ```
> Users can test their device for IPC functionality by calling cuDeviceGetAttribute with CU_DEVICE_ATTRIBUTE_IPC_EVENT_SUPPORTED.
> ```
> 
> I used the attribute because we use `cuIpcGetMemHandle` in the registration method. Is it really incorrect usage for this case?

@Akshay-Venkatesh do you propose to use `CU_DEVICE_ATTRIBUTE_INTEGRATED` in case of CUDA12 too?
I think `CU_DEVICE_ATTRIBUTE_IPC_EVENT_SUPPORTED` can be used according to the documentation.

PR #9336, opened 2023-09-01
UCP/API: Implement rkey compare and symmetric key backends

tvegas1:
## What
Add ucp_rkey_compare, uct_rkey_compare, comparison for all transports and ib symmetric enablement.

## Why ?
Added at once with tests, to run sanities.

## How ?
Could be later split in multiple parts if needed.


yosefe:
@tvegas1 pls split the PR, at least to UCP and UCT part.
UCT part can be split by transport/md.
We can review them in parallel.

yosefe:
@tvegas1 is this PR still relevant?

tvegas1:
was split in multiple pr

PR #9338, opened 2023-09-03
TOOLS/PERF: make ucp perftest independent from allocators

roiedanino:
## What
Removing usage of allocators in ucp perftest.

## Why ?
So adding new memory types will be easier as memory handling will be managed only through UCP / UCT api's

## How ?
perftest object will have a pointer to `memcpy` function, anything else related to allocators will be used only in UCT perftest.

### DIFF
--- a/src/tools/perf/lib/libperf.c
+++ b/src/tools/perf/lib/libperf.c
@@ -44,6 +44,18 @@
         return UCS_ERR_UNSUPPORTED; \
     }
 
+#define UCX_PERF_VERBOSE(_log_level, _params, _func, ...) \
+    ({ \
+        ucs_status_t _status = _func(__VA_ARGS__); \
+        if (((_status) != UCS_OK) && \
+            ((_params)->flags & UCX_PERF_TEST_FLAG_VERBOSE)) { \
+            ucs_##_log_level("%s() failed: %s", UCS_PP_MAKE_STRING(_func), \
+                             ucs_status_string(_status)); \
+        } \
+        _status; \
+    })
+
+
 typedef struct {
     union {
         struct {
@@ -208,22 +220,6 @@ void ucx_perf_test_prepare_new_run(ucx_perf_context_t *perf,
         perf->timing_queue[i] = 0;
     }
     ucx_perf_test_start_clock(perf);
-}
-
-static void ucx_perf_test_init(ucx_perf_context_t *perf,
-                               const ucx_perf_params_t *params)
-{
-    perf->params = *params;
-
-    ucs_debug("set send allocator by send mem type %s",
-              ucs_memory_type_names[params->send_mem_type]);
-    perf->send_allocator = ucx_perf_mem_type_allocators[params->send_mem_type];
-
-    ucs_debug("set recv allocator by recv mem type %s",
-              ucs_memory_type_names[params->recv_mem_type]);
-    perf->recv_allocator = ucx_perf_mem_type_allocators[params->recv_mem_type];
-
-    ucx_perf_test_prepare_new_run(perf, params);
 }
 
 void ucx_perf_calc_result(ucx_perf_context_t *perf, ucx_perf_result_t *result)
@@ -848,6 +844,12 @@ static void ucp_perf_worker_progress(void *arg)
     }
 }
 
+static int ucp_perf_test_is_host_mem_only(const ucx_perf_params_t *params)
+{
+    return (params->send_mem_type == UCS_MEMORY_TYPE_HOST) &&
+           (params->recv_mem_type == UCS_MEMORY_TYPE_HOST);
+}
+
 static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
                                               ucp_params_t *ucp_params)
 {
@@ -898,6 +900,11 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
         ucp_params->features |= UCP_FEATURE_WAKEUP;
     }
 
+    if (!ucp_perf_test_is_host_mem_only(params)) {
+        /* Required as sn is being copied over put / get operations */
+        ucp_params->features |= UCP_FEATURE_RMA;
+    }
+
     status = ucx_perf_test_check_params(params);
     if (status != UCS_OK) {
         return status;
@@ -906,32 +913,38 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
     return UCS_OK;
 }
 
-static void ucp_perf_test_destroy_eps(ucx_perf_context_t* perf)
+static ucs_status_ptr_t
+ucp_perf_test_destroy_ep(ucp_ep_h ep, const ucp_rkey_h *rkeys, unsigned num_rkeys, unsigned index)
 {
-    unsigned i, thread_count = perf->params.thread_count;
-    unsigned num_in_prog     = 0;
-    ucs_status_ptr_t **reqs  = ucs_alloca(thread_count * sizeof(*reqs));
+    ucp_request_param_t ep_close_params = {0};
     ucs_status_ptr_t *req;
-    ucs_status_t status;
+    unsigned i;
 
-    for (i = 0; i < thread_count; ++i) {
-        if (perf->ucp.tctx[i].perf.ucp.rkey != NULL) {
-            ucp_rkey_destroy(perf->ucp.tctx[i].perf.ucp.rkey);
-        }
-
-        if (perf->ucp.tctx[i].perf.ucp.ep != NULL) {
-            req = ucp_ep_close_nb(perf->ucp.tctx[i].perf.ucp.ep,
-                                  UCP_EP_CLOSE_MODE_FLUSH);
-
-            if (UCS_PTR_IS_PTR(req)) {
-                reqs[num_in_prog++] = req;
-            } else if (UCS_PTR_STATUS(req) != UCS_OK) {
-                ucs_warn("failed to close ep %p on thread %d: %s\n",
-                         perf->ucp.tctx[i].perf.ucp.ep, i,
-                         ucs_status_string(UCS_PTR_STATUS(req)));
-            }
+    for (i = 0; i < num_rkeys; ++i) {
+        if (rkeys[i] != NULL) {
+            ucp_rkey_destroy(rkeys[i]);
         }
     }
+
+    if (ep != NULL) {
+        req = ucp_ep_close_nbx(ep, &ep_close_params);
+        if (UCS_PTR_IS_PTR(req)) {
+            return req;
+        } else if (UCS_PTR_STATUS(req) != UCS_OK) {
+            ucs_warn("failed to close ep %p on thread %d: %s\n", ep, index,
+                     ucs_status_string(UCS_PTR_STATUS(req)));
+        }
+    }
+
+    return NULL;
+}
+
+static void ucp_perf_release_requests_in_progress(ucx_perf_context_t *perf,
+                                                  ucs_status_ptr_t **reqs,
+                                                  unsigned num_in_prog)
+{
+    ucs_status_t status;
+    unsigned i;
 
     while (num_in_prog != 0) {
         ucp_perf_worker_progress(perf);
@@ -943,6 +956,46 @@ static void ucp_perf_test_destroy_eps(ucx_perf_context_t* perf)
             }
         }
     }
+}
+
+static void ucp_perf_test_destroy_self_eps(ucx_perf_context_t *perf)
+{
+    unsigned thread_count   = perf->params.thread_count;
+    unsigned num_in_prog    = 0;
+    ucs_status_ptr_t **reqs = ucs_alloca(thread_count * sizeof(*reqs));
+    ucs_status_ptr_t *req;
+    unsigned i;
+
+    for (i = 0; i < thread_count; ++i) {
+        ucp_rkey_h self_rkeys[] = {perf->ucp.tctx[i].perf.ucp.self_send_rkey,
+                                   perf->ucp.tctx[i].perf.ucp.self_recv_rkey};
+        req = ucp_perf_test_destroy_ep(perf->ucp.tctx[i].perf.ucp.self_ep,
+                                       self_rkeys, 2, i);
+        if (req != NULL) {
+            reqs[num_in_prog++] = req;
+        }
+    }
+
+    ucp_perf_release_requests_in_progress(perf, reqs, num_in_prog);
+}
+
+static void ucp_perf_test_destroy_eps(ucx_perf_context_t *perf)
+{
+    unsigned thread_count   = perf->params.thread_count;
+    unsigned num_in_prog    = 0;
+    ucs_status_ptr_t **reqs = ucs_alloca(thread_count * sizeof(*reqs));
+    ucs_status_ptr_t *req;
+    unsigned i;
+
+    for (i = 0; i < thread_count; ++i) {
+        req = ucp_perf_test_destroy_ep(perf->ucp.tctx[i].perf.ucp.ep,
+                                       &perf->ucp.tctx[i].perf.ucp.rkey, 1, i);
+        if (req != NULL) {
+            reqs[num_in_prog++] = req;
+        }
+    }
+
+    ucp_perf_release_requests_in_progress(perf, reqs, num_in_prog);
 }
 
 static ucs_status_t
@@ -986,14 +1039,10 @@ static ucs_status_t ucp_perf_test_rkey_pack(ucx_perf_context_t *perf,
     ucs_status_t status;
 
     if (features & (UCP_FEATURE_RMA | UCP_FEATURE_AMO32 | UCP_FEATURE_AMO64)) {
-        status = ucp_rkey_pack(perf->ucp.context, perf->ucp.recv_memh,
-                               rkey_buffer, rkey_size);
+        status = UCX_PERF_VERBOSE(error, &perf->params, ucp_rkey_pack,
+                                  perf->ucp.context, perf->ucp.recv_memh,
+                                  rkey_buffer, rkey_size);
         if (status != UCS_OK) {
-            if (perf->params.flags & UCX_PERF_TEST_FLAG_VERBOSE) {
-                ucs_error("ucp_rkey_pack() failed: %s",
-                           ucs_status_string(status));
-            }
-
             return status;
         }
     } else {
@@ -1011,14 +1060,10 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
     ucs_status_t status;
 
     if (rkey_size > 0) {
-        status = ucp_ep_rkey_unpack(thread->perf.ucp.ep, rkey_buffer,
-                                    &thread->perf.ucp.rkey);
+        status = UCX_PERF_VERBOSE(fatal, params, ucp_ep_rkey_unpack,
+                                  thread->perf.ucp.ep, rkey_buffer,
+                                  &thread->perf.ucp.rkey);
         if (status != UCS_OK) {
-            if (params->flags & UCX_PERF_TEST_FLAG_VERBOSE) {
-                ucs_fatal("ucp_rkey_unpack() failed: %s",
-                           ucs_status_string(status));
-            }
-
             return status;
         }
     } else {
@@ -1026,6 +1071,28 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
     }
 
     return UCS_OK;
+}
+
+static ucs_status_t ucp_perf_test_create_self_rkey(ucx_perf_context_t *perf,
+                                                   ucp_mem_h memh,
+                                                   ucp_rkey_h *rkey)
+{
+    void *rkey_buffer                                      = NULL;
+    size_t rkey_size                                       = 0;
+    ucp_memh_buffer_release_params_t buffer_release_params = {0};
+    ucs_status_t status;
+
+    status = UCX_PERF_VERBOSE(fatal, &perf->params, ucp_rkey_pack,
+                              perf->ucp.context, memh, &rkey_buffer,
+                              &rkey_size);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    status = UCX_PERF_VERBOSE(fatal, &perf->params, ucp_ep_rkey_unpack,
+                              perf->ucp.self_ep, rkey_buffer, rkey);
+    ucp_memh_buffer_release(rkey_buffer, &buffer_release_params);
+    return status;
 }
 
 static ucs_status_t ucp_perf_test_receive_remote_data(ucx_perf_context_t *perf,
@@ -1053,11 +1120,14 @@ static ucs_status_t ucp_perf_test_receive_remote_data(ucx_perf_context_t *perf,
 
     /* Initialize all endpoints and rkeys to NULL to handle error flow */
     for (i = 0; i < thread_count; i++) {
-        perf->ucp.tctx[i].perf.ucp.ep   = NULL;
-        perf->ucp.tctx[i].perf.ucp.rkey = NULL;
+        perf->ucp.tctx[i].perf.ucp.ep             = NULL;
+        perf->ucp.tctx[i].perf.ucp.rkey           = NULL;
+        perf->ucp.tctx[i].perf.ucp.self_ep        = NULL;
+        perf->ucp.tctx[i].perf.ucp.self_send_rkey = NULL;
+        perf->ucp.tctx[i].perf.ucp.self_recv_rkey = NULL;
     }
 
-    /* receive the data from the remote peer, extract the address from it
+    /* Receive the data from the remote peer, extract the address from it
      * (along with additional wireup info) and create an endpoint to the peer */
     rte_call(perf, recv, peer_index, buffer, buffer_size, req);
 
@@ -1079,12 +1149,10 @@ static ucs_status_t ucp_perf_test_receive_remote_data(ucx_perf_context_t *perf,
             ep_params.err_mode        = UCP_ERR_HANDLING_MODE_PEER;
         }
 
-        status = ucp_ep_create(perf->ucp.tctx[i].perf.ucp.worker, &ep_params,
-                               &perf->ucp.tctx[i].perf.ucp.ep);
+        status = UCX_PERF_VERBOSE(error, &perf->params, ucp_ep_create,
+                                  perf->ucp.tctx[i].perf.ucp.worker, &ep_params,
+                                  &perf->ucp.tctx[i].perf.ucp.ep);
         if (status != UCS_OK) {
-            if (perf->params.flags & UCX_PERF_TEST_FLAG_VERBOSE) {
-                ucs_error("ucp_ep_create() failed: %s", ucs_status_string(status));
-            }
             goto err_free_eps_buffer;
         }
 
@@ -1138,13 +1206,10 @@ static ucs_status_t ucp_perf_test_send_local_data(ucx_perf_context_t *perf,
     /* get the worker address created for every thread and send it to the remote
      * peer */
     for (i = 0; i < thread_count; i++) {
-        status = ucp_worker_get_address(perf->ucp.tctx[i].perf.ucp.worker,
-                                        &address, &address_length);
+        status = UCX_PERF_VERBOSE(error, &perf->params, ucp_worker_get_address,
+                                  perf->ucp.tctx[i].perf.ucp.worker, &address,
+                                  &address_length);
         if (status != UCS_OK) {
-            if (perf->params.flags & UCX_PERF_TEST_FLAG_VERBOSE) {
-                ucs_error("ucp_worker_get_address() failed: %s",
-                          ucs_status_string(status));
-            }
             goto err_free_workers_vec;
         }
 
@@ -1257,6 +1322,62 @@ static ucs_status_t ucp_perf_test_flush_workers(ucx_perf_context_t *perf)
     return ctx.status;
 }
 
+static ucs_status_t ucp_perf_test_setup_self_endpoints(ucx_perf_context_t *perf)
+{
+    ucp_worker_h worker = perf->ucp.tctx[0].perf.ucp.worker;
+    ucs_status_t status;
+    ucp_ep_params_t ep_params;
+    ucx_perf_context_t *thread_perf;
+    ucp_worker_attr_t worker_attr = {
+        .field_mask = UCP_WORKER_ATTR_FIELD_ADDRESS
+    };
+    unsigned i;
+
+    status = UCX_PERF_VERBOSE(error, &perf->params, ucp_worker_query, worker,
+                              &worker_attr);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    ep_params.field_mask = UCP_EP_PARAM_FIELD_REMOTE_ADDRESS;
+    ep_params.address    = worker_attr.address;
+
+    for (i = 0; i < perf->params.thread_count; ++i) {
+        thread_perf = &perf->ucp.tctx[i].perf;
+        status = UCX_PERF_VERBOSE(error, &perf->params, ucp_ep_create, worker,
+                                  &ep_params, &thread_perf->ucp.self_ep);
+        if (status != UCS_OK) {
+            goto err_destroy_eps;
+        }
+
+        status = ucp_perf_test_create_self_rkey(thread_perf, thread_perf->ucp.send_memh,
+                                                &thread_perf->ucp.self_send_rkey);
+        if (status != UCS_OK) {
+            goto err_destroy_eps;
+        }
+
+        status = ucp_perf_test_create_self_rkey(thread_perf, thread_perf->ucp.recv_memh,
+                                                &thread_perf->ucp.self_recv_rkey);
+        if (status != UCS_OK) {
+            goto err_destroy_eps;
+        }
+    }
+
+    /* Global default ep will point to the ep of the 1st thread */
+    perf->ucp.self_ep        = perf->ucp.tctx[0].perf.ucp.self_ep;
+    perf->ucp.self_send_rkey = perf->ucp.tctx[0].perf.ucp.self_send_rkey;
+    perf->ucp.self_recv_rkey = perf->ucp.tctx[0].perf.ucp.self_recv_rkey;
+
+    status = UCS_OK;
+    goto out;
+
+err_destroy_eps:
+    ucp_perf_test_destroy_self_eps(perf);
+out:
+    ucp_worker_release_address(worker, worker_attr.address);
+    return status;
+}
+
 static ucs_status_t ucp_perf_test_setup_endpoints(ucx_perf_context_t *perf,
                                                   uint64_t features)
 {
@@ -1279,25 +1400,33 @@ static ucs_status_t ucp_perf_test_setup_endpoints(ucx_perf_context_t *perf,
         return UCS_ERR_UNSUPPORTED;
     }
 
-    /* pack the local endpoints data and send to the remote peer */
+    /* Pack the local endpoints data and send to the remote peer */
     status = ucp_perf_test_send_local_data(perf, features);
     if (status != UCS_OK) {
         goto err;
     }
 
-    /* receive remote peer's endpoints' data and connect to them */
+    /* Receive remote peer's endpoints' data and connect to them */
     status = ucp_perf_test_receive_remote_data(perf, peer_index);
     if (status != UCS_OK) {
         goto err;
     }
 
-    /* sync status across all processes */
+    /* Sync status across all processes */
     status = ucx_perf_test_exchange_status(perf, UCS_OK);
     if (status != UCS_OK) {
         goto err_destroy_eps;
     }
 
-    /* force wireup completion */
+    /* Create self_ep used for copying into non-host memory types*/
+    if (!ucp_perf_test_is_host_mem_only(&perf->params)) {
+        status = ucp_perf_test_setup_self_endpoints(perf);
+        if (status != UCS_OK) {
+            goto err_destroy_eps;
+        }
+    }
+
+    /* Force wireup completion */
     return ucp_perf_test_flush_workers(perf);
 
 err_destroy_eps:
@@ -1310,6 +1439,7 @@ err:
 static void ucp_perf_test_cleanup_endpoints(ucx_perf_context_t *perf)
 {
     ucp_perf_barrier(perf);
+    ucp_perf_test_destroy_self_eps(perf);
     ucp_perf_test_destroy_eps(perf);
 }
 
@@ -1684,11 +1814,10 @@ static ucs_status_t ucp_perf_setup(ucx_perf_context_t *perf)
         }
     }
 
-    status = ucp_perf_test_setup_endpoints(perf, ucp_params.features);
+    status = UCX_PERF_VERBOSE(error, &perf->params,
+                              ucp_perf_test_setup_endpoints, perf,
+                              ucp_params.features);
     if (status != UCS_OK) {
-        if (perf->params.flags & UCX_PERF_TEST_FLAG_VERBOSE) {
-            ucs_error("Failed to setup endpoints: %s", ucs_status_string(status));
-        }
         goto err_destroy_workers;
     }
 
@@ -1723,6 +1852,46 @@ ucx_perf_funcs_t ucx_perf_funcs[] = {
                           ucp_perf_test_dispatch, ucp_perf_thread_barrier}
 };
 
+ucs_status_t ucx_perf_allocators_init(ucx_perf_context_t *perf,
+                                      const ucx_perf_params_t *params)
+{
+    if ((params->send_mem_type == UCS_MEMORY_TYPE_RDMA) ||
+        (params->recv_mem_type == UCS_MEMORY_TYPE_RDMA)) {
+        return UCS_OK;
+    }
+
+    ucs_debug("set send allocator by send mem type %s",
+              ucs_memory_type_names[params->send_mem_type]);
+    perf->send_allocator = ucx_perf_mem_type_allocators[params->send_mem_type];
+
+    ucs_debug("set recv allocator by recv mem type %s",
+              ucs_memory_type_names[params->recv_mem_type]);
+    perf->recv_allocator = ucx_perf_mem_type_allocators[params->recv_mem_type];
+
+    if ((perf->send_allocator == NULL) || (perf->recv_allocator == NULL)) {
+        ucs_error("Unsupported memory types %s<->%s",
+                  ucs_memory_type_names[params->send_mem_type],
+                  ucs_memory_type_names[params->recv_mem_type]);
+        return UCS_ERR_UNSUPPORTED;
+    }
+
+    if (perf->send_allocator->mem_type != UCS_MEMORY_TYPE_HOST) {
+        ucs_diag("UCT tests also copy one-byte value from %s memory to "
+                 "%s send memory, which may impact performance results",
+                 ucs_memory_type_names[UCS_MEMORY_TYPE_HOST],
+                 ucs_memory_type_names[perf->send_allocator->mem_type]);
+    }
+
+    if (perf->recv_allocator->mem_type != UCS_MEMORY_TYPE_HOST) {
+        ucs_diag("UCT tests also copy one-byte value from %s recv memory "
+                 "to %s memory, which may impact performance results",
+                 ucs_memory_type_names[perf->recv_allocator->mem_type],
+                 ucs_memory_type_names[UCS_MEMORY_TYPE_HOST]);
+    }
+
+    return ucx_perf_allocators_init_thread(perf);
+}
+
 ucs_status_t ucx_perf_run(const ucx_perf_params_t *params,
                           ucx_perf_result_t *result)
 {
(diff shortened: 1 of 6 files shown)


gleon99:
@roiedanino ,
seems many mutual changes with https://github.com/openucx/ucx/pull/9189/files.
How do they work together, which one should be merged first?
Also - there are some changes in UCT flow (`ucx_perf_allocators_init`) in this PR and not the UCT one. Is that intentional?

src/tools/perf/lib/Makefile.am
```diff
@@ -18,6 +18,24 @@ libucxperf_la_LIBADD   = \
...
+# Add ROCM dependencies
+if HAVE_HIP
+libucxperf_la_CPPFLAGS += $(HIP_CPPFLAGS)
+libucxperf_la_CFLAGS   += $(HIP_CFLAGS)
+libucxperf_la_LDFLAGS  += $(HIP_LDFLAGS) $(HIP_LIBS) -version-info $(SOVERSION) \
+                          $(patsubst %, -Xlinker %, -L$(ROCM_ROOT)/lib -rpath $(ROCM_ROOT)/hip/lib -rpath $(ROCM_ROOT)/lib) \
+                          $(patsubst %, -Xlinker %, --enable-new-dtags) \
+                          $(patsubst %, -Xlinker %, -rpath $(ROCM_ROOT)/lib64)
+endif
+
```
yosefe, line 38: we cannot add direct dependency on cuda/rocm because then if UCX is built with cuda support but used on a system without cuda libs, it will not work

src/tools/perf/lib/libperf_thread.c
```diff
@@ -30,17 +30,19 @@ static ucs_status_t ucx_perf_thread_run_test(void* arg)
...
 
-    /* new threads need explicit device association */
-    status = perf->send_allocator->init(perf);
-    if (status != UCS_OK) {
-        goto out;
-    }
-
-    if (perf->send_allocator != perf->recv_allocator) {
-        status = perf->recv_allocator->init(perf);
+    if (params->api == UCX_PERF_API_UCT) {
```
gleon99, line 33: In the UCT PR you deleted this flow, but here it is under the condition of `UCT` (?)

src/tools/perf/perftest_params.c
```diff
@@ -16,12 +16,34 @@
...
+    }
+}
+
 static void print_memory_type_usage(void)
 {
     ucs_memory_type_t it;
 
     ucs_memory_type_for_each(it) {
-        if (ucx_perf_mem_type_allocators[it] != NULL) {
+        if (check_mem_type_support(it)) {
```
gleon99, line 46: This func is totally different in the UCT PR. How is that going to work?

roiedanino:
> @roiedanino , seems many mutual changes with https://github.com/openucx/ucx/pull/9189/files. How do they work together, which one should be merged first?

Yes, this is intentional, Making UCT support any memory type is currently on hold because changing memset is problematic as the correct way to do it is put_short but we need to check that it is supported by the selected transport layer.
So @yosefe suggested that I'll open a separate PR updating only UCP layer `perftest` and keep the other one on hold

> Also - there are some changes in UCT flow (`ucx_perf_allocators_init`) in this PR and not the UCT one. Is that intentional?
Yes, the changes are needed so UCT will go through allocators init() and UCP won't

src/tools/perf/perftest_params.c
roiedanino: I think this approach is more efficient than what I did in the other PR

gleon99: But in such case need to change it in the other PR as well, otherwise we'll merge temporary code and then have merge conflicts.

src/tools/perf/lib/Makefile.am
roiedanino: So `libperf_memcopy.c` should be a module?

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -370,29 +370,27 @@ class ucp_perf_test_runner {
...
         const void *ptr            = sn_ptr(buffer, length);
+        ucp_request_param_t param  = {0};
         psn_t sn;
 
         if (mem_type == UCS_MEMORY_TYPE_HOST) {
             return *(const volatile psn_t*)ptr;
         } else {
-            m_perf.recv_allocator->memcpy(&sn, UCS_MEMORY_TYPE_HOST, ptr,
-                                          mem_type, sizeof(sn));
+            ucp_get_nbx(m_perf.ucp.ep, &sn, sizeof(sn), (uint64_t)ptr, m_perf.ucp.rkey, &param);
```
yosefe, line 379: need to wait for completion

src/tools/perf/perftest_params.c
yosefe: if something is complited it's not good enough, it also has to be supported in runtime.
we can fail if ucp_mem_map fails to allocate the given memory type.

src/tools/perf/lib/libperf.c
```diff
@@ -1749,42 +1790,14 @@ ucs_status_t ucx_perf_run(const ucx_perf_params_t *params,
...
-
-    status = perf->send_allocator->init(perf);
-    if (status != UCS_OK) {
-        goto out_free;
-    }
-
-    if (perf->send_allocator != perf->recv_allocator) {
-        status = perf->recv_allocator->init(perf);
+        ucx_perf_global_init();
+        status = ucx_perf_allocators_init(perf, params);
```
yosefe, line 1795: why do we need this change? we're not supposed to change UCT flow..

roiedanino: I didn't change UCT flow but made some of the flow exclusive to UCT

src/tools/perf/lib/ucp_tests.cc
roiedanino: can I use ucp_get and ucp_put instead?

src/tools/perf/perftest_params.c
roiedanino: Okay, I changed to a solution similar to what I did in #9189

src/tools/perf/lib/libperf.c
yosefe: is this change really necessary? what happens if you don't change it?

src/tools/perf/lib/ucp_tests.cc
yosefe: no since they are deprecated

src/tools/perf/lib/libperf.c
roiedanino: UCP will try to call `perf->send_allocator->init()` and will depend on the existence of allocators

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -366,33 +366,74 @@ class ucp_perf_test_runner {
...
+
+        do {
+            ret    = ucp_worker_progress(worker);
+            status = ucp_request_check_status(request);
+            if (status != UCS_INPROGRESS) {
+                ucp_request_free(request);
+                if (ucs_unlikely(UCS_OK != status)) {
+                    return status;
+                }
+                break;
```
roiedanino, line 383: break is redundant

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -366,31 +366,68 @@ class ucp_perf_test_runner {
...
         if (mem_type == UCS_MEMORY_TYPE_HOST) {
             return *(const volatile psn_t*)ptr;
         } else {
-            ucp_get_nbx(m_perf.ucp.ep, &sn, sizeof(sn), (uint64_t)ptr, m_perf.ucp.rkey, &param);
+            request = ucp_get_nbx(m_perf.ucp.ep, &sn, sizeof(sn), (uint64_t)ptr,
+                                  m_perf.ucp.rkey, &param);
+            status  = ucp_perftest_wait_for_completion(request,
+                                                       m_perf.ucp.worker);
+            if (status != UCS_OK) {
+                ucs_warn("%s: Failed to read sn from memory type: %s",
```
yosefe, line 401: can we add the warning message inside ucp_perftest_wait_for_completion?

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -366,31 +366,68 @@ class ucp_perf_test_runner {
         return UCS_PTR_BYTE_OFFSET(buffer, length - sizeof(psn_t));
     }
 
+    UCS_F_ALWAYS_INLINE ucs_status_t ucp_perftest_wait_for_completion(
+            ucs_status_ptr_t request, ucp_worker_h worker)
+    {
+        ucs_status_t status;
```
yosefe, line 372: first, check if the request is NULL or error. See other places that wait for a request.

yosefe (changes requested):

src/tools/perf/perftest_params.c
```diff
@@ -10,40 +10,20 @@
 
 #include "perftest.h"
 
+#include <ucp/core/ucp_mm.h>
```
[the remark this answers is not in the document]
yosefe: there is no good place to share this function.
To reduce code duplication, try to avoid calling ucp_init and use existing context in each place

src/tools/perf/lib/libperf.c
```diff
@@ -1723,13 +1722,55 @@ ucx_perf_funcs_t ucx_perf_funcs[] = {
...
+                 ucs_memory_type_names[perf->recv_allocator->mem_type],
+                 ucs_memory_type_names[UCS_MEMORY_TYPE_HOST]);
+    }
+
+    status = perf->send_allocator->init(perf);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    if (perf->send_allocator != perf->recv_allocator) {
```
gleon99, line 1758: Instead of
```
    if (perf->send_allocator != perf->recv_allocator) {
        status = perf->recv_allocator->init(perf);
        if (status != UCS_OK) {
            return status;
        }
    }

    return UCS_OK;
```

can have:
```
if (perf->send_allocator == perf->recv_allocator) {
    return UCS_OK;
}

return perf->recv_allocator->init(perf);
```
WDYT?

src/tools/perf/perftest_params.c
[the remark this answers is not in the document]
roiedanino: Those functions are being called from the `usage()` printing function being called before context creation

src/tools/perf/lib/libperf.c
roiedanino: Will change 👍

src/tools/perf/lib/libperf.c
```diff
@@ -1749,42 +1787,14 @@ ucs_status_t ucx_perf_run(const ucx_perf_params_t *params,
...
-    }
-
-    status = perf->send_allocator->init(perf);
-    if (status != UCS_OK) {
-        goto out_free;
-    }
-
-    if (perf->send_allocator != perf->recv_allocator) {
-        status = perf->recv_allocator->init(perf);
+        ucx_perf_global_init();
```
yosefe, line 1791: pls move this line back to its previous location

src/tools/perf/lib/libperf_thread.c
```diff
@@ -30,17 +30,19 @@ static ucs_status_t ucx_perf_thread_run_test(void* arg)
...
         if (status != UCS_OK) {
             goto out;
         }
+
+        if (perf->send_allocator != perf->recv_allocator) {
+            status = perf->recv_allocator->init(perf);
+            if (status != UCS_OK) {
+                goto out;
+            }
+        }
```
yosefe, line 45: can we extract common code with ucx_perf_allocators_init()?
the common function could be called ucx_perf_allocators_init_thread()

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -366,33 +366,70 @@ class ucp_perf_test_runner {
...
+    ucp_perftest_wait_for_completion(ucs_status_ptr_t request,
+                                     ucp_worker_h worker,
+                                     ucs_memory_type_t mem_type,
+                                     const char *desc)
+    {
+        ucs_status_t status;
+
+        if (request == NULL) {
+            return;
+        }
```
yosefe, line 379: this can be removed

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -366,33 +366,70 @@ class ucp_perf_test_runner {
...
+                status = ucp_request_check_status(request);
+            } while (status == UCS_INPROGRESS);
+
+            ucp_request_free(request);
+        } else {
+            status = UCS_PTR_STATUS(request);
+        }
+
+        if (status != UCS_OK) {
+            ucs_warn("%s: Failed to %s, memory_type: %s",
```
yosefe, line 393:
```c
ucs_warn("failed to %s(memory_type=%s): %s", operation_name, 
         ucs_memory_type_names[mem_type], ucs_status_string(status));
```

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -366,33 +366,70 @@ class ucp_perf_test_runner {
         return UCS_PTR_BYTE_OFFSET(buffer, length - sizeof(psn_t));
     }
 
+    UCS_F_ALWAYS_INLINE void
+    ucp_perftest_wait_for_completion(ucs_status_ptr_t request,
```
yosefe, line 370: this is a member function. just call it request_wait

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -366,33 +366,70 @@ class ucp_perf_test_runner {
...
 
         if (mem_type == UCS_MEMORY_TYPE_HOST) {
             *(volatile psn_t*)ptr = sn;
         } else {
-            allocator->memcpy(ptr, mem_type, &sn, UCS_MEMORY_TYPE_HOST,
-                              sizeof(sn));
+            request = ucp_put_nbx(m_perf.ucp.ep, &sn, sizeof(sn), (uint64_t)ptr,
+                                  m_perf.ucp.rkey, &param);
+
+            ucp_perftest_wait_for_completion(request, m_perf.ucp.worker,
```
yosefe, line 431: why need to pass worker? it's already a member

yosefe (changes requested):

src/tools/perf/lib/libperf.c
```diff
@@ -191,6 +191,7 @@ void ucx_perf_test_prepare_new_run(ucx_perf_context_t *perf,
 {
     unsigned i;
 
+    perf->params            = *params;
```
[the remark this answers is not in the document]
roiedanino: I wanted this line to be in both UCT and UCP flows, I'll move it to ucx_perf_run

src/tools/perf/lib/libperf.c
roiedanino: Can I move it into `ucx_perf_allocators_init`?

yosefe: IMO previous location is better: 1. no code change, 2. global_init could in theory initialize other things (because of its name)

yosefe (dismissed):

roiedanino:
@yosefe Can I squash?

yosefe:
> @yosefe Can I squash?

Yes

yosefe (approved):

src/tools/perf/lib/libperf.c
```diff
@@ -894,7 +894,8 @@ static void ucp_perf_test_destroy_eps(ucx_perf_context_t* perf)
 {
     unsigned i, thread_count = perf->params.thread_count;
     unsigned num_in_prog     = 0;
-    ucs_status_ptr_t **reqs  = ucs_alloca(thread_count * sizeof(*reqs));
+    ucs_status_ptr_t **reqs  = ucs_alloca(2 * /* (ep + self_ep) */
```
yosefe, line 897: why need to allocate for self_ep?
there should be only one request for self_ep, so it could be a single local var

src/tools/perf/lib/libperf.c
```diff
@@ -915,6 +920,19 @@ static void ucp_perf_test_destroy_eps(ucx_perf_context_t* perf)
                          ucs_status_string(UCS_PTR_STATUS(req)));
             }
         }
+
+        if (perf->ucp.tctx[i].perf.ucp.self_ep != NULL) {
```
yosefe, line 924: extract common code with lines 911-923

src/tools/perf/lib/libperf.c
```diff
@@ -1012,6 +1030,64 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+    status = ucp_worker_get_address(worker, &address, &address_length);
+    if (status != UCS_OK) {
+        ucs_error("ucp_worker_get_address() failed: %s",
+                  ucs_status_string(status));
+        return status;
+    }
+    ep_params.field_mask = UCP_EP_PARAM_FIELD_REMOTE_ADDRESS;
+    ep_params.address    = address;
+
+    if (perf->params.flags & UCX_PERF_TEST_FLAG_ERR_HANDLING) {
```
yosefe, line 1052: no need for err handling on self ep

src/tools/perf/lib/libperf.c
```diff
@@ -1012,6 +1030,64 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+        ep_params.field_mask     |= UCP_EP_PARAM_FIELD_ERR_HANDLER |
+                                    UCP_EP_PARAM_FIELD_ERR_HANDLING_MODE;
+        ep_params.err_handler.cb  = ucp_perf_test_err_handler;
+        ep_params.err_handler.arg = NULL;
+        ep_params.err_mode        = UCP_ERR_HANDLING_MODE_PEER;
+    }
+
+    status = ucp_ep_create(worker, &ep_params, &perf->ucp.self_ep);
+
```
yosefe, line 1061: remove space line

src/tools/perf/lib/libperf.c
```diff
@@ -1012,6 +1030,64 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+    if (status != UCS_OK) {
+        if (perf->params.flags & UCX_PERF_TEST_FLAG_VERBOSE) {
+            ucs_error("ucp_ep_create() failed: %s", ucs_status_string(status));
+        }
+        return status;
+    }
+    status = ucp_rkey_pack(perf->ucp.context, perf->ucp.recv_memh, &rkey_buffer,
```
yosefe, line 1068: 1. space line before, not after
2. why self_rkey is for recv_memh - what happens if we need to copy to send buffer?

src/tools/perf/lib/libperf.c
```diff
@@ -1012,6 +1030,64 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+                           &rkey_size);
+
+    if (status != UCS_OK) {
+        if (perf->params.flags & UCX_PERF_TEST_FLAG_VERBOSE) {
+            ucs_fatal("ucp_rkey_pack() failed: %s", ucs_status_string(status));
+        }
+
+        return status;
+    }
+    status = ucp_ep_rkey_unpack(perf->ucp.self_ep, rkey_buffer,
```
yosefe, line 1078: space line before

src/tools/perf/lib/libperf.c
```diff
@@ -1012,6 +1030,64 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+                                &perf->ucp.self_rkey);
+    if (status != UCS_OK) {
+        if (perf->params.flags & UCX_PERF_TEST_FLAG_VERBOSE) {
+            ucs_fatal("ucp_rkey_unpack() failed: %s",
+                      ucs_status_string(status));
+        }
+
+        return status;
+    }
+    return UCS_OK;
```
yosefe, line 1088: space line before

src/tools/perf/lib/libperf.c
roiedanino: We need a self_ep for each thread, don't we? As far as I understand m_perf is thread-local

yosefe: self_ep - yes, request - no

src/tools/perf/lib/libperf.c
```diff
@@ -890,48 +890,49 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
...
+{
+    ucs_status_ptr_t *req;
+
+    if (rkey != NULL) {
+        ucp_rkey_destroy(rkey);
+    }
+
+    if (ep != NULL) {
+        req = ucp_ep_close_nb(ep, UCP_EP_CLOSE_MODE_FLUSH);
+
```
yosefe, line 904: remove space line

src/tools/perf/lib/libperf.c
```diff
@@ -890,48 +890,49 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
...
+        if (UCS_PTR_IS_PTR(req)) {
+            return req;
+        } else if (UCS_PTR_STATUS(req) != UCS_OK) {
+            ucs_warn("failed to close ep %p on thread %d: %s\n", ep, index,
+                     ucs_status_string(UCS_PTR_STATUS(req)));
+        }
+    }
+    return NULL;
```
yosefe, line 912: space line before

src/tools/perf/lib/libperf.c
```diff
@@ -890,48 +890,49 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
...
+}
+
 static void ucp_perf_test_destroy_eps(ucx_perf_context_t* perf)
 {
     unsigned i, thread_count = perf->params.thread_count;
     unsigned num_in_prog     = 0;
-    ucs_status_ptr_t **reqs  = ucs_alloca(2 * /* (ep + self_ep) */
-                                          thread_count * sizeof(*reqs));
+    ucs_status_ptr_t **reqs  = ucs_alloca(thread_count * sizeof(*reqs));
```
yosefe, line 919: I was mistaken, we do need a request also for self_ep. missed that it's a destroy flow and it's a close request.
so need to do `reqs[num_in_prog++] = req;` for it instead of releasing it.

src/tools/perf/lib/libperf.c
```diff
@@ -906,31 +890,53 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
...
-            } else if (UCS_PTR_STATUS(req) != UCS_OK) {
-                ucs_warn("failed to close ep %p on thread %d: %s\n",
-                         perf->ucp.tctx[i].perf.ucp.ep, i,
-                         ucs_status_string(UCS_PTR_STATUS(req)));
-            }
+        if (self_req != NULL) {
+            ucp_request_release(self_req);
         }
+
+        ucp_rkey_destroy(perf->ucp.tctx[i].perf.ucp.self_recv_rkey);
```
yosefe, line 939: 1. do it before destroying the ep.
2. could it be NULL?
3. maybe pass the keys that should be destroyed to ucp_perf_test_destroy_ep - as "const ucp_rkey_h *rkeys, unsigned num_rkeys"

src/tools/perf/lib/libperf.c
```diff
@@ -1028,6 +1034,71 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
 
+static ucs_status_t ucp_perf_test_create_self_rkey(ucx_perf_context_t *perf,
+                                                   ucp_mem_h memh,
+                                                   ucp_rkey_h *rkey)
+{
+    void *rkey_buffer = NULL;
+    size_t rkey_size  = 0;
+    ucs_status_t status;
+
+    status = ucp_rkey_pack(perf->ucp.context, memh, //perf->ucp.recv_memh,
```
yosefe, line 1045: remove comment

src/tools/perf/lib/libperf.c
```diff
@@ -1028,6 +1034,71 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+        if (perf->params.flags & UCX_PERF_TEST_FLAG_VERBOSE) {
+            ucs_error("ucp_ep_create() failed: %s", ucs_status_string(status));
+        }
+        return status;
+    }
+
+    status = ucp_perf_test_create_self_rkey(perf, perf->ucp.send_memh,
+                                            &perf->ucp.self_send_rkey);
+    if (status != UCS_OK) {
+        return status;
```
yosefe, line 1095: need cleanup flow if some stage fails (with goto)

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -366,33 +366,62 @@ class ucp_perf_test_runner {
         return UCS_PTR_BYTE_OFFSET(buffer, length - sizeof(psn_t));
     }
 
+    UCS_F_ALWAYS_INLINE void request_wait(ucs_status_ptr_t request,
```
yosefe, line 369: does it need to be inline?

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -649,10 +675,8 @@ class ucp_perf_test_runner {
             return;
         }
 
-        write_sn(m_perf.send_buffer, m_perf.params.send_mem_type, length, sn,
-                 m_perf.send_allocator);
-        write_sn(m_perf.recv_buffer, m_perf.params.recv_mem_type, length, sn,
-                 m_perf.recv_allocator);
+        write_sn(m_perf.send_buffer, m_perf.params.send_mem_type, length, sn);
+        write_sn(m_perf.recv_buffer, m_perf.params.recv_mem_type, length, sn);
```
yosefe, line 679: do we need to use self_recv_memh here?

src/tools/perf/lib/libperf.c
roiedanino: But the calling function will do a cleanup for any `status != UCS_OK`:
image

yosefe: nevertheless, each function should cleanup after itself if it fails

src/tools/perf/lib/ucp_tests.cc
roiedanino: Why?

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -419,12 +419,17 @@ class ucp_perf_test_runner {
...
+                                  (uint64_t)ptr, rkey, &param);
             request_wait(request, mem_type, "write_sn");
         }
     }
 
+    UCS_F_ALWAYS_INLINE void write_sn(void *buffer, ucs_memory_type_t mem_type,
+                                      size_t length, psn_t sn)
+    {
+        write_sn_ext(buffer, mem_type, length, sn, m_perf.ucp.self_send_rkey);
+    }
```
yosefe, line 431: why not always pass rkey to write_sn? IMO this wrapper is redundant..

src/tools/perf/lib/libperf.c
```diff
@@ -1028,6 +1035,79 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+    ucs_status_t status;
+    ucp_ep_params_t ep_params;
+
+    status = ucp_worker_get_address(worker, &address, &address_length);
+    if (status != UCS_OK) {
+        ucs_error("ucp_worker_get_address() failed: %s",
+                  ucs_status_string(status));
+        return status;
+    }
+    ep_params.field_mask = UCP_EP_PARAM_FIELD_REMOTE_ADDRESS;
```
yosefe, line 1081: space line before

src/tools/perf/lib/libperf.c
```diff
@@ -1028,6 +1035,79 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+
+    status = ucp_perf_test_create_self_rkey(perf, perf->ucp.send_memh,
+                                            &perf->ucp.self_send_rkey);
+    if (status != UCS_OK) {
+        goto err_destroy_ep;
+    }
+
+    status = ucp_perf_test_create_self_rkey(perf, perf->ucp.recv_memh,
+                                            &perf->ucp.self_recv_rkey);
+    if (status == UCS_OK) {
```
yosefe, line 1100: if (status != UCS_OK) {
    goto err_destroy_rkeys;
}

return UCS_OK;

src/tools/perf/lib/libperf.c
```diff
@@ -1028,6 +1035,79 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+        return UCS_OK;
+    }
+
+err_destroy_ep:
+    ucp_perf_test_destroy_eps(perf);
```
yosefe, line 1105: destroy only self ep

src/tools/perf/lib/libperf.c
```diff
@@ -1028,6 +1035,79 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+err_release_address:
+    ucp_worker_release_address(worker, address);
```
yosefe, line 1107: IMO we need to release the address also in case of success

src/tools/perf/lib/ucp_tests.cc
roiedanino: because in most cases it will be self_send_rkey, I'll remove the wrapper anyway

src/tools/perf/lib/libperf.c
```diff
@@ -1064,7 +1064,7 @@ static ucs_status_t ucp_perf_test_create_self_rkey(ucx_perf_context_t *perf,
     return status;
 }
 
-static ucs_status_t ucp_perf_test_setup_self_endpoint(ucx_perf_context_t *perf)
+static ucs_status_t ucp_perf_test_setup_self_endpoint(ucx_perf_context_t *perf, int thread_index)
```
yosefe, line 1067: line seems too long

src/tools/perf/lib/libperf.c
```diff
@@ -1097,12 +1098,17 @@ static ucs_status_t ucp_perf_test_setup_self_endpoint(ucx_perf_context_t *perf)
 
     status = ucp_perf_test_create_self_rkey(perf, perf->ucp.recv_memh,
                                             &perf->ucp.self_recv_rkey);
-    if (status == UCS_OK) {
-        return UCS_OK;
+    if (status != UCS_OK) {
+        goto err_destroy_rkey;
```
yosefe, line 1102: pls rename label to err_destroy_send_rkey

yosefe (commented):
pls squash,
in next PR let's make it run in CI when device memory is available

yosefe (approved):

src/tools/perf/lib/libperf.c
```diff
@@ -1257,6 +1347,29 @@ static ucs_status_t ucp_perf_test_flush_workers(ucx_perf_context_t *perf)
     return ctx.status;
 }
 
+static ucs_status_t ucp_perf_test_setup_self_endpoints(ucx_perf_context_t *perf)
+{
+    int i;
+    ucs_status_t status;
+
+    for (i = 0; i < perf->params.thread_count; ++i) {
+        status = ucp_perf_test_setup_self_endpoint(&perf->ucp.tctx[i].perf, i);
```
yosefe, line 1356: we can unite ucp_perf_test_setup_self_endpoint into ucp_perf_test_setup_self_endpoints - since need to get worker address only once, and ep_params are the same, it can reduce overall amount of code

src/tools/perf/lib/libperf.c
```diff
@@ -1257,6 +1347,29 @@ static ucs_status_t ucp_perf_test_flush_workers(ucx_perf_context_t *perf)
...
+        if (status != UCS_OK) {
+            goto err_destroy_eps;
+        }
+    }
+
+    /* global default ep will point to the ep of the 1st thread */
```
yosefe, line 1362: begin comments with capital case (multiple locations)

src/tools/perf/lib/libperf.c
yosefe: BTW, maybe we can create self endpoints only if we have non-host memory?

roiedanino: Can I use worker and address of tctx[0] for all of the eps?

src/ucp/wireup/select.c
```diff
@@ -1203,10 +1204,21 @@ ucp_wireup_add_amo_lanes(const ucp_wireup_select_params_t *select_params,
...
+                select_params, ep_init_flags, &criteria, mem_type, tl_bitmap,
+                UCP_LANE_TYPE_AMO, select_ctx);
+        if ((status != UCS_OK) && (mem_type == UCS_MEMORY_TYPE_HOST)) {
+            return status;
+        }
+    }
+
+    return UCS_OK;
+
+    // return ucp_wireup_add_memaccess_lanes(select_params, ep_init_flags,
```
roiedanino, line 1218: delete

yosefe:
@rakhmets can you pls review?

src/ucp/wireup/select.c
```diff
@@ -1203,10 +1204,16 @@ ucp_wireup_add_amo_lanes(const ucp_wireup_select_params_t *select_params,
...
+    ucs_memory_type_for_each(mem_type) {
+        ucs_status_t status = ucp_wireup_add_memaccess_lanes(
+                select_params, ep_init_flags, &criteria, mem_type, tl_bitmap,
+                UCP_LANE_TYPE_AMO, select_ctx);
+        if ((status != UCS_OK) && (mem_type == UCS_MEMORY_TYPE_HOST)) {
+            return status;
+        }
+    }
+
+    return UCS_OK;
```
rakhmets, line 1216: 1. Could you please explain the use case covered by this code.
2. Does it fix a specific issue? If so, should it be added as a separate PR with a test?
3. Can it cause a wire compatibility issue?

rakhmets, line 1216: This is a code duplication with `ucp_wireup_add_rma_lanes`. Please make it as a separate method.

src/ucp/wireup/select.c
```diff
@@ -1203,10 +1204,16 @@ ucp_wireup_add_amo_lanes(const ucp_wireup_select_params_t *select_params,
...
-    return ucp_wireup_add_memaccess_lanes(select_params, ep_init_flags,
-                                          &criteria, UCS_MEMORY_TYPE_HOST,
-                                          tl_bitmap, UCP_LANE_TYPE_AMO,
-                                          select_ctx);
+    ucs_memory_type_for_each(mem_type) {
+        ucs_status_t status = ucp_wireup_add_memaccess_lanes(
+                select_params, ep_init_flags, &criteria, mem_type, tl_bitmap,
+                UCP_LANE_TYPE_AMO, select_ctx);
+        if ((status != UCS_OK) && (mem_type == UCS_MEMORY_TYPE_HOST)) {
+            return status;
```
rakhmets, line 1212: I think it is better to add a debug message to the log when `(status != UCS_OK) && (mem_type != UCS_MEMORY_TYPE_HOST)`.

src/ucp/wireup/select.c
roiedanino: Sorry, those changes were unnecessary, I accidentally pushed experimental code not solving any issue currently

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1386,15 +1399,20 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
         return;
     }
 
+    devx_memh = ucs_derived_of(memh, uct_ib_mlx5_devx_mem_t);
+    status    = uct_ib_mlx5_devx_reg_atomic_key(md, devx_memh);
+    if (status == UCS_OK) {
+        /* Enable device memory only if atomics are available*/
+        md->super.cap_flags |= UCT_MD_FLAG_ALLOC;
+        uct_ib_mlx5_devx_obj_destroy(devx_memh->atomic_dvmr, "DM-ATOMIC-KSM");
```
roiedanino, line 1407: I need to delete this line too

src/tools/perf/lib/libperf.c
```diff
@@ -1028,6 +1041,35 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
 
+static ucs_status_t ucp_perf_test_create_self_rkey(ucx_perf_context_t *perf,
+                                                   ucp_mem_h memh,
+                                                   ucp_rkey_h *rkey)
+{
+    void *rkey_buffer = NULL;
+    size_t rkey_size  = 0;
+    ucs_status_t status;
+
+    status = ucp_rkey_pack(perf->ucp.context, memh, &rkey_buffer, &rkey_size);
```
rakhmets, line 1052: `ucp_rkey_pack` is deprecated. Please use `ucp_memh_pack`.

src/tools/perf/lib/libperf.c
```diff
@@ -1028,6 +1041,35 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+        goto out;
+    }
+
+    status = ucp_ep_rkey_unpack(perf->ucp.self_ep, rkey_buffer, rkey);
+    if (status != UCS_OK) {
+        if (perf->params.flags & UCX_PERF_TEST_FLAG_VERBOSE) {
+            ucs_fatal("ucp_rkey_unpack() failed: %s",
+                      ucs_status_string(status));
+        }
+    }
```
rakhmets, line 1066: Maybe add a macro. And use it here and in `ucp_perf_test_rkey_pack`, `ucp_perf_test_rkey_unpack`.

src/tools/perf/lib/libperf.c
```diff
@@ -898,6 +882,12 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
         ucp_params->features |= UCP_FEATURE_WAKEUP;
     }
 
+    if ((params->send_mem_type != UCS_MEMORY_TYPE_HOST) ||
+        (params->recv_mem_type != UCS_MEMORY_TYPE_HOST)) {
```
rakhmets, line 886:
```
static int ucp_perf_test_is_host_mem(const ucx_perf_params_t *params)
{
    return (params->send_mem_type != UCS_MEMORY_TYPE_HOST) ||
           (params->recv_mem_type != UCS_MEMORY_TYPE_HOST);
}
```

Here and in `ucx_perf_allocators_init`.

src/tools/perf/lib/libperf.c
```diff
@@ -906,30 +896,53 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
...
+    unsigned i;
+
+    for (i = 0; i < num_rkeys; ++i) {
+        if (rkeys[i] != NULL) {
+            ucp_rkey_destroy(rkeys[i]);
+        }
+    }
+
+    if (ep != NULL) {
+        req = ucp_ep_close_nb(ep, UCP_EP_CLOSE_MODE_FLUSH);
```
rakhmets, line 912: `ucp_ep_close_nb` is deprecated. Please use `ucp_ep_close_nbx` instead.

src/tools/perf/lib/libperf.c
```diff
@@ -1257,6 +1302,64 @@ static ucs_status_t ucp_perf_test_flush_workers(ucx_perf_context_t *perf)
...
+{
+    ucp_worker_h worker = perf->ucp.tctx[0].perf.ucp.worker;
+    size_t address_length;
+    ucp_address_t *address;
+    ucs_status_t status;
+    ucp_ep_params_t ep_params;
+    ucx_perf_context_t *thread_perf;
+    unsigned i;
+
+    status = ucp_worker_get_address(worker, &address, &address_length);
```
rakhmets, line 1315: `ucp_worker_get_address` is deprecated. Please use `ucp_worker_query` instead.

src/tools/perf/lib/libperf.c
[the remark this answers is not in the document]
roiedanino: ucp_rkey.c:587 `ucs_fatal("packing rkey using ucp_memh_pack() is unsupported");`

I don't think I have any choice here but to use `ucp_rkey_pack`

src/tools/perf/lib/libperf.c
```diff
@@ -1028,6 +1048,30 @@ ucp_perf_test_rkey_unpack(ucx_perf_thread_context_t *thread,
...
+    status = ucp_rkey_pack(perf->ucp.context, memh, &rkey_buffer, &rkey_size);
+    if (status != UCS_OK) {
+        UCX_PERF_VERBOSE(fatal, &perf->params, "ucp_rkey_pack", status);
+        return status;
+    }
+
+    status = ucp_ep_rkey_unpack(perf->ucp.self_ep, rkey_buffer, rkey);
+    if (status != UCS_OK) {
+        UCX_PERF_VERBOSE(fatal, &perf->params, "ucp_ep_rkey_unpack", status);
+    }
```
rakhmets, line 1069:
```suggestion
#define UCX_PERF_VERBOSE(_log_level, _params, _func, ...) \
    ({ \
        ucs_status_t _status = _func(__VA_ARGS__); \
        if ((_status != UCS_OK) && \
            ((_params)->flags & UCX_PERF_TEST_FLAG_VERBOSE)) { \
            ucs_##_log_level("%s() failed: %s", UCS_PP_MAKE_STRING(_func), \
                             ucs_status_string(_status)); \
        } \
        _status; \
    })
...
    status = UCX_PERF_VERBOSE(fatal, &perf->params, ucp_ep_rkey_unpack,
                              perf->ucp.self_ep, rkey_buffer, rkey);
```

src/tools/perf/lib/libperf.c
```diff
@@ -1257,6 +1300,61 @@ static ucs_status_t ucp_perf_test_flush_workers(ucx_perf_context_t *perf)
...
+    perf->ucp.self_send_rkey = perf->ucp.tctx[0].perf.ucp.self_send_rkey;
+    perf->ucp.self_recv_rkey = perf->ucp.tctx[0].perf.ucp.self_recv_rkey;
+
+    ucp_worker_release_address(worker, worker_attr.address);
+    return UCS_OK;
+
+err_destroy_eps:
+    ucp_perf_test_destroy_eps(perf);
+    ucp_worker_release_address(worker, worker_attr.address);
+    return status;
```
rakhmets, line 1355:
```suggestion
    status = UCS_OK;
    goto out;

err_destroy_eps:
    ucp_perf_test_destroy_eps(perf);
out:
    ucp_worker_release_address(worker, worker_attr.address);
    return status;
```

src/tools/perf/lib/libperf.c
```diff
@@ -1257,6 +1300,61 @@ static ucs_status_t ucp_perf_test_flush_workers(ucx_perf_context_t *perf)
...
+    ucp_worker_attr_t worker_attr = {
+        .field_mask = UCP_WORKER_ATTR_FIELD_ADDRESS
+    };
+    unsigned i;
+
+    status = ucp_worker_query(worker, &worker_attr);
+    if (status != UCS_OK) {
+        ucs_error("ucp_worker_query() failed: %s", ucs_status_string(status));
+        return status;
+    }
```
rakhmets, line 1318:
```suggestion
    status = UCX_PERF_VERBOSE(error, &perf->params, ucp_worker_query, worker,
                              &worker_attr);
    if (status != UCS_OK) {
        return status;
    }
```

src/tools/perf/lib/libperf.c
```diff
@@ -1053,11 +1098,14 @@ static ucs_status_t ucp_perf_test_receive_remote_data(ucx_perf_context_t *perf,
 
     /* Initialize all endpoints and rkeys to NULL to handle error flow */
     for (i = 0; i < thread_count; i++) {
-        perf->ucp.tctx[i].perf.ucp.ep   = NULL;
-        perf->ucp.tctx[i].perf.ucp.rkey = NULL;
+        perf->ucp.tctx[i].perf.ucp.ep             = NULL;
+        perf->ucp.tctx[i].perf.ucp.rkey           = NULL;
+        perf->ucp.tctx[i].perf.ucp.self_ep        = NULL;
+        perf->ucp.tctx[i].perf.ucp.self_send_rkey = NULL;
+        perf->ucp.tctx[i].perf.ucp.self_recv_rkey = NULL;
```
rakhmets, line 1105: Seems that this should be in `ucp_perf_test_setup_self_endpoints`.

src/tools/perf/lib/libperf.c
```diff
@@ -1279,25 +1377,33 @@ static ucs_status_t ucp_perf_test_setup_endpoints(ucx_perf_context_t *perf,
...
     if (status != UCS_OK) {
         goto err_destroy_eps;
     }
 
-    /* force wireup completion */
+    /* Create self_ep used for copying into non-host memory types*/
+    if (!ucp_perf_test_is_host_mem_only(&perf->params)) {
+        status = ucp_perf_test_setup_self_endpoints(perf);
+        if (status != UCS_OK) {
+            goto err_destroy_eps;
```
rakhmets, line 1402: `ucp_perf_test_destroy_eps` will be called twice in case of error inside `ucp_perf_test_setup_self_endpoints`.
Please split clean-up of ep, rkey, and self_ep, self_send_rkey, self_recv_rkey into two methods.

src/tools/perf/lib/libperf.c
roiedanino: `ucp_perf_test_setup_self_endpoints` is being called conditionally, we would want those attributes to be NULL otherwise, so in destruction, we will be able to check whether we should release those eps and rkeys.

src/tools/perf/perftest_params.c
```diff
@@ -21,11 +21,8 @@ static void print_memory_type_usage(void)
     ucs_memory_type_t it;
 
     ucs_memory_type_for_each(it) {
-        if (ucx_perf_mem_type_allocators[it] != NULL) {
-            printf("                        %s - %s\n",
-                   ucs_memory_type_names[it],
-                   ucs_memory_type_descs[it]);
-        }
+        printf("                        %s - %s\n", ucs_memory_type_names[it],
+               ucs_memory_type_descs[it]);
```
rakhmets, line 25: I think that `ucx_perftest` will not work with all available memory types unconditionally.

Before the changes `ucx_perftest -h` showed:
if CUDA is not loaded:
```
     -m <send mem type>[,<recv mem type>]
                    memory type of message for sender and receiver (host)
                        host - System memory
```
if CUDA is loaded:
```
     -m <send mem type>[,<recv mem type>]
                    memory type of message for sender and receiver (host)
                        host - System memory
                        cuda - NVIDIA GPU memory
                        cuda-managed - NVIDIA GPU managed/unified memory
```

After the changes `ucx_perftest -h` will unconditionally show:
```
     -m <send mem type>[,<recv mem type>]
                    memory type of message for sender and receiver (host)
                        host - System memory
                        cuda - NVIDIA GPU memory
                        cuda-managed - NVIDIA GPU managed/unified memory
                        rocm - AMD/ROCm GPU memory
                        rocm-managed - AMD/ROCm GPU managed memory
                        rdma - RDMA device memory
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2137,7 +2137,7 @@ uct_ib_mlx5_devx_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
         return UCS_ERR_INVALID_PARAM;
     }
 
-    if (flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA) {
+    if ((flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA) || (memh->dm != NULL)) {
```
rakhmets, line 2140: Is this done intentionally? If so, why is this needed?

roiedanino: Device memory depends on an indirect key, the only way to access it is through RDMA operations.

rakhmets (approved):

PR #9339, opened 2023-09-04
UCP/PROTO: Calculate perf for ack messages as parallel stages

ivankochin:
## What
Changes the way how do we calculate ACK performance for multi protocols. Previously we simply added the ack perf to the perf of the proto, but the calculating it as the parallel stages (choosing the slowest stage in case of multi perf type) seems the correct way to do that.

## Why ?
To improve perf estimations for RNDV protos. Without this patch multi benchmarks with many PPN prefer RNDV through AM instead of pure RNDV.

## Illustration
| Before| After|
|--------|--------|
| image | image|

### DIFF
--- a/src/ucp/proto/proto_init.c
+++ b/src/ucp/proto/proto_init.c
@@ -189,13 +189,13 @@ ucp_proto_perf_envelope_make(const ucp_proto_perf_list_t *perf_list,
 }
 
 ucs_status_t
-ucp_proto_init_parallel_stages(const ucp_proto_common_init_params_t *params,
+ucp_proto_init_parallel_stages(const ucp_proto_init_params_t *params,
                                size_t range_start, size_t range_end,
                                size_t frag_size, double bias,
                                const ucp_proto_perf_range_t **stages,
-                               unsigned num_stages)
+                               unsigned num_stages, unsigned flags)
 {
-    ucp_proto_caps_t *caps      = params->super.caps;
+    ucp_proto_caps_t *caps      = params->caps;
     ucs_linear_func_t bias_func = ucs_linear_func_make(0.0, 1.0 - bias);
     UCS_ARRAY_DEFINE_ONSTACK(stage_list, ucp_proto_perf_list, 16);
     UCS_ARRAY_DEFINE_ONSTACK(concave, ucp_proto_perf_envelope, 16);
@@ -226,7 +226,7 @@ ucp_proto_init_parallel_stages(const ucp_proto_common_init_params_t *params,
             /* For multi-fragment protocols, we need to apply the fragment
              * size to the performance function linear factor.
              */
-            if (!(params->flags & UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG)) {
+            if (!(flags & UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG)) {
                 perf[perf_type].m += perf[perf_type].c / frag_size;
             }
         }
@@ -268,10 +268,10 @@ ucp_proto_init_parallel_stages(const ucp_proto_common_init_params_t *params,
         range             = &caps->ranges[caps->num_ranges];
         range->max_length = elem->max_length;
         if (fabs(bias) > UCP_PROTO_PERF_EPSILON) {
-            range->node   = ucp_proto_perf_node_new_data(params->super.proto_name,
+            range->node   = ucp_proto_perf_node_new_data(params->proto_name,
                                                          "bias %f", bias);
         } else {
-            range->node   = ucp_proto_perf_node_new_data(params->super.proto_name,
+            range->node   = ucp_proto_perf_node_new_data(params->proto_name,
                                                          "");
         }
 
@@ -698,8 +698,8 @@ ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
     parallel_stages[2] = &recv_perf;
 
     /* Add ranges representing sending single fragment */
-    status = ucp_proto_init_parallel_stages(params, 0, frag_size, frag_size,
-                                            0.0, parallel_stages, 3);
+    status = ucp_proto_init_parallel_stages(&params->super, 0, frag_size, frag_size,
+                                            0.0, parallel_stages, 3, params->flags);
     if (status != UCS_OK) {
         goto out_deref_recv_perf;
     }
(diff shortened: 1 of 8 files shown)


ivankochin:
Need to test performance before merge

src/ucp/proto/proto_init.c
```diff
@@ -189,14 +189,14 @@ ucp_proto_perf_envelope_make(const ucp_proto_perf_list_t *perf_list,
...
                                size_t range_start, size_t range_end,
                                size_t frag_size, double bias,
                                const ucp_proto_perf_range_t **stages,
-                               unsigned num_stages)
+                               unsigned num_stages, unsigned flags)
 {
-    ucp_proto_caps_t *caps      = params->super.caps;
-    ucs_linear_func_t bias_func = ucs_linear_func_make(0.0, 1.0 - bias);
+    ucp_proto_caps_t *caps        = params->caps;
+    ucs_linear_func_t bias_func   = ucs_linear_func_make(0.0, 1.0 - bias);
```
ivankochin, line 199:
```suggestion
    ucp_proto_caps_t *caps      = params->caps;
    ucs_linear_func_t bias_func = ucs_linear_func_make(0.0, 1.0 - bias);
```

ivankochin:
The changes in gdr_copy perf estimations are required to fix performance of this patch on DGX2.

brminich (dismissed):

ivankochin:
Performance was checked. Ready for review.

ivankochin:
@brminich @yosefe pls review

brminich (dismissed):

src/ucp/rndv/proto_rndv.c
```diff
@@ -489,38 +491,40 @@ ucs_status_t ucp_proto_rndv_rts_reset(ucp_request_t *req)
...
                                       &receive_time);
     if (status != UCS_OK) {
         return status;
     }
+    ack_func = ucs_linear_func_make(send_time + receive_time, 0);
 
     ack_perf[UCP_PROTO_PERF_TYPE_SINGLE] =
-            ucs_linear_func_make(send_time + receive_time, 0);
     ack_perf[UCP_PROTO_PERF_TYPE_MULTI] =
+            ucs_linear_func_add(ack_func, overhead);
```
yosefe, line 510: why changed MULTI to include receive_time?
also, according to this code, SINGLE should be equal to MULTI, but according to the perf graph image, they are different:
image

src/ucp/proto/proto_init.c
```diff
@@ -189,13 +189,13 @@ ucp_proto_perf_envelope_make(const ucp_proto_perf_list_t *perf_list,
 }
 
 ucs_status_t
-ucp_proto_init_parallel_stages(const ucp_proto_common_init_params_t *params,
+ucp_proto_init_parallel_stages(const ucp_proto_init_params_t *params,
                                size_t range_start, size_t range_end,
                                size_t frag_size, double bias,
                                const ucp_proto_perf_range_t **stages,
-                               unsigned num_stages)
+                               unsigned num_stages, unsigned flags)
```
yosefe, line 196: can we pass frag_size==SIZE_MAX when need to pass flags with UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG? to avoid adding extra parameter

src/ucp/rndv/proto_rndv.c
ivankochin: As I understand, for RTS,RTR and ACK messages we don't calculate the send/xfer/recv overheads as parallel stages in MULTI mode to simplify our performance estimation model. But according to my perf measurements on different setups, the most part of overhead on these messages is transfer time, which means that it can be executed simultaneously with other stages in case of MULTI mode.

That's why the closest estimation without adding extra-complexity would be setting the full overhead in MULTI mode too. 

BTW we already do the same for RTS, but there we set the full overhead to all modes including CPU. (pls see here: https://github.com/openucx/ucx/blob/master/src/ucp/rndv/proto_rndv.c#L337)

Regarding the graph: thanks, you are right, it is outdated. I will update the new one to the description.

yosefe: But receive_time also includes the xfer time. So including it in MULTI perf is as if ACKs cannot be sent in parallel. Is that what you meant to do?

src/ucp/proto/proto_init.c
ivankochin: It seems like we can, but I think that passing additional argument is less evil then passing `frag_size` that is higher then `max_length` in some cases. It would bring additional complexity to `ucp_proto_common_init_caps` and `ucp_proto_init_parallel_stages` logic.

src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -119,10 +120,6 @@ uct_gdr_copy_estimate_perf(uct_iface_h iface, uct_perf_attr_t *perf_attr)
                 perf_attr->bandwidth.dedicated = 250.0 * UCS_MBYTE;
                 perf_attr->bandwidth.shared    = 0;
                 break;
-            case UCT_EP_OP_PUT_SHORT:
-                perf_attr->bandwidth.dedicated = 0;
-                perf_attr->bandwidth.shared    = 10200.0 * UCS_MBYTE;
-                break;
```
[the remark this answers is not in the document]
ivankochin: It was several weeks ago, but as I remember I changed it basing on results got from gdr_copy internal benchmarks. I can recheck if needed.

src/ucp/rndv/proto_rndv.c
ivankochin: Discussed offline.

brminich (dismissed):

src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -112,20 +112,17 @@ uct_gdr_copy_iface_query(uct_iface_h tl_iface, uct_iface_attr_t *iface_attr)
...
-            default:
-                perf_attr->bandwidth.dedicated = 0;
-                perf_attr->bandwidth.shared    =
-                        UCT_GDR_COPY_IFACE_DEFAULT_BANDWIDTH;
-            }
-        } else {
+        switch (op) {
+        case UCT_EP_OP_GET_SHORT:
+        case UCT_EP_OP_GET_ZCOPY:
```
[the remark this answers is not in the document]
ivankochin: Done.

yosefe (approved):

yosefe (approved):

PR #9341, opened 2023-09-04
UCS/SYS/TOPO: Use default distance if sysfs path fails.

rakhmets:
## What
`ucs_topo_get_distance_sysfs` returns default distance if sysfs path provider fails.
Plus tracing different code paths of `ucs_topo_get_distance_sysfs`.

## Why ?
The changes are needed to properly handle UCX usage on integrated GPUs.
Assert is triggered without these changes.

### DIFF
--- a/src/ucs/sys/topo/base/topo.c
+++ b/src/ucs/sys/topo/base/topo.c
@@ -345,30 +345,34 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
     /* If one of the devices is unknown, we assume near topology */
     if ((device1 == UCS_SYS_DEVICE_ID_UNKNOWN) ||
         (device2 == UCS_SYS_DEVICE_ID_UNKNOWN) || (device1 == device2)) {
-        *distance = ucs_topo_default_distance;
-        return UCS_OK;
+        goto default_distance;
     }
 
     status = ucs_topo_sys_dev_to_sysfs_path(device1, path1, sizeof(path1));
     if (status != UCS_OK) {
-        return status;
+        ucs_debug("failed to get sysfs path for %s",
+                  ucs_topo_sys_device_get_name(device1));
+        goto default_distance;
     }
 
     status = ucs_topo_sys_dev_to_sysfs_path(device2, path2, sizeof(path2));
     if (status != UCS_OK) {
-        return status;
+        ucs_debug("failed to get sysfs path for %s",
+                  ucs_topo_sys_device_get_name(device2));
+        goto default_distance;
     }
 
     ucs_path_get_common_parent(path1, path2, common_path);
     if (ucs_topo_is_sys_root(common_path)) {
         ucs_topo_sys_root_distance(distance);
+        return UCS_OK;
     } else if (ucs_topo_is_pci_root(common_path)) {
         ucs_topo_pci_root_distance(path1, path2, distance);
-    } else {
-        *distance = ucs_topo_default_distance;
+        return UCS_OK;
     }
 
-    return UCS_OK;
+default_distance:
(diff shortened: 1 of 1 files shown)


src/ucs/sys/topo/base/topo.c
```diff
@@ -345,30 +345,37 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
...
-        *distance = ucs_topo_default_distance;
+        return UCS_OK;
     }
 
-    return UCS_OK;
+default_distance:
+    ucs_trace("default distance between %s and %s",
+              ucs_topo_sys_device_get_name(device1),
+              ucs_topo_sys_device_get_name(device2));
+    return ucs_topo_get_distance_default(device1, device2, distance);
```
brminich, line 378: minor: i'd add space line here

src/ucs/sys/topo/base/topo.c
```diff
@@ -345,30 +345,37 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
...
         (device2 == UCS_SYS_DEVICE_ID_UNKNOWN) || (device1 == device2)) {
-        *distance = ucs_topo_default_distance;
-        return UCS_OK;
+        goto default_distance;
     }
 
     status = ucs_topo_sys_dev_to_sysfs_path(device1, path1, sizeof(path1));
     if (status != UCS_OK) {
-        return status;
+        ucs_trace("failed to get sysfs path for %s",
```
brminich, line 353: maybe better debug, as it is not supposed to be a frequent message?

rakhmets: `ucs_topo_get_distance` is called from `ucp_proto_common_get_lane_distance`, which is called from other methods: ucp_proto_rndv_ctrl_select_remote_proto, ucp_proto_common_get_lane_perf...
Is it still worth to make the level of verbosity debug?

src/ucs/sys/topo/base/topo.c
rakhmets: Added

src/ucs/sys/topo/base/topo.c
rakhmets: Changed to debug.

brminich (dismissed):

src/ucs/sys/topo/base/topo.c
```diff
@@ -345,30 +345,38 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
...
     } else if (ucs_topo_is_pci_root(common_path)) {
         ucs_topo_pci_root_distance(path1, path2, distance);
-    } else {
-        *distance = ucs_topo_default_distance;
+        return UCS_OK;
     }
 
-    return UCS_OK;
+default_distance:
+    ucs_debug("default distance between %s and %s",
```
yosefe, line 375: IMO better remove this log print, UCP layer should print it already

rakhmets: Removed

yosefe (approved):

agirault:
Will this totally remove the need for this patch (both in ucp: `proto_rndv.c` and `proto_common.c`)?

https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/patches/ucx_error_handling_topo_get_distance.patch

rakhmets:
> Will this totally remove the need for this patch (both in ucp: `proto_rndv.c` and `proto_common.c`)?
> 
> https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/patches/ucx_error_handling_topo_get_distance.patch

Yes, because the PR fixes the issue in ucs_topo_get_distance.

PR #9342, opened 2023-09-04
UCS/MEMORY: Make fail of numa node file opening silent.

rakhmets:
## What
 Make fail of numa node file opening silent.

## Why ?
There is no need to print error message here, because the case is handled on the next lines by returning `UCS_NUMA_NODE_DEFAULT`.

### DIFF
--- a/src/ucs/memory/numa.c
+++ b/src/ucs/memory/numa.c
@@ -147,7 +147,7 @@ ucs_numa_node_t ucs_numa_node_of_device(const char *dev_path)
     long parsed_node;
     ucs_status_t status;
 
-    status = ucs_read_file_number(&parsed_node, 0, "%s/numa_node", dev_path);
+    status = ucs_read_file_number(&parsed_node, 1, "%s/numa_node", dev_path);
 
     if ((status != UCS_OK) || (parsed_node < 0) ||
         (parsed_node >= UCS_NUMA_NODE_MAX)) {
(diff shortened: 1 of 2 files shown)


src/ucs/memory/numa.c
```diff
@@ -147,7 +147,7 @@ ucs_numa_node_t ucs_numa_node_of_device(const char *dev_path)
     long parsed_node;
     ucs_status_t status;
 
-    status = ucs_read_file_number(&parsed_node, 0, "%s/numa_node", dev_path);
+    status = ucs_read_file_number(&parsed_node, 1, "%s/numa_node", dev_path);
```
brminich, line 150: it's better to add some print on the line 154 then

brminich: but it does not print anything when silent arg is provided

rakhmets: Yes. I meant to update `ucs_read_file_vararg`, and add else case to print the message with other verbosity level.

brminich: it's also an option, but i'd check where it is used to avoid trace flooding

src/ucs/memory/numa.c
```diff
@@ -147,10 +147,16 @@ ucs_numa_node_t ucs_numa_node_of_device(const char *dev_path)
...
+    if (status != UCS_OK) {
+        ucs_trace("failed to read number from %s/numa_node: %s", dev_path,
```
yosefe, line 152: maybe replace "int silent" parameter by "ucs_log_level_t err_level"?

PR #9344, opened 2023-09-05
UCT/API: Add rkey_compare for non-ib, non-sm transports

tvegas1:
## What
Add rkey_compare implementation for non-shared memory and non-ib transports.

## Why ?
All transports must implement a valid comparison function. UCT tests coming when all transports will implement rkey_compare.

## How ?
- Using memcmp with clearing any padding byte seems more future proof.
- Add MD flag attribute to set expectations. This can be used, for instance, by testing infra to set proper expectations when comparing keys unpacked twice, for instance.


tvegas1:
will use common compare approach instead.

PR #9345, opened 2023-09-05
UCP/MEMORY: Don't register when looking up RMA capable host memory allocator

tvegas1:
## What
Don't register memory when looking up host memory allocator as it can cause issues with cuda.

## Why ?
Caching the md index for host memory involves alloc/mem_reg/mem_dereg. This is performed at send time during proto v2 RTR mtype init. With cuda_copy `cudaHostUnregister()` is called for mem_dereg.

With nccl-test and UCX plugin seems we have deadlock situation:
- thread 1: pending for completion for cuda job: `testStreamSynchronize() { .. cudaStreamQuery(streams[i]); .. }`, not ready
- thread 2: the sender that progresses the test, but stuck on `cudaHostUnregister()`, as it seems synchronuous unregister cannot progress while cuda job is ongoing.

When forcing thread 1 to finish/timeout, the call to `cudaHostUnregister()` in thread 2 completes properly.
- thread 1 unblocks thread 2 when setting cuda malloced flag: `ncclCommAbort() { ... *comm->abortFlag = 1; ... }`

## How ?
Keep allocation-based probing (sysv huge tlb case), but don't try to register memory.

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -599,11 +599,29 @@ ucp_memh_rcache_get(ucs_rcache_t *rcache, void *address, size_t length,
     return UCS_OK;
 }
 
+static ucp_md_index_t ucp_mem_get_md_index(ucp_context_h context,
+                                           const uct_md_h md,
+                                           uct_alloc_method_t method)
+{
+    ucp_md_index_t md_index;
+
+    if (method != UCT_ALLOC_METHOD_MD) {
+        return UCP_NULL_RESOURCE;
+    }
+
+    for (md_index = 0; md_index < context->num_mds; md_index++) {
+        if (md == context->tl_mds[md_index].md) {
+            return md_index;
+        }
+    }
+
+    return UCP_NULL_RESOURCE;
+}
+
 static ucs_status_t ucp_memh_create_from_mem(ucp_context_h context,
                                              const uct_allocated_memory_t *mem,
                                              ucp_mem_h *memh_p)
 {
-    ucp_md_index_t md_index;
     ucs_status_t status;
     ucp_mem_h memh;
 
@@ -613,27 +631,16 @@ static ucs_status_t ucp_memh_create_from_mem(ucp_context_h context,
         return status;
     }
 
-    if (mem->method != UCT_ALLOC_METHOD_MD) {
-        goto out;
-    }
-
-    for (md_index = 0; md_index < context->num_mds; md_index++) {
-        if (mem->md != context->tl_mds[md_index].md) {
-            continue;
-        }
-
-        memh->alloc_md_index = md_index;
-        memh->uct[md_index]  = mem->memh;
-        memh->md_map        |= UCS_BIT(md_index);
+    memh->alloc_md_index = ucp_mem_get_md_index(context, mem->md, mem->method);
+    if (memh->alloc_md_index != UCP_NULL_RESOURCE) {
+        memh->uct[memh->alloc_md_index] = mem->memh;
+        memh->md_map                   |= UCS_BIT(memh->alloc_md_index);
         ucs_trace("allocated address %p length %zu on md[%d]=%s %p",
-                  mem->address, mem->length, md_index,
-                  context->tl_mds[md_index].rsc.md_name, memh->uct[md_index]);
-        break;
+                  mem->address, mem->length, memh->alloc_md_index,
+                  context->tl_mds[memh->alloc_md_index].rsc.md_name,
+                  memh->uct[memh->alloc_md_index]);
     }
 
-    ucs_assert(memh->alloc_md_index != UCP_NULL_RESOURCE);
-
-out:
     *memh_p = memh;
     return UCS_OK;
 }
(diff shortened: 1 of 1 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -1463,30 +1463,22 @@ void ucp_mem_rcache_cleanup(ucp_context_h context)
...
-                                        memh_reg_flags, "get_alloc_md_id",
-                                        &alloc_memh);
-        if (status != UCS_OK) {
-            return status;
-        }
-
-        context->alloc_md_index_initialized     = 1;
-        context->alloc_md_index[alloc_mem_type] = alloc_memh->alloc_md_index;
+    memh_reg_flags = UCT_MD_MEM_ACCESS_RMA | UCT_MD_MEM_FLAG_HIDE_ERRORS;
+    status         = ucp_memh_alloc(context, NULL, 1, UCS_MEMORY_TYPE_HOST,
```
yosefe, line 1473: this may trigger set (or using) GPU device context during ucp_init() and we don't want that
maybe we can find the alloc md index without doing cudaHostRegister? or even without actually allocating anything?

yosefe (changes requested):

src/ucp/core/ucp_mm.c
tvegas1: Updated without the registration part, but still exercising alloc path. cuda/rocm alloc early refuse host memory allocation so we should be good on that side? But if we want for instance to be able to use sysv/huge tlb actual function, I did not see simple way to test without allocating. Else we could just pick mmap, no alloc, but I think this would depart from current behavior.

src/ucp/core/ucp_context.c
```diff
@@ -2174,6 +2172,13 @@ ucs_status_t ucp_init_version(unsigned api_major_version, unsigned api_minor_ver
     context->uuid             = ucs_generate_uuid((uintptr_t)context);
     context->next_memh_reg_id = 0;
 
+    /* Find MD to use for host memory allocation */
+    if (ucp_mem_get_host_alloc_md_index(context) != UCS_OK) {
+        ucs_error("could not find MD for host memory allocation: %s",
+                  ucs_status_string(status));
+        goto err_free_res;
```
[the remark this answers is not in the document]
brminich: no, we have to fail just `ucp_proto_rndv_rtr_mtype_init`

src/ucp/core/ucp_mm.c
```diff
@@ -613,26 +636,16 @@ static ucs_status_t ucp_memh_create_from_mem(ucp_context_h context,
...
-        goto out;
-    }
-
-    for (md_index = 0; md_index < context->num_mds; md_index++) {
-        if (mem->md != context->tl_mds[md_index].md) {
-            continue;
-        }
-
+    md_index = ucp_mem_get_md_index(context, mem->md, mem->method);
+    if (md_index != UCP_NULL_RESOURCE) {
```
yosefe, line 640: can we remove the local var md_index and the "out" label?

tvegas1: fixed both and retested that original issue is still fixed.

src/ucp/core/ucp_mm.c
```diff
@@ -1467,22 +1478,23 @@ ucs_status_t
...
+                                  &mem);
         if (status != UCS_OK) {
             return status;
         }
 
-        context->alloc_md_index_initialized     = 1;
-        context->alloc_md_index[alloc_mem_type] = alloc_memh->alloc_md_index;
-        ucp_memh_cleanup(context, alloc_memh);
+        context->alloc_md_index_initialized = 1;
+        context->alloc_md_index[alloc_mem_type] =
```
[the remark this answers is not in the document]
tvegas1: fixed

brminich (dismissed):

src/ucp/core/ucp_mm.c
```diff
@@ -599,11 +599,33 @@ ucp_memh_rcache_get(ucs_rcache_t *rcache, void *address, size_t length,
...
+    }
+
+    for (index = 0; index < context->num_mds; index++) {
+        if (md == context->tl_mds[index].md) {
+            md_index = index;
+            break;
+        }
+    }
+
+    ucs_assert(md_index != UCP_NULL_RESOURCE);
```
yosefe, line 620: IMO this assertion is not needed

src/ucp/core/ucp_mm.c
```diff
@@ -599,11 +599,33 @@ ucp_memh_rcache_get(ucs_rcache_t *rcache, void *address, size_t length,
     return UCS_OK;
 }
 
+static ucp_md_index_t ucp_mem_get_md_index(ucp_context_h context,
+                                           const uct_md_h md,
+                                           uct_alloc_method_t method)
+{
+    ucp_md_index_t md_index = UCP_NULL_RESOURCE;
+    ucp_md_index_t index;
```
yosefe, line 607: why need also "index" variable?
IMO we can just use "md_index" as the loop variable

src/ucp/core/ucp_mm.c
```diff
@@ -599,11 +599,33 @@ ucp_memh_rcache_get(ucs_rcache_t *rcache, void *address, size_t length,
...
+
+    if (method != UCT_ALLOC_METHOD_MD) {
+        goto out;
+    }
+
+    for (index = 0; index < context->num_mds; index++) {
+        if (md == context->tl_mds[index].md) {
+            md_index = index;
+            break;
```
yosefe, line 616: here we can "return md_index"

src/ucp/core/ucp_mm.c
tvegas1: removed index

src/ucp/core/ucp_mm.c
tvegas1: returning md_index

src/ucp/core/ucp_mm.c
tvegas1: removed

brminich (approved):

yosefe (approved):
pls squash

yosefe:
@tvegas1 can you pls fix the commit and PR title to reflect the latest changes?

yosefe (approved):

brminich (approved):

PR #9346, opened 2023-09-05
UCT/CUDA/CUDA_COPY: Set device context within mem alloc.

rakhmets:
## What
Set primary device context as current context for the calling thread, if there is no active context during memory allocation.

### DIFF
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.c
@@ -205,13 +205,47 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_dereg,
     return UCS_OK;
 }
 
-static ucs_status_t uct_cuda_copy_mem_alloc(uct_md_h md, size_t *length_p,
-                                            void **address_p,
-                                            ucs_memory_type_t mem_type,
-                                            unsigned flags,
-                                            const char *alloc_name,
-                                            uct_mem_h *memh_p)
+static ucs_status_t uct_cuda_copy_md_set_ctx(uct_cuda_copy_md_t *md)
 {
+    ucs_status_t status;
+    CUdevice dev;
+    CUcontext ctx;
+
+    ucs_assertv(!md->cuda_ctx_retained,
+                "md %p: cuda primary context has already been retained", md);
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&dev, 0));
+    if (status != UCS_OK) {
+        goto err;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuDevicePrimaryCtxRetain(&ctx, dev));
+    if (status != UCS_OK) {
+        goto err;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxSetCurrent(ctx));
+    if (status != UCS_OK) {
+        goto err_release;
+    }
+
+    md->cuda_ctx_retained = 1;
+    ucs_debug("md %p: cuda primary context retained on gpu %d", md, dev);
+
+    return UCS_OK;
+
+err_release:
+    UCT_CUDADRV_FUNC_LOG_ERR(cuDevicePrimaryCtxRelease(dev));
+err:
+    return status;
+}
+
+static ucs_status_t
+uct_cuda_copy_mem_alloc(uct_md_h tl_md, size_t *length_p, void **address_p,
+                        ucs_memory_type_t mem_type, unsigned flags,
+                        const char *alloc_name, uct_mem_h *memh_p)
+{
+    uct_cuda_copy_md_t *md = ucs_derived_of(tl_md, uct_cuda_copy_md_t);
     ucs_status_t status;
 
     if ((mem_type != UCS_MEMORY_TYPE_CUDA_MANAGED) &&
@@ -220,8 +254,10 @@ static ucs_status_t uct_cuda_copy_mem_alloc(uct_md_h md, size_t *length_p,
     }
 
     if (!uct_cuda_base_is_context_active()) {
-        ucs_error("attempt to allocate cuda memory without active context");
-        return UCS_ERR_NO_DEVICE;
+        status = uct_cuda_copy_md_set_ctx(md);
+        if (status != UCS_OK) {
+            return UCS_ERR_NO_DEVICE;
+        }
     }
 
     if (mem_type == UCS_MEMORY_TYPE_CUDA) {
@@ -250,6 +286,13 @@ static ucs_status_t uct_cuda_copy_mem_free(uct_md_h md, uct_mem_h memh)
 
 static void uct_cuda_copy_md_close(uct_md_h uct_md) {
     uct_cuda_copy_md_t *md = ucs_derived_of(uct_md, uct_cuda_copy_md_t);
+    CUdevice dev;
+
+    if (md->cuda_ctx_retained &&
+        (UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&dev, 0)) == UCS_OK) &&
+        (UCT_CUDADRV_FUNC_LOG_ERR(cuDevicePrimaryCtxRelease(dev)) == UCS_OK)) {
+        ucs_debug("md %p: cuda primary context released on gpu %d", md, dev);
+    }
 
     ucs_free(md);
 }
@@ -585,6 +628,7 @@ uct_cuda_copy_md_open(uct_component_t *component, const char *md_name,
     md->config.max_reg_ratio    = config->max_reg_ratio;
     md->config.pref_loc         = config->pref_loc;
     md->config.dmabuf_supported = 0;
+    md->cuda_ctx_retained       = 0;
 
     dmabuf_supported = uct_cuda_copy_md_is_dmabuf_supported();
     if ((config->enable_dmabuf == UCS_YES) && !dmabuf_supported) {
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.h
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.h
@@ -24,6 +24,7 @@ typedef enum {
  */
 typedef struct uct_cuda_copy_md {
     struct uct_md               super;           /* Domain info */
+    int                         cuda_ctx_retained;
     struct {
         ucs_on_off_auto_value_t alloc_whole_reg; /* force return of allocation
                                                     range even for small bar


roiedanino (dismissed):

brminich (dismissed):

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -205,6 +205,39 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_dereg,
...
 }
 
+static ucs_status_t uct_cuda_copy_md_set_ctx(uct_cuda_copy_md_t *md)
+{
+    ucs_status_t status;
+    CUdevice dev;
+    CUcontext ctx;
+
+    ucs_assertv(!md->cuda_ctx_retained,
+                "cuda context has been already retain for the md %p", md);
```
gleon99, line 215: Minor: **retained**

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -250,6 +287,13 @@ static ucs_status_t uct_cuda_copy_mem_free(uct_md_h md, uct_mem_h memh)
 
 static void uct_cuda_copy_md_close(uct_md_h uct_md) {
     uct_cuda_copy_md_t *md = ucs_derived_of(uct_md, uct_cuda_copy_md_t);
+    CUdevice dev;
+
+    if (md->cuda_ctx_retained) {
```
gleon99, line 292: Maybe simplify to: `if (A && B)` ?
(Relying on `&&` order of execution is a definition in the C standard..)

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -220,8 +253,12 @@ static ucs_status_t uct_cuda_copy_mem_alloc(uct_md_h md, size_t *length_p,
     }
 
     if (!uct_cuda_base_is_context_active()) {
-        ucs_error("attempt to allocate cuda memory without active context");
-        return UCS_ERR_NO_DEVICE;
+        ucs_debug("attempt to allocate cuda memory without active context");
```
gleon99, line 256: Maybe rephrase the message, as it now represents a "lazy init" rather than some "wrong condition".

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -205,6 +205,39 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_dereg,
...
+    ucs_status_t status;
+    CUdevice dev;
+    CUcontext ctx;
+
+    ucs_assertv(!md->cuda_ctx_retained,
+                "cuda context has been already retain for the md %p", md);
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&dev, 0));
+    if (status != UCS_OK) {
+        goto ret;
```
yosefe, line 219: "ret" -> "err"

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -205,6 +205,39 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_dereg,
...
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxSetCurrent(ctx));
+    if (status != UCS_OK) {
+        goto release_ctx;
+    }
+
+    md->cuda_ctx_retained = 1;
+    return UCS_OK;
+
+release_ctx:
```
yosefe, line 235: err_release:

src/uct/cuda/cuda_copy/cuda_copy_md.c
yosefe: I would also print this message from uct_cuda_copy_md_set_ctx, when it succeeds

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -220,8 +253,12 @@ static ucs_status_t uct_cuda_copy_mem_alloc(uct_md_h md, size_t *length_p,
     }
 
     if (!uct_cuda_base_is_context_active()) {
-        ucs_error("attempt to allocate cuda memory without active context");
-        return UCS_ERR_NO_DEVICE;
+        ucs_debug("attempt to allocate cuda memory without active context");
+        status = uct_cuda_copy_md_set_ctx(
+                ucs_derived_of(md, uct_cuda_copy_md_t));
```
yosefe, line 258: rename 'md' to 'tl_md' and add local var
uct_cuda_copy_md_t *md = ucs_derived_of(...)

src/uct/cuda/cuda_copy/cuda_copy_md.h
```diff
@@ -32,6 +32,7 @@ typedef struct uct_cuda_copy_md {
         int                     dmabuf_supported;
         uct_cuda_pref_loc_t     pref_loc;
     } config;
+    int cuda_ctx_retained;
```
yosefe, line 35: move it to after 'super' and align on column

src/uct/cuda/cuda_copy/cuda_copy_md.c
rakhmets: Done

src/uct/cuda/cuda_copy/cuda_copy_md.h
rakhmets: Fixed.

src/uct/cuda/cuda_copy/cuda_copy_md.c
rakhmets: Fixed

src/uct/cuda/cuda_copy/cuda_copy_md.c
rakhmets: Fixed

src/uct/cuda/cuda_copy/cuda_copy_md.c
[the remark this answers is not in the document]
rakhmets: Updated the message.

src/uct/cuda/cuda_copy/cuda_copy_md.c
rakhmets: Updated the message and moved it to uct_cuda_copy_md_set_ctx.

src/uct/cuda/cuda_copy/cuda_copy_md.c
rakhmets: Done.

src/uct/cuda/cuda_copy/cuda_copy_md.c
rakhmets: Updated the message.

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -212,39 +212,43 @@ static ucs_status_t uct_cuda_copy_md_set_ctx(uct_cuda_copy_md_t *md)
...
     }
 
     status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxSetCurrent(ctx));
     if (status != UCS_OK) {
-        goto release_ctx;
+        goto err_release;
     }
 
     md->cuda_ctx_retained = 1;
+    ucs_debug("retained the primary context on the gpu %d and made it current "
```
gleon99, line 233: 1. IMO the message is too long.
2. Maybe include the `md` as well?
How about:
`md %p: cuda context retained on gpu %d`

yosefe: "md %p: cuda primary context retained on gpu %d"

rakhmets: reduced the message.
I left the word 'primary' because this is a special context. And made debug messages identical:
```
ucs_debug("md %p: retained the primary context on the gpu %d", md, dev);
ucs_debug("md %p: released the primary context on the gpu %d", md, dev);
```

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -250,6 +288,14 @@ static ucs_status_t uct_cuda_copy_mem_free(uct_md_h md, uct_mem_h memh)
...
 static void uct_cuda_copy_md_close(uct_md_h uct_md) {
     uct_cuda_copy_md_t *md = ucs_derived_of(uct_md, uct_cuda_copy_md_t);
+    CUdevice dev;
+
+    if (md->cuda_ctx_retained &&
+        (UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&dev, 0)) == UCS_OK) &&
+        (UCT_CUDADRV_FUNC_LOG_ERR(cuDevicePrimaryCtxRelease(dev)) == UCS_OK)) {
+        md->cuda_ctx_retained = 0;
+        ucs_debug("released the primary context on the gpu %d", dev);
+    }
```
brminich, line 298: looks a bit strange:
1. why is it needed to clear `md->cuda_ctx_retained` right before freeing this memory?
2. do we want to print something if we failed to release the context?

rakhmets: 1. removed
2. we print error from CUDA driver API calls.

src/uct/cuda/cuda_copy/cuda_copy_md.c
rakhmets:
```
ucs_assertv(!md->cuda_ctx_retained,
                "md %p: cuda primary context has already been retained", md);
ucs_debug("md %p: cuda primary context retained on gpu %d", md, dev);
ucs_debug("md %p: cuda primary context released on gpu %d", md, dev);
 ```

brminich (approved):

gleon99 (approved):

rakhmets:
@yosefe do you have any other comments? Or is it ok to squash?

yosefe (approved):
pls sqaush

Akshay-Venkatesh:
@rakhmets @yosefe Apologies for seeing the PR review request late. I thought we had discussed offline the following:
1. If calling thread doesn't have a context associated with it and it fails with a call to ucp_mem_map, then this is expected behaviour and it is up to the user (or the test which showed this behaviour) to ensure that the calling thread has the right device context before calling ucp_mem_map with alloc and mem_type=cuda flags.
2. This PR decides to retain [primary context of device 0](https://github.com/openucx/ucx/pull/9346/files#diff-f990528a14758f5a40bc106bd5fb5888513d667e4fc0b80256fe09dee83015c0R217). What happens if the process has one thread that has created a context on non-zero device and another thread belonging to the same context is the thread calling ucp_mem_map with alloc and mem_type=cuda flags? I think we would have the undesired effect of retaining primary context of device 0 when the caller of ucp_mem_map was probably trying to allocate memory on the non-zero device context that the other thread of the calling process already has a reference to. Also, by not only retaining context but setting it as the current context against the calling thread, you potentially cause the calling process to have the side effect of managing multiple device contexts when this was unintended. In general, when we have multi-GPU support per ucp context in UCX, we shouldn't set context and leave it that way when the thread returns from some UCP/UCT API but instead the thread should go back to whichever state it was in before making the UCP/UCT API call. In other words, if there was no context before then there should be no context after; if there was a device context  on device x before then the state should be unaltered afterwards.
3. Primary context is a resource that CUDA runtime API primarily manages. As UCX is expected to run with applications that use either driver API alone or runtime API alone or combination, UCX cannot assume that it can retain primary context. For instance what if it is an application that uses only driver API? We would end up touching a runtime resource that we should've never touched. Please refer to [primary context details here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#initialization).

IMO, we should revert this PR and explicitly mention in ucp_mem_map API that a valid device context needs to be set against the calling thread if allocation on a specific device is desired. Other attempts to do so should result in a failure. Please lmk your thoughts on this when possible.

yosefe:
@Akshay-Venkatesh @rakhmets what if we set cuda driver context (if it isn't set) only during ucp_mem_map() - for cuMemAlloc - and revert it before returning from UCX to the application?

Akshay-Venkatesh:
@yosefe That doesn't address the second case [here](https://github.com/openucx/ucx/pull/9346#issuecomment-1714347976) where user intends to run the application using non-zero device context. Concrete use case is when the process starts up and thread 0, it initializes some resources on device context corresponding to device 1 (say because CUDA_VISIBLE_DEVICES=0,1). Then another thread on the same process calls ucp_mem_map and which would lead to memory being allocated on device 0 which may not be intended by the user. We could potentially lookup all threads owned by the process by looking at /proc/<pid> and seeing if there is a context associated with any of the threads (not sure how we could do this if at all). There is also the complexity when there are different threads owned by the process and they manage two or more unique device contexts. We would get rid of all this ambiguity if we just failed when ucp_mem_map gets called without a device context set against it. One possibility is to provide UUID/BDF as a parameter to ucp_mem_map when wanting to allocate on device memory if the intent is to allocate on a specific device without wanting to explicitly set a device context against the calling thread. IMO caller must be responsible for setting a device context before they make a UCP call to allocate memory on that device. If there are instances where this behaviour accidentally works today, then we should discourage that and fail.

rakhmets:
>> @Akshay-Venkatesh @rakhmets what if we set cuda driver context (if it isn't set) only during ucp_mem_map() - for cuMemAlloc - and revert it before returning from UCX to the application?

> @yosefe That doesn't address the second case [here](https://github.com/openucx/ucx/pull/9346#issuecomment-1714347976) where user intends to run the application using non-zero device context. Concrete use case is when the process starts up and thread 0 initializes some resources on device context corresponding to device 1 (say because CUDA_VISIBLE_DEVICES=0,1). Then thread 1 on the same process calls ucp_mem_map would lead to memory being allocated on device 0 which may not be intended by the user. We could potentially lookup all threads owned by the process by looking at /proc/ and seeing if there is a context associated with any of the threads (not sure how we could do this if at all). There is also the complexity when there are different threads owned by the process and they manage two or more unique device contexts. We would get rid of all this ambiguity if we just failed when ucp_mem_map gets called without a device context set against it. One possibility is to provide UUID/BDF as a parameter to ucp_mem_map when wanting to allocate on device memory if the intent is to allocate on a specific device without wanting to explicitly set a device context against the calling thread. IMO caller must be responsible for setting a device context before they make a UCP call to allocate memory on that device. If there are instances where this behaviour accidentally works today, then we should discourage that and fail.
> 
> Another immediate issue is with how uct_mem_alloc gets used for allocating device buffers for staging managed memory transfers. If the application has two threads - one thread (t0) managing use of cuda managed buffers on device x!=0; another thread (t1) for managing communication using UCX. In this case, with the current logic, we would allocate all device buffers on device0 and hence we would stage managed memory pipeline transfers from potentially device x -> device 0 and then issuing network operations. Vice versa moving from device 0 -> device x on the receive path. This is problematic because we ideally want to stage through device x instead of device 0 (because intra-GPU transfers have the highest bandwidth).

If user binds CUDA context to the CPU thread before using UCX in the thread everything is OK. The user can either use CUDA Driver or CUDA Runtime API. cuCtxGetCurrent returns the context, and there is nothing to do from UCX side.

If user doesn't bind CUDA context to the CPU thread before using UCX in the thread, we may have the issues described by Akshay. Potentially, before and after the changes in this PR.

The issue is only if user doesn't bind the context to the CPU thread. This is not an optimal use case, but we still can support it.
We can retain primary device context. However, I don't see a proper way to return to the state which was before using UCX. Because releasing primary context may lead to primary device context reset...
We can create context in this case. And use it inside UCX.

If we don't create context in UCX, user doesn't bind the context to the CPU thread, the user will see an error... We can go this way. Users have to update their code in this case, by adding call to cudaSetDevice, or cuCtxSetCurrent in the CPU thread using UCX.

Akshay-Venkatesh:
> The issue is only if user doesn't bind the context to the CPU thread. This is not an optimal use case, but we still can support it.

We shouldn't support this case. 

> We can retain primary device context. However, I don't see a proper way to return to the state which was before using UCX. Because releasing primary context may lead to primary device context reset... We can create context in this case. And use it inside UCX.

Refer to case 3 in my comment where cuda runtime isn't used. UCX shouldn't be directly creating runtime resources 


> If we don't create context in UCX, user doesn't bind the context to the CPU thread, the user will see an error... We can go this way. Users have to update their code in this case, by adding call to cudaSetDevice, or cuCtxSetCurrent in the CPU thread using UCX.

Better to fail than lead to unexpected behavior. If we create a context on device 0 and user data resides on a device that isn't peer accessible, there can be failures too.

PR #9352, opened 2023-09-06
UCT/IB: do not check iface ID if it is not present in address

evgeny-leksikov:
## What
do not check interface ID if it is not present in address

## Why ?
```
UCX_IB_ADDR_TYPE=ib_global gtest --gtest_filter=ud/test_ucp_sockaddr.listen/0 --gtest_repeat=100
...
ib_iface.c:794  Assertion `gid->global.interface_id != 0' failed
```

### DIFF
--- a/src/uct/ib/base/ib_iface.c
+++ b/src/uct/ib/base/ib_iface.c
@@ -789,8 +789,10 @@ void uct_ib_iface_fill_ah_attr_from_gid_lid(uct_ib_iface_t *iface, uint16_t lid,
         ah_attr->src_path_bits = path_bits;
     }
 
-    if (iface->config.force_global_addr ||
-        (iface->gid_info.gid.global.subnet_prefix != gid->global.subnet_prefix)) {
+    if ((gid != NULL) &&
+        (iface->config.force_global_addr ||
+         (iface->gid_info.gid.global.subnet_prefix !=
+          gid->global.subnet_prefix))) {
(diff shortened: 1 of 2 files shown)


src/uct/ib/base/ib_iface.c
```diff
@@ -791,7 +791,6 @@ void uct_ib_iface_fill_ah_attr_from_gid_lid(uct_ib_iface_t *iface, uint16_t lid,
 
     if (iface->config.force_global_addr ||
         (iface->gid_info.gid.global.subnet_prefix != gid->global.subnet_prefix)) {
-        ucs_assert_always(gid->global.interface_id != 0);
```
yosefe, line 794: but it means we create ah_attr with global=1 and interface_id=0, doesn't it cause error during communication?
maybe we should use iface->config.force_global_addr only when packing the address, and not when creating ah_attr?

evgeny-leksikov: I found that we dont pack interface id only in case of eth link layer, AFAUI `iface->config.force_global_addr` should be ignorred then since `gid.raw` is used

evgeny-leksikov: in case of connect by sockaddr, rdmacm component is not aware about IB iface conffiguration and address is packed with parameters provided by rdmacm

yosefe (approved):

PR #9353, opened 2023-09-07
UCP/UCS: Fix ep configs race

brminich:
## What
Fixes a race for ep configs array.
Main thread may access some of ep configs, while async thread adding new element to ep config array. Since UCS array use realloc to extend the array, main thread may try to acces invalid memory. 

## How ?
Use calloc + memcpy + free for extending array (and call free on the main thread for ep configs array)

### DIFF
--- a/src/ucp/core/ucp_worker.c
+++ b/src/ucp/core/ucp_worker.c
@@ -2055,6 +2055,23 @@ ucp_worker_ep_config_short_init(ucp_worker_h worker, ucp_ep_config_t *ep_config,
     max_eager_short->memtype_on  = proto_short.max_length_host_mem;
 }
 
+static unsigned ucp_worker_ep_config_free_cb(void *arg)
+{
+    ucs_free(arg);
+    return 1;
+}
+
+static int
+ucp_worker_ep_config_filter(const ucs_callbackq_elem_t *elem, void *arg)
+{
+    if (elem->cb == ucp_worker_ep_config_free_cb) {
+        ucp_worker_ep_config_free_cb(elem->arg);
+        return 1;
+    }
+
+    return 0;
+}
+
 /* All the ucp endpoints will share the configurations. No need for every ep to
  * have its own configuration (to save memory footprint). Same config can be used
  * by different eps.
@@ -2072,6 +2089,7 @@ ucs_status_t ucp_worker_get_ep_config(ucp_worker_h worker,
     ucp_memtype_thresh_t *tag_max_short;
     ucp_lane_index_t tag_exp_lane;
     unsigned tag_proto_flags;
+    void *old_ep_cfg_buf;
     ucs_status_t status;
 
     ucs_assertv_always(key->num_lanes > 0,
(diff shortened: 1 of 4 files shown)


src/ucs/datastruct/array.c
```diff
@@ -13,31 +13,51 @@
...
 #define UCS_ARRAY_GROW_FACTOR   2
 
 
-ucs_status_t ucs_array_grow(void **buffer_p, size_t *capacity_p,
-                            size_t min_capacity, size_t value_size,
-                            const char *array_name, const char *value_name)
+ucs_status_t
+ucs_array_grow(void **buffer_p, size_t *capacity_p, size_t min_capacity,
+               size_t value_size, ucs_array_free_callback_t free_cb,
+               void *free_cb_arg, const char *array_name,
```
yosefe, line 26: maybe simpler to return the released pointer instead of defining a callback?
if nothing was released - return NULL

src/ucs/datastruct/array.c
```diff
@@ -13,31 +13,51 @@
...
 
-    new_capacity     = ucs_max(*capacity_p * UCS_ARRAY_GROW_FACTOR, min_capacity);
+    current_capacity = *capacity_p;
+    new_capacity     = ucs_max(current_capacity * UCS_ARRAY_GROW_FACTOR,
+                               min_capacity);
     aligned_capacity = (new_capacity + ~UCS_ARRAY_CAP_MASK) & UCS_ARRAY_CAP_MASK;
 
-    new_buffer       = ucs_realloc(*buffer_p, value_size * aligned_capacity,
-                                   array_name);
+    new_buffer       = ucs_calloc(aligned_capacity, value_size, array_name);
```
yosefe, line 37: why not malloc?

src/ucs/datastruct/array.c
```diff
@@ -13,31 +13,51 @@
...
+    if (free_cb != NULL) {
+        free_cb(*buffer_p, free_cb_arg);
+    } else {
+        ucs_free(*buffer_p);
+    }
+
+out:
     *buffer_p   = new_buffer;
     *capacity_p = aligned_capacity;
+
```
yosefe, line 61: revert?

src/ucs/datastruct/array.inl
```diff
@@ -23,52 +23,58 @@
...
+        old_buffer = (void*)array->buffer; \
+        status     = ucs_array_grow((void**)&array->buffer, &capacity, min_capacity, \
+                                    sizeof(_value_type), UCS_PP_MAKE_STRING(_name), \
+                                    UCS_PP_MAKE_STRING(_value_type)); \
         if (status != UCS_OK) { \
             return status; \
         } \
         \
         array->capacity = capacity; \
+        if (old_buffer_p != NULL) { \
```
yosefe, line 46: maybe invert if/else?

src/ucp/core/ucp_worker.c
```diff
@@ -2118,10 +2111,19 @@ ucs_status_t ucp_worker_get_ep_config(ucp_worker_h worker,
...
-                                      ucp_worker_ep_config_array_free_cb, worker,
-                                      return UCS_ERR_NO_MEMORY);
-    status    = ucp_ep_config_init(worker, ep_config, key);
+    old_ep_cfg_buf = NULL;
+    ep_config      = ucs_array_append_safe(ep_config_arr, &worker->ep_config,
+                                           &old_ep_cfg_buf,
+                                           return UCS_ERR_NO_MEMORY);
+    if (old_ep_cfg_buf != NULL) {
+        /* Schedule release of old ep configs array backing buffer on the main
+         * thread (this func can be called by async thread).
```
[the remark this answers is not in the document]
brminich: can you elaborate? Just add this to the end of comment? (does not look clear if yes)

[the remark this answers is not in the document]
yosefe: yes, add it to the end of the comment, to explain why returning a pointer to the old buffer makes it thread-safe
usually when thinking about thread-safe data struct it means it can be freely used from multiple threads at the same time, using some kind of sync/lock, and this is not the case here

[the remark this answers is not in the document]
brminich:
```
Schedule release of old ep configs array backing buffer on the main thread (this func can be called by async thread).
So the main thread can still access the old configuration buffer if it's modified by async thread.
```

yosefe (approved):

PR #9359, opened 2023-09-11
Reverted UCT/CUDA/CUDA_COPY: Set device context within mem alloc.

rakhmets:
## What
Reverted https://github.com/openucx/ucx/pull/9346

## Why ?
Please find reasons here https://github.com/openucx/ucx/pull/9346#issuecomment-1714347976

### DIFF
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.c
@@ -202,47 +202,13 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_dereg,
     return UCS_OK;
 }
 
-static ucs_status_t uct_cuda_copy_md_set_ctx(uct_cuda_copy_md_t *md)
+static ucs_status_t uct_cuda_copy_mem_alloc(uct_md_h md, size_t *length_p,
+                                            void **address_p,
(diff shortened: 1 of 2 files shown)


yosefe (commented):
can't it lead to memory leak ?

rakhmets:
> can't it lead to memory leak ?

It shouldn't. The primary context is not created by the UCX. It is only retained. cuDevicePrimaryCtxRelease is a safe way to reset the primary context (cuDevicePrimaryCtxReset). But we do not need to do this inside UCX.

Even calling for cudaDeviceReset is not mandatory by the way.
( cudaDeviceReset behaves as cuDevicePrimaryCtxReset, if the primary context is a current context for the calling thread. )

yosefe:
@rakhmets According to Cuda doc, we should release the primary context:
["A retained context should always be released once the user is done using it."](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__PRIMARY__CTX.html#group__CUDA__PRIMARY__CTX_1gf2a8bc16f8df0c88031f6a1ba3d6e8ad)
Maybe it would help if cuda_ctx_retained flag would be per thread?

yosefe (approved):

yosefe:
@roiedanino pls note that now we need to set cuda context from perftest

PR #9360, opened 2023-09-11
UCT/BASE: Removed code duplication related to dummy mem reg.

rakhmets:
## What
Removed duplicate code related to dummy memory registration.


src/uct/base/uct_md.c
```diff
@@ -592,6 +592,10 @@ ucs_status_t uct_md_dummy_mem_reg(uct_md_h md, void *address, size_t length,
                                   const uct_md_mem_reg_params_t *params,
                                   uct_mem_h *memh_p)
 {
+    /* For testing we have to make sure that
+     * memh_h != UCT_MEM_HANDLE_NULL
```
yosefe, line 596: what is memh_h?
maybe ucs_assert(memh_p != NULL)?

rakhmets: Changed to memh_p. This comment was copied from uct_cma_mem_reg.

yosefe: maybe it's not relevant anymore? can you check what happens if we remove the static assert?

PR #9366, opened 2023-09-12
UCT/ROCM: Add parameter to control max elements in sigpool

nileshnegi:
## What
Adds parameter (UCX_ROCM_COPY_SIGPOOL_MAX_ELEMS) to control the maximum number of elements in hsa signal pool (ucs_mpool_params). Changes from 1024 (default) to parameter.

## Why ?
Running the osu_bw benchmark beyond the standard 4MB message length for D-H communication leads to an error due to the uct/rocm_copy component running out of pre-allocated hsa-signals. Increasing the number of pre-allocated signals fixes the problem. This PR introduces a parameter for controlling the number of elements in the signal pool.

## How ?
_It is optional but for complex PRs please provide information about the design,
architecture, approach, etc._

@edgargabriel

### DIFF
--- a/src/uct/rocm/copy/rocm_copy_ep.c
+++ b/src/uct/rocm/copy/rocm_copy_ep.c
@@ -196,6 +196,12 @@ ucs_status_t uct_rocm_copy_ep_zcopy(uct_ep_h tl_ep, uint64_t remote_addr,
     }
 
     rocm_copy_signal = ucs_mpool_get(&iface->signal_pool);
+    if (rocm_copy_signal == NULL) {
+        ucs_error("increase the maximum number of signal pool elements with "
+                  "UCX_ROCM_COPY_SIGPOOL_MAX_ELEMS");
+        return UCS_ERR_IO_ERROR;
+    }
+
     hsa_signal_store_screlease(rocm_copy_signal->signal, 1);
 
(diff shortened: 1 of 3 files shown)


src/uct/rocm/copy/rocm_copy_ep.c
```diff
@@ -196,6 +196,12 @@ ucs_status_t uct_rocm_copy_ep_zcopy(uct_ep_h tl_ep, uint64_t remote_addr,
     }
 
     rocm_copy_signal = ucs_mpool_get(&iface->signal_pool);
+    if (rocm_copy_signal == NULL) {
+        ucs_error("Increase the maximum number of signal pool elements with "
```
yosefe, line 200: small case

nileshnegi: done, thanks!

src/uct/rocm/copy/rocm_copy_iface.h
```diff
@@ -26,6 +26,7 @@ typedef struct uct_rocm_copy_iface {
         size_t                  h2d_thresh;
         int                     enable_async_zcopy;
         double                  latency;
+        unsigned                sigpool_max_elems;
```
yosefe, line 29: seems it's unused

nileshnegi: done, thanks!

yosefe (approved):

PR #9371, opened 2023-09-13
BINDINGS/GO: Set CUDA device before allocating CUDA memory.

rakhmets:
## What
Set CUDA device before allocating CUDA memory using UCX.

### DIFF
--- a/bindings/go/src/examples/perftest/perftest.go
+++ b/bindings/go/src/examples/perftest/perftest.go
@@ -16,6 +16,7 @@ import (
 	"time"
 	. "ucx"
 	"unsafe"
+	. "cuda"
 )
 
 type PerfTestParams struct {
@@ -89,9 +90,16 @@ func initContext() {
 	perfTest.context, _ = NewUcpContext(params)
 }
 
+func tryCudaSetDevice() {
+	if perfTestParams.memType == UCS_MEMORY_TYPE_CUDA {
+		if ret := CudaSetDevice(); ret != nil {
+			panic(ret)
+		}
+	}
+}
+
 func initMemory() error {
 	var err error
-	var dummyMemh *UcpMemory
 	memTypeMask, _ := perfTest.context.MemoryTypesMask()
 
 	if !IsMemTypeSupported(perfTestParams.memType, memTypeMask) {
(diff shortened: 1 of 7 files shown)


bindings/go/src/examples/perftest/perftest.go
```diff
@@ -232,6 +230,10 @@ func serverAmRecvHandler(header unsafe.Pointer, headerSize uint64, data *UcpAmDa
 	if data.IsDataValid() {
 		atomic.AddUint32(&perfTest.numCompletedRequests, 1)
 	} else {
+		if perfTestParams.memType == UCS_MEMORY_TYPE_CUDA && !CudaCheckCurrentContext() {
+			CudaSetDevice();
```
ivankochin, line 234: minor: WDYT about checking `CudaCheckCurrentContext` or `cuCtxGetDevice` during `CudaSetDevice` and do nothing if context (or device) is set?

ivankochin: I would like to check for `CudaSetDevice` result and call `panic` in case of error (here and inside `serverStart`). Or even always do panic in case if `CudaSetDevice` fails.

rakhmets:
> I understand that currently the issue with perftest is on server side only, but I think it worth to set CUDA device for `clientThreadDoIter` too to avoid errors in the future (e.g. in case of using `rndv/put/zcopy`).

Done.

bindings/go/src/examples/perftest/perftest.go
rakhmets: Done.

bindings/go/src/examples/perftest/perftest.go
```diff
@@ -91,26 +92,23 @@ func initContext() {
...
 
 	mmapParams := &UcpMmapParams{}
-	// Allocate dummy host memory, to initialize cuda context if the memType is Cuda.
-	mmapParams.SetMemoryType(UCS_MEMORY_TYPE_HOST).Allocate()
-	mmapParams.SetLength(1)
-	dummyMemh, err = perfTest.context.MemMap(mmapParams)
-	if err != nil {
-		return err
-	}
-	dummyMemh.Close()
```
ivankochin, line 102: Minor: remove blank line

ivankochin (dismissed):

bindings/go/src/examples/perftest/perftest.go
rakhmets: Removed

bindings/go/src/examples/perftest/perftest.go
```diff
@@ -280,6 +289,12 @@ func clientThreadDoIter(i int, t uint) {
 	var request *UcpRequest
 	requestParams := (&UcpRequestParams{}).SetMemType(perfTestParams.memType)
 
+	if perfTestParams.memType == UCS_MEMORY_TYPE_CUDA {
+		if ret := CudaSetDevice(); ret != nil {
+			panic(ret)
+		}
+	}
```
yosefe, line 296: maybe wrap this in a function? seems like code duplication..

rakhmets: added helper function.

yosefe (approved):

PR #9375, opened 2023-09-19
BINDINGS/GO: Change perftest server completions counting logic

ivankochin:
## What
Changes the counting of completed requests inside the go perftest logic to avoid hanging on the server side.

## Why ?
`progressWorker` call the `ucp_worker_progress` until it returns `0`. So during one `progressWorker` call we can increase `perfTest.numCompletedRequests` several times, but after end of the `WaitGroup.Wait()` we increase `i` only by `1`. So we can miss some completed requests if they were handled during one `progressWorker` call.

Also, the server side didn't take warmup iterations into attention.

## How ?
Increase `i` by number of completed requests

### DIFF
--- a/bindings/go/src/examples/perftest/perftest.go
+++ b/bindings/go/src/examples/perftest/perftest.go
@@ -258,21 +258,18 @@ func serverStart() error {
 		perfTest.perThreadWorkers[t+1].SetAmRecvHandler(t, UCP_AM_FLAG_WHOLE_MSG, serverAmRecvHandler)
 	}
 
-	for i := uint(0); i < perfTestParams.numIterations; i += 1 {
-		perfTest.numCompletedRequests = 0
-
-		perfTest.wg.Add(int(perfTestParams.numThreads + 1))
-		for t := uint(0); t < perfTestParams.numThreads+1; t += 1 {
-			go func(tid uint) {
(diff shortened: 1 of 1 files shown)


bindings/go/src/examples/perftest/perftest.go
```diff
@@ -272,6 +273,7 @@ func serverStart() error {
 		}
 
 		perfTest.wg.Wait()
+		i += perfTest.numCompletedRequests
```
yosefe, line 276: but now we increase i by "numThreads" on each iteration, seems wrong

ivankochin: `perfTest.numCompletedRequests` isn't guaranteed to be `numThreads` each iteration since we do `ucp_worker_progress()` on each thread until it returns `0`.

And even if `perfTest.numCompletedRequests` equal to `numThreads`, `commonCompRequests` is the number of all completions for all threads.

yosefe: 1. can we launch all threads once, and wait until perfTest.numCompletedRequests reaches commonCompRequests?
2. remove the variable I
3. rename commonCompRequests to totalNumRequests

yosefe (approved):
pls squash

PR #9376, opened 2023-09-19
BINDINGS/GO: Improve perftest error detection

ivankochin:
## What
Add panic failures in case of errors.

## Why ?
Currently, Golang perftest client can end successfully even if endpoint error happened and all the requests returns error status.

### DIFF
--- a/bindings/go/src/examples/perftest/perftest.go
+++ b/bindings/go/src/examples/perftest/perftest.go
@@ -136,7 +136,8 @@ func initWorker(i int) {
 
 func epErrorHandling(ep *UcpEp, status UcsStatus) {
 	if status != UCS_ERR_CONNECTION_RESET {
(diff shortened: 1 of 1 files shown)


bindings/go/src/examples/perftest/perftest.go
```diff
@@ -288,6 +289,10 @@ func clientThreadDoIter(i int, t uint) {
 	for request.GetStatus() == UCS_INPROGRESS {
 		progressWorker(int(t))
 	}
+	for request.GetStatus() != UCS_OK {
```
yosefe, line 292: I think it should be "if" and not "for"

ivankochin, line 292:
```suggestion
	if request.GetStatus() != UCS_OK {
```

PR #9377, opened 2023-09-19
BINDINGS/GO: Set error status for failed requests

ivankochin:
## What
Set error status for non-pointer failed requests 

## Why ?
Currently we return `error` but don't update the status field inside returned request.

### DIFF
--- a/bindings/go/src/ucx/request.go
+++ b/bindings/go/src/ucx/request.go
@@ -54,19 +54,19 @@ func NewRequest(request C.ucs_status_ptr_t, callbackId uint64, immidiateInfo int
 		ucpRequest.request = unsafe.Pointer(uintptr(request))
 		ucpRequest.Status = UCS_INPROGRESS
 	} else {
-		requestStatus := UcsStatus(int64(uintptr(request)))
+		ucpRequest.Status = UcsStatus(int64(uintptr(request)))
 		if callback, found := deregister(callbackId); found {
 			switch callback := callback.(type) {
 			case UcpSendCallback:
-				callback(ucpRequest, requestStatus)
+				callback(ucpRequest, ucpRequest.Status)
 			case UcpTagRecvCallback:
-				callback(ucpRequest, requestStatus, immidiateInfo.(*UcpTagRecvInfo))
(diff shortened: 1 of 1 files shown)


bindings/go/src/ucx/request.go
```diff
@@ -64,9 +64,10 @@ func NewRequest(request C.ucs_status_ptr_t, callbackId uint64, immidiateInfo int
 			case UcpAmDataRecvCallback:
 				callback(ucpRequest, requestStatus, uint64(immidiateInfo.(C.size_t)))
 			}
-			if requestStatus != C.UCS_OK {
-				return ucpRequest, NewUcxError(requestStatus)
-			}
+		}
+		if requestStatus != UCS_OK {
+			ucpRequest.Status = requestStatus
```
brminich, line 69: do we need to set status if it's `UCS_OK` or it is a default value?

ivankochin: `UCS_OK` is equal to 0, so this is default one that is assigned during the creation

brminich: ok, but i'd not rely on the fact that UCS_OK is zero
imo, would be cleaner to always initialize status, but it's up to you

ivankochin, line 69:
```suggestion
		ucpRequest.Status = requestStatus
		if requestStatus != UCS_OK {
```

bindings/go/src/ucx/request.go
```diff
@@ -64,9 +64,10 @@ func NewRequest(request C.ucs_status_ptr_t, callbackId uint64, immidiateInfo int
 			case UcpAmDataRecvCallback:
 				callback(ucpRequest, requestStatus, uint64(immidiateInfo.(C.size_t)))
 			}
-			if requestStatus != C.UCS_OK {
-				return ucpRequest, NewUcxError(requestStatus)
-			}
+		}
+		ucpRequest.Status = requestStatus
+		if requestStatus != UCS_OK {
```
yosefe, line 69: can we remove requestStatus variable altogether, and always use ucpRequest.Status instead?

yosefe (approved):

PR #9385, opened 2023-09-22
UCP/CORE: Usage Tracker is used by ucp_worker

shasson5:
## What
Usage Tracker is used by ucp_worker

## Why ?
Prepare final stage of RC/DC Support


src/ucp/core/ucp_worker.c
```diff
@@ -39,7 +39,7 @@
 #include <time.h>
 
 
-#define UCP_WORKER_KEEPALIVE_ITER_SKIP 32
+#define UCP_WORKER_SLOW_CB_ITER_SKIP 32
```
yosefe, line 42: UCP_WORKER_PROGRESS_TIMER_SKIP_COUNT

src/ucp/core/ucp_worker.c
```diff
@@ -2385,6 +2385,100 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+
+    if (!worker->context->config.ext.usage_tracker_enable) {
+        return UCS_OK;
+    }
+
+    params.promote_capacity = UCS_USAGE_TRACKER_PROMOTE_CAPACITY;
+    params.promote_thresh   = UCS_USAGE_TRACKER_PROMOTE_THRESH;
+    params.remove_thresh    = UCS_USAGE_TRACKER_REMOVE_THRESH;
+    params.exp_decay.m      = UCS_USAGE_TRACKER_EXP_DECAY_MULTIPLIER;
+    params.exp_decay.c      = UCS_USAGE_TRACKER_EXP_DECAY_CONST;
```
yosefe, line 2449: IMO no need to define constants since we use them only once

src/ucp/core/ucp_worker.c
```diff
@@ -2385,6 +2385,100 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+
+    ucs_usage_tracker_touch_key(worker->usage_tracker.handle, ep);
+    worker->usage_tracker.samples_count++;
+
+    if ((worker->usage_tracker.samples_count %
+         UCS_USAGE_TRACKER_SAMPLES_COUNT_PER_RUN) == 0) {
+        worker->usage_tracker.running    = 0;
+        worker->usage_tracker.last_round = ucs_get_time();
+
+        uct_worker_progress_register_safe(worker->uct,
```
yosefe, line 2429: maybe progress the usage tracker directly from here, instead of registering a progress callback?

shasson5: this code is fast path, shouldn't it bail out as soon as possible?

src/ucp/core/ucp_request.c
```diff
@@ -754,6 +754,8 @@ ucs_status_t ucp_request_progress_wrapper(uct_pending_req_t *self)
     } else {
         ucp_trace_req(req, "progress protocol %s returned: %s", proto->name,
                       ucs_status_string(status));
+
+        ucp_worker_track_ep_usage(req->send.ep->worker, req->send.ep);
```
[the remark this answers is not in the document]
shasson5: is there a field in ucp_request which indicate fragment num? I see "offset" in multiple places.

src/ucp/core/ucp_worker.c
```diff
@@ -2385,6 +2385,75 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+    ucs_usage_tracker_touch_key(worker->usage_tracker.handle, req->send.ep);
+    worker->usage_tracker.samples_count++;
+
+    if ((worker->usage_tracker.samples_count % 100000) == 0) {
+        ucs_usage_tracker_progress(worker->usage_tracker.handle);
+        worker->usage_tracker.last_round = ucs_get_time();
+    }
+}
+
+static ucs_status_t ucp_worker_create_usage_tracker(ucp_worker_h worker)
```
gleon99, line 2416: `usage_tracker_create`

src/ucp/core/ucp_worker.h
```diff
@@ -361,6 +362,14 @@ typedef struct ucp_worker {
         /* Number of failed endpoints */
         uint64_t                     ep_failures;
     } counters;
+
+    struct {
+        ucs_usage_tracker_h          handle;
```
gleon99, line 367: Missing docs

src/ucp/core/ucp_worker.c
```diff
@@ -2385,6 +2385,75 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+        return status;
+    }
+
+    worker->usage_tracker.iter_count    = 0;
+    worker->usage_tracker.samples_count = 0;
+    worker->usage_tracker.last_round    = ucs_get_time();
+    return UCS_OK;
+}
+
+static void ucp_worker_destroy_usage_tracker(ucp_worker_h worker)
```
gleon99, line 2446: `usage_tracker_destroy`

src/ucp/core/ucp_worker.c
```diff
@@ -2385,6 +2385,75 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+{
+    if (!worker->context->config.ext.usage_tracker_enable) {
+        return;
+    }
+
+    ucs_usage_tracker_destroy(worker->usage_tracker.handle);
+    uct_worker_progress_unregister_safe(worker->uct,
```
gleon99, line 2453: Where is the call to `register`?

src/ucp/core/ucp_worker.c
```diff
@@ -2385,6 +2385,99 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
                             ucs_min(max_am_header, UINT32_MAX) : 0ul;
 }
 
+static unsigned ucp_worker_progress_usage_tracker(void *arg)
```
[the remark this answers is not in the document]
shasson5: was removed in the last commit

src/ucp/core/ucp_worker.c
gleon99: Unresolved

src/ucp/core/ucp_context.c
```diff
@@ -392,6 +392,15 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "(inf - check all endpoints on every round, must be greater than 0)",
    ucs_offsetof(ucp_context_config_t, keepalive_num_eps), UCS_CONFIG_TYPE_UINT},
 
+  {"DYNAMIC_TL_SWITCH_ENABLE", "n", "Enable dynamic transport switching",
+   ucs_offsetof(ucp_context_config_t, dynamic_tl_switch_enable), UCS_CONFIG_TYPE_BOOL},
+
+  {"DYNAMIC_TL_SWITCH_INTERVAL", "3s",
```
gleon99, line 398: Can unite these 2 params, see other places with intervals and special value for "disabled".

src/ucp/core/ucp_context.h
```diff
@@ -136,6 +136,10 @@ typedef struct ucp_context_config {
     /** Maximal number of endpoints to check on every keepalive round
      * (0 - disabled, inf - check all endpoints on every round) */
     unsigned                               keepalive_num_eps;
+    /** Enable dynamic transport switching */
+    int                                    dynamic_tl_switch_enable;
+    /** Time period between dynamic transport switching rounds */
+    ucs_time_t                             dynamic_tl_switch_interval;
```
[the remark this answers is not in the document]
shasson5: for keep_alive "0" is error and UCS_TIME_INFINITY is disabled.

src/ucp/core/ucp_worker.c
```diff
@@ -2387,38 +2387,41 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
 
 void ucp_worker_track_ep_usage(ucp_request_t *req)
 {
-    ucp_worker_h worker = req->send.ep->worker;
+    static unsigned sample_count_per_round = 100000;
```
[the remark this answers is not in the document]
shasson5: this const is intended to define how many samples we take in each round, how does 'skip' fit here?

src/ucp/core/ucp_worker.c
shasson5: I already removed this function call. why unresolved?

src/ucp/core/ucp_worker.c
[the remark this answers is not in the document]
shasson5: > And maybe make it `#define` as well.

why better?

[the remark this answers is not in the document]
gleon99: > this const is intended to define how many samples we take in each round, how does 'skip' fit here?

Ok.

But still, why make it a variable and not a `#define`?

src/ucp/core/ucp_worker.c
```diff
@@ -2414,6 +2414,78 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+    params.promote_thresh   = 10;
+    params.remove_thresh    = 0.2;
+    params.exp_decay.m      = 0.8;
+    params.exp_decay.c      = 0.2;
+    params.promote_cb       =
+            (ucs_usage_tracker_elem_update_cb_t)ucs_empty_function;
+    params.demote_cb        =
+            (ucs_usage_tracker_elem_update_cb_t)ucs_empty_function;
+
+    status = ucs_usage_tracker_create(&params, &worker->usage_tracker.handle);
```
gleon99, line 2468: Maybe have a var for `handle`, and only copy to the `worker` upon success?

src/ucp/core/ucp_worker.h
```diff
@@ -362,6 +363,14 @@ typedef struct ucp_worker {
         /* Number of failed endpoints */
         uint64_t                     ep_failures;
     } counters;
+
+    struct {
+        ucs_usage_tracker_h          handle;              /* Usage tracker handle */
+        unsigned                     iter_count;          /* Number of progress iterations to skip,
+                                                           * used to minimize call of ucs_get_time */
+        unsigned                     sample_count;        /* Number of samples taken in each round */
```
gleon99, line 371: `samples to take`

src/ucp/core/ucp_worker.c
```diff
@@ -2414,6 +2414,78 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+    }
+
+    req->flags |= UCP_REQUEST_FLAG_USAGE_TRACKED;
+    ucs_usage_tracker_touch_key(worker->usage_tracker.handle, req->send.ep);
+    worker->usage_tracker.sample_count++;
+
+    if (ucs_unlikely((worker->usage_tracker.sample_count %
+                      sample_count_per_round) == 0)) {
+        ucs_usage_tracker_progress(worker->usage_tracker.handle);
+        worker->usage_tracker.last_round = ucs_get_time();
```
gleon99, line 2444: Put `ucs_get_time()` result in a var?

src/ucp/core/ucp_worker.c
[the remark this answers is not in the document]
shasson5: for better performance?

src/ucp/core/ucp_worker.c
```diff
@@ -2414,6 +2414,78 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+{
+    ucs_usage_tracker_params_t params = {0};
+    ucs_status_t status;
+
+    if (worker->context->config.ext.dynamic_tl_switch_interval ==
+        UCS_TIME_INFINITY) {
+        return UCS_OK;
+    }
+
+    params.promote_capacity = 20;
```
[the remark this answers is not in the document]
shasson5: documentation already exists in usage_tracker.h

src/ucp/tag/tag_send.c
```diff
@@ -153,6 +153,10 @@ ucp_tag_send_inline(ucp_ep_h ep, const void *buffer, size_t length,
 {
     ucs_status_t status;
 
+    if (ucs_unlikely(ucp_proto_is_progress_wrapper_enabled(ep->worker))) {
```
yosefe, line 156: this disables fast-path short by default, can you suggest another approach?

shasson5:
@gleon99 @yosefe can review?

src/ucp/core/ucp_worker.c
```diff
@@ -2419,6 +2428,87 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+    }
+
+    req->flags |= UCP_REQUEST_FLAG_USAGE_TRACKED;
+    ucs_usage_tracker_touch_key(worker->usage_tracker.handle, req->send.ep);
+    worker->usage_tracker.sample_count++;
+
+    if (ucs_unlikely((worker->usage_tracker.sample_count %
+                      sample_count_per_round) == 0)) {
+        ucs_usage_tracker_progress(worker->usage_tracker.handle);
+        worker->usage_tracker.last_round = now;
```
yosefe, line 2461: shouldn't we do it outside the if condition?

src/ucp/core/ucp_worker.h
```diff
@@ -364,6 +365,17 @@ typedef struct ucp_worker {
...
+    struct {
+        ucs_usage_tracker_h          handle;              /* Usage tracker handle */
+        unsigned                     iter_count;          /* Number of progress iterations to skip,
+                                                           * used to minimize calls to ucs_get_time */
+        unsigned                     sample_count;        /* Number of samples to take in each round */
+        ucs_time_t                   last_round;          /* Last round timestamp */
+    } usage_tracker;
+    
+    /* Progress wrapper enabled */
+    int                              progress_wrapper_enabled;
```
yosefe, line 378: can use worker->flags (or better move it to ucp_context)

src/ucp/core/ucp_worker.c
shasson5: you want to take 100K samples in each round

src/ucp/core/ucp_worker.c
```diff
@@ -2419,6 +2428,87 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+
+    if (ucs_likely((req->send.ep->worker->usage_tracker.iter_count++ %
+                    UCP_WORKER_PROGRESS_TIMER_SKIP_COUNT) != 0)) {
+        return;
+    }
+
+    worker = req->send.ep->worker;
+    now    = ucs_get_time();
+
+    if (ucs_likely((now - worker->usage_tracker.last_round) <
```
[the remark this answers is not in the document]
shasson5: there are 3 brackets:
1) `if` block
2) left side of `<`
3) ucs_likely block

src/ucp/core/ucp_worker.c
```diff
@@ -2419,6 +2428,87 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
+void ucp_worker_track_ep_usage(ucp_request_t *req)
+{
+    static unsigned sample_count_per_round = 100000;
+    ucp_worker_h worker;
+    ucs_time_t now;
+
+    if (ucs_likely((req->send.ep->worker->usage_tracker.iter_count++ %
+                    UCP_WORKER_PROGRESS_TIMER_SKIP_COUNT) != 0)) {
+        return;
+    }
```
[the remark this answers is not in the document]
shasson5: also why not turn `ucp_worker_track_ep_usage` to be inline instead?

src/ucp/core/ucp_context.c
```diff
@@ -419,6 +419,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, dynamic_tl_switch_interval),
    UCS_CONFIG_TYPE_TIME_UNITS},
 
+  {"DYNAMIC_TL_SAMPLES_PER_ROUND", "100000",
+    "Number of usage samples taken every transport switching round. Must be\n"
```
yosefe, line 423: indent on column

src/ucp/core/ucp_context.c
```diff
@@ -1918,6 +1930,28 @@ static double ucp_context_get_memcpy_bw()
...
 }
 
+static int
+ucp_dynamic_tl_switch_config_valid(const ucp_context_config_t *config)
+{
+    const char *err_str;
+
+    if (config->dynamic_tl_switch_interval == 0) {
+        err_str = "UCX_DYNAMIC_TL_SWITCH_INTERVAL";
+        goto err;
```
yosefe, line 1940: IMO it would be simpler to print error message and exit, without goto
```c
    if (config->dynamic_tl_switch_interval == 0) {
        ucs_error("UCX_DYNAMIC_TL_SWITCH_INTERVAL must be > 0");
        return 0;
    }
```

src/ucp/core/ucp_request.c
```diff
@@ -746,6 +746,11 @@ ucs_status_t ucp_request_progress_wrapper(uct_pending_req_t *self)
                   req->send.state.dt_iter.length);
 
     ucs_log_indent(1);
+
+    if (ucs_unlikely(ucp_worker_should_track_usage(req->send.ep->worker))) {
+        ucp_worker_track_ep_usage(req);
+    }
```
yosefe, line 752: maybe unite it to one inline function? the inline part will check TIMER_SKIP and USAGE_TRACEKED flag. Probably also need to check if usage tracker is enabled.
The non-inline part will do all the rest.

shasson5: what is the logic behind this division? why not make all the function inline?

src/ucp/core/ucp_worker.inl
```diff
@@ -240,11 +240,17 @@ ucp_worker_default_address_pack_flags(ucp_worker_h worker)
...
-ucp_worker_should_track_usage(ucp_worker_h worker)
+static UCS_F_ALWAYS_INLINE void ucp_worker_track_ep_usage(ucp_request_t *req)
 {
-    return (worker->usage_tracker.iter_count++ %
-            UCP_WORKER_PROGRESS_TIMER_SKIP_COUNT) == 0;
+    ucp_worker_h worker = req->send.ep->worker;
+
+    if (ucs_likely(!ucp_context_usage_tracker_enabled(worker->context) ||
+                   ((worker->usage_tracker.iter_count++ %
+                     UCP_WORKER_PROGRESS_TIMER_SKIP_COUNT) == 0))) {
```
yosefe, line 249: should be != 0

yosefe (approved):
pls squash

src/ucp/core/ucp_context.c
```diff
@@ -2114,6 +2147,9 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
              ((context->config.ext.max_rma_lanes > 1) ||
               context->config.ext.proto_enable));
 
+    context->config.progress_wrapper_enabled =
+            ucs_log_is_enabled(UCS_LOG_LEVEL_TRACE_REQ) ||
+            ucp_context_usage_tracker_enabled(context);
```
ivankochin, line 2152: This config field causes short proto disabling short if `UCS_LOG_LEVEL_TRACE_REQ` log level is enabled. I don't think that make proto selection dependent on log level is a good idea.

E.g. for protov1 setting `UCX_LOG_LEVEL=req` is one of the simplest ways to understand what procol was chosen for MPI calls. 

Can we avoid that somehow?

PR #9386, opened 2023-09-24
AZP: Extend perf testing

Alexey-Rivkin:
## What
Extend UCX perf pipeline to include also GPU, intra-node, and osu_mbw_mr tests.

[Successful test run](https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=70200&view=logs&j=94be751c-9e4c-5c28-d13e-4c720246b05d&t=91332361-356c-5009-c13f-158bec505cbf&l=70)

### DIFF
--- a/buildlib/azure-pipelines-perf.yml
+++ b/buildlib/azure-pipelines-perf.yml
@@ -10,8 +10,6 @@ pr:
 variables:
 - name: WorkDir
   value: /hpc/scrap/azure/$(Build.DefinitionName)/$(Build.BuildId)-$(Build.BuildNumber)
-- name: perf_nodes
-  value: vulcan03,vulcan04
 - name: threshold
   value: 5
 
@@ -53,6 +51,7 @@ stages:
     jobs:
       - job: Perf
         displayName: Performance testing
+        timeoutInMinutes: 180
         pool:
           name: MLNX
           demands:
@@ -85,32 +84,19 @@ stages:
             displayName: Debug info
             workingDirectory: $(WorkDir)
 
-          - bash: |
-              set -eE
-              # Extract commit SHAs from the PR's merge message.
-              # SHA_After=$(Build.SourceVersion)
-              SHA_After=$(echo $(Build.SourceVersionMessage) | awk '{print $2}')
-              SHA_Before=$(echo $(Build.SourceVersionMessage) | awk '{print $4}')
+          - template: tools/perf-common.yml
+            parameters:
+              Name: Build-UCX
 
-              python3 -m venv VENV
-              source VENV/bin/activate
-              cd $(WorkDir)/PerfX
-              pip3 install -q -r requirements.txt
-              module load /hpc/local/etc/modulefiles/hpcx-gcc
+          - template: tools/perf-common.yml
+            parameters:
+              Name: Perf-test-multi-node
 
-              set -xeE -o pipefail
-              python3 -u perfx.py \
-                --source-branch $(Build.SourceBranch) \
-                --config ucx-rdmz.yml \
-                --before "$SHA_Before" \
-                --after "$SHA_After" \
-                --module hpcx-gcc \
-                --filter \
-                --extra-ompi-opts " -H $(perf_nodes)" | tee $(WorkDir)/results.txt
-            displayName: Perf test
-            workingDirectory: $(WorkDir)
+          - template: tools/perf-common.yml
+            parameters:
+              Name: Perf-test-single-node
 
-          - script: $(WorkDir)/ucx/buildlib/tools/perf_results.py $(WorkDir)/results.txt $(threshold)
+          - script: $(WorkDir)/ucx/buildlib/tools/perf_results.py $(WorkDir)/results-Perf\*.txt $(threshold)
             displayName: Results analyzer
             workingDirectory: $(WorkDir)
 
@@ -119,6 +105,7 @@ stages:
               cat perfx.log
             displayName: Print raw log
             workingDirectory: $(WorkDir)/PerfX
+            condition: always()
 
           - task: PublishBuildArtifacts@1
             inputs:
--- a/buildlib/tools/perf-common.yml
+++ b/buildlib/tools/perf-common.yml
@@ -1,0 +1,44 @@
+parameters:
+  Name: ''
+
+steps:
+  - bash: |
+      set -eE -o pipefail
+      SHA_After=$(echo $(Build.SourceVersionMessage) | awk '{print $2}')
+      SHA_Before=$(echo $(Build.SourceVersionMessage) | awk '{print $4}')
+
+      module=""
+      perfxParams=""
+      perf_nodes="vulcan03:24,vulcan04:24"
+
+      case "${{ parameters.Name }}" in
+        "Build-UCX")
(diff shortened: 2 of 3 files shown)


Alexey-Rivkin:
Successful test run:
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=70188

yosefe:
@rakhmets @ivankochin can you pls review?

buildlib/azure-pipelines-perf.yml
```diff
@@ -106,11 +109,43 @@ stages:
...
+              # Pick random node
+              IFS=',' read -ra nodes <<< "$(perf_nodes)"
+              node=${nodes[RANDOM % 2]}
+
+              python3 -u perfx.py \
+                --skip-update \
+                --skip-build \
+                --use-cache \
+                --source-branch $(Build.SourceBranch) \
+                --config ucx-rdmz.yml \
```
rakhmets, line 137: I think that the config should be stored in the place where perfx should be used. Otherwise, we should update perfx repo, if we want to change configuration of UCX tests. Moreover, you should check perfx repo to find out testing parameter, e.g. the list of benchmarks.

I know that this comment is not related to this PR, because it was added earlier. However, it can be discussed and addressed in the future.

buildlib/azure-pipelines-perf.yml
```diff
@@ -106,11 +109,43 @@ stages:
...
+                --before "$SHA_Before" \
+                --after "$SHA_After" \
+                --module hpcx-gcc \
+                --filter \
+                --omb-cuda \
+                --setup rdmz-intra \
+                --extra-ompi-opts " -H $node,$node" | tee $(WorkDir)/results-singlenode.txt
+            displayName: Perf test - single node
+            workingDirectory: $(WorkDir)
```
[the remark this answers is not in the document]
Alexey-Rivkin: Not much in common between the two steps:
perfx.py flags differ significantly, modules, venv init - not the same all of them.

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: Adding the PerfX architect: @amaslenn

amaslenn: I agree with @rakhmets.

The only thing I worry about in this case is compatibility: I don't understand how Perfx is delivered to this pipeline and it might happen that it doesn't support new config option and maybe it dropped support for another.

Explicit git checkout with a tag should solve it though.

buildlib/azure-pipelines-perf.yml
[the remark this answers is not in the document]
ivankochin: Do I understand correctly that `--skip-build` flag is set because build happens during the first perfX execution?

Is it possible to create some "Initialization" step to download requirements and perform build to make the code for multi and single node mode common?

Also, is it possible to add CUDA tests for single-node too?

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: There is an explicit checkout:
```
resources:
  repositories:
  - repository: PerfX
    type: github
    name: Mellanox-lab/PerfX
    endpoint: Mellanox-lab
    ref: ucx_extend
```

buildlib/azure-pipelines-perf.yml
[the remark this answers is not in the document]
amaslenn: Perfx also has `--skip-run` :)
So should be possible to introduce such initialisation.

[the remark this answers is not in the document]
Alexey-Rivkin: `--skip-run` didn't go well when I tried it, but we can further debug it.

All tests get run in both multi-node and single-node modes.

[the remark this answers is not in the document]
ivankochin: Then it's strange that we do load `hpcx-gcc` module for both single and multi node runs, but `dev/cuda-latest` only for multi.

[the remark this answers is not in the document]
Alexey-Rivkin: `hpcx-gcc` module provides OMPI.
CUDA module is only needed for the build phase.

Alexey-Rivkin:
@rakhmets @ivankochin
Now we have a separate step to build UCX using PerfX first. Two more consequent steps run tests in multi-node and single-node modes.
[New working build](https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=70275&view=logs&j=94be751c-9e4c-5c28-d13e-4c720246b05d)

ivankochin:
As I can see, now we have more in common between single and multi node executions. The differences I see are picking the random node for single node case + `--setup rdmz-intra` option. So it should be simpler to move the common code to the separate method.

From my point of view, even choosing random node is not necessery. @rakhmets WDYT?

Alexey-Rivkin:
The YAML format provides limited options. Further streamlining the code might compromise its readability and debuggability.

Alexey-Rivkin:
Can merge both single-node and multi-node runs into a single step, so they share env setup. 
The disadvantage is that it will not be immediately clear in which mode the test is failing.

Alexey-Rivkin:
@rakhmets @ivankochin I stand corrected. A separate template with a bit of logic simplifies the main pipeline tremendously.

buildlib/azure-pipelines-perf.yml
```diff
@@ -11,7 +11,7 @@ variables:
 - name: WorkDir
   value: /hpc/scrap/azure/$(Build.DefinitionName)/$(Build.BuildId)-$(Build.BuildNumber)
 - name: perf_nodes
-  value: vulcan03,vulcan04
+  value: vulcan03:24,vulcan04:24
```
rakhmets, line 14: Can be moved to perf-common.yml.

rakhmets (approved):

buildlib/tools/perf-common.yml
```diff
@@ -0,0 +1,43 @@
...
+steps:
+  - bash: |
+      set -eE -o pipefail
+      SHA_After=$(echo $(Build.SourceVersionMessage) | awk '{print $2}')
+      SHA_Before=$(echo $(Build.SourceVersionMessage) | awk '{print $4}')
+
+      module=""
+      perfxParams=""
+
+      if [[ "${{ parameters.displayName }}" == "Build-UCX" ]]; then
```
ivankochin, line 14: Minor: Use switch-case expression instead of if-else

ivankochin (approved):

ivankochin (approved):

buildlib/tools/perf-common.yml
Alexey-Rivkin: Done

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: Can move it directly to Bash though, thanks!

ivankochin (approved):

amaslenn (approved):

yosefe (approved):

yosefe:
@Alexey-Rivkin pls squash

Alexey-Rivkin:
squashed

yosefe (approved):

PR #9399, opened 2023-10-03
UCP/PROTO: Eliminate copy and bandwidth overheads for empty messages

ivankochin:
## What
- Remove overheads that are applicable only for non-empty messages for 0 size estimations.
- Enable short protocol for non-host memory types on empty messages

## Why ?
To improve performance prediction model.

### DIFF
--- a/src/ucp/proto/proto_common.c
+++ b/src/ucp/proto/proto_common.c
@@ -415,21 +415,10 @@ static ucp_lane_index_t ucp_proto_common_find_lanes_internal(
               ucs_string_buffer_cstr(&sel_param_strb));
     ucs_log_indent(1);
 
-    if (flags & UCP_PROTO_COMMON_INIT_FLAG_HDR_ONLY) {
-        /* Skip send payload check */
-    } else if (flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY) {
-        if ((select_param->dt_class == UCP_DATATYPE_GENERIC)) {
-            /* Generic/IOV datatype cannot be used with zero-copy send */
-            ucs_trace("datatype %s cannot be used with zcopy",
-                      ucp_datatype_class_names[select_param->dt_class]);
-            goto out;
-        }
-    } else if (!UCP_MEM_IS_ACCESSIBLE_FROM_CPU(select_param->mem_type) &&
-               (memtype_op == UCT_EP_OP_LAST)) {
-        /* If zero-copy is off, the memory must be host-accessible for
-         * non-generic type (for generic type there is no buffer to access) */
-        ucs_trace("memory type %s with datatype %s is not supported",
-                  ucs_memory_type_names[select_param->mem_type],
+    if ((flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY) &&
+        (select_param->dt_class == UCP_DATATYPE_GENERIC)) {
+        /* Generic/IOV datatype cannot be used with zero-copy send */
+        ucs_trace("datatype %s cannot be used with zcopy",
                   ucp_datatype_class_names[select_param->dt_class]);
         goto out;
     }
@@ -661,9 +650,7 @@ void ucp_proto_request_zcopy_completion(uct_completion_t *self)
 
 int ucp_proto_is_short_supported(const ucp_proto_select_param_t *select_param)
 {
-    /* Short protocol requires contig/host */
-    return (select_param->dt_class == UCP_DATATYPE_CONTIG) &&
-           UCP_MEM_IS_HOST(select_param->mem_type);
+    return (select_param->dt_class == UCP_DATATYPE_CONTIG);
 }
 
 void ucp_proto_trace_selected(ucp_request_t *req, size_t msg_length)
--- a/src/ucp/proto/proto_init.c
+++ b/src/ucp/proto/proto_init.c
@@ -220,7 +220,8 @@ ucp_proto_init_parallel_stages(const ucp_proto_init_params_t *params,
             /* For multi-fragment protocols, we need to apply the fragment
              * size to the performance function linear factor.
              */
-            if (!(flags & UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG)) {
+            if (!(flags & UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG) &&
+                (frag_size > 0)) {
                 perf[perf_type].m += perf[perf_type].c / frag_size;
             }
         }
@@ -478,7 +479,7 @@ ucp_proto_init_buffer_copy_time(ucp_worker_h worker, const char *title,
 static ucs_status_t
 ucp_proto_common_init_send_perf(const ucp_proto_common_init_params_t *params,
                                 const ucp_proto_common_tl_perf_t *tl_perf,
-                                ucp_md_map_t reg_md_map,
+                                ucp_md_map_t reg_md_map, int empty_msg,
                                 ucp_proto_perf_range_t *send_perf)
 {
     ucp_proto_perf_node_t *child_perf_node;
@@ -497,7 +498,8 @@ ucp_proto_common_init_send_perf(const ucp_proto_common_init_params_t *params,
         ucp_proto_init_memreg_time(params, reg_md_map, &send_overhead,
                                    &child_perf_node);
         ucp_proto_perf_node_own_child(send_perf->node, &child_perf_node);
-    } else if (params->flags & UCP_PROTO_COMMON_INIT_FLAG_RKEY_PTR) {
+    } else if ((params->flags & UCP_PROTO_COMMON_INIT_FLAG_RKEY_PTR) ||
+               empty_msg) {
         send_overhead = UCS_LINEAR_FUNC_ZERO;
     } else {
         ucs_assert(reg_md_map == 0);
@@ -575,7 +577,7 @@ ucp_proto_common_init_xfer_perf(const ucp_proto_common_init_params_t *params,
 static ucs_status_t
 ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
                                 const ucp_proto_common_tl_perf_t *tl_perf,
-                                ucp_md_map_t reg_md_map,
+                                ucp_md_map_t reg_md_map, int empty_msg,
                                 ucp_proto_perf_range_t *recv_perf)
 {
     const ucp_proto_select_param_t *select_param = params->super.select_param;
@@ -593,7 +595,8 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
         (params->flags & UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS) ||
         /* Count only send completion time without waiting for a response */
         ((op_attr_mask & UCP_OP_ATTR_FLAG_FAST_CMPL) &&
-         !(params->flags & UCP_PROTO_COMMON_INIT_FLAG_RESPONSE))) {
+         !(params->flags & UCP_PROTO_COMMON_INIT_FLAG_RESPONSE)) ||
+        empty_msg) {
         recv_overhead = UCS_LINEAR_FUNC_ZERO;
     } else {
         if (params->flags & UCP_PROTO_COMMON_INIT_FLAG_RECV_ZCOPY) {
@@ -640,6 +643,69 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
     return UCS_OK;
 }
 
+static ucs_status_t
+ucp_proto_init_single_frag_ranges(const ucp_proto_common_init_params_t *params,
+                                  const ucp_proto_common_tl_perf_t *tl_perf,
+                                  ucp_proto_perf_node_t *const tl_perf_node,
+                                  ucp_md_map_t reg_md_map, size_t range_start,
+                                  size_t frag_size)
+{
+    int empty_msg = (frag_size == 0);
+    ucp_proto_perf_range_t xfer_perf, send_perf, recv_perf;
+    const ucp_proto_perf_range_t *parallel_stages[3];
+    ucs_status_t status;
+
+    /* Network transfer time */
+    ucp_proto_common_init_xfer_perf(params, tl_perf, tl_perf_node, &xfer_perf);
+
+    /* Sender overhead */
+    status = ucp_proto_common_init_send_perf(params, tl_perf, reg_md_map,
+                                             empty_msg, &send_perf);
+    if (status != UCS_OK) {
+        goto out_deref_xfer_perf;
+    }
+
+    /* Receiver overhead */
+    status = ucp_proto_common_init_recv_perf(params, tl_perf, reg_md_map,
+                                             empty_msg, &recv_perf);
+    if (status != UCS_OK) {
+        goto out_deref_send_perf;
+    }
+
+    parallel_stages[0] = &send_perf;
+    parallel_stages[1] = &xfer_perf;
+    parallel_stages[2] = &recv_perf;
+
+    /* Add ranges representing sending single fragment */
+    status = ucp_proto_init_parallel_stages(&params->super, range_start, frag_size,
+                                            frag_size, 0.0, parallel_stages, 3,
+                                            params->flags);
+
+    ucp_proto_perf_node_deref(&recv_perf.node);
+out_deref_send_perf:
+    ucp_proto_perf_node_deref(&send_perf.node);
+out_deref_xfer_perf:
+    ucp_proto_perf_node_deref(&xfer_perf.node);
+    return status;
+}
+
+static int
+ucp_proto_common_check_mem_access(const ucp_proto_common_init_params_t *params)
+{
+    uint8_t mem_type = params->super.select_param->mem_type;
+
+    /*
+     * - HDR_ONLY protocols don't need access to payload memory
+     * - ZCOPY protocols don't need to copy payload memory
+     * - CPU accessible memory doesn't require memtype_op
+     * - memtype_op should be defined to valid op if memory is CPU inaccessible
+     */
+    return (params->flags & UCP_PROTO_COMMON_INIT_FLAG_HDR_ONLY) ||
+           (params->flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY) ||
+           UCP_MEM_IS_ACCESSIBLE_FROM_CPU(mem_type) ||
+           (params->memtype_op != UCT_EP_OP_LAST);
+}
+
 ucs_status_t
 ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
                            const ucp_proto_common_tl_perf_t *tl_perf,
@@ -647,10 +713,8 @@ ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
                            ucp_md_map_t reg_md_map)
 {
     ucp_proto_caps_t *caps = params->super.caps;
-    ucp_proto_perf_range_t xfer_perf, send_perf, recv_perf;
-    const ucp_proto_perf_range_t *parallel_stages[3];
-    ucs_status_t status;
     size_t frag_size;
+    ucs_status_t status;
 
     ucs_trace("caps" UCP_PROTO_TIME_FMT(send_pre_overhead)
               UCP_PROTO_TIME_FMT(send_post_overhead)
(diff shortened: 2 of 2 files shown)


src/ucp/proto/proto_init.c
```diff
@@ -648,18 +653,60 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
...
+    parallel_stages[0] = &send_perf;
+    parallel_stages[1] = &xfer_perf;
+    parallel_stages[2] = &recv_perf;
+
+    /* Add ranges representing sending single fragment */
+    status = ucp_proto_init_parallel_stages(params, range_start, frag_size,
+                                            frag_size, 0.0, parallel_stages, 3);
+
+out_deref_recv_perf:
+    ucp_proto_perf_node_deref(&recv_perf.node);
```
ivankochin, line 693:
```suggestion
    ucp_proto_perf_node_deref(&recv_perf.node);
```

src/ucp/proto/proto_init.c
```diff
@@ -668,40 +714,27 @@ ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
...
-    ucp_proto_common_init_base_caps(params, tl_perf->min_length);
-
-    parallel_stages[0] = &send_perf;
-    parallel_stages[1] = &xfer_perf;
-    parallel_stages[2] = &recv_perf;
-
     /* Add ranges representing sending single fragment */
-    status = ucp_proto_init_parallel_stages(params, 0, frag_size, frag_size,
-                                            0.0, parallel_stages, 3);
+    ucs_assert(frag_size >= 1);
```
brminich, line 733: minor `(frag_size != 0)` looks better

src/ucp/proto/proto_init.c
```diff
@@ -226,7 +226,8 @@ ucp_proto_init_parallel_stages(const ucp_proto_common_init_params_t *params,
             /* For multi-fragment protocols, we need to apply the fragment
              * size to the performance function linear factor.
              */
-            if (!(params->flags & UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG)) {
+            if (!(params->flags & UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG) &&
+                (frag_size > 0)) {
```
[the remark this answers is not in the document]
ivankochin: This is not about selection, we do this procedure for all the protocols. This change was made to avoid division by zero.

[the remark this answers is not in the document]
brminich: ok

src/ucp/proto/proto_init.c
```diff
@@ -648,18 +653,59 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
...
+                                  ucp_proto_perf_node_t *const tl_perf_node,
+                                  ucp_md_map_t reg_md_map, size_t range_start,
+                                  size_t frag_size) {
+    ucp_proto_perf_range_t xfer_perf, send_perf, recv_perf;
+    const ucp_proto_perf_range_t *parallel_stages[3];
+    ucs_status_t status;
+
+    /* Network transfer time */
+    ucp_proto_common_init_xfer_perf(params, tl_perf, tl_perf_node,
+                                    (frag_size == 0), &xfer_perf);
```
yosefe, line 668: minor:
int empty_msg = (frag_size == 0);

src/ucp/proto/proto_init.c
```diff
@@ -668,40 +714,27 @@ ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
...
-    if (status != UCS_OK) {
-        goto out_deref_xfer_perf;
-    }
+    /* Initialize capabilities */
+    ucp_proto_common_init_base_caps(params, tl_perf->min_length);
 
-    /* Receiver overhead */
-    status = ucp_proto_common_init_recv_perf(params, tl_perf, reg_md_map,
-                                             &recv_perf);
+    /* Add ranges representing sending empty message */
```
yosefe, line 720: maybe do it only if min_length == 0 ?

src/ucp/proto/proto_init.c
```diff
@@ -659,24 +657,28 @@ ucp_proto_init_single_frag_ranges(const ucp_proto_common_init_params_t *params,
                                   ucp_proto_perf_node_t *const tl_perf_node,
                                   ucp_md_map_t reg_md_map, size_t range_start,
                                   size_t frag_size) {
+    int empty_msg = (frag_size == 0);
     ucp_proto_perf_range_t xfer_perf, send_perf, recv_perf;
     const ucp_proto_perf_range_t *parallel_stages[3];
     ucs_status_t status;
 
+    if (empty_msg && params->super.caps->min_length > 0) {
```
yosefe, line 665: add ( ) around >

yosefe (dismissed):

yosefe:
@ivankochin pls squash

src/ucp/proto/proto_init.c
```diff
@@ -548,7 +550,7 @@ ucp_proto_common_init_xfer_perf(const ucp_proto_common_init_params_t *params,
     op_attr_mask = ucp_proto_select_op_attr_unpack(select_param->op_attr);
 
     if ((op_attr_mask & UCP_OP_ATTR_FLAG_FAST_CMPL) &&
-        !(params->flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY)) {
+         !(params->flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY)) {
```
brminich, line 553: pls revert, previous indent was correct

src/ucp/proto/proto_init.c
brminich: unresolved

src/ucp/proto/proto_init.c
```diff
@@ -648,18 +651,63 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
...
+                                  ucp_proto_perf_node_t *const tl_perf_node,
+                                  ucp_md_map_t reg_md_map, size_t range_start,
+                                  size_t frag_size) {
+    int empty_msg = (frag_size == 0);
+    ucp_proto_perf_range_t xfer_perf, send_perf, recv_perf;
+    const ucp_proto_perf_range_t *parallel_stages[3];
+    ucs_status_t status;
+
+    if (empty_msg && (params->super.caps->min_length > 0)) {
+        return UCS_OK;
```
brminich, line 666: when can it happen?

ivankochin: Theoretically it can happen for atomic protocols (see `ucp_proto_amo_init`)

brminich: maybe change to assert then? We do not expect 0-sized atomics

yosefe: get_zcopy can also have min_length>0

brminich: then this protocol should not be considered for 0-size message range

yosefe: 1. right, but the initialization is called for the whole protocol, not for a specific range
2. in UCT level, when scatter2cqe is enabled (by default), get_zcopy cannot be done for 0-64B message

brminich: then we have to start range from min_length rather than from 1
discussed that with Ivan offline

contrib/configure-devel
```diff
@@ -19,5 +19,6 @@ $basedir/../configure \
 	--enable-frame-pointer \
 	--enable-stats \
 	--enable-debug-data \
+	--enable-debug \
```
yosefe, line 22: revert?

src/ucp/proto/proto_init.c
ivankochin: After the discussion with @brminich it was decided to move this if outside of this function.

src/ucp/proto/proto_init.c
```diff
@@ -648,60 +651,94 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
...
+ucs_status_t
+ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
+                           const ucp_proto_common_tl_perf_t *tl_perf,
+                           ucp_proto_perf_node_t *const tl_perf_node,
+                           ucp_md_map_t reg_md_map)
+{
+    ucp_proto_caps_t *caps = params->super.caps;
+    ucs_status_t status;
+    size_t frag_size;
+    ucs_trace("caps" UCP_PROTO_TIME_FMT(send_pre_overhead)
```
yosefe, line 707: space line before

src/ucp/proto/proto_init.c
```diff
@@ -648,60 +651,94 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
...
 
+    /* Get fragment size */
+    frag_size = ucs_min(params->max_length,
+                        tl_perf->max_frag - params->hdr_size);
+    if (frag_size == 0) {
+        /* Empty msg range is already initialized */
+        return UCS_OK;
+    }
+
+    ucs_assert_always(frag_size > caps->min_length);
```
yosefe, line 735: assertv

src/ucp/proto/proto_init.c
ivankochin: Space line before already included to `UCP_PROTO_TIME_FMT`

yosefe: I mean before there ucs_trace

src/ucp/proto/proto_init.c
ivankochin: Done.

src/ucp/proto/proto_init.c
ivankochin: Sorry, I misread it first time. Done.

contrib/configure-devel
ivankochin: Done.

brminich (approved):

src/ucp/proto/proto_init.c
```diff
@@ -732,7 +733,9 @@ ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
         return UCS_OK;
     }
 
-    ucs_assert_always(frag_size > caps->min_length);
+    ucs_assertv_always(frag_size > caps->min_length,
+                       "frag_size=%ld, caps->min_length=%ld",
```
yosefe, line 737: "frag_size=%zu caps->min_length=%zu",

ivankochin:
```suggestion
                       "frag_size=%zu caps->min_length=%zu",
```

ivankochin: Done.

yosefe (approved):

src/ucp/proto/proto_init.c
```diff
@@ -668,40 +713,35 @@ ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
...
+    }
 
-    /* Initialize capabilities */
-    ucp_proto_common_init_base_caps(params, tl_perf->min_length);
-
-    parallel_stages[0] = &send_perf;
-    parallel_stages[1] = &xfer_perf;
-    parallel_stages[2] = &recv_perf;
-
+    ucs_assertv_always(frag_size > caps->min_length,
```
ivankochin, line 736:
```suggestion
    ucs_assertv_always(frag_size >= caps->min_length,
```

yosefe (dismissed):

brminich (dismissed):

src/ucp/proto/proto_common.c
```diff
@@ -661,9 +661,21 @@ void ucp_proto_request_zcopy_completion(uct_completion_t *self)
 
 int ucp_proto_is_short_supported(const ucp_proto_select_param_t *select_param)
 {
-    /* Short protocol requires contig/host */
-    return (select_param->dt_class == UCP_DATATYPE_CONTIG) &&
-           UCP_MEM_IS_HOST(select_param->mem_type);
```
yosefe, line 666: how does it work when sending small GPU buffer (for example, 8B)? won't we try to use short fast-path send?

ivankochin: In case of small GPU buffer `ucp_proto_short_memtype_params` will set `max_length` to 0. So as I understand we would be able to use short only for empty messages.

yosefe: But then it seem that when memtype cache is not empty (which is always the case now), fast-path short will be completely disabled, also for host memory?

ivankochin: As it was discussed, that patch affects only proto initialization flow, so it shouldn't break the logic

src/ucp/am/eager_single.c
```diff
@@ -115,6 +113,8 @@ ucp_am_eager_short_proto_init_common(const ucp_proto_init_params_t *init_params,
         .tl_cap_flags        = UCT_IFACE_FLAG_AM_SHORT
     };
 
+    ucp_proto_short_memtype_params(&params.super, select_param->mem_type);
```
yosefe, line 116: maybe we can move this to proto common logic?
i.e keep `super.memtype_op = UCT_EP_OP_LAST` definition, and the common logic that checks memtype_op value will set max_length to 0 if select_param->mem_type is not HOST (instead of returning UNSUPPORTED)

src/ucp/proto/proto_init.c
```diff
@@ -648,60 +651,98 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
...
+        status = ucp_proto_init_single_frag_ranges(params, tl_perf,
+                                                   tl_perf_node, reg_md_map,
+                                                   0, 0);
+        if (status != UCS_OK) {
+            return status;
+        }
+    }
 
+    /* Get fragment size */
+    frag_size = ucs_min(params->max_length,
```
brminich, line 731: i'd leave this assert to properly catch if hdr_size is larger than max_frag
`ucs_assert(tl_perf->max_frag >= params->hdr_size);`

ivankochin: Done.

src/ucp/tag/eager_single.c
```diff
@@ -55,8 +55,8 @@ ucp_proto_eager_short_init(const ucp_proto_init_params_t *init_params)
         .super.cfg_thresh    = UCS_MEMUNITS_AUTO,
         .super.cfg_priority  = 0,
         .super.min_length    = 0,
-        .super.max_length    = SIZE_MAX,
         .super.min_iov       = 0,
+        .super.max_length    = SIZE_MAX,
```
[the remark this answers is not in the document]
ivankochin: Done.

src/ucp/am/eager_single.c
```diff
@@ -99,8 +99,8 @@ ucp_am_eager_short_proto_init_common(const ucp_proto_init_params_t *init_params,
         .super.cfg_thresh    = UCS_MEMUNITS_AUTO,
         .super.cfg_priority  = 0,
         .super.min_length    = 0,
-        .super.max_length    = SIZE_MAX,
         .super.min_iov       = 0,
+        .super.max_length    = SIZE_MAX,
```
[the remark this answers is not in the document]
ivankochin: Done.

src/ucp/proto/proto_common.c
```diff
@@ -661,9 +660,19 @@ void ucp_proto_request_zcopy_completion(uct_completion_t *self)
...
+}
+
+int ucp_proto_mem_is_accessible(const ucp_proto_select_param_t *select_param,
+                                uct_ep_operation_t memtype_op)
+{
+    if (UCP_MEM_IS_ACCESSIBLE_FROM_CPU(select_param->mem_type)) {
+        return 1;
+    } else if (memtype_op != UCT_EP_OP_LAST) {
+        return 1;
+    }
```
yosefe, line 673: return UCP_MEM_IS_ACCESSIBLE_FROM_CPU(select_param->mem_type) ||
   (memtype_op != UCT_EP_OP_LAST);

src/ucp/proto/proto_common.c
```diff
@@ -424,8 +424,7 @@ static ucp_lane_index_t ucp_proto_common_find_lanes_internal(
                       ucp_datatype_class_names[select_param->dt_class]);
             goto out;
         }
-    } else if (!UCP_MEM_IS_ACCESSIBLE_FROM_CPU(select_param->mem_type) &&
-               (memtype_op == UCT_EP_OP_LAST)) {
+    } else if (!ucp_proto_mem_is_accessible(select_param, memtype_op)) {
```
yosefe, line 427: ucp_proto_common_is_mem_accessible

src/ucp/proto/proto_single.c
```diff
@@ -62,10 +62,31 @@ ucp_proto_single_init_priv(const ucp_proto_single_init_params_t *params,
     return status;
 }
 
-ucs_status_t ucp_proto_single_init(const ucp_proto_single_init_params_t *params)
+static void
+ucp_proto_single_check_memory_access(ucp_proto_single_init_params_t *params)
```
yosefe, line 66: why is it only for proto single? i think it should be in proto_common

src/ucp/proto/proto_common.c
```diff
@@ -661,9 +660,19 @@ void ucp_proto_request_zcopy_completion(uct_completion_t *self)
...
-           UCP_MEM_IS_HOST(select_param->mem_type);
+    return (select_param->dt_class == UCP_DATATYPE_CONTIG);
+}
+
+int ucp_proto_mem_is_accessible(const ucp_proto_select_param_t *select_param,
+                                uct_ep_operation_t memtype_op)
+{
+    if (UCP_MEM_IS_ACCESSIBLE_FROM_CPU(select_param->mem_type)) {
+        return 1;
+    } else if (memtype_op != UCT_EP_OP_LAST) {
```
yosefe, line 671: could short protocol be selected for 0-byte GPU memory send?

ivankochin: Yes, it can

src/ucp/proto/proto_common.c
ivankochin: Done.

src/ucp/proto/proto_common.c
ivankochin: Done.

src/ucp/proto/proto_single.c
ivankochin: Is there any sense to enable multi-fragment protocols on empty messages?

yosefe: some protocols (like put sw) have only multi fragment implementation

ivankochin: Moved to proto_common ranges initialization logic.

ivankochin:
Performance checked. @yosefe pls review

src/ucp/proto/proto_init.c
```diff
@@ -640,60 +643,115 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
...
+    uint8_t mem_type = params->super.select_param->mem_type;
+
+    /* HDR_ONLY and ZCOPY protocols don't need access to memory */
+    if ((params->flags & UCP_PROTO_COMMON_INIT_FLAG_HDR_ONLY) ||
+        (params->flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY)) {
+        return 1;
+    }
+
+    return UCP_MEM_IS_ACCESSIBLE_FROM_CPU(mem_type) ||
+           (params->memtype_op != UCT_EP_OP_LAST);
```
yosefe, line 704: can we unite to single "return" statement (and document each condition)?

src/ucp/proto/proto_init.c
```diff
@@ -640,60 +643,115 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
...
+            return status;
+        }
+    }
+
+    /* Get fragment size */
+    ucs_assert(tl_perf->max_frag >= params->hdr_size);
+    frag_size = ucs_min(params->max_length,
+                        tl_perf->max_frag - params->hdr_size);
+    if ((frag_size == 0) || !ucp_proto_common_check_mem_access(params)) {
+        return (caps->min_length == 0) ? UCS_OK : UCS_ERR_UNSUPPORTED;
```
yosefe, line 743: pls document

src/ucp/proto/proto_init.c
```diff
@@ -640,60 +643,115 @@ ucp_proto_common_init_recv_perf(const ucp_proto_common_init_params_t *params,
...
+    }
 
+    ucs_assertv_always(frag_size >= caps->min_length,
+                       "frag_size=%zu caps->min_length=%zu",
+                       frag_size, caps->min_length);
```
yosefe, line 748: space line after

src/ucp/proto/proto_common.c
```diff
@@ -661,9 +650,7 @@ void ucp_proto_request_zcopy_completion(uct_completion_t *self)
 
 int ucp_proto_is_short_supported(const ucp_proto_select_param_t *select_param)
 {
-    /* Short protocol requires contig/host */
-    return (select_param->dt_class == UCP_DATATYPE_CONTIG) &&
-           UCP_MEM_IS_HOST(select_param->mem_type);
+    return (select_param->dt_class == UCP_DATATYPE_CONTIG);
```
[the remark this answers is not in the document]
ivankochin: As I understand, we can enable it for iov, but generic requires calling custom pack/unpack (as bcopy do), and we do not support generic even for zcopy.

BTW eager/singe/zcopy also disabled for IOV, should we enable it for empty messages?

yosefe (dismissed):

src/ucp/proto/proto_common.c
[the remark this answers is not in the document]
ivankochin: Discussed offline.

I think it could be done in the separate PR. I want to propose the following solution (https://github.com/ivankochin/ucx/pull/5) which would be common for all the protocols.

ivankochin:
Approval was dismissed by mistake, the code base is the same.

brminich (approved):

PR #9407, opened 2023-10-05
UCP/PROTOV2: Prefer native proto implementation

brminich:
Prefer native RMA and AMO protocols over emulations to avoid selections like below:
```
 7543 [1696516991.258413] [jazz20:103988:0]   +----------------------------+-------------------------------------------------------------+
 7544 [1696516991.258425] [jazz20:103988:0]   | 0x2d0aca0 intra-node cfg#2 | remote memory write by ucp_put* from host memory to host    |
 7545 [1696516991.258428] [jazz20:103988:0]   +----------------------------+------------------------------------------+------------------+
 7546 [1696516991.258430] [jazz20:103988:0]   |                     0..220 | short                                    | rc_mlx5/mlx5_0:1 |
 7547 [1696516991.258433] [jazz20:103988:0]   |                   221..349 | software emulation                       | rc_mlx5/mlx5_0:1 |
 7548 [1696516991.258435] [jazz20:103988:0]   |                   350..inf | zero-copy                                | knem/memory      |
 7549 [1696516991.258438] [jazz20:103988:0]   +----------------------------+------------------------------------------+------------------+

```

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -455,6 +455,13 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, reg_nb_mem_types),
    UCS_CONFIG_TYPE_BITMAP(ucs_memory_type_names)},
 
+  {"PREFER_OFFLOAD", "y",
+   "Prefer transports capable of remote memory access for RMA and AMO operations.\n"
+   "The value is interpreted as follows:\n"
+   " 'y' : Prefer transports with native RMA/AMO support (if available)\n"
+   " 'n' : Select RMA/AMO lanes according to performance charasteristics",
+   ucs_offsetof(ucp_context_config_t, prefer_offload), UCS_CONFIG_TYPE_BOOL},
+
   {NULL}
 };
 
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -164,6 +164,8 @@ typedef struct ucp_context_config {
     char                                   *proto_info_dir;
     /** Memory types that perform non-blocking registration by default */
     uint64_t                               reg_nb_mem_types;
+    /** Prefer native RMA transports for RMA/AMO protocols */
+    int                                    prefer_offload;
 } ucp_context_config_t;
 
 
--- a/src/ucp/rma/amo_sw.c
+++ b/src/ucp/rma/amo_sw.c
@@ -408,7 +408,7 @@ ucp_proto_amo_sw_init(const ucp_proto_init_params_t *init_params, unsigned flags
         .super.super         = *init_params,
         .super.latency       = 1.2e-6,
         .super.overhead      = 40e-9,
-        .super.cfg_thresh    = 0,
+        .super.cfg_thresh    = ucp_proto_sw_rma_cfg_thresh(worker->context, 0),
         .super.cfg_priority  = 20,
         .super.min_length    = sizeof(uint32_t),
         .super.max_length    = sizeof(uint64_t),
--- a/src/ucp/rma/get_am.c
+++ b/src/ucp/rma/get_am.c
@@ -79,7 +79,8 @@ ucp_proto_get_am_bcopy_init(const ucp_proto_init_params_t *init_params)
         .super.super         = *init_params,
         .super.latency       = 0,
         .super.overhead      = 40e-9,
-        .super.cfg_thresh    = context->config.ext.bcopy_thresh,
+        .super.cfg_thresh    = ucp_proto_sw_rma_cfg_thresh(
+                                   context, context->config.ext.bcopy_thresh),
         .super.cfg_priority  = 20,
         .super.min_length    = 0,
         .super.max_length    = SIZE_MAX,
--- a/src/ucp/rma/put_am.c
+++ b/src/ucp/rma/put_am.c
@@ -81,7 +81,8 @@ ucp_proto_put_am_bcopy_init(const ucp_proto_init_params_t *init_params)
         .super.super         = *init_params,
         .super.latency       = 0,
         .super.overhead      = 40e-9,
-        .super.cfg_thresh    = context->config.ext.bcopy_thresh,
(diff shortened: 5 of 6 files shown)


rakhmets:
Can it be done by changing the performance estimation of "emulating" protocols? Maybe the estimation is too optimistic.
I think that "emulating" protocol can be used if "native" is not available for some reasons.

brminich:
> Can it be done by changing the performance estimation of "emulating" protocols? Maybe the estimation is too optimistic. I think that "emulating" protocol can be used if "native" is not available for some reasons.

I do not think it is proper way to proceed here. In this case using native protocol is important to to have real RMA, even the performance is worse than emulation

yosefe:
Perf offline discussion, will make this configurable (default: prefer offload)

src/ucp/rma/rma_sw.c
```diff
@@ -15,9 +15,25 @@
...
 #include <ucp/proto/proto_common.inl>
 
 
+int ucp_rma_sw_proto_init_check_op(const ucp_proto_init_params_t *init_params,
+                                   uint64_t op_id_mask)
+{
+    ucp_context_h context = init_params->worker->context;
+
+    return (context->config.ext.rma_offload != UCS_YES) &&
+           ucp_proto_init_check_op(init_params, UCS_BIT(UCP_OP_ID_PUT));
```
rakhmets, line 28:
```suggestion
           ucp_proto_init_check_op(init_params, op_id_mask);
```

src/ucp/core/ucp_context.c
```diff
@@ -455,6 +455,15 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, reg_nb_mem_types),
    UCS_CONFIG_TYPE_BITMAP(ucs_memory_type_names)},
 
+  {"RMA_OFFLOAD", "try",
+   "Prefer transports capable of remote memory access for RMA and AMO operations.\n"
```
gleon99, line 459: IMO calling it "prefer" is inconsistent with the meaning of 'y' value. It is more like "force" (e.g. as in `UCX_DC_MLX5_DCI_FULL_HANDSHAKE`).

src/ucp/wireup/select.c
```diff
@@ -918,7 +919,7 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_add_memaccess_lanes(
 
     /* If could not find registered memory access lane, try to use emulation */
     if (status != UCS_OK) {
-        if (!select_params->allow_am) {
+        if (!allow_am) {
```
gleon99, line 922: How did that work until now? The wrong value (`select_params->allow_am`) was used?

brminich: `select_params->allow_am` is almost the same (see line 890),
the only difference was exported memh feature, which is not widely tested currently

src/ucp/core/ucp_context.c
```diff
@@ -455,6 +455,15 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, reg_nb_mem_types),
    UCS_CONFIG_TYPE_BITMAP(ucs_memory_type_names)},
 
+  {"RMA_OFFLOAD", "try",
```
yosefe, line 458: maybe PROTO_OFFLOAD or OFFLOAD_PROTO - to set mode for offload protocols for all cases including HWTM for example?

src/ucp/proto/proto_common.h
```diff
@@ -57,7 +57,11 @@ typedef enum {
     UCP_PROTO_COMMON_INIT_FLAG_CAP_SEG_SIZE  = UCS_BIT(8),
 
     /* Supports error handling */
-    UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING  = UCS_BIT(9)
+    UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING  = UCS_BIT(9),
+
+    /* The protocol emulates requested functionality (e.g. sw implementation of
+     * RMA ops) */
+    UCP_PROTO_COMMON_INIT_FLAG_EMULATION     = UCS_BIT(10)
```
yosefe, line 64: maybe invert the flag - UCP_PROTO_COMMON_INIT_FLAG_OFFLOAD

src/ucp/proto/proto_select.c
```diff
@@ -113,6 +113,12 @@ static ucs_status_t ucp_proto_thresholds_next_range(
         proto_perf[proto_id] = range->perf[perf_type];
         max_length           = ucs_min(max_length, range->max_length);
 
+        if (caps->flags & UCP_PROTO_COMMON_INIT_FLAG_EMULATION) {
```
yosefe, line 116: don't we need to check the context config before disabling?

src/ucp/rma/rma_sw.c
```diff
@@ -15,9 +15,25 @@
...
+
+    return (context->config.ext.rma_offload != UCS_YES) &&
+           ucp_proto_init_check_op(init_params, op_id_mask);
+}
+
+unsigned ucp_rma_sw_emulation_flag(ucp_context_h context)
+{
+    return (context->config.ext.rma_offload != UCS_NO) ?
+           UCP_PROTO_COMMON_INIT_FLAG_EMULATION : 0;
+}
```
yosefe, line 35: I think the protocol should exhibit the right flag regardless of ep config, and the common protocol selection logic should prefer offload according to disabled_proto_mask

src/ucp/core/ucp_context.c
```diff
@@ -455,6 +455,15 @@ static ucs_config_field_t ucp_context_config_table[] = {
...
+   "Select transports capable of remote memory access for RMA and AMO operations.\n"
+   "The value is interpreted as follows:\n"
+   " 'y'   : Use only transports with native RMA/AMO support,\n"
+   "         fail if such transports are not available\n"
```
yosefe, line 462: why do we need such mode?

yosefe (approved):
pls squash

PR #9409, opened 2023-10-09
AZP: Lock resources to prevent double start

Alexey-Rivkin:
## What
Prevent results tampering due to parallel start on 2+ PRs.

### DIFF
--- a/buildlib/azure-pipelines-perf.yml
+++ b/buildlib/azure-pipelines-perf.yml
@@ -8,8 +8,10 @@ pr:
     - '*'
 
 variables:
+- name: SharedDir
+  value: /hpc/scrap/azure/$(Build.DefinitionName)
 - name: WorkDir
-  value: /hpc/scrap/azure/$(Build.DefinitionName)/$(Build.BuildId)-$(Build.BuildNumber)
+  value: $(SharedDir)/$(Build.BuildId)-$(Build.BuildNumber)
 - name: threshold
   value: 5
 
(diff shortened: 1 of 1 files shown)


buildlib/azure-pipelines-perf.yml
```diff
@@ -31,6 +33,28 @@ stages:
           - ucx_perf
 
         steps:
+          - bash: |
+              set -xeE
+              lock_file="$(SharedDir)/on-air.lock"
```
artemry-nv, line 38: Aren't there any standard GHA mechanisms to prevent parallel workflow runs?
CC @B-a-S

Alexey-Rivkin: Nope. First thing I checked.

Alexey-Rivkin:
> @Alexey-Rivkin which tests fail without this PR?

No test fails. The purpose of this PR is to prevent concurrent tests for different PRs, skewing final results.

yosefe:
> > @Alexey-Rivkin which tests fail without this PR?
> 
> No test fails. The purpose of this PR is to prevent concurrent tests for different PRs, skewing final results.

Can't we limit number of agents to 1?

Alexey-Rivkin:
> > > @Alexey-Rivkin which tests fail without this PR?
> > 
> > 
> > No test fails. The purpose of this PR is to prevent concurrent tests for different PRs, skewing final results.
> 
> Can't we limit number of agents to 1?

Good point - done that.
I would still prefer to have this locking logic for added safety.

yosefe:
> > > > @Alexey-Rivkin which tests fail without this PR?
> > > 
> > > 
> > > No test fails. The purpose of this PR is to prevent concurrent tests for different PRs, skewing final results.
> > 
> > 
> > Can't we limit number of agents to 1?
> 
> Good point - done that. I would still prefer to have this locking logic for added safety.

If not needed, I'd prefer to not add more logic to the testing infra

Alexey-Rivkin:
Required as a prerequisite to https://github.com/openucx/ucx/pull/9431

PR #9411, opened 2023-10-10
UCT/SCOPY: Adjust latency overhead

ivankochin:
## What
Adjusting the latency overhead for `knem` and `cma`.

## Why
To bring the performance estimation model closer to the real picture.

After the measurements on different platforms it was turned out that the real latency overhead for knem and cma transports are 0.5 usec. Method of measurements:
```
ucx_perftest localhost -t get -d memory -x knem -D zcopy -s 1
```


yosefe (approved):

ivankochin:
Disabled automerge because want to check perf for degradations before merging

PR #9415, opened 2023-10-11
UCS/ARRAY: Simplify array implementation

yosefe:
## Why
- Simplify code using ucs_array
- Preparation for adding dynamic bitmap data structure over ucs_array
   - Dynamic bitmap is needed to add protocol variants; with variants the number of instantiated protocols can exceed 64


yosefe:
@evgeny-leksikov @tvegas1 can you pls review?

src/ucs/datastruct/array.c
```diff
@@ -47,7 +48,17 @@ ucs_status_t ucs_array_grow(void **buffer_p, size_t *capacity_p,
     memcpy(new_buffer, *buffer_p, current_capacity * value_size);
 
 out:
+    if (old_buffer_p == NULL) {
+        ucs_array_free(old_buffer);
```
tvegas1, line 52: maybe `ucs_array_buffer_free()` with actual function since `ucs_array_{grow,resize,...}()` operate on the `ucs_array_s()` itself?

src/ucs/datastruct/string_buffer.c
```diff
@@ -68,60 +64,59 @@ void ucs_string_buffer_appendf(ucs_string_buffer_t *strb, const char *fmt, ...)
     va_list ap;
     int ret;
 
-    ucs_array_reserve(string_buffer, &strb->str,
-                      ucs_array_length(&strb->str) + UCS_STRING_BUFFER_RESERVE);
+    ucs_array_reserve(strb, ucs_array_length(strb) + UCS_STRING_BUFFER_RESERVE);
```
tvegas1, line 67: what about changing the macro to simply add free space, in order to use it like `ucs_array_reserve(strb, UCS_STRING_BUFFER_RESERVE)`? seems we currently mostly use it with the added `ucs_array_length()` anyways

test/gtest/ucs/test_datatype.cc
```diff
@@ -1185,26 +1185,44 @@ void test_array::test_fixed(ucs_array_t(test_1int) *array, size_t capacity)
...
 
+UCS_TEST_F(test_array, resize) {
+    static const size_t NUM_ELEMS = 10;
+    static const int VALUE1       = 896;
+    static const int VALUE2       = 1234;
+    test_1int_t test_array;
+
+    ucs_array_init_dynamic(&test_array);
+    *ucs_array_append(&test_array, FAIL()) = VALUE1;
+    ucs_array_resize(&test_array, NUM_ELEMS, VALUE2, FAIL());
```
tvegas1, line 1215: seems it's the only usage so we plan to use it in a subsequent patch?

src/ucs/datastruct/array.c
yosefe: did not understand the comment.. can you pls elaborate?

src/ucs/datastruct/string_buffer.c
yosefe: There are few places where we reserve with the actual size:
https://github.com/openucx/ucx/blob/bd8433bfee918cf346578f1690cde030d16611e6/src/ucp/core/ucp_am.c#L204-L204
https://github.com/openucx/ucx/blob/bd8433bfee918cf346578f1690cde030d16611e6/src/ucp/core/ucp_worker.c#L2522-L2522

tvegas1: i see

src/ucs/datastruct/array.c
tvegas1: actually i now see that `ucs_array_free()` is already a function below. minor: regarding naming, since `ucs_array_{append,last,reserve}()` all operate on `ucs_array_s(<type>)`, i would expect we keep `ucs_array_free` naming if we need to oerate/free `ucs_array_s()` object.

tvegas1: yes ;)

tvegas1 (approved):

PR #9418, opened 2023-10-12
GTEST/COMMON: Add HIP_VERSION condition for ROCm compatibility

nileshnegi:
## What
Adds condition based on HIP_VERSION for using `type` instead of `memoryType` in `hipPointerAttribute_t`.

## Why ?
From ROCm 5.5 onwards, `memoryType` is deprecated and `type` is used instead in `hipPointerAttribute_t`.

### DIFF
--- a/test/gtest/common/mem_buffer.cc
+++ b/test/gtest/common/mem_buffer.cc
@@ -50,7 +50,8 @@
 #endif
 
 #if HAVE_ROCM
-#  include <hip_runtime.h>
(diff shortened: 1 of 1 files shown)


yosefe (approved):

edgargabriel:
Apologies, I am disabling auto-merge on this issue since it might need an additional change, we discovered yesterday with Nilesh a second problem.  Just trying to minimize the number of PRs that we will have to backport to 1.15 branch

PR #9420, opened 2023-10-14
UCT/MM: fix possible sending in out-of-order.

razor1991:
## What
I found that osu_gatherv sometimes hangs when using UCX_TLS=sm in a single aarch64 node (Kunpeng 920). 
Command is below:
`mpirun -np 128 -x UCX_TLS=sm osu_gatherv -m 65536:65535`
When I reduced the fifo size, the probability increased.
`mpirun -np 128 -x UCX_TLS=sm -x UCX_MM_FIFO_SIZE=2 osu_gatherv -m 65536:65535`

## Why ?
When receiver invoked am callback return not OK in function `uct_mm_iface_process_recv`, here ucx will assign a new receive descriptor to FIFO element and update fifo tail. In aarch64 machine, the sender may get the new tail while pack data to a old FIFO element.

## How ?
memory fence should be added before tail update.

### DIFF
--- a/src/uct/sm/mm/base/mm_iface.c
+++ b/src/uct/sm/mm/base/mm_iface.c
@@ -1,6 +1,7 @@
 /**
  * Copyright (c) UT-Battelle, LLC. 2014-2015. ALL RIGHTS RESERVED.
  * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2021. ALL RIGHTS RESERVED.
+ * Copyright (C) Huawei Technologies Co., Ltd. 2023. ALL RIGHTS RESERVED.
  * See file LICENSE for terms.
  */
 
@@ -243,6 +244,10 @@ uct_mm_progress_fifo_tail(uct_mm_iface_t *iface)
     if (iface->read_index & iface->fifo_release_factor_mask) {
         return;
     }
+
+    /* memory barrier - make sure that the memory is flushed before update the
+     * FIFO tail */
+    ucs_memory_cpu_store_fence();
 
     iface->recv_fifo_ctl->tail = iface->read_index;
 }


shamisp:
Can you please clarify  what are the exact instructions you are trying to order here ? Have your tried to use `ucs_compiler_fence` instead of store barrier ? thanks

yosefe (approved):

razor1991:
> Can you please clarify what are the exact store instructions you are trying to order here ? Have your tried to use `ucs_compiler_fence` instead of store barrier ? thanks

Sure. This problem occurs when we use mm bcopy.

**Receiver when ucp is not tag matched,**
1) uct/mm will `assign a new receive descriptor to this FIFO element`. 
image
2) Update the `iface->recv_fifo_ctl->tail`.
image

No fence between 1) and 2), but `elem->desc = desc->info` should be ensured to executed before update tail.

**Sender,**
1) Update cached tail in `uct_mm_ep_update_cached_tail`.
image
2) Get remote element in function `uct_mm_ep_get_remote_elem`.
3) Pack data to the address that the elem point to.

In some cases, receiver update tail firstly, then sender executed 1), 2), 3), then receiver `assign a new receive descriptor`, which will pack data to an unexpected address.

**So there should be a load fence between receiver 1) and 2), to make sure 1) is exexuted before 2)**

yosefe:
@razor1991 can you pls rebase the PR on master branch?

tvegas1:
@razor1991, on reader side, don't we also need to add corresponding cpu load fence between `uct_mm_ep_update_cached_tail()` and `uct_mm_ep_get_remote_elem()` for ordering to be maintained across cpu? (currently load fence is _before_ `uct_mm_ep_update_cached_tail()`)

shamisp:
The patch makes sense @yosefe . There are two stores to the shared memory that we have to order. The descriptor store , followed by tail update.

razor1991:
> @razor1991, on reader side, don't we also need to add corresponding cpu load fence between `uct_mm_ep_update_cached_tail()` and `uct_mm_ep_get_remote_elem()` for ordering to be maintained across cpu? (currently load fence is _before_ `uct_mm_ep_update_cached_tail()`)

head is update with `ucs_atomic_cswap64`, I think the sender side is OK.

tvegas1:
i see, `ucs_atomic_cswap64` has acq/rel semantics

tvegas1 (approved):

yosefe (approved):

razor1991:
2 checks failed, but I don't know how to deal with these.

brminich:
This failure is unrelated to your changes. Once we fix our CI, we will restart testing for this PR

PR #9423, opened 2023-10-16
UCP: Fix proto select hash race

brminich:
## What
When the ucp_ep_config array is extended by the async thread, the main thread may modify ep_config->proto_select.hash by adding new protocol selections. These newly added elements will not be reflected in the new ep_config_array and will be released together with the old ep_config array. Using a pointer to the hash in proto_select fixes this issue because the hash will persist in memory regardless of ep_config array reallocations.

## Why ?
Fixes [RM3633376](https://redmine.mellanox.com/issues/3633376) (internal link)

### DIFF
--- a/src/ucp/proto/proto_select.c
+++ b/src/ucp/proto/proto_select.c
@@ -700,9 +700,9 @@ ucp_proto_select_lookup_slow(ucp_worker_h worker,
     int khret;
 
     key.param = *select_param;
-    khiter    = kh_get(ucp_proto_select_hash, &proto_select->hash, key.u64);
-    if (khiter != kh_end(&proto_select->hash)) {
-        select_elem = &kh_value(&proto_select->hash, khiter);
+    khiter    = kh_get(ucp_proto_select_hash, proto_select->hash, key.u64);
+    if (khiter != kh_end(proto_select->hash)) {
+        select_elem = &kh_value(proto_select->hash, khiter);
         goto out;
     }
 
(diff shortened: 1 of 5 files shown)


src/ucp/proto/proto_select.c
```diff
@@ -734,7 +738,7 @@ ucp_proto_select_lookup_slow(ucp_worker_h worker,
 
 ucs_status_t ucp_proto_select_init(ucp_proto_select_t *proto_select)
 {
-    kh_init_inplace(ucp_proto_select_hash, &proto_select->hash);
+    proto_select->hash = kh_init(ucp_proto_select_hash);
```
tvegas1, line 741: for completeness we need to handle allocation failure?

src/ucp/proto/proto_select.c
```diff
@@ -19,6 +19,10 @@
 #include <ucp/core/ucp_worker.inl>
 #include <ucs/datastruct/array.inl>
 
+__KHASH_IMPL(ucp_proto_select_hash, kh_inline,  khint64_t,
```
tvegas1, line 22: Is that intentional to change `KHASH_TYPE`/`KHASH_IMPL` to `KHASH_DECLARE`/`__KHASH_IMPL`? Seems now `proto_select.inl` would not inline khash functions anymore?

brminich: right, reverted in place (force pushed), because the patch is small.

tvegas1 (approved):

tvegas1 (approved):

yosefe (approved):

PR #9424, opened 2023-10-17
UCT/MD/IB: Fix indirect key creation for MT registered memory

ivankochin:
## What
Support indirect keys creation for memory regions that were created by multiple treads.

## Why ?
MT memory registration can happen (or not) implicitly for large buffers, but the current code doesn't support indirect keys creation for such registration (it always assumes that MR was registered as contig buffer by one thread)

## How ?
Create `uct_ib_mlx5_devx_reg_ksm_data` wrapper that separates KSM creation method by MR flags.

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -112,38 +112,69 @@ uct_ib_mlx5_devx_reg_ksm(uct_ib_mlx5_md_t *md, int atomic, uint64_t address,
     return UCS_OK;
 }
 
-static ucs_status_t
-uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md, int atomic, void *address,
-                              uct_ib_mlx5_devx_ksm_data_t *ksm_data,
-                              size_t length, uint64_t iova, uint32_t mkey_index,
-                              const char *reason, struct mlx5dv_devx_obj **mr_p,
-                              uint32_t *mkey)
+static void uct_ib_mlx5_devx_ksm_log(uct_ib_mlx5_md_t *md, int atomic,
+                                     void *address, size_t length,
+                                     uint64_t iova, ucs_status_t status, int mt)
 {
-    void *mr_address = address;
+    ucs_debug("KSM %s-thread memory registration status \"%s\" "
+              "range %p..%p iova 0x%" PRIx64 "%s on %s",
+              mt ? "multi" : "single", ucs_status_string(status), address,
+              UCS_PTR_BYTE_OFFSET(address, length), iova,
+              atomic ? " atomic" : "", uct_ib_device_name(&md->super.dev));
+}
+
+/*
+ * Register KSM data for given memory handle. Can work only with MRs
+ * that were registered in multi-threaded mode.
+ */
+static ucs_status_t
+uct_ib_mlx5_devx_reg_ksm_data_mt(uct_ib_mlx5_md_t *md, int atomic,
+                                 void *address,
+                                 uct_ib_mlx5_devx_ksm_data_t *ksm_data,
+                                 uint64_t iova, uint32_t mkey_index,
+                                 const char *reason,
+                                 struct mlx5dv_devx_obj **mr_p, uint32_t *mkey)
+{
+    struct ibv_mr *last_mr = ksm_data->mrs[ksm_data->mr_num - 1];
+    uint64_t iova_offset   = iova - (uint64_t)address;
+    void *mr_address       = address;
+    size_t list_size       = ksm_data->mr_num;
     ucs_status_t status;
     char *in;
     void *klm;
     int i;
 
-    status = uct_ib_mlx5_alloc_mkey_inbox(ksm_data->mr_num, &in);
+    /* Add offset to workaround CREATE_MKEY range check issue */
+    if (iova_offset > 0) {
+        ++list_size;
+    }
+
+    status = uct_ib_mlx5_alloc_mkey_inbox(list_size, &in);
     if (status != UCS_OK) {
-        return UCS_ERR_NO_MEMORY;
+        goto out;
     }
 
     klm = UCT_IB_MLX5DV_ADDR_OF(create_mkey_in, in, klm_pas_mtt);
     for (i = 0; i < ksm_data->mr_num; i++) {
         UCT_IB_MLX5DV_SET64(klm, klm, address, (uintptr_t)mr_address);
-        UCT_IB_MLX5DV_SET(klm, klm, byte_count, ksm_data->mrs[i]->length);
         UCT_IB_MLX5DV_SET(klm, klm, mkey, ksm_data->mrs[i]->lkey);
         klm        = UCS_PTR_BYTE_OFFSET(klm, UCT_IB_MLX5DV_ST_SZ_BYTES(klm));
         mr_address = UCS_PTR_BYTE_OFFSET(mr_address, ksm_data->mrs[i]->length);
     }
 
-    status = uct_ib_mlx5_devx_reg_ksm(md, atomic, iova, length,
-                                      ksm_data->mr_num,
-                                      ksm_data->mrs[0]->length, in, mkey_index,
-                                      reason, mr_p, mkey);
+    if (iova_offset > 0) {
+        UCT_IB_MLX5DV_SET64(klm, klm, address, (uintptr_t)mr_address);
+        UCT_IB_MLX5DV_SET(klm, klm, mkey, last_mr->lkey);
+    }
+
+    status = uct_ib_mlx5_devx_reg_ksm(md, atomic, iova, ksm_data->length,
+                                      list_size, ksm_data->mrs[0]->length, in,
+                                      mkey_index, reason, mr_p, mkey);
     ucs_free(in);
+
+    uct_ib_mlx5_devx_ksm_log(md, atomic, address, ksm_data->length, iova,
+                             status, 1);
+out:
     return status;
 }
 
@@ -180,12 +211,17 @@ uct_ib_mlx5_devx_reg_ksm_data_addr(uct_ib_mlx5_md_t *md, struct ibv_mr *mr,
     return status;
 }
 
+/*
+ * Register KSM data for given memory handle. Can work only with MRs
+ * that were registered in single-threaded mode.
+ */
 static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_contig(
         uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mr_t *mr, void *address,
         uint64_t iova, int atomic, uint32_t mkey_index, const char *reason,
         struct mlx5dv_devx_obj **mr_p, uint32_t *mkey)
 {
     size_t mr_length = mr->super.ib->length;
+    ucs_status_t status;
     uint64_t ksm_address;
     uint64_t ksm_iova;
     size_t ksm_length;
@@ -202,10 +238,13 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_contig(
     list_size = ucs_div_round_up(ksm_length + ucs_get_page_size(),
                                  UCT_IB_MD_MAX_MR_SIZE);
 
-    return uct_ib_mlx5_devx_reg_ksm_data_addr(md, mr->super.ib, ksm_address,
-                                              ksm_length, ksm_iova, atomic,
-                                              list_size, mkey_index, reason,
-                                              mr_p, mkey);
+    status = uct_ib_mlx5_devx_reg_ksm_data_addr(md, mr->super.ib, ksm_address,
+                                                ksm_length, ksm_iova, atomic,
+                                                list_size, mkey_index, reason,
+                                                mr_p, mkey);
+
+    uct_ib_mlx5_devx_ksm_log(md, atomic, address, mr_length, ksm_iova, status, 0);
+    return status;
 }
 
 static void *
@@ -362,6 +401,37 @@ static void uct_ib_mlx5_devx_mr_lru_cleanup(uct_ib_mlx5_md_t *md)
     kh_destroy_inplace(rkeys, &md->lru_rkeys.hash);
 }
 
+/*
+ * Register KSM data for given memory handle. Distinguish the way of KSM creation
+ * structures filling by checking UCT_IB_MEM_MULTITHREADED flag.
+ */
+static ucs_status_t
+uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md,
+                              uct_ib_mlx5_devx_mem_t *memh,
+                              uct_ib_mr_type_t mr_type, uint32_t iova_offset,
+                              int atomic, uint32_t mkey_index,
+                              const char *reason, struct mlx5dv_devx_obj **mr_p,
+                              uint32_t *mkey)
+{
+    uct_ib_mlx5_devx_mr_t *mr = &memh->mrs[mr_type];
+    void *address             = uct_ib_mlx5_devx_memh_base_address(memh);
+    uint64_t iova             = (uint64_t)address + iova_offset;
+    ucs_status_t status;
+
+    if (memh->super.flags & UCT_IB_MEM_MULTITHREADED) {
+        status = uct_ib_mlx5_devx_reg_ksm_data_mt(md, atomic, address,
+                                                  mr->ksm_data, iova,
+                                                  mkey_index, reason, mr_p,
+                                                  mkey);
+    } else {
+        status = uct_ib_mlx5_devx_reg_ksm_data_contig(md, mr, address, iova,
+                                                      atomic, mkey_index,
+                                                      reason, mr_p, mkey);
+    }
+
+    return status;
+}
+
 UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_indirect_key,
                         (md, memh), uct_ib_mlx5_md_t *md,
                         uct_ib_mlx5_devx_mem_t *memh)
@@ -372,11 +442,10 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_indirect_key,
                 md->super.name);
 
     do {
-        status = uct_ib_mlx5_devx_reg_ksm_data_contig(
-                md, &memh->mrs[UCT_IB_MR_DEFAULT],
-                uct_ib_mlx5_devx_memh_base_address(memh),
-                (uint64_t)memh->address, 0, 0, "indirect key",
-                &memh->indirect_dvmr, &memh->indirect_rkey);
+        status = uct_ib_mlx5_devx_reg_ksm_data(md, memh, UCT_IB_MR_DEFAULT, 0,
+                                               0, 0, "indirect key",
+                                               &memh->indirect_dvmr,
+                                               &memh->indirect_rkey);
         if (status != UCS_OK) {
             break;
         }
@@ -420,12 +489,9 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_atomic_key,
                         uct_ib_mlx5_devx_mem_t *memh)
 {
     uct_ib_mr_type_t mr_type = uct_ib_devx_get_atomic_mr_type(&md->super, memh);
-    uct_ib_mlx5_devx_mr_t *mr = &memh->mrs[mr_type];
-    uint8_t mr_id             = uct_ib_md_get_atomic_mr_id(&md->super);
-    uint32_t atomic_offset    = uct_ib_md_atomic_offset(mr_id);
+    uint8_t mr_id            = uct_ib_md_get_atomic_mr_id(&md->super);
+    uint32_t atomic_offset   = uct_ib_md_atomic_offset(mr_id);
     uint32_t mkey_index;
-    uint64_t iova;
-    ucs_status_t status;
     int is_atomic;
 
     if (memh->smkey_mr != NULL) {
@@ -436,31 +502,11 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_atomic_key,
     }
 
     is_atomic = memh->super.flags & UCT_IB_MEM_ACCESS_REMOTE_ATOMIC;
-    iova      = (uint64_t)memh->address + atomic_offset;
 
-    if (memh->super.flags & UCT_IB_MEM_MULTITHREADED) {
-        return uct_ib_mlx5_devx_reg_ksm_data(md, is_atomic, memh->address,
-                                             mr->ksm_data, mr->ksm_data->length,
-                                             iova, mkey_index,
-                                             "multi-thread atomic key",
-                                             &memh->atomic_dvmr,
-                                             &memh->atomic_rkey);
-    }
-
-    status = uct_ib_mlx5_devx_reg_ksm_data_contig(
-            md, mr, uct_ib_mlx5_devx_memh_base_address(memh), iova, is_atomic,
-            mkey_index, "atomic key", &memh->atomic_dvmr, &memh->atomic_rkey);
-    if (status != UCS_OK) {
-        return status;
-    }
-
-    ucs_debug("KSM registered memory %p..%p lkey 0x%x offset 0x%x%s on %s rkey "
-              "0x%x",
-              memh->address,
-              UCS_PTR_BYTE_OFFSET(memh->address, mr->super.ib->length),
-              mr->super.ib->lkey, atomic_offset, is_atomic ? " atomic" : "",
-              uct_ib_device_name(&md->super.dev), memh->atomic_rkey);
-    return UCS_OK;
+    return uct_ib_mlx5_devx_reg_ksm_data(md, memh, mr_type, atomic_offset,
+                                         is_atomic, mkey_index, "atomic key",
+                                         &memh->atomic_dvmr,
+                                         &memh->atomic_rkey);
 }
 
 static ucs_status_t
@@ -506,10 +552,10 @@ uct_ib_mlx5_devx_reg_mt(uct_ib_mlx5_md_t *md, void *address, size_t length,
         goto err_free;
     }
 
-    status = uct_ib_mlx5_devx_reg_ksm_data(md, is_atomic, address, ksm_data,
-                                           length, (uint64_t)address, 0,
-                                           "multi-thread key", &ksm_data->dvmr,
-                                           mkey_p);
+    status = uct_ib_mlx5_devx_reg_ksm_data_mt(md, is_atomic, address, ksm_data,
+                                              (uint64_t)address, 0,
+                                              "multi-thread key",
+                                              &ksm_data->dvmr, mkey_p);
     if (status != UCS_OK) {
         goto err_dereg;
     }
@@ -2022,6 +2068,7 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_exported_key,
         goto out_umem_mr;
     }
 
+    ucs_assert(!(memh->super.flags & UCT_IB_MEM_MULTITHREADED));
     status = uct_ib_mlx5_devx_reg_ksm_data_contig(md,
                                                   &memh->mrs[UCT_IB_MR_DEFAULT],
                                                   memh->address,
--- a/test/gtest/uct/ib/test_ib_md.cc
+++ b/test/gtest/uct/ib/test_ib_md.cc
@@ -27,6 +27,8 @@ protected:
                                 uct_rkey_t *rkey_p = NULL);
     void check_smkeys(uct_rkey_t rkey1, uct_rkey_t rkey2);
 
+    void test_mkey_pack_mt(bool invalidate);
+    void test_mkey_pack_mt_internal(unsigned access_mask, bool invalidate);
     void test_smkey_reg_atomic(void);
 
 private:
@@ -222,6 +224,80 @@ void test_ib_md::test_smkey_reg_atomic(void)
     EXPECT_UCS_OK(uct_md_mem_dereg(md(), memh2));
     EXPECT_UCS_OK(uct_md_mem_dereg(md(), memh3));
     ucs_mmap_free(buffer, size);
+}
+
+void test_ib_md::test_mkey_pack_mt_internal(unsigned access_mask,
+                                            bool invalidate)
+{
+    constexpr size_t size = UCS_MBYTE;
+    unsigned pack_flags, dereg_flags;
+    void *buffer;
+    int ret;
+    uct_mem_h memh;
+
+    if (!check_invalidate_support(access_mask)) {
+        UCS_TEST_SKIP_R("mkey invalidation isn't supported");
+    }
+
+    if (!has_ksm()) {
+        UCS_TEST_SKIP_R("KSM is required for MT registration");
+    }
+
+    ret = ucs_posix_memalign(&buffer, size, size, "mkey_pack_mt");
+    ASSERT_EQ(0, ret) << "Allocation failed";
+
+    if (invalidate) {
+        pack_flags  = UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA;
+        dereg_flags = UCT_MD_MEM_DEREG_FLAG_INVALIDATE;
+    } else {
+        pack_flags = dereg_flags = 0;
+    }
+
+    ASSERT_UCS_OK(reg_mem(access_mask, buffer, size, &memh));
+
+    uct_ib_mem_t *ib_memh = (uct_ib_mem_t*)memh;
+    EXPECT_TRUE(ib_memh->flags & UCT_IB_MEM_MULTITHREADED);
+
+    std::vector<uint8_t> rkey(md_attr().rkey_packed_size);
+    uct_md_mkey_pack_params_t pack_params;
+    pack_params.field_mask = UCT_MD_MKEY_PACK_FIELD_FLAGS;
+    pack_params.flags      = pack_flags;
+    ASSERT_UCS_OK(uct_md_mkey_pack_v2(md(), memh, buffer, size,
+                                      &pack_params, rkey.data()));
+
+    uct_md_mem_dereg_params_t params;
+    params.field_mask  = UCT_MD_MEM_DEREG_FIELD_MEMH |
+                         UCT_MD_MEM_DEREG_FIELD_COMPLETION |
+                         UCT_MD_MEM_DEREG_FIELD_FLAGS;
+    params.memh        = memh;
+    params.flags       = dereg_flags;
+    comp().comp.func   = dereg_cb;
+    comp().comp.count  = 1;
+    comp().comp.status = UCS_OK;
+    comp().self        = this;
+    params.comp        = &comp().comp;
+    ASSERT_UCS_OK(uct_md_mem_dereg_v2(md(), &params));
+
+    ucs_free(buffer);
+}
+
+void test_ib_md::test_mkey_pack_mt(bool invalidate)
+{
+    test_mkey_pack_mt_internal(UCT_MD_MEM_ACCESS_REMOTE_ATOMIC, invalidate);
+    test_mkey_pack_mt_internal(UCT_MD_MEM_ACCESS_RMA, invalidate);
+    test_mkey_pack_mt_internal(UCT_MD_MEM_ACCESS_ALL, invalidate);
+}
+
+UCS_TEST_P(test_ib_md, pack_mkey_mt, "IB_REG_MT_THRESH=128K",
+           "IB_REG_MT_CHUNK=128K")
+{
+    test_mkey_pack_mt(false);
+}
+
+UCS_TEST_P(test_ib_md, pack_mkey_mt_invalidate, "IB_REG_MT_THRESH=128K",
+           "IB_REG_MT_CHUNK=128K")
+{
+    test_mkey_pack_mt(true);
 }
 
 UCS_TEST_P(test_ib_md, smkey_reg_atomic)
--- a/test/gtest/uct/ib/test_ib_xfer.cc
+++ b/test/gtest/uct/ib/test_ib_xfer.cc
@@ -7,6 +7,10 @@
 #include <uct/test_p2p_rma.h>
 #include <uct/test_p2p_mix.h>
 
+#include <uct/ib/base/ib_md.h>
+#ifdef HAVE_MLX5_DV
+#include <uct/ib/mlx5/ib_mlx5.h>
+#endif
 
 class uct_p2p_rma_test_xfer : public uct_p2p_rma_test {};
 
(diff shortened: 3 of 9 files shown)


pentschev (approved):
Thanks @ivankochin , this seems to resolve the issue I was experiencing. I'm +1 on this change.

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -222,4 +225,73 @@ UCS_TEST_P(test_ib_md, smkey_reg_atomic)
...
+        dereg_flags = UCT_MD_MEM_DEREG_FLAG_INVALIDATE;
+    } else {
+        pack_flags = dereg_flags = 0;
+    }
+
+    ucs_status_t status = reg_mem(access_mask, buffer.data(), size, &memh);
+    ASSERT_UCS_OK(status);
+
+    /* memh isn't always registered as multithreaded due to the following error:
+       mlx5dv_devx_obj_create(CREATE_MKEY, mode=KSM) failed, syndrome 0x103e77: Remote I/O error
```
brminich, line 256: syndrom is `klm va need to be aligned to max atomic size`

contrib/configure-devel
```diff
@@ -19,5 +19,6 @@ $basedir/../configure \
 	--enable-frame-pointer \
 	--enable-stats \
 	--enable-debug-data \
+	--enable-debug \
 	--enable-mt \
```
ivankochin, line 23:
```suggestion
	--enable-mt \
```

test/gtest/uct/ib/test_ib_xfer.cc
```diff
@@ -129,13 +130,46 @@ UCS_TEST_P(uct_p2p_mix_test_alloc_methods, mix1000)
...
+
+        uct_p2p_mix_test::random_op(sendbuf, recvbuf);
+    }
+
+
+    virtual void init() override {
+        uct_p2p_mix_test::init();
+
+        /* Large buffer can cause failures because of too many chunks 
+         *registration */
```
brminich, line 157: alignment

test/gtest/uct/test_p2p_mix.h
```diff
@@ -56,15 +56,14 @@ class uct_p2p_mix_test : public uct_p2p_test {
...
-    void random_op(const mapped_buffer &sendbuf, const mapped_buffer &recvbuf);
+    virtual void random_op(const mapped_buffer &sendbuf, const mapped_buffer &recvbuf);
 
     void run(unsigned count);
 
     virtual void init();
 
     virtual void cleanup();
 
-private:
```
brminich, line 67: seems like only `m_send_size` needs to be protected, others can remain private

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -556,6 +567,7 @@ static void uct_ib_mlx5_devx_reg_symmetric(uct_ib_mlx5_md_t *md,
     uint32_t symmetric_rkey;
     ucs_status_t status;
 
+    ucs_assert(!(memh->super.flags & UCT_IB_MEM_MULTITHREADED));
```
brminich, line 570: @tvegas1 do we need to update symmetric key creation? To check whether original memh was registered in multiple threads or not

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -362,6 +363,40 @@ static void uct_ib_mlx5_devx_mr_lru_cleanup(uct_ib_mlx5_md_t *md)
     kh_destroy_inplace(rkeys, &md->lru_rkeys.hash);
 }
 
+static ucs_status_t
+uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh,
```
brminich, line 367: minor: why do we need `data` in the naming? maybe:
`uct_ib_mlx5_devx_reg_ksm`
`uct_ib_mlx5_devx_reg_ksm_mt`
`uct_ib_mlx5_devx_reg_ksm_contig`

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
tvegas1: if ksm (symmetric key) in front of ksm (multi-thread) is not supported, i think we should either:
1. prevent MT when `((flags & UCT_MD_MEM_SYMMETRIC_RKEY) && (md->flags & UCT_IB_MLX5_MD_FLAG_MKEY_BY_NAME_RESERVE)` is true
2. prevent symmetric key when multi-thread flag is set on the memh

tvegas1: started #9447 to skip altogether mt reg for smkey.

brminich:
@tvegas1 can you pls review

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1977,12 +1987,12 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_exported_key,
         goto out_umem_mr;
     }
 
-    status = uct_ib_mlx5_devx_reg_ksm_data_contig(md,
-                                                  &memh->mrs[UCT_IB_MR_DEFAULT],
-                                                  memh->address,
-                                                  (uint64_t)memh->address, 0, 0,
-                                                  "exported key", &cross_mr,
-                                                  &exported_lkey);
+    ucs_assert(!(memh->super.flags & UCT_IB_MEM_MULTITHREADED));
```
ivankochin, line 1990: BTW, do someone know should we care about that in case of `uct_ib_mlx5_devx_reg_exported_key` or not? This call also always assumes that MR was create by single thread.

Should we use `uct_ib_mlx5_devx_reg_ksm` wrapper? Or assert is OK for now?

brminich: i'm not sure it is even supported. You can leave an assert for now

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -362,6 +362,39 @@ static void uct_ib_mlx5_devx_mr_lru_cleanup(uct_ib_mlx5_md_t *md)
...
+    uint64_t iova             = (uint64_t)memh->address + iova_offset;
+    void *address             = uct_ib_mlx5_devx_memh_base_address(memh);
+    ucs_status_t status;
+    size_t length;
+
+    if (memh->super.flags & UCT_IB_MEM_MULTITHREADED) {
+        ucs_assertv_always(mr->ksm_data->length > iova_offset,
+                           "mr->ksm_data->length=%ld iova_offset=%u",
+                           mr->ksm_data->length, iova_offset);
+        length = mr->ksm_data->length - iova_offset;
```
brminich, line 381: why?

ivankochin: We set iova (which is `address + iova_offset`) to address field during atomic ksm creation. If I pass full length, then iova + length exceed the registered buffer size and test fails during `devx_create_obj()` call.

@Artemy-Mellanox WDYT is that OK to pass `mr->ksm_data->length - iova_offset` as length?

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -222,4 +225,75 @@ UCS_TEST_P(test_ib_md, smkey_reg_atomic)
...
+    }
+
+    if (invalidate) {
+        pack_flags  = UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA;
+        dereg_flags = UCT_MD_MEM_DEREG_FLAG_INVALIDATE;
+    } else {
+        pack_flags = dereg_flags = 0;
+    }
+
+    ucs_status_t status = reg_mem(access_mask, buffer, size, &memh);
```
tvegas1, line 256: could remove status and expect directly on function calls

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
tvegas1: seems we were not doing it previously on `uct_ib_mlx5_devx_reg_atomic_key()` code path

test/gtest/uct/ib/test_ib_md.cc
ivankochin:
```suggestion
    ASSERT_UCS_OK(reg_mem(access_mask, buffer, size, &memh));
```

Do you mean something like that?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -362,6 +362,39 @@ static void uct_ib_mlx5_devx_mr_lru_cleanup(uct_ib_mlx5_md_t *md)
...
+        status = uct_ib_mlx5_devx_reg_ksm_mt(md, atomic, address, mr->ksm_data,
+                                             length, iova, mkey_index, reason,
+                                             mr_p, mkey);
+    } else {
+        length = mr->super.ib->length;
+        status = uct_ib_mlx5_devx_reg_ksm_contig(md, mr, address, iova, atomic,
+                                                 mkey_index, reason, mr_p, mkey);
+    }
+
+    ucs_debug("KSM registered memory %p..%p offset 0x%x%s on %s rkey 0x%x",
```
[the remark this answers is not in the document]
ivankochin: Done.

test/gtest/uct/ib/test_ib_md.cc
tvegas1: yes

ivankochin: Done.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -113,11 +113,11 @@ uct_ib_mlx5_devx_reg_ksm(uct_ib_mlx5_md_t *md, int atomic, uint64_t address,
 }
 
 static ucs_status_t
-uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md, int atomic, void *address,
-                              uct_ib_mlx5_devx_ksm_data_t *ksm_data,
-                              size_t length, uint64_t iova, uint32_t mkey_index,
-                              const char *reason, struct mlx5dv_devx_obj **mr_p,
-                              uint32_t *mkey)
+uct_ib_mlx5_devx_reg_ksm_mt(uct_ib_mlx5_md_t *md, int atomic, void *address,
```
yosefe, line 116: IMO uct_ib_mlx5_devx_reg_ksm_data is a better name - since this function accepts ksm_data structure to register. It's not actually doing a multi-threaded registration

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -55,10 +55,10 @@ static ucs_status_t uct_ib_mlx5_alloc_mkey_inbox(int list_size, char **in_p)
 }
 
 static ucs_status_t
-uct_ib_mlx5_devx_reg_ksm(uct_ib_mlx5_md_t *md, int atomic, uint64_t address,
-                         size_t length, int list_size, size_t entity_size,
-                         char *in, uint32_t mkey_index, const char *reason,
-                         struct mlx5dv_devx_obj **mr_p, uint32_t *mkey)
+uct_ib_mlx5_devx_create_ksm(uct_ib_mlx5_md_t *md, int atomic, uint64_t address,
```
yosefe, line 58: IMO it's worth adding 1-2 lines of documentation to each KSM registration function to document what it's doing in high level. The names are very similar to each other..

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: agree, seems it was mr->ksm_data->length previously, @ivankochin is this another bug fix?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -362,6 +362,41 @@ static void uct_ib_mlx5_devx_mr_lru_cleanup(uct_ib_mlx5_md_t *md)
...
+        length = mr->super.ib->length;
+        status = uct_ib_mlx5_devx_reg_ksm_contig(md, mr, address, iova, atomic,
+                                                 mkey_index, reason, mr_p, mkey);
+    }
+
+    if (status == UCS_OK) {
+        ucs_debug("KSM registered memory %p..%p offset 0x%x%s on %s rkey 0x%x",
+                  address, UCS_PTR_BYTE_OFFSET(address, length),
+                  iova_offset, atomic ? " atomic" : "",
+                  uct_ib_device_name(&md->super.dev), *mkey);
```
yosefe, line 395: maybe put the debug message inside uct_ib_mlx5_devx_reg_ksm_mt and uct_ib_mlx5_devx_reg_ksm_contig, to distinguish between them in the log?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Yes, without the subtraction of the offset I get the following error: 
```
create mkey: window size exceed translations_octword_size
```

yosefe: why is it needed not after the change?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Done.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Done.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -408,12 +491,9 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_atomic_key,
                         uct_ib_mlx5_devx_mem_t *memh)
 {
     uct_ib_mr_type_t mr_type  = uct_ib_md_get_atomic_mr_type(&md->super);
-    uct_ib_mlx5_devx_mr_t *mr = &memh->mrs[mr_type];
```
brminich, line 411: pls fix alignment by `=` (now there is extra space)

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1987,15 +2046,17 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_exported_key,
...
-                                                  memh->address,
-                                                  (uint64_t)memh->address, 0, 0,
+                                                  memh->address, 0, 0, 0,
                                                   "exported key", &cross_mr,
                                                   &exported_lkey);
     if (status != UCS_OK) {
         return status;
     }
+    ucs_assert(exported_lkey != 0);
+    ucs_assert(cross_mr != NULL);
```
brminich, line 2059: why needed? Does it fix some valgrind error?

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -224,6 +226,78 @@ void test_ib_md::test_smkey_reg_atomic(void)
...
+test_ib_md::test_mkey_pack_mt_internal(unsigned access_mask, bool invalidate)
+{
+    constexpr size_t size = UCS_MBYTE;
+    unsigned pack_flags, dereg_flags;
+    void* buffer;
+    int ret;
+    uct_mem_h memh;
+
+    if ((access_mask & UCT_MD_MEM_ACCESS_REMOTE_ATOMIC) && is_bf_arm()) {
+        UCS_TEST_SKIP_R("FIXME: AMO reg key bug on BF device, skipping");
```
brminich, line 239: what is the issue?

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -224,6 +226,78 @@ void test_ib_md::test_smkey_reg_atomic(void)
...
+    uct_md_mem_dereg_params_t params;
+    params.field_mask  = UCT_MD_MEM_DEREG_FIELD_MEMH |
+                         UCT_MD_MEM_DEREG_FIELD_COMPLETION |
+                         UCT_MD_MEM_DEREG_FIELD_FLAGS;
+    params.memh        = memh;
+    params.flags       = dereg_flags;
+    comp().comp.func   = dereg_cb;
+    comp().comp.count  = 1;
+    comp().comp.status = UCS_OK;
+    comp().self        = this;
```
brminich, line 281: do we really need completion in this test?

test/gtest/uct/ib/test_ib_xfer.cc
```diff
@@ -129,13 +133,81 @@ UCS_TEST_P(uct_p2p_mix_test_alloc_methods, mix1000)
...
+        /* We need at least two chunks */
+        m_send_size = ucs_max(m_send_size, reg_mt_chunk + 1);
+    }
+
+    virtual void cleanup() override {
+        uct_p2p_mix_test::cleanup();
+
+        pop_config();
+    }
+
```
brminich, line 193: extra line

test/gtest/uct/test_p2p_mix.cc
```diff
@@ -161,6 +188,13 @@ void uct_p2p_mix_test::random_op(const mapped_buffer &sendbuf,
     }
 }
 
+
+void uct_p2p_mix_test::check_buffers(const mapped_buffer &sendbuf,
+                                     const mapped_buffer &recvbuf)
+{
+    /* Empty function that is overrided in descendants */
```
brminich, line 195: `overridden`

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Before my patch, both `exported_lkey` and `cross_mr` were only defined in the beginning of this function. After adding the logging func in the end of `uct_ib_mlx5_devx_reg_ksm_data_contig`, compilation on Ubuntu 18.04 and SLES 15 started to fail (all other platforms passed) with message that these variables can be uninitialized.

So I decided to initialize them and check that we change them by the right way by these asserts.

test/gtest/uct/ib/test_ib_md.cc
ivankochin: I copied that message from `test_md::test_reg_mem` since it's doing the pretty similar thing. I didn't catch this error by myself.

test/gtest/uct/ib/test_ib_xfer.cc
ivankochin: Removed.

test/gtest/uct/test_p2p_mix.cc
ivankochin: Nice catch! Done.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -419,13 +503,10 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_atomic_key,
                         (md, memh), uct_ib_mlx5_md_t *md,
                         uct_ib_mlx5_devx_mem_t *memh)
 {
-    uct_ib_mr_type_t mr_type = uct_ib_devx_get_atomic_mr_type(&md->super, memh);
-    uct_ib_mlx5_devx_mr_t *mr = &memh->mrs[mr_type];
+    uct_ib_mr_type_t mr_type  = uct_ib_md_get_atomic_mr_type(&md->super);
     uint8_t mr_id             = uct_ib_md_get_atomic_mr_id(&md->super);
     uint32_t atomic_offset    = uct_ib_md_atomic_offset(mr_id);
```
ivankochin, line 508:
```suggestion
    uct_ib_mr_type_t mr_type = uct_ib_md_get_atomic_mr_type(&md->super);
    uint8_t mr_id            = uct_ib_md_get_atomic_mr_id(&md->super);
    uint32_t atomic_offset   = uct_ib_md_atomic_offset(mr_id);
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -419,13 +503,10 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_atomic_key,
                         (md, memh), uct_ib_mlx5_md_t *md,
                         uct_ib_mlx5_devx_mem_t *memh)
 {
-    uct_ib_mr_type_t mr_type = uct_ib_devx_get_atomic_mr_type(&md->super, memh);
-    uct_ib_mlx5_devx_mr_t *mr = &memh->mrs[mr_type];
-    uint8_t mr_id             = uct_ib_md_get_atomic_mr_id(&md->super);
-    uint32_t atomic_offset    = uct_ib_md_atomic_offset(mr_id);
+    uct_ib_mr_type_t mr_type = uct_ib_md_get_atomic_mr_type(&md->super);
```
ivankochin, line 506:
```suggestion
    uct_ib_mr_type_t mr_type = uct_ib_devx_get_atomic_mr_type(&md->super, memh);
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Here is the log of failure:

```
/__w/1/s/contrib/../src/uct/ib/mlx5/dv/ib_mlx5dv_md.c: In function 'uct_ib_mlx5_devx_reg_exported_key':
/__w/1/s/contrib/../src/uct/ib/mlx5/dv/ib_mlx5dv_md.c:487:17: error: 'exported_lkey' may be used uninitialized in this function [-Werror=maybe-uninitialized]
     return mkey >> 8;
            ~~~~~^~~~
/__w/1/s/contrib/../src/uct/ib/mlx5/dv/ib_mlx5dv_md.c:2071:14: note: 'exported_lkey' was declared here
     uint32_t exported_lkey;
              ^~~~~~~~~~~~~
/__w/1/s/contrib/../src/uct/ib/mlx5/dv/ib_mlx5dv_md.c:2109:25: error: 'cross_mr' may be used uninitialized in this function [-Werror=maybe-uninitialized]
     memh->cross_mr      = cross_mr;
     ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
/__w/1/s/contrib/../src/uct/ib/mlx5/dv/ib_mlx5dv_md.c:2070:29: note: 'cross_mr' was declared here
     struct mlx5dv_devx_obj *cross_mr;
                             ^~~~~~~~
```

It appeared after adding `uct_ib_mlx5_devx_ksm_log` function to `uct_ib_mlx5_devx_reg_ksm_data_contig` which isn't touch any uninitialized memory. Removing log function solves the issue.

Initialization of these variables seems the best way to solve this issue.

ivankochin: I found that the issue can be solved by removing information about `mkey` from `uct_ib_mlx5_devx_ksm_log ` function. That information is already printed inside `uct_ib_mlx5_devx_reg_ksm` in case of successfull registration, so, I think it can be removed here.

test/gtest/uct/ib/test_ib_md.cc
brminich: maybe it's not relevant anymore and can be removed? Can you try remove it and see whether CI passes.
@Artemy-Mellanox are you aware about this issue?

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -224,6 +226,79 @@ void test_ib_md::test_smkey_reg_atomic(void)
...
+    unsigned pack_flags, dereg_flags;
+    void *buffer;
+    int ret;
+    uct_mem_h memh;
+
+    if ((access_mask & UCT_MD_MEM_ACCESS_REMOTE_ATOMIC) && is_bf_arm()) {
+        UCS_TEST_SKIP_R("FIXME: AMO reg key bug on BF device, skipping");
+        return;
+    }
+
```
ivankochin, line 242:
```suggestion

```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -362,6 +401,40 @@ static void uct_ib_mlx5_devx_mr_lru_cleanup(uct_ib_mlx5_md_t *md)
...
+                              const char *reason, struct mlx5dv_devx_obj **mr_p,
+                              uint32_t *mkey)
+{
+    uct_ib_mlx5_devx_mr_t *mr = &memh->mrs[mr_type];
+    void *address             = uct_ib_mlx5_devx_memh_base_address(memh);
+    ucs_status_t status;
+
+    if (memh->super.flags & UCT_IB_MEM_MULTITHREADED) {
+        ucs_assertv_always(mr->ksm_data->length > iova_offset,
+                           "mr->ksm_data->length=%ld iova_offset=%u",
```
brminich, line 422:
```suggestion
                           "mr->ksm_data->length=%zu iova_offset=%u",
```

brminich (dismissed):

ivankochin:
@Artemy-Mellanox  pls review

Artemy-Mellanox (dismissed):

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -224,6 +226,74 @@ void test_ib_md::test_smkey_reg_atomic(void)
...
+    ucs_free(buffer);
+}
+
+void test_ib_md::test_mkey_pack_mt(bool invalidate)
+{
+    test_mkey_pack_mt_internal(UCT_MD_MEM_ACCESS_REMOTE_ATOMIC, invalidate);
+    test_mkey_pack_mt_internal(md_flags_remote_rma, invalidate);
+    test_mkey_pack_mt_internal(UCT_MD_MEM_ACCESS_ALL, invalidate);
+    test_mkey_pack_mt_internal(UCT_MD_MEM_ACCESS_RMA, invalidate);
```
yosefe, line 288: what is the difference between UCT_MD_MEM_ACCESS_RMA and md_flags_remote_rma?

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -224,6 +226,74 @@ void test_ib_md::test_smkey_reg_atomic(void)
...
+}
+
+UCS_TEST_P(test_ib_md, pack_mkey_mt, "REG_MT_THRESH=128K", "REG_MT_CHUNK=128K")
+{
+    test_mkey_pack_mt(false);
+    test_mkey_pack_mt(true);
```
yosefe, line 294: maybe separate to 2 test functions (invalidate and no invalidate)

test/gtest/uct/ib/test_ib_xfer.cc
```diff
@@ -129,13 +133,86 @@ UCS_TEST_P(uct_p2p_mix_test_alloc_methods, mix1000)
...
+        uint32_t mt_flag = UCT_IB_MEM_MULTITHREADED;
+        ASSERT_TRUE(((uct_ib_mem_t*)sendbuf.memh())->flags & mt_flag);
+        ASSERT_TRUE(((uct_ib_mem_t*)recvbuf.memh())->flags & mt_flag);
+    }
+
+    bool check_md_flags()
+    {
+#if HAVE_DEVX
+        auto *ib_mlx5_md = ucs_derived_of(sender().md(), uct_ib_mlx5_md_t);
```
yosefe, line 162: 1. do this after checking md is mlx5
2. check by md ops function pointer (need mlx5 devx md) rather than by MLX5 PRM flag

test/gtest/uct/ib/test_ib_xfer.cc
```diff
@@ -129,13 +133,86 @@ UCS_TEST_P(uct_p2p_mix_test_alloc_methods, mix1000)
...
+        modify_config("REG_MT_CHUNK", ucs::to_string(reg_mt_chunk));
+
+        uct_p2p_mix_test::init();
+
+        if (!check_md_flags()) {
+            UCS_TEST_SKIP_R("KSM and indirect atomics are required for MT "
+                            "registration");
+        }
+
+        /* Large buffer can cause failures because of too many chunks */
```
yosefe, line 187: why too many chunks is a failure?

test/gtest/uct/ib/test_ib_xfer.cc
```diff
@@ -129,13 +133,86 @@ UCS_TEST_P(uct_p2p_mix_test_alloc_methods, mix1000)
...
+        m_send_size = ucs_min(m_send_size, 256 * reg_mt_chunk);
+        /* We need at least two chunks */
+        m_send_size = ucs_max(m_send_size, reg_mt_chunk + 1);
+    }
+
+    virtual void cleanup() override
+    {
+        uct_p2p_mix_test::cleanup();
+
```
yosefe, line 196: remove space line

yosefe (commented):
need to make sure it does not increase testing time

test/gtest/uct/ib/test_ib_xfer.cc
ivankochin: I changed it to the following check:
```c
        auto *ib_md = ucs_derived_of(sender().md(), uct_ib_md_t);
        if (strcmp(ib_md->name, UCT_IB_MD_NAME(mlx5))) {
            return false;
        }
```

I think it should be enough, isn't it?

test/gtest/uct/test_p2p_mix.cc
```diff
@@ -27,47 +29,72 @@ void uct_p2p_mix_test::completion_callback(uct_completion_t *comp)
...
+void *uct_p2p_mix_test::atomic_get_send_addr(const mapped_buffer &buf,
+                                             size_t op_size) {
+    return buf.ptr();
+}
+
+template<>
+void *uct_p2p_mix_test::atomic_get_send_addr<true>(const mapped_buffer &buf,
+                                                   size_t op_size) {
+    return UCS_PTR_BYTE_OFFSET(buf.ptr(), buf.length() - op_size);
+}
```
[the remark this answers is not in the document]
ivankochin: The idea here is to use atomic operations that are performed on the last bytes of the big buffer to check that we registered the whole buffer space and all the pointer logic (when the iova not equal to address) works correctly.

test/gtest/uct/test_p2p_mix.cc
```diff
@@ -27,47 +29,72 @@ void uct_p2p_mix_test::completion_callback(uct_completion_t *comp)
...
 }
 
 ucs_status_t uct_p2p_mix_test::put_bcopy(const mapped_buffer &sendbuf,
                                          const mapped_buffer &recvbuf,
                                          uct_completion_t *comp)
 {
-    ssize_t packed_len;
-    packed_len = uct_ep_put_bcopy(sender().ep(0), mapped_buffer::pack,
-                                  (void*)&sendbuf, recvbuf.addr(), recvbuf.rkey());
+    bcopy_pack_arg pack_arg = {sendbuf.ptr(), m_max_bcopy, sendbuf.mem_type()};
```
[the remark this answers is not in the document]
ivankochin: Done.

test/gtest/uct/test_p2p_mix.cc
```diff
@@ -172,73 +206,92 @@ void uct_p2p_mix_test::run(unsigned count) {
     mapped_buffer sendbuf(m_send_size, 0, sender());
     mapped_buffer recvbuf(m_send_size, 0, receiver());
 
+    check_buffers(sendbuf, recvbuf);
```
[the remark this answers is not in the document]
ivankochin: Do you mean to make `mapped_buffer` constructor virtual? And then create some descendant `mapped_buffer_check_mt` with flags checking logic inside the constructor?

test/gtest/uct/test_p2p_mix.cc
```diff
@@ -172,73 +206,92 @@ void uct_p2p_mix_test::run(unsigned count) {
...
     if (sender().iface_attr().cap.atomic64.fop_flags & UCS_BIT(UCT_ATOMIC_OP_SWAP)) {
-        m_avail_send_funcs.push_back(&uct_p2p_mix_test::atomic_fop<uint64_t, UCT_ATOMIC_OP_SWAP>);
+        m_avail_send_funcs.push_back(&uct_p2p_mix_test::atomic_fop<uint64_t, UCT_ATOMIC_OP_SWAP, use_last_byte>);
     }
     if (sender().iface_attr().cap.atomic32.fop_flags & UCS_BIT(UCT_ATOMIC_OP_SWAP)) {
-        m_avail_send_funcs.push_back(&uct_p2p_mix_test::atomic_fop<uint32_t, UCT_ATOMIC_OP_SWAP>);
+        m_avail_send_funcs.push_back(&uct_p2p_mix_test::atomic_fop<uint32_t, UCT_ATOMIC_OP_SWAP, use_last_byte>);
+    }
+
+    if (sender().iface_attr().cap.atomic64.fop_flags & UCS_BIT(UCT_ATOMIC_OP_CSWAP)) {
```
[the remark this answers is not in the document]
ivankochin: Done.

test/gtest/uct/test_p2p_mix.cc
```diff
@@ -172,73 +206,92 @@ void uct_p2p_mix_test::run(unsigned count) {
...
+        m_avail_send_funcs.push_back(&uct_p2p_mix_test::cswap64<use_last_byte>);
+    }
+}
+
+size_t uct_p2p_mix_test::max_buffer_size() {
```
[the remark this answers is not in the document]
ivankochin: Done.

test/gtest/uct/test_p2p_mix.cc
```diff
@@ -172,73 +206,92 @@ void uct_p2p_mix_test::run(unsigned count) {
...
+    uct_p2p_test::init();
+    ucs_status_t status = uct_iface_set_am_handler(receiver().iface(), AM_ID,
+                                                   am_callback, NULL,
+                                                   UCT_CB_FLAG_ASYNC);
+    ASSERT_UCS_OK(status);
+
+    m_max_short = m_max_bcopy = m_max_zcopy = max_buffer_size();
+    if (sender().iface_attr().cap.flags & UCT_IFACE_FLAG_AM_SHORT) {
+        m_max_short = ucs_min(m_max_short, sender().iface_attr().cap.am.max_short);
+
```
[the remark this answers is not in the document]
ivankochin: Done.

test/gtest/uct/test_p2p_mix.cc
```diff
@@ -172,73 +206,92 @@ void uct_p2p_mix_test::run(unsigned count) {
...
+    if (sender().iface_attr().cap.flags & UCT_IFACE_FLAG_PUT_SHORT) {
+        m_avail_send_funcs.push_back(&uct_p2p_mix_test::put_short);
+        m_max_short = ucs_min(m_max_short, sender().iface_attr().cap.put.max_short);
+    }
+    if (sender().iface_attr().cap.flags & UCT_IFACE_FLAG_PUT_BCOPY) {
+        m_avail_send_funcs.push_back(&uct_p2p_mix_test::put_bcopy);
+        m_max_bcopy = ucs_min(m_max_bcopy, sender().iface_attr().cap.put.max_bcopy);
+    }
+
+    m_send_size = ucs_max(m_max_zcopy, ucs_max(m_max_short, m_max_bcopy));
```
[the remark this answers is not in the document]
ivankochin: Done!

test/gtest/uct/test_p2p_mix.h
```diff
@@ -56,18 +59,36 @@ class uct_p2p_mix_test : public uct_p2p_test {
...
+
     void run(unsigned count);
 
     virtual void init();
 
     virtual void cleanup();
 
+    size_t max_buffer_size();
+
+    size_t                   m_send_size;
```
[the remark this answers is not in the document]
ivankochin: We use it in `uct_p2p_mix_test_mt` so it should be protected.

test/gtest/uct/test_p2p_mix.h
```diff
@@ -56,18 +59,36 @@ class uct_p2p_mix_test : public uct_p2p_test {
...
+    virtual void
+    check_buffers(const mapped_buffer &sendbuf, const mapped_buffer &recvbuf);
+
     void run(unsigned count);
 
     virtual void init();
 
     virtual void cleanup();
 
+    size_t max_buffer_size();
```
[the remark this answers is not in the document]
ivankochin: Done.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -112,22 +112,46 @@ uct_ib_mlx5_devx_reg_ksm(uct_ib_mlx5_md_t *md, int atomic, uint64_t address,
...
 static ucs_status_t
-uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md, int atomic, void *address,
-                              uct_ib_mlx5_devx_ksm_data_t *ksm_data,
-                              size_t length, uint64_t iova, uint32_t mkey_index,
-                              const char *reason, struct mlx5dv_devx_obj **mr_p,
-                              uint32_t *mkey)
+uct_ib_mlx5_devx_reg_ksm_data_mt(uct_ib_mlx5_md_t *md, int atomic,
+                                 void *address,
+                                 uct_ib_mlx5_devx_ksm_data_t *ksm_data,
+                                 uint32_t iova_offset, uint32_t mkey_index,
```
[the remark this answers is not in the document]
yosefe: can we keep uct_ib_mlx5_devx_reg_ksm_data with iova_offset, but change it to iova in the lower-level functions that accept "void *address" as an argument?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -112,22 +112,46 @@ uct_ib_mlx5_devx_reg_ksm(uct_ib_mlx5_md_t *md, int atomic, uint64_t address,
...
     ucs_status_t status;
     char *in;
     void *klm;
     int i;
 
-    status = uct_ib_mlx5_alloc_mkey_inbox(ksm_data->mr_num, &in);
+    /* Add offset to workaround CREATE_MKEY range check issue */
+    if (iova_offset > 0) {
+        ++list_size;
+    }
```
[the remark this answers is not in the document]
yosefe: can we share the same code to workaround single and multi threaded cases?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -139,11 +163,18 @@ uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md, int atomic, void *address,
         mr_address = UCS_PTR_BYTE_OFFSET(mr_address, ksm_data->mrs[i]->length);
     }
 
-    status = uct_ib_mlx5_devx_reg_ksm(md, atomic, iova, length,
-                                      ksm_data->mr_num,
-                                      ksm_data->mrs[0]->length, in, mkey_index,
-                                      reason, mr_p, mkey);
+    if (iova_offset > 0) {
+        UCT_IB_MLX5DV_SET64(klm, klm, address, (uintptr_t)mr_address);
+        UCT_IB_MLX5DV_SET(klm, klm, mkey, last_mr->lkey);
```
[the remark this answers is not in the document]
yosefe: so could we remove it also from line 160?

test/gtest/uct/test_p2p_mix.cc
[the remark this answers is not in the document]
yosefe: the intention is to avoid adding mem_type() method to mem_buffer class

test/gtest/uct/test_p2p_mix.cc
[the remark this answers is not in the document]
yosefe: I mean instead of adding "check_buffers", add a virtual method "alloc_buffer"

test/gtest/uct/test_p2p_mix.cc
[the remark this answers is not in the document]
yosefe: right, but why need to change the operations themselves?
can we create a mapped_buffer that points to the last bytes of some other buffer?

test/gtest/uct/ib/test_ib_xfer.cc
yosefe: ?

yosefe (commented):
pls see replies to comments

test/gtest/uct/test_p2p_mix.cc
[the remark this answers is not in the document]
ivankochin: As we discussed there is no local `copy_from` method, so it can be left as it is done now.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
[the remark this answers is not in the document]
ivankochin: Removed from line 160.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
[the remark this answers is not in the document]
ivankochin: As I understand, this `ucs_div_round_up` covers also another workaround (for alignment) for single threaded registration case. MT registration still doesn't work in that case. 

So it may be complicated to unify them in this patch. Maybe later, when MT registration for unaligned buffers would work, it can be possible.

test/gtest/uct/test_p2p_mix.cc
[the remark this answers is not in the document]
ivankochin: Done.

test/gtest/uct/ib/test_ib_xfer.cc
ivankochin: Changed the comment.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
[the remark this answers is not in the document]
ivankochin: Done.

yosefe (commented):
remove use_last_byte from test_p2p_mix
otherwise, LGTM

test/gtest/uct/test_p2p_mix.cc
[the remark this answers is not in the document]
ivankochin: Done.

test/gtest/uct/ib/test_ib_xfer.cc
```diff
@@ -212,6 +212,12 @@ UCS_TEST_P(uct_p2p_mix_test_mt, mix1000)
     run(1000);
 }
 
+UCS_TEST_P(uct_p2p_mix_test_mt, mix1000_last_byte_offset)
+{
+    /* Alloc 2 chunks buffer, but perform all the operation on the last 8 bytes */
+    run(1000, reg_mt_chunk * 2 - 8, 8);
```
yosefe, line 218: use ( ) around * expression

test/gtest/uct/uct_test.cc
```diff
@@ -1411,6 +1416,23 @@ uct_test::mapped_buffer::mapped_buffer(size_t size, uint64_t seed,
...
+    m_mem.md       = other.m_mem.md;
+    m_mem.memh     = other.m_mem.memh;
+    m_mem.mem_type = other.m_mem.mem_type;
+    m_mem.length   = other.m_mem.length;
+    m_buf          = other.m_buf;
+    m_end          = other.m_end;
+    m_rkey.rkey    = other.m_rkey.rkey;
+    m_rkey.handle  = other.m_rkey.handle;
+
+    other.fill_empty();
```
yosefe, line 1433: rename fill_empty -> reset

yosefe (commented):
also there is a build failure

test/gtest/uct/uct_test.cc
ivankochin: Done.

test/gtest/uct/ib/test_ib_xfer.cc
ivankochin: Done.

yosefe (approved):

yosefe (approved):

yosefe (approved):

yosefe (dismissed):

yosefe:
@ivankochin can u pls squash?

ivankochin:
I did the rebase to resolve the merge conflicts. The only change is add `IB_` prefix for all the modified env variables in `test_ib_xfer`.

yosefe:
> I did the rebase to resolve the merge conflicts. The only change is add `IB_` prefix for all the modified env variables in `test_ib_xfer`.

Why not a merge commit?

ivankochin:
> > I did the rebase to resolve the merge conflicts. The only change is add `IB_` prefix for all the modified env variables in `test_ib_xfer`.
> 
> Why not a merge commit?

To patch was already squased to I did it to prevent merge commit creation and the squashing it since the merge conflict solution was straightforward for `test_ib_xfer`. But I completely forgot about required changres in `test_ib_md`, so I created one more commit yesterday evening.

yosefe:
> To patch was already squased so I did it to prevent merge commit creation and squashing it since the merge conflict solution was straightforward for `test_ib_xfer`. But I completely forgot about required changres in `test_ib_md`, so I created one more commit yesterday evening.

Even in this case, need to use a merge commit, since there is no good way to review the conflict resolution changes

yosefe (approved):

PR #9427, opened 2023-10-18
UCP/RNDV: Fixed rendezvous protocol query.

rakhmets:
## What
Fixed rendezvous protocol query.

## Why ?
Fixed proto info output.
Before:
```
+---------------------------+--------------------------------------------------------------+
| perftest intra-node cfg#2 | active message by ucp_am_send* from host memory              |
+---------------------------+-----------------------------------------------+--------------+
|                     0..92 | short                                         | sysv/memory  |
|                  93..2824 | copy-in                                       | sysv/memory  |
|                 2825..inf | (?) rendezvous copy from mapped remote memory | xpmem/memory |
+---------------------------+-----------------------------------------------+--------------+
+---------------------------+--------------------------------------------------------+
| perftest intra-node cfg#2 | active message by ucp_am_send*(rndv) from cuda/GPU0    |
+---------------------------+-----------------------------------------------------+--+
|                    0..inf | (?) rendezvous no data fetch                        |  |
+---------------------------+-----------------------------------------------------+--+
```
After:
```
+---------------------------+---------------------------------------------------------------------------------------------------------+
| perftest intra-node cfg#2 | active message by ucp_am_send* from host memory                                                         |
+---------------------------+--------------------------------------------------+------------------------------------------------------+
|                     0..92 | short                                            | sysv/memory                                          |
|                  93..2824 | copy-in                                          | sysv/memory                                          |
|               2825..91537 | (?) rendezvous copy from mapped remote memory    | xpmem/memory                                         |
|  91538..5713982720063549K | (?) rendezvous zero-copy flushed write to remote | 35% on knem/memory and 65% on rc_mlx5/mlx5_0:1/path0 |
|  5851118305345074177..inf | (?) rendezvous zero-copy read from remote        | 35% on knem/memory and 65% on rc_mlx5/mlx5_0:1/path0 |
+---------------------------+--------------------------------------------------+------------------------------------------------------+
+---------------------------+--------------------------------------------------------------------------------------------------------------------+
| perftest intra-node cfg#2 | active message by ucp_am_send*(rndv) from cuda/GPU0                                                                |
+---------------------------+--------------------------------------------------+-----------------------------------------------------------------+
|                         0 | (?) rendezvous no data fetch                     |                                                                 |
|                    1..159 | (?) rendezvous fragmented copy-in copy-out       | sysv/memory                                                     |
|                160..27214 | (?) rendezvous zero-copy read from remote        | 50% on rc_mlx5/mlx5_0:1/path0 and 50% on rc_mlx5/mlx5_1:1/path0 |
|                27215..inf | (?) rendezvous zero-copy flushed write to remote | cuda_ipc/cuda                                                   |
+---------------------------+--------------------------------------------------+-----------------------------------------------------------------+
```

### DIFF
--- a/src/ucp/rndv/proto_rndv.c
+++ b/src/ucp/rndv/proto_rndv.c
@@ -462,7 +462,7 @@ void ucp_proto_rndv_rts_query(const ucp_proto_query_params_t *params,
                                 params->msg_length, &remote_attr);
 
     attr->is_estimation  = 1;
-    attr->max_msg_length = SIZE_MAX;
+    attr->max_msg_length = remote_attr.max_msg_length;
     attr->lane_map       = UCS_BIT(rpriv->lane);
 
     ucs_snprintf_safe(attr->desc, sizeof(attr->desc), "rendezvous %s",


src/ucp/proto/proto_debug.c
```diff
@@ -198,6 +205,77 @@ static int ucp_proto_debug_is_info_enabled(ucp_context_h context,
...
+                     sizeof(proto_attr_dst->desc));
+    ucs_strncpy_safe(proto_attr_dst->config, proto_attr_src->config,
+                     sizeof(proto_attr_dst->config));
+    proto_attr_dst->is_estimation = proto_attr_src->is_estimation;
+    dst->range_start              = src->range_start;
+    dst->range_end                = src->range_end;
+}
+
+static int
+ucp_proto_debug_cmp_range_info(const ucp_proto_debug_range_info_t *range_info_1,
```
brminich, line 268: maybe `ucp_proto_debug_range_info_equal`? Or return zero if they compare equal to be aligned with standard `cmp` functions.

src/ucp/proto/proto_debug.c
```diff
@@ -198,6 +205,77 @@ static int ucp_proto_debug_is_info_enabled(ucp_context_h context,
...
+                               const ucp_proto_debug_range_info_t *range_info_2)
+{
+    const ucp_proto_query_attr_t *proto_attr_1 = &range_info_1->proto_attr;
+    const ucp_proto_query_attr_t *proto_attr_2 = &range_info_2->proto_attr;
+
+    return proto_attr_1->is_estimation == proto_attr_2->is_estimation &&
+           strcmp(proto_attr_1->config, proto_attr_2->config) == 0 &&
+           strcmp(proto_attr_1->desc, proto_attr_2->desc) == 0;
```
brminich, line 276:
```suggestion
    return (proto_attr_1->is_estimation == proto_attr_2->is_estimation) &&
           (strcmp(proto_attr_1->config, proto_attr_2->config) == 0) &&
           (strcmp(proto_attr_1->desc, proto_attr_2->desc) == 0);
```

src/ucp/proto/proto_debug.c
```diff
@@ -198,6 +205,77 @@ static int ucp_proto_debug_is_info_enabled(ucp_context_h context,
...
+
+static int
+ucp_proto_debug_cmp_range_info(const ucp_proto_debug_range_info_t *range_info_1,
+                               const ucp_proto_debug_range_info_t *range_info_2)
+{
+    const ucp_proto_query_attr_t *proto_attr_1 = &range_info_1->proto_attr;
+    const ucp_proto_query_attr_t *proto_attr_2 = &range_info_2->proto_attr;
+
+    return proto_attr_1->is_estimation == proto_attr_2->is_estimation &&
+           strcmp(proto_attr_1->config, proto_attr_2->config) == 0 &&
```
brminich, line 275: do you assume that `lane_map` is also equal if configs are equal? If yes, can we add some assert ?

src/ucp/proto/proto_debug.c
```diff
@@ -198,6 +205,77 @@ static int ucp_proto_debug_is_info_enabled(ucp_context_h context,
...
+    const ucp_proto_query_attr_t *proto_attr_src = &src->proto_attr;
+    ucp_proto_query_attr_t *proto_attr_dst       = &dst->proto_attr;
+
+    ucs_strncpy_safe(proto_attr_dst->desc, proto_attr_src->desc,
+                     sizeof(proto_attr_dst->desc));
+    ucs_strncpy_safe(proto_attr_dst->config, proto_attr_src->config,
+                     sizeof(proto_attr_dst->config));
+    proto_attr_dst->is_estimation = proto_attr_src->is_estimation;
+    dst->range_start              = src->range_start;
+    dst->range_end                = src->range_end;
```
brminich, line 264: maybe copy the whole `ucp_proto_query_attr_t`?

src/ucp/proto/proto_debug.c
```diff
@@ -190,6 +192,76 @@ static int ucp_proto_debug_is_info_enabled(ucp_context_h context,
...
+static int
+ucp_proto_debug_get_range_info(ucp_worker_h worker,
+                               const ucp_proto_select_elem_t *select_elem,
+                               size_t msg_length, size_t *range_end,
+                               ucp_proto_query_attr_t *proto_attr)
+{
+    const ucp_proto_perf_range_t *range;
+
+    range      = ucp_proto_perf_range_search(select_elem, msg_length);
+    *range_end = range->max_length;
```
yosefe, line 204: can we say this fix is mostly this (set range_end to range->max_length instead of proto_attr.max_msg_length) and the rest is mostly refactoring?
In this case, can we fix rndv protocol query to set proto_attr.max_msg_length in a way that will distinguish between different remote protocols?

rakhmets: If we only change `proto_attr.max_msg_length` by `range->max_length`, then we will get many duplicated rows. Because there can be many ranges for the same combination of protocol and device.

Yes, I agree. The query for rendezvous should be fixed instead. Currently we set `ucp_proto_query_attr_t::max_msg_length` to `SIZE_MAX` in query method. Which is not correct.

yosefe: yes, we need to set it to the max range that the same protocol (including remote) is used. this way, will also not have duplicates.

src/ucp/proto/proto_debug.c
rakhmets: Reverted changes.

src/ucp/proto/proto_debug.c
rakhmets: Reverted.

src/ucp/proto/proto_debug.c
rakhmets: Reverted.

src/ucp/proto/proto_debug.c
rakhmets: Fixed query.

brminich (approved):

brminich:
can you please squash?

PR #9433, opened 2023-10-23
AZP: Handle multiple log files

Alexey-Rivkin:
## What
Save PerfX log file per job. 
Preserve as artifacts and show in the build log.

### Before:
perfx.log got overwritten, so only the last one was handled.

### DIFF
--- a/buildlib/azure-pipelines-perf.yml
+++ b/buildlib/azure-pipelines-perf.yml
@@ -101,17 +101,14 @@ stages:
             workingDirectory: $(WorkDir)
 
           - bash: |
-              set -x
-              cat perfx.log
-            displayName: Print raw log
+              set -xe
+              sep_line="######################"
+              for file in *.log; do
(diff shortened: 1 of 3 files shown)


buildlib/azure-pipelines-perf.yml
```diff
@@ -101,19 +101,16 @@ stages:
             workingDirectory: $(WorkDir)
 
           - bash: |
-              set -x
-              cat perfx.log
-            displayName: Print raw log
+              set -xe
+              sep_line="######################"
+              for file in *.log; do
```
artemry-nv, line 106: Check it's safe according to shellcheck

artemry-nv (approved):

buildlib/azure-pipelines-perf.yml
Alexey-Rivkin: It is

yosefe (approved):

PR #9440, opened 2023-10-24
UCP/TAG: Populate tag_info on immediate completion

### LINKED ISSUE #9425: TAG API doesn't populate `ucp_tag_recv_info_t` struct
### Describe the bug
According to the [`ucp_tag_recv_nbx` docs](https://openucx.readthedocs.io/en/master/api/file/ucp_8h.html?highlight=ucp_tag_recv_nbx#_CPPv416ucp_tag_recv_nbx12ucp_worker_hPv6size_t9ucp_tag_t9ucp_tag_tPK19ucp_request_param_t), we should get message details in `param.recv_info.tag_info` if the request is completed immediately:
> `NULL` - The receive operation was completed immediately. In this case, if `param->recv_info.tag_info` is specified in the param, the value to which it points is updated with the information about the received message.

However, the following example always prints that both `recv_info.length` and `recv_info.sender_tag` are `0` (as per initial value) after the message completes immediately.

```c++
ucp_request_param_t param{0};
param.op_attr_mask =
  UCP_OP_ATTR_FIELD_CALLBACK | UCP_OP_ATTR_FIELD_RECV_INFO | UCP_OP_ATTR_FIELD_USER_DATA;
param.user_data = user_data;
param.cb.recv   = cb;

ucp_tag_recv_info_t recv_info{0};
param.recv_info.tag_info = &recv_info;

request = ucp_tag_recv_nbx(worker_handle, buffer, length, tag, tagMask, &param);

if (UCS_PTR_STATUS((ucs_status_ptr_t)(request)) == UCS_OK)
  ucxx_warn("Request completed, length: %lu (expected: %lu), tag: 0x%x (expected: 0x%x)",
            recv_info.length,
            length,
            recv_info.sender_tag,
            tag);
```

### Setup and versions
- UCX `v1.14.1` and `v1.15.0`

### Additional information (depending on the issue)
According to @brminich , this seems to be unimplemented currently.

tvegas1:
## What
Return `tag_info` data on immediate completion when requested.

Fixes #9425 

## Why ?
Match existing documentation:
`NULL - The receive operation was completed immediately. In this case, if param->recv_info.tag_info is specified in the param, the value to which it points is updated with the information about the received message.`
## How ?
Populate field on immediate completion when receive info is passed and immediate completion is not prevented.

### DIFF
--- a/src/ucp/tag/tag_recv.c
+++ b/src/ucp/tag/tag_recv.c
@@ -103,7 +103,18 @@ static UCS_F_ALWAYS_INLINE ucs_status_ptr_t ucp_tag_recv_common(
         req->status = status;
         UCS_PROFILE_REQUEST_EVENT(req, "complete_imm_tag_recv", 0);
 
-        ucp_request_imm_cmpl_param(param, req, recv, &req->recv.tag.info);
+        if (param->op_attr_mask & UCP_OP_ATTR_FLAG_NO_IMM_CMPL) {
+            return ucp_request_prevent_imm_cmpl(param, req, recv,
+                                                &req->recv.tag.info);
+        }
+
+        if ((param->op_attr_mask & UCP_OP_ATTR_FIELD_RECV_INFO) &&
+            (status == UCS_OK)) {
+            *param->recv_info.tag_info = req->recv.tag.info;
+        }
+
+        ucp_request_put_param(param, req);
+        return UCS_STATUS_PTR(status);
     }
 
     /* TODO: allocate request only in case if flag
--- a/test/gtest/ucp/test_ucp_tag.cc
+++ b/test/gtest/ucp/test_ucp_tag.cc
@@ -306,6 +306,11 @@ test_ucp_tag::recv(entity &receiver, recv_type_t type, void *buffer,
     param.datatype     = datatype;
 
     switch (type) {
+    case RECV_IMM:
+        param.op_attr_mask      &= ~UCP_OP_ATTR_FLAG_NO_IMM_CMPL;
+        param.op_attr_mask      |= UCP_OP_ATTR_FIELD_RECV_INFO;
+        param.recv_info.tag_info = info;
+        // Fallthrough
     case RECV_B:
     case RECV_NB:
         param.op_attr_mask |= UCP_OP_ATTR_FIELD_CALLBACK |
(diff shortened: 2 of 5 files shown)


src/ucp/tag/tag_recv.c
```diff
@@ -100,6 +100,12 @@ static UCS_F_ALWAYS_INLINE ucs_status_ptr_t ucp_tag_recv_common(
                                                  recv_len, 1, param);
         ucp_recv_desc_release(rdesc);
 
+        if ((param->op_attr_mask & UCP_OP_ATTR_FIELD_RECV_INFO) &&
+            !(param->op_attr_mask & UCP_OP_ATTR_FLAG_NO_IMM_CMPL) &&
```
yosefe, line 104: ucp_request_imm_cmpl_param already checks UCP_OP_ATTR_FLAG_NO_IMM_CMPL flag, can we unite with it? to avoid extra branch

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -306,6 +306,11 @@ test_ucp_tag::recv(entity &receiver, recv_type_t type, void *buffer,
     param.datatype     = datatype;
 
     switch (type) {
+    case RECV_IMM:
+        param.op_attr_mask      &= ~UCP_OP_ATTR_FLAG_NO_IMM_CMPL;
+        param.op_attr_mask      |= UCP_OP_ATTR_FIELD_RECV_INFO;
+        param.recv_info.tag_info = info;
+        // fallthrough
```
yosefe, line 313: capital case

src/ucp/tag/tag_recv.c
tvegas1: split macro and using first half

test/gtest/ucp/test_ucp_tag.cc
tvegas1: fixed

yosefe (approved):

tvegas1:
do we need to backport to v1.15 or v1.14?

yosefe:
> do we need to backport to v1.15 or v1.14?

@pentschev ?

brminich (approved):

pentschev:
I think we can live without it until v1.16, so let's not worry about backporting it. If something changes I'll then open a backport PR for it.

pentschev:
And thanks @tvegas1 for working on this!

PR #9441, opened 2023-10-25
UCS/RCACHE: Add support for dynamic region alignment

Artemy-Mellanox:
## Why ?
Some transports (e.g. gdrcopy) require special alignment. 

## How ?
Regions in the UCP rcache will select the maximum alignment from the MD registrations they comprise. When merging regions, the rcache will also choose the maximum alignment.

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -582,9 +582,10 @@ ucp_memh_create(ucp_context_h context, void *address, size_t length,
     return UCS_OK;
 }
 
-static ucs_status_t
-ucp_memh_rcache_get(ucs_rcache_t *rcache, void *address, size_t length,
-                    ucs_memory_type_t mem_type, ucp_mem_h *memh_p)
+static ucs_status_t ucp_memh_rcache_get(ucs_rcache_t *rcache, void *address,
+                                        size_t length,
+                                        ucs_memory_type_t mem_type,
+                                        size_t alignment, ucp_mem_h *memh_p)
 {
     ucp_mem_attr_t attr = {
         .mem_type = mem_type
@@ -592,8 +593,8 @@ ucp_memh_rcache_get(ucs_rcache_t *rcache, void *address, size_t length,
     ucs_rcache_region_t *rregion;
     ucs_status_t status;
 
-    status = ucs_rcache_get(rcache, address, length, PROT_READ | PROT_WRITE,
-                            &attr, &rregion);
+    status = ucs_rcache_get(rcache, address, length, alignment,
+                            PROT_READ | PROT_WRITE, &attr, &rregion);
     if (status != UCS_OK) {
         return status;
     }
@@ -713,26 +714,18 @@ err:
     return status;
 }
 
-static void
-ucp_memh_reg_align(ucp_context_h context, void **address_p, size_t *length_p,
-                   ucp_md_map_t reg_md_map)
+static size_t ucp_memh_reg_align(ucp_context_h context, ucp_md_map_t reg_md_map)
 {
     size_t reg_align = UCS_RCACHE_MIN_ALIGNMENT;
     const uct_md_attr_v2_t *md_attr;
     ucp_md_index_t md_index;
-    void *start, *end;
 
     ucs_for_each_bit(md_index, reg_md_map) {
         md_attr   = &context->tl_mds[md_index].attr;
         reg_align = ucs_max(md_attr->reg_alignment, reg_align);
     }
 
-    start = ucs_align_down_pow2_ptr(*address_p, reg_align);
-    end   = ucs_align_up_pow2_ptr(UCS_PTR_BYTE_OFFSET(*address_p, *length_p),
-                                  reg_align);
-
-    *address_p = start;
-    *length_p  = UCS_PTR_BYTE_DIFF(start, end);
+    return reg_align;
 }
 
 ucs_status_t ucp_memh_get_slow(ucp_context_h context, void *address,
@@ -745,6 +738,7 @@ ucs_status_t ucp_memh_get_slow(ucp_context_h context, void *address,
     size_t reg_length;
     ucs_status_t status;
     ucs_memory_info_t mem_info;
+    size_t reg_align;
 
     if (context->config.ext.reg_whole_alloc_bitmap & UCS_BIT(mem_type)) {
         ucp_memory_detect_internal(context, address, length, &mem_info);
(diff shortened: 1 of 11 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -731,8 +732,9 @@ ucp_memh_reg_align(ucp_context_h context, void **address_p, size_t *length_p,
     end   = ucs_align_up_pow2_ptr(UCS_PTR_BYTE_OFFSET(*address_p, *length_p),
                                   reg_align);
 
-    *address_p = start;
-    *length_p  = UCS_PTR_BYTE_DIFF(start, end);
+    *address_p   = start;
+    *length_p    = UCS_PTR_BYTE_DIFF(start, end);
```
yosefe, line 736: do we still need to modify address/len? i'd expect rcache to do it now

src/ucs/memory/rcache.c
```diff
@@ -780,13 +776,18 @@ ucs_rcache_check_overlap(ucs_rcache_t *rcache, ucs_pgt_addr_t *start,
     /* coverity[double_unlock] */
     ucs_rcache_check_gc_list(rcache, 1);
 
+    ucs_list_head_init(&region_list);
+    ucs_trace("find region1 %lx..%lx", *start, *end);
```
yosefe, line 780: move this log to ucs_rcache_find_regions ad print rcache name

src/ucs/memory/rcache.c
```diff
@@ -844,14 +845,28 @@ ucs_rcache_check_overlap(ucs_rcache_t *rcache, ucs_pgt_addr_t *start,
...
                 rcache, region, UCS_RCACHE_REGION_PUT_FLAG_IN_PGTABLE);
     }
+
+    ucs_list_head_init(&region_list);
+    ucs_rcache_find_regions(rcache, *start, old_start, &region_list);
+    ucs_rcache_find_regions(rcache, old_end, *end - 1, &region_list);
+
+    if (!ucs_list_is_empty(&region_list)) {
+        lookup = 0;
+        goto again;
```
yosefe, line 867: add comment

test/gtest/ucs/test_rcache.cc
```diff
@@ -717,7 +718,6 @@ class test_rcache_with_limit : public test_rcache {
         ucs_rcache_params_t params = test_rcache::rcache_params();
         params.max_regions         = 2;
         params.max_size            = 1000;
-        params.alignment           = 16;
```
yosefe, line 720: let's add some gtests for alignment

src/ucp/core/ucp_mm.c
Artemy-Mellanox: currently, it does it during region creation and lookup may return a region without proper alignment

yosefe: after this PR i'd expect to get address/len of the region from rcache, also we may have a merge..

Artemy-Mellanox: what if a region for a not-yet-aligned address/len already exists in rcache? it might have been created with a smaller alignment e.g. without gdrcopy in md_map and now we are adding gdrcopy. in this case lookup will return it as is and gdrcopy reg_mem will receive it unaligned

yosefe: maybe rcache lookup should fallback to slowpath if region->alignment is less than alignment param

src/ucp/core/ucp_mm.c
```diff
@@ -1702,6 +1681,7 @@ ucp_memh_import(ucp_context_h context, const void *export_mkey_buffer,
 
             rregion = ucs_rcache_lookup_unsafe(rcache, unpacked_memh.address,
                                                unpacked_memh.length,
+                                               UCS_RCACHE_MIN_ALIGNMENT,
```
yosefe, line 1684: maybe pass 1 ?

src/ucp/core/ucp_mm.inl
```diff
@@ -50,8 +66,9 @@ ucp_memh_get(ucp_context_h context, void *address, size_t length,
 
     if (ucs_likely(context->rcache != NULL)) {
         UCP_THREAD_CS_ENTER(&context->mt_lock);
-        rregion = ucs_rcache_lookup_unsafe(context->rcache, address, length,
-                                           PROT_READ | PROT_WRITE);
+        reg_align = ucp_memh_reg_align(context, reg_md_map);
```
yosefe, line 69: seems this adds extra overhead to rcache lookup

src/ucs/memory/rcache.c
```diff
@@ -858,8 +855,11 @@ ucs_rcache_check_overlap(ucs_rcache_t *rcache, ucs_pgt_addr_t *start,
                 rcache, region, UCS_RCACHE_REGION_PUT_FLAG_IN_PGTABLE);
     }
 
+    /* Overlapping regions might necessitate a larger alignment, which,
+     * in turn, can result in even more overlapping regions.
+     */
     ucs_list_head_init(&region_list);
-    ucs_rcache_find_regions(rcache, *start, old_start, &region_list);
+    ucs_rcache_find_regions(rcache, *start, old_start - 1, &region_list);
```
yosefe, line 862: why changed to -1?

Artemy-Mellanox: ucs_rcache_find_regions() expects an inclusive end, see other uses.

src/ucs/memory/rcache.inl
```diff
@@ -14,7 +14,7 @@ ucs_rcache_region_test(ucs_rcache_region_t *region, int prot, size_t alignment)
 {
     return (region->flags & UCS_RCACHE_REGION_FLAG_REGISTERED) &&
            ucs_test_all_flags(region->prot, prot) &&
-           (region->alignment >= alignment);
+           ((alignment == 1) || (region->alignment >= alignment));
```
yosefe, line 17: why need special case for alignment == 1? is it just for optimization?
maybe pass alignment==0 and then compiler could deduce the condition is always true because region->alignment is unsigned?

src/ucs/memory/rcache.c
```diff
@@ -844,14 +842,31 @@ ucs_rcache_check_overlap(ucs_rcache_t *rcache, ucs_pgt_addr_t *start,
...
                                 "merge 0x%lx..0x%lx "UCS_RCACHE_PROT_FMT" with",
                                 *start, *end, UCS_RCACHE_PROT_ARG(*prot));
-        *start  = ucs_min(*start, region->super.start);
-        *end    = ucs_max(*end,   region->super.end);
-        *merged = 1;
+        *alignment = ucs_max(*alignment, region->alignment);
+        *start     = ucs_align_down_pow2(ucs_min(*start, region->super.start),
+                                         *alignment);
+        *end       = ucs_align_up_pow2(ucs_max(*end, region->super.end),
+                                       *alignment);
```
yosefe, line 849: IMO we don't need to align up/down here, but only after finished going over all overlapping regions

src/ucs/memory/rcache.c
yosefe: can we write the code as while loop that stops when the list of overlapping regions is empty?

yosefe (approved):

PR #9443, opened 2023-10-26
UCP/GTEST: Use strong fence with protov2

brminich:
## What
Use strong fence with proto V2

## Why ?
With proto v2 UCP can use different RMA lanes for different sizes, thus fence can't be a no-op

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -2061,7 +2061,8 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
     context->config.worker_strong_fence =
             (context->config.ext.fence_mode == UCP_FENCE_MODE_STRONG) ||
             ((context->config.ext.fence_mode == UCP_FENCE_MODE_AUTO) &&
-             (context->config.ext.max_rma_lanes > 1));
+             ((context->config.ext.max_rma_lanes > 1) ||
+              context->config.ext.proto_enable));
 
     return UCS_OK;
 
--- a/test/gtest/ucp/test_ucp_rma.cc
+++ b/test/gtest/ucp/test_ucp_rma.cc
@@ -364,3 +364,63 @@ UCS_TEST_P(test_ucp_rma_reg_nb, get_nonblocking) {
 }
 
 UCP_INSTANTIATE_TEST_CASE(test_ucp_rma_reg_nb)
+
+
+class test_ucp_rma_order : public test_ucp_rma {
+public:
+    static void get_test_variants(std::vector<ucp_test_variant>& variants) {
+        add_variant(variants, UCP_FEATURE_RMA);
+    }
+
+    virtual void init() {
+        test_ucp_memheap::init();
+    }
+
+    void put_nbx(void *sbuf, size_t size, uint64_t target, ucp_rkey_h rkey,
+                 uint32_t flags) {
+        ucp_request_param_t param = { .op_attr_mask = flags };
+
+        ucs_status_ptr_t sptr = ucp_put_nbx(sender().ep(), sbuf, size, target,
+                                            rkey, &param);
+        request_release(sptr);
+    }
+
+    void test_ordering(size_t size, uint32_t put_flags) {
+        mem_buffer sbuf(size, UCS_MEMORY_TYPE_HOST);
+        mapped_buffer rbuf(size, receiver());
+
+        rbuf.memset(0);
+        sbuf.memset(CHAR_MAX);
+
+        ucs::handle<ucp_rkey_h> rkey;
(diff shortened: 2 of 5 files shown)


test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -364,3 +364,63 @@ UCS_TEST_P(test_ucp_rma_reg_nb, get_nonblocking) {
...
+        for (uint8_t iter = 0; iter < 50; ++iter) {
+            put_nbx(sbuf.ptr(), sbuf.size(), (uint64_t)rbuf.ptr(), rkey,
+                    put_flags);
+            sender().fence();
+            // Update first and last bytes of the target buffer and make sure
+            // it is not overwritten by the sender buffer (note fence() above)
+            put_nbx(&iter, sizeof(iter), (uint64_t)last, rkey, put_flags);
+            put_nbx(&iter, sizeof(iter), (uint64_t)first, rkey, put_flags);
+            flush_workers();
+            EXPECT_EQ(*first, iter);
```
tvegas1, line 411: add message with size on assert failure?

test/gtest/ucp/ucp_test.cc
```diff
@@ -1236,6 +1236,15 @@ ucs::handle<ucp_rkey_h> ucp_test::mapped_buffer::rkey(const entity& entity) cons
     return ucs::handle<ucp_rkey_h>(rkey, ucp_rkey_destroy);
 }
 
+void ucp_test::mapped_buffer::rkey(const entity& entity, ucs::handle<ucp_rkey_h> &rkey) const
+{
+    ucp_rkey_h ucp_rkey;
+
+    ucs_status_t status = ucp_ep_rkey_unpack(entity.ep(), m_rkey_buffer, &ucp_rkey);
+    ASSERT_UCS_OK(status);
```
tvegas1, line 1244: could remove `status`

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -364,3 +364,63 @@ UCS_TEST_P(test_ucp_rma_reg_nb, get_nonblocking) {
...
+        add_variant(variants, UCP_FEATURE_RMA);
+    }
+
+    virtual void init() {
+        test_ucp_memheap::init();
+    }
+
+    void put_nbx(void *sbuf, size_t size, uint64_t target, ucp_rkey_h rkey,
+                 uint32_t flags) {
+        ucp_request_param_t param = {};
```
tvegas1, line 381: maybe `param = { .op_attr_mask = flags }`

tvegas1 (approved):

src/ucp/core/ucp_context.c
```diff
@@ -2061,7 +2061,8 @@ static ucs_status_t ucp_fill_config(ucp_context_h context,
     context->config.worker_strong_fence =
             (context->config.ext.fence_mode == UCP_FENCE_MODE_STRONG) ||
             ((context->config.ext.fence_mode == UCP_FENCE_MODE_AUTO) &&
-             (context->config.ext.max_rma_lanes > 1));
+             ((context->config.ext.max_rma_lanes > 1) ||
+              (context->config.ext.proto_enable)));
```
yosefe, line 2065: extra ( )

tvegas1 (approved):

test/gtest/ucp/test_ucp_wireup.cc
```diff
@@ -304,12 +304,14 @@ void test_ucp_wireup::send_nb(ucp_ep_h ep, size_t length, int repeat,
             reqs.push_back(req);
         }
 
-        ucs_status_t status = ucp_worker_fence(ep->worker);
-        ASSERT_UCS_OK(status);
+        ucp_request_param_t param = {};
+        void *req = ucp_ep_flush_nbx(ep, &param);
+        ASSERT_UCS_PTR_OK(req);
```
yosefe, line 309: why changed fence to flush?

brminich: otherwise it hangs with ud/tcp, because our strong fence is blocking and we need to progress both peers for it to complete

yosefe: this can cause a deadlock with MPI/shmem as well
can we make the fence non-blocking? switch to wireup_ep like we when need to resolve remote ep?

brminich: why is it a problem? with emulation we need progress on both sides for other operations as well

yosefe: add FIXME comment why it's a flush and not a fence

yosefe (approved):
pls squash

PR #9444, opened 2023-10-26
GIT/GITIGNORE: Add test_dlopen and test_no_cuda_ctx

ivankochin:
## What
Add new entries to `.gitignore` which are automatically created during the UCX build:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,8 @@ contrib/ucx
 bindings/java/pom.xml
 src/ucs/ucs_stats_parser
 test/gtest/gtest
+test/gtest/ucm/test_dlopen/.noinst/
+test/gtest/ucm/test_dlopen/rpath-subdir/.noinst/
 build-*
 ucx*tar.gz
 src/tools/info/build_config.h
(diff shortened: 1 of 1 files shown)


tvegas1 (approved):

.gitignore
```diff
@@ -42,6 +42,8 @@ contrib/ucx
 bindings/java/pom.xml
 src/ucs/ucs_stats_parser
 test/gtest/gtest
+test/gtest/ucm/test_dlopen/.noinst/
+test/gtest/ucm/test_dlopen/rpath-subdir/.noinst/
```
yosefe, line 46: how were these files created? should we ignore them also in other directories?

ivankochin: As I understand, these two directories were created because only two these makefiles set `noinst_libdir` variable for some purpose. Yeah, I think it may be useful to ignore them in other directories too.

ivankochin:
This MR was closed because the feature branch was accidentally created in the openucx/ucx repo instead of fork one. Pls see #9448

PR #9447, opened 2023-10-26
UCT/IB: Skip multi-thread memory registration for symmetric key

tvegas1:
## What
Skip multi-threaded memory registration when symmetric key is requested and FW supports it.

## Why ?
Simplest approach, could later be improved to generate KSM(MT) with specific mkey_index using KSM wrapper introduced by related #9424.

## How ?
Add corresponding tests and assertion.

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -556,6 +556,8 @@ static void uct_ib_mlx5_devx_reg_symmetric(uct_ib_mlx5_md_t *md,
     uint32_t symmetric_rkey;
     ucs_status_t status;
 
+    ucs_assert(!(memh->super.flags & UCT_IB_MEM_MULTITHREADED));
+
     /* Best effort, only allocate in the range below the atomic keys. */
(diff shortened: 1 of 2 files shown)


src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -556,6 +556,8 @@ static void uct_ib_mlx5_devx_reg_symmetric(uct_ib_mlx5_md_t *md,
     uint32_t symmetric_rkey;
     ucs_status_t status;
 
+    ucs_assert((memh->super.flags & UCT_IB_MEM_MULTITHREADED) == 0);
```
brminich, line 559:
```suggestion
    ucs_assert(!(memh->super.flags & UCT_IB_MEM_MULTITHREADED));
```

brminich (dismissed):
👍 besides minor comment

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
tvegas1: force pushed

brminich (approved):

ivankochin (approved):

yosefe (approved):

PR #9449, opened 2023-10-27
UCT/MD/IB: Support MT KSM registration for unaligned buffers

ivankochin:
## What
Now we do multithreading registration without alignment of the memory region which can cause errors during KSM creation. This patch would register chunk with regard to the alignement.

Current state:
image

After the patch:
image

# How?

1. Register first direct MR up to `md->config.mt_reg_chunk` alignment to meet this alignment for all further MRs.
2. Register N bytes before first direct MR to meet `md->dev.atomic_align` alignement.

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -230,6 +230,7 @@ typedef struct {
     uct_ib_md_t                   *md;
     void                          *address;
     size_t                        length;
+    size_t                        first_mr_size;
     const uct_md_mem_reg_params_t *params;
     uint64_t                      access_flags;
     struct ibv_mr                 **mrs;
@@ -297,14 +298,14 @@ uct_ib_md_print_mem_reg_err_msg(const char *title, void *address, size_t length,
 void *uct_ib_md_mem_handle_thread_func(void *arg)
 {
     uct_ib_md_mem_reg_thread_t *ctx = arg;
-    size_t max_chunk                = ctx->md->config.mt_reg_chunk;
+    size_t chunk_size               = ctx->md->config.mt_reg_chunk;
     ucs_time_t UCS_V_UNUSED t0      = ucs_get_time();
+    void UCS_V_UNUSED *start        = ctx->address;
     int mr_idx                      = 0;
+    size_t length                   = ctx->first_mr_size;
     ucs_status_t status;
-    size_t length;
 
     while (ctx->length > 0) {
-        length = ucs_min(ctx->length, max_chunk);
         if (ctx->params != NULL) {
             status = uct_ib_reg_mr(ctx->md, ctx->address, length, ctx->params,
                                    ctx->access_flags, NULL, &ctx->mrs[mr_idx]);
@@ -319,12 +320,13 @@ void *uct_ib_md_mem_handle_thread_func(void *arg)
         }
         ctx->address = UCS_PTR_BYTE_OFFSET(ctx->address, length);
         ctx->length -= length;
+        length       = ucs_min(ctx->length, chunk_size);
         mr_idx++;
     }
 
-    ucs_trace("%s %p..%p took %f usec\n",
+    ucs_trace("%s %p..%p (first_mr_size %zu) took %f usec\n",
               (ctx->params != NULL) ? "reg_mr" : "dereg_mr",
-              ctx->mrs[0]->addr, ctx->address,
+              start, ctx->address, ctx->first_mr_size,
               ucs_time_to_usec(ucs_get_time() - t0));
     return UCS_STATUS_PTR(UCS_OK);
 
@@ -339,10 +341,10 @@ err:
 ucs_status_t
 uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
                             const uct_md_mem_reg_params_t *params,
-                            uint64_t access_flags, struct ibv_mr **mrs)
+                            uint64_t access_flags, size_t mr_num,
+                            struct ibv_mr **mrs)
 {
     size_t chunk_size = md->config.mt_reg_chunk;
-    int mr_num        = ucs_div_round_up(length, chunk_size);
     int thread_num_mrs, thread_num, thread_idx, mr_idx, cpu_id;
     ucs_sys_cpuset_t parent_set, thread_set;
     uct_ib_md_mem_reg_thread_t *ctxs, *ctx;
@@ -350,6 +352,8 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
     pthread_attr_t attr;
     ucs_status_t status;
     void *thread_status;
+    uint64_t offset;
+    size_t padding;
     int ret;
 
     status = ucs_sys_pthread_getaffinity(&parent_set);
@@ -378,6 +382,7 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
     status = UCS_OK;
     mr_idx = 0;
     cpu_id = 0;
+    offset = 0;
     for (thread_idx = 0; thread_idx < thread_num; thread_idx++) {
         /* calculate number of mrs for each thread so each one will
          * get proportional amount */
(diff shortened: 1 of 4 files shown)


src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -522,17 +538,193 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_atomic_key,
...
+    return UCS_STATUS_PTR(status);
+}
+
+static ucs_status_t
+uct_ib_mlx5_devx_handle_mr_list_mt(uct_ib_md_t *md, void *address,
+                                  const uct_md_mem_reg_params_t *params,
+                                  uint64_t access_flags,
+                                  uct_ib_mlx5_devx_ksm_data_t *ksm_data)
+{
+    size_t chunk_size    = md->config.mt_reg_chunk;
```
ivankochin, line 598:
```suggestion
    size_t chunk_size = md->config.mt_reg_chunk;
```

yosefe:
ASAN error seems relevant:
```
[     INFO ] memory_type:host 65 53687091 =================================================================
==24766==ERROR: AddressSanitizer: heap-use-after-free on address 0x60600104a690 at pc 0x7f8ef3572d14 bp 0x7f8ee3c8fbe0 sp 0x7f8ee3c8fbd0
READ of size 8 at 0x60600104a690 thread T1729
    #0 0x7f8ef3572d13 in uct_ib_mlx5_devx_mem_handle_thread_func /__w/2/s/contrib/../src/uct/ib/mlx5/dv/ib_mlx5dv_md.c:578
    #1 0x7f8ef2fcbb42 in start_thread nptl/pthread_create.c:442
    #2 0x7f8ef305cbb3 in __clone (/lib/x86_64-linux-gnu/libc.so.6+0x125bb3)
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -522,17 +538,194 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_atomic_key,
                                          &memh->atomic_rkey);
 }
 
+void *uct_ib_mlx5_devx_mem_handle_thread_func(void *arg)
```
yosefe, line 541: why needed to move these functions to ib_mlx5dv_md.c?

ivankochin: 1. To make `ksm_data` an argument for `uct_ib_mlx5_devx_handle_mr_list_mt` since this structure contains `length`, `mr_num` and `mrs[]` arguments that are required in this function.
2. `ib_mlx5dv_md.c` is the only place where these functions are used, so I decided that this would be better place for them.

yosefe: 1. Why is it needed to add ksm_data parameter to these functions to fix alignment?
2. We discussed it a while ago with @Artemy-Mellanox  and since these functions are generic we decided to keep them in ib_ms.c
3. Is that refactoring really needed to fix alignment issue?

ivankochin: Moved these functions back to `ib_md.c`

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -175,20 +180,20 @@ uct_ib_mlx5_devx_reg_ksm_data_mt(uct_ib_mlx5_md_t *md, void *address,
     ucs_carray_for_each(mr, ksm_data->mrs, ksm_data->mr_num) {
         uct_ib_mlx5_devx_klm_entry_set(&klm, mr - ksm_data->mrs, mr_address,
                                        *mr);
-        mr_address = UCS_PTR_BYTE_OFFSET(mr_address, (*mr)->length);
+        mr_address = UCS_PTR_BYTE_OFFSET(mr_address, chunk_size);
     }
 
-    if ((void*)iova != address) {
+    if (ksm_iova + length > (uintptr_t)mr_address) {
```
yosefe, line 186: why need to add length?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -175,20 +180,20 @@ uct_ib_mlx5_devx_reg_ksm_data_mt(uct_ib_mlx5_md_t *md, void *address,
     ucs_carray_for_each(mr, ksm_data->mrs, ksm_data->mr_num) {
         uct_ib_mlx5_devx_klm_entry_set(&klm, mr - ksm_data->mrs, mr_address,
                                        *mr);
-        mr_address = UCS_PTR_BYTE_OFFSET(mr_address, (*mr)->length);
+        mr_address = UCS_PTR_BYTE_OFFSET(mr_address, chunk_size);
```
yosefe, line 183: why need to change?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -528,11 +533,16 @@ uct_ib_mlx5_devx_reg_mt(uct_ib_mlx5_md_t *md, void *address, size_t length,
...
+    size_t chunk_size    = md->super.config.mt_reg_chunk;
+    size_t atomic_prefix = (uintptr_t)address % md->super.dev.atomic_align;
+    void* prefix_address = (void*)((uintptr_t)address - atomic_prefix);
+    size_t align_padding = ucs_padding((uintptr_t)prefix_address, chunk_size);
     uct_ib_mlx5_devx_ksm_data_t *ksm_data;
     ucs_status_t status;
     int dmabuf_fd;
+    int mr_num;
+
+    length += atomic_prefix;
```
yosefe, line 545: seems it didn't exist in previous code, is that a bug fix?

shamisp:
@ivankochin  - this one of the best patch description I have seen ! thanks for doing this.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Before this patch MT registration worked successfully only for buffers aligned to `mt_reg_chunk` so `atomic_prefix` was always 0.

src/uct/ib/base/ib_md.c
```diff
@@ -362,6 +374,13 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
         return UCS_ERR_UNSUPPORTED;
     }
 
+    /* During the deregistration NULL can be passed as address, in that
+       case address can be obtained from the first registered MR */
+    if (address == NULL) {
```
yosefe, line 379: can we pass the right address also in case of deregistration?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: why can't we compare  iova!=address like before?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -156,8 +157,12 @@ uct_ib_mlx5_devx_reg_ksm_data_mt(uct_ib_mlx5_md_t *md, void *address,
                                  uct_ib_mlx5_devx_ksm_data_t *ksm_data,
                                  struct mlx5dv_devx_obj **mr_p, uint32_t *mkey)
 {
-    void *mr_address = address;
-    size_t list_size = ksm_data->mr_num;
+    size_t chunk_size     = md->super.config.mt_reg_chunk;
+    intptr_t chunk_prefix = (intptr_t)address % chunk_size;
+    void* mr_address      = UCS_PTR_BYTE_OFFSET(address, -chunk_prefix);
+    uint64_t ksm_iova     = iova - chunk_prefix;
+    size_t length         = ksm_data->length + chunk_prefix;
```
yosefe, line 164: do we really have to "duplicate" this logic here in addition to what we have in uct_ib_mlx5_devx_reg_mt?
maybe pass aligned address and iova from uct_ib_mlx5_devx_reg_mt to this function, instead of also doing the padding/alignment here?

ivankochin: Done in accordance to offline discussion

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Reverted.

PR #9450, opened 2023-10-27
UCT/TCP: Order network interface listing to produce stable sys_dev number

tvegas1:
## What
Use a sorted list for network interface probing of `tcp` transport.

## Why ?
We need the `sys_dev` identifier allocation to be consistent on identical nodes. The `ib` transport is already listing interfaces in an ordered way, and `tcp` is the first to create `bdf` to `sys_dev` mapping.

Scripts rely on that value to be comparable across identical nodes (jazz). Also in proto v2, `sys_dev` is part of the key comparison, so this could help in key deduplication.

## How ?
Sort interface listing by name.


brminich (dismissed):

brminich:
wire compat may be affected though

yosefe:
Which scripts rely on sys_dev being consistent? We defined it as process-local value..
Regarding symmetric key - is it possible to not rely on it for rkey compare?

tvegas1:
> Which scripts rely on sys_dev being consistent? We defined it as process-local value.. Regarding symmetric key - is it possible to not rely on it for rkey compare?

Summarizing discussion: Initial report is from scripts that compare `ucx_info` outputs across identical nodes to check sanity. Also, while not an issue at scale due to randomness, a more predictible `sys_dev` value help rkey compare. On proto v2, `sys_dev` is part of `ucp_rkey_config_key_t` so it cannot be excluded easily.

src/uct/tcp/tcp_iface.c
```diff
@@ -870,35 +870,25 @@ ucs_status_t uct_tcp_query_devices(uct_md_h md,
...
-        if (entry == NULL) {
-            if (errno != 0) {
-                ucs_error("readdir(%s) failed: %m", UCT_TCP_IFACE_NETDEV_DIR);
-                ucs_free(devices);
-                status = UCS_ERR_IO_ERROR;
-                goto out_closedir;
-            }
-            break; /* no more items */
-        }
+    for (i = 0; i < n; i++) {
```
yosefe, line 890: maybe
`ucs_carray_for_each(entry, entries, n)`

then, can avoid extra variable "i"

src/uct/tcp/tcp_iface.c
```diff
@@ -948,8 +938,12 @@ ucs_status_t uct_tcp_query_devices(uct_md_h md,
     *devices_p     = devices;
     status         = UCS_OK;
 
-out_closedir:
-    closedir(dir);
+out_release:
+    for (i = 0; i < n; i++) {
```
yosefe, line 942: minor: ucs_carray_for_each

tvegas1: fixed

yosefe (approved):

PR #9453, opened 2023-10-30
UCT/IB: added atomic_mem_types attributes to md_query_v2

roiedanino:
## What
added atomic_mem_types attributes to md_query_v2 and validated memory types selected for atomic operations supporting it.

## Why ?
Device memory (`UCS_MEMORY_TYPE_RDMA`) might support allocation but not atomic operations on certain MOFED firmware versions.

## How ?
Adding atomic_mem_type attribute, so in md_query_v2 we would know which memory types supports atomics

## Issue
When the allocated memory type has no md that supports atomics for that memory type, no proto would be selected for atomic operations and we will fallback to reconfig proto failure

### DIFF
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -567,13 +567,7 @@ enum {
      * received from different peers are compared equal, they can be used
      * interchangeably, avoiding the need to keep all of them in memory.
      */
-    UCP_MEM_MAP_SYMMETRIC_RKEY = UCS_BIT(3),
-
-    /**
-     * (?) Indicating that allocated memory type should support atomic operations
-     * only available when setting UCP_MEM_MAP_ALLOCATE
-    */
-    UCP_MEM_MAP_DEVICE_MEMORY_ATOMICS = UCS_BIT(4),
+    UCP_MEM_MAP_SYMMETRIC_RKEY = UCS_BIT(3)
 };
 
 
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -567,7 +567,13 @@ enum {
      * received from different peers are compared equal, they can be used
      * interchangeably, avoiding the need to keep all of them in memory.
      */
-    UCP_MEM_MAP_SYMMETRIC_RKEY = UCS_BIT(3)
+    UCP_MEM_MAP_SYMMETRIC_RKEY = UCS_BIT(3),
+
+    /**
+     * (?) Indicating that allocated memory type should support atomic operations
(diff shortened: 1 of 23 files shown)


src/ucp/api/ucp.h
```diff
@@ -567,7 +567,13 @@ enum {
...
      * interchangeably, avoiding the need to keep all of them in memory.
      */
-    UCP_MEM_MAP_SYMMETRIC_RKEY = UCS_BIT(3)
+    UCP_MEM_MAP_SYMMETRIC_RKEY = UCS_BIT(3),
+
+    /**
+     * (?) Indicating that allocated memory type should support atomic operations
+     * only available when setting UCP_MEM_MAP_ALLOCATE
+    */
+    UCP_MEM_MAP_DEVICE_MEMORY_ATOMICS = UCS_BIT(4),
```
roiedanino, line 576: delete

src/ucp/proto/proto_common.c
```diff
@@ -507,27 +507,42 @@ static ucp_lane_index_t ucp_proto_common_find_lanes_internal(
...
-                ucs_trace("%s: no support of dst md map 0x%" PRIx64,
-                          lane_desc, rkey_config_key->md_map);
+                ucs_trace("%s: no support of dst md map 0x%" PRIx64, lane_desc,
+                          rkey_config_key->md_map);
                 continue;
             }
 
-            if (!(md_attr->flags & UCT_MD_FLAG_NEED_RKEY) &&
+            if (
+                //!(md_attr->flags & UCT_MD_FLAG_NEED_RKEY) &&
```
roiedanino, line 524: delete

src/ucp/proto/proto_common.c
```diff
@@ -507,27 +507,42 @@ static ucp_lane_index_t ucp_proto_common_find_lanes_internal(
...
             }
-
             if (((md_attr->flags & UCT_MD_FLAG_NEED_RKEY) ||
                  (flags & UCP_PROTO_COMMON_INIT_FLAG_RKEY_PTR)) &&
-                !(rkey_config_key->md_map &
-                  UCS_BIT(ep_config_key->lanes[lane].dst_md_index))) {
+                (!(rkey_config_key->md_map &
+                   UCS_BIT(ep_config_key->lanes[lane].dst_md_index))) &&
+                (!(md_attr->alloc_mem_types &
+                   UCS_BIT(rkey_config_key->mem_type)))) {
```
roiedanino, line 515: check if those changes are necessary

PR #9459, opened 2023-11-01
UCT/IB/MLX5/DV: disable device memory if atomics are not available

roiedanino:
UCT/IB/MLX5/DV: disable device memory if atomics are not available
UCP/PROTO: changed proto_reconfig assert to ucs_error

## What
_Describe what this PR is doing._ 
Disabled device memory if atomics are not available and replaced proto_reconfig assert to `ucs_error` informative log so `UCX` won't crash if some of the protocols are missing.

## Why ?
Fixing the following bug:
when memic is available and atomics are not supported by FW, proto tries to create atomic `rkey` and fails

### DIFF
--- a/src/ucp/proto/proto_reconfig.c
+++ b/src/ucp/proto/proto_reconfig.c
@@ -34,9 +34,22 @@ static ucs_status_t ucp_proto_reconfig_progress(uct_pending_req_t *self)
 {
     ucp_request_t *req = ucs_container_of(self, ucp_request_t, send.uct);
     ucp_ep_h ep        = req->send.ep;
+    UCS_STRING_BUFFER_ONSTACK(strb, 256);
 
     /* This protocol should not be selected for valid and connected endpoint */
-    ucs_assert(!(ep->flags & UCP_EP_FLAG_REMOTE_CONNECTED));
+    if (ep->flags & UCP_EP_FLAG_REMOTE_CONNECTED) {
+        ucp_ep_config_name(ep->worker, req->send.proto_config->ep_cfg_index,
+                           &strb);
+        ucs_string_buffer_appendf(&strb, " | ");
+        ucp_proto_select_info_str(ep->worker,
+                                  req->send.proto_config->rkey_cfg_index,
+                                  &req->send.proto_config->select_param,
+                                  ucp_operation_names, &strb);
+        ucs_error("cannot find remote protocol for: %s",
+                  ucs_string_buffer_cstr(&strb));
+        ucp_request_complete_send(req, UCS_ERR_CANCELED);
+        return UCS_OK;
+    }
 
     if (ep->cfg_index != req->send.proto_config->ep_cfg_index) {
         ucp_trace_req(req,
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -403,11 +403,23 @@ static UCS_F_ALWAYS_INLINE uint32_t uct_ib_mlx5_mkey_index(uint32_t mkey)
     return mkey >> 8;
 }
 
+static UCS_F_ALWAYS_INLINE uct_ib_mr_type_t uct_ib_devx_get_atomic_mr_type(
+        uct_ib_md_t *md, const uct_ib_mlx5_devx_mem_t *memh)
+{
+#if HAVE_IBV_DM
+    /* Device memory only supports default mr */
+    if (memh->dm != NULL) {
+        return UCT_IB_MR_DEFAULT;
+    }
+#endif
+    return uct_ib_md_get_atomic_mr_type(md);
+}
+
 UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_atomic_key,
                         (md, memh), uct_ib_mlx5_md_t *md,
                         uct_ib_mlx5_devx_mem_t *memh)
 {
-    uct_ib_mr_type_t mr_type  = uct_ib_md_get_atomic_mr_type(&md->super);
+    uct_ib_mr_type_t mr_type = uct_ib_devx_get_atomic_mr_type(&md->super, memh);
     uct_ib_mlx5_devx_mr_t *mr = &memh->mrs[mr_type];
     uint8_t mr_id             = uct_ib_md_get_atomic_mr_id(&md->super);
     uint32_t atomic_offset    = uct_ib_md_atomic_offset(mr_id);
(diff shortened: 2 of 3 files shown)


src/ucp/proto/proto_reconfig.c
```diff
@@ -34,9 +34,24 @@ static ucs_status_t ucp_proto_reconfig_progress(uct_pending_req_t *self)
 {
     ucp_request_t *req = ucs_container_of(self, ucp_request_t, send.uct);
     ucp_ep_h ep        = req->send.ep;
+    ucs_string_buffer_t strb;
 
     /* This protocol should not be selected for valid and connected endpoint */
-    ucs_assert(!(ep->flags & UCP_EP_FLAG_REMOTE_CONNECTED));
+    if (ep->flags & UCP_EP_FLAG_REMOTE_CONNECTED) {
+        ucs_string_buffer_init(&strb);
```
yosefe, line 41: to simplify the code, can use UCS_STRING_BUFFER_ONSTACK

src/ucp/proto/proto_reconfig.c
```diff
@@ -34,9 +34,24 @@ static ucs_status_t ucp_proto_reconfig_progress(uct_pending_req_t *self)
...
+        ucp_ep_config_name(ep->worker, req->send.proto_config->ep_cfg_index,
+                           &strb);
+        ucs_string_buffer_appendf(&strb, " | ");
```
yosefe, line 44: can you pls post an example of a full error message?
the "|" character seems weird

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -407,7 +407,10 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_atomic_key,
                         (md, memh), uct_ib_mlx5_md_t *md,
                         uct_ib_mlx5_devx_mem_t *memh)
 {
-    uct_ib_mr_type_t mr_type  = uct_ib_md_get_atomic_mr_type(&md->super);
+    /* dm only supports default mr */
+    uct_ib_mr_type_t mr_type  = memh->dm != NULL ?
+                                         UCT_IB_MR_DEFAULT :
+                                         uct_ib_md_get_atomic_mr_type(&md->super);
```
yosefe, line 413: shall we pass memh->dm or memh to uct_ib_md_get_atomic_mr_type?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1386,15 +1390,20 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
         return;
     }
 
+    devx_memh = ucs_derived_of(memh, uct_ib_mlx5_devx_mem_t);
+    status    = uct_ib_mlx5_devx_reg_atomic_key(md, devx_memh);
+    if (status == UCS_OK) {
+        /* Enable device memory only if atomics are available*/
+        md->super.cap_flags |= UCT_MD_FLAG_ALLOC;
+        uct_ib_mlx5_devx_obj_destroy(devx_memh->atomic_dvmr, "DM-ATOMIC-KSM");
```
yosefe, line 1398: shouldn't uct_ib_mlx5_devx_device_mem_free release atomic_dvmr?
if not, what happens in normal flow when user allocates DM, packs with atomic, and then releases the DM ?

roiedanino: atomic_dvmr is not allocated within `uct_ib_mlx5_devx_device_mem_alloc`, so I thought it should be freed as part of `uct_ib_mlx5_devx_mem_dereg` (.mem_dereg) called from `ucp_mem_unmap`

roiedanino: UCT level API user is calling `uct_md_mkey_pack_v2` if he wants to use atomics, and has to call `uct_md_mem_dereg_v2` accordingly to release the atomic / indirect keys created in `mkey_pack`.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
roiedanino: yes, but it will have to be in a new function because `uct_ib_md_get_atomic_mr_type` is in /ib level and here we want to pass devx memh

src/ucp/proto/proto_reconfig.c
roiedanino: `UCX  ERROR cannot find remote protocol for: 0x3b6f530 self cfg#0 | amo_fetch into host memory from rdma, arg in host memory`

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1330,33 +1342,37 @@ static ucs_status_t
...
     if (ret != 0) {
         ucs_warn("munmap(address=%p, length=%zu) failed: %m", ib_memh->address,
                  length);
     }
 
-    ret = UCS_PROFILE_CALL(ibv_free_dm, ib_memh->dm);
+    status = uct_ib_mlx5_devx_mem_dereg(uct_md, &dereg_params);
+    if (status != UCS_OK) {
+        ucs_warn("%s: uct_ib_mlx5_devx_mem_dereg() failed", ucs_status_string(status));
+    }
```
yosefe, line 1368: though it works, I think it would be nicer to have a common helper function for both uct_ib_mlx5_devx_device_mem_free and uct_ib_mlx5_devx_mem_dereg. This common function will release only what is really relevant for both cases. For example, it will not check UCT_IB_MEM_IMPORTED, memh->umem

test/gtest/uct/test_md.cc
```diff
@@ -436,6 +443,9 @@ UCS_TEST_SKIP_COND_P(test_md, alloc,
             if (mem_type == UCS_MEMORY_TYPE_HOST) {
                 memset(address, 0xBB, size);
             }
+
+            uct_md_mkey_pack_v2(md(), mem.memh, &pack_params,&key);
```
yosefe, line 447: 1. rename params to alloc_params
2. space before &key
3. do we need to pass the buffer to pack to?

roiedanino: I don't think I understand (3), if we don't pass `key` (the key buffer) mkey_pack_v2 will return INVALID_PARAM status
I will add an assertion for the returned status

yosefe: pls ignore (3), i misread the code

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2073,7 +2073,9 @@ uct_ib_mlx5_devx_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
     flags = UCS_PARAM_VALUE(UCT_MD_MKEY_PACK_FIELD, params, flags, FLAGS, 0);
     if (flags & UCT_MD_MKEY_PACK_FLAG_EXPORT) {
 #if HAVE_IBV_DM
-        ucs_assert_always(memh->dm == NULL);
+        if (memh->dm == NULL) {
+            return UCS_ERR_INVALID_PARAM;
```
yosefe, line 2077: pls also print a nice error message

yosefe: please add more details to the message:
`ucs_error("%s: cannot export memory allocated on the device (address %p length %zu)", uct_ib_device_name(...), ...);`

yosefe (approved):

yosefe (approved):

PR #9465, opened 2023-11-03
UCP/PROTO: Calculate perf for ack messages as parallel stages - v1.16

ivankochin:
## What
Backport of #9339


ivankochin:
It was decided not to backport perf fixes to 1.16

PR #9466, opened 2023-11-03
UCP: Fix lane selection for wireup ack message

brminich:
## What
Wireup ack can only be sent with wireup lane If AM lane is not requested

## Why ?
Fixes [3655964](https://redmine.mellanox.com/issues/3655964) (internal link)

Problem popped up after #9445, because with that fix AM is not always selected (if not requested)

### DIFF
--- a/src/ucp/wireup/wireup.c
+++ b/src/ucp/wireup/wireup.c
@@ -78,18 +78,21 @@ const char* ucp_wireup_msg_str(uint8_t msg_type)
 
 static ucp_lane_index_t ucp_wireup_get_msg_lane(ucp_ep_h ep, uint8_t msg_type)
 {
-    ucp_context_h   context           = ep->worker->context;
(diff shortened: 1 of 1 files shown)


yosefe (commented):
can we add gtest for it?

yosefe:
> can we add gtest for it?

Happens when gtest is running on Jazz. Can we repro on CI node?

brminich:
> > can we add gtest for it?
> 
> Happens when gtest is running on Jazz. Can we repro on CI node?

it reproduced on jazz because xpmem is selected for RMA. 

On CI nodes (e.g. swx-rdmz-ucx-new-01) rc_mlx5 is selected for rma (p2p) and then we also select AM lane, because of this [line](https://github.com/openucx/ucx/blob/master/src/ucp/wireup/select.c#L1374)

yosefe (approved):

PR #9467, opened 2023-11-06
UCP/PROTO: Replace ucp protocol in the middle of send operation (offset > 0)

shasson5:
## What
Support replacing ucp protocol in the middle of send operation (offset > 0).

## Why ?
Needed for several features: RC/DC, memic, EP reconfig.

### DIFF
--- a/src/ucp/am/eager_multi.c
+++ b/src/ucp/am/eager_multi.c
@@ -34,7 +34,8 @@ ucp_am_eager_multi_bcopy_proto_init(const ucp_proto_init_params_t *init_params)
         .super.send_op       = UCT_EP_OP_AM_BCOPY,
         .super.memtype_op    = UCT_EP_OP_GET_SHORT,
         .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_CAP_SEG_SIZE |
-                               UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING,
+                               UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING |
+                               UCP_PROTO_COMMON_INIT_FLAG_RESUME,
         .super.exclude_map   = 0,
         .max_lanes           = context->config.ext.max_eager_lanes,
         .initial_reg_md_map  = 0,
--- a/src/ucp/proto/proto_common.c
+++ b/src/ucp/proto/proto_common.c
@@ -715,7 +715,9 @@ void ucp_proto_common_zcopy_adjust_min_frag_always(ucp_request_t *req,
     }
 }
 
-ucs_status_t ucp_proto_request_init(ucp_request_t *req)
+ucs_status_t
+ucp_proto_request_init(ucp_request_t *req,
+                       const ucp_proto_select_param_t *select_param)
 {
     ucp_ep_h ep         = req->send.ep;
     ucp_worker_h worker = ep->worker;
@@ -736,40 +738,32 @@ ucs_status_t ucp_proto_request_init(ucp_request_t *req)
     }
 
     /* Select from protocol hash according to saved request parameters */
-    return ucp_proto_request_lookup_proto(
-            worker, ep, req, proto_select, rkey_cfg_index,
-            &req->send.proto_config->select_param, msg_length);
-}
-
-/**
- * Current implementation of @ref ucp_proto_t::reset supports only the case
- * that no data was sent yet.
- *
- * @param req   request to check
- */
-void ucp_proto_request_check_reset_state(const ucp_request_t *req)
-{
-    ucs_assertv_always(
-            ucp_datatype_iter_is_begin(&req->send.state.dt_iter),
-            "request %p: cannot reset the state after sending %zu bytes", req,
-            req->send.state.dt_iter.offset);
+    return ucp_proto_request_lookup_proto(worker, ep, req, proto_select,
+                                          rkey_cfg_index, select_param,
+                                          msg_length);
 }
 
 void ucp_proto_request_restart(ucp_request_t *req)
 {
+    const ucp_proto_config_t *proto_config = req->send.proto_config;
+    ucp_proto_select_param_t select_param  = proto_config->select_param;
     ucs_status_t status;
 
     ucp_trace_req(req, "proto %s at stage %d restarting",
-                  req->send.proto_config->proto->name, req->send.proto_stage);
+                  proto_config->proto->name, req->send.proto_stage);
 
-    ucp_proto_request_check_reset_state(req);
-    status = req->send.proto_config->proto->reset(req);
+    status = proto_config->proto->reset(req);
     if (status != UCS_OK) {
         ucs_assert_always(status == UCS_ERR_CANCELED);
         return;
     }
 
-    status = ucp_proto_request_init(req);
+    /* Select a protocol with resume request support */
+    if (!ucp_datatype_iter_is_begin(&req->send.state.dt_iter)) {
+        select_param.op_id_flags |= UCP_PROTO_SELECT_OP_FLAG_RESUME;
+    }
+
+    status = ucp_proto_request_init(req, &select_param);
     if (status == UCS_OK) {
         ucp_request_send(req);
     } else {
--- a/src/ucp/proto/proto_common.inl
+++ b/src/ucp/proto/proto_common.inl
@@ -46,6 +46,10 @@ ucp_proto_bcopy_send_func_status(ssize_t packed_size)
 static UCS_F_ALWAYS_INLINE void
 ucp_proto_msg_multi_request_init(ucp_request_t *req)
 {
+    if (!ucp_datatype_iter_is_begin(&req->send.state.dt_iter)) {
+        return;
+    }
+
     req->send.msg_proto.message_id = req->send.ep->worker->am_message_id++;
 }
 
--- a/src/ucp/rma/get_offload.c
+++ b/src/ucp/rma/get_offload.c
@@ -113,6 +113,19 @@ ucp_proto_get_offload_bcopy_init(const ucp_proto_init_params_t *init_params)
                                 init_params->priv_size);
 }
 
+static void ucp_proto_get_offload_reset(ucp_request_t *req)
+{
+    /* get_am protocol will use this field after reset, so it must be
+     * initialized */
+    req->send.state.completed_size = 0;
+}
+
+static ucs_status_t ucp_proto_get_offload_bcopy_reset(ucp_request_t *req)
+{
+    ucp_proto_get_offload_reset(req);
+    return ucp_proto_request_bcopy_reset(req);
+}
+
 ucp_proto_t ucp_get_offload_bcopy_proto = {
     .name     = "get/bcopy",
     .desc     = UCP_PROTO_COPY_OUT_DESC,
@@ -121,7 +134,7 @@ ucp_proto_t ucp_get_offload_bcopy_proto = {
     .query    = ucp_proto_multi_query,
     .progress = {ucp_proto_get_offload_bcopy_progress},
     .abort    = ucp_proto_abort_fatal_not_implemented,
-    .reset    = ucp_proto_request_bcopy_reset
+    .reset    = ucp_proto_get_offload_bcopy_reset
 };
 
 static UCS_F_ALWAYS_INLINE ucs_status_t
@@ -206,6 +219,12 @@ ucp_proto_get_offload_zcopy_init(const ucp_proto_init_params_t *init_params)
                                 init_params->priv_size);
 }
 
+static ucs_status_t ucp_proto_get_offload_zcopy_reset(ucp_request_t *req)
+{
+    ucp_proto_get_offload_reset(req);
+    return ucp_proto_request_zcopy_reset(req);
+}
+
 ucp_proto_t ucp_get_offload_zcopy_proto = {
     .name     = "get/zcopy",
     .desc     = UCP_PROTO_ZCOPY_DESC,
@@ -214,5 +233,5 @@ ucp_proto_t ucp_get_offload_zcopy_proto = {
     .query    = ucp_proto_multi_query,
     .progress = {ucp_proto_get_offload_zcopy_progress},
     .abort    = ucp_proto_abort_fatal_not_implemented,
-    .reset    = ucp_proto_request_zcopy_reset
+    .reset    = ucp_proto_get_offload_zcopy_reset
 };
--- a/src/ucp/rndv/rndv_put.c
+++ b/src/ucp/rndv/rndv_put.c
@@ -19,27 +19,6 @@
 
 #define UCP_PROTO_RNDV_PUT_DESC "write to remote"
 
-
-enum {
-    /* Initial stage for put zcopy is sending the data */
-    UCP_PROTO_RNDV_PUT_ZCOPY_STAGE_SEND = UCP_PROTO_STAGE_START,
-
-    /* Initial stage for put memtype is copy the data to the fragment */
-    UCP_PROTO_RNDV_PUT_MTYPE_STAGE_COPY = UCP_PROTO_STAGE_START,
-
-    /* Flush all lanes to ensure remote delivery */
-    UCP_PROTO_RNDV_PUT_STAGE_FLUSH,
-
-    /* Send ATP without fence (could be done after a flush) */
-    UCP_PROTO_RNDV_PUT_STAGE_ATP,
-
-    /* Send ATP with fence (could be done if using send lanes for ATP) */
-    UCP_PROTO_RNDV_PUT_STAGE_FENCED_ATP,
-
-    /* Memtype only: send the fragment to the remote side */
-    UCP_PROTO_RNDV_PUT_MTYPE_STAGE_SEND
-};
-
 typedef struct ucp_proto_rndv_put_priv {
     uct_completion_callback_t  put_comp_cb;
     uct_completion_callback_t  atp_comp_cb;
@@ -428,6 +407,23 @@ ucp_proto_rndv_put_zcopy_query(const ucp_proto_query_params_t *params,
                       UCP_PROTO_ZCOPY_DESC, put_desc);
 }
 
+static ucs_status_t ucp_proto_rndv_put_zcopy_reset(ucp_request_t *request)
+{
+    const ucp_proto_rndv_put_priv_t *rpriv = request->send.proto_config->priv;
+    ucp_datatype_iter_t *dt_iter           = &request->send.state.dt_iter;
+    ucp_lane_map_t acked_map;
+
+    if (request->send.rndv.put.atp_map != 0) {
+        acked_map       = request->send.rndv.put.atp_map ^ rpriv->atp_map;
+        dt_iter->offset = ucs_popcount(acked_map);
+    } else {
+        dt_iter->offset = dt_iter->length;
+    }
+
+    request->flags &= ~UCP_REQUEST_FLAG_PROTO_INITIALIZED;
+    return UCS_OK;
+}
+
 ucp_proto_t ucp_rndv_put_zcopy_proto = {
     .name     = "rndv/put/zcopy",
     .desc     = NULL,
@@ -441,7 +437,7 @@ ucp_proto_t ucp_rndv_put_zcopy_proto = {
         [UCP_PROTO_RNDV_PUT_STAGE_FENCED_ATP] = ucp_proto_rndv_put_common_fenced_atp_progress,
     },
     .abort    = ucp_proto_request_zcopy_abort,
-    .reset    = (ucp_request_reset_func_t)ucp_proto_reset_fatal_not_implemented
+    .reset    = ucp_proto_rndv_put_zcopy_reset
 };
 
 
--- a/test/gtest/ucp/test_ucp_request.cc
+++ b/test/gtest/ucp/test_ucp_request.cc
@@ -11,6 +11,7 @@ extern "C" {
 #include <ucp/core/ucp_request.inl>
 #include <ucp/core/ucp_worker.h>
 #include <ucp/proto/proto_common.h>
+#include <ucp/rndv/proto_rndv.h>
 }
 
 
@@ -117,7 +118,7 @@ public:
     {
     }
 
-    void init() override
+    virtual void init() override
     {
         if (!m_ucp_config->ctx.proto_enable) {
             UCS_TEST_SKIP_R("reset is not supported for proto v1");
@@ -130,13 +131,15 @@ public:
         ucp_test::init();
         modify_config("TCP_SNDBUF", "8K", IGNORE_IF_NOT_EXIST);
         modify_config("IB_TX_QUEUE_LEN", "65", IGNORE_IF_NOT_EXIST);
+        modify_config("MM_FIFO_SIZE", "64", IGNORE_IF_NOT_EXIST);
         create_entity(true);
+        create_entity();
 
         sender().connect(&receiver(), get_ep_params());
         receiver().connect(&sender(), get_ep_params());
     }
 
-    void cleanup() override
+    virtual void cleanup() override
     {
         m_rkeys.clear();
         m_rbufs.clear();
@@ -248,6 +251,7 @@ public:
         ASSERT_FALSE(UCS_PTR_IS_ERR(request));
         if (request != NULL) {
             wait_for_value(&m_completed, true);
+            ASSERT_TRUE(m_completed);
             ucp_request_release(request);
         }
 
@@ -255,12 +259,8 @@ public:
 
         UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(ep->worker);
         for (auto &req : m_pending) {
-            if (req->send.state.dt_iter.offset > 0) {
-                ucp_request_send(req);
-            } else {
-                ucp_proto_request_restart(req);
-                restart_count++;
-            }
+            ucp_proto_request_restart(req);
+            restart_count++;
         }
         UCP_WORKER_THREAD_CS_EXIT_CONDITIONAL(ep->worker);
 
@@ -284,7 +284,6 @@ public:
             rreq = ucp_tag_recv_nbx(receiver().worker(), rbuf->ptr(),
                                     rbuf->size(), 0, 0, &param);
             ASSERT_FALSE(UCS_PTR_IS_ERR(rreq));
-            reqs.push_back(rreq);
             break;
         case RMA_GET:
             param.op_attr_mask = UCP_OP_ATTR_FLAG_NO_IMM_CMPL;
@@ -310,16 +309,20 @@ public:
 
         ASSERT_FALSE(UCS_PTR_IS_ERR(sreq));
         reqs.push_back(sreq);
+        reqs.push_back(rreq);
     }
 
-    void wait_recv(operation_t op, std::vector<void*> &reqs)
+    void wait_reqs(operation_t op, std::vector<void*> &reqs)
     {
+        size_t reqs_count = reqs.size() / 2;
+
         if (op == STREAM) {
-            for (unsigned i = 0; i < reqs.size(); ++i) {
+            for (unsigned i = 0; i < reqs_count; ++i) {
                 get_stream_data(*m_rbufs[i].get());
             }
         } else if (op == AM) {
-            wait_for_value(&m_am_cb_cnt, reqs.size());
+            wait_for_value(&m_am_cb_cnt, reqs_count);
+            ASSERT_EQ(m_am_cb_cnt, reqs_count);
         }
 
         requests_wait(reqs);
@@ -336,10 +339,44 @@ public:
         }
     }
 
-    void reset_protocol(operation_t op, bool sync = false)
+    virtual void wait_and_restart(const std::vector<void*> &reqs)
     {
-        static const unsigned reqs_count = 1000;
-        static const size_t msg_size     = UCS_KBYTE * 64;
+        wait_any(reqs, [this](const void *ureq) {
+            const ucp_request_t *req = get_request(ureq, true);
+            if (req == NULL) {
+                return false;
+            }
+
+            return is_request_in_the_middle(req);
+        });
+
+        restart(sender().ep());
+    }
+
+    typedef std::function<bool(const void*)> predicate_t;
+
+    void wait_any(const std::vector<void*> &reqs, const predicate_t &predicate)
+    {
+        const double timeout      = 10;
+        const ucs_time_t deadline = ucs::get_deadline(timeout);
+
+        while (ucs_get_time() < deadline) {
+            for (auto &req : reqs) {
+                if (predicate(req)) {
+                    return;
+                }
+            }
+
+            progress();
+        }
+
+        ASSERT_LT(ucs_get_time(), deadline);
+    }
+
+    void reset_protocol(operation_t op, bool sync = false,
+                        unsigned reqs_count = 1000)
+    {
+        static const size_t msg_size = UCS_KBYTE * 70;
 
         for (int i = 0; i < reqs_count; ++i) {
             mapped_buffer *rbuf = new mapped_buffer(msg_size, receiver());
@@ -352,12 +389,12 @@ public:
            data */
         std::vector<void*> reqs;
         send_requests(1, reqs, op, sync);
-        wait_recv(op, reqs);
+        wait_reqs(op, reqs);
 
         /* Send all messages */
         send_requests(reqs_count, reqs, op, sync);
-        restart(sender().ep());
-        wait_recv(op, reqs);
+        wait_and_restart(reqs);
+        wait_reqs(op, reqs);
         flush_ep(sender());
 
         for (int i = 0; i < reqs_count; ++i) {
@@ -387,6 +424,22 @@ public:
         }
     }
 
+    unsigned count_tl_with_caps(uint64_t capability)
+    {
+        /* EPs with no RMA transport use single type protocol */
+        const auto config  = ucp_ep_config(sender().ep());
+        unsigned rma_count = 0;
+
+        for (ucp_lane_index_t lane = 0; lane < config->key.num_lanes; ++lane) {
+            if (ucp_ep_get_iface_attr(sender().ep(), lane)->cap.flags &
+                capability) {
+                rma_count++;
+            }
+        }
+
+        return rma_count;
+    }
+
     static void get_test_variants(std::vector<ucp_test_variant> &variants)
     {
         add_variant_with_value(variants,
@@ -396,6 +449,25 @@ public:
     }
 
 protected:
+    bool is_request_in_the_middle(const ucp_request_t *req)
+    {
+        const ucp_datatype_iter_t *dt_iter = &req->send.state.dt_iter;
+
+        return !ucp_datatype_iter_is_begin(dt_iter) &&
+               !ucp_datatype_iter_is_end(dt_iter);
+    }
+
+    const ucp_request_t *get_request(const void *ureq, bool is_send)
+    {
+        if (ureq == NULL) {
+            return NULL;
+        }
+
+        ucp_request_t *req = (ucp_request_t*)ureq - 1;
+        return (!!(req->flags & UCP_REQUEST_FLAG_PROTO_SEND) == is_send) ? req :
+                                                                           NULL;
+    }
+
     std::vector<std::vector<uint8_t>>           m_sbufs;
     std::vector<std::unique_ptr<mapped_buffer>> m_rbufs;
     std::vector<ucs::handle<ucp_rkey_h>>        m_rkeys;
@@ -410,15 +482,19 @@ UCS_TEST_P(test_proto_reset, tag_eager_multi_bcopy, "ZCOPY_THRESH=inf",
     reset_protocol(TAG);
 }
 
-UCS_TEST_P(test_proto_reset, get_offload_bcopy_to_get_am_bcopy,
-           "ZCOPY_THRESH=inf", "RNDV_THRESH=inf")
+UCS_TEST_P(test_proto_reset, get_offload_bcopy, "ZCOPY_THRESH=inf",
+           "RNDV_THRESH=inf")
 {
+    if (count_tl_with_caps(UCT_IFACE_FLAG_GET_BCOPY) == 0) {
+        UCS_TEST_SKIP_R("no RMA transports found");
+    }
+
     skip_no_pending_rma();
     reset_protocol(RMA_GET);
 }
 
-UCS_TEST_P(test_proto_reset, put_offload_bcopy_to_put_am_bcopy,
-           "ZCOPY_THRESH=inf", "RNDV_THRESH=inf")
+UCS_TEST_P(test_proto_reset, put_offload_bcopy, "ZCOPY_THRESH=inf",
+           "RNDV_THRESH=inf", "RMA_ZCOPY_MAX_SEG_SIZE=1024")
 {
     skip_no_pending_rma();
     reset_protocol(RMA_PUT);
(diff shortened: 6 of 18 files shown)


src/ucp/proto/proto_common.c
```diff
@@ -183,6 +185,14 @@ ucp_proto_common_get_frag_size(const ucp_proto_common_init_params_t *params,
         *max_frag_p = ucs_min(ucp_proto_common_get_seg_size(params, lane),
                               *max_frag_p);
     }
+
+    if (ucs_test_all_flags(params->flags,
+                           UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS |
+                                   UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY) &&
+        (context->config.ext.rma_zcopy_seg_size != UCS_MEMUNITS_AUTO)) {
+        *max_frag_p = ucs_min(context->config.ext.rma_zcopy_seg_size,
```
evgeny-leksikov, line 193: maybe warn that the requested value was adjustment to minimal allowed one?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -100,3 +103,478 @@ UCS_TEST_P(test_ucp_request, test_request_query)
...
+
+        ASSERT_UCS_OK(ucp_ep_rkey_unpack(sender().ep(), rkey_buffer, &m_rkey));
+        ucp_rkey_buffer_release(rkey_buffer);
+    }
+
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_with_value(variants,
+                               UCP_FEATURE_TAG | UCP_FEATURE_RMA |
+                                       UCP_FEATURE_STREAM | UCP_FEATURE_AM,
```
evgeny-leksikov, line 324: alignment

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -100,3 +103,478 @@ UCS_TEST_P(test_ucp_request, test_request_query)
...
+
+    static const size_t mb_to_bytes = 1048576;
+    size_t m_msg_size;
+    ucp_mem_h m_memh;
+    ucp_rkey_h m_rkey;
+    std::vector<char> m_sbuf;
+    std::vector<char> m_rbuf;
+};
+
+#ifdef ENABLE_DEBUG_DATA
```
evgeny-leksikov, line 336: why some tests below don't work without debug data but `reset_protocol(TAG, "rndv/am/bcopy");` does work? can we enable all tests unconditionally?

src/ucp/core/ucp_ep.h
```diff
@@ -885,4 +885,14 @@ ucs_status_t ucp_ep_query_sockaddr(ucp_ep_h ucp_ep, ucp_ep_attr_t *attr);
...
 
+/**
+ * @brief Restarts all pending requests by selecting a new protocol for each
+ *        one and reset old protocol.
+ *
+ * @param [in] ucp_ep           Endpoint object.
+ *
+ * @return Error code as defined by @ref ucs_status_t
+ */
+ucs_status_t ucp_ep_restart_nb(ucp_ep_h ep);
```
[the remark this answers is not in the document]
shasson5: actually I think anyway I need to add a completion callback as a param, so _nb should be ok, no?

src/ucp/am/eager_multi.c
```diff
@@ -34,7 +34,8 @@ ucp_am_eager_multi_bcopy_proto_init(const ucp_proto_init_params_t *init_params)
         .super.send_op       = UCT_EP_OP_AM_BCOPY,
         .super.memtype_op    = UCT_EP_OP_GET_SHORT,
         .super.flags         = UCP_PROTO_COMMON_INIT_FLAG_CAP_SEG_SIZE |
-                               UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING,
+                               UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING |
+                               UCP_PROTO_COMMON_INIT_FLAG_RESUME,
```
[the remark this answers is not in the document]
shasson5: will be addressed later on as discussed

src/ucp/core/ucp_ep.h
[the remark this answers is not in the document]
evgeny-leksikov: It looks confusing to me because conflicting with API functions naming but actually it's just internal function

test/gtest/ucp/test_ucp_request.cc
shasson5: actually it's needed for 2 tests: rndv_get and atp (I moved it to cover only those tests).
Those tests need to access `req->recv.proto_rndv_request` field, which is only present under DEBUG_DATA

src/ucp/core/ucp_ep.c
```diff
@@ -1393,6 +1393,50 @@ static void ucp_ep_failed_destroy(uct_ep_h ep)
...
+    tmp_q = ucs_malloc(sizeof(*tmp_q), "ep_restart");
+    if (tmp_q == NULL) {
+        return UCS_ERR_NO_MEMORY;
+    }
+
+    param.user_data = tmp_q;
+    ucs_queue_head_init(tmp_q);
+    ucp_ep_purge_lanes(ep, ucp_request_purge_enqueue_cb, tmp_q);
+
+    request = ucp_ep_flush_nbx(ep, &param);
```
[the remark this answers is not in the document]
shasson5: the internal function does not allow passing arg to the callback (only inside param).

src/ucp/core/ucp_ep.h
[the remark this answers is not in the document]
shasson5: ok I see, will change

src/ucp/core/ucp_ep.c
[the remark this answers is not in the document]
evgeny-leksikov: `ucp_ep_flush_nbx` is just a wrapper of `ucp_ep_flush_internal` with locking for MT-UCX. It's redundant in case of internal call which is already under the lock.
https://github.com/openucx/ucx/blob/6b27efac95c749291a8ce51d6b887f29bbedef15/src/ucp/rma/flush.c#L450

src/ucp/proto/proto_common.c
shasson5: seems it cause issues with the test. is there any way to disable specific warning in gtest?

test/gtest/ucp/test_ucp_request.cc
evgeny-leksikov: ok, pls add the comment

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -100,3 +103,478 @@ UCS_TEST_P(test_ucp_request, test_request_query)
...
+UCP_INSTANTIATE_TEST_CASE_TLS(test_proto_reset_rndv_get, ib, "ib")
+
+
+class test_proto_reset_atp : public test_proto_reset {
+public:
+    void init()
+    {
+        test_proto_reset::init();
+        static const size_t kb_to_bytes = 1024;
+        m_msg_size                      = 50 * kb_to_bytes;
```
evgeny-leksikov, line 460: UCS_KBYTE

src/ucp/proto/proto_common.c
evgeny-leksikov: 1. why test does wrong thing?
2. yes, see usage of `scoped_log_handler` in error handling tests

shasson5: in the I modify `rma_zcopy_seg_size` value so now it gets a warning

evgeny-leksikov: The question is why do you set `rma_zcopy_seg_size ` to unsupported/wrong value which should be adjusted?
> Did you mean to use ucs_warn?

yes, `ucs_warn` or `ucs_diag`

src/ucp/proto/proto_common.c
```diff
@@ -186,13 +186,20 @@ ucp_proto_common_get_frag_size(const ucp_proto_common_init_params_t *params,
...
         (context->config.ext.rma_zcopy_seg_size != UCS_MEMUNITS_AUTO)) {
         *max_frag_p = ucs_min(context->config.ext.rma_zcopy_seg_size,
                               *max_frag_p);
     }
+
+    if (*max_frag_p < *min_frag_p) {
+        ucs_warn("max tl fragment size (%lu) cannot be smaller than (%lu), "
+                 "using min value supported by transport",
+                 *max_frag_p, *min_frag_p);
+    }
```
evgeny-leksikov, line 202: From my understanding the check should be before as the value has been adjusted and under
```
(context->config.ext.rma_zcopy_seg_size != UCS_MEMUNITS_AUTO)
```
condition, otherwise this is a dead code which never executed

src/ucp/core/ucp_ep.h
```diff
@@ -885,4 +887,26 @@ ucs_status_t ucp_ep_query_sockaddr(ucp_ep_h ucp_ep, ucp_ep_attr_t *attr);
...
+ * @brief Restarts all pending requests by selecting a new protocol for each
+ *        one and reset old protocol.
+ *
+ * @param [in] ep   Endpoint object.
+ * @param [in] cb   Callback to be called upon completion.
+ * @param [in] arg  Argument to be passed to callback.
+ *
+ * @return Error code as defined by @ref ucs_status_t
+ */
+ucs_status_t ucp_ep_pending_schedule_restart(ucp_ep_h ep,
```
evgeny-leksikov, line 908: Where is this function called besides tests? The PR seems incomplete...

src/ucp/proto/proto_common.c
evgeny-leksikov: not addressed

shasson5: you mean before `ucs_min(context->config.ext.rma_zcopy_seg_size`?

src/ucp/core/ucp_ep.h
shasson5: it is used in RC/DC while reconfiguring EP

evgeny-leksikov: Do you mean other PR? pls add a reference

src/ucp/proto/proto_common.c
evgeny-leksikov: yes, after adjustment it does not make sense

src/ucp/proto/proto_common.c
```diff
@@ -183,6 +185,20 @@ ucp_proto_common_get_frag_size(const ucp_proto_common_init_params_t *params,
...
+                           UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS |
+                           UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY) &&
+        (context->config.ext.rma_zcopy_seg_size != UCS_MEMUNITS_AUTO)) {
+        if (*max_frag_p < context->config.ext.rma_zcopy_seg_size) {
+            ucs_warn("max tl fragment size (%lu) cannot be smaller than (%lu)",
+                     *max_frag_p, *min_frag_p);
+        } else {
+            *max_frag_p = ucs_min(context->config.ext.rma_zcopy_seg_size,
+                                     *max_frag_p);
+        }
```
evgeny-leksikov, line 200: pls keep things simpler, for example
```
if (a > max_val) {
    warn("Requested value of \"a\" is greater than maximum allowed value %d, it has been adjusted", max_val);
    a = max_val;
}
```

src/ucp/core/ucp_ep.h
shasson5: it's WIP, will update once I add it.

src/ucp/proto/proto_common.c
```diff
@@ -191,9 +191,11 @@ ucp_proto_common_get_frag_size(const ucp_proto_common_init_params_t *params,
...
         (context->config.ext.rma_zcopy_seg_size != UCS_MEMUNITS_AUTO)) {
-        if (*max_frag_p < context->config.ext.rma_zcopy_seg_size) {
-            ucs_warn("max tl fragment size (%lu) cannot be smaller than (%lu)",
-                     *max_frag_p, *min_frag_p);
+        if (context->config.ext.rma_zcopy_seg_size < *min_frag_p) {
+            ucs_warn("requested fragment size is smaller than minimum allowed "
+                     "(%lu), adjusting it",
+                     *min_frag_p);
+            *max_frag_p = *min_frag_p;
         } else {
```
evgeny-leksikov, line 199: is not this branch redundant now?

shasson5: I don't think so, for example. if `context->config.ext.rma_zcopy_seg_size > max_frag_p` then you would not want to take the value of `context->config.ext.rma_zcopy_seg_size`

evgeny-leksikov: then logic should be the same as with min value

shasson5: you mean:
`*max_frag_p = ucs_min(ucs_max(context->config.ext.rma_zcopy_seg_size, *min_frag_p), *max_frag_p);`?

shasson5:
@evgeny-leksikov PR is ready for review

src/ucp/core/ucp_ep.h
evgeny-leksikov: are cb/arg indeed required to be parameterized?

src/ucp/proto/proto_common.c
evgeny-leksikov: I mean:
```
if (a < min_val) {
    warn(...);
    a = min_val;
} else if (a > max_val) {
    warn(...);
    a = max_val;
}
```

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -100,3 +103,489 @@ UCS_TEST_P(test_ucp_request, test_request_query)
...
+    std::vector<char> m_rbuf;
+    unsigned m_completed;
+    ucp_mem_h m_memh;
+    ucp_rkey_h m_rkey;
+};
+
+UCS_TEST_P(test_proto_reset, eager_zcopy, "ZCOPY_THRESH=0", "RNDV_THRESH=inf")
+{
+    reset_protocol(TAG, "egr/multi/bcopy");
+}
```
evgeny-leksikov, line 349: IMO better to keep test name the same as protocol name, in some cases current naming looks confusing

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -100,3 +103,489 @@ UCS_TEST_P(test_ucp_request, test_request_query)
...
+        /* Backup uct ops for all lanes */
+        for (ucp_lane_index_t lane = 0; lane < num_lanes; ++lane) {
+            m_ops.push_back(ucp_ep_get_lane(ep, lane)->iface->ops);
+        }
+
+        hook_uct_cbs();
+
+        /* Wait for rndv_put initialization */
+        while (std::string(req->send.proto_config->proto->name) !=
+               "rndv/put/zcopy") {
```
evgeny-leksikov, line 544:
```suggestion
        /* Wait for rndv_put initialization */
        const std::string rndv_put_zcopy_name("rndv/put/zcopy");
        while (rndv_put_zcopy_name != req->send.proto_config->proto->name) {
```

src/ucp/rndv/rndv_put.c
```diff
@@ -129,26 +129,33 @@ static size_t ucp_proto_rndv_put_common_pack_atp(void *dest, void *arg)
...
 }
 
+static size_t
+ucp_proto_rndv_put_common_ack_size(ucp_request_t *req, ucp_lane_map_t atp_map)
+{
+    const ucp_proto_rndv_put_priv_t *rpriv = req->send.proto_config->priv;
+
+    /* When we need to send multiple ATP messages: each will acknowledge 1 byte,
+       except the last ATP which will acknowledge the remaining payload size.
+       This is simpler than keeping track of how much was sent on each lane */
```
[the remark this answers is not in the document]
shasson5: AFAIU we need to ack each lane because you need to verify they all finished sending all the data.
anyway this logic was already in the code, I just refactored it out in order to use it in `ucp_proto_rndv_put_zcopy_reset`

test/gtest/ucp/test_ucp_request.cc
shasson5: test name is source protocol and argument name is dest protocol

src/ucp/core/ucp_ep.h
shasson5: actually it seems that A ucp_ep API is not needed (because of the usage in RC/DC).
so will remove it alltogether

evgeny-leksikov: so, not addressed?

src/ucp/proto/proto_common.c
evgeny-leksikov: could you pls describe the logic you are implementing here? It seems like we think differently

src/ucp/rndv/rndv_put.c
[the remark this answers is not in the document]
evgeny-leksikov: ok

src/ucp/rndv/rndv_put.c
```diff
@@ -414,6 +421,22 @@ ucp_proto_rndv_put_zcopy_query(const ucp_proto_query_params_t *params,
                       UCP_PROTO_ZCOPY_DESC, put_desc);
 }
 
+static ucs_status_t ucp_proto_rndv_put_zcopy_reset(ucp_request_t *request)
+{
+    const ucp_proto_rndv_put_priv_t *rpriv = request->send.proto_config->priv;
+    ucp_lane_map_t acked_map               = rpriv->atp_map;
+    size_t acked_size                      = 0;
+
+    while (acked_map != request->send.rndv.put.atp_map) {
```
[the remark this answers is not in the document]
evgeny-leksikov: then maybe something like:
```
acked_map = rpriv->atp_map & ~(rpriv->atp_map & request->send.rndv.put.atp_map);
ucs_for_each_bit(i, acked_map) {
    acked_size += ucp_proto_rndv_put_common_ack_size(request, acked_map);
    acked_map &= ~UCS_BIT(i);
}
```

test/gtest/ucp/test_ucp_request.cc
evgeny-leksikov: how does `eager_zcopy` match `egr/multi/bcopy`?

shasson5: as a first step we decided to support only bcopy protocols after reset

evgeny-leksikov: Let's make naming of tests consistent to what it's testing

shasson5: you mean combine src +dst proto?

evgeny-leksikov: what is src and dst? "sender and receiver" or switch "from to"? it's not clear without digging into implementation

shasson5: how about `eager_zcopy_to_eager_bcopy`?

src/ucp/rndv/rndv_put.c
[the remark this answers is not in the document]
shasson5: will not work because `acked_map` should start from full `atp_map`

src/ucp/proto/proto_common.c
shasson5: I added a new config var (`rma_zcopy_seg_size`) in order to limit seg size in case TL support very large seg size.
The reason for this new config is to simplify testing of reset_request scenraios

src/ucp/am/eager_multi.c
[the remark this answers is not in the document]
shasson5: done

src/ucp/core/ucp_ep.h
shasson5: done

src/ucp/proto/proto_common.c
evgeny-leksikov: I mean the logic in this particular if/else section, how do you adjust the parameter

src/ucp/proto/proto_common.inl
```diff
@@ -46,6 +46,11 @@ ucp_proto_bcopy_send_func_status(ssize_t packed_size)
 static UCS_F_ALWAYS_INLINE void
 ucp_proto_msg_multi_request_init(ucp_request_t *req)
 {
+    if (ucp_proto_select_op_flags(&req->send.proto_config->select_param) &
+        UCP_PROTO_SELECT_OP_FLAG_RESUME) {
+        return;
+    }
```
evgeny-leksikov, line 52: won't it cause ordering issues?

src/ucp/rndv/rndv_put.c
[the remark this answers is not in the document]
evgeny-leksikov: so can change `acked_map` initial value?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -100,3 +103,510 @@ UCS_TEST_P(test_ucp_request, test_request_query)
...
+
+    void wait_receive(operation_e op)
+    {
+        if (op == STREAM) {
+            get_stream_data();
+        } else if (op == AM) {
+            while (m_rbuf != m_sbuf) {
+                progress();
+            }
+        }
```
evgeny-leksikov, line 220: pls add timeout to avoid hanged test

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -100,3 +103,510 @@ UCS_TEST_P(test_ucp_request, test_request_query)
...
+        EXPECT_EQ(m_sbuf, m_rbuf);
+    }
+
+    virtual void
+    wait_and_restart(void *sreq, void *rreq, const std::string &expected_proto)
+    {
+        ucp_request_t *req                 = (ucp_request_t*)sreq - 1;
+        const ucp_datatype_iter_t *dt_iter = &req->send.state.dt_iter;
+
+        while (dt_iter->offset == 0) {
```
evgeny-leksikov, line 324: `wait_for_value`

test/gtest/ucp/test_ucp_request.cc
evgeny-leksikov: how to differentiate single/multi? IMO better to have full exact name

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -100,3 +103,510 @@ UCS_TEST_P(test_ucp_request, test_request_query)
...
+        return UCS_OK;
+    }
+
+    static ucs_status_t zcopy_no_resource(uct_ep_h ep, const uct_iov_t *iov,
+                                          size_t iovcnt, uint64_t remote_addr,
+                                          uct_rkey_t rkey,
+                                          uct_completion_t *comp)
+    {
+        return UCS_ERR_NO_RESOURCE;
+    }
```
evgeny-leksikov, line 542: can reuse `ucs_empty_function_return_<status>`

src/ucp/proto/proto_common.c
shasson5: The logic is intended to force an upper bound to max_frag value. 
If I understand correctly you think it should also effect min_frag value. 
but it is only supposed to effect max_frag.

src/ucp/proto/proto_common.inl
shasson5: AFAIU it will prevent existing requests from getting a new message id.
new requests will not be effected.

src/ucp/rndv/rndv_put.c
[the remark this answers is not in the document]
shasson5: so how can you stop iteration when you reach acked_map == atp_map?

test/gtest/ucp/test_ucp_request.cc
shasson5: I don't wait for a specific value, I wait for any value other then zero

src/ucp/proto/proto_common.c
evgeny-leksikov: then the code should be simplified, pls make it straight forward according what it's doing

src/ucp/proto/proto_common.inl
evgeny-leksikov: got it, maybe add an assert to make sure that `req->send.msg_proto.message_id` is initialized?

src/ucp/rndv/rndv_put.c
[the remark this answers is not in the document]
evgeny-leksikov: AFAIU it's possible to calculate all required bits in advance and do the loop thru all of them

test/gtest/ucp/test_ucp_request.cc
evgeny-leksikov: ok, so maybe add something like `wait_for_changed_value(addr, value)`? anyway need to add timeout to all loops like this to avoid hanged test in CI

src/ucp/rndv/rndv_put.c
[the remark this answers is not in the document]
yosefe: we can compare request->send.rndv.put.atp_map to 0
if they are different, it means we did not ack all lanes yet, so ack_size is popcount(rpriv->atp_map) - 1
otherwise, (atp_map is 0) we acked all lanes, so offset is dt_iter length

src/ucp/proto/proto_common.c
```diff
@@ -183,6 +185,25 @@ ucp_proto_common_get_frag_size(const ucp_proto_common_init_params_t *params,
         *max_frag_p = ucs_min(ucp_proto_common_get_seg_size(params, lane),
                               *max_frag_p);
     }
+
+    /* Truncate maximum fragment size according to user configuration. */
+    if (ucs_test_all_flags(params->flags,
+                           UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS |
+                           UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY) &&
+        (context->config.ext.rma_zcopy_seg_size != UCS_MEMUNITS_AUTO)) {
+        if (context->config.ext.rma_zcopy_seg_size < *min_frag_p) {
```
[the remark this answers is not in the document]
yosefe: we can change the var name to RMA_ZCOPY_MAX_SEG_SIZE, and define that it forces an upper bound on the seg size
if max_frag becomes lower than min_frag, the protocol should be disqualified (may already happen today)

*max_frag_p = ucs_min(*max_frag_p, context->config.ext.rma_zcopy_seg_size);

src/ucp/proto/proto_common.inl
yosefe: check ucp_datatype_iter_is_begin

src/ucp/proto/proto_common.c
```diff
@@ -744,34 +765,27 @@ ucs_status_t ucp_proto_request_init(ucp_request_t *req)
...
 {
     ucs_status_t status;
+    ucp_proto_config_t *proto_config;
 
     ucp_trace_req(req, "proto %s at stage %d restarting",
                   req->send.proto_config->proto->name, req->send.proto_stage);
 
-    ucp_proto_request_check_reset_state(req);
-    status = req->send.proto_config->proto->reset(req);
+    proto_config = (ucp_proto_config_t*)req->send.proto_config;
```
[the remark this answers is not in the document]
shasson5: casting needed to convert from const to non-const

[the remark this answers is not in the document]
yosefe: can we make proto_config const?

src/ucp/rma/get_offload.c
```diff
@@ -113,6 +113,14 @@ ucp_proto_get_offload_bcopy_init(const ucp_proto_init_params_t *init_params)
                                 init_params->priv_size);
 }
 
+static ucs_status_t ucp_proto_get_offload_bcopy_reset(ucp_request_t *req)
+{
+    /* get_am protocol will use this field after reset, so it must be
+     * initialized */
+    req->send.state.completed_size = 0;
```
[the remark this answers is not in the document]
shasson5: yes

src/ucp/proto/proto_common.c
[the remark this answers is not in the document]
shasson5: we need to modify op_id_flags in line 785

src/ucp/rma/get_offload.c
```diff
@@ -206,6 +214,14 @@ ucp_proto_get_offload_zcopy_init(const ucp_proto_init_params_t *init_params)
                                 init_params->priv_size);
 }
 
+static ucs_status_t ucp_proto_get_offload_zcopy_reset(ucp_request_t *req)
+{
+    /* get_am protocol will use this field after reset, so it must be
+     * initialized */
+    req->send.state.completed_size = 0;
```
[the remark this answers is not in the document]
shasson5: RESUME flag in this case is in get_am which is the proto after reset

src/ucp/proto/proto_common.c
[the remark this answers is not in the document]
yosefe: we should not modify existing proto_config, the intention is to really make it const

src/ucp/rma/get_offload.c
```diff
@@ -113,12 +113,17 @@ ucp_proto_get_offload_bcopy_init(const ucp_proto_init_params_t *init_params)
...
     /* get_am protocol will use this field after reset, so it must be
      * initialized */
     req->send.state.completed_size = 0;
-    return ucp_proto_request_bcopy_reset(req);
+    return ucp_proto_request_zcopy_reset(req);
+}
+
+static ucs_status_t ucp_proto_get_offload_bcopy_reset(ucp_request_t *req)
+{
+    return ucp_proto_get_offload_reset(req);
```
yosefe, line 126: now this function calls ucp_proto_request_zcopy_reset, seems wrong

yosefe (commented):
unresolved:
- remove tests
- remove seg size
- simplify rndv

shasson5:
the tests here are the full tests, if I remove them, they will be lost.
seg_size will be rebased after we merge its PR.
not sure what means to simplify rndv.

yosefe (commented):
The code of main reset looks good 👍 , waiting for rebase on seg_size and tests

yosefe:
CI failures seem relevant https://github.com/openucx/ucx/pull/9467

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -558,13 +559,16 @@ class test_proto_reset_rndv_get : public test_proto_reset {
...
                 (pair.rreq->recv.proto_rndv_request == NULL)) {
                 return false;
             }
 
             const ucp_request_t *rndv_req = pair.rreq->recv.proto_rndv_request;
-            return rndv_req->send.state.dt_iter.offset > 0;
+
+            return (rndv_req->send.state.dt_iter.offset > 0) &&
+                   (rndv_req->send.state.dt_iter.offset <
+                    rndv_req->send.state.dt_iter.length);
```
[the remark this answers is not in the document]
shasson5: AFAIU completed requests will enter this code with offest == length.
so I would like to skip those requests and iterate until I get to a request with offset < length

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -387,6 +459,12 @@ class test_proto_reset : public ucp_test {
         }
     }
 
+    bool no_rma_transport()
+    {
+        return has_transport("ud_v") || has_transport("ud_x") ||
+               has_transport("tcp");
```
evgeny-leksikov, line 465:
```suggestion
        return has_any_transport({ "ud_v", "ud_x", "tcp" });
```

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +527,209 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
            "RNDV_THRESH=inf")
 {
     reset_protocol(AM);
 }
 
-UCS_TEST_P(test_proto_reset, rndv_put, "RNDV_THRESH=0", "RNDV_SCHEME=put_zcopy")
+UCS_TEST_P(test_proto_reset, rndv_put, "RNDV_THRESH=0", "RNDV_SCHEME=put_zcopy",
+           "RMA_ZCOPY_MAX_SEG_SIZE=1024")
 {
+    skip_no_pending_rma();
```
evgeny-leksikov, line 572: `UCS_TEST_SKIP_COND_P`?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +527,209 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+
+        restart(receiver().ep());
+    }
+};
+
+UCS_TEST_SKIP_COND_P(test_proto_reset_rndv_get, rndv_get, no_rma_transport(),
+                     "RNDV_THRESH=0", "RNDV_SCHEME=get_zcopy",
+                     "RMA_ZCOPY_MAX_SEG_SIZE=1024")
+{
+    skip_no_pending_rma();
```
evgeny-leksikov, line 605: can this condition be added to `UCS_TEST_SKIP_COND_P`?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +527,209 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+public:
+    void init() override
+    {
+        test_proto_reset::init();
+    }
+
+    void cleanup() override
+    {
+        test_proto_reset::cleanup();
+    }
```
evgeny-leksikov, line 621: why needed?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +527,209 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+            m_ops.push_back(ucp_ep_get_lane(ep, lane)->iface->ops);
+        }
+
+        hook_uct_cbs();
+
+        /* Wait for rndv_put initialization */
+        auto &pair = wait_for_condition(pairs, [](const request_pair_t &pair) {
+            std::string rndv_put_zcopy_name("rndv/put/zcopy");
+            return rndv_put_zcopy_name ==
+                   pair.sreq->send.proto_config->proto->name;
```
evgeny-leksikov, line 683:
```suggestion
bool res = req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED'
if (res) {
    ASSERT_EQ("rndv/put/zcopy", pair.sreq->send.proto_config->proto->name);
}
return res;
```

test/gtest/ucp/test_ucp_request.cc
shasson5: need to specify the reason for skip (`UCS_TEST_SKIP_R("RMA transport does not support pending queue");`)

test/gtest/ucp/test_ucp_request.cc
shasson5: same as below comment

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -309,37 +336,82 @@ class test_proto_reset : public ucp_test {
...
+        auto &pair = wait_for_condition(pairs, [](const request_pair_t &pair) {
+            if (pair.sreq == request_pair_t::from_user(NULL)) {
+                return false;
+            }
+
+            ucp_datatype_iter_t *dt_iter = &pair.sreq->send.state.dt_iter;
+            return (dt_iter->offset > 0) && (dt_iter->offset < dt_iter->length);
+        });
+
+        ASSERT_FALSE(invalid_pair == pair);
```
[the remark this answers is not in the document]
evgeny-leksikov: IMHO it should not be a problem:
```
bool operator!=(pair1, pair2) const
{
    return !(pair1 == pair2);
}
```

test/gtest/ucp/test_ucp_request.cc
shasson5: seems like entities are not yet initialized in that stage.

evgeny-leksikov (approved):

shasson5:
@yosefe can squash?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -113,6 +113,35 @@ class test_proto_reset : public ucp_test {
...
+        }
+
+        bool operator!=(const request_pair_t &other) const
+        {
+            return (sreq != other.sreq) || (rreq != other.rreq);
+        }
+
+        ucp_request_t *sreq;
+        ucp_request_t *rreq;
+    };
```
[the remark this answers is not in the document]
shasson5: you mean remove `request_pair_t `? or just to_user/from_user?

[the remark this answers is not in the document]
yosefe: remove request pair and keep requests in a "flat" vector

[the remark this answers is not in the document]
shasson5: but I need to associate sreq -> rreq.
IMO it's less convenient to do that with a flat vector

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -309,37 +336,82 @@ class test_proto_reset : public ucp_test {
...
 
         for (int i = 0; i < reqs_count; ++i) {
-            send_nb(m_sbufs[i], m_rbufs[i].get(), op, sync, reqs);
+            send_nb(m_sbufs[i], m_rbufs[i].get(), op, sync, pairs);
+        }
+    }
+
+    virtual void wait_and_restart(const std::vector<request_pair_t> &pairs)
+    {
+        auto &pair = wait_for_condition(pairs, [](const request_pair_t &pair) {
```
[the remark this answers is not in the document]
yosefe: could be either of them

test/gtest/ucp/test_ucp_request.cc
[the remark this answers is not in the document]
yosefe: why do we need such association? IMO we can reset the ep and then expect all reqs complete successfully.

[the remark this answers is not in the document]
shasson5: we need association for RNDV ATP test, because we wait on both sender and receiver sides:
test_proto_reset_atp::wait_and_restart

test/gtest/ucp/test_ucp_request.cc
yosefe: maybe we can check UCT capabilities in worker->wiface[i], rather than by transport names?

test/gtest/ucp/test_ucp_request.cc
shasson5: not sure how is this the same?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +526,198 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+        }
+
+        m_req = pair.sreq;
+
+        hook_uct_cbs();
+        restart(ep);
+        restore_uct_cbs();
+    }
+
+    static ucp_request_t *m_req;
```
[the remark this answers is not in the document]
shasson5: it's already protected

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +526,198 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+    std::vector<uct_iface_ops> m_ops;
+};
+
+ucp_request_t *test_proto_reset_atp::m_req;
+
+UCS_TEST_P(test_proto_reset_atp, rndv_put, "RNDV_THRESH=0",
+           "RNDV_SCHEME=put_zcopy", "RMA_ZCOPY_MAX_SEG_SIZE=1024")
+{
+    if (count_resources(sender(), "rc_mlx5") <= 1) {
```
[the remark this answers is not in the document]
shasson5: AFAIR it is intended  to enable multiple ATP messages.

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -387,6 +459,11 @@ class test_proto_reset : public ucp_test {
         }
     }
 
+    bool no_rma_transport()
+    {
+        return has_any_transport({"ud_v", "ud_x", "tcp"});
```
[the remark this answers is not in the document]
shasson5: > What if we have both rc and ud? is it still considered as no RMA transport? 
No, we have ib_shm, but it returns false from this function.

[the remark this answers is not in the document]
shasson5: > Also, why need to skip transports without RMA?

Because they use get_am which is a single type proto.

test/gtest/ucp/test_ucp_request.cc
shasson5: you mean regardless to the location of the call?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +526,198 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+        });
+
+        restore_uct_cbs();
+
+        /* Wait until ATP stage starts */
+        static const unsigned send_stage = 0;
+        const double timeout             = 10;
+        ucs_time_t deadline              = ucs::get_deadline(timeout);
+
+        while ((pair.sreq->send.proto_stage == send_stage) &&
```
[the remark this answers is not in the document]
shasson5: existing progress will send all ATPs instead of just the first one.

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +526,198 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+        }
+
+        /* One more progress to send the first ATP */
+        pair.sreq->send.uct.func(&pair.sreq->send.uct);
+
+        const ucp_request_t *rndv_req = pair.rreq->recv.proto_rndv_request;
+        deadline                      = ucs::get_deadline(timeout);
+
+        /* Wait until receiver gets the ATP message */
+        while (rndv_req->send.state.completed_size == 0 &&
```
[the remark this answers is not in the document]
shasson5: same as above

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +526,198 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+                     "RNDV_THRESH=0", "RNDV_SCHEME=get_zcopy",
+                     "RMA_ZCOPY_MAX_SEG_SIZE=1024")
+{
+    skip_no_pending_rma();
+    reset_protocol(TAG);
+}
+
+UCP_INSTANTIATE_TEST_CASE(test_proto_reset_rndv_get)
+
+class test_proto_reset_atp : public test_proto_reset {
```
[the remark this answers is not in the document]
shasson5: it's documented inside the function.

shasson5:
Comments fixed @yosefe

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -459,10 +429,20 @@ class test_proto_reset : public ucp_test {
         }
     }
 
-    bool no_rma_transport()
+    unsigned count_rma_transports(uint64_t capability)
```
yosefe, line 432: maybe rename to count_tl_with_caps since the function itself is not really specific to RMA caps

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -459,10 +429,20 @@ class test_proto_reset : public ucp_test {
...
     {
-        return has_transport("ud_v") || has_transport("ud_x") ||
-               has_transport("tcp");
+        /* EPs with no RMA transport use single type protocol */
+        const auto config  = ucp_ep_config(sender().ep());
+        unsigned rma_count = 0;
+
+        for (ucp_lane_index_t lane = 0; lane < config->key.num_lanes; ++lane) {
+            uct_iface_attr_t *attr = ucp_ep_get_iface_attr(sender().ep(), lane);
```
yosefe, line 439: minor: no need the helper var attr

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -580,46 +566,54 @@ UCP_INSTANTIATE_TEST_CASE(test_proto_reset)
...
-            return (rndv_req->send.state.dt_iter.offset > 0) &&
-                   (rndv_req->send.state.dt_iter.offset <
-                    rndv_req->send.state.dt_iter.length);
+            if (rndv_req == NULL) {
+                return false;
+            }
+
+            const ucp_datatype_iter_t *dt_iter = &rndv_req->send.state.dt_iter;
+            return !ucp_datatype_iter_is_begin(dt_iter) &&
+                   !ucp_datatype_iter_is_end(dt_iter);
```
yosefe, line 589: can we share some common code with line 341-355?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -580,46 +566,54 @@ UCP_INSTANTIATE_TEST_CASE(test_proto_reset)
...
+        UCS_TEST_SKIP_R("no RMA transports found");
+    }
+
     skip_no_pending_rma();
     reset_protocol(TAG);
 }
 
 UCP_INSTANTIATE_TEST_CASE(test_proto_reset_rndv_get)
 
+/* This test is intended to check resseting request during ATP phase for
```
yosefe, line 609: typo: resetting

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -644,87 +638,101 @@ class test_proto_reset_atp : public test_proto_reset {
...
+    }
+
+    static ucs_status_t uct_am_short(uct_ep_h ep, uint8_t id, uint64_t header,
+                                     const void *payload, unsigned length)
+    {
+        if (is_second_atp(id)) {
+            return UCS_ERR_NO_RESOURCE;
+        }
+
+        return m_ops[ep].ep_am_short(ep, id, header, payload, length);
```
yosefe, line 661: - Add a member var to count ATPs (so test works when running multiple variants/iterations)
- Remove is_second_atp to simplify:
```c
/* Pause on 2nd ATP message */
if ((id == UCP_AM_ID_RNDV_ATP) && (++m_atp_count == 2)) {
     return UCS_ERR_NO_RESOURCE;
}

return m_ops[ep].ep_am_short(ep, id, header, payload, length);
```

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -644,87 +638,101 @@ class test_proto_reset_atp : public test_proto_reset {
...
-ucp_request_t *test_proto_reset_atp::m_req;
+std::vector<uct_pending_req_t*> test_proto_reset_atp::m_reqs;
+std::map<uct_ep_h, uct_iface_ops> test_proto_reset_atp::m_ops;
 
 UCS_TEST_P(test_proto_reset_atp, rndv_put, "RNDV_THRESH=0",
            "RNDV_SCHEME=put_zcopy", "RMA_ZCOPY_MAX_SEG_SIZE=1024")
 {
-    if (count_resources(sender(), "rc_mlx5") <= 1) {
-        UCS_TEST_SKIP_R("Less than 2 RC resources are found");
+    if (count_rma_transports(UCT_IFACE_FLAG_PUT_ZCOPY) <= 1) {
```
yosefe, line 734: minor: < 2

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -644,87 +638,101 @@ class test_proto_reset_atp : public test_proto_reset {
...
-        /* Wait for rndv_put initialization */
-        auto &pair = wait_for_condition(pairs, [](const request_pair_t &pair) {
-            std::string rndv_put_zcopy_name("rndv/put/zcopy");
-            return rndv_put_zcopy_name ==
-                   pair.sreq->send.proto_config->proto->name;
-        });
-
-        restore_uct_cbs();
+        /* Wait until first ATP was sent and received by remote side */
+        static const unsigned send_stage = 3;
```
yosefe, line 687: can we use a UCP constant instead of 3?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -644,87 +638,101 @@ class test_proto_reset_atp : public test_proto_reset {
...
+            ucp_request_t *req = (ucp_request_t*)ureq - 1;
+            if (!(req->flags & UCP_REQUEST_FLAG_PROTO_SEND)) {
+                return false;
+            }
 
-        while ((pair.sreq->send.proto_stage == send_stage) &&
-               (ucs_get_time() < deadline)) {
-            pair.sreq->send.uct.func(&pair.sreq->send.uct);
-        }
+            if (req->send.proto_stage != send_stage) {
```
yosefe, line 698: unite with line 694 using ||

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -644,87 +638,101 @@ class test_proto_reset_atp : public test_proto_reset {
...
 
-        /* Wait until receiver gets the ATP message */
-        while (rndv_req->send.state.completed_size == 0 &&
-               (ucs_get_time() < deadline)) {
-            receiver().progress();
-        }
+            ucp_request_t *rreq           = (ucp_request_t*)user_rreq - 1;
+            const ucp_request_t *rndv_req = rreq->recv.proto_rndv_request;
+            if (rndv_req == NULL) {
+                return false;
```
yosefe, line 714: why need to find the matching receive request?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -644,87 +638,101 @@ class test_proto_reset_atp : public test_proto_reset {
...
 
-        hook_uct_cbs();
-        restart(ep);
+        restart(sender().ep());
         restore_uct_cbs();
     }
 
-    static ucp_request_t *m_req;
-    std::vector<uct_iface_ops> m_ops;
+    static std::vector<uct_pending_req_t*> m_reqs;
```
yosefe, line 724: m_pending_reqs ?

test/gtest/ucp/test_ucp_request.cc
shasson5: In 341 I test send request and in 583 recv request. which means:
1) check PROTO_SEND exist for send request, but not exist for recv request.
2) data_iter for send request is done on the request itself, in contrast with recv request which check internal RNDV request.

test/gtest/ucp/test_ucp_request.cc
shasson5: it is only defined in rndv_put.c

test/gtest/ucp/test_ucp_request.cc
shasson5: it's unsafe to unite them because if the request is recv request then `req->send.proto_stage` may conain garbage

test/gtest/ucp/test_ucp_request.cc
shasson5: how else can wait for receiver to process it?

test/gtest/ucp/test_ucp_request.cc
yosefe: some of the code looks similar so could use a helper function

test/gtest/ucp/test_ucp_request.cc
yosefe: so need to move it to h file, otherwise it can change in UCP and break the test, the number "3" looks fragile

test/gtest/ucp/test_ucp_request.cc
yosefe: C defines a short circuit evaluation

test/gtest/ucp/test_ucp_request.cc
yosefe: i don't think we must wait for it here. can do it before the test finishes. as long as we call progress, the recv reqs will also progress

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +524,207 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+                    UCP_PROTO_RNDV_PUT_STAGE_FENCED_ATP);
+        });
+
+        restart(sender().ep());
+        restore_uct_cbs();
+    }
+
+    static unsigned m_atp_count;
+    static std::vector<uct_pending_req_t*> m_pending_reqs;
+    static std::map<uct_ep_h, uct_iface_ops> m_ops;
```
yosefe, line 711: can it be non-static member vars?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +524,207 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+};
+
+std::vector<uct_pending_req_t*> test_proto_reset_atp::m_pending_reqs;
+std::map<uct_ep_h, uct_iface_ops> test_proto_reset_atp::m_ops;
+unsigned test_proto_reset_atp::m_atp_count;
```
yosefe, line 716: why needed?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -448,47 +524,207 @@ UCS_TEST_P(test_proto_reset, am_eager_multi_bcopy, "ZCOPY_THRESH=inf",
...
+        m_pending_reqs.clear();
+    }
+
+protected:
+    void wait_and_restart(const std::vector<void*> &reqs) override
+    {
+        hook_uct_cbs();
+
+        /* Wait until first ATP was sent */
+        wait_any(reqs, [reqs](const void *ureq) {
```
yosefe, line 694: why need to capture "reqs"?

test/gtest/ucp/test_ucp_request.cc
shasson5: need to access it inside hooks.

yosefe: so need to clear pending reqs and m_ops after every test (in cleanup function)

test/gtest/ucp/test_ucp_request.cc
shasson5: compiler error otherwise:
`undefined reference to `test_proto_reset_atp::m_atp_count`

test/gtest/ucp/test_ucp_request.cc
yosefe: 1. better to clear at exit to release memory after the test
2. I didn't find the place where m_ops is cleared

test/gtest/ucp/test_ucp_request.cc
yosefe: ok

shasson5:
@yosefe comments fixed

yosefe (approved):

yosefe:
@shasson5 pls squash

src/ucp/proto/proto_common.c
```diff
@@ -193,6 +193,12 @@ ucp_proto_common_get_frag_size(const ucp_proto_common_init_params_t *params,
         *max_frag_p = ucs_min(*max_frag_p,
                               context->config.ext.rma_zcopy_max_seg_size);
     }
+
+    if (*max_frag_p < *min_frag_p) {
+        ucs_warn("max tl fragment size (%lu) cannot be smaller than (%lu), "
+                 "using min value supported by transport",
+                 *max_frag_p, *min_frag_p);
+    }
```
yosefe, line 201: seems these lines were added after force push, why is it needed?

PR #9469, opened 2023-11-07
UCP/CONTEXT: If context name is not provided, use counter instead of pointer

yosefe:


src/ucp/core/ucp_context.c
```diff
@@ -1831,7 +1834,8 @@ static void ucp_apply_params(ucp_context_h context, const ucp_params_t *params,
         ucs_snprintf_zero(context->name, UCP_ENTITY_NAME_MAX, "%s",
                           params->name);
     } else {
-        ucs_snprintf_zero(context->name, UCP_ENTITY_NAME_MAX, "%p", context);
+        ucs_snprintf_zero(context->name, UCP_ENTITY_NAME_MAX, "ucp_context_%lu",
+                          ucs_atomic_fadd64(&context_counter, 1));
```
brminich, line 1838: why is atomic needed here?

yosefe: in case multiple threads call ucp_init at the same time. context_counter is a global variable

brminich (approved):

PR #9471, opened 2023-11-07
UCP: Minor code cleanup

yosefe:


src/ucp/core/ucp_worker.h
```diff
@@ -4,10 +4,10 @@
  * See file LICENSE for terms.
  */
 
-
 #ifndef UCP_WORKER_H_
 #define UCP_WORKER_H_
 
+#include "ucp_types.h"
```
brminich, line 10: does it fix something?

yosefe: removed

brminich (approved):

PR #9475, opened 2023-11-07
UCT/TCP: Filtered out bridge devices.

rakhmets:
## What
Filtered out bridge devices.
Added control `UCX_TCP_BRIDGE_DEVICES`. The default value is set to `no`. It means that the function querying the list of available TCP devices doesn't include bridge devices to the list.

## Why ?
The lane associated with the bridge device may be selected for transferring data. However, the lane is unreachable.

### DIFF
--- a/src/uct/tcp/tcp.h
+++ b/src/uct/tcp/tcp.h
@@ -465,6 +465,7 @@ typedef struct uct_tcp_md {
     struct {
         int         af_prio_count;
         sa_family_t af_prio_list[2];
+        uint8_t     bridge_enable;
     } config;
 } uct_tcp_md_t;
 
@@ -475,6 +476,7 @@ typedef struct uct_tcp_md {
 typedef struct uct_tcp_md_config {
     uct_md_config_t                   super;
     UCS_CONFIG_STRING_ARRAY_FIELD(af) af_prio;
+    int                               bridge_enable;
 } uct_tcp_md_config_t;
 
 
--- a/src/uct/tcp/tcp_iface.c
+++ b/src/uct/tcp/tcp_iface.c
@@ -862,6 +862,17 @@ static UCS_CLASS_DEFINE_NEW_FUNC(uct_tcp_iface_t, uct_iface_t, uct_md_h,
                                  uct_worker_h, const uct_iface_params_t*,
                                  const uct_iface_config_t*);
 
+static int uct_tcp_is_bridge(const char *if_name)
+{
+    char path[PATH_MAX];
+    struct stat st;
+
+    ucs_snprintf_safe(path, PATH_MAX, UCT_TCP_IFACE_NETDEV_DIR "/%s/bridge",
+                      if_name);
+
+    return (stat(path, &st) == 0) && S_ISDIR(st.st_mode);
+}
+
 ucs_status_t uct_tcp_query_devices(uct_md_h md,
                                    uct_tl_device_resource_t **devices_p,
                                    unsigned *num_devices_p)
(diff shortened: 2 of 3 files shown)


src/uct/tcp/tcp_md.c
```diff
@@ -21,6 +21,9 @@ static ucs_config_field_t uct_tcp_md_config_table[] = {
      "Priority of address families used for socket connections",
      ucs_offsetof(uct_tcp_md_config_t, af_prio), UCS_CONFIG_TYPE_STRING_ARRAY},
 
+    {"BRIDGE_DEVICES", "n", "Use bridge devices",
+    ucs_offsetof(uct_tcp_md_config_t, bridge_devices), UCS_CONFIG_TYPE_BOOL},
```
brminich: alignment

brminich (approved):

tvegas1 (approved):

src/uct/tcp/tcp_iface.c
```diff
@@ -877,7 +894,9 @@ ucs_status_t uct_tcp_query_devices(uct_md_h md,
     char path_buffer[PATH_MAX];
     ucs_sys_device_t sys_dev;
 
-    n = scandir(UCT_TCP_IFACE_NETDEV_DIR, &entries, NULL, alphasort);
+    n = scandir(UCT_TCP_IFACE_NETDEV_DIR, &entries,
+                tcp_md->config.bridge_devices ? NULL : uct_tcp_filter_bridges,
```
yosefe, line 898: it's nice to pass filter function to scandir - but there are other checks below: line 915 and 920-930 which are not in the filter function. So either all should be in the filter function, or none of them

src/uct/tcp/tcp_md.c
```diff
@@ -14,13 +14,16 @@
...
-    {"", "", NULL,
-     ucs_offsetof(uct_tcp_md_config_t, super), UCS_CONFIG_TYPE_TABLE(uct_md_config_table)},
+    {"", "", NULL, ucs_offsetof(uct_tcp_md_config_t, super),
+     UCS_CONFIG_TYPE_TABLE(uct_md_config_table)},
 
     {"AF_PRIO", "inet,inet6",
      "Priority of address families used for socket connections",
      ucs_offsetof(uct_tcp_md_config_t, af_prio), UCS_CONFIG_TYPE_STRING_ARRAY},
 
+    {"BRIDGE_DEVICES", "n", "Use bridge devices",
```
yosefe, line 24: "BRIDGE_ENABLE", "n", "Enable using bridge devices"

src/uct/tcp/tcp.h
```diff
@@ -465,6 +465,7 @@ typedef struct uct_tcp_md {
     struct {
         int         af_prio_count;
         sa_family_t af_prio_list[2];
+        int         bridge_devices;
```
yosefe, line 468: uint8_t

rakhmets: fixed

src/uct/tcp/tcp_md.c
rakhmets: fixed

src/uct/tcp/tcp_iface.c
rakhmets: Removed filter function.
It is better from code scalability perspective. Because it is not possible to add filter by condition without code duplication.

tvegas1 (approved):

yosefe (approved):
pls squash

PR #9477, opened 2023-11-08
UCS/TOPO: Report NUMA node distance for sys root only

ivankochin:
## What
Detect same NUMA distance when devices has common sys root part only.

## Why ?
When devices have distance distance closer then sys_root and pci (e.g. PXB), NUMA distance is reported, when the right thing is to report default distance.

### DIFF
--- a/src/ucs/sys/topo/base/topo.c
+++ b/src/ucs/sys/topo/base/topo.c
@@ -410,14 +410,17 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
     if (ucs_topo_is_pci_root(common_path)) {
         ucs_topo_pci_root_distance(path1, path2, distance);
         return UCS_OK;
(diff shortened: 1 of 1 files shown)


yosefe (approved):
pls squash

src/ucs/sys/topo/base/topo.c
```diff
@@ -410,14 +410,17 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
...
+        if (ucs_topo_is_same_numa_node(device1, device2)) {
+            ucs_topo_common_numa_node_distance(distance);
+            return UCS_OK;
+        }
+
         ucs_topo_sys_root_distance(distance);
         return UCS_OK;
     }
 
+    /* Report best perf for common PCI bridge or sysfs parsing error */
```
brminich, line 423: 👍

brminich (approved):

PR #9478, opened 2023-11-08
UCS/SYS: Fix loading UCS modules for executable file with dot in name

dmitrygx:
## What

This patch fixes loading UCS modules for executable fileы with dot in name.

## Why ?

If some executable file contains `.` in the name (e.g. ".py" in case python file with UCX bindings), UCX assumes that it is file extension name (like ".so" or ".dll") and changes default `module_ext` value from ".so" to parsed value. It leads to the case that UCX modules can't be loaded.

## How ?

1. Remove code that searches `.` in the name of executable file and update `module_ext` value.
2. Define share library file extension through `shrext` from configure.ac.


yosefe:
does it happen when UCS is linked statically into another module?

dmitrygx:
> does it happen when UCS is linked statically into another module?

no, it happens with dynamically loaded libraries (aka modules), but when an executable file has `.` in its name, e.g. `ucx_perftest.py`

yosefe (dismissed):

dmitrygx:
@yosefe @brminich could you merge pls? CLA has already been signed. Thanks!

src/ucs/config/global_opts.c
```diff
@@ -192,6 +193,12 @@ static ucs_config_field_t ucs_global_opts_table[] = {
   "Directory to search for loadable modules",
   ucs_offsetof(ucs_global_opts_t, module_dir), UCS_CONFIG_TYPE_STRING},
 
+
```
yosefe, line 196: Pls remove extra space lines before and after MODULE_EXT field
otherwise LGTM

PR #9481, opened 2023-11-09
UCT/MM: Add trace for posix transport when shm has no space left

tvegas1:
## What
Trace when there is not enough space left on the shm for posix transport.

## Why ?
When `/dev/shm` is small and ppn is high, posix transport is not able to perform allocations. Add trace because posix usage is limited without allocation capability (no reg) and there are already many traces on that codepath anyways.

## How ?
Using `df -h /dev/shm` showed 64MB size.


src/uct/sm/mm/posix/mm_posix.c
```diff
@@ -141,8 +141,12 @@ uct_posix_md_query(uct_md_h tl_md, uct_md_attr_v2_t *md_attr)
     }
 
     shm_size = shm_statvfs.f_bsize * shm_statvfs.f_bavail;
-    uct_mm_md_query(&md->super, md_attr,
-                    (shm_size < posix_config->shm_min_size) ? 0 : shm_size);
+    if (shm_size < posix_config->shm_min_size) {
+        ucs_trace("md alloc disabled: only %zu bytes left in shm", shm_size);
```
brminich, line 145: maybe it can even be a debug trace, because it is not supposed to be invoked frequintly

tvegas1: ok for me, was putting trace to err on the safe side since this is called on the mem alloc for md method code path, @yosefe what do you think?

yosefe: should be ucs_debug because it is enabled in release mode

tvegas1: fixed

tvegas1: force-pushed for this simple one.

brminich (approved):

yosefe (approved):

PR #9483, opened 2023-11-12
UCT/IB/MLX5: using vst4q_u64 vector copy (SIMD) for arm neon

roiedanino:
## What
using `vst4q_u64` vector copy (SIMD) for arm neon instead of `int16x8_t` word copy in `bf_copy_bb`

## Why ?
Matching [rdma-core ](https://github.com/linux-rdma/rdma-core/blob/f16330dc0e079a6670a31fa54e89abb52994fd9d/util/mmio.h#L215) blueflame copy


yosefe:
@jgunthorpe FYI

PR #9494, opened 2023-11-15
CONTRIB: Add squash script to use after PR approval

tvegas1:
## What
Add a script to use for squashing branches after PR approval.

## Why ?
There can be one or more commits interleaved with zero or more merges. It's easier to use a script to do squashing.

## How ?
Auto-detect most recent base point. The commit message can later be amended. Usage:
```
./contrib/squash_commit.sh -h
usage: ./ucx_commit/contrib/squash_commit.sh -m <commit_message> [-r <remote_branch>] [-p <new_parent>]

Arguments
  -m <commit_message>     commit message to use
  -r <remote_branch>      upstream master branch to merge into
  -p <new_parent>         new base to use for squashed commit

Sample
    ./ucx_commit/contrib/squash_commit.sh -m "Commit message" -r upstream/master
    ./ucx_commit/contrib/squash_commit.sh -m "Commit message" -p adef30

```

### DIFF
--- a/contrib/squash_commit.sh
+++ b/contrib/squash_commit.sh
@@ -1,0 +1,77 @@
+#!/bin/bash -eEu
+#
+# Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2023. ALL RIGHTS RESERVED.
+# See file LICENSE for terms.
+#
+# Prepare one commit after PR has been approved.
+#
+
+message=
+parent=
+remote=
+
+usage() {
+    echo "usage: $0 -m <commit_message> [-r <remote_branch>] [-p <new_parent>]" >&2
+    echo "" >&2
+    echo "Arguments" >&2
+    echo "  -m <commit_message>     commit message to use" >&2
+    echo "  -r <remote_branch>      upstream master branch to merge into" >&2
+    echo "  -p <new_parent>         new base to use for squashed commit" >&2
+    echo "" >&2
+    echo "Sample" >&2
+    echo "    $0 -m \"Commit message\" -r upstream/master" >&2
+    echo "    $0 -m \"Commit message\" -p adef30" >&2
+    exit 1
+}
+
+param_checks() {
+    if [ -z "$message" ]
+    then
+        usage
+    fi
+
+    if [ -z "$parent" ] && [ -z "$remote" ]
+    then
+        echo "error: parent must be specified when remote is missing (-p)" >&2
+        exit 1
+    fi
+
+    if [ -z "$parent" ]
+    then
+        parent=$(git merge-base HEAD "$remote")
+    fi
+
+    object=$(git cat-file -t "$parent" || :)
+
+    if [ "$object" != "commit" ]
+    then
+        echo "error: parent sha1 $parent is not a commit" >&2
+        exit 1
+    fi
+}
+
+while getopts "hm:r:p:" o; do
+    case "$o" in
+    m)
+        message="$OPTARG"
+        ;;
+    r)
+        remote="$OPTARG"
+        ;;
+    p)
+        parent="$OPTARG"
+        ;;
+    *)
+        usage
+        ;;
+    esac
+done
+
+param_checks
+
+revision=$(git rev-parse HEAD)
+tree=$(git rev-parse "$revision"^{tree})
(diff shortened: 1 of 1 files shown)


contrib/squash_commit.sh
```diff
@@ -0,0 +1,102 @@
...
+then
+    echo "error: working directory has changes (run 'git reset --hard'?)" >&2
+    exit 1
+fi
+
+revision=$(git rev-parse HEAD)
+echo "Using revision $revision"
+
+set -x
+git clean -fdx
```
brminich, line 89: imo, this is not really required and may delete some oimportant info (logs, diffs, core, etc)

contrib/squash_commit.sh
```diff
@@ -0,0 +1,102 @@
...
+git reset "$parent"
+git add --all
+git commit -m "$message"
```
brminich, line 92:
```suggestion
git commit -am "$message"
```

contrib/squash_commit.sh
```diff
@@ -0,0 +1,102 @@
...
+        echo "error: parent must be specified when remote is missing (-p)" >&2
+        exit 1
+    fi
+
+    if [ -z "$parent" ]
+    then
+        parent=$(git merge-base HEAD "$remote")
+    fi
+
+    typ=$(git cat-file -t "$parent" || :)
```
brminich, line 44: minor: `type`?

tvegas1: type is keyword, replaced by object instead

contrib/squash_commit.sh
tvegas1: removed altogether

contrib/squash_commit.sh
tvegas1: for commits with added files, we later do add --all, so we need to clean the tree first. but agree on the deletion, pushing simpler approach to avoid the need to clean files.

contrib/squash_commit.sh
```diff
@@ -0,0 +1,77 @@
...
+        echo "error: parent must be specified when remote is missing (-p)" >&2
+        exit 1
+    fi
+
+    if [ -z "$parent" ]
+    then
+        parent=$(git merge-base HEAD "$remote")
+    fi
+
+    object=$(git cat-file -t "$parent" || :)
```
brminich, line 44: is `|| :` really needed here? What can be an error besides wrong `parent`? If it's wrong parent than object will be empty anyway

tvegas1: if the `parent` is bad as it can be input from command line, the git cli fails with an error message and script aborts since we have `bash -eE`

brminich: ok

brminich (dismissed):

brminich:
@Alexey-Rivkin, can you pls review as well?

contrib/squash_commit.sh
```diff
@@ -0,0 +1,77 @@
...
+done
+
+param_checks
+
+revision=$(git rev-parse HEAD)
+tree=$(git rev-parse $revision^{tree})
+commit=$(git commit-tree $tree -m "$message" -p $parent)
+
+git reset $commit
+git log -2 --stat
```
Alexey-Rivkin, line 77: No newline at the script end?

contrib/squash_commit.sh
```diff
@@ -0,0 +1,77 @@
...
+    *)
+        usage
+        ;;
+    esac
+done
+
+param_checks
+
+revision=$(git rev-parse HEAD)
+tree=$(git rev-parse $revision^{tree})
```
Alexey-Rivkin, line 73: Shellcheck suggests adding more quotes around variable expansions, e.g., 
`git rev-parse "$revision"^{tree}`

tvegas1: fixed all according to shellcheck

contrib/squash_commit.sh
tvegas1: there is an \x0a so we should be good

brminich (approved):

brminich:
can you please squash once approved by @Alexey-Rivkin?

Alexey-Rivkin (approved):

PR #9495, opened 2023-11-16
TEST/RKEY: Check rkey distance with accordance to fp8 precision

ivankochin:
## What
Check rkey and topo distance difference doesn't cross `UCS_FP8_PRECISION` ratio.

## Why ?
Now test uses hardcoded values, but with big BW difference can be more then 600 MB/s

Can be tested on rock partition by the following command: `GTEST_FILTER=*rc/test_ucp_mmap.reg_mem_type/0* ./test/gtest/gtest`

### DIFF
--- a/src/ucp/core/ucp_types.h
+++ b/src/ucp/core/ucp_types.h
@@ -117,16 +117,23 @@ extern const ucp_tl_bitmap_t ucp_tl_bitmap_min;
                    UCP_MAX_RESOURCES)
 
 
+#define UCS_FP8_MIN_BW  (512 * UCS_MBYTE)
+#define UCS_FP8_MAX_BW  (4 * UCS_TBYTE)
+#define UCS_FP8_MIN_LAT UCS_BIT(4)
+#define UCS_FP8_MAX_LAT UCS_BIT(17)
+#define UCS_FP8_MIN_OVH UCS_BIT(0)
+#define UCS_FP8_MAX_OVH UCS_BIT(12)
+
 /* Pack bandwidth as bytes/second, range: 512 MB/s to 4 TB/s */
-UCS_FP8_DECLARE_TYPE(BANDWIDTH, 512 * UCS_MBYTE, 4 * UCS_TBYTE)
+UCS_FP8_DECLARE_TYPE(BANDWIDTH, UCS_FP8_MIN_BW, UCS_FP8_MAX_BW)
 
 
 /* Pack latency as nanoseconds, range: 16 nsec to 131 usec */
-UCS_FP8_DECLARE_TYPE(LATENCY, UCS_BIT(4), UCS_BIT(17))
+UCS_FP8_DECLARE_TYPE(LATENCY, UCS_FP8_MIN_LAT, UCS_FP8_MAX_LAT)
 
 
 /* Pack overhead as nanoseconds, range: 1 nsec to 4 usec */
-UCS_FP8_DECLARE_TYPE(OVERHEAD, UCS_BIT(0), UCS_BIT(12))
+UCS_FP8_DECLARE_TYPE(OVERHEAD, UCS_FP8_MIN_OVH, UCS_FP8_MAX_OVH)
 
 
 /**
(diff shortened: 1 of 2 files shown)


brminich:
test failures are relevant

test/gtest/ucp/test_ucp_mmap.cc
```diff
@@ -355,18 +356,28 @@ bool test_ucp_mmap::disable_proto() const
...
-    EXPECT_NEAR(dist1.bandwidth, dist2.bandwidth, 600e6); /* 600 MBs accuracy */
-    EXPECT_NEAR(dist1.latency, dist2.latency, 20e-9); /* 20 nsec accuracy */
+    double allowed_diff_ratio = 1 - UCS_FP8_PRECISION;
+
+    /* Even if topo_value crosses the pack borders, it still can be not
+       capped in case of successfull cache entry catch */
+    if (((topo_value <= pack_min) && (rkey_value != pack_min)) ||
+        ((topo_value >= pack_max) && (rkey_value != pack_max))) {
+        EXPECT_NEAR(rkey_value, topo_value, topo_value * allowed_diff_ratio);
+    }
```
brminich, line 374: why are only out of range values checked now?

ivankochin: Good catch, thanks, fixed!

ivankochin: Good catch, thanks! Fixed.

test/gtest/ucp/test_ucp_mmap.cc
```diff
@@ -355,18 +356,32 @@ bool test_ucp_mmap::disable_proto() const
...
     /* Expect the implementation to always provide a reasonable precision w.r.t.
      * real-world bandwidth and latency ballpark numbers.
      */
-    EXPECT_NEAR(dist1.bandwidth, dist2.bandwidth, 600e6); /* 600 MBs accuracy */
-    EXPECT_NEAR(dist1.latency, dist2.latency, 20e-9); /* 20 nsec accuracy */
+    double allowed_diff_ratio = 1 - UCS_FP8_PRECISION;
+
+    if (rkey_value == pack_min) {
+        /* Capped by pack_min, no cache entry */
+        EXPECT_LE(topo_value, pack_min);
```
brminich, line 371: better to do casting here

src/ucp/core/ucp_types.h
```diff
@@ -117,16 +117,23 @@ extern const ucp_tl_bitmap_t ucp_tl_bitmap_min;
                    UCP_MAX_RESOURCES)
 
 
+#define UCS_FP8_MIN_BW  512 * UCS_MBYTE
```
tvegas1, line 120: nit: add parenthesis for completeness?

test/gtest/ucp/test_ucp_mmap.cc
```diff
@@ -355,18 +358,32 @@ bool test_ucp_mmap::disable_proto() const
...
+                                             size_t pack_max)
 {
     /* Expect the implementation to always provide a reasonable precision w.r.t.
      * real-world bandwidth and latency ballpark numbers.
      */
-    EXPECT_NEAR(dist1.bandwidth, dist2.bandwidth, 600e6); /* 600 MBs accuracy */
-    EXPECT_NEAR(dist1.latency, dist2.latency, 20e-9); /* 20 nsec accuracy */
+    double allowed_diff_ratio = 1 - UCS_FP8_PRECISION;
+
+    if (rkey_value == pack_min) {
```
brminich, line 371: it's also not casted

test/gtest/ucp/test_ucp_mmap.cc
```diff
@@ -355,18 +358,32 @@ bool test_ucp_mmap::disable_proto() const
...
+        /* Capped by pack_min, no cache entry */
+        EXPECT_LE(std::lround(topo_value), pack_min);
+    } else if (rkey_value == pack_max) {
```
brminich, line 374: it's also not casted

test/gtest/ucp/test_ucp_mmap.cc
ivankochin: In that case we set only exponent part for rkey value, so casting isn't required

brminich: ok, seems there should be no warnings

brminich (approved):

PR #9496, opened 2023-11-16
UCP/MM: Fix alignment when rcache disabled

Artemy-Mellanox:
## Why ?
With the addition of the dynamic alignment feature, MDs now specify alignment requirements and expect the registering region to align accordingly. By default, UCP relies on rcache to align the region. However, in cases where rcache is disabled, UCP should manage alignment on its own.

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -453,6 +453,9 @@ ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
     ucs_log_level_t err_level;
     ucp_md_index_t md_index;
     ucs_status_t status;
+    void *reg_address;
+    size_t reg_length;
+    size_t reg_align;
 
     if (reg_md_map == 0) {
         return UCS_OK;
@@ -505,12 +508,20 @@ ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
                                      UCT_MD_MEM_REG_FIELD_DMABUF_OFFSET;
         }
 
-        status = uct_md_mem_reg_v2(context->tl_mds[md_index].md, address,
-                                   length, &reg_params, &memh->uct[md_index]);
+        reg_address = address;
+        reg_length  = length;
+
+        if (context->rcache == NULL) {
+            reg_align = ucs_max(context->tl_mds[md_index].attr.reg_alignment, 1);
+            ucs_align_ptr_range(&reg_address, &reg_length, reg_align);
+        }
+
+        status = uct_md_mem_reg_v2(context->tl_mds[md_index].md, reg_address,
+                                   reg_length, &reg_params, &memh->uct[md_index]);
         if (ucs_unlikely(status != UCS_OK)) {
-            ucp_memh_register_log_fail(err_level, address, length, mem_type,
-                                       reg_params.dmabuf_fd, md_index, context,
-                                       status);
+            ucp_memh_register_log_fail(err_level, reg_address, reg_length,
+                                       mem_type, reg_params.dmabuf_fd, md_index,
+                                       context, status);
             if (uct_flags & UCT_MD_MEM_FLAG_HIDE_ERRORS) {
                 continue;
             }
@@ -521,7 +532,7 @@ ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
 
         ucs_trace("register address %p length %zu dmabuf-fd %d flags %ld "
                   "on md[%d]=%s %p",
-                  address, length,
+                  reg_address, reg_length,
                   (dmabuf_md_map & UCS_BIT(md_index)) ? reg_params.dmabuf_fd :
                                                         UCT_DMABUF_FD_INVALID,
                   reg_params.flags,
@@ -758,15 +769,15 @@ ucs_status_t ucp_memh_get_slow(ucp_context_h context, void *address,
     } else {
         status = ucp_memh_rcache_get(context->rcache, reg_address, reg_length,
                                      mem_type, reg_align, &memh);
+
+        ucs_assert(memh->mem_type == mem_type);
+        ucs_assert(ucs_padding((intptr_t)ucp_memh_address(memh), reg_align) == 0);
+        ucs_assert(ucs_padding(ucp_memh_length(memh), reg_align) == 0);
     }
 
     if (status != UCS_OK) {
         goto out;
     }
-
-    ucs_assert(memh->mem_type == mem_type);
-    ucs_assert(ucs_padding((intptr_t)ucp_memh_address(memh), reg_align) == 0);
-    ucs_assert(ucs_padding(ucp_memh_length(memh), reg_align) == 0);
 
     ucs_trace(
             "memh_get_slow: %s address %p/%p length %zu/%zu %s md_map %" PRIx64
--- a/test/gtest/ucp/test_ucp_rma.cc
+++ b/test/gtest/ucp/test_ucp_rma.cc
@@ -112,6 +112,14 @@ protected:
         return (100 * UCS_MBYTE) / ucs::test_time_multiplier();
     }
 
+    virtual unsigned size_step() {
+        return 4;
+    }
+
+    virtual unsigned max_iters() {
+        return 100;
+    }
+
     void test_mem_types(send_func_t send_func, size_t min_size = 128,
                         size_t max_size = default_max_size()) {
         const std::vector<std::vector<ucs_memory_type_t> >& pairs =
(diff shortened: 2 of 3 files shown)


tvegas1:
is adding a gtest possible for this one?

Artemy-Mellanox:
> is adding a gtest possible for this one?

it is covered by `shm_ib/test_ucp_mmap.reg_mem_type/3`

src/ucp/core/ucp_mm.c
```diff
@@ -437,6 +437,32 @@ static void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
     ucs_free(memh);
 }
 
+ucs_status_t ucp_memh_register_one(ucp_context_h context, ucp_mem_h memh,
```
ivankochin, line 440: I would prefer to have something like `ucp_memh_check_alignment` which would just adjust address and length. This function should be called inside `ucp_memh_register` before `uct_md_mem_reg_v2`. WDYT?

src/ucp/core/ucp_mm.c
```diff
@@ -437,6 +437,32 @@ static void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
...
+
+    if (context->rcache == NULL) {
+        reg_align   = ucs_max(context->tl_mds[md_index].attr.reg_alignment,
+                              UCS_RCACHE_MIN_ALIGNMENT);
+        reg_address = ucs_align_down_pow2_ptr(address, reg_align);
+        reg_length  = UCS_PTR_BYTE_DIFF(reg_address,
+                                        ucs_align_up_pow2_ptr(
+                                            UCS_PTR_BYTE_OFFSET(address,
+                                                                length),
+                                            reg_align));
```
ivankochin, line 456: Can we simplify this part by introducing additional variables like `buffer_end` and/or `aligned_buffer_end` to make it like:

```c
     buffer_end = UCS_PTR_BYTE_OFFSET(address, length);
     aligned_buffer_end = ucs_align_up_pow2_ptr(buffer_end, reg_align);
     reg_length = UCS_PTR_BYTE_DIFF(reg_address, aligned_buffer_end);
```

ivankochin:
> > is adding a gtest possible for this one?
> 
> it is covered by `shm_ib/test_ucp_mmap.reg_mem_type/3`

Do you know why CI didn't catch it during dynamic alignment testing?

src/ucp/core/ucp_mm.c
```diff
@@ -437,6 +437,32 @@ static void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
...
+                                   void *address, size_t length,
+                                   ucp_md_index_t md_index,
+                                   uct_md_mem_reg_params_t *reg_params)
+{
+    size_t reg_length, reg_align;
+    void *reg_address;
+
+    if (context->rcache == NULL) {
+        reg_align   = ucs_max(context->tl_mds[md_index].attr.reg_alignment,
+                              UCS_RCACHE_MIN_ALIGNMENT);
```
brminich, line 450: why is this value used as a minimum?

Artemy-Mellanox: It's minimum supported by pgtable implementation, same logic in `ucp_memh_reg_align()`

brminich: isn't it only relevant for rcache use-case?

brminich: Regarding `ucp_memh_reg_align()`, can you pls also check why [this assert](https://github.com/openucx/ucx/blob/master/src/ucp/core/ucp_mm.c#L768) does not fail when rcache is disabled?

src/ucp/core/ucp_mm.c
ivankochin: Would it looks ugly if we save both aligned and original pointers on the `ucp_memh_register` stack and then align only the required part. E.g.:

```C
ucp_memh_register(...) {
    void *address     = ucp_memh_address(memh);
    size_t length     = ucp_memh_length(memh);
    void *reg_address = address;
    size_t reg_length = length;

...

    ucp_memh_adjust_alignment(&reg_address, &reg_length);

...

    return uct_md_mem_reg_v2(context->tl_mds[md_index].md, reg_address,
                             reg_length, reg_params, &memh->uct[md_index]);
}

```

Even with the current approach I want to propose to rename `ucp_memh_register_one` to `ucp_memh_reg_align` or `ucp_memh_register_internal`. WDYT?

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -237,6 +240,7 @@ class test_ucp_rma : public test_ucp_memheap {
                            rkey, param);
     }
 
+protected:
```
ivankochin, line 243: Minor: It think it would be better to keep all the protected members in one place.

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -347,23 +331,83 @@ UCS_TEST_P(test_ucp_rma, get_blocking_zcopy, "ZCOPY_THRESH=0") {
...
+            modify_config("IB_MLX5_DEVX_OBJECTS", "", SETENV_IF_NOT_EXIST);
+        }
+        if (get_variant_value() & NO_RCACHE) {
+            modify_config("RCACHE_ENABLE", "n");
+        }
+        test_ucp_rma::init();
+    }
+
+protected:
+    void test_reg(send_func_t send_func, bool host_only) {
```
ivankochin, line 358: As I can see, `host_only` arg is always set to `host_only()` call. Can we check `host_only()` within the function and remove the argument?

src/ucp/core/ucp_mm.c
brminich: `ucp_memh_reg_align` looks good, but then better to use `ucp_memh_register_align[ed]` for consistency

src/ucp/core/ucp_mm.c
```diff
@@ -437,6 +437,29 @@ static void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
...
+    void *reg_address, *end;
+
+    if (context->rcache == NULL) {
+        reg_align   = ucs_max(context->tl_mds[md_index].attr.reg_alignment, 1);
+        reg_address = ucs_align_down_pow2_ptr(address, reg_align);
+        end         = UCS_PTR_BYTE_OFFSET(address, length);
+        end         = ucs_align_up_pow2_ptr(end, reg_align);
+        reg_length  = UCS_PTR_BYTE_DIFF(reg_address, end);
+    } else {
+        reg_address = address;
```
yosefe, line 455: I would add UCX function like:
ucs_align_ptr_range(void **address_p, size_t *length_p, size_t alignment)

And then do this before calling uct_md_mem_reg_v2 (in ucp_memh_register):
```c
if (rcache == NULL) {
  reg_align = ucs_max(context->tl_mds[md_index].attr.reg_alignment, 1);
  ucs_align_ptr_range(&address, &length, reg_align);
}
```

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -347,23 +331,83 @@ UCS_TEST_P(test_ucp_rma, get_blocking_zcopy, "ZCOPY_THRESH=0") {
...
 
-UCS_TEST_P(test_ucp_rma_reg_nb, put_blocking) {
-    test_message_sizes(static_cast<send_func_t>(&test_ucp_rma::put_b));
+
+class test_ucp_rma_reg : public test_ucp_rma {
+public:
+    static void get_test_variants(std::vector<ucp_test_variant>& variants) {
+        add_variant_with_value(variants, UCP_FEATURE_RMA, 0, "");
+        add_variant_with_value(variants, UCP_FEATURE_RMA, NON_BLOCK, "nb");
+        add_variant_with_value(variants, UCP_FEATURE_RMA, NON_BLOCK | NO_DEVX, "no_devx");
```
yosefe, line 340: maybe also add
NON_BLOCK | NO_RCACHE

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -347,23 +331,83 @@ UCS_TEST_P(test_ucp_rma, get_blocking_zcopy, "ZCOPY_THRESH=0") {
...
+protected:
+    void test_reg(send_func_t send_func, bool host_only) {
+        if (host_only) {
+            test_message_sizes(send_func, 128, default_max_size(),
+                               UCS_MEMORY_TYPE_HOST, UCS_MEMORY_TYPE_HOST, 0);
+            return;
+        }
+
+        const std::vector<ucs_memory_type_t>& mem_types =
+                mem_buffer::supported_mem_types();
```
yosefe, line 366: `host_only() ? {UCS_MEMORY_TYPE_HOST} : mem_buffer::supported_mem_types();`

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -347,23 +331,83 @@ UCS_TEST_P(test_ucp_rma, get_blocking_zcopy, "ZCOPY_THRESH=0") {
...
+                               mem_types[i], mem_types[i], 0);
+        }
+    }
+
+    virtual unsigned size_step() {
+        return 16;
+    }
+
+    virtual unsigned max_iters() {
+        return get_variant_value() & NO_RCACHE ? 1 : 10;
```
yosefe, line 379: 1. why without rcache only 1 iter?
2. add ( )

src/ucp/core/ucp_mm.c
```diff
@@ -505,8 +508,16 @@ ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
                                      UCT_MD_MEM_REG_FIELD_DMABUF_OFFSET;
         }
 
-        status = uct_md_mem_reg_v2(context->tl_mds[md_index].md, address,
-                                   length, &reg_params, &memh->uct[md_index]);
+        reg_address = address;
+        reg_length  = length;
```
yosefe, line 512: why need to copy to reg_address/reg_length?
IMO better to print log messages with the actual (aligned) address we are passing down to UCT

Artemy-Mellanox: need modify (align) addresses according to specific md before passing down or do I miss something?

yosefe: ok, probably don't want to change parameters.. so it means using reg_address/reg_length in the log/error messages instead of address/length

yosefe (approved):

brminich (approved):

PR #9497, opened 2023-11-17
UCT/CUDA/GDR_COPY: Release registration during rcache invalidation

Artemy-Mellanox:
## Why ?
Work-around gdrcopy [issue](https://redmine.mellanox.com/issues/3606732#change-28285685) with non-contig memory registration.

## How ?
As the gdrcopy TL is exclusively utilized during DT pack/unpack operations, it remains uninvolved in in-flight operations and can be released irrespective of rcache reference counting.


src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -116,6 +116,7 @@ uct_cuda_copy_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
                                       UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
     md_attr->dmabuf_mem_types       = md->config.dmabuf_supported ?
                                       UCS_BIT(UCS_MEMORY_TYPE_CUDA) : 0;
+    md_attr->stack_mem_types        = UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
```
brminich, line 119: why is it different with `cache_mem_types`?

PR #9499, opened 2023-11-20
UCP/PROTO: Do not change multi-fragment perf for 1 fragment range

ivankochin:
## What
Move additional multi-fragment overheads to multi-fragment ranges.

## Why ?
Currently multi-fragmented protocols changes perf by applying the fragment size to the performance function linear factor even for 1 fragment range which is wrong. We should calculate one-fragment perf using the provided untouched data for all the protocols and apply the additional PPLN overheads on multi-frag ranges.


brminich:
pls fix conflict in `proto_init.c`

brminich (approved):

brminich (approved):

brminich:
can you please squash?

ivankochin:
Customer confirmed that this patch fixes the issue.

yosefe (approved):

PR #9500, opened 2023-11-20
UCT/IB/UD: Completed flush immediately if no AM posted.

rakhmets:
## What
Completed flush operation immediately, if no Active Messages have been sent on the UD endpoint.

Introduced flag to distinguish UD endpoints from which Active Messages were sent.

## Why ?


src/uct/ib/ud/base/ud_inl.h
```diff
@@ -195,6 +200,7 @@ uct_ud_am_skb_common(uct_ud_iface_t *iface, uct_ud_ep_t *ep, uint8_t id,
 
     UCT_CHECK_AM_ID(id);
 
+    uct_ud_ep_set_am_flag(ep);
```
yosefe, line 203: can we do it in uct_ud_iface_complete_tx_skb?

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1215,6 +1215,13 @@ ucs_status_t uct_ud_ep_flush(uct_ep_h ep_h, unsigned flags,
         goto out;
     }
 
+    if ((ep->flags & UCT_UD_EP_FLAG_AM_POSTED) == 0) {
```
yosefe, line 1218: if (!(ep->flags & UCT_UD_EP_FLAG_AM_POSTED)) {
because it is a flag

src/uct/ib/ud/base/ud_ep.c
```diff
@@ -1215,6 +1215,13 @@ ucs_status_t uct_ud_ep_flush(uct_ep_h ep_h, unsigned flags,
...
+        ucs_trace("ep %p: not a single active message was posted", ep);
```
yosefe, line 1219: IMO better remove this debug message

rakhmets: Removed

src/uct/ib/ud/base/ud_inl.h
rakhmets: `{ud|ud_x}\test_ucp_peer_failure_keepalive.kill_receiver` will fail in this case.

`uct_ud_ep_get_tx_skb` returns `NULL`. So `uct_ud_{verbs|mlx5}_ep_am_bcopy` returns `UCS_ERR_NO_RESOURCE` right after that without setting the flag.

```
ucp_ep.c:1258 Bug: pending request 0x474ee60 (???) on ep 0x7f2669423000 should have been flushed

/auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/ucp/core/ucp_ep.c: [ ucp_destroyed_ep_pending_purge() ]
      ...
     1254 
     1255 void ucp_destroyed_ep_pending_purge(uct_pending_req_t *self, void *arg)
     1256 {
==>  1257     ucs_bug("pending request %p (%s) on ep %p should have been flushed",
     1258             self, ucs_debug_get_symbol_name(self->func), arg);
     1259 }
     1260 

==== backtrace (tid:  18110) ====
 0 0x000000000005f886 ucp_destroyed_ep_pending_purge()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/ucp/core/ucp_ep.c:1257
 1 0x000000000018900d uct_ud_ep_pending_purge_cb()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/uct/ib/ud/base/ud_ep.c:1752
 2 0x000000000005669a ucs_arbiter_group_purge()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/ucs/datastruct/arbiter.c:135
 3 0x00000000001892e8 uct_ud_ep_pending_purge()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/uct/ib/ud/base/ud_ep.c:1777
 4 0x0000000000060da0 uct_ep_pending_purge()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/uct/api/uct.h:3209
 5 0x000000000021975a ucp_wireup_send_ep_removed()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/ucp/wireup/wireup.c:870
 6 0x000000000021a04a ucp_wireup_msg_handler()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/ucp/wireup/wireup.c:940
 7 0x000000000017ed1e uct_iface_invoke_am()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/uct/base/uct_iface.h:942
 8 0x0000000000185d6a uct_ib_iface_invoke_am_desc()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/uct/ib/base/ib_iface.h:391
 9 0x0000000000185d6a uct_ud_ep_process_rx()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/uct/ib/ud/base/ud_ep.c:1064
10 0x000000000019263e uct_ud_verbs_iface_poll_rx()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/uct/ib/ud/verbs/ud_verbs.c:419
11 0x000000000019263e uct_ud_verbs_iface_progress()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/uct/ib/ud/verbs/ud_verbs.c:460
12 0x000000000008718a ucs_callbackq_dispatch()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/ucs/datastruct/callbackq.h:215
13 0x000000000009484b uct_worker_progress()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../src/uct/api/uct.h:2787
14 0x0000000000b50362 ucp_test_base::entity::progress()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../test/gtest/ucp/ucp_test.cc:1037
15 0x0000000000b4b100 ucp_test::progress()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../test/gtest/ucp/ucp_test.cc:167
16 0x0000000000b4b16e ucp_test::progress()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../test/gtest/ucp/ucp_test.cc:175
17 0x0000000000b4b199 ucp_test::short_progress_loop()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../test/gtest/ucp/ucp_test.cc:180
18 0x00000000008adcc6 ucp_test::wait_for_flag<unsigned long>()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../test/gtest/ucp/ucp_test.h:344
19 0x00000000008a43b4 test_ucp_peer_failure_keepalive_kill_receiver_Test::test_body()  /auto/mtrswgwork/rakhmetshin/ucx/contrib/../test/gtest/ucp/test_ucp_peer_failure.cc:599
```

yosefe: I see, it's a problem to return OK from flush when have no send resources.
BTW, why the code in https://github.com/yosefe/ucx/blob/c15da99d44c9b0a64070a20587eb4abd5e3f24b5/src/uct/ib/ud/base/ud_ep.c#L1146-L1155 is not causing flush to return UCS_OK ? according to the comment it should return OK if nothing was sent

rakhmets: We may set `UCT_UD_EP_FLAG_CONNECTED` after receiving `CREQ`. E.g. in case of two UD endpoints send `CREQ` simultaneous. But `ACK` message still have to be sent, so queue is not empty.

rakhmets: But we increase psn when sending creq.
Moreover, adding any additional check to the line 1151 will not completely solve the issue. We will still have hang in this case:
Two UD endpoints send CREQ simultaneously. UCT_UD_EP_FLAG_CONNECTED is set. uct_ud_ep_is_connected retruns true. But the queue is not empty, because ACK message still have to be sent. flush will not return UCS_OK in this case as well.

yosefe: hmmm
maybe take another approach: in UCP layer, no need to flush lanes that belong of a UCT iface that was not activated yet (if it was not activated, it means we didn't not have protocol to send on it yet)

PR #9501, opened 2023-11-21
UCP/PROTO: Unify data type checks

ivankochin:
## What
Unify the way of checking memory and data types validness. Allow sending empty messages on all data and memory types for all protocols.

## How
By introducing `inv_datatypes` map field in the protocol initialization parameters.

`ucp_proto_common_check_range` calls during single and multi initialization aren't required since all the requirements would be checked during `caps_init()`. These calls were added to avoid extra code execution in case of wrong datatype/memory type.


ivankochin:
Fixes this comment: https://github.com/openucx/ucx/pull/9399#discussion_r1392306104

ivankochin:
The CI failures caused by `ucp_datatype_iter_is_class` check `dt_mask`. That check probably can be removed since we decide what datatypes are supported during the proto initialization stage now.

ivankochin:
@yosefe @brminich WDYT about the approach suggested in the current patch?

PR #9506, opened 2023-11-24
UCP: Fine-grained intra/inter config for rendezvous

iyastreb:
## What
Introduced intra/inter-node fine-grained configuration for rendezvous switching.

## Why ?
Current implementation provides coarse-grained config (`UCX_RNDV_THRESH`) to set rendezvous switching threshold. There is a demand from some clients to introduce fine-grained config allowing to set this threshold separately for intra-node and inter-node communication.

## How ?
We introduce 2 new fine-grained configs for intra-node (`UCX_RNDV_INTRA_THRESH`) and inter-node (`UCX_RNDV_INTER_THRESH`) communication. If these configs are set to non-auto values, they have precedence over coarse-grained value (`UCX_RNDV_THRESH`).

**Testing**
Currently there is no easy way to test inter-node endpoint in gtest. So manual testing was performed to make sure the proper rendezvous thresholds are applied (with `ucp_client_server` and `ucp_hello_world`)

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -169,9 +169,26 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Threshold for switching from eager to rendezvous protocol",
    ucs_offsetof(ucp_context_config_t, rndv_thresh), UCS_CONFIG_TYPE_MEMUNITS},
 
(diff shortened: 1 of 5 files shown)


tvegas1 (commented):
seem commit title should be like `UCP/RNDV: Fine-grained..` for the two commits. (https://github.com/openucx/ucx/wiki/Guidance-for-contributors)

PR #9507, opened 2023-11-24
UCP: Fine-grained intra/inter config for rendezvous

iyastreb:
## What
Introduced intra/inter-node fine-grained configuration for rendezvous switching.
This change is **only supported by proto v2**. For proto v1 only coarse grained config is supported (e.g. `UCX_RNDV_THRESH=1000`)

## Why ?
Current implementation provides coarse-grained config (`UCX_RNDV_THRESH`) to set rendezvous switching threshold. There is a demand from some clients to introduce fine-grained config allowing to set this threshold separately for intra-node and inter-node communication.

## How ?
We use newly developed key-value parser in the config, that is used to configure fine-grained thresholds for intra and inter communication. Coarse-grained config (value for all) is still acceptable. Fine-grained config has precedence over coarse-grained one
```
UCX_RNDV_THRESH=1000  // 1000 is the limit for intra and inter
UCX_RNDV_THRESH=1000,intra:500  // 500 is the limit for intra, 1000 for inter
UCX_RNDV_THRESH=intra:100,inter:200  // explicit
```

**Testing**
Currently there is no easy way to test inter-node endpoint in gtest. So manual testing was performed to make sure the proper rendezvous thresholds are applied (with `ucp_client_server` and `ucp_hello_world`)

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -166,8 +166,14 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, bcopy_thresh), UCS_CONFIG_TYPE_MEMUNITS},
 
   {"RNDV_THRESH", UCS_VALUE_AUTO_STR,
-   "Threshold for switching from eager to rendezvous protocol",
-   ucs_offsetof(ucp_context_config_t, rndv_thresh), UCS_CONFIG_TYPE_MEMUNITS},
+   "Threshold for switching from eager to rendezvous protocol", 0,
+    UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_MEMUNITS,
+        {"intra", "threshold for intra-node communication",
+         ucs_offsetof(ucp_context_config_t, rndv_intra_thresh)},
+        {"inter", "threshold for inter-node communication",
+         ucs_offsetof(ucp_context_config_t, rndv_inter_thresh)},
+        {NULL}
+  )},
 
   {"RNDV_SEND_NBR_THRESH", "256k",
    "Threshold for switching from eager to rendezvous protocol in ucp_tag_send_nbr().\n"
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -2183,7 +2183,7 @@ ucp_ep_config_set_am_rndv_thresh(ucp_worker_h worker,
     ucs_assert(config->key.am_lane != UCP_NULL_LANE);
     ucs_assert(config->key.lanes[config->key.am_lane].rsc_index != UCP_NULL_RESOURCE);
 
-    if (context->config.ext.rndv_thresh == UCS_MEMUNITS_AUTO) {
+    if (context->config.ext.rndv_inter_thresh == UCS_MEMUNITS_AUTO) {
         /* auto - Make UCX calculate the AM rndv threshold on its own.*/
         status = ucp_ep_config_calc_rndv_thresh(worker, config,
                                                 config->key.am_bw_lanes,
@@ -2196,8 +2196,8 @@ ucp_ep_config_set_am_rndv_thresh(ucp_worker_h worker,
         rndv_local_thresh = context->config.ext.rndv_send_nbr_thresh;
         ucs_trace("active message rendezvous threshold is %zu", rndv_thresh);
     } else {
-        rndv_thresh       = context->config.ext.rndv_thresh;
-        rndv_local_thresh = context->config.ext.rndv_thresh;
+        rndv_thresh       = context->config.ext.rndv_inter_thresh;
+        rndv_local_thresh = context->config.ext.rndv_inter_thresh;
     }
 
     min_thresh     = ucs_max(iface_attr->cap.am.min_zcopy, min_rndv_thresh);
@@ -2233,7 +2233,7 @@ ucp_ep_config_set_rndv_thresh(ucp_worker_t *worker, ucp_ep_config_t *config,
 
     iface_attr = ucp_worker_iface_get_attr(worker, rsc_index);
 
-    if (context->config.ext.rndv_thresh == UCS_MEMUNITS_AUTO) {
+    if (context->config.ext.rndv_inter_thresh == UCS_MEMUNITS_AUTO) {
         /* auto - Make UCX calculate the RMA (get_zcopy) rndv threshold on its own.*/
         status = ucp_ep_config_calc_rndv_thresh(worker, config,
                                                 config->key.am_bw_lanes,
@@ -2244,8 +2244,8 @@ ucp_ep_config_set_rndv_thresh(ucp_worker_t *worker, ucp_ep_config_t *config,
 
         rndv_local_thresh = context->config.ext.rndv_send_nbr_thresh;
     } else {
-        rndv_thresh       = context->config.ext.rndv_thresh;
-        rndv_local_thresh = context->config.ext.rndv_thresh;
+        rndv_thresh       = context->config.ext.rndv_inter_thresh;
+        rndv_local_thresh = context->config.ext.rndv_inter_thresh;
     }
 
     min_thresh = ucs_max(iface_attr->cap.get.min_zcopy, min_rndv_thresh);
@@ -2371,8 +2371,8 @@ ucp_ep_config_max_short(ucp_context_t *context, uct_iface_attr_t *iface_attr,
     }
 
     if ((rndv_thresh != NULL) &&
-        (context->config.ext.rndv_thresh != UCS_MEMUNITS_AUTO)) {
-        /* Adjust max_short if rndv_thresh is set externally. Note local and
+        (context->config.ext.rndv_inter_thresh != UCS_MEMUNITS_AUTO)) {
+        /* Adjust max_short if rndv_inter_thresh is set externally. Note local and
          * remote threshold values are the same if set externally, so can
          * compare with just one of them. */
         ucs_assert(rndv_thresh->remote == rndv_thresh->local);
--- a/src/ucp/core/ucp_ep.inl
+++ b/src/ucp/core/ucp_ep.inl
@@ -251,6 +251,13 @@ ucp_ep_config_key_has_cm_lane(const ucp_ep_config_key_t *config_key)
     return config_key->cm_lane != UCP_NULL_LANE;
 }
 
+static UCS_F_ALWAYS_INLINE int
+ucp_ep_config_is_inter_node(const ucp_ep_config_key_t *config_key)
+{
+    return !(config_key->flags & (UCP_EP_CONFIG_KEY_FLAG_SELF |
+                                  UCP_EP_CONFIG_KEY_FLAG_INTRA_NODE));
+}
+
 static inline int ucp_ep_has_cm_lane(ucp_ep_h ep)
 {
     return (ep->cfg_index != UCP_WORKER_CFG_INDEX_NULL) &&
--- a/src/ucp/core/ucp_worker.c
+++ b/src/ucp/core/ucp_worker.c
@@ -331,7 +331,8 @@ static ucs_status_t ucp_worker_wakeup_init(ucp_worker_h worker,
      */
     if ((events & UCP_WAKEUP_TAG_SEND) ||
         ((events & UCP_WAKEUP_TAG_RECV) &&
-         (context->config.ext.rndv_thresh != UCS_MEMUNITS_INF)))
+         ((context->config.ext.rndv_intra_thresh != UCS_MEMUNITS_INF) ||
+          (context->config.ext.rndv_inter_thresh != UCS_MEMUNITS_INF))))
     {
         worker->uct_events |= UCT_EVENT_SEND_COMP;
     }
@@ -2159,6 +2160,11 @@ ucs_status_t ucp_worker_get_ep_config(ucp_worker_h worker,
                                         UCP_FEATURE_AM, UCP_OP_ID_AM_SEND,
                                         UCP_PROTO_FLAG_AM_SHORT, key->am_lane,
                                         &ep_config->am_u.max_eager_short);
+
+        ucp_worker_ep_config_short_init(worker, ep_config, ep_cfg_index,
+                                        UCP_FEATURE_AM, UCP_OP_ID_AM_SEND_REPLY,
+                                        UCP_PROTO_FLAG_AM_SHORT, key->am_lane,
+                                        &ep_config->am_u.max_reply_eager_short);
     }
 
     ucp_worker_print_used_tls(worker, ep_cfg_index);
--- a/src/ucp/rndv/proto_rndv.c
+++ b/src/ucp/rndv/proto_rndv.c
@@ -387,19 +387,29 @@ out:
     return status;
 }
 
-static size_t ucp_proto_rndv_thresh(const ucp_proto_init_params_t *init_params)
+size_t ucp_proto_rndv_thresh(const ucp_proto_init_params_t *init_params)
 {
     const ucp_proto_select_param_t *select_param = init_params->select_param;
     const ucp_context_config_t *cfg = &init_params->worker->context->config.ext;
+    size_t rndv_thresh;
 
-    if ((cfg->rndv_thresh == UCS_MEMUNITS_AUTO) &&
+    /* Explicit configuration for rendezvous threshold (either fine-grained or
+     * coarse-grained) has precedence over rndv_send_nbr_thresh
+     */
+    if (ucp_ep_config_is_inter_node(init_params->ep_config_key)) {
+        rndv_thresh = cfg->rndv_inter_thresh;
+    } else {
+        rndv_thresh = cfg->rndv_intra_thresh;
+    }
+
+    if ((rndv_thresh == UCS_MEMUNITS_AUTO) &&
         (ucp_proto_select_op_attr_unpack(select_param->op_attr) &
          UCP_OP_ATTR_FLAG_FAST_CMPL) &&
         ucs_likely(UCP_MEM_IS_HOST(select_param->mem_type))) {
-        return cfg->rndv_send_nbr_thresh;
+        rndv_thresh = cfg->rndv_send_nbr_thresh;
     }
 
-    return cfg->rndv_thresh;
+    return rndv_thresh;
 }
 
 ucs_status_t
--- a/test/gtest/ucp/test_ucp_tag.cc
+++ b/test/gtest/ucp/test_ucp_tag.cc
@@ -17,6 +17,7 @@ extern "C" {
 #include <ucs/arch/atomic.h>
 #include <ucs/memory/rcache.h>
 #include <ucs/memory/rcache_int.h>
+#include <ucp/proto/proto_select.inl>
 }
 
 #include <sys/mman.h>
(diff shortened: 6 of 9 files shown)


brminich:
pls check that rndv is preferred over short protocols on fast path in `ucp_tag_send_nbx` and `ucp_am_send_nbx`

iyastreb:
> pls check that rndv is preferred over short protocols on fast path in `ucp_tag_send_nbx` and `ucp_am_send_nbx`

I have addressed this review comment
It turned out that for tag matching all the thresholds were already properly applied.
For AM however I had to slightly change initialisation routine, so that max_reply_eager_short threshold is properly initialised. Updated all the tests, now also checking tag offloading setup

src/ucp/core/ucp_ep.inl
```diff
@@ -251,6 +251,13 @@ ucp_ep_config_key_has_cm_lane(const ucp_ep_config_key_t *config_key)
     return config_key->cm_lane != UCP_NULL_LANE;
 }
 
+static UCS_F_ALWAYS_INLINE int
+ucp_ep_config_key_is_inter_node(const ucp_ep_config_key_t *config_key)
```
brminich, line 255:
```suggestion
ucp_ep_config_is_inter_node(const ucp_ep_config_key_t *config_key)
```

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,116 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
 
     void init() {
+        /* TODO: Currently all the tests are for intra-node communication only.
+         * Find a way to create inter-node endpoint on a single node */
         test_ucp_tag::init();
-        check_offload_support(m_test_offload);
+
+        if (m_test_offload) {
+            check_offload_support(m_test_offload);
+        }
```
brminich, line 439: Seems the check is not needed. 

```suggestion
        check_offload_support(m_test_offload);
```

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,116 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
     }
 
-    static void get_test_variants(std::vector<ucp_test_variant>& variants) {
-        add_variant_with_value(variants, get_ctx_params(), 0, "");
-        add_variant_with_value(variants, get_ctx_params(), 1, "offload");
+    static void get_test_variants(std::vector<ucp_test_variant>& variants) {\
```
brminich, line 442:
```suggestion
    static void get_test_variants(std::vector<ucp_test_variant>& variants) {
```

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,116 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
+                EXPECT_EQ(cfg->tag.offload.max_eager_short.memtype_on + 1,
+                          min_rndv);
+                EXPECT_EQ(cfg->tag.offload.max_eager_short.memtype_off + 1,
+                          min_rndv);
+            } else {
+                EXPECT_LE(cfg->tag.offload.max_eager_short.memtype_on + 1,
+                          cfg_thresh);
+                EXPECT_LE(cfg->tag.offload.max_eager_short.memtype_off + 1,
+                          cfg_thresh);
+            }
```
brminich, line 486:
```suggestion
            cfg_thresh = ucs_max(min_rndv, cfg_thresh);
            EXPECT_LE(cfg->tag.offload.max_eager_short.memtype_on + 1,
                      cfg_thresh);
            EXPECT_LE(cfg->tag.offload.max_eager_short.memtype_off + 1,
                      cfg_thresh);
```

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,116 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
+    }
+
+    void check_am_rndv(size_t cfg_thresh)
+    {
+        ucp_ep_config_t *cfg = ucp_ep_config(sender().ep());
+
+        EXPECT_LE(cfg->am_u.max_eager_short.memtype_on + 1, cfg_thresh);
+        EXPECT_LE(cfg->am_u.max_eager_short.memtype_off + 1, cfg_thresh);
+        EXPECT_LE(cfg->am_u.max_reply_eager_short.memtype_on + 1, cfg_thresh);
+        EXPECT_LE(cfg->am_u.max_reply_eager_short.memtype_off + 1, cfg_thresh);
```
brminich, line 500: maybe define a short wrapper function, like 
```
check_short_thresh (ucp_memtype_thresh_t *thresh, size_t cfg_thresh)
{
    EXPECT_LE(thresh->memtype_on + 1, cfg_thresh);
    EXPECT_LE(thresh->memtype_off + 1, cfg_thresh);
}
```
And use it for checks in all funcs of these test

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,116 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
+        ucp_ep_config_t *cfg = ucp_ep_config(sender().ep());
+        ucp_proto_select_t *proto_select = &cfg->proto_select;
+        ucp_proto_select_elem_t value;
+        size_t msg_length = cfg_thresh;
+
+        /* Check tag matching thresholds */
+        if (m_test_offload) {
+            /* There is a lower bound for rndv threshold with tag offload.
+             * We should not send messages smaller than this limit */
+            size_t min_rndv = ucp_ep_tag_offload_min_rndv_thresh(sender().ucph(),
```
brminich, line 514: minor: maybe it can be done once at init stage and saved in some member variable

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -482,8 +577,49 @@ UCS_TEST_P(test_ucp_tag_limits, check_max_short_zcopy_thresh_zero, "ZCOPY_THRESH
...
+           "RNDV_INTRA_THRESH=20")
+{
+    check_rndv_startup_config(UCS_MEMUNITS_AUTO, 20, UCS_MEMUNITS_AUTO);
+    check_rndv_threshold(20);
+}
 
+UCS_TEST_P(test_ucp_tag_limits, check_rndv_intra_thresh_large,
+           "RNDV_INTRA_THRESH=2000")
+{
+    check_rndv_startup_config(UCS_MEMUNITS_AUTO, 2000, UCS_MEMUNITS_AUTO);
```
brminich, line 602: maybe also check `inf`?

test/gtest/ucp/test_ucp_tag.cc
iyastreb: Here the idea was to verify the exact match (with `EXPECT_EQ`) when configured rndv threshold is below the min_rndv limit. Otherwise check is more relaxed with unconditional `EXPECT_LE`.
I'm ok with this change, just want to make sure we are on the same page

test/gtest/ucp/test_ucp_tag.cc
iyastreb: sure

test/gtest/ucp/test_ucp_tag.cc
iyastreb: right, good idea indeed

test/gtest/ucp/test_ucp_tag.cc
iyastreb: sure, will be added

test/gtest/ucp/test_ucp_tag.cc
brminich: ok

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,133 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
+        check_short_thresh(cfg->am_u.max_eager_short, cfg_thresh);
+        check_short_thresh(cfg->am_u.max_reply_eager_short, cfg_thresh);
+    }
+
+    void check_ep_proto_rndv(size_t cfg_thresh, bool expect_rndv)
+    {
+        ucp_ep_config_t *cfg = ucp_ep_config(sender().ep());
+        ucp_proto_select_t *proto_select = &cfg->proto_select;
+        ucp_proto_select_elem_t value;
+        size_t msg_length = cfg_thresh;
```
brminich, line 510: According to coding style initialized vars go first:
```suggestion
        ucp_ep_config_t *cfg = ucp_ep_config(sender().ep());
        ucp_proto_select_t *proto_select = &cfg->proto_select;
        size_t msg_length = cfg_thresh;
        ucp_proto_select_elem_t value;
```

brminich (dismissed):

src/ucp/core/ucp_context.c
```diff
@@ -169,9 +169,26 @@ static ucs_config_field_t ucp_context_config_table[] = {
...
+   ucs_offsetof(ucp_context_config_t, rndv_intra_thresh), UCS_CONFIG_TYPE_MEMUNITS},
+
+  {"RNDV_INTER_THRESH", UCS_VALUE_AUTO_STR,
+   "Threshold for switching from eager to rendezvous protocol for inter-node\n"
+   "communication.\n"
+   " auto   - Threshold is defined by UCX_RNDV_THRESH\n"
+   " 0..inf - Explicit threshold for inter-node communication, has precedence\n"
+   "          over UCX_RNDV_THRESH config",
+   ucs_offsetof(ucp_context_config_t, rndv_inter_thresh), UCS_CONFIG_TYPE_MEMUNITS},
+
```
ivankochin, line 187: The separate settings for inter/intra modes can be useful not only for RNDV_THRESH but also for many system setting like transport bandwidth and latency. And creation of two additional variables for each of such cases seems not scalable for me.

I would propose to develop the new parser for such settings to let user set them in the following way or consider different more scalable way to create separate inter/intra node setting in common:

```
UCX_RNDV_THRESH=intra:1024,inter:256K
```

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,133 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
+
+        check_short_thresh(cfg->am_u.max_eager_short, cfg_thresh);
+        check_short_thresh(cfg->am_u.max_reply_eager_short, cfg_thresh);
+    }
+
+    void check_ep_proto_rndv(size_t cfg_thresh, bool expect_rndv)
+    {
+        ucp_ep_config_t *cfg = ucp_ep_config(sender().ep());
+        ucp_proto_select_t *proto_select = &cfg->proto_select;
+        size_t msg_length = cfg_thresh;
```
ivankochin, line 509: Align assignments

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,133 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
+            /* Find index of the corresponding ucp_proto_threshold_elem_t
+             * to handle the given message size */
+            unsigned idx = 0;
+            const ucp_proto_threshold_elem_t *th;
+            for (; msg_length > value.thresholds[idx].max_msg_length; ++idx) {
+                th = &value.thresholds[idx];
+                /* Assert no rndv before expected limit */
+                EXPECT_EQ(nullptr, strstr(th->proto_config.proto->name, "rndv"));
+            }
+            th = &value.thresholds[idx];
```
ivankochin, line 528: Minor: It seems like test doesn't required any information from `ucp_proto_threshold_elem_t` except `proto_config::name` and `proto_config::cfg_thresh` so there is no need to use logic based on threshold entity.

Pls consider this:
```suggestion
            const ucp_proto_config_t *proto_config;
            for (; msg_length > value.thresholds[idx].max_msg_length; ++idx) {
                proto_config = &value.thresholds[idx].proto_config;
                /* Assert no rndv before expected limit */
                EXPECT_EQ(nullptr, strstr(proto_config->proto->name, "rndv"));
            }
            proto_config = &value.thresholds[idx].proto_config;
```

Or even this:
```suggestion
            std::string proto_name;
            size_t cfg_thresh;
            
            for (; msg_length > value.thresholds[idx].max_msg_length; ++idx) {
                proto_name = value.thresholds[idx].proto_config.proto->name;
                /* Assert no rndv before expected limit */
                EXPECT_EQ(std::string::npos, proto_name.find("rndv"));
            }
            proto_name       = value.thresholds[idx].proto_config.proto->name;
            proto_cfg_thresh = value.thresholds[idx].proto_config.cfg_thresh;
```

test/gtest/ucp/test_ucp_tag.cc
iyastreb: done

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -482,8 +594,61 @@ UCS_TEST_P(test_ucp_tag_limits, check_max_short_zcopy_thresh_zero, "ZCOPY_THRESH
...
+}
+
+UCS_TEST_P(test_ucp_tag_limits, check_rndv_inter_thresh_common,
+           "RNDV_THRESH=1000", "RNDV_INTER_THRESH=30")
+{
+    check_rndv_startup_config(1000, UCS_MEMUNITS_AUTO, 30);
+    check_rndv_threshold(1000);
+}
+
+UCP_INSTANTIATE_TEST_CASE(test_ucp_tag_limits)
```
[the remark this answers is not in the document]
iyastreb: fixed

test/gtest/ucp/test_ucp_tag.cc
iyastreb: sure, makes sense

src/ucp/rndv/proto_rndv.c
```diff
@@ -388,11 +388,24 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+     */
+    if (ucp_ep_config_is_inter_node(init_params->ep_config_key)) {
+        if (cfg->rndv_inter_thresh != UCS_MEMUNITS_AUTO) {
+            return cfg->rndv_inter_thresh;
+        }
+    } else {
+        if (cfg->rndv_intra_thresh != UCS_MEMUNITS_AUTO) {
+            return cfg->rndv_intra_thresh;
+        }
+    }
```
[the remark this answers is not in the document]
iyastreb: Yes, as we agreed with Mikhail, this new functionality is supposed to be used in protov2 only.
I've added the appropriate description

src/ucp/core/ucp_context.c
iyastreb: Thanks, this idea looks interesting to me!

It adds a bit of complexity, but might be useful if there is real use case for this feature in other configs.

As discussed, we can even consider using YAML or JSON format for this config. Maybe YAML would be more suitable for environment variable config, because it's more concise and does not require quotes.
So the config would be defined like:
```
UCX_RNDV_THRESH=8k     // Coarse-grained config for all
UCX_RNDV_THRESH={intra:inf,inter:256k}  // Explicit fine-grained config for both
UCX_RNDV_THRESH={inter:256k}  // implicit "auto" value for intra
```
Let's discuss this topic with others to check:
1. If this feature really useful in other configs? Would it be over-engineering?
2. Should we use 3rd party library for parsing this config. YAML seems like a good candidate

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,131 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
-        m_env.push_back(new ucs::scoped_setenv("UCX_RC_TM_ENABLE",
-                                               ucs::to_string(m_test_offload).c_str()));
+        if (m_test_offload) {
+            m_env.push_back(new ucs::scoped_setenv("UCX_RC_TM_ENABLE", "y"));
+        }
+        m_min_rndv = 0;
     }
 
     void init() {
+        /* TODO: Currently all the tests are for intra-node communication only.
```
ivankochin, line 434: WDYT, would the following way help to imitate inter-node testing?

```c
    ucp_worker_cfg_index_t ep_cfg_index                  = sender().ep()->cfg_index;
    ucs_array_elem(&worker->ep_config, ep_cfg_index).key = 0;
```

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -482,6 +592,59 @@ UCS_TEST_P(test_ucp_tag_limits, check_max_short_zcopy_thresh_zero, "ZCOPY_THRESH
...
+    /* check that inter-node config is ignored for intra-node */
+    check_ep_proto_rndv(30, false);
+}
+
+UCS_TEST_P(test_ucp_tag_limits, check_rndv_inter_thresh_common,
+           "RNDV_THRESH=1000,inter:30")
+{
+    check_rndv_startup_config(1000, 30);
+    check_rndv_threshold(1000);
+}
```
ivankochin, line 646: As I can see almost all the tests constains the same scenario:
1. Set RNDV thresholds environement
2. check_rndv_startup_config
3. check_rndv_threshold

WDYT, would it has sense to unify all this checks under one test and do something like:
```c++
struct rndv_thresh {
    size_t inter;
    size_t intra;
    const char *string;
}

UCS_TEST_P(test_ucp_tag_limits, check_rndv_thresh) {
    std::vector<rndv_thresh> env_variants = {
        {0, 0, 0, "0"},
        {UCS_MEMUNITS_AUTO, 20, UCS_MEMUNITS_AUTO, "intra:20"},
        ....
    }

    for (const auto& env: env_variants) {
         UCS_TEST_MESSAGE << " Check RNDV thresholds for UCX_RNDV_THRESH=" << env.string;
         ucp_config_modify(config, "RNDV_THRESH", env.string);
         check_rndv_startup_config(env.intra, env.inter);
         check_rndv_threshold(env.intra);
    }
}
```

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,131 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
+        ucp_ep_config_t *cfg = ucp_ep_config(sender().ep());
+
+        if (m_test_offload) {
+            /* If configured threshold is less than min_rndv, then expect exact
+             * min_rndv limit for short messages */
+            if (cfg_thresh < m_min_rndv) {
+                EXPECT_EQ(cfg->tag.offload.max_eager_short.memtype_on + 1,
+                          m_min_rndv);
+                EXPECT_EQ(cfg->tag.offload.max_eager_short.memtype_off + 1,
+                          m_min_rndv);
```
ivankochin, line 486: Minor:

```suggestion
        ucp_ep_config_t *cfg  = ucp_ep_config(sender().ep());
        auto &max_eager_short = cfg->tag.offload.max_eager_short;
        if (m_test_offload) {
            /* If configured threshold is less than min_rndv, then expect exact
             * min_rndv limit for short messages */
            if (cfg_thresh < m_min_rndv) {
                EXPECT_EQ(max_eager_short.memtype_on + 1, m_min_rndv);
                EXPECT_EQ(max_eager_short.memtype_off + 1, m_min_rndv);
```

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -423,22 +424,131 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
+    void check_rndv_threshold(size_t cfg_thresh)
+    {
+        ucp_context_config_t *cfg = &sender().worker()->context->config.ext;
+        if (cfg->proto_enable) {
+            /* Check proto_v2 rndv thresholds only when this protocol is
+             * enabled, otherwise these checks are irrelevant */
+            check_tag_rndv_v2(cfg_thresh);
+            check_am_rndv_v2(cfg_thresh);
+        }
+        check_ep_proto_rndv(cfg_thresh, true);
```
ivankochin, line 550: Wouldn't `check_ep_proto_rndv` fail in case if `cfg->proto_enable` is not true? As I undestand it fully relies on `proto_select` which exists only for `proto_enable`.

brminich:
can this failure be related?
```
[ RUN      ] all/test_ucp_cuda.sparse_regions/0 <all>
/scrap/azure/agent-03/AZP_WORKSPACE/1/s/contrib/../test/gtest/common/mem_buffer.cc:223: Failure
cudaMalloc(&ptr, size) failed: unknown error: size=4096
[  FAILED  ] all/test_ucp_cuda.sparse_regions/0, where GetParam() = all (345 ms)
```

iyastreb:
> can this failure be related?
> 
> ```
> [ RUN      ] all/test_ucp_cuda.sparse_regions/0 <all>
> /scrap/azure/agent-03/AZP_WORKSPACE/1/s/contrib/../test/gtest/common/mem_buffer.cc:223: Failure
> cudaMalloc(&ptr, size) failed: unknown error: size=4096
> [  FAILED  ] all/test_ucp_cuda.sparse_regions/0, where GetParam() = all (345 ms)
> ```

I don't think so.
I quickly checked the source code of that test, it seems to be completely unrelated to rendezvous feature

src/ucp/core/ucp_ep.c
```diff
@@ -2183,7 +2183,7 @@ ucp_ep_config_set_am_rndv_thresh(ucp_worker_h worker,
     ucs_assert(config->key.am_lane != UCP_NULL_LANE);
     ucs_assert(config->key.lanes[config->key.am_lane].rsc_index != UCP_NULL_RESOURCE);
 
-    if (context->config.ext.rndv_thresh == UCS_MEMUNITS_AUTO) {
+    if (context->config.ext.rndv_inter_thresh == UCS_MEMUNITS_AUTO) {
```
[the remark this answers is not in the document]
iyastreb: Yes, that's the idea.

Basically I considered 2 approaches:
1. Keep common variable `rndv_thresh` together with fine-grained settings.
(+) This way we could keep the existing code the same
(-) It would complicate the key-value parser by adding a concept of "common config" offset

2. Replace common config with fine-grained settings
(-) We need to adapt the existing code for protov1
(+) We have cleaner design for key-value feature

I raised this question last year during the meeting, and IIRC Yossi supported the second approach, therefore this change

src/ucp/core/ucp_worker.c
```diff
@@ -331,7 +331,8 @@ static ucs_status_t ucp_worker_wakeup_init(ucp_worker_h worker,
      */
     if ((events & UCP_WAKEUP_TAG_SEND) ||
         ((events & UCP_WAKEUP_TAG_RECV) &&
-         (context->config.ext.rndv_thresh != UCS_MEMUNITS_INF)))
+         (context->config.ext.rndv_intra_thresh != UCS_MEMUNITS_INF) &&
+         (context->config.ext.rndv_inter_thresh != UCS_MEMUNITS_INF)))
```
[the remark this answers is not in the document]
iyastreb: You're right, will use OR instead

src/ucp/core/ucp_worker.c
```diff
@@ -2159,6 +2160,11 @@ ucs_status_t ucp_worker_get_ep_config(ucp_worker_h worker,
                                         UCP_FEATURE_AM, UCP_OP_ID_AM_SEND,
                                         UCP_PROTO_FLAG_AM_SHORT, key->am_lane,
                                         &ep_config->am_u.max_eager_short);
+
+        ucp_worker_ep_config_short_init(worker, ep_config, ep_cfg_index,
```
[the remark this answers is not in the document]
iyastreb: Sure.
It's related to this PR, but a bit indirectly.
While testing this little feature I discovered that existing `max_short` limits are correctly adjusted based on current RNDV value. Namely the following short limits were properly adjusted:
```
tag.max_eager_short
tag.offload.max_eager_short
am_u.max_eager_short
```
The adjustment basically means that short limit can't exceed the RNDV threshold. My understanding is that this adjustment is needed because we want switch to RNDV protocol even if configured value of RNDV threshold is below some "short limits".
However the very similar short threshold `am_u.max_reply_eager_short` was not adjusted accordingly. We discussed this with Mikhail and agreed that the common adjustment must be done for this limit as well

test/gtest/ucp/test_ucp_tag.cc
iyastreb: Mmm, honestly I wouldn't go this way for 2 reasons:

1. Apart from the function calls that you mentioned, each test performs initialization part which is not trivial: look at `test_ucp_tag_limits::init`. It creates sender and receiver, init eps, connect them...
With your approach we do this init only once, and then just change the config.
But I doubt that changing config affects already created eps. ?
If so, then we cannot test runtime config

2. Even if p1 worked out, personally I would still prefer to have these tests independent of each other, for debugability reason.

test/gtest/ucp/test_ucp_tag.cc
iyastreb: Thanks for the hint, will try that

src/ucp/core/ucp_ep.c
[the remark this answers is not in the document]
ivankochin: Completely agree with the current approach, but I have an idea how to possibly improve it.

I suggest to add `proto_disable_rndv_thresh` field to `ucp_context_config` structure, set it to `rndv_inter_thresh` inside `ucp_fill_config()`, also leave the comment with something like "this variable is used for protov1 only because protov2 supports different thresholds for inter/intra nodes" near the initialization place and then use this variable in all protov1 related places.

I understand that this is debatable suggestion because of extra-field which is used only for better code self-documentation so it is up to you.

But anyway I think it makes sense to insert some comments about that and maybe check somewhere whether we use protov1 and have different inter/intra thresholds and print warning in that case.

test/gtest/ucp/test_ucp_tag.cc
ivankochin: I think the first point can be potentially solved by explicit `test::init()` and `test::cleanup()` calls, but that can be complicated....

Ok, up to you :)

test/gtest/ucp/test_ucp_tag.cc
iyastreb: Thanks for spotting! Well, these tests pass CI jobs with proto_disabled

But I think they pass because inside `check_ep_proto_rndv` we iterate over protocols hash map
```
        kh_foreach_value(proto_select->hash, value, {
```
which is supposedly empty in case of proto_disable. So essentially this test may test nothing for v1)

I will run this test with PROTO disabled and let you know the outcome.
If it's like that, then all runtime tests are v2 specific and I'll update the code accordingly

test/gtest/ucp/test_ucp_tag.cc
iyastreb: No, it does not build, but if it does, it will be useless.

The problem is that code is pretty much tightly coupled in that part.
The interesting part happens in `ucp_wireup_init_lanes` function that does:
```
ucp_wireup_select_lanes
  ucp_wireup_construct_lanes
    ucp_wireup_select_set_locality_flags
      assign UCP_EP_CONFIG_KEY_FLAG_INTRA_NODE
```
After that the EP is considered as local
And soon after that, in the same function we initialize the config, that discovers all the protocols: `ucp_worker_get_ep_config`.

So if we want to mock inter-node communication, we need to change ep config between these 2 calls.
How? I see some alternatives
1. play with `ucp_wireup_check_config_intersect` which is in the middle?
2. introduce new config param, e.g. force_inter
3. use mock library/framework. e.g. cpputest
4. mock at linker level, e.g. by using weak link for `ucp_wireup_check_config_intersect` function

So far I don't see an easy way to mock inter-node communication, I've already spent few days in November trying to mock it, but no luck. For now we agreed with Mikhail that I don't spend more time on this, I just tested inter-node communication manually. But IMO this might be useful in other places too

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -482,6 +595,59 @@ UCS_TEST_P(test_ucp_tag_limits, check_max_short_zcopy_thresh_zero, "ZCOPY_THRESH
...
+    check_ep_proto_rndv_v2(30, false);
+}
+
+UCS_TEST_P(test_ucp_tag_limits, check_rndv_inter_thresh_common,
+           "RNDV_THRESH=1000,inter:30")
+{
+    check_rndv_startup_config(1000, 30);
+    check_rndv_threshold(1000);
+}
+
```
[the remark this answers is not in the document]
iyastreb: sure, will be added

brminich (dismissed):

yosefe (approved):

yosefe:
@iyastreb pls squash

yosefe (approved):

PR #9509, opened 2023-11-24
UCT/CUDA/BASE: Get device for current context only if context is active.

rakhmets:
## What
Added condition to call `cuCtxGetDevice` only if there is a CUDA context bound to the calling CPU thread. Otherwise, `cuDeviceGet(ordinal = 0)` is used in query method.

## Why ?
Previously `cuCtxGetDevice` could return `CUDA_ERROR_INVALID_CONTEXT` error. The error was not traced, but could be caught by `compute-sanitaizer`.

### DIFF
--- a/src/uct/cuda/base/cuda_iface.c
+++ b/src/uct/cuda/base/cuda_iface.c
@@ -36,12 +36,23 @@ uct_cuda_base_query_devices_common(
         uct_md_h md, uct_device_type_t dev_type,
         uct_tl_device_resource_t **tl_devices_p, unsigned *num_tl_devices_p)
 {
-    ucs_sys_device_t sys_device = UCS_SYS_DEVICE_ID_UNKNOWN;
+    ucs_sys_device_t sys_device;
     CUdevice cuda_device;
+    ucs_status_t status;
 
-    if (cuCtxGetDevice(&cuda_device) == CUDA_SUCCESS) {
-        uct_cuda_base_get_sys_dev(cuda_device, &sys_device);
+    if (uct_cuda_base_is_context_active()) {
+        status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetDevice(&cuda_device));
+    } else {
+        ucs_debug("using locality information of cuda device 0 as no context is"
+                  " currently active");
+        status = UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&cuda_device, 0));
     }
+
+    if (status != UCS_OK) {
+        return UCS_ERR_NO_DEVICE;
+    }
+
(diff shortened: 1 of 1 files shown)


brminich (dismissed):

src/uct/cuda/base/cuda_iface.c
```diff
@@ -39,8 +39,13 @@ uct_cuda_base_query_devices_common(
...
     CUdevice cuda_device;
 
-    if (cuCtxGetDevice(&cuda_device) == CUDA_SUCCESS) {
+    if (uct_cuda_base_is_context_active()) {
+        if (UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetDevice(&cuda_device)) != UCS_OK) {
+            return UCS_ERR_NO_DEVICE;
+        }
         uct_cuda_base_get_sys_dev(cuda_device, &sys_device);
+    } else {
+        ucs_debug("querying devices without active context");
```
yosefe, line 48: this means we won't know the sys device of the GPU? it can affect protocols and devices selection..

rakhmets: Yes. And this is current behavior. `cuCtxGetDevice` returns `CUDA_ERROR_INVALID_CONTEXT`, if there is no active device.
We can try to add here `cuDeviceGet(&cuda_device, 0)`, assuming that 0-th device will be used...

yosefe: Yes, let's do this, and also update the debug message to print that no device is selected and we're taking locality information from device 0

src/uct/cuda/base/cuda_iface.c
```diff
@@ -36,13 +36,23 @@ uct_cuda_base_query_devices_common(
...
+    ucs_sys_device_t sys_device;
     CUdevice cuda_device;
+    ucs_status_t status;
 
-    if (cuCtxGetDevice(&cuda_device) == CUDA_SUCCESS) {
-        uct_cuda_base_get_sys_dev(cuda_device, &sys_device);
+    if (uct_cuda_base_is_context_active()) {
+        status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetDevice(&cuda_device));
+    } else {
+        ucs_debug("no device is selected, zero device will be used");
```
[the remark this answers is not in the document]
rakhmets: Fixed.

yosefe (approved):

src/uct/cuda/base/cuda_iface.c
rakhmets: Reverted this change because of wire incompatibility issue. https://github.com/openucx/ucx/pull/9529

PR #9489, opened 2023-11-14
UCT/IB/DC: add env for reverse_sl (default: same value of sl)

roiedanino:
## What
Set reverse_sl to the same value of sl (service level) configured by `UCX_IB_SL` env.

## Why ?
Currently on certain device configurations reverse_sl is 0 if not set.

PR #9510, opened 2023-11-26
UCT/IB: fixed reverse_sl feature - removed code-dup and parentheses

roiedanino:
## What
Fixing code duplication when setting reverse_sl in different ib transports.

## Why ?
PR https://github.com/openucx/ucx/pull/9489 was already merged

### DIFF
--- a/src/uct/ib/rc/accel/rc_mlx5_iface.c
+++ b/src/uct/ib/rc/accel/rc_mlx5_iface.c
@@ -826,10 +826,6 @@ UCS_CLASS_INIT_FUNC(uct_rc_mlx5_iface_common_t, uct_iface_ops_t *tl_ops,
     status = uct_ib_mlx5_iface_select_sl(&self->super.super,
                                          &mlx5_config->super,
                                          &rc_config->super);
-    self->super.super.config.reverse_sl = (rc_config->super.reverse_sl ==
-                                           UCS_ULUNITS_AUTO) ?
-                                                  self->super.super.config.sl :
-                                                  rc_config->super.reverse_sl;
 
     if (status != UCS_OK) {
         return status;
--- a/src/uct/ib/base/ib_iface.c
+++ b/src/uct/ib/base/ib_iface.c
@@ -1303,15 +1303,16 @@ uint8_t uct_ib_iface_config_select_sl(const uct_ib_iface_config_t *ib_config)
     return (uint8_t)ib_config->sl;
 }
 
-uint8_t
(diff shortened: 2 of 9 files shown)


src/uct/ib/rc/accel/rc_mlx5_iface.c
```diff
@@ -826,10 +826,8 @@ UCS_CLASS_INIT_FUNC(uct_rc_mlx5_iface_common_t, uct_iface_ops_t *tl_ops,
     status = uct_ib_mlx5_iface_select_sl(&self->super.super,
                                          &mlx5_config->super,
                                          &rc_config->super);
-    self->super.super.config.reverse_sl = (rc_config->super.reverse_sl ==
-                                           UCS_ULUNITS_AUTO) ?
-                                                  self->super.super.config.sl :
-                                                  rc_config->super.reverse_sl;
+    self->super.super.config.reverse_sl = uct_ib_iface_config_select_reverse_sl(
+            &rc_config->super, self->super.super.config.sl);
```
yosefe, line 830: can we pass just ib_config struct to this function? it has both sl and reverse_sl

roiedanino: The issue is in some places this function is being called after selecting an sl:
`self->super.super.config.sl = uct_ib_iface_config_select_sl(&config->super);`

In those cases, we would want to use the selected sl and not the configured one, `self->super.super.config` is not the same struct as `rc_config->super`.

`self->super.super.config.sl` default value is 16 (an illegal SL)
while the default for `rc_config->super.sl`, coming from env configuration would be: `UCS_ULUNITS_AUTO`

yosefe: can we unite the selection of sl and reverse_sl to a single function that operates on ib_config and/or ib_iface?
if needed it can accept additional parameters such as mask of sl with AR, on transports that support it

test/gtest/uct/ib/test_ib.cc
```diff
@@ -557,6 +557,7 @@ class test_uct_ib_sl_utils : public test_uct_ib_utils {
         uint16_t ooo_sl_mask = (test_ooo_sl_mask !=
                                 m_ooo_sl_mask_not_detected) ?
                                static_cast<uint16_t>(test_ooo_sl_mask) : 0;
+
```
roiedanino, line 560: remove

roiedanino:
> @roiedanino seems there are many unrelated changes in the commit

Oops probably clang, will fix it

yosefe (approved):

PR #9512, opened 2023-11-29
UCP/WIREUP: Consider local distance during slow lanes dropping

ivankochin:
## What
Make lanes filtering function aware of local distance.

## Why ?
Remote BW value already capped by distance during the comparison, but local BW doesn't consider distance. That can cause reconfiguration errors because the peers can calculate different BW numbers for the same lane.


pentschev (commented):
I can confirm this resolves the issue I was experiencing on my end. Thanks @ivankochin !

src/ucp/wireup/select.c
```diff
@@ -1469,12 +1469,16 @@ static double ucp_wireup_get_lane_bw(ucp_worker_h worker,
...
+    ucp_worker_iface_t *wiface = ucp_worker_iface(worker, sinfo->rsc_index);
     double bw_local, bw_remote;
 
-    iface_attr = ucp_worker_iface_get_attr(worker, sinfo->rsc_index);
-    bw_local   = ucp_tl_iface_bandwidth(context, &iface_attr->bandwidth);
+    if (address->dst_version < 17) {
+        bw_local = ucp_tl_iface_bandwidth(worker->context,
+                                          &wiface->attr.bandwidth);
+    } else {
+        bw_local = ucp_wireup_iface_bw_distance(wiface);
```
brminich, line 1479: pls add a short comment what it fixes
👍 otherwise

yosefe (approved):

src/ucp/wireup/select.c
ivankochin: Done.

brminich (approved):

PR #9513, opened 2023-11-29
UCS/RCACHE: Synchronize events

Artemy-Mellanox:
## Why ?
The UCP rcache currently depends on UCP synchronization, making the pgt_lock unreliable for memory event handling.

## How ?
Always queue event's inv_q to handle it by upcoming region creation process.

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -1435,6 +1435,7 @@ ucp_mem_rcache_create(ucp_context_h context, const char *name,
                       ucs_rcache_t **rcache_p, ucs_rcache_params_t *rcache_params)
 {
     rcache_params->region_struct_size = ucp_memh_size(context);
+    rcache_params->flags             |= UCS_RCACHE_FLAG_SYNC_EVENTS;
     rcache_params->ucm_events         = UCM_EVENT_VM_UNMAPPED |
                                         UCM_EVENT_MEM_TYPE_FREE;
     rcache_params->context            = context;
@@ -1716,13 +1717,12 @@ ucp_memh_import(ucp_context_h context, const void *export_mkey_buffer,
                  * exists in the RCACHE of imported regions, it means that
                  * an exported memory handle has already been destroyed for a
                  * given address, but an imported memory handle hasn't been
-                 * retrieved from the RCACHE yet. So, it should have reference
-                 * counter == 1, since ucp_mem_unmap() should be invoked for
-                 * unused imported memory handles and then - for corresponding
-                 * exported ones */
-                ucs_assertv(rregion->refcount == 1, "%u", rregion->refcount);
+                 * removed from the RCACHE yet. So, it had refcount == 1 and
+                 * now it should be 2. */
+                ucs_assertv(rregion->refcount == 2, "%u", rregion->refcount);
                 ucs_rcache_region_invalidate(rcache, rregion,
                                              ucs_empty_function, NULL);
+                ucs_rcache_region_put_unsafe(rcache, rregion);
             }
         }
 
--- a/test/gtest/common/test_helpers.cc
+++ b/test/gtest/common/test_helpers.cc
@@ -628,17 +628,16 @@ uint16_t get_port() {
     return port;
 }
 
-void *mmap_fixed_address(size_t length) {
-    void *ptr = mmap(NULL, length, PROT_READ | PROT_WRITE,
-                     MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
-    if (ptr == MAP_FAILED) {
-        return nullptr;
+mmap_fixed_address::mmap_fixed_address(size_t length) : m_length(length) {
+    m_ptr = mmap(NULL, m_length, PROT_READ | PROT_WRITE,
+                 MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    if (m_ptr == MAP_FAILED) {
+        UCS_TEST_ABORT("mmap failed to allocate memory region");
     }
+}
 
-    munmap(ptr, length);
-
-    /* coverity[use_after_free] */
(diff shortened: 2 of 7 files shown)


test/gtest/ucp/test_ucp_mmap.cc
```diff
@@ -861,10 +861,7 @@ UCS_TEST_P(test_ucp_mmap, fixed) {
 
     for (int i = 0; i < 1000 / ucs::test_time_multiplier(); ++i) {
         size_t size = (i + 1) * ((i % 2) ? 1000 : 1);
-        void *ptr   = ucs::mmap_fixed_address(size);
-        if (ptr == nullptr) {
-            UCS_TEST_ABORT("mmap failed to allocate memory region");
-        }
+        void *ptr   = (void *)0x300000000000;
```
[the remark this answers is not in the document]
Artemy-Mellanox: this change breaking `ucs::mmap_fixed_address()` because of interference with `ucs_mpool_get(&rcache->mp)`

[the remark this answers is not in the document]
yosefe: why?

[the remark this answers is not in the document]
Artemy-Mellanox: there are no guarantee that address returned by ucs::mmap_fixed_address() is not occupied, see explanation in comment above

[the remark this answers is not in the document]
yosefe: can we provide a workaround for `ucs_mpool_get(&rcache->mp)` as well?
since there is no guarantee that 0x300000000000 will work on all systems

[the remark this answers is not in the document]
yosefe: if it pass CI then it's OK
but need to make sure we release the address eventually and not leak it

src/ucp/core/ucp_mm.c
```diff
@@ -1595,7 +1600,7 @@ ucp_memh_import_slow(ucp_context_h context, ucs_rcache_t *existing_rcache,
 
             ucs_rcache_set_default_params(&rcache_params);
 
-            status = ucp_mem_rcache_create(context, rcache_name, &rcache,
+            status = ucp_mem_rcache_create(context, rcache_name, &rcache, 0,
```
yosefe, line 1603: why need to avoid events for imported memory in this PR, does it solve some CI failure?

Artemy-Mellanox: `shm_ib/test_ucp_mmap_export.reg_export_and_reimport/2` started to fail on `swx-rain03-bf1`.
isn't it correct fix? this rcache caching memory registrations of foreign host, why should we invalidate basing on memory events happening in local host?

test/gtest/uct/test_mem.cc
```diff
@@ -209,9 +209,10 @@ UCS_TEST_P(test_mem, md_fixed) {
                     status = uct_mem_free(&uct_mem);
                 } else {
                     EXPECT_EQ(status, UCS_ERR_NO_MEMORY);
+                    munmap(curr_addr, page_size);
```
yosefe, line 212: I think mmap_fixed_address should release the memory, or use another approach - since the test maps and unmaps memory, it may still hit the issue that something else gets mapped to this range.

test/gtest/common/test_helpers.cc
```diff
@@ -628,17 +628,16 @@ uint16_t get_port() {
     return port;
 }
 
-void *mmap_fixed_address(size_t length) {
-    void *ptr = mmap(NULL, length, PROT_READ | PROT_WRITE,
-                     MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
-    if (ptr == MAP_FAILED) {
-        return nullptr;
+mmap_fixed_address::mmap_fixed_address(size_t length) : m_length(length) {
```
yosefe, line 631: nice.

test/gtest/common/test_helpers.h
```diff
@@ -347,8 +347,15 @@ uint16_t get_port();
...
-
+class mmap_fixed_address {
+public:
+    mmap_fixed_address(size_t length);
+    ~mmap_fixed_address();
+    void* operator*() const { return m_ptr; }
+private:
+    void *m_ptr;
+    size_t m_length;
+};
```
yosefe, line 358: another space line after (2 in H file)

test/gtest/uct/test_mem.cc
```diff
@@ -154,10 +154,7 @@ UCS_TEST_P(test_mem, md_fixed) {
     uct_alloc_method_t       meth;
     size_t                   length      = 1;
     size_t                   n_success;
-    void*                    p_addr      = ucs::mmap_fixed_address(page_size * 2 * n_tryes);
-    if (p_addr == nullptr) {
-        UCS_TEST_ABORT("mmap failed to allocate memory region");
-    }
+    ucs::mmap_fixed_address  p_addr(page_size * n_tryes);
```
yosefe, line 157: can we revert to prev allocation size?

test/gtest/uct/test_mem.cc
```diff
@@ -230,10 +227,8 @@ UCS_TEST_P(test_mem, mmap_fixed) {
     uct_alloc_method_t      meth;
     size_t                  length      = 1;
     size_t                  n_success;
-    void*                   p_addr      = ucs::mmap_fixed_address(page_size * 2 * (n_tryes + 1));
-    if (p_addr == nullptr) {
-        UCS_TEST_ABORT("mmap failed to allocate memory region");
-    }
+    ucs::mmap_fixed_address p_addr(page_size * n_tryes);
```
yosefe, line 230: can we revert to prev allocation size?

yosefe (approved):

PR #9518, opened 2023-11-30
TEST/UCT/MD: Prevent rkey buffer overflow

ivankochin:
## What
Allocate `rkey_packed_size` rkey.

## Why ?
To prevent buffer overflow during `uct_md_mkey_pack_v2` reported by ASAN check.

### DIFF
--- a/test/gtest/uct/test_md.cc
+++ b/test/gtest/uct/test_md.cc
@@ -385,6 +385,7 @@ UCS_TEST_SKIP_COND_P(test_md, alloc,
     uct_md_h md_ref               = md();
     uct_alloc_method_t method     = UCT_ALLOC_METHOD_MD;
     unsigned num_alloc_failures   = 0;
+    std::vector<char> key(md_attr().rkey_packed_size);
     size_t size, orig_size;
     ucs_status_t status;
     void *address;
@@ -392,7 +393,6 @@ UCS_TEST_SKIP_COND_P(test_md, alloc,
     uct_allocated_memory_t mem;
     uct_mem_alloc_params_t alloc_params;
     uct_md_mkey_pack_params_t pack_params;
-    uint64_t key;
 
     alloc_params.field_mask = UCT_MEM_ALLOC_PARAM_FIELD_FLAGS |
                               UCT_MEM_ALLOC_PARAM_FIELD_ADDRESS |
@@ -444,8 +444,11 @@ UCS_TEST_SKIP_COND_P(test_md, alloc,
                 memset(address, 0xBB, size);
             }
 
-            status = uct_md_mkey_pack_v2(md(), mem.memh, &pack_params, &key);
-            ASSERT_UCS_OK(status);
+            if (md_attr().rkey_packed_size != 0) {
+                status = uct_md_mkey_pack_v2(md(), mem.memh, &pack_params,
+                                             key.data());
+                ASSERT_UCS_OK(status);
+            }
 
             uct_mem_free(&mem);
         }


yosefe (approved):
good catch!

brminich (dismissed):

src/uct/base/uct_md.c
```diff
@@ -303,8 +303,24 @@ ucs_status_t uct_config_modify(void *config, const char *name, const char *value
...
     if (ENABLE_PARAMS_CHECK) {
-        return ((md != NULL) && (memh != NULL) && (mkey_buffer != NULL)) ?
+        if (md == NULL) {
+            return UCS_ERR_INVALID_PARAM;
+        }
+
+        md_attr.field_mask = UCT_MD_ATTR_FIELD_RKEY_PACKED_SIZE;
+        status = uct_md_query_v2(md, &md_attr);
+        ucs_assertv(status == UCS_OK, "uct_md_query_v2 status: %s",
+                    ucs_status_string(status));
```
ivankochin, line 317:
```suggestion
        ucs_assertv_always(status == UCS_OK, "uct_md_query_v2 status: %s",
                           ucs_status_string(status));
```

brminich (dismissed):

src/uct/base/uct_md.c
```diff
@@ -303,8 +303,24 @@ ucs_status_t uct_config_modify(void *config, const char *name, const char *value
...
+        }
+
+        md_attr.field_mask = UCT_MD_ATTR_FIELD_RKEY_PACKED_SIZE;
+        status = uct_md_query_v2(md, &md_attr);
+        ucs_assertv_always(status == UCS_OK, "uct_md_query_v2 status: %s",
+                           ucs_status_string(status));
+
+        if (md_attr.rkey_packed_size == 0) {
+            return UCS_OK;
+        }
```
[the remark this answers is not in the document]
ivankochin: Removed as discussed

test/gtest/uct/test_md.cc
```diff
@@ -444,8 +444,11 @@ UCS_TEST_SKIP_COND_P(test_md, alloc,
                 memset(address, 0xBB, size);
             }
 
-            status = uct_md_mkey_pack_v2(md(), mem.memh, &pack_params, &key);
-            ASSERT_UCS_OK(status);
+            if (md_attr().flags & UCT_MD_FLAG_NEED_RKEY) {
```
yosefe, line 447: why needed?
maybe some MD needs to fix its rkey packed size?

ivankochin: This check is needed for cuda_copy MD.

yosefe: so what's the problem if rkey_packed_size=0?

yosefe: > why is checking rkey_size preferrable over checking cap?

According to UCT Api definition, uct_rkey_pack should work even if NEED_RKEY cap is not set.
We could also allocate 1-byte buffer in rkey_packed_size is 0 and do uct_rkey_pack to it

ivankochin: But it requires additional logic on the user side.

I think it is better not to check `(mkey_buffer != NULL)` condition inside `uct_md_mkey_pack_params_check` if `rkey_packed_size==0` as it was done in this commit: https://github.com/openucx/ucx/pull/9518/commits/660070f82c87b560f520e96df1af4a7138cc320f

WDYT?

brminich: imo, it looks weird to call function which is not supposed to be called

brminich: > > why is checking rkey_size preferrable over checking cap?
> 
> According to UCT Api definition, uct_rkey_pack should work even if NEED_RKEY cap is not set. We could also allocate 1-byte buffer in rkey_packed_size is 0 and do uct_rkey_pack to it

where is it stated explicitly? Why would we need to pack key if it is not really needed?

yosefe: The goal of gtest in is test the APIs as much as possible, even is something is "not needed"
so we should call and test uct_rkey_pack as much as possible
here I don't see a reason to skip calling it, we can pass non-NULL pointer or 1-byte buffer if rkey_packed_size==0

brminich: ok, but where is defined that `uct_rkey_pack` always works regardless of md caps? Seems  to be inconsistent with other caps/apis.

ivankochin: @brminich @yosefe maybe then it would be better to completely remove `(mkey_buffer != NULL)` check from `uct_md_mkey_pack_params_check`? 

Then the call wouldn't fail if someone pass NULL ptr in case of `rkey_packed_size==0`, but would fail in the internal `md->ops->mkey_pack()` call if NULL was passed but `rkey_packed_size!=0`.

brminich: imo, we should not allow to pass null buffers to the API which is expected to pack something there

yosefe: agree with @brminich to keep the NULL check

ivankochin, line 447:
```suggestion
            if (md_attr().rkey_packed_size != 0) {
```

yosefe (approved):

brminich (approved):

PR #9521, opened 2023-12-01
IO-DEMO: fix grep regexp in runner script

evgeny-leksikov:
## What
fix grep regexp in runner script

## Why ?
"\x00"-does not work for grep > 3.0 version.


evgeny-leksikov:
@maria-magazinnik could you pls review?

yosefe:
@evgeny-leksikov can we replace by something else? since we could have wrong match: server_1 could also match server_10 for example

evgeny-leksikov:
AFAIU 
`  -z, --null-data           a data line ends in 0 byte, not newline`
does what's needed

yosefe (approved):

PR #9525, opened 2023-12-04
UCP/AM: Allow AM handlers registration in any order

iyastreb:
## What
Fix the issue introduced recently in https://github.com/openucx/ucx/pull/9415 
The issue happens when registering AM handlers out of message ID order. For example, when client registers AM handlers in the order IDs [3, 4, 1], then only the last ID remains in the AM handlers array. Handlers [3, 4] are sliced away, which is manifested as
```
ucp_am.c:234  UCX  WARN  UCP Active Message was received with id : 3, but there is no registered callback for that id
ucp_am.c:234  UCX  WARN  UCP Active Message was received with id : 4, but there is no registered callback for that id
```

## Why ?
This happens because in current logic we unconditionally resize an array of AM handlers when adding new element to it. This works in most of the cases, when we register handlers ordered by message ID. However, if we register messages out of order, then the handlers array is truncated to the last registered ID.

Issue exists only in master, not exposed to v1.16 or below

### DIFF
--- a/src/ucp/core/ucp_am.c
+++ b/src/ucp/core/ucp_am.c
@@ -199,8 +199,12 @@ static ucs_status_t ucp_worker_set_am_handler_common(ucp_worker_h worker,
         return UCS_ERR_INVALID_PARAM;
     }
 
-    ucs_array_resize(&worker->am.cbs, id + 1, empty_am_handler,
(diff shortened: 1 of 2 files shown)


brminich:
can you please squash the commits?

brminich (approved):

PR #9529, opened 2023-12-05
UCT/CUDA/BASE: Do not use cuda device 0 if no context is active

ivankochin:
## What
Avoid using locality information of cuda device 0 if no active context detected.

## Why ?
Usage of 0 device breaks the wire compatibility because master now detect the distance to the CUDA device even if there is no active context. But older version cannot detect such distance and set it to default now which causes incompatibility.

Fixes the following issue (reproduces only on swx-rdmz-ucx-gpu-01): https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=72705&view=logs&j=fd287c5c-dcf2-52b8-cd0f-25b007d2717e&t=64aaf147-dfa5-5aa9-0220-0479128fc8a3

### DIFF
--- a/src/uct/cuda/base/cuda_iface.c
+++ b/src/uct/cuda/base/cuda_iface.c
@@ -36,23 +36,21 @@ uct_cuda_base_query_devices_common(
         uct_md_h md, uct_device_type_t dev_type,
         uct_tl_device_resource_t **tl_devices_p, unsigned *num_tl_devices_p)
 {
-    ucs_sys_device_t sys_device;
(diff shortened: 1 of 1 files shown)


brminich (approved):

brminich:
@rakhmets pls review

rakhmets (approved):

yosefe:
@rakhmets do you know why the original PR did not fail wire compatibility tests?

brminich:
> @rakhmets do you know why the original PR did not fail wire compatibility tests?

@ivankochin found that `gpu01` and `gpu02` have differenty configs. Tests pass on gpu02 and fail on gpu01 (or vice versa)

PR #9530, opened 2023-12-05
UCP: Report invalidation error if no rcache

ivankochin:
## What
Add assert that tell user about memory invalidation unavailability if no rcache is presented.

## Why ?
If no rcache was initialized, memory invalidation is unsupported. Without this checks, `ucp_memh_invalidate` would fail with segfault in this case.

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -2503,14 +2503,6 @@ ucs_status_t ucp_ep_config_init(ucp_worker_h worker, ucp_ep_config_t *config,
     if (status != UCS_OK) {
         goto err;
     }
-
-    if ((key->err_mode == UCP_ERR_HANDLING_MODE_PEER) &&
-        (context->rcache == NULL)) {
-        ucs_warn("worker %p: memory invalidation is requested but not supported"
-                 " since rcache isn't initialized. Invalidation case will cause"
-                 " runtime failure.", worker);
-    }
-
     if (config->key.flags & UCP_EP_CONFIG_KEY_FLAG_INTERMEDIATE) {
         short_am_cap_flag  = 0;
         short_tag_cap_flag = 0;
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -2503,6 +2503,7 @@ ucs_status_t ucp_ep_config_init(ucp_worker_h worker, ucp_ep_config_t *config,
     if (status != UCS_OK) {
         goto err;
     }
+
     if (config->key.flags & UCP_EP_CONFIG_KEY_FLAG_INTERMEDIATE) {
         short_am_cap_flag  = 0;
         short_tag_cap_flag = 0;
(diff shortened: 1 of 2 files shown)


src/ucp/core/ucp_ep.c
```diff
@@ -2504,6 +2504,13 @@ ucs_status_t ucp_ep_config_init(ucp_worker_h worker, ucp_ep_config_t *config,
         goto err;
     }
 
+    if ((key->err_mode == UCP_ERR_HANDLING_MODE_PEER) &&
+        (context->rcache == NULL)) {
+        ucs_warn("worker %p: memory invalidation is requested but not supported"
+                 " since rcache isn't initialized. Invalidation case will cause"
+                 " runtime failure.", worker);
```
brminich, line 2511: i'd also return error here (then can use `ucs_error` instead of `warn`)

yosefe: maybe we print this message in case of that runtime failure and not during initialization?

src/ucp/core/ucp_ep.c
```diff
@@ -2504,6 +2504,13 @@ ucs_status_t ucp_ep_config_init(ucp_worker_h worker, ucp_ep_config_t *config,
...
+    }
+
```
ivankochin, line 2513:
```suggestion

```

src/ucp/core/ucp_ep.c
```diff
@@ -2503,7 +2503,6 @@ ucs_status_t ucp_ep_config_init(ucp_worker_h worker, ucp_ep_config_t *config,
     if (status != UCS_OK) {
         goto err;
     }
-
     if (config->key.flags & UCP_EP_CONFIG_KEY_FLAG_INTERMEDIATE) {
```
ivankochin, line 2506:
```suggestion

    if (config->key.flags & UCP_EP_CONFIG_KEY_FLAG_INTERMEDIATE) {
```

src/ucp/core/ucp_ep.c
brminich: ok

src/ucp/core/ucp_mm.c
```diff
@@ -370,6 +370,8 @@ void ucp_memh_invalidate(ucp_context_h context, ucp_mem_h memh,
 
     ucs_assert(memh->parent == NULL);
     ucs_assert(!(memh->flags & UCP_MEMH_FLAG_IMPORTED));
+    ucs_assertv_always(context->rcache != NULL,
+                       "rcache is required for memory invalidation");
```
yosefe, line 374: BTW, what if we just deregister the memory (when rcache is not enabled)? Won't it allow using invalidation also without rcache?

yosefe:
@ivankochin is this still relevant?

ivankochin:
> @ivankochin is this still relevant?

Yes, I think it is still relevant, I just don't have enough time to check whether we can deregister memory in case of rcache absence. And as I understand, it is not high priority, so it can be postponed for some time.

gleon99:

ivankochin:
@gleon99 Privet, Lyonya! Please feel free to close any of my PRs for UCX if you think they’re no longer valuable. I’d love to come back to them someday, but I’m afraid I just don’t have the time right now. :(

PR #9536, opened 2023-12-07
TOOLS/PERF: Skeleton for DPU daemon

iyastreb:
## What
This is the first part of the "Support XGVMI in ucx perftest" story (original PR: https://github.com/openucx/ucx/pull/8767). 
The scope of this change-set is:
 - daemon skeleton implementation
 - appropriate config options for perftest
 - address existing comments from the original PR

## Why ?

### DIFF
--- a/src/tools/perf/api/libperf.h
+++ b/src/tools/perf/api/libperf.h
@@ -217,11 +217,14 @@ typedef struct ucx_perf_params {
     } uct;
 
     struct {
-        unsigned               nonblocking_mode; /* TBD */
-        ucp_perf_datatype_t    send_datatype;
-        ucp_perf_datatype_t    recv_datatype;
-        size_t                 am_hdr_size; /* UCP Active Message header size
-                                               (not included in message size) */
+        ucp_perf_datatype_t     send_datatype;
+        ucp_perf_datatype_t     recv_datatype;
+        size_t                  am_hdr_size; /* UCP Active Message header size
+                                                (not included in message size) */
+        struct sockaddr_storage dmn_local_addr;  /* IP and port of local daemon,
+                                                    used to offload communication */
+        struct sockaddr_storage dmn_remote_addr; /* IP and port of remote daemon,
+                                                    used to offload communication */
     } ucp;
 
 } ucx_perf_params_t;
--- a/src/tools/perf/perftest.c
+++ b/src/tools/perf/perftest.c
@@ -167,6 +167,8 @@ static int safe_recv(int sock, void *data, size_t size,
 
 ucs_status_t init_test_params(perftest_params_t *params)
 {
+    static const struct sockaddr_storage empty_addr = {};
+
     memset(params, 0, sizeof(*params));
     params->super.api               = UCX_PERF_API_LAST;
     params->super.command           = UCX_PERF_CMD_LAST;
@@ -194,6 +196,8 @@ ucs_status_t init_test_params(perftest_params_t *params)
     params->super.ucp.send_datatype = UCP_PERF_DATATYPE_CONTIG;
     params->super.ucp.recv_datatype = UCP_PERF_DATATYPE_CONTIG;
     params->super.ucp.am_hdr_size   = 0;
+    params->super.ucp.dmn_local_addr  = empty_addr;
+    params->super.ucp.dmn_remote_addr = empty_addr;
     strcpy(params->super.uct.dev_name, TL_RESOURCE_NAME_NONE);
     strcpy(params->super.uct.tl_name,  TL_RESOURCE_NAME_NONE);
 
--- a/src/tools/perf/perftest_daemon.c
+++ b/src/tools/perf/perftest_daemon.c
@@ -1,0 +1,251 @@
+/**
+ * Copyright (C) NVIDIA 2024.  ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#ifdef HAVE_CONFIG_H
+#  include "config.h"
+#endif
+
+#include "perftest.h"
+
+#include <ucp/api/ucp.h>
+#include <ucs/debug/log.h>
+#include <ucs/sys/sock.h>
+#include <ucs/sys/string.h>
+#include <ucs/type/serialize.h>
+#include <tools/perf/api/libperf.h>
+#include <string.h>
+#include <arpa/inet.h>
+#include <sys/socket.h>
+#include <unistd.h>
+#include <signal.h>
+
+typedef struct ucp_perf_daemon_context_t {
+    ucp_context_h                      context;
+    ucp_worker_h                       worker;
+    ucp_listener_h                     listener;
+    uint16_t                           port;
+    ucp_ep_h                           host_ep;
+    ucp_ep_h                           peer_ep;
+    struct sockaddr_storage            peer_address;
+    ucp_mem_h                          send_memh;
+    ucp_mem_h                          recv_memh;
+    void                               *rx_address;
+} ucp_perf_daemon_context_t;
+
+static volatile int terminated = 0;
+
+static void
+ucp_perf_daemon_ep_close(ucp_perf_daemon_context_t *ctx, ucp_ep_h ep)
+{
+    ucp_request_param_t param;
+    ucs_status_ptr_t close_req;
+    ucs_status_t status;
+
+    if (NULL == ep) {
+        return;
+    }
+
+    param.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS;
+    param.flags        = UCP_EP_CLOSE_FLAG_FORCE;
+    close_req          = ucp_ep_close_nbx(ep, &param);
+
+    if (UCS_PTR_IS_PTR(close_req)) {
+        do {
+            ucp_worker_progress(ctx->worker);
+            status = ucp_request_check_status(close_req);
+        } while (status == UCS_INPROGRESS);
+        ucp_request_free(close_req);
+    } else {
+        status = UCS_PTR_STATUS(close_req);
+    }
+
+    if (status != UCS_OK) {
+        ucs_error("daemon failed to close ep %p", ep);
+    }
+}
+
+static void ucp_perf_daemon_err_cb(void *arg, ucp_ep_h ep, ucs_status_t status)
+{
+    terminated = 1;
+}
+
+static void
+ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
+                                      void *arg)
+{
+    ucp_perf_daemon_context_t *ctx = arg;
+    ucp_ep_params_t ep_params;
+    ucs_status_t status;
+    ucp_ep_h ep;
+
+    ep_params.field_mask      = UCP_EP_PARAM_FIELD_ERR_HANDLER  |
+                                UCP_EP_PARAM_FIELD_CONN_REQUEST |
+                                UCP_EP_PARAM_FIELD_ERR_HANDLING_MODE;
+    ep_params.err_mode        = UCP_ERR_HANDLING_MODE_PEER;
+    ep_params.conn_request    = conn_request;
+    ep_params.err_handler.cb  = ucp_perf_daemon_err_cb;
+    ep_params.err_handler.arg = ctx;
+
+    status = ucp_ep_create(ctx->worker, &ep_params, &ep);
+    if (status != UCS_OK) {
+        ucs_error("failed to create an endpoint on the daemon: %s",
+                  ucs_status_string(status));
+    }
+}
+
+static void ucp_perf_daemon_cleanup(ucp_perf_daemon_context_t *ctx)
+{
+    if (NULL != ctx->send_memh) {
+        /* coverity[check_return] */
+        ucp_mem_unmap(ctx->context, ctx->send_memh);
+    }
+
+    if (NULL != ctx->recv_memh) {
+        /* coverity[check_return] */
+        ucp_mem_unmap(ctx->context, ctx->recv_memh);
+    }
+
+    ucp_perf_daemon_ep_close(ctx, ctx->peer_ep);
+    ucp_perf_daemon_ep_close(ctx, ctx->host_ep);
+    ucp_listener_destroy(ctx->listener);
+    ucp_worker_destroy(ctx->worker);
+    ucp_cleanup(ctx->context);
+}
+
+static ucs_status_t ucp_perf_daemon_init(ucp_perf_daemon_context_t *ctx)
+{
+    ucp_listener_params_t listen_params = {};
+    ucp_worker_params_t worker_params   = {};
+    ucp_params_t ucp_params;
+    struct sockaddr_in listen_addr;
+    ucp_config_t *config;
+    ucs_status_t status;
+
+    ucp_params.field_mask = UCP_PARAM_FIELD_FEATURES;
+    ucp_params.features   = UCP_FEATURE_AM | UCP_FEATURE_EXPORTED_MEMH;
+
+    status = ucp_config_read(NULL, NULL, &config);
+    if (status != UCS_OK) {
+        ucs_error("daemon failed to read UCP context: %s",
+                  ucs_status_string(status));
+        goto err;
+    }
+
+    status = ucp_init(&ucp_params, config, &ctx->context);
+    ucp_config_release(config);
+    if (status != UCS_OK) {
+        ucs_error("daemon failed to init UCP: %s", ucs_status_string(status));
+        goto err;
+    }
+
+    status = ucp_worker_create(ctx->context, &worker_params, &ctx->worker);
+    if (status != UCS_OK) {
+        ucs_error("failed to create worker: %s", ucs_status_string(status));
+        goto err_free_ctx;
+    }
+
+    listen_addr.sin_family      = AF_INET;
+    listen_addr.sin_addr.s_addr = INADDR_ANY;
+    listen_addr.sin_port        = htons(ctx->port);
+
+    listen_params.field_mask       = UCP_LISTENER_PARAM_FIELD_SOCK_ADDR |
+                                     UCP_LISTENER_PARAM_FIELD_CONN_HANDLER;
+    listen_params.sockaddr.addr    = (const struct sockaddr*)&listen_addr;
+    listen_params.sockaddr.addrlen = sizeof(listen_addr);
+    listen_params.conn_handler.cb  = ucp_perf_daemon_server_conn_handle_cb;
+    listen_params.conn_handler.arg = ctx;
+
+    status = ucp_listener_create(ctx->worker, &listen_params, &ctx->listener);
+    if (status != UCS_OK) {
+        ucs_error("failed to listen: %s", ucs_status_string(status));
+        goto err_free_worker;
+    }
+
+    return UCS_OK;
+
+err_free_worker:
+    ucp_worker_destroy(ctx->worker);
+err_free_ctx:
+    ucp_cleanup(ctx->context);
+err:
+    return status;
+}
+
+static void ucp_perf_daemon_signal_terminate_handler(int signo)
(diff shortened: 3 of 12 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -988,7 +988,7 @@ void ucp_memh_put_slow(ucp_context_h context, ucp_mem_h memh)
 
 ucs_status_t ucp_mem_unmap(ucp_context_h context, ucp_mem_h memh)
 {
-    if (ucp_memh_is_zero_length(memh)) {
+    if (NULL == memh || ucp_memh_is_zero_length(memh)) {
```
brminich, line 991:
```suggestion
    if ((NULL == memh) || ucp_memh_is_zero_length(memh)) {
```

BTW, is this situation possible? `(memh == NULL)`

src/tools/perf/perftest_params.c
```diff
@@ -253,6 +267,95 @@ static ucs_status_t parse_ucp_datatype_params(const char *opt_arg,
...
+
+    return set_sockaddr(ip_addr_str, port, addr);
+}
+
+static ucs_status_t daemon_addrs_init(ucx_perf_params_t *params)
+{
+    struct sockaddr_storage *local_addr  = &params->ucp.dmn_local_addr;
+    struct sockaddr_storage *remote_addr = &params->ucp.dmn_remote_addr;
+
+    /* Return if both daemon addressed are not configured */
```
brminich, line 316:
```suggestion
    /* Return if both daemon addresses are not configured */
```

src/tools/perf/perftest_daemon.c
```diff
@@ -0,0 +1,243 @@
...
+    ucp_mem_unmap(ctx->context, ctx->recv_memh);
+
+    ucp_perf_daemon_ep_close(ctx, ctx->peer_ep);
+    ucp_perf_daemon_ep_close(ctx, ctx->client_ep);
+    ucp_listener_destroy(ctx->listener);
+    ucp_worker_destroy(ctx->worker);
+    ucp_cleanup(ctx->context);
+}
+
+static int ucp_perf_daemon_init(ucp_perf_daemon_context_t *ctx)
```
brminich, line 112: better to make it return `ucs_status_t` to make it consistent with `ucp_perf_daemon_parse_cmd`

src/tools/perf/perftest_daemon.c
```diff
@@ -0,0 +1,243 @@
+/**
+ * Copyright (C) NVIDIA 2022.  ALL RIGHTS RESERVED.
```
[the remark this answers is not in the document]
iyastreb: ok

src/tools/perf/api/libperf.h
```diff
@@ -217,11 +216,13 @@ typedef struct ucx_perf_params {
     } uct;
 
     struct {
-        unsigned               nonblocking_mode; /* TBD */
-        ucp_perf_datatype_t    send_datatype;
-        ucp_perf_datatype_t    recv_datatype;
-        size_t                 am_hdr_size; /* UCP Active Message header size
-                                               (not included in message size) */
+        unsigned                nonblocking_mode; /* TBD */
```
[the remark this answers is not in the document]
iyastreb: sure

src/tools/perf/perftest.c
```diff
@@ -194,6 +196,8 @@ ucs_status_t init_test_params(perftest_params_t *params)
     params->super.ucp.send_datatype = UCP_PERF_DATATYPE_CONTIG;
     params->super.ucp.recv_datatype = UCP_PERF_DATATYPE_CONTIG;
     params->super.ucp.am_hdr_size   = 0;
+    params->super.ucp.dmn_local_addr  = empty_addr;
```
ivankochin, line 199: Minor: I don't think that `empty_addr` is required here and would prefer to use {}.

src/tools/perf/api/libperf.h
```diff
@@ -217,11 +216,12 @@ typedef struct ucx_perf_params {
...
-        ucp_perf_datatype_t    send_datatype;
-        ucp_perf_datatype_t    recv_datatype;
-        size_t                 am_hdr_size; /* UCP Active Message header size
-                                               (not included in message size) */
+        ucp_perf_datatype_t     send_datatype;
+        ucp_perf_datatype_t     recv_datatype;
+        size_t                  am_hdr_size; /* UCP Active Message header size
+                                                (not included in message size) */
+        struct sockaddr_storage dmn_local_addr;
+        struct sockaddr_storage dmn_remote_addr;
```
ivankochin, line 224: WDYT, maybe it would be better to add both `dmn_local/remote_addr` fields in the separate PR and make this PR contain `perftest_daemon` changes only. I mean that in the current patch we only introduce and define them to default value, so, from my point of view, it makes sense to introduce them in the patch, where we would have some logic connected to these variables.

If you disagree with the point described above, pls consider adding comment that says that these variables aren't used for now.

src/tools/perf/perftest_daemon.c
```diff
@@ -0,0 +1,243 @@
...
+#include <unistd.h>
+#include <signal.h>
+
+typedef struct ucp_perf_daemon_context_t {
+    ucp_context_h                      context;
+    ucp_worker_h                       worker;
+    ucp_listener_h                     listener;
+    uint16_t                           port;
+    ucp_ep_h                           client_ep;
+    ucp_ep_h                           peer_ep;
```
ivankochin, line 30: Do I understand correctly that `peer_ep` is EP connected to another daemon and `client_ep` is EP connected to host?

If so, I would like to propose to rename `client_ep` to `host_ep`, WDYT?

src/tools/perf/api/libperf.h
iyastreb: I would keep them in this PR for 2 reasons:
1. In this PR we introduce new command line arguments and parser exactly for these fields (`--daemon-local` and `--daemon-remote`). So it's not just for assigning default values)
IMO these fields are actually "used" for parsing, but not for communication yet
2. If I remove these addresses and related CLI and parsing, then the PR would be around 200 LOC. Committing in such a small chunks comes with a lot of overhead, so I would prefer to keep a reasonable PR size. So that overall we have ~3 commits for this feature

As for the comment - well, if you have strong opinion I can add it, but I hardly see the benefit.
I mean I will have to remember to remove it in the subsequent commit..
Please let me know

src/tools/perf/perftest.c
iyastreb: I would also prefer to use `{}` here, and that was my first attempt actually.
But C compiler complains that this syntax is not valid:
```
perftest.c:197:41: error: expected expression before ‘{’ token
     params->super.ucp.dmn_local_addr  = {};
```
So this syntax works for initialization, but not for assignment.
Therefore we have 2 options left: assign temporary struct, or memset. I prefer the first one

src/tools/perf/perftest_daemon.c
```diff
@@ -0,0 +1,243 @@
...
+err_free_ctx:
+    ucp_cleanup(ctx->context);
+err:
+    return status;
+}
+
+static void ucp_perf_daemon_signal_terminate_handler(int signo)
+{
+    char msg[64];
+    ssize_t ret __attribute__((unused));
```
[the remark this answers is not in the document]
iyastreb: Leftover from the original commit, will be removed

src/tools/perf/perftest_params.c
```diff
@@ -630,7 +740,7 @@ ucs_status_t parse_opts(struct perftest_context *ctx, int mpi_initialized,
         }
     }
 
-    return UCS_OK;
+    return daemon_addrs_init(&ctx->params.super);
```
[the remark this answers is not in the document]
iyastreb: Sure, makes sense

src/tools/perf/perftest_params.c
```diff
@@ -253,6 +267,95 @@ static ucs_status_t parse_ucp_datatype_params(const char *opt_arg,
     return UCS_OK;
 }
 
+static ucs_status_t set_sockaddr(const char *ip_addr_str, uint16_t port,
```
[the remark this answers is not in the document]
iyastreb: Yes, great idea, thanks!
I will use `ucs_sock_ipstr_to_sockaddr` for IP address and `ucs_sockaddr_set_port` for port

src/tools/perf/perftest_params.c
```diff
@@ -253,6 +267,95 @@ static ucs_status_t parse_ucp_datatype_params(const char *opt_arg,
...
+    }
+
+    ucs_error("invalid address '%s'", ip_addr_str);
+
+    return UCS_ERR_INVALID_PARAM;
+}
+
+static ucs_status_t parse_ip_addr_port(const char *str,
+                                       struct sockaddr_storage *addr,
+                                       uint16_t port)
```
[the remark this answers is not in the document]
iyastreb: Well, the idea was to make a reusable function, so that we can use it in other places.
This pattern `IP:port` seems to be quite common.
What do you think about moving this function to `ucs/sock` module?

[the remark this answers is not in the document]
iyastreb: Btw, for IPv6 addresses this string representation `IP:port` is ambiguous.
It seems we will have to use this format for IPv6: `[IPv6]:port`, which as [SO say](https://softwareengineering.stackexchange.com/questions/311413/how-should-we-represent-an-ipv6-address-with-port-number-in-text) is used in RFCs 5952, 3986

src/tools/perf/api/libperf.h
ivankochin: Oh, sorry I missed that part, my bad :)

src/tools/perf/perftest_daemon.c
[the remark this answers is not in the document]
iyastreb: Ok, that one is a bit more tricky.
When I remove it, it builds fine on my local machine, but [build fails of CI nodes](https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=73204&view=logs&j=8cc70074-80e3-506d-e80d-a542b02430e7&t=da02301d-00be-54d1-aa97-9bf0b310a3f6) with error:
```
perftest_daemon.c:176:10: error: ignoring return value of 'write', declared with attribute warn_unused_result [-Werror=unused-result]
     write(STDOUT_FILENO, msg, strlen(msg) + 1);
```
So I guess we need to keep it, it's done the same way as in io_demo.cc

src/ucp/core/ucp_mm.c
```diff
@@ -988,7 +988,7 @@ void ucp_memh_put_slow(ucp_context_h context, ucp_mem_h memh)
 
 ucs_status_t ucp_mem_unmap(ucp_context_h context, ucp_mem_h memh)
 {
-    if (ucp_memh_is_zero_length(memh)) {
+    if ((NULL == memh) || ucp_memh_is_zero_length(memh)) {
```
yosefe, line 991: why needed?

src/ucs/sys/sock.c
```diff
@@ -824,6 +824,44 @@ ucs_status_t ucs_sock_ipstr_to_sockaddr(const char *ip_str,
...
+{
+    /* Address is treated as IPv6 if it contains more than one colon */
+    const char *ch = strchr(ip_port_str, ':');
+    if (NULL != ch) {
+        return (NULL != strchr(++ch, ':'));
+    }
+    return 0;
+}
+
+ucs_status_t ucs_sock_ipportstr_to_sockaddr(const char *ip_port_str,
```
yosefe, line 837: can't we use ucs_sock_ipstr_to_sockaddr?

src/tools/perf/perftest_daemon.c
```diff
@@ -0,0 +1,243 @@
...
+        usage(&ctx, ucs_basename(argv[0]));
+        return EXIT_FAILURE;
+    }
+
+    new_sigaction.sa_handler = ucp_perf_daemon_signal_terminate_handler;
+    new_sigaction.sa_flags   = 0;
+    sigemptyset(&new_sigaction.sa_mask);
+
+    sigaction(SIGINT, &new_sigaction, NULL);
+    sigaction(SIGHUP, &new_sigaction, NULL);
```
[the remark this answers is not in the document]
iyastreb: To gracefully terminate the process when parent terminal is closed.

Well, my understanding is that this code was copy-pasted from a similar [VFS daemon](https://github.com/openucx/ucx/blob/master/src/tools/vfs/vfs_server.c#L319): 
that also registers handlers for all these signals.

src/tools/perf/perftest_daemon.c
```diff
@@ -0,0 +1,243 @@
...
+    sigaction(SIGTERM, &new_sigaction, NULL);
+
+    if (ucp_perf_daemon_init(&ctx) != UCS_OK) {
+        ucs_fatal("failed to initalize");
```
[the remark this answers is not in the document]
iyastreb: Ok, changed to:
```
        ucs_error("failed to initalize");
        return EXIT_FAILURE;
```

src/ucs/sys/sock.h
```diff
@@ -435,6 +435,27 @@ ucs_status_t ucs_sock_ipstr_to_sockaddr(const char *ip_str,
...
+ * [IPv6]:port  e.g. [::1]:1339
+ *
+ * Port is optional parameter in string, if not set, default_port is used
+ *
+ * @param [in]  ip_port_str  A string to take IP address and port from.
+ * @param [in]  default_port Default port to use if string doesn't have it.
+ * @param [out] sa_storage   sockaddr storage filled with the IP address,
+ *                           address family and port.
+ *
+ * @return UCS_OK if @a ip_str has a valid IP address or UCS_ERR_INVALID_ADDR
```
ivankochin, line 451:
```suggestion
 * @return UCS_OK if @a ip_port_str has a valid IP address or UCS_ERR_INVALID_ADDR
```

src/ucs/sys/sock.c
ivankochin: BTW, can we rely on `getaddrinfo` to handle both port and IP address?

src/tools/perf/perftest_params.c
```diff
@@ -253,6 +267,54 @@ static ucs_status_t parse_ucp_datatype_params(const char *opt_arg,
     return UCS_OK;
 }
 
+static ucs_status_t parse_daemon_addrs(ucx_perf_params_t *params)
```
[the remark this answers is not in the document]
ivankochin: I think that we should remove it to something more common like `verify_init_params` since it checks not only daemon addresses.

src/ucs/sys/sock.c
yosefe: won't it be simpler to just split by ':' and use last argument as port number?
why need the brackets notation with IPv6?

yosefe: AFAIR getaddrinfo can also handle hostnames

src/ucp/core/ucp_mm.c
iyastreb: ok, done

src/ucs/sys/sock.h
iyastreb: ok

src/tools/perf/perftest_daemon.c
```diff
@@ -0,0 +1,243 @@
+/**
+ * Copyright (C) NVIDIA 2023.  ALL RIGHTS RESERVED.
```
brminich, line 2: now `2024` )
here and in all other files

test/gtest/ucs/test_sock.cc
```diff
@@ -394,6 +410,44 @@ UCS_TEST_F(test_socket, str_sockaddr_str) {
...
+    check_ip_port("192.168.100.100:60000", 1234, UCS_OK, "192.168.100.100:60000");
+
+    /* Check wrong IPv4:port addresses */
+    {
+        socket_err_exp_str = "invalid address";
+        scoped_log_handler log_handler(socket_error_handler);
+
+        check_ip_port("", 0, UCS_ERR_INVALID_ADDR, "");
+        check_ip_port(":9999", 0, UCS_ERR_INVALID_ADDR, "");
+        check_ip_port("1.2.3.", 0, UCS_ERR_INVALID_ADDR, "");
```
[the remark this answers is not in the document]
iyastreb: ok, will add

src/ucs/sys/sock.c
iyastreb: Yes, it would me much simpler if we could just split by colon.
But there are 2 reasons why we can't do this:
1) IPv6 itself may contain colons (`::1`, `::`)
2) Ok, can we split just the **last** colon and use it as a port?
No, because port argument is optional. It may or may not be present in the config

I'm checking getaddrinfo

brminich (dismissed):
👍 besides a couple of comments from Yossi

src/ucs/sys/sock.c
iyastreb: My understanding is that `getaddrinfo` would not be really helpful here, because it expects to receive a port as a separate argument. But we don't have a port as a separate string, that's why we have parsing - that's the whole thing here to split IP and port from the same string

The complexity of this function comes from the facts that:
1. we use colon as a separator, which interferes with IPv6 string representation
2. IP can be either IPv4 or IPv6
3. port arg is optional

If we really want to make it simpler, then we can use some other symbol as an IP-port separator: (`/`, `;`, `-`, `%`). E.g. `192.168.1.1/13311`
Personally I don't really like it

ivankochin (approved):
LGTM!

src/ucs/sys/sock.c
yosefe: Ok, according to https://en.wikipedia.org/wiki/IPv6_address and https://datatracker.ietf.org/doc/html/rfc3986, we should use square brackets indeed.
Can we add support for getaddrinfo? maybe even in another PR.

src/ucs/sys/sock.h
```diff
@@ -448,14 +448,26 @@ ucs_status_t ucs_sock_ipstr_to_sockaddr(const char *ip_str,
...
+/**
+ * Extract port from a given string and validate it.
+ *
+ * @param [in]  port_str A string to take port from.
+ * @param [out] port     Variable to store parsed port.
+ *
+ * @return UCS_OK if @a port_str has a valid port,
+ *         UCS_ERR_INVALID_ADDR otherwise.
+ */
+ucs_status_t ucs_port_from_string(const char *port_str, uint16_t *port);
```
yosefe, line 468: ucs_sock_port_from_string

src/ucs/sys/sock.c
iyastreb: Yes, let's consider support for `getaddrinfo` in another PR, this one already grew up beyond 500LOC

src/ucs/sys/sock.h
iyastreb: sure

test/gtest/ucs/test_sock.cc
```diff
@@ -421,8 +421,19 @@ UCS_TEST_F(test_socket, sock_ipportstr_to_sockaddr) {
...
+        check_ip_port("1.1.1.256:9999", 0, UCS_ERR_INVALID_ADDR, "");
+    }
+    {
+        socket_err_exp_str = "invalid port";
+        scoped_log_handler log_handler(socket_error_handler);
+
+        check_ip_port("1.1.1.1:-1", 0, UCS_ERR_INVALID_ADDR, "");
+        check_ip_port("1.1.1.1:65536", 0, UCS_ERR_INVALID_ADDR, "");
+        check_ip_port("1.1.1.1:1111111", 0, UCS_ERR_INVALID_ADDR, "");
+        check_ip_port("1.1.1.1:randomstring", 0, UCS_ERR_INVALID_ADDR, "");
```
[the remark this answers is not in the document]
iyastreb: ok

yosefe (approved):

PR #9538, opened 2023-12-07
AZP/RELEASE: Fix JUCX with ARM build

Alexey-Rivkin:
## What
Fix JUCX package publishing, so it will include support for ARM.

## How ?
1. Copy both libraries into the resources folder to match the resources section in pom.xml.
2. Copy specific files, not a wildcard, so `make `will fail if one of them is missing.
3. Added a validation target to check the final Jar before publishing.

### DIFF
--- a/bindings/java/src/main/native/Makefile.am
+++ b/bindings/java/src/main/native/Makefile.am
@@ -103,6 +103,7 @@ set-version:
 repack-jar: $(jarfile)
 	$(RM) -r $(java_build_dir)/repack_dir
 	unzip -o $(jarfile) -d $(java_build_dir)/repack_dir
+	\cp -v $(topdir)/libjucx_*.so/libjucx_*.so $(topdir)/bindings/java/resources
 	\cp -v $(topdir)/libjucx_*.so/libjucx_*.so $(java_build_dir)/repack_dir
 	jar -cf $(jarfile) -C $(java_build_dir)/repack_dir .
 	jar tf $(jarfile)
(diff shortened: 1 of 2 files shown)


bindings/java/src/main/native/Makefile.am
```diff
@@ -103,22 +103,29 @@ set-version:
...
 	@make repack-jar
 
+check-jar:
+	@test `jar tf $(jarfile) | grep -c libjucx_` -eq 2
+
 # Publish JUCX jar to maven central
 publish-snapshot:
 	@make set-version JUCX_VERSION=@VERSION@-SNAPSHOT
+	@make repack-jar
+	@make check-jar
```
yosefe, line 122: don't we want to publish only the release?
@petro-rudenko WDYT?

bindings/java/src/main/native/Makefile.am
```diff
@@ -103,22 +103,32 @@ set-version:
 repack-jar: $(jarfile)
 	$(RM) -r $(java_build_dir)/repack_dir
 	unzip -o $(jarfile) -d $(java_build_dir)/repack_dir
-	\cp -v $(topdir)/libjucx_*.so/libjucx_*.so $(java_build_dir)/repack_dir
+	\cp $(topdir)/libjucx_amd64.so/libjucx_amd64.so $(topdir)/bindings/java/resources
```
yosefe, line 106: 1. why need to copy both to resources and repack_dir?
2. can we copy with wildcard? anyway we have test step to make sure the files are present

Alexey-Rivkin: 1. The resources dir is used by Maven while rebuilding Jars. The repack_dir is used while repacking. This is a temporary solution before we externalize both repack and check steps to CI.
2. Changed to wildcards.

yosefe (approved):
pls squash

amaslenn (approved):

Alexey-Rivkin:
@yosefe Should we port this fix to v1.16.x branch?

yosefe:
> @yosefe Should we port this fix to v1.16.x branch?

@Alexey-Rivkin yes, please

PR #9539, opened 2023-12-07
AZP/RELEASE: Fix JUCX with ARM build-v1.15.x

Alexey-Rivkin:
## What
Fix JUCX package publishing, so it will include support for ARM.

## How ?
1. Copy both libraries into the resources folder to match the resources section in pom.xml.
2. Copy specific files, not a wildcard, so `make `will fail if one of them is missing.
3. Added a validation target to check the final Jar before publishing.


Alexey-Rivkin:
@yosefe  Should I create the v1.15.0.1 release to publish the JUCX with ARM support?

amaslenn (approved):

yosefe:
@Alexey-Rivkin would it help if we create 1.16.0-rc1 release?
Do we need to merge any fixes to 1.16.x branch?

PR #9544, opened 2023-12-11
AZP/RELEASE: Do not publish JUCX for RCs

Alexey-Rivkin:
## What
Skip the JUCX publish job for RC tags.


Alexey-Rivkin:
Closing due to change in requirements: we do publish JUCX for RCs now.

PR #9084, opened 2023-05-17
UCP/GTEST: Enable protov2 by default (wire-compat is not addressed)

brminich:
## What
Enable protov2 by default

## Why ?
Just enable proto v2 by default and update corresponding test.
All wire-compat issues will be updated in other PRs

PR #9545, opened 2023-12-12
CI fixes

yosefe:
## Why
1. Fix assertion in rcx/test_ucp_tag_mem_type.reuse_buffers_mrail/15. The receive buffer could have been registered as part of tag offload and moved from recv iterator to rndv send request iterator.
2. Fix the condition that enable testing protov2 specific features: need to check UCP configuration rather than test variant, since protov2 can be disabled by env var UCX_PROTO_ENABLE=n

cc @shasson5

### DIFF
--- a/test/gtest/ucp/test_ucp_atomic.cc
+++ b/test/gtest/ucp/test_ucp_atomic.cc
@@ -17,8 +17,7 @@ class test_ucp_atomic : public test_ucp_memheap {
 public:
     /* Test variants */
     enum {
-        ATOMIC_MODE   = UCS_MASK(2),
-        DISABLE_PROTO = UCS_BIT(2)
+        ATOMIC_MODE = UCS_MASK(2)
     };
 
     static void get_test_variants(std::vector<ucp_test_variant>& variants,
@@ -34,9 +33,6 @@ public:
     static void get_test_variants(std::vector<ucp_test_variant>& variants)
     {
         get_test_variants(variants, 0, "");
-        if (!RUNNING_ON_VALGRIND) {
-            get_test_variants(variants, DISABLE_PROTO, "/proto_v1");
-        }
     }
 
     struct send_func_data {
@@ -134,11 +130,6 @@ protected:
                         (atomic_mode == UCP_ATOMIC_MODE_GUESS)  ? "guess" :
                         "";
         modify_config("ATOMIC_MODE", atomic_mode_cfg);
-
-        if (get_variant_value() & DISABLE_PROTO) {
-            modify_config("PROTO_ENABLE", "n");
-        }
-
         test_ucp_memheap::init();
     }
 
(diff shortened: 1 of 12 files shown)


test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -182,7 +182,7 @@ class test_ucp_rma : public test_ucp_memheap {
     }
 
     bool disable_proto() {
-        return !m_ucp_config->ctx.proto_enable;
+        return !is_proto_enabled();
```
rakhmets, line 185: I guess it's a bug here (from this PR https://github.com/openucx/ucx/pull/9084).
This method should check test variant value, and the config value should be set based on return value.
Thus, it should be:
 ```C
 bool disable_proto() const
{
        return get_variant_value() & DISABLE_PROTO;
}
```

And then `!is_proto_enabled()` should be used in this file at these lines: 295 and 314.

File: test/gtest/ucp/test_ucp_mmap.cc [hunk 2]
```diff
@@ -424,7 +424,7 @@ void test_ucp_mmap::test_rkey_proto(ucp_mem_h memh)
     ASSERT_UCS_OK(status);
 
     /* Check rkey configuration */
-    if (!disable_proto() && m_ucp_config->ctx.proto_enable) {
+    if (!disable_proto() && is_proto_enabled()) {
```
rakhmets, line 427: Guess that should be enough:
```suggestion
    if (is_proto_enabled()) {
```

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -9,6 +9,7 @@
 
 extern "C" {
 #include <ucp/core/ucp_types.h> /* for atomic mode */
+#include <ucp/core/ucp_context.h>
```
rakhmets, line 12: This line can be removed.

test/gtest/ucp/test_ucp_rma.cc
brminich: config maybe set externally (e.g. in proto_disable roce tests in CI)

rakhmets: I see.
But we don't use `DISABLE_PROTO` variant value in these tests.

test/gtest/ucp/test_ucp_rma.cc
```diff
@@ -182,7 +182,7 @@ class test_ucp_rma : public test_ucp_memheap {
     }
 
     bool disable_proto() {
-        return !m_ucp_config->ctx.proto_enable;
+        return get_variant_value() & DISABLE_PROTO;
```
brminich, line 185: seems it is only used in `init` method now.
maybe it's better for check for `get_variant_value() & DISABLE_PROTO` there directly?

brminich: @yosefe, do you think it's worth to cnosider?
seems such method is removed in other places

yosefe: done

rakhmets (dismissed):

PR #9546, opened 2023-12-12
UCP/RMA: Do not flush UCT EP if UCT iface is not active.

rakhmets:
## What
Added condition to flush UCT endpoint only if corresponding UCT interface was activated.
The condition is only relevant if new protocol infrastructure was enabled.

## Why ?


src/ucp/rma/flush.c
```diff
@@ -63,6 +64,28 @@ static int ucp_ep_flush_is_completed(ucp_request_t *req)
...
+        return 1;
+    }
+
+    if (ucp_wireup_ep_test(ucp_ep_get_lane(ep, lane))) {
+        return 1;
+    }
+
+    wiface = ucp_worker_iface(worker, ucp_ep_get_rsc_index(ep, lane));
+    return (wiface == NULL) ||
+           (wiface->flags & UCP_WORKER_IFACE_FLAG_KEEP_ACTIVE);
```
yosefe, line 86: 1. maybe instead of proto_enable and IFACE_FLAG_KEEP_ACTIVE, we could check wiface->activate_count directly?
2. how could wiface be NULL here? since we have ep on it..

src/ucp/rma/flush.c
```diff
@@ -63,6 +64,28 @@ static int ucp_ep_flush_is_completed(ucp_request_t *req)
...
+static int uct_ep_flush_is_needed(ucp_ep_h ep, ucp_lane_index_t lane)
+{
+    ucp_worker_h worker = ep->worker;
+    ucp_worker_iface_t *wiface;
+
+    if (!worker->context->config.ext.proto_enable) {
+        return 1;
+    }
+
+    if (ep->flags & UCP_EP_FLAG_INTERNAL) {
```
yosefe, line 76: why need different handling for internal ep?

src/ucp/rma/flush.c
```diff
@@ -63,6 +64,28 @@ static int ucp_ep_flush_is_completed(ucp_request_t *req)
     return (req->send.state.uct_comp.count == 0) && req->send.flush.sw_done;
 }
 
+static int uct_ep_flush_is_needed(ucp_ep_h ep, ucp_lane_index_t lane)
```
iyastreb, line 67: can we pass ep by const?

rakhmets: Added `const` keyword.

src/ucp/rma/flush.c
rakhmets: We always activate wiface for mem type endpoints.

tvegas1:
Seems the perftest MAD failure could be related to PR.

gleon99:
Hi @rakhmets - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #9548, opened 2023-12-15
UCT/IB: Add FLID based routing support

brminich:
## What
Use FLID as a destination lid when it is present in site-local gid

## Why ?
Support IB routing


yosefe:
replaced by #9551

PR #9549, opened 2023-12-15
UCT/CUDA-IPC: Ignore ka failure during ep creation

brminich:
## What
Error handling is not always needed by the UCT user (e.g. MPI apps). Ignore ka initialization failure, which enables cuda-ipc multinode for MPI flows.


brminich:
@Akshay-Venkatesh

yosefe (approved):

PR #9555, opened 2023-12-19
UCS/CONFIG: Key-value config parser

iyastreb:
## What
Advanced key-value parser for UCX config.

Usage example:
```
  {"TEMP", "20", "Temperature", 0,
    UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_UINT,
        {"front", "front temperature", ucs_offsetof(car_opts_t, temp_front)},
        {"rear",  "rear temperature",  ucs_offsetof(car_opts_t, temp_rear)},
        {NULL}
    )},
```

## Why ?
There are several use cases when config environment variables may contain multiple values (e.g. `RNDV_FRAG_ALLOC_COUNT= host:128,cuda:128`). However there are several drawbacks with current implementation: storing intermediate strings, 2-phase processing, syntax is not enforced.
Advanced key-value config parser aims to solve these issues.

## How ?
[Design considerations document](https://nvidia.sharepoint.com/:p:/r/sites/NBU-UCX/_layouts/15/Doc.aspx?sourcedoc=%7B232d86f5-c9e1-4a3c-8e40-7b6a6f3cf161%7D&action=edit&wdPreviousSession=61b5f70b-d600-ddd5-ef57-aaed06ac60c2&wdOrigin=TEAMS-MAGLEV.p2p_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1702363341203&web=1).

Some considerations:
1. Outer config has an offset which is supposed to be zero. One idea was to hide it with macro and enforce to be 0, e.g.:
```
UCS_CONFIG_KEY_VALUE("TEMP", "20", "Temperature", UCS_CONFIG_TYPE_UINT,
        {"front", "front temperature", ucs_offsetof(car_opts_t, temp_front)},
        {"rear",  "rear temperature",  ucs_offsetof(car_opts_t, temp_rear)},
        {NULL}
),
```
For now I decided not implement it, because: a) config becomes less intuitive b) common offset may be useful? c) IMO macro syntax is overkill here

2. Default value is set only in outer config, which may contain both generic value and/or fine-grained configs
3. **Partial setting is NOT allowed**. E.g. `TEMP=rear:10` is not accepted, value must initialize all keys
4. Config is parsed into a temporary buffer and applied exactly once per key. This is done to avoid re-allocation for dynamic values.

### DIFF
--- a/src/ucs/config/parser.c
+++ b/src/ucs/config/parser.c
@@ -1026,8 +1026,8 @@ int ucs_config_sscanf_table(const char *buf, void *dest, const void *arg)
         name  = strtok_r(token, "=", &saveptr2);
         value = strtok_r(NULL,  "=", &saveptr2);
         if (name == NULL || value == NULL) {
-            ucs_free(tokens);
             ucs_error("Could not parse list of values in '%s' (token: '%s')", buf, token);
+            ucs_free(tokens);
             return 0;
         }
 
@@ -1066,7 +1066,190 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
     snprintf(buf, max, "Table");
 }
 
+/**
+ * Return number of allowed keys in the given config
+ */
+static size_t ucs_config_key_count(const ucs_config_key_field_t *cfg_key)
+{
+    size_t count = 0;
+    for (; cfg_key->name; ++cfg_key, ++count);
+    return count;
+}
+
+/**
+ * Find index of the given key in the config
+ */
+static int ucs_config_key_find(const ucs_config_key_field_t *cfg_key,
+                               const char *key)
+{
+    int idx;
+
+    for (idx = 0; cfg_key->name; ++cfg_key, ++idx) {
+        if (!strcmp(cfg_key->name, key)) {
+            return idx;
+        }
+    }
+    return -1;
+}
+
+int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    size_t buf_len                            = strlen(buf) + 1;
+    const char *TOKEN_DELIM                   = ",";
+    const char *VALUE_DELIM                   = ":";
+    const char *default_value                 = NULL;
+    const char *key, *value, **values;
+    char *token, *tokens;
+    size_t key_count;
+    int idx;
+
+    /* values is the array of key-specific settings */
+    key_count = ucs_config_key_count(param->keys);
+    ucs_assert(key_count > 0);
+    values = ucs_alloca(key_count * sizeof(*values));
+    memset(values, 0, key_count * sizeof(*values));
+
+    /* Writable copy of the input buffer on stack */
+    tokens = ucs_alloca(buf_len);
+    ucs_strncpy_safe(tokens, buf, buf_len);
+
+    tokens = ucs_string_split(tokens, TOKEN_DELIM, 1, &token);
+    while (NULL != token) {
+        ucs_string_split(token, VALUE_DELIM, 2, &key, &value);
+
+        if (NULL == value) {
+            /* No value means key is the default value for all */
+            default_value = key;
+        } else {
+            /* Specific key config is present */
+            if ((idx = ucs_config_key_find(param->keys, key)) < 0) {
+                ucs_error("key '%s' is not supported", key);
+                return 0;
+            }
+
+            values[idx] = value;
+        }
+
+        tokens = ucs_string_split(tokens, TOKEN_DELIM, 1, &token);
+    }
+
+    /* Validate and apply settings, revert in case of errors */
+    for (idx = 0; idx < key_count; ++idx) {
+        value = (values[idx] == NULL) ? default_value : values[idx];
+        if (NULL == value) {
+            ucs_error("no value configured for key '%s'", param->keys[idx].name);
+        } else if ((param->parser.read(value,
+                                       UCS_PTR_BYTE_OFFSET(dest, param->keys[idx].offset),
+                                       param->parser.arg) == 1)) {
+            continue;
+        }
+
+        /* Revert already made settings */
+        while (idx-- > 0) {
+            param->parser.release(UCS_PTR_BYTE_OFFSET(dest, param->keys[idx].offset),
+                                  param->parser.arg);
+        }
+
+        return 0;
+    }
+
+    return 1;
+}
+
+int ucs_config_sprintf_key_value(char *buf, size_t max, const void *src,
+                                 const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    char value_buf[256]                       = "";
+    const ucs_config_key_field_t *cfg_key;
+    UCS_STRING_BUFFER_FIXED(strb, buf, max);
+
+    for (cfg_key = param->keys; NULL != cfg_key->name; ++cfg_key) {
+        if (!param->parser.write(value_buf, sizeof(value_buf),
+                                 UCS_PTR_BYTE_OFFSET(src, cfg_key->offset),
+                                 param->parser.arg)) {
+            /* Error is logged by parser impl */
+            return 0;
+        }
+
+        ucs_string_buffer_appendf(&strb, "%s:%s,", cfg_key->name, value_buf);
+    }
+
+    ucs_string_buffer_rtrim(&strb, ",");
+    return 1;
+}
+
+ucs_status_t ucs_config_clone_key_value(const void *src, void *dest,
+                                        const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    ucs_status_t status                       = UCS_OK;
+    const ucs_config_key_field_t *cfg_key;
+
+    for (cfg_key = param->keys; NULL != cfg_key->name; ++cfg_key) {
+        status = param->parser.clone(UCS_PTR_BYTE_OFFSET(src, cfg_key->offset),
+                                     UCS_PTR_BYTE_OFFSET(dest, cfg_key->offset),
+                                     param->parser.arg);
+        if (UCS_OK != status) {
+            /* Error is logged by parser impl */
+            break;
+        }
+    }
+    return status;
+}
+
+void ucs_config_release_key_value(void *ptr, const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    const ucs_config_key_field_t *cfg_key;
+
+    for (cfg_key = param->keys; NULL != cfg_key->name; ++cfg_key) {
+        param->parser.release(UCS_PTR_BYTE_OFFSET(ptr, cfg_key->offset),
+                              param->parser.arg);
+    }
+}
+
+void ucs_config_help_key_value(char *buf, size_t max, const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    char syntax_buf[256]                      = "";
+    const ucs_config_key_field_t *cfg_key;
+    UCS_STRING_BUFFER_FIXED(strb, buf, max);
+
+    ucs_string_buffer_appendf(&strb, "comma-separated list of value or key:value"
+                              " pairs, where key is one of [");
+
+    for (cfg_key = param->keys; NULL != cfg_key->name; ++cfg_key) {
+        ucs_string_buffer_appendf(&strb, "%s,", cfg_key->name);
+    }
+
+    ucs_string_buffer_rtrim(&strb, ",");
+    ucs_string_buffer_appendf(&strb, "] and value is: ");
+
+    param->parser.help(syntax_buf, sizeof(syntax_buf), param->parser.arg);
+    ucs_string_buffer_appendf(&strb, "%s. A value without a key is the default.",
+                              syntax_buf);
+}
+
+void ucs_config_doc_key_value(ucs_string_buffer_t *strb, const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    const ucs_config_key_field_t *cfg_key;
+
+    for (cfg_key = param->keys; NULL != cfg_key->name; ++cfg_key) {
+        ucs_string_buffer_appendf(strb, " %-*s- %s\n",
+                                  UCS_CONFIG_PARSER_DOCSTR_WIDTH,
+                                  cfg_key->name, cfg_key->doc);
+    }
+    ucs_string_buffer_rtrim(strb, "\n");
+}
+
 void ucs_config_release_nop(void *ptr, const void *arg)
+{
+}
+
+void ucs_config_doc_nop(ucs_string_buffer_t *strb, const void *arg)
 {
 }
 
@@ -1090,6 +1273,21 @@ static inline int ucs_config_is_table_field(const ucs_config_field_t *field)
     return (field->parser.read == ucs_config_sscanf_table);
 }
 
+/**
+ * Allocates multi-line documentation text string, caller is responsible for
+ * freeing the memory.
+ */
+static char *ucs_config_get_doc(const ucs_config_field_t *field)
+{
+    ucs_string_buffer_t strb;
+    ucs_string_buffer_init(&strb);
+
+    ucs_string_buffer_appendf(&strb, "%s\n", field->doc);
+    field->parser.doc(&strb, field->parser.arg);
+
+    return ucs_string_buffer_extract_mem(&strb);
+}
+
 static void ucs_config_print_doc_line_by_line(const ucs_config_field_t *field,
                                               void (*cb)(int num, const char *line, void *arg),
                                               void *arg)
@@ -1097,7 +1295,7 @@ static void ucs_config_print_doc_line_by_line(const ucs_config_field_t *field,
     char *doc, *line, *p;
     int num;
 
-    line = doc = ucs_strdup(field->doc, "config_doc");
+    line = doc = ucs_config_get_doc(field);
     p = strchr(line, '\n');
     num = 0;
     while (p != NULL) {
@@ -1112,7 +1310,8 @@ static void ucs_config_print_doc_line_by_line(const ucs_config_field_t *field,
 }
 
 static ucs_status_t
-ucs_config_parser_parse_field(ucs_config_field_t *field, const char *value, void *var)
+ucs_config_parser_parse_field(const ucs_config_field_t *field,
+                              const char *value, void *var)
 {
     char syntax_buf[256];
     int ret;
--- a/src/ucs/config/parser.h
+++ b/src/ucs/config/parser.h
@@ -46,6 +46,7 @@ typedef struct ucs_config_parser {
     ucs_status_t             (*clone)(const void *src, void *dest, const void *arg);
     void                     (*release)(void *ptr, const void *arg);
     void                     (*help)(char *buf, size_t max, const void *arg);
+    void                     (*doc)(ucs_string_buffer_t *strb, const void *arg);
     const void               *arg;
 } ucs_config_parser_t;
 
@@ -63,6 +64,23 @@ typedef struct ucs_config_field {
     size_t                   offset;
     ucs_config_parser_t      parser;
 } ucs_config_field_t;
+
+
+typedef struct {
+    const char               *name;   /* Key name */
+    const char               *doc;    /* Documentation */
+    size_t                   offset;  /* Storage offset */
+} ucs_config_key_field_t;
+
+
+/**
+ * Describes structure that is passed to key-value parser functions as an
+ * argument.
+ */
+typedef struct {
+    ucs_config_parser_t          parser;  /* Parser impl */
+    const ucs_config_key_field_t *keys;   /* Array of allowed key fields */
+} ucs_config_key_value_param_t;
 
 
 typedef struct ucs_config_cached_key {
@@ -259,9 +277,18 @@ ucs_status_t ucs_config_clone_table(const void *src, void *dest, const void *arg
 void ucs_config_release_table(void *ptr, const void *arg);
 void ucs_config_help_table(char *buf, size_t max, const void *arg);
 
+/* Key-value parser functions */
+int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg);
+int ucs_config_sprintf_key_value(char *buf, size_t max, const void *src, const void *arg);
+ucs_status_t ucs_config_clone_key_value(const void *src, void *dest, const void *arg);
+void ucs_config_release_key_value(void *ptr, const void *arg);
+void ucs_config_help_key_value(char *buf, size_t max, const void *arg);
+void ucs_config_doc_key_value(ucs_string_buffer_t *strb, const void *arg);
+
 ucs_status_t ucs_config_clone_log_comp(const void *src, void *dst, const void *arg);
 
 void ucs_config_release_nop(void *ptr, const void *arg);
+void ucs_config_doc_nop(ucs_string_buffer_t *strb, const void *arg);
 void ucs_config_help_generic(char *buf, size_t max, const void *arg);
 
 #define UCS_CONFIG_DEPRECATED_FIELD_OFFSET SIZE_MAX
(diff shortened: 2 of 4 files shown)


src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+    /* values is a temporary array value storage for the keys.
+     * In values array we remember the last applicable setting for each key.
+     * We need it to perform actual setting exactly once per each key */
+    key_count = ucs_config_key_value_count(param->keys);
+    values = ucs_alloca(key_count * sizeof(*values));
+    memset(values, 0, key_count * sizeof(*values));
+
+    /* Writable copy of the input buffer on stack */
+    tokens = ucs_alloca(buf_len);
+    ucs_strncpy_safe(tokens, buf, buf_len);
```
yosefe, line 1128: maybe use string buffer? it has tokenize api

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+
+    tokens = ucs_string_split(tokens, TOKEN_DELIM, 1, &token);
+    while (NULL != token) {
+        ucs_string_split(token, VALUE_DELIM, 2, &key, &value);
+
+        /* Remember last value in the values array */
+        if (UCS_OK != ucs_config_key_value_set(param->keys, values, key_count,
+                                               key, value)) {
+            ucs_error("Key '%s' is not supported", key);
```
yosefe, line 1137: no need to print another error

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+
+    /* Writable copy of the input buffer on stack */
+    tokens = ucs_alloca(buf_len);
+    ucs_strncpy_safe(tokens, buf, buf_len);
+
+    tokens = ucs_string_split(tokens, TOKEN_DELIM, 1, &token);
+    while (NULL != token) {
+        ucs_string_split(token, VALUE_DELIM, 2, &key, &value);
+
+        /* Remember last value in the values array */
```
yosefe, line 1134: We should do it only for values without keys

src/ucs/config/parser.h
```diff
@@ -65,6 +65,23 @@ typedef struct ucs_config_field {
 } ucs_config_field_t;
 
 
+typedef struct {
+    const char               *name;   /* Key name */
+    const char               *doc;    /* Documentation */
+    size_t                   offset;  /* Storage offset */
```
yosefe, line 71: can we use a static array instead of offset? i.e assume all values are stored sequentially, according to the order of keys in the keys array

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+
+    if (NULL == value) {
+        /* Empty value means that key is a generic value for all keys */
+        for (index = 0; index < key_count; ++index) {
+            values[index] = key;
+        }
+        return UCS_OK;
+    } else {
+        /* Key configured, save value for specific key */
+        for (index = 0; index < key_count; ++index, ++cfg_key) {
```
yosefe, line 1098: can we reuse ucs_string_find_in_list?

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+    const char **values;
+    char *token, *tokens;
+    const char *key, *value;
+
+    /* values is a temporary array value storage for the keys.
+     * In values array we remember the last applicable setting for each key.
+     * We need it to perform actual setting exactly once per each key */
+    key_count = ucs_config_key_value_count(param->keys);
+    values = ucs_alloca(key_count * sizeof(*values));
+    memset(values, 0, key_count * sizeof(*values));
```
yosefe, line 1124: i think we can make it simpler:
- go over the tokens in the input string
- for each key:value pair:
- if key is provided - find key index in the list and save value in "dest" in the right place by calling param->parser.read, and mark it as initialized (in a bitmap array / uint8_t array on the stack)
- otherwise, if key is not provided (value only) - read the value to a temp buffer (the default value)
- after going over the tokens, assign the default value to each un-initialized entry in "dest"

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+
+int ucs_config_sprintf_key_value(char *buf, size_t max, const void *src,
+                                 const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    char value_buf[256]                       = "";
+    const ucs_config_key_field_t *cfg_key;
+    ucs_string_buffer_t strb;
+
+    ucs_string_buffer_init_fixed(&strb, buf, max);
```
yosefe, line 1167: minor: can init inline by UCS_STRING_BUFFER_FIXED macro

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+
+    for (cfg_key = param->keys; NULL != cfg_key->name; ++cfg_key) {
+        if (!param->parser.write(value_buf, sizeof(value_buf),
+                                 src + cfg_key->offset, param->parser.arg)) {
+            /* Error is logged by parser impl */
+            return 0;
+        }
+        if (cfg_key != param->keys) {
+            ucs_string_buffer_appendf(&strb, ",");
+        }
```
yosefe, line 1177: can add "," every time, and then use ucs_string_buffer_rtrim

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+
+ucs_status_t ucs_config_clone_key_value(const void *src, void *dest,
+                                        const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    ucs_status_t status                       = UCS_OK;
+    const ucs_config_key_field_t *cfg_key;
+
+    for (cfg_key = param->keys; NULL != cfg_key->name; ++cfg_key) {
+        status = param->parser.clone(src + cfg_key->offset,
```
yosefe, line 1191: can we extract common code with ucs_config_clone_array?

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+}
+
+void ucs_config_release_key_value(void *ptr, const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    const ucs_config_key_field_t *cfg_key;
+
+    for (cfg_key = param->keys; NULL != cfg_key->name; ++cfg_key) {
+        param->parser.release(ptr + cfg_key->offset, param->parser.arg);
+    }
```
yosefe, line 1208: can we extract common code with ucs_config_release_array?

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+void ucs_config_help_key_value(char *buf, size_t max, const void *arg)
+{
+    const ucs_config_key_value_param_t *param = arg;
+    char syntax_buf[256]                      = "";
+    const ucs_config_key_field_t *cfg_key;
+    ucs_string_buffer_t strb;
+
+    ucs_string_buffer_init_fixed(&strb, buf, max);
+
+    ucs_string_buffer_appendf(&strb, "generic value for all keys and/or"
```
yosefe, line 1220: generic --> default

src/ucs/config/parser.c
```diff
@@ -1066,6 +1066,173 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+                              " comma-separated list of key:value pairs,"
+                              " where key is in [");
```
yosefe, line 1222: "in" -> "one of"

src/ucs/config/parser.c
```diff
@@ -1090,14 +1257,44 @@ static inline int ucs_config_is_table_field(const ucs_config_field_t *field)
...
+static char *ucs_config_get_doc(const ucs_config_field_t *field)
+{
+    const ucs_config_key_value_param_t *param;
+    const ucs_config_key_field_t *cfg_key;
+    ucs_string_buffer_t strb;
+
+    ucs_string_buffer_init(&strb);
+    ucs_string_buffer_appendf(&strb, "%s", field->doc);
+
+    if (ucs_config_is_key_value_field(field)) {
```
yosefe, line 1278: can't it be part of ucs_config_help_key_value?

src/ucs/config/parser.c
```diff
@@ -1112,12 +1309,27 @@ static void ucs_config_print_doc_line_by_line(const ucs_config_field_t *field,
...
+ucs_config_parser_parse_field(const ucs_config_field_t *field,
+                              const char *value, void *var)
 {
-    char syntax_buf[256];
+    char value_buf[512]  = "";
+    char syntax_buf[256] = "";
     int ret;
 
-    ret = field->parser.read(value, var, field->parser.arg);
+    if (ucs_config_is_key_value_field(field)) {
```
yosefe, line 1319: should not have a special case here, it should be part of ucs_config_sscanf_key_value

src/ucs/config/parser.h
```diff
@@ -400,6 +424,16 @@ void ucs_config_help_generic(char *buf, size_t max, const void *arg);
                                     (ucs_field_type(ucs_config_parser_t, help))   ucs_empty_function_do_assert, \
                                     ""}
 
+#define UCS_CONFIG_TYPE_KEY_VALUE(PARSER, ...) \
```
yosefe, line 427: PARSER --> _value

src/ucs/config/parser.h
```diff
@@ -400,6 +424,16 @@ void ucs_config_help_generic(char *buf, size_t max, const void *arg);
...
+    {ucs_config_sscanf_key_value, ucs_config_sprintf_key_value, \
+     ucs_config_clone_key_value,  ucs_config_release_key_value, \
+     ucs_config_help_key_value, \
+        (const ucs_config_key_value_param_t[]) { \
```
yosefe, line 431: why need type cast?

src/ucs/config/parser.c
iyastreb: ok, removed

src/ucs/config/parser.c
iyastreb: Unfortunately we can't.
Because we search in NULL terminated array of key configs (`ucs_config_key_field_t *`), not in array of strings

src/ucs/config/parser.c
iyastreb: ok

src/ucs/config/parser.c
iyastreb: sure

src/ucs/config/parser.c
iyastreb: done

src/ucs/config/parser.h
iyastreb: ok

src/ucs/config/parser.h
iyastreb: Please correct me if I misunderstood your idea. I didn't understand the role of 'static array' in your proposal.
Currently we have this design:
```
typedef struct {
    int             temp_front;
    int             temp_rear;
} car_opts_t;

  {"TEMP", "20", "Temperature", 0,
    UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_UINT,
        {"front", "front temperature", ucs_offsetof(car_opts_t, temp_front)},
        {"rear",  "rear temperature",  ucs_offsetof(car_opts_t, temp_rear)},
        {NULL}
    )},

```
Now what you propose would be something like:
```
typedef struct {
    int             temp_front;
    int             temp_rear;
// OR you think of this?
    int    temps[2];
} car_opts_t;

  {"TEMP", "20", "Temperature", ucs_offsetof(car_opts_t, temp_front), << offset to the first field
    UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_UINT,
        sizeof(int),  << we also need to store the size of the entry, right?
        {"front", "front temperature"},  << implicit offset 0
        {"rear",  "rear temperature"},  << implicit offset 0 + sizeof(entry)
        {NULL}
    )},

```
Did I get right your idea?

If I get it right, I have some concerns:
1. If we require that storage is always an array (e.g. `int    temps[2];`) then in C code it might not be obvious what temperature we are using by writing `temp[0]`. We might need to come up with enum in this case:
```
typedef enum {
  UCS_CONFIG_TEMP_FRONT,
  UCS_CONFIG_TEMP_REAR,
  CONFIG_TEMP_SIZE
}
int    temps[CONFIG_TEMP_SIZE];
int access = temps[UCS_CONFIG_TEMP_FRONT];
```
2. If we rely on sequential order of fields in the structure - that should work well for primitives, but is it safe regarding alignment or more complex types? There is another issue with this approach: variable access becomes implicit. There is no explicit reference to variable temp_rear. This is error-prone IMO

Also it's not very clear what are benefits of this approach (If I get it right of course)?

src/ucs/config/parser.c
iyastreb: Yes, I would also prefer it that way.
But with current design we cannot access `field->dfl_value` from the parser.
The easiest way would be to add one more argument to parser.read signature, so that it accepts also default value:
```
    int                      (*read) (const char *buf, void *dest, const void * dfl_value, const void *arg);

```
Would it be ok?

src/ucs/config/parser.c
iyastreb: It can, but help is also used to warn about config errors, so the consequence would be:
When user makes an error in config, we show him a help message, that would be much larger, e.g.
```
 Allowed keys:
    front      front temperature
    rear       rear temperature
    syntax:    default value for all keys and/or comma-separated list of key:value pairs, where key is one of allowed keys and value is: unsigned integer
``` 
Please confirm if this is desired behavior

src/ucs/config/parser.c
iyastreb: thanks, good idea indeed

src/ucs/config/parser.h
yosefe: very cool!

src/ucs/config/parser.c
yosefe: why do we print "allowed keys" and not the list of keys?
anyway we should not have special cases for config types in the general parsing code

src/ucs/config/parser.c
iyastreb: As the comment below explains, this is needed to support *partial setting*.
Supposedly we want to allow users to specify only subset of all the supported keys.
For instance, in our TEMP example with allowed keys [front,rear] I can make this setting:
`TEMP=rear:15`  (what is set for `front`?)
This is what I called a *partial setting*, meaning that my config does NOT provide explicit values for all the allowed keys.

Should we allow that? I thought *yes*, but I don't have a strong opinion on that. What's your take on that?
If it's not what we want to support, I just remove this code.

If we allow that, then we need to initialize those remaining keys somehow. And inside parser.read I have no access to default value.
So the idea was to concatenate the default value, followed by current setting. This way we guarantee that a) all fields are initialized, b) current setting is applied on top of default, so it has precedence.

yosefe: i think we should not allow that.
when setting a new config value, the default value in the table becomes irrelevant (this is how all configs work today)

src/ucs/config/parser.c
yosefe: Doc:
```
Specifies car temperatures.
- front : front temperature 
- rear  : rear temperature
```

Syntax:
```
Comma-separated list of value or key:value pairs, where key is one of {front, rear} and 
value is unsigned integer. A value without a key is the default.
```

iyastreb: Ok!
Another agreement is to add new `doc` function pointer to parser, so that we don't need to handle specific case here

src/ucs/config/parser.c
iyastreb: Ok, removing

src/ucs/config/parser.c
iyastreb: Done.
I just slightly changed the formatting to resemble existing configs:
```
#
# MMAP hook mode
#  none   - Don't set mmap hooks.
#  reloc  - Use ELF relocation table to set hooks.
#  bistro - Use binary instrumentation to set hooks.

# Cuda memory hook modes. A combination of:
#  none   - Don't set Cuda hooks.
#  reloc  - Use ELF relocation table to set hooks. In this mode, if any

# Module unload prevention mode
#  lazy - use RTLD_LAZY flag to add reference to module.
#  now  - use RTLD_NOW flag to add reference to module.

# Communication scheme in RNDV protocol.
#  get_zcopy - use get_zcopy scheme in RNDV protocol.
#  put_zcopy - use put_zcopy scheme in RNDV protocol.
```
So now it looks like this:
```
#
# Temperature
#  front     - front temperature
#  rear      - rear temperature
#
# syntax:    comma-separated list of value or key:value pairs, where key is one of [front,rear] and value is: unsigned integer. A value without a key is the default.
#
UCX_TEMP=front:17,rear:16
```

src/ucs/config/parser.c
iyastreb: LGTM, applying this change

iyastreb: Btw, currently I cannot easily declare string buffer of dynamic size on stack, because `UCS_ARRAY_ALLOC_ONSTACK` explicitly requires a compile-time constant for capacity.
May we can relax this requirement, and instead of checking at compile-time, perform check on runtime by using `ucs_alloca`?
```
 #define UCS_ARRAY_ALLOC_ONSTACK(_array_type, _capacity) \
     ({ \
         typedef ucs_typeof(*((_array_type*)NULL)->buffer) value_t; \
-        UCS_STATIC_ASSERT((_capacity) * sizeof(value_t) <= \
-                          UCS_ALLOCA_MAX_SIZE); \
-        (value_t*)alloca((_capacity) * sizeof(value_t)); \
+        (value_t*)ucs_alloca((_capacity) * sizeof(value_t)); \
     })
```
This would allow us to define arrays of dynamic size on stack
Any concerns?

iyastreb: Ok, I tried to use string buffer here, but I don't like the final result.
The reason is that tokenize algorithm of string buffer is less suitable for this particular use case. Basically here we do 2 phase tokenising: first we split on comma delimiter, and then on colon:
```
    tokens = ucs_string_split(tokens, TOKEN_DELIM, 1, &token);
    while (NULL != token) {
        ucs_string_split(token, VALUE_DELIM, 2, &key, &value);
        tokens = ucs_string_split(tokens, TOKEN_DELIM, 1, &token);
    }
```
ucs_string_split is convenient because it does 2 things at once: extracts token AND return string after the current token. This way I can split the current token as I want, and this does not affect behavior.
In string buffer tokenisation is implemented slightly differently: it just returns the current token.
This is fine if we don't split it further.. But if I split it further into key and value, then it breaks the iteration:
```
15:rear:50
token = 15, key = 15, value = NULL
token = rear:50, key = rear, value = 50
token = 50 << This happens because we split on the previous step
```
Well, we can fix it by manually advancing token pointer by certain size, but looks worse than `ucs_string_split`.
So I would keep the existing implementation for simplicity

yosefe: do you think it's possible to extend string_buffer tokenize api to support multi level tokenization? it could be useful.. if the change is too big we can keep the parse it as is

iyastreb: Well, look useful to me too. But I would definitely do it in a separate PR, because this one is already above 500 LOC:) I can do it together with array improvements (separate bit field for fixed flag), or in a separate PR.

Btw, what do you think about my previous comment about array on stack: https://github.com/openucx/ucx/pull/9555#discussion_r1432853798
?

src/ucs/config/parser.c
iyastreb: Ok, initially this looks good to me.
QQ about bitmaps: do we have an easy way to allocate bitmap of arbitrary size on stack?
Looking at bitmap.h I didn't find such API, but probably just overlooked.

In past I've created this super-simple bitset, that allows to create any dimensional bitmaps on stack:
```
#define TL3_BIT_MASK(bit)       (1 << ((bit) & (CHAR_BIT - 1)))
#define TL3_BIT_SET(set, bit)   (set[(bit)/CHAR_BIT] |= TL3_BIT_MASK(bit))
#define TL3_BIT_RESET(set, bit) (set[(bit)/CHAR_BIT] &= ~TL3_BIT_MASK(bit))
#define TL3_BIT_GET(set, bit)   (set[(bit)/CHAR_BIT] &  TL3_BIT_MASK(bit))
#define TL3_BITSET_SIZE(bits)   ((bits + CHAR_BIT - 1) / CHAR_BIT)

    uint8_t static_bitset[TL3_BITSET_SIZE(compile_time_size)];
    uint8_t *dynamic_bitset = alloca(TL3_BITSET_SIZE(dynamic_size));
```
would it be useful to have in UCX?

src/ucs/config/parser.c
yosefe: > Btw, what do you think about my previous comment about array on stack: [#9555 (comment)](https://github.com/openucx/ucx/pull/9555#discussion_r1432853798)

Sounds good to me

src/ucs/config/parser.c
```diff
@@ -1134,22 +1134,30 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
     }
 
-    /* Now apply stored temporary values, exactly once per each key */
+    /* Verify that there is a value for each key. Do it before actual set, so
+     * that we don't need to revert in case of failure */
     for (index = 0; index < key_count; ++index) {
         if (NULL == values[index]) {
-            ucs_error("No value configured for key '%s'",
-                      param->keys[index].name);
+            ucs_error("Partial setting is not allowed. No value configured for"
```
yosefe, line 1146: I would remote the string "Partial setting is not allowed. "

src/ucs/config/parser.c
```diff
@@ -1134,22 +1134,30 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
         }
+    }
+
+    /* Now apply stored temporary values, exactly once per each key */
+    for (index = 0; index < key_count; ++index) {
         if (param->parser.read(values[index], dest + param->keys[index].offset,
                                param->parser.arg) != 1) {
-            /* Error is logged by parser impl */
+            /* Revert already made settings */
+            while (index --> 0) {
```
yosefe, line 1157: code style

src/ucs/config/parser.c
```diff
@@ -1162,21 +1170,17 @@ int ucs_config_sprintf_key_value(char *buf, size_t max, const void *src,
...
         if (!param->parser.write(value_buf, sizeof(value_buf),
                                  src + cfg_key->offset, param->parser.arg)) {
             /* Error is logged by parser impl */
             return 0;
         }
-        if (cfg_key != param->keys) {
-            ucs_string_buffer_appendf(&strb, ",");
-        }
-        ucs_string_buffer_appendf(&strb, "%s:%s", cfg_key->name, value_buf);
+        ucs_string_buffer_appendf(&strb, "%s:%s,", cfg_key->name, value_buf);
```
yosefe, line 1181: space line before (after a block)

src/ucs/config/parser.c
```diff
@@ -1162,21 +1170,17 @@ int ucs_config_sprintf_key_value(char *buf, size_t max, const void *src,
...
     }
+    ucs_string_buffer_rtrim(&strb, ",");
```
yosefe, line 1183: space line before (after a block)

src/ucs/config/parser.c
```diff
@@ -1213,30 +1217,43 @@ void ucs_config_help_key_value(char *buf, size_t max, const void *arg)
...
+                              " pairs, where key is one of [");
 
     for (cfg_key = param->keys; NULL != cfg_key->name; ++cfg_key) {
-        if (cfg_key != param->keys) {
-            ucs_string_buffer_appendf(&strb, ",");
-        }
-        ucs_string_buffer_appendf(&strb, "%s", cfg_key->name);
+        ucs_string_buffer_appendf(&strb, "%s,", cfg_key->name);
     }
+    ucs_string_buffer_rtrim(&strb, ",");
```
yosefe, line 1228: space line before (after a block)

src/ucs/config/parser.c
yosefe: actually I plan to add dynamic bitmap based on ucs array

src/ucs/config/parser.c
```diff
@@ -1143,19 +1143,20 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
+                      param->keys[index].name);
             return 0;
         }
     }
 
     /* Now apply stored temporary values, exactly once per each key */
     for (index = 0; index < key_count; ++index) {
-        if (param->parser.read(values[index], dest + param->keys[index].offset,
+        if (param->parser.read(values[index],
+                               (char *)dest + param->keys[index].offset,
```
yosefe, line 1155: use UCS_PTR_BYTE_OFFSET or another wrapper macro on top of it

src/ucs/config/parser.c
```diff
@@ -1066,10 +1066,231 @@ void ucs_config_help_table(char *buf, size_t max, const void *arg)
...
+    const char *TOKEN_DELIM                   = ",";
+    const char *VALUE_DELIM                   = ":";
+    const char *common_value                  = NULL;
+    size_t key_count, index;
+    uint8_t *key_set;
+    char *token, *tokens;
+    const char *key, *value;
+
+    key_count = ucs_config_key_value_count(param->keys);
+    key_set = ucs_alloca(key_count * sizeof(*key_set));
```
brminich, line 1144:
```suggestion
    key_count = ucs_config_key_value_count(param->keys);
    key_set   = ucs_alloca(key_count * sizeof(*key_set));
```

test/gtest/ucs/test_config.cc
```diff
@@ -245,7 +265,7 @@ class test_config : public ucs::test {
                          const char *message, va_list ap)
     {
         // Ignore errors that invalid input parameters as it is expected
-        if (level == UCS_LOG_LEVEL_WARN) {
+        if (level == UCS_LOG_LEVEL_WARN || level == UCS_LOG_LEVEL_ERROR) {
```
brminich, line 268:
```suggestion
        if ((level == UCS_LOG_LEVEL_WARN) || (level == UCS_LOG_LEVEL_ERROR)) {
```

brminich (dismissed):

src/ucs/config/parser.c
```diff
@@ -1131,38 +1152,47 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
         }
+
         tokens = ucs_string_split(tokens, TOKEN_DELIM, 1, &token);
     }
 
-    /* Verify that there is a value for each key. Do it before actual set, so
-     * that we don't need to revert in case of failure */
+    /* Set common value if any */
+    if (NULL != common_value) {
+        if (UCS_OK != ucs_config_key_value_set_common(param, dest, key_set,
```
yosefe, line 1169: the wrapper ucs_config_key_value_set_common seems redundant
after moving its code here we can unite it with next loop:

```
for (index = 0; index < key_count; ++index) {
    if (key_set[index]) {
    } else if (NULL != common_value) {
       parser->read(..., common_value);
    } else {
       error("no value configured...");
    }
}
```

src/ucs/config/parser.c
```diff
@@ -1076,33 +1076,56 @@ static size_t ucs_config_key_value_count(const ucs_config_key_field_t *cfg_key)
...
+                return UCS_ERR_INVALID_PARAM;
             }
+
+            if (param->parser.read(value,
+                                   UCS_PTR_BYTE_OFFSET(dest, cfg_key->offset),
+                                   param->parser.arg) != 1) {
+                return UCS_ERR_INVALID_PARAM;
+            }
+
+            key_set[index] = 1;
```
yosefe, line 1124: IMO better not pass key_set to this function, it became too complicated.
a simple function can search for a key by name and return its index or -1 if not found

src/ucs/config/parser.c
```diff
@@ -1112,7 +1348,8 @@ static void ucs_config_print_doc_line_by_line(const ucs_config_field_t *field,
 }
 
 static ucs_status_t
-ucs_config_parser_parse_field(ucs_config_field_t *field, const char *value, void *var)
+ucs_config_parser_parse_field(const ucs_config_field_t *field,
+                              const char *value, void *var)
```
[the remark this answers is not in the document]
iyastreb: sure

src/ucs/config/parser.c
iyastreb: ok

src/ucs/config/parser.c
iyastreb: Yes, it's doable of course, but then the `ucs_config_sscanf_key_value` function does everything and becomes quite lengthy.

Initially I liked the idea of using flags, but at the end IMO it makes the code longer, more involved and harder to reason about. So after implementing this proposal, I think the original approach was actually cleaner:
read in a single place, no gotos.
Just compare with what we have now:
```
    tokens = ucs_string_split(tokens, TOKEN_DELIM, 1, &token);
    while (NULL != token) {
        ucs_string_split(token, VALUE_DELIM, 2, &key, &value);

        /* Remember last value in the values array */
        if (UCS_OK != ucs_config_key_value_set(param->keys, values, key_count,
                                               key, value)) {
            return 0;
        }
        tokens = ucs_string_split(tokens, TOKEN_DELIM, 1, &token);
    }

    /* Verify that there is a value for each key. Do it before actual set, so
     * that we don't need to revert in case of failure */
    for (index = 0; index < key_count; ++index) {
        if (NULL == values[index]) {
            ucs_error("No value configured for key '%s'",
                      param->keys[index].name);
            return 0;
        }
    }

    /* Now apply stored temporary values, exactly once per each key */
    for (index = 0; index < key_count; ++index) {
        if (param->parser.read(values[index],
                               (char *)dest + param->keys[index].offset,
                               param->parser.arg) != 1) {
            /* Revert already made settings */
            while (index-- > 0) {
                param->parser.release((char *)dest + param->keys[index].offset,
                                      param->parser.arg);
            }
            return 0;
        }
    }
    return 1;
```

iyastreb: Ok, I have just pushed the updated code that addresses your comments.
But personally I would prefer to rollback to the version without flags - it's just not worth it

yosefe: I agree the above code (wo the flags looks cleaner),
1. maybe remove ucs_config_key_value_set helper function?
2. seems we still need to revert if parsing failed, so maybe can unite the check that each key has a value with the parsing?

iyastreb: Ok, thanks, I will revert
1. ok
2. will try

src/ucs/config/parser.c
```diff
@@ -1115,59 +1118,42 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
         if (NULL == value) {
-            common_value = key;
+            /* No value means key is the default value for all */
+            default_value = key;
         } else {
-            if ((key_index = ucs_config_key_find(param->keys, key)) < 0) {
-                goto err;
+            /* Specific key config is present */
+            if ((index = ucs_config_key_find(param->keys, key)) < 0) {
+                return 0;
```
yosefe, line 1127: shall we print an error?

src/ucs/config/parser.c
```diff
@@ -1115,59 +1118,42 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
-                goto err;
-            }
-
-            if (param->parser.read(value,
-                                   UCS_PTR_BYTE_OFFSET(dest, param->keys[key_index].offset),
-                                   param->parser.arg) != 1) {
-                goto err;
-            }
-            key_set[key_index] = 1;
+            values[index] = value;
```
yosefe, line 1129: space line before

src/ucs/config/parser.c
```diff
@@ -1095,16 +1098,16 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
     const char *VALUE_DELIM                   = ":";
-    const char *common_value                  = NULL;
-    size_t key_count, index;
-    uint8_t *key_set;
+    const char *default_value                 = NULL;
+    const char *key, *value, **values;
     char *token, *tokens;
-    const char *key, *value;
-    int key_index;
+    size_t key_count;
```
yosefe, line 1104: the name 'index' is not recommended since it's a name of C library function

src/ucs/config/parser.c
```diff
@@ -1115,59 +1118,42 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
-        if (key_set[index]) {
-            param->parser.release(
-                UCS_PTR_BYTE_OFFSET(dest, param->keys[index].offset),
-                param->parser.arg);
+        /* Revert already made settings */
+        while (index-- > 0) {
+            param->parser.release(UCS_PTR_BYTE_OFFSET(dest, param->keys[index].offset),
+                                  param->parser.arg);
         }
+        return 0;
```
yosefe, line 1154: space line before

src/ucs/config/parser.c
```diff
@@ -1115,59 +1118,42 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
-            key_set[index] = 1;
+        value = values[index] ? values[index] : default_value;
+
+        if (NULL == value) {
+            ucs_error("No value configured for key '%s'", param->keys[index].name);
         } else {
-            ucs_error("No value configured for key '%s'",
-                      param->keys[index].name);
-            goto err;
+            if ((param->parser.read(value,
```
yosefe, line 1142: else if { .. on same line

src/ucs/config/parser.c
```diff
@@ -1115,59 +1118,42 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
-            if (param->parser.read(common_value,
-                                   UCS_PTR_BYTE_OFFSET(dest, param->keys[index].offset),
-                                   param->parser.arg) != 1) {
-                goto err;
-            }
-            key_set[index] = 1;
+        value = values[index] ? values[index] : default_value;
+
+        if (NULL == value) {
+            ucs_error("No value configured for key '%s'", param->keys[index].name);
```
yosefe, line 1140: smallcase

src/ucs/config/parser.c
```diff
@@ -1115,59 +1118,42 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
-            continue;
-        } else if (NULL != common_value) {
-            if (param->parser.read(common_value,
-                                   UCS_PTR_BYTE_OFFSET(dest, param->keys[index].offset),
-                                   param->parser.arg) != 1) {
-                goto err;
-            }
-            key_set[index] = 1;
+        value = values[index] ? values[index] : default_value;
+
```
yosefe, line 1138: remove space line

src/ucs/config/parser.c
```diff
@@ -1115,59 +1118,42 @@ int ucs_config_sscanf_key_value(const char *buf, void *dest, const void *arg)
...
-        if (key_set[index]) {
-            continue;
-        } else if (NULL != common_value) {
-            if (param->parser.read(common_value,
-                                   UCS_PTR_BYTE_OFFSET(dest, param->keys[index].offset),
-                                   param->parser.arg) != 1) {
-                goto err;
-            }
-            key_set[index] = 1;
+        value = values[index] ? values[index] : default_value;
```
yosefe, line 1137: `value = (values[index] == NULL) ? default_value : values[index];`

src/ucs/config/parser.c
iyastreb: Well, initially I had an error here:
```
ucs_error("Key '%s' is not supported", key);
```
But you suggested to remove it (`no need to print another error`): https://github.com/openucx/ucx/pull/9555#discussion_r1431286438

Should I put it back?

src/ucs/config/parser.c
iyastreb: sure

src/ucs/config/parser.c
iyastreb: ok

src/ucs/config/parser.c
iyastreb: ok

src/ucs/config/parser.c
iyastreb: done

src/ucs/config/parser.c
iyastreb: ok

src/ucs/config/parser.c
iyastreb: added

src/ucs/config/parser.c
iyastreb: sure

src/ucs/config/parser.c
yosefe: yes: the previous function ucs_config_key_value_set() used to print an error, but the new function ucs_config_key_find() does not print an error anymore.

yosefe (commented):
LGTM besides https://github.com/openucx/ucx/pull/9555#discussion_r1451459435

src/ucs/config/parser.c
iyastreb: Actually `ucs_config_key_value_set()` [didn't](https://github.com/openucx/ucx/pull/9555/commits/a925c6df4b72769fb8b4fbceba207e5086c8d198) print an error, but just returned a status.

Anyway, sure, I will add it back

yosefe (approved):

PR #9556, opened 2023-12-19
AZP: Show the full mpirun path before running tests

Alexey-Rivkin:


amaslenn (approved):

yosefe (approved):

contrib/test_jenkins.sh
```diff
@@ -475,9 +475,11 @@ run_ucx_perftest() {
 		if [ $with_mpi -eq 1 ]
 		then
 			# Run UCP performance test
+			which mpirun
```
artemry-nv, line 478: `command -v <cmd>` is the preferred way according to shellcheck (`which` is not a standard utility).

Alexey-Rivkin: Checked before and 'which' is present in all images that we use.

artemry-nv: Double checked - looks like shellcheck now doesn't raise this warning.

Issue #8892, opened 2023-02-19
Automate CLA


bkmgit:
### Describe the bug
Before making contributions, one needs to send a paper CLA agreement.  This can be time consuming. Consider automating and making use of tools such as:
- https://github.com/finos/cla-bot
- https://cla-assistant.io/
As an example see:
https://cla.salesforce.com/sign-cla

shamisp:
@bkmgit thanks for suggestion. It has been on our todo list for a while now.
As for now, you can email the CLA to `admin at ucfconsortium.org`

uranix:
@shamisp Is it still possible to sign individual CLA by email?

PR #9558, opened 2023-12-19
BUILD/CONFIG: Keep CFLAGS and CXXFLAGS separate

yselkowitz:
## What
Preventing CFLAGS from polluting CXXFLAGS in cases where they differ.

## Why ?
If configured with CFLAGS containing flags which are specific to C and invalid in CXX, this results in errors when compiling src/tools/perf/lib/uc[pt]_tests.cc:
```
cc1plus: error: '-Werror=' argument '-Werror=implicit-function-declaration' is not valid for C++ [-Werror]
cc1plus: error: '-Werror=' argument '-Werror=implicit-int' is not valid for C++ [-Werror]
```
This is currently breaking the build in Fedora 40 (rawhide) and ELN (the future RHEL 10).


yosefe:
@shamisp do we need CLA?

shamisp:
checking

yselkowitz:
Ping?

yselkowitz:
Ping?

bkmgit:
@yselkowitz Please see https://openucx.org/license/ there is a pdf with the contributer agreement that needs to be signed before changes can be merged.

bkmgit:
See also https://github.com/openucx/ucx/issues/8892

shamisp:
We have CLA with IBM but RH has to check if this they covered.

bkmgit:
Done in https://github.com/openucx/ucx/commit/be0246c7bc76abc8b29daeccc8fe7a9cf83edfd3

PR #9504, opened 2023-11-22
UCP: API - adding priority attribute to ucp requests

roiedanino:
## What
Adding priority attribute to `ucp_request_param_t`

## Why ?
Upcoming feature - high-priority messages, on supporting transports.

PR #9559, opened 2023-12-20
UCT/IB: Adding support for priorities in IB

roiedanino:
## What
Adding support for priority uct eps in IB

## Why ?
Support for the new priority messages feature: https://github.com/openucx/ucx/pull/9504 

## How ?
The user will set a dedicated SL by setting `UCX_IB_HIGH_PRIORITY_SL` env/config, UCP layer will create additional UCT priority eps in the supported transports.


src/uct/ib/rc/accel/rc_mlx5_ep.c
```diff
@@ -800,6 +800,9 @@ uct_rc_mlx5_ep_connect_to_ep_v2(uct_ep_h tl_ep,
                                         ep->super.path_index, &ah_attr,
                                         &path_mtu);
     ucs_assert(path_mtu != UCT_IB_ADDRESS_INVALID_PATH_MTU);
+    ah_attr.sl = ep->super.sl;
+
+    ucs_warn("PRIORITY SL: %u", ep->super.sl);
```
roiedanino, line 805: delete

PR #9564, opened 2023-12-25
UCS/BITMAP/TEST: Initial implementation of dynamic bitmap structure

yosefe:
## Why
Initial implementation of dynamic bitmap based on ucs_array and sharing common code with static bitmap.
Dynamic bitmap is needed for adding protocol variants (since total number of instantiated protocol variants is not known in advance)

### DIFF
--- a/src/ucs/datastruct/bitmap.h
+++ b/src/ucs/datastruct/bitmap.h
@@ -96,6 +96,30 @@ _ucs_bitmap_word_index(size_t bitmap_words, size_t bit_index)
 
 
 /**
+ * Assert that bit index is inside the expected range.
+ *
+ * @param _bit_index  Index to check.
+ * @param _num_words  Number of words in the bitmap
+ * @param _cmp        Comparison operator - should be < or <=
+ */
+#define UCS_BITMAP_CHECK_INDEX(_bit_index, _num_words, _cmp) \
+    ucs_assertv((_bit_index) _cmp (UCS_BITMAP_BITS_IN_WORD * (_num_words)), \
+                "bit_index=%zu num_words=%zu", (_bit_index), (_num_words))
+
+
+/**
+ * Check that destination number of words is at least as large as the source.
+ *
+ * @param _dst_num_words   Nunber of words in the destination bitmap.
+ * @param _src_num_words   Nunber of words in the source bitmap.
+ */
+#define UCS_BITMAP_CHECK_DST_NUM_WORDS(_dst_num_words, _src_num_words) \
+    ucs_assertv((_dst_num_words) >= (_src_num_words), \
+                "dst_num_words=%zu src_num_words=%zu", (_dst_num_words), \
+                (_src_num_words));
+
+
+/**
  * Given a bitmap and a bit index, get the whole word that contains it
  *
  * @param _bitmap    Take the word from this bitmap
@@ -567,6 +591,75 @@ _UCS_BITMAP_DECLARE_TYPE(64)
 _UCS_BITMAP_DECLARE_TYPE(128)
 _UCS_BITMAP_DECLARE_TYPE(256)
 
+
+/* Helper function to set all bitmap bits to 0 */
+static UCS_F_ALWAYS_INLINE void
+ucs_bitmap_bits_reset_all(ucs_bitmap_word_t *bits, size_t num_words)
+{
+    memset(bits, 0, num_words * sizeof(ucs_bitmap_word_t));
+}
+
+
+/* Helper function to return work mask of bit at given index */
+static UCS_F_ALWAYS_INLINE ucs_bitmap_word_t
+ucs_bitmap_word_bit_mask(size_t bit_index)
+{
+    return UCS_BIT(bit_index % UCS_BITMAP_BITS_IN_WORD);
+}
+
+
+/* Helper function to return word index of a given bit index */
+static UCS_F_ALWAYS_INLINE size_t ucs_bitmap_word_index(size_t bit_index,
+                                                        size_t num_words)
+{
+    UCS_BITMAP_CHECK_INDEX(bit_index, num_words, <);
+    return bit_index / UCS_BITMAP_BITS_IN_WORD;
+}
+
+
+/* Helper function to retrieve the value of a single bit */
+static UCS_F_ALWAYS_INLINE int
+ucs_bitmap_bits_get(const ucs_bitmap_word_t *bits, size_t num_words,
+                    size_t bit_index)
+{
+    const ucs_bitmap_word_t mask = ucs_bitmap_word_bit_mask(bit_index);
+    return !!(bits[ucs_bitmap_word_index(bit_index, num_words)] & mask);
+}
+
+
+/* Helper function to set the value of a single bit to 1 */
+static UCS_F_ALWAYS_INLINE void
+ucs_bitmap_bits_set(ucs_bitmap_word_t *bits, size_t num_words, size_t bit_index)
+{
+    const ucs_bitmap_word_t mask = ucs_bitmap_word_bit_mask(bit_index);
+    bits[ucs_bitmap_word_index(bit_index, num_words)] |= mask;
+}
+
+
+/* Helper function to set the value of a single bit to 0 */
+static UCS_F_ALWAYS_INLINE void ucs_bitmap_bits_reset(ucs_bitmap_word_t *bits,
+                                                      size_t num_words,
+                                                      size_t bit_index)
+
+{
+    const ucs_bitmap_word_t mask = ucs_bitmap_word_bit_mask(bit_index);
+    bits[ucs_bitmap_word_index(bit_index, num_words)] &= ~mask;
+}
+
+
+/* Helper function to test if all bitmap bits are 0 */
+static UCS_F_ALWAYS_INLINE int
+ucs_bitmap_bits_is_zero(const ucs_bitmap_word_t *bits, size_t num_words)
+{
+    const ucs_bitmap_word_t *bits_word;
+
+    ucs_carray_for_each(bits_word, bits, num_words) {
+        if (*bits_word != 0) {
+            return 0;
+        }
+    }
+    return 1;
+}
 
 END_C_DECLS
 
--- a/src/ucs/datastruct/dynamic_bitmap.h
+++ b/src/ucs/datastruct/dynamic_bitmap.h
@@ -1,0 +1,151 @@
+
+/**
+ * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2023. ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#ifndef UCS_DYNAMIC_BITMAP_H_
+#define UCS_DYNAMIC_BITMAP_H_
+
+#include <ucs/datastruct/array.h>
+#include <ucs/datastruct/bitmap.h>
+#include <ucs/sys/math.h>
+
+BEGIN_C_DECLS
+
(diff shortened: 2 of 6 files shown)


yosefe:
@gleon99 can you pls review?

src/ucs/datastruct/bitmap.h
```diff
@@ -95,6 +95,30 @@ _ucs_bitmap_word_index(size_t bitmap_words, size_t bit_index)
     ((_bit_index) < _UCS_BITMAP_NUM_WORDS(_bitmap) * UCS_BITMAP_BITS_IN_WORD)
 
 
+/**
+ * Assert if bit index is outside of the expected range.
```
gleon99, line 99: "Assert if" -> " [ Assert that | Check whether ] the "

src/ucs/datastruct/bitmap.h
```diff
@@ -95,6 +95,30 @@ _ucs_bitmap_word_index(size_t bitmap_words, size_t bit_index)
...
+ *
+ * @param _bit_index  Index to check.
+ * @param _num_word   Number of words in the bitmap
```
gleon99, line 102: `_num_words`

src/ucs/datastruct/bitmap.h
```diff
@@ -95,6 +95,30 @@ _ucs_bitmap_word_index(size_t bitmap_words, size_t bit_index)
...
+ * @param _cmp        Comparison operator - should be < or <=
+ */
+#define UCS_BITMAP_CHECK_INDEX(_bit_index, _num_words, _cmp) \
+    ucs_assertv((_bit_index) _cmp (UCS_BITMAP_BITS_IN_WORD *(_num_words)), \
```
gleon99, line 106: `* (_num_words)` (that's not a pointer)

src/ucs/datastruct/bitmap.h
gleon99: `outside` -> `inside`

src/ucs/datastruct/bitmap.h
```diff
@@ -568,6 +592,75 @@ _UCS_BITMAP_DECLARE_TYPE(128)
 _UCS_BITMAP_DECLARE_TYPE(256)
 
 
+/* Helper function to set all bitmap bits to 0 */
+static UCS_F_ALWAYS_INLINE void
+ucs_bitmap_bits_clear(ucs_bitmap_word_t *bits, size_t num_words)
```
gleon99, line 597: Maybe `ucs_bitmap_clear`?
Removing `_bits` saves redundant chars. and doesn't introduce any ambiguity.

src/ucs/datastruct/bitmap.h
```diff
@@ -568,6 +592,75 @@ _UCS_BITMAP_DECLARE_TYPE(128)
...
+                                                        size_t num_words)
+{
+    UCS_BITMAP_CHECK_INDEX(bit_index, num_words, <);
+    return bit_index / UCS_BITMAP_BITS_IN_WORD;
+}
+
+
+/* Helper function to retrieve the value of a single bit */
+static UCS_F_ALWAYS_INLINE int
+ucs_bitmap_bits_get(const ucs_bitmap_word_t *bits, size_t num_words,
```
gleon99, line 622: As it returns the value of a (single) bit, maybe rename to `_bits` or similar?

src/ucs/datastruct/bitmap.h
```diff
@@ -568,6 +592,75 @@ _UCS_BITMAP_DECLARE_TYPE(128)
...
+static UCS_F_ALWAYS_INLINE void
+ucs_bitmap_bits_set(ucs_bitmap_word_t *bits, size_t num_words, size_t bit_index)
+{
+    const ucs_bitmap_word_t mask = ucs_bitmap_word_bit_mask(bit_index);
+    bits[ucs_bitmap_word_index(bit_index, num_words)] |= mask;
+}
+
+
+/* Helper function to set the value of a single bit to 0 */
+static UCS_F_ALWAYS_INLINE void ucs_bitmap_bits_reset(ucs_bitmap_word_t *bits,
```
gleon99, line 640: func name inconsistency:
single bit to 0 is `reset`,
all bits to 0 is `clear`

src/ucs/datastruct/bitmap.h
```diff
@@ -568,6 +592,75 @@ _UCS_BITMAP_DECLARE_TYPE(128)
...
+    const ucs_bitmap_word_t mask = ucs_bitmap_word_bit_mask(bit_index);
+    bits[ucs_bitmap_word_index(bit_index, num_words)] &= ~mask;
+}
+
+
+/* Helper function to test if all bitmap bits are 0 */
+static UCS_F_ALWAYS_INLINE int
+ucs_bitmap_bits_is_zero(const ucs_bitmap_word_t *bits, size_t num_words)
+{
+    const ucs_bitmap_word_t *word;
```
gleon99, line 654: Minor:
In many other places there's `ucs_bitmap_word_t` **bits** and not **word**. Maybe rename for consistency

src/ucs/datastruct/dynamic_bitmap.h
```diff
@@ -0,0 +1,143 @@
+
+#ifndef UCS_DYNAMIC_BITMAP_H_
+#define UCS_DYNAMIC_BITMAP_H_
+
+#include <ucs/datastruct/bitmap.h>
+#include <ucs/datastruct/array.h>
```
gleon99, line 6: Order

src/ucs/datastruct/dynamic_bitmap.h
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * @param [inout] bitmap   Bitmap to reserve space in.
+ */
+static UCS_F_ALWAYS_INLINE void
+ucs_dynamic_bitmap_reserve(ucs_dynamic_bitmap_t *bitmap, size_t num_bits)
+{
+    size_t num_words = ucs_div_round_up(num_bits, UCS_BITMAP_BITS_IN_WORD);
+
+    if (num_words > ucs_array_length(bitmap)) {
+        ucs_array_resize(bitmap, num_words, 0, );
```
gleon99, line 62: Are we sure we want empty "failed action" here?

src/ucs/datastruct/dynamic_bitmap.h
```diff
@@ -0,0 +1,143 @@
...
+ * @param [in] bitmap    Read value from this bitmap.
+ * @param [in] bit_index Bit index to read.
+ *
+ * @return Bit value (0 or 1)
+ */
+static UCS_F_ALWAYS_INLINE size_t
+ucs_dynamic_bitmap_get(const ucs_dynamic_bitmap_t *bitmap, size_t bit_index)
+{
+    if (!ucs_dynamic_bitmap_is_in_range(bitmap, bit_index)) {
+        return 0;
```
gleon99, line 89: `0` is a **valid** possible bit value, which is also returned in case of an error.
Are we sure it is indeed the desired behavior? Might hide invalid calls..

src/ucs/datastruct/dynamic_bitmap.h
```diff
@@ -0,0 +1,143 @@
+
```
gleon99, line 1: Missing copyright?

src/ucs/datastruct/dynamic_bitmap.h
```diff
@@ -0,0 +1,143 @@
...
+        return 0;
+    }
+
+    return ucs_bitmap_bits_get(ucs_array_begin(bitmap),
+                               ucs_array_length(bitmap), bit_index);
+}
+
+
+/**
+ * Set the a bit in the bitmap to 1.
```
gleon99, line 98: `the a`? + below

src/ucs/datastruct/static_bitmap.h
```diff
@@ -0,0 +1,116 @@
...
+ * Get the number of words in a given bitmap
+ *
+ * @param _bitmap Bitmap variable to get words number
+ *
+ * @return Number of words
+ */
+#define UCS_STATIC_BITMAP_NUM_WORDS(_bitmap) \
+    ucs_static_array_size((_bitmap).bits)
+
+
```
gleon99, line 41: Docs missing. Intentionally?

src/ucs/datastruct/static_bitmap.h
```diff
@@ -0,0 +1,116 @@
...
+    ((ucs_bitmap_word_t*)(_bitmap_ptr)->bits), \
+            UCS_STATIC_BITMAP_NUM_WORDS(*(_bitmap_ptr))
+
+
+#define UCS_STATIC_BITMAP_BITS_CARGS(_bitmap_ptr) \
+    ((const ucs_bitmap_word_t*)(_bitmap_ptr)->bits), \
+            UCS_STATIC_BITMAP_NUM_WORDS(*(_bitmap_ptr))
+
+
+#define _UCS_STATIC_BITMAP_FUNC_CALL(_uid, _bitmap, _func, ...) \
```
gleon99, line 52: `uid` like "user id"? Why?

src/ucs/datastruct/static_bitmap.h
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @return Bit value (0 or 1)
+ */
+#define UCS_STATIC_BITMAP_GET(_bitmap, _bit_index) \
+    UCS_STATIC_BITMAP_FUNC_CALL(UCS_PP_UNIQUE_ID, _bitmap, \
+                                ucs_bitmap_bits_get, _bit_index)
+
+
+/**
+ * Set the a bit in the bitmap to 1.
```
gleon99, line 84: `the a`

test/gtest/ucs/test_bitmap.cc
```diff
@@ -6,6 +6,8 @@
 
 #include <common/test.h>
 #include <ucs/datastruct/bitmap.h>
+#include <ucs/datastruct/static_bitmap.h>
```
gleon99, line 9: Order

src/ucs/datastruct/bitmap.h
yosefe: wanted to distinguish between functions that operated on "bits, num_words" from functions that operate on a bitmap struct. the ones with "_bits" are helper functions

src/ucs/datastruct/bitmap.h
yosefe: _bits here is because of the "bits" argument

src/ucs/datastruct/bitmap.h
yosefe: "word" is when it's used to access a single word, and "bits" is when it is used as an array of words
do you think we should use "const ucs_bitmap_word_t *words" for example, instead of "const ucs_bitmap_word_t *bits"?

src/ucs/datastruct/dynamic_bitmap.h
yosefe: it's not really an error. I wanted to assume that un-allocated bits are considered to be 0

src/ucs/datastruct/static_bitmap.h
yosefe: like unique id, to create a temp variable name that will be different from others

src/ucs/datastruct/bitmap.h
gleon99: > do you think we should use "const ucs_bitmap_word_t *words" for example, instead of "const ucs_bitmap_word_t *bits"?

Yes, that sounds good. `const ucs_bitmap_word_t *words` is both meaningful & consistent, IMO.

gleon99 (approved):

PR #9566, opened 2023-12-26
AZP/POC: build with no CUDA module

Alexey-Rivkin:
## What
POC to reduce env modules usage.
Use images with pre-installed CUDA instead of loading `dev/cuda11.4` module.

### DIFF
--- a/buildlib/pr/main.yml
+++ b/buildlib/pr/main.yml
@@ -25,10 +25,10 @@ resources:
       image: rdmz-harbor.rdmz.labs.mlnx/ucx/coverity:mofed-5.1-2.3.8.0
       options: $(DOCKER_OPT_ARGS) $(DOCKER_OPT_VOLUMES)
     - container: rhel76
-      image: rdmz-harbor.rdmz.labs.mlnx/ucx/x86_64/rhel7.6/builder:mofed-5.0-1.0.0.0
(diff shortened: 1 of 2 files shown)


buildlib/pr/main.yml
```diff
@@ -37,16 +37,16 @@ resources:
       image: rdmz-harbor.rdmz.labs.mlnx/ucx/x86_64/rhel7.2/builder:mofed-5.0-1.0.0.0
       options: $(DOCKER_OPT_ARGS) $(DOCKER_OPT_VOLUMES)
     - container: rhel82
-      image: rdmz-harbor.rdmz.labs.mlnx/ucx/x86_64/rhel8.2/builder:mofed-5.0-1.0.0.0
+      image: rdmz-harbor.rdmz.labs.mlnx/ucx/x86_64/rhel8.2/builder-mofed-5.0-1.0.0.0:1
```
artemry-nv, line 40: Where are Dockerfiles for the new images?

Alexey-Rivkin: Those were never found. Installed CUDA over existing images instead of building from scratch.

artemry-nv: As we discussed investigate whether it's safe to use `docker commit` (potential docker layer issues). It may make sense to create a Dockerfile and use hpcx images as base.

Issue #9372, opened 2023-09-15
problem using device memory with get_ppln rendezvous Protocol 


edgargabriel:
### Describe the bug
As part of our internal test script for UCX we also run through all possible rendezvous protocols of UCX (put_zcopy, get_zcopy, put_ppln, get_ppln, am, rkey_ptr). With v1 protocols, all versions work and pass our tests.  With v2 protocols, we run into a problem when using the get_ppln protocol  for device memory (all other protocol types work correctly otherwise).

The problem seems to be that we get a host memory address passed to the rocm_ipc component for data transfer, which leads to an error/abort. The memory passed in stems from allocating the `ucp_rndv_frags` which in `ucp_proto_rndv_mtype_request_init()` is hard coded to be host memory type.

I am not entirely sure what the proper solution for this issue would be, or whether this is a way to circumvent this problem (other than not using get_ppln rendezvous protocol for this scenario).

### Setup and versions
ucx master with v2 protocols. OS is probably irrelevant.

### Additional information (depending on the issue)
[ucx_get_ppln_ipc.txt](https://github.com/openucx/ucx/files/12622926/ucx_get_ppln_ipc.txt)

edgargabriel:
I wanted to check back whether somebody has a suggestion on what the issue could be. The issue is still observed with UCX 1.16.0-RC1

edgargabriel:
issue seems fixed in ucx 1.18.0rc1

PR #9568, opened 2023-12-28
UCT/ROCM: adjust rocm/ipc performance parameters

edgargabriel:
## What
introduce two new parameters in rocm/ipc that allow to fine tune the protocol selection. 

## Why ?
Help to resolve the issue seen with v2 protocols for short message intra-node collective operations.


edgargabriel:
setting the DoNotMerge flag since I might have a better solution than the min_zcopy (which did most of the heavy lifting in this fix)

edgargabriel:
Lets stay for now with this solution here to deal with the  intra-node collectives performance issue that I'm looking into.

An alternative approach that I explored was to set the pre_send_overhead to a non-zero value in rocm_ipc, that gave me similar performance improvements as the solution in this pr.  However, I started to hit issue https://github.com/openucx/ucx/issues/9372 with the pre_send_overhead in a number of tests. Until we find a solution to the problem in that issues, this solution here (which mostly relies on setting a minimum message length for the rocm/ipc/zcopy operations) seems to be more stable.

This pr will also have to go to the 1.16 branch.

yosefe (approved):

yosefe:
pls check code style in CI
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=73762&view=logs&s=42f352bd-ef90-5911-18f7-6fd3843ebbaa&j=cc064a77-22b5-56bf-ecc0-70b5fe764261

edgargabriel:
@yosefe thank you, I think I fixed the issues

yosefe (approved):

Issue #9570, opened 2023-12-29
rndv_put.c:493  Assertion `req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED' failed


edgargabriel:
### Describe the bug
With ucx 1.16.0-rc1, I get an error when the rocm zcopy operation returns that the operation has finished right away (e.g. because the ROCM_COPY_ENABLE_ASYNC_ZCOPY flag has been set to 'no'). 

The error is:

```
[hyd-7c-zt05-02:1002786:0:1002786]    rndv_put.c:493  Assertion `req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED' failed

/home/egabriel/ucx-1.16.0/build/src/ucp/../../../src/ucp/rndv/rndv_put.c: [ ucp_proto_rndv_put_mtype_send_progress() ]
      ...
      490     ucp_request_t *req = ucs_container_of(uct_req, ucp_request_t, send.uct);
      491     const ucp_proto_rndv_put_priv_t *rpriv;
      492
==>   493     ucs_assert(req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED);
      494
      495     rpriv = req->send.proto_config->priv;
      496     return ucp_proto_multi_progress(req, &rpriv->bulk.mpriv,

==== backtrace (tid:1002786) ====
 0 0x000000000011e391 ucp_proto_rndv_put_mtype_send_progress()  /home/egabriel/ucx-1.16.0/build/src/ucp/../../../src/ucp/rndv/rndv_put.c:493
 1 0x000000000011cfd3 ucp_request_try_send()  /home/egabriel/ucx-1.16.0/build/../src/ucp/core/ucp_request.inl:305
 2 0x000000000011cfd3 ucp_request_send()  /home/egabriel/ucx-1.16.0/build/../src/ucp/core/ucp_request.inl:328
 3 0x000000000011cfd3 ucp_proto_rndv_put_mtype_pack_completion()  /home/egabriel/ucx-1.16.0/build/src/ucp/../../../src/ucp/rndv/rndv_put.c:446
 4 0x000000000011e2b3 ucp_invoke_uct_completion()  /home/egabriel/ucx-1.16.0/build/../src/ucp/core/ucp_request.inl:977
 5 0x000000000011e2b3 ucp_proto_rndv_mtype_copy()  /home/egabriel/ucx-1.16.0/build/src/ucp/../../../src/ucp/rndv/rndv_mtype.inl:137
[...]

PR #9571, opened 2023-12-29
UCP: set FLAG_PROTO_INITIALIZED before mtype_copy

edgargabriel:
## What
set the UCP_REQUEST_FLAG_PROTO_INITIALIZE flag in `ucp_proto_rndv_put_mtype_copy_progress()` before calling `ucp_proto_rndv_mtype_copy()`. In case the mytpe_pack/copy operation reports immediate completion, the progress function invoked later might fail because the UCP_REQUEST_FLAG_PROTO_INITIALIZED flag has not yet been set on the request otherwise

## Why ?
Fixes issue #9570

### DIFF
--- a/src/ucp/rndv/rndv_put.c
+++ b/src/ucp/rndv/rndv_put.c
@@ -472,13 +472,13 @@ ucp_proto_rndv_put_mtype_copy_progress(uct_pending_req_t *uct_req)
     }
 
     ucp_proto_rndv_put_common_request_init(req);
+    req->flags |= UCP_REQUEST_FLAG_PROTO_INITIALIZED;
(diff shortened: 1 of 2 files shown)


yosefe:
Good catch! do we need the same fix in ucp_proto_rndv_rkey_ptr_mtype_copy_progress?

edgargabriel:
probably yes, shall I add that to this pull request?

yosefe:
> probably yes, shall I add that to this pull request?

Yes please :)

edgargabriel:
done, thanks!

yosefe (approved):

PR #9574, opened 2023-12-31
UCT/API/V2: Introduce uct_iface_query_v2 API

roiedanino:
## What
API PR for uct_iface_query_v2 

## Why ?
To enable adding new fields to be queried by uct_iface_query without breaking compatibility (field_mask)
We plan to add new field to specify the number of supported active message priorities.

PR #9576, opened 2024-01-03
UCT/API: added priority field to uct_ep_params

roiedanino:
(cherry picked from commit c13e2aade3ae320f9543e5307b10c39cebd73c21)

## What
Adding a priority field to UCT eps

## Why ?
UCT eps need to be aware of their own priority as this info will be needed in underneath implementation, in IB for example different SLs might be used for different priorities

### DIFF
--- a/src/uct/api/uct.h
+++ b/src/uct/api/uct.h
@@ -969,7 +969,10 @@ enum uct_ep_params_field {
     UCT_EP_PARAM_FIELD_PRIV_DATA_LENGTH           = UCS_BIT(15),
 
     /** Enables @ref uct_ep_params::local_sockaddr */
-    UCT_EP_PARAM_FIELD_LOCAL_SOCKADDR             = UCS_BIT(16)
+    UCT_EP_PARAM_FIELD_LOCAL_SOCKADDR             = UCS_BIT(16),
+
+    /** Enables @ref uct_ep_params::priority */
+    UCT_EP_PARAM_FIELD_PRIORITY                   = UCS_BIT(17)
 };
 
 
@@ -1420,6 +1423,15 @@ struct uct_ep_params {
      * @ref UCT_IFACE_FLAG_CONNECT_TO_SOCKADDR capability.
      */
     const ucs_sock_addr_t             *local_sockaddr;
+
+    /**
+     * On supported transports, an additional uct_ep is created
+     * for higher priority messages, which can be handled differently in
+     * the underlying transport.
+     * @note The interface in this routine requires the 
+     * @ref UCT_IFACE_FLAG_MESSAGE_PRIORITY capability.  
+     */
+    unsigned                                 priority;
 };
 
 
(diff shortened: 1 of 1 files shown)


src/uct/api/uct.h
```diff
@@ -1420,6 +1423,13 @@ struct uct_ep_params {
      * @ref UCT_IFACE_FLAG_CONNECT_TO_SOCKADDR capability.
      */
     const ucs_sock_addr_t             *local_sockaddr;
+
+    /**
+     * On supported transports, an additional uct_ep is created
+     * for higher priority messages, which can be handled differently in
+     * the underlying transport.
+     */
+    unsigned                                 priority;
```
yosefe, line 1432: pls align on column

src/uct/api/uct.h
```diff
@@ -1420,6 +1423,13 @@ struct uct_ep_params {
      * @ref UCT_IFACE_FLAG_CONNECT_TO_SOCKADDR capability.
      */
     const ucs_sock_addr_t             *local_sockaddr;
+
+    /**
+     * On supported transports, an additional uct_ep is created
+     * for higher priority messages, which can be handled differently in
+     * the underlying transport.
```
yosefe, line 1430: This documents the usage and not the parameter
It should be something like: 
`Priority level of operations on this endpoint, supported values are 0..@ref uct_iface_attr_v2_t::max_priority-1. if not specified, the priority is considered to be 0, which is the lowest priority.`

yosefe (approved):

yosefe:
Build failures are because need to merge #9574 first

src/uct/api/uct.h
```diff
@@ -1397,29 +1400,37 @@ struct uct_ep_params {
...
-    const ucs_sock_addr_t             *local_sockaddr;
+    const ucs_sock_addr_t                   *local_sockaddr;
+
+    /**
+     * Priority level of operations on this endpoint, supported values 
+     * are 0.. @ref uct_iface_attr_v2_t::max_priority -1.
+     * if not specified, the priority is considered to be 0,
+     * which is the lowest priority.
+     */
+    unsigned                                priority;
```
brminich, line 1433: how will these priorities be specified in transport operations?
will it be added in another PR?

roiedanino: each transport will define it's own priority semantics, in IB eps will have sl attribute determined at ep creation:

`ep->sl = iface->config.priority_sls[ep_params->priority]`

It will be implemented in a separate PR.

brminich: a ok, so a single ep will support just one priority (description is confusing a bit)
do you think it is worth to add prio to `uct_ep_attr` to be able to query it?

roiedanino: At least for now, I don't think we are going to need it in uct_ep_attr, as I didn't see any use for that in the POC

brminich (approved):

PR #9580, opened 2024-01-04
UCS/BITMAP/TEST: Implement dynamic bitmap data structure

yosefe:
## Why
Implementation of a bitmap that can grow as needed

### DIFF
--- a/src/ucs/datastruct/bitmap.h
+++ b/src/ucs/datastruct/bitmap.h
@@ -21,6 +21,11 @@ BEGIN_C_DECLS
 typedef uint64_t ucs_bitmap_word_t;
 
 
+/* Binary operation on bitmap words */
+typedef ucs_bitmap_word_t (*ucs_bitmap_binary_op_t)(ucs_bitmap_word_t,
+                                                    ucs_bitmap_word_t);
+
+
 /*
  * Bits number in a single bitmap word
  */
@@ -600,6 +605,14 @@ ucs_bitmap_bits_reset_all(ucs_bitmap_word_t *bits, size_t num_words)
 }
 
 
+/* Helper function to set all bitmap bits to 1 */
+static UCS_F_ALWAYS_INLINE void
+ucs_bitmap_bits_set_all(ucs_bitmap_word_t *bits, size_t num_words)
+{
+    memset(bits, 0xff, num_words * sizeof(ucs_bitmap_word_t));
+}
+
+
 /* Helper function to return work mask of bit at given index */
 static UCS_F_ALWAYS_INLINE ucs_bitmap_word_t
 ucs_bitmap_word_bit_mask(size_t bit_index)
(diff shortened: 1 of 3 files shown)


src/ucs/datastruct/bitmap.h
```diff
@@ -661,6 +669,181 @@ ucs_bitmap_bits_is_zero(const ucs_bitmap_word_t *bits, size_t num_words)
...
+    UCS_BITMAP_CHECK_INDEX(bit_index, num_words, <=);
+    return ucs_bitmap_bits_popcount(bits, last_word) +
+           ((mask == 0) ? 0 : ucs_popcount(bits[last_word] & mask));
+}
+
+
+/* Helper function to copy between bitmap bit arrays. Destination size must
+   be at lease the source size */
+static UCS_F_ALWAYS_INLINE void
+ucs_bitmap_bits_copy(const ucs_bitmap_word_t *src_bits, size_t src_num_words,
```
iyastreb, line 705: I noticed that in other functions destination params come first, followed by source ones.
In this one the order is the opposite.
Maybe it makes sense to maintain the same order `(dst, src)` in all similar functions to avoid confusion?

test/gtest/ucs/test_bitmap.cc
```diff
@@ -352,6 +352,22 @@ class test_dynamic_bitmap : public ucs::test {
...
+
+    static void set(ucs_dynamic_bitmap_t &bitmap,
+                    const std::initializer_list<size_t> &offsets)
+    {
+        for (auto offset : offsets) {
+            ucs_dynamic_bitmap_set(&bitmap, offset);
+        }
+    }
+
+    void expect_bits(const std::initializer_list<size_t> &offsets)
```
iyastreb, line 364: IMO it would be better to have static function that takes bitmap by const reference, rather than using member `m_bitmap`. This way it's more reusable.
If you prefer to keep it as is, then this function may be const:
`void expect_bits(const std::initializer_list<size_t> &offsets) const`

src/ucs/datastruct/bitmap.h
yosefe: right. but actually this function can be moved to a later PR since it's needed only by static_bitmap

test/gtest/ucs/test_bitmap.cc
yosefe: changed to const

src/ucs/datastruct/bitmap.h
```diff
@@ -661,6 +669,181 @@ ucs_bitmap_bits_is_zero(const ucs_bitmap_word_t *bits, size_t num_words)
...
+    ucs_bitmap_word_t mask = ucs_bitmap_word_bit_mask(bit_index) - 1;
+
+    UCS_BITMAP_CHECK_INDEX(bit_index, num_words, <=);
+    return ucs_bitmap_bits_popcount(bits, last_word) +
+           ((mask == 0) ? 0 : ucs_popcount(bits[last_word] & mask));
+}
+
+
+/* Helper function to copy between bitmap bit arrays. Destination size must
+   be at lease the source size */
```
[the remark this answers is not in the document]
yosefe: will fix in next PR (removed the whole function)

iyastreb (approved):

yosefe:
@gleon99 can you pls review as well?

test/gtest/ucs/test_bitmap.cc
```diff
@@ -368,3 +384,105 @@ UCS_TEST_F(test_dynamic_bitmap, test_get_set_reset)
 
     EXPECT_TRUE(ucs_dynamic_bitmap_is_zero(&m_bitmap));
 }
+
```
tvegas1, line 387: maybe we need to add tests for at least some of the `ucs_bitmap_bits_*()` functions?

yosefe: they should all be tested by static_bitmap/dynamic_bitmap, since they are internal helper function IMO they don't deserve a separate testing..

tvegas1 (approved):

test/gtest/ucs/test_bitmap.cc
```diff
@@ -368,3 +384,105 @@ UCS_TEST_F(test_dynamic_bitmap, test_get_set_reset)
...
+    ucs_dynamic_bitmap_t bitmap2;
+    ucs_dynamic_bitmap_init(&bitmap2);
+
+    set(m_bitmap, {1, 3, 4});
+    set(bitmap2, {2, 3, 4, 5});
+
+    ucs_dynamic_bitmap_xor_inplace(&m_bitmap, &bitmap2);
+    expect_bits({1, 2, 5});
+
+    ucs_dynamic_bitmap_cleanup(&bitmap2);
```
gleon99, line 487: Maybe add (at least some basic sanity) scenarios for testing "zero-padding" in the various operations?

src/ucs/datastruct/bitmap.h
```diff
@@ -661,6 +669,164 @@ ucs_bitmap_bits_is_zero(const ucs_bitmap_word_t *bits, size_t num_words)
...
+           ((mask == 0) ? 0 : ucs_popcount(bits[last_word] & mask));
+}
+
+
+/* Helper function to set the bitmap array to a mask up to the given index */
+static UCS_F_ALWAYS_INLINE void ucs_bitmap_bits_mask(ucs_bitmap_word_t *bits,
+                                                     size_t num_words,
+                                                     size_t bit_index)
+{
+    size_t last_word = bit_index / UCS_BITMAP_BITS_IN_WORD;
```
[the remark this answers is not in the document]
yosefe: cannot use const here since it's incremented in the function

src/ucs/datastruct/bitmap.h
```diff
@@ -661,6 +669,164 @@ ucs_bitmap_bits_is_zero(const ucs_bitmap_word_t *bits, size_t num_words)
...
+    size_t word_index;
+
+    UCS_BITMAP_CHECK_DST_NUM_WORDS(dst_num_words, src_num_words);
+
+    for (word_index = 0; word_index < src_num_words; ++word_index) {
+        dst_bits[word_index] = ~src_bits[word_index];
+    }
+
+    if (dst_num_words > src_num_words) {
+        /* Set remaining bits in destination to 1 */
```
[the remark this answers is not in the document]
yosefe: not sure what you mean here?

gleon99 (approved):

PR #9581, opened 2024-01-04
BINDINGS/GO: Locked OS thread before setting CUDA device.

rakhmets:
## What
Locked OS thread before setting CUDA device.

## Why ?
This is an attempt to fix the following failure:
```
/scrap/azure/agent-08/AZP_WORKSPACE/_temp/ucx-73847/bin/goperftest -p=30800 -m=cuda
...
/scrap/azure/agent-08/AZP_WORKSPACE/_temp/ucx-73847/bin/goperftest -p=30800 -m=cuda -i=localhost
...
     cuda_ipc_md.c:212  UCX  ERROR   cuCtxGetDevice(&this_device) failed: invalid device context
        ucp_rkey.c:902  UCX  ERROR   failed to unpack remote key from remote md[5]: Input/output error
 Caught signal 11 (Segmentation fault: address not mapped to object at address 0x38)
```
Goroutine is an abstraction which allows to develop multi-threaded applications. There is no guarantee that goroutine will be executed on the same CPU thread for the entire time. What can result an issue, because CUDA device (aka cuContext) can be bound to the calling CPU thread.

### DIFF
--- a/bindings/go/src/examples/perftest/perftest.go
+++ b/bindings/go/src/examples/perftest/perftest.go
@@ -17,6 +17,7 @@ import (
 	. "ucx"
 	"unsafe"
 	. "cuda"
+	"runtime"
 )
 
 type PerfTestParams struct {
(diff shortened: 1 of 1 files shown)


ivankochin (approved):
LGTM!

yosefe (approved):

PR #9582, opened 2024-01-04
UCS/ARRAY: Use bit field for array is_fixed flag

iyastreb:
## What
2 improvements in this PR:
- Use bit field in ucs_array for storing `is_fixed` flag
- Allow to create array of dynamic (not just compile-time constant) size on stack 

## Why ?
Before this change we used lower bit of ucs_array `capacity` field to store the "fixed" flag.
There are 2 issues with this design choice:
- debugability: while you see certain value in the debugger for array capacity (e.g. say 3), this might not be the REAL capacity of an array, because LSB is simply disregarded when checking capacity. This is ambiguous and error-prone
- can't use full capacity of fixed array in some cases.
The problem here is that despite the REAL capacity is 3, ucs_array_capacity returns 2, because it discards the LSB. For dynamic arrays this is not an issue - it will just grow a bit in advance (still not good). But for fixed array this IS an issue: we cannot use full capacity of an array, and we cannot allocate more by definition.

For the second improvement - previously we were able to create arrays of compile-time-constant size on stack. However this does not allow user to create arrays of runtime size on stack, which is sometimes desirable. The idea here is to use ucs_alloca that imposes array size constraints at runtime.

## How ?
We discussed with Yossi whether we should use LSB or MSB of ucs_array capacity for storing "fixed" flag. I measured performance of both approaches and don't see any difference. However I have a slight preference of placing new bit field `is_fixed` after the `capacity` field (= LSB approach), just to maintain the same order of assignment.
Currently we have `{ buf, size, capacity }` static initializer for array, so I just wanted to keep the order these first three fields the same as before. This is safer option IMO

### DIFF
--- a/src/ucs/datastruct/array.c
+++ b/src/ucs/datastruct/array.c
@@ -21,18 +21,25 @@
 
 
 ucs_status_t ucs_array_grow(void **buffer_p, size_t *capacity_p,
-                            size_t min_capacity, size_t value_size,
-                            void **old_buffer_p, const char *array_name)
+                            size_t min_capacity, size_t max_capacity,
+                            size_t value_size, void **old_buffer_p,
+                            const char *array_name)
 {
-    void *old_buffer = *buffer_p;
-    size_t new_capacity, aligned_capacity, current_capacity;
+    void *old_buffer        = *buffer_p;
+    size_t current_capacity = *capacity_p;
+    size_t new_capacity;
     void *new_buffer;
 
-    current_capacity = *capacity_p;
-    new_capacity     = ucs_max(current_capacity * UCS_ARRAY_GROW_FACTOR,
-                               min_capacity);
-    aligned_capacity = (new_capacity + ~UCS_ARRAY_CAP_MASK) & UCS_ARRAY_CAP_MASK;
-    new_buffer       = ucs_malloc(aligned_capacity * value_size, array_name);
+    if (min_capacity > max_capacity) {
+        ucs_error("failed to grow %s from %zu to %zu, maximum capacity is %zu",
+                  array_name, current_capacity, min_capacity, max_capacity);
+        return UCS_ERR_EXCEEDS_LIMIT;
+    }
+
+    new_capacity = ucs_max(current_capacity * UCS_ARRAY_GROW_FACTOR,
+                           min_capacity);
+    new_capacity = ucs_min(new_capacity, max_capacity);
+    new_buffer   = ucs_malloc(new_capacity * value_size, array_name);
     if (new_buffer == NULL) {
         ucs_error("failed to grow %s from %zu to %zu elems of %zu bytes",
                   array_name, current_capacity, new_capacity, value_size);
@@ -43,8 +50,6 @@ ucs_status_t ucs_array_grow(void **buffer_p, size_t *capacity_p,
         goto out;
     }
 
-    ucs_assert(!(current_capacity & UCS_ARRAY_CAP_FLAG_FIXED));
-
     memcpy(new_buffer, *buffer_p, current_capacity * value_size);
 
 out:
@@ -54,7 +59,7 @@ out:
         *old_buffer_p = old_buffer;
     }
     *buffer_p   = new_buffer;
-    *capacity_p = aligned_capacity;
+    *capacity_p = new_capacity;
     return UCS_OK;
 }
 
--- a/src/ucs/datastruct/array.h
+++ b/src/ucs/datastruct/array.h
@@ -8,9 +8,11 @@
 #define UCS_ARRAY_H_
 
 #include <ucs/sys/compiler_def.h>
+#include <ucs/sys/math.h>
 #include <ucs/sys/preprocessor.h>
 #include <ucs/type/status.h>
 #include <stddef.h>
+#include <limits.h>
 
 BEGIN_C_DECLS
 
@@ -25,7 +27,8 @@ BEGIN_C_DECLS
     struct { \
         _value_type *buffer; \
         _index_type length; \
-        _index_type capacity; \
+        _index_type capacity : ((CHAR_BIT * sizeof(_index_type)) - 1); \
+        uint8_t     is_fixed : 1; \
     }
 
 
@@ -47,7 +50,7 @@ BEGIN_C_DECLS
  * calling @ref ucs_array_cleanup_dynamic()
  */
 #define UCS_ARRAY_DYNAMIC_INITIALIZER \
-    { NULL, 0, 0 }
+    { NULL, 0, 0, 0 }
 
 
 /**
@@ -67,7 +70,7 @@ BEGIN_C_DECLS
  * @endcode
  */
 #define UCS_ARRAY_FIXED_INITIALIZER(_buffer, _capacity) \
-    { (_buffer), 0, ucs_array_init_fixed_capacity(_capacity) }
+    { (_buffer), 0, (_capacity), 1 }
 
 
 /**
@@ -82,9 +85,7 @@ BEGIN_C_DECLS
 #define UCS_ARRAY_ALLOC_ONSTACK(_array_type, _capacity) \
     ({ \
         typedef ucs_typeof(*((_array_type*)NULL)->buffer) value_t; \
-        UCS_STATIC_ASSERT((_capacity) * sizeof(value_t) <= \
-                          UCS_ALLOCA_MAX_SIZE); \
-        (value_t*)alloca((_capacity) * sizeof(value_t)); \
+        (value_t*)ucs_alloca((_capacity) * sizeof(value_t)); \
     })
 
 
@@ -112,13 +113,6 @@ BEGIN_C_DECLS
 
 
 /**
- * Helper function to initialize capacity field of a fixed-size array.
- */
-#define ucs_array_init_fixed_capacity(_capacity) \
-    (((_capacity) & UCS_ARRAY_CAP_MASK) | UCS_ARRAY_CAP_FLAG_FIXED)
-
-
-/**
  * Initialize a dynamic array. Such array can grow automatically to accommodate
  * for more elements.
  *
@@ -129,6 +123,7 @@ BEGIN_C_DECLS
         (_array)->buffer   = NULL; \
         (_array)->length   = 0; \
         (_array)->capacity = 0; \
+        (_array)->is_fixed = 0; \
     }
 
 
@@ -141,9 +136,14 @@ BEGIN_C_DECLS
  */
 #define ucs_array_init_fixed(_array, _buffer, _capacity) \
     { \
+        ucs_assertv((_capacity) <= ucs_array_max_capacity(_array), \
+                    "capacity=%zu is out of range [0, %zu]", \
+                    (size_t)(_capacity), \
+                    (size_t)ucs_array_max_capacity(_array)); \
         (_array)->buffer   = (_buffer); \
         (_array)->length   = 0; \
-        (_array)->capacity = ucs_array_init_fixed_capacity(_capacity); \
+        (_array)->capacity = (_capacity); \
+        (_array)->is_fixed = 1; \
     }
 
 
@@ -188,6 +188,7 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
     ({ \
         ucs_status_t _reserve_status; \
         size_t _capacity; \
+        UCS_STATIC_ASSERT(ucs_is_unsigned_type(ucs_typeof((_array)->length))); \
         \
         if (ucs_likely((_min_capacity)) <= ucs_array_capacity(_array)) { \
             ucs_array_old_buffer_set_null((void**)(_old_buffer_p)); \
@@ -198,6 +199,7 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
             _capacity       = (_array)->capacity; \
             _reserve_status = ucs_array_grow((void**)&(_array)->buffer, \
                                              &_capacity, _min_capacity, \
+                                             ucs_array_max_capacity(_array), \
                                              sizeof(*(_array)->buffer), \
                                              (void**)(_old_buffer_p), \
                                              UCS_PP_MAKE_STRING(_array)); \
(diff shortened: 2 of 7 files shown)


src/ucs/debug/assert.h
```diff
@@ -69,9 +69,9 @@ BEGIN_C_DECLS
 
 #define UCS_ENABLE_ASSERT 0
 
-#define ucs_bug(...)
-#define ucs_assert(...)
-#define ucs_assertv(...)
+#define ucs_bug(...)     do {} while(0)
+#define ucs_assert(...)  do {} while(0)
+#define ucs_assertv(...) do {} while(0)
```
iyastreb, line 74: This is needed to silence cppcheck complaining about uninitialised value.
When built in release mode we used empty value for these macros. And when this macro is used in statement expression, then cppcheck complains about uninitialized value
```
#define ucs_alloca(_size) \
    ({ \
        ucs_assertv((_size) <= UCS_ALLOCA_MAX_SIZE, "alloca(%zu)", (size_t)(_size)); \
        alloca(_size); \
    })
```
unfolded as:
```
    ({
        ;
        alloca(_size);
    })
```
Despite this is valid statement expression, cppcheck does not like it due to redundant semicolon. So this change is needed to silence that warning

src/ucs/datastruct/array.h
```diff
@@ -143,7 +136,8 @@ BEGIN_C_DECLS
     { \
         (_array)->buffer   = (_buffer); \
         (_array)->length   = 0; \
-        (_array)->capacity = ucs_array_init_fixed_capacity(_capacity); \
```
tvegas1, line 146: minor, not strictly related to patch, could add `ucs_assertv((_array)->capacity == _capacity)` after assignment, also after call to `ucs_array_grow()` in same file.

tvegas1 (approved):

src/ucs/datastruct/array.h
```diff
@@ -25,7 +25,8 @@ BEGIN_C_DECLS
     struct { \
         _value_type *buffer; \
         _index_type length; \
-        _index_type capacity; \
+        _index_type capacity : ((CHAR_BIT * sizeof(_index_type)) - 1); \
```
yosefe, line 28: can we add some overflow checks that capacity does not grow above _index_type/2 ?

src/ucs/datastruct/array.h
```diff
@@ -314,14 +308,14 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
...
-    ((_array)->capacity & UCS_ARRAY_CAP_MASK)
+    ((_array)->capacity)
 
 
 /**
  * @return Whether this is a fixed-length array
  */
 #define ucs_array_is_fixed(_array) \
-    ((_array)->capacity & UCS_ARRAY_CAP_FLAG_FIXED)
+    ((_array)->is_fixed)
```
yosefe, line 318: maybe remove these wrappers now?

src/ucs/datastruct/array.h
```diff
@@ -314,14 +308,14 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
  * @return Current capacity of the array
  */
 #define ucs_array_capacity(_array) \
-    ((_array)->capacity & UCS_ARRAY_CAP_MASK)
+    ((_array)->capacity)
```
brminich, line 311: minor: can be on the same line (it is short now)

src/ucs/datastruct/array.h
brminich: - imo, using wrappers looks cleaner
- can be on the same line

iyastreb: While working on the fix I've already considered both ideas: remove macros and put it on the same line, and finally I decided to keep as it is now, because it's consistent with the rest of the macros in this file:
```
#define ucs_array_length(_array) \
    ((_array)->length)

#define ucs_array_elem(_array, _index) \
    ((_array)->buffer[_index])

#define ucs_array_begin(_array) \
    ((_array)->buffer)
```
So in order to keep this file consistent, I would just keep it as it is now.
Unless you have strong opinion on that

src/ucs/datastruct/array.h
iyastreb: sure, good idea indeed!

src/ucs/datastruct/array.h
iyastreb: Sure, it can, but then it will be different style from other similar macros declarations.
I would prefer to keep the same style across this header file

src/ucs/datastruct/array.h
yosefe: ok

src/ucs/datastruct/array.h
iyastreb: Right, I see what you mean. Will add those checks where makes sense

src/ucs/datastruct/array.c
```diff
@@ -21,18 +21,26 @@
...
-    new_buffer       = ucs_malloc(aligned_capacity * value_size, array_name);
+    new_capacity     = ucs_min(new_capacity, max_capacity);
+
+    if (min_capacity > new_capacity) {
+        ucs_error("failed to grow %s from %zu to %zu, maximum capacity is %zu",
+                  array_name, current_capacity, min_capacity, max_capacity);
+        return UCS_ERR_EXCEEDS_LIMIT;
+    }
+
+    new_buffer       = ucs_malloc(new_capacity * value_size, array_name);
```
brminich, line 43: minor:
```suggestion
    new_buffer = ucs_malloc(new_capacity * value_size, array_name);
```

brminich (approved):

src/ucs/datastruct/array.c
```diff
@@ -31,6 +32,14 @@ ucs_status_t ucs_array_grow(void **buffer_p, size_t *capacity_p,
     current_capacity = *capacity_p;
     new_capacity     = ucs_max(current_capacity * UCS_ARRAY_GROW_FACTOR,
                                min_capacity);
+    new_capacity     = ucs_min(new_capacity, max_capacity);
+
+    if (min_capacity > new_capacity) {
```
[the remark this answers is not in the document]
iyastreb: This condition would be always FALSE, because of the line above:
```
new_capacity     = ucs_min(new_capacity, max_capacity);
```
Here the `new_capacity` is what we can allocate, it never exceeds the maximum allowed.
`min_capacity` is the minimum capacity requested by the caller.
So here we check if client request cannot be fulfilled, therefore this condition.

src/ucs/datastruct/array.h
```diff
@@ -192,11 +196,15 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
...
                                              &_capacity, _min_capacity, \
+                                             ucs_array_max_capacity(_array), \
                                              sizeof(*(_array)->buffer), \
                                              (void**)(_old_buffer_p), \
                                              UCS_PP_MAKE_STRING(_array)); \
             if (_reserve_status == UCS_OK) { \
                 (_array)->capacity = _capacity; \
+                ucs_assertv((_array)->capacity == _capacity, \
+                            "Reserved capacity=%zu mismatch array capacity=%zu", \
+                            _capacity, (size_t)((_array)->capacity)); \
```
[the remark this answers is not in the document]
iyastreb: ok, will be fixed

src/ucs/datastruct/array.c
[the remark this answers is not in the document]
yosefe: i suggest to remove the line above, why need to calc min and then check condition, can't we just check if requested allocation is within the limit?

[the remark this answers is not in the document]
iyastreb: Mmm, no, I believe the line with `min` is needed.

Look, imagine you have an array of 100 elements with capacity=100 and max_capacity=127 and you want to add a new single element to it.
Then on growing, the algorithm would try to double the array size:
```
new_capacity = ucs_max(current_capacity * UCS_ARRAY_GROW_FACTOR, min_capacity); // = 200
```
200 is above the max_capacity, but there is no error here, because we can still allocate the maximum allowed (so 127). So `min` here guarantees that we don't allocate beyond max_capacity.

src/ucs/datastruct/array.h
```diff
@@ -408,6 +402,23 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
...
 
+/**
+ * Modify array capacity (maximum number of elements in the array)
+ *
+ * @param _array        Array whose capacity to set
+ * @param _new_capacity New capacity to set
+ */
+#define ucs_array_set_capacity(_array, _new_capacity) \
+    { \
+        ucs_assertv((0 <= (_new_capacity)) && \
```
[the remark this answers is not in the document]
iyastreb: I haven't seen warnings on my build machine, but let's see if CI produces them

src/ucs/datastruct/array.h
```diff
@@ -408,6 +402,23 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
...
+                    ((_new_capacity) <= ucs_array_max_capacity(_array)), \
```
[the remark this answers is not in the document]
iyastreb: Now this macro is called from 2 places: `ucs_array_init_fixed` and `ucs_array_grow`.
I think this check is needed mostly for the first case. For array grow it's basically redundant

src/ucs/datastruct/array.h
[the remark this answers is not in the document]
iyastreb: yes, you are right, coverity complains, so I will remove the first condition

src/ucs/datastruct/array.c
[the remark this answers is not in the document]
yosefe: so maybe first check if min_capacity>max_capacity and if yes print error?

src/ucs/datastruct/array.h
```diff
@@ -390,6 +402,22 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
     }
 
 
+/**
+ * Modify array capacity (maximum number of elements in the array)
+ *
+ * @param _array        Array whose capacity to set
+ * @param _new_capacity New capacity to set
+ */
+#define ucs_array_set_capacity(_array, _new_capacity) \
```
yosefe, line 411: now this macro is used only once so can remove it

iyastreb: ok

src/ucs/datastruct/array.c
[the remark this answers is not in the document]
iyastreb: yes, it would also work, will change

yosefe (approved):

src/ucs/datastruct/array.h
```diff
@@ -321,16 +322,10 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
...
+ * fixed-size or not, the maximum capacity range is reduced by 1 bit.
  */
 #define ucs_array_max_capacity(_array) \
-    ({ \
-        typedef ucs_typeof((_array)->length) length_t; \
-        (ucs_is_signed_type(length_t) ? \
-            (UCS_BIT(CHAR_BIT * sizeof(length_t) - 2) - 1) : \
-            (UCS_BIT(CHAR_BIT * sizeof(length_t) - 1) - 1)); \
-    })
+    (UCS_BIT(CHAR_BIT * sizeof((_array)->length) - 1) - 1)
```
yosefe, line 328: maybe use UCS_MASK_SAFE?

test/gtest/ucs/test_string.cc
```diff
@@ -9,6 +9,7 @@ extern "C" {
 #include <ucs/debug/memtrack_int.h>
 #include <ucs/datastruct/string_buffer.h>
 #include <ucs/datastruct/string_set.h>
+#include <ucs/sys/compiler.h>
```
yosefe, line 12: why needed? we didn't add new code to this file

iyastreb: Right, but now we use `ucs_alloca` in array.h, and BlueField build fails to find that function unless explicitly include this header

src/ucs/datastruct/array.h
iyastreb: ok

test/gtest/ucs/test_string.cc
yosefe: well, that's why i used UCS_STATIC_ASSERT rather than ucs_alloca

iyastreb: Yes, exactly, but the consequence of using `UCS_STATIC_ASSERT` is that we cannot allocate array of non-compile-time constant on stack

yosefe: ok, let's keep it as is then

src/ucs/datastruct/array.h
```diff
@@ -325,7 +326,7 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
  * fixed-size or not, the maximum capacity range is reduced by 1 bit.
  */
 #define ucs_array_max_capacity(_array) \
-    (UCS_BIT(CHAR_BIT * sizeof((_array)->length) - 1) - 1)
+    UCS_MASK_SAFE(CHAR_BIT * sizeof((_array)->length) - 1)
```
yosefe, line 329: use ( ) around * expression

src/ucs/sys/compiler_def.h
```diff
@@ -179,10 +179,10 @@
 /**
  * @param _type   Integer type.
  *
- * @return Whether integer type is signed.
+ * @return Whether integer type is unsigned.
  */
-#define ucs_is_signed_type(_type) \
-    ((_type)(-1) < (_type)(0))
+#define ucs_is_unsigned_type(_type) \
```
yosefe, line 184: add gtest for this macro

src/ucs/datastruct/array.h
iyastreb: ok

src/ucs/sys/compiler_def.h
iyastreb: sure, makes sense

yosefe (dismissed):

brminich (dismissed):

yosefe:
@iyastreb pls squash

yosefe (approved):

brminich (approved):

PR #9585, opened 2024-01-09
UCT/IB: Auto-detect GID idx with FLID routing

brminich:
## What
Use routable gid when FLID routing is configured and enabled

## Why ?
better use experience, no need to set GID manually for FLID based routing to work properly

### DIFF
--- a/src/uct/ib/base/ib_device.h
+++ b/src/uct/ib/base/ib_device.h
@@ -53,6 +53,10 @@
 #define UCT_IB_SITE_LOCAL_PREFIX          be64toh(0xfec0000000000000ul) /* IBTA 4.1.1 12b */
 #define UCT_IB_SITE_LOCAL_MASK            be64toh(0xffffffffffff0000ul) /* IBTA 4.1.1 12b */
 #define UCT_IB_SITE_LOCAL_FLID_MASK       be64toh(0xffffffff00000000ul) /* site-local + flid */
+#define UCT_IB_GUID_OPENIB_OUI            0x001405 /* An OUI is a 24 bit globally unique assigned
+                                                      number referenced by various standards.
+                                                      IB_OPENIB_OUI is part of the routable alias
+                                                      GUID built by SM. */
 #define UCT_IB_DEFAULT_ROCEV2_DSCP        106  /* Default DSCP for RoCE v2 */
 #define UCT_IB_ROCE_UDP_SRC_PORT_BASE     0xC000
 #define UCT_IB_CQE_SL_PKTYPE_MASK         0x7 /* SL for IB or packet type
@@ -65,6 +69,9 @@
 #define UCT_IB_DEVICE_SYSFS_GID_NDEV_FMT  UCT_IB_DEVICE_SYSFS_GID_ATTR_PFX "/ndevs/%d"
 #define UCT_IB_DEVICE_ECE_DEFAULT         0x0         /* default ECE */
 #define UCT_IB_DEVICE_ECE_MAX             0xffffffffU /* max ECE */
+#define UCT_IB_DEVICE_DEFAULT_GID_INDEX 0   /* The gid index used by default for an IB/RoCE port */
+#define UCT_IB_DEVICE_ROUTABLE_FLID_GID_INDEX 1 /* The gid index used by default
+                                                   with FLID based IB routing */
 
 
 enum {
--- a/src/uct/ib/base/ib_iface.c
+++ b/src/uct/ib/base/ib_iface.c
@@ -1286,6 +1286,46 @@ out_mask_info_failed:
     return UCS_OK;
 }
 
+static unsigned uct_ib_iface_gid_index(uct_ib_iface_t *iface,
+                                       unsigned long cfg_gid_index)
+{
+    uct_ib_device_t *dev = uct_ib_iface_device(iface);
+    uint8_t port_num     = iface->config.port_num;
+    int gid_tbl_len      = uct_ib_device_port_attr(dev, port_num)->gid_tbl_len;
+    unsigned gid_index   = UCT_IB_DEVICE_DEFAULT_GID_INDEX;
+    uct_ib_device_gid_info_t gid_info;
+    ucs_status_t status;
+    uint32_t oui;
+
+    if (cfg_gid_index != UCS_ULUNITS_AUTO) {
+        return cfg_gid_index;
+    }
+
+    if (!iface->config.flid_enabled ||
+        (gid_tbl_len <= UCT_IB_DEVICE_ROUTABLE_FLID_GID_INDEX)) {
+        goto out;
+    }
+
+    status = uct_ib_device_query_gid_info(
+                          dev->ibv_context, uct_ib_device_name(dev), port_num,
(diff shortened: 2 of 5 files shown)


src/uct/ib/base/ib_iface.c
```diff
@@ -1286,6 +1286,41 @@ uct_ib_iface_init_roce_addr_prefix(uct_ib_iface_t *iface,
...
+    if (uct_ib_iface_gid_extract_flid(&gid_info.gid) == 0) {
+        goto out;
+    }
+
+    oui = be32toh(gid_info.gid.global.interface_id & 0xffffff) >> 8;
+    if (oui == UCT_IB_OPENIB_OUI) {
+        gid_index = UCT_IB_MD_DEFAULT_ROUTE_GID_INDEX;
+    }
+
+out:
```
iyastreb, line 1320: Minor: personally I would avoid using goto when possible, but up to you

src/uct/ib/base/ib_device.h
```diff
@@ -53,6 +53,7 @@
 #define UCT_IB_SITE_LOCAL_PREFIX          be64toh(0xfec0000000000000ul) /* IBTA 4.1.1 12b */
 #define UCT_IB_SITE_LOCAL_MASK            be64toh(0xffffffffffff0000ul) /* IBTA 4.1.1 12b */
 #define UCT_IB_SITE_LOCAL_FLID_MASK       be64toh(0xffffffff00000000ul) /* site-local + flid */
+#define UCT_IB_OPENIB_OUI                 0x001405
```
yosefe, line 56: is it part of GID/GUID? let's add to the name

src/uct/ib/base/ib_iface.c
```diff
@@ -1286,6 +1286,41 @@ uct_ib_iface_init_roce_addr_prefix(uct_ib_iface_t *iface,
     return UCS_OK;
 }
 
+static unsigned uct_ib_iface_gid_index(uct_ib_iface_t *iface)
```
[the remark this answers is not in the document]
yosefe: we don't usually do that

src/uct/ib/base/ib_iface.c
```diff
@@ -1286,6 +1286,41 @@ uct_ib_iface_init_roce_addr_prefix(uct_ib_iface_t *iface,
...
+    uint32_t oui;
+
+    if (!iface->config.flid_enabled ||
+        (gid_tbl_len <= UCT_IB_MD_DEFAULT_ROUTE_GID_INDEX)) {
+        goto out;
+    }
+
+    status = uct_ib_device_query_gid_info(
+                          dev->ibv_context, uct_ib_device_name(dev), port_num,
+                          UCT_IB_MD_DEFAULT_ROUTE_GID_INDEX, &gid_info);
```
yosefe, line 1306: shall we search all valid GIDs, not only index 1?

src/uct/ib/base/ib_iface.c
```diff
@@ -1307,10 +1342,9 @@ uct_ib_iface_init_gid_info(uct_ib_iface_t *iface,
             goto out;
         }
     } else {
-        gid_info->gid_index             = (cfg_gid_index ==
-                                           UCS_ULUNITS_AUTO) ?
-                                          UCT_IB_MD_DEFAULT_GID_INDEX :
-                                          cfg_gid_index;
+        gid_info->gid_index             = (cfg_gid_index != UCS_ULUNITS_AUTO) ?
+                                          cfg_gid_index :
```
yosefe, line 1346: maybe unite this logic between IB/ROCE - that if cfg did index is not auto, use it

iyastreb (approved):

tvegas1 (approved):

src/uct/ib/base/ib_iface.c
brminich: afaiu, it is not necessary.
need to change?

yosefe: ok, so maybe rename to UCT_IB_DEVICE_FLID_GID_INDEX?

brminich: ok

brminich: btw, i'd make it `UCT_IB_MD_DEFAULT_FLID_ROUTE_GID_INDEX` then, because when FLID routing is configured, all GIDS contain FLID, but we have to mention that we need routable GID

yosefe: UCT_IB_DEVICE_ROUTABLE_FLID_GID_INDEX

src/uct/ib/base/ib_device.h
brminich: it's a specific value we look for in GID.
do you propose something like `UCT_IB_GID_OPENIB_OUI`? 
imo, this looks a bit confusing because it is not a gid, but some specific ID

yosefe: so can we provide some documentation for it?

src/uct/ib/base/ib_iface.c
brminich: shall we rename `UCT_IB_MD_DEFAULT_GID_INDEX` then? Otherwise we would have different name patterns for different gid indexes. So either we have 
```
UCT_IB_MD_DEFAULT_GID_INDEX
UCT_IB_MD_ROUTABLE_FLID_GID_INDEX
```
or
```
UCT_IB_DEVICE_DEFAULT_GID_INDEX
UCT_IB_DEVICE_ROUTABLE_FLID_GID_INDEX
```

src/uct/ib/base/ib_iface.c
brminich: tried that looks heavier than it was. 
we will have to check iface type (roce or not) at least twice, also will need to change `uct_ib_iface_init_roce_gid_info` and `uct_ib_device_select_gid` to unify with ib part

src/uct/ib/base/ib_iface.c
yosefe: IMO 2nd option, and move it to ib_device.h

src/uct/ib/base/ib_device.h
brminich: removed this one and used `IB_OPENIB_OUI` from `mad.h`. Also added comment

src/uct/ib/base/ib_iface.c
```diff
@@ -24,6 +24,7 @@
 #include <ucs/debug/log.h>
 #include <ucs/time/time.h>
 #include <ucs/sys/sock.h>
+#include <infiniband/mad.h>
```
yosefe, line 27: why needed? seems it can fail the build in some envs

src/uct/ib/base/ib_iface.c
yosefe: maybe move the AUTO check to uct_ib_iface_gid_index?

src/uct/ib/base/ib_iface.c
brminich: leftover, tried to use value from this header before

yosefe (approved):

PR #9588, opened 2024-01-10
AZP/RELEASE: Refactor GPG sign and publish of Jar files - 1.16.x

Alexey-Rivkin:
## What
Refactoring the process for releasing Jar files to enable build, sign, and publishing per platform.

## Why ?
Failures in Maven publish otherwise.

## How ?
1. Move forward from CentOS-7 to CentOS-8 due to the outdated GPG agent.
2. Add the 'pinentry' package in Docker.
3. Configure the GPG agent to use pinentry.
4. Add filename suffix for non-x86 platforms.

### DIFF
--- a/buildlib/dockers/centos-release.Dockerfile
+++ b/buildlib/dockers/centos-release.Dockerfile
@@ -5,27 +5,28 @@ FROM nvidia/cuda:${CUDA_VERSION}-devel-centos${OS_VERSION}
 RUN yum install -y \
     autoconf \
     automake \
+    environment-modules \
+    ethtool \
     file \
+    fuse-libs \
     gcc-c++ \
     git \
     glibc-devel \
     libtool \
+    libusbx \
+    lsof \
     make \
     maven \
     numactl-devel \
+    pinentry \
+    python36 \
     rdma-core-devel \
     rpm-build \
     tcl \
     tcsh \
     tk \
-    wget \
-    libusbx \
-    fuse-libs \
-    python36 \
-    lsof \
-    ethtool \
-    environment-modules \
     valgrind-devel \
+    wget \
     && yum clean all
 
 # MOFED
--- a/buildlib/jucx/jucx-build.yml
+++ b/buildlib/jucx/jucx-build.yml
@@ -8,6 +8,9 @@ parameters:
 jobs:
   - job: jucx_build_${{ parameters.arch }}
     displayName: JUCX build ${{ parameters.arch }}
+    variables:
+      ${{ if ne(parameters.arch, 'amd64') }}:
+        SUFFIX: "-${{ parameters.arch }}"
 
     # we need to use lowest version for compatibility
     container: ${{ parameters.container }}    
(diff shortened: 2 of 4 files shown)


buildlib/dockers/centos-release.Dockerfile
amaslenn, line 1: If this file is updated, shouldn't you also update tag somewhere?

buildlib/jucx/jucx-build.yml
```diff
@@ -43,7 +46,7 @@ jobs:
           MAVEN_VERSION=${TAG:1}
           make -C bindings/java/src/main/native/ package JUCX_VERSION=${MAVEN_VERSION}
         displayName: Build JUCX
-        condition: eq(variables['Build.Reason'], 'PullRequest')
+        condition: eq(variables['Build.Reason'], 'IndividualCI')
```
amaslenn, line 49: Is it for debug?

buildlib/dockers/centos-release.Dockerfile
Alexey-Rivkin: For a significant change - sure. Only a single tiny package was added by this change, no collateral damage expected.

buildlib/jucx/jucx-build.yml
Alexey-Rivkin: This is to only run when the condition is right. 
[Sample skip log](https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=74349&view=logs&j=63c6b53b-67c9-50a3-0a67-f7e911956bf3&t=cf9a9368-6788-584f-400e-578b0d707aeb&l=1)

buildlib/dockers/centos-release.Dockerfile
amaslenn: But this breaks reproducibility: without new package new logic should fail. Old logic should probably work in both envs. While it looks minor in this case, overriding container tags is a very bad and dangerous practice.

Alexey-Rivkin: Old logic did not work, thus the change. Agree with the best practice emphasized. It's important to note that the tag was already overwritten in this instance.

amaslenn (approved):

buildlib/jucx/jucx-build.yml
```diff
@@ -82,12 +85,20 @@ jobs:
           cp $(publicKey.secureFilePath)  ${{ parameters.gpg_dir }}/pubring.gpg
           cp $(privateKey.secureFilePath) ${{ parameters.gpg_dir }}/secring.gpg
           export GNUPGHOME=${{ parameters.gpg_dir }}
+
+          # GPG agent config
```
artemry-nv, line 89: Is this a reference config? Where is it from?

artemry-nv (approved):

buildlib/jucx/jucx-build.yml
Alexey-Rivkin: Based on experimenting with various solutions recommended in online forums.

PR #9592, opened 2024-01-11
UCP/RCACHE: Fix ibv_reg_mr error by registering memory in rcache callback

yosefe:
## Why ?
Fix sporadic ibv_reg_mr() failure with "Cannot allocate memory" or "Bad address"
[Internal issue](https://redmine.mellanox.com/issues/3605689)

## How ?
The rcache memory registration callback is expected to fail if the memory region contains invalid addresses. In this case, rcache will revert a merge operation and register only the minimal requested region. Currently UCP rcache memory registration callback does not do that, so rcache may merge with invalid memory area, leading to a registration error later on.

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -25,6 +25,13 @@
 #include <string.h>
 #include <inttypes.h>
 
+/* Context for rcache memory registration callback */
+typedef struct {
+    ucs_memory_type_t mem_type;    /* Memory type */
+    ucp_md_map_t      reg_md_map;  /* Map of memory domains to be registered */
+    unsigned          uct_flags;   /* UCT memory registration flags */
+    const char        *alloc_name; /* Memory allocation name */
+} ucp_mem_rcache_reg_ctx_t;
 
 ucp_mem_dummy_handle_t ucp_mem_dummy_handle = {
     .memh = {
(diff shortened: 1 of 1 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -1632,8 +1661,8 @@ ucp_memh_import_slow(ucp_context_h context, ucs_rcache_t *existing_rcache,
         }
 
         status = ucp_memh_rcache_get(rcache, unpacked->address,
-                                     unpacked->length, unpacked->mem_type,
-                                     UCS_RCACHE_MIN_ALIGNMENT, &memh);
+                                     unpacked->length, UCS_RCACHE_MIN_ALIGNMENT,
+                                     unpacked->mem_type, 0, 0, "", &memh);
```
brminich, line 1665: why `md_map` is 0 here?

brminich (commented):
is gtest possible?

src/ucp/core/ucp_mm.c
```diff
@@ -1386,20 +1398,37 @@ ucp_mem_rcache_mem_reg_cb(void *ctx, ucs_rcache_t *rcache, void *arg,
...
+        uct_flags |= UCT_MD_MEM_FLAG_HIDE_ERRORS;
+    }
+
+    status = ucp_memh_register(context, memh, reg_ctx->reg_md_map, uct_flags,
+                               reg_ctx->alloc_name);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    if (reg_ctx->reg_md_map != memh->md_map) {
```
Artemy-Mellanox, line 1427: if UCT_MD_MEM_FLAG_HIDE_ERRORS we return UCS_OK even if not all registered

src/ucp/core/ucp_mm.c
yosefe: because we don't need to register an imported key

src/ucp/core/ucp_mm.c
yosefe: but when rcache was in UCT, we failed if it was not registered. we have to fail here to indicate that merge should be aborted..
anyway, reworked the function to return failure in such case in the next commit

Artemy-Mellanox (approved):

PR #9594, opened 2024-01-11
AZP: Remove "new" machines from CI - master

Alexey-Rivkin:
## What

Remove duplicate testing on "new" machines.

## Why ?
Requested to remove "new" machines.


yosefe:
This PR is probably not needed anymore on master branch since we removed Valgrind on new and replaced it by ASAN

PR #9596, opened 2024-01-11
UCT/ROCM: set parameters dependent on GPU type

edgargabriel:
## What
set the bandwidth values in the rocm/copy performance estimate functions based on the GPU type that we found. We use the same parameters for MI300A and MI300X at the moment, but the infrastructure is in place in case we need to distinguish between them at a later point.

## Why ?
Performance fixes with ucx 1.16 for non MI200 GPUs

### DIFF
--- a/src/uct/rocm/base/rocm_base.c
+++ b/src/uct/rocm/base/rocm_base.c
@@ -501,6 +501,54 @@ ucs_status_t uct_rocm_base_get_link_type(hsa_amd_link_info_type_t *link_type)
     return UCS_OK;
 }
 
+uct_rocm_amd_gpu_product_t uct_rocm_base_get_gpu_product(void)
+{
+    uct_rocm_amd_gpu_product_t gpu_product = UCT_ROCM_AMD_GPU_MI200;
+    char product_name[64];
+    char gfx_name[64];
+    hsa_status_t status;
+
+    /* fetching data from GPU 0, assuming all GPUs on a node are
+       identical */
+    status = hsa_agent_get_info(uct_rocm_base_agents.gpu_agents[0],
+                                (hsa_agent_info_t)
+                                        HSA_AMD_AGENT_INFO_PRODUCT_NAME,
+                                (void*)product_name);
+    if (status != HSA_STATUS_SUCCESS) {
+        ucs_debug("Error in hsa_agent_info %d", status);
+        return gpu_product;
+    }
+
+    if (NULL != strstr(product_name, "MI300A")) {
+        gpu_product = UCT_ROCM_AMD_GPU_MI300A;
+        ucs_trace("found MI300A GPU");
+    } else if (NULL != strstr(product_name, "MI300X")) {
+        gpu_product = UCT_ROCM_AMD_GPU_MI300X;
+        ucs_trace("found MI300X GPU");
+    } else {
+        /* In case product_name is not set correctly, query the gfx
+           architecture name */
+        status = hsa_agent_get_info(uct_rocm_base_agents.gpu_agents[0],
+                                    (hsa_agent_info_t)HSA_AGENT_INFO_NAME,
+                                    (void*)gfx_name);
+        if (status != HSA_STATUS_SUCCESS) {
+            ucs_debug("error in hsa_agent_info %d", status);
+            return gpu_product;
+        }
+
+        if (NULL != strstr(gfx_name, "gfx94")) {
+            /* This is an MI300 GPU, but cannot say whether its the A or X
+               variant. Assuming A variant for now*/
+            gpu_product = UCT_ROCM_AMD_GPU_MI300A;
+            ucs_trace("found gfx94* GPU, assuming MI300A");
+        } else {
+            ucs_trace("assuming MI100/MI200 GPU");
+        }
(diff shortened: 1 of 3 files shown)


edgargabriel:
@yosefe could you please have a look at this pr? I would love this pr in the next 1.16 rc if possible.

src/uct/rocm/base/rocm_base.h
```diff
@@ -13,6 +13,10 @@
 #include <hsa.h>
 #include <hsa_ext_amd.h>
 
+/* First one is used as the default value */
+#define AMD_GPU_MI200  UCS_BIT(1)
+#define AMD_GPU_MI300A UCS_BIT(2)
+#define AMD_GPU_MI300X UCS_BIT(3)
```
yosefe, line 19: 1. add UCT_ROCM_ prefix
2. why bits and not enum?

src/uct/rocm/base/rocm_base.c
```diff
@@ -501,6 +501,52 @@ ucs_status_t uct_rocm_base_get_link_type(hsa_amd_link_info_type_t *link_type)
...
+        /* In case product_name is not set correctly, query the gfx
+           architecture name */
+        status = hsa_agent_get_info(uct_rocm_base_agents.gpu_agents[0],
+                                    (hsa_agent_info_t)HSA_AGENT_INFO_NAME,
+                                    (void*)gfx_name);
+        if (status != HSA_STATUS_SUCCESS) {
+            ucs_debug("Error in hsa_agent_info %d", status);
+            return;
+        }
+        if (NULL != strstr(gfx_name, "gfx94")) {
```
yosefe, line 539: space line before

src/uct/rocm/base/rocm_base.c
```diff
@@ -501,6 +501,52 @@ ucs_status_t uct_rocm_base_get_link_type(hsa_amd_link_info_type_t *link_type)
...
+        *gpu_product = AMD_GPU_MI300X;
+        ucs_trace("found MI300X GPU");
+    } else {
+        /* In case product_name is not set correctly, query the gfx
+           architecture name */
+        status = hsa_agent_get_info(uct_rocm_base_agents.gpu_agents[0],
+                                    (hsa_agent_info_t)HSA_AGENT_INFO_NAME,
+                                    (void*)gfx_name);
+        if (status != HSA_STATUS_SUCCESS) {
+            ucs_debug("Error in hsa_agent_info %d", status);
```
yosefe, line 536: smallcase

src/uct/rocm/base/rocm_base.c
```diff
@@ -501,6 +501,52 @@ ucs_status_t uct_rocm_base_get_link_type(hsa_amd_link_info_type_t *link_type)
...
+                                        HSA_AMD_AGENT_INFO_PRODUCT_NAME,
+                                (void*)product_name);
+    if (status != HSA_STATUS_SUCCESS) {
+        ucs_debug("Error in hsa_agent_info %d", status);
+        return;
+    }
+
+    if (NULL != strstr(product_name, "MI300A")) {
+        *gpu_product = AMD_GPU_MI300A;
+        ucs_trace("found MI300A GPU");
```
yosefe, line 525: should we print the GPU index or bus id?

src/uct/rocm/base/rocm_base.c
```diff
@@ -501,6 +501,52 @@ ucs_status_t uct_rocm_base_get_link_type(hsa_amd_link_info_type_t *link_type)
     return UCS_OK;
 }
 
+void uct_rocm_base_get_gpu_product(int *gpu_product)
```
yosefe, line 504: maybe return value (instead of void) ?

src/uct/rocm/base/rocm_base.c
edgargabriel: we are not truly checking for a particular GPU, in fact GPU Id is not passed in into the function. I can hard code that we check GPU zero if you, but there is limited value in that.  We are assuming at the moment that all GPUs on a node are of the same type.

edgargabriel:
@yosefe thank you for your review, I think I addressed all your comments, except for the 'adding the GPU id' to the error message, I commented however why I think it is not truly helpful in that routine.

yosefe:
@edgargabriel pls don't force-push during a review, see https://github.com/openucx/ucx/blob/master/.github/CONTRIBUTING.md

src/uct/rocm/base/rocm_base.c
yosefe: ok, maybe document this assumption before line 511, seems hsa_agent_get_info(uct_rocm_base_agents.gpu_agents[0]) is getting info from GPU 0?

edgargabriel:
> @edgargabriel pls don't force-push during a review, see https://github.com/openucx/ucx/blob/master/.github/CONTRIBUTING.md

@yosefe ok, sorry, I did it again this time, but will hold back with squashing next until the review is complete.

src/uct/rocm/base/rocm_base.c
edgargabriel: I added a comment along those lines.

PR #9607, opened 2024-01-16
UCP/RNDV: Implement minimum chunk size per lane on proto v2

tvegas1:
## What
Implement minimum chunk size for multi-lane RNDV protos v2.

## Why ?
Not splitting leads to performance gains on smaller sizes.

## How ?
When maximum fragment size is not reached, simply make sure that we always use the remaining size or the minimal specified chunk size (default 16kB).

Tested:
- osu_mbw_mr 1:1, 16:16, RNDV thresh 0, RNDV thresh default, rock and jazz
  - UCX_MIN_RNDV_CHUNK_SIZE 64b, 16k, 256k
- osu_latency unchanged


src/ucp/rndv/proto_rndv.inl
```diff
@@ -244,19 +244,29 @@ ucp_proto_rndv_bulk_max_payload(ucp_request_t *req,
...
+    size_t min_rndv_chunk_size;
 
     if (ucs_likely(total_length < max_frag_sum)) {
-        /* Each lane sends less than its maximal fragment size */
+        /**
+         * Each lane sends less than its maximal fragment size but more than the
+         * minimal chunk size
+         */
+        min_rndv_chunk_size =
+                req->send.ep->worker->context->config.ext.min_rndv_chunk_size;
```
yosefe, line 255: maybe can do it simpler (and less overhead):
```c
if (ucs_likely(total_length <= min_rndv_chunk_size)) {
    max_payload = total_length - total_offset;
} else if (ucs_likely(total_length < max_frag_sum)) {
... (keep code as it is today)
```

brminich (commented):
i think we also need to reflect this change in perf estimations and update `ucp_proto_multi_init`, so that `perf.max_frag` takes into account mimimal rndv chunk value as a max fragment

src/ucp/rndv/proto_rndv.inl
```diff
@@ -244,19 +244,29 @@ ucp_proto_rndv_bulk_max_payload(ucp_request_t *req,
...
+                    "total_offset=%zu total_length=%zu weight_sum=%zu%% ",
+                    req, min_rndv_chunk_size, scaled_length, total_offset,
+                    total_length,
                     ucp_proto_multi_scaled_length(lpriv->weight_sum, 100));
-
         max_payload = scaled_length - total_offset;
+        if (max_payload < min_rndv_chunk_size) {
+            max_payload = ucs_min(min_rndv_chunk_size,
+                                  total_length - total_offset);
+        }
```
yosefe, line 269: max_payload=ucs_max(scaled_length, min_rndv_chunk_size)

tvegas1: fixed, kindly double check

src/ucp/rndv/proto_rndv.inl
```diff
@@ -244,19 +244,30 @@ ucp_proto_rndv_bulk_max_payload(ucp_request_t *req,
...
+                    total_length,
                     ucp_proto_multi_scaled_length(lpriv->weight_sum, 100));
-
-        max_payload = scaled_length - total_offset;
+        /**
+         * max_payload is later capped by remaining request length when
+         * advancing datatype iterator
+         */
+        max_payload = ucs_max(min_rndv_chunk_size,
+                              scaled_length - total_offset);
```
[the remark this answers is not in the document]
tvegas1: As I understand, `scaled_length` is a proportion of `total_length`. But the weight used is `weight_sum` which is the sum of all `weights` of previous lanes (works because we only use every lane at most once in this `if` block). As consequence `scaled_length` is the sum of all data sent on previous lanes plus current one, so we need to substract what has already been done.

To remove that offset, I guess we should also change:
`scaled_length = ucp_proto_multi_scaled_length(lpriv->weight_sum, total_length);`
to
`scaled_length = ucp_proto_multi_scaled_length(lpriv->weight, total_length);`

Are we sure to want to do it in this patch? I fear some rounding issues.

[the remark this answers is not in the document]
yosefe: right, scaled_length is cumulative.

yosefe (approved):

yosefe:
@tvegas1 pls squash

brminich (approved):

PR #9608, opened 2024-01-16
CI: Move Valgrind check to a separate stage

ivankochin:
## What
Moves tests execution with enabled Valgrind to the separate stage of CI.

## Why
In that scheme, if valgrind fails but tests without valgrind passed, there is no need to rerun the tests without valgrind.

### DIFF
--- a/contrib/test_jenkins.sh
+++ b/contrib/test_jenkins.sh
@@ -9,14 +9,15 @@
 #
 #
 # Environment variables set by Jenkins CI:
-#  - WORKSPACE           : path to work dir
-#  - BUILD_NUMBER        : jenkins build number
-#  - JOB_URL             : jenkins job url
-#  - EXECUTOR_NUMBER     : number of executor within the test machine
-#  - JENKINS_RUN_TESTS   : whether to run unit tests
-#  - RUN_TESTS           : same as JENKINS_RUN_TESTS, but for Azure
-#  - JENKINS_TEST_PERF   : whether to validate performance
-#  - JENKINS_NO_VALGRIND : set this to disable valgrind tests
+#  - WORKSPACE         : path to work dir
+#  - BUILD_NUMBER      : jenkins build number
+#  - JOB_URL           : jenkins job url
+#  - EXECUTOR_NUMBER   : number of executor within the test machine
+#  - JENKINS_RUN_TESTS : whether to run unit tests
+#  - RUN_TESTS         : same as JENKINS_RUN_TESTS, but for Azure
+#  - JENKINS_TEST_PERF : whether to validate performance
+#  - ASAN_CHECK        : set to enable Address Sanitizer instrumentation build
+#  - VALGRIND_CHECK    : set to enable running tests with Valgrind
 #
 # Optional environment variables (could be set by job configuration):
 #  - nworkers : number of parallel executors
@@ -35,12 +36,14 @@ if [ -z "$BUILD_NUMBER" ]; then
 	JENKINS_RUN_TESTS=yes
 	JENKINS_TEST_PERF=1
 	TIMEOUT=""
-	TIMEOUT_VALGRIND=""
 else
 	echo "Running under jenkins"
 	WS_URL=$JOB_URL/ws
-	TIMEOUT="timeout 200m"
-	TIMEOUT_VALGRIND="timeout 240m"
+	if [[ "$VALGRIND_CHECK" == "yes" ]]; then
+		TIMEOUT="timeout 240m"
+	else
+		TIMEOUT="timeout 200m"
+	fi
 fi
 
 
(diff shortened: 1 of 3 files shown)


contrib/test_jenkins.sh
```diff
@@ -1119,6 +1131,25 @@ run_asan_check() {
...
 }
 
+run_valgrind_check() {
+	if [[ $(uname -m) =~ "aarch" ]] || [[ $(uname -m) =~ "ppc" ]]
+	then
+		echo "==== Skip valgrind tests ===="
+		return
+	fi
+
+	# Load newer valgrind if naative is older than 3.10
```
tvegas1, line 1141: typo

contrib/test_jenkins.sh
```diff
@@ -1119,6 +1131,25 @@ run_asan_check() {
 	run_gtest "default"
 }
 
+run_valgrind_check() {
+	if [[ $(uname -m) =~ "aarch" ]] || [[ $(uname -m) =~ "ppc" ]]
+	then
+		echo "==== Skip valgrind tests ===="
```
tvegas1, line 1137: shall we add `uname -a` output to the echo line to provide more details? would it make sense to instead never call `run_valgrind_check` for `aarch` or `ppc`?

contrib/test_jenkins.sh
```diff
@@ -1119,6 +1131,25 @@ run_asan_check() {
...
+		return
+	fi
+
+	# Load newer valgrind if naative is older than 3.10
+	if ! (echo "valgrind-3.10.0"; valgrind --version) | sort -CV
```
tvegas1, line 1142: would `if ! { echo "valgrind-3.10.0"; valgrind --version; } | sort -CV; then ...` work, to avoid sub shell?

contrib/test_jenkins.sh
```diff
@@ -923,40 +915,60 @@ run_gtest() {
...
+#
+# Run the test suite (gtest)
+# Arguments: <compiler-name> <make-target> [configure-flags]
+#
+run_gtest_make() {
+	compiler_name=$1
+	make_target=$2
+
+	set_common_test_flags
+	set_gtest_test_flags
```
rakhmets, line 952: `unset GTEST_EXTRA_ARGS` is missing before running valgrind tests.

contrib/test_jenkins.sh
ivankochin: Seems like it works, but why is that better?

contrib/test_jenkins.sh
ivankochin: Fixed.

contrib/test_jenkins.sh
ivankochin: Added `uname -a` to output.

From my point of view this place for this check is better then [that one](https://github.com/openucx/ucx/pull/9608/files#diff-a938c4972801b864bb71586317dfb5b5eac57b6edf63694b0816154de0ab856bR1165). Or do you propose to move it somewhere else?

contrib/test_jenkins.sh
ivankochin: Good catch, fixed!

contrib/test_jenkins.sh
tvegas1: yes agree. I was thinking moving the check outside of .sh script altogether, maybe to yaml, but that might not be easy

rakhmets (approved):

contrib/test_jenkins.sh
```diff
@@ -892,71 +895,82 @@ run_malloc_hook_gtest() {
...
-		$AFFINITY $TIMEOUT_VALGRIND make -C test/gtest test_valgrind
-		module unload tools/valgrind-3.12.0
-	else
-		echo "==== Not running valgrind tests with $compiler_name compiler ===="
-	fi
-
 	unset GTEST_REPORT_LONGEST_TESTS
 	unset GTEST_TOTAL_SHARDS
 	unset GTEST_SHARD_INDEX
+
```
brminich, line 927: imo not needed

contrib/test_jenkins.sh
```diff
@@ -892,71 +895,82 @@ run_malloc_hook_gtest() {
...
 }
 
-#
-# Run the test suite (gtest)
-# Arguments: <compiler-name> [configure-flags]
-#
-run_gtest() {
-	compiler_name=$1
-
+set_common_test_flags() {
```
brminich, line 898: 3 of these flags are also gtest flags, maybe add gtest to the name

contrib/test_jenkins.sh
```diff
@@ -892,71 +895,82 @@ run_malloc_hook_gtest() {
...
 	export GTEST_UCT_TCP_FASTEST_DEV=1
 	export OMP_NUM_THREADS=4
+}
 
-	# Run specific tests
-	do_distributed_task 1 4 run_malloc_hook_gtest
-	do_distributed_task 2 4 run_gtest_watchdog_test 5 60 300
-	do_distributed_task 3 4 test_memtrack
-
+set_gtest_test_flags() {
```
brminich, line 906: you can call `set_common_test_flags` from here, since this func is never used alone

tvegas1 (approved):

contrib/test_jenkins.sh
ivankochin: Done

contrib/test_jenkins.sh
ivankochin: Done

contrib/test_jenkins.sh
ivankochin: Removed.

brminich (approved):

yosefe (approved):

PR #9613, opened 2024-01-18
GTEST/COMMON: check support for hipMallocPitch

edgargabriel:
## What
gfx94* architectures do not support hipMallocPitch, so disable them in gtest. Add a check for image support on the GPU, which is  required for hipMallocPitch.

### DIFF
--- a/test/gtest/common/mem_buffer.cc
+++ b/test/gtest/common/mem_buffer.cc
@@ -121,6 +121,24 @@ bool mem_buffer::is_rocm_managed_supported()
 #endif
 }
 
+bool mem_buffer::is_rocm_malloc_pitch_supported()
+{
+#if HAVE_ROCM
+    hipError_t ret;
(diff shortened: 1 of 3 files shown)


test/gtest/common/mem_buffer.cc
```diff
@@ -121,6 +121,31 @@ bool mem_buffer::is_rocm_managed_supported()
...
+    }
+
+    if (imageSupport == 1) {
+        /* gfx94* architectures do not support hipMallocPitch.
+           Make sure that we have image support, which is
+           required for hipMallocPitch */
+        return true;
+    }
+
+    return false;
```
yosefe, line 143: maybe just "return (imageSupport == 1)" ?

edgargabriel: done, thanks!

yosefe (approved):
pls squash

PR #9624, opened 2024-01-23
TOOLS/PERF: Second batch commit for DPU xgvmi daemon

iyastreb:
## What
Second batch commit for DPU xgvmi daemon
It contains handlers for host-to-peer and peer-to-peer messages no the daemon side.

For reference:
The first batch commit https://github.com/openucx/ucx/pull/9536 is merged.
The original changes: https://github.com/openucx/ucx/pull/8767

## Why ?

### DIFF
--- a/src/tools/perf/api/libperf.h
+++ b/src/tools/perf/api/libperf.h
@@ -98,6 +98,67 @@ enum {
 };
 
 
+/*
+ * Active message ID used for performance tests.
+ */
+#define UCP_PERF_AM_ID 1
+
+
+/*
+ *   +------+                                                        +------+
+ *   |HOST 0|-------------------UCP_PERF_AM_ID(AM)------------------>|HOST 1|
+ *   +------+        +---------+                  +---------+        +------+
+ *      |            |DPU DMN 0|                  |DPU DMN 1|            |
+ *      |            +---------+                  +---------+            |
+ *      |                 |                            |                 |
+ *      |------INIT------>|                            |<------INIT------|
+ *      |                 |----------PEER_INIT-------->|                 |
+ *      |                 |                            |                 |
+ *      |----SEND_REQ---->|                            |<--RECV_REQ(TM)--|
+ *      |                 |---------PEER_TX(AM)------->|                 |
+ *      |                 |----------(TM msg)--------->|                 |
+ *      |<---SEND_CMPL----|                            |---RECV_CMPL---->|
+ *      |                 |                            |                 |
+ *      |-------FIN------>|                            |<------FIN-------|
+ *
+ * (AM) used only when AM communication is configured
+ * (TM) used only when tag matching communication is configured
+ */
+#define UCP_FOREACH_PERF_DAEMON_AM_ID(_macro) \
+    /* 1) No DPU offloading case: this message is used to transfer data \
+     *    between hosts \
+     * 2) DPU offloading case: Sender DPU->Receiver DPU, (optional) AM message \
+     *    to transfer data between DPU peers when AM communication is used. \
+     *    For tag matching this message isn't used */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_PEER_TX,   UCP_PERF_AM_ID) \
+    /* Host->DPU daemon initialization message */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_INIT,      UCP_PERF_AM_ID + 1) \
+    /* Host->DPU daemon, initiate data transfer to the remote peer */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_SEND_REQ,  UCP_PERF_AM_ID + 2) \
+    /* DPU daemon->Host, ack the data tx operation initiated by SEND_REQ */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_SEND_CMPL, UCP_PERF_AM_ID + 3) \
+    /* Host->DPU daemon, (optional) initiate data receive. Not used when \
+     * DPU peers communicate by AM. Will be used by tag matching */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_RECV_REQ,  UCP_PERF_AM_ID + 4) \
+    /* DPU daemon->Host, ack the receive completion. This message is sent \
+     * unsolicited in case of AM communication. For tag matching transport it \
+     * confirms the completion of RECV_REQ request */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_RECV_CMPL, UCP_PERF_AM_ID + 5) \
+    /* Host->DPU daemon finalize message */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_FIN,       UCP_PERF_AM_ID + 6) \
+    /* Sender DPU->Receiver DPU, message to establish communication between \
+     * DPU peers */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_PEER_INIT, UCP_PERF_AM_ID + 7)
+
+
+#define UCP_PERF_DAEMON_ENUMIFY(ID, VALUE) ID = VALUE,
+
+
+typedef enum {
+    UCP_FOREACH_PERF_DAEMON_AM_ID(UCP_PERF_DAEMON_ENUMIFY)
+} ucp_perf_daemon_am_id_t;
+
+
 #define UCT_PERF_TEST_PARAMS_FMT             "%s/%s"
 #define UCT_PERF_TEST_PARAMS_ARG(_params)    (_params)->uct.tl_name, \
                                              (_params)->uct.dev_name
@@ -221,6 +282,8 @@ typedef struct ucx_perf_params {
         ucp_perf_datatype_t     recv_datatype;
         size_t                  am_hdr_size; /* UCP Active Message header size
                                                 (not included in message size) */
+        int                     is_daemon_mode;  /* Whether DPU offloading daemon
+                                                    is configured */
         struct sockaddr_storage dmn_local_addr;  /* IP and port of local daemon,
                                                     used to offload communication */
         struct sockaddr_storage dmn_remote_addr; /* IP and port of remote daemon,
@@ -230,9 +293,18 @@ typedef struct ucx_perf_params {
 } ucx_perf_params_t;
 
 
+typedef struct {
+    uint64_t addr;
+    uint64_t length;
+} ucp_perf_daemon_req_t;
+
+
 /* Allocators for each memory type */
 typedef struct ucx_perf_allocator ucx_perf_allocator_t;
 extern const ucx_perf_allocator_t* ucx_perf_mem_type_allocators[];
+
+
+const char *ucp_perf_daemon_am_id_name(ucp_perf_daemon_am_id_t id);
 
 
 /**
--- a/src/tools/perf/lib/libperf_int.h
+++ b/src/tools/perf/lib/libperf_int.h
@@ -109,6 +109,7 @@ struct ucx_perf_context {
             unsigned long              remote_addr;
             ucp_mem_h                  send_memh;
             ucp_mem_h                  recv_memh;
+            ucp_perf_daemon_req_t      daemon_req;
             ucp_dt_iov_t               *send_iov;
             ucp_dt_iov_t               *recv_iov;
             void                       *am_hdr;
--- a/src/tools/perf/lib/ucp_tests.cc
+++ b/src/tools/perf/lib/ucp_tests.cc
@@ -23,7 +23,7 @@ class ucp_perf_test_runner {
 public:
     typedef uint8_t psn_t;
 
-    static const unsigned AM_ID     = 1;
+    static const unsigned AM_ID     = UCP_PERF_AM_ID;
     static const ucp_tag_t TAG      = 0x1337a880u;
     static const ucp_tag_t TAG_MASK = (FLAGS & UCX_PERF_TEST_FLAG_TAG_WILDCARD) ?
                                       0 : (ucp_tag_t)-1;
@@ -47,7 +47,11 @@ public:
 
         ucs_assert_always(m_max_outstanding > 0);
 
-        set_am_handler(am_data_handler, this, UCP_AM_FLAG_WHOLE_MSG);
+        set_am_handler(AM_ID, am_data_handler, this, UCP_AM_FLAG_WHOLE_MSG);
+        set_am_handler(UCP_PERF_DAEMON_AM_ID_SEND_CMPL,
+                       am_daemon_send_ack_handler, this, UCP_AM_FLAG_WHOLE_MSG);
+        set_am_handler(UCP_PERF_DAEMON_AM_ID_RECV_CMPL,
+                       am_daemon_recv_ack_handler, this, UCP_AM_FLAG_WHOLE_MSG);
 
         if (CMD == UCX_PERF_CMD_ADD) {
             m_atomic_op = UCP_ATOMIC_OP_ADD;
@@ -64,17 +68,20 @@ public:
 
     ~ucp_perf_test_runner()
     {
-        set_am_handler(NULL, this, 0);
+        set_am_handler(UCP_PERF_DAEMON_AM_ID_RECV_CMPL, NULL, NULL, 0);
+        set_am_handler(UCP_PERF_DAEMON_AM_ID_SEND_CMPL, NULL, NULL, 0);
+        set_am_handler(AM_ID, NULL, NULL, 0);
     }
 
-    void set_am_handler(ucp_am_recv_callback_t cb, void *arg, unsigned flags)
+    void set_am_handler(unsigned id, ucp_am_recv_callback_t cb, void *arg,
+                        unsigned flags)
     {
         if (CMD == UCX_PERF_CMD_AM) {
-            ucp_am_handler_param_t param;
+            ucp_am_handler_param_t param = {};
             param.field_mask = UCP_AM_HANDLER_PARAM_FIELD_ID |
                                UCP_AM_HANDLER_PARAM_FIELD_CB |
                                UCP_AM_HANDLER_PARAM_FIELD_ARG;
-            param.id         = AM_ID;
+            param.id         = id;
             param.cb         = cb;
             param.arg        = arg;
 
@@ -326,6 +333,36 @@ public:
         return UCS_OK;
     }
 
+    static ucs_status_t
+    am_daemon_send_ack_handler(void *arg, const void *header,
+                               size_t header_length, void *data, size_t length,
+                               const ucp_am_recv_param_t *param)
+    {
+        ucp_perf_test_runner *test = (ucp_perf_test_runner*)arg;
+
+        ucs_assertv(length == 0ul, "length=%zu", length);
+        ucs_assert(!(param->recv_attr & UCP_AM_RECV_ATTR_FLAG_RNDV));
+
+        test->send_completed();
+
+        return UCS_OK;
+    }
+
+    static ucs_status_t
+    am_daemon_recv_ack_handler(void *arg, const void *header,
+                               size_t header_length, void *data, size_t length,
+                               const ucp_am_recv_param_t *param)
+    {
+        ucp_perf_test_runner *test = (ucp_perf_test_runner*)arg;
+
+        ucs_assertv(length == 0ul, "length=%zu", length);
+        ucs_assert(!(param->recv_attr & UCP_AM_RECV_ATTR_FLAG_RNDV));
+
+        test->recv_completed();
+
+        return UCS_OK;
+    }
+
     void UCS_F_ALWAYS_INLINE send_started()
     {
         ++m_sends_outstanding;
@@ -427,6 +464,28 @@ public:
         }
     }
 
+    UCS_F_ALWAYS_INLINE ucs_status_t send_daemon_req(void *buffer,
+                                                     unsigned length)
+    {
+        ucp_ep_h ep               = m_perf.ucp.ep;
+        ucp_request_param_t param = {
+            .op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS,
+            .flags        = UCP_AM_SEND_FLAG_EAGER
+        };
+        ucs_status_ptr_t req;
+
+        ucs_assertv(CMD == UCX_PERF_CMD_AM, "cmd %d", CMD);
+
+        req = ucp_am_send_nbx(ep, UCP_PERF_DAEMON_AM_ID_SEND_REQ, NULL, 0,
+                              &m_perf.ucp.daemon_req,
+                              sizeof(m_perf.ucp.daemon_req), &param);
+        if (UCS_PTR_IS_PTR(req)) {
+            ucp_request_free(req);
+        }
+
+        return UCS_PTR_STATUS(req);
+    }
+
     ucs_status_t UCS_F_ALWAYS_INLINE
     send(ucp_ep_h ep, void *buffer, unsigned length, ucp_datatype_t datatype,
          psn_t sn, uint64_t remote_addr, ucp_rkey_h rkey, bool get_info = false)
@@ -435,8 +494,14 @@ public:
                                                 &m_send_params;
         uint64_t value             = 0;
         void *request;
+        ucs_status_t status;
 
         wait_send_window(1);
+
+        if (m_perf.params.ucp.is_daemon_mode) {
+            status = send_daemon_req(buffer, length);
+            goto out;
+        }
 
         /* coverity[switch_selector_expr_is_constant] */
         switch (CMD) {
@@ -493,8 +558,11 @@ public:
             ucp_request_release(request);
         }
 
+        status = UCS_OK;
+
+    out:
         send_started();
-        return UCS_OK;
+        return status;
     }
 
     ucs_status_t UCS_F_ALWAYS_INLINE
--- a/src/tools/perf/perftest_daemon.c
+++ b/src/tools/perf/perftest_daemon.c
@@ -11,6 +11,7 @@
 #include "perftest.h"
 
 #include <ucp/api/ucp.h>
+#include <ucp/core/ucp_mm.h>
 #include <ucs/debug/log.h>
 #include <ucs/sys/sock.h>
 #include <ucs/sys/string.h>
@@ -26,21 +27,34 @@ typedef struct ucp_perf_daemon_context_t {
     ucp_context_h                      context;
     ucp_worker_h                       worker;
     ucp_listener_h                     listener;
+    sa_family_t                        ai_family;
     uint16_t                           port;
     ucp_ep_h                           host_ep;
     ucp_ep_h                           peer_ep;
     struct sockaddr_storage            peer_address;
     ucp_mem_h                          send_memh;
     ucp_mem_h                          recv_memh;
-    void                               *rx_address;
 } ucp_perf_daemon_context_t;
 
 static volatile int terminated = 0;
 
+const char *ucp_perf_daemon_am_id_name(ucp_perf_daemon_am_id_t am_id)
+{
+#define UCP_PERF_DAEMON_NAME_CASE(ID, _) \
+    case ID: \
+        return #ID;
+
+    switch (am_id) {
+        UCP_FOREACH_PERF_DAEMON_AM_ID(UCP_PERF_DAEMON_NAME_CASE);
+    default:
+        return "UNKNOWN";
+    }
+}
+
 static void
 ucp_perf_daemon_ep_close(ucp_perf_daemon_context_t *ctx, ucp_ep_h ep)
 {
-    ucp_request_param_t param;
+    ucp_request_param_t param = {};
     ucs_status_ptr_t close_req;
     ucs_status_t status;
 
@@ -81,7 +95,7 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
     ucs_status_t status;
     ucp_ep_h ep;
 
-    ep_params.field_mask      = UCP_EP_PARAM_FIELD_ERR_HANDLER  |
+    ep_params.field_mask      = UCP_EP_PARAM_FIELD_ERR_HANDLER |
                                 UCP_EP_PARAM_FIELD_CONN_REQUEST |
                                 UCP_EP_PARAM_FIELD_ERR_HANDLING_MODE;
     ep_params.err_mode        = UCP_ERR_HANDLING_MODE_PEER;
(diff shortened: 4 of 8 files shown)


src/tools/perf/lib/ucp_tests.cc
```diff
@@ -482,19 +547,20 @@ class ucp_perf_test_runner {
         default:
             return UCS_ERR_INVALID_PARAM;
         }
-
```
brminich, line 485: imo, no need to remove

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -482,19 +547,20 @@ class ucp_perf_test_runner {
...
         if (!UCS_PTR_IS_PTR(request)) {
             /* coverity[overflow] */
             return UCS_PTR_STATUS(request);
         }
-
```
brminich, line 490: imo, no need to remove

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,287 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+                                        UCP_PERF_DAEMON_AM_ID_PEER_INIT);
+}
+
+static void ucp_perf_daemon_check_am_msg(const ucp_am_recv_param_t *param,
+                                         size_t header_length,
+                                         int check_reply_ep,
+                                         int check_rndv,
+                                         ucp_perf_daemon_am_id_t am_id)
+{
+    ucs_debug("message received: %s", ucp_perf_daemon_am_id_name(am_id));
```
brminich, line 228:
```suggestion
    ucs_data("message received: %s", ucp_perf_daemon_am_id_name(am_id));
```

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,287 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+{
+    ucp_perf_daemon_context_t *ctx  = arg;
+    void *ptr                       = data;
+    void *value;
+    uint16_t value_length;
+
+    ucp_perf_daemon_check_am_msg(param, header_length, 1, 0,
+                                 UCP_PERF_DAEMON_AM_ID_HOST_INIT);
+
+    if (ctx->peer_ep != NULL) {
```
brminich, line 283: why checking `peer_ep` here?

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,287 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+    if (value_length != 0) {
+        ucp_perf_daemon_create_peer_ep(ctx, value, value_length);
+    }
+
+    ptr            = ucp_perf_unpack_length_value(ptr, &value_length, &value);
+    ctx->send_memh = ucp_perf_daemon_memh_import(ctx, value);
+    ptr            = ucp_perf_unpack_length_value(ptr, &value_length, &value);
+    ctx->recv_memh = ucp_perf_daemon_memh_import(ctx, value);
+
+    ucs_assert_always(UCS_PTR_BYTE_DIFF(data, ptr) <= length);
```
brminich, line 299: better use assertv and print the values

src/tools/perf/perftest_daemon.c
iyastreb: I borrowed this code from the initial implementation, and IMO it is correct (at least at this stage).

On startup host process connects to daemon and send HOST_INIT message to it, containing IP:port of the peer daemon and exported memory handles. So on the initial HOST_INIT the remote daemon peer is not yet known, and this is expected behavior.

What if `peer_ep` is already resolved? That could mean that this is a stale daemon process that was already configured once, and now we receive duplicate HOST_INIT message. Or it receives HOST_INIT from multiple host processes. IMO at this stage of the daemon we can just abort the process, this is what we do.

Maybe in the future we need to better handle that, any thoughts?
Like persistent daemon service, or multiple host processes supported

src/tools/perf/perftest_daemon.c
iyastreb: sure

tvegas1:
i think we need to add `ucx_perftest_daemon` to `.gitignore` file

iyastreb:
> i think we need to add `ucx_perftest_daemon` to `.gitignore` file

Right, will be added.
For me git does not complain because I build from a dedicated build-* folder (not from the WS root).

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+                                         int check_reply_ep,
+                                         int check_rndv,
+                                         ucp_perf_daemon_am_id_t am_id)
+{
+    ucs_trace_data("message received: %s", ucp_perf_daemon_am_id_name(am_id));
+
+    if (check_rndv) {
+        if (!(param->recv_attr & UCP_AM_RECV_ATTR_FLAG_RNDV)) {
+            ucs_fatal("am message received with unsupported eager protocol");
+        }
```
brminich, line 233:
```suggestion
        ucs_assertv(param->recv_attr & UCP_AM_RECV_ATTR_FLAG_RNDV,
                    "am message received with unsupported eager protocol");
```

src/tools/perf/perftest_daemon.c
brminich: ok, can you pls add a short comment here for this check?

src/tools/perf/perftest_daemon.c
iyastreb: Found better option to handle both cases:
```
    ucs_assertv(check_rndv == !!(param->recv_attr & UCP_AM_RECV_ATTR_FLAG_RNDV),
                "am message received with unsupported %s protocol",
                (check_rndv ? "eager" : "rndv"));
```

src/tools/perf/perftest_daemon.c
iyastreb: sure

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,29 @@ enum {
 };
 
 
+#define FOREACH_UCP_PERF_DAEMON_AM_ID \
+    /* Host to daemon communication */  \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_INIT,     0) \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_REQ,      1) \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_SEND_ACK, 2) \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_RECV_ACK, 3) \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_FIN,      4) \
```
brminich, line 107: btw, where do we send INIT and FIN?

iyastreb: we send them from libperf.c, which will be in the 3rd batch commit

iyastreb: It's possible, but if I bring those changes in this PR, they will come with all the remaining daemon part (allocation of memory handles, export memory) which is another 250-300LOC, and basically we don't need 3rd commit in this case. This is also an option, I let you decide

I was happy with current split, because 2nd commit concludes on daemon process side, and all the remaining changes are in host process side different files.

brminich: ok

brminich (dismissed):

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,29 @@ enum {
...
+\
+    /* Peer to peer daemon communication */ \
+    X(UCP_PERF_DAEMON_AM_ID_PEER_INIT,     5) \
+    X(UCP_PERF_DAEMON_AM_ID_PEER_OP,       6)
+
+
+typedef enum {
+#define X(ID, VALUE) ID = VALUE,
+    FOREACH_UCP_PERF_DAEMON_AM_ID
```
yosefe, line 116: can we reuse UCS_PP_FOREACH or UCS_PP_FOREACH_SEP ?

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,29 @@ enum {
...
+#undef X
+} ucp_perf_daemon_am_id_t;
+
+
+const char *ucp_perf_daemon_am_id_name(ucp_perf_daemon_am_id_t id);
```
yosefe, line 121: move this prototype below, along with other prototypes

src/tools/perf/lib/libperf_int.h
```diff
@@ -221,6 +222,12 @@ static UCS_F_ALWAYS_INLINE void ucx_perf_update(ucx_perf_context_t *perf,
     }
 }
 
+static UCS_F_ALWAYS_INLINE int
+ucp_perf_is_daemon_mode(const ucx_perf_params_t *params)
+{
+    return (0 != params->ucp.dmn_local_addr.ss_family);
```
yosefe, line 228: 1. use AF_UNSPEC instead of 0
2. maybe there could be a better way to check if it's a daemon? like some flag?

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,29 @@ enum {
 };
 
 
+#define FOREACH_UCP_PERF_DAEMON_AM_ID \
+    /* Host to daemon communication */  \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_INIT,     0) \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_REQ,      1) \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_SEND_ACK, 2) \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_RECV_ACK, 3) \
```
yosefe, line 106: why request is not separated to send/recv but ack is separated to send/recv?

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
     }
 }
 
+static void
+ucp_perf_daemon_set_am_recv_handler(ucp_perf_daemon_context_t *ctx,
+                                    ucp_perf_daemon_am_id_t am_id,
+                                    ucp_am_recv_callback_t cb)
+{
+    ucp_am_handler_param_t param = {};
```
yosefe, line 115: no need to init

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+
+    param.field_mask = UCP_AM_HANDLER_PARAM_FIELD_ID    |
+                       UCP_AM_HANDLER_PARAM_FIELD_CB    |
+                       UCP_AM_HANDLER_PARAM_FIELD_FLAGS |
+                       UCP_AM_HANDLER_PARAM_FIELD_ARG;
+    param.id         = am_id;
+    param.cb         = cb;
+    param.flags      = UCP_AM_FLAG_WHOLE_MSG;
+    param.arg        = ctx;
+    status           = ucp_worker_set_am_recv_handler(ctx->worker, &param);
```
yosefe, line 126: space line before not after

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+
+    ucs_assertv_always(status == UCS_OK, "status=%s", ucs_status_string(status));
```
yosefe, line 128: probably need better error handling flow than fatal assertion

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+}
+
+static void
+ucp_perf_daemon_send_am_eager_reply(ucp_ep_h ep, ucp_perf_daemon_am_id_t am_id)
+{
+    ucp_request_param_t param = {};
```
yosefe, line 134: no init

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+    ucs_status_ptr_t sptr;
+
+    param.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS;
+    param.flags        = UCP_AM_SEND_FLAG_REPLY | UCP_AM_SEND_FLAG_EAGER;
+
+    sptr = ucp_am_send_nbx(ep, am_id, NULL, 0ul, NULL, 0ul, &param);
```
yosefe, line 140: sport -> sreq

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+    if (UCS_PTR_IS_PTR(sptr)) {
+        ucp_request_free(sptr);
+    } else if (UCS_PTR_IS_ERR(sptr)) {
+        ucs_fatal("failed to send am id %u: %s", am_id,
+                  ucs_status_string(UCS_PTR_STATUS(sptr)));
```
yosefe, line 145: can we have better err handling?

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+static ucp_mem_h
+ucp_perf_daemon_memh_import(ucp_perf_daemon_context_t *ctx, void *packed_memh)
+{
+    ucp_mem_map_params_t params = {};
+    ucs_status_t status;
+    ucp_mem_h memh;
+
+    params.field_mask           = UCP_MEM_MAP_PARAM_FIELD_EXPORTED_MEMH_BUFFER;
+    params.exported_memh_buffer = packed_memh;
+    status                      = ucp_mem_map(ctx->context, &params, &memh);
```
yosefe, line 250: space line before

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+    }
+
+    ucp_perf_unpack_length_value(&ptr, &value_length, &value);
+    ctx->send_memh = ucp_perf_daemon_memh_import(ctx, value);
+    ucp_perf_unpack_length_value(&ptr, &value_length, &value);
+    ctx->recv_memh = ucp_perf_daemon_memh_import(ctx, value);
+
+    ucs_assertv_always(UCS_PTR_BYTE_DIFF(data, ptr) <= length,
+                       "data=%p ptr=%p length=%zu", data, ptr, length);
+
```
yosefe, line 301: remove space line

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+    params.exported_memh_buffer = packed_memh;
+    status                      = ucp_mem_map(ctx->context, &params, &memh);
+    if (status != UCS_OK) {
+        ucs_fatal("failed to import memory (%s)", ucs_status_string(status));
+    }
+
+    return memh;
+}
+
+static void
```
yosefe, line 258: pls document this function

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+    ucp_perf_daemon_req_t *req     = data;
+    ucs_status_t status;
+
+    ucp_perf_daemon_check_am_msg(param, header_length, 0, 0,
+                                 UCP_PERF_DAEMON_AM_ID_HOST_REQ);
+    ucs_assertv(length >= sizeof(*req), "length=%lu", length);
+    ucs_assert(ctx->peer_ep != NULL);
+
+    status = ucp_perf_daemon_handle_request(ctx, req);
+
```
yosefe, line 321: remove space line

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+ucp_perf_daemon_host_fin_handler(void *arg, const void *header,
+                                 size_t header_length,
+                                 void *data, size_t length,
+                                 const ucp_am_recv_param_t *param)
+{
+    ucp_perf_daemon_check_am_msg(param, header_length, 0, 0,
+                                 UCP_PERF_DAEMON_AM_ID_HOST_FIN);
+
+    terminated = 1;
+
```
yosefe, line 339: remove space line

src/tools/perf/perftest_daemon.c
```diff
@@ -96,6 +107,288 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
...
+                                  size_t length,
+                                  const ucp_am_recv_param_t *param)
+{
+    ucp_perf_daemon_context_t *ctx = arg;
+
+    ucp_perf_daemon_check_am_msg(param, header_length, 1, 0,
+                                 UCP_PERF_DAEMON_AM_ID_PEER_INIT);
+
+    ctx->peer_ep = param->reply_ep;
+
```
yosefe, line 355: remove space line

src/tools/perf/perftest_daemon.c
```diff
@@ -131,14 +424,14 @@ static ucs_status_t ucp_perf_daemon_init(ucp_perf_daemon_context_t *ctx)
...
-        goto err;
+        return status;
     }
 
     status = ucp_init(&ucp_params, config, &ctx->context);
     ucp_config_release(config);
     if (status != UCS_OK) {
         ucs_error("daemon failed to init UCP: %s", ucs_status_string(status));
-        goto err;
+        return status;
```
yosefe, line 434: why changed?

src/tools/perf/api/libperf.h
iyastreb: ok

src/tools/perf/perftest_daemon.c
iyastreb: ok

src/tools/perf/perftest_daemon.c
iyastreb: done

src/tools/perf/perftest_daemon.c
iyastreb: removed

src/tools/perf/perftest_daemon.c
iyastreb: renamed

src/tools/perf/perftest_daemon.c
iyastreb: ok

src/tools/perf/perftest_daemon.c
iyastreb: ok

src/tools/perf/perftest_daemon.c
iyastreb: done

src/tools/perf/perftest_daemon.c
iyastreb: ok

src/tools/perf/perftest_daemon.c
iyastreb: ok

src/tools/perf/perftest_daemon.c
iyastreb: To remove one extra goto label, that was used to just return the error.
IMO this way code is simpler, cause we don't need to scroll 50 lines below to see that function just returns error.

yosefe:
@iyastreb seems Coverity failure is relevant

src/tools/perf/perftest_daemon.c
yosefe: no need to align = now

src/tools/perf/perftest_daemon.c
```diff
@@ -123,26 +123,26 @@ ucp_perf_daemon_set_am_recv_handler(ucp_perf_daemon_context_t *ctx,
     param.cb         = cb;
     param.flags      = UCP_AM_FLAG_WHOLE_MSG;
     param.arg        = ctx;
-    status           = ucp_worker_set_am_recv_handler(ctx->worker, &param);
 
+    status           = ucp_worker_set_am_recv_handler(ctx->worker, &param);
```
yosefe, line 127: no need to align =

src/tools/perf/perftest_daemon.c
yosefe: our convention is that if a function has err handling flow with goto, all error flows direct to same place

yosefe (commented):
many comments are unhandled

iyastreb:
> @iyastreb seems Coverity failure is relevant

@yosefe 
Ah yes, it happens after we removed params initialisation on line 134 as you suggested.
Coverity detected that there is a potential flow where `param.datatype` is used without being explicitly initialized.
IMO the safest option would be just put back zero initialisation?
```
ucp_request_param_t param = {};
```

src/tools/perf/perftest_daemon.c
iyastreb: As discussed above, maybe we can postpone error handling until we support connection re-establishment?

src/tools/perf/api/libperf.h
yosefe:
```diff
diff --git a/src/tools/perf/api/libperf.h b/src/tools/perf/api/libperf.h
index f3150d3dd2..dfe4438586 100644
--- a/src/tools/perf/api/libperf.h
+++ b/src/tools/perf/api/libperf.h
@@ -98,23 +98,13 @@ enum {
 };


-#define FOREACH_UCP_PERF_DAEMON_AM_ID \
-    /* Host to daemon communication */  \
-    X(UCP_PERF_DAEMON_AM_ID_HOST_INIT,     0) \
-    X(UCP_PERF_DAEMON_AM_ID_HOST_REQ,      1) \
-    X(UCP_PERF_DAEMON_AM_ID_HOST_SEND_ACK, 2) \
-    X(UCP_PERF_DAEMON_AM_ID_HOST_RECV_ACK, 3) \
-    X(UCP_PERF_DAEMON_AM_ID_HOST_FIN,      4) \
-\
-    /* Peer to peer daemon communication */ \
-    X(UCP_PERF_DAEMON_AM_ID_PEER_INIT,     5) \
-    X(UCP_PERF_DAEMON_AM_ID_PEER_OP,       6)
-
+#define UCP_PERF_DAEMON_AM_IDS \
+    HOST_INIT, HOST_REQ, HOST_SEND_ACK, HOST_RECV_ACK, HOST_FIN, PEER_INIT, \
+            PEER_OP

 typedef enum {
-#define X(ID, VALUE) ID = VALUE,
-    FOREACH_UCP_PERF_DAEMON_AM_ID
-#undef X
+#define UCP_PERF_DAEMON_AM_ID_DEFINE(_, _id) UCP_PERF_DAEMON_AM_ID_##_id
+    UCS_PP_FOREACH_SEP(UCP_PERF_DAEMON_AM_ID_DEFINE, _, UCP_PERF_DAEMON_AM_IDS)
 } ucp_perf_daemon_am_id_t;


diff --git a/src/tools/perf/perftest_daemon.c b/src/tools/perf/perftest_daemon.c
index 594f34db3b..06873cb462 100644
--- a/src/tools/perf/perftest_daemon.c
+++ b/src/tools/perf/perftest_daemon.c
@@ -40,10 +40,12 @@ static volatile int terminated = 0;

 const char *ucp_perf_daemon_am_id_name(ucp_perf_daemon_am_id_t am_id)
 {
+#define UCP_PERF_DAEMON_AM_ID_STR(_, _id) \
+    case UCP_PERF_DAEMON_AM_ID_##_id: \
+        return #_id;
+
     switch (am_id) {
-#define X(ID, _) case ID: return #ID;
-        FOREACH_UCP_PERF_DAEMON_AM_ID
-#undef X
+    UCS_PP_FOREACH(UCP_PERF_DAEMON_AM_ID_STR, _, UCP_PERF_DAEMON_AM_IDS)
         default: return "UNKNOWN";
     }
 }
```

src/tools/perf/perftest_daemon.c
yosefe: can it be error message without fatal? and propagate the error to main loop, event without proper handling?
fatal error can create core dump, backtrace, etc, and not sure this is the right way here

src/tools/perf/api/libperf.h
yosefe: yes, i think we need better names, something that looks like (name) and (name)_ACK

src/tools/perf/api/libperf.h
iyastreb: Ok, this is exactly what I was talking about)
In this diff you removed the id to make it suitable for `UCS_PP_FOREACH`.
Ok, that works, but the idea of X-macro is exactly the opposite: it is to group several related items in a single place, that's the whole point.

For example, for each message type we know in advance whether to expect reply-ep or rndv. So we can extend our X-macro to enforce this behavior:
```diff
+/* NAME, ID, REPLY-EP, RNDV */                                                                                                                                                                                                    
 #define FOREACH_UCP_PERF_DAEMON_AM_ID \                                                                                                                                                                                           
     /* Host to daemon communication */  \                                                                                                                                                                                         
-    X(UCP_PERF_DAEMON_AM_ID_HOST_INIT,     0) \                                                                                                                                                                                   
-    X(UCP_PERF_DAEMON_AM_ID_HOST_REQ,      1) \                                                                                                                                                                                   
-    X(UCP_PERF_DAEMON_AM_ID_HOST_SEND_ACK, 2) \                                                                                                                                                                                   
-    X(UCP_PERF_DAEMON_AM_ID_HOST_RECV_ACK, 3) \                                                                                                                                                                                   
-    X(UCP_PERF_DAEMON_AM_ID_HOST_FIN,      4) \                                                                                                                                                                                   
+    X(UCP_PERF_DAEMON_AM_ID_HOST_INIT,     0, 1, 0) \                                                                                                                                                                             
+    X(UCP_PERF_DAEMON_AM_ID_HOST_REQ,      1, 0, 0) \                                                                                                                                                                             
+    X(UCP_PERF_DAEMON_AM_ID_HOST_SEND_ACK, 2, 0, 0) \                                                                                                                                                                             
+    X(UCP_PERF_DAEMON_AM_ID_HOST_RECV_ACK, 3, 0, 0) \                                                                                                                                                                             
+    X(UCP_PERF_DAEMON_AM_ID_HOST_FIN,      4, 0, 0) \                                                                                                                                                                             
 \                                                                                                                                                                                                                                 
     /* Peer to peer daemon communication */ \                                                                                                                                                                                     
-    X(UCP_PERF_DAEMON_AM_ID_PEER_INIT,     5) \                                                                                                                                                                                   
-    X(UCP_PERF_DAEMON_AM_ID_PEER_OP,       6)                                                                                                                                                                                     
+    X(UCP_PERF_DAEMON_AM_ID_PEER_INIT,     5, 1, 0) \                                                                                                                                                                             
+    X(UCP_PERF_DAEMON_AM_ID_PEER_OP,       6, 0, 1)                                                                                                                                                                               
                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                   
 typedef enum {                                                                                                                                                                                                                    
-#define X(ID, VALUE) ID = VALUE,                                                                                                                                                                                                  
+#define X(ID, VALUE, REPLY_EP, RNDV) ID = VALUE,                                                                                                                                                                                  
     FOREACH_UCP_PERF_DAEMON_AM_ID                                                                                                                                                                                                 
 #undef X                                                                                                                                                                                                                          
 } ucp_perf_daemon_am_id_t;                                                                                                                                                                                                        

diff --git a/src/tools/perf/perftest_daemon.c b/src/tools/perf/perftest_daemon.c                                                                                                                                                   
index 594f34db3..293fa1e25 100644                                                                                                                                                                                                  
--- a/src/tools/perf/perftest_daemon.c                                                                                                                                                                                             
+++ b/src/tools/perf/perftest_daemon.c                                                                                                                                                                                             
@@ -38,10 +38,25 @@ typedef struct ucp_perf_daemon_context_t {                                                                                                                                                                     
                                                                                                                                                                                                                                   
 static volatile int terminated = 0;                                                                                                                                                                                               
                                                                                                                                                                                                                                   
+static int ucp_perf_daemon_am_id_reply_ep[] = {                                                                                                                                                                                   
+#define X(ID, VALUE, REPLY_EP, RNDV) \                                                                                                                                                                                            
+    [ID] = REPLY_EP,                                                                                                                                                                                                              
+#undef X                                                                                                                                                                                                                          
+};                                                     
+                                                       
+static int ucp_perf_daemon_am_id_rndv[] = {
+#define X(ID, VALUE, REPLY_EP, RNDV) \
+    [ID] = RNDV,                                       
+#undef X                                               
+};                                                     
+                                                       
@@ -221,10 +236,11 @@ static void ucp_perf_daemon_create_peer_ep(ucp_perf_daemon_context_t *ctx,                 
                                                        
 static void ucp_perf_daemon_check_am_msg(const ucp_am_recv_param_t *param,                                      
                                          size_t header_length,                                                  
-                                         int check_reply_ep,                                                    
-                                         int check_rndv,                                                        
                                          ucp_perf_daemon_am_id_t am_id)                                         
 {                                                      
+    int check_rndv     = ucp_perf_daemon_am_id_rndv[am_id];                                                     
+    int check_reply_ep = ucp_perf_daemon_am_id_reply_ep[am_id];                                                 
+                                                       
     ucs_trace_data("message received: %s", ucp_perf_daemon_am_id_name(am_id));                                  
                                                        
     ucs_assertv(check_rndv == !!(param->recv_attr & UCP_AM_RECV_ATTR_FLAG_RNDV),                                
@@ -283,7 +299,7 @@ ucp_perf_daemon_host_init_handler(void *arg, const void *header,                             
     void *value;                                       
     uint16_t value_length;                             
                                                        
-    ucp_perf_daemon_check_am_msg(param, header_length, 1, 0,                                                    
+    ucp_perf_daemon_check_am_msg(param, header_length,
                                  UCP_PERF_DAEMON_AM_ID_HOST_INIT);                                              
                                                        
     /* Peer EP must not be initialized on receiving HOST_INIT message.                                          
@@ -321,7 +337,7 @@ ucp_perf_daemon_host_req_handler(void *arg, const void *header,                              
     ucp_perf_daemon_req_t *req     = data;
     ucs_status_t status;                               
                                                        
-    ucp_perf_daemon_check_am_msg(param, header_length, 0, 0,                                                    
+    ucp_perf_daemon_check_am_msg(param, header_length,
                                  UCP_PERF_DAEMON_AM_ID_HOST_REQ);                                               
     ucs_assertv(length >= sizeof(*req), "length=%lu", length);                                                  
     ucs_assert(ctx->peer_ep != NULL);
```
So the advantage of X-macro is that you can easily extend your "enum" with more related fields. That's why using x-macro in enum declaration is so widely used practise.

Ok, that was the last attempt, if you really have strong opinion I will apply your diff

yosefe: Can you pls give example of places using X macro?
Also, are you planning to use X macro as described above (with additional fields) in subsequent PRs?

src/tools/perf/perftest_daemon.c
iyastreb: ok, will be fixed

src/tools/perf/api/libperf.h
iyastreb: Ok, after live discussion the agreement is to use X-macro flavour with explicit argument:
```
#define UCP_FOREACH_PERF_DAEMON_AM_ID(X) \
    X(UCP_PERF_DAEMON_AM_ID_HOST_INIT,     0) \
```

src/tools/perf/api/libperf.h
iyastreb: Following our conversation, I updated all the ID names and improved the documentation, please check

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -23,7 +23,7 @@ class ucp_perf_test_runner {
 public:
     typedef uint8_t psn_t;
 
-    static const unsigned AM_ID     = 1;
+    static const unsigned AM_ID     = UCP_PERF_DAEMON_AM_ID_PEER_TX;
```
brminich, line 26: it's a bit weird because daemon is not relevant in this flow

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,60 @@ enum {
...
+ *      |<--HOST_SEND_ACK-|                            |-HOST_RECV_ACK-->|
+ *      |                 |                            |                 |
+ *      |----HOST_FIN---->|                            |<----HOST_FIN----|
+ *
+ * (AM) used only when AM communication is configured
+ * (T) used only when tag matching communication is configured
+ */
+#define UCP_FOREACH_PERF_DAEMON_AM_ID(X) \
+    /* Host->DPU daemon AM initialization message */ \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_INIT,     0) \
```
brminich, line 122: maybe better avoid `HOST` in names, because it is actually a daemon who is being initialized.

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,60 @@ enum {
...
+    X(UCP_PERF_DAEMON_AM_ID_HOST_SEND,     1) \
+    /* DPU daemon->Host AM to ack the data tx operation initiated by HOST_SEND */ \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_SEND_ACK, 2) \
+    /* Host->DPU daemon, (optional) AM to initiate data receive. Not used when \
+     * DPU peers communicate by AM. Will be used by tag matching */ \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_RECV,     3) \
+    /* DPU daemon->Host AM to ack the receive completion. This message is sent \
+     * unsolicited in case of AM communication. For tag matching transport it \
+     * confirms the completion of HOST_RECV request */ \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_RECV_ACK, 4) \
```
brminich, line 133: maybe:

```suggestion
    X(UCP_PERF_DAEMON_AM_ID_SEND_REQ,     1) \
    /* DPU daemon->Host AM to ack the data tx operation initiated by HOST_SEND */ \
    X(UCP_PERF_DAEMON_AM_ID_SEND_ACK, 2) \
    /* Host->DPU daemon, (optional) AM to initiate data receive. Not used when \
     * DPU peers communicate by AM. Will be used by tag matching */ \
    X(UCP_PERF_DAEMON_AM_ID_RECV_REQ,     3) \
    /* DPU daemon->Host AM to ack the receive completion. This message is sent \
     * unsolicited in case of AM communication. For tag matching transport it \
     * confirms the completion of HOST_RECV request */ \
    X(UCP_PERF_DAEMON_AM_ID_RECV_ACK, 4) \
```

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,60 @@ enum {
...
+    /* Host->DPU daemon AM finalize message */ \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_FIN,      5) \
```
brminich, line 135: maybe:

```suggestion
    X(UCP_PERF_DAEMON_AM_ID_DAEMON_FIN,      5) \
```

src/tools/perf/api/libperf.h
```diff
@@ -98,29 +98,60 @@ enum {
...
-    /* Peer to peer daemon communication */ \
-    X(UCP_PERF_DAEMON_AM_ID_PEER_INIT,     5) \
-    X(UCP_PERF_DAEMON_AM_ID_PEER_OP,       6)
+    /* Host->DPU daemon, (optional) AM to initiate data receive. Not used when \
+     * DPU peers communicate by AM. Will be used by tag matching */ \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_RECV,     3) \
+    /* DPU daemon->Host AM to ack the receive completion. This message is sent \
+     * unsolicited in case of AM communication. For tag matching transport it \
+     * confirms the completion of HOST_RECV request */ \
+    X(UCP_PERF_DAEMON_AM_ID_HOST_RECV_ACK, 4) \
```
yosefe, line 133: maybe rename 
UCP_PERF_DAEMON_AM_ID_HOST_SEND_ACK -> UCP_PERF_DAEMON_AM_ID_HOST_SEND_CMPL
UCP_PERF_DAEMON_AM_ID_HOST_RECV_ACK -> UCP_PERF_DAEMON_AM_ID_HOST_RECV_CMPL

src/tools/perf/api/libperf.h
```diff
@@ -98,29 +98,60 @@ enum {
...
+ *      |---HOST_SEND---->|                            |<--HOST_RECV(T)--|
+ *      |                 |--------PEER_TX(AM)-------->|                 |
+ *      |<--HOST_SEND_ACK-|                            |-HOST_RECV_ACK-->|
+ *      |                 |                            |                 |
+ *      |----HOST_FIN---->|                            |<----HOST_FIN----|
+ *
+ * (AM) used only when AM communication is configured
+ * (T) used only when tag matching communication is configured
+ */
+#define UCP_FOREACH_PERF_DAEMON_AM_ID(X) \
```
yosefe, line 120: rename X to _macro

src/tools/perf/lib/libperf_int.h
```diff
@@ -225,7 +225,7 @@ static UCS_F_ALWAYS_INLINE void ucx_perf_update(ucx_perf_context_t *perf,
 static UCS_F_ALWAYS_INLINE int
 ucp_perf_is_daemon_mode(const ucx_perf_params_t *params)
 {
-    return (0 != params->ucp.dmn_local_addr.ss_family);
+    return params->ucp.is_daemon_mode;
```
yosefe, line 228: can remove the wrapper ucp_perf_is_daemon_mode now

src/tools/perf/perftest_daemon.c
```diff
@@ -123,27 +124,35 @@ ucp_perf_daemon_set_am_recv_handler(ucp_perf_daemon_context_t *ctx,
...
-    } else if (UCS_PTR_IS_ERR(sptr)) {
-        ucs_fatal("failed to send am id %u: %s", am_id,
-                  ucs_status_string(UCS_PTR_STATUS(sptr)));
+    sreq = ucp_am_send_nbx(ep, am_id, NULL, 0ul, NULL, 0ul, &param);
+    if (UCS_PTR_IS_PTR(sreq)) {
+        ucp_request_free(sreq);
+    } else if (UCS_PTR_IS_ERR(sreq)) {
+        ucs_error("failed to send am id %u: %s", am_id,
+                  ucs_status_string(UCS_PTR_STATUS(sreq)));
+        return UCS_PTR_STATUS(sreq);
```
yosefe, line 152: :+1:

src/tools/perf/perftest_daemon.c
```diff
@@ -247,14 +257,25 @@ ucp_perf_daemon_memh_import(ucp_perf_daemon_context_t *ctx, void *packed_memh)
...
+        ucs_error("failed to import memory (%s)", ucs_status_string(status));
+        return NULL;
     }
 
     return memh;
 }
 
+/*
+ * From the given buffer unpack an array of bytes.
+ * This array is encoded as 2 values: length of the array as uint16_t, followed
```
yosefe, line 272: maybe uint64_t instead of uint16_t?

src/tools/perf/perftest_daemon.c
```diff
@@ -247,14 +257,25 @@ ucp_perf_daemon_memh_import(ucp_perf_daemon_context_t *ctx, void *packed_memh)
...
-        ucs_fatal("failed to import memory (%s)", ucs_status_string(status));
+        ucs_error("failed to import memory (%s)", ucs_status_string(status));
+        return NULL;
     }
 
     return memh;
 }
 
+/*
+ * From the given buffer unpack an array of bytes.
```
yosefe, line 271: Unpack array length and array contents from a buffer.

rename the function to ucp_perf_unpack_array
maybe return ucs_array_t?

src/tools/perf/perftest_daemon.c
```diff
@@ -408,6 +430,32 @@ static void ucp_perf_daemon_cleanup(ucp_perf_daemon_context_t *ctx)
...
 
+static ucs_status_t
+ucp_perf_daemon_set_am_handlers(ucp_perf_daemon_context_t *ctx)
+{
+    struct {
+        ucp_perf_daemon_am_id_t id;
+        ucp_am_recv_callback_t  cb;
+    } *ptr, handlers[] = {
+        { UCP_PERF_DAEMON_AM_ID_HOST_INIT, ucp_perf_daemon_host_init_handler },
+        { UCP_PERF_DAEMON_AM_ID_HOST_SEND, ucp_perf_daemon_host_send_handler  },
```
yosefe, line 441: extra space before }

src/tools/perf/perftest_daemon.c
```diff
@@ -408,6 +430,32 @@ static void ucp_perf_daemon_cleanup(ucp_perf_daemon_context_t *ctx)
...
+        { UCP_PERF_DAEMON_AM_ID_HOST_FIN,  ucp_perf_daemon_host_fin_handler  },
+        { UCP_PERF_DAEMON_AM_ID_PEER_INIT, ucp_perf_daemon_peer_init_handler },
+        { UCP_PERF_DAEMON_AM_ID_PEER_TX,   ucp_perf_daemon_peer_tx_handler   },
+        {}
+    };
+    ucs_status_t status;
+
+    for (ptr = handlers; NULL != ptr->cb; ++ptr) {
```
yosefe, line 449: we could avoid extra NULL ptr and use ucs_static_array_size

src/tools/perf/perftest_daemon.c
iyastreb: Ok for documentation and rename.

For `ucs_array_t`- I hesitate a bit, because it will also require to make symmetric change in the "pack" counterpart (which will be part of batch commit 3). Currently on the host side we have `void *, size_t` pair representing a memory block, which we get from the existing function:
```
ucp_memh_pack(ucp_mem_h memh, const ucp_memh_pack_params_t *params,
              void **buffer_p, size_t *buffer_size_p)
```
And the result of that function we store inside the new structure:
```
typedef struct {
    void   *address;
    size_t length;
} ucx_perf_exported_mem_t;
```
So yes, technically this is an array. And if we want to replace it with `ucs_array`, then the code will be a bit more verbose:
```
UCS_ARRAY_DECLARE_TYPE(ucs_byte_array, uint64_t, uint8_t);

static void
ucp_perf_unpack_byte_array(void **ptr, ucs_byte_array *array)
{
    array->length = *ucs_serialize_next(ptr, uint64_t);
    array->buffer = (length != 0) ?
                        ucs_serialize_next_raw(ptr, void, array->length) : NULL;
    array->capacity = array->length;
    array->is_fixed = 1;
}

static void
ucp_perf_pack_byte_array(void **ptr, const ucs_byte_array *array)
{
    *ucs_serialize_next(ptr, uint64_t) = array->length;
    if (array->length > 0) {
        memcpy(ucs_serialize_next_raw(ptr, void, array->length),
               array->buffer, array->length);
    }
}

static size_t ucp_perf_pack_daemon_info(const ucx_perf_context_t *perf,
                                        void *buffer, size_t buffer_size,
                                        const struct sockaddr_storage *address)
{
    void *p               = buffer;
    size_t address_length = (address != NULL) ? sizeof(*address) : 0;
    ucs_byte_array addr_bytes = UCS_ARRAY_FIXED_INITIALIZER(address, address_length);

    ucp_perf_pack_byte_array(&p, &addr_bytes);
    ucp_perf_pack_byte_array(&p, &perf->ucp.send_exported_mem);
    ucp_perf_pack_byte_array(&p, &perf->ucp.recv_exported_mem);
```
So in the pack side we need to perform conversion from `void *, size_t` into `byte_array` in 2 places.
Plus we need to store one extra field for `capacity`.
That's why I'm not 100% sure about this change, but I'm not against it either

If you like this approach - please let me know, I'll put in in place

src/tools/perf/lib/libperf_int.h
iyastreb: sure

src/tools/perf/api/libperf.h
iyastreb: done

src/tools/perf/api/libperf.h
iyastreb: ok

src/tools/perf/api/libperf.h
iyastreb: Well, since you proposed to exclude HOST from enum IDs, I would keep maybe `UCP_PERF_DAEMON_AM_ID_FIN`?
Otherwise DAEMON appears twice in a single ID

src/tools/perf/api/libperf.h
iyastreb: ok

src/tools/perf/api/libperf.h
iyastreb: Ok, I will add REQ

src/tools/perf/lib/ucp_tests.cc
iyastreb: Yes, naming is a bit confusing here..
If I get it right, Yossi's proposal was to use the same AM ID both for DPU-to-DPU communication and for host-to-host one. So the easiest option I found is just to remove the `DAEMON` part from that enum ID, so that it is now: `UCP_PERF_AM_ID_PEER_TX`
Please let me know if it fits

iyastreb:
@yosefe please review

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,61 @@ enum {
...
+    /* Host->DPU daemon, initiate data transfer to the remote peer */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_SEND_REQ,      2) \
+    /* DPU daemon->Host, ack the data tx operation initiated by HOST_SEND */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_SEND_REQ_CMPL, 3) \
+    /* Host->DPU daemon, (optional) initiate data receive. Not used when \
+     * DPU peers communicate by AM. Will be used by tag matching */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_RECV_REQ,      4) \
+    /* DPU daemon->Host, ack the receive completion. This message is sent \
+     * unsolicited in case of AM communication. For tag matching transport it \
+     * confirms the completion of HOST_RECV request */ \
```
brminich, line 139:
```suggestion
     * confirms the completion of _RECV_REQ request */ \
```

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,61 @@ enum {
...
+     *    between hosts \
+     * 2) DPU offloading case: Sender DPU->Receiver DPU, (optional) AM message \
+     *    to transfer data between DPU peers when AM communication is used. \
+     *    For tag matching this message isn't used */ \
+    _macro(UCP_PERF_AM_ID_PEER_TX,              0) \
+    /* Host->DPU daemon initialization message */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_INIT,          1) \
+    /* Host->DPU daemon, initiate data transfer to the remote peer */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_SEND_REQ,      2) \
+    /* DPU daemon->Host, ack the data tx operation initiated by HOST_SEND */ \
```
brminich, line 132:
```suggestion
    /* DPU daemon->Host, ack the data tx operation initiated by SEND_REQ */ \
```

src/tools/perf/api/libperf.h
```diff
@@ -98,6 +98,61 @@ enum {
...
+ * (AM) used only when AM communication is configured
+ * (TM) used only when tag matching communication is configured
+ */
+#define UCP_FOREACH_PERF_DAEMON_AM_ID(_macro) \
+    /* 1) No DPU offloading case: this message is used to transfer data \
+     *    between hosts \
+     * 2) DPU offloading case: Sender DPU->Receiver DPU, (optional) AM message \
+     *    to transfer data between DPU peers when AM communication is used. \
+     *    For tag matching this message isn't used */ \
+    _macro(UCP_PERF_AM_ID_PEER_TX,              0) \
```
brminich, line 127: maybe better to define some general `UCP_PERF_AM_ID` in this file, and then start this enum from that value + 1?

src/tools/perf/api/libperf.h
```diff
@@ -124,25 +130,25 @@ enum {
      * 2) DPU offloading case: Sender DPU->Receiver DPU, (optional) AM message \
      *    to transfer data between DPU peers when AM communication is used. \
      *    For tag matching this message isn't used */ \
-    _macro(UCP_PERF_AM_ID_PEER_TX,              0) \
+    _macro(UCP_PERF_DAEMON_AM_ID_PEER_TX,       UCP_PERF_AM_ID) \
```
brminich, line 133: maybe no need to define now and can always use `UCP_PERF_AM_ID`?

iyastreb: IMO it's better to have it inside enum for consistency:
 - to uniformly treat all the IDs (including ID to string conversion) by ucp_perf_daemon_am_id_t
 - to enforce distinct values (it's indirectly enforced by switch in ucp_perf_daemon_am_id_name)

src/tools/perf/api/libperf.h
```diff
@@ -98,50 +98,57 @@ enum {
...
      *    to transfer data between DPU peers when AM communication is used. \
      *    For tag matching this message isn't used */ \
-    X(UCP_PERF_DAEMON_AM_ID_PEER_TX,       7)
+    _macro(UCP_PERF_DAEMON_AM_ID_PEER_TX,       UCP_PERF_AM_ID) \
+    /* Host->DPU daemon initialization message */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_INIT,          UCP_PERF_AM_ID + 1) \
+    /* Host->DPU daemon, initiate data transfer to the remote peer */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_SEND_REQ,      UCP_PERF_AM_ID + 2) \
+    /* DPU daemon->Host, ack the data tx operation initiated by SEND_REQ */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_SEND_REQ_CMPL, UCP_PERF_AM_ID + 3) \
```
yosefe, line 139: UCP_PERF_DAEMON_AM_ID_SEND_CMPL

src/tools/perf/api/libperf.h
```diff
@@ -98,50 +98,57 @@ enum {
...
+    /* Host->DPU daemon, (optional) initiate data receive. Not used when \
+     * DPU peers communicate by AM. Will be used by tag matching */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_RECV_REQ,      UCP_PERF_AM_ID + 4) \
+    /* DPU daemon->Host, ack the receive completion. This message is sent \
+     * unsolicited in case of AM communication. For tag matching transport it \
+     * confirms the completion of RECV_REQ request */ \
+    _macro(UCP_PERF_DAEMON_AM_ID_RECV_REQ_CMPL, UCP_PERF_AM_ID + 5) \
```
yosefe, line 146: UCP_PERF_DAEMON_AM_ID_RECV_CMPL

yosefe (approved):

brminich (dismissed):

yosefe (approved):

yosefe:
@iyastreb I noticed there are multiple relevant clang-format errors. can you pls fix?

yosefe (changes requested):
fix clang

iyastreb:
> @iyastreb I noticed there are multiple relevant clang-format errors. can you pls fix?

@yosefe Ok, fixed 95% of them that make sense to fix
clang-format still compains in a couple of places, but IMO it's not worth changing

yosefe (approved):

PR #9625, opened 2024-01-23
CONFIG/M4: Disable MAD support to prevent libibmad/libumad dependency leak

tvegas1:
## What
Fix dependency leak from rpm build machine, when `libibmad-devel` and `libumad-devel` are installed.

## Why ?
RPM users might not have those libs installed. Valid also for non-rpm package builders.

## How ?
Force disable in v1.16, later can be default guess (or rather yes for more predictability), then `dlopen()` at runtime.

### DIFF
--- a/config/m4/mad.m4
+++ b/config/m4/mad.m4
@@ -9,9 +9,9 @@
 #
 AC_ARG_WITH([mad],
             [AS_HELP_STRING([--with-mad=(DIR)],
-                [Enable Infiniband MAD support (default is guess).])],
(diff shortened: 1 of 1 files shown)


tvegas1:
disable for now, once sanity passed and approved, will backport in 1.16

yosefe:
IMO, this is not enough - the issue can happen also without RPM installation (with local installation such as HPC-X). We need to disable MAD usage in configure script

yosefe (approved):
pls squash and port to v1.16.x

yosefe (approved):

PR #9628, opened 2024-01-24
TOOLS/PERF: dlopen MAD related libibmad/libibumad

tvegas1:
## What
Build with headers, and dlopen() at runtime.

## Why ?
Not all machines have that libibmad and/or libibumad installed.

## How ?
Use typeof to track function type from header, and keep actual code out of macros.

### Tested
- lid/giud running modes, tested missing lib, tested enabling debug (for extern ibdebug).
- ldd/readelf -a do not show dependency


tvegas1:
as per @brminich suggestion, will investigate other approach, by introducing `libperftest_mad.so` still built with `-libmad -libumad` and dlopened (likely leveraging `ucs_module_init`).

tvegas1:
updated as #9637, now implemented as a separate perftest module.

PR #9632, opened 2024-01-25
UCP: Fix rndv put protocol

brminich:
## What
Receive completion may be invoked before all the data arrives. This can happen because uct_ep_flush does not guarantee remote completion. Thus, atp can be sent via some SM transport, while data still traversing PCIe. Using UCT_FLUSH_FLAG_REMOTE leads to UCP worker address increase (+2 bytes for each dc tl and +3 bytes for each rc tl), so is not an option.

## Why ?
Fixes [RM3720250](https://redmine.mellanox.com/issues/3720250) (Nvidia internal link)


src/ucp/rndv/rndv_put.c
```diff
@@ -273,26 +274,37 @@ ucp_proto_rndv_put_common_init(const ucp_proto_init_params_t *init_params,
...
+        if (((iface_attr->cap.flags & UCT_IFACE_FLAG_AM_SHORT) &&
+             (iface_attr->cap.am.max_short >= atp_size)) ||
+            ((iface_attr->cap.flags & UCT_IFACE_FLAG_AM_BCOPY) &&
+             (iface_attr->cap.am.max_bcopy >= atp_size))) {
+            atp_map |= UCS_BIT(lane);
+        }
     }
 
+    /* Use fence only if all lanes support sending ATP */
+    use_fence &= (rpriv->bulk.mpriv.lane_map == atp_map);
```
yosefe, line 296: maybe move the logic of line 279 into this line

src/ucp/rndv/rndv_put.c
```diff
@@ -273,26 +274,37 @@ ucp_proto_rndv_put_common_init(const ucp_proto_init_params_t *init_params,
...
     rpriv     = params.super.super.priv;
     send_atp  = !ucp_proto_rndv_init_params_is_ppln_frag(init_params);
     use_fence = send_atp && !context->config.ext.rndv_put_force_flush;
 
-    /* Check if all potential lanes support sending ATP */
+    /* Check which lanes support sending ATP */
     lane_idx  = 0;
-    while (use_fence && (lane_idx < rpriv->bulk.mpriv.num_lanes)) {
+    atp_map   = 0;
+    while (lane_idx < rpriv->bulk.mpriv.num_lanes) {
```
yosefe, line 284: this can be converted for for loop now

src/ucp/rndv/rndv_put.c
```diff
@@ -274,14 +274,12 @@ ucp_proto_rndv_put_common_init(const ucp_proto_init_params_t *init_params,
...
-    use_fence = send_atp && !context->config.ext.rndv_put_force_flush;
+    rpriv    = params.super.super.priv;
+    send_atp = !ucp_proto_rndv_init_params_is_ppln_frag(init_params);
 
     /* Check which lanes support sending ATP */
-    lane_idx  = 0;
-    atp_map   = 0;
-    while (lane_idx < rpriv->bulk.mpriv.num_lanes) {
+    for (lane_idx = 0, atp_map = 0; lane_idx < rpriv->bulk.mpriv.num_lanes;
+         ++lane_idx) {
```
yosefe, line 282: remove ++ from next line
init atp_map before the loop

src/ucp/rndv/rndv_put.c
```diff
@@ -273,26 +274,37 @@ ucp_proto_rndv_put_common_init(const ucp_proto_init_params_t *init_params,
...
-    use_fence = send_atp && !context->config.ext.rndv_put_force_flush;
+    rpriv    = params.super.super.priv;
+    send_atp = !ucp_proto_rndv_init_params_is_ppln_frag(init_params);
 
-    /* Check if all potential lanes support sending ATP */
-    lane_idx  = 0;
-    while (use_fence && (lane_idx < rpriv->bulk.mpriv.num_lanes)) {
+    /* Check which lanes support sending ATP */
+    atp_map = 0;
+    for (lane_idx = 0; lane_idx < rpriv->bulk.mpriv.num_lanes; ++lane_idx) {
```
yosefe, line 282: remove ++ from next line

yosefe (approved):

PR #9638, opened 2024-01-27
UCT/CUDA: Don't register memory without active context - v1.16.x

yosefe:
## Why
Fix Cuda host memory registration error (from ucp_mem_map) when there is no active context.
For example, ails with shmem applications, like so:
```
[r-hpc-gpu08:62403:0]    cuda_copy_md.c:173  UCX  ERROR cudaHostRegister() failed: invalid argument
[r-hpc-gpu08:62403:0]          ucp_mm.c:70   UCX  ERROR failed to register address 0xff000000 (host) length 11202985984 on md[3]=cuda_cpy: Input/output error (md supports: host|cuda|cuda-managed)
```

Fixes internal issue [3748762](https://redmine.mellanox.com/issues/3748762)

## Note
The fix is needed only on v1.16.x branch, since on master branch it was already fixed by #8590

### DIFF
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.c
@@ -158,6 +158,11 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_reg,
     CUresult result;
     ucs_status_t status;
 
+    if (!uct_cuda_base_is_context_active()) {
+        ucs_debug("attempt to register memory without active context");
+        return uct_md_dummy_mem_reg(md, address, length, params, memh_p);
+    }
+
     result = cuPointerGetAttribute(&memType, CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
                                    (CUdeviceptr)(address));
     if ((result == CUDA_SUCCESS) && ((memType == CU_MEMORYTYPE_HOST)    ||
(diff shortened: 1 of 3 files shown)


yosefe:
@brminich @rakhmets can you pls review?

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -158,6 +158,11 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_mem_reg,
     CUresult result;
     ucs_status_t status;
 
+    if (!uct_cuda_base_is_context_active()) {
+        ucs_debug("attempt to register memory without active context");
+        return uct_md_dummy_mem_reg(md, address, length, params, memh_p);
+    }
```
brminich, line 164: does it fix the problem whith that shmem app (i guess with perf penalty then)?
btw, why did we remove it in 1.16  - i see the same code is present  in master

rakhmets: It should fix the issue. UCX tries to register host memory using all available MDs. It was agreed that we should do dummy memory registration when there is no active CUDA context (aka CUDA device is not set).
1.16 branch was created before https://github.com/openucx/ucx/pull/8590 was merged to master branch.

rakhmets (approved):

brminich (approved):

PR #9643, opened 2024-01-29
TEST/APPS: Modify test_fuzzy_match.py and test_ucx_tls.py to use default python version

shasson5:
## What
Fixed test_fuzzy_match.py and test_ucx_tls.py python version reference.

## Why ?
Default python for RHEL8 is python3.

### DIFF
--- a/test/apps/test_fuzzy_match.py
+++ b/test/apps/test_fuzzy_match.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/env python
 
 #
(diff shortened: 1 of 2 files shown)


gleon99:
@shasson5  also in `test_ucx_tls.py`?

test/apps/test_fuzzy_match.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/python
```
yosefe, line 1: `#!/usr/bin/env python`

yosefe (approved):
pls squash

PR #9645, opened 2024-01-30
UCT/CUDA_COPY: add multi-device support in cuda_copy

Akshay-Venkatesh:
## What/Why?
Allow a single UCP context to handle multiple CUDA devices for cuda_copy transport. This enables use cases under Legion/Realm, OpenACC, and MPI workloads that prefer 1:N process-to-GPU mapping than the default current 1:1 mapping.

## How ?
CUDA stream and event resources which were previously tied to iface now are tied to each newly detected cuda device context. When resources are needed, context ID is looked up using a hashtable and appropriate resources are picked.

## TODO
1. ~~Need a way to detect if cuda context is destroyed before destroying stream/event resources associated with that context~~ (not going to cleanup resources and leave it to the OS to handle it)
2. ~~Need to check if stream bitmap is needed for flush operations and flush each individually using streamsync~~ (removed)

### DIFF
--- a/src/uct/cuda/cuda_copy/cuda_copy_ep.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_ep.c
@@ -44,9 +44,11 @@ UCS_CLASS_DEFINE_DELETE_FUNC(uct_cuda_copy_ep_t, uct_ep_t);
     ucs_trace_data("%s [ptr %p len %zu] to 0x%" PRIx64, _name, (_iov)->buffer, \
                    (_iov)->length, (_remote_addr))
 
-ucs_status_t uct_cuda_copy_init_stream(CUstream *stream)
+
+static UCS_F_ALWAYS_INLINE ucs_status_t
+uct_cuda_copy_init_stream(CUstream *stream)
 {
-    if (*stream != 0) {
+    if (ucs_likely(*stream != NULL)) {
         return UCS_OK;
     }
 
@@ -55,19 +57,18 @@ ucs_status_t uct_cuda_copy_init_stream(CUstream *stream)
 }
 
 static UCS_F_ALWAYS_INLINE CUstream *
-uct_cuda_copy_get_stream(uct_cuda_copy_iface_t *iface,
+uct_cuda_copy_get_stream(uct_cuda_copy_ctx_rsc_t *ctx_rsc,
                          ucs_memory_type_t src_type, ucs_memory_type_t dst_type)
 {
-    CUstream *stream = NULL;
+    CUstream *stream;
     ucs_status_t status;
 
     ucs_assert((src_type < UCS_MEMORY_TYPE_LAST) &&
                (dst_type < UCS_MEMORY_TYPE_LAST));
 
-    stream = &iface->queue_desc[src_type][dst_type].stream;
-
+    stream = &ctx_rsc->queue_desc[src_type][dst_type].stream;
     status = uct_cuda_copy_init_stream(stream);
-    if (status != UCS_OK) {
+    if (ucs_unlikely(status != UCS_OK)) {
         return NULL;
     }
 
@@ -97,6 +98,28 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
     return mem_info.type;
 }
 
+static UCS_F_ALWAYS_INLINE ucs_status_t uct_cuda_copy_ctx_rsc_get(
+        uct_cuda_copy_iface_t *iface, uct_cuda_copy_ctx_rsc_t **ctx_rsc_p)
+{
+    unsigned long long ctx_id;
+    CUresult result;
+    khiter_t iter;
+
+    result = uct_cuda_base_ctx_get_id(NULL, &ctx_id);
+    if (ucs_unlikely(result != CUDA_SUCCESS)) {
+        UCT_CUDADRV_LOG(cuCtxGetId, UCS_LOG_LEVEL_ERROR, result);
+        return UCS_ERR_IO_ERROR;
+    }
+
+    iter = kh_get(cuda_copy_ctx_rscs, &iface->ctx_rscs, ctx_id);
+    if (ucs_likely(iter != kh_end(&iface->ctx_rscs))) {
+        *ctx_rsc_p = kh_value(&iface->ctx_rscs, iter);
+        return UCS_OK;
+    }
+
+    return uct_cuda_copy_ctx_rsc_create(iface, ctx_id, ctx_rsc_p);
+}
+
 static UCS_F_ALWAYS_INLINE ucs_status_t
 uct_cuda_copy_post_cuda_async_copy(uct_ep_h tl_ep, void *dst, void *src,
                                    size_t length, uct_completion_t *comp)
@@ -110,45 +133,45 @@ uct_cuda_copy_post_cuda_async_copy(uct_ep_h tl_ep, void *dst, void *src,
     ucs_memory_type_t dst_type;
     CUstream *stream;
     ucs_queue_head_t *event_q;
+    uct_cuda_copy_ctx_rsc_t *ctx_rsc;
 
     if (!length) {
         return UCS_OK;
     }
 
-    /* ensure context is set before creating events/streams */
-    if (iface->cuda_context == NULL) {
-        UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetCurrent(&iface->cuda_context));
-        if (iface->cuda_context == NULL) {
-            ucs_error("attempt to perform cuda memcpy without active context");
-            return UCS_ERR_IO_ERROR;
-        }
+    status = uct_cuda_copy_ctx_rsc_get(iface, &ctx_rsc);
+    if (ucs_unlikely(status != UCS_OK)) {
+        return status;
     }
 
     src_type = uct_cuda_copy_get_mem_type(base_iface->md, src, length);
     dst_type = uct_cuda_copy_get_mem_type(base_iface->md, dst, length);
-    q_desc   = &iface->queue_desc[src_type][dst_type];
+    q_desc   = &ctx_rsc->queue_desc[src_type][dst_type];
     event_q  = &q_desc->event_queue;
-    stream   = uct_cuda_copy_get_stream(iface, src_type, dst_type);
-    if (stream == NULL) {
+    stream   = uct_cuda_copy_get_stream(ctx_rsc, src_type, dst_type);
+    if (ucs_unlikely(stream == NULL)) {
         ucs_error("stream for src %s dst %s not available",
                    ucs_memory_type_names[src_type],
                    ucs_memory_type_names[dst_type]);
         return UCS_ERR_IO_ERROR;
     }
 
-    cuda_event = ucs_mpool_get(&iface->cuda_event_desc);
+    cuda_event = ucs_mpool_get(&ctx_rsc->event_mp);
     if (ucs_unlikely(cuda_event == NULL)) {
-        ucs_error("Failed to allocate cuda event object");
+        ucs_error("failed to allocate cuda event object");
         return UCS_ERR_NO_MEMORY;
     }
 
     status = UCT_CUDADRV_FUNC_LOG_ERR(
             cuMemcpyAsync((CUdeviceptr)dst, (CUdeviceptr)src, length, *stream));
+    if (ucs_unlikely(UCS_OK != status)) {
+        return status;
+    }
 
     status = UCT_CUDADRV_FUNC_LOG_ERR(
             cuEventRecord(cuda_event->event, *stream));
-    if (UCS_OK != status) {
-        return UCS_ERR_IO_ERROR;
+    if (ucs_unlikely(UCS_OK != status)) {
+        return status;
     }
 
     if (ucs_queue_is_empty(event_q)) {
@@ -209,24 +232,43 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_ep_put_zcopy,
 
 }
 
-UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_ep_put_short,
-                 (tl_ep, buffer, length, remote_addr, rkey),
-                 uct_ep_h tl_ep, const void *buffer, unsigned length,
-                 uint64_t remote_addr, uct_rkey_t rkey)
+static UCS_F_ALWAYS_INLINE ucs_status_t uct_cuda_copy_ep_rma_short(
+        uct_ep_h tl_ep, CUdeviceptr dst, CUdeviceptr src, unsigned length)
 {
-    uct_cuda_copy_iface_t *iface = ucs_derived_of(tl_ep->iface, uct_cuda_copy_iface_t);
-    CUstream *stream             = &iface->short_stream;
+    uct_cuda_copy_iface_t *iface = ucs_derived_of(tl_ep->iface,
+                                                  uct_cuda_copy_iface_t);
+    uct_cuda_copy_ctx_rsc_t *ctx_rsc;
     ucs_status_t status;
+    CUstream *stream;
 
-    status = uct_cuda_copy_init_stream(stream);
-    if (status != UCS_OK) {
+    status = uct_cuda_copy_ctx_rsc_get(iface, &ctx_rsc);
+    if (ucs_unlikely(status != UCS_OK)) {
         return status;
     }
 
-    UCT_CUDADRV_FUNC_LOG_ERR(cuMemcpyAsync((CUdeviceptr)remote_addr,
-                                           (CUdeviceptr)buffer, length,
-                                           *stream));
-    status = UCT_CUDADRV_FUNC_LOG_ERR(cuStreamSynchronize(*stream));
+    stream = &ctx_rsc->short_stream;
+    status = uct_cuda_copy_init_stream(stream);
+    if (ucs_unlikely(status != UCS_OK)) {
+        return status;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemcpyAsync(dst, src, length, *stream));
+    if (ucs_unlikely(status != UCS_OK)) {
+        return status;
+    }
+
+    return UCT_CUDADRV_FUNC_LOG_ERR(cuStreamSynchronize(*stream));
+}
+
+UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_ep_put_short,
+                 (tl_ep, buffer, length, remote_addr, rkey), uct_ep_h tl_ep,
+                 const void *buffer, unsigned length, uint64_t remote_addr,
+                 uct_rkey_t rkey)
+{
+    ucs_status_t status;
+
+    status = uct_cuda_copy_ep_rma_short(tl_ep, (CUdeviceptr)remote_addr,
+                                        (CUdeviceptr)buffer, length);
 
     UCT_TL_EP_STAT_OP(ucs_derived_of(tl_ep, uct_base_ep_t), PUT, SHORT, length);
     ucs_trace_data("PUT_SHORT size %d from %p to %p",
@@ -239,19 +281,10 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_ep_get_short,
                  uct_ep_h tl_ep, void *buffer, unsigned length,
                  uint64_t remote_addr, uct_rkey_t rkey)
 {
-    uct_cuda_copy_iface_t *iface = ucs_derived_of(tl_ep->iface, uct_cuda_copy_iface_t);
-    CUstream *stream             = &iface->short_stream;
     ucs_status_t status;
 
-    status = uct_cuda_copy_init_stream(stream);
-    if (status != UCS_OK) {
-        return status;
-    }
-
-    UCT_CUDADRV_FUNC_LOG_ERR(cuMemcpyAsync((CUdeviceptr)buffer,
-                                           (CUdeviceptr)remote_addr, length,
-                                           *stream));
-    status = UCT_CUDADRV_FUNC_LOG_ERR(cuStreamSynchronize(*stream));
+    status = uct_cuda_copy_ep_rma_short(tl_ep, (CUdeviceptr)buffer,
+                                        (CUdeviceptr)remote_addr, length);
 
     UCT_TL_EP_STAT_OP(ucs_derived_of(tl_ep, uct_base_ep_t), GET, SHORT, length);
     ucs_trace_data("GET_SHORT size %d from %p to %p",
--- a/src/uct/cuda/cuda_copy/cuda_copy_iface.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_iface.c
@@ -151,33 +151,6 @@ static ucs_status_t uct_cuda_copy_iface_query(uct_iface_h tl_iface,
     return UCS_OK;
 }
 
-static ucs_status_t uct_cuda_copy_sync_streams(uct_cuda_copy_iface_t *iface)
-{
-    CUstream stream;
-    uint32_t stream_index;
-    ucs_memory_type_t src_mem_type, dst_mem_type;
-    ucs_status_t status;
-
-    UCS_STATIC_BITMAP_FOR_EACH_BIT(stream_index, &iface->streams_to_sync) {
-        src_mem_type = stream_index / UCS_MEMORY_TYPE_LAST;
-        if ((src_mem_type >= UCS_MEMORY_TYPE_LAST)) {
-            break;
-        }
-
-        dst_mem_type = stream_index % UCS_MEMORY_TYPE_LAST;
-        stream       = iface->queue_desc[src_mem_type][dst_mem_type].stream;
-        status       = UCT_CUDADRV_FUNC_LOG_ERR(cuStreamSynchronize(stream));
-        if (status != UCS_OK) {
-            return status;
-        }
-
-        UCS_STATIC_BITMAP_RESET(&iface->streams_to_sync,
-                                uct_cuda_copy_flush_bitmap_idx(src_mem_type,
-                                                               dst_mem_type));
-    }
-
-    return UCS_OK;
-}
 
 static ucs_status_t uct_cuda_copy_iface_flush(uct_iface_h tl_iface, unsigned flags,
                                               uct_completion_t *comp)
@@ -185,15 +158,9 @@ static ucs_status_t uct_cuda_copy_iface_flush(uct_iface_h tl_iface, unsigned fla
     uct_cuda_copy_iface_t *iface = ucs_derived_of(tl_iface, uct_cuda_copy_iface_t);
     uct_cuda_copy_queue_desc_t *q_desc;
     ucs_queue_iter_t iter;
-    ucs_status_t status;
 
     if (comp != NULL) {
         return UCS_ERR_UNSUPPORTED;
-    }
-
-    status = uct_cuda_copy_sync_streams(iface);
-    if (status != UCS_OK) {
-        return status;
     }
 
     ucs_queue_for_each_safe(q_desc, iter, &iface->active_queue, queue) {
@@ -206,7 +173,6 @@ static ucs_status_t uct_cuda_copy_iface_flush(uct_iface_h tl_iface, unsigned fla
 
     UCT_TL_IFACE_STAT_FLUSH(ucs_derived_of(tl_iface, uct_base_iface_t));
     return UCS_OK;
-
 }
 
 static UCS_F_ALWAYS_INLINE unsigned
@@ -323,21 +289,6 @@ static ucs_status_t uct_cuda_copy_iface_event_fd_arm(uct_iface_h tl_iface,
     return UCS_OK;
 }
 
-static ucs_status_t
-uct_cuda_copy_ep_flush(uct_ep_h tl_ep, unsigned flags, uct_completion_t *comp)
-{
-    uct_cuda_copy_iface_t *iface = ucs_derived_of(tl_ep->iface,
-                                                  uct_cuda_copy_iface_t);
-    ucs_status_t status;
-
-    status = uct_cuda_copy_sync_streams(iface);
-    if (status != UCS_OK) {
-        return status;
-    }
-
-    return uct_base_ep_flush(tl_ep, flags, comp);
-}
-
 static uct_iface_ops_t uct_cuda_copy_iface_ops = {
     .ep_get_short             = uct_cuda_copy_ep_get_short,
     .ep_put_short             = uct_cuda_copy_ep_put_short,
@@ -345,7 +296,7 @@ static uct_iface_ops_t uct_cuda_copy_iface_ops = {
     .ep_put_zcopy             = uct_cuda_copy_ep_put_zcopy,
     .ep_pending_add           = (uct_ep_pending_add_func_t)ucs_empty_function_return_busy,
     .ep_pending_purge         = (uct_ep_pending_purge_func_t)ucs_empty_function,
-    .ep_flush                 = uct_cuda_copy_ep_flush,
+    .ep_flush                 = uct_base_ep_flush,
     .ep_fence                 = uct_base_ep_fence,
     .ep_create                = UCS_CLASS_NEW_FUNC_NAME(uct_cuda_copy_ep_t),
     .ep_destroy               = UCS_CLASS_DELETE_FUNC_NAME(uct_cuda_copy_ep_t),
@@ -363,30 +314,46 @@ static uct_iface_ops_t uct_cuda_copy_iface_ops = {
     .iface_is_reachable       = uct_base_iface_is_reachable
 };
 
-static void uct_cuda_copy_event_desc_init(ucs_mpool_t *mp, void *obj, void *chunk)
+static void
+uct_cuda_copy_event_desc_init(ucs_mpool_t *mp, void *obj, void *chunk)
 {
-    uct_cuda_copy_event_desc_t *base = (uct_cuda_copy_event_desc_t *) obj;
-    ucs_status_t status;
+    uct_cuda_copy_event_desc_t *base = (uct_cuda_copy_event_desc_t*)obj;
 
     memset(base, 0 , sizeof(*base));
-    status = UCT_CUDADRV_FUNC_LOG_ERR(
+    UCT_CUDADRV_FUNC_LOG_ERR(
             cuEventCreate(&base->event, CU_EVENT_DISABLE_TIMING));
-    if (UCS_OK != status) {
-        ucs_error("cuEventCreate Failed");
+}
+
+static int uct_cuda_copy_is_ctx_valid(uct_cuda_copy_ctx_rsc_t *ctx_rsc)
+{
+#if CUDA_VERSION >= 12000
+    unsigned long long ctx_id;
+    CUresult result;
+
+    result = uct_cuda_base_ctx_get_id(ctx_rsc->ctx, &ctx_id);
+    if (result == CUDA_ERROR_CONTEXT_IS_DESTROYED) {
+        return 0;
+    } else if (result != CUDA_SUCCESS) {
+        UCT_CUDADRV_LOG(cuCtxGetId, UCS_LOG_LEVEL_WARN, result);
+        return 0;
     }
+
+    return ctx_id == ctx_rsc->ctx_id;
+#else
+    /* Best effort check on older Cuda versions */
+    return uct_cuda_base_is_context_valid(ctx_rsc->ctx);
+#endif
 }
 
 static void uct_cuda_copy_event_desc_cleanup(ucs_mpool_t *mp, void *obj)
 {
-    uct_cuda_copy_event_desc_t *base = (uct_cuda_copy_event_desc_t *) obj;
-    uct_cuda_copy_iface_t *iface     = ucs_container_of(mp,
-                                                        uct_cuda_copy_iface_t,
-                                                        cuda_event_desc);
-    CUcontext cuda_context;
+    uct_cuda_copy_event_desc_t *base = obj;
+    uct_cuda_copy_ctx_rsc_t *ctx_rsc = ucs_container_of(mp,
+                                                        uct_cuda_copy_ctx_rsc_t,
+                                                        event_mp);
 
-    UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetCurrent(&cuda_context));
-    if (uct_cuda_base_context_match(cuda_context, iface->cuda_context)) {
-        UCT_CUDADRV_FUNC_LOG_ERR(cuEventDestroy(base->event));
+    if (uct_cuda_copy_is_ctx_valid(ctx_rsc)) {
+        UCT_CUDADRV_FUNC_LOG_WARN(cuEventDestroy(base->event));
     }
 }
 
@@ -480,6 +447,76 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
     .ep_is_connected       = uct_base_ep_is_connected
 };
 
+ucs_status_t uct_cuda_copy_ctx_rsc_create(uct_cuda_copy_iface_t *iface,
+                                          unsigned long long ctx_id,
+                                          uct_cuda_copy_ctx_rsc_t **ctx_rsc_p)
+{
+    CUcontext ctx;
+    ucs_status_t status;
+    ucs_kh_put_t ret;
+    khiter_t iter;
+    uct_cuda_copy_ctx_rsc_t *ctx_rsc;
+    ucs_mpool_params_t mp_params;
+    ucs_memory_type_t src, dst;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetCurrent(&ctx));
+    if (status != UCS_OK) {
+        return status;
+    } else if (ctx == NULL) {
+        ucs_error("no cuda context bound to calling thread");
+        return UCS_ERR_IO_ERROR;
+    }
+
+    iter = kh_put(cuda_copy_ctx_rscs, &iface->ctx_rscs, ctx_id, &ret);
+    if (ret == UCS_KH_PUT_FAILED) {
+        ucs_error("failed to allocate cuda context resource hash entry");
+        return UCS_ERR_NO_MEMORY;
+    }
+
+    ucs_assertv_always(ret != UCS_KH_PUT_KEY_PRESENT,
+                       "the key has already been added iface=%p key=%llu",
+                       iface, ctx_id);
+
+    ctx_rsc = ucs_malloc(sizeof(*ctx_rsc), "uct_cuda_copy_ctx_rsc_t");
+    if (ctx_rsc == NULL) {
+        ucs_error("failed to allocate cuda context resource struct");
+        status = UCS_ERR_NO_MEMORY;
+        goto err_del_iter;
+    }
+
+    ucs_mpool_params_reset(&mp_params);
+    mp_params.elem_size       = sizeof(uct_cuda_copy_event_desc_t);
+    mp_params.elems_per_chunk = 128;
+    mp_params.max_elems       = iface->config.max_cuda_events;
+    mp_params.ops             = &uct_cuda_copy_event_desc_mpool_ops;
+    mp_params.name            = "cuda_copy_event_descriptors";
+
+    status = ucs_mpool_init(&mp_params, &ctx_rsc->event_mp);
+    if (status != UCS_OK) {
+        goto err_free_ctx_rsc;
+    }
+
+    ucs_memory_type_for_each(src) {
+        ucs_memory_type_for_each(dst) {
+            ctx_rsc->queue_desc[src][dst].stream = NULL;
+            ucs_queue_head_init(&ctx_rsc->queue_desc[src][dst].event_queue);
+        }
+    }
+
+    ctx_rsc->short_stream            = NULL;
+    ctx_rsc->ctx                     = ctx;
+    ctx_rsc->ctx_id                  = ctx_id;
+    kh_value(&iface->ctx_rscs, iter) = ctx_rsc;
+    *ctx_rsc_p                       = ctx_rsc;
+    return UCS_OK;
+
+err_free_ctx_rsc:
+    ucs_free(ctx_rsc);
+err_del_iter:
+    kh_del(cuda_copy_ctx_rscs, &iface->ctx_rscs, iter);
+    return UCS_ERR_NO_MEMORY;
+}
+
 static UCS_CLASS_INIT_FUNC(uct_cuda_copy_iface_t, uct_md_h md, uct_worker_h worker,
                            const uct_iface_params_t *params,
                            const uct_iface_config_t *tl_config)
@@ -487,8 +524,6 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_copy_iface_t, uct_md_h md, uct_worker_h work
     uct_cuda_copy_iface_config_t *config = ucs_derived_of(tl_config,
                                                           uct_cuda_copy_iface_config_t);
     ucs_status_t status;
-    ucs_memory_type_t src, dst;
-    ucs_mpool_params_t mp_params;
 
     UCS_CLASS_CALL_SUPER_INIT(uct_cuda_iface_t, &uct_cuda_copy_iface_ops,
                               &uct_cuda_copy_iface_internal_ops, md, worker,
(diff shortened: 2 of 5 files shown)


src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -55,22 +55,27 @@ ucs_status_t uct_cuda_copy_init_stream(CUstream *stream)
 }
 
 static UCS_F_ALWAYS_INLINE CUstream *
-uct_cuda_copy_get_stream(uct_cuda_copy_iface_t *iface,
+uct_cuda_copy_get_stream(uct_cuda_copy_per_ctx_rsc_t *ctx_rsc,
                          ucs_memory_type_t src_type, ucs_memory_type_t dst_type)
 {
-    CUstream *stream = NULL;
+    static pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
+    CUstream *stream            = NULL;
```
SeyedMir, line 62: Is the NULL assignment necessary? Line 70 will always overwrite it.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -109,34 +114,41 @@ uct_cuda_copy_post_cuda_async_copy(uct_ep_h tl_ep, void *dst, void *src,
...
+    if (current_ctx == NULL) {
+        ucs_error("attempt to perform cuda memcpy without active context");
+        return UCS_ERR_IO_ERROR;
+    } else {
+        status = uct_cuda_copy_get_ctx_rscs(iface, current_ctx, &ctx_rsc);
+        if (UCS_OK != status) {
+            ucs_error("unable to get resources associated with cuda context");
             return UCS_ERR_IO_ERROR;
         }
     }
```
SeyedMir, line 137: This block of code (lines 128-137) is repeated in put_short and get_short as well. Maybe put it in a function or macro?

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -109,34 +114,41 @@ uct_cuda_copy_post_cuda_async_copy(uct_ep_h tl_ep, void *dst, void *src,
...
+
     /* ensure context is set before creating events/streams */
-    if (iface->cuda_context == NULL) {
-        UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetCurrent(&iface->cuda_context));
-        if (iface->cuda_context == NULL) {
-            ucs_error("attempt to perform cuda memcpy without active context");
+    UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetCurrent(&current_ctx));
+    if (current_ctx == NULL) {
+        ucs_error("attempt to perform cuda memcpy without active context");
+        return UCS_ERR_IO_ERROR;
```
SeyedMir, line 130: Should we return error or attempt to set the context as we have the buffers? Though, it may not be in the scope of this PR.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -455,94 +457,175 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
     .ep_is_connected       = uct_base_ep_is_connected
 };
 
-static UCS_CLASS_INIT_FUNC(uct_cuda_copy_iface_t, uct_md_h md, uct_worker_h worker,
-                           const uct_iface_params_t *params,
-                           const uct_iface_config_t *tl_config)
+void uct_cuda_copy_cleanup_per_ctx_rscs(uct_cuda_copy_per_ctx_rsc_t *ctx_rsc)
```
SeyedMir, line 460: Seems like this can be a static function.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -455,94 +457,175 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
     }
 
-    self->id                     = ucs_generate_uuid((uintptr_t)self);
-    self->config.max_poll        = config->max_poll;
-    self->config.max_cuda_events = config->max_cuda_events;
-    self->config.bandwidth       = config->bandwidth;
-    UCS_BITMAP_CLEAR(&self->streams_to_sync);
+}
+
+ucs_status_t uct_cuda_copy_init_per_ctx_rscs(uct_cuda_copy_iface_t *iface,
```
SeyedMir, line 501: Seems like this can be a static function.
Also, I think this function does not need the `iface`. The `max_cuda_events` value can be passed directly as an argument instead. If you decide to keep passing the `iface`, then let's add `const`.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -455,94 +457,175 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
-    self->cuda_context = 0;
+    ctx_rsc->short_stream = 0;
+    ctx_rsc->cuda_ctx     = cuda_ctx;
+    ctx_rsc->ctx_id       = ctx_id;
 
     return UCS_OK;
 }
 
-static UCS_CLASS_CLEANUP_FUNC(uct_cuda_copy_iface_t)
+ucs_status_t uct_cuda_copy_get_ctx_rscs(uct_cuda_copy_iface_t *iface,
```
SeyedMir, line 537: Should this have `_per_` in the name to be consistent with `init_per_ctx_rscs` and `cleanup_per_ctx_rscs` functions?

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -455,94 +457,175 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
+
+                UCT_CUDADRV_FUNC_LOG_ERR(cuStreamDestroy(*stream));
+            }
+        }
+
+        if (ctx_rsc->short_stream) {
+            UCT_CUDADRV_FUNC_LOG_ERR(cuStreamDestroy(ctx_rsc->short_stream));
+        }
+
+        ucs_mpool_cleanup(&ctx_rsc->cuda_event_desc, 1);
```
SeyedMir, line 495: Shouldn't this mpool cleanup be called even if the `ctx_rsc->cuda_ctx` is not valid anymore?

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -373,14 +374,22 @@ uct_cuda_copy_md_query_attributes(uct_cuda_copy_md_t *md, const void *address,
         goto out_default_range;
     }
 
+    /* GetAddressRange requires context to be set. On DGXA100 it takes 0.03 us
+     * to push and pop the context associated with address (which should be
+     * non-NULL if we are at this point)*/
+    cuCtxPushCurrent(cuda_mem_ctx);
+
     cu_err = cuMemGetAddressRange(&base_address, &alloc_length,
                                   (CUdeviceptr)address);
```
SeyedMir, line 383: Shall we move `cuCtxPopCurrent(&cuda_popped_ctx);` to here? Because we want to pop the pushed context regardless of the success or failure of `cuMemGetAddressRange`

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -55,22 +55,27 @@ ucs_status_t uct_cuda_copy_init_stream(CUstream *stream)
...
-uct_cuda_copy_get_stream(uct_cuda_copy_iface_t *iface,
+uct_cuda_copy_get_stream(uct_cuda_copy_per_ctx_rsc_t *ctx_rsc,
                          ucs_memory_type_t src_type, ucs_memory_type_t dst_type)
 {
-    CUstream *stream = NULL;
+    static pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
+    CUstream *stream            = NULL;
     ucs_status_t status;
 
+    pthread_mutex_lock(&lock);
```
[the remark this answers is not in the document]
Akshay-Venkatesh: @rakhmets As we're modifying context resource and multiple threads may want the same stream, we don't want the same stream variable to be initialized more than once.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
Akshay-Venkatesh: Not needed. Will leave it uninitialized.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -55,22 +55,27 @@ ucs_status_t uct_cuda_copy_init_stream(CUstream *stream)
...
 
-    stream = &iface->queue_desc[src_type][dst_type].stream;
+    stream = &ctx_rsc->queue_desc[src_type][dst_type].stream;
 
     status = uct_cuda_copy_init_stream(stream);
     if (status != UCS_OK) {
         return NULL;
     }
 
+    pthread_mutex_unlock(&lock);
```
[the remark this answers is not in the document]
Akshay-Venkatesh: Will fix this.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
Akshay-Venkatesh: Scope for this PR is for the thread to have set the right context. Will address this in a follow up PR.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -240,9 +269,25 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_copy_ep_get_short,
...
+    if (current_ctx == NULL) {
+        ucs_error("attempt to perform cuda memcpy without active context");
+        return UCS_ERR_IO_ERROR;
+    } else {
+        status = uct_cuda_copy_get_ctx_rscs(iface, current_ctx, &ctx_rsc);
+        if (UCS_OK != status) {
+            ucs_error("unable to get resources associated with cuda context");
+            return UCS_ERR_IO_ERROR;
+        }
+    }
```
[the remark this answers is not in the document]
Akshay-Venkatesh: Will change to inline function.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
Akshay-Venkatesh: If context is not present, then context could've been destroyed by the user before ucp_context_destroy or MPI_Finalize.

src/uct/cuda/cuda_copy/cuda_copy_md.c
Akshay-Venkatesh: Thanks for the suggestion. @brminich made the same suggestion too.

Akshay-Venkatesh (commented):
Thanks for the feedback. I'll make these changes today.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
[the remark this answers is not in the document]
brminich: but UCT is not thread-safe and in UCP we have global lock per operation.
is there a specific use-cases requiring this lock?

Akshay-Venkatesh:
@brminich I see one of the commits had an extra colon and 2 commit style tests are failing because of that. Would it be ok to rebase? I can wait to do this until all the reviewers have had a chance to look at my comments and code changes.

cc @rakhmets @SeyedMir

SeyedMir:
@Akshay-Venkatesh Rebase is fine with me.

brminich:
@Akshay-Venkatesh, no problem from my side

rakhmets (commented):
Rebase is OK for me too.

Akshay-Venkatesh:
@brminich @rakhmets @SeyedMir 

FYI, in https://github.com/openucx/ucx/pull/9645/commits/dd8b66d905c3363cec94554c8f16d70a2966adb9 I had to remove all code that does EventDestroy or StreamDestroy as CUDA doesn't have a way to query if a give CUcontext has been destroyed or not and calling Stream/EventDestroy on streams/events whose context has been destroyed is potentially unsafe. For this reason we will have to leave it to the point when the process is cleaned up. This should be safe from UCX's viewpoint as all UCT resources are tied to some UCP context and there isn't a concern of reusing streams/events that haven't been cleaned up (as they are not global).

src/uct/cuda/cuda_copy/cuda_copy_ep.c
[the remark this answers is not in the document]
rakhmets: I believe the lock can be removed.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -97,6 +103,38 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
...
+                                       uct_cuda_copy_per_ctx_rsc_t **ctx_rsc)
+{
+    CUcontext current_ctx;
+    ucs_status_t status;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetCurrent(&current_ctx));
+    if (status != UCS_OK) {
+        return status;
+    } else if (current_ctx == NULL) {
+        ucs_error("attempt to perform cuda memcpy without active context");
```
rakhmets, line 117: I think this log message can be confusing since the function can be called by different callers. It is better to print the following message: `there is no cuda context bound to the calling cpu thread`.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -97,6 +103,38 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
     return mem_info.type;
 }
 
+static inline
+ucs_status_t uct_cuda_copy_get_ctx_rsc(uct_cuda_copy_iface_t *iface,
+                                       uct_cuda_copy_per_ctx_rsc_t **ctx_rsc)
```
rakhmets, line 108:
```suggestion
static UCS_F_ALWAYS_INLINE ucs_status_t
uct_cuda_copy_get_ctx_rsc(uct_cuda_copy_iface_t *iface,
                          uct_cuda_copy_per_ctx_rsc_t **ctx_rsc)
```

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -455,15 +450,101 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
+    }
+
+    *ctx_rsc_p = ctx_rsc;
+    return UCS_OK;
+
+err_free_ctx:
+    ucs_free(ctx_rsc);
+err_kh_del:
+    kh_del(cuda_copy_ctx_rscs, &iface->ctx_rscs, iter);
+out:
```
rakhmets, line 537:
```suggestion
err:
```

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -373,14 +374,21 @@ uct_cuda_copy_md_query_attributes(uct_cuda_copy_md_t *md, const void *address,
...
+    cuCtxPopCurrent(&cuda_popped_ctx);
     if (cu_err != CUDA_SUCCESS) {
         ucs_error("cuMemGetAddressRange(%p) error: %s", address,
                   uct_cuda_base_cu_get_error_string(cu_err));
         return UCS_ERR_INVALID_ADDR;
     }
 
+
```
rakhmets, line 391: Please remove extra empty line.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -455,15 +450,101 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
+    ctx_rsc->short_stream = 0;
+    ctx_rsc->cuda_ctx     = cuda_ctx;
+    ctx_rsc->ctx_id       = ctx_id;
+
+    return UCS_OK;
+}
+
+ucs_status_t uct_cuda_copy_get_per_ctx_rscs(uct_cuda_copy_iface_t *iface,
+                                            CUcontext cuda_ctx,
+                                            uct_cuda_copy_per_ctx_rsc_t **ctx_rsc_p)
```
rakhmets, line 492: Please remove the declaration from the .h file, and make the function static.
```suggestion
static ucs_status_t
uct_cuda_copy_get_per_ctx_rscs(uct_cuda_copy_iface_t *iface, CUcontext cuda_ctx,
                               uct_cuda_copy_per_ctx_rsc_t **ctx_rsc_p)
```

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -48,23 +49,30 @@ typedef struct uct_cuda_copy_queue_desc {
     ucs_queue_elem_t            queue;
 } uct_cuda_copy_queue_desc_t;
 
+typedef struct uct_cuda_copy_per_ctx_rsc {
+    CUcontext                   cuda_ctx;
+    unsigned long long          ctx_id;
```
rakhmets, line 54: Please remove `ctx_id` filed.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -131,6 +131,7 @@ static ucs_status_t uct_cuda_copy_iface_query(uct_iface_h tl_iface,
     return UCS_OK;
 }
 
+#if 0
```
rakhmets, line 134: Please remove unused code.

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -48,23 +49,30 @@ typedef struct uct_cuda_copy_queue_desc {
...
+    /* pool of cuda events to check completion of memcpy operations */
+    ucs_mpool_t                 cuda_event_desc;
+    /* stream used to issue short operations */
+    CUstream                    short_stream;
+    /* array of queue descriptors for each src/dst memory type combination */
+    uct_cuda_copy_queue_desc_t  queue_desc[UCS_MEMORY_TYPE_LAST][UCS_MEMORY_TYPE_LAST];
+} uct_cuda_copy_per_ctx_rsc_t;
+
+
+KHASH_MAP_INIT_INT64(cuda_copy_ctx_rscs, struct uct_cuda_copy_per_ctx_rsc*);
```
brminich, line 64: you can store `uct_cuda_copy_per_ctx_rsc` (not the pointer), then you would need to do alloc/free during put.

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -48,23 +49,30 @@ typedef struct uct_cuda_copy_queue_desc {
     ucs_queue_elem_t            queue;
 } uct_cuda_copy_queue_desc_t;
 
+typedef struct uct_cuda_copy_per_ctx_rsc {
+    CUcontext                   cuda_ctx;
+    unsigned long long          ctx_id;
+    /* pool of cuda events to check completion of memcpy operations */
+    ucs_mpool_t                 cuda_event_desc;
```
brminich, line 56: do we really need to have this mpool per context? Maybe one common mpool is enough?

Akshay-Venkatesh: No event and stream resources are associated with a context.

src/uct/cuda/cuda_copy/cuda_copy_iface.h
Akshay-Venkatesh: @rakhmets But we use this as the key to get the right context from the hashmap. Are you suggesting that we use a different key?

src/uct/cuda/cuda_copy/cuda_copy_iface.h
Akshay-Venkatesh: @brminich Will incorporate this change.

src/uct/cuda/cuda_copy/cuda_copy_iface.h
rakhmets: No, I suggest using ctx_id as key, without storing to a value.

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -48,23 +49,31 @@ typedef struct uct_cuda_copy_queue_desc {
     ucs_queue_elem_t            queue;
 } uct_cuda_copy_queue_desc_t;
 
+typedef struct uct_cuda_copy_ctx_rsc {
+    /* pool of cuda events to check completion of memcpy operations */
+    ucs_mpool_t                cuda_event_desc;
```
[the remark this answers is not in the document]
rakhmets: Yes, we do. `cuda_event_desc` contains CUDA event. CUDA event and CUDA stream must be from the same context otherwise `cuEventRecord` returns `CUDA_ERROR_INVALID_HANDLE`. The new gtests cover the issue.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -46,32 +46,31 @@ UCS_CLASS_DEFINE_DELETE_FUNC(uct_cuda_copy_ep_t, uct_ep_t);
 
 ucs_status_t uct_cuda_copy_init_stream(CUstream *stream)
 {
-    if (*stream != 0) {
+    if (*stream != NULL) {
```
brminich, line 49: maybe also make this func inline?

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -108,45 +144,41 @@ uct_cuda_copy_post_cuda_async_copy(uct_ep_h tl_ep, void *dst, void *src,
...
         return UCS_ERR_NO_MEMORY;
     }
 
     status = UCT_CUDADRV_FUNC_LOG_ERR(
-            cuMemcpyAsync((CUdeviceptr)dst, (CUdeviceptr)src, length, *stream));
+            cuMemcpyAsync((CUdeviceptr)dst, (CUdeviceptr)src, length, stream));
 
-    status = UCT_CUDADRV_FUNC_LOG_ERR(
-            cuEventRecord(cuda_event->event, *stream));
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuEventRecord(cuda_event->event, stream));
```
brminich, line 181: is it possible that `cuMemcpyAsync` fails, but `cuEventRecord` succeeds? 
`status` will b eoverwritten then

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -313,29 +283,14 @@ static ucs_status_t uct_cuda_copy_iface_event_fd_arm(uct_iface_h tl_iface,
...
-
 static uct_iface_ops_t uct_cuda_copy_iface_ops = {
     .ep_get_short             = uct_cuda_copy_ep_get_short,
     .ep_put_short             = uct_cuda_copy_ep_put_short,
     .ep_get_zcopy             = uct_cuda_copy_ep_get_zcopy,
     .ep_put_zcopy             = uct_cuda_copy_ep_put_zcopy,
     .ep_pending_add           = (uct_ep_pending_add_func_t)ucs_empty_function_return_busy,
     .ep_pending_purge         = (uct_ep_pending_purge_func_t)ucs_empty_function,
-    .ep_flush                 = uct_cuda_copy_ep_flush,
+    .ep_flush                 = uct_base_ep_flush,
```
brminich, line 293: why is flush not needed anymore?

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -97,6 +97,43 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
     return mem_info.type;
 }
 
+static UCS_F_ALWAYS_INLINE ucs_status_t uct_cuda_copy_get_ctx_rsc(
+        uct_cuda_copy_iface_t *iface, uct_cuda_copy_ctx_rsc_t **ctx_rsc)
+{
+    CUcontext current_ctx;
+    ucs_status_t status;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetCurrent(&current_ctx));
```
yosefe, line 106: /* TODO take context from buffer */

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -97,6 +97,43 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
...
+    if (status != UCS_OK) {
+        return status;
+    } else if (current_ctx == NULL) {
```
yosefe, line 109: unlikely

src/uct/cuda/cuda_copy/cuda_copy_ep.c
yosefe: unlikely

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -469,15 +410,90 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
+#if CUDA_VERSION >= 12000
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetId(cuda_ctx, &ctx_id));
+    if (status != UCS_OK) {
+        goto err;
+    }
+#else
+    ctx_id = 0;
+#endif
+
+    iter = kh_put(cuda_copy_ctx_rscs, &iface->ctx_rscs, ctx_id, &ret);
```
yosefe, line 465: 1. map by ctx pointer instead of id (we assume cannot get pointer of released context)
 check as assert that ctx_id is equal to cached value
2. make 1st part of this function inline and use kh_get

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -97,6 +97,43 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
...
+}
+
+static UCS_F_ALWAYS_INLINE ucs_status_t
+uct_cuda_copy_get_short_stream(uct_cuda_copy_iface_t *iface, CUstream *stream)
+{
+    uct_cuda_copy_ctx_rsc_t *ctx_rsc;
+    ucs_status_t status;
+
+    status = uct_cuda_copy_get_ctx_rsc(iface, &ctx_rsc);
+    if (status != UCS_OK) {
```
yosefe, line 124: unlikely

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -97,6 +97,43 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
...
+        return status;
+    }
+
+    status = uct_cuda_copy_init_stream(&ctx_rsc->short_stream);
+    if (status != UCS_OK) {
```
yosefe, line 129: unlikely

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -455,7 +396,7 @@ static ucs_mpool_ops_t uct_cuda_copy_event_desc_mpool_ops = {
     .chunk_alloc   = ucs_mpool_chunk_malloc,
     .chunk_release = ucs_mpool_chunk_free,
     .obj_init      = uct_cuda_copy_event_desc_init,
-    .obj_cleanup   = uct_cuda_copy_event_desc_cleanup,
+    .obj_cleanup   = (ucs_mpool_obj_cleanup_func_t)ucs_empty_function,
```
yosefe, line 399: need cleanup?

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -494,69 +510,25 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_copy_iface_t, uct_md_h md, uct_worker_h work
...
-
-        if (self->short_stream) {
-            UCT_CUDADRV_FUNC_LOG_ERR(cuStreamDestroy(self->short_stream));
-        }
-    }
+    kh_foreach_value(&self->ctx_rscs, ctx_rsc, {
+        ucs_mpool_cleanup(&ctx_rsc.cuda_event_desc, 1);
+    });
 
-    ucs_mpool_cleanup(&self->cuda_event_desc, 1);
```
yosefe, line 559: need to cleanup

test/gtest/uct/cuda/test_switch_cuda_device.cc
```diff
@@ -143,3 +144,59 @@ UCS_TEST_P(test_switch_cuda_device, detect_mem_type_cuda_fabric)
...
+
+    if (num_devices < 2) {
+        UCS_TEST_SKIP_R("less than two cuda devices available");
+    }
+
+    uct_p2p_rma_test::test_xfer(send, length, flags, mem_type);
+
+    int current_device;
+    ASSERT_EQ(cudaGetDevice(&current_device), cudaSuccess);
+    ASSERT_EQ(cudaSetDevice((current_device + 1) % num_devices), cudaSuccess);
```
yosefe, line 169: instead can we run xfer tests on all avail cuda devices? (up to 2)? instead of adding another test class
maybe even using c++ iterator

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -469,15 +452,114 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
+    uct_cuda_copy_iface_mpool_priv_t *mpool_priv;
+
+#if CUDA_VERSION >= 12000
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetId(ctx, &ctx_rsc->ctx_id));
+    if (status != UCS_OK) {
+        return status;
+    }
+#endif
+
+    status = uct_cuda_copy_ctx_create_validator(ctx, &ctx_rsc->validator);
```
brminich, line 491: can you keep validator just in mpool_priv and initialize it after mpool creation?
why is it needed to also keep it in the context?

rakhmets: Removed extra field.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -97,40 +98,41 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
...
+    *ctx_rsc_p = &kh_value(&iface->ctx_rscs, iter);
+
+#if CUDA_VERSION >= 12000
+    ucs_assertv((result = cuCtxGetId(ctx, &ctx_id)) == CUDA_SUCCESS,
+                "cuCtxGetId failed %s",
+                uct_cuda_base_cu_get_error_string(result));
+    ucs_assertv(ctx_id == (*ctx_rsc_p)->ctx_id,
+                "context id %llu does not match the cached value %llu", ctx_id,
+                (*ctx_rsc_p)->ctx_id);
+#endif
```
yosefe, line 134: if context id is changed, drop the hash value and call uct_cuda_copy_ctx_rsc_create

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -23,6 +23,15 @@
 #define UCT_CUDA_COPY_IFACE_LATENCY  ucs_linear_func_make(8e-6, 0)
 
 
+typedef struct uct_cuda_copy_ctx_mpool_priv {
+#if CUDA_VERSION >= 12000
+    CUcontext validator;
+#else
+    CUdeviceptr validator;
+#endif
+} uct_cuda_copy_ctx_mpool_priv_t;
```
yosefe, line 32: move validator to ctx_rsc
we can use ucs_container_of(mp, uct_cuda_copy_ctx_rsc_t, even_desc_mp) to get the ctx from obj_destroy

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -442,49 +512,34 @@ static ucs_status_t uct_cuda_copy_init_ctx_rsc(unsigned int max_cuda_events,
...
+                       "the key %lu has already been added", (uintptr_t)ctx);
+
     ctx_rsc = &kh_value(&iface->ctx_rscs, iter);
-    if (ret != UCS_KH_PUT_KEY_PRESENT) {
-        status = uct_cuda_copy_init_ctx_rsc(iface->config.max_cuda_events,
-                                            ctx_rsc);
-        if (status != UCS_OK) {
-            goto err_kh_del;
-        }
+    status  = uct_cuda_copy_ctx_rsc_init(ctx, iface->config.max_cuda_events,
```
yosefe, line 534: consider uniting uct_cuda_copy_ctx_rsc_init into uct_cuda_copy_ctx_rsc_create

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -517,15 +572,62 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_copy_iface_t, uct_md_h md, uct_worker_h work
...
+    valid_ctx  = uct_cuda_copy_ctx_rsc_valid(mpool_priv->validator);
+
+    ucs_memory_type_for_each(src) {
+        ucs_memory_type_for_each(dst) {
+            event_q = &ctx_rsc->queue_desc[src][dst].event_queue;
+            if (!ucs_queue_is_empty(event_q)) {
+                ucs_warn("stream destroyed but queue not empty");
+            }
+
+            uct_cuda_copy_stream_destroy(&ctx_rsc->queue_desc[src][dst].stream,
```
yosefe, line 602: to consider - IMO passing valid_ctx to uct_cuda_copy_stream_destroy is redundant - if context is destroyed we don't have to go over all streams in the first place
but on other hand need to warn if stream event q is not empty

yosefe (commented):
add context switch to test_xfer

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -353,30 +304,43 @@ static uct_iface_ops_t uct_cuda_copy_iface_ops = {
...
+#if CUDA_VERSION >= 12000
+    unsigned long long ctx_id;
+    CUresult result;
+
+    result = cuCtxGetId(ctx_rsc->ctx, &ctx_id);
+    if (result == CUDA_ERROR_CONTEXT_IS_DESTROYED) {
+        return 0;
+    } else if (result != CUDA_SUCCESS) {
+        UCT_CUDADRV_LOG(cuCtxGetId, UCS_LOG_LEVEL_WARN, result);
+        return 0;
```
brminich, line 329: can we just use `UCT_CUDADRV_FUNC`?

brminich: ok

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -97,6 +98,40 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
...
+    CUcontext ctx;
+    ucs_status_t status;
+    khiter_t iter;
+    unsigned long long ctx_id;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetCurrent(&ctx));
+    if (ucs_unlikely(status != UCS_OK)) {
+        return status;
+    } else if (ucs_unlikely(ctx == NULL)) {
+        ucs_error("no context bound to calling thread");
```
yosefe, line 113: 1. "no cuda context bound to calling thread"
2. didn't we decide in this case to use primary context of device 0?

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -97,6 +98,40 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
...
+        return UCS_ERR_IO_ERROR;
+    }
+
+#if CUDA_VERSION >= 12000
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetId(ctx, &ctx_id));
```
yosefe, line 118: From cuCtxGetId doc:
 Returns in ctxId the unique Id which is associated with a given context. The Id is unique for the life of the program for this instance of CUDA. **If context is supplied as NULL and there is one current, the Id of the current context is returned**.

So maybe we can avoid the extra call to cuCtxGetCurrent and pass NULL?

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -313,29 +279,14 @@ static ucs_status_t uct_cuda_copy_iface_event_fd_arm(uct_iface_h tl_iface,
...
-
 static uct_iface_ops_t uct_cuda_copy_iface_ops = {
     .ep_get_short             = uct_cuda_copy_ep_get_short,
     .ep_put_short             = uct_cuda_copy_ep_put_short,
     .ep_get_zcopy             = uct_cuda_copy_ep_get_zcopy,
     .ep_put_zcopy             = uct_cuda_copy_ep_put_zcopy,
     .ep_pending_add           = (uct_ep_pending_add_func_t)ucs_empty_function_return_busy,
     .ep_pending_purge         = (uct_ep_pending_purge_func_t)ucs_empty_function,
-    .ep_flush                 = uct_cuda_copy_ep_flush,
+    .ep_flush                 = uct_base_ep_flush,
```
yosefe, line 289: did we check it doesn't break ucx_perftest ?

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -469,15 +433,62 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
+    ucs_kh_put_t ret;
+    khiter_t iter;
+    uct_cuda_copy_ctx_rsc_t *ctx_rsc;
+    ucs_status_t status;
+    ucs_mpool_params_t mp_params;
+    ucs_memory_type_t src, dst;
+
+    iter = kh_put(cuda_copy_ctx_rscs, &iface->ctx_rscs, ctx_id, &ret);
+    if (ret == UCS_KH_PUT_FAILED) {
+        ucs_error("cannot allocate hash entry");
```
yosefe, line 450: "cannot allocate cuda context hash entry"

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -469,15 +433,62 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
+        return UCS_ERR_NO_MEMORY;
+    }
+
+    ucs_assertv_always(ret != UCS_KH_PUT_KEY_PRESENT,
+                       "the key %llu has already been added", ctx_id);
+
+    ctx_rsc = &kh_value(&iface->ctx_rscs, iter);
+
+    ucs_mpool_params_reset(&mp_params);
```
yosefe, line 459: "iface=%p key=%llu"

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -469,15 +433,62 @@ static uct_iface_internal_ops_t uct_cuda_copy_iface_internal_ops = {
...
+    status = ucs_mpool_init(&mp_params, &ctx_rsc->cuda_event_desc);
+    if (status != UCS_OK) {
+        kh_del(cuda_copy_ctx_rscs, &iface->ctx_rscs, iter);
+        return status;
+    }
+
+    ucs_memory_type_for_each(src) {
+        ucs_memory_type_for_each(dst) {
+            ctx_rsc->queue_desc[src][dst].stream = NULL;
+            ucs_queue_head_init(&ctx_rsc->queue_desc[src][dst].event_queue);
```
yosefe, line 475: i think this is a problem to place event_queue inside the hash table value because when the hash size changes the back pointers become invalid

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -494,69 +505,61 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_copy_iface_t, uct_md_h md, uct_worker_h work
...
+    kh_init_inplace(cuda_copy_ctx_rscs, &self->ctx_rscs);
 
     ucs_queue_head_init(&self->active_queue);
 
+    return UCS_OK;
+}
+
+static void uct_cuda_copy_stream_destroy(CUstream *stream, int valid_ctx)
+{
+    if ((*stream == NULL) || (!valid_ctx)) {
```
yosefe, line 517: 1. extra ( )
2. we probably want to set *stream=NULL also is valid_ctx = 0. or not set it at all since ctx_rsc is destroyed
3. suggest rename stream to stream_p

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -494,69 +505,61 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_copy_iface_t, uct_md_h md, uct_worker_h work
...
+    ucs_memory_type_t src, dst;
+    ucs_queue_head_t *event_q;
+
     ucs_memory_type_for_each(src) {
         ucs_memory_type_for_each(dst) {
-            self->queue_desc[src][dst].stream = 0;
-            ucs_queue_head_init(&self->queue_desc[src][dst].event_queue);
+            event_q = &ctx_rsc->queue_desc[src][dst].event_queue;
+            if (!ucs_queue_is_empty(event_q)) {
+                ucs_warn("stream destroyed but queue not empty");
```
yosefe, line 535: ucs_warn("cuda context %d stream[%d][%d] being destroyed with %zu outstanding events", ctx_id, src, dst, ucs_queue_length(event_q));

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -494,69 +505,61 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_copy_iface_t, uct_md_h md, uct_worker_h work
...
-            }
-        }
-
-        if (self->short_stream) {
-            UCT_CUDADRV_FUNC_LOG_ERR(cuStreamDestroy(self->short_stream));
-        }
-    }
+    kh_foreach_key(&self->ctx_rscs, ctx_id, {
+        iter    = kh_get(cuda_copy_ctx_rscs, &self->ctx_rscs, ctx_id);
+        ctx_rsc = &kh_value(&self->ctx_rscs, iter);
```
yosefe, line 558: given the queue head/hash issue, maybe the hash table should point to ctx_rsc and not store the struct directly. then we can use kh_foreach_value here

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -49,22 +51,34 @@ typedef struct uct_cuda_copy_queue_desc {
...
+    /* stream used to issue short operations */
+    CUstream                   short_stream;
+    /* array of queue descriptors for each src/dst memory type combination */
+    uct_cuda_copy_queue_desc_t queue_desc[UCS_MEMORY_TYPE_LAST]
+                                         [UCS_MEMORY_TYPE_LAST];
+    /* CUDA context handle */
+    CUcontext                  ctx;
```
yosefe, line 63: minor: make it the 1st struct field

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -49,22 +51,34 @@ typedef struct uct_cuda_copy_queue_desc {
 } uct_cuda_copy_queue_desc_t;
 
 
+typedef struct uct_cuda_copy_ctx_rsc {
+    /* pool of cuda events to check completion of memcpy operations */
+    ucs_mpool_t                cuda_event_desc;
```
yosefe, line 56: maybe rename to event_mp?

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -49,22 +51,34 @@ typedef struct uct_cuda_copy_queue_desc {
...
+           kh_int64_hash_func, kh_int64_hash_equal);
+
+
 typedef struct uct_cuda_copy_iface {
     uct_cuda_iface_t            super;
     /* used to store uuid and check iface reachability */
     uct_cuda_copy_iface_addr_t  id;
-    /* pool of cuda events to check completion of memcpy operations */
-    ucs_mpool_t                 cuda_event_desc;
+    /* CUDA context resources */
```
yosefe, line 76: /* CUDA resources per context */

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -105,4 +119,19 @@ uct_cuda_copy_flush_bitmap_idx(ucs_memory_type_t src_mem_type,
...
+ * @param [in]  ctx       CUDA context handle
+ * @param [in]  ctx_id    CUDA context Id
+ * @param [out] ctx_rsc_p Returned pointer to context resources
+ *
+ * @return Error code as defined by @ref ucs_status_t.
+ */
+ucs_status_t uct_cuda_copy_ctx_rsc_create(uct_cuda_copy_iface_t *iface,
+                                          CUcontext ctx,
+                                          unsigned long long ctx_id,
+                                          uct_cuda_copy_ctx_rsc_t **ctx_rsc_p);
```
yosefe, line 136: space line after

test/gtest/uct/cuda/test_switch_cuda_device.cc
```diff
@@ -143,3 +144,68 @@ UCS_TEST_P(test_switch_cuda_device, detect_mem_type_cuda_fabric)
...
+    if (num_devices < 1) {
+        UCS_TEST_SKIP_R("no cuda devices available");
+    }
+
+    CUcontext current_ctx;
+    ASSERT_EQ(cuCtxGetCurrent(&current_ctx), CUDA_SUCCESS);
+
+    CUdevice device;
+    if (current_ctx != NULL) {
+        ASSERT_EQ(cuCtxGetDevice(&device), CUDA_SUCCESS);
```
yosefe, line 170: why not always use cuDeviceGet(&device, 0)?
IMO better to not use/assume anything about context being set from previous calls, and always create/use primary context explicitly.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
rakhmets: We need a context to store it in the hash map value, thus we need to call `cuCtxGetCurrent` here.

yosefe: > We need a context to store it in the hash map value, thus we need to call `cuCtxGetCurrent` here.

Only in slow path when we initi ctx_rsc

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -117,22 +114,21 @@ static UCS_F_ALWAYS_INLINE ucs_status_t uct_cuda_copy_ctx_rsc_get(
...
 
-    iter = kh_get(cuda_copy_ctx_rscs, &iface->ctx_rscs, (khint64_t)ctx);
+#if CUDA_VERSION >= 12000
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetId(ctx, &ctx_id));
+    if (ucs_unlikely(status != UCS_OK)) {
+        return status;
+    }
+#else
+    ctx_id = 0;
+#endif
```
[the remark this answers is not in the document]
rakhmets: Added.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
rakhmets: Updated hash map value type.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
rakhmets: The following command line doesn't hang: ucx_perftest -t ucp_am_lat -m cuda,host -s 400 -l.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -447,14 +430,21 @@ ucs_status_t uct_cuda_copy_ctx_rsc_create(uct_cuda_copy_iface_t *iface,
...
-                       "the key %llu has already been added", ctx_id);
+                       "the key has already been added iface=%p key=%llu",
+                       iface, ctx_id);
 
-    ctx_rsc = &kh_value(&iface->ctx_rscs, iter);
+    ctx_rsc = ucs_malloc(sizeof(*ctx_rsc), "uct_cuda_copy_ctx_rsc_t");
+    if (ctx_rsc == NULL) {
+        ucs_error("failed to allocate cuda context resource struct");
+        status = UCS_ERR_NO_MEMORY;
+        kh_del(cuda_copy_ctx_rscs, &iface->ctx_rscs, iter);
```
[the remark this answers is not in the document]
rakhmets: Done.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -101,35 +101,31 @@ uct_cuda_copy_get_mem_type(uct_md_h md, void *address, size_t length)
...
 
-    iter = kh_get(cuda_copy_ctx_rscs, &iface->ctx_rscs, ctx_id);
-    if (ucs_unlikely(iter == kh_end(&iface->ctx_rscs))) {
-        return uct_cuda_copy_ctx_rsc_create(iface, ctx, ctx_id, ctx_rsc_p);
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetCurrent(&ctx));
+    if (status != UCS_OK) {
+        return status;
+    } else if (ctx == NULL) {
+        ucs_error("no cuda context bound to calling thread");
+        return UCS_ERR_IO_ERROR;
```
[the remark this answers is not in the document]
rakhmets: Moved inside the method.

src/uct/cuda/cuda_copy/cuda_copy_ep.c
```diff
@@ -102,13 +102,11 @@ static UCS_F_ALWAYS_INLINE ucs_status_t uct_cuda_copy_ctx_rsc_get(
...
     unsigned long long ctx_id;
-    ucs_status_t status;
     khiter_t iter;
-    CUcontext ctx;
 
-    status = uct_cuda_base_get_ctx_id(NULL, &ctx_id);
-    if (ucs_unlikely(status != UCS_OK)) {
-        return status;
+    if (ucs_unlikely(uct_cuda_base_ctx_get_id(NULL, &ctx_id) != CUDA_SUCCESS)) {
+        ucs_error("failed to get the id of the current cuda context");
```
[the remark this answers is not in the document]
rakhmets: Fixed.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -322,7 +341,7 @@ static void uct_cuda_copy_event_desc_cleanup(ucs_mpool_t *mp, void *obj)
                                                         uct_cuda_copy_ctx_rsc_t,
                                                         event_mp);
 
-    if (uct_cuda_base_is_context_valid(ctx_rsc->ctx)) {
+    if (uct_cuda_copy_is_ctx_valid(ctx_rsc->ctx, ctx_rsc->ctx_id)) {
```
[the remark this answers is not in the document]
rakhmets: Fixed.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -314,6 +314,25 @@ uct_cuda_copy_event_desc_init(ucs_mpool_t *mp, void *obj, void *chunk)
...
+    if (result == CUDA_ERROR_CONTEXT_IS_DESTROYED) {
+        return 0;
+    } else if (result != CUDA_SUCCESS) {
+        UCT_CUDADRV_LOG(cuCtxGetId, UCS_LOG_LEVEL_WARN, result);
+        return 0;
+    }
+
+    return ctx_id == ctx_id_to_cmp;
+#else
+    return uct_cuda_base_is_context_valid(ctx);
```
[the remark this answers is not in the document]
rakhmets: Added the comment.

yosefe (approved):

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -49,22 +51,36 @@ typedef struct uct_cuda_copy_queue_desc {
...
+    /* stream used to issue short operations */
+    CUstream                   short_stream;
+    /* array of queue descriptors for each src/dst memory type combination */
+    uct_cuda_copy_queue_desc_t queue_desc[UCS_MEMORY_TYPE_LAST]
+                                         [UCS_MEMORY_TYPE_LAST];
+} uct_cuda_copy_ctx_rsc_t;
+
+
+/* Hash map for CUDA context resources. The key is the CUDA context Id. */
+KHASH_INIT(cuda_copy_ctx_rscs, unsigned long long, uct_cuda_copy_ctx_rsc_t*, 1,
```
brminich, line 70: can't we keep `uct_cuda_copy_ctx_rsc_t` in the hash (not a pointer) to avoid malloc/free in our code?
afaiu it was like that before in this PR, what is the reason to change?

yosefe: the event queue contains list head which gets corrupted when hash is resized

brminich (approved):

rakhmets:
> Some failures seem relevant:
> 
> ```
> 2025-03-18T09:10:10.6391431Z [ RUN      ] tcp/test_ucp_atomic32.misaligned_post/1 <tcp,cuda_copy,rocm_copy/device>
> 2025-03-18T09:10:10.7557476Z /__w/1/s/contrib/../test/gtest/ucp/test_ucp_memheap.cc:51: Failure
> 2025-03-18T09:10:10.7558631Z Error: Input/output error
> 2025-03-18T09:10:10.7559513Z [     INFO ] host->host ADD 
> 2025-03-18T09:10:10.7604080Z [  FAILED  ] tcp/test_ucp_atomic32.misaligned_post/1, where GetParam() = tcp,cuda_copy,rocm_copy/device (122 ms)
> 2025-03-18T09:10:10.7605274Z [ RUN      ] tcp/test_ucp_atomic32.post/1 <tcp,cuda_copy,rocm_copy/device>
> 2025-03-18T09:10:10.8674141Z [1742289010.866831] [swx-rdmz-ucx-gpu-02:31773:0]    cuda_copy_md.c:165  UCX  ERROR cuMemHostRegister_v2(address, length, 0x01) failed: context is destroyed
> 2025-03-18T09:10:10.8675852Z [1742289010.866876] [swx-rdmz-ucx-gpu-02:31773:0]          ucp_mm.c:76   UCX  ERROR failed to register address 0x62300021e900 (host) length 6404 on md[1]=cuda_cpy: Input/output error (md supports: host|cuda|cuda-managed)
> 2025-03-18T09:10:10.8677741Z /__w/1/s/contrib/../test/gtest/ucp/test_ucp_memheap.cc:51: Failure
> 2025-03-18T09:10:10.8678762Z Error: Input/output error
> 2025-03-18T09:10:10.8679729Z [     INFO ] host->host ADD 
> ```

Yes, the problem is in the new added test. It is not safe to call `cudaDeviceReset()` in the test. I will revert back to creating/destroying a separate context to test cleanup flow.

yosefe (approved):

rakhmets:
The approval was dismissed after resolving conflicts with recently merged PR https://github.com/openucx/ucx/pull/10155.

brminich (approved):

yosefe (approved):

PR #9646, opened 2024-01-31
GIT: Add ucx_perftest_daemon file to .gitignore

ivankochin:


.gitignore
```diff
@@ -48,6 +48,7 @@ ucx*tar.gz
 src/tools/info/build_config.h
 src/tools/info/ucx_info
 src/tools/perf/ucx_perftest
+src/tools/perf/ucx_perftest_daemon
```
iyastreb, line 51: It's already part of https://github.com/openucx/ucx/pull/9624
Thomas suggested this change a week ago
For me git does not complain because I build from a dedicated build-* folder (not from the WS root).

PR #9648, opened 2024-01-31
UCT/TCP: verbose assertion on connection establishment

evgeny-leksikov:
## What
verbose assertion on connection establishment

## Why ?
the patch was used in debugging of https://github.com/openucx/ucx/issues/9547


tvegas1 (approved):

src/uct/tcp/tcp_cm.c
```diff
@@ -597,7 +599,13 @@ uct_tcp_cm_handle_conn_req(uct_tcp_ep_t **ep_p,
             goto out_destroy_ep;
         }
     } else {
-        ucs_assert(uct_tcp_cm_ep_accept_conn(ep));
+        ucs_assertv(uct_tcp_cm_ep_accept_conn(ep),
```
yosefe, line 602: code inside uct_tcp_cm_ep_accept_conn looks very similar? maybe unite and print more info such as conn_state and result of compare?

evgeny-leksikov: yes but `cmp < 0` is not always bad case, printing message to a buffer on any result of compare will add extra overhead for good case. I added a func to reuse the code

src/uct/tcp/tcp_cm.c
```diff
@@ -416,12 +416,25 @@ void uct_tcp_cm_remove_ep(uct_tcp_iface_t *iface, uct_tcp_ep_t *ep)
     ep->flags &= ~UCT_TCP_EP_FLAG_ON_MATCH_CTX;
 }
 
+static const char*
+uct_tcp_cm_format_addrs_pair(char *buffer, size_t max_length, const char *format,
```
iyastreb, line 420: What about another approach:
```
#define SOCKADDR_TO_STR(sock_addr) \
    ({ \
        char buf[UCS_SOCKADDR_STRING_LEN + 1]; \
        ucs_sockaddr_str(sock_addr, buf, UCS_SOCKADDR_STRING_LEN); \
        buf; \
    })

ucs_assertv_always(..., "ucs_sockaddr_cmp(%s, %s) failed",
                    SOCKADDR_TO_STR(s1), SOCKADDR_TO_STR(s2))
```

iyastreb: No, this does not work, need to allocate with alloca instead:
```
#define SOCKADDR_TO_STR(sock_addr) \
    ({ \
        char *buf = ucs_alloca(UCS_SOCKADDR_STRING_LEN); \
        ucs_sockaddr_str(sock_addr, buf, UCS_SOCKADDR_STRING_LEN); \
        buf; \
    })

Comment (iyastreb) [reply src/uct/tcp/tcp_cm.c:420]: Wait, now it might not be safe to call `SOCKADDR_TO_STR` in `ucs_assertv_always`, because of:
> Do not use alloca inside the arguments of a function call
https://www.gnu.org/software/libc/manual/html_node/Variable-Size-Automatic.html#index-alloca-function

So the code would be:
```
s1_str = SOCKADDR_TO_STR(s1);
s2_str = SOCKADDR_TO_STR(s2);
ucs_assertv_always(..., "ucs_sockaddr_cmp(%s, %s) failed", s1_str, s2_str)

evgeny-leksikov: Won't `buf` be freed after its scope of visibility, i.e.  `}` inside of macro?

iyastreb: That's a compound expression, should be safe wrt alloca. The entire ucs_array is based on this idea:
```
#define UCS_ARRAY_ALLOC_ONSTACK(_array_type, _capacity) \
    ({ \
        typedef ucs_typeof(*((_array_type*)NULL)->buffer) value_t; \
        (value_t*)ucs_alloca((_capacity) * sizeof(value_t)); \
    })

Comment (evgeny-leksikov) [reply src/uct/tcp/tcp_cm.c:420]: Ok, just variable allocation on stack looks safe vs a macro which is expected to be used like a function, but there is a limitation that it can't be used as an argument of other function

Comment (yosefe) [reply src/uct/tcp/tcp_cm.c:420]: maybe pass ep as parameter instead of s1,s2?
`static const char *uct_tcp_cm_ep_addrs_str(uct_tcp_ep_t *ep, char *buffer, size_t max_length)`

File: src/uct/tcp/tcp_cm.c [hunk 3]
```diff
@@ -416,12 +416,34 @@ void uct_tcp_cm_remove_ep(uct_tcp_iface_t *iface, uct_tcp_ep_t *ep)
...
+    uct_tcp_iface_t *iface             = ucs_derived_of(ep->super.super.iface,
+                                                        uct_tcp_iface_t);
+    const struct sockaddr* local_addr  = (const void *)&iface->config.ifaddr;
+    const struct sockaddr* remote_addr = (const void *)&ep->peer_addr;
+    char remote_addr_str[UCS_SOCKADDR_STRING_LEN];
+    char local_addr_str[UCS_SOCKADDR_STRING_LEN];
+
+    ucs_assertv(max_length >=
+                (2 * UCS_SOCKADDR_STRING_LEN + sizeof(addrs_format)),
+                "a buffer is too small");
```
yosefe, line 432: IMO it's not needed

src/uct/tcp/tcp_cm.c
```diff
@@ -416,12 +416,34 @@ void uct_tcp_cm_remove_ep(uct_tcp_iface_t *iface, uct_tcp_ep_t *ep)
...
+    return buffer;
+}
+
 int uct_tcp_cm_ep_accept_conn(uct_tcp_ep_t *ep)
 {
     uct_tcp_iface_t *iface = ucs_derived_of(ep->super.super.iface,
                                             uct_tcp_iface_t);
-    char str_local_addr[UCS_SOCKADDR_STRING_LEN];
-    char str_remote_addr[UCS_SOCKADDR_STRING_LEN];
+    char err_str[UCS_SOCKADDR_STRING_LEN * 2 + 32];
```
yosefe, line 446: define a constant
`#define UCT_TCP_CM_SOCKADDR_STR_LEN ((UCS_SOCKADDR_STRING_LEN * 2) + 32)`

yosefe (approved):

PR #9472, opened 2023-11-07
UCT/IB/MLX5: Refactor KSM functions and extend logging

yosefe:
## Why
Improve KSM debug logging

PR #9649, opened 2024-01-31
UCT/IB: Fix address value when creating exporting key

brminich:
## What
Fixes a regression introduced by #9472, resulted in incorrect ksm start address


iyastreb (approved):
Great job! XGVMI is working again with this patch

yosefe (approved):

PR #9650, opened 2024-01-31
UCP/CORE: Fixed wiface activation counter.

rakhmets:
## What
Added usage counter to ep configuration structure to track the number of endpoints using the configuration.
Reverted https://github.com/openucx/ucx/pull/9104.
Activated UCP worker interfaces when setting endpoint configuration to an endpoint.

## Why ?
The changes are required to properly deactivate worker interface when it is unused.
Currently we may deactivate worker interface even if there are still active endpoints.

PR #9651, opened 2024-01-31
UCP/WIREUP: Activate wiface if lane is wireup message lane.

rakhmets:
## What
Activate wiface corresponding to the lane that is selected as wireup message lane.

## Why ?

## How ?
Previously tried to fix issue on UCT layer: https://github.com/openucx/ucx/pull/9500 and on UCP layer https://github.com/openucx/ucx/pull/9546.
This is another attempt to fix the hang in UCP flush. The issue can be fixed only if wiface activation counter will be fixed by https://github.com/openucx/ucx/pull/9650.


yosefe:
@rakhmets per the f2f discussion, can we close this PR?

rakhmets:
> @rakhmets per the f2f discussion, can we close this PR?

Yes we can. And I will re-open PR to not flush lanes with inactive interface.

PR #9652, opened 2024-01-31
UCP/WIREUP: Add rma_bw lanes in smaller chunks for each memtype

SeyedMir:
## What
Add rma_bw lanes in smaller chunks for each memtype.

## Why ?
The rma_bw lanes are added for each memtype, starting from host. Thus, the lanes added for host memory can exhaust the limit for the total number of lanes, leading to cuda_ipc lane to be missed. Adding the lanes in smaller chunks iteratively avoids this issue.

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1761,11 +1761,11 @@ ucp_wireup_add_am_bw_lanes(const ucp_wireup_select_params_t *select_params,
     }
 
     ucp_wireup_dev_usage_count_init(&dev_count);
-    dev_count.excl_lane = am_lane;
-    bw_info.dev_count   = &dev_count;
-    num_am_bw_lanes     = ucp_wireup_add_bw_lanes(select_params, &bw_info,
-                                                  ucp_tl_bitmap_max,
-                                                  select_ctx, 0);
+    dev_count.excl_lane       = am_lane;
+    bw_info.dev_count         = &dev_count;
+    num_am_bw_lanes = ucp_wireup_add_bw_lanes(select_params, &bw_info,
+                                              ucp_tl_bitmap_max,
+                                              select_ctx, 0);
     return ((am_lane != UCP_NULL_LANE) || (num_am_bw_lanes > 0)) ? UCS_OK :
            UCS_ERR_UNREACHABLE;
 }
@@ -1892,7 +1892,7 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
     dev_count = ucs_malloc(UCS_MEMORY_TYPE_LAST * sizeof(*dev_count),
                            "rma_bw_lanes_dev_count");
     if (dev_count == NULL) {
-        ucs_log(UCS_LOG_LEVEL_ERROR, "failed to allocate dev_count");
+        ucs_error("failed to allocate dev_count");
         return UCS_ERR_NO_MEMORY;
     }
 
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1846,6 +1846,7 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
         UCP_RNDV_MODE_PUT_ZCOPY
     };
     ucp_wireup_dev_usage_count *dev_count;
+    ucp_wireup_dev_usage_count dev_count_rkey_ptr;
     ucp_wireup_select_bw_info_t bw_info;
     ucs_memory_type_t mem_type;
     unsigned max_lanes, lanes_batch;
@@ -1889,13 +1890,6 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
 
     bw_info.md_map = 0;
 
-    dev_count = ucs_malloc(UCS_MEMORY_TYPE_LAST * sizeof(*dev_count),
-                           "rma_bw_lanes_dev_count");
-    if (dev_count == NULL) {
-        ucs_error("failed to allocate dev_count");
-        return UCS_ERR_NO_MEMORY;
-    }
-
     /* check rkey_ptr */
     if (!(ep_init_flags & UCP_EP_INIT_FLAG_MEM_TYPE) &&
          (context->config.ext.rndv_mode == UCP_RNDV_MODE_AUTO)) {
@@ -1906,14 +1900,14 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
          * Allow selecting additional lanes in case the remote memory will not be
          * registered with this memory domain, i.e with GPU memory.
          */
-        ucp_wireup_dev_usage_count_init(&dev_count[0]);
+        ucp_wireup_dev_usage_count_init(&dev_count_rkey_ptr);
         bw_info.local_dev_bitmap           = UINT64_MAX;
         bw_info.remote_dev_bitmap          = UINT64_MAX;
         bw_info.criteria.title             = "obtain remote memory pointer";
         bw_info.criteria.local_cmpt_flags |= UCT_COMPONENT_FLAG_RKEY_PTR;
         bw_info.criteria.lane_type         = UCP_LANE_TYPE_RKEY_PTR;
-        bw_info.criteria.arg               = &dev_count[0];
-        bw_info.dev_count                  = &dev_count[0];
+        bw_info.criteria.arg               = &dev_count_rkey_ptr;
+        bw_info.dev_count                  = &dev_count_rkey_ptr;
         bw_info.max_lanes                  = 1;
 
         UCP_CONTEXT_MEM_CAP_TLS(context, UCS_MEMORY_TYPE_HOST, access_mem_types,
@@ -1973,6 +1967,13 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
             allow_extra_path      = 1;
             break;
         }
+    }
+
+    dev_count = ucs_malloc(UCS_MEMORY_TYPE_LAST * sizeof(*dev_count),
+                           "rma_bw_lanes_dev_count");
+    if (dev_count == NULL) {
+        ucs_error("failed to allocate dev_count");
+        return UCS_ERR_NO_MEMORY;
     }
 
     ucs_memory_type_for_each(mem_type) {
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1996,9 +1996,9 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
         /* Add lanes that can access the memory by short operations */
         added_lanes = 0;
 
-        while (max_lanes > 0) {
+        while (added_lanes < max_lanes) {
             added_lanes_prev  = added_lanes;
-            bw_info.max_lanes = ucs_min(max_lanes, lanes_batch);
+            bw_info.max_lanes = ucs_min(max_lanes - added_lanes, lanes_batch);
             UCS_STATIC_BITMAP_RESET_ALL(&tl_bitmap);
             ucs_memory_type_for_each(mem_type) {
                 UCP_CONTEXT_MEM_CAP_TLS(context, mem_type, reg_mem_types,
@@ -2020,7 +2020,6 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
             if (added_lanes == added_lanes_prev) {
                 break;
             }
-            max_lanes -= bw_info.max_lanes;
         }
 
         if (added_lanes /* There are selected lanes */ ||
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1889,7 +1889,7 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
          * Allow selecting additional lanes in case the remote memory will not be
          * registered with this memory domain, i.e with GPU memory.
          */
-        memset(&dev_count, 0, sizeof(dev_count));
+        memset(dev_count, 0, sizeof(dev_count));
         bw_info.local_dev_bitmap           = UINT64_MAX;
         bw_info.remote_dev_bitmap          = UINT64_MAX;
         bw_info.criteria.title             = "obtain remote memory pointer";
@@ -1905,7 +1905,7 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
                                 UCP_NULL_LANE, select_ctx, 0);
     }
 
-    memset(&dev_count, 0, sizeof(dev_count));
+    memset(dev_count, 0, sizeof(dev_count));
     bw_info.local_dev_bitmap          = UINT64_MAX;
     bw_info.remote_dev_bitmap         = UINT64_MAX;
     bw_info.criteria.title            = "high-bw remote memory access";
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1518,7 +1518,8 @@ ucp_wireup_add_fast_lanes(ucp_worker_h worker,
     /* Calculate max BW of existing lanes of the corresponding mem type */
     for (lane_desc = select_ctx->lane_descs;
      lane_desc < select_ctx->lane_descs + select_ctx->num_lanes; ++lane_desc) {
-        if (lane_desc->rsc_index != UCP_NULL_RESOURCE) {
+        if ((lane_desc->rsc_index != UCP_NULL_RESOURCE) &&
+            (lane_desc->lane_types & UCS_BIT(lane_type))) {
             md_index = context->tl_rscs[lane_desc->rsc_index].md_index;
             md_attr  = &context->tl_mds[md_index].attr;
             if (md_attr->reg_mem_types & lane_reg_mem_types) {
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1723,33 +1723,6 @@ ucp_wireup_add_am_bw_lanes(const ucp_wireup_select_params_t *select_params,
         return UCS_OK;
     }
 
-    /* Select one lane for active messages */
-    ucp_wireup_criteria_init(&bw_info.criteria);
-    bw_info.criteria.title              = "high-bw active messages";
-    bw_info.criteria.calc_score         = ucp_wireup_am_bw_score_func;
-    bw_info.criteria.lane_type          = UCP_LANE_TYPE_AM_BW;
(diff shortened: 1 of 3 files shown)


SeyedMir:
@brminich One issue I'm running to is that local and remote device bitmaps will not be preserved in different iterations of the new loop. The is because we don't modify the bitmaps in bw_info; we modify them locally in add_bw_lanes. However, if I update add_bw_lanes so that we modify those bitmaps in the bw_info, then updates for one memtype will affect the bitmap for other memtypes too.

src/ucp/wireup/select.c
```diff
@@ -1872,20 +1877,27 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
-            UCS_BITMAP_OR_INPLACE(&tl_bitmap, mem_type_tl_bitmap);
+        while (lanes_iter--) {
+            added_lanes_prev = added_lanes;
+            UCS_BITMAP_CLEAR(&tl_bitmap);
+            ucs_memory_type_for_each(mem_type) {
+                UCP_CONTEXT_MEM_CAP_TLS(context, mem_type, reg_mem_types,
+                                        mem_type_tl_bitmap);
+
+                bw_info.criteria.reg_mem_types = UCS_BIT(mem_type);
+                added_lanes                   += ucp_wireup_add_bw_lanes(
```
tvegas1, line 1889: due to `dev_count` being declared inside `ucp_wireup_add_bw_lanes()`, when now calling multiple times for a given memtype, i think we could end-up failing to track local and remote max usage count against `dev_num_path` properly?

brminich:
> @brminich One issue I'm running to is that local and remote device bitmaps will not be preserved in different iterations of the new loop. The is because we don't modify the bitmaps in bw_info; we modify them locally in add_bw_lanes. However, if I update add_bw_lanes so that we modify those bitmaps in the bw_info, then updates for one memtype will affect the bitmap for other memtypes too.

i think it should be ok, because if the same device is suitable for several memory types, we would want to initiate just one set of lanes.

src/ucp/wireup/select.c
SeyedMir: That's right. Maybe we should initialize define `dev_count` in add_rma_bw (and other functions making call to add_bw_lanes) and pass it to add_bw_lanes.

SeyedMir:
@yosefe @brminich can you please take another look at this PR so we can have it merged for 1.17? Thanks

brminich (commented):
general comment is that we need to restrict this change to versions v1.17 or newer to preserve wire compat

src/ucp/wireup/select.c
```diff
@@ -1836,14 +1837,18 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
     bw_info.criteria.title            = "high-bw remote memory access";
     bw_info.criteria.lane_type        = UCP_LANE_TYPE_RMA_BW;
     bw_info.criteria.local_cmpt_flags = 0;
+    bw_info.max_lanes                 = ucs_min(2, context->config.ext.max_rndv_lanes);
```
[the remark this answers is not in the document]
SeyedMir: That's an outdated change. I currently set `bw_info.max_lanes = 1` to add lanes one by one.

src/ucp/wireup/select.c
```diff
@@ -1876,17 +1901,25 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
+        max_lanes   = UCP_PROTO_MAX_LANES;
+        lanes_batch = context->config.ext.rma_bw_lanes_batch;
     } else {
-        bw_info.max_lanes = context->config.ext.max_rndv_lanes;
+        max_lanes   = lanes_batch = context->config.ext.max_rndv_lanes;
     }
 
+    bw_info.max_lanes = lanes_batch;
+    lanes_batch_rem   = max_lanes % lanes_batch;
+    lanes_batch_num   = (max_lanes / lanes_batch) + (lanes_batch_rem ? 1 : 0);
```
yosefe, line 1921: this approach of counting is more error-prone,
should be something like:
```
while (max_lanes > 0) {
  num_lanes = ... add_lanes(.., ucs_min(max_lanes, lanes_latch)) ...
  max_lanes -= num_lanes;
}
```
also, may need to modify how added_lanes is used

src/ucp/wireup/select.c
```diff
@@ -1507,18 +1509,39 @@ ucp_wireup_add_fast_lanes(ucp_worker_h worker,
...
+            md_index = context->tl_rscs[lane_desc->rsc_index].md_index;
+            md_attr  = &context->tl_mds[md_index].attr;
+            if (md_attr->reg_mem_types & lane_reg_mem_types) {
+                lane_bw = ucp_wireup_get_lane_bw(worker, lane_desc->rsc_index,
+                                                 lane_desc->addr_index,
+                                                 select_params->address);
+                max_bw  = ucs_max(lane_bw, max_bw);
+            }
+        }
+    }
```
yosefe, line 1531: how about keeping max_bw in ucp_wireup_select_context_t?
in any case, seems redundant to do bot this loop and also the loop in lines 1534-1539

src/ucp/wireup/select.c
SeyedMir: Very good suggestion. Updated accordingly.

src/ucp/wireup/select.c
SeyedMir: I tried to do that, but the problem is that we're dealing with a mem_type bitmap, not a single mem_type. The `ucp_wireup_criteria_t` in `bw_info` has a `reg_mem_types` bitmap. Though, it seems like that bitmap is set to only one mem_type in the existing code.

src/ucp/wireup/select.c
```diff
@@ -1570,28 +1593,22 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
                         ucp_wireup_select_context_t *select_ctx,
                         unsigned allow_extra_path)
 {
-    ucp_rsc_index_t skip_dev_index       = UCP_NULL_RESOURCE;
-    ucp_ep_h ep                          = select_params->ep;
-    ucp_context_h context                = ep->worker->context;
-    ucp_wireup_dev_usage_count dev_count = {};
+    ucp_rsc_index_t skip_dev_index        = UCP_NULL_RESOURCE;
```
tvegas1, line 1596: i think `skip_dev_index` must also be maintained at same level as `dev_count`.

src/ucp/wireup/select.c
```diff
@@ -1678,6 +1696,8 @@ ucp_wireup_add_am_bw_lanes(const ucp_wireup_select_params_t *select_params,
     ucp_context_h context  = ep->worker->context;
     unsigned ep_init_flags = ucp_wireup_ep_init_flags(select_params,
                                                       select_ctx);
+    ucp_wireup_dev_usage_count dev_count = {};
```
tvegas1, line 1699: We might have to track `dev_count` per memory type. Previously we were starting with `dev_count` local and remote equal to zero for each mem type.
Now we can have host local from 0 to 2, cuda from 2 to 4, local from 4 to 6...

src/ucp/wireup/select.c
```diff
@@ -2006,6 +2000,7 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
             if (added_lanes == added_lanes_prev) {
                 break;
             }
+            max_lanes -= bw_info.max_lanes;
```
yosefe, line 2003: I think we need to remove the actual number of lanes we have
for example, we can avoid updating max_lanes here, and line 1938 would be:
`bw_info.max_lanes = ucs_min(max_lanes - added_lanes, lanes_batch);`
or, we can add new local variable "num_lanes" that would save the value returned from ucp_wireup_add_bw_lanes, and here we would do:
`max_lanes -= num_lanes`

src/ucp/wireup/select.c
SeyedMir: I asked about this before and @brminich said it should be ok https://github.com/openucx/ucx/pull/9652#issuecomment-1920994559

src/ucp/wireup/select.c
SeyedMir: I'm not sure if I understand. The loop condition won't change if we don't update `max_lanes`. Also, decreasing `max_lanes` by `added_lanes` (instead of `bw_info.max_lanes`) will result in additional and unnecessary iteration of the `while` loop because `added_lanes` can be smaller than `bw_info.max_lanes`.

src/ucp/wireup/select.c
SeyedMir: But unlike `dev_count`, we don't need to know about the `skip_dev_index` from one iteration to another. It will always be the same, right?

src/ucp/wireup/select.c
```diff
@@ -1507,18 +1509,39 @@ ucp_wireup_add_fast_lanes(ucp_worker_h worker,
...
+    const ucp_wireup_lane_desc_t *lane_desc;
+    const uct_md_attr_v2_t *md_attr;
+
+    /* Calculate max BW of existing lanes of the corresponding mem type */
+    for (lane_desc = select_ctx->lane_descs;
+     lane_desc < select_ctx->lane_descs + select_ctx->num_lanes; ++lane_desc) {
+        if (lane_desc->rsc_index != UCP_NULL_RESOURCE) {
+            md_index = context->tl_rscs[lane_desc->rsc_index].md_index;
+            md_attr  = &context->tl_mds[md_index].attr;
+            if (md_attr->reg_mem_types & lane_reg_mem_types) {
```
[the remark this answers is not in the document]
SeyedMir: Added the check for `lane_type`.

src/ucp/wireup/select.c
tvegas1: after reading comment, I synced with @brminich.

src/ucp/wireup/select.c
tvegas1: yes we have one am lane and we need to exclude it once. with two rock nodes i see we that we fail to create the second path with run below:
- ` UCX_TLS=dc,cuda  UCX_RMA_BW_LANES_BATCH=1 ./rfs/bin/ucx_perftest -t tag_bw rock05`

src/ucp/wireup/select.c
SeyedMir: @tvegas1 I made changes to track `dev_count` per memory type because otherwise, `am_lane` exclusion for one memory type will incorrectly increase `dev_count` for every other memory type.
Also, I had to set `am_lane` to `UCP_NULL_LANE` after the first iteration of rma_bw lanes addition. Otherwise, `dev_count` will be wrongfully incremented in each iteration.

src/ucp/wireup/select.c
```diff
@@ -1833,7 +1833,7 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
         UCP_RNDV_MODE_GET_ZCOPY,
         UCP_RNDV_MODE_PUT_ZCOPY
     };
-    ucp_wireup_dev_usage_count dev_count;
+    ucp_wireup_dev_usage_count dev_count[UCS_MEMORY_TYPE_LAST];
```
tvegas1, line 1838: can remove `&` in front of each `dev_count` in memcpy

src/ucp/wireup/select.c
```diff
@@ -1648,25 +1669,24 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
 
             /* Remove that skipped path from score computation */
-            dev_count.local_skip[dev_index]      = 1;
-            dev_count.remote_skip[ae->dev_index] = 1;
+            dev_count->local_skip[dev_index]      = 1;
+            dev_count->remote_skip[ae->dev_index] = 1;
         }
 
-        if (dev_count.local[dev_index] >= local_num_paths) {
+        if (dev_count->local[dev_index] >= local_num_paths) {
```
tvegas1, line 1676: we also need to make sure that local_num_paths/remote_num_paths also get +1 line 1668/1669. this can be done by also moving skip_dev_index at same level as dev_count, maybe there is simpler way like checking local_skip/remote_skip

src/ucp/wireup/select.c
```diff
@@ -1729,9 +1760,12 @@ ucp_wireup_add_am_bw_lanes(const ucp_wireup_select_params_t *select_params,
         }
     }
 
-    num_am_bw_lanes = ucp_wireup_add_bw_lanes(select_params, &bw_info,
-                                              ucp_tl_bitmap_max, am_lane,
-                                              select_ctx, 0);
+    ucp_wireup_dev_usage_count_init(&dev_count);
+    dev_count.excl_lane = am_lane;
+    bw_info.dev_count   = &dev_count;
```
yosefe, line 1765: pls move these 3 lines right before line 1740

src/ucp/wireup/select.c
```diff
@@ -1851,9 +1887,14 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
 
-    bw_info.local_dev_bitmap  = UINT64_MAX;
-    bw_info.remote_dev_bitmap = UINT64_MAX;
-    bw_info.md_map            = 0;
+    bw_info.md_map = 0;
+
+    dev_count = ucs_malloc(UCS_MEMORY_TYPE_LAST * sizeof(*dev_count),
+                           "rma_bw_lanes_dev_count");
+    if (dev_count == NULL) {
+        ucs_log(UCS_LOG_LEVEL_ERROR, "failed to allocate dev_count");
```
yosefe, line 1895: ucs_error

src/ucp/wireup/select.c
```diff
@@ -1865,26 +1906,34 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
          * Allow selecting additional lanes in case the remote memory will not be
          * registered with this memory domain, i.e with GPU memory.
          */
+        ucp_wireup_dev_usage_count_init(&dev_count[0]);
```
yosefe, line 1909: use &dev_count[UCS_MEMORY_TYPE_HOST], or define another variable - rkey_ptr_dev_count - specifically for using it in rkey_ptr lane selection

src/ucp/wireup/select.c
```diff
@@ -1865,26 +1906,34 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
 
         UCP_CONTEXT_MEM_CAP_TLS(context, UCS_MEMORY_TYPE_HOST, access_mem_types,
                                 tl_bitmap);
         ucp_wireup_add_bw_lanes(select_params, &bw_info, tl_bitmap,
-                                UCP_NULL_LANE, select_ctx, 0);
+                                select_ctx, 0);
     }
 
+    bw_info.local_dev_bitmap          = UINT64_MAX;
+    bw_info.remote_dev_bitmap         = UINT64_MAX;
```
yosefe, line 1926: why dupliced these 2 lines instead of keeping them as they were?

src/ucp/wireup/select.c
```diff
@@ -1926,6 +1975,11 @@ ucp_wireup_add_rma_bw_lanes(const ucp_wireup_select_params_t *select_params,
         }
     }
 
+    ucs_memory_type_for_each(mem_type) {
+        ucp_wireup_dev_usage_count_init(&dev_count[mem_type]);
```
yosefe, line 1979: 1. maybe it would be nicer to pass excl_lane as a parameter to ucp_wireup_dev_usage_count_init
2. I think we need to do it before line 2006, inside the loop that goes over rndv_modes

src/ucp/wireup/select.c
```diff
@@ -56,23 +56,26 @@ typedef struct {
...
 } ucp_wireup_dev_usage_count;
 
 
+typedef struct {
+    ucp_wireup_criteria_t criteria;
+    uint64_t                   local_dev_bitmap;
+    uint64_t                   remote_dev_bitmap;
+    ucp_md_map_t               md_map;
+    unsigned                   max_lanes;
+    ucp_wireup_dev_usage_count *dev_count;
```
yosefe, line 75: why needed? it's already in criteria.arg
then will also not need to move this sruct definition

src/ucp/wireup/select.c
```diff
@@ -1618,55 +1648,54 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
...
-            local_dev_bitmap &= ~UCS_BIT(dev_index);
+            bw_info->local_dev_bitmap &= ~UCS_BIT(dev_index);
         }
 
         /* Assume symmetric num_paths */
-        if (dev_count.remote[ae->dev_index] >= remote_num_paths) {
+        if (dev_count->remote[ae->dev_index] >= remote_num_paths) {
             /* exclude remote device if reached max concurrency level */
-            remote_dev_bitmap &= ~UCS_BIT(ae->dev_index);
+            bw_info->remote_dev_bitmap &= ~UCS_BIT(ae->dev_index);
```
tvegas1, line 1693: i think `local_dev_bitmap` and `remote_dev_bit_map` also has to be tracked by mem_type. with this change, a given mem_type can now consume paths of another mem_type, since `bw_info` is global to all mem_types.

src/ucp/wireup/select.c
SeyedMir: Done. But had to move up am_lane finding logic as well.

src/ucp/wireup/select.c
SeyedMir: Done.

src/ucp/wireup/select.c
SeyedMir: Done.

src/ucp/wireup/select.c
SeyedMir: Because those bitmaps are updated by `add_bw_lanes` so we need to reset them after rkey_ptr.

src/ucp/wireup/select.c
SeyedMir: Done.

src/ucp/wireup/select.c
tvegas1: yeah maybe we can just add array of struct called `ucp_wireup_add_bw_lanes_state_t` instantiated per mem_type above `ucp_wireup_add_bw_lane()` and move dev_count and others inside.

SeyedMir: Yes, please see my latest commit and let me know.

SeyedMir:
I've been debugging a case where the selected lanes are different with smaller vs. larger batch size. The root cause of it is that the `ucp_wireup_rma_bw_score_func()` and `ucp_wireup_add_fast_lanes()` calculate the bandwidth differently. As a result, the rma_bw lane selected by `ucp_wireup_select_transport` (which will be of highest score) may not be the lane with the highest bandwidth (from `ucp_wireup_add_fast_lanes()` point of view).

Akshay-Venkatesh:
Closing this PR as https://github.com/openucx/ucx/pull/9814 addresses the original problem.

PR #9655, opened 2024-02-02
TEST/UCT: add imp/exp memh testing in uct_p2p_rma_test

evgeny-leksikov:
## What
add imp/exp memh testing in uct_p2p_rma_test

## Why ?
test coverage

## How ?
create exported rkey for non-loopback entities

### DIFF
--- a/test/gtest/uct/uct_test.cc
+++ b/test/gtest/uct/uct_test.cc
@@ -820,7 +820,7 @@ void uct_test::reduce_tl_send_queues()
 
 uct_test::entity::entity(const resource& resource, uct_iface_config_t *iface_config,
                          uct_iface_params_t *params, uct_md_config_t *md_config) :
-    m_resource(resource)
+    m_resource(resource), m_exp_memh(NULL)
 {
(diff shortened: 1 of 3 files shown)


test/gtest/uct/uct_test.cc
```diff
@@ -1018,6 +1018,48 @@ void uct_test::entity::rkey_release(const uct_rkey_bundle *rkey_bundle) const
...
+void uct_test::entity::memh_attach(void *rkey_buffer,
+                                   uct_rkey_bundle *rkey_bundle)
+{
+    uct_md_mem_attach_params_t attach_params = {0};
+    ucs_status_t status                      =
+            uct_md_mem_attach(m_md, rkey_buffer, &attach_params, &m_exp_memh);
+    ASSERT_UCS_OK(status);
+    free(rkey_buffer);
+
+    void *rkey_repack_buffer = malloc(md_attr().rkey_packed_size);
```
iyastreb, line 1052: I would consider std::array or std::vector here for auto memory handling
It will be also exception safe

PR #9658, opened 2024-02-04
TEST/JENKINS: Don't fail if nvidia-smi does not find devices

yosefe:
## Why
try_load_cuda_env() should not cause a fatal failure if no cuda devices are present


yosefe:
@rakhmets can you pls review?

rakhmets (approved):

PR #9660, opened 2024-02-06
UCP/PROTO: Calculate all stages perf during the final proto initialization

ivankochin:
## What
This is the first stage of MULTI mode performance calculation rework. Includes CTRL messages overheads to RNDV protocols performance. Removes performance calculation from RTS/RTR protocols, they became fully relied on `remote_proto_lookup` output.

## Why ?
Now such protocols as `rndv/get/zcopy` and `rndv/put/zcopy` during the performance estimations take into attention only RNDV + ACK stages. But for more precise calculation CTRL messages overhead should participate in proto performance too, since they contains both sender and receiver overhead.

### DIFF
--- a/src/ucp/proto/proto_init.c
+++ b/src/ucp/proto/proto_init.c
@@ -188,13 +188,12 @@ ucp_proto_perf_envelope_make(const ucp_proto_perf_list_t *perf_list,
 }
 
 ucs_status_t
-ucp_proto_init_parallel_stages(const ucp_proto_init_params_t *params,
-                               size_t range_start, size_t range_end,
+ucp_proto_init_parallel_stages(size_t range_start, size_t range_end,
                                size_t frag_size, double bias,
                                const ucp_proto_perf_range_t **stages,
-                               unsigned num_stages)
+                               unsigned num_stages, const char *proto_name,
+                               ucp_proto_caps_t *caps)
 {
-    ucp_proto_caps_t *caps      = params->caps;
     ucs_linear_func_t bias_func = ucs_linear_func_make(0.0, 1.0 - bias);
     UCS_ARRAY_DEFINE_ONSTACK(ucp_proto_perf_envelope_t, concave, 16);
     UCS_ARRAY_DEFINE_ONSTACK(ucp_proto_perf_list_t, stage_list, 16);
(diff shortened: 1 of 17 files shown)


src/ucp/proto/proto_init.c
```diff
@@ -188,13 +188,12 @@ ucp_proto_perf_envelope_make(const ucp_proto_perf_list_t *perf_list,
...
 
 ucs_status_t
-ucp_proto_init_parallel_stages(const ucp_proto_init_params_t *params,
-                               size_t range_start, size_t range_end,
+ucp_proto_init_parallel_stages(size_t range_start, size_t range_end,
                                size_t frag_size, double bias,
                                const ucp_proto_perf_range_t **stages,
-                               unsigned num_stages)
+                               unsigned num_stages, const char *proto_name,
+                               ucp_proto_caps_t *caps)
```
iyastreb, line 195: caps can be const?

src/ucp/proto/proto_init.c
```diff
@@ -188,13 +188,12 @@ ucp_proto_perf_envelope_make(const ucp_proto_perf_list_t *perf_list,
 }
 
 ucs_status_t
-ucp_proto_init_parallel_stages(const ucp_proto_init_params_t *params,
-                               size_t range_start, size_t range_end,
+ucp_proto_init_parallel_stages(size_t range_start, size_t range_end,
```
iyastreb, line 191: IMO it was better with single `params` argument, now it's already 8 args
It is also consistent with the rest of the functions taking the whole structure.
This way it's easier to introduce new changes without breaking the code
But up to you

src/ucp/rndv/proto_rndv.c
```diff
@@ -306,85 +363,47 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
-            goto out_deref_perf_node;
-        }
-
-        ucp_proto_perf_node_deref(&remote_perf.node);
+        range  = &caps->ranges[caps->num_ranges];
+        *range = *remote_range;
 
-        min_length = range_max_length - 1;
-    } while ((remote_range++)->max_length < max_length);
+        range->node = ucp_proto_perf_node_new_data(init_params->proto_name,
```
iyastreb, line 397: should we check for NULL return here?

PR #9664, opened 2024-02-06
UCP/FLUSH: Proceed to next EP if current flushed immediately

Artemy-Mellanox:
## Why ?
Fix [3677246](https://redmine.mellanox.com/issues/3677246) (internal issue)


src/ucp/rma/flush.c
```diff
@@ -737,7 +743,8 @@ static ucs_status_t ucp_worker_fence_strong(ucp_worker_h worker)
 
     do {
         ucp_worker_progress(worker);
-    } while (request->flush_worker.comp_count > 1);
+    } while ((request->flush_worker.comp_count > 1) ||
+             !ucp_worker_flush_all_eps_scheduled(request, worker));
```
brminich, line 747: do you think it is worth to swap conditions for better perf?

brminich (dismissed):

src/ucp/rma/flush.c
yosefe: so it means request->flush_worker.comp_count is not really checking flush completion?
maybe we can wait for flush request completion in a better way?

yosefe: discussed offline to make this function to reuse code from blocking flush

src/ucp/rma/flush.c
```diff
@@ -720,31 +720,10 @@ static ucs_status_t ucp_worker_fence_weak(ucp_worker_h worker)
...
-    if (ucs_unlikely(!UCS_PTR_IS_PTR(status_ptr))) {
-        status = UCS_PTR_STATUS(status_ptr);
-        goto out;
-    }
-
-    request = (ucp_request_t*)status_ptr - 1;
-
-    do {
-        ucp_worker_progress(worker);
-    } while (request->flush_worker.comp_count > 1);
```
brminich, line 740: this condition was an issue, right?

brminich (approved):

yosefe (approved):

yosefe:
@Artemy-Mellanox pls squash

PR #9665, opened 2024-02-06
UCT/IB/DC: dynamic allocation of dci pools by configuration

roiedanino:
## What
Instead of creating all dci pools (and dcis) on iface init, dci pools will be created dynamically on demand by configuration

## Why ?
As part of the active messages priority feature - creating different pools with different SLs is required, enabling dedicated dci pools for different lanes with different priorities, as we can't change SL of a dci (qp) after creation / connection.

## How ?
For each new dc ep created - a matching dci pool will be created unless a dci pool with the same exact configuration already exists in the hash table, in which the configuration is the key and the value is pool_index, so eps which share configuration will also share a dci pool.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.c
+++ b/src/uct/ib/dc/dc_mlx5.c
@@ -146,6 +146,12 @@ ucs_config_field_t uct_dc_mlx5_iface_config_table[] = {
     {NULL}
 };
 
+static uint8_t
+uct_dc_mlx5_max_rd_atomic(const uct_dc_mlx5_iface_addr_t *if_addr)
+{
+    return (if_addr->flags & UCT_DC_MLX5_IFACE_ADDR_MAX_RD_ATOMIC_16) ? 16 : 64;
+}
+
 static ucs_status_t
 uct_dc_mlx5_ep_create_connected(const uct_ep_params_t *params, uct_ep_h* ep_p)
 {
@@ -153,11 +159,13 @@ uct_dc_mlx5_ep_create_connected(const uct_ep_params_t *params, uct_ep_h* ep_p)
                                                 uct_dc_mlx5_iface_t);
     const uct_ib_address_t *ib_addr;
     const uct_dc_mlx5_iface_addr_t *if_addr;
+    uct_dc_mlx5_dci_config_t dci_config;
     ucs_status_t status;
     int is_global;
     uct_ib_mlx5_base_av_t av;
     struct mlx5_grh_av grh_av;
     unsigned path_index;
+    uint8_t max_rd_atomic, if_addr_max_rd_atomic;
 
     ucs_trace_func("");
 
@@ -165,6 +173,12 @@ uct_dc_mlx5_ep_create_connected(const uct_ep_params_t *params, uct_ep_h* ep_p)
     ib_addr    = (const uct_ib_address_t *)params->dev_addr;
     if_addr    = (const uct_dc_mlx5_iface_addr_t *)params->iface_addr;
     path_index = UCT_EP_PARAMS_GET_PATH_INDEX(params);
+
+    if_addr_max_rd_atomic = uct_dc_mlx5_max_rd_atomic(if_addr);
+    max_rd_atomic         = ucs_min(iface->super.super.config.max_rd_atomic,
+                                    if_addr_max_rd_atomic);
+
+    uct_dc_mlx5_init_dci_config(&dci_config, path_index, max_rd_atomic);
 
     status = uct_ud_mlx5_iface_get_av(&iface->super.super.super,
                                       &iface->ud_common, ib_addr, path_index,
@@ -175,10 +189,10 @@ uct_dc_mlx5_ep_create_connected(const uct_ep_params_t *params, uct_ep_h* ep_p)
 
     if (is_global) {
         return UCS_CLASS_NEW(uct_dc_mlx5_grh_ep_t, ep_p, iface, if_addr, &av,
-                             path_index, &grh_av);
+                             path_index, &grh_av, &dci_config);
     } else {
         return UCS_CLASS_NEW(uct_dc_mlx5_ep_t, ep_p, iface, if_addr, &av,
-                             path_index);
+                             path_index, &dci_config);
     }
 }
 
@@ -253,6 +267,7 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface, int poll_flags)
     uint8_t dci_index;
     struct mlx5_cqe64 *cqe;
     uint16_t hw_ci;
+    uct_dc_dci_t *dci;
     UCT_DC_MLX5_TXQP_DECL(txqp, txwq);
 
     cqe = uct_ib_mlx5_poll_cq(&iface->super.super.super,
@@ -267,8 +282,10 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface, int poll_flags)
     ucs_memory_cpu_load_fence();
 
     dci_index = uct_dc_mlx5_iface_dci_find(iface, cqe);
-    txqp      = &iface->tx.dcis[dci_index].txqp;
-    txwq      = &iface->tx.dcis[dci_index].txwq;
+    dci       = uct_dc_mlx5_iface_dci(iface, dci_index);
+    ucs_assert_always(uct_dc_mlx5_is_dci_valid(dci));
+    txqp      = &dci->txqp;
+    txwq      = &dci->txwq;
     hw_ci     = ntohs(cqe->wqe_counter);
 
     ucs_trace_poll("dc iface %p tx_cqe: dci[%d] txqp %p hw_ci %d",
@@ -282,8 +299,7 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface, int poll_flags)
      *       to avoid OOO sends when this is the only missing resource.
      */
     uct_dc_mlx5_iface_dci_put(iface, dci_index);
-    uct_dc_mlx5_iface_progress_pending(iface,
-                                       iface->tx.dcis[dci_index].pool_index);
+    uct_dc_mlx5_iface_progress_pending(iface, dci->pool_index);
     uct_dc_mlx5_iface_check_tx(iface);
     uct_ib_mlx5_update_db_cq_ci(&iface->super.cq[UCT_IB_DIR_TX]);
 
@@ -344,18 +360,14 @@ static void uct_ib_mlx5dv_dci_qp_init_attr(uct_ib_qp_init_attr_t *qp_attr,
     uct_ib_mlx5dv_dc_qp_init_attr(dv_attr, MLX5DV_DCTYPE_DCI);
 }
 
-static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
-                                                 uint8_t pool_index,
-                                                 uint8_t dci_index,
-                                                 uint8_t path_index,
-                                                 int full_handshake)
+ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
+                                          uint8_t dci_index, int connect)
 {
-    uct_ib_iface_t *ib_iface           = &iface->super.super.super;
-    uct_ib_mlx5_qp_attr_t attr         = {};
+    uct_ib_iface_t *ib_iface   = &iface->super.super.super;
+    uct_ib_mlx5_qp_attr_t attr = {};
+    uct_ib_mlx5_md_t *md = ucs_derived_of(ib_iface->super.md, uct_ib_mlx5_md_t);
+    uct_dc_dci_t *dci    = uct_dc_mlx5_iface_dci(iface, dci_index);
     ucs_status_t status;
-    uct_ib_mlx5_md_t *md               = ucs_derived_of(ib_iface->super.md,
-                                                        uct_ib_mlx5_md_t);
-    uct_dc_dci_t *dci                  = &iface->tx.dcis[dci_index];
 #if HAVE_DC_DV
     uct_ib_device_t *dev               = uct_ib_iface_device(ib_iface);
     struct mlx5dv_qp_init_attr dv_attr = {};
@@ -370,7 +382,8 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
     if (md->flags & UCT_IB_MLX5_MD_FLAG_DEVX_DCI) {
         attr.super.max_inl_cqe[UCT_IB_DIR_RX] = 0;
         attr.uidx                        = htonl(dci_index) >> UCT_IB_UIDX_SHIFT;
-        attr.full_handshake              = full_handshake;
+        attr.full_handshake              = iface->flags &
+                                           UCT_DC_MLX5_IFACE_FLAG_DCI_FULL_HANDSHAKE;
         attr.rdma_wr_disabled            = (iface->flags & UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT) &&
                                            (md->flags & UCT_IB_MLX5_MD_FLAG_NO_RDMA_WR_OPTIMIZED);
         attr.log_num_dci_stream_channels = ucs_ilog2(iface->tx.num_dci_channels);
@@ -434,12 +447,11 @@ init_qp:
         goto err_qp;
     }
 
-    dci->pool_index = pool_index;
-    dci->path_index = path_index;
-
-    status = uct_dc_mlx5_iface_dci_connect(iface, dci);
-    if (status != UCS_OK) {
-        goto err;
+    if (connect) {
+        status = uct_dc_mlx5_iface_dci_connect(iface, dci);
+        if (status != UCS_OK) {
+            goto err;
+        }
     }
 
     if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
@@ -475,19 +487,21 @@ err_put_res_domain:
 }
 
 #if HAVE_DC_DV
-ucs_status_t uct_dc_mlx5_iface_dci_connect(uct_dc_mlx5_iface_t *iface,
-                                           uct_dc_dci_t *dci)
+ucs_status_t
+uct_dc_mlx5_iface_dci_connect(uct_dc_mlx5_iface_t *iface, uct_dc_dci_t *dci)
 {
     uct_ib_mlx5_md_t *md = ucs_derived_of(iface->super.super.super.super.md,
                                           uct_ib_mlx5_md_t);
     uct_ib_device_t *dev = uct_ib_iface_device(&iface->super.super.super);
+    uct_dc_mlx5_dci_config_t *config =
+            &iface->tx.dci_pool[dci->pool_index].config;
     struct ibv_qp_attr attr;
     long attr_mask;
     ucs_status_t status;
 
     if (md->flags & UCT_IB_MLX5_MD_FLAG_DEVX) {
         return uct_dc_mlx5_iface_devx_dci_connect(iface, &dci->txwq.super,
-                                                  dci->path_index);
+                                                  config);
     }
 
     ucs_assert(dci->txwq.super.type == UCT_IB_MLX5_OBJ_TYPE_VERBS);
@@ -533,7 +547,7 @@ ucs_status_t uct_dc_mlx5_iface_dci_connect(uct_dc_mlx5_iface_t *iface,
     attr.timeout        = iface->super.super.config.timeout;
     attr.rnr_retry      = iface->super.super.config.rnr_retry;
     attr.retry_cnt      = iface->super.super.config.retry_cnt;
-    attr.max_rd_atomic  = iface->super.super.config.max_rd_atomic;
+    attr.max_rd_atomic  = config->max_rd_atomic;
     attr_mask           = IBV_QP_STATE      |
                           IBV_QP_SQ_PSN     |
                           IBV_QP_TIMEOUT    |
@@ -733,93 +747,159 @@ void uct_dc_mlx5_cleanup_rx(uct_rc_iface_t *rc_iface)
 static void uct_dc_mlx5_iface_dci_pool_destroy(uct_dc_mlx5_dci_pool_t *dci_pool)
 {
     ucs_arbiter_cleanup(&dci_pool->arbiter);
-    ucs_free(dci_pool->stack);
+    ucs_array_cleanup_dynamic(&dci_pool->stack);
 }
 
-static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface,
-                                           uint8_t num_dci_pools,
-                                           uint8_t num_dcis)
+static void
+uct_dc_mlx5_destroy_dci(uct_dc_mlx5_iface_t *iface, uct_dc_dci_t *dci)
 {
     uct_ib_mlx5_md_t *md = ucs_derived_of(iface->super.super.super.super.md,
                                           uct_ib_mlx5_md_t);
+
+    uct_rc_txqp_cleanup(&iface->super.super, &dci->txqp);
+    uct_ib_mlx5_destroy_qp(md, &dci->txwq.super);
+
+    if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
+        ucs_arbiter_group_cleanup(&dci->arb_group);
+    }
+    uct_ib_mlx5_qp_mmio_cleanup(&dci->txwq.super, dci->txwq.reg);
+
+    dci->txwq.super.qp_num = UCT_IB_INVALID_QPN;
+}
+
+static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface)
+{
+    uint8_t num_dcis = ucs_array_length(&iface->tx.dcis);
+    uct_dc_dci_t *dci;
     uint8_t pool_index, dci_index;
 
-    ucs_assert(num_dci_pools <= iface->tx.num_dci_pools);
-    ucs_assert(num_dcis <= uct_dc_mlx5_iface_total_ndci(iface));
-
     for (dci_index = 0; dci_index < num_dcis; dci_index++) {
-        uct_rc_txqp_cleanup(&iface->super.super,
-                            &iface->tx.dcis[dci_index].txqp);
-        uct_ib_mlx5_destroy_qp(md, &iface->tx.dcis[dci_index].txwq.super);
-
-        if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
-            ucs_arbiter_group_cleanup(&iface->tx.dcis[dci_index].arb_group);
+        dci = uct_dc_mlx5_iface_dci(iface, dci_index);
+        if (!uct_dc_mlx5_is_dci_valid(dci)) {
+            continue;
         }
-        uct_ib_mlx5_qp_mmio_cleanup(&iface->tx.dcis[dci_index].txwq.super,
-                                    iface->tx.dcis[dci_index].txwq.reg);
+
+        uct_dc_mlx5_destroy_dci(iface, dci);
     }
 
-    for (pool_index = 0; pool_index < num_dci_pools; pool_index++) {
+    for (pool_index = 0; pool_index < iface->tx.num_dci_pools; pool_index++) {
         uct_dc_mlx5_iface_dci_pool_destroy(&iface->tx.dci_pool[pool_index]);
     }
 
-    ucs_free(iface->tx.dcis);
+    ucs_array_cleanup_dynamic(&iface->tx.dcis);
+}
+
+static void
+uct_dc_mlx5_dump_dci_pool_config(const uct_dc_mlx5_dci_config_t *config)
+{
+    ucs_debug("dci pool config: (path_index=%u, max_rd_atomic=%u)",
+              config->path_index, config->max_rd_atomic);
+}
+
+static void
+uct_dc_mlx5_iface_create_dci_pool(uct_dc_mlx5_iface_t *iface,
+                                  const uct_dc_mlx5_dci_config_t *config,
+                                  uint8_t *pool_index_p)
+{
+    const uint8_t pool_index = iface->tx.num_dci_pools;
+    uint8_t pool_size        = iface->tx.ndci;
+    uct_dc_mlx5_dci_pool_t *dci_pool;
+
+    ucs_assertv_always(iface->tx.num_dci_pools <
+                               UCT_DC_MLX5_IFACE_MAX_DCI_POOLS,
+                       "num_dci_pools=%d, UCT_DC_MLX5_IFACE_MAX_DCI_POOLS=%d",
+                       iface->tx.num_dci_pools,
+                       UCT_DC_MLX5_IFACE_MAX_DCI_POOLS);
+    ucs_debug("creating dci pool %u with %u QPs", pool_index, pool_size);
+    uct_dc_mlx5_dump_dci_pool_config(config);
+
+    dci_pool                    = &iface->tx.dci_pool[pool_index];
+    dci_pool->stack_top         = 0;
+    dci_pool->release_stack_top = -1;
+    dci_pool->config            = *config;
+    ucs_arbiter_init(&dci_pool->arbiter);
+    ucs_array_init_dynamic(&dci_pool->stack);
+
+    iface->tx.num_dci_pools++;
+    *pool_index_p = pool_index;
+}
+
+uint32_t uct_dc_mlx5_dci_config_hash(const uct_dc_mlx5_dci_config_t *dci_config)
+{
+    return dci_config->path_index | (dci_config->max_rd_atomic << 8);
+}
+
+ucs_status_t
+uct_dc_mlx5_dci_pool_get_or_create(uct_dc_mlx5_iface_t *iface,
+                                   const uct_dc_mlx5_dci_config_t *dci_config,
+                                   uint8_t *pool_index_p)
+{
+    uint32_t config_hash = uct_dc_mlx5_dci_config_hash(dci_config);
+    khiter_t khit;
+    ucs_kh_put_t kh_put_ret;
+
+    khit = kh_put(uct_dc_mlx5_config_hash, &iface->dc_config_hash, config_hash,
+                  &kh_put_ret);
+    if (kh_put_ret == UCS_KH_PUT_KEY_PRESENT) {
+        *pool_index_p = kh_value(&iface->dc_config_hash, khit);
+        return UCS_OK;
+    }
+
+    if (kh_put_ret == UCS_KH_PUT_FAILED) {
+        ucs_error("failed to kh_put a new dci configuration");
+        return UCS_ERR_EXCEEDS_LIMIT;
+    }
+
+    uct_dc_mlx5_iface_create_dci_pool(iface, dci_config, pool_index_p);
+    kh_value(&iface->dc_config_hash, khit) = *pool_index_p;
+
+    return UCS_OK;
+}
+
+ucs_status_t uct_dc_mlx5_iface_resize_and_fill_dcis(uct_dc_mlx5_iface_t *iface,
+                                                    uint16_t size)
+{
+    uct_dc_dci_t empty_dci = {{{0}}};
+
+    empty_dci.txwq.super.qp_num = UCT_IB_INVALID_QPN;
+    ucs_array_resize(&iface->tx.dcis, size, empty_dci,
+                     ucs_error("%p: could not resize dcis array to %u", iface,
+                               size);
+                     return UCS_ERR_NO_MEMORY);
+
+    return UCS_OK;
 }
 
 static ucs_status_t
 uct_dc_mlx5_iface_dcis_create(uct_dc_mlx5_iface_t *iface,
                               const uct_dc_mlx5_iface_config_t *config)
 {
-    const uint8_t num_paths = iface->super.super.super.num_paths;
-    uint8_t dci_index       = 0;
-    uct_dc_mlx5_dci_pool_t *dci_pool;
-    uint8_t pool_index, i;
     ucs_status_t status;
+    uct_dc_dci_t *dci;
 
-    iface->tx.dcis = ucs_calloc(uct_dc_mlx5_iface_total_ndci(iface),
-                                sizeof(*iface->tx.dcis), "dcis");
-    if (iface->tx.dcis == NULL) {
-        status = UCS_ERR_NO_MEMORY;
-        goto err;
+    ucs_array_init_dynamic(&iface->tx.dcis);
+    /* Reserving max size as a temporary solution due to pointer-copy issue in ucs_array
+       TODO: fix when ucs_array_t supports complex types 
+    */
+    status = uct_dc_mlx5_iface_resize_and_fill_dcis(
+            iface, UCT_DC_MLX5_IFACE_MAX_DCI_POOLS * iface->tx.ndci);
+    if (status != UCS_OK) {
+        return status;
     }
 
-    for (pool_index = 0; pool_index < iface->tx.num_dci_pools; pool_index++) {
-        ucs_debug("creating dci pool %u with %u QPs", pool_index, iface->tx.ndci);
+    ucs_array_length(&iface->tx.dcis) = 0;
 
-        dci_pool        = &iface->tx.dci_pool[pool_index];
-        dci_pool->stack = ucs_calloc(iface->tx.ndci, sizeof(*dci_pool->stack),
-                                     "dci pool stack");
-        if (dci_pool->stack == NULL) {
-            status = UCS_ERR_NO_MEMORY;
-            goto err_dcis_destroy;
-        }
-
-        ucs_arbiter_init(&dci_pool->arbiter);
-        dci_pool->stack_top         = 0;
-        dci_pool->release_stack_top = -1;
-
-        for (i = 0; i < iface->tx.ndci; ++i) {
-            status = uct_dc_mlx5_iface_create_dci(
-                    iface, pool_index, dci_index, pool_index % num_paths,
-                    iface->flags & UCT_DC_MLX5_IFACE_FLAG_DCI_FULL_HANDSHAKE);
-            if (status != UCS_OK) {
-                goto err_dci_pool_destroy;
-            }
-
-            dci_pool->stack[i] = dci_index;
-            ++dci_index;
-        }
+    status = uct_dc_mlx5_iface_create_dci(iface, 0, 0);
+    if (status != UCS_OK) {
+        return status;
     }
 
-    iface->tx.bb_max = iface->tx.dcis[0].txwq.bb_max;
+    dci = uct_dc_mlx5_iface_dci(iface, 0);
+
+    iface->tx.bb_max = dci->txwq.bb_max;
+    uct_dc_mlx5_destroy_dci(iface, dci);
+
     return UCS_OK;
-
-err_dci_pool_destroy:
-    uct_dc_mlx5_iface_dci_pool_destroy(dci_pool);
-err_dcis_destroy:
-    uct_dc_mlx5_iface_dcis_destroy(iface, pool_index, dci_index);
-err:
-    return status;
 }
 
 void uct_dc_mlx5_iface_set_quota(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_iface_config_t *config)
@@ -870,7 +950,7 @@ static void uct_dc_mlx5_iface_vfs_refresh(uct_iface_h tl_iface)
         dci_pool = &iface->tx.dci_pool[pool_index];
         ucs_vfs_obj_add_dir(iface, dci_pool, "dci_pool/%d", pool_index);
         for (i = 0; i < iface->tx.ndci; ++i) {
-            dci = &iface->tx.dcis[dci_index];
+            dci = uct_dc_mlx5_iface_dci(iface, dci_index);
             ucs_vfs_obj_add_dir(dci_pool, dci, "%d", dci_index);
             uct_ib_mlx5_txwq_vfs_populate(&dci->txwq, dci);
             uct_rc_txqp_vfs_populate(&dci->txqp, dci);
@@ -958,6 +1038,7 @@ uct_dc_mlx5_iface_get_address(uct_iface_h tl_iface, uct_iface_addr_t *iface_addr
 
 static inline ucs_status_t uct_dc_mlx5_iface_flush_dcis(uct_dc_mlx5_iface_t *iface)
 {
+    size_t num_dcis;
     int i;
 
     if (kh_size(&iface->tx.fc_hash) != 0) {
@@ -966,8 +1047,10 @@ static inline ucs_status_t uct_dc_mlx5_iface_flush_dcis(uct_dc_mlx5_iface_t *ifa
         return UCS_INPROGRESS;
     }
 
-    for (i = 0; i < iface->tx.ndci * iface->tx.num_dci_pools; i++) {
-        if (uct_dc_mlx5_iface_flush_dci(iface, i) != UCS_OK) {
+    num_dcis = ucs_array_length(&iface->tx.dcis);
+    for (i = 0; i < num_dcis; i++) {
+        if (uct_dc_mlx5_is_dci_valid(uct_dc_mlx5_iface_dci(iface, i)) &&
+            uct_dc_mlx5_iface_flush_dci(iface, i) != UCS_OK) {
             return UCS_INPROGRESS;
         }
     }
@@ -1001,6 +1084,8 @@ ucs_status_t uct_dc_mlx5_iface_flush(uct_iface_h tl_iface, unsigned flags, uct_c
 
 ucs_status_t uct_dc_mlx5_iface_init_fc_ep(uct_dc_mlx5_iface_t *iface)
 {
+    uint8_t pool_index = 0;
+    uct_dc_mlx5_dci_config_t dci_config;
     ucs_status_t status;
     uct_dc_mlx5_ep_t *ep;
 
@@ -1010,6 +1095,7 @@ ucs_status_t uct_dc_mlx5_iface_init_fc_ep(uct_dc_mlx5_iface_t *iface)
         status =  UCS_ERR_NO_MEMORY;
         goto err;
     }
+
     /* We do not have any peer address at this point, so init basic subclasses
      * only (for statistics, iface, etc) */
     status = UCS_CLASS_INIT(uct_base_ep_t, (void*)(&ep->super),
@@ -1020,8 +1106,17 @@ ucs_status_t uct_dc_mlx5_iface_init_fc_ep(uct_dc_mlx5_iface_t *iface)
         goto err_free;
     }
 
-    ep->flags = 0;
-    status    = uct_dc_mlx5_ep_basic_init(iface, ep);
+    uct_dc_mlx5_init_dci_config(&dci_config, 0,
+                                iface->super.super.config.max_rd_atomic);
+    status = uct_dc_mlx5_dci_pool_get_or_create(iface, &dci_config,
+                                                &pool_index);
+    if (status != UCS_OK) {
+        goto err_cleanup;
+    }
+
+    ep->flags = pool_index & UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK;
+
+    status = uct_dc_mlx5_ep_basic_init(iface, ep);
     if (status != UCS_OK) {
         ucs_error("FC ep init failed %s", ucs_status_string(status));
         goto err_cleanup;
@@ -1041,6 +1136,7 @@ err:
 static void uct_dc_mlx5_iface_cleanup_fc_ep(uct_dc_mlx5_iface_t *iface)
 {
     uct_dc_mlx5_ep_t *fc_ep = iface->tx.fc_ep;
+    uct_dc_dci_t *fc_dci    = uct_dc_mlx5_iface_dci(iface, fc_ep->dci);
     uct_rc_iface_send_op_t *op;
     ucs_queue_iter_t iter;
     uct_rc_txqp_t *txqp;
@@ -1049,8 +1145,13 @@ static void uct_dc_mlx5_iface_cleanup_fc_ep(uct_dc_mlx5_iface_t *iface)
     ucs_arbiter_group_cleanup(&fc_ep->arb_group);
     uct_rc_fc_cleanup(&fc_ep->fc);
 
+    if ((fc_ep->dci != UCT_DC_MLX5_EP_NO_DCI) &&
+        !uct_dc_mlx5_is_dci_valid(fc_dci)) {
+        goto out;
+    }
+
     if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
-        txqp = &iface->tx.dcis[fc_ep->dci].txqp;
+        txqp = &fc_dci->txqp;
         ucs_queue_for_each_safe(op, iter, &txqp->outstanding, queue) {
             if (op->handler == uct_dc_mlx5_ep_fc_pure_grant_send_completion) {
                 ucs_queue_del_iter(&txqp->outstanding, iter);
@@ -1059,13 +1160,13 @@ static void uct_dc_mlx5_iface_cleanup_fc_ep(uct_dc_mlx5_iface_t *iface)
         }
     } else if (fc_ep->dci != UCT_DC_MLX5_EP_NO_DCI) {
         /* All outstanding operations on this DCI are FC_PURE_GRANT packets */
-        txqp = &iface->tx.dcis[fc_ep->dci].txqp;
+        txqp = &fc_dci->txqp;
         uct_rc_txqp_purge_outstanding(&iface->super.super, txqp,
                                       /* complete with OK to avoid re-sending */
-                                      UCS_OK,
-                                      iface->tx.dcis[fc_ep->dci].txwq.sw_pi, 0);
+                                      UCS_OK, fc_dci->txwq.sw_pi, 0);
     }
 
+out:
     UCS_CLASS_CLEANUP(uct_base_ep_t, fc_ep);
     ucs_free(fc_ep);
 }
@@ -1235,9 +1336,9 @@ static void uct_dc_mlx5_dci_handle_failure(uct_dc_mlx5_iface_t *iface,
     }
 
     if (ep == NULL) {
-        uct_ib_mlx5_completion_with_err(&iface->super.super.super,
-                                        (uct_ib_mlx5_err_cqe_t*)cqe,
-                                        &iface->tx.dcis[dci_index].txwq, level);
+        uct_ib_mlx5_completion_with_err(
+                &iface->super.super.super, (uct_ib_mlx5_err_cqe_t*)cqe,
+                &uct_dc_mlx5_iface_dci(iface, dci_index)->txwq, level);
         return;
     }
 
@@ -1456,7 +1557,6 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
     ucs_status_t status;
     unsigned tx_cq_size;
     unsigned num_dci_channels;
-    int max_dcis;
 
     ucs_trace_func("");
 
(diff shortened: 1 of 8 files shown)


src/uct/ib/dc/dc_mlx5.c
```diff
@@ -773,67 +779,137 @@ static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface,
...
+
+err_pool_destroy:
+    uct_dc_mlx5_iface_dci_pool_destroy(dci_pool);
+    uct_dc_mlx5_iface_dcis_destroy(iface, pool_index, iface->tx.dci_counter);
+
+    return status;
+}
+
+/**
+ * Checks wether dci pool config present in the dc_config_hash and then either returning
```
gleon99, line 845: Checks whether dci pool config is present in dc_config_hash and returns the matching pool index or creates a new one

gleon99: Move docs to `.h`

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -773,67 +779,137 @@ static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface,
...
+        ucs_error("failed to kh_put a new dci configuration");
+        return UCS_ERR_EXCEEDS_LIMIT;
+    }
+
+    pool_size = dci_config->key.is_keepalive ? UCT_DC_MLX5_KEEPALIVE_NUM_DCIS :
+                                               iface->tx.ndci;
+    status    = uct_dc_mlx5_iface_create_dci_pool(iface, dci_config, pool_size,
+                                                  pool_index_p);
+    if (status != UCS_OK) {
+        return status;
```
gleon99, line 877: `kh_del`?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1724,24 +1803,21 @@ uct_dc_mlx5_dci_keepalive_handle_failure(uct_dc_mlx5_iface_t *iface,
...
-    }
+    sl             = iface->super.super.super.config.sl;
+    port_affinity  = iface->tx.port_affinity;
+
+    uct_dc_mlx5_init_dci_config_key(&dci_config, sl, port_affinity, 0, 1);
+    uct_dc_mlx5_dci_pool_get_or_create(iface, &dci_config, &pool_index);
 
     iface->flags        |= UCT_DC_MLX5_IFACE_FLAG_KEEPALIVE;
-    iface->keepalive_dci = dci_index;
+    iface->keepalive_dci = iface->tx.dci_counter - 1;
```
gleon99, line 1820: Can this place be reached with `dci_counter` 0?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -773,67 +779,137 @@ static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface,
...
+        ucs_error("failed to allocate dci_pool stack");
+        return UCS_ERR_NO_MEMORY;
+    }
+
+    ucs_arbiter_init(&dci_pool->arbiter);
+    dci_pool->stack_top         = 0;
+    dci_pool->release_stack_top = -1;
+
+    for (i = 0; i < pool_size; ++i) {
+        status = uct_dc_mlx5_iface_create_dci(
```
gleon99, line 820: We discussed to do "lazy" creation..

gleon99:
@roiedanino maybe add some tests?

src/uct/ib/dc/dc_mlx5.c
roiedanino: Yes, it's in progress I will add WIP-DNM label

src/uct/ib/dc/dc_mlx5.c
roiedanino: No, in the current state of the PR a pool is being created with a single dci so dc_counter is at least 1.

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -247,6 +248,25 @@ typedef struct uct_dc_mlx5_ep_fc_entry {
...
+    UCT_DC_MLX5_EP_TYPE_FC
+} UCS_S_PACKED uct_dc_mlx5_ep_type_t;
+
+typedef union uct_dc_mlx5_dci_config {
+    struct {
+        uint8_t sl;
+        uint8_t port_affinity;
+        uint8_t path_index;
+        uint8_t ep_type;
+        uint8_t max_rd_atomic;
```
roiedanino, line 263: should be a single bit because it can be either 16 or 64, maybe we can unite it with ep_type and port_affinity?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -177,6 +181,14 @@ uct_dc_mlx5_ep_create_connected(const uct_ep_params_t *params, uct_ep_h* ep_p)
     if_addr    = (const uct_dc_mlx5_iface_addr_t *)params->iface_addr;
     path_index = UCT_EP_PARAMS_GET_PATH_INDEX(params);
 
+    if_addr_max_rd_atomic =
+            if_addr->flags & UCT_DC_MLX5_IFACE_ADDR_MAX_RD_ATOMIC_16 ? 16 : 64;
```
yosefe, line 185: add ( )

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -177,6 +181,14 @@ uct_dc_mlx5_ep_create_connected(const uct_ep_params_t *params, uct_ep_h* ep_p)
...
+    max_rd_atomic = ucs_min(iface->super.super.config.max_rd_atomic,
+                            if_addr_max_rd_atomic);
+
+    uct_dc_mlx5_init_dci_config_key(&dci_config, sl, port_affinity, path_index,
+                                    0, max_rd_atomic);
```
yosefe, line 190: maybe pass iface param to uct_dc_mlx5_init_dci_config_key?
to move more "common" logic of initializing the config key inside uct_dc_mlx5_init_dci_config_key

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -771,71 +792,123 @@ static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface,
...
 
+ucs_status_t
+uct_dc_mlx5_iface_create_dci_pool(uct_dc_mlx5_iface_t *iface,
+                                  const uct_dc_mlx5_dci_config_t *config,
+                                  uint8_t pool_size, uint8_t *pool_index_p)
+{
+    const int pool_index = iface->tx.num_dci_pools;
+    uct_dc_mlx5_dci_pool_t *dci_pool;
+
+    ucs_assert(iface->tx.num_dci_pools < UCT_DC_MLX5_IFACE_MAX_DCI_POOLS);
```
yosefe, line 814: assertv

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -771,71 +792,123 @@ static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface,
...
+    }
+
+    it = kh_put(uct_dc_mlx5_config_hash, &iface->dc_config_hash,
+                dci_config->u64, &kh_put_ret);
+    if (kh_put_ret == UCS_KH_PUT_FAILED) {
+        ucs_error("failed to kh_put a new dci configuration");
+        return UCS_ERR_EXCEEDS_LIMIT;
+    }
+
+    pool_size = (dci_config->key.flags & UCT_DC_MLX5_DCI_CONFIG_KEEPALIVE) ?
```
yosefe, line 861: 1. can uct_dc_mlx5_iface_create_dci_pool calc pool_size internally?
2. as a preparation task, we can remove keepalive from DC since it's not useful there anyway @brminich FYI

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -771,71 +792,123 @@ static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface,
...
-            status = UCS_ERR_NO_MEMORY;
-            goto err_dcis_destroy;
-        }
+    iface->tx.dci_counter = 0;
 
-        ucs_arbiter_init(&dci_pool->arbiter);
-        dci_pool->stack_top         = 0;
-        dci_pool->release_stack_top = -1;
+    /* Create a 'default' dci pool to be used when no eps were created */
+    uct_dc_mlx5_init_dci_config_key(&dci_config,
```
yosefe, line 894: why needed?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -974,7 +1048,8 @@ static inline ucs_status_t uct_dc_mlx5_iface_flush_dcis(uct_dc_mlx5_iface_t *ifa
         return UCS_INPROGRESS;
     }
 
-    for (i = 0; i < iface->tx.ndci * iface->tx.num_dci_pools; i++) {
+    num_dcis = uct_dc_mlx5_iface_max_total_dcis(iface);
```
yosefe, line 1051: init inline

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1028,7 +1105,19 @@ ucs_status_t uct_dc_mlx5_iface_init_fc_ep(uct_dc_mlx5_iface_t *iface)
...
+    max_rd_atomic  = iface->super.super.config.max_rd_atomic;
+
+    uct_dc_mlx5_init_dci_config_key(&dci_config, sl, port_affinity, 0, 0,
+                                    max_rd_atomic);
+
+    status = uct_dc_mlx5_dci_pool_get_or_create(iface, &dci_config,
+                                                &pool_index);
+    if (status != UCS_OK) {
+        goto err_cleanup;
+    }
```
yosefe, line 1119: space line after

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1057,6 +1146,11 @@ static void uct_dc_mlx5_iface_cleanup_fc_ep(uct_dc_mlx5_iface_t *iface)
     ucs_arbiter_group_cleanup(&fc_ep->arb_group);
     uct_rc_fc_cleanup(&fc_ep->fc);
 
+    if ((fc_ep->dci != UCT_DC_MLX5_EP_NO_DCI) &&
+        (!iface->tx.dcis[fc_ep->dci].initialized)) {
```
yosefe, line 1150: 1. extra () 
2. why need to check initialized?

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -487,6 +533,9 @@ uct_dc_mlx5_iface_dci_has_outstanding(uct_dc_mlx5_iface_t *iface, int dci_index)
 {
     uct_rc_txqp_t *txqp;
 
+    if (!iface->tx.dcis[dci_index].initialized) {
```
yosefe, line 536: seems there are many places checking "initialized" but i would expect we will not get to them in the first place with an uninitialized dci - since an uninitialized dci should not be allocated by any endpoint

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -471,6 +483,8 @@ static ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
     }
 
     uct_rc_txqp_available_set(&dci->txqp, dci->txwq.bb_max);
+    ++iface->tx.dci_counter;
+    dci->initialized = 1;
```
yosefe, line 487: if we have iface->tx.dci_counter, why need also initialized field? we could check if dci_index<dci_counter

roiedanino: Not in random policy

src/uct/ib/dc/dc_mlx5.h
roiedanino: But in random policy, any dci_index can be allocated and we need to track it somehow, another option would be to allocate all dcis ahead when random policy is used

yosefe: i see. so maybe reuse another field, such as qp number, to check if the dci is initialized?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -577,14 +643,18 @@ uct_dc_mlx5_iface_dci_get(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
+    ucs_status_t status;
 
     ucs_assert(!iface->super.super.config.tx_moderation);
 
     if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
         /* Silence Coverity - in random policy the endpoint always has an
          * assigned DCI */
         ucs_assert(ep->dci != UCT_DC_MLX5_EP_NO_DCI);
-
+        status = uct_dc_mlx5_dci_pool_add_dci(iface, pool_index, ep->dci);
```
yosefe, line 654: in random need to alloc dci in ep create

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -567,6 +570,69 @@ uct_dc_mlx5_iface_dci_detach(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
+{
+    /* take a first available dci from stack.
+     * There is no need to check txqp because
+     * dci must have resources to transmit.
+     */
+    uint8_t pool_index           = uct_dc_mlx5_ep_pool_index(ep);
+    uct_dc_mlx5_dci_pool_t *pool = &iface->tx.dci_pool[pool_index];
+    ucs_status_t status;
+
+    status = uct_dc_mlx5_dci_pool_add_dci(iface, pool_index, iface->tx.dci_counter);
```
yosefe, line 615: instead of adding this on fast path,check pool size in uct_dc_mlx5_iface_dci_can_alloc and in needed extend the pool there

src/uct/ib/dc/dc_mlx5.c
roiedanino: I need to delete sl and port affinity from the config key anyway, the other parameters are not derived from iface

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -733,93 +749,171 @@ void uct_dc_mlx5_cleanup_rx(uct_rc_iface_t *rc_iface)
...
+    uint8_t pool_index;
+    int i;
 
-    iface->tx.dcis = ucs_calloc(uct_dc_mlx5_iface_total_ndci(iface),
-                                sizeof(*iface->tx.dcis), "dcis");
-    if (iface->tx.dcis == NULL) {
-        status = UCS_ERR_NO_MEMORY;
-        goto err;
+    ucs_array_init_dynamic(&iface->tx.dcis);
+    status = ucs_array_reserve(&iface->tx.dcis, max_num_dcis);
```
yosefe, line 888: why need to reserve up front?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -733,93 +749,171 @@ void uct_dc_mlx5_cleanup_rx(uct_rc_iface_t *rc_iface)
...
-                    iface->flags & UCT_DC_MLX5_IFACE_FLAG_DCI_FULL_HANDSHAKE);
-            if (status != UCS_OK) {
-                goto err_dci_pool_destroy;
-            }
+    /* Create a 'default' dci pool to be used when no eps were created */
+    uct_dc_mlx5_init_dci_config_key(&dci_config, 0, 0,
+                                    iface->super.super.config.max_rd_atomic);
+    status = uct_dc_mlx5_dci_pool_get_or_create(iface, &dci_config, &pool_index);
+    if(status != UCS_OK) {
+        return status;
```
yosefe, line 905: why need such default pool? i'd expect all pools to be created on demand

src/uct/ib/dc/dc_mlx5.c
yosefe: don't we need to flush only existing DCIs?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1023,7 +1121,15 @@ ucs_status_t uct_dc_mlx5_iface_init_fc_ep(uct_dc_mlx5_iface_t *iface)
         goto err_free;
     }
 
-    ep->flags = 0;
+    max_rd_atomic = iface->super.super.config.max_rd_atomic;
```
yosefe, line 1124: can we remove max_rd_atomic helper var?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1023,7 +1121,15 @@ ucs_status_t uct_dc_mlx5_iface_init_fc_ep(uct_dc_mlx5_iface_t *iface)
...
+    uct_dc_mlx5_init_dci_config_key(&dci_config, 0, 0, max_rd_atomic);
+    status = uct_dc_mlx5_dci_pool_get_or_create(iface, &dci_config,
+                                                &pool_index);
+    if (status != UCS_OK) {
+        goto err_cleanup;
+    }
+
+    ep->flags = pool_index;
```
yosefe, line 1132: can we remove pool_index helper var?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1542,18 +1655,17 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
...
-        self->tx.num_dci_pools = self->super.super.super.num_paths;
     }
 
     if ((params->field_mask & UCT_IFACE_PARAM_FIELD_FEATURES) &&
         !(params->features & UCT_IFACE_FEATURE_PUT)) {
         self->flags |= UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT;
     }
-
-    max_dcis = ucs_min(INT8_MAX, UINT8_MAX / self->tx.num_dci_pools);
+    max_dcis = ucs_min(INT8_MAX, UINT8_MAX / UCT_DC_MLX5_IFACE_MAX_DCI_POOLS);
```
yosefe, line 1668: revert?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1611,8 +1723,7 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
     return UCS_OK;
 
 err_destroy_fc_ep_and_dcis:
-    uct_dc_mlx5_iface_dcis_destroy(self, self->tx.num_dci_pools,
-                                   uct_dc_mlx5_iface_total_ndci(self));
+    uct_dc_mlx5_iface_dcis_destroy(self, self->tx.num_dci_pools);
```
yosefe, line 1726: maybe remove 2nd arg for uct_dc_mlx5_iface_dcis_destroy? seems it's always self->tx.num_dci_pools so the function can take it from the iface pointer

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -40,8 +41,8 @@ struct ibv_ravh {
 #define UCT_DC_MLX5_IFACE_TXQP_DCI_GET(_iface, _dci, _txqp, _txwq) \
     { \
         ucs_assert(_dci != UCT_DC_MLX5_EP_NO_DCI); \
-        _txqp = &(_iface)->tx.dcis[_dci].txqp; \
-        _txwq = &(_iface)->tx.dcis[_dci].txwq; \
+        _txqp = &ucs_array_elem(&(_iface)->tx.dcis, _dci).txqp; \
```
yosefe, line 44: maybe add a helper function : `uct_dc_mlx5_iface_dci(iface, dci_index)` to return pointer to a given dci by index
seems it can be used in many places

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -240,6 +241,23 @@ typedef struct uct_dc_mlx5_ep_fc_entry {
...
+    UCT_DC_MLX5_DCI_CONFIG_MAX_RD_ATOMIC_IS_64 = UCS_BIT(1)
+};
+
+typedef union uct_dc_mlx5_dci_config {
+    struct {
+        uint8_t path_index;
+        uint8_t flags;
+        uint8_t padding[6];
+    } key;
+    uint64_t u64;
```
yosefe, line 255: IMO defining this as union is redundant since it is slow path code. it's ok even if it will not fit into 64 bit.
we can define helper function to calc 64-bit (or 32-bit) hash key from it instead

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -240,6 +241,23 @@ typedef struct uct_dc_mlx5_ep_fc_entry {
 
 KHASH_MAP_INIT_INT64(uct_dc_mlx5_fc_hash, uct_dc_mlx5_ep_fc_entry_t);
 
+enum uct_dc_mlx5_dci_config_flags {
+    UCT_DC_MLX5_DCI_CONFIG_KEEPALIVE           = UCS_BIT(0),
```
yosefe, line 245: is keep alive still needed?

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -256,20 +274,23 @@ KHASH_MAP_INIT_INT64(uct_dc_mlx5_fc_hash, uct_dc_mlx5_ep_fc_entry_t);
...
-    int8_t        stack_top;         /* dci stack top */
-    uint8_t       *stack;            /* LIFO of indexes of available dcis */
-    ucs_arbiter_t arbiter;           /* queue of requests waiting for DCI */
-    int8_t        release_stack_top; /* releasing dci's stack,
-                                        points to last DCI to release
-                                        or -1 if no DCI's to release */
+    int8_t                   stack_top;         /* dci stack top */
+    uct_dc_mlx5_pool_stack_t stack;             /* LIFO of indexes of available dcis */
+    ucs_arbiter_t            arbiter;           /* queue of requests waiting for DCI */
+    int8_t                   release_stack_top; /* releasing dci's stack,
```
yosefe, line 280: move this after "stack_top" field, to reduce struct sizer.
pls move new DC structs to ucx_info / type_info and check their sizes - pls post them here and let's review they are reasonable

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -321,6 +344,8 @@ struct uct_dc_mlx5_iface {
         uint8_t                   port_affinity;
     } rx;
 
+    khash_t(uct_dc_mlx5_config_hash) dc_config_hash;
```
yosefe, line 347: need to adjust all fields indentation accordingly

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -366,15 +391,45 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
 
 void uct_dc_mlx5_iface_reset_dci(uct_dc_mlx5_iface_t *iface, uint8_t dci_index);
 
+ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
+                                          uint8_t pool_index, uint8_t dci_index,
```
yosefe, line 395: 1. can we pass dci_config instead of all these params?
2. is pool_index still needed?

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -366,15 +391,45 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
...
+                                          uint8_t path_index,
+                                          int full_handshake);
+
+ucs_status_t
+uct_dc_mlx5_iface_create_dci_pool(uct_dc_mlx5_iface_t *iface,
+                                  const uct_dc_mlx5_dci_config_t *config,
+                                  uint8_t *pool_index_p);
```
yosefe, line 402: why needed in H file?

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -366,15 +391,45 @@ void uct_dc_mlx5_iface_set_ep_failed(uct_dc_mlx5_iface_t *iface,
...
+*/
+ucs_status_t
+uct_dc_mlx5_dci_pool_get_or_create(uct_dc_mlx5_iface_t *iface,
+                                   const uct_dc_mlx5_dci_config_t *dci_config,
+                                   uint8_t *pool_index_p);
+
+static UCS_F_ALWAYS_INLINE void uct_dc_mlx5_dci_set_invalid(uct_dc_dci_t *dci)
+{
+    dci->txwq.super.qp_num = UCT_IB_INVALID_QPN;
+}
```
yosefe, line 416: IMO this wrapper is redundant

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -487,10 +547,17 @@ uct_dc_mlx5_iface_flush_dci(uct_dc_mlx5_iface_t *iface, int dci_index)
...
+                   ucs_array_elem(&iface->tx.dcis, dci_index).txqp.available,
+                   iface->tx.bb_max);
+    ucs_assertv(uct_rc_txqp_unsignaled(
+                        &ucs_array_elem(&iface->tx.dcis, dci_index).txqp) == 0,
                 "unsignalled send is not supported!!!");
     return UCS_INPROGRESS;
 }
 
+static UCS_F_ALWAYS_INLINE uint8_t
+uct_dc_mlx5_dci_config_get_max_rd_atomic(const uct_dc_mlx5_dci_config_t *dci_config) {
```
yosefe, line 559: code style

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -408,10 +464,10 @@ uct_dc_mlx5_iface_fill_ravh(struct ibv_ravh *ravh, uint32_t dct_num)
 }
 #endif
 
-static UCS_F_ALWAYS_INLINE uint8_t
-uct_dc_mlx5_iface_total_ndci(uct_dc_mlx5_iface_t *iface)
+static UCS_F_ALWAYS_INLINE size_t
+uct_dc_mlx5_iface_max_total_dcis(uct_dc_mlx5_iface_t *iface)
```
yosefe, line 468: I'd expect this function will not be needed anymore since we keep all dis in ucs_array and we know how much we have

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -472,10 +529,13 @@ uct_dc_mlx5_iface_dci_waitq(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
 static UCS_F_ALWAYS_INLINE int
 uct_dc_mlx5_iface_dci_has_outstanding(uct_dc_mlx5_iface_t *iface, int dci_index)
 {
-    uct_rc_txqp_t *txqp;
+    uct_dc_dci_t *dci = &ucs_array_elem(&iface->tx.dcis, dci_index);
+
+    if (!uct_dc_mlx5_is_dci_valid(dci)) {
```
yosefe, line 534: is it really possible we will call this function with an invalid dci_index?

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1255,6 +1265,7 @@ UCS_CLASS_CLEANUP_FUNC(uct_dc_mlx5_ep_t)
 {
     uct_dc_mlx5_iface_t *iface = ucs_derived_of(self->super.super.iface,
                                                 uct_dc_mlx5_iface_t);
+    uct_dc_dci_t *ep_dci = &ucs_array_elem(&iface->tx.dcis, self->dci);
```
yosefe, line 1268: rename to dci for consistency with other places

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -331,9 +392,22 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
 uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
 {
+    ucs_assert_always(pool_index < iface->tx.num_dci_pools);
     return iface->tx.dci_pool[pool_index].stack_top < iface->tx.ndci;
 }
 
+static UCS_F_ALWAYS_INLINE int uct_dc_mlx5_iface_dci_ensure_pool_size(uct_dc_mlx5_iface_t *iface,
+                                                  uint8_t pool_index)
+{
+    ucs_status_t status = uct_dc_mlx5_dci_pool_add_dci(iface, pool_index,
```
yosefe, line 402: 1. this function name is weird
2. a call to slow path function cannot be done before checking some condition, it's adding fast path overhead

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1273,28 +1284,35 @@ UCS_CLASS_CLEANUP_FUNC(uct_dc_mlx5_ep_t)
...
     /* TODO should be removed by flush */
     uct_rc_txqp_purge_outstanding(&iface->super.super,
-                                  &iface->tx.dcis[self->dci].txqp, UCS_ERR_CANCELED,
-                                  iface->tx.dcis[self->dci].txwq.sw_pi, 1);
-    ucs_assert(ucs_queue_is_empty(&iface->tx.dcis[self->dci].txqp.outstanding));
-    iface->tx.dcis[self->dci].ep = NULL;
+                                  &ep_dci->txqp, UCS_ERR_CANCELED,
+                                  ep_dci->txwq.sw_pi, 1);
+    ucs_assert(ucs_queue_is_empty(&ep_dci->txqp.outstanding));
+    ep_dci->ep = NULL;
```
yosefe, line 1296: maybe we can remove this

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -298,13 +300,69 @@ uct_dc_mlx5_ep_from_dci(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
...
+    uct_dc_mlx5_dci_config_t pool_config = {
+        .u64 = pool->config_key
+    };
+    ucs_status_t status;
+    uint8_t* stack_ptr;
+
+    ucs_assertv(ucs_array_length(&pool->stack) <= iface->tx.ndci,
+                "stack length exceeded ndci");
+
+    if ((ucs_array_capacity(&pool->stack) == ucs_array_length(&pool->stack)) ||
```
yosefe, line 344: this looks weird

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -298,13 +300,69 @@ uct_dc_mlx5_ep_from_dci(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
...
+            pool_config.key.path_index,
+            iface->flags & UCT_DC_MLX5_IFACE_FLAG_DCI_FULL_HANDSHAKE);
+
+    if (status != UCS_OK) {
+        ucs_error("iface %p: failed to create dci %u at pool %u", iface,
+                  dci_index, pool_index);
+        return status;
+    }
+
+    stack_ptr  = ucs_array_append(&pool->stack, return UCS_ERR_NO_MEMORY);
```
yosefe, line 360: stack_ptr var is reundant

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -298,13 +300,69 @@ uct_dc_mlx5_ep_from_dci(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
...
+    *stack_ptr = dci_index;
+
```
yosefe, line 362: remove space line

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -314,11 +372,14 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
         ep->dci_channel_index = 0;
+        uct_dc_mlx5_dci_pool_add_dci(iface, uct_dc_mlx5_ep_pool_index(ep),
+                                     ep->dci);
     } else if (uct_dc_mlx5_iface_is_hw_dcs(iface)) {
         ucs_assertv(iface->tx.ndci == 1, "ndci=%u", iface->tx.ndci);
         ep->dci               = 0;
-        ep->dci_channel_index = iface->tx.dcis[ep->dci].next_channel_index++ %
-                                iface->tx.num_dci_channels;
+        ep->dci_channel_index =
+                ucs_array_elem(&iface->tx.dcis, ep->dci).next_channel_index++ %
```
yosefe, line 381: helper var for dci object

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -331,9 +392,22 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 static UCS_F_ALWAYS_INLINE int
 uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
 {
+    ucs_assert_always(pool_index < iface->tx.num_dci_pools);
```
yosefe, line 395: adds overhead to fast path

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -549,22 +596,49 @@ uct_dc_mlx5_iface_dci_detach(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
+    ep->dci    = UCT_DC_MLX5_EP_NO_DCI;
+    ep->flags &= ~UCT_DC_MLX5_EP_FLAG_TX_WAIT;
+    ucs_array_elem(&iface->tx.dcis, dci_index).ep = NULL;
 
     uct_dc_mlx5_iface_dci_schedule_release(iface, dci_index);
 
     return 1;
 }
 
+static inline ucs_status_t uct_dc_mlx5_iface_dci_alloc(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
```
yosefe, line 608: 1. why needed to move this function?
2. why needed to add return value to it?

src/uct/ib/dc/dc_mlx5_ep.h
roiedanino: This means all dcis qps are already allocated, no?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -481,13 +493,17 @@ ucs_status_t uct_dc_mlx5_iface_dci_connect(uct_dc_mlx5_iface_t *iface,
     uct_ib_mlx5_md_t *md = ucs_derived_of(iface->super.super.super.super.md,
                                           uct_ib_mlx5_md_t);
     uct_ib_device_t *dev = uct_ib_iface_device(&iface->super.super.super);
+    uct_dc_mlx5_dci_pool_t *dci_pool = &iface->tx.dci_pool[dci->pool_index];
+    uct_dc_mlx5_dci_config_t dci_config = {
+        .u64 = dci_pool->config_key
+    };
```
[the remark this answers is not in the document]
roiedanino: and copy the entire struct instead of int? wouldn't that be less efficient?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -733,93 +749,171 @@ void uct_dc_mlx5_cleanup_rx(uct_rc_iface_t *rc_iface)
...
+                                    iface->super.super.config.max_rd_atomic);
+    status = uct_dc_mlx5_dci_pool_get_or_create(iface, &dci_config, &pool_index);
+    if(status != UCS_OK) {
+        return status;
+    }
 
-            dci_pool->stack[i] = dci_index;
-            ++dci_index;
-        }
+    ucs_assert_always(pool_index < UCT_DC_MLX5_IFACE_MAX_DCI_POOLS);
```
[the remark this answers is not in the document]
roiedanino: We need to know what will be the next dci_index somehow

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -751,67 +749,89 @@ void uct_dc_mlx5_cleanup_rx(uct_rc_iface_t *rc_iface)
...
-            ucs_arbiter_group_cleanup(&iface->tx.dcis[dci_index].arb_group);
+            ucs_arbiter_group_cleanup(&dci->arb_group);
         }
-        uct_ib_mlx5_qp_mmio_cleanup(&iface->tx.dcis[dci_index].txwq.super,
-                                    iface->tx.dcis[dci_index].txwq.reg);
-        iface->tx.dcis[dci_index].initialized = 0;
+        uct_ib_mlx5_qp_mmio_cleanup(&dci->txwq.super, dci->txwq.reg);
+
+        uct_dc_mlx5_dci_set_invalid(dci);
+        dci = NULL;
```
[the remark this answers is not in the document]
roiedanino: To avoid valgrind issues

src/uct/ib/dc/dc_mlx5.c
roiedanino: This will cause a division by 0 if we don't create at least 1 dci pool: https://github.com/openucx/ucx/pull/9665#discussion_r1590279124

yosefe: i think the check in line 1669 is not relevant anymore since the number of pools is dynamic

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -431,9 +473,9 @@ uct_dc_mlx5_iface_dci_find(uct_dc_mlx5_iface_t *iface, struct mlx5_cqe64 *cqe)
     }
 
     qp_num = ntohl(cqe->sop_drop_qpn) & UCS_MASK(UCT_IB_QPN_ORDER);
-    ndci   = uct_dc_mlx5_iface_total_ndci(iface);
+    ndci   = ucs_array_capacity(&iface->tx.dcis);
```
roiedanino, line 476: change to `ucs_array_length(...)`

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -298,26 +300,91 @@ uct_dc_mlx5_ep_from_dci(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
...
+        empty_dci.txwq.super.qp_num = UCT_IB_INVALID_QPN;
+        dcis_array_size             = ucs_max(ep->dci + 1,
+                                              ucs_array_length(&iface->tx.dcis));
+        ucs_array_resize(&iface->tx.dcis, dcis_array_size, empty_dci,
+                         ucs_error("%p: failed to create ep %p - could not "
+                                   "resize dcis array to %u",
+                                   iface, ep, ep->dci);
+                         return UCS_ERR_NO_MEMORY);
+        if (!uct_dc_mlx5_is_dci_valid(
+                    &ucs_array_elem(&iface->tx.dcis, ep->dci))) {
```
roiedanino, line 379: use `uct_dc_mlx5_iface_dci(...)`

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -275,83 +267,86 @@ UCS_ARRAY_DECLARE_TYPE(uct_dc_mlx5_pool_stack_t, uint8_t, uint8_t);
...
+    ucs_arbiter_t            arbiter;           /* queue of requests waiting for DCI */
+    uct_dc_mlx5_dci_config_t config;
 } uct_dc_mlx5_dci_pool_t;
 
 
 UCS_ARRAY_DECLARE_TYPE(uct_dc_dci_array_t, uint16_t, uct_dc_dci_t);
 
 struct uct_dc_mlx5_iface {
-    uct_rc_mlx5_iface_common_t    super;
+    uct_rc_mlx5_iface_common_t super;
```
yosefe, line 282: pls adjust indent

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -275,83 +267,86 @@ UCS_ARRAY_DECLARE_TYPE(uct_dc_mlx5_pool_stack_t, uint8_t, uint8_t);
...
-        uint8_t                   dci_counter;
+        uint8_t                      num_dci_channels;
     } tx;
 
     struct {
-        uct_ib_mlx5_qp_t          dct;
+        uct_ib_mlx5_qp_t dct;
 
-        uint8_t                   port_affinity;
+        uint8_t          port_affinity;
```
yosefe, line 339: pls adjust indent

src/uct/ib/dc/dc_mlx5.h
yosefe: WDYT?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -300,86 +300,92 @@ uct_dc_mlx5_ep_from_dci(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
...
-            pool_config.key.path_index,
-            iface->flags & UCT_DC_MLX5_IFACE_FLAG_DCI_FULL_HANDSHAKE);
-
+    status = uct_dc_mlx5_iface_create_dci(iface, dci_index, dci);
     if (status != UCS_OK) {
         ucs_error("iface %p: failed to create dci %u at pool %u", iface,
                   dci_index, pool_index);
         return status;
     }
+    dci->path_index = pool->config.path_index;
```
yosefe, line 335: space line before

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -300,86 +300,92 @@ uct_dc_mlx5_ep_from_dci(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
...
+    dci->pool_index = pool_index;
 
-    stack_ptr  = ucs_array_append(&pool->stack, return UCS_ERR_NO_MEMORY);
-    *stack_ptr = dci_index;
+    if(!uct_dc_mlx5_iface_is_dci_rand(iface)) {
```
yosefe, line 338: space after if
check clang-format..

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -389,23 +395,47 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
                           UCS_STATS_ARG(ep->super.stats));
 }
 
-static UCS_F_ALWAYS_INLINE int
+typedef enum {
+    UCT_DC_MLX5_CAN_NOT_ALLOC, /* can't use existing or create new dcis */
+    UCT_DC_MLX5_CAN_ALLOC, /* can use existing dcis for allocation */
+    UCT_DC_MLX5_SHOULD_ALLOC /* can allocate another dci by creating another dci */
+} uct_dc_mlx5_can_alloc_res_t;
```
yosefe, line 402: seems over complication to introduce another set of return codes

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -300,86 +300,92 @@ uct_dc_mlx5_ep_from_dci(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
...
     return UCS_OK;
 }
 
+ucs_status_t static UCS_F_ALWAYS_INLINE
+uct_dc_mlx5_dci_pool_append_dci(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
+{
+    uct_dc_dci_t dci  = {{{0}}};
+    uint8_t dci_index = ucs_array_length(&iface->tx.dcis);
+
+    dci.txwq.super.qp_num = UCT_IB_INVALID_QPN;
```
yosefe, line 351: seems like the logic of adding empty dci appears 3 times in the code, can we unite?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -598,47 +653,21 @@ uct_dc_mlx5_iface_dci_detach(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 
     ep->dci    = UCT_DC_MLX5_EP_NO_DCI;
     ep->flags &= ~UCT_DC_MLX5_EP_FLAG_TX_WAIT;
-    ucs_array_elem(&iface->tx.dcis, dci_index).ep = NULL;
+    // uct_dc_mlx5_iface_dci(iface, ep->dci)->ep = NULL;
```
yosefe, line 656: remove?

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1090,6 +1090,8 @@ ucs_status_t uct_dc_mlx5_ep_fc_pure_grant_send(uct_dc_mlx5_ep_t *ep,
         return UCS_ERR_NO_RESOURCE;
     }
 
+    ucs_assert(ep->dci < ucs_array_length(&iface->tx.dcis));
```
yosefe, line 1093: assertv

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -878,41 +864,26 @@ static ucs_status_t
...
-    status = uct_dc_mlx5_dci_pool_get_or_create(iface, &dci_config, &pool_index);
-    if(status != UCS_OK) {
-        return status;
-    }
-
-    ucs_assert_always(pool_index < UCT_DC_MLX5_IFACE_MAX_DCI_POOLS);
+    ucs_array_resize(&iface->tx.dcis,
+                     UCT_DC_MLX5_IFACE_MAX_DCI_POOLS * iface->tx.ndci, dci,
+                     return UCS_ERR_NO_MEMORY);
+    ucs_array_length(&iface->tx.dcis) = 0;
```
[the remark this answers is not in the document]
roiedanino: To avoid copy when increasing array size while being able to track the actual length of `dcis` using `ucs_array_length`

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -493,17 +490,22 @@ ucs_status_t uct_dc_mlx5_iface_dci_connect(uct_dc_mlx5_iface_t *iface,
...
-    uct_dc_mlx5_dci_config_t dci_config = {
-        .u64 = dci_pool->config_key
-    };
+    uct_dc_mlx5_dci_config_t config;
     struct ibv_qp_attr attr;
     long attr_mask;
     ucs_status_t status;
 
+    /* in case dci is not attached to a dci pool*/
+    if (dci->pool_index != UCT_DC_MLX5_IFACE_INVALID_POOL_INDEX) {
```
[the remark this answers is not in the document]
roiedanino: In `uct_dc_mlx5_iface_dcis_create` just for getting `bb_max`

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1616,13 +1590,13 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
...
-    self->tx.num_dci_pools                 = 1;
-    self->flags                            = 0;
-    self->tx.av_fl_mlid                    = self->super.super.super.path_bits[0] & 0x7f;
+    self->tx.fc_seq                     = 0;
+    self->tx.fc_hard_req_timeout        = config->fc_hard_req_timeout;
+    self->tx.fc_hard_req_resend_time    = ucs_get_time();
+    self->tx.fc_hard_req_progress_cb_id = UCS_CALLBACKQ_ID_NULL;
+    self->tx.num_dci_pools              = 0;
+    self->flags                         = 0;
+    self->tx.av_fl_mlid = self->super.super.super.path_bits[0] & 0x7f;
```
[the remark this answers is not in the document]
roiedanino: clang-format won't allow because of line length

src/uct/ib/dc/dc_mlx5_ep.h
roiedanino: I need somehow to distinguish between those cases otherwise we will need to check the same condition twice:

```c
static UCS_F_ALWAYS_INLINE uct_dc_mlx5_can_alloc_res_t
uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
{
    uct_dc_mlx5_dci_pool_t *pool = &iface->tx.dci_pool[pool_index];

    if (ucs_likely(pool->stack_top < ucs_array_length(&pool->stack))) {
        return UCT_DC_MLX5_CAN_ALLOC;
    }

    if ((ucs_array_length(&pool->stack) == iface->tx.ndci) ||
        (pool->stack_top == iface->tx.ndci)) {
        return UCT_DC_MLX5_CAN_NOT_ALLOC;
    }

    return UCT_DC_MLX5_SHOULD_ALLOC;
}

static UCS_F_ALWAYS_INLINE uct_dc_mlx5_can_alloc_res_t
uct_dc_mlx5_iface_dci_can_alloc_or_create(uct_dc_mlx5_iface_t *iface,
                                          uint8_t pool_index)
{
    uct_dc_mlx5_can_alloc_res_t can_alloc =
            uct_dc_mlx5_iface_dci_can_alloc(iface, pool_index);
    ucs_status_t status;
    /** Here we need to check whether to append or not, a boolean response would make me check the same condition 
         again */
    if (can_alloc != UCT_DC_MLX5_SHOULD_ALLOC) {
        return can_alloc;
    }

    status = uct_dc_mlx5_dci_pool_append_dci(iface, pool_index);
    if (status != UCS_OK) {
        return UCT_DC_MLX5_CAN_NOT_ALLOC;
    }

    return UCT_DC_MLX5_CAN_ALLOC;
}

```

Alternative:
```c
uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
{
    uct_dc_mlx5_dci_pool_t *pool = &iface->tx.dci_pool[pool_index];

    if (ucs_likely(pool->stack_top < ucs_array_length(&pool->stack))) {
        return 1;
    }

    if ((ucs_array_length(&pool->stack) == iface->tx.ndci) ||
        (pool->stack_top == iface->tx.ndci)) {
        return 0;
    }
   return 1;
}

static UCS_F_ALWAYS_INLINE uct_dc_mlx5_can_alloc_res_t
uct_dc_mlx5_iface_dci_can_alloc_or_create(uct_dc_mlx5_iface_t *iface,
                                          uint8_t pool_index)
{
   uct_dc_mlx5_dci_pool_t *pool = &iface->tx.dci_pool[pool_index];
   int can_alloc = uct_dc_mlx5_iface_dci_can_alloc(iface, pool_index);
    ucs_status_t status;
   
    if (can_alloc && pool->stack_top < ucs_array_length(&pool->stack)) {
        return 1;
    }

    status = uct_dc_mlx5_dci_pool_append_dci(iface, pool_index);
    if (status != UCS_OK) {
        return 0;
    }

    return 1;
}

```

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -733,93 +752,138 @@ void uct_dc_mlx5_cleanup_rx(uct_rc_iface_t *rc_iface)
...
+{
+    ucs_status_t status;
+
+    UCT_DC_MLX5_DECL_EMPTY_DCI(empty_dci);
+
+    ucs_array_init_dynamic(&iface->tx.dcis);
+    ucs_array_resize(&iface->tx.dcis,
+                     UCT_DC_MLX5_IFACE_MAX_DCI_POOLS * iface->tx.ndci,
+                     empty_dci, return UCS_ERR_NO_MEMORY);
+    ucs_array_length(&iface->tx.dcis) = 0;
```
roiedanino, line 876: Add `TODO:` comment as it will be fixed in the future

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -490,22 +490,23 @@ ucs_status_t uct_dc_mlx5_iface_dci_connect(uct_dc_mlx5_iface_t *iface,
...
-    uct_dc_mlx5_dci_config_t config;
+    uct_dc_mlx5_dci_config_t *config, temp_config;
     struct ibv_qp_attr attr;
     long attr_mask;
     ucs_status_t status;
 
     /* in case dci is not attached to a dci pool*/
     if (dci->pool_index != UCT_DC_MLX5_IFACE_INVALID_POOL_INDEX) {
-        config = iface->tx.dci_pool[dci->pool_index].config;
+        config = &iface->tx.dci_pool[dci->pool_index].config;
```
yosefe, line 500: can we avoid calling this function (uct_dc_mlx5_iface_dci_connect) if pool_index is invalid?
maybe can also remove UCT_DC_MLX5_IFACE_INVALID_POOL_INDEX define

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -62,6 +62,11 @@ struct ibv_ravh {
         (_self)->flags |= UCT_DC_MLX5_IFACE_FLAG_##_flag_name##_FULL_HANDSHAKE; \
     }
 
+#define UCT_DC_MLX5_DECL_EMPTY_DCI(_empty_dci) \
+    uct_dc_dci_t _empty_dci      = {{{0}}}; \
```
yosefe, line 66: this macro has both var declaration and assign statements so it must always be placed after all var declarations
do we really have to set all its fields to 0?
maybe define a function uct_dc_mlx5_init_empty_dci that sets pool_index and qp_num only?

src/uct/ib/dc/dc_mlx5.h
```diff
@@ -517,9 +522,9 @@ uct_dc_mlx5_iface_dci_has_outstanding(uct_dc_mlx5_iface_t *iface, int dci_index)
 {
     uct_dc_dci_t *dci = uct_dc_mlx5_iface_dci(iface, dci_index);
 
-    if (!uct_dc_mlx5_is_dci_valid(dci)) {
-        return 0;
-    }
+    // if (!uct_dc_mlx5_is_dci_valid(dci)) {
+    //     return 0;
+    // }
```
yosefe, line 527: remove?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -298,26 +300,91 @@ uct_dc_mlx5_ep_from_dci(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
...
+                  dci_index, pool_index);
+        return status;
+    }
+
+    dci->path_index = pool->config.path_index;
+    dci->pool_index = pool_index;
+
+    if (!uct_dc_mlx5_iface_is_dci_rand(iface)) {
+        *ucs_array_append(&pool->stack, return UCS_ERR_NO_MEMORY) = dci_index;
+    }
```
yosefe, line 341: IMO better move it outside this function, to uct_dc_mlx5_dci_pool_append_dci
and:
- rename uct_dc_mlx5_dci_pool_add_dci to uct_dc_mlx5_dci_pool_init_dci
- rename unite uct_dc_mlx5_dci_pool_append_dci into uct_dc_mlx5_iface_dci_can_alloc_or_create since it's called only from there

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -328,10 +395,42 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
+    if (ucs_likely(pool->stack_top < ucs_array_length(&pool->stack))) {
+        return 1;
+    }
+
+    if ((ucs_array_length(&pool->stack) == iface->tx.ndci) ||
+        (pool->stack_top == iface->tx.ndci)) {
+        return 0;
+    }
+
+    return 1;
```
yosefe, line 412: return (ucs_likely(pool->stack_top < ucs_array_length(&pool->stack))) || 
((ucs_array_length(&pool->stack) < iface->tx.ndci) && (pool->stack_top != iface->tx.ndci)
BTW, why needed to check pool->stack_top? it wasn't checked in previous code

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -328,10 +395,42 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
+static UCS_F_ALWAYS_INLINE int
+uct_dc_mlx5_iface_dci_can_alloc_or_create(uct_dc_mlx5_iface_t *iface,
+                                          uint8_t pool_index)
+{
+    int can_alloc = uct_dc_mlx5_iface_dci_can_alloc(iface, pool_index);
+    uct_dc_mlx5_dci_pool_t *pool = &iface->tx.dci_pool[pool_index];
+    ucs_status_t status;
+
+    if (!can_alloc ||
+        ucs_likely(pool->stack_top < ucs_array_length(&pool->stack))) {
```
yosefe, line 424: we already check stack_top in uct_dc_mlx5_iface_dci_can_alloc, why need again?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -328,10 +395,42 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
+
+static UCS_F_ALWAYS_INLINE int
+uct_dc_mlx5_iface_dci_can_alloc_or_create(uct_dc_mlx5_iface_t *iface,
+                                          uint8_t pool_index)
+{
+    int can_alloc = uct_dc_mlx5_iface_dci_can_alloc(iface, pool_index);
+    uct_dc_mlx5_dci_pool_t *pool = &iface->tx.dci_pool[pool_index];
+    ucs_status_t status;
+
+    if (!can_alloc ||
```
yosefe, line 423: can_alloc==1 is the fast path so need to return immediately if it's 1

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -370,8 +469,8 @@ uct_dc_mlx5_iface_progress_pending(uct_dc_mlx5_iface_t *iface,
          * NOTE: in case of rand dci allocation policy, dci_waitq is always
          * empty.
          */
-        if (uct_dc_mlx5_iface_dci_can_alloc(iface, pool_index) &&
-            !uct_dc_mlx5_iface_is_dci_shared(iface)) {
+        if (!uct_dc_mlx5_iface_is_dci_shared(iface) &&
```
yosefe, line 472: why switched order?

src/uct/ib/dc/dc_mlx5_ep.h
roiedanino: We need to distinguish between 3 cases:

1. We can alloc because there are free slots nothing is need to be done
2. We can alloc by appending to the stack / dcis array
3. We can't alloc at all

src/uct/ib/dc/dc_mlx5_ep.h
roiedanino: **can remove 408 and unify conditions

roiedanino: can_alloc -> revert to only check stack_top
can_alloc_or_create -> check stack_top and the rest of the cases

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -861,30 +856,47 @@ uct_dc_mlx5_dci_pool_get_or_create(uct_dc_mlx5_iface_t *iface,
...
        TODO: fix when ucs_array_t supports complex types 
     */
-    ucs_array_resize(&iface->tx.dcis,
-                     UCT_DC_MLX5_IFACE_MAX_DCI_POOLS * iface->tx.ndci,
-                     empty_dci, return UCS_ERR_NO_MEMORY);
+    status = uct_dc_mlx5_iface_resize_and_fill_dcis(
+            iface, UCT_DC_MLX5_IFACE_MAX_DCI_POOLS * iface->tx.ndci);
+    if (status != UCS_OK) {
+        return status;
+    }
```
yosefe, line 888: space line after

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -400,36 +379,41 @@ uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
...
         return 0;
     }
 
+    status = uct_dc_mlx5_dci_pool_init_dci(iface, pool_index, dci_index);
+    if (status != UCS_OK) {
+        return 0;
+    }
+
+    *ucs_array_append(&pool->stack, return UCS_ERR_NO_MEMORY) = dci_index;
+
```
yosefe, line 416: remove space line

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -599,6 +583,9 @@ uct_dc_mlx5_iface_dci_alloc(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 
     ucs_assert(!uct_dc_mlx5_iface_is_dci_shared(iface));
     ucs_assert(pool->release_stack_top < pool->stack_top);
+    ucs_assertv(pool->stack_top < ucs_array_length(&pool->stack),
+                "stack top: %u, length: %u", pool->stack_top,
```
yosefe, line 587: "stack_top=%u, array_length(stack)=%u"

yosefe (approved):

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -267,8 +282,10 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface, int poll_flags)
     ucs_memory_cpu_load_fence();
 
     dci_index = uct_dc_mlx5_iface_dci_find(iface, cqe);
-    txqp      = &iface->tx.dcis[dci_index].txqp;
-    txwq      = &iface->tx.dcis[dci_index].txwq;
+    dci       = uct_dc_mlx5_iface_dci(iface, dci_index);
+    ucs_assert_always(uct_dc_mlx5_is_dci_valid(dci));
```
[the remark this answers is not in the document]
roiedanino: https://github.com/openucx/ucx/pull/9998

PR #9674, opened 2024-02-08
UCT/IB/BASE: Fix error message for QP create

dmitrygx:
## What

This PR fixes minor issue for printed error trace when `ibv_create_qp()` function failed
 in UCT/IB base code.

## Why ?

When debugging some issue noticed that string representation of `errno` is printed twice:
```
[1707374670.430287] [***:1409 :0]        ib_iface.c:1031 UCX  ERROR   iface=0x55555564cfa0: failed to create RC QP TX wr:256 sge:5 inl:64 resp:64 RX wr:0 sge:1 resp:64: File exists failed: File exists
```

## How ?

Remove extra `%m` passed to `uct_ib_check_memlock_limit_msg()` macro, because it prints an error string taken as `%s` from `strerror()`.


tvegas1 (approved):

brminich (approved):

shamisp:
we have the cla

yosefe (approved):

PR #9681, opened 2024-02-09
DOCS: Clarify `ucp_put_nbx` and `ucp_get_nbx`

pentschev:
## What
Clarify some of the aspects of `ucp_put_nbx` and `ucp_get_nbx`.

## Why ?
Guarantees are currently not well-documented.

### DIFF
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -3655,10 +3655,10 @@ ucs_status_ptr_t ucp_tag_msg_recv_nbx(ucp_worker_h worker, void *buffer,
  * are a bit-wise OR of the @ref ucp_op_attr_t enumeration.
  *
  * @note The completion of a put operation signals the local @e buffer can be
- * reused. The completion of the operation on the remote endpoint requires use
+ * reused. The completion of the operation on the remote address requires use
  * of @ref ucp_worker_flush_nbx "ucp_worker_flush_nbx()" or
- * @ref ucp_ep_flush_nbx "ucp_ep_flush_nbx()", after which the data in
- * @e remote_addr is guaranteed to be available.
+ * @ref ucp_ep_flush_nbx "ucp_ep_flush_nbx()", after completion of which the
+ * data in @e remote_addr is guaranteed to be available.
  *
  * @param [in]  ep           Remote endpoint handle.
  * @param [in]  buffer       Pointer to the local source address.
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -3638,24 +3638,27 @@ ucs_status_ptr_t ucp_tag_msg_recv_nbx(ucp_worker_h worker, void *buffer,
  *
  * This routine initiates a storage of contiguous block of data that is
  * described by the local address @a buffer in the remote contiguous memory
- * region described by @a remote_addr address and the @ref ucp_rkey_h "memory
- * handle" @a rkey.  The routine returns immediately and @b does @b not
- * guarantee re-usability of the source address @e buffer. If the operation is
- * completed immediately the routine return UCS_OK, otherwise UCS_INPROGRESS
- * or an error is returned to user. If the put operation completes immediately,
- * the routine returns UCS_OK and the call-back routine @a param.cb.send is
- * @b not invoked. If the operation is @b not completed immediately and no
- * error is reported, then the UCP library will schedule invocation of the
- * call-back routine @a param.cb.send upon completion of the put operation.
- * In other words, the completion of a put operation can be signaled by the
- * return code or execution of the call-back.
+ * region described by @a remote_addr address and the
+ * @ref ucp_rkey_h "memory handle" rkey.  The routine returns immediately and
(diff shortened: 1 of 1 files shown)


pentschev:
cc @brminich

src/ucp/api/ucp.h
```diff
@@ -3638,24 +3638,27 @@ ucs_status_ptr_t ucp_tag_msg_recv_nbx(ucp_worker_h worker, void *buffer,
...
+ * signaled by the return code or execution of the call-back.
  * Immediate completion signals can be fine-tuned via the
  * @ref ucp_request_param_t.op_attr_mask field in the
  * @ref ucp_request_param_t structure. The values of this field
  * are a bit-wise OR of the @ref ucp_op_attr_t enumeration.
  *
- * @note A user can use @ref ucp_worker_flush_nb "ucp_worker_flush_nb()"
- * in order to guarantee re-usability of the source address @e buffer.
+ * @note The completion of a put operation signals the local @e buffer can be
+ * reused. The completion of the operation on the remote endpoint requires use
```
brminich, line 3658: i'd not mention remote endpoint, because we may not even have one

src/ucp/api/ucp.h
```diff
@@ -3707,6 +3710,12 @@ ucs_status_ptr_t ucp_put_nbx(ucp_ep_h ep, const void *buffer, size_t count,
  *
  * @note A user can use @ref ucp_worker_flush_nb "ucp_worker_flush_nb()"
  * in order to guarantee re-usability of the source address @e buffer.
+ * @note The completion of a get operation signals the local @e buffer holds the
+ * the expected data and that both local @e buffer and remote @remote_addr are
```
brminich, line 3714: i'd not mention any guarantee about remote buffer

brminich (commented):
@yosefe, @shamisp

src/ucp/api/ucp.h
pentschev: You're right, sorry, I mixed the fact that we have a _local_ endpoint. Would replacing "remote endpoint" with "remote worker" be correct?

src/ucp/api/ucp.h
pentschev: This is something that isn't clear to me. As discussed offline, if the get operation completes successfully and the local buffer is now ready for use, to me that intuitively would mean "the remote buffer use is also completed". Is that not right? What else could make it unsafe to reuse the remote buffer once the get operation is confirmed successful in the local worker (who initiated and confirmed the operation complete)?

src/ucp/api/ucp.h
brminich: it's actually just remote address

src/ucp/api/ucp.h
brminich: it sounds a bit odd to mention it, as this process does not own/manage that memory

src/ucp/api/ucp.h
```diff
@@ -3638,24 +3638,27 @@ ucs_status_ptr_t ucp_tag_msg_recv_nbx(ucp_worker_h worker, void *buffer,
...
  * @ref ucp_request_param_t.op_attr_mask field in the
  * @ref ucp_request_param_t structure. The values of this field
  * are a bit-wise OR of the @ref ucp_op_attr_t enumeration.
  *
- * @note A user can use @ref ucp_worker_flush_nb "ucp_worker_flush_nb()"
- * in order to guarantee re-usability of the source address @e buffer.
+ * @note The completion of a put operation signals the local @e buffer can be
+ * reused. The completion of the operation on the remote endpoint requires use
+ * of @ref ucp_worker_flush_nbx "ucp_worker_flush_nbx()" or
+ * @ref ucp_ep_flush_nbx "ucp_ep_flush_nbx()", after which the data in
```
SeyedMir, line 3660: "after which" may be a bit misleading. Might be better to say "after completion of which" or something like that.

src/ucp/api/ucp.h
pentschev: Fixed in https://github.com/openucx/ucx/pull/9681/commits/e6634c54776e6a076ed16a9ef218965b24ec84f1

src/ucp/api/ucp.h
pentschev: Exactly, plus it's not clear even from a testing standpoint, what are the guarantees that are provided by UCP? If I just wait on the `ucp_get_nbx`, am I safe to destroy everything immediately? If not, I think it's worth documenting exactly what is required to ensure what the user is doing is enough/safe.

src/ucp/api/ucp.h
pentschev: I see your point. My idea was essentially "after which" actually means "you have to read the docs for those docs and see what they say it's safe", but I concur and made the change in https://github.com/openucx/ucx/pull/9681/commits/e6634c54776e6a076ed16a9ef218965b24ec84f1.

yosefe (approved):

yosefe:
@shamisp can you pls review as well?

brminich (approved):

shamisp:
I like it +1

PR #9684, opened 2024-02-13
UCT/IB/DC: bugfix - dci_pool_index have only 3 bits in ep->flags (it's max value is 15)

roiedanino:
## What
Adding dci_pool_index attribute to `uct_dc_mlx5_ep_t` instead of using the first 3 bits of `ep->flags`.

## Why ?
`UCT_DC_MLX5_IFACE_MAX_DCI_POOLS` is defined to be 16, meaning 3 bits aren't enough, this patch fixes the issue by adding a separate field for dci_pool_index.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.c
+++ b/src/uct/ib/dc/dc_mlx5.c
@@ -1580,6 +1580,8 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
     UCT_DC_MLX5_CHECK_FORCE_FULL_HANDSHAKE(self, config, dct, DCT, status, err);
 
     ucs_assert(self->tx.num_dci_pools <= UCT_DC_MLX5_IFACE_MAX_DCI_POOLS);
+    UCS_STATIC_ASSERT((UCT_DC_MLX5_IFACE_MAX_DCI_POOLS - 1) <=
+                      UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK);
 
     /* create DC target */
     status = uct_dc_mlx5_iface_create_dct(self, config);
--- a/src/uct/ib/dc/dc_mlx5_ep.h
+++ b/src/uct/ib/dc/dc_mlx5_ep.h
@@ -27,35 +27,35 @@
 
 enum uct_dc_mlx5_ep_flags {
     /* DCI pool EP assigned to according to it's lag port */
-    UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK     = UCS_MASK(3),
+    UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK     = UCS_MASK(5),
 
     /* EP is in the tx_wait state. See description of the dcs+quota dci
        selection policy above */
-    UCT_DC_MLX5_EP_FLAG_TX_WAIT             = UCS_BIT(3),
+    UCT_DC_MLX5_EP_FLAG_TX_WAIT             = UCS_BIT(5),
 
     /* EP has GRH address. Used by dc_mlx5 endpoint */
-    UCT_DC_MLX5_EP_FLAG_GRH                 = UCS_BIT(4),
+    UCT_DC_MLX5_EP_FLAG_GRH                 = UCS_BIT(6),
 
     /* Keepalive Request scheduled: indicates that keepalive request
      * is scheduled in outstanding queue and no more keepalive actions
      * are needed */
-    UCT_DC_MLX5_EP_FLAG_KEEPALIVE_POSTED    = UCS_BIT(5),
+    UCT_DC_MLX5_EP_FLAG_KEEPALIVE_POSTED    = UCS_BIT(7),
 
     /* Flush cancel was executed on EP */
-    UCT_DC_MLX5_EP_FLAG_FLUSH_CANCEL        = UCS_BIT(6),
+    UCT_DC_MLX5_EP_FLAG_FLUSH_CANCEL        = UCS_BIT(8),
 
     /* Error handler already called or flush(CANCEL) disabled it */
-    UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED = UCS_BIT(7),
+    UCT_DC_MLX5_EP_FLAG_ERR_HANDLER_INVOKED = UCS_BIT(9),
 
     /* EP supports flush remote operation */
-    UCT_DC_MLX5_EP_FLAG_FLUSH_RKEY          = UCS_BIT(8),
+    UCT_DC_MLX5_EP_FLAG_FLUSH_RKEY          = UCS_BIT(10),
 
     /* Flush remote operation should be invoked */
-    UCT_DC_MLX5_EP_FLAG_FLUSH_REMOTE        = UCS_BIT(9),
+    UCT_DC_MLX5_EP_FLAG_FLUSH_REMOTE        = UCS_BIT(11),
 
 #if UCS_ENABLE_ASSERT
     /* EP was invalidated without DCI */
-    UCT_DC_MLX5_EP_FLAG_INVALIDATED         = UCS_BIT(10)
+    UCT_DC_MLX5_EP_FLAG_INVALIDATED         = UCS_BIT(12)
 #else
     UCT_DC_MLX5_EP_FLAG_INVALIDATED         = 0
 #endif
--- a/test/gtest/uct/ib/test_dc.cc
+++ b/test/gtest/uct/ib/test_dc.cc
@@ -440,6 +440,32 @@ UCS_TEST_P(test_dc, rand_dci_pending_purge) {
     flush();
 }
 
+UCS_TEST_P(test_dc, pool_indices)
+{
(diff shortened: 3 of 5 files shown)


test/gtest/uct/ib/test_dc.cc
```diff
@@ -440,6 +440,32 @@ UCS_TEST_P(test_dc, rand_dci_pending_purge) {
...
+UCS_TEST_P(test_dc, pool_indices)
+{
+    constexpr uint8_t max_dci_pools = UCT_DC_MLX5_IFACE_MAX_DCI_POOLS;
+    entity *rand_e;
+    uint8_t num_dci_pools;
+    uct_dc_mlx5_ep_t *ep;
+    uint8_t expected_pool_index;
+    int i;
+
+    /* set num paths to max */
```
roiedanino, line 452: capital letter

test/gtest/uct/ib/test_dc.cc
```diff
@@ -440,6 +440,32 @@ UCS_TEST_P(test_dc, rand_dci_pending_purge) {
...
+    modify_config("IB_NUM_PATHS", std::to_string(max_dci_pools).c_str());
+
+    rand_e        = create_rand_entity();
+    num_dci_pools = dc_iface(rand_e)->tx.num_dci_pools;
+
+    EXPECT_EQ(num_dci_pools, max_dci_pools);
+
+    /* create a single ep for each pool */
```
roiedanino, line 460: capital letter

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1259,7 +1259,8 @@ UCS_CLASS_INIT_FUNC(uct_dc_mlx5_ep_t, uct_dc_mlx5_iface_t *iface,
 
     self->av.dqp_dct      = av->dqp_dct | htonl(remote_dctn);
     self->av.rlid         = av->rlid;
-    self->flags           = path_index % iface->tx.num_dci_pools;
+    self->dci_pool_index  = path_index % iface->tx.num_dci_pools;
+    self->flags           = self->dci_pool_index;
```
roiedanino, line 1263: set to 0

Artemy-Mellanox (commented):
shouldn't we remove UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK now?

gleon99:
@roiedanino  given that we already mess with the flags enum there, maybe expand the dci pool index to 4 bits instead of adding a separate field?

roiedanino:
> @roiedanino given that we already mess with the flags enum there, maybe expand the dci pool index to 4 bits instead of adding a separate field?

The cons are:
1. Encountering the same issue next time MAX_DCI_POOLS is increased.
2. Extra opcode (bitwise AND) each time the pool index is being accessed.
3. Index is not semantically a flag - less intuitive.

I guess the benefit is having a smaller ep struct.

roiedanino:
> I guess the benefit is having a smaller ep struct.

Although now we can decrease flags to use `uint8_t` instead of `uint16_t`

brminich:
> > I guess the benefit is having a smaller ep struct.
> 
> Although now we can decrease flags to use `uint8_t` instead of `uint16_t`

imo, better to keep flags as `uint16_t` and use 4 bits for dci pool index. We do not need 8 bits for dci pool index and this way we can keep the rest bits for flags.

test/gtest/uct/uct_test.cc
```diff
@@ -1291,7 +1291,9 @@ void uct_test::entity::connect_to_ep(unsigned index, entity& other,
     }
 }
 
-void uct_test::entity::connect_to_iface(unsigned index, entity& other) {
+void uct_test::entity::connect_to_iface(unsigned index, entity &other,
+                                        unsigned path_index)
```
yosefe, line 1295: shall we add path_index also for connect_to_ep flow, to make it consistent?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -440,6 +440,32 @@ UCS_TEST_P(test_dc, rand_dci_pending_purge) {
...
+
+    rand_e        = create_rand_entity();
+    num_dci_pools = dc_iface(rand_e)->tx.num_dci_pools;
+
+    EXPECT_EQ(num_dci_pools, max_dci_pools);
+
+    /* Create an ep for each pool */
+    for (i = 0; i < max_dci_pools; i++) {
+        rand_e->connect_to_iface(i, *m_e2, i);
+        expected_pool_index = i % num_dci_pools;
```
yosefe, line 463: maybe expected_pool_index could be the current pool index, before the send operation, and not necessarily assume that pool assignment is round robin?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -440,6 +440,32 @@ UCS_TEST_P(test_dc, rand_dci_pending_purge) {
...
+    entity *rand_e;
+    uint8_t num_dci_pools;
+    uct_dc_mlx5_ep_t *ep;
+    uint8_t expected_pool_index;
+    int i;
+
+    /* Set num paths to max */
+    modify_config("IB_NUM_PATHS", std::to_string(max_dci_pools).c_str());
+
+    rand_e        = create_rand_entity();
```
yosefe, line 455: minor: could define vars like "auto num_dci_pools = dc_iface..." or "for (int i = 0; i < max_dci_pools..."

test/gtest/uct/ib/test_dc.cc
roiedanino: I'm not sure what you mean by:
> the current pool index, before the send operation

the pool index is being set internally on ep creation so it will be like comparing pool_index(ep) with itself, no?

yosefe: I see.. any way to change the test to not expect specific pool index?
maybe check that pool index is in a given range?

yosefe: what happens if there are not enough bits?
maybe we can check at least one DCI uses the last pool?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -440,6 +440,32 @@ UCS_TEST_P(test_dc, rand_dci_pending_purge) {
     flush();
 }
 
+UCS_TEST_P(test_dc, pool_indices)
+{
+    constexpr uint8_t max_dci_pools       = UCT_DC_MLX5_IFACE_MAX_DCI_POOLS;
```
yosefe, line 445: maybe define (or static assert) UCT_DC_MLX5_IFACE_MAX_DCI_POOLS according to UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -27,35 +27,35 @@
 
 enum uct_dc_mlx5_ep_flags {
     /* DCI pool EP assigned to according to it's lag port */
-    UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK     = UCS_MASK(3),
+    UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK     = UCS_MASK(5),
```
yosefe, line 30: maybe define (or static assert) UCT_DC_MLX5_IFACE_MAX_DCI_POOLS according to UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -1580,6 +1580,9 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
     UCT_DC_MLX5_CHECK_FORCE_FULL_HANDSHAKE(self, config, dct, DCT, status, err);
 
     ucs_assert(self->tx.num_dci_pools <= UCT_DC_MLX5_IFACE_MAX_DCI_POOLS);
+    UCS_STATIC_ASSERT((UCT_DC_MLX5_IFACE_MAX_DCI_POOL_INDEX &
```
yosefe, line 1583: No need to define UCT_DC_MLX5_IFACE_MAX_DCI_POOL_INDEX just for the assert.
we can
UCS_STATIC_ASSERT((UCT_DC_MLX5_IFACE_MAX_DCI_POOLS - 1) <= UCT_DC_MLX5_EP_FLAG_POOL_INDEX_MASK)

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -27,35 +27,35 @@
...
 
     /* EP is in the tx_wait state. See description of the dcs+quota dci
        selection policy above */
-    UCT_DC_MLX5_EP_FLAG_TX_WAIT             = UCS_BIT(3),
+    UCT_DC_MLX5_EP_FLAG_TX_WAIT             = UCS_BIT(6),
```
brminich, line 34: why not starting from 5?

yosefe (approved):

src/uct/ib/dc/dc_mlx5_ep.h
roiedanino: Sorry, will fix

brminich (approved):

yosefe:
@roiedanino can you pls squash?

PR #9685, opened 2024-02-13
TOOLS/PERF: Third batch commit for DPU xgvmi daemon

iyastreb:
## What
Third batch commit for DPU xgvmi daemon

For reference:
First batch commit https://github.com/openucx/ucx/pull/9536 is merged.
Second batch commit https://github.com/openucx/ucx/pull/9624 is merged.
Original changes: https://github.com/openucx/ucx/pull/8767

## Why ?

### DIFF
--- a/src/tools/perf/lib/libperf.c
+++ b/src/tools/perf/lib/libperf.c
@@ -15,7 +15,9 @@
 #include <ucs/debug/log.h>
 #include <ucs/arch/bitops.h>
 #include <ucs/sys/module.h>
+#include <ucs/sys/sock.h>
 #include <ucs/sys/string.h>
+#include <ucs/type/serialize.h>
 #include <tools/perf/lib/libperf_int.h>
 
 #include <string.h>
@@ -905,6 +907,10 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
         ucp_params->features |= UCP_FEATURE_RMA;
     }
 
+    if (params->ucp.is_daemon_mode) {
+        ucp_params->features |= UCP_FEATURE_AM;
+    }
+
     status = ucx_perf_test_check_params(params);
     if (status != UCS_OK) {
         return status;
@@ -914,33 +920,43 @@ static ucs_status_t ucp_perf_test_fill_params(ucx_perf_params_t *params,
 }
 
 static ucs_status_ptr_t
-ucp_perf_test_destroy_ep(ucp_ep_h ep, const ucp_rkey_h *rkeys, unsigned num_rkeys, unsigned index)
+ucp_perf_test_destroy_ep(ucp_ep_h ep, unsigned index, int daemon_mode)
 {
     ucp_request_param_t ep_close_params = {0};
-    ucs_status_ptr_t *req;
-    unsigned i;
+    ucp_request_param_t dmn_fin_param   = {0};
+    ucs_status_ptr_t req;
 
-    for (i = 0; i < num_rkeys; ++i) {
-        if (rkeys[i] != NULL) {
-            ucp_rkey_destroy(rkeys[i]);
+    if (NULL == ep) {
+        return NULL;
+    }
+
+    if (daemon_mode) {
+        dmn_fin_param.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS;
+        dmn_fin_param.flags        = UCP_AM_SEND_FLAG_EAGER;
+        req = ucp_am_send_nbx(ep, UCP_PERF_DAEMON_AM_ID_FIN, NULL, 0, NULL, 0,
+                              &dmn_fin_param);
+        if (UCS_PTR_IS_PTR(req)) {
+            ucp_request_free(req);
+        } else if (UCS_PTR_STATUS(req) != UCS_OK) {
+            ucs_warn("failed to send FIN message to daemon ep %p "
+                     "on thread %d: %s\n",
+                     ep, index, ucs_status_string(UCS_PTR_STATUS(req)));
         }
     }
 
-    if (ep != NULL) {
-        req = ucp_ep_close_nbx(ep, &ep_close_params);
-        if (UCS_PTR_IS_PTR(req)) {
-            return req;
-        } else if (UCS_PTR_STATUS(req) != UCS_OK) {
-            ucs_warn("failed to close ep %p on thread %d: %s\n", ep, index,
-                     ucs_status_string(UCS_PTR_STATUS(req)));
-        }
+    req = ucp_ep_close_nbx(ep, &ep_close_params);
+    if (UCS_PTR_IS_PTR(req)) {
+        return req;
+    } else if (UCS_PTR_STATUS(req) != UCS_OK) {
+        ucs_warn("failed to close ep %p on thread %d: %s\n", ep, index,
+                 ucs_status_string(UCS_PTR_STATUS(req)));
     }
 
     return NULL;
 }
 
 static void ucp_perf_release_requests_in_progress(ucx_perf_context_t *perf,
-                                                  ucs_status_ptr_t **reqs,
+                                                  ucs_status_ptr_t *reqs,
                                                   unsigned num_in_prog)
 {
     ucs_status_t status;
@@ -958,19 +974,27 @@ static void ucp_perf_release_requests_in_progress(ucx_perf_context_t *perf,
     }
 }
 
+static inline void ucp_perf_test_rkey_destroy(ucp_rkey_h rkey)
+{
+    if (NULL != rkey) {
+        ucp_rkey_destroy(rkey);
+    }
+}
+
 static void ucp_perf_test_destroy_self_eps(ucx_perf_context_t *perf)
 {
-    unsigned thread_count   = perf->params.thread_count;
-    unsigned num_in_prog    = 0;
-    ucs_status_ptr_t **reqs = ucs_alloca(thread_count * sizeof(*reqs));
-    ucs_status_ptr_t *req;
+    unsigned thread_count  = perf->params.thread_count;
+    unsigned num_in_prog   = 0;
+    ucs_status_ptr_t *reqs = ucs_alloca(thread_count * sizeof(*reqs));
+    ucs_status_ptr_t req;
     unsigned i;
 
     for (i = 0; i < thread_count; ++i) {
-        ucp_rkey_h self_rkeys[] = {perf->ucp.tctx[i].perf.ucp.self_send_rkey,
-                                   perf->ucp.tctx[i].perf.ucp.self_recv_rkey};
-        req = ucp_perf_test_destroy_ep(perf->ucp.tctx[i].perf.ucp.self_ep,
-                                       self_rkeys, 2, i);
+        ucp_perf_test_rkey_destroy(perf->ucp.tctx[i].perf.ucp.self_send_rkey);
+        ucp_perf_test_rkey_destroy(perf->ucp.tctx[i].perf.ucp.self_recv_rkey);
+
+        req = ucp_perf_test_destroy_ep(perf->ucp.tctx[i].perf.ucp.self_ep, i,
+                                       perf->params.ucp.is_daemon_mode);
         if (req != NULL) {
             reqs[num_in_prog++] = req;
         }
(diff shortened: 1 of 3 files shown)


src/tools/perf/lib/libperf.c
```diff
@@ -1267,6 +1282,164 @@ static ucs_status_t ucp_perf_test_send_local_data(ucx_perf_context_t *perf,
...
+    if (status != UCS_OK) {
+        ucs_error("failed to create endpoint: %s", ucs_status_string(status));
+        return status;
+    }
+
+    /* Create and send HOST_INIT message to the local DPU daemon */
+    status = ucp_perf_daemon_init_send(perf, perf->ucp.tctx[0].perf.ucp.ep,
+                                       remote_dmn_addr);
+    if (status != UCS_OK) {
+        /* Error is logged */
```
brminich, line 1435: what does this comment mean? I'd remove it

src/tools/perf/lib/libperf.c
```diff
@@ -1400,16 +1573,23 @@ static ucs_status_t ucp_perf_test_setup_endpoints(ucx_perf_context_t *perf,
...
-    if (status != UCS_OK) {
-        goto err;
-    }
+    if (perf->params.ucp.is_daemon_mode) {
+        status = ucp_perf_setup_daemon_endpoints(perf);
+        if (status != UCS_OK) {
+            goto err;
+        }
+    } else {
+        /* pack the local endpoints data and send to the remote peer */
```
brminich, line 1582:
```suggestion
        /* Pack the local endpoints data and send to the remote peer */
```

src/tools/perf/lib/libperf.c
```diff
@@ -1400,16 +1573,23 @@ static ucs_status_t ucp_perf_test_setup_endpoints(ucx_perf_context_t *perf,
...
+        status = ucp_perf_test_send_local_data(perf, features);
+        if (status != UCS_OK) {
+            goto err;
+        }
 
-    /* Receive remote peer's endpoints' data and connect to them */
-    status = ucp_perf_test_receive_remote_data(perf, peer_index);
-    if (status != UCS_OK) {
-        goto err;
+        /* receive remote peer's endpoints' data and connect to them */
```
brminich, line 1588:
```suggestion
        /* Receive remote peer's endpoints' data and connect to them */
```

src/tools/perf/lib/libperf.c
iyastreb: It explains why we don't log an error here - because it was logged already in the function
But ok to remove

src/tools/perf/lib/libperf.c
iyastreb: ok

src/tools/perf/lib/libperf.c
iyastreb: ok

brminich (dismissed):

src/tools/perf/lib/libperf.c
```diff
@@ -958,38 +980,31 @@ static void ucp_perf_release_requests_in_progress(ucx_perf_context_t *perf,
...
 
-    for (i = 0; i < thread_count; ++i) {
-        ucp_rkey_h self_rkeys[] = {perf->ucp.tctx[i].perf.ucp.self_send_rkey,
-                                   perf->ucp.tctx[i].perf.ucp.self_recv_rkey};
-        req = ucp_perf_test_destroy_ep(perf->ucp.tctx[i].perf.ucp.self_ep,
-                                       self_rkeys, 2, i);
-        if (req != NULL) {
-            reqs[num_in_prog++] = req;
-        }
+#define UCP_RKEY_DESTROY(_rkey) \
```
yosefe, line 992: maybe define inline function?

src/tools/perf/lib/libperf.c
```diff
@@ -958,38 +980,31 @@ static void ucp_perf_release_requests_in_progress(ucx_perf_context_t *perf,
...
+    unsigned thread_count  = perf->params.thread_count;
+    unsigned num_in_prog   = 0;
+    ucs_status_ptr_t *reqs = ucs_alloca(thread_count * sizeof(*reqs));
+    ucs_status_ptr_t req;
     unsigned i;
 
-    for (i = 0; i < thread_count; ++i) {
-        ucp_rkey_h self_rkeys[] = {perf->ucp.tctx[i].perf.ucp.self_send_rkey,
-                                   perf->ucp.tctx[i].perf.ucp.self_recv_rkey};
-        req = ucp_perf_test_destroy_ep(perf->ucp.tctx[i].perf.ucp.self_ep,
```
yosefe, line 972: where do we destroy `perf->ucp.tctx[i].perf.ucp.self_ep` now?

src/tools/perf/lib/libperf.c
```diff
@@ -958,38 +980,31 @@ static void ucp_perf_release_requests_in_progress(ucx_perf_context_t *perf,
...
-    unsigned thread_count   = perf->params.thread_count;
-    unsigned num_in_prog    = 0;
-    ucs_status_ptr_t **reqs = ucs_alloca(thread_count * sizeof(*reqs));
-    ucs_status_ptr_t *req;
-    unsigned i;
-
     for (i = 0; i < thread_count; ++i) {
-        req = ucp_perf_test_destroy_ep(perf->ucp.tctx[i].perf.ucp.ep,
-                                       &perf->ucp.tctx[i].perf.ucp.rkey, 1, i);
+        if (ep_domain & UCP_EP_DOMAIN_SELF) {
```
yosefe, line 998: why need this check, isn't it enough we're checking rkey!=NULL?

src/tools/perf/lib/libperf.c
```diff
@@ -1267,6 +1282,163 @@ static ucs_status_t ucp_perf_test_send_local_data(ucx_perf_context_t *perf,
...
+ * Pack array length and array contents into a buffer.
+ *
+ * @param [out] ptr    Pointer to the buffer.
+ * @param [in]  length Length of the array.
+ * @param [in]  value  Pointer to the array.
+ */
+static void ucp_perf_pack_array(void **ptr, uint64_t length, const void *value)
+{
+    *ucs_serialize_next(ptr, uint64_t) = length;
+    if (length > 0) {
```
yosefe, line 1301: why needed ? to avoid static check/coverity error?

src/tools/perf/lib/libperf.c
```diff
@@ -1267,6 +1282,163 @@ static ucs_status_t ucp_perf_test_send_local_data(ucx_perf_context_t *perf,
...
+        memcpy(ucs_serialize_next_raw(ptr, void, length), value, length);
+    }
+}
+
+static size_t ucp_perf_pack_daemon_info(const ucx_perf_context_t *perf,
+                                        void *buffer, size_t buffer_size,
+                                        const struct sockaddr_storage *address)
+{
+    void *p               = buffer;
+    size_t address_length = (address != NULL) ? sizeof(*address) : 0;
```
yosefe, line 1311: maybe pass the length? the actual ipv4/ipv6 length is not really as big as sockaddr_storage

src/tools/perf/lib/libperf.c
```diff
@@ -1267,6 +1282,163 @@ static ucs_status_t ucp_perf_test_send_local_data(ucx_perf_context_t *perf,
...
+    int is_local_sender       = (group_index % 2) != 0;
+    int is_remote_sender      = (peer_group_index % 2) != 0;
+    ucp_ep_params_t ep_params = {};
+    struct sockaddr_storage *local_dmn_addr;
+    struct sockaddr_storage *remote_dmn_addr;
+    ucs_status_t status;
+
+    if (!perf->params.ucp.is_daemon_mode) {
+        return UCS_OK;
+    }
```
yosefe, line 1391: IMO better check it outside the function

src/tools/perf/lib/libperf.c
```diff
@@ -980,33 +981,43 @@ static void ucp_perf_release_requests_in_progress(ucx_perf_context_t *perf,
...
-
     for (i = 0; i < thread_count; ++i) {
+        thread_perf = &perf->ucp.tctx[i].perf;
         if (ep_domain & UCP_EP_DOMAIN_SELF) {
-            UCP_RKEY_DESTROY(perf->ucp.tctx[i].perf.ucp.self_send_rkey);
-            UCP_RKEY_DESTROY(perf->ucp.tctx[i].perf.ucp.self_recv_rkey);
+            ucp_perf_test_rkey_destroy(thread_perf->ucp.self_send_rkey);
+            ucp_perf_test_rkey_destroy(thread_perf->ucp.self_recv_rkey);
+
+            if ((req = ucp_perf_test_destroy_ep(thread_perf->ucp.self_ep, i,
```
yosefe, line 1008: IMO better separate "req = ..." to previous line. same for line 1017

src/tools/perf/lib/libperf.c
```diff
@@ -1323,7 +1338,7 @@ static size_t ucp_perf_pack_daemon_info(const ucx_perf_context_t *perf,
 
 static size_t ucp_perf_daemon_info_size(const ucx_perf_context_t *perf)
 {
-    return sizeof(uint64_t) + sizeof(struct sockaddr_storage) +
+    return sizeof(uint64_t) + UCS_IPV6_SOCKADDR_LEN +
```
yosefe, line 1341: why IPv6?

src/tools/perf/lib/libperf.c
yosefe: won't the next line `memcpy(ucs_serialize_next_raw(ptr, void, length), value, length);` be no-operation anyway when length==0?

src/tools/perf/lib/libperf.c
```diff
@@ -1267,6 +1282,163 @@ static ucs_status_t ucp_perf_test_send_local_data(ucx_perf_context_t *perf,
     return status;
 }
 
+static void err_cb(void *arg, ucp_ep_h ep, ucs_status_t status)
```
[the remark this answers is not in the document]
yosefe: so can we print something like this :
```c
static void ucp_perf_err_cb(void *arg, ucp_ep_h ep, ucs_status_t status)
 {
     ucs_diag("failed to communicate with the daemon on %s:%d: %s", daemon_hostname, daemon_port
              ucs_status_string(status));
 }
```

src/tools/perf/lib/libperf.c
yosefe: so maybe split it to different functions, as it was before?

src/tools/perf/lib/libperf.c
iyastreb: This function returns buffer size large enough to hold the message to the daemon.
Since address that we pass can be either IPv4 or IPv6 or NULL, I decided to use the larger size that would fit all cases. Since the diff between the 2 is just 12 bytes, I thought it's not a big deal if we allocate a bit more just in case. This way we don't need to check whether address is IPv4/IPv6/empty in advance.
I think the same pattern is used in rkey serialization

If you believe it makes sense to calculate the precise buffer length before allocation - let me know, I'll modify the code accordingly

src/tools/perf/lib/libperf.c
iyastreb: Yes, it will be no-op
Personally I would still prefer to have that explicit check before the function call to make the intention clear. But if you have strong opinion on that - I can remove that condition, not a big deal anyway

src/tools/perf/lib/libperf.c
iyastreb: Ok, agree, after all I see more code with new approach
Reverted

src/tools/perf/lib/libperf.c
iyastreb: ok

src/tools/perf/lib/libperf.c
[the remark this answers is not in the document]
iyastreb: done

src/tools/perf/lib/libperf.c
yosefe: let's see how complex it is to pass exact address size. maybe reuse that size value also in the send function

iyastreb: ok

src/tools/perf/lib/libperf.c
iyastreb: Static check complains on this change. So I put the condition back
```
static checkers found errors:
Error: CLANG_WARNING:
/src/tools/perf/lib/libperf.c:1326:5: warning: Null pointer passed to 2nd parameter expecting 'nonnull' [core.NonNullParamChecker]
#    memcpy(ucs_serialize_next_raw(ptr, void, length), value, length);
```

src/tools/perf/lib/libperf.c
```diff
@@ -988,36 +981,41 @@ static inline void ucp_perf_test_rkey_destroy(ucp_rkey_h rkey)
...
-            ucp_perf_test_rkey_destroy(thread_perf->ucp.self_send_rkey);
-            ucp_perf_test_rkey_destroy(thread_perf->ucp.self_recv_rkey);
-
-            if ((req = ucp_perf_test_destroy_ep(thread_perf->ucp.self_ep, i,
-                                                daemon_mode)) != NULL) {
-                reqs[num_in_prog++] = req;
-            }
+        ucp_perf_test_rkey_destroy(perf->ucp.tctx[i].perf.ucp.self_send_rkey);
+        ucp_perf_test_rkey_destroy(perf->ucp.tctx[i].perf.ucp.self_recv_rkey);
+        req = ucp_perf_test_destroy_ep(perf->ucp.tctx[i].perf.ucp.self_ep, i,
```
yosefe, line 995: space line before

src/tools/perf/lib/libperf.c
```diff
@@ -988,36 +981,41 @@ static inline void ucp_perf_test_rkey_destroy(ucp_rkey_h rkey)
...
+{
+    unsigned thread_count  = perf->params.thread_count;
+    unsigned num_in_prog   = 0;
+    ucs_status_ptr_t *reqs = ucs_alloca(thread_count * sizeof(*reqs));
+    ucs_status_ptr_t req;
+    unsigned i;
+
+    for (i = 0; i < thread_count; ++i) {
+        ucp_perf_test_rkey_destroy(perf->ucp.tctx[i].perf.ucp.rkey);
+        req = ucp_perf_test_destroy_ep(perf->ucp.tctx[i].perf.ucp.ep, i,
```
yosefe, line 1015: space line before

src/tools/perf/lib/libperf.c
iyastreb: done

src/tools/perf/lib/libperf.c
iyastreb: done

yosefe (approved):

yosefe (approved):

PR #9695, opened 2024-02-18
UCP/PROTO: Move non-rndv protocols to use probe() instead of init()

yosefe:
## Why
Next phase of adding protocol variants, moving from init() to probe()

### DIFF
--- a/src/ucp/proto/proto_multi.c
+++ b/src/ucp/proto/proto_multi.c
@@ -19,8 +19,8 @@
 
 
 ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
-                                  ucp_proto_multi_priv_t *mpriv,
-                                  size_t *priv_size_p)
+                                  ucp_proto_caps_t *caps,
+                                  ucp_proto_multi_priv_t *mpriv)
 {
     ucp_context_h context         = params->super.super.worker->context;
     const double max_bw_ratio     = context->config.ext.multi_lane_max_ratio;
@@ -213,10 +213,6 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
     }
     ucs_assert(mpriv->num_lanes == ucs_popcount(lane_map));
 
-    /* Fill the size of private data according to number of used lanes */
-    *priv_size_p = sizeof(ucp_proto_multi_priv_t) +
-                   (mpriv->num_lanes * sizeof(*lpriv));
-
     /* After this block, 'perf_node' and 'lane_perf_nodes[]' have extra ref */
     if (mpriv->num_lanes == 1) {
         perf_node = lanes_perf_nodes[ucs_ffs64(lane_map)];
@@ -231,8 +227,7 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
     }
 
     status = ucp_proto_common_init_caps(&params->super, &perf, perf_node,
-                                        reg_md_map);
-
+                                        reg_md_map, caps);
     /* Deref unused nodes */
     for (i = 0; i < num_lanes; ++i) {
         ucp_proto_perf_node_deref(&lanes_perf_nodes[lanes[i]]);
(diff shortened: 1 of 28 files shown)


yosefe:
@ivankochin @brminich can you pls review?

src/ucp/proto/proto_multi.c
```diff
@@ -242,17 +237,44 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
     return status;
 }
 
+size_t
+ucp_proto_multi_priv_size(const ucp_proto_multi_priv_t *mpriv, size_t full_size)
+{
+    ucs_assert(full_size >= sizeof(ucp_proto_multi_priv_t));
```
brminich, line 243: looks hacky, is it possible to calculate just `mpriv->num_lanes*ucs_field_sizeof(ucp_proto_multi_priv_t, lanes[0])` and the rest calculate separatly in each place?

ivankochin:
Not directly connected to this patch, but anyway. As I understand, at the current moment, failing of probing or initialization doesn't  entail any traces printing. Maybe it worth adding some logging which would print status of  initialization or how many variants were added by probe? e.g. extend `ucp_proto_select_add_proto`

src/ucp/proto/proto_multi.h
```diff
@@ -18,6 +18,33 @@
...
+    ({ \
+        typedef ucs_typeof(*(_priv)) _type; \
+        \
+        /* Make sure _mpriv is the last field in _type */ \
+        UCS_STATIC_ASSERT((ucs_offsetof(_type, _mpriv) + \
+                           sizeof(ucp_proto_multi_priv_t)) == sizeof(_type)); \
+        \
+        /* Subtract the full size of _mpriv and add the actual size. */ \
+        sizeof(_type) - sizeof(ucp_proto_multi_priv_t) + \
+                ucp_proto_multi_priv_size(&(_priv)->_mpriv); \
```
brminich, line 44: maybe `ucs_offsetof(_type, _mpriv) + ucp_proto_multi_priv_size(&(_priv)->_mpriv)`?

src/ucp/proto/proto_multi.h
```diff
@@ -65,7 +95,7 @@ typedef struct {
     ucp_lane_index_t            num_lanes;    /* Number of lanes to use */
     size_t                      align_thresh; /* Cached value of threshold for
                                                  enabling data split alignment */
-    ucp_proto_multi_lane_priv_t lanes[0];     /* Array of lanes */
+    ucp_proto_multi_lane_priv_t lanes[UCP_MAX_LANES]; /* Array of lanes */
```
ivankochin, line 98: Why changed to `UCP_MAX_LENGTH`? 

Minor: Align comments for other fields.

src/ucp/proto/proto_select.c
```diff
@@ -716,8 +716,10 @@ void ucp_proto_select_add_proto(const ucp_proto_init_params_t *init_params,
     ucp_proto_id_t proto_id                       = init_params->proto_id;
     ucp_proto_init_elem_t *init_elem;
     const char *proto_name;
-    size_t priv_offset;
+    char min_length_str[64];
+    char thresh_str[64];
```
ivankochin, line 721: Extra blank line?

src/ucp/proto/proto_multi.h
ivankochin: I thought that rpriv is always `UCP_PROTO_PRIV_MAX` buffer allocated on the `ucp_proto_select _init_protocols` side. Maybe I am missing something, but I cannot find the place where we allocate `ucp_proto_multi_priv_t` using alloca.

Is that change made for some future improvements?

yosefe: For example here: https://github.com/openucx/ucx/pull/9695/files#diff-a430ba778d55dee29e9281cc476b4563ea1732a2462db82afb8dcd1c0bd1972aR249 we place ucp_proto_multi_priv_t  on the stack; if it had zero-size array of lanes we could not do that - we had to allocate with alloca and add space for the lanes (which create more code).

brminich (dismissed):

ivankochin (approved):

src/ucp/proto/proto_multi.h
ivankochin: Understood, thanks!

src/ucp/proto/proto_select.c
yosefe: > Extra blank line?

fixed

PR #9696, opened 2024-02-19
CMAKE: Add targets once in case of multiple find_package()

tvegas1:
## What
Check target existence before adding them.

## Why ?
Triggers error when `find_package()` is invoked multiple times.

## How ?
Derived from #9622, but since we don't generate the ucx cmake target file, take simpler approach.


brminich (approved):

yosefe (approved):

vyasr:
Confirming that this should be practically equivalent to #9622.

Issue #9626, opened 2024-01-24
ucp_put_nbx() Fails to Deliver Data when UCP_OP_ATTR_FIELD_DATATYPE is not Set and UCP_OP_ATTR_FLAG_NO_IMM_CMPL is set


weixuanzh:
### Describe the bug
When ucp_put_nbx() is called using UCP_OP_ATTR_FLAG_NO_IMM_CMPL and without setting UCP_OP_ATTR_FIELD_DATATYPE in parameters, the remote receiving buffer never receives the message, even though ucp_put_nbx() is completed both locally and remotely.

Related Code (from ucx/src/ucp/rma/rma_send.c)

    ucp_worker_h worker     = ep->worker;
    size_t contig_length    = 0;
    ucp_datatype_t datatype = ucp_dt_make_contig(1);
    ucp_ep_rma_config_t *rma_config;
    ucs_status_ptr_t ret;
    ucs_status_t status;
    ucp_request_t *req;
    uint32_t attr_mask;

    UCP_REQUEST_CHECK_PARAM(param);
    UCP_RMA_CHECK_PTR(worker->context, buffer, count);
    UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);

    ucs_trace_req("put_nbx buffer %p count %zu remote_addr %" PRIx64
                  " rkey %p to %s cb %p",
                  buffer, count, remote_addr, rkey, ucp_ep_peer_name(ep),
                  ucp_request_param_send_callback(param));

    attr_mask = param->op_attr_mask &
                (UCP_OP_ATTR_FIELD_DATATYPE | UCP_OP_ATTR_FLAG_NO_IMM_CMPL);

    if (worker->context->config.ext.proto_enable) {
        status = ucp_put_send_short(ep, buffer, count, remote_addr, rkey, param);
        if (ucs_likely(status != UCS_ERR_NO_RESOURCE)) {
            ret = UCS_STATUS_PTR(status);
            goto out_unlock;
        }

        req = ucp_request_get_param(worker, param,
                                    {ret = UCS_STATUS_PTR(UCS_ERR_NO_MEMORY);
                                    goto out_unlock;});
        req->send.rma.rkey        = rkey;
        req->send.rma.remote_addr = remote_addr;

        if (ucs_likely(attr_mask == 0)) {
            contig_length = count;
        } else if (attr_mask & UCP_OP_ATTR_FIELD_DATATYPE) {
            datatype = param->datatype;
[...]

PR #9700, opened 2024-02-20
UCP/PROTOV2: Don't depend on NO_IMM_CMPL for DATATYPE checks

tvegas1:
## What
Check datatype field independently from no immediate completion flag.

## Why ?
Reported by #9626.

### DIFF
--- a/src/ucp/rma/rma_send.c
+++ b/src/ucp/rma/rma_send.c
@@ -250,7 +250,6 @@ ucs_status_ptr_t ucp_put_nbx(ucp_ep_h ep, const void *buffer, size_t count,
     ucs_status_ptr_t ret;
     ucs_status_t status;
     ucp_request_t *req;
-    uint32_t attr_mask;
 
     UCP_REQUEST_CHECK_PARAM(param);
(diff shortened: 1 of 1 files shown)


tvegas1:
> seems that `UCP_OP_ATTR_FLAG_NO_IMM_CMPL` is not really supported for `put_nbx` can you pls fix it in this pr then?

it is supported inside `ucp_put_send_short()` so we should be good i think.

brminich:
> > seems that `UCP_OP_ATTR_FLAG_NO_IMM_CMPL` is not really supported for `put_nbx` can you pls fix it in this pr then?
> 
> it is supported inside `ucp_put_send_short()` so we should be good i think.

indeed, then can we move `attr_mask` initialization to protov1 section to avoid unneeded overhead for protov2?

tvegas1:
> indeed, then can we move `attr_mask` initialization to protov1 section to avoid unneeded overhead for protov2?

makes sense, removed it altogether.

yosefe (approved):

yosefe:
@tvegas1 can you pls squash?

PR #9702, opened 2024-02-20
CMAKE: Add targets once in case of multiple find_package() - v1.16.x

tvegas1:
## What
Backport #9696 to support multiple cmake `find_package()` invocations.


rakhmets:
> > we'd also need `NEWS` file update under rc3?
> 
> I've just created a pre-release... But I can remove tag and RC to avoid having two RCs for two single PRs.

Please ignore my suggestion to remove the existing tag.
NEWS should be updated by adding section for RC4.

cc @shasson5

gleon99 (approved):

PR #9678, opened 2024-02-08
UCT/GGA: initial commit for gga_mlx5 transport

evgeny-leksikov:
## What
initial commit for gga_mlx5 transport

## Why ?
new HW capabilities support

PR #9703, opened 2024-02-20
UCT/GGA: md api implementation

evgeny-leksikov:
## What
MD API implementation of GGA transport

## Why ?
continue https://github.com/openucx/ucx/pull/9678

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -47,7 +47,7 @@ static const char *uct_ib_devx_objs[] = {
     NULL
 };
 
-static ucs_config_field_t uct_ib_md_config_table[] = {
+ucs_config_field_t uct_ib_md_config_table[] = {
     {"", "", NULL,
      ucs_offsetof(uct_ib_md_config_t, super), UCS_CONFIG_TYPE_TABLE(uct_md_config_table)},
 
@@ -187,7 +187,6 @@ extern uct_tl_t UCT_TL_NAME(rc_verbs);
 extern uct_tl_t UCT_TL_NAME(rc_mlx5);
 extern uct_tl_t UCT_TL_NAME(ud_verbs);
 extern uct_tl_t UCT_TL_NAME(ud_mlx5);
-extern uct_tl_t UCT_TL_NAME(gga_mlx5);
 
 static uct_tl_t *uct_ib_tls[] = {
 #ifdef HAVE_TL_DC
@@ -198,9 +197,6 @@ static uct_tl_t *uct_ib_tls[] = {
 #endif
 #if defined (HAVE_TL_RC) && defined (HAVE_MLX5_DV)
     &UCT_TL_NAME(rc_mlx5),
-#ifdef HAVE_DEVX
-    &UCT_TL_NAME(gga_mlx5),
-#endif
 #endif
 #ifdef HAVE_TL_UD
     &UCT_TL_NAME(ud_verbs),
@@ -737,9 +733,9 @@ ucs_status_t uct_ib_verbs_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
     return UCS_OK;
 }
 
-static ucs_status_t uct_ib_rkey_unpack(uct_component_t *component,
-                                       const void *rkey_buffer,
-                                       uct_rkey_t *rkey_p, void **handle_p)
+ucs_status_t uct_ib_rkey_unpack(uct_component_t *component,
+                                const void *rkey_buffer, uct_rkey_t *rkey_p,
+                                void **handle_p)
 {
     uint64_t packed_rkey = *(const uint64_t*)rkey_buffer;
 
@@ -808,9 +804,9 @@ int uct_ib_device_is_accessible(struct ibv_device *device)
     return uct_ib_device_is_supported(device);
 }
 
-static ucs_status_t uct_ib_query_md_resources(uct_component_t *component,
-                                              uct_md_resource_desc_t **resources_p,
-                                              unsigned *num_resources_p)
+ucs_status_t uct_ib_query_md_resources(uct_component_t *component,
+                                       uct_md_resource_desc_t **resources_p,
+                                       unsigned *num_resources_p)
 {
     UCS_MODULE_FRAMEWORK_DECLARE(uct_ib);
     int num_resources = 0;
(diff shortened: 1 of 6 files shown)


src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -988,6 +988,49 @@ uct_ib_mlx5_devx_md_get_counter_set_id(uct_ib_mlx5_md_t *md, uint8_t port_num)
 
 #endif
 
+/**
+ * DEVX MD API
+ */
+#if HAVE_DEVX
```
tvegas1, line 994: shall we reuse the existing `#if HAVE_DEVX` above line 820?

evgeny-leksikov: the existing `#if HAVE_DEVX` block defines `DEVX QP API` and have its stubs in`else` block which is not required for `MD API`, so defining it in separate section looks better to me.

evgeny-leksikov:
Related CI failure is addressed in https://github.com/openucx/ucx/pull/9706

src/uct/ib/rc/accel/gga_mlx5.c
```diff
@@ -59,6 +59,43 @@ static UCS_CLASS_DEFINE_NEW_FUNC(uct_gga_mlx5_iface_t, uct_iface_t, uct_md_h,
...
+                                     const uct_ib_md_config_t *md_config,
+                                     struct uct_ib_md **md_p)
+{
+    ucs_status_t status;
+
+    if (md_config->devx == UCS_NO) {
+        return UCS_ERR_UNSUPPORTED;
+    }
+
+    status = uct_ib_mlx5_devx_md_open(ibv_device, md_config, md_p);
```
yosefe, line 72: maybe gga can use same md object? why need different md for gga?

src/uct/ib/base/ib_md.c
```diff
@@ -1080,29 +1107,24 @@ ucs_status_t uct_ib_md_open(uct_component_t *component, const char *md_name,
...
-            break;
-        } else if (status != UCS_ERR_UNSUPPORTED) {
-            goto out_free_dev_list;
-        }
-        ucs_debug("%s: md open by '%s' failed, trying next", md_name,
-                  uct_ib_ops[i]->name);
+    if (component == &uct_ib_component) {
+        status = uct_ib_component_md_open(ib_device, md_config, md_name, &md);
+    } else {
+        status = uct_ib_mlx5_gga_md_open(ib_device, md_config, &md);
```
yosefe, line 1113: can we call this function directly from gga code (so gga will not point to uct_ib_md_open, but to its own func)?

src/uct/ib/rc/accel/gga_mlx5.c
```diff
@@ -34,11 +45,157 @@ ucs_config_field_t uct_gga_mlx5_iface_config_table[] = {
...
+                         uct_rkey_t *rkey_p, void **handle_p)
+{
+    const uct_ib_md_packed_mkey_t *mkey = rkey_buffer;
+    uct_rkey_bundle_t *rkey_bundle;
+    uct_gga_mlx5_rkey_handle_t *rkey_handle;
+
+    rkey_bundle = ucs_malloc(sizeof(*rkey_bundle), "gga_rkey_bundle");
+    if (rkey_bundle == NULL) {
+        return UCS_ERR_NO_MEMORY;
+    }
```
yosefe, line 78: can we avoid allocating uct_rkey_bundle_t?

src/uct/ib/rc/accel/gga_mlx5.c
```diff
@@ -34,11 +45,157 @@ ucs_config_field_t uct_gga_mlx5_iface_config_table[] = {
...
+        .field_mask = UCT_MD_MEM_DEREG_FIELD_MEMH,
+        .memh       = rkey_handle->memh
+    };
+    ucs_status_t status;
+
+    if (rkey_handle->memh == NULL) {
+        return;
+    }
+
+    status = uct_ib_mlx5_devx_mem_dereg(&rkey_handle->md->super.super, &params);
```
yosefe, line 116: can we use more internal functions of ib_mlx5? for example, in case of gga there should not be atomic MR

src/uct/ib/rc/accel/gga_mlx5.h
```diff
@@ -0,0 +1,31 @@
...
+
+#else /* HAVE_DEVX */
+
+static inline ucs_status_t
+uct_ib_mlx5_gga_md_open(struct ibv_device *ibv_device,
+                        const uct_ib_md_config_t *md_config,
+                        struct uct_ib_md **md_p)
+{
+    return UCS_ERR_UNSUPPORTED;
+}
```
yosefe, line 27: why needed? can we avoid defining gga component altogether if no devx?

src/uct/ib/rc/accel/gga_mlx5.c
evgeny-leksikov: using mpool?

src/uct/ib/rc/accel/gga_mlx5.h
evgeny-leksikov: because of `ib_md.c:1113` (I will try to refactor), gga component won't be defined without devx.

src/uct/ib/rc/accel/gga_mlx5.c
evgeny-leksikov: Then better to define different memh types as well. WDYT if it will be done in separate PR since it might be large and sensitive change?

yosefe: ok

src/uct/ib/rc/accel/gga_mlx5.c
```diff
@@ -91,12 +82,8 @@ uct_gga_mlx5_rkey_unpack(uct_component_t *component, const void *rkey_buffer,
...
     rkey_handle->rkey_ob.type   = NULL;
 
-    rkey_bundle->handle      = rkey_handle;
-    rkey_bundle->rkey        = (uintptr_t)rkey_bundle;
-    rkey_bundle->type        = NULL;
-
-    *rkey_p   = rkey_bundle->rkey;
-    *handle_p = rkey_bundle->handle;
+    *rkey_p   = (uintptr_t)rkey_handle;
+    *handle_p = rkey_handle;
```
yosefe, line 86: IMO the handle can be NULL - and use rkey

src/uct/ib/rc/accel/gga_mlx5.c
```diff
@@ -265,23 +254,55 @@ static uct_md_ops_t uct_mlx5_gga_md_ops = {
...
-                                     const uct_ib_md_config_t *md_config,
-                                     struct uct_ib_md **md_p)
+static ucs_status_t
+uct_ib_mlx5_gga_md_open(uct_component_t *component, const char *md_name,
+                        const uct_md_config_t *uct_md_config, uct_md_h *md_p)
 {
-    ucs_status_t status;
+    const uct_ib_md_config_t *md_config = ucs_derived_of(uct_md_config,
+                                                         uct_ib_md_config_t);
+    ucs_status_t status = UCS_ERR_UNSUPPORTED;
```
yosefe, line 263: why need to initialize?

src/uct/ib/rc/accel/gga_mlx5.c
```diff
@@ -265,23 +254,55 @@ static uct_md_ops_t uct_mlx5_gga_md_ops = {
...
-    status = uct_ib_mlx5_devx_md_open(ibv_device, md_config, md_p);
+    /* Get device list from driver */
+    ib_device_list = ibv_get_device_list(&num_devices);
+    if (ib_device_list == NULL) {
+        ucs_debug("Failed to get GGA device list, assuming no devices are present");
+        status = UCS_ERR_NO_DEVICE;
+        goto out;
+    }
+
+    status = uct_ib_get_device_by_name(ib_device_list, num_devices, md_name,
```
yosefe, line 282: maybe unite ibv_get_device_list and uct_ib_get_device_by_name to a single function?

evgeny-leksikov: I didn't implement this way because the device is a pointer to device list element which is used later in `uct_ib_mlx5_devx_md_open` and similar in IB, so the list cannot be freed in the united function.

yosefe: it's a pointer to a pointer, so should be ok

evgeny-leksikov: https://linux.die.net/man/3/ibv_get_device_list
```
Client code should open all the devices it intends to use with ibv_open_device() before calling ibv_free_device_list(). Once it frees the array with ibv_free_device_list(), it will be able to use only the open devices; pointers to unopened devices will no longer be valid.
```

yosefe (approved):

PR #9706, opened 2024-02-21
UCT/IB/GGA: fix compilation

evgeny-leksikov:
## What
fix compilation without devx

## Why ?
gga_mlx5.c file was wrongly added under `HAVE_MLX5_DV` condition instead of `HAVE_DEVX`

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -198,7 +198,9 @@ static uct_tl_t *uct_ib_tls[] = {
 #endif
 #if defined (HAVE_TL_RC) && defined (HAVE_MLX5_DV)
     &UCT_TL_NAME(rc_mlx5),
+#ifdef HAVE_DEVX
     &UCT_TL_NAME(gga_mlx5),
+#endif
 #endif
 #ifdef HAVE_TL_UD
     &UCT_TL_NAME(ud_verbs),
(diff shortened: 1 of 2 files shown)


src/uct/ib/base/ib_md.c
```diff
@@ -190,20 +190,22 @@ extern uct_tl_t UCT_TL_NAME(ud_mlx5);
...
+#if HAVE_DEVX
     &UCT_TL_NAME(gga_mlx5),
 #endif
-#ifdef HAVE_TL_UD
+#endif
+#if HAVE_TL_UD
     &UCT_TL_NAME(ud_verbs),
 #endif
-#if defined (HAVE_TL_UD) && defined (HAVE_MLX5_HW_UD)
+#if HAVE_TL_UD && HAVE_MLX5_HW_UD
```
yosefe, line 208: why needed to change?

evgeny-leksikov: AFAIR `HAVE_XXX` definitions are always defined to `0` or `1` and check it with `#if defined HAVE_XXX` is not correct

yosefe: this seems like unrelated fix that needs another PR?

evgeny-leksikov: The change regarding to `&UCT_TL_NAME(gga_mlx5)` at line 201 is related and since the patch is small I decided to unite, but OK, I'll split

evgeny-leksikov: just checked these definitions - they are undefined if built on system without IB or for example, with `--with-devx=no` configure parameter, otherwise they are defined as `1`
Also, C99 standard says that "all remaining identifiers ...  are replaced with the pp-number 0", so it seems to be safe to use `#if HAVE_XXX` as well as `#ifdef`

yosefe (approved):

PR #9708, opened 2024-02-23
CI: Enable ASAN roce testing

ivankochin:
## What
Enable ASAN testing for `ucx_roce` setup.

### DIFF
--- a/buildlib/pr/main.yml
+++ b/buildlib/pr/main.yml
@@ -458,13 +458,6 @@ stages:
         demands: ucx_roce -equals yes
         test_perf: 0
         valgrind_check: yes
-    - template: tests.yml
-      parameters:
-        name: roce_proto_disable
-        demands: ucx_roce -equals yes
-        test_perf: 0
-        proto_enable: no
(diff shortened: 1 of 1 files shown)


buildlib/pr/main.yml
```diff
@@ -432,6 +432,21 @@ stages:
...
+        container: ubuntu2204_ib
+        asan_check: yes
+    - template: tests.yml
+      parameters:
+        name: roce_proto_disable
+        demands: ucx_roce -equals yes
+        test_perf: 0
+        proto_enable: no
+        container: ubuntu2204_ib
+        asan_check: yes
```
yosefe, line 449: IMO this is not needed - we can remove to reduce testing time
@brminich WDYT?

brminich: maybe we can leave ASAN but remove valgrind for proto disable instead?

ivankochin: Done.

yosefe (approved):

PR #9712, opened 2024-02-26
UCP/RNDV: Make sure lane max_frag is never overriden

tvegas1:
## What
Remove assert and enforce lane `max_frag`.

## Why ?
After 4909af673af7f2e3dcd2f98410213374f53d9fca, assertion is not valid anymore since `total_offset` can grow faster than the `scaled_length` which comes from pre-computed weights.

## How ?
Remove assert line, don't explicitly check for underflow, but make sure that `max_payload` always remains below lane `max_frag`.

Report from internal [3786221](https://redmine.mellanox.com/issues/3786221):
```
proto_rndv.inl:264  Assertion `scaled_length >= total_offset' failed: req=0x1ddc3c0 \
 min_rndv_chunk_size=16384 scaled_length=13108 total_offset=16384 total_length=32768 weight_sum=41%
```


tvegas1:
Unfortunately, I did not find a reasonable way to adapt the assertion.

src/ucp/rndv/proto_rndv.inl
```diff
@@ -256,18 +256,13 @@ ucp_proto_rndv_bulk_max_payload(ucp_request_t *req,
...
-                    total_length,
-                    ucp_proto_multi_scaled_length(lpriv->weight_sum, 100));
         /**
          * max_payload is later capped by remaining request length when
-         * advancing datatype iterator
+         * advancing datatype iterator, potential underflow is fine.
          */
         max_payload = ucs_max(min_rndv_chunk_size,
                               scaled_length - total_offset);
+        max_payload = ucs_min(max_payload, lpriv->max_frag);
```
brminich, line 265: afair it is cut later in `dt_iter_next`, so the decision in prev PR was to avoid extra branch here

tvegas1: Yes, we still don't check if `min_rndv_chunk_size` is bigger than the remaining length at this will be enforced in `dt_iter_next`. And `min_rndv_chunk_size` was assumed to be smaller than `lrpiv->max_frag` anyways.

But now we realize that it's actually possible that `max_payload` is a huge number due to underflow in the substraction line 264. Even if it is later cut, for bigger requests sizes, the number could still be bigger than `lpriv->max_frag`.

 Do you mean that `lpriv->max_frag` is later enforced on the obtained `max_payload`?

brminich (dismissed):

src/ucp/rndv/proto_rndv.inl
yosefe: if scaled_length < total_offset, line 264 would result in a negative value, how would that work given that all variables are unsigned types?

tvegas1: all values are unsigned with defined underflow, so if `scaled_length` < `total_offset`, we get huge value, capped at lane `max_frag` by line 265. Later, dt code path make sure we remain within request length too.

we could add a check if explicit behavior is more desirable.

yosefe: ok. but compared to the situation before https://github.com/openucx/ucx/commit/4909af673af7f2e3dcd2f98410213374f53d9fca, we have 2 more branches now. Any way we can optimize this?

tvegas1: tried to make it more explicit.

tvegas1: now that I reread, I wonder if we should instead go away from using `scaled_length` values (accumulated weights) to rather use individual weights. With current patch, I think we risk forcing using `lpriv->max_frag` for the lane after which `min_rndv_chunk_size` was used. Maybe the lengths considered are anyways small so we don't really care and the PR is good enough. wdyt?

tvegas1: finally fixed as discussed offline by pre-computing at multi init time.

src/ucp/proto/proto_multi.c
```diff
@@ -28,14 +30,14 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
...
-    double max_bandwidth, max_frag_ratio;
+    double max_bandwidth, max_frag_ratio, min_bandwidth;
     ucp_lane_index_t i, lane, num_lanes;
     ucp_proto_multi_lane_priv_t *lpriv;
     ucp_proto_perf_node_t *perf_node;
     size_t max_frag, min_length;
     ucp_lane_map_t lane_map;
     ucp_md_map_t reg_md_map;
-    uint32_t weight_sum;
+    uint32_t weight_sum, min_end_offset;
```
yosefe, line 40: min_end_offset should be size_t

src/ucp/proto/proto_multi.c
```diff
@@ -202,14 +209,17 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
...
-        mpriv->max_frag_sum += lpriv->max_frag;
-        lpriv->weight_sum    = weight_sum;
-        lpriv->max_frag_sum  = mpriv->max_frag_sum;
-        lpriv->opt_align     = ucp_proto_multi_get_lane_opt_align(params, lane);
-        mpriv->align_thresh  = ucs_max(mpriv->align_thresh,
-                                       lpriv->opt_align);
+        weight_sum           += lpriv->weight;
+        min_end_offset       += lane_perf->bandwidth *
+                                context->config.ext.min_rndv_chunk_size /
+                                min_bandwidth;
```
yosefe, line 215: minor: 
```c
        min_end_offset       += (lane_perf->bandwidth / min_bandwidth) *
                                context->config.ext.min_rndv_chunk_size;
```

src/ucp/rndv/proto_rndv.inl
```diff
@@ -243,31 +243,28 @@ ucp_proto_rndv_bulk_max_payload(ucp_request_t *req,
...
     if (ucs_likely(total_length < max_frag_sum)) {
         /**
          * Each lane sends less than its maximal fragment size but more than the
          * minimal chunk size
          */
-        min_rndv_chunk_size =
-                req->send.ep->worker->context->config.ext.min_rndv_chunk_size;
         scaled_length = ucp_proto_multi_scaled_length(lpriv->weight_sum,
                                                       total_length);
+        length        = ucs_max(scaled_length, lpriv->min_end_offset);
```
yosefe, line 255: rename length to "end_offset"
could also rename "scaled_length" to "weight_offset" or similar..

tvegas1: fixed

src/ucp/proto/proto_multi.c
tvegas1: fixed

src/ucp/proto/proto_multi.c
tvegas1: maybe nitpick, but i'd keep multiplications before divisions to try to minimize rounding errors, let me know else i can change it.

yosefe (approved):

brminich (approved):

yosefe:
@tvegas1 can you pls squash?

yosefe (approved):

PR #9713, opened 2024-02-27
UCP/PROTO: Add RNDV performance variants

ivankochin:
## What
Represent each possible RNDV proto that can be chosen after CTRL message as a separate variant.

## Why ?
That's the first step of MULTI performance estimation rework. The main reason of that patch is that we need to add CTRL msg perf to remote proto perf before selection the best one.

## How ?
The implementation part that worth noting is that remote variant `priv` which is required for `query` and `progress` calls is copied after the CTRL message priv buffer. `remote_proto_config::priv` cannot be filled during initialization, since both CTRL and remote variant `priv` buffers will be copied out during `ucp_proto_select_add_proto`.

There is another patch (https://github.com/openucx/ucx/pull/9728) that does the same change, but store all the initialization proto info inside `ucp_proto_select_elem_t` so it make it possible to assign variant config priv pointers during initialization. I created both of them to show difference between two approaches.

### DIFF
--- a/src/ucp/am/rndv.c
+++ b/src/ucp/am/rndv.c
@@ -59,21 +59,22 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_am_rndv_proto_progress, (self),
     return ucp_proto_am_handle_user_header_send_status(req, status);
 }
 
-ucs_status_t ucp_am_rndv_rts_init(const ucp_proto_init_params_t *init_params)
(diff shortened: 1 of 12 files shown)


ivankochin:
Closing this PR to avoid confusing during the review. Pls review https://github.com/openucx/ucx/pull/9728. First two commits of #9728 contains this PR changes.

PR #9714, opened 2024-02-27
UCT/IB/MLX5: Prevent compiler to replace SSE instructions by memmove()

tvegas1:
## What
Prevent compiler to replace SSE instructions with `memmove()`. This is continuation of #9692.

## Why ?
With `-O2` compiler can replace code by `memmove()`, even when SSE types are used.

## How ?
Use structure assignment like the non-SSE case to avoid `memmove()` optimization.

### Alternative
Patch below also works, but we want to avoid to selectively and globally disable optimizations:
```diff
diff --git a/src/uct/ib/Makefile.am b/src/uct/ib/Makefile.am
index 3974e3784..cad31a1bc 100644
--- a/src/uct/ib/Makefile.am
+++ b/src/uct/ib/Makefile.am
@@ -9,7 +9,7 @@ SUBDIRS = . rdmacm

 module_LTLIBRARIES    = libuct_ib.la
 libuct_ib_la_CPPFLAGS = $(BASE_CPPFLAGS) $(IBVERBS_CPPFLAGS)
-libuct_ib_la_CFLAGS   = $(BASE_CFLAGS)
+libuct_ib_la_CFLAGS   = $(BASE_CFLAGS) -fno-tree-loop-distribute-patterns
 libuct_ib_la_LIBADD   = $(top_builddir)/src/ucs/libucs.la \
                         $(top_builddir)/src/uct/libuct.la
 libuct_ib_la_LDFLAGS  = $(IBVERBS_LDFLAGS) -version-info $(SOVERSION)
```

### DIFF
--- a/src/uct/ib/mlx5/ib_mlx5.inl
+++ b/src/uct/ib/mlx5/ib_mlx5.inl
@@ -504,16 +504,20 @@ size_t uct_ib_mlx5_set_data_seg_iov(uct_ib_mlx5_txwq_t *txwq,
 static UCS_F_ALWAYS_INLINE void uct_ib_mlx5_bf_copy_bb(void * restrict dst,
                                                        void * restrict src)
 {
-#if defined( __SSE4_2__)
-    UCS_WORD_COPY(__m128i, dst, __m128i, src, MLX5_SEND_WQE_BB);
-#elif defined(__ARM_NEON)
+#if defined(__ARM_NEON)
     vst4q_u64(dst, vld4q_u64(src));
 #else
+#if defined(__SSE4_2__)
+    typedef __m128i uct_ib_mlx5_send_wqe_bb_block_t;
+#else
+    typedef uint8_t uct_ib_mlx5_send_wqe_bb_block_t;
+#endif
+    /* Prevent compiler to replace by memmove() */
     typedef struct {
-        uint8_t data[MLX5_SEND_WQE_BB];
+        uct_ib_mlx5_send_wqe_bb_block_t
+                data[MLX5_SEND_WQE_BB / sizeof(uct_ib_mlx5_send_wqe_bb_block_t)];
     } UCS_S_PACKED uct_ib_mlx5_send_wqe_bb_t;
 
-    /* Prevent the compiler to replace by memmove() */
     UCS_WORD_COPY(uct_ib_mlx5_send_wqe_bb_t, dst,
                   uct_ib_mlx5_send_wqe_bb_t, src,
                   MLX5_SEND_WQE_BB);


tvegas1:
@QuesarVII, thanks I could repro the issue. Could you please try if PR fixes all reproductions? It could be also interesting to independently confirm that the Makefile patch mentioned in Description also fixes all reproductions.

QuesarVII:
Yes, I can confirm that the PR resolves the issue.  I tested a few different MPI jobs of different sizes and quantities of communication and they are all working.

I can also confirm the no-tree-loop Makefile patch works. I tested with just that patch, without PR 9714 and 9692 being applied.

Thanks for resolving this!

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -505,7 +505,11 @@ static UCS_F_ALWAYS_INLINE void uct_ib_mlx5_bf_copy_bb(void * restrict dst,
                                                        void * restrict src)
 {
 #if defined( __SSE4_2__)
-    UCS_WORD_COPY(__m128i, dst, __m128i, src, MLX5_SEND_WQE_BB);
+    UCS_STATIC_ASSERT(MLX5_SEND_WQE_BB == 64);
+    _mm_storeu_si128((__m128i *)dst, *(__m128i *)src);
+    _mm_storeu_si128(((__m128i *)dst + 1), *((__m128i *)src + 1));
+    _mm_storeu_si128(((__m128i *)dst + 2), *((__m128i *)src + 2));
+    _mm_storeu_si128(((__m128i *)dst + 3), *((__m128i *)src + 3));
```
yosefe, line 512: maybe use same method as in lines 516-519?

tvegas1: fixed, makes sense to try to avoid explict instructions, retested `-O2`/`-O3` on submitter testbed. Double checked `sizeof(__m128i) == 16`.

yosefe (approved):

brminich (approved):

tvegas1:
@yosefe, I guess we need it in v1.16.x also, but no need to update NEWS file since rc4 is not yet out?

yosefe:
> @yosefe, I guess we need it in v1.16.x also, but no need to update NEWS file since rc4 is not yet out?

Pls port it to v1.16.x and update the NEWS as part of the PR

rakhmets:
> > @yosefe, I guess we need it in v1.16.x also, but no need to update NEWS file since rc4 is not yet out?
> 
> Pls port it to v1.16.x and update the NEWS as part of the PR

I think it is better to update NEWS once right before the publishing of RC. Otherwise, one more PR may be required just to update the date.

rakhmets (approved):

yosefe (approved):

PR #9715, opened 2024-02-27
TOOLS/PERF: do not override UCT device on server side

evgeny-leksikov:
## What
do not override UCT device on server side of ucx_perf_test

## Why ?
Client overrides all parameters on server side. In case if desired devices on client and server are different, the server will try to use wrong device.


src/tools/perf/perftest.c
```diff
@@ -449,6 +455,12 @@ static ucs_status_t setup_sock_rte_p2p(struct perftest_context *ctx)
             status = UCS_ERR_IO_ERROR;
             goto err_close_connfd;
         }
+
+        if (ctx->params.super.api == UCX_PERF_API_UCT) {
```
tvegas1, line 459: do we need to do such backup for the three supported RTE (mad/sock/mpi)?

evgeny-leksikov: I don't see if mpi does exchange of parameters, but it might be relevant for mad

evgeny-leksikov:
The PR was re-based to resolve conflicts... @tvegas1 could you pls take a look?

src/tools/perf/mad/perftest_mad.c
```diff
@@ -551,10 +551,9 @@ static ucs_status_t perftest_mad_accept(perftest_mad_rte_group_t *rte_group,
     lid = rte_group->dst_port.lid;
     ucs_debug("MAD: accept: remote lid:%d/0x%02x", lid, lid);
 
-    /* Import receive params, preserving passed flags */
-    needed_flags = ctx->params.super.flags & UCX_PERF_TEST_FLAG_ERR_HANDLING;
+    backup_local_params(&ctx->params, &params_backup);
```
tvegas1, line 554: Unfortunately this will break runtime as perftest_mad.c is bundled in libperftest_mad.so, and the loader will fail to resolve `backup_local_params()`/`restore_local_params()`. We are in the process of adding MAD-based runs in CI to catch those issues in the future.

Could you please move the added function in the corresponding header as static inlines?

src/tools/perf/perftest.c
```diff
@@ -332,6 +332,29 @@ static ucs_status_t setup_sock_rte_loopback(struct perftest_context *ctx)
     return UCS_OK;
 }
 
+void backup_local_params(const perftest_params_t *params,
+                         perftest_params_local_backup_t *backup)
```
tvegas1, line 336: maybe rename to `perftest_param_backup_local/restore()` and reuse `perftest_params_t` instead of adding `perftest_params_local_backup_t`?

src/tools/perf/perftest.c
```diff
@@ -332,6 +332,29 @@ static ucs_status_t setup_sock_rte_loopback(struct perftest_context *ctx)
...
+}
+
+void restore_local_params(perftest_params_t *params,
+                          const perftest_params_local_backup_t *backup)
+{
+    if (params->super.api == UCX_PERF_API_UCT) {
+        strcpy(params->super.uct.dev_name, backup->dev_name);
+    }
+
+    params->super.flags |= backup->needed_flags;
```
iyastreb, line 355: shouldn't we assign here instead of append?

src/tools/perf/perftest.h
```diff
@@ -53,4 +60,10 @@ void print_progress(void *UCS_V_UNUSED rte_group,
                     const ucx_perf_result_t *result, void *arg,
                     const char *extra_info, int final, int is_multi_thread);
 
+void backup_local_params(const perftest_params_t *params,
```
tvegas1, line 63: maybe add as static inline in `perftest_context.h`?

tvegas1 (commented):
Build mad support: `--with-mad`

src/tools/perf/perftest.c
evgeny-leksikov: According to existing logic, they should be or'ed. In fact, this is just sync of error handling enabling from client side to server.

src/tools/perf/perftest.c
evgeny-leksikov: I don't think that we need to backup entire `perftest_params_t`, reusing the structure may lead to wrong assumption that all field are valid after backup.

src/tools/perf/mad/perftest_mad.c
evgeny-leksikov: fixed

src/tools/perf/perftest.h
evgeny-leksikov: done

src/tools/perf/perftest.c
tvegas1: since functions are now in header, shall we add `perftest_params_*()` prefix to all of them?

src/tools/perf/mad/perftest_mad.c
```diff
@@ -551,10 +551,9 @@ static ucs_status_t perftest_mad_accept(perftest_mad_rte_group_t *rte_group,
     lid = rte_group->dst_port.lid;
     ucs_debug("MAD: accept: remote lid:%d/0x%02x", lid, lid);
 
-    /* Import receive params, preserving passed flags */
-    needed_flags = ctx->params.super.flags & UCX_PERF_TEST_FLAG_ERR_HANDLING;
+    perftest_params_local_backup(&ctx->params, &params_backup);
     memcpy(&ctx->params, buf, sizeof(ctx->params));
-    ctx->params.super.flags |= needed_flags;
+    perftest_params_local_restore(&ctx->params, params_backup);
```
tvegas1, line 556: i think we miss `&` in front of `params_backup`?

src/tools/perf/perftest_context.h
```diff
@@ -60,6 +60,39 @@ struct perftest_context {
...
+    backup->needed_flags = params->super.flags &
+                           UCX_PERF_TEST_FLAG_ERR_HANDLING;
+}
+
+
+static inline void
+perftest_params_local_restore(perftest_params_t *params,
+                              const perftest_params_local_backup_t *backup)
+{
+    if (params->super.api == UCX_PERF_API_UCT) {
```
tvegas1, line 88: for my understanding, are we sure this `UCX_PERF_API_UCT` value is not remove from receiving params content? would it make sense to also embed it in `backup` somehow?

evgeny-leksikov: I don't think that it's legal and robust to run server with API_UCT parameters but client with API_UCP ones, probably we should add such test and related error handling, but this is out of scope of this PR

tvegas1 (commented):
tested that mad rte works

src/tools/perf/perftest.c
```diff
@@ -866,9 +936,13 @@ int main(int argc, char **argv)
...
+    if (status != UCS_OK) {
+        goto out;
+    }
 
     /* Parse command line */
-    status = parse_opts(&ctx, mpi_initialized, argc, argv);
+    status = parse_opts(&ctx, argc, argv);
     if (status != UCS_OK) {
         ret = (status == UCS_ERR_CANCELED) ? 0 : -127;
         goto out;
```
evgeny-leksikov, line 948:
```suggestion
        goto out_ctx_free;
```

tvegas1 (commented):
seems `mpirun -np 2  ./rfs/bin/ucx_perftest -t tag_lat -n 100` does not work anymore with PR.

built on jazz: `git clean -xdfxdf && ./autogen.sh && ./contrib/configure-release --prefix=$(pwd)/rfs --with-mpi --with-mad  --with-xpmem && make -j && make install`

src/tools/perf/perftest.c
```diff
@@ -213,6 +214,52 @@ ucs_status_t init_test_params(perftest_params_t *params)
...
+    ucx_perf_rte_t *local_rte = ctx->params.super.rte;
+    unsigned needed_flags;
+    char local_dev_name[UCT_DEVICE_NAME_MAX];
+    ucs_status_t status;
+
+    if ((ctx->server_addr != NULL) || ctx->mpi) {
+        return UCS_OK;
+    }
+
+    if (ctx->params.super.api != ctx->peer_params.super.api) {
```
tvegas1, line 236: seems now mad and socket fail because `ctx->params.super.api` is `UCX_PERF_API_LAST`

tvegas1 (approved):
lgtm, but why not continuing with previous approach?

evgeny-leksikov:
> lgtm, but why not continuing with previous approach?

 - this one is simpler
 - previous approach requires to handle every rte implementation outside of rte call, that's a bug prone

evgeny-leksikov:
@yosefe could you please review?

src/tools/perf/mad/perftest_mad.c
```diff
@@ -551,10 +551,9 @@ static ucs_status_t perftest_mad_accept(perftest_mad_rte_group_t *rte_group,
     lid = rte_group->dst_port.lid;
     ucs_debug("MAD: accept: remote lid:%d/0x%02x", lid, lid);
 
-    /* Import receive params, preserving passed flags */
-    needed_flags = ctx->params.super.flags & UCX_PERF_TEST_FLAG_ERR_HANDLING;
+    perftest_params_local_backup(&ctx->params, &params_backup);
     memcpy(&ctx->params, buf, sizeof(ctx->params));
-    ctx->params.super.flags |= needed_flags;
+    perftest_params_local_restore(&ctx->params, &params_backup);
```
yosefe, line 556: maybe instead of backup/restore there will be a single pertest_params_copy() function. that function will backup, (to local vars), do memcpy, and restore from local vars - all internally. I think it will simplify the usage.
when receiving from socket, receive into a local struct and then use the new function to copy into ctx->params.

evgeny-leksikov: if I understand correctly - do you propose to implement different behavior for mad and socket rte?
> receive into a local struct and then use the new function to copy into `ctx->params`

Just `memcpy` won't work because some parameters should be overwritten, for example `msg_size_list` (which furthermore allocated separately) but other one like device name should be saved from initial parameters and third type is `flags` which should be combined (OR'ed). IMO it does not look simpler.

src/tools/perf/perftest_context.h
```diff
@@ -60,13 +60,55 @@ struct perftest_context {
...
+    char uct_dev_name[UCT_DEVICE_NAME_MAX];
+    unsigned needed_flags;
+
+    if (dst->super.api != src->super.api) {
+        return UCS_ERR_INVALID_PARAM;
+    }
+
+    /* backup required dst parameters */
+    if (dst->super.api == UCX_PERF_API_UCT) {
+        strcpy(uct_dev_name, dst->super.uct.dev_name);
```
yosefe, line 83: ucs_strncpy_safe or memcpy

src/tools/perf/perftest_context.h
```diff
@@ -60,13 +60,55 @@ struct perftest_context {
...
+    }
+
+    needed_flags = (dst->super.flags & UCX_PERF_TEST_FLAG_ERR_HANDLING);
+
+    /* copy src to dst */
```
yosefe, line 88: remove comment

src/tools/perf/perftest_context.h
```diff
@@ -60,13 +60,55 @@ struct perftest_context {
...
+    perftest_params_release_msg_size_list(dst);
+    memcpy(dst, src, sizeof(*src));
```
yosefe, line 90: space line before and after

tvegas1 (approved):

yosefe (dismissed):

tvegas1 (approved):

src/tools/perf/perftest.c
```diff
@@ -169,7 +169,11 @@ ucs_status_t init_test_params(perftest_params_t *params)
     static const struct sockaddr_storage empty_addr = {};
 
     memset(params, 0, sizeof(*params));
-    params->super.api               = UCX_PERF_API_LAST;
+
+    /* UCT tests require specified device (-d) on both sides - client and server,
```
yosefe, line 173: the api should be detected by test code, maybe this is what the merge function should check?

src/tools/perf/perftest_context.h
```diff
@@ -76,6 +76,7 @@ perftest_params_merge(perftest_params_t *dst, const perftest_params_t *src)
     unsigned needed_flags;
 
     if (dst->super.api != src->super.api) {
+        ucs_error("test parameters merge: source and destination API mismatch");
```
yosefe, line 79: print both values of apis that don't match

src/tools/perf/perftest_params.c
```diff
@@ -327,6 +327,7 @@ ucs_status_t parse_test_params(perftest_params_t *params, char opt,
     case 'd':
         ucs_snprintf_zero(params->super.uct.dev_name,
                           sizeof(params->super.uct.dev_name), "%s", opt_arg);
+        params->super.api = UCX_PERF_API_UCT;
```
yosefe, line 330: it should be decided by test name

src/tools/perf/perftest_context.h
```diff
@@ -69,15 +70,36 @@ perftest_params_release_msg_size_list(perftest_params_t *params)
...
 
+static inline ucs_status_t
+perftest_params_check_api(ucx_perf_api_t dst_api, ucx_perf_api_t src_api)
+{
+    if (src_api == UCX_PERF_API_LAST) {
+        ucs_error(PARAMS_MERGE_PREFIX": source API must be defined");
+        return UCS_ERR_INVALID_PARAM;
+    }
+
+    if ((dst_api != src_api) ||
```
tvegas1, line 81: i might be confused but if server options are only `-K mlx5_0:1`, dst_api is UCX_PERF_API_LAST, but we still fail? If yes I think the check should rather be:
```
if ((dst_api != UCX_PERF_API_LAST) && (dst_api != src_api)) {
```

src/tools/perf/perftest_context.h
```diff
@@ -104,10 +97,7 @@ perftest_params_merge(perftest_params_t *dst, const perftest_params_t *src)
     }
 
     /* restore required overwritten parameters */
-    if (dst->super.api == UCX_PERF_API_UCT) {
-        strcpy(dst->super.uct.dev_name, uct_dev_name);
-    }
-
+    strcpy(dst->super.uct.dev_name, uct_dev_name);
```
yosefe, line 100: ucs_strncpy_safe

src/tools/perf/perftest_context.h
```diff
@@ -9,11 +9,13 @@
 
 #include "api/libperf.h"
 #include "lib/libperf_int.h"
+#include <ucs/sys/string.h>
 
 
 #define MAX_BATCH_FILES         32
 #define MAX_CPUS                1024
 
+#define PARAMS_MERGE_PREFIX    "test parameters merge"
```
yosefe, line 18: where is it used?

tvegas1 (approved):

yosefe (approved):

PR #9720, opened 2024-02-29
UCP: Fix rndv ppln protocols selection - v1.16.x

brminich:
## What
Do not select pipeline rndv protocols when sender buffer is not contig (in this case address is zero in RTS header)

### DIFF
--- a/src/ucp/rndv/rndv_ppln.c
+++ b/src/ucp/rndv/rndv_ppln.c
@@ -59,7 +59,8 @@ ucp_proto_rndv_ppln_init(const ucp_proto_init_params_t *init_params)
     if ((select_param->dt_class != UCP_DATATYPE_CONTIG) ||
         !ucp_proto_init_check_op(init_params, UCP_PROTO_RNDV_OP_ID_MASK) ||
         !ucp_proto_common_init_check_err_handling(&err_params) ||
-        ucp_proto_rndv_init_params_is_ppln_frag(init_params)) {
+        ucp_proto_rndv_init_params_is_ppln_frag(init_params) ||
+        (init_params->rkey_cfg_index == UCP_WORKER_CFG_INDEX_NULL)) {
         return UCS_ERR_UNSUPPORTED;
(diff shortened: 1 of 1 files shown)


src/ucp/rndv/proto_rndv.c
```diff
@@ -707,7 +708,7 @@ ucp_proto_rndv_send_reply(ucp_worker_h worker, ucp_request_t *req,
         rkey           = NULL;
     }
 
-    ucp_proto_select_param_init(&sel_param, op_id, op_attr_mask, 0,
+    ucp_proto_select_param_init(&sel_param, op_id, op_attr_mask, op_flags,
```
rakhmets, line 711: Should we do the same fix in master branch?

rakhmets (dismissed):

src/ucp/rndv/proto_rndv.c
brminich: yes, together with relevant gtest

src/ucp/proto/proto_select.h
```diff
@@ -28,6 +28,10 @@
  * Relevant for UCP_OP_ID_RNDV_SEND and UCP_OP_ID_RNDV_RECV. */
 #define UCP_PROTO_SELECT_OP_FLAG_PPLN_FRAG (UCP_PROTO_SELECT_OP_FLAGS_BASE << 0)
 
+/* Restrict selection of rendezvous pipeline protocols (e.g. when sender buffer
+ * is non-contig). */
+#define UCP_PROTO_SELECT_OP_FLAG_NO_PPLN   (UCP_PROTO_SELECT_OP_FLAGS_BASE << 1)
```
yosefe, line 33: why need a special flag? can't rndv/ppln disqualify itself if the buffer is not contig according to select_param dt_class?

brminich: fixed in place

rakhmets (approved):

PR #9723, opened 2024-03-01
UCT/IB/MLX5: Prevent compiler to replace SSE instructions by memmove() - v1.16.x

tvegas1:
## What
Backport #9714.

### DIFF
--- a/src/uct/ib/mlx5/ib_mlx5.inl
+++ b/src/uct/ib/mlx5/ib_mlx5.inl
@@ -504,16 +504,20 @@ size_t uct_ib_mlx5_set_data_seg_iov(uct_ib_mlx5_txwq_t *txwq,
 static UCS_F_ALWAYS_INLINE void uct_ib_mlx5_bf_copy_bb(void * restrict dst,
                                                        void * restrict src)
 {
-#if defined( __SSE4_2__)
-    UCS_WORD_COPY(__m128i, dst, __m128i, src, MLX5_SEND_WQE_BB);
-#elif defined(__ARM_NEON)
+#if defined(__ARM_NEON)
     UCS_WORD_COPY(int16x8_t, dst, int16x8_t, src, MLX5_SEND_WQE_BB);
 #else
+#if defined(__SSE4_2__)
+    typedef __m128i uct_ib_mlx5_send_wqe_bb_block_t;
+#else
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -11,6 +11,11 @@
 ### Features:
 ### Bugfixes:
 
+## 1.16.0-rc5 (March 1, 2024)
+### Bugfixes:
+#### UCT
+* Fixed Infiniband posting error when compiler replaces SSE code with memmove()
+
```
rakhmets, line 18: We haven't released RC4 yet.
This changes can be added to RC4. And it is better to make a single PR for changes in NEWS right before the creation of an RC. @shasson5 will update the NEWS for RC4 and publish the release.

tvegas1: removed NEWS changes, ask was from @yosefe for combined PR, I guess we can do as you suggest.

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -504,16 +504,20 @@ size_t uct_ib_mlx5_set_data_seg_iov(uct_ib_mlx5_txwq_t *txwq,
 static UCS_F_ALWAYS_INLINE void uct_ib_mlx5_bf_copy_bb(void * restrict dst,
                                                        void * restrict src)
 {
-#if defined( __SSE4_2__)
-    UCS_WORD_COPY(__m128i, dst, __m128i, src, MLX5_SEND_WQE_BB);
-#elif defined(__ARM_NEON)
+#if defined(__ARM_NEON)
     UCS_WORD_COPY(int16x8_t, dst, int16x8_t, src, MLX5_SEND_WQE_BB);
```
rakhmets, line 508: `vst4q_u64(dst, vld4q_u64(src));` ?

tvegas1: if we need it in v1.16.x i'd rather backport it as a separate commit, @yosefe I backport that one? it was not coming from a bug report unless i'm mistaken.

rakhmets: I see... that was in a different PR. Yes, agree that this doesn't belong to this PR.

rakhmets (approved):

yosefe (approved):

yosefe:
@tvegas1 can you pls squash?

PR #9728, opened 2024-03-05
UCP/PROTO: Add RNDV performance variants - extend selection hash map

ivankochin:
## What
This PR is the copy of #9713 with storing proto initialization info inside the selection hash map.

## Why?
That allows get rid of of copying variant priv after the protocol priv and completely fill variant proto config (including priv) during the initialization. That also reduces the number of changes and avoids explicit const removing from priv of PPLN protos.


src/ucp/rndv/proto_rndv.c
```diff
@@ -363,28 +346,143 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+    if (status != UCS_OK) {
+        goto out;
+    }
+    remote_proto_init = &remote_proto_select->proto_init;
+
+    /* Add variants for each remote proto */
+    ucs_array_for_each(remote_proto, &remote_proto_init->protocols) {
+        remote_caps = &remote_proto->caps;
+        /* Skip empty variants and reconfig proto */
+        if ((remote_caps->num_ranges == 0) || (remote_proto->proto_id == 0)) {
```
brminich, line 421: can reconfig be the only option?

src/ucp/rndv/proto_rndv.c
```diff
@@ -363,28 +346,143 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+            continue;
+        }
+
+        last_range = remote_caps->ranges + (remote_caps->num_ranges - 1);
+        max_length = ucs_min(params->super.max_length, last_range->max_length);
+        min_length = ucs_max(params->super.min_length,
+                              remote_proto->caps.min_length);
```
brminich, line 428:
```suggestion
        min_length = ucs_max(params->super.min_length,
                             remote_proto->caps.min_length);
```

src/ucp/rndv/proto_rndv.c
```diff
@@ -363,28 +346,143 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+             * params->super.cfg_thresh and remote_proto->cfg_thresh.
+             *    E.g. RNDV_THRESH=0 and RNDV_MODE=get_zcopy
+             *    In this case all the variants would have 0 cfg_thresh, but
+             *    get_zcopy should be preferred over other RNDV protocols.
+             * 
+             * +1 is required in case if remote variant cfg_priority is 0.
+             */
+            if (remote_proto->cfg_thresh != UCS_MEMUNITS_INF) {
+                cfg_priority += remote_proto->cfg_priority + 1;
+            }
```
brminich, line 457: can we do that using thresh? If we take max of thresh values, all rndv variants will be `inf` except the one which is selected by `UCX_RNDV_SCHEME`

src/ucp/rndv/proto_rndv.c
```diff
@@ -363,28 +346,143 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+        }
+
+        ucp_proto_common_init_base_caps(&params->super, min_length);
+
+        status = ucp_proto_rndv_ctrl_add_remote_variant(params, &ctrl_perf,
+                                                        remote_caps->ranges,
+                                                        min_length, max_length);
+        ucp_proto_perf_node_deref(&ctrl_perf.node);
+        if (status != UCS_OK) {
+            break;
```
brminich, line 473: why not continue?

src/ucp/rndv/rndv_ats.c
```diff
@@ -24,7 +24,7 @@ ucp_proto_rndv_ats_init(const ucp_proto_init_params_t *init_params)
 
     *init_params->priv_size = sizeof(ucp_proto_rndv_ack_priv_t);
 
-    caps.cfg_thresh   = 0;
+    caps.cfg_thresh   = UCS_MEMUNITS_AUTO;
```
brminich, line 27: why?

src/ucp/rndv/rndv_ppln.c
```diff
@@ -74,55 +76,79 @@ ucp_proto_rndv_ppln_init(const ucp_proto_init_params_t *init_params)
...
+        ucp_proto_select_add_proto(init_params, proto->cfg_thresh,
+                                   proto->cfg_priority,
+                                   init_params->caps, init_params->priv,
+                                   sizeof(*rpriv));
+    }
 
-    return status;
+out:
+    if (status != UCS_OK) {
+        ucs_debug("error during rndv ppln probing, status \"%s\"",
```
brminich, line 149: would not it be always unsupported? i'd remove this print

src/ucp/rndv/rndv_ppln.c
```diff
@@ -247,8 +274,16 @@ static ucs_status_t ucp_proto_rndv_ppln_progress(uct_pending_req_t *uct_req)
             return UCS_OK;
         }
 
+        /* Avoid sending message less than `frag_proto_min_length` */
+        next_frag = rpriv->frag_size;
+        msg_left  = req->send.state.dt_iter.length - req->send.state.dt_iter.offset;
+        if ((msg_left > rpriv->frag_size) &&
+            (msg_left < rpriv->frag_size + rpriv->frag_proto_min_length)) {
+            next_frag = rpriv->frag_proto_min_length;
+        }
```
brminich, line 283: is it an unrelated fix?

src/ucp/rndv/proto_rndv.c
ivankochin: In that case there are no variant added so the protocol would be unavailable.

src/ucp/rndv/proto_rndv.c
ivankochin: No we cannot, imagine the case when `UCX_RNDV_THRESH=0` and `UCX_RNDV_SCHEME=get_zcopy` but `rndv/get/zcopy` is unavailable. If we use only `cfg_thresh` all the RNDV protocols in that case would have `inf` threshold and so the eager would be used. 

I understand that this is a kind of a corner case, but I don't think that changing current behavior is a good idea.

src/ucp/rndv/proto_rndv.c
ivankochin: As I know, fail inside `add_parallel_ranges()` can be caused only by absense of memory for allocation, so I put break here since further protos will return error in that case.

I can change it to `continue`, if it seems more logically correct from your point of view.

src/ucp/rndv/rndv_ats.c
ivankochin: With this patch, RNDV variants are compared together with eager one, so setting 0 here would force empty ATS over short which is not good decision for performance.

src/ucp/rndv/rndv_ppln.c
ivankochin: It also can be error code from `ucp_proto_rndv_ack_init`

src/ucp/rndv/rndv_ppln.c
ivankochin: No, it is related, let me explain the case. 

In the master we have `ucp_proto_select_elem_t` which contains protocols for every message size. Now we have only one proto which has it's own `min_length`.

Lets imagine that we are sending 300 bytes using RNDV PPLN proto and the frag size is 130 bytes. Then if we use master logic, we send two 130 bytes fragments using RNDV-get and one 40 bytes remainder (300 - 130*2) using RNDV-put since RNDV-get cannot send such small chunks (min_length of RNDV-get is 129).

In the new approach we don't have ability to choose different proto for small remainders. So I solved this problem by sending min_length first, so the next fragment would leave a remainder at all. 

It also can be done by increasing the current fragment, but we need to care about max supported message length for current proto in that case.

src/ucp/rndv/proto_rndv.c
```diff
@@ -265,58 +250,45 @@ ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+    rpriv->lane             = lane;
+    rpriv->packed_rkey_size = ucp_rkey_packed_size(
+            init_params->worker->context, rpriv->md_map,
+            init_params->select_param->sys_dev, rpriv->sys_dev_map);
+
+    /* Use only memory domains for which the unpacking of the remote key was
+     * successful */
+    if (init_params->rkey_config_key != NULL) {
+        rpriv->md_map &= ~init_params->rkey_config_key->unreachable_md_map;
+    }
```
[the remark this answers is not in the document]
ivankochin: Good catch, thanks! Done.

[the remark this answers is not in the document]
ivankochin: You are right, but I would like to leave at here for at least some time.

src/ucp/proto/proto.h
```diff
@@ -61,11 +61,9 @@ typedef struct ucp_proto_perf_node ucp_proto_perf_node_t;
 /* Key for selecting a protocol */
 typedef struct ucp_proto_select_param ucp_proto_select_param_t;
 
-
 /* Context for probing a protocol for a specific selection key */
 typedef struct ucp_proto_probe_ctx ucp_proto_probe_ctx_t;
 
-
```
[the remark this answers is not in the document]
ivankochin: Done.

src/ucp/rndv/proto_rndv.c
ivankochin: Done.

src/ucp/rndv/proto_rndv.c
ivankochin: Created separate a patch to simplify this logic: https://github.com/openucx/ucx/pull/9815

ivankochin: The patch that was supposed to simplify that logic cannot be merged since many tests rely on that logic and some of them cannot be workarounded (see the PR mentioner above).

brminich: is there any other drawback? Using thresholds still lokks better for me

test/gtest/ucp/test_ucp_tag.cc
```diff
@@ -503,26 +503,48 @@ class test_ucp_tag_limits : public test_ucp_tag {
...
+                                                   cfg_thresh);
+
             for (; msg_length > value.thresholds[idx].max_msg_length; ++idx) {
                 proto_config = &value.thresholds[idx].proto_config;
+                /* Skip check that rndv is disabled for tag offload use case.
+                 * With tag offload rndv protocol might still be used for large
+                 * messages, because of HWTM limit for eager transfers.
+                 * e.g. RNDV_THRESH=inf + HWTM */
+                if((m_test_offload) &&
+                   (value.thresholds[idx].max_msg_length >= m_tag_min_rndv)) {
```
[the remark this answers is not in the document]
ivankochin: This check is true but too relaxed, here should be something like:
```c
    (value.thresholds[idx].max_msg_length >= sender().iface_attr().cap.tag.eager.max_zcopy)) {
```

But there is no simple way to get this `max_zcopy` value on UCP layer. So it was decided to completely disable this test in case when HWTM is enabled.

src/ucp/rndv/proto_rndv.c
ivankochin: It was decided to try setting `cfg_priority` to the current protocol value for all variants, and initialize `cfg_thresh` as  `max(protocol_cfg_thresh, variant_cfg_thresh)`. Need to check whether it would cause many workarounds in gtest logic.

src/ucp/rndv/proto_rndv.c
```diff
@@ -263,98 +250,94 @@ ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+}
 
-    if (!ucp_proto_select_get_valid_range(priv->remote_proto.thresholds,
-                                          &min_length, &max_length)) {
-        status = UCS_ERR_UNSUPPORTED;
-        goto out;
-    }
+static ucs_status_t
+ucp_proto_rndv_ctrl_range_init(const ucp_proto_rndv_ctrl_init_params_t *params,
+                               ucp_proto_rndv_ctrl_priv_t *priv,
```
yosefe, line 283: can we make int const*?

src/ucp/rndv/proto_rndv.c
```diff
@@ -365,41 +348,136 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
-    status = UCS_OK;
+    return UCS_OK;
+}
 
-out_deref_perf_node:
-    ucp_proto_perf_node_deref(&ctrl_perf.node);
-out:
-    return status;
+void
+ucp_proto_rndv_set_variant_config(const ucp_proto_init_params_t *init_params,
```
yosefe, line 370: why do we need to initialize a ucp_proto_config_t struct?

src/ucp/rndv/rndv_ppln.c
```diff
@@ -245,8 +255,16 @@ static ucs_status_t ucp_proto_rndv_ppln_progress(uct_pending_req_t *uct_req)
             return UCS_OK;
         }
 
+        /* Avoid sending message less than `frag_proto_min_length` */
```
yosefe, line 258: why is it needed?

src/ucp/rndv/proto_rndv.c
```diff
@@ -365,41 +348,136 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+                                  const ucp_proto_init_elem_t *proto,
+                                  const ucp_proto_select_param_t *select_param,
+                                  void *priv, ucp_proto_config_t *cfg)
+{
+    cfg->priv           = priv;
+    cfg->cfg_thresh     = proto->cfg_thresh;
```
yosefe, line 376: can we remove cfg_thresh from ucp_proto_config_t?

src/ucp/rndv/proto_rndv.c
ivankochin: For `query()` interface calls in case of RTS remote proto initialization and for both `query()` and `progress()` in case of RNDV PPLN underlying proto initialization.

src/ucp/rndv/rndv_ppln.c
ivankochin: See explanation here: https://github.com/openucx/ucx/pull/9728#discussion_r1522764256

We can discuss it in more details if needed.

src/ucp/rndv/proto_rndv.c
ivankochin: Done.

src/ucp/rndv/rndv_ppln.c
```diff
@@ -247,8 +274,16 @@ static ucs_status_t ucp_proto_rndv_ppln_progress(uct_pending_req_t *uct_req)
...
+        next_frag = rpriv->frag_size;
+        msg_left  = req->send.state.dt_iter.length - req->send.state.dt_iter.offset;
+        if ((msg_left > rpriv->frag_size) &&
+            (msg_left < rpriv->frag_size + rpriv->frag_proto_min_length)) {
```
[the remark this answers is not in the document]
ivankochin: Done.

src/ucp/proto/proto_select.h
```diff
@@ -111,13 +131,13 @@ typedef struct {
  */
 typedef struct {
     /* Array of which protocol to use for different message sizes */
-    const ucp_proto_threshold_elem_t *thresholds;
+    const ucp_proto_threshold_elem_t  *thresholds;
 
     /* Estimated performance for the selected protocols */
-    ucp_proto_perf_range_t           *perf_ranges;
+    ucp_proto_perf_range_t            *perf_ranges;
```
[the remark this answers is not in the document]
ivankochin: Yes, that seems to be right, as I see the only place where `perf_ranges` are used is `ucp_proto_perf_range_search` but this search can be done using `select_elem->thresholds`. Do you want me to remove it in that patch or in separate one?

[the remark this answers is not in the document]
yosefe: yes, in a separate pr, after changing proto_caps to perf

src/ucp/rndv/rndv_ppln.c
```diff
@@ -84,43 +85,51 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
...
-                                         frag_overhead / frag_max_length);
-    status = ucp_proto_rndv_ack_init(init_params, UCP_PROTO_RNDV_ATS_NAME,
-                                     &ppln_caps, ppln_overhead, &rpriv.ack,
-                                     &caps);
-    if (status == UCS_OK) {
-        ucp_proto_select_add_proto(init_params,
-                                   thresh_elem->proto_config.cfg_thresh, 0,
-                                   &caps, &rpriv, sizeof(rpriv));
-    }
+        frag_range = proto->caps.ranges + (proto->caps.num_ranges - 1);
```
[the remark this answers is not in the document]
ivankochin: `ucp_proto_perf_range_search` is supposed to use in select_elem and the current logic just gets the last range from caps. We cannot search through `select_elem` since it contains envelope and we need to handle only one certain variant here.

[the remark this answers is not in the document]
yosefe: ok

src/ucp/rndv/proto_rndv.c
```diff
@@ -365,41 +348,122 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+                             remote_proto->caps.min_length);
+        if (max_length < min_length) {
+            continue;
+        }
+
+        remote_priv = &ucs_array_elem(&remote_proto_init->priv_buf,
+                                      remote_proto->priv_offset);
+        ucp_proto_rndv_set_variant_config(init_params, remote_proto,
+                                          &remote_select_param, remote_priv,
+                                          &rpriv->remote_proto_config);
```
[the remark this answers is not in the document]
ivankochin: That's not seem to be possible since we also need to obtain pointer to `priv` buffer, `ep_cfg_index` and `remote_select_param` to call remote protocol query during RTS/RTR query call.

We can store these variables separately, but from my point of view it's simpler and more straightforward to store the whole `ucp_proto_config_t` structure.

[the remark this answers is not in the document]
yosefe: ok

src/ucp/rndv/rndv_ppln.c
```diff
@@ -245,8 +255,16 @@ static ucs_status_t ucp_proto_rndv_ppln_progress(uct_pending_req_t *uct_req)
             return UCS_OK;
         }
 
+        /* Avoid sending message less than `frag_proto_min_length` */
+        next_frag = rpriv->frag_size;
+        msg_left  = req->send.state.dt_iter.length - req->send.state.dt_iter.offset;
+        if ((msg_left > rpriv->frag_size) &&
+            (msg_left < (rpriv->frag_size + rpriv->frag_proto_min_length))) {
+            next_frag = rpriv->frag_proto_min_length;
+        }
```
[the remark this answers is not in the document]
ivankochin: Done.

src/ucp/rndv/rndv_ppln.c
```diff
@@ -84,43 +85,51 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
...
 
-    /* Find the performance range of sending one fragment */
-    if (!ucp_proto_select_get_valid_range(select_elem->thresholds,
-                                          &frag_min_length, &frag_max_length)) {
-        return;
-    }
+    /* Add each proto as a separate variant */
+    ucs_array_for_each(proto, &select_elem->proto_init.protocols) {
+        /* Skip empty variants and reconfig proto */
+        if ((proto->caps.num_ranges == 0) || (proto->proto_id == 0)) {
```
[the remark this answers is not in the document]
ivankochin: I added assert for `proto->caps.num_ranges > 0` but left reconfig proto skip since this proto has infinite latency. If reconfig proto passed forward as other protocols do, then some arithmetic operations will happen with infinite latency and that causes weird failures.

src/ucp/rndv/proto_rndv.c
```diff
@@ -365,41 +348,122 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+ucp_proto_rndv_set_variant_config(const ucp_proto_init_params_t *init_params,
+                                  const ucp_proto_init_elem_t *proto,
+                                  const ucp_proto_select_param_t *select_param,
+                                  void *priv, ucp_proto_config_t *cfg)
+{
+    cfg->priv           = priv;
+    cfg->ep_cfg_index   = init_params->ep_cfg_index;
+    cfg->proto          = ucp_protocols[proto->proto_id];
+    cfg->rkey_cfg_index = init_params->rkey_cfg_index;
+    cfg->select_param   = *select_param;
```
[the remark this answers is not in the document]
ivankochin: Done.

src/ucp/rndv/proto_rndv.c
```diff
@@ -415,18 +415,32 @@ void ucp_proto_rndv_ctrl_probe(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+    ucs_trace("add variants for %s: min_length=%zu max_length=%zu",
+              ucp_proto_id_field(init_params->proto_id, name),
+              params->super.min_length, params->super.max_length);
+    ucs_log_indent(+1);
     ucs_array_for_each(remote_proto, &remote_proto_init->protocols) {
-        remote_caps = &remote_proto->caps;
-        /* Skip empty variants and reconfig proto */
-        if ((remote_caps->num_ranges == 0) || (remote_proto->proto_id == 0)) {
+        /* Skip reconfig proto */
+        if (remote_proto->proto_id == 0) {
```
yosefe, line 424: can it be a better check than comparing proto_id to 0? for example, UCP_PROTO_FLAG_INVALID?

src/ucp/rndv/proto_rndv.c
```diff
@@ -415,18 +415,32 @@ void ucp_proto_rndv_ctrl_probe(const ucp_proto_rndv_ctrl_init_params_t *params,
...
             continue;
         }
-
-        last_range = remote_caps->ranges + (remote_caps->num_ranges - 1);
-        max_length = ucs_min(params->super.max_length, last_range->max_length);
-        min_length = ucs_max(params->super.min_length,
-                             remote_proto->caps.min_length);
+        ucs_assertv(remote_proto->caps.num_ranges > 0,
```
yosefe, line 427: space line before

src/ucp/rndv/proto_rndv.c
```diff
@@ -439,31 +453,43 @@ void ucp_proto_rndv_ctrl_probe(const ucp_proto_rndv_ctrl_init_params_t *params,
...
         }
 
-        cfg_thresh = params->super.cfg_thresh;
-        if (cfg_thresh == UCS_MEMUNITS_AUTO) {
-            cfg_thresh = remote_proto->cfg_thresh;
-        } else if (remote_proto->cfg_thresh != UCS_MEMUNITS_AUTO) {
-            cfg_thresh = ucs_max(cfg_thresh, remote_proto->cfg_thresh);
+        cfg_thresh   = params->super.cfg_thresh;
+        cfg_priority = params->super.cfg_priority;
+        if (remote_proto->cfg_thresh != UCS_MEMUNITS_INF &&
```
yosefe, line 474: add ( )

src/ucp/rndv/proto_rndv.c
```diff
@@ -439,31 +453,43 @@ void ucp_proto_rndv_ctrl_probe(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+            remote_proto->cfg_thresh != UCS_MEMUNITS_AUTO) {
+            cfg_thresh       = ucs_max(cfg_thresh, remote_proto->cfg_thresh);
+            /* Set priority to remote proto only if RNDV_SCHEME or
+             * BCOPY/ZCOPY thresh is set to force these settings */
+            if (remote_proto->cfg_thresh != UCS_MEMUNITS_AUTO) {
```
yosefe, line 479: why check it again in addition to line 475?

src/ucp/rndv/rndv_ppln.c
```diff
@@ -87,10 +87,12 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
 
     /* Add each proto as a separate variant */
     ucs_array_for_each(proto, &select_elem->proto_init.protocols) {
-        /* Skip empty variants and reconfig proto */
-        if ((proto->caps.num_ranges == 0) || (proto->proto_id == 0)) {
+        /* Skip reconfig proto */
+        if (proto->proto_id == 0) {
```
yosefe, line 91: check proto flag INVALID

src/ucp/rndv/rndv_ppln.c
```diff
@@ -255,16 +256,14 @@ static ucs_status_t ucp_proto_rndv_ppln_progress(uct_pending_req_t *uct_req)
...
-        /* Avoid sending message less than `frag_proto_min_length` */
-        next_frag = rpriv->frag_size;
-        msg_left  = req->send.state.dt_iter.length - req->send.state.dt_iter.offset;
-        if ((msg_left > rpriv->frag_size) &&
-            (msg_left < (rpriv->frag_size + rpriv->frag_proto_min_length))) {
-            next_frag = rpriv->frag_proto_min_length;
-        }
+        /* Avoid sending message less than `frag_proto_min_length` by
+           overlapping previous fragment */
+        req->send.state.dt_iter.offset = ucs_min(
```
yosefe, line 261: maybe encapsulate it in a new datatype_iter function

yosefe: yes, maybe add a variant of ucp_datatype_iter_next_slice that also accepts min_frag and returns overlap value (which is 0 if no overlap or not last frag)

ivankochin: Done

src/ucp/rndv/proto_rndv.c
```diff
@@ -365,41 +348,147 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+              ucp_proto_id_field(init_params->proto_id, name),
+              params->super.min_length, params->super.max_length);
+    ucs_log_indent(+1);
+    ucs_array_for_each(remote_proto, &remote_proto_init->protocols) {
+        if (ucp_proto_id_field(remote_proto->proto_id, flags) &
+            UCP_PROTO_FLAG_INVALID) {
+            continue;
+        }
+
+        ucs_assertv(remote_proto->caps.num_ranges > 0,
```
brminich, line 428: no need for assertV, if it fails, the value is `0`

src/ucp/rndv/proto_rndv.c
```diff
@@ -365,41 +348,147 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+                    remote_proto->caps.num_ranges);
+
+        proto_name = ucp_proto_id_field(remote_proto->proto_id, name);
+        caps       = &remote_proto->caps;
+        max_length = caps->ranges[caps->num_ranges - 1].max_length;
+        min_length = caps->min_length;
+        ucs_trace("%s proto variant: min_length=%zu max_length=%zu "
+                  "cfg_thresh=%zu cfg_priority=%u", proto_name, 
+                  min_length, max_length, remote_proto->cfg_thresh,
+                  remote_proto->cfg_priority);
```
brminich, line 439: maybe move it below after checking `max_len > min_len`

src/ucp/rndv/proto_rndv.c
```diff
@@ -365,41 +348,147 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+        if ((remote_proto->cfg_thresh != UCS_MEMUNITS_INF) &&
+            (remote_proto->cfg_thresh != UCS_MEMUNITS_AUTO)) {
+            /* Consider remote priority and threshold only if RNDV_SCHEME or
+             * BCOPY/ZCOPY thresh are set to force these settings */
+            cfg_priority = remote_proto->cfg_priority;
+            cfg_thresh   = ucs_max(cfg_thresh, remote_proto->cfg_thresh);
+        }
 
-    ucp_proto_common_add_proto(&params->super, &caps, &priv, sizeof(priv));
+        ucs_assertv(params->super.cfg_priority <= remote_proto->cfg_priority,
```
brminich, line 483: why?

src/ucp/rndv/proto_rndv.c
```diff
@@ -365,41 +348,147 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+                "remote_proto=%s params->super.cfg_priority=%u "
+                "remote_proto->cfg_priority=%u", proto_name,
+                params->super.cfg_priority, remote_proto->cfg_priority);
```
brminich, line 486:
```suggestion
        ucs_assertv(params->super.cfg_priority <= remote_proto->cfg_priority,
                    "remote_proto=%s params->super.cfg_priority=%u "
                    "remote_proto->cfg_priority=%u", proto_name,
                    params->super.cfg_priority, remote_proto->cfg_priority);
```

src/ucp/rndv/proto_rndv.c
ivankochin: It is a design requirement, since if variant cfg_thresh not inf or auto, then we took cfg_priority from variant to force it among others. That's why it is required that it should be higher than params->cfg_priority

src/ucp/rndv/proto_rndv.c
brminich: but you do not print `params->super.max_length` and `params->super.min_length`, so it maybe hard to see why some proto was disabled. Maybe add separate traces for the normal case and when there are not available ranges

src/ucp/rndv/proto_rndv.c
```diff
@@ -365,41 +348,145 @@ ucp_proto_rndv_ctrl_init(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+        }
+
+        cfg_thresh   = params->super.cfg_thresh;
+        cfg_priority = params->super.cfg_priority;
+        if ((remote_proto->cfg_thresh != UCS_MEMUNITS_INF) &&
+            (remote_proto->cfg_thresh != UCS_MEMUNITS_AUTO)) {
+            /* Consider remote priority and threshold only if RNDV_SCHEME or
+             * BCOPY/ZCOPY thresh are set to force these settings */
+            cfg_priority = remote_proto->cfg_priority;
+            cfg_thresh   = ucs_max(cfg_thresh, remote_proto->cfg_thresh);
```
brminich, line 478: `cfg_thresh` can be `AUTO` or `INF` here and you will take this value.

src/ucp/rndv/proto_rndv.c
brminich: maybe better to add comment about it here.

src/ucp/rndv/proto_rndv.c
ivankochin: I print them on line 418

brminich: ok

src/ucp/rndv/proto_rndv.c
```diff
@@ -475,9 +475,15 @@ void ucp_proto_rndv_ctrl_probe(const ucp_proto_rndv_ctrl_init_params_t *params,
             /* Consider remote priority and threshold only if RNDV_SCHEME or
              * BCOPY/ZCOPY thresh are set to force these settings */
             cfg_priority = remote_proto->cfg_priority;
-            cfg_thresh   = ucs_max(cfg_thresh, remote_proto->cfg_thresh);
+            cfg_thresh   = (cfg_thresh == UCS_MEMUNITS_AUTO) ?
```
brminich, line 478: can it be `INF`?

ivankochin: It can be inf only in case when RNDV_THRESH=inf, so it is OK to take this value

src/ucp/rndv/rndv_ppln.c
```diff
@@ -199,7 +199,9 @@ void ucp_proto_rndv_ppln_send_frag_complete(ucp_request_t *freq, int send_ack)
 
 static ucs_status_t ucp_proto_rndv_recv_ppln_reset(ucp_request_t *req)
 {
-    ucs_assert(req->send.rndv.ppln.ack_data_size == 0);
+    ucs_assertv(req->send.rndv.ppln.ack_data_size <= 0,
```
yosefe, line 202: how can it be < 0? maybe try to avoid it?

ivankochin: That's what we have discussed, we subtract `overlap` from this field during the `progress()` so when it sums up during completion, the size became the exact size of message, so it can be negative, but not less than `-priv->frag_proto_min_length`. I can add this check too, I mean `-frag_proto_min_length < ack_data_size <= 0`, WDYT?

src/ucp/dt/datatype_iter.inl
```diff
@@ -515,17 +513,24 @@ ucp_datatype_iter_next_ptr(const ucp_datatype_iter_t *dt_iter,
...
+        dt_iter->offset -= overlap;
+        length          += overlap;
+    }
+
+    ucp_datatype_iter_slice(dt_iter, dt_iter->offset, length, sliced_dt_iter);
 
-    length = ucp_datatype_iter_next(dt_iter, max_length, next_iter);
-    ucp_datatype_iter_slice(dt_iter, dt_iter->offset, length, sliced_dt_iter,
-                            sg_count);
+    return (overlap > 0) ? overlap : 0;
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/dt/datatype_iter.inl
```diff
@@ -515,17 +513,24 @@ ucp_datatype_iter_next_ptr(const ucp_datatype_iter_t *dt_iter,
...
+                                     size_t max_length, size_t min_frag,
+                                     ucp_datatype_iter_t *sliced_dt_iter,
+                                     ucp_datatype_iter_t *next_iter)
 {
-    size_t length;
+    size_t  length  = ucp_datatype_iter_next(dt_iter, max_length, next_iter);
+    ssize_t overlap = min_frag - length;
+
+    /* Overlap previous fragment to respect min_frag requirement */
+    if (overlap > 0) {
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/rndv/rndv_ppln.c
yosefe: ok

yosefe (approved):

brminich (approved):

PR #9735, opened 2024-03-10
AZP/TEST: Add MAD tests

Alexey-Rivkin:
## What
Add ucx_perftest between two nodes with MAD mode.

## Why ?
[HPCINFRA-1704](https://jirasw.nvidia.com/browse/HPCINFRA-1704) (Internal link).

## How
High-level sequence of operations:
2. Build UCX with MAD on both nodes.
3. Run the server side in Docker.
4. Extract server-side target values for LID, GUID and pass them as vars to the client-side tasks.
5. Run a client-side test using LID.
6. Restart the server-side.
7. Run a client-side test using GUID.
8. Stop the server-side.

Note the stage locking, to prevent parallel test execution.

### DIFF
--- a/buildlib/pr/mad_tests.yml
+++ b/buildlib/pr/mad_tests.yml
@@ -1,0 +1,112 @@
+jobs:
+  - job: SetupServer
+    displayName: Setup Server
+    pool:
+      name: MLNX
+      demands: mad_server
+    workspace:
+      clean: outputs
+    steps:
+      - checkout: self
+        clean: true
+        fetchDepth: 100
+        retryCountOnTaskFailure: 5
+      - task: Bash@3
+        name: Set_Vars
+        inputs:
+          targetType: "inline"
+          script: |
+            source ./buildlib/tools/test_mad.sh
+            set_vars
+        displayName: Set Vars
+      - bash: |
+          source ./buildlib/tools/test_mad.sh
+          build_ucx_in_docker
+          docker_run_srv
+        displayName: Setup Server
+
+  - job: SetupClient
+    displayName: Setup Client
+    pool:
+      name: MLNX
+      demands: mad_client
+    workspace:
+      clean: outputs
+    steps:
+      - checkout: self
+        clean: true
+        fetchDepth: 100
+        retryCountOnTaskFailure: 5
+      - bash: |
+          source ./buildlib/tools/test_mad.sh
+          build_ucx
+        displayName: Setup Client
+
+  - job: TestLid
+    dependsOn:
+      - SetupServer
+      - SetupClient
+    displayName: Test Lid
+    timeoutInMinutes: 10
+    pool:
+      name: MLNX
+      demands: mad_client
+    variables:
+      LID: $[ dependencies.SetupServer.outputs['Set_Vars.LID'] ]
+      HCA: $[ dependencies.SetupServer.outputs['Set_Vars.HCA'] ]
+    steps:
+      - checkout: none
+      - bash: |
+          source ./buildlib/tools/test_mad.sh
+          run_mad_test lid:$(LID)
+        env:
+          HCA: $(HCA)
+        displayName: Test LID
+
+  - job: ServerRestart
+    dependsOn: TestLid
+    displayName: Server Restart
+    pool:
+      name: MLNX
+      demands: mad_server
+    steps:
+      - checkout: none
+      - bash: |
+          source ./buildlib/tools/test_mad.sh
+          docker_run_srv
+        displayName: Server Restart
+
+  - job: TestGuid
+    dependsOn:
+      - SetupServer
+      - ServerRestart
+    displayName: Test Guid
+    timeoutInMinutes: 10
+    pool:
+      name: MLNX
+      demands: mad_client
+    variables:
+      GUID: $[ dependencies.SetupServer.outputs['Set_Vars.GUID'] ]
+      HCA: $[ dependencies.SetupServer.outputs['Set_Vars.HCA'] ]
+    steps:
+      - checkout: none
+      - bash: |
+          source ./buildlib/tools/test_mad.sh
+          run_mad_test guid:$(GUID)
+        env:
+          HCA: $(HCA)
+        displayName: Test GUID
+
+  - job: ServerStop
+    dependsOn: TestGuid
+    displayName: Server Stop
+    condition: always()
+    pool:
+      name: MLNX
+      demands: mad_server
+    steps:
+      - checkout: none
+      - bash: |
+          source ./buildlib/tools/test_mad.sh
+          docker_stop_srv
+        displayName: Server Stop
--- a/buildlib/pr/main.yml
+++ b/buildlib/pr/main.yml
@@ -255,6 +255,15 @@ stages:
               long_test: $(long_test)
               test_static: $(test_static)
 
+  - stage: ucx_perftest_mad_rte
+    dependsOn: [Static_check]
+    displayName: ucx_perftest over MAD RTE
+    lockBehavior: sequential
+    variables:
+      - group: concurrency_lock
+    jobs:
+    - template: mad_tests.yml
+
   - stage: WireCompat
     dependsOn: [Static_check]
     jobs:
--- a/buildlib/tools/test_mad.sh
+++ b/buildlib/tools/test_mad.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+set -exE -o pipefail
+
+IMAGE="rdmz-harbor.rdmz.labs.mlnx/ucx/x86_64/rhel8.2/builder:mofed-5.0-1.0.0.0"
+
+if [ -z "$BUILD_SOURCESDIRECTORY" ]; then
+    echo "Not running in Azure"
+    exit 1
+fi
+cd "$BUILD_SOURCESDIRECTORY"
+
+build_ucx() {
+    ./autogen.sh
+    ./contrib/configure-release \
+        --prefix="$PWD"/install \
+        --with-mad \
+        --without-valgrind \
+        --without-go \
+        --without-java
+    make -s -j"$(nproc)"
+    make install
+}
+
+build_ucx_in_docker() {
+    docker run --rm \
+        --name ucx_build_"$BUILD_BUILDID" \
+        -e BUILD_SOURCESDIRECTORY="$BUILD_SOURCESDIRECTORY" \
+        -v "$PWD":"$PWD" -w "$PWD" \
+        -v /hpc/local:/hpc/local \
+        $IMAGE \
+        bash -c "source ./buildlib/tools/test_mad.sh && build_ucx"
+
+    sudo chown -R swx-azure-svc:ecryptfs "$PWD"
+}
+
+docker_run_srv() {
+    HCA=$(detect_hca)
+    docker_stop_srv
+    docker run --rm \
+        --detach \
+        --net=host \
+        --name ucx_perftest_"$BUILD_BUILDID" \
+        -e BUILD_SOURCESDIRECTORY="$BUILD_SOURCESDIRECTORY" \
+        -v "$PWD":"$PWD" -w "$PWD" \
+        -v /hpc/local:/hpc/local \
+        --ulimit memlock=-1:-1 --device=/dev/infiniband/ \
+        $IMAGE \
+        bash -c "${PWD}/install/bin/ucx_perftest -K ${HCA}"
+}
+
+docker_stop_srv() {
+    docker stop ucx_perftest_"$BUILD_BUILDID" || true
+}
+
+set_vars() {
+    set +x
+    HCA=$(detect_hca)
+    # Replace ':' with space for 'ibstat' format
+    HCA_DEV=${HCA/:/ }
+    # shellcheck disable=SC2086
+    LID=$(ibstat $HCA_DEV | grep Base | awk '{print $NF}')
+    # shellcheck disable=SC2086
+    GUID=$(ibstat $HCA_DEV | grep GUID | awk '{print $NF}')
+    echo "##vso[task.setvariable variable=LID;isOutput=true]$LID"
+    echo "##vso[task.setvariable variable=GUID;isOutput=true]$GUID"
+    echo "##vso[task.setvariable variable=HCA;isOutput=true]$HCA"
+    echo "LID: $LID"
+    echo "GUID: $GUID"
+    echo "HCA: $HCA"
+}
+
+run_mad_test() {
+    local ib_address="$1"
+    sudo chmod 777 /dev/infiniband/umad*
+    "$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" -e "$ib_address"
+}
+
+detect_hca() {
+    # Detect first active HCA port
+    ibv_devinfo | awk '/hca_id:/ {hca=$2} /port:/ {port=$2} /PORT_ACTIVE/ {print hca ":" port; exit}'
+}


Alexey-Rivkin:
Test build:
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=77684

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+set -exE -o pipefail
+
+cd "$BUILD_SOURCESDIRECTORY"
+
+run_mad_server() {
+	build_ucx
```
tvegas1: indent

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+    setup
+    srv_stop
+    funcname
+    export HCA="$HCA"
+	sudo -E bash -c 'envsubst < "buildlib/tools/ucx_perftest.template" \
```
tvegas1: indent, 5 lines

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+    # shellcheck disable=SC2086
+    GUID=$(ibstat $HCA | grep GUID | awk '{print $NF}')
+    echo "LID: $LID"
+    echo "GUID: $GUID"
+    echo "##vso[task.setvariable variable=LID;isOutput=true]$LID"
+    echo "##vso[task.setvariable variable=GUID;isOutput=true]$GUID"
+}
+
+run_mad_test_lid() {
+	build_ucx
```
tvegas1: indent

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+    setup
+    funcname
+	"$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" -e lid:"$LID"
```
tvegas1: indent

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+}
+
+run_mad_test_guid() {
+    funcname
+	"$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" guid:"$GUID"
```
tvegas1: indent

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+}
+
+srv_restart() {
+    funcname
+    sudo systemctl stop ucx_perftest
```
tvegas1, line 64: should we ignore return code of `stop` in case it would not be there?

buildlib/pr/mad_tests.yml
```diff
@@ -0,0 +1,91 @@
...
+      - bash: ./contrib/test_mad.sh run_mad_server
+        env:
+          HCA: ${{ parameters.HCA }}
+        displayName: Server Up          
+
+      - task: Bash@3
+        name: Set_Vars
+        inputs:
+          targetType: "inline"
+          script: ./contrib/test_mad.sh set_vars
```
tvegas1, line 25: for my understanding: why not sourcing contrib/test_mad.sh then calling set_vars?

buildlib/tools/ucx_perftest.template
```diff
@@ -0,0 +1,23 @@
...
+Environment=UCX_RDMA_CM_TIMEOUT=5
+Restart=always
+RestartSec=500ms
+User=root
+Group=root
+StartLimitInterval=10s
+StartLimitBurst=100
+StandardOutput=${PWD}/ucx_perftest.log
+StandardError=${PWD}/ucx_perftest.log
+ExecStart=${PWD}/install/bin/ucx_perftest -e -K ${HCA}
```
tvegas1, line 23: double quotes around HCA bash variable and around PWD?

buildlib/pr/mad_tests.yml
Alexey-Rivkin: Sourcing for each step is repetitive. Why not directly execute the required functions?

buildlib/tools/ucx_perftest.template
Alexey-Rivkin: Done

contrib/test_mad.sh
Alexey-Rivkin: If service doesn't run at this point, then something is wrong and I'd better know about it.
In contrast, we do ignore results of srv_stop().

contrib/test_mad.sh
Alexey-Rivkin: Unified indent by auto-formating.
While seemingly the same, some indents used Tabs, while others 4 spaces.

contrib/test_mad.sh
Alexey-Rivkin: Unified indent by auto-formating.

contrib/test_mad.sh
Alexey-Rivkin: Unified indent by auto-formating.

contrib/test_mad.sh
Alexey-Rivkin: Unified indent by auto-formating.

buildlib/pr/mad_tests.yml
dpressle: I agree with Thomas, sourcing bash functions and then run them is best practice and what everyone is familier with, your implementation is a little bit confusing and very unusual.
However, i understand that if we need to source it each task/step it might look like repetitive work, but some time we can have repetitive in favor of readable code

buildlib/pr/main.yml
```diff
@@ -188,6 +188,13 @@ stages:
     jobs:
     - template: static_checks.yml
 
+  - stage: MAD_Tests
+    dependsOn: [Static_check]
+    jobs:
+    - template: mad_tests.yml
+      parameters:
+        HCA: "mlx5_0:1"
```
dpressle, line 196: why we need this param here? who will control it from main.yml scope? Do we usually manage hard coded parameters from main.yml?

buildlib/tools/ucx_perftest.template
dpressle, line 1: can we have more meaningful name for this file? its the server service right? 
are all files under tools dir service related?

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+    echo "GUID: $GUID"
+    echo "##vso[task.setvariable variable=LID;isOutput=true]$LID"
+    echo "##vso[task.setvariable variable=GUID;isOutput=true]$GUID"
+}
+
+run_mad_test_lid() {
+    build_ucx
+    setup
+    funcname
+    "$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" -e lid:"$LID"
```
dpressle, line 54: PWD should be set globally by this command:
WORKDIRECTORY=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )

This is the best practice approach in bash for setting the current working directory and since you are using it multiple times it should be global var

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+}
+
+run_mad_test_guid() {
+    funcname
+    "$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" guid:"$GUID"
+}
+
+srv_restart() {
+    funcname
+    sudo systemctl stop ucx_perftest
```
dpressle, line 64: maybe call srv_stop function to reduce duplicated code?

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+run_mad_server() {
+    build_ucx
+    setup
+    srv_stop
+    funcname
+    export HCA="$HCA"
+    sudo -E bash -c 'envsubst < "buildlib/tools/ucx_perftest.template" \
+        > /etc/systemd/system/ucx_perftest.service'
+    sudo systemctl daemon-reload
+    sudo systemctl start ucx_perftest
```
dpressle, line 15: maybe call srv_start function to reduce duplicated code?

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+
+run_mad_test_guid() {
+    funcname
+    "$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" guid:"$GUID"
+}
+
+srv_restart() {
+    funcname
+    sudo systemctl stop ucx_perftest
+    sudo systemctl start ucx_perftest
```
dpressle, line 65: maybe call srv_start function to reduce duplicated code?

contrib/test_mad.sh
```diff
@@ -0,0 +1,83 @@
...
+    # shellcheck disable=SC2086
+    GUID=$(ibstat $HCA | grep GUID | awk '{print $NF}')
+    echo "LID: $LID"
+    echo "GUID: $GUID"
+    echo "##vso[task.setvariable variable=LID;isOutput=true]$LID"
+    echo "##vso[task.setvariable variable=GUID;isOutput=true]$GUID"
+}
+
+run_mad_test_lid() {
+    build_ucx
```
dpressle, line 51: why we build ucx as part of this test? it should be separated function, what if someone decide to run only guid tests it will not work

buildlib/tools/ucx_perftest.template
tvegas1: suggesting rename to `ucx_perftest.service`

Alexey-Rivkin: Renaming the file. 
Would you suggest another dir to place it?

buildlib/tools/ucx_perftest.template
tvegas1: checked, seems it could be at `./contrib/ucx_perftest.service` since it can be make non-CI specific

buildlib/pr/mad_tests.yml
dpressle, line 1: I would like to see separate setup and build stages instead of hiding them in some test stage

contrib/test_mad.sh
Alexey-Rivkin: $PWD is already a globally available environment variable.
In this script we ensure a proper PWD value by `cd "$BUILD_SOURCESDIRECTORY"`, which is an Azure built-in.
Please let me know if you still want to introduce more variables.

buildlib/pr/mad_tests.yml
Alexey-Rivkin: Done

buildlib/pr/mad_tests.yml
Alexey-Rivkin: Done

buildlib/pr/main.yml
Alexey-Rivkin: Moved the hardcoded var into the Bash script.

buildlib/tools/ucx_perftest.template
Alexey-Rivkin: DOne

contrib/test_mad.sh
Alexey-Rivkin: A separate “Setup” job has been created, including UCX build and system setup.

contrib/test_mad.sh
Alexey-Rivkin: Done

buildlib/pr/main.yml
```diff
@@ -188,6 +188,11 @@ stages:
     jobs:
     - template: static_checks.yml
 
+  - stage: MAD_Tests
```
dpressle, line 191: I want to suggest something that i am trying to promote in our pipelines for some time - its not a blocker for this PR but a reminder for our self.

I would like us to use the "display name" key everywhere, the reason for this is to have consistant and unified naming convention of steps/jobs/stages etc in code and in UI.

The idea is to be able to have different naming convention for stage/job/step names one for pipeline code and one for UI display, for example:

UI should present the stage name as:
"Mad Tests"
instead of the "coding" way of:
"MAD_Tests"

And for internal pipeline use we should have it always as same convention (we should agree on one but see my suggestion below in the example as it is in AZP examples), so in this example the code should look like this:

- stage: MadTests
   displayName: Mad Tests

Alexey-Rivkin: Sure

buildlib/pr/mad_tests.yml
```diff
@@ -0,0 +1,114 @@
...
+    displayName: Test Lid
+    timeoutInMinutes: 10
+    pool:
+      name: MLNX
+      demands: mad_client
+    variables:
+      LID: $[ dependencies.ServerUp.outputs['Set_Vars.LID'] ]
+    steps:
+      - checkout: none
+      - bash: $(Build.SourcesDirectory)/install/bin/ucx_perftest -t tag_bw -e -K $(HCA) -e lid:$(LID)
```
tvegas1, line 71: tbh, I preferred previous version where commands were coming from sourced bash, easier to run/test on dev machines.

tvegas1:
lgtm

buildlib/pr/mad_tests.yml
Alexey-Rivkin: Your wish has been granted.

tvegas1 (approved):

dpressle (approved):

tvegas1:
@yosefe, could you please have a look?

contrib/ucx_perftest.service
```diff
@@ -0,0 +1,23 @@
...
+ 
+[Service]
+Type=simple
+Environment=UCX_KEEPALIVE_INTERVAL=2
+Environment=UCX_DC_MLX5_TIMEOUT=5
+Environment=UCX_RC_VERBS_TIMEOUT=5
+Environment=UCX_RC_MLX5_TIMEOUT=5
+Environment=UCX_UD_VERBS_TIMEOUT=5
+Environment=UCX_UD_MLX5_TIMEOUT=5
+Environment=UCX_RDMA_CM_TIMEOUT=5
```
yosefe, line 14: why do we need these vars?

contrib/test_mad.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+set -exE -o pipefail
+
+export HCA="mlx5_0:1"
```
yosefe, line 4: can we detect first active port instead of hardcoded?

contrib/ucx_perftest.service
```diff
@@ -0,0 +1,23 @@
...
+Restart=always
+RestartSec=500ms
+User=root
+Group=root
+StartLimitInterval=10s
+StartLimitBurst=100
+StandardOutput="${PWD}"/ucx_perftest.log
+StandardError="${PWD}"/ucx_perftest.log
```
yosefe, line 22: seems it causes errors in Azure:
```
Mar 12 22:13:03 vulcan03 systemd[1]: /etc/systemd/system/ucx_perftest.service:21: Failed to parse output specifier, ignoring: "/scrap/azure/agent-02/AZP_WORKSPACE/2/s"/ucx_perftest.log
Mar 12 22:13:03 vulcan03 systemd[1]: /etc/systemd/system/ucx_perftest.service:22: Failed to parse output specifier, ignoring: "/scrap/azure/agent-02/AZP_WORKSPACE/2/s"/ucx_perftest.log
Mar 12 22:13:03 vulcan03 systemd[1]: /etc/systemd/system/ucx_perftest.service:21: Failed to parse output specifier, ignoring: "/scrap/azure/agent-02/AZP_WORKSPACE/2/s"/ucx_perftest.log
Mar 12 22:13:03 vulcan03 systemd[1]: /etc/systemd/system/ucx_perftest.service:22: Failed to parse output specifier, ignoring: "/scrap/azure/agent-02/AZP_WORKSPACE/2/s"/ucx_perftest.log
Mar 12 22:13:03 vulcan03 systemd[1]: /etc/systemd/system/ucx_perftest.service:21: Failed to parse output specifier, ignoring: "/scrap/azure/agent-02/AZP_WORKSPACE/2/s"/ucx_perftest.log
```

contrib/test_mad.sh
```diff
@@ -0,0 +1,78 @@
...
+    funcname
+    "$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" -e lid:"$LID"
+}
+
+run_mad_test_guid() {
+    funcname
+    "$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" guid:"$GUID"
+}
+
+srv_start() {
```
yosefe, line 58: what is the reason to use a system service?
this requires running the test as root, and it's not always desired

contrib/ucx_perftest.service
tvegas1: when running perftest with `-e`, we rely on the transport to call the error handler and terminate. using shorter timeout helps emulate more synchronuous behavior. else when perftest client ctrl-c/aborts, it can be tens for seconds before server stops, restarts and becomes available again.

yosefe: why do we need to run perftest with -e?

tvegas1: without `-e` perftest never terminates if something went wrong during the test (client abort/interrupt...), so running it as a service would not work.

yosefe: i think we should not run it as a service

tvegas1: ok, for that aspect I'll let Alexey comment

Alexey-Rivkin: Daemonizing the server side allows orchestrating client-server testing in CI. 
Running a background process in Azure is not allowed otherwise. 
We successfully use the same approach with SparkDPU.

yosefe: when we run io_demo, for example, we don't use a system service, but we run a process in the background. is that a wrong approach?

Alexey-Rivkin: I see that "io_demo" uses SSH to run commands on nodes. While this is a valid solution, it's more complex and error-prone.
Laying out the test flow in minimalistic Yaml steps results in straightforward code and separate logs, both are readable and make troubleshooting easier.
Simplicity and manageability are the keys to sustainable devops.

yosefe (commented):
per offline discussion:
- run the test in a container
- move files from contrib/ to buildlib/
- remove special UCX env vars for shorter timeout

contrib/test_mad.sh
Alexey-Rivkin: Running the server side on Docker instead of as a system service.

contrib/ucx_perftest.service
Alexey-Rivkin: The server daemon is not used now.

contrib/test_mad.sh
Alexey-Rivkin: Done, see `detect_hca()` in the Shell script.

contrib/ucx_perftest.service
Alexey-Rivkin: Removed "TIMEOUT" vars.
The server side now runs on Docker.

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+set -exE -o pipefail
+
+IMAGE="rdmz-harbor.rdmz.labs.mlnx/ucx/x86_64/rhel8.2/builder:mofed-5.0-1.0.0.0"
+cd "$BUILD_SOURCESDIRECTORY"
```
dpressle, line 5: what about checking this var exists?

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+    set -x
+}
+
+detect_hca() {
+    echo "Detect first active HCA port"
+    HCA="$(ibv_devinfo | awk '/hca_id:/ {hca=$2} /port:/ {port=$2} /PORT_ACTIVE/ {print hca ":" port; exit}')"
+    export HCA
+}
+
+"$@"
```
dpressle, line 93: why you need this?

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+
+funcname() {
+    set +x
+    echo "==== Running: ${FUNCNAME[1]} ===="
+    set -x
+}
+
+detect_hca() {
+    echo "Detect first active HCA port"
+    HCA="$(ibv_devinfo | awk '/hca_id:/ {hca=$2} /port:/ {port=$2} /PORT_ACTIVE/ {print hca ":" port; exit}')"
```
dpressle, line 89: Will this be a problem that we re-calculate the value in different stages of this build?
Is it possible we can get different results in different stages?

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+        -v "$PWD":"$PWD" -w "$PWD" \
+        -v /hpc/local:/hpc/local \
+        --gpus all --ulimit memlock=-1:-1 --device=/dev/infiniband/ \
+        $IMAGE \
+        bash -c "source ./buildlib/tools/test_mad.sh && \
+                ${PWD}/install/bin/ucx_perftest -K ${HCA}"
+}
+
+docker_stop_srv() {
+    docker stop ucx_perftest_"$BUILD_BUILDID" || true
```
dpressle, line 51: stop is not enough we should remove the container as well, we dont want to have dangling containers all the time

buildlib/pr/mad_tests.yml
```diff
@@ -0,0 +1,113 @@
...
+  - job: ServerRestart
+    dependsOn: TestLid
+    displayName: Server Restart
+    pool:
+      name: MLNX
+      demands: mad_server
+    steps:
+      - checkout: none
+      - bash: |
+          docker stop ucx_perftest_$(Build.BuildId)
```
dpressle, line 76: business logic should not be done here but in the bash script

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+}
+
+set_vars() {
+    set +x
+    detect_hca
+    # Replace ':' with space for 'ibstat' format
+    HCA=${HCA/:/ }
+    # shellcheck disable=SC2086
+    LID=$(ibstat $HCA | grep Base | awk '{print $NF}')
+    # shellcheck disable=SC2086
```
dpressle, line 61: why disable? its only adding double quotes on $HCA

Alexey-Rivkin: Adding double quotes breaks the functionality

buildlib/pr/mad_tests.yml
Alexey-Rivkin: Done

buildlib/tools/test_mad.sh
Alexey-Rivkin: We have a daily cleanup of Docker resources.

buildlib/tools/test_mad.sh
Alexey-Rivkin: Done

buildlib/tools/test_mad.sh
Alexey-Rivkin: Removed as not needed anymore.

buildlib/tools/test_mad.sh
Alexey-Rivkin: Docker has error handling of its own.

dpressle: How is this related to Docker? this is executed every time you source this file

buildlib/tools/test_mad.sh
dpressle: So we will have many stopped containers every day until they are cleaned? this is bad practice, each run should do its best to clean after it self, the daily cleanup is extra cleanup not instead

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,92 @@
...
+funcname() {
+    set +x
+    echo "==== Running: ${FUNCNAME[1]} ===="
+    set -x
+}
+
+detect_hca() {
+    echo "Detect first active HCA port"
+    HCA="$(ibv_devinfo | awk '/hca_id:/ {hca=$2} /port:/ {port=$2} /PORT_ACTIVE/ {print hca ":" port; exit}')"
+    export HCA
```
dpressle, line 91: so why we need export now?

dpressle (commented):
Why do we have "Error response from daemon: No such container: ucx_perftest_79028" in server restart?

buildlib/tools/test_mad.sh
Alexey-Rivkin: Check out the `--rm ` flag.

buildlib/tools/test_mad.sh
Alexey-Rivkin: What's the purpose of checking if this var exists?

buildlib/tools/test_mad.sh
dpressle: so why do you stop it? if we need to clean than we need to clean we stop is not cleanning

Alexey-Rivkin:
> Why do we have "Error response from daemon: No such container: ucx_perftest_79028" in server restart? https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=79028&view=logs&j=551595a6-c33f-5841-440b-a6858ba5ebd6&t=67dc2fe8-6bfb-54a0-4bcb-22f44968bb72&l=24

Server-side exit following a test. The 'docker stop' command is to confirm the killing.

dpressle:
Please check this error
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=79028&view=logs&j=c0372fa8-9ef6-5b2d-27ac-eb9499b58ab4&t=f6e32584-77ba-5c83-c97d-60655a1c1d38&l=21

buildlib/tools/test_mad.sh
Alexey-Rivkin: Containers get cleaned once stopped.

Alexey-Rivkin:
> Please check this error https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=79028&view=logs&j=c0372fa8-9ef6-5b2d-27ac-eb9499b58ab4&t=f6e32584-77ba-5c83-c97d-60655a1c1d38&l=21

Cosmetic error, resolved by adding a mount.

Alexey-Rivkin:
> > Why do we have "Error response from daemon: No such container: ucx_perftest_79028" in server restart? https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=79028&view=logs&j=551595a6-c33f-5841-440b-a6858ba5ebd6&t=67dc2fe8-6bfb-54a0-4bcb-22f44968bb72&l=24
> 
> Server-side exit following a test. The 'docker stop' command is to confirm the killing.

@tvegas1 please confirm this is expected behavior

buildlib/tools/test_mad.sh
dpressle: Robustness, technical debt, separation of business logic from flow management - all the things we are talking about all the time ...

dpressle:
@yosefe please review, thanks.

buildlib/tools/test_mad.sh
Alexey-Rivkin: If you have a specific concern, I could address that.

yosefe: if this var is undefined the script will go to home dir and do things there

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+set -exE -o pipefail
+
+IMAGE="rdmz-harbor.rdmz.labs.mlnx/ucx/x86_64/rhel8.2/builder:mofed-5.0-1.0.0.0"
+cd "$BUILD_SOURCESDIRECTORY"
+
+node_setup() {
+    funcname
+    sudo chmod 777 /dev/infiniband/umad*
```
yosefe, line 9: This should be done by admin, not by test job

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+build_ucx_in_docker() {
+    docker run --rm \
+        --name ucx_build_"$BUILD_BUILDID" \
+        -v "$PWD":"$PWD" -w "$PWD" \
+        -v /hpc/local:/hpc/local \
+        $IMAGE \
+        bash -c "source ./buildlib/tools/test_mad.sh && build_ucx"
+
+    sudo chown -R swx-azure-svc:ecryptfs "$PWD"
+}
```
yosefe, line 34: can we do it by Azure "container:" field, like in other build jobs, instead of running docker manually?

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+    docker run --rm \
+        --detach \
+        --net=host \
+        -e HCA="$HCA" \
+        --name ucx_perftest_"$BUILD_BUILDID" \
+        -v "$PWD":"$PWD" -w "$PWD" \
+        -v /hpc/local:/hpc/local \
+        --gpus all --ulimit memlock=-1:-1 --device=/dev/infiniband/ \
+        $IMAGE \
+        bash -c "source ./buildlib/tools/test_mad.sh && \
```
yosefe, line 48: why need to source test_mad.sh?

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+    sudo chown -R swx-azure-svc:ecryptfs "$PWD"
+}
+
+docker_run_srv() {
+    detect_hca
+    docker_stop_srv
+    docker run --rm \
+        --detach \
+        --net=host \
+        -e HCA="$HCA" \
```
yosefe, line 42: why needed?

buildlib/tools/test_mad.sh
yosefe: IMO better return the value as printing to stdout and caller of this function will do HCA=$(detect_hca)

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+
+docker_stop_srv() {
+    docker stop ucx_perftest_"$BUILD_BUILDID" || true
+}
+
+set_vars() {
+    set +x
+    detect_hca
+    # Replace ':' with space for 'ibstat' format
+    HCA_IBSTAT=${HCA/:/ }
```
yosefe, line 60: rename HCA_IBSTAT -> HCA_DEV

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+
+run_mad_test_lid() {
+    funcname
+    "$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" -e lid:"$LID"
+}
+
+run_mad_test_guid() {
+    funcname
+    "$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" guid:"$GUID"
+}
```
yosefe, line 81: IMO these wrappers are redundant

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,93 @@
...
+    detect_hca
+    docker_stop_srv
+    docker run --rm \
+        --detach \
+        --net=host \
+        -e HCA="$HCA" \
+        --name ucx_perftest_"$BUILD_BUILDID" \
+        -v "$PWD":"$PWD" -w "$PWD" \
+        -v /hpc/local:/hpc/local \
+        --gpus all --ulimit memlock=-1:-1 --device=/dev/infiniband/ \
```
yosefe, line 46: minor: no need to pass gpu

buildlib/tools/test_mad.sh
dpressle: And what if this folder gets recreated (reboot???)? test will fail, and we will need to manually chmod it.

buildlib/tools/test_mad.sh
dpressle: IMO, we should separate business logic from flow, having all the logic in this script helps by allowing easy way of running the tests without having to reverse engineer the AZP yaml.

buildlib/tools/test_mad.sh
Alexey-Rivkin: Removed, thx.

buildlib/tools/test_mad.sh
Alexey-Rivkin: Removed, thx.

buildlib/tools/test_mad.sh
Alexey-Rivkin: Removed, thx.

buildlib/tools/test_mad.sh
Alexey-Rivkin: Merged into a single function `run_mad_test()` with params.

buildlib/tools/test_mad.sh
Alexey-Rivkin: Done

buildlib/tools/test_mad.sh
Alexey-Rivkin: Had to resort to the manual Docker run commands to run the service container in a detached mode. 
To build UCX in Docker we can indeed use a separate job with a "container:" field. 
As Daniel mentioned, this would affect the code's readability.

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,88 @@
...
+    echo "##vso[task.setvariable variable=LID;isOutput=true]$LID"
+    echo "##vso[task.setvariable variable=GUID;isOutput=true]$GUID"
+    echo "##vso[task.setvariable variable=HCA;isOutput=true]$HCA"
+    echo "LID: $LID"
+    echo "GUID: $GUID"
+    echo "HCA: $HCA"
+}
+
+run_mad_test() {
+    test_type="$1"
```
dpressle, line 74: if you use arguments you should validate them and make them local to function, something like:
local -r test_type=${1:-}
local -r ib_add=${2:-}
[[ -z $test_type ]] && { exit 1; }
[[ -z $ib_add ]] && { exit 1; }

buildlib/pr/mad_tests.yml
```diff
@@ -0,0 +1,112 @@
...
+      name: MLNX
+      demands: mad_client
+    variables:
+      LID: $[ dependencies.SetupServer.outputs['Set_Vars.LID'] ]
+      HCA: $[ dependencies.SetupServer.outputs['Set_Vars.HCA'] ]
+    steps:
+      - checkout: none
+      - bash: |
+          source ./buildlib/tools/test_mad.sh
+          run_mad_test lid $(LID)
```
tvegas1, line 61: we might as well either pass `lid:$(LID)` or `guid:$(GUID)` to simplify argument handling in the called function?

tvegas1:
> > > Why do we have "Error response from daemon: No such container: ucx_perftest_79028" in server restart? https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=79028&view=logs&j=551595a6-c33f-5841-440b-a6858ba5ebd6&t=67dc2fe8-6bfb-54a0-4bcb-22f44968bb72&l=24
> > 
> > 
> > Server-side exit following a test. The 'docker stop' command is to confirm the killing.
> 
> @tvegas1 please confirm this is expected behavior

Yes it's expected behaviour, after a successful test, server and client both exit successfully.

buildlib/tools/test_mad.sh
Alexey-Rivkin: The `ucx_perftest` binary has built-in params validation.
Added `local` to limit var's scope.

buildlib/tools/test_mad.sh
yosefe: so maybe the test itself should run as root to be able to access this device node
in any case the test should not perform changes on the system

buildlib/tools/test_mad.sh
yosefe: usually our build jobs are in a separate task with "container:" tag. so i think for consistency this one should do it as well.

buildlib/tools/test_mad.sh
Alexey-Rivkin: As umad files are created dynamically, permissions must be set before running the test.
Otherwise we end up with errors:
[`UCX  ERROR mad_rpc_open_port(ca="mlx5_0" ca_port=1 mgmt_classes=IB_SA_CLASS) failed: Permission denied`](https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=79418&view=logs&j=fc80f01c-33c4-5912-cc2b-a3708338f199&t=0a6a587e-5e7b-5fe5-5647-b8ca39406e3c&l=19)

yosefe: the mad files should be created by openibd service, what do you mean by they are created dynamically?

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,88 @@
...
+    local ib_add="$1"
+    sudo chmod 777 /dev/infiniband/umad*
+    "$PWD"/install/bin/ucx_perftest -t tag_bw -e -K "$HCA" -e "$ib_add"
+}
+
+funcname() {
+    set +x
+    echo "==== Running: ${FUNCNAME[1]} ===="
+    set -x
+}
```
yosefe, line 83: seems redundant, it's used only once

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,88 @@
...
+    echo "##vso[task.setvariable variable=LID;isOutput=true]$LID"
+    echo "##vso[task.setvariable variable=GUID;isOutput=true]$GUID"
+    echo "##vso[task.setvariable variable=HCA;isOutput=true]$HCA"
+    echo "LID: $LID"
+    echo "GUID: $GUID"
+    echo "HCA: $HCA"
+}
+
+run_mad_test() {
+    local ib_add="$1"
```
yosefe, line 74: what is "ib_add"? maybe "ib_address"?

buildlib/tools/test_mad.sh
```diff
@@ -0,0 +1,88 @@
...
+        $IMAGE \
+        bash -c "${PWD}/install/bin/ucx_perftest -K ${HCA}"
+}
+
+docker_stop_srv() {
+    docker stop ucx_perftest_"$BUILD_BUILDID" || true
+}
+
+set_vars() {
+    set +x
```
yosefe, line 57: why needed?

buildlib/pr/main.yml
```diff
@@ -255,6 +255,15 @@ stages:
               long_test: $(long_test)
               test_static: $(test_static)
 
+  - stage: MadTests
+    dependsOn: [Static_check]
+    displayName: Mad Tests
```
yosefe, line 260: ucx_perftest over MAD RTE

Alexey-Rivkin: Done

buildlib/tools/test_mad.sh
Alexey-Rivkin: Azure quirks:
https://developercommunity.visualstudio.com/t/pipeline-variable-incorrectly-inserts-single-quote/375679#T-N394968

buildlib/tools/test_mad.sh
Alexey-Rivkin: Yes, changed

buildlib/tools/test_mad.sh
Alexey-Rivkin: Done

buildlib/pr/main.yml
```diff
@@ -255,9 +255,9 @@ stages:
               long_test: $(long_test)
               test_static: $(test_static)
 
-  - stage: MadTests
+  - stage: ucx_perftestOverMAD_RTE
```
yosefe, line 258: ucx_perftest_mad_rte

buildlib/tools/test_mad.sh
```diff
@@ -10,7 +10,7 @@ fi
 cd "$BUILD_SOURCESDIRECTORY"
 
 build_ucx() {
-    funcname
+    echo "==== Running: ${FUNCNAME[1]} ===="
```
yosefe, line 13: IMO can remove this

yosefe (approved):

yosefe:
@Alexey-Rivkin can you pls squash?

PR #9739, opened 2024-03-11
RELEASE: Update NEWS 1.16 RC4

shasson5:
## What
Update NEWS 1.16 RC4

## Why ?
Prepare v.16 RC4

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -13,8 +13,14 @@
 
 ## 1.16.0-rc4 (February 21, 2024)
 ### Bugfixes:
-#### UCT
-* Fixed Infiniband posting error when compiler optimizes code with memmove()
+#### UCP
+* Disabled rendezvous pipeline protocol selection when using non-contiguous buffer
+#### RDMA CORE (IB, ROCE, etc.)
+* Fixed mlx5 WQE posting error due to compiler memory copy optimizations
+#### GPU (CUDA, ROCM)
+* Fixed cuda_ipc transport being disabled if a CUDA device is not set during initialization
+#### UCM
+* Fixed compilation error when building on PPC64
 #### Packaging
 * Fixed already existing target error when using cmake find_package(ucx) twice
 


NEWS
```diff
@@ -13,8 +13,14 @@
 
 ## 1.16.0-rc4 (February 21, 2024)
 ### Bugfixes:
+#### UCP
+* Disabled rendezvous pipeline protocol selection when using non-contiguous buffer.
 #### UCT
 * Fixed Infiniband posting error when compiler optimizes code with memmove()
+#### GPU (CUDA, ROCM)
+* Fixed CUDA IPC disabling when fails to initialize.
```
yosefe, line 21: this is not clear, need to rephrase, what fails to initialize and in which cases?

NEWS
```diff
@@ -13,8 +13,14 @@
...
+#### Build
+* Fixed compilation error when building on PPC64.
```
yosefe, line 23: isn't it UCM fix? should be in UCM section

NEWS
```diff
@@ -13,8 +13,14 @@
 
 ## 1.16.0-rc4 (February 21, 2024)
 ### Bugfixes:
+#### UCP
+* Disabled rendezvous pipeline protocol selection when using non-contiguous buffer.
 #### UCT
 * Fixed Infiniband posting error when compiler optimizes code with memmove()
+#### GPU (CUDA, ROCM)
+* Fixed CUDA IPC disabling when CUDA device is not set for the process.
```
yosefe, line 21: Fixed cuda_ipc transport being disabled if a CUDA device is not set during initialization

NEWS
```diff
@@ -13,8 +13,14 @@
 
 ## 1.16.0-rc4 (February 21, 2024)
 ### Bugfixes:
+#### UCP
```
tvegas1, line 16: can we also add details for #9723?

yosefe (approved):

NEWS
shasson5: isn't it covered by the comment for #9705 ?

shasson5:
@yosefe can merge?

NEWS
yosefe: Maybe rephrase it to "Fixed mlx5 WQE posting error due to compiler memory copy optimizations"
and rename the section from "UCT" to "RDMA CORE" like in line 56

yosefe (approved):

shasson5:
@yosefe squashed

PR #9743, opened 2024-03-12
UCT/IB/MLX5/DV: fixed memic iova in ksm registration

roiedanino:
## What
In ksm registration: setting iova to `(uint64_t)memh->address` in case of memic.
In perftest: check allocators are not NULL before initializing them

## Why ?
MEMIC-relate bugfixes - 
1. Fix in ksm registration is necessary for rdma memory to work
2. Fix in perftest avoids a segfault when using multiple threads with RDMA memory

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -241,7 +241,7 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_addr(
  * that were registered in single-threaded mode.
  */
 static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_contig(
-        uct_ib_mlx5_md_t *md, void *address, uint32_t iova_offset, int atomic,
+        uct_ib_mlx5_md_t *md, void *address, uint64_t iova, int atomic,
         uint32_t mkey_index, const char *reason, uct_ib_mlx5_devx_mr_t *mr,
         struct mlx5dv_devx_obj **mr_p, uint32_t *mkey)
 {
@@ -254,7 +254,7 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_contig(
     /* FW requires indirect atomic MR address and length to be aligned
      * to max supported atomic argument size */
     ksm_address = ucs_align_down_pow2((uint64_t)address, UCT_IB_MD_MAX_MR_SIZE);
-    ksm_iova    = ksm_address + iova_offset;
+    ksm_iova    = ksm_address - (uint64_t)address + iova;
     ksm_length  = mr_length + (uint64_t)address - ksm_address;
     ksm_length  = ucs_align_up(ksm_length, md->super.dev.atomic_align);
 
@@ -436,16 +436,16 @@ uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md,
 {
     uct_ib_mlx5_devx_mr_t *mr = &memh->mrs[mr_type];
     void *address             = uct_ib_mlx5_devx_memh_base_address(memh);
+    uint64_t iova             = (uint64_t)memh->address + iova_offset;
 
     if (memh->super.flags & UCT_IB_MEM_MULTITHREADED) {
-        return uct_ib_mlx5_devx_reg_ksm_data_mt(md, address,
-                                                (uint64_t)address + iova_offset,
-                                                atomic, mkey_index, reason,
+        return uct_ib_mlx5_devx_reg_ksm_data_mt(md, address, iova, atomic,
+                                                mkey_index, reason,
                                                 mr->ksm_data, mr_p, mkey);
     } else {
-        return uct_ib_mlx5_devx_reg_ksm_data_contig(md, address, iova_offset,
-                                                    atomic, mkey_index, reason,
-                                                    mr, mr_p, mkey);
+        return uct_ib_mlx5_devx_reg_ksm_data_contig(md, address, iova, atomic,
+                                                    mkey_index, reason, mr,
+                                                    mr_p, mkey);
     }
 }
 
@@ -650,7 +650,7 @@ static void uct_ib_mlx5_devx_reg_symmetric(uct_ib_mlx5_md_t *md,
     /* Best effort, only allocate in the range below the atomic keys. */
     while (md->smkey_index < md->super.mkey_by_name_reserve.size) {
         status = uct_ib_mlx5_devx_reg_ksm_data_contig(
-                md, address, 0,
+                md, address, (uint64_t)address,
                 (memh->super.flags & UCT_IB_MEM_ACCESS_REMOTE_ATOMIC),
                 md->super.mkey_by_name_reserve.base + md->smkey_index,
                 "symmetric-key", &memh->mrs[UCT_IB_MR_DEFAULT], &smkey_mr,
@@ -2097,7 +2097,8 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_exported_key,
     }
 
     ucs_assert(!(memh->super.flags & UCT_IB_MEM_MULTITHREADED));
-    status = uct_ib_mlx5_devx_reg_ksm_data_contig(md, memh->address, 0, 0, 0,
+    status = uct_ib_mlx5_devx_reg_ksm_data_contig(md, memh->address,
+                                                  (uint64_t)memh->address, 0, 0,
                                                   "exported-key",
                                                   &memh->mrs[UCT_IB_MR_DEFAULT],
                                                   &cross_mr, &exported_lkey);


src/tools/perf/lib/libperf_thread.c
```diff
@@ -148,15 +148,22 @@ ucs_status_t ucx_perf_thread_spawn(ucx_perf_context_t *perf,
 
 ucs_status_t ucx_perf_allocators_init_thread(ucx_perf_context_t *perf)
 {
+    ucs_status_t status;
     /* New threads need explicit device association */
-    ucs_status_t status = perf->send_allocator->init(perf);
-    if (status != UCS_OK) {
-        return status;
+    if (perf->send_allocator != NULL) {
```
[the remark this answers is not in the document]
tvegas1: if it has to be null we could use template always returning `UCS_OK` instead.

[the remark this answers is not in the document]
roiedanino: @yosefe When using RDMA memory:
> Fix in perftest avoids a segfault when using multiple threads with RDMA memory

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -237,20 +237,22 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_addr(
...
-    uint64_t ksm_iova;
+    uint64_t ksm_iova, ksm_iova_base;
     size_t ksm_length;
     int list_size;
 
     /* FW requires indirect atomic MR address and length to be aligned
      * to max supported atomic argument size */
     ksm_address = ucs_align_down_pow2((uint64_t)address, UCT_IB_MD_MAX_MR_SIZE);
-    ksm_iova    = ksm_address + iova_offset;
+    ksm_iova_base = memh->dm != NULL ? (uint64_t)memh->address : ksm_address;
```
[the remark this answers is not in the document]
roiedanino: When using RDMA memory (dm != NULL), we want to map the reserved address range (memh->address) to 0 which is the actual RDMA base address for it.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -237,20 +237,22 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_addr(
  * that were registered in single-threaded mode.
  */
 static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_contig(
```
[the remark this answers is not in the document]
roiedanino: Will change it as well, thanks

src/tools/perf/lib/libperf_thread.c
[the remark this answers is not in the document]
yosefe: > @yosefe When using RDMA memory:
> 
> > Fix in perftest avoids a segfault when using multiple threads with RDMA memory

Is this a PR/commit? can you pls point to it?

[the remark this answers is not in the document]
roiedanino: 2 commits caused different issues here:
6fbfdea4f89d7cc09e7ba6ed1971329c68ba00f7 - assumed `iova` is always `= address + offset` while when using memic the base address should be 0. 

fc85a133f3f84feda80d33ef6a5ce1bb52916e1c - creating all self_eps with the same worker address (issue was identified only when using multiple threads obviously).

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -432,16 +446,17 @@ uct_ib_mlx5_devx_reg_ksm_data(uct_ib_mlx5_md_t *md,
...
     if (memh->super.flags & UCT_IB_MEM_MULTITHREADED) {
         return uct_ib_mlx5_devx_reg_ksm_data_mt(md, address,
                                                 (uint64_t)address + iova_offset,
                                                 atomic, mkey_index, reason,
                                                 mr->ksm_data, mr_p, mkey);
     } else {
-        return uct_ib_mlx5_devx_reg_ksm_data_contig(md, address, iova_offset,
-                                                    atomic, mkey_index, reason,
-                                                    mr, mr_p, mkey);
+        return uct_ib_mlx5_devx_reg_ksm_data_contig(md, address, iova, atomic,
```
yosefe, line 457: shouldn't we pass iova to uct_ib_mlx5_devx_reg_ksm_data_mt (line 452) as well? instead of `(uint64_t)address + iova_offset` ?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -276,6 +274,22 @@ uct_ib_mlx5_devx_memh_base_address(const uct_ib_mlx5_devx_mem_t *memh)
...
+
+#if HAVE_IBV_DM
+    if (memh->dm != NULL) {
+        /* Device memory memory key is zero based */
+        return (uint64_t)memh->address + iova_offset;
+    }
+#endif
+
+    address = (uint64_t)uct_ib_mlx5_devx_memh_base_address(memh);
+    return ucs_align_down_pow2(address, UCT_IB_MD_MAX_MR_SIZE) + iova_offset;
```
yosefe, line 290: uct_ib_mlx5_devx_reg_ksm_data_contig should adjust the iova it receives according to alignment - instead of doing it here.

src/tools/perf/lib/libperf_thread.c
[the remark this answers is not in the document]
yosefe: > 2 commits caused different issues here: [6fbfdea](https://github.com/openucx/ucx/commit/6fbfdea4f89d7cc09e7ba6ed1971329c68ba00f7) - assumed `iova` is always `= address + offset` while when using memic the base address should be 0.
> 
> [fc85a13](https://github.com/openucx/ucx/commit/fc85a133f3f84feda80d33ef6a5ce1bb52916e1c) - creating all self_eps with the same worker address (issue was identified only when using multiple threads obviously).

1. According to https://github.com/openucx/ucx/blob/8668850ca49311024949c793b18ea5ec24b96987/src/tools/perf/lib/libperf.c#L2074-L2079, send_allocator and recv_allocator cannot be NULL during the test. So why need to check for NULL here?

2. How this segfault was not caught in CI? Maybe we need to add multi-threaded perftest?

src/tools/perf/lib/libperf.c
```diff
@@ -1520,26 +1520,32 @@ static ucs_status_t ucp_perf_test_flush_workers(ucx_perf_context_t *perf)
...
+        if (status != UCS_OK) {
+            if (i > 0) {
+                goto err_destroy_eps;
+            }
+            return status;
+        }
+
+        ep_params.field_mask = UCP_EP_PARAM_FIELD_REMOTE_ADDRESS;
+        ep_params.address    = worker_attr.address;
+
```
yosefe, line 1548: remove space line

src/tools/perf/lib/libperf.c
```diff
@@ -1520,26 +1520,32 @@ static ucs_status_t ucp_perf_test_flush_workers(ucx_perf_context_t *perf)
...
-
-    ep_params.field_mask = UCP_EP_PARAM_FIELD_REMOTE_ADDRESS;
-    ep_params.address    = worker_attr.address;
-
     for (i = 0; i < perf->params.thread_count; ++i) {
         thread_perf = &perf->ucp.tctx[i].perf;
+        worker      = thread_perf->ucp.worker;
+        /* coverity[use_after_free] */
+        status      = UCX_PERF_VERBOSE(error, &perf->params, ucp_worker_query,
+                                       worker, &worker_attr);
```
yosefe, line 1538: So the problem here was that ep from all threads were connected to the "main" worker and not the "thread" worker? How is this fix related to MEMIC?

src/tools/perf/lib/libperf.c
```diff
@@ -1520,26 +1520,32 @@ static ucs_status_t ucp_perf_test_flush_workers(ucx_perf_context_t *perf)
...
+        if (status != UCS_OK) {
+            if (i > 0) {
+                goto err_destroy_eps;
```
yosefe, line 1541: maybe we can always `goto err_destroy_eps` in case of error?

src/tools/perf/lib/libperf.c
```diff
@@ -1557,6 +1563,8 @@ static ucs_status_t ucp_perf_test_setup_self_endpoints(ucx_perf_context_t *perf)
         if (status != UCS_OK) {
             goto err_destroy_eps;
         }
+
+        ucp_worker_release_address(worker, worker_attr.address);
```
yosefe, line 1567: should we set address to NULL after we release it, to avoid double-free in line 1580?
(Maybe Coverity warning was justified?)

roiedanino: Sounds good

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
roiedanino: I think not, because data_contig is aligning down the base address while it was not the logic in `uct_ib_mlx5_devx_reg_ksm_data_mt`

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
roiedanino: So we will have to pass `iova_offset` separately to ksm_data_contig

src/tools/perf/lib/libperf.c
yosefe: maybe it has to be a separate fix in another PR?
does it have some dependency with MEMIC fix?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: instead, maybe we could add the difference between address and ksm_address (i.e how much alignment has moved the address) to iova argument?

roiedanino:
@yosefe 
https://github.com/openucx/ucx/pull/9743#discussion_r1530249046

This logic explains it:
```c
ucs_status_t ucx_perf_allocators_init(ucx_perf_context_t *perf,
                                      const ucx_perf_params_t *params)
{
    if ((params->send_mem_type == UCS_MEMORY_TYPE_RDMA) ||
        (params->recv_mem_type == UCS_MEMORY_TYPE_RDMA)) {
        return UCS_OK;
    }
```
We need the allocators initialization also in UCP for the CUDA / ROCM context. the multithreading logic is shared with the UCT perftests which are using the allocators in runtime, I think we can't avoid that null check.

src/tools/perf/lib/libperf.c
roiedanino: We can split it but we won't be able to test memic with multithreading using just one of those fixes, what about separating this PR into 2 different commits when squashing?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -276,6 +274,22 @@ uct_ib_mlx5_devx_memh_base_address(const uct_ib_mlx5_devx_mem_t *memh)
...
 
+static uint64_t uct_ib_mlx5_devx_memh_iova(const uct_ib_mlx5_devx_mem_t *memh,
+                                           uint32_t iova_offset)
+{
+    uint64_t address;
+
+#if HAVE_IBV_DM
+    if (memh->dm != NULL) {
+        /* Device memory memory key is zero based */
+        return (uint64_t)memh->address + iova_offset;
```
ivankochin, line 285: Should we align to `memh->address` to `UCT_IB_MD_MAX_MR_SIZE`  in that case?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: After one of the recent patches it does (but it aligns to `chunk_size` not to `UCT_IB_MD_MAX_MR_SIZE`):

https://github.com/openucx/ucx/blob/master/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c#L162

src/tools/perf/lib/libperf.c
roiedanino: Will make this PR the memic fix and open a new PR for the perftest

roiedanino: https://github.com/openucx/ucx/pull/9766

yosefe (commented):
can we add a gtest to expose the bug?

roiedanino:
> can we add a gtest to expose the bug?

We can create the test but it will only fail CI when the new FW will be used in CI

yosefe:
> > can we add a gtest to expose the bug?
> 
> We can create the test but it will only fail CI when the new FW will be used in CI

Let's add it (van test on rock cluster in meantime)

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
roiedanino: No

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -241,20 +241,18 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_addr(
...
     size_t mr_length = mr->super.ib->length;
     uint64_t ksm_address;
-    uint64_t ksm_iova;
     size_t ksm_length;
     int list_size;
 
     /* FW requires indirect atomic MR address and length to be aligned
      * to max supported atomic argument size */
     ksm_address = ucs_align_down_pow2((uint64_t)address, UCT_IB_MD_MAX_MR_SIZE);
-    ksm_iova    = ksm_address + iova_offset;
```
yosefe, line 257: `ksm_iova = ksm_address - address + iova;`

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
roiedanino: Recent update should be good for both cases

yosefe (approved):

ivankochin:
> > > can we add a gtest to expose the bug?
> > 
> > 
> > We can create the test but it will only fail CI when the new FW will be used in CI
> 
> Let's add it (can test on rock cluster in meantime)

What about adding gtest for that fix? If I understand the quoted discussion correctly, it should be added, isn't it?

yosefe:
> > > > can we add a gtest to expose the bug?
> > > 
> > > 
> > > We can create the test but it will only fail CI when the new FW will be used in CI
> > 
> > 
> > Let's add it (can test on rock cluster in meantime)
> 
> What about adding gtest for that fix? If I understand the quoted discussion correctly, it should be added, isn't it?

We decided to add tests in another PR since this fix is needed for the release

PR #9745, opened 2024-03-12
UCT/UCP: UMR approach for XGVMI

iyastreb:
## What
UMR approach for XGVMI.
This part contains all the core components of the UMR approach:
 - UMR primitives: CQ, QP, init, destroy
 - UMR mkey pool for the exporter part
 - UMR hash map for the importer side

This changeset triggers UMR workflow:
In the exporter (host process), in `uct_ib_mlx5_devx_reg_exported_key`:
 - Find existing UMR mkey in the pool, or create a new one
 - Bind passed UMR with given MR (if MR length is below 2GB and not MT)
 - Otherwise (large or MT MR) fallback to KSM registration
 - Mark UMR mkey tag (8 LBS) so that we distinguish it on the importer side

On the importer side (DPU process), when receiving new rkey (`uct_ib_mlx5_devx_mem_attach`):
 - If passed mkey is UMR key, then lookup alias in the UMR hash map
 - If alias is found, just use it and don't create a new one
 - Otherwise create a new alias, and save it into the UMR hash map (if passed mkey is UMR one)

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -395,8 +395,8 @@ uct_ib_md_mlx5_devx_mr_lru_push(uct_ib_mlx5_md_t *md, uint32_t rkey, void *mr)
               kh_size(&md->lru_rkeys.hash), rkey, mr);
 
     if ((++md->lru_rkeys.count % md->super.config.max_idle_rkey_count) == 0) {
-        /* modulo prime number to avoid resonance with mkey */
-        md->mkey_tag = (md->mkey_tag + 1) % 251;
+        /* Increment mkey tag in order to mitigate the rkey collision risk */
+        md->mkey_tag = (md->mkey_tag + 1) % UCT_IB_MLX5_MKEY_TAG_MAX;
     }
 
     return UCS_OK;
@@ -895,6 +895,508 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
     return UCS_OK;
 }
 
+static ucs_status_t uct_ib_mlx5_devx_umr_create_cq(uct_ib_mlx5_md_t *md)
+{
+    md->umr.cq = ibv_create_cq(md->super.dev.ibv_context, 1, NULL, NULL, 0);
+    if (NULL == md->umr.cq) {
+        ucs_error("%s: ibv_create_cq() failed to create UMR CQ: %m",
+                  uct_ib_mlx5_dev_name(md));
+        return UCS_ERR_IO_ERROR;
+    }
+
+    ucs_trace("%s: created UMR CQ %p", uct_ib_mlx5_dev_name(md), md->umr.cq);
+    return UCS_OK;
+}
+
+static ucs_status_t uct_ib_mlx5_devx_umr_create_qp(uct_ib_mlx5_md_t *md)
+{
+    struct mlx5dv_qp_init_attr mlx5_qp_attr = {};
+    struct ibv_qp_init_attr_ex qp_attr_ex   = {};
+
+    qp_attr_ex.send_cq             = md->umr.cq;
+    qp_attr_ex.recv_cq             = md->umr.cq;
+    qp_attr_ex.srq                 = NULL;
+    qp_attr_ex.cap.max_send_wr     = 1;
+    qp_attr_ex.cap.max_recv_wr     = 1;
+    qp_attr_ex.cap.max_send_sge    = 1;
+    qp_attr_ex.cap.max_recv_sge    = 1;
+    qp_attr_ex.cap.max_inline_data = sizeof(struct mlx5_wqe_umr_klm_seg);
+    qp_attr_ex.qp_type             = IBV_QPT_RC;
+    qp_attr_ex.comp_mask           = IBV_QP_INIT_ATTR_SEND_OPS_FLAGS |
+                                     IBV_QP_INIT_ATTR_PD;
+    qp_attr_ex.pd                  = md->super.pd;
+    qp_attr_ex.send_ops_flags      = IBV_QP_EX_WITH_SEND;
+
+    mlx5_qp_attr.comp_mask         = MLX5DV_QP_INIT_ATTR_MASK_SEND_OPS_FLAGS;
+    mlx5_qp_attr.send_ops_flags    = MLX5DV_QP_EX_WITH_MR_LIST |
+                                     MLX5DV_QP_EX_WITH_MR_INTERLEAVED;
+
+    md->umr.qp = mlx5dv_create_qp(md->super.dev.ibv_context, &qp_attr_ex,
+                                  &mlx5_qp_attr);
+    if (NULL == md->umr.qp) {
+        ucs_error("%s: mlx5dv_create_qp() failed to create UMR QP: %m",
+                  uct_ib_mlx5_dev_name(md));
+        return UCS_ERR_IO_ERROR;
+    }
+
+    ucs_trace("%s: created UMR QP QPN 0x%x", uct_ib_mlx5_dev_name(md),
+              md->umr.qp->qp_num);
+    return UCS_OK;
+}
+
+static ucs_status_t uct_ib_mlx5_devx_umr_modify_qp(uct_ib_mlx5_md_t *md)
+{
+    uct_ib_device_t *ibdev     = &md->super.dev;
+    struct ibv_qp_attr qp_attr = {};
+    uint8_t port_num;
+    struct ibv_port_attr *port_attr;
+    int attr_mask;
+    int ret;
+
+    port_num  = ibdev->first_port;
+    port_attr = uct_ib_device_port_attr(ibdev, port_num);
+
+    /* Modify QP to INIT state */
+    attr_mask               = IBV_QP_STATE |
+                              IBV_QP_PKEY_INDEX |
+                              IBV_QP_PORT |
+                              IBV_QP_ACCESS_FLAGS;
+    qp_attr.qp_state        = IBV_QPS_INIT;
+    qp_attr.pkey_index      = 0;
+    qp_attr.port_num        = port_num;
+    qp_attr.qp_access_flags = UCT_IB_MEM_ACCESS_FLAGS;
+
+    ret = ibv_modify_qp(md->umr.qp, &qp_attr, attr_mask);
+    if (ret) {
+        ucs_error("%s: ibv_modify_qp(UMR QP 0x%x) failed to modify to INIT: %m",
+                  uct_ib_device_name(ibdev), md->umr.qp->qp_num);
+        return UCS_ERR_IO_ERROR;
+    }
+
+    /* Modify to RTR */
+    attr_mask                  = IBV_QP_STATE |
+                                 IBV_QP_DEST_QPN |
+                                 IBV_QP_PATH_MTU |
+                                 IBV_QP_RQ_PSN |
+                                 IBV_QP_MIN_RNR_TIMER |
+                                 IBV_QP_MAX_DEST_RD_ATOMIC |
+                                 IBV_QP_AV;
+    qp_attr.qp_state           = IBV_QPS_RTR;
+    qp_attr.dest_qp_num        = md->umr.qp->qp_num;
+    qp_attr.path_mtu           = IBV_MTU_512;
+    qp_attr.rq_psn             = 0;
+    qp_attr.min_rnr_timer      = 7;
+    qp_attr.max_dest_rd_atomic = 1;
+    qp_attr.ah_attr.port_num   = port_num;
+    qp_attr.ah_attr.dlid       = port_attr->lid;
+    qp_attr.ah_attr.is_global  = 1;
+
+    if (UCS_OK != uct_ib_device_query_gid(ibdev, port_num,
+                                          UCT_IB_DEVICE_DEFAULT_GID_INDEX,
+                                          &qp_attr.ah_attr.grh.dgid,
+                                          UCS_LOG_LEVEL_ERROR)) {
+        return UCS_ERR_IO_ERROR;
+    }
+
+    ret = ibv_modify_qp(md->umr.qp, &qp_attr, attr_mask);
+    if (ret) {
+        ucs_error("%s: ibv_modify_qp(UMR QP 0x%x) failed to modify to RTR: %m",
+                  uct_ib_device_name(ibdev), md->umr.qp->qp_num);
+        return UCS_ERR_IO_ERROR;
+    }
+
+    /* Modify to RTS */
+    attr_mask             = IBV_QP_STATE |
+                            IBV_QP_SQ_PSN |
+                            IBV_QP_TIMEOUT |
+                            IBV_QP_RNR_RETRY |
+                            IBV_QP_RETRY_CNT |
+                            IBV_QP_MAX_QP_RD_ATOMIC;
+    qp_attr.qp_state      = IBV_QPS_RTS;
+    qp_attr.sq_psn        = 0;
+    qp_attr.timeout       = 7;
+    qp_attr.rnr_retry     = 7;
+    qp_attr.retry_cnt     = 7;
+    qp_attr.max_rd_atomic = 1;
+
+    ret = ibv_modify_qp(md->umr.qp, &qp_attr, attr_mask);
+    if (ret) {
+        ucs_error("%s: ibv_modify_qp(UMR QP 0x%x) failed to modify to RTS: %m",
+                  uct_ib_device_name(ibdev), md->umr.qp->qp_num);
+        return UCS_ERR_IO_ERROR;
+    }
+
+    ucs_trace("%s: initialized UMR QP 0x%x", uct_ib_device_name(ibdev),
+              md->umr.qp->qp_num);
+    return UCS_OK;
+}
+
+static ucs_status_t uct_ib_mlx5_devx_umr_init_export(uct_ib_mlx5_md_t *md)
+{
+    ucs_status_t status;
+
+    status = uct_ib_mlx5_devx_umr_create_cq(md);
+    if (UCS_OK != status) {
+        goto err;
+    }
+
+    status = uct_ib_mlx5_devx_umr_create_qp(md);
+    if (UCS_OK != status) {
+        goto err_destroy_cq;
+    }
+
+    status = uct_ib_mlx5_devx_umr_modify_qp(md);
+    if (UCS_OK != status) {
+        goto err_destroy_qp;
+    }
+
+    /* Initialize indirect UMR mkey pool */
+    ucs_list_head_init(&md->umr.mkey_pool);
+
+    return UCS_OK;
+
+err_destroy_qp:
+    uct_ib_destroy_qp(md->umr.qp);
+err_destroy_cq:
+    ibv_destroy_cq(md->umr.cq);
+err:
+    return status;
+}
+
+static void
+uct_ib_mlx5_devx_umr_mkey_alias_destroy(uct_ib_mlx5_md_t *md,
+                                        uct_ib_mlx5_devx_umr_alias_t *umr_alias)
+{
+    ucs_status_t status;
+
+    ucs_trace("%s: destroy " UCT_IB_MLX5_UMR_ALIAS_FMT,
+              uct_ib_mlx5_dev_name(md), UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias));
+
+    status = uct_ib_mlx5_devx_obj_destroy(umr_alias->cross_mr, "MKEY_ALIAS");
+    if (UCS_OK != status) {
+        ucs_warn("%s: uct_ib_mlx5_devx_obj_destroy(" UCT_IB_MLX5_UMR_ALIAS_FMT
+                 ") failed with error '%s': %m", uct_ib_mlx5_dev_name(md),
+                 UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias),
+                 ucs_status_string(status));
+    }
+}
+
+static void
+uct_ib_mlx5_devx_umr_mkey_destroy(uct_ib_mlx5_md_t *md,
+                                  uct_ib_mlx5_devx_umr_mkey_t *umr_mkey)
+{
+    int ret;
+
+    ucs_trace("%s: destroy " UCT_IB_MLX5_UMR_MKEY_FMT, uct_ib_mlx5_dev_name(md),
+              UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey));
+
+    ret = mlx5dv_destroy_mkey(umr_mkey->mkey);
+    if (0 != ret) {
+        ucs_warn("%s: mlx5dv_destroy_mkey("UCT_IB_MLX5_UMR_MKEY_FMT") returned "
+                 "%d: %m", uct_ib_mlx5_dev_name(md),
+                 UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey), ret);
+    }
+
+    ucs_free(umr_mkey);
+}
+
+static uct_ib_mlx5_devx_umr_mkey_t *
+uct_ib_mlx5_devx_umr_mkey_create(uct_ib_mlx5_md_t *md)
+{
+    struct mlx5dv_mkey_init_attr mkey_init_attr = {
+        .pd           = md->super.pd,
+        .create_flags = MLX5DV_MKEY_INIT_ATTR_FLAGS_INDIRECT,
+        .max_entries  = 1
+    };
+    uct_ib_mlx5_devx_umr_mkey_t *umr_mkey;
+    ucs_status_t status;
+
+    /* Lazy init UMR related objects for exporter */
+    if (NULL == md->umr.qp) {
+        status = uct_ib_mlx5_devx_umr_init_export(md);
+        if (UCS_OK != status) {
+            return NULL;
+        }
+    }
+
+    umr_mkey = ucs_malloc(sizeof(*umr_mkey), "uct_ib_mlx5_devx_umr_mkey_t");
+    if (NULL == umr_mkey) {
+        ucs_error("%s: failed to allocate UMR mkey", uct_ib_mlx5_dev_name(md));
+        return NULL;
+    }
+
+    umr_mkey->mkey = mlx5dv_create_mkey(&mkey_init_attr);
+    if (NULL == umr_mkey->mkey) {
+        ucs_error("%s: mlx5dv_create_mkey() failed to create UMR mkey: %m",
+                  uct_ib_mlx5_dev_name(md));
+        ucs_free(umr_mkey);
+        return NULL;
+    }
+
+    /* Mark mkey as UMR to distinguish it on the importer side */
+    umr_mkey->mkey->lkey |= UCT_IB_MLX5_MKEY_TAG_UMR;
+    umr_mkey->mkey->rkey |= UCT_IB_MLX5_MKEY_TAG_UMR;
+
+    status = uct_ib_mlx5_devx_allow_xgvmi_access(md, umr_mkey->mkey->lkey, 1);
+    if (status != UCS_OK) {
+        /* Reset XGVMI capability flag */
+        md->flags &= ~UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI;
+        uct_ib_mlx5_devx_umr_mkey_destroy(md, umr_mkey);
+        return NULL;
+    }
+
+    ucs_trace("%s: created " UCT_IB_MLX5_UMR_MKEY_FMT,
+              uct_ib_mlx5_dev_name(md), UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey));
+    return umr_mkey;
+}
+
+static void uct_ib_mlx5_devx_umr_cleanup(uct_ib_mlx5_md_t *md)
+{
+    uct_ib_mlx5_devx_umr_alias_t mkey_alias;
+    uct_ib_mlx5_devx_umr_mkey_t *item, *tmp;
+    int ret;
+
+    /* Destroy UMR mkey hash if present */
+    if (NULL != md->umr.mkey_hash) {
+        ucs_trace("%s: destroy UMR mkey hash with %d elements",
+                  uct_ib_mlx5_dev_name(md), kh_size(md->umr.mkey_hash));
+
+        kh_foreach_value(md->umr.mkey_hash, mkey_alias, {
+            uct_ib_mlx5_devx_umr_mkey_alias_destroy(md, &mkey_alias);
+        });
+
+        kh_destroy(umr_mkey_map, md->umr.mkey_hash);
+    }
+
+    /* Destroy UMR mkey pool if not empty */
+    if (!ucs_list_is_empty(&md->umr.mkey_pool)) {
+        ucs_trace("%s: destroy UMR mkey pool with %lu elements",
+                  uct_ib_mlx5_dev_name(md),
+                  ucs_list_length(&md->umr.mkey_pool));
+
+        ucs_list_for_each_safe(item, tmp, &md->umr.mkey_pool, super) {
+            ucs_list_del(&item->super);
+            uct_ib_mlx5_devx_umr_mkey_destroy(md, item);
+        }
+    }
+
+    if (NULL != md->umr.qp) {
+        ret = ibv_destroy_qp(md->umr.qp);
+        if (ret != 0) {
+            ucs_warn("%s: ibv_destroy_qp(UMR QP) returned %d: %m",
+                     uct_ib_mlx5_dev_name(md), ret);
+        }
+    }
+
+    if (NULL != md->umr.cq) {
+        uct_ib_destroy_cq(md->umr.cq, uct_ib_mlx5_dev_name(md));
+    }
+}
+
+static ucs_status_t
+uct_ib_mlx5_devx_umr_post_sync(uct_ib_mlx5_md_t *md, struct ibv_send_wr *wr,
+                               uct_ib_mlx5_devx_umr_mkey_t *umr_mkey,
+                               const char *desc)
+{
+    ucs_time_t start_time = ucs_get_time();
+    struct ibv_wc wc;
+    struct ibv_send_wr *bad_wr;
+    int ret;
+
+    ret = ibv_post_send(md->umr.qp, wr, &bad_wr);
+    if (ret != 0) {
+        ucs_error("%s: ibv_post_send(UMR QP, %s, "UCT_IB_MLX5_UMR_MKEY_FMT") "
+                  "returned %d: %m", uct_ib_mlx5_dev_name(md), desc,
+                  UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey), ret);
+        return UCS_ERR_IO_ERROR;
+    }
+
+    /* Poll for completion with timeout */
+    while ((ret = ibv_poll_cq(md->umr.cq, 1, &wc)) == 0) {
+        if ((ucs_get_time() - start_time) > ucs_time_from_sec(30)) {
+            ucs_error("%s: ibv_poll_cq(UMR CQ, %s, "UCT_IB_MLX5_UMR_MKEY_FMT") "
+                      "timed out", uct_ib_mlx5_dev_name(md), desc,
+                      UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey));
+            return UCS_ERR_TIMED_OUT;
+        }
+    }
+
+    if ((ret < 0) || (wc.status != IBV_WC_SUCCESS)) {
+        ucs_error("%s: ibv_poll_cq(UMR CQ, %s, "UCT_IB_MLX5_UMR_MKEY_FMT") "
+                  "returned %d, status: %s", uct_ib_mlx5_dev_name(md), desc,
+                  UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey), ret,
+                  uct_ib_wc_status_str(wc.status));
+        return UCS_ERR_IO_ERROR;
+    }
+
+    return UCS_OK;
+}
+
+static ucs_status_t
+uct_ib_mlx5_devx_umr_mkey_invalidate(uct_ib_mlx5_md_t *md,
+                                     uct_ib_mlx5_devx_umr_mkey_t *umr_mkey)
+{
+    struct ibv_send_wr wr = {
+        .wr_id           = 0,
+        .next            = NULL,
+        .num_sge         = 0,
+        .opcode          = IBV_WR_LOCAL_INV,
+        .send_flags      = IBV_SEND_INLINE | IBV_SEND_SIGNALED,
+        .invalidate_rkey = umr_mkey->mkey->lkey
+    };
+    ucs_status_t status;
+
+    status = uct_ib_mlx5_devx_umr_post_sync(md, &wr, umr_mkey, "invalidation");
+    if (status != UCS_OK) {
+        uct_ib_mlx5_devx_umr_mkey_destroy(md, umr_mkey);
+        return status;
+    }
+
+    /* Add UMR mkey to mkey pool for reuse & cleanup */
+    ucs_list_add_head(&md->umr.mkey_pool, &umr_mkey->super);
+    ucs_trace("%s: put " UCT_IB_MLX5_UMR_MKEY_FMT " into pool (%ld)",
+              uct_ib_mlx5_dev_name(md), UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey),
+              ucs_list_length(&md->umr.mkey_pool));
+    return UCS_OK;
+}
+
+static uct_ib_mlx5_devx_umr_mkey_t *
+uct_ib_mlx5_devx_umr_mkey_pool_get(uct_ib_mlx5_md_t *md)
+{
+    uct_ib_mlx5_devx_umr_mkey_t *umr_mkey;
+
+    if (ucs_list_is_empty(&md->umr.mkey_pool)) {
+        umr_mkey = uct_ib_mlx5_devx_umr_mkey_create(md);
+    } else {
+        umr_mkey = ucs_list_extract_head(&md->umr.mkey_pool,
+                                         uct_ib_mlx5_devx_umr_mkey_t, super);
+        ucs_trace("%s: extracted from pool " UCT_IB_MLX5_UMR_MKEY_FMT,
+                  uct_ib_mlx5_dev_name(md), UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey));
+    }
+
+    return umr_mkey;
+}
+
+static uct_ib_mlx5_devx_umr_mkey_t *
+uct_ib_mlx5_devx_umr_mkey_bind(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mr_t *mr)
+{
+    struct ibv_mr *ib_mr  = mr->super.ib;
+    struct ibv_mw mw      = {
+        .rkey = 0,
+        .type = IBV_MW_TYPE_1
+    };
+    struct ibv_send_wr wr = {
+        .wr_id      = 0,
+        .next       = NULL,
+        .opcode     = IBV_WR_BIND_MW,
+        .send_flags = IBV_SEND_INLINE | IBV_SEND_SIGNALED,
+        .bind_mw    = {
+            .mw        = &mw,
+            .rkey      = 0,
+            .bind_info = {
+                .mr              = ib_mr,
+                .length          = ib_mr->length,
+                .addr            = (uint64_t)ib_mr->addr,
+                .mw_access_flags = UCT_IB_MLX5_MD_UMEM_ACCESS
+            }
+        }
+    };
+    ucs_status_t status;
+    uct_ib_mlx5_devx_umr_mkey_t *umr_mkey;
+
+    umr_mkey = uct_ib_mlx5_devx_umr_mkey_pool_get(md);
+    if (NULL == umr_mkey) {
+        return NULL;
+    }
+
+    /* Set UMR lkey to be registered with given MR */
+    mw.rkey         = umr_mkey->mkey->lkey;
+    wr.bind_mw.rkey = umr_mkey->mkey->lkey;
+
+    status = uct_ib_mlx5_devx_umr_post_sync(md, &wr, umr_mkey, "registration");
+    if (status != UCS_OK) {
+        uct_ib_mlx5_devx_umr_mkey_invalidate(md, umr_mkey);
+        return NULL;
+    }
+
+    ucs_trace("%s: registered addr %p length %zu lkey 0x%x to "
+              UCT_IB_MLX5_UMR_MKEY_FMT, uct_ib_mlx5_dev_name(md), ib_mr->addr,
+              ib_mr->length, ib_mr->lkey, UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey));
+    return umr_mkey;
+}
+
+static UCS_F_ALWAYS_INLINE uint64_t
+uct_ib_mlx5_devx_umr_mkey_hash_code(const uct_ib_md_packed_mkey_t *packed_mkey)
+{
+    return ((uint64_t)packed_mkey->vhca_id << 32) | packed_mkey->lkey;
+}
+
+/*
+ * Indicate that given mkey is an UMR mkey.
+ */
+static UCS_F_ALWAYS_INLINE int uct_ib_mlx5_devx_mkey_is_umr(uint32_t lkey)
+{
+    return UCT_IB_MLX5_MKEY_TAG_UMR == (lkey & 0xff);
+}
+
+static ucs_status_t
+uct_ib_mlx5_devx_umr_mkey_hash_find(uct_ib_mlx5_md_t *md,
+                                    const uct_ib_md_packed_mkey_t *packed_mkey,
+                                    uct_ib_mlx5_devx_umr_alias_t *umr_alias)
+{
+    uint64_t input_key;
+    khiter_t khiter;
+
+    if (NULL == md->umr.mkey_hash) {
+        return UCS_ERR_NO_ELEM;
+    }
+
+    input_key = uct_ib_mlx5_devx_umr_mkey_hash_code(packed_mkey);
+    khiter    = kh_get(umr_mkey_map, md->umr.mkey_hash, input_key);
+
+    if (khiter == kh_end(md->umr.mkey_hash)) {
+        return UCS_ERR_NO_ELEM;
+    }
+
+    *umr_alias = kh_val(md->umr.mkey_hash, khiter);
+    ucs_trace("%s: found " UCT_IB_MLX5_UMR_ALIAS_FMT " for index 0x%x in hash",
+              uct_ib_mlx5_dev_name(md), UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias),
+              uct_ib_mlx5_mkey_index(packed_mkey->lkey));
+    return UCS_OK;
+}
+
+static ucs_status_t
+uct_ib_mlx5_devx_umr_mkey_hash_put(uct_ib_mlx5_md_t *md,
+                                   const uct_ib_md_packed_mkey_t *packed_mkey,
+                                   const uct_ib_mlx5_devx_umr_alias_t *umr_alias)
+{
+    uint64_t input_key = uct_ib_mlx5_devx_umr_mkey_hash_code(packed_mkey);
+    int ret;
+    khiter_t khiter;
+
+    /* Lazy init UMR mkey alias hash map */
+    if (NULL == md->umr.mkey_hash) {
+        md->umr.mkey_hash = kh_init(umr_mkey_map);
+        if (NULL == md->umr.mkey_hash) {
+            return UCS_ERR_NO_MEMORY;
+        }
+    }
+
+    khiter = kh_put(umr_mkey_map, md->umr.mkey_hash, input_key, &ret);
+    if (ret == UCS_KH_PUT_FAILED) {
+        ucs_error("%s: failed to add " UCT_IB_MLX5_UMR_ALIAS_FMT " to hash map",
+                  uct_ib_mlx5_dev_name(md),
+                  UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias));
+        return UCS_ERR_NO_MEMORY;
+    }
+
+    kh_val(md->umr.mkey_hash, khiter) = *umr_alias;
+    ucs_trace("%s: added " UCT_IB_MLX5_UMR_ALIAS_FMT " for index 0x%x to hash",
+              uct_ib_mlx5_dev_name(md), UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias),
+              uct_ib_mlx5_mkey_index(packed_mkey->lkey));
+    return UCS_OK;
+}
+
 ucs_status_t
 uct_ib_mlx5_devx_mem_dereg(uct_md_h uct_md,
                            const uct_md_mem_dereg_params_t *params)
@@ -934,6 +1436,14 @@ uct_ib_mlx5_devx_mem_dereg(uct_md_h uct_md,
                   uct_ib_device_name(&md->super.dev), memh->cross_mr,
                   memh->exported_lkey);
         status = uct_ib_mlx5_devx_obj_destroy(memh->cross_mr, "CROSS_MR");
+        if (status != UCS_OK) {
+            return status;
+        }
+    }
+
+    if (memh->exported_umr_mkey != NULL) {
+        status = uct_ib_mlx5_devx_umr_mkey_invalidate(md,
+                                                      memh->exported_umr_mkey);
         if (status != UCS_OK) {
             return status;
         }
@@ -1793,6 +2303,12 @@ ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
     md->super.name       = UCT_IB_MD_NAME(mlx5);
     md->super.vhca_id    = vhca_id;
 
+    /* Zero init UMR related fields, will lazy init on first use */
+    md->umr.cq        = NULL;
+    md->umr.qp        = NULL;
+    md->umr.mkey_hash = NULL;
+    ucs_list_head_init(&md->umr.mkey_pool);
+
     ksm_atomic = 0;
     if (md->flags & UCT_IB_MLX5_MD_FLAG_KSM) {
         md->super.cap_flags |= UCT_MD_FLAG_INVALIDATE_RMA;
@@ -1868,6 +2384,7 @@ void uct_ib_mlx5_devx_md_close(uct_md_h tl_md)
     uct_ib_mlx5_md_buf_free(md, md->zero_buf, &md->zero_mem);
     ucs_mpool_cleanup(&md->dbrec_pool, 1);
     ucs_recursive_spinlock_destroy(&md->dbrec_lock);
+    uct_ib_mlx5_devx_umr_cleanup(md);
     uct_ib_md_close_common(&md->super);
     uct_ib_mlx5_devx_mr_lru_cleanup(md);
     uct_ib_md_free(&md->super);
@@ -1977,9 +2494,8 @@ err:
     return 0;
 }
 
-static ucs_status_t
+ucs_status_t
 uct_ib_mlx5_devx_allow_xgvmi_access(uct_ib_mlx5_md_t *md,
-                                    uct_ib_mlx5_devx_mem_t *memh,
                                     uint32_t exported_lkey, int silent)
 {
     char in[UCT_IB_MLX5DV_ST_SZ_BYTES(allow_other_vhca_access_in)]   = {0};
@@ -2072,7 +2588,7 @@ static ucs_status_t uct_ib_mlx5_devx_xgvmi_umem_mr(uct_ib_mlx5_md_t *md,
     exported_lkey = (UCT_IB_MLX5DV_GET(create_mkey_out, out, mkey_index) << 8) |
                     md->mkey_tag;
 
-    status = uct_ib_mlx5_devx_allow_xgvmi_access(md, memh, exported_lkey, 0);
+    status = uct_ib_mlx5_devx_allow_xgvmi_access(md, exported_lkey, 0);
     if (status != UCS_OK) {
         goto err_cross_mr_destroy;
     }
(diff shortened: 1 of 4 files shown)


src/uct/ib/base/ib_md.h
```diff
@@ -79,6 +79,7 @@ typedef struct uct_ib_md_ext_config {
     int                      prefer_nearest_device; /**< Give priority for near
                                                          device */
     int                      enable_indirect_atomic; /** Enable indirect atomic */
+    int                      max_inline_klm_list; /* Maximal length of inline KLM list */
```
brminich: i'd remove that, because it is always 1 currently

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,407 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+destroy_mkey:
+    if (mlx5dv_destroy_mkey(umr->umr_mkey) != 0) {
+        ucs_error("%s: failed to destroy UMR mkey %p lkey 0x%x: %m",
+                  uct_ib_device_name(&md->super.dev), umr->umr_mkey, mkey);
+    }
+    free(umr);
+    return UCS_ERR_IO_ERROR;
+}
+
+static ucs_status_t
```
brminich, line 1719: why need to move this func?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1730,6 +2054,18 @@ static ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
         goto err_lru_cleanup;
     }
 
+    if (md->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI) {
+        status = uct_ib_mlx5_devx_umr_init(md);
+        if (status != UCS_OK && status != UCS_ERR_UNSUPPORTED) {
```
brminich, line 2054:
```suggestion
        if ((status != UCS_OK) && (status != UCS_ERR_UNSUPPORTED)) {
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1730,6 +2054,18 @@ static ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
...
+            goto err_md_close;
+        }
+    } else {
+        md->umr_cq        = NULL;
+        md->umr_qp        = NULL;
+        md->umr_mkey_hash = NULL;
```
brminich, line 2060: should we also set the to null if `(md->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI)`, but  `uct_ib_mlx5_devx_umr_init` returned `UCS_ERR_UNSUPPORTED`

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
 #endif
 }
 
+static ucs_status_t uct_ib_mlx5_devx_umr_create_cq(uct_ib_mlx5_md_t *md)
+{
+    uct_ib_device_t *ibdev = &md->super.dev;
+
+    ucs_trace("%s: using ibv_context %p PD %p", uct_ib_device_name(ibdev),
+              ibdev->ibv_context, md->super.pd);
```
brminich, line 1416: i'd remove this trace or combine somehow with the one below (line 1424)

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+    return UCS_OK;
+}
+
+static ucs_status_t uct_ib_mlx5_devx_umr_create_qp(uct_ib_mlx5_md_t *md)
+{
+    uct_ib_device_t *ibdev                  = &md->super.dev;
+    struct mlx5dv_qp_init_attr mlx5_qp_attr = {};
+    struct ibv_qp_init_attr_ex qp_attr_ex   = {};
+
+    /* Create QP */
```
brminich, line 1434: pls remove this comment

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+    qp_attr_ex.recv_cq             = md->umr_cq;
+    qp_attr_ex.srq                 = NULL;
+    qp_attr_ex.cap.max_send_wr     = 1;
+    qp_attr_ex.cap.max_recv_wr     = 1;
+    qp_attr_ex.cap.max_send_sge    = 1;
+    qp_attr_ex.cap.max_recv_sge    = 1;
+    qp_attr_ex.cap.max_inline_data = sizeof(struct mlx5_wqe_umr_klm_seg);
+    qp_attr_ex.qp_type             = IBV_QPT_RC;
+    qp_attr_ex.comp_mask          |= (IBV_QP_INIT_ATTR_SEND_OPS_FLAGS |
+                                      IBV_QP_INIT_ATTR_PD);
```
brminich, line 1445: no need for `|`
```suggestion
    qp_attr_ex.comp_mask          = (IBV_QP_INIT_ATTR_SEND_OPS_FLAGS |
                                      IBV_QP_INIT_ATTR_PD);
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+    qp_attr_ex.pd                  = md->super.pd;
+    qp_attr_ex.send_ops_flags     |= IBV_QP_EX_WITH_SEND;
+
+    mlx5_qp_attr.comp_mask        |= MLX5DV_QP_INIT_ATTR_MASK_SEND_OPS_FLAGS;
```
brminich, line 1449:
```suggestion
    qp_attr_ex.send_ops_flags     = IBV_QP_EX_WITH_SEND;

    mlx5_qp_attr.comp_mask      = MLX5DV_QP_INIT_ATTR_MASK_SEND_OPS_FLAGS;
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+                                  &mlx5_qp_attr);
+    if (NULL == md->umr_qp) {
+        ucs_error("%s: failed to create UMR QP (qp): %m",
+                  uct_ib_device_name(ibdev));
+        return UCS_ERR_IO_ERROR;
+    }
+
+    ucs_trace("%s: created UMR QP (qp) QPN 0x%x", uct_ib_device_name(ibdev),
+              md->umr_qp->qp_num);
+    return UCS_OK;
```
brminich, line 1463:
```suggestion
    ucs_trace("%s: created UMR QP (qp) QPN 0x%x", uct_ib_device_name(ibdev),
              md->umr_qp->qp_num);
              
    return UCS_OK;
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+static ucs_status_t uct_ib_mlx5_devx_umr_modify_qp(uct_ib_mlx5_md_t *md)
+{
+    uct_ib_device_t *ibdev     = &md->super.dev;
+    struct ibv_qp_attr qp_attr = {};
+    uint8_t port_num;
+    struct ibv_port_attr *port_attr;
+    int attr_mask;
+    int ret;
+
+    /* TODO: fix port selection - active port should be used */
```
iyastreb, line 1475: I kept this comment for now just to bring your attention to the port selection.
In principle we should select the active port, but my understanding is:
- As per our discussion with Evgeny, each device has just a single port. There are no actual devices with multiple ports, basically for each port we create a separate device. Right?
- In other places for QP creation ([uct_rc_mlx5_devx_create_cmd_qp](https://github.com/openucx/ucx/blob/master/src/uct/ib/rc/accel/rc_mlx5_common.c#L269)) first port is taken
- Mikhail proposed to reuse API for active port finding, but it's tightly coupled with iface, and cannot be easily reused from device

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+        .next            = NULL,
+        .num_sge         = 0,
+        .opcode          = IBV_WR_LOCAL_INV,
+        .send_flags      = IBV_SEND_INLINE | IBV_SEND_SIGNALED,
+        .invalidate_rkey = mkey
+    };
+    int cq_ret;
+    struct ibv_send_wr *bad_wr;
+
+    if (ibv_post_send(md->umr_qp, &wr, &bad_wr) != UCS_OK) {
```
brminich, line 1687: need to check for `int` value defined by IB verbs rather than comparing with UCS codes (as it is not a UCX API function)

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+    ret = ibv_modify_qp(md->umr_qp, &qp_attr, attr_mask);
+    if (ret) {
+        ucs_error("%s: failed to modify UMR QP (0x%x) to RTS: %m",
+                  uct_ib_device_name(ibdev), md->umr_qp->qp_num);
+        return UCS_ERR_IO_ERROR;
+    }
+
+    ucs_trace("%s: initialized UMR QP 0x%x", uct_ib_device_name(ibdev),
+              md->umr_qp->qp_num);
+    return UCS_OK;
```
brminich, line 1551:
```suggestion
    ucs_trace("%s: initialized UMR QP 0x%x", uct_ib_device_name(ibdev),
              md->umr_qp->qp_num);
              
    return UCS_OK;
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+{
+    uct_ib_device_t *ibdev = &md->super.dev;
+    uct_ib_mlx5_devx_umr_alias_t mkey_alias;
+    uct_ib_mlx5_devx_umr_t *item, *tmp;
+    ucs_status_t status;
+
+    if (NULL != md->umr_mkey_hash) {
+        if (kh_size(md->umr_mkey_hash) != 0) {
+            ucs_info("%s: UMR mkey hash has %d elements",
+                     uct_ib_device_name(ibdev), kh_size(md->umr_mkey_hash));
```
brminich, line 1600: isn't it always supposed to be non-empty? why need this print?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+        }
+
+        /* Destroy umr_mkey_hash content */
+        kh_foreach_value(md->umr_mkey_hash, mkey_alias, {
+            /* Deregister each alias (value) */
+            ucs_trace("%s: deregistered UMR mkey alias %p lkey 0x%x",
+                    uct_ib_device_name(ibdev), mkey_alias.cross_mr,
+                    mkey_alias.lkey);
```
brminich, line 1608:
```suggestion
            ucs_trace("%s: deregistered UMR mkey alias %p lkey 0x%x",
                      uct_ib_device_name(ibdev), mkey_alias.cross_mr,
                      mkey_alias.lkey);
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+
+            status = uct_ib_mlx5_devx_obj_destroy(mkey_alias.cross_mr,
+                                                  "MKEY_ALIAS");
+            if (UCS_OK != status) {
+                ucs_error("%s: failed to deregister UMR alias %p lkey 0x%x: %m",
+                        uct_ib_device_name(ibdev), mkey_alias.cross_mr,
+                        mkey_alias.lkey);
```
brminich, line 1615:
```suggestion
                ucs_error("%s: failed to deregister UMR alias %p lkey 0x%x: %m",
                          uct_ib_device_name(ibdev), mkey_alias.cross_mr,
                          mkey_alias.lkey);
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+        });
+
+        kh_destroy(umr_mkey_map, md->umr_mkey_hash);
+        md->umr_mkey_hash = NULL;
+        ucs_trace("%s: destroyed umr_mkey_hash", uct_ib_device_name(ibdev));
+    }
+
+    if (!ucs_list_is_empty(&md->umr_mkey_pool)) {
+        ucs_info("%s: UMR mkey pool has %lu elements",
+                 uct_ib_device_name(ibdev),
```
brminich, line 1626: imo no need to print it

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1485,6 +1408,402 @@ static void uct_ib_mlx5dv_check_dm_ksm_reg(uct_ib_mlx5_md_t *md)
...
+    if (ibv_post_send(md->umr_qp, &wr, &bad_wr) != UCS_OK) {
+        ucs_error("%s: failed to invalidate indirect mkey 0x%x: %m",
+                  uct_ib_device_name(&md->super.dev), mkey);
+        goto destroy_mkey;
+    }
+
+    /* Poll for completion */
+    while ((cq_ret = ibv_poll_cq(md->umr_cq, 1, &wc)) == 0);
+
+    if (cq_ret < 0 || wc.status != IBV_WC_SUCCESS) {
```
brminich, line 1696:
```suggestion
    if ((cq_ret < 0) || (wc.status != IBV_WC_SUCCESS)) {
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -858,6 +858,313 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+    uct_ib_device_t *ibdev = &md->super.dev;
+    uct_ib_mlx5_devx_umr_alias_t mkey_alias;
+    uct_ib_mlx5_devx_umr_t *item, *tmp;
+    ucs_status_t status;
+
+    if (NULL != md->umr_mkey_hash) {
+        if (kh_size(md->umr_mkey_hash) != 0) {
+            ucs_info("%s: UMR mkey hash has %d elements",
+                     uct_ib_device_name(ibdev), kh_size(md->umr_mkey_hash));
+        }
```
brminich, line 1049: it will always be non-empty, no need to print it in info level (anyway you will print each entry in trace)

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -858,6 +858,313 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+        kh_destroy(umr_mkey_map, md->umr_mkey_hash);
+        md->umr_mkey_hash = NULL;
+        ucs_trace("%s: destroyed umr_mkey_hash", uct_ib_device_name(ibdev));
+    }
+
+    if (!ucs_list_is_empty(&md->umr_mkey_pool)) {
+        ucs_info("%s: UMR mkey pool has %lu elements",
+                 uct_ib_device_name(ibdev),
+                 ucs_list_length(&md->umr_mkey_pool));
+    }
```
brminich, line 1076: no need to print in info

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: No, it's not necessarily non-empty.
UMR hash map is populated only when process imports UMR key (e.g. DPU offloading daemon side).

But this hash map object is unconditionally created on all the MD devices supporting indirect xgvmi, even on the exporter side, even if not really using xgvmi feature at all. In all those cases, UMR hash map remains empty.

Why need this print - well, just for debugging purposes. When daemon is terminated it's useful to see the amount of mkeys imported.
Maybe log level can be lowered to debug or trace?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: fixed

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: for now I lowered log level to trace

iyastreb: Well, it was useful for me during development phase - so that I can quickly compare the amount of keys created on exporter and received on importer. If we have only log entry per element - it would be harder to compare.

If you don't like extra `if`, maybe we just print trace unconditionally:
```
    if (NULL != md->umr_mkey_hash) {
        ucs_trace("%s: UMR mkey hash has %d elements",
                  uct_ib_device_name(ibdev), kh_size(md->umr_mkey_hash));
```
??

brminich: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -858,6 +858,491 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+        if (NULL == umr) {
+            ucs_error("%s: failed to allocate UMR mkey",
+                      uct_ib_device_name(&md->super.dev));
+            return NULL;
+        }
+
+        if (uct_ib_mlx5_devx_umr_mkey_create(md, &umr->umr_mkey) != UCS_OK) {
+            ucs_free(umr);
+            return NULL;
+        }
```
brminich, line 1210: maybe we can allow xgvmi access here when creating a new UMR? then we would not need `from_pool` var

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -858,6 +858,491 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+uct_ib_mlx5_devx_umr_mkey_hash_put(uct_ib_mlx5_md_t *md,
+                                   const uct_ib_md_packed_mkey_t *packed_mkey,
+                                   const uct_ib_mlx5_devx_umr_alias_t *umr_alias)
+{
+    uint64_t input_key = uct_ib_mlx5_devx_umr_mkey_hash_code(packed_mkey);
+    int ret;
+    khint_t k;
+
+    k = kh_put(umr_mkey_map, md->umr_mkey_hash, input_key, &ret);
+    if (ret < 0) {
```
brminich, line 1331:
```suggestion
    if (ret == UCS_KH_PUT_FAILED) {
        ucs_assert(ret != UCS_KH_PUT_KEY_PRESENT);
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Ok, that looks good initially, but it's not that straightforward on practice.
The reason is that we have a fallback to umem registration in case xgvmi access can't be added:
```
    status = uct_ib_mlx5_devx_allow_xgvmi_access(md, memh, exported_lkey, 1);
    if (status == UCS_OK) {
        goto out;
    }

    ucs_debug("%s: indirect cross gvmi not supported, fallback to DEVX UMEM",
              uct_ib_device_name(&md->super.dev));
    mlx5dv_devx_obj_destroy(cross_mr);
    md->flags &= ~UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI;

out_umem_mr:
    return uct_ib_mlx5_devx_xgvmi_umem_mr(md, memh);
```
Currently ig xgvmi fails once, then capability is removed and it falls back to umem_mr forever.
Now, if we try to allow xgvmi when creating UMR key, I'm afraid we can break that logic.
I will experiment a bit more, so far I didn't find a safe option to implement it like you suggested

iyastreb: I believe there is no easy way to call `uct_ib_mlx5_devx_allow_xgvmi_access` during UMR creation, without changing existing behavior. There is one option however to get rid of `from_pool` variable: we can return `UCS_ERR_ALREADY_EXISTS` error code from the `bind` function when the resulting UMR mkey was present in the pool. Would it make sense to you?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -876,6 +876,501 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+                  "%zu lkey 0x%x: %m", uct_ib_device_name(ibdev),
+                  ib_mr->addr, ib_mr->length, ib_mr->lkey);
+
+        goto invalidate;
+    }
+
+    /* Poll for completion */
+    while ((ret = ibv_poll_cq(md->umr_cq, 1, &wc)) == 0);
+
+    if (ret < 0 || wc.status != IBV_WC_SUCCESS) {
```
brminich, line 1290:
```suggestion
    if ((ret < 0) || (wc.status != IBV_WC_SUCCESS)) {
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
brminich: is it possible to try to allow xgvmi access during umr creation and update some cap flag on md if fail? So that we do not need to pass `from-pool` thru the whole stack, but check some flag instead?

iyastreb: Ok, I think using some md flag is not a good option, for several reasons.
But elaborating this idea - there is better alternative actually! We can add `xgvmi_access` flag to the UMR mkey type itself:
```
typedef struct {
    ucs_list_link_t    super;
    struct mlx5dv_mkey *umr_mkey;
    /* Indicates whether XGVMI access had been allowed for this mkey */
    int     xgvmi_access;
} uct_ib_mlx5_devx_umr_t;
```
Sounds good?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Ok, I implemented that, and I like the approach.
In principle, if we want to save a bit of memory, we can re-use mkey tag (8 LSB of mkey) for this purpose.
Assuming we remove `is_xgvmi_allowed` field, we can mark mkey as UMR only after allowing xgvmi access:
```
static void uct_ib_mlx5_devx_umr_mkey_set_xgvmi_allowed(uct_ib_mlx5_devx_umr_t *umr)
    /* Set marker for UMR key to distinguish it on the importer side */
    umr->umr_mkey->lkey  |= UCT_IB_MLX5_MKEY_TAG_UMR;
    umr->umr_mkey->rkey  |= UCT_IB_MLX5_MKEY_TAG_UMR;
}

// And to detect whether xgvmi is already set we can reuse existing function:
static UCS_F_ALWAYS_INLINE int uct_ib_mlx5_devx_is_umr_mkey(uint32_t lkey)
{
    return (UCT_IB_MLX5_MKEY_TAG_UMR == (lkey & 0xFF));
}

Comment (iyastreb) [reply src/uct/ib/mlx5/dv/ib_mlx5dv_md.c]: I've just implemented and committed the aforementioned optimisation, please check

File: src/uct/ib/mlx5/dv/ib_mlx5dv_md.c [hunk 23]
```diff
@@ -876,6 +876,517 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+    while ((ret = ibv_poll_cq(md->umr_cq, 1, &wc)) == 0);
+
+    if ((ret < 0) || (wc.status != IBV_WC_SUCCESS)) {
+        ucs_error("%s: failed to poll CQ for UMR mkey registration for addr %p "
+                  " len %zu lkey 0x%x ret %d status %d",
+                  uct_ib_device_name(ibdev), ib_mr->addr, ib_mr->length,
+                  ib_mr->lkey, ret, wc.status);
+
+        goto invalidate;
+    }
```
brminich, line 1292: the sme sequence is in `uct_ib_mlx5_devx_umr_mkey_invalidate`, maybe a separate inline function is worth creating?

brminich (commented):
👍 besides a minor comment

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Thanks, totally makes sense, added

brminich (approved):

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+    return umr->umr_mkey->lkey;
+}
+
+static ucs_status_t
+uct_ib_mlx5_devx_umr_mkey_hash_find(uct_ib_mlx5_md_t *md,
+                                    const uct_ib_md_packed_mkey_t *packed_mkey,
+                                    uct_ib_mlx5_devx_umr_alias_t *umr_alias)
+{
+    uint64_t input_key = uct_ib_mlx5_devx_umr_mkey_hash_code(packed_mkey);
+    khint_t k          = kh_get(umr_mkey_map, md->umr_mkey_hash, input_key);
```
yosefe, line 1355: khiter_t khiter = kh_get(...

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+    if ((k == kh_end(md->umr_mkey_hash)) || (!kh_exist(md->umr_mkey_hash, k))) {
+        return UCS_ERR_NO_ELEM;
+    }
+
+    *umr_alias = kh_val(md->umr_mkey_hash, k);
+    ucs_trace("%s: found " UCT_IB_MLX5_UMR_ALIAS_FMT " for index 0x%x in hash",
+              uct_ib_device_name(&md->super.dev),
+              UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias),
+              uct_ib_mlx5_mkey_index(packed_mkey->lkey));
+
```
yosefe, line 1365: remove space

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+}
+
+static ucs_status_t
+uct_ib_mlx5_devx_umr_mkey_hash_put(uct_ib_mlx5_md_t *md,
+                                   const uct_ib_md_packed_mkey_t *packed_mkey,
+                                   const uct_ib_mlx5_devx_umr_alias_t *umr_alias)
+{
+    uint64_t input_key = uct_ib_mlx5_devx_umr_mkey_hash_code(packed_mkey);
+    int ret;
+    khint_t k;
```
yosefe, line 1376: khiter_t khiter;

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+    umr->umr_mkey->rkey |= UCT_IB_MLX5_MKEY_TAG_UMR;
+    return umr->umr_mkey->lkey;
+}
+
+static ucs_status_t
+uct_ib_mlx5_devx_umr_mkey_hash_find(uct_ib_mlx5_md_t *md,
+                                    const uct_ib_md_packed_mkey_t *packed_mkey,
+                                    uct_ib_mlx5_devx_umr_alias_t *umr_alias)
+{
+    uint64_t input_key = uct_ib_mlx5_devx_umr_mkey_hash_code(packed_mkey);
```
yosefe, line 1354: IMO it's a better usage to define umr_mkey_map hash table as uct_ib_md_packed_mkey_t key type (instead of uint64_t), instead of calling uct_ib_mlx5_devx_umr_mkey_hash_code directly outside the hash

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+    int ret;
+    khint_t k;
+
+    k = kh_put(umr_mkey_map, md->umr_mkey_hash, input_key, &ret);
+    if (ret == UCS_KH_PUT_FAILED) {
+        ucs_error("%s: failed to add " UCT_IB_MLX5_UMR_ALIAS_FMT " to hash map",
+                  uct_ib_device_name(&md->super.dev),
+                  UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias));
+
```
yosefe, line 1383: remove space

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+        return UCS_ERR_NO_MEMORY;
+    } else if (ret == UCS_KH_PUT_KEY_PRESENT) {
```
yosefe, line 1385: no need else

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+        ucs_error("%s: " UCT_IB_MLX5_UMR_ALIAS_FMT " already exists in hash "
+                  "map", uct_ib_device_name(&md->super.dev),
+                  UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias));
+
```
yosefe, line 1389: remove space

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+        return UCS_ERR_ALREADY_EXISTS;
+    }
+
+    kh_val(md->umr_mkey_hash, k) = *umr_alias;
+    ucs_trace("%s: added " UCT_IB_MLX5_UMR_ALIAS_FMT " for index 0x%x to hash "
+              "map", uct_ib_device_name(&md->super.dev),
+              UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias),
+              uct_ib_mlx5_mkey_index(packed_mkey->lkey));
+
```
yosefe, line 1398: remove space

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1767,6 +2278,18 @@ ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
         goto err_lru_cleanup;
     }
 
+    md->umr_cq        = NULL;
+    md->umr_qp        = NULL;
+    md->umr_mkey_hash = NULL;
+    ucs_list_head_init(&md->umr_mkey_pool);
+
+    if (md->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI) {
+        status = uct_ib_mlx5_devx_umr_init(md);
```
yosefe, line 2287: maybe initialize it on first usage?
if not lets move it to a later stage in md init

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2100,46 +2625,79 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_exported_key,
...
-    if (status != UCS_OK) {
-        return status;
+    /* Current bind impl (IBV_WR_BIND_MW) attaches a single KLM segment, so:
+     * - IBV_WR_BIND_MW supports the maximum region length of 2GB
+     * - IBV_WR_BIND_MW does not support multi-segment (multi-threaded) MRs
+     * For these use cases we fallback to KSM
+     */
+    if ((mr->super.ib->length > UCT_IB_MD_MAX_MR_SIZE) ||
+        (memh->super.flags & UCT_IB_MEM_MULTITHREADED)) {
+        status = uct_ib_mlx5_devx_reg_ksm_data(md, memh, UCT_IB_MR_DEFAULT, 0,
```
yosefe, line 2652: why changed uct_ib_mlx5_devx_reg_ksm_data_contig to uct_ib_mlx5_devx_reg_ksm_data?

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -247,13 +247,43 @@ typedef union {
...
+
+
 typedef struct {
     uct_ib_mem_t            super;
     void                    *address;
     struct mlx5dv_devx_obj  *atomic_dvmr;
     struct mlx5dv_devx_obj  *indirect_dvmr;
     struct mlx5dv_devx_umem *umem;
     struct mlx5dv_devx_obj  *cross_mr;
+    uct_ib_mlx5_devx_umr_t  *umr_mr;
```
yosefe, line 286: maybe it should be called exported_umr_mkey?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2100,46 +2625,79 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_exported_key,
...
+    } else {
+        status = uct_ib_mlx5_devx_umr_mkey_bind(md, mr, &umr);
+        if (status != UCS_OK) {
+            return status;
+        }
+
+        exported_lkey = umr->umr_mkey->lkey;
+
+        /* Return if mkey is already marked as XGVMI */
+        if (uct_ib_mlx5_devx_umr_mkey_is_xgvmi(exported_lkey)) {
```
yosefe, line 2667: it which case it's not XGVMI?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2100,46 +2625,79 @@ UCS_PROFILE_FUNC_ALWAYS(ucs_status_t, uct_ib_mlx5_devx_reg_exported_key,
...
 
     ucs_debug("%s: indirect cross gvmi not supported, fallback to DEVX UMEM",
               uct_ib_device_name(&md->super.dev));
-    mlx5dv_devx_obj_destroy(cross_mr);
+
+    if (cross_mr != NULL) {
+        mlx5dv_devx_obj_destroy(cross_mr);
+    }
+
+    if (umr != NULL) {
```
yosefe, line 2689: maybe use else if here?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2273,6 +2832,16 @@ ucs_status_t uct_ib_mlx5_devx_mem_attach(uct_md_h uct_md,
     hdr       = UCT_IB_MLX5DV_ADDR_OF(create_alias_obj_in, in, hdr);
     alias_ctx = UCT_IB_MLX5DV_ADDR_OF(create_alias_obj_in, in, alias_ctx);
 
+    /* If received mkey is UMR key, find alias in the UMR hash map */
+    if (uct_ib_mlx5_devx_umr_mkey_is_xgvmi(packed_mkey->lkey)) {
```
yosefe, line 2836: maybe this if condition is not needed, or move it inside uct_ib_mlx5_devx_umr_mkey_hash_find

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2290,11 +2859,11 @@ ucs_status_t uct_ib_mlx5_devx_mem_attach(uct_md_h uct_md,
...
-    memh->cross_mr = uct_ib_mlx5_devx_obj_create(md->super.dev.ibv_context, in,
-                                                 sizeof(in), out, sizeof(out),
-                                                 "MKEY_ALIAS",
-                                                 uct_md_attach_log_lvl(flags));
-    if (memh->cross_mr == NULL) {
+    umr_alias.cross_mr = uct_ib_mlx5_devx_obj_create(md->super.dev.ibv_context,
+                                                     in, sizeof(in), out,
+                                                     sizeof(out), "MKEY_ALIAS",
+                                                     uct_md_attach_log_lvl(flags));
+    if (umr_alias.cross_mr == NULL) {
```
yosefe, line 2866: don't we need to set memh->cross_mr?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2273,6 +2832,16 @@ ucs_status_t uct_ib_mlx5_devx_mem_attach(uct_md_h uct_md,
     hdr       = UCT_IB_MLX5DV_ADDR_OF(create_alias_obj_in, in, hdr);
     alias_ctx = UCT_IB_MLX5DV_ADDR_OF(create_alias_obj_in, in, alias_ctx);
 
+    /* If received mkey is UMR key, find alias in the UMR hash map */
+    if (uct_ib_mlx5_devx_umr_mkey_is_xgvmi(packed_mkey->lkey)) {
+        status = uct_ib_mlx5_devx_umr_mkey_hash_find(md, packed_mkey,
```
yosefe, line 2837: major: maybe the hash table could hold full memh as a value, instead of just lkey+cross_mr?

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -247,13 +247,43 @@ typedef union {
...
 
+/* Data structure to hold the UMR MR (on the host) item in the mkey pool */
+typedef struct {
+    ucs_list_link_t    super;
+    struct mlx5dv_mkey *umr_mkey;
+} uct_ib_mlx5_devx_umr_t;
+
+
+#define UCT_IB_MLX5_UMR_FMT "UMR mkey %p index 0x%x"
+#define UCT_IB_MLX5_UMR_ARG(umr) \
```
yosefe, line 258: _umr

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -247,13 +247,43 @@ typedef union {
...
+
+/* Data structure to hold the UMR mkey alias on the DPU item in the hash map */
+typedef struct {
+    struct mlx5dv_devx_obj *cross_mr;
+    uint32_t               lkey;
+} uct_ib_mlx5_devx_umr_alias_t;
+
+
+#define UCT_IB_MLX5_UMR_ALIAS_FMT "UMR mkey alias %p index 0x%x"
+#define UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias) \
```
yosefe, line 270: _umr_alias

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -330,6 +373,17 @@ typedef struct uct_ib_mlx5_md {
...
+    /* CQ for indirect (UMR) mkeys */
+    struct ibv_cq            *umr_cq;
+    /* QP for indirect (UMR) mkeys */
+    struct ibv_qp            *umr_qp;
+
+    /* Indirect (UMR) mkey pool (on the host) */
+    ucs_list_link_t          umr_mkey_pool;
+
+    /* Hash map of indirect mkey (from the host) to mkey alias (on the DPU) */
+    khash_t(umr_mkey_map)    *umr_mkey_hash;
```
yosefe, line 386: maybe put all umr stuff under sub-struct "umr" ?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+
+        ucs_list_for_each_safe(item, tmp, &md->umr_mkey_pool, super) {
+            ucs_list_del(&item->super);
+            uct_ib_mlx5_devx_umr_mkey_destroy(md, item);
+        }
+    }
+
+    if (NULL != md->umr_qp) {
+        if (ibv_destroy_qp(md->umr_qp) != 0) {
+            ucs_error("%s: failed to destroy UMR QP: %m",
```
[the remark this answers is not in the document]
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+                      uct_ib_device_name(ibdev));
+        }
+
+        md->umr_qp = NULL;
+    }
+
+    if (NULL != md->umr_cq) {
+        if (ibv_destroy_cq(md->umr_cq) != 0) {
+            ucs_error("%s: failed to destroy UMR CQ: %m",
```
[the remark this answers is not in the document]
iyastreb: ok

src/uct/ib/mlx5/ib_mlx5.h
iyastreb: renamed

src/uct/ib/mlx5/ib_mlx5.h
iyastreb: renamed

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: It's not XGVMI allowed when we are creating a new UMR mkey, not reusing existing mkey from the pool. For such a newly created key we need to explicitly allow XGVMI with `uct_ib_mlx5_devx_allow_xgvmi_access` call below.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: `uct_ib_mlx5_devx_reg_ksm_data` is a helper function that supports both contiguous and MT MRs, while `uct_ib_mlx5_devx_reg_ksm_data_contig` does support only the contiguous MRs. We discussed this with Ivan and Mikhail and agreed that we need to support MT memory handles here as well, like in other places

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: done

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: done

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+{
+    return (uint64_t)packed_mkey->vhca_id << 32 | packed_mkey->lkey;
+}
+
+/*
+ * Indicate that given mkey is an UMR mkey and XGVMI enabled.
+ */
+static UCS_F_ALWAYS_INLINE int uct_ib_mlx5_devx_umr_mkey_is_xgvmi(uint32_t lkey)
+{
+    return (UCT_IB_MLX5_MKEY_TAG_UMR == (lkey & 0xFF));
```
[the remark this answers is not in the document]
iyastreb: done

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+        uct_ib_mlx5_devx_umr_mkey_invalidate(md, *umr);
+        *umr = NULL;
+        return status;
+    }
+
+    ucs_trace("%s: registered addr %p length %zu lkey 0x%x to "
+              UCT_IB_MLX5_UMR_FMT, uct_ib_device_name(&md->super.dev),
+              ib_mr->addr, ib_mr->length, ib_mr->lkey,
+              UCT_IB_MLX5_UMR_ARG(*umr));
+
```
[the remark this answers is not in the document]
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+uct_ib_mlx5_devx_umr_mkey_pool_get(uct_ib_mlx5_md_t *md)
+{
+    uct_ib_mlx5_devx_umr_t *umr;
+
+    if (ucs_list_is_empty(&md->umr_mkey_pool)) {
+        umr = uct_ib_mlx5_devx_umr_mkey_create(md);
+    } else {
+        umr = ucs_list_extract_head(&md->umr_mkey_pool, uct_ib_mlx5_devx_umr_t,
+                                    super);
+
```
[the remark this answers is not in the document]
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+        uct_ib_mlx5_devx_umr_mkey_destroy(md, umr);
+        return status;
+    }
+
+    /* Add UMR mkey to mkey pool for reuse & cleanup */
+    ucs_list_add_head(&md->umr_mkey_pool, &umr->super);
+    ucs_trace("%s: put " UCT_IB_MLX5_UMR_FMT " into pool (%ld)",
+              uct_ib_device_name(&md->super.dev), UCT_IB_MLX5_UMR_ARG(umr),
+              ucs_list_length(&md->umr_mkey_pool));
+
```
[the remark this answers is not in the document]
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+    /* Poll for completion */
+    while ((ret = ibv_poll_cq(md->umr_cq, 1, &wc)) == 0);
+
+    if ((ret < 0) || (wc.status != IBV_WC_SUCCESS)) {
+        ucs_error("%s: failed to poll CQ operation '%s' for "
+                  UCT_IB_MLX5_UMR_FMT ": ret %d status %d",
+                  uct_ib_device_name(&md->super.dev), desc,
+                  UCT_IB_MLX5_UMR_ARG(umr), ret, wc.status);
+
```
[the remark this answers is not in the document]
iyastreb: done

iyastreb (commented):
Addressed the first part of the comments

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -247,13 +247,43 @@ typedef union {
 } uct_ib_mlx5_devx_mr_t;
 
 
+/* Data structure to hold the UMR MR (on the host) item in the mkey pool */
+typedef struct {
+    ucs_list_link_t    super;
+    struct mlx5dv_mkey *umr_mkey;
+} uct_ib_mlx5_devx_umr_t;
```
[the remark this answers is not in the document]
iyastreb: Ok, makes sense, renaming

src/uct/ib/mlx5/ib_mlx5.h
iyastreb: sounds good, renamed

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Yes, we need to set it only when received mkey is NOT UMR mkey.
Thanks for spotting
Before my change `memh->cross_mr` field was used:
1. on the exporter side - to keep the created exported mkey, which is destroyed on `mem_unmap`
2. on the importer side - to keep the imported mkey alias, which is again destroyed on `mem_unmap`

But with UMR approach the situation is different on the importer side.
The whole idea was to keep the created aliases in the the hash map, so that their lifetimes are much longer than the lifetimes of memory handles. So with UMR approach we don't bind memh with alias anymore, and this is on purpose.
However we still need to support non-UMR mkeys, and this is what I missed here. When imported mkey is NOT UMR one, we still need to bind alias with memh, to make sure it gets destroyed on memh unmap:
```
    /* If received mkey is UMR key, store alias in the UMR hash map */
    if (uct_ib_mlx5_devx_umr_mkey_is_xgvmi(packed_mkey->lkey)) {
        umr_alias.lkey = memh->super.lkey;

        status = uct_ib_mlx5_devx_umr_mkey_hash_put(md, packed_mkey, &umr_alias);
        if (UCS_OK != status) {
            goto err_cross_mr_destroy;
        }
    } else {
        /* Not UMR mkey: move cross_mr ownership to memh */
        memh->cross_mr = umr_alias.cross_mr;
    }

File: src/uct/ib/mlx5/dv/ib_mlx5dv_md.c (hunk 39 above)
Comment (iyastreb) [reply src/uct/ib/mlx5/dv/ib_mlx5dv_md.c:2837]: Mmm, to me this idea seems controversial to the original design.
My understanding is: we keep the hash map of mkey aliases on the importer side, so that we can reuse them with different memory handles. This is the principal use case for UMR approach, also called "temporary workflow" or "non-persistent workflow". We gain performance because we reuse the same UMR mkey on the exporter side, and corresponding mkey alias on the importer side for *different* memory handles.
Now if we bind alias to memh, that would break the design I'm afraid.

**Review by iyastreb — COMMENTED** [review]
Second batch of comments

File: src/uct/ib/mlx5/ib_mlx5.h (hunk 42 above)
Comment (iyastreb) [reply src/uct/ib/mlx5/ib_mlx5.h:386]: Ok, I've put it like:
```
    struct {
        /* CQ for indirect (UMR) mkeys */
        struct ibv_cq            *cq;
        /* QP for indirect (UMR) mkeys */
        struct ibv_qp            *qp;
        /* Indirect (UMR) mkey pool (on the host) */
        ucs_list_link_t          mkey_pool;
        /* Hash map of indirect mkey (from the host) to mkey alias (on the DPU) */
        khash_t(umr_mkey_map)    *mkey_hash;
    } umr;

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Ok, I see, I think we can remove this condition

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Lazy initialization is an interesting idea IMO.
Maybe we can elaborate it even a bit more. Basically on the exporter side we don't even need to allocate a aliases hash map, and on the importer side we don't really need QP,CQ, mkey pool. So in principle we can init just the exporter part of the first UMR mkey create. And we can init just the importer part on the first alias put into hash map. Let me implement that if you agree

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Ok, there are 2 ways to check it directly:
1. Out param `int *from_pool`, in the `uct_ib_mlx5_devx_umr_mkey_bind` function, indicating that returned mkey is from the pool. This was the initial impl
2. Add field `int xgvmi_allowed` to the `uct_ib_mlx5_devx_umr_mkey_t`. This is the most explicit way, we discussed it with Mikhail. The only downside is that it consumes a bit more memory comparing to current implicit impl. But ok

I guess you prefer the 2nd option, so I'm implementing it

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Yes, I see your point, but I'm not sure if it's simpler.
My concern is related to the lifetime and usage of those memory handles. So far the typical workflow on the daemon side for those memory handles is:
```
in INIT or SEND_REQ message handler:
    imported_memh = ucp_mem_map(buf); // internally creates new memh in mem_attach
    ...
    ucp_am_send_nbx(imported_memh); // send request to the receiver daemon
    ...
    ucp_mem_unmap(imported_memh); // cleanup the resource
```
With "regular workflow" we allocate memh once in INIT handler, and deallocate in FIN handler. But with "temporary workflow" we allocate on SEND_REQ, and deallocate once receive ACK from the peer daemon. So it's the user who controls the lifetime of those memory handles with API.
Now you propose to make those memory handles persistent. Ok, but then what happens if client destroys the handle with `ucp_mem_unmap` call? My assumption is that we need to destroy the memh object, and keep persistent only the alias mkey - to be reused in the future. This is current impl.
But your approach is also doable: we just need to check in `ucp_mem_unmap` call if memh being destroyed is present in the hash map, and if it is, then do nothing. Then this persistent memh can be destroyed only after removing from hash map. Is it what you are proposing?
Is it simpler? I don't know, it changes a bit the API semantics.
I'm not sure if there are other consequences of those persistent memory handles - like if you call `ucp_mem_map` several times, you always get the same object returned. It can be ambiguous.

IMO current design is simpler, as it does not change the semantics of existing stuff. Please let me know your thoughts

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+              ib_mr->addr, ib_mr->length, ib_mr->lkey,
+              UCT_IB_MLX5_UMR_ARG(*umr));
+
+    return UCS_OK;
+}
+
+static UCS_F_ALWAYS_INLINE uint64_t
+uct_ib_mlx5_devx_umr_mkey_hash_code(const uct_ib_md_packed_mkey_t *packed_mkey)
+{
+    return (uint64_t)packed_mkey->vhca_id << 32 | packed_mkey->lkey;
```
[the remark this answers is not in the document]
iyastreb: Ok for brackets
Regarding `|` - look, it seems this pattern is used everywhere in similar places, here are just a few examples from the same file:
```
return ((getpid() & 0xff) << 8) | UCT_IB_MD_INVALID_FLUSH_RKEY;
(UCT_IB_MLX5DV_GET(create_mkey_out, out, mkey_index) << 8) | md->mkey_tag;
(UCT_IB_MLX5DV_GET(create_alias_obj_out, out, hdr.obj_id) << 8) | md->mkey_tag;
```
So I would keep it the same for consistency

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -895,6 +895,510 @@ uct_ib_mlx5_devx_dereg_keys(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh)
...
+}
+
+static ucs_status_t
+uct_ib_mlx5_devx_umr_mkey_hash_find(uct_ib_mlx5_md_t *md,
+                                    const uct_ib_md_packed_mkey_t *packed_mkey,
+                                    uct_ib_mlx5_devx_umr_alias_t *umr_alias)
+{
+    uint64_t input_key = uct_ib_mlx5_devx_umr_mkey_hash_code(packed_mkey);
+    khint_t k          = kh_get(umr_mkey_map, md->umr_mkey_hash, input_key);
+    if ((k == kh_end(md->umr_mkey_hash)) || (!kh_exist(md->umr_mkey_hash, k))) {
```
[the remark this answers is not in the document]
iyastreb: 1. ok
2. removed, leftover from Yong POC
3. removed

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Do you mean that the whole `else if (PRESENT){ }` block is not needed?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: Yes, it looks like a good idea.
I implemented it in a test project, and I don't really like the outcome.
```
static inline uint64_t
uct_ib_mlx5_devx_umr_mkey_hash_code2(uct_ib_md_packed_mkey_t packed_mkey)
{
    return ((uint64_t)packed_mkey.vhca_id << 32) | packed_mkey.lkey;
}

static int uct_ib_mlx5_devx_umr_mkey_equal2(uct_ib_md_packed_mkey_t a,
                                            uct_ib_md_packed_mkey_t b)
{
    // return (a.vhca_id == b.vhca_id) && (a.lkey == b.lkey); <-- slow
    return memcmp(&a, &b, sizeof(a)) == 0; // is it always safe wrt padding?
}

KHASH_INIT(test_map2, uct_ib_md_packed_mkey_t, uint64_t, 1,
           uct_ib_mlx5_devx_umr_mkey_hash_code2, uct_ib_mlx5_devx_umr_mkey_equal2);
```
First, it's more verbose, cause you have to provide custom `hash_equals` function.
Then, lookup is a bit slower then with the original impl:
```
test_map1: 1138
test_map2: 1334
```
The reason is that equal function calls memcmp, which is slower then the plain integers comparison. So I would keep the original impl, which is shorter and faster.

FYI comparison test: /labhome/iyastrebov/ws/misc/hash.c

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: done

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: I refactored existing code to split the flows:
- Separate flow for KSM exported key
- Separate flow for UMR exported key - now we create valid mkey and allow xgvmi during creation.
- KSM or UMR may reset XGVMI capability flag if fail to allow xgvmi -> in this case we fallback to UMEM regsitration

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1023,195 +1018,224 @@ static ucs_status_t uct_ib_mlx5_devx_umr_modify_qp(uct_ib_mlx5_md_t *md)
...
+                                        uct_ib_mlx5_devx_umr_alias_t *umr_alias)
+{
+    ucs_status_t status;
+
+    ucs_trace("%s: destroy " UCT_IB_MLX5_UMR_ALIAS_FMT,
+              uct_ib_mlx5_dev_name(md), UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias));
+
+    status = uct_ib_mlx5_devx_obj_destroy(umr_alias->cross_mr, "MKEY_ALIAS");
+    if (UCS_OK != status) {
+        ucs_error("%s: failed to destroy " UCT_IB_MLX5_UMR_ALIAS_FMT ": %m",
```
yosefe, line 1062: ucs_warn

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1023,195 +1018,224 @@ static ucs_status_t uct_ib_mlx5_devx_umr_modify_qp(uct_ib_mlx5_md_t *md)
...
-
-    ucs_trace("%s: created " UCT_IB_MLX5_UMR_FMT,
-              uct_ib_device_name(&md->super.dev), UCT_IB_MLX5_UMR_ARG(umr));
-
-    return umr;
-}
-
-static void uct_ib_mlx5_devx_umr_mkey_destroy(uct_ib_mlx5_md_t *md,
-                                              uct_ib_mlx5_devx_umr_t *umr)
+static ucs_status_t uct_ib_mlx5_devx_umr_init_import(uct_ib_mlx5_md_t *md)
```
yosefe, line 1100: IMO this small wrapper is redundant

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1023,195 +1018,224 @@ static ucs_status_t uct_ib_mlx5_devx_umr_modify_qp(uct_ib_mlx5_md_t *md)
...
         });
 
-        kh_destroy(umr_mkey_map, md->umr_mkey_hash);
-        md->umr_mkey_hash = NULL;
+        kh_destroy(umr_mkey_map, md->umr.mkey_hash);
     }
 
     /* Destroy UMR mkey pool if not empty */
-    if (!ucs_list_is_empty(&md->umr_mkey_pool)) {
+    if (!ucs_list_is_empty(&md->umr.mkey_pool)) {
```
yosefe, line 1130: we can remove is_empty check and the debug trace, since having non-empty pool is not a leak

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1023,195 +1018,224 @@ static ucs_status_t uct_ib_mlx5_devx_umr_modify_qp(uct_ib_mlx5_md_t *md)
...
 
     if ((ret < 0) || (wc.status != IBV_WC_SUCCESS)) {
         ucs_error("%s: failed to poll CQ operation '%s' for "
-                  UCT_IB_MLX5_UMR_FMT ": ret %d status %d",
-                  uct_ib_device_name(&md->super.dev), desc,
-                  UCT_IB_MLX5_UMR_ARG(umr), ret, wc.status);
-
+                  UCT_IB_MLX5_UMR_MKEY_FMT ": ret %d status %d: %s",
+                  uct_ib_mlx5_dev_name(md), desc,
+                  UCT_IB_MLX5_UMR_MKEY_ARG(umr_mkey), ret, wc.status,
```
yosefe, line 1237: no need to print numeric status , string is enough

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2621,85 +2623,87 @@ static ucs_status_t uct_ib_mlx5_devx_xgvmi_umem_mr(uct_ib_mlx5_md_t *md,
...
+        } else {
+            status = uct_ib_mlx5_devx_xgvmi_umr_mr(md, memh);
+        }
 
-    if (umr != NULL) {
-        uct_ib_mlx5_devx_umr_mkey_destroy(md, umr);
-    }
+        /* If KSM or UMR implementation fail to enable XGVMI, this capability
+         * flag is removed by impl, and then we fallback to UMEM impl */
+        if (md->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI) {
```
yosefe, line 2697: it seems a bit unexpected check
maybe we could check (status != UCS_ERR_UNSUPPORTED) ?

iyastreb: Yes, I have already elaborated this idea.
In short, it's a bit more risky but doable, we just need to be careful, because UCS_ERR_UNSUPPORTED might be returned from other functions (not from "allow xgvmi").
For instance:
```
uct_ib_mlx5_devx_reg_xgvmi_ksm_mr
    uct_ib_mlx5_devx_reg_ksm_data
        uct_ib_mlx5_devx_reg_ksm_data_contig
            uct_ib_mlx5_devx_reg_ksm_data_addr
                uct_ib_mlx5_devx_reg_ksm: line 97: if (mr == NULL) return UCS_ERR_UNSUPPORTED
```
This we should not treat as XGVMI failure, at least it is not treated like that with current impl.
So we can implement it like that:
```
uct_ib_mlx5_devx_reg_xgvmi_ksm_mr()
{
    status = uct_ib_mlx5_devx_reg_ksm_data(); // This may result in UCS_ERR_UNSUPPORTED
    if (status != UCS_OK) {
        return UCS_ERR_IO_ERROR;   << return generic error, lose precise error
    }

    status = uct_ib_mlx5_devx_allow_xgvmi_access(md, exported_lkey, 1);
    if (status != UCS_OK) {
        return UCS_ERR_UNSUPPORTED;
    }
}
```
And then we need to change the `umr_bind` function to return ucs_status_t, and make sure we return UCS_ERR_UNSUPPORTED only in case of xgvmi failure.
It's a bit risky, because some future modification can break this logic. So I'm not sure this is a superior alternative..

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1097,58 +1063,39 @@ static ucs_status_t uct_ib_mlx5_devx_umr_init_export(uct_ib_mlx5_md_t *md)
...
 
-    if (NULL != md->umr.qp) {
-        ret = ibv_destroy_qp(md->umr.qp);
-        if (ret != 0) {
-            ucs_warn("%s: failed to destroy UMR QP, error %d: %m",
-                     uct_ib_mlx5_dev_name(md), ret);
-        }
+    ret = mlx5dv_destroy_mkey(umr_mkey->mkey);
+    if (0 != ret) {
+        ucs_warn("%s: failed to destroy " UCT_IB_MLX5_UMR_MKEY_FMT ": %m",
```
yosefe, line 1094: log that mlx5dv_destroy_mkey failed

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1023,195 +1018,224 @@ static ucs_status_t uct_ib_mlx5_devx_umr_modify_qp(uct_ib_mlx5_md_t *md)
...
-    uct_ib_mlx5_devx_umr_t *item, *tmp;
+    uct_ib_mlx5_devx_umr_mkey_t *item, *tmp;
+    int ret;
 
     /* Destroy UMR mkey hash if present */
-    if (NULL != md->umr_mkey_hash) {
+    if (NULL != md->umr.mkey_hash) {
         ucs_trace("%s: destroy UMR mkey hash with %d elements",
-                  uct_ib_device_name(ibdev), kh_size(md->umr_mkey_hash));
+                  uct_ib_mlx5_dev_name(md), kh_size(md->umr.mkey_hash));
```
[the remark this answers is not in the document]
yosefe: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2328,18 +2315,18 @@ ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
...
     md->umr.cq        = NULL;
     md->umr.qp        = NULL;
     md->umr.mkey_hash = NULL;
     ucs_list_head_init(&md->umr.mkey_pool);
 
+    /* Enable relaxed order only if we would be able to create an indirect key
+       (with offset) for strict order access */
+    uct_ib_md_parse_relaxed_order(&md->super, md_config, ksm_atomic);
+
+    uct_ib_mlx5_devx_init_flush_mr(md);
```
yosefe, line 2328: there was some confusion, when asked to move "md->umr.cq = NULL;" after line 2316 i meant the line before you applied more review comments. so it would be after "md->super.vhca_id = vhca_id;"

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: done, also for the function above

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok, got it, done

yosefe (approved):
pls squash :)

PR #9746, opened 2024-03-13
UCT/DC: Print error when max_rd_atomic is different on peers

brminich:
## What
RDMA_READ or/and atomic operations may fail if peers are using dcis configured with different values of max_rd_atomic. Since certain FW in CX-7 this value was increased from 16 up to 64. This PR checks whether just one of the peers is using old value (16) of max_rd_atomics and prints an error if it is different from the local value.
Later, when dynamic dci allocation is introduced, ep will create a new dci pool matching peers config rather than just printing an error.

### DIFF
--- a/src/uct/ib/dc/dc_mlx5.h
+++ b/src/uct/ib/dc/dc_mlx5.h
@@ -67,12 +67,13 @@ typedef struct uct_dc_mlx5_iface  uct_dc_mlx5_iface_t;
 
 
 typedef enum {
-    UCT_DC_MLX5_IFACE_ADDR_HW_TM      = UCS_BIT(0),
-    UCT_DC_MLX5_IFACE_ADDR_DC_V1      = UCS_BIT(1),
-    UCT_DC_MLX5_IFACE_ADDR_DC_V2      = UCS_BIT(2),
-    UCT_DC_MLX5_IFACE_ADDR_FLUSH_RKEY = UCS_BIT(3),
-    UCT_DC_MLX5_IFACE_ADDR_DC_VERS    = UCT_DC_MLX5_IFACE_ADDR_DC_V1 |
-                                        UCT_DC_MLX5_IFACE_ADDR_DC_V2
+    UCT_DC_MLX5_IFACE_ADDR_HW_TM            = UCS_BIT(0),
+    UCT_DC_MLX5_IFACE_ADDR_DC_V1            = UCS_BIT(1),
+    UCT_DC_MLX5_IFACE_ADDR_DC_V2            = UCS_BIT(2),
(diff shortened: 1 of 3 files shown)


src/uct/ib/dc/dc_mlx5.h
```diff
@@ -67,12 +68,13 @@ typedef struct uct_dc_mlx5_iface  uct_dc_mlx5_iface_t;
...
-    UCT_DC_MLX5_IFACE_ADDR_DC_V1      = UCS_BIT(1),
-    UCT_DC_MLX5_IFACE_ADDR_DC_V2      = UCS_BIT(2),
-    UCT_DC_MLX5_IFACE_ADDR_FLUSH_RKEY = UCS_BIT(3),
-    UCT_DC_MLX5_IFACE_ADDR_DC_VERS    = UCT_DC_MLX5_IFACE_ADDR_DC_V1 |
-                                        UCT_DC_MLX5_IFACE_ADDR_DC_V2
+    UCT_DC_MLX5_IFACE_ADDR_HW_TM                = UCS_BIT(0),
+    UCT_DC_MLX5_IFACE_ADDR_DC_V1                = UCS_BIT(1),
+    UCT_DC_MLX5_IFACE_ADDR_DC_V2                = UCS_BIT(2),
+    UCT_DC_MLX5_IFACE_ADDR_FLUSH_RKEY           = UCS_BIT(3),
+    UCT_DC_MLX5_IFACE_ADDR_MAX_RD_ATOMIC_LEGACY = UCS_BIT(4),
```
yosefe, line 75: maybe explicitly state what is the value?
e.g UCT_DC_MLX5_IFACE_ADDR_MAX_RD_ATOMIC_16
consider it can be changed by user config..

brminich: renaming is fine, but  what do you mean by user config?

yosefe: > renaming is fine, but what do you mean by user config?

That since it can be changed by user config, it can still be "legacy" (e.g old device) but with a value different that 16, or new device with a value of 16

yosefe: > > renaming is fine, but what do you mean by user config?
> 
> That since it can be changed by user config, it can still be "legacy" (e.g old device) but with a value different that 16, or new device with a value of 16

So we should rename "legacy" to 16

brminich: imo legacy is more self-explanatory, but can change to 16 instead.

brminich: fixed in place because patch is small

yosefe (approved):

PR #9748, opened 2024-03-13
UCP: Fix rndv ppln protocols selection

brminich:
## What
Do not select pipeline rndv protocols when sender buffer is not contig (in this case address is zero and there is no rkey in RTS header)

cheryy-picked from #9720 + added gtest

### DIFF
--- a/src/ucp/rndv/rndv_rtr.c
+++ b/src/ucp/rndv/rndv_rtr.c
@@ -400,7 +400,8 @@ ucp_proto_rndv_rtr_mtype_init(const ucp_proto_init_params_t *init_params)
     size_t frag_size;
     ucp_md_index_t md_index;
 
-    if (!ucp_proto_rndv_op_check(init_params, UCP_OP_ID_RNDV_RECV, 1)) {
+    if (!ucp_proto_rndv_op_check(init_params, UCP_OP_ID_RNDV_RECV, 1) ||
+        (init_params->rkey_cfg_index == UCP_WORKER_CFG_INDEX_NULL)) {
         return UCS_ERR_UNSUPPORTED;
     }
 


test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1848,7 +1849,9 @@ class test_ucp_am_nbx_rndv_memtype : public test_ucp_am_nbx_rndv {
 
 UCS_TEST_P(test_ucp_am_nbx_rndv_memtype, rndv)
 {
-    test_am_send_recv_memtype(64 * UCS_KBYTE);
+    ucp_context_config_t *cfg = &sender().worker()->context->config.ext;
+    size_t size               = cfg->rndv_frag_size[UCS_MEMORY_TYPE_HOST] + 1;
```
tvegas1, line 1853: for my understanding: part of the test is that we make `size` too big for lane/rkey to be used?

tvegas1 (approved):

test/gtest/ucp/test_ucp_am.cc
brminich: i just make sure that we are using more than 1 fragment when pipeline protocol is used (should be always selected for large cuda-managed transfers)

src/ucp/rndv/rndv_ppln.c
```diff
@@ -59,7 +59,8 @@ ucp_proto_rndv_ppln_init(const ucp_proto_init_params_t *init_params)
     if ((select_param->dt_class != UCP_DATATYPE_CONTIG) ||
         !ucp_proto_init_check_op(init_params, UCP_PROTO_RNDV_OP_ID_MASK) ||
         !ucp_proto_common_init_check_err_handling(&err_params) ||
-        ucp_proto_rndv_init_params_is_ppln_frag(init_params)) {
+        ucp_proto_rndv_init_params_is_ppln_frag(init_params) ||
+        (init_params->rkey_cfg_index == UCP_WORKER_CFG_INDEX_NULL)) {
```
yosefe, line 63: I would expect ppln protocol to not be selected because rndv/put/mtype and rndv/get/mtype required rkey

src/ucp/rndv/rndv_get.c
```diff
@@ -61,8 +61,8 @@ ucp_proto_rndv_get_common_init(const ucp_proto_init_params_t *init_params,
     };
 
     if ((init_params->select_param->dt_class != UCP_DATATYPE_CONTIG) ||
-        !ucp_proto_rndv_op_check(init_params, UCP_OP_ID_RNDV_RECV,
-                                 support_ppln)) {
+        !ucp_proto_rndv_op_rkey_check(init_params, UCP_OP_ID_RNDV_RECV,
```
yosefe, line 64: how did it work so far, without this additional check?

brminich: correct, missed this check https://github.com/openucx/ucx/blob/master/src/ucp/proto/proto_common.c#L504
so new check is needed for rtr_mtype only
updated in place

yosefe (approved):

yosefe:
per offline discussion, will separate the new test to another PR, since its failure could be unrelated

brminich:
> per offline discussion, will separate the new test to another PR, since its failure could be unrelated

fixed in place

yosefe (approved):

yosefe:
Generally speaking, we should enable rtr/mtype protocol to handle a sender selecting rndv/am protocol. since we don't have a guarantee the sender will not do it (if when its datatype is contig)

PR #9751, opened 2024-03-14
UCT/CUDA: Removed uct_ep_check support from CUDA_IPC

shasson5:
## What
uct_ep_check implementations were removed from CUDA_IPC.

## Why ?
uct_ep_check API is used for error handling mechanism.
As part of the lane selection process, we select a single keepalive lane, which is used to check remote EP connection status.
Only transports with one of the following caps can be used as keepalive lane:
1) CONNECT_TO_EP + EP_CHECK flag.
2) CONNECT_TO_IFACE + AM_BCOPY flags.
3)  CONNECT_TO_EP + EP_KEEPALIVE flags.
CUDA_IPC doesn't fulfill those conditions, so cannot be used as keepalive lane.

### DIFF
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
@@ -201,10 +201,9 @@ static ucs_status_t uct_cuda_ipc_iface_query(uct_iface_h tl_iface,
     iface_attr->ep_addr_len             = 0;
     iface_attr->max_conn_priv           = 0;
     iface_attr->cap.flags               = UCT_IFACE_FLAG_ERRHANDLE_PEER_FAILURE |
-                                          UCT_IFACE_FLAG_EP_CHECK               |
-                                          UCT_IFACE_FLAG_CONNECT_TO_IFACE       |
-                                          UCT_IFACE_FLAG_PENDING                |
-                                          UCT_IFACE_FLAG_GET_ZCOPY              |
+                                          UCT_IFACE_FLAG_CONNECT_TO_IFACE |
+                                          UCT_IFACE_FLAG_PENDING          |
+                                          UCT_IFACE_FLAG_GET_ZCOPY        |
                                           UCT_IFACE_FLAG_PUT_ZCOPY;
     iface_attr->cap.event_flags         = UCT_IFACE_FLAG_EVENT_SEND_COMP |
                                           UCT_IFACE_FLAG_EVENT_RECV      |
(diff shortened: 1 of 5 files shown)


src/uct/sm/mm/base/mm_ep.c
```diff
@@ -563,14 +563,3 @@ ucs_status_t uct_mm_ep_flush(uct_ep_h tl_ep, unsigned flags,
...
     return UCS_OK;
 }
-
-ucs_status_t
-uct_mm_ep_check(uct_ep_h tl_ep, unsigned flags, uct_completion_t *comp)
-{
-    uct_mm_ep_t *ep = ucs_derived_of(tl_ep, uct_mm_ep_t);
-
-    UCT_EP_KEEPALIVE_CHECK_PARAM(flags, comp);
-    uct_ep_keepalive_check(tl_ep, &ep->keepalive, ep->fifo_ctl->pid, flags,
```
rakhmets, line 573: I think we still may have an issue with calling `uct_ep_keepalive_init` inside mm endpoint initialization.
E.g. when PID namespace is not shared between two containers running on the same host.

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -200,11 +200,9 @@ static ucs_status_t uct_cuda_ipc_iface_query(uct_iface_h tl_iface,
     iface_attr->device_addr_len         = sizeof(uint64_t);
     iface_attr->ep_addr_len             = 0;
     iface_attr->max_conn_priv           = 0;
-    iface_attr->cap.flags               = UCT_IFACE_FLAG_ERRHANDLE_PEER_FAILURE |
```
brminich, line 203: i think removing `UCT_IFACE_FLAG_ERRHANDLE_PEER_FAILURE` cap means cuda_ipc can't be used for rndv whith client-server apps. Did you check it is still selected?

yosefe:
failure seems relevant:
```
2024-03-27T14:20:43.2436331Z [ RUN      ] test_obj_size.size <> <>
2024-03-27T14:20:43.2436761Z /__w/1/s/contrib/../test/gtest/common/test_obj_size.cc:80: Failure
2024-03-27T14:20:43.2437175Z Expected equality of these values:
2024-03-27T14:20:43.2437553Z   (size_t)24
2024-03-27T14:20:43.2437917Z     Which is: 24
2024-03-27T14:20:43.2438282Z   sizeof(uct_cuda_ipc_ep_t)
2024-03-27T14:20:43.2438629Z     Which is: 16
2024-03-27T14:20:43.2443667Z [  FAILED  ] test_obj_size.size, where TypeParam =  and GetParam() =  (0 ms)
```

yosefe (approved):

rakhmets:
Please update the title and the description.

rakhmets:
> Those transports (CMA,CUDA_IPC,MM) doesn't support CONNECT_TO_EP, so they cannot be used as keepalive lane.

Do we need any changes in CMA and MM transports?

PR #9752, opened 2024-03-15
TOOLS/PERF: CI test for perftest daemon

iyastreb:
## What
First CI test for perftest_daemon.

## Why ?
XGVMI functionality must be covered by CI tests to prevent regressions.

## How ?
Some refactoring was needed to allow daemon to support perftest batch mode:
- Moved command line arguments from `TEST_PARAMS_ARGS` (i.e. per-test options) to global options. Reason: these daemon offloading options are immutable during the process lifetime, and should not be changed on per-test basis. It's pretty much like existing options for receiver IP and port.
- Moved daemon-related config params into a separate struct
- Added `daemon-keep-alive` option, to keep daemon alive across multiple test runs
- Support "reconnection" for daemon clients. That means that TCP connection between host and daemon is not reestablished, but it's possible to reconnect to daemon from another EP (INIT and FIN are reentrant)
- Support for batch mode: unless `daemon-keep-alive` option is specified - kill daemon at the end of the batch run

### DIFF
--- a/contrib/test_jenkins.sh
+++ b/contrib/test_jenkins.sh
@@ -561,6 +561,79 @@ run_ucx_perftest() {
 	fi
 }
 
+start_perftest_daemon() {
+	daemon_exe="$1"
+
+	# Find daemon port
+	dmn_port="$server_port"
+	step_server_port
+
+	# We explicitly disable cuda transport, because it's not p2p and therefore
+	# imposes INVALIDATE_RMA flag for all lanes (@see ucp_ep_config_init).
+	# However invalidating of imported rkeys is not supported by the daemon.
+	# TODO: Should we support invalidation of imported keys?
+	# Normally cuda_ipc cannot be used to communicate between host and DPU,
+	# unless we run both processes on host for testing purposes.
+
+	# Mandatory options to run the daemon
+	# - UCX_RNDV_THRESH=0 is needed to enforce RNDV protocol usage, as it's the
+	# only supported protocol between host and DPU
+	# - UCX_RNDV_SCHEME=put_zcopy. On low buffer dimensions (below ~8KB) UCX
+	# prefers bcopy over zero-copy, but bcopy workflow is not supported by DPU
+	# daemon. The workaround is to force rendezvous scheme to use zero-copy.
+	# get_zcopy option is not good enough, because bcopy is still selected for
+	# tiny messages (below 64 bytes)
+	dmn_env="UCX_TLS=^cuda UCX_TCP_CM_REUSEADDR=y UCX_RNDV_THRESH=0 UCX_RNDV_SCHEME=put_zcopy"
+
+	# Run the daemon
+	env $dmn_env $daemon_exe -p $dmn_port &
+
+	# Return the daemon pid and port
+	eval "$2=$!"
+	eval "$3=$dmn_port"
+}
+
+#
+# Run UCX performance daemon test
+#
+run_ucx_perftest_with_daemon() {
+	ucx_inst_ptest=$ucx_inst/share/ucx/perftest
+
+	ucx_perftest="$ucx_inst/bin/ucx_perftest"
+	ucx_perftest_daemon="$ucx_inst/bin/ucx_perftest_daemon"
+	ucp_test_args="-b $ucx_inst_ptest/test_types_ucp_daemon"
+
+	devices="$(get_ib_bf_devices $(get_active_ib_devices))"
+	for ucx_dev in $devices
+	do
+		echo "==== Running ucx_perftest over a daemon on $ucx_dev ===="
+		ip_addr=$(get_rdma_device_ip_addr $ucx_dev)
+		if [ -z "$ip_addr" ]
+		then
+			echo "Cannot find IPv4 address for device $ucx_dev"
+			continue
+		fi
+
+		export UCX_NET_DEVICES=$ucx_dev
+
+		# Start client and server daemons
+		start_perftest_daemon $ucx_perftest_daemon server_dmn_pid server_dmn_port
+		start_perftest_daemon $ucx_perftest_daemon client_dmn_pid client_dmn_port
+
+		ucp_client_args="-g $ip_addr:$client_dmn_port -G $ip_addr:$server_dmn_port $(hostname)"
+
+		run_client_server_app "$ucx_perftest" "$ucp_test_args" "$ucp_client_args" 0 0
+
+		kill ${client_dmn_pid} || true # ignore failure
+		kill ${server_dmn_pid} || true # ignore failure
+		wait $client_dmn_pid || true
+		wait $server_dmn_pid || true
+
+		unset UCX_TLS
+		unset UCX_NET_DEVICES
+	done
+}
+
 #
 # Test malloc hooks with mpi
 #
@@ -1099,6 +1172,7 @@ run_tests() {
 	do_distributed_task 2 4 test_init_mt
 	do_distributed_task 3 4 run_ucp_client_server
 	do_distributed_task 0 4 test_no_cuda_context
+	do_distributed_task 1 4 run_ucx_perftest_with_daemon
 
 	# long devel tests
 	do_distributed_task 0 4 run_ucp_hello
--- a/contrib/ucx_perftest_config/test_types_ucp_daemon
+++ b/contrib/ucx_perftest_config/test_types_ucp_daemon
@@ -1,0 +1,5 @@
+#
+# UCP daemon basic
+#
+ucp_daemon_bw_1                        -t ucp_am_bw -w 0 -n 10000 -s 1
+ucp_daemon_bw_8k                       -t ucp_am_bw -w 0 -n 10000 -s 8192
--- a/src/tools/perf/api/libperf.h
+++ b/src/tools/perf/api/libperf.h
@@ -118,8 +118,6 @@ enum {
  *      |                 |---------PEER_TX(AM)------->|                 |
  *      |                 |----------(TM msg)--------->|                 |
  *      |<---SEND_CMPL----|                            |---RECV_CMPL---->|
- *      |                 |                            |                 |
- *      |-------FIN------>|                            |<------FIN-------|
  *
  * (AM) used only when AM communication is configured
  * (TM) used only when tag matching communication is configured
@@ -144,11 +142,9 @@ enum {
      * unsolicited in case of AM communication. For tag matching transport it \
      * confirms the completion of RECV_REQ request */ \
     _macro(UCP_PERF_DAEMON_AM_ID_RECV_CMPL, UCP_PERF_AM_ID + 5) \
-    /* Host->DPU daemon finalize message */ \
-    _macro(UCP_PERF_DAEMON_AM_ID_FIN,       UCP_PERF_AM_ID + 6) \
     /* Sender DPU->Receiver DPU, message to establish communication between \
      * DPU peers */ \
-    _macro(UCP_PERF_DAEMON_AM_ID_PEER_INIT, UCP_PERF_AM_ID + 7)
+    _macro(UCP_PERF_DAEMON_AM_ID_PEER_INIT, UCP_PERF_AM_ID + 6)
 
 
 #define UCP_PERF_DAEMON_ENUMIFY(ID, VALUE) ID = VALUE,
--- a/src/tools/perf/perftest_daemon.c
+++ b/src/tools/perf/perftest_daemon.c
@@ -62,10 +62,10 @@ ucp_perf_daemon_ep_close(ucp_perf_daemon_context_t *ctx, ucp_ep_h ep)
         return;
     }
 
+    ucs_debug("closing ep %p", ep);
     param.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS;
     param.flags        = UCP_EP_CLOSE_FLAG_FORCE;
     close_req          = ucp_ep_close_nbx(ep, &param);
-
     if (UCS_PTR_IS_PTR(close_req)) {
         do {
             ucp_worker_progress(ctx->worker);
@@ -81,9 +81,39 @@ ucp_perf_daemon_ep_close(ucp_perf_daemon_context_t *ctx, ucp_ep_h ep)
     }
 }
 
+static void ucp_perf_daemon_cleanup_connections(ucp_perf_daemon_context_t *ctx)
+{
+    ucs_trace("cleaning up daemon connections");
+
+    if (NULL != ctx->send_memh) {
+        /* coverity[check_return] */
+        ucp_mem_unmap(ctx->context, ctx->send_memh);
+        ctx->send_memh = NULL;
+    }
+
+    if (NULL != ctx->recv_memh) {
+        /* coverity[check_return] */
+        ucp_mem_unmap(ctx->context, ctx->recv_memh);
+        ctx->recv_memh = NULL;
+    }
+
+    ucp_perf_daemon_ep_close(ctx, ctx->host_ep);
+    ctx->host_ep = NULL;
+}
+
 static void ucp_perf_daemon_err_cb(void *arg, ucp_ep_h ep, ucs_status_t status)
 {
-    terminated = 1;
+    ucp_perf_daemon_context_t *ctx = arg;
+
+    if (ep == ctx->host_ep) {
+        ucs_debug("ep %p: host error handler called with status %s", ep,
+                  ucs_status_string(status));
+        ucp_perf_daemon_cleanup_connections(arg);
+    } else {
+        ucs_error("ep %p: peer error handler called with status %s", ep,
+                  ucs_status_string(status));
+        terminated = 1;
+    }
 }
 
 static void
@@ -107,6 +137,8 @@ ucp_perf_daemon_server_conn_handle_cb(ucp_conn_request_h conn_request,
     if (status != UCS_OK) {
         ucs_error("failed to create an endpoint on the daemon: %s",
                   ucs_status_string(status));
+    } else {
+        ucs_debug("new ep %p accepted", ep);
     }
 }
 
@@ -129,7 +161,7 @@ ucp_perf_daemon_set_am_recv_handler(ucp_perf_daemon_context_t *ctx,
 
     status = ucp_worker_set_am_recv_handler(ctx->worker, &param);
     if (UCS_OK != status) {
-        ucs_error("failed to set am recv handler: %s",
+        ucs_error("ucp_worker_set_am_recv_handler() failed with error: %s",
                   ucs_status_string(status));
         return status;
     }
@@ -138,7 +170,8 @@ ucp_perf_daemon_set_am_recv_handler(ucp_perf_daemon_context_t *ctx,
 }
 
 static ucs_status_t
-ucp_perf_daemon_send_am_eager_reply(ucp_ep_h ep, ucp_perf_daemon_am_id_t am_id)
+ucp_perf_daemon_send_am_eager_reply(ucp_ep_h ep, ucp_perf_daemon_am_id_t am_id,
+                                    ucs_status_t status)
 {
     ucp_request_param_t param = {};
     ucs_status_ptr_t sreq;
@@ -146,11 +179,12 @@ ucp_perf_daemon_send_am_eager_reply(ucp_ep_h ep, ucp_perf_daemon_am_id_t am_id)
     param.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS;
     param.flags        = UCP_AM_SEND_FLAG_REPLY | UCP_AM_SEND_FLAG_EAGER;
 
-    sreq = ucp_am_send_nbx(ep, am_id, NULL, 0ul, NULL, 0ul, &param);
+    sreq = ucp_am_send_nbx(ep, am_id, NULL, 0ul, &status, sizeof(status),
+                           &param);
     if (UCS_PTR_IS_PTR(sreq)) {
         ucp_request_free(sreq);
     } else if (UCS_PTR_IS_ERR(sreq)) {
-        ucs_error("failed to send am id %u: %s", am_id,
+        ucs_error("ucp_am_send_nbx(AM_ID=%u) failed with: %s", am_id,
                   ucs_status_string(UCS_PTR_STATUS(sreq)));
         return UCS_PTR_STATUS(sreq);
     }
@@ -163,11 +197,15 @@ ucp_perf_daemon_send_cb(void *request, ucs_status_t status, void *user_data)
 {
     ucp_perf_daemon_context_t *ctx = user_data;
 
-    if (ucs_likely(status == UCS_OK)) {
-        ucp_perf_daemon_send_am_eager_reply(ctx->host_ep,
-                                            UCP_PERF_DAEMON_AM_ID_SEND_CMPL);
+    if (status != UCS_OK) {
+        ucs_error("SEND operation failed: %s", ucs_status_string(status));
+    } else {
+        ucs_trace_data("SEND completed");
     }
 
+    ucp_perf_daemon_send_am_eager_reply(ctx->host_ep,
+                                        UCP_PERF_DAEMON_AM_ID_SEND_CMPL,
+                                        status);
     ucp_request_free(request);
 }
 
@@ -198,11 +236,15 @@ static void ucp_perf_daemon_recv_cb(void *request, ucs_status_t status,
 {
     ucp_perf_daemon_context_t *ctx = user_data;
 
-    if (ucs_likely(status == UCS_OK)) {
-        ucp_perf_daemon_send_am_eager_reply(ctx->host_ep,
-                                            UCP_PERF_DAEMON_AM_ID_RECV_CMPL);
+    if (status != UCS_OK) {
+        ucs_error("RECV operation failed: %s", ucs_status_string(status));
+    } else {
+        ucs_trace_data("RECV completed");
     }
 
+    ucp_perf_daemon_send_am_eager_reply(ctx->host_ep,
+                                        UCP_PERF_DAEMON_AM_ID_RECV_CMPL,
+                                        status);
     ucp_request_free(request);
 }
 
@@ -228,13 +270,15 @@ ucp_perf_daemon_create_peer_ep(ucp_perf_daemon_context_t *ctx,
 
     status = ucp_ep_create(ctx->worker, &ep_params, &ctx->peer_ep);
     if (status != UCS_OK) {
-        ucs_error("daemon failed to create an endpoint: %s",
-                  ucs_status_string(status));
+        ucs_error("ucp_ep_create() failed with: %s", ucs_status_string(status));
         return status;
+    } else {
+        ucs_debug("peer ep %p created", ctx->peer_ep);
     }
 
     return ucp_perf_daemon_send_am_eager_reply(ctx->peer_ep,
-                                               UCP_PERF_DAEMON_AM_ID_PEER_INIT);
+                                               UCP_PERF_DAEMON_AM_ID_PEER_INIT,
+                                               UCS_OK);
 }
 
 static void ucp_perf_daemon_check_am_msg(const ucp_am_recv_param_t *param,
@@ -302,11 +346,10 @@ ucp_perf_daemon_init_handler(void *arg, const void *header,
     ucp_perf_daemon_check_am_msg(param, header_length, 1, 0,
                                  UCP_PERF_DAEMON_AM_ID_INIT);
 
-    /* Peer EP must not be initialized on receiving HOST_INIT message.
+    /* Host EP must not be initialized on receiving HOST_INIT message.
      * Otherwise it means that daemon was already initialized from host. For now
-     * we don't support multiple host processes, or connection reestablishment
-     */
-    if (ctx->peer_ep != NULL) {
+     * we don't support multiple host processes */
+    if (ctx->host_ep != NULL) {
         ucs_error("duplicate daemon init req");
         return UCS_ERR_ALREADY_EXISTS;
     }
@@ -314,7 +357,9 @@ ucp_perf_daemon_init_handler(void *arg, const void *header,
     ctx->host_ep = param->reply_ep;
 
     ucp_perf_unpack_array(&ptr, &value_length, &value);
-    if (value_length != 0) {
+    /* Peer EP is persistent during daemon lifetime, create it only if it does
+     * not exist yet */
+    if ((value_length != 0) && (ctx->peer_ep == NULL)) {
         status = ucp_perf_daemon_create_peer_ep(ctx, value, value_length);
         if (status != UCS_OK) {
             return status;
@@ -354,18 +399,6 @@ ucp_perf_daemon_send_handler(void *arg, const void *header,
 }
 
 static ucs_status_t
-ucp_perf_daemon_fin_handler(void *arg, const void *header, size_t header_length,
-                            void *data, size_t length,
-                            const ucp_am_recv_param_t *param)
-{
-    ucp_perf_daemon_check_am_msg(param, header_length, 0, 0,
-                                 UCP_PERF_DAEMON_AM_ID_FIN);
-
-    terminated = 1;
-    return UCS_OK;
-}
-
-static ucs_status_t
 ucp_perf_daemon_peer_init_handler(void *arg, const void *header,
                                   size_t header_length, void *data,
                                   size_t length,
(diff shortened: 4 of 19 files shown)


src/tools/perf/perftest_params.c
```diff
@@ -670,6 +684,23 @@ ucs_status_t parse_opts(struct perftest_context *ctx, int mpi_initialized,
             ctx->mpi = atoi(optarg) && mpi_initialized;
             break;
 #endif
+        case 'g': /* handles daemon-local long option as well */
+            status = ucs_sock_ipportstr_to_sockaddr(
+                optarg, DEFAULT_DAEMON_PORT, &ctx->dmn_info.dmn_local_addr);
+            if (status != UCS_OK) {
+                goto err;
```
tvegas1, line 691: add log detailing failure?

src/tools/perf/perftest_params.c
```diff
@@ -670,6 +684,23 @@ ucs_status_t parse_opts(struct perftest_context *ctx, int mpi_initialized,
...
+            }
+            break;
+        case 'G': /* handles daemon-remote long option as well */
+            status = ucs_sock_ipportstr_to_sockaddr(
+                optarg, DEFAULT_DAEMON_PORT, &ctx->dmn_info.dmn_remote_addr);
+            if (status != UCS_OK) {
+                goto err;
```
tvegas1, line 698: add log detailing failure?

src/tools/perf/perftest_run.c
```diff
@@ -262,10 +262,28 @@ static ucs_status_t read_batch_file(FILE *batch_file, const char *file_name,
...
 
+static size_t get_test_count(FILE *batch_file)
+{
+    size_t test_count = 0;
+
+#define MAX_ARG_SIZE 2048
+    char buf[MAX_ARG_SIZE];
+
+    while (fgets(buf, sizeof(buf) - 1, batch_file) != NULL) {
+        if ((buf[0] != '#') && (strtok(buf, " \t\n\r"))) {
```
tvegas1, line 273: what is the format of batch_file? with lines containing only one space, we would count them as test? same with leading spaces before `#`?

src/tools/perf/api/libperf.h
```diff
@@ -236,6 +236,24 @@ typedef struct ucx_perf_rte {
 } ucx_perf_rte_t;
 
 
+typedef struct {
+    /* Whether DPU offloading daemon is configured */
+    int                     is_daemon_mode;
+    /* Whether to keep daemon alive at the end of the test run */
+    int                     is_keep_alive;
```
brminich, line 243: maybe rename to `is_keep_running` to avoid confusion with keepalive protocols?

src/tools/perf/perftest_daemon.c
```diff
@@ -409,18 +448,10 @@ ucp_perf_daemon_peer_tx_handler(void *arg, const void *header,
 
 static void ucp_perf_daemon_cleanup(ucp_perf_daemon_context_t *ctx)
 {
-    if (NULL != ctx->send_memh) {
-        /* coverity[check_return] */
-        ucp_mem_unmap(ctx->context, ctx->send_memh);
-    }
+    ucs_trace_data("cleaning up daemon");
```
brminich, line 451: maybe just `trace` because it is not on a data path?

src/tools/perf/api/libperf.h
iyastreb: ok, renamed, + all the related variables and comments

src/tools/perf/perftest_daemon.c
iyastreb: ok, changed here and in ucp_perf_daemon_cleanup_connections

brminich (dismissed):

src/tools/perf/perftest_run.c
iyastreb: Yes, you are right, now I see
Ideally we want to reuse the same code for parsing, let me try to implement that

iyastreb: done

tvegas1 (approved):

src/tools/perf/perftest_run.c
```diff
@@ -210,6 +210,26 @@ static void print_test_name(struct perftest_context *ctx)
     }
 }
 
+static int read_args(char *buf, char **argv, size_t argv_size)
```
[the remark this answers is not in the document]
iyastreb: Yes, I thought about this, but still decided to keep `size_t`.
My idea was to be consistent with `int main(int argc, char *const argv[])` signature, that's why we return `int` from the function - to be used as `argc` later on.
But `argv_size` is not gonna be used anywhere apart from `read_args`, there is no corresponding argument in main. It represents the size of the allocated array, which is normally is `size_t`. For instance, I could call `read_args` function like that:
`read_args(buf, argv, ucs_static_array_size(argv))`
So I thought `size_t` would be more appropriate here

contrib/ucx_perftest_config/test_types_ucp_daemon
```diff
@@ -0,0 +1,4 @@
+#
+# UCP daemon basic
+#
+ucp_daemon_bw                        -t ucp_am_bw -w 0 -n 10000 -s 8192
```
yosefe, line 4: why can't we reuse test_types_ucp? or just run ucx_perftest without batch file (since in this batch file there is only one test)

contrib/test_jenkins.sh
```diff
@@ -574,6 +574,70 @@ run_ucx_perftest() {
 	fi
 }
 
+start_daemon() {
```
yosefe, line 577: CI should be similar to how ucx_perftest MAD is run - using different steps in Azure pipelines jobs, including a step that cleans up the deamon in case of failure
we can move CI scripts from this PR to next PR

src/tools/perf/perftest_params.c
```diff
@@ -670,6 +684,23 @@ ucs_status_t parse_opts(struct perftest_context *ctx, int mpi_initialized,
...
+            }
+            break;
+        case 'k': /* handles daemon-keep-running long option as well */
+            ctx->dmn_info.is_keep_running = 1;
```
yosefe, line 702: do we really need separate keep running option? I'd expect the daemon will keep running until the client finishes the batch file and send a FIN command that tells the daemon to exit

src/tools/perf/api/libperf.h
```diff
@@ -297,23 +315,12 @@ typedef struct ucx_perf_params {
         ucp_perf_datatype_t     recv_datatype;
         size_t                  am_hdr_size; /* UCP Active Message header size
                                                 (not included in message size) */
-        int                     is_daemon_mode;  /* Whether DPU offloading daemon
-                                                    is configured */
-        struct sockaddr_storage dmn_local_addr;  /* IP and port of local daemon,
-                                                    used to offload communication */
-        struct sockaddr_storage dmn_remote_addr; /* IP and port of remote daemon,
-                                                    used to offload communication */
+        ucp_perf_daemon_info_t  dmn_info;    /* DPU offloading daemon info */
```
yosefe, line 318: why need  a separate struct?

src/tools/perf/perftest_context.h
```diff
@@ -50,6 +50,7 @@ struct perftest_context {
     unsigned                     num_cpus;
     unsigned                     cpus[MAX_CPUS];
     unsigned                     flags;
+    ucp_perf_daemon_info_t       dmn_info;
```
yosefe, line 53: why need to add this - isn't this already part of "params" ?

src/tools/perf/perftest_daemon.c
```diff
@@ -347,15 +348,48 @@ ucp_perf_daemon_send_handler(void *arg, const void *header,
     return UCS_OK;
 }
 
+static void ucp_perf_daemon_cleanup_connections(ucp_perf_daemon_context_t *ctx)
```
yosefe, line 351: instead of adding new function "ucp_perf_daemon_cleanup_connections", can we mov this code directly to fin_handler, so it will be symmetric to init_handler?

src/tools/perf/perftest_daemon.c
```diff
@@ -347,15 +348,48 @@ ucp_perf_daemon_send_handler(void *arg, const void *header,
...
+    ucp_perf_daemon_ep_close(ctx, ctx->host_ep);
+    ctx->host_ep = NULL;
+}
+
 static ucs_status_t
 ucp_perf_daemon_fin_handler(void *arg, const void *header, size_t header_length,
                             void *data, size_t length,
                             const ucp_am_recv_param_t *param)
 {
+    uint8_t keep_running;
```
yosefe, line 379: maybe define struct type for the FIN message?

src/tools/perf/perftest_run.c
yosefe, line 213: maybe rename to read_batch_args?

src/tools/perf/perftest_run.c
```diff
@@ -262,10 +273,30 @@ static ucs_status_t read_batch_file(FILE *batch_file, const char *file_name,
     return UCS_OK;
 }
 
+static size_t get_test_count(FILE *batch_file)
```
yosefe, line 276: instead of doing this, maybe we could send a FIN message to the daemon to exit, after reaching the end of the batch file? that means we would send most FIN messages after getting the next test from the batch file (instead of immediately after running the test), and the last FIN will be outside of the loop

src/tools/perf/perftest_params.c
iyastreb: Of course we can live without it. I added this option with the following ideas in mind:
1. (minor) It allows us to run multiple batches in a row, like a regular sender-receiver perftest
2. It allows us to run persistent daemon service, and execute multiple tests without restarting 2 daemon processes for each run. Personally I found this useful when manually testing performance on different message sizes. With this option I just have to restart sender and receiver, but not their daemons
3. It brings no complexity, but can be useful

But if we change design like you proposed in other comment, this option might be obsolete

contrib/ucx_perftest_config/test_types_ucp_daemon
iyastreb: For now we cannot reuse `test_types_ucp` because of multiple limitations:
- tag matching is not supported
- only contig buffers are supported
- only single thread is supported
- latency tests need to be adjusted to daemon workflow
If/when these limitations are addressed, then we can reuse this test suite.

Regarding batch vs single test, I'm absolutely sure that we need a batch mode for the future improvements (out-of-the-box XGVMI task). For instance, we need to add a test for short messages once we force XGVMI to always use RNDV. I also plan to add am_lat tests to this batch in the future. So IMO having a separate test batch is the right choice

src/tools/perf/api/libperf.h
iyastreb: To copy global-params daemon config into per-test params, please see my other comment on this

src/tools/perf/perftest_context.h
iyastreb: There are 2 types of perftest params:
- global ones, configured once for all tests: remote server, port, AF, etc
- per-test params, configured per each individual test: test name, msg size, etc

Initially I put daemon params to the per-test params, which seems wrong. Logically we need a single daemon configured per batch on test (just like receiver server), and this config must not be allowed on per-test basis. That's why daemon config is now stored in the global params.
However, we still need to copy this config to per-test params, because at runtime perftest needs this information in order to connect to the daemon. I didn't find a clean way to access global properties from the perftest `ucp_perf_setup`. So I thought that an easiest option is just to copy this struct from global-params to per-test-params.

src/tools/perf/perftest_daemon.c
iyastreb: It was implemented almost like you said, before this change. All the cleanup code was in `ucp_perf_daemon_cleanup`. Now the idea was to support `keep_running` mode, so that host process may connect to the daemon multiple times. This means that disconnection step is now separate from the overall cleanup, and now is encapsulated in `ucp_perf_daemon_cleanup_connections` function.
This function is called from 2 places: from FIN message handler, and from the main cleanup. So I cannot keep it inside the FIN handler

src/tools/perf/perftest_context.h
iyastreb: I reflected a bit more on that, and I think you are right, I will remove this duplication.

src/tools/perf/api/libperf.h
iyastreb: will be reverted

src/tools/perf/perftest_params.c
iyastreb: ok, I'll remove it

src/tools/perf/perftest_run.c
iyastreb: sure

src/tools/perf/perftest_daemon.c
yosefe: ok

contrib/ucx_perftest_config/test_types_ucp_daemon
yosefe: ok

src/tools/perf/perftest_daemon.c
iyastreb: ok

src/tools/perf/perftest_run.c
iyastreb: I see what you mean, but I think this would not fly because it is perf test who owns EP-to-daemon. Individual perf test is self-contained, in the sense that on startup it initializes the EP to daemon, and at the end of the test it destroys that EP. So there is no mean to send FIN message over this EP afterwards, after the test cleanup is executed.

But in principle I agree that daemon is a kind of shared resource, and ideally it should be "owned" not by an individual test, but rather by a test runner. However this means that we need to have EP-to-daemon outside of the perf-test, somewhere in perftest_run.c:
```
perftest_run:run_test
    run_test_recurs
        ucx_perf_run
            ucp_perf_setup: create EP2D, -> [INIT]
            ucp_perf_cleanup: -> [FIN keep-running], destroy EP2D // never kill daemon from test, as it's shared resource
        ucx_perf_run ...
        ucx_perf_run ...
    kill_daemon: create EP2D, -> [FIN kill], destroy EP2D
```
That's the cleanest option IMO, what do you think?

yosefe: I think it's a good idea!

iyastreb: done

contrib/test_jenkins.sh
```diff
@@ -561,6 +561,67 @@ run_ucx_perftest() {
...
+#
+run_ucx_perftest_daemon() {
+	ucx_inst_ptest=$ucx_inst/share/ucx/perftest
+
+	ucx_perftest="$ucx_inst/bin/ucx_perftest"
+	ucx_perftest_daemon="$ucx_inst/bin/ucx_perftest_daemon"
+	ucp_test_args="-b $ucx_inst_ptest/test_types_ucp_daemon"
+
+	devices="$(get_active_ib_devices)"
+	devices="$(get_ib_bf_devices $devices)"
```
yosefe, line 593: to avoid overriding a var
devices="$(get_ib_bf_devices $(get_active_ib_devices))"

contrib/test_jenkins.sh
```diff
@@ -561,6 +561,67 @@ run_ucx_perftest() {
 	fi
 }
 
+start_daemon() {
```
yosefe, line 564: rename to start_perftest_daemon

contrib/test_jenkins.sh
```diff
@@ -561,6 +561,67 @@ run_ucx_perftest() {
...
+
+	# Return the daemon pid and port
+	eval "$2=$!"
+	eval "$3=$dmn_port"
+}
+
+#
+# Run UCX performance daemon test
+#
+run_ucx_perftest_daemon() {
```
yosefe, line 585: run_ucx_perftest_with_daemon
or
run_ucx_perftest_over_daemon

contrib/test_jenkins.sh
```diff
@@ -561,6 +561,67 @@ run_ucx_perftest() {
...
+	ucx_perftest_daemon="$ucx_inst/bin/ucx_perftest_daemon"
+	ucp_test_args="-b $ucx_inst_ptest/test_types_ucp_daemon"
+
+	devices="$(get_active_ib_devices)"
+	devices="$(get_ib_bf_devices $devices)"
+
+	my_devices=$(get_my_tasks $devices)
+	for ucx_dev in $my_devices
+	do
+		echo "==== Running ucx_perftest_daemon kit on $ucx_dev ===="
```
yosefe, line 598: "==== Running ucx_perftest over a daemon on $ucx_dev ===="

contrib/test_jenkins.sh
```diff
@@ -561,6 +561,67 @@ run_ucx_perftest() {
...
+		export UCX_NET_DEVICES=$ucx_dev,lo
```
yosefe, line 599: why need lo ?

contrib/test_jenkins.sh
```diff
@@ -561,6 +561,67 @@ run_ucx_perftest() {
...
+
+		# We explicitly disable cuda transport, because it's not p2p and therefore
+		# imposes INVALIDATE_RMA flag for all lanes (@see ucp_ep_config_init).
+		# However invalidating of imported rkeys is not supported by the daemon.
+		# TODO: Should we support invalidation of imported keys?
+		# Normally cuda_ipc cannot be used to communicate between host and DPU,
+		# unless we run both processes on host for testing purposes.
+		export UCX_TLS=^cuda
```
yosefe, line 607: why this is not part of dmn_env?

src/tools/perf/perftest_daemon.c
```diff
@@ -51,38 +51,25 @@ const char *ucp_perf_daemon_am_id_name(ucp_perf_daemon_am_id_t am_id)
...
-        } while (status == UCS_INPROGRESS);
-        ucp_request_free(close_req);
-    } else {
-        status = UCS_PTR_STATUS(close_req);
-    }
-
-    if (status != UCS_OK) {
-        ucs_error("daemon failed to close ep %p", ep);
-    }
+    ucp_ep_close_nbx(ep, &param);
```
yosefe, line 65: why not waiting for the close request?

src/tools/perf/perftest_daemon.c
```diff
@@ -163,8 +154,15 @@ ucp_perf_daemon_send_cb(void *request, ucs_status_t status, void *user_data)
 {
     ucp_perf_daemon_context_t *ctx = user_data;
 
+    if (status != UCS_OK) {
+        ucs_error("PEER_TX operation failed: %s", ucs_status_string(status));
```
yosefe, line 158: "SEND operation failed: %s"

src/tools/perf/perftest_params.c
```diff
@@ -312,7 +301,28 @@ static ucs_status_t verify_daemon_params(ucx_perf_params_t *params)
...
+
+static ucs_status_t init_daemon_params(ucx_perf_params_t *params)
+{
+    struct sockaddr_storage *local_addr  = &params->ucp.dmn_local_addr;
+    struct sockaddr_storage *remote_addr = &params->ucp.dmn_remote_addr;
+
+    /* Return if both daemon addresses are not configured */
+    if ((0 == local_addr->ss_family) && (0 == remote_addr->ss_family)) {
+        return UCS_OK;
+    }
```
yosefe, line 315: shouldn't this be in check_daemon_params?

src/tools/perf/perftest_run.c
```diff
@@ -319,6 +319,24 @@ static ucs_status_t run_test_recurs(struct perftest_context *ctx,
...
+
+    perf->params                     = ctx->params.super;
+    perf->params.ucp.is_keep_running = 0;
+    perf->params.command             = UCX_PERF_CMD_AM;
+    perf->params.max_outstanding     = 1;
+
+    ucx_perf_funcs[UCX_PERF_API_UCP].setup(perf);
+    ucx_perf_funcs[UCX_PERF_API_UCP].cleanup(perf);
+    free(perf);
+}
```
yosefe, line 338: seems like over complicated to me. maybe just remove keep_running from all places, and kill the daemon from the test script? FIN message would never kill the daemon on itself.

src/tools/perf/perftest_daemon.c
iyastreb: ok

iyastreb (commented):
Part 1 of comments, will address the rest in the next batch

src/tools/perf/perftest_daemon.c
iyastreb: I looked at the impl of `ucp_ep_close_nbx`. When we close with the `UCP_EP_CLOSE_FLAG_FORCE` flag, then returned request is just dummy:
```
    if (ucp_request_param_flags(param) & UCP_EP_CLOSE_FLAG_FORCE) {
        ucp_ep_discard_lanes(ep, UCS_ERR_CANCELED);
        ucp_ep_disconnected(ep, 1);
    } else {
        request = ucp_ep_flush_internal(ep, 0, param, NULL,
                                        ucp_ep_close_flushed_callback, "close");
```

contrib/test_jenkins.sh
iyastreb: Right, in principle that should work without loopback interface, we can just specify `0.0.0.0` address for daemons.
However, this didn't work out of the box. When host tries to connect to daemon, I see an error on daemon side:
```
    tcp_listener.c:136  UCX  DEBUG created a TCP listener 0x1cfaec0 on cm 0x1cd6400 with fd: 59 listening on 0.0.0.0:1338
   tcp_sockcm_ep.c:1128 UCX  DEBUG server created an endpoint on tcp_sockcm 0x1cd6400 id: 0 state: 1
           async.c:232  UCX  DEBUG added async handler 0x7ff8380011c0 [id=60 ref 1] uct_tcp_sa_data_handler() to hash
           async.c:494  UCX  DEBUG listening to async event fd 60 events 0x5 mode thread_spinlock
   tcp_sockcm_ep.c:360  UCX  DEBUG ep 0x7ff8380008c0 (fd=60 state=1): remote peer (127.0.0.1:44684) disconnected/rejected (Endpoint is not connected)
```
So on the daemon server side the remote peer (i.e. host client process) is still processed as `127.0.0.1:44684`.
Honestly I didn't spent much time debugging it, just enabled loopback interface to make it working.

Why 0.0.0.0 works for host-to-host communication?
Because there we don't use UCP EP, but plain TCP socket communication

contrib/test_jenkins.sh
iyastreb: Right, it's sufficient to have this env var only in daemon environment
Fixed

contrib/test_jenkins.sh
yosefe: can we listen on the network address of $ucx_dev instead of 0.0.0.0 ?

src/tools/perf/perftest_params.c
yosefe: can we document this distinction in the names of the functions and/or in some comment before thr functions themselves?

src/tools/perf/perftest_run.c
yosefe: let's discuss it

yosefe: per offline discussion, daemon should keep running and close the session with a specific client when getting disconnected (error callback is called on the endpoint)

src/tools/perf/perftest_params.c
iyastreb: sure

src/tools/perf/perftest_run.c
iyastreb: All done, it was quite significant redesign after all.
FIN message is redundant, we can close connection on EP disconnect, so removed FIN completely.
Now daemona are permanent processes (if started manually), which are killed explicitly from test_jenkins.
Also added mechanism to detect stale entries in alias hash by unique ID (which is generated per IB MD, then passed through `uct_ib_md_packed_mkey_t`, then stored in mkey alias).
Process ID does not work for this purpose, because a single process may close MD (and destroy mkey UMR pool), and then reopen it - like happens when we execute test batch.

contrib/test_jenkins.sh
```diff
@@ -582,38 +589,38 @@ start_daemon() {
...
 
-		# We explicitly disable cuda transport, because it's not p2p and therefore
-		# imposes INVALIDATE_RMA flag for all lanes (@see ucp_ep_config_init).
-		# However invalidating of imported rkeys is not supported by the daemon.
-		# TODO: Should we support invalidation of imported keys?
-		# Normally cuda_ipc cannot be used to communicate between host and DPU,
-		# unless we run both processes on host for testing purposes.
-		export UCX_TLS=^cuda
+		export UCX_NET_DEVICES=$ucx_dev
+		export UCX_MAX_RNDV_LANES=1
```
yosefe, line 612: why needed?

contrib/test_jenkins.sh
```diff
@@ -561,15 +561,22 @@ run_ucx_perftest() {
...
+	# We explicitly disable cuda transport, because it's not p2p and therefore
+	# imposes INVALIDATE_RMA flag for all lanes (@see ucp_ep_config_init).
+	# However invalidating of imported rkeys is not supported by the daemon.
+	# TODO: Should we support invalidation of imported keys?
+	# Normally cuda_ipc cannot be used to communicate between host and DPU,
+	# unless we run both processes on host for testing purposes.
+
 	# Mandatory options to run the daemon
-	dmn_env="UCX_TCP_CM_REUSEADDR=y UCX_MAX_RNDV_LANES=1 UCX_RNDV_THRESH=0 UCX_RNDV_SCHEME=get_zcopy"
+	dmn_env="UCX_TLS=^cuda UCX_TCP_CM_REUSEADDR=y UCX_RNDV_THRESH=0 UCX_RNDV_SCHEME=put_zcopy"
```
yosefe, line 579: why need to set UCX_RNDV_SCHEME? pls add to the comment

src/tools/perf/perftest_daemon.c
```diff
@@ -62,15 +65,56 @@ static void ucp_perf_daemon_ep_close(ucp_ep_h ep)
...
 
-    terminated = 1;
+    if (ep == ctx->host_ep) {
+        ucs_debug("ep %p: host error handler called with status %s", ep,
+                  ucs_status_string(status));
+        ucp_perf_daemon_cleanup_connections(arg);
+    } else {
+        ucs_error("ep %p: peer error handler called with status %s", ep,
+                  ucs_status_string(status));
+
```
yosefe, line 115: minor: remove space line

src/uct/ib/base/ib_md.h
```diff
@@ -15,7 +15,7 @@
 #include <ucs/stats/stats.h>
 
 #define UCT_IB_MD_MAX_MR_SIZE        0x80000000UL
-#define UCT_IB_MD_PACKED_RKEY_SIZE   sizeof(uint64_t)
+#define UCT_IB_MD_PACKED_RKEY_SIZE   2 * sizeof(uint64_t)
```
yosefe, line 18: why needed?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1408,9 +1408,20 @@ uct_ib_mlx5_devx_umr_mkey_hash_find(uct_ib_mlx5_md_t *md,
     }
 
     *umr_alias = kh_val(md->umr.mkey_hash, khiter);
-    ucs_trace("%s: found " UCT_IB_MLX5_UMR_ALIAS_FMT " for index 0x%x in hash",
-              uct_ib_mlx5_dev_name(md), UCT_IB_MLX5_UMR_ALIAS_ARG(umr_alias),
-              uct_ib_mlx5_mkey_index(packed_mkey->lkey));
+    /* Check if alias is from the same MD as packed_mkey */
```
yosefe, line 1411: space line before

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2353,6 +2363,7 @@ ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
     md->flags           |= UCT_IB_MLX5_MD_FLAGS_DEVX_OBJS(md_config->devx_objs);
     md->super.name       = UCT_IB_MD_NAME(mlx5);
     md->super.vhca_id    = vhca_id;
+    md->super.uuid       = ucs_generate_uuid(0);
```
yosefe, line 2366: ucs_generate_uuid((uintptr_t)(md))

src/uct/ib/base/ib_md.h
```diff
@@ -161,6 +162,7 @@ typedef struct uct_ib_md {
 
 
 typedef struct uct_ib_md_packed_mkey {
+    uint64_t uuid;
```
yosefe, line 165: md_uuid

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -266,13 +266,14 @@ typedef struct {
 typedef struct {
     struct mlx5dv_devx_obj *cross_mr;
     uint32_t               lkey;
+    uint64_t               uuid;
```
yosefe, line 269: md_uuid

contrib/test_jenkins.sh
iyastreb: It works without this option set, removing

src/tools/perf/perftest_daemon.c
iyastreb: done

src/uct/ib/base/ib_md.h
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: done

src/uct/ib/mlx5/ib_mlx5.h
iyastreb: fixed

src/uct/ib/base/ib_md.h
iyastreb: This is needed to reflect the change in `uct_ib_md_packed_mkey_t`: we added a new `md_uuid` field to it, and the structure size increased by 8B.
```C
typedef struct uct_ib_md_packed_mkey {
    uint64_t md_uuid;  << newly added field
    uint32_t lkey;
    uint16_t vhca_id;
} UCS_S_PACKED uct_ib_md_packed_mkey_t;
```
And uuid field was added as a way to distinguish stale entries in alias cache on DPU side. Here is our discussion: https://github.com/openucx/ucx/pull/9752#discussion_r1711229304

yosefe: 1. does it break wire compatibility for non-xGVMI use cases?
2. can we use sizeof() instead of such define?
3. wasn't the previous size 6 bytes and not 8?

iyastreb: I have pushed the backward compatible approach described in p1

src/uct/api/v2/uct_v2.h
```diff
@@ -544,6 +546,11 @@ typedef struct uct_md_mem_attach_params {
      * @ref uct_md_mem_attach_flags_t.
      */
     uint64_t                     flags;
+
+    /**
+     * Size of the memory key.
+     */
+    uint8_t                      mkey_size;
```
yosefe, line 553: size_t

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2937,6 +2937,25 @@ uct_ib_mlx5_devx_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
...
+    packed_mkey.vhca_id = tmp->vhca_id;
+    packed_mkey.md_uuid = 0;
+
+    /* Backward compatibility: extract MD UUID from mkey buffer only if it's
+     * present there */
+    if (mkey_size > ucs_offsetof(uct_ib_md_packed_mkey_t, md_uuid)) {
+        packed_mkey.md_uuid = tmp->md_uuid;
+    }
+
+    return packed_mkey;
```
yosefe, line 2956: IMO we should use "serialize" functions, since the uuid field may not be accessible and it's too dangerous to assign it to "tmp" pointer where last field may not be accessible

yosefe (approved):

PR #9757, opened 2024-03-17
UCP/MM: Fix assertion

Artemy-Mellanox:
## Why ?
Those assertions are true only if ucp_memh_rcache_get succeeded.

### DIFF
--- a/src/ucm/bistro/bistro.c
+++ b/src/ucm/bistro/bistro.c
@@ -17,7 +17,7 @@
 #include <ucm/bistro/bistro.h>
 #include <ucm/bistro/bistro_int.h>
 #include <ucs/type/serialize.h>
-#include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
 #include <ucs/time/time.h>
 #include <ucs/arch/atomic.h>
 
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -792,10 +792,6 @@ ucs_status_t ucp_memh_get_slow(ucp_context_h context, void *address,
         status = ucp_memh_rcache_get(context->rcache, reg_address, reg_length,
                                      reg_align, mem_type, reg_md_map, uct_flags,
                                      alloc_name, &memh);
-
-        ucs_assert(memh->mem_type == mem_type);
-        ucs_assert(ucs_padding((intptr_t)ucp_memh_address(memh), reg_align) == 0);
-        ucs_assert(ucs_padding(ucp_memh_length(memh), reg_align) == 0);
     }
 
     if (status != UCS_OK) {
@@ -808,6 +804,14 @@ ucs_status_t ucp_memh_get_slow(ucp_context_h context, void *address,
             alloc_name, address, ucp_memh_address(memh), length,
             ucp_memh_length(memh), ucs_memory_type_names[mem_type], reg_md_map,
             uct_flags);
+
+    ucs_assertv(mem_type == memh->mem_type, "mem_type=%s got memh->mem_type=%s",
+                ucs_memory_type_names[mem_type],
+                ucs_memory_type_names[memh->mem_type]);
+    if (context->rcache != NULL) {
+        ucs_ptr_check_align(ucp_memh_address(memh), ucp_memh_length(memh),
+                            reg_align);
+    }
 
     status = ucp_memh_register(context, memh, reg_md_map, uct_flags,
                                alloc_name);
--- a/src/ucs/datastruct/mpool.c
+++ b/src/ucs/datastruct/mpool.c
@@ -13,7 +13,7 @@
 #include "queue.h"
 
 #include <ucs/debug/log.h>
-#include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
 #include <ucs/sys/checker.h>
 #include <ucs/sys/sys.h>
 #include <ucs/arch/cpu.h>
--- a/src/ucs/datastruct/mpool_set.inl
+++ b/src/ucs/datastruct/mpool_set.inl
@@ -12,7 +12,7 @@
 #include <ucs/datastruct/mpool.inl>
 #include <ucs/debug/assert.h>
 #include <ucs/arch/bitops.h>
-#include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
 
 
 /**
--- a/src/ucs/datastruct/pgtable.c
+++ b/src/ucs/datastruct/pgtable.c
@@ -14,7 +14,7 @@
 #include <ucs/debug/assert.h>
 #include <ucs/debug/log.h>
 #include <ucs/debug/memtrack_int.h>
-#include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
 #include <string.h>
 
 
--- a/src/ucs/memory/memtype_cache.c
+++ b/src/ucs/memory/memtype_cache.c
@@ -18,7 +18,7 @@
 #include <ucs/debug/memtrack_int.h>
 #include <ucs/stats/stats.h>
 #include <ucs/sys/sys.h>
-#include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
 #include <ucm/api/ucm.h>
 
 
--- a/src/ucs/sys/math.h
+++ b/src/ucs/sys/math.h
@@ -47,48 +47,8 @@ BEGIN_C_DECLS
 #define ucs_is_pow2(_n) \
     (((_n) > 0) && ucs_is_pow2_or_zero(_n))
 
-#define ucs_padding(_n, _alignment) \
-    ( ((_alignment) - (_n) % (_alignment)) % (_alignment) )
-
-#define ucs_align_down(_n, _alignment) \
-    ( (_n) - ((_n) % (_alignment)) )
-
-#define ucs_align_up(_n, _alignment) \
-    ( (_n) + ucs_padding(_n, _alignment) )
-
-#define ucs_align_down_pow2(_n, _alignment) \
-    ( (_n) & ~((_alignment) - 1) )
-
-#define ucs_align_up_pow2(_n, _alignment) \
-    ucs_align_down_pow2((_n) + (_alignment) - 1, _alignment)
-
-#define ucs_align_down_pow2_ptr(_ptr, _alignment) \
-    ((ucs_typeof(_ptr))ucs_align_down_pow2((uintptr_t)(_ptr), (_alignment)))
-
-#define ucs_align_up_pow2_ptr(_ptr, _alignment) \
-    ((ucs_typeof(_ptr))ucs_align_up_pow2((uintptr_t)(_ptr), (_alignment)))
-
-#define ucs_roundup_pow2(_n) \
-    ({ \
-        ucs_typeof(_n) pow2; \
-        ucs_assert((_n) >= 1); \
-        for (pow2 = 1; pow2 < (_n); pow2 <<= 1); \
-        pow2; \
-    })
-
-#define ucs_rounddown_pow2(_n) (ucs_roundup_pow2(_n + 1) / 2)
-
 #define ucs_signum(_n) \
     (((_n) > (ucs_typeof(_n))0) - ((_n) < (ucs_typeof(_n))0))
-
-#define ucs_roundup_pow2_or0(_n) \
-    ( ((_n) == 0) ? 0 : ucs_roundup_pow2(_n) )
-
-/* Return values: 0 - aligned, non-0 - unaligned */
-#define ucs_check_if_align_pow2(_n, _p) ((_n) & ((_p) - 1))
-
-/* Return values: off-set from the alignment */
-#define ucs_padding_pow2(_n, _p) ucs_check_if_align_pow2(_n, _p)
 
 #define UCS_MASK_SAFE(_i) \
     (((_i) >= 64) ? ((uint64_t)(-1)) : UCS_MASK(_i))
@@ -107,18 +67,6 @@ static UCS_F_ALWAYS_INLINE size_t ucs_double_to_sizet(double value, size_t max)
     return (round_value < (double)max) ? ((size_t)round_value) : max;
 }
 
-static UCS_F_ALWAYS_INLINE void
-ucs_align_ptr_range(void **address_p, size_t *length_p, size_t alignment)
-{
-    void *start, *end;
-
-    start = ucs_align_down_pow2_ptr(*address_p, alignment);
-    end   = ucs_align_up_pow2_ptr(UCS_PTR_BYTE_OFFSET(*address_p, *length_p),
-                                  alignment);
-
-    *address_p = start;
-    *length_p  = UCS_PTR_BYTE_DIFF(start, end);
-}
 
 /**
  * Convert flags without a branch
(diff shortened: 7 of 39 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -792,16 +792,17 @@ ucs_status_t ucp_memh_get_slow(ucp_context_h context, void *address,
...
-        ucs_assert(ucs_padding((intptr_t)ucp_memh_address(memh), reg_align) == 0);
-        ucs_assert(ucs_padding(ucp_memh_length(memh), reg_align) == 0);
     }
 
     if (status != UCS_OK) {
         goto out;
     }
 
+    ucs_assert(memh->mem_type == mem_type);
+    ucs_assert((context->rcache) == NULL ||
```
yosefe, line 802: maybe put these assertions under `if (context->rcache == NULL) {`?
also may use this opportunity to convert them into ucs_assertv

Artemy-Mellanox: up to now we've tucked the assertion logic right into the assertions themselves
if you're thinking of switching this up for new code, let's give everyone a heads-up
otherwise you'll have to bring it up each time, as we've see here and [there](https://github.com/openucx/ucx/pull/9449#discussion_r1509907227)

yosefe: it depends on the assertion condition itself. in this particular case, putting the rcache!=NULL check inside the assertion logic makes it more complicated, unites 2 assertions to one (so when it fails, it's harder to know which one failed), and the check or "rcache == NULL" is not the essence of the assertion itself, but just the condition under which we want to perform the check (otherwise, memh is not even valid)

tvegas1: also use `ucs_assertv()` instead and dump all fields (length, address, align and padding)?

tvegas1 (approved):
code style mandates using equal instead of space to get like: `<field>=%p`

src/ucp/core/ucp_mm.c
```diff
@@ -809,6 +805,16 @@ ucs_status_t ucp_memh_get_slow(ucp_context_h context, void *address,
             ucp_memh_length(memh), ucs_memory_type_names[mem_type], reg_md_map,
             uct_flags);
 
+    ucs_assertv(mem_type == memh->mem_type, "mem_type expected %s got %s",
+                ucs_memory_type_names[mem_type],
+                ucs_memory_type_names[memh->mem_type]);
+    if (context->rcache != NULL) {
+        ucs_assertv(ucs_padding((intptr_t)ucp_memh_address(memh), reg_align) == 0,
+                    "address %p align %zu", ucp_memh_address(memh), reg_align);
```
yosefe, line 813: address=%p align=%zu

src/ucp/core/ucp_mm.c
```diff
@@ -809,6 +805,16 @@ ucs_status_t ucp_memh_get_slow(ucp_context_h context, void *address,
...
+        ucs_assertv(ucs_padding(ucp_memh_length(memh), reg_align) == 0,
+                    "length %zu align %zu", ucp_memh_length(memh), reg_align);
```
yosefe, line 815: length=%zu align=%zu

yosefe (approved):

tvegas1 (approved):

brminich (approved):

Artemy-Mellanox:
we have similar code in `uct_gdr_copy_mem_reg()`
maybe move it to *ucs/sys/math.h* along with `ucs_align_ptr_range()` and reuse in both cases?

yosefe:
> we have similar code in `uct_gdr_copy_mem_reg()` maybe move it to _ucs/sys/math.h_ along with `ucs_align_ptr_range()` and reuse in both cases?

Sounds good, it can be an inline function like ucs_ptr_check_align

src/ucp/core/ucp_mm.c
```diff
@@ -809,6 +805,14 @@ ucs_status_t ucp_memh_get_slow(ucp_context_h context, void *address,
             ucp_memh_length(memh), ucs_memory_type_names[mem_type], reg_md_map,
             uct_flags);
 
+    ucs_assertv(mem_type == memh->mem_type, "mem_type=%s got memh->mem_type=%s",
+                ucs_memory_type_names[mem_type],
+                ucs_memory_type_names[memh->mem_type]);
+    if (context->rcache != NULL) {
```
yosefe, line 811: BTW, why the assertion is not valid when not using rcache?

src/ucs/sys/math.h
```diff
@@ -120,6 +121,15 @@ ucs_align_ptr_range(void **address_p, size_t *length_p, size_t alignment)
     *length_p  = UCS_PTR_BYTE_DIFF(start, end);
 }
 
+static UCS_F_ALWAYS_INLINE void
+ucs_ptr_check_align(void *address, size_t length, size_t alignment)
+{
+    ucs_assertv(ucs_padding((intptr_t)address, alignment) == 0,
```
yosefe, line 127: maybe uintptr_t ?

src/ucp/core/ucp_mm.c
Artemy-Mellanox: when rcache is disabled it doesn't align regions

yosefe: doesn't seem right
why we must align only with rcache?

yosefe:
failure seems relevant
https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=78196&view=logs&j=087a5e36-b289-5cda-d2a1-de6f38127104&t=3360d7e5-c1ed-5196-84c0-e08ae279e4f5
```
==== Running profiling example  ====
In file included from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/datastruct/array.h:11:0,
                 from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/datastruct/string_buffer.h:12,
                 from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/config/parser.h:13,
                 from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/arch/global_opts.h:11,
                 from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/config/global_opts.h:15,
                 from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/profile/profile_defs.h:10,
                 from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/profile/profile_on.h:10,
                 from ../test/apps/profiling/ucx_profiling.c:7:
/scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/sys/math.h:19:30: fatal error: ucs/debug/assert.h: No such file or directory
 #include <ucs/debug/assert.h>
```

src/ucp/core/ucp_mm.c
Artemy-Mellanox: my understanding was that by turning off rcache, we're trying to avoid any unexpected quirks it might bring along
take for instance when we're dealing with rcache and registering on multiple MDs we need to make sure regions are aligned just right to match what the MDs need
this ensures that later on rcache lookups can fetch the region properly
however this alignment can sometimes lead to new headaches
so I thought it'd be neat if disabling rcache also turns off this alignment feature
that way if we disable rcache and the issue disappears we'll have a quick way to tell if it was caused by rcache or not

yosefe: but according to UCT API, the registration must use the alignment specified in uct_md_attr_v2_t.reg_alignment - regardless of caching on higher level
maybe this API field needs to be defined in a better way?

Artemy-Mellanox: sure, but [if rcache is disabled we align registration to uct_md_attr_v2_t.reg_alignment for specific MDs later on](https://github.com/openucx/ucx/pull/9757/files#diff-9136600ae0324c21c07ccead535b761ee5a5113c2817d02af656b33ce992ba8dR519) so their won't interfere

yosefe:
> > failure seems relevant https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=78196&view=logs&j=087a5e36-b289-5cda-d2a1-de6f38127104&t=3360d7e5-c1ed-5196-84c0-e08ae279e4f5
> > ```
> > ==== Running profiling example  ====
> > In file included from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/datastruct/array.h:11:0,
> >                  from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/datastruct/string_buffer.h:12,
> >                  from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/config/parser.h:13,
> >                  from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/arch/global_opts.h:11,
> >                  from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/config/global_opts.h:15,
> >                  from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/profile/profile_defs.h:10,
> >                  from /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/profile/profile_on.h:10,
> >                  from ../test/apps/profiling/ucx_profiling.c:7:
> > /scrap/azure/agent-02/AZP_WORKSPACE/2/s/install/include/ucs/sys/math.h:19:30: fatal error: ucs/debug/assert.h: No such file or directory
> >  #include <ucs/debug/assert.h>
> > ```
> 
> need to place the new macro in the noinst header? I can't find the appropriate one. add a new one?

Maybe in assert.h?

yosefe:
> > > need to place the new macro in the noinst header? I can't find the appropriate one. add a new one?
> > 
> > 
> > Maybe in assert.h?
> 
> than assert.h will depend on ucs_padding() from math.h btw. math.h weakly depends in ucs_round_pow2() on ucs_assert() maybe split math.h into a minimal part which needs to be exported and the rest make noinst?

maybe math.h - public stuff w/o pointer arithmetic

pointer.h / ptr.h / ptr_arith.h / ptr_math.h - pointer arithmetic, internal, depends on assert.h, includes pointer align / check and ucs_round/align* macros?

src/ucm/bistro/bistro.c
```diff
@@ -18,6 +18,7 @@
 #include <ucm/bistro/bistro_int.h>
 #include <ucs/type/serialize.h>
 #include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 21: remove math.h?

src/ucs/datastruct/mpool.c
```diff
@@ -14,6 +14,7 @@
 
 #include <ucs/debug/log.h>
 #include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 17: remove math.h?

src/ucs/datastruct/mpool_set.inl
```diff
@@ -13,6 +13,7 @@
 #include <ucs/debug/assert.h>
 #include <ucs/arch/bitops.h>
 #include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 16: remove math.h?

src/ucs/datastruct/pgtable.c
```diff
@@ -15,6 +15,7 @@
 #include <ucs/debug/log.h>
 #include <ucs/debug/memtrack_int.h>
 #include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 18: remove math.h?

src/ucs/memory/memtype_cache.c
```diff
@@ -19,6 +19,7 @@
 #include <ucs/stats/stats.h>
 #include <ucs/sys/sys.h>
 #include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 22: remove math.h?

src/ucs/sys/ptr_arith.h
```diff
@@ -0,0 +1,100 @@
...
+*
+* See file LICENSE for terms.
+*/
+
+#ifndef UCS_PTR_ARITH_H
+#define UCS_PTR_ARITH_H
+
+#include "math.h"
+#include "compiler_def.h"
+#include <ucs/debug/assert.h>
```
yosefe, line 12: space line before

src/ucs/sys/ptr_arith.h
```diff
@@ -0,0 +1,100 @@
+/**
+* Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2024. ALL RIGHTS RESERVED.
+*
+* See file LICENSE for terms.
+*/
+
+#ifndef UCS_PTR_ARITH_H
+#define UCS_PTR_ARITH_H
+
+#include "math.h"
```
yosefe, line 10: why needed?

src/ucs/sys/ptr_arith.h
```diff
@@ -0,0 +1,100 @@
...
+#include "compiler_def.h"
+#include <ucs/debug/assert.h>
+
+#include <stddef.h>
+#include <stdint.h>
+
+BEGIN_C_DECLS
+
+/** @file ptr_arith.h */
```
yosefe, line 19: remove

src/ucs/sys/sys.c
```diff
@@ -13,6 +13,7 @@
 
 #include <ucs/algorithm/crc.h>
 #include <ucs/sys/checker.h>
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 16: why needed? it's already in sys.h

src/ucs/time/timer_wheel.c
```diff
@@ -14,6 +14,7 @@
 #include <ucs/debug/log.h>
 #include <ucs/debug/memtrack_int.h>
 #include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 17: remove math.h?

src/uct/base/uct_mem.c
```diff
@@ -14,6 +14,7 @@
 #include <ucs/arch/cpu.h>
 #include <ucs/profile/profile.h>
 #include <ucs/sys/math.h>
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 17: remove math.h?

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -4,6 +4,8 @@
  * See file LICENSE for terms.
  */
 
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 7: include it after ib_mlx5.h

test/gtest/uct/test_md.cc
```diff
@@ -12,6 +12,7 @@
 #include <uct/api/v2/uct_v2.h>
 extern "C" {
 #include <ucs/time/time.h>
+#include <ucs/sys/ptr_arith.h>
```
yosefe, line 15: remove math.h?

yosefe (commented):
in general maybe no need to add ptr_arith.h to so many places, and try removing math.h includes

src/ucs/sys/sys.c
Artemy-Mellanox: where in sys.h?

Artemy-Mellanox:
> in general maybe no need to add ptr_arith.h to so many places, and try removing math.h includes

yosefe (approved):
pls squash

yosefe (approved):

PR #9761, opened 2024-03-18
UCT/MD/IB: Fix indirect key creation for MT registered memory - v1.16.x

ivankochin:
## What
Backport of https://github.com/openucx/ucx/pull/9424


roiedanino:
> do we need MEMIC fixes here? cc @roiedanino

Yes, memic won't work without the fixes here: https://github.com/openucx/ucx/pull/9743

ivankochin:
> > do we need MEMIC fixes here? cc @roiedanino
> 
> Yes, memic won't work without the fixes here: #9743

@yosefe should the memic fix #9743 be added to this PR or backported later as a separate one?

yosefe:
after fixing MEMIC in master, need to add it to this PR

yosefe:
per offline discussion to minimize risk on v1.16.x release branch, will disable MT registration as a workaround

PR #9764, opened 2024-03-19
JUCX: Expose EP local socket address to jucx

ofirfarjun7:
## What
Expose EP local socket address in jucx

## Why ?
To allow configuration of local socket address from jucx

## How ?
Extend jucx API.


gleon99 (approved):

yosefe (commented):
can we add it to Jucx unit test?

ofirfarjun7:
Here UcpEndpointTest.java?
I think we can add something.

yosefe:
> Here `UcpEndpointTest.java`? I think we can add something.

Yes

PR #9767, opened 2024-03-20
UCT/IB/DEVX/MD: init atomic_rkey on mem_attach

evgeny-leksikov:
## What
init atomic_rkey on mem_attach

## Why ?
zcopy using imported memh can wrongly resolve rkey

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -738,6 +738,30 @@ static ucs_status_t uct_ib_mlx5_devx_dereg_mr(uct_ib_mlx5_md_t *md,
     }
 }
 
+static ucs_status_t uct_ib_mlx5_devx_memh_alloc(uct_ib_mlx5_md_t *md,
+                                                size_t length, unsigned flags,
+                                                size_t mr_size,
+                                                uct_ib_mlx5_devx_mem_t **memh_p)
+{
+    uct_ib_mlx5_devx_mem_t *memh;
+    uct_ib_mem_t *ib_memh;
+    ucs_status_t status;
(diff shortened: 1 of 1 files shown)


src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2251,7 +2256,9 @@ ucs_status_t uct_ib_mlx5_devx_mem_attach(uct_md_h uct_md,
         goto err;
     }
 
-    memh      = ucs_derived_of(ib_memh, uct_ib_mlx5_devx_mem_t);
+    memh = ucs_derived_of(ib_memh, uct_ib_mlx5_devx_mem_t);
+    uct_ib_mlx5_devx_mem_reset_keys(memh);
```
rakhmets, line 2260: What does the issue look like to the user?
Is it possible to add a test to cover the issue?

evgeny-leksikov: > What does the issue look like to the user?

Wrong rkey -> remote access error

> Is it possible to add a test to cover the issue?

Will be covered with GGA transport which uses attached memh for RMA

rakhmets (approved):

tvegas1 (approved):

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: Maybe add common function uct_ib_mlx5_devx_memh_alloc() tat will unite the common code in lines 2254-2260 and 760-767?

yosefe (approved):

rakhmets (approved):

PR #9772, opened 2024-03-21
UCP/PROTO: Improve logging of unimplemented abort/reset

yosefe:
## Why
Change error message from
` Fatal: abort request 0x123456 proto rndv/rts, not implemented`
to
`Fatal: 'abort' is not implemented for protocol rndv/rts (req: 0x123456)`

### DIFF
--- a/src/ucp/proto/proto_common.c
+++ b/src/ucp/proto/proto_common.c
@@ -845,8 +845,8 @@ ucs_status_t ucp_proto_request_zcopy_id_reset(ucp_request_t *req)
 static void ucp_proto_stub_fatal_not_implemented(const char *func_name,
                                                  ucp_request_t *req)
 {
-    ucs_fatal("%s request %p proto %s, not implemented", func_name, req,
-              req->send.proto_config->proto->name);
+    ucs_fatal("'%s' is not implemented for protocol %s (req: %p)", func_name,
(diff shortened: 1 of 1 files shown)


tvegas1 (approved):

src/ucp/proto/proto_common.c
```diff
@@ -845,8 +845,8 @@ ucs_status_t ucp_proto_request_zcopy_id_reset(ucp_request_t *req)
 static void ucp_proto_stub_fatal_not_implemented(const char *func_name,
                                                  ucp_request_t *req)
 {
-    ucs_fatal("%s request %p proto %s, not implemented", func_name, req,
-              req->send.proto_config->proto->name);
+    ucs_fatal("'%s' is not implemented for protocol %s (req: %p)", func_name,
```
ivankochin, line 848: Minor: `ucp_trace_req` has the following format: "req %p:  message text". Maybe it worth making this one consistent with `ucp_trace_req`?

ivankochin (approved):

brminich (approved):

src/ucp/proto/proto_common.c
yosefe: not sure, because this is a Fatal error message and not a regular log message

PR #9773, opened 2024-03-21
UCP/PROTOV2: Enable using device staging buffers for ppln protos

brminich:
## What
Allocate staging buffers for ppln protocols according to UCX_RNDV_FRAG_MEM_TYPE. (currently it is parsed by v1 protos only)

### DIFF
--- a/src/ucp/rndv/rndv_rtr.c
+++ b/src/ucp/rndv/rndv_rtr.c
@@ -393,6 +393,7 @@ ucp_proto_rndv_rtr_mtype_init(const ucp_proto_init_params_t *init_params)
     ucp_proto_rndv_rtr_priv_t *rpriv = init_params->priv;
     const uint64_t rndv_modes        = UCS_BIT(UCP_RNDV_MODE_PUT_PIPELINE);
     ucp_context_h context            = init_params->worker->context;
+    ucs_memory_type_t frag_mem_type  = context->config.ext.rndv_frag_mem_type;
     ucp_proto_perf_node_t *unpack_perf_node;
     ucs_linear_func_t unpack_time;
     ucp_md_map_t md_map, dummy_md_map;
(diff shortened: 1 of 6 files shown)


src/ucp/rndv/rndv_rtr.c
```diff
@@ -410,25 +411,31 @@ ucp_proto_rndv_rtr_mtype_init(const ucp_proto_init_params_t *init_params)
...
-    status = ucp_mm_get_alloc_md_index(context, &md_index);
-    if (status != UCS_OK) {
-        return status;
+
+    if (frag_mem_type == UCS_MEMORY_TYPE_HOST) {
+        /* Make sure that key of the md used to allocate a bounce buffer will be
+         * packed to the RTR rkey. It can be used by 2-stage pipeline
+         * (rkey_ptr_mtype) protocol.
+         */
+        status = ucp_mm_get_alloc_md_index(context, &md_index);
```
ivankochin, line 427: As I can see,  that is the only place where `ucp_mm_get_alloc_md_index` is called.  And it assumes that `alloc_mem_type` is host. I would suggest to add `alloc_mem_type` argument for this function and return 0 md_index in case if it is not host.

brminich: we could extend `ucp_mm_get_alloc_md_index` to support other memory types, but in this case i want to explicitly initialize that md_map for host chunks only, because 2-stage ppln does not support using other memory types chunks

brminich: currently we do not need any other mem type chunk in `ucp_mm_get_alloc_md_index`.
Imo the code is easier to read and less error prone when we clearly pack md_index when it is needed only.
What is the reason to extend it if it is not supposed to be used?

yosefe: because we don't want to create dependency between different protocols, and have less "special cases" in the code

yosefe: also, seems extending ucp_mm_get_alloc_md_index would be quite straight forward:
https://github.com/openucx/ucx/blob/1960e7c0b1d852255bcd2a793d7c8c0ed7a5de07/src/ucp/core/ucp_mm.c#L1567

yosefe (approved):

PR #9776, opened 2024-03-22
CONFIG: Moved Fujitsu ARM specific values to ucx.conf.

rakhmets:
## What
Added section for configuration values for Fujitsu ARM to `ucx.conf`.
Added `rkey_ptr` sub-option to `UCX_PROTO_OVERHEAD` configuration variable.


src/ucp/core/ucp_context.c
```diff
@@ -60,7 +60,10 @@
 
 #define UCP_AM_HANDLER_ENTRY(_id) [_id] = &ucp_am_handler_##_id,
 
-#define UCP_CPU_EST_BCOPY_BW_DEFAULT (5800 * UCS_MBYTE)
+
+#define UCP_CPU_EST_BCOPY_BW_DEFAULT_NEW (7000 * UCS_MBYTE)
+#define UCP_CPU_EST_BCOPY_BW_DEFAULT     (5800 * UCS_MBYTE)
```
brminich, line 65: maybe:
```suggestion
#define UCP_CPU_EST_BCOPY_BW_DEFAULT    (7000 * UCS_MBYTE)
#define UCP_CPU_EST_BCOPY_BW_DEFAULT_LEGACY     (5800 * UCS_MBYTE)
```

src/uct/ib/base/ib_iface.h
```diff
@@ -31,6 +31,26 @@
 #define UCT_IB_COUNTER_SET_ID_INVALID      UINT8_MAX
 
 
+#define UCT_IB_SEND_OVERHEAD_BCOPY     "bcopy"
+#define UCT_IB_SEND_OVERHEAD_SQE       "sqe"
+#define UCT_IB_SEND_OVERHEAD_DB        "db"
+#define UCT_IB_SEND_OVERHEAD_WQE_FETCH "wqe_fetch"
+#define UCT_IB_SEND_OVERHEAD_WQE_POST  "wqe_post"
```
brminich, line 38: add `_STR` postfix?

src/uct/ib/rc/verbs/rc_verbs_iface.c
```diff
@@ -230,7 +234,9 @@ uct_rc_verbs_iface_query(uct_iface_h tl_iface, uct_iface_attr_t *iface_attr)
 
     iface_attr->cap.flags |= UCT_IFACE_FLAG_EP_CHECK;
     iface_attr->latency.m += 1e-9;  /* 1 ns per each extra QP */
-    iface_attr->overhead   = 75e-9; /* Software overhead */
+
+    /* Software overhead */
+    iface_attr->overhead = UCT_RC_VERBS_IFACE_OVERHEAD;
```
brminich, line 239: we need to take  it from config value (in case it is modified for some platform)

src/uct/ib/ud/accel/ud_mlx5.c
```diff
@@ -627,7 +630,7 @@ uct_ud_mlx5_iface_query(uct_iface_h tl_iface, uct_iface_attr_t *iface_attr)
         return status;
     }
 
-    iface_attr->overhead = 80e-9; /* Software overhead */
+    iface_attr->overhead = UCT_UD_MLX5_IFACE_OVERHEAD; /* Software overhead */
```
brminich, line 633: we need to take  it from config value (in case it is modified for some platform)

src/uct/ib/ud/verbs/ud_verbs.c
```diff
@@ -535,7 +539,7 @@ uct_ud_verbs_iface_query(uct_iface_h tl_iface, uct_iface_attr_t *iface_attr)
         return status;
     }
 
-    iface_attr->overhead = 105e-9; /* Software overhead */
+    iface_attr->overhead = UCT_UD_VERBS_IFACE_OVERHEAD; /* Software overhead */
```
brminich, line 542: we need to take  it from config value (in case it is modified for some platform)

brminich (commented):
failures seem relevant

src/uct/ib/rc/verbs/rc_verbs_iface.c
rakhmets: It will brake wire compatibility, won't it?

src/ucp/core/ucp_context.c
rakhmets: Updated names.

src/ucp/core/ucp_context.c
```diff
@@ -61,7 +61,10 @@
 
 #define UCP_AM_HANDLER_ENTRY(_id) [_id] = &ucp_am_handler_##_id,
 
-#define UCP_CPU_EST_BCOPY_BW_DEFAULT (5800 * UCS_MBYTE)
+
```
yosefe, line 64: remove spaceline

src/ucp/core/ucp_context.c
```diff
@@ -1925,7 +1921,9 @@ ucp_fill_rndv_frag_config(const ucp_context_config_names_t *config,
 
 static double ucp_context_get_memcpy_bw()
 {
-    return ucp_context_est_bcopy_bw[ucs_arch_get_cpu_vendor()];
+    return (ucs_arch_get_cpu_vendor() == UCS_CPU_VENDOR_AMD) ?
```
yosefe, line 1924: maybe rename this function to ucp_context_get_protov1_memcpy_bw() ?

src/uct/base/uct_iface.h
```diff
@@ -248,6 +248,9 @@ enum {
     UCT_CHECK_PARAM((_flags) == 0, "Unsupported flags: %x", (_flags));
 
 
+#define UCT_IFACE_SEND_OVERHEAD_STR "SEND_OVERHEAD"
```
yosefe, line 251: IMO it's overkill to define constant for this string

src/uct/ib/base/ib_iface.c
```diff
@@ -205,19 +205,22 @@ ucs_config_field_t uct_ib_iface_config_table[] = {
...
-    {"cqe", "estimated overhead of processing a work request completion",
+    {UCT_IB_SEND_OVERHEAD_SQE_STR,
+     "estimated overhead of processing a work request completion",
      ucs_offsetof(uct_ib_iface_config_t, send_overhead.cqe)},
-    {"db", "estimated overhead of writing a doorbell to PCI",
+    {UCT_IB_SEND_OVERHEAD_DB_STR,
+     "estimated overhead of writing a doorbell to PCI",
      ucs_offsetof(uct_ib_iface_config_t, send_overhead.db)},
-    {"wqe_fetch", "estimated overhead of fetching a wqe",
+    {UCT_IB_SEND_OVERHEAD_WQE_FETCH_STR, "estimated overhead of fetching a wqe",
```
yosefe, line 221: IMO it's overkill to define constants for these strings

yosefe (commented):
can you pls confirm `ucx_info -d` output is not changed after this PR?

src/uct/ib/rc/verbs/rc_verbs_iface.c
brminich: we have to decide what to do with wire compat, but we need to take this value from the config, otherwise the config per platform feature does not work

rakhmets: Why doesn't it work?
We fill `uct_perf_attr_t` using `uct_iface_estimate_perf`.
Previously we set `perf_attr->send_pre_overhead` to `iface_attr.overhead`.
After this changes we will set this value according the value in the configuration. But I left `iface_attr.overhead` unchanged. Why do you think this is not correct?

rakhmets:
> @rakhmets pls avoid force push during review, use a merge commit if have conflicts

The latest force-push has no changes. But the commits are different. There is a merge of master branch instead of rebasing. And the latest fixes of review comments are in the commit after merging.

yosefe (approved):

brminich (dismissed):

PR #9780, opened 2024-03-26
UCP/WIREUP: Calculate RMA score with regard to local distance BW - v1.16.x

ivankochin:
## What
Backport of https://github.com/openucx/ucx/pull/9474

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -984,13 +984,20 @@ static double ucp_wireup_rma_score_func(const ucp_worker_iface_t *wiface,
                                         void *arg)
 {
     /* best for 4k messages */
+    double local_bw;
+
+    if (unpacked_addr->dst_version < 17) {
+        local_bw = ucp_tl_iface_bandwidth(wiface->worker->context,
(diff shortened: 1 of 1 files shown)


yosefe (approved):

src/ucp/wireup/select.c
```diff
@@ -984,13 +984,20 @@ static double ucp_wireup_rma_score_func(const ucp_worker_iface_t *wiface,
...
 {
     /* best for 4k messages */
+    double local_bw;
+
+    if (unpacked_addr->dst_version < 17) {
+        local_bw = ucp_tl_iface_bandwidth(wiface->worker->context,
+                                          &wiface->attr.bandwidth);
+    } else {
+        local_bw = ucp_wireup_iface_bw_distance(wiface);
+    }
```
brminich, line 994: but this is for version 1.16, why need to compare with 17?
I think in version 1.17 we expect 1.16 to always do `ucp_tl_iface_bandwidth`

ivankochin:
This fix shouldn't be merged since it would break wire compatibility

PR #9781, opened 2024-03-26
UCP/WIREUP: Consider local distance during slow lanes dropping - v1.16.x - revert

ivankochin:
## What
Reverting the backport of https://github.com/openucx/ucx/pull/9512 since it breaks wire compatibility.

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1461,18 +1461,12 @@ static double ucp_wireup_get_lane_bw(ucp_worker_h worker,
                                      const ucp_wireup_select_info_t *sinfo,
                                      const ucp_unpacked_address_t *address)
 {
-    ucp_worker_iface_t *wiface = ucp_worker_iface(worker, sinfo->rsc_index);
+    ucp_context_h context = worker->context;
+    const uct_iface_attr_t *iface_attr;
     double bw_local, bw_remote;
 
-    if (address->dst_version < 17) {
-        bw_local = ucp_tl_iface_bandwidth(worker->context,
-                                          &wiface->attr.bandwidth);
-    } else {
-        /* Compare BW including local distance to prevent EP reconfiguration
-         * since remote distance is included to remote_addr*/
-        bw_local = ucp_wireup_iface_bw_distance(wiface);
-    }
-
+    iface_attr = ucp_worker_iface_get_attr(worker, sinfo->rsc_index);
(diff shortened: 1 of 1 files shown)


yosefe:
Do we need to revert it on master branch?
How did you find the issue? Why we didn't catch it in CI?

ivankochin:
> Do we need to revert it on master branch?
No, we didn't, the fix works OK on master.

> How did you find the issue?
It was just a logical conclusion. The original fixes that works for master checked that UCX version is >= 17 and then applied new logic, with older UCX version it leaves the logic as is to avoid different lanes performance on old UCX. If we somehow change that for UCX 1.16, it will break the logic for master branch.

> Why we didn't catch it in CI?
I think this error can require some specific setup to be reproduced.

yosefe:
> I think this error can require some specific setup to be reproduced.

Maybe because CI doesn't test current v1.16.x vs master branch?
@brminich WDYT?

ivankochin:
> > I think this error can require some specific setup to be reproduced.
> 
> Maybe because CI doesn't test current v1.16.x vs master branch? @brminich WDYT?

Yes, that should be the case, but also it should be the system with several NICs and these NICs should have different distance to GPU.

The original issue can be reproduced on DGX1

yosefe:
ok
but probably need to fix the commit title by force push

ivankochin:
> ok but probably need to fix the commit title by force push

What's wrong with the commit title?

brminich:
> > I think this error can require some specific setup to be reproduced.
> 
> Maybe because CI doesn't test current v1.16.x vs master branch? @brminich WDYT?

we do test it https://github.com/openucx/ucx/blob/master/buildlib/pr/wire_compat.yml#L11
i think we need a specific server config to get it

brminich (approved):

yosefe:
> > > I think this error can require some specific setup to be reproduced.
> > 
> > 
> > Maybe because CI doesn't test current v1.16.x vs master branch? @brminich WDYT?
> 
> we do test it https://github.com/openucx/ucx/blob/master/buildlib/pr/wire_compat.yml#L11 i think we need a specific server config to get it

I think in the current CI config - master PR would have failed after we merged something bad to v1.16.x
but we need v1.16.x to fail. maybe need to test master branch vs. release **tags**, and release branches vs. master?

ivankochin:
@yosefe pls review

yosefe (approved):

PR #9782, opened 2024-03-26
UCP/PROTO: Add latency conversion to nsec before FP8 pack/unpack

ivankochin:
## What
Add conversion of latency to nanoseconds to make FP8 pack/unpack correctly

## Why ?
Latency FP8 pack/unpack supports value represented in nanoseconds

Latency value without this patch after FP8 pack 😄 :
image

### DIFF
--- a/src/ucp/proto/proto_common.c
+++ b/src/ucp/proto/proto_common.c
@@ -128,7 +128,8 @@ ucp_proto_common_get_sys_dev(const ucp_proto_init_params_t *params,
 static void
 ucp_proto_common_fp8_pack_unpack_distance(ucs_sys_dev_distance_t *distance)
 {
-    distance->latency   = UCS_FP8_PACK_UNPACK(LATENCY, distance->latency);
+    double nsec         = distance->latency * UCS_NSEC_PER_SEC;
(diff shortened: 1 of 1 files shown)


ivankochin:
@yosefe pls review

yosefe (approved):

yosefe:
@ivankochin do we need a backport to v1.16.x?

ivankochin:
> @ivankochin do we need a backport to v1.16.x?

No since the code that is fixed by this PR exists only in master

Artemy-Mellanox:
@ivankochin @yosefe shouldn't `ucp_wireup_tl_iface_latency()` be fixed in the same way? Looks like it converting `lat_lossy` to nanoseconds but how does it convert `local_lat` to seconds?

ivankochin:
> @ivankochin @yosefe shouldn't `ucp_wireup_tl_iface_latency()` be fixed in the same way? Looks like it converting `lat_lossy` to nanoseconds but how does it convert `local_lat` to seconds?

Thanks for noticing that, you are definitely right, here is the fix: https://github.com/openucx/ucx/pull/9792

PR #9795, opened 2024-04-02
BINDINGS/GO: Fixed dependencies in Makefile.

rakhmets:
## What
Fixed dependencies in Makefile.


yosefe (commented):
can we add this to "build" stage CI - so the test will reproduce the build issue before this fix?

rakhmets:
> can we add this to "build" stage CI - so the test will reproduce the build issue before this fix?

I now doubt that this fixes the compilation issue. Need to check this.

PR #9800, opened 2024-04-03
UCP/PROTO: Remove redundant frag_size argument from parallel_stages()

ivankochin:
## What
Remove `frag_size` argument from `ucp_proto_init_parallel_stages` API.

## Why?
Since this argument participates only in logs and perf tree information. Can be removed since `ucp_proto_common_add_ppln_range` already prints info about fragment size.

### DIFF
--- a/src/ucp/proto/proto_init.c
+++ b/src/ucp/proto/proto_init.c
@@ -187,7 +187,7 @@ ucp_proto_perf_envelope_make(const ucp_proto_perf_list_t *perf_list,
 
 ucs_status_t
 ucp_proto_init_parallel_stages(const char *proto_name, size_t range_start,
-                               size_t range_end, size_t frag_size, double bias,
+                               size_t range_end, double bias,
                                const ucp_proto_perf_range_t **stages,
                                unsigned num_stages, ucp_proto_caps_t *caps)
 {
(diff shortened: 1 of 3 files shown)


src/ucp/proto/proto_init.c
```diff
@@ -665,9 +656,8 @@ ucp_proto_init_single_frag_ranges(const ucp_proto_common_init_params_t *params,
     parallel_stages[2] = &recv_perf;
 
     /* Add ranges representing sending single fragment */
-    status = ucp_proto_init_parallel_stages(proto_name, range_start, frag_size,
-                                            frag_size, 0.0, parallel_stages, 3,
-                                            caps);
+    status = ucp_proto_init_parallel_stages(proto_name, range_start, range_end,
```
yosefe, line 659: seems we had a bug when passed frag_size as range_end?

yosefe (dismissed):

ivankochin:
@yosefe approve was removed due to merge commit (there were merge conflicts). Could you please review it one more time?

yosefe (approved):

brminich (approved):

yosefe:
@ivankochin pls squash

PR #9802, opened 2024-04-07
UCP/TAG: Call user callback from tag_recv only if NO_IMM_CMPL is passed

yosefe:
## Why
Fix JVM segfault in Java CI tests when gdrcopy driver is not loaded

## How
ucp_tag_recv_nbx() can call the user callback when the request is completed immediately (with self transport), even though Jucx does not pass NO_IMM_CMPL flag in request params. This leads to calling set_native_id method on a released object.

The segfault looks like this:
```
Stack: [0x00007ffa692de000,0x00007ffa693de000],  sp=0x00007ffa693da420,  free space=1009k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [libjvm.so+0x6d2a6f]  jni_invoke_nonstatic(JNIEnv_*, JavaValue*, _jobject*, JNICallType, _jmethodID*, JNI_ArgumentPusher*, Thread*)+0x1af
V  [libjvm.so+0x6d709f]  jni_CallVoidMethodV+0x10f
C  [libjucx_amd64.so+0x61b1]  JNIEnv_::CallVoidMethod(_jobject*, _jmethodID*, ...)+0x81
C  [libjucx_amd64.so+0x7706]  process_request(JNIEnv_*, ucp_request_param_t const*, void*)+0x196
C  [libjucx_amd64.so+0xa6bb]  Java_org_openucx_jucx_ucp_UcpWorker_recvTaggedNonBlockingNative+0xdb
j  org.openucx.jucx.ucp.UcpWorker.recvTaggedNonBlockingNative(JJJJJLorg/openucx/jucx/UcxCallback;Lorg/openucx/jucx/ucp/UcpRequestParams;)Lorg/openucx/jucx/ucp/UcpRequest;+0
j  org.openucx.jucx.ucp.UcpWorker.recvTaggedNonBlocking(JJJJLorg/openucx/jucx/UcxCallback;Lorg/openucx/jucx/ucp/UcpRequestParams;)Lorg/openucx/jucx/ucp/UcpRequest;+17
j  org.openucx.jucx.ucp.UcpWorker.recvTaggedNonBlocking(JJJJLorg/openucx/jucx/UcxCallback;)Lorg/openucx/jucx/ucp/UcpRequest;+10
j  org.openucx.jucx.ucp.UcpWorker.recvTaggedNonBlocking(Ljava/nio/ByteBuffer;JJLorg/openucx/jucx/UcxCallback;)Lorg/openucx/jucx/ucp/UcpRequest;+32
j  org.openucx.jucx.UcxTest$MemoryBlock.getData()Ljava/nio/ByteBuffer;+61
j  org.openucx.jucx.UcpEndpointTest.testActiveMessages(I)V+540
```

### DIFF
--- a/src/ucp/core/ucp_request.inl
+++ b/src/ucp/core/ucp_request.inl
@@ -264,6 +264,8 @@ ucp_request_complete_tag_recv(ucp_request_t *req, ucs_status_t status)
                   req->recv.tag.info.sender_tag, req->recv.tag.info.length,
                   ucs_status_string(status));
     UCS_PROFILE_REQUEST_EVENT(req, "complete_tag_recv", status);
+    /* coverity[address_free] */
+    /* coverity[offset_free] */
     ucp_request_complete(req, recv.tag.cb, status, &req->recv.tag.info,
                          req->user_data);
 }
--- a/src/ucp/tag/tag_recv.c
+++ b/src/ucp/tag/tag_recv.c
@@ -60,8 +60,6 @@ static UCS_F_ALWAYS_INLINE ucs_status_ptr_t ucp_tag_recv_common(
         ucp_tag_t tag_mask, ucp_request_t *req, ucp_recv_desc_t *rdesc,
         const ucp_request_param_t *param, const char *debug_name)
 {
-    uint32_t req_flags = (param->op_attr_mask & UCP_OP_ATTR_FIELD_CALLBACK) ?
-                         UCP_REQUEST_FLAG_CALLBACK : 0;
     ucp_request_queue_t *req_queue;
     size_t hdr_len, recv_len;
     ucs_status_t status;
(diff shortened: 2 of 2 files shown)


src/ucp/tag/tag_recv.c
```diff
@@ -143,16 +129,27 @@ static UCS_F_ALWAYS_INLINE ucs_status_ptr_t ucp_tag_recv_common(
     req->recv.op_attr       = param->op_attr_mask;
     req->recv.tag.tag       = tag;
     req->recv.tag.tag_mask  = tag_mask;
-    if (param->op_attr_mask & UCP_OP_ATTR_FIELD_CALLBACK) {
-        req->recv.tag.cb = param->cb.recv;
```
tvegas1, line 147: to confirm, `req->recv.tag.cb` is never checked for NULL/used unless `UCP_REQUEST_FLAG_CALLBACK` is set on `req->flags`?

src/ucp/core/ucp_request.inl
```diff
@@ -264,6 +264,8 @@ ucp_request_complete_tag_recv(ucp_request_t *req, ucs_status_t status)
                   req->recv.tag.info.sender_tag, req->recv.tag.info.length,
                   ucs_status_string(status));
     UCS_PROFILE_REQUEST_EVENT(req, "complete_tag_recv", status);
+    /* coverity[address_free] */
```
tvegas1, line 267: for which field are those two comments for?

tvegas1 (commented):
would adding a gtest forcing eager or using small size, with/without no immediate completion make sense, or you think it would be too brittle?

src/ucp/tag/tag_recv.c
yosefe: yes, if UCP_REQUEST_FLAG_CALLBACK  is not set the CB should not be accessed at all

src/ucp/core/ucp_request.inl
yosefe: there are bogus Coverity errors that calling a completion callback may release the request (because coverity thinks the completion callback can resolve to some internal function such as ucp_cm_server_conn_request_progress

yosefe:
> would adding a gtest forcing eager or using small size, with/without no immediate completion make sense, or you think it would be too brittle?

Tried to write a gtest (expect that send/recv callbacks are not called), but so far it didn't reproduce the issue. So IMO better merge the fix first to resolve CI failures and fix UCP recv behavior

brminich (approved):

src/ucp/tag/tag_recv.c
```diff
@@ -143,16 +129,27 @@ static UCS_F_ALWAYS_INLINE ucs_status_ptr_t ucp_tag_recv_common(
...
+                                 ucp_tag_rndv_rts_from_rdesc(rdesc),
+                                 rdesc->length);
+            UCP_WORKER_STAT_RNDV(worker, RX_UNEXP, 1);
+            ucp_recv_desc_release(rdesc);
+        } else {
+            ucp_tag_recv_eager_multi(worker, req, rdesc);
+        }
+
+        if (req->flags & UCP_REQUEST_FLAG_COMPLETED) {
+            goto out_completed;
```
tvegas1, line 150: i think we need to set `status = req->status` here, because we return `status` at the end of the function at line 184 (and also use it in case we need to set `param->recv.tag.info`)

yosefe: thanks, good catch!

yosefe: fixed by force push due to small pr

brminich (approved):

tvegas1 (approved):

PR #9370, opened 2023-09-13
UCT/API: Introduce migratable_mem_types

Akshay-Venkatesh:
## What/Why ?
Introduce migratable_mem_types field that is populated by MDs. These mem_types would include the list of the memory types whose pages can migrate between host and PCIe device memory. When the MDs identify specific allocation types are migratable, UCP can add such memory types to a list which would be candidates for nonblocking registration (for example through ODP under IB transport). Today nonblocking registration is applied to memory types listed in reg_nb_mem_types field of UCP context's external config variable but with MDs identifying which memory types migrate, the decision is dynamically made by MDs.

## How ?
Instead of having the user list memory types which need to go through on-demand registration for pages that migrate, instead users have decide if migratable memory should be registered (default is to let the runtime decide based on the platform) using `UCX_REG_MIGRATABLE_MEM` env var. `UCX_REG_MIGRATABLE_MEM=off` would imply that none of the memory types would undergo non-blocking registration; `UCX_REG_MIGRATABLE_MEM=on` would imply that all memory types that can be registered without explicit pinning will be considered. 

## Next steps
This PR doesn't add capability in different MDs (like cuda_copy) to detect the specific system and populate migratable_mem_types. This can be done potentially using capabilities added in https://github.com/openucx/ucx/pull/9314 or by directly making query calls in the specific MD in a follow up PR.

PR #9805, opened 2024-04-08
UCP/CONTEXT: Add host,cuda_managed to ODP memtypes if Grace CPU is detected

Akshay-Venkatesh:
## What
Simpler PR to enable ODP for host/cuda_managed memory on Grace compared to https://github.com/openucx/ucx/pull/9370

## Out-of-scope and deferred to sysadmins:
1. Detecting if ODPv2 is enabled on the system
2. Detecting if Hopper(+) is present
3. Other checks to ensure host and cuda_managed memory actually works correctly with ODPv2

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -16,6 +16,7 @@
 #include <ucs/config/parser.h>
 #include <ucs/algorithm/crc.h>
 #include <ucs/arch/atomic.h>
+#include <ucs/arch/cpu.h>
 #include <ucs/datastruct/mpool.inl>
 #include <ucs/datastruct/queue.h>
 #include <ucs/datastruct/string_set.h>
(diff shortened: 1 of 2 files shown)


Akshay-Venkatesh:
@yosefe @brminich I created this alternative PR to simplify review and enable ODPv2 for host/cuda_managed memory on GH.

src/ucp/core/ucp_context.c
```diff
@@ -476,6 +477,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, reg_nb_mem_types),
    UCS_CONFIG_TYPE_BITMAP(ucs_memory_type_names)},
 
+  {"POPULATE_NONBLOCK_MEMTYPES", "try",
```
brminich, line 480: we already have REG_NONBLOCK_MEM_TYPES for this purpose.
We thought to add platform specific config, similar to what @rakhmets implemented in #9776

yosefe:
@Akshay-Venkatesh do we still need this PR?

PR #9806, opened 2024-04-09
JUCX/WORKER: Support null request in progressRequest()

yosefe:
## Why
Fix null pointer exception in Jucx CI

### DIFF
--- a/bindings/java/src/main/java/org/openucx/jucx/ucp/UcpWorker.java
+++ b/bindings/java/src/main/java/org/openucx/jucx/ucp/UcpWorker.java
@@ -143,7 +143,7 @@ public class UcpWorker extends UcxNativeStruct implements Closeable {
      * Blocking progress for request until it's not completed.
      */
     public void progressRequest(UcpRequest request) throws Exception {
-        while (!request.isCompleted()) {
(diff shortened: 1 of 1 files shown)


yosefe:
@rakhmets can you pls review?

brminich (approved):

rakhmets:
Is this a new issue? Why do we need to allow the user to pass null object to the method?

tvegas1 (approved):

yosefe:
> Is this a new issue? Why do we need to allow the user to pass null object to the method?

This issue was introduced by my recent fix of waiting for requests completion in Java unit tests. An alternative way would be to check for null before calling progressRequest method, to handle immediate completion.

yosefe:
> > > Is this a new issue? Why do we need to allow the user to pass null object to the method?
> > 
> > 
> > This issue was introduced by my recent fix of waiting for requests completion in Java unit tests. An alternative way would be to check for null before calling progressRequest method, to handle immediate completion.
> 
> Maybe it is better to fix it in the test than changing the implementation of the API method?

Will add helper function in the test then

yosefe:
The fix is actually not relevant. Jucx always returns a request object and always calls user callback in case of an immediate completion

PR #9807, opened 2024-04-09
CONFIG: Added configuration file with section for Grace Hopper.

rakhmets:
## What
Added configuration file with section for Grace Hopper, setting `REG_NONBLOCK_MEM_TYPES=host,cuda-managed`.

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -1,0 +1,10 @@
+#
+# Copyright (C) 2024 NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+#
+# See file LICENSE for terms.
+#
+
+conf_filesdir = $(ucx_config_dir)
+conf_files_DATA = ucx.conf
+
+EXTRA_DIST = $(conf_files_DATA)
--- a/config/ucx.conf
+++ b/config/ucx.conf
@@ -1,0 +1,3 @@
+[Grace Hopper]
+Product name=*Grace*Hopper*
+UCX_REG_NONBLOCK_MEM_TYPES=host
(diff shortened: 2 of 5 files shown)


config/ucx.conf
```diff
@@ -0,0 +1,3 @@
+[Grace Hopper]
+Product name=*Grace Hopper*
```
brminich, line 2: maybe this
```suggestion
Product name=*Grace*Hopper*
```

brminich (dismissed):

config/Makefile.am
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (C) 2024 NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+#
+# See file LICENSE for terms.
+#
+
+conf_files_dir = $(ucx_config_dir)
+conf_files__DATA = ucx.conf
```
yosefe, line 8: i think need single underscore, not a double

config/ucx.conf
```diff
@@ -0,0 +1,3 @@
+[Grace Hopper]
+Product name=*Grace*Hopper*
+UCX_REG_NONBLOCK_MEM_TYPES=host,cuda-managed
```
yosefe, line 3: why cuda-managed?

config/Makefile.am
rakhmets: With single it will be:
```
conf_filesdir = $(ucx_config_dir)
conf_files_DATA = ucx.conf
```

yosefe: IMO it is better

config/ucx.conf
rakhmets: Removed.

yosefe (approved):

brminich (approved):

PR #9810, opened 2024-04-10
UCT/DC: Report max_rd_atomic issue as diag instead of error

brminich:
## What
Report max_rd_atomic inconsistency as a diag rather than error message

### DIFF
--- a/src/uct/ib/dc/dc_mlx5_ep.c
+++ b/src/uct/ib/dc/dc_mlx5_ep.c
@@ -1270,9 +1270,9 @@ UCS_CLASS_INIT_FUNC(uct_dc_mlx5_ep_t, uct_dc_mlx5_iface_t *iface,
 
     if ((iface->super.super.config.max_rd_atomic == 16) !=
         !!(if_addr->flags & UCT_DC_MLX5_IFACE_ADDR_MAX_RD_ATOMIC_16)) {
-        ucs_error("max_rd_attomic values do not match on peers (local is %u), "
-                  "set UCX_RC_MAX_RD_ATOMIC=16 to resolve this issue.",
-                  iface->super.super.config.max_rd_atomic);
+        ucs_diag("max_rd_atomic values do not match on peers (local is %u), "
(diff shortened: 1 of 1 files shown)


src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1270,9 +1270,9 @@ UCS_CLASS_INIT_FUNC(uct_dc_mlx5_ep_t, uct_dc_mlx5_iface_t *iface,
 
     if ((iface->super.super.config.max_rd_atomic == 16) !=
         !!(if_addr->flags & UCT_DC_MLX5_IFACE_ADDR_MAX_RD_ATOMIC_16)) {
-        ucs_error("max_rd_attomic values do not match on peers (local is %u), "
```
tvegas1, line 1273: we can also fix the typo

brminich: fixed

src/uct/ib/dc/dc_mlx5_ep.c
```diff
@@ -1270,9 +1270,9 @@ UCS_CLASS_INIT_FUNC(uct_dc_mlx5_ep_t, uct_dc_mlx5_iface_t *iface,
...
-                  "set UCX_RC_MAX_RD_ATOMIC=16 to resolve this issue.",
-                  iface->super.super.config.max_rd_atomic);
+        ucs_diag("max_rd_atomic values do not match on peers (local is %u), "
+                 "set UCX_RC_MAX_RD_ATOMIC=16 to resolve this issue.",
```
tvegas1, line 1274: sorry just noticed, we should also change `UCX_RC_MAX_RD_ATOMIC` to `UCX_DC_MLX5_MAX_RD_ATOMIC`, as this is DC specific?

brminich: UCX_RC_MAX_RD_ATOMIC is a common name for both rc and dc, but you are right that it's better to mention just DC related var

yosefe (approved):

PR #9815, opened 2024-04-11
UCP/PROTO: Simplify RNDV_SCHEME env logic

ivankochin:
## What
Fail protocol initialization if `RNDV_SCHEME` doesn't contain it's RNDV mode.

## Why
This simplifies proto initialization logic which significantly affects logic for https://github.com/openucx/ucx/pull/9728


test/gtest/ucp/ucp_test.cc
```diff
@@ -462,6 +462,15 @@ int ucp_test::get_variant_thread_type() const {
     return GetParam().variant.thread_type;
 }
 
+bool ucp_test::supports_caps(uint64_t caps) const {
+    for (const auto &entity : entities()) {
+        if (!entity->has_lane_with_caps(caps)) {
+            return false;
+        }
+    }
+    return true;
```
rakhmets, line 471:
```suggestion
    return std::all_of(
            entities().begin(), entities().end(),
            [caps] (const entity *e) { return e->has_lane_with_caps(caps); });
```

test/gtest/ucp/test_ucp_tag_xfer.cc
```diff
@@ -808,199 +818,210 @@ UCS_TEST_P(test_ucp_tag_xfer, send_contig_recv_contig_exp_rndv_probe, "RNDV_THRE
...
+UCP_TEST_TAG_RNDV_GENERIC(test_ucp_tag_xfer,
+                          send_contig_recv_generic_unexp_rndv_truncated,
+                          "ZCOPY_THRESH=1248576") {
     test_run_xfer(true, false, false, false, true);
 }
 
-UCS_TEST_P(test_ucp_tag_xfer, send_contig_recv_generic_unexp_sync_rndv,
-           "RNDV_THRESH=1000", "ZCOPY_THRESH=1248576") {
+UCP_TEST_TAG_RNDV_GENERIC(test_ucp_tag_xfer, send_contig_recv_generic_unexp_sync_rndv,
+                         "ZCOPY_THRESH=1248576") {
```
iyastreb, line 955: minor: aligment

test/gtest/ucp/ucp_test.h
```diff
@@ -333,6 +333,9 @@ class ucp_test : public ucp_test_base,
     // Return context parameters of the current test variant
     const ucp_params_t& get_variant_ctx_params() const;
 
+    // Check whether sender and receiver supports certain caps
+    bool supports_caps(uint64_t rndv_modes) const;
```
iyastreb, line 337: minor: maybe name it `support_caps` or `is_support_caps`?
It's unusual to have `s` suffix for verbs in function names. E.g. `free_memory`, not `frees_memory`

src/ucp/rndv/rndv_get.c
```diff
@@ -30,7 +30,7 @@ ucp_proto_rndv_get_common_init(const ucp_proto_init_params_t *init_params,
     ucp_context_t *context               = init_params->worker->context;
     ucp_proto_multi_init_params_t params = {
         .super.super         = *init_params,
-        .super.cfg_thresh    = ucp_proto_rndv_cfg_thresh(context, rndv_modes),
+        .super.cfg_thresh    = UCS_MEMUNITS_AUTO,
```
iyastreb, line 33: Q: with older implementation, the resulting threshold could be 0 or AUTO, while with the new one it is always AUTO.
Is this change on purpose? May it introduce regressions?

ivankochin: This change is why this patch was created. In master branch if RNDV_SHCEME is set to certain RNDV mode, and this protocol is unavailable, UCX falls back to other RNDV.

It was decide to remove this fall back logic and just don't initialize RNDV modes which are not set in RNDV_SCHEME.

test/gtest/ucp/ucp_test.h
ivankochin: Done.

Issue #9482, opened 2023-11-09
Failed to build: "cannot define new methods on non-local type *C.ucp_request_param_t"


banana-bred:
### Describe the bug
UCX fails to build.

### Steps to Reproduce
Followed the build instructions
```
$ ./autogen.sh
$ ./contrib/configure-release --prefix=$(pwd)/RightHere
$ make -j8
<SNIP>
cd /home/user/.cache/paru/clone/openucx/src/ucx/bindings/go/src/ucx ;\
go build
# ucx
./request.go:32:10: cannot define new methods on non-local type *C.ucp_request_param_t
# ucx
../../ucx/request.go:32:10: cannot define new methods on non-local type *C.ucp_request_param_t
make[2]: *** [Makefile:622: build] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: *** [Makefile:639: goperftest] Error 1
make[2]: Leaving directory '/home/user/.cache/paru/clone/openucx/src/ucx/bindings/go'
make[1]: *** [Makefile:769: all-recursive] Error 1
make[1]: Leaving directory '/home/user/.cache/paru/clone/openucx/src/ucx'
make: *** [Makefile:635: all] Error 2
```

### Setup and versions
OS: Artix Linux x86_64

### Additional information 
```
$ echo go glibc binutils openmpi gcc rdma-core numactl | tr ' ' '\n' | xargs -I{} sh -c 'echo {} $(pacman -Qi {}  | grep Version)' | column -t
go         Version  :  2:1.21.4-1
glibc      Version  :  2.38-7
binutils   Version  :  2.41-3
openmpi    Version  :  4.1.5-5
gcc        Version  :  13.2.1-3
rdma-core  Version  :  44.0-2
numactl    Version  :  2.0.16-1

```
[config.log](https://github.com/openucx/ucx/files/13313223/config.log)

ParoXoN:
I ran into the very same issue with the same set of libs on Arch x86_64:
```
$ echo go glibc binutils openmpi gcc rdma-core numactl | tr ' ' '\n' | xargs -I{} sh -c 'echo {} $(pacman -Qi {}  | grep Version)' | column -t
go         Version  :  2:1.21.4-1
glibc      Version  :  2.38-7
binutils   Version  :  2.41-3
openmpi    Version  :  4.1.5-5
gcc        Version  :  13.2.1-3
rdma-core  Version  :  44.0-2
numactl    Version  :  2.0.16-1
```

[...]

PR #9817, opened 2024-04-11
BINDINGS/GO: Fixed go build failure.

rakhmets:
## What
Fixed failure:
```
request.go:32:10: cannot define new methods on non-local type *C.ucp_request_param_t
```

Checked with go1.22.2.

## Why?
https://github.com/openucx/ucx/issues/9482

### DIFF
--- a/bindings/go/src/ucx/request.go
+++ b/bindings/go/src/ucx/request.go
@@ -29,7 +29,7 @@ func (p *UcpRequestParams) SetMemType(memType UcsMemoryType) *UcpRequestParams {
 	return p
 }
 
-func (p *C.ucp_request_param_t) SetMemType(params *UcpRequestParams) {
+func setMemType(params *UcpRequestParams, p *C.ucp_request_param_t) {
 	if (params != nil) && params.memTypeSet {
 		p.op_attr_mask = C.UCP_OP_ATTR_FIELD_MEMORY_TYPE
 		p.memory_type = C.ucs_memory_type_t(params.memType)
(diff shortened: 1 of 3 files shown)


bindings/go/src/ucx/request.go
```diff
@@ -29,7 +29,7 @@ func (p *UcpRequestParams) SetMemType(memType UcsMemoryType) *UcpRequestParams {
 	return p
 }
 
-func (p *C.ucp_request_param_t) SetMemType(params *UcpRequestParams) {
+func SetMemType(params *UcpRequestParams, p *C.ucp_request_param_t) {
```
ivankochin, line 32: Do I understand correctly that this is some kind of casting function, which casts Go structure to C one?

I mean is that function for internal use only? Or is it supposed to be used by customers? If latter, then this change would break API compatibility.

rakhmets: The function sets field to a C-structure based on values of Go-structure passed as a parameter.
Originally, it was written in a way that allows to have the following code:
```go
func foo(goParams *UcpRequestParams, cParams *C.ucp_request_param_t) {
    ...
    cParams.SetMemType(goRequestParams)
    ...
}
```
However, go doesn't allow to extend an existing type (`C.ucp_request_param_t`) in another package (`ucx`).
The older go version don't raise an error, unlike the newer ones.
Yes this change would break compatibility. But the code still needs to be modified so that it can be compiled.

yosefe: does it mean we need the same fix when setting all other fields of UcpRequestParams ?

ivankochin: How do you think, does it worth it to add some compile-time logic to leave the old version of function on place for old versions of go?

yosefe: AFAIK this is an internal function , let's see how we can "hide" it

ivankochin (approved):

bindings/go/src/ucx/request.go
rakhmets: Made it a private.

yosefe (approved):
pls squash

PR #9818, opened 2024-04-12
UCP: Register staging buffers on all MDs supporting registration

brminich:
## What

Register rndv staging buffers on all MDs supporting registration (even if non-blocking registration is requested for the given memory type, but MD does not support it)

Fixes [RM3861726](https://redmine.mellanox.com/issues/3861726) (internal link)

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -1565,12 +1565,20 @@ ucp_add_component_resources(ucp_context_h context, ucp_rsc_index_t cmpt_index,
             }
 
             ucs_memory_type_for_each(mem_type) {
-                if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
-                    !(md_attr->reg_nonblock_mem_types & UCS_BIT(mem_type))) {
-                    continue;
-                }
-
                 if (md_attr->flags & UCT_MD_FLAG_REG) {
+                    if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
+                        !(md_attr->reg_nonblock_mem_types & UCS_BIT(mem_type))) {
+                        if (md_attr->reg_mem_types & UCS_BIT(mem_type)) {
+                            /* Keep map of MDs supporting blocking registration
+                             * if non-blocking registration is requested for the
+                             * given memory type. In some cases blocking
+                             * registration maybe required anyway (e.g. internal
+                             * staging buffers for rndv pipeline protocols). */
+                            context->reg_block_md_map[mem_type] |= UCS_BIT(md_index);
+                        }
+                        continue;
+                    }
+
                     if (md_attr->reg_mem_types & UCS_BIT(mem_type)) {
                         context->reg_md_map[mem_type] |= UCS_BIT(md_index);
                     }
@@ -1696,6 +1704,7 @@ static ucs_status_t ucp_fill_resources(ucp_context_h context,
 
     ucs_memory_type_for_each(mem_type) {
         context->reg_md_map[mem_type]           = 0;
+        context->reg_block_md_map[mem_type]     = 0;
         context->cache_md_map[mem_type]         = 0;
         context->dmabuf_mds[mem_type]           = UCP_NULL_RESOURCE;
         context->alloc_md[mem_type].md_index    = UCP_NULL_RESOURCE;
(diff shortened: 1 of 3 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -718,16 +719,19 @@ ucp_memh_init_from_parent(ucp_mem_h memh, ucp_md_map_t parent_md_map)
 }
 
 static ucs_status_t ucp_memh_init_uct_reg(ucp_context_h context, ucp_mem_h memh,
-                                          unsigned uct_flags,
+                                          unsigned uct_flags, ucp_md_map_t md_map,
```
yosefe, line 722: maybe we could use UCT_MD_MEM_FLAG_LOCK in uct_flags for rndv_frag memory pools, and this would also allow using MDs that support only blocking registration?

src/ucp/core/ucp_context.c
```diff
@@ -1565,12 +1565,13 @@ ucp_add_component_resources(ucp_context_h context, ucp_rsc_index_t cmpt_index,
...
             ucs_memory_type_for_each(mem_type) {
-                if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
-                    !(md_attr->reg_nonblock_mem_types & UCS_BIT(mem_type))) {
-                    continue;
-                }
-
                 if (md_attr->flags & UCT_MD_FLAG_REG) {
+                    if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
+                        !(md_attr->reg_nonblock_mem_types & UCS_BIT(mem_type))) {
+                        context->reg_block_md_map[mem_type] |= UCS_BIT(md_index);
```
yosefe, line 1571: seems we don't check md_attr->reg_mem_types here, which leads for example to registering host memory with gdr_copy

src/ucp/core/ucp_mm.c
brminich: fixed in place as the patch is small

src/ucp/core/ucp_context.c
```diff
@@ -1565,12 +1565,15 @@ ucp_add_component_resources(ucp_context_h context, ucp_rsc_index_t cmpt_index,
...
-                if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
-                    !(md_attr->reg_nonblock_mem_types & UCS_BIT(mem_type))) {
-                    continue;
-                }
-
                 if (md_attr->flags & UCT_MD_FLAG_REG) {
+                    if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
+                        !(md_attr->reg_nonblock_mem_types & UCS_BIT(mem_type))) {
+                        if (md_attr->reg_mem_types & UCS_BIT(mem_type)) {
+                            context->reg_block_md_map[mem_type] |= UCS_BIT(md_index);
```
yosefe, line 1572: pls add doc

src/ucp/core/ucp_mm.c
```diff
@@ -502,7 +502,8 @@ ucp_memh_register_internal(ucp_context_h context, ucp_mem_h memh,
     }
 
     ucs_for_each_bit(md_index, reg_md_map) {
-        ucs_assertv(context->reg_md_map[mem_type] & UCS_BIT(md_index),
+        ucs_assertv((context->reg_md_map[mem_type] |
+                     context->reg_block_md_map[mem_type]) & UCS_BIT(md_index),
```
yosefe, line 506: print reg_block_md_map in the assertv

src/ucp/core/ucp_mm.c
```diff
@@ -722,12 +723,19 @@ static ucs_status_t ucp_memh_init_uct_reg(ucp_context_h context, ucp_mem_h memh,
                                           const char *alloc_name)
 {
     ucs_memory_type_t mem_type = memh->mem_type;
-    ucp_md_map_t reg_md_map    = context->reg_md_map[mem_type] & ~memh->md_map;
-    ucp_md_map_t cache_md_map  = context->cache_md_map[mem_type] & reg_md_map;
+    ucp_md_map_t reg_md_map    = context->reg_md_map[mem_type];
```
yosefe, line 726: maybe not initialize it? it's changed during the function

yosefe (approved):

PR #9827, opened 2024-04-17
CI: Remove RoCE Valgrind testing

ivankochin:
## What
Remove Valgrind testing on RoCE

## Why ?
It's already covered by ASAN and fails due to timeout sometimes.

### DIFF
--- a/buildlib/pr/main.yml
+++ b/buildlib/pr/main.yml
@@ -450,16 +450,6 @@ stages:
         asan_check: yes
 
 
-  - stage: Valgrind
-    dependsOn: [Static_check]
-    jobs:
-    - template: tests.yml
-      parameters:
-        name: roce
-        demands: ucx_roce -equals yes
-        test_perf: 0
-        valgrind_check: yes
-
 #  - stage: Cuda_compatible
 #    dependsOn: [Static_check]
 #    jobs:


yosefe (approved):

rakhmets:
@ivankochin can we remove all valgrind related code from scripts?

ivankochin:
> @ivankochin can we remove all valgrind related code from scripts?

Yes, I think we can, but the question is should we? I can imagine the scenario when test_jenkins.sh can be used again to test with valgrind. So I will prefer to leave it on it's place, in case we would still need it later. @yosefe, what do you think about that question?

ivankochin:
> @ivankochin can we remove all valgrind related code from scripts?

Yes, I think we can, but the question is should we? I can imagine the scenario when test_jenkins.sh can be used again to test with valgrind. So I will prefer to leave it on it's place.

yosefe:
> > @ivankochin can we remove all valgrind related code from scripts?
> 
> Yes, I think we can, but the question is should we? I can imagine the scenario when test_jenkins.sh can be used again to test with valgrind. So I will prefer to leave it on it's place, in case we would still need it later. @yosefe, what do you think about that question?

Valgrind tests are still running in QA, and we want to allow other applications that are using UCX to be run with valgrind without false alarms. So we want to keep them.

PR #9829, opened 2024-04-17
UCT/BASE: Also distribute UCT v2 header

tvegas1:
## What
Make UCT v2 header available for applications to include it.


yosefe:
ucx_v2.h is not installed intentionally, which APIs in it are needed?

yosefe:
> For my understanding: why not exporting the v2 header?

In order to allow changes in the V2 API, we don't distribute it for now

PR #9833, opened 2024-04-18
AZP/TEST: Add server-side log prints

Alexey-Rivkin:
## What
Get server logs from ephemeral Docker containers.

## Why ?
Currently, there is no way to see stdout+stderr of the server container.

## How ?
Redirect server output to a file on the workspace volume.

### DIFF
--- a/buildlib/tools/test_mad.sh
+++ b/buildlib/tools/test_mad.sh
@@ -34,9 +34,10 @@ build_ucx_in_docker() {
 }
 
 docker_run_srv() {
+    local test_name="$1"
     HCA=$(detect_hca)
-    docker_stop_srv
-    docker run --rm \
+    docker run \
+        --rm \
         --detach \
         --net=host \
         --name ucx_perftest_"$BUILD_BUILDID" \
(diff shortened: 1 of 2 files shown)


dpressle (commented):
Seems like over compilated, why not simply redirect process output to a file on the workspace volume?

buildlib/tools/test_mad.sh
```diff
@@ -45,7 +46,11 @@ docker_run_srv() {
         -v /hpc/local:/hpc/local \
         --ulimit memlock=-1:-1 --device=/dev/infiniband/ \
         $IMAGE \
-        bash -c "${PWD}/install/bin/ucx_perftest -K ${HCA}"
+        bash -c "${PWD}/install/bin/ucx_perftest -K ${HCA} > ${PWD}/ucx_perf_srv.log 2>&1"
```
dpressle, line 49: since we are starting the server 2 times, we should probably generate 2 different logs for each test.

Alexey-Rivkin: Added unique names for the server log files to be on the safe side.

Alexey-Rivkin:
> Seems like over compilated, why not simply redirect process output to a file on the workspace volume?

Good idea, thanks.

buildlib/tools/test_mad.sh
```diff
@@ -45,7 +47,13 @@ docker_run_srv() {
         -v /hpc/local:/hpc/local \
         --ulimit memlock=-1:-1 --device=/dev/infiniband/ \
         $IMAGE \
-        bash -c "${PWD}/install/bin/ucx_perftest -K ${HCA}"
+        bash -c "${PWD}/install/bin/ucx_perftest -K ${HCA} > \
+            ${PWD}/ucx_perf_srv_${test_name}_${BUILD_BUILDID}.log 2>&1"
+}
+
+srv_logs() {
```
yosefe, line 54: maybe we can print the logs during server stop stage?

Alexey-Rivkin: Good idea. The only drawback is that you will need to know where to look for logs, rather than dedicated steps now.

yosefe: Add ".. and print logs" to the stage name :)

yosefe (approved):

PR #9836, opened 2024-04-19
NEWS: Add 1.16.0 and 1.17.0-rc1 sections - v1.17.x

yosefe:
## Why
Preparation for making 1.17.0 releases

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,5 @@
 #
-## Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2021. ALL RIGHTS RESERVED.
+## Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001. ALL RIGHTS RESERVED.
 ## Copyright (C) UT-Battelle, LLC. 2014-2019. ALL RIGHTS RESERVED.
 ## Copyright (C) ARM Ltd. 2017-2021.  ALL RIGHTS RESERVED.
 ##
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -1,5 +1,5 @@
 #
-## Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2021. ALL RIGHTS RESERVED.
+## Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2024. ALL RIGHTS RESERVED.
```
rakhmets, line 2:
```suggestion
## Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001. ALL RIGHTS RESERVED.
```

yosefe: Why do you suggest to revert it?

rakhmets: It should be only the year of creation according to the guidance.

rakhmets (approved):

NEWS
```diff
@@ -1,5 +1,5 @@
 #
-## Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2021. ALL RIGHTS RESERVED.
+## Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001. ALL RIGHTS RESERVED.
```
tvegas1, line 2: shouldn't we add -2024?

tvegas1: just saw previous comments

tvegas1 (approved):

NEWS
rakhmets: It should be only the year of creation according to the guidance.

PR #9838, opened 2024-04-22
GTEST/UCT: Unit tests for UMR XGVMI

iyastreb:
## What
Few unit tests for XGVMI UMR approach to check that:
- Lazy initialization works as expected
- All UMR invariants are respected
- UMR mkey pool is properly used on exporter when XGVMI is supported and when it's not
- UMR mkey hash map is properly used on importer side with and without XGVMI support
- All allocated objects are freed

We introduce new config option UCX_XGVMI_UMR_ENABLE option as a part of this commit, which enables XGVMI UMR workflow.

### DIFF
--- a/test/gtest/uct/ib/test_devx.cc
+++ b/test/gtest/uct/ib/test_devx.cc
@@ -30,11 +30,11 @@ public:
         }
     }
 
-    uct_ib_mlx5_md_t *md() {
+    uct_ib_mlx5_md_t *md() const {
         return ucs_derived_of(m_e->md(), uct_ib_mlx5_md_t);
     }
 
-    uct_priv_worker_t *worker() {
+    uct_priv_worker_t *worker() const {
         return ucs_derived_of(m_e->worker(), uct_priv_worker_t);
     }
 };
@@ -49,3 +49,234 @@ UCS_TEST_P(test_devx, dbrec)
 }
 
 UCT_INSTANTIATE_IB_TEST_CASE(test_devx);
+
+
+class test_devx_umr_mkey : public test_devx {
+public:
+    void init() {
+        test_devx::init();
+    }
+
+    bool check_xgvmi() const {
+        if (md()->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI) {
+            uct_ib_mlx5_devx_mem_t *memh = create_memh(1);
+            ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), memh));
+            /* XGVMI capability might be dropped by this point! */
+            destroy_memh(memh);
+        }
+
+        return md()->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI;
+    }
+
+    void skip_no_xgvmi() const {
+        if (!check_xgvmi()) {
+            UCS_TEST_SKIP_R("XGVMI capability is not supported");
+        }
+    }
+
+    uct_ib_mlx5_devx_mem_t *create_memh(size_t length) const {
+        void *addr                     = ucs_malloc(length, "test");
+        uct_md_mem_reg_params_t params = {};
+        uct_mem_h memh;
+
+        EXPECT_NE(nullptr, addr);
+        ASSERT_UCS_OK(uct_ib_mlx5_devx_mem_reg(m_e->md(), addr, length, &params,
+                                               &memh));
+        return ucs_derived_of(memh, uct_ib_mlx5_devx_mem_t);
+    }
+
+    uct_ib_mlx5_devx_mem_t *
+    import_memh(const uct_ib_mlx5_devx_mem_t *exported_memh) const {
+        uct_ib_md_packed_mkey_t packed_mkey = {
+            .lkey    = exported_memh->exported_lkey,
+            .vhca_id = md()->super.vhca_id
+        };
+
+        uct_md_mem_attach_params_t params = {};
+        uct_ib_mlx5_devx_mem_t *imported_memh;
+        ASSERT_UCS_OK(uct_ib_mlx5_devx_mem_attach(m_e->md(), &packed_mkey,
+                                                  &params,
+                                                  (uct_mem_h*)&imported_memh));
+        return imported_memh;
+    }
+
+    void destroy_memh(uct_ib_mlx5_devx_mem_t *memh) const {
+        void *addr = memh->address;
+        uct_md_mem_dereg_params_t params;
+
+        params.field_mask = UCT_MD_MEM_DEREG_FIELD_MEMH;
+        params.memh       = (uct_mem_h)memh;
+
+        ASSERT_UCS_OK(uct_ib_mlx5_devx_mem_dereg(m_e->md(), &params));
+        ucs_free(addr);
+    }
+
+    void check_umr_init(bool init_export, bool init_import) const {
+        EXPECT_EQ(init_export, nullptr != md()->umr.qp);
+        EXPECT_EQ(init_export, nullptr != md()->umr.cq);
+        if (!init_export) {
+            EXPECT_TRUE(ucs_list_is_empty(&md()->umr.mkey_pool));
+        }
+
+        EXPECT_EQ(init_import, nullptr != md()->umr.mkey_hash);
+    }
+
+    static void check_memh_export(uct_ib_mlx5_devx_mem_t *memh, bool is_umr) {
+        EXPECT_EQ(is_umr,
+                  UCT_IB_MLX5_MKEY_TAG_UMR == (memh->exported_lkey & 0xff));
+        EXPECT_EQ(is_umr, nullptr != memh->exported_umr_mkey);
+        EXPECT_EQ(is_umr, nullptr == memh->cross_mr);
+    }
+
+    void perf_export_import_mkey(size_t memh_count, size_t memh_size,
+                                 ucs_time_t &export_time,
+                                 ucs_time_t &import_time) {
+        std::vector<uct_ib_mlx5_devx_mem_t*> exported_memh, imported_memh;
+        exported_memh.reserve(memh_count);
+        imported_memh.reserve(memh_count);
+
+        for (size_t i = 0; i < memh_count; ++i) {
+            exported_memh.push_back(create_memh(memh_size));
+        }
+
+        ucs_time_t start_time = ucs_get_time();
+        for (auto it : exported_memh) {
+            ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), it));
+            check_memh_export(it, true);
+        }
+        export_time = ucs_get_time() - start_time;
+
+        start_time = ucs_get_time();
+        for (auto it : exported_memh) {
+            imported_memh.push_back(import_memh(it));
+        }
+        import_time = ucs_get_time() - start_time;
+
+        for (size_t i = 0; i < memh_count; ++i) {
+            destroy_memh(imported_memh[i]);
+            destroy_memh(exported_memh[i]);
+        }
+    }
+};
+
+UCS_TEST_P(test_devx_umr_mkey, lazy_init)
+{
+    check_umr_init(false, false);
+}
+
+UCS_TEST_P(test_devx_umr_mkey, export_mkey_no_xgvmi, "IB_XGVMI_UMR_ENABLE=n")
+{
+    skip_no_xgvmi();
+
+    uct_ib_mlx5_devx_mem_t *memh = create_memh(1024);
+
+    ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), memh));
+
+    /* UMR objects remains uninitialized, exported key is not UMR one */
+    check_umr_init(false, false);
+    check_memh_export(memh, false);
+    destroy_memh(memh);
+}
+
+UCS_TEST_P(test_devx_umr_mkey, export_mkey_xgvmi)
+{
+    skip_no_xgvmi();
+
+    uct_ib_mlx5_devx_mem_t *memh = create_memh(1024);
+    ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), memh));
+    /* XGVMI capability might be dropped by this point! */
+
+    /* UMR export transport objects must be initialized */
+    check_umr_init(true, false);
+    /* UMR mkey pool must be empty */
+    EXPECT_TRUE(ucs_list_is_empty(&md()->umr.mkey_pool));
+    /* Whether mkey is UMR depends on XGVMI capability */
+    check_memh_export(memh, true);
+
+    destroy_memh(memh);
+
+    /* After release of memh, associated UMR mkey is moved back to mkey pool */
+    EXPECT_EQ(1, ucs_list_length(&md()->umr.mkey_pool));
+}
+
+UCS_TEST_P(test_devx_umr_mkey, import_mkey_no_umr, "IB_XGVMI_UMR_ENABLE=n")
+{
+    skip_no_xgvmi();
+
+    /* Export any mkey */
+    uct_ib_mlx5_devx_mem_t *exported_memh = create_memh(1024);
+    ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), exported_memh));
+
+    /* Import as non-UMR mkey */
+    uct_ib_mlx5_devx_mem_t *imported_memh = import_memh(exported_memh);
+    EXPECT_NE(nullptr, imported_memh);
+
+    /* UMR hash map remains uninitialized */
+    EXPECT_EQ(nullptr, md()->umr.mkey_hash);
+
+    destroy_memh(imported_memh);
+    destroy_memh(exported_memh);
+}
+
+UCS_TEST_P(test_devx_umr_mkey, import_mkey_umr)
+{
+    skip_no_xgvmi();
+
+    /* Export UMR mkey */
+    uct_ib_mlx5_devx_mem_t *exported_memh = create_memh(1024);
+    ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), exported_memh));
+    check_memh_export(exported_memh, true);
+
+    /* Import first UMR mkey */
+    uct_ib_mlx5_devx_mem_t *imported_memh = import_memh(exported_memh);
+
+    check_umr_init(true, true);
+    EXPECT_EQ(1, kh_size(md()->umr.mkey_hash));
+
+    /* Second import with the same packed mkey, should reuse alias */
+    uct_ib_mlx5_devx_mem_t *imported_memh2 = import_memh(exported_memh);
+    EXPECT_NE(imported_memh, imported_memh2);
+    EXPECT_EQ(imported_memh->super.lkey, imported_memh2->super.lkey);
+    EXPECT_EQ(1, kh_size(md()->umr.mkey_hash));
+
+    destroy_memh(imported_memh);
+    destroy_memh(imported_memh2);
+    destroy_memh(exported_memh);
+}
+
+UCS_TEST_P(test_devx_umr_mkey, perf_export_import_mkey_xgvmi)
+{
+    skip_no_xgvmi();
+
+    const size_t MEMH_COUNT = RUNNING_ON_VALGRIND ? 100 : 1000;
+    ucs_time_t export_create_time, import_create_time;
+    perf_export_import_mkey(MEMH_COUNT, 1024, export_create_time,
+                            import_create_time);
+
+    /* After release all allocated UMR keys go back to the mkey pool */
+    EXPECT_EQ(MEMH_COUNT, ucs_list_length(&md()->umr.mkey_pool));
+
+    /* Now reuse the keys for different handles of different size */
+    ucs_time_t export_reuse_time, import_reuse_time;
+    perf_export_import_mkey(MEMH_COUNT, 2048, export_reuse_time,
+                            import_reuse_time);
+
+    /* Reusing existing UMR mkeys is normally 200x faster than creating new
+     * ones, given jitter we check that it's at least 10 times faster */
+    EXPECT_LT(export_reuse_time, export_create_time / 10);
+    EXPECT_LT(export_reuse_time, export_create_time / 10);
+
+    char buf[1024] = {};
+    snprintf(buf, sizeof(buf), "Create %zu UMR mkeys, export: %.3f ms, import: "
+             "%.3f ms", MEMH_COUNT, ucs_time_to_msec(export_create_time),
+             ucs_time_to_msec(import_create_time));
+    UCS_TEST_MESSAGE << buf;
+    snprintf(buf, sizeof(buf), "Reuse %zu UMR mkeys, export: %.3f ms, import: "
+             "%.3f ms", MEMH_COUNT, ucs_time_to_msec(export_reuse_time),
+            ucs_time_to_msec(import_reuse_time));
+    UCS_TEST_MESSAGE << buf;
+
+    EXPECT_EQ(MEMH_COUNT, ucs_list_length(&md()->umr.mkey_pool));
+}
+
+UCT_INSTANTIATE_IB_TEST_CASE(test_devx_umr_mkey);
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -166,6 +166,10 @@ ucs_config_field_t uct_ib_md_config_table[] = {
      "Number of indexes in a symmetric block. More can lead to less contention.",
      ucs_offsetof(uct_ib_md_config_t, ext.smkey_block_size), UCS_CONFIG_TYPE_UINT},
 
+    {"XGVMI_UMR_ENABLE", "y",
+     "Enable UMR optimization for XGVMI mkeys export/import.",
+     ucs_offsetof(uct_ib_md_config_t, xgvmi_umr_enable), UCS_CONFIG_TYPE_BOOL},
+
     {NULL}
 };
 
--- a/src/uct/ib/base/ib_md.h
+++ b/src/uct/ib/base/ib_md.h
@@ -188,6 +188,7 @@ typedef struct uct_ib_md_config {
     unsigned                 devx_objs;    /**< Objects to be created by DevX */
     ucs_ternary_auto_value_t mr_relaxed_order; /**< Allow reorder memory accesses */
     int                      enable_gpudirect_rdma; /**< Enable GPUDirect RDMA */
+    int                      xgvmi_umr_enable; /**< Enable UMR workflow for XGVMI */
 } uct_ib_md_config_t;
 
 /**
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -2303,6 +2303,11 @@ ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
     md->super.name       = UCT_IB_MD_NAME(mlx5);
     md->super.vhca_id    = vhca_id;
 
+    if ((md->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI) &&
+        (md_config->xgvmi_umr_enable == UCS_YES)) {
+        md->flags |= UCT_IB_MLX5_MD_FLAG_XGVMI_UMR;
+    }
+
     /* Zero init UMR related fields, will lazy init on first use */
     md->umr.cq        = NULL;
     md->umr.qp        = NULL;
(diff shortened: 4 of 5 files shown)


test/gtest/uct/ib/test_devx.cc
```diff
@@ -48,4 +48,202 @@ UCS_TEST_P(test_devx, dbrec)
...
 }
 
+class test_devx_umr_mkey : public test_devx {
+public:
+    void init() {
+        test_devx::init();
+    }
+
+    uct_ib_mlx5_devx_mem_t *create_memh(size_t length) const {
+        void *addr                     = ucs_malloc(length, "test");
```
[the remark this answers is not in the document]
iyastreb: sure

test/gtest/uct/ib/test_devx.cc
```diff
@@ -48,4 +48,202 @@ UCS_TEST_P(test_devx, dbrec)
...
+    void destroy_memh(uct_ib_mlx5_devx_mem_t *memh) const {
+        void *addr                       = memh->address;
+        uct_md_mem_dereg_params_t params = {};
+
+        params.field_mask = UCT_MD_MEM_DEREG_FIELD_MEMH;
+        params.memh       = (uct_mem_h)memh;
+
+        ASSERT_UCS_OK(uct_ib_mlx5_devx_mem_dereg(m_e->md(), &params));
+
+        if (nullptr != addr) {
```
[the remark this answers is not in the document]
iyastreb: no, for imported memory handles it's NULL

test/gtest/uct/ib/test_devx.cc
```diff
@@ -48,4 +48,202 @@ UCS_TEST_P(test_devx, dbrec)
...
+    check_memh_export(memh, true);
+    destroy_memh(memh);
+
+    /* After release of memh, associated UMR mkey is moved back to mkey pool */
+    ASSERT_EQ(1, ucs_list_length(&md()->umr.mkey_pool));
+}
+
+UCS_TEST_P(test_devx_umr_mkey, export_mkey_xgvmi_reuse)
+{
+    if (!(md()->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI)) {
```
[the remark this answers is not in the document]
iyastreb: Yes, indeed that is a good option!
I tried that initially, but if failed with exception, because `m_e` was created only in `init`, but this check-for-skip-test is called after object constructor but before `init`. So I moved `m_e` object creation to the ctor and it works like that

test/gtest/uct/ib/test_devx.cc
```diff
@@ -48,4 +48,202 @@ UCS_TEST_P(test_devx, dbrec)
...
+
+UCS_TEST_P(test_devx_umr_mkey, import_mkey_no_umr)
+{
+    /* Export any mkey */
+    uct_ib_mlx5_devx_mem_t *exported_memh = create_memh(1024);
+    ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), exported_memh));
+
+    /* Construct packed mkey without UMR tag */
+    uct_ib_md_packed_mkey_t packed_mkey = {
+        .lkey = exported_memh->exported_lkey & ~UCT_IB_MLX5_MKEY_TAG_UMR,
```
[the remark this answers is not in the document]
iyastreb: I tested it manually, but I was advised by @brminich to not commit such a large buffer allocation (2GB), because it may slow down / break the tests.

test/gtest/uct/ib/test_devx.cc
```diff
@@ -48,4 +48,202 @@ UCS_TEST_P(test_devx, dbrec)
...
+    ASSERT_EQ(MEMH_COUNT - REUSE_COUNT, ucs_list_length(&md()->umr.mkey_pool));
+
+    for (size_t i = 0; i < REUSE_COUNT; ++i) {
+        destroy_memh(memh[i]);
+    }
+
+    ASSERT_EQ(MEMH_COUNT, ucs_list_length(&md()->umr.mkey_pool));
+}
+
+UCS_TEST_P(test_devx_umr_mkey, import_mkey_no_umr)
```
tvegas1, line 184: seems that test is failing on `ALLOW_OTHER_VHCA_ACCESS` on non-supported xgvmi FW, called from `uct_ib_mlx5_devx_xgvmi_umem_mr()`.

We might need to fix it in separate PR with either enabling `silent` or removing `uct_ib_mlx5_devx_allow_xgvmi_access()` when non supported.

test/gtest/uct/ib/test_devx.cc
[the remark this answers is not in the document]
iyastreb: Yes, this is exactly what I've implemented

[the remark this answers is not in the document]
tvegas1: i see, i misunderstood you last comment.

test/gtest/uct/ib/test_devx.cc
```diff
@@ -48,4 +48,202 @@ UCS_TEST_P(test_devx, dbrec)
...
+    ASSERT_NE(imported_memh, imported_memh2);
+    ASSERT_EQ(imported_memh->super.lkey, imported_memh2->super.lkey);
+    ASSERT_EQ(1, kh_size(md()->umr.mkey_hash));
+
+    destroy_memh(imported_memh);
+    destroy_memh(imported_memh2);
+    destroy_memh(exported_memh);
+}
+
 UCT_INSTANTIATE_IB_TEST_CASE(test_devx);
```
brminich, line 263: pls move it before the new class, typically we initialize test classes right after their (and their tests) definitions

iyastreb: sure

test/gtest/uct/ib/test_devx.cc
```diff
@@ -48,17 +50,42 @@ UCS_TEST_P(test_devx, dbrec)
...
+    }
+
+    /*
+     * XGVMI capability advertised by device might be dropped on the initial
+     * operation. This function performs dummy XGVMI operation to detect
+     * whether XGVMI is actually operational
+     */
+    bool check_xgvmi_effective() const {
+        if (check_xgvmi()) {
+            uct_ib_mlx5_devx_mem_t *memh = create_memh(1);
```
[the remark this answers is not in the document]
iyastreb: This change would have some side-effects:
- `check_xgvmi_effective` calls `reg_export` which internally initializes UMR export objects, and populates mkey pool with one element. So after this call we would not be able to check the pristine state (like we do in `lazy_init` test)
- with this change we would not be able to test cases where XGVMI capability is advertised, but not actually supported. Currently I have one test that covers this use case: `export_mkey_xgvmi`. I think it's desirable to have a possibility to test it

So I would keep it as is, calling this function only where necessary

brminich (dismissed):

tvegas1 (approved):

test/gtest/uct/ib/test_devx.cc
```diff
@@ -16,12 +16,14 @@ class test_devx : public uct_test {
 public:
     entity* m_e;
 
+    test_devx() : m_e(create_entity(0))
+    {
+        m_entities.push_back(m_e);
```
yosefe, line 21: IMO it better be during init() and not in actor

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+    bool check_xgvmi() const {
+        return md()->flags & UCT_IB_MLX5_MD_FLAG_INDIRECT_XGVMI;
+    }
+
+    /*
+     * XGVMI capability advertised by device might be dropped on the initial
+     * operation. This function performs dummy XGVMI operation to detect
+     * whether XGVMI is actually operational
+     */
+    bool check_xgvmi_effective() const {
```
yosefe, line 71: can we unite check_xgvmi_effective and check_xgvmi to a single function?

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+    uct_ib_mlx5_devx_mem_t *memh = create_memh(1024);
+    ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), memh));
+    /* XGVMI capability might be dropped by this point! */
+
+    /* UMR export transport objects must be initialized */
+    check_umr_init(true, false);
+    /* UMR mkey pool must be empty */
+    ASSERT_TRUE(ucs_list_is_empty(&md()->umr.mkey_pool));
+    /* Whether mkey is UMR depends on XGVMI capability */
+    check_memh_export(memh, check_xgvmi());
```
yosefe, line 157: function is skipped if check_xgvmi is false. so why. check it here again? (multiple locations)

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+    void destroy_memh(uct_ib_mlx5_devx_mem_t *memh) const {
+        void *addr                       = memh->address;
+        uct_md_mem_dereg_params_t params = {};
+
+        params.field_mask = UCT_MD_MEM_DEREG_FIELD_MEMH;
+        params.memh       = (uct_mem_h)memh;
+
+        ASSERT_UCS_OK(uct_ib_mlx5_devx_mem_dereg(m_e->md(), &params));
+
+        if (nullptr != addr) {
```
yosefe, line 103: ucs_free ignores NULL anyway, no need to check

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+
+        EXPECT_NE(nullptr, addr);
+        ASSERT_UCS_OK(uct_ib_mlx5_devx_mem_reg(m_e->md(), addr, length, &params,
+                                               &memh));
+        return ucs_derived_of(memh, uct_ib_mlx5_devx_mem_t);
+    }
+
+    void destroy_memh(uct_ib_mlx5_devx_mem_t *memh) const {
+        void *addr                       = memh->address;
+        uct_md_mem_dereg_params_t params = {};
```
yosefe, line 96: why need to init?

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+
+        ASSERT_UCS_OK(uct_ib_mlx5_devx_mem_dereg(m_e->md(), &params));
+
+        if (nullptr != addr) {
+            ucs_free(addr);
+        }
+    }
+
+    void check_umr_init(bool init_export, bool init_import) const {
+        EXPECT_EQ(init_export, nullptr != md()->umr.qp);
```
yosefe, line 109: if (init_export) {
  EXPECT_NE(nullptr, umr.qp);
  EXPECT_NE(nullptr, umr.cq);
} else {
   EXPECT_TRUE(ucs_list_is_empty(&md()->umr.mkey_pool));
}

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+        EXPECT_EQ(init_export, nullptr != md()->umr.cq);
+        if (!init_export) {
+            EXPECT_TRUE(ucs_list_is_empty(&md()->umr.mkey_pool));
+        }
+
+        EXPECT_EQ(init_import, nullptr != md()->umr.mkey_hash);
```
yosefe, line 115: seems redundant, we initialize the hash on demand anyway, and it's implementation detail

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+    }
+
+    static void check_memh_export(uct_ib_mlx5_devx_mem_t *memh, bool is_umr) {
+        EXPECT_EQ(is_umr, UCT_IB_MLX5_MKEY_TAG_UMR ==
+                          (memh->exported_lkey & 0xff));
+        EXPECT_EQ(is_umr, nullptr != memh->exported_umr_mkey);
+        EXPECT_EQ(is_umr, nullptr == memh->cross_mr);
+    }
```
yosefe, line 123: I think this check is too low-level, we should check import/export works under certain conditions, no need to check implementation details such as tag or struct fields
too low-level checks will make it harder to refactor the code while preserving the UCT level functionality

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+UCS_TEST_SKIP_COND_P(test_devx_umr_mkey, export_mkey_xgvmi, !check_xgvmi())
+{
+    uct_ib_mlx5_devx_mem_t *memh = create_memh(1024);
+    ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), memh));
+    /* XGVMI capability might be dropped by this point! */
+
+    /* UMR export transport objects must be initialized */
+    check_umr_init(true, false);
+    /* UMR mkey pool must be empty */
+    ASSERT_TRUE(ucs_list_is_empty(&md()->umr.mkey_pool));
```
yosefe, line 155: why is it important the mkey pool is empty?

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+    /* Whether mkey is UMR depends on XGVMI capability */
+    check_memh_export(memh, check_xgvmi());
+
+    destroy_memh(memh);
+
+    /* After release of memh, associated UMR mkey is moved back to mkey pool */
+    ASSERT_EQ(check_xgvmi() ? 1 : 0, ucs_list_length(&md()->umr.mkey_pool));
```
yosefe, line 162: why is it needed to check?

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+}
+
+UCS_TEST_SKIP_COND_P(test_devx_umr_mkey, export_mkey_xgvmi_reuse,
+                     !check_xgvmi_effective())
+{
+    const size_t MEMH_COUNT = 10;
+    uct_ib_mlx5_devx_mem_t *memh[MEMH_COUNT];
```
yosefe, line 169: maybe std::vector?

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+    }
+
+    /* After release all allocated UMR keys go back to the mkey pool */
+    ASSERT_EQ(MEMH_COUNT, ucs_list_length(&md()->umr.mkey_pool));
+
+    /* Now reuse the keys for different handles of different size */
+    const size_t REUSE_COUNT = 5;
+    for (size_t i = 0; i < REUSE_COUNT; ++i) {
+        memh[i] = create_memh(2048);
+        ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), memh[i]));
```
yosefe, line 191: can we check performance or somehow check there is no FW flow when reusing an existing key?
checking mkey_pool list length is too low level and indirect

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +51,216 @@ UCS_TEST_P(test_devx, dbrec)
...
+                     !check_xgvmi_effective())
+{
+    /* Export any mkey */
+    uct_ib_mlx5_devx_mem_t *exported_memh = create_memh(1024);
+    ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), exported_memh));
+
+    /* Construct packed mkey without UMR tag */
+    uct_ib_md_packed_mkey_t packed_mkey = {
+        .lkey = exported_memh->exported_lkey & ~UCT_IB_MLX5_MKEY_TAG_UMR,
+        .vhca_id = md()->super.vhca_id
```
yosefe, line 214: can we disable UMR export in another way? maybe by configuration?
it's too low-level to use mkey tags in the test

yosefe (commented):
as general comment, make the tests more high level to capture more flows in less tests

test/gtest/uct/ib/test_devx.cc
iyastreb: Good question!
Regarding checking FW flow - that's a classical case for mocking, we could easily do this if we used mocking library/framework (that we discussed some time ago). Then we would write something like:
`mock().expectNcalls(0, "uct_ib_mlx5_devx_obj_create");`
But without mocking framework I've no idea how we can check that.

wrt check performance - is it fast and reliable way to guarantee that FW code is not triggered? I don't think so..
So IMO the only fast and reliable way to check this requirement is to check the pool size

test/gtest/uct/ib/test_devx.cc
iyastreb: These are the options that I see:
1. Add configuration e.g. `UCX_UMR_ENABLE` like you said. IMO this is the least desirable option, because there is no use case in production workflow for this option (right?), so we add it in production code purely for unit testing purposes, which seems wrong.
2. Using mocking library. We would be able to mock internal function call like:
`mock().expectOneCall("uct_ib_mlx5_is_xgvmi_supported").andReturnValue(0);`
3. Create MR of size > 2GB, in this case it will be handled by KSM workflow. Mikhail advised me not to do that in unit test, because it can be an expensive call.
4. Remove UMR tag from the mkey - like is done today. IMO it's ok to do low-level things in unit tests - they are low-level tests by definition.

Do you have a preference?

test/gtest/uct/ib/test_devx.cc
iyastreb: There are several comments mentioning low-level, so let me respond here.
Of course it's low level by definition, these are unit tests, not integration ones.
First I agree with you that having excessive amount of unit tests, that check implementation, not behaviour, can be hard to maintain. And ROI of those tests can be negative. I believe it's not the case here.

> we should check import/export works under certain conditions

We already have high-level CI tests for XGVMI that check that, right? There are however many things that we cannot check easily with integration tests. The fact that "it works" does not necessarily mean that it works correctly. I give you a few real examples when integration test "works":
- How do we make sure that UMR approach is used, not KSM? You cannot reliably verify that at high-level
- How to make sure that mkeys are actually reused and stored in the pool? You don't see this at high level
- Same for importer side - check that we don't create new alias each time, but reuse one from hash map
- Check that we don't allocate objects unless we need to
- Finally it takes much longer to run integration tests, so we cannot test all the possible corner cases

The aim is to complement integration tests with the reasonable amount of low-level tests. with the intention to:
1. Code feature requirements and assumptions, so that they exist not just in the design documents or developers' heads.
2. Execute different workflows to check all the internal invariants and assertions
3. Provide the minimum set of tests that cover all flows and requirements, and are executed instantly.
4. Provide a safety net to actually make a refactoring **easier**. It's super hard to change the code that is not covered by tests sufficiently. IMO it's better to break and fix a unit test than code in production

test/gtest/uct/ib/test_devx.cc
yosefe: why do we care about testing "fake-gvmi" devices? IMO we should check if xgvmi is "really" supported (or not)

test/gtest/uct/ib/test_devx.cc
yosefe: i'd suggest that check_xgvmi will trigger some flow that checks if xgvmi is really supported. or remove it altogether and add a skip during the test itself if xgvmi is dropped

test/gtest/uct/ib/test_devx.cc
yosefe: i think it's not important to check if umr objects are initialized, but just try to use them (it should fail if they are not initialized)

test/gtest/uct/ib/test_devx.cc
yosefe: > * Validate import of non-UMR mkey - to make sure that UMR workflow is NOT used, and UMR alias hash map remains unallocated

Wouldn't the import fail anyway if UMR workflow is used for non-UMR mkey?

test/gtest/uct/ib/test_devx.cc
yosefe: Maybe we can check that UMR is used and mpool keys are reused by time measurement or counting FW commands. We expect that we could import mkeys fast and without FW command.

test/gtest/uct/ib/test_devx.cc
yosefe: > 1. Add configuration e.g. `UCX_UMR_ENABLE` like you said. IMO this is the least desirable option, because there is no use case in production workflow for this option (right?), so we add it in production code purely for unit testing purposes, which seems wrong.

Actually it can be useful to add option to enable/disable this feature in case we find SW/HE issues with this flow. Since it can be disabled without hurting functionality.

test/gtest/uct/ib/test_devx.cc
iyastreb: Because those "fake-xgvmi" devices are real devices that we support. For instance, when I run my test on rock, I see that both mlx5_0 and mlx5_2 (MT4129 == ConnectX7?) are examples of those devices. I've no idea why they advertise XGVMI capability. But since we **can** run export mkey on those devices, I prefer to guarantee that it works as expected (= not using UMR approach).

test/gtest/uct/ib/test_devx.cc
iyastreb: No, it does not fail at all. It can successfully create an alias for non-UMR keys and store it inside the hash map

mem_attach supports both UMR mkeys and KSM/UMEM generated mkeys. That's why we distinguish between them by checking the mkey tag. So after importing a key I want to make sure that UMR approach was not triggered for non-UMR key

test/gtest/uct/ib/test_devx.cc
iyastreb: Time measurement is not reliable method I think. Also it's error prone, because overall performance depends on too many different calls.
First, I don't even know how we can count these FW commands?
Second, counting FW commands - it may be even less reliable in this case, because we want to differentiate between UMR and non=UMR workflow. But both workflows call FW commands (for creating mkey and then allow XGVMI).

test/gtest/uct/ib/test_devx.cc
iyastreb: ok, adding an option then

test/gtest/uct/ib/test_devx.cc
yosefe: why is it a problem that  UMR approach is triggered for non-UMR key?

test/gtest/uct/ib/test_devx.cc
yosefe: UMR should not do FW flow in most cases, this is the goal of using it.. so let's test it

test/gtest/uct/ib/test_devx.cc
yosefe: would ASAN catch it if we had a memory leak?

test/gtest/uct/ib/test_devx.cc
iyastreb: This would be a real problem (and for some time during development it was really the case)
With UMR approach we store all the received aliases in the hash map, and we destroy them only at the end of process lifetime. So if we do this for non-UMR key - we would store alias to non-UMR key forever. This is not a classical memory leak, because we destroy those objects at the end, but this is abusive memory usage for no reason.

test/gtest/uct/ib/test_devx.cc
iyastreb: True, UMR aims to minimize the number of FW calls. We could check that FW is not called in `export_mkey_xgvmi_reuse` test, it would be a nice test. The only problem is that I don't know how to count those FW calls... Do you have a clue?

We could use mocking framework for that, but this is a BIG change

test/gtest/uct/ib/test_devx.cc
iyastreb: Well, if we had a true memory leak, it must be catch by LSAN.
Here however I don't check for a memory leak. All the objects stored in the mkey pool or alias hash map will be destroyed at the end of the process, so no memory leak will be detected in this particular case. I just validate the invariant: pool must not grow up on export.
Again, I don't consider this check as super important, so we can remove it if you dislike it

test/gtest/uct/ib/test_devx.cc
iyastreb: Added perf test, as we agreed, measuring CPU cycles of export & import

test/gtest/uct/ib/test_devx.cc
iyastreb: ok, done

test/gtest/uct/ib/test_devx.cc
iyastreb: well, the most important use case here is to check that UMR objects are NOT initialized (when they should not be)

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +49,235 @@ UCS_TEST_P(test_devx, dbrec)
...
+    ASSERT_EQ(MEMH_COUNT, ucs_list_length(&md()->umr.mkey_pool));
+
+    /* Now reuse the keys for different handles of different size */
+    ucs_time_t export_reuse_time, import_reuse_time;
+    perf_export_import_mkey(MEMH_COUNT, 2048, export_reuse_time,
+                            import_reuse_time);
+
+    /* Reusing existing UMR mkeys is normally 200x faster than creating new
+     * ones, given jitter we check that it's at least 50 times faster */
+    ASSERT_GT(export_create_time / 50, export_reuse_time);
```
yosefe, line 273: 1. IMO we can reduce it to 10 times, or check vs. some known time content
2. inverse order since export_reuse_time is what we test:
EXPECT_LT(export_reuse_time, export_create_time / 10)

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +49,235 @@ UCS_TEST_P(test_devx, dbrec)
...
+    ASSERT_GT(import_create_time / 50, import_reuse_time);
+    UCS_TEST_MESSAGE << "Export create time: " << export_create_time;
```
yosefe, line 275: IMO we can print it in a more compact way + convert to milliseconds

iyastreb: Ok, now it's in this format:
```
[     INFO ] Create 1000 UMR mkeys, export: 397.802 ms, import: 215.757 ms       
[     INFO ] Reuse 1000 UMR mkeys, export: 1.784 ms, import: 0.421 ms                                            
```

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +49,235 @@ UCS_TEST_P(test_devx, dbrec)
...
+
+UCS_TEST_P(test_devx_umr_mkey, import_mkey_no_umr, "IB_XGVMI_UMR_ENABLE=n")
+{
+    if (!check_xgvmi()) {
+        UCS_TEST_SKIP_R("XGVMI capability is not supported");
+    }
+
+    /* Export any mkey */
+    uct_ib_mlx5_devx_mem_t *exported_memh = create_memh(1024);
+    ASSERT_UCS_OK(uct_ib_mlx5_devx_reg_exported_key(md(), exported_memh));
```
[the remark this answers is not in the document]
iyastreb: I think it's redundant, because export workflow we check already in export_mkey_no_xgvmi test

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +49,234 @@ UCS_TEST_P(test_devx, dbrec)
...
+    perf_export_import_mkey(MEMH_COUNT, 2048, export_reuse_time,
+                            import_reuse_time);
+
+    /* Reusing existing UMR mkeys is normally 200x faster than creating new
+     * ones, given jitter we check that it's at least 10 times faster */
+    EXPECT_LT(export_reuse_time, export_create_time / 10);
+    EXPECT_LT(export_reuse_time, export_create_time / 10);
+
+    char buf[1024] = {};
+    sprintf(buf, "Create %zu UMR mkeys, export: %.3f ms, import: %.3f ms",
```
yosefe, line 270: snprintf / ucs_snprintf_safe

test/gtest/uct/ib/test_devx.cc
```diff
@@ -49,3 +49,234 @@ UCS_TEST_P(test_devx, dbrec)
...
+            MEMH_COUNT, ucs_time_to_msec(export_create_time),
+            ucs_time_to_msec(import_create_time));
+    UCS_TEST_MESSAGE << buf;
+    sprintf(buf, "Reuse %zu UMR mkeys, export: %.3f ms, import: %.3f ms",
```
yosefe, line 274: snprintf / ucs_snprintf_safe

test/gtest/uct/ib/test_devx.cc
iyastreb: sure

test/gtest/uct/ib/test_devx.cc
iyastreb: done

yosefe (approved):

PR #9840, opened 2024-04-23
UCT/GTEST: Fixed multiple tests for gdr_copy transport

iyastreb:
## What
This is the fix for [RM#3873368](https://redmine.mellanox.com/issues/3873368).
The issue is always reproducible on rock, when building and running UCX with the following modules:
```
module load hpcx-env
module load hpcx-env/cuda
module load hpcx-env/gdrcopy
```
With `gdr_copy` MD configured, memory registration fails in several test suites:
- 3 tests in test_md
- 2 tests in uct_test
- test_uct_loopback_cuda

The root cause was always the same: CUDA memory of arbitrary size was allocated, and then this memory is registered with `uct_md_mem_reg` without any alignment. However this does not work for gdr_copy transport, because it's required to register memory aligned by GPU_PAGE_SIZE (64k in this case).

I fixed the memory registration in all those places the same way it's done in ucp_mm module: by alignment using `ucs_align_ptr_range` API

### DIFF
--- a/src/tools/perf/cuda/cuda_alloc.c
+++ b/src/tools/perf/cuda/cuda_alloc.c
@@ -12,6 +12,8 @@
 
 #include <cuda_runtime.h>
 #include <ucs/sys/compiler.h>
+#include <ucs/sys/ptr_arith.h>
+#include <uct/api/v2/uct_v2.h>
 
 
 static ucs_status_t ucx_perf_cuda_init(ucx_perf_context_t *perf)
@@ -69,15 +71,27 @@ uct_perf_cuda_alloc_reg_mem(const ucx_perf_context_t *perf,
                             unsigned flags,
                             uct_allocated_memory_t *alloc_mem)
 {
+    uct_md_attr_v2_t md_attr = {.field_mask = UCT_MD_ATTR_FIELD_REG_ALIGNMENT};
+    void *reg_address;
     ucs_status_t status;
+
+    status = uct_md_query_v2(perf->uct.md, &md_attr);
+    if (status != UCS_OK) {
+        ucs_error("uct_md_query_v2() returned %d", status);
+        return status;
+    }
 
     status = ucx_perf_cuda_alloc(length, mem_type, &alloc_mem->address);
     if (status != UCS_OK) {
         return status;
     }
 
-    status = uct_md_mem_reg(perf->uct.md, alloc_mem->address,
-                            length, flags, &alloc_mem->memh);
+    /* Register memory respecting MD reg_alignment */
+    reg_address = alloc_mem->address;
+    ucs_align_ptr_range(&reg_address, &length, md_attr.reg_alignment);
+
+    status = uct_md_mem_reg(perf->uct.md, reg_address, length, flags,
+                            &alloc_mem->memh);
     if (status != UCS_OK) {
         cudaFree(alloc_mem->address);
         ucs_error("failed to register memory");
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.c
@@ -100,25 +100,21 @@ uct_cuda_copy_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
 {
     uct_cuda_copy_md_t *md = ucs_derived_of(uct_md, uct_cuda_copy_md_t);
 
-    md_attr->flags                  = UCT_MD_FLAG_REG | UCT_MD_FLAG_ALLOC;
-    md_attr->reg_mem_types          = UCS_BIT(UCS_MEMORY_TYPE_HOST) |
-                                      UCS_BIT(UCS_MEMORY_TYPE_CUDA) |
-                                      UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
-    md_attr->reg_nonblock_mem_types = 0;
-    md_attr->cache_mem_types        = UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
-    md_attr->alloc_mem_types        = UCS_BIT(UCS_MEMORY_TYPE_CUDA) |
-                                      UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
-    md_attr->access_mem_types       = UCS_BIT(UCS_MEMORY_TYPE_CUDA) |
-                                      UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
-    md_attr->detect_mem_types       = UCS_BIT(UCS_MEMORY_TYPE_CUDA) |
-                                      UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
-    md_attr->dmabuf_mem_types       = md->config.dmabuf_supported ?
-                                      UCS_BIT(UCS_MEMORY_TYPE_CUDA) : 0;
+    uct_md_base_md_query(md_attr);
+    md_attr->flags            = UCT_MD_FLAG_REG | UCT_MD_FLAG_ALLOC;
+    md_attr->reg_mem_types    = UCS_BIT(UCS_MEMORY_TYPE_HOST) |
+                                UCS_BIT(UCS_MEMORY_TYPE_CUDA) |
+                                UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
+    md_attr->cache_mem_types  = UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
+    md_attr->alloc_mem_types  = UCS_BIT(UCS_MEMORY_TYPE_CUDA) |
+                                UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
+    md_attr->access_mem_types = UCS_BIT(UCS_MEMORY_TYPE_CUDA) |
+                                UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
+    md_attr->detect_mem_types = UCS_BIT(UCS_MEMORY_TYPE_CUDA) |
+                                UCS_BIT(UCS_MEMORY_TYPE_CUDA_MANAGED);
+    md_attr->dmabuf_mem_types = md->config.dmabuf_supported ?
+                                UCS_BIT(UCS_MEMORY_TYPE_CUDA) : 0;
     md_attr->max_alloc        = SIZE_MAX;
-    md_attr->max_reg          = ULONG_MAX;
-    md_attr->rkey_packed_size = 0;
-    md_attr->reg_cost         = UCS_LINEAR_FUNC_ZERO;
-    memset(&md_attr->local_cpus, 0xff, sizeof(md_attr->local_cpus));
     return UCS_OK;
 }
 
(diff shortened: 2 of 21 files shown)


test/gtest/uct/test_md.cc
```diff
@@ -66,12 +66,18 @@ void* test_md::alloc_thread(void *arg)
 ucs_status_t test_md::reg_mem(unsigned flags, void *address, size_t length,
                               uct_mem_h *memh_p)
 {
+    /* Register memory respecting MD reg_alignment */
+    size_t reg_align  = ucs_max(md_attr().reg_alignment, 1);
+    void *reg_address = address;
```
tvegas1, line 71: can simply pass length and addr to ucx_align_ptr_range()

src/tools/perf/cuda/cuda_alloc.c
```diff
@@ -69,15 +71,31 @@ uct_perf_cuda_alloc_reg_mem(const ucx_perf_context_t *perf,
                             unsigned flags,
                             uct_allocated_memory_t *alloc_mem)
 {
+    uct_md_attr_v2_t md_attr = { .field_mask = UINT64_MAX };
```
tvegas1, line 74: since this is non-test code, maybe we want to use only `UCT_MD_ATTR_FIELD_REG_ALIGNMENT`.

tvegas1 (commented):
cosmetic

src/tools/perf/cuda/cuda_alloc.c
iyastreb: Thanks, makes sense

test/gtest/uct/test_md.cc
iyastreb: Right, done

tvegas1 (approved):

test/gtest/uct/test_md.cc
```diff
@@ -276,7 +280,8 @@ void test_md::dereg_cb(uct_completion_t *comp)
 bool test_md::is_gpu_ipc() const
 {
     return (GetParam().md_name == "cuda_ipc") ||
-	   (GetParam().md_name == "rocm_ipc");
+           (GetParam().md_name == "rocm_ipc") ||
+           (GetParam().md_name == "gdr_copy");
```
brminich, line 284: why is it needed? it's not really a gpu_ipc transport

test/gtest/uct/uct_test.cc
```diff
@@ -967,7 +968,13 @@ void uct_test::entity::mem_type_reg(uct_allocated_memory_t *mem,
...
 {
     if (md_attr().reg_mem_types & UCS_BIT(mem->mem_type)) {
-        ucs_status_t status = uct_md_mem_reg(m_md, mem->address, mem->length,
+        /* Register memory respecting MD reg_alignment */
+        size_t reg_align  = ucs_max(md_attr().reg_alignment, 1);
+        void *reg_address = mem->address;
+        size_t reg_length = mem->length;
+        ucs_align_ptr_range(&reg_address, &reg_length, reg_align);
+
+        ucs_status_t status = uct_md_mem_reg(m_md, reg_address, reg_length,
```
brminich, line 977: better to use `uct_md_mem_reg_v2`

test/gtest/uct/test_md.cc
brminich: can you update the test instead. E.g. like this:
query md attrs with `uct_md_query_v2`
skip the test if `md_attr->reg_mem_types` does not contain host memory

src/tools/perf/cuda/cuda_alloc.c
```diff
@@ -69,15 +71,31 @@ uct_perf_cuda_alloc_reg_mem(const ucx_perf_context_t *perf,
...
+
     status = ucx_perf_cuda_alloc(length, mem_type, &alloc_mem->address);
     if (status != UCS_OK) {
         return status;
     }
 
-    status = uct_md_mem_reg(perf->uct.md, alloc_mem->address,
-                            length, flags, &alloc_mem->memh);
+    /* Register memory respecting MD reg_alignment */
+    reg_align   = ucs_max(md_attr.reg_alignment, 1);
```
yosefe, line 92: why needed? is there an MD that returns 0 here? if yes it should be fixed

src/tools/perf/cuda/cuda_alloc.c
```diff
@@ -69,15 +71,31 @@ uct_perf_cuda_alloc_reg_mem(const ucx_perf_context_t *perf,
...
+    reg_address = alloc_mem->address;
+    reg_length  = length;
```
yosefe, line 94: IMO we can remove these local helper vars

test/gtest/uct/test_md.cc
```diff
@@ -66,6 +66,10 @@ void* test_md::alloc_thread(void *arg)
 ucs_status_t test_md::reg_mem(unsigned flags, void *address, size_t length,
                               uct_mem_h *memh_p)
 {
+    /* Register memory respecting MD reg_alignment */
+    size_t reg_align  = ucs_max(md_attr().reg_alignment, 1);
```
yosefe, line 70: 1. same here, need to fix UCT MD if it returns 0
2. we should make sure the test allocates aligned memory buffer and not just align during registration. Otherwise, we could try to register pages not in the process address space.

test/gtest/uct/test_md.cc
yosefe: Right, as @brminich the right way to to check if the MD can register host memory .
Maybe after this, we could remove is_gpu_ipc condition as well to cleanup the code

test/gtest/uct/uct_test.cc
```diff
@@ -967,7 +968,13 @@ void uct_test::entity::mem_type_reg(uct_allocated_memory_t *mem,
                                     unsigned mem_flags) const
 {
     if (md_attr().reg_mem_types & UCS_BIT(mem->mem_type)) {
-        ucs_status_t status = uct_md_mem_reg(m_md, mem->address, mem->length,
+        /* Register memory respecting MD reg_alignment */
+        size_t reg_align  = ucs_max(md_attr().reg_alignment, 1);
+        void *reg_address = mem->address;
+        size_t reg_length = mem->length;
+        ucs_align_ptr_range(&reg_address, &reg_length, reg_align);
+
```
yosefe, line 976: 1. code duplication with test_md
2. cannot register outside the allocated buffer

yosefe (changes requested):

src/tools/perf/cuda/cuda_alloc.c
iyastreb: Well, as I see it's 0s for all MDs, expect gdr_copy:
```
[     INFO ] >>> knem 0
[     INFO ] >>> cma 0
[     INFO ] >>> posix 0
[     INFO ] >>> sysv 0
[     INFO ] >>> xpmem 0
[     INFO ] >>> cuda_cpy 0
[     INFO ] >>> cuda_ipc 0
[     INFO ] >>> mlx5_0 0
[     INFO ] >>> mlx5_1 0
[     INFO ] >>> mlx5_2 0
[     INFO ] >>> gdr_copy 65536
```
And the second reason: I looked at the existing handling of this param in production code: https://github.com/openucx/ucx/blob/master/src/ucp/core/ucp_mm.c#L525
```
        if (context->rcache == NULL) {
            reg_align = ucs_max(context->tl_mds[md_index].attr.reg_alignment, 1);
            ucs_align_ptr_range(&reg_address, &reg_length, reg_align);
        }
```
Should we set this attribute to 1 everywhere by default?
I quickly checked places where this field is used, should be ok to change it to 1

src/tools/perf/cuda/cuda_alloc.c
iyastreb: reg_align - we can remove it if agree to init reg_alignment to 1 for all MDs
reg_address - we cannot remove it, because we don't want to modify `alloc_mem->address`
reg_length - this is safe to remove

test/gtest/uct/test_md.cc
iyastreb: Ok, agree, will be fixed

test/gtest/uct/test_md.cc
iyastreb: 1. ok
2. hmm, are we absolutely sure about this?
Looking at [production code](https://github.com/openucx/ucx/blob/master/src/ucp/core/ucp_mm.c#L525), it seems we do exactly that, or I misunderstood it?
Finally we just register the whole page, so it must be within the process address space.

test/gtest/uct/uct_test.cc
iyastreb: 1. will think how to get rid of the code duplication
2. Let's discuss this in person?

src/tools/perf/cuda/cuda_alloc.c
yosefe: > Should we set this attribute to 1 everywhere by default?
Yes

test/gtest/uct/test_md.cc
yosefe: 2. ok

src/tools/perf/cuda/cuda_alloc.c
yosefe: ok

test/gtest/uct/uct_test.cc
iyastreb: 1. there is no duplication with test_md anymore
2. it's ok as we agreed

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -118,6 +118,7 @@ uct_cuda_copy_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
     md_attr->max_reg          = ULONG_MAX;
     md_attr->rkey_packed_size = 0;
     md_attr->reg_cost         = UCS_LINEAR_FUNC_ZERO;
+    md_attr->reg_alignment    = UCS_PGT_REG_ALIGNMENT;
```
Artemy-Mellanox, line 121: Maybe we could add some base_md_query() function in uct/base/uct_md.c to set default values like this, and then just call it from all MDs before tweaking specific values. That would make it easier to add or tweak fields that only differ in one MD.

yosefe: sounds good!

iyastreb: done

src/ucs/datastruct/pgtable.h
```diff
@@ -28,6 +28,7 @@
 #define UCS_PGT_ADDR_ALIGN         (1ul << UCS_PGT_ADDR_SHIFT)
 #define UCS_PGT_ADDR_ORDER          (sizeof(ucs_pgt_addr_t) * 8)
 #define UCS_PGT_ADDR_MAX           ((ucs_pgt_addr_t)-1)
+#define UCS_PGT_REG_ALIGNMENT      1
```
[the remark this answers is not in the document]
iyastreb: ok

src/uct/cuda/gdr_copy/gdr_copy_md.c
```diff
@@ -37,6 +37,7 @@ static ucs_config_field_t uct_gdr_copy_md_config_table[] = {
 static ucs_status_t
 uct_gdr_copy_md_query(uct_md_h md, uct_md_attr_v2_t *md_attr)
 {
+    uct_md_base_md_query(md_attr);
```
yosefe, line 40: all the "*mem_types" fields can also be set to 0 by default

iyastreb: ok, will be fixed

yosefe (approved):

tvegas1 (approved):

brminich (approved):

PR #8225, opened 2022-05-15
UCT/IB: Remove IB registration methods

Artemy-Mellanox:

PR #7583, opened 2021-10-26
UCP/PROTO: Adjust performace estimations

Artemy-Mellanox:

PR #9273, opened 2023-08-07
UCP/AZP: Enhance wire compat CI and fix issues

brminich:
## What
- Enable sa_data format by default
- Fix wire compatibility CI
- Define wire/release version and add to the worker address, so it is possible to get remote peer version and adjust proto selection algorithms accordingly
       
## Why ?
to preserve wire compatibility with older UCX versions

PR #9847, opened 2024-04-26
UCP/CORE: Use rcache overhead env for protocol perf estimation.

rakhmets:
## What
Removed `uct_md_rcache_overhead`. The method became unused after merging https://github.com/openucx/ucx/pull/8225.
Added logic to use previously unused `UCX_RCACHE_OVERHEAD` environment variable. The variable can be used to configure the `rcache lookup` node (applied suggestion https://github.com/openucx/ucx/pull/8995#discussion_r1160421878).
Added `UCX_RCACHE_OVERHEAD` to `[Fujitsu ARM]` section to restore adjustments made by https://github.com/openucx/ucx/pull/7583.

### DIFF
--- a/config/ucx.conf
+++ b/config/ucx.conf
@@ -11,3 +11,4 @@ UCX_IB_PCI_RELAXED_ORDERING=try
 UCX_IB_SEND_OVERHEAD=bcopy:5ns,cqe:50ns,db:400ns,wqe_fetch:350ns,wqe_post:100ns
 UCX_MM_SEND_OVERHEAD=am_short:40ns,am_bcopy:220ns
 UCX_MM_RECV_OVERHEAD=am_short:40ns,am_bcopy:220ns
+UCX_RCACHE_OVERHEAD=360ns
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -181,6 +181,8 @@ typedef struct ucp_context_config {
     double                                 proto_overhead_rndv_rtr;
     double                                 proto_overhead_sw;
     double                                 proto_overhead_rkey_ptr;
+    /** Registration cache lookup overhead estimation */
+    double                                 rcache_overhead;
 } ucp_context_config_t;
 
 
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -1568,6 +1568,9 @@ ucs_status_t ucp_mem_rcache_init(ucp_context_h context,
         }
     }
 
+    context->config.ext.rcache_overhead = ucs_time_units_to_sec(
+            rcache_config->overhead, UCP_RCACHE_OVERHEAD_DEFAULT);
+
     return UCS_OK;
 
 err_rcache_destroy:
--- a/src/ucp/proto/proto_init.c
+++ b/src/ucp/proto/proto_init.c
@@ -315,7 +315,8 @@ void ucp_proto_init_memreg_time(const ucp_proto_common_init_params_t *params,
     if (context->rcache != NULL) {
         perf_node = ucp_proto_perf_node_new_data("rcache lookup", "");
 
-        *memreg_time = UCP_RCACHE_LOOKUP_FUNC;
+        *memreg_time = ucs_linear_func_make(context->config.ext.rcache_overhead,
+                                            0);
 
         ucp_proto_perf_node_add_data(perf_node, "lookup", *memreg_time);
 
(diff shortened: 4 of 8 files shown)


src/ucp/wireup/select.c
```diff
@@ -1295,7 +1295,7 @@ ucp_wireup_mem_reg_cost(ucp_context_t *context,
     }
 
     if (remote_addr->dst_version > UCP_RELEASE_LEGACY) {
-        return UCP_RCACHE_LOOKUP_FUNC;
```
rakhmets, line 1298: It seems that it shouldn't be configurable, to avoid wire compatibility issues.

@brminich

src/ucp/core/ucp_mm.c
```diff
@@ -1568,6 +1568,9 @@ ucs_status_t ucp_mem_rcache_init(ucp_context_h context,
         }
     }
 
+    context->config.ext.rcache_overhead = ucs_linear_func_make(
+            ucs_time_units_to_sec(rcache_config->overhead, 50.0e-9), 0);
```
yosefe, line 1572: seems the prev default was 180e-9, why changed?

ivankochin: As I understand it is changed to 50 since before that patch that value was mostly used for UCT level rcache, but now it is used for UCP level rcache, and the latest measurements on Selene shown that it is 50 nanoseconds which is exposed inside `UCP_RCACHE_LOOKUP_FUNC`

src/ucp/proto/proto_init.c
```diff
@@ -315,7 +315,7 @@ void ucp_proto_init_memreg_time(const ucp_proto_common_init_params_t *params,
     if (context->rcache != NULL) {
         perf_node = ucp_proto_perf_node_new_data("rcache lookup", "");
 
-        *memreg_time = UCP_RCACHE_LOOKUP_FUNC;
```
ivankochin, line 318: The only another usage of `UCP_RCACHE_LOOKUP_FUNC` is inside `ucp_wireup_mem_reg_cost`, shouldn't we use the value from the config there too? If so, then `UCP_RCACHE_LOOKUP_FUNC` can be removed.

config/ucx.conf
```diff
@@ -11,3 +11,4 @@ UCX_IB_PCI_RELAXED_ORDERING=try
 UCX_IB_SEND_OVERHEAD=bcopy:5ns,cqe:50ns,db:400ns,wqe_fetch:350ns,wqe_post:100ns
 UCX_MM_SEND_OVERHEAD=am_short:40ns,am_bcopy:220ns
 UCX_MM_RECV_OVERHEAD=am_short:40ns,am_bcopy:220ns
+UCX_RCACHE_OVERHEAD=360ns
```
ivankochin, line 14: In which scenarios does this config used? And why do we set 360ns here?

src/ucp/proto/proto_init.c
rakhmets: I'm afraid that it will break wire compatibility.
@brminich fixed similar issue here: https://github.com/openucx/ucx/pull/9273.

brminich (approved):

src/ucp/proto/proto_init.c
ivankochin: But shouldn't it be same values? Maybe it worth applying the new value by wire compatible way (with peer version check)?

brminich: we have it defined in wire-compatible way in wireup (`ucp_wireup_mem_reg_cost`). here wire compatibility is not important.

ivankochin (dismissed):

src/ucp/core/ucp_context.h
```diff
@@ -181,6 +181,8 @@ typedef struct ucp_context_config {
     double                                 proto_overhead_rndv_rtr;
     double                                 proto_overhead_sw;
     double                                 proto_overhead_rkey_ptr;
+    /** Registration cache lookup overhead estimation */
+    ucs_linear_func_t                      rcache_overhead;
```
yosefe, line 185: maybe save it as "double" and not a linear func?

src/ucp/core/ucp_mm.c
yosefe: shall we reuse (or delete) UCP_RCACHE_LOOKUP_FUNC?

rakhmets: Updated macro.

yosefe (approved):

brminich (approved):

PR #9850, opened 2024-04-28
TOOLS/PERF/TEST: Fix build without openmp

### LINKED ISSUE #9848: OpenMP required by `libperf.c`'s `ucx_perf_do_warmup`
Most code in UCX that uses OpenMP is guarded to check that OpenMP is enabled. If it is not, there is fallback logic that doesn't use OpenMP

For example this happens in `libperf_thread.c` where an OpenMP block is provided if OpenMP is found

https://github.com/openucx/ucx/blob/3c6b20db9fec8204cfe9f2744aa519efef9de8f8/src/tools/perf/lib/libperf_thread.c#L21-L25

Otherwise a non-OpenMP fallback is provided

https://github.com/openucx/ucx/blob/3c6b20db9fec8204cfe9f2744aa519efef9de8f8/src/tools/perf/lib/libperf_thread.c#L138-L147

As a result any `#pragma omp` lines live in the OpenMP enabled block

https://github.com/openucx/ucx/blob/3c6b20db9fec8204cfe9f2744aa519efef9de8f8/src/tools/perf/lib/libperf_thread.c#L44

So are not encountered outside it

<hr>

With `libperf.c` however the OpenMP `#include` is optional

https://github.com/openucx/ucx/blob/3c6b20db9fec8204cfe9f2744aa519efef9de8f8/src/tools/perf/lib/libperf.c#L26-L28

Though later in the code `#pragma omp` is not guarded

https://github.com/openucx/ucx/blob/3c6b20db9fec8204cfe9f2744aa519efef9de8f8/src/tools/perf/lib/libperf.c#L1702-L1706

As a result building without OpenMP leads to errors like this

```C
make[3]: Entering directory '/home/conda/feedstock_root/build_artifacts/ucx-split_1714181624237/work/src/tools/perf/lib'
  CC       libucxperf_la-libperf.lo
  CC       libucxperf_la-libperf_memory.lo
libperf.c: In function 'ucx_perf_do_warmup':
libperf.c:1356: error: ignoring '#pragma omp barrier' [-Werror=unknown-pragmas]
 1356 | #pragma omp barrier
      | 
libperf.c:1357: error: ignoring '#pragma omp single' [-Werror=unknown-pragmas]
 1357 | #pragma omp single copyprivate(status)
      | 
libperf.c:1360: error: ignoring '#pragma omp barrier' [-Werror=unknown-pragmas]
 1360 | #pragma omp barrier
      | 
```

This comes from a UCX 1.16.0 build. Here is [the full build log]( https://github.com/openucx/ucx/files/15136778/ucx-split-feedstock_20240426.14_linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11.txt )

<hr>

Would it be possible to guard the OpenMP code above and provide a fallback to allow building without OpenMP?

yosefe:
## Why
Fixes #9848

### DIFF
--- a/buildlib/tools/builds.sh
+++ b/buildlib/tools/builds.sh
@@ -303,6 +303,10 @@ build_no_devx() {
 	build_gcc --with-devx=no
 }
 
+build_no_openmp() {
+	build_gcc --disable-openmp
+}
+
 build_gcc_debug_opt() {
 	build_gcc CFLAGS=-Og CXXFLAGS=-Og
 }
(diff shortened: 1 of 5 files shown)


buildlib/tools/builds.sh
```diff
@@ -427,8 +431,9 @@ do_task() {
 az_init_modules
 prepare_build
 
-[ "${long_test}" = "yes" ] && prog=5 || prog=12
+[ "${long_test}" = "yes" ] && prog=5 || prog=10
```
tvegas1, line 434: maybe we could use array of commands, and set `prog` to 100/array_length:
- `tests=('check_config_h', 'check_inst_headers', ...)` + long runs and then:
- `for test in "${tests[@]}"; do`

src/tools/perf/lib/libperf.c
```diff
@@ -1696,14 +1696,17 @@ ucx_perf_do_warmup(ucx_perf_context_t *perf, const ucx_perf_params_t *params)
         /* Stop when reaching the deadline */
         stop_status = (ucs_get_time() > deadline) ? UCS_OK : UCS_INPROGRESS;
 
-        if (params->thread_count == 1) {
-            status = ucx_perf_test_exchange_status(perf, stop_status);
-        } else {
-#pragma omp barrier
-#pragma omp single copyprivate(status)
+#if _OPENMP
+        if (params->thread_count != 1) {
```
tvegas1, line 1700: can we remove `if (params->thread_count != 1`? If we have only one thread, the sequence barrier/single copyprivate/barrier would still be correct. do you see be perf issue there?

src/tools/perf/lib/libperf.c
```diff
@@ -1696,19 +1696,17 @@ ucx_perf_do_warmup(ucx_perf_context_t *perf, const ucx_perf_params_t *params)
...
-            /* Synchronize on whether to continue or stop the warmup phase */
-            status = ucx_perf_test_exchange_status(perf, stop_status);
-#  pragma omp barrier
-        } else
 #endif
         {
             status = ucx_perf_test_exchange_status(perf, stop_status);
         }
-
+#if _OPENMP
```
tvegas1, line 1707: we can remove { and } line 1704 and 1706

yosefe: right, since the pragma applies to a single statement, but for consistency of how we use pragma in other places and avoid potential bug of adding more statements, IMO {} is beter

tvegas1 (dismissed):

yosefe:
@tvegas1 can you pls review again?

buildlib/tools/builds.sh
```diff
@@ -408,47 +412,45 @@ build_fuse() {
...
+for ((i=0;i<${num_tests};++i))
+do
+	test_name=${tests[$i]}
+
+	# update progress indicator
+	progress=$(( (i * 100) / num_tests ))
+	echo "##vso[task.setprogress value=${progress};]${test_name}"
+
+	# cleanup build dir before the task
+	[ -n "${ucx_build_dir}" ] && rm -rf ${ucx_build_dir}/*
```
tvegas1, line 452: detail: i think using `-d` instead `-n` would cover case where dir is not specified (and we don't want to rm -rf the root), and also if dir does not exist

tvegas1 (approved):
cosmetic comment since dir exists in practice

tvegas1 (approved):

PR #9863, opened 2024-05-06
UCP/CORE: Fix for activate count

iyastreb:
There is a bug introduced in recent commit https://github.com/openucx/ucx/pull/9650, which is exposed in 2 ways:
- assertion failure in debug build in function `ucp_worker_iface_deactivate : wiface->activate_count > 0`
- in release mode process hangs
Revert of that commit solves the connectivity issue both for daemon CI use case and RFI.

The issue happens on the receiver side, only when there are multiple connections from the same sender. In this case there are multiple wireup_eps created and they reuse the same ep_config. However, increments and decrements of interface reference counter are not consistent:
```
First sender connect:
  ucp_worker_get_ep_config -> Load new config
    ucp_worker_ep_config_short_init -> Select protocols
      ucp_wiface_process_for_each_lane -> activate_count ++
  ucp_ep_set_cfg_index -> Set config to EP
    ucp_ep_config_activate_worker_ifaces
      ucp_wiface_process_for_each_lane -> activate_count ++
First sender disconnect
  ucp_wireup_process_request -> Handle disconnect event
    ucp_ep_set_cfg_index -> Update config, remove active interfaces
      ucp_wiface_process_for_each_lane -> activate_count --
  ucp_wireup_eps_progress -> Wireup EP gets deleted
    ucp_proxy_ep_replace
      ucp_wireup_ep_t_cleanup
        ucp_wireup_ep_discard_aux_ep
          ucp_worker_iface_unprogress_ep -> activate_count --  

Now the second client connects and reuses the same ep_config:
  ucp_worker_get_ep_config -> Reuse existing config
    No increment happens here, just return <<< Fix here?
  ucp_ep_set_cfg_index -> Set config to EP
    ucp_ep_config_activate_worker_ifaces
      ucp_wiface_process_for_each_lane -> activate_count ++
Second sender disconnect
  ucp_wireup_process_request -> Handle disconnect event
    ucp_ep_set_cfg_index -> Update config, remove active interfaces
      ucp_wiface_process_for_each_lane -> activate_count -- (HERE refcount reaches 0, interface gets deactivated)
  ucp_wireup_eps_progress -> Wireup EP gets deleted
    ucp_proxy_ep_replace
      ucp_wireup_ep_t_cleanup
        ucp_wireup_ep_discard_aux_ep
          ucp_worker_iface_unprogress_ep -> activate_count -- (Failure: activate_count == 0)
```
As we agreed after internal discussion, the idea is to avoid iface activation during config creation. Instead, we do iface activation on `ucp_ep_set_cfg_index`. Also add missing iface activation for AUX EP

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -3812,6 +3812,76 @@ ucs_status_t ucp_ep_realloc_lanes(ucp_ep_h ep, unsigned new_num_lanes)
     return UCS_OK;
 }
 
+static void
+ucp_ep_select_short_init(ucp_worker_h worker, ucp_worker_cfg_index_t cfg_index,
+                         unsigned feature_flag, ucp_operation_id_t op_id,
+                         unsigned proto_flags, ucp_lane_index_t exp_lane,
+                         ucp_memtype_thresh_t *max_eager_short)
+{
+    ucp_ep_config_t *ep_config = ucp_worker_ep_config(worker, cfg_index);
+    ucp_proto_select_short_t proto_short;
+
+    if (worker->context->config.features & feature_flag) {
+        ucp_proto_select_short_init(worker, &ep_config->proto_select,
+                                    cfg_index, UCP_WORKER_CFG_INDEX_NULL,
+                                    op_id, proto_flags, &proto_short);
+
+        /* Short protocol should be either disabled, or use expected lane */
+        ucs_assertv((proto_short.max_length_host_mem < 0) ||
+                    (proto_short.lane == exp_lane),
+                    "max_length_host_mem %ld, lane %d",
+                    proto_short.max_length_host_mem, proto_short.lane);
+    } else {
+        ucp_proto_select_short_disable(&proto_short);
+    }
+
+    max_eager_short->memtype_off = proto_short.max_length_unknown_mem;
+    max_eager_short->memtype_on  = proto_short.max_length_host_mem;
+}
+
+static void ucp_ep_config_proto_init(ucp_worker_h worker,
+                                     ucp_worker_cfg_index_t cfg_index)
+{
+    ucp_ep_config_t *ep_config = ucp_worker_ep_config(worker, cfg_index);
+    ucp_ep_config_key_t *key   = &ep_config->key;
+
+    ucp_memtype_thresh_t *tag_max_short;
+    ucp_lane_index_t tag_exp_lane;
+    unsigned tag_proto_flags;
+
+    /* Do protocol init once per EP config and only for protov2 */
+    if ((!worker->context->config.ext.proto_enable) ||
+        (ep_config->proto_init_flags & UCP_EP_PROTO_INITIALIZED)) {
+        return;
+    }
+
+    ep_config->proto_init_flags |= UCP_EP_PROTO_INITIALIZED;
+
+    if (ucp_ep_config_key_has_tag_lane(key)) {
+        tag_proto_flags = UCP_PROTO_FLAG_TAG_SHORT;
+        tag_max_short   = &ep_config->tag.offload.max_eager_short;
+        tag_exp_lane    = key->tag_lane;
+    } else {
+        tag_proto_flags = UCP_PROTO_FLAG_AM_SHORT;
+        tag_max_short   = &ep_config->tag.max_eager_short;
+        tag_exp_lane    = key->am_lane;
+    }
+
+    ucp_ep_select_short_init(worker, cfg_index,
+                             UCP_FEATURE_TAG, UCP_OP_ID_TAG_SEND,
+                             tag_proto_flags, tag_exp_lane, tag_max_short);
+
+    ucp_ep_select_short_init(worker, cfg_index,
+                             UCP_FEATURE_AM, UCP_OP_ID_AM_SEND,
+                             UCP_PROTO_FLAG_AM_SHORT, key->am_lane,
+                             &ep_config->am_u.max_eager_short);
+
+    ucp_ep_select_short_init(worker, cfg_index,
+                             UCP_FEATURE_AM, UCP_OP_ID_AM_SEND_REPLY,
+                             UCP_PROTO_FLAG_AM_SHORT, key->am_lane,
+                             &ep_config->am_u.max_reply_eager_short);
+}
+
 void ucp_ep_set_cfg_index(ucp_ep_h ep, ucp_worker_cfg_index_t cfg_index)
 {
     if (ep->cfg_index != UCP_WORKER_CFG_INDEX_NULL) {
@@ -3820,4 +3890,5 @@ void ucp_ep_set_cfg_index(ucp_ep_h ep, ucp_worker_cfg_index_t cfg_index)
 
     ep->cfg_index = cfg_index;
     ucp_ep_config_activate_worker_ifaces(ep->worker, cfg_index);
+    ucp_ep_config_proto_init(ep->worker, cfg_index);
 }
--- a/src/ucp/core/ucp_ep.h
+++ b/src/ucp/core/ucp_ep.h
@@ -370,6 +370,12 @@ typedef struct {
 KHASH_DECLARE(ucp_ep_peer_mem_hash, uint64_t, ucp_ep_peer_mem_data_t);
 
 
+typedef enum {
+    /* Protocol initialization was done */
+    UCP_EP_PROTO_INITIALIZED = UCS_BIT(0),
+} ucp_ep_init_flags_t;
+
+
 struct ucp_ep_config {
 
     /* A key which uniquely defines the configuration, and all other fields of
@@ -476,6 +482,9 @@ struct ucp_ep_config {
 
     /* Bitmap of lanes selected by the protocols */
     ucp_lane_map_t                proto_lane_map;
+
+    /* EP initialization flags from @ref ucp_ep_init_flags_t */
+    unsigned                      proto_init_flags;
 
     /* Number of endpoints using this configuration */
     unsigned                      ep_count;
--- a/src/ucp/wireup/wireup_ep.c
+++ b/src/ucp/wireup/wireup_ep.c
@@ -204,6 +204,9 @@ UCS_CLASS_DEFINE_NAMED_NEW_FUNC(ucp_wireup_ep_create, ucp_wireup_ep_t, uct_ep_t,
 void ucp_wireup_ep_set_aux(ucp_wireup_ep_t *wireup_ep, uct_ep_h uct_ep,
                            ucp_rsc_index_t rsc_index, int is_p2p)
 {
+    ucp_worker_iface_t *wiface =
+        ucp_worker_iface(wireup_ep->super.ucp_ep->worker, rsc_index);
+
     ucs_assert(!ucp_wireup_ep_test(uct_ep));
     wireup_ep->aux_ep        = uct_ep;
     wireup_ep->aux_rsc_index = rsc_index;
@@ -211,6 +214,8 @@ void ucp_wireup_ep_set_aux(ucp_wireup_ep_t *wireup_ep, uct_ep_h uct_ep,
     if (is_p2p) {
         wireup_ep->flags |= UCP_WIREUP_EP_FLAG_AUX_P2P;
     }
+
+    ucp_worker_iface_progress_ep(wiface);
 }
 
 static ucs_status_t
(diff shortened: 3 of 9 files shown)


src/ucp/core/ucp_ep.h
```diff
@@ -477,6 +477,9 @@ struct ucp_ep_config {
     /* Bitmap of lanes selected by the protocols */
     ucp_lane_map_t                proto_lane_map;
 
+    /* Indicate that protocol initialization was done for config */
+    int                           is_proto_initialized;
```
brminich, line 481: maybe define common flags field + enum with one element (so we can use/add more flags in the future)

src/ucp/core/ucp_ep.c
```diff
@@ -3812,6 +3812,70 @@ ucs_status_t ucp_ep_realloc_lanes(ucp_ep_h ep, unsigned new_num_lanes)
...
+                    proto_short.max_length_host_mem, proto_short.lane);
+    } else {
+        ucp_proto_select_short_disable(&proto_short);
+    }
+
+    max_eager_short->memtype_off = proto_short.max_length_unknown_mem;
+    max_eager_short->memtype_on  = proto_short.max_length_host_mem;
+}
+
+static void ucp_ep_select_short_init_all(ucp_ep_h ep)
```
yosefe, line 3842: 1. rename it to ucp_ep_config_proto_init
2. remove ep parameter. pass only worker and cfg_index
3. check proto_enable inside this function

src/ucp/wireup/wireup_ep.c
```diff
@@ -204,6 +204,10 @@ UCS_CLASS_DEFINE_NAMED_NEW_FUNC(ucp_wireup_ep_create, ucp_wireup_ep_t, uct_ep_t,
 void ucp_wireup_ep_set_aux(ucp_wireup_ep_t *wireup_ep, uct_ep_h uct_ep,
                            ucp_rsc_index_t rsc_index, int is_p2p)
 {
+    ucp_worker_iface_t *wiface = ucp_worker_iface(wireup_ep->super.ucp_ep->worker, rsc_index);
+
+    ucp_worker_iface_progress_ep(wiface);
```
yosefe, line 209: IMOP better move it to the end of the function

src/ucp/wireup/wireup_ep.c
```diff
@@ -204,6 +204,10 @@ UCS_CLASS_DEFINE_NAMED_NEW_FUNC(ucp_wireup_ep_create, ucp_wireup_ep_t, uct_ep_t,
 void ucp_wireup_ep_set_aux(ucp_wireup_ep_t *wireup_ep, uct_ep_h uct_ep,
                            ucp_rsc_index_t rsc_index, int is_p2p)
 {
+    ucp_worker_iface_t *wiface = ucp_worker_iface(wireup_ep->super.ucp_ep->worker, rsc_index);
```
yosefe, line 207: line seems too long

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3232,3 +3235,82 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_err_sender,
...
+    {
+        return test_ucp_sockaddr::is_skip_interface(ifa) ||
+               ucs::is_rdmacm_netdev(ifa->ifa_name);
+    }
+
+    template <typename Func>
+    void wait_progress(Func func)
+    {
+        ucs_time_t deadline = ucs::get_deadline();
+        scoped_log_handler slh(detect_error_logger);
```
yosefe, line 3260: why needed?

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3232,3 +3235,82 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_err_sender,
...
+    }
+
+    virtual int is_skip_interface(const struct ifaddrs *ifa) const
+    {
+        return test_ucp_sockaddr::is_skip_interface(ifa) ||
+               ucs::is_rdmacm_netdev(ifa->ifa_name);
+    }
+
+    template <typename Func>
+    void wait_progress(Func func)
```
yosefe, line 3257: can we move it to ucp_test class?

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3232,3 +3235,82 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_err_sender,
...
+    {
+        ucs_time_t deadline = ucs::get_deadline();
+        scoped_log_handler slh(detect_error_logger);
+        while (!func() && (ucs_get_time() < deadline)) {
+            progress();
+        };
+    }
+
+    ucp_worker_iface_t *get_test_iface(entity &e) const
```
yosefe, line 3266: IMO we can check if any iface is activated, not check only the one we connect to. because we have TEST_MODIFIER_CM_USE_ALL_DEVICES, other devices can be used as well.

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3232,3 +3235,82 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_err_sender,
...
+        receiver().reset_err_num();
+    }
+};
+
+UCS_TEST_P(test_ucp_sockaddr_iface_activate, iface_activate_count)
+{
+    create_entity(true);
+    listen(cb_type());
+    client_connect_disconnect(0);
+    client_connect_disconnect(1);
```
yosefe, line 3313: why need to use different index (1) here, if index 0 is already disconnected?
maybe reuse index 0 and then could avoid adding sender_index parameter to different function

test/gtest/ucp/test_ucp_sockaddr.cc
iyastreb: This does not work like that, at least on the receiver side, just tested.
```
[     INFO ] receiver after connect:                                                                                                                                                                                               
[     INFO ] iface 0: ens10f0 acount: 0                                                                                                                                                                                                    
[     INFO ] iface 1: ib0 acount: 0                                                                                                                                                                                                        
[     INFO ] iface 2: ib1 acount: 0                                                                                                                                                                                                        
[     INFO ] iface 3: lo acount: 1                                                                                                                                                                                                         
```
activate_count gets incremented only on the interface we connect to.

src/ucp/core/ucp_ep.h
```diff
@@ -477,8 +482,8 @@ struct ucp_ep_config {
     /* Bitmap of lanes selected by the protocols */
     ucp_lane_map_t                proto_lane_map;
 
-    /* Indicate that protocol initialization was done for config */
-    int                           is_proto_initialized;
+    /* EP initialization flags */
+    ucp_ep_init_flags_t           proto_init_flags;
```
yosefe, line 486: unsigned (since it's not really an enum)

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -619,6 +618,18 @@ class test_ucp_sockaddr : public ucp_test {
...
+    template <typename Func>
+    void wait_progress(Func func)
+    {
+        ucs_time_t deadline = ucs::get_deadline();
+
+        while (!func() && (ucs_get_time() < deadline)) {
+            progress();
+        };
+
+        EXPECT_GT(deadline, ucs_get_time());
```
yosefe, line 630: can remove space lines

test/gtest/ucp/test_ucp_sockaddr.cc
yosefe: i think in some tests it may work like that, anyway seems more fail-safe to me

test/gtest/ucp/test_ucp_sockaddr.cc
iyastreb: ok, removed just the first one
IIRC we need a space line after code block within {} brackets

src/ucp/core/ucp_ep.h
iyastreb: right, fixed

test/gtest/ucp/test_ucp_sockaddr.cc
yosefe: @iyastreb seems this comment was not addressed , WDYT?

src/ucp/core/ucp_ep.h
```diff
@@ -370,6 +370,11 @@ typedef struct {
 KHASH_DECLARE(ucp_ep_peer_mem_hash, uint64_t, ucp_ep_peer_mem_data_t);
 
 
+typedef enum {
+    UCP_EP_PROTO_INIT = UCS_BIT(0), /* Protocol initialization was done */
```
brminich, line 374: maybe
```suggestion
    UCP_EP_PROTO_INITIALIZED = UCS_BIT(0), /* Protocol initialization was done */
```

brminich (approved):

test/gtest/ucp/test_ucp_sockaddr.cc
yosefe: I think it is better to check activate count globally and not for a specific device since it will make code simpler and also cover more cases

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3283,27 +3282,33 @@ class test_ucp_sockaddr_iface_activate : public test_ucp_sockaddr {
         /* Check state before connection establishment */
         ucp_worker_iface_t *sender_iface = get_test_iface(sender());
         ucp_worker_iface_t *server_iface = get_test_iface(receiver());
-        EXPECT_EQ(0, ucp_worker_iface_is_activated(server_iface));
+        int receiver_ep_count            = receiver().get_num_eps();
+        if (receiver_ep_count == 0) {
+            EXPECT_EQ(0, ucp_worker_iface_is_activated(sender_iface));
```
yosefe, line 3287: EXPECT_FALSE (multiple locations)

test/gtest/ucp/test_ucp_sockaddr.cc
iyastreb: Ok, I implemented that
For that to work I had to explicitly disconnect both client and receiver EPs, while previously I was only disconnecting on the client side. With receiver disconnect we can check that all interfaces are deactivated.
(This updated test still reproduces the issue on the branch without fix, so we are covered here)

yosefe: > Ok, I implemented that For that to work I had to explicitly disconnect both client and receiver EPs, while previously I was only disconnecting on the client side. With receiver disconnect we can check that all interfaces are deactivated. (This updated test still reproduces the issue on the branch without fix, so we are covered here)

Seems testing this discovered an existing issue that activate_count is not back at 0 when client disconnects. However I guess we don't have to fix it in this PR , can add TODO/FIXME note in the test

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3261,32 +3261,23 @@ class test_ucp_sockaddr_iface_activate : public test_ucp_sockaddr {
                ucs::is_rdmacm_netdev(ifa->ifa_name);
     }
 
-    ucp_worker_iface_t *get_test_iface(entity &e) const
+    bool is_interface_activated(entity &e) const
```
yosefe, line 3264: is_any_interface_activated

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3232,3 +3243,71 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_err_sender,
...
+        EXPECT_TRUE(is_interface_activated(sender()));
+        EXPECT_TRUE(is_interface_activated(receiver()));
+
+        one_sided_disconnect(sender());
+        one_sided_disconnect(receiver(), 0, receiver_ep_count);
+        /* Wait for EP disconnect completion which might happen after
+         * one_sided_disconnect */
+        wait_progress([&] {
+            return !is_interface_activated(receiver());
+        });
```
yosefe, line 3299: why need to wait for ! activated?
seems that after one_sided_disconnect the activation count should be already set to 0 because that function waits for disconnect to complete

test/gtest/ucp/test_ucp_sockaddr.cc
iyastreb: ok

test/gtest/ucp/test_ucp_sockaddr.cc
iyastreb: done

test/gtest/ucp/test_ucp_sockaddr.cc
iyastreb: Yes, this activation logic is complex and there are many case and combinations, that I discovered when reproducing this particular issue. I'm sure it deserves a separate task to validate that this feature works as expected. Which is out of scope of this issue. For now I add a TODO note

test/gtest/ucp/test_ucp_sockaddr.cc
iyastreb: Yes, it was needed with one sided disconnect from the client side, but with disconnect on the receiver side it's redundant

yosefe (approved):

brminich (approved):

PR #9814, opened 2024-04-11
UCP/WIREUP: Increase UCP_MAX_LANES to 64

ivankochin:
## What
Increase number of maximal lanes per EP to 64.

## Why ?
To allow collecting information about all lanes on systems with many transports/devices.

PR #9864, opened 2024-05-07
UCP/PROTOV1: Remove redundant fieds from request structure

ivankochin:
## What
Removes redundant fields from request structure.

## Why ?
To avoid increasing `ucp_request` structure size after merging https://github.com/openucx/ucx/pull/9814

## Notes

Performance was tested on osu_latency, osu_bw and osu_mbw_mr benchmarks

### DIFF
--- a/src/ucp/rndv/rndv.c
+++ b/src/ucp/rndv/rndv.c
@@ -500,7 +500,7 @@ static ucp_lane_index_t ucp_rndv_zcopy_get_lane(ucp_request_t *rndv_req,
         return UCP_NULL_LANE;
     }
 
-    lane_idx   = ucs_ffs64_safe(rndv_req->send.lanes_map_avail);
+    lane_idx   = rndv_req->send.multi_lane_idx;
     ucs_assert(lane_idx < UCP_MAX_LANES);
     rkey       = rndv_req->send.rndv.rkey;
 
@@ -515,10 +515,12 @@ static ucp_lane_index_t ucp_rndv_zcopy_get_lane(ucp_request_t *rndv_req,
 
 static void ucp_rndv_zcopy_next_lane(ucp_request_t *rndv_req)
 {
-    rndv_req->send.lanes_map_avail    &= rndv_req->send.lanes_map_avail - 1;
-    if (!rndv_req->send.lanes_map_avail) {
-        rndv_req->send.lanes_map_avail = rndv_req->send.rndv.zcopy.lanes_map_all;
-    }
+    ucp_lane_map_t lanes_map_all = rndv_req->send.rndv.zcopy.lanes_map_all;
+    ucp_lane_map_t lane_map;
+    lane_map = lanes_map_all & ~UCS_MASK(rndv_req->send.multi_lane_idx + 1);
+
+    rndv_req->send.multi_lane_idx = ucs_ffs32((lane_map > 0) ? lane_map :
+                                                               lanes_map_all);
 }
 
 static ucs_status_t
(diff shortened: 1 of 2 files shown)


src/ucp/rndv/rndv.c
```diff
@@ -515,10 +515,12 @@ static ucp_lane_index_t ucp_rndv_zcopy_get_lane(ucp_request_t *rndv_req,
 
 static void ucp_rndv_zcopy_next_lane(ucp_request_t *rndv_req)
 {
-    rndv_req->send.lanes_map_avail    &= rndv_req->send.lanes_map_avail - 1;
-    if (!rndv_req->send.lanes_map_avail) {
-        rndv_req->send.lanes_map_avail = rndv_req->send.rndv.zcopy.lanes_map_all;
-    }
+    ucp_lane_map_t lanes_map_all = rndv_req->send.rndv.zcopy.lanes_map_all;
+    ucp_lane_map_t lane_map;
```
tvegas1, line 519: merge with line below

tvegas1 (approved):

src/ucp/rndv/rndv.c
ivankochin: It will cross 80 characters line length limit then

src/ucp/rndv/rndv.c
```diff
@@ -529,6 +531,7 @@ ucp_rndv_progress_rma_zcopy_common(ucp_request_t *req, ucp_lane_index_t lane,
     ucp_ep_h ep             = req->send.ep;
     uct_ep_h uct_ep         = ucp_ep_get_lane(ep, lane);
     ucp_ep_config_t *config = ucp_ep_config(ep);
+    size_t lanes_count      = ucs_popcount(req->send.rndv.zcopy.lanes_map_all);
```
yosefe, line 534: popcount is not converted to CPU instruction when not passing --enable-optimizations flag
so we should avoid it in fast path
maybe we can keep it in the request, it's not so big anyway

ivankochin:
```c
union {
    struct {
        /* Actual lanes map */
        ucp_lane_map_t lanes_map_all;

        /* Actual lanes count */
        uint8_t        lanes_count;
    } zcopy;

    struct {
        /* Data start offset of this request */
        size_t offset;
    } rtr;
};
```
That is the part of `ucp_request` where `lanes_count` is stored. If size of `ucp_lane_map_t` is less than 8, everything is OK since `lanes_count` + `lanes_map_all` fields took no more than `offset`, so union has 8 bytes size.

But if `ucp_lane_map_t` has 8 bytes, `lanes_count` became the 9th byte, so the union size became 16 bytes .

The only possible option that I see is to move that 1-byte field to another place inside `ucp_request` sturcture, but all other fields inside `struct rndv {...}` have 8-byte size, so on any place it would create 7 byte padding.

Is that so significant to have that value prestored in that field instead of `popcount` on that place? I mean that RMA flow which is often used for big messages and that's only for protov1, WDYT?

yosefe: ok, though maybe worth converting `ucs_assert_always(lanes_count > 0);` to `ucs_assert(ucs_popcount(lanes_count) > 0);` and remove the local var lanes_count, so it will be calculated only when needed in release mode

ivankochin: Done.

tvegas1 (dismissed):

yosefe (approved):

src/ucp/rndv/rndv.c
```diff
@@ -1844,7 +1845,7 @@ UCS_PROFILE_FUNC(ucs_status_t, ucp_rndv_progress_rma_put_zcopy, (self),
     ucp_request_t *sreq = ucs_container_of(self, ucp_request_t, send.uct);
     uct_rkey_t uct_rkey;
 
-    ucs_assert_always(sreq->send.rndv.zcopy.lanes_count > 0);
+    ucs_assert_always(ucs_popcount(sreq->send.rndv.zcopy.lanes_map_all) > 0);
```
brminich, line 1848: why not changed to assert?

ivankochin, line 1848:
```suggestion
    ucs_assert(ucs_popcount(sreq->send.rndv.zcopy.lanes_map_all) > 0);
```

ivankochin: Good catch, removed!

yosefe (approved):

tvegas1 (approved):

PR #9870, opened 2024-05-12
CI/MAD: Run Docker as a user to avoid creating files as root 

Alexey-Rivkin:
## What
Regain ownership of files in the working directory after running custom "docker run" commands.

## Why ?
CI fails, as it's unable to clean up root files left on hosts by the Docker container.

Findings:
```
[swx-azure-svc@vulcan03 11:00:53 azure]$ find . -user root
./agent-02/AZP_WORKSPACE/3/s/ucx_perf_srv_LID_80635.log
./agent-02/AZP_WORKSPACE/3/s/ucx_perf_srv_GUID_80635.log
```

### DIFF
--- a/buildlib/tools/test_mad.sh
+++ b/buildlib/tools/test_mad.sh
@@ -23,23 +23,24 @@ build_ucx() {
 
 build_ucx_in_docker() {
     docker run --rm \
+        --user "$(id -u)":"$(id -g)" \
         --name ucx_build_"$BUILD_BUILDID" \
         -e BUILD_SOURCESDIRECTORY="$BUILD_SOURCESDIRECTORY" \
         -v "$PWD":"$PWD" -w "$PWD" \
         -v /hpc/local:/hpc/local \
         $IMAGE \
         bash -c "source ./buildlib/tools/test_mad.sh && build_ucx"
-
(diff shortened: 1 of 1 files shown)


buildlib/tools/test_mad.sh
```diff
@@ -48,6 +48,8 @@ docker_run_srv() {
         $IMAGE \
         bash -c "${PWD}/install/bin/ucx_perftest -K ${HCA} > \
             ${PWD}/ucx_perf_srv_${test_name}_${BUILD_BUILDID}.log 2>&1"
+
+    sudo chown -R swx-azure-svc:ecryptfs "$PWD"
```
yosefe, line 52: why docker creates files outside the container? maybe we can avoid it?

tvegas1: if it is needed to be able to later retrieve logs, it could be possible to either:
- use `docker ... bash -c "umask 0011; ${PWD}/install/bin/ucx_perftest ... "`
- run docker without root user like `docker --user $(id -u swx-azure-svc):$(id -g swx-azure-svc) ...`.

yosefe (approved):
pls squash

tvegas1 (approved):

dpressle (approved):

PR #9871, opened 2024-05-13
UCP/PROTO/TEST: Add protocol performance data strucutre

yosefe:
## Why
Unite common code for protocol performance manipulations
@ivankochin @brminich


src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,400 @@
...
+    char            name[UCP_PROTO_DESC_STR_MAX];
+
+    /* List of segments */
+    ucs_list_link_t segments;
+};
+
+/*
+ * Iterate over the segments of the protocol perf structure function.
+ */
+#define ucp_proto_perf_segment_foreach(_perf, _seg) \
```
tvegas1, line 50: swap params?

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,400 @@
...
+
+    status = ucp_proto_perf_segment_new(perf, seg_end + 1, seg->end, &new_seg);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    for (factor_id = 0; factor_id < UCP_PROTO_PERF_FACTOR_LAST; factor_id++) {
+        new_seg->factors[factor_id] = seg->factors[factor_id];
+    }
+    new_seg->node = ucp_proto_perf_node_dup(seg->node);
```
tvegas1, line 124: we might need to handle error path like using `goto fail` or there is risk to silently endup with incomplete perf segments? there are other instances.

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,400 @@
...
+#if ENABLE_ASSERT
+    ucp_proto_perf_segment_t *seg;
+    size_t min_start;
+
+    min_start = 0;
+    ucp_proto_perf_segment_foreach(perf, seg) {
+        ucs_assertv((seg->start >= min_start) && (seg->start <= seg->end),
+                    "perf=%p seg->start=%zu seg->end=%zu min_start=%zu", perf,
+                    seg->start, seg->end, min_start);
+        min_start = seg->end + 1;
```
tvegas1, line 75: do we have a case where `seg->end` is UINT64_MAX? if yes we might want to check that when min_start is lower than seg->end, we are the last segment?

src/ucp/proto/proto_perf.h
```diff
@@ -0,0 +1,160 @@
...
+#define UCP_PROTO_PERF_H_
+
+#include "proto.h"
+
+#include <ucs/datastruct/linear_func.h>
+#include <ucs/datastruct/string_buffer.h>
+#include <ucs/type/status.h>
+
+
+/* Protocol performance data streucture over multiple ranges */
```
tvegas1, line 17: typo

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,400 @@
...
+                                        NULL);
+        ucp_proto_perf_segment_add_func(perf, seg, UCP_PROTO_PERF_FACTOR_MULTI,
+                                        range->perf[UCP_PROTO_PERF_TYPE_MULTI],
+                                        NULL);
+        ucp_proto_perf_segment_add_func(perf, seg,
+                                        UCP_PROTO_PERF_FACTOR_LOCAL_CPU,
+                                        range->perf[UCP_PROTO_PERF_TYPE_CPU],
+                                        NULL);
+
+        range_start = range->max_length + 1;
```
tvegas1, line 237: just to double check: proto_caps->ranges is sorted by range

src/ucp/proto/proto_perf.c
yosefe: w.r.t perf_node, a failure to create it is not considered fatal (because perf tree is for debug); in case of failure it will remain NULL and ignored later on.

ivankochin:
I think it can be useful to have function that sums up two perf structures (maybe make it inplace)?

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,400 @@
...
+    /* Linear function representing value of each contributing factor */
+    ucs_linear_func_t     factors[UCP_PROTO_PERF_FACTOR_LAST];
+};
+
+/*
+ * Protocol performance structure.
+ */
+struct ucp_proto_perf {
+    /* Name of the protocol */
+    char            name[UCP_PROTO_DESC_STR_MAX];
```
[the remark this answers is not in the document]
yosefe: because we are adding perf nodes automatically to the segments in it, while in proto_caps we get a ready perf node

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,400 @@
...
+
+struct ucp_proto_perf_segment {
+    /* List element */
+    ucs_list_link_t       list;
+
+    /* Start value of this segment (inclusive) */
+    size_t                start;
+
+    /* End value of this segment (inclusive) */
+    size_t                end;
```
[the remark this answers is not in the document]
yosefe: yes, it's done in order to support holes, including starting from certain min_length and that for some ranges a protocol would not be found

src/ucp/proto/proto_perf.h
```diff
@@ -0,0 +1,160 @@
...
+ * @param [out] perf_p      Filled with the new performance data structure.
+ */
+ucs_status_t ucp_proto_perf_create(const char *name, ucp_proto_perf_t **perf_p);
+
+
+/**
+ * Destroy a performance data structure and cleanup associated memory.
+ * allocations.
+ * The referece counts of any perf_node objects passed to
+ * @ref ucp_proto_perf_add_func will be adjusted accordingly.
```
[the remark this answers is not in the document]
yosefe: the intention was to mention ucp_proto_perf_add_func - perf_node is passed to it

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -343,3 +344,218 @@ UCS_TEST_P(test_perf_node, replace_null)
...
+    UCS_STRING_BUFFER_ONSTACK(strb, 256);
+    ucp_proto_perf_dump(perf, &strb);
+    return os << ucs_string_buffer_cstr(&strb);
+}
+
+class test_proto_perf : public test_ucp_proto {
+protected:
+    ucp_proto_perf_segment_t *find_lb(size_t start) const
+    {
+        return ucp_proto_perf_find_segment_lb(m_perf, start);
```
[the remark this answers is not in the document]
yosefe: I found it useful when replacing proto_init->caps with perf struct: even if we cannot find a range at the given msg_length point, we still need to know the next msg_length where a range exists, to start calculating thresholds from it next time. if we have more cases when need "exact" match we can add API ucp_proto_perf_get_segment().

yosefe:
> I think it can be useful to have function that sums up two perf structures (maybe make it inplace)?

Probably yes, but since this PR is quite big IMO it's better to add it later, maybe even along with using it in rndv perf

src/ucp/proto/proto_perf.c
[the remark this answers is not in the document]
ivankochin: Are there at least theoretical cases for protocols with holes? 

I am asking since from my point of view, supporting holes can cause extra complication during summarization of two separate perf structures. I mean if we have two perf structures for different protocol stages and we need to sum up them, therefore they seem to need to conform each other (have same global start, end and holes). Maybe supporting global start and end would be enough? WDYT?

src/ucs/datastruct/list.h
```diff
@@ -114,13 +114,30 @@ static inline int ucs_list_is_empty(ucs_list_link_t *head)
     return head->next == head;
 }
 
+/**
+ * @return Whether the list contains only elem.
+ */
+static inline int
+ucs_list_is_first(ucs_list_link_t *head, ucs_list_link_t *elem)
```
tvegas1, line 121: minor: const for both params here and below

src/ucs/datastruct/list.h
```diff
@@ -114,13 +114,30 @@ static inline int ucs_list_is_empty(ucs_list_link_t *head)
...
+}
+
 /**
  * @return Whether the list contains only elem.
  */
 static inline int ucs_list_is_only(ucs_list_link_t *head,
                                    ucs_list_link_t *elem)
 {
-    return (head->next == elem) && (elem->next == head);
+    return ucs_list_is_first(head, elem) && ucs_list_is_last(head, elem);
```
tvegas1, line 140: could add const's line 137

src/ucp/proto/proto_perf.c
tvegas1: I see, I assume there is no ambiguity ontroduced in the dump for perf segments/nodes.

tvegas1 (commented):
typo in PR title and commit would need to be fixed while squashing

src/ucs/datastruct/list.h
yosefe: it would make it inconsistent with other list functions, so IMO better not add it

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,411 @@
...
+        goto err;
+    }
+
+    range_start = proto_caps->min_length;
+    ucs_carray_for_each(range, proto_caps->ranges, proto_caps->num_ranges) {
+        ucs_assertv(range->max_length >= range_start,
+                    "proto_caps=%p range[%zd].max_length=%zu range_start=%zu",
+                    proto_caps, range - proto_caps->ranges, range->max_length,
+                    range_start);
+        if (range->max_length < range_start) {
```
brminich, line 223: there is an assert above that it is not possible

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,411 @@
...
+            if (status != UCS_OK) {
+                return status;
+            }
+
+            ucs_list_insert_before(&seg->list, &new_seg->list);
+            seg = new_seg;
+        } else {
+            /* Split the first affected segment */
+            if (start > seg->start) {
+                ucs_assertv(start > 0, "perf=%p start=%zu", perf, start);
```
brminich, line 300: `start` and `seg->start` are both `size_t`, so `start` can never be 0 if it is greater than `seg->start`

brminich (approved):

tvegas1 (dismissed):

src/ucp/proto/proto_perf.h
```diff
@@ -0,0 +1,160 @@
...
+typedef enum {
+    UCP_PROTO_PERF_FACTOR_LOCAL_CPU,
+    UCP_PROTO_PERF_FACTOR_REMOTE_CPU,
+    UCP_PROTO_PERF_FACTOR_TL_SEND,
+    UCP_PROTO_PERF_FACTOR_TL_RECEIVE,
+    UCP_PROTO_PERF_FACTOR_LATENCY,
+    UCP_PROTO_PERF_FACTOR_SINGLE, /* For compatibility; to remove */
+    UCP_PROTO_PERF_FACTOR_MULTI,  /* For compatibility; to remove */
+    UCP_PROTO_PERF_FACTOR_LAST
+} ucp_proto_perf_factor_id_t;
```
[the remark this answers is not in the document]
ivankochin: Regarding "pre" and "post", we can trait CPU factors as sum of "pre" and "post" for MULTI perf case and build envelope basing on this sum. For SINGLE perf case "post" part happens in parallel with all other factors except CPU factor. Both these cases can be handled during SINGLE and/or MULTI perf final envelope calculation if they are separated.

So if we leave that as is, should we include `send_post_overhead` to `LOCAL_CPU` factor?

I am OK to leave it without separation, but as I know cuda copy has `send_post_overhead` equal to 1.8 usec which seems significant overhead, so it can cause performance degradation in some cases.

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -343,3 +344,218 @@ UCS_TEST_P(test_perf_node, replace_null)
...
+    caps.ranges[0].perf[UCP_PROTO_PERF_TYPE_SINGLE] = tl_send_func;
+    caps.ranges[0].perf[UCP_PROTO_PERF_TYPE_MULTI]  = UCS_LINEAR_FUNC_ZERO;
+    caps.ranges[0].perf[UCP_PROTO_PERF_TYPE_CPU]    = UCS_LINEAR_FUNC_ZERO;
+    caps.ranges[0].node                             = NULL;
+
+    caps.ranges[1].max_length                       = 2999;
+    caps.ranges[1].perf[UCP_PROTO_PERF_TYPE_SINGLE] = UCS_LINEAR_FUNC_ZERO;
+    caps.ranges[1].perf[UCP_PROTO_PERF_TYPE_MULTI]  = UCS_LINEAR_FUNC_ZERO;
+    caps.ranges[1].perf[UCP_PROTO_PERF_TYPE_CPU]    = local_cpu_func;
+    caps.ranges[1].node                             = NULL;
```
ivankochin, line 427: I understand that patch is huge already, but WDYTA adding perf nodes allocation to check how they splitted/allocated/deleted?

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -343,3 +344,218 @@ UCS_TEST_P(test_perf_node, replace_null)
...
+            zero_perf_factors &= ~UCS_BIT(factor.first);
+        }
+
+        unsigned factor_id;
+        ucs_for_each_bit(factor_id, zero_perf_factors) {
+            expect_perf(start, end,
+                        static_cast<ucp_proto_perf_factor_id_t>(factor_id),
+                        UCS_LINEAR_FUNC_ZERO);
+        }
+    }
```
ivankochin, line 401: Minor:

```suggestion

    void
    expect_perf(size_t start, size_t end,
                const std::unordered_map<int, ucs_linear_func_t> factors)
    {
        int factor_id = UCP_PROTO_PERF_FACTOR_LOCAL_CPU;
        for (; factor_id < UCP_PROTO_PERF_FACTOR_LAST; ++factor_id) {
            ucs_linear_func_t expected_perf = UCS_LINEAR_FUNC_ZERO;
            auto entry                      = factors.find(factor_id);
            if (entry != std::end(factors)) {
                expected_perf = entry->second;
            }
            expect_perf(start, end, (ucp_proto_perf_factor_id_t)factor_id,
                        expected_perf);
        }
    }

```

test/gtest/ucp/test_ucp_proto.cc
yosefe: it's tested to some extend in the intersection tests, IMO it's better to make more testing in a separate PR

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -435,6 +431,8 @@ UCS_TEST_P(test_proto_perf, from_caps)
     expect_perf(2000, 2999,
                 {{UCP_PROTO_PERF_FACTOR_LOCAL_CPU, local_cpu_func}});
     expect_empty_range(3000, SIZE_MAX);
+
+    ucp_proto_perf_destroy(m_perf);
```
ivankochin, line 435: Move it to `cleanup()` func for this test and remove `cleanup()` from `test_proto_perf_add_func`

tvegas1 (dismissed):

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -509,29 +513,30 @@ UCS_TEST_P(test_proto_perf_add_func, intersect_first)
 UCS_TEST_P(test_proto_perf_add_func, intersect_last)
 {
     /*
-     * 500   1000         1999     3000  3499   3999
-     *  |     |             |       |      |      |
-     *  |     +-- tl_send +-+       +- tl_recv ---+
+     * 500   1000         1999    3000  3499   3999
+     *  |     |             |      |      |      |
+     *  |     +- local_tl +-+      +- remote_tl --+
```
ivankochin, line 518: Minor:
```suggestion
     *  |     |             |      |      |       |
     *  |     +- local_tl +-+      +- remote_tl --+
```

src/ucp/proto/proto_perf.c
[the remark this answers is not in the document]
ivankochin: After the discussion it was decided to proceed with holes support.

yosefe:
@ivankochin ok to squash?

brminich (approved):

tvegas1 (dismissed):

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,406 @@
...
+                                ucp_proto_perf_factor_id_t factor_id,
+                                ucs_linear_func_t func,
+                                ucp_proto_perf_node_t *perf_node)
+{
+    ucs_linear_func_add_inplace(&seg->factors[factor_id], func);
+
+    if (seg->node == NULL) {
+        ucp_proto_perf_segment_create_node(perf, seg);
+    } else {
+        ucp_proto_perf_node_update_data(perf_node,
```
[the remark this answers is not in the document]
ivankochin: ?

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,406 @@
...
+            continue;
+        }
+
+        status = ucp_proto_perf_segment_new(perf, range_start,
+                                            range->max_length, &seg);
+        if (status != UCS_OK) {
+            goto err_destroy;
+        }
+
+        seg->node = range->node;
```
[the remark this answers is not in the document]
ivankochin: ?

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,406 @@
...
+                                     size_t end,
+                                     ucp_proto_perf_factor_id_t factor_id,
+                                     ucs_linear_func_t func,
+                                     ucp_proto_perf_node_t *perf_node)
+{
+    ucp_proto_perf_segment_t *seg, *new_seg;
+    ucs_status_t status;
+    size_t seg_end;
+
+    ucp_proto_perf_check(perf);
```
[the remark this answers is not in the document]
ivankochin: ?

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,406 @@
...
+
+    /*                   __________        _________________
+     * perf before:     |__________|      |_________________|
+     *                __________________
+     * range to add: |__________________|
+     *                ______________________________________
+     * perf after:   |__|__________|____|_|_____|__________|
```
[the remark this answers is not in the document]
ivankochin: ?

ivankochin:
> @ivankochin ok to squash?

There are still several unresolved comments that I wanted to discuss. I placed question marks under them.

ivankochin:
> @ivankochin ok to squash?

There are still several unresolved comments that I want to discuss. I placed question marks under them.

src/ucp/proto/proto_perf.c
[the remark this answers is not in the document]
yosefe: re 2. the "from_caps" method was done for temporary compatibility with existing code so IMO better keep original perf node name to keep same names in the tree structure

src/ucp/proto/proto_perf.c
```diff
@@ -0,0 +1,406 @@
...
+                                     seg->factors[factor_id]);
+    }
+}
+
+static void
+ucp_proto_perf_segment_add_func(ucp_proto_perf_t *perf,
+                                ucp_proto_perf_segment_t *seg,
+                                ucp_proto_perf_factor_id_t factor_id,
+                                ucs_linear_func_t func,
+                                ucp_proto_perf_node_t *perf_node)
```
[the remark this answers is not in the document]
yosefe: it's perf_node because it corresponds to the function "func", since in the function parameters there is no mention of child/parent node/funcs

yosefe:
> > @ivankochin ok to squash?
> 
> There are still several unresolved comments that I wanted to discuss. I placed question marks under them.

@ivankochin thanks, fixed

ivankochin (approved):

PR #9872, opened 2024-05-13
UCP/WIREUP: Fix AM BW multi-lanes selection

ivankochin:
## What
- Fixes lanes selection process for multi-lane AM BW case
- Extend testing for max lanes case

## Why ?
- Currently, UCX chooses `MAX_LANES - 1` number of lanes since `am_lane` is passed to `ucp_wireup_add_bw_lanes` as `excl_lane` parameter, so UCX shouldn't subtract extra one lane.
-  During `ucp_wireup_construct_lanes` execution UCX places all the `am_bw_lanes` to one position further to set `am_lane` to `am_bw_lanes[0]` later, but `am_lane` is already part of `am_bw_lanes`, so we should place it as usual, but doesn't include to sorted range.

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -1710,11 +1710,11 @@ ucp_wireup_add_am_bw_lanes(const ucp_wireup_select_params_t *select_params,
     bw_info.local_dev_bitmap  = UINT64_MAX;
     bw_info.remote_dev_bitmap = UINT64_MAX;
     bw_info.md_map            = 0;
-    bw_info.max_lanes         = context->config.ext.max_eager_lanes - 1;
+    bw_info.max_lanes         = context->config.ext.max_eager_lanes;
     /* rndv/am/zcopy proto should take max_rndv_lanes value into account */
     if (context->config.ext.proto_enable) {
         bw_info.max_lanes = ucs_max(bw_info.max_lanes,
-                                    context->config.ext.max_rndv_lanes - 1);
+                                    context->config.ext.max_rndv_lanes);
     }
 
     /* am_bw_lane[0] is am_lane, so don't re-select it here */
@@ -1728,6 +1728,7 @@ ucp_wireup_add_am_bw_lanes(const ucp_wireup_select_params_t *select_params,
             break;
         }
     }
+    ucs_assertv(am_lane != UCP_NULL_LANE, "am_lane=%u", am_lane);
 
     num_am_bw_lanes = ucp_wireup_add_bw_lanes(select_params, &bw_info,
                                               ucp_tl_bitmap_max, am_lane,
@@ -2384,9 +2385,8 @@ ucp_wireup_construct_lanes(const ucp_wireup_select_params_t *select_params,
             ucs_assert(key->am_lane == UCP_NULL_LANE);
             key->am_lane = lane;
         }
-        if ((select_ctx->lane_descs[lane].lane_types & UCS_BIT(UCP_LANE_TYPE_AM_BW)) &&
-            (lane < UCP_MAX_LANES - 1)) {
-            key->am_bw_lanes[lane + 1] = lane;
+        if ((select_ctx->lane_descs[lane].lane_types & UCS_BIT(UCP_LANE_TYPE_AM_BW))) {
+            key->am_bw_lanes[lane] = lane;
         }
         if (select_ctx->lane_descs[lane].lane_types & UCS_BIT(UCP_LANE_TYPE_RMA)) {
             key->rma_lanes[lane] = lane;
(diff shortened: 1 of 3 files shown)


src/ucp/wireup/select.c
```diff
@@ -2384,9 +2385,8 @@ ucp_wireup_construct_lanes(const ucp_wireup_select_params_t *select_params,
             ucs_assert(key->am_lane == UCP_NULL_LANE);
             key->am_lane = lane;
         }
-        if ((select_ctx->lane_descs[lane].lane_types & UCS_BIT(UCP_LANE_TYPE_AM_BW)) &&
-            (lane < UCP_MAX_LANES - 1)) {
-            key->am_bw_lanes[lane + 1] = lane;
+        if ((select_ctx->lane_descs[lane].lane_types & UCS_BIT(UCP_LANE_TYPE_AM_BW))) {
+            key->am_bw_lanes[lane] = lane;
```
brminich, line 2389: what if am lane is not selected also as `AM_BW` lane? it will not be set in this array then

ivankochin: With the new logic (last commit) AM lane is never selected as AM_BW since it passed as excluded lane.

src/ucp/wireup/select.c
```diff
@@ -1725,13 +1725,15 @@ ucp_wireup_add_am_bw_lanes(const ucp_wireup_select_params_t *select_params,
             /* do not continue searching since we found AM lane (and there is
              * only one AM lane) */
             am_lane = lane_desc_idx;
+            select_ctx->lane_descs[lane_desc_idx].path_index = 0;
```
brminich, line 1728: why needed?

src/ucp/wireup/select.c
brminich: why do you start with index 0 then if you overwrite it with pure AM lane below

ivankochin: If I understand the code correctly, it would never start with 0 since 0 would always be `am_lane` and `am_lane` cannot contain `UCP_LANE_TYPE_AM_BW` since it passed as `excl_lane` parameter during `add_bw_lanes()` call

src/ucp/wireup/select.c
ivankochin: In case when `am_lane` has undefined `path_index`, it would be merged with one of `am_bw_lanes`

brminich: ok, why do we need to fix that?
also this change breaks wire compatibility, need to add some protection to avoid that

ivankochin: In test case where we have `IB_NUM_PATHS=16` and `MAX_EAGER_LANES=16` without this change UCX first will select 1 lane for AM, then will select 15 lanes for AM_BW, but since AM lane has undefined path_index, it would be merged with one of selected AM_BW lanes, so the resulting number of lanes would be 15.

test/gtest/ucp/test_ucp_tag_xfer.cc
```diff
@@ -1232,41 +1237,45 @@ class multi_rail_max : public test_ucp_tag_xfer {
...
 
-    const uint32_t max_lanes = 16;
-};
-
-UCS_TEST_P(multi_rail_max, max_lanes, "IB_NUM_PATHS?=16", "TM_SW_RNDV=y",
-           "RNDV_THRESH=1", "MIN_RNDV_CHUNK_SIZE=1", "MULTI_PATH_RATIO=0.0001")
-{
-    if (is_proto_enabled()) {
-        UCS_TEST_SKIP_R("TM_SW_RNDV has no effect with proto v2");
+    void test_max_lanes(bool is_eager)
```
brminich, line 1240: you do not use this argument

test/gtest/ucp/test_ucp_tag_xfer.cc
```diff
@@ -1232,41 +1237,45 @@ class multi_rail_max : public test_ucp_tag_xfer {
...
-        /* Verify that each lane sent something, except the active message lane
-           that could be used only for control messages */
-        if (lane == num_lanes - 1) {
-            EXPECT_GT(sender_tx + receiver_tx, 0); // last lane sends the rest
-        } else if (lane != ucp_ep_get_am_lane(sender().ep())) {
-            EXPECT_GE(sender_tx + receiver_tx, chunk_size);
-        }
-    }
+UCS_TEST_P(multi_rail_max, max_lanes_rndv, "RNDV_THRESH=1",
+           "MIN_RNDV_CHUNK_SIZE=1", )
```
brminich, line 1272:
```suggestion
           "MIN_RNDV_CHUNK_SIZE=1")
```

src/ucp/wireup/select.c
brminich: i'm afraid this would affect other use-case, like selecting other type of lanes (e.g. amo or rma). With this code they ma not be merged with AM lane anymore, which is not good in terms of resources consumption

PR #9876, opened 2024-05-15
UCM: Print errno when mremap fails

brminich:
## What
Print errno when mremap fails

## Why
Better debugging experience


src/ucm/util/sys.c
```diff
@@ -191,7 +191,8 @@ void ucm_parse_proc_self_maps(ucm_proc_maps_cb_t cb, void *arg)
             buffer = ucm_orig_mremap(buffer, buffer_size, buffer_size * 2,
                                      MREMAP_MAYMOVE, NULL);
             if (buffer == MAP_FAILED) {
-                ucm_fatal("failed to allocate maps buffer(size=%zu)", buffer_size);
+                ucm_fatal("failed to allocate maps buffer(size=%zu): %m",
```
yosefe, line 194: maybe print full params:
"failed to reallocate buffer for reading proc/self/maps from %p/%zu to size %zu"

yosefe (approved):

tvegas1 (approved):

PR #9877, opened 2024-05-15
GTEST/COMMON: Cache CUDA device BAR1 available size

tvegas1:
## What
Cache the obtained BAR1 available size.

## Why ?
The `nvmlInit_v2()` seems to `fork()/wait()` which is causing issues when running under valgrind.

Internal link: [3886807](https://redmine.mellanox.com/issues/3886807)

## How ?
Only request available size once on a given run.

### DIFF
--- a/test/gtest/common/mem_buffer.cc
+++ b/test/gtest/common/mem_buffer.cc
@@ -196,38 +196,29 @@ void mem_buffer::set_device_context()
     device_set = true;
 }
 
-size_t mem_buffer::get_bar1_free_size()
-{
-    /* All gtest CUDA tests explicitly assume that all memory allocations are
-     * done on the device 0. The same assumption is followed here. */
-    size_t available_size = SIZE_MAX;
+size_t mem_buffer::m_bar1_free_size = SIZE_MAX;
 
+void mem_buffer::get_bar1_free_size_nvml()
+{
 #if HAVE_CUDA
     nvmlDevice_t device;
     nvmlBAR1Memory_t bar1mem;
 
     if (NVML_CALL(nvmlInit_v2()) != UCS_OK) {
-        return available_size;
+        return;
     }
 
-    if (NVML_CALL(nvmlDeviceGetHandleByIndex(0, &device)) != UCS_OK) {
-        /* For whatever reason we cannot open device handle.
-         * As a result let's assume there is no limit on the size
-         * and in the worse case scenario gtest will fail in runtime */
-        return available_size;
+    /* Assume no size limit in case of failure, in the worst case scenario
+     * gtest will fail in runtime */
+    if (NVML_CALL(nvmlDeviceGetHandleByIndex(0, &device)) == UCS_OK) {
+        if (NVML_CALL(nvmlDeviceGetBAR1MemoryInfo(device, &bar1mem)) ==
+            UCS_OK) {
+            mem_buffer::m_bar1_free_size = (size_t)bar1mem.bar1Free;
+        }
     }
-
-    if (NVML_CALL(nvmlDeviceGetBAR1MemoryInfo(device, &bar1mem)) != UCS_OK) {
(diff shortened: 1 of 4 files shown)


test/gtest/common/mem_buffer.cc
```diff
@@ -200,31 +200,39 @@ size_t mem_buffer::get_bar1_free_size()
 {
     /* All gtest CUDA tests explicitly assume that all memory allocations are
      * done on the device 0. The same assumption is followed here. */
-    size_t available_size = SIZE_MAX;
+    static size_t available_size = SIZE_MAX;
 
 #if HAVE_CUDA
-    nvmlDevice_t device;
-    nvmlBAR1Memory_t bar1mem;
+    static bool cached_free_size = false;
```
yosefe, line 206: 1. cached_free_size -> cached_available_size
2. maybe extract the code that access NVML to separate helper function?

yosefe (commented):
cc @shamisp

tvegas1:
ASAN failures look very much related but they are not as they are also found on CI failures for #9870:
```
2024-05-15T18:12:48.0241690Z     #1 0x7fcbb242d595  (/usr/lib64/libnvidia-ml.so.1+0x11c595)
2024-05-15T18:12:48.0242359Z     #2 0x7fcbb232bceb in nvmlInitWithFlags (/usr/lib64/libnvidia-ml.so.1+0x1aceb)
2024-05-15T18:12:48.0242914Z     #3 0x7fcbb525ae74 in uct_cuda_ipc_get_device_nvlinks /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_iface.c:138
```

yosefe:
> ASAN failures look very much related but they are not as they are also found on CI failures for #9870:
> 
> ```
> 2024-05-15T18:12:48.0241690Z     #1 0x7fcbb242d595  (/usr/lib64/libnvidia-ml.so.1+0x11c595)
> 2024-05-15T18:12:48.0242359Z     #2 0x7fcbb232bceb in nvmlInitWithFlags (/usr/lib64/libnvidia-ml.so.1+0x1aceb)
> 2024-05-15T18:12:48.0242914Z     #3 0x7fcbb525ae74 in uct_cuda_ipc_get_device_nvlinks /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_iface.c:138
> ```

What if we get the BAR1 size during startup, and not on-demand when running tests?

test/gtest/common/mem_buffer.cc
tvegas1: fixed

yosefe:
> > What if we get the BAR1 size during startup, and not on-demand when running tests?
> 
> If I get it right, you are suggesting to implement getting BAR1 size at test startup and not on-demand? It is possible but since we don't know when/if the data will be needed, I though keeping it like that.
> 
> Also putting it in some test init, for example, would still call it many times, whereas here with the static, it's only ever called once per gtest run at most.

I mean to do it before tests are running. Maybe this way we can avoid calling nvmlInit() twice (without destroy) and avoid the ASAN error

tvegas1:
addressed, but since other failure comes from `uct_cuda_ipc_get_device_nvlinks()`, not from get bar1 test function, there is possibility that leak will persist.

test/gtest/common/mem_buffer.cc
```diff
@@ -196,38 +196,31 @@ void mem_buffer::set_device_context()
     device_set = true;
 }
 
-size_t mem_buffer::get_bar1_free_size()
+size_t mem_buffer::m_bar1_free_size;
```
yosefe, line 199: shall we init to SIZE_MAX instead of setting it in get_bar1_free_size_nvml?

yosefe (approved):

shamisp:
Just calling it once in gtest should be okay, unless you have some other cuda app running in parallel.

PR #9886, opened 2024-05-19
AZP: Re-enable running tests on BlueField

yosefe:
## Why
Failures should be fixed after FW upgrade


yosefe:
@gleon99 can you pls review?

PR #9889, opened 2024-05-21
UCP/PROTO/PERF: Add function to aggregate multiple performances

yosefe:
## Why
Support combining multiple protocol parts

### DIFF
--- a/src/ucp/proto/proto_perf.c
+++ b/src/ucp/proto/proto_perf.c
@@ -12,7 +12,9 @@
 #include "proto_debug.h"
 
 #include <ucs/datastruct/list.h>
+#include <ucs/debug/log.h>
 #include <ucs/debug/memtrack_int.h>
+#include <ucs/sys/compiler.h>
 #include <ucs/sys/string.h>
 
 
@@ -61,10 +63,10 @@ static const char *ucp_proto_perf_factor_names[] = {
     [UCP_PROTO_PERF_FACTOR_LAST]       = NULL
 };
 
-static void ucp_proto_perf_check(ucp_proto_perf_t *perf)
+static void ucp_proto_perf_check(const ucp_proto_perf_t *perf)
 {
 #if ENABLE_ASSERT
-    ucp_proto_perf_segment_t *seg;
+    const ucp_proto_perf_segment_t *seg;
     size_t min_start;
 
     min_start = 0;
@@ -135,41 +137,28 @@ static ucs_status_t ucp_proto_perf_segment_split(const ucp_proto_perf_t *perf,
     return UCS_OK;
 }
 
-static void ucp_proto_perf_segment_create_node(ucp_proto_perf_t *perf,
-                                               ucp_proto_perf_segment_t *seg)
+static void ucp_proto_perf_segment_add_funcs(
+        ucp_proto_perf_t *perf, ucp_proto_perf_segment_t *seg,
+        const ucs_linear_func_t funcs[UCP_PROTO_PERF_FACTOR_LAST],
+        uint64_t factors_bitmap, ucp_proto_perf_node_t *perf_node)
 {
-    ucp_proto_perf_factor_id_t factor_id;
+    unsigned factor_id;
 
-    ucs_assert(seg->node == NULL);
-
-    seg->node = ucp_proto_perf_node_new_data(perf->name, "");
+    /* Create a performance node for this segment if it does not exist yet */
     if (seg->node == NULL) {
-        return;
+        seg->node = ucp_proto_perf_node_new_data(perf->name, "");
     }
 
-    for (factor_id = 0; factor_id < UCP_PROTO_PERF_FACTOR_LAST; factor_id++) {
-        ucp_proto_perf_node_add_data(seg->node,
-                                     ucp_proto_perf_factor_names[factor_id],
-                                     seg->factors[factor_id]);
-    }
-}
-
-static void
-ucp_proto_perf_segment_add_func(ucp_proto_perf_t *perf,
-                                ucp_proto_perf_segment_t *seg,
-                                ucp_proto_perf_factor_id_t factor_id,
-                                ucs_linear_func_t func,
-                                ucp_proto_perf_node_t *perf_node)
-{
-    ucs_linear_func_add_inplace(&seg->factors[factor_id], func);
-
-    if (seg->node == NULL) {
-        ucp_proto_perf_segment_create_node(perf, seg);
-    } else {
+    /* Add the functions to the segment and the performance node */
+    ucs_for_each_bit(factor_id, factors_bitmap) {
+        ucs_assert(factor_id < UCP_PROTO_PERF_FACTOR_LAST); /* For Coverity */
+        ucs_linear_func_add_inplace(&seg->factors[factor_id], funcs[factor_id]);
         ucp_proto_perf_node_update_data(seg->node,
                                         ucp_proto_perf_factor_names[factor_id],
                                         seg->factors[factor_id]);
     }
+
+    /* Add the child performance node to the segment performance node */
     ucp_proto_perf_node_add_child(seg->node, perf_node);
 }
 
@@ -203,6 +192,7 @@ ucs_status_t ucp_proto_perf_from_caps(const char *name,
                                       const ucp_proto_caps_t *proto_caps,
                                       ucp_proto_perf_t **perf_p)
 {
+    ucs_linear_func_t funcs[UCP_PROTO_PERF_FACTOR_LAST];
     const ucp_proto_perf_range_t *range;
     ucp_proto_perf_segment_t *seg;
     ucp_proto_perf_t *perf;
@@ -231,16 +221,18 @@ ucs_status_t ucp_proto_perf_from_caps(const char *name,
         ucp_proto_perf_node_ref(seg->node);
         ucs_list_add_tail(&perf->segments, &seg->list);
 
-        ucp_proto_perf_segment_add_func(perf, seg, UCP_PROTO_PERF_FACTOR_SINGLE,
-                                        range->perf[UCP_PROTO_PERF_TYPE_SINGLE],
-                                        NULL);
-        ucp_proto_perf_segment_add_func(perf, seg, UCP_PROTO_PERF_FACTOR_MULTI,
-                                        range->perf[UCP_PROTO_PERF_TYPE_MULTI],
-                                        NULL);
-        ucp_proto_perf_segment_add_func(perf, seg,
-                                        UCP_PROTO_PERF_FACTOR_LOCAL_CPU,
-                                        range->perf[UCP_PROTO_PERF_TYPE_CPU],
-                                        NULL);
+        funcs[UCP_PROTO_PERF_FACTOR_LOCAL_CPU] =
+                range->perf[UCP_PROTO_PERF_TYPE_CPU];
+        funcs[UCP_PROTO_PERF_FACTOR_SINGLE] =
+                range->perf[UCP_PROTO_PERF_TYPE_SINGLE];
+        funcs[UCP_PROTO_PERF_FACTOR_MULTI] =
+                range->perf[UCP_PROTO_PERF_TYPE_MULTI];
+        ucp_proto_perf_segment_add_funcs(
+                perf, seg, funcs,
+                UCS_BIT(UCP_PROTO_PERF_FACTOR_LOCAL_CPU) |
+                UCS_BIT(UCP_PROTO_PERF_FACTOR_SINGLE) |
+                UCS_BIT(UCP_PROTO_PERF_FACTOR_MULTI),
+                NULL);
 
         range_start = range->max_length + 1;
     }
@@ -255,11 +247,10 @@ err:
     return status;
 }
 
-ucs_status_t ucp_proto_perf_add_func(ucp_proto_perf_t *perf, size_t start,
-                                     size_t end,
-                                     ucp_proto_perf_factor_id_t factor_id,
-                                     ucs_linear_func_t func,
-                                     ucp_proto_perf_node_t *perf_node)
+ucs_status_t ucp_proto_perf_add_funcs(
+        ucp_proto_perf_t *perf, size_t start, size_t end,
+        const ucs_linear_func_t funcs[UCP_PROTO_PERF_FACTOR_LAST],
+        uint64_t factors_bitmap, ucp_proto_perf_node_t *perf_node)
 {
     ucp_proto_perf_segment_t *seg, *new_seg;
     ucs_status_t status;
@@ -317,7 +308,8 @@ ucs_status_t ucp_proto_perf_add_func(ucp_proto_perf_t *perf, size_t start,
                     seg->start);
         ucs_assertv(end >= seg->end, "end=%zu seg->end=%zu", end, seg->end);
 
-        ucp_proto_perf_segment_add_func(perf, seg, factor_id, func, perf_node);
+        ucp_proto_perf_segment_add_funcs(perf, seg, funcs, factors_bitmap,
+                                         perf_node);
         start = seg->end + 1;
         seg   = ucs_list_next(&seg->list, ucp_proto_perf_segment_t, list);
     }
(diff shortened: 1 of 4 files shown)


yosefe:
@ivankochin @tvegas1 can you pls take a look?

src/ucp/proto/proto_perf.c
```diff
@@ -340,6 +334,96 @@ ucs_status_t ucp_proto_perf_add_func(ucp_proto_perf_t *perf, size_t start,
...
+        status = ucp_proto_perf_segment_new(perf, start, end, &new_seg);
+        if (status != UCS_OK) {
+            goto err_destroy;
+        }
+
+        ucs_list_add_tail(&perf->segments, &new_seg->list);
+        for (i = 0; i < num_elems; ++i) {
+            seg = ucs_container_of(pos[i], ucp_proto_perf_segment_t, list);
+            ucp_proto_perf_segment_add_funcs(
+                    perf, new_seg, UCS_MASK(UCP_PROTO_PERF_FACTOR_LAST),
```
brminich, line 410: minor: we could have a define for this mask, like `UCP_PROTO_PERF_FACTOR_ALL`

src/ucp/proto/proto_perf.c
```diff
@@ -231,16 +224,16 @@ ucs_status_t ucp_proto_perf_from_caps(const char *name,
...
-                                        range->perf[UCP_PROTO_PERF_TYPE_SINGLE],
-                                        NULL);
-        ucp_proto_perf_segment_add_func(perf, seg, UCP_PROTO_PERF_FACTOR_MULTI,
-                                        range->perf[UCP_PROTO_PERF_TYPE_MULTI],
-                                        NULL);
-        ucp_proto_perf_segment_add_func(perf, seg,
-                                        UCP_PROTO_PERF_FACTOR_LOCAL_CPU,
-                                        range->perf[UCP_PROTO_PERF_TYPE_CPU],
-                                        NULL);
+        /* funcs array should be according to the order of factor_id bits */
```
tvegas1, line 227: to remove that constraint, what about instead simply pass `funcs` being defined as
```
ucs_linear_func_t funcs[UCP_PROTO_PERF_FACTOR_LAST] = {
  [UCP_PROTO_PERF_TYPE_CPU] = range->perf[UCP_PROTO_PERF_TYPE_CPU],
...
};
```

and either only use non-null pointers or keep the bitfield?

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -563,4 +609,72 @@ UCS_TEST_P(test_proto_perf_add_func, intersect_middle)
...
+    UCS_TEST_MESSAGE << m_perf.get();
+
+    expect_empty_range(0, 999);
+    expect_perf(1000, 1999,
+                {{UCP_PROTO_PERF_FACTOR_LOCAL_CPU, local_cpu_func},
+                 {UCP_PROTO_PERF_FACTOR_LOCAL_TL, local_tl_func}});
+    expect_empty_range(2000, SIZE_MAX);
+}
+
+UCS_TEST_P(test_proto_perf, agg3)
```
tvegas1, line 641: do we have a case with multiple intersections being output with ranges overlaps? maybe add other ranges to generate multiple intersections.

src/ucp/proto/proto_perf.c
```diff
@@ -340,6 +334,96 @@ ucs_status_t ucp_proto_perf_add_func(ucp_proto_perf_t *perf, size_t start,
...
+        goto out;
+    }
+
+    pos = ucs_alloca(sizeof(*pos) * num_elems);
+    for (i = 0; i < num_elems; ++i) {
+        pos[i] = perf_elems[i]->segments.next;
+    }
+
+    start = 0;
+    for (;;) {
```
tvegas1, line 366: i think we can probably streamline with less nesting like:
```C
i = 0, start = 0, end = INT_MAX;

while (!list_empty(pos[i])) {
  seg = container_of(pos[i]);

  if (seg->end < start) {
    pos[i] = list_next(pos[i]);
    continue;
  }

  if (start < seg->start) {
    start = seg->start;
    end   = seg->end;
    i     = 0;
    continue;
  }

  end = min(seg->end, end);
  if (++i == num_elems) {
    < Create intersection segment >

    start = end + 1;
    end   = INT_MAX;
  }
}

```

src/ucp/proto/proto_perf.c
```diff
@@ -135,41 +136,32 @@ static ucs_status_t ucp_proto_perf_segment_split(const ucp_proto_perf_t *perf,
...
-{
-    ucs_linear_func_add_inplace(&seg->factors[factor_id], func);
-
-    if (seg->node == NULL) {
-        ucp_proto_perf_segment_create_node(perf, seg);
-    } else {
+    /* Add the functions to the segment and the performance node */
+    func = funcs;
+    ucs_for_each_bit(factor_id, factors_bitmap) {
+        ucs_assert(factor_id < UCP_PROTO_PERF_FACTOR_LAST);
```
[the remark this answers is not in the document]
yosefe: not sure since this assert is mostly for fixing Coverity false positive. adding comment.

src/ucp/proto/proto_perf.c
yosefe: it's used only once so I'm not sure it's worth it

src/ucp/proto/proto_perf.c
yosefe: then if the user wants to define few funcs, they'd still have to allocate an array of the maximal length.

src/ucp/proto/proto_perf.c
ivankochin: I would also like to propose the following solution:

```c
    curr            = 0;
    segments[curr]  = ucs_list_head(&perf_elems[curr]->segments,
                                    ucp_proto_perf_segment_t, list);
    while (&segments[curr]->list != &perf_elems[curr]->segments) {
        end = segments[curr]->end;
        for (i = 0; i < num_elems; ++i) {
            if (i == curr) {
                continue;
            }

            segments[i] = ucp_proto_perf_find_segment_lb(perf_elems[i],
                                                         segments[curr]->start);
            if (segments[i] == NULL) {
                goto out;
            }

            if (segments[i]->start > segments[curr]->start) {
                end  = seg->end;
                curr = i;
                i    = 0;
            }

            end = ucs_min(end, seg->end);
        }

        // Create [segments[curr]->start, end] intersection segment

        segments[curr] = ucs_list_next(segments[curr]->list,
                                       ucp_proto_perf_segment_t, list);
    }
```

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -409,13 +464,25 @@ class test_proto_perf : public ucp_test {
         }
     }
 
+private:
+    static perf_ptr_t make_perf(ucp_proto_perf_t *perf)
+    {
+        return perf_ptr_t(perf, ucp_proto_perf_destroy);
+    }
```
ivankochin, line 471: Minor:
```suggestion
    static perf_ptr_t make_perf_shared_ptr(ucp_proto_perf_t *perf)
    {
        return {perf, ucp_proto_perf_destroy};
    }
```

1. Extend the name since the function doesn't make any `perf` but just creates shared ptr on top of it (`make_perf_ptr` or `make_perf_shared` seems also good options for me)

src/ucp/proto/proto_perf.c
ivankochin: But each segment already contains all the funcs and even if it would be used explicitly (not during aggregation of two perf structures) that funcs will be probably allocated on stack (and even optimized-away by compiler) so is that a significant drawback?

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -363,23 +370,69 @@ class test_proto_perf : public ucp_test {
...
+
+    void add_func(perf_ptr_t perf, size_t start, size_t end,
+                  ucp_proto_perf_factor_id_t factor_id, ucs_linear_func_t func)
+    {
+        ucs_linear_func_t funcs[1] = {func};
+        ucs_status_t status = ucp_proto_perf_add_funcs(perf.get(), start, end,
+                                                       UCS_BIT(factor_id),
+                                                       funcs, NULL);
+        ASSERT_UCS_OK(status);
```
ivankochin, line 420: Is that possible to make it this way?
```suggestion
        ASSERT_UCS_OK(ucp_proto_perf_add_funcs(
                perf.get(), start, end, UCS_BIT(factor_id), &func, NULL));
```

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -363,23 +370,69 @@ class test_proto_perf : public ucp_test {
...
+        return make_perf(perf);
+    }
+
+    void add_func(size_t start, size_t end,
+                  ucp_proto_perf_factor_id_t factor_id, ucs_linear_func_t func)
+    {
+        add_func(m_perf, start, end, factor_id, func);
+    }
+
+    void add_func(perf_ptr_t perf, size_t start, size_t end,
```
ivankochin, line 413: I think it worth adding tests that adds several funcs by one call of `ucp_proto_perf_add_funcs`

src/ucp/proto/proto_perf.c
yosefe: IMO the drawback is to initialize an array with NULL pointers / "empty" values. I guess we can use a bitmap to say which fields are valid.

test/gtest/ucp/test_ucp_proto.cc
yosefe: added random tests

src/ucp/proto/proto_perf.c
yosefe: fixed according to @tvegas1 suggestion, since @ivankochin suggestion has more nested loops (ucp_proto_perf_find_segment_lb searches from the start of each perf_elem)

src/ucp/proto/proto_perf.c
yosefe: fixed

yosefe:
@tvegas1 @ivankochin can you pls re-review?

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -563,4 +605,224 @@ UCS_TEST_P(test_proto_perf_add_func, intersect_middle)
...
+    for (auto point : sample_points) {
+        auto expected_result = expected_aggregate_result(perfs, point);
+        if (expected_result) {
+            expect_perf(point, point, *expected_result.get());
+        } else {
+            expect_empty_range(point, point);
+        }
+    }
+
+    if (testing::Test::HasFailure()) {
```
tvegas1, line 811: maybe dump start/end list in case of failure to ease analysis?

tvegas1 (commented):
one minor comment

test/gtest/ucp/test_ucp_proto.cc
yosefe: what is the start/end list?

brminich (dismissed):

test/gtest/ucp/test_ucp_proto.cc
tvegas1: each `ucp_proto_perf_segment_t`'s `start`/`end` for each element

src/ucp/proto/proto_perf.c
ivankochin: I am OK with @tvegas1 solution, but it worth noticing that the problem with `ucp_proto_perf_find_segment_lb` can be solved by passing start pointer to this function.

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -563,4 +605,224 @@ UCS_TEST_P(test_proto_perf_add_func, intersect_middle)
...
+                if (agg.find(factor_id) == agg.end()) {
+                    agg[factor_id] = func;
+                } else {
+                    ucs_linear_func_add_inplace(&agg[factor_id], func);
+                }
+            }
+        }
+    }
+
+    return std::unique_ptr<perf_factors_map_t>(new perf_factors_map_t(agg));
```
ivankochin, line 781: Minor:
```suggestion
    return {new perf_factors_map_t(agg)};
```

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -563,4 +605,224 @@ UCS_TEST_P(test_proto_perf_add_func, intersect_middle)
...
+
+test_proto_perf::perf_ptr_t
+test_proto_perf_aggregate::generate_random_perf(unsigned max_segments)
+{
+    perf_ptr_t perf = create();
+    size_t start    = 0;
+    auto points     = generate_random_sequence(max_segments);
+    for (auto point : points) {
+        if (ucs::rand_range(4) > 0) {
+            add_func(perf, start, point, UCP_PROTO_PERF_FACTOR_LOCAL_TL,
```
ivankochin, line 722: What do you think about adding functions to several different factors? Seems like `expected_aggregate_result` already support checking such combinations.

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -563,4 +605,224 @@ UCS_TEST_P(test_proto_perf_add_func, intersect_middle)
...
+        ucp_proto_perf_segment_t *seg = find_lb(perf, point);
+        if ((seg == NULL) || (point < ucp_proto_perf_segment_start(seg))) {
+            return nullptr; // Point not found in one of the perf objects
+        }
+
+        for (int factor_id = 0; factor_id < UCP_PROTO_PERF_FACTOR_LAST;
+             ++factor_id) {
+            auto func = ucp_proto_perf_segment_func(
+                    seg, (ucp_proto_perf_factor_id_t)factor_id);
+            if (!ucs_linear_func_is_zero(func, UCP_PROTO_PERF_EPSILON)) {
```
ivankochin, line 771: Can we remove this check?

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -563,4 +605,224 @@ UCS_TEST_P(test_proto_perf_add_func, intersect_middle)
...
+{
+    perf_ptr_t perf = create();
+    size_t start    = 0;
+    auto points     = generate_random_sequence(max_segments);
+    for (auto point : points) {
+        if (ucs::rand_range(4) > 0) {
+            add_func(perf, start, point, UCP_PROTO_PERF_FACTOR_LOCAL_TL,
+                     local_tl_func);
+        }
+        start = point + 1;
```
ivankochin, line 725: 1. If I understand correctly this test doesn't create ranges with holes between segments. Can we randomly add little holes between segments (they shouldn't be too large to not reduce probability of ranges intersection)?
2. Seems like `generate_random_sequence()` can return same neighboring points if `rand_range` returns 0, so after setting start to `point + 1` test will try to add func where start > end

test/gtest/ucp/test_ucp_proto.cc
yosefe: it's already printed when dumping the perf structure

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -348,75 +349,137 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_perf_node, all, "all")
...
+    void
+    expect_perf(size_t start, size_t end, const perf_factors_map_t &factors)
     {
-        ucp_proto_perf_segment_t *seg = find_lb(start);
-        ASSERT_NE(nullptr, seg);
+        const ucp_proto_perf_segment_t *seg = find_lb(m_perf, start);
+        ASSERT_NE(nullptr, seg) << "start=" << start;
+
         EXPECT_GE(start, ucp_proto_perf_segment_start(seg));
         EXPECT_LE(end, ucp_proto_perf_segment_end(seg));
```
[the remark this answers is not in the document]
yosefe: when we don't care about how segments are split (during aggregation test), we test specific point and not a rage, so this scenario cannot happen

test/gtest/ucp/test_ucp_proto.cc
yosefe: it's used to simplify the output in case of getting invalid result

test/gtest/ucp/test_ucp_proto.cc
yosefe: 1. there are holes: segment is added only if rand(4)>0. so there is 25% probability of a hold
2. added +1 to generate_random_sequence

test/gtest/ucp/test_ucp_proto.cc
yosefe: sounds good

test/gtest/ucp/test_ucp_proto.cc
yosefe: it's not possible since constructor is explicit

tvegas1 (dismissed):

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -348,75 +349,137 @@ UCP_INSTANTIATE_TEST_CASE_TLS(test_perf_node, all, "all")
...
+    static const ucs_linear_func_t local_tl_func;
+    static const ucs_linear_func_t remote_tl_func;
+    static const ucs_linear_func_t local_cpu_func;
+    perf_ptr_t m_perf;
 };
 
-UCS_TEST_P(test_proto_perf, from_caps)
+const ucs_linear_func_t test_proto_perf::local_tl_func  = {10, 1};
+const ucs_linear_func_t test_proto_perf::remote_tl_func = {20, 2};
+const ucs_linear_func_t test_proto_perf::local_cpu_func = {30, 3};
```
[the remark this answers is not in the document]
yosefe: it leads to unresolved symbol errors

Issue #9888, opened 2024-05-20
UCX fails to build with Intel compilers in optimized mode


bosilca:
Recent Intel compilers (I checked starting from 2023.2) automatically enable fast-math mode for optimized builds. This mode allows optimizations for floating-point arithmetic that assume that arguments and results are not NaNs or +-Infs. While most compilers will simply translate isnan and isinf to always false, recent Intel compilers took a different approach resulting in the following errors while building UCX in optimized mode.

```
.../ucx-master/src/ucs/datastruct/linear_func.h:171:9: error: explicit comparison with NaN in fast floating point mode [-Werror,-Wtautological-constant-compare]
    if (isnan(x) || isinf(x)) {
        ^~~~~~~~
.../ucx-master/src/ucs/datastruct/linear_func.h:171:21: error: explicit comparison with infinity in fast floating point mode [-Werror,-Wtautological-constant-compare]
    if (isnan(x) || isinf(x)) {
```
 
A quick hack is to force Intel compilers to switch in precise floating point mode for the incriminating function (`ucs_linear_func_compose`). 

```diff
diff --git a/src/ucs/datastruct/linear_func.h b/src/ucs/datastruct/linear_func.h
index 31e168df1..0808f775e 100644
--- a/src/ucs/datastruct/linear_func.h
+++ b/src/ucs/datastruct/linear_func.h
@@ -157,6 +157,7 @@ ucs_linear_func_compose(ucs_linear_func_t outer, ucs_linear_func_t inner)
  *         double value.
  *
  */
+#pragma float_control(precise,on)
 static inline ucs_status_t
 ucs_linear_func_intersect(ucs_linear_func_t func1, ucs_linear_func_t func2,
                           double *x_intersect)
```

bosilca:
Unfortunately, the proposed "solution" only works for the Intel compilers, all other compilers will choke with an unknown-pragma error. I fiddled a little around and it seems the culprit is the extremely restrictive -Werror flag added by configure. Without it I was able to successfully compile in all cases.

tvegas1:
Can you please try to partially disable fast-math optimization with `-fno-finite-math-only`? Else we could wrap isnan()/isinf() with `__FINITE_MATH_ONLY__` being false, but seems that ICX does not use that define, do you find anything related that would be set by ICX?

bosilca:
Adding `CFLAGS=-fno-finite-math-only` solves the problem.

Few additional issues:
[...]

PR #9893, opened 2024-05-22
BUILD/CONFIG: Build with Intel Compiler

tvegas1:
## What
Build UCX with Intel ICC/ICX compilers.

## Why ?
Basic Intel compiler support.

## How ?
Relax needed options, for Intel compiler only.

### Tested
Build tested only with `configure-devel` and `configure-release`.
```
icc -V
Intel(R) C Intel(R) 64 Compiler for applications running on Intel(R) 64, Version 17.0.4.196 Build 20170411
icx -V
Intel(R) C Compiler for applications running on Intel(R) 64, Version 2021.1 NextGen Beta Build 20200304
```
```
./autogen.sh; CXX=icpc CC=icc ./contrib/configure-devel
./autogen.sh; CXX=icpx CC=icx ./contrib/configure-devel
```

### DIFF
--- a/buildlib/tools/builds.sh
+++ b/buildlib/tools/builds.sh
@@ -110,23 +110,32 @@ build_release_pkg() {
 	cd -
 }
 
+build_icc_check() {
+	cc=$1
+	cxx=$2
+
+	if $cc -v
+	then
+		echo "==== Build with Intel compiler $cc ===="
+		${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst CC=$cc CXX=$cxx
+		$MAKEP
+		make_clean distclean
+	else
+		azure_log_warning "Not building with Intel compiler $cc"
+	fi
+}
+
 #
 # Build with Intel compiler
 #
 build_icc() {
-	if az_module_load $INTEL_MODULE && icc -v
+	if az_module_load $INTEL_MODULE
 	then
-		echo "==== Build with Intel compiler ===="
-		${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst CC=icc CXX=icpc
-		$MAKEP
-		make_clean distclean
-
-		echo "==== Build with Intel compiler (clang) ===="
-		${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst CC=clang CXX=clang++
-		$MAKEP
-		make_clean distclean
+		build_icc_check icc icpc
+		build_icc_check icx icpx
+		build_icc_check clang clang++
 	else
-		azure_log_warning "Not building with Intel compiler"
+		azure_log_warning "Not building with Intel compilers"
 	fi
 	az_module_unload $INTEL_MODULE
 }
--- a/config/m4/compiler.m4
+++ b/config/m4/compiler.m4
@@ -282,7 +282,7 @@ AC_DEFUN([CHECK_DEPRECATED_DECL_FLAG],
                                   int main(int argc, char** argv) { return f(); }
                             ]])],
                            [AC_MSG_RESULT([yes])
-                            $2="${$2} $1"],
+                            $2="$1"],
                            [AC_MSG_RESULT([no])])
          CFLAGS="$SAVE_CFLAGS"
 ])
@@ -296,6 +296,10 @@ ADD_COMPILER_FLAGS_IF_SUPPORTED([[-diag-error 10006],
                                  [-diag-error 10148]],
                                 [AC_LANG_SOURCE([[int main(int argc, char **argv){return 0;}]])])
 
+#
+# Disable all deprecations if a finer grained approach does not work
+#
+CFLAGS_NO_DEPRECATED="-Wno-deprecated"
 
 CHECK_DEPRECATED_DECL_FLAG([-diag-disable 1478], CFLAGS_NO_DEPRECATED) # icc
 CHECK_DEPRECATED_DECL_FLAG([-Wno-deprecated-declarations], CFLAGS_NO_DEPRECATED) # gcc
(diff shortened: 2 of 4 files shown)


config/m4/compiler.m4
```diff
@@ -563,6 +567,16 @@ ADD_COMPILER_FLAGS_IF_SUPPORTED([[-Wno-missing-field-initializers],
                                  [-Winvalid-pch]],
                                 [AC_LANG_SOURCE([[int main(int argc, char **argv){return 0;}]])])
 
+#
+# Intel Compiler specifics
+#
+if ! $CC -V | grep -q Intel.*Compiler; then
```
yosefe, line 573: why need to check for intel compiler - I think ADD_COMPILER_FLAGS_IF_SUPPORTED would not add unsupported flags anyway?

tvegas1: yes we could, but since they relax code checking constraints, i thought to do it in a selective way since the failure is specific to intel compiler behavior i think:
1. this one is to be able to build with inline
2. this one is to address reported issue #9888 
3. this is for allow isnan() which is a macro in math.h skipped with icx compiler that uses empty parameter

bosilca (approved):
UCX can compile in release mode with `Intel(R) oneAPI DPC++/C++ Compiler 2024.0.2 (2024.0.2.20231213)`

buildlib/tools/builds.sh
```diff
@@ -110,23 +110,39 @@ build_release_pkg() {
...
-		${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst CC=clang CXX=clang++
-		$MAKEP
-		make_clean distclean
+		if icc -v
+		then
+			echo "==== Build with Intel compiler (clang) ===="
+			${WORKSPACE}/contrib/configure-devel --prefix=$ucx_inst CC=clang CXX=clang++
+			$MAKEP
+			make_clean distclean
+		fi
```
yosefe, line 143: can we reuse build_icc_check here?

config/m4/compiler.m4
yosefe: Not sure I understand the explanation.. do you mean ADD_COMPILER_FLAGS_IF_SUPPORTED will not always work correctly for non-intel compilers?

src/ucs/Makefile.am
```diff
@@ -15,7 +15,7 @@ bin_PROGRAMS        =
 libucs_la_CPPFLAGS = $(BASE_CPPFLAGS) $(BFD_CPPFLAGS) \
                      -DUCX_MODULE_DIR=\"$(moduledir)\" \
                      -DUCX_CONFIG_DIR=\"$(ucx_config_dir)\"
-libucs_la_CFLAGS   = $(BASE_CFLAGS) $(BFD_CFLAGS)
+libucs_la_CFLAGS   = $(BASE_CFLAGS) $(BFD_CFLAGS) $(CFLAGS_NO_DEPRECATED)
```
yosefe, line 18: what in UCS is deprecated? maybe we can avoid it?
maybe we can put isnan+isinf inside a preprocessor macro that check is fast-path float is disabled?

config/m4/compiler.m4
tvegas1: As you suggested, simply adding to existing list, without checking for icc first, would work I think. But since it relaxes the compiler checks, it could lead to less strict compiling warning/errors.

Moreover the reason for relaxing those requirements appear to be very icc/icx specific behavior, for instance:
- Wno-language-extension-token is needed otherwise `inline` is not accepted (but __typeof__ etc become allowed).
- -fno-finite-math-only is needed otherwise isnan isinf is producing error on icx, other compilers simply replace those by true when applicable
- -Wno-recommended-options seems to be icx only anyways (recommends -fopenmp -> -qopenmp)
- Wno-c99-extentions, math.h shipped with icx defines isnan() as a macro with an empty argument which triggers an error

buildlib/tools/builds.sh
tvegas1: not exactly as `if icc -v then` is different from `CC=clang`

src/ucs/Makefile.am
tvegas1: It is not isnan/isinf, it is ".. use readdir() instead of readdir_r() which is deprecated ..", coming from added `icx` builds.

buildlib/tools/builds.sh
yosefe: maybe it could work also with clang? or pass another $3 arg. code is very similar..

src/ucs/Makefile.am
yosefe: > It is not isnan/isinf, it is ".. use readdir() instead of readdir_r() which is deprecated ..", coming from added `icx` builds.

so maybe let's change to readdir_r? IMO it's better to fix than adding compiler flags to ignore

config/m4/compiler.m4
yosefe: so maybe need a AC_LANG_SOURCE style test to test if each thing is needed?
re isnan/isinf - maybe we can fix it in the code?

buildlib/tools/builds.sh
tvegas1: makes sense, will simply replace icc -v by clang -v and run anyway

src/ucs/Makefile.am
tvegas1: it's the other way around, ucs uses readdir_r(). we'd need to move to readdir(), which appears to not always be reentrant for it's internal buffers, depending on libcs/versions, so I'd keep that change for another PR.

config/m4/compiler.m4
tvegas1: could be done, but since every compiler/version will have its own needs, i think it might be more maintainable to simply add as is.

for isnan/isinf, i thought about this. there is `__FINITE_MATH_ONLY__ ` but icx does not define it...

src/ucs/Makefile.am
yosefe: According to https://man7.org/linux/man-pages/man3/readdir_r.3.html, seems we can use readdir() as it is also thread-safe in modern implementations.

config/m4/compiler.m4
yosefe: ok

src/ucs/Makefile.am
tvegas1: ok, replaced with `readdir()`.

buildlib/tools/builds.sh
tvegas1: fixed

brminich (approved):

yosefe (approved):

PR #9907, opened 2024-05-28
UCS/SOCK: Added subnet match support

shasson5:
## What
Added new API for subnet match support.

## Why ?
To allow subnet matching from all parts of code.
Before this PR, subnet matching was done only in ib_iface_roce_is_reachable, but there is a demand to support it also for
matching with user configured subnets (ie. for GID selection).

This is the first PR of the GID subnet filter feature.

### DIFF
--- a/src/ucs/sys/sock.c
+++ b/src/ucs/sys/sock.c
@@ -951,6 +951,38 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
                   UCS_IPV4_ADDR_LEN : UCS_IPV6_ADDR_LEN);
 }
 
+int ucs_sockaddr_is_same_subnet(const struct sockaddr *sa1,
+                                const struct sockaddr *sa2, unsigned prefix_len)
+{
+    const void *ipaddr1, *ipaddr2;
+    size_t addr_size, addr_size_bits;
+
+    if (sa1->sa_family != sa2->sa_family) {
+        ucs_debug("different addr_family: s1 %s s2 %s",
+                  ucs_sockaddr_address_family_str(sa1->sa_family),
+                  ucs_sockaddr_address_family_str(sa2->sa_family));
+        return 0;
+    }
+
+    /* Get address size */
+    if (ucs_sockaddr_inet_addr_sizeof(sa1, &addr_size) != UCS_OK) {
+        return 0;
+    }
+
+    addr_size_bits = addr_size * CHAR_BIT;
+
+    /* Truncate prefix_len if it exceeds address size */
+    prefix_len = ucs_min(prefix_len, addr_size_bits);
+
+    ipaddr1 = ucs_sockaddr_get_inet_addr(sa1);
+    ipaddr2 = ucs_sockaddr_get_inet_addr(sa2);
+    ucs_assertv((ipaddr1 != NULL) && (ipaddr2 != NULL), "ipaddr1=%p ipaddr2=%p",
+                ipaddr1, ipaddr2);
+
+    /* Check if the addresses have matching prefixes */
+    return ucs_bitwise_is_equal(ipaddr1, ipaddr2, prefix_len);
+}
+
 ucs_status_t ucs_sockaddr_set_inaddr_any(struct sockaddr *saddr, sa_family_t af)
 {
     struct sockaddr_in *sa_in;
--- a/src/ucs/sys/sock.h
+++ b/src/ucs/sys/sock.h
@@ -540,6 +540,21 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2);
 
 
 /**
+ * Checks if two IP addresses are in the same subnet
+ *
+ * @param [in]  sa1         Pointer to sockaddr structure #1.
+ * @param [in]  sa2         Pointer to sockaddr structure #2.
+ * @param [in]  prefix_len  Subnet prefix length in bits.
+ *
+ * @return 1 if both addresses are in the same subnet.
+ *         0 if not
+ */
+int ucs_sockaddr_is_same_subnet(const struct sockaddr *sa1,
+                                const struct sockaddr *sa2,
+                                unsigned prefix_len);
+
+
+/**
  * Indicate if given IP address is INADDR_ANY (IPV4) or in6addr_any (IPV6)
  *
  * @param [in]   addr       Pointer to sockaddr structure.
--- a/src/uct/ib/base/ib_iface.c
+++ b/src/uct/ib/base/ib_iface.c
@@ -614,6 +614,56 @@ ucs_status_t uct_ib_iface_get_device_address(uct_iface_h tl_iface,
     return UCS_OK;
 }
 
+static void uct_ib_iface_log_subnet_info(const struct sockaddr_storage *sa1,
+                                         const struct sockaddr_storage *sa2,
+                                         unsigned prefix_len, int matched)
+{
+    UCS_STRING_BUFFER_ONSTACK(info, UCS_SOCKADDR_STRING_LEN * 2 + 60);
+
+    ucs_string_buffer_appendf(&info, "IP addresses");
+    if (!matched) {
+        ucs_string_buffer_appendf(&info, " do not");
+    }
+
+    ucs_string_buffer_appendf(&info,
+                              " match with a %u-bit prefix, addresses: ",
+                              prefix_len);
+
+    ucs_string_buffer_append_saddr(&info, (struct sockaddr*)sa1);
+    ucs_string_buffer_appendf(&info, " ");
+    ucs_string_buffer_append_saddr(&info, (struct sockaddr*)sa2);
+    ucs_debug("%s", ucs_string_buffer_cstr(&info));
+}
+
+static ucs_status_t
+uct_ib_iface_roce_to_sockaddr(sa_family_t af, const void *gid,
+                              struct sockaddr_storage *sock_storage)
+{
+    struct sockaddr *sa = (struct sockaddr *)sock_storage;
+    const uint8_t *inet_addr;
+    size_t addr_size;
+    ucs_status_t status;
+
+    /* Set address family */
+    sa->sa_family = af;
+
+    /* Set port to 0 as it's not relevant for RoCE */
+    status = ucs_sockaddr_set_port(sa, 0);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    /* Get address size */
+    status = ucs_sockaddr_inet_addr_size(af, &addr_size);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    /* Set IP address */
+    inet_addr = UCS_PTR_BYTE_OFFSET(gid, sizeof(union ibv_gid) - addr_size);
+    return ucs_sockaddr_set_inet_addr(sa, inet_addr);
+}
+
 static int
 uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
                                const uct_ib_address_t *remote_ib_addr,
@@ -622,14 +672,11 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
     sa_family_t local_ib_addr_af         = local_gid_info->roce_info.addr_family;
     uct_ib_roce_version_t local_roce_ver = local_gid_info->roce_info.ver;
     uint8_t remote_ib_addr_flags         = remote_ib_addr->flags;
+    struct sockaddr_storage sa_local, sa_remote;
     uct_ib_roce_version_t remote_roce_ver;
     sa_family_t remote_ib_addr_af;
     char local_str[128], remote_str[128];
-    uint8_t *local_addr, *remote_addr;
-    ucs_status_t status;
-    size_t addr_offset;
-    size_t addr_size;
-    int ret;
+    int matched;
 
     /* check for wildcards in the RoCE version (RDMACM or non-RoCE cases) */
     if ((uct_ib_address_flags_get_roce_version(remote_ib_addr_flags)) ==
@@ -640,17 +687,6 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
     /* check for zero-sized netmask */
     if (prefix_bits == 0) {
         return 1;
-    }
-
-    /* check the address family */
-    remote_ib_addr_af = uct_ib_address_flags_get_roce_af(remote_ib_addr_flags);
-
-    if (local_ib_addr_af != remote_ib_addr_af) {
-        ucs_assert(local_ib_addr_af != 0);
-        ucs_debug("different addr_family detected. local %s remote %s",
-                  ucs_sockaddr_address_family_str(local_ib_addr_af),
-                  ucs_sockaddr_address_family_str(remote_ib_addr_af));
-        return 0;
     }
 
     /* check the RoCE version */
@@ -674,31 +710,23 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
         return 1; /* We assume it is, but actually there's no good test */
     }
 
-    status = ucs_sockaddr_inet_addr_size(local_ib_addr_af, &addr_size);
-    if (status != UCS_OK) {
-        ucs_error("failed to detect RoCE address size");
+    remote_ib_addr_af = uct_ib_address_flags_get_roce_af(remote_ib_addr_flags);
+    if ((uct_ib_iface_roce_to_sockaddr(local_ib_addr_af, &local_gid_info->gid,
+                                          &sa_local) != UCS_OK) ||
+        (uct_ib_iface_roce_to_sockaddr(remote_ib_addr_af, remote_ib_addr + 1,
+                                          &sa_remote) != UCS_OK)) {
         return 0;
     }
 
-    addr_offset = sizeof(union ibv_gid) - addr_size;
-    local_addr  = UCS_PTR_BYTE_OFFSET(&local_gid_info->gid, addr_offset);
-    remote_addr = UCS_PTR_BYTE_OFFSET(&remote_ib_addr->flags + 1, addr_offset);
+    matched = ucs_sockaddr_is_same_subnet((struct sockaddr*)&sa_local,
+                                          (struct sockaddr*)&sa_remote,
+                                          prefix_bits);
 
-    /* sanity check on the subnet mask size (bits belonging to the prefix) */
-    ucs_assert((prefix_bits / 8) <= addr_size);
+    if (ucs_log_is_enabled(UCS_LOG_LEVEL_DEBUG)) {
+        uct_ib_iface_log_subnet_info(&sa_local, &sa_remote, prefix_bits, matched);
+    }
 
-    /* check if the addresses have matching prefixes */
-    ret = ucs_bitwise_is_equal(local_addr, remote_addr, prefix_bits);
-
-    ucs_debug(ret ? "IP addresses match with a %u-bit prefix: local IP is %s,"
-                    " remote IP is %s" :
-                    "IP addresses do not match with a %u-bit prefix. local IP"
-                    " is %s, remote IP is %s",
-              prefix_bits,
-              inet_ntop(local_ib_addr_af, local_addr, local_str, 128),
-              inet_ntop(remote_ib_addr_af, remote_addr, remote_str, 128));
-
-    return ret;
+    return matched;
 }
 
 int uct_ib_iface_is_same_device(const uct_ib_address_t *ib_addr, uint16_t dlid,
--- a/test/gtest/ucs/test_sock.cc
+++ b/test/gtest/ucs/test_sock.cc
@@ -53,6 +53,48 @@ protected:
             EXPECT_EQ(exp_str, test_str);
         }
     }
+
+    void verify_subnet_match(const char *ip_addr1,
+                             const char *ip_addr2,
+                             unsigned longest_common_prefix) const
+    {
+        struct sockaddr_storage saddr1, saddr2;
+        int matched;
+
+        ASSERT_UCS_OK(ucs_sock_ipstr_to_sockaddr(ip_addr1, &saddr1));
+        ASSERT_UCS_OK(ucs_sock_ipstr_to_sockaddr(ip_addr2, &saddr2));
+
+        size_t addr_size;
+        ASSERT_UCS_OK(ucs_sockaddr_inet_addr_sizeof((struct sockaddr *)&saddr1,
+                                                    &addr_size));
+        size_t addr_size_bits = addr_size * CHAR_BIT;
+
+        /* Add extra iterations to check prefix truncation */
+        constexpr size_t extra_iters = 10;
+
+        for (size_t prefix = 0; prefix <= addr_size_bits + extra_iters; ++prefix) {
+            matched = ucs_sockaddr_is_same_subnet((struct sockaddr*)&saddr1,
+                                                  (struct sockaddr*)&saddr2,
+                                                  prefix);
+            EXPECT_EQ(std::min(prefix, addr_size_bits) <= longest_common_prefix,
+                      matched)
+                    << "failed to match subnet: ip1=" << ip_addr1
+                    << " ip2=" << ip_addr2 << " prefix=" << prefix
+                    << " max_common=" << longest_common_prefix
+                    << " (matched=" << matched << ")";
+        }
+    }
+
+    void verify_subnet_error(sa_family_t af1, sa_family_t af2, unsigned prefix) const
+    {
+        struct sockaddr_storage storage1 = {0}, storage2 = {0};
+        struct sockaddr *sa1             = (struct sockaddr *)&storage1;
+        struct sockaddr *sa2             = (struct sockaddr *)&storage2;
+
+        sa1->sa_family = af1;
+        sa2->sa_family = af2;
+        EXPECT_FALSE(ucs_sockaddr_is_same_subnet(sa1, sa2, prefix));
+    }
 };
 
 UCS_TEST_F(test_socket, sockaddr_sizeof) {
@@ -672,6 +714,44 @@ UCS_TEST_F(test_socket, sockaddr_cmp_err) {
 
     sockaddr_cmp_err_test((const struct sockaddr*)&sa_un,
                           (const struct sockaddr*)&sa_in);
+}
+
+UCS_TEST_F(test_socket, subnet_match) {
+    /* IPv4 non byte-aligned */
+    verify_subnet_match("192.168.122.28", "192.168.122.13", 27);
+    verify_subnet_match("192.168.3.28", "192.168.7.13", 21);
+    verify_subnet_match("192.45.3.28", "192.14.7.13", 10);
+
+    /* IPv4 byte-aligned */
+    verify_subnet_match("192.168.173.157", "10.33.44.100", 0);
+    verify_subnet_match("192.168.173.157", "192.33.44.100", 8);
+    verify_subnet_match("192.168.173.157", "192.168.44.100", 16);
+    verify_subnet_match("192.168.173.157", "192.168.173.100", 24);
+    verify_subnet_match("192.168.173.157", "192.168.173.157", 32);
+
+    /* IPv6 non byte-aligned */
+    verify_subnet_match("fe80::218:e7ff:fe16:fb97", "fe80::219:e7ff:fe16:fb97",
+                        79);
+    verify_subnet_match("fe80::218:e7ff:fef8:fb97", "fe80::218:e7ff:feff:fb97",
+                        109);
+
+    /* IPv6 byte-aligned */
+    verify_subnet_match("fe80::218:e7ff:fe16:fb97", "fe80::218:e77f:fe16:fb97",
+                        88);
+    verify_subnet_match("fe80::218:e7ff:fe16:fb97", "fe80::218:e7ff:fe16:fb97",
+                        128);
+}
+
+UCS_TEST_F(test_socket, subnet_match_failure) {
+    scoped_log_handler log_handler(socket_error_handler);
+
+    /* Unknown address family */
+    socket_err_exp_str = "unknown address family: ";
+    verify_subnet_error(AF_UNIX, AF_INET, 10);
+    verify_subnet_error(AF_INET, AF_UNIX, 10);
+
+    /* Different address families */
+    verify_subnet_error(AF_INET, AF_INET6, 10);
 }
 
 static void sockaddr_get_ipstr_check(const struct sockaddr *sockaddr,
--- a/src/ucs/datastruct/string_buffer.c
+++ b/src/ucs/datastruct/string_buffer.c
@@ -15,6 +15,7 @@
 #include <ucs/debug/memtrack_int.h>
 #include <ucs/sys/string.h>
 #include <ucs/sys/math.h>
+#include <ucs/sys/sock.h>
 #include <string.h>
 #include <ctype.h>
 
@@ -144,6 +145,15 @@ void ucs_string_buffer_append_iovec(ucs_string_buffer_t *strb,
                                   iov[iov_index].iov_len);
     }
     ucs_string_buffer_rtrim(strb, "|");
+}
+
+void ucs_string_buffer_append_saddr(ucs_string_buffer_t *strb,
+                                    const struct sockaddr *sa)
+{
+    char sockstr[UCS_SOCKADDR_STRING_LEN];
+
+    ucs_sockaddr_str(sa, sockstr, UCS_SOCKADDR_STRING_LEN);
+    ucs_string_buffer_appendf(strb, "%s", sockstr);
 }
 
 static int ucs_string_buffer_match_charset(char ch, const char *charset)
--- a/src/ucs/datastruct/string_buffer.h
+++ b/src/ucs/datastruct/string_buffer.h
@@ -10,6 +10,7 @@
 #include <ucs/sys/compiler_def.h>
 #include <ucs/type/status.h>
 #include <ucs/datastruct/array.h>
+#include <sys/socket.h>
 #include <sys/uio.h>
 #include <stdint.h>
 #include <stddef.h>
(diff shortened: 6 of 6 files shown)


test/gtest/ucs/test_sock.cc
```diff
@@ -674,6 +719,48 @@ UCS_TEST_F(test_socket, sockaddr_cmp_err) {
...
+    socket_err_exp_str = "unknown address family: ";
+    verify_subnet_error(AF_UNIX, AF_INET, 10, UCS_ERR_INVALID_PARAM);
+    verify_subnet_error(AF_INET, AF_UNIX, 10, UCS_ERR_INVALID_PARAM);
+
+    /* Different address families */
+    verify_subnet_error(AF_INET, AF_INET6, 10, UCS_ERR_INVALID_PARAM);
+
+    /* Prefix out of range */
+    socket_err_exp_str = "prefix is bigger than address size";
+    verify_subnet_error(AF_INET, AF_INET, 45, UCS_ERR_EXCEEDS_LIMIT);
```
tvegas1, line 761: test with 33 for v4 and 129 for v6?

test/gtest/ucs/test_sock.cc
```diff
@@ -674,6 +719,48 @@ UCS_TEST_F(test_socket, sockaddr_cmp_err) {
                           (const struct sockaddr*)&sa_in);
 }
 
+UCS_TEST_F(test_socket, subnet_match) {
```
tvegas1, line 722: maybe add similar tests for case where af matches but subnet itself does not match?

src/ucs/sys/sock.c
```diff
@@ -951,6 +959,66 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+    }
+
+    ipaddr1 = ucs_sockaddr_get_inet_addr(sa1);
+    ipaddr2 = ucs_sockaddr_get_inet_addr(sa2);
+    ucs_assert((ipaddr1 != NULL) && (ipaddr2 != NULL));
+
+    /* Check if the addresses have matching prefixes */
+    matched = ucs_bitwise_is_equal(ipaddr1, ipaddr2, prefix_len);
+
+    if ((ucs_sockaddr_get_ipstr(sa1, ip1_str, UCS_SOCKADDR_STRING_LEN) ==
```
tvegas1, line 1006: check for debug enablement before generating ipstr with `ucs_log_is_enabled(log_level)`

src/ucs/sys/sock.c
```diff
@@ -951,6 +959,67 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+    }
+
+    ipaddr1 = ucs_sockaddr_get_inet_addr(sa1);
+    ipaddr2 = ucs_sockaddr_get_inet_addr(sa2);
+    ucs_assert((ipaddr1 != NULL) && (ipaddr2 != NULL));
+
+    /* Check if the addresses have matching prefixes */
+    matched = ucs_bitwise_is_equal(ipaddr1, ipaddr2, prefix_len);
+
+    if (ucs_log_is_enabled(UCS_LOG_LEVEL_DEBUG) &&
```
tvegas1, line 1006: maybe dump ucs_debug() with empty/default string if one of the get_ipstr fail

src/ucs/sys/sock.c
```diff
@@ -951,6 +959,67 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+        ucs_error("prefix is bigger than address size (in bits): prefix=%u"
+                  " size=%lu",
+                  prefix_len, addr_size_bits);
+        status = UCS_ERR_EXCEEDS_LIMIT;
+        goto out;
+    }
+
+    ipaddr1 = ucs_sockaddr_get_inet_addr(sa1);
+    ipaddr2 = ucs_sockaddr_get_inet_addr(sa2);
+    ucs_assert((ipaddr1 != NULL) && (ipaddr2 != NULL));
```
tvegas1, line 1001: ucs_assert_v()

src/ucs/sys/sock.c
```diff
@@ -951,6 +959,67 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+    if (ucs_log_is_enabled(UCS_LOG_LEVEL_DEBUG) &&
+        (ucs_sockaddr_get_ipstr(sa1, ip1_str, UCS_SOCKADDR_STRING_LEN) ==
+         UCS_OK) &&
+        (ucs_sockaddr_get_ipstr(sa2, ip2_str, UCS_SOCKADDR_STRING_LEN) ==
+         UCS_OK)) {
+        ucs_debug(matched ? "IP addresses match with a %u-bit prefix: local"
+                            " IP is %s, remote IP is %s" :
+                            "IP addresses do not match with a %u-bit prefix:"
+                            " local IP is %s, remote IP is %s",
+                  prefix_len, ip1_str, ip2_str);
```
brminich, line 1015:
```suggestion
        ucs_debug("IP addresses%s match with a %u-bit prefix: local"
                  " IP is %s, remote IP is %s", matched ? "" : " do not"
                  prefix_len, ip1_str, ip2_str);
```

shasson5: maybe use string builder? because we also need to distinguish when ip_str is NULL/non-NULL

brminich (approved):

tvegas1 (dismissed):

src/ucs/sys/sock.c
```diff
@@ -951,6 +959,87 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+        goto out;
+    }
+
+    /* Get address size */
+    ucs_assert_always(ucs_sockaddr_inet_addr_sizeof(sa1, &addr_size) == UCS_OK);
+    addr_size_bits = addr_size * UCS_SOCKET_BYTE_TO_BITS;
+
+    /* Check subnet mask size (bits belonging to the prefix) */
+    if (prefix_len > addr_size_bits) {
+        ucs_error("prefix is bigger than address size (in bits): prefix=%u"
```
gleon99, line 1022: bigger -> longer

src/ucs/sys/sock.c
```diff
@@ -951,6 +959,87 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+    }
+
+    ipaddr1 = ucs_sockaddr_get_inet_addr(sa1);
+    ipaddr2 = ucs_sockaddr_get_inet_addr(sa2);
+    ucs_assertv((ipaddr1 != NULL) && (ipaddr2 != NULL), "ipaddr1=%p ipaddr2=%p",
+                ipaddr1, ipaddr2);
+
+    /* Check if the addresses have matching prefixes */
+    matched = ucs_bitwise_is_equal(ipaddr1, ipaddr2, prefix_len);
+    ucs_sockaddr_print_subnet_info(sa1, sa2, prefix_len, matched);
```
gleon99, line 1036: What is the rationale in calling a print func when we only compare? I noticed the `if (!ucs_log_is_enabled...` there, and yet...

test/gtest/ucs/test_sock.cc
```diff
@@ -17,6 +17,22 @@ static std::string socket_err_exp_str;
 
 class test_socket : public ucs::test {
 public:
+    static void fill_sockaddr(sa_family_t af, const char *ip_addr,
```
gleon99, line 20: Usually, in a `fill_something` func, the first arg is `something`

test/gtest/ucs/test_sock.cc
```diff
@@ -53,6 +69,48 @@ class test_socket : public ucs::test {
             EXPECT_EQ(exp_str, test_str);
         }
     }
+
+    void verify_subnet_match(sa_family_t af, const char *ip_addr1,
+                             const char *ip_addr2, unsigned max_common_prefix)
+    {
+        struct sockaddr_storage storage1 = {0};
```
gleon99, line 76: Is the init needed?

brminich (dismissed):

test/gtest/ucs/test_sock.cc
```diff
@@ -53,6 +69,48 @@ class test_socket : public ucs::test {
...
+        struct sockaddr_storage storage2 = {0};
+        struct sockaddr *saddr1          = (struct sockaddr*)&storage1;
+        struct sockaddr *saddr2          = (struct sockaddr*)&storage2;
+        int matched;
+
+        fill_sockaddr(af, ip_addr1, 1234, saddr1);
+        fill_sockaddr(af, ip_addr2, 1234, saddr2);
+
+        static const size_t byte_to_bits = 8;
```
gleon99, line 85: `constexpr`

test/gtest/ucs/test_sock.cc
```diff
@@ -53,6 +69,48 @@ class test_socket : public ucs::test {
...
+        size_t addr_size_bits = get_addr_size(saddr1) * byte_to_bits;
+
+        for (size_t prefix = 0; prefix < addr_size_bits; ++prefix) {
```
gleon99, line 88: Seems that `prefix` is an index, while `max_common_prefix` is a length, so in case of 0 the result is wrong (0 <= 0 is true...)

test/gtest/ucs/test_sock.cc
```diff
@@ -53,6 +69,48 @@ class test_socket : public ucs::test {
             EXPECT_EQ(exp_str, test_str);
         }
     }
+
+    void verify_subnet_match(sa_family_t af, const char *ip_addr1,
```
gleon99, line 73: `static`? `const`?

test/gtest/ucs/test_sock.cc
```diff
@@ -53,6 +69,48 @@ class test_socket : public ucs::test {
...
+                                                    &matched));
+            EXPECT_EQ(prefix <= max_common_prefix, matched)
+                    << "failed to match subnet: ip1=" << ip_addr1
+                    << " ip2=" << ip_addr2 << " prefix=" << prefix
+                    << " max_common=" << max_common_prefix
+                    << " (matched=" << matched << ")";
+        }
+    }
+
+    void verify_subnet_error(sa_family_t af1, sa_family_t af2, unsigned prefix,
```
gleon99, line 99: `static`? `const`?

src/ucs/sys/sock.c
shasson5: not sure I understand, how does it make difference if we compare or do something else?

gleon99: why print smth when only a comparison is requested? that's werid..

test/gtest/ucs/test_sock.cc
shasson5: the convention is passing output params after input params

src/ucs/sys/sock.c
```diff
@@ -887,19 +888,28 @@ ucs_status_t ucs_sock_port_from_string(const char *port_str, uint16_t *port)
     return UCS_OK;
 }
 
-int ucs_sockaddr_cmp(const struct sockaddr *sa1,
-                     const struct sockaddr *sa2,
-                     ucs_status_t *status_p)
+static int ucs_sockaddr_are_both_known_af(const struct sockaddr *sa1,
```
iyastreb, line 891: IMO it's not practical (nor logical) to have this function checking 2 addresses. Finally it's more verbose and cumbersome, I would propose to have just a function checking single address
```
static int ucs_sockaddr_check_af(const struct sockaddr *sa)
{
    if (!ucs_sockaddr_is_known_af(sa)
        error_message
        return 0
    return 1
}

File: src/ucs/sys/sock.c [hunk 16]
```diff
@@ -887,19 +888,28 @@ ucs_status_t ucs_sock_port_from_string(const char *port_str, uint16_t *port)
...
+        return 0;
+    }
+
+    return 1;
+}
+
+int ucs_sockaddr_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2,
+                     ucs_status_t *status_p)
+{
+    int result     = 1;
```
iyastreb, line 908: alignment on =

src/ucs/sys/sock.c
```diff
@@ -951,6 +959,87 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+                                       const struct sockaddr *sa2,
+                                       unsigned prefix_len, int *is_match_p)
+{
+    int matched         = 0;
+    ucs_status_t status = UCS_OK;
+    const void *ipaddr1, *ipaddr2;
+    size_t addr_size, addr_size_bits;
+
+    if (!ucs_sockaddr_are_both_known_af(sa1, sa2)) {
+        status = UCS_ERR_INVALID_PARAM;
```
iyastreb, line 1004: do we really need goto in this function? why not just return the status?

test/gtest/ucs/test_sock.cc
shasson5: actually prefix is also length, and the result should be 'true' in that case. (prefix = 0 => always match).
but seems that the case where prefix == addr_size_bits is not covered by the loop.

test/gtest/ucs/test_sock.cc
shasson5: why static?

src/ucs/sys/sock.c
shasson5: for debug purposes

src/ucs/sys/sock.c
shasson5: won't it be problematic to add `*is_equal_p = is_equal;` before each return?

iyastreb: Do you mean setting `is_match_p` (is_equal_p is from another function)?
Ok, I see, I thought it's not important to set that variable in case of an error, but if you think it's useful please keep the current impl

src/ucs/sys/sock.c
iyastreb: IMO it's ok to have a function just for error message, but for a single argument
With 2 arguments it's not clear which one is failing (or both?), and you have to check again

shasson5: how about moving the error message to `ucs_sockaddr_is_known_af`?

iyastreb: IMO it's not desirable, because this function is called from `uct_rdmacm_cm_ep_str`, and it basically checks whether IP address is initialized or not:
```
    if ((local_addr != NULL) && ucs_sockaddr_is_known_af(local_addr)) {
        ucs_sockaddr_str(local_addr, local_ip_port_str, UCS_SOCKADDR_STRING_LEN);
    } else {
        ucs_strncpy_safe(local_ip_port_str, "<invalid>", UCS_SOCKADDR_STRING_LEN);
    }
```
So personally I would propose to add just a simple static function with error message in sock.c, and replace all the invocations of `ucs_sockaddr_is_known_af` inside sock.c with this new function.
```
int ucs_sockaddr_check_known_af(const struct sockaddr *sa)
{
    int res = ucs_sockaddr_is_known_af(sa);
    if (!res) {
        ucs_error("<invalid address family>", ...)
    }
    return res;
}
And then replace all 7 places where ucs_sockaddr_is_known_af is used inside sock.c
```

test/gtest/ucs/test_sock.cc
```diff
@@ -71,23 +71,22 @@ class test_socket : public ucs::test {
...
-                             const char *ip_addr2, unsigned max_common_prefix)
+                             const char *ip_addr2,
+                             unsigned max_common_prefix) const
     {
-        struct sockaddr_storage storage1 = {0};
-        struct sockaddr_storage storage2 = {0};
-        struct sockaddr *saddr1          = (struct sockaddr*)&storage1;
-        struct sockaddr *saddr2          = (struct sockaddr*)&storage2;
+        struct sockaddr_storage storage1, storage2;
+        struct sockaddr *saddr1 = (struct sockaddr*)&storage1;
```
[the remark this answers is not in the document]
shasson5: there's dependency between sockaddr_storage to sockaddr *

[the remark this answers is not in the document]
gleon99: Right. Then, according to our conventions, both have to be uninitialized - then init both.

src/ucs/sys/sock.c
```diff
@@ -951,6 +956,87 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+
+    ucs_string_buffer_appendf(&info, "IP addresses");
+    if (!matched) {
+        ucs_string_buffer_appendf(&info, " do not");
+    }
+
+    ucs_string_buffer_appendf(&info, " match with a %u-bit prefix,",
+                              prefix_len);
+
+    ucs_carray_for_each(sa, saddrs, ucs_static_array_size(saddrs)) {
```
yosefe, line 978: maybe no need array and for loop, since it's always 2 addresses?

src/ucs/sys/sock.c
```diff
@@ -951,6 +956,87 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+    const void *ipaddr1, *ipaddr2;
+    size_t addr_size, addr_size_bits;
+
+    if (!ucs_sockaddr_check_known_af(sa1) ||
+        !ucs_sockaddr_check_known_af(sa2)) {
+        status = UCS_ERR_INVALID_PARAM;
+        goto out;
+    }
+
+    if (sa1->sa_family != sa2->sa_family) {
```
yosefe, line 1002: we can first check sa_family is the same, and that ucs_sockaddr_inet_addr_sizeof is OK. after checking that, no need to check ucs_sockaddr_check_known_af any more.

src/ucs/sys/sock.c
```diff
@@ -951,6 +956,87 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+    ucs_carray_for_each(sa, saddrs, ucs_static_array_size(saddrs)) {
+        status = ucs_sockaddr_get_ipstr(*sa, ip_str, UCS_SOCKADDR_STRING_LEN);
+        ucs_string_buffer_appendf(&info, " sa%u: %s", ++sa_index,
+                                  (status == UCS_OK) ? ip_str : "<null>");
+    }
+
+    ucs_debug("%s", ucs_string_buffer_cstr(&info));
+}
+
+ucs_status_t ucs_sockaddr_is_same_subnet(const struct sockaddr *sa1,
```
yosefe, line 987: can we reuse this function in uct_ib_iface_roce_is_reachable ?

src/ucs/sys/sock.c
```diff
@@ -951,6 +956,87 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+    /* Get address size */
+    ucs_assert_always(ucs_sockaddr_inet_addr_sizeof(sa1, &addr_size) == UCS_OK);
+    addr_size_bits = addr_size * CHAR_BIT;
+
+    /* Check subnet mask size (bits belonging to the prefix) */
+    if (prefix_len > addr_size_bits) {
+        ucs_error("prefix is longer than address size (in bits): prefix=%u"
+                  " size=%lu",
+                  prefix_len, addr_size_bits);
+        status = UCS_ERR_EXCEEDS_LIMIT;
```
yosefe, line 1019: IMO the error here should be UCS_ERR_INVALID_PARAM

src/ucs/sys/sock.c
```diff
@@ -951,6 +956,87 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
+    ipaddr2 = ucs_sockaddr_get_inet_addr(sa2);
+    ucs_assertv((ipaddr1 != NULL) && (ipaddr2 != NULL), "ipaddr1=%p ipaddr2=%p",
+                ipaddr1, ipaddr2);
+
+    /* Check if the addresses have matching prefixes */
+    is_same = ucs_bitwise_is_equal(ipaddr1, ipaddr2, prefix_len);
+
+    if (ucs_log_is_enabled(UCS_LOG_LEVEL_DEBUG)) {
+        ucs_sockaddr_log_subnet_info(sa1, sa2, prefix_len, is_same);
+    }
```
yosefe, line 1033: seems the log is excessive, in UCS functions we don't usually print such logs, the log message has to include some information about the context in which the address is compared and what it means the address is matching or not

src/ucs/sys/sock.h
```diff
@@ -539,6 +539,25 @@ int ucs_sockaddr_cmp(const struct sockaddr *sa1,
...
+ * @param [in]  prefix_len  Subnet prefix length in bits.
+ * @param [out] is_same_p   Pointer to a boolean value indicating if both
+ *                          addresses are in the same subnet.
+ *
+ * @return UCS_OK on success or one of the following errors:
+ *         UCS_ERR_INVALID_PARAM - wrong or mismatched address families.
+ *         UCS_ERR_EXCEEDS_LIMIT - prefix_len is too big (exceeds size of
+ *                                 address).
+ */
+ucs_status_t ucs_sockaddr_is_same_subnet(const struct sockaddr *sa1,
```
yosefe, line 556: maybe just return if same subnet or not?
if there is any kind of error in the parameters, can return 0
if prefix is longer than address length, can check according to the actual address length

test/gtest/ucs/test_sock.cc
```diff
@@ -17,6 +17,22 @@ static std::string socket_err_exp_str;
 
 class test_socket : public ucs::test {
 public:
+    static void fill_sockaddr(struct sockaddr *sa, sa_family_t af,
+                              const char *ip_addr, unsigned port)
+    {
+        sa->sa_family = af;
+        inet_pton(af, ip_addr,
+                  const_cast<void*>(ucs_sockaddr_get_inet_addr(sa)));
```
yosefe, line 25: why need const_cast?

test/gtest/ucs/test_sock.cc
```diff
@@ -17,6 +17,22 @@ static std::string socket_err_exp_str;
...
+        ASSERT_UCS_OK(ucs_sockaddr_set_port(sa, port));
+        ASSERT_TRUE(ucs_sockaddr_get_inet_addr(sa) != NULL);
```
yosefe, line 27: why needed?

test/gtest/ucs/test_sock.cc
```diff
@@ -17,6 +17,22 @@ static std::string socket_err_exp_str;
 
 class test_socket : public ucs::test {
 public:
+    static void fill_sockaddr(struct sockaddr *sa, sa_family_t af,
```
yosefe, line 20: maybe use ucs_sock_ipstr_to_sockaddr?

test/gtest/ucs/test_sock.cc
```diff
@@ -17,6 +17,22 @@ static std::string socket_err_exp_str;
...
+                              const char *ip_addr, unsigned port)
+    {
+        sa->sa_family = af;
+        inet_pton(af, ip_addr,
+                  const_cast<void*>(ucs_sockaddr_get_inet_addr(sa)));
+        ASSERT_UCS_OK(ucs_sockaddr_set_port(sa, port));
+        ASSERT_TRUE(ucs_sockaddr_get_inet_addr(sa) != NULL);
+    }
+
+    static size_t get_addr_size(const struct sockaddr *sa)
```
yosefe, line 30: maybe use ucs_sockaddr_sizeof?

test/gtest/ucs/test_sock.cc
```diff
@@ -53,6 +69,50 @@ class test_socket : public ucs::test {
...
+        size_t addr_size_bits = get_addr_size(saddr1) * CHAR_BIT;
+
+        for (size_t prefix = 0; prefix <= addr_size_bits; ++prefix) {
+            ASSERT_UCS_OK(ucs_sockaddr_is_same_subnet(saddr1, saddr2, prefix,
+                                                      &matched));
+            EXPECT_EQ(prefix <= longest_common_prefix, matched)
+                    << "failed to match subnet: ip1=" << ip_addr1
+                    << " ip2=" << ip_addr2 << " prefix=" << prefix
+                    << " max_common=" << longest_common_prefix
+                    << " (matched=" << matched << ")";
```
yosefe, line 96: why needed? prefix is defined by addr_size_bits
can iterate from 0 to std::min(addr_size_bits, longest_common_prefix)

test/gtest/ucs/test_sock.cc
```diff
@@ -53,6 +69,50 @@ class test_socket : public ucs::test {
...
+    void verify_subnet_match(sa_family_t af, const char *ip_addr1,
+                             const char *ip_addr2,
+                             unsigned longest_common_prefix) const
+    {
+        struct sockaddr_storage storage1, storage2;
+        struct sockaddr *saddr1;
+        struct sockaddr *saddr2;
+        int matched;
+
+        saddr1 = (struct sockaddr*)&storage1;
```
yosefe, line 82: `struct sockaddr *saddr1 = (struct sockaddr*)&storage1;`

test/gtest/ucs/test_sock.cc
```diff
@@ -53,6 +69,50 @@ class test_socket : public ucs::test {
...
+                    << " (matched=" << matched << ")";
+        }
+    }
+
+    void verify_subnet_error(sa_family_t af1, sa_family_t af2, unsigned prefix,
+                             ucs_status_t expected_status) const
+    {
+        struct sockaddr_storage storage1, storage2;
+        struct sockaddr *sa1 = (struct sockaddr*)&storage1;
+        struct sockaddr *sa2 = (struct sockaddr*)&storage2;
```
yosefe, line 105: seems the addresses are not initialized

src/ucs/sys/sock.c
shasson5: it is referenced in PR's description:
`The code in IB reachability function will be refactored to use the new API in the next PR.`

test/gtest/ucs/test_sock.cc
shasson5: please see:
https://github.com/openucx/ucx/pull/9907#discussion_r1626020044

test/gtest/ucs/test_sock.cc
shasson5: what do you mean? they are initialized with the address of the stack vars

src/ucs/sys/sock.c
shasson5: this function was removed.

src/ucs/sys/sock.c
yosefe: IMO better avoid code duplication already in this PR and it will also improve coverage by existing tests

test/gtest/ucs/test_sock.cc
yosefe: I mean the contents of storage1, 2

src/ucs/sys/sock.c
shasson5: That will prevent separate error handling in the caller context. (which is needed for ib_iface_is_reachable).

shasson5: logic changed as a result of below comment:
https://github.com/openucx/ucx/pull/9907#discussion_r1654742740

test/gtest/ucs/test_sock.cc
shasson5: because `ucs_sockaddr_get_inet_addr` returns `const void *`

test/gtest/ucs/test_sock.cc
shasson5: the idea is to check all possible sizes and decide whether they should be matched or not.
`longest_common_prefix` is used as the threshold between matched/unmatched subnet.

test/gtest/ucs/test_sock.cc
shasson5: they are not required to be, as this test checks AF errors

test/gtest/ucs/test_sock.cc
shasson5: function was removed

src/ucs/sys/sock.c
```diff
@@ -887,25 +887,19 @@ ucs_status_t ucs_sock_port_from_string(const char *port_str, uint16_t *port)
...
 }
 
-static int ucs_sockaddr_check_known_af(const struct sockaddr *sa)
-{
-    if (!ucs_sockaddr_is_known_af(sa)) {
-        ucs_error("unknown address family: %d", sa->sa_family);
-        return 0;
-    }
-
-    return 1;
```
yosefe: why removed this function? it seems useful to print the error from it

src/ucs/sys/sock.c
```diff
@@ -956,68 +951,34 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2)
...
-    /* Check subnet mask size (bits belonging to the prefix) */
+    /* Truncate prefix_len if it exceeds address size */
     if (prefix_len > addr_size_bits) {
-        ucs_error("prefix is longer than address size (in bits): prefix=%u"
-                  " size=%lu",
-                  prefix_len, addr_size_bits);
-        status = UCS_ERR_EXCEEDS_LIMIT;
-        goto out;
+        ucs_debug("prefix is longer than address size (in bits): prefix=%u"
+                  " size=%lu", prefix_len, addr_size_bits);
```
yosefe, line 980: not sure we need to keep this debug print

src/ucs/sys/sock.h
```diff
@@ -542,20 +542,16 @@ int ucs_sockaddr_ip_cmp(const struct sockaddr *sa1, const struct sockaddr *sa2);
...
- *         UCS_ERR_EXCEEDS_LIMIT - prefix_len is too big (exceeds size of
- *                                 address).
+ * @return 1 if both addresses are in the same subnet.
+ *         0 if not
  */
-ucs_status_t ucs_sockaddr_is_same_subnet(const struct sockaddr *sa1,
-                                         const struct sockaddr *sa2,
-                                         unsigned prefix_len, int *is_same_p);
+int ucs_sockaddr_is_same_subnet(const struct sockaddr_storage *sa1,
+                                const struct sockaddr_storage *sa2,
```
yosefe, line 553: IMO better keep it as const struct sockaddr * since it's input parameter and not output

src/uct/ib/base/ib_iface.c
```diff
@@ -614,6 +614,58 @@ ucs_status_t uct_ib_iface_get_device_address(uct_iface_h tl_iface,
...
+
+    ucs_string_buffer_appendf(&info, "IP addresses");
+    if (!matched) {
+        ucs_string_buffer_appendf(&info, " do not");
+    }
+
+    ucs_string_buffer_appendf(&info, " match with a %u-bit prefix,",
+                              prefix_len);
+
+    ucs_sockaddr_str((struct sockaddr *)sa1, sockstr, UCS_SOCKADDR_STRING_LEN);
```
yosefe, line 632: maybe add new string_buffer function, and use it like so:
`ucs_string_buffer_append_saddr(&info, (struct sockaddr *)sa1);`

src/uct/ib/base/ib_iface.c
```diff
@@ -642,17 +691,6 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
-    /* check the address family */
-    remote_ib_addr_af = uct_ib_address_flags_get_roce_af(remote_ib_addr_flags);
-
-    if (local_ib_addr_af != remote_ib_addr_af) {
-        ucs_assert(local_ib_addr_af != 0);
-        ucs_debug("different addr_family detected. local %s remote %s",
-                  ucs_sockaddr_address_family_str(local_ib_addr_af),
-                  ucs_sockaddr_address_family_str(remote_ib_addr_af));
-        return 0;
-    }
```
yosefe, line 654: 👍

test/gtest/ucs/test_sock.cc
```diff
@@ -70,48 +54,45 @@ class test_socket : public ucs::test {
...
 
-        for (size_t prefix = 0; prefix <= addr_size_bits; ++prefix) {
-            ASSERT_UCS_OK(ucs_sockaddr_is_same_subnet(saddr1, saddr2, prefix,
-                                                      &matched));
-            EXPECT_EQ(prefix <= longest_common_prefix, matched)
-                    << "failed to match subnet: ip1=" << ip_addr1
+        for (size_t prefix = 0; prefix <= addr_size_bits + extra_iters; ++prefix) {
+            matched = ucs_sockaddr_is_same_subnet(&saddr1, &saddr2, prefix);
+            EXPECT_EQ(std::min(prefix,addr_size_bits) <= longest_common_prefix,
+                    matched) << "failed to match subnet: ip1=" << ip_addr1
```
yosefe, line 78: code style: space after `,` and indentation seem weird

test/gtest/ucs/test_sock.cc
```diff
@@ -53,6 +53,47 @@ class test_socket : public ucs::test {
...
+                    << " ip2=" << ip_addr2 << " prefix=" << prefix
+                    << " max_common=" << longest_common_prefix
+                    << " (matched=" << matched << ")";
+        }
+    }
+
+    void verify_subnet_error(sa_family_t af1, sa_family_t af2, unsigned prefix) const
+    {
+        struct sockaddr_storage storage1, storage2;
```
yosefe, line 87: IMO better init storage1 and storage2 to empty/zero address just in case

src/ucs/sys/sock.c
shasson5: it's no longer related to the PR..

yosefe: ok

yosefe (approved):
pls squash

yosefe:
failures seems relevant:
```
4-07-17T18:14:31.9181482Z [ RUN      ] shm_ib/test_ucp_loopback.envelope/17 <shm,ib/str_lat>
2024-07-17T18:14:32.0629570Z [1721240072.061894] [swx-rdmz-ucx-gpu-02:52008:0]            sock.c:699  UCX  ERROR   unknown address family: 32769
2024-07-17T18:14:32.0631039Z [1721240072.061921] [swx-rdmz-ucx-gpu-02:52008:0]            sock.c:699  UCX  ERROR   unknown address family: 32769
2024-07-17T18:14:32.0631984Z [1721240072.061932] [swx-rdmz-ucx-gpu-02:52008:0]            sock.c:699  UCX  ERROR   unknown address family: 32769
```

yosefe (approved):

PR #9915, opened 2024-05-30
UCP/WIREUP: Only skip AM lanes for RMA with supported remote addresses

tvegas1:
## What
Only skip AM lanes with appropriate with peers >= v1.17.

## Why ?
Maintain wire compatibility.

## How ?
Check unpacked remote address version.

### Tested
- same version, am lanes skipped
- older version, am lanes not skipped


brminich:
@yosefe, ok to merge?

PR #9919, opened 2024-05-31
UCT/ROCM: fix test_md_fork

edgargabriel:
## What
fix for the test_md_fork test of gtest for ROCm devices and memory domains.

### DIFF
--- a/src/uct/rocm/copy/rocm_copy_md.c
+++ b/src/uct/rocm/copy/rocm_copy_md.c
@@ -433,7 +433,7 @@ uct_rocm_copy_md_open(uct_component_h component, const char *md_name,
         rcache_params.ucm_event_priority = md_config->rcache.event_prio;
         rcache_params.context            = md;
         rcache_params.ops                = &uct_rocm_copy_rcache_ops;
-        rcache_params.flags              = 0;
+        rcache_params.flags              = UCS_RCACHE_FLAG_PURGE_ON_FORK;
 
         status = ucs_rcache_create(&rcache_params, "rocm_copy", NULL, &md->rcache);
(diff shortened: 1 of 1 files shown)


test/gtest/uct/test_md.cc
```diff
@@ -1067,7 +1067,11 @@ UCS_TEST_SKIP_COND_P(test_md_fork, fork,
         memset(buf, 42, REG_SIZE);
         /* child touch the page */
         EXPECT_EQ(0, memcmp(page, buf, REG_SIZE));
+#if HAVE_ROCM
+        _exit(0);
+#else
         exit(0);
+#endif
```
yosefe, line 1074: why needed?

yosefe: @edgargabriel can we avoid this in the code and solve in a better way?

edgargabriel: @yosefe yes, I was told that the runtime fix have a solution for the fork cleanup problem implemented, and I am just waiting to be able to test that. If that works indeed, I would just keep the first part of the PR and remove the '_exit' from it

yosefe: > @yosefe yes, I was told that the runtime folks have a fix for the fork cleanup problem implemented, and I am just waiting to be able to test that. If that works indeed, I would just keep the first part of the PR and remove the '_exit' from it

sounds good

edgargabriel:
@yosefe I've updated the pr to contain only the change to the rcache_params, and removed the change to gtest. Please have look when you have a chance.

yosefe (approved):
pls squash

PR #9926, opened 2024-06-03
UCP/PROTO: Make perf nodes names more informative

ivankochin:
## What
Place full protocol chain to the perf nodes' names.

## Why ?
Makes protocol graph debugging more convenient.

| Before| After|
|--------|--------|
| image| image|

### DIFF
--- a/src/ucp/proto/proto_debug.c
+++ b/src/ucp/proto/proto_debug.c
@@ -33,7 +33,7 @@ struct ucp_proto_perf_node {
     ucp_proto_perf_node_type_t                    type;
 
     /* Name of the range */
-    const char                                    *name;
+    char                                          name[UCP_PROTO_DESC_STR_MAX];
 
     /* Description of the range */
     char                                          desc[UCP_PROTO_DESC_STR_MAX];
@@ -512,9 +512,11 @@ ucp_proto_perf_node_new(ucp_proto_perf_node_type_t type, const char *name,
     }
 
     perf_node->type     = type;
-    perf_node->name     = name;
     perf_node->refcount = 1;
     ucs_array_init_dynamic(&perf_node->children);
+
+    ucs_assert(name != NULL);
+    ucs_strncpy_safe(perf_node->name, name, sizeof(perf_node->name));
     ucs_vsnprintf_safe(perf_node->desc, sizeof(perf_node->desc), desc_fmt, ap);
 
     return perf_node;
@@ -899,7 +901,8 @@ ucp_proto_perf_graph_dump_recurs(ucp_proto_perf_node_t *perf_node,
     /* Description */
     if (!ucs_string_is_empty(perf_node->desc)) {
         ucs_string_buffer_appendf(
-                strb, "<font face=\"calibri\" point-size=\"11\">%s<br/></font>",
+                strb, "<font face=\"calibri\" point-size=\"11\">%s"
+                UCP_PROTO_PERF_NODE_NEW_LINE"</font>",
                 perf_node->desc);
     }
 
--- a/src/ucp/proto/proto_debug.h
+++ b/src/ucp/proto/proto_debug.h
@@ -10,6 +10,9 @@
 #include "proto_select.h"
 
 
+/* String to place new line inside performance graph */
+#define UCP_PROTO_PERF_NODE_NEW_LINE "<br/>"
+
 /* Format string to display a protocol time */
 #define UCP_PROTO_TIME_FMT(_time_var) " " #_time_var ": %.2f ns"
 #define UCP_PROTO_TIME_ARG(_time_val) ((_time_val) * 1e9)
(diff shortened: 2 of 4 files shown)


yosefe:
do we really need it? if we look at the child (or descendant) nodes of a protocol node, we could see the protocol chain

ivankochin:
> do we really need it? if we look at the child (or descendant) nodes of a protocol node, we could see the protocol chain

There is no strong need for this improvement, but FMPOV the line of "RTS" labels at the top levels looks ugly and uninformative. One who wants just to understand what particular proto was chosen in some particular case needs to dig deeper down to the graph to find the node with the protocol name (which even doesn't highlighted).

And with that patch top level contains the full chain of protocols which help to understand the whole picture.

brminich:
> > do we really need it? if we look at the child (or descendant) nodes of a protocol node, we could see the protocol chain
> 
> There is no strong need for this improvement, but FMPOV the line of "RTS" labels at the top levels looks ugly and uninformative. One who wants just to understand what particular proto was chosen in some particular case needs to dig deeper down to the graph to find the node with the protocol name (which even doesn't highlighted).
> 
> And with that patch top level contains the full chain of protocols which help to understand the whole picture.

i tend to agree that it looks better than existing naming

yosefe:
Maybe there is a way to make the description more compact, so the size of the tree node will not increase?

yosefe (commented):
in general LGTM

src/ucp/proto/proto_debug.c
```diff
@@ -512,9 +514,13 @@ ucp_proto_perf_node_new(ucp_proto_perf_node_type_t type, const char *name,
     }
 
     perf_node->type     = type;
-    perf_node->name     = name;
     perf_node->refcount = 1;
     ucs_array_init_dynamic(&perf_node->children);
+
+    if (name == NULL) {
+        name = null_string;
```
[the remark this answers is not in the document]
ivankochin: Do you mean the following change?
```suggestion
        name = "(null)";
```

src/ucp/proto/proto_debug.h
```diff
@@ -10,6 +10,9 @@
 #include "proto_select.h"
 
 
+/* String to place new line inside performance graph */
+#define UCP_PROTO_PERF_NODE_NEW_LINE "<BR/>"
```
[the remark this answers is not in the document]
ivankochin: Do you mean the following change?
```suggestion
#define ucp_proto_perf_node_new_line "<BR/>"
```

You are right that is used only once, but I think it worth to create macro for that for code self-documentation purpose

src/ucp/proto/proto_debug.c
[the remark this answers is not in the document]
yosefe: yes

src/ucp/proto/proto.h
```diff
@@ -33,7 +33,6 @@
 /* Threshold for considering two performance values as equal */
 #define UCP_PROTO_PERF_EPSILON     1e-15
```
ivankochin, line 35:
```suggestion


```

brminich (approved):

src/ucp/proto/proto_debug.c
```diff
@@ -512,9 +512,13 @@ ucp_proto_perf_node_new(ucp_proto_perf_node_type_t type, const char *name,
     }
 
     perf_node->type     = type;
-    perf_node->name     = name;
     perf_node->refcount = 1;
     ucs_array_init_dynamic(&perf_node->children);
+
+    if (name == NULL) {
```
yosefe, line 518: in which flow the name can be NULL?

src/ucp/proto/proto_debug.c
```diff
@@ -512,9 +512,13 @@ ucp_proto_perf_node_new(ucp_proto_perf_node_type_t type, const char *name,
...
+        name = "(null)";
+    }
```
ivankochin, line 520:
```suggestion
    ucs_assert(name != NULL);
```

src/ucp/proto/proto_debug.c
ivankochin: Changed to assert

ivankochin: Answering you question: `rndv/rkey_ptr` proto set `params.send_op` to `UCT_EP_OP_LAST` which causes setting `uct_ep_operation_names[UCT_EP_OP_LAST]` which is NULL to name of the `ucp_proto_common_lane_perf_node` result node.

ivankochin:
@yosefe pls review

src/uct/base/uct_iface.c
```diff
@@ -39,7 +39,7 @@ const char *uct_ep_operation_names[] = {
     [UCT_EP_OP_RNDV_ZCOPY]   = "rndv_zcopy",
     [UCT_EP_OP_ATOMIC_POST]  = "atomic_post",
     [UCT_EP_OP_ATOMIC_FETCH] = "atomic_fetch",
-    [UCT_EP_OP_LAST]         = NULL
+    [UCT_EP_OP_LAST]         = "unspecified"
```
yosefe, line 42: maybe we should not create a perf_node when op==UCT_EP_OP_LAST? i.e skip calling ucp_proto_perf_node_new in this case, when name is supposedly NULL

ivankochin: Done.

yosefe (approved):

PR #9930, opened 2024-06-04
UCS/DATASTRUCT: disable memcopy when passing _old_buffer_p to array_grow

roiedanino:
## What
Skip calling `memcpy` when passing `_old_buffer_p` to `ucs_array_append_safe` / `ucs_array_reserve_safe`, making it callers responsibility copying the data to the new allocated buffer.

## Why ?
Some arrays may contain pointers or pointer-based data structures (e.g. linked lists), for which a`memcpy` won't do, and the inner data structures would break.

### DIFF
--- a/src/ucs/datastruct/array.h
+++ b/src/ucs/datastruct/array.h
@@ -179,8 +179,10 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
  * @param _array         Array to reserve buffer for.
  * @param _min_capacity  Minimal capacity to reserve.
  * @param _old_buffer_p  If the array was reallocated, and this parameter is
- *                       non-NULL, the previous buffer will not be released, and
- *                       instread it will be returned in *_old_buffer_p.
+ *                       non-NULL, the previous buffer will not be released,
+ *                       instead it will be returned in *_old_buffer_p,
+ *                       and the caller should copy the contents of the previous buffer 
+ *                       to the new array buffer.
  *
  * @return UCS_OK if successful, UCS_ERR_NO_MEMORY if cannot allocate the array.
  */
--- a/test/gtest/ucs/test_datatype.cc
+++ b/test/gtest/ucs/test_datatype.cc
@@ -1124,6 +1124,7 @@ void test_array::generate_linked_list(int size, simple_elem_t *head)
         simple_elem_t *iter = new simple_elem_t;
         iter->i             = i;
         ucs_list_add_tail(&head->list, &iter->list);
+        /* coverity[leaked_storage] */
     }
 }
 
--- a/test/gtest/ucs/test_datatype.cc
+++ b/test/gtest/ucs/test_datatype.cc
@@ -1098,14 +1098,101 @@ static std::ostream& operator<<(std::ostream& os, const test_value_type_t &v) {
     return os << "<" << v.num1 << "," << v.num2 << ">";
 }
 
+typedef struct {
+    int             i;
+    ucs_list_link_t list;
+} simple_elem_t;
 
 UCS_ARRAY_DECLARE_TYPE(test_2num_t, unsigned, test_value_type_t);
 UCS_ARRAY_DECLARE_TYPE(test_1int_t, size_t, int);
+UCS_ARRAY_DECLARE_TYPE(test_list_links_array_t, unsigned, simple_elem_t);
 
 class test_array : public test_datatype {
 protected:
     void test_fixed(test_1int_t *array, size_t capacity);
+    void generate_linked_list(int size, simple_elem_t *head);
+    void copy_array_of_linked_lists(void *dst, void *src, int size);
+    void cleanup_array_of_linked_lists(test_list_links_array_t *test_array);
 };
+
+/* generate a list of numbers in a certain size */
+void test_array::generate_linked_list(int size, simple_elem_t *head)
+{
+    ucs_list_head_init(&head->list);
+
+    for (int i = 0; i < size; i++) {
+        simple_elem_t *iter = new simple_elem_t;
+        iter->i             = i;
+        ucs_list_add_tail(&head->list, &iter->list);
+    }
+}
+
+void test_array::copy_array_of_linked_lists(void *dst, void *src, int size)
+{
+    simple_elem_t *src_array_of_lists = static_cast<simple_elem_t*>(src);
+    simple_elem_t *dst_array_of_lists = static_cast<simple_elem_t*>(dst);
+
+    for (int i = 0; i < size; ++i) {
+        ucs_list_head_init(&dst_array_of_lists[i].list);
+        ucs_list_splice_tail(&dst_array_of_lists[i].list,
+                             &src_array_of_lists[i].list);
+    }
+}
+
+void test_array::cleanup_array_of_linked_lists(
+        test_list_links_array_t *test_array)
+{
+    simple_elem_t *tmp;
+    simple_elem_t *head;
+    simple_elem_t *elem;
+    simple_elem_t *list_elem;
+    ucs_array_for_each(head, test_array) {
+        ucs_list_for_each_safe(list_elem, tmp, &head->list, list) {
+            elem = (simple_elem_t*)list_elem;
+            ucs_list_del(&elem->list);
+            delete list_elem;
+        }
+    }
+
+    ucs_array_cleanup_dynamic(test_array);
+}
+
+UCS_TEST_F(test_array, dynamic_array_grow_of_list_link_elements) {
+    constexpr size_t NUM_ELEMENTS = 100;
+    constexpr size_t LIST_SIZE    = 200;
+    test_list_links_array_t test_array;
+
+    ucs_array_init_dynamic(&test_array);
+
+    /* create and fill array of linked lists */
+    for (size_t i = 0; i < NUM_ELEMENTS; ++i) {
+        void *old_buffer    = NULL;
+        simple_elem_t *elem = ucs_array_append_safe(&test_array, &old_buffer,
+                                                    FAIL());
+        generate_linked_list(LIST_SIZE, elem);
+
+        if (old_buffer != NULL) {
+            copy_array_of_linked_lists(test_array.buffer, old_buffer, i);
+            ucs_free(old_buffer);
+        }
+        EXPECT_EQ(i + 1, ucs_array_length(&test_array));
+    }
+
+    EXPECT_EQ(NUM_ELEMENTS, ucs_array_length(&test_array));
+
+    /* validate array integrity */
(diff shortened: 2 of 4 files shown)


yosefe (commented):
pls add a gtest with ucs_list element to demonstrate the usage

test/gtest/ucs/test_datatype.cc
```diff
@@ -1098,15 +1098,111 @@ static std::ostream& operator<<(std::ostream& os, const test_value_type_t &v) {
...
+        iter   = (simple_elem_t*)ucs_calloc(1, sizeof(simple_elem_t),
+                                            "list element");
+        iter->i = i;
+        ucs_list_insert_after(&tmp->list, &iter->list);
+        tmp = iter;
+    }
+}
+
+void test_array::copy_array_of_linked_lists(int size, simple_elem_t *dst,
+                                            simple_elem_t *src)
```
iyastreb, line 1139: src can be const

test/gtest/ucs/test_datatype.cc
```diff
@@ -1098,15 +1098,111 @@ static std::ostream& operator<<(std::ostream& os, const test_value_type_t &v) {
...
+    ucs_array_for_each(head, test_array) {
+        ucs_list_for_each(list_elem, &head->list, list) {
+            elem = (simple_elem_t*)list_elem;
+            ucs_list_del(&elem->list);
+            ucs_free(elem);
+        }
+    }
+
+    ucs_array_cleanup_dynamic(test_array);
+
```
iyastreb, line 1161: minor: extra empty line

iyastreb:
General comment.
I understand the importance of this fix, but it still looks a bit strange to me, because of 2 things:
1) IMO it is still error prone to leave the copy part to the caller. It means that other existing API functions would never work properly: ucs_array_append, ucs_array_resize. If you go this way, then maybe during array initialization we may keep a function pointer to the array-resize. Then it will be called automatically each time array is re-allocated with the new address
Still this is not what I would propose..
2) If someone has a collection of objects of complex type, then the obvious choices are:
- just store array of pointers, not objects, so that objects content is never copied
- store them in the linked list
Because it looks like the wrong data structure choice, if you need to copy the whole content of the vector

gleon99:
Per discussion with @roiedanino and @iyastreb , we'll only support "trivial" types without custom user ctor + need to add a comment in docs.

src/ucs/datastruct/array.h
```diff
@@ -17,6 +17,14 @@
 BEGIN_C_DECLS
 
 
+typedef enum {
+    UCS_ARRAY_TYPE_DYNAMIC, /* Array size can be dynamically increased */
+    UCS_ARRAY_TYPE_FIXED,   /* Fixed-size array*/
+    UCS_ARRAY_TYPE_COMPLEX /* Array may contain complex elements such as linked lists or pointers,
```
iyastreb, line 23: I'm not sure this is the best way to solve the problem. There are several drawbacks with this new array type:
- you cannot use regular API for adding elements
- every time you add an element you have to check the buffer and manually copy elements around - this is ugly and error prone

Why don't we impose copy behavior once, during array creation? This way when reallocating array elements we don't need to care about copying - array should call copy impl when needed. IMO we should be able to use one API to add elements, not 2 mutually exclusive.

Why don't we add just an function pointer to the copy function to the array? I guess the only reason is to keep the array size small. Maybe we could do something like that:

```
typedef void (*copy_fn_t)(void *, const void *);

typedef struct {
    copy_fn_t copy;
} array_ops_t;

#define ucs_array_impl(_index_type, _value_type, ops_size)             \
    struct                                                             \
    {                                                                  \
        _value_type *buffer;                                           \
        _index_type length;                                            \
        _index_type capacity : ((CHAR_BIT * sizeof(_index_type)) - 1); \
        uint8_t is_fixed : 1;                                          \
        array_ops_t ops[ops_size]; << Zero sized array by default \
}

#define ucs_array_complex_s -> ucs_array_impl(1) when ops are provided
#define ucs_array_s -> ucs_array_impl(0) when ops are not provided

Then we can distinguish between them by sizeof
```

gleon99: @roiedanino let's discuss on the team meeting today.

1. One of the drawbacks with @iyastreb  approach is that the "copy callback" might be called arbitrary numbers of times, "hidden" inside the grow func, w/o the user having any control or knowledge of it.

1. IMO the enum is somewhat problematic, as it mixes 2 different properties of the array: (1) whether it is fixed size or expandable, (2) whether the element datatype is trivial or complex; and indeed - "complex" **implies** "dynamic" in this context, which is confusing terminology. Maybe something like "generic/user/custom datatype" would be more appropriate naming (akin to `dt_generic.h`.)? Separate to 2 different properties?

iyastreb: Here is POC that better explains the idea, not production ready:

```diff
diff --git a/src/ucs/datastruct/array.h b/src/ucs/datastruct/array.h
index 5f545b5..0349e30 100644
--- a/src/ucs/datastruct/array.h
+++ b/src/ucs/datastruct/array.h
@@ -16,6 +16,21 @@
 
 BEGIN_C_DECLS
 
+typedef void (*array_copy_fn_t)(void *dst, const void *src, size_t length);
+
+typedef struct {
+    array_copy_fn_t copy_fn;
+} ucs_array_ops_t;
+
+#define ucs_array_s_impl(_index_type, _value_type, ops_size) \
+    struct { \
+        _value_type *buffer; \
+        _index_type length; \
+        _index_type capacity : ((CHAR_BIT * sizeof(_index_type)) - 1); \
+        uint8_t     is_fixed : 1; \
+        ucs_array_ops_t ops[ops_size]; \
+    }
+
 
 /**
  * Define an array structure.
@@ -24,13 +39,10 @@ BEGIN_C_DECLS
  * @param _value_type  Type of array value.
 */
 #define ucs_array_s(_index_type, _value_type) \
-    struct { \
-        _value_type *buffer; \
-        _index_type length; \
-        _index_type capacity : ((CHAR_BIT * sizeof(_index_type)) - 1); \
-        uint8_t     is_fixed : 1; \
-    }
+    ucs_array_s_impl(_index_type, _value_type, 0)
 
+#define ucs_array_complex_s(_index_type, _value_type) \
+    ucs_array_s_impl(_index_type, _value_type, 1)
 
 /**
  * Declare an array type.
@@ -206,6 +218,9 @@ ucs_array_old_buffer_set_null(void **old_buffer_p)
             if (_reserve_status == UCS_OK) { \
                 (_array)->capacity = _capacity; \
             } \
+            if (sizeof((_array)->ops)) { \
+                ((_array)->ops[0].copy_fn)((_array)->buffer, _old_buffer_p, _capacity); \
+            } \
         } \
         _reserve_status; \
     })
diff --git a/test/gtest/ucs/test_datatype.cc b/test/gtest/ucs/test_datatype.cc
index 0e7a935..9d7dc76 100644
--- a/test/gtest/ucs/test_datatype.cc
+++ b/test/gtest/ucs/test_datatype.cc
@@ -1288,6 +1288,22 @@ UCS_TEST_F(test_array, add_above_max_capacity) {
     ucs_array_cleanup_dynamic(&test_array);
 }
 
+UCS_TEST_F(test_array, complex)
+{
+    ucs_array_complex_s(uint64_t, int) array;
+    ucs_array_init_dynamic(&array);
+
+    // should be assigned in in ucs_array_init_complex
+    array.ops[0].copy_fn = [](void *dst, const void *src, size_t length) {
+        UCS_TEST_MESSAGE << "SPECIFIC COPY";
+    };
+
+    for (int i = 0; i < 1000; ++i) {
+        int *new_elem = ucs_array_append(&array, FAIL());
+        *new_elem = i;
+    }
+}
+
 class test_datatype_ptr_map : public test_datatype {
```

I also believe we don't need any *_safe functions, instead all the API functions should be safe. With this approach we can achieve that

iyastreb: @gleon99
Right, the idea is that user sets up copy function once, and then it's called by an array impl every time we need to grow an array. What's really important is that we have a uniform array API for all types of data, not 2 mutually exclusive set of functions that you can use only for plain/complex arrays.
Btw copy semantics can be also used if we want to implement generic remove API.

IMO encapsulation of copy semantics ("hidden" inside array) is actually more an advantage - because user specifies it once, and don't have to remember to use it in all the places where it's needed (which is copy-paste and error prone).

roiedanino: I agree and even suggested a similar solution, but even increasing the struct size just for the "extended" array can affect performance

iyastreb: Where from the overhead will come?
I believe the impact on performance is negligible. We increase a complex array size by 8 bytes, and this copy function will be used only in a very rare case when reallocating memory returns new address.

Let me leave all my comments here before my PTO.
- Regarding complexity concern expressed by Leonid: well, I agree that we should keep it simple, and to me that means we should not even support complex types. But if we decide to support complex types in array, then complexity is already there) With current approach we just move complexity to the caller

- For me the biggest advantage is that we can stick to the uniform API, no matter what kind of data is in array.
- There is no memory overhead for plain arrays
- Another thing is that we configure copy function once, and that is implicitly used when needed. No code duplication, less error prone.
- Apart from the copy semantics we can also consider adding a function to destroy an element - that would automatically free elements removed from the container.
- If one day we decide to implement remove array API - then we just reuse existing function pointers

test/gtest/ucs/test_datatype.cc
```diff
@@ -1098,15 +1098,110 @@ static std::ostream& operator<<(std::ostream& os, const test_value_type_t &v) {
     return os << "<" << v.num1 << "," << v.num2 << ">";
 }
 
+typedef struct {
+    int             i;
+    ucs_list_link_t list;
+} simple_elem_t;
```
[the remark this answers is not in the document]
yosefe: ok

test/gtest/ucs/test_datatype.cc
```diff
@@ -1125,37 +1123,38 @@ void test_array::generate_linked_list(int size, simple_elem_t *head)
...
+                             &src_array_of_lists[i].list);
     }
 }
 
-void test_array::cleanup_array_of_linked_lists(test_list_links_array_t *test_array)
+void test_array::cleanup_array_of_linked_lists(
+        test_list_links_array_t *test_array)
 {
-    simple_elem_t *prev = NULL;
+    simple_elem_t *tmp = NULL;
```
yosefe, line 1147: no need to init tmp

yosefe (approved):

yosefe (approved):

yosefe (approved):

PR #9932, opened 2024-06-04
BUILD/ASAN: Preload a single ASAN version

Alexey-Rivkin:
## What
Only use a single ASAN version with LD_PRELOAD.

## Why ?
Some machines had both libasan6 and libasan8 installed, so LD_PRELOAD tried to load them both:
`env LD_PRELOAD=/lib/aarch64-linux-gnu/libasan.so.8 /lib/aarch64-linux-gnu/libasan.so.6:  ...`

Even though libasan.so.6 was present, the linker stopped after finding libasan.so.8 and never reached the path to libasan.so.6. This resulted in error:
`env: ‘/lib/aarch64-linux-gnu/libasan.so.6:’ No such file or directory`

### DIFF
--- a/test/gtest/Makefile.am
+++ b/test/gtest/Makefile.am
@@ -33,7 +33,7 @@ export ASAN_OPTIONS=protect_shadow_gap=0
 
 # LD_PRELOAD puts ASAN first in initial library list which is required for
 # correct ASAN work.
-ASAN_LD_PRELOAD=$(shell ldconfig -p | grep libasan.so | tr ' ' '\n' | grep /)
+ASAN_LD_PRELOAD=$(shell ldconfig -p | grep libasan.so | tr ' ' '\n' | grep / | head -1)
 TEST_ENV=LD_PRELOAD=$(ASAN_LD_PRELOAD):$(LD_PRELOAD)
 endif
 


yosefe (approved):
@ivankochin IMO this is good to have anyway

tvegas1 (approved):

ivankochin:
I thought about writing the similar patch but I don't think that this is a good idea. FMPOV choosing one version asan between several existing can be buggy, since if UCX was compiled with asan 6 and later LD_PRELOAD is set to asan 8 it would cause many failures.

ivankochin:
The best logic that comes to my mind is to report mindfull message when we detect several ASAN so (or somehow understand which one should be chosen)

yosefe:
> The best logic that comes to my mind is to report mindfull message when we detect several ASAN so (or somehow understand which one should be chosen)

Can we tell (using ldd, objdump or similar) with which ASAN version UCX was compiled?

ivankochin:
> > The best logic that comes to my mind is to report mindfull message when we detect several ASAN so (or somehow understand which one should be chosen)
> 
> Can we tell (using ldd, objdump or similar) with which ASAN version UCX was compiled?

I am investigating this right now, but I think it is better to roll back this patch for now to get explicit error in case when several ASAN version exists.

yosefe:
> > > The best logic that comes to my mind is to report mindfull message when we detect several ASAN so (or somehow understand which one should be chosen)
> > 
> > 
> > Can we tell (using ldd, objdump or similar) with which ASAN version UCX was compiled?
> 
> I am investigating this right now, but I think it is better to roll back this patch for now to get explicit error in case when several ASAN version exists.

IMO it's better to avoid false alarms in CI for now (which is achieved by this patch), until we have a better fix

PR #9942, opened 2024-06-07
UCS/ASYNC: Skip released handler in poll missed process

jeynmann:
## What
Skip events in poll missed process if its handler has already been removed.

## Why ?
Async handler might be removed while events missed.
When the handler be called in poll missed process and same memory is reused by the other worker, the handler will try to access wrong resources.

## How ?
Set MPMC queue element's value to -1 if handler removed, then skip the events in poll missed process.

### DIFF
--- a/src/ucs/async/async.c
+++ b/src/ucs/async/async.c
@@ -104,11 +104,27 @@ static inline uint64_t ucs_async_missed_event_pack(int id,
     return ((uint64_t)id << UCS_ASYNC_MISSED_QUEUE_SHIFT) | (uint32_t)events;
 }
 
+static inline int ucs_async_missed_event_unpack_id(uint64_t value)
+{
+    return value >> UCS_ASYNC_MISSED_QUEUE_SHIFT;
+}
+
+static inline int ucs_async_missed_event_unpack_events(uint64_t value)
+{
+    return value & UCS_ASYNC_MISSED_QUEUE_MASK;
+}
+
 static inline void ucs_async_missed_event_unpack(uint64_t value, int *id_p,
                                                  int *events_p)
 {
-    *id_p     = value >> UCS_ASYNC_MISSED_QUEUE_SHIFT;
-    *events_p = value & UCS_ASYNC_MISSED_QUEUE_MASK;
+    *id_p     = ucs_async_missed_event_unpack_id(value);
+    *events_p = ucs_async_missed_event_unpack_events(value);
+}
+
+static int ucs_async_missed_event_is_same(uint64_t value, void *arg)
+{
+    return ucs_async_missed_event_unpack_id(value) ==
+           ((ucs_async_handler_t *)arg)->id;
 }
 
 static void ucs_async_handler_hold(ucs_async_handler_t *handler)
@@ -530,6 +546,7 @@ err:
 ucs_status_t ucs_async_remove_handler(int id, int is_sync)
 {
     ucs_async_handler_t *handler;
+    ucs_async_context_t *async;
     ucs_status_t status;
 
     /* We can't find the async handle mode without taking a read lock, which in
@@ -566,6 +583,12 @@ ucs_status_t ucs_async_remove_handler(int id, int is_sync)
              * for the async handler to complete */
             sched_yield();
         }
+    }
+
+    async = handler->async;
+    if (async != NULL) {
+        ucs_mpmc_queue_remove_if(&async->missed, ucs_async_missed_event_is_same,
+                                 handler);
     }
 
     ucs_async_handler_put(handler);
--- a/src/ucs/datastruct/mpmc.c
+++ b/src/ucs/datastruct/mpmc.c
@@ -17,6 +17,8 @@
 #include <ucs/debug/memtrack_int.h>
 
 
+#define UCS_MPMC_INVALID_VALUE -1
+
 ucs_status_t ucs_mpmc_queue_init(ucs_mpmc_queue_t *mpmc)
 {
     ucs_queue_head_init(&mpmc->queue);
@@ -52,27 +54,44 @@ ucs_status_t ucs_mpmc_queue_push(ucs_mpmc_queue_t *mpmc, uint64_t value)
     return UCS_OK;
 }
 
-
 ucs_status_t ucs_mpmc_queue_pull(ucs_mpmc_queue_t *mpmc, uint64_t *value_p)
 {
+    ucs_status_t status = UCS_ERR_NO_PROGRESS;
     ucs_mpmc_elem_t *elem;
-    ucs_status_t status;
 
     if (ucs_queue_is_empty(&mpmc->queue)) {
-        return UCS_ERR_NO_PROGRESS;
+        return status;
     }
 
     ucs_spin_lock(&mpmc->lock);
-    if (!ucs_queue_is_empty(&mpmc->queue)) {
-        elem     = ucs_queue_pull_elem_non_empty(&mpmc->queue,
-                                                 ucs_mpmc_elem_t, super);
-        *value_p = elem->value;
-        status   = UCS_OK;
+    while (!ucs_queue_is_empty(&mpmc->queue)) {
+        elem = ucs_queue_pull_elem_non_empty(&mpmc->queue, ucs_mpmc_elem_t,
+                                             super);
+        if (elem->value != UCS_MPMC_INVALID_VALUE) {
+            *value_p = elem->value;
+            status   = UCS_OK;
+            ucs_free(elem);
+            break;
+        }
+
         ucs_free(elem);
-    } else {
-        status = UCS_ERR_NO_PROGRESS;
     }
     ucs_spin_unlock(&mpmc->lock);
 
     return status;
+}
+
+void ucs_mpmc_queue_remove_if(ucs_mpmc_queue_t *mpmc,
+                              ucs_mpmc_queue_predicate_t predicate, void *arg)
+{
+    ucs_mpmc_elem_t *elem;
+    ucs_queue_iter_t iter;
+
+    ucs_spin_lock(&mpmc->lock);
+    ucs_queue_for_each_safe(elem, iter, &mpmc->queue, super) {
+        if (predicate(elem->value, arg)) {
+            elem->value = UCS_MPMC_INVALID_VALUE;
+        }
+    }
+    ucs_spin_unlock(&mpmc->lock);
 }
--- a/src/ucs/datastruct/mpmc.h
+++ b/src/ucs/datastruct/mpmc.h
@@ -32,6 +32,17 @@ typedef struct ucs_mpmc_elem {
 
 
 /**
+ * MPMC queue element value predicate.
+ *
+ * @param [in] value MPMC queue element value to check.
+ * @param [in] arg   User-defined argument.
+ *
+ * @return Predicate result value - nonzero means "true", zero means "false".
+ */
+typedef int (*ucs_mpmc_queue_predicate_t)(uint64_t value, void *arg);
+
+
+/**
  * Initialize MPMC queue.
  *
  * @param length   Queue length.
@@ -62,6 +73,19 @@ ucs_status_t ucs_mpmc_queue_push(ucs_mpmc_queue_t *mpmc, uint64_t value);
  *                            or another thread removed the current item.
  */
 ucs_status_t ucs_mpmc_queue_pull(ucs_mpmc_queue_t *mpmc, uint64_t *value_p);
+
+
+/**
+ * Remove all elements from the MPMC queue with the given value for which the
+ * given predicate returns "true" (nonzero) value.
+ * This can be used from any context and any thread.
+ *
+ * @param  [in] mpmc      MPMC queue.
+ * @param  [in] predicate Predicate to check candidates for removal.
+ * @param  [in] arg       User-defined argument for the predicate.
+ */
+void ucs_mpmc_queue_remove_if(ucs_mpmc_queue_t *mpmc,
+                              ucs_mpmc_queue_predicate_t predicate, void *arg);
 
 
 /**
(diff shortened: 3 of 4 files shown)


brminich:
@jeynmann, pls fix commit titles to trigger CI.
 Seems the 1st commit format is incorrect - need to use capitl letters in the very beginning of the title, like described here:
  
 https://github.com/openucx/ucx/wiki/Guidance-for-contributors

brminich:
pls check CI failures. Coverity error looks relevant:
```
+ '[' 1 -gt 0 ']'
+ cov-format-errors --dir /home/swx-azure-svc_azpcontainer/82170-20240610.12/build/cov_build_devel --emacs-style
/__w/1/s/test/gtest/ucs/test_async.cc:388
  Type: Calling risky function (DC.WEAK_CRYPTO)

/__w/1/s/test/gtest/ucs/test_async.cc:388:
  dont_call: "rand" should not be used for security-related applications, because linear congruential algorithms are too easy to break.
/__w/1/s/test/gtest/ucs/test_async.cc:388:
  remediation: Use a compliant random number generator, such as "/dev/random" or "/dev/urandom" on Unix-like systems, and CNG (Cryptography API: Next Generation) on Windows.

+ cp -ar /home/swx-azure-svc_azpcontainer/82170-20240610.12/build/cov_build_devel /__w/1/s/cov_build_devel
+ echo 'not ok 1 Coverity Detected 1 failures'
```

jeynmann:
@evgeny-leksikov
Please help to take a review when you're free.

evgeny-leksikov (approved):

src/ucs/async/async.c
```diff
@@ -104,11 +106,21 @@ static inline uint64_t ucs_async_missed_event_pack(int id,
     return ((uint64_t)id << UCS_ASYNC_MISSED_QUEUE_SHIFT) | (uint32_t)events;
 }
 
+static inline int ucs_async_missed_event_unpack_id(uint64_t value)
+{
+    return value >> UCS_ASYNC_MISSED_QUEUE_SHIFT;
+}
+
+static inline int ucs_async_missed_event_unpack_event(uint64_t value)
```
gleon99, line 114: `event` -> `events`?

src/ucs/async/async.c
```diff
@@ -568,6 +582,17 @@ ucs_status_t ucs_async_remove_handler(int id, int is_sync)
         }
     }
 
+    async = handler->async;
+    if (async != NULL) {
+        ucs_mpmc_queue_block(&async->missed);
+        ucs_mpmc_queue_for_each(elem, &async->missed) {
```
gleon99, line 588: Need to add `ucs_mpmc_queue_for_each` in `.clang-format` (`ForEachMacros`)

src/ucs/datastruct/mpmc.h
```diff
@@ -31,6 +31,29 @@ typedef struct ucs_mpmc_elem {
 } ucs_mpmc_elem_t;
 
 
+/**
+ * Iterate over MPMC queue elements. The queue must not be modified during the
+ * iteration.
+ *
+ * @param elem Variable which will hold point to the MPMC queue element.
```
gleon99, line 38: [out], [in] where needed

src/ucs/datastruct/mpmc.h
```diff
@@ -31,6 +31,29 @@ typedef struct ucs_mpmc_elem {
...
+ * @param mpmc_queue MPMC Queue to iterate on.
+ */
+#define ucs_mpmc_queue_for_each(elem, mpmc_queue) \
+    ucs_queue_for_each(elem, &(mpmc_queue)->queue, super)
+
+
+/**
+ * Lock MPMC queue.
+ */
+void ucs_mpmc_queue_block(ucs_mpmc_queue_t *mpmc);
```
gleon99, line 48: Maybe rename no lock/unlock?

test/gtest/ucs/test_async.cc
```diff
@@ -413,6 +476,21 @@ class test_async_mt : public test_async {
         return (void*)-1;
     }
 
+    static void *repeat_create_destroy_func(void *arg)
+    {
+        test_async_mt *self = reinterpret_cast<test_async_mt*>(arg);
+
+        for (unsigned index = 0; index < NUM_THREADS; ++index) {
```
gleon99, line 483: Would this look better with `std::for_each`?

test/gtest/ucs/test_async.cc
```diff
@@ -349,10 +375,46 @@ class test_async_mt : public test_async {
         return result;
     }
 
-    void spawn() {
+    int repeat_create_destroy_run(unsigned index) {
+        static_assert(std::is_base_of<local_event, LOCAL>::value, "");
+        LOCAL *le;
+        m_ev[index] = le = new LOCAL(GetParam());
```
gleon99, line 381: Is the only reason for using `new` here is that `LOCAL` is template? If so, I think can be eliminated by templating the function.

test/gtest/ucs/test_async.cc
```diff
@@ -95,10 +94,27 @@ class base_event : public base_async {
...
+            push_event();
+        }
+    }
+
     void reset() {
         ucs_async_pipe_drain(&m_event_pipe);
     }
 
+    void create_event() {
+        ucs_status_t status = ucs_async_pipe_create(&m_event_pipe);
```
iyastreb, line 108: Q: What happens when we call this function on already "created" event?
Shouldn't we check if object is already initialized, and if so - release the previous state first?

test/gtest/ucs/test_async.cc
```diff
@@ -349,10 +375,46 @@ class test_async_mt : public test_async {
...
+            le->create_event(); /* same memory could be used by other thread */
+            le->set_event();
+            suspend(20 - wait);
+            le->check_miss();
+        }
+
+        check_is_blocked(le, false);
+
+        int result = le->count();
+        delete le;
```
iyastreb, line 409: pls consider using std::scoped_ptr to get rid of explicit delete
You can also just `return le->count();` in this case without saving in tmp var

test/gtest/ucs/test_async.cc
jeynmann: IMO, double initialization is an error.
The life time of event is managed by `base_event`, since we call create/destroy method inside the constructor/destructor.
``` c
    base_event(ucs_async_mode_t mode) : base_async(mode) {
        create_event();
    }
    ...
    virtual ~base_event() {
        destroy_event();
    }
```
To avoid double initialization, I added assertion inside the `create_event` method, see
``` c
EXPECT_EQ(0, ucs_atomic_cswap32(&m_event_valid, 0, 1));
```
In addition, if we explicitly call `create_event`, which means we want to manage the event by ourself, then we're responsible for `destroy_event`.

src/ucs/async/async.c
yosefe: maybe instead of exposing ucs_mpmc_queue_lock, ucs_mpmc_queue_unlock and foreach, can add ucs_mpmc_remove_if(mpmc, predicate_cb, arg) to take a callback that returns 0/1 whether the given element should be removed?

test/gtest/ucs/test_async.cc
```diff
@@ -95,10 +94,27 @@ class base_event : public base_async {
         ucs_async_pipe_push(&m_event_pipe);
     }
 
+    void push_event_if_valid() {
+        if (m_event_valid == 1) {
```
yosefe, line 98: if (m_event_valid) {

test/gtest/ucs/test_async.cc
```diff
@@ -222,12 +239,20 @@ class local_event : public local,
...
     local_event(ucs_async_mode_t mode) : local(mode), base_event(mode) {
+        set_event();
+    }
+
+    void set_event() {
         set_handler(&m_async);
     }
 
-    ~local_event() {
+    void unset_event() {
```
yosefe, line 249: why set_event/unset_event wrappers?

test/gtest/ucs/test_async.cc
```diff
@@ -308,6 +333,7 @@ public ucs::test_base {
 template<typename LOCAL>
 class test_async_mt : public test_async {
 protected:
+    using ThreadFunc                  = void *(*)(void*);
```
yosefe, line 336: seems redundant type def - used only once

test/gtest/ucs/test_async.cc
```diff
@@ -811,6 +887,19 @@ UCS_TEST_SKIP_COND_P(test_async_event_mt, multithread,
     EXPECT_GE(min_count, exp_min_count);
 }
 
+UCS_TEST_SKIP_COND_P(test_async_event_mt, multithread_create_destroy,
```
yosefe, line 890: how long does this test run?

test/gtest/ucs/test_async.cc
jeynmann: 1. `m_async` is proteted member of `base_event`, cannot be accessed inside test case.
2. If rename `set_event()` to `set_handler()`, it will ***overshadow*** the `set_handler()` of `base_async`, which might make method confusing.

test/gtest/ucs/test_async.cc
jeynmann: 600-800ms from my test.

yosefe:
@jeynmann pls force push the last PR to fix the commit title

src/ucs/async/async.c
```diff
@@ -123,6 +121,11 @@ static inline void ucs_async_missed_event_unpack(uint64_t value, int *id_p,
     *events_p = ucs_async_missed_event_unpack_events(value);
 }
 
+static inline int ucs_async_missed_event_is_same(uint64_t value, void *arg)
+{
+    return ucs_async_missed_event_unpack_id(value) == (int)(uintptr_t)arg;
```
yosefe, line 126: can remote (int)

src/ucs/async/async.c
```diff
@@ -123,6 +121,11 @@ static inline void ucs_async_missed_event_unpack(uint64_t value, int *id_p,
     *events_p = ucs_async_missed_event_unpack_events(value);
 }
 
+static inline int ucs_async_missed_event_is_same(uint64_t value, void *arg)
```
yosefe, line 124: does not need to be inline

src/ucs/async/async.c
```diff
@@ -584,13 +586,8 @@ ucs_status_t ucs_async_remove_handler(int id, int is_sync)
...
     if (async != NULL) {
-        ucs_mpmc_queue_lock(&async->missed);
-        ucs_mpmc_queue_for_each(elem, &async->missed) {
-            if (ucs_async_missed_event_unpack_id(elem->value) == handler->id) {
-                elem->value = UCS_ASYNC_INVALID_EVENT;
-            }
-        }
-        ucs_mpmc_queue_unlock(&async->missed);
+        ucs_mpmc_queue_remove_if(&async->missed, ucs_async_missed_event_is_same,
+                                 (void *)(uintptr_t)handler->id);
```
yosefe, line 590: can pass a pointer to the handler itself as "arg", it could simplify the code a bit

src/ucs/datastruct/mpmc.c
```diff
@@ -86,3 +76,18 @@ ucs_status_t ucs_mpmc_queue_pull(ucs_mpmc_queue_t *mpmc, uint64_t *value_p)
 
     return status;
 }
+
+void ucs_mpmc_queue_remove_if(ucs_mpmc_queue_t *mpmc,
+                              ucs_mpmc_queue_predicate_t pred, void *arg)
```
yosefe, line 81: pls rename "pred" to "predicate"

yosefe (dismissed):

src/ucs/async/async.c
```diff
@@ -638,6 +638,10 @@ void __ucs_async_poll_missed(ucs_async_context_t *async)
             break;
         }
 
+        if (value == UCS_MPMC_INVALID_VALUE) {
```
yosefe, line 641: instead of exposing UCS_MPMC_INVALID_VALUE from mpmc.h, can we update ucs_mpmc_queue_is_empty and ucs_mpmc_queue_pull to ignore this value?

test/gtest/ucs/test_async.cc
gleon99: Then why use `new` at all, and not alloc on stack?

src/ucs/async/async.c
jeynmann: Do we need check it in `ucs_mpmc_queue_is_empty`, too?
This will add lock/unlock inside `ucs_mpmc_queue_is_empty`.
Maybe we can do this check only in `ucs_mpmc_queue_pull`, since it will return `UCS_ERR_NO_PROGRESS` if queue is empty, WDYT?

src/ucs/datastruct/mpmc.c
```diff
@@ -52,25 +54,28 @@ ucs_status_t ucs_mpmc_queue_push(ucs_mpmc_queue_t *mpmc, uint64_t value)
...
     ucs_spin_lock(&mpmc->lock);
-    if (!ucs_queue_is_empty(&mpmc->queue)) {
-        elem     = ucs_queue_pull_elem_non_empty(&mpmc->queue,
-                                                 ucs_mpmc_elem_t, super);
-        *value_p = elem->value;
-        status   = UCS_OK;
+    while (!ucs_queue_is_empty(&mpmc->queue)) {
+        elem = ucs_queue_pull_elem_non_empty(&mpmc->queue, ucs_mpmc_elem_t,
+                                             super);
+
```
yosefe, line 70: remove space line

src/ucs/async/async.c
yosefe: ok

yosefe (approved):

yosefe (approved):

yosefe:
@jeynmann pls squash

PR #9389, opened 2023-09-26
UCS/TOPO: Calculate distance for common NUMA node separately

ivankochin:
## What
Detect the case when two devices have no common PCI root but belong to one NUMA.

## Why ?
BW in cross-NUMA and intra-NUMA cases differs significantly.

## How ?
Check `numa_node` file content inside PCI device path.

PR #7390, opened 2021-09-12
UCS/TOPO: More accurate topology distance estimation

yosefe:
## Why
Better estimation for system topology: the main factor for system distance, when traversing sysfs device tree, is who is the common ancestor - rather than the path distance to it.

When using this PR, this is the topology distance estimation:

Case 1: Supermicro X10DGQ, Intel(R) Xeon(R) CPU E5-2650 v4 @ 2.20GHz

<details>
<summary>(Open details)</summary>

```
$ nvidia-smi topo -m
	GPU0	GPU1	mlx5_0	mlx5_1	CPU Affinity	NUMA Affinity
GPU0	 X 	SYS	PHB	PHB	0-11	0
GPU1	SYS	 X 	SYS	SYS	12-23	1
mlx5_0	PHB	SYS	 X 	PIX		
mlx5_1	PHB	SYS	PIX	 X 		

$ ./build-devel/src/tools/info/ucx_info -T
#
# System topology:
#
# +--------+----------+----------+----------+
# |        |          |          |          |
# |   MB/s |     GPU0 |     GPU1 |   mlx5_0 |
# |        |          |          |          |
# +--------+----------+----------+----------+
# |        |          |          |          |
# |   GPU0 |        - |    220.0 |   3500.0 |
# |        |          |          |          |
# +--------+----------+----------+----------+
# |        |          |          |          |
# |   GPU1 |    220.0 |        - |    220.0 |
# |        |          |          |          |
# +--------+----------+----------+----------+
# |        |          |          |          |
# | mlx5_0 |   3500.0 |    220.0 |        - |
# |        |          |          |          |
# +--------+----------+----------+----------+
#

```
</details>

Case 2: NVIDIA DGX-1 with V100-32, Intel(R) Xeon(R) CPU E5-2698 v4 @ 2.20GHz

<details>
<summary>(Open details)</summary>

```
$ nvidia-smi topo -m
	GPU0	GPU1	GPU2	GPU3	GPU4	GPU5	GPU6	GPU7	mlx5_0	mlx5_1	mlx5_2	mlx5_3	CPU Affinity	NUMA Affinity
GPU0	 X 	NV1	NV1	NV2	NV2	SYS	SYS	SYS	PIX	PHB	SYS	SYS	0-19	0
GPU1	NV1	 X 	NV2	NV1	SYS	NV2	SYS	SYS	PIX	PHB	SYS	SYS	0-19	0
[...]

PR #9947, opened 2024-06-10
UCS/SYS/TOPO: Added options to set distance between devices.

rakhmets:
## What
Added options to set estimated latency and bandwidth values according to the distance within the sysfs device tree.
Moved platform specific values to `ucx.conf` file.

### DIFF
--- a/config/ucx.conf
+++ b/config/ucx.conf
@@ -2,6 +2,7 @@
 CPU model=Grace
 UCX_REG_NONBLOCK_MEM_TYPES=host
 UCX_IB_MLX5_DEVX_OBJECTS=
+UCX_DISTANCE_BW=auto,sys:16500MBs
 
 [Fujitsu ARM]
 CPU vendor=Fujitsu ARM
@@ -12,3 +13,15 @@ UCX_IB_SEND_OVERHEAD=bcopy:5ns,cqe:50ns,db:400ns,wqe_fetch:350ns,wqe_post:100ns
 UCX_MM_SEND_OVERHEAD=am_short:40ns,am_bcopy:220ns
 UCX_MM_RECV_OVERHEAD=am_short:40ns,am_bcopy:220ns
 UCX_RCACHE_OVERHEAD=360ns
+
+[AMD Rome]
+CPU model=Rome
+UCX_DISTANCE_BW=auto,sys:5100MBs
+
+[AMD Milan]
+CPU model=Milan
+UCX_DISTANCE_BW=auto,sys:5100MBs
+
+[AMD Genoa]
+CPU model=Genoa
+UCX_DISTANCE_BW=auto,sys:5100MBs
--- a/src/ucs/config/global_opts.c
+++ b/src/ucs/config/global_opts.c
@@ -74,6 +74,22 @@ static UCS_CONFIG_DEFINE_ARRAY(signo,
                                UCS_CONFIG_TYPE_SIGNO);
 
 
+#define UCS_DISTANCE_KEYS_DESCRIPTION(_field) \
+    {"phb", \
+     "connection traversing PCIe as well as a PCIe Host Bridge (typically " \
+     "the CPU)", \
+     ucs_offsetof(ucs_global_opts_t, dist.phb._field)}, \
+    {"node", \
+     "connection traversing PCIe as well as the interconnect between PCIe " \
+     "Host Bridges within a NUMA node", \
+     ucs_offsetof(ucs_global_opts_t, dist.node._field)}, \
+    {"sys", \
+     "connection traversing PCIe as well as the SMP interconnect between " \
+     "NUMA nodes", \
+     ucs_offsetof(ucs_global_opts_t, dist.sys._field)}, \
+    {NULL}
+
+
 static ucs_config_field_t ucs_global_opts_table[] = {
  {"LOG_LEVEL", "warn",
   "UCS logging level. Messages with a level higher or equal to the selected "
@@ -190,6 +206,16 @@ static ucs_config_field_t ucs_global_opts_table[] = {
   "Comma-separated list of providers for detecting system topology.\n"
   "The list order decides the priority of the providers.",
   ucs_offsetof(ucs_global_opts_t, topo_prio), UCS_CONFIG_TYPE_STRING_ARRAY},
+
+ {"DISTANCE_LAT", "phb:300ns,node:300ns,sys:500ns",
+  "Estimated latency between system devices", 0,
+  UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_TIME,
+                            UCS_DISTANCE_KEYS_DESCRIPTION(latency))},
+
+ {"DISTANCE_BW", "auto",
+  "Estimated bandwidth between system devices", 0,
+  UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_BW,
+                            UCS_DISTANCE_KEYS_DESCRIPTION(bandwidth))},
 
  {NULL}
 };
--- a/src/ucs/config/global_opts.h
+++ b/src/ucs/config/global_opts.h
@@ -12,6 +12,7 @@
 #include <ucs/stats/stats_fwd.h>
 #include <ucs/type/status.h>
 #include <ucs/sys/compiler_def.h>
+#include <ucs/sys/topo/base/topo.h>
 #include <ucs/arch/global_opts.h>
 #include <stddef.h>
 #include <stdio.h>
@@ -147,6 +148,19 @@ typedef struct {
        values. */
     size_t                     rcache_stat_min;
     size_t                     rcache_stat_max;
+
+    /* Estimated latency and bandwidth between devices according to distance
+       within the sysfs device tree */
+    struct {
+        /* Connection traversing PCIe as well as a PCIe Host Bridge */
+        ucs_sys_dev_distance_t phb;
+        /* Connection traversing PCIe as well as the interconnect between PCIe
+           Host Bridges within a NUMA node */
+        ucs_sys_dev_distance_t node;
+        /* Connection traversing PCIe as well as the SMP interconnect between
+           NUMA nodes */
+        ucs_sys_dev_distance_t sys;
+    } dist;
 } ucs_global_opts_t;
 
 
--- a/src/ucs/sys/topo/base/topo.c
+++ b/src/ucs/sys/topo/base/topo.c
@@ -332,26 +332,16 @@ static int ucs_topo_is_pci_root(const char *path)
     return count == strlen(path);
 }
 
-static void ucs_topo_sys_root_distance(ucs_sys_dev_distance_t *distance)
+static void ucs_topo_set_distance(const ucs_sys_dev_distance_t *distance_cfg,
+                                  double bw_auto,
+                                  ucs_sys_dev_distance_t *distance)
 {
-    distance->latency = 500e-9;
-    switch (ucs_arch_get_cpu_model()) {
-    case UCS_CPU_MODEL_AMD_ROME:
-    case UCS_CPU_MODEL_AMD_MILAN:
-    case UCS_CPU_MODEL_AMD_GENOA:
-        distance->bandwidth = 5100 * UCS_MBYTE;
-        break;
-    case UCS_CPU_MODEL_NVIDIA_GRACE:
-        distance->bandwidth = 16500 * UCS_MBYTE;
-        break;
-    default:
-        distance->bandwidth = 220 * UCS_MBYTE;
-        break;
-    }
+    distance->latency   = distance_cfg->latency;
+    distance->bandwidth = UCS_CONFIG_DBL_IS_AUTO(distance_cfg->bandwidth) ?
+                                  bw_auto : distance_cfg->bandwidth;
 }
 
-static void ucs_topo_pci_root_distance(const char *path1, const char *path2,
-                                       ucs_sys_dev_distance_t *distance)
+static double ucs_topo_pci_root_bw(const char *path1, const char *path2)
 {
     size_t path_distance = ucs_path_calc_distance(path1, path2);
 
@@ -359,16 +349,7 @@ static void ucs_topo_pci_root_distance(const char *path1, const char *path2,
                    path_distance);
     ucs_assertv(path_distance > 0, "path1=%s path2=%s", path1, path2);
 
-    /* TODO set latency/bandwidth by CPU model */
-    distance->latency   = 300e-9;
-    distance->bandwidth = ucs_min(3500.0 * UCS_MBYTE,
-                                  (19200.0 * UCS_MBYTE) / path_distance);
-}
-
-static void ucs_topo_common_numa_node_distance(ucs_sys_dev_distance_t *distance)
-{
-    distance->latency   = 300e-9;
-    distance->bandwidth = 17000 * UCS_MBYTE;
+    return ucs_min(3500.0 * UCS_MBYTE, (19200.0 * UCS_MBYTE) / path_distance);
 }
 
 static int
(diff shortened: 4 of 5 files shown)


src/ucs/sys/topo/base/topo.c
```diff
@@ -357,15 +347,15 @@ static void ucs_topo_pci_root_distance(const char *path1, const char *path2,
     ucs_assertv(path_distance > 0, "path1=%s path2=%s", path1, path2);
 
     /* TODO set latency/bandwidth by CPU model */
-    distance->latency   = 300e-9;
-    distance->bandwidth = ucs_min(3500.0 * UCS_MBYTE,
-                                  (19200.0 * UCS_MBYTE) / path_distance);
+    distance->latency   = ucs_global_opts.pci_root_lat;
```
tvegas1, line 350: maybe remove TODO?

src/ucs/config/global_opts.c
```diff
@@ -191,6 +191,39 @@ static ucs_config_field_t ucs_global_opts_table[] = {
   "The list order decides the priority of the providers.",
   ucs_offsetof(ucs_global_opts_t, topo_prio), UCS_CONFIG_TYPE_STRING_ARRAY},
 
+  {"PCI_ROOT_LAT", "300ns",
+   "Estimated latency between devices with the same PCI domain and bus number",
+   ucs_offsetof(ucs_global_opts_t, pci_root_lat), UCS_CONFIG_TYPE_TIME},
+
+  {"PCI_ROOT_BW_BASE", "19200MBs",
+   "Divisible part of the estimated bandwidth between devices with the same\n"
+   "PCI domain and bus number. The path between devices is the divisor of the\n"
```
ivankochin, line 200: Minor:
```suggestion
   "PCI domain and bus number. The length of the path between devices is the divisor of the\n"
```

or 

```suggestion
   "PCI domain and bus number. The path distance between devices is the divisor of the\n"
```

src/ucs/sys/topo/base/topo.c
```diff
@@ -334,17 +334,7 @@ static int ucs_topo_is_pci_root(const char *path)
...
-    case UCS_CPU_MODEL_AMD_ROME:
-    case UCS_CPU_MODEL_AMD_MILAN:
-    case UCS_CPU_MODEL_AMD_GENOA:
-        distance->bandwidth = 5100 * UCS_MBYTE;
-        break;
-    default:
-        distance->bandwidth = 220 * UCS_MBYTE;
-        break;
-    }
+    *distance = ucs_global_opts.sys_root;
```
ivankochin, line 337: Minor: this function is called only once and does only one action so maybe it worth to just make assignment in place of function call and remove this function?

Same for `ucs_topo_common_numa_node_distance`

src/ucs/config/global_opts.c
```diff
@@ -191,6 +191,39 @@ static ucs_config_field_t ucs_global_opts_table[] = {
...
+   "estimation.",
+   ucs_offsetof(ucs_global_opts_t, pci_root_bw_base), UCS_CONFIG_TYPE_BW},
+
+  {"PCI_ROOT_BW_MAX", "3500MBs",
+   "Maximum estimated bandwidth between devices with the same PCI domain and\n"
+   "bus number",
```
ivankochin, line 206: I think current descriptions of `PCI_ROOT_BW_BASE` and `PCI_ROOT_BW_MAX` may be little confusing for person who will read that without knowledge of the formula in `topo.c` since they kind of intersecting.

I want to propose the following (maybe the description still isn't perfect, but that's more about renaming of variable):

```suggestion
  {"PCI_ROOT_BW_UPPER_BOUND", "3500MBs",
   "Upper bound of estimated bandwidth between devices with the same PCI domain and\n"
   "bus number.  Final PCI root bandwidth estimations are capped by this value.",
```

src/ucs/sys/topo/base/topo.c
rakhmets: Removed.

src/ucs/config/global_opts.c
rakhmets: I've updated the description. I think the name `PCI_ROOT_BW_MAX` fits well. Because this is a maximum value. So the name match the description, and the purpose. I changed the other variable. I guess it was more unclear than this one. Please check the new name and description.

src/ucs/sys/topo/base/topo.c
rakhmets: I think it is still relevant. However, we can customize it via `ucx.conf`.

src/ucs/config/global_opts.c
```diff
@@ -191,38 +191,38 @@ static ucs_config_field_t ucs_global_opts_table[] = {
...
-   ucs_offsetof(ucs_global_opts_t, common_numa.bandwidth), UCS_CONFIG_TYPE_BW},
-
-  {"SYS_ROOT_LAT", "500ns",
-   "Estimated latency between devices residing on different NUMA nodes",
-   ucs_offsetof(ucs_global_opts_t, sys_root.latency), UCS_CONFIG_TYPE_TIME},
-
-  {"SYS_ROOT_BW", "220MBs",
-   "Estimated bandwidth between devices residing on different NUMA nodes",
-   ucs_offsetof(ucs_global_opts_t, sys_root.bandwidth), UCS_CONFIG_TYPE_BW},
+ {"PCI_ROOT_BW_K", "19200MBs",
```
ivankochin, line 198: What does `_K` means? Maybe change to `_COEFF`?

ivankochin:
Another idea that came to my mind is that maybe we don't need so many environment variables that do the same thing but for different cases. Maybe better to define something like `DISTANCE_BW=numa:17000MBs` and `DISTANCE_LAT=sys:500ns` using key-value config type as it is done for RNDV inter/intra thresh?

I understand that there can be complications with PCI case since it contains also BW coefficient, but I think it can be a separate config value.

src/ucs/config/global_opts.c
```diff
@@ -191,6 +191,39 @@ static ucs_config_field_t ucs_global_opts_table[] = {
...
+
+  {"COMMON_NUMA_LAT", "300ns",
+   "Estimated latency between devices with different PCI domains and/or bus\n"
+   "numbers residing on the same NUMA node.",
+   ucs_offsetof(ucs_global_opts_t, common_numa.latency), UCS_CONFIG_TYPE_TIME},
+
+  {"COMMON_NUMA_BW", "17000MBs",
+   "Estimated bandwidth between devices with different PCI domains and/or bus\n"
+   "numbers residing on the same NUMA node.",
+   ucs_offsetof(ucs_global_opts_t, common_numa.bandwidth), UCS_CONFIG_TYPE_BW},
```
brminich: maybe call them `NODE_..` or `PHB_..`, like in the output of `nvidia-smi`

src/ucs/config/global_opts.c
```diff
@@ -191,6 +191,39 @@ static ucs_config_field_t ucs_global_opts_table[] = {
   "The list order decides the priority of the providers.",
   ucs_offsetof(ucs_global_opts_t, topo_prio), UCS_CONFIG_TYPE_STRING_ARRAY},
 
+  {"PCI_ROOT_LAT", "300ns",
+   "Estimated latency between devices with the same PCI domain and bus number",
+   ucs_offsetof(ucs_global_opts_t, pci_root_lat), UCS_CONFIG_TYPE_TIME},
+
+  {"PCI_ROOT_BW_BASE", "19200MBs",
```
brminich: maybe call it like `PXB_BW` `PXB_LAT`?

src/ucs/config/global_opts.c
```diff
@@ -191,6 +191,41 @@ static ucs_config_field_t ucs_global_opts_table[] = {
   "The list order decides the priority of the providers.",
   ucs_offsetof(ucs_global_opts_t, topo_prio), UCS_CONFIG_TYPE_STRING_ARRAY},
 
+ {"DISTANCE_LAT", "pci:300ns,numa:300ns,sys:500ns",
```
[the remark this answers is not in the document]
rakhmets: Fixed.

src/ucs/config/global_opts.c
rakhmets: Fixed.

src/ucs/config/global_opts.c
rakhmets: Fixed.

src/ucs/config/global_opts.c
```diff
@@ -191,6 +191,41 @@ static ucs_config_field_t ucs_global_opts_table[] = {
...
+      {"numa",
+       "latency between devices with different PCI domains and/or bus numbers "
+       "residing on the same NUMA node",
+       ucs_offsetof(ucs_global_opts_t, common_numa.latency)},
+      {"sys", "latency between devices residing on different NUMA nodes",
+       ucs_offsetof(ucs_global_opts_t, sys_root.latency)},
+      {NULL}
+ )},
+
+ {"DISTANCE_BW", "pci_k:19200MBs,pci_max:3500MBs,numa:17000MBs,sys:220MBs",
```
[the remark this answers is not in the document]
rakhmets: Fixed. But kept pxb_k/pxb_max for now.

src/ucs/config/global_opts.c
rakhmets: I left the calculation unchanged. The logic was added by this PR https://github.com/openucx/ucx/pull/7390.

src/ucs/sys/topo/base/topo.c
```diff
@@ -412,11 +392,11 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
         return UCS_OK;
     } else if (ucs_topo_is_sys_root(common_path)) {
         if (ucs_topo_is_same_numa_node(device1, device2)) {
-            ucs_topo_common_numa_node_distance(distance);
+            *distance = ucs_global_opts.dist_phb;
```
ivankochin, line 395: If we are talking in terms of `nvidia-smi` tool, this distance was designed to represent not PHB but NODE distance.

src/ucs/config/global_opts.c
```diff
@@ -191,6 +191,41 @@ static ucs_config_field_t ucs_global_opts_table[] = {
...
+       "maximum estimated bandwidth between devices with the same PCI domain "
+       "and bus number",
+       ucs_offsetof(ucs_global_opts_t, dist_pxb_bw_max)},
+      {"phb",
+       "bandwidth between devices with different PCI domains and/or bus "
+       "numbers residing on the same NUMA node",
+       ucs_offsetof(ucs_global_opts_t, dist_phb.bandwidth)},
+      {"sys", "bandwidth between devices residing on different NUMA nodes",
+       ucs_offsetof(ucs_global_opts_t, dist_sys.bandwidth)},
+      {NULL}
```
ivankochin, line 226: Maybe it worth using same with `nvidia-smi` distances' descriptions:

image

src/ucs/sys/topo/base/topo.c
rakhmets: Changed to `node`.

brminich (commented):
i think we need to limit BW of devices connected to the same switch as well according to PCI generation BW.
But it can be done in a separate PR

👍

brminich (approved):

src/ucs/config/global_opts.c
```diff
@@ -191,6 +191,42 @@ static ucs_config_field_t ucs_global_opts_table[] = {
...
+       "Host Bridges within a NUMA node",
+       ucs_offsetof(ucs_global_opts_t, dist_node.latency)},
+      {"sys",
+       "connection traversing PCIe as well as the SMP interconnect between "
+       "NUMA nodes",
+       ucs_offsetof(ucs_global_opts_t, dist_sys.latency)},
+      {NULL}
+ )},
+
+ {"DISTANCE_BW", "phb:auto,node:17000MBs,sys:220MBs",
```
ivankochin, line 212: WDYT does it worth to elaborate what "auto" means in that case?

config/ucx.conf
```diff
@@ -12,3 +13,15 @@ UCX_IB_SEND_OVERHEAD=bcopy:5ns,cqe:50ns,db:400ns,wqe_fetch:350ns,wqe_post:100ns
 UCX_MM_SEND_OVERHEAD=am_short:40ns,am_bcopy:220ns
 UCX_MM_RECV_OVERHEAD=am_short:40ns,am_bcopy:220ns
 UCX_RCACHE_OVERHEAD=360ns
+
+[AMD Rome]
+CPU model=Rome
+UCX_DISTANCE_BW=phb:auto,node:17000MBs,sys:5100MBs
```
yosefe, line 19: could it be something like
`UCX_DISTANCE_BW=auto,node:17000MBs,sys:5100MBs`
so what is not specified explicitly is auto

config/ucx.conf
```diff
@@ -12,3 +13,15 @@ UCX_IB_SEND_OVERHEAD=bcopy:5ns,cqe:50ns,db:400ns,wqe_fetch:350ns,wqe_post:100ns
...
+
+[AMD Milan]
+CPU model=Milan
+UCX_DISTANCE_BW=phb:auto,node:17000MBs,sys:5100MBs
+
+[AMD Genoa]
+CPU model=Genoa
+UCX_DISTANCE_BW=phb:auto,node:17000MBs,sys:5100MBs
```
yosefe, line 27: 1. Do we need something for Intel / ARM?
2. Are these settings based on DGX servers? since it depends if there is PCIe switch on the platform, not only on the system CPU

src/ucs/config/global_opts.c
```diff
@@ -191,6 +191,42 @@ static ucs_config_field_t ucs_global_opts_table[] = {
...
+       ucs_offsetof(ucs_global_opts_t, dist_phb.bandwidth)},
+      {"node",
+       "connection traversing PCIe as well as the interconnect between PCIe "
+       "Host Bridges within a NUMA node",
+       ucs_offsetof(ucs_global_opts_t, dist_node.bandwidth)},
+      {"sys",
+       "connection traversing PCIe as well as the SMP interconnect between "
+       "NUMA nodes",
+       ucs_offsetof(ucs_global_opts_t, dist_sys.bandwidth)},
+      {NULL}
```
yosefe, line 227: can we unite code with DISTANCE_LAT definition, maybe using a macro, to avoid duplicating the description strings?

src/ucs/config/global_opts.h
```diff
@@ -147,6 +148,12 @@ typedef struct {
        values. */
     size_t                     rcache_stat_min;
     size_t                     rcache_stat_max;
+
+    /* Estimated latency and bandwidth between devices according to distance
+       within the sysfs device tree */
+    ucs_sys_dev_distance_t     dist_phb;
+    ucs_sys_dev_distance_t     dist_node;
+    ucs_sys_dev_distance_t     dist_sys;
```
yosefe, line 156: pls add short desc for each field

src/ucs/sys/topo/base/topo.c
```diff
@@ -415,11 +394,11 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
         return UCS_OK;
     } else if (ucs_topo_is_sys_root(common_path)) {
         if (ucs_topo_is_same_numa_node(device1, device2)) {
-            ucs_topo_common_numa_node_distance(distance);
+            *distance = ucs_global_opts.dist_node;
             return UCS_OK;
         }
 
-        ucs_topo_sys_root_distance(distance);
+        *distance = ucs_global_opts.dist_sys;
```
yosefe, line 401: *distance can contain AUTO value here (-2), how is it handled?

config/ucx.conf
rakhmets: 1. Probably. Please note that I have not introduced any new values in this PR. The values are moved to the configuration file from topo.c file.
2. The values were introduced by the following PRs: https://github.com/openucx/ucx/pull/7390 and https://github.com/openucx/ucx/pull/9389.

yosefe: ok

src/ucs/config/global_opts.c
rakhmets: I think it will be redundant. I don't think that we should write all the values and formulas here. auto means that the value will be calculated automatically. so, this should be enough.

src/ucs/sys/topo/base/topo.c
rakhmets: Updated code to handle `auto` values.

src/ucs/config/global_opts.h
```diff
@@ -147,6 +148,17 @@ typedef struct {
...
+    /* Estimated latency and bandwidth between devices according to distance
+       within the sysfs device tree */
+    /* Connection traversing PCIe as well as a PCIe Host Bridge */
+    ucs_sys_dev_distance_t     dist_phb;
+    /* Connection traversing PCIe as well as the interconnect between PCIe Host
+       Bridges within a NUMA node */
+    ucs_sys_dev_distance_t     dist_node;
+    /* Connection traversing PCIe as well as the SMP interconnect between NUMA
+       nodes */
+    ucs_sys_dev_distance_t     dist_sys;
```
brminich, line 161: maybe define a struct `dist`?

src/ucs/sys/topo/base/topo.c
```diff
@@ -338,7 +338,8 @@ static void ucs_topo_set_distance(const ucs_sys_dev_distance_t *distance_cfg,
 {
     distance->latency   = distance_cfg->latency;
     distance->bandwidth = UCS_CONFIG_DBL_IS_AUTO(distance_cfg->bandwidth) ?
-                          bw_auto : distance_cfg->bandwidth;
+                                  bw_auto :
+                                  distance_cfg->bandwidth
```
brminich, line 342: i'd leave it as before. I.e.
```suggestion
                                  bw_auto : distance_cfg->bandwidth
```

brminich (dismissed):

yosefe (approved):

yosefe (approved):

yosefe:
@rakhmets pls squash

PR #9949, opened 2024-06-11
UCP/WIREUP: Support EP reconfiguration for non-reused p2p scenarios

shasson5:
## What
Support partial EP reconfiguration: only non-reused, fully wired-up scenarios. (Lanes must not be reused and must all be p2p).

## Why ?
PR 2 for EP reconfiguration.

## How ?
Add a new function for replacing wireup lane by moving internal AUX EP to the new lane.
The following steps are performed:
1) Set uct_ep to be used for new lane:
   a) for CM flow -> use CM EP.
   b) otherwise -> create a new wireup_ep.
2) Move aux ep from old config to new wireup lane.
3) Delete old wireup lane.
4) Select lane to be used as new wireup lane:
    a) for CM flow -> use CM lane.
    b) otherwise -> select a random lane. (Will be modified in next PRs to look for a non-reused lane).

### DIFF
--- a/src/ucp/wireup/wireup.c
+++ b/src/ucp/wireup/wireup.c
@@ -1355,6 +1355,141 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
     }
 }
 
+static int
+ucp_wireup_are_all_lanes_p2p(ucp_ep_h ep, const ucp_ep_config_key_t *key)
+{
+    ucp_lane_index_t lane;
+    ucp_rsc_index_t rsc_index;
+
+    for (lane = 0; lane < key->num_lanes; ++lane) {
+        rsc_index = ucp_ep_get_rsc_index(ep, lane);
+        ucs_assert(rsc_index != UCP_NULL_RESOURCE);
+
+        if (ucp_worker_is_tl_2iface(ep->worker, rsc_index)) {
+            return 0;
+        }
+    }
+
+    return 1;
+}
+
+static unsigned
+ucp_ep_num_reused_lanes(ucp_ep_h ep, const ucp_lane_index_t *reuse_lane_map)
+{
+    unsigned num_reused = 0;
+    ucp_lane_index_t lane;
+
+    for (lane = 0; lane < ucp_ep_num_lanes(ep); ++lane) {
+        num_reused += (reuse_lane_map[lane] != UCP_NULL_LANE);
+    }
+
+    return num_reused;
+}
+
+static int
+ucp_wireup_check_is_reconfigurable(ucp_ep_h ep,
+                                   const ucp_ep_config_key_t *new_key,
+                                   const ucp_unpacked_address_t *remote_address,
+                                   const unsigned *addr_indices)
+{
+    ucp_lane_index_t reuse_lane_map[UCP_MAX_LANES];
+    const ucp_ep_config_key_t *old_key;
+
+    if (ucp_ep_has_cm_lane(ep)) {
+        return 1;
+    }
+
+    old_key = &ucp_ep_config(ep)->key;
+
+    /* Verify both old/new configurations have only p2p lanes */
+    if (!ucp_wireup_are_all_lanes_p2p(ep, old_key) ||
+        !ucp_wireup_are_all_lanes_p2p(ep, new_key) ||
+        (old_key->num_lanes != new_key->num_lanes)) {
+        return 0;
+    }
+
+    ucp_ep_config_lanes_intersect(old_key, new_key, ep, remote_address,
+                                  addr_indices, reuse_lane_map);
+
+    /* For now, reconfig is supported only if no lanes are reused */
+    return ucp_ep_num_reused_lanes(ep, reuse_lane_map) == 0;
+}
+
+static ucp_lane_index_t
+ucp_wireup_find_non_reused_lane(ucp_ep_h ep, const ucp_ep_config_key_t *key,
+                                const ucp_lane_index_t *reuse_lane_map)
+{
+    if (ucp_ep_has_cm_lane(ep)) {
+        return key->cm_lane;
+    }
+
+    /* Just use first lane, as only non-reused lanes are allowed at the
+     * moment. */
+    ucs_assert(key->num_lanes > 0);
+    ucs_assert(ucp_ep_num_reused_lanes(ep, reuse_lane_map) == 0);
+    return 0;
+}
+
+static ucs_status_t
+ucp_wireup_replace_wireup_msg_lane(ucp_ep_h ep, ucp_ep_config_key_t *key,
+                                   uct_ep_h *new_uct_eps,
+                                   const ucp_lane_index_t *reuse_lane_map)
+{
+    uct_ep_h uct_ep = NULL;
+    ucp_lane_index_t old_lane, new_wireup_lane;
+    ucp_wireup_ep_t *old_wireup_ep, *new_wireup_ep;
+    ucp_rsc_index_t aux_rsc_index;
+    int is_p2p;
+    ucs_status_t status;
+
+    /* Get old wireup lane */
+    old_lane      = ucp_wireup_get_msg_lane(ep, UCP_WIREUP_MSG_REQUEST);
+    old_wireup_ep = ucp_wireup_ep(ucp_ep_get_lane(ep, old_lane));
+    ucs_assert_always(old_wireup_ep != NULL);
+
+    /* Set wireup EP for new configuration's wireup lane */
+    if (ucp_ep_has_cm_lane(ep)) {
+        /* Use existing EP from CM lane */
+        new_wireup_ep = ucp_ep_get_cm_wireup_ep(ep);
+        ucs_assert(new_wireup_ep != NULL);
+        aux_rsc_index = ucp_ep_get_rsc_index(ep, old_lane);
+    } else {
+        /* Create new EP for non-CM flow */
+        status = ucp_wireup_ep_create(ep, &uct_ep);
+        if (status != UCS_OK) {
+            return status;
+        }
+
+        new_wireup_ep = ucp_wireup_ep(uct_ep);
+        aux_rsc_index = ucp_wireup_ep_get_aux_rsc_index(
+                &old_wireup_ep->super.super);
+    }
+
+    ucs_assert(aux_rsc_index != UCP_NULL_RESOURCE);
+    is_p2p = ucp_ep_config_connect_p2p(ep->worker, &ucp_ep_config(ep)->key,
+                                       aux_rsc_index);
+
+    /* Move aux EP to new wireup lane */
+    ucp_wireup_ep_set_aux(new_wireup_ep,
+                          ucp_wireup_ep_extract_msg_ep(old_wireup_ep),
+                          aux_rsc_index, is_p2p);
+
+    /* Remove old wireup_ep as it's not needed anymore.
+     * NOTICE: Next two lines are intentionally not merged with the lane
+     * removal loop in ucp_wireup_check_config_intersect, because of future
+     * support for non-wireup EPs reconfiguration (which will modify this
+     * code). */
+    uct_ep_destroy(&old_wireup_ep->super.super);
+    ucp_ep_set_lane(ep, old_lane, NULL);
+
+    /* Select CM/non-reused lane as new wireup lane */
+    new_wireup_lane = ucp_wireup_find_non_reused_lane(ep, key, reuse_lane_map);
+
+    new_uct_eps[new_wireup_lane] = &new_wireup_ep->super.super;
+    key->wireup_msg_lane         = new_wireup_lane;
+    return UCS_OK;
+}
+
 static ucs_status_t
 ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
                                   const ucp_unpacked_address_t *remote_address,
@@ -1362,9 +1497,8 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
                                   ucp_lane_map_t *connect_lane_bitmap,
                                   ucs_queue_head_t *replay_pending_queue)
 {
-    uct_ep_h new_uct_eps[UCP_MAX_LANES]                = { NULL };
-    ucp_lane_index_t reuse_lane_map[UCP_MAX_LANES]     = { UCP_NULL_LANE };
-    ucp_wireup_ep_t *cm_wireup_ep                      = NULL;
+    uct_ep_h new_uct_eps[UCP_MAX_LANES]            = {NULL};
+    ucp_lane_index_t reuse_lane_map[UCP_MAX_LANES] = {UCP_NULL_LANE};
     ucp_ep_config_key_t *old_key;
     ucp_lane_index_t lane, reuse_lane;
     uct_ep_h uct_ep;
@@ -1372,8 +1506,9 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
 
     *connect_lane_bitmap = UCS_MASK(new_key->num_lanes);
 
-    if (!ucp_ep_has_cm_lane(ep) ||
-        (ep->cfg_index == UCP_WORKER_CFG_INDEX_NULL)) {
+    if ((ep->cfg_index == UCP_WORKER_CFG_INDEX_NULL) ||
+        !ucp_wireup_check_is_reconfigurable(ep, new_key, remote_address,
+                                            addr_indices)) {
         /* nothing to intersect with */
         return ucp_ep_realloc_lanes(ep, new_key->num_lanes);
     }
@@ -1384,43 +1519,36 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
         ucs_assert(ucp_ep_get_lane(ep, lane) != NULL);
     }
 
-    cm_wireup_ep = ucp_ep_get_cm_wireup_ep(ep);
-    ucs_assert(cm_wireup_ep != NULL);
-
     old_key = &ucp_ep_config(ep)->key;
     ucp_ep_config_lanes_intersect(old_key, new_key, ep, remote_address,
                                   addr_indices, reuse_lane_map);
 
-    /* CM lane has to be reused by the new EP configuration */
-    ucs_assert(reuse_lane_map[ucp_ep_get_cm_lane(ep)] != UCP_NULL_LANE);
+    if (ucp_ep_has_cm_lane(ep)) {
+        /* CM lane has to be reused by the new EP configuration */
+        ucs_assert(reuse_lane_map[ucp_ep_get_cm_lane(ep)] != UCP_NULL_LANE);
+        /* wireup lane hasn't been selected by the new configuration: only this
+         * function should select it */
+        ucs_assertv(new_key->wireup_msg_lane == UCP_NULL_LANE,
+                    "new_key->wireup_msg_lane=%u", new_key->wireup_msg_lane);
+    }
+
     /* wireup lane has to be selected for the old configuration */
     ucs_assert(old_key->wireup_msg_lane != UCP_NULL_LANE);
-    /* wireup lane hasn't been selected by the new configuration: only this
-     * function should select it */
-    ucs_assert(new_key->wireup_msg_lane == UCP_NULL_LANE);
 
-    /* set the correct WIREUP MSG lane in case of CM */
+    /* set the correct WIREUP MSG lane */
     reuse_lane = reuse_lane_map[old_key->wireup_msg_lane];
     if (reuse_lane != UCP_NULL_LANE) {
         /* previous wireup lane is part of the new configuration, so reuse it */
         new_key->wireup_msg_lane = reuse_lane;
     } else /* old wireup lane won't be reused */ {
         /* previous wireup lane is not part of new configuration, so add it as
-         * auxiliary endpoint inside cm lane, to be able to continue wireup
-         * messages exchange */
-        new_key->wireup_msg_lane = new_key->cm_lane;
-        reuse_lane               = old_key->wireup_msg_lane;
-        uct_ep                   = ucp_ep_get_lane(ep, reuse_lane);
-        ucp_wireup_ep_set_aux(cm_wireup_ep,
-                              ucp_wireup_ep_extract_next_ep(uct_ep),
-                              old_key->lanes[reuse_lane].rsc_index,
-                              ucp_ep_is_lane_p2p(ep, reuse_lane));
-
-        /* reset the UCT EP from the previous WIREUP lane and destroy its WIREUP EP,
-         * since it's not needed anymore in the new configuration, UCT EP will be
-         * used for sending WIREUP MSGs in the new configuration */
-        uct_ep_destroy(uct_ep);
-        ucp_ep_set_lane(ep, reuse_lane, NULL);
+         * auxiliary endpoint inside CM/non-reused lane, to be able to
+         * continue wireup messages exchange */
+        status = ucp_wireup_replace_wireup_msg_lane(ep, new_key, new_uct_eps,
+                                                    reuse_lane_map);
+        if (status != UCS_OK) {
+            return status;
+        }
     }
 
     /* Need to discard only old lanes that won't be used anymore in the new
@@ -1466,6 +1594,34 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
     return UCS_OK;
 }
 
+static ucs_status_t
+ucp_wireup_try_select_lanes(ucp_ep_h ep, unsigned ep_init_flags,
+                            const ucp_tl_bitmap_t *tl_bitmap,
+                            const ucp_unpacked_address_t *remote_address,
+                            unsigned *addr_indices, ucp_ep_config_key_t *key,
+                            ucp_rsc_index_t *dst_md_storage)
+{
+    ucs_status_t status;
+
+    ucp_ep_config_key_reset(key);
+    ucp_ep_config_key_set_err_mode(key, ep_init_flags);
+    ucp_ep_config_key_init_flags(key, ep_init_flags);
+    key->dst_version  = remote_address->dst_version;
+    key->dst_md_cmpts = dst_md_storage;
+
+    status = ucp_wireup_select_lanes(ep, ep_init_flags, *tl_bitmap,
+                                     remote_address, addr_indices, key, 1);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    /* Get all reachable MDs from full remote address list and join with
+     * current ep configuration
+     */
+    ucp_wireup_get_reachable_mds(ep, ep_init_flags, remote_address, key);
+    return UCS_OK;
+}
+
 ucs_status_t ucp_wireup_init_lanes(ucp_ep_h ep, unsigned ep_init_flags,
                                    const ucp_tl_bitmap_t *local_tl_bitmap,
                                    const ucp_unpacked_address_t *remote_address,
@@ -1482,6 +1638,7 @@ ucs_status_t ucp_wireup_init_lanes(ucp_ep_h ep, unsigned ep_init_flags,
     ucs_status_t status;
     char str[32];
     ucs_queue_head_t replay_pending_queue;
+    ucp_rsc_index_t dst_mds_mem[UCP_MAX_MDS];
 
     tl_bitmap = UCS_STATIC_BITMAP_AND(*local_tl_bitmap,
                                       worker->context->tl_bitmap);
@@ -1489,19 +1646,22 @@ ucs_status_t ucp_wireup_init_lanes(ucp_ep_h ep, unsigned ep_init_flags,
 
     ucs_trace("ep %p: initialize lanes", ep);
     ucs_log_indent(1);
-
-    ucp_ep_config_key_reset(&key);
-    ucp_ep_config_key_set_err_mode(&key, ep_init_flags);
-    ucp_ep_config_key_init_flags(&key, ep_init_flags);
     ucp_wireup_eps_pending_extract(ep, &replay_pending_queue);
 
-    key.dst_version = remote_address->dst_version;
+    status = ucp_wireup_try_select_lanes(ep, ep_init_flags, &tl_bitmap,
+                                         remote_address, addr_indices, &key,
+                                         dst_mds_mem);
+    if (status != UCS_OK) {
+        goto out;
+    }
 
-    /* Allow to choose only the lanes that were already chosen for case
-     * without CM to prevent reconfiguration error.
-     */
     if ((ep->cfg_index != UCP_WORKER_CFG_INDEX_NULL) &&
-        !ucp_ep_has_cm_lane(ep)) {
+        !ucp_ep_config_is_equal(&ucp_ep_config(ep)->key, &key) &&
+        !ucp_wireup_check_is_reconfigurable(ep, &key, remote_address,
+                                            addr_indices)) {
+        /* Allow to choose only the lanes that were already chosen for case
+         * without CM to prevent reconfiguration error.
+         */
         UCS_STATIC_BITMAP_RESET_ALL(&current_tl_bitmap);
         for (lane = 0; lane < ucp_ep_config(ep)->key.num_lanes; ++lane) {
             rsc_idx = ucp_ep_config(ep)->key.lanes[lane].rsc_index;
@@ -1514,23 +1674,22 @@ ucs_status_t ucp_wireup_init_lanes(ucp_ep_h ep, unsigned ep_init_flags,
                                 &worker->context->tl_rscs[rsc_idx].tl_rsc));
         }
         UCS_STATIC_BITMAP_AND_INPLACE(&tl_bitmap, current_tl_bitmap);
+
+        status = ucp_wireup_try_select_lanes(ep, ep_init_flags, &tl_bitmap,
+                                             remote_address, addr_indices, &key,
+                                             dst_mds_mem);
+        if (status != UCS_OK) {
+            goto out;
+        }
     }
 
-    status = ucp_wireup_select_lanes(ep, ep_init_flags, tl_bitmap,
-                                     remote_address, addr_indices, &key, 1);
+    status = ucp_wireup_check_config_intersect(ep, &key, remote_address,
+                                               addr_indices,
+                                               &connect_lane_bitmap,
+                                               &replay_pending_queue);
     if (status != UCS_OK) {
         goto out;
     }
-
-    ucp_wireup_check_config_intersect(ep, &key, remote_address, addr_indices,
-                                      &connect_lane_bitmap,
-                                      &replay_pending_queue);
-
-    /* Get all reachable MDs from full remote address list and join with
-     * current ep configuration
-     */
-    key.dst_md_cmpts = ucs_alloca(sizeof(*key.dst_md_cmpts) * UCP_MAX_MDS);
-    ucp_wireup_get_reachable_mds(ep, ep_init_flags, remote_address, &key);
 
     /* Load new configuration */
     status = ucp_worker_get_ep_config(worker, &key, ep_init_flags,
@@ -1552,8 +1711,8 @@ ucs_status_t ucp_wireup_init_lanes(ucp_ep_h ep, unsigned ep_init_flags,
     cm_idx = ep->ext->cm_idx;
 
     if ((ep->cfg_index != UCP_WORKER_CFG_INDEX_NULL) &&
-        /* reconfiguration is allowed for CM flow */
-        !ucp_ep_has_cm_lane(ep)) {
+        !ucp_wireup_check_is_reconfigurable(ep, &key, remote_address,
+                                            addr_indices)) {
         /*
          * TODO handle a case where we have to change lanes and reconfigure the ep:
          *
--- a/src/ucp/wireup/wireup_ep.c
+++ b/src/ucp/wireup/wireup_ep.c
@@ -52,11 +52,35 @@ static ssize_t ucp_wireup_ep_bcopy_send_func(uct_ep_h uct_ep)
     return UCS_ERR_NO_RESOURCE;
 }
 
+static int ucp_wireup_ep_is_next_ep_active(ucp_wireup_ep_t *wireup_ep)
+{
+    return (wireup_ep->flags & UCP_WIREUP_EP_FLAG_READY) ||
+           (wireup_ep->aux_ep == NULL);
+}
+
+uct_ep_h ucp_wireup_ep_extract_msg_ep(ucp_wireup_ep_t *wireup_ep)
+{
+    uct_ep_h msg_ep;
+
+    ucs_assertv(ucs_queue_is_empty(&wireup_ep->pending_q), "queue_length=%zu",
+                ucs_queue_length(&wireup_ep->pending_q));
+    uct_ep_pending_purge(&wireup_ep->super.super, NULL, NULL);
+
+    if (ucp_wireup_ep_is_next_ep_active(wireup_ep)) {
+        return ucp_wireup_ep_extract_next_ep(&wireup_ep->super.super);
+    }
+
+    msg_ep                   = wireup_ep->aux_ep;
+    wireup_ep->aux_ep        = NULL;
+    wireup_ep->aux_rsc_index = UCP_NULL_RESOURCE;
+    return msg_ep;
+}
+
 uct_ep_h ucp_wireup_ep_get_msg_ep(ucp_wireup_ep_t *wireup_ep)
 {
     uct_ep_h wireup_msg_ep;
 
-    if ((wireup_ep->flags & UCP_WIREUP_EP_FLAG_READY) || (wireup_ep->aux_ep == NULL)) {
+    if (ucp_wireup_ep_is_next_ep_active(wireup_ep)) {
         wireup_msg_ep = wireup_ep->super.uct_ep;
     } else {
         wireup_msg_ep = wireup_ep->aux_ep;
@@ -488,8 +512,9 @@ ucs_status_t ucp_wireup_ep_connect(uct_ep_h uct_ep, unsigned ep_init_flags,
               UCT_TL_RESOURCE_DESC_ARG(
                       &worker->context->tl_rscs[rsc_index].tl_rsc));
 
-    /* we need to create an auxiliary transport only for active messages */
-    if (connect_aux) {
+    /* We need to create an auxiliary transport only for active messages.
+       Skip this step if auxiliary already exists. */
+    if (connect_aux && (wireup_ep->aux_ep == NULL)) {
         status = ucp_wireup_ep_connect_aux(wireup_ep, ep_init_flags,
                                            remote_address);
         if (status != UCS_OK) {
--- a/test/gtest/ucp/test_ucp_ep_reconfig.cc
+++ b/test/gtest/ucp/test_ucp_ep_reconfig.cc
@@ -1,0 +1,246 @@
+/**
+* Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2024. ALL RIGHTS RESERVED.
+*
+* See file LICENSE for terms.
+*/
+
+#include "ucp_test.h"
+#include "common/test.h"
+
+extern "C" {
+#include <ucp/core/ucp_ep.inl>
+#include <ucp/wireup/address.h>
+#include <ucp/wireup/wireup_ep.h>
+}
+
+class test_ucp_ep_reconfig : public ucp_test {
+protected:
+    using address_t = std::pair<void*, ucp_unpacked_address_t>;
+    using address_p = std::unique_ptr<address_t, void (*)(address_t*)>;
+
+    class entity : public ucp_test_base::entity {
+    public:
+        entity(const ucp_test_param &test_param, ucp_config_t *ucp_config,
+               const ucp_worker_params_t &worker_params,
+               const ucp_test_base *test_owner, bool exclude_ifaces) :
+            ucp_test_base::entity(test_param, ucp_config, worker_params,
+                                  test_owner),
+            m_exclude_ifaces(exclude_ifaces)
+        {
+        }
+
+        void connect(const ucp_test_base::entity *other,
+                     const ucp_ep_params_t &ep_params, int ep_idx = 0,
+                     int do_set_ep = 1) override;
+
+        void verify_configuration(const entity &other) const;
+
+        bool is_reconfigured() const
+        {
+            return m_cfg_index != ep()->cfg_index;
+        }
+
+    private:
+        void store_config();
+        ucp_tl_bitmap_t ep_tl_bitmap() const;
+        address_p get_address(const ucp_tl_bitmap_t &tl_bitmap) const;
+        bool is_lane_connected(ucp_ep_h ep, ucp_lane_index_t lane_idx,
+                               const entity &other) const;
+
+        ucp_worker_cfg_index_t m_cfg_index = UCP_WORKER_CFG_INDEX_NULL;
+        std::vector<uct_ep_h>  m_uct_eps;
+        bool                   m_exclude_ifaces;
+    };
+
+    void create_entity(bool push_front, bool exclude_ifaces)
+    {
+        auto e = new entity(GetParam(), m_ucp_config, get_worker_params(), this,
+                            exclude_ifaces);
+
+        if (push_front) {
+            m_entities.push_front(e);
+        } else {
+            m_entities.push_back(e);
+        }
+    }
+
+    void create_entities_and_connect(bool exclude_ifaces)
+    {
+        create_entity(true, exclude_ifaces);
+        create_entity(false, exclude_ifaces);
+        sender().connect(&receiver(), get_ep_params());
+        receiver().connect(&sender(), get_ep_params());
+    }
+
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_TAG, 0, "");
+    }
+
+    void send_recv()
+    {
+        static const unsigned num_iterations = 100;
+        static const size_t msg_sizes[]      = {8, 1024, 16384, UCS_MBYTE};
+        const ucp_request_param_t param      = {
+            .op_attr_mask = UCP_OP_ATTR_FLAG_NO_IMM_CMPL
+        };
+
+        for (auto msg_size : msg_sizes) {
+            for (unsigned i = 0; i < num_iterations; ++i) {
+                mem_buffer sbuf(msg_size, UCS_MEMORY_TYPE_HOST, i);
+                mem_buffer rbuf(msg_size, UCS_MEMORY_TYPE_HOST, ucs::rand());
+
+                void *sreq = ucp_tag_send_nbx(sender().ep(), sbuf.ptr(),
+                                              msg_size, 0, &param);
+                void *rreq = ucp_tag_recv_nbx(receiver().worker(), rbuf.ptr(),
+                                              msg_size, 0, 0, &param);
+                request_wait(rreq);
+                request_wait(sreq);
+                rbuf.pattern_check(i);
+            }
+        }
+    }
+};
+
+void test_ucp_ep_reconfig::entity::store_config()
+{
+    for (ucp_lane_index_t lane = 0; lane < ucp_ep_num_lanes(ep()); ++lane) {
+        uct_ep_h uct_ep = ucp_ep_get_lane(ep(), lane);
+
+        /* Store UCT endpoints in order to compare with next configuration */
+        if (ucp_wireup_ep_test(uct_ep)) {
+            m_uct_eps.push_back(ucp_wireup_ep(uct_ep)->super.uct_ep);
+        } else {
+            m_uct_eps.push_back(uct_ep);
+        }
+    }
+
+    m_cfg_index = ep()->cfg_index;
+}
+
+ucp_tl_bitmap_t test_ucp_ep_reconfig::entity::ep_tl_bitmap() const
+{
+    if (ep() == NULL) {
+        return UCS_STATIC_BITMAP_ZERO_INITIALIZER;
+    }
+
+    ucp_tl_bitmap_t tl_bitmap = UCS_STATIC_BITMAP_ZERO_INITIALIZER;
+    for (ucp_lane_index_t lane = 0; lane < ucp_ep_num_lanes(ep()); ++lane) {
+        UCS_STATIC_BITMAP_SET(&tl_bitmap, ucp_ep_get_rsc_index(ep(), lane));
+    }
+
+    return tl_bitmap;
+}
+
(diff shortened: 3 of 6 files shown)


yosefe:
Errors seem relevant:
```
024-07-17T17:33:17.3872029Z [ RUN      ] tcp/test_ucp_loopback.envelope/29 <tcp/am0_lat>
2024-07-17T17:33:17.3872555Z [1721237597.127722] [swx-rdmz-ucx-gpu-02:52008:0]            sock.c:699  UCX  ERROR   unknown address family: 32769
2024-07-17T17:33:17.3873342Z [1721237597.127750] [swx-rdmz-ucx-gpu-02:52008:0]            sock.c:699  UCX  ERROR   unknown address family: 32769
```

gleon99:
@shasson5  please see `clang-format` output.

test/gtest/ucp/ucp_test.h
```diff
@@ -81,10 +81,11 @@ class ucp_test_base : public ucs::test_base {
                const ucp_worker_params_t& worker_params,
                const ucp_test_base* test_owner);
 
-        ~entity();
+        virtual ~entity();
```
gleon99, line 84: Making these `virtual` can negatively impact overall CI running time. Please measure.

An alternative would be to replace inheritance (`test_ucp_configure::entity`) with composition / friend class.

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+                     int do_set_ep = 1) override;
+
+        void verify(const entity &other) const;
+
+        bool is_reconfigured() const
+        {
+            return m_cfg_index != ep()->cfg_index;
+        }
+
+        static const entity &to_reconfigured(const ucp_test_base::entity &e)
```
gleon99, line 44: It's not "reconfigured"..

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+        std::vector<uct_ep_h>  m_uct_eps;
+    };
+
+    void init() override
+    {
+        ucp_test::init();
+
+        /* Check presence of IB devices using rc_verbs, as all devices must
+         * support it. */
+        if (!has_resource(sender(), "rc_verbs")) {
```
gleon99, line 66: `has_any_transport`..

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        m_entities.clear();
+    }
+
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_values(variants, get_test_variants_feature, 0);
+        add_variant_values(variants, get_test_variants_feature, 1, "excl_if");
```
gleon99, line 77: Clearer name (`exclude_iface`?)

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+            UCS_TEST_SKIP_R("IB transport is not present");
+        }
+
+        m_entities.clear();
+    }
+
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_values(variants, get_test_variants_feature, 0);
```
gleon99, line 76: Maybe have `add_variant_with_value` called here twice, then can remove `get_test_variants_feature` below?

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+        add_variant_values(variants, get_test_variants_feature, 1, "excl_if");
+    }
+
+    static void
+    get_test_variants_feature(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_TAG, 0, "");
+    }
+
+    bool exclude_iface() const
```
gleon99, line 86: `is_`

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+    {
+        return get_variant_value(1);
+    }
+
+    void create_entity()
```
gleon99, line 91: The base class also has such method..

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+
+        for (unsigned i = 0; i < num_iterations; ++i) {
+            void *sreq = ucp_tag_send_nbx(sender().ep(), sbuf.c_str(), msg_size,
+                                          0, &param);
+            void *rreq = ucp_tag_recv_nbx(receiver().worker(),
+                                          (void*)rbuf.c_str(), msg_size, 0, 0,
+                                          &param);
+            request_wait(rreq);
+            request_wait(sreq);
+            EXPECT_EQ(sbuf, rbuf);
```
gleon99, line 117: After the first iteration this comparison is meaningless..

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    return tl_bitmap;
+}
+
+void test_ucp_reconfigure::entity::connect(const ucp_test_base::entity *other,
+                                           const ucp_ep_params_t &ep_params,
+                                           int ep_idx, int do_set_ep)
+{
+    auto r_test      = static_cast<const test_ucp_reconfigure*>(m_test);
```
gleon99, line 157: Maybe `*self` (consistencty with similar places)

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+                                  test_owner),
+            m_cfg_index(UCP_WORKER_CFG_INDEX_NULL)
+        {
+        }
+
+        void connect(const ucp_test_base::entity *other,
+                     const ucp_ep_params_t &ep_params, int ep_idx = 0,
+                     int do_set_ep = 1) override;
+
+        void verify(const entity &other) const;
```
gleon99, line 37: Name unclear (what gets verified?)

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+    ASSERT_UCS_OK(ucp_address_unpack(worker(), address->first, flags,
+                                     &address->second));
+    return address;
+}
+
+UCS_TEST_P(test_ucp_reconfigure, basic)
+{
+    create_entity();
+    create_entity();
+    send_recv();
```
gleon99, line 256: Maybe the `connect` should be in something like `create_entities_and_connect`, but definitely separated from `send_recv`.

To support future more sophisticated scenarios.

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+    } else {
+        EXPECT_FALSE(e1.is_reconfigured());
+        EXPECT_FALSE(e2.is_reconfigured());
+    }
+
+    e1.verify(e2);
+    e2.verify(e1);
+}
+
+UCP_INSTANTIATE_TEST_CASE_TLS(test_ucp_reconfigure, rc, "rc_v,rc_x");
```
gleon99, line 272: `"rc"`?

src/ucp/wireup/wireup.c
```diff
@@ -1343,25 +1343,166 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
     }
 }
 
+static int
+ucp_wireup_are_all_lanes_p2p(ucp_ep_h ep, const ucp_ep_config_key_t *key)
```
gleon99, line 1347: On the one hand, this func is only used in `wireup`.
On the other hand, there's nothing "wireup-specific" here, it can also reside in `ucp_ep.c` and be used from other places, if needed, and it is anywhere not inlined. wdyt?

src/ucp/wireup/wireup.c
```diff
@@ -1343,25 +1343,166 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+        if (reuse_lane_map[lane] != UCP_NULL_LANE) {
+            return 0;
+        }
+    }
+
+    return 1;
+}
+
+/* Check whether resource restriction is required */
+static int ucp_wireup_should_restrict_resources(
```
gleon99, line 1400: `restrict` is unclear. What does it mean? ++docs and/or rename, please

src/ucp/wireup/wireup_ep.c
```diff
@@ -52,11 +52,31 @@ static ssize_t ucp_wireup_ep_bcopy_send_func(uct_ep_h uct_ep)
     return UCS_ERR_NO_RESOURCE;
 }
 
+static int ucp_wireup_ep_is_next_ep_active(ucp_wireup_ep_t *wireup_ep)
+{
+    return (wireup_ep->flags & UCP_WIREUP_EP_FLAG_READY) ||
+           (wireup_ep->aux_ep == NULL);
+}
+
+uct_ep_h ucp_wireup_ep_extract_msg_ep(ucp_wireup_ep_t *wireup_ep)
```
gleon99, line 61: `extract` is unclear.
How is it different from `get`? What does it add?

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: what if the first iteration fails to fill the buffer and the others succeed?

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: base class initializes ucp_test_base::entity instance instead of test_ucp_reconfigure::entity

src/ucp/wireup/wireup.c
shasson5: bare in mind that it's a temp function which will be removed next PRs, so not sure it's relevant

test/gtest/ucp/ucp_test.h
shasson5: from what I measured it seems perf impact is negligible.

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,272 @@
...
+            std::unique_ptr<address_pair_t, void (*)(address_pair_t*)>;
+
+    class entity : public ucp_test_base::entity {
+    public:
+        entity(const ucp_test_param &test_params, ucp_config_t *ucp_config,
+               const ucp_worker_params_t &worker_params,
+               const ucp_test *test_owner) :
+            ucp_test_base::entity(test_params, ucp_config, worker_params,
+                                  test_owner),
+            m_cfg_index(UCP_WORKER_CFG_INDEX_NULL)
```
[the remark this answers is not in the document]
shasson5: base class ctor will create base class object, instead of derived object

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: `has_any_transport` checks test params, `has_resource` checks device list.

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: I think better keep variant names short, to avoid too long lines when gtest prints the full test name (possibly combined with other variant axis)

src/ucp/wireup/wireup_ep.c
shasson5: on the same weight as `ucp_wireup_ep_extract_next_ep`

src/ucp/wireup/wireup.c
```diff
@@ -1372,43 +1512,34 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
...
                                   addr_indices, reuse_lane_map);
 
-    /* CM lane has to be re-used by the new EP configuration */
-    ucs_assert(reuse_lane_map[ucp_ep_get_cm_lane(ep)] != UCP_NULL_LANE);
+    if (ucp_ep_has_cm_lane(ep)) {
+        /* CM lane has to be re-used by the new EP configuration */
+        ucs_assert(reuse_lane_map[ucp_ep_get_cm_lane(ep)] != UCP_NULL_LANE);
+        /* wireup lane hasn't been selected by the new configuration: only this
+         * function should select it */
+        ucs_assert(new_key->wireup_msg_lane == UCP_NULL_LANE);
```
brminich, line 1524:
```suggestion
        ucs_assertv(new_key->wireup_msg_lane == UCP_NULL_LANE, "new_key->wireup_msg_lane=%u",
                    new_key->wireup_msg_lane);
```

shasson5: that's code from master. 
Was marked as "new added" code for some wrong reason.

src/ucp/wireup/wireup.c
```diff
@@ -1343,25 +1343,165 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+     * NOTICE: Next two lines are intentionally not merged with the lane
+     * removal loop in ucp_wireup_check_config_intersect, because of future
+     * support for non-wireup EPs reconfiguration (which will modify this
+     * code). */
+    uct_ep_destroy(&old_ep->super.super);
+    ucp_ep_set_lane(ep, old_lane, NULL);
+
+    /* Select CM/non-reused lane as new wireup lane */
+    new_lane              = ucp_wireup_find_non_reused_lane(ep, key);
+    new_uct_eps[new_lane] = &new_ep->super.super;
```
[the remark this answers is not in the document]
shasson5: I think it may complicate the flow, because you need to return [new_lane,new_ep] from this function and then add extra "if" block in `ucp_wireup_check_config_intersect` in order to properly initialize `new_uct_eps`.

And anyway the value of `new_uct_eps` will not be overwritten, because in `ucp_wireup_check_config_intersect` we only handle "reused" lanes.

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,184 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+        ucs_assert(new_ep != NULL);
+        aux_rsc_index = ucp_ep_get_rsc_index(ep, old_lane);
+        is_p2p        = ucp_ep_is_lane_p2p(ep, old_lane);
+    } else {
+        /* Create new EP for non-CM flow */
+        status = ucp_wireup_ep_create(ep, &uct_ep);
+        if (status != UCS_OK) {
+            return status;
+        }
+        new_ep        = ucp_wireup_ep(uct_ep);
```
evgeny-leksikov, line 1476: pls add empty line after `}`

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,184 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+
+    /* Move aux EP to new wireup lane */
+    msg_ep = ucp_wireup_ep_extract_msg_ep(old_ep, &pending_queue);
+    ucp_wireup_ep_set_aux(new_ep, msg_ep, aux_rsc_index, is_p2p);
+
+    /* Add wireup messages to pending queue */
+    ucs_queue_for_each_extract(req, &pending_queue, send.uct.priv, 1) {
+        status = uct_ep_pending_add(&new_ep->super.super, &req->send.uct, 0);
+        if (status != UCS_OK) {
+            goto destroy_ep;
```
evgeny-leksikov, line 1492: rest of pending queue wont be handled in case of error

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,184 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+    new_lane              = ucp_wireup_find_non_reused_lane(ep, key);
+    new_uct_eps[new_lane] = &new_ep->super.super;
+    key->wireup_msg_lane  = new_lane;
+    return UCS_OK;
+
+destroy_ep:
+    if (uct_ep != NULL) {
+        uct_ep_destroy(uct_ep);
+    }
+    return status;
```
evgeny-leksikov, line 1514: empty line

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+#include "ucp_test.h"
+#include "common/test.h"
+
+extern "C" {
+#include <ucp/core/ucp_ep.inl>
+#include <ucp/wireup/address.h>
+#include <ucp/wireup/wireup_ep.h>
+}
+
+class test_ucp_reconfigure : public ucp_test {
```
evgeny-leksikov, line 16: `test_ucp_ep_reconfigure`

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    EXPECT_EQ(reused, is_reconfigured() ? 0 : num_lanes);
+}
+
+test_ucp_reconfigure::address_p
+test_ucp_reconfigure::entity::get_address(bool ep_only) const
+{
+    const ucp_tl_bitmap_t tl_bitmap = ep_only ? ep_tl_bitmap() :
+                                                ucp_tl_bitmap_max;
```
evgeny-leksikov, line 227: maybe move `tl_bitmap` to parameter? IMO it clearer than "ep_only" flag.

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+
+    void cleanup()
+    {
+        m_sender.cleanup();
+        m_receiver.cleanup();
+        ucp_test::cleanup();
+    }
+
+    entity m_sender;
+    entity m_receiver;
```
evgeny-leksikov, line 84: can `entities_storage` class be reused?

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant_with_value(variants, UCP_FEATURE_TAG, 0, "");
+        add_variant_with_value(variants, UCP_FEATURE_TAG, 1, "exclude_iface");
+    }
+
+    bool is_exclude_iface() const
+    {
+        return get_variant_value();
```
evgeny-leksikov, line 95: pls compare to exact value to avoid wrong behavior after adding new variant

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    void send_recv()
+    {
+        static constexpr unsigned num_iterations = 100;
+        static constexpr size_t msg_size         = 16 * UCS_KBYTE;
```
evgeny-leksikov, line 101: why `constexpr`? maybe just `const`?

test/gtest/ucp/test_ucp_reconfigure.cc
evgeny-leksikov: reuse `pattern_fill`/`pattern_check`

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+}
+
+ucp_tl_bitmap_t test_ucp_reconfigure::entity::ep_tl_bitmap() const
+{
+    ucp_tl_bitmap_t tl_bitmap = UCS_STATIC_BITMAP_ZERO_INITIALIZER;
+
+    if (ep() == NULL) {
+        return tl_bitmap;
+    }
+
```
evgeny-leksikov, line 144: minor:
```suggestion
    if (ep() == NULL) {
        return UCS_STATIC_BITMAP_ZERO_INITIALIZER;
    }

    ucp_tl_bitmap_t tl_bitmap = UCS_STATIC_BITMAP_ZERO_INITIALIZER;
```

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+                                               &m_ep));
+
+    m_ep->conn_sn = 0;
+    ASSERT_TRUE(ucp_ep_match_insert(worker(), m_ep, worker_addr->second.uuid,
+                                    m_ep->conn_sn, UCS_CONN_MATCH_QUEUE_EXP));
+
+    ASSERT_UCS_OK(ucp_wireup_send_request(m_ep));
+    UCS_ASYNC_UNBLOCK(&worker()->async);
+
+    store_config();
```
evgeny-leksikov, line 176: should not store be under lock? AFAIU ep can be asynchronously reconfigured once wireup started

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+}
+
+bool test_ucp_reconfigure::entity::has_matching_lane(ucp_ep_h ep,
+                                                     ucp_lane_index_t lane_idx,
+                                                     const entity &other) const
+{
+    const auto lane     = &ucp_ep_config(ep)->key.lanes[lane_idx];
+    auto context        = m_worker_entity->ucph();
+    const auto resource = &context->tl_rscs[ucp_ep_get_rsc_index(ep, lane_idx)];
+    auto addr           = other.get_address(true);
```
evgeny-leksikov, line 186: as I mentioned above `other.get_address(tl_bit_map);` looks simpler and might be extended with smth like
`other.get_address(other.get_eps_tl_bit_map());`

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+        const entity &other) const
+{
+    unsigned reused                  = 0;
+    const ucp_lane_index_t num_lanes = ucp_ep_num_lanes(ep());
+
+    EXPECT_EQ(num_lanes, ucp_ep_num_lanes(other.ep()));
+
+    for (ucp_lane_index_t lane = 0; lane < num_lanes; ++lane) {
+        /* Verify local and remote lanes are identical */
+        EXPECT_TRUE(has_matching_lane(ep(), lane, other));
```
evgeny-leksikov, line 212: maybe return "matching lane" instead of `bool`? then add it into std::set which can test by size, uniq and values

test/gtest/ucp/test_ucp_reconfigure.cc
```diff
@@ -0,0 +1,264 @@
...
+{
+    const ucp_tl_bitmap_t tl_bitmap = ep_only ? ep_tl_bitmap() :
+                                                ucp_tl_bitmap_max;
+    size_t addr_len;
+
+    address_p address(new address_t(nullptr, {0}), [](address_t *addr) {
+        ucs_free(addr->first);
+        ucs_free(addr->second.address_list);
+        delete addr;
+    });
```
evgeny-leksikov, line 234: it looks unreadable, maybe add a wrapper class with clear API?

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,184 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+
+static ucp_lane_index_t
+ucp_wireup_find_non_reused_lane(ucp_ep_h ep, const ucp_ep_config_key_t *key)
+{
+    if (ucp_ep_has_cm_lane(ep)) {
+        return key->cm_lane;
+    }
+
+    /* Just use first lane, as only non-reused lanes are allowed at the
+     * moment. */
```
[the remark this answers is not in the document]
shasson5: in line 1402 we verify that no lanes are reused.

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,184 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+    ucp_rsc_index_t aux_rsc_index;
+    ucs_queue_head_t pending_queue;
+    int is_p2p;
+    ucs_status_t status;
+    ucp_request_t *req;
+
+    /* Get old wireup lane */
+    old_lane = ucp_wireup_get_msg_lane(ep, UCP_WIREUP_MSG_REQUEST);
+    old_ep   = ucp_wireup_ep(ucp_ep_get_lane(ep, old_lane));
+    ucs_assert(old_ep != NULL);
```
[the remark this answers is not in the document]
shasson5: this will make it present in release config, why is that required?

src/ucp/wireup/wireup.c
shasson5: why is it relevant?

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: IIUC, it cannot be used because `test_ucp_reconfigure::entity` does not derive from `ucp_test_base::entity`

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: better that it would be initialized at compile time

src/ucp/wireup/wireup.c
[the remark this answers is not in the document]
evgeny-leksikov: so this function can work unexpectedly and requires extra check outside, can we test in here?

src/ucp/wireup/wireup.c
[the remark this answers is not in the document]
evgeny-leksikov: 1) if you add the assertion for debug then it makes sense, right?
2) this is not a fast path and extra branch is not critical

src/ucp/wireup/wireup.c
evgeny-leksikov: in case of error it leads to memory leak and probably hang due to uncompleted lost requests

test/gtest/ucp/test_ucp_reconfigure.cc
evgeny-leksikov: 1) BTW, what's the reason to add new entity class? do we really need it?
2) `entities_storage` is a template, so it can be used with any type of entities.

test/gtest/ucp/test_ucp_reconfigure.cc
evgeny-leksikov: local constant will be optimized the same way by compiler, just replaced with the value calculated in compile time

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: why unreachable? IMO wrapper class is overhead here

evgeny-leksikov: ok, up to other reviewers, it looks ugly to me

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: std::set will not detect repetitions

evgeny-leksikov: exactly, `std::set` cannot contain duplicates, why is it possible? should not all lanes be unique?

shasson5: they are not unique because of paths

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: 1. it's required for controlling tl_bitmap which creates asymmetric configs in local vs remote.
2. `entities_storage` is already used by `ucp_test` which is the base class of `test_ucp_reconfigure`, so `test_ucp_reconfigure` cannot add another `entities_storage` with a different entity type

test/gtest/ucp/test_ucp_ep_reconfigure.cc
```diff
@@ -92,33 +92,36 @@ class test_ucp_reconfigure : public ucp_test {
 
     bool is_exclude_iface() const
     {
-        return get_variant_value();
+        return get_variant_value(0);
```
evgeny-leksikov, line 95: get_variant_value() == 1, also pls add an enum to replace magic numbers

test/gtest/ucp/test_ucp_ep_reconfigure.cc
```diff
@@ -92,33 +92,36 @@ class test_ucp_reconfigure : public ucp_test {
...
+        static const unsigned num_iterations = 100;
+        static const size_t msg_size         = 16 * UCS_KBYTE;
+        static const uint64_t seed           = 0xdeadbeef;
+        const ucp_request_param_t param      = {
             .op_attr_mask = UCP_OP_ATTR_FLAG_NO_IMM_CMPL
         };
 
         for (unsigned i = 0; i < num_iterations; ++i) {
-            std::string sbuf(msg_size, 'a'), rbuf(msg_size, 'b');
+            mem_buffer sbuf(msg_size, UCS_MEMORY_TYPE_HOST, seed);
```
evgeny-leksikov, line 108: maybe `seed + 1` to add variability to test

test/gtest/ucp/test_ucp_reconfigure.cc
evgeny-leksikov: 1. can existing entity class be extended with missed functionality?
2. so we have common `entities_storage` and local members of completely different type of entities, IMO need to reuse base class since there is a set of methods which is based on common entities storage and widely used in tests, like is_loopback(), skip_<smth>(), etc

test/gtest/ucp/test_ucp_reconfigure.cc
evgeny-leksikov: then how do you control if this is a re-used lane or newly created but with the same transport?

src/ucp/wireup/wireup.c
```diff
@@ -1438,12 +1446,24 @@ ucp_wireup_find_non_reused_lane(ucp_ep_h ep, const ucp_ep_config_key_t *key)
...
-static ucs_status_t ucp_wireup_replace_wireup_msg_lane(ucp_ep_h ep,
-                                                       ucp_ep_config_key_t *key,
-                                                       uct_ep_h *new_uct_eps)
+static void ucp_wireup_purge_pending_requests(ucs_queue_head_t *pending_queue)
+{
+    ucp_request_t *req;
+
+    ucs_queue_for_each_extract(req, pending_queue, send.uct.priv, 1) {
+        ucs_free(req->send.buffer);
+        ucp_request_mem_free(req);
```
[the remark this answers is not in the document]
shasson5: these are internal wireup requests

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: 1. IMO it's better not to extend the existing entity class only to support this specific case. better it's all contained in test_reconfigure.
2. OK

test/gtest/ucp/test_ucp_reconfigure.cc
shasson5: In wireup.c we count how many times we are using the same address entry and expect it will be max_paths.
In test_reconfigure I cannot keep counter per address entry in std::set.

shasson5: It can be done with std::map or std::pair but it's a bit too complicated for now IMO

test/gtest/ucp/test_ucp_ep_reconfigure.cc
shasson5: you mean seed +i?

test/gtest/ucp/test_ucp_ep_reconfigure.cc
```diff
@@ -64,6 +64,11 @@ class test_ucp_ep_reconfigure : public ucp_test {
         ucp_ep_h                     m_ep            = nullptr;
     };
 
+    typedef enum {
+        METHOD_NONE,
+        METHOD_EXCLUDE_IFACES
+    } method_t;
```
evgeny-leksikov, line 70: method -> variant
none -> default

test/gtest/ucp/test_ucp_ep_reconfigure.cc
```diff
@@ -105,17 +111,17 @@ class test_ucp_ep_reconfigure : public ucp_test {
         };
 
         for (unsigned i = 0; i < num_iterations; ++i) {
-            mem_buffer sbuf(msg_size, UCS_MEMORY_TYPE_HOST, seed);
-            std::string rbuf(msg_size, 'a');
+            mem_buffer sbuf(msg_size, UCS_MEMORY_TYPE_HOST, seed + i);
+            mem_buffer rbuf(msg_size, UCS_MEMORY_TYPE_HOST);
+            mem_buffer::memset(rbuf.ptr(), msg_size, 0, UCS_MEMORY_TYPE_HOST);
```
evgeny-leksikov, line 116: maybe `mem_buffer rbuf(msg_size, UCS_MEMORY_TYPE_HOST, ucs::rand());` ?

src/ucp/wireup/wireup.c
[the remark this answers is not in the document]
shasson5: there are 2 types of pending queue for wireup ep:
1) queue for user reqs (handled in line 1671)
2) queue for wireup messages (which is internal).

test/gtest/ucp/test_ucp_ep_reconfigure.cc
```diff
@@ -0,0 +1,268 @@
...
+
+            void *sreq = ucp_tag_send_nbx(m_sender.ep(), sbuf.ptr(), msg_size,
+                                          0, &param);
+            void *rreq = ucp_tag_recv_nbx(m_receiver.worker(), rbuf.ptr(),
+                                          msg_size, 0, 0, &param);
+            request_wait(rreq);
+            request_wait(sreq);
+            mem_buffer::pattern_check(rbuf.ptr(), msg_size, ucs::rand(),
+                                      UCS_MEMORY_TYPE_HOST);
```
evgeny-leksikov, line 124:
```suggestion
            mem_buffer sbuf(msg_size, UCS_MEMORY_TYPE_HOST, i);
            mem_buffer rbuf(msg_size, UCS_MEMORY_TYPE_HOST, ucs::rand());

            void *sreq = ucp_tag_send_nbx(m_sender.ep(), sbuf.ptr(), msg_size,
                                          0, &param);
            void *rreq = ucp_tag_recv_nbx(m_receiver.worker(), rbuf.ptr(),
                                          msg_size, 0, 0, &param);
            request_wait(rreq);
            request_wait(sreq);
            rbuf.pattern_check(i);
```

src/ucp/wireup/wireup.c
gleon99: This code is indeed from master. Marked as diff because of indentation (`if` block).
IMO the suggestions makes sense.

test/gtest/ucp/test_ucp_reconfigure.cc
gleon99: As a compromise, we can keep this code but add some doc above.

gleon99:
@yosefe please look, as discussed

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,208 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+    } else {
+        /* Create new EP for non-CM flow */
+        status = ucp_wireup_ep_create(ep, &uct_ep);
+        if (status != UCS_OK) {
+            return status;
+        }
+
+        new_ep        = ucp_wireup_ep(uct_ep);
+        aux_rsc_index = ucp_wireup_ep_get_aux_rsc_index(&old_ep->super.super);
+        is_p2p        = old_ep->flags & UCP_WIREUP_EP_FLAG_AUX_P2P;
```
yosefe, line 1499: can we use ucp_ep_config_connect_p2p to calculate is_p2p, in both cm and non-cm case?
we can test it by 
`ucs_assert(is_p2p == ucp_ep_config_connect_p2p(worker, okd_key, aux_rsc_index)` after the if/else and run tests

src/ucp/wireup/wireup_ep.c
```diff
@@ -52,11 +52,50 @@ static ssize_t ucp_wireup_ep_bcopy_send_func(uct_ep_h uct_ep)
...
+                                      ucs_queue_head_t *pending_queue)
+{
+    uct_ep_h msg_ep = ucp_wireup_ep_get_msg_ep(wireup_ep);
+
+    uct_ep_pending_purge(&wireup_ep->super.super,
+                         ucp_wireup_ep_append_pending_cb, pending_queue);
+    ucs_assert(wireup_ep->pending_count == 0);
+
+    if (ucp_wireup_ep_is_next_ep_active(wireup_ep)) {
+        ucp_proxy_ep_set_uct_ep(&wireup_ep->super, NULL, 0, UCP_NULL_RESOURCE);
```
yosefe, line 85: can we use ucp_wireup_ep_extract_next_ep() instead of ucp_proxy_ep_set_uct_ep directly?

src/ucp/wireup/wireup_ep.c
```diff
@@ -52,11 +52,50 @@ static ssize_t ucp_wireup_ep_bcopy_send_func(uct_ep_h uct_ep)
...
+}
+
+uct_ep_h ucp_wireup_ep_extract_msg_ep(ucp_wireup_ep_t *wireup_ep,
+                                      ucs_queue_head_t *pending_queue)
+{
+    uct_ep_h msg_ep = ucp_wireup_ep_get_msg_ep(wireup_ep);
+
+    uct_ep_pending_purge(&wireup_ep->super.super,
+                         ucp_wireup_ep_append_pending_cb, pending_queue);
+    ucs_assert(wireup_ep->pending_count == 0);
```
yosefe, line 82: seems redundant, it already exists in ucp_wireup_ep_pending_purge

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,208 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+
+    /* Move aux EP to new wireup lane */
+    msg_ep = ucp_wireup_ep_extract_msg_ep(old_ep, &pending_queue);
+    ucp_wireup_ep_set_aux(new_ep, msg_ep, aux_rsc_index, is_p2p);
+
+    /* Add wireup messages to pending queue */
+    ucs_queue_for_each_extract(req, &pending_queue, send.uct.priv, 1) {
+        status = uct_ep_pending_add(&new_ep->super.super, &req->send.uct, 0);
+        if (status != UCS_OK) {
+            ucp_wireup_purge_pending_requests(&pending_queue);
```
yosefe, line 1513: in this context, "req" is a general send request and not a wireup request message (since wireup requests are released in ucp_wireup_ep_pending_purge) so we cannot do ucs_free(req->send.buffer) in ucp_wireup_purge_pending_requests
we probably need to do "ucp_request_send" and not uct_ep_pending_add by adding these requests to replay_pending_queue

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,208 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+    }
+
+    ucs_assert(aux_rsc_index != UCP_NULL_RESOURCE);
+    ucs_queue_head_init(&pending_queue);
+
+    /* Move aux EP to new wireup lane */
+    msg_ep = ucp_wireup_ep_extract_msg_ep(old_ep, &pending_queue);
+    ucp_wireup_ep_set_aux(new_ep, msg_ep, aux_rsc_index, is_p2p);
+
+    /* Add wireup messages to pending queue */
```
yosefe, line 1509: ucp_wireup_ep_extract_msg_ep is currently NOT returning wireup message requests

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,208 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+    uct_ep_h msg_ep;
+    ucp_rsc_index_t aux_rsc_index;
+    ucs_queue_head_t pending_queue;
+    int is_p2p;
+    ucs_status_t status;
+    ucp_request_t *req;
+
+    /* Get old wireup lane */
+    old_lane = ucp_wireup_get_msg_lane(ep, UCP_WIREUP_MSG_REQUEST);
+    old_ep   = ucp_wireup_ep(ucp_ep_get_lane(ep, old_lane));
```
yosefe, line 1480: old_wireup_ep?

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,208 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+    /* Remove old wireup_ep as it's not needed anymore.
+     * NOTICE: Next two lines are intentionally not merged with the lane
+     * removal loop in ucp_wireup_check_config_intersect, because of future
+     * support for non-wireup EPs reconfiguration (which will modify this
+     * code). */
+    uct_ep_destroy(&old_ep->super.super);
+    ucp_ep_set_lane(ep, old_lane, NULL);
+
+    /* Select CM/non-reused lane as new wireup lane */
+    new_lane              = ucp_wireup_find_non_reused_lane(ep, key,
```
yosefe, line 1527: new_wireup_lane?

src/ucp/wireup/wireup.c
```diff
@@ -1355,25 +1355,208 @@ static void ucp_wireup_discard_uct_eps(ucp_ep_h ep, uct_ep_h *uct_eps,
...
+                                                            reuse_lane_map);
+    new_uct_eps[new_lane] = &new_ep->super.super;
+    key->wireup_msg_lane  = new_lane;
+    return UCS_OK;
+
+destroy_ep:
+    if (uct_ep != NULL) {
+        uct_ep_destroy(uct_ep);
+    }
+
```
yosefe, line 1537: remove spaceline

src/ucp/wireup/wireup_ep.c
```diff
@@ -52,11 +52,50 @@ static ssize_t ucp_wireup_ep_bcopy_send_func(uct_ep_h uct_ep)
...
+           (wireup_ep->aux_ep == NULL);
+}
+
+uct_ep_h ucp_wireup_ep_extract_msg_ep(ucp_wireup_ep_t *wireup_ep,
+                                      ucs_queue_head_t *pending_queue)
+{
+    uct_ep_h msg_ep = ucp_wireup_ep_get_msg_ep(wireup_ep);
+
+    uct_ep_pending_purge(&wireup_ep->super.super,
+                         ucp_wireup_ep_append_pending_cb, pending_queue);
```
yosefe, line 81: seems ucp_wireup_ep_pending_purge (called from here) already calls ucp_wireup_ep_pending_req_release which decrements pending_count by 1

src/ucp/wireup/wireup_ep.c
shasson5: solved, let's discuss tomorrow and I will explain

src/ucp/wireup/wireup.c
shasson5: same as above comment

src/ucp/wireup/wireup.c
shasson5: removed

src/ucp/wireup/wireup_ep.c
shasson5: replaced

test/gtest/ucp/test_ucp_reconfigure.cc
gleon99: Unresolved

shasson5: line 219

src/ucp/wireup/wireup_ep.c
```diff
@@ -52,11 +52,70 @@ static ssize_t ucp_wireup_ep_bcopy_send_func(uct_ep_h uct_ep)
...
+    return (wireup_ep->flags & UCP_WIREUP_EP_FLAG_READY) ||
+           (wireup_ep->aux_ep == NULL);
+}
+
+uct_ep_h ucp_wireup_ep_extract_msg_ep(ucp_wireup_ep_t *wireup_ep)
+{
+    uct_ep_h msg_ep = ucp_wireup_ep_get_msg_ep(wireup_ep);
+
+    ucp_wireup_ep_pending_purge(&wireup_ep->super.super,
+                                ucs_empty_function_do_assert_void, NULL);
```
yosefe, line 102: IMO we can just assert that wireup_ep->pending_q length is 0. and call uct_ep_pending_purge. then no need to move the functions ucp_wireup_ep_pending_purge, ucp_wireup_ep_pending_req_release

src/ucp/wireup/wireup_ep.c
```diff
@@ -52,11 +52,70 @@ static ssize_t ucp_wireup_ep_bcopy_send_func(uct_ep_h uct_ep)
...
+
+    if (ucp_wireup_ep_is_next_ep_active(wireup_ep)) {
+        ucp_wireup_ep_extract_next_ep(&wireup_ep->super.super);
+    } else {
+        wireup_ep->aux_ep        = NULL;
+        wireup_ep->aux_rsc_index = UCP_NULL_RESOURCE;
+    }
+
+    return msg_ep;
```
yosefe, line 111: don't use the function ucp_wireup_ep_get_msg_ep
and then:

```suggestion
    if (ucp_wireup_ep_is_next_ep_active(wireup_ep)) {
        return ucp_wireup_ep_extract_next_ep(&wireup_ep->super.super);
    }
    
    msg_ep                   = wireup->aux_ep;
    wireup_ep->aux_ep        = NULL;
    wireup_ep->aux_rsc_index = UCP_NULL_RESOURCE;
    return msg_ep;
```

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -18,37 +18,26 @@ class test_ucp_ep_reconfigure : public ucp_test {
...
     public:
-        void connect(const entity &other, bool is_exclude_iface);
-
-        void verify_configuration(const entity &other) const;
-
-        bool is_reconfigured() const
+        entity(const ucp_test_param &test_param, ucp_config_t *ucp_config,
+               const ucp_worker_params_t &worker_params,
+               const ucp_test_base *test_owner, bool exclude_ifaces) :
+            ucp_test_base::entity(test_param, ucp_config, worker_params,
```
yosefe, line 26: i'm not sure we need the prefix ucp_test_base:: in this context (multiple locations)

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -0,0 +1,255 @@
...
+#include "ucp_test.h"
+#include "common/test.h"
+
+extern "C" {
+#include <ucp/core/ucp_ep.inl>
+#include <ucp/wireup/address.h>
+#include <ucp/wireup/wireup_ep.h>
+}
+
+class test_ucp_ep_reconfigure : public ucp_test {
```
yosefe, line 16: rename to test_ucp_ep_reconfig

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -0,0 +1,255 @@
...
+    typedef enum {
+        VARIANT_DEFAULT,
+        VARIANT_EXCLUDE_IFACES
+    } variant_t;
+
+    void create_entities_and_connect()
+    {
+        m_entities.push_front(new entity(GetParam(), m_ucp_config,
+                                         get_worker_params(), this,
+                                         is_exclude_ifaces()));
```
yosefe, line 64: minor: maybe add helper function in this class to create and push new entity

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -0,0 +1,255 @@
...
+{
+    auto r_other     = static_cast<const entity*>(other);
+    auto worker_addr = r_other->get_address(ucp_tl_bitmap_max);
+    ucp_tl_bitmap_t tl_bitmap;
+    unsigned addr_indices[UCP_MAX_LANES];
+    ucp_ep_h ucp_ep;
+
+    tl_bitmap = m_exclude_ifaces ?
+                        UCS_STATIC_BITMAP_NOT(r_other->ep_tl_bitmap()) :
+                        ucp_tl_bitmap_max;
```
yosefe, line 150: if the variant is not exclude_ifaces, how is it different than normal connection flow? what is causing a reconfiguration in this case?

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -0,0 +1,255 @@
...
+        const entity &other) const
+{
+    unsigned reused                  = 0;
+    const ucp_lane_index_t num_lanes = ucp_ep_num_lanes(ep());
+
+    EXPECT_EQ(num_lanes, ucp_ep_num_lanes(other.ep()));
+
+    for (ucp_lane_index_t lane = 0; lane < num_lanes; ++lane) {
+        /* Verify local and remote lanes are identical */
+        EXPECT_TRUE(has_matching_lane(ep(), lane, other));
```
yosefe, line 203: do we really need this? isn't it enough to test that communication send/recv is working?
same for checking num_lanes

shasson5: how is it possible to check all lanes without using multiple protocols/message sizes?

test/gtest/ucp/test_ucp_ep_reconfig.cc
shasson5: The idea was to check a flow without reconfig, but I indeed removed it in the next PRs because it is covered by other UCP tests.
I'll backport the removal to this PR.

test/gtest/ucp/test_ucp_ep_reconfig.cc
shasson5: compiler error is generated without it.
How should it be distinguished from reconfig::entity?

yosefe: right

src/ucp/wireup/wireup.c
```diff
@@ -1642,6 +1594,34 @@ ucp_wireup_check_config_intersect(ucp_ep_h ep, ucp_ep_config_key_t *new_key,
...
+    status = ucp_wireup_select_lanes(ep, ep_init_flags, *tl_bitmap,
+                                     remote_address, addr_indices, key, 1);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    /* Get all reachable MDs from full remote address list and join with
+     * current ep configuration
+     */
+    key->dst_md_cmpts = dst_md_storage;
```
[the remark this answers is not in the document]
shasson5: you mean move it to be before 1609?

test/gtest/ucp/test_ucp_ep_reconfig.cc
yosefe: IMO we can do basic test like active message or tag match on several message sizes and extend it
since communication ability is something we need to test anyway to ensure configuration is working

src/ucp/wireup/wireup_ep.c
shasson5: ok but notice it requires duplication of `ucs_assert(wireup_ep->pending_count == 0);` as well

shasson5: and also `ucp_wireup_ep_pending_req_release` still required to be moved because it is passed also in `uct_ep_pending_purge`

src/ucp/wireup/wireup.c
```diff
@@ -1658,27 +1638,30 @@ ucs_status_t ucp_wireup_init_lanes(ucp_ep_h ep, unsigned ep_init_flags,
...
+        goto out;
+    }
 
-    /* Allow to choose only the lanes that were already chosen for case
-     * without CM to prevent reconfiguration error.
-     */
     if ((ep->cfg_index != UCP_WORKER_CFG_INDEX_NULL) &&
-        !ucp_wireup_can_reconfigure(ep, ep_init_flags, &tl_bitmap,
-                                    remote_address)) {
+        !ucp_ep_config_is_equal(&ucp_ep_config(ep)->key, &key) &&
```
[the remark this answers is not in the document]
shasson5: it's not possible as some places check (cfg_index == NULL) and some check (cfg_index != NULL)

test/gtest/ucp/test_ucp_ep_reconfig.cc
shasson5: maybe better as a complement?

test/gtest/ucp/test_ucp_ep_reconfig.cc
shasson5: removed variant, but left flag, because will be needed to distinguish from another reconf method in next PR

src/ucp/wireup/wireup_ep.c
yosefe: why need to check pending_count==0? ucp_wireup_ep_pending_purge would release them anyway so it doesn't have to be empty IMO

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -52,60 +52,58 @@ class test_ucp_ep_reconfigure : public ucp_test {
...
+                mem_buffer sbuf(msg_size, UCS_MEMORY_TYPE_HOST, i);
+                mem_buffer rbuf(msg_size, UCS_MEMORY_TYPE_HOST, ucs::rand());
+
+                void *sreq = ucp_tag_send_nbx(sender().ep(), sbuf.ptr(),
+                                              msg_size, 0, &param);
+                void *rreq = ucp_tag_recv_nbx(receiver().worker(), rbuf.ptr(),
+                                              msg_size, 0, 0, &param);
+                request_wait(rreq);
+                request_wait(sreq);
+                rbuf.pattern_check(i);
```
yosefe, line 100: 👍

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -52,60 +52,58 @@ class test_ucp_ep_reconfigure : public ucp_test {
...
-    {
-        return get_variant_value() == VARIANT_EXCLUDE_IFACES;
+        add_variant_with_value(variants, UCP_FEATURE_TAG, 0, "");
     }
 
     void send_recv()
     {
         static const unsigned num_iterations = 100;
-        static const size_t msg_size         = 16 * UCS_KBYTE;
+        static const size_t msg_sizes[]      = {8, 1024, 16384};
```
yosefe, line 84: pls add 1MB size

test/gtest/ucp/test_ucp_ep_reconfig.cc
shasson5: I see, so removed the num_lanes check

shasson5:
@yosefe please approve

yosefe (approved):
pls squash

yosefe (approved):

PR #9960, opened 2024-06-16
UCP/AM: Limit active message id to uint16_t

amastbaum:
## What
When registering an AM handler, ensure the id fits within the `uint16_t` range.

## Why ?
The AM API accepts the AM handler id as an `unsigned int`, but internally the id is constrained to `uint16_t`. This difference can cause multiple ids (up to 2^16) to map to the same internal id, ultimately referring to the same AM handler (the last one registered).

## How ?
When registering AM handler, ensure `param->id`'s 16 msb are not set. If any of them are non-zero, the function returns an error.

### DIFF
--- a/src/ucp/core/ucp_am.c
+++ b/src/ucp/core/ucp_am.c
@@ -316,6 +316,17 @@ ucp_am_zcopy_pack_user_header(ucp_request_t *req)
     return UCS_OK;
 }
 
+static ucs_status_t ucp_am_check_id(unsigned am_id)
+{
+    if (ENABLE_PARAMS_CHECK && ucs_unlikely(am_id > UINT16_MAX)) {
+        ucs_error("invalid AM id %u, must be in range [0, %u]",
+                  am_id, UINT16_MAX);
+        return UCS_ERR_INVALID_PARAM;
+    }
+
+    return UCS_OK;
+}
+
 ucs_status_t ucp_worker_set_am_recv_handler(ucp_worker_h worker,
                                             const ucp_am_handler_param_t *param)
 {
@@ -326,6 +337,11 @@ ucs_status_t ucp_worker_set_am_recv_handler(ucp_worker_h worker,
     if (!ucs_test_all_flags(param->field_mask, UCP_AM_HANDLER_PARAM_FIELD_ID |
                                                UCP_AM_HANDLER_PARAM_FIELD_CB)) {
         return UCS_ERR_INVALID_PARAM;
+    }
+
+    status = ucp_am_check_id(param->id);
+    if (status != UCS_OK) {
+        return status;
     }
 
     id    = param->id;
@@ -948,6 +964,11 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
     UCP_CONTEXT_CHECK_FEATURE_FLAGS(worker->context, UCP_FEATURE_AM,
                                     return UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM));
     UCP_REQUEST_CHECK_PARAM(param);
+
+    status = ucp_am_check_id(id);
+    if (status != UCS_OK) {
+        return UCS_STATUS_PTR(status);
+    }
 
     UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);
 
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -418,9 +418,9 @@ protected:
         }
     }
 
-    void set_am_data_handler(entity &e, uint16_t am_id,
-                             ucp_am_recv_callback_t cb, void *arg,
-                             unsigned flags = 0)
+    ucs_status_t set_am_data_handler_internal(entity &e, unsigned am_id,
+                                              ucp_am_recv_callback_t cb, void *arg,
+                                              unsigned flags = 0)
     {
         ucp_am_handler_param_t param;
 
@@ -437,7 +437,14 @@ protected:
             param.flags       = flags;
         }
 
-        ASSERT_UCS_OK(ucp_worker_set_am_recv_handler(e.worker(), &param));
+        return ucp_worker_set_am_recv_handler(e.worker(), &param);
+    }
+
+    void set_am_data_handler(entity &e, uint16_t am_id,
+                             ucp_am_recv_callback_t cb, void *arg,
+                             unsigned flags = 0)
+    {
+        ASSERT_UCS_OK(set_am_data_handler_internal(e, am_id, cb, arg, flags));
     }
 
     void check_header(const void *header, size_t header_length)
@@ -448,11 +455,11 @@ protected:
 
     ucs_status_ptr_t
     update_counter_and_send_am(const void *header, size_t header_length,
-                               const void *buffer, size_t count,
+                               const void *buffer, size_t count, unsigned am_id,
                                const ucp_request_param_t *param)
     {
         m_send_counter++;
-        return ucp_am_send_nbx(sender().ep(), TEST_AM_NBX_ID, header,
+        return ucp_am_send_nbx(sender().ep(), am_id, header,
                                header_length, buffer, count, param);
     }
 
@@ -478,6 +485,7 @@ protected:
         ucs_status_ptr_t sptr = update_counter_and_send_am(hdr, hdr_length,
                                                            dt_desc.buf(),
                                                            dt_desc.count(),
+                                                           TEST_AM_NBX_ID,
                                                            &param);
         return sptr;
     }
@@ -692,6 +700,46 @@ protected:
     ucp_mem_h                       m_rx_memh;
 };
 
+class test_ucp_am_id : public test_ucp_am_nbx {
+protected:
+    void reset_counters()
+    {
+        test_ucp_am_nbx::reset_counters();
+        m_recv_counter_cb = 0;
+    }
+
+    void test_am_id_handler()
+    {
+        reset_counters();
+
+        ucs_status_t status = set_am_data_handler_internal(
+                                            receiver(), 0xffff0001,
+                                            am_id_overflow_data_cb, this);
+        EXPECT_EQ(UCS_ERR_INVALID_PARAM, status);
+
+        ucp_request_param_t param;
+        param.op_attr_mask = 0;
+
+        ucs_status_ptr_t sptr = update_counter_and_send_am(NULL, 0ul, NULL, 0,
+                                                           0xffff0001, &param);
+        EXPECT_EQ(UCS_PTR_STATUS(sptr), UCS_ERR_INVALID_PARAM);
+        EXPECT_EQ(0, m_recv_counter_cb);
+    }
+
+    static ucs_status_t am_id_overflow_data_cb(void *arg, const void *header,
+                                               size_t header_length,
+                                               void *data, size_t length,
+                                               const ucp_am_recv_param_t *param)
+    {
+        test_ucp_am_id *self = reinterpret_cast<test_ucp_am_id*>(arg);
+        self->m_recv_counter_cb++;
+        return self->am_data_handler(header, header_length,
+                                     data, length, param);
+    }
+
+    volatile size_t m_recv_counter_cb;
+};
+
 void test_ucp_am_nbx::test_datatypes(std::function<void()> test_f,
                                      const std::vector<ucp_dt_type> &datatypes)
 {
@@ -796,6 +844,14 @@ UCS_TEST_P(test_ucp_am_nbx, am_header_error)
     EXPECT_EQ(UCS_PTR_STATUS(sptr), UCS_ERR_INVALID_PARAM);
 }
 #endif
+
+UCS_TEST_P(test_ucp_am_id, am_id_overflow)
+{
+    scoped_log_handler wrap_err(wrap_errors_logger);
+    test_am_id_handler();
+}
+
+UCP_INSTANTIATE_TEST_CASE(test_ucp_am_id)
 
 UCS_TEST_P(test_ucp_am_nbx, zero_send)
 {
@@ -1732,7 +1788,8 @@ UCS_TEST_SKIP_COND_P(test_ucp_am_nbx_rndv, invalid_recv_desc,
 
     param.op_attr_mask = 0ul;
     ucs_status_ptr_t sptr = update_counter_and_send_am(NULL, 0ul, &data,
-                                                       sizeof(data), &param);
+                                                       sizeof(data),
+                                                       TEST_AM_NBX_ID, &param);
 
     wait_receives();
 
@@ -1769,7 +1826,9 @@ UCS_TEST_P(test_ucp_am_nbx_rndv, reject_rndv)
 
         ucs_status_ptr_t sptr = update_counter_and_send_am(NULL, 0ul,
                                                            sbuf.data(),
-                                                           sbuf.size(), &param);
+                                                           sbuf.size(),
+                                                           TEST_AM_NBX_ID,
+                                                           &param);
 
         EXPECT_EQ(m_status, request_wait(sptr));
         EXPECT_EQ(m_recv_counter, m_send_counter);
--- a/src/ucp/api/ucp.h
+++ b/src/ucp/api/ucp.h
@@ -1881,6 +1881,7 @@ typedef struct ucp_am_handler_param {
 
     /**
      * Active Message id.
+     * @warning Value must be between 0 and UINT16_MAX.
      */
     unsigned                 id;
 


amastbaum:
@gleon99 it's ready for review (I think I cannot edit the reviewers in the PR).

src/ucp/core/ucp_am.c
```diff
@@ -951,6 +955,11 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
 
     UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);
 
+    if (id & 0xffff0000) {
+        ret = UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM);
```
iyastreb, line 959: I guess we need to log an error with `ucs_error` macro

src/ucp/core/ucp_am.c
```diff
@@ -328,6 +328,10 @@ ucs_status_t ucp_worker_set_am_recv_handler(ucp_worker_h worker,
         return UCS_ERR_INVALID_PARAM;
     }
 
+    if (param->id & 0xffff0000) {
```
iyastreb, line 331: minor: maybe `if (param->id != (uint16_t)param->id)` to make it more explicit?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +691,148 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+        if (flags != 0) {
+            param.field_mask |= UCP_AM_HANDLER_PARAM_FIELD_FLAGS;
+            param.flags       = flags;
+        }
+
+        return ucp_worker_set_am_recv_handler(e.worker(), &param);
+    }
+
+    void test_am_id_handler(size_t size, size_t header_size = 0ul,
+                            unsigned flags = 0, unsigned data_cb_flags = 0,
```
iyastreb, line 733: I guess we don't really need all default arguments

gleon99:
> amastbaum

@amastbaum plz check your email, added you to maintainers group.

src/ucp/core/ucp_am.c
amastbaum: I see that using `ucs_error` fails the test automatically (I wanted to specifically check for a failure when the id is too large). @iyastreb  What is preferable? issuing a ucs_error message or testing this case? or is there another solution?

src/ucp/core/ucp_am.c
tvegas1: `if (param->id > UINT16_MAX)`

src/ucp/core/ucp_am.c
tvegas1: for test to ignore error, you could use `scoped_log_handler slh(hide_errors_logger);`

tvegas1: maybe use UINT16_MAX

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -619,7 +619,6 @@ class test_ucp_am_nbx : public test_ucp_am_base {
         return am_data_rndv_handler(data, length);
     }
 
-
```
tvegas1, line 622: seems unrelated

src/ucp/core/ucp_am.c
```diff
@@ -328,6 +328,12 @@ ucs_status_t ucp_worker_set_am_recv_handler(ucp_worker_h worker,
         return UCS_ERR_INVALID_PARAM;
     }
 
+    if (param->id > UINT16_MAX) {
+        ucs_error("AM handler id is too large (%u). "
```
iyastreb, line 332: I would rephrase it like:
`AM handler id %u is outside of allowed range [0, %u], id, UINT16_MAX`
But up to you

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +692,141 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+        mem_buffer sbuf(size, tx_memtype());
+        sbuf.pattern_fill(SEED);
+        m_hdr.resize(header_size);
+        ucs::fill_random(m_hdr);
+        reset_counters();
+        ucp_mem_h memh = NULL;
+
+        EXPECT_EQ(UCS_OK, set_am_id_overflow_data_handler(receiver(),
+                                        0x1, am_id_overflow_data_cb_1,
+                                        this, 0));
```
iyastreb, line 744: I think you don't need to pass 2 last parameters - `this` can be taken from the `set_am_id_overflow_data_handler`
And `flags` are not used, so no need to have that param at all

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +692,141 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+        param.op_attr_mask = UCP_OP_ATTR_FIELD_DATATYPE;
+        param.datatype     = sdt_desc.dt();
+
+        if (memh != NULL) {
+            param.op_attr_mask |= UCP_OP_ATTR_FIELD_MEMH;
+            param.memh          = memh;
+        }
+
+        m_send_counter++;
+        ucs_status_ptr_t sptr = ucp_am_send_nbx(sender().ep(), 0x1, m_hdr.data(),
```
iyastreb, line 765: Since we are testing AM IDs, I'm not sure if we need to register and allocate memory.
I would propose to implement the minimal test just for this particular improvement without memory registration/flags/mem_buffer.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +692,111 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+                  !!(rx_param->recv_attr & UCP_AM_RECV_ATTR_FIELD_REPLY_EP));
+        EXPECT_EQ(has_reply_ep, rx_param->reply_ep != NULL);
+
+        if (1 == handler_num) m_recv_counter_cb_1++;
+        else if (2 == handler_num) m_recv_counter_cb_2++;
+
+        return UCS_OK;
+    }
+
+    static ucs_status_t am_id_overflow_data_cb_1(void *arg, const void *header,
```
gleon99, line 776: Code repetition.
Maybe use lambda or `std::bind` to eliminate  it?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +692,111 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+protected:
+
+    void reset_counters()
+    {
+        test_ucp_am_nbx::reset_counters();
+        m_recv_counter_cb_1 = 0;
+        m_recv_counter_cb_2 = 0;
+    }
+
+    ucs_status_t set_am_id_overflow_data_handler(entity &e, unsigned am_id,
```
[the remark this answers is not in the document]
amastbaum: It's different in that `set_am_data_handler` receives `am_id` as `uint16_t` instead of `unsigned int`, and it also asserts successful registration of data handler while `set_am_id_overflow_data_handler` only returns status, so the opposite case can also be tested when using this function. I can change its name though as there is no overflow here.. or maybe I can change `set_am_data_handler` so it would be more generic.
What do you think?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +692,111 @@ class test_ucp_am_nbx : public test_ucp_am_base {
     ucp_mem_h                       m_rx_memh;
 };
 
+class test_ucp_am_id : public test_ucp_am_nbx {
+public:
+    test_ucp_am_id()
```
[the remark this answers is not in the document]
amastbaum: I guess I was trying to mimic the behavior of `test_ucp_am_nbx` and call `reset_counters()` as done there, but on a second thought `reset_counters()` is called at the beginning of `test_am_id_handler()` anyway..

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +692,111 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+
+        EXPECT_EQ(UCS_OK, set_am_id_overflow_data_handler(receiver(),
+                                        0x1, am_id_overflow_data_cb_1));
+        EXPECT_NE(UCS_OK, set_am_id_overflow_data_handler(receiver(),
+                                        0xffff0001, am_id_overflow_data_cb_2));
+
+        ucp_request_param_t param;
+        param.op_attr_mask = 0;
+
+        m_send_counter++;
```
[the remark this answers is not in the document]
amastbaum: Right.. I just avoided adding a parameter to `update_counter_and_send_am` but for no reason :)

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +692,111 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+                           UCP_AM_HANDLER_PARAM_FIELD_CB |
+                           UCP_AM_HANDLER_PARAM_FIELD_ARG;
+        param.id         = am_id;
+        param.cb         = cb;
+        param.arg        = this;
+
+        return ucp_worker_set_am_recv_handler(e.worker(), &param);
+    }
+
+    void test_am_id_handler(size_t size, size_t header_size = 0ul)
```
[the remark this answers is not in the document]
amastbaum: no reason actually

test/gtest/ucp/test_ucp_am.cc
amastbaum: @gleon99 I was trying to use either lambda or `std::bind` to define a function of type `ucp_am_recv_callback_t` but it didn't work and became too messy.. I am pushing the code without this change for now, but would like to talk about it, maybe I have some gaps in this matter

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +693,69 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+        test_ucp_am_nbx::reset_counters();
+        m_recv_counter_cb_1 = 0;
+        m_recv_counter_cb_2 = 0;
+    }
+
+    void test_am_id_handler()
+    {
+        reset_counters();
+
+        EXPECT_UCS_OK(test_ucp_am_nbx::set_am_data_handler(
```
gleon99, line 709: `test_ucp_am_nbx::` really needed?

test/gtest/ucp/test_ucp_am.cc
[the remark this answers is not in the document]
gleon99: I see.
Than maybe have `set_am_data_handler` in the base class, which would call `set_am_data_handler_internal` - which will do the job and return rc.
`set_am_data_handler` will `ASSERT_...` so that no change in existing code is needed.

Your new func `set_am_..something` in the subclass will call `set_am_data_handler_internal` as well.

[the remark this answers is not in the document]
amastbaum: Oh, I missed this comment yesterday. I can revert it

test/gtest/ucp/test_ucp_am.cc
gleon99: Per our f2f discussion, lambda or `std::bind` won't work here because of C/C++ types interoperability limitation.

src/ucp/core/ucp_am.c
```diff
@@ -951,6 +957,13 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
 
     UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);
 
+    if (id > UINT16_MAX) {
```
brminich, line 960: maybe 
```suggestion
    if (ENABLE_PARAMS_CHECK && ucs_unlikely(id > UINT16_MAX)) {
```
to avoid extra overhead on fast path, like it is done in other funcs in this file (e.g. `ucp_am_recv_data_nbx`)

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +700,69 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+    }
+
+    void test_am_id_handler()
+    {
+        reset_counters();
+
+        EXPECT_UCS_OK(set_am_data_handler_internal(
+                                        receiver(), 0x1,
+                                        am_id_overflow_data_cb_1, this));
+        EXPECT_NE(UCS_OK, set_am_data_handler_internal(
```
brminich, line 719:
```suggestion
        EXPECT_EQ(UCS_ERR_INVALID_PARAM, set_am_data_handler_internal(
```

src/ucp/core/ucp_am.c
amastbaum: @brminich When is `ENABLE_PARAMS_CHECK` set to 0?

brminich: it is a configure option which is set to `0` for release builds
```
# ./configure --help | grep params
  --disable-params-check  Disable checking user parameters passed to API,
# cat ./contrib/configure-release | grep params
        --disable-params-check \
```

brminich (approved):

src/ucp/core/ucp_am.c
amastbaum: But I think we want to check this parameter in release as well. Don't we?

amastbaum: I see. Thanks

src/ucp/core/ucp_am.c
```diff
@@ -328,6 +328,12 @@ ucs_status_t ucp_worker_set_am_recv_handler(ucp_worker_h worker,
         return UCS_ERR_INVALID_PARAM;
     }
 
+    if (param->id > UINT16_MAX) {
+        ucs_error("Invalid AM id %u, must be in range [0, %u]",
```
yosefe, line 332: small case

src/ucp/core/ucp_am.c
```diff
@@ -951,6 +957,13 @@ UCS_PROFILE_FUNC(ucs_status_ptr_t, ucp_am_send_nbx,
 
     UCP_WORKER_THREAD_CS_ENTER_CONDITIONAL(worker);
 
+    if (ENABLE_PARAMS_CHECK && ucs_unlikely(id > UINT16_MAX)) {
+        ret = UCS_STATUS_PTR(UCS_ERR_INVALID_PARAM);
+        ucs_error("Invalid AM id %u, must be in range [0, %u]",
```
yosefe, line 962: 1. small case
2. try to unite common code with ucp_worker_set_am_recv_handler(), a helper function like ucp_am_check_id(unsigned am_id) that will print the error message and return failure status if am id is bad

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +700,69 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+        test_ucp_am_nbx::reset_counters();
+        m_recv_counter_cb_1 = 0;
+        m_recv_counter_cb_2 = 0;
+    }
+
+    void test_am_id_handler()
+    {
+        reset_counters();
+
+        EXPECT_UCS_OK(set_am_data_handler_internal(
```
yosefe, line 716: can were just use set_am_data_handler?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +700,69 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+                                              NULL, 0ul, NULL, 0, 0x1, &param);
+        wait_receives();
+        ASSERT_UCS_OK(request_wait(sptr));
+        EXPECT_EQ(m_recv_counter_cb_1, 1);
+        EXPECT_EQ(m_recv_counter_cb_2, 0);
+
+        sptr = update_counter_and_send_am(NULL, 0ul, NULL, 0, 0xffff0001, &param);
+        EXPECT_EQ(UCS_PTR_STATUS(sptr), UCS_ERR_INVALID_PARAM);
+        EXPECT_EQ(m_recv_counter_cb_1, 1);
+        EXPECT_EQ(m_recv_counter_cb_2, 0);
```
yosefe, line 736: I don't think we need to test the "good" flow, there are quite a lot of ucp ap tests already, in the new tests we can just check the erroneous cases of passing invalid am id

test/gtest/ucp/test_ucp_am.cc
amastbaum: set_am_data_handler only receives uint16_t as a handler id (this is the original implementation, and it was suggested to keep it that way although now I am not sure why), and I wanted to reproduce a scenario where the user passes 2 different ids that are eventually translated to the same handler function, and for that I need to use a function that receives unsigned int.

yosefe: I think we need to change set_am_data_handler to accept unsigned am_im

src/ucp/core/ucp_am.c
amastbaum: 2. so we prefer to use ENABLE_PARAMS_CHECK condition on both cases? (fast **and slow** path)?

test/gtest/ucp/test_ucp_am.cc
amastbaum: Another difference I forgot to mention here is that set_am_data_handler_internal doesn't check/expect success while set_am_data_handler does. All the other functions that call set_am_data_handler assume it succeeds except for this function (test_am_id_handler) so this is another reason for dividing it into 2 functions.

src/ucp/core/ucp_am.c
amastbaum: @yosefe 
2. after adding a helper function
```
static ucs_status_t ucp_am_check_id(unsigned am_id)
{
    if (ENABLE_PARAMS_CHECK && ucs_unlikely(am_id > UINT16_MAX)) {
        printf("invalid AM id %u, must be in range [0, %u]",
                  am_id, UINT16_MAX);
        return UCS_ERR_INVALID_PARAM;
    }

    return UCS_OK;
}
```
I get a very strange warning that fails gtest and it seems to be unrelated (at least not directly related).
This is the output:

```
$ GTEST_FILTER=*rc/*test_ucp_am_id.am_id_overflow* ./test/gtest/gtest

[     INFO ] Using random seed of 10211
Note: Google Test filter = *rc/*test_ucp_am_id.am_id_overflow*
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from rc/test_ucp_am_id
[ RUN      ] rc/test_ucp_am_id.am_id_overflow/0 <rc_v>
invalid AM id 4294901761, must be in range [0, 65535]
invalid AM id 4294901761, must be in range [0, 65535]
[1719478243.466888] [rock28:19096:0]        spinlock.c:29   UCX  WARN  ucs_recursive_spinlock_destroy() failed: busy
common/test.cc:366: Failure
Failed
Got 0 errors and 1 warnings during the test
[     INFO ] < type/spinlock.c:29 ucs_recursive_spinlock_destroy() failed: busy >
[  FAILED  ] rc/test_ucp_am_id.am_id_overflow/0, where GetParam() = rc_v (347 ms)
[ RUN      ] rc/test_ucp_am_id.am_id_overflow/1 <rc_v/proto_v1>
invalid AM id 4294901761, must be in range [0, 65535]
invalid AM id 4294901761, must be in range [0, 65535]
[1719478243.741695] [rock28:19096:0]        spinlock.c:29   UCX  WARN  ucs_recursive_spinlock_destroy() failed: busy
common/test.cc:366: Failure
Failed
Got 0 errors and 1 warnings during the test
[     INFO ] < type/spinlock.c:29 ucs_recursive_spinlock_destroy() failed: busy >
[  FAILED  ] rc/test_ucp_am_id.am_id_overflow/1, where GetParam() = rc_v/proto_v1 (274 ms)
[----------] 2 tests from rc/test_ucp_am_id (621 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test suite ran. (623 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] rc/test_ucp_am_id.am_id_overflow/0, where GetParam() = rc_v
[  FAILED  ] rc/test_ucp_am_id.am_id_overflow/1, where GetParam() = rc_v/proto_v1

 2 FAILED TESTS
```

Have you seen something like this?

Unless it's a silly mistake and I missed something, please let me know if you want me to keep searching for the cause in the scope of this PR.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +700,44 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+    {
+        test_ucp_am_nbx::reset_counters();
+        m_recv_counter_cb = 0;
+    }
+
+    void test_am_id_handler()
+    {
+        reset_counters();
+
+        EXPECT_EQ(UCS_ERR_INVALID_PARAM, set_am_data_handler_internal(
```
yosefe, line 715: ucs_status_t status = set_am_data_handler_internal(...)
EXPECT_EQ(UCS_ERR_INVALID_PARAM, status)

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +700,44 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+                                         receiver(), 0xffff0001,
+                                         am_id_overflow_data_cb, this));
+
+        ucp_request_param_t param;
+        param.op_attr_mask = 0;
+
+        ucs_status_ptr_t sptr = update_counter_and_send_am(NULL, 0ul, NULL, 0, 0xffff0001, &param);
```
yosefe, line 722: line seems too long - pls check clang-format output

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -692,6 +700,44 @@ class test_ucp_am_nbx : public test_ucp_am_base {
...
+        EXPECT_EQ(UCS_PTR_STATUS(sptr), UCS_ERR_INVALID_PARAM);
+        EXPECT_EQ(m_recv_counter_cb, 0);
```
yosefe, line 724: EXPECT_EQ(0, m_recv_counter_cb);

yosefe (approved):
pls squash

PR #9963, opened 2024-06-17
BUILD/CONFIGURE: enforce 8KB stack threshold on compilation

michal-shalev:
## What?
Improves stack memory usage by introducing an 8KB threshold for stack frames using the GCC flag `-Wframe-larger-than=8192`. The fixes for functions that exceed this limit were merged:
- [x] https://github.com/openucx/ucx/pull/10128
- [x] https://github.com/openucx/ucx/pull/10157
- [x] https://github.com/openucx/ucx/pull/10130
- [x] https://github.com/openucx/ucx/pull/10131
- [x] https://github.com/openucx/ucx/pull/10127
- [x] https://github.com/openucx/ucx/pull/10274

## Why?
Customers with limited stack sizes have recently experienced stack overflow issues when running UCX.
The goal is to ensure that the entire UCX runs within a defined stack size threshold.
Mitigates the risk of stack overflow, particularly in environments with limited stack sizes.

## How?
- [x] Include -Wframe-larger-than=8192 in the build configuration to trigger warnings for large stack frames.
- [x] Review and refactor functions with large stack frames or move large variables to the heap.

### DIFF
--- a/config/m4/compiler.m4
+++ b/config/m4/compiler.m4
@@ -451,6 +451,23 @@ AS_IF([test "x$enable_gcov" = xyes],
 
 
 #
+# Enable stack usage check
+#
+AC_ARG_ENABLE([stack-usage-check],
+        AS_HELP_STRING([--enable-stack-usage-check], [Enable stack usage check with -Wframe-larger-than=8192]),
+        [],
+        [enable_stack_usage_check=yes])
+
+AS_IF([test "x$enable_stack_usage_check" = xyes],
+      [AS_IF([test "x$enable_asan" = xyes],
+             [AC_MSG_NOTICE([Stack usage check was not added because ASAN is enabled])],
+             [ADD_COMPILER_FLAG_IF_SUPPORTED([stack_usage_check],
+                                             [-Wframe-larger-than=8192],
+                                             [AC_LANG_SOURCE([[int main(int argc, char** argv) { return 0; }]])])])],
+      [AC_MSG_NOTICE([Stack usage check is disabled])])
+
+
+#
 # Check for C++ support
 #
 CHECK_CXX_COMP()


src/ucs/debug/log.c
```diff
@@ -145,15 +145,24 @@ static void ucs_log_get_file_name(char *log_file_name, size_t max, int idx)
...
-    char new_log_file_name[PATH_MAX];
+    char *old_log_file_name = NULL;
+    char *new_log_file_name = NULL;
     int idx, ret;
 
+    old_log_file_name = (char *)ucs_malloc(PATH_MAX, "old_log_file_name");
+    new_log_file_name = (char *)ucs_malloc(PATH_MAX, "new_log_file_name");
+
+    if (old_log_file_name == NULL || new_log_file_name == NULL) {
+        ucs_free(old_log_file_name);
```
tvegas1, line 156: I guess no need to free when we `ucs_fatal()`.

src/ucs/debug/log.c
```diff
@@ -145,15 +145,24 @@ static void ucs_log_get_file_name(char *log_file_name, size_t max, int idx)
...
-    char old_log_file_name[PATH_MAX];
-    char new_log_file_name[PATH_MAX];
+    char *old_log_file_name = NULL;
+    char *new_log_file_name = NULL;
     int idx, ret;
 
+    old_log_file_name = (char *)ucs_malloc(PATH_MAX, "old_log_file_name");
+    new_log_file_name = (char *)ucs_malloc(PATH_MAX, "new_log_file_name");
+
+    if (old_log_file_name == NULL || new_log_file_name == NULL) {
```
tvegas1, line 155: coding style mandates parenthesis around each check

src/ucs/debug/log.c
```diff
@@ -145,15 +145,24 @@ static void ucs_log_get_file_name(char *log_file_name, size_t max, int idx)
 
 static void ucs_log_file_rotate()
 {
-    char old_log_file_name[PATH_MAX];
-    char new_log_file_name[PATH_MAX];
+    char *old_log_file_name = NULL;
```
tvegas1, line 148: no need to initialize

src/ucs/debug/log.c
tvegas1: valid for all cases.

src/ucs/debug/log.c
```diff
@@ -145,15 +145,24 @@ static void ucs_log_get_file_name(char *log_file_name, size_t max, int idx)
...
+    char *new_log_file_name = NULL;
     int idx, ret;
 
+    old_log_file_name = (char *)ucs_malloc(PATH_MAX, "old_log_file_name");
```
iyastreb, line 152: in C this cast from `void *` is redundant

src/ucs/debug/log.c
```diff
@@ -164,30 +173,40 @@ static void ucs_log_file_rotate()
...
+                              PATH_MAX, idx);
         ucs_log_get_file_name(new_log_file_name,
-                              sizeof(new_log_file_name), idx + 1);
+                              PATH_MAX, idx + 1);
 
         if (access(old_log_file_name, W_OK) != 0) {
+            ucs_free(old_log_file_name);
+            ucs_free(new_log_file_name);
             ucs_fatal("unable to write to %s", old_log_file_name);
         }
```
ivankochin, line 184: Printing log file names after freeing them seems like not a good idea. Also there is a lot of code duplication, so I would propose something like the following:

```suggestion
        if (access(old_log_file_name, W_OK) != 0) {
            ucs_error("unable to write to %s", old_log_file_name);
            goto err_out;
        }
        
...

err_out:
    ucs_free(old_log_file_name);
    ucs_free(new_log_file_name);
    if (err_msg) {
        ucs_fatal("ucs_log_file_rotate() fatal error (see logs before)");
    }
```

src/ucs/debug/log.c
michal-shalev: done

src/ucs/debug/log.c
michal-shalev: done

src/ucs/debug/log.c
michal-shalev: done

src/ucs/debug/log.c
michal-shalev: Thank you :), please check now @ivankochin

src/ucs/debug/log.c
michal-shalev: Thank you :), please check now @tvegas1

src/ucs/debug/log.c
```diff
@@ -164,30 +172,41 @@ static void ucs_log_file_rotate()
...
         if (access(new_log_file_name, W_OK) != 0) {
-            ucs_fatal("unable to write to %s", new_log_file_name);
+            ucs_error("unable to write to %s", new_log_file_name);
+            goto err_out;
         }
     }
+
+    ucs_free(old_log_file_name);
+    ucs_free(new_log_file_name);
+
```
tvegas1, line 205: missing return here ;)

src/ucs/debug/log.c
```diff
@@ -145,15 +145,23 @@ static void ucs_log_get_file_name(char *log_file_name, size_t max, int idx)
 
 static void ucs_log_file_rotate()
 {
-    char old_log_file_name[PATH_MAX];
-    char new_log_file_name[PATH_MAX];
+    char *old_log_file_name;
+    char *new_log_file_name;
     int idx, ret;
 
+    old_log_file_name = ucs_malloc(PATH_MAX, "old_log_file_name");
```
tvegas1, line 152: you could malloc upon variable declaration

src/ucs/debug/log.c
```diff
@@ -164,30 +172,41 @@ static void ucs_log_file_rotate()
...
         ucs_log_get_file_name(old_log_file_name,
-                              sizeof(old_log_file_name), idx);
+                              PATH_MAX, idx);
         ucs_log_get_file_name(new_log_file_name,
-                              sizeof(new_log_file_name), idx + 1);
+                              PATH_MAX, idx + 1);
 
         if (access(old_log_file_name, W_OK) != 0) {
-            ucs_fatal("unable to write to %s", old_log_file_name);
+            ucs_error("unable to write to %s", old_log_file_name);
```
tvegas1, line 180: since we fatal anyways, maybe set `msg = "..."` then print with `ucs_fatal(".. %s.. ", msg);`

src/ucs/debug/log.c
```diff
@@ -164,30 +172,41 @@ static void ucs_log_file_rotate()
...
+            ucs_error("unable to write to %s", new_log_file_name);
+            goto err_out;
         }
     }
+
+    ucs_free(old_log_file_name);
+    ucs_free(new_log_file_name);
+
+err_out:
+    ucs_free(old_log_file_name);
```
tvegas1, line 207: no need to free i guess, since there would be so many leaked resources aynways..

src/ucs/debug/log.c
```diff
@@ -164,30 +172,41 @@ static void ucs_log_file_rotate()
...
+    ucs_free(new_log_file_name);
+    ucs_fatal("ucs_log_file_rotate() fatal error (see logs before)");
```
tvegas1, line 209: please see previous suggestion

src/ucs/debug/log.c
tvegas1: or just replace ucs_error() by ucs_fatal()

src/ucs/debug/log.c
michal-shalev: ofc >< thanks, done

src/ucs/vfs/base/vfs_obj.c
```diff
@@ -259,13 +259,21 @@ static ucs_status_t ucs_vfs_node_add(void *parent_obj, ucs_vfs_node_type_t type,
...
-    char abs_path_buf[PATH_MAX];
+    char *rel_path_buf = ucs_malloc(PATH_MAX, "rel_path_buf");
+    char *abs_path_buf = ucs_malloc(PATH_MAX, "abs_path_buf");
     char *token, *next_token;
+    ucs_status_t status = UCS_OK;
+
+    if ((rel_path_buf == NULL) || (abs_path_buf == NULL)) {
+        ucs_error("Failed to allocate memory for path buffers");
+        status = UCS_ERR_NO_MEMORY;
+        goto out;
```
tvegas1, line 270: remote `out` label and use `return UCS_ERR_NO_MEMORY` instead, rename `out_free_buffers` to `out`

src/ucs/vfs/base/vfs_obj.c
```diff
@@ -259,13 +259,21 @@ static ucs_status_t ucs_vfs_node_add(void *parent_obj, ucs_vfs_node_type_t type,
                                      va_list ap, ucs_vfs_node_t **new_node)
 {
     ucs_vfs_node_t *current_node;
```
tvegas1, line 261: coding guideline mandates `current_node` be after initialized variables

src/ucs/sys/topo/base/topo.c
```diff
@@ -383,8 +383,16 @@ ucs_topo_get_distance_sysfs(ucs_sys_device_t device1,
...
+    char *path1 = ucs_malloc(PATH_MAX, "path1");
+    char *path2 = ucs_malloc(PATH_MAX, "path2");
+    char *common_path = ucs_malloc(PATH_MAX, "common_path");
+    ucs_status_t status = UCS_OK;
+
+    if ((path1 == NULL) || (path2 == NULL) || (common_path == NULL)) {
+        ucs_error("Failed to allocate memory for path buffers");
+        status = UCS_ERR_NO_MEMORY;
+        goto out;
+    }
```
ivankochin, line 395: That condition can lead to memory leaks (e.g. when only `common_path` allocation failed so we need to free two other pointers before calling return)

michal-shalev: fixed

config/m4/compiler.m4
```diff
@@ -10,7 +10,20 @@
 #
 # Initialize CFLAGS
 #
-BASE_CFLAGS="-g -Wall -Werror"
+# BASE_CFLAGS sets compilation flags for GCC:
+# -g: Debug information
+# -Wall: All common warnings
+# -Werror: Treat warnings as errors
+# -Wframe-larger-than=8192: Warn if stack frame size exceeds 8192 bytes (8 KB)
+BASE_CFLAGS="-g -Wall -Werror -Wframe-larger-than=8192"
```
tvegas1, line 18: might as well remove comments since the flags are so common

src/tools/perf/lib/uct_tests.cc
```diff
@@ -773,7 +773,7 @@ class uct_perf_test_runner {
    TEST_CASE_ALL_OSD(_perf, _case, UCT_PERF_DATA_LAYOUT_BCOPY) \
    TEST_CASE_ALL_OSD(_perf, _case, UCT_PERF_DATA_LAYOUT_ZCOPY)
 
-ucs_status_t uct_perf_test_dispatch(ucx_perf_context_t *perf)
+ucs_status_t __attribute__ ((optimize("O1"))) uct_perf_test_dispatch(ucx_perf_context_t *perf)
```
tvegas1, line 776: why is attribute needed here?

src/ucs/debug/log.c
tvegas1: replace ucs_error() by ucs_fatal()

src/ucs/vfs/base/vfs_obj.c
tvegas1: need to move `current_node` line 264

src/ucs/vfs/base/vfs_obj.c
tvegas1: return here.

test/gtest/common/googletest/gtest-param-test.h
```diff
@@ -457,6 +457,11 @@ internal::CartesianProductHolder<Generator...> Combine(const Generator&... g) {
 #define GTEST_GET_FIRST_(first, ...) first
 #define GTEST_GET_SECOND_(first, second, ...) second
 
+// Modified to reduce stack frame size per test file.
```
tvegas1, line 460: is there a way to remove changes to googletest (disabling check for it)? or make it outside of googletest by including `gtest-param-test.h`?

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -1044,24 +1044,62 @@ ucs_status_t ucp_perf_test_dispatch(ucx_perf_context_t *perf)
...
-        );
+    );
+    return UCS_ERR_INVALID_PARAM;
+}
+
+ucs_status_t ucp_perf_test_dispatch(ucx_perf_context_t *perf)
+{
+    ucs_status_t status;
+
+    status = dispatch_osd(perf);
```
tvegas1, line 1084: maybe replace by an array of function pointers

test/gtest/ucs/test_bitops.cc
```diff
@@ -90,6 +90,8 @@ UCS_TEST_F(test_bitops, ptr_ctz) {
     ASSERT_EQ(88, ucs_count_ptr_trailing_zero_bits(buffer, 160));
 }
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wframe-larger-than="
```
tvegas1, line 94: to avoid all pragma in test/gtest, is there a way to not set the -Wframe-larger-than when compiling gtests?

tvegas1 (commented):
maybe remove #pragma in gtest by not specifing Wframe-large-than when building .cc files

src/ucs/debug/log.c
michal-shalev: replaced ucs_error() by ucs_fatal()

src/ucs/vfs/base/vfs_obj.c
michal-shalev: done

src/ucs/vfs/base/vfs_obj.c
michal-shalev: there could be a case where one of the buffer was allocated successfully and the other one failed, so I think we still need to free, so I changed it to "goto out", let me know what do you think

config/m4/compiler.m4
michal-shalev: removed "-Wall", "-Werror" and "-g"

src/tools/perf/lib/uct_tests.cc
michal-shalev: Without optimization the stack frame size > 8192 bytes. I tried using "new" at "src/tools/perf/lib/ucp_tests.cc" to make the runner use the heap instead of the stack and it compiled successfully on the HPC node, but then the CI failed because of the "new" and I couldn't reproduce it.
At the moment I changed it to pragma instead of the attribute.

test/gtest/ucs/test_bitops.cc
michal-shalev: There are a few macros that I can change to make those tests use the heap instead of the stack, just like I did in the commit "TEST/GTEST: reduce stack frame size in INSTANTIATE_TEST_SUITE_P macro".
Those macros are "UCS_TEST_P_" and "UCS_TEST_P_", I'm trying to make the syntax work, it's challenging so I added the pragma to see if I can make the CI pass meanwhile.
The reason I want to reduce the stack allocation in the tests is that later on I will check the stack usage on run time alongside running the tests and if the stack usage of the tests are initially above the threshold then it'll be harder to add a rule for the CI to catch cases of large stack usage during run time.

src/tools/perf/lib/ucp_tests.cc
michal-shalev: thank you, done

config/m4/compiler.m4
```diff
@@ -447,6 +447,26 @@ AS_IF([test "x$enable_gcov" = xyes],
...
+#
+AC_ARG_ENABLE([stack-usage-check],
+        AS_HELP_STRING([--enable-stack-usage-check], [Enable stack usage check with -Wframe-larger-than=8192]),
+        [],
+        [enable_stack_usage_check=yes])
+
+AS_IF([test "x$enable_stack_usage_check" = xyes],
+      [ADD_COMPILER_FLAG_IF_SUPPORTED([stack_usage_check], [-Wframe-larger-than=8192],
+                                      [AC_LANG_SOURCE([[int main(int argc, char** argv){return 0;}]])])],
+      [:])
```
michal-shalev, line 469: Force flag only in builds.sh

src/ucs/vfs/sock/vfs_sock.c
```diff
@@ -40,17 +41,23 @@ void ucs_vfs_sock_get_address(struct sockaddr_un *un_addr)
 
 int ucs_vfs_sock_mkdir(const char *sock_path, ucs_log_level_t log_level)
 {
-    char sock_path_dir[PATH_MAX];
+    char *sock_path_dir = (char *)ucs_malloc(PATH_MAX, "sock_path_dir");
```
iyastreb, line 44: 1. In C there is no need to cast from `void *`(in particular from `ucs_malloc`). This can be applied to all modifications
2. In this function there is a memory leak in case `mkdir` returns an error.

iyastreb:
I want to express a few concerns about this change.
1. This change might be error prone, because you need to change the flow of every function from return into goto err. It's easy to overlook a memory leaks, like I found above (I checked only 20% of the change).
2. We have a lot of boilerplate code..
3. Changing function flow due to buffer allocations seems like an overkill to me..
4. Replacing stack allocated arrays with malloc/free might not be the best choice for performance reasons.

Can we just use thread local temporary buffer instead of malloc/free?
So that we reuse the same buffer and don't have to allocate it again and again
Simplest form could be like:
```
/* In case we're using a C99 compiler */
#ifndef thread_local
#define thread_local __thread
#endif

char * ucs_get_tmp_path()
{
    static thread_local char path[PATH_MAX];
    return path;
}
```
This way we have a thread safe buffer that is not allocated on stack. And also we can keep existing functions behaviour unchanged, and without tons of boilerplate code.
+ @yosefe What do you think?

gleon99:
> I want to express a few concerns about this change.
> 
> 1. This change might be error prone, because you need to change the flow of every function from return into goto err. It's easy to overlook a memory leaks, like I found above (I checked only 20% of the change).
> 2. We have a lot of boilerplate code..
> 3. Changing function flow due to buffer allocations seems like an overkill to me..
> 4. Replacing stack allocated arrays with malloc/free might not be the best choice for performance reasons.
> 
> Can we just use thread local temporary buffer instead of malloc/free? So that we reuse the same buffer and don't have to allocate it again and again Simplest form could be like:
> 
> ```c
> /* In case we're using a C99 compiler */
> #ifndef thread_local
> #define thread_local __thread
> #endif
> 
> char * ucs_get_tmp_path()
> {
>     static thread_local char path[PATH_MAX];
>     return path;
> }
> ```
> 
> This way we have a thread safe buffer that is not allocated on stack. And also we can keep existing functions behaviour unchanged, and without tons of boilerplate code.
> 
> * @yosefe What do you think?

@iyastreb  thanks for the comments,
the concerns seem legitimate but the PR is still a draft. Let's address them when the PR is stable and ready for review.

src/ucp/proto/proto_multi.c
```diff
@@ -40,17 +40,31 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
     uint32_t weight_sum;
     ucs_status_t status;
 
+    lanes_perf_nodes = ucs_malloc(UCP_PROTO_MAX_LANES *
+                                  sizeof(*lanes_perf_nodes),
+                                  "lanes_perf_nodes");
+    lanes_perf = ucs_malloc(UCP_PROTO_MAX_LANES * sizeof(*lanes_perf),
+                            "lanes_perf");
+    if ((lanes_perf_nodes == NULL) || (lanes_perf == NULL)) {
+        ucs_free(lanes_perf_nodes);
```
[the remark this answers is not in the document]
michal-shalev: done

src/ucp/proto/proto_debug.c
```diff
@@ -979,29 +979,38 @@ ucp_proto_select_write_info(ucp_worker_h worker,
...
     ucp_proto_query_attr_t proto_attr;
     size_t range_start, range_end;
     ucs_string_buffer_t dot_strb;
-    char dir_path[PATH_MAX];
+    char *dir_path, *range_start_str, *range_end_str;
     FILE *fp;
     int ret;
 
+    dir_path = (char *)ucs_malloc(PATH_MAX, "dir_path");
+    range_start_str = (char *)ucs_malloc(64, "range_start_str");
```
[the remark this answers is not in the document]
michal-shalev: done

src/ucs/vfs/sock/vfs_sock.c
michal-shalev: done

src/ucs/debug/log.c
michal-shalev: Outdated, done in other PRs

src/ucs/debug/log.c
michal-shalev: Outdated, done in other PRs

src/ucs/debug/log.c
michal-shalev: Outdated, done in other PRs

src/ucs/vfs/base/vfs_obj.c
michal-shalev: Outdated, done in other PRs

src/ucs/vfs/base/vfs_obj.c
michal-shalev: Outdated, done in other PRs

config/m4/compiler.m4
michal-shalev: Outdated, done in other PRs

src/tools/perf/lib/uct_tests.cc
michal-shalev: Outdated, done in other PRs

test/gtest/common/googletest/gtest-param-test.h
michal-shalev: Outdated, done in other PRs

src/tools/perf/lib/ucp_tests.cc
michal-shalev: Outdated, done in other PRs

test/gtest/ucs/test_bitops.cc
michal-shalev: Outdated, done in other PRs

src/ucs/vfs/sock/vfs_sock.c
michal-shalev: Outdated, done in other PRs

src/ucp/proto/proto_debug.c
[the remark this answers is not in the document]
michal-shalev: Outdated, done in other PRs

src/ucp/proto/proto_multi.c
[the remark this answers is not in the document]
michal-shalev: Outdated, done in other PRs

michal-shalev:
@iyastreb @tvegas1 please re-review

config/m4/compiler.m4
```diff
@@ -449,6 +449,27 @@ AS_IF([test "x$enable_gcov" = xyes],
                                        [AC_LANG_SOURCE([[int main(int argc, char** argv){return 0;}]])])],
       [:])
 
+#
```
yosefe, line 452: pls use 2 space lines before/after the new code block

config/m4/compiler.m4
```diff
@@ -449,6 +449,27 @@ AS_IF([test "x$enable_gcov" = xyes],
...
+# Enable stack usage check
+#
+# To suppress -Wframe-larger-than=8192 for specific functions, use:
+# #pragma GCC diagnostic push
+# #pragma GCC diagnostic ignored "-Wframe-larger-than="
+# void specific_function() {
+#     /* Function code */
+# }
+# #pragma GCC diagnostic pop
```
yosefe, line 461: do we really want to advise users to do that?

config/m4/compiler.m4
```diff
@@ -449,6 +449,27 @@ AS_IF([test "x$enable_gcov" = xyes],
...
+AC_ARG_ENABLE([stack-usage-check],
+        AS_HELP_STRING([--enable-stack-usage-check], [Enable stack usage check with -Wframe-larger-than=8192]),
+        [],
+        [enable_stack_usage_check=yes])
+
+AS_IF([test "x$enable_stack_usage_check" = xyes && test "x$enable_asan" != xyes],
+      [ADD_COMPILER_FLAG_IF_SUPPORTED([stack_usage_check],
+                                      [-Wframe-larger-than=8192],
+                                      [AC_LANG_SOURCE([[int main(int argc, char** argv){return 0;}]])])],
+      [AC_MSG_NOTICE([Stack usage check was not added, either disabled or ASAN is enabled])])
```
yosefe, line 472: maybe print a separate message when stack usage is disabled because of ASAN

config/m4/compiler.m4
```diff
@@ -449,6 +449,27 @@ AS_IF([test "x$enable_gcov" = xyes],
...
+#     /* Function code */
+# }
+# #pragma GCC diagnostic pop
+#
+AC_ARG_ENABLE([stack-usage-check],
+        AS_HELP_STRING([--enable-stack-usage-check], [Enable stack usage check with -Wframe-larger-than=8192]),
+        [],
+        [enable_stack_usage_check=yes])
+
+AS_IF([test "x$enable_stack_usage_check" = xyes && test "x$enable_asan" != xyes],
```
yosefe, line 468: use "-a" instead of "&&", see other usages of AS_IF in the codebase

michal-shalev: Changed `&&` to nested `AS_IF`

config/m4/compiler.m4
michal-shalev: done

config/m4/compiler.m4
michal-shalev: Maybe as a tmp solution? they can also use the config flag I guess... removed it

config/m4/compiler.m4
michal-shalev: done

config/m4/compiler.m4
```diff
@@ -449,27 +449,22 @@ AS_IF([test "x$enable_gcov" = xyes],
...
-                                      [-Wframe-larger-than=8192],
-                                      [AC_LANG_SOURCE([[int main(int argc, char** argv){return 0;}]])])],
-      [AC_MSG_NOTICE([Stack usage check was not added, either disabled or ASAN is enabled])])
+AS_IF([test "x$enable_stack_usage_check" = xyes],
+      [AS_IF([test "x$enable_asan" = xyes],
+             [AC_MSG_NOTICE([Stack usage check was not added because ASAN is enabled])],
+             [ADD_COMPILER_FLAG_IF_SUPPORTED([stack_usage_check],
+                                             [-Wframe-larger-than=8192],
+                                             [AC_LANG_SOURCE([[int main(int argc, char** argv) { return 0; }]])])])],
+      [AC_MSG_NOTICE([Stack usage check was not added, as it is disabled])])
```
yosefe, line 467: Just [Stack usage check is disabled]

michal-shalev: done

yosefe (approved):

PR #9967, opened 2024-06-19
UCT/IB: Retry ODP registration unlimitedly

Artemy-Mellanox:
## What
Keep trying to register memory. If it return EAGAIN it means following registration should succeed.

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -158,9 +158,9 @@ ucs_config_field_t uct_ib_md_config_table[] = {
      "Maximal number of invalidated memory keys that are kept idle before reuse.",
      ucs_offsetof(uct_ib_md_config_t, ext.max_idle_rkey_count), UCS_CONFIG_TYPE_UINT},
 
-    {"REG_RETRY_CNT", "7",
+    {"REG_RETRY_CNT", "inf",
      "Number of memory registration attempts.",
-     ucs_offsetof(uct_ib_md_config_t, ext.reg_retry_cnt), UCS_CONFIG_TYPE_UINT},
+     ucs_offsetof(uct_ib_md_config_t, ext.reg_retry_cnt), UCS_CONFIG_TYPE_ULUNITS},
 
     {"SMKEY_BLOCK_SIZE", "8",
      "Number of indexes in a symmetric block. More can lead to less contention.",
@@ -462,7 +462,7 @@ ucs_status_t uct_ib_reg_mr(uct_ib_md_t *md, void *address, size_t length,
                            struct ibv_mr **mr_p)
 {
     ucs_time_t UCS_V_UNUSED start_time = ucs_get_time();
-    unsigned retry                     = 0;
+    unsigned long retry                = 0;
     size_t UCS_V_UNUSED dmabuf_offset;
     const char *title;
     struct ibv_mr *mr;
@@ -511,7 +511,7 @@ ucs_status_t uct_ib_reg_mr(uct_ib_md_t *md, void *address, size_t length,
     }
 
     ucs_trace("%s(pd=%p addr=%p len=%zu fd=%d offset=%zu access=0x%" PRIx64 "):"
-              " mr=%p lkey=0x%x retry=%d took %.3f ms",
+              " mr=%p lkey=0x%x retry=%lu took %.3f ms",
               title, md->pd, address, length, dmabuf_fd, dmabuf_offset,
               access_flags, mr, mr->lkey, retry,
               ucs_time_to_msec(ucs_get_time() - start_time));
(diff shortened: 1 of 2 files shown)


src/uct/ib/base/ib_md.c
```diff
@@ -158,7 +158,7 @@ ucs_config_field_t uct_ib_md_config_table[] = {
      "Maximal number of invalidated memory keys that are kept idle before reuse.",
      ucs_offsetof(uct_ib_md_config_t, ext.max_idle_rkey_count), UCS_CONFIG_TYPE_UINT},
 
-    {"REG_RETRY_CNT", "7",
+    {"REG_RETRY_CNT", "0",
```
yosefe, line 161: maybe use "inf"?

src/uct/ib/base/ib_md.c
```diff
@@ -488,10 +488,15 @@ ucs_status_t uct_ib_reg_mr(uct_ib_md_t *md, void *address, size_t length,
...
              * with invalidation */
+            if (retry > 0) {
+                usleep(1000);
+            }
+
             mr = UCS_PROFILE_CALL_ALWAYS(ibv_reg_mr, md->pd, address, length,
                                          access_flags);
         } while ((mr == NULL) && (errno == EAGAIN) &&
-                 (retry++ < md->config.reg_retry_cnt));
+                 ((md->config.reg_retry_cnt == 0) ||
```
tvegas1, line 498: else could set default value to UINT_MAX and remove this added check

src/uct/ib/base/ib_md.c
```diff
@@ -495,8 +496,7 @@ ucs_status_t uct_ib_reg_mr(uct_ib_md_t *md, void *address, size_t length,
             mr = UCS_PROFILE_CALL_ALWAYS(ibv_reg_mr, md->pd, address, length,
                                          access_flags);
         } while ((mr == NULL) && (errno == EAGAIN) &&
-                 ((md->config.reg_retry_cnt == 0) ||
-                  (retry++ < md->config.reg_retry_cnt)));
```
tvegas1: for completeness, maybe make  `retry` also `unsigned long`?

buildlib/tools/coverity.sh
```diff
@@ -63,7 +63,7 @@ run_coverity() {
 	if [ "${ucx_build_type}" == "devel" ]; then
 		cov-manage-emit --dir $cov_build --tu-pattern "file('.*/test/gtest/common/googletest/*')" delete
 	fi
-	cov-analyze --jobs $parallel_jobs $COV_OPT --security --concurrency --dir $cov_build
+	cov-analyze --jobs $parallel_jobs $COV_OPT --disable SLEEP --security --concurrency --dir $cov_build
```
yosefe, line 66: can we add a local coverity suppression as a code comment instead?

Artemy-Mellanox: i had no luck with local suppression; this checker only understands annotations at the lock acquisition location, and not at the sleep location. btw - maybe sched_yield() instead of sleep will be enough?

src/uct/ib/base/ib_md.c
brminich: @Artemy-Mellanox
wdyt about making it `inf` for gtest only? I'd not set inf by default for all cases

Artemy-Mellanox: assumption is that this behavior is not specific to gtest, a regular application will experience the same crashes and therefore we make the `inf` default

brminich: do not we expect it to succeed with the reasonable number of attempts? Imo it is better to crash after X attempts rather than hang

Artemy-Mellanox: Imo this would be justified if we encountered a hang situation or at least expected it

brminich: hang is quite generic syndrome, may be not easy to find a correlation with this code when it occurs.
My point is to set it to some relatively large value, like `128` or `256`, but ok, it's up to you

buildlib/tools/coverity.sh
yosefe: what is the coverity error?
sched_yield - not sure is needed - we call ibv_reg_mr sys call anyway and AFAIK kernel should call schedule before returning from syscall

Artemy-Mellanox: might sleep while holding lock
[output.log](https://github.com/user-attachments/files/15987033/output.log)

yosefe: seems it is 
lock: "pthread_rwlock_wrlock" locks "rcache->pgt_lock
but we're not supposed to hold rcache lock during memory registration
so maybe need some fix in rcache?

Artemy-Mellanox: following out-of-band talk, will try to remove the sleep and relay on syscall resched

yosefe (dismissed):

yosefe (approved):
pls squash

tvegas1 (approved):

PR #9978, opened 2024-06-26
UCT/CUDA_IPC: Redesign peer accessible cache

iyastreb:
## What
Redesigned peer accessible cache in uct_cuda_ipc component.
There were several design flaws in existing implementation:
- uct_cuda_ipc_component stores a reference to the last created MD. However if last MD is closed this leads to the dangling pointer (root cause of VASP application segfault)
- peer accessible cache is not thread safe, fails in case of multiple contexts
- cache is stored per each cuda_ipc MD, but in fact only the last created MD is used for caching
- minor: improved lookup to use hash map instead of an array based search

Consideration: locking
- global cache, single lock (current impl)
- cache in TLS
- cache per UCT worker

UPD:
I tested lock overhead with ucx_perftest on 4-16 threads with different message sizes (64-4096).
There is **no visible performance drop** when using global lock for peer accessibility cache.

## Why ?
Fix for https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=4705183
([RM#3955117](https://redmine.mellanox.com/issues/3955117))

### DIFF
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -28,6 +28,58 @@ static ucs_config_field_t uct_cuda_ipc_md_config_table[] = {
     {NULL}
 };
 
+static uct_cuda_ipc_dev_cache_t *uct_cuda_ipc_create_dev_cache(int dev_num)
+{
+    uct_cuda_ipc_dev_cache_t *cache;
+    ucs_status_t status;
+    int i, num_devices;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGetCount(&num_devices));
+    if (UCS_OK != status) {
+        ucs_error("cuDeviceGetCount() failed: %s", ucs_status_string(status));
+        return NULL;
+    }
+
+    cache = ucs_malloc(sizeof(*cache) + (num_devices * sizeof(uint8_t)),
+                       "uct_cuda_ipc_dev_cache_t");
+    if (cache == NULL) {
+        ucs_error("failed to allocate memory for uct_cuda_ipc_dev_cache_t");
+        return NULL;
+    }
+
+    cache->dev_num = dev_num;
+    for (i = 0; i < num_devices; ++i) {
+        cache->accessible[i] = UCS_TRY;
+    }
+
+    return cache;
+}
+
+static uct_cuda_ipc_dev_cache_t *
+uct_cuda_ipc_get_dev_cache(uct_cuda_ipc_component_t *component,
+                           const CUuuid *uuid)
+{
+    khash_t(cuda_ipc_uuid_hash) *hash = &component->uuid_hash;
+    uct_cuda_ipc_dev_cache_t *cache;
+    khiter_t iter;
+    int ret;
+
+    iter = kh_put(cuda_ipc_uuid_hash, hash, *uuid, &ret);
+    if (ret == UCS_KH_PUT_KEY_PRESENT) {
+        return kh_val(hash, iter);
+    } else if ((ret == UCS_KH_PUT_BUCKET_EMPTY) ||
+               (ret == UCS_KH_PUT_BUCKET_CLEAR)) {
+        cache = uct_cuda_ipc_create_dev_cache(kh_size(hash) - 1);
+        if (NULL != cache) {
+            kh_val(hash, iter) = cache;
+        }
+        return cache;
+    } else {
+        ucs_error("kh_put(cuda_ipc_uuid_hash) failed with %d", ret);
+        return NULL;
+    }
+}
+
 static ucs_status_t
 uct_cuda_ipc_md_query(uct_md_h md, uct_md_attr_v2_t *md_attr)
 {
@@ -115,101 +167,35 @@ found:
                                                     memh->dev_num));
 }
 
-static inline int uct_cuda_ipc_uuid_equals(const CUuuid* a, const CUuuid* b)
-{
-    int64_t *a0 = (int64_t *) a->bytes;
-    int64_t *b0 = (int64_t *) b->bytes;
-    return (a0[0] == b0[0]) && (a0[1] == b0[1]) ? 1 : 0;
-}
-
-static inline void uct_cuda_ipc_uuid_copy(CUuuid* dst, const CUuuid* src)
-{
-    int64_t *a = (int64_t *) src->bytes;
-    int64_t *b = (int64_t *) dst->bytes;
-    *b++ = *a++;
-    *b   = *a;
-}
-
-ucs_status_t uct_cuda_ipc_get_unique_index_for_uuid(int* idx,
-                                                    uct_cuda_ipc_md_t* md,
-                                                    uct_cuda_ipc_rkey_t *rkey)
-{
-    int i;
-    int num_devices;
-    int original_capacity, new_capacity;
-    int original_count, new_count;
-
-    for (i = 0; i < md->uuid_map_size; i++) {
-        if (uct_cuda_ipc_uuid_equals(&rkey->uuid, &md->uuid_map[i])) {
-            *idx = i;
-            return UCS_OK; /* found */
-        }
-    }
-
-    if (ucs_unlikely(md->uuid_map_size == md->uuid_map_capacity)) {
-        /* reallocate on demand */
-        UCT_CUDA_IPC_DEVICE_GET_COUNT(num_devices);
-        original_capacity     = md->uuid_map_capacity;
-        new_capacity          = md->uuid_map_capacity ?
-                                (md->uuid_map_capacity * 2) : 16;
-        original_count        = original_capacity * num_devices;
-        new_count             = new_capacity * num_devices;
-        md->uuid_map_capacity = new_capacity;
-        md->uuid_map          = ucs_realloc(md->uuid_map,
-                                            new_capacity * sizeof(CUuuid),
-                                            "uct_cuda_ipc_uuid_map");
-        if (md->uuid_map == NULL) {
-            return UCS_ERR_NO_MEMORY;
-        }
-
-        md->peer_accessible_cache = ucs_realloc(md->peer_accessible_cache,
-                                                new_count *
-                                                sizeof(ucs_ternary_auto_value_t),
-                                                "uct_cuda_ipc_peer_accessible_cache");
-        if (md->peer_accessible_cache == NULL) {
-            return UCS_ERR_NO_MEMORY;
-        }
-
-        for (i = original_count; i < new_count; i++) {
-            md->peer_accessible_cache[i] = UCS_TRY;
-        }
-    }
-
-    /* Add new mapping */
-    uct_cuda_ipc_uuid_copy(&md->uuid_map[md->uuid_map_size], &rkey->uuid);
-    *idx = md->uuid_map_size++;
-
-    return UCS_OK;
-}
-
-static ucs_status_t uct_cuda_ipc_is_peer_accessible(uct_cuda_ipc_component_t *mdc,
-                                                    uct_cuda_ipc_rkey_t *rkey)
+static ucs_status_t
+uct_cuda_ipc_is_peer_accessible(uct_cuda_ipc_component_t *component,
+                                uct_cuda_ipc_rkey_t *rkey)
 {
     CUdevice this_device;
     ucs_status_t status;
-    int peer_idx;
-    int num_devices;
-    ucs_ternary_auto_value_t *accessible;
     void *d_mapped;
+    uct_cuda_ipc_dev_cache_t *cache;
+    uint8_t *accessible;
 
-    status = uct_cuda_ipc_get_unique_index_for_uuid(&peer_idx, mdc->md, rkey);
-    if (ucs_unlikely(status != UCS_OK)) {
+    status = UCT_CUDADRV_FUNC_LOG_DEBUG(cuCtxGetDevice(&this_device));
+    if (UCS_OK != status) {
+        return status;
+    }
+
+    pthread_mutex_lock(&component->lock);
+
+    cache = uct_cuda_ipc_get_dev_cache(component, &rkey->uuid);
+    if (ucs_unlikely(NULL == cache)) {
+        status = UCS_ERR_NO_RESOURCE;
         goto err;
     }
 
     /* overwrite dev_num with a unique ID; this means that relative remote
      * device number of multiple peers do not map on the same stream and reduces
      * stream sequentialization */
-    rkey->dev_num = peer_idx;
-
-    UCT_CUDA_IPC_DEVICE_GET_COUNT(num_devices);
-    if (UCT_CUDADRV_FUNC_LOG_DEBUG(cuCtxGetDevice(&this_device)) != UCS_OK) {
-        status = UCS_ERR_UNREACHABLE;
-        goto err;
-    }
-
-    accessible = &mdc->md->peer_accessible_cache[peer_idx * num_devices + this_device];
-    if (*accessible == UCS_TRY) { /* unchecked, add to cache */
+    rkey->dev_num = cache->dev_num;
+    accessible    = &cache->accessible[this_device];
+    if (ucs_unlikely(*accessible == UCS_TRY)) { /* unchecked, add to cache */
 
         /* Check if peer is reachable by trying to open memory handle. This is
          * necessary when the device is not visible through CUDA_VISIBLE_DEVICES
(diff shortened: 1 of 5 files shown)


src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -339,52 +355,50 @@ uct_cuda_ipc_md_open(uct_component_t *component, const char *md_name,
         .detect_memory_type = ucs_empty_function_return_unsupported
     };
 
-    uct_cuda_ipc_md_t* md;
-    uct_cuda_ipc_component_t* com;
-
-    md = ucs_calloc(1, sizeof(uct_cuda_ipc_md_t), "uct_cuda_ipc_md");
+    uct_md_t *md = ucs_calloc(1, sizeof(*md), "uct_cuda_ipc_md");
```
brminich, line 358: minor: typically we do not allocate mem as initialization, i'd just declare a var here

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -28,6 +29,94 @@ static ucs_config_field_t uct_cuda_ipc_md_config_table[] = {
...
+    return (khint32_t)(i64[0]);
+}
+
+KHASH_INIT(cuda_ipc_uuid_hash, CUuuid, uct_cuda_ipc_dev_cache_t*, 1,
+           uct_cuda_ipc_uuid_hash_func, uct_cuda_ipc_uuid_equals);
+
+typedef struct {
+    khash_t(cuda_ipc_uuid_hash) hash;
+    ucs_recursive_spinlock_t    lock;
+} uct_cuda_ipc_uuid_cache_t;
```
brminich, line 58: maybe make it a part of ipc component?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: ok

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -28,6 +29,94 @@ static ucs_config_field_t uct_cuda_ipc_md_config_table[] = {
...
+
+static uct_cuda_ipc_uuid_cache_t uct_cuda_ipc_uuid_cache;
+
+static uct_cuda_ipc_dev_cache_t *uct_cuda_ipc_create_dev_cache(int dev_num)
+{
+    uct_cuda_ipc_dev_cache_t *cache;
+    ucs_status_t status;
+    int i, num_devices;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGetCount(&num_devices));
```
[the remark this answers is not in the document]
iyastreb: The only reason is that `UCT_CUDA_IPC_DEVICE_GET_COUNT` returns ucs_status_t in case of an error, while this function returns a pointer to the created cache.
I could use this macro only if convert function to return status, in UCX code we do it both ways, so I preferred the one with shorter signature. But it you prefer to return a status - please let me know

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: ok, will be done

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -28,6 +29,94 @@ static ucs_config_field_t uct_cuda_ipc_md_config_table[] = {
...
+    if (UCS_OK != status) {
+        ucs_error("cuDeviceGetCount() failed: %s", ucs_status_string(status));
+        return NULL;
+    }
+
+    cache = ucs_malloc(sizeof(*cache) + num_devices,
```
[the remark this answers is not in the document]
yosefe: also add ( )

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -28,6 +29,94 @@ static ucs_config_field_t uct_cuda_ipc_md_config_table[] = {
...
+{
+    int64_t *a64 = (int64_t *)a.bytes;
+    int64_t *b64 = (int64_t *)b.bytes;
+    return (a64[0] == b64[0]) && (a64[1] == b64[1]);
+}
+
+static UCS_F_ALWAYS_INLINE khint32_t uct_cuda_ipc_uuid_hash_func(CUuuid key)
+{
+    int64_t *i64 = (int64_t *)key.bytes;
+    return (khint32_t)(i64[0]);
```
yosefe, line 49: maybe use other bits of the uuid in the hash?

test/gtest/uct/cuda/test_cuda_ipc_md.cc
```diff
@@ -0,0 +1,107 @@
...
+#include <cuda_runtime.h>
+
+extern "C" {
+#include <uct/cuda/cuda_ipc/cuda_ipc_md.h>
+}
+
+class test_cuda_ipc_md : public test_md {
+protected:
+    static uct_cuda_ipc_rkey_t
+    unpack(uct_md_h md, int64_t uuid0 = 0, int64_t uuid1 = 0)
```
yosefe, line 18: why need to set default value of uuid?
maybe pass it as CUuuid?

test/gtest/uct/cuda/test_cuda_ipc_md.cc
```diff
@@ -0,0 +1,107 @@
...
+        EXPECT_EQ(CUDA_SUCCESS, cuMemFree(ptr));
+        return rkey;
+    }
+};
+
+class cuda_context {
+public:
+    cuda_context()
+    {
+        if (cudaSetDevice(0) != cudaSuccess) {
```
yosefe, line 54: can it affect the cuda context set in https://github.com/openucx/ucx/blob/6b84982510a89631a215e4fc7b1e19693a89794e/test/gtest/common/mem_buffer.cc#L181

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -28,6 +29,94 @@ static ucs_config_field_t uct_cuda_ipc_md_config_table[] = {
...
+
+static uct_cuda_ipc_dev_cache_t *uct_cuda_ipc_get_dev_cache(const CUuuid *uuid)
+{
+    uct_cuda_ipc_dev_cache_t *cache   = NULL;
+    khash_t(cuda_ipc_uuid_hash) *hash = &uct_cuda_ipc_uuid_cache.hash;
+    khiter_t iter;
+    int ret;
+
+    iter = kh_put(cuda_ipc_uuid_hash, hash, *uuid, &ret);
+    switch (ret) {
```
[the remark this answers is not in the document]
iyastreb: ok

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -28,6 +29,94 @@ static ucs_config_field_t uct_cuda_ipc_md_config_table[] = {
...
+    case UCS_KH_PUT_BUCKET_EMPTY:
+    case UCS_KH_PUT_BUCKET_CLEAR:
+        cache = uct_cuda_ipc_create_dev_cache(kh_size(hash) - 1);
+        if (NULL != cache) {
+            kh_val(hash, iter) = cache;
+        } else {
+            kh_del(cuda_ipc_uuid_hash, hash, iter);
```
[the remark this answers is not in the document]
iyastreb: By this point in time we have added a new key in the hashmap with empty value.
Now we have 2 options with subtle difference in behavior:
1) we can either remove from the map a pair {key -> NULL}, which I think is more logical. The important point is that by removing we also decrease the hashmap size, which is used to assing unique ID for dev_num (see line 100). This way we maintain in the hashmap only entries with valid unique IDs.
2) we can keep pair {key -> NULL} in the map.
It will have some side effect on unique ID assignment: we might have gaps in the IDs -> suboptimal GPU distribution across the streams. On the other side, next time someone tries to create a new entry with this UUID, we take a null entry from cache and don't even try to create.

I thought that p1 is desirable behaviour, therefore this logic. If you think p2 is the way to go - please let me know. Not a big deal anyway

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -115,101 +204,34 @@ uct_cuda_ipc_mkey_pack(uct_md_h md, uct_mem_h tl_memh, void *address,
...
+    uint8_t *accessible;
+
+    status = UCT_CUDADRV_FUNC_LOG_DEBUG(cuCtxGetDevice(&this_device));
+    if (UCS_OK != status) {
+        return status;
+    }
 
-    status = uct_cuda_ipc_get_unique_index_for_uuid(&peer_idx, mdc->md, rkey);
-    if (ucs_unlikely(status != UCS_OK)) {
+    ucs_recursive_spin_lock(&uct_cuda_ipc_uuid_cache.lock);
```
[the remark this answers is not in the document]
iyastreb: no need, will use plain spinlock instead

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -339,52 +355,50 @@ uct_cuda_ipc_md_open(uct_component_t *component, const char *md_name,
...
+
+UCS_STATIC_CLEANUP {
+    uct_cuda_ipc_dev_cache_t *cache;
+
+    kh_foreach_value(&uct_cuda_ipc_uuid_cache.hash, cache, {
+        free(cache);
+    })
+    kh_destroy_inplace(cuda_ipc_uuid_hash, &uct_cuda_ipc_uuid_cache.hash);
+    ucs_recursive_spinlock_destroy(&uct_cuda_ipc_uuid_cache.lock);
+}
```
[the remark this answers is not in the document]
iyastreb: Right, I see the potential problem, these calls may be thrown away by a linker. Do we have the proper way to handle it?

I ask because I found this pattern used in cuda_ipc_cache.c, but it's also used in other places (event.c, vfs_obj.c), so my assumption was that it's a valid thing to use. If it's not the case, maybe we should come up with approach that handles all those cases?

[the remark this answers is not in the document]
yosefe: use KHASH_STATIC_INITIALIZER, make the lock pthread_mutex_t, and use PTHREAD_MUTEX_INITIALIZER

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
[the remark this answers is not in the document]
yosefe: i think p2 is less code and we won't try to add the same uuid over and over again

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -86,32 +55,29 @@ static uct_cuda_ipc_dev_cache_t *uct_cuda_ipc_create_dev_cache(int dev_num)
...
+    uct_cuda_ipc_dev_cache_t *cache;
     khiter_t iter;
     int ret;
 
     iter = kh_put(cuda_ipc_uuid_hash, hash, *uuid, &ret);
-    switch (ret) {
-    case UCS_KH_PUT_BUCKET_EMPTY:
-    case UCS_KH_PUT_BUCKET_CLEAR:
+    if (ret == UCS_KH_PUT_KEY_PRESENT) {
+        cache = kh_val(hash, iter);
```
yosefe, line 69: return kh_val(hash, iter);

we can add return in each if/else and eliminate the last return and make the code a bit simpler to follow

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -86,32 +55,29 @@ static uct_cuda_ipc_dev_cache_t *uct_cuda_ipc_create_dev_cache(int dev_num)
...
-
-    case UCS_KH_PUT_KEY_PRESENT:
-        cache = kh_val(hash, iter);
-        break;
-
-    default:
+    } else {
         ucs_error("kh_put(cuda_ipc_uuid_hash) failed with %d", ret);
-        break;
+        cache = NULL;
```
yosefe, line 80: return NULL

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -354,51 +323,57 @@ uct_cuda_ipc_md_open(uct_component_t *component, const char *md_name,
...
+        .flags              = 0,
+        .md_vfs_init        =
+                (uct_component_md_vfs_init_func_t)ucs_empty_function
     },
-    .cm_config          = UCS_CONFIG_EMPTY_GLOBAL_LIST_ENTRY,
-    .tl_list            = UCT_COMPONENT_TL_LIST_INITIALIZER(&uct_cuda_ipc_component),
-    .flags              = 0,
-    .md_vfs_init        = (uct_component_md_vfs_init_func_t)ucs_empty_function
+    .uuid_hash              = KHASH_STATIC_INITIALIZER,
+    .lock                   = {0}
```
yosefe, line 362: it's not correct on all systems, sometimes the unlocked value is -1
must switch to pthread_mutex_t

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
```diff
@@ -9,11 +9,43 @@
...
+{
+    int64_t *a64 = (int64_t *)a.bytes;
+    int64_t *b64 = (int64_t *)b.bytes;
+    return (a64[0] == b64[0]) && (a64[1] == b64[1]);
+}
+
+static UCS_F_ALWAYS_INLINE khint32_t uct_cuda_ipc_uuid_hash_func(CUuuid key)
+{
+    int64_t *i64 = (int64_t *)key.bytes;
+    return kh_int64_hash_func(i64[0] & i64[1]);
```
yosefe, line 33: use ^ instead of &
with &, the probability of 0 bits is higher than 1 bits so it's not uniform

iyastreb: sure, thanks, didn't know that

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: ok

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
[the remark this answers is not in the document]
iyastreb: done

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: ok

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: ok

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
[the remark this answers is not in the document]
iyastreb: ok

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -384,7 +355,17 @@ uct_cuda_ipc_component_t uct_cuda_ipc_component = {
...
+    .uuid_hash              = KHASH_STATIC_INITIALIZER,
+    .lock                   = PTHREAD_MUTEX_INITIALIZER
 };
 UCT_COMPONENT_REGISTER(&uct_cuda_ipc_component.super);
 
+UCS_STATIC_CLEANUP {
+    uct_cuda_ipc_dev_cache_t *cache;
+
+    kh_foreach_value(&uct_cuda_ipc_component.uuid_hash, cache, {
+        free(cache);
```
yosefe, line 367: ucs_free

iyastreb: sure

yosefe (approved):

Akshay-Venkatesh (approved):

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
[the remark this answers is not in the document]
iyastreb: This would not change anything, because internally khash uses `khint_t` (which is `typedef khint32_t khint_t`) as a hash.
https://github.com/openucx/ucx/blob/master/src/ucs/datastruct/khash.h#L263
Btw, in the last version the impl uses all the bits to compute the hash, but still returns 32 bit integer:
```
static UCS_F_ALWAYS_INLINE khint32_t uct_cuda_ipc_uuid_hash_func(CUuuid key)
{
    int64_t *i64 = (int64_t *)key.bytes;
    return kh_int64_hash_func(i64[0] ^ i64[1]);
}
```

PR #9867, opened 2024-05-09
UCT/CUDA: Detect fabric vmm memory

Akshay-Venkatesh:
## What/Why?
Breaking https://github.com/openucx/ucx/pull/9787 into parts. This part isolates:
1. Fabric memory detection in cuda_copy transport
2. Change UCM cuda hooks to treat mallocasync memory as CUDA memory and not CUDA managed memory.
3. Add ability to allocate fabric pinned memory for rendezvous fragments

## Notes
1. This will break existing mallocasync support which applies managed memory pipeline protocols
2. Currently EGM fabric VMM allocations are treated as device memory for subsequent PRs to leverage NVLINKs for EGM transfers. In a subsequent PR, this change will be reverted once it is confirmed that cuda_ipc supporting host memory doesn't break UCP protocolsV2.

PR #9787, opened 2024-03-29
UCT/CUDA_IPC: Add VMM/mallocasync support over fabric handles

Akshay-Venkatesh:
## What
Allow VMM/Mallocasync created by user with FABRIC handles to take advantage of cuda-ipc transport

PR #9982, opened 2024-06-28
UCT/CUDA_IPC: Fabric memory support

Akshay-Venkatesh:
## What/Why?
Second part of https://github.com/openucx/ucx/pull/9787 and follow up to https://github.com/openucx/ucx/pull/9867. This implements IPC-support for fabric handle associated memory using the newer import/export API and caches import operations similar to legacy IPC handles.

Note: need to solve https://redmine.mellanox.com/issues/4005305 before merging

### DIFF
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
@@ -77,6 +77,30 @@ uct_cuda_ipc_cache_region_collect_callback(const ucs_pgtable_t *pgtable,
     ucs_list_add_tail(list, &region->list);
 }
 
+static ucs_status_t uct_cuda_ipc_close_memhandle(uct_cuda_ipc_cache_region_t *region)
+{
+    ucs_status_t status;
+
+#if HAVE_CUDA_FABRIC
+    if (region->key.ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_VMM) {
+        status = UCT_CUDADRV_FUNC_LOG_WARN(cuMemUnmap(
+                    (CUdeviceptr)region->mapped_addr, region->key.b_len));
+        if (status == UCS_OK) {
+            return UCT_CUDADRV_FUNC_LOG_WARN(cuMemAddressFree(
+                        (CUdeviceptr)region->mapped_addr, region->key.b_len));
+        }
+    } else if (region->key.ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_MEMPOOL) {
+        return UCT_CUDADRV_FUNC_LOG_WARN(cuMemPoolDestroy(region->key.ph.pool));
+    } else
+#endif
+    {
+        return UCT_CUDADRV_FUNC_LOG_WARN(cuIpcCloseMemHandle(
+                    (CUdeviceptr)region->mapped_addr));
+    }
+
+    return status;
+}
+
 static void uct_cuda_ipc_cache_purge(uct_cuda_ipc_cache_t *cache)
 {
     int active = uct_cuda_base_is_context_active();
@@ -88,32 +112,152 @@ static void uct_cuda_ipc_cache_purge(uct_cuda_ipc_cache_t *cache)
                       &region_list);
     ucs_list_for_each_safe(region, tmp, &region_list, list) {
         if (active) {
-            UCT_CUDADRV_FUNC_LOG_ERR(
-                    cuIpcCloseMemHandle((CUdeviceptr)region->mapped_addr));
+            uct_cuda_ipc_close_memhandle(region);
         }
         ucs_free(region);
     }
     ucs_trace("%s: cuda ipc cache purged", cache->name);
 }
 
-static ucs_status_t uct_cuda_ipc_open_memhandle(const uct_cuda_ipc_rkey_t *key,
-                                                CUdeviceptr *mapped_addr)
+static ucs_status_t
+uct_cuda_ipc_open_memhandle_legacy(CUipcMemHandle memh,
+                                   CUdeviceptr *mapped_addr)
 {
     CUresult cuerr;
-    ucs_status_t status;
 
-    cuerr = cuIpcOpenMemHandle(mapped_addr, key->ph,
+    cuerr = cuIpcOpenMemHandle(mapped_addr, memh,
                                CU_IPC_MEM_LAZY_ENABLE_PEER_ACCESS);
-    if (cuerr == CUDA_SUCCESS) {
-        status = UCS_OK;
-    } else {
+    if (cuerr != CUDA_SUCCESS) {
         ucs_debug("cuIpcOpenMemHandle() failed: %s",
                   uct_cuda_base_cu_get_error_string(cuerr));
-        status = (cuerr == CUDA_ERROR_ALREADY_MAPPED) ? UCS_ERR_ALREADY_EXISTS :
-                                                        UCS_ERR_INVALID_PARAM;
+        return (cuerr == CUDA_ERROR_ALREADY_MAPPED) ?
+            UCS_ERR_ALREADY_EXISTS : UCS_ERR_INVALID_PARAM;
     }
 
+    return UCS_OK;
+}
+
+#if HAVE_CUDA_FABRIC
+static ucs_status_t uct_cuda_ipc_init_access_desc(CUmemAccessDesc *access_desc)
+{
+    access_desc->location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+    access_desc->flags         = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
+
+    return UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetDevice(&access_desc->location.id));
+}
+
+static ucs_status_t
+uct_cuda_ipc_open_memhandle_vmm(uct_cuda_ipc_rkey_t *key,
+                                CUdeviceptr *mapped_addr)
+{
+    CUmemAccessDesc access_desc = {};
+    ucs_status_t status;
+    CUdeviceptr dptr;
+    CUmemGenericAllocationHandle handle;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemImportFromShareableHandle(&handle,
+                (void*)&key->ph.handle.fabric_handle,
+                CU_MEM_HANDLE_TYPE_FABRIC));
+    if (status != UCS_OK) {
+        goto out;
+    }
+
+    status =
+        UCT_CUDADRV_FUNC_LOG_ERR(cuMemAddressReserve(&dptr, key->b_len, 0, 0, 0));
+    if (status != UCS_OK) {
+        goto release_handle;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemMap(dptr, key->b_len, 0, handle, 0));
+    if (status != UCS_OK) {
+        goto release_va_range;
+    }
+
+    status = uct_cuda_ipc_init_access_desc(&access_desc);
+    if (status != UCS_OK) {
+        goto unmap_range;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemSetAccess(dptr, key->b_len, &access_desc, 1));
+    if (status != UCS_OK) {
+        goto unmap_range;
+    }
+
+    *mapped_addr = dptr;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemRelease(handle));
+    if (status == UCS_OK) {
+        goto out;
+    }
+
+unmap_range:
+    cuMemUnmap(dptr, key->b_len);
+release_va_range:
+    cuMemAddressFree(dptr, key->b_len);
+release_handle:
+    cuMemRelease(handle);
+out:
     return status;
+}
+
+static ucs_status_t
+uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
+                                    CUdeviceptr *mapped_addr)
+{
+    CUmemAccessDesc access_desc = {};
+    CUdeviceptr dptr;
+    ucs_status_t status;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportFromShareableHandle(
+                &key->ph.pool, (void *)&key->ph.handle.fabric_handle,
+                CU_MEM_HANDLE_TYPE_FABRIC, 0));
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(&dptr,
+                key->ph.pool, (CUmemPoolPtrExportData*)&key->ph.ptr));
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    status = uct_cuda_ipc_init_access_desc(&access_desc);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolSetAccess(key->ph.pool,
+                &access_desc, 1));
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    *mapped_addr = dptr;
+
+    return UCS_OK;
+}
+#endif
+
+static ucs_status_t uct_cuda_ipc_open_memhandle(uct_cuda_ipc_rkey_t *key,
+                                                CUdeviceptr *mapped_addr)
+{
+
+#if HAVE_CUDA_FABRIC
+    ucs_trace("key handle type %u", key->ph.handle_type);
+
+    if (key->ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_LEGACY) {
+        return uct_cuda_ipc_open_memhandle_legacy(key->ph.handle.legacy,
+                                                  mapped_addr);
+    } else if (key->ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_VMM) {
+        return uct_cuda_ipc_open_memhandle_vmm(key, mapped_addr);
+    } else if (key->ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_MEMPOOL) {
+        return uct_cuda_ipc_open_memhandle_mempool(key, mapped_addr);
+    } else {
+        return UCS_ERR_INVALID_PARAM;
+    }
+#else
+    return uct_cuda_ipc_open_memhandle_legacy(key->ph, mapped_addr);
+#endif
 }
 
 static void uct_cuda_ipc_cache_invalidate_regions(uct_cuda_ipc_cache_t *cache,
@@ -122,6 +266,7 @@ static void uct_cuda_ipc_cache_invalidate_regions(uct_cuda_ipc_cache_t *cache,
     ucs_list_link_t region_list;
     ucs_status_t status;
     uct_cuda_ipc_cache_region_t *region, *tmp;
+    int handle_type;
 
     ucs_list_head_init(&region_list);
     ucs_pgtable_search_range(&cache->pgtable, (ucs_pgt_addr_t)from,
@@ -134,8 +279,19 @@ static void uct_cuda_ipc_cache_invalidate_regions(uct_cuda_ipc_cache_t *cache,
             ucs_error("failed to remove address:%p from cache (%s)",
                       (void *)region->key.d_bptr, ucs_status_string(status));
         }
-        UCT_CUDADRV_FUNC_LOG_ERR(
-                cuIpcCloseMemHandle((CUdeviceptr)region->mapped_addr));
+
+        status = uct_cuda_ipc_close_memhandle(region);
+        if (status != UCS_OK) {
+#if HAVE_CUDA_FABRIC
+            handle_type = region->key.ph.handle_type;
+#else
+            handle_type = 1; /* legacy memory type is the only valid type */
+#endif
+            ucs_error("failed to close memhandle for base addr:%p type:%d (%s)",
+                      (void *)region->key.d_bptr, handle_type,
+                      ucs_status_string(status));
+        }
+
         ucs_free(region);
     }
     ucs_trace("%s: closed memhandles in the range [%p..%p]",
@@ -215,8 +371,7 @@ ucs_status_t uct_cuda_ipc_unmap_memhandle(pid_t pid, uintptr_t d_bptr,
                       (void *)region->key.d_bptr, ucs_status_string(status));
         }
         ucs_assert(region->mapped_addr == mapped_addr);
-        status = UCT_CUDADRV_FUNC_LOG_ERR(
-                cuIpcCloseMemHandle((CUdeviceptr)region->mapped_addr));
+        status = uct_cuda_ipc_close_memhandle(region);
         ucs_free(region);
     }
 
@@ -225,13 +380,20 @@ ucs_status_t uct_cuda_ipc_unmap_memhandle(pid_t pid, uintptr_t d_bptr,
 }
 
 UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_ipc_map_memhandle, (key, mapped_addr),
-                 const uct_cuda_ipc_rkey_t *key, void **mapped_addr)
+                 uct_cuda_ipc_rkey_t *key, void **mapped_addr)
 {
     uct_cuda_ipc_cache_t *cache;
     ucs_status_t status;
     ucs_pgt_region_t *pgt_region;
     uct_cuda_ipc_cache_region_t *region;
     int ret;
+    size_t cmp_size;
+
+#if HAVE_CUDA_FABRIC
+    cmp_size = sizeof(key->ph.handle);
+#else
+    cmp_size = sizeof(key->ph);
+#endif
 
     status = uct_cuda_ipc_get_remote_cache(key->pid, &cache);
     if (status != UCS_OK) {
@@ -244,7 +406,7 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_ipc_map_memhandle, (key, mapped_addr),
     if (ucs_likely(pgt_region != NULL)) {
         region = ucs_derived_of(pgt_region, uct_cuda_ipc_cache_region_t);
         if (memcmp((const void *)&key->ph, (const void *)&region->key.ph,
-                   sizeof(key->ph)) == 0) {
+                   cmp_size) == 0) {
             /*cache hit */
             ucs_trace("%s: cuda_ipc cache hit addr:%p size:%lu region:"
                       UCS_PGT_REGION_FMT, cache->name, (void *)key->d_bptr,
@@ -269,8 +431,7 @@ UCS_PROFILE_FUNC(ucs_status_t, uct_cuda_ipc_map_memhandle, (key, mapped_addr),
             }
 
             /* close memhandle */
-            UCT_CUDADRV_FUNC_LOG_ERR(
-                    cuIpcCloseMemHandle((CUdeviceptr)region->mapped_addr));
+            uct_cuda_ipc_close_memhandle(region);
             ucs_free(region);
         }
     }
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
@@ -74,14 +74,55 @@ static ucs_status_t uct_cuda_ipc_iface_get_address(uct_iface_h tl_iface,
     return UCS_OK;
 }
 
+#if HAVE_CUDA_FABRIC
+static int uct_cuda_ipc_iface_is_mnnvl_supported(uct_cuda_ipc_md_t *md)
+{
+    CUdevice cu_device;
+    int coherent;
+    ucs_status_t status;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&cu_device, 0));
+    if (status != UCS_OK) {
+        return 0;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(
+            cuDeviceGetAttribute(&coherent,
+                                 CU_DEVICE_ATTRIBUTE_PAGEABLE_MEMORY_ACCESS_USES_HOST_PAGE_TABLES,
+                                 cu_device));
+    if (status != UCS_OK) {
+        return 0;
+    }
+
+    return coherent && (md->enable_mnnvl != UCS_NO);
+}
+#endif
+
 static int
 uct_cuda_ipc_iface_is_reachable_v2(const uct_iface_h tl_iface,
                                    const uct_iface_is_reachable_params_t *params)
 {
-    return uct_iface_is_reachable_params_addrs_valid(params) &&
-           (ucs_get_system_id() == *((const uint64_t*)params->device_addr)) &&
-           (getpid() != *(pid_t*)params->iface_addr) &&
-           uct_iface_scope_is_reachable(tl_iface, params);
+#if HAVE_CUDA_FABRIC
+    uct_base_iface_t *base_iface = ucs_derived_of(tl_iface, uct_base_iface_t);
+    uct_cuda_ipc_md_t *md        = ucs_derived_of(base_iface->md, uct_cuda_ipc_md_t);
+#endif
+
+    if (!uct_iface_is_reachable_params_addrs_valid(params) ||
+        (getpid() == *(pid_t*)params->iface_addr) ||
+        !uct_iface_scope_is_reachable(tl_iface, params)) {
+        return 0;
+    };
+
+#if HAVE_CUDA_FABRIC
+    if (uct_cuda_ipc_iface_is_mnnvl_supported(md)) {
+        /* multi-node nvlink is supported and enabled */
+        return 1;
+    }
+#endif
+
+    /* Not fabric capable or multi-node nvlink disabled, so iface has to be on
+     * the same node for cuda-ipc to be reachable */
+    return ucs_get_system_id() == *((const uint64_t*)params->device_addr);
 }
 
 static double uct_cuda_ipc_iface_get_bw()
@@ -556,10 +597,18 @@ static UCS_CLASS_CLEANUP_FUNC(uct_cuda_ipc_iface_t)
 
 ucs_status_t
 uct_cuda_ipc_query_devices(
-        uct_md_h md, uct_tl_device_resource_t **tl_devices_p,
+        uct_md_h uct_md, uct_tl_device_resource_t **tl_devices_p,
         unsigned *num_tl_devices_p)
 {
-    return uct_cuda_base_query_devices_common(md, UCT_DEVICE_TYPE_SHM,
+    uct_device_type_t dev_type = UCT_DEVICE_TYPE_SHM;
+#if HAVE_CUDA_FABRIC
+    uct_cuda_ipc_md_t *md      = ucs_derived_of(uct_md, uct_cuda_ipc_md_t);
+
+    if (uct_cuda_ipc_iface_is_mnnvl_supported(md)) {
+        dev_type = UCT_DEVICE_TYPE_NET;
+    }
+#endif
+    return uct_cuda_base_query_devices_common(uct_md, dev_type,
                                               tl_devices_p, num_tl_devices_p);
 }
 
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -25,6 +25,10 @@ static ucs_config_field_t uct_cuda_ipc_md_config_table[] = {
     {"", "", NULL,
      ucs_offsetof(uct_cuda_ipc_md_config_t, super), UCS_CONFIG_TYPE_TABLE(uct_md_config_table)},
 
+    {"ENABLE_MNNVL", "try",
+     "Enable multi-node NVLINK capabilities.",
+     ucs_offsetof(uct_cuda_ipc_md_config_t, enable_mnnvl), UCS_CONFIG_TYPE_TERNARY},
+
     {NULL}
 };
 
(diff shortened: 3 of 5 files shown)


src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -48,6 +48,76 @@ typedef struct uct_cuda_ipc_remote_cache {
...
+
+typedef struct uct_cuda_ipc_rem_mpool_cache {
+    khash_t(cuda_ipc_rem_mpool_cache) hash;
+    ucs_recursive_spinlock_t          lock;
+} uct_cuda_ipc_rem_mpool_cache_t;
+
+uct_cuda_ipc_rem_mpool_cache_t uct_cuda_ipc_rem_mpool_cache;
+
+static ucs_status_t
+uct_cuda_ipc_get_rem_mpool_cache(CUmemFabricHandle *fabric_handle,
```
yosefe, line 77: can we reuse the existing IPC remote cache instead of adding a new one?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -96,26 +166,137 @@ static void uct_cuda_ipc_cache_purge(uct_cuda_ipc_cache_t *cache)
...
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemSetAccess(dptr, key->b_len, &access_desc, 1));
+    if (status != UCS_OK) {
+        cuMemUnmap(dptr, key->b_len);
+        goto release_va_range;
+    }
+
+    *mapped_addr = dptr;
+
+    return UCT_CUDADRV_FUNC_LOG_ERR(cuMemRelease(handle));
```
yosefe, line 227: if cuMemRelease fails here, there is no cleanup

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -96,26 +166,137 @@ static void uct_cuda_ipc_cache_purge(uct_cuda_ipc_cache_t *cache)
...
+#if HAVE_CUDA_FABRIC
+    ucs_trace("key handle type %u", key->ph.handle_type);
+
+    if (key->ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_LEGACY) {
+        return uct_cuda_ipc_open_memhandle_legacy(key->ph.handle.legacy, mapped_addr);
+    } else if (key->ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_VMM) {
+        return uct_cuda_ipc_open_memhandle_vmm(key, mapped_addr);
+    } else if (key->ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_MEMPOOL) {
+        return uct_cuda_ipc_open_memhandle_mempool(key, mapped_addr);
+    } else {
```
yosefe, line 292: can we put only the non-legacy if..else parts under the ifdef/endif scope?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -50,21 +50,92 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
 {
     uct_cuda_ipc_lkey_t *key;
     ucs_status_t status;
+#if HAVE_CUDA_FABRIC
+    CUmemAllocationHandleType fabric_type = CU_MEM_HANDLE_TYPE_FABRIC;
```
yosefe, line 54: initialized vars should be first. maybe can avoid this init?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -50,21 +50,92 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
...
+            if (status == UCS_OK) {
+                status = UCT_CUDADRV_FUNC_LOG_ERR(cuPointerGetAttribute(
+                            &allowed_handle_types,
+                            CU_POINTER_ATTRIBUTE_ALLOWED_HANDLE_TYPES,
+                            (CUdeviceptr)addr));
+                if ((status != UCS_OK) ||
+                    (!(allowed_handle_types | fabric_type))) {
+                    status = UCS_ERR_NO_RESOURCE;
+                    goto err;
+                }
```
yosefe, line 95: why is this needed?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -50,21 +50,92 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
...
+    /* cuda_ipc can handle VMM, mallocasync, and legacy pinned device so need to
+     * pack appropriate handle */
+    status =
+        UCT_CUDADRV_FUNC(cuMemRetainAllocationHandle(&handle, addr), UCS_LOG_LEVEL_DIAG);
+    if (status != UCS_OK) {
+        /* Try legacy IPC first */
+        status =
+            UCT_CUDADRV_FUNC_LOG_DEBUG(cuIpcGetMemHandle(&key->ph.handle.legacy,
+                        (CUdeviceptr)addr));
+        if (status == UCS_OK) {
```
yosefe, line 80: can we unite common code with 133-137?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: IPC mechanism with memory pools involves two steps in the import phase. One expensive step to import the actual memory pool to allows local GPU to be capable of accessing locally mapped memory pool; and another low overhead step to import remote pointer. The exporting process is expected to make several allocations and deallocations from the memory pool and pack associated pointer and memory pool handle in the rkey. Importing process checks for remote memory pool handle and performs a mapping only when not present. 

The existing IPC remote cache relies on handle as key and returns a VA as value and directly uses offsets from the VA range for copy purposes. This doesn't translate to remote memory pool cache as we need imported memory and use that with the exported pointer [here](https://github.com/openucx/ucx/pull/9982/files#diff-561a7e4fa208245423bb161bc7a1cc05fbce0d65a10ef13c77e366f38c541826R256) to import associated VA. Hence there isn't a one-to-one mapping between the two remote caches.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: Will fix this.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: Will do

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
Akshay-Venkatesh: Ok will directly use the enum and remove the variable altogether.

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -50,21 +50,92 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
...
+    CUmemoryPool mempool;
+    int allowed_handle_types;
+#endif
 
     key = ucs_malloc(sizeof(*key), "uct_cuda_ipc_lkey_t");
     if (key == NULL) {
         return UCS_ERR_NO_MEMORY;
     }
 
+    memset(key, 0, sizeof(*key));
```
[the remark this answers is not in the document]
Akshay-Venkatesh: Will do

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
Akshay-Venkatesh: Will give that a shot.

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
Akshay-Venkatesh: Because user may allocate memory pool without fabric capabilities in which case cuda-ipc UCT cannot support fabric handle based IPC.

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
```diff
@@ -52,11 +52,34 @@ typedef struct {
...
+    UCT_CUDA_IPC_KEY_HANDLE_TYPE_VMM,
+    UCT_CUDA_IPC_KEY_HANDLE_TYPE_MEMPOOL
+} uct_cuda_ipc_key_handle_t;
+
+
+typedef struct uct_cuda_ipc_md_handle {
+    union {
+        CUipcMemHandle    legacy; /* Legacy IPC handle */
+        CUmemFabricHandle vmm;    /* VMM export handle */
+        CUmemFabricHandle mempool;/* MallocAsync handle */
```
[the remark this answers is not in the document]
Akshay-Venkatesh: Can have a common variable `fabric_handle`. Will change this.

Akshay-Venkatesh:
@yosefe / @brminich looks like tests are failing for reasons outside PR logic. Would you mind restarting failing tests?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -218,14 +219,18 @@ uct_cuda_ipc_open_memhandle_vmm(const uct_cuda_ipc_rkey_t *key,
...
-        goto release_va_range;
+        goto unmap_range;
     }
 
     *mapped_addr = dptr;
 
-    return UCT_CUDADRV_FUNC_LOG_ERR(cuMemRelease(handle));
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemRelease(handle));
+    if (status == UCS_OK) {
+        return status;
```
yosefe, line 229: goto line 238 so there will be single place where this function returns

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -280,21 +286,19 @@ uct_cuda_ipc_open_memhandle_mempool(const uct_cuda_ipc_rkey_t *key,
...
     } else if (key->ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_MEMPOOL) {
         return uct_cuda_ipc_open_memhandle_mempool(key, mapped_addr);
-    } else {
-        return UCS_ERR_NO_RESOURCE;
-    }
-#else
-    return uct_cuda_ipc_open_memhandle_legacy(key->ph, mapped_addr);
+    } else
 #endif
+    return uct_cuda_ipc_open_memhandle_legacy(*legacy_handle, mapped_addr);
```
yosefe, line 301: actually better revert, I didn't notice legacy_handle location is different depending on HAVE_CUDA_FABRIC

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -128,15 +89,56 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
...
         ucs_trace("packed vmm fabric handle for %p", addr);
+        goto common_path;
+    } else {
+        status = UCT_CUDADRV_FUNC_LOG_ERR(cuPointerGetAttribute(&mempool,
+                    CU_POINTER_ATTRIBUTE_MEMPOOL_HANDLE, (CUdeviceptr)addr));
+        if (status == UCS_OK) {
+            status = UCT_CUDADRV_FUNC_LOG_ERR(cuPointerGetAttribute(
+                        &allowed_handle_types,
+                        CU_POINTER_ATTRIBUTE_ALLOWED_HANDLE_TYPES,
+                        (CUdeviceptr)addr));
```
yosefe, line 100: maybe we should check this, before getting the mempool handle?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -128,15 +89,56 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
...
+            if ((status != UCS_OK) ||
+                (!(allowed_handle_types | CU_MEM_HANDLE_TYPE_FABRIC))) {
```
yosefe, line 102: should it be & here instead of | ?
if yes, seems this code was not tested, can we add test with different cuda mempool types?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -128,15 +89,56 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
...
+                status = UCS_ERR_NO_RESOURCE;
+                goto err;
+            }
+
+            status =
+                UCT_CUDADRV_FUNC_LOG_DEBUG(cuMemPoolExportToShareableHandle(
+                            (void *)&key->ph.handle.fabric_handle, mempool,
+                            CU_MEM_HANDLE_TYPE_FABRIC, 0));
+            if (UCS_OK != status) {
+                /* VMM, mempool queries failed. Try legacy IPC handle */
```
yosefe, line 112: seems we should get here only if CU_MEM_HANDLE_TYPE_FABRIC is included in allowed_handle_types, so the fallback to legacy should happen when checking allowed_handle_types and not here. Or no need to check allowed_handle_types at all.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
yosefe: maybe we could still unite them, if the hash key will contain the handle type?

Akshay-Venkatesh: @yosefe Today, we use exporter-side base pointer to lookup cache and perform import operation if base poiner isn't found. This works fine for legacy cuda-ipc because there is one independent backing for a given allocation. Two allocations can never have a common backing. So performing expensive import operation for each new unique base pointer is justified (we also need to check that the handle for given base pointer matches with cached handle to detect VA recycling). 

For mempools, many sub-allocations will have the same common backing (the memory pool). The first time we encounter a base pointer corresponding to a sub-allocation from the mempool, we import the whole mempool and then use that imported mempool and exported data pointer to get local translation of the associated base pointer of the sub-allocation at the importer side. If another sub-allocation from the same mempool is encountered at the importing process, if we use existing IPC remote cache logic, we will re-import an already imported mempool (paying unnecessary cost) and get local translation. Instead, it is better to check if a given mempool has been imported. Therefore we have a new cache that takes exported handle as input and returns imported handle as output. This way, any new sub-allocation from already imported memory pool will avoid import operation and just data pointer is imported. Since the old cache takes base pointer as input and the new cache introduced takes exporter handle as input, I don't see an immediate way of unifying the cache as we do need a common key type for a common cache.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -235,7 +235,7 @@ uct_cuda_ipc_open_memhandle_vmm(const uct_cuda_ipc_rkey_t *key,
     cuMemAddressFree(dptr, key->b_len);
 release_handle:
     cuMemRelease(handle);
-
+exit:
```
yosefe, line 238: rename to out:

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -93,25 +103,13 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
...
-                        &allowed_handle_types,
-                        CU_POINTER_ATTRIBUTE_ALLOWED_HANDLE_TYPES,
-                        (CUdeviceptr)addr));
-            if ((status != UCS_OK) ||
-                (!(allowed_handle_types | CU_MEM_HANDLE_TYPE_FABRIC))) {
-                status = UCS_ERR_NO_RESOURCE;
-                goto err;
-            }
-
+        if ((status == UCS_OK) && (mempool != 0)) {
```
yosefe, line 106: could mempool be 0 even if status is OK?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -125,9 +123,7 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
             ucs_trace("packed mempool handle and export pointer for %p", addr);
             goto common_path;
         } else {
-            /* VMM, mempool queries failed. Try legacy IPC handle */
-            key->ph.handle_type = UCT_CUDA_IPC_KEY_HANDLE_TYPE_LEGACY;
-            goto legacy_path;
+            goto err;
```
yosefe, line 126: maybe if (status!=OK) goto err? to avoid the else

Akshay-Venkatesh: @yosefe I'm assuming that this is the suggested change:
```diff
             key->ph.handle_type = UCT_CUDA_IPC_KEY_HANDLE_TYPE_MEMPOOL;
             ucs_trace("packed mempool handle and export pointer for %p", addr);
             goto common_path;
-        } else {
+        }
+
+        if (status != UCS_OK) {
             goto err;

```

This would miss the case where status is UCS_OK and mempool is 0. In this case too, we want to fail because legacy_ipc check, vmm check, and mempool check all failed. For this reason, shouldn't we keep the else case as it is?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -48,24 +48,93 @@ static ucs_status_t
...
 
-    status = UCT_CUDADRV_FUNC(cuIpcGetMemHandle(&key->ph, (CUdeviceptr)addr),
+    legacy_handle = (CUipcMemHandle*)&key->ph;
+    UCT_CUDADRV_FUNC_LOG_ERR(cuMemGetAddressRange(&key->d_bptr, &key->b_len,
+                (CUdeviceptr)addr));
+
+#if HAVE_CUDA_FABRIC
+    /* cuda_ipc can handle VMM, mallocasync, and legacy pinned device so need to
+     * pack appropriate handle */
+    legacy_handle = &key->ph.handle.legacy;
```
brminich, line 72: minor: maybe move it to line 78? i.e. initialize  if `legacy_capable` only

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
brminich: i think the idea is to check the following on line 106
```
if ((status != UCS_OK) || (mempool == 0)) {
    goto err;
}
```
this way else is not needed and nesting can be reduced

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -79,8 +79,10 @@ uct_cuda_ipc_iface_is_reachable_v2(const uct_iface_h tl_iface,
                                    const uct_iface_is_reachable_params_t *params)
 {
     return uct_iface_is_reachable_params_addrs_valid(params) &&
+#if !HAVE_CUDA_FABRIC
```
brminich, line 82: maybe we can introduce an env var which would manage MNNVL enablement (i.e. make devices on different nodes unreachable if it is disabled)

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -96,26 +166,145 @@ static void uct_cuda_ipc_cache_purge(uct_cuda_ipc_cache_t *cache)
...
+static ucs_status_t uct_cuda_ipc_open_memhandle(const uct_cuda_ipc_rkey_t *key,
+                                                CUdeviceptr *mapped_addr)
+{
+
+#if HAVE_CUDA_FABRIC
+    ucs_trace("key handle type %u", key->ph.handle_type);
+
+    if (key->ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_LEGACY) {
+        return uct_cuda_ipc_open_memhandle_legacy(key->ph.handle.legacy,
+                mapped_addr);
```
brminich, line 295: minor:
```suggestion
        return uct_cuda_ipc_open_memhandle_legacy(key->ph.handle.legacy,
                                                  mapped_addr);
```

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -48,6 +48,76 @@ typedef struct uct_cuda_ipc_remote_cache {
...
+            goto err;
+        }
+        kh_val(&uct_cuda_ipc_rem_mpool_cache.hash, khiter) = mpool;
+        *imported_mpool = mpool;
+        *key_present    = 0;
+    } else if (khret == UCS_KH_PUT_KEY_PRESENT) {
+        *imported_mpool = kh_val(&uct_cuda_ipc_rem_mpool_cache.hash, khiter);
+        *key_present    = 1;
+    } else {
+        ucs_error("unable to use cuda_ipc remote_cache hash");
```
brminich, line 111:
```suggestion
        ucs_error("unable to use cuda_ipc remote_cache hash: %d", khret);
```

brminich (commented):
a couple of minor styling comments

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
yosefe: so maybe convert it to assertion to show this isn't expected?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -561,8 +561,11 @@ uct_cuda_ipc_query_devices(
         uct_md_h md, uct_tl_device_resource_t **tl_devices_p,
         unsigned *num_tl_devices_p)
 {
+    uct_cuda_ipc_md_t *ipc_md = ucs_derived_of(md, uct_cuda_ipc_md_t);
 #if HAVE_CUDA_FABRIC
-    uct_device_type_t dev_type = UCT_DEVICE_TYPE_NET;
+    uct_device_type_t dev_type =
+        ((ipc_md->enable_mnnvl == UCS_TRY) || (ipc_md->enable_mnnvl == UCS_YES))
```
yosefe, line 567: why needed?
seems it does not really affect VMM memory allocation/usage in cuda_ipc flows..

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
```diff
@@ -22,6 +22,7 @@ typedef struct uct_cuda_ipc_md {
     ucs_ternary_auto_value_t *peer_accessible_cache;
     int                      uuid_map_size;
     int                      uuid_map_capacity;
+    int                      enable_mnnvl;
```
yosefe, line 25: ucs_ternary_auto_value_t

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
```diff
@@ -39,6 +40,7 @@ extern uct_cuda_ipc_component_t uct_cuda_ipc_component;
  */
 typedef struct uct_cuda_ipc_md_config {
     uct_md_config_t super;
+    int             enable_mnnvl;
```
yosefe, line 43: ucs_ternary_auto_value_t

yosefe:
@Akshay-Venkatesh also as we discussed let's remove the caching from this PR

Akshay-Venkatesh:
> @Akshay-Venkatesh also as we discussed let's remove the caching from this PR

@yosefe I've removed the cache now. Please let me know if the changes look good. As discussed offline, I'll create a follow up PR to add mempool cache logic back once this PR is approved.

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
Akshay-Venkatesh: @yosefe The motivation behind `enable_mnnvl` was to address [this ](https://github.com/openucx/ucx/pull/9982#discussion_r1671823426) request. 
So while this variable doesn't affect VMM allocation/usage, it does result in disabling MNNVL reachability. I've removed these changes for this PR and will address this separately.

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
Akshay-Venkatesh: Will fix this in the follow-up PR.

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
Akshay-Venkatesh: Will fix this in the follow-up PR.

Akshay-Venkatesh:
@yosefe As checks are passing and there aren't more comments, let me know if I can squash the comments. Thanks

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -48,24 +48,91 @@ static ucs_status_t
...
+
+        status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemRelease(handle));
+        if (status != UCS_OK) {
+            goto err;
+        }
+
+        key->ph.handle_type = UCT_CUDA_IPC_KEY_HANDLE_TYPE_VMM;
+        ucs_trace("packed vmm fabric handle for %p", addr);
+        goto common_path;
+    } else {
```
yosefe, line 103: so we assume that if cuMemRetainAllocationHandle failed, it's a mempool?
can we check it explicitly by some poster query or as part of cuPointerGetAttribute?\
in any case need to add documentation in the code that describes the logic (in high level)

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
```diff
@@ -52,27 +52,50 @@ typedef struct {
 } uct_cuda_ipc_memh_t;
 
 
+#if HAVE_CUDA_FABRIC
+typedef enum uct_cuda_ipc_key_handle {
+    UCT_CUDA_IPC_KEY_HANDLE_TYPE_ERROR = 0,
+    UCT_CUDA_IPC_KEY_HANDLE_TYPE_LEGACY,
+    UCT_CUDA_IPC_KEY_HANDLE_TYPE_VMM,
+    UCT_CUDA_IPC_KEY_HANDLE_TYPE_MEMPOOL
```
yosefe, line 60: pls document the enum values

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
```diff
@@ -52,27 +52,50 @@ typedef struct {
...
+
+
+typedef struct uct_cuda_ipc_md_handle {
+    union {
+        CUipcMemHandle        legacy;        /* Legacy IPC handle */
+        CUmemFabricHandle     fabric_handle; /* VMM/Mallocasync export handle */
+    } handle;
+    CUmemPoolPtrExportData    ptr;
+    CUmemoryPool              pool;
+    uct_cuda_ipc_key_handle_t handle_type;
```
yosefe, line 71: I'd move handle_type to be the first field in the struct, since it decides the usage of the rest of the fields

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
```diff
@@ -52,27 +52,50 @@ typedef struct {
...
  */
 typedef struct {
-    CUipcMemHandle  ph;      /* Memory handle of GPU memory */
-    CUdeviceptr     d_bptr;  /* Allocation base address */
-    size_t          b_len;   /* Allocation size */
-    ucs_list_link_t link;
+    uct_cuda_ipc_md_handle_t  ph;     /* Memory handle of GPU memory */
+    CUdeviceptr               d_bptr; /* Allocation base address */
+    size_t                    b_len;  /* Allocation size */
+    ucs_list_link_t           link;
```
yosefe, line 85: pls check clang-format

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
Akshay-Venkatesh: We conclude mempool if it's not legacy ipc memory (viz cudaMalloc), and it is not cuMemcreate/VMM memory (because retainAllocHandle failed).

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -48,24 +48,91 @@ static ucs_status_t
...
+                (CUdeviceptr)addr));
+
+#if HAVE_CUDA_FABRIC
+    /* cuda_ipc can handle VMM, mallocasync, and legacy pinned device so need to
+     * pack appropriate handle */
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuPointerGetAttribute(&legacy_capable,
+                CU_POINTER_ATTRIBUTE_IS_LEGACY_CUDA_IPC_CAPABLE,
+                (CUdeviceptr)addr));
+    if (legacy_capable) {
```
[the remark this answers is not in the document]
Akshay-Venkatesh: Legacy IPC (cudaMalloc) within the node, new IPC across nodes.

src/uct/cuda/cuda_ipc/cuda_ipc_md.h
Akshay-Venkatesh: I don't see issues with clang-format (https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=84170&view=logs&j=cc064a77-22b5-56bf-ecc0-70b5fe764261&t=aedfd754-44a6-53e6-6843-8659d800fee2&l=214)

Akshay-Venkatesh:
@yosefe the latest changes have `enable_mnnvl` env var added back and necessary reachability checks. Let me know if the changes look good. Thanks

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -77,6 +77,29 @@ uct_cuda_ipc_cache_region_collect_callback(const ucs_pgtable_t *pgtable,
     ucs_list_add_tail(list, &region->list);
 }
 
+static ucs_status_t uct_cuda_ipc_close_memhandle(uct_cuda_ipc_cache_region_t *region)
+{
+    ucs_status_t status;
```
yosefe, line 82: space line after

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -77,6 +77,29 @@ uct_cuda_ipc_cache_region_collect_callback(const ucs_pgtable_t *pgtable,
...
+        if (status == UCS_OK) {
+            return UCT_CUDADRV_FUNC_LOG_WARN(cuMemAddressFree(
+                        (CUdeviceptr)region->mapped_addr, region->key.b_len));
+        }
+    } else if (region->key.ph.handle_type == UCT_CUDA_IPC_KEY_HANDLE_TYPE_MEMPOOL) {
+        return UCT_CUDADRV_FUNC_LOG_WARN(cuMemPoolDestroy(region->key.ph.pool));
+    } else
+#endif
+    {
+        status = UCT_CUDADRV_FUNC_LOG_WARN(cuIpcCloseMemHandle(
```
yosefe, line 96: return ...

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -74,15 +74,46 @@ static ucs_status_t uct_cuda_ipc_iface_get_address(uct_iface_h tl_iface,
...
+    ucs_status_t status;
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&cu_device, 0));
+    if (status != UCS_OK) {
+        return 0;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(
+            cuDeviceGetAttribute(&coherent,
+                                 CU_DEVICE_ATTRIBUTE_PAGEABLE_MEMORY_ACCESS_USES_HOST_PAGE_TABLES,
```
yosefe, line 91: is this an indirect indication that MNNVL is supported?
is there a more direct check? like trying to allocate fabric memory?

Akshay-Venkatesh: This is an indirect check. Fabric memory allocation doesn't suffice. 

On non-MNNVL capable systems which have hardware coherence, the checks [here](https://github.com/openucx/ucx/pull/9982/files#diff-561a7e4fa208245423bb161bc7a1cc05fbce0d65a10ef13c77e366f38c541826R231) will suffice. This is where we require protov2 to fail gracefully and fallback to IB-based rendezvous transfers between nodes. 

This check was added mainly to avoid attempting MNNVL on non-coherent systems.

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -561,10 +592,13 @@ uct_cuda_ipc_query_devices(
         uct_md_h md, uct_tl_device_resource_t **tl_devices_p,
         unsigned *num_tl_devices_p)
 {
-#if HAVE_CUDA_FABRIC
-    uct_device_type_t dev_type = UCT_DEVICE_TYPE_NET;
-#else
     uct_device_type_t dev_type = UCT_DEVICE_TYPE_SHM;
+#if HAVE_CUDA_FABRIC
+    uct_cuda_ipc_md_t *ipc_md  = ucs_derived_of(md, uct_cuda_ipc_md_t);
```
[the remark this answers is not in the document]
Akshay-Venkatesh: md is already declared.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -134,8 +269,7 @@ static void uct_cuda_ipc_cache_invalidate_regions(uct_cuda_ipc_cache_t *cache,
             ucs_error("failed to remove address:%p from cache (%s)",
                       (void *)region->key.d_bptr, ucs_status_string(status));
         }
-        UCT_CUDADRV_FUNC_LOG_ERR(
-                cuIpcCloseMemHandle((CUdeviceptr)region->mapped_addr));
+        uct_cuda_ipc_close_memhandle(region);
```
brminich, line 272: why not checking return code?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -94,7 +94,7 @@ static int uct_cuda_ipc_iface_is_coherent()
         return 0;
     }
 
-    return coherent;
+    return (coherent && (md->enable_mnnvl != UCS_NO));
```
yosefe, line 97: pls remove the external ( )

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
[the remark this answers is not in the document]
yosefe: per convention in other places, let's rename the parameter from md to uct_md (and ipc_md to md)

Akshay-Venkatesh (commented):
addressed comments

brminich (dismissed):

yosefe (approved):

Akshay-Venkatesh:
@yosefe shall I squash commits?

Akshay-Venkatesh:
@yosefe and @brminich I created a branch https://github.com/openucx/ucx/pull/10036 by applying https://github.com/openucx/ucx/pull/9982.diff on top of current main branch. Squashing the commits in this branch proved painful after the merges in between. Hope it's ok  to ignore this and just merge https://github.com/openucx/ucx/pull/10036 into main branch.

yosefe:
@Akshay-Venkatesh it's better to squash and rebase in the current PR. Moving the changes to a new PR makes it harder to review the changes and track history.

Akshay-Venkatesh:
@yosefe 

@brminich provided the fix for  https://redmine.mellanox.com/issues/4005305 and IMB/OMB tests are passing now. I've pushed the commit now. Let me know if the changes look good and if there is a further squash needed.

FYI, there were some fixes needed in openmpi collective layers and memory detection layers as well. I've started PRs for those and those fixes are needed for OMB/IMB collective tests to pass.

Akshay-Venkatesh:
@yosefe/ @brminich  errors look unrelated. Would you mind restarting the tests?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -134,8 +278,13 @@ static void uct_cuda_ipc_cache_invalidate_regions(uct_cuda_ipc_cache_t *cache,
             ucs_error("failed to remove address:%p from cache (%s)",
                       (void *)region->key.d_bptr, ucs_status_string(status));
         }
-        UCT_CUDADRV_FUNC_LOG_ERR(
-                cuIpcCloseMemHandle((CUdeviceptr)region->mapped_addr));
+
+        status = uct_cuda_ipc_close_memhandle(region);
+        if (status != UCS_OK) {
+            ucs_error("failed to close memhandle for base addr:%p (%s)",
```
brminich, line 284: minor: maybe also print `region->key.ph.handle_type`

brminich (commented):
👍 besides minor comment

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -74,14 +74,52 @@ static ucs_status_t uct_cuda_ipc_iface_get_address(uct_iface_h tl_iface,
...
+                uct_iface_scope_is_reachable(tl_iface, params);
+
+#if HAVE_CUDA_FABRIC
+    if (uct_cuda_ipc_iface_is_mnnvl_supported(md)) {
+        return reachable;
+    }
+#endif
+
+    return reachable &&
+        (ucs_get_system_id() == *((const uint64_t*)params->device_addr));
```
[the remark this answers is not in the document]
brminich: this does not seem to be correct. If we have `HAVE_CUDA_FABRIC` and mnvl is not supported (e.g. disabled by env var), this function will always return 0

brminich (approved):

brminich:
error looks relevant
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I/__w/1/s/contrib/../src/uct/cuda -I../../.. -DCPU_FLAGS= -I/__w/1/s/contrib/../src -I/home/swx-azure-svc_azpcontainer/85835-20240814.18/build -I/home/swx-azure-svc_azpcontainer/85835-20240814.18/build/src -O3 -g -Wall -Werror -funwind-tables -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-label -Wno-long-long -Wno-endif-labels -Wno-sign-compare -Wno-multichar -Wno-deprecated-declarations -Winvalid-pch -Wno-language-extension-token -fno-finite-math-only -Wno-recommended-option -Wno-c99-extensions -Wno-pointer-sign -Werror-implicit-function-declaration -Wno-format-zero-length -Wnested-externs -Wshadow -Werror=declaration-after-statement -MT cuda_ipc/libuct_cuda_la-cuda_ipc_cache.lo -MD -MP -MF cuda_ipc/.deps/libuct_cuda_la-cuda_ipc_cache.Tpo -c /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c  -fPIC -DPIC -DUCX_SHARED_LIB -o cuda_ipc/.libs/libuct_cuda_la-cuda_ipc_cache.o
In file included from /__w/1/s/contrib/../src/ucs/debug/log.h:14:0,
                 from /__w/1/s/contrib/../src/uct/base/uct_iface.h:20,
                 from /__w/1/s/contrib/../src/uct/cuda/base/cuda_md.h:9,
                 from /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_md.h:10,
                 from /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.h:14,
                 from /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c:11:
/__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c: In function 'uct_cuda_ipc_cache_invalidate_regions':
/__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c:285:65: error: 'uct_cuda_ipc_md_handle_t' has no member named 'handle_type'
                       (void *)region->key.d_bptr, region->key.ph.handle_type,
                                                                 ^
/__w/1/s/contrib/../src/ucs/debug/log_def.h:35:84: note: in definition of macro 'ucs_log_component'
                              (ucs_log_level_t)(_level), _comp_log_config, _fmt, ## __VA_ARGS__); \
                                                                                    ^
/__w/1/s/contrib/../src/ucs/debug/log_def.h:44:37: note: in expansion of macro 'ucs_log'
 #define ucs_error(_fmt, ...)        ucs_log(UCS_LOG_LEVEL_ERROR, _fmt, ## __VA_ARGS__)
                                     ^
/__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c:284:13: note: in expansion of macro 'ucs_error'
             ucs_error("failed to close memhandle for base addr:%p type:%d (%s)",
             ^
/__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c: At top level:
cc1: error: unrecognized command line option "-Wno-c99-extensions" [-Werror]
```

Akshay-Venkatesh:
> error looks relevant
> 
> ```
> libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I/__w/1/s/contrib/../src/uct/cuda -I../../.. -DCPU_FLAGS= -I/__w/1/s/contrib/../src -I/home/swx-azure-svc_azpcontainer/85835-20240814.18/build -I/home/swx-azure-svc_azpcontainer/85835-20240814.18/build/src -O3 -g -Wall -Werror -funwind-tables -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-label -Wno-long-long -Wno-endif-labels -Wno-sign-compare -Wno-multichar -Wno-deprecated-declarations -Winvalid-pch -Wno-language-extension-token -fno-finite-math-only -Wno-recommended-option -Wno-c99-extensions -Wno-pointer-sign -Werror-implicit-function-declaration -Wno-format-zero-length -Wnested-externs -Wshadow -Werror=declaration-after-statement -MT cuda_ipc/libuct_cuda_la-cuda_ipc_cache.lo -MD -MP -MF cuda_ipc/.deps/libuct_cuda_la-cuda_ipc_cache.Tpo -c /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c  -fPIC -DPIC -DUCX_SHARED_LIB -o cuda_ipc/.libs/libuct_cuda_la-cuda_ipc_cache.o
> In file included from /__w/1/s/contrib/../src/ucs/debug/log.h:14:0,
>                  from /__w/1/s/contrib/../src/uct/base/uct_iface.h:20,
>                  from /__w/1/s/contrib/../src/uct/cuda/base/cuda_md.h:9,
>                  from /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_md.h:10,
>                  from /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.h:14,
>                  from /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c:11:
> /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c: In function 'uct_cuda_ipc_cache_invalidate_regions':
> /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c:285:65: error: 'uct_cuda_ipc_md_handle_t' has no member named 'handle_type'
>                        (void *)region->key.d_bptr, region->key.ph.handle_type,
>                                                                  ^
> /__w/1/s/contrib/../src/ucs/debug/log_def.h:35:84: note: in definition of macro 'ucs_log_component'
>                               (ucs_log_level_t)(_level), _comp_log_config, _fmt, ## __VA_ARGS__); \
>                                                                                     ^
> /__w/1/s/contrib/../src/ucs/debug/log_def.h:44:37: note: in expansion of macro 'ucs_log'
>  #define ucs_error(_fmt, ...)        ucs_log(UCS_LOG_LEVEL_ERROR, _fmt, ## __VA_ARGS__)
>                                      ^
> /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c:284:13: note: in expansion of macro 'ucs_error'
>              ucs_error("failed to close memhandle for base addr:%p type:%d (%s)",
>              ^
> /__w/1/s/contrib/../src/uct/cuda/cuda_ipc/cuda_ipc_cache.c: At top level:
> cc1: error: unrecognized command line option "-Wno-c99-extensions" [-Werror]
> ```

@brminich that `handle_type` field was missing a guard. Can you please approve/reject again?

brminich (approved):

Akshay-Venkatesh:
@yosefe gentle ping to check if latest commits look good

yosefe (approved):

yosefe (approved):

Akshay-Venkatesh:
@yosefe @brminich These errors look unrelated.

Issue #9890, opened 2024-05-21
UCX 1.17.0RC1 gtest hang


edgargabriel:
### Describe the bug
While testing UCX 1.17.0RC1 we hit a problem with UCX gtest, which seems to end up in an endless loop during initialization and hangs.

### Setup and versions
- OS version (e.g Linux distro) + CPU architecture (x86_64/aarch64/ppc64le/...)
  ubuntu 22.04, x86_64 AMD EPYC 7V12
- For RDMA/IB/RoCE related issues:
    - Driver version:
```
$ sudo dpkg -s rdma-core
Package: rdma-core
Status: install ok installed
Priority: optional
Section: net
Installed-Size: 229
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: amd64
Version: 39.0-1

$ ibstat
CA 'rocep177s0'
        CA type: MT4117
        Number of ports: 1
        Firmware version: 14.25.8102
        Hardware version: 0
        Node GUID: 0xf46b8c6c2f580001
        System image GUID: 0xf46b8c6c2f580001
        Port 1:
                State: Active
                Physical state: LinkUp
                Rate: 40
                Base lid: 0
                LMC: 0
                SM lid: 0
                Capability mask: 0x00010000
                Port GUID: 0xf66b8cfffe6c2f58
                Link layer: Ethernet
```

### Additional information (depending on the issue)
- Output of `ucx_info -d`  attached to ticket

edgargabriel:
[ucx_info.txt](https://github.com/openucx/ucx/files/15390891/ucx_info.txt)

shasson5:
@edgargabriel can specify which gtest are you running? (with full command line)

shasson5:
@evgeny-leksikov can check?

edgargabriel:
Any command line argument will generate the issue on this platform, include just doing ```./gtest```

gtest gets stuck ```uct_test_base::enum_resources()``` (resp. ```uct_md_query_tl_resources``` invoked from there) for the ```gga_mlx5``` tl. For whatever reason the ```component->tl_list``` seems to be ill formed and the loop at https://github.com/openucx/ucx/blob/master/src/uct/base/uct_md.c#L93 never terminates. (I tried to count the number of entries in the tl_list by inserting a ```ucs_list_length()``` in the code, but ```ucs_list_length``` also doesn't return, so its really the list that has an issue)

[...]

PR #9984, opened 2024-07-01
GTEST: disable-new-dtags

evgeny-leksikov:
## What
disable-new-dtags to use RPATH instead of RUN_PATH to avoid loading of wrong libucp.so if it's accessible by LD_LIBRARY_PATH

## Why ?
fixes https://github.com/openucx/ucx/issues/9890

## How ?
add the parameter to the linker command

### DIFF
--- a/test/gtest/Makefile.am
+++ b/test/gtest/Makefile.am
@@ -82,6 +82,11 @@ gtest_CPPFLAGS = \
 	$(OPENMP_CFLAGS)
 
 gtest_LDFLAGS  = $(GTEST_LDFLAGS) -no-install $(LDFLAGS_DYNAMIC_LIST_DATA)
+
+# disable-new-dtags to force using RPATH instead of RUNPATH to avoid loading of
+# wrong shared libraries accessible by LD_LIBRARY_PATH
+gtest_LDFLAGS += -Wl,--disable-new-dtags
+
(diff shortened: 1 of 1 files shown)


test/gtest/Makefile.am
```diff
@@ -81,7 +81,9 @@ gtest_CPPFLAGS = \
 	$(GTEST_CPPFLAGS) \
 	$(OPENMP_CFLAGS)
 
-gtest_LDFLAGS  = $(GTEST_LDFLAGS) -no-install $(LDFLAGS_DYNAMIC_LIST_DATA)
+gtest_LDFLAGS  = \
+	$(GTEST_LDFLAGS) -no-install $(LDFLAGS_DYNAMIC_LIST_DATA) \
+	-Wl,--disable-new-dtags
```
yosefe, line 86: maybe add comment why we need it

evgeny-leksikov:
@edgargabriel could you please check the fix for https://github.com/openucx/ucx/issues/9890?

edgargabriel:
> @edgargabriel could you please check the fix for #9890?

will do, it might take me a day to get to it however

edgargabriel:
@evgeny-leksikov I can confirm that this pr fixes the gtest issue that observed on that particular platform. Thank you!

PR #9986, opened 2024-07-01
UCT/IB/MLX5: suggest user to increase PF_LOG_BAR_SIZE when UAR allocation fails

michal-shalev:
## What ?
Improve error logging for UAR allocation failure in `uct_ib_mlx5_devx_alloc_uar`.

## Why ?
To inform users to increase `PF_LOG_BAR_SIZE` and reboot when UAR allocation fails due to insufficient memory.
see also: https://github.com/open-mpi/ompi/issues/6854

## How ?
Update the error message in `uct_ib_mlx5_devx_alloc_uar` to suggest increasing `PF_LOG_BAR_SIZE` and mention the reboot requirement.

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -1883,20 +1883,6 @@ static void uct_ib_mlx5_md_port_counter_set_id_init(uct_ib_mlx5_md_t *md)
     }
 }
 
-static int uct_ib_mlx5_check_uar(uct_ib_mlx5_md_t *md)
-{
-    uct_ib_mlx5_devx_uar_t uar;
-    ucs_status_t status;
-
-    status = uct_ib_mlx5_devx_uar_init(&uar, md, 0);
-    if (status != UCS_OK) {
-        return UCS_ERR_UNSUPPORTED;
-    }
-
-    uct_ib_mlx5_devx_uar_cleanup(&uar);
-    return UCS_OK;
-}
-
 ucs_status_t
 uct_ib_mlx5_devx_device_mem_alloc(uct_md_h uct_md, size_t *length_p,
                                   void **address_p, ucs_memory_type_t mem_type,
@@ -2127,7 +2113,7 @@ ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
         goto err_free_context;
     }
 
-    status = uct_ib_mlx5_check_uar(md);
+    status = uct_ib_mlx5_devx_check_uar(md);
     if (status != UCS_OK) {
         goto err_free_md;
     }
--- a/src/uct/ib/mlx5/ib_mlx5.c
+++ b/src/uct/ib/mlx5/ib_mlx5.c
@@ -543,31 +543,91 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
 }
 
 #if HAVE_DEVX
-static ucs_status_t
-uct_ib_mlx5_devx_alloc_uar(uct_ib_mlx5_md_t *md, unsigned flags, int log_level,
-                           char *title, char *fallback,
-                           struct mlx5dv_devx_uar **uar_p)
+static ucs_status_t uct_ib_mlx5_devx_alloc_uar(uct_ib_mlx5_md_t *md,
+                                               uint32_t flags,
+                                               struct mlx5dv_devx_uar **uar_p)
 {
+    const char *uar_type_str      = (flags == UCT_IB_MLX5_UAR_ALLOC_TYPE_WC) ?
+                                    "WC" : "NC_DEDICATED";
+    ucs_log_level_t err_log_level = UCS_LOG_LEVEL_DIAG;
+    UCS_STRING_BUFFER_ONSTACK(strb, 256);
     struct mlx5dv_devx_uar *uar;
-    char buf[512];
+    ucs_status_t status;
+    int err;
 
     uar = mlx5dv_devx_alloc_uar(md->super.dev.ibv_context, flags);
-    if (uar == NULL) {
-        sprintf(buf, "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x(%s)) "
-                "failed: %m", uct_ib_device_name(&md->super.dev), flags, title);
-        if (fallback == NULL) {
-            ucs_log(log_level, "%s", buf);
-        } else {
-            ucs_log(log_level, "%s, fallback to %s", buf, fallback);
-        }
-
-        return UCS_ERR_NO_MEMORY;
+    if (uar != NULL) {
+        *uar_p = uar;
+        return UCS_OK;
     }
 
-    *uar_p = uar;
-    return UCS_OK;
+    err = errno;
+    ucs_string_buffer_appendf(&strb,
+                              "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x)"
+                              "type=%s failed: %s. ",
+                              uct_ib_device_name(&md->super.dev),
+                              flags,
+                              uar_type_str,
+                              strerror(err));
+    switch (err) {
+    case ENOMEM:
+        ucs_string_buffer_appendf(&strb,
+                                  "Consider increasing PF_LOG_BAR_SIZE "
+                                  "using mlxconfig tool (requires reboot)");
+        err_log_level = UCS_LOG_LEVEL_ERROR;
+        status        = UCS_ERR_NO_MEMORY;
+        break;
+    case EOPNOTSUPP:
+        status = UCS_ERR_UNSUPPORTED;
+        break;
+    case EINVAL:
+        status = UCS_ERR_INVALID_PARAM;
+        break;
+    default:
+        status = UCS_ERR_NO_MEMORY;
+        break;
+    }
+
+    ucs_log(err_log_level, "%s", ucs_string_buffer_cstr(&strb));
+    return status;
 }
 #endif
+
+ucs_status_t uct_ib_mlx5_devx_check_uar(uct_ib_mlx5_md_t *md)
+{
+#if HAVE_DEVX
+    uct_ib_mlx5_devx_uar_t uar;
+    ucs_status_t status;
+
+    status = uct_ib_mlx5_devx_alloc_uar(md,
+                                        UCT_IB_MLX5_UAR_ALLOC_TYPE_WC,
+                                        &uar.uar);
+    if (status == UCS_ERR_UNSUPPORTED) {
+        status = uct_ib_mlx5_devx_alloc_uar(md,
+                                            UCT_IB_MLX5_UAR_ALLOC_TYPE_NC,
+                                            &uar.uar);
+        if (status == UCS_ERR_UNSUPPORTED) {
+            ucs_error("%s: both WC and NC_DEDICATED UAR allocation types "
+                      " are not supported", uct_ib_device_name(&md->super.dev));
+            return status;
+        } else if (status != UCS_OK) {
+            return status;
+        }
+        /* NC_DEDICATED is supported - the flag is automatically set to 0 */
+    } else if (status != UCS_OK) {
+        /* The error is unrelated to the UAR allocation type, no fallback */
+        return status;
+    } else {
+        /* WC is supported - set the flag to 1 */
+        md->flags |= UCT_IB_MLX5_MD_FLAG_UAR_USE_WC;
+    }
+
+    uct_ib_mlx5_devx_uar_cleanup(&uar);
+    return UCS_OK;
+#else
+    return UCS_ERR_UNSUPPORTED;
+#endif
+}
 
 ucs_status_t uct_ib_mlx5_devx_uar_init(uct_ib_mlx5_devx_uar_t *uar,
                                        uct_ib_mlx5_md_t *md,
@@ -575,16 +635,16 @@ ucs_status_t uct_ib_mlx5_devx_uar_init(uct_ib_mlx5_devx_uar_t *uar,
 {
 #if HAVE_DEVX
     ucs_status_t status;
+    uint32_t flags;
 
-    status = uct_ib_mlx5_devx_alloc_uar(md, UCT_IB_MLX5_UAR_ALLOC_TYPE_WC,
-                                        UCS_LOG_LEVEL_DEBUG, "WC", "NC",
-                                        &uar->uar);
-    if (status != UCS_OK) {
-        status = uct_ib_mlx5_devx_alloc_uar(md, UCT_IB_MLX5_UAR_ALLOC_TYPE_NC,
-                                            UCS_LOG_LEVEL_ERROR, "NC", NULL,
-                                            &uar->uar);
+    /* Use UCT_IB_MLX5_MD_FLAG_UAR_USE_WC to determine the supported UAR allocation type */
+    if (md->flags & UCT_IB_MLX5_MD_FLAG_UAR_USE_WC) {
+        flags = UCT_IB_MLX5_UAR_ALLOC_TYPE_WC;
+    } else {
+        flags = UCT_IB_MLX5_UAR_ALLOC_TYPE_NC;
     }
 
+    status = uct_ib_mlx5_devx_alloc_uar(md, flags, &uar->uar);
     if (status != UCS_OK) {
         return status;
     }
--- a/src/uct/ib/mlx5/ib_mlx5.h
+++ b/src/uct/ib/mlx5/ib_mlx5.h
@@ -78,10 +78,10 @@
 #  define UCT_IB_MLX5_UAR_ALLOC_TYPE_WC 0x0
 #endif
 
-#if HAVE_DECL_MLX5DV_UAR_ALLOC_TYPE_NC
-#  define UCT_IB_MLX5_UAR_ALLOC_TYPE_NC MLX5DV_UAR_ALLOC_TYPE_NC
+#if HAVE_DECL_MLX5DV_UAR_ALLOC_TYPE_NC_DEDICATED
+#  define UCT_IB_MLX5_UAR_ALLOC_TYPE_NC MLX5DV_UAR_ALLOC_TYPE_NC_DEDICATED
 #else
-#  define UCT_IB_MLX5_UAR_ALLOC_TYPE_NC 0x1
+#  define UCT_IB_MLX5_UAR_ALLOC_TYPE_NC (1U << 31)
 #endif
 
 #define UCT_IB_MLX5_OPMOD_EXT_ATOMIC(_log_arg_size) \
@@ -197,9 +197,11 @@ enum {
     UCT_IB_MLX5_MD_FLAG_MMO_DMA              = UCS_BIT(15),
     /* Device supports XGVMI UMR workflow */
     UCT_IB_MLX5_MD_FLAG_XGVMI_UMR            = UCS_BIT(16),
+    /* Device supports UAR WC allocation type */
+    UCT_IB_MLX5_MD_FLAG_UAR_USE_WC           = UCS_BIT(17),
 
     /* Object to be created by DevX */
-    UCT_IB_MLX5_MD_FLAG_DEVX_OBJS_SHIFT  = 17,
+    UCT_IB_MLX5_MD_FLAG_DEVX_OBJS_SHIFT  = 18,
     UCT_IB_MLX5_MD_FLAG_DEVX_RC_QP       = UCT_IB_MLX5_MD_FLAG_DEVX_OBJS(RCQP),
     UCT_IB_MLX5_MD_FLAG_DEVX_RC_SRQ      = UCT_IB_MLX5_MD_FLAG_DEVX_OBJS(RCSRQ),
     UCT_IB_MLX5_MD_FLAG_DEVX_DCT         = UCT_IB_MLX5_MD_FLAG_DEVX_OBJS(DCT),
@@ -857,6 +859,8 @@ void uct_ib_mlx5_verbs_srq_cleanup(uct_ib_mlx5_srq_t *srq, struct ibv_srq *verbs
 int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
                              uct_ib_mlx5_md_t *md,
                              uct_ib_mlx5_mmio_mode_t mmio_mode);
+
+ucs_status_t uct_ib_mlx5_devx_check_uar(uct_ib_mlx5_md_t *md);
 
 ucs_status_t uct_ib_mlx5_devx_uar_init(uct_ib_mlx5_devx_uar_t *uar,
                                        uct_ib_mlx5_md_t *md,
--- a/src/uct/ib/configure.m4
+++ b/src/uct/ib/configure.m4
@@ -170,7 +170,7 @@ AS_IF([test "x$with_ib" = "xyes"],
                            MLX5DV_CQ_INIT_ATTR_MASK_CQE_SIZE,
                            MLX5DV_QP_CREATE_ALLOW_SCATTER_TO_CQE,
                            MLX5DV_UAR_ALLOC_TYPE_BF,
-                           MLX5DV_UAR_ALLOC_TYPE_NC,
+                           MLX5DV_UAR_ALLOC_TYPE_NC_DEDICATED,
                            mlx5dv_devx_umem_reg_ex],
                                   [], [], [[#include <infiniband/mlx5dv.h>]])
                        AC_CHECK_MEMBERS([struct mlx5dv_cq.cq_uar],


src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -554,7 +554,9 @@ uct_ib_mlx5_devx_alloc_uar(uct_ib_mlx5_md_t *md, unsigned flags, int log_level,
     uar = mlx5dv_devx_alloc_uar(md->super.dev.ibv_context, flags);
     if (uar == NULL) {
         sprintf(buf, "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x(%s)) "
-                "failed: %m", uct_ib_device_name(&md->super.dev), flags, title);
+                "failed: %m. This may be due to insufficient UAR space. "
+                "Consider increasing PF_LOG_BAR_SIZE from the default value of 5.",
```
yosefe, line 558: 1. "Consider increasing PF_LOG_BAR_SIZE using mlxconfig tool",
2. should we also check errno to have more specific error condition?
3. i think there could be other places that may fail because of the same reason, for example "mlx5dv_devx_obj_create(CQ)"

michal-shalev: 1. added the "mlxconfig tool" to the string.
2. do you mean check "ENOMEM"?
3. added this, checking for other places

michal-shalev: done

michal-shalev: 2. It looks like `mlx5dv_devx_alloc_uar` -> `_mlx5dv_devx_alloc_uar` -> `mlx5_attach_dedicated_uar` -> `mlx5_alloc_dyn_uar` -> `execute_ioctl` -> `ioctl`
it returns the `errno` but I don't know if we can know for sure that it'll set `ENOMEM` when `PF_LOG_BAR_SIZE` is insufficient. 
3. `mlx5dv_devx_obj_create(CQ)` doesn't fail because of `PF_LOG_BAR_SIZE`, in the email "
After applying the PF_LOG_BAR_SIZE parameter update".

**Regarding non-DEVX flows**
`uct_worker_tl_data_get` -> `uct_ib_mlx5_iface_get_res_domain` -> `uct_ib_mlx5_res_domain_init` -> `ibv_alloc_td` -> `mlx5_alloc_td` -> `mlx5_attach_dedicated_uar`-> `mlx5_alloc_dyn_uar` -> `execute_ioctl` -> `ioctl`
If that fails due to insufficient `PF_LOG_BAR_SIZE`, `uct_ib_mlx5_res_domain_init`  function does not fail. Instead, it returns `UCS_OK` and logs a debug message. 
The fallback mechanism uses the PD directly from the MD structure. This ensures continued operation, though with potentially reduced performance, especially for users sending a high volume of messages to different destinations. 
This is because the shared UAR may not handle the high message rate as efficiently.
I can add this if `ibv_alloc_td` fails in `uct_ib_mlx5_res_domain_init`:
`ucs_log(log_level, "Consider increasing PF_LOG_BAR_SIZE using mlxconfig tool");
`But it might not be noticeable, because entire flow doesn't fail, leading to missed logs for affected users.
Another suggestion - change the log level from `ucs_debug` to `ucs_warn`:
`ucs_warn("ibv_alloc_td() on %s failed: %m. Falling back to using the PD directly from the MD structure, which may reduce performance.", uct_ib_device_name(&md->dev))`
and then the first suggestion would be more noticeable.

gleon99: @michal-shalev  per our offline discussion,
- (2) ENOMEM - please check whether it indeed returns.
- Other flows - IMO generally we shouldn't change the behavior in this PR. Flows should not turn from failing to succeeding as a result of this PR, neither logging levels should change. Only add a suggestion for the user, in the already existing log level.

michal-shalev: @gleon99 I'll write a test that calls `mlx5dv_devx_alloc_uar` until it fails and checks the errno to make sure what it returns exactly.
Other flows - I'll add a suggestion.

michal-shalev: Ready for re-review @gleon99 @Artemy-Mellanox

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,25 +543,56 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+static ucs_status_t uct_ib_mlx5_devx_alloc_uar(uct_ib_mlx5_md_t *md, 
+                                               uint32_t flags, 
+                                               const char *allocation_type, 
+                                               const char *fallback,
+                                               struct mlx5dv_devx_uar **uar_p)
 {
     struct mlx5dv_devx_uar *uar;
-    char buf[512];
+    ucs_status_t status;
+    const char *pf_log_bar_size_suggestion = ". Consider increasing PF_LOG_BAR_SIZE using mlxconfig tool (requires reboot)";
```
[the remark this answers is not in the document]
gleon99: Maybe rename to `increase_log_bar_msg` or even `increase_bar_msg`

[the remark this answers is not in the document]
gleon99: Maybe terminate the msg with `.` to avoid the weird `". Falling back"` and similar, below?

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,25 +543,56 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
-            ucs_log(log_level, "%s", buf);
-        } else {
-            ucs_log(log_level, "%s, fallback to %s", buf, fallback);
-        }
 
-        return UCS_ERR_NO_MEMORY;
+        ucs_string_buffer_init(&buf);
+
+        err = errno;
+        ucs_string_buffer_appendf(&buf, "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x, type=%s) failed: %s",
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,25 +543,56 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+                                  uct_ib_device_name(&md->super.dev), flags, allocation_type, strerror(err));
+        
+        switch (err) {
+            case ENOMEM:
+                /* The UAR allocation failed due to insufficient PF_LOG_BAR_SIZE, no need to fallback since the result will be the same */
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,25 +543,56 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+                err_log_level = UCS_LOG_LEVEL_ERROR;
+                status = UCS_ERR_NO_MEMORY;
+                break;
+
+            case EOPNOTSUPP:
+                /* The UAR allocation failed due to unsupported allocation type, if there's no fallback, it's an error */
+                if (fallback) {
+                    ucs_string_buffer_appendf(&buf, ". Falling back to %s", fallback);
+                }
+                err_log_level = fallback ? UCS_LOG_LEVEL_WARN : UCS_LOG_LEVEL_ERROR;
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,25 +543,56 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+                                               uint32_t flags, 
+                                               const char *allocation_type, 
+                                               const char *fallback,
+                                               struct mlx5dv_devx_uar **uar_p)
 {
     struct mlx5dv_devx_uar *uar;
-    char buf[512];
+    ucs_status_t status;
+    const char *pf_log_bar_size_suggestion = ". Consider increasing PF_LOG_BAR_SIZE using mlxconfig tool (requires reboot)";
+    ucs_log_level_t err_log_level = UCS_LOG_LEVEL_WARN;
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,25 +543,56 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+                err_log_level = fallback ? UCS_LOG_LEVEL_WARN : UCS_LOG_LEVEL_ERROR;
+                status = UCS_ERR_UNSUPPORTED;
+                break;
+
+            case EINVAL:
+                status = UCS_ERR_INVALID_PARAM;
+                break;
+
+            default:
+                status = UCS_ERR_IO_ERROR;
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,25 +543,56 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+
+        err = errno;
+        ucs_string_buffer_appendf(&buf, "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x, type=%s) failed: %s",
+                                  uct_ib_device_name(&md->super.dev), flags, allocation_type, strerror(err));
+        
+        switch (err) {
+            case ENOMEM:
+                /* The UAR allocation failed due to insufficient PF_LOG_BAR_SIZE, no need to fallback since the result will be the same */
+                ucs_string_buffer_appendf(&buf, "%s", pf_log_bar_size_suggestion);
+                err_log_level = UCS_LOG_LEVEL_ERROR;
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -197,6 +197,8 @@ enum {
     UCT_IB_MLX5_MD_FLAG_MMO_DMA              = UCS_BIT(15),
     /* Device supports XGVMI UMR workflow */
     UCT_IB_MLX5_MD_FLAG_XGVMI_UMR            = UCS_BIT(16),
+    /* Device supports UAR allocation type (1 for WC, 0 for NC_DEDICATED) */
+    UCT_IB_MLX5_MD_FLAG_UAR_ALLOC_TYPE       = UCS_BIT(17),
 
     /* Object to be created by DevX */
     UCT_IB_MLX5_MD_FLAG_DEVX_OBJS_SHIFT  = 17,
```
Artemy-Mellanox, line 204: need increment this value

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -78,10 +78,10 @@
 #  define UCT_IB_MLX5_UAR_ALLOC_TYPE_WC 0x0
 #endif
 
-#if HAVE_DECL_MLX5DV_UAR_ALLOC_TYPE_NC
-#  define UCT_IB_MLX5_UAR_ALLOC_TYPE_NC MLX5DV_UAR_ALLOC_TYPE_NC
+#if HAVE_DECL_MLX5DV_UAR_ALLOC_TYPE_NC_DEDICATED
+#  define UCT_IB_MLX5_UAR_ALLOC_TYPE_NC_DEDICATED MLX5DV_UAR_ALLOC_TYPE_NC_DEDICATED
```
Artemy-Mellanox, line 82: I would keep the name of UCX constant without _DEDICATED

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -602,6 +663,14 @@ ucs_status_t uct_ib_mlx5_devx_uar_init(uct_ib_mlx5_devx_uar_t *uar,
 void uct_ib_mlx5_devx_uar_cleanup(uct_ib_mlx5_devx_uar_t *uar)
 {
 #if HAVE_DEVX
+    if (uar == NULL) {
```
Artemy-Mellanox, line 666: it shouldn't be
uct_ib_mlx5_devx_uar_cleanup() problem, the caller shouldn't call it with NULL

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,109 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+        status = uct_ib_mlx5_devx_alloc_uar(md, UCT_IB_MLX5_UAR_ALLOC_TYPE_NC_DEDICATED, &uar.uar);    
+        if (status == UCS_ERR_UNSUPPORTED) {
+            ucs_error("Both WC and NC_DEDICATED UAR allocation types are not supported");
+            return status;
+        } else if (status != UCS_OK) {
+            return status;
+        } 
+
+        /* NC_DEDICATED is supported - set the flag to 0 */
+        md->flags &= ~UCT_IB_MLX5_MD_FLAG_UAR_ALLOC_TYPE;
```
Artemy-Mellanox, line 616: no need, it is 0 before calling this function

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1885,15 +1885,11 @@ static void uct_ib_mlx5_md_port_counter_set_id_init(uct_ib_mlx5_md_t *md)
 
 static int uct_ib_mlx5_check_uar(uct_ib_mlx5_md_t *md)
```
Artemy-Mellanox, line 1886: why not call uct_ib_mlx5_devx_check_supported_uar_alloc_type() directly? uct_ib_mlx5_check_uar() looks redundant

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,109 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+                /* UAR allocation failed due to unsupported allocation type, if there's no fallback, it's an error */
+                status = UCS_ERR_UNSUPPORTED;
+                break;
+
+            case EINVAL:
+                status = UCS_ERR_INVALID_PARAM;
+                break;
+
+            default:
+                status = UCS_ERR_NO_MEMORY;
```
Artemy-Mellanox, line 584: why default - no memory?

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -197,6 +197,8 @@ enum {
     UCT_IB_MLX5_MD_FLAG_MMO_DMA              = UCS_BIT(15),
     /* Device supports XGVMI UMR workflow */
     UCT_IB_MLX5_MD_FLAG_XGVMI_UMR            = UCS_BIT(16),
+    /* Device supports UAR allocation type (1 for WC, 0 for NC_DEDICATED) */
+    UCT_IB_MLX5_MD_FLAG_UAR_ALLOC_TYPE       = UCS_BIT(17),
```
Artemy-Mellanox, line 201: better more sound name 
UCT_IB_MLX5_MD_FLAG_UAR_USE_WC?

michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: @gleon99 asked to keep the original logic, previously this function returned this status in the general failure context

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
michal-shalev: removed uct_ib_mlx5_check_uar, done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: So do we need this check in the caller?

src/uct/ib/mlx5/ib_mlx5.h
michal-shalev: keep the original HAVE_DECL_MLX5DV_UAR_ALLOC_TYPE_NC? why? also, UCT_IB_MLX5_UAR_ALLOC_TYPE_NC? or UCT_IB_MLX5_UAR_ALLOC_TYPE_NC_DEDICATED?

src/uct/ib/mlx5/ib_mlx5.h
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.h
Artemy-Mellanox: in the context of mlx5dv macros (and HAVE_DECL_ derived from it) it's important to distinguish between nc and nc_dedicated. 
In the UCX context, we need proper NC const so in this case use of dedicated is only implementation detail

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -198,10 +198,10 @@ enum {
     /* Device supports XGVMI UMR workflow */
     UCT_IB_MLX5_MD_FLAG_XGVMI_UMR            = UCS_BIT(16),
     /* Device supports UAR allocation type (1 for WC, 0 for NC_DEDICATED) */
-    UCT_IB_MLX5_MD_FLAG_UAR_ALLOC_TYPE       = UCS_BIT(17),
+    UCT_IB_MLX5_MD_FLAG_UAR_USE_WC       = UCS_BIT(17),
```
yosefe, line 201: align = on column

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2127,7 +2113,7 @@ ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
         goto err_free_context;
     }
 
-    status = uct_ib_mlx5_check_uar(md);
+    status = uct_ib_mlx5_devx_check_supported_uar_alloc_type(md);
```
yosefe, line 2116: let's keep the previous short function name "uct_ib_mlx5_check_uar"

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
 #if HAVE_DEVX
-static ucs_status_t
-uct_ib_mlx5_devx_alloc_uar(uct_ib_mlx5_md_t *md, unsigned flags, int log_level,
-                           char *title, char *fallback,
-                           struct mlx5dv_devx_uar **uar_p)
+static ucs_status_t uct_ib_mlx5_devx_alloc_uar(uct_ib_mlx5_md_t *md, 
+                                               uint32_t flags,
+                                               struct mlx5dv_devx_uar **uar_p) 
 {
+    const char *pf_log_bar_size_suggestion = "Consider increasing PF_LOG_BAR_SIZE using mlxconfig tool (requires reboot)";
```
yosefe, line 550: 1. line too long - pls check code style
2. why need to define a variable for it?

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+    const char *allocation_type            = (flags == UCT_IB_MLX5_UAR_ALLOC_TYPE_WC) ? "WC" : "NC_DEDICATED";
```
yosefe, line 551: 1. line seems too long
2. rename to uar_type_str

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+    ucs_log_level_t err_log_level          = UCS_LOG_LEVEL_DIAG;
     struct mlx5dv_devx_uar *uar;
-    char buf[512];
+    ucs_status_t status;
+    ucs_string_buffer_t buf;
```
yosefe, line 555: 1. rename buf to strb for consistency with other places
2. can init it inline by UCS_STRING_BUFFER_ONSTACK, then no need cleanup

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
-        } else {
-            ucs_log(log_level, "%s, fallback to %s", buf, fallback);
+        ucs_string_buffer_init(&buf);
+
+        err = errno;
+        ucs_string_buffer_appendf(&buf, "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x) type=%s failed: %s. ",
+                                  uct_ib_device_name(&md->super.dev), flags, allocation_type, strerror(err));
+        
+        switch (err) {
+            case ENOMEM:
```
yosefe, line 567: no need to indent 'case' keyword, pls check clang-format output

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
-                "failed: %m", uct_ib_device_name(&md->super.dev), flags, title);
-        if (fallback == NULL) {
-            ucs_log(log_level, "%s", buf);
-        } else {
-            ucs_log(log_level, "%s, fallback to %s", buf, fallback);
+        ucs_string_buffer_init(&buf);
+
+        err = errno;
+        ucs_string_buffer_appendf(&buf, "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x) type=%s failed: %s. ",
+                                  uct_ib_device_name(&md->super.dev), flags, allocation_type, strerror(err));
```
yosefe, line 564: we can print the UAR type as string, we already have it, instead of flags value as hex

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+        
+        switch (err) {
+            case ENOMEM:
+                /* UAR allocation failed due to insufficient PF_LOG_BAR_SIZE; fallback will produce the same result */
+                ucs_string_buffer_appendf(&buf, "%s", pf_log_bar_size_suggestion);
+                err_log_level = UCS_LOG_LEVEL_ERROR;
+                status        = UCS_ERR_NO_MEMORY;
+                break;
+
```
yosefe, line 573: remove space line after 'break' - in multiple places

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+                status = UCS_ERR_INVALID_PARAM;
+                break;
+
+            default:
+                status = UCS_ERR_NO_MEMORY;
+                break;
         }
-
-        return UCS_ERR_NO_MEMORY;
+        ucs_log(err_log_level, "%s", ucs_string_buffer_cstr(&buf));
```
yosefe, line 587: space line before

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+        ucs_string_buffer_cleanup(&buf);
+        return status;
```
yosefe, line 589: i think we can simplify this code without using string buffer and local variable to save errno

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
 
+ucs_status_t uct_ib_mlx5_devx_check_supported_uar_alloc_type(uct_ib_mlx5_md_t *md)
+{
+#if HAVE_DEVX
+    uct_ib_mlx5_devx_uar_t uar;
+    ucs_status_t status;
+
+    status = uct_ib_mlx5_devx_alloc_uar(md, UCT_IB_MLX5_UAR_ALLOC_TYPE_WC, &uar.uar);
+    if (status == UCS_ERR_UNSUPPORTED) {
+        ucs_debug("WC UAR allocation type is not supported, falling back to NC_DEDICATED");
```
yosefe, line 605: 1. print device name in all log messages in this function
2. "falling back to NC_DEDICATED" is misleading because it also may be unsupported. IMO better just print the failure and not the future action
3. uct_ib_mlx5_devx_alloc_uar already prints a lot of log messages and has logic to print the PF_LOG_BAR_SIZE suggestion. I think here we need to use mlx5dv_devx_alloc_uar directly

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,108 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+    } else if (status != UCS_OK) {
+        /* The error is not related to the UAR allocation type, no need to check the fallback */
+        return status;
+    } else {
+        /* WC is supported - set the flag to 1 */
+        md->flags |= UCT_IB_MLX5_MD_FLAG_UAR_USE_WC;
+    }
+
+    uct_ib_mlx5_devx_uar_cleanup(&uar);
+
```
yosefe, line 625: remove space line

src/uct/ib/mlx5/ib_mlx5.c
yosefe: we don't need to check for NULL in this function, a NULL value would be bug in the code and we do not need to have protection against it

gleon99: @yosefe let's keep at least the second one (the internal `uar` check)..

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: 1. done
2. changed to inline

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: mlx5dv_devx_alloc_uar gets device and flags as arguments, additionally there's a string for the type

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: 1. done
2. removed it
3. IMO the entire logic of uct_ib_mlx5_devx_alloc_uar should be used here, because mlx5dv_devx_uar can fail for different reasons, it can reach ENOMEM here as well.

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: removed the first null check

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
michal-shalev: renamed to "uct_ib_mlx5_devx_check_uar"

src/uct/ib/mlx5/ib_mlx5.h
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.h
michal-shalev: renamed "UCT_IB_MLX5_UAR_ALLOC_TYPE_NC_DEDICATED" to "UCT_IB_MLX5_UAR_ALLOC_TYPE_NC"

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: removed the local variable to save errno, I can do something like this to remove the string buffer:
`#if HAVE_DEVX
static ucs_status_t uct_ib_mlx5_devx_alloc_uar(uct_ib_mlx5_md_t *md,
                                               uint32_t flags,
                                               struct mlx5dv_devx_uar **uar_p)
{
    const char *uar_type_str      = (flags == UCT_IB_MLX5_UAR_ALLOC_TYPE_WC) ?
                                    "WC" : "NC_DEDICATED";
    struct mlx5dv_devx_uar *uar;

    uar = mlx5dv_devx_alloc_uar(md->super.dev.ibv_context, flags);
    if (uar == NULL) {
        switch (errno) {
        case ENOMEM:
            ucs_log(UCS_LOG_LEVEL_ERROR,
                    "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x) type=%s failed: %s. "
                    "Consider increasing PF_LOG_BAR_SIZE using mlxconfig tool "
                    "(requires reboot)",
                    uct_ib_device_name(&md->super.dev), flags, uar_type_str,
                    strerror(errno));
            return UCS_ERR_NO_MEMORY;
        case EOPNOTSUPP:
            ucs_log(UCS_LOG_LEVEL_DIAG,
                    "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x) type=%s failed: %s",
                    uct_ib_device_name(&md->super.dev), flags, uar_type_str,
                    strerror(errno));
            return UCS_ERR_UNSUPPORTED;
        case EINVAL:
            ucs_log(UCS_LOG_LEVEL_DIAG,
                    "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x) type=%s failed: %s",
                    uct_ib_device_name(&md->super.dev), flags, uar_type_str,
                    strerror(errno));
            return UCS_ERR_INVALID_PARAM;
        default:
            ucs_log(UCS_LOG_LEVEL_DIAG,
                    "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x) type=%s failed: %s",
                    uct_ib_device_name(&md->super.dev), flags, uar_type_str,
                    strerror(errno));
            return UCS_ERR_NO_MEMORY;
        }
    }

    *uar_p = uar;
    return UCS_OK;
}
#endif`

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -602,6 +659,10 @@ ucs_status_t uct_ib_mlx5_devx_uar_init(uct_ib_mlx5_devx_uar_t *uar,
 void uct_ib_mlx5_devx_uar_cleanup(uct_ib_mlx5_devx_uar_t *uar)
 {
 #if HAVE_DEVX
+    if (uar->uar == NULL) {
+        ucs_error("uar->uar is NULL");
+        return;
+    }
```
yosefe, line 665: pls remove this

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -197,9 +197,11 @@ enum {
     UCT_IB_MLX5_MD_FLAG_MMO_DMA              = UCS_BIT(15),
     /* Device supports XGVMI UMR workflow */
     UCT_IB_MLX5_MD_FLAG_XGVMI_UMR            = UCS_BIT(16),
+    /* Device supports UAR allocation type (1 for WC, 0 for NC_DEDICATED) */
```
Artemy-Mellanox, line 200: with flag name USE_WC no need to explain that 1 is for WC and 0 for NC

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,105 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
-                           char *title, char *fallback,
-                           struct mlx5dv_devx_uar **uar_p)
+static ucs_status_t uct_ib_mlx5_devx_alloc_uar(uct_ib_mlx5_md_t *md,
+                                               uint32_t flags,
+                                               struct mlx5dv_devx_uar **uar_p)
 {
+    const char *uar_type_str      = (flags == UCT_IB_MLX5_UAR_ALLOC_TYPE_WC) ?
+                                    "WC" : "NC_DEDICATED";
+    ucs_log_level_t err_log_level = UCS_LOG_LEVEL_DIAG;
+    UCS_STRING_BUFFER_ONSTACK(strb, 512);
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,105 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
 
     uar = mlx5dv_devx_alloc_uar(md->super.dev.ibv_context, flags);
     if (uar == NULL) {
-        sprintf(buf, "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x(%s)) "
-                "failed: %m", uct_ib_device_name(&md->super.dev), flags, title);
-        if (fallback == NULL) {
-            ucs_log(log_level, "%s", buf);
-        } else {
-            ucs_log(log_level, "%s, fallback to %s", buf, fallback);
+        ucs_string_buffer_appendf(&strb,
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,105 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+                                  "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x)"
+                                  " type=%s failed: %s. ",
+                                  uct_ib_device_name(&md->super.dev),
+                                  flags,
+                                  uar_type_str,
+                                  strerror(errno));
+
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,105 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
+    uct_ib_mlx5_devx_uar_t uar;
+    ucs_status_t status;
+
+    status = uct_ib_mlx5_devx_alloc_uar(md,
+                                        UCT_IB_MLX5_UAR_ALLOC_TYPE_WC,
+                                        &uar.uar);
+    if (status == UCS_ERR_UNSUPPORTED) {
+        status = uct_ib_mlx5_devx_alloc_uar(md, UCT_IB_MLX5_UAR_ALLOC_TYPE_NC, &uar.uar);
+        if (status == UCS_ERR_UNSUPPORTED) {
+            ucs_error("device=%s both WC and NC_DEDICATED UAR allocation types "
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.h
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.c
```diff
@@ -543,48 +543,105 @@ int uct_ib_mlx5_devx_uar_cmp(uct_ib_mlx5_devx_uar_t *uar,
...
-            ucs_log(log_level, "%s", buf);
-        } else {
-            ucs_log(log_level, "%s, fallback to %s", buf, fallback);
+        ucs_string_buffer_appendf(&strb,
+                                  "mlx5dv_devx_alloc_uar(device=%s, flags=0x%x)"
+                                  " type=%s failed: %s. ",
+                                  uct_ib_device_name(&md->super.dev),
+                                  flags,
+                                  uar_type_str,
+                                  strerror(errno));
```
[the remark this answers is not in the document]
michal-shalev: done

src/uct/ib/mlx5/ib_mlx5.h
michal-shalev: @Artemy-Mellanox is it okay now?

src/uct/ib/mlx5/ib_mlx5.c
michal-shalev: removed the second null check as well @yosefe @Artemy-Mellanox

yosefe (approved):

yosefe (approved):

PR #9989, opened 2024-07-02
GTEST: Mock protocol selection tests

iyastreb:
## What
Mock infrastructure and reference test for protocol selection

## Why ?
To enable writing tests for protocol selection, that are executed on different HW

### DIFF
--- a/src/ucs/sys/topo/base/topo.c
+++ b/src/ucs/sys/topo/base/topo.c
@@ -94,6 +94,9 @@ static ucs_topo_global_ctx_t ucs_topo_global_ctx;
 /* Global list of topology detectors */
 UCS_LIST_HEAD(ucs_sys_topo_providers_list);
 
+/* Selected topo provider */
+static ucs_sys_topo_provider_t *ucs_sys_topo_provider = NULL;
+
 
 /* According to NUMA distance definition distances are normalized to 10
  * and the relative distance correlates with the latency.
@@ -104,27 +107,31 @@ static inline double ucs_topo_sysfs_numa_distance_to_latency(double distance)
     return distance * 10e-9;
 }
 
+void ucs_sys_topo_reset_provider()
+{
+    ucs_sys_topo_provider = NULL;
+}
+
 static ucs_sys_topo_provider_t *ucs_sys_topo_get_provider()
 {
-    static ucs_sys_topo_provider_t *provider = NULL;
     ucs_sys_topo_provider_t *list_provider;
     unsigned i;
 
-    if (provider != NULL) {
-        return provider;
+    if (ucs_sys_topo_provider != NULL) {
+        return ucs_sys_topo_provider;
     }
 
     for (i = 0; i < ucs_global_opts.topo_prio.count; ++i) {
         ucs_list_for_each(list_provider, &ucs_sys_topo_providers_list, list) {
             if (!strcmp(ucs_global_opts.topo_prio.names[i],
                         list_provider->name)) {
-                provider = list_provider;
-                return provider;
+                ucs_sys_topo_provider = list_provider;
+                return ucs_sys_topo_provider;
             }
         }
     }
 
-    return provider;
+    return ucs_sys_topo_provider;
 }
 
 static ucs_status_t
--- a/src/ucs/sys/topo/base/topo.h
+++ b/src/ucs/sys/topo/base/topo.h
@@ -73,6 +73,12 @@ extern ucs_list_link_t ucs_sys_topo_providers_list;
 
 
 /**
+ * Reset the internal singleton system topology provider.
+ */
+void ucs_sys_topo_reset_provider(void);
+
+
+/**
  * Find system device by pci bus id.
  *
  * @param [in]  bus_id  pointer to bus id of the device of interest.
--- a/test/gtest/ucp/test_ucp_proto_mock.cc
+++ b/test/gtest/ucp/test_ucp_proto_mock.cc
@@ -1,0 +1,318 @@
+/**
+ * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2024. ALL RIGHTS RESERVED.
+ *
+ * See file LICENSE for terms.
+ */
+
+#include "ucp_test.h"
+
+extern "C" {
+#include <ucp/core/ucp_ep.inl>
+#include <uct/base/uct_iface.h>
+#include <ucp/proto/proto_select.inl>
+}
+
+class mock_iface {
+public:
+    /* Can't use std::function due to coverity errors */
+    using iface_attr_func_t = void (*)(uct_iface_attr&);
+
+    mock_iface()
+    {
+        ucs_assert(m_self == nullptr);
+        m_self = this;
+    }
+
+    ~mock_iface()
+    {
+        cleanup();
+        m_self = nullptr;
+    }
+
+    void cleanup()
+    {
+        m_mock.cleanup();
+    }
+
+    void set_mock_iface_attr(const std::string &tl_name, iface_attr_func_t cb)
+    {
+        mock_tl(tl_name);
+        m_iface_attrs_funcs[tl_name] = cb;
+    }
+
+private:
+    void mock_tl(const std::string &tl_name)
+    {
+        uct_component_h component;
+        ucs_list_for_each(component, &uct_components_list, list) {
+            uct_tl_t *tl;
+            ucs_list_for_each(tl, &component->tl_list, list) {
+                if (tl_name == tl->name) {
+                    EXPECT_EQ(0, m_tls.count(tl_name));
+                    m_tls[tl_name] = tl;
+                    m_mock.setup(&tl->iface_open, iface_open_mock);
+                }
+            }
+        }
+    }
+
+    static ucs_status_t
+    iface_open_mock(uct_md_h md, uct_worker_h worker,
+                    const uct_iface_params_t *params,
+                    const uct_iface_config_t *config, uct_iface_h *iface_p)
+    {
+        uct_tl_t *tl = m_self->m_tls[params->mode.device.tl_name];
+        ucs_status_t status;
+
+        status = m_self->m_mock.orig_func(&tl->iface_open, md, worker, params,
+                                          config, iface_p);
+        if (status != UCS_OK) {
+            return status;
+        }
+
+        uct_base_iface_t *base      = ucs_derived_of(*iface_p, uct_base_iface_t);
+        m_self->m_iface_names[base] = params->mode.device.tl_name;
+        m_self->m_mock.setup(&(*iface_p)->ops.iface_query, iface_query_mock);
+        return status;
+    }
+
+    static ucs_status_t
+    iface_query_mock(uct_iface_h iface, uct_iface_attr_t *iface_attr)
+    {
+        ucs_status_t status = m_self->m_mock.orig_func(
+                                    &iface->ops.iface_query, iface, iface_attr);
+        if (status != UCS_OK) {
+            return status;
+        }
+
+        uct_base_iface_t *base  = ucs_derived_of(iface, uct_base_iface_t);
+        std::string &iface_name = m_self->m_iface_names[base];
+        auto it                 = m_self->m_iface_attrs_funcs.find(iface_name);
+        (it->second)(*iface_attr);
+        return status;
+    }
+
+    /* We have to use singleton to mock C functions */
+    static mock_iface *m_self;
+
+    ucs::mock                                           m_mock;
+    std::unordered_map<std::string, uct_tl_t *>         m_tls;
+    std::unordered_map<uct_base_iface_t *, std::string> m_iface_names;
+    std::unordered_map<std::string, iface_attr_func_t>  m_iface_attrs_funcs;
+};
+
+mock_iface *mock_iface::m_self = nullptr;
+
+struct proto_select_data {
+    size_t      range_start;
+    size_t      range_end;
+    std::string desc;
+};
+
+using proto_select_data_vec_t = std::vector<proto_select_data>;
+
+class test_ucp_proto_mock : public ucp_test, public mock_iface {
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant(variants, UCP_FEATURE_TAG | UCP_FEATURE_AM);
+    }
+
+    virtual void init() override
+    {
+        /* This test is for dynamic protocol selection available only in v2 */
+        if (!is_proto_enabled()) {
+            UCS_TEST_SKIP_R("Proto v2 is disabled");
+        }
+
+        /* Reset topo provider to force reload from config */
+        ucs_sys_topo_reset_provider();
+
+        /*
+         * By default TOPO_PRIO="sysfs,default"
+         * We keep only default config to always have the same topo distances
+         * when test is being executed on different machines.
+         */
+        modify_config("TOPO_PRIO", "default");
+
+        /* Currently only 1 RNDV lane is supported */
+        modify_config("MAX_RNDV_LANES", "1");
+
+        ucp_test::init();
+        connect();
+    }
+
+    virtual void cleanup() override
+    {
+        mock_iface::cleanup();
+        ucp_test::cleanup();
+        /* Reset topo provider to not affect subsequent tests */
+        ucs_sys_topo_reset_provider();
+    }
+
+    static void check_ep_config(entity &e, const proto_select_data_vec_t &data,
+                                ucp_proto_select_key_t key)
+    {
+        ucp_ep_config_t *config = ucp_worker_ep_config(e.worker(), 0);
+        check_proto_select(e.worker(), config->proto_select, data, key);
+    }
+
+    static void check_rkey_config(entity &e,
+                                  const proto_select_data_vec_t &data,
+                                  ucp_proto_select_key_t key)
+    {
+        ucp_rkey_config_t *config = &e.worker()->rkey_config[0];
+        check_proto_select(e.worker(), config->proto_select, data, key);
+    }
+
+    /*
+     * Helper function that returns a key that matches any protocol selection.
+     * It is used to first create the default key matching any protocol, and
+     * then the fine-grained setting can be applied to match specific use case.
+     */
+    static ucp_proto_select_key_t any_key()
+    {
+        ucp_proto_select_key_t key;
+        /* We set all key params to UINT8_MAX, meaning default value, was not
+         * explicitly configured with a fine-grained setting */
+        key.u64 = UINT64_MAX;
+        return key;
+    }
+
+    void connect()
+    {
+        sender().connect(&receiver(), get_ep_params());
+        wait_for_cond([this] {
+            return (receiver().worker()->ep_config.length > 0);
+        }, [this] { progress(); });
+
+        EXPECT_EQ(1, sender().worker()->ep_config.length);
+        EXPECT_EQ(1, sender().worker()->rkey_config_count);
+        EXPECT_EQ(1, receiver().worker()->ep_config.length);
+        EXPECT_EQ(1, receiver().worker()->rkey_config_count);
+    }
+
+protected:
+    static bool
+    select_elem_match(ucp_worker_h worker, const ucp_proto_select_elem_t &elem,
+                      const proto_select_data &data, size_t range_start)
+    {
+        ucp_proto_query_attr_t attr;
+        if (!ucp_proto_select_elem_query(worker, &elem, range_start, &attr)) {
+            return false;
+        }
+
+        return (data.range_end == attr.max_msg_length) &&
+               (data.desc == attr.desc);
+    }
+
+    static void dump_select_elem(ucp_worker_h worker,
+                                 const ucp_proto_select_elem_t &elem)
+    {
+        size_t range_end = -1;
+        size_t range_start;
+        do {
+            range_start = range_end + 1;
+            ucp_proto_query_attr_t attr;
+            if (ucp_proto_select_elem_query(worker, &elem, range_start, &attr)) {
+                UCS_TEST_MESSAGE << range_start <<  "-" << attr.max_msg_length
+                                 << " desc: " << attr.desc << ", config: "
+                                 << attr.config;
+            }
+
+            range_end = attr.max_msg_length;
+        } while (range_end != SIZE_MAX);
+    }
+
+    static void check_proto_select_elem(ucp_worker_h worker,
+                                        const ucp_proto_select_elem_t &elem,
+                                        const proto_select_data_vec_t &data)
+    {
+        for (auto &it : data) {
+            EXPECT_TRUE(select_elem_match(worker, elem, it, it.range_start));
+            /* As we cannot get range_start directly, we assert that protocol
+             * is different at that range */
+            if (it.range_start > 0) {
+                EXPECT_FALSE(select_elem_match(worker, elem, it,
+                                               it.range_start - 1));
+            }
+        }
+    }
+
+    static void check_proto_select(ucp_worker_h worker,
+                                   const ucp_proto_select_t &proto_select,
+                                   const proto_select_data_vec_t &data,
+                                   ucp_proto_select_key_t key)
+    {
+        ucp_proto_select_elem_t select_elem;
+        ucp_proto_select_key_t select_key;
+
+        kh_foreach(proto_select.hash, select_key.u64, select_elem,
+            if (key_match(key, select_key)) {
+                check_proto_select_elem(worker, select_elem, data);
+            });
+
+        if (testing::Test::HasFailure()) {
+            kh_foreach(proto_select.hash, select_key.u64, select_elem,
+                if (key_match(key, select_key)) {
+                    UCS_TEST_MESSAGE << "Key op flags: "
+                                     << (int)select_key.param.op_id_flags
+                                     << ", attr: "
+                                     << (int)select_key.param.op_attr;
+                    dump_select_elem(worker, select_elem);
+                });
+        }
+    }
+
+    static bool
+    key_match(ucp_proto_select_key_t req, ucp_proto_select_key_t actual)
+    {
+        /*
+         * For each field of the protocol selection key check whether it has a
+         * default value (UINT8_MAX - meaning a wildcard), or the fine-grained
+         * setting matches was provided and matches the actual value.
+         */
+#define CMP_FIELD(FIELD) \
+        if ((UINT8_MAX != req.param.FIELD) && \
+            (req.param.FIELD != actual.param.FIELD)) { \
+            return false; \
+        }
+
+        CMP_FIELD(op_id_flags);
+        CMP_FIELD(op_attr);
+        CMP_FIELD(dt_class);
+        CMP_FIELD(mem_type);
+        return true;
+    }
+};
+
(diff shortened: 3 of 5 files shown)


test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+            ucp_proto_query_attr_t proto_attr;
+            int valid = ucp_proto_select_elem_query(worker, &select_elem,
+                                                    range_start, &proto_attr);
+            range_end = proto_attr.max_msg_length;
+            if (valid) {
+            char buf[1024];
+                sprintf(buf, "  %zu-%zu desc: %s, config: %s", range_start,
+                range_end, proto_attr.desc, proto_attr.config);
+                UCS_TEST_MESSAGE << buf;
+            }
```
ivankochin, line 363:
```suggestion
            if (ucp_proto_select_elem_query(worker, &select_elem,
                                            range_start, &proto_attr) {
                UCS_TEST_MESSAGE << range_start <<  "-" << proto_attr.max_msg_length << " dest: " << proto_attr.desc << ", config: " << proto_attr.config;
            }
            range_end = proto_attr.max_msg_length;
```

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+    static void
+    check_proto_select_elem(ucp_worker_h worker,
+                            const ucp_proto_select_elem_t &select_elem,
+                            const proto_select_data_vec_t &data)
+    {
+        dump_select_elem(worker, select_elem);
+        for (auto &it : data) {
+            size_t start = string_to_memunits(it.range_start);
+            EXPECT_TRUE(select_elem_match(worker, select_elem, it, start));
+            /* As we cannot get range_start directly, we assert that protocol
```
ivankochin, line 376: Why we cannot get range_start directly?

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+    static ucs_status_t iface_open_mock(uct_md_h md, uct_worker_h worker,
+                                        const uct_iface_params_t *params,
+                                        const uct_iface_config_t *config,
+                                        uct_iface_h *iface_p)
+    {
+        mock_component *self = find(md->component);
+        uct_tl_t *tl         = self->find_tl(params);
+        ucs_status_t status  = mock_t::actual_call(&tl->iface_open, md, worker,
+                                                   params, config, iface_p);
+        if (status == UCS_OK) {
```
gleon99, line 191: Minor: invert branches in all these places. Better to have:
```
if (status != UCS_OK) {
    return status;
}

...
...
return UCS_OK;
```

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+#include <ucp/proto/proto_select.inl>
+#include <ucs/sys/topo/base/topo.h>
+}
+
+struct mock_t {
+public:
+    typedef void (*func_t)(void);
+    typedef std::map<func_t *, std::unique_ptr<mock_t>> mock_map_t;
+
+    ~mock_t()
```
gleon99, line 23: Can move to `private`, under the ctor?

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+    static void
+    check_ep_config(entity &e, const proto_select_data_vec_t &data,
+                    ucp_proto_select_key_t key)
+    {
+        ucp_ep_config_t *config = ucp_worker_ep_config(e.worker(), 0);
+        check_proto_select(e.worker(), config->proto_select, data, key);
+    }
+
+    static void
+    check_rkey_config(entity &e, const proto_select_data_vec_t &data,
```
ivankochin, line 292: Never called?

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+        }
+
+        return status;
+    }
+
+    uct_component_h                           m_component;
+    uint64_t                                  m_flags_orig;
+    std::map<uct_base_iface_t *, std::string> m_iface_names;
+    std::map<std::string, iface_attr_func_t>  m_iface_attrs_funcs;
+    std::map<std::string, perf_attr_func_t>   m_perf_attrs_funcs;
```
ivankochin, line 243: Is there a necessity to be able to set different values for different transport devices? WDYTA removing all this maps and replacing them by one instance of `iface_attr_func_t` and `perf_attr_func_t` which would be used to mock all the corresponding calls?

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+extern "C" {
+#include <ucp/core/ucp_ep.inl>
+#include <uct/base/uct_component.h>
+#include <uct/base/uct_iface.h>
+#include <uct/base/uct_md.h>
+#include <ucp/proto/proto_select.inl>
+#include <ucs/sys/topo/base/topo.h>
+}
+
+struct mock_t {
```
ivankochin, line 18: If I understand correctly both `mock_t` and `mock_component` entities are used for two purposes where first is storage of mock and origin pointer or component and operations under them and another are their static interfaces which are supposed to store some global containers with `mock_t` or `mock_component` and making some operations with them.

So I want to propose to split these to API to two different entities: make `mock_t` and `mock_component` APIs as minimalistic as possible and move all the containers and operations with outside of the classes (maybe create separate workspace for them). To incapsulate map containers we can use unnamed namespaces.

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+        g_mock_map.clear();
+    }
+
+    template <typename Fn>
+    static mock_t *setup(Fn *orig_ptr, Fn mock)
+    {
+        func_t *ptr = reinterpret_cast<func_t *>(orig_ptr);
+        auto it = g_mock_map.find(ptr);
+        if (it == g_mock_map.end()) {
+            mock_t * new_mock = new mock_t(ptr, reinterpret_cast<func_t>(mock));
```
gleon99, line 39: There's `make_unique` since c++14. Currently we use c++11.
Let's see if [this one](https://github.com/openucx/ucx/pull/10018) gets merged soon. If yes - can consider using `make_unique` here then.

gleon99: Ignore.

test/gtest/ucp/test_ucp_proto_select.cc
iyastreb: Yes, it's never called so far.
To me the goal of this commit is to provide meaningful infra for writing mock tests for protocol selection. Currently I have just one test that is just an example. I think API for checking rkey config would be useful in the future

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+        g_component_map.clear();
+        mock_t::cleanup();
+    }
+
+    uct_component_t * operator->() const
+    {
+        return m_component;
+    }
+
+    void set_mock_perf_attr(const std::string &iface_name, perf_attr_func_t cb)
```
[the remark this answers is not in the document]
iyastreb: Yes, so far it's never used.
I thought of this function as an API for extended protocol selection tests. But if we stick to mocking `iface_attr_t`, then probably this one we can remove.

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+    {
+        func_t *ptr = reinterpret_cast<func_t *>(orig_ptr);
+        auto it = g_mock_map.find(ptr);
+        if (it == g_mock_map.end()) {
+            mock_t * new_mock = new mock_t(ptr, reinterpret_cast<func_t>(mock));
+            g_mock_map[ptr] = std::unique_ptr<mock_t>(new_mock);
+            return new_mock;
+        }
+
+        return it->second.get();
```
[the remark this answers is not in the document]
iyastreb: Ok, there are 2 points here:

1. Why do we mock the same functions multiple times?
This happens unintentionally, just because UCT creates multiple interfaces of the same transport type. Since we mock iface open with `iface_open_mock`, we need to replace iface_query with mock. But if we create more than one iface of the same type - this situation happens. It is not an error in this sense.

2. Why don't we overwrite?
Yes, we can overwrite, it looks logical

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+
+    template <typename Fn, typename... Args>
+    static ucs_status_t actual_call(Fn *orig_ptr, Args&&... args)
+    {
+        mock_t *mock = find(reinterpret_cast<func_t *>(orig_ptr)); 
+        Fn orig_func = mock ? reinterpret_cast<Fn>(mock->m_orig) : *orig_ptr;
+        return orig_func(std::forward<Args>(args)...);
+    }
+
+    static mock_t *find(func_t *orig_ptr)
```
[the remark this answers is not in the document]
iyastreb: ok, why not

test/gtest/ucp/test_ucp_proto_select.cc
iyastreb: ok

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+        return status;
+    }
+
+    uct_component_h                           m_component;
+    uint64_t                                  m_flags_orig;
+    std::map<uct_base_iface_t *, std::string> m_iface_names;
+    std::map<std::string, iface_attr_func_t>  m_iface_attrs_funcs;
+    std::map<std::string, perf_attr_func_t>   m_perf_attrs_funcs;
+
+    typedef std::map<uct_component_h,
```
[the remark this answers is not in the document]
iyastreb: done

test/gtest/ucp/test_ucp_proto_select.cc
iyastreb: Because this field does not exist neither in `ucp_proto_query_attr_t` nor in `ucp_proto_threshold_elem_t`. We kind of keep only the upper limit, and lower limit we deduce from the array: e.g. check `ucp_proto_select_elem_info`, `ucp_proto_select_write_info`

test/gtest/ucp/test_ucp_proto_select.cc
iyastreb: ok, applied with some minor changes

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+            }
+        } while (range_end != SIZE_MAX);
+    }
+
+    static void
+    check_proto_select_elem(ucp_worker_h worker,
+                            const ucp_proto_select_elem_t &select_elem,
+                            const proto_select_data_vec_t &data)
+    {
+        dump_select_elem(worker, select_elem);
```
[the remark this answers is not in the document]
iyastreb: Yes, we need something like that, will be fixed
This dumping is very useful when writing new tests. Maybe dumping on error is sufficient for this purpose too

test/gtest/ucp/test_ucp_proto_select.cc
iyastreb: It's hard to say whether it's really needed, it depends what kind of tests we wish to create.
This gives us possibility to mock multiple transport/ifaces within the same test, which I think might be useful to:
1. Test proto selection between multiple ifaces (e.g. when we switch from one iface to another)
2. Test lane selection
I thought it's important to have this possibility (especially if it does not come with huge complexity). Only if we are 100% sure that those tests will never be needed - then we can simplify and stick to a single instance.

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,447 @@
...
+    ~mock_component()
+    {
+        m_component->flags = m_flags_orig;
+    }
+
+    static mock_component &setup(const std::string &name)
+    {
+        uct_component_h component = find_uct_component(name);
+        if (nullptr == component) {
+            UCS_TEST_SKIP_R("Component " + name + " not found");
```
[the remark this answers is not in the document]
iyastreb: Sure, if we consider simplest use case with single transport - this skip makes no sense.
I was thinking of cases with multiple transports specified, like implicit "all" or specific "self,shmem,tcp,ib".
In this case - is it guaranteed that ALL those transports exist?
I was not sure, therefore this relaxed skip.

test/gtest/ucp/test_ucp_proto_select.cc
iyastreb: I refactored the code taking into account your feedback (also from our discussion). Key changes:
1. Split mock_t into outer mock { and inner struct mock_func }
2. Removed static global map - in fact we can instantiate mock per test, no need in global instance
3. Moved mock into test_helpers
4. Split mock_component into 2 entities:
- mock_component_test - This is the base class for component mocking, containing mock instance + implements global singleton map(), and setup_mock functions
- mock_component - this is API and infra for individual component mocking

Hope that helps

test/gtest/ucp/test_ucp_proto_select.cc
ivankochin: As I can see, `select_elem_match` for `start - 1` would always return `false` since we compare `data.range_end` for current proto with `proto_attr.max_msg_length` which obviously would be different at `start - 1` point. So even if there are the same `desc` and `config` it would return false. AFAIU it would never happen now since `select_elem::thresholds` elements are merged together if same proto is selected, but I am not sure whether it is a good idea to rely on that.

test/gtest/ucp/test_ucp_proto_select.cc
```diff
@@ -0,0 +1,386 @@
...
+mock_component &mock_component_test::setup_mock(const std::string &name)
+{
+    uct_component_h component = find_uct_component(name);
+    if (nullptr == component) {
+        UCS_TEST_SKIP_R("Component " + name + " not found");
+    }
+
+    mock_component *mock_comp = find(component);
+    if (nullptr == mock_comp) {
+        mock_comp = new mock_component(component, m_mock);
```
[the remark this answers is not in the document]
iyastreb: > Why can't each mock_component contain its own mock instance?

That would be a good idea, if we are 100% sure that component mocks are fully disjoint. But we cannot guarantee that, because different components may point to the same `uct_tl_t` instances, and therefore we may end mocking the same pointer multiple times.
What's wrong with mocking the same pointer twice?
It's a problem to revert mock changes to the original state, for example:
```C++
orig_fun_ptr = &orig_fun;
// remember CURRENT value (&orig_fun), restores on destroy: orig_fun_ptr = &orig_fun;
mock_func(&orig_fun_ptr, mock_fun);
// second mock, same ptr
// remember CURRENT value (&mock_fun), restores on destroy: orig_fun_ptr = &mock_fun;
mock_func(&orig_fun_ptr, mock_fun);
// which one is destroyed first?
```
Now you have a situation when behaviour depends on destructor order. We don't want that. We must use single mock instance for all UCT components

> Or static instance?

Yes, static would work.
However to me it seems that `m_mock` field in fixture is superior alternative, because it's automatically cleaned up upon test case completion. And also I try to avoid global variables when possible, IMO it's better to have explicit dependency injection here.

brminich:
failure in  `rcx/test_ucp_proto_mock.mock_iface_attr/0` looks relevant

iyastreb:
> failure in `rcx/test_ucp_proto_mock.mock_iface_attr/0` looks relevant

Yes, this PR is open for a long time, and new changes in master break test assumptions.
Fixed

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -0,0 +1,408 @@
...
+                    dump_select_elem(worker, select_elem);
+                });
+        }
+    }
+
+    static bool
+    key_match(ucp_proto_select_key_t req, ucp_proto_select_key_t actual)
+    {
+#define CMP_FIELD(FIELD) \
+        if ((UINT8_MAX != req.param.FIELD) && \
```
yosefe, line 364: why comparing to UINT8_MAX?

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -0,0 +1,408 @@
...
+            iface_attr.cap.am.max_short = 208;
+            iface_attr.bandwidth.shared = 10000000000;
+            iface_attr.latency.c        = 0.000006;
+            iface_attr.latency.m        = 0.000000001;
+        });
+
+    ucp_test::init();
+    if (!has_resource(sender(), "rc_mlx5", "mlx5_0:1")) {
+        UCS_TEST_SKIP_R("no mlx5_0:1 device");
+    }
```
yosefe, line 391: can we unite it with mock setup?

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -0,0 +1,408 @@
...
+        CMP_FIELD(op_attr);
+        CMP_FIELD(dt_class);
+        CMP_FIELD(mem_type);
+        return true;
+    }
+};
+
+UCS_TEST_P(test_ucp_proto_mock, mock_iface_attr, "NET_DEVICES=mlx5_0:1,lo")
+{
+    setup_mock("ib").set_mock_iface_attr("rc_mlx5/mlx5_0:1",
```
yosefe, line 379: can we implement mock so that passing component name is not required?

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -0,0 +1,408 @@
...
+public:
+    static void get_test_variants(std::vector<ucp_test_variant> &variants)
+    {
+        add_variant(variants, UCP_FEATURE_TAG | UCP_FEATURE_AM);
+    }
+
+    virtual void init() override
+    {
+        modify_config("PROTO_ENABLE", "y");
+        modify_config("MAX_RNDV_LANES", "1");
```
yosefe, line 223: why are these needed?

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -0,0 +1,408 @@
...
+    {
+        ucp_proto_select_key_t key;
+        key.u64 = UINT64_MAX;
+        return key;
+    }
+
+    void connect()
+    {
+        sender().connect(&receiver(), get_ep_params());
+        for (int i = 0; i < 10; ++i) {
```
yosefe, line 263: why needed and why 10 is enough?

iyastreb: Right, it's ugly, better wait until certain condition happens, will reimplement this

src/ucs/sys/topo/base/topo.h
```diff
@@ -72,6 +72,10 @@ typedef ucs_status_t
 extern ucs_list_link_t ucs_sys_topo_providers_list;
 
 
+/* Selected topo provider */
+extern struct ucs_sys_topo_provider *ucs_sys_topo_provider;
```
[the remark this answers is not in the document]
iyastreb: Ok, I think it's better to have a reset API rather then extending `ucs_sys_topo_get_provider` with reset flag, because the latter would require one extra check in the production flow

test/gtest/ucp/test_ucp_proto_mock.cc
iyastreb: No, unfortunately we can't..
Because we must setup mock before ucp_init, in order to intercept all the iface opening and perf estimation functions. So we cannot check sender resources at that point.
But I can implement this skip slightly differently, so that we just call a single function here (`check_mocks`) that makes sure ALL the configured mocks got initialized and skip test otherwise, without explicit call to `has_resource`

test/gtest/ucp/test_ucp_proto_mock.cc
iyastreb: done

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -8,130 +8,80 @@
...
-        uct_tl_t *tl;
-        ucs_list_for_each(tl, &m_component->tl_list, list) {
-            if (0 == strcmp(tl->name, params->mode.device.tl_name)) {
-                return tl;
+        uct_component_h component;
+        ucs_list_for_each(component, &uct_components_list, list) {
+            uct_tl_t *tl;
+            ucs_list_for_each(tl, &component->tl_list, list) {
+                if (tl_name == tl->name) {
+                    EXPECT_EQ(0, m_tls.count(tl_name));
```
yosefe, line 60: why needed?

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -178,39 +128,27 @@ class mock_component {
         return status;
     }
 
-    uct_component_h                                     m_component;
-    ucs::mock                                          &m_mock;
+    /* We have to use singleton to mock C functions */
+    static mock_iface *singleton;
```
yosefe, line 132: maybe m_signleton or m_instance?

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -219,20 +157,22 @@ class test_ucp_proto_mock : public ucp_test, public mock_component_test {
 
     virtual void init() override
     {
+        /* Reset topo provider to force reload from config */
+        ucs_sys_topo_reset_provider();
+        modify_config("TOPO_PRIO", "default");
```
yosefe, line 162: can u document each env var why is it needed?

test/gtest/ucp/test_ucp_proto_mock.cc
yosefe: ok, can you pls document this in any_key and key_match functions?

iyastreb: sure

test/gtest/ucp/test_ucp_proto_mock.cc
iyastreb: sure, done

test/gtest/ucp/test_ucp_proto_mock.cc
iyastreb: maybe shorter `m_self`?

test/gtest/ucp/test_ucp_proto_mock.cc
iyastreb: Currently mock works with a single transport, so this is just a check that we don't try to mock multiple ones. In the new impl it's slightly different

iyastreb:
> @iyastreb can you pls revert the refactoring and only fix review comments and refactor in next pr?

Ok, done

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -159,8 +124,18 @@ class test_ucp_proto_mock : public ucp_test, public mock_iface {
...
         ucs_sys_topo_reset_provider();
+
+        /*
+         * By default TOPO_PRIO="sysfs,default"
+         * We keep only default config to always have the same topo distances
+         * when test is being executed on different machines.
+         */
         modify_config("TOPO_PRIO", "default");
+
+        /* This test is for dynamic protocol selection available only in v2 */
```
yosefe, line 135: maybe we skill the test altogether if protov2 is disabled? since the default is that it's enabled

iyastreb: good idea, done

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -0,0 +1,343 @@
...
+        unsigned max_dev_path   = iface_attr->dev_num_paths;
+        (it->second)(*iface_attr);
+
+        /*
+         * It's hard to mock more paths than are actually supported by the
+         * hardware, so in this case we want to skip the test. Returning an
+         * error or skipping the test directly from this point is problematic,
+         * therefore we remember the error and check it in skip_on_error() later
+         */
+        if (iface_attr->dev_num_paths > max_dev_path) {
```
brminich, line 107: why can't we mock more paths? What is the issue?

test/gtest/ucp/test_ucp_proto_mock.cc
```diff
@@ -0,0 +1,408 @@
...
+        key.u64 = UINT64_MAX;
+        return key;
+    }
+
+    void connect()
+    {
+        sender().connect(&receiver(), get_ep_params());
+        for (int i = 0; i < 10; ++i) {
+            progress();
+        }
```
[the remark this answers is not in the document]
iyastreb: I think this is stale comment, this is fixed already

src/ucs/sys/topo/base/topo.c
```diff
@@ -54,7 +54,7 @@ typedef struct {
 /*
  * Structure needed to define a topology module implementation
  */
-typedef struct {
+typedef struct ucs_sys_topo_provider {
```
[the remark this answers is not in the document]
iyastreb: outdated

test/gtest/ucp/test_ucp_proto_mock.cc
iyastreb: There are 2 difficulties:

1. I noticed that dev_num_paths are implicitly "aligned" between IB transports (e.g. between `rc_mlx5` and `ud_mlx5`), in function `ucp_address_gather_devices`. There are 2 possible workarounds for this problem: either we mock both transports (which comes with just another problem: mocking of different TLs is hard to do); or we mock MD query to return some mock MD name for rc_mlx5.
I tried both options, I like more the second one, and by using it we can create as many dev paths as we want

2. wireup tries to connect all the lanes for all the paths, and it fails to do it with mocked paths (Destination is unreachable). It works well when real device supports that amount of paths. But when it exceeds it, then new paths are successfully created (e.g. mock_1:0/path2), but then wireup fails -> connection is terminating -> it hangs forever. Maybe it deserves a bit more investigation, I think we can find solution in the future. For now I just explicitly disable this

yosefe: can we force num paths to 1?

iyastreb: We can force it in this PR if you want. But this is a serious limitation, which makes mocking useless for the task I'm working on. Having multiple paths is crucial to reproduce existing bugs in protocol selection:
- issue with lane sorting by BW, which happens only when device with multiple paths is present
- issue with incorrect performance estimation when multiple paths get assigned the entire device performance

brminich: maybe you can set `IB_NUM_PATHS` to the value you need in the particular test?

iyastreb: Thank you so much for this proposal, it works like a charm!
This is exactly what I needed to mock multiple paths. So I remove all these checks

brminich (approved):

PR #9845, opened 2024-04-26
UCT/GGA: basic iface and ep infrastructure

evgeny-leksikov:
## What
Implement and test basic GGA iface and ep infrastructure

## Why ?
Continue https://github.com/openucx/ucx/pull/9811

PR #9993, opened 2024-07-05
UCT/GGA: put_zcopy and get_zcopy implementation

evgeny-leksikov:
## What
Implement and test GGA put_zcopy and get_zcopy

## Why ?
continue https://github.com/openucx/ucx/pull/9845


src/uct/ib/rc/base/rc_iface.c
```diff
@@ -1055,6 +1047,24 @@ void uct_rc_iface_vfs_refresh(uct_iface_h iface)
     }
 }
 
+void
+uct_rc_iface_adjust_max_get_zcopy(uct_rc_iface_t *iface,
```
iyastreb, line 1051: Would it be easier and more explicit if this function just returns a value instead of internal assignment?
```C
size_t uct_rc_iface_get_max_get_zcopy(...)

iface->config.max_get_zcopy = uct_rc_iface_get_max_get_zcopy(...);
```

test/gtest/uct/uct_test.h
```diff
@@ -117,6 +117,11 @@ class uct_test : public testing::TestWithParam<const resource*>,
     uct_test();
     virtual ~uct_test();
 
+    virtual bool has_transport(const std::string& tl_name) const;
+
+    /* Call this wrapper right before uct_config_modify */
+    virtual std::string wrap_config_str(const std::string& str) const;
```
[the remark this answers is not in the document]
evgeny-leksikov: IMO this approach adds complexity without significant benefit. Once you need to add a new wrapper (1 simple thing), you need to control 2 functions (2 hops in the code to understand logic) instead of 1 which is simple enough.

src/uct/ib/mlx5/ib_mlx5_log.c
```diff
@@ -237,6 +237,31 @@ static unsigned uct_ib_mlx5_parse_dseg(void **dseg_p, void *qstart, void *qend,
...
+static size_t
+uct_ib_mlx5_dump_data_seg(const struct mlx5_wqe_data_seg *seg,
+                          const char *prefix, char *buf, size_t max)
+{
+    snprintf(buf, max, " %s[va 0x%"PRIx64" len %d lkey 0x%x]", prefix,
+             be64toh(seg->addr), be32toh(seg->byte_count), be32toh(seg->lkey));
+    return strlen(buf);
+}
+
+static void uct_ib_mlx5_dump_dma_wqe(const void *wqe, char *buf, size_t max)
```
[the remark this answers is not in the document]
evgeny-leksikov: extra macros disimprove readability, but thank for pointing to redundant `strlen()`, it can be replaced by just `return snprintf(...);`

src/uct/ib/rc/base/rc_iface.c
evgeny-leksikov: then need to duplicate `warn(...)` since assignment becomes outside this function

iyastreb: I think `warn` can still be inside the function:
```C
{
    if (config->tx.max_get_zcopy == UCS_MEMUNITS_AUTO) {
        return max_tl_get_zcopy;
    } else if (config->tx.max_get_zcopy <= max_tl_get_zcopy) {
        return config->tx.max_get_zcopy;
    } else {
        ucs_warn("%s_iface on %s:%d: reduced max_get_zcopy to %zu",
                 tl_name, dev_name, iface->super.config.port_num,
                 max_tl_get_zcopy);
        return max_tl_get_zcopy;
    }
}
```

src/uct/ib/mlx5/ib_mlx5_log.c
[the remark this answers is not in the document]
iyastreb: `snprintf` in theory may return negative number, but I think it's mostly safe here.
Still I think the easiest is to keep just one call, but up to you

src/uct/ib/rc/base/rc_iface.c
evgeny-leksikov: it's too early to warn since the value is not changed in this scope, we cannot predict caller's behavior

src/uct/ib/mlx5/ib_mlx5_log.c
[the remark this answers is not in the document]
evgeny-leksikov: ok

src/uct/ib/mlx5/ib_mlx5_log.c
```diff
@@ -16,6 +16,12 @@
 #include <string.h>
 
 
+#define UCT_IB_MLX5_DMA_SEG_FMT           "va 0x%"PRIx64" len %d lkey 0x%x"
```
yosefe, line 19: len can be %u

src/uct/ib/mlx5/rc/rc_mlx5.inl
```diff
@@ -874,15 +881,46 @@ void uct_rc_mlx5_txqp_dptr_post_iov(uct_rc_mlx5_iface_common_t *iface, int qp_ty
...
+    case MLX5_OPCODE_MMO|UCT_RC_MLX5_OPCODE_FLAG_MMO_PUT:
+    case MLX5_OPCODE_MMO|UCT_RC_MLX5_OPCODE_FLAG_MMO_GET:
+        gga_wqe                  = txwq->curr;
+#if ENABLE_DEBUG_DATA
+        gga_wqe->padding         = 0xdeadbeef;
+#endif
+        gga_wqe->be_opaque_lkey  = htobe32(opaque_mr->lkey);
+        gga_wqe->be_opaque_vaddr = htobe64((uintptr_t)opaque_mr->addr);
+
+        ucs_assert(iovcnt == 1);
```
yosefe, line 897: assertv

src/uct/ib/mlx5/rc/rc_mlx5.inl
```diff
@@ -874,15 +881,46 @@ void uct_rc_mlx5_txqp_dptr_post_iov(uct_rc_mlx5_iface_common_t *iface, int qp_ty
...
 
+#if HAVE_MLX5_MMO
+    case MLX5_OPCODE_MMO|UCT_RC_MLX5_OPCODE_FLAG_MMO_PUT:
+    case MLX5_OPCODE_MMO|UCT_RC_MLX5_OPCODE_FLAG_MMO_GET:
+        gga_wqe                  = txwq->curr;
+#if ENABLE_DEBUG_DATA
+        gga_wqe->padding         = 0xdeadbeef;
+#endif
+        gga_wqe->be_opaque_lkey  = htobe32(opaque_mr->lkey);
+        gga_wqe->be_opaque_vaddr = htobe64((uintptr_t)opaque_mr->addr);
```
yosefe, line 895: minor: maybe we can precompute the byte order switch during mem reg or rkey unpack?

src/uct/ib/mlx5/rc/rc_mlx5.inl
```diff
@@ -874,15 +881,46 @@ void uct_rc_mlx5_txqp_dptr_post_iov(uct_rc_mlx5_iface_common_t *iface, int qp_ty
...
+
+        ucs_assert(iovcnt == 1);
+        if (opcode_flags & UCT_RC_MLX5_OPCODE_FLAG_MMO_PUT) {
```
yosefe, line 898: maybe we can simplify the code by defining pointer like "struct mlx5_wqe_data_seg *src,*dest" and point them to gather/scatter fields according to flags. then fill src and dest with uct_ib_mlx5_set_data_seg

test/gtest/uct/ib/test_dc.cc
```diff
@@ -531,11 +540,15 @@ class test_dc_flow_control : public test_rc_flow_control {
...
     }
 
-    virtual void test_pending_grant(int16_t wnd)
+    virtual void test_pending_grant(int16_t wnd) override
     {
         test_rc_flow_control::test_pending_grant(wnd);
         flush();
     }
+
+    virtual std::string wrap_config_str(const std::string &str) const override {
```
yosefe, line 549: why need wrap_config_str? it seems hacky.. i'd expect a config without transport name to work..

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -275,9 +301,13 @@ static UCS_CLASS_INIT_FUNC(uct_gga_mlx5_ep_t, const uct_ep_params_t *params)
 
     UCS_CLASS_CALL_SUPER_INIT(uct_rc_mlx5_base_ep_t, params);
 
-    self->dma_opaque.mr = ibv_reg_mr(md->super.pd, self->dma_opaque.buf,
-                                     UCT_GGA_MLX5_OPAQUE_BUF_LEN,
-                                     IBV_ACCESS_LOCAL_WRITE);
+    self->dma_opaque.mr =
+            ibv_reg_mr(md->super.pd,
+                       ucs_align_up_pow2_ptr(&self->dma_opaque.buf[0],
+                                             UCS_SYS_CACHE_LINE_SIZE),
```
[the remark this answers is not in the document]
evgeny-leksikov: yes, reg_mr < bufer size, it's needed for alignment

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -388,6 +418,102 @@ uct_gga_mlx5_ep_connect_to_ep_v2(uct_ep_h tl_ep,
...
+    UCT_CHECK_LENGTH(uct_iov_total_length(iov, iovcnt), 0, MAX_GGA_MSG_SIZE,
+                     "put_zcopy");
+
+    /* rkey resolution doesn't depend on available resources */
+    status = uct_gga_mlx5_rkey_resolve(md, rkey_handle);
+    if (ucs_unlikely(status != UCS_OK)) {
+        return status;
+    }
+
+    UCT_RC_CHECK_RES(&iface->super, &ep->super);
```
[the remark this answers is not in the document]
evgeny-leksikov: why? we can resolve the key while waiting for resources

test/gtest/uct/ib/test_dc.cc
evgeny-leksikov: let's see if it might be fixed in config (https://github.com/openucx/ucx/pull/10057)

evgeny-leksikov:
the PR has dependency to https://github.com/openucx/ucx/pull/10057

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -36,8 +36,9 @@ typedef struct {
 } uct_gga_mlx5_iface_config_t;
 
 typedef struct {
-    uint8_t       buf[UCT_GGA_MLX5_OPAQUE_BUF_LEN];
-    struct ibv_mr *mr;
+    uint8_t             buf[UCT_GGA_MLX5_OPAQUE_BUF_LEN];
+    uct_dma_opaque_mr_t opaque_mr;
```
yosefe, line 40: uct_ib_mlx5_dma_opaque_mr_t

src/uct/ib/mlx5/gga/gga_mlx5.c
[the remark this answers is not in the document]
yosefe: won't it it better to allocate opaque_buf separately using posix_memalign for example?

src/uct/ib/mlx5/gga/gga_mlx5.c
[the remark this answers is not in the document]
yosefe: to avoid the extra overhead of resolving the key every time we go to pending queue

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -511,7 +637,15 @@ static UCS_CLASS_INIT_FUNC(uct_gga_mlx5_iface_t,
...
-    config->super.super.fc.enable = 0; /* FC requires AM capability */
+    config->super.super.fc.enable        = 0; /* FC requires AM capability */
+    self->super.config.atomic_fence_flag = UCT_IB_MLX5_WQE_CTRL_FLAG_FENCE;
+    self->super.super.config.fence_mode  = UCT_RC_FENCE_MODE_AUTO;
+
+    uct_rc_iface_adjust_max_get_zcopy(&self->super.super, &config->super.super,
+            MAX_GGA_MSG_SIZE,
+            uct_ib_device_name(uct_ib_iface_device(&self->super.super.super)),
+            "gga");
+
```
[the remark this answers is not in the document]
yosefe: seems unresolved, do we really need to call uct_rc_iface_adjust_max_get_zcopy here?

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -236,6 +238,16 @@ enum {
...
 
 #if HAVE_DEVX
+typedef struct uct_ib_mlx5_dma_wqe {
+    struct mlx5_wqe_ctrl_seg ctrl;
+    uint32_t                 padding;   /* unused for dma */
+    uint32_t                 be_opaque_lkey;
+    uint64_t                 be_opaque_vaddr;
+    struct mlx5_wqe_data_seg gather;
+    struct mlx5_wqe_data_seg scatter;
+} UCS_S_PACKED UCS_V_ALIGNED(sizeof(void*)) uct_ib_mlx5_dma_wqe_t;
```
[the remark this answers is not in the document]
yosefe: BTW, why do we meed tp declare it's aligned? we don't do this for other WQE types

test/gtest/uct/uct_test.cc
```diff
@@ -1625,16 +1625,32 @@ void test_uct_iface_attrs::basic_iov_test()
 
     EXPECT_FALSE(max_iov_map.empty());
 
-    if (max_iov_map.find("am")  != max_iov_map.end()) {
+    if (m_e->iface_attr().cap.flags & (UCT_IFACE_FLAG_AM_SHORT |
+                                       UCT_IFACE_FLAG_AM_BCOPY |
+                                       UCT_IFACE_FLAG_AM_ZCOPY)) {
+        EXPECT_NE(max_iov_map.end(), max_iov_map.find("am"));
```
iyastreb, line 1631: alternatively you can check the presence:
`EXPECT_EQ(map.count("key"), 1)`

test/gtest/uct/uct_test.h
```diff
@@ -478,7 +479,11 @@ class test_uct_iface_attrs : public uct_test {
     rc_verbs,           \
     dc_mlx5,            \
     ud_verbs,           \
-    ud_mlx5,            \
+    ud_mlx5
+
+
+#define UCT_TEST_IB_AND_GGA_TLS \
```
iyastreb, line 485: Q: maybe add `gga_mlx5` to the  list of TLS and skip when it's not available?

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -237,7 +240,23 @@ enum {
 };
 
 
+typedef struct uct_dma_opaque_mr {
```
iyastreb, line 243: minor: here and below: struct names seem redundant

src/uct/ib/mlx5/gga/gga_mlx5.c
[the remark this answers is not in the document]
evgeny-leksikov: why is it better? the buffer is used on fast path, built-in buffer should improve memory locality, though I dont think performance impact can be proven by benchmarking...

src/uct/ib/mlx5/gga/gga_mlx5.c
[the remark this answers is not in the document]
evgeny-leksikov: when the key is resolved - it's just testing of one value, pending path has much higher overhead, on other hand we can overlap first "heavy" resolution.

src/uct/ib/mlx5/ib_mlx5.h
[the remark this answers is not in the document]
evgeny-leksikov: to suppress compiling warning which is threaded as an error, probably we dont pass other WQEs to functions without type cast or they are allocated differently

src/uct/ib/mlx5/gga/gga_mlx5.c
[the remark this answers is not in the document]
evgeny-leksikov: I think so sinse we have configurable parameter inherited from RC

test/gtest/uct/uct_test.cc
evgeny-leksikov: why is it better? we are testing presence but not amount, IMO `find` is better here also this check was in initial code

test/gtest/uct/uct_test.h
evgeny-leksikov: init/cleanup also takes time

src/uct/ib/mlx5/gga/gga_mlx5.c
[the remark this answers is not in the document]
evgeny-leksikov: > we don't need get_zcopy flow control for GGA, like we have for RDMA_READ

Also, this is not for FC
https://github.com/openucx/ucx/blob/b1a268bdbd09838f6c20105a2a0ff2d41de38b82/src/uct/api/uct.h#L1074

test/gtest/uct/uct_test.cc
iyastreb: Essentially in map count also checks the presence, because the returned value can vary between 0 and 1.
For the the only pros is that it's shorter and looks more readable, but up to you

src/uct/ib/mlx5/gga/gga_mlx5.c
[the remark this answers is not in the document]
yosefe: max_zcopy is for "incast avoidance" for RDMA_READ, IMO it is not needed

src/uct/ib/mlx5/ib_mlx5.h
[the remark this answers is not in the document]
yosefe: can you pls post the error if it's not aligned?

src/uct/ib/mlx5/gga/gga_mlx5.c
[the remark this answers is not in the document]
yosefe: the code will be simpler IMO, less memory waste, and we don't really access the pointer contents so it should not matter. actually if hw is the only one accessing the pointer, it will be more efficient if it will be separated from the memory that the CPU is accessing to avoid false sharing

src/uct/ib/mlx5/gga/gga_mlx5.c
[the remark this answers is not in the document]
yosefe: what does it mean "can overlap first "heavy" resolution"?
pls see rc_mlx5_ep.c - it checks resources first, then resolves the rkey. so IMO better to align gga to such order as well.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2879,9 +2879,9 @@ uct_ib_mlx5_devx_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
      * memory region and the hardware supports it.
      */
     if (ucs_unlikely(memh->atomic_dvmr == NULL) &&
+        !(memh->super.flags & UCT_IB_MEM_IMPORTED) &&
```
yosefe, line 2882: why moved?

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -105,7 +104,8 @@ uct_gga_mlx5_rkey_unpack(uct_component_t *component, const void *rkey_buffer,
     }
 
     rkey_handle->packed_mkey    = *mkey;
-    /* memh and rkey_ob will be initialized on demand */
+    /* memh and rkey_ob will be initialized on demand since MD
+     * (required for memh import) is not available at this point */
```
yosefe, line 108: does it work if user opens 2 UCT MDs, and tries to use same UCT rkey on both MDs?

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -297,26 +300,37 @@ static UCS_CLASS_INIT_FUNC(uct_gga_mlx5_ep_t, const uct_ep_params_t *params)
...
 
-    self->dma_opaque.mr =
-            ibv_reg_mr(md->super.pd,
-                       ucs_align_up_pow2_ptr(&self->dma_opaque.buf[0],
-                                             UCS_SYS_CACHE_LINE_SIZE),
-                       UCT_GGA_MLX5_OPAQUE_MR_LEN,
-                       IBV_ACCESS_LOCAL_WRITE);
+    ret = ucs_posix_memalign(&self->dma_opaque.buf, UCS_SYS_CACHE_LINE_SIZE,
+                             UCT_GGA_MLX5_OPAQUE_BUF_LEN, "gga_dma_opaque_buf");
+    if (ret) {
```
yosefe, line 310: ret != 0

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -297,26 +300,37 @@ static UCS_CLASS_INIT_FUNC(uct_gga_mlx5_ep_t, const uct_ep_params_t *params)
...
+        ucs_error("failed to allocate DMA/MMO opaque buffer");
```
yosefe, line 311: minor: print the size

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -249,15 +250,25 @@ enum {
...
-typedef struct uct_ib_mlx5_dma_wqe {
-    struct mlx5_wqe_ctrl_seg ctrl;
-    uint32_t                 padding;   /* unused for dma */
-    uint32_t                 be_opaque_lkey;
-    uint64_t                 be_opaque_vaddr;
-    struct mlx5_wqe_data_seg gather;
-    struct mlx5_wqe_data_seg scatter;
-} UCS_S_PACKED UCS_V_ALIGNED(sizeof(void*)) uct_ib_mlx5_dma_wqe_t;
+/**
+ * full QWE format:
```
yosefe, line 261: QWE -> WQE

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -249,15 +250,25 @@ enum {
...
+ * struct mlx5_wqe_ctrl_seg;
+ * struct uct_ib_mlx5_dma_qwe_seg;
+ * struct mlx5_wqe_data_seg gather;
+ * struct mlx5_wqe_data_seg scatter;
+ */
+struct uct_ib_mlx5_dma_qwe_seg {
```
yosefe, line 267: 1. uct_ib_mlx5_dma_seg
2. use typedef?

src/uct/ib/mlx5/ib_mlx5_log.c
```diff
@@ -313,6 +343,13 @@ static void uct_ib_mlx5_wqe_dump(uct_ib_iface_t *iface, void *wqe, void *qstart,
                            s, ends - s);
     s += strlen(s);
 
+#if HAVE_MLX5_MMO
+    if (opcode == MLX5_OPCODE_MMO) {
+        uct_ib_mlx5_dump_dma_wqe(wqe, s, ends - s);
+        return;
+    }
+#endif
```
yosefe, line 351: can we unitfy the code of printing DMA/GGA wqe with the existing code flow? instead of making a special case for it.
Seems the only "special" thing about GGA is that it has DMA segment. So it can be treated same way as atomic/rdma segments for example.

src/uct/ib/mlx5/rc/rc_mlx5.inl
```diff
@@ -887,30 +889,33 @@ void uct_rc_mlx5_txqp_dptr_post_iov(uct_rc_mlx5_iface_common_t *iface, int qp_ty
 #if HAVE_MLX5_MMO
     case MLX5_OPCODE_MMO|UCT_RC_MLX5_OPCODE_FLAG_MMO_PUT:
     case MLX5_OPCODE_MMO|UCT_RC_MLX5_OPCODE_FLAG_MMO_GET:
-        gga_wqe                  = txwq->curr;
+        dma_seg                  = UCS_PTR_TYPE_OFFSET(ctrl, *ctrl);
```
yosefe, line 892: use next_seg to align with existing code

src/uct/ib/mlx5/rc/rc_mlx5.inl
```diff
@@ -887,30 +889,33 @@ void uct_rc_mlx5_txqp_dptr_post_iov(uct_rc_mlx5_iface_common_t *iface, int qp_ty
...
-                                     iov[0].length,
-                                     uct_ib_memh_get_lkey(iov[0].memh));
+            dma_src_buf = (const void*)remote_addr;
+            dma_src_key = rkey;
+            dma_dst_buf = iov[0].buffer;
+            dma_dst_key = uct_ib_memh_get_lkey(iov[0].memh);
         }
 
-        wqe_size = sizeof(*gga_wqe);
+        dptr = UCS_PTR_TYPE_OFFSET(dma_seg, *dma_seg);
```
yosefe, line 912: shall we use uct_ib_mlx5_txwq_wrap_exact ?

evgeny-leksikov: why? can single WQE be wrapped by segments?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
yosefe: why not, isn't it registered by uct_ib_mlx5_devx_mem_reg?

src/uct/ib/mlx5/gga/gga_mlx5.c
yosefe: > IIUC you mean use rkey with 2 different EPs which created on different ifaces + different MDs? this is a good question, I'll see how to add such test case.

Yes, this is what i've meant. Seems need a hash per md of rkey to imported key.

src/uct/ib/mlx5/rc/rc_mlx5.inl
yosefe: so pls use uct_ib_mlx5_txwq_wrap_none

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
evgeny-leksikov: `uct_ib_mlx5_devx_mem_attach` works for `UCT_IB_MEM_IMPORTED`

src/uct/ib/base/ib_log.c
```diff
@@ -70,8 +58,8 @@ void uct_ib_log_dump_sg_list(uct_ib_iface_t *iface, uct_am_trace_type_t type,
         if (inline_bitmap & UCS_BIT(i)) {
             snprintf(s, ends - s, " [inl len %d]", sg_list[i].length);
         } else {
-            snprintf(s, ends - s, " %s[va 0x%"PRIx64" len %d lkey 0x%x]",
-                     uct_ib_mlx5_dump_get_dseg_prefix(op_flags, i),
+            snprintf(s, ends - s, " %.1s[va 0x%"PRIx64" len %d lkey 0x%x]",
+                     sg_prefixes == NULL ? "" : &sg_prefixes[i],
```
yosefe, line 62: 1. %1s is cool
2. pls add ( ) around == operation

yosefe (approved):

PR #9995, opened 2024-07-07
UCP/WIREUP/IB: provide a reason when a resource's lane is unreachable

amastbaum:
## What
During the wireup process, provide a reason when a resource's lane is unreachable.

## Why
Users need more information about why a device is not reachable after an unsuccessful connection establishment. This information should be passed from UCT to UCP. 

## How ?
Pass a string to select/search_lane functions so in case the wireup process fails, the reason will be printed out in an upper layer.

### DIFF
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -387,7 +387,8 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
         const ucp_wireup_criteria_t *criteria, ucp_tl_bitmap_t tl_bitmap,
         uint64_t remote_md_map, uint64_t local_dev_bitmap,
         uint64_t remote_dev_bitmap, int show_error,
-        ucp_wireup_select_info_t *select_info)
+        ucp_wireup_select_info_t *select_info, char *info_str,
+        size_t info_str_size)
 {
     UCS_STRING_BUFFER_ONSTACK(missing_flags_str,
                               UCP_WIREUP_MAX_FLAGS_STRING_SIZE);
@@ -599,7 +600,8 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
         UCS_STATIC_BITMAP_FOR_EACH_BIT(addr_index, &rsc_addr_index_map) {
             ae = &address->address_list[addr_index];
             if (!ucp_wireup_is_reachable(ep, select_params->ep_init_flags,
-                                         rsc_index, ae)) {
+                                         rsc_index, ae, info_str,
+                                         info_str_size)) {
                 /* Must be reachable device address, on same transport */
                 continue;
             }
@@ -625,8 +627,6 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
         /* If a local resource cannot reach any of the remote addresses,
          * generate debug message. */
         if (!is_reachable) {
-            ucs_trace(UCT_TL_RESOURCE_DESC_FMT" : unreachable ",
-                      UCT_TL_RESOURCE_DESC_ARG(resource));
             snprintf(p, endp - p, UCT_TL_RESOURCE_DESC_FMT" - %s, ",
                      UCT_TL_RESOURCE_DESC_ARG(resource),
                      ucs_status_string(UCS_ERR_UNREACHABLE));
@@ -908,7 +908,7 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_add_memaccess_lanes(
     status = ucp_wireup_select_transport(select_ctx, select_params,
                                          &mem_criteria, tl_bitmap,
                                          remote_md_map, UINT64_MAX, UINT64_MAX,
-                                         !allow_am, &select_info);
+                                         !allow_am, &select_info, NULL, 0);
     if (status == UCS_OK) {
         /* Add to the list of lanes */
         status = ucp_wireup_add_lane(select_params, &select_info, lane_type,
@@ -954,7 +954,8 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_add_memaccess_lanes(
         status = ucp_wireup_select_transport(select_ctx, select_params,
                                              &mem_criteria, tl_bitmap,
                                              remote_md_map, UINT64_MAX,
-                                             UINT64_MAX, 0, &select_info);
+                                             UINT64_MAX, 0, &select_info,
+                                             NULL, 0);
         /* Break if: */
         /* - transport selection wasn't OK */
         if ((status != UCS_OK) ||
@@ -1400,7 +1401,8 @@ ucp_wireup_is_am_required(const ucp_wireup_select_params_t *select_params,
 static ucs_status_t
 ucp_wireup_add_am_lane(const ucp_wireup_select_params_t *select_params,
                        ucp_wireup_select_info_t *am_info,
-                       ucp_wireup_select_context_t *select_ctx)
+                       ucp_wireup_select_context_t *select_ctx,
+                       char *info_string, size_t info_string_length)
 {
     ucp_worker_h worker            = select_params->ep->worker;
     ucp_tl_bitmap_t tl_bitmap      = select_params->tl_bitmap;
@@ -1440,7 +1442,8 @@ ucp_wireup_add_am_lane(const ucp_wireup_select_params_t *select_params,
         status = ucp_wireup_select_transport(select_ctx, select_params,
                                              &criteria, tl_bitmap, UINT64_MAX,
                                              UINT64_MAX, UINT64_MAX, 1,
-                                             am_info);
+                                             am_info, info_string,
+                                             info_string_length);
         if (status != UCS_OK) {
             return status;
         }
@@ -1623,7 +1626,8 @@ ucp_wireup_add_bw_lanes(const ucp_wireup_select_params_t *select_params,
             status = ucp_wireup_select_transport(select_ctx, select_params,
                                                  &bw_info->criteria, tl_bitmap,
                                                  UINT64_MAX, local_dev_bitmap,
-                                                 remote_dev_bitmap, 0, sinfo);
+                                                 remote_dev_bitmap, 0, sinfo,
+                                                 NULL, 0);
             if (status != UCS_OK) {
                 ucs_array_pop_back(&sinfo_array);
                 break;
@@ -2046,7 +2050,7 @@ ucp_wireup_add_tag_lane(const ucp_wireup_select_params_t *select_params,
     status = ucp_wireup_select_transport(select_ctx, select_params, &criteria,
                                          ucp_tl_bitmap_max, UINT64_MAX,
                                          UINT64_MAX, UINT64_MAX, 0,
-                                         &select_info);
+                                         &select_info, NULL, 0);
     if ((status == UCS_OK) &&
         (ucp_score_cmp(select_info.score,
                        am_info->score) >= 0)) {
@@ -2191,7 +2195,7 @@ ucp_wireup_add_keepalive_lane(const ucp_wireup_select_params_t *select_params,
 
     status = ucp_wireup_select_transport(select_ctx, select_params, &criteria,
                                          *tl_bitmap, UINT64_MAX, UINT64_MAX,
-                                         UINT64_MAX, 0, &select_info);
+                                         UINT64_MAX, 0, &select_info, NULL, 0);
     if (status == UCS_OK) {
         return ucp_wireup_add_lane(select_params, &select_info,
                                    UCP_LANE_TYPE_KEEPALIVE, /* show error */ 1,
@@ -2213,7 +2217,8 @@ ucp_wireup_select_context_init(ucp_wireup_select_context_t *select_ctx)
 static UCS_F_NOINLINE ucs_status_t
 ucp_wireup_search_lanes(const ucp_wireup_select_params_t *select_params,
                         ucp_err_handling_mode_t err_mode,
-                        ucp_wireup_select_context_t *select_ctx)
+                        ucp_wireup_select_context_t *select_ctx,
+                        char *info_string, size_t info_string_length)
 {
     ucp_wireup_select_info_t am_info;
     ucs_status_t status;
@@ -2239,7 +2244,8 @@ ucp_wireup_search_lanes(const ucp_wireup_select_params_t *select_params,
 
     /* Add AM lane only after RMA/AMO was selected to be aware
      * about whether they need emulation over AM or not */
-    status = ucp_wireup_add_am_lane(select_params, &am_info, select_ctx);
+    status = ucp_wireup_add_am_lane(select_params, &am_info, select_ctx,
+                                    info_string, info_string_length);
     if (status != UCS_OK) {
         return status;
     }
@@ -2502,6 +2508,9 @@ ucp_wireup_select_lanes(ucp_ep_h ep, unsigned ep_init_flags,
 {
     ucp_worker_h worker                = ep->worker;
     ucp_tl_bitmap_t scalable_tl_bitmap = worker->scalable_tl_bitmap;
+    /* TO DO: remove initializtion after all ucp_wireup_add_X_lanes functions
+       will support specifying a reason */
+    char wireup_info[256]              = {0};
     ucp_wireup_select_context_t select_ctx;
     ucp_wireup_select_params_t select_params;
     ucs_status_t status;
@@ -2512,7 +2521,8 @@ ucp_wireup_select_lanes(ucp_ep_h ep, unsigned ep_init_flags,
         ucp_wireup_select_params_init(&select_params, ep, ep_init_flags,
                                       remote_address, scalable_tl_bitmap, 0);
         status = ucp_wireup_search_lanes(&select_params, key->err_mode,
-                                         &select_ctx);
+                                         &select_ctx, wireup_info,
+                                         sizeof(wireup_info));
         if (status == UCS_OK) {
             goto out;
         }
@@ -2525,8 +2535,13 @@ ucp_wireup_select_lanes(ucp_ep_h ep, unsigned ep_init_flags,
     ucp_wireup_select_params_init(&select_params, ep, ep_init_flags,
                                   remote_address, tl_bitmap, show_error);
     status = ucp_wireup_search_lanes(&select_params, key->err_mode,
-                                     &select_ctx);
+                                     &select_ctx, wireup_info,
+                                     sizeof(wireup_info));
     if (status != UCS_OK) {
+        if (wireup_info[0] != '\0') {
+            ucs_diag("destination is unreachable [%s]", wireup_info);
+        }
+
         return status;
     }
 
@@ -2567,7 +2582,7 @@ ucp_wireup_select_aux_transport(ucp_ep_h ep, unsigned ep_init_flags,
     status = ucp_wireup_select_transport(&select_ctx, &select_params, &criteria,
                                          ucp_tl_bitmap_max, UINT64_MAX,
                                          UINT64_MAX, UINT64_MAX, 0,
-                                         select_info);
+                                         select_info, NULL, 0);
     if (status == UCS_OK) {
         return UCS_OK;
     }
@@ -2577,5 +2592,6 @@ ucp_wireup_select_aux_transport(ucp_ep_h ep, unsigned ep_init_flags,
     ucp_wireup_fill_aux_criteria(&criteria, ep_init_flags, 0);
     return ucp_wireup_select_transport(&select_ctx, &select_params, &criteria,
                                        ucp_tl_bitmap_max, UINT64_MAX,
-                                       UINT64_MAX, UINT64_MAX, 1, select_info);
+                                       UINT64_MAX, UINT64_MAX, 1, select_info,
+                                       NULL, 0);
 }
--- a/src/ucp/wireup/wireup.c
+++ b/src/ucp/wireup/wireup.c
@@ -1243,7 +1243,8 @@ static void ucp_wireup_print_config(ucp_worker_h worker,
 
 int ucp_wireup_is_reachable(ucp_ep_h ep, unsigned ep_init_flags,
                             ucp_rsc_index_t rsc_index,
-                            const ucp_address_entry_t *ae)
+                            const ucp_address_entry_t *ae,
+                            char *info_str, size_t info_str_size)
 {
     ucp_context_h context      = ep->worker->context;
     ucp_worker_iface_t *wiface = ucp_worker_iface(ep->worker, rsc_index);
@@ -1254,11 +1255,22 @@ int ucp_wireup_is_reachable(ucp_ep_h ep, unsigned ep_init_flags,
         .iface_addr  = ae->iface_addr,
     };
 
-    return (context->tl_rscs[rsc_index].tl_name_csum == ae->tl_name_csum) &&
-           (/* assume reachability is checked by CM, if EP selects lanes
-             * during CM phase */
-            (ep_init_flags & UCP_EP_INIT_CM_PHASE) ||
-            uct_iface_is_reachable_v2(wiface->iface, &params));
+    if (info_str != NULL) {
+        params.field_mask        |=
+                                UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING |
+                                UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING_LENGTH;
+        params.info_string        = info_str;
+        params.info_string_length = info_str_size;
+    }
+
+    if (context->tl_rscs[rsc_index].tl_name_csum != ae->tl_name_csum) {
+        return 0;
+    }
+
+    /* assume reachability is checked by CM, if EP selects lanes
+     * during CM phase */
+    return (ep_init_flags & UCP_EP_INIT_CM_PHASE) ||
+           uct_iface_is_reachable_v2(wiface->iface, &params);
 }
 
 static void
@@ -1280,7 +1292,7 @@ ucp_wireup_get_reachable_mds(ucp_ep_h ep, unsigned ep_init_flags,
     ae_dst_md_map = 0;
     UCS_STATIC_BITMAP_FOR_EACH_BIT(rsc_index, &context->tl_bitmap) {
         ucp_unpacked_address_for_each(ae, remote_address) {
-            if (ucp_wireup_is_reachable(ep, ep_init_flags, rsc_index, ae)) {
+            if (ucp_wireup_is_reachable(ep, ep_init_flags, rsc_index, ae, NULL, 0)) {
                 ae_dst_md_map         |= UCS_BIT(ae->md_index);
                 dst_md_index           = context->tl_rscs[rsc_index].md_index;
                 ae_cmpts[ae->md_index] = context->tl_mds[dst_md_index].cmpt_index;
--- a/src/uct/base/uct_iface.c
+++ b/src/uct/base/uct_iface.c
@@ -242,6 +242,26 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
     return !memcmp(device_addr, dev_addr, attr.device_addr_len);
 }
 
+void uct_iface_fill_info_str_buf(const uct_iface_is_reachable_params_t *params,
+                                 const char *fmt, ...)
+{
+    char *info_str_buf      = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                              params, info_string, INFO_STRING,
+                                              NULL);
+    size_t info_str_buf_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                              params, info_string_length,
+                                              INFO_STRING_LENGTH, 0);
+    va_list ap;
+
+    if (info_str_buf == NULL) {
+        return;
+    }
+
+    va_start(ap, fmt);
+    ucs_vsnprintf_safe(info_str_buf, info_str_buf_len, fmt, ap);
+    va_end(ap);
+}
+
 int uct_iface_is_reachable_params_valid(
         const uct_iface_is_reachable_params_t *params, uint64_t flags)
 {
@@ -289,6 +309,17 @@ int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
                               const uct_iface_is_reachable_params_t *params)
 {
     const uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
+    char *info_str_buf            = UCS_PARAM_VALUE(
+                                          UCT_IFACE_IS_REACHABLE_FIELD, params,
+                                          info_string, INFO_STRING, NULL);
+    size_t info_str_buf_len       = UCS_PARAM_VALUE(
+                                          UCT_IFACE_IS_REACHABLE_FIELD, params,
+                                          info_string_length,
+                                          INFO_STRING_LENGTH, 0);
+
+    if ((info_str_buf != NULL) && (info_str_buf_len > 0)) {
+        info_str_buf[0] = '\0';
+    }
 
     return iface->internal_ops->iface_is_reachable_v2(tl_iface, params);
 }
--- a/src/uct/base/uct_iface.h
+++ b/src/uct/base/uct_iface.h
@@ -912,6 +912,9 @@ void uct_iface_get_local_address(uct_iface_local_addr_ns_t *addr_ns,
 int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
                                  ucs_sys_namespace_type_t sys_ns_type);
 
+void uct_iface_fill_info_str_buf(const uct_iface_is_reachable_params_t *params,
+                                 const char *fmt, ...);
+
 int uct_iface_is_reachable_params_valid(
         const uct_iface_is_reachable_params_t *params, uint64_t flags);
 
--- a/src/uct/ib/base/ib_iface.c
+++ b/src/uct/ib/base/ib_iface.c
@@ -667,7 +667,8 @@ uct_ib_iface_roce_to_sockaddr(sa_family_t af, const void *gid,
 static int
 uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
                                const uct_ib_address_t *remote_ib_addr,
-                               unsigned prefix_bits)
+                               unsigned prefix_bits,
+                               const uct_iface_is_reachable_params_t *params)
 {
     sa_family_t local_ib_addr_af         = local_gid_info->roce_info.addr_family;
     uct_ib_roce_version_t local_roce_ver = local_gid_info->roce_info.ver;
@@ -695,14 +696,16 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
     remote_roce_ver = uct_ib_address_flags_get_roce_version(remote_ib_addr_flags);
 
     if (local_roce_ver != remote_roce_ver) {
-        ucs_debug("different RoCE versions detected. local %s (gid=%s)"
-                  "remote %s (gid=%s)",
-                  uct_ib_roce_version_str(local_roce_ver),
-                  uct_ib_gid_str(&local_gid_info->gid, local_str,
-                                 sizeof(local_str)),
-                  uct_ib_roce_version_str(remote_roce_ver),
-                  uct_ib_gid_str((union ibv_gid*)(remote_ib_addr + 1),
-                                 remote_str, sizeof(remote_str)));
+        uct_iface_fill_info_str_buf(
+                        params,
+                        "different RoCE versions detected. local %s (gid=%s)"
+                        "remote %s (gid=%s)",
+                        uct_ib_roce_version_str(local_roce_ver),
+                        uct_ib_gid_str(&local_gid_info->gid, local_str,
+                                       sizeof(local_str)),
+                        uct_ib_roce_version_str(remote_roce_ver),
+                        uct_ib_gid_str((union ibv_gid*)(remote_ib_addr + 1),
+                                       remote_str, sizeof(remote_str)));
         return 0;
     }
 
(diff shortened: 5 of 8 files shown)


gleon99:
Please edit PR description:
1. Capitals, grammar, styling.
2. Add the "Why?" section, as per our conventions.

src/uct/ib/base/ib_iface.c
```diff
@@ -787,26 +829,37 @@ int uct_ib_iface_is_reachable_v2(const uct_iface_h tl_iface,
...
                 params, UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR)) {
+        uct_iface_populate_info_str_buf(params, "invalid is_reachable params");
         return 0;
     }
 
     device_addr = (const uct_ib_address_t*)
             UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params, device_addr,
                             DEVICE_ADDR, NULL);
     if (device_addr == NULL) {
+        uct_iface_populate_info_str_buf(params, "invalid ib device address");
```
gleon99, line 840: `ib` -> `IB`

src/ucp/wireup/select.c
```diff
@@ -419,6 +421,13 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
     uint8_t priority;
     ucp_md_index_t md_index;
 
+    wireup_info = (char *)ucs_malloc(UCP_WIREUP_INFO_STRING_SIZE,
```
gleon99, line 424: Why not on stack?

src/uct/base/uct_iface.c
```diff
@@ -242,6 +242,27 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
     return !memcmp(device_addr, dev_addr, attr.device_addr_len);
 }
 
+void uct_iface_populate_info_str_buf(
+                                const uct_iface_is_reachable_params_t *params,
+                                const char *fmt, ...)
+{
+    char *info_str_buf = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
```
gleon99, line 249: Align

src/uct/base/uct_iface.c
```diff
@@ -242,6 +242,27 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
     return !memcmp(device_addr, dev_addr, attr.device_addr_len);
 }
 
+void uct_iface_populate_info_str_buf(
```
gleon99, line 245: Maybe `fill` instead?

gleon99: Maybe make it a macro instead? Then it is inlined and somewhat simpler.

src/ucp/wireup/select.c
```diff
@@ -594,12 +603,15 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
             UCS_STATIC_BITMAP_AND_INPLACE(&rsc_addr_index_map, addr_index_map);
         }
 
+        wireup_info[0] = '\0';
```
gleon99, line 606: Can remove? (see `uct_iface_is_reachable_params_valid`)

src/uct/ib/base/ib_iface.c
```diff
@@ -787,26 +829,37 @@ int uct_ib_iface_is_reachable_v2(const uct_iface_h tl_iface,
 
     if (!uct_iface_is_reachable_params_valid(
                 params, UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR)) {
+        uct_iface_populate_info_str_buf(params, "invalid is_reachable params");
```
gleon99, line 832: Can be more specific here.

src/uct/ib/base/ib_iface.c
```diff
@@ -764,16 +796,26 @@ static int uct_ib_iface_dev_addr_is_reachable(uct_ib_iface_t *iface,
         }
 
         /* Check FLID route: is enabled locally, and remote GID has it */
-        return (uct_ib_iface_is_flid_enabled(iface) &&
-                uct_ib_iface_gid_extract_flid(&params.gid) != 0);
+        if (!uct_ib_iface_is_flid_enabled(iface) ||
+            !uct_ib_iface_gid_extract_flid(&params.gid)) {
+            uct_iface_populate_info_str_buf(is_reachable_params,
+                                            "invalid flid route");
```
gleon99, line 802: `FLID`

src/uct/base/uct_iface.c
```diff
@@ -242,6 +242,27 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
...
+                                    NULL);
+    size_t info_str_buf_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                        params, info_string_length,
+                                        INFO_STRING_LENGTH, 0);
+
+    va_list args;
+    va_start(args, fmt);
+
+    if (info_str_buf && info_str_buf_len > 0 && info_str_buf[0] == '\0') {
+        snprintf(info_str_buf, info_str_buf_len, fmt, args);
```
gleon99, line 260: `ucs_snprintf_safe`

gleon99: AFAIU, it might be called several times in a single flow (e.g `uct_ib_iface_dev_addr_is_reachable` then in the caller `uct_ib_iface_is_reachable_v2`). The strings overwrite each other?

src/ucp/wireup/select.c
```diff
@@ -397,6 +398,7 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
     ucp_context_h context                         = worker->context;
     ucp_wireup_select_info_t sinfo                = {0};
     int found                                     = 0;
+    char *wireup_info;
```
yosefe, line 401: move it after local_iface_flags since initialized vars should before un-initialized ones

src/ucp/wireup/select.c
```diff
@@ -625,8 +637,8 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
         /* If a local resource cannot reach any of the remote addresses,
          * generate debug message. */
         if (!is_reachable) {
-            ucs_trace(UCT_TL_RESOURCE_DESC_FMT" : unreachable ",
-                      UCT_TL_RESOURCE_DESC_ARG(resource));
+            ucs_diag(UCT_TL_RESOURCE_DESC_FMT" : unreachable [%s]",
+                     UCT_TL_RESOURCE_DESC_ARG(resource), wireup_info);
```
yosefe, line 641: can wireup_info be empty string here?

src/ucp/wireup/select.c
```diff
@@ -639,6 +651,8 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
         *(p - 2) = '\0'; /* trim last "," */
     }
 
+    free(wireup_info);
```
yosefe, line 654: ucs_free

src/ucp/wireup/select.c
yosefe: no need to use (char*)

src/ucp/wireup/wireup.c
```diff
@@ -1243,22 +1243,33 @@ static void ucp_wireup_print_config(ucp_worker_h worker,
...
 {
     ucp_context_h context      = ep->worker->context;
     ucp_worker_iface_t *wiface = ucp_worker_iface(ep->worker, rsc_index);
     uct_iface_is_reachable_params_t params = {
         .field_mask  = UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR |
                        UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR,
-        .device_addr = ae->dev_addr,
-        .iface_addr  = ae->iface_addr,
+        .device_addr        = ae->dev_addr,
+        .iface_addr         = ae->iface_addr,
```
yosefe, line 1255: pls revert the indentation

src/uct/base/uct_iface.c
```diff
@@ -242,6 +242,27 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
...
+                                    params, info_string, INFO_STRING,
+                                    NULL);
+    size_t info_str_buf_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                        params, info_string_length,
+                                        INFO_STRING_LENGTH, 0);
+
+    va_list args;
+    va_start(args, fmt);
+
+    if (info_str_buf && info_str_buf_len > 0 && info_str_buf[0] == '\0') {
```
yosefe, line 259: 1. pls use ( ) around internal expressions
2. why need to check info_str_buf[0] == '\0' ?
3. use (info_str_buf != NULL) - check for NULL explicitly
4. probably no need to check info_str_buf_len > 0, it's already handled by ucs_snprintf_safe

src/uct/base/uct_iface.c
```diff
@@ -242,6 +242,27 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
...
+                                const char *fmt, ...)
+{
+    char *info_str_buf = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                    params, info_string, INFO_STRING,
+                                    NULL);
+    size_t info_str_buf_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                        params, info_string_length,
+                                        INFO_STRING_LENGTH, 0);
+
+    va_list args;
```
yosefe, line 256: 1. rename args to ap
2. space line after

src/uct/ib/base/ib_iface.c
```diff
@@ -650,6 +651,11 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
         ucs_debug("different addr_family detected. local %s remote %s",
                   ucs_sockaddr_address_family_str(local_ib_addr_af),
                   ucs_sockaddr_address_family_str(remote_ib_addr_af));
+        uct_iface_populate_info_str_buf(
```
yosefe, line 654: can we move the ucs_debug message inside uct_iface_populate_info_str_buf function? since essentially they are printing the same thing

src/uct/ib/base/ib_iface.c
```diff
@@ -754,6 +784,8 @@ static int uct_ib_iface_dev_addr_is_reachable(uct_ib_iface_t *iface,
         !((params.pkey | iface->pkey) & UCT_IB_PKEY_MEMBERSHIP_MASK) ||
         /* PKEY values have to be equal */
         ((params.pkey ^ iface->pkey) & UCT_IB_PKEY_PARTITION_MASK)) {
+        uct_iface_populate_info_str_buf(is_reachable_params,
+                                        "none of the pkeys has full membership");
```
yosefe, line 788: let's split to 2 cases:
1. "both local and remote pkeys 0x%x have partial membership"
2. "local pkey 0x%x differs from remote pkey 0x%x"

src/uct/ib/base/ib_iface.c
yosefe: 1. pls use uct_ib_iface_gid_extract_flid == 0 (instead of !uct_ib_iface_gid_extract_flid)
2. "IB subnet prefix differs 0x%"PRIx64" vs 0x%"PRIx64" and FLID routing is disabled"

src/uct/ib/base/ib_iface.c
```diff
@@ -764,16 +796,26 @@ static int uct_ib_iface_dev_addr_is_reachable(uct_ib_iface_t *iface,
...
          * and address families have to be the same */
         return uct_ib_iface_roce_is_reachable(&iface->gid_info, ib_addr,
-                                              iface->addr_prefix_bits);
+                                              iface->addr_prefix_bits,
+                                              is_reachable_params);
     } else {
         /* local and remote have different link layers and therefore are unreachable */
+        uct_iface_populate_info_str_buf(
+                            is_reachable_params,
+                            "local and remote have different link layers");
```
yosefe, line 818: "link layers differ %s vs %s"

src/uct/ib/base/ib_iface.c
yosefe: in case of invalid params, not sure we want to return any info string

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -921,6 +921,9 @@ uct_dc_mlx5_iface_is_reachable_v2(const uct_iface_h tl_iface,
         same_version = ((addr->flags & UCT_DC_MLX5_IFACE_ADDR_DC_VERS) ==
                         iface->version_flag);
         if (!same_version || !same_tm) {
+            uct_iface_populate_info_str_buf(
+                        params, "incompatible tag matching and dc version");
```
yosefe, line 925: need to split to different case and print the different versions or the different support of HW TM

src/uct/ib/rc/accel/rc_mlx5_iface.c
```diff
@@ -713,6 +713,9 @@ uct_rc_mlx5_iface_is_reachable_v2(const uct_iface_h tl_iface,
                                  iface_addr, IFACE_ADDR, NULL);
     /* Check hardware tag matching compatibility */
     if ((iface_addr != NULL) && (my_type != *(uint8_t*)iface_addr)) {
+        uct_iface_populate_info_str_buf(
+                            params, "incompatible hardware tag matching");
```
yosefe, line 717: need to print what is not compatible

yosefe (commented):
can you pls post example error messages from UCP level?
Maybe add guest for it?

src/ucp/wireup/select.c
amastbaum: It can't be empty unless we choose to not return info string in some cases (such invalid params).. 
At first I was making sure it's filled in every scenario but you're right that the code doesn't ensure that

src/uct/base/uct_iface.c
amastbaum: @gleon99  This is why I am checking `info_str_buf` is empty (`info_str_buf[0] == '\0'`), so it won't override an existing reason. I assume the first time `info_str_buf` is filled happens in the deepest level of this check (== the most descriptive one), yet I want to provide a more basic reason in case it wasn't filled earlier in the flow.

@yosefe it answers your other question about why we need `info_str_buf[0] == '\0'` .

src/uct/base/uct_iface.h
```diff
@@ -256,6 +256,21 @@ enum {
 UCS_CLASS_DECLARE(uct_iface_h, uct_iface_ops_t, uct_md_h);
 UCS_CLASS_DECLARE(uct_ep_t, uct_iface_h);
 
+/**
+ * 
+ */
+#define UCT_IFACE_FILL_INFO_STR_BUF(params, fmt, ...) do { \
```
yosefe, line 262: can it be a function instead of a macro?

src/uct/base/uct_iface.h
```diff
@@ -256,6 +256,21 @@ enum {
...
+    char *info_str_buf      = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, \
+                                              params, info_string, INFO_STRING, \
+                                              NULL); \
+    size_t info_str_buf_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, \
+                                              params, info_string_length, \
+                                              INFO_STRING_LENGTH, 0); \
+    if ((info_str_buf != NULL) && (info_str_buf[0] == '\0')) { \
```
yosefe, line 269: why need to check info_str_buf[0] ?

src/ucp/wireup/select.c
```diff
@@ -2539,8 +2539,13 @@ ucp_wireup_select_lanes(ucp_ep_h ep, unsigned ep_init_flags,
     ucp_wireup_select_params_init(&select_params, ep, ep_init_flags,
                                   remote_address, tl_bitmap, show_error);
     status = ucp_wireup_search_lanes(&select_params, key->err_mode,
-                                     &select_ctx);
+                                     &select_ctx, wireup_info,
+                                     sizeof(wireup_info));
     if (status != UCS_OK) {
+        if (wireup_info[0] != '\0') {
+            ucs_diag("Destination is unreachable [%s]", wireup_info);
```
yosefe, line 2546: small case

src/ucp/wireup/select.c
```diff
@@ -2519,14 +2517,16 @@ ucp_wireup_select_lanes(ucp_ep_h ep, unsigned ep_init_flags,
     ucp_wireup_select_context_t select_ctx;
     ucp_wireup_select_params_t select_params;
     ucs_status_t status;
+    char wireup_info[UCP_WIREUP_INFO_STRING_SIZE] = {0};
```
yosefe, line 2520: why need to initialize it?

src/ucp/wireup/select.c
```diff
@@ -603,15 +596,18 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
             UCS_STATIC_BITMAP_AND_INPLACE(&rsc_addr_index_map, addr_index_map);
         }
 
-        wireup_info[0] = '\0';
+        if (info_str != NULL) {
+            info_str[0] = '\0';
+        }
```
yosefe, line 601: can we remove?

src/uct/ib/base/ib_iface.c
```diff
@@ -693,7 +682,7 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
     status = ucs_sockaddr_inet_addr_size(local_ib_addr_af, &addr_size);
     if (status != UCS_OK) {
         ucs_error("failed to detect RoCE address size");
-        uct_iface_populate_info_str_buf(
+        UCT_IFACE_FILL_INFO_STR_BUF(
```
yosefe, line 685: do we also need the ucs_error above?

src/uct/ib/base/ib_iface.c
```diff
@@ -780,12 +769,20 @@ static int uct_ib_iface_dev_addr_is_reachable(
...
-                                        "none of the pkeys has full membership");
+    /* at least one PKEY has to be with full membership */
+    if (!((params.pkey | iface->pkey) & UCT_IB_PKEY_MEMBERSHIP_MASK)) {
+        UCT_IFACE_FILL_INFO_STR_BUF(
+                    is_reachable_params,
+                    "both local and remote pkeys (0x%x, 0x%x) "
+                    "have partial membership",
+                    iface->pkey, params.pkey);
+        return 0;
+    } else if (/* PKEY values have to be equal */
```
yosefe, line 780: can remove the else since have a return in line 779

src/uct/ib/base/ib_iface.c
```diff
@@ -796,10 +793,16 @@ static int uct_ib_iface_dev_addr_is_reachable(
...
         /* Check FLID route: is enabled locally, and remote GID has it */
-        if (!uct_ib_iface_is_flid_enabled(iface) ||
-            !uct_ib_iface_gid_extract_flid(&params.gid)) {
-            uct_iface_populate_info_str_buf(is_reachable_params,
-                                            "invalid flid route");
+        if (!uct_ib_iface_is_flid_enabled(iface)) {
+            UCT_IFACE_FILL_INFO_STR_BUF(is_reachable_params,
+                                        "FLID routing is disabled");
+            return 0;
+        } else if (uct_ib_iface_gid_extract_flid(&params.gid) == 0) {
```
yosefe, line 800: can remove the else since have a return in line 799

src/uct/ib/base/ib_iface.c
```diff
@@ -813,9 +816,12 @@ static int uct_ib_iface_dev_addr_is_reachable(
                                               is_reachable_params);
     } else {
         /* local and remote have different link layers and therefore are unreachable */
-        uct_iface_populate_info_str_buf(
-                            is_reachable_params,
-                            "local and remote have different link layers");
+        UCT_IFACE_FILL_INFO_STR_BUF(
+                        is_reachable_params,
+                        "link layers differ %s (local) vs %s (remote)",
+                        is_local_eth ? "ethernet" : "IB",
```
yosefe, line 822: ethernet -> RoCE

src/uct/ib/base/ib_iface.c
```diff
@@ -813,9 +816,12 @@ static int uct_ib_iface_dev_addr_is_reachable(
...
+                        ib_addr->flags & UCT_IB_ADDRESS_FLAG_LINK_LAYER_ETH ?
+                        "ethernet" : "IB");
```
yosefe, line 824: ethernet -> RoCE

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -920,10 +920,19 @@ uct_dc_mlx5_iface_is_reachable_v2(const uct_iface_h tl_iface,
...
-                        params, "incompatible tag matching and dc version");
-
+        if (!same_version) {
+            UCT_IFACE_FILL_INFO_STR_BUF(
+                        params,
+                        "incompatible dc version, %u (local) vs. %u (remote)",
+                        iface->version_flag,
+                        addr->flags & UCT_DC_MLX5_IFACE_ADDR_DC_VERS);
+            return 0;
+        } else if (!same_tm) {
```
yosefe, line 930: can remove else here too

src/uct/ib/rc/accel/rc_mlx5_iface.c
```diff
@@ -707,14 +707,21 @@ uct_rc_mlx5_iface_is_reachable_v2(const uct_iface_h tl_iface,
...
-        uct_iface_populate_info_str_buf(
-                            params, "incompatible hardware tag matching");
+    if ((iface_addr != NULL) && (my_type != remote_type)) {
+        UCT_IFACE_FILL_INFO_STR_BUF(
+                    params, "incompatible hardware tag matching. "
+                    "%s (local) vs %s (remote)",
+                    my_type == UCT_RC_MLX5_IFACE_ADDR_TYPE_TM ?
+                                        "tag matching" : "basic",
+                    remote_type == UCT_RC_MLX5_IFACE_ADDR_TYPE_TM ?
+                                        "tag matching" : "basic");
```
yosefe, line 724: 1. use ( ) around == expression
2. can we unite common code in these 2 arguments?

src/uct/base/uct_iface.h
amastbaum: I originally made it a function but Leonid suggested otherwise (https://github.com/openucx/ucx/pull/9995#discussion_r1674333725). I am actually not sure what is preferable here and if the advantages of using a macro are worth it this time because it will not be called many times and it's a slow path so maybe we don't need to limit function calls that much

src/uct/base/uct_iface.h
amastbaum: If this routine is called more than once in a flow, it'd probably be a case where the first call is in the deepest level of the tree and the second call is a more generic description (for cases that no specific description is provided), and if so then we would like to choose the first description. But after a second look at the code I don't find such cases so I'll remove it

src/ucp/wireup/select.c
amastbaum: no need

src/uct/base/uct_iface.h
yosefe: the advantage of function is easier to debug/maintain, analyze backtrace, etc. since it is not fast path macro does not have an advantage IMO, function with `...` is simple enough, @gleon99 ?

src/uct/base/uct_iface.h
yosefe: thanks, in general the call should ignore whatever was in info_str_buf before and overwrite it

amastbaum: This is right as long as info_str_buf is not called more than once in nested calls, so if we go this way we need to make sure of it in every flow

src/ucp/wireup/select.c
amastbaum: I wanted to reset it somewhere to prevent printing trash at the end. It could happen if all the transports fail to wireup in early stages before any is_reachable function is called (for example if `uct_iface_is_reachable_params_addrs_valid` fails and we don't print anything to info_str). I think I can remove this part and maybe just initialize the string when declaring it (here actually https://github.com/openucx/ucx/pull/9995#discussion_r1683132250)

src/uct/ib/base/ib_iface.c
amastbaum: No.. unless we want to provide something to uct users..?

src/uct/ib/base/ib_iface.c
amastbaum: You're right

src/ucp/wireup/select.c
yosefe: we can initialize it to empty string in uct_iface_is_reachable_v2() and then transports could (possibly) override

src/uct/ib/base/ib_iface.c
yosefe: so let's remove line 684 `ucs_error("failed to detect RoCE address size");`

src/uct/ib/rc/accel/rc_mlx5_iface.c
amastbaum: 2. I guess only with using a map such `static const tm_type_to_str[2] = {"basic", "tag matching"};` but it requires protecting it before accessing etc. What do you think?

amastbaum: I can at least reduce it to one comparison. I guess it may make it more efficient, this way
```
(my_type == UCT_RC_MLX5_IFACE_ADDR_TYPE_TM) ?
                    "tag matching" : "basic",
(my_type == UCT_RC_MLX5_IFACE_ADDR_TYPE_TM) ?
                    "basic" : "tag matching");
```

yosefe: maybe it can go without protection, since the value has to be 0 or 1 in a way that cannot be affected by user input?

src/uct/base/uct_iface.c
```diff
@@ -242,6 +242,26 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
...
+void uct_iface_fill_info_str_buf(const uct_iface_is_reachable_params_t *params,
+                                 const char *fmt, ...)
+{
+    char *info_str_buf      = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                              params, info_string, INFO_STRING,
+                                              NULL);
+    size_t info_str_buf_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                              params, info_string_length,
+                                              INFO_STRING_LENGTH, 0);
+
```
yosefe, line 254: pls remove space line

src/uct/base/uct_iface.c
```diff
@@ -242,6 +242,26 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
...
+    va_list ap;
+
+    va_start(ap, fmt);
+    if ((info_str_buf != NULL) && (info_str_buf[0] == '\0')) {
+        ucs_vsnprintf_safe(info_str_buf, info_str_buf_len, fmt, ap);
+    }
+
```
yosefe, line 261: pls remove space line

src/uct/base/uct_iface.c
```diff
@@ -242,6 +242,26 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
...
+                                              params, info_string, INFO_STRING,
+                                              NULL);
+    size_t info_str_buf_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                              params, info_string_length,
+                                              INFO_STRING_LENGTH, 0);
+
+    va_list ap;
+
+    va_start(ap, fmt);
+    if ((info_str_buf != NULL) && (info_str_buf[0] == '\0')) {
```
yosefe, line 258: why need to check info_str_buf[0] == '\0'?

src/uct/base/uct_iface.c
```diff
@@ -290,6 +310,10 @@ int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
 {
     const uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
 
+    if (params->info_string != NULL) {
```
yosefe, line 313: use UCS_PARAM_VALUE and do it only if info_str_len>0

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
```diff
@@ -31,6 +31,7 @@ enum {
      * compatibility */
 };
 
+static const char *tm_type_to_str[2] = {"basic", "tag matching"};
```
yosefe, line 34: pls move it inside uct_rc_mlx5_iface_is_reachable_v2 function
and can remove "2" and just use "[]"

src/uct/base/uct_iface.c
amastbaum: Oh, I mistakenly put it back.

src/uct/base/uct_iface.c
amastbaum: without checking that it's not NULL?

yosefe: in addition to checking for NULL

src/uct/base/uct_iface.c
```diff
@@ -309,9 +307,13 @@ int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
...
     const uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
-
-    if (params->info_string != NULL) {
-        params->info_string[0] = '\0';
+    char *info_str_buf = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params,
+                                         info_string, INFO_STRING, NULL);
+    size_t info_str_buf_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                              params, info_string_length,
+                                              INFO_STRING_LENGTH, 0);
+    if (info_str_buf && (info_str_buf_len > 0)) {
```
yosefe, line 315: space line before
also:
`if ((info_str_buf != NULL) && (info_str_buf_len > 0)) {`

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
```diff
@@ -647,6 +646,7 @@ uct_rc_mlx5_iface_is_reachable_v2(const uct_iface_h tl_iface,
     uint8_t my_type = uct_rc_mlx5_iface_get_address_type(tl_iface);
     uint8_t remote_type;
     const uct_iface_addr_t *iface_addr;
+    static const char *tm_type_to_str[] = {"basic", "tag matching"};
```
yosefe, line 649: pls move it before my_type since initialized vars should be before uninitialized ones

src/ucp/wireup/wireup.c
```diff
@@ -1243,17 +1243,25 @@ static void ucp_wireup_print_config(ucp_worker_h worker,
...
     };
 
+    if (info_str != NULL) {
+        params.field_mask |= UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING |
+                             UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING_LENGTH;
+        params.info_string = info_str;
```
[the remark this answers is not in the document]
gleon99: Unresolved

src/uct/ib/base/ib_iface.c
```diff
@@ -794,19 +845,28 @@ int uct_ib_iface_is_reachable_v2(const uct_iface_h tl_iface,
             UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params, device_addr,
                             DEVICE_ADDR, NULL);
     if (device_addr == NULL) {
+        uct_iface_fill_info_str_buf(params, "invalid IB device address");
         return 0;
     }
 
-    if (!uct_ib_iface_dev_addr_is_reachable(iface, device_addr)) {
+    if (!uct_ib_iface_dev_addr_is_reachable(iface, device_addr, params)) {
+        uct_iface_fill_info_str_buf(params, "unreachable ib device address");
```
gleon99, line 853: `ib` -> `IB`

src/uct/base/uct_iface.c
```diff
@@ -289,6 +307,15 @@ int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
                               const uct_iface_is_reachable_params_t *params)
 {
     const uct_base_iface_t *iface = ucs_derived_of(tl_iface, uct_base_iface_t);
+    char *info_str_buf = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params,
```
gleon99, line 310: Align `=`

src/uct/base/uct_iface.c
```diff
@@ -242,6 +242,24 @@ static int uct_iface_is_same_device(const uct_iface_h iface,
...
+{
+    char *info_str_buf      = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                              params, info_string, INFO_STRING,
+                                              NULL);
+    size_t info_str_buf_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD,
+                                              params, info_string_length,
+                                              INFO_STRING_LENGTH, 0);
+    va_list ap;
+
+    va_start(ap, fmt);
```
gleon99, line 256: Maybe instead:
```
if (buf == NULL) {
    return;
}

va_start(..);
..
```

wdyt?

amastbaum: Yes, it's a good idea

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
```diff
@@ -650,9 +650,10 @@ uct_rc_mlx5_iface_is_reachable_v2(const uct_iface_h tl_iface,
 
     iface_addr = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params,
                                  iface_addr, IFACE_ADDR, NULL);
-    remote_type = *(uint8_t*)iface_addr;
+
     /* Check hardware tag matching compatibility */
-    if ((iface_addr != NULL) && (my_type != remote_type)) {
+    if ((iface_addr != NULL) &&
+        (remote_type = *(uint8_t*)iface_addr != my_type)) {
```
amastbaum, line 656: @gleon99 if it's too messy I can just dereference twice..

src/ucp/wireup/select.c
```diff
@@ -2509,10 +2508,12 @@ ucp_wireup_select_lanes(ucp_ep_h ep, unsigned ep_init_flags,
 {
     ucp_worker_h worker                = ep->worker;
     ucp_tl_bitmap_t scalable_tl_bitmap = worker->scalable_tl_bitmap;
+    /* TO DO: remove initializtion after all ucp_wireup_add_X_lanes functions
+       will support specifying a reason */
+    char wireup_info[256]              = {0};
```
yosefe, line 2513: do we really need to initialize, given that uct_iface_is_reachable_v2() already initializes it to NULL?

amastbaum: Currently we do, because `ucp_wireup_add_am_lane` is the only one (out of all `ucp_wireup_add_*_lane` functions) that supports filling info_str. All the rest don't receive info_str and don't initialize it (until I'll add the other transports). So if we don't initialize it and one of them fails before `ucp_wireup_add_am_lane` is called, `ucp_wireup_search_lanes` returns a string filled with trash.

yosefe (approved):

src/ucp/wireup/select.c
yosefe: ok
do you plan to add support for other ucp_wireup_add_*_lane in next pr?

amastbaum: @yosefe yes.

yosefe:
@amastbaum pls see the merge conflicts

yosefe:
ASAN failure seems relevant:
```
[ RUN      ] udx/test_ucp_wireup_errh_peer.msg_before_ep_create/9 <ud_x/tag,unified,addr_v2>
=================================================================
==43336==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xffff7014d87d at pc 0xffff977e1c54 bp 0xfffff33b14d0 sp 0xfffff33b1528
READ of size 16 at 0xffff7014d87d thread T0
    #0 0xffff977e1c50 in __interceptor_inet_ntop ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:2517
    #1 0xffff96f2d734 in uct_ib_iface_roce_is_reachable /scrap/azure/agent-02/AZP_WORKSPACE/1/s/contrib/../src/uct/ib/base/ib_iface.c:752
    #2 0xffff96f2f890 in uct_ib_iface_dev_addr_is_reachable /scrap/azure/agent-02/AZP_WORKSPACE/1/s/contrib/../src/uct/ib/base/ib_iface.c:861
    #3 0xffff96f2f890 in uct_ib_iface_is_reachable_v2 /scrap/azure/agent-02/AZP_WORKSPACE/1/s/contrib/../src/uct/ib/base/ib_iface.c:896
```

src/uct/ib/base/ib_iface.c
```diff
@@ -822,19 +879,28 @@ int uct_ib_iface_is_reachable_v2(const uct_iface_h tl_iface,
...
-           uct_ib_iface_is_same_device(device_addr,
-                                       uct_ib_iface_port_attr(iface)->lid,
-                                       &iface->gid_info.gid);
+    if ((scope == UCT_IFACE_REACHABILITY_SCOPE_DEVICE) &&
+        !uct_ib_iface_is_same_device(device_addr,
+                                     uct_ib_iface_port_attr(iface)->lid,
+                                     &iface->gid_info.gid)) {
+        uct_iface_fill_info_str_buf(
+                params, "same device is expected in device reachability scope");
+
```
yosefe, line 899: remove space line

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
```diff
@@ -643,13 +643,22 @@ static int
...
+
     /* Check hardware tag matching compatibility */
-    if ((iface_addr != NULL) && (my_type != *(uint8_t*)iface_addr)) {
+    if ((iface_addr != NULL) &&
+        ((remote_type = *(uint8_t*)iface_addr) != my_type)) {
+        uct_iface_fill_info_str_buf(
+                    params, "incompatible hardware tag matching. "
+                    "%s (local) vs %s (remote)",
+                    tm_type_to_str[my_type], tm_type_to_str[remote_type]);
+
```
yosefe, line 661: remove space line

yosefe (approved):

amastbaum:
@yosefe I accidentally merged 'master' with my branch before squashing previous commits (when I had to solve some merge conflicts) so there are still some old commits left in history

yosefe:
> @yosefe I accidentally merged 'master' with my branch before squashing previous commits (when I had to solve some merge conflicts) so there are still some old commits left in history

It's ok to merge master branch, currently need to rebase on master branch - see https://github.com/openucx/ucx/wiki/Guidance-for-contributors#squashing-a-pr-with-merge-commits-in-history

PR #9992, opened 2024-07-04
UCS/TOPO: Set higher bandwidth for different NUMA nodes on Grace

tvegas1:
## What
Bandwidth on different NUMA nodes is almost identical to the ones on same NUMA nodes on Grace.

Issue: [3972096](https://redmine.mellanox.com/issues/3972096)

## Why ?
With GPU0 on NUMA 0, mlx5_0/mlx5_3/mlx5_7/mlx5_9 on NUMA 0/1/2/3. Without that patch, mlx5_0 is seen so fast that other interfaces are not used for D2H transfers (RNDV get), and bandwidth is capped at <50GBs (400Gbps CX7).

Adding huge `UCX_MULTI_LANE_MAX_RATIO` leads to using many interfaces but with traffic still going through mlx5_0:1 for >90% of it.

PR #9996, opened 2024-07-08
UCS/TOPO: Set higher bandwidth for different NUMA nodes on Grace - v1.17.x

tvegas1:
## What
backport #9992


tvegas1:
workaround in case we want to skip backport:
- `-x UCX_IB_PCI_BW=mlx5_0:1Gbps,mlx5_3:1Gbps,mlx5_7:1Gbps,mlx5_9:1Gbps -x UCX_RNDV_THRESH=1k`

yosefe:
@brminich do we need to port https://github.com/openucx/ucx/pull/9962 to v1.17.x?

brminich:
> @brminich do we need to port #9962 to v1.17.x?

let me check why it is failing on 1.17

brminich:
we also need to analyze a regressions associated with this change

PR #9999, opened 2024-07-09
UCP/GTEST: Fix rtr_mtype abort()

brminich:
## What
- Fix abort() function for rtr_mtype (need to set status in the super request, so it is propogated to receive request)
- Add gtest for testing error handling with ppln protocol

### DIFF
--- a/src/ucp/rndv/rndv_rtr.c
+++ b/src/ucp/rndv/rndv_rtr.c
@@ -277,11 +277,11 @@ static size_t ucp_proto_rndv_rtr_mtype_pack(void *dest, void *arg)
     return sizeof(*rtr) + packed_rkey_size;
 }
 
-static void ucp_proto_rndv_rtr_mtype_complete(ucp_request_t *req)
+static void ucp_proto_rndv_rtr_mtype_complete(ucp_request_t *req, int abort)
 {
     ucs_mpool_put_inline(req->send.rndv.mdesc);
     if (ucp_proto_rndv_request_is_ppln_frag(req)) {
-        ucp_proto_rndv_ppln_recv_frag_complete(req, 0, 0);
+        ucp_proto_rndv_ppln_recv_frag_complete(req, 0, abort);
     } else {
         ucp_proto_rndv_rtr_common_complete(req, UCS_BIT(UCP_DATATYPE_CONTIG));
     }
@@ -290,15 +290,13 @@ static void ucp_proto_rndv_rtr_mtype_complete(ucp_request_t *req)
 static void
 ucp_proto_rndv_rtr_mtype_abort(ucp_request_t *req, ucs_status_t status)
 {
-    req->status = status;
-    ucp_send_request_id_release(req);
-    ucs_mpool_put_inline(req->send.rndv.mdesc);
+    ucp_request_t *super_req = ucp_request_get_super(req);
 
-    if (ucp_proto_rndv_request_is_ppln_frag(req)) {
-        ucp_proto_rndv_ppln_recv_frag_complete(req, 0, 1);
-    } else {
-        ucp_proto_rndv_rtr_common_complete(req, UCS_BIT(UCP_DATATYPE_CONTIG));
-    }
+    super_req->status = status;
+
+    /*TODO: Invalidate memh */
+    ucp_send_request_id_release(req);
+    ucp_proto_rndv_rtr_mtype_complete(req, 1);
 }
 
 static ucs_status_t ucp_proto_rndv_rtr_mtype_reset(ucp_request_t *req)
@@ -315,7 +313,7 @@ static void ucp_proto_rndv_rtr_mtype_copy_completion(uct_completion_t *uct_comp)
 {
     ucp_request_t *req = ucs_container_of(uct_comp, ucp_request_t,
                                           send.state.uct_comp);
-    ucp_proto_rndv_rtr_mtype_complete(req);
+    ucp_proto_rndv_rtr_mtype_complete(req, 0);
 }
 
 static void
@@ -325,7 +323,7 @@ ucp_proto_rndv_rtr_mtype_data_received(ucp_request_t *req, int in_buffer)
     if (in_buffer) {
         /* Data was already placed in user buffer because the sender responded
            with RNDV_DATA packets */
-        ucp_proto_rndv_rtr_mtype_complete(req);
+        ucp_proto_rndv_rtr_mtype_complete(req, 0);
     } else {
         /* Data was not placed in user buffer, which means it was placed to
            the remote address we published - the rendezvous fragment */
--- a/test/gtest/ucp/test_ucp_peer_failure.cc
+++ b/test/gtest/ucp/test_ucp_peer_failure.cc
@@ -27,7 +27,9 @@ public:
     ucp_ep_params_t get_ep_params();
 
 protected:
-    static const int AM_ID = 0;
+    static const int AM_ID        = 0;
+    static const uint64_t TX_SEED = 0x1111111111111111lu;
+    static const uint64_t RX_SEED = 0x2222222222222222lu;
 
     enum {
         TEST_AM  = UCS_BIT(0),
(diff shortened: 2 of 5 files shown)


test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -75,14 +77,17 @@ class test_ucp_peer_failure : public ucp_test {
...
-    ucs::handle<ucp_rkey_h> m_stable_rkey, m_failing_rkey;
+    void init_buffers(size_t msg_size);
+
+    void                        *m_sreq, *m_rreq;
+    std::queue<void *>          m_am_rndv_descs;
+    size_t                      m_am_rx_count;
+    size_t                      m_err_count;
+    ucs_memory_type_t           m_mem_type;
+    ucs_status_t                m_err_status;
+    std::unique_ptr<mem_buffer> m_sbuf, m_rbuf;
```
ivankochin, line 88: Why needed to store `unique_ptr` to `mem_buffer` as I can see buffer is structure that already contains pointer to allocated memory. If that is required for `reset()` functionality, I think it is better to realize `mem_buffer::reset()` method for that purpose.

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -830,40 +843,68 @@ class test_ucp_peer_failure_rndv_abort : public test_ucp_peer_failure {
...
 
 UCP_INSTANTIATE_TEST_CASE(test_ucp_peer_failure_rndv_abort)
+
+
+class test_ucp_peer_failure_rndv_ppln_abort :
+      public test_ucp_peer_failure_rndv_abort {
+public:
+    static void get_test_variants(variant_vec_t &variants)
+    {
+        if (!mem_buffer::is_gpu_supported()) {
```
ivankochin, line 884: What does it mean to return nothing from `get_test_variants()`? Does it mean to skip this test? Maybe better to move it to init and add `_SKIP_TEST()` then?

test/gtest/ucp/ucp_test.cc
```diff
@@ -1182,6 +1182,30 @@ bool ucp_test_base::entity::has_lane_with_caps(uint64_t caps) const
...
+    for (auto i = 0; i < config->key.num_lanes; ++i) {
+        const auto lane = config->key.rma_bw_lanes[i];
+        if (lane == UCP_NULL_LANE) {
+            break;
+        }
+        if (ucp_ep_md_attr(ep, lane)->reg_mem_types & UCS_BIT(mem_type)) {
+            const auto iface_attr =
+                ucp_worker_iface_get_attr(this->worker(),
+                        ucp_ep_get_rsc_index(ep, lane));
+            if (iface_attr->cap.flags & UCT_IFACE_FLAG_PUT_ZCOPY) {
```
ivankochin, line 1200: Is `UCT_IFACE_FLAG_PUT_ZCOPY` only possible option? As I know RNDV through AM can work with PPLN too.

brminich: the aim is to check ppln with put using specific fragments memory type

test/gtest/ucp/test_ucp_peer_failure.cc
brminich: we do not initialize the test at all, imo it is less intrusive. Also we already do the same in UCP AM eager and rndv memtype tests

test/gtest/ucp/test_ucp_peer_failure.cc
brminich: imo it is more convinient

src/ucp/rndv/rndv_rtr.c
```diff
@@ -277,11 +277,14 @@ static size_t ucp_proto_rndv_rtr_mtype_pack(void *dest, void *arg)
     return sizeof(*rtr) + packed_rkey_size;
 }
 
-static void ucp_proto_rndv_rtr_mtype_complete(ucp_request_t *req)
+static UCS_F_ALWAYS_INLINE void
+ucp_proto_rndv_rtr_mtype_complete(ucp_request_t *req, int abort)
 {
-    ucs_mpool_put_inline(req->send.rndv.mdesc);
+    if (!abort || (req->send.rndv.mdesc != NULL)) {
```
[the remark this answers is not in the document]
brminich: it's a tiny optimization, because abort is passed as a constant, so in normal flow this branch is supposed to be eliminated

test/gtest/ucp/ucp_test.cc
ivankochin: Then maybe worth renaiming to `is_rndv_ppln_put_supported` (and rename tests accordingly)

brminich: renamed

ivankochin (dismissed):

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -95,9 +100,11 @@ test_ucp_peer_failure::test_ucp_peer_failure() :
     m_rreq(nullptr),
     m_am_rx_count(0),
     m_err_count(0),
+    m_mem_type(UCS_MEMORY_TYPE_HOST),
     m_err_status(UCS_OK)
 {
-    ucs::fill_random(m_sbuf);
+    m_sbuf.reset();
+    m_rbuf.reset();
```
yosefe, line 107: why needed?

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -830,40 +843,68 @@ class test_ucp_peer_failure_rndv_abort : public test_ucp_peer_failure {
...
+            return;
+        }
+
+        test_ucp_peer_failure_rndv_abort::get_test_variants(variants);
+    }
+
+    void init() override
+    {
+        modify_config("RNDV_SCHEME", "put_ppln");
+        modify_config("RNDV_PIPELINE_SHM_ENABLE", "n");
```
yosefe, line 894: why? pls add comment

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -830,40 +843,68 @@ class test_ucp_peer_failure_rndv_abort : public test_ucp_peer_failure {
...
+        m_mem_type = UCS_MEMORY_TYPE_CUDA;
```
yosefe, line 895: why only cuda?

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -830,40 +843,68 @@ class test_ucp_peer_failure_rndv_abort : public test_ucp_peer_failure {
...
+        test_ucp_peer_failure_rndv_abort::init();
+    }
+};
+
+UCS_TEST_P(test_ucp_peer_failure_rndv_put_ppln_abort, rtr_mtype)
+{
+    if (!sender().is_rndv_put_ppln_supported()) {
+        UCS_TEST_SKIP_R("RNDV pipeline is not supported");
+    }
```
yosefe, line 904: maybe do it during init() ?

test/gtest/ucp/ucp_test.cc
```diff
@@ -1182,6 +1182,30 @@ bool ucp_test_base::entity::has_lane_with_caps(uint64_t caps) const
     return false;
 }
 
+bool ucp_test_base::entity::is_rndv_put_ppln_supported() const
+{
+    const auto ep              = this->ep();
```
yosefe, line 1187: do we really need this var?

test/gtest/ucp/ucp_test.cc
```diff
@@ -1182,6 +1182,30 @@ bool ucp_test_base::entity::has_lane_with_caps(uint64_t caps) const
...
+    const auto config          = ucp_ep_config(ep);
+    ucs_memory_type_t mem_type = this->ucph()->config.ext.rndv_frag_mem_type;
```
yosefe, line 1189: "this->" is probably not needed here

test/gtest/ucp/test_ucp_peer_failure.cc
brminich: it's just to test a bug in abort flow, do you suggest to also add `cuda-managed`?

brminich:
@yosefe, ok to squash?

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -75,14 +77,17 @@ class test_ucp_peer_failure : public ucp_test {
...
-    std::string             m_sbuf, m_rbuf;
-    mem_handle_t            m_stable_memh, m_failing_memh;
-    ucs::handle<ucp_rkey_h> m_stable_rkey, m_failing_rkey;
+    void init_buffers(size_t msg_size);
+
+    void                        *m_sreq, *m_rreq;
+    std::queue<void *>          m_am_rndv_descs;
+    size_t                      m_am_rx_count;
+    size_t                      m_err_count;
+    ucs_memory_type_t           m_mem_type;
```
yosefe, line 86: can it be a virtual function ? since it does not change

yosefe (approved):

PR #10000, opened 2024-07-09
AZP: Multiple wire-compat test fixes

brminich:
## What
Some fixes in wire-compat testing:
- Use tag API in client-server example to make sure that RMA BW lanes are established
- Use `UCX_TCP_BRIDGE_ENABLE=y` to not fail `check_tls_perf_caps.sh`, because bridge devices are not shown in `ucx_info -d` since #9487
- Do not try to connect to different subnet addresses with RoCE

### DIFF
--- a/buildlib/tools/check_tls_perf_caps.sh
+++ b/buildlib/tools/check_tls_perf_caps.sh
@@ -15,6 +15,8 @@ LD_LIBRARY_PATH=${UCX_PR_PATH}/lib "${UCX_PR_PATH}"/bin/ucx_info -v
 LD_LIBRARY_PATH=${UCX_LEGACY_PATH}/lib old_out=$("${UCX_LEGACY_PATH}"/bin/ucx_info -d)
 LD_LIBRARY_PATH=${UCX_PR_PATH}/lib new_out=$("${UCX_PR_PATH}"/bin/ucx_info -d)
 
+export UCX_TCP_BRIDGE_ENABLE=y #bridge devices are not shown by default since v1.16
+
 res=true
 for tl_name in $(echo "${old_out}" | grep Transport | awk '{print $3}')
 do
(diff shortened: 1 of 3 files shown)


buildlib/tools/check_tls_perf_caps.sh
```diff
@@ -15,6 +15,8 @@ LD_LIBRARY_PATH=${UCX_PR_PATH}/lib "${UCX_PR_PATH}"/bin/ucx_info -v
 LD_LIBRARY_PATH=${UCX_LEGACY_PATH}/lib old_out=$("${UCX_LEGACY_PATH}"/bin/ucx_info -d)
 LD_LIBRARY_PATH=${UCX_PR_PATH}/lib new_out=$("${UCX_PR_PATH}"/bin/ucx_info -d)
 
+export UCX_TCP_BRIDGE_ENABLE=y #bridge devices are not shown by default since v1.16
```
yosefe, line 18: why would we want to run on bridge device?

yosefe: it can try running on ephemeral docker network devices and result in failures

brminich: to be able to compare perf settings with UCX v1.15 (there is no such var there and default behavior is yes)

brminich: we do not run anything here, just execute `ucx_info -d`, so I think it is safe

yosefe: ok

yosefe (dismissed):

yosefe (dismissed):

yosefe:
> AZP: do not test clietn-server with v1.14

Why?

brminich:
> > AZP: do not test clietn-server with v1.14
> 
> Why?

it's broken with tag api  (fails ci)

brminich:
squashed, because PR is small

yosefe (approved):

PR #10003, opened 2024-07-09
PERF: add fence to UCT/STREAM_UNI test before sentinel ack

evgeny-leksikov:
## What
add fence to UCT/STREAM_UNI test before sentinel ack

## Why ?
if TL does not guarantee put-after-put ordering then test may hang due to wrong value in recv buffer

### DIFF
--- a/src/tools/perf/lib/uct_tests.cc
+++ b/src/tools/perf/lib/uct_tests.cc
@@ -377,6 +377,15 @@ public:
         };
     }
 
+    void fence(int peer_index)
+    {
+        ucs_status_t status = uct_ep_fence(m_perf.uct.peers[peer_index].ep, 0);
+
+        if (status != UCS_OK) {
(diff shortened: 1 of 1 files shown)


src/tools/perf/lib/libperf.c
```diff
@@ -336,6 +336,15 @@ static ucs_status_t ucx_perf_test_check_params(ucx_perf_params_t *params)
     return UCS_OK;
 }
 
+void uct_perf_ep_fence(ucx_perf_context_t *perf, int peer_index)
+{
+    uct_ep_h ep = perf->uct.peers[peer_index].ep;
```
yosefe, line 341: maybe remove ep var?
moreover, I'm not sure we need this wrapper, it's used only once, can call uct_ep_fence directly from uct_perf_test_runner::fence

yosefe (approved):

PR #10005, opened 2024-07-10
UCT/IB/DC: apply append_safe mechanism to dcis array, added ucs_array_resize_safe for similar reasons

roiedanino:
## What
Used `ucs_array_append_safe` when adding a new DCI, and `ucs_array_resize_safe` in random policy

## Why ?
Applying changes from #9930 to `dcis` array and avoiding the reservation of the entire array (a fix that was used as a workaround to `ucs_array_t` copy issue)

### DIFF
--- a/src/ucs/datastruct/list.h
+++ b/src/ucs/datastruct/list.h
@@ -7,6 +7,7 @@
 #ifndef UCS_LIST_H_
 #define UCS_LIST_H_
 
+#include <ucs/debug/debug.h>
 #include <ucs/sys/compiler_def.h>
 
 
@@ -15,6 +16,8 @@ BEGIN_C_DECLS
 #define UCS_LIST_INITIALIZER(_prev, _next) \
     { (_prev), (_next) }
 
+#define UCS_LIST_CHECK_ELEM_IS_VALID(_elem) \
+    UCS_ASAN_ADDRESS_IS_VALID(_elem, sizeof(ucs_list_link_t));
 
 /**
  * Declare an empty list
@@ -111,6 +114,7 @@ static inline void ucs_list_del(ucs_list_link_t *elem)
  */
 static inline int ucs_list_is_empty(const ucs_list_link_t *head)
 {
+    UCS_LIST_CHECK_ELEM_IS_VALID(head->next);
     return head->next == head;
 }
 
@@ -120,6 +124,8 @@ static inline int ucs_list_is_empty(const ucs_list_link_t *head)
 static inline int
 ucs_list_is_first(const ucs_list_link_t *head, const ucs_list_link_t *elem)
 {
+    UCS_LIST_CHECK_ELEM_IS_VALID(head);
+    UCS_LIST_CHECK_ELEM_IS_VALID(elem->prev);
     return elem->prev == head;
 }
 
@@ -129,6 +135,8 @@ ucs_list_is_first(const ucs_list_link_t *head, const ucs_list_link_t *elem)
 static inline int
 ucs_list_is_last(const ucs_list_link_t *head, const ucs_list_link_t *elem)
 {
+    UCS_LIST_CHECK_ELEM_IS_VALID(head);
+    UCS_LIST_CHECK_ELEM_IS_VALID(elem->next);
     return elem->next == head;
 }
 
@@ -177,7 +185,9 @@ static inline unsigned long ucs_list_length(ucs_list_link_t *head)
     unsigned long length;
     ucs_list_link_t *ptr;
 
+    UCS_LIST_CHECK_ELEM_IS_VALID(head->next);
     for (ptr = head->next, length = 0; ptr != head; ptr = ptr->next, ++length);
+
     return length;
 }
 
@@ -215,6 +225,7 @@ static inline unsigned long ucs_list_length(ucs_list_link_t *head)
  * Iterate over members of the list.
  */
 #define ucs_list_for_each(_elem, _head, _member) \
+    UCS_LIST_CHECK_ELEM_IS_VALID(_head); \
     for (_elem = ucs_container_of((_head)->next, ucs_typeof(*_elem), _member); \
         &(_elem)->_member != (_head); \
         _elem = ucs_container_of((_elem)->_member.next, ucs_typeof(*_elem), _member))
@@ -223,6 +234,7 @@ static inline unsigned long ucs_list_length(ucs_list_link_t *head)
  * Iterate over members of the list, the user may invalidate the current entry.
  */
 #define ucs_list_for_each_safe(_elem, _telem, _head, _member) \
+    UCS_LIST_CHECK_ELEM_IS_VALID(_head); \
     for (_elem = ucs_container_of((_head)->next, ucs_typeof(*_elem), _member), \
         _telem = ucs_container_of(_elem->_member.next, ucs_typeof(*_elem), _member); \
         &_elem->_member != (_head); \
--- a/src/ucs/datastruct/queue.h
+++ b/src/ucs/datastruct/queue.h
@@ -10,7 +10,12 @@
 #include "queue_types.h"
 
 #include <ucs/debug/assert.h>
+#include <ucs/debug/debug.h>
 #include <stddef.h>
+
+
+#define UCS_QUEUE_CHECK_HEAD_IS_VALID(_head) \
+    UCS_ASAN_ADDRESS_IS_VALID((_head)->ptail, sizeof(*(_head)->ptail));
 
 
 /**
@@ -31,10 +36,10 @@ static inline void ucs_queue_head_init(ucs_queue_head_t *queue)
  */
 static inline size_t ucs_queue_length(ucs_queue_head_t *queue)
 {
+    size_t length = 0;
     ucs_queue_elem_t **pelem;
-    size_t length;
-
-    length = 0;
+    
+    UCS_QUEUE_CHECK_HEAD_IS_VALID(queue);
     for (pelem = &queue->head; pelem != queue->ptail; pelem = &(*pelem)->next) {
         ++length;
     }
@@ -47,6 +52,8 @@ static inline size_t ucs_queue_length(ucs_queue_head_t *queue)
 static inline int
 ucs_queue_is_tail(ucs_queue_head_t *queue, ucs_queue_elem_t *elem)
 {
+    UCS_QUEUE_CHECK_HEAD_IS_VALID(queue);
+    UCS_ASAN_ADDRESS_IS_VALID(elem, sizeof(*elem));
     return queue->ptail == &elem->next;
 }
 
@@ -55,6 +62,7 @@ ucs_queue_is_tail(ucs_queue_head_t *queue, ucs_queue_elem_t *elem)
  */
 static inline int ucs_queue_is_empty(const ucs_queue_head_t *queue)
 {
+    UCS_QUEUE_CHECK_HEAD_IS_VALID(queue);
     return queue->ptail == &queue->head;
 }
 
@@ -208,6 +216,7 @@ static inline void ucs_queue_splice(ucs_queue_head_t *queue,
  * @param member  Member inside 'elem' which is the queue link.
  */
 #define ucs_queue_for_each(elem, queue, member) \
+    UCS_QUEUE_CHECK_HEAD_IS_VALID(queue); \
     /* we set `ptail` field to queue address to not subtract NULL pointer */ \
     for (*(queue)->ptail = (ucs_queue_elem_t*)(void*)(queue), \
              elem = ucs_container_of((queue)->head, ucs_typeof(*elem), member); \
@@ -225,6 +234,7 @@ static inline void ucs_queue_splice(ucs_queue_head_t *queue,
  * @param member  Member inside 'elem' which is the queue link.
  */
 #define ucs_queue_for_each_safe(elem, iter, queue, member) \
+    UCS_QUEUE_CHECK_HEAD_IS_VALID(queue); \
     for (iter = &(queue)->head, \
          elem = ucs_container_of(*iter, ucs_typeof(*elem), member); \
           iter != (queue)->ptail; \
@@ -242,6 +252,7 @@ static inline void ucs_queue_splice(ucs_queue_head_t *queue,
  * TODO optimize
  */
 #define ucs_queue_for_each_extract(elem, queue, member, cond) \
+    UCS_QUEUE_CHECK_HEAD_IS_VALID(queue); \
     for (elem = ucs_container_of((queue)->head, ucs_typeof(*elem), member); \
          \
          !ucs_queue_is_empty(queue) && (cond) && ucs_queue_pull_non_empty(queue); \
@@ -265,6 +276,7 @@ static inline ucs_queue_iter_t ucs_queue_iter_next(ucs_queue_iter_t i)
 
 static inline int ucs_queue_iter_end(ucs_queue_head_t *q, ucs_queue_iter_t i)
 {
+    UCS_QUEUE_CHECK_HEAD_IS_VALID(q);
     return i == q->ptail;
 }
 
--- a/src/ucs/debug/debug.h
+++ b/src/ucs/debug/debug.h
@@ -9,6 +9,12 @@
 
 #include <ucs/sys/compiler_def.h>
 
+#ifdef __SANITIZE_ADDRESS__
+#include <sanitizer/asan_interface.h>
+#include <ucs/debug/assert.h>
+#endif
+
+
 BEGIN_C_DECLS
 
 /**
@@ -17,6 +23,15 @@ BEGIN_C_DECLS
  * @param signum   Signal number to disable handling.
  */
 void ucs_debug_disable_signal(int signum);
+
+
+#ifdef __SANITIZE_ADDRESS__
+#define UCS_ASAN_ADDRESS_IS_VALID(_ptr, _size) \
+    ucs_assertv(!__asan_region_is_poisoned((void*)(_ptr), _size), "%s: %p", \
+                #_ptr, (void*)(_ptr))
+#else
+#define UCS_ASAN_ADDRESS_IS_VALID(_ptr, _size)
+#endif
 
 END_C_DECLS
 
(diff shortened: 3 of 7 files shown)


src/uct/ib/dc/dc_mlx5.c
```diff
@@ -879,17 +886,14 @@ uct_dc_mlx5_iface_dcis_create(uct_dc_mlx5_iface_t *iface,
...
-    status = uct_dc_mlx5_iface_resize_and_fill_dcis(
-            iface, UCT_DC_MLX5_IFACE_MAX_DCI_POOLS * iface->tx.ndci);
+    status = uct_dc_mlx5_iface_resize_and_fill_dcis(iface, 1);
     if (status != UCS_OK) {
         return status;
     }
 
     ucs_array_length(&iface->tx.dcis) = 0;
 
+
```
roiedanino, line 896: remove

src/ucs/datastruct/list.h
```diff
@@ -111,6 +111,7 @@ static inline void ucs_list_del(ucs_list_link_t *elem)
  */
 static inline int ucs_list_is_empty(const ucs_list_link_t *head)
 {
+    UCS_ASAN_ADDRESS_IS_VALID(head, sizeof(ucs_list_link_t));
```
yosefe, line 114: IMO we also need it in:
- ucs_list_length
- ucs_list_for_each
- ucs_list_for_each_safe
- ucs_list_is_first
- ucs_list_is_last

src/ucs/debug/debug.h
```diff
@@ -18,6 +19,17 @@ BEGIN_C_DECLS
  */
 void ucs_debug_disable_signal(int signum);
 
+#ifdef __SANITIZE_ADDRESS__
+#define UCS_ASAN_ADDRESS_IS_VALID(_ptr, _size) \
+    do { \
+        char _buffer[_size]; \
+        memcpy(_buffer, _ptr, _size); \
```
yosefe, line 26: can we avoid actually doing memcpy?
how about checking __asan_region_is_poisoned for example?
BTW, are we sure some compiler will not optimize-out this memory copy?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -879,17 +886,14 @@ uct_dc_mlx5_iface_dcis_create(uct_dc_mlx5_iface_t *iface,
     uct_dc_dci_t *dci;
 
     ucs_array_init_dynamic(&iface->tx.dcis);
-    /* Reserving max size as a temporary solution due to pointer-copy issue in ucs_array
-       TODO: fix when ucs_array_t supports complex types 
-    */
-    status = uct_dc_mlx5_iface_resize_and_fill_dcis(
-            iface, UCT_DC_MLX5_IFACE_MAX_DCI_POOLS * iface->tx.ndci);
+    status = uct_dc_mlx5_iface_resize_and_fill_dcis(iface, 1);
```
yosefe, line 889: why do we need one DCI? can we not call this function at all?

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -900,6 +904,8 @@ uct_dc_mlx5_iface_dcis_create(uct_dc_mlx5_iface_t *iface,
     iface->tx.bb_max = dci->txwq.bb_max;
     uct_dc_mlx5_destroy_dci(iface, dci);
 
+    ucs_array_length(&iface->tx.dcis) = 0;
+
```
yosefe, line 908: remove space line

src/uct/ib/dc/dc_mlx5.c
```diff
@@ -900,6 +904,8 @@ uct_dc_mlx5_iface_dcis_create(uct_dc_mlx5_iface_t *iface,
     iface->tx.bb_max = dci->txwq.bb_max;
     uct_dc_mlx5_destroy_dci(iface, dci);
 
+    ucs_array_length(&iface->tx.dcis) = 0;
```
yosefe, line 907: why needed?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -382,6 +382,40 @@ uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
     return ucs_likely(pool->stack_top < iface->tx.ndci);
 }
 
+static UCS_F_ALWAYS_INLINE void
+uct_dc_mlx5_iface_dcis_array_copy(void *dst, void *src, size_t length)
+{
+    int i;
+    uct_dc_dci_t *src_dcis = (uct_dc_dci_t*)src;
+    uct_dc_dci_t *dst_dcis = (uct_dc_dci_t*)dst;
```
yosefe, line 390: initialized vars should be first

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -382,6 +382,40 @@ uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
     return ucs_likely(pool->stack_top < iface->tx.ndci);
 }
 
+static UCS_F_ALWAYS_INLINE void
+uct_dc_mlx5_iface_dcis_array_copy(void *dst, void *src, size_t length)
```
yosefe, line 386: can we make src a "const void*" and change the definitions accordingly if needed?

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -382,6 +382,40 @@ uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
...
+{
+    int i;
```
yosefe, line 388: size_t

test/gtest/uct/ib/test_dc.cc
```diff
@@ -500,8 +500,9 @@ class test_dc_flow_control : public test_rc_flow_control {
         uct_dc_mlx5_iface_t *iface      = ucs_derived_of(e->iface(),
                                                          uct_dc_mlx5_iface_t);
         uct_dc_mlx5_pool_stack_t *stack = &iface->tx.dci_pool[0].stack;
+        size_t dcis_length              = ucs_array_length(&iface->tx.dcis);
 
-        for (int i = 0; i < iface->tx.ndci; ++i) {
+        for (int i = 0; i < dcis_length; ++i) {
```
yosefe, line 505: can we use ucs_array_for_each? also below

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -382,6 +382,40 @@ uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
...
+    *ucs_array_append_safe(&iface->tx.dcis, &old_buffer_p,
+                           return UCS_ERR_NO_MEMORY) = empty_dci;
+
+    if (old_buffer_p) {
+        uct_dc_mlx5_iface_dcis_array_copy(iface->tx.dcis.buffer, old_buffer_p,
+                                          old_length);
+        ucs_array_buffer_free(old_buffer_p);
+    }
+    return UCS_OK;
+}
```
yosefe, line 418: maybe not add this function, it seems like a code duplication, and revert to calling uct_dc_mlx5_iface_resize_and_fill_dcis(iface, dci_index + 1);

src/uct/ib/dc/dc_mlx5.c
roiedanino: `uct_dc_mlx5_iface_create_dci` is relying on the fact that there is enough space in `dcis` array

src/uct/ib/dc/dc_mlx5.c
roiedanino: because dci 0 was destroyed and that place is effectively empty

roiedanino:
Waiting for #10017 to merge (it will fix compilation issues)

src/ucs/datastruct/list.h
```diff
@@ -129,6 +132,8 @@ ucs_list_is_first(const ucs_list_link_t *head, const ucs_list_link_t *elem)
 static inline int
 ucs_list_is_last(const ucs_list_link_t *head, const ucs_list_link_t *elem)
 {
+    UCS_ASAN_ADDRESS_IS_VALID(head, sizeof(ucs_list_link_t));
```
yosefe, line 135: maybe add helper macro UCS_LIST_CHECK_ELEM_IS_VALID
to avoid passing `sizeof(ucs_list_link_t)` every time

src/ucs/datastruct/list.h
```diff
@@ -223,6 +234,8 @@ static inline unsigned long ucs_list_length(ucs_list_link_t *head)
  * Iterate over members of the list, the user may invalidate the current entry.
  */
 #define ucs_list_for_each_safe(_elem, _telem, _head, _member) \
+    UCS_ASAN_ADDRESS_IS_VALID(_head, sizeof(ucs_list_link_t)); \
+    \
```
yosefe, line 238: remove space line

src/ucs/datastruct/list.h
```diff
@@ -177,7 +182,12 @@ static inline unsigned long ucs_list_length(ucs_list_link_t *head)
     unsigned long length;
     ucs_list_link_t *ptr;
 
-    for (ptr = head->next, length = 0; ptr != head; ptr = ptr->next, ++length);
+    UCS_ASAN_ADDRESS_IS_VALID(head, sizeof(ucs_list_link_t));
+
```
yosefe, line 186: remove space line

src/ucs/debug/debug.h
```diff
@@ -9,15 +9,32 @@
...
  * Disable signal handling in UCS for signal.
  * Previous signal handler is set.
  * @param signum   Signal number to disable handling.
  */
 void ucs_debug_disable_signal(int signum);
 
+
+#ifdef __SANITIZE_ADDRESS__
+#define UCS_ASAN_ADDRESS_IS_VALID(_ptr, _size) \
+    ucs_assert(!__asan_region_is_poisoned((void*)_ptr, _size))
```
yosefe, line 33: assertv to print the ptr
add () around _ptr usage

src/uct/ib/dc/dc_mlx5_ep.h
yosefe: I see, we could assign the queue head from the old buffer to a temp local var on the stack
BTW, do we need to add ASAN macro to queue.h ?

roiedanino: Wouldn't that still override the const modifier?
> BTW, do we need to add ASAN macro to queue.h ?

Yes we probably need

yosefe: > Wouldn't that still override the const modifier?

No because we don't really need to update the old buffer (make the old list empty). Using a local var as a copy of the head will allow the source buffer stay const because now only this local var will change

src/ucs/datastruct/list.h
```diff
@@ -8,13 +8,15 @@
 #define UCS_LIST_H_
 
 #include <ucs/sys/compiler_def.h>
-
+#include <ucs/debug/debug.h>
```
gleon99, line 11: includes order

src/uct/ib/dc/dc_mlx5_ep.h
```diff
@@ -382,6 +382,23 @@ uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
...
+{
+    const uct_dc_dci_t *src_dcis = (const uct_dc_dci_t*)src;
+    uct_dc_dci_t *dst_dcis       = (uct_dc_dci_t*)dst;
+    size_t i;
+
+    memcpy(dst_dcis, src_dcis, sizeof(uct_dc_dci_t) * length);
+
+    /* txqp is a queue and need to splice tail */
+    for (i = 0; i < length; ++i) {
+        ucs_queue_head_init(&dst_dcis[i].txqp.outstanding);
```
gleon99, line 396: remove line

roiedanino: What blank line?

src/uct/ib/dc/dc_mlx5_ep.h
roiedanino: gtest: `test_datatype, queue_splice` is expecting source length to be 0 after calling splice, can I just modify the test?

src/ucs/datastruct/list.h
```diff
@@ -121,6 +123,8 @@ static inline int ucs_list_is_empty(const ucs_list_link_t *head)
 static inline int
 ucs_list_is_first(const ucs_list_link_t *head, const ucs_list_link_t *elem)
 {
+    UCS_LIST_CHECK_ELEM_IS_VALID(head);
+    UCS_LIST_CHECK_ELEM_IS_VALID(elem);
```
yosefe, line 127: no need to check elem, we deref it
we need to check elem->prev
same comment for ucs_list_is_last

src/ucs/datastruct/list.h
```diff
@@ -111,7 +113,7 @@ static inline void ucs_list_del(ucs_list_link_t *elem)
  */
 static inline int ucs_list_is_empty(const ucs_list_link_t *head)
 {
-    UCS_ASAN_ADDRESS_IS_VALID(head, sizeof(ucs_list_link_t));
+    UCS_LIST_CHECK_ELEM_IS_VALID(head);
```
yosefe, line 116: we need to check head->next is a valid pointer..

src/ucs/datastruct/list.h
```diff
@@ -178,7 +184,11 @@ static inline unsigned long ucs_list_length(ucs_list_link_t *head)
     unsigned long length;
     ucs_list_link_t *ptr;
 
-    for (ptr = head->next, length = 0; ptr != head; ptr = ptr->next, ++length);
+    UCS_LIST_CHECK_ELEM_IS_VALID(head);
+    for (ptr = head->next, length = 0; ptr != head; ptr = ptr->next, ++length) {
+        UCS_LIST_CHECK_ELEM_IS_VALID(ptr);
```
yosefe, line 189: given the logic above, what is the right thing to check here?

src/ucs/datastruct/queue.h
```diff
@@ -155,12 +166,14 @@ static inline ucs_queue_elem_t *ucs_queue_pull(ucs_queue_head_t *queue)
...
 {
+    ucs_queue_head_t temp = *new_elems;
+
     if (!ucs_queue_is_empty(new_elems)) {
-        *queue->ptail = new_elems->head;
-        queue->ptail = new_elems->ptail;
-        new_elems->ptail = &new_elems->head;
+        *queue->ptail = temp.head;
+        queue->ptail = temp.ptail;
+        temp.ptail = &temp.head;
```
yosefe, line 176: we should not change splice function, but change its usage to introduce local var

test/gtest/ucs/test_datatype.cc
```diff
@@ -518,7 +518,8 @@ UCS_TEST_F(test_datatype, queue_splice) {
 
     ucs_queue_splice(&newq, &head);
 
-    EXPECT_EQ((unsigned long)0, ucs_queue_length(&head));
+    /* head should not be affected by the operation */
+    EXPECT_EQ((unsigned long)3, ucs_queue_length(&head));
```
yosefe, line 522: revert

src/ucs/datastruct/list.h
roiedanino: as ptr is always being dereferenced, maybe we don't need a check inside the loop

src/ucs/datastruct/queue.h
```diff
@@ -10,9 +10,16 @@
 #include "queue_types.h"
 
 #include <ucs/debug/assert.h>
+#include <ucs/debug/debug.h>
 #include <stddef.h>
 
 
+#define UCS_QUEUE_CHECK_HEAD_IS_VALID(_elem) \
+    UCS_ASAN_ADDRESS_IS_VALID(_elem, sizeof(ucs_queue_head_t));
```
yosefe, line 18: we should check only queue->ptail is valid pointer

src/ucs/datastruct/queue.h
```diff
@@ -10,9 +10,16 @@
...
+
+#define UCS_QUEUE_CHECK_ELEM_IS_VALID(_elem) \
+    UCS_ASAN_ADDRESS_IS_VALID(_elem, sizeof(ucs_queue_elem_t));
```
yosefe, line 21: this is probably not needed

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -383,10 +383,10 @@ uct_dc_mlx5_iface_dci_can_alloc(uct_dc_mlx5_iface_t *iface, uint8_t pool_index)
 }
 
 static UCS_F_ALWAYS_INLINE void
-uct_dc_mlx5_iface_dcis_array_copy(void *dst, const void *src, size_t length)
+uct_dc_mlx5_iface_dcis_array_copy(void *dst, void *src, size_t length)
 {
-    const uct_dc_dci_t *src_dcis = (const uct_dc_dci_t*)src;
-    uct_dc_dci_t *dst_dcis       = (uct_dc_dci_t*)dst;
+    uct_dc_dci_t *src_dcis = (uct_dc_dci_t*)src;
```
yosefe, line 388: 1. keep it const*. we ca do
```
ucs_queue_head_t tmp_q;
...
tmp_q = src_dcis[i].txqp.outstanding;
ucs_queue_splice(&dst_dcis[i].txqp.outstanding, &tmp_q);
```

src/ucs/debug/debug.h
```diff
@@ -9,15 +9,31 @@
...
 
+#ifdef __SANITIZE_ADDRESS__
+#include <sanitizer/asan_interface.h>
+#include <ucs/debug/assert.h>
+#endif
+
+
 BEGIN_C_DECLS
 
+
```
yosefe, line 20: why added?

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
roiedanino: That will cause a segfault as `outstanding` will point to `tmp_q`, but `tmp_q` will be dead at the end of this scope

yosefe: outstanding should not point to tmp_q after splice

src/ucs/debug/debug.h
roiedanino: Its reverting a deletion of that line

yosefe (approved):

PR #10007, opened 2024-07-11
UCT/CUDA_COPY: enable rdma flag for fabric allocations

Akshay-Venkatesh:
## Why ?
Allow cumemcreate memory allocations to be registered with IB.

### DIFF
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.c
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.c
@@ -207,6 +207,7 @@ uct_cuda_copy_mem_alloc_fabric(uct_cuda_copy_md_t *md,
     ucs_log_level_t log_level   = (md->config.enable_fabric == UCS_YES) ?
                                   UCS_LOG_LEVEL_ERROR : UCS_LOG_LEVEL_DEBUG;
     ucs_status_t status;
+    int allowed_types;
 
     if (!(flags & UCT_MD_MEM_FLAG_HIDE_ERRORS) &&
         (md->config.enable_fabric == UCS_YES)) {
@@ -215,10 +216,11 @@ uct_cuda_copy_mem_alloc_fabric(uct_cuda_copy_md_t *md,
         log_level = UCS_LOG_LEVEL_DEBUG;
     }
 
-    prop.type                 = CU_MEM_ALLOCATION_TYPE_PINNED;
-    prop.requestedHandleTypes = CU_MEM_HANDLE_TYPE_FABRIC;
-    prop.location.type        = CU_MEM_LOCATION_TYPE_DEVICE;
-    prop.location.id          = cu_device;
+    prop.type                            = CU_MEM_ALLOCATION_TYPE_PINNED;
+    prop.requestedHandleTypes            = CU_MEM_HANDLE_TYPE_FABRIC;
+    prop.location.type                   = CU_MEM_LOCATION_TYPE_DEVICE;
+    prop.location.id                     = cu_device;
+    prop.allocFlags.gpuDirectRDMACapable = 1;
 
     if (md->granularity == SIZE_MAX) {
         status = UCT_CUDADRV_FUNC(cuMemGetAllocationGranularity(
@@ -264,9 +266,19 @@ uct_cuda_copy_mem_alloc_fabric(uct_cuda_copy_md_t *md,
         goto err_mem_unmap;
     }
 
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuPointerGetAttribute(&allowed_types,
+                CU_POINTER_ATTRIBUTE_ALLOWED_HANDLE_TYPES, alloc_handle->ptr));
+    if (status != UCS_OK) {
+            goto err_mem_unmap;
+    } else if (!(allowed_types & CU_MEM_HANDLE_TYPE_FABRIC)) {
+        ucs_error("allocated memory at %p does not have fabric property",
+                  (void*)alloc_handle->ptr);
+        goto err_mem_unmap;
+    }
+
     alloc_handle->is_vmm = 1;
 
-    ucs_trace("allocated vmm fabric memory at %p of size %ld\n",
+    ucs_trace("allocated vmm fabric memory at %p of size %ld",
               (void*)alloc_handle->ptr, alloc_handle->length);
     return UCS_OK;
 


src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -221,10 +221,11 @@ uct_cuda_copy_mem_alloc_fabric(uct_cuda_copy_md_t *md,
...
 
-    prop.type                 = CU_MEM_ALLOCATION_TYPE_PINNED;
-    prop.requestedHandleTypes = CU_MEM_HANDLE_TYPE_FABRIC;
-    prop.location.type        = CU_MEM_LOCATION_TYPE_DEVICE;
-    prop.location.id          = cu_device;
+    prop.type                            = CU_MEM_ALLOCATION_TYPE_PINNED;
+    prop.requestedHandleTypes            = CU_MEM_HANDLE_TYPE_FABRIC;
+    prop.location.type                   = CU_MEM_LOCATION_TYPE_DEVICE;
+    prop.location.id                     = cu_device;
+    prop.allocFlags.gpuDirectRDMACapable = 1;
```
SeyedMir, line 228: Will the allocation fail on a system with no IB?

Akshay-Venkatesh: @SeyedMir 
No. `gpuDirectRDMACapable` is a request. CUDA isn't obliged to honor it and report errors if rdma enablement fails. Instead user is expected to check attribute `CU_POINTER_ATTRIBUTE_IS_GPU_DIRECT_RDMA_CAPABLE` once allocation succeeds to see if RDMA capability is enabled. 

But thanks for raising this as we do need to check that allocation has handle fabric allowed with it and we should return a failure if that isn't possible. I've made relevant changes now.

brminich (dismissed):

src/uct/cuda/cuda_copy/cuda_copy_md.c
yosefe: What about systems without RDMA (i.e TCP+shared mem only), maybe passing such flag could lead to failure?

yosefe: Maybe we should not fail if the allocation is not RDMA capable, and let it fail later if actually used with RDMA

brminich:
@yosefe pls review?

src/uct/cuda/cuda_copy/cuda_copy_md.c
yosefe: 1. does it mean users have to pass this flag if they allocate fabric memory and pass it to UCX/MPI?
2. why it doesn't fail on new systems?

Akshay-Venkatesh: 1. Indeed they are expected to pass a flag during allocation call
2. Under what scenarios do the tests currently fail? In general, passing a flag is a request to the allocator. The allocation doesn't fail if the request was made and the allocator wasn't able to honor the request. It is only at the point of trying to register that we are able to detect if the buffer is RDMA-capable.

Akshay-Venkatesh: > What about systems without RDMA (i.e TCP+shared mem only), maybe passing such flag could lead to failure?

It should not cause a failure.

> Maybe we should not fail if the allocation is not RDMA capable, and let it fail later if actually used with RDMA

We wouldn't fail the allocation and I agree that we should just fail at the point of registration.

yosefe: > 2\. Under what scenarios do the tests currently fail? In general, passing a flag is a request to the allocator. The allocation doesn't fail if the request was made and the allocator wasn't able to honor the request. It is only at the point of trying to register that we are able to detect if the buffer is RDMA-capable.

Well, today we don't pass RDMA flag to the allocation, and CI passes, does it mean we never tried to register VMM rdnv fragments with IB? Or the issue would be exposed on specific type of system/GPU/Cuda.. ?

Akshay-Venkatesh: This should show up with a VMM allocation on any platform. If we try to register device VMM memory and the flag is not set by caller, then IB registration call should return a failure.

ovidiusm:
Received confirmation that applying this PR fixes the issue seen on DGX V100 systems (ibv_reg_mr failing)

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -268,6 +270,12 @@ uct_cuda_copy_mem_alloc_fabric(uct_cuda_copy_md_t *md,
         goto err_mem_unmap;
     }
 
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuPointerGetAttribute(&allowed_types,
+                CU_POINTER_ATTRIBUTE_ALLOWED_HANDLE_TYPES, alloc_handle->ptr));
+    if ((status != UCS_OK) || !(allowed_types & CU_MEM_HANDLE_TYPE_FABRIC)) {
+        goto err_mem_unmap;
+    }
```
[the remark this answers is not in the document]
Akshay-Venkatesh: @yosefe addressed the second part. As we discussed offline, no need to check first part as we want ibv_reg_(dmabuf_)mr to fail instead.

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -273,12 +273,13 @@ uct_cuda_copy_mem_alloc_fabric(uct_cuda_copy_md_t *md,
     status = UCT_CUDADRV_FUNC_LOG_ERR(cuPointerGetAttribute(&allowed_types,
                 CU_POINTER_ATTRIBUTE_ALLOWED_HANDLE_TYPES, alloc_handle->ptr));
     if ((status != UCS_OK) || !(allowed_types & CU_MEM_HANDLE_TYPE_FABRIC)) {
+        ucs_debug("unable to allocate vmm memory with fabric property");
```
yosefe, line 276:
```c
if (status != UCS_OK) {
    goto err_mem_unmap;
} else if (!(allowed_types & CU_MEM_HANDLE_TYPE_FABRIC)) {
    ucs_error("allocated memory at %p does not have fabric property", alloc_handle->ptr);
    goto err_mem_unmap;
}
```

pentschev:
I confirmed this fixed an issue I've seen locally which reproduces on both 1.18.0 and `master`. The following reproduces the error on a DGX-1:

```
$ CUDA_VISIBLE_DEVICES=0,1 UCX_TLS=tcp,rc,cuda_copy ucx_perftest -m cuda -t tag_bw -s 1000000 & CUDA_VISIBLE_DEVICES=0,1 UCX_TLS=tcp,rc,cuda_copy ucx_perftest -m cuda -t tag_bw -s 1000000 localhost
[2] 271204
[1743605233.803141] [dgx13:271205:0]        perftest.c:800  UCX  WARN  CPU affinity is not set (bound to 80 cpus). Performance may be impacted.
[1743605233.803277] [dgx13:271204:0]        perftest.c:800  UCX  WARN  CPU affinity is not set (bound to 80 cpus). Performance may be impacted.
Waiting for connection...
Accepted connection from 127.0.0.1:41062
+----------------------------------------------------------------------------------------------------------+
+--------------+--------------+------------------------------+---------------------+-----------------------+
| API:          protocol layer                                                                             |
|              |              |       overhead (usec)        |   bandwidth (MB/s)  |  message rate (msg/s) |
| Test:         tag match bandwidth                                                                        |
+--------------+--------------+----------+---------+---------+----------+----------+-----------+-----------+
| Data layout:  (automatic)                                                                                |
| Send memory:  cuda                                                                                       |
|    Stage     | # iterations | 50.0%ile | average | overall |  average |  overall |  average  |  overall  |
| Recv memory:  cuda                                                                                       |
+--------------+--------------+----------+---------+---------+----------+----------+-----------+-----------+
| Message size: 1000000                                                                                    |
| Window size:  32                                                                                         |
+----------------------------------------------------------------------------------------------------------+
[1743605234.405892] [dgx13:271204:0]           ib_md.c:287  UCX  ERROR ibv_reg_mr(address=0x7fa88da00000, length=2097152, access=0xf) failed: Bad address
[1743605234.405909] [dgx13:271204:0]          ucp_mm.c:76   UCX  ERROR failed to register address 0x7fa88da00000 (cuda) length 2097152 on md[2]=mlx5_0: Input/output error (md supports: host|cuda)
[1743605234.406104] [dgx13:271204:0]         libperf.c:2064 UCX  WARN  ucp test failed to allocate memory
[1743605234.406218] [dgx13:271205:0]           ib_md.c:287  UCX  ERROR ibv_reg_mr(address=0x7f9303a00000, length=2097152, access=0xf) failed: Bad address
[1743605234.406235] [dgx13:271205:0]          ucp_mm.c:76   UCX  ERROR failed to register address 0x7f9303a00000 (cuda) length 2097152 on md[2]=mlx5_0: Input/output error (md supports: host|cuda)
[1743605234.406435] [dgx13:271205:0]         libperf.c:2064 UCX  WARN  ucp test failed to allocate memory
[1743605234.409474] [dgx13:271204:0]    perftest_run.c:345  UCX  ERROR Failed to run test: Out of memory
[1743605234.409641] [dgx13:271205:0]    perftest_run.c:345  UCX  ERROR Failed to run test: Out of memory
```

Setting `UCX_CUDA_COPY_ENABLE_FABRIC=no` with `master` gets rid of the issue above, using this PR without setting `UCX_CUDA_COPY_ENABLE_FABRIC` resolves it too. Can we ensure this makes goes in 1.19.0?

src/uct/cuda/cuda_copy/cuda_copy_md.c
```diff
@@ -264,9 +266,19 @@ uct_cuda_copy_mem_alloc_fabric(uct_cuda_copy_md_t *md,
...
     }
 
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuPointerGetAttribute(&allowed_types,
+                CU_POINTER_ATTRIBUTE_ALLOWED_HANDLE_TYPES, alloc_handle->ptr));
+    if (status != UCS_OK) {
+            goto err_mem_unmap;
+    } else if (!(allowed_types & CU_MEM_HANDLE_TYPE_FABRIC)) {
+        ucs_error("allocated memory at %p does not have fabric property",
+                   (void*)alloc_handle->ptr);
+        goto err_mem_unmap;
```
yosefe, line 276: pls fix code style (alignment)

yosefe:
replaced by #10616

PR #10012, opened 2024-07-14
UCT/KNEM/CUDA_IPC/SELF: Fixed uct_ep_is_connected

shasson5:
## What
Fixed uct_ep_is_connected API for 3 transports: knem, self, cuda_ipc.

## Why ?
These transports should support is_connected API in order for EP reconfiguration to work properly.

## How
1) The macro `UCT_EP_IS_CONNECTED_CHECK_DEV_IFACE_ADDRS` is removed from `uct_base_ep_is_connected`, in order to allow transports without dev_addr to use it (eg. knem).
2) In cuda_ipc/self -> updated `ep_is_connected` callback.

### DIFF
--- a/src/uct/base/uct_iface.c
+++ b/src/uct/base/uct_iface.c
@@ -310,9 +310,21 @@ int uct_base_iface_is_reachable(const uct_iface_h tl_iface,
 int uct_base_ep_is_connected(const uct_ep_h tl_ep,
                              const uct_ep_is_connected_params_t *params)
 {
-    UCT_EP_IS_CONNECTED_CHECK_DEV_IFACE_ADDRS(params);
-    return uct_base_iface_is_reachable(tl_ep->iface, params->device_addr,
-                                       params->iface_addr);
+    uct_iface_is_reachable_params_t is_reachable_params = {0};
+
+    if (params->field_mask & UCT_EP_IS_CONNECTED_FIELD_DEVICE_ADDR) {
+        is_reachable_params.field_mask |=
+                UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR;
+        is_reachable_params.device_addr = params->device_addr;
+    }
+
+    if (params->field_mask & UCT_EP_IS_CONNECTED_FIELD_IFACE_ADDR) {
+        is_reachable_params.field_mask |=
+                UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR;
+        is_reachable_params.iface_addr  = params->iface_addr;
+    }
+
+    return uct_iface_is_reachable_v2(tl_ep->iface, &is_reachable_params);
 }
 
 int uct_ep_is_connected(uct_ep_h ep, const uct_ep_is_connected_params_t *params)
--- a/src/uct/sm/scopy/cma/cma_iface.c
+++ b/src/uct/sm/scopy/cma/cma_iface.c
@@ -79,7 +79,8 @@ uct_cma_iface_is_reachable_v2(const uct_iface_h tl_iface,
     ucs_cma_iface_ext_device_addr_t *iface_addr;
     pid_t peer_pid;
 
-    if (!uct_iface_is_reachable_params_addrs_valid(params)) {
+    if (!uct_iface_is_reachable_params_addrs_valid(params) ||
+        (params->device_addr == NULL) || (params->iface_addr == NULL)) {
         return 0;
     }
 
--- a/test/gtest/uct/test_uct_ep.cc
+++ b/test/gtest/uct/test_uct_ep.cc
@@ -29,8 +29,14 @@ protected:
 
     void create_sender()
     {
-        m_sender = uct_test::create_entity(0);
-        m_entities.push_back(m_sender);
+        /* Self/Memtype transports must connect to the same ucp_worker */
+        if ((GetParam()->dev_type == UCT_DEVICE_TYPE_SELF) ||
+            (GetParam()->dev_type == UCT_DEVICE_TYPE_ACC)) {
+            m_sender = m_receiver;
+        } else {
+            m_sender = uct_test::create_entity(0);
+            m_entities.push_back(m_sender);
+        }
     }
 
     void connect()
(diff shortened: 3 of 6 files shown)


shasson5:
@yosefe @gleon99 please review

src/uct/base/uct_iface.c
```diff
@@ -310,7 +310,6 @@ int uct_base_iface_is_reachable(const uct_iface_h tl_iface,
 int uct_base_ep_is_connected(const uct_ep_h tl_ep,
                              const uct_ep_is_connected_params_t *params)
 {
-    UCT_EP_IS_CONNECTED_CHECK_DEV_IFACE_ADDRS(params);
```
tvegas1, line 313: since `params->device_addr` and `params->iface_addr` are still passed to `uct_base_iface_is_reachable()`, shall instead make sure that they are actively set (maybe to NULL for TLs where there is no address) by keeping `UCT_EP_IS_CONNECTED_CHECK_DEV_IFACE_ADDRS()` line 313?

shasson5: I cannot know in this function which addresses should be set because it's per TL.
And I don't think check for NULL for unset addresses are required by the UCT API (AFAIU, non-set addresses can br uninitialized pointers).

tvegas1: ok but the check is removed, so now supporting transports cannot check anymore if the parameter was actually provided. i guess this is fine for transport that do not have address, but do you think it could be worth keeping the check and expect transports without addr to pass NULL pointers and setting the flag?

shasson5: I think it's problematic. For example look at knem_iface.c line 84:
`.ep_is_connected       = uct_base_ep_is_connected`
This function is called from wireup.c line 1911:
`        return uct_ep_is_connected(ucp_wireup_get_tl_ep(uct_ep), &params);`
The second arg (`params`) is initialized in `ucp_wireup_fill_is_connected_params`, so theoretically we can pass NULL for non-provided addresses.
But UCT API can also be called from MPI and other apps, and unlike UCP we cannot demand that they will pass NULL pointers.

BTW, the checks for the validity of the pointers are done in is_reachable which is called by `uct_base_ep_is_connected`

tvegas1: I see. Is there a way to somehow keep the check for supporting transports (the one that mandate flags+params being set)?

Because `uct_base_ep_is_connected()` calls `uct_base_iface_is_reachable()` which potentially falsely adds `UCT_IFACE_IS_REACHABLE_FIELD_{IFACE,DEVICE_ADDR}` and pointers that might not have been present/valid in the first place, if this check is removed.

yosefe: IMO it's ok since we call uct_iface_is_reachable_v2 that also checks the input parameters, per transport

test/gtest/uct/test_uct_ep.cc
```diff
@@ -29,8 +29,13 @@ class test_uct_ep : public uct_test {
...
     {
-        m_sender = uct_test::create_entity(0);
-        m_entities.push_back(m_sender);
+        /* Self/Memtype transports must connect to the same ucp_worker */
+        if (has_transport("self") || has_gpu()) {
+            m_sender = m_receiver;
+        } else {
+            m_sender = uct_test::create_entity(0);
+            m_entities.push_back(m_sender);
+        }
```
yosefe, line 38: maybe check dev_type for SELF or ACC
like here:
https://github.com/openucx/ucx/blob/d34bcea701db54f257088594febd1a21d4c7aa95/test/gtest/uct/uct_p2p_test.cc#L36

test/gtest/uct/test_uct_ep.cc
```diff
@@ -135,49 +198,23 @@ UCS_TEST_SKIP_COND_P(test_uct_ep, disconnect_after_send,
...
+    auto e = create_entity(0);
+    m_entities.push_back(e);
 
-    if (iface_attr.cap.flags & UCT_IFACE_FLAG_CONNECT_TO_EP) {
-        ep_addr.resize(iface_attr.ep_addr_len);
-        ASSERT_UCS_OK(uct_ep_get_address(m_receiver->ep(0),
-                                         (uct_ep_addr_t*)ep_addr.data()));
-        params.ep_addr     = (uct_ep_addr_t*)ep_addr.data();
-        params.field_mask |= UCT_EP_IS_CONNECTED_FIELD_EP_ADDR;
+    if (has_cma() || has_cuda_ipc() || has_transport("knem")) {
```
yosefe, line 208: why does it depend on the transport? can you pls add comment?

test/gtest/uct/test_uct_ep.cc
```diff
@@ -240,4 +277,5 @@ UCS_TEST_SKIP_COND_P(test_uct_ep, destroy_entity_after_send,
     }
 }
 
-UCT_INSTANTIATE_NO_SELF_TEST_CASE(test_uct_ep)
+UCT_INSTANTIATE_TEST_CASE(test_uct_ep)
+UCT_INSTANTIATE_CUDA_IPC_TEST_CASE(test_uct_ep)
```
yosefe, line 281: maybe use UCT_INSTANTIATE_CUDA_TEST_CASE?

src/uct/base/uct_iface.c
tvegas1: ok app or ucp directly calling `uct_iface_is_reachable_v2()` is still ok with per transport checks. but I think there is this case, app (or maybe ucp):
1. `uct_ep_is_connected()` calls function pointer: `uct_base_ep_is_connected()`, with the now removed flags check.
2. `uct_base_ep_is_connected()` now does not check flags, calls `uct_base_iface_is_reachable()` which _unconditionally_ adds flags with pointer, before calling `uct_iface_is_reachable_v2()`, which cannot check flags, since they are always added.

shasson5: I see, so I think the best thing is to avoid using `uct_base_iface_is_reachable` here, and instead call directly to `uct_iface_is_reachable_v2` (and set only addresses which are passed).

test/gtest/uct/test_uct_ep.cc
shasson5: it's mem_type only at the moment:
https://github.com/openucx/ucx/blob/4655c0445fe5fcce9e4eeffbaea1b8f35baf9d5d/test/gtest/uct/test_uct_perf.cc#L282

can instead rename CUDA -> CUDA_MEMTYPE and then add cuda_ipc to CUDA_TEST_CASE

shasson5: or even better -> rename UCT_TEST_TLS to UCT_TEST_NO_CUDA_IPC_TLS
and then add new UCT_TEST_TLS with cuda_ipc included.

yosefe: Why not instantiate on cuda_ipc as well?

shasson5: It causes failure in some unrelated tests

yosefe: I guess we can use UCT_TEST_NO_CUDA_IPC_TLS, but it's not expected cuda_ipc would fail, need to understand why

test/gtest/uct/uct_test.h
```diff
@@ -599,6 +599,14 @@ class test_uct_iface_attrs : public uct_test {
     UCT_INSTANTIATE_RC_TEST_CASE(_test_case) \
     _UCT_INSTANTIATE_TEST_CASE(_test_case, dc_mlx5)
 
+/**
+ * Instantiate the parametrized test case for CUDA_IPC.
+ *
```
yosefe, line 604: TODO
add cuda_ipc to UCT_INSTANTIATE_TEST_CASE

yosefe (approved):

yosefe (approved):

src/uct/sm/scopy/cma/cma_iface.c
```diff
@@ -83,6 +83,10 @@ uct_cma_iface_is_reachable_v2(const uct_iface_h tl_iface,
         return 0;
     }
 
+    /* Suppress Coverity warning that params->device_addr/iface_addr could be
+     * NULL and dereferenced */
+    ucs_assert((params->device_addr != NULL) && (params->iface_addr != NULL));
```
yosefe, line 88: IMO better check it in uct_iface_is_reachable_params_addrs_valid

src/uct/base/uct_iface.c
```diff
@@ -267,9 +267,14 @@ int uct_iface_is_reachable_params_valid(
...
         const uct_iface_is_reachable_params_t *params)
 {
-    return uct_iface_is_reachable_params_valid(
+    int res = uct_iface_is_reachable_params_valid(
             params, UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR |
                             UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR);
+
+    /* Suppress Coverity warning that params->device_addr/iface_addr could be
+     * NULL and dereferenced */
+    ucs_assert((params->device_addr != NULL) && (params->iface_addr != NULL));
```
yosefe, line 276: instead of assert, shouldn't we return an error?

yosefe (approved):

yosefe:
@shasson5 it seems the failures are related to failing to create memtype_ep lane, probably because of the change in UCT reachability logic in the last commit

shasson5:
reverted

yosefe:
> reverted

IMO we still need to return error when an address is not provided.. Also, I wonder why the assertion does not fail in this case?

shasson5:
>  Also, I wonder why the assertion does not fail in this case?
as a condition for this PR?

shasson5:
> > Also, I wonder why the assertion does not fail in this case?
 As a condition for this PR?

yosefe (approved):

shasson5:
@yosefe can merge?

PR #10017, opened 2024-07-16
UCS/SYS/COMPILER_DEF: use __func__ instead of __FUNCTION__ to meet standard

roiedanino:
## What
Use `__func__` instead of `__FUNCTION__` on supported C versions

## Why ?
To meet [C standard](https://gcc.gnu.org/onlinedocs/gcc/Function-Names.html)

### DIFF
--- a/src/ucm/cuda/cudamem.c
+++ b/src/ucm/cuda/cudamem.c
@@ -36,7 +36,7 @@
         ret = ucm_orig_##_name(ptr_arg, UCM_FUNC_PASS_ARGS(__VA_ARGS__)); \
         if (ret == (_success)) { \
             ptr = _ref ptr_arg; \
-            ucm_trace("%s(" _args_fmt ") allocated %p", __FUNCTION__, \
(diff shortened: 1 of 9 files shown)


src/ucs/sys/compiler_def.h
```diff
@@ -250,4 +250,15 @@
         *(_b) = __tmp; \
     }
 
+/*
+ * Function name macro
+ */
+
+#if __GNUC__ > 2 || \
```
tvegas1, line 257: since we already have unprotected `__func__` usage in UCX codebase and gcc 2.95 dates from 2001, I think we can simply replace `__FUNCTION__` occurences by `__func__`.

roiedanino: Thanks!

tvegas1 (approved):

gleon99 (approved):
Please squash

yosefe (approved):

ivankochin (approved):

PR #10018, opened 2024-07-17
TEST/GTEST: Use c++17

gleon99:
## What
Use C++17 standard in GTEST

## Why ?
Bump to a newer standard, to allow usage of advanced language features and optimizations.


ivankochin:
Many developers still uses RHEL 7 which by default has only gcc4.8.5 that doesn't support even C++14. So do we really want to make impossible gtest compilation on default RHEL 7 distribution?

gleon99:
> Many developers still uses RHEL 7 which by default has only gcc4.8.5 that doesn't support even C++14. So do we really want to make impossible gtest compilation on default RHEL 7 distribution?

Good point, thanks!

[Here](https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/rhel-7-end-of-maintenance#:~:text=Red%20Hat%C2%AE%20Enterprise%20Linux,the%20newest%20features%20and%20benefits.) it says "Red Hat® Enterprise Linux® 7  reached the end of its maintenance support 2 phase on June 30, 2024".

@yosefe wdyt?

yosefe:
> > Many developers still uses RHEL 7 which by default has only gcc4.8.5 that doesn't support even C++14. So do we really want to make impossible gtest compilation on default RHEL 7 distribution?
> 
> Good point, thanks!
> 
> [Here](https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/rhel-7-end-of-maintenance#:~:text=Red%20Hat%C2%AE%20Enterprise%20Linux,the%20newest%20features%20and%20benefits.) it says "Red Hat® Enterprise Linux® 7 reached the end of its maintenance support 2 phase on June 30, 2024".
> 
> @yosefe wdyt?

We still want to support RHEL7 for the time being

PR #10024, opened 2024-07-21
TEST/CALLBACKQ: Fix test_callbackq.test_callbackq test failures

yosefe:
## Why
Fix test failures like:
```
2024-07-10T10:31:07.4474989Z [ RUN      ] test_callbackq.oneshot_mt/mt_10 <> <>
2024-07-10T10:31:07.4478332Z /scrap/azure/agent-04/AZP_WORKSPACE/1/s/contrib/../test/gtest/ucs/test_callbackq.cc:360: Failure
2024-07-10T10:31:07.4478820Z Expected equality of these values:
2024-07-10T10:31:07.4479038Z   1u
2024-07-10T10:31:07.4479254Z     Which is: 1
2024-07-10T10:31:07.4479513Z   ctx.count
2024-07-10T10:31:07.4479810Z     Which is: 1
2024-07-10T10:31:07.4481242Z [  FAILED  ] test_callbackq.oneshot_mt/mt_10, where TypeParam =  and GetParam() =  (1 ms)
2024-07-10T10:31:07.4481612Z [ RUN      ] test_callbackq.remove_self_safe <> <>
```

### DIFF
--- a/test/gtest/common/test.h
+++ b/test/gtest/common/test.h
@@ -75,6 +75,8 @@ protected:
 
     typedef std::vector<ucs_global_opts_t> config_stack_t;
 
+    static constexpr double DEFAULT_TIMEOUT_SEC = 10.0;
+
     void SetUpProxy();
     void TearDownProxy();
     void TestBodyProxy();
@@ -87,6 +89,30 @@ protected:
     virtual void check_skip_test() = 0;
 
     virtual void test_body() = 0;
+
+    template<typename Cond, typename Wait>
+    void wait_for_cond(Cond cond, Wait wait,
+                       double timeout = DEFAULT_TIMEOUT_SEC) const
+    {
+        ucs_time_t deadline = get_deadline(timeout);
+        while ((ucs_get_time() < deadline) && (!cond())) {
+            wait();
+        }
+    }
+
+    template<typename T, typename Wait>
+    void wait_for_value(const volatile T *var, T value, Wait wait,
+                        double timeout = DEFAULT_TIMEOUT_SEC) const
+    {
+        wait_for_cond([var, value]() { return *var == value; }, wait, timeout);
+    }
+
+    template<typename T>
+    void wait_for_value(const volatile T *var, T value,
+                        double timeout = DEFAULT_TIMEOUT_SEC) const
+    {
+        wait_for_value(var, value, sched_yield, timeout);
+    }
 
     static ucs_log_func_rc_t
     common_logger(ucs_log_level_t log_level_to_handle, bool print,
--- a/test/gtest/configure.m4
+++ b/test/gtest/configure.m4
@@ -16,11 +16,17 @@ CHECK_COMPILER_FLAG([--diag_suppress 186], [--diag_suppress 186],
                     [AC_LANG_SOURCE([[int main(int argc, char** argv){return 0;}]])],
                     [GTEST_CXXFLAGS="$GTEST_CXXFLAGS --diag_suppress 186"],
                     [])
-                    
+
 # error #236: controlling expression is constant
 CHECK_COMPILER_FLAG([--diag_suppress 236], [--diag_suppress 236],
                     [AC_LANG_SOURCE([[int main(int argc, char** argv){return 0;}]])],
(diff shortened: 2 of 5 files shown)


yosefe:
@tvegas1 @iyastreb can you pls review?

test/gtest/common/test.h
```diff
@@ -88,6 +90,17 @@ class test_base {
...
+    template<typename T>
+    void wait_for_value(volatile T *var, T value,
+                        double timeout = DEFAULT_TIMEOUT_SEC) const
+    {
+        ucs_time_t deadline = ucs_get_time() + (ucs_time_from_sec(timeout) *
+                                                ucs::test_time_multiplier());
+        while ((ucs_get_time() < deadline) && (*var != value)) {
+            sched_yield();
+        }
+    }
```
iyastreb, line 102: I counted 3 exiting reincarnations on exactly this function in UCX, and now this is the 4th.
```
ucp_test::wait_for_value
2 overloads in uct_test::wait_for_value
```
Apart from that there are multiple variations of the same function with slightly different names/behavior:
```
wait_progress
wait_for_rx_sn
many more...
```
Maybe we need to generalize it so that we don't have dozens of the similar functions?
Something like:
```C++
    template<typename Cond, typename Wait>
    void wait_for(Cond cond, Wait wait,
                        double timeout = DEFAULT_TIMEOUT_SEC) const
    {
        ucs_time_t deadline = ucs_get_time() + (ucs_time_from_sec(timeout) *
                                                ucs::test_time_multiplier());
        while ((ucs_get_time() < deadline) && (!cond())) {
            wait();
        }
    }

    template<typename T, Wait>
    void wait_for_value(volatile T *var, T value, Wait wait,
                        double timeout = DEFAULT_TIMEOUT_SEC) const
    {
        wait_for([&](){ return (*var == value); }, wait, timeout);
    }

Usage:
wait_for_value(..., sched_yield);

We can have overloads/default values for different wait strategies, or maybe leave it explicit
```

test/gtest/common/test.h
```diff
@@ -88,6 +90,17 @@ class test_base {
...
 
+    template<typename T>
+    void wait_for_value(volatile T *var, T value,
+                        double timeout = DEFAULT_TIMEOUT_SEC) const
+    {
+        ucs_time_t deadline = ucs_get_time() + (ucs_time_from_sec(timeout) *
+                                                ucs::test_time_multiplier());
+        while ((ucs_get_time() < deadline) && (*var != value)) {
+            sched_yield();
+        }
```
iyastreb, line 101: We don't expect timeout to happen, so maybe add a check at the end:
`EXPECT_GT(deadline, ucs_get_time());`

test/gtest/common/test.h
```diff
@@ -88,6 +90,17 @@ class test_base {
 
     virtual void test_body() = 0;
 
+    template<typename T>
+    void wait_for_value(volatile T *var, T value,
+                        double timeout = DEFAULT_TIMEOUT_SEC) const
+    {
+        ucs_time_t deadline = ucs_get_time() + (ucs_time_from_sec(timeout) *
```
tvegas1, line 97: maybe leverage/generalize `get_deadline()` from `test_helpers.cc`?

tvegas1 (commented):
maybe we can merge similar functions to some extent (wait_for_flag/wait_for_bits/wait_for_value..), gain might not be huge as code is small.

test/gtest/common/test.h
yosefe: done

test/gtest/common/test.h
yosefe: fixed

test/gtest/common/test.h
yosefe: existing tests do that check already, so i'd prefer to not make such significant change in this PR

iyastreb (dismissed):

tvegas1 (dismissed):

tvegas1 (dismissed):

yosefe:
@tvegas1 can you pls re-review the small build fix?

tvegas1 (approved):

PR #10029, opened 2024-07-25
UCP/RNDV: Fix sending too many ATPs for small messages

yosefe:
## Why
Fix internal issue [4003121](https://redmine.mellanox.com/issues/4003121) and [3998165](https://redmine.mellanox.com/issues/3998165)

### DIFF
--- a/src/ucp/dt/datatype_iter.inl
+++ b/src/ucp/dt/datatype_iter.inl
@@ -619,23 +619,15 @@ ucp_datatype_iter_copy_position(ucp_datatype_iter_t *dt_iter,
 }
 
 /*
- * Check if the next iterator has reached the end
- */
-static UCS_F_ALWAYS_INLINE int
-ucp_datatype_iter_is_end_position(const ucp_datatype_iter_t *dt_iter,
-                                  const ucp_datatype_iter_t *pos_iter)
-{
-    ucs_assert(dt_iter->offset <= dt_iter->length);
-    return pos_iter->offset == dt_iter->length;
-}
-
-/*
  * Check if the iterator has reached the end
  */
 static UCS_F_ALWAYS_INLINE int
 ucp_datatype_iter_is_end(const ucp_datatype_iter_t *dt_iter)
 {
-    return ucp_datatype_iter_is_end_position(dt_iter, dt_iter);
+    ucs_assertv(dt_iter->offset <= dt_iter->length,
+                "dt_iter=%p dt_iter->offset=%zu dt_iter->length=%zu", dt_iter,
+                dt_iter->offset, dt_iter->length);
+    return dt_iter->offset == dt_iter->length;
 }
 
 static UCS_F_ALWAYS_INLINE void
@@ -645,6 +637,20 @@ ucp_datatype_iter_rewind(ucp_datatype_iter_t *dt_iter, unsigned dt_mask)
     if (ucp_datatype_iter_is_class(dt_iter, UCP_DATATYPE_IOV, dt_mask)) {
         dt_iter->type.iov.iov_index  = 0;
         dt_iter->type.iov.iov_offset = 0;
+    }
+}
+
+static UCS_F_ALWAYS_INLINE void
+ucp_datatype_iter_seek(ucp_datatype_iter_t *dt_iter, size_t offset,
+                       unsigned dt_mask)
+{
+    ucs_assertv(offset <= dt_iter->length,
+                "dt_iter=%p offset=%zu dt_iter->length=%zu", dt_iter,
+                offset, dt_iter->length);
+    if (ucp_datatype_iter_is_class(dt_iter, UCP_DATATYPE_IOV, dt_mask)) {
+        ucp_datatype_iter_iov_seek(dt_iter, offset);
+    } else {
+        dt_iter->offset = offset;
     }
 }
 
--- a/src/ucp/rndv/rndv_put.c
+++ b/src/ucp/rndv/rndv_put.c
@@ -126,26 +126,40 @@ ucp_proto_rndv_put_common_atp_send(ucp_request_t *req, ucp_lane_index_t lane)
 {
     const ucp_proto_rndv_put_priv_t *rpriv = req->send.proto_config->priv;
     ucp_proto_rndv_put_atp_pack_ctx_t pack_ctx;
+    ucs_status_t status;
+
+    /* Make sure the sum of ack_size field in all ATP messages we send will not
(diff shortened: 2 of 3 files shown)


src/ucp/dt/datatype_iter.inl
```diff
@@ -648,6 +650,20 @@ ucp_datatype_iter_rewind(ucp_datatype_iter_t *dt_iter, unsigned dt_mask)
     }
 }
```
ivankochin, line 652: Add some description?

src/ucp/dt/datatype_iter.inl
```diff
@@ -648,6 +650,20 @@ ucp_datatype_iter_rewind(ucp_datatype_iter_t *dt_iter, unsigned dt_mask)
...
+ucp_datatype_iter_seek(ucp_datatype_iter_t *dt_iter, size_t offset,
+                       unsigned dt_mask)
+{
+    ucs_assertv(offset <= dt_iter->length,
+                "dt_iter=%p offset=%zu dt_iter->length=%zu", dt_iter,
+                offset, dt_iter->length);
+    if (ucp_datatype_iter_is_class(dt_iter, UCP_DATATYPE_IOV, dt_mask)) {
+        ucp_datatype_iter_iov_seek(dt_iter, offset);
+    } else {
+        dt_iter->offset = offset;
```
ivankochin, line 663: Does it work for GENERIC datatype?

src/ucp/rndv/rndv_put.c
```diff
@@ -416,21 +425,25 @@ ucp_proto_rndv_put_zcopy_query(const ucp_proto_query_params_t *params,
...
 {
-    const ucp_proto_rndv_put_priv_t *rpriv = request->send.proto_config->priv;
-    ucp_datatype_iter_t *dt_iter           = &request->send.state.dt_iter;
-    ucp_lane_map_t atp_sent_map;
+    const ucp_proto_rndv_put_priv_t *rpriv = req->send.proto_config->priv;
 
-    atp_sent_map = rpriv->atp_map & UCS_MASK(request->send.rndv.put.atp_lane);
-    if (atp_sent_map == rpriv->atp_map) {
-        dt_iter->offset = dt_iter->length;
+    if (req->send.rndv.put.atp_count == rpriv->atp_num_lanes) {
```
ivankochin, line 432: AFAIU we can return `UCS_OK` from `ucp_proto_rndv_put_common_atp_send` if `req->send.rndv.put.atp_count >= req->send.state.dt_iter.length` so it would be considered here as not all ATPs were sent which is wrong.

src/ucp/rndv/rndv_put.c
```diff
@@ -126,26 +126,34 @@ ucp_proto_rndv_put_common_atp_send(ucp_request_t *req, ucp_lane_index_t lane)
 {
     const ucp_proto_rndv_put_priv_t *rpriv = req->send.proto_config->priv;
     ucp_proto_rndv_put_atp_pack_ctx_t pack_ctx;
+    ucs_status_t status;
 
     pack_ctx.req = req;
 
+    if (req->send.rndv.put.atp_count >= req->send.state.dt_iter.length) {
```
ivankochin, line 133: 1. Do I understand correctly that there is no correctness problem in calculating `atp_sent_map`(and therefore `atp_sent_count`) as `rpriv->atp_map & UCS_MASK(request->send.rndv.put.atp_lane);` and you introduced `atp_count` request field mostly for performance purposes?
2. I understand that this if does its job, but AFAIK it works only for `min_length==1`. So I would suggest to create a more general case when `atp_count` compared with `dt_iter.length / min_length`. Or maybe add assertion that `min_length == 1` and probably add some comment to explain what does this if do.

yosefe: 1. atp_count was added to check the actual number of atp messages sent, so it will not exceed "length". if we counted it using "lane" and mask, it would involve popcount bit operation which is expensive
2. unfortunately, setting min_length to 1 exposes issue with incorrect handling of overlap on sender side (when receiving RTR). so have to keep min_length==0 for rndv/put protocol. so IMO we can keep the same behavior as the previous code - which is to ignore 0-length ATP - and set min_length=1 when the overlap issue is fixed

src/ucp/rndv/rndv_put.c
yosefe: it would mean we sent all ATPs we needed to send, since we should not send more ATPs than the request length

src/ucp/rndv/rndv_put.c
```diff
@@ -126,26 +126,34 @@ ucp_proto_rndv_put_common_atp_send(ucp_request_t *req, ucp_lane_index_t lane)
 {
     const ucp_proto_rndv_put_priv_t *rpriv = req->send.proto_config->priv;
     ucp_proto_rndv_put_atp_pack_ctx_t pack_ctx;
+    ucs_status_t status;
+
+    if (req->send.rndv.put.atp_count >= req->send.state.dt_iter.length) {
```
ivankochin, line 131:
```suggestion
    ucs_assertv(req->send.rndv.put.atp_count <= req->send.state.dt_iter.length, ...);
    if (req->send.rndv.put.atp_count == req->send.state.dt_iter.length) {
```

ivankochin (approved):

src/ucp/rndv/rndv_put.c
```diff
@@ -126,26 +126,40 @@ ucp_proto_rndv_put_common_atp_send(ucp_request_t *req, ucp_lane_index_t lane)
...
+    status = ucp_proto_am_bcopy_single_send(req, UCP_AM_ID_RNDV_ATP, lane,
+                                            ucp_proto_rndv_put_common_pack_atp,
+                                            &pack_ctx,
+                                            sizeof(ucp_rndv_ack_hdr_t), 0);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    ++req->send.rndv.put.atp_count;
+    return UCS_OK;
```
brminich, line 162: one line shorter:
```suggestion
    if (status == UCS_OK) {
        ++req->send.rndv.put.atp_count;      
    }

    return status;
```

yosefe: the suggestion is shorter but the current implementation is according to code style where error flow is checked first

brminich (approved):

ivankochin:
Is that possible to add some tests for that? E.g. create 8 rndv lanes, specify RNDV_SCHEME=put_zcopy and send 4 bytes?

PR #10030, opened 2024-07-26
TEST/RNDV: Run memtype rndv tests on 1-byte message

yosefe:
## Why
- Increase test coverage
- Force rendezvous protocol by flag rather than by threshold, which is the preferred way

### DIFF
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -612,7 +612,7 @@ protected:
         EXPECT_LT(m_recv_counter, m_send_counter);
         check_header(header, header_length);
 
-        bool has_reply_ep = get_send_flag();
+        bool has_reply_ep = get_send_flag() & UCP_AM_SEND_FLAG_REPLY;
 
         EXPECT_EQ(has_reply_ep,
                   !!(rx_param->recv_attr & UCP_AM_RECV_ATTR_FIELD_REPLY_EP));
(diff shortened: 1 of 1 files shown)


test/gtest/ucp/test_ucp_am.cc
```diff
@@ -612,7 +612,7 @@ class test_ucp_am_nbx : public test_ucp_am_base {
         EXPECT_LT(m_recv_counter, m_send_counter);
         check_header(header, header_length);
 
-        bool has_reply_ep = get_send_flag();
+        bool has_reply_ep = get_send_flag() & UCP_AM_SEND_FLAG_REPLY;
```
yosefe, line 615: bugfix

yosefe:
@ivankochin can you pls review?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1896,19 +1896,24 @@ class test_ucp_am_nbx_rndv_memtype : public test_ucp_am_nbx_rndv {
         return test_ucp_am_nbx_eager_memtype::get_test_variants(variants);
     }
 
-    void init()
+    void init() override
     {
         modify_config("RNDV_THRESH", "128");
```
ivankochin, line 1901: Can this threshold setting be removed since now `UCP_AM_SEND_FLAG_RNDV` is passed?

ivankochin (approved):

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1896,19 +1896,23 @@ class test_ucp_am_nbx_rndv_memtype : public test_ucp_am_nbx_rndv {
...
     {
-        modify_config("RNDV_THRESH", "128");
         test_ucp_am_nbx::init();
     }
 
 private:
-    virtual ucs_memory_type_t tx_memtype() const
+    unsigned get_send_flag() const override
+    {
+        return test_ucp_am_nbx_rndv::get_send_flag() | UCP_AM_SEND_FLAG_RNDV;
```
brminich, line 1907: maybe set this flag in `test_ucp_am_nbx_rndv` instead and remove setting `RNDV_THRESH` from its constructor?

yosefe: it may invalidate some tests that check for the specific threshold - for example rndv_flag_zero_send, just_header_rndv, header_and_data_rndv

brminich (approved):

PR #10041, opened 2024-08-02
Fix failing test for MAD that breaks build

jeroen-mostert:
On at least one distro I've tested (Void, in case it matters) the shell chokes on using the Bash-specific `==` syntax rather than POSIX Bourne `=`. This happens whether or not `--with-mad` is specified.

### DIFF
--- a/config/m4/mad.m4
+++ b/config/m4/mad.m4
@@ -14,7 +14,7 @@ AC_ARG_WITH([mad],
             [with_mad=guess])
 
 mad_happy=no
-AS_IF([test "x$with_mad" == "xno"],
+AS_IF([test "x$with_mad" = "xno"],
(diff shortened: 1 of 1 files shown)


jeroen-mostert:
Reading the contrib guidelines, be advised that I won't be signing a document to remove a single equality sign. I give everyone in this world permission to use my improvement and incorporate it themselves, if they so wish, or throw it away, if they're a stickler for the rules. :P

tvegas1:
> Reading the contrib guidelines, be advised that I won't be signing a document to remove a single equality sign. I give everyone in this world permission to use my improvement and incorporate it themselves, if they so wish, or throw it away, if they're a stickler for the rules. :P

Thanks for the report, started another PR to fix that.

tvegas1:
Will be fixed with #10043.

PR #10047, opened 2024-08-05
COVERITY: Allow iterative runs for faster turnaround time

tvegas1:
## What
Allow iterative runs for coverity script.

## Why ?
A full run takes more than 30 minutes. Using the added `INCREMENTAL=yes` allows rebuild with html generation within 6 minutes.

## How ?
Split configure and cleanup part.
### Tested
```
# First run
$ parallel_jobs=32 ENABLE_JAVA=no BUILD_ID=scrap/build ./buildlib/tools/coverity.sh devel

# Following runs
$ INCREMENTAL=yes parallel_jobs=32 ENABLE_JAVA=no BUILD_ID=scrap/build ./buildlib/tools/coverity.sh devel
```

### DIFF
--- a/buildlib/tools/coverity.sh
+++ b/buildlib/tools/coverity.sh
@@ -1,10 +1,44 @@
-#!/bin/bash -eExl
+#!/bin/bash -eEl
 
 realdir=$(realpath $(dirname $0))
 source ${realdir}/common.sh
 source ${realdir}/../az-helpers.sh
 
 COV_MODULE="tools/cov-2019.12"
+
+usage() {
+	echo "Usage: $0 <release|devel> [--clean] [<configure opts>]"
+	echo
+	echo "Commands:"
+	echo "    release    build in release mode"
+	echo "    devel      build in development mode"
+	echo
+	echo "Options:"
+	echo "    --clean    start from a full build"
+	exit 1
+}
+
+parse_args() {
+	if [ "$#" -lt 1 ]; then
+		usage
+	fi
+
+	mode=$1
+	shift
+
+	if [ "$mode" != release -a "$mode" != devel ]; then
+		usage
+	fi
+
+	if [ "$#" -gt 0 -a "$1" = "--clean" ]; then
+		clean=yes
+		shift
+	else
+		clean=no
+	fi
+
+	configure_opts=$*
+}
 
 #
 # Run Coverity and report errors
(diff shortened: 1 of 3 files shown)


yosefe:
Is anyone using coverity.sh outside of CI?

buildlib/tools/coverity.sh
```diff
@@ -79,5 +92,14 @@ run_coverity() {
 	return $rc
 }
 
-prepare_build
+az_init_modules
+modules_for_coverity
```
brminich, line 96: seems like you can still keep it inside `run_coverity`

buildlib/tools/coverity.sh
```diff
@@ -44,31 +44,44 @@ modules_for_coverity_unload() {
...
 
 	xpmem_root=$(module show $XPMEM_MODULE 2>&1 | awk '/CPATH/ {print $3}' | sed -e 's,/include,,')
 	with_xpmem="--with-xpmem=$xpmem_root"
 
-	${WORKSPACE}/contrib/configure-$ucx_build_type --prefix=$ucx_inst --with-cuda --with-gdrcopy --with-java $with_xpmem
+	if [ "${ENABLE_JAVA:-yes}" = yes ]; then
+		java_opt="--with-java"
+	else
+		java_opt="--without-java"
+	fi
```
yosefe, line 57: I think that if we're adapting this script to usage without CI, make it more generic:
- splitting build/analyze is good step
- instead of using ENABLE_JAVA as env var, pass build options and control options using command line args. and document the usage with --help
- avoid checking INCREMENTAL env var; just let users set env vars outside the script

buildlib/tools/coverity.sh
```diff
@@ -79,5 +92,14 @@ run_coverity() {
 	return $rc
 }
 
-prepare_build
+az_init_modules
```
yosefe, line 95: how does it work when running outside the CI? seems this script is not really standalone

tvegas1: tbh, i did not see anything very az specific. there is az_module_load, but it's only logging that is AZ specific.

yosefe: can we remove it then?

buildlib/tools/coverity.sh
tvegas1: i think for coverity one coverity-build needs one analyze so we don't generally need to analyze repeatedly on a given build. what about doing like below:

```
$ coverity.sh release --no-clean --no-java
$ coverity.sh release
$ coverity.sh devel

$ coverity.sh
$ coverity.sh help
coverity.sh <release|devel> [--no-clean] [--no-java]
  --no-clean     reuse the existing build
  --no-java      do not enable java in the build
```

yosefe: Maybe "--clean" (opt-in instead of opt-out)
instead of --no-java we can pass custom config args (--with-java or --without-java if needed; if neither is passed, configure will use guess as default mode)

buildlib/tools/coverity.sh
tvegas1: it's difficult because of all naming az_module_load which in practice does a module load with hypothetical `az_log_warning` on failure.

But after checking, even that logging function does finally check for AZ presence:
```
# Report an warning message to Azure pipeline
function azure_log_warning() {
    test "x$RUNNING_IN_AZURE" = "xno" && return
    msg=$1
    set +x
    echo "##vso[task.logissue type=warning]${msg}"
}
```

tvegas1: tbh, even being able to trigger azure logs locally (setting RUNNING_IN_AZURE=1) could be useful when developping script

buildlib/tools/coverity.sh
tvegas1: makes sense, something like:
```
$ coverity.sh help
coverity.sh <release|devel> [--clean] [<configure opts>]
  --clean     reuse the existing build
```

yosefe: yeah, every unrecognized option (e.g not --clean) is passed to configure

buildlib/tools/coverity.sh
yosefe: ok

buildlib/tools/coverity.sh
tvegas1: i think configure step needs it because it load gdrcopy headers etc.

brminich (approved):

yosefe (approved):

tvegas1:
confirmed with the test PR #10061 that coverity failures still make CI properly fail.

PR #10050, opened 2024-08-07
UCS/VFS: inotify watch from private user folder

iyastreb:
## What
Make VFS unix socket path configurable with UCX_VFS_SOCK_PATH option
By default put listening socket in dedicated folder: `/run/user/<user_id>/ucx/vfs.sock`

## Why ?
We use inotify watch from the client side to detect VFS server presence. When we run inotify watch in a public folder (/tmp) with many files, then we experience the following issues:
- inotify_add_watch can take more than a second to start (in NVHPC cluster), and the main thread is delayed
- we wake VFS fuse thread on all the file create events, even unrelated to us

## How ?
Added option UCX_VFS_SOCK_PATH to configure socket path on the client side. Before it was only configurable on VFS daemon side with -l option.
Use %i wildcard that represents user_id. Ideally we should use `ucs_fill_filename_template` here, but ucs library is not linked by vfs_sock, and it's not desirable to add this dependency. 
By default the folder for VFS socket is `/run/user/<user_id>/ucx`, and both client and server creates this folder - this way we guarantee that no interference happens with other processes, and no wakeups on irrelevant files created

### DIFF
--- a/src/tools/vfs/vfs_daemon.h
+++ b/src/tools/vfs/vfs_daemon.h
@@ -25,18 +25,8 @@ enum {
 };
 
 
-#define vfs_error(_fmt, ...) \
-    { \
-        fprintf(stderr, "Error: " _fmt "\n", ##__VA_ARGS__); \
-    }
-
-
-#define vfs_log(_fmt, ...) \
-    { \
-        if (g_opts.verbose) { \
-            fprintf(stderr, "Debug: " _fmt "\n", ##__VA_ARGS__); \
-        } \
-    }
+#define vfs_error ucs_error
+#define vfs_log   ucs_debug
 
 
 typedef struct {
@@ -45,7 +35,6 @@ typedef struct {
     int        verbose;
     const char *mountpoint_dir;
     const char *mount_opts;
-    const char *sock_path;
 } vfs_opts_t;
 
 
--- a/src/tools/vfs/vfs_main.c
+++ b/src/tools/vfs/vfs_main.c
@@ -10,6 +10,7 @@
 
 #include "vfs_daemon.h"
 
+#include <ucs/debug/log_def.h>
 #include <sys/wait.h>
 #include <limits.h>
 #include <stdlib.h>
@@ -22,8 +23,7 @@ vfs_opts_t g_opts = {
     .foreground     = 0,
     .verbose        = 0,
     .mountpoint_dir = VFS_DEFAULT_MOUNTPOINT_DIR,
-    .mount_opts     = "",
-    .sock_path      = NULL
+    .mount_opts     = ""
 };
 
 const char *vfs_action_names[] = {
@@ -254,6 +254,11 @@ static int vfs_listen(int silent_addinuse_err)
         goto out;
     }
 
+    ret = ucs_vfs_sock_mkdir(g_sockaddr.sun_path, UCS_LOG_LEVEL_ERROR);
+    if (ret != 0) {
+        goto out;
+    }
+
     listen_fd = socket(AF_UNIX, SOCK_STREAM, 0);
     if (listen_fd < 0) {
         ret = -errno;
@@ -387,9 +392,6 @@ out:
 
 static void vfs_usage()
 {
-    struct sockaddr_un sock_addr = {};
-
-    ucs_vfs_sock_get_address(&sock_addr);
     printf("Usage:   ucx_vfs [options]  [action]\n");
     printf("\n");
     printf("Options:\n");
@@ -398,8 +400,6 @@ static void vfs_usage()
     printf("  -o <opts>  Pass these mount options to mount.fuse\n");
     printf("  -f         Do not daemonize; run in foreground\n");
     printf("  -v         Enable verbose logging (requires -f)\n");
-    printf("  -l <path>  Set listening unix socket path (default: %s)\n",
-           sock_addr.sun_path);
     printf("\n");
     printf("Actions:\n");
     printf("   start     Run the daemon and listen for connection from UCX\n");
@@ -414,7 +414,7 @@ static int vfs_parse_args(int argc, char **argv)
     const char *action_str;
     int c, i;
 
-    while ((c = getopt(argc, argv, "d:o:vfl:h")) != -1) {
+    while ((c = getopt(argc, argv, "d:o:vfh")) != -1) {
         switch (c) {
         case 'd':
             g_opts.mountpoint_dir = optarg;
@@ -427,9 +427,6 @@ static int vfs_parse_args(int argc, char **argv)
             break;
         case 'f':
             g_opts.foreground = 1;
-            break;
-        case 'l':
-            g_opts.sock_path = optarg;
             break;
         case 'h':
         default:
@@ -494,14 +491,7 @@ int main(int argc, char **argv)
         fuse_daemonize(0);
     }
 
-    if (g_opts.sock_path == NULL) {
-        ucs_vfs_sock_get_address(&g_sockaddr);
-    } else {
-        g_sockaddr.sun_family = AF_UNIX;
-        memset(g_sockaddr.sun_path, 0, sizeof(g_sockaddr.sun_path));
-        strncpy(g_sockaddr.sun_path, g_opts.sock_path,
-                sizeof(g_sockaddr.sun_path) - 1);
-    }
+    ucs_vfs_sock_get_address(&g_sockaddr);
 
     switch (g_opts.action) {
     case VFS_DAEMON_ACTION_START:
--- a/src/ucs/sys/string.c
+++ b/src/ucs/sys/string.c
@@ -48,45 +48,39 @@ void ucs_fill_filename_template(const char *tmpl, char *buf, size_t max)
         length = ucs_min(pp - pf, end - p);
         strncpy(p, pf, length);
         p += length;
+        /* default length of the modifier (e.g. %p) */
+        length = 2;
 
         switch (*(pp + 1)) {
         case 'p':
             snprintf(p, end - p, "%d", getpid());
-            pf = pp + 2;
-            p += strlen(p);
             break;
         case 'h':
             snprintf(p, end - p, "%s", ucs_get_host_name());
-            pf = pp + 2;
-            p += strlen(p);
             break;
         case 'c':
             snprintf(p, end - p, "%02d", ucs_get_first_cpu());
-            pf = pp + 2;
-            p += strlen(p);
             break;
         case 't':
             t = time(NULL);
             strftime(p, end - p, "%Y-%m-%d-%H-%M-%S", localtime(&t));
-            pf = pp + 2;
-            p += strlen(p);
             break;
         case 'u':
             snprintf(p, end - p, "%s", ucs_basename(ucs_get_user_name()));
-            pf = pp + 2;
-            p += strlen(p);
             break;
         case 'e':
             snprintf(p, end - p, "%s", ucs_basename(ucs_get_exe()));
-            pf = pp + 2;
-            p += strlen(p);
+            break;
+        case 'i':
+            snprintf(p, end - p, "%u", geteuid());
             break;
         default:
             *(p++) = *pp;
-            pf = pp + 1;
+            length = 1;
             break;
         }
 
+        pf = pp + length;
         p += strlen(p);
     }
     *p = 0;
--- a/src/ucs/vfs/fuse/vfs_fuse.c
+++ b/src/ucs/vfs/fuse/vfs_fuse.c
@@ -250,14 +250,16 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
     ucs_status_t status;
     ssize_t nread;
     size_t offset;
+    int ret;
 
     pthread_mutex_lock(&ucs_vfs_fuse_context.mutex);
 
-    /* copy path components to 'dir_buf' and 'watch_filename' */
-    ucs_strncpy_safe(dir_buf, path, sizeof(dir_buf));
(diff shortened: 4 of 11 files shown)


src/ucs/vfs/sock/vfs_sock.c
```diff
@@ -25,25 +25,27 @@ typedef struct {
...
-    struct passwd *pw;
-    uid_t euid;
+    if (NULL != strstr(tmpl, "%i")) {
+        snprintf(buf, max, tmpl, (int)geteuid());
+    } else {
+        strncpy(buf, tmpl, max);
+    }
+}
 
+void ucs_vfs_sock_get_address(struct sockaddr_un *un_addr, const char *tmpl)
```
tvegas1, line 39: tmpl -> template?

src/ucs/vfs/fuse/vfs_fuse.c
```diff
@@ -328,8 +337,9 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
             }
 
             ucs_trace("file '%s' created", event->name);
+            /* event->len is multiply of 16, not the string length */
```
tvegas1, line 340: "is a multiple of 16"

src/ucs/sys/string.c
```diff
@@ -81,6 +81,11 @@ void ucs_fill_filename_template(const char *tmpl, char *buf, size_t max)
             pf = pp + 2;
             p += strlen(p);
             break;
+        case 'i':
+            snprintf(p, end - p, "%u", geteuid());
+            pf = pp + 2;
+            p += strlen(p);
```
tvegas1, line 87: seems we could remove all `p += strlen(p);` in the switch cases and rely on line 95 that does the same?

iyastreb: Yes, true.
In fact this function is not used at all by the VFS, so I just improved it as we discussed with Yossi. If we keep these changes, I'll apply the change you suggested

src/ucs/vfs/fuse/vfs_fuse.c
iyastreb: sure

src/ucs/vfs/sock/vfs_sock.c
iyastreb: I put `tmpl` like in `ucs_fill_filename_template`, and I think the reason is not to interfere with C++ template keyword (even if it's C)

src/ucs/vfs/sock/vfs_sock.c
```diff
@@ -25,25 +25,27 @@ typedef struct {
...
 
-void ucs_vfs_sock_get_address(struct sockaddr_un *un_addr)
+/* Ideally we should use ucs_fill_filename_template here, but ucs library is
+ * not linked by vfs_sock */
+static void ucs_vfs_sock_template(const char *tmpl, char *buf, size_t max)
 {
-    struct passwd *pw;
-    uid_t euid;
+    if (NULL != strstr(tmpl, "%i")) {
+        snprintf(buf, max, tmpl, (int)geteuid());
```
yosefe, line 33: what if %i appears more than once, or we have other instances of format sequence (%-something) in the string?

src/ucs/vfs/sock/vfs_sock.c
```diff
@@ -25,25 +25,27 @@ typedef struct {
...
-    pw   = getpwuid(euid);
-    if (pw != NULL) {
-        /* By name */
-        snprintf(un_addr->sun_path, sizeof(un_addr->sun_path) - 1,
-                 "/tmp/ucx-vfs-%s.sock", pw->pw_name);
-    } else {
-        /* By number */
-        snprintf(un_addr->sun_path, sizeof(un_addr->sun_path) - 1,
-                 "/tmp/ucx-vfs-%u.sock", euid);
+    if (NULL == tmpl) {
```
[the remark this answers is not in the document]
iyastreb: ok

src/ucs/vfs/fuse/vfs_fuse.c
```diff
@@ -259,6 +260,14 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
                      sizeof(watch_filename));
     watch_dirname = dirname(dir_buf);
 
+    /* Create directory path */
+    ret = mkdir(watch_dirname, S_IRWXU);
+    if ((ret < 0) && (errno != EEXIST)) {
+        ucs_diag("failed to create directory '%s': %m", watch_dirname);
+        status = UCS_ERR_IO_ERROR;
+        goto out;
+    }
```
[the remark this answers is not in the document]
iyastreb: yes, will be done

src/ucs/vfs/sock/vfs_sock.c
iyastreb: > what if %i appears more than once, or we have other instances of format sequence (%-something) in the string?

well, then obviously it does not work.. Then we want to use UCS template function, and we can either:
- move it to vfs_sock module (then it means we need to move all other functions used by it)
- move impl to header (and also for all functions used inside it)
- link UCS with VFS daemon (and disable VFS_ENABLE to prevent spawning VFS fuse thread. Maybe also remove custom vfs_error logging)
What do you prefer?

src/ucs/vfs/sock/vfs_sock.c
tvegas1: ah yes makes sense

src/ucs/vfs/sock/vfs_sock.c
yosefe: what if we include string.c directly in the ucx_vfs tool build?

iyastreb: I evaluated each approach, and I think that the right way is to link UCS with VFS daemon.
In this case we can also get rid of code duplication in VFS.
I think we can just disable loading UCS modules, maybe with weak linkage

iyastreb: > what if we include string.c directly in the ucx_vfs tool build?

No good..
Because that template also depends on many other functions (the most problematic one is `ucs_get_first_cpu`) which are defined in sys.c.

yosefe: ok let's try to link with UCS and see how it goes

iyastreb: pushed
IMO it looks good, removed quite some code duplication

iyastreb: If you like the approach, we can also get rid of separate lib `libucs_vfs_sock`: it should be part of UCS.
We can also maybe refactor `verbose` option, because now we can use regular UCX_LOG_LEVEL

yosefe: maybe in next PR, since the goal of this PR is customer bugfix so let's not delay it..

src/tools/vfs/vfs_main.c
```diff
@@ -246,30 +246,26 @@ static int vfs_unlink_socket(int silent_notexist)
...
 
-    strncpy(dir_buf, g_sockaddr.sun_path, sizeof(dir_buf));
-    sock_dirname = dirname(dir_buf);
-    ret          = mkdir(sock_dirname, S_IRWXU);
-    if ((ret < 0) && (errno != EEXIST)) {
-        ret = -errno;
-        vfs_error("failed to create directory '%s': %m", sock_dirname);
+    ret = ucs_vfs_sock_mkdir(g_sockaddr.sun_path, dir, sizeof(dir));
+    if (ret != 0) {
+        ucs_error("failed to create directory '%s': %m", dir);
```
yosefe, line 261: move it inside ucs_vfs_sock_mkdir

src/tools/vfs/vfs_daemon.h
```diff
@@ -25,27 +25,12 @@ enum {
...
 
-#define vfs_error(_fmt, ...) \
-    { \
-        fprintf(stderr, "Error: " _fmt "\n", ##__VA_ARGS__); \
-    }
-
-
-#define vfs_log(_fmt, ...) \
-    { \
-        if (g_opts.verbose) { \
```
yosefe, line 36: seems g_opts.verbose is not relevant any more, let's revert this refactoring for now

src/ucs/sys/init.c
```diff
@@ -83,7 +83,9 @@ static UCS_F_NOOPTIMIZE void ucs_check_cpu_flags(void)
     }
 }
 
-static void ucs_modules_load()
+void ucs_modules_load();
+
+void __attribute__((weak)) ucs_modules_load()
```
yosefe, line 88: why needed? pls add comment

src/ucs/vfs/sock/vfs_sock.c
```diff
@@ -10,42 +10,45 @@
...
-        strncpy(buf, tmpl, max);
-    }
+    memset(un_addr, 0, sizeof(*un_addr));
+    un_addr->sun_family = AF_UNIX;
+    ucs_fill_filename_template(ucs_global_opts.vfs_sock_path, un_addr->sun_path,
+                               sizeof(un_addr->sun_path));
 }
 
-void ucs_vfs_sock_get_address(struct sockaddr_un *un_addr, const char *tmpl)
+int ucs_vfs_sock_mkdir(const char *sock_path, char *dir, size_t dir_size)
```
yosefe, line 40: can we avoid passing 'dir' argument?

src/ucs/vfs/sock/vfs_sock.c
```diff
@@ -10,42 +10,45 @@
...
 {
-    memset(un_addr, 0, sizeof(*un_addr));
-    un_addr->sun_family = AF_UNIX;
+    int ret;
 
-    if (NULL == tmpl) {
-        tmpl = UCX_VFS_SOCK_DEFAULT_PATH;
+    strncpy(dir, sock_path, dir_size);
```
yosefe, line 44: ucs_strncpy_safe

src/ucs/vfs/sock/vfs_sock.c
```diff
@@ -10,42 +10,45 @@
...
+    dirname(dir);
+    ret = mkdir(dir, S_IRWXU);
```
yosefe, line 46: `ret = mkdir(dirname(sock_path_copy), S_IRWXU);`

src/ucs/sys/init.c
iyastreb: It's needed to disable UCS module loading from VFS (which defines strong symbol `ucs_modules_load`)
Ok, will add a comment

src/ucs/vfs/sock/vfs_sock.c
iyastreb: sure

src/ucs/vfs/sock/vfs_sock.c
iyastreb: Ok, I'll get rid of it

src/ucs/vfs/sock/vfs_sock.c
iyastreb: ok

src/ucs/sys/init.c
```diff
@@ -83,7 +83,11 @@ static UCS_F_NOOPTIMIZE void ucs_check_cpu_flags(void)
     }
 }
 
-static void ucs_modules_load()
+void ucs_modules_load();
+
+/* Weak linker symbol to allow override it in certain binaries (ucx_vfx), which
```
tvegas1, line 88: typo -> vfx->vfs

iyastreb: merci, fixed

src/ucs/vfs/sock/vfs_sock.c
```diff
@@ -27,23 +32,25 @@ typedef struct {
...
-                 "/tmp/ucx-vfs-%u.sock", euid);
+int ucs_vfs_sock_mkdir(const char *sock_path)
+{
+    char sock_path_dir[PATH_MAX];
+    int ret;
+
+    ucs_strncpy_safe(sock_path_dir, sock_path, sizeof(sock_path_dir));
+    ret = mkdir(dirname(sock_path_dir), S_IRWXU);
+    if ((ret < 0) && (errno != EEXIST)) {
+        ucs_diag("failed to create directory '%s': %m", sock_path_dir);
```
yosefe, line 49: maybe pass log_level as argument, so it will be ERROR when using from ucx_vfs tool?

iyastreb: ok, sure

src/ucs/vfs/fuse/vfs_fuse.c
```diff
@@ -280,6 +279,8 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
     }
 
     /* Watch for new files in 'watch_dir' */
+    ucs_strncpy_safe(watch_dir, path, sizeof(watch_dir));
+    dirname(watch_dir);
```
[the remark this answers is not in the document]
iyastreb: ok, you're right, fixed

src/ucs/vfs/fuse/vfs_fuse.c
```diff
@@ -278,13 +282,11 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
         goto out;
     }
 
-    /* Watch for new files in 'watch_dir' */
-    ucs_strncpy_safe(watch_dir, path, sizeof(watch_dir));
-    dirname(watch_dir);
+    /* Watch for new files in 'watch_dirname' */
```
yosefe, line 285: since we use watch_dirname only here, maybe move lines 257-261 to here

yosefe (approved):
pls squash

PR #10051, opened 2024-08-08
UCP/PROTO: Performance factors based proto selection

ivankochin:
## What
Currently MULTI mode performance calculated between several ranges passed to `_parallel_stages()` function and sum of CPU performance type. That patch changes that to another approach, when during the protocol initialization, all the overheads accumulated in parallel factors that can be overlapped one by another. And result MULTI performance prediction calculates as longest parallel performance factor.

## Why ?
To make MULTI mode performance prediction more precise.

## How ?
In that patch a protocol performance stored inside `ucp_proto_perf_t` structure which is a list of segments and each segment is a combination of performance factors, each of which represented as a linear function.


src/ucp/proto/proto_perf.h
```diff
@@ -144,6 +144,25 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+ */
+ucs_status_t ucp_proto_perf_aggregate2(const char *name,
+                                       const ucp_proto_perf_t *perf1,
+                                       const ucp_proto_perf_t *perf2,
+                                       ucp_proto_perf_t **perf_p);
+
+
+/**
+ * Create a proto perf structure that is equal to @a remote_perf but all
+ * local factors' values are turned to remote ones and vice versa.
```
rakhmets, line 158:
```suggestion
 * Create a proto perf structure based on @a remote_perf, converting the values 
 * of local factors to remote ones and vice versa.
```

src/ucp/proto/proto_perf.c
```diff
@@ -414,6 +414,67 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+    ucs_status_t status;
+
+    ucp_proto_perf_check(remote_perf);
+
+    status = ucp_proto_perf_create("remote", &perf);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    /* Turn local factors to remote and vice versa */
```
rakhmets, line 442: I'd prefer `convert` here.

src/ucp/proto/proto_perf.c
```diff
@@ -414,6 +414,67 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+        perf_factors[UCP_PROTO_PERF_FACTOR_LOCAL_MTYPE_COPY] =
+                remote_seg->perf_factors[UCP_PROTO_PERF_FACTOR_REMOTE_MTYPE_COPY];
+        perf_factors[UCP_PROTO_PERF_FACTOR_LATENCY] =
+                remote_seg->perf_factors[UCP_PROTO_PERF_FACTOR_LATENCY];
+        perf_factors[UCP_PROTO_PERF_FACTOR_REMOTE_MTYPE_COPY] =
+                remote_seg->perf_factors[UCP_PROTO_PERF_FACTOR_LOCAL_MTYPE_COPY];
+        perf_factors[UCP_PROTO_PERF_FACTOR_REMOTE_TL] =
+                remote_seg->perf_factors[UCP_PROTO_PERF_FACTOR_LOCAL_TL];
+        perf_factors[UCP_PROTO_PERF_FACTOR_REMOTE_CPU] =
+                remote_seg->perf_factors[UCP_PROTO_PERF_FACTOR_LOCAL_CPU];
```
rakhmets, line 457: How about using a static array of pairs?

src/ucp/proto/proto_common.c
```diff
@@ -255,9 +255,9 @@ void ucp_proto_common_lane_perf_node(ucp_context_h context,
             UCT_TL_RESOURCE_DESC_FMT, UCT_TL_RESOURCE_DESC_ARG(tl_rsc));
 
     if (perf_attr->field_mask & UCT_PERF_ATTR_FIELD_BANDWIDTH) {
-        ucp_proto_perf_node_add_bandwidth(perf_node, "bw/proc",
+        ucp_proto_perf_node_add_bandwidth(perf_node, "bw/dedicated",
                                           perf_attr->bandwidth.dedicated);
-        ucp_proto_perf_node_add_bandwidth(perf_node, "bw/node",
+        ucp_proto_perf_node_add_bandwidth(perf_node, "bw/shared",
```
rakhmets, line 260: Maybe in a separate PR?

src/ucp/proto/proto_common.c
```diff
@@ -646,12 +646,16 @@ ucp_lane_index_t ucp_proto_common_find_lanes_with_min_frag(
...
 {
-    ucp_proto_select_add_proto(&params->super, params->cfg_thresh,
-                               params->cfg_priority, proto_caps, priv,
-                               priv_size);
+    ucs_status_t status = ucp_proto_select_add_proto(&params->super,
+                                                     params->cfg_thresh,
+                                                     params->cfg_priority,
+                                                     perf, priv, priv_size);
+    if (status != UCS_OK) {
+        ucp_proto_perf_destroy(perf);
```
rakhmets, line 657: Please call destroy on the same level as create.

src/ucp/proto/proto_perf.c
ivankochin: Not sure that I understand you correctly. Do you mean create two enums one is for factors and another for side?

Like:
```
typedef enum {
  UCP_PROTO_PERF_SIDE_LOCAL,
  UCP_PROTO_PERF_SIDE_REMOTE,
  UCP_PROTO_PERF_SIDE_LAST
} ucp_proto_perf_side_t;

typedef enum {
    UCP_PROTO_PERF_FACTOR_CPU,
    UCP_PROTO_PERF_FACTOR_TL,
    UCP_PROTO_PERF_FACTOR_MTYPE_COPY,
    UCP_PROTO_PERF_FACTOR_LATENCY,
    UCP_PROTO_PERF_FACTOR_LAST
} ucp_proto_perf_factor_id_t;
```

And then store them inside segment as two dimensional array?

```
typedef ucs_linear_func_t ucp_proto_perf_factors_t[UCP_PROTO_PERF_FACTOR_LAST][UCP_PROTO_PERF_SIDE_LAST];
```

src/ucp/proto/proto_common.c
ivankochin: Please elaborate, do you suggest to return status from `ucp_proto_common_add_proto` and destroy perf on one level higher if `status != OK` ?

ivankochin: Done

src/ucp/proto/proto_common.c
ivankochin: Done.

src/ucp/proto/proto_perf.c
ivankochin: Done

src/ucp/proto/proto_perf.h
ivankochin: Done

src/ucp/proto/proto_perf.c
rakhmets: I mean two dimensional array: 3x2. And:
```C
for (i in size of array) {
    perf_factors[array[i][0]] = remote_seg->perf_factors[array[i][1]];
    perf_factors[array[i][1]] = remote_seg->perf_factors[array[i][0]];
}

perf_factors[UCP_PROTO_PERF_FACTOR_LATENCY] = remote_seg->perf_factors[UCP_PROTO_PERF_FACTOR_LATENCY];
```

src/ucp/proto/proto_common.c
rakhmets: Yes, exactly.
`ucp_proto_common_add_proto` is just a wrapper of `ucp_proto_select_add_proto`. It is better to change the signature of the function to return the status.
The current implementation will work only because `ucp_proto_common_add_proto` is called in the end of a caller function. But it may lead to an error, when the caller function will try to use `perf` after it was destroyed inside `ucp_proto_common_add_proto`.

brminich: btw, do we really need this wrapper then?

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,251 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+                                      ucs_popcount(reg_md_map));
+    ucp_proto_perf_node_deref(&reg_perf_node);
+    return status;
+}
+
+static ucp_proto_perf_factor_id_t
+ucp_proto_buffer_copy_get_cpu_factor_id(int local) {
+    return (local) ? UCP_PROTO_PERF_FACTOR_LOCAL_CPU :
+                     UCP_PROTO_PERF_FACTOR_REMOTE_CPU;
+}
```
brminich, line 325: add new line

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,251 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+    status = ucp_proto_perf_add_funcs(perf, range_start, range_end,
+                                      perf_factors, reg_perf_node,
+                                      perf_node_name, "%u mds",
+                                      ucs_popcount(reg_md_map));
+    ucp_proto_perf_node_deref(&reg_perf_node);
+    return status;
+}
+
+static ucp_proto_perf_factor_id_t
+ucp_proto_buffer_copy_get_cpu_factor_id(int local) {
```
brminich, line 322: i'd rename it to
```suggestion
ucp_proto_buffer_copy_cpu_factor_id(int local) {
```

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,251 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+    return (local) ? UCP_PROTO_PERF_FACTOR_LOCAL_CPU :
+                     UCP_PROTO_PERF_FACTOR_REMOTE_CPU;
+}
+static ucp_proto_perf_factor_id_t
+ucp_proto_buffer_copy_get_factor_id(ucs_memory_type_t local_mem_type,
```
brminich, line 327: i'd rename it to 
```suggestion
ucp_proto_buffer_copy_factor_id(ucs_memory_type_t local_mem_type,
```

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,251 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+                                    ucs_memory_type_t remote_mem_type,
+                                    uct_ep_operation_t memtype_op,
+                                    int local)
+{
+    int h2h      = UCP_MEM_IS_HOST(local_mem_type) &&
+                   UCP_MEM_IS_HOST(remote_mem_type);
+    /* RNDV mtype protocols which do async copy set ZCOPY as `memtype_op`
+     * while eager procols that imply blocking copy set SHORT */
+    int blocking = (memtype_op == UCT_EP_OP_GET_SHORT) ||
```
brminich, line 336: there is the same check on line 437 in this file. Maybe create an inline function for it?

src/ucp/proto/proto_init.c
```diff
@@ -698,66 +534,69 @@ ucp_proto_common_check_mem_access(const ucp_proto_common_init_params_t *params)
...
-    return UCS_OK;
+    if (range_end > 0) {
+        /* Add buffer copy/register cost for non-empty messages */
+        status = ucp_proto_init_add_buffer_perf(params, ucs_max(1, range_start),
+                                                range_end, reg_md_map, perf);
+        if (status != UCS_OK) {
+            goto out;
+        }
+
+        /* Add range that represent sending many fragments */
```
brminich, line 582:
```suggestion
        /* Add range that represents sending many fragments */
```

src/ucp/proto/proto_common.c
ivankochin: Removed

src/ucp/proto/proto.h
```diff
@@ -203,20 +127,20 @@ typedef struct {
...
-    char   desc[UCP_PROTO_DESC_STR_MAX];
+    char                           desc[UCP_PROTO_DESC_STR_MAX];
 
     /* Protocol configuration in the range, such as devices and transports */
-    char   config[UCP_PROTO_CONFIG_STR_MAX];
+    char                           config[UCP_PROTO_CONFIG_STR_MAX];
 
     /* Map of used lanes */
-    ucp_lane_map_t lane_map;
+    ucp_lane_map_t                 lane_map;
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/proto/proto_debug.c
```diff
@@ -445,28 +429,15 @@ void ucp_proto_config_info_str(ucp_worker_h worker,
...
+    if (range == NULL || range->start > msg_length) {
+        ucs_string_buffer_appendf(strb, " - not available");
         return;
     }
 
-    /* Find the relevant performance range */
-    range     = ucp_proto_perf_range_search(select_elem, msg_length);
-    bandwidth = ucp_proto_select_calc_bandwidth(&proto_config->select_param,
-                                                range, msg_length);
+    bandwidth = msg_length / ucs_linear_func_apply(range->value, msg_length);
```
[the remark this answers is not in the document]
ivankochin: I found couple of places where we are doing same operations to calculate BW and it seems like in all of these places can divide result by 0. But since all these places use this result for printing, and result of 0 division would be INF I think that is OK ("inf" would be printed, which is basically true in that case. see https://godbolt.org/z/66cqTo7YT).

If we want to consider infinite BW as an issue, we can add assert to catch such cases. WDYT?

src/ucp/proto/proto_init.c
ivankochin: These checks are not the same. Check on 437 is:
```c
    if ((memtype_op == UCT_EP_OP_GET_SHORT) ||
        (memtype_op == UCT_EP_OP_GET_ZCOPY)) {
```

And it check whether operation PUT or GET, while this check defines whether operation blocking or not

src/ucp/proto/proto_debug.c
```diff
@@ -445,28 +429,15 @@ void ucp_proto_config_info_str(ucp_worker_h worker,
...
-    ucs_assert(new_key_cfg_index == proto_config->rkey_cfg_index);
-    select_elem = ucp_proto_select_lookup_slow(worker, proto_select, 1,
-                                               proto_config->ep_cfg_index,
-                                               proto_config->rkey_cfg_index,
-                                               &proto_config->select_param);
-    if (select_elem == NULL) {
+    /* Find the relevant performance range */
+    range = ucp_proto_flat_perf_find_lb(proto_config->init_elem->flat_perf,
+                                        msg_length);
+    if (range == NULL || range->start > msg_length) {
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/proto/proto_debug.c
```diff
@@ -1007,39 +992,43 @@ ucp_proto_select_write_info(ucp_worker_h worker,
...
+            range_start_str, range_end_str);
+
+    ucs_snprintf_safe(file_name, sizeof(file_name), "%s/%s_%s_%s_%s.dot",
+                      dir_path, ucs_string_buffer_cstr(&ep_cfg_strb),
+                      ucs_string_buffer_cstr(&sel_param_strb), range_start_str,
+                      range_end_str);
+
+    UCS_DYNAMIC_BITMAP_FOR_EACH_BIT(proto_idx, proto_mask) {
+        proto     = &ucs_array_elem(&proto_init->protocols, proto_idx);
+        range     = ucp_proto_flat_perf_find_lb(proto->flat_perf, range_start);
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/rndv/proto_rndv.c
```diff
@@ -185,20 +185,26 @@ static ucs_status_t ucp_proto_rndv_ctrl_select_remote_proto(
...
 
+    status = ucp_proto_perf_create("rndv_ctrl", &perf);
+    if (status != UCS_OK) {
+        return status;
+    }
+
     if (lane == UCP_NULL_LANE) {
-        *send_time = *receive_time = 0;
-        return UCS_OK;
+        goto out;
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/rndv/proto_rndv.c
```diff
@@ -208,17 +214,33 @@ ucp_proto_rndv_ctrl_perf(const ucp_proto_init_params_t *params,
...
+    perf_factors[UCP_PROTO_PERF_FACTOR_REMOTE_CPU] =
+            ucs_linear_func_make(perf_attr.recv_overhead, 0.0);
+    perf_factors[UCP_PROTO_PERF_FACTOR_LATENCY] = ucs_linear_func_make(
+            ucp_tl_iface_latency(worker->context, &perf_attr.latency), 0.0);
+
+    status = ucp_proto_perf_add_funcs(perf, init_params->min_length,
+                                      init_params->max_length, perf_factors,
+                                      NULL, name, "");
+    if (status != UCS_OK) {
+        return status;
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/rndv/proto_rndv.c
```diff
@@ -278,115 +300,128 @@ ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+        ucp_proto_init_elem_t *remote_proto, const void *remote_priv)
 {
-    const char *variant_name = ucp_proto_perf_node_name(remote_range->node);
-    const char *proto_name   = ucp_proto_id_field(params->super.super.proto_id,
-                                                  name);
-    const ucp_proto_perf_range_t *parallel_stages[2];
-    UCS_STRING_BUFFER_ONSTACK(agg_buf, 256);
-    size_t range_max_length;
+    const char *proto_name = ucp_proto_id_field(remote_proto->proto_id, name);
+    size_t num_elems  = 0;
```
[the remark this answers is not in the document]
ivankochin: Done

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -665,7 +665,8 @@ UCS_TEST_F(test_proto_perf, single_func)
     expect_empty_range(2000, SIZE_MAX);
 }
 
-UCS_TEST_F(test_proto_perf, to_inf) {
+UCS_TEST_F(test_proto_perf, to_inf)
+{
```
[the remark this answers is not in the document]
ivankochin: Removed

src/ucp/proto/proto_init.c
ivankochin: Done

src/ucp/proto/proto_init.c
ivankochin: Done

src/ucp/proto/proto_init.c
ivankochin: Done

src/ucp/proto/proto_perf.c
ivankochin: I found several ways how to do this operation instead of current way (option 0) and want to discuss with you what is the best option from your point of view. 

Let's start with your suggestion (option 1), if I don't miss something, it would look like the following:

```c
    static const size_t factors_to_convert = 3;
    ucp_proto_perf_factor_id_t convertaion_map[][2] = {
        {UCP_PROTO_PERF_FACTOR_LOCAL_CPU, UCP_PROTO_PERF_FACTOR_REMOTE_CPU},
        {UCP_PROTO_PERF_FACTOR_LOCAL_TL, UCP_PROTO_PERF_FACTOR_REMOTE_TL},
        {UCP_PROTO_PERF_FACTOR_LOCAL_MTYPE_COPY,
         UCP_PROTO_PERF_FACTOR_REMOTE_MTYPE_COPY}
    };

    ucp_proto_perf_segment_foreach(remote_seg, remote_perf) {
        for (factor_idx = 0; factor_idx < factors_to_convert; ++factor_idx) {
            perf_factors[convertaion_map[factor_idx][0]] =
                    remote_seg->perf_factors[convertaion_map[factor_idx][1]];
            perf_factors[convertaion_map[factor_idx][1]] =
                    remote_seg->perf_factors[convertaion_map[factor_idx][0]];
        }
        perf_factors[UCP_PROTO_PERF_FACTOR_LATENCY] =
                remote_seg->perf_factors[UCP_PROTO_PERF_FACTOR_LATENCY];
        ...
    }
```

FMPOV it looks better inside the loop but I don't really like separate `factors_to_convert` variable. Another option (option 2) with defining `convertation_map` is using `ucs_swap`:

```c
    static const size_t factors_to_convert = 3;
    ucp_proto_perf_factor_id_t convertaion_map[][2] = {
        {UCP_PROTO_PERF_FACTOR_LOCAL_CPU, UCP_PROTO_PERF_FACTOR_REMOTE_CPU},
        {UCP_PROTO_PERF_FACTOR_LOCAL_TL, UCP_PROTO_PERF_FACTOR_REMOTE_TL},
        {UCP_PROTO_PERF_FACTOR_LOCAL_MTYPE_COPY,
         UCP_PROTO_PERF_FACTOR_REMOTE_MTYPE_COPY}
    };

    ucp_proto_perf_segment_foreach(remote_seg, remote_perf) {
        /* copy segments */
        for (fid = 0; fid < UCP_PROTO_PERF_FACTOR_LAST; fid++) {
            perf_factors[fid] = remote_seg->perf_factors[fid];
        }
        
        /* convert segments */
        for (factor_idx = 0; factor_idx < factors_to_convert; ++factor_idx) {
            ucs_swap(&perf_factors[convertaion_map[factor_idx][0]],
                     &perf_factors[convertaion_map[factor_idx][1]]);
        }
        ...
    }
```

Also I want to propose another solution (option 3) with `ucs_swap` but without `convertaion_map`:
```c
    ucp_proto_perf_segment_foreach(remote_seg, remote_perf) {
        /* copy segments */
        for (fid = 0; fid < UCP_PROTO_PERF_FACTOR_LAST; fid++) {
            perf_factors[fid] = remote_seg->perf_factors[fid];
        }
        
        /* convert segments */
        for (factor_idx = 0; factor_idx < factors_to_convert; ++factor_idx) {
            ucs_swap(&perf_factors[UCP_PROTO_PERF_FACTOR_LOCAL_CPU],
                     &perf_factors[UCP_PROTO_PERF_FACTOR_REMOTE_CPU]);
            ucs_swap(&perf_factors[UCP_PROTO_PERF_FACTOR_LOCAL_TL],
                     &perf_factors[UCP_PROTO_PERF_FACTOR_REMOTE_TL]);
            ucs_swap(&perf_factors[UCP_PROTO_PERF_FACTOR_LOCAL_MTYPE_COPY],
                     &perf_factors[UCP_PROTO_PERF_FACTOR_REMOTE_MTYPE_COPY]);
        }
        ...
    }
```

What do you think would be the best for this place?

P.S. I checked compilation of all these option, but didn't check correctness of the logic on practice. Despite this fact I think all of them should be valid.

rakhmets: I think the first one is the simplest one. You can use `ucs_static_array_size` to avoid extra variable.

ivankochin: Done

src/ucp/proto/proto_init.c
ivankochin: Done

src/ucs/sys/topo/base/topo.c
rakhmets, line 1: Seems like it should be in another PR.
The PR is already big enough. Let's try to reduce it as much as possible.

test/gtest/ucp/test_ucp_mmap.cc
rakhmets, line 1: The same as above.

src/ucp/proto/proto_debug.h
```diff
@@ -62,8 +53,8 @@ void ucp_proto_select_perf_str(const ucs_linear_func_t *perf, char *time_str,
                                size_t bw_str_max);
 
 
-void ucp_proto_select_init_trace_caps(const ucp_proto_init_params_t *init_params,
-                                      const ucp_proto_caps_t *proto_caps,
+void ucp_proto_select_init_trace_perf(const ucp_proto_init_params_t *init_params,
```
rakhmets, line 56: `ucp_proto_debug_...` ?

src/ucp/proto/proto_debug.h
```diff
@@ -100,6 +91,11 @@ ucp_proto_perf_node_new(ucp_proto_perf_node_type_t type,
                         unsigned selected_child, const char *name,
                         const char *desc_fmt, va_list ap);
 
+ucp_proto_perf_node_t *
+ucp_proto_perf_node_new(ucp_proto_perf_node_type_t type,
+                        unsigned selected_child, const char *name,
+                        const char *desc_fmt, va_list ap);
```
rakhmets, line 97:
```suggestion
```

src/ucs/memory/rcache.c
rakhmets, line 1: Why do we need these changes in this PR?

src/ucp/proto/proto_debug.h
```diff
@@ -172,4 +168,12 @@ void ucp_proto_select_elem_trace(ucp_worker_h worker,
                                  ucp_proto_select_elem_t *select_elem);
 
 
+void ucp_proto_select_write_info(
```
rakhmets, line 171: `ucp_proto_debug_...`

src/ucp/proto/proto_init.h
```diff
@@ -68,44 +49,40 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
- * @param [in]    proto_name    Protocol name, for debugging.
- * @param [in]    range_start   Range interval start.
- * @param [in]    range_end     Range interval end.
- * @param [in]    bias          Performance bias (0 - no bias).
- * @param [in]    stages        Array of parallel stages performance ranges.
- * @param [in]    num_stages    Number of parallel stages in the protocol.
- * @param [inout] caps          Filled with protocol performance data.
- */
+ucp_proto_common_init_params_t
+ucp_proto_common_params_init(const ucp_proto_init_params_t *init_params);
```
rakhmets, line 53: Maybe `ucp_proto_common_init_params`, and move it to `ucp_proto_common.h`

src/ucp/proto/proto_init.h
```diff
@@ -68,44 +49,40 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+
+
 ucs_status_t
-ucp_proto_init_parallel_stages(const char *proto_name, size_t range_start,
-                               size_t range_end, double bias,
-                               const ucp_proto_perf_range_t **stages,
-                               unsigned num_stages, ucp_proto_caps_t *caps);
+ucp_proto_common_add_ppln_perf(ucp_proto_perf_t *perf,
```
rakhmets, line 57: It seems it should be in `ucp_perf.h`, and renamed to `ucp_perf_...`.

src/ucp/proto/proto_init.h
```diff
@@ -68,44 +49,40 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+                                    ucs_memory_type_t local_mem_type,
+                                    ucs_memory_type_t remote_mem_type,
+                                    uct_ep_operation_t memtype_op,
+                                    size_t range_start, size_t range_end,
+                                    int local, ucp_proto_perf_t *perf);
 
 
 ucs_status_t
-ucp_proto_common_init_caps(const ucp_proto_common_init_params_t *params,
+ucp_proto_common_init_perf(const ucp_proto_common_init_params_t *params,
```
rakhmets, line 81: It seems this method is also in the wrong place.
It should be `ucp_proto_perf_init`, and moved to `ucp_proto_perf.h`.

src/ucs/memory/rcache.c
ivankochin: Coverity suddenly started to show double-lock/unlock errors for this PR despite the fact that these code was no affected and there is obviously no double lock/unlock...

This changed are intended to suppress these issues (that didn't help, need to find new way to suppress them).

src/ucs/sys/topo/base/topo.c
ivankochin: These changes were initially introduced in @yosefe [PR](https://github.com/openucx/ucx/pull/9944/files#diff-a5f27f8a976d15fc19f888549386fdcf7f26f672b657710be7e7bcf94c211cdbR88) that was an origin PR for these chagnes. @yosefe do you remember why did you do these changes and whether they are necessary for this PR?

src/ucp/proto/proto_debug.h
ivankochin: There are many functions in `proto_debug.h` which don't have `ucp_proto_debug_` prefix. Even function that was in that place before didn't have it. Why it should be applied here?

rakhmets: I suggested this name, since you have changed it. So, I thought it would be better to start to use the proper naming.
However, I now have doubts about the name. On one hand, I would prefer to use the common prefix for the methods declared in the same file. On the other hand, there is not a single method named according to this pattern in this file.

src/ucp/proto/proto_debug.h
ivankochin: Done

src/ucp/proto/proto_debug.h
ivankochin: Let's make a common decision for all such cases in [that conversation](https://github.com/openucx/ucx/pull/10051#discussion_r1734616108)

src/ucp/proto/proto_init.h
ivankochin: I am agree that we can consider renaming this function to meet header file prefix (e.g. `ucp_proto_init_perf`) but disagree that it should be part of `proto_perf.h` file. I think that this function on the right place since it contain protocols related logic of extending perf structure .

FMPOV `proto_perf.h` file was designed to contain `ucp_proto_perf_t` structure interfaces which is unrelated to certain protocols logic.

I agree to move `ppln` (your previous comment) logic to `proto_perf.h` file since to implement this logic we need only entities defined in that header file. But implementing `ucp_proto_common_init_perf` requires `ucp_proto_common_init_params_t` and `ucp_proto_common_tl_perf_t` which I am tend consider as protocol-specific stuff.

src/ucp/proto/proto_init.h
ivankochin: Done

src/ucp/proto/proto_init.h
ivankochin: Done

src/ucp/proto/proto_select.c
```diff
@@ -89,23 +74,26 @@ static ucs_status_t ucp_proto_thresholds_next_range(
...
-        proto = &ucs_array_elem(&proto_init->protocols, proto_idx);
-        if (msg_length < proto->caps.min_length) {
-            ucs_trace(
-                    "skipping proto %s with min_length %zu for msg_length %zu",
-                    ucp_proto_id_field(proto->proto_id, name),
-                    proto->caps.min_length, msg_length);
-            max_length = ucs_min(max_length, proto->caps.min_length - 1);
+        proto      = &ucs_array_elem(&proto_init->protocols, proto_idx);
+        proto_name = ucp_proto_id_field(proto->proto_id, name);
+        ucs_assert_always(proto_name != NULL);
```
yosefe, line 79: really needed?

src/ucp/proto/proto_select.c
```diff
@@ -709,16 +630,35 @@ void ucp_proto_select_add_proto(const ucp_proto_init_params_t *init_params,
...
 
     memset(init_elem, 0, sizeof(*init_elem));
     init_elem->proto_id     = proto_id;
     init_elem->priv_offset  = priv_offset;
     init_elem->cfg_thresh   = cfg_thresh;
     init_elem->cfg_priority = cfg_priority;
-    init_elem->caps         = *proto_caps;
+    init_elem->perf         = perf;
+
+    if (ucp_proto_select_op_attr_unpack(init_params->select_param->op_attr) &
```
yosefe, line 643: maybe add local var op_attr_flags?

src/ucp/proto/proto_select.c
```diff
@@ -747,8 +687,9 @@ void ucp_proto_select_short_init(ucp_worker_h worker,
     ssize_t max_short_signed;
     const uint32_t *op_attribute;
 
+    /* Initialize to disabled state to make coverity happy */
+    ucp_proto_select_short_disable(proto_short);
```
yosefe, line 691: so can we remove ucp_proto_select_short_disable from the end of this function?

src/ucp/rndv/proto_rndv.c
```diff
@@ -278,115 +303,134 @@ ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+                              params->ctrl_msg_name,
+                              ucp_proto_perf_name(remote_proto->perf));
+    ucs_trace("%s proto variant: cfg_thresh=%zu cfg_priority=%u", variant_name,
+              remote_proto->cfg_thresh, remote_proto->cfg_priority);
 
-        ucs_trace("%s: max %zu remote-op %s" UCP_PROTO_PERF_FUNC_TYPES_FMT,
-                  proto_name, remote_range->max_length, variant_name,
-                  UCP_PROTO_PERF_FUNC_TYPES_ARG(remote_range->perf));
+    /* Set send_overheads to the time to send and receive RTS message */
+    overhead = ucs_linear_func_make(params->super.overhead * 2, 0.0);
```
yosefe, line 345: maybe the RTS receive overhead should be part of REMOTE_CPU and not the LOCAL_CPU?

src/ucp/rndv/proto_rndv.c
```diff
@@ -278,115 +303,134 @@ ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params,
...
-        min_length = range_max_length + 1;
-    } while ((remote_range++)->max_length < max_length);
+    status = ucp_proto_perf_remote(remote_proto->perf, &remote_perf);
+    if (status != UCS_OK) {
+        goto out_destroy_ctrl_perf;
+    }
 
-    return UCS_OK;
-}
+    perf_elems[num_elems++]     = ctrl_perf;
```
yosefe, line 366: maybe use ucs_array (onstack) for perf_elems?

src/ucp/rndv/proto_rndv.c
```diff
@@ -278,115 +303,134 @@ ucp_proto_rndv_ctrl_init_priv(const ucp_proto_rndv_ctrl_init_params_t *params,
...
+    status = ucp_proto_select_add_proto(&params->super.super, cfg_thresh,
+                                        cfg_priority, perf, rpriv, priv_size);
+    if (status == UCS_OK) {
+        /* `perf` ownership was moved to `init_params->ctx`, so it shouldn't be
+         * destroyed now */
+        goto out_destroy_remote_perf;
+    }
+
+out_destroy_perf:
+    ucp_proto_perf_destroy(perf);
```
yosefe, line 417: this looks a bit weird (not following the convention)

src/ucp/rndv/proto_rndv.c
```diff
@@ -717,13 +637,26 @@ ucp_proto_rndv_bulk_init(const ucp_proto_multi_init_params_t *init_params,
...
                                   mpriv->align_thresh + mpriv->min_frag);
 
-    /* Add ack latency */
-    status = ucp_proto_rndv_ack_init(&init_params->super.super, ack_name,
-                                     &multi_caps,
-                                     ucs_linear_func_make(150e-9, 0),
-                                     &rpriv->super, caps);
-    ucp_proto_select_caps_cleanup(&multi_caps);
+    status = ucp_proto_rndv_ack_init(&init_params->super, ack_name,
+                                     ucs_linear_func_make(50e-9, 0),
```
yosefe, line 641: overhead changed from 150e-9 to 50e-9, is that intentional?

src/ucp/rndv/proto_rndv.c
```diff
@@ -717,13 +637,26 @@ ucp_proto_rndv_bulk_init(const ucp_proto_multi_init_params_t *init_params,
...
+        goto out_destroy_bulk_perf;
+    }
+
+    if (rpriv->super.lane == UCP_NULL_LANE) {
+        /* Add perf without ACK in case of pipeline */
+        *perf_p = bulk_perf;
+        return UCS_OK;
+    }
+
+    perf_name = ucp_proto_id_field(init_params->super.super.proto_id, name);
```
yosefe, line 653: proto_name

src/ucp/rndv/rndv_ats.c
```diff
@@ -13,45 +13,39 @@
...
 
-    status = ucp_proto_rndv_ack_init(init_params, UCP_PROTO_RNDV_ATS_NAME,
-                                     &input_caps, UCS_LINEAR_FUNC_ZERO, &priv,
-                                     &caps);
-    ucp_proto_select_caps_cleanup(&input_caps);
-
-    if (status != UCS_OK) {
+    status = ucp_proto_rndv_ack_init(&params, UCP_PROTO_RNDV_ATS_NAME,
+                                     UCS_LINEAR_FUNC_ZERO, &perf, &priv);
+    if ((status != UCS_OK) || (priv.lane == UCP_NULL_LANE)) {
```
yosefe, line 39: if priv.lane == UCP_NULL_LANE, shall we release perf? or assert it is NULL?

src/ucp/rndv/rndv_get.c
```diff
@@ -71,13 +71,19 @@ ucp_proto_rndv_get_common_probe(const ucp_proto_init_params_t *init_params,
...
 
     priv_size = UCP_PROTO_MULTI_EXTENDED_PRIV_SIZE(&rpriv, mpriv);
-    ucp_proto_common_add_proto(&params.super, &caps, &rpriv, priv_size);
+    status    = ucp_proto_select_add_proto(&params.super.super,
+                                           params.super.cfg_thresh,
+                                           params.super.cfg_priority, perf,
+                                           &rpriv, priv_size);
+    if (status != UCS_OK) {
+        ucp_proto_perf_destroy(perf);
+    }
```
yosefe, line 86: maybe should add a wrapper to destroy perf if failed to add?
or maybe ucp_proto_select_add_proto should return void and destroy perf itself if failed to add?
since this construct repeats itself in multiple places

src/ucp/rndv/rndv_ppln.c
```diff
@@ -85,52 +85,81 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
...
-                ppln_overhead, &rpriv.ack, &caps);
-        if (status == UCS_OK) {
-            ucp_proto_select_add_proto(init_params, proto->cfg_thresh,
-                                       proto->cfg_priority,
-                                       &caps, &rpriv, sizeof(rpriv));
+        ppln_overhead = ucs_linear_func_make(frag_overhead,
+                                             frag_overhead / rpriv.frag_size);
+        status        = ucp_proto_rndv_ack_init(&ack_params,
+                                                UCP_PROTO_RNDV_ATS_NAME,
+                                                ppln_overhead, &ack_perf,
```
yosefe, line 139: seems weird the overhead of pipelining is part of the "ATS" node, i would expect it will be in the same level as the ATS

src/ucp/rndv/rndv_ppln.c
```diff
@@ -85,52 +85,81 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
...
+                                                &rpriv.ack);
+        if ((status != UCS_OK) || (rpriv.ack.lane == UCP_NULL_LANE)) {
+            goto out_destroy_ppln_perf;
         }
-        ucp_proto_select_caps_cleanup(&ppln_caps);
+
+        status = ucp_proto_perf_aggregate2(
+                ucp_proto_id_field(init_params->proto_id, name),
```
yosefe, line 146: shall we add local var proto_name? it is used in several places in this function

src/ucp/proto/proto_select.c
```diff
@@ -339,41 +291,23 @@ static ucs_status_t ucp_proto_select_elem_add_envelope(
...
-        UCS_DYNAMIC_BITMAP_FOR_EACH_BIT(child_proto_idx, proto_mask) {
-            proto       = &ucs_array_elem(&proto_init->protocols,
-                                          child_proto_idx);
-            child_range = ucp_proto_caps_range_find(&proto->caps, range_start);
-            ucp_proto_perf_node_add_child(range->node, child_range->node);
+        /* Print detailed protocol selection data to a user-configured path */
+        if (!internal && !ucs_string_is_empty(proto_info_dir)) {
+            ucp_proto_select_write_info(worker, proto_init, proto_mask,
+                                        proto_idx, &thresh_elem->proto_config,
+                                        range_start, envelope_elem->max_length);
```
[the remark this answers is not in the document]
ivankochin: That was changed because in the master we create select perf node and store it in structure array called `ucp_proto_ranges_t` which was a part of `ucp_proto_select_elem_t`. Now there is no `ucp_proto_perf_range_t` structure since there is no `ucp_proto_caps_t` structure and therefore we do not have place to store selection perf nodes together with `max_length` that allows us to build performance graph. 

That's why I decided to build selection perf node on flight during `_add_envelope` function, draw the perf graph for some certain range and then deref this selection perf node. Since for that I need to obtain `proto_mask` for certain ranges I decided to move it here.

We still can create selection perf nodes in that place, store them together with meta information in some container (like it was done with `ucp_proto_ranges_t` and propagate all this information to `ucp_proto_select_elem_trace` but I am not sure that this worth it.

test/gtest/ucp/test_ucp_proto.cc
```diff
@@ -247,15 +247,13 @@ UCS_TEST_P(test_perf_node, basic)
...
-    ucp_proto_perf_node_add_child(n3, NULL);
 
     ucp_proto_perf_node_t *tmp = NULL;
-    ucp_proto_perf_node_own_child(n3, &tmp);
+    ucp_proto_perf_node_own_child(n1, &tmp);
 
     /* NULL parent is ignored */
     ucp_proto_perf_node_add_child(NULL, n1);
     ucp_proto_perf_node_add_child(NULL, n2);
-    ucp_proto_perf_node_add_child(NULL, n3);
```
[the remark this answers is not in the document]
ivankochin: Since there is no `ucp_proto_init_parallel_stages`, I completely removed the code related to `compose` performance node.

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,176 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+                                      perf_node_name, "%u mds",
+                                      ucs_popcount(reg_md_map));
+    ucp_proto_perf_node_deref(&reg_perf_node);
+    return status;
+}
+
+static ucp_proto_perf_factor_id_t
+ucp_proto_buffer_copy_cpu_factor_id(int local) {
+    return (local) ? UCP_PROTO_PERF_FACTOR_LOCAL_CPU :
+                     UCP_PROTO_PERF_FACTOR_REMOTE_CPU;
```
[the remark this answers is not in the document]
ivankochin: 2,3 - ok, thanks

1 - I am OK with moving `ucp_proto_buffer_copy_cpu_factor_id` to `proto_perf.h` since it doesn't relies on other entities like memory ops and types but I think that it is better to leave `ucp_proto_buffer_copy_factor_id` on the current place. In that PR I tried to place perf structure common interfaces to `proto_perf.h` and place all the logic of defining perf for certain proto in `proto_init.c`, WDYT?

src/ucp/proto/proto_init.c
```diff
@@ -698,66 +458,67 @@ ucp_proto_common_check_mem_access(const ucp_proto_common_init_params_t *params)
...
+                                                range_end, reg_md_map, perf);
+        if (status != UCS_OK) {
+            goto out;
+        }
+
+        /* Add range that represents sending many fragments */
+        if ((range_end < params->max_length) &&
+            !(params->flags & UCP_PROTO_COMMON_INIT_FLAG_SINGLE_FRAG)) {
+            frag_seg = ucp_proto_perf_segment_last(perf);
+            status   = ucp_perf_add_ppln(perf, frag_seg, params->max_length);
```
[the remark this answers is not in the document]
ivankochin: We can, but for RNDV pipeline we still need to find fragment segment on the outer level since we define size of fragment as `ucp_proto_perf_segment_end(frag_seg)`.

src/ucp/proto/proto_perf.c
```diff
@@ -414,6 +419,113 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+                               size_t max_length)
+{
+    ucp_proto_perf_factors_t factors = UCP_PROTO_PERF_FACTORS_INITIALIZER;
+    ucp_proto_perf_factor_id_t factor_id, max_factor_id;
+    ucs_linear_func_t factor_func;
+    double max_value;
+    size_t frag_size;
+    char frag_str[64];
+
+    ucs_assert(frag_seg != NULL);
```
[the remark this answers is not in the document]
ivankochin: AFAIR I placed it here after catching some bug, so yes, I think it is useful

src/ucp/proto/proto_common.c
```diff
@@ -16,6 +16,21 @@
 #include <uct/api/v2/uct_v2.h>
 
 
+ucp_proto_common_init_params_t
+ucp_proto_common_init_params(const ucp_proto_init_params_t *init_params)
+{
+    ucp_proto_common_init_params_t params = {
+        .super         = *init_params,
```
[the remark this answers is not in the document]
ivankochin: Initialized all the fields explicitly

src/ucp/proto/proto_select.c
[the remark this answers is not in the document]
ivankochin: Added TODO

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,176 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
-                  UCP_PROTO_PERF_FUNC_TYPES_ARG((*stage_elem)->perf),
-                  UCP_PROTO_PERF_FUNC_ARG(perf_elem));
+    if (/* Don't care about receiver time for one-sided remote access */
+        (params->flags & UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS) ||
+        /* Count only send completion time without waiting for a response */
+        ((op_attr_mask & UCP_OP_ATTR_FLAG_FAST_CMPL) &&
+         !(params->flags & UCP_PROTO_COMMON_INIT_FLAG_RESPONSE))) {
+        return UCS_LINEAR_FUNC_ZERO;
     }
+    return ucs_linear_func_make(tl_perf->recv_overhead, 0);
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,176 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
-                  UCP_PROTO_PERF_FUNC_FMT,
-                  stage_elem - stages,
-                  ucp_proto_perf_node_name((*stage_elem)->node),
-                  UCP_PROTO_PERF_FUNC_TYPES_ARG((*stage_elem)->perf),
-                  UCP_PROTO_PERF_FUNC_ARG(perf_elem));
+    if (/* Don't care about receiver time for one-sided remote access */
+        (params->flags & UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS) ||
+        /* Count only send completion time without waiting for a response */
+        ((op_attr_mask & UCP_OP_ATTR_FLAG_FAST_CMPL) &&
+         !(params->flags & UCP_PROTO_COMMON_INIT_FLAG_RESPONSE))) {
```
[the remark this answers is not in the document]
ivankochin: United

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,176 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
-    *perf_elem = sum_cpu_perf;
+static ucs_status_t
+ucp_proto_init_add_tl_perf(const ucp_proto_common_init_params_t *params,
+                           const ucp_proto_common_tl_perf_t *tl_perf,
+                           ucp_proto_perf_node_t *const tl_perf_node,
+                           size_t range_start, size_t range_end,
+                           ucp_proto_perf_t *perf)
+{
+    ucp_proto_perf_factors_t perf_factors = UCP_PROTO_PERF_FACTORS_INITIALIZER;
+    const double send_overhead = tl_perf->send_pre_overhead +
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,176 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+ucp_proto_buffer_copy_cpu_factor_id(int local) {
+    return (local) ? UCP_PROTO_PERF_FACTOR_LOCAL_CPU :
+                     UCP_PROTO_PERF_FACTOR_REMOTE_CPU;
+}
+
+static ucp_proto_perf_factor_id_t
+ucp_proto_buffer_copy_factor_id(ucs_memory_type_t local_mem_type,
+                                    ucs_memory_type_t remote_mem_type,
+                                    uct_ep_operation_t memtype_op,
+                                    int local)
```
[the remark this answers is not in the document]
ivankochin: Fixed

src/ucp/proto/proto_init.c
```diff
@@ -196,180 +115,176 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+    int blocking = (memtype_op == UCT_EP_OP_GET_SHORT) ||
+                   (memtype_op == UCT_EP_OP_PUT_SHORT);
+    if (h2h || blocking) {
+        return ucp_proto_buffer_copy_cpu_factor_id(local);
+    }
+
+    ucs_assertv((memtype_op == UCT_EP_OP_GET_ZCOPY) ||
+                (memtype_op == UCT_EP_OP_PUT_ZCOPY), "memtype_op=%d",
+                memtype_op);
+    return (local) ? UCP_PROTO_PERF_FACTOR_LOCAL_MTYPE_COPY :
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/proto/proto_init.c
```diff
@@ -437,11 +348,13 @@ ucp_proto_init_buffer_copy_time(ucp_worker_h worker, const char *title,
...
-    ucs_assert(perf_attr.recv_overhead < 1e-15);
-    copy_time->c = ucp_tl_iface_latency(context, &perf_attr.latency) +
-                   perf_attr.send_pre_overhead + perf_attr.send_post_overhead +
-                   perf_attr.recv_overhead;
-    copy_time->m = 1.0 / ucp_tl_iface_bandwidth(context, &perf_attr.bandwidth);
+    ucs_assert(perf_attr.recv_overhead < UCP_PROTO_PERF_EPSILON);
+
+    perf_factors[ucp_proto_buffer_copy_cpu_factor_id(local)].c =
+            perf_attr.send_pre_overhead + perf_attr.send_post_overhead;
+    perf_factors[factor_id] = ucs_linear_func_make(
```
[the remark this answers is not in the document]
ivankochin: Good catch, fixed, thanks!

src/ucp/proto/proto_multi.c
```diff
@@ -260,17 +263,25 @@ size_t ucp_proto_multi_priv_size(const ucp_proto_multi_priv_t *mpriv)
 
 void ucp_proto_multi_probe(const ucp_proto_multi_init_params_t *params)
 {
+    const char *perf_name = ucp_proto_id_field(params->super.super.proto_id,
```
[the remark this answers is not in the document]
ivankochin: Done.

src/ucp/proto/proto_perf.c
```diff
@@ -228,6 +229,10 @@ int ucp_proto_perf_is_empty(const ucp_proto_perf_t *perf)
     return ucs_list_is_empty(&perf->segments);
 }
 
+const char *ucp_proto_perf_name(const ucp_proto_perf_t *perf) {
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/proto/proto_perf.c
```diff
@@ -414,6 +419,113 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+            max_value     = factors[factor_id].c;
+        }
+    }
+
+    /* Turn longest factor overhead to dynamic only */
+    factors[max_factor_id]    = ucp_proto_perf_segment_func(frag_seg,
+                                                            max_factor_id);
+    factors[max_factor_id].m += factors[max_factor_id].c / frag_size;
+    factors[max_factor_id].c  = 0;
+
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/proto/proto_perf.c
```diff
@@ -414,6 +419,113 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+    size_t frag_size;
+    char frag_str[64];
+
+    ucs_assert(frag_seg != NULL);
+    frag_size = ucp_proto_perf_segment_end(frag_seg);
+    ucs_assertv(frag_size < max_length, "frag_size=%zu max_length=%zu",
+                frag_size, max_length);
+    ucs_assertv(ucp_proto_perf_find_segment_lb(perf, frag_size + 1) == NULL,
+                "ppln range already contains perf data frag_size=%zu",
+                frag_size);
```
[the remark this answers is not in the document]
yosefe: i think proto_perf struct should be more generic and either return error code (and protocol logic will do fatal failure) or support the case where something already exists and remove it

src/ucp/proto/proto_perf.c
```diff
@@ -414,6 +419,113 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+                                       const ucp_proto_perf_t *perf1,
+                                       const ucp_proto_perf_t *perf2,
+                                       ucp_proto_perf_t **perf_p)
+{
+    const ucp_proto_perf_t *perf_elems[2] = {perf1, perf2};
+
+    return ucp_proto_perf_aggregate(name, perf_elems, 2, perf_p);
+}
+
+ucs_status_t ucp_perf_add_ppln(ucp_proto_perf_t *perf,
```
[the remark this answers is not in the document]
ivankochin: The only gtest I can imagine is a testing through reference implementation which would check that current logic didn't changed, but I am not sure whether we need such type of testing. I think better testing for that function is regular performance degradation testing.

src/ucp/proto/proto_select.c
ivankochin: AFAIU no, because of the for loop defined later since one of iterations in that for loop can overwrite `proto_short` data and next one can call `out_disable` so we would need to overwrite this field to disable again.

src/ucp/rndv/proto_rndv.c
ivankochin: You mean that we need to add one portion of `params->super.overhead` to LOCAL_CPU and the same portion to LOCAL_CPU  instead of multiplying it here by 2 and adding whole this overhead to LOCAL_CPU?

src/ucp/rndv/proto_rndv.c
ivankochin: AFAIU it would require introducing separate type for UCS_ARRAY_ALLOC_ONSTACK or defining separate buffer on stack to store data separately there (for UCS_ARRAY_FIXED_INITIALIZER). Do you think that one of these options are better than usual C array or am I missing a way to detect ucs_array on stack?

src/ucp/rndv/proto_rndv.c
ivankochin: The logic here is that we shouldn't destroy `perf` structure in case of success because ownership of that function was moved to  `init_params->ctx`. What is unclear here?

src/ucp/rndv/proto_rndv.c
ivankochin: Yes, that helped to fix a couple of performance issues.

src/ucp/rndv/rndv_ats.c
ivankochin: The only case when status is OK and priv.lane is NULL is case when PPLN is detected. I changed this condition to `perf == NULL`.

src/ucp/rndv/rndv_get.c
ivankochin: I had such wrapper before but @rakhmets mentioned that this makes things less straightforward and adds more error 
prone since we create perf on one level and destroy it on another. See here https://github.com/openucx/ucx/pull/10051#discussion_r1722944168

src/ucp/rndv/rndv_ppln.c
ivankochin: Do you mean to just move ppln_overhead to a separate perf structure(and node) on the same level?

src/ucp/rndv/rndv_ppln.c
ivankochin: I see that it used only once for `init_params->proto_id` and once for `proto->proto_id`

src/ucp/proto/proto_perf.c
```diff
@@ -414,6 +419,113 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+    /* Turn all factors overheads to constant and choose longest one */
+    max_factor_id = 0;
+    max_value     = -DBL_MAX;
+    for (factor_id = 0; factor_id < UCP_PROTO_PERF_FACTOR_LAST; factor_id++) {
+        factor_func          = ucp_proto_perf_segment_func(frag_seg, factor_id);
+        factors[factor_id].c = ucs_linear_func_apply(factor_func, frag_size);
+        if ((factors[factor_id].c > max_value) &&
+            (factor_id != UCP_PROTO_PERF_FACTOR_LATENCY)) {
+            max_factor_id = factor_id;
+            max_value     = factors[factor_id].c;
```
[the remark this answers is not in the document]
ivankochin: Removed

src/ucp/proto/proto_perf.c
```diff
@@ -414,6 +419,113 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+    ucp_proto_perf_check(remote_perf);
+
+    status = ucp_proto_perf_create("remote", &perf);
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    /* Convert local factors to remote and vice versa */
+    ucp_proto_perf_segment_foreach(remote_seg, remote_perf) {
+        for (fidx = 0; fidx < ucs_static_array_size(convert_map); ++fidx) {
```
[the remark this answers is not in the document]
ivankochin: Done

src/ucp/proto/proto_single.c
```diff
@@ -19,9 +19,11 @@
 
 
 ucs_status_t ucp_proto_single_init(const ucp_proto_single_init_params_t *params,
-                                   ucp_proto_caps_t *caps,
+                                   ucp_proto_perf_t **perf_p,
                                    ucp_proto_single_priv_t *spriv)
 {
+    const char *perf_name = ucp_proto_id_field(params->super.super.proto_id,
```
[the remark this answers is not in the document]
ivankochin: Changed

src/ucp/proto/proto_perf.c
```diff
@@ -414,6 +419,113 @@ ucs_status_t ucp_proto_perf_aggregate(const char *name,
...
+    factors[max_factor_id].c  = 0;
+
+
+    ucs_memunits_to_str(frag_size, frag_str, sizeof(frag_str));
+    return ucp_proto_perf_add_funcs(perf, frag_size + 1, max_length, factors,
+                                    ucp_proto_perf_segment_node(frag_seg),
+                                    "pipeline", "frag size: %s", frag_str);
+}
+
+ucs_status_t ucp_proto_perf_remote(const ucp_proto_perf_t *remote_perf,
```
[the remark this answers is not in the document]
ivankochin: Added gtest but it works by a different way then suggested since suggested one can miss a case when this func does nothing

src/ucp/proto/proto_select.c
ivankochin: Removed

src/ucp/proto/proto_select.c
ivankochin: Added

src/ucp/rndv/proto_rndv.c
ivankochin: Done

src/ucp/proto/proto_perf.c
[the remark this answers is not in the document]
ivankochin: Changed signature to get two perf structures and return frag segment

src/ucp/rndv/rndv_get.c
ivankochin: Changed back to destroying perf inside `ucp_proto_select_add_proto`

src/ucp/proto/proto_init.h
ivankochin: Renamed but left in the `proto_init` file

src/ucp/proto/proto_perf.h
```diff
@@ -177,8 +177,9 @@ ucs_status_t ucp_proto_perf_aggregate2(const char *name,
  * All the factors except longest one turn into constant fragment overhead
  * due to overlapping (1 and 3 from example).
  *
- * Longest factor still saves the slope but it's constant part turns
- * to dynamic since it start to depend on number of sent fragments
+ * Longest factor still saves the linear function part but has additional
+ * overhead turned to dynamic since it start to depend on number of sent
```
brminich, line 181: starts

src/ucp/rndv/rndv_ppln.c
```diff
@@ -127,13 +139,15 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
...
+        }
+
         /* Add ATS overhead */
-        ppln_overhead = ucs_linear_func_make(frag_overhead,
-                                             frag_overhead / rpriv.frag_size);
-        status        = ucp_proto_rndv_ack_init(&ack_params,
-                                                UCP_PROTO_RNDV_ATS_NAME,
-                                                ppln_overhead, &ack_perf,
-                                                &rpriv.ack);
+        status  = ucp_proto_rndv_ack_init(&ack_params,UCP_PROTO_RNDV_ATS_NAME,
```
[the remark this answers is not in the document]
ivankochin:
```suggestion
        status  = ucp_proto_rndv_ack_init(&ack_params, UCP_PROTO_RNDV_ATS_NAME,
```

src/ucp/proto/proto_perf.h
ivankochin:
```suggestion
 * overhead turned to dynamic since it starts to depend on number of sent
```

src/ucp/rndv/proto_rndv.c
```diff
@@ -589,7 +589,7 @@ ucs_status_t ucp_proto_rndv_rts_reset(ucp_request_t *req)
 
 ucs_status_t
 ucp_proto_rndv_ack_init(const ucp_proto_common_init_params_t *init_params,
-                        const char *name, ucs_linear_func_t overhead,
+                        const char *name, double overhead,
```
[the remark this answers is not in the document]
ivankochin: Before that patch we were adding ppln overhead to ATS and since this overhead has linear part we were needed to pass linear function here. After decision to make this ppln overhead part of `ppln_perf` there is no need to get linear function as argument. 

I always though that linear overhead (which depends on the message size) for acknowledge performance looks weird so decided to change this argument to double.

src/ucp/proto/proto_init.c
```diff
@@ -279,10 +272,12 @@ ucp_proto_init_add_buffer_copy_time(ucp_worker_h worker, const char *title,
...
                                     int local, ucp_proto_perf_t *perf)
 {
-    ucp_proto_perf_factors_t perf_factors = UCP_PROTO_PERF_FACTORS_INITIALIZER;
-    ucp_context_h context                 = worker->context;
-    ucp_proto_perf_factor_id_t factor_id  = ucp_proto_buffer_copy_factor_id(
-            local_mem_type, remote_mem_type, memtype_op, local);
+    ucp_context_h context                            = worker->context;
+    ucp_proto_perf_factors_t perf_factors            =
+            UCP_PROTO_PERF_FACTORS_INITIALIZER;
+    ucp_proto_perf_factor_id_t buffer_copy_factor_id =
```
yosefe, line 278: minor: initialize it after the declaration to reduce indentation

src/ucp/proto/proto_perf.c
```diff
@@ -429,49 +430,64 @@ ucs_status_t ucp_proto_perf_aggregate2(const char *name,
...
+ * RTS sends can potentially make another factor the longest one due
+ * to different factors overheads in RTS.
+ * 
+ * That can be potentially solved by calculating all the pipeline logic in the
+ * end during initialization of `flat_perf` structure but it is unclear how to
+ * distinguish which parts of the factors are part of pipeline and which aren't
+ * at that moment.
+ */
+const ucp_proto_perf_segment_t *
+ucp_proto_perf_add_ppln(ucp_proto_perf_t *perf, ucp_proto_perf_t *ppln_perf,
```
yosefe, line 449: can "perf" be const?

src/ucp/proto/proto_perf.c
```diff
@@ -429,49 +430,64 @@ ucs_status_t ucp_proto_perf_aggregate2(const char *name,
...
 
     ucs_memunits_to_str(frag_size, frag_str, sizeof(frag_str));
-    return ucp_proto_perf_add_funcs(perf, frag_size + 1, max_length, factors,
-                                    ucp_proto_perf_segment_node(frag_seg),
-                                    "pipeline", "frag size: %s", frag_str);
+    status = ucp_proto_perf_add_funcs(ppln_perf, frag_size + 1, max_length,
+                                      factors,
+                                      ucp_proto_perf_segment_node(frag_seg),
+                                      "pipeline", "frag size: %s", frag_str);
+    return (status == UCS_OK) ? frag_seg : NULL;
```
yosefe, line 490:
```c
if (status != UCS_OK) {
   return NULL;
}

return drag_seg;
```

src/ucp/rndv/rndv_ats.c
```diff
@@ -34,18 +34,14 @@ static void ucp_proto_rndv_ats_probe(const ucp_proto_init_params_t *init_params)
         return;
     }
 
-    status = ucp_proto_rndv_ack_init(&params, UCP_PROTO_RNDV_ATS_NAME,
-                                     UCS_LINEAR_FUNC_ZERO, &perf, &priv);
-    if ((status != UCS_OK) || (priv.lane == UCP_NULL_LANE)) {
+    status = ucp_proto_rndv_ack_init(&params, UCP_PROTO_RNDV_ATS_NAME, 0, &perf,
+                                     &priv);
+    if ((status != UCS_OK) || (perf == NULL)) {
```
yosefe, line 39: can status be UCS_OK and perf==NULL? i would expect status to be an error in this case

src/ucp/rndv/rndv_ppln.c
```diff
@@ -85,52 +97,76 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
...
 
-        frag_range = proto->caps.ranges + (proto->caps.num_ranges - 1);
-        if (frag_range->max_length == 0) {
+        /* Add the single range of the pipeline protocol to ppln_caps */
+        status = ucp_proto_perf_create("pipeline", &ppln_perf);
+        if (status != UCS_OK) {
             continue;
         }
 
+        first_seg = ucp_proto_perf_find_segment_lb(proto->perf, 0);
```
yosefe, line 116: initialize first_seg right before line 123 - where it is used

src/ucp/rndv/rndv_ats.c
ivankochin: status can be OK and perf can be NULL in case if `UCP_PROTO_SELECT_OP_FLAG_PPLN_FRAG` is set (protocol trying to be initialized for using as a part of pipeline)

src/ucp/rndv/rndv_ppln.c
ivankochin: Done

src/ucp/proto/proto_perf.c
ivankochin: Done

src/ucp/proto/proto_perf.c
ivankochin: Yes, done

src/ucp/proto/proto_init.c
ivankochin: Done

src/ucp/proto/proto_init.c
```diff
@@ -115,19 +115,19 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+ucp_proto_count_recv_overhead(const ucp_proto_common_init_params_t *params,
+                              uint32_t op_attr_mask)
 {
     if (/* Don't care about receiver time for one-sided remote access */
         (params->flags & UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS) ||
         /* Count only send completion time without waiting for a response */
         ((op_attr_mask & UCP_OP_ATTR_FLAG_FAST_CMPL) &&
          !(params->flags & UCP_PROTO_COMMON_INIT_FLAG_RESPONSE))) {
-        return UCS_LINEAR_FUNC_ZERO;
+        return 0;
```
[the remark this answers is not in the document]
ivankochin: Done, but I think that conditions are more readable and easier to understand in that order, so I changed `ucp_proto_count_recv_overhead` to `ucp_proto_init_skip_recv_overhead`

src/ucp/proto/proto_init.c
```diff
@@ -115,19 +115,19 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
     return UCS_OK;
 }
 
-static ucs_linear_func_t
-ucp_proto_init_tl_recv_ovh(const ucp_proto_common_init_params_t *params,
-                           const ucp_proto_common_tl_perf_t *tl_perf,
-                           uint32_t op_attr_mask)
+static int
+ucp_proto_count_recv_overhead(const ucp_proto_common_init_params_t *params,
```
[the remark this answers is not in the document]
ivankochin: Changed, but to `ucp_proto_init_skip_recv_overhead` due to next comment

yosefe (approved):

src/ucp/proto/proto_debug.h
rakhmets: Let's keep the current names.

src/ucp/proto/proto_init.c
```diff
@@ -196,196 +115,178 @@ ucp_proto_perf_envelope_make(const ucs_linear_func_t *funcs, uint64_t funcs_num,
...
+    if (!ucp_proto_init_skip_recv_overhead(params, op_attr_mask)) {
+        perf_factors[UCP_PROTO_PERF_FACTOR_REMOTE_CPU].c +=
+                tl_perf->recv_overhead;
+    }
 
-        ++caps->num_ranges;
-        range_start = range->max_length + 1;
+    /* With non-zcopy fast-completion we don't count transport time */
+    if (!(op_attr_mask & UCP_OP_ATTR_FLAG_FAST_CMPL) ||
+        (params->flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY)) {
```
brminich, line 166: should we also do it for `params->flags & UCP_PROTO_COMMON_INIT_FLAG_RESPONSE`?

yosefe: actually, for UCP_PROTO_COMMON_INIT_FLAG_RESPONSE we should count the BW as REMOTE_TL instead of LOCAL_TL, but we can do such tuning in next PR IMO.

brminich: ok

brminich (approved):

PR #10052, opened 2024-08-08
UCT/GDRCOPY: Require rcache by default

brminich:
## What
Require rcache to be used by gdr copy md by default

## Why ?
Otherwise there maybe an issue where small memory region to be registered is aligned to 64K and IB registration fails with the following errors
```
ib_md.c:276  UCX  ERROR ibv_reg_dmabuf_mr(address=0x7f271a600000, length=65536, access=0xf) failed: Invalid argument
ucp_mm.c:70   UCX  ERROR failed to register address 0x7f271a600000 (cuda) length 65536 dmabuf_fd 98 on md[2]=mlx5_0: Input/output error (md supports: host|cuda)
```
## How
 Keep `UCX_GDR_COPY_RCACHE` env var, but make it bool instead of ternary. So it is always enabled by default, but there is still a possibility to debug/use gdrcopy transport without rcache

### DIFF
--- a/src/uct/cuda/gdr_copy/gdr_copy_md.c
+++ b/src/uct/cuda/gdr_copy/gdr_copy_md.c
@@ -42,8 +42,8 @@ static ucs_config_field_t uct_gdr_copy_md_config_table[] = {
      "shared across all contexts",
      ucs_offsetof(uct_gdr_copy_md_config_t, shared), UCS_CONFIG_TYPE_BOOL},
 
-    {"RCACHE", "try", "Enable using memory registration cache",
-     ucs_offsetof(uct_gdr_copy_md_config_t, enable_rcache), UCS_CONFIG_TYPE_TERNARY},
+    {"RCACHE", "yes", "Enable using memory registration cache",
+     ucs_offsetof(uct_gdr_copy_md_config_t, enable_rcache), UCS_CONFIG_TYPE_BOOL},
 
     {"MEM_REG_OVERHEAD", "16us", "Memory registration overhead", /* TODO take default from device */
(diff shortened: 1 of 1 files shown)


Artemy-Mellanox (approved):

tvegas1 (approved):

ivankochin (dismissed):

src/uct/cuda/gdr_copy/gdr_copy_md.c
```diff
@@ -42,8 +42,8 @@ static ucs_config_field_t uct_gdr_copy_md_config_table[] = {
      "shared across all contexts",
      ucs_offsetof(uct_gdr_copy_md_config_t, shared), UCS_CONFIG_TYPE_BOOL},
 
-    {"RCACHE", "try", "Enable using memory registration cache",
-     ucs_offsetof(uct_gdr_copy_md_config_t, enable_rcache), UCS_CONFIG_TYPE_TERNARY},
+    {"RCACHE", "yes", "Enable using memory registration cache",
+     ucs_offsetof(uct_gdr_copy_md_config_t, enable_rcache), UCS_CONFIG_TYPE_BOOL},
```
rakhmets, line 46: Is it safe to change the option type from `ternary` to `bool`? How do we handle the case when the new version will be run with `RCACHE=try`? If there is a crash, then it is better to keep it `ternary`, but print a message that `try` is treated as `yes`, and will be removed.

Artemy-Mellanox: we haven't released a version with a ternary variant yet, so maybe it's not that big problem

rakhmets (approved):

yosefe (approved):

PR #10053, opened 2024-08-10
UCP/MM: Selectively exclude mem type MDs from rcache

Artemy-Mellanox:
## What
Bypass UCP rcache when using gdr_copy for mem_type operations.

## Why?
gdr_copy has its own rcache and double rcache restores internal issue [3606732](https://redmine.mellanox.com/issues/3606732). The root cause of this issue is the insufficient support for overlapping regions in the gdr_copy driver. With single caching, the memory handle's reference count tracks the number of in-flight operations, ensuring no overlapping memory regions for mem_type operations. However, with double caching, the reference count in the secondary cache corresponds to the number of primary memory regions, which does not provide the same guarantee.

## How?
Introduce a static user memh flow for mem_type operations and mark gdr_copy as non-cacheable of UCP.

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -368,7 +368,6 @@ static void ucp_memh_dereg(ucp_context_h context, ucp_mem_h memh,
 
         memh->uct[md_index] = NULL;
     }
-    memh->md_map &= ~md_map;
 
     if ((memh->flags & UCP_MEMH_FLAG_MLOCKED) &&
         (context->gva_md_map[memh->mem_type] & memh->md_map) == 0) {
@@ -417,9 +416,19 @@ static void ucp_memh_put_rcache(ucp_context_h context, ucp_mem_h memh)
     UCP_THREAD_CS_EXIT(&context->mt_lock);
 }
 
+static void ucp_memh_dereg_all(ucp_context_h context, ucp_mem_h memh)
+{
+    if (memh->parent == memh) {
+        ucp_memh_dereg(context, memh, memh->md_map);
+    } else {
+        /* Have a parent memory handle from rcache */
+        ucp_memh_dereg(context, memh, memh->md_map & ~memh->parent->md_map);
+        ucp_memh_put_rcache(context, memh->parent);
+    }
+}
+
 static void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
 {
-    ucp_md_map_t md_map = memh->md_map;
     uct_allocated_memory_t mem;
     ucs_status_t status;
 
@@ -433,18 +442,12 @@ static void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
 
     if (mem.method == UCT_ALLOC_METHOD_MD) {
         ucs_assert(memh->alloc_md_index != UCP_NULL_RESOURCE);
-        mem.md   = context->tl_mds[memh->alloc_md_index].md;
-        mem.memh = memh->uct[memh->alloc_md_index];
-        md_map  &= ~UCS_BIT(memh->alloc_md_index);
+        mem.md        = context->tl_mds[memh->alloc_md_index].md;
+        mem.memh      = memh->uct[memh->alloc_md_index];
+        memh->md_map &= ~UCS_BIT(memh->alloc_md_index);
     }
 
-    /* Have a parent memory handle from rcache */
-    if (memh->parent != memh) {
-        ucp_memh_dereg(context, memh, md_map & ~memh->parent->md_map);
-        ucp_memh_put_rcache(context, memh->parent);
-    } else {
-        ucp_memh_dereg(context, memh, md_map);
-    }
+    ucp_memh_dereg_all(context, memh);
 
     /* If the memory was also allocated, release it */
     if (memh->alloc_method != UCT_ALLOC_METHOD_LAST) {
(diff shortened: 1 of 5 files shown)


src/ucp/core/ucp_mm.h
```diff
@@ -27,6 +27,8 @@
    existing region */
 #define UCP_MM_UCT_ACCESS_MASK UCT_MD_MEM_ACCESS_ALL
 
+#define UCP_MM_UCT_NON_CACHED UCS_BIT(31)
```
brminich, line 30: imo, such flag needs to be defined in UCT instead

src/uct/cuda/gdr_copy/gdr_copy_md.c
```diff
@@ -62,7 +62,7 @@ uct_gdr_copy_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
     uct_md_base_md_query(md_attr);
     md_attr->flags            = UCT_MD_FLAG_REG | UCT_MD_FLAG_NEED_RKEY;
     md_attr->reg_mem_types    = UCS_BIT(UCS_MEMORY_TYPE_CUDA);
-    md_attr->cache_mem_types  = UCS_BIT(UCS_MEMORY_TYPE_CUDA);
+    md_attr->cache_mem_types  = 0;
```
brminich, line 65: should not be this enough to avoid UCP rcache usage?

src/ucp/core/ucp_mm.c
```diff
@@ -825,6 +826,7 @@ static ucs_status_t ucp_memh_init_uct_reg(ucp_context_h context, ucp_mem_h memh,
         }
 
         ucp_memh_init_from_parent(memh, cache_md_map);
+        memh->flags |= UCP_MEMH_FLAG_NON_CACHED;
```
yosefe, line 829: we set this in both if/else, should it be outside to do it only once?
or even do it from ucp_mem_map/ucp_memh_create_from_mem/ucp_memh_import when they call ucp_memh_create?

src/ucp/core/ucp_mm.c
```diff
@@ -1140,7 +1143,7 @@ ucs_status_t ucp_mem_map(ucp_context_h context, const ucp_mem_map_params_t *para
 
 void ucp_memh_put_slow(ucp_context_h context, ucp_mem_h memh)
 {
-    ucs_assert(context->rcache == NULL);
+    ucs_assert(memh->flags & UCP_MEMH_FLAG_NON_CACHED);
```
yosefe, line 1146: shall we do also an (inverse) assert when releasing memh from rcache?

src/ucp/core/ucp_mm.c
```diff
@@ -1180,9 +1184,12 @@ ucs_status_t ucp_mem_type_reg_buffers(ucp_worker_h worker, void *remote_addr,
     tl_md  = &context->tl_mds[md_index];
     cmpt   = context->tl_cmpts[tl_md->cmpt_index].cmpt;
 
+    if (!(context->cache_md_map[mem_type] & UCS_BIT(md_index))) {
+        uct_flags |= UCP_MM_UCT_NON_CACHED;
+    }
```
yosefe, line 1189: can we combine cached and non-cached MDs in the same memh?

src/ucp/core/ucp_mm.h
yosefe: can we avoid this flag, since we already have UCP_MEMH_FLAG_NON_CACHED?

src/ucp/dt/dt.c
```diff
@@ -31,10 +31,10 @@ UCS_PROFILE_FUNC_VOID(ucp_mem_type_unpack,
                       ucp_worker_h worker, void *buffer, const void *recv_data,
                       size_t recv_length, ucs_memory_type_t mem_type)
 {
-    ucp_ep_h ep         = worker->mem_type_ep[mem_type];
+    ucp_ep_h ep                 = worker->mem_type_ep[mem_type];
+    ucp_mem_dummy_handle_t memh = {};
```
brminich, line 35: why is this needed?
(also in `ucp_mem_type_pack`)

Artemy-Mellanox: in PR desc I call it "static user memh"

src/ucp/core/ucp_mm.c
```diff
@@ -843,6 +841,29 @@ static ucs_status_t ucp_memh_init_uct_reg(ucp_context_h context, ucp_mem_h memh,
...
+    ucp_md_map_t reg_md_map    = context->reg_md_map[mem_type];
+    ucs_status_t status;
+
+    if (uct_flags & UCT_MD_MEM_FLAG_LOCK) {
+        reg_md_map |= context->reg_block_md_map[mem_type];
+    }
+
+    reg_md_map &= ~memh->md_map;
+    status = ucp_memh_init_uct_reg(context, memh, reg_md_map, uct_flags,
+                                   alloc_name);
```
yosefe, line 858: maybe pass "reg_md_map &  ~memh->md_map" to ucp_memh_init_uct_reg instead of modifying reg_md_map?

src/ucp/core/ucp_mm.c
```diff
@@ -843,6 +841,29 @@ static ucs_status_t ucp_memh_init_uct_reg(ucp_context_h context, ucp_mem_h memh,
...
+    if (status != UCS_OK) {
+        return status;
+    }
+
+    ucs_assert(ucp_memh_is_user_memh(memh));
```
yosefe, line 863: maybe do it in the beginning of the function?

src/ucp/core/ucp_mm.c
```diff
@@ -1177,12 +1197,15 @@ ucs_status_t ucp_mem_type_reg_buffers(ucp_worker_h worker, void *remote_addr,
         goto out;
     }
 
+    memh->super.super.start = (uintptr_t)remote_addr;
+    memh->super.super.end   = (uintptr_t)remote_addr + length;
+    ucp_memh_init(memh, context, 0, UCT_MD_MEM_ACCESS_ALL,
```
yosefe, line 1202: can we skip ucp_memory_detect that is inside ucp_memh_init?

src/ucp/dt/dt.c
yosefe: why need to initialize it to 0?

src/ucp/dt/dt.c
```diff
@@ -59,19 +59,19 @@ UCS_PROFILE_FUNC_VOID(ucp_mem_type_unpack,
...
 }
 
 UCS_PROFILE_FUNC_VOID(ucp_mem_type_pack,
                       (worker, dest, src, length, mem_type),
                       ucp_worker_h worker, void *dest, const void *src,
                       size_t length, ucs_memory_type_t mem_type)
 {
-    ucp_ep_h ep         = worker->mem_type_ep[mem_type];
+    ucp_ep_h ep                 = worker->mem_type_ep[mem_type];
+    ucp_mem_dummy_handle_t memh = {};
```
yosefe, line 71: why need to initialize?

src/uct/cuda/gdr_copy/gdr_copy_md.c
yosefe: shall we remove line 65?

src/ucp/dt/dt.c
Artemy-Mellanox: user memh flow expect it to be zeroed

yosefe: can we avoid this extra overhead?

src/ucp/dt/dt.c
```diff
@@ -59,30 +60,31 @@ UCS_PROFILE_FUNC_VOID(ucp_mem_type_unpack,
...
 }
 
 UCS_PROFILE_FUNC_VOID(ucp_mem_type_pack,
                       (worker, dest, src, length, mem_type),
                       ucp_worker_h worker, void *dest, const void *src,
                       size_t length, ucs_memory_type_t mem_type)
 {
-    ucp_ep_h ep                 = worker->mem_type_ep[mem_type];
-    ucp_mem_dummy_handle_t memh = {};
+    ucp_ep_h ep         = worker->mem_type_ep[mem_type];
```
yosefe, line 71: can remove some spaces before =

src/ucp/dt/dt.c
```diff
@@ -31,22 +31,23 @@ UCS_PROFILE_FUNC_VOID(ucp_mem_type_unpack,
                       ucp_worker_h worker, void *buffer, const void *recv_data,
                       size_t recv_length, ucs_memory_type_t mem_type)
 {
-    ucp_ep_h ep                 = worker->mem_type_ep[mem_type];
-    ucp_mem_dummy_handle_t memh = {};
+    ucp_ep_h ep         = worker->mem_type_ep[mem_type];
```
yosefe, line 34: can remove some spaces before =

yosefe (approved):

src/ucp/core/ucp_mm.c
```diff
@@ -417,9 +416,19 @@ static void ucp_memh_put_rcache(ucp_context_h context, ucp_mem_h memh)
     UCP_THREAD_CS_EXIT(&context->mt_lock);
 }
 
+static void ucp_memh_dereg_all(ucp_context_h context, ucp_mem_h memh)
+{
+    if (memh->parent == memh) {
+        ucp_memh_dereg(context, memh, memh->md_map);
+    } else {
```
iyastreb, line 423: Should we handle here the case when parent == NULL?

yosefe (approved):

PR #10054, opened 2024-08-11
UCP/WIREUP: specify a reason for wireup failure - part 2

amastbaum:
## What
During the wireup process, provide a reason when a resource's lane is unreachable (second part after https://github.com/openucx/ucx/pull/9995. Includes all the transports other than IB).

## Why ?
Users need more information about why a device is not reachable after an unsuccessful connection establishment. This information should be passed from UCT to UCP.

## How ?
Pass a string to select/search_lane functions so in case the wireup process fails, the reason will be printed out in an upper layer.

### DIFF
--- a/src/uct/base/uct_iface.c
+++ b/src/uct/base/uct_iface.c
@@ -301,8 +301,16 @@ int uct_iface_scope_is_reachable(const uct_iface_h iface,
 
     ucs_assert(params->field_mask & UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR);
 
-    return (scope == UCT_IFACE_REACHABILITY_SCOPE_NETWORK) ||
-           uct_iface_is_same_device(iface, params->device_addr);
+    if (scope == UCT_IFACE_REACHABILITY_SCOPE_NETWORK) {
+        return 1;
+    }
+
+    if (!uct_iface_is_same_device(iface, params->device_addr)) {
+        uct_iface_fill_info_str_buf(params, "device address is different");
+        return 0;
+    }
+
+    return 1;
 }
 
 int uct_iface_is_reachable_v2(const uct_iface_h tl_iface,
@@ -992,7 +1000,8 @@ void uct_iface_get_local_address(uct_iface_local_addr_ns_t *addr_ns,
 }
 
 int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
-                                 ucs_sys_namespace_type_t sys_ns_type)
+                                 ucs_sys_namespace_type_t sys_ns_type,
+                                 const uct_iface_is_reachable_params_t *params)
 {
     uct_iface_local_addr_ns_t my_addr = {};
 
@@ -1003,6 +1012,9 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
     /* Check if both processes are on same host and both of them are in root (or
      * non-root) pid namespace */
     if (addr_ns->super.id != my_addr.super.id) {
+        uct_iface_fill_info_str_buf(params,
+                                    "different host id local %lu vs %lu",
+                                    my_addr.super.id, addr_ns->super.id);
         return 0;
     }
 
@@ -1012,7 +1024,14 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
 
     /* We are in non-root PID namespace - return 1 if ID of namespaces are the
      * same */
-    return addr_ns->sys_ns == my_addr.sys_ns;
+    if (addr_ns->sys_ns != my_addr.sys_ns) {
+        uct_iface_fill_info_str_buf(
+                    params,
+                    "different pid namespaces %"PRIx64" vs %"PRIx64"",
+                    my_addr.sys_ns, addr_ns->sys_ns);
+        return 0;
+    }
+    return 1;
 }
 
 void uct_iface_mpool_config_copy(ucs_mpool_params_t *mp_params,
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
@@ -107,11 +107,14 @@ uct_cuda_ipc_iface_is_reachable_v2(const uct_iface_h tl_iface,
     uct_cuda_ipc_md_t *md        = ucs_derived_of(base_iface->md, uct_cuda_ipc_md_t);
 #endif
 
-    if (!uct_iface_is_reachable_params_addrs_valid(params) ||
-        (getpid() == *(pid_t*)params->iface_addr) ||
-        !uct_iface_scope_is_reachable(tl_iface, params)) {
+    if (!uct_iface_is_reachable_params_addrs_valid(params)) {
         return 0;
-    };
+    }
+
+    if (getpid() == *(pid_t*)params->iface_addr) {
+        uct_iface_fill_info_str_buf(params, "same process");
+        return 0;
+    }
 
 #if HAVE_CUDA_FABRIC
     if (uct_cuda_ipc_iface_is_mnnvl_supported(md)) {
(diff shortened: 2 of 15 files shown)


yosefe:
@shasson5 can you pls review?

src/uct/tcp/tcp_iface.c
```diff
@@ -208,19 +209,30 @@ uct_tcp_iface_is_reachable_v2(const uct_iface_h tl_iface,
 
     tcp_dev_addr = (uct_tcp_device_addr_t*)params->device_addr;
     if (iface->config.ifaddr.ss_family != tcp_dev_addr->sa_family) {
+        uct_iface_fill_info_str_buf(
+                params, "socket address family types differ, %d vs %d",
```
yosefe, line 213: different address family %d vs %d

src/uct/ze/copy/ze_copy_iface.c
```diff
@@ -41,8 +41,18 @@ uct_ze_copy_iface_is_reachable_v2(const uct_iface_h tl_iface,
...
-    return (addr != NULL) && (iface->id == *addr) &&
-           uct_iface_scope_is_reachable(tl_iface, params);
+    if (addr == NULL) {
+        uct_iface_fill_info_str_buf(params, "device address is empty");
+        return 0;
+    }
+
+    if (iface->id != *addr) {
+        uct_iface_fill_info_str_buf(params,
+                                    "iface id and device address differ");
```
yosefe, line 51: "different iface id"

src/uct/base/uct_iface.c
```diff
@@ -1003,6 +1013,13 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
     /* Check if both processes are on same host and both of them are in root (or
      * non-root) pid namespace */
     if (addr_ns->super.id != my_addr.super.id) {
+        if (info_string != NULL) {
```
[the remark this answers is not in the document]
amastbaum: uct_iface_fill_info_str_buf receives (uct_iface_is_reachable_params_t *) while uct_iface_local_is_reachable doesn't hold this variable. I thought maybe to create one function that prints a string to a buffer and another wrapper function that validates  uct_iface_is_reachable_params_t fields but I didn't know if this is a overhead (it results in creating va_list twice, etc.)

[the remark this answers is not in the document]
yosefe: maybe we can pass params to uct_iface_local_is_reachable?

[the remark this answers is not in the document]
amastbaum: You're right, I am not sure why I didn't do it in the first place

yosefe:
failure seems relevant:
```
2024-08-26T07:53:11.2247900Z [ RUN      ] tcp/test_uct_sockaddr_err_handle_non_exist_ip.conn_to_non_exist_ip/2 </lo>
2024-08-26T07:53:11.2250959Z [     INFO ] Testing tcp on 0.0.0.0:49036 interface lo
2024-08-26T07:55:18.5225372Z [       OK ] tcp/test_uct_sockaddr_err_handle_non_exist_ip.conn_to_non_exist_ip/2 (127298 ms)
2024-08-26T07:55:18.5256477Z [----------] 1 test from tcp/test_uct_sockaddr_err_handle_non_exist_ip (127298 ms total)
2024-08-26T07:55:18.5257776Z 
2024-08-26T07:55:18.5258713Z [----------] 1 test from cuda_ipc/test_uct_ep
2024-08-26T07:55:18.5261351Z [ RUN      ] cuda_ipc/test_uct_ep.is_connected/0 <cuda_ipc/cuda>
2024-08-26T07:55:18.5262355Z /__w/1/s/contrib/../test/gtest/uct/test_uct_ep.cc:220: Failure
2024-08-26T07:55:18.5262982Z Value of: is_connected_to_sender(*m_receiver)
2024-08-26T07:55:18.5263335Z   Actual: false
2024-08-26T07:55:18.5263591Z Expected: true
2024-08-26T07:55:18.5263890Z [  FAILED  ] cuda_ipc/test_uct_ep.is_connected/0, where GetParam() = cuda_ipc/cuda (0 ms)
2024-08-26T07:55:18.5264403Z [----------] 1 test from cuda_ipc/test_uct_ep (0 ms total)
```

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -107,11 +107,20 @@ uct_cuda_ipc_iface_is_reachable_v2(const uct_iface_h tl_iface,
...
-        (getpid() == *(pid_t*)params->iface_addr) ||
-        !uct_iface_scope_is_reachable(tl_iface, params)) {
+    if (!uct_iface_is_reachable_params_addrs_valid(params)) {
         return 0;
-    };
+    }
+
+    if (getpid() == *(pid_t*)params->iface_addr) {
+        uct_iface_fill_info_str_buf(params,
+                                    "different pid %d vs %d", getpid(),
```
yosefe, line 116: "same process"
since cuda_ipc cannot connect within the same process - check is == not !=

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -107,11 +107,20 @@ uct_cuda_ipc_iface_is_reachable_v2(const uct_iface_h tl_iface,
...
+                                    *(pid_t*)params->iface_addr);
+        return 0;
+    }
+
+    if (!uct_iface_scope_is_reachable(tl_iface, params)) {
```
yosefe, line 121: maybe unite this with the 'if' in line 110

src/uct/base/uct_iface.c
```diff
@@ -301,8 +301,17 @@ int uct_iface_scope_is_reachable(const uct_iface_h iface,
...
 
-    return (scope == UCT_IFACE_REACHABILITY_SCOPE_NETWORK) ||
-           uct_iface_is_same_device(iface, params->device_addr);
+    if (scope == UCT_IFACE_REACHABILITY_SCOPE_NETWORK) {
+        return 1;
+    }
+
+    if (!uct_iface_is_same_device(iface, params->device_addr)) {
+        uct_iface_fill_info_str_buf(
+                params, "local and remote devices differ in a device scope");
```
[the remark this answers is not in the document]
yosefe: unresolved

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
amastbaum: It means switching the order between checking the process id and calling uct_iface_scope_is_reachable (which does more checks). Do you think it's ok?

yosefe: actually, i think we need to fix an issue introduced by another PR and move uct_iface_scope_is_reachable to the end of the function (return uct_iface_scope_is_reachable instead of returning 1) because we want to call uct_iface_scope_is_reachable only after making sure the address is reachable

yosefe (dismissed):

yosefe:
coverity failure seems relevant

amastbaum:
@yosefe it needs to be approved again. Thanks

src/uct/rocm/ipc/rocm_ipc_iface.c
```diff
@@ -70,10 +70,22 @@ static int
...
+    }
+
+    if (ucs_get_system_id() != *((const uint64_t*)params->device_addr)) {
+        uct_iface_fill_info_str_buf(params,
+                                    "the device addr is from another machine");
+        return 0;
+    }
+
+    if (getpid() == *(pid_t*)params->iface_addr) {
+        uct_iface_fill_info_str_buf(params,
```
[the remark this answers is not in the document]
gleon99: @yosefe it's not different (`==`)

src/uct/sm/mm/base/mm_iface.c
```diff
@@ -138,10 +138,17 @@ uct_mm_iface_is_reachable_v2(const uct_iface_h tl_iface,
     }
 
     iface_addr = (void*)params->iface_addr;
+    if (iface_addr == NULL) {
+        uct_iface_fill_info_str_buf(params, "iface address is empty");
+        return 0;
+    }
+
+    if (!uct_mm_md_mapper_ops(md)->is_reachable(md, iface_addr->fifo_seg_id,
```
gleon99, line 146: What's the reason for this change? Seems the logic is the same (`return A && B && C`)

amastbaum: You're right, I initially changed it so I could write an error message to the string but eventually I did it internally and didn't switch it back the original version

gleon99:
@amastbaum  please squash

yosefe (approved):

yosefe (approved):

PR #10056, opened 2024-08-13
UCP/ATOMIC: Test ODP/KSM interoperability

ivankochin:
## What
Change for the UCP atomic tests allows to simply reproduce issue of interoperability of ODPv1, KSM and atomic operations.
Originally this issue was reported on GG systems ([RM4025026](https://redmine.mellanox.com/issues/4025026)) but I was able to reproduce it on x64 CPU with IB device. 

## Why ?
To extend test coverage.

### DIFF
--- a/test/gtest/ucp/test_ucp_atomic.cc
+++ b/test/gtest/ucp/test_ucp_atomic.cc
@@ -9,6 +9,7 @@
 
 extern "C" {
 #include <ucp/core/ucp_types.h> /* for atomic mode */
+#include <ucp/core/ucp_ep.inl>
 #include <ucp/core/ucp_mm.h>
 }
 
@@ -17,7 +18,8 @@ class test_ucp_atomic : public test_ucp_memheap {
 public:
     /* Test variants */
     enum {
-        ATOMIC_MODE = UCS_MASK(2)
+        ATOMIC_MODE = UCS_MASK(2),
+        ODP         = UCS_BIT(2)
     };
 
     static void get_test_variants(std::vector<ucp_test_variant>& variants,
@@ -28,6 +30,12 @@ public:
         add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_CPU, "cpu" + name);
         add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_DEVICE, "device" + name);
         add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_GUESS, "guess" + name);
+
+        /* Do not force using device atomics when testing ODP, since some
+         * configurations do not support it. Allow a fallback to SW emulation
+         * in this case.
+         */
+        add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_GUESS | ODP, "guess/odp");
     }
 
     static void get_test_variants(std::vector<ucp_test_variant>& variants)
@@ -124,17 +132,28 @@ protected:
 
     virtual void init() {
         int atomic_mode = get_variant_value() & ATOMIC_MODE;
+        int odp         = get_variant_value() & ODP;
         const char *atomic_mode_cfg =
                         (atomic_mode == UCP_ATOMIC_MODE_CPU)    ? "cpu" :
                         (atomic_mode == UCP_ATOMIC_MODE_DEVICE) ? "device" :
                         (atomic_mode == UCP_ATOMIC_MODE_GUESS)  ? "guess" :
                         "";
         modify_config("ATOMIC_MODE", atomic_mode_cfg);
+
+        if (odp) {
+            modify_config("REG_NONBLOCK_MEM_TYPES", "host");
+            modify_config("IB_MLX5_DEVX_OBJECTS", "", SETENV_IF_NOT_EXIST);
+        }
         test_ucp_memheap::init();
+
+        if (odp && !(check_amo_odp_supported(sender(), UCS_MEMORY_TYPE_HOST))) {
+            cleanup();
+            UCS_TEST_SKIP_R("No AMO lanes with ODP support");
+        }
     }
 
     static unsigned default_num_iters() {
-        return ucs_max(100 / ucs::test_time_multiplier(), 1);
+        return ucs_max(32000 / ucs::test_time_multiplier(), 1);
     }
 
     void test(send_func_t send_func, uint64_t op_mask,
@@ -144,6 +163,21 @@ protected:
     }
 
 private:
+    bool check_amo_odp_supported(const entity &e, ucs_memory_type_t mem_type)
+    {
+        const ucp_ep_config_key_t *key = &ucp_ep_config(e.ep())->key;
+        ucp_lane_index_t lane;
+
+        for (lane = 0; key->amo_lanes[lane] != UCP_NULL_LANE; ++lane) {
+            if (ucp_ep_md_attr(e.ep(), lane)->reg_nonblock_mem_types &
+                UCS_BIT(mem_type)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
     static T atomic_op_result(ucp_atomic_op_t op, T x, T y, T z) {
         switch (op) {
         case UCP_ATOMIC_OP_ADD:
@@ -208,6 +242,11 @@ private:
             if (!UCP_MEM_IS_HOST(send_mem_type) || !UCP_MEM_IS_HOST(recv_mem_type)) {
                 /* Memory type atomics are fully supported only with new protocols */
                 if (!is_proto_enabled()) {
+                    continue;
+                }
+
+                /* ODP is currently tested only for host memory */
+                if (get_variant_value() & ODP) {
                     continue;
                 }
 


test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -134,7 +134,7 @@ class test_ucp_atomic : public test_ucp_memheap {
     }
 
     static unsigned default_num_iters() {
-        return ucs_max(100 / ucs::test_time_multiplier(), 1);
+        return (128 * 1024) / ucs::test_time_multiplier();
```
tvegas1, line 137: for completeness, shall we still keep explicitly making sure it's > 1?

ivankochin: In many places where we use `test_time_multiplier` we don't compare result with 1 and I think it doesn't have sense to check it if number of iterations is 128K.

tvegas1 (dismissed):

yosefe (dismissed):

test/gtest/ucp/test_ucp_atomic.cc
brminich: i think @tvegas1 suggestion is valid, given that `ucs::test_time_multiplier()` can be changed later in a way that num_iters will be 0 here

ivankochin: I think to avoid such case we need to introduce something like `size_t ucs::iters_num(size_t input_num_iters)` which should care about dividing to `ucs::test_time_multiplier()` and compare result with 0 since as I said before there are many places where such error can occur and FMPOV it is better to fix them all together.

brminich: i do not see how having issuea in other places justify introducing one more, but it's a minor comment, feel free to ignore

brminich (dismissed):

test/gtest/ucp/test_ucp_atomic.cc
ivankochin:
```suggestion
        return ucs_max((128 * 1024) / ucs::test_time_multiplier(), 1);
```

yosefe (approved):

yosefe:
@ivankochin can you pls squash?

ivankochin:
> @ivankochin can you pls squash?

`test_ucp_atomic32.post` failures in CI seems related. I just don't have time right now to debug them. Is there any sense to squash if CI failures are relevant?

ivankochin:
> @ivankochin can you pls squash to trigger re-test? MPS should solve the timeouts

Done

yosefe:
@ivankochin seems there are too many iterations and causing a timeout (only in this PR), can we decrease and still reproduce the issue?
```
2024-08-29T17:16:48.3427583Z [----------] 3 tests from rcx/test_ucp_atomic64
2024-08-29T17:16:48.3428270Z [ RUN      ] rcx/test_ucp_atomic64.fetch/1 <rc_x,cuda_copy,rocm_copy/device>
2024-08-29T17:16:51.4343726Z [     INFO ] host->host ADD SWAP CSWAP AND OR XOR 
2024-08-29T17:17:02.9181288Z [     INFO ] host->cuda ADD SWAP CSWAP AND OR XOR 
2024-08-29T17:17:28.4053388Z [     INFO ] cuda->host ADD SWAP CSWAP AND OR XOR 
2024-08-29T17:18:01.0575721Z [     INFO ] cuda->cuda ADD SWAP CSWAP AND OR XOR 
2024-08-29T17:20:29.1133998Z [     INFO ] cuda-managed->host ADD SWAP CSWAP AND OR XOR 
2024-08-29T17:21:22.7518796Z [     INFO ] cuda-managed->cuda ADD SWAP CSWAP AND OR XOR 
2024-08-29T17:21:25.5912215Z [     INFO ] host->host ADD SWAP CSWAP AND OR XOR 
2024-08-29T17:21:38.9541593Z [     INFO ] host->cuda ADD SWAP CSWAP AND OR XOR 
2024-08-29T17:27:35.3484505Z [     INFO ] cuda->host ADD SWAP CSWAP AND OR XOR 
2024-08-29T17:31:48.2757437Z /__w/2/s/contrib/../test/gtest/common/test_helpers.cc:55: Failure
2024-08-29T17:31:48.2758213Z Failed
2024-08-29T17:31:48.2758757Z Connection timed out - abort testing
2024-08-29T17:31:48.2759706Z [swx-rdmz-ucx-gpu-02:49909:0:49909] Caught signal 6 (Aborted: tkill(2) or tgkill(2))
2024-08-29T17:31:48.2875618Z ==== backtrace (tid:  49909) ====
2024-08-29T17:31:48.2876571Z  0  /__w/2/s/build-test/src/ucs/.libs/libucs.so.0(ucs_handle_error+0x12c) [0x7f44bb37cdfc]
```

yosefe (approved):

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -124,17 +126,26 @@ class test_ucp_atomic : public test_ucp_memheap {
...
         const char *atomic_mode_cfg =
                         (atomic_mode == UCP_ATOMIC_MODE_CPU)    ? "cpu" :
                         (atomic_mode == UCP_ATOMIC_MODE_DEVICE) ? "device" :
                         (atomic_mode == UCP_ATOMIC_MODE_GUESS)  ? "guess" :
                         "";
         modify_config("ATOMIC_MODE", atomic_mode_cfg);
+
+        if (odp) {
+            modify_config("REG_NONBLOCK_MEM_TYPES", "host,cuda-managed");
+            modify_config("IB_ODP_MEM_TYPES", "host,cuda-managed",
```
brminich, line 139: afair, we can't set odp for cuda-managed on non-grace system

ivankochin: Thanks, removed.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -144,6 +144,8 @@ static void uct_ib_mlx5_devx_klm_entry_set(void **klm_p, size_t klm_idx,
               mr->addr, mr->length, mr->lkey);
     UCT_IB_MLX5DV_SET64(klm, klm, address, (uintptr_t)address);
     UCT_IB_MLX5DV_SET(klm, klm, mkey, mr->lkey);
+    UCT_IB_MLX5DV_SET(klm, klm, byte_count, mr->length);
```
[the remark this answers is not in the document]
ivankochin: After discussion with @Artemy-Mellanox it was decided that we need to set `UCT_IB_MD_MAX_MR_SIZE` as `byte_count` in single-thread case and `chunk_size` in multi-threaded since it should be the same with `entity_size` ksm argument.

yosefe (approved):
nice catch!!

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -232,6 +265,13 @@ class test_ucp_atomic : public test_ucp_memheap {
                 }
             }
 
+            /* Skip test if ODP requested but not supported */
+            if ((get_variant_value() & ODP) &&
```
ivankochin, line 269: This condition was introduced to avoid running device atomic + odp variant on bluefield device since IB devices don't support ODP there and device atomics disables SW atomic emulation so there is protocol to send message in that case.

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -196,6 +208,27 @@ class test_ucp_atomic : public test_ucp_memheap {
...
+
+        for (int i = 0; i < config->key.num_lanes; ++i) {
+            auto lane = config->key.amo_lanes[i];
+            if (lane == UCP_NULL_LANE) {
+                break;
+            }
+
+            auto md_index = ucp_ep_md_index(e.ep(), lane);
+            // std::cout << "lane " << lane << " md_index " << md_index << " ctx->reg_md_map[mem_type] " << ctx->reg_md_map[mem_type] << std::endl;
+            ucs_info("lane=%u md_index=%u ctx->reg_md_map[mem_type]=%zu", lane, md_index, ctx->reg_md_map[mem_type]);
```
brminich, line 223: leftovers

ivankochin, line 223:
```suggestion
            auto md_index = ucp_ep_md_index(e.ep(), lane);
```

test/gtest/ucp/test_ucp_atomic.cc
yosefe: Let's discuss it: if ODP is not supported, maybe we should skipp all ODP tests? But on other hand, this test is on UCP level, so maybe the issue is missing fallback logic in UCP?

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -28,6 +30,7 @@ class test_ucp_atomic : public test_ucp_memheap {
         add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_CPU, "cpu" + name);
         add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_DEVICE, "device" + name);
         add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_GUESS, "guess" + name);
+        add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_CPU | ODP, "cpu/odp" + name);
```
yosefe, line 33: is the mode of cpu atomics+ODP able to reproduce the original issue of KSM+ODP?
seems to me that no, since it will not issue atomic operations on KSM+ODP rkey, even when ODP is supported..
maybe we want to set the mode "guess" and adjust the logic of selecting the atomic mode to also consider ODP?

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -124,17 +127,31 @@ class test_ucp_atomic : public test_ucp_memheap {
...
                         (atomic_mode == UCP_ATOMIC_MODE_DEVICE) ? "device" :
                         (atomic_mode == UCP_ATOMIC_MODE_GUESS)  ? "guess" :
                         "";
         modify_config("ATOMIC_MODE", atomic_mode_cfg);
+
+        if (odp) {
+            // Do not test device atomics with ODP since SW atomic emulation
+            // is disabled with device atomics and there can be configurations
+            // without atomic lanes with ODP support which would lead to failure
+            ASSERT_NE(atomic_mode, UCP_ATOMIC_MODE_DEVICE);
```
yosefe, line 142: not sure this assert is needed. and i'd move the comment to get_test_variants

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -124,17 +127,31 @@ class test_ucp_atomic : public test_ucp_memheap {
...
+
+            modify_config("REG_NONBLOCK_MEM_TYPES", "host");
+            modify_config("IB_ODP_MEM_TYPES", "host",
+                          SETENV_IF_NOT_EXIST);
+            modify_config("IB_MLX5_DEVX_OBJECTS", "", SETENV_IF_NOT_EXIST);
+
```
yosefe, line 148: remove space line

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -124,17 +127,31 @@ class test_ucp_atomic : public test_ucp_memheap {
...
+        if (odp) {
+            // Do not test device atomics with ODP since SW atomic emulation
+            // is disabled with device atomics and there can be configurations
+            // without atomic lanes with ODP support which would lead to failure
+            ASSERT_NE(atomic_mode, UCP_ATOMIC_MODE_DEVICE);
+
+            modify_config("REG_NONBLOCK_MEM_TYPES", "host");
+            modify_config("IB_ODP_MEM_TYPES", "host",
+                          SETENV_IF_NOT_EXIST);
+            modify_config("IB_MLX5_DEVX_OBJECTS", "", SETENV_IF_NOT_EXIST);
```
yosefe, line 147: maybe devx enable/disable can be a test variant?

yosefe (commented):
since the test is debatable, i suggest to move the unit test to next PR, so we can merge the fix

ivankochin:
> since the test is debatable, i suggest to move the unit test to next PR, so we can merge the fix

Pls see fix without changing test here: https://github.com/openucx/ucx/pull/10215

test/gtest/ucp/test_ucp_atomic.cc
ivankochin: done

test/gtest/ucp/test_ucp_atomic.cc
ivankochin: done

test/gtest/ucp/test_ucp_atomic.cc
ivankochin: done

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -124,17 +131,29 @@ class test_ucp_atomic : public test_ucp_memheap {
...
         const char *atomic_mode_cfg =
                         (atomic_mode == UCP_ATOMIC_MODE_CPU)    ? "cpu" :
                         (atomic_mode == UCP_ATOMIC_MODE_DEVICE) ? "device" :
                         (atomic_mode == UCP_ATOMIC_MODE_GUESS)  ? "guess" :
                         "";
         modify_config("ATOMIC_MODE", atomic_mode_cfg);
+
+        if (odp) {
+            modify_config("REG_NONBLOCK_MEM_TYPES", "host");
+            modify_config("IB_ODP_MEM_TYPES", "host", SETENV_IF_NOT_EXIST);
```
yosefe, line 144: why needed?

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -124,17 +131,29 @@ class test_ucp_atomic : public test_ucp_memheap {
...
+            modify_config("IB_MLX5_DEVX_OBJECTS", "", SETENV_IF_NOT_EXIST);
+        }
         test_ucp_memheap::init();
+
+        if (odp && !(check_amo_odp_supported(sender(), UCS_MEMORY_TYPE_HOST))) {
+            cleanup();
```
yosefe, line 150: what is not cleaned up by the destructor?

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -144,6 +163,21 @@ class test_ucp_atomic : public test_ucp_memheap {
     }
 
 private:
+    bool check_amo_odp_supported(const entity &e, ucs_memory_type_t mem_type)
+    {
+        ucp_ep_config_key_t *key = &ucp_ep_config(e.ep())->key;
```
yosefe, line 168: const

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -211,6 +247,11 @@ class test_ucp_atomic : public test_ucp_memheap {
                     continue;
                 }
 
+                /* ODP variant affects only host buffer flow */
+                if (odp) {
```
yosefe, line 251:
```suggestion
                /* ODP is currently tested only for host memory */
                if (get_variant_value() & ODP) {
```

test/gtest/ucp/test_ucp_atomic.cc
```diff
@@ -28,6 +30,11 @@ class test_ucp_atomic : public test_ucp_memheap {
         add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_CPU, "cpu" + name);
         add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_DEVICE, "device" + name);
         add_variant_with_value(variants, features, variant | UCP_ATOMIC_MODE_GUESS, "guess" + name);
+
+        // Do not test device atomics with ODP since SW atomic emulation
+        // disabled with device atomics and there can be configurations
+        // without atomic lanes with ODP support which would lead to failure
```
yosefe, line 36:
```suggestion
        /* Do not force using device atomics when testing ODP, since some
         * configurations do not support it. Allow a fallback to SW emulation
         * in this case.
         */
```

test/gtest/ucp/test_ucp_atomic.cc
yosefe: maybe this warning (warn_existing_eps) should be moved to test_base::TearDownProxy, after calling cleanup:
https://github.com/openucx/ucx/blob/97b551b607ff7adfeadc0c2367a9688e5d95438f/test/gtest/common/test.cc#L345-L347

yosefe: decided to keep it as other test use same method, and fix all in another PR

yosefe (approved):

PR #10057, opened 2024-08-13
UCS/CONFIG: mask variables ignoring table prefix on set value

evgeny-leksikov:
## What
add pattern match to ucs_config

## Why ?
https://github.com/openucx/ucx/pull/9993#discussion_r1697056724

### DIFF
--- a/src/ucs/config/parser.c
+++ b/src/ucs/config/parser.c
@@ -1387,6 +1387,27 @@ ucs_config_parser_set_default_values(void *opts, ucs_config_field_t *fields)
     return UCS_OK;
 }
 
+static int
+ucs_config_prefix_name_match(const char *prefix, size_t prefix_len,
+                             const char *name, const char *pattern)
+{
+    const char *match_name;
+    char *full_name;
+    size_t full_name_len;
+
+    if (prefix_len == 0) {
+        match_name = name;
+    } else {
+        full_name_len = prefix_len + strlen(name) + 1;
+        full_name     = ucs_alloca(full_name_len);
+
+        ucs_snprintf_safe(full_name, full_name_len, "%s%s", prefix, name);
+        match_name = full_name;
+    }
+
+    return !fnmatch(pattern, match_name, 0);
+}
+
 /**
  * table_prefix == NULL  -> unused
  */
@@ -1417,8 +1438,8 @@ ucs_config_parser_set_value_internal(void *opts, ucs_config_field_t *fields,
             /* Check with sub-table prefix */
             if (recurse) {
                 status = ucs_config_parser_set_value_internal(var, sub_fields,
-                                                             name, value,
-                                                             field->name, 1);
+                                                              name, value,
+                                                              field->name, 1);
                 if (status == UCS_OK) {
                     ++count;
                 } else if (status != UCS_ERR_NO_ELEM) {
@@ -1429,17 +1450,16 @@ ucs_config_parser_set_value_internal(void *opts, ucs_config_field_t *fields,
             /* Possible override with my prefix */
             if (table_prefix != NULL) {
                 status = ucs_config_parser_set_value_internal(var, sub_fields,
-                                                             name, value,
-                                                             table_prefix, 0);
+                                                              name, value,
+                                                              table_prefix, 0);
                 if (status == UCS_OK) {
                     ++count;
                 } else if (status != UCS_ERR_NO_ELEM) {
                     return status;
                 }
             }
-        } else if (((table_prefix == NULL) || !strncmp(name, table_prefix, prefix_len)) &&
-                   !strcmp(name + prefix_len, field->name))
-        {
+        } else if (ucs_config_prefix_name_match(table_prefix, prefix_len,
+                                                field->name, name)) {
             if (ucs_config_is_deprecated_field(field)) {
                 return UCS_ERR_NO_ELEM;
             }
--- a/test/gtest/ucs/test_config.cc
+++ b/test/gtest/ucs/test_config.cc
@@ -61,6 +61,7 @@ typedef struct {
 } coach_opts_t;
 
 typedef struct {
+    const char      *model;
     unsigned        volume;
     unsigned long   power;
 } engine_opts_t;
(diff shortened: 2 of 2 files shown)


evgeny-leksikov:
unrelated failures will fixed by https://github.com/openucx/ucx/pull/10064

src/ucs/config/parser.c
```diff
@@ -1437,8 +1437,12 @@ ucs_config_parser_set_value_internal(void *opts, ucs_config_field_t *fields,
                     return status;
                 }
             }
-        } else if (((table_prefix == NULL) || !strncmp(name, table_prefix, prefix_len)) &&
-                   !strcmp(name + prefix_len, field->name))
+        } else if (/* check exact match */
```
iyastreb, line 1440: I propose to add a few simple unit tests in `test_config.cc` to cover this behaviour

evgeny-leksikov: good point, pls share your ideas about test cases

test/gtest/ucs/test_config.cc
```diff
@@ -329,8 +329,15 @@ class test_config : public ucs::test {
             delete [] m_value;
         }
 
-        ucs_status_t set(const char *name, const char *value) {
-            return ucs_config_parser_set_value(&m_opts, car_opts_table, NULL,
+        ucs_status_t set(const char *name, const char *value)
```
iyastreb, line 332: since prefix is optional arg, maybe put it last with default value:
```C++
ucs_status_t set(const char *name, const char *value, const char *prefix = NULL)
```

test/gtest/ucs/test_config.cc
```diff
@@ -579,6 +586,35 @@ UCS_TEST_F(test_config, set_get) {
...
+    ASSERT_UCS_OK(status);
+    ASSERT_EQ(model3, mazda.get("MODEL"));
+
+    // with prefix but ignore by mask
+    status = mazda.set("MAZDA_", "*MODEL", model6.c_str());
+    ASSERT_UCS_OK(status);
+    ASSERT_EQ(model6, mazda.get("MODEL"));
+
+    // without prefix and ignore by mask
+    status = mazda.set("*MODEL", modelcx5.c_str());
```
iyastreb, line 613: Here we set just a single config field, this is not very relevant coverage for wildcard IMO.
It seems this improvement is all about covering multiple values with single wildcard, right?
So this should be tested, e.g. by extending `car_opts_table` with another table with prefix and checking that wildcard setting is applied for both

src/ucs/config/parser.c
```diff
@@ -1429,16 +1429,21 @@ ucs_config_parser_set_value_internal(void *opts, ucs_config_field_t *fields,
...
+                   (((table_prefix == NULL) ||
+                     !strncmp(name, table_prefix, prefix_len)) &&
+                    !strcmp(name + prefix_len, field->name)) ||
+                   /* ignore prefix */
```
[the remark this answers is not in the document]
yosefe: maybe use "fnmatch" for checking?

test/gtest/ucs/test_config.cc
evgeny-leksikov: I prefer don't change logical ordering of args

iyastreb (approved):

src/ucs/config/parser.c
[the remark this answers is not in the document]
evgeny-leksikov: @yosefe It will add variability like fully supported wildcards, I'm not sure if this is what we want in addition to current behavior of prefix+name, up to you.

evgeny-leksikov:
@yosefe is it ok to squash?

src/ucs/config/parser.c
```diff
@@ -1429,16 +1429,20 @@ ucs_config_parser_set_value_internal(void *opts, ucs_config_field_t *fields,
...
                 }
             }
-        } else if (((table_prefix == NULL) || !strncmp(name, table_prefix, prefix_len)) &&
-                   !strcmp(name + prefix_len, field->name))
+        } else if (/* check exact match */
+                   (((table_prefix == NULL) ||
+                     !strncmp(name, table_prefix, prefix_len)) &&
+                    !strcmp(name + prefix_len, field->name)) ||
+                   /* check for wildcard */
+                   !fnmatch(name, field->name, 0))
```
yosefe, line 1445: IMO we should check the match between name and table_prefix concatenated with field->name

src/ucs/config/parser.c
```diff
@@ -1437,13 +1464,8 @@ ucs_config_parser_set_value_internal(void *opts, ucs_config_field_t *fields,
...
             }
-        } else if (/* check exact match */
-                   (((table_prefix == NULL) ||
-                     !strncmp(name, table_prefix, prefix_len)) &&
-                    !strcmp(name + prefix_len, field->name)) ||
-                   /* check for wildcard */
-                   !fnmatch(name, field->name, 0))
-        {
+        } else if (ucs_config_prefix_name_match(table_prefix, prefix_len,
+                                                field->name, name)) {
```
yosefe, line 1468: To simplify the code, WDYT?

```c
else if (((prefix_len == 0) && !fnmatch(name, field->name)) ||
         ucs_config_prefix_name_match(table_prefix, prefix_len,
                                      field->name, name)) {
}
```

and rename ucs_config_prefix_name_match_no_check to ucs_config_prefix_name_match

src/ucs/config/parser.c
```diff
@@ -1387,6 +1387,33 @@ ucs_config_parser_set_default_values(void *opts, ucs_config_field_t *fields)
     return UCS_OK;
 }
 
+static int
+ucs_config_name_match_no_check(const char *name, const char *pattern)
+{
+    return !fnmatch(pattern, name, 0);
+}
```
yosefe, line 1394: IMO this wrapper is redundant

evgeny-leksikov: this wrapper was added:
1) to improve readability of `ucs_config_prefix_name_match`
2) to have external `fnmatch` call and matching ligic in single place since `fnmatch` has a perameter which changes its behaviour (can be changed in one place and missed in other one)

basically, all above is minor polishing, so the wrapper can be removed if you are disagree

src/ucs/config/parser.c
evgeny-leksikov: I think it does not look like simplifucation, in this case:
1) name matching ligic is brocken up in two places
2) it complecates entire `ucs_config_parser_set_value_internal` function and `if () else if ()` statements

yosefe (dismissed):

evgeny-leksikov:
squashed + small code style fix

PR #10058, opened 2024-08-13
CI: Add codespell check job

ivankochin:
## What
Add spell check to CI

## Why ?
To reduce number of spelling issues.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,3 @@
+[codespell]
+skip = ./test/gtest/common,./src/ucs/datastruct/sglib.h,./docs/doxygen/Architecture*,./docs/doxygen/header.tex.in,./src/ucm/ptmalloc286,./config/m4/ax_prog_doxygen.m4,./VENV/*
+ignore-words-list = hsa,inout,scoll,warmup,pullrequest,shft
--- a/buildlib/pr/codestyle.yml
+++ b/buildlib/pr/codestyle.yml
@@ -70,3 +70,23 @@ jobs:
               azure_complete_with_issues "Code is not formatted according to the code style, see $url for more info."
           fi
         condition: eq(variables['Build.Reason'], 'PullRequest')
+
+  - job: codespell
+    displayName: codespell check
+    pool:
+      name: MLNX
+      demands:
+      - ucx_docker -equals yes
+    container: fedora
+    steps:
+      - checkout: self
+        clean: true
+        fetchDepth: 100
+        retryCountOnTaskFailure: 5
+
+      - bash: |
+          python3 -m venv /tmp/codespell_env
+          source /tmp/codespell_env/bin/activate
+          pip3 install codespell
+          codespell
+        displayName: codespell test
--- a/bindings/go/src/ucx/worker.go
+++ b/bindings/go/src/ucx/worker.go
@@ -97,7 +97,7 @@ func (w *UcpWorker) Query(attrs ...UcpWorkerAttribute) (*UcpWorkerAttributes, er
 // is over, as part of the wake-up mechanism.
 //
 // The worker must be armed before waiting on an event (must be re-armed after
-// it has been signaled for re-use) with UcpWorker.Arm().
+// it has been signaled for reuse) with UcpWorker.Arm().
 // The events triggering a signal of the file descriptor from
 // UcpWorker.GetEfd() depend on the interfaces used by the worker and
 // defined in the transport layer, and typically represent a request completion
--- a/bindings/java/src/main/java/org/openucx/jucx/ucp/UcpWorker.java
+++ b/bindings/java/src/main/java/org/openucx/jucx/ucp/UcpWorker.java
@@ -190,7 +190,7 @@ public class UcpWorker extends UcxNativeStruct implements Closeable {
      * as part of the wake-up mechanism.
      *
      * The worker must be armed before waiting on an event (must be re-armed after it has been
-     * signaled for re-use) with {@link UcpWorker#arm()}. The events triggering a signal of the
+     * signaled for reuse) with {@link UcpWorker#arm()}. The events triggering a signal of the
      * file descriptor from {@link UcpWorker#getEventFD()} depend on the interfaces used by
      * the worker and defined in the transport layer, and typically represent a request completion
      * or newly available resources. It can also be triggered by calling {@link UcpWorker#signal()}.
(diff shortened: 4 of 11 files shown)


buildlib/pr/codestyle.yml
```diff
@@ -70,3 +70,22 @@ jobs:
...
+      - ucx_docker -equals yes
+    container: fedora
+    steps:
+      - checkout: self
+        clean: true
+        fetchDepth: 100
+        retryCountOnTaskFailure: 5
+
+      - bash: |
+          sudo pip3 install codespell
```
tvegas1, line 88: maybe instead add those commands as a shell script so that we can run it on the development machine too?

ivankochin: Now these commands take 4 lines, and it can be reduced to two lines (pip install + codespell run). Do you think that running a script would significantly simplify running on development machine against running two commands?

tvegas1: up to you, but it cannot hurt, specially since scripts start small and grow with the time.

yosefe: can we make it part of the docker image?
or use VENV like here: https://github.com/openucx/ucx/blob/d346ea2c58b1aa40a882e0b94a46ace43aa0d647/buildlib/tools/perf-common.yml#L31-L35
To avoid changing the root file system

ivankochin: Yes, we can, but what is the problem with installing package inside container?

yosefe: IMO it may be error prone to be run accidentally without container

buildlib/pr/codestyle.yml
```diff
@@ -70,3 +70,22 @@ jobs:
...
+          export SKIPPED_DIRS="./test/gtest/common,./src/ucs/datastruct/sglib.h,./docs/doxygen/Architecture*,./docs/doxygen/header.tex.in,./src/ucm/ptmalloc286,./config/m4/ax_prog_doxygen.m4"
+          export IGNORED_WORDS="hsa,inout,scoll,warmup,pullrequest,re-use,shft"
```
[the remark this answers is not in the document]
yosefe: IMO better add a codespell config file - .codespellrc
https://github.com/codespell-project/codespell?tab=readme-ov-file#using-a-config-file

buildlib/pr/codestyle.yml
ivankochin: Done

buildlib/pr/codestyle.yml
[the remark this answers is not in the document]
ivankochin: > maybe change re-use to reuse if it is treated as a typo
Ok, done

> IMO better add a codespell config file - .codespellrc
Added .codespellrc

.codespellrc
```diff
@@ -0,0 +1,3 @@
+[codespell]
+skip = ./test/gtest/common,./src/ucs/datastruct/sglib.h,./docs/doxygen/Architecture*,./docs/doxygen/header.tex.in,./src/ucm/ptmalloc286,./config/m4/ax_prog_doxygen.m4,./VENV/*
+ignore-words-list = hsa,inout,scoll,warmup,pullrequest,re-use,shft
```
ivankochin, line 3:
```suggestion
ignore-words-list = hsa,inout,scoll,warmup,pullrequest,shft
```

tvegas1 (approved):

ivankochin:
@yosefe @tvegas1 @brminich pls review

buildlib/pr/codestyle.yml
```diff
@@ -70,3 +70,25 @@ jobs:
...
+    pool:
+      name: MLNX
+      demands:
+      - ucx_docker -equals yes
+    container: fedora
+    steps:
+      - checkout: self
+        clean: true
+        fetchDepth: 100
+        retryCountOnTaskFailure: 5
```
brminich, line 85: it's a minor comment, but maybe we do not need to set this values to be that large. @Alexey-Rivkin what are the defaults for `fetch` and `retry`? Can we just use defaults here?

tvegas1 (approved):

buildlib/pr/codestyle.yml
Alexey-Rivkin: The default for retry is 0, we made it 5 to overcome pure connectivity issues.
A fetch depth of 100 is Ok too, less might fail with the "commit not found" error if the PR was shelved for a while.

buildlib/pr/codestyle.yml
```diff
@@ -70,3 +70,25 @@ jobs:
...
+
+      - bash: |
+          python3 -m venv /tmp/codespell_env
+          source /tmp/codespell_env/bin/activate
+          pip3 install codespell
+          codespell
+          test_exit_code=$?
```
Alexey-Rivkin, line 92: Do we really need to store an exit code?
If `codespell` command fails, Azure will fail the job.

buildlib/pr/codestyle.yml
```diff
@@ -70,3 +70,25 @@ jobs:
...
+          rm -rf /tmp/codespell_env
```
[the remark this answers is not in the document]
ivankochin: Thanks, removed cleanup

buildlib/pr/codestyle.yml
ivankochin: That was needed because of env cleanup, now it should work without saving exit code

buildlib/pr/codestyle.yml
```diff
@@ -70,3 +70,25 @@ jobs:
...
+      demands:
+      - ucx_docker -equals yes
+    container: fedora
+    steps:
+      - checkout: self
+        clean: true
+        fetchDepth: 100
+        retryCountOnTaskFailure: 5
+
+      - bash: |
```
Alexey-Rivkin, line 87: Cosmetic: It's advised to set a `displayName` per step.

ivankochin: Done

Alexey-Rivkin (approved):

buildlib/pr/codestyle.yml
brminich: is it really needed. I see no other job in this file do that

Alexey-Rivkin:
```
arivkin@mtr-vdi-007:ucx (test_PR) $ grep -r displayName * | wc -l     
101
```
Some do it right.
I would add displayName to other steps in this file as well.

ivankochin: AFAIU it is a cosmetic change, so it is not strongly required, do you propose to remove it?

brminich (approved):

ivankochin:
@yosefe pls review

yosefe (approved):
pls sqaush

PR #10061, opened 2024-08-14
Test coverity failure

tvegas1:
## What
Make CI coverity fail to confirm.


tvegas1:
confirmed that coverity failures are still triggered

PR #10062, opened 2024-08-15
UCT/MD/IB: Add ODP support for verbs and mlx5dv

ivankochin:
## What
Support ODPv1 for verbs and mlx5dv MDs.

## Why ?
To provide workaround in case of issues with DEVX ODP support.

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -1203,6 +1203,35 @@ static void uct_ib_md_check_dmabuf(uct_ib_md_t *md)
 #endif
 }
 
+int uct_ib_md_check_odp_common(uct_ib_md_t *md, const char **reason_ptr)
+{
+    if (IBV_ACCESS_ON_DEMAND == 0) {
+        *reason_ptr = "IBV_ACCESS_ON_DEMAND is not supported";
+        return 0;
+    }
+
+    if (!IBV_DEVICE_HAS_ODP(&md->dev)) {
+        *reason_ptr = "device does not support IBV_ACCESS_ON_DEMAND";
+        return 0;
+    }
+
+    return 1;
+}
+
+void uct_ib_md_check_odp(uct_ib_md_t *md)
+{
+    const char *device_name = uct_ib_device_name(&md->dev);
+    const char *reason;
+
+    if (!uct_ib_md_check_odp_common(md, &reason)) {
+        ucs_debug("%s: ODP is disabled because %s", device_name, reason);
+        return;
+    }
+
+    md->reg_nonblock_mem_types = UCS_BIT(UCS_MEMORY_TYPE_HOST);
+    ucs_debug("%s: ODP is supported, version 1", device_name);
+}
+
 ucs_status_t uct_ib_md_open_common(uct_ib_md_t *md,
                                    struct ibv_device *ib_device,
                                    const uct_ib_md_config_t *md_config)
@@ -1463,6 +1492,7 @@ static ucs_status_t uct_ib_verbs_md_open(struct ibv_device *ibv_device,
 
     uct_ib_md_ece_check(md);
     uct_ib_md_parse_relaxed_order(md, md_config, 0);
+    uct_ib_md_check_odp(md);
 
     *p_md = md;
     return UCS_OK;
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -1609,13 +1609,7 @@ static void uct_ib_mlx5_devx_check_odp(uct_ib_mlx5_md_t *md,
     struct ibv_mr *mr;
     uint8_t version;
 
-    if (IBV_ACCESS_ON_DEMAND == 0) {
-        reason = "IBV_ACCESS_ON_DEMAND is not supported";
-        goto no_odp;
-    }
-
-    if (!IBV_DEVICE_HAS_ODP(&md->super.dev)) {
-        reason = "device does not support ODP";
+    if (!uct_ib_md_check_odp_common(&md->super, &reason)) {
         goto no_odp;
     }
 
@@ -3199,6 +3193,7 @@ static ucs_status_t uct_ib_mlx5dv_md_open(struct ibv_device *ibv_device,
 
     uct_ib_md_parse_relaxed_order(&md->super, md_config, 0);
     uct_ib_md_ece_check(&md->super);
+    uct_ib_md_check_odp(&md->super);
 
     md->super.flush_rkey = uct_ib_mlx5_flush_rkey_make();
 
--- a/test/gtest/ucp/test_ucp_mmap.cc
+++ b/test/gtest/ucp/test_ucp_mmap.cc
@@ -106,6 +106,11 @@ public:
             modify_config("PROTO_ENABLE", "n");
         }
 
+        if (get_variant_value() == VARIANT_MAP_NONBLOCK) {
+            // ODPv1 cannot interact with DEVX objects
+            modify_config("IB_MLX5_DEVX_OBJECTS", "", SETENV_IF_NOT_EXIST);
+        }
+
         if (get_variant_value() == VARIANT_NO_RCACHE) {
             modify_config("RCACHE_ENABLE", "n");
             ucp_test::init(); // Init UCP with rcache disabled
@@ -547,7 +552,11 @@ UCS_TEST_P(test_ucp_mmap, reg_mem_type) {
 
     for (int i = 0; i < 1000 / ucs::test_time_multiplier(); ++i) {
         size_t size    = ucs::rand() % UCS_MBYTE;
-        alloc_mem_type = mem_types.at(ucs::rand() % mem_types.size());
+        auto flags     = mem_map_flags();
+        /* Test ODP registration with host buffer only */
+        alloc_mem_type = (flags & VARIANT_MAP_NONBLOCK) ?
+                UCS_MEMORY_TYPE_HOST :
+                mem_types.at(ucs::rand() % mem_types.size());
         mem_buffer buf(size, alloc_mem_type);
         mem_buffer::pattern_fill(buf.ptr(), size, 0, alloc_mem_type);
 
@@ -561,7 +570,7 @@ UCS_TEST_P(test_ucp_mmap, reg_mem_type) {
         params.address     = buf.ptr();
         params.length      = size;
         params.memory_type = alloc_mem_type;
-        params.flags       = mem_map_flags();
+        params.flags       = flags;
 
         status = ucp_mem_map(sender().ucph(), &params, &memh);
         ASSERT_UCS_OK(status);
@@ -943,8 +952,7 @@ UCS_TEST_P(test_ucp_mmap, fixed) {
     }
 }
 
-UCS_TEST_P(test_ucp_mmap, gva_allocate, "GVA_ENABLE=y",
-           "IB_MLX5_DEVX_OBJECTS?=")
+UCS_TEST_P(test_ucp_mmap, gva_allocate, "GVA_ENABLE=y")
 {
     for (auto mem_type : mem_buffer::supported_mem_types()) {
         ucp_md_map_t md_map = sender().ucph()->gva_md_map[mem_type];
@@ -969,7 +977,7 @@ UCS_TEST_P(test_ucp_mmap, gva_allocate, "GVA_ENABLE=y",
     }
 }
 
-UCS_TEST_P(test_ucp_mmap, gva, "GVA_ENABLE=y", "IB_MLX5_DEVX_OBJECTS?=")
+UCS_TEST_P(test_ucp_mmap, gva, "GVA_ENABLE=y")
 {
     std::list<void*> bufs;
     ucp_mem_h first     = NULL;
--- a/test/gtest/uct/ib/test_ib_md.cc
+++ b/test/gtest/uct/ib/test_ib_md.cc
@@ -311,3 +311,18 @@ UCS_TEST_P(test_ib_md, smkey_reg_atomic_mt, "IB_REG_MT_THRESH=1k",
 }
 
 _UCT_MD_INSTANTIATE_TEST_CASE(test_ib_md, ib)
+
+class test_ib_md_non_blocking : public test_md_non_blocking {
+};
+
+UCS_TEST_P(test_ib_md_non_blocking, reg_advise_odp_no_devx, "IB_MLX5_DEVX=no")
+{
+    test_nb_reg_advise();
+}
+
+UCS_TEST_P(test_ib_md_non_blocking, reg_odp_no_devx, "IB_MLX5_DEVX=no")
+{
+    test_nb_reg();
+}
+
+_UCT_MD_INSTANTIATE_TEST_CASE(test_ib_md_non_blocking, ib)
--- a/test/gtest/uct/test_md.cc
+++ b/test/gtest/uct/test_md.cc
@@ -1137,44 +1137,16 @@ UCS_TEST_P(test_md_memlock_limit, md_open)
 
 UCT_MD_INSTANTIATE_TEST_CASE(test_md_memlock_limit)
 
-class test_md_non_blocking : public test_md
-{
-protected:
-    void init() override {
-        /* ODPv1 IB feature can work only for certain DEVX configuration */
-        modify_config("IB_MLX5_DEVX_OBJECTS", "dct,dcsrq", IGNORE_IF_NOT_EXIST);
-        test_md::init();
-    }
-};
-
 UCS_TEST_SKIP_COND_P(test_md_non_blocking, reg_advise,
                      !check_caps(UCT_MD_FLAG_REG | UCT_MD_FLAG_ADVISE))
 {
-    test_reg_advise(UCS_KBYTE, UCS_KBYTE, 0, true);
-    test_reg_advise(UCS_KBYTE, UCS_KBYTE / 2, 0, true);
-    test_reg_advise(UCS_KBYTE, UCS_KBYTE / 2, UCS_KBYTE / 4, true);
-
-    /*
-     * TODO: These tests should be enabled
-     * when https://redmine.mellanox.com/issues/336376 fixed
-
-     * test_reg_advise(UCS_MBYTE, UCS_MBYTE, 0, true);
-     * test_reg_advise(UCS_MBYTE, UCS_MBYTE / 2, 0, true);
-     * test_reg_advise(UCS_MBYTE, UCS_MBYTE / 2, UCS_MBYTE / 4, true);
-     */
+    test_nb_reg_advise();
 }
 
 UCS_TEST_SKIP_COND_P(test_md_non_blocking, reg,
                      !check_caps(UCT_MD_FLAG_REG))
 {
-    test_reg_advise(UCS_KBYTE, 0, 0, true);
-
-    /*
-     * TODO: This test should be enabled
-     * when https://redmine.mellanox.com/issues/336376 fixed
-
-     * test_reg_advise(UCS_MBYTE, 0, 0, true);
-     */
+    test_nb_reg();
 }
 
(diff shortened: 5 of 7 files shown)


src/uct/ib/base/ib_md.c
```diff
@@ -1192,6 +1192,31 @@ static void uct_ib_md_check_dmabuf(uct_ib_md_t *md)
 #endif
 }
 
+static void uct_ib_md_check_odp(uct_ib_md_t *md)
+{
+    uct_md_mem_reg_params_t params = {};
+    const static size_t buffer_size = 8;
```
brminich, line 1198:
```suggestion
    uct_md_mem_reg_params_t params  = {};
    const static size_t buffer_size = 8;
```

src/uct/ib/base/ib_md.c
```diff
@@ -1192,6 +1192,31 @@ static void uct_ib_md_check_dmabuf(uct_ib_md_t *md)
...
+    char buffer[buffer_size];
```
brminich, line 1199: why not just `uint64_t`?

src/uct/ib/base/ib_md.c
```diff
@@ -1192,6 +1192,31 @@ static void uct_ib_md_check_dmabuf(uct_ib_md_t *md)
...
+
+    if (IBV_ACCESS_ON_DEMAND == 0) {
+        ucs_debug("IBV_ACCESS_ON_DEMAND is not supported");
+        return;
+    }
+
+    status = uct_ib_reg_mr(md, buffer, buffer_size, &params,
+                           IBV_ACCESS_ON_DEMAND, NULL, &mr);
+    if (status != UCS_OK) {
+        ucs_debug("%s: ODP is not supported", uct_ib_device_name(&md->dev));
```
brminich, line 1211: does it really fail if ODP is not supported or just silently ignored?

test/gtest/uct/test_md.cc
```diff
@@ -1036,6 +1047,63 @@ UCS_TEST_SKIP_COND_P(test_md, rkey_compare,
...
+    void init() override {
+        modify_config("IB_MLX5_DEVX", "no");
+    }
+};
+
+UCS_TEST_SKIP_COND_P(test_md_odp_no_devx, reg, !check_caps(UCT_MD_FLAG_REG)) {
+    test_odp();
+}
+
+_UCT_MD_INSTANTIATE_TEST_CASE(test_md_odp_no_devx, ib)
```
brminich, line 1105: seems to be ib specific, why is it in common `test_md`?

test/gtest/uct/test_md.cc
```diff
@@ -1036,6 +1047,63 @@ UCS_TEST_SKIP_COND_P(test_md, rkey_compare,
...
+                            std::string(GetParam().md_name));
+        }
+
+        for (unsigned i = 1; i <= 300; ++i) {
+            size_t size = UCS_KBYTE * i;
+
+            std::vector<char> fill_buffer(size, 0);
+            ucs::fill_random(fill_buffer);
+
+            alloc_memory(&address, size, &fill_buffer[0], UCS_MEMORY_TYPE_HOST);
```
brminich, line 1070: use `mem_type` or remove that var

test/gtest/uct/test_md.cc
```diff
@@ -1036,6 +1047,63 @@ UCS_TEST_SKIP_COND_P(test_md, rkey_compare,
 
 UCT_MD_INSTANTIATE_TEST_CASE(test_md)
 
+class test_md_odp : public test_md {
```
brminich, line 1050: why is this test needed?
does it fail without your changes?

src/uct/ib/base/ib_md.c
ivankochin: In that case I just copied behavior from `uct_ib_mlx5_devx_check_odp` so ODP is not supported it will print debug message and leave `reg_nonblock_mem_types` empty.

src/uct/ib/base/ib_md.c
ivankochin: NA

src/uct/ib/base/ib_md.c
ivankochin: Changed

test/gtest/uct/test_md.cc
ivankochin: I found that ODP is tested already, so test part was significantly changed

ivankochin:
`test_ucp_mmap.reg_mem_type` probably will fail with cuda memtype even with devx after last commit. That case test tries to register CUDA memory with `UCP_MEM_MAP_NONBLOCK` flag through `ucp_mem_map` call. It passed before since ODP never worked there because of existing DEVX objects.

It fails since `context->reg_md_map[mem_type] ` fills by `md_attr->reg_mem_types` IB MD is inside reg_md_map for CUDA memory. 

This causes `uct_ib_mlx5_devx_mem_reg` call for CUDA memory, which calls `uct_ib_memh_alloc` that set `UCT_IB_MEM_FLAG_ODP` if `md->reg_nonblock_mem_types` contains **UCS_MEMORY_TYPE_HOST** despite the fact that we are registering CUDA memory. So that leads to call `ibv_reg_mr` with `IBV_ACCESS_ON_DEMAND` flag for CUDA buffer.

Question is what would be the expected behavior for passing CUDA memory to `ucp_mem_map` and `uct_md_mem_reg_v2`?

Variants that come to my mind:
- Do not register memory using this MD (Applicable for `ucp_mem_map`)
- Gracefully fail with meaningful message (I would expect that for `uct_md_mem_reg_v2`, but this call knows nothing about memory type right now)
- Register without ODP flag


@yosefe, @brminich WDYT?

yosefe:
> `test_ucp_mmap.reg_mem_type` probably will fail with cuda memtype even with devx after last commit. That case test tries to register CUDA memory with `UCP_MEM_MAP_NONBLOCK` flag through `ucp_mem_map` call. It passed before since ODP never worked there because of existing DEVX objects.
> 
> It fails since `context->reg_md_map[mem_type] ` fills by `md_attr->reg_mem_types` IB MD is inside reg_md_map for CUDA memory.
> 
> This causes `uct_ib_mlx5_devx_mem_reg` call for CUDA memory, which calls `uct_ib_memh_alloc` that set `UCT_IB_MEM_FLAG_ODP` if `md->reg_nonblock_mem_types` contains **UCS_MEMORY_TYPE_HOST** despite the fact that we are registering CUDA memory. So that leads to call `ibv_reg_mr` with `IBV_ACCESS_ON_DEMAND` flag for CUDA buffer.
> 
> Question is what would be the expected behavior for passing CUDA memory to `ucp_mem_map` and `uct_md_mem_reg_v2`?
> 
> Variants that come to my mind:
> 
> * Do not register memory using this MD (Applicable for `ucp_mem_map`)
> * Gracefully fail with meaningful message (I would expect that for `uct_md_mem_reg_v2`, but this call knows nothing about memory type right now)
> * Register without ODP flag
> 
> @yosefe, @brminich WDYT?

I would expect UCP to not register with NONBLOCK a memory type that is not part of reg_nonblock_mem_types for that MD. Here there are 2 options: either don't register on this MD, or register without NONBLOCK. Maybe we can do it according to UCX_REG_NONBLOCK_MEM_TYPES config: if the mem type is in UCX_REG_NONBLOCK_MEM_TYPES, don't register it. if it's not there, register it without NONBLOCK (ODP)

yosefe:
> I would expect UCP to not register with NONBLOCK a memory type that is not part of reg_nonblock_mem_types for that MD. Here there are 2 options: either don't register on this MD, or register without NONBLOCK. Maybe we can do it according to UCX_REG_NONBLOCK_MEM_TYPES config: if the mem type is in UCX_REG_NONBLOCK_MEM_TYPES, don't register it. if it's not there, register it without NONBLOCK (ODP)

And in addition i would check in UCT layer that NONBLOCK flag is not passed when the mem type is not in reg_nonblock_mem_types (in md_addt_v2)

ivankochin:
> > I would expect UCP to not register with NONBLOCK a memory type that is not part of reg_nonblock_mem_types for that MD. Here there are 2 options: either don't register on this MD, or register without NONBLOCK. Maybe we can do it according to UCX_REG_NONBLOCK_MEM_TYPES config: if the mem type is in UCX_REG_NONBLOCK_MEM_TYPES, don't register it. if it's not there, register it without NONBLOCK (ODP)
> 
> And in addition i would check in UCT layer that NONBLOCK flag is not passed when the mem type is not in reg_nonblock_mem_types (in md_addt_v2)

At this moment I decided to disable testing for CUDA buffers with ODP, I think this problem worth creation separate PR to cover it.

test/gtest/ucp/test_ucp_mmap.cc
```diff
@@ -547,7 +552,11 @@ UCS_TEST_P(test_ucp_mmap, reg_mem_type) {
 
     for (int i = 0; i < 1000 / ucs::test_time_multiplier(); ++i) {
         size_t size    = ucs::rand() % UCS_MBYTE;
-        alloc_mem_type = mem_types.at(ucs::rand() % mem_types.size());
+        auto flags     = mem_map_flags();
+        /* Test ODP registration with host buffer only */
+        alloc_mem_type = (flags & VARIANT_MAP_NONBLOCK) ?
+                UCS_MEMORY_TYPE_HOST :
```
brminich, line 558: so the issue with cuda mem still exists right?

test/gtest/uct/test_md.cc
```diff
@@ -1150,31 +1153,19 @@ class test_md_non_blocking : public test_md
...
-
-    /*
-     * TODO: This test should be enabled
-     * when https://redmine.mellanox.com/issues/336376 fixed
-
-     * test_reg_advise(UCS_MBYTE, 0, 0, true);
-     */
+    for (auto size: {UCS_KBYTE, UCS_MBYTE}) {
+        test_reg_advise(size, 0, 0, true);
+    }
```
brminich, line 1168: this can be a member of `test_md` and `test_ib_md` could also use it.

ivankochin: I am not sure that I got your point. You propose to create new function in `test_md` that run `test_reg_advise` for different sizes and call it from `test_ib_md`?

src/uct/ib/base/ib_md.c
```diff
@@ -1192,6 +1192,30 @@ static void uct_ib_md_check_dmabuf(uct_ib_md_t *md)
 #endif
 }
 
+void uct_ib_md_check_odp(uct_ib_md_t *md)
```
yosefe, line 1195: There is Verbs API to check ODP, https://docs.nvidia.com/networking/display/mlnxofedv492240/optimized+memory+access#src-39263878_OptimizedMemoryAccess-QueryCapabilities.1
why not use it?

test/gtest/ucp/test_ucp_mmap.cc
```diff
@@ -106,6 +106,11 @@ class test_ucp_mmap : public ucp_test {
             modify_config("PROTO_ENABLE", "n");
         }
 
+        if (get_variant_value() == VARIANT_MAP_NONBLOCK) {
+            // ODPv1 cannot interact with DEVX objects
+            modify_config("IB_MLX5_DEVX_OBJECTS", "", SETENV_IF_NOT_EXIST);
```
yosefe, line 111: 1. why needed? i would expect ODP would be reported as not supported by devx md if it's only odpv1, but still be supported by mlx5dv/verbs md
2. why SETENV_IF_NOT_EXIST and not SKIP_IF_NOT_EXIST ?

test/gtest/ucp/test_ucp_mmap.cc
yosefe: 1. we should randomize the memory type based on md_attr.reg_nb_mem_types, not just assume host mem and only host mem is always supported
2. seems this is an existing issue with the test, why did it not happen before this PR?

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -310,4 +310,22 @@ UCS_TEST_P(test_ib_md, smkey_reg_atomic_mt, "IB_REG_MT_THRESH=1k",
...
 }
 
+UCS_TEST_P(test_ib_md, reg_advise_odp_no_devx, "IB_MLX5_DEVX=no")
+{
+    ASSERT_EQ(md_attr().reg_nonblock_mem_types, UCS_BIT(UCS_MEMORY_TYPE_HOST));
+    for (auto size: {UCS_KBYTE, UCS_MBYTE}) {
+        test_reg_advise(size, size, 0, true);
+        test_reg_advise(size, size / 2, 0, true);
+        test_reg_advise(size, size / 2, size / 4, true);
+    }
```
yosefe, line 320: maybe move this common part to a helper function in test_md.h/.cc?

test/gtest/uct/test_md.cc
brminich: you have wholly copy pasted tests in two classes. Imo, making the test a member function in a base class is worth it

test/gtest/ucp/test_ucp_mmap.cc
ivankochin: 1. `test_ucp_mmap` is UCP test which has ODP variant for all test cases, so we don't specify `DEVX_OBJECTS` for all test cases it would use DEVX MD by default but without ODP support (if we are talking about ODPv1 which is the only option now). So in that general case without setting this variable ODP flag would be just ignored which is probably not what we want to test right here.
2. AFAIR `SETENV_IF_NOT_EXIST` wouldn't work since at that point it has access only to UCP config so we need to set ENV variable so it would be taken into account during UCT initialization. I will double check it.

test/gtest/ucp/test_ucp_mmap.cc
ivankochin: 1. That's UCP test which is working with many MDs. So do you mean to iterate somehow through all MDs and collect map from all the supported MDs?
2. That test had no `DEVX_OBJECTS` variable set, so ODP flag was disabled for IB MD.

src/uct/ib/base/ib_md.c
ivankochin: Am I missing anything or the only API to check it which is present there is `ibv_exp_query_device` which is part of experimental verbs API which was completely removed from UCX some time ago?

test/gtest/uct/test_md.cc
ivankochin: Unified

test/gtest/uct/ib/test_ib_md.cc
ivankochin: Done

src/uct/ib/base/ib_md.c
yosefe: right, this pointer to exp verbs is irrelevant, pls see https://github.com/linux-rdma/rdma-core/blob/dff9ddeee32e74152230313cb021b1a722cdcc06/libibverbs/verbs.h#L347

test/gtest/ucp/test_ucp_mmap.cc
yosefe: 2. there is code in UCP to forward config to UCT

yosefe:
@ivankochin the test failures seems specific to this PR

test/gtest/ucp/test_ucp_mmap.cc
ivankochin: 2. Are you sure that we really have the code which propagates config from UCP to UCT level? I tried to find it or check by debugger which tables are affected by `ucp_config_modify_internal` but had no success. BTW why `SETENV_IF_NOT_EXIST ` looks worse than `SKIP_IF_NOT_EXIST` from your point of view?

src/uct/ib/base/ib_md.c
ivankochin: If I understand the things correctly this structure is already used to check ODP in DEVX related code (see `IBV_DEVICE_HAS_ODP`) but there is only one of many checks that helps us to detect ODP feature and it seems like in DEVX MD case this check is not enough to tell whether ODP available or not. 

So the question is whether checking this field would be enough to be sure that ODP is supported in verbs MD case? And if yes, why it is not enough to check that field only for DEVX MD case?

src/uct/ib/base/ib_md.c
```diff
@@ -1203,6 +1203,34 @@ static void uct_ib_md_check_dmabuf(uct_ib_md_t *md)
...
+
+    if (uct_ib_md_check_odp_common(md, &reason)) {
+        md->reg_nonblock_mem_types = UCS_BIT(UCS_MEMORY_TYPE_HOST);
+        ucs_debug("%s: ODP is supported, version 1",
+                  uct_ib_device_name(&md->dev));
+    } else {
+        ucs_debug("%s: ODP is disabled because %s",
+                  uct_ib_device_name(&md->dev), reason);
+    }
+}
```
yosefe, line 1232:
```suggestion
    if (!uct_ib_md_check_odp_common(md, &reason)) {
        ucs_debug("%s: ODP is disabled because %s",
                  uct_ib_device_name(&md->dev), reason);
        return;
    }

    md->reg_nonblock_mem_types = UCS_BIT(UCS_MEMORY_TYPE_HOST);
    ucs_debug("%s: ODP is supported, version 1",
              uct_ib_device_name(&md->dev));
}
```

src/uct/ib/base/ib_md.c
```diff
@@ -1203,6 +1203,34 @@ static void uct_ib_md_check_dmabuf(uct_ib_md_t *md)
...
 }
 
+int uct_ib_md_check_odp_common(uct_ib_md_t *md, const char **reason_ptr) {
+    if (IBV_ACCESS_ON_DEMAND == 0) {
+        *reason_ptr = "IBV_ACCESS_ON_DEMAND is not supported";
+        return 0;
+    }
+
+    if (!IBV_DEVICE_HAS_ODP(&md->dev)) {
+        *reason_ptr = "device does not support ODP";
```
yosefe, line 1213: "device does not support IBV_ACCESS_ON_DEMAND"

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1609,13 +1609,7 @@ static void uct_ib_mlx5_devx_check_odp(uct_ib_mlx5_md_t *md,
     struct ibv_mr *mr;
     uint8_t version;
 
-    if (IBV_ACCESS_ON_DEMAND == 0) {
-        reason = "IBV_ACCESS_ON_DEMAND is not supported";
-        goto no_odp;
-    }
-
-    if (!IBV_DEVICE_HAS_ODP(&md->super.dev)) {
-        reason = "device does not support ODP";
```
yosefe, line 1618: maybe keep this check

test/gtest/ucp/test_ucp_mmap.cc
yosefe: 2. oks see the function ucp_apply_uct_config_list

ivankochin: 2. `ucp_test::modify_config` uses `ucp_config_modify_internal` which doesn't propagates environment to UCT layer.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
ivankochin: Why? It is included to `uct_ib_md_check_odp_common`

src/uct/ib/base/ib_md.c
ivankochin: Done

src/uct/ib/base/ib_md.c
ivankochin: Done

src/uct/ib/base/ib_md.c
```diff
@@ -1203,6 +1203,34 @@ static void uct_ib_md_check_dmabuf(uct_ib_md_t *md)
 #endif
 }
 
+int uct_ib_md_check_odp_common(uct_ib_md_t *md, const char **reason_ptr) {
```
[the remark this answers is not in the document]
ivankochin: Done

yosefe (approved):

brminich (approved):

PR #10065, opened 2024-08-15
UCT/CUDA: Detect async allocations as managed (plus env var)

brminich:
## What
Detect async memory allocations as cuda-managed memory to avoid cuda-ipc transport being selected.

## Why ?
Fixes a regression from #9867 

```
Allocate async cuda memory
Allocate async cuda memory
cuda_ipc_md.c:112  UCX  ERROR cuIpcGetMemHandle(&key->ph, (CUdeviceptr)addr) failed: invalid argument
proto_common.inl:373  UCX  ERROR failed to pack remote key: Input/output error
cuda_ipc_md.c:112  UCX  ERROR cuIpcGetMemHandle(&key->ph, (CUdeviceptr)addr) failed: invalid argument
proto_common.inl:373  UCX  ERROR failed to pack remote key: Input/output error
proto_rndv.c:850  Assertion `(address != 0) == (rkey_length > 0)' failed: rts rts->address=0x320000000 rkey_length=0

```

### DIFF
--- a/src/uct/cuda/cuda_copy/cuda_copy_md.h
+++ b/src/uct/cuda/cuda_copy/cuda_copy_md.h
@@ -35,6 +35,7 @@ typedef struct uct_cuda_copy_md {
         int                      dmabuf_supported;
         ucs_ternary_auto_value_t enable_fabric;
         uct_cuda_pref_loc_t      pref_loc;
+        int                      cuda_async_managed;
     } config;
 } uct_cuda_copy_md_t;
 
@@ -48,6 +49,7 @@ typedef struct uct_cuda_copy_md_config {
     ucs_ternary_auto_value_t    enable_dmabuf;
     ucs_ternary_auto_value_t    enable_fabric;
     uct_cuda_pref_loc_t         pref_loc;
+    ucs_memory_type_t           cuda_async_mem_type;
 } uct_cuda_copy_md_config_t;
 
 /**
--- a/src/ucm/cuda/cudamem.c
+++ b/src/ucm/cuda/cudamem.c
@@ -24,8 +24,8 @@
 
 
 /* Create a body of CUDA memory allocation replacement function */
-#define UCM_CUDA_ALLOC_FUNC(_name, _mem_type, _retval, _success, _size, \
-                            _ptr_type, _ref, _args_fmt, ...) \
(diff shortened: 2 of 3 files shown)


src/uct/cuda/cuda_copy/cuda_copy_md.h
```diff
@@ -35,6 +35,7 @@ typedef struct uct_cuda_copy_md {
         int                      dmabuf_supported;
         ucs_ternary_auto_value_t enable_fabric;
         uct_cuda_pref_loc_t      pref_loc;
+        int                      cuda_async_managed;
```
iyastreb, line 38: minor: wouldn't it be more consistent to have also `ucs_memory_type_t` here?

src/uct/cuda/cuda_copy/cuda_copy_md.h
```diff
@@ -48,6 +49,7 @@ typedef struct uct_cuda_copy_md_config {
     ucs_ternary_auto_value_t    enable_dmabuf;
     ucs_ternary_auto_value_t    enable_fabric;
     uct_cuda_pref_loc_t         pref_loc;
+    ucs_memory_type_t           cuda_async_mem_type;
```
iyastreb, line 52: Maybe it's worth adding a simple unit test to cover this change?

src/uct/cuda/cuda_copy/cuda_copy_md.h
brminich: not sure why it is better. We cache/store on md only what is needed. Also we do not need to check memory types again in memory detection flow

src/uct/cuda/cuda_copy/cuda_copy_md.h
brminich: let me try

src/uct/cuda/cuda_copy/cuda_copy_md.h
yosefe: i think it should be ucs_memory_type_t because this is what the parser expects for enums and also we used it for rndv_frag_mem_type

brminich: this is not a config field (so parser and rndv_frag_mem_type are not relevant), i use `ucs_memory_type_t` for config. This is what we save on MD for later evalution. IMO, having  a boolean looks better.

yosefe (approved):

PR #10072, opened 2024-08-20
BUILD/CONFIG: fix intel compiler check

michal-shalev:
image

## What
Fix intel compiler check 

## Why ?
The "-V" flag is not supported by gcc and it causes an error for gcc compiler when running configure-devel/release.

## How ?
Changed to "--version" that is supported by both.


tvegas1 (approved):
checked that it works with `intel/ics-17.0.4`, maybe using pattern `ICC` would work too

config/m4/compiler.m4
```diff
@@ -570,7 +570,7 @@ ADD_COMPILER_FLAGS_IF_SUPPORTED([[-Wno-missing-field-initializers],
 #
 # Intel Compiler specifics
 #
-if ! $CC -V | grep -q Intel.*Compiler; then
+if ! $CC --version | grep -q Intel; then
```
yosefe, line 573: i would convert | to |& to work on compilers that still do not support --version flag

michal-shalev: added &

yosefe (dismissed):

tvegas1 (changes requested):
actually, I think there are existing issues with original script, meaning introduced by #9893:
1. block in the `if` is meant to execute when compiler is intel, so the `!` should be removed
2. this was possible to introduce because `isc -V` outputs on stderr, so grep would fail and we would always enter the block, for gcc and intel compiler
3. we don't need those flags in gcc as they relax code constraints where it's not needed

now `--version` appears to now dump on stdout, so this should make intel compiler checks fail in CI. Finally I think we should replace the `if` line by:

`if $CC --version 2>&1 | grep -q Intel; then`

tvegas1 (approved):

config/m4/compiler.m4
michal-shalev: @yosefe please re-check changed to:
`if $CC --version 2>&1 | grep -q Intel; then`

yosefe (approved):

PR #10074, opened 2024-08-20
UCP/CONFIG: limit transport availability check to ALLOW mode configuration lists only

amastbaum:
## What
Limit the transport availability check to configuration lists set to ALLOW mode only.

## Why ?
Currently ucp_context checks the availability of a user configured transport even when the configuration list is set to NEGATE (i.e. all the transports in this list must NOT be used). This creates a problem when the user excludes an already unavailable transport, for instance `UCX_TLS=^gdr_copy`.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -1838,8 +1838,11 @@ static ucs_status_t ucp_fill_resources(ucp_context_h context,
         }
 
         ucp_get_aliases_set(&avail_tls);
-        ucp_report_unavailable(&config->tls.array, tl_cfg_mask, "", "transport",
-                               &avail_tls);
+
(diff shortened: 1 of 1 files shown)


amastbaum:
@tvegas1 since I've added a fix to `ucp_context.c` I think it might be better to add tests to `test_ucp_context.cc`, as the logic I want to test is related to ucp_context and to the way it uses `ucs_config_allow_list_t`. What do you think?

tvegas1:
> @tvegas1 since I've added a fix to `ucp_context.c` I think it might be better to add tests to `test_ucp_context.cc`, as the logic I want to test is related to ucp_context and to the way it uses `ucs_config_allow_list_t`. What do you think?

ok sure

gleon99:
@amastbaum  as discussed f2f, please fix the PR (and commit) title

yosefe (approved):

PR #10075, opened 2024-08-20
UCP/RNDV/PPLN: Advertise error handling capability for put_mtype/ppln protocols

iyastreb:
## What
Using UCX in this combination: 1.18+proto_v2+CUDA async+error handling, causes memory transfer via host instead of bounce buffers, making transfers prohibitively slow. Unlike with proto_v1, pipeline protocol (rndv_ppln over rnvd_(put/rtr)_mtype) is not being selected, because it doesn't support error handling.
As a result, memory copy is done through host staging buffers which drastically affect performance. The proper solution would be to enable error handling for aforementioned protocols with rkey invalidation.

This PR is the first step towards the real solution: here we just advertise support for error handling without providing invalidation flow. Subsequent PR will provide the invalidation functionality.

Related PRs: https://github.com/openucx/ucx/pull/8666, https://github.com/openucx/ucx/pull/9999

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -354,6 +354,11 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Use two stage pipeline rendezvous protocol for intra-node GPU to GPU transfers",
    ucs_offsetof(ucp_context_config_t, rndv_shm_ppln_enable), UCS_CONFIG_TYPE_BOOL},
 
+  {"RNDV_PIPELINE_ERROR_HANDLING", "n",
+   "Allow using error handling protocol in the rendezvous pipeline protocol\n"
+   "even if invalidation workflow isn't supported",
+   ucs_offsetof(ucp_context_config_t, rndv_errh_ppln_enable), UCS_CONFIG_TYPE_BOOL},
+
   {"FLUSH_WORKER_EPS", "y",
    "Enable flushing the worker by flushing its endpoints. Allows completing\n"
    "the flush operation in a bounded time even if there are new requests on\n"
--- a/src/ucp/rndv/proto_rndv.c
+++ b/src/ucp/rndv/proto_rndv.c
@@ -1002,3 +1002,14 @@ void ucp_proto_rndv_bulk_request_init_lane_idx(
 
     req->send.multi_lane_idx = lane_idx - 1;
 }
+
+void ucp_proto_rndv_stub_abort(ucp_request_t *req, ucs_status_t status)
+{
+    /* FIXME: Proper abort functionality is not implemented yet.
+     * This stub function is used to advertise error-handling capability for
+     * pipeline protocols, but proper implementation is to be done */
+    ucs_diag("aborting rendezvous request %p with status %s. This may lead to "
+             "data corruption, since invalidation workflow isn't implemented",
+             req, ucs_status_string(status));
+    ucp_invoke_uct_completion(&req->send.state.uct_comp, status);
+}
--- a/src/ucp/rndv/rndv_ppln.c
+++ b/src/ucp/rndv/rndv_ppln.c
@@ -52,8 +52,7 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
 {
     ucp_worker_h worker                          = init_params->worker;
     const ucp_proto_select_param_t *select_param = init_params->select_param;
-    ucp_proto_common_init_params_t ack_params    =
-            ucp_proto_common_init_params(init_params);
+    ucp_proto_common_init_params_t ack_params;
     ucp_proto_perf_t *ppln_perf, *ack_perf, *result_perf;
     const ucp_proto_perf_segment_t *frag_seg, *first_seg;
     const ucp_proto_select_elem_t *select_elem;
@@ -67,6 +66,11 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
     void *frag_proto_priv;
     ucs_status_t status;
     uint8_t proto_flags;
+
+    ack_params = ucp_proto_common_init_params(init_params);
+    if (worker->context->config.ext.rndv_errh_ppln_enable) {
+        ack_params.flags |= UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING;
+    }
 
     if ((select_param->dt_class != UCP_DATATYPE_CONTIG) ||
         !ucp_proto_init_check_op(init_params, UCP_PROTO_RNDV_OP_ID_MASK) ||
@@ -208,7 +212,9 @@ ucp_proto_rndv_ppln_frag_complete(ucp_request_t *freq, int send_ack, int abort,
         req->send.rndv.ppln.ack_data_size += freq->send.state.dt_iter.length;
     }
 
-    if (!ucp_proto_rndv_frag_complete(req, freq, title) && !abort) {
+    /* In case of abort we don't destroy super request until all fragments are
+     * completed */
+    if (!ucp_proto_rndv_frag_complete(req, freq, title)) {
         return;
     }
 
@@ -218,7 +224,7 @@ ucp_proto_rndv_ppln_frag_complete(ucp_request_t *freq, int send_ack, int abort,
 
     ucp_datatype_iter_cleanup(&req->send.state.dt_iter, 1, UCP_DT_MASK_ALL);
 
-    if ((req->send.rndv.ppln.ack_data_size > 0) && !abort) {
+    if (!abort && (req->send.rndv.ppln.ack_data_size > 0)) {
         ucp_proto_request_set_stage(req, UCP_PROTO_RNDV_PPLN_STAGE_ACK);
         ucp_request_send(req);
     } else {
@@ -341,7 +347,7 @@ ucp_proto_t ucp_rndv_send_ppln_proto = {
         [UCP_PROTO_RNDV_PPLN_STAGE_SEND] = ucp_proto_rndv_ppln_progress,
         [UCP_PROTO_RNDV_PPLN_STAGE_ACK]  = ucp_proto_rndv_send_ppln_atp_progress,
     },
-    .abort    = ucp_proto_abort_fatal_not_implemented,
+    .abort    = ucp_proto_rndv_stub_abort,
     .reset    = (ucp_request_reset_func_t)ucp_proto_reset_fatal_not_implemented
 };
 
@@ -394,6 +400,6 @@ ucp_proto_t ucp_rndv_recv_ppln_proto = {
         [UCP_PROTO_RNDV_PPLN_STAGE_SEND] = ucp_proto_rndv_ppln_progress,
         [UCP_PROTO_RNDV_PPLN_STAGE_ACK]  = ucp_proto_rndv_recv_ppln_ats_progress,
     },
-    .abort    = ucp_proto_abort_fatal_not_implemented,
+    .abort    = ucp_proto_rndv_stub_abort,
     .reset    = ucp_proto_rndv_ppln_reset
 };
--- a/src/ucp/rndv/rndv_put.c
+++ b/src/ucp/rndv/rndv_put.c
@@ -564,11 +564,15 @@ ucp_proto_rndv_put_mtype_probe(const ucp_proto_init_params_t *init_params)
     ucp_md_map_t mdesc_md_map;
     ucs_status_t status;
     size_t frag_size;
+    unsigned flags;
 
     status = ucp_proto_rndv_mtype_init(init_params, &mdesc_md_map, &frag_size);
     if (status != UCS_OK) {
         return;
     }
+
+    flags = init_params->worker->context->config.ext.rndv_errh_ppln_enable ?
+                UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING : 0;
 
     if (ucp_proto_rndv_init_params_is_ppln_frag(init_params)) {
         comp_cb = ucp_proto_rndv_put_mtype_frag_completion;
@@ -578,7 +582,7 @@ ucp_proto_rndv_put_mtype_probe(const ucp_proto_init_params_t *init_params)
 
     ucp_proto_rndv_put_common_probe(init_params,
                                     UCS_BIT(UCP_RNDV_MODE_PUT_PIPELINE),
-                                    frag_size, UCT_EP_OP_GET_ZCOPY, 0,
+                                    frag_size, UCT_EP_OP_GET_ZCOPY, flags,
                                     mdesc_md_map, comp_cb, 1,
                                     UCP_WORKER_STAT_RNDV_PUT_MTYPE_ZCOPY);
 }
@@ -606,6 +610,6 @@ ucp_proto_t ucp_rndv_put_mtype_proto = {
         [UCP_PROTO_RNDV_PUT_STAGE_ATP]        = ucp_proto_rndv_put_common_atp_progress,
         [UCP_PROTO_RNDV_PUT_STAGE_FENCED_ATP] = ucp_proto_rndv_put_common_fenced_atp_progress,
     },
-    .abort    = ucp_proto_abort_fatal_not_implemented,
+    .abort    = ucp_proto_rndv_stub_abort,
     .reset    = (ucp_request_reset_func_t)ucp_proto_reset_fatal_not_implemented
 };
--- a/test/gtest/common/test_helpers.h
+++ b/test/gtest/common/test_helpers.h
@@ -28,6 +28,7 @@
 #include <vector>
 #include <string>
 #include <algorithm>
+#include <unordered_map>
 #include <sys/socket.h>
 #include <dirent.h>
 #include <stdint.h>
@@ -1052,6 +1053,76 @@ const std::vector<std::vector<ucs_memory_type_t> >& supported_mem_type_pairs();
  * Skip test if address sanitizer is enabled
  */
 void skip_on_address_sanitizer();
+
+/**
+ * Class for mocking C function pointers
+ * The class is used to mock C function pointers, by replacing the original
+ * function pointer with a mock function pointer. The original function pointer
+ * is restored when the object is destroyed.
+ */
+class mock {
+public:
+    using func = void (*)();
+
+    ~mock()
+    {
+        cleanup();
+    }
+
+    void cleanup()
+    {
+        m_mock_map.clear();
+    }
+
+    template<typename Fn> void setup(Fn *orig_ptr, Fn mock)
+    {
+        func *ptr = reinterpret_cast<func*>(orig_ptr);
+        auto it   = m_mock_map.find(ptr);
+        if (it == m_mock_map.end()) {
+            m_mock_map[ptr] = std::unique_ptr<mock_func>(
+                                                new mock_func(ptr, (func)mock));
+        } else {
+            *orig_ptr = mock;
+        }
+    }
+
+    template<typename Fn, typename... Args>
+    ucs_status_t orig_func(Fn *orig_ptr, Args &&...args) const
+    {
+        auto it    = m_mock_map.find((func*)orig_ptr);
+        Fn orig_fn = (it == m_mock_map.end()) ? *orig_ptr :
+                                                (Fn)it->second->m_orig;
+        return orig_fn(std::forward<Args>(args)...);
+    }
+
+private:
+    struct mock_func {
+        mock_func(func *orig_ptr, func mock) :
+            m_orig_ptr(orig_ptr), m_orig(*orig_ptr)
+        {
+            *m_orig_ptr = mock;
+        }
+
+        ~mock_func()
+        {
+            /* Reset value for mocked function pointer.
+             * Due to this side effect we make the class non-copyable, to avoid
+             * implementing copy/move semantics */
+            *m_orig_ptr = m_orig;
+        }
+
+        /* Non-copyable, non-moveable */
+        mock_func(const mock_func&) = delete;
+        mock_func(mock_func&&) = delete;
+        mock_func &operator=(const mock_func&) = delete;
+        mock_func &operator=(mock_func&&) = delete;
+
+        func *m_orig_ptr;
+        func m_orig;
+    };
+
+    std::unordered_map<func*, std::unique_ptr<mock_func>> m_mock_map;
+};
 
 } // ucs
 
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -1992,6 +1992,9 @@ public:
         modify_config("RNDV_THRESH", "128");
         modify_config("RNDV_SCHEME", "put_ppln");
         modify_config("RNDV_PIPELINE_SHM_ENABLE", "n");
+        /* FIXME: Advertise error handling support for RNDV PPLN protocol.
+         * Remove this once invalidation workflow is implemented. */
+        modify_config("RNDV_PIPELINE_ERROR_HANDLING", "y");
         test_ucp_am_nbx::init();
     }
 
@@ -2007,10 +2010,21 @@ public:
             return;
         }
 
-        add_variant_with_value(variants, UCP_FEATURE_AM, 0, "");
+        add_variant_values(variants, test_ucp_am_base::get_test_variants,
+                           UCP_ERR_HANDLING_MODE_NONE);
+        add_variant_values(variants, test_ucp_am_base::get_test_variants,
+                           UCP_ERR_HANDLING_MODE_PEER, "errh");
     }
 
 protected:
+    virtual ucp_ep_params_t get_ep_params() override
+    {
+        ucp_ep_params_t ep_params = test_ucp_am_nbx_rndv::get_ep_params();
+        ep_params.field_mask     |= UCP_EP_PARAM_FIELD_ERR_HANDLING_MODE;
+        ep_params.err_mode        = get_err_mode();
+        return ep_params;
+    }
+
     void test_ppln_send(ucs_memory_type_t mem_type, size_t num_frags,
                         uint64_t stats_cntr_value)
     {
@@ -2048,6 +2062,11 @@ private:
 
         EXPECT_EQ(exp_value, value) << "counter is "
                                     << stats_node->cls->counter_names[cntr];
+    }
+
+    ucp_err_handling_mode_t get_err_mode() const
+    {
+        return static_cast<ucp_err_handling_mode_t>(get_variant_value(1));
     }
 
     ucs_memory_type_t m_mem_type;
--- a/test/gtest/ucp/test_ucp_peer_failure.cc
+++ b/test/gtest/ucp/test_ucp_peer_failure.cc
@@ -13,6 +13,7 @@ extern "C" {
 #include <ucp/core/ucp_ep.inl>    /* for testing EP RNDV configuration */
 #include <ucp/core/ucp_request.h> /* for debug */
 #include <ucp/core/ucp_worker.h>  /* for testing memory consumption */
+#include <ucp/rndv/proto_rndv.h>
 }
 
 #include <unordered_map>
@@ -647,7 +648,8 @@ public:
 protected:
     enum class rndv_mode {
         rndv_get,
-        rndv_put
+        rndv_put,
+        put_ppln
     };
 
     void init()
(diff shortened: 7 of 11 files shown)


test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1981,7 +1981,9 @@ UCP_INSTANTIATE_TEST_CASE_GPU_AWARE(test_ucp_am_nbx_rndv_memtype_disable_zcopy);
 #ifdef ENABLE_STATS
 class test_ucp_am_nbx_rndv_ppln : public test_ucp_am_nbx_rndv {
 public:
-    test_ucp_am_nbx_rndv_ppln() : m_mem_type(UCS_MEMORY_TYPE_HOST) {}
+    test_ucp_am_nbx_rndv_ppln() : m_mem_type(UCS_MEMORY_TYPE_HOST) {
+        modify_config("PROTO_ENABLE", "y");
```
brminich, line 1985: it is not needed, the test is skipped for protov1  (see `init` on line 1990)

iyastreb: The problem is the base class `test_ucp_am_base` constructor, that treats any variant value as proto v1. To ignore that behaviour of the base class I revert setting back to proto v2.
```C
    test_ucp_am_base()
    {
        if (get_variant_value()) {
            modify_config("PROTO_ENABLE", "n");
        }
    }
```
This seems to be the least intrusive way to fix it, I was afraid of changing that base class due to many subclasses reusing it

evgeny-leksikov: so, maybe just skip wrong variants and add missed if any?

iyastreb: It's not about skipping wrong variants.
Look, as a part of this commit I've added a new variant (error handling).
But, the base class misinterprets this variant - it treats **any non-zero variant of level 0 as proto_v1** which is wrong IMO (see code above). It makes all my error-handling tests = proto_v1 tests, which are then skipped.
So I see the following options to fix it:
1. The easiest is what's implemented here: just always force proto_v2, and ignore base class misinterpretation of new variant
2. use level 0 variant for proto_v1 (to conform the base class interpretation), and add level 1 variant for error handling. Then obviously we skip all proto_v1 tests.
3. Change impl of the base class ctor, and all the subclasses using it
Any other ideas?
I thought option 1 is the best choice here

evgeny-leksikov: The code above looks weird to me, from first look it seems like something temporary and hacky to pass CI on transition from proto_v1 to proto_v2.
If we need some derived tests working only with proto_v1 and others with proto_v2, I would suggest to change inheritance correspondingly.

brminich: this class defines its own variants, so should modify base class probably to not set it when it is not needed

rakhmets: You can introduce enum.
```
enum {
    VARIANT_DEFAULT = 0,
    VARIANT_PROTO_V1,
    VARIANT_ERR_HANDLING
};

class test_ucp_am_base : public ucp_test {
...
    test_ucp_am_base()
    {
        if (get_variant_value() == VARIANT_PROTO_V1) {
            modify_config("PROTO_ENABLE", "n");
        }
    }

    static void get_test_variants(variant_vec_t &variants)
    {
        add_variant_with_value(variants, UCP_FEATURE_AM, VARIANT_DEFAULT, "");
        if (!RUNNING_ON_VALGRIND) {
            add_variant_with_value(variants, UCP_FEATURE_AM, VARIANT_PROTO_V1, "proto_v1");
        }
    }
...
};

...

class test_ucp_am_nbx_rndv_ppln : public test_ucp_am_nbx_rndv {
...
    static void get_test_variants(variant_vec_t &variants)
    {
        if (!mem_buffer::is_gpu_supported()) {
            return;
        }

        add_variant_with_value(variants, UCP_FEATURE_AM, VARIANT_DEFAULT , "");
        add_variant_with_value(variants, UCP_FEATURE_AM, VARIANT_ERR_HANDLING, "error-handling");
    }

    get_err_mode() {
        return (get_variant_value() == VARIANT_ERR_HANDLING) ? UCP_ERR_HANDLING_MODE_PEER : UCP_ERR_HANDLING_MODE_NONE;
    }
```

iyastreb: Thanks for the code, I agree that this is the right way of refactoring

iyastreb: After reflecting a bit more on this, I realised that enum approach might not be the best candidate..
It allows to choose just a single option (default, proto_v1, errh) for level 0, but there are tests in this file that use both combinations: proto_v1 AND errh: e.g. `test_ucp_am_nbx_dts` So this enum does not really map well with this. TBH I would stay with initial approach that is one-liner: ignore base class protocol setting and treat it as error-handling mode

rakhmets: I would recommend to use enum anyway. The code in the base class (which was partially written by) is difficult to extend. A good practice is to add specific value using `add_variant_with_value`, and compare with it when calling `get_variant_value`.

And if it is required to multiply the test cases it should be done like it already done:
```
static void get_test_variants(variant_vec_t &variants)
{
    add_variant_values(variants, base_class_name::get_test_variants, VARIANT_DEFAULT);
    add_variant_values(variants, base_class_name::get_test_variants, VARIANT_ERR_HANDLING, "error-handling");
}

get_err_mode() {
    return (get_variant_value(<index>) == VARIANT_ERR_HANDLING) ? UCP_ERR_HANDLING_MODE_PEER : UCP_ERR_HANDLING_MODE_NONE;
}
```

rakhmets: side note. I think the

src/ucp/core/ucp_context.c
```diff
@@ -354,6 +354,11 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Use two stage pipeline rendezvous protocol for intra-node GPU to GPU transfers",
    ucs_offsetof(ucp_context_config_t, rndv_shm_ppln_enable), UCS_CONFIG_TYPE_BOOL},
 
+  {"RNDV_PIPELINE_ERRH_ENABLE", "n",
```
ivankochin, line 357: An idea came to my mind that maybe it worth to introduce more common variable that allows using all the protocols which don't have implemented abort even if error handling is enabled. That control will allow customers to use all the protocols if they are OK that any abort call will cause hard error. Also it can be useful for testing and debugging from my point of view.

src/ucp/rndv/rndv_ppln.c
```diff
@@ -305,7 +314,7 @@ ucp_proto_t ucp_rndv_send_ppln_proto = {
         [UCP_PROTO_RNDV_PPLN_STAGE_SEND] = ucp_proto_rndv_ppln_progress,
         [UCP_PROTO_RNDV_PPLN_STAGE_ACK]  = ucp_proto_rndv_send_ppln_atp_progress,
     },
-    .abort    = ucp_proto_abort_fatal_not_implemented,
+    .abort    = ucp_proto_rndv_ppln_abort,
```
ivankochin, line 317: Maybe it would be better to leave that aborts pointers set to `ucp_proto_abort_fatal_not_implemented` so when it is called, user will  get hard error instead of half-ready abortion implementation without memory invalidation?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -2050,6 +2065,34 @@ class test_ucp_am_nbx_rndv_ppln : public test_ucp_am_nbx_rndv {
...
+                    ucp_worker_iface_get_attr(sender().worker(),
+                                              ucp_ep_get_rsc_index(ep, lane));
+                if (iface_attr->cap.flags & UCT_IFACE_FLAG_PUT_ZCOPY) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
```
brminich, line 2089: why is it needed?
there is `ucp_test_base::entity::is_rndv_put_ppln_supported()`

src/ucp/core/ucp_context.c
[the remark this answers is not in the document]
iyastreb: done

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1981,7 +1981,8 @@ UCP_INSTANTIATE_TEST_CASE_GPU_AWARE(test_ucp_am_nbx_rndv_memtype_disable_zcopy);
 #ifdef ENABLE_STATS
 class test_ucp_am_nbx_rndv_ppln : public test_ucp_am_nbx_rndv {
 public:
-    test_ucp_am_nbx_rndv_ppln() : m_mem_type(UCS_MEMORY_TYPE_HOST) {}
+    test_ucp_am_nbx_rndv_ppln() : m_mem_type(UCS_MEMORY_TYPE_HOST) {
+    }
```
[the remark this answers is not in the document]
iyastreb: ok

test/gtest/ucp/test_ucp_am.cc
iyastreb: Thanks for spotting this, this is not needed, and not actually used, it's just a leftover after rebasing on master

src/ucp/rndv/rndv_ppln.c
iyastreb: For now I would keep it for 2 reasons:
1. As discussed on the meeting, we don't want to crash the application in case of remote peer failure.
2. Current protocol selection logic explicitly requires that abort function pointer is set to a different function:
```C
void ucp_proto_select_add_proto
...
    if (init_params->ep_config_key->err_mode != UCP_ERR_HANDLING_MODE_NONE) {
        ucs_assertv(ucp_protocols[proto_id]->abort !=
                            ucp_proto_abort_fatal_not_implemented,
                    "error handling is enabled, but %s doesn't support abort()",
                    proto_name);
    }
```

src/ucp/core/ucp_context.c
iyastreb: I think this idea might be useful, but maybe it's out of scope of this PR. Because in this particular case we plan to implement proper support for error handling and therefore this env var will be removed in the subsequent commit. So for these protocols it's not gonna be used anymore, and for the other ones - it can be addressed separately with clear scope & testing

brminich (approved):

brminich (dismissed):

src/ucp/rndv/rndv_ppln.c
evgeny-leksikov: maybe add warning message to `ucp_proto_rndv_ppln_abort` and `ucp_proto_rndv_put_mtype_abort`? BTW, they are 100% identical, so can be replaced with one common function, for example `ucp_proto_ppln_abort `

iyastreb: Ok for warning message
Ok for common function

src/ucp/rndv/proto_rndv.inl
```diff
@@ -386,4 +386,15 @@ ucp_proto_rndv_recv_complete(ucp_request_t *req)
     return ucp_proto_rndv_recv_complete_status(req, rreq->status);
 }
 
+static UCS_F_ALWAYS_INLINE void
+ucp_proto_rndv_ppln_abort(ucp_request_t *req, ucs_status_t status)
```
evgeny-leksikov, line 390: minor: this function is supposed to be called by pointer on non-fast path, no need inline and static, i.e. it can be declared in some common `.h` and defined in `.c`

iyastreb: Done

src/ucp/rndv/proto_rndv.c
```diff
@@ -1071,3 +1071,13 @@ void ucp_proto_rndv_bulk_request_init_lane_idx(
 
     req->send.multi_lane_idx = lane_idx - 1;
 }
+
+void ucp_proto_rndv_ppln_abort(ucp_request_t *req, ucs_status_t status)
+{
+    /* FIXME: Proper abort functionality is not implemented yet.
+     * This stub function is used to advertise error-handling capability for
+     * pipeline protocols, but proper implementation is to be done */
+    ucs_warn("aborting ppln request %p with status %s", req,
```
ivankochin, line 1080: WDYT about making warning more informative and describe what it can lead to? 
```suggestion
    ucs_warn("rndv ppln request %p was aborted with status %s. "
                    "This may lead to memory leaks since rndv ppln abortion "
                    "mocking was enabled", req,
```

src/ucp/rndv/rndv_put.c
```diff
@@ -605,6 +609,6 @@ ucp_proto_t ucp_rndv_put_mtype_proto = {
         [UCP_PROTO_RNDV_PUT_STAGE_ATP]        = ucp_proto_rndv_put_common_atp_progress,
         [UCP_PROTO_RNDV_PUT_STAGE_FENCED_ATP] = ucp_proto_rndv_put_common_fenced_atp_progress,
     },
-    .abort    = ucp_proto_abort_fatal_not_implemented,
+    .abort    = ucp_proto_rndv_ppln_abort,
```
ivankochin, line 612: Not sure that current name of the function is a good one since `rndv/put/mtype` is not always a part of pipeline. Maybe rename to `ucp_proto_stub_abort`?

src/ucp/core/ucp_context.c
```diff
@@ -354,6 +354,11 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Use two stage pipeline rendezvous protocol for intra-node GPU to GPU transfers",
    ucs_offsetof(ucp_context_config_t, rndv_shm_ppln_enable), UCS_CONFIG_TYPE_BOOL},
 
+  {"RNDV_PIPELINE_ERROR_HANDLING", "n",
```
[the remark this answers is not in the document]
iyastreb: Personally I would not add "mock" prefix, but let's see what others think

src/ucp/rndv/proto_rndv.c
iyastreb: My understanding is that it may lead to memory corruption rather than memory leak.
Because when we detect that EP is in failed state, we should invalidate the memh we exposed to prevent stale writes to it. Since we don't do invalidation yet, these writes may appear and damage chunk elements already used for different purpose.
So in worst case the same element is reused for some other operation and we can get data corrupted.
I can update the warning message with these details if you agree

src/ucp/rndv/rndv_put.c
iyastreb: ok, ucp_proto_rndv_stub_abort then

src/ucp/rndv/proto_rndv.c
ivankochin: Yeah, AFAIK that is true too. I thought about memory leaks because we are using mempools and I am not sure that we are returning them back in case of this abortion stub calling. 

Mentioning memory corruption is OK from my point of view.

brminich: pls use `ucs_diag` instead of `ucs_warn`

evgeny-leksikov: > pls use `ucs_diag` instead of `ucs_warn`

why? AFAIU this callback is not expected to be called, IMO warning is ok here.

brminich: it can be called during error handling flow 
(even without invalidation better not to print something to warning)

iyastreb: done

evgeny-leksikov: It can be, but unexpected and may lead to undefined behaviour.
Is invalidation the only missed thing on this flow? if so then just need to move this warning under "connected-to-iface" condition.

brminich: i had am impression that invalidation is the only missing piece (ilia pls confirm).
I'd still leave it as `diag`, because even with `connect-to-iface` invalidation may not be really required. 
The point is that this how things worked in protov1 and users were satisfied with this behavior. Also we already protected  this (not fully complete) code with env var to be able to disable such behavior.

iyastreb: Yes, I confirm that only invalidation is missing
IMO it's not really a big deal, as subsequent commit will remove this code anyway..

ivankochin: +1 for leaving as warning (or at least debug message)

brminich: > +1 for leaving as warning (or at least debug message)

we are not going to remove it. We are duscussing the log level to be used

evgeny-leksikov: if you are sure that `connect-to-ep` works as expected, then this message will only confuse with any level.

I see 2 potential issues here:
1) completion if single fragment might be not enough to complete full ppl protocol. It may be relevant for any type of transport. (@iyastreb confirmed that this is not the case).
2) invalidation flow is valid only for `connect-to-iface` transport.
So, 
1) that's the reason to have it for `connect-to-ep`?
2) invalidation flow may lead to hardly reproducible data corruption bug, IMO warning is preferable because on re-run (with log level diag) error flow may be not invoked at all or invoked but without data corruption.

ivankochin: > > +1 for leaving as warning (or at least debug message)
> 
> we are not going to remove it. We are duscussing the log level to be used

I understand, but I think that we need to print a warning message instead of diag to explicitly show user that something went wrong and call that was not supposed to be called was called. FMPOV if we didn't print warning in that case in protov1 it doesn't mean that we should keep this behavior now.

iyastreb: let's discuss it on meeting, this discussion becomes too long

src/ucp/core/ucp_context.c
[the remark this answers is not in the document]
yosefe: maybe RNDV_PPLN_IGNORE_FAILURES?

src/ucp/rndv/proto_rndv.c
```diff
@@ -1071,3 +1071,14 @@ void ucp_proto_rndv_bulk_request_init_lane_idx(
 
     req->send.multi_lane_idx = lane_idx - 1;
 }
+
+void ucp_proto_rndv_stub_abort(ucp_request_t *req, ucs_status_t status)
+{
+    /* FIXME: Proper abort functionality is not implemented yet.
+     * This stub function is used to advertise error-handling capability for
+     * pipeline protocols, but proper implementation is to be done */
+    ucs_diag("aborting ppln request %p with status %s. This may lead to data "
```
yosefe, line 1080: ppln -> rendezvous 
since this function is rndv common code

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -782,10 +753,17 @@ class test_ucp_peer_failure_rndv_abort : public test_ucp_peer_failure {
         switch (mode) {
         case rndv_mode::rndv_get:
             m_proto_name    = "rndv/get/zcopy";
+            m_proto_stage   = UCP_PROTO_STAGE_START;
```
yosefe, line 756: maybe rename m_proto_stage to m_proto_xfer_stage, document that it is the protocol stage during which data transfer happens, and here use UCP_PROTO_STAGE_START instead of UCP_PROTO_STAGE_START

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -666,35 +668,17 @@ class test_ucp_peer_failure_rndv_abort : public test_ucp_peer_failure {
         set_am_handler(receiver());
     }
 
-    template<typename Func>
-    void for_each_protos_progress(Func func)
+    static void mock_protos_progress(ucs::mock &mock)
```
yosefe, line 671: setup_progress_mock

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -703,26 +687,16 @@ class test_ucp_peer_failure_rndv_abort : public test_ucp_peer_failure {
         return UCS_ERR_CONNECTION_RESET;
     }
 
-    void replace_rndv_ops(ucp_ep_h ep)
+    static void mock_rndv_ops(ucp_ep_h ep, ucs::mock &mock)
```
yosefe, line 690: pls check code style in clang-format

test/gtest/common/test_helpers.h
```diff
@@ -1053,6 +1054,77 @@ const std::vector<std::vector<ucs_memory_type_t> >& supported_mem_type_pairs();
...
 
+/**
+ * Class for mocking C function pointers
+ * The class is used to mock C function pointers, by replacing the original
+ * function pointer with a mock function pointer. The original function pointer
+ * is restored when the object is destroyed.
+ */
+class mock {
+public:
+    using func = void (*)();
```
yosefe, line 1065: maybe it can be a template parameter for IMO this typedef is redundant, could just use "void*"

test/gtest/common/test_helpers.h
```diff
@@ -1053,6 +1054,77 @@ const std::vector<std::vector<ucs_memory_type_t> >& supported_mem_type_pairs();
...
+        if (it == m_mock_map.end()) {
+            mock_func *new_mock = new mock_func(ptr, (func)mock);
+            m_mock_map[ptr]     = std::unique_ptr<mock_func>(new_mock);
+        } else {
+            it->second->m_mock = (func)mock;
+        }
+    }
+
+    template <typename Fn, typename... Args>
+    ucs_status_t actual_call(Fn *orig_ptr, Args&&... args) const
```
yosefe, line 1090: maybe rename "actual_call" to "orig_func" ?

test/gtest/common/test_helpers.h
```diff
@@ -1053,6 +1054,77 @@ const std::vector<std::vector<ucs_memory_type_t> >& supported_mem_type_pairs();
...
+    {
+        m_mock_map.clear();
+    }
+
+    template <typename Fn> void setup(Fn *orig_ptr, Fn mock)
+    {
+        func *ptr    = reinterpret_cast<func*>(orig_ptr);
+        auto it      = m_mock_map.find(ptr);
+        if (it == m_mock_map.end()) {
+            mock_func *new_mock = new mock_func(ptr, (func)mock);
```
yosefe, line 1082: do we really need the local var new_mock?

test/gtest/common/test_helpers.h
```diff
@@ -1053,6 +1054,77 @@ const std::vector<std::vector<ucs_memory_type_t> >& supported_mem_type_pairs();
...
+
+    void cleanup()
+    {
+        m_mock_map.clear();
+    }
+
+    template <typename Fn> void setup(Fn *orig_ptr, Fn mock)
+    {
+        func *ptr    = reinterpret_cast<func*>(orig_ptr);
+        auto it      = m_mock_map.find(ptr);
```
yosefe, line 1080: too many spaces before =

test/gtest/common/test_helpers.h
```diff
@@ -1053,6 +1054,77 @@ const std::vector<std::vector<ucs_memory_type_t> >& supported_mem_type_pairs();
...
+        if (it == m_mock_map.end()) {
+            mock_func *new_mock = new mock_func(ptr, (func)mock);
+            m_mock_map[ptr]     = std::unique_ptr<mock_func>(new_mock);
+        } else {
+            it->second->m_mock = (func)mock;
```
yosefe, line 1085: don't we want to replace *orig_ptr here?
do we have to save m_mock in mock_func as a member var?

test/gtest/common/test_helpers.h
iyastreb: ok

test/gtest/common/test_helpers.h
iyastreb: In this class we can keep mocks of different function types, so it cannot be a template argument.

It's not recommended to use `void *` for function pointers, and it gives an appropriate warning when built with pedantic: `warning: initializing 'void *' with an expression of type 'void (*)()' converts between void pointer and function pointer`. So the safest option is to use dedicated function pointer, as suggested here: https://stackoverflow.com/questions/36645660/why-cant-i-cast-a-function-pointer-to-void

src/ucp/rndv/proto_rndv.c
iyastreb: done

src/ucp/core/ucp_context.c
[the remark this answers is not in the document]
iyastreb: I don't know, sounds a bit broad.. which failures?
maybe RNDV_PIPELINE_UNSAFE_ERROR_HANDLING?

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -752,19 +754,19 @@ class test_ucp_peer_failure_rndv_abort : public test_ucp_peer_failure {
 
         switch (mode) {
         case rndv_mode::rndv_get:
-            m_proto_name    = "rndv/get/zcopy";
-            m_proto_stage   = UCP_PROTO_STAGE_START;
-            m_peer_to_close = &self->sender();
+            m_proto_name       = "rndv/get/zcopy";
+            m_proto_xfer_stage = UCP_PROTO_STAGE_START;
```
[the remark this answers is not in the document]
iyastreb: yes, thanks, I moved UCP_PROTO_RNDV_GET_STAGE_FETCH to header, but forgot to use it here

src/ucp/rndv/rndv_ppln.c
```diff
@@ -52,11 +52,10 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
 {
     ucp_worker_h worker                          = init_params->worker;
     const ucp_proto_select_param_t *select_param = init_params->select_param;
-    ucp_proto_common_init_params_t err_params    = {
-        .super = *init_params,
-        .flags = worker->context->config.ext.rndv_errh_ppln_enable ?
-                    UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING : 0
-    };
+    ucp_proto_common_init_params_t ack_params    =
+            ucp_proto_common_init_params(init_params);
```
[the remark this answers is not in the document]
iyastreb: ok

test/gtest/ucp/test_ucp_peer_failure.cc
```diff
@@ -748,24 +722,23 @@ class test_ucp_peer_failure_rndv_abort : public test_ucp_peer_failure {
         auto *req   = ucs_container_of(uct_req, ucp_request_t, send.uct);
         auto stage  = req->send.proto_stage;
         auto *proto = req->send.proto_config->proto;
+        ucs::mock mock;
+
+        ucs_debug("progress_wrapper %s stage %d", proto->name, (int)stage);
```
brminich, line 727: is it really needed? Typically we do not add debug prints to tests

iyastreb: not really, removing

brminich (approved):

src/ucp/core/ucp_context.c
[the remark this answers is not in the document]
yosefe: maybe RNDV_PPLN_ERROR_HANDLING?

The short form PPLN also appears in UCX_RNDV_SCHEME values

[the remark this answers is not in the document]
iyastreb: ok, sure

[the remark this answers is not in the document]
brminich:
```
[mikhailb@rock20 ucx]$ ucx_info -c | grep PIPELINE
UCX_RNDV_PIPELINE_SEND_THRESH=inf
UCX_RNDV_PIPELINE_SHM_ENABLE=y
[mikhailb@rock20 ucx]$ ucx_info -c | grep PPLN
```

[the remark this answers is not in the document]
iyastreb: wait, it's dissonant with previous vars:
```C
  {"RNDV_PIPELINE_SEND_THRESH", "inf",
   "RNDV size threshold to enable sender side pipeline for mem type",
   ucs_offsetof(ucp_context_config_t, rndv_pipeline_send_thresh), UCS_CONFIG_TYPE_MEMUNITS},

  {"RNDV_PIPELINE_SHM_ENABLE", "y",
   "Use two stage pipeline rendezvous protocol for intra-node GPU to GPU transfers",
   ucs_offsetof(ucp_context_config_t, rndv_shm_ppln_enable), UCS_CONFIG_TYPE_BOOL},

  {"RNDV_PPLN_ERROR_HANDLING", "n",
```

PR #10076, opened 2024-08-20
BUILD: print test name on failure in builds.sh

michal-shalev:
## What
Added logic to `builds.sh` to print the test name if a test fails.

## Why ?
To quickly identify which test failed, improving debugging efficiency.

## How ?
Inserted a conditional check after each test in builds.sh to print the test name and exit on failure.

### DIFF
--- a/buildlib/tools/builds.sh
+++ b/buildlib/tools/builds.sh
@@ -461,5 +461,5 @@ do
 	[ -d "${ucx_build_dir}" ] && rm -rf ${ucx_build_dir}/*
 
 	# run the test
-	${test_name}
+	$test_name || { azure_log_error "Test failed: $test_name"; exit 1; }
 done


buildlib/tools/builds.sh
```diff
@@ -461,5 +461,5 @@ do
 	[ -d "${ucx_build_dir}" ] && rm -rf ${ucx_build_dir}/*
 
 	# run the test
-	${test_name}
+	${test_name} || { echo "Test failed: ${test_name}"; exit 1; }
```
tvegas1, line 464: if we exit, we cannot get a list of failing tests at once, which could be painful?

tvegas1 (commented):
maybe don't

buildlib/tools/builds.sh
michal-shalev: I can remove the "exit 1;" part

yosefe: the test would already exit after first failure, to not consume CI resources
https://github.com/openucx/ucx/blob/master/buildlib/tools/builds.sh#L1

tvegas1: I see

michal-shalev: removed the "exit 1;" part @tvegas1 @yosefe

yosefe: pls keep the exit

michal-shalev: done @yosefe

tvegas1 (dismissed):

yosefe (dismissed):

buildlib/tools/builds.sh
gleon99, line 464: Maybe better use `azure_log` helpers here instead, @yosefe  wdyt?

michal-shalev: Changed `echo` to `azure_log_error`, please check @gleon99 @yosefe

gleon99 (approved):

gleon99:
Please squash

yosefe (approved):

PR #10079, opened 2024-08-21
UCP/MMAP: Avoid noblocking registration for MDs which do not support it

ivankochin:
## What
Currently `ucp_mem_map` registers MDs according to `reg_md_map` only. But if CUDA buffer was provided with nonblocking registration flag, `ucp_mem_map` will call UCT registration with this flag for all MDs in `reg_md_map`.

This patch reworks this logic by the following way:
- If `mem_type` was specified in `UCX_REG_NONBLOCK_MEM_TYPES`, then register only MDs that support nonblocking registration
- Else do not pass nonblocking flag for MDs that don't support it (since this flag documented as a hint)

## Why ?
To avoid weird  CUDA memory ODP registration case.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -1605,6 +1605,11 @@ ucp_add_component_resources(ucp_context_h context, ucp_rsc_index_t cmpt_index,
                         context->reg_md_map[mem_type] |= UCS_BIT(md_index);
                     }
 
+                    if (md_attr->reg_nonblock_mem_types & UCS_BIT(mem_type)) {
+                        context->reg_nonblock_md_map[mem_type] |=
+                                UCS_BIT(md_index);
+                    }
+
                     if (md_attr->cache_mem_types & UCS_BIT(mem_type)) {
                         context->cache_md_map[mem_type] |= UCS_BIT(md_index);
                     }
@@ -1732,6 +1737,7 @@ static ucs_status_t ucp_fill_resources(ucp_context_h context,
     ucs_memory_type_for_each(mem_type) {
         context->reg_md_map[mem_type]           = 0;
         context->reg_block_md_map[mem_type]     = 0;
+        context->reg_nonblock_md_map[mem_type]  = 0;
         context->cache_md_map[mem_type]         = 0;
         context->gva_md_map[mem_type]           = 0;
         context->dmabuf_mds[mem_type]           = UCP_NULL_RESOURCE;
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -320,6 +320,10 @@ typedef struct ucp_context {
        ucp_mem_map() will register memory for all those domains. */
     ucp_md_map_t                  reg_md_map[UCS_MEMORY_TYPE_LAST];
 
+    /* Map of MDs that provide nonblocking registration for given memory type,
+       ucp_mem_map() will register memory for all those domains. */
+    ucp_md_map_t                  reg_nonblock_md_map[UCS_MEMORY_TYPE_LAST];
+
     /* Map of MDs that provide blocking registration for given memory type.
      * This map is initialized if non-blocking registration is requested for
      * the memory type (thus reg_md_map contains only MDs supporting
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -534,17 +534,20 @@ ucp_memh_register_internal(ucp_context_h context, ucp_mem_h memh,
     }
 
     reg_md_map = ~memh->md_map & md_map;
+    if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
+        !(uct_flags & UCT_MD_MEM_FLAG_LOCK)) {
+        /* Do not register MDs that doesn't support nonblock registration if
+         * this memtype exists in `config.ext.reg_nb_mem_types`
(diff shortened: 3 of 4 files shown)


src/ucp/core/ucp_context.h
```diff
@@ -320,6 +320,10 @@ typedef struct ucp_context {
        ucp_mem_map() will register memory for all those domains. */
     ucp_md_map_t                  reg_md_map[UCS_MEMORY_TYPE_LAST];
 
+    /* Map of MDs that provide nonblocking registration for given memory type,
+       ucp_mem_map() will register memory for all those domains. */
+    ucp_md_map_t                  reg_nonblock_md_map[UCS_MEMORY_TYPE_LAST];
```
rakhmets, line 325: The description for the next field `reg_block_md_map` is confusing.
It is said that `reg_md_map` contains only MDs supporting non-blocking registration. Which means that the new map is redundant.

src/ucp/core/ucp_mm.c
```diff
@@ -580,7 +583,13 @@ ucp_memh_register_internal(ucp_context_h context, ucp_mem_h memh,
                     context->reg_md_map[mem_type],
                     context->reg_block_md_map[mem_type]);
 
+        reg_params.flags      = uct_flags | memh->uct_flags;
         reg_params.field_mask = UCT_MD_MEM_REG_FIELD_FLAGS;
+        if (!(context->reg_nonblock_md_map[mem_type] & md_index)) {
+            /* Ignore nonblcok flag if MD doesn't support it */
+            reg_params.flags &= !UCT_MD_MEM_FLAG_NONBLOCK;
```
tvegas1, line 590: !1 is 0, so we remove everything unless I misread?

src/ucp/core/ucp_mm.c
```diff
@@ -580,7 +583,13 @@ ucp_memh_register_internal(ucp_context_h context, ucp_mem_h memh,
                     context->reg_md_map[mem_type],
                     context->reg_block_md_map[mem_type]);
 
+        reg_params.flags      = uct_flags | memh->uct_flags;
         reg_params.field_mask = UCT_MD_MEM_REG_FIELD_FLAGS;
+        if (!(context->reg_nonblock_md_map[mem_type] & md_index)) {
+            /* Ignore nonblcok flag if MD doesn't support it */
```
tvegas1, line 589: typo

src/ucp/core/ucp_context.h
tvegas1: maybe add something like: `.. when nonblocking is requested, registration will only be done for MD part of the Map`

src/ucp/core/ucp_context.c
```diff
@@ -1605,6 +1605,11 @@ ucp_add_component_resources(ucp_context_h context, ucp_rsc_index_t cmpt_index,
                         context->reg_md_map[mem_type] |= UCS_BIT(md_index);
                     }
 
+                    if (md_attr->reg_nonblock_mem_types & UCS_BIT(mem_type)) {
```
tvegas1, line 1608: can you please confirm if my understanding is correct:
- `reg_mem_types` is set if blocking, non-blocking registration or both, are supported
  - `reg_md_map` mirrors that
- `reg_nonblock_mem_types` is set if non-blocking is supported
  - `reg_nonblock_md_map` mirrors that
- `reg_block_md_map` is only applicable if mem_type is part of `reg_nb_mem_types`.
  - it is then set if registration is supported but not the non-blocking mode

src/ucp/core/ucp_mm.c
```diff
@@ -534,17 +534,20 @@ ucp_memh_register_internal(ucp_context_h context, ucp_mem_h memh,
     }
 
     reg_md_map = ~memh->md_map & md_map;
-    if (reg_md_map == 0) {
-        return UCS_OK;
-    }
-
     if ((context->config.ext.reg_nb_mem_types & UCS_BIT(mem_type)) &&
```
tvegas1, line 537: So `mem_type` found in `reg_nb_mem_types` forces non-blocking registration, and only MD that support them.

Without that, it is still possible to manually pass `UCT_MD_MEM_FLAG_NONBLOCK`, which tries non-block on MDs that support it, else tries with blocking reg, is that correct?

src/ucp/core/ucp_mm.c
ivankochin: My bad, thanks for noticing!

src/ucp/core/ucp_context.c
ivankochin: My own understanding is that nonblock registration is a feature of usual (blocking) registration, so `reg_nonblock_mem_types` is always a subset of `reg_mem_types`. And `reg_mem_types` contains mem types for which usual (blocking) registration can be performed by this MD.

Understanding of other fields looks correct FMPOV.

src/ucp/core/ucp_mm.c
ivankochin: Yes, that's correct. According to this logic `reg_nb_mem_types` forces nonblock registration only and `UCT_MD_MEM_FLAG_NONBLOCK` considered us hint only.

brminich: AFAIU the issue is that we pass `UCT_MD_MEM_FLAG_NONBLOCK` to UCT registration for cuda memory. How is that possible given this check for `context->config.ext.reg_nb_mem_types`?

PR #10082, opened 2024-08-22
UCT/CUDA_IPC: Cache for mempool import operation

Akshay-Venkatesh:
## What
Follow up to https://github.com/openucx/ucx/pull/9982. This PR caches the operation that imports remotely exported handle for a custom CUDA memory pool as the mapping operation via `cuMemPoolImportFromShareableHandle` is expensive.

### DIFF
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
@@ -48,6 +48,40 @@ typedef struct uct_cuda_ipc_remote_cache {
 
 uct_cuda_ipc_remote_cache_t uct_cuda_ipc_remote_cache;
 
+#if HAVE_CUDA_FABRIC
+static UCS_F_ALWAYS_INLINE int
+uct_cuda_ipc_rem_mpool_cache_hash_equal(CUmemFabricHandle key1,
+                                        CUmemFabricHandle key2)
+{
+    return !memcmp(&key1, &key2, sizeof(key1));
+}
+
+/* djb2 hash function */
+static UCS_F_ALWAYS_INLINE khint_t
+uct_cuda_ipc_rem_mpool_cache_hash_func(CUmemFabricHandle key)
+{
+    khint_t hash = 5381;
+    int i;
+
+    for (i = 0; i < sizeof(key.data); i++) {
+        hash = ((hash << 5) + hash) + key.data[i];
+    }
+
+    return hash;
+}
+
+KHASH_INIT(cuda_ipc_rem_mpool_cache, CUmemFabricHandle, CUmemoryPool, 1,
+           uct_cuda_ipc_rem_mpool_cache_hash_func,
+           uct_cuda_ipc_rem_mpool_cache_hash_equal)
+
+typedef struct {
+    khash_t(cuda_ipc_rem_mpool_cache) hash;
+    pthread_rwlock_t                  lock;
+} uct_cuda_ipc_rem_mpool_cache_t;
+
+static uct_cuda_ipc_rem_mpool_cache_t uct_cuda_ipc_rem_mpool_cache;
+#endif
+
 static ucs_pgt_dir_t *uct_cuda_ipc_cache_pgt_dir_alloc(const ucs_pgtable_t *pgtable)
 {
     void *ptr;
@@ -200,41 +234,100 @@ out:
     return status;
 }
 
-static ucs_status_t
-uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
-                                    CUdeviceptr *mapped_addr)
+static ucs_status_t cuda_ipc_rem_mpool_cache_create(uct_cuda_ipc_rkey_t *key,
+                                                    CUmemoryPool *mpool,
+                                                    CUdeviceptr *mapped_addr)
 {
     CUmemAccessDesc access_desc = {};
     CUdeviceptr dptr;
     ucs_status_t status;
 
     status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportFromShareableHandle(
-                &key->ph.pool, (void *)&key->ph.handle.fabric_handle,
+                mpool, (void *)&key->ph.handle.fabric_handle,
                 CU_MEM_HANDLE_TYPE_FABRIC, 0));
     if (status != UCS_OK) {
-        return status;
+        goto err;
     }
 
-    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(&dptr,
-                key->ph.pool, (CUmemPoolPtrExportData*)&key->ph.ptr));
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(&dptr, *mpool,
+                (CUmemPoolPtrExportData*)&key->ph.ptr));
     if (status != UCS_OK) {
-        return status;
+        goto err_free_mpool;
     }
 
     status = uct_cuda_ipc_init_access_desc(&access_desc);
     if (status != UCS_OK) {
-        return status;
+        goto err_free_ptr;
     }
 
-    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolSetAccess(key->ph.pool,
-                &access_desc, 1));
+    status = UCT_CUDADRV_FUNC_LOG_ERR(
+                cuMemPoolSetAccess(*mpool, &access_desc, 1));
     if (status != UCS_OK) {
-        return status;
+        goto err_free_ptr;
     }
 
     *mapped_addr = dptr;
-
     return UCS_OK;
+
+err_free_ptr:
+    cuMemFree(dptr);
+err_free_mpool:
+    cuMemPoolDestroy(*mpool);
+err:
+    return status;
+}
+
+static ucs_status_t
+uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
+                                    CUdeviceptr *mapped_addr)
+{
+    khash_t(cuda_ipc_rem_mpool_cache) *hash = &uct_cuda_ipc_rem_mpool_cache.hash;
+    const CUmemFabricHandle *hkey           = &key->ph.handle.fabric_handle;
+    ucs_status_t status                     = UCS_OK;
+    CUmemoryPool mpool;
+    khiter_t khiter;
+    int khret;
+
+    pthread_rwlock_rdlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+    khiter = kh_get(cuda_ipc_rem_mpool_cache, hash, *hkey);
+    if (ucs_likely(khiter != kh_end(hash))) {
+        key->ph.pool = kh_val(hash, khiter);
+        ucs_assert(key->ph.pool != NULL);
+        goto out_import_pointer;
+    }
+
+    /* acquire writer lock as we did not find cached mpool in hashmap */
+    pthread_rwlock_unlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+    pthread_rwlock_wrlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+
+    khiter = kh_put(cuda_ipc_rem_mpool_cache, hash, *hkey, &khret);
+    if ((khret == UCS_KH_PUT_BUCKET_EMPTY) ||
+        (khret == UCS_KH_PUT_BUCKET_CLEAR)) {
+        status = cuda_ipc_rem_mpool_cache_create(key, &mpool, mapped_addr);
+        if (status != UCS_OK) {
+            /* Remove stale key from the hash if failed to construct new value.
+             * Otherwise next lookup will return NULL value. */
+            kh_del(cuda_ipc_rem_mpool_cache, hash, khiter);
+            goto err;
+        }
+
+        kh_val(hash, khiter) = mpool;
+        key->ph.pool         = mpool;
+    } else if (khret == UCS_KH_PUT_KEY_PRESENT) {
+        key->ph.pool = kh_val(hash, khiter);
+    } else {
+        ucs_error("unable to use cuda_ipc remote_cache hash: %d", khret);
+        status = UCS_ERR_NO_RESOURCE;
+        goto err;
+    }
+
+out_import_pointer:
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(mapped_addr,
+                key->ph.pool, (CUmemPoolPtrExportData*)&key->ph.ptr));
+
+err:
+    pthread_rwlock_unlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+    return status;
 }
 #endif
 
@@ -573,10 +666,30 @@ void uct_cuda_ipc_destroy_cache(uct_cuda_ipc_cache_t *cache)
 UCS_STATIC_INIT {
     ucs_recursive_spinlock_init(&uct_cuda_ipc_remote_cache.lock, 0);
     kh_init_inplace(cuda_ipc_rem_cache, &uct_cuda_ipc_remote_cache.hash);
+
+#if HAVE_CUDA_FABRIC
+    pthread_rwlock_init(&uct_cuda_ipc_rem_mpool_cache.lock, NULL);
+    /* Assumption: If import process succeeds, then the two nodes are in the
+     * same domain. Within a domain, fabric handles are expected to be unique.
+     * For this reason, there is no need to maintain a hashmap per peer OS as
+     * key collisions are not expected to occur. */
+    kh_init_inplace(cuda_ipc_rem_mpool_cache, &uct_cuda_ipc_rem_mpool_cache.hash);
+#endif
 }
 
 UCS_STATIC_CLEANUP {
     uct_cuda_ipc_cache_t *rem_cache;
+
+#if HAVE_CUDA_FABRIC
+    CUmemoryPool mpool;
+
+    kh_foreach_value(&uct_cuda_ipc_rem_mpool_cache.hash, mpool, {
+        cuMemPoolDestroy(mpool);
+    });
+    kh_destroy_inplace(cuda_ipc_rem_mpool_cache,
+                       &uct_cuda_ipc_rem_mpool_cache.hash);
+    pthread_rwlock_destroy(&uct_cuda_ipc_rem_mpool_cache.lock);
+#endif
 
     kh_foreach_value(&uct_cuda_ipc_remote_cache.hash, rem_cache, {
         uct_cuda_ipc_destroy_cache(rem_cache);
--- a/test/gtest/uct/cuda/test_cuda_ipc_md.cc
+++ b/test/gtest/uct/cuda/test_cuda_ipc_md.cc
@@ -14,14 +14,13 @@ extern "C" {
 
 class test_cuda_ipc_md : public test_md {
 protected:
-    static uct_cuda_ipc_rkey_t unpack(uct_md_h md, int64_t uuid)
+    static uct_cuda_ipc_rkey_t
+    unpack_common(uct_md_h md, int64_t uuid, CUdeviceptr ptr, size_t size)
     {
-        CUdeviceptr ptr;
-        EXPECT_EQ(CUDA_SUCCESS, cuMemAlloc(&ptr, 64));
         uct_mem_h memh;
-        EXPECT_UCS_OK(md->ops->mem_reg(md, (void *)ptr, 64, NULL, &memh));
         uct_cuda_ipc_rkey_t rkey;
-        EXPECT_UCS_OK(md->ops->mkey_pack(md, memh, (void *)ptr, 64, NULL,
+        EXPECT_UCS_OK(md->ops->mem_reg(md, (void *)ptr, size, NULL, &memh));
+        EXPECT_UCS_OK(md->ops->mkey_pack(md, memh, (void *)ptr, size, NULL,
                                          &rkey));
 
         int64_t *uuid64 = (int64_t *)rkey.uuid.bytes;
@@ -37,10 +36,66 @@ protected:
         params.field_mask = UCT_MD_MEM_DEREG_FIELD_MEMH;
         params.memh       = memh;
         EXPECT_UCS_OK(md->ops->mem_dereg(md, &params));
+        return rkey;
+    }
 
+    static uct_cuda_ipc_rkey_t unpack(uct_md_h md, int64_t uuid)
+    {
+        CUdeviceptr ptr;
+        EXPECT_EQ(CUDA_SUCCESS, cuMemAlloc(&ptr, 64));
+        uct_cuda_ipc_rkey_t rkey = unpack_common(md, uuid, ptr, 64);
         EXPECT_EQ(CUDA_SUCCESS, cuMemFree(ptr));
         return rkey;
     }
+
+#if HAVE_CUDA_FABRIC
+    static void alloc_mempool(CUdeviceptr *ptr, CUmemoryPool *mpool,
+                              CUstream *cu_stream, size_t size)
+    {
+        CUmemPoolProps pool_props = {};
+        CUmemAccessDesc map_desc;
+        CUdevice cu_device;
+
+        EXPECT_EQ(CUDA_SUCCESS, cuCtxGetDevice(&cu_device));
+
+        pool_props.allocType     = CU_MEM_ALLOCATION_TYPE_PINNED;
+        pool_props.location.id   = (int)cu_device;
+        pool_props.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+        pool_props.handleTypes   = CU_MEM_HANDLE_TYPE_FABRIC;
+        pool_props.maxSize       = size;
+        map_desc.flags           = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
+        map_desc.location        = pool_props.location;
+
+        EXPECT_EQ(CUDA_SUCCESS,
+                  cuStreamCreate(cu_stream, CU_STREAM_NON_BLOCKING));
+        EXPECT_EQ(CUDA_SUCCESS, cuMemPoolCreate(mpool, &pool_props));
+        EXPECT_EQ(CUDA_SUCCESS, cuMemPoolSetAccess(*mpool, &map_desc, 1));
+        EXPECT_EQ(CUDA_SUCCESS,
+                  cuMemAllocFromPoolAsync(ptr, size, *mpool, *cu_stream));
+        EXPECT_EQ(CUDA_SUCCESS, cuStreamSynchronize(*cu_stream));
+    }
+
+    static void
+    free_mempool(CUdeviceptr *ptr, CUmemoryPool *mpool, CUstream *cu_stream)
+    {
+        EXPECT_EQ(CUDA_SUCCESS, cuMemFree(*ptr));
+        EXPECT_EQ(CUDA_SUCCESS, cuMemPoolDestroy(*mpool));
+        EXPECT_EQ(CUDA_SUCCESS, cuStreamDestroy(*cu_stream));
+    }
+
+    static uct_cuda_ipc_rkey_t unpack_masync(uct_md_h md, int64_t uuid)
+    {
+        size_t size = 4 * UCS_MBYTE;
+        CUdeviceptr ptr;
+        CUmemoryPool mpool;
+        CUstream cu_stream;
+
+        alloc_mempool(&ptr, &mpool, &cu_stream, size);
+        uct_cuda_ipc_rkey_t rkey = unpack_common(md, uuid, ptr, size);
+        free_mempool(&ptr, &mpool, &cu_stream);
+        return rkey;
+    }
+#endif
 };
 
 UCS_MT_TEST_P(test_cuda_ipc_md, multiple_mds, 8)
@@ -68,5 +123,41 @@ UCS_MT_TEST_P(test_cuda_ipc_md, multiple_mds, 8)
         EXPECT_EQ(i, rkey.dev_num);
     }
 }
+
+#if HAVE_CUDA_FABRIC
+UCS_MT_TEST_P(test_cuda_ipc_md, multiple_mds_mempool, 8)
+{
+    cuda_context cuda_ctx;
+    ucs::handle<uct_md_h> md;
+    UCS_TEST_CREATE_HANDLE(uct_md_h, md, uct_md_close, uct_md_open,
+                           GetParam().component, GetParam().md_name.c_str(),
+                           m_md_config);
+
+    CUdeviceptr ptr;
+    CUmemoryPool mpool, q_mpool;
+    CUstream cu_stream;
+    CUmemFabricHandle fabric_handle;
+    CUresult cu_err;
+
+    alloc_mempool(&ptr, &mpool, &cu_stream, 64);
+    EXPECT_EQ(CUDA_SUCCESS, (cuPointerGetAttribute((void*)&q_mpool,
+                    CU_POINTER_ATTRIBUTE_MEMPOOL_HANDLE, ptr)));
+
+    cu_err = cuMemPoolExportToShareableHandle((void*)&fabric_handle, q_mpool,
+                                              CU_MEM_HANDLE_TYPE_FABRIC, 0);
+    free_mempool(&ptr, &mpool, &cu_stream);
+
+    if (cu_err == CUDA_SUCCESS) {
+        for (int64_t i = 0; i < 64; ++i) {
+            /* We get unique dev_num on new UUID */
+            uct_cuda_ipc_rkey_t rkey = unpack_masync(md, i + 1);
+            EXPECT_EQ(i, rkey.dev_num);
+            /* Subsequent call with the same UUID returns value from cache */
+            rkey = unpack_masync(md, i + 1);
+            EXPECT_EQ(i, rkey.dev_num);
+        }
+    }
+}
+#endif
 
 _UCT_MD_INSTANTIATE_TEST_CASE(test_cuda_ipc_md, cuda_ipc);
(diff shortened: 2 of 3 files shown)


src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -204,32 +274,36 @@ static ucs_status_t
...
+                *key->ph.pool, (CUmemPoolPtrExportData*)&key->ph.ptr));
     if (status != UCS_OK) {
         return status;
     }
 
-    status = uct_cuda_ipc_init_access_desc(&access_desc);
-    if (status != UCS_OK) {
-        return status;
-    }
+    if (!key_present) {
```
iyastreb, line 296: If it needs to be called once during hash element creation, then I propose:
- add a function that creates mpool object (uct_cuda_ipc_get_rem_mpool_create), so it calls all these functions inside:
  ucs_malloc
  cuMemPoolImportFromShareableHandle
  uct_cuda_ipc_init_access_desc
  cuMemPoolSetAccess
- call it from `uct_cuda_ipc_get_rem_mpool_cache`
- get rid of `key_present` return arg

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -48,6 +48,76 @@ typedef struct uct_cuda_ipc_remote_cache {
...
+           uct_cuda_ipc_rem_mpool_cache_hash_equal);
+
+typedef struct uct_cuda_ipc_rem_mpool_cache {
+    khash_t(cuda_ipc_rem_mpool_cache) hash;
+    ucs_recursive_spinlock_t          lock;
+} uct_cuda_ipc_rem_mpool_cache_t;
+
+uct_cuda_ipc_rem_mpool_cache_t uct_cuda_ipc_rem_mpool_cache;
+
+static ucs_status_t
```
iyastreb, line 76: Maybe just return `CUmemoryPool *` from this function?

Akshay-Venkatesh (commented):
@iyastreb Looks like my review didn't go through for some reason. Let me know if I've responded to your questions.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
iyastreb: Another potential flaw is race condition: we lock mutex only during cache creation, but not during setting access. This means that it's possible that from another thread you might retrieve an object from cache, which is constructed but without access set yet.

To overcome this and previous issue I think it makes sense to redesign this function:
```
uct_cuda_ipc_open_memhandle_mempool
    // try to find an element in cache, without updating the cache
    cache = uct_cuda_ipc_find_rem_mpool_cache (lock - unlock)
    if (cache) {
        // This is hot path
        return ImportPointer(cache)
    }
    // cache is not found, construct a new one
    cache = uct_cuda_ipc_create_rem_mpool_cache
                       ucs_malloc
                       cuMemPoolImportFromShareableHandle
                       ImportPointer
                       uct_cuda_ipc_init_access_desc
                       cuMemPoolSetAccess
    // Need to check that cache is not present - may be inserted by another thread
    uct_cuda_ipc_insert_rem_mpool_cache (cache, lock - unlock)
```
Yes, in slow path we lock twice, but for a very short period of time.
In hot path there is just one lock  
This way we:
1. acquire the lock only for the khash get/put operations, and not holding it during the malloc and cuMemPoolImportFromShareableHandle calls
2. Always have fully consistent objects in cache, not half-constructed
3. Update operation is atomic, no stale cache entries possible, no race conditions

iyastreb: Ok, now it's consistent and correct, but I'm not sure about performance. Few things to consider:
1. Single lock for the entire time-consuming operation - which means that we block all the other threads even for disjoint requests
2. Spin lock might not be the good choice for long operations, due to busy waiting
3. Global singleton lock (even for readers) might be a source of contention and poor scalability - maybe consider RW lock?

2 important things that I don't know:
- what is the overall delay of the operation in worst case? ~1ms?
- how many different fabric_handle we are supposed to handle? 10/100/1000?

If delay is small and amount of entries is low, then please disregard my comment.
If the overall delay is considerably long, and amount of updates is high, then ideal solution would be: use **RW lock** wiht 2 fine-grained locks:
```
uct_cuda_ipc_open_memhandle_mempool
    // try to find an element in cache, without updating the cache
    // READ lock is scalable across the READERS, after cache is updated we don't
    // even block the readers
    cache = uct_cuda_ipc_find_rem_mpool_cache (READ lock - unlock)
    if (cache) {
        // This is hot path
        return ImportPointer(cache)
    }
    // cache is not found, construct a new one
    cache = uct_cuda_ipc_create_rem_mpool_cache
                       ucs_malloc
                       cuMemPoolImportFromShareableHandle
                       ImportPointer
                       uct_cuda_ipc_init_access_desc
                       cuMemPoolSetAccess
    // Need to check that cache is not present - may be inserted by another thread
    // Only this part is mutually exclusive = blocks all threads, but for a very short period of time
    uct_cuda_ipc_insert_rem_mpool_cache (cache, WRITE lock - unlock)
```

Akshay-Venkatesh: @iyastreb Do the latest changes look good?

Akshay-Venkatesh: @iyastreb 
>> 2 important things that I don't know:
    what is the overall delay of the operation in worst case? ~1ms?
    how many different fabric_handle we are supposed to handle? 10/100/1000?

Worst case will be ~1ms but it depends on driver internal operations. This cache is expected to handle in the range of 10-100 fabric handles with majority operations being reads rather than updates. 

Thanks for RW lock usage suggestion. I'm not sure if khash datastructure is thread-safe where one of the threads could be updating the hashmap while others can read unrelated keys and associated values (for eg, what happens if one of the updates needs to allocate more memory and update hashmap metadata?). For this reason, in commit https://github.com/openucx/ucx/pull/10082/commits/fa00e3129627e2efdc5d1abbda2581c931bbd6be, I switched to using a single RW lock for readers and writers. All readers use rdlock to check if key is in hash and use imported pool to generate mapped pointer if key is present. If not, one writer uses wrlock to update the hash and use imported mpool for mapped pointer. This doesn't allow concurrency between readers and writers of unrelated keys but it should permit multiple readers to benefit from not waiting for other reads to complete while there are no writes. 

Let me know if the changes look good and if you have other thoughts.

iyastreb: Thank you for addressing my comments! Now it is safe.
You are right, khash impl is not thread-safe for disjoint updates, so in our case a writer blocks all the readers, which is expected. For disjoint access parallelism we would need to use concurrent data structures (or STM) which we don't have at disposal

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -205,36 +239,79 @@ uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
...
+            goto err;
+        }
+
+        status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportFromShareableHandle(
+               mpool, (void *)&key->ph.handle.fabric_handle, CU_MEM_HANDLE_TYPE_FABRIC, 0));
+        if (status != UCS_OK) {
+            ucs_free(mpool);
+            goto err;
+        }
+        kh_val(&uct_cuda_ipc_rem_mpool_cache.hash, khiter) = mpool;
```
iyastreb, line 283: empty line before, and align on =

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -205,36 +239,79 @@ uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
...
+                    *key->ph.pool, (CUmemPoolPtrExportData*)&key->ph.ptr));
+        goto err;
     }
 
-    status = uct_cuda_ipc_init_access_desc(&access_desc);
-    if (status != UCS_OK) {
-        return status;
+    /* acquire writer lock as we did not find cached mpool in hashmap */
+    pthread_rwlock_unlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+    pthread_rwlock_wrlock(&uct_cuda_ipc_rem_mpool_cache.lock);
```
iyastreb, line 261: Just an idea, I'm not requesting this change: 
This is correct logic, but can we do better in terms of scalability?
What if we don't acquire a write lock until we construct a new object?
This way we drastically reduce the scope of the write lock - it will be held only during khash update basically:
```
pthread_rwlock_unlock
// Construct new object
mpool = ucs_malloc(sizeof(*mpool), "cuda_ipc_mempool");
cuMemPoolImportFromShareableHandle
uct_cuda_ipc_init_access_desc
cuMemPoolSetAccess
// now acquire the write lock and update the hashmap
pthread_rwlock_wrlock
khiter = kh_put
// Here we need to check if khash is updated already for this key
if (khret == UCS_KH_PUT_KEY_PRESENT)
  ucs_free(mpool); // Release tmp mpool, and use a value from hash
```
The pros is scalability, the cons is more complexity and potentially more CPU cycles (can lead to parallel object creation for the same key). Just to consider, up to you

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -573,11 +650,30 @@ void uct_cuda_ipc_destroy_cache(uct_cuda_ipc_cache_t *cache)
...
+#endif
 }
 
 UCS_STATIC_CLEANUP {
     uct_cuda_ipc_cache_t *rem_cache;
 
+#if HAVE_CUDA_FABRIC
+    CUmemoryPool *mpool;
+    kh_foreach_value(&uct_cuda_ipc_rem_mpool_cache.hash, mpool, {
+        cuMemPoolDestroy(*mpool);
```
iyastreb, line 670: Should we also destroy all keys allocated with `ucs_strndup`?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -205,36 +239,79 @@ uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
...
-    status = uct_cuda_ipc_init_access_desc(&access_desc);
-    if (status != UCS_OK) {
-        return status;
+    /* acquire writer lock as we did not find cached mpool in hashmap */
+    pthread_rwlock_unlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+    pthread_rwlock_wrlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+
+    khiter = kh_put(cuda_ipc_rem_mpool_cache,
+                    &uct_cuda_ipc_rem_mpool_cache.hash,
+                    ucs_strndup((char*)&key->ph.handle.fabric_handle,
```
iyastreb, line 265: Is it safe in terms of a memory leak? What is `kh_put` fails?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -48,6 +48,40 @@ typedef struct uct_cuda_ipc_remote_cache {
...
+
+    for (i = 0; i < sizeof(CUmemFabricHandle); i++) {
+        c = *s++;
+        hash = ((hash << 5) + hash) + c;
+    }
+
+    return hash;
+}
+
+KHASH_INIT(cuda_ipc_rem_mpool_cache, kh_cstr_t, CUmemoryPool*, 1,
```
iyastreb, line 73: Given that string key requires alloc/free operations - wouldn't be better just to have `CUmemFabricHandle` as a key? And then treat it as a byte array in hash function

test/gtest/uct/cuda/test_cuda_ipc_md.cc
```diff
@@ -41,6 +41,60 @@ class test_cuda_ipc_md : public test_md {
         EXPECT_EQ(CUDA_SUCCESS, cuMemFree(ptr));
         return rkey;
     }
+
+#if HAVE_CUDA_FABRIC
+    static uct_cuda_ipc_rkey_t unpack_masync(uct_md_h md, int64_t uuid)
+    {
+        size_t size = (1 << 22);
```
iyastreb, line 48: align =

test/gtest/uct/cuda/test_cuda_ipc_md.cc
```diff
@@ -41,6 +41,60 @@ class test_cuda_ipc_md : public test_md {
...
+        map_desc.flags           = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
+        map_desc.location        = pool_props.location;
+
+        EXPECT_EQ(CUDA_SUCCESS, cuStreamCreate(&cu_stream, CU_STREAM_NON_BLOCKING));
+        EXPECT_EQ(CUDA_SUCCESS, cuMemPoolCreate(&mpool, &pool_props));
+        EXPECT_EQ(CUDA_SUCCESS, cuMemPoolSetAccess(mpool, &map_desc, 1));
+        EXPECT_EQ(CUDA_SUCCESS, cuMemAllocFromPoolAsync(&ptr, size, mpool, cu_stream));
+        EXPECT_EQ(CUDA_SUCCESS, cuStreamSynchronize(cu_stream));
+
+        uct_mem_h memh;
```
iyastreb, line 72: The part below is identical to existing `unpack` method, except the allocation/free part.
Maybe for code reuse purpose we encapsulate the common part into `unpack_impl` method, that does only reg-pack-unpack-dereg without memory alloc/free?

test/gtest/uct/cuda/test_cuda_ipc_md.cc
iyastreb, line 48: You can use `UCS_MBYTE` etc.
minor: maybe for unit testing purposes it's enough to allocate few KBs?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -205,36 +239,79 @@ uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
...
-                &access_desc, 1));
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(mapped_addr,
+                *key->ph.pool, (CUmemPoolPtrExportData*)&key->ph.ptr));
     if (status != UCS_OK) {
-        return status;
+        goto err;
     }
 
-    *mapped_addr = dptr;
+    if (khret != UCS_KH_PUT_KEY_PRESENT) {
```
brminich, line 299: is it possible to move to line 283 above and remove this if?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -205,36 +239,79 @@ uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
...
+        status = uct_cuda_ipc_init_access_desc(&access_desc);
+        if (status != UCS_OK) {
+            goto err;
```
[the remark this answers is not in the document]
Akshay-Venkatesh: Does it not suffice that all mpool objects are freed when the hash map is destroyed during cleanup?
```C
672     kh_foreach_value(&uct_cuda_ipc_rem_mpool_cache.hash, mpool, {
673         cuMemPoolDestroy(*mpool);
674         ucs_free(mpool);
675     });
```

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: @iyastreb This was previously the case but it ran into cache lookup issues when the actual key was freed. This is why I had to duplicate a copy of the key once per key, value insertion.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: Thanks for the suggestions @iyastreb. 

Multiple threads calling `ImportFromShareableHandle` is undesirable because: 
1. This PR is aiming to avoid multiple calls to ImportFromShareable/SetMemAccess as they are expensive by caching the calls to these operations. Not having a lock/unlock around these calls would negate that purpose.
2. It would lead to multiple mappings of the same remote allocation and all but one thread have to release their mappings after the subsequent lookup. Ultimately only one of the threads has done useful work.

As the costly critical path is stressed once per unique key, IMO it's better to keep existing logic. Let me know if you think there is still room for optimization.

test/gtest/uct/cuda/test_cuda_ipc_md.cc
Akshay-Venkatesh: I can go lower but internally all allocations between 1byte and 4MB take about the same time. I think it's ok to use 4MB here.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: Fixed this.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -573,11 +652,33 @@ void uct_cuda_ipc_destroy_cache(uct_cuda_ipc_cache_t *cache)
...
 
+#if HAVE_CUDA_FABRIC
+    CUmemoryPool *mpool;
+    const char *key;
+
+    kh_foreach_value(&uct_cuda_ipc_rem_mpool_cache.hash, mpool, {
+        cuMemPoolDestroy(*mpool);
+        ucs_free(mpool);
+    });
+    kh_foreach_key(&uct_cuda_ipc_rem_mpool_cache.hash, key, free((void*)key));
```
iyastreb, line 677: ucs_free, casting is redundant

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
iyastreb: Ok, I see. Still dynamic string does not look like the best choice, because of malloc/free overhead and associated error handling. The size is always the same for all entries, so maybe we just declare a key like that:
```C
typedef struct {
    uint8_t bytes[sizeof(CUmemFabricHandle)];
} cuda_ipc_rem_mpool_cache_key_t;
```

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -205,36 +239,81 @@ uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
...
 
-    status = uct_cuda_ipc_init_access_desc(&access_desc);
-    if (status != UCS_OK) {
-        return status;
+    /* acquire writer lock as we did not find cached mpool in hashmap */
+    pthread_rwlock_unlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+    pthread_rwlock_wrlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+
+    hkey   = ucs_strndup((char*)&key->ph.handle.fabric_handle,
+                         sizeof(CUmemFabricHandle), "fabric_handle");
```
iyastreb, line 265: then we also need to check if ucs_strndup returned NULL..
Better avoid that with struct I described above

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: Compiler complains without the cast.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: Hi @iyastreb, 
This is still not the common case and the overhead of allocating the key occurs whenever key was found in the hashmap. If I understand correctly, your suggestion is to avoid the call to `ucs_strndup` and have a local variable (of type `cuda_ipc_rem_mpool_cache_key_t`) in the function to avoid heap allocation overhead. However, the reason why have to duplicate the fabric handle (`key->ph.handle.fabric_handle`) and not just use it is because it appears that the hashmap expects the keys to have the same lifetime as the hashmap (maybe because the hashmap data structure doesn't make a copy of the key). As the fabric handle lifetime expires after the associated rkey's role completes, using a copy of the fabric handle allows the keys to persist until a later point when we destroy both the keys and the hashmap itself. For this reason, I think dynamic allocation of they hash key is unavoidable. Let me know if I've mistaken your suggestion. For now, I'll only add logic to handle ucs_strndup returning NULL.

Akshay-Venkatesh:
@yosefe Please provide a review when possible. Most if not all comments have been addressed until now.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
iyastreb: Actually khash does copy keys, and key can be anything: number, pointer, structure, whatever:
```C
#define __KHASH_TYPE(name, khkey_t, khval_t) \
	typedef struct kh_##name##_s { \
		khkey_t *keys; \

	SCOPE khint_t kh_put_##name(kh_##name##_t *h, khkey_t key, int *ret) \
	{																	\
...
			h->keys[x] = key;											\
```
So it can perfectly work with structures as well. I admit that there could be some consequences of having structure as a key:
 - khash preallocates key array, which may provoke higher memory usage.
 But in this case I think it's negligible, because even on maximum size of 100 with dynamic allocation we store:
 128x8B (pointer key) + 100x64B (dynamic key) = 7424B
 128x64B (fixed struct key) = 8192B
 - key is passed **by value** to hash_code and equals functions
This is optimised away by compiler (I checked that), and code without pointer actually works faster.

Let me do this little change along with some other minor fixes

Akshay-Venkatesh: > Actually khash does copy keys, and key can be anything: number, pointer, structure, whatever:

@iyastreb Thanks for the explanation. If khash does perform a copy, do we need an explicit structure(`cuda_ipc_rem_mpool_cache_key_t`) for the key? Why not just use `CUmemFabricHandle`? Also, if copies are being performed, was it the case that previously pointers to the key were being copied as opposed to the key contents which caused the failures that I saw?

iyastreb: But this is exactly what I proposed previously: https://github.com/openucx/ucx/pull/10082#discussion_r1743138544
You said "This was previously the case but it ran into cache lookup issues when the actual key was freed". But it seems that previously you tried to store the key by **pointer**, not by value.
Of course I agree that it's better to keep things simple and just use `CUmemFabricHandle` as a key.
I implemented this change

iyastreb (dismissed):

test/gtest/uct/cuda/test_cuda_ipc_md.cc
```diff
@@ -67,6 +122,34 @@ UCS_MT_TEST_P(test_cuda_ipc_md, multiple_mds, 8)
...
+    alloc_mempool(&ptr, &mpool, &cu_stream, 64);
+    EXPECT_EQ(CUDA_SUCCESS, (cuPointerGetAttribute((void*)&q_mpool,
+                    CU_POINTER_ATTRIBUTE_MEMPOOL_HANDLE, ptr)));
+
+    cu_err = cuMemPoolExportToShareableHandle((void*)&fabric_handle, q_mpool,
+                                              CU_MEM_HANDLE_TYPE_FABRIC, 0);
+    free_mempool(&ptr, &mpool, &cu_stream);
+
+    if (cu_err == CUDA_SUCCESS) {
+        for (int64_t i = 0; i < 64; ++i) {
```
brminich, line 142: why is `int_64_t` needed?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -200,41 +234,107 @@ uct_cuda_ipc_open_memhandle_vmm(uct_cuda_ipc_rkey_t *key,
...
+
+    pthread_rwlock_rdlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+    khiter = kh_get(cuda_ipc_rem_mpool_cache, hash, hkey);
+    if (ucs_likely(khiter != kh_end(hash))) {
+        key->ph.pool = kh_val(hash, khiter);
+        ucs_assert(key->ph.pool != NULL);
+
+        status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(mapped_addr,
+                    key->ph.pool, (CUmemPoolPtrExportData*)&key->ph.ptr));
+        goto err;
```
brminich, line 301: do you think it is performance-wise reasonable trying to `kh_get` at first with rd lock?
You could just do put immediately and take existing value if `(khret == UCS_KH_PUT_KEY_PRESENT)`
This would be a bit more intrusive, but code will be simpler

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -200,41 +234,107 @@ uct_cuda_ipc_open_memhandle_vmm(uct_cuda_ipc_rkey_t *key,
...
+    }
+
+    /* acquire writer lock as we did not find cached mpool in hashmap */
+    pthread_rwlock_unlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+    pthread_rwlock_wrlock(&uct_cuda_ipc_rem_mpool_cache.lock);
+
+    khiter = kh_put(cuda_ipc_rem_mpool_cache, hash, hkey, &khret);
+    if ((khret == UCS_KH_PUT_BUCKET_EMPTY) ||
+        (khret == UCS_KH_PUT_BUCKET_CLEAR)) {
+        status = cuda_ipc_rem_mpool_cache_create(key, &mpool, mapped_addr);
```
brminich, line 311: seems like you can avoid using pointer to pointer and pass just a pointer to some local `CUmemoryPool` var, which will be stored in hash anyway

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -200,41 +234,107 @@ uct_cuda_ipc_open_memhandle_vmm(uct_cuda_ipc_rkey_t *key,
...
+        kh_val(hash, khiter) = *mpool;
+        key->ph.pool         = *mpool;
+    } else if (khret == UCS_KH_PUT_KEY_PRESENT) {
+        key->ph.pool = kh_val(hash, khiter);
+        status       = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(
+                            mapped_addr, key->ph.pool,
+                            (CUmemPoolPtrExportData*)&key->ph.ptr));
+        if (status != UCS_OK) {
+            goto err;
+        }
```
brminich, line 328: minor: can remove this if

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
iyastreb: Let me respond here, because it was me who suggested to use rwlock.
The aim of using rwlock here was to minimise the contention, so that multiple readers can progress simultaneously without blocking each other at all (when cache is populated). Typically it's desirable to have less global contention points. I don't think it complicates a lot

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
iyastreb: right, let me fix that, since I added this function

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
iyastreb: ok

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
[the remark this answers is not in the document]
iyastreb: done

brminich: ok

test/gtest/uct/cuda/test_cuda_ipc_md.cc
iyastreb: well, this is to match `unpack_masync` function that takes `int64_t` as input arg

iyastreb (dismissed):

brminich (commented):
:+1 besides minor comments

brminich (dismissed):

Akshay-Venkatesh:
@iyastreb @brminich Thanks for the reviews and changes. I just tested with the latest commit and all MNNVL tests are passing as expected.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -200,41 +234,96 @@ uct_cuda_ipc_open_memhandle_vmm(uct_cuda_ipc_rkey_t *key,
...
-        return status;
+        goto err;
     }
 
-    *mapped_addr = dptr;
-
     return UCS_OK;
+
+err:
+    return status;
```
yosefe, line 271: why replaced "return status" by "goto err"?
do we need revert the import operation of a subsequent API call (such as cuMemPoolSetAccess) fails?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -200,41 +234,96 @@ uct_cuda_ipc_open_memhandle_vmm(uct_cuda_ipc_rkey_t *key,
...
+}
+
+static ucs_status_t
+uct_cuda_ipc_open_memhandle_mempool(uct_cuda_ipc_rkey_t *key,
+                                    CUdeviceptr *mapped_addr)
+{
+    khash_t(cuda_ipc_rem_mpool_cache) *hash = &uct_cuda_ipc_rem_mpool_cache.hash;
+    CUmemFabricHandle hkey                  = key->ph.handle.fabric_handle;
```
yosefe, line 279: can it be const ptr to fabric handle?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -200,41 +234,96 @@ uct_cuda_ipc_open_memhandle_vmm(uct_cuda_ipc_rkey_t *key,
...
+            goto out;
+        }
+
+        kh_val(hash, khiter) = mpool;
+        key->ph.pool         = mpool;
+    } else if (khret == UCS_KH_PUT_KEY_PRESENT) {
+        key->ph.pool = kh_val(hash, khiter);
+        status       = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(
+                            mapped_addr, key->ph.pool,
+                            (CUmemPoolPtrExportData*)&key->ph.ptr));
```
yosefe, line 317: seems cuMemPoolImportPointer is called in 3 different flows:
1. value is found using kh_get
2. value is added in cuda_ipc_rem_mpool_cache_create
3. value is found using kh_put (here)

can we unite these flows, so it will be called once, after all the khash checks? for example add label "out_import_pointer:"

test/gtest/uct/cuda/test_cuda_ipc_md.cc
```diff
@@ -67,6 +122,34 @@ UCS_MT_TEST_P(test_cuda_ipc_md, multiple_mds, 8)
         rkey = unpack(md, i + 1);
         EXPECT_EQ(i, rkey.dev_num);
     }
+
+#if HAVE_CUDA_FABRIC
```
yosefe, line 126: maybe create a separate test function for mempool allocations?

yosefe (commented):
so the operation of importing a single pointer from a memory pool, or fabric memory without mpool, are not considered expensive?

test/gtest/uct/cuda/test_cuda_ipc_md.cc
iyastreb: done

brminich (approved):

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: @yosefe This is required order as it stands. Can we keep this flow for now and we can reconsider post-release?

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
Akshay-Venkatesh: I think it makes sense to call `cuMemPoolDestroy(mpool)` before `goto err` as per @yosefe's suggestion. @iyastreb would you mind making the change?

iyastreb: np, I'll do it

yosefe (dismissed):
pls squash

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
iyastreb: I added a bit more than that:
```C
err_free_ptr:
    cuMemFree(dptr);   << Also destroy imported memory
err_free_mpool:
    cuMemPoolDestroy(*mpool);
    return status;
```
As per documentation of `cuMemPoolImportPointer `:
> The imported memory must not be accessed before the allocation operation completes in the exporting process. The imported memory must be freed from all importing processes before being freed in the exporting process. The pointer may be freed with cuMemFree or cuMemFreeAsync.

src/uct/cuda/cuda_ipc/cuda_ipc_cache.c
```diff
@@ -200,41 +234,99 @@ uct_cuda_ipc_open_memhandle_vmm(uct_cuda_ipc_rkey_t *key,
...
         return status;
     }
 
-    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(&dptr,
-                key->ph.pool, (CUmemPoolPtrExportData*)&key->ph.ptr));
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuMemPoolImportPointer(&dptr, *mpool,
+                (CUmemPoolPtrExportData*)&key->ph.ptr));
     if (status != UCS_OK) {
-        return status;
+        goto err_free_mpool;
```
yosefe, line 255: shall we change line 249 to "goto err" for consistency?

iyastreb: sure

brminich (approved):

PR #10071, opened 2024-08-20
AZP: Use different ports in one wire-compat job

brminich:
## What
Use different ports in one wire-compat job to avoid CI failures like
```
[1724129739.884304] [swx-rdmz-ucx-new-02:17268:0]          parser.c:2307 UCX  WARN  (set UCX_WARN_UNUSED_ENV_VARS=n to suppress this warning)
[1724129739.884905] [swx-rdmz-ucx-new-02:17268:0]          parser.c:2314 UCX  INFO  UCX_* env variables: UCX_WARN_UNUSED_ENV_VARS=n UCX_TLS=^sm UCX_CM_REUSEADDR=y UCX_IB_ROCE_LOCAL_SUBNET=y UCX_LOG_LEVEL=info
[1724129740.029045] [swx-rdmz-ucx-new-02:17268:0] rdmacm_listener.c:83   UCX  DIAG  rdma_bind_addr(addr=0.0.0.0:22000) failed: Cannot assign requested address
[1724129740.029084] [swx-rdmz-ucx-new-02:17268:0]    ucp_listener.c:242  UCX  DIAG  failed to create UCT listener on CM 0x10e1830 (component rdmacm) with address 0.0.0.0:22000 status Device is busy
failed to listen (Device is busy)
```

PR #10089, opened 2024-08-26
AZP: Use different ports in one wire-compat job - v1.17.x

brminich:
## What
port #10071

### DIFF
--- a/buildlib/tools/test_wire_compat.sh
+++ b/buildlib/tools/test_wire_compat.sh
@@ -8,10 +8,11 @@
 exe_name=$1
 client_opt=$2
 common_opt=$3
-port=$((10000 + 1000 * ${AZP_AGENT_ID}))
+port=$((10000 + 1000 * ${AZP_AGENT_ID} + 100 * ${WIRE_COMPAT_STAGE_ID}))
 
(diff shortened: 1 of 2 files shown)


buildlib/tools/test_wire_compat.sh
```diff
@@ -8,10 +8,11 @@
 exe_name=$1
 client_opt=$2
 common_opt=$3
-port=$((10000 + 1000 * ${AZP_AGENT_ID}))
+port=$((10000 + 1000 * ${AZP_AGENT_ID} + 100 * ${WIRE_COMPAT_STAGE_ID}))
```
iyastreb, line 11: I know I'm late with my comment, since it's porting existing change.
But is it reliable way to get free port, and avoid collisions?
What if those ports are taken by other services running on the machine?
I remember implementing something similar in the past, and the approach was:
- Check `LOCAL_PORT_MIN LOCAL_PORT_MAX` from `/proc/sys/net/ipv4/ip_local_port_range`
- Check if port is used with `netstat -an | grep "LISTEN "`
- Synchronize between multiple processes by using `flock` on common file

brminich: i think the proposed solution is good enough for now. We can change it if some issue arises

yosefe: we can implement a better solution on master branch

yosefe (approved):

PR #10091, opened 2024-08-26
UCS/VFS: Detect deletion of the monitored folder

iyastreb:
## What
Fix for https://redmine.mellanox.com/issues/4044773
We see that fuse thread is not interrupted from the main thread dtor, and remains waiting for data in the read:
```
Thread 2 (Thread 0x7fb4cffff700 (LWP 420349)):
#0  0x00007fb4de0abaa4 in read () from /lib64/libpthread.so.0
#1  0x00007fb4de6448f6 in ucs_vfs_fuse_wait_for_path (path=path@entry=0x7fb4cfffecd2 "/run/user/90540/ucx/vfs.sock") at vfs_fuse.c:311
#2  0x00007fb4de644c2d in ucs_vfs_fuse_thread_func (arg=<optimized out>) at vfs_fuse.c:418
Thread 1 (Thread 0x7fb4de7fb8c0 (LWP 420284)):
#0  0x00007fb4de0a36bd in __pthread_timedjoin_ex () from /lib64/libpthread.so.0
#1  0x00007fb4de6443d5 in ucs_fuse_thread_stop () at vfs_fuse.c:510
```
The only reason it can happen is that folder that we monitor (`/run/user/<user_id>/ucx`) gets deleted/recreated in the meantime. Currently we see this happening when tests are executed under `mtt` account.

## Why ?
We must get notified when `/run/user/<user_id>/ucx` folder gets deleted, so we must also handle **IN_IGNORED** event in addition to IN_CREATE

## How ?
On receiving this event we can either:
1. return error, so that VFS becomes efficiently disabled
2. return UCS_OK, so that we try again to create and monitor that folder
For now option 1 is implemented

### DIFF
--- a/src/ucs/vfs/fuse/vfs_fuse.c
+++ b/src/ucs/vfs/fuse/vfs_fuse.c
@@ -254,6 +254,16 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
 
     pthread_mutex_lock(&ucs_vfs_fuse_context.mutex);
 
+    /* Check 'stop' flag before entering the loop. If the main thread sets
+     * 'stop' flag before this thread created 'inotify_fd' fd, the execution
+     * of the thread has to be stopped, otherwise - the thread hangs waiting
+     * for the data on 'inotify_fd' fd.
+     */
+    if (ucs_vfs_fuse_context.stop) {
+        status = UCS_ERR_CANCELED;
+        goto out;
+    }
+
     /* Create directory path */
     ret = ucs_vfs_sock_mkdir(path, UCS_LOG_LEVEL_DIAG);
     if (ret != 0) {
@@ -282,23 +292,15 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
                      sizeof(watch_filename));
     watch_dirname = dirname(dir_buf);
 
-    /* Watch for new files in 'watch_dirname' */
+    /* Watch for new files in 'watch_dirname' and monitor if this watch gets
+     * deleted explicitly or implicitly */
     ucs_vfs_fuse_context.watch_desc = inotify_add_watch(
-            ucs_vfs_fuse_context.inotify_fd, watch_dirname, IN_CREATE);
+            ucs_vfs_fuse_context.inotify_fd, watch_dirname,
+            IN_CREATE | IN_IGNORED);
     if (ucs_vfs_fuse_context.watch_desc < 0) {
         ucs_error("inotify_add_watch(%s) failed: %m", watch_dirname);
         status = UCS_ERR_IO_ERROR;
         goto out_close_inotify_fd;
-    }
-
-    /* Check 'stop' flag before entering the loop. If the main thread sets
-     * 'stop' flag before this thread created 'inotify_fd' fd, the execution
-     * of the thread has to be stopped, otherwise - the thread hangs waiting
-     * for the data on 'inotify_fd' fd.
-     */
-    if (ucs_vfs_fuse_context.stop) {
-        status = UCS_ERR_CANCELED;
-        goto out_close_watch_id;
     }
 
     /* Read events from inotify channel and exit when either the main thread set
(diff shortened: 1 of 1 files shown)


src/ucs/vfs/fuse/vfs_fuse.c
```diff
@@ -330,6 +332,15 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
         for (offset  = 0; offset < nread;
              offset += (sizeof(*event) + event->len)) {
             event = UCS_PTR_BYTE_OFFSET(event_buf, offset);
+
+            /* Watch was removed explicitly (inotify_rm_watch) or automatically
+             * (file was deleted, or file system was unmounted). */
+            if (event->mask & IN_IGNORED) {
+                ucs_debug("inotify watch on '%s' was removed", watch_dirname);
```
yosefe, line 339: shall we check stop flag?

iyastreb: no, we shouldn't
Because we check it already on line 317 - in that case no matter what's event - we cancel the waiting.
Then we remain holding the mutex - so stop flag is not possible to set.
Here we are guaranteed that `stop` flag is not set, so we mostly check if the folder was removed from the filesystem (means implicitly, not from VFS)

Btw, regarding stop flag, we can also simplify the code above: move lines 294-302 to be right after mutex acquisition, so that when `stop` flag is set we detect it early and don't even bother creating folder, initializing inotify etc

brminich (dismissed):

src/ucs/vfs/fuse/vfs_fuse.c
```diff
@@ -282,25 +292,17 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
                      sizeof(watch_filename));
     watch_dirname = dirname(dir_buf);
 
-    /* Watch for new files in 'watch_dirname' */
+    /* Watch for new files in 'watch_dirname', and monitor if this watch gets
```
ivankochin, line 295:
```suggestion
    /* Watch for new files in 'watch_dirname' and monitor if this watch gets
```

src/ucs/vfs/fuse/vfs_fuse.c
yosefe: maybe we should try re-creating the watch in case it was removed?

src/ucs/vfs/fuse/vfs_fuse.c
```diff
@@ -254,6 +254,16 @@ static ucs_status_t ucs_vfs_fuse_wait_for_path(const char *path)
...
 
+    /* Check 'stop' flag before entering the loop. If the main thread sets
+     * 'stop' flag before this thread created 'inotify_fd' fd, the execution
+     * of the thread has to be stopped, otherwise - the thread hangs waiting
+     * for the data on 'inotify_fd' fd.
+     */
+    if (ucs_vfs_fuse_context.stop) {
+        status = UCS_ERR_CANCELED;
+        goto out;
+    }
```
yosefe, line 265: after moving checking of stop flag here, i think there could be a race, leading to a hang: the main thread will set stop=1 after the VFS thread already passed this point. Then, ucs_vfs_fuse_context.inotify_fd is still -1, so the main thread will not call inotify_rm_watch. Then, this VFS thread will set a watch and never wake up from it.
That's why the previos code checked stop flag only **after** creating a watch.

iyastreb: This assumption is wrong.
Bot fuse and main threads use mutex to synchronise access, which means that main thread cannot update the state in between, while fuse thread is holding the lock. They cannot execute mutually exclusive blocks simultaneously, that's the whole point of having mutex
```
fuse thread:
  lock
    check stop
   <<< NO interruption is possible here
    check inotify_fd = inotify_init
  unlock

main thread:
  lock
    stop = 1
    check inotify_fd
  unlock
```

yosefe: ok

src/ucs/vfs/fuse/vfs_fuse.c
iyastreb: I was thinking about this, and even implemented (and tested) this approach, it's really just a matter of 2 LOCs. Then I decided to not go this way, because:
- deletion of socket folder is not a part of the normal workflow, and it is not handled gracefully by VFS daemon. Daemon does not re-establish a socket listen if it gets deleted
- continue reconnection effort means we need to re-create the folder, try to connect to socket (and then redo watch) - this may consume CPU resources
- Currently function `ucs_vfs_sock_mkdir` that we use to create a folder, does not create a path (like `mkdir -p`), but creates just the last folder from path. So if the whole user folder is deleted (like happens with `mtt` user), we cannot recreate this path anyway

So I thought that it's not worth to implement re-connection in this case. But I don't have strong opinion on that, please tell me if you do

yosefe: ok, let'd keep as is, i guess if the folder is removed it's not a case we have to support

yosefe (approved):

PR #10094, opened 2024-08-26
ARCH/X86: Add CPU Model for MI300A - v1.16.x

edgargabriel:
## What
The MI300A has 24 Zen4 cores, which are architecturally (mostly) identical to the Genoa processors. They have however a different CPU id. Add the CPU ID to recognize CPU on this architecture.

## How ?
(cherry picked from commit 8b2113067bc5d158561c9564282d52c5e79d3bfe)

### DIFF
--- a/src/ucs/arch/x86_64/cpu.c
+++ b/src/ucs/arch/x86_64/cpu.c
@@ -472,7 +472,7 @@ ucs_cpu_model_t ucs_arch_get_cpu_model()
                 break;
             }
             break;
-        /* AMD Zen3 */
+        /* AMD Zen3/Zen4 */
         case 0x19:
             switch (model) {
             case 0x00:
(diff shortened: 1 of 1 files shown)


tvegas1 (approved):

brminich:
is there a reason to port it to 1.16? @yosefe, are we going to have another 1.16.x release?

yosefe:
> is there a reason to port it to 1.16? @yosefe, are we going to have another 1.16.x release?

Indeed we don't plan another release for 1.16.x. @edgargabriel did you intend to port it to 1.17.x?

edgargabriel:
@brminich @yosefe I did not anticipate another 1.16 release. However, we see benefits on having this fix on the 1.16 branch as well in case somebody is doing a git clone of this branch.

yosefe:
> @brminich @yosefe I did not anticipate another 1.16 release. However, we see benefits on having this fix on the 1.16 branch as well in case somebody is doing a git clone of this branch.

Unfortunately CI will not pass for v1.16.x branch, since it was not extended to support recent CI env changes. We don't usually maintain more than one release branch.

edgargabriel:
ok, in that case I will close this pr. We did something similar for 1.15 a while back (i.e. there is a fix in 1.15 that was committed to the branch after the 1.15 release date and is not part of any official release), and I followed the same model here. But if there are technical reasons why this cannot be accomplished in this case, we can skip this pr.

PR #10101, opened 2024-08-28
UCT/IB/MLX5/DC: dci_channel_index is not limited by num_dci_channels

roiedanino:
## What
Removed limitation of `dci_channel_index` by `num_dci_channels`

## Why ?
Improvement in performance, utilization of all 256 DCI channels

### DIFF
--- a/src/uct/ib/mlx5/dc/dc_mlx5_ep.h
+++ b/src/uct/ib/mlx5/dc/dc_mlx5_ep.h
@@ -368,8 +368,7 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
         }
         ep->dci               = 0;
         dci                   = uct_dc_mlx5_iface_dci(iface, ep->dci);
-        ep->dci_channel_index = dci->next_channel_index++ %
-                                iface->tx.num_dci_channels;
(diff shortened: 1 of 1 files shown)


src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -368,8 +368,7 @@ uct_dc_mlx5_ep_basic_init(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
         }
         ep->dci               = 0;
         dci                   = uct_dc_mlx5_iface_dci(iface, ep->dci);
-        ep->dci_channel_index = dci->next_channel_index++ %
-                                iface->tx.num_dci_channels;
+        ep->dci_channel_index = dci->next_channel_index++;
```
gleon99, line 371: Can't it grow infinitely?

roiedanino: dci_channel_index is uint8_t so it will go back to 0 after 255

brminich: how is it handled in HW? Do they do module by channel num themselves?

roiedanino:
> is there a specific test case that improved because of this change?

osu_all2all (when setting policy to hw_dcs)

PR #10104, opened 2024-08-28
UCP: Enable ppln protos with cuda buffers by default

brminich:
## What
Enable using cuda staging buffers for pipeline protocols by default.

## Why ?
Using ppln protocols with cuda staging buffers provides better performance for the following cases:
- sending data from cuda-managed memory to cuda memory
- sending data between different nodes where MNNVLK is available
- etc

## How ?
Rename `RNDV_FRAG_MEM_TYPE` to `RNDV_FRAG_MEM_TYPES` and allow to specify a list of memory types. Then initialize RNDV_RECEIVE protocol variants (rtr_mtype and get_mtype) with these memory types. RNDV_SEND protocol has to use the mem type which corresponds to the rkey from received RTR, because:
- We use different fragment sizes for different mem_types, so rtr_mtype and put_mtype sizes will be different which breaks the protocol
- Currently there is no strong use-case to support ppln protocols with different mem_type fragments being used by peers

### DIFF
--- a/src/ucp/am/eager_multi.c
+++ b/src/ucp/am/eager_multi.c
@@ -37,6 +37,7 @@ ucp_am_eager_multi_bcopy_proto_probe(const ucp_proto_init_params_t *init_params)
                                UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING |
                                UCP_PROTO_COMMON_INIT_FLAG_RESUME,
         .super.exclude_map   = 0,
+        .super.reg_mem_type  = UCS_MEMORY_TYPE_UNKNOWN,
         .max_lanes           = context->config.ext.max_eager_lanes,
         .initial_reg_md_map  = 0,
         .first.lane_type     = UCP_LANE_TYPE_AM,
@@ -197,6 +198,7 @@ ucp_am_eager_multi_zcopy_proto_probe(const ucp_proto_init_params_t *init_params)
                                UCP_PROTO_COMMON_INIT_FLAG_CAP_SEG_SIZE |
                                UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING,
         .super.exclude_map   = 0,
+        .super.reg_mem_type  = init_params->select_param->mem_type,
         .max_lanes           = context->config.ext.max_eager_lanes,
         .initial_reg_md_map  = 0,
         .opt_align_offs      = UCP_PROTO_COMMON_OFFSET_INVALID,
--- a/src/ucp/proto/proto_common.c
+++ b/src/ucp/proto/proto_common.c
@@ -433,8 +433,9 @@ ucp_lane_index_t
 ucp_proto_common_find_lanes(const ucp_proto_init_params_t *params,
                             uct_ep_operation_t memtype_op, unsigned flags,
                             ptrdiff_t max_iov_offs, size_t min_iov,
-                            ucp_lane_type_t lane_type, uint64_t tl_cap_flags,
-                            ucp_lane_index_t max_lanes,
+                            ucp_lane_type_t lane_type,
+                            ucs_memory_type_t reg_mem_type,
+                            uint64_t tl_cap_flags, ucp_lane_index_t max_lanes,
                             ucp_lane_map_t exclude_map, ucp_lane_index_t *lanes)
 {
     UCS_STRING_BUFFER_ONSTACK(sel_param_strb, UCP_PROTO_SELECT_PARAM_STR_MAX);
@@ -457,7 +458,7 @@ ucp_proto_common_find_lanes(const ucp_proto_init_params_t *params,
     }
 
     ucp_proto_select_info_str(params->worker, params->rkey_cfg_index,
-                              params->select_param, ucp_operation_names,
+                              select_param, ucp_operation_names,
                               &sel_param_strb);
 
     num_lanes = 0;
@@ -518,25 +519,29 @@ ucp_proto_common_find_lanes(const ucp_proto_init_params_t *params,
         }
 
         /* Check memory registration capabilities for zero-copy case */
-        if (flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY) {
+        if (reg_mem_type != UCS_MEMORY_TYPE_UNKNOWN) {
+            ucs_assertv((reg_mem_type == select_param->mem_type) ||
+                        !(flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY),
+                        "flags=0x%x reg_mem_type=%s select_param->mem_type=%s",
+                        flags, ucs_memory_type_names[reg_mem_type],
+                        ucs_memory_type_names[select_param->mem_type]);
+
             if (md_attr->flags & UCT_MD_FLAG_NEED_MEMH) {
                 /* Memory domain must support registration on the relevant
                  * memory type */
-                if (!(context->reg_md_map[select_param->mem_type] &
-                      UCS_BIT(md_index))) {
+                if (!(context->reg_md_map[reg_mem_type] & UCS_BIT(md_index))) {
                     ucs_trace("%s: md %s cannot register %s memory", lane_desc,
                               context->tl_mds[md_index].rsc.md_name,
-                              ucs_memory_type_names[select_param->mem_type]);
+                              ucs_memory_type_names[reg_mem_type]);
                     continue;
                 }
-            } else if (!(md_attr->access_mem_types &
-                         UCS_BIT(select_param->mem_type))) {
+            } else if (!(md_attr->access_mem_types & UCS_BIT(reg_mem_type))) {
                 /*
                  * Memory domain which does not require a registration for zero
                  * copy operation must be able to access the relevant memory type
                  */
                 ucs_trace("%s: no access to mem type %s", lane_desc,
-                          ucs_memory_type_names[select_param->mem_type]);
+                          ucs_memory_type_names[reg_mem_type]);
                 continue;
             }
         }
@@ -629,11 +634,11 @@ ucp_lane_index_t ucp_proto_common_find_lanes_with_min_frag(
     const uct_iface_attr_t *iface_attr;
     size_t tl_min_frag, tl_max_frag;
 
-    num_lanes = ucp_proto_common_find_lanes(&params->super, params->memtype_op,
-                                            params->flags, params->max_iov_offs,
-                                            params->min_iov, lane_type,
-                                            tl_cap_flags, max_lanes,
-                                            exclude_map, lanes);
+    num_lanes = ucp_proto_common_find_lanes(
+                   &params->super, params->memtype_op, params->flags,
+                   params->max_iov_offs, params->min_iov, lane_type,
+                   params->reg_mem_type, tl_cap_flags, max_lanes, exclude_map,
+                   lanes);
 
     num_valid_lanes = 0;
     for (lane_index = 0; lane_index < num_lanes; ++lane_index) {
--- a/src/ucp/proto/proto_common.h
+++ b/src/ucp/proto/proto_common.h
@@ -61,6 +61,7 @@ typedef enum {
 
     /* Supports starting the request when its datatype iterator offset is > 0 */
     UCP_PROTO_COMMON_INIT_FLAG_RESUME        = UCS_BIT(10),
+    UCP_PROTO_COMMON_KEEP_MD_MAP             = UCS_BIT(11)
 } ucp_proto_common_init_flags_t;
 
 
@@ -120,6 +121,13 @@ typedef struct {
 
     /* Map of unsuitable lanes */
     ucp_lane_map_t          exclude_map;
+
+    /* Memory type of the buffer used for data transfer on the transport level.
+     * If UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY flag is set, it is expected to
+     * be the user buffer memory type. Alternatively, it refers to the type of
+     * memory used for bounce buffers (either in the UCP or UCT layer) where
+     * data needs to be copied as part of the protocol. */
+    ucs_memory_type_t       reg_mem_type;
 } ucp_proto_common_init_params_t;
 
 
@@ -259,8 +267,9 @@ ucp_lane_index_t
 ucp_proto_common_find_lanes(const ucp_proto_init_params_t *params,
                             uct_ep_operation_t memtype_op, unsigned flags,
                             ptrdiff_t max_iov_offs, size_t min_iov,
-                            ucp_lane_type_t lane_type, uint64_t tl_cap_flags,
-                            ucp_lane_index_t max_lanes,
+                            ucp_lane_type_t lane_type,
+                            ucs_memory_type_t reg_mem_type,
+                            uint64_t tl_cap_flags, ucp_lane_index_t max_lanes,
                             ucp_lane_map_t exclude_map,
                             ucp_lane_index_t *lanes);
 
--- a/src/ucp/proto/proto_init.c
+++ b/src/ucp/proto/proto_init.c
@@ -376,6 +376,7 @@ ucp_proto_init_add_buffer_perf(const ucp_proto_common_init_params_t *params,
                                ucp_md_map_t reg_md_map, ucp_proto_perf_t *perf)
 {
     const ucp_proto_select_param_t *select_param = params->super.select_param;
+    ucs_memory_type_t buffer_mem_type;
     ucs_memory_type_t recv_mem_type;
     uint32_t op_attr_mask;
     ucs_status_t status;
@@ -390,8 +391,17 @@ ucp_proto_init_add_buffer_perf(const ucp_proto_common_init_params_t *params,
         }
     } else if (!(params->flags & UCP_PROTO_COMMON_INIT_FLAG_RKEY_PTR)) {
         ucs_assert(reg_md_map == 0);
+
+        /* TODO: This mem_type initialization is specific to put and get mtype
+         * protocols. Consider moving it to the corresponding probe functions.
+         */
+        if (params->reg_mem_type != UCS_MEMORY_TYPE_UNKNOWN) {
+            buffer_mem_type = params->reg_mem_type;
+        } else {
+            buffer_mem_type = UCS_MEMORY_TYPE_HOST;
+        }
         status = ucp_proto_init_add_buffer_copy_time(
-                params->super.worker, "local copy", UCS_MEMORY_TYPE_HOST,
+                params->super.worker, "local copy", buffer_mem_type,
                 select_param->mem_type, params->memtype_op, range_start,
                 range_end, 1, perf);
         if (status != UCS_OK) {
@@ -423,11 +433,8 @@ ucp_proto_init_add_buffer_perf(const ucp_proto_common_init_params_t *params,
             params->super.worker, "remote copy", UCS_MEMORY_TYPE_HOST,
             recv_mem_type, UCT_EP_OP_PUT_SHORT, range_start, range_end, 0,
             perf);
-    if (status != UCS_OK) {
-        return status;
-    }
 
-    return UCS_OK;
+    return status;
 }
 
 static int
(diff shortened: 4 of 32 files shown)


src/ucp/rndv/rndv_mtype.inl
```diff
@@ -169,18 +169,24 @@ static UCS_F_ALWAYS_INLINE ucs_status_t ucp_proto_rndv_mtype_copy(
...
-    ucp_ep_h mtype_ep          = ucp_proto_rndv_mtype_ep(params->worker,
-                                                         mem_type);
+    ucp_ep_h mtype_ep;
     ucp_lane_index_t lane;
     ucp_rsc_index_t rsc_index;
     const char *tl_name;
 
+    /* Make coverity happy */
+    ucs_assertv(frag_mem_type < UCS_MEMORY_TYPE_UNKNOWN, "frag_mem_type = %u",
+                frag_mem_type);
```
yosefe, line 186: 1. maybe move this inside ucp_proto_rndv_mtype_ep ?
2. remove space before and after =

src/ucp/rndv/rndv_put.c
```diff
@@ -345,8 +347,9 @@ ucp_proto_rndv_put_common_probe(const ucp_proto_init_params_t *init_params,
     if (send_atp) {
         ucs_assert(rpriv.atp_map != 0);
     }
-    rpriv.atp_num_lanes = ucs_popcount(rpriv.atp_map);
-    rpriv.stat_counter  = stat_counter;
+    rpriv.atp_num_lanes      = ucs_popcount(rpriv.atp_map);
+    rpriv.stat_counter       = stat_counter;
+    rpriv.bulk.frag_mem_type = frag_mem_type;
```
yosefe, line 352: i'd expect ucp_proto_rndv_bulk_init to set "bulk.frag_mem_type" field

src/ucp/rndv/rndv_put.c
```diff
@@ -514,6 +519,7 @@ ucp_proto_rndv_put_mtype_copy_progress(uct_pending_req_t *uct_req)
     req->flags |= UCP_REQUEST_FLAG_PROTO_INITIALIZED;
     ucp_proto_rndv_mtype_copy(req, req->send.rndv.mdesc->ptr,
                               ucp_proto_rndv_mtype_get_req_memh(req),
+                              req->send.rndv.mdesc->memh->mem_type,
```
yosefe, line 522: maybe reuse the mem_type from line 512 by caching it in a local var?

src/ucp/rndv/rndv_put.c
```diff
@@ -563,8 +569,20 @@ ucp_proto_rndv_put_mtype_probe(const ucp_proto_init_params_t *init_params)
...
 
-    status = ucp_proto_rndv_mtype_init(init_params, &mdesc_md_map, &frag_size);
+    if (init_params->rkey_config_key == NULL) {
+        frag_mem_type = UCS_MEMORY_TYPE_HOST;
+    } else {
+        /* Can initialize only the same fragment type as received in RTR
+         * because pipeline protocols assume that both peers use the same
+         * fragment sizes (and they are different for different memory types by
+         * default). */
+        frag_mem_type = init_params->rkey_config_key->mem_type;
```
yosefe, line 581: what happens if the memory type received in the RTR is not of a fragment, but belongs to a user buffer?

src/ucp/rndv/rndv_rkey_ptr.c
```diff
@@ -240,8 +242,9 @@ ucp_proto_rndv_rkey_ptr_mtype_probe(const ucp_proto_init_params_t *init_params)
         return;
     }
 
-    status = ucp_proto_rndv_mtype_init(init_params, &mdesc_md_map,
-                                       &params.super.max_length);
+    /* 2-stage ppln protocols work with host staging buffers only */
+    status = ucp_proto_rndv_mtype_init(init_params, UCS_MEMORY_TYPE_HOST,
```
yosefe, line 246: matbe we should check the supported memory types of rkey_ptr lane?

src/ucp/am/eager_multi.c
```diff
@@ -37,6 +37,7 @@ ucp_am_eager_multi_bcopy_proto_probe(const ucp_proto_init_params_t *init_params)
                                UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING |
                                UCP_PROTO_COMMON_INIT_FLAG_RESUME,
         .super.exclude_map   = 0,
+        .super.reg_mem_type  = UCS_MEMORY_TYPE_UNKNOWN,
```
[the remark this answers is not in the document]
brminich: what is the meaning of having a bitmap here? we are not supposed to use different memory type fragments in one variant of protocol

src/ucp/core/ucp_mm.c
```diff
@@ -1700,27 +1700,28 @@ ucs_status_t
...
                                   alloc_mem_type, "get_alloc_md_id",
                                   &mem);
-        if (status != UCS_OK) {
-            return status;
+        if (status == UCS_OK) {
+            context->alloc_md[alloc_mem_type].md_index =
+                              ucp_mem_get_md_index(context, mem.md, mem.method);
+            uct_mem_free(&mem);
+        } else {
+            context->alloc_md[alloc_mem_type].md_index = UCP_NULL_RESOURCE;
```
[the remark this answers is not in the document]
brminich: seems to be safer to set it explicitly here. Remove?

src/ucp/proto/proto_common.c
```diff
@@ -494,25 +495,23 @@ ucp_proto_common_find_lanes(const ucp_proto_init_params_t *params,
         }
 
         /* Check memory registration capabilities for zero-copy case */
-        if (flags & UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY) {
+        if (reg_mem_type != UCS_MEMORY_TYPE_UNKNOWN) {
```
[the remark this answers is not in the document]
brminich: ok

src/ucp/proto/proto_init.c
```diff
@@ -494,10 +495,18 @@ ucp_proto_common_init_send_perf(const ucp_proto_common_init_params_t *params,
         send_overhead = UCS_LINEAR_FUNC_ZERO;
     } else {
         ucs_assert(reg_md_map == 0);
+
+        if ((params->flags & UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS) &&
+            (params->reg_mem_type != UCS_MEMORY_TYPE_UNKNOWN)) {
+            remote_mem_type = params->reg_mem_type;
+        } else {
+            remote_mem_type = UCS_MEMORY_TYPE_HOST;
+        }
```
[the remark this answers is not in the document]
brminich: can you pls elaborate? Are you suggesting to add one more param to `ucp_proto_common_init_params_t`?
Imo, it looks ok - if no send_zcopy, but reg_mem_type is still set this means we have to do copy to this mem and use it for net transfer

src/ucp/rndv/proto_rndv.h
```diff
@@ -68,6 +68,7 @@ typedef struct {
  */
 typedef struct {
     ucp_proto_rndv_ack_priv_t super;
+    ucs_memory_type_t         frag_mem_type;
```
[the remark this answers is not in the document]
brminich: ok

src/ucp/rndv/rndv.c
```diff
@@ -21,6 +21,27 @@
 #include <ucs/datastruct/queue.h>
 
 
+static UCS_F_ALWAYS_INLINE int
+ucp_rndv_frag_memtype(ucp_context_t *context)
```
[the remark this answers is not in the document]
brminich: ok

src/ucp/rndv/rndv_get.c
```diff
@@ -125,7 +128,7 @@ ucp_proto_rndv_get_zcopy_probe(const ucp_proto_init_params_t *init_params)
             UCT_EP_OP_LAST,
             UCP_PROTO_COMMON_INIT_FLAG_SEND_ZCOPY |
             UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING,
-            0, 0);
+            0, 0, init_params->select_param->mem_type);
```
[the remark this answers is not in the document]
brminich: it's a new `reg_mem_type` which is equal to `select_params->mem_type` for `send_zopy`. I can rename it

src/ucp/rndv/rndv_rkey_ptr.c
brminich: why? we can update the code if/when new 2-stage ppln scheme is implemented

src/ucp/rndv/rndv_put.c
brminich: the code would still be correct. If it is cuda memory, cuda fragment will be selected (which is perf efficient). Otherwise, host chunk will be used (also beneficial on receiver)

src/ucp/rndv/rndv_put.c
brminich: ok

src/ucp/rndv/rndv_mtype.inl
```diff
@@ -82,6 +79,7 @@ static ucp_ep_h ucp_proto_rndv_req_mtype_ep(ucp_request_t *req)
     ucp_ep_h mem_type_ep;
 
     mem_type_ep = ucp_proto_rndv_mtype_ep(req->send.ep->worker,
+                                          req->send.rndv.mdesc->memh->mem_type,
```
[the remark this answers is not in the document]
yosefe: ok

src/ucp/rndv/rndv_rkey_ptr.c
yosefe: ok

src/ucp/core/ucp_mm.c
```diff
@@ -1696,31 +1696,66 @@ void ucp_mem_rcache_cleanup(ucp_context_h context)
...
 {
-    ucs_status_t status;
+    ucs_status_t status = UCS_OK;
     uct_allocated_memory_t mem;
 
     if (!context->alloc_md[alloc_mem_type].initialized) {
+        context->alloc_md[alloc_mem_type].initialized = 1;
+
+        if ((alloc_mem_type != UCS_MEMORY_TYPE_HOST) &&
+            !ucp_mm_alloc_memtype_device_ready(context, alloc_mem_type)) {
```
brminich, line 1736: use HIDE_ERRORS for alloc instead

src/ucp/core/ucp_mm.c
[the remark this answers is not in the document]
brminich: add assert instead

src/ucp/proto/proto_init.c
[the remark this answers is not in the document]
brminich: rename to frag_mem_type

src/ucp/rndv/rndv_get.c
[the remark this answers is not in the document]
brminich: rename to send_mem_type

src/ucp/rndv/rndv_get.c
```diff
@@ -275,6 +278,7 @@ ucp_proto_rndv_get_mtype_fetch_completion(uct_completion_t *uct_comp)
 
     ucp_proto_rndv_mtype_copy(req, req->send.rndv.mdesc->ptr,
                               ucp_proto_rndv_mtype_get_req_memh(req),
+                              req->send.rndv.mdesc->memh->mem_type,
```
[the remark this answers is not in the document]
brminich: try to refactor

src/ucp/rndv/rndv_put.c
```diff
@@ -564,8 +571,20 @@ ucp_proto_rndv_put_mtype_probe(const ucp_proto_init_params_t *init_params)
     ucp_md_map_t mdesc_md_map;
     ucs_status_t status;
     size_t frag_size;
+    ucs_memory_type_t frag_mem_type;
+
+    if (init_params->rkey_config_key == NULL) {
+        frag_mem_type = UCS_MEMORY_TYPE_HOST;
```
brminich, line 577: return immediately here as not supported

src/ucp/am/eager_multi.c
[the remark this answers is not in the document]
brminich: renamed to send_mem_type, but dif not convert to bitmap. To use bitmap we will always have to take 1st bit with ffs and assert that only one bit left, which looks uglier imo

src/ucp/am/eager_multi.c
```diff
@@ -37,7 +37,7 @@ ucp_am_eager_multi_bcopy_proto_probe(const ucp_proto_init_params_t *init_params)
                                UCP_PROTO_COMMON_INIT_FLAG_ERR_HANDLING |
                                UCP_PROTO_COMMON_INIT_FLAG_RESUME,
         .super.exclude_map   = 0,
-        .super.reg_mem_type  = UCS_MEMORY_TYPE_UNKNOWN,
+        .super.send_mem_type = UCS_MEMORY_TYPE_UNKNOWN,
```
yosefe, line 40: IMO for bcopy protocols, super.send_mem_type should be always host memory. which is ok because if SEND_ZCOPY is not set, we should not check for registration capability anyway. and it will be consistent with memory copy performance estimation.

src/ucp/core/ucp_mm.c
```diff
@@ -1700,28 +1700,27 @@ ucs_status_t
...
                                   alloc_mem_type, "get_alloc_md_id",
                                   &mem);
-        if (status == UCS_OK) {
-            context->alloc_md[alloc_mem_type].md_index =
-                              ucp_mem_get_md_index(context, mem.md, mem.method);
-            uct_mem_free(&mem);
-        } else {
-            context->alloc_md[alloc_mem_type].md_index = UCP_NULL_RESOURCE;
+        if (status != UCS_OK) {
+            return status;
```
yosefe, line 1713: why should we fail here? shouldn't we keep it as NULL_RESOURCE?

src/ucp/proto/proto_common.c
```diff
@@ -519,23 +519,29 @@ ucp_proto_common_find_lanes(const ucp_proto_init_params_t *params,
         }
 
         /* Check memory registration capabilities for zero-copy case */
-        if (reg_mem_type != UCS_MEMORY_TYPE_UNKNOWN) {
+        if (send_mem_type != UCS_MEMORY_TYPE_UNKNOWN) {
```
yosefe, line 522: check also for SEND_ZCOPY here, for consistency, since the name of the field is not reg_mem_type anymore. and remove the next assert in line 523

src/ucp/proto/proto_common.h
```diff
@@ -122,8 +122,12 @@ typedef struct {
     /* Map of unsuitable lanes */
     ucp_lane_map_t          exclude_map;
 
-    /* Memory type that the transport should be capable of registering. */
-    ucs_memory_type_t       reg_mem_type;
+    /* Memory type that the transport should be capable of registering.
+     * In case of zcopy protocols (e.g. rndv/put/zcopy, am/egr/single/zcopy,
+     * etc) it should be the same type as user buffer, which is specified
+     * in select_param->mem_type. For mtype protocols (rndv/put/mtype and
+     * rndv/get/mtype) it represents fragment memory type. */
```
yosefe, line 129: Memory type of the buffer used data transfer on the transport level. If SEND_ZCOPY is set, it is expected to be the user buffer memory type. Alternatively, it refers to the type of memory used for bounce buffers (either in the UCP or UCT layer) where data needs to be copied as part of the protocol.

src/ucp/proto/proto_init.c
```diff
@@ -451,8 +390,16 @@ ucp_proto_common_init_xfer_perf(const ucp_proto_common_init_params_t *params,
         }
     } else if (!(params->flags & UCP_PROTO_COMMON_INIT_FLAG_RKEY_PTR)) {
         ucs_assert(reg_md_map == 0);
+
+        if ((params->flags & UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS) &&
+            (params->send_mem_type != UCS_MEMORY_TYPE_UNKNOWN)) {
+            remote_mem_type = params->send_mem_type;
+        } else {
+            remote_mem_type = UCS_MEMORY_TYPE_HOST;
```
yosefe, line 398: can we always use  params->send_mem_type without extra conditions? and initialize the protocol parameters accordingly?

src/ucp/proto/proto_init.c
```diff
@@ -371,73 +371,12 @@ ucp_proto_init_add_buffer_copy_time(ucp_worker_h worker, const char *title,
...
-                                const ucp_proto_common_tl_perf_t *tl_perf,
-                                ucp_proto_perf_node_t *const tl_perf_node,
-                                ucp_proto_perf_range_t *xfer_perf)
+ucp_proto_init_add_buffer_perf(const ucp_proto_common_init_params_t *params,
+                               size_t range_start, size_t range_end,
+                               ucp_md_map_t reg_md_map, ucp_proto_perf_t *perf)
 {
     const ucp_proto_select_param_t *select_param = params->super.select_param;
-    ucs_memory_type_t recv_mem_type;
+    ucs_memory_type_t remote_mem_type;
```
yosefe, line 379: pls rename to frag_mem_type or buffer_mem_type

src/ucp/proto/proto_init.c
```diff
@@ -451,8 +390,16 @@ ucp_proto_common_init_xfer_perf(const ucp_proto_common_init_params_t *params,
         }
     } else if (!(params->flags & UCP_PROTO_COMMON_INIT_FLAG_RKEY_PTR)) {
         ucs_assert(reg_md_map == 0);
+
+        if ((params->flags & UCP_PROTO_COMMON_INIT_FLAG_REMOTE_ACCESS) &&
+            (params->send_mem_type != UCS_MEMORY_TYPE_UNKNOWN)) {
+            remote_mem_type = params->send_mem_type;
```
yosefe, line 396: pls don't reuse "remote_mem_type" variable here - it has different meaning. pls use new variable "buffer_mem_type"

src/ucp/rndv/proto_rndv.h
```diff
@@ -68,6 +68,9 @@ typedef struct {
  */
 typedef struct {
     ucp_proto_rndv_ack_priv_t super;
+
+    /* Memory type of fragment buffers which are used by get/mtype and put/mtype
+     * protocols */
```
yosefe, line 73: maybe need another priv struct for get/mtype and put/mtype, since zcopy protocols do not need it?

src/ucp/proto/proto_common.c
brminich: it broke the original fix. The idea was to check for registration ability even if SEND_ZCOPY is not set (to work with mtype protocols). Revert?

src/ucp/rndv/proto_rndv.h
brminich: imo, code will look uglier. Need to remove `ucp_proto_select_add_proto` from `ucp_proto_rndv_get_common_probe` and `ucp_proto_rndv_put_common_probe` which increases code duplication. Or pass more parameters, like `priv`, `priv_size` and `priv_offset_to_bulk` to these `common_probe` routines

src/ucp/rndv/rndv_mtype.inl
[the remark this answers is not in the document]
brminich: what do you mean? There is no such var defined in this func

src/ucp/proto/proto_init.c
```diff
@@ -416,18 +427,15 @@ ucp_proto_init_add_buffer_perf(const ucp_proto_common_init_params_t *params,
 
     /* Receiver has to copy data.
      * Assume same memory type as sender if no rkey */
-    recv_mem_type = (params->super.rkey_config_key == NULL) ?
+    buffer_mem_type = (params->super.rkey_config_key == NULL) ?
```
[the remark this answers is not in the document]
brminich: imo, using one var looks simpler, but up to you - i'll declare the 2d

brminich:
@yosefe, ok to squash?

src/ucp/rndv/rndv_mtype.inl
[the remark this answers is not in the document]
yosefe: right, it was another func

yosefe (approved):

yosefe (approved):

yosefe (approved):

PR #10107, opened 2024-08-28
CI: Improve debug logs

michal-shalev:
WIP! 
Do not review yet!


buildlib/az-helpers.sh
```diff
@@ -247,3 +257,149 @@ get_arch() {
...
+    echo -e "${COLORS[GREEN]}User:${COLORS[NC]} $(whoami)"
+}
+
+function log_machine_info() {
+    echo -e "\n${COLORS[CYAN]}Machine Information${COLORS[NC]}"
+    echo -e "${COLORS[BLUE]}----------------------------${COLORS[NC]}"
+    if [ -f /.dockerenv ]; then
+        echo -e "${COLORS[GREEN]}Running in container:${COLORS[NC]} Yes"
+        echo -e "${COLORS[GREEN]}Agent Machine Name:${COLORS[NC]} $AGENT_MACHINENAME"
+        echo -e "${COLORS[GREEN]}Container Image:${COLORS[NC]} $(awk -F= '/^PRETTY_NAME=/ {gsub(/"/, "", $2); print $2}' /etc/os-release)"
```
ivankochin, line 283: Is that possible to somehow also obtain docker image path here or maybe even show the full command to run the docker manually?

Something like:
```
Container image path: rdmz-harbor.rdmz.labs.mlnx/ucx/centos8-mofed23-cuda11_asan:1
CMD to run container: docker run --rm -ti -v `pwd`:`pwd` -w `pwd` --cap-add=SYS_PTRACE -v /hpc/local:/hpc/local -v /auto/sw_tools:/auto/sw_tools --gpus all --net=host   --ulimit memlock=-1:-1 --device=/dev/infiniband/  rdmz-harbor.rdmz.labs.mlnx/ucx/centos8-mofed23-cuda11_asan:1 /bin/bash

```

buildlib/az-helpers.sh
```diff
@@ -247,3 +257,149 @@ get_arch() {
...
+        log_docker_variables
+    else
+        echo -e "${COLORS[GREEN]}Running in container:${COLORS[NC]} No"
+        echo -e "${COLORS[GREEN]}Host Machine:${COLORS[NC]} $(uname -n)"
+    fi
+    echo -e "${COLORS[GREEN]}OS:${COLORS[NC]} $(uname -s) ($(uname -o))"
+    echo -e "${COLORS[GREEN]}Kernel:${COLORS[NC]} $(uname -r) ($(uname -v))"
+    echo -e "${COLORS[GREEN]}Architecture:${COLORS[NC]} $(uname -m)"
```
ivankochin, line 291: Maybe print full information from lscpu instead of arch only?

Also may be useful to print nvidia-smi and ibv_devinfo (or even lshca) outputs.

buildlib/az-helpers.sh
```diff
@@ -247,3 +257,149 @@ get_arch() {
...
+}
+
+function log_docker_variables() {
+    if env | grep -q 'DOCKER_OPT_'; then
+        echo -e "${COLORS[GREEN]}Docker Variables:${COLORS[NC]}"
+        for var in DOCKER_OPT_VOLUMES DOCKER_OPT_IB DOCKER_OPT_GPU DOCKER_OPT_ARGS; do
```
ivankochin, line 297: Is that all possible docker related env variables? Maybe just echo content from all the values found by grep as you do for UCX variables?

buildlib/az-helpers.sh
```diff
@@ -247,3 +257,149 @@ get_arch() {
...
+    done
+}
+
+function log_loaded_modules() {
+    echo -e "\n${COLORS[CYAN]}Modules Loaded${COLORS[NC]}"
+    echo -e "${COLORS[BLUE]}----------------------------${COLORS[NC]}"
+    if [ ${#loaded_modules[@]} -eq 0 ]; then
+        echo -e "${COLORS[RED]}    No modules loaded.${COLORS[NC]}"
+    else
+        for module in "${loaded_modules[@]}"; do
```
ivankochin, line 360: Is that possible to print output of `module list` here instead of reminding loaded modules to special variable?

ivankochin:
Basically looks like a good addition to our CI but I think that most important part is about how to reproduce the same results in for exact job manually. In perfect world I would like to see list of commands that should be executed to reproduce the CI issue in the same environment manually.

buildlib/az-helpers.sh
michal-shalev: that's exactly what I'm trying to add, I may have to add an environment variable

buildlib/az-helpers.sh
michal-shalev: Yes, you can see it in `main.yml`

buildlib/az-helpers.sh
michal-shalev: No, because those modules are unloaded before the exit log, which is why I added a variable.
Will add a comment to explain it though.

michal-shalev: Added a comment

buildlib/az-helpers.sh
michal-shalev: Added `log_lshca_info()` and `log_ibv_info()` LMKWDYT

buildlib/az-helpers.sh
michal-shalev: Added an environment variable for it

PR #10108, opened 2024-08-28
BUILD: Add libtool -no-suppress flag

michal-shalev:
## What?
Added the `-no-suppress` flag to the build process of libraries to prevent the suppression of duplicate compiler warnings.

## Why?
Currently, some errors aren't visible when the build fails, making debugging difficult. By default, libtool suppresses duplicate warnings when compiling each file twice. The `-no-suppress` flag ensures all warnings are displayed. 
[More details here](https://lists.gnu.org/archive/html/autoconf/2005-02/msg00017.html).

## How?
Defined the `LT_CFLAGS` flag in `configure.ac` and applied it in `Makefile.am` files to disable libtool's warning suppression behavior.


configure.ac
```diff
@@ -85,6 +85,9 @@ LT_LIB_M
 AC_C_RESTRICT
 AC_FUNC_STRERROR_R
 
+LIBTOOL_NO_SUPPRESS="-no-suppress"
```
yosefe, line 88: rename it to LT_CPPFLAGS and define it along with CPPFLAGS in compiler.m4

gleon99: Perhaps `_CFLAGS` (as `CPPFLAGS` are initialized from `CFLAGS` in other places, not the opposite).

michal-shalev: Renamed `LIBTOOL_NO_SUPPRESS` to `LT_CFLAGS`
Moved the definition from `configure.ac` to `compiler.m4`
@yosefe @gleon99

config/m4/compiler.m4
```diff
@@ -598,6 +598,10 @@ ADD_COMPILER_FLAGS_IF_SUPPORTED([[-Wno-pointer-sign],
                                 [AC_LANG_SOURCE([[int main(int argc, char **argv){return 0;}]])])
 
 
+# Prevent libtool from suppression of warnings
+LT_CFLAGS="-no-suppress"
```
yosefe, line 602: move this to line 14, after BASE_CFLAGS

config/m4/compiler.m4
```diff
@@ -598,6 +598,10 @@ ADD_COMPILER_FLAGS_IF_SUPPORTED([[-Wno-pointer-sign],
...
+AC_SUBST(LT_CFLAGS)
```
yosefe, line 603: move this after line 605, `AC_SUBST([BASE_CFLAGS])`

config/m4/compiler.m4
michal-shalev: done

config/m4/compiler.m4
michal-shalev: done

src/tools/perf/cuda/Makefile.am
```diff
@@ -7,8 +7,8 @@
 if HAVE_CUDA
 
 module_LTLIBRARIES               = libucx_perftest_cuda.la
-libucx_perftest_cuda_la_CPPFLAGS = $(BASE_CPPFLAGS) $(CUDA_CPPFLAGS)
-libucx_perftest_cuda_la_CFLAGS   = $(BASE_CFLAGS) $(CUDA_CFLAGS)
+libucx_perftest_cuda_la_CPPFLAGS = $(BASE_CPPFLAGS) $(CUDA_CPPFLAGS) $(LT_CFLAGS)
```
yosefe, line 10: no need to add to libucx_perftest_cuda_la_CPPFLAGS. need to add only to libucx_perftest_cuda_la_CFLAGS
same for all the rest

michal-shalev: Should I add it only to the CFLAGS in general? in all the Makefile.am files? @yosefe

yosefe: yes

michal-shalev: done

yosefe (approved):
pls squash

PR #10112, opened 2024-08-30
UCT/GDRCOPY: Extend ENV variables interface to set platform specific perf estimations

ivankochin:
## What
Add option to set gdrcopy BW and LAT and set them to platform specific value on GH systems.

## Why
To optimize performance estimations on different platforms

### DIFF
--- a/config/ucx.conf
+++ b/config/ucx.conf
@@ -2,6 +2,8 @@
 CPU model=Grace
 UCX_REG_NONBLOCK_MEM_TYPES=host
 UCX_IB_MLX5_DEVX_OBJECTS=
+UCX_GDR_COPY_BW=0MBs,get_dedicated:30GBs,put_dedicated:30GBs
+UCX_GDR_COPY_LAT=30e-9
 UCX_DISTANCE_BW=auto,sys:16500MBs
 
 [Fujitsu ARM]
--- a/src/uct/cuda/gdr_copy/gdr_copy_iface.c
+++ b/src/uct/cuda/gdr_copy/gdr_copy_iface.c
@@ -16,16 +16,35 @@
 #include <ucs/sys/string.h>
 
 
-#define UCT_GDR_COPY_IFACE_DEFAULT_BANDWIDTH (6911.0 * UCS_MBYTE)
-#define UCT_GDR_COPY_IFACE_OVERHEAD          0
-#define UCT_GDR_COPY_IFACE_GET_LATENCY       ucs_linear_func_make(1.4e-6, 0)
-#define UCT_GDR_COPY_IFACE_PUT_LATENCY       ucs_linear_func_make(0.4e-6, 0)
+#define UCT_GDR_COPY_IFACE_OVERHEAD 0
 
 static ucs_config_field_t uct_gdr_copy_iface_config_table[] = {
 
     {"", "", NULL,
      ucs_offsetof(uct_gdr_copy_iface_config_t, super),
      UCS_CONFIG_TYPE_TABLE(uct_iface_config_table)},
+
+    {"BW", "0MBs,get_dedicated:250MBs,put_shared:6911MBs",
+     "Estimated memory copy bandwidth", 0,
+     UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_BW,
+        {"get_dedicated", "dedicated get bandwidth",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, get_bw.dedicated)},
+        {"get_shared", "shared get bandwidth",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, get_bw.shared)},
+        {"put_dedicated", "dedicated put bandwidth",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, put_bw.dedicated)},
+        {"put_shared", "shared put bandwidth",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, put_bw.shared)},
+        {NULL})},
+
+    {"LAT", "get:1.4e-6,put:0.4e-6",
+     "Estimated latency", 0,
+     UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_TIME,
+        {"get", "get latency",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, get_latency)},
+        {"put", "put latency",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, put_latency)},
+        {NULL})},
 
     {NULL}
 };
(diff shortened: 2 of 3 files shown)


src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -27,6 +26,19 @@ static ucs_config_field_t uct_gdr_copy_iface_config_table[] = {
      ucs_offsetof(uct_gdr_copy_iface_config_t, super),
      UCS_CONFIG_TYPE_TABLE(uct_iface_config_table)},
 
+    {"BW", "0MBs,get_dedicated:250MBs,put_shared:6911MBs",
```
yosefe, line 29: why default is 0?

src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -27,6 +26,19 @@ static ucs_config_field_t uct_gdr_copy_iface_config_table[] = {
...
+     "Estimated p2p memory copy bandwidth", 0,
```
yosefe, line 30: how is the term "p2p" related to gdrcopy?

src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -206,6 +213,8 @@ static UCS_CLASS_INIT_FUNC(uct_gdr_copy_iface_t, uct_md_h md, uct_worker_h worke
     }
 
     self->id = ucs_generate_uuid((uintptr_t)self);
+    self->config.get_bw = gdr_config->get_bw;
```
yosefe, line 216: align = on column

src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -193,6 +198,8 @@ static UCS_CLASS_INIT_FUNC(uct_gdr_copy_iface_t, uct_md_h md, uct_worker_h worke
                            const uct_iface_config_t *tl_config)
 {
     ucs_status_t status;
+    uct_gdr_copy_iface_config_t *gdr_config =
```
yosefe, line 201: initialized vars should be 1st

config/ucx.conf
```diff
@@ -2,6 +2,7 @@
 CPU model=Grace
 UCX_REG_NONBLOCK_MEM_TYPES=host
 UCX_IB_MLX5_DEVX_OBJECTS=
+UCX_GDR_COPY_BW=get_dedicated:15GBs,put_dedicated:15GBs
```
yosefe, line 5: is there a scenario where shared bw is not 0 ?

src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -27,6 +26,19 @@ static ucs_config_field_t uct_gdr_copy_iface_config_table[] = {
...
+     "Estimated p2p memory copy bandwidth", 0,
+     UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_BW,
+        {"get_dedicated", "dedicated get bandwidth",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, get_bw.dedicated)},
+        {"get_shared", "shared get bandwidth",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, get_bw.shared)},
+        {"put_dedicated", "dedicated put bandwidth",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, put_bw.dedicated)},
+        {"put_shared", "shared put bandwidth",
+         ucs_offsetof(uct_gdr_copy_iface_config_t, put_bw.shared)},
```
brminich, line 39: i think it should be just put and get. Whether it is shared or dedicated should be set/decided implicitly in the transport

config/ucx.conf
ivankochin: That's the default scenario (see here https://github.com/openucx/ucx/pull/10112/files#diff-d74cb886a55c337f1b0cebb7a424dd4da3e7a88745b3e6282aaf3a1a2bd54a04R5)

ivankochin, line 5:
```suggestion
UCX_GDR_COPY_BW=0MBs,get_dedicated:15GBs,put_dedicated:15GBs
```

src/uct/cuda/gdr_copy/gdr_copy_iface.c
ivankochin: Default is 0 only for values that are not mentioned later. That's how current key-value parser works. I mean that according to the master branch code, `get_shared` and `put_dedicated` are 0, so it is the shortest way to define them to that value.

config/ucx.conf
yosefe: According to `UCX_GDR_COPY_BW=0MBs,get_dedicated:15GBs,put_dedicated:15GBs`, seems shared is 0?

ivankochin: Yes, that's right, shared is 0 and dedicated for both put and get is 15GBs

yosefe: so if shared is always 0 for gdrcopy, maybe we need to have configuration only for dedicated bw?

ivankochin: Shared is 0 for GH systems but it is not zero for common case: https://github.com/openucx/ucx/pull/10112/files#diff-f52d5bd2e1444dd959c6c78c1e2c7db60af22046db17517e03e62e65b9b8ef9fR29

yosefe (dismissed):

brminich (approved):

config/ucx.conf
```diff
@@ -2,6 +2,8 @@
 CPU model=Grace
 UCX_REG_NONBLOCK_MEM_TYPES=host
 UCX_IB_MLX5_DEVX_OBJECTS=
+UCX_GDR_COPY_BW=0MBs,get_dedicated:30GBs,put_dedicated:30GBs
```
SeyedMir, line 5: Is the 30GBs from actual measurements?

ivankochin: Yes, I got results close to 30GBs by measuring `copybw` benchmark with sizes > 1MB.

src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -39,6 +37,15 @@ static ucs_config_field_t uct_gdr_copy_iface_config_table[] = {
          ucs_offsetof(uct_gdr_copy_iface_config_t, put_bw.shared)},
         {NULL})},
 
+    {"LAT", "get:1.4e-6,put:0.4e-6",
```
yosefe, line 40: maybe rename to LATENCY?

brminich (approved):

src/uct/cuda/gdr_copy/gdr_copy_iface.c
ivankochin: I like your idea but I made it to be consistent with existent env variables (see `UCX_ROCM_COPY_LAT` and `UCX_ROCM_IPC_LAT`)

yosefe: @rakhmets @brminich WDYT? what is more consistent with existing env vars?

brminich: we already have `LAT`
ucx_info -c | grep LAT
UCX_DISTANCE_LAT=phb:0.30us,node:0.30us,sys:0.50us
 so, imo `LAT` is ok

yosefe: ok

yosefe (approved):

PR #10113, opened 2024-09-01
TEST/COMPAT: Fix wire compat test failures

yosefe:
## Why
Fix CI failures in wire compat jobs

image

### DIFF
--- a/buildlib/tools/check_tls_perf_caps.sh
+++ b/buildlib/tools/check_tls_perf_caps.sh
@@ -8,28 +8,36 @@
 realdir="$(realpath "$(dirname "$0")")"
 source "${realdir}"/../az-helpers.sh
 
-# check that correct UCX versions are used
-LD_LIBRARY_PATH=${UCX_LEGACY_PATH}/lib "${UCX_LEGACY_PATH}"/bin/ucx_info -v
-LD_LIBRARY_PATH=${UCX_PR_PATH}/lib "${UCX_PR_PATH}"/bin/ucx_info -v
-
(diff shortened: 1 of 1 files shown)


buildlib/tools/check_tls_perf_caps.sh
```diff
@@ -8,28 +8,36 @@
...
 res=true
-for tl_name in $(echo "${old_out}" | grep Transport | awk '{print $3}')
+for tl_name in $(grep Transport ${legacy_info_file} | awk '{print $3}')
 do
-    old_tl_caps=$(echo "$old_out" | grep -A 8 "Transport: $tl_name")
-    new_tl_caps=$(echo "$new_out" | grep -A 8 "Transport: $tl_name")
-    for device in $(echo "${old_tl_caps}" | grep Device | awk '{print $3}')
+    old_tl_caps=$(grep -A 8 "Transport: $tl_name" ${legacy_info_file} || true)
+    new_tl_caps=$(grep -A 8 "Transport: $tl_name" ${pr_info_file}     || true)
+    for device in  $(echo "${old_tl_caps}" | grep Device | awk '{print $3}')
```
brminich, line 33:
```suggestion
    for device in $(echo "${old_tl_caps}" | grep Device | awk '{print $3}')
```

brminich (approved):
👍 besides minor comment

buildlib/tools/check_tls_perf_caps.sh
yosefe: per offline discussion will merge as is to unblock CI

PR #10116, opened 2024-09-02
UCS/CONFIG: Use uint64_t to parse configuration bitmaps

tvegas1:
## What
Align all users to `uint64_t` and validate bit ranges.

We were previously using `uint64_t` for one configuration instead of the then expected `unsigned`.

### DIFF
--- a/src/ucm/api/ucm.h
+++ b/src/ucm/api/ucm.h
@@ -213,7 +213,7 @@ typedef struct ucm_global_config {
     ucm_mmap_hook_mode_t mmap_hook_mode;              /* MMAP hook mode */
     int                  enable_malloc_hooks;         /* Enable installing malloc hooks */
     int                  enable_malloc_reloc;         /* Enable installing malloc relocations */
-    int                  cuda_hook_modes;             /* Bitmap of allowed cuda hooks modes */
+    uint64_t             cuda_hook_modes;             /* Bitmap of allowed cuda hooks modes */
     int                  enable_dynamic_mmap_thresh;  /* Enable adaptive mmap threshold */
     size_t               alloc_alignment;             /* Alignment for memory allocations */
     int                  dlopen_process_rpath;        /* Process RPATH section in dlopen hook */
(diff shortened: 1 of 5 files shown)


brminich:
does it fix some issue, or is it found during code review? Maybe it s worth to always use `uint64_t`?

tvegas1:
This is found while reading code. I guess there is potential issue on big endian machines. Yes, I can migrate all users to `uint64_t`.

src/ucm/api/ucm.h
```diff
@@ -213,7 +213,7 @@ typedef struct ucm_global_config {
     ucm_mmap_hook_mode_t mmap_hook_mode;              /* MMAP hook mode */
     int                  enable_malloc_hooks;         /* Enable installing malloc hooks */
     int                  enable_malloc_reloc;         /* Enable installing malloc relocations */
-    int                  cuda_hook_modes;             /* Bitmap of allowed cuda hooks modes */
+    unsigned             cuda_hook_modes;             /* Bitmap of allowed cuda hooks modes */
```
brminich, line 216: uint64_t?

brminich (commented):
seems also `ucs_global_opts_t->handle_errors` and `ucs_global_opts_t->profile_mode`  are not updated

tvegas1:
fixed, forgot to add before push

src/ucm/api/ucm.h
tvegas1: fixed thx

src/ucs/config/global_opts.h
```diff
@@ -105,7 +105,7 @@ typedef struct {
     size_t                     memtrack_limit;
 
     /* Profiling mode */
-    unsigned                   profile_mode;
+    uint64_t                   profile_mode;
```
brminich, line 108: also `handle_errors` in this struct

tvegas1: thanks, somehow I had changed it but lost some in the meantime.

brminich (approved):

yosefe (approved):

PR #10117, opened 2024-09-02
CI: improve Coverity error reporting

michal-shalev:
## What?
Improves error reporting in buildlib/tools/coverity.sh to make Coverity errors more visible in the Azure DevOps CI pipeline.

## Why?
Currently, identifying the specific Coverity errors causing CI failures is difficult because they are buried in the log output. This change helps to quickly see the errors without scrolling through the logs.

## How?
Modifies the script to capture Coverity errors and output them at the end, making the cause of the failure immediately clear in the CI logs.

## Before
image

## After
image

### DIFF
--- a/buildlib/tools/coverity.sh
+++ b/buildlib/tools/coverity.sh
@@ -104,21 +104,19 @@ run_coverity() {
 	fi
 	cov-analyze --jobs $parallel_jobs $COV_OPT --security --concurrency --dir $cov_build
 	nerrors=$(cov-format-errors --dir $cov_build | awk '/Processing [0-9]+ errors?/ { print $2 }')
-	rc=$(($rc+$nerrors))
 
 	if [ $nerrors -gt 0 ]; then
-		cov-format-errors --dir $cov_build --emacs-style
+		error_log_file="$WORKSPACE/cov_error_log.txt"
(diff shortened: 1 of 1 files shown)


buildlib/tools/coverity.sh
```diff
@@ -135,4 +133,9 @@ else
 	cd "${ucx_build_dir}"
 fi
 
-run_coverity "$mode"
+run_coverity "$mode" || {
+	set +x
+	azure_log_error "Coverity found $nerrors issues: $cov_error_log"
```
yosefe, line 138: saving the errors in a shell variable could be problematic since their length is limited
maybe we can just execute `cov-format-errors --dir $cov_build --emacs-style` at a later stage?

michal-shalev: we can't execute it after unloading the modules so I created an error log file in case the are errors.

ivankochin (approved):

yosefe (dismissed):

michal-shalev:
@yosefe waiting for your reply on https://github.com/openucx/ucx/pull/10145#issuecomment-2404368115

michal-shalev:
I removed the "debug instructions" and squashed @yosefe

yosefe (approved):

PR #10119, opened 2024-09-03
UCP/MM: Set reg_id when memh created

Artemy-Mellanox:
## What
Set memh reg_id when memh is created.

## Why ?
Currently reg_id is reset every ucp_memh_get_slow() call e.g. when there is a pending memory event.


brminich (approved):

brminich:
@Artemy-Mellanox can you pls port to v1.17x?

Issue #10129, opened 2024-09-05
UCX compilation error: 'uct_ib_mlx5_devx_mem_t' has no member named 'dm'


JiakunYan:
I am getting the following error when compiling UCX:

```
     1061    mlx5/dv/ib_mlx5dv_md.c: In function 'uct_ib_mlx5_devx_xgvmi_umem_mr':
  >> 1062    mlx5/dv/ib_mlx5dv_md.c:1930:13: error: 'uct_ib_mlx5_devx_mem_t' has no member named 'dm'
     1063     1930 |     if (memh->dm != NULL) {
     1064          |             ^~
     1065    mlx5/dv/ib_mlx5dv_md.c: In function 'uct_ib_mlx5_devx_reg_exported_key_inner':
  >> 1066    mlx5/dv/ib_mlx5dv_md.c:2014:13: error: 'uct_ib_mlx5_devx_mem_t' has no member named 'dm'
     1067     2014 |     if (memh->dm != NULL) {
     1068          |             ^~
     1069    mlx5/dv/ib_mlx5dv_md.c: In function 'uct_ib_mlx5_devx_md_query':
     1070    mlx5/dv/ib_mlx5dv_md.c:2238:18: warning: unused variable 'md' [-Wunused-variable]
     1071     2238 |     uct_ib_md_t *md = ucs_derived_of(uct_md, uct_ib_md_t);
     1072          |                  ^~
  >> 1073    make[4]: *** [Makefile:1023: mlx5/dv/libuct_ib_la-ib_mlx5dv_md.lo] Error 1
```

After examining the source code, I did find the corresponding lines are not properly guided with `#if HAVE_IBV_DM`.

I am using the following spack spec:
```
ucx+verbs+rc+ud+mlx5_dv+ib_hw_tm+dc+thread_multiple+knem+cma+xpmem+rdmacm
```

JiakunYan:
I can bypass this error with `+dm+ added to my spack spec, but this does seem like a bug to me.

roiedanino:
Hi @JiakunYan,
Can you please provide the UCX version / commit that you are using?
Thanks!

PR #10133, opened 2024-09-08
UCT/IB/MLX5/DV: added missing conditional compilation checks for dm

roiedanino:
## What
Added missing compilation checks for `HAVE_IBV_DM` reported by #10129

## Why ?
To avoid compilation issues when dm is not supported / ucx is configured with `--without-dm` flag

### DIFF
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -269,15 +269,24 @@ static ucs_status_t uct_ib_mlx5_devx_reg_ksm_data_contig(
                                               list_size, mr_p, mkey);
 }
 
+static UCS_F_ALWAYS_INLINE int
+uct_ib_mlx5_devx_has_dm(const uct_ib_mlx5_devx_mem_t *memh)
+{
+    #if HAVE_IBV_DM
+        return memh->dm != NULL;
+    #else
+        return 0;
+    #endif
+}
+
 static void *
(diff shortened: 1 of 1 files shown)


src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2942,7 +2949,9 @@ ucs_status_t uct_ib_mlx5_devx_mem_attach(uct_md_h uct_md,
 ucs_status_t
 uct_ib_mlx5_devx_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
 {
+#if HAVE_IBV_DM
```
yosefe, line 2952: can we use UCS_V_UNUSED instead?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2824,7 +2831,7 @@ uct_ib_mlx5_devx_mkey_pack(uct_md_h uct_md, uct_mem_h uct_memh,
         return UCS_ERR_INVALID_PARAM;
     }
 
-    if ((flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA) || (memh->dm != NULL)) {
+    if ((flags & UCT_MD_MKEY_PACK_FLAG_INVALIDATE_RMA) || uct_ib_mlx5_devx_has_dm(memh)) {
```
yosefe, line 2834: line too long

yosefe (approved):

PR #10134, opened 2024-09-08
UCT/GGA: fix component loading

evgeny-leksikov:
## What
fix GGA component loading

## Why ?
too late GGA component registration (found with `ucx_perftest`)

## How ?
move IB load from query_md_resources to `so` ctor

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -793,14 +793,11 @@ ucs_status_t uct_ib_query_md_resources(uct_component_t *component,
                                        uct_md_resource_desc_t **resources_p,
                                        unsigned *num_resources_p)
 {
-    UCS_MODULE_FRAMEWORK_DECLARE(uct_ib);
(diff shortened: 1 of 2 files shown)


src/uct/ib/base/ib_md.c
```diff
@@ -1517,8 +1514,11 @@ uct_component_t uct_ib_component = {
 
 void UCS_F_CTOR uct_ib_init()
 {
+    UCS_MODULE_FRAMEWORK_DECLARE(uct_ib);
     ssize_t i;
 
+    UCS_MODULE_FRAMEWORK_LOAD(uct_ib, 0);
```
tvegas1, line 1520: I think the load might need to be placed after line 1523 below, to preserve current ordering for component and ib ops.

evgeny-leksikov: fixed

yosefe (approved):

PR #10135, opened 2024-09-08
UCT: check reachability using the routing table

amastbaum:
## What
Add a new reachability check mode for IP-based lanes.

## Why ?
Currently, the reachability of IP-based lanes is only verified by checking if the local and remote addresses share the same subnet. If they do, the remote address is considered reachable.
However, we want to introduce an additional reachability check (and mode) that considers non-local addresses. This will allow us to determine if the remote address is reachable based on routing information, not just subnet matching.

## How ?
When the reachability mode is set to "route," the routing table will be consulted to verify reachability based on the specified interface and destination address.

### DIFF
--- a/src/ucs/Makefile.am
+++ b/src/ucs/Makefile.am
@@ -131,6 +131,7 @@ noinst_HEADERS = \
 	sys/iovec.h \
 	sys/iovec.inl \
 	sys/ptr_arith.h \
+	sys/netlink.h \
 	time/time.h \
 	time/timerq.h \
 	time/timer_wheel.h \
@@ -205,6 +206,7 @@ libucs_la_SOURCES = \
 	sys/sock.c \
 	sys/topo/base/topo.c \
 	sys/stubs.c \
+	sys/netlink.c \
 	sys/uid.c \
 	time/time.c \
 	time/timer_wheel.c \
--- a/src/ucs/sys/netlink.c
+++ b/src/ucs/sys/netlink.c
@@ -1,0 +1,237 @@
+/**
+* Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2025. ALL RIGHTS RESERVED.
+*
+* See file LICENSE for terms.
+*/
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "netlink.h"
+
+#include <ucs/debug/log.h>
+#include <ucs/sys/compiler.h>
+#include <ucs/sys/sock.h>
+#include <ucs/type/status.h>
+#include <ucs/debug/memtrack_int.h>
+
+#include <errno.h>
+#include <linux/netlink.h>
+#include <linux/rtnetlink.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+
+typedef struct {
+    const struct sockaddr *sa_remote;
+    int                    if_index;
+    int                    found;
+} ucs_netlink_route_info_t;
+
+
+/**
+ * Callback function for parsing individual netlink messages.
+ *
+ * @param [in] nlh    Pointer to the netlink message header.
+ * @param [in] nl_msg Pointer to the netlink message payload.
+ * @param [in] arg    User-provided argument passed through from the caller.
+ *
+ * @return UCS_OK if parsing is complete, UCS_INPROGRESS if there are more
+ *         messages to be parsed, or error code otherwise.
+ */
+typedef ucs_status_t (*ucs_netlink_parse_cb_t)(const struct nlmsghdr *nlh,
+                                               const void *nl_msg, void *arg);
+
+static ucs_status_t ucs_netlink_socket_init(int *fd_p, int protocol)
+{
+    struct sockaddr_nl sa = {.nl_family = AF_NETLINK};
+    ucs_status_t status;
+
+    status = ucs_socket_create(AF_NETLINK, SOCK_RAW, protocol, fd_p);
+    if (status != UCS_OK) {
+        ucs_error("failed to create netlink socket: %s",
+                  ucs_status_string(status));
+        goto err;
+    }
+
+    if (bind(*fd_p, (struct sockaddr *)&sa, sizeof(sa)) < 0) {
+        ucs_error("failed to bind netlink socket %d: %m", *fd_p);
+        status = UCS_ERR_IO_ERROR;
+        goto err_close_socket;
+    }
+
+    return UCS_OK;
+
+err_close_socket:
+    ucs_close_fd(fd_p);
+err:
+    return status;
+}
+
+static ucs_status_t
+ucs_netlink_parse_msg(const void *msg, size_t msg_len,
+                      ucs_netlink_parse_cb_t parse_cb, void *arg)
+{
+    ucs_status_t status        = UCS_INPROGRESS;
+    const struct nlmsghdr *nlh = (const struct nlmsghdr *)msg;
+
+    while ((status == UCS_INPROGRESS) && NLMSG_OK(nlh, msg_len) &&
+           (nlh->nlmsg_type != NLMSG_DONE)) {
+        if (nlh->nlmsg_type == NLMSG_ERROR) {
+            struct nlmsgerr *err = (struct nlmsgerr *)NLMSG_DATA(nlh);
+            ucs_error("received error response from netlink err=%d: %s\n",
+                      err->error, strerror(-err->error));
+            return UCS_ERR_IO_ERROR;
+        }
+
+        status = parse_cb(nlh, NLMSG_DATA(nlh), arg);
+        nlh    = NLMSG_NEXT(nlh, msg_len);
+    }
+
+    return UCS_OK;
+}
+
+static ucs_status_t
+ucs_netlink_send_request(int protocol, unsigned short nlmsg_type,
+                         const void *protocol_header, size_t header_length,
+                         ucs_netlink_parse_cb_t parse_cb, void *arg)
+{
+    struct nlmsghdr nlh = {0};
+    char *recv_msg      = NULL;
+    size_t recv_msg_len = 0;
+    int netlink_fd      = -1;
+    ucs_status_t status;
+    struct iovec iov[2];
+    size_t bytes_sent;
+
+    status = ucs_netlink_socket_init(&netlink_fd, protocol);
+    if (status != UCS_OK) {
+        goto out;
+    }
+
+    nlh.nlmsg_len   = NLMSG_LENGTH(header_length);
+    nlh.nlmsg_type  = nlmsg_type;
+    nlh.nlmsg_flags = NLM_F_REQUEST | NLM_F_DUMP;
+    iov[0].iov_base = &nlh;
+    iov[0].iov_len  = sizeof(nlh);
+    iov[1].iov_base = (void *)protocol_header;
+    iov[1].iov_len  = header_length;
+
+    do {
+        status = ucs_socket_sendv_nb(netlink_fd, iov, 2, &bytes_sent);
+    } while (status == UCS_ERR_NO_PROGRESS);
+
+    if (status != UCS_OK) {
+        ucs_error("failed to send netlink message on fd=%d: %s",
+                  netlink_fd, ucs_status_string(status));
+        goto out;
+    }
+
+    /* get message size */
+    status = ucs_socket_recv_nb(netlink_fd, NULL, MSG_PEEK | MSG_TRUNC,
+                                &recv_msg_len);
+    if (status != UCS_OK) {
+        ucs_error("failed to get netlink message size %d (%s)",
+                  status, ucs_status_string(status));
+        goto out;
+    }
+
+    recv_msg = ucs_malloc(recv_msg_len, "netlink recv message");
+    if (recv_msg == NULL) {
+        ucs_error("failed to allocate a buffer for netlink receive message of"
+                  " size %zu", recv_msg_len);
+        goto out;
+    }
+
+    status = ucs_socket_recv(netlink_fd, recv_msg, recv_msg_len);
+    if (status != UCS_OK) {
+        ucs_error("failed to receive netlink message on fd=%d: %s",
+                  netlink_fd, ucs_status_string(status));
+        goto out;
+    }
+
+    status = ucs_netlink_parse_msg(recv_msg, recv_msg_len, parse_cb, arg);
+
+out:
+    ucs_close_fd(&netlink_fd);
+    ucs_free(recv_msg);
+    return status;
+}
+
+static ucs_status_t
+ucs_netlink_get_route_info(const struct rtattr *rta, int len, int *if_index_p,
+                           const void **dst_in_addr)
+{
+    *if_index_p  = -1;
+    *dst_in_addr = NULL;
+
+    for (; RTA_OK(rta, len); rta = RTA_NEXT(rta, len)) {
+        if (rta->rta_type == RTA_OIF) {
+            *if_index_p = *((const int *)RTA_DATA(rta));
+        } else if (rta->rta_type == RTA_DST) {
+            *dst_in_addr = RTA_DATA(rta);
+        }
+    }
+
+    if ((*if_index_p == -1) || (*dst_in_addr == NULL)) {
+        ucs_diag("invalid routing table entry");
+        return UCS_ERR_INVALID_PARAM;
+    }
+
+    return UCS_OK;
+}
+
+static ucs_status_t
+ucs_netlink_parse_rt_entry_cb(const struct nlmsghdr *nlh, const void *nl_msg,
+                              void *arg)
+{
+    ucs_netlink_route_info_t *info = (ucs_netlink_route_info_t *)arg;
+    int rule_iface;
+    const void *dst_in_addr;
+
+    if (ucs_netlink_get_route_info(RTM_RTA((const struct rtmsg *)nl_msg),
+                                   RTM_PAYLOAD(nlh), &rule_iface,
+                                   &dst_in_addr) != UCS_OK) {
+        return UCS_INPROGRESS;
+    }
+
+    if (rule_iface != info->if_index) {
+        return UCS_INPROGRESS;
+    }
+
+    if (ucs_bitwise_is_equal(ucs_sockaddr_get_inet_addr(info->sa_remote),
+                             dst_in_addr,
+                             ((const struct rtmsg *)nl_msg)->rtm_dst_len)) {
+        info->found = 1;
+        return UCS_OK;
+    }
+
+    return UCS_INPROGRESS;
+}
+
+int ucs_netlink_route_exists(const char *if_name,
+                             const struct sockaddr *sa_remote)
+{
+    ucs_netlink_route_info_t info = {0};
+    struct rtmsg rtm              = {0};
+    int iface_index;
+
+    iface_index = if_nametoindex(if_name);
+    if (iface_index == 0) {
+        ucs_error("failed to get interface index (errno %d)", errno);
+        goto out;
+    }
+
+    rtm.rtm_family = sa_remote->sa_family;
+    rtm.rtm_table  = RT_TABLE_MAIN;
+
+    info.if_index  = iface_index;
+    info.sa_remote = sa_remote;
+
+    ucs_netlink_send_request(NETLINK_ROUTE, RTM_GETROUTE, &rtm, sizeof(rtm),
+                             ucs_netlink_parse_rt_entry_cb, &info);
+
+out:
+    return info.found;
+}
--- a/src/ucs/sys/sock.c
+++ b/src/ucs/sys/sock.c
@@ -80,7 +80,7 @@ ucs_status_t ucs_netif_ioctl(const char *if_name, unsigned long request,
 
     ucs_strncpy_zero(if_req->ifr_name, if_name, sizeof(if_req->ifr_name));
 
-    status = ucs_socket_create(AF_INET, SOCK_STREAM, &fd);
+    status = ucs_socket_create(AF_INET, SOCK_STREAM, 0, &fd);
     if (status != UCS_OK) {
         goto out;
     }
@@ -209,9 +209,9 @@ out:
     return ret;
 }
 
-ucs_status_t ucs_socket_create(int domain, int type, int *fd_p)
+ucs_status_t ucs_socket_create(int domain, int type, int protocol, int *fd_p)
 {
-    int fd = socket(domain, type, 0);
+    int fd = socket(domain, type, protocol);
     if (fd < 0) {
         ucs_socket_print_error_info("socket create failed", errno);
         return UCS_ERR_IO_ERROR;
@@ -465,7 +465,7 @@ ucs_status_t ucs_socket_server_init(const struct sockaddr *saddr, socklen_t sock
 
     /* Create the server socket for accepting incoming connections */
     fd     = -1; /* Suppress compiler warning */
-    status = ucs_socket_create(saddr->sa_family, SOCK_STREAM, &fd);
+    status = ucs_socket_create(saddr->sa_family, SOCK_STREAM, 0, &fd);
     if (status != UCS_OK) {
         goto err;
     }
@@ -596,9 +596,9 @@ ucs_socket_handle_io(int fd, const void *data, size_t count,
 
 static inline ucs_status_t
 ucs_socket_do_io_nb(int fd, void *data, size_t *length_p,
-                    ucs_socket_io_func_t io_func, const char *name)
+                    ucs_socket_io_func_t io_func, const char *name, int flags)
 {
-    ssize_t ret = io_func(fd, data, *length_p, MSG_NOSIGNAL);
+    ssize_t ret = io_func(fd, data, *length_p, MSG_NOSIGNAL | flags);
     return ucs_socket_handle_io(fd, data, *length_p, length_p, 0,
                                 ret, errno, name);
 }
@@ -611,7 +611,7 @@ ucs_socket_do_io_b(int fd, void *data, size_t length,
     ucs_status_t status;
 
     do {
-        status = ucs_socket_do_io_nb(fd, data, &cur_cnt, io_func, name);
+        status = ucs_socket_do_io_nb(fd, data, &cur_cnt, io_func, name, 0);
         done_cnt += cur_cnt;
         ucs_assert(done_cnt <= length);
         cur_cnt = length - done_cnt;
@@ -638,7 +638,7 @@ ucs_socket_do_iov_nb(int fd, struct iovec *iov, size_t iov_cnt, size_t *length_p
 ucs_status_t ucs_socket_send_nb(int fd, const void *data, size_t *length_p)
 {
     return ucs_socket_do_io_nb(fd, (void*)data, length_p,
-                               (ucs_socket_io_func_t)send, "send");
+                               (ucs_socket_io_func_t)send, "send", 0);
 }
 
 /* recv is declared as 'always_inline' on some platforms, it leads to
@@ -648,9 +648,10 @@ static ssize_t ucs_socket_recv_io(int fd, void *data, size_t size, int flags)
     return recv(fd, data, size, flags);
 }
 
-ucs_status_t ucs_socket_recv_nb(int fd, void *data, size_t *length_p)
+ucs_status_t ucs_socket_recv_nb(int fd, void *data, int flags, size_t *length_p)
 {
-    return ucs_socket_do_io_nb(fd, data, length_p, ucs_socket_recv_io, "recv");
+    return ucs_socket_do_io_nb(fd, data, length_p, ucs_socket_recv_io,
+                               "recv", flags);
 }
 
 ucs_status_t ucs_socket_send(int fd, const void *data, size_t length)
--- a/src/uct/ib/base/ib_iface.c
+++ b/src/uct/ib/base/ib_iface.c
@@ -23,6 +23,7 @@
 #include <ucs/type/serialize.h>
 #include <ucs/debug/log.h>
 #include <ucs/time/time.h>
+#include <ucs/sys/netlink.h>
 #include <ucs/sys/sock.h>
 #include <string.h>
 #include <stdlib.h>
@@ -40,6 +41,13 @@ const char *uct_ib_mtu_values[] = {
     [UCT_IB_MTU_2048]       = "2048",
     [UCT_IB_MTU_4096]       = "4096",
     [UCT_IB_MTU_LAST]       = NULL
+};
+
+const char *uct_ib_reachability_modes[] = {
+    [UCT_IB_REACHABILITY_MODE_ROUTE]        = "route",
+    [UCT_IB_REACHABILITY_MODE_LOCAL_SUBNET] = "local_subnet",
+    [UCT_IB_REACHABILITY_MODE_ALL]          = "all",
+    [UCT_IB_REACHABILITY_MODE_LAST]         = NULL
 };
 
 enum {
@@ -180,6 +188,13 @@ ucs_config_field_t uct_ib_iface_config_table[] = {
    "address and a netmask in the form x.x.x.x/y.\n"
    "It must not be used together with UCX_IB_GID_INDEX.",
    ucs_offsetof(uct_ib_iface_config_t, rocev2_subnet_filter), UCS_CONFIG_TYPE_ALLOW_LIST},
+
+  {"ROCE_REACHABILITY_MODE", "route",
+   "The mode used for performing the reachability check\n"
+   " - route        - all routable addresses are assumed as reachable\n"
+   " - local_subnet - only addresses within the interface's subnet are assumed as reachable.\n"
+   " - all          - all addresses are assumed as reachable, without any check",
+   ucs_offsetof(uct_ib_iface_config_t, reachability_mode), UCS_CONFIG_TYPE_ENUM(uct_ib_reachability_modes)},
 
   {"ROCE_PATH_FACTOR", "1",
    "Multiplier for RoCE LAG UDP source port calculation. The UDP source port\n"
@@ -620,8 +635,8 @@ ucs_status_t uct_ib_iface_get_device_address(uct_iface_h tl_iface,
     return UCS_OK;
 }
 
-static void uct_ib_iface_log_subnet_info(const struct sockaddr_storage *sa1,
-                                         const struct sockaddr_storage *sa2,
+static void uct_ib_iface_log_subnet_info(const struct sockaddr *sa1,
+                                         const struct sockaddr *sa2,
                                          unsigned prefix_len, int matched)
 {
     UCS_STRING_BUFFER_ONSTACK(info, UCS_SOCKADDR_STRING_LEN * 2 + 60);
@@ -635,84 +650,58 @@ static void uct_ib_iface_log_subnet_info(const struct sockaddr_storage *sa1,
                               " match with a %u-bit prefix, addresses: ",
                               prefix_len);
 
-    ucs_string_buffer_append_saddr(&info, (struct sockaddr*)sa1);
+    ucs_string_buffer_append_saddr(&info, sa1);
     ucs_string_buffer_appendf(&info, " ");
-    ucs_string_buffer_append_saddr(&info, (struct sockaddr*)sa2);
+    ucs_string_buffer_append_saddr(&info, sa2);
     ucs_debug("%s", ucs_string_buffer_cstr(&info));
 }
 
 static int
-uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
-                               const uct_ib_address_t *remote_ib_addr,
-                               unsigned prefix_bits,
-                               const uct_iface_is_reachable_params_t *params)
+uct_ib_iface_roce_is_routable(uct_ib_iface_t *iface, int gid_index,
+                              struct sockaddr *sa_remote,
+                              const uct_iface_is_reachable_params_t *params)
 {
-    sa_family_t local_ib_addr_af         = local_gid_info->roce_info.addr_family;
-    uct_ib_roce_version_t local_roce_ver = local_gid_info->roce_info.ver;
-    uint8_t remote_ib_addr_flags         = remote_ib_addr->flags;
-    struct sockaddr_storage sa_local, sa_remote;
-    uct_ib_roce_version_t remote_roce_ver;
-    sa_family_t remote_ib_addr_af;
-    char local_str[128], remote_str[128];
-    int matched;
+    uct_ib_device_t *dev = uct_ib_iface_device(iface);
+    uint8_t port_num     = iface->config.port_num;
+    char ndev_name[IFNAMSIZ];
+    char remote_str[128];
+    ucs_status_t status;
 
-    /* check for wildcards in the RoCE version (RDMACM or non-RoCE cases) */
-    if ((uct_ib_address_flags_get_roce_version(remote_ib_addr_flags)) ==
-         UCT_IB_DEVICE_ROCE_ANY) {
-        return 1;
+    status = uct_ib_device_get_roce_ndev_name(dev, port_num, gid_index,
+                                              ndev_name, sizeof(ndev_name));
+    if (status != UCS_OK) {
+        uct_iface_fill_info_str_buf(params,
+                                    "couldn't get network interface name");
+        return 0;
     }
 
-    /* check for zero-sized netmask */
+    if (!ucs_netlink_route_exists(ndev_name, sa_remote)) {
+        uct_iface_fill_info_str_buf(params, "remote address %s is not routable",
+                                    ucs_sockaddr_str(sa_remote, remote_str, 128));
+        return 0;
+    }
+
+    return 1;
+}
+
+static int
+uct_ib_iface_roce_is_local_subnet(int prefix_bits,
+                                  const struct sockaddr *sa_local,
+                                  const struct sockaddr *sa_remote,
+                                  const uct_iface_is_reachable_params_t *params)
+{
+    int matched;
+    char local_str[128], remote_str[128];
+
+    /* A 0-bit netmask implies any address is reachable */
     if (prefix_bits == 0) {
         return 1;
     }
 
-    /* check the RoCE version */
-    ucs_assert(local_roce_ver != UCT_IB_DEVICE_ROCE_ANY);
-
-    remote_roce_ver = uct_ib_address_flags_get_roce_version(remote_ib_addr_flags);
-
-    if (local_roce_ver != remote_roce_ver) {
-        uct_iface_fill_info_str_buf(
-                        params,
-                        "different RoCE versions detected. local %s (gid=%s)"
-                        "remote %s (gid=%s)",
-                        uct_ib_roce_version_str(local_roce_ver),
-                        uct_ib_gid_str(&local_gid_info->gid, local_str,
-                                       sizeof(local_str)),
-                        uct_ib_roce_version_str(remote_roce_ver),
-                        uct_ib_gid_str((union ibv_gid*)(remote_ib_addr + 1),
-                                       remote_str, sizeof(remote_str)));
-        return 0;
-    }
-
-    if (local_gid_info->roce_info.ver != UCT_IB_DEVICE_ROCE_V2) {
-        return 1; /* We assume it is, but actually there's no good test */
-    }
-
-    remote_ib_addr_af = uct_ib_address_flags_get_roce_af(remote_ib_addr_flags);
-    if ((uct_ib_device_roce_gid_to_sockaddr(local_ib_addr_af,
-                                            &local_gid_info->gid,
-                                            &sa_local) != UCS_OK)) {
-        uct_iface_fill_info_str_buf(
-               params, "Couldn't convert local RoCE address to socket address");
-        return 0;
-    }
-
-    if (uct_ib_device_roce_gid_to_sockaddr(remote_ib_addr_af,
-                                           remote_ib_addr + 1,
-                                           &sa_remote) != UCS_OK) {
-        uct_iface_fill_info_str_buf(
-               params, "Couldn't convert remote RoCE address to socket address");
-        return 0;
-    }
-
-    matched = ucs_sockaddr_is_same_subnet((struct sockaddr*)&sa_local,
-                                          (struct sockaddr*)&sa_remote,
-                                          prefix_bits);
+    matched = ucs_sockaddr_is_same_subnet(sa_local, sa_remote, prefix_bits);
 
     if (ucs_log_is_enabled(UCS_LOG_LEVEL_DEBUG)) {
-        uct_ib_iface_log_subnet_info(&sa_local, &sa_remote, prefix_bits, matched);
+        uct_ib_iface_log_subnet_info(sa_local, sa_remote, prefix_bits, matched);
     }
 
     if (!matched) {
@@ -720,14 +709,94 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
                     params,
                     "IP addresses do not match with a %u-bit prefix. local IP"
                     " is %s, remote IP is %s",
-                    prefix_bits,
-                    ucs_sockaddr_str((struct sockaddr *)&sa_local,
-                                     local_str, 128),
-                    ucs_sockaddr_str((struct sockaddr *)&sa_remote,
-                                     remote_str, 128));
+                    prefix_bits, ucs_sockaddr_str(sa_local, local_str, 128),
+                    ucs_sockaddr_str(sa_remote, remote_str, 128));
     }
 
     return matched;
+}
+
+static int
+uct_ib_iface_roce_is_reachable(uct_ib_iface_t *iface,
+                               const uct_ib_address_t *remote_ib_addr,
+                               const uct_iface_is_reachable_params_t *params)
+{
+    uct_ib_device_gid_info_t local_gid_info = iface->gid_info;
+    sa_family_t local_ib_addr_af            = local_gid_info.roce_info.addr_family;
+    uct_ib_roce_version_t local_roce_ver    = local_gid_info.roce_info.ver;
+    uint8_t remote_ib_addr_flags            = remote_ib_addr->flags;
+    struct sockaddr_storage sa_local, sa_remote;
+    uct_ib_roce_version_t remote_roce_ver;
+    sa_family_t remote_ib_addr_af;
+    char local_str[128], remote_str[128];
+
+    remote_roce_ver = uct_ib_address_flags_get_roce_version(remote_ib_addr_flags);
+
+    /* check for wildcards in the RoCE version (RDMACM or non-RoCE cases)
+       or for reachability mode 'all' (no reachibility check is needed) */
+    if ((remote_roce_ver == UCT_IB_DEVICE_ROCE_ANY) ||
+        (iface->config.reachability_mode == UCT_IB_REACHABILITY_MODE_ALL)) {
+        return 1;
+    }
+
+    /* check the RoCE version */
+    ucs_assert(local_roce_ver != UCT_IB_DEVICE_ROCE_ANY);
+
+    if (local_roce_ver != remote_roce_ver) {
+        uct_iface_fill_info_str_buf(
+                        params,
+                        "different RoCE versions detected. local %s (gid=%s)"
+                        "remote %s (gid=%s)",
+                        uct_ib_roce_version_str(local_roce_ver),
+                        uct_ib_gid_str(&local_gid_info.gid, local_str,
+                                       sizeof(local_str)),
+                        uct_ib_roce_version_str(remote_roce_ver),
+                        uct_ib_gid_str((union ibv_gid*)(remote_ib_addr + 1),
+                                       remote_str, sizeof(remote_str)));
+        return 0;
+    }
+
+    if (local_roce_ver != UCT_IB_DEVICE_ROCE_V2) {
+        return 1; /* We assume it is, but actually there's no good test */
+    }
+
+    remote_ib_addr_af = uct_ib_address_flags_get_roce_af(remote_ib_addr_flags);
+    if (local_ib_addr_af != remote_ib_addr_af) {
+        uct_iface_fill_info_str_buf(
+                    params, "different IP versions, local %s vs remote %s\n",
+                    local_ib_addr_af == AF_INET ? "IPv4": "IPv6",
+                    remote_ib_addr_af == AF_INET ? "IPv4": "IPv6");
+        return 0;
+    }
+
+    if ((uct_ib_device_roce_gid_to_sockaddr(local_ib_addr_af,
+                                            &local_gid_info.gid,
+                                            &sa_local) != UCS_OK)) {
+        uct_iface_fill_info_str_buf(
+               params, "couldn't convert local RoCE address to socket address");
+        return 0;
+    }
+
+    if (uct_ib_device_roce_gid_to_sockaddr(remote_ib_addr_af,
+                                           remote_ib_addr + 1,
+                                           &sa_remote) != UCS_OK) {
+        uct_iface_fill_info_str_buf(
+               params, "couldn't convert remote RoCE address to socket address");
+        return 0;
+    }
+
+    if (iface->config.reachability_mode == UCT_IB_REACHABILITY_MODE_ROUTE) {
+        return uct_ib_iface_roce_is_routable(iface, local_gid_info.gid_index,
+                                             (struct sockaddr *)&sa_remote,
+                                             params);
+    }
+
+    ucs_assert(iface->config.reachability_mode ==
+               UCT_IB_REACHABILITY_MODE_LOCAL_SUBNET);
+    return uct_ib_iface_roce_is_local_subnet(iface->addr_prefix_bits,
+                                             (const struct sockaddr *)&sa_local,
+                                             (const struct sockaddr *)&sa_remote,
+                                             params);
 }
 
 int uct_ib_iface_is_same_device(const uct_ib_address_t *ib_addr, uint16_t dlid,
@@ -827,8 +896,7 @@ static int uct_ib_iface_dev_addr_is_reachable(
         /* there shouldn't be a lid and the UCT_IB_ADDRESS_FLAG_LINK_LAYER_ETH
          * flag should be on. If reachable, the remote and local RoCE versions
          * and address families have to be the same */
-        return uct_ib_iface_roce_is_reachable(&iface->gid_info, ib_addr,
-                                              iface->addr_prefix_bits,
+        return uct_ib_iface_roce_is_reachable(iface, ib_addr,
                                               is_reachable_params);
     } else {
         /* local and remote have different link layers and therefore are unreachable */
@@ -1321,8 +1389,14 @@ uct_ib_iface_init_roce_addr_prefix(uct_ib_iface_t *iface,
 
     ucs_assert(uct_ib_iface_is_roce(iface));
 
+    /* Override the original value of reachability_mode if
+       rocev2_local_subnet is enabled to maintain backward compatibility */
+    iface->config.reachability_mode = config->rocev2_local_subnet ?
+                                      UCT_IB_REACHABILITY_MODE_LOCAL_SUBNET :
+                                      config->reachability_mode;
+
     if ((gid_info->roce_info.ver != UCT_IB_DEVICE_ROCE_V2) ||
-        !config->rocev2_local_subnet) {
+        (iface->config.reachability_mode != UCT_IB_REACHABILITY_MODE_LOCAL_SUBNET)) {
         iface->addr_prefix_bits = 0;
         return UCS_OK;
     }
--- a/src/uct/ib/base/ib_iface.h
+++ b/src/uct/ib/base/ib_iface.h
@@ -107,6 +107,18 @@ enum {
     UCT_IB_ADDRESS_PACK_FLAG_PKEY          = UCS_BIT(5)
 };
 
+
+/**
+ * Reachability mode.
+ */
+typedef enum {
+    UCT_IB_REACHABILITY_MODE_ROUTE        = 0,
+    UCT_IB_REACHABILITY_MODE_LOCAL_SUBNET = 1,
+    UCT_IB_REACHABILITY_MODE_ALL          = 2,
+    UCT_IB_REACHABILITY_MODE_LAST
+} uct_ib_iface_reachability_mode_t;
+
+
 enum {
     UCT_IB_IFACE_STAT_RX_COMPLETION,
     UCT_IB_IFACE_STAT_TX_COMPLETION,
@@ -175,60 +187,63 @@ struct uct_ib_iface_config {
     } rx;
 
     /* Inline space to reserve in CQ */
-    size_t                  inl[UCT_IB_DIR_LAST];
+    size_t                           inl[UCT_IB_DIR_LAST];
 
     /* Change the address type */
-    int                     addr_type;
+    int                              addr_type;
 
     /* Force global routing */
-    int                     is_global;
+    int                              is_global;
 
     /* Use FLID based routing */
-    int                     flid_enabled;
+    int                              flid_enabled;
 
     /* IB SL to use (default: AUTO) */
-    unsigned long           sl;
+    unsigned long                    sl;
 
     /* IB Traffic Class to use */
-    unsigned long           traffic_class;
+    unsigned long                    traffic_class;
 
     /* IB hop limit / TTL */
-    unsigned                hop_limit;
+    unsigned                         hop_limit;
 
     /* Number of paths to expose for the interface  */
-    unsigned long           num_paths;
+    unsigned long                    num_paths;
 
     /* Whether to check RoCEv2 reachability by IP address and local subnet */
-    int                     rocev2_local_subnet;
+    int                              rocev2_local_subnet;
 
     /* Length of subnet prefix for reachability check */
-    unsigned long           rocev2_subnet_pfx_len;
+    unsigned long                    rocev2_subnet_pfx_len;
+
+    /* Mode used for performing reachability check */
+    uct_ib_iface_reachability_mode_t reachability_mode;
 
     /* List of included/excluded subnets to filter RoCE GID entries by */
-    ucs_config_allow_list_t rocev2_subnet_filter;
+    ucs_config_allow_list_t          rocev2_subnet_filter;
 
     /* Multiplier for RoCE LAG UDP source port calculation */
-    unsigned                roce_path_factor;
+    unsigned                         roce_path_factor;
 
     /* Ranges of path bits */
     UCS_CONFIG_ARRAY_FIELD(ucs_range_spec_t, ranges) lid_path_bits;
 
     /* IB PKEY to use */
-    unsigned                     pkey;
+    unsigned                         pkey;
 
     /* Path MTU size */
-    uct_ib_mtu_t                 path_mtu;
+    uct_ib_mtu_t                     path_mtu;
 
     /* QP counter set ID */
-    unsigned long                counter_set_id;
+    unsigned long                    counter_set_id;
 
     /* IB reverse SL (default: AUTO - same value as sl) */
-    unsigned long                reverse_sl;
+    unsigned long                    reverse_sl;
 
     /**
      * Estimated overhead of preparing a work request and posting it to the NIC
      */
-    uct_ib_iface_send_overhead_t send_overhead;
+    uct_ib_iface_send_overhead_t     send_overhead;
 };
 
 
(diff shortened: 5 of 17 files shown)


src/ucs/Makefile.am
```diff
@@ -206,6 +207,7 @@ libucs_la_SOURCES = \
 	sys/topo/base/topo.c \
 	sys/stubs.c \
 	sys/uid.c \
+	sys/ucs_netlink.c \
```
shasson5, line 210: move this line before `sys/uid.c`

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,119 @@
+#ifdef HAVE_CONFIG_H
```
shasson5, line 1: add copyright

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,119 @@
...
+#include <stdlib.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <errno.h>
+#include <arpa/inet.h>
+#include <net/if.h>
+#include <linux/netlink.h>
+#include <ucs/type/status.h>
+#include <ucs/debug/log.h>
+
```
shasson5, line 17: add spaceline

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,119 @@
...
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <errno.h>
+#include <arpa/inet.h>
+#include <net/if.h>
+#include <linux/netlink.h>
+#include <ucs/type/status.h>
+#include <ucs/debug/log.h>
```
shasson5, line 16: 1) order alphabeticly
2) check if some includes can be removed
3) same in other files

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,119 @@
...
+
+ucs_status_t netlink_socket_create(struct netlink_socket *nl_sock, int protocol)
+{
+    int fd;
```
shasson5, line 20: all function names should start with: `ucs_netlink`

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,45 @@
...
+    UCS_NL_STATUS_OK = 0,
+    UCS_NL_STATUS_DONE = 1,
+    UCS_NL_STATUS_ERROR = 2,
+} ucs_nl_parse_status_t;
+
+struct netlink_socket {
+    int fd;
+    struct sockaddr_nl local;
+    struct sockaddr_nl peer;
+};
```
shasson5, line 20: no need for this struct (only `fd` is required across functions)

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,119 @@
...
+
+    fd = socket(AF_NETLINK, SOCK_RAW, protocol);
+    if (fd < 0) {
+        ucs_diag("failed to create a socket\n");
+        return UCS_ERR_IO_ERROR;
+    }
```
shasson5, line 26: use `ucs_socket_create` instead.

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,119 @@
...
+    return UCS_OK;
+}
+
+void netlink_socket_close(struct netlink_socket *nl_sock)
+{
+    if (nl_sock->fd >= 0) {
+        close(nl_sock->fd);
+        nl_sock->fd = -1;
+    }
+}
```
shasson5, line 48: use `ucs_close_fd` instead.

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,119 @@
...
+                          struct netlink_message *msg)
+{
+    struct nlmsghdr *nlh = (struct nlmsghdr *)msg->buf;
+
+    /* send the request */
+    if (send(nl_sock->fd, nlh, nlh->nlmsg_len, 0) < 0) {
+        ucs_diag("failed to send netlink message\n");
+        close(nl_sock->fd);
+        return UCS_ERR_IO_ERROR;
+    }
```
shasson5, line 74: use `ucs_socket_send`

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,119 @@
...
+
+    memset(&msg->buf, 0, sizeof(msg->buf));
+    ret = recv(nl_sock->fd, &msg->buf, sizeof(msg->buf), 0);
+    if (ret < 0) {
+        return -errno;
+    }
+
+    if (ret == 0) {
+        return -ENODATA;
+    }
```
shasson5, line 91: use `ucs_socket_recv`

src/ucs/sys/ucs_netlink.c
amastbaum: The problem is that I want ucs_netlink user to be able to specify the netlink protocol to use (NETLINK_ROUTE, NETLINK_FIREWALL, NETLINK_RDMA, etc.). Specifically in its current use it doesn't matter because NETLINK_ROUTE is equal to 0 (which is the default value defined in `ucs_socket_create`) but in other cases I want to enable it..

shasson5: right, so need to add `protocol` argument to `ucs_socket_create`

gleon99:
@amastbaum 
as discussed f2f, let's first understand the recv / buffer len, then consider whether we need to change anything in UCS sock.

src/uct/ib/base/ib_iface.h
```diff
@@ -107,6 +107,18 @@ enum {
     UCT_IB_ADDRESS_PACK_FLAG_PKEY          = UCS_BIT(5)
 };
 
+
+/**
+ * Reachability mode.
+ */
+typedef enum uct_ib_reachability_mode {
```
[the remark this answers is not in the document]
amastbaum: @iyastreb What do you mean?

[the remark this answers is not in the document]
iyastreb: since you have a typedef, the enum name is redundant:
```C
typedef enum {
} uct_ib_reachability_mode_t;
```
Then you access it everywhere by typedef, not by `enum uct_ib_reachability_mode`, so the enum name is actually redundant.

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1038,168 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+
+static void parse_nl_route_cb(struct nlmsghdr *nlh, void *arg)
+{
+    struct route_info *info = (struct route_info*)arg;
+    struct rtmsg *rtm       = NLMSG_DATA(nlh);
+    struct rtattr *rta[RTA_MAX + 1];
+    int *oif;
+
+    rtm->rtm_family = info->family;
+    rtm->rtm_table = RT_TABLE_MAIN;
```
shasson5, line 1092: seems it can be removed

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1038,168 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+
+    if (rtm->rtm_family != info->family) {
+        return;
+    }
+
+    netlink_parse_rtattr(rta, RTA_MAX, RTM_RTA(rtm), RTM_PAYLOAD(nlh));
+
```
shasson5, line 1099: change this function to `ucs_netlink_get_route_info` and return only relevant attributes (iface index + dst address)

src/ucs/sys/ucs_netlink.c
shasson5: unresolved

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,134 @@
...
+    for (nlh = (struct nlmsghdr*)msg->buf; NLMSG_OK(nlh, msg_len);
+         nlh = NLMSG_NEXT(nlh, msg_len)) {
+        if (nlh->nlmsg_type == NLMSG_DONE) {
+            return UCS_NL_STATUS_DONE;
+        }
+
+        if (nlh->nlmsg_type == NLMSG_ERROR) {
+            struct nlmsgerr *err = (struct nlmsgerr*)NLMSG_DATA(nlh);
+            ucs_diag("failed to parse netlink message header (%d)", err->error);
+            return UCS_NL_STATUS_ERROR;
```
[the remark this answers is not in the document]
amastbaum: I first wanted to support the parsing of both bulk messages and multiple message headers so I wanted to distinguish between their return values..

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1038,168 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+    for (; RTA_OK(rta, len); rta = RTA_NEXT(rta, len)) {
+        if (rta->rta_type <= max) {
+            attrs[rta->rta_type] = rta;
+        }
+    }
+
+    return 0;
+}
+
+static void create_ipv6_mask(struct in6_addr *mask, int prefix_len)
```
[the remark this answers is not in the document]
iyastreb: thanks, `uint8_t` then

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,130 @@
...
+ *         messages may follow,
+ *         UCS_NL_STATUS_DONE if this was the last message in a multi-part
+ *         message sequence, or UCS_NL_STATUS_ERROR otherwise.
+ */
+ucs_nl_parse_status_t
+ucs_netlink_parse_msg(struct netlink_message *msg, size_t msg_len,
+                      void (*parse_cb)(struct nlmsghdr *h, void *arg),
+                      void *arg);
+
+#endif // UCS_NETLINK_H
```
iyastreb, line 130: endline

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1038,168 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+    }
+
+    ucs_netlink_send_msg_create(&msg, RTM_GETROUTE, NLM_F_REQUEST | NLM_F_DUMP,
+                                sizeof(struct rtmsg));
+
+    rtm             = (struct rtmsg*)NLMSG_DATA(msg.buf);
+    rtm->rtm_family = info.family;
+    rtm->rtm_table  = RT_TABLE_MAIN;
+
+    ret = ucs_netlink_send(&nl_sock, &msg);
```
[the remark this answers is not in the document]
amastbaum: @shasson5 I mentioned it yesterday but have a problem initializing `struct rtmsg`'s fields. What do you think?

src/ucs/sys/sock.c
```diff
@@ -582,28 +582,49 @@ ucs_socket_handle_io(int fd, const void *data, size_t count,
...
 ucs_socket_do_io_nb(int fd, void *data, size_t *length_p,
-                    ucs_socket_io_func_t io_func, const char *name)
+                    ucs_socket_io_func_t io_func, const char *name, int flags)
 {
-    ssize_t ret = io_func(fd, data, *length_p, MSG_NOSIGNAL);
+    ssize_t ret;
+
+    if (flags == 0) {
+        flags = MSG_NOSIGNAL;
+    }
```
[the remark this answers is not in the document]
amastbaum: I put it under a condition by mistake. This flag was originally set when no other flags were supported in `ucs_socket_do_io_b`. I am changing it to `ret = io_func(fd, data, *length_p, flags | MSG_NOSIGNAL);`

src/ucs/sys/sock.c
```diff
@@ -582,28 +582,49 @@ ucs_socket_handle_io(int fd, const void *data, size_t count,
...
     do {
-        status = ucs_socket_do_io_nb(fd, data, &cur_cnt, io_func, name);
+        status = ucs_socket_do_io_nb(fd, data, &cur_cnt, io_func, name, flags);
         done_cnt += cur_cnt;
-        ucs_assert(done_cnt <= length);
-        cur_cnt = length - done_cnt;
-    } while ((done_cnt < length) &&
+        if ((flags & MSG_TRUNC) == 0) {
+            ucs_assert(done_cnt <= *length);
+        }
```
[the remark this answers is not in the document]
amastbaum: But I guess as long as it supports receiving flags we should handle MSG_TRUNC, otherwise it crashes for `done_cnt <= *length`.
Another option is to not support flags at all

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,74 @@
...
+#ifdef HAVE_CONFIG_H
+#  include "config.h"
+#endif
+
+#include "ucs_netlink.h"
+
+#include <ucs/sys/sock.h>
+#include <ucs/type/status.h>
+#include <ucs/debug/log.h>
+#include <ucs/debug/memtrack_int.h>
```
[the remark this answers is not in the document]
amastbaum: I need it for `ucs_malloc`

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,105 @@
...
+    return UCS_OK;
+}
+
+void ucs_netlink_socket_close(struct netlink_socket *nl_sock)
+{
+    if (nl_sock->fd >= 0) {
+        ucs_close_fd(&nl_sock->fd);
+        nl_sock->fd = -1;
+    }
+}
```
shasson5, line 55: can remove this function and instead use `ucs_close_fd` directly

src/ucs/sys/ucs_netlink.c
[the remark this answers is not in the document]
shasson5: can use `ucs_alloca` instead

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,80 @@
...
+                    ((struct nlmsgerr*)NLMSG_DATA(nlh))->error); \
+            action; \
+        } \
+    } while (0)
+
+
+struct netlink_socket {
+    int fd;
+};
+
```
shasson5, line 35: can remove and instead use `fd` directly

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,80 @@
...
+ucs_status_t
+ucs_netlink_socket_create(struct netlink_socket *nl_sock, int protocol);
+
+
+/**
+ * Creates a netlink socket.
+ *
+ * @param [in]  nl_sock   Pointer to the netlink socket to be closed.
+ */
+void ucs_netlink_socket_close(struct netlink_socket *nl_sock);
```
shasson5, line 55: these functions are not API anymore

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,88 @@
...
+
+    nlh              = (struct nlmsghdr *)send_msg;
+    nlh->nlmsg_len   = NLMSG_LENGTH(nl_protocol_hdr_size);
+    nlh->nlmsg_type  = nlmsg_type;
+    nlh->nlmsg_flags = NLM_F_REQUEST | NLM_F_DUMP;
+    nlh->nlmsg_seq   = 1;
+    nlh->nlmsg_pid   = getpid();
+    memcpy(NLMSG_DATA(send_msg), nl_protocol_hdr, nl_protocol_hdr_size);
+
+    ret = ucs_socket_send(fd, nlh, nlh->nlmsg_len);
```
shasson5, line 73: replace with io vector as we discussed

amastbaum: @shasson5 For some reason didn't upload the newest changes yesterday..

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,96 @@
...
+        ucs_diag("failed to open netlink socket");
+        return ret;
+    }
+
+    memset(iov, 0, sizeof(iov));
+    nlh.nlmsg_len   = send_msg_len;
+    nlh.nlmsg_type  = nlmsg_type;
+    nlh.nlmsg_flags = NLM_F_REQUEST | NLM_F_DUMP;
+    nlh.nlmsg_seq   = 1;
+    nlh.nlmsg_pid   = getpid();
```
shasson5, line 69: 1) `nlh.nlmsg_pid = 0`
2) better to init the whole struct and then init specific fields: `struct nlmsghdr nlh = {0}`

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,96 @@
...
+
+    return UCS_OK;
+}
+
+ucs_status_t
+ucs_netlink_send_cmd(int protocol, void *nl_protocol_hdr,
+                     size_t nl_protocol_hdr_size, char *recv_msg_buf,
+                     size_t *recv_msg_buf_len, unsigned short nlmsg_type)
+{
+    ucs_status_t ret;
```
shasson5, line 52: `ucs_status_t status;`

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,96 @@
...
+        ucs_close_fd(fd);
+        ucs_diag("failed to bind netlink socket %d", *fd);
+        return UCS_ERR_IO_ERROR;
+    }
+
+    return UCS_OK;
+}
+
+ucs_status_t
+ucs_netlink_send_cmd(int protocol, void *nl_protocol_hdr,
```
shasson5, line 48: add `const` where possible

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,96 @@
...
+    size_t send_msg_len = NLMSG_LENGTH(nl_protocol_hdr_size);
+
+    ret = ucs_netlink_socket_create(&fd, NETLINK_ROUTE);
+    if (ret != UCS_OK) {
+        ucs_diag("failed to open netlink socket");
+        return ret;
+    }
+
+    memset(iov, 0, sizeof(iov));
+    nlh.nlmsg_len   = send_msg_len;
```
shasson5, line 65: `nlh.nlmsg_len = NLMSG_LENGTH(nl_protocol_hdr_size);`

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,96 @@
...
+    nlh.nlmsg_flags = NLM_F_REQUEST | NLM_F_DUMP;
+    nlh.nlmsg_seq   = 1;
+    nlh.nlmsg_pid   = getpid();
+    iov[0].iov_base = &nlh;
+    iov[0].iov_len  = sizeof(nlh);
+    iov[1].iov_base = nl_protocol_hdr;
+    iov[1].iov_len  = nl_protocol_hdr_size;
+
+    do {
+        ret = ucs_socket_sendv_nb(fd, iov, 2, &send_msg_len);
```
shasson5, line 76: `send_msg_len` => `dummy`

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,54 @@
...
+         elem = NLMSG_NEXT(elem, len))
+
+#define ucs_netlink_handle_parse_error(nlh, action) \
+    do { \
+        if (nlh->nlmsg_type == NLMSG_ERROR) { \
+            ucs_diag("failed to parse netlink message header (%d)", \
+                    ((struct nlmsgerr*)NLMSG_DATA(nlh))->error); \
+            action; \
+        } \
+    } while (0)
```
shasson5, line 29: maybe check errors inside `ucs_netlink_send_cmd`?

src/ucs/sys/ucs_netlink.c
amastbaum: `iovec::iov_base` is of type `void *` so can't use const. `recv_msg_buf` cannot be const as well

src/ucs/sys/ucs_netlink.c
amastbaum: I wanted to use another variable so I could send its pointer to `ucs_socket_sendv_nb`. I know can technically send `&nlh.nlmsg_len` but for some reason it feels unsafe.. is there a reason for that or should I use `&nlh.nlmsg_len`?

src/ucs/sys/ucs_netlink.h
amastbaum: @shasson5  Here I check errors for each entry in the routing table so it has to be called from uct_iface.c

shasson5: right but we can add another "pre" iteration over the array, inside netlink infrastructure (because we only have to deal with MSG_ERROR header and not with rtmsg header).

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,54 @@
...
+#ifndef UCS_NETLINK_H
+#define UCS_NETLINK_H
+
+#include <ucs/type/status.h>
+
+#include <stddef.h>
+#include <linux/netlink.h>
+
+
+#define ucs_netlink_foreach(elem, msg, len) \
```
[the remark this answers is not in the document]
gleon99: Regarding netlink/rt separation - as we discussed together with Shachar.

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,54 @@
...
+    for (elem = (struct nlmsghdr *)msg; \
+         NLMSG_OK(elem, len) && (elem->nlmsg_type != NLMSG_DONE) && \
+         (elem->nlmsg_type != NLMSG_ERROR); \
+         elem = NLMSG_NEXT(elem, len))
+
+#define ucs_netlink_handle_parse_error(nlh, action) \
```
gleon99, line 22: `handle_parse` weird name. Maybe only "handle"

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,54 @@
...
+
+
+/**
+ * Sends and receives a netlink message using a user allocated buffer.
+ *
+ * @param [in]    protocol         The communication protocol to be used
+ *                                 (NETLINK_ROUTE, NETLINK_NETFILTER, etc.).
+ * @param [in]    nl_protocol_hdr  A struct that holds nl protocol specific
+ *                                 details and is placed in nlmsghdr.
+ * @param [in]    nl_protocol_hdr_size Protocol struct size.
```
gleon99, line 39: Alignment

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @return UCS_OK if received successfully, or error code otherwise.
+ */
+ucs_status_t
+ucs_netlink_send_cmd(int protocol, void *nl_protocol_hdr,
+                     size_t nl_protocol_hdr_size, char *recv_msg_buf,
+                     size_t *recv_msg_buf_len, unsigned short nlmsg_type);
+
+
+#endif // UCS_NETLINK_H
```
gleon99, line 54: `//` ?

src/uct/base/uct_iface.c
```diff
@@ -18,10 +18,11 @@
 #include <uct/api/v2/uct_v2.h>
 #include <ucs/async/async.h>
 #include <ucs/sys/string.h>
+#include <ucs/sys/ucs_netlink.h>
 #include <ucs/time/time.h>
 #include <ucs/debug/debug_int.h>
 #include <ucs/vfs/base/vfs_obj.h>
-
+#include <linux/rtnetlink.h>
```
gleon99, line 25: Order

src/uct/base/uct_iface.c
```diff
@@ -81,6 +82,8 @@ static ucs_stats_class_t uct_iface_stats_class = {
 };
 #endif
 
+#define NETLINK_MESSAGE_MAX_SIZE 8195
```
gleon99, line 85: Where does it come from? Doc plz

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1037,163 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+    union {
+        struct in_addr  ipv4;
+        struct in6_addr ipv6;
+    } remote_addr;
+
+    int prefix_len;
+    int reachable;
+};
+
+static int netlink_get_route_info(int **if_idx, void **dst_in_addr,
```
gleon99, line 1053: IMO `uct_iface.c` is not the place for it (in fact, it has nothing UCX specific).
Anyway, after moving, don't forget to prefix func names with module name.

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1037,163 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+                                  struct rtattr *rta, int len)
+{
+    *if_idx = NULL;
+    *dst_in_addr = NULL;
+
+    for (; RTA_OK(rta, len); rta = RTA_NEXT(rta, len)) {
+        if (rta->rta_type == RTA_OIF) {
+            *if_idx = RTA_DATA(rta);
+        }
```
gleon99, line 1062: `} else if {`

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1037,163 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+    } remote_addr;
+
+    int prefix_len;
+    int reachable;
+};
+
+static int netlink_get_route_info(int **if_idx, void **dst_in_addr,
+                                  struct rtattr *rta, int len)
+{
+    *if_idx = NULL;
```
gleon99, line 1056: Alignment

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1037,163 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+    for (; RTA_OK(rta, len); rta = RTA_NEXT(rta, len)) {
+        if (rta->rta_type == RTA_OIF) {
+            *if_idx = RTA_DATA(rta);
+        }
+        else if (rta->rta_type == RTA_DST) {
+            *dst_in_addr = RTA_DATA(rta);
+        }
+    }
+
+    return 0;
```
gleon99, line 1068: Why not void?

gleon99: Maybe better make it return UCS status, then don't need to check for NULL outside.

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1037,163 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+}
+
+static void create_ipv6_mask(struct in6_addr *mask, unsigned char prefix_len)
+{
+    int i;
```
gleon99, line 1073: Space

src/uct/base/uct_iface.c
amastbaum: @gleon99 But then I need to check the status outside of the function anyway, and I am also not sure if it's an error more than just an entry with missing properties. Which error status would you return then?

src/uct/ib/base/ib_iface.h
```diff
@@ -115,6 +127,7 @@ enum {
     UCT_IB_IFACE_STAT_LAST
 };
 
+
```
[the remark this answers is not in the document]
amastbaum: @shasson5 There are 2 space lines after this enum, as should be between definitions in header files, am I wrong?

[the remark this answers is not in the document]
shasson5: that's right, but in general it's better not to fix issues which are not directly related to the PR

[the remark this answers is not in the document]
amastbaum: Oh, I see

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,255 @@
...
+    status = ucs_netlink_parse_msg(recv_msg, recv_msg_len,
+                                   ucs_netlink_parse_rt_entry_cb, &info);
+    if (status != UCS_OK) {
+        ucs_error("failed to parse netlink route message (%s)",
+                  ucs_status_string(status));
+        goto out;
+    }
+
+out:
+    if (recv_msg != NULL) {
```
shasson5, line 250: null check is not required

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    return UCS_INPROGRESS;
+}
+
+int ucs_netlink_rule_exists(const char *iface,
+                            const struct sockaddr *sa_remote)
+{
+    char *recv_msg         = NULL;
+    struct route_info info = {0};
```
shasson5, line 208: `struct route_info` => `struct ucs_netlink_route_info_t`

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,255 @@
...
+static int ucs_rtnetlink_is_rule_matching(struct rtmsg *rtm, size_t rtm_len,
+                                          const struct sockaddr *sa_remote,
+                                          int iface_index)
+{
+    int   rule_iface;
+    void *dst_in_addr;
+
+    if (rtm->rtm_family != sa_remote->sa_family) {
+        return 0;
+    }
```
shasson5, line 175: address family is guaranteed to match because that's what we passed in the netlink request (line 214)

src/uct/ib/base/ib_iface.h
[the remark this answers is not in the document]
shasson5: unresolved

src/uct/ib/base/ib_iface.c
```diff
@@ -1322,11 +1360,20 @@ uct_ib_iface_init_roce_addr_prefix(uct_ib_iface_t *iface,
...
     }
 
+    /* Override the original value of reachability_mode if
+       rocev2_local_subnet is enabled to maintain backward compatibility */
+    if (config->rocev2_local_subnet) {
+        iface->config.reachability_mode = UCT_IB_REACHABILITY_MODE_LOCAL_SUBNET;
+    } else {
+        iface->config.reachability_mode = config->reachability_mode;
+    }
+
```
shasson5, line 1376: `iface->config.reachability_mode = config->rocev2_local_subnet ? UCT_IB_REACHABILITY_MODE_LOCAL_SUBNET : config->reachability_mode`

src/uct/base/uct_iface.c
```diff
@@ -1034,6 +1036,23 @@ int uct_iface_local_is_reachable(uct_iface_local_addr_ns_t *addr_ns,
...
+
+    if (!ucs_netlink_rule_exists(iface, sa_remote)) {
+        uct_iface_fill_info_str_buf(
+                params, "remote address %s is not routable",
+                ucs_sockaddr_str(sa_remote, ip_str, 128));
+
+        return 0;
+    }
+
+    return 1;
```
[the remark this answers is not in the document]
amastbaum: On a second thought maybe it's better to keep it because this function will also be used by roce

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+#define NETLINK_MESSAGE_MAX_SIZE 8195
+
+struct route_info {
+    const struct sockaddr *sa_remote;
+    int                    if_index;
+    int                    matching;
+};
+
+
+static void ucs_rtnetlink_get_route_info(int **if_idx, void **dst_in_addr,
```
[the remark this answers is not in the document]
shasson5: unresolved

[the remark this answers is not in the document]
amastbaum: It was resolved here https://github.com/openucx/ucx/pull/10135/commits/7b9b8f7796917dce97838331f47f14426e8b0d50

src/uct/base/uct_iface.c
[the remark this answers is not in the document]
shasson5: better to put it only where used, and if needed move it in next PR

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+
+    if (rule_iface != iface_index) {
+        return 0;
+    }
+
+    return ucs_bitwise_is_equal(ucs_sockaddr_get_inet_addr(sa_remote),
+                             dst_in_addr, rtm->rtm_dst_len);
+}
+
+ucs_status_t
```
shasson5, line 186: static

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+    if (ucs_rtnetlink_is_rule_matching((struct rtmsg*)nl_msg, RTM_PAYLOAD(nlh),
+                                       info->sa_remote, info->if_index)) {
+        info->matching = 1;
+        return UCS_OK;
+    }
+
+    return UCS_INPROGRESS;
+}
+
+int ucs_netlink_rule_exists(const char *iface,
```
shasson5, line 200: `const char *iface` => `const char *iface_name`

src/uct/base/uct_iface.c
```diff
@@ -17,7 +17,9 @@
 #include <uct/api/uct.h>
 #include <uct/api/v2/uct_v2.h>
 #include <ucs/async/async.h>
+#include <ucs/sys/sock.h>
 #include <ucs/sys/string.h>
+#include <ucs/sys/ucs_netlink.h>
```
shasson5, line 22: revert

shasson5 (commented):
@yosefe @gleon99 please review

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+#include <ucs/type/status.h>
+
+#include <errno.h>
+#include <linux/netlink.h>
+#include <linux/rtnetlink.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+
+/* *******************************************************
```
gleon99, line 26: Better remove, we don't usually do such boxes.

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+ * General Netlink utilities                             *
+ * ***************************************************** */
+
+static ucs_status_t ucs_netlink_socket_init(int *fd, int protocol)
```
gleon99, line 30: `fd` -> `fd_p` (as inside `ucs_socket_create`)

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+        status = UCS_ERR_IO_ERROR;
+        goto err_close_socket;
+    }
+
+    return UCS_OK;
+
+err_close_socket:
+    ucs_close_fd(fd);
+err:
+    *fd = -1;
```
gleon99, line 55: IMO redundant

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+    return status;
+}
+
+ucs_status_t ucs_netlink_send_cmd(int protocol, unsigned short nlmsg_type,
+                                  void *nl_protocol_hdr,
+                                  size_t nl_protocol_hdr_size,
+                                  char *recv_msg_buf, size_t *recv_msg_buf_len)
+{
+    struct nlmsghdr nlh = {0};
```
gleon99, line 64: `nlh` -> more descriptive name?

gleon99: Is init really needed?

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+    nlh.nlmsg_type  = nlmsg_type;
+    nlh.nlmsg_flags = NLM_F_REQUEST | NLM_F_DUMP;
+    iov[0].iov_base = &nlh;
+    iov[0].iov_len  = sizeof(nlh);
+    iov[1].iov_base = nl_protocol_hdr;
+    iov[1].iov_len  = nl_protocol_hdr_size;
+
+    do {
+        status = ucs_socket_sendv_nb(fd, iov, 2, &dummy);
+    } while (status == UCS_ERR_NO_PROGRESS);
```
gleon99, line 85: Maybe not?

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+out:
+    ucs_close_fd(&fd);
+    return status;
+}
+
+ucs_status_t ucs_netlink_parse_msg(void *msg, size_t msg_len,
+                                   ucs_netlink_parse_cb_t parse_cb, void *arg)
+{
+    struct nlmsghdr *nlh;
+    ucs_status_t status = UCS_INPROGRESS;
```
gleon99, line 111: Initialized first

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+ucs_status_t ucs_netlink_send_cmd(int protocol, unsigned short nlmsg_type,
+                                  void *nl_protocol_hdr,
+                                  size_t nl_protocol_hdr_size,
+                                  char *recv_msg_buf, size_t *recv_msg_buf_len)
+{
+    struct nlmsghdr nlh = {0};
+    ucs_status_t status;
+    int fd;
+    struct iovec iov[2];
+    size_t dummy;
```
gleon99, line 68: Rename

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+        struct nlmsgerr *err = (struct nlmsgerr*)NLMSG_DATA(nlh);
+        ucs_error("failed to parse netlink message header (%d)", err->error);
+        return UCS_ERR_IO_ERROR;
+    }
+
+    return UCS_OK;
+}
+
+
+/* *******************************************************
```
gleon99, line 130: Remove

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+
+    for (nlh = (struct nlmsghdr*)msg;
+         (status == UCS_INPROGRESS) && NLMSG_OK(nlh, msg_len) &&
+         (nlh->nlmsg_type != NLMSG_DONE) && (nlh->nlmsg_type != NLMSG_ERROR);
+         nlh = NLMSG_NEXT(nlh, msg_len)) {
+        status = parse_cb(nlh, NLMSG_DATA(nlh), arg);
+    }
+
+    if (nlh->nlmsg_type == NLMSG_ERROR) {
+        struct nlmsgerr *err = (struct nlmsgerr*)NLMSG_DATA(nlh);
```
gleon99, line 121: Move to top of func

gleon99: `(struct nlmsgerr*)NLMSG_DATA` -> `(struct nlmsgerr *)NLMSG_DATA`

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,248 @@
...
+
+    return UCS_OK;
+}
+
+
+/* *******************************************************
+ * Route Netlink utilities                               *
+ * ***************************************************** */
+
+#define NETLINK_MESSAGE_MAX_SIZE 8195
```
gleon99, line 134: Move this and struct to top + maybe to `.h`

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,88 @@
...
+
+
+/**
+ * Callback function for parsing individual netlink messages.
+ *
+ * @param [in] nlh    Pointer to the netlink message header.
+ * @param [in] nl_msg Pointer to the netlink message payload.
+ * @param [in] arg    User-provided argument passed through from the caller.
+ *
+ * @return UCS_INPROGRESS if parsing should continue to the next message,
```
gleon99, line 26: Maybe simplify:
`UCS_OK is parsing is complete, UCS_INPROGRESS is there are more messages to be parsed, or error code otherwise\in case of error`

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,88 @@
...
+ * @param [in]    arg       Pointer to the callback function arguments.
+ *
+ * @return UCS_OK if parsed successfully, or error code otherwise.
+ */
+ucs_status_t ucs_netlink_parse_msg(void *msg, size_t msg_len,
+                                   ucs_netlink_parse_cb_t parse_cb, void *arg);
+
+
+/**
+ * For a given network interface name and a destination address, return
```
gleon99, line 74: Check whether a X exists for given Y..

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,88 @@
...
+ * whether a routing table rule exists.
+ *
+ * @param [in]    iface       Pointer to the name of the interface.
```
gleon99, line 77: Too many spaces

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,88 @@
...
+ * @param [in]    sa_remote   Pointer to the destination address.
+ *
+ * @return 1 if such rule exists, or 0 otherwise.
```
gleon99, line 80: rephrase `such`

src/uct/ib/base/ib_iface.c
```diff
@@ -42,6 +43,13 @@ const char *uct_ib_mtu_values[] = {
     [UCT_IB_MTU_LAST]       = NULL
 };
 
+const char *uct_ib_reachability_mode[] = {
```
gleon99, line 46: `modes`

src/uct/ib/base/ib_iface.c
```diff
@@ -181,6 +189,13 @@ ucs_config_field_t uct_ib_iface_config_table[] = {
    "It must not be used together with UCX_IB_GID_INDEX.",
    ucs_offsetof(uct_ib_iface_config_t, rocev2_subnet_filter), UCS_CONFIG_TYPE_ALLOW_LIST},
 
+  {"ROCE_REACHABILITY_MODE", "route",
+   "The mode used for performing the reachability check\n"
+   " - route        - all routable addresses will be assumed as reachable\n"
+   " - local_subnet - only addresses within the interface's subnet will be assumed as reachable.\n"
+   " - none         - no reachability check will be done. all addresses are reachable",
```
gleon99, line 196: `all addresses are assumes as reachable, without any check`

src/uct/ib/base/ib_iface.c
```diff
@@ -645,25 +660,29 @@ static int
...
     sa_family_t remote_ib_addr_af;
     char local_str[128], remote_str[128];
     int matched;
+    char ndev_name[IFNAMSIZ];
+    ucs_status_t status;
 
-    /* check for wildcards in the RoCE version (RDMACM or non-RoCE cases) */
-    if ((uct_ib_address_flags_get_roce_version(remote_ib_addr_flags)) ==
-         UCT_IB_DEVICE_ROCE_ANY) {
+    if (iface->config.reachability_mode == UCT_IB_REACHABILITY_MODE_NONE) {
```
gleon99, line 679: Merge 2 `if`s (keep comment + add new)

src/uct/ib/base/ib_iface.c
```diff
@@ -695,18 +714,43 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
+                                              local_gid_info->gid_index,
+                                              ndev_name, sizeof(ndev_name));
+    if (status != UCS_OK) {
+        uct_iface_fill_info_str_buf(
+               params, "couldn't get network interface name");
         return 0;
     }
 
+    if (iface->config.reachability_mode == UCT_IB_REACHABILITY_MODE_ROUTE) {
+        if (!ucs_netlink_rule_exists(ndev_name, (struct sockaddr*)&sa_remote)) {
```
gleon99, line 739: Invert branches -> less overall indented lines

src/uct/ib/base/ib_iface.h
```diff
@@ -175,60 +187,63 @@ struct uct_ib_iface_config {
...
 
     /* Whether to check RoCEv2 reachability by IP address and local subnet */
-    int                     rocev2_local_subnet;
+    int                              rocev2_local_subnet;
 
     /* Length of subnet prefix for reachability check */
-    unsigned long           rocev2_subnet_pfx_len;
+    unsigned long                    rocev2_subnet_pfx_len;
+
+    /* The mode used for performing the reachability check */
```
gleon99, line 219: Remove both `the` (consistency)

src/ucs/sys/ucs_netlink.c
shasson5: how will check interrupts otherwise?

src/ucs/sys/ucs_netlink.c
amastbaum: Agree

src/ucs/sys/ucs_netlink.c
amastbaum: Yes, it is, because other fields should be initialized to 0

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,54 @@
...
+* See file LICENSE for terms.
+*/
+
+#ifndef UCS_NETLINK_H
+#define UCS_NETLINK_H
+
+#include <ucs/type/status.h>
+
+#include <stddef.h>
+#include <linux/netlink.h>
```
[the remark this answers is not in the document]
amastbaum: Which order? I thought it should be ordered alphabetically per level of the tree, for instance

a/d
a/e
c

src/ucs/sys/ucs_netlink.h
amastbaum: It's aligned, isn't it?

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,238 @@
...
+}
+
+struct ucs_netlink_route_info_t {
+    const struct sockaddr *sa_remote;
+    int                    if_index;
+    int                    matching;
+};
+
+
+static ucs_status_t ucs_rtnetlink_get_route_info(struct rtattr *rta, int len,
```
gleon99, line 133: Let's discuss netlink / rnetlink separation

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,238 @@
...
+    for (; RTA_OK(rta, len); rta = RTA_NEXT(rta, len)) {
+        if (rta->rta_type == RTA_OIF) {
+            *if_idx = *((int*)RTA_DATA(rta));
+        } else if (rta->rta_type == RTA_DST) {
+            *dst_in_addr = RTA_DATA(rta);
+        }
+    }
+
+    if ((*if_idx == -1) || (*dst_in_addr == NULL)) {
+        ucs_debug("either iface index or dest address are missing "
```
gleon99, line 148: `ucs_diag` + add specific info in the string

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,238 @@
...
+                                     &dst_in_addr) != UCS_OK) {
+        return 0;
+    }
+
+    if (rule_iface != iface_index) {
+        return 0;
+    }
+
+    return ucs_bitwise_is_equal(ucs_sockaddr_get_inet_addr(sa_remote),
+                             dst_in_addr, rtm->rtm_dst_len);
```
gleon99, line 173: format

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,238 @@
...
+    struct rtmsg rtm                     = {0};
+    ucs_status_t status;
+    size_t recv_msg_len;
+    int iface_index;
+
+    rtm.rtm_family = sa_remote->sa_family;
+    rtm.rtm_table  = RT_TABLE_MAIN;
+
+    recv_msg_len = NETLINK_MESSAGE_MAX_SIZE;
+    recv_msg = ucs_malloc(NETLINK_MESSAGE_MAX_SIZE, "netlink recv message");
```
gleon99, line 204: align

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,238 @@
...
+    status = ucs_netlink_send_cmd(NETLINK_ROUTE, RTM_GETROUTE, &rtm,
+                                  sizeof(rtm), recv_msg, &recv_msg_len);
+    if (status != UCS_OK) {
+        ucs_error("failed to send netlink route message (%s)",
+                  ucs_status_string(status));
+        goto out;
+    }
+
+    iface_index = if_nametoindex(iface_name);
+    if (iface_index == 0) {
```
gleon99, line 219: + errno

src/uct/ib/base/ib_iface.c
```diff
@@ -181,6 +189,13 @@ ucs_config_field_t uct_ib_iface_config_table[] = {
    "It must not be used together with UCX_IB_GID_INDEX.",
    ucs_offsetof(uct_ib_iface_config_t, rocev2_subnet_filter), UCS_CONFIG_TYPE_ALLOW_LIST},
 
+  {"ROCE_REACHABILITY_MODE", "route",
+   "The mode used for performing the reachability check\n"
+   " - route        - all routable addresses will be assumed as reachable\n"
```
gleon99, line 194: future tense -> present

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,238 @@
...
+{
+    ucs_status_t status = UCS_INPROGRESS;
+    struct nlmsghdr *nlh;
+
+    for (nlh = (struct nlmsghdr*)msg;
+         (status == UCS_INPROGRESS) && NLMSG_OK(nlh, msg_len) &&
+         nlh->nlmsg_type != NLMSG_DONE;
+         nlh = NLMSG_NEXT(nlh, msg_len)) {
+        if (nlh->nlmsg_type == NLMSG_ERROR) {
+            struct nlmsgerr *err = (struct nlmsgerr *)NLMSG_DATA(nlh);
```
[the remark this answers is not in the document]
amastbaum: What do you mean? `nlmsg_type` should be checked for each `nlh`

src/ucs/sys/ucs_netlink.c
amastbaum: `*dst_in_addr` is `void *` pointing to a "raw" IP address, so in order to print it I need to pass its family type from outside of the function. Is it worth it?

src/ucs/sys/ucs_rtnetlink.c
```diff
@@ -0,0 +1,135 @@
...
+    if (status != UCS_OK) {
+        ucs_error("failed to parse netlink route message (%s)",
+                  ucs_status_string(status));
+        goto out;
+    }
+
+out:
+    ucs_free(recv_msg);
+    return info.matching;
+}
```
gleon99, line 135: Add newline

src/ucs/sys/ucs_rtnetlink.c
```diff
@@ -0,0 +1,135 @@
...
+    status = ucs_netlink_send_cmd(NETLINK_ROUTE, RTM_GETROUTE, &rtm,
+                                  sizeof(rtm), recv_msg, &recv_msg_len);
+    if (status != UCS_OK) {
+        ucs_error("failed to send netlink route message (%s)",
+                  ucs_status_string(status));
+        goto out;
+    }
+
+    iface_index = if_nametoindex(iface_name);
+    if (iface_index == 0) {
```
gleon99, line 116: Maybe move it to top? It doesn't depend on alloc or any netlink cmd action -> check it first..

src/ucs/sys/ucs_rtnetlink.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @return 1 if rule exists, or 0 otherwise.
+ */
+int ucs_netlink_rule_exists(const char *iface,
+                            const struct sockaddr *sa_remote);
+
+
+END_C_DECLS
+
+#endif /* UCS_RTNETLINK_H */
```
gleon99, line 32: Add newline

src/uct/ib/base/ib_iface.c
```diff
@@ -181,6 +189,13 @@ ucs_config_field_t uct_ib_iface_config_table[] = {
    "It must not be used together with UCX_IB_GID_INDEX.",
    ucs_offsetof(uct_ib_iface_config_t, rocev2_subnet_filter), UCS_CONFIG_TYPE_ALLOW_LIST},
 
+  {"ROCE_REACHABILITY_MODE", "route",
+   "The mode used for performing the reachability check\n"
+   " - route        - all routable addresses are assumed as reachable\n"
+   " - local_subnet - only addresses within the interface's subnet are assumed as reachable.\n"
+   " - none         - all addresses are assumed as reachable, without any check",
```
gleon99, line 196: Maybe `none` -> `all`?
The semantics here suggest the "mode X" means "X address is assumed to be reachable",
thus "mode none" implies "none is reachable", while in fact the opposite is true.

src/uct/ib/base/ib_iface.c
```diff
@@ -695,18 +712,44 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
+        }
+
+        uct_iface_fill_info_str_buf(
+               params, "remote address %s is not routable",
+               ucs_sockaddr_str((struct sockaddr *)&sa_remote, remote_str, 128));
+
         return 0;
     }
 
+    /* check for zero-sized netmask */
```
gleon99, line 748: This comment repeats the code below it, without adding any info.
Please either make it more informative, e.g. "An 0-bit netmask implies any address is reachable" or remove it.

gleon99 (approved):
@yosefe ?

src/ucs/Makefile.am
```diff
@@ -131,6 +131,8 @@ noinst_HEADERS = \
 	sys/iovec.h \
 	sys/iovec.inl \
 	sys/ptr_arith.h \
+	sys/ucs_netlink.h \
+	sys/ucs_rtnetlink.h \
```
yosefe, line 135: 1. can we rename the H file to "netlink.h"? even though it's the same name as the system header i think it should be ok since they have diferent path
2. maybe unite these 2 headers / c files to one: netlink.h, netlink.c

src/ucs/sys/ucs_netlink.h
```diff
@@ -0,0 +1,72 @@
...
+ * @param [in]    nl_protocol_hdr_size Protocol struct size.
+ * @param [out]   recv_msg_buf         The buffer that will hold the received
+ *                                     message.
+ * @param [inout] recv_msg_buf_len     Pointer to the size of the buffer and to
+ *                                     store the length of the received message.
+ *
+ * @return UCS_OK if received successfully, or error code otherwise.
+ */
+ucs_status_t ucs_netlink_send_cmd(int protocol, unsigned short nlmsg_type,
+                                  void *nl_protocol_hdr,
```
yosefe, line 50: could be const void*?

src/ucs/sys/ucs_rtnetlink.c
```diff
@@ -0,0 +1,135 @@
...
+#include <ucs/debug/memtrack_int.h>
+#include <ucs/sys/ucs_rtnetlink.h>
+#include <ucs/sys/ucs_netlink.h>
+#include <ucs/sys/sock.h>
+#include <ucs/type/status.h>
+
+#include <errno.h>
+#include <linux/rtnetlink.h>
+
+#define NETLINK_MESSAGE_MAX_SIZE 8195
```
yosefe, line 21: can we get it from a system header file? why is it the right value?

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,120 @@
...
+    ucs_status_t status;
+
+    status = ucs_socket_create(AF_NETLINK, SOCK_RAW, protocol, fd_p);
+    if (status != UCS_OK) {
+        ucs_error("failed to create netlink socket (%s)",
+                  ucs_status_string(status));
+        goto err;
+    }
+
+    sa.nl_family = AF_NETLINK;
```
yosefe, line 36: can we set it during the initialization?

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,120 @@
...
+
+    if (bind(*fd_p, (struct sockaddr *)&sa, sizeof(sa)) < 0) {
+        ucs_error("failed to bind netlink socket %d", *fd_p);
```
yosefe, line 39: pls print errno using %m

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,120 @@
...
+                  ucs_status_string(status));
+        goto out;
+    }
+
+    do {
+        status = ucs_socket_recv_nb(fd, recv_msg_buf, recv_msg_buf_len);
+    } while (status == UCS_ERR_NO_PROGRESS);
+
+    if (status != UCS_OK) {
+        ucs_error("failed to receive netlink message");
```
yosefe, line 91: shall we print the error string like in line 81?

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,120 @@
...
+    } while (status == UCS_ERR_NO_PROGRESS);
+
+    if (status != UCS_OK) {
+        ucs_error("failed to send netlink message (%s)",
+                  ucs_status_string(status));
+        goto out;
+    }
+
+    do {
+        status = ucs_socket_recv_nb(fd, recv_msg_buf, recv_msg_buf_len);
```
yosefe, line 87: how is partial recv handled? i'd expect the next iteration will begin from the offset according to number of bytes received in the previous iteration. same for send.
probably needed to use ucs_socket_send / ucs_socket_recv?

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,120 @@
...
+                                   ucs_netlink_parse_cb_t parse_cb, void *arg)
+{
+    ucs_status_t status = UCS_INPROGRESS;
+    struct nlmsghdr *nlh;
+
+    for (nlh = (struct nlmsghdr *)msg;
+         (status == UCS_INPROGRESS) && NLMSG_OK(nlh, msg_len) &&
+         nlh->nlmsg_type != NLMSG_DONE;
+         nlh = NLMSG_NEXT(nlh, msg_len)) {
+        if (nlh->nlmsg_type == NLMSG_ERROR) {
```
yosefe, line 110: IMO a while loop would be more readable

src/ucs/sys/ucs_netlink.c
```diff
@@ -0,0 +1,120 @@
...
+            struct nlmsgerr *err = (struct nlmsgerr *)NLMSG_DATA(nlh);
+            ucs_error("failed to parse netlink message header (%d)", err->error);
```
yosefe, line 112: is there a way to print the error string?

src/ucs/sys/ucs_rtnetlink.h
```diff
@@ -0,0 +1,32 @@
...
+#include <netinet/in.h>
+
+BEGIN_C_DECLS
+
+
+/**
+ * Check whether a routing table rule exists for a given network
+ * interface name and a destination address.
+ *
+ * @param [in]  iface      Pointer to the name of the interface.
```
yosefe, line 21: maybe rename to if_name for consistency?
is it something like "eth0" ?

src/ucs/sys/ucs_rtnetlink.c
```diff
@@ -0,0 +1,135 @@
...
+                            const struct sockaddr *sa_remote)
+{
+    char *recv_msg                       = NULL;
+    struct ucs_netlink_route_info_t info = {0};
+    struct rtmsg rtm                     = {0};
+    ucs_status_t status;
+    size_t recv_msg_len;
+    int iface_index;
+
+    iface_index = if_nametoindex(iface_name);
```
yosefe, line 97: can we have some sort of caching so that if is_reachable is called multiple times, even to different destinations, we will not send netlink command every time?
seems we could read the routing table just once?

src/ucs/sys/ucs_rtnetlink.c
```diff
@@ -0,0 +1,135 @@
...
+#include <errno.h>
+#include <linux/rtnetlink.h>
+
+#define NETLINK_MESSAGE_MAX_SIZE 8195
+
+
+struct ucs_netlink_route_info_t {
+    const struct sockaddr *sa_remote;
+    int                    if_index;
+    int                    matching;
```
yosefe, line 27: matching -> found

src/ucs/sys/ucs_rtnetlink.c
```diff
@@ -0,0 +1,135 @@
...
+
+static int ucs_rtnetlink_is_rule_matching(struct rtmsg *rtm, size_t rtm_len,
+                                          const struct sockaddr *sa_remote,
+                                          int iface_index)
+{
+    int   rule_iface;
+    void *dst_in_addr;
+
+    if (ucs_rtnetlink_get_route_info(RTM_RTA(rtm), rtm_len, &rule_iface,
+                                     &dst_in_addr) != UCS_OK) {
```
yosefe, line 61: minor: seems the code would be smaller if we removed ucs_rtnetlink_get_route_info helper function and put its code directly in ucs_rtnetlink_is_rule_matching, which is already small enough - but up to you

src/ucs/sys/ucs_rtnetlink.h
```diff
@@ -0,0 +1,32 @@
...
+ * @param [in]  sa_remote  Pointer to the destination address.
+ *
+ * @return 1 if rule exists, or 0 otherwise.
+ */
+int ucs_netlink_rule_exists(const char *iface,
```
yosefe, line 26: maybe rename to ucs_netlink_route_exists which is a bit more explicit?

src/uct/ib/base/ib_iface.c
```diff
@@ -695,18 +712,44 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
     if (uct_ib_device_roce_gid_to_sockaddr(remote_ib_addr_af,
                                            remote_ib_addr + 1,
                                            &sa_remote) != UCS_OK) {
         uct_iface_fill_info_str_buf(
-               params, "Couldn't convert remote RoCE address to socket address");
+               params, "couldn't convert remote RoCE address to socket address");
+        return 0;
+    }
+
+    status = uct_ib_device_get_roce_ndev_name(dev, port_num,
```
yosefe, line 727: the function uct_ib_iface_roce_is_reachable is already quite big, maybe split it to sub-function, so that each one handles a different mode of reachability?

src/uct/ib/base/ib_iface.c
```diff
@@ -695,18 +712,44 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
 
+    if (iface->config.reachability_mode == UCT_IB_REACHABILITY_MODE_ROUTE) {
+        if (ucs_netlink_rule_exists(ndev_name, (struct sockaddr *)&sa_remote)) {
+            return 1;
+        }
+
+        uct_iface_fill_info_str_buf(
+               params, "remote address %s is not routable",
+               ucs_sockaddr_str((struct sockaddr *)&sa_remote, remote_str, 128));
+
```
yosefe, line 744: remove space line

src/uct/ib/base/ib_iface.c
```diff
@@ -1322,11 +1366,18 @@ uct_ib_iface_init_roce_addr_prefix(uct_ib_iface_t *iface,
...
+        (!config->rocev2_local_subnet &&
+        (config->reachability_mode != UCT_IB_REACHABILITY_MODE_LOCAL_SUBNET))) {
         iface->addr_prefix_bits = 0;
         return UCS_OK;
     }
 
+    /* Override the original value of reachability_mode if
+       rocev2_local_subnet is enabled to maintain backward compatibility */
+    iface->config.reachability_mode = config->rocev2_local_subnet ?
+                                      UCT_IB_REACHABILITY_MODE_LOCAL_SUBNET :
```
yosefe, line 1378: maybe unite common logic with lines 1369-1370 so a single helpe function will return the reachability mode based on config, and it would consider both rocev2_local_subnet and reachability_mode fields for backward compat?

src/ucs/sys/ucs_rtnetlink.c
amastbaum: @yosefe I couldn't find a good enough prediction for the rt-netlink message size nor a definition from a system header file, so I did a couple of things:
  - I used netlink usage examples and saw that 8195 was a pretty common max size
  - I measured routing table (rt-netlink response) max sizes on various machines and it didn't exceed this number. The maximum size I found was half of it.

yosefe: is there way to increase the size at runtime if it is not enough, or handle incoming message in chunks?

gleon99: [### ](https://www.perplexity.ai/search/netlink-message-max-size-8k-bafGz1cIQYCUtsqMhi42dg)

The Kernel safely assumes it's 8K. Do we really need to accommodate for size increase?

amastbaum: Thanks @gleon99! I'll add a comment to clarify the size.

@yosefe I had a few other versions of the buffer allocation and size determination and after some discussions (with @shasson5 as well) we decided it's not worth the 2 requests required (in any case, even when the buffer is originally enough because you need to flush the message from the socket) and that this solution will support the majority of cases. So considering @gleon99's comment I guess we can be even more sure about it.

yosefe: according to the comment in https://github.com/torvalds/linux/blob/master/include/linux/netlink.h#L265, seems it is the limit for a single message, not necessarily the entire payload of the reply data, which could span multiple messages.
we should add some protection/error/warning if we're not able to receive the entire reply in 8K buffer. Since this PR is already big enough, we can solve this in next PR.

amastbaum:
@yosefe I was mistaken in thinking that the total message size is written at the beginning of the message. Instead, it represents the size of each individual netlink message (i.e. a single routing table entry).
The only approach I’ve found so far is to call `ucs_socket_recv_nb` with the `MSG_TRUNC | MSG_PEEK` flags which allow you to determine the entire message size (the one stored in the receive queue) without fully reading or flushing it.

yosefe:
> @yosefe I was mistaken in thinking that the total message size is written at the beginning of the message. Instead, it represents the size of each individual netlink message (i.e. a single routing table entry). The only approach I’ve found so far is to call `ucs_socket_recv_nb` with the `MSG_TRUNC | MSG_PEEK` flags which allow you to determine the entire message size (the one stored in the receive queue) without fully reading or flushing it.

Sounds good - let's use MSG_TRUNC | MSG_PEEK

src/uct/ib/base/ib_iface.c
```diff
@@ -635,43 +650,100 @@ static void uct_ib_iface_log_subnet_info(const struct sockaddr_storage *sa1,
...
+    }
+
+    return matched;
+}
+
 static int
 uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
                                const uct_ib_address_t *remote_ib_addr,
                                unsigned prefix_bits,
+                               uct_ib_iface_t *iface,
```
yosefe, line 722: 1. iface should be added as the first parameter
2. since iface parameter is added, other params such as local_gid_info and prefix_bits can be removed since they are part of the iface

src/uct/ib/base/ib_iface.c
```diff
@@ -695,39 +767,28 @@ uct_ib_iface_roce_is_reachable(const uct_ib_device_gid_info_t *local_gid_info,
...
-                    " is %s, remote IP is %s",
-                    prefix_bits,
-                    ucs_sockaddr_str((struct sockaddr *)&sa_local,
-                                     local_str, 128),
-                    ucs_sockaddr_str((struct sockaddr *)&sa_remote,
-                                     remote_str, 128));
-    }
-
-    return matched;
+    return uct_ib_iface_roce_is_reachable_in_local_subnet(
```
yosefe, line 789: before this line - check or assert that config.reachability_mode is local subnet

src/uct/ib/base/ib_iface.h
```diff
@@ -340,29 +340,28 @@ struct uct_ib_iface {
...
+        uint8_t                      port_num;
+        uint8_t                      sl;
+        uint8_t                      reverse_sl;
+        uint8_t                      traffic_class;
+        uint8_t                      hop_limit;
+        uint8_t                      qp_type;
+        uint8_t                      force_global_addr;
+        uint8_t                      flid_enabled;
+        enum ibv_mtu                 path_mtu;
+        uint8_t                      counter_set_id;
```
yosefe, line 363: need to restore the indentation - all fields should be aligned on same column

src/ucs/sys/netlink.c
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    return UCS_INPROGRESS;
+}
+
+int ucs_netlink_route_exists(const char *if_name,
+                             const struct sockaddr *sa_remote)
+{
+    ucs_netlink_route_info_t info = {0};
+    struct rtmsg rtm              = {0};
```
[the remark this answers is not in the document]
amastbaum: Because it has many fields that are not assigned with anything and if they're assigned with trash values it fails the request

yosefe (approved):

yosefe (dismissed):

yosefe (approved):

PR #10138, opened 2024-09-09
UCT/IB/MLX5/DC: introducing dcs_hybrid policy

roiedanino:
## What
Introducing a new dci allocation policy - dcs_hybrid which will behave the same as dcs_quota except using a dedicated HW dci when available instead of waiting for a dci from the pool to be available

## Why ?
Utilizing hw resources instead of waiting for another dci, improves latency for certain message sizes

## How ?
In case `dci_alloc_or_create` returned 0 - use the dedicated HW dci with a new dci channel.

### DIFF
--- a/src/uct/ib/mlx5/dc/dc_mlx5.c
+++ b/src/uct/ib/mlx5/dc/dc_mlx5.c
@@ -29,11 +29,12 @@
 
 
 const char *uct_dc_tx_policy_names[] = {
-    [UCT_DC_TX_POLICY_DCS]       = "dcs",
-    [UCT_DC_TX_POLICY_DCS_QUOTA] = "dcs_quota",
-    [UCT_DC_TX_POLICY_RAND]      = "rand",
-    [UCT_DC_TX_POLICY_HW_DCS]    = "hw_dcs",
-    [UCT_DC_TX_POLICY_LAST]      = NULL
+    [UCT_DC_TX_POLICY_DCS]        = "dcs",
+    [UCT_DC_TX_POLICY_DCS_QUOTA]  = "dcs_quota",
+    [UCT_DC_TX_POLICY_DCS_HYBRID] = "dcs_hybrid",
+    [UCT_DC_TX_POLICY_RAND]       = "rand",
+    [UCT_DC_TX_POLICY_HW_DCS]     = "hw_dcs",
+    [UCT_DC_TX_POLICY_LAST]       = NULL
 };
 
 static const char *uct_dct_affinity_policy_names[] = {
@@ -250,7 +251,7 @@ static ucs_status_t uct_dc_mlx5_iface_query(uct_iface_h tl_iface, uct_iface_attr
 
     /* Error handling is not supported with random dci policy
      * TODO: Fix */
-    if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
         iface_attr->cap.flags &= ~(UCT_IFACE_FLAG_ERRHANDLE_PEER_FAILURE |
                                    UCT_IFACE_FLAG_ERRHANDLE_ZCOPY_BUF    |
                                    UCT_IFACE_FLAG_ERRHANDLE_REMOTE_MEM);
@@ -366,7 +367,8 @@ static void uct_ib_mlx5dv_dci_qp_init_attr(uct_ib_qp_init_attr_t *qp_attr,
 }
 
 ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
-                                          uint8_t dci_index, int connect)
+                                          uint8_t dci_index, int connect,
+                                          uint8_t num_dci_channels)
 {
     uct_ib_iface_t *ib_iface   = &iface->super.super.super;
     uct_ib_mlx5_qp_attr_t attr = {};
@@ -391,7 +393,7 @@ ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
                                            UCT_DC_MLX5_IFACE_FLAG_DCI_FULL_HANDSHAKE;
         attr.rdma_wr_disabled            = (iface->flags & UCT_DC_MLX5_IFACE_FLAG_DISABLE_PUT) &&
                                            (md->flags & UCT_IB_MLX5_MD_FLAG_NO_RDMA_WR_OPTIMIZED);
-        attr.log_num_dci_stream_channels = ucs_ilog2(iface->tx.num_dci_channels);
+        attr.log_num_dci_stream_channels = ucs_ilog2(num_dci_channels);
         status = uct_ib_mlx5_devx_create_qp(ib_iface,
                                             &iface->super.cq[UCT_IB_DIR_TX],
                                             &iface->super.cq[UCT_IB_DIR_RX],
@@ -459,7 +461,7 @@ init_qp:
         }
     }
 
-    if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
         ucs_arbiter_group_init(&dci->arb_group);
     } else {
         dci->ep = NULL;
@@ -765,7 +767,7 @@ uct_dc_mlx5_destroy_dci(uct_dc_mlx5_iface_t *iface, uct_dc_dci_t *dci)
     uct_rc_txqp_cleanup(&iface->super.super, &dci->txqp);
     uct_ib_mlx5_destroy_qp(md, &dci->txwq.super);
 
-    if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
         ucs_arbiter_group_cleanup(&dci->arb_group);
     }
     uct_ib_mlx5_qp_mmio_cleanup(&dci->txwq.super, dci->txwq.reg);
@@ -886,8 +888,8 @@ ucs_status_t uct_dc_mlx5_iface_resize_and_fill_dcis(uct_dc_mlx5_iface_t *iface,
 }
 
 static ucs_status_t
-uct_dc_mlx5_iface_dcis_create(uct_dc_mlx5_iface_t *iface,
-                              const uct_dc_mlx5_iface_config_t *config)
+uct_dc_mlx5_iface_init_dcis_array(uct_dc_mlx5_iface_t *iface,
+                                  const uct_dc_mlx5_iface_config_t *config)
 {
     ucs_status_t status;
     uct_dc_dci_t *dci;
@@ -901,7 +903,7 @@ uct_dc_mlx5_iface_dcis_create(uct_dc_mlx5_iface_t *iface,
 
     ucs_array_length(&iface->tx.dcis) = 0;
 
-    status = uct_dc_mlx5_iface_create_dci(iface, 0, 0);
+    status = uct_dc_mlx5_iface_create_dci(iface, 0, 0, 1);
     if (status != UCS_OK) {
         return status;
     }
@@ -1178,7 +1180,7 @@ static void uct_dc_mlx5_iface_cleanup_fc_ep(uct_dc_mlx5_iface_t *iface)
         goto out;
     }
 
-    if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
         txqp = &fc_dci->txqp;
         ucs_queue_for_each_safe(op, iter, &txqp->outstanding, queue) {
             if (op->handler == uct_dc_mlx5_ep_fc_pure_grant_send_completion) {
@@ -1355,7 +1357,7 @@ static void uct_dc_mlx5_dci_handle_failure(uct_dc_mlx5_iface_t *iface,
     uct_dc_mlx5_ep_t *ep;
     ucs_log_level_t  level;
 
-    if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
+    if (uct_dc_mlx5_is_dci_shared(iface, dci_index)) {
         ep    = NULL;
         level = UCS_LOG_LEVEL_FATAL; /* error handling is not supported with rand dci */
     } else {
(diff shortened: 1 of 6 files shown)


src/uct/ib/mlx5/dc/dc_mlx5_ep.c
```diff
@@ -1395,7 +1396,11 @@ uct_dc_mlx5_iface_dci_do_pending_wait(ucs_arbiter_t *arbiter,
                 uct_dc_mlx5_iface_dci(iface, ep->dci)->pool_index);
 
     if (!uct_dc_mlx5_iface_dci_can_alloc_or_create(iface, pool_index)) {
-        return UCS_ARBITER_CB_RESULT_STOP;
+        if (uct_dc_mlx5_set_ep_to_hw_dcs(iface, ep) == UCS_OK) {
+            return UCS_ARBITER_CB_RESULT_DESCHED_GROUP;
```
yosefe, line 1400: shouldn't we go to line 1408 to sched for TX?

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -266,6 +267,19 @@ uct_dc_mlx5_iface_is_dci_shared(const uct_dc_mlx5_iface_t *iface)
     return iface->tx.policy >= UCT_DC_TX_POLICY_SHARED_FIRST;
 }
 
+static UCS_F_ALWAYS_INLINE int
+uct_dc_mlx5_iface_is_dcs_quota_or_hybrid(const uct_dc_mlx5_iface_t *iface)
+{
+    return (iface->tx.policy == UCT_DC_TX_POLICY_DCS_QUOTA) ||
+           (iface->tx.policy == UCT_DC_TX_POLICY_DCS_HYBRID);
```
yosefe, line 274: can we use bit operations to optimize?

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -339,11 +365,22 @@ uct_dc_mlx5_dci_pool_init_dci(uct_dc_mlx5_iface_t *iface, uint8_t pool_index,
     return UCS_OK;
 }
 
+static UCS_F_ALWAYS_INLINE void
+uct_dc_mlx5_ep_init_hw_dcs(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
+{
+    if (ucs_array_is_empty(&iface->tx.dcis)) {
+        uct_dc_mlx5_iface_resize_and_fill_dcis(iface, 1);
+        uct_dc_mlx5_dci_pool_init_dci(iface, uct_dc_mlx5_ep_pool_index(ep),
+                                      UCT_DC_MLX5_HW_DCI);
```
yosefe, line 374: does it mean HW DCS is the first DCI, that is most likely to be used?

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -14,6 +14,7 @@
 #include "dc_mlx5.h"
 
 #define UCT_DC_MLX5_EP_NO_DCI ((uint8_t)-1)
+#define UCT_DC_MLX5_HW_DCI    0
```
yosefe, line 17: UCT_DC_MLX5_HW_DCI_INDEX
since this is the array index

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -315,18 +329,30 @@ uct_dc_mlx5_init_dci_config(uct_dc_mlx5_dci_config_t *dci_config,
     dci_config->max_rd_atomic = max_rd_atomic;
 }
 
+static UCS_F_ALWAYS_INLINE int
+uct_dc_mlx5_is_hw_dci(const uct_dc_mlx5_iface_t *iface, uint8_t dci)
+{
+    return (uct_dc_mlx5_iface_is_hybrid(iface) && (dci == UCT_DC_MLX5_HW_DCI));
```
yosefe, line 335: extra ()

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -544,7 +580,8 @@ uct_dc_mlx5_iface_dci_put(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
 
     ucs_assert(dci_index != UCT_DC_MLX5_EP_NO_DCI);
 
-    if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_dci_shared(iface) ||
+        uct_dc_mlx5_is_hw_dci(iface, dci_index)) {
```
yosefe, line 584: can we avoid extra branches here? this is fast path

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -666,6 +703,29 @@ uct_dc_mlx5_iface_dci_detach(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
+
+    if (!uct_dc_mlx5_iface_is_hybrid(iface)) {
+        return UCS_ERR_NO_RESOURCE;
+    }
+
+    if (ep->dci == UCT_DC_MLX5_HW_DCI) {
+        return UCS_OK;
+    }
+
+    if (uct_dc_mlx5_iface_dci_has_tx_resources(iface, UCT_DC_MLX5_HW_DCI)) {
```
yosefe, line 719: can we try to reduce number of branches on the fast path?

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
roiedanino: It doesn't mean that, because in alloc_or_create we are allocating by the length of dcis array:
image

roiedanino:
Add hw_dci attribute to iface which will be -1 if hybrid was not selected

src/uct/ib/mlx5/dc/dc_mlx5.h
```diff
@@ -331,6 +331,9 @@ struct uct_dc_mlx5_iface {
         uint8_t                      av_fl_mlid;
 
         uint8_t                      num_dci_channels;
+
+        /* used in hybrid dcs policy otherwise -1 */
+        int8_t                       hybrid_hw_dci;
```
yosefe, line 339: are we limited to 127 DCIs? maybe make it uint8_t?

roiedanino: Maybe it should be uint16_t because I don't want this value to intersect with `UCT_DC_MLX5_EP_NO_DCI` which is a `(uint8_t) -1`

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -901,7 +903,7 @@ uct_dc_mlx5_iface_dcis_create(uct_dc_mlx5_iface_t *iface,
 
     ucs_array_length(&iface->tx.dcis) = 0;
 
-    status = uct_dc_mlx5_iface_create_dci(iface, 0, 0);
+    status = uct_dc_mlx5_iface_create_dci(iface, 0, 0, 1);
```
brminich, line 906: why just one channel and not `self->tx.num_dci_channels`?

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -266,6 +267,19 @@ uct_dc_mlx5_iface_is_dci_shared(const uct_dc_mlx5_iface_t *iface)
...
+{
+    return UCS_BIT(iface->tx.policy) & (UCS_BIT(UCT_DC_TX_POLICY_DCS_QUOTA) |
+                                        UCS_BIT(UCT_DC_TX_POLICY_DCS_HYBRID));
+}
+
+static UCS_F_ALWAYS_INLINE int
+uct_dc_mlx5_iface_is_hybrid(const uct_dc_mlx5_iface_t *iface)
+{
+    return iface->tx.policy == UCT_DC_TX_POLICY_DCS_HYBRID;
+}
```
brminich, line 281: I'd do this:
```suggestion
static UCS_F_ALWAYS_INLINE int
uct_dc_mlx5_iface_is_hybrid(const uct_dc_mlx5_iface_t *iface)
{
    return iface->tx.policy == UCT_DC_TX_POLICY_DCS_HYBRID;
}

static UCS_F_ALWAYS_INLINE int
uct_dc_mlx5_iface_is_dcs_quota_or_hybrid(const uct_dc_mlx5_iface_t *iface)
{
    return (iface->tx.policy == UCT_DC_TX_POLICY_DCS_QUOTA) ||;
            uct_dc_mlx5_iface_is_hybrid(iface);
    }

```

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -677,7 +722,8 @@ uct_dc_mlx5_iface_dci_get(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 
     ucs_assert(!iface->super.super.config.tx_moderation);
 
-    if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_dci_shared(iface) ||
+        uct_dc_mlx5_is_hw_dci(iface, ep->dci)) {
```
brminich, line 726: seems these two check are frequently used together. separate func is worth creating

brminich (commented):
can you create some DC specific gtests for this new mode?

src/uct/ib/mlx5/dc/dc_mlx5.c
roiedanino: It doesn't matter as this dci is only created to query `bb_max` and then destroyed (lines 913-914)

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
roiedanino: @yosefe wanted optimize that check as its being called in fast path

brminich: ok

src/uct/ib/mlx5/dc/dc_mlx5.c
brminich: ic, probably need to change the name of this func, as it is quite confusing currently

src/uct/ib/mlx5/dc/dc_mlx5.h
yosefe: ok

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -335,6 +335,13 @@ uct_dc_mlx5_is_hw_dci(const uct_dc_mlx5_iface_t *iface, uint8_t dci)
     return dci == iface->tx.hybrid_hw_dci;
 }
 
+static UCS_F_ALWAYS_INLINE int
+uct_dc_mlx5_is_dci_shared_or_hw(const uct_dc_mlx5_iface_t *iface, uint8_t dci)
+{
+    return uct_dc_mlx5_iface_is_dci_shared(iface) ||
+           uct_dc_mlx5_is_hw_dci(iface, dci);
```
yosefe, line 342: can we optimize this to be a single branch? it is used in uct_dc_mlx5_iface_dci_get

roiedanino: I guess replacing `||` with `+`?

yosefe: both these checks are checking the policy field, maybe we can a bitmap based check or something like "policy >= UCT_DC_TX_POLICY_DCS_HYBRID" ?

roiedanino: is_hw_dci just checks `dci == iface->tx.hybrid_hw_dci` so `>= DCS_HYBRID` won't help, we can just use bitwise `|` between is_shared and is_hw_dci if that's somehow better

src/uct/ib/mlx5/dc/dc_mlx5.h
```diff
@@ -212,6 +212,7 @@ typedef struct uct_dc_dci {
     uint8_t                       path_index; /* Path index */
     uint8_t                       next_channel_index; /* next DCI channel index
                                                          to be used by EP */
+    uint8_t                       is_shared; /* Indicates that this specific dci is shared, regardless of policy*/
```
yosefe, line 215: 1. space before *
2. line seems too long
3. maybe make it uint8_t flags?

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -366,6 +365,8 @@ uct_dc_mlx5_dci_pool_init_dci(uct_dc_mlx5_iface_t *iface, uint8_t pool_index,
         return status;
     }
 
+    dci->is_shared  = uct_dc_mlx5_iface_is_policy_shared(iface) ||
```
yosefe, line 368: set it later - initialize according to fields order in the struct

src/uct/ib/mlx5/dc/dc_mlx5.h
roiedanino: It seems like we can't avoid that extra branch because now we don't know if dci is `UCT_DC_MLX5_EP_NO_DCI`, resulting in sanitizer failure (accessing dcis[255])

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -666,18 +708,38 @@ uct_dc_mlx5_iface_dci_detach(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
-    uct_dc_dci_t *dci  = uct_dc_mlx5_iface_dci(iface, ep->dci);
+    uct_dc_dci_t *dci;
     ucs_arbiter_t *waitq;
     uct_rc_txqp_t *txqp;
     int16_t available;
 
     ucs_assert(!iface->super.super.config.tx_moderation);
 
-    if (uct_dc_mlx5_iface_is_dci_shared(iface)) {
+    if (ep->dci == UCT_DC_MLX5_EP_NO_DCI) {
```
yosefe, line 736: remove assert+comment in line 745?

yosefe: pls also update or remove the comment in that line

yosefe (approved):

PR #10141, opened 2024-09-10
UCT/CUDA: Advertise MNNVL inter-node capability with shm device type

tvegas1:
## What
Keep cuda device type as SHM and advertise MNNVL as SHM with inter-node capability to avoid `cuda_ipc` transport deactivation when specifying `UCX_NET_DEVICES`.

## Why ?
We do not need to be able to list/exclude MNNVL from/with `UCX_NET_DEVICES=^cuda` as there is already the specific configuration `ENABLE_MNNVL` to control that.

## How ?
Sanity tested, but not end-to-end tested.

### DIFF
--- a/src/uct/api/uct.h
+++ b/src/uct/api/uct.h
@@ -332,7 +332,10 @@ typedef struct uct_tl_resource_desc {
                                                 (e.g. UCT_DEVICE_TYPE_NET for a network interface) */
     ucs_sys_device_t         sys_device;   /**< The identifier associated with the device
                                                 bus_id as captured in ucs_sys_bus_id_t struct */
+    uint8_t                  flags;        /**< Associated flags to the resource */
 } uct_tl_resource_desc_t;
+
+#define UCT_TL_RESOURCE_FLAG_INTER_NODE UCS_BIT(0) /**< Inter-node capability */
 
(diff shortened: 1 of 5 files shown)


src/uct/api/uct.h
```diff
@@ -332,8 +332,11 @@ typedef struct uct_tl_resource_desc {
                                                 (e.g. UCT_DEVICE_TYPE_NET for a network interface) */
     ucs_sys_device_t         sys_device;   /**< The identifier associated with the device
                                                 bus_id as captured in ucs_sys_bus_id_t struct */
+    uint8_t                  flags;        /**< Associated flags to the resource */
 } uct_tl_resource_desc_t;
 
+#define UCT_TL_RESOURCE_FLAG_INTER_NODE UCS_BIT(0) /**< Inter-node capability */
```
iyastreb, line 338: Does it make sense to declare a enum with name, and reference that name in the `flags` field documentation?
Otherwise it's hard to understand afterwards which enum/values are supposed to be set

tvegas1:
@yosefe, for ABI compatibility, we could add something like `uct_md_query_tl_resources_v2()`, using `tl->query_devices_v2()`?

tvegas1:
Implementing with #10161 instead.

PR #10142, opened 2024-09-10
TOOLS/PERF/LIB: add hints for devices and transports in UCT perftests

roiedanino:
## What
List available devices / transports when UCT perftest cannot find the one required by the user

## Why ?
User experience and ease of use of UCT perftests.

### DIFF
--- a/src/tools/perf/lib/libperf.c
+++ b/src/tools/perf/lib/libperf.c
@@ -12,8 +12,10 @@
 #  include "config.h"
 #endif
 
-#include <ucs/debug/log.h>
 #include <ucs/arch/bitops.h>
+#include <ucs/datastruct/string_buffer.h>
+#include <ucs/datastruct/string_set.h>
+#include <ucs/debug/log.h>
 #include <ucs/sys/module.h>
 #include <ucs/sys/sock.h>
 #include <ucs/sys/string.h>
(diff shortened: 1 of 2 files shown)


src/tools/perf/lib/libperf.c
```diff
@@ -1718,8 +1720,21 @@ ucx_perf_do_warmup(ucx_perf_context_t *perf, const ucx_perf_params_t *params)
...
+    if (strlen(resource_list) > 0) {
+        strncat(resource_list, ",", max_rsc_len);
+    }
+
+    strncat(resource_list, rsc, max_rsc_len - 1);
+}
+
 static ucs_status_t uct_perf_create_md(ucx_perf_context_t *perf)
 {
+    char resources_hint[UCT_PERF_HINT_MAX_LEN] = {0};
```
yosefe, line 1736: can we use ucs_string_buffer ?

src/tools/perf/lib/libperf.c
```diff
@@ -1778,14 +1793,25 @@ static ucs_status_t uct_perf_create_md(ucx_perf_context_t *perf)
...
+                        perf->uct.md   = md;
+                        status         = UCS_OK;
+                        goto out_release_components_list;
+                    }
+                    hints_are_devices = 1;
+                    uct_perf_append_resource(resources_hint,
+                                             sizeof(resources_hint),
+                                             tl_resources[tl_index].dev_name);
+                } else if (!strcmp(perf->params.uct.dev_name,
+                                   tl_resources[tl_index].dev_name)) {
```
yosefe, line 1811: seens duplicate check with lines 1798-1799, can we refactor the code to unify?

src/tools/perf/lib/libperf.c
```diff
@@ -1796,6 +1822,15 @@ static ucs_status_t uct_perf_create_md(ucx_perf_context_t *perf)
 
     ucs_error("Cannot use "UCT_PERF_TEST_PARAMS_FMT,
               UCT_PERF_TEST_PARAMS_ARG(&perf->params));
+
+    if (strlen(resources_hint)) {
+        ucs_error("Available %s: %s",
+                  hints_are_devices ? "devices" : "transports", resources_hint);
+    } else {
+        ucs_error("Please check available transports and devices using "
```
yosefe, line 1830: maybe worth printing something here as well?
such as
Available transports: rc_mlx5, rc_verbs, ud_mlx5, ...
Available devices: mlx5_0:1, mlx5_1:1, ...

src/tools/perf/lib/libperf.c
```diff
@@ -1796,9 +1864,27 @@ static ucs_status_t uct_perf_create_md(ucx_perf_context_t *perf)
 
     ucs_error("Cannot use "UCT_PERF_TEST_PARAMS_FMT,
               UCT_PERF_TEST_PARAMS_ARG(&perf->params));
+
+    if (!is_valid_device && !is_valid_tl) {
```
yosefe, line 1868: maybe something like
if (!ucs_string_set_is_empty(devices_hint)) {
   // print available devices
} else if (!ucs_string_set_is_empty(tls_hint)) {
    // print available tls
} else {
    // print all devices and transprots
}

src/ucs/datastruct/string_set.h
```diff
@@ -23,6 +23,13 @@ KHASH_INIT(ucs_string_set, char*, char, 0, kh_str_hash_func, kh_str_hash_equal)
 typedef khash_t(ucs_string_set) ucs_string_set_t;
 
 
+/*
+ *  Check if ucs_string_set_t is empty
+ */
+#define ucs_string_set_is_empty(_string_set_ptr) \
```
yosefe, line 29: IMO better make it a regular function. could be inline for performance.

src/ucs/datastruct/string_set.h
```diff
@@ -26,8 +26,10 @@ typedef khash_t(ucs_string_set) ucs_string_set_t;
 /*
  *  Check if ucs_string_set_t is empty
  */
-#define ucs_string_set_is_empty(_string_set_ptr) \
-    (kh_size(_string_set_ptr) == 0)
+static UCS_F_ALWAYS_INLINE int ucs_string_set_is_empty(ucs_string_set_t *sset)
```
yosefe, line 29: 1. can it be const ptr?
2. usually inline functions are defined after the non-inline ones

yosefe (approved):

PR #10143, opened 2024-09-10
UCT/CUDA: Avoid crash in MPI_Finalize when CUDA context is destroyed

iyastreb:
## What
There is a crash in UCX when we destroy CUDA interfaces (cuda_ipc_iface and cuda_copy_iface) with CUDA context already closed by an application.
The goal of this fix is to avoid crash of UCX, while still printing the errors and release non-CUDA related objects

## Why ?
There is such a use case in Fortran client app:
1. CALL ACC_SHUTDOWN(ACC_DEVICE_NVIDIA)  << This essentially closes CUDA context
2. call MPI_Finalize(ierr)  << Here CUDA context is closed and we cannot release the resources
If we reverse the order of these operations then the issue disappears.

## How ?
Now we check CUDA context for validity using some dummy operation on it: `cuCtxGetApiVersion`

### DIFF
--- a/src/uct/cuda/base/cuda_iface.h
+++ b/src/uct/cuda/base/cuda_iface.h
@@ -76,10 +76,22 @@ static UCS_F_ALWAYS_INLINE int uct_cuda_base_is_context_active()
 }
 
 
+static UCS_F_ALWAYS_INLINE int uct_cuda_base_is_context_valid(CUcontext ctx)
+{
+    unsigned version;
+    ucs_status_t status;
+
+    /* Check if CUDA context is valid by running a dummy operation on it */
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuCtxGetApiVersion(ctx, &version));
+    return (status == UCS_OK);
+}
+
+
 static UCS_F_ALWAYS_INLINE int uct_cuda_base_context_match(CUcontext ctx1,
                                                            CUcontext ctx2)
 {
-    return ((ctx1 != NULL) && (ctx1 == ctx2));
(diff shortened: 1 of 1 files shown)


src/uct/cuda/base/cuda_iface.h
```diff
@@ -76,10 +76,22 @@ static UCS_F_ALWAYS_INLINE int uct_cuda_base_is_context_active()
...
+    return (status == UCS_OK);
+}
+
+
 static UCS_F_ALWAYS_INLINE int uct_cuda_base_context_match(CUcontext ctx1,
                                                            CUcontext ctx2)
 {
-    return ((ctx1 != NULL) && (ctx1 == ctx2));
+    return ((ctx1 != NULL) && (ctx1 == ctx2) &&
+            uct_cuda_base_is_context_valid(ctx1));
```
yosefe, line 94: now it will be called from multiple places, including progress, do we really want it?

iyastreb: I checked that, `uct_cuda_base_context_match` is called from 2 places:
1. UCS_CLASS_CLEANUP_FUNC (uct_cuda_ipc_iface_t, uct_cuda_copy_iface_t)
2. uct_cuda_copy_event_desc_cleanup, uct_cuda_ipc_event_desc_cleanup
These functions are invoked on memory pool chunks cleanup, when we deallocate the chunk. Normally this happens at UCX destroy. So it should not add any overhead to the hot path
Or I missed something?

yosefe: right, i mixed with other function

yosefe (approved):

PR #10144, opened 2024-09-11
UCT/IB/MLX5/DC: New env for dcis array initial capacity (defaults to max)

roiedanino:
## What
Sets dcis array initial capacity to max by default

## Why ?
Currently there are some use-after-free issues when the array resize occurs, as some functions are holding a pointer to a dcis before and after the reallocation of the array.

### DIFF
--- a/src/uct/ib/mlx5/dc/dc_mlx5.c
+++ b/src/uct/ib/mlx5/dc/dc_mlx5.c
@@ -131,6 +131,11 @@ ucs_config_field_t uct_dc_mlx5_iface_config_sub_table[] = {
      ucs_offsetof(uct_dc_mlx5_iface_config_t, num_dci_channels),
      UCS_CONFIG_TYPE_UINT},
 
+    {"DCIS_INITIAL_CAPACITY", "inf",
+     "Initial capacity of DCIs array.",
+     ucs_offsetof(uct_dc_mlx5_iface_config_t, dcis_initial_capacity),
+     UCS_CONFIG_TYPE_UINT},
+
     {NULL}
 };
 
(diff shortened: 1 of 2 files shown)


src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -131,6 +131,11 @@ ucs_config_field_t uct_dc_mlx5_iface_config_sub_table[] = {
      ucs_offsetof(uct_dc_mlx5_iface_config_t, num_dci_channels),
      UCS_CONFIG_TYPE_UINT},
 
+    {"DCIS_INITIAL_CAPACITY", "512",
```
yosefe, line 134: maybe it should be "inf" or "auto" which will actually use "ndci" value?

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -886,7 +891,8 @@ uct_dc_mlx5_iface_dcis_create(uct_dc_mlx5_iface_t *iface,
     uct_dc_dci_t *dci;
 
     ucs_array_init_dynamic(&iface->tx.dcis);
-    status = uct_dc_mlx5_iface_resize_and_fill_dcis(iface, 1);
+    status = uct_dc_mlx5_iface_resize_and_fill_dcis(
+            iface, iface->tx.dcis_initial_capacity);
```
iyastreb, line 895: Currently resize of this DCI array leads to pointer-use-after-free, which is sometimes hard to understand. So I believe we should disable this flow with FATAL. Because if we crash anyway after reallocation, better to do it explicitly
```
    ucs_array_resize_safe(&iface->tx.dcis, size, empty_dci,
                          ucs_error("%p: could not resize dcis array to %u",
                                    iface, size);
                          return UCS_ERR_NO_MEMORY, &old_buffer_p);
    if (old_buffer_p) {
        /* FIXME: explain the reason here */
        ucs_fatal("currently DCI array reallocation is not supported");
        uct_dc_mlx5_iface_dcis_array_copy(iface->tx.dcis.buffer, old_buffer_p,

```

roiedanino: How about `ucs_warn`? because some scenarios won't cause a use-after-free and aborting won't be necessary in those cases

iyastreb: Ok, I see, I was not aware of that, I thought it always crashes on non-initial reallocation
If there are cases when it does not crash, then `ucs_warn` is not needed I think, maybe `ucs_diag`

yosefe (approved):
pls squash

iyastreb (approved):

PR #10013, opened 2024-07-15
UCT/IB/MLX5: Add multi_path config to enable Adaptive Routing

tvegas1:
## What
Add adaptive routing support for `rc_x` and `dc` transports on RoCE by the use of `multi_path` and `multi_path_force` parameters.

## Why ?
Need to implement `UCX_IB_AR_ENABLE=auto/no/yes` for RoCE.

## How ?
Use PRM to add:
- HCA Cap/Cap2 multi_path force/rc/rcx/dc handling
  - force option allows overriding nvconfig/sm..
- skip setting multipath parameters upon QP creation
- on RoCE: dc: set both for DCT context and DC QP INIT2RTR transition as specified in PRM
- on RoCE: rc_x: set multipath parameters on QP INIT2RTR transition as specified in PRM

### Tested
On RoCE cluster tested UCX_IB_AR_ENABLE=auto/no/yes for rc_x and dc. Could not confirm actual throughput improvement.

PR #10149, opened 2024-09-17
DOC/FAQ: Add section for adaptive routing on RoCE

tvegas1:
## What
Document after merging #10013.

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -272,10 +272,12 @@ Yes.
 
 #### What do I need to do to run UCX with adaptive routing?
 
-When adaptive routing is configured on an Infiniband fabric, it is enabled per SL 
-(IB Service Layer).  
-Setting `UCX_IB_SL=<sl-num>` will make UCX run on the given
-service level and utilize adaptive routing. 
+Setting UCX_IB_AR_ENABLE activates adaptive routing for both InfiniBand and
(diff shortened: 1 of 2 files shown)


brminich (approved):

docs/source/faq.md
```diff
@@ -275,7 +275,8 @@ Yes.
 When adaptive routing is configured on an Infiniband fabric, it is enabled per SL 
 (IB Service Layer).  
 Setting `UCX_IB_SL=<sl-num>` will make UCX run on the given
-service level and utilize adaptive routing. 
+service level and utilize adaptive routing. On RoCE, setting `UCX_IB_AR_ENABLE`
+can activate adaptive routing without forcing any SL.
```
yosefe, line 279: Need to update the whole paragraph, can use chatgpt to rephrase:
UCX_IB_AR_ENABLE enables adataptive routing on both Infiniband and RoCE clusters. On Infiniband, it will try to select the first SL that has adaptive routing enabled, and on RoCE it will enable adaptive routing if supported by HW configuration. if UCX_IB_AR_ENABLE is set to yes, and adaptive routing is not supported by the underlying network, it will cause an error. When set to try, lack of support will be silently ignored.

tvegas1: fixed

yosefe (approved):

PR #10152, opened 2024-09-17
UCT/GGA: Initialize RoCE multipath configuration when creating interface

tvegas1:
## What
Initialize adaptive routing configuration when creating GGA interface. Relates to #10013.

## Why?
Avoid random QP modify failures on CI.

### DIFF
--- a/src/uct/ib/mlx5/gga/gga_mlx5.c
+++ b/src/uct/ib/mlx5/gga/gga_mlx5.c
@@ -496,6 +496,7 @@ static UCS_CLASS_INIT_FUNC(uct_gga_mlx5_iface_t,
             ucs_derived_of(tl_config, uct_gga_mlx5_iface_config_t);
     uct_ib_mlx5_md_t *md                = ucs_derived_of(tl_md, uct_ib_mlx5_md_t);
     uct_ib_iface_init_attr_t init_attr  = {};
+    ucs_status_t status;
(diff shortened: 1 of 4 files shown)


src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -509,6 +510,13 @@ static UCS_CLASS_INIT_FUNC(uct_gga_mlx5_iface_t,
                               &config->super.super, &config->rc_mlx5_common,
                               &init_attr);
 
+    status = uct_rc_mlx5_dp_ordering_ooo_init(
+            &self->super, UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_RC,
+            &config->rc_mlx5_common, "gga");
+    if (status != UCS_OK) {
+        return UCS_ERR_INVALID_PARAM;
```
evgeny-leksikov, line 517: why not `return status;`? 
the same to other usages - maybe `return UCS_ERR_INVALID_PARAM;` from `uct_rc_mlx5_dp_ordering_ooo_init`?

tvegas1: fixed

evgeny-leksikov (approved):

brminich (approved):

PR #10154, opened 2024-09-17
UCT/API: Add new MD resource query API

tvegas1:
## What
Introduce UCT MD transport resource descriptor query v2. Relates to #10141.

## Why ?
More parameters need to be queried and UCT ABI cannot change.

## How ?
Add all mandatory fields as parameters, use new `params` for future parameter (in and out).

### DIFF
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -1052,43 +1052,8 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
 
 /**
  * @ingroup UCT_RESOURCE
- * @brief Capability flags of @ref uct_tl_resource_desc_t.
- *
- * The enumeration defines bit mask of capabilities in @ref
- * uct_tl_resource_desc_v2_t::flags, set by @ref uct_md_query_tl_resources_v2.
- */
-typedef enum {
-    /**
-     * If set, the resource supports inter-node communications.
-     */
-    UCT_TL_RESOURCE_DESC_FLAG_INTER_NODE = UCS_BIT(0)
-} uct_md_query_tl_resources_flags_t;
-
-
-/**
- * @ingroup UCT_RESOURCE
- * @brief Parameters passed to @ref uct_md_query_tl_resources_v2.
- */
-typedef struct {
-    /**
-     * Mask of valid fields which must currently be set to zero.
-     * Future fields not specified in this mask will be ignored.
-     * Provides ABI compatibility with respect to adding new fields.
-     */
-    uint64_t field_mask;
-} uct_md_query_tl_resources_params_t;
-
-
-/**
- * @ingroup UCT_RESOURCE
  * @brief Communication resource descriptor.
  *
- * Resource descriptor is an object representing the network resource.
- * Resource descriptor could represent a stand-alone communication resource
- * such as an HCA port, network interface, or multiple resources such as
- * multiple network interfaces or communication ports. It could also represent
- * virtual communication resources that are defined over a single physical
- * network interface.
  */
 typedef struct uct_tl_resource_desc_v2 {
     /**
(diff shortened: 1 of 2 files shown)


src/uct/api/v2/uct_v2.h
```diff
@@ -1050,6 +1050,83 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
                               const uct_iface_is_reachable_params_t *params);
 
 
+/**
+ * @ingroup UCT_RESOURCE
+ * @brief Parameters passed to @ref uct_md_query_tl_resources_v2.
+ */
+typedef struct uct_md_query_tl_resources_params {
+    /**
+     * Mask of valid fields which currently has to be zero.
```
ivankochin, line 1059:
```suggestion
     * Mask of valid fields which currently have to be zero.
```

tvegas1: I am not a native english speaker, but I think "have" is for I/We/You/They. Here it is "It" (the Mask) that should be zero. Replaced by "must" anyways ;)

src/uct/api/v2/uct_v2.h
```diff
@@ -1050,6 +1050,83 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
...
+/**
+ * @ingroup UCT_RESOURCE
+ * @brief Release the list of resources returned from @ref uct_md_query_tl_resources_v2.
+ *
+ * This routine releases the memory associated with the list of resources
+ * allocated by @ref uct_md_query_tl_resources_v2.
+ *
+ * @param [in] resources  Array of resource descriptors to release.
+ */
+void uct_release_tl_resource_list_v2(uct_tl_resource_desc_v2_t *resources);
```
brminich, line 1127: do we really need it? Maybe keep using v1 version?

src/uct/api/v2/uct_v2.h
ivankochin: Yeah, I am not a native english speaker too :)

I thought that in that case the subject is "fields" which are plural and not "mask". Anyway, replacing by "must" works in both cases. Thanks!

src/uct/api/v2/uct_v2.h
tvegas1: i think we need it if type is different.

brminich: right the type changed

brminich (dismissed):

ivankochin (dismissed):

src/uct/api/v2/uct_v2.h
```diff
@@ -1050,6 +1050,83 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
...
+ * @param [in]  params          Parameters as defined in @ref
+ *                              uct_md_query_tl_resources_params_t.
+ *
+ * @return Error code.
+ */
+ucs_status_t
+uct_md_query_tl_resources_v2(uct_md_h md,
+                             uct_tl_resource_desc_v2_t **resources_p,
+                             unsigned *num_resources_p,
+                             uct_md_query_tl_resources_params_t *params);
```
rakhmets, line 1115: Copied from the PR with the implementation of this API.

I think it is better to re-arrange in and out parameters. So that you'll have in-parameters at first place, and out-parameters at the end of the parameters list. I see that you wanted to make the new interface as similar as possible to the old one. It will look better when both interfaces are used. But eventually, when only the new one remains, it will be worse.

src/uct/api/v2/uct_v2.h
```diff
@@ -1050,6 +1050,83 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
...
+ * @brief Query for transport resources.
+ *
+ * This routine queries the @ref uct_md_h "memory domain" for communication
+ * resources that are available for it.
+ *
+ * @param [in]  md              Handle to memory domain.
+ * @param [out] resources_p     Filled with a pointer to an array of resource
+ *                              descriptors.
+ * @param [out] num_resources_p Filled with the number of resources in the array.
+ * @param [in]  params          Parameters as defined in @ref
```
rakhmets, line 1106: Maybe change [in] to [inout] for `params`.

src/uct/api/v2/uct_v2.h
tvegas1: fixed

src/uct/api/v2/uct_v2.h
tvegas1: fixed

src/uct/api/v2/uct_v2.h
```diff
@@ -1050,6 +1050,83 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
...
+ * @ingroup UCT_RESOURCE
+ * @brief Parameters passed to @ref uct_md_query_tl_resources_v2.
+ */
+typedef struct uct_md_query_tl_resources_params {
+    /**
+     * Mask of valid fields which must currently be set to zero.
+     * Future fields not specified in this mask will be ignored.
+     * Provides ABI compatibility with respect to adding new fields.
+     */
+    uint64_t                       field_mask;
```
rakhmets, line 1063: This file contains different styles of alignment of the fields of structures. So, I suggest to use the same style in this PR.
```suggestion
    uint64_t field_mask;
```

tvegas1: fixed

rakhmets (dismissed):

src/uct/api/v2/uct_v2.h
```diff
@@ -1050,6 +1050,91 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
...
+ * @brief Communication resource descriptor.
+ *
+ * Resource descriptor of a standalone communication resource with extraneous
+ * flags.
+ */
+typedef struct uct_tl_resource_desc_v2 {
+    /**
+     * Main resource descriptor
+     */
+    uct_tl_resource_desc_t desc;
```
rakhmets, line 1093: Sorry, I noticed the issue only after I started to review the implementation.
V2 API shouldn't depend on V1 API. If the new structure has to replace the old one, then it is better to write the code without using the old API. Except for the old interface, which uses this structure.

tvegas1: fixed

rakhmets (dismissed):

iyastreb (dismissed):

src/uct/api/v2/uct_v2.h
```diff
@@ -1050,6 +1050,112 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
...
+ *                                uct_md_query_tl_resources_params_t.
+ * @param [out]   resources_p     Filled with a pointer to an array of resource
+ *                                descriptors.
+ * @param [out]   num_resources_p Filled with the number of resources in the array.
+ *
+ * @return Error code.
+ */
+ucs_status_t
+uct_md_query_tl_resources_v2(uct_md_h md,
+                             uct_md_query_tl_resources_params_t *params,
```
yosefe, line 1142: const?

src/uct/api/v2/uct_v2.h
```diff
@@ -1050,6 +1050,112 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
...
+ * @ingroup UCT_RESOURCE
+ * @brief Parameters passed to @ref uct_md_query_tl_resources_v2.
+ */
+typedef struct {
+    /**
+     * Mask of valid fields which must currently be set to zero.
+     * Future fields not specified in this mask will be ignored.
+     * Provides ABI compatibility with respect to adding new fields.
+     */
+    uint64_t field_mask;
```
yosefe, line 1078: do we need to define bits for it?

src/uct/api/v2/uct_v2.h
```diff
@@ -1050,6 +1050,112 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
...
+ * @brief Communication resource descriptor.
+ *
+ * Resource descriptor is an object representing the network resource.
+ * Resource descriptor could represent a stand-alone communication resource
+ * such as an HCA port, network interface, or multiple resources such as
+ * multiple network interfaces or communication ports. It could also represent
+ * virtual communication resources that are defined over a single physical
+ * network interface.
+ */
+typedef struct uct_tl_resource_desc_v2 {
```
yosefe, line 1093: how would we make this struct extendable without breaking ABI?

src/uct/api/v2/uct_v2.h
tvegas1: are we sure we never need to output any field?

src/uct/api/v2/uct_v2.h
tvegas1: currently we have none, so we expect zero value.

src/uct/api/v2/uct_v2.h
tvegas1: we can force highest alignment (or pack) and add version field first, currently expected to be zero?

tvegas1: fixed by returning it in `params`.

src/uct/api/v2/uct_v2.h
tvegas1: fixed added one bitfield for returning `resources`.

brminich (commented):
maybe implement it in a similar way to component query?

tvegas1:
#10183

PR #10155, opened 2024-09-17
UCT/CUDA: Update cuda_copy perf estimates for Grace

SeyedMir:
## What
Update cuda_copy perf estimates for Grace-Hopper

## Why ?
The bandwidth and latency values will be different for PCIe versus C2C links that connect CPU and GPU.

## How ?
Update the cuda_copy bw config and UCX_CUDA_COPY_BW.


src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -407,16 +417,17 @@ uct_cuda_copy_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr)
...
             (dst_mem_type == UCS_MEMORY_TYPE_CUDA)) {
-            perf_attr->bandwidth.shared = (zcopy ? 8300.0 : 7900.0) * UCS_MBYTE;
+            perf_attr->bandwidth.shared = zcopy ? iface->config.bw.h2d :
+                                          iface->config.bw.h2d * 0.95;
         } else if ((src_mem_type == UCS_MEMORY_TYPE_CUDA) &&
                    (dst_mem_type == UCS_MEMORY_TYPE_HOST)) {
-            perf_attr->bandwidth.shared = (zcopy ? 11660.0 : 9320.0) *
-                                          UCS_MBYTE;
+            perf_attr->bandwidth.shared = zcopy ? iface->config.bw.d2h :
+                                          iface->config.bw.d2h * 0.95;
```
ivankochin, line 425: Why bcopy BW is slower than zcopy one? BTW `11660 * 0.95` is not equal to `9320`. Maybe we need to introduce to different env variables like `BCOPY_BW` and `ZCOPY_BW` to control this values accurately. Or if we are OK with changing performance in common case, maybe better just not to distinguish bcopy/zcopy perf and set one value in both cases?

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -37,14 +37,24 @@ static ucs_config_field_t uct_cuda_copy_iface_config_table[] = {
...
+    {"BW", "h2d:8300MBs,d2h:11660MBs,d2d:320GBs,other:10000MBs",
+     "Effective memory bandwidth", 0,
+     UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_BW,
+         {"h2d", "host to device bandwidth",
+          ucs_offsetof(uct_cuda_copy_iface_config_t, bw.h2d)},
+         {"d2h", "device to host bandwidth",
+          ucs_offsetof(uct_cuda_copy_iface_config_t, bw.d2h)},
+         {"d2d", "device to device bandwidth",
+          ucs_offsetof(uct_cuda_copy_iface_config_t, bw.d2d)},
+         {"other", "any other src-dest memory types bandwidth",
```
ivankochin, line 51: Minor

```suggestion
         {"other", "any other memory types combinations bandwidth",
```

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -37,14 +37,24 @@ static ucs_config_field_t uct_cuda_copy_iface_config_table[] = {
      "Max number of cuda events. -1 is infinite",
      ucs_offsetof(uct_cuda_copy_iface_config_t, max_cuda_events), UCS_CONFIG_TYPE_UINT},
 
-    {"BW", "10000MBs",
-     "Effective memory bandwidth",
-     ucs_offsetof(uct_cuda_copy_iface_config_t, bandwidth), UCS_CONFIG_TYPE_BW},
+    /* TODO: 1. Add separate keys for shared and dedicated bandwidth
+             2. Remove the "other" key (use pref_loc for managed memory) */
+    {"BW", "h2d:8300MBs,d2h:11660MBs,d2d:320GBs,other:10000MBs",
```
ivankochin, line 42: Is that possible to remove other and define value for other memory types combinations through value without label?

```suggestion
    {"BW", "10000MBs,h2d:8300MBs,d2h:11660MBs,d2d:320GBs",
```

src/uct/cuda/cuda_copy/cuda_copy_iface.c
SeyedMir: It's actually not zcopy vs. bcopy; it's zcopy vs. short. Unlike zcopy, put/get short operations invoke `cuStreamSynchronize` per operation. Therefore, we want to advertise a slightly lower bw for the short vs zcopy operation for cuda_copy.
I'm not sure what 9320 represents. Why do you want it to be equal to 9320?

ivankochin: Thanks for the explanation. I am thinking about whether we need this difference to be made by this way because each change to performance estimation without proper performance testing can lead to unforeseen degradation in some cases. So if we want to tune performance for GH systems only I would like to leave performance for other platforms untouched. Or if we are OK to change performance on all platforms in that PR, I am wondering whether this 5% difference really matters or we can follow this KISS principle and set the same values for both zcopy and short cases.

@brminich @yosefe WDYT?

src/uct/cuda/cuda_copy/cuda_copy_iface.h
```diff
@@ -87,7 +92,12 @@ typedef struct uct_cuda_copy_iface_config {
     uct_iface_config_t      super;
     unsigned                max_poll;
     unsigned                max_cuda_events;
-    double                  bandwidth;
+    struct {
```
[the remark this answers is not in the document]
SeyedMir: I thought about doing it that way initially in fact. But, then I thought that bw matrix will have entries for types that are completely irrelevant to CUDA, and there will be much more entries than the four in this struct. Having said that, I'm not strongly against using the matrix.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -37,14 +37,24 @@ static ucs_config_field_t uct_cuda_copy_iface_config_table[] = {
...
+    {"BW", "10000MBs,h2d:8300MBs,d2h:11660MBs,d2d:320GBs",
+     "Effective memory bandwidth", 0,
+     UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_BW,
+         {"h2d", "host to device bandwidth",
+          ucs_offsetof(uct_cuda_copy_iface_config_t, bw.h2d)},
+         {"d2h", "device to host bandwidth",
+          ucs_offsetof(uct_cuda_copy_iface_config_t, bw.d2h)},
+         {"d2d", "device to device bandwidth",
+          ucs_offsetof(uct_cuda_copy_iface_config_t, bw.d2d)},
+         {"other", "any other memory types combinations bandwidth",
```
brminich, line 51: no need for `other` key - you can set default value without any key (see https://github.com/openucx/ucx/blob/master/src/uct/cuda/gdr_copy/gdr_copy_iface.c#L27)

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -37,14 +37,24 @@ static ucs_config_field_t uct_cuda_copy_iface_config_table[] = {
      "Max number of cuda events. -1 is infinite",
      ucs_offsetof(uct_cuda_copy_iface_config_t, max_cuda_events), UCS_CONFIG_TYPE_UINT},
 
-    {"BW", "10000MBs",
-     "Effective memory bandwidth",
-     ucs_offsetof(uct_cuda_copy_iface_config_t, bandwidth), UCS_CONFIG_TYPE_BW},
+    /* TODO: 1. Add separate keys for shared and dedicated bandwidth
+             2. Remove the "other" key (use pref_loc for managed memory) */
```
brminich, line 41: other can be removed

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -134,7 +144,7 @@ static ucs_status_t uct_cuda_copy_iface_query(uct_iface_h tl_iface,
 
     iface_attr->latency                 = UCT_CUDA_COPY_IFACE_LATENCY;
     iface_attr->bandwidth.dedicated     = 0;
-    iface_attr->bandwidth.shared        = iface->config.bandwidth;
+    iface_attr->bandwidth.shared        = iface->config.bw.other;
```
brminich, line 147: maybe we can use just constant here to preserve wire compatibility and avoid adding `other` key?
@yosefe, wdyt?

src/uct/cuda/cuda_copy/cuda_copy_iface.c
brminich: imo, the proposed variant is ok for simplicity

src/uct/cuda/cuda_copy/cuda_copy_iface.c
SeyedMir: @brminich I removed the `other` key and added a constant instead.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -407,16 +416,17 @@ uct_cuda_copy_estimate_perf(uct_iface_h tl_iface, uct_perf_attr_t *perf_attr)
         perf_attr->bandwidth.dedicated = 0;
         if ((src_mem_type == UCS_MEMORY_TYPE_HOST) &&
             (dst_mem_type == UCS_MEMORY_TYPE_CUDA)) {
-            perf_attr->bandwidth.shared = (zcopy ? 8300.0 : 7900.0) * UCS_MBYTE;
+            perf_attr->bandwidth.shared = zcopy ? iface->config.bw.h2d :
+                                          iface->config.bw.h2d * 0.95;
```
brminich, line 420: does 0.95 come from overhead of `cuStreamSynchronize` for short protocols?

SeyedMir: Yes

brminich (approved):

src/uct/cuda/cuda_copy/cuda_copy_iface.c
yosefe: maybe smth like
stream_sync_factor = zcopy ? 1 : 0.95;
...
perf_attr->bandwidth.shared = iface->config.bw.h2d * stream_sync_factor;

src/uct/cuda/cuda_copy/cuda_copy_iface.h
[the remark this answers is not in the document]
yosefe: maybe define a common struct type to also use it in uct_cuda_copy_iface_t?

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -37,9 +38,17 @@ static ucs_config_field_t uct_cuda_copy_iface_config_table[] = {
      "Max number of cuda events. -1 is infinite",
      ucs_offsetof(uct_cuda_copy_iface_config_t, max_cuda_events), UCS_CONFIG_TYPE_UINT},
 
-    {"BW", "10000MBs",
-     "Effective memory bandwidth",
-     ucs_offsetof(uct_cuda_copy_iface_config_t, bandwidth), UCS_CONFIG_TYPE_BW},
+    /* TODO: Add separate keys for shared and dedicated bandwidth */
+    {"BW", "h2d:8300MBs,d2h:11660MBs,d2d:320GBs",
```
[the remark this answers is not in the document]
SeyedMir: Added the `other` key back to get c2c perf for managed memory. Also updated config values.
@brminich and @yosefe please let me know what you think.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
SeyedMir: We need "other" because the existing code has a branch for each of h2d, d2h, d2d, and "else". The bw value set for h2d and "else" are not the same. My understanding is that we don't want this PR to change the already existing bw settings for non-grace platforms.
I can change it from "other" to "default", but I personally find "default" a bit confusing because the default value of the keys are set in line 42.

brminich: copying my old comment here
```
no need for other key - you can set default value without any key (see https://github.com/openucx/ucx/blob/master/src/uct/cuda/gdr_copy/gdr_copy_iface.c#L27)
```

rakhmets: @brminich
as @SeyedMir already mentioned: we need a key to set the value, that is used to initialize bw performance attribute in "else" case (not in [h2d, d2h, d2d]).

You can set default values for the existing tags. E.g. `UCX_CUDA_COPY_BW=12345` sets `12345` to all declared tags.

We have four values to configure in the existing code of performance attributes query. If we want to keep old behavior, and just extend it, then we need to add four tags to the existing configuration variable.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -491,7 +502,10 @@ static UCS_CLASS_INIT_FUNC(uct_cuda_copy_iface_t, uct_md_h md, uct_worker_h work
     self->id                     = ucs_generate_uuid((uintptr_t)self);
     self->config.max_poll        = config->max_poll;
     self->config.max_cuda_events = config->max_cuda_events;
-    self->config.bandwidth       = config->bandwidth;
+    self->config.bw.h2d          = config->bw.h2d;
+    self->config.bw.d2h          = config->bw.d2h;
+    self->config.bw.d2d          = config->bw.d2d;
+    self->config.bw.other        = config->bw.other;
```
rakhmets, line 508:
```suggestion
    self->config.bw              = config->bw;
```

rakhmets (approved):

src/uct/cuda/cuda_copy/cuda_copy_iface.c
brminich: ok with 4-th tag then. IMO either is ok (`other` or `default`)

SeyedMir:
@yosefe @brminich some tests failed and they don't seem relevant to the changes in this PR AFAICT.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
SeyedMir: Can't use `default` because it's a keyword. What else do you suggest?

SeyedMir:
@yosefe please let me know if you have more comments or I should squash.

src/uct/cuda/cuda_copy/cuda_copy_iface.c
```diff
@@ -37,14 +37,24 @@ static ucs_config_field_t uct_cuda_copy_iface_config_table[] = {
...
+    {"BW", "10000MBs,h2d:8300MBs,d2h:11660MBs,d2d:320GBs",
+     "Effective memory bandwidth", 0,
+     UCS_CONFIG_TYPE_KEY_VALUE(UCS_CONFIG_TYPE_BW,
+         {"h2d", "host to device bandwidth",
+          ucs_offsetof(uct_cuda_copy_iface_config_t, bw.h2d)},
+         {"d2h", "device to host bandwidth",
+          ucs_offsetof(uct_cuda_copy_iface_config_t, bw.d2h)},
+         {"d2d", "device to device bandwidth",
+          ucs_offsetof(uct_cuda_copy_iface_config_t, bw.d2d)},
+         {"dflt", "any other memory types combinations bandwidth",
```
yosefe, line 51: "default"

rakhmets (approved):

brminich (approved):

yosefe:
@SeyedMir pls squash

yosefe (approved):

PR #10156, opened 2024-09-18
TOOLS/INFO: reduce stack memory usage in print_type_info()

michal-shalev:
## What?
This PR reduces stack memory usage in `print_type_info()`.

## Why?
Customers with limited stack sizes have recently experienced stack overflow issues when running UCX. Moving these allocations to the heap prevents such overflows.
`print_type_info()` exceeds the threshold.

image

## How?
It splits `print_type_info()` into a few helper functions.

### DIFF
--- a/src/tools/info/type_info.c
+++ b/src/tools/info/type_info.c
@@ -87,7 +87,7 @@ static void print_size(const char *name, size_t size)
     print_size(UCS_PP_QUOTE(_type) "." UCS_PP_QUOTE(_field), \
                ucs_field_sizeof(_type, _field))
 
-void print_type_info(const char * tl_name)
+static void print_ucs(const char *tl_name)
 {
     if (tl_name == NULL) {
         printf("UCS:\n");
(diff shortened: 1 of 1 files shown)


src/tools/info/type_info.c
```diff
@@ -87,7 +87,7 @@ static void print_size(const char *name, size_t size)
     print_size(UCS_PP_QUOTE(_type) "." UCS_PP_QUOTE(_field), \
                ucs_field_sizeof(_type, _field))
 
-void print_type_info(const char * tl_name)
+static void print_ucs(const char *tl_name)
```
Artemy-Mellanox, line 90: Just curious - making `print_size()` non-inline won't solve stack issue here?

michal-shalev: It was the first thing I tried, I think it's related to the excessive macro calling

michal-shalev: To reproduce this you'll need to run the full CI here: https://github.com/openucx/ucx/pull/9963 without this commit https://github.com/openucx/ucx/pull/9963/commits/2388f0db1d1e710b3a69adab3a0526ea9e734f63
@Artemy-Mellanox

yosefe: @michal-shalev can you pls copy/paste the CI failure in print_type_info without this fix?

michal-shalev: I couldn't reproduce it, but it failed in the CI, I'll document these cases in the future.
@Artemy-Mellanox @yosefe

michal-shalev:
Reproduced in CI:
image
@Artemy-Mellanox @yosefe

michal-shalev:
@Artemy-Mellanox @yosefe @gleon99 
I'm closing this PR because we have decided to disable the GCC flag `-Wframe-larger-than=8192` when compiling with AddressSanitizer.

PR #10158, opened 2024-09-18
AZP: Retry git clone

Alexey-Rivkin:
## What
Retry 'git clone' command in case of a failure.

## Why ?
Sometimes we get a git clone error:
```
git clone --depth 1 -b v1.16.x https://github.com/openucx/ucx.git /__w/2/s/ucx_v1.16.x_88170_gpu
Cloning into '/__w/2/s/ucx_v1.16.x_88170_gpu'...
error: RPC failed; result=18, HTTP code = 200
fatal: The remote end hung up unexpectedly 
```
Switching to the built-in Azure function will prevent such failures using the retry logic.

### DIFF
--- a/buildlib/az-helpers.sh
+++ b/buildlib/az-helpers.sh
@@ -247,3 +247,23 @@ get_arch() {
         echo "$arch"
     fi
 }
+
+git_clone_with_retry() {
+    local branch="$1"
+    local target_dir="$2"
+    local depth="$3"
+    local max_attempts=5
+
+    for attempt in $(seq 1 $max_attempts); do
+        echo "Attempt $attempt of $max_attempts: Cloning UCX (branch: $branch)"
+        if git clone --depth "$depth" -b "$branch" "$BUILD_REPOSITORY_URI" "$target_dir"; then
+            echo "Clone successful"
+            return 0
+        fi
+        echo "Clone failed. Retrying in 5 seconds..."
+        sleep 5
+    done
+
+    echo "Failed to clone UCX after $max_attempts attempts"
+    return 1
+}
--- a/buildlib/pr/wire_compat.yml
+++ b/buildlib/pr/wire_compat.yml
@@ -72,8 +72,10 @@ jobs:
       - bash: chmod u+rwx $(System.DefaultWorkingDirectory)/$(ucx_current)/bin/ucx_info
       - bash: |
           set -eEx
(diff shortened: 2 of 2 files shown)


buildlib/az-helpers.sh
```diff
@@ -247,3 +247,23 @@ get_arch() {
         echo "$arch"
     fi
 }
+
+git_clone_with_retry() {
```
dpressle, line 251: fix indentation (see other functions in this file)

dpressle (changes requested):

buildlib/az-helpers.sh
Alexey-Rivkin: Changed indentation to match the existing code style.

dpressle: existing indentation uses 4 spaces you use tabs, its not the same

Alexey-Rivkin: Good catch! Fixed.

buildlib/az-helpers.sh
```diff
@@ -247,3 +247,23 @@ get_arch() {
...
+    local branch="$1"
+    local target_dir="$2"
+    local max_attempts=5
+    local repo_url="https://github.com/openucx/ucx.git"
```
yosefe, line 255: can we take it from some azure env var?

buildlib/az-helpers.sh
```diff
@@ -247,3 +247,23 @@ get_arch() {
...
+
+    for attempt in $(seq 1 $max_attempts); do
+        echo "Attempt $attempt of $max_attempts: Cloning UCX (branch: $branch)"
+        if git clone --depth 1 -b "$branch" "$repo_url" "$target_dir"; then
```
yosefe, line 259: maybe pass depth as a parameter?

buildlib/az-helpers.sh
Alexey-Rivkin: Done, thank you for the great idea!

buildlib/az-helpers.sh
Alexey-Rivkin: Done, 10x!

yosefe (approved):

yosefe:
@Alexey-Rivkin pls squash

dpressle (approved):

PR #10163, opened 2024-09-19
UCP/PROTO: Reverted short protocol disable with log level trace

shasson5:
## What
Reverted short protocol disable with log level trace

## Why ?
Prevent perf degradation

### DIFF
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -2450,7 +2450,7 @@ ucp_ep_config_max_short(ucp_context_t *context, uct_iface_attr_t *iface_attr,
     ssize_t cfg_max_short;
 
     if (!(iface_attr->cap.flags & short_flag) ||
-        context->config.progress_wrapper_enabled) {
+        ucp_context_usage_tracker_enabled(context)) {
         return -1;
     }
 


gleon99:
@shasson5 as discussed, please test it and ensure it works.

yosefe (commented):
the fix in this PR affects fast-path short threshold, but AFAIU the issue is also that short protocol is also not selected when a request is allocated.
anyway, it is ok fast-path short to be disabled when log level is set to TRACE

shasson5:
Only in proto V1 those values are used for non-fast path in addition to fast path.
For v2 it does not effect non-fast path

yosefe:
> Only in proto V1 those values are used for non-fast path in addition to fast path. For v2 it does not effect non-fast path

so why need to fix proto_select.c which is protov2?

yosefe (approved):

yosefe:
@shasson5 can you pls squash?

src/ucp/core/ucp_ep.c
```diff
@@ -2450,7 +2450,7 @@ ucp_ep_config_max_short(ucp_context_t *context, uct_iface_attr_t *iface_attr,
     ssize_t cfg_max_short;
 
     if (!(iface_attr->cap.flags & short_flag) ||
-        context->config.progress_wrapper_enabled) {
+        ucp_context_usage_tracker_enabled(context)) {
```
ivankochin, line 2453: Could you please explain why should we disable short protocol in case when `ucp_context_usage_tracker_enabled` is true?

shasson5: Because we cannot count packets which bypass non-fast path flow. (without init ucp_req)

PR #10166, opened 2024-09-22
CI: add AddressSanitizer load check in test_jenkins

michal-shalev:
## What?
Added a check to ensure AddressSanitizer is loaded before running tests in `run_asan_check()` in `test_jenkins.sh`. 

## Why?
To prevent tests from running without AddressSanitizer enabled in `run_asan_check()`, ensuring memory issues are caught during CI.

## How?
Added a check for `libasan.so` in the `run_asan_check()` function. If AddressSanitizer is not loaded, the script logs an error and exits.

### DIFF
--- a/contrib/test_jenkins.sh
+++ b/contrib/test_jenkins.sh
@@ -1131,6 +1131,13 @@ run_test_proto_disable() {
 
 run_asan_check() {
 	build devel --enable-gtest --enable-asan --without-valgrind
+
+	if ! ldd ${WORKSPACE}/build-test/test/gtest/gtest | grep -q "libasan.so"
+	then
+		azure_log_error "Error: ASan is not loaded."
+		exit 1
+	fi
+
 	run_gtest "default"
 }
 


contrib/test_jenkins.sh
```diff
@@ -1131,6 +1131,13 @@ run_test_proto_disable() {
 
 run_asan_check() {
 	build devel --enable-gtest --enable-asan --without-valgrind
+
+	if ! ldd ${ucx_inst}/src/tools/info/.libs/ucx_info | grep -q "libasan.so"
```
yosefe, line 1135: maybe we can do ldd on gtest binary?

contrib/test_jenkins.sh
```diff
@@ -1132,7 +1132,7 @@ run_test_proto_disable() {
 run_asan_check() {
 	build devel --enable-gtest --enable-asan --without-valgrind
 
-	if ! ldd ${ucx_inst}/src/tools/info/.libs/ucx_info | grep -q "libasan.so"
+	if ! ldd ${WORKSPACE}/test/gtest/gtest | grep -q "libasan.so"
```
yosefe, line 1135: i think need to check for gtest in ${ucx_inst}

contrib/test_jenkins.sh
michal-shalev: yes

contrib/test_jenkins.sh
michal-shalev: done

yosefe (dismissed):

PR #10167, opened 2024-09-22
UCT/IB/MLX5: Add devx DDP support

roiedanino:
## What
Add DDP support for RC and DC over DevX
## Why ?
Supporting OOO recv wqes out of the box.

### DIFF
--- a/src/ucs/sys/compiler_def.h
+++ b/src/ucs/sys/compiler_def.h
@@ -80,6 +80,9 @@
 /* Mask of bits 0..i-1 */
 #define UCS_MASK(_i)             (((_i) >= 64) ? ~0 : (UCS_BIT(_i) - 1))
 
+/* The i-th bit */
+#define UCS_BIT_GET(_value, _i)  (!!((_value) & UCS_BIT(_i)))
+
 /*
  * Enable compiler checks for printf-like formatting.
  *
--- a/src/uct/ib/mlx5/dc/dc_mlx5_devx.c
+++ b/src/uct/ib/mlx5/dc/dc_mlx5_devx.c
@@ -49,15 +49,14 @@ ucs_status_t uct_dc_mlx5_iface_devx_create_dct(uct_dc_mlx5_iface_t *iface)
                               ib_iface->config.max_inl_cqe[UCT_IB_DIR_RX], 1));
     UCT_IB_MLX5DV_SET(dctc, dctc, atomic_mode,
                       uct_ib_mlx5_get_atomic_mode(ib_iface));
-    if (uct_ib_iface_is_roce(&iface->super.super.super)) {
-        UCT_IB_MLX5DV_SET(dctc, dctc, dp_ordering_0,
-                          ucs_ternary_auto_value_is_yes_or_try(
-                                  ib_iface->config.dp_ordering_ooo));
-        UCT_IB_MLX5DV_SET(dctc, dctc, dp_ordering_1, 0);
-        UCT_IB_MLX5DV_SET(dctc, dctc, dp_ordering_force,
-                          ucs_ternary_auto_value_is_yes_or_no(
-                                  ib_iface->config.dp_ordering_ooo));
-    } else {
+    UCT_IB_MLX5DV_SET(dctc, dctc, dp_ordering_0,
+                      UCS_BIT_GET(iface->super.config.dp_ordering, 0));
+    UCT_IB_MLX5DV_SET(dctc, dctc, dp_ordering_1,
+                      UCS_BIT_GET(iface->super.config.dp_ordering, 1));
+    UCT_IB_MLX5DV_SET(dctc, dctc, dp_ordering_force,
+                      iface->super.config.dp_ordering_force);
+
+    if (!uct_ib_iface_is_roce(&iface->super.super.super)) {
         UCT_IB_MLX5DV_SET(dctc, dctc, pkey_index, ib_iface->pkey_index);
     }
 
@@ -145,6 +144,9 @@ ucs_status_t uct_dc_mlx5_iface_devx_dci_connect(uct_dc_mlx5_iface_t *iface,
     UCT_IB_MLX5DV_SET(qpc, qpc, atomic_mode,
                       uct_ib_mlx5_get_atomic_mode(&rc_iface->super));
     UCT_IB_MLX5DV_SET(qpc, qpc, rae, true);
+
+    uct_ib_mlx5_devx_set_qpc_dp_ordering(md, qpc, &iface->super);
+
     if (uct_ib_iface_is_roce(&rc_iface->super)) {
         UCT_IB_MLX5DV_SET(qpc, qpc, primary_address_path.eth_prio,
                           rc_iface->super.config.sl);
@@ -152,9 +154,6 @@ ucs_status_t uct_dc_mlx5_iface_devx_dci_connect(uct_dc_mlx5_iface_t *iface,
             uct_ib_mlx5_devx_set_qpc_port_affinity(md, dci_config->path_index,
                                                    qpc, &opt_param_mask);
         }
-
-        uct_ib_mlx5_devx_set_qpc_dp_ordering(
-                qpc, rc_iface->super.config.dp_ordering_ooo);
     } else {
         UCT_IB_MLX5DV_SET(qpc, qpc, primary_address_path.sl,
                           rc_iface->super.config.sl);
--- a/src/uct/ib/mlx5/dv/ib_mlx5_dv.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5_dv.c
@@ -462,16 +462,6 @@ ucs_status_t uct_ib_mlx5_devx_query_ooo_sl_mask(uct_ib_mlx5_md_t *md,
     return UCS_OK;
 }
 
-void uct_ib_mlx5_devx_set_qpc_dp_ordering(
-        void *qpc, ucs_ternary_auto_value_t dp_ordering_ooo)
-{
-    UCT_IB_MLX5DV_SET(qpc, qpc, dp_ordering_0,
-                      ucs_ternary_auto_value_is_yes_or_try(dp_ordering_ooo));
-    UCT_IB_MLX5DV_SET(qpc, qpc, dp_ordering_1, 0);
-    UCT_IB_MLX5DV_SET(qpc, qpc, dp_ordering_force,
-                      ucs_ternary_auto_value_is_yes_or_no(dp_ordering_ooo));
-}
-
 void uct_ib_mlx5_devx_set_qpc_port_affinity(uct_ib_mlx5_md_t *md,
                                             uint8_t path_index, void *qpc,
                                             uint32_t *opt_param_mask)
--- a/src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
+++ b/src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
@@ -323,7 +323,12 @@ struct uct_ib_mlx5_cmd_hca_cap_bits {
     uint8_t    log_max_transport_domain[0x5];
     uint8_t    reserved_at_328[0x3];
     uint8_t    log_max_pd[0x5];
-    uint8_t    reserved_at_330[0xb];
+    uint8_t    dp_ordering_ooo_all_ud[0x1];
+    uint8_t    dp_ordering_ooo_all_uc[0x1];
+    uint8_t    dp_ordering_ooo_all_xrc[0x1];
+    uint8_t    dp_ordering_ooo_all_dc[0x1];
+    uint8_t    dp_ordering_ooo_all_rc[0x1];
+    uint8_t    reserved_at_335[0x6];
     uint8_t    log_max_xrcd[0x5];
 
     uint8_t    nic_receive_steering_discard[0x1];
--- a/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
+++ b/src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
@@ -1895,12 +1895,20 @@ static void uct_ib_mlx5_devx_check_dp_ordering(uct_ib_mlx5_md_t *md, void *cap,
                                                void *cap_2,
                                                uct_ib_device_t *dev)
 {
-    if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_rw_rc)) {
-        md->flags |= UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_RC;
+    if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_all_rc)) {
+        md->dp_ordering_cap.rc = UCT_IB_MLX5_DP_ORDERING_OOO_ALL;
+    } else if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_rw_rc)) {
+        md->dp_ordering_cap.rc = UCT_IB_MLX5_DP_ORDERING_OOO_RW;
+    } else {
+        md->dp_ordering_cap.rc = UCT_IB_MLX5_DP_ORDERING_IBTA;
     }
 
-    if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_rw_dc)) {
-        md->flags |= UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_DC;
+    if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_all_dc)) {
+        md->dp_ordering_cap.dc = UCT_IB_MLX5_DP_ORDERING_OOO_ALL;
+    } else if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_rw_dc)) {
+        md->dp_ordering_cap.dc = UCT_IB_MLX5_DP_ORDERING_OOO_RW;
+    } else {
+        md->dp_ordering_cap.dc = UCT_IB_MLX5_DP_ORDERING_IBTA;
     }
 
     if ((cap_2 != NULL) &&
@@ -1911,8 +1919,7 @@ static void uct_ib_mlx5_devx_check_dp_ordering(uct_ib_mlx5_md_t *md, void *cap,
     ucs_debug("%s: dp_ordering support: force=%d ooo_rw_rc=%d ooo_rw_dc=%d",
               uct_ib_device_name(dev),
               !!(md->flags & UCT_IB_MLX5_MD_FLAG_DP_ORDERING_FORCE),
-              !!(md->flags & UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_RC),
-              !!(md->flags & UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_DC));
+              md->dp_ordering_cap.rc, md->dp_ordering_cap.dc);
 }
 
 static void uct_ib_mlx5_devx_check_mkey_by_name(uct_ib_mlx5_md_t *md,
@@ -3189,7 +3196,8 @@ out:
 
 static uct_ib_md_ops_t uct_ib_mlx5_md_ops;
 
-static void uct_ib_mlx5dv_query_ddp(struct ibv_context *ctx, uct_ib_mlx5_md_t *md)
+static ucs_status_t 
+uct_ib_mlx5dv_check_ddp(struct ibv_context *ctx, uct_ib_mlx5_md_t *md)
 {
 #ifdef HAVE_OOO_RECV_WRS
     struct mlx5dv_context ctx_dv = {
@@ -3200,13 +3208,21 @@ static void uct_ib_mlx5dv_query_ddp(struct ibv_context *ctx, uct_ib_mlx5_md_t *m
     ret = mlx5dv_query_device(ctx, &ctx_dv);
     if (ret != 0) {
         ucs_error("mlx5dv_query_device: Failed to query device capabilities, ret=%d\n", ret);
-        return;
+        return UCS_ERR_NO_RESOURCE;
     }
 
     if (ctx_dv.ooo_recv_wrs_caps.max_rc > 0) {
-        md->flags |= UCT_IB_MLX5_MD_FLAG_DDP;
+        md->dp_ordering_cap.rc = UCT_IB_MLX5_DP_ORDERING_OOO_ALL;
     }
+
+    if (ctx_dv.ooo_recv_wrs_caps.max_dct > 0) {
+        md->dp_ordering_cap.dc = UCT_IB_MLX5_DP_ORDERING_OOO_ALL;
+    }
+#else
+    md->dp_ordering_cap.rc = UCT_IB_MLX5_DP_ORDERING_IBTA;
+    md->dp_ordering_cap.dc = UCT_IB_MLX5_DP_ORDERING_IBTA;
 #endif
+    return UCS_OK;
 }
 
 static ucs_status_t uct_ib_mlx5dv_md_open(struct ibv_device *ibv_device,
@@ -3244,7 +3260,10 @@ static ucs_status_t uct_ib_mlx5dv_md_open(struct ibv_device *ibv_device,
         goto err_md_free;
     }
 
-    uct_ib_mlx5dv_query_ddp(ctx, md);
+    status = uct_ib_mlx5dv_check_ddp(ctx, md);
+    if (status != UCS_OK) {
+        goto err_md_free;
+    }
 
     if (IBV_DEVICE_ATOMIC_HCA(dev)) {
         dev->atomic_arg_sizes = sizeof(uint64_t);
--- a/src/uct/ib/mlx5/gga/gga_mlx5.c
+++ b/src/uct/ib/mlx5/gga/gga_mlx5.c
@@ -678,7 +678,8 @@ static UCS_CLASS_INIT_FUNC(uct_gga_mlx5_iface_t,
                               &init_attr);
 
     status = uct_rc_mlx5_dp_ordering_ooo_init(
-            &self->super, UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_RC,
+            &self->super,
+            ucs_min(md->dp_ordering_cap.rc, UCT_IB_MLX5_DP_ORDERING_OOO_RW),
             &config->rc_mlx5_common, "gga");
     if (status != UCS_OK) {
         return status;
--- a/src/uct/ib/mlx5/ib_mlx5.h
+++ b/src/uct/ib/mlx5/ib_mlx5.h
@@ -208,17 +208,11 @@ enum {
     UCT_IB_MLX5_MD_FLAG_UAR_USE_WC           = UCS_BIT(17),
     /* Device supports implicit ODP with PCI relaxed order */
     UCT_IB_MLX5_MD_FLAG_GVA_RO               = UCS_BIT(18),
-    /* RoCE supports out-of-order RDMA for RC */
-    UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_RC = UCS_BIT(19),
-    /* RoCE supports out-of-order RDMA for DC */
-    UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_DC = UCS_BIT(20),
-    /* RoCE supports forcing ordering configuration */
-    UCT_IB_MLX5_MD_FLAG_DP_ORDERING_FORCE     = UCS_BIT(21),
-    /* Device supports DDP (OOO data placement)*/
-    UCT_IB_MLX5_MD_FLAG_DDP                   = UCS_BIT(22),
+    /* Device supports forcing ordering configuration */
+    UCT_IB_MLX5_MD_FLAG_DP_ORDERING_FORCE     = UCS_BIT(19),
 
     /* Object to be created by DevX */
-    UCT_IB_MLX5_MD_FLAG_DEVX_OBJS_SHIFT  = 23,
+    UCT_IB_MLX5_MD_FLAG_DEVX_OBJS_SHIFT  = 20,
     UCT_IB_MLX5_MD_FLAG_DEVX_RC_QP       = UCT_IB_MLX5_MD_FLAG_DEVX_OBJS(RCQP),
     UCT_IB_MLX5_MD_FLAG_DEVX_RC_SRQ      = UCT_IB_MLX5_MD_FLAG_DEVX_OBJS(RCSRQ),
     UCT_IB_MLX5_MD_FLAG_DEVX_DCT         = UCT_IB_MLX5_MD_FLAG_DEVX_OBJS(DCT),
@@ -393,6 +387,16 @@ KHASH_MAP_INIT_INT(rkeys, uct_ib_mlx5_mem_lru_entry_t*);
 #endif
 
 
+typedef enum {
+    /* IBTA-compliant ordering semantics */
+    UCT_IB_MLX5_DP_ORDERING_IBTA    = 0x0,
+    /* Out-of-order RDMA reads and writes */
+    UCT_IB_MLX5_DP_ORDERING_OOO_RW  = 0x1,
+    /* Out-of-order RDMA read/write/send/recv (DDP) */
+    UCT_IB_MLX5_DP_ORDERING_OOO_ALL = 0x2,
+} uct_ib_mlx5_dp_ordering_t;
+
+
 /**
  * MLX5 IB memory domain.
  */
@@ -438,6 +442,12 @@ typedef struct uct_ib_mlx5_md {
     uint8_t                  max_rd_atomic_dc;
     uint8_t                  log_max_dci_stream_channels;
     uint32_t                 smkey_index;
+    struct {
+        /* Max dp ordering level per transport, 
+           as listed in uct_ib_mlx5_dp_ordering_t */
+        uint8_t              rc;
+        uint8_t              dc;
+    } dp_ordering_cap;
 } uct_ib_mlx5_md_t;
 
 
@@ -992,9 +1002,6 @@ uct_ib_mlx5_devx_general_cmd(struct ibv_context *context,
 ucs_status_t uct_ib_mlx5_devx_query_ooo_sl_mask(uct_ib_mlx5_md_t *md,
                                                 uint8_t port_num,
                                                 uint16_t *ooo_sl_mask_p);
-
-void uct_ib_mlx5_devx_set_qpc_dp_ordering(
-        void *qpc, ucs_ternary_auto_value_t dp_ordering_ooo);
 
 void uct_ib_mlx5_devx_set_qpc_port_affinity(uct_ib_mlx5_md_t *md,
                                             uint8_t path_index, void *qpc,
--- a/src/uct/ib/mlx5/rc/rc_mlx5_common.c
+++ b/src/uct/ib/mlx5/rc/rc_mlx5_common.c
@@ -64,12 +64,13 @@ ucs_config_field_t uct_rc_mlx5_common_config_table[] = {
    ucs_offsetof(uct_rc_mlx5_iface_common_config_t, exp_backoff),
    UCS_CONFIG_TYPE_UINT},
 
-  {"SRQ_TOPO", "cyclic,cyclic_emulated",
+  {"SRQ_TOPO", "cyclic,cyclic_emulated,list",
    "List of SRQ topology types in order of preference. Supported types are:\n"
    "\n"
    "list              SRQ is organized as a buffer containing linked list of WQEs.\n"
    "\n"
    "cyclic            SRQ is organized as a continuous array of WQEs. Requires DEVX.\n"
+   "                  cannot be used with DDP enabled.\n"
    "\n"
    "cyclic_emulated   SRQ is organized as a continuous array of WQEs, but HW\n"
    "                  treats it as a linked list. Doesn`t require DEVX.",
@@ -81,6 +82,11 @@ ucs_config_field_t uct_rc_mlx5_common_config_table[] = {
     UCS_PP_MAKE_STRING(UCT_RC_MLX5_MAX_LOG_ACK_REQ_FREQ) ".",
    ucs_offsetof(uct_rc_mlx5_iface_common_config_t, log_ack_req_freq),
    UCS_CONFIG_TYPE_UINT},
+
+  {"DDP_ENABLE", "try",
+   "Enable direct data placement\n",
+   ucs_offsetof(uct_rc_mlx5_iface_common_config_t, ddp_enable), 
+   UCS_CONFIG_TYPE_TERNARY},
 
   {NULL}
 };
@@ -596,21 +602,12 @@ void uct_rc_mlx5_release_desc(uct_recv_desc_t *self, void *desc)
 
 ucs_status_t
 uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
-                                 uint64_t tl_flag,
+                                 uct_ib_mlx5_dp_ordering_t dp_ordering_cap,
                                  uct_rc_mlx5_iface_common_config_t *config,
                                  const char *tl_name)
 {
-    uct_ib_mlx5_md_t *md = uct_ib_mlx5_iface_md(&iface->super.super);
-    int dp_ordering_ooo, dp_ordering_ooo_force;
-
-    if (!uct_ib_iface_is_roce(&iface->super.super)) {
-        iface->super.super.config.dp_ordering_ooo = UCS_AUTO;
-        return UCS_OK;
-    }
-
-    dp_ordering_ooo       = !!(md->flags & tl_flag);
-    dp_ordering_ooo_force = !!(md->flags &
-                               UCT_IB_MLX5_MD_FLAG_DP_ORDERING_FORCE);
+    uct_ib_mlx5_md_t *md      = uct_ib_mlx5_iface_md(&iface->super.super);
+    uct_ib_mlx5_dp_ordering_t min_forced_ordering;
 
     /*
      * HCA has an mlxreg admin configuration to force enable adaptive routing
@@ -618,42 +615,54 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
      *
      * HCA cap/cap_2 booleans:
      * - if dp_ordering_ooo is set, QPC/DCTC can enable AR.
-     * - if dp_ordering_ooo_force is set, QPC/DCTC can request mlxreg
+     * - if UCT_IB_MLX5_MD_FLAG_DP_ORDERING_FORCE is set, QPC/DCTC can request mlxreg
      *   configuration override, useful to force disable.
-     *
-     * QP modify behavior with returned values:
-     * - UCS_AUTO: Do not affect existing system behavior.
-     * - UCS_NO  : Force AR disabling on the QP if supported. QP modify will
-     *   return error on failure.
-     * - UCS_TRY : Set AR to enable, ignored if any failure.
-     * - UCS_YES : Force AR enabling on the QP if supported. QP modify will
-     *   return error on failure.
      */
+    iface->config.dp_ordering_force =
+            ucs_ternary_auto_value_is_yes_or_no(config->super.ar_enable) ||
+            ucs_ternary_auto_value_is_yes_or_no(config->ddp_enable);
 
-    if ((config->super.ar_enable == UCS_TRY) && dp_ordering_ooo) {
-        iface->super.super.config.dp_ordering_ooo = UCS_TRY;
-    } else if (config->super.ar_enable == UCS_NO) {
-        if (!dp_ordering_ooo_force) {
-            goto failure;
-        }
-
-        iface->super.super.config.dp_ordering_ooo = UCS_NO;
-    } else if (config->super.ar_enable == UCS_YES) {
-        if (!dp_ordering_ooo_force || !dp_ordering_ooo) {
-            goto failure;
-        }
-
-        iface->super.super.config.dp_ordering_ooo = UCS_YES;
-    } else {
-        iface->super.super.config.dp_ordering_ooo = UCS_AUTO;
+    /*
+     * Fail at the following cases:
+     * - Want to force configuration but can't force it
+     * - Want to force enable or force disable adaptive routing but it is unavailable
+     * - Want to force disable adaptive routing but force enabling DDP
+    */
+    if (iface->config.dp_ordering_force &&
+        !(md->flags & UCT_IB_MLX5_MD_FLAG_DP_ORDERING_FORCE)) {
+        goto failure;
     }
+
+    if (config->ddp_enable == UCS_YES) {
+        min_forced_ordering = UCT_IB_MLX5_DP_ORDERING_OOO_ALL;
+    } else if (config->super.ar_enable == UCS_YES) {
+        min_forced_ordering = UCT_IB_MLX5_DP_ORDERING_OOO_RW;
+    } else {
+        min_forced_ordering = UCT_IB_MLX5_DP_ORDERING_IBTA;
+    }
+
+    if (min_forced_ordering > dp_ordering_cap) {
+        goto failure;
+    }
+
+    if (config->super.ar_enable == UCS_NO) {
+        iface->config.dp_ordering = UCT_IB_MLX5_DP_ORDERING_IBTA;
+        return UCS_OK;
+    }
+
+    iface->config.dp_ordering = config->ddp_enable ?
+                                        UCT_IB_MLX5_DP_ORDERING_OOO_ALL :
+                                        UCT_IB_MLX5_DP_ORDERING_OOO_RW;
+
+    iface->config.dp_ordering = ucs_min(iface->config.dp_ordering,
+                                        dp_ordering_cap);
 
     return UCS_OK;
 
 failure:
-    ucs_error("%s: cannot set ar_enable=%d for RoCE on %s",
+    ucs_error("%s: cannot set ar_enable=%d, ddp_enable=%d, capability=%d on %s",
               uct_ib_device_name(&md->super.dev), config->super.ar_enable,
-              tl_name);
+              config->ddp_enable, dp_ordering_cap, tl_name);
     return UCS_ERR_INVALID_PARAM;
 }
 
@@ -1055,9 +1064,6 @@ void uct_rc_mlx5_common_fill_dv_qp_attr(uct_rc_mlx5_iface_common_t *iface,
                                         struct mlx5dv_qp_init_attr *dv_attr,
                                         unsigned scat2cqe_dir_mask)
 {
-    uct_ib_mlx5_md_t UCS_V_UNUSED *md = uct_ib_mlx5_iface_md(
-            &iface->super.super);
-
 #if HAVE_DECL_MLX5DV_QP_CREATE_ALLOW_SCATTER_TO_CQE
     if ((scat2cqe_dir_mask & UCS_BIT(UCT_IB_DIR_RX)) &&
         (iface->super.super.config.max_inl_cqe[UCT_IB_DIR_RX] == 0)) {
@@ -1088,7 +1094,7 @@ void uct_rc_mlx5_common_fill_dv_qp_attr(uct_rc_mlx5_iface_common_t *iface,
     }
 
 #ifdef HAVE_OOO_RECV_WRS
-    if (md->flags & UCT_IB_MLX5_MD_FLAG_DDP) {
+    if (iface->config.dp_ordering == UCT_IB_MLX5_DP_ORDERING_OOO_ALL) {
         dv_attr->create_flags |= MLX5DV_QP_CREATE_OOO_DP;
         dv_attr->comp_mask    |= MLX5DV_QP_INIT_ATTR_MASK_QP_CREATE_FLAGS;
     }
@@ -1268,4 +1274,15 @@ int uct_rc_mlx5_iface_commom_clean(uct_ib_mlx5_cq_t *mlx5_cq,
     uct_ib_mlx5_update_db_cq_ci(mlx5_cq);
 
     return nfreed;
+}
+
+void uct_ib_mlx5_devx_set_qpc_dp_ordering(uct_ib_mlx5_md_t *md, void *qpc,
+                                          uct_rc_mlx5_iface_common_t *iface)
+{
+    UCT_IB_MLX5DV_SET(qpc, qpc, dp_ordering_0,
+                      UCS_BIT_GET(iface->config.dp_ordering, 0));
+    UCT_IB_MLX5DV_SET(qpc, qpc, dp_ordering_1,
+                      UCS_BIT_GET(iface->config.dp_ordering, 1));
+    UCT_IB_MLX5DV_SET(qpc, qpc, dp_ordering_force,
+                      iface->config.dp_ordering_force);
 }
--- a/src/uct/ib/mlx5/rc/rc_mlx5_common.h
+++ b/src/uct/ib/mlx5/rc/rc_mlx5_common.h
@@ -409,6 +409,8 @@ typedef struct uct_rc_mlx5_iface_common {
         uint8_t                        atomic_fence_flag;
         uct_rc_mlx5_srq_topo_t         srq_topo;
         uint8_t                        log_ack_req_freq;
+        uint8_t                        dp_ordering;
+        uint8_t                        dp_ordering_force;
     } config;
     UCS_STATS_NODE_DECLARE(stats)
 } uct_rc_mlx5_iface_common_t;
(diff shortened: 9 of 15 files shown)


src/uct/ib/mlx5/dc/dc_mlx5_devx.c
```diff
@@ -145,21 +146,24 @@ ucs_status_t uct_dc_mlx5_iface_devx_dci_connect(uct_dc_mlx5_iface_t *iface,
     UCT_IB_MLX5DV_SET(qpc, qpc, atomic_mode,
                       uct_ib_mlx5_get_atomic_mode(&rc_iface->super));
     UCT_IB_MLX5DV_SET(qpc, qpc, rae, true);
+
```
roiedanino, line 149: delete

roiedanino:
** Add env for setting DDP on/off for debugging purposes (on by default when available)

src/uct/ib/mlx5/dc/dc_mlx5_devx.c
```diff
@@ -49,11 +49,12 @@ ucs_status_t uct_dc_mlx5_iface_devx_create_dct(uct_dc_mlx5_iface_t *iface)
                               ib_iface->config.max_inl_cqe[UCT_IB_DIR_RX], 1));
     UCT_IB_MLX5DV_SET(dctc, dctc, atomic_mode,
                       uct_ib_mlx5_get_atomic_mode(ib_iface));
-    if (uct_ib_iface_is_roce(&iface->super.super.super)) {
+    if (md->flags & UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_DC) {
```
tvegas1, line 52: Line 61, `pkey_index` must be set specifically when not on RoCE.

src/uct/ib/mlx5/dc/dc_mlx5_devx.c
```diff
@@ -49,11 +49,12 @@ ucs_status_t uct_dc_mlx5_iface_devx_create_dct(uct_dc_mlx5_iface_t *iface)
...
         UCT_IB_MLX5DV_SET(dctc, dctc, dp_ordering_0,
                           ucs_ternary_auto_value_is_yes_or_try(
                                   ib_iface->config.dp_ordering_ooo));
-        UCT_IB_MLX5DV_SET(dctc, dctc, dp_ordering_1, 0);
+        UCT_IB_MLX5DV_SET(dctc, dctc, dp_ordering_1,
```
tvegas1, line 56: values supported for the two-bits fields `dp_ordering` are 0, 1 or 2, or is there are a later PRM version?

Used: Table 3731

src/uct/ib/mlx5/dc/dc_mlx5_devx.c
```diff
@@ -152,14 +153,16 @@ ucs_status_t uct_dc_mlx5_iface_devx_dci_connect(uct_dc_mlx5_iface_t *iface,
...
         }
-
-        uct_ib_mlx5_devx_set_qpc_dp_ordering(
-                qpc, rc_iface->super.config.dp_ordering_ooo);
     } else {
         UCT_IB_MLX5DV_SET(qpc, qpc, primary_address_path.sl,
                           rc_iface->super.config.sl);
     }
 
+    if (md->flags & UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_DC) {
```
tvegas1, line 161: seems the if() check is not needed as `...super.config.dp_ordering_ooo` is already including it.

src/uct/ib/mlx5/dv/ib_mlx5_dv.c
```diff
@@ -463,11 +463,12 @@ ucs_status_t uct_ib_mlx5_devx_query_ooo_sl_mask(uct_ib_mlx5_md_t *md,
...
 
 void uct_ib_mlx5_devx_set_qpc_dp_ordering(
-        void *qpc, ucs_ternary_auto_value_t dp_ordering_ooo)
+        void *qpc, ucs_ternary_auto_value_t dp_ordering_ooo,
+        uct_ib_iface_t *iface)
 {
     UCT_IB_MLX5DV_SET(qpc, qpc, dp_ordering_0,
                       ucs_ternary_auto_value_is_yes_or_try(dp_ordering_ooo));
-    UCT_IB_MLX5DV_SET(qpc, qpc, dp_ordering_1, 0);
+    UCT_IB_MLX5DV_SET(qpc, qpc, dp_ordering_1, !uct_ib_iface_is_roce(iface));
```
tvegas1, line 471: same as above

tvegas1 (commented):
The setting of dp_ordering value is done with `UCX_IB_AR_ENABLE`. On non-RoCE, it is already performing a selection of the first SL with AR enabled. Seems this PR is bypassing this existing behavior, is that ok?

src/uct/ib/mlx5/dc/dc_mlx5_devx.c
roiedanino: I don't know if there is a later PRM but we looked here: https://lore.kernel.org/netdev/f0e05a6d-6f9c-4351-af7a-7227fb3998be@nvidia.com/T/ 
and it seems only dp_ordering_1 is relevant for our case, and it also shows x4 BW in RC

src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
```diff
@@ -324,6 +324,12 @@ struct uct_ib_mlx5_cmd_hca_cap_bits {
     uint8_t    reserved_at_328[0x3];
     uint8_t    log_max_pd[0x5];
     uint8_t    reserved_at_330[0xb];
+    uint8_t    dp_ordering_ooo_all_ud[0x1];
```
tvegas1, line 327: this seems wrong, we need to remove the line 326 `uint8_t reserved_at_330[0xb]` I think

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1880,6 +1880,18 @@ static void uct_ib_mlx5_devx_check_xgvmi(uct_ib_mlx5_md_t *md, void *cap_2,
     }
 }
 
+static ucs_status_t
+uct_ib_mlx5_ddp_check_status(uct_ib_mlx5_md_t *md,
+                             ucs_ternary_auto_value_t ddp_enable)
+{
+    if (!(md->flags &
+          (UCT_IB_MLX5_MD_FLAG_DDP_RC | UCT_IB_MLX5_MD_FLAG_DDP_DC))) {
+        return (ddp_enable == UCS_YES) ? UCS_ERR_UNSUPPORTED : UCS_OK;
```
tvegas1, line 1889: shall we use try/yes for force=0/1? forcing means failure is causing error, else ignored.

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -201,17 +201,19 @@ enum {
...
-    /* RoCE supports out-of-order RDMA for DC */
+    /* Device supports out-of-order RDMA for DC */
     UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_DC = UCS_BIT(20),
-    /* RoCE supports forcing ordering configuration */
+    /* Device supports forcing ordering configuration */
     UCT_IB_MLX5_MD_FLAG_DP_ORDERING_FORCE     = UCS_BIT(21),
-    /* Device supports DDP (OOO data placement)*/
-    UCT_IB_MLX5_MD_FLAG_DDP                   = UCS_BIT(22),
+    /* Device supports direct data placement for RC */
+    UCT_IB_MLX5_MD_FLAG_DDP_RC                = UCS_BIT(22),
```
tvegas1, line 211: maybe we should unify both namings (ddp / dp_ordering_ooo)

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
```diff
@@ -930,6 +930,9 @@ UCS_CLASS_INIT_FUNC(uct_rc_mlx5_iface_t,
         return status;
     }
 
+    uct_rc_mlx5_ddp_init_if_available(&self->super, md,
```
tvegas1, line 933: shall we also add to gga INIT_FUNC(*iface?

src/uct/ib/mlx5/rc/rc_mlx5_common.h
```diff
@@ -132,6 +132,15 @@ enum {
     UCT_RC_MLX5_CQE_APP_OP_TM_CONSUMED_MSG   = 0xA
 };
 
+enum {
+    /* IBTA-compliant ordering semantics */
+    UCT_IB_MLX5_ORDERING_IBTA,
```
tvegas1, line 137: i think it's best to put explicit values 0, 1 and 2 as they are specified in the PRM

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -408,6 +410,10 @@ typedef struct uct_ib_mlx5_md {
     uint8_t                  max_rd_atomic_dc;
     uint8_t                  log_max_dci_stream_channels;
     uint32_t                 smkey_index;
+    struct {
+        uint32_t max_rc;
```
tvegas1, line 414: maybe add comments to details what they are for

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -594,23 +594,27 @@ void uct_rc_mlx5_release_desc(uct_recv_desc_t *self, void *desc)
...
-    uct_ib_mlx5_md_t *md = uct_ib_mlx5_iface_md(&iface->super.super);
-    int dp_ordering_ooo, dp_ordering_ooo_force;
-
-    if (!uct_ib_iface_is_roce(&iface->super.super)) {
-        iface->super.super.config.dp_ordering_ooo = UCS_AUTO;
-        return UCS_OK;
-    }
+    uct_ib_mlx5_md_t *md      = uct_ib_mlx5_iface_md(&iface->super.super);
+    int dp_ordering_ooo       = !!(md->flags & tl_flag);
+    int dp_ordering_ooo_force = !!(md->flags &
```
tvegas1, line 614: i think always forcing ordering when fw supports might trigger error

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -620,32 +624,18 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
...
-        }
-
-        iface->super.super.config.dp_ordering_ooo = UCS_NO;
-    } else if (config->super.ar_enable == UCS_YES) {
-        if (!dp_ordering_ooo_force || !dp_ordering_ooo) {
-            goto failure;
-        }
+    if ((/* Want to force configuration but can't force it*/
+         ucs_ternary_auto_value_is_yes_or_no(config->super.ar_enable) &&
+         !dp_ordering_ooo_force) ||
```
tvegas1, line 631: i think we should set dp_ordering_force to zero if ar_enable is not yes/no.

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
tvegas1: or rather merge this function inside `uct_rc_mlx5_dp_ordering_ooo_init()`?

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -1088,9 +1078,15 @@ void uct_rc_mlx5_common_fill_dv_qp_attr(uct_rc_mlx5_iface_common_t *iface,
     }
 
 #ifdef HAVE_OOO_RECV_WRS
-    if (md->flags & UCT_IB_MLX5_MD_FLAG_DDP) {
+    if ((md->flags & UCT_IB_MLX5_MD_FLAG_DDP_RC) ||
+        (md->flags & UCT_IB_MLX5_MD_FLAG_DDP_DC)) {
```
tvegas1, line 1082: this might not work if we are in DC setup context, and only `(md->flags & UCT_IB_MLX5_MD_FLAG_DDP_RC)` is set. we will then set create_flags erroneously i think

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -2329,6 +2362,14 @@ ucs_status_t uct_ib_mlx5_devx_md_open(struct ibv_device *ibv_device,
 
     uct_ib_mlx5_devx_check_dp_ordering(md, cap, cap_2, dev);
 
+    if (ucs_ternary_auto_value_is_yes_or_try(md_config->ddp_enable)) {
+        status = uct_ib_mlx5_devx_check_ddp(md, cap, cap_2, dev,
```
tvegas1, line 2366: if it's "try", then we should not `goto err_lru_cleanup;` on error.

roiedanino: Its covered in: `uct_ib_mlx5_ddp_check_status`

src/uct/ib/mlx5/dv/ib_mlx5_dv.c
```diff
@@ -164,6 +164,9 @@ ucs_status_t uct_ib_mlx5_devx_create_qp(uct_ib_iface_t *iface,
 
     max_tx = uct_ib_mlx5_devx_sq_length(attr->super.cap.max_send_wr);
     len_tx = max_tx * MLX5_SEND_WQE_BB;
+    if (md->dv_ooo_recv_cap.max_rc > 0) {
+        attr->super.cap.max_recv_wr = md->dv_ooo_recv_cap.max_rc;
```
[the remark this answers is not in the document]
roiedanino: When using OOO recv we have to query the maximum qp capacity, this feature relies on that

src/uct/ib/mlx5/dv/ib_mlx5_ifc.h
roiedanino: Wow, thanks, missed that

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
roiedanino: Because this is a cx-8 feature YES => fail if there is no support for ddp at all
No won't even enter the ddp query function and won't set ddp on at all, (ddp is not enabled out of the box)

src/uct/ib/mlx5/ib_mlx5.h
roiedanino: The issue is they are entirely different capabilities we need somehow to distinguish between, they are queried differently

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
roiedanino: It is only available for rc and dc so I'm not 100% sure about this, @yosefe ?

src/uct/ib/mlx5/rc/rc_mlx5_common.h
roiedanino: Agree

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
yosefe: ddp is not relevant for gga transport

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -620,34 +624,23 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
      * - if dp_ordering_ooo is set, QPC/DCTC can enable AR.
      * - if dp_ordering_ooo_force is set, QPC/DCTC can request mlxreg
      *   configuration override, useful to force disable.
-     *
-     * QP modify behavior with returned values:
```
tvegas1, line 624: let's keep the comment reworded instead of commenting inside the if conditions

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -620,34 +624,23 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
...
+            config->super.ar_enable);
 
-        iface->super.super.config.dp_ordering_ooo = UCS_NO;
-    } else if (config->super.ar_enable == UCS_YES) {
-        if (!dp_ordering_ooo_force || !dp_ordering_ooo) {
-            goto failure;
-        }
+    if ((/* Want to force configuration but can't force it*/
+         iface->config.force_ordering && !dp_ordering_ooo_force) ||
+        /* Want to force enable of adaptive routing but its unavailable */
```
tvegas1, line 634: ...Want to force enable or force _disable_ of adaptive..

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -620,34 +624,23 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
...
+        ((config->super.ar_enable == UCS_YES) && !dp_ordering_ooo)) {
+        goto failure;
+    }
 
-        iface->super.super.config.dp_ordering_ooo = UCS_YES;
-    } else {
-        iface->super.super.config.dp_ordering_ooo = UCS_AUTO;
+    if (dp_ordering_ooo) {
```
tvegas1, line 639: i understand that `dp_ordering_ooo` is the value reported in cap by FW.

in case ar_enable=no or auto, we need to set `ordering_level` to 0 (IBTA) even if md supports ar, otherwise we cannot force disable anymore and/or ar is always requested/enabled.

src/uct/ib/base/ib_md.c
```diff
@@ -176,6 +176,11 @@ ucs_config_field_t uct_ib_md_config_table[] = {
      ucs_offsetof(uct_ib_md_config_t, ext.odp.mem_types),
      UCS_CONFIG_TYPE_BITMAP(ucs_memory_type_names)},
 
+    {"DDP_ENABLE", "try",
+     "Enable direct data placement\n",
```
tvegas1, line 180: add maybe something like: "... when AR is enabled"

src/uct/ib/mlx5/rc/rc_mlx5_common.h
```diff
@@ -132,6 +132,15 @@ enum {
     UCT_RC_MLX5_CQE_APP_OP_TM_CONSUMED_MSG   = 0xA
 };
 
+enum {
+    /* IBTA-compliant ordering semantics */
+    UCT_IB_MLX5_ORDERING_IBTA    = 0x0,
```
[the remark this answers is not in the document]
roiedanino: Going to need to use this enum in md level (moving from flags to variables)

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -1088,9 +1091,12 @@ void uct_rc_mlx5_common_fill_dv_qp_attr(uct_rc_mlx5_iface_common_t *iface,
     }
 
 #ifdef HAVE_OOO_RECV_WRS
-    if (md->flags & UCT_IB_MLX5_MD_FLAG_DDP) {
+    if (iface->config.dp_ordering == UCT_IB_MLX5_DP_ORDERING_OOO_ALL) {
         dv_attr->create_flags |= MLX5DV_QP_CREATE_OOO_DP;
         dv_attr->comp_mask    |= MLX5DV_QP_INIT_ATTR_MASK_QP_CREATE_FLAGS;
+
+        // qp_attr->cap.max_recv_wr = is_dc ? md->dv_ooo_recv_cap.max_dct :
+        //                                    md->dv_ooo_recv_cap.max_rc;
```
roiedanino, line 1099: default should be SIZE_MAX, what happens if not being set at all?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -3151,13 +3171,21 @@ static void uct_ib_mlx5dv_query_ddp(struct ibv_context *ctx, uct_ib_mlx5_md_t *m
     ret = mlx5dv_query_device(ctx, &ctx_dv);
     if (ret != 0) {
         ucs_error("mlx5dv_query_device: Failed to query device capabilities, ret=%d\n", ret);
-        return;
+        return UCS_ERR_NO_RESOURCE;
     }
 
+    // md->dv_ooo_cap.max_rc  = ctx_dv.ooo_recv_wrs_caps.max_rc;
+    // md->dv_ooo_cap.max_dct = ctx_dv.ooo_recv_wrs_caps.max_rc;
```
roiedanino, line 3178: default should be SIZE_MAX, what happens if not being set at all?

src/uct/ib/mlx5/dv/ib_mlx5_dv.c
```diff
@@ -164,6 +164,9 @@ ucs_status_t uct_ib_mlx5_devx_create_qp(uct_ib_iface_t *iface,
 
     max_tx = uct_ib_mlx5_devx_sq_length(attr->super.cap.max_send_wr);
     len_tx = max_tx * MLX5_SEND_WQE_BB;
+    // if (md->dv_ooo_recv_cap.max_rc > 0) {
+        // attr->super.cap.max_recv_wr = md->dv_ooo_recv_cap.max_rc;
+    // }
```
roiedanino, line 169: default should be SIZE_MAX, what happens if not being set at all?

src/ucs/sys/compiler_def.h
```diff
@@ -80,6 +80,9 @@
 /* Mask of bits 0..i-1 */
 #define UCS_MASK(_i)             (((_i) >= 64) ? ~0 : (UCS_BIT(_i) - 1))
 
+/* The i-th bit */
+#define UCS_GET_BIT(_value, _i)  (!!(UCS_BIT(_i) & (_value)))
```
tvegas1, line 84: maybe `UCS_BIT_IS_SET`

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1884,12 +1884,31 @@ static void uct_ib_mlx5_devx_check_dp_ordering(uct_ib_mlx5_md_t *md, void *cap,
                                                void *cap_2,
                                                uct_ib_device_t *dev)
 {
+    md->dv_ooo_cap.max_dp_ordering_rc = UCT_IB_MLX5_DP_ORDERING_IBTA;
+    md->dv_ooo_cap.max_dp_ordering_dc = UCT_IB_MLX5_DP_ORDERING_IBTA;
+
     if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_rw_rc)) {
-        md->flags |= UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_RC;
+        md->dv_ooo_cap.max_dp_ordering_rc =
```
tvegas1, line 1891: This implies that RW is always supported when ALL is supported.
Instantiate for RC/DC:
```
if (max_dp_ordering_dc == OOO_ALL) {
  goto done;
}

if (MLX5DV_GET(all_dc)) {
  max_dp_ordering_dc = OOO_ALL;
} else if (MLX5_DV_GET(rw_dc)) {
  max_dp_ordering_dc = RW_ALL;
} else {
  max_dp_ordering_dc = IBTA;
}
```
max is actually the most relaxed ordering so it could make sense to use cap/limit/mode wording instead.

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -511,7 +511,9 @@ static UCS_CLASS_INIT_FUNC(uct_gga_mlx5_iface_t,
                               &init_attr);
 
     status = uct_rc_mlx5_dp_ordering_ooo_init(
-            &self->super, UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_RC,
+            &self->super,
+            ucs_min(md->dv_ooo_cap.max_dp_ordering_rc,
```
tvegas1, line 515: `(...max_dp_ordering_rc >= OOO_RW)` to get AR enabled boolean passed as parameter?

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -620,40 +618,53 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
...
-        }
-
-        iface->super.super.config.dp_ordering_ooo = UCS_NO;
-    } else if (config->super.ar_enable == UCS_YES) {
-        if (!dp_ordering_ooo_force || !dp_ordering_ooo) {
-            goto failure;
-        }
+    iface->config.dp_ordering = UCT_IB_MLX5_DP_ORDERING_IBTA;
+
+    min_forced_ordering = (config->ddp_enable == UCS_YES) ?
```
tvegas1, line 630: requested_ordering maybe

tvegas1:
```
if (ddp_enable == YES) {
  min_forced_ordering = OOO_ALL;
} else if (ar_enable YES or TRY) {
  min_forced_ordering = OOO_RW;
} else {
  min_forced_ordering = IBTA:
}
```
and compare min_forced ordering with ordering_cap

tvegas1: and maybe put sanitisation before.

src/uct/ib/mlx5/rc/rc_mlx5_common.h
```diff
@@ -132,6 +132,7 @@ enum {
     UCT_RC_MLX5_CQE_APP_OP_TM_CONSUMED_MSG   = 0xA
 };
 
+
```
tvegas1, line 135: remove

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
roiedanino: But it is that way, DDP cannot be enabled without AR support

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -596,21 +601,14 @@ void uct_rc_mlx5_release_desc(uct_recv_desc_t *self, void *desc)
...
-    if (!uct_ib_iface_is_roce(&iface->super.super)) {
-        iface->super.super.config.dp_ordering_ooo = UCS_AUTO;
-        return UCS_OK;
-    }
-
-    dp_ordering_ooo       = !!(md->flags & tl_flag);
-    dp_ordering_ooo_force = !!(md->flags &
-                               UCT_IB_MLX5_MD_FLAG_DP_ORDERING_FORCE);
+    uct_ib_mlx5_md_t *md      = uct_ib_mlx5_iface_md(&iface->super.super);
+    int dp_ordering_ooo_force = !!(md->flags &
```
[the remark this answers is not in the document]
roiedanino: Impossible, `UCT_IB_MLX5_MD_FLAG_DP_ORDERING_FORCE` is already `UCS_BIT(19)` , the macro assumes an index

src/uct/ib/mlx5/rc/rc_mlx5_common.c
roiedanino: Min forced ordering is being used in the sanitization:
image

I think that name is less ambiguous as `requested_ordering` might be a `try`, and the condition above shouldn't fail if we set DDP=try or AR=try and they are not available

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
tvegas1: agree "This implies that RW is always supported when ALL is supported." is just side note. I still think using if/else like below could be clearer:
```
if (max_dp_ordering_dc == OOO_ALL) {
  goto done;
}

if (MLX5DV_GET(all_dc)) {
  max_dp_ordering_dc = OOO_ALL;
} else if (MLX5_DV_GET(rw_dc)) {
  max_dp_ordering_dc = RW_ALL;
} else {
  max_dp_ordering_dc = IBTA;
}
```

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1884,31 +1884,27 @@ static void uct_ib_mlx5_devx_check_dp_ordering(uct_ib_mlx5_md_t *md, void *cap,
                                                void *cap_2,
                                                uct_ib_device_t *dev)
 {
-    md->dv_ooo_cap.max_dp_ordering_rc = UCT_IB_MLX5_DP_ORDERING_IBTA;
-    md->dv_ooo_cap.max_dp_ordering_dc = UCT_IB_MLX5_DP_ORDERING_IBTA;
+    md->dp_ordering_cap.rc = UCT_IB_MLX5_DP_ORDERING_IBTA;
+    md->dp_ordering_cap.dc      = UCT_IB_MLX5_DP_ORDERING_IBTA;
```
tvegas1, line 1888: indent

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1884,31 +1884,27 @@ static void uct_ib_mlx5_devx_check_dp_ordering(uct_ib_mlx5_md_t *md, void *cap,
...
 
     if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_rw_rc)) {
-        md->dv_ooo_cap.max_dp_ordering_rc =
-                ucs_max(UCT_IB_MLX5_DP_ORDERING_OOO_RW,
-                        md->dv_ooo_cap.max_dp_ordering_rc);
+        md->dp_ordering_cap.rc = ucs_max(UCT_IB_MLX5_DP_ORDERING_OOO_RW,
```
tvegas1, line 1891: We can remove the ucs_max() calls by checking in order:
```
if (MLX5DV_GET(all_dc)) {
  md->dp_ordering_cap.dc = OOO_ALL;
} else if (MLX5_DV_GET(rw_dc)) {
  md->dp_ordering_cap.dc = RW_ALL;
} else {
  md->dp_ordering_cap.dc = IBTA;
}
```

src/uct/ib/mlx5/rc/rc_mlx5_common.h
```diff
@@ -486,11 +485,11 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
                                  const char *tl_name);
 
 #if IBV_HW_TM
-        void uct_rc_mlx5_handle_unexp_rndv(uct_rc_mlx5_iface_common_t *iface,
-                                           struct ibv_tmh *tmh, uct_tag_t tag,
-                                           struct mlx5_cqe64 *cqe,
-                                           unsigned flags, unsigned byte_len,
-                                           int poll_flags);
+void uct_rc_mlx5_handle_unexp_rndv(uct_rc_mlx5_iface_common_t *iface,
+                                    struct ibv_tmh *tmh, uct_tag_t tag,
```
tvegas1, line 489: restore

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -621,19 +621,18 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
      * 
      * Fail at the following cases:
      * - Want to force configuration but can't force it
-     * - Want to force enable or force disable of adaptive routing but its unavailable
+     * - Want to force enable or force disable adaptive routing but it is unavailable
      * - Want to force disable adaptive routing but force enabling DDP
      */
 
     iface->config.dp_ordering = UCT_IB_MLX5_DP_ORDERING_IBTA;
```
tvegas1, line 628: maybe no need to initialize here

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -621,19 +621,18 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
...
-                                  UCT_IB_MLX5_DP_ORDERING_IBTA;
-    min_forced_ordering = (config->super.ar_enable == UCS_YES) ?
-                                  ucs_max(min_forced_ordering,
-                                          UCT_IB_MLX5_DP_ORDERING_OOO_RW) :
-                                  min_forced_ordering;
+    if (config->ddp_enable == UCS_YES) {
+        min_forced_ordering = UCT_IB_MLX5_DP_ORDERING_OOO_ALL;
+    } else if (config->super.ar_enable == UCS_YES) {
+        min_forced_ordering = UCT_IB_MLX5_DP_ORDERING_OOO_RW;
+    }
```
tvegas1, line 635: add the one related sanity check here to reduce list of other conditions below?
`if (min_forced_ordering > dp_ordering_cap) {`

src/uct/ib/mlx5/rc/rc_mlx5_common.c
tvegas1: ok for naming

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -596,21 +601,14 @@ void uct_rc_mlx5_release_desc(uct_recv_desc_t *self, void *desc)
 
 ucs_status_t
 uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
-                                 uint64_t tl_flag,
+                                 uct_ib_mlx5_dp_ordering_t dp_ordering_cap,
```
tvegas1, line 604: need to make boolean as a parameter and setting OOO_RW if true in that function

test/gtest/uct/ib/test_ib.cc
```diff
@@ -439,26 +439,66 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
+            modify_config("IB_AR_ENABLE", sl_supports_ar ? "y" : "n");
         }
+
         modify_config("IB_SL", ucs::to_string(static_cast<uint16_t>(sl)));
 
-        test_uct_ib::init();
-        send_recv_short();
-        test_uct_ib::cleanup();
+        for (int test_ddp = 0; test_ddp <= has_ddp; ++test_ddp) {
+            set_ddp_enable(test_ddp && sl_supports_ar);
```
tvegas1, line 497: break if not supported?

roiedanino: Why?
The goal here is to do 2 iterations when ddp is supported: one with ddp and another one without ddp,
when ddp is not supported we want to do a single iteration - without ddp, just to test AR / different SLs

tvegas1: from what I understand when `has_dpp=1` and `sl_supports_ar=0`:
- test_ddp=0: set_ddp_enable(0); init(); recv_short(); cleanup();
- test_dpp=1: set_ddp_enable(1 && 0): recv_short(); cleanup(); 

so I think we do the same sequence twice. maybe if we use `for () {}` then check ` <= (has_ddp && sl_supports_ar)` and `set_ddp_enable(test_ddp)`?

src/uct/ib/mlx5/rc/rc_mlx5_common.c
roiedanino: I didn't understand why, as dp_ordering_cap would indicate that

test/gtest/uct/ib/test_ib.cc
```diff
@@ -439,26 +439,67 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
 };
 
 UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
+    bool has_ddp = transport_has_ddp();
+
     // go over all SLs, check UCTs could be initialized on a specific SL
     // and able to send/recv traffic
-    for (uint8_t sl = 0; sl < UCT_IB_SL_NUM; ++sl)  {
+    for (uint8_t sl = 0; sl < UCT_IB_SL_NUM; ++sl) {
+        uint16_t sl_supports_ar = (m_ooo_sl_mask & UCS_BIT(sl));
```
tvegas1, line 486: nit: `UCS_BIT_IS_SET()` ;)

src/uct/ib/mlx5/rc/rc_mlx5_common.c
tvegas1: I see that we now pass the second parameter as a boolean in the PR:
```
    status = uct_rc_mlx5_dp_ordering_ooo_init(&self->super,
                                              (md->dp_ordering_cap.rc >=
                                               UCT_IB_MLX5_DP_ORDERING_OOO_RW),
                                               &config->rc_mlx5_common, "gga");
```

So the the second argument should become boolean, and we could test it and set variables to `UCT_IB_MLX5_DP_ORDERING_OOO_RW` in this function accordingly.

Now `UCT_IB_MLX5_DP_ORDERING_OOO_RW` happens to be 1 and match the boolean value, but it is really a coincidence and we should not rely on.

roiedanino: That case is only in gga where ddp is not relevant so we are limiting it here to OOO_RW,
that's why it was `ucs_min(md->dp_ordering_cap.rc, UCT_IB_MLX5_DP_ORDERING_OOO_RW)` to make sure we are not passing OOO_ALL here

tvegas1: First it was set to `ucs_min(md->dp_ordering_cap.rc, UCT_IB_MLX5_DP_ORDERING_OOO_RW)`. Since it's only either OOO_ALL or OOO_RW, I suggested to use (... >= OOO_RW). So now it is boolean, and we need to change the prototype of the function.

roiedanino: but for `rc_mlx5` and `dc_mlx5` it can be either `IBTA` `OOO_RW` or `OOO_ALL` so I don't see how a boolean cover all of those options

roiedanino: I will revert it to `ucs_min` in gga

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -1884,12 +1884,19 @@ static void uct_ib_mlx5_devx_check_dp_ordering(uct_ib_mlx5_md_t *md, void *cap,
...
+        md->dp_ordering_cap.rc = UCT_IB_MLX5_DP_ORDERING_OOO_RW;
     }
 
-    if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_rw_dc)) {
-        md->flags |= UCT_IB_MLX5_MD_FLAG_DP_ORDERING_OOO_RW_DC;
+    if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_all_dc)) {
+        md->dp_ordering_cap.dc = UCT_IB_MLX5_DP_ORDERING_OOO_ALL;
+    } else if (UCT_IB_MLX5DV_GET(cmd_hca_cap, cap, dp_ordering_ooo_rw_dc)) {
+        md->dp_ordering_cap.dc = UCT_IB_MLX5_DP_ORDERING_OOO_RW;
     }
```
tvegas1, line 1900: maybe `else { ..ordering_cap.dc = .._IBTA; }` and don't init line 1887/1888.

src/uct/ib/mlx5/ib_mlx5.h
```diff
@@ -408,6 +412,12 @@ typedef struct uct_ib_mlx5_md {
     uint8_t                  max_rd_atomic_dc;
     uint8_t                  log_max_dci_stream_channels;
     uint32_t                 smkey_index;
+    struct {
+        /* Max dp ordering level per transport, 
+           as listed in uct_ib_mlx5_dp_ordering_t */
+        uint8_t              rc;
```
tvegas1, line 418: use `uct_ib_mlx5_dp_ordering_t` for clarity?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -3151,13 +3158,18 @@ static void uct_ib_mlx5dv_query_ddp(struct ibv_context *ctx, uct_ib_mlx5_md_t *m
     ret = mlx5dv_query_device(ctx, &ctx_dv);
     if (ret != 0) {
         ucs_error("mlx5dv_query_device: Failed to query device capabilities, ret=%d\n", ret);
-        return;
+        return UCS_ERR_NO_RESOURCE;
     }
 
     if (ctx_dv.ooo_recv_wrs_caps.max_rc > 0) {
-        md->flags |= UCT_IB_MLX5_MD_FLAG_DDP;
+        md->dp_ordering_cap.rc = UCT_IB_MLX5_DP_ORDERING_OOO_ALL;
```
tvegas1, line 3165: for my understanding: with mlx5dv md, we only support either `_OOO_ALL` or `_IBTA`?

tvegas1 (approved):
minor comments

src/uct/ib/mlx5/ib_mlx5.h
roiedanino: The issue is that in those kinds of structs, we want the struct size to be as small as possible, and enum type take up to 64 bits.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
roiedanino: right

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
tvegas1: Compiler will probably do what it wants anyways ;) and I think it should not matter for setting up MD.

src/uct/ib/mlx5/ib_mlx5.h
tvegas1: I assume we don't have size issue for MD structure as we don't have that many instances (and enum is probably 1 byte).

tvegas1 (dismissed):

roiedanino:
Critical: SRQ_TOPOLOGY has to be set to `list` if ddp is enabled

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -620,40 +621,58 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
...
+     * - Want to force enable or force disable adaptive routing but it is unavailable
+     * - Want to force disable adaptive routing but force enabling DDP
      */
+    iface->config.force_ordering = ucs_ternary_auto_value_is_yes_or_no(
+                                           config->super.ar_enable) ||
+                                   (config->ddp_enable == UCS_YES);
+
+    if ((iface->config.force_ordering && !dp_ordering_ooo_force) ||
+        ((config->super.ar_enable == UCS_NO) &&
+         (config->ddp_enable == UCS_YES))) {
```
yosefe, line 636: need a comment to explain this check, move the comment from lines 626-638 to here?

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -620,40 +621,58 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
...
+        goto failure;
+    }
 
-    if ((config->super.ar_enable == UCS_TRY) && dp_ordering_ooo) {
-        iface->super.super.config.dp_ordering_ooo = UCS_TRY;
-    } else if (config->super.ar_enable == UCS_NO) {
-        if (!dp_ordering_ooo_force) {
-            goto failure;
-        }
+    min_forced_ordering       = UCT_IB_MLX5_DP_ORDERING_IBTA;
```
yosefe, line 640: remove spaces before =, or add this line as "else" in line 646

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -620,40 +621,58 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
...
+                                        UCT_IB_MLX5_DP_ORDERING_OOO_RW;
+
+    iface->config.dp_ordering = ucs_min(iface->config.dp_ordering,
+                                        dp_ordering_cap);
+
+    ucs_assertv_always((iface->config.srq_topo !=
+                        UCT_RC_MLX5_SRQ_TOPO_CYCLIC) ||
+                               (iface->config.dp_ordering !=
+                                UCT_IB_MLX5_DP_ORDERING_OOO_ALL),
+                       "SRQ topology cyclic is not supported with DDP");
```
yosefe, line 668: can simplify this code? and convert to ucs_error

src/uct/ib/mlx5/rc/rc_mlx5_devx.c
```diff
@@ -243,6 +243,15 @@ uct_rc_mlx5_devx_init_rx_common(uct_rc_mlx5_iface_common_t *iface,
 
     iface->rx.srq.db = &iface->rx.srq.devx.dbrec->db[MLX5_RCV_DBR];
 
+    if (iface->config.srq_topo == UCT_RC_MLX5_SRQ_TOPO_AUTO) {
+        ddp_enabled            = (md->dp_ordering_cap.rc ==
+                                  UCT_IB_MLX5_DP_ORDERING_OOO_ALL) ||
+                                 (md->dp_ordering_cap.dc ==
+                                  UCT_IB_MLX5_DP_ORDERING_OOO_ALL);
```
yosefe, line 250: can we do this according to the actual transport? maybe pass the ddp_enabled value from caller?

src/uct/ib/mlx5/rc/rc_mlx5_devx.c
```diff
@@ -243,6 +243,15 @@ uct_rc_mlx5_devx_init_rx_common(uct_rc_mlx5_iface_common_t *iface,
...
+        iface->config.srq_topo = ddp_enabled ? UCT_RC_MLX5_SRQ_TOPO_LIST :
+                                               UCT_RC_MLX5_SRQ_TOPO_CYCLIC;
```
yosefe, line 252: srq_topo is currently a priority-based list. so we should say "cyclic" is not supported with DDP and move to the next priority  - instead of adding new "auto" value

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
```diff
@@ -385,14 +385,18 @@ uct_rc_mlx5_iface_parse_srq_topo(uct_ib_mlx5_md_t *md,
             return UCS_OK;
         } else if (!strcasecmp(config->srq_topo.types[i], "cyclic")) {
             /* real cyclic list requires DevX support */
-            if (!(md->flags & UCT_IB_MLX5_MD_FLAG_DEVX_RC_SRQ)) {
+            if (!(md->flags & UCT_IB_MLX5_MD_FLAG_DEVX_RC_SRQ) ||
+                (config->ddp_enable == UCS_YES)) {
```
yosefe, line 389: what if ddp_enable is "try" , and it's enabled eventually?

test/gtest/uct/ib/test_ib.cc
```diff
@@ -440,11 +440,46 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
+        int dc_has_ddp          = has_transport("dc_mlx5") &&
+                                  (ib_md->dp_ordering_cap.dc ==
+                                   UCT_IB_MLX5_DP_ORDERING_OOO_ALL);
+        return rc_has_ddp || dc_has_ddp;
+    }
+
+    void set_ddp_enable(bool enable)
+    {
+        if (has_transport("rc_mlx5")) {
+            modify_config("RC_MLX5_DDP_ENABLE", enable ? "y" : "n");
```
yosefe, line 469: 1. if we set just RC_MLX5_DDP_ENABLE, IMO it should enable for both RC and DC, because of inheritance.
2. when "enable=true", don't we want to set to "try"?

test/gtest/uct/ib/test_ib.cc
```diff
@@ -458,18 +493,23 @@ UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
...
 
-        test_uct_ib::init();
-        send_recv_short();
-        test_uct_ib::cleanup();
+        for (int test_ddp = 0; test_ddp <= (has_ddp && sl_supports_ar);
+             ++test_ddp) {
+            set_ddp_enable(test_ddp);
+            test_uct_ib::init();
+            send_recv_short();
+            test_uct_ib::cleanup();
```
yosefe, line 511: this looks weird, need to create a new test class and instantiate it with several variants

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -82,6 +85,11 @@ ucs_config_field_t uct_rc_mlx5_common_config_table[] = {
    ucs_offsetof(uct_rc_mlx5_iface_common_config_t, log_ack_req_freq),
    UCS_CONFIG_TYPE_UINT},
 
+  {"DDP_ENABLE", "try",
+   "Enable direct data placement when AR is enabled.\n",
```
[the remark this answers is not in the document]
roiedanino: It seems that the SL ooo mask has nothing to do with the DDP feature, so I'll change the description accordingly

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -3200,13 +3208,18 @@ static void uct_ib_mlx5dv_query_ddp(struct ibv_context *ctx, uct_ib_mlx5_md_t *m
...
+    }
+
+    if (ctx_dv.ooo_recv_wrs_caps.max_dct > 0) {
+        md->dp_ordering_cap.dc = UCT_IB_MLX5_DP_ORDERING_OOO_ALL;
```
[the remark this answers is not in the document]
yosefe: and where do we set these values if ctx_dv.ooo_recv_wrs_caps.max_rc==0 or ctx_dv.ooo_recv_wrs_caps.max_dct==0?

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -64,18 +64,16 @@ ucs_config_field_t uct_rc_mlx5_common_config_table[] = {
    ucs_offsetof(uct_rc_mlx5_iface_common_config_t, exp_backoff),
    UCS_CONFIG_TYPE_UINT},
 
-  {"SRQ_TOPO", "auto",
+  {"SRQ_TOPO", "cyclic,cyclic_emulated",
```
yosefe, line 67: add "list" as the least prioritized option

src/uct/ib/mlx5/rc/rc_mlx5_common.c
```diff
@@ -661,11 +657,10 @@ uct_rc_mlx5_dp_ordering_ooo_init(uct_rc_mlx5_iface_common_t *iface,
...
                                         dp_ordering_cap);
 
-    ucs_assertv_always((iface->config.srq_topo !=
-                        UCT_RC_MLX5_SRQ_TOPO_CYCLIC) ||
-                               (iface->config.dp_ordering !=
-                                UCT_IB_MLX5_DP_ORDERING_OOO_ALL),
-                       "SRQ topology cyclic is not supported with DDP");
+    if ((iface->config.srq_topo == UCT_RC_MLX5_SRQ_TOPO_CYCLIC) &&
+        (iface->config.dp_ordering == UCT_IB_MLX5_DP_ORDERING_OOO_ALL)) {
+        ucs_error("SRQ topology cyclic is not supported with DDP");
```
yosefe, line 662: need to return error code and ensure iface create function fails

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
```diff
@@ -374,7 +374,8 @@ static ucs_status_t uct_rc_mlx5_iface_tag_recv_cancel(uct_iface_h tl_iface,
 static ucs_status_t
 uct_rc_mlx5_iface_parse_srq_topo(uct_ib_mlx5_md_t *md,
                                  uct_rc_mlx5_iface_common_config_t *config,
-                                 uct_rc_mlx5_srq_topo_t *topo_p)
+                                 uct_rc_mlx5_srq_topo_t *topo_p,
+                                 uint8_t has_ddp)
```
yosefe, line 378: int has_ddp

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
```diff
@@ -421,7 +418,9 @@ static ucs_status_t uct_rc_mlx5_iface_preinit(uct_rc_mlx5_iface_common_t *iface,
     ucs_status_t status;
 
     status = uct_rc_mlx5_iface_parse_srq_topo(md, mlx5_config,
-                                              &iface->config.srq_topo);
+                                              &iface->config.srq_topo,
+                                              !!(init_attr->flags &
+                                                 UCT_IB_DDP_SUPPORTED));
```
yosefe, line 423: what if DDP is supported, but disabled by configuration? would we also fallback to cyclic_emulated (we should not)?

test/gtest/uct/ib/test_ib.cc
```diff
@@ -462,55 +462,58 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
     uint16_t m_ooo_sl_mask;
 };
 
-UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
-    bool has_ddp                 = transport_has_ddp();
-    const char *max_avail_sl_str = getenv("GTEST_MAX_IB_SL");
-    uint8_t sl, max_avail_sl;
+UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config_without_ddp)
+{
+    modify_config("RC_MLX5_DDP_ENABLE", "no");
```
yosefe, line 502:
```suggestion
UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config_without_ddp, "RC_MLX5_DDP_ENABLE?=no")
{
```

test/gtest/uct/ib/test_ib.cc
```diff
@@ -462,55 +462,58 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
-             ++test_ddp) {
-            set_ddp_enable(test_ddp);
-            test_uct_ib::init();
-            send_recv_short();
-            test_uct_ib::cleanup();
-        }
+    if (has_transport("rc_mlx5")) {
+        modify_config("RC_MLX5_DDP_ENABLE", "try");
     }
+
```
yosefe, line 515: could be something like
```c
UCS_TEST_SKIP_COND_P(test_uct_ib_sl, check_ib_sl_config_with_ddp, !has_ddp(),
                     "RC_MLX5_DDP_ENABLE?=y")
```

test/gtest/uct/ib/test_ib.cc
roiedanino:
```
[ RUN      ] dc_mlx5/test_uct_ib_sl.check_ib_sl_config_with_ddp/0 <dc_mlx5/mlx5_0:1>
[     SKIP ] (RC_MLX5_DDP_ENABLE is not a valid configuration)
```

It looks like DC is skipping, meaning we do need to use both variables

test/gtest/uct/ib/test_ib.cc
roiedanino:
```
[ RUN      ] dc_mlx5/test_uct_ib_sl.check_ib_sl_config_with_ddp/0 <dc_mlx5/mlx5_0:1>
[     SKIP ] (RC_MLX5_DDP_ENABLE is not a valid configuration)
```

It looks like DC is skipping, meaning we do need to use both variables

test/gtest/uct/ib/test_ib.cc
roiedanino: `transport_has_ddp` need to query md, which is not initialized yet at the constructor

src/uct/ib/mlx5/rc/rc_mlx5_iface.c
```diff
@@ -417,10 +416,10 @@ static ucs_status_t uct_rc_mlx5_iface_preinit(uct_rc_mlx5_iface_common_t *iface,
...
     ucs_status_t status;
 
-    status = uct_rc_mlx5_iface_parse_srq_topo(md, mlx5_config,
-                                              &iface->config.srq_topo,
-                                              !!(init_attr->flags &
-                                                 UCT_IB_DDP_SUPPORTED));
+    status = uct_rc_mlx5_iface_parse_srq_topo(
+            md, mlx5_config, &iface->config.srq_topo,
+            (init_attr->flags & UCT_IB_DDP_SUPPORTED) &&
+                    (mlx5_config->ddp_enable != UCS_NO));
```
yosefe, line 422: 1. maybe just pass mlx5_config to uct_rc_mlx5_iface_parse_srq_topo
2. i think it would be better to set srq topology according to the same logic we set iface->config.dp_ordering

test/gtest/.gdb_history
```diff
@@ -0,0 +1,40 @@
+frame 9
```
yosefe, line 1: pls remove..

yosefe (approved):

tvegas1 (approved):

PR #10168, opened 2024-09-23
UCT/IB/MLX5: correct array address grammar semantic

changchengx:
1. There's no need to get one dim array address as array member start address.
2. unify rmac & rgid start address with the access method in function uct_ib_mlx5_set_dgram_seg_grh

## What
```correct one dim array address grammar semantic usage```.
There's no function error in the original code implementation.

## Why ?
```align the rmac & rgid start address with the access method in``` [uct_ib_mlx5_set_dgram_seg_grh](https://github.com/openucx/ucx/blob/v1.17.0/src/uct/ib/mlx5/ib_mlx5.inl#L340)


changchengx:
[struct mlx5_wqe_av](https://github.com/linux-rdma/rdma-core/blob/v53.0/providers/mlx5/mlx5dv.h#L1282) is defined here in rdma-core.

yosefe (approved):

PR #10171, opened 2024-09-23
UCT/IB: Fix teardown path on memory handler callback

tvegas1:
## What
Fix error handling path on multi-threaded registration.

## Why ?
Currently we see segfault in the thread because we try to teardown the first failing registration.
More generally, a thread can fail to start or fail to properly register all its tasked `mrs`.

## How ?

- Make sure we cannot dereg twice.
- Make sure we only teardown `mrs` entries corresponding to started threads.
- Make sure started threads return with their `mrs` all initialized.

### DIFF
--- a/src/uct/ib/base/ib_md.c
+++ b/src/uct/ib/base/ib_md.c
@@ -224,6 +224,11 @@ typedef struct {
     struct ibv_mr                 **mrs;
 } uct_ib_md_mem_reg_thread_t;
 
+typedef struct {
+    uct_ib_md_mem_reg_thread_t ctx;
+    int                        mr_count;
+} uct_ib_md_mem_reg_context_t;
+
 ucs_status_t uct_ib_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
 {
     uct_ib_md_t *md              = ucs_derived_of(uct_md, uct_ib_md_t);
@@ -282,13 +287,22 @@ uct_ib_md_print_mem_reg_err_msg(const char *title, void *address, size_t length,
     ucs_log(level, "%s", ucs_string_buffer_cstr(&msg));
 }
 
+static void uct_ib_dereg_mrs(struct ibv_mr **mrs, int count)
+{
+    int i;
+
+    for (i = 0; i < count; i++) {
+        (void)uct_ib_dereg_mr(mrs[i]);
+    }
+}
+
 void *uct_ib_md_mem_handle_thread_func(void *arg)
 {
+    int mr_idx                      = 0;
     uct_ib_md_mem_reg_thread_t *ctx = arg;
     size_t chunk_size               = ctx->md->config.mt_reg_chunk;
     ucs_time_t UCS_V_UNUSED t0      = ucs_get_time();
     void UCS_V_UNUSED *start        = ctx->address;
-    int mr_idx                      = 0;
     size_t length                   = ctx->first_mr_size;
     ucs_status_t status;
 
@@ -302,7 +316,7 @@ void *uct_ib_md_mem_handle_thread_func(void *arg)
         } else {
             status = uct_ib_dereg_mr(ctx->mrs[mr_idx]);
             if (status != UCS_OK) {
-                goto err;
+                ucs_warn("failed to deregister mr_idx=%d", mr_idx);
             }
         }
         ctx->address = UCS_PTR_BYTE_OFFSET(ctx->address, length);
@@ -315,13 +329,13 @@ void *uct_ib_md_mem_handle_thread_func(void *arg)
               (ctx->params != NULL) ? "reg_mr" : "dereg_mr",
               start, ctx->address, ctx->first_mr_size,
               ucs_time_to_usec(ucs_get_time() - t0));
-    return UCS_STATUS_PTR(UCS_OK);
+    return (void*)(intptr_t)mr_idx;
 
 err_dereg:
-    for (; mr_idx >= 0; --mr_idx) {
-        uct_ib_dereg_mr(ctx->mrs[mr_idx]);
-    }
-err:
+    uct_ib_dereg_mrs(ctx->mrs, mr_idx);
+
+    ucs_assertv(UCS_STATUS_IS_ERR(status),
+                "Cannot return in-progress memory thread handling");
     return UCS_STATUS_PTR(status);
 }
 
@@ -334,7 +348,8 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
     size_t chunk_size = md->config.mt_reg_chunk;
     int thread_num_mrs, thread_num, thread_idx, mr_idx, cpu_id;
     ucs_sys_cpuset_t parent_set, thread_set;
-    uct_ib_md_mem_reg_thread_t *ctxs, *ctx;
+    uct_ib_md_mem_reg_context_t *context;
+    uct_ib_md_mem_reg_thread_t *ctx;
     char UCS_V_UNUSED affinity_str[64];
     pthread_attr_t attr;
     ucs_status_t status;
@@ -359,8 +374,8 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
               ucs_make_affinity_str(&parent_set, affinity_str,
                                     sizeof(affinity_str)));
 
-    ctxs = ucs_calloc(thread_num, sizeof(*ctxs), "ib mr ctxs");
-    if (ctxs == NULL) {
+    context = ucs_calloc(thread_num, sizeof(*context), "ib mr context");
+    if (context == NULL) {
         return UCS_ERR_NO_MEMORY;
     }
 
@@ -375,7 +390,7 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
          * get proportional amount */
         thread_num_mrs    = ucs_div_round_up(mr_num - mr_idx,
                                              thread_num - thread_idx);
-        ctx               = &ctxs[thread_idx];
+        ctx               = &context[thread_idx].ctx;
         ctx->md           = md;
         ctx->address      = UCS_PTR_BYTE_OFFSET(address, offset);
         ctx->params       = params;
@@ -389,7 +404,6 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
         ucs_assertv((ctx->address == address) || (padding == 0),
                     "thread_idx=%d address=%p padding=%zu",
                     thread_idx, address, padding);
-
         ctx->length        = (thread_num_mrs - 1) * chunk_size +
                              ctx->first_mr_size;
         ctx->length        = ucs_min(ctx->length, length - offset);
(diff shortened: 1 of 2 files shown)


src/uct/ib/base/ib_md.c
```diff
@@ -432,8 +445,10 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
 
     if (status != UCS_OK) {
         for (mr_idx = 0; mr_idx < mr_num; mr_idx++) {
-            /* coverity[check_return] */
-            uct_ib_dereg_mr(mrs[mr_idx]);
+            if (mrs[mr_idx] != NULL) {
+                /* coverity[check_return] */
+                uct_ib_dereg_mr(mrs[mr_idx]);
+            }
```
iyastreb, line 451: maybe add `mrs[mr_idx] = NULL`?

src/uct/ib/base/ib_md.c
```diff
@@ -318,9 +321,18 @@ void *uct_ib_md_mem_handle_thread_func(void *arg)
...
 err_dereg:
-    for (; mr_idx >= 0; --mr_idx) {
-        uct_ib_dereg_mr(ctx->mrs[mr_idx]);
+    for (i = 0; i < mr_idx; i++) {
+        uct_ib_dereg_mr(ctx->mrs[i]);
+        ctx->mrs[i] = NULL;
+    }
+
+    while (ctx->length > 0) {
+        ctx->mrs[mr_idx] = NULL;
```
iyastreb, line 330: Finally all `mrs` should be NULL, right?
Maybe just use memset for the whole array?

tvegas1: I am assuming that we want to avoid always setting the array to zero, else this would be simpler. @yosefe wdyt?

src/uct/ib/base/ib_md.c
tvegas1: not sure, because if we return an error in this function, all `mrs` are considered invalid anyways.

brminich (approved):

src/uct/ib/base/ib_md.c
yosefe: IMO better to not initialize all mrs to NULL when registration fails, but instead move the cleanup loop from here to lines 438-440 flow - if one thread failed, cleanup all the other threads. We can do this easier by extracting the deregistration loop to a helper function.

iyastreb (dismissed):

src/uct/ib/base/ib_md.c
```diff
@@ -288,27 +289,25 @@ void *uct_ib_md_mem_handle_thread_func(void *arg)
...
             if (status != UCS_OK) {
                 goto err_dereg;
             }
+
         } else {
-            status = uct_ib_dereg_mr(ctx->mrs[mr_idx]);
-            if (status != UCS_OK) {
-                goto err;
-            }
+            (void)uct_ib_dereg_mr(ctx->mrs[ctx->mr_idx]);
```
yosefe, line 305: maybe print a warning message?

src/uct/ib/base/ib_md.c
```diff
@@ -219,6 +219,7 @@ typedef struct {
     void                          *address;
     size_t                        length;
     size_t                        first_mr_size;
+    int                           mr_idx;
```
yosefe, line 222: why needed? I would expect the thread will not exit with a partial registration: either it return UCS_OK, or returns an error and reverts all regisrtations

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -310,4 +310,64 @@ UCS_TEST_P(test_ib_md, smkey_reg_atomic_mt, "IB_REG_MT_THRESH=1k",
...
+    size_t upper_size       = (size - (size / 2)) & ~align_mask;
+    void *start, *mid, *last;
+    uct_mem_h memh;
+
+    auto mmap_anon = [](void *ptr, size_t size) {
+        ptr = mmap(ptr, size, PROT_READ | PROT_WRITE,
+                   MAP_PRIVATE | MAP_ANONYMOUS |
+                           ((ptr != nullptr) ? MAP_FIXED : 0),
+                   -1, 0);
+        return ptr != MAP_FAILED ? ptr : nullptr;
```
yosefe, line 330: add ( )

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -310,4 +310,64 @@ UCS_TEST_P(test_ib_md, smkey_reg_atomic_mt, "IB_REG_MT_THRESH=1k",
...
 
+UCS_TEST_SKIP_COND_P(test_ib_md, mt_fail,
+                     has_ksm() &&
+                             check_invalidate_support(UCT_MD_MEM_ACCESS_RMA),
+                     "IB_REG_MT_THRESH=128K", "IB_REG_MT_CHUNK=16K")
+{
+    size_t size             = UCS_MBYTE;
+    const size_t align_mask = (8 * UCS_KBYTE) - 1;
+    size_t lower_size       = ((size / 3)) & ~align_mask;
+    size_t upper_size       = (size - (size / 2)) & ~align_mask;
```
yosefe, line 321: (size - (size / 2)) -> (size / 2)

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -310,4 +310,64 @@ UCS_TEST_P(test_ib_md, smkey_reg_atomic_mt, "IB_REG_MT_THRESH=1k",
...
+                   MAP_PRIVATE | MAP_ANONYMOUS |
+                           ((ptr != nullptr) ? MAP_FIXED : 0),
+                   -1, 0);
+        return ptr != MAP_FAILED ? ptr : nullptr;
+    };
+
+    /* Find an available VMA */
+    start = mmap_anon(nullptr, size);
+    ASSERT_NE(nullptr, start);
+    mid  = reinterpret_cast<char*>(start) + lower_size;
```
yosefe, line 336: maybe use UCS_PTR_BYTE_OFFSET?

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -310,4 +310,64 @@ UCS_TEST_P(test_ib_md, smkey_reg_atomic_mt, "IB_REG_MT_THRESH=1k",
...
+    last = reinterpret_cast<char*>(start) + size - upper_size;
+    munmap(start, size);
+
+    /* Allocate lower half and upper half */
+    /* coverity[pass_freed_arg] */
+    start = mmap_anon(start, lower_size);
+    last  = mmap_anon(last, upper_size);
+    EXPECT_NE(nullptr, start);
```
yosefe, line 344: 1. move it right after mmap_anon(start, lower_size)
2. maybe better just munmap between "start+lower_size" until "last"?
 you assume that after we munmap the block in line 338, no other thread will try to allocate in that area

src/uct/ib/base/ib_md.c
tvegas1: yes. It is for the case where there is at least one failing thread. it will then be used to teardown the other threads that had succeeded.

src/uct/ib/base/ib_md.c
tvegas1: fixed

test/gtest/uct/ib/test_ib_md.cc
tvegas1: fixed

test/gtest/uct/ib/test_ib_md.cc
tvegas1: fixed

test/gtest/uct/ib/test_ib_md.cc
tvegas1: fixed

test/gtest/uct/ib/test_ib_md.cc
tvegas1: fixed both

src/uct/ib/base/ib_md.c
yosefe: can we check this by the return value of the thread function, instead of adding this var to the context?

tvegas1: we have to store per-thread because it's only when they are all joined that we know if one failed and we need to tear them down, so I think we cannot.

pthread_join could take return code and set it in `ctx` but seems better to have the thread update it itself.

tvegas1:
@yosefe, could you please check it?

test/gtest/uct/ib/test_ib_md.cc
yosefe: unresolved

src/uct/ib/base/ib_md.c
yosefe: Right, we can wait untill all threads joined before deciding if need to clean up.
I think it's cleaner that "ctx" will only be used as input to the thread function, and not as both input and output. And the thread function can either fully succeed or fail (without "leftovers"), and the caller of the thread function will not need to check "ctx" internals in order to clean up. Similarly to what we do with regular (non-thread) functions.

test/gtest/uct/ib/test_ib_md.cc
tvegas1: now fixed

src/uct/ib/base/ib_md.c
```diff
@@ -422,22 +428,26 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
     }
 
     for (thread_idx = 0; thread_idx < thread_num; thread_idx++) {
-        ctx = &ctxs[thread_idx];
+        ctx = &context[thread_idx].ctx;
         pthread_join(ctx->thread, &thread_status);
+
         if (UCS_PTR_IS_ERR(thread_status)) {
             status = UCS_PTR_STATUS(thread_status);
+            context[thread_idx].mr_count = -1;
```
yosefe, line 436: maybe set mr_count to 0?

src/uct/ib/base/ib_md.c
```diff
@@ -295,38 +295,44 @@ void *uct_ib_md_mem_handle_thread_func(void *arg)
...
+    while (mr_idx-- > 0) {
+        (void)uct_ib_dereg_mr(ctx->mrs[mr_idx]);
     }
 
+    ucs_assertv(UCS_STATUS_IS_ERR(status),
+                "Cannot return in-progress memory thread handling");
     return UCS_STATUS_PTR(status);
 }
 
+typedef struct {
```
yosefe, line 331: IMO better move it to the beginning of this file

src/uct/ib/base/ib_md.c
```diff
@@ -420,23 +428,28 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
...
-
-    if (status != UCS_OK) {
-        for (mr_idx = 0; mr_idx < mr_num; mr_idx++) {
-            /* coverity[check_return] */
-            uct_ib_dereg_mr(mrs[mr_idx]);
+    if ((status != UCS_OK) && (params != NULL)) {
+        for (thread_idx = 0; thread_idx < thread_num; thread_idx++) {
+            for (mr_idx = 0; mr_idx < context[thread_idx].mr_count; mr_idx++) {
+                (void)uct_ib_dereg_mr(context[thread_idx].ctx.mrs[mr_idx]);
+            }
```
yosefe, line 446: can we share common code with uct_ib_md_mem_handle_thread_func cleanup flow?

tvegas1: fixed

src/uct/ib/base/ib_md.c
tvegas1: fixed

src/uct/ib/base/ib_md.c
tvegas1: fixed

src/uct/ib/base/ib_md.c
tvegas1: fixed

yosefe (approved):

PR #10172, opened 2024-09-24
UCP/CONTEXT/CONFIG: create env vars to force MULTI_SEND and FAST_CMPL

michal-shalev:
## What?
Added environment variables (`UCX_FORCE_FAST_CMPL` and `UCX_FORCE_MULTI_SEND`) to enforce the `UCP_OP_ATTR_FLAG_FAST_CMPL` and `UCP_OP_ATTR_FLAG_MULTI_SEND` flags globally for all UCX operations.

## Why?
To simplify configuration and improve performance by enabling these optimizations globally without needing to modify each operation individually.

## How?
Added `FORCE_FAST_CMPL` and `FORCE_MULTI_SEND` to `ucp_context_config_t` and updated the environment parsing logic to apply these flags automatically during protocol initialization.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -140,6 +140,13 @@ const char *ucp_object_versions[] = {
     [UCP_OBJECT_VERSION_LAST] = NULL
 };
 
+const char *ucp_extra_op_attr_flags_names[] = {
+    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_NO_IMM_CMPL)]    = "no_imm_cmpl",
+    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_FAST_CMPL)]      = "fast_cmpl",
+    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_FORCE_IMM_CMPL)] = "force_imm_cmpl",
+    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_MULTI_SEND)]     = "multi_send",
+    NULL
+};
 
 static UCS_CONFIG_DEFINE_ARRAY(memunit_sizes, sizeof(size_t),
                                UCS_CONFIG_TYPE_MEMUNITS);
@@ -530,6 +537,13 @@ static ucs_config_field_t ucp_context_config_table[] = {
   {"GVA_PREFETCH", "y",
    "Prefetch memory when using global VA MR",
    ucs_offsetof(ucp_context_config_t, gva_prefetch), UCS_CONFIG_TYPE_BOOL},
+
+  {"EXTRA_OP_ATTR_FLAGS", "",
+   "Additional send/receive operation flags that are added for each request"
+   "in addition to what is set explicitly by the user. \n"
+   "Possible values are: no_imm_cmpl, fast_cmpl, force_imm_cmpl, multi_send.",
+   ucs_offsetof(ucp_context_config_t, extra_op_attr_flags),
+   UCS_CONFIG_TYPE_BITMAP(ucp_extra_op_attr_flags_names)},
 
   {NULL}
 };
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -42,6 +42,15 @@ enum {
     UCP_TL_RSC_FLAG_AUX = UCS_BIT(0)
 };
 
+#define UCP_OP_ATTR_INDEX_MASK (UCP_OP_ATTR_FLAG_NO_IMM_CMPL    | \
+                                UCP_OP_ATTR_FLAG_FORCE_IMM_CMPL | \
+                                UCP_OP_ATTR_FLAG_FAST_CMPL      | \
+                                UCP_OP_ATTR_FLAG_MULTI_SEND)
+
+#define UCP_OP_ATTR_INDEX(_op_attr_flag) \
+    (ucs_ilog2(ucp_proto_select_op_attr_pack((_op_attr_flag), \
+                                             UCP_OP_ATTR_INDEX_MASK)))
+
 
 typedef struct ucp_context_config {
     /** Threshold for switching UCP to buffered copy(bcopy) protocol */
@@ -191,6 +200,8 @@ typedef struct ucp_context_config {
     double                                 proto_overhead_rkey_ptr;
     /** Registration cache lookup overhead estimation */
     double                                 rcache_overhead;
+    /** UCP extra operation attributes flags */
+    uint64_t                               extra_op_attr_flags;
 } ucp_context_config_t;
 
 
--- a/src/ucp/proto/proto_select.c
+++ b/src/ucp/proto/proto_select.c
@@ -456,12 +456,15 @@ ucp_proto_select_elem_init(ucp_worker_h worker, int internal,
                            const ucp_proto_select_param_t *select_param,
                            ucp_proto_select_elem_t *select_elem)
 {
+    ucp_proto_select_param_t select_param_copy = *select_param;
     UCS_STRING_BUFFER_ONSTACK(sel_param_strb, UCP_PROTO_SELECT_PARAM_STR_MAX);
     UCS_STRING_BUFFER_ONSTACK(config_name_strb, UCP_PROTO_SELECT_PARAM_STR_MAX);
     ucp_proto_select_init_protocols_t proto_init;
     ucs_status_t status;
 
-    ucp_proto_select_info_str(worker, rkey_cfg_index, select_param,
+    select_param_copy.op_attr |= worker->context->config.ext.extra_op_attr_flags;
+
+    ucp_proto_select_info_str(worker, rkey_cfg_index, &select_param_copy,
                               ucp_operation_names, &sel_param_strb);
     ucp_ep_config_name(worker, ep_cfg_index, &config_name_strb);
 
@@ -472,7 +475,7 @@ ucp_proto_select_elem_init(ucp_worker_h worker, int internal,
     ucs_log_indent(1);
 
     status = ucp_proto_select_init_protocols(worker, ep_cfg_index,
-                                             rkey_cfg_index, select_param,
+                                             rkey_cfg_index, &select_param_copy,
                                              &proto_init);
     if (status != UCS_OK) {
         goto out;
@@ -480,7 +483,7 @@ ucp_proto_select_elem_init(ucp_worker_h worker, int internal,
 
     status = ucp_proto_select_elem_init_thresh(worker, select_elem, &proto_init,
                                                ep_cfg_index, rkey_cfg_index,
-                                               select_param, internal);
+                                               &select_param_copy, internal);
     if (status != UCS_OK) {
         goto out_cleanup_proto_init;
     }
@@ -489,7 +492,7 @@ ucp_proto_select_elem_init(ucp_worker_h worker, int internal,
 
     if (!internal) {
         ucp_proto_select_elem_trace(worker, ep_cfg_index, rkey_cfg_index,
-                                    select_param, select_elem);
+                                    &select_param_copy, select_elem);
     }
 
     status = UCS_OK;
--- a/src/ucp/proto/proto_select.h
+++ b/src/ucp/proto/proto_select.h
@@ -43,6 +43,11 @@
 #define UCP_PROTO_SELECT_PARAM_STR_MAX 128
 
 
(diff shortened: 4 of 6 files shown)


src/ucp/core/ucp_context.c
```diff
@@ -140,6 +140,11 @@ const char *ucp_object_versions[] = {
     [UCP_OBJECT_VERSION_LAST] = NULL
 };
 
+static const char *ucp_force_flags_names[] = {
+    [UCP_FORCE_FLAG_MULTI_SEND] = "multi_send",
+    [UCP_FORCE_FLAG_FAST_CMPL]  = "fast_cmpl",
```
ivankochin, line 145:
```suggestion
    [UCP_OP_ATTR_FLAG_MULTI_SEND] = "UCP_OP_ATTR_FLAG_MULTI_SEND",
    [UCP_OP_ATTR_FLAG_FAST_CMPL]  = "UCP_OP_ATTR_FLAG_FAST_CMPL",
```

1. Is that possible to use flags as indexes for names directly? If yes then code in `proto_select.c` can look like:
```c
    ucp_proto_select_param_t select_param_copy = *select_param;
    select_param_copy.op_attr |= worker->context->config.ext.force_request_flags;
```

2. I am not sure that we need to shorten flags' names here. WDYT about setting names for flags to their full version?

src/ucp/core/ucp_context.c
```diff
@@ -522,6 +527,11 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Prefetch memory when using global VA MR",
    ucs_offsetof(ucp_context_config_t, gva_prefetch), UCS_CONFIG_TYPE_BOOL},
 
+  {"FORCE_REQUEST_FLAGS", "none",
```
ivankochin, line 530: Have you tested it manually, it seems like there should be empty default value instead of "none". 

```suggestion
  {"FORCE_REQUEST_FLAGS", "",
```

Also I tried to set `UCX_FORCE_REQUEST_FLAGS=fast_cmpl` and get the following issue:

image

src/ucp/proto/proto_select.c
```diff
@@ -461,7 +461,17 @@ ucp_proto_select_elem_init(ucp_worker_h worker, int internal,
     ucp_proto_select_init_protocols_t proto_init;
     ucs_status_t status;
 
-    ucp_proto_select_info_str(worker, rkey_cfg_index, select_param,
+    ucp_proto_select_param_t select_param_copy = *select_param;
```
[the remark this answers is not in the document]
michal-shalev: done

src/ucp/proto/proto_select.c
```diff
@@ -461,7 +461,17 @@ ucp_proto_select_elem_init(ucp_worker_h worker, int internal,
...
+
+    if (worker->context->config.ext.force_request_flags & UCP_FORCE_FLAG_MULTI_SEND) {
```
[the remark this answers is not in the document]
michal-shalev: Added a helper function

src/ucp/core/ucp_context.h
```diff
@@ -189,6 +196,9 @@ typedef struct ucp_context_config {
     double                                 proto_overhead_rkey_ptr;
     /** Registration cache lookup overhead estimation */
     double                                 rcache_overhead;
+    /** Bitmask for forced request flags */
+    uint32_t force_request_flags;
```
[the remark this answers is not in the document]
michal-shalev: done

src/ucp/core/ucp_context.c
michal-shalev: Changed to "auto"
image

src/ucp/core/ucp_context.c
michal-shalev: 1. `op_attr` in `ucp_proto_select_param` is `uint8_t`, while `ucp_op_attr_t` contains `UCP_OP_ATTR_FLAG_MULTI_SEND=UCS_BIT(19)` meaning I have to use unpack, unless I'm missing something, please check the fixes and LMKWDYT.
2. kept the short names

yosefe: we can define ucp_proto_select_op_attr_pack() as a macro, then use something like
```c
[ucs_ilog2(ucp_proto_select_op_attr_pack(UCP_OP_ATTR_FLAG_MULTI_SEND))] = "multi_send",
...
```

we can also define a macro to express `ucs_ilog2(ucp_proto_select_op_attr_pack(_value))` to shorten the code.

This way the flags will be defined with their "packed" values.
so we could just add (bitwise or) the force_op_attr_flags value into select_param->op_attr

src/ucp/proto/proto_select.c
```diff
@@ -449,27 +449,37 @@ ucp_proto_select_wiface_activate(ucp_worker_h worker,
     ep_config->proto_lane_map |= lane_map;
 }
 
+static void apply_force_op_attr_flags(ucp_proto_select_param_t *select_param,
```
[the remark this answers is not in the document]
michal-shalev: No need for a helper function after all

src/ucp/core/ucp_context.c
michal-shalev: please check now @yosefe

ivankochin:
I was thinking about adding some functionality to this PR, probably it would be useful not only to force using some flag but also remove flag if one is defined. I would propose syntax that is used for `UCX_TLS`, I mean we can use `^` symbol to force removing some flags. WDYT?

michal-shalev:
@ivankochin @yosefe 
Added "no_imm_cmpl" and "force_imm_cmpl" as well, ATM the 4 flags are working (checked with GDB), but when I tried to pass the wrong flag name it causes a core dump:
image

src/ucp/core/ucp_context.c
```diff
@@ -140,6 +140,13 @@ const char *ucp_object_versions[] = {
     [UCP_OBJECT_VERSION_LAST] = NULL
 };
 
+const char *ucp_force_flags_names[] = {
+    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_NO_IMM_CMPL)] = "no_imm_cmpl",
```
yosefe, line 144: align = on column

src/ucp/core/ucp_context.c
```diff
@@ -522,6 +529,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Prefetch memory when using global VA MR",
    ucs_offsetof(ucp_context_config_t, gva_prefetch), UCS_CONFIG_TYPE_BOOL},
 
+  {"FORCE_REQUEST_FLAGS", "",
+   "Defines forced request flags. Possible flags are: no_imm_cmpl, fast_cmpl, "
+   "force_imm_cmpl, multi_send",
```
yosefe, line 534: I think UCS_CONFIG_TYPE_BITMAP already prints the list of possible values

src/ucp/proto/proto_select.h
```diff
@@ -19,7 +19,9 @@
  * selection decision.
  */
 #define UCP_PROTO_SELECT_OP_ATTR_BASE   UCP_OP_ATTR_FLAG_NO_IMM_CMPL
-#define UCP_PROTO_SELECT_OP_ATTR_MASK   (UCP_OP_ATTR_FLAG_FAST_CMPL | \
+#define UCP_PROTO_SELECT_OP_ATTR_MASK   (UCP_OP_ATTR_FLAG_NO_IMM_CMPL    | \
+                                         UCP_OP_ATTR_FLAG_FAST_CMPL      | \
+                                         UCP_OP_ATTR_FLAG_FORCE_IMM_CMPL | \
```
yosefe, line 24: adding these 2 flags to the mask will mean that when they are used it will create a different select_param key, and it's not the desired behavior
we need that ucp_proto_select_op_attr_pack will use the "full" mask only from UCP_OP_ATTR_INDEX context

src/ucp/proto/proto_select.h
```diff
@@ -43,6 +45,11 @@
 #define UCP_PROTO_SELECT_PARAM_STR_MAX 128
 
 
+/* Pack operation attributes from uint32_t to a uint8_t */
+#define ucp_proto_select_op_attr_pack(_op_attr) \
+    (((_op_attr) & UCP_PROTO_SELECT_OP_ATTR_MASK) / \
+    UCP_PROTO_SELECT_OP_ATTR_BASE)
```
yosefe, line 51: pls check code style in clang format, seems like hav alignment issue

michal-shalev:
@ivankochin as discussed, I'll open another ticket for the improvement

src/ucp/core/ucp_context.c
michal-shalev: `UCX_MEMTYPE_REG_WHOLE_ALLOC_TYPES` uses `UCS_CONFIG_TYPE_BITMAP` as well, so it is consistent with the format used for other configuration options in the table.
`ucx_info -f`:
image

for comparison:
image

src/ucp/core/ucp_context.c
michal-shalev: done

src/ucp/proto/proto_select.h
michal-shalev: I'll check why, thank you, done

src/ucp/proto/proto_select.h
michal-shalev: Instead of changing the flags in `UCP_PROTO_SELECT_OP_ATTR_MASK` (which would affect protocol selection - something we don't want), I created a new mask, `UCP_OP_ATTR_INDEX_MASK`, that includes all flags, including `UCP_OP_ATTR_FLAG_NO_IMM_CMPL` and `UCP_OP_ATTR_FLAG_FORCE_IMM_CMPL`. This mask is only used in `UCP_OP_ATTR_INDEX` to modify `op_attr` in `ucp_proto_select_elem_init()`, which affects the protocol’s internal behavior after the protocol is already selected based on the original key.

yosefe (approved):
pls squash

src/ucp/core/ucp_context.c
```diff
@@ -522,6 +529,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Prefetch memory when using global VA MR",
    ucs_offsetof(ucp_context_config_t, gva_prefetch), UCS_CONFIG_TYPE_BOOL},
 
+  {"FORCE_REQUEST_FLAGS", "",
```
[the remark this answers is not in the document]
michal-shalev: Thank you, I picked EXTRA_OP_ATTR_FLAGS thank you!

src/ucp/core/ucp_context.c
```diff
@@ -522,6 +529,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
...
+   "Defines forced request flags. Possible flags are: no_imm_cmpl, fast_cmpl, "
```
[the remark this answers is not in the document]
michal-shalev: done

src/ucp/core/ucp_context.h
```diff
@@ -42,6 +42,15 @@ enum {
...
 };
 
+#define UCP_OP_ATTR_INDEX_MASK (UCP_OP_ATTR_FLAG_NO_IMM_CMPL    | \
+                                UCP_OP_ATTR_FLAG_FORCE_IMM_CMPL | \
+                                UCP_OP_ATTR_FLAG_FAST_CMPL      | \
+                                UCP_OP_ATTR_FLAG_MULTI_SEND)
+
+#define UCP_OP_ATTR_INDEX(_op_attr_flag) \
+    (ucs_ilog2(ucp_proto_select_op_attr_pack((_op_attr_flag), \
+                                              UCP_OP_ATTR_INDEX_MASK)))
```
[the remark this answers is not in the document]
michal-shalev: done

src/ucp/core/ucp_context.h
```diff
@@ -189,6 +198,8 @@ typedef struct ucp_context_config {
     double                                 proto_overhead_rkey_ptr;
     /** Registration cache lookup overhead estimation */
     double                                 rcache_overhead;
+    /** UCP forced request flags */
+    uint64_t                                force_op_attr_flags;
```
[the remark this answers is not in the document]
michal-shalev: done

yosefe (approved):

brminich (commented):
this PR is missing support of forcing `UCP_OP_ATTR_FLAG_NO_IMM_CMPL` and `UCP_OP_ATTR_FLAG_FORCE_IMM_CMPL`.  Is it intentionally?

yosefe:
merge after 1.18.x branch

PR #10173, opened 2024-09-24
GTEST/COMMON: Added test for cuda async allocated buffers.

rakhmets:
## What
Extended UCP AM RNDV tests by adding buffers allocated by `cudaMallocAsync`.

### DIFF
--- a/test/gtest/common/mem_buffer.cc
+++ b/test/gtest/common/mem_buffer.cc
@@ -221,7 +221,7 @@ void mem_buffer::get_bar1_free_size_nvml()
 #endif
 }
 
-void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type)
+void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type, bool async)
 {
     void *ptr;
 
@@ -238,7 +238,18 @@ void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type)
         return ptr;
 #if HAVE_CUDA
     case UCS_MEMORY_TYPE_CUDA:
-        CUDA_CALL(cudaMalloc(&ptr, size), ": size=" << size);
+        if (async) {
+#if CUDART_VERSION >= 11020
+            CUDA_CALL(cudaMallocAsync(&ptr, size, 0), ": size=" << size);
+            cudaStreamSynchronize(0);
+#else
+            UCS_TEST_ABORT("asynchronous allocation for " +
+                           std::string(ucs_memory_type_names[mem_type]) +
+                           " memory type is not supported");
+#endif
+        } else {
+            CUDA_CALL(cudaMalloc(&ptr, size), ": size=" << size);
+        }
         return ptr;
     case UCS_MEMORY_TYPE_CUDA_MANAGED:
         CUDA_CALL(cudaMallocManaged(&ptr, size), ": size=" << size);
@@ -258,7 +269,7 @@ void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type)
     }
 }
 
-void mem_buffer::release(void *ptr, ucs_memory_type_t mem_type)
+void mem_buffer::release(void *ptr, ucs_memory_type_t mem_type, bool async)
 {
     try {
         switch (mem_type) {
@@ -267,6 +278,19 @@ void mem_buffer::release(void *ptr, ucs_memory_type_t mem_type)
             break;
 #if HAVE_CUDA
         case UCS_MEMORY_TYPE_CUDA:
+            if (async) {
+#if CUDART_VERSION >= 11020
+                cudaStreamSynchronize(0);
+                CUDA_CALL(cudaFreeAsync(ptr, 0), ": ptr=" << ptr);
+#else
+                UCS_TEST_ABORT("asynchronous release for " +
+                               std::string(ucs_memory_type_names[mem_type]) +
+                               " memory type is not supported");
+#endif
+            } else {
+                CUDA_CALL(cudaFree(ptr), ": ptr=" << ptr);
+            }
+            break;
         case UCS_MEMORY_TYPE_CUDA_MANAGED:
             CUDA_CALL(cudaFree(ptr), ": ptr=" << ptr);
             break;
@@ -513,6 +537,15 @@ bool mem_buffer::compare(const void *expected, const void *buffer,
 std::string mem_buffer::mem_type_name(ucs_memory_type_t mem_type)
 {
     return ucs_memory_type_names[mem_type];
+}
+
+bool mem_buffer::is_async_supported(ucs_memory_type_t mem_type)
+{
+#if CUDART_VERSION >= 11020
+    return mem_type == UCS_MEMORY_TYPE_CUDA;
+#else
+    return false;
+#endif
 }
 
 mem_buffer::mem_buffer(size_t size, ucs_memory_type_t mem_type) :
--- a/test/gtest/common/mem_buffer.h
+++ b/test/gtest/common/mem_buffer.h
@@ -28,10 +28,11 @@ public:
     static bool is_mem_type_supported(ucs_memory_type_t mem_type);
 
     /* allocate buffer of a given memory type */
-    static void *allocate(size_t size, ucs_memory_type_t mem_type);
-
+    static void *
+    allocate(size_t size, ucs_memory_type_t mem_type, bool async = false);
     /* release buffer of a given memory type */
-    static void release(void *ptr, ucs_memory_type_t mem_type);
+    static void
+    release(void *ptr, ucs_memory_type_t mem_type, bool async = false);
 
     /* fill pattern in a host-accessible buffer */
     static void pattern_fill(void *buffer, size_t length, uint64_t seed);
@@ -102,6 +103,11 @@ public:
     {
         return m_bar1_free_size;
     }
+
+    /**
+     * Check whether asynchronous operations are supported for the memory type
+     */
+    static bool is_async_supported(ucs_memory_type_t mem_type);
 
     mem_buffer(size_t size, ucs_memory_type_t mem_type);
     mem_buffer(size_t size, ucs_memory_type_t mem_type, uint64_t seed);
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -366,6 +366,16 @@ protected:
         return UCS_MEMORY_TYPE_HOST;
     }
 
+    virtual bool tx_memtype_async() const
+    {
+        return false;
+    }
+
+    virtual bool rx_memtype_async() const
+    {
+        return false;
+    }
+
     void reset_counters()
     {
         m_send_counter = 0;
@@ -495,8 +505,9 @@ protected:
                            unsigned flags = 0, unsigned data_cb_flags = 0,
                            uint32_t op_attr_mask = 0)
     {
-        mem_buffer sbuf(size, tx_memtype());
-        sbuf.pattern_fill(SEED);
+        auto sbuf = mem_buffer::allocate(size, tx_memtype(),
+                                         tx_memtype_async());
+        mem_buffer::pattern_fill(sbuf, size, SEED, tx_memtype());
         m_hdr.resize(header_size);
         ucs::fill_random(m_hdr);
         reset_counters();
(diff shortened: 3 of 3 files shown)


test/gtest/common/mem_buffer.cc
```diff
@@ -221,7 +221,7 @@ void mem_buffer::get_bar1_free_size_nvml()
 #endif
 }
 
-void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type)
+void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type, bool async)
```
iyastreb, line 224: Adding `async` argument may look misleading, because it can be passed for any memory type.
Maybe we can use extended enum for memory type?
```C
typedef enum {
    MEMORY_TYPE_CUDA_ASYNC = UCS_MEMORY_TYPE_LAST
} ucs_memory_type_ext_t;
```
And use this enum in allocate and release methods

test/gtest/common/mem_buffer.cc
```diff
@@ -239,6 +239,12 @@ void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type)
 #if HAVE_CUDA
     case UCS_MEMORY_TYPE_CUDA:
         CUDA_CALL(cudaMalloc(&ptr, size), ": size=" << size);
+        if (!async) {
```
iyastreb, line 242: very minor: maybe for readability it's more natural to first handle `if (async)` then else branch

test/gtest/common/mem_buffer.cc
```diff
@@ -268,7 +274,12 @@ void mem_buffer::release(void *ptr, ucs_memory_type_t mem_type)
 #if HAVE_CUDA
         case UCS_MEMORY_TYPE_CUDA:
         case UCS_MEMORY_TYPE_CUDA_MANAGED:
-            CUDA_CALL(cudaFree(ptr), ": ptr=" << ptr);
+            if (!async) {
```
iyastreb, line 277: async branch is not applicable to CUDA managed memory?

test/gtest/common/mem_buffer.h
```diff
@@ -103,8 +105,10 @@ class mem_buffer {
         return m_bar1_free_size;
     }
 
-    mem_buffer(size_t size, ucs_memory_type_t mem_type);
-    mem_buffer(size_t size, ucs_memory_type_t mem_type, uint64_t seed);
+    mem_buffer(size_t size, ucs_memory_type_t mem_type, bool async = false);
+    mem_buffer(size_t size, ucs_memory_type_t mem_type, bool async,
```
brminich, line 109: why not adding it to the end and also default to false?

rakhmets: It will not be compiled in this case. Because this will be unclear to the compiler which method to call in case of pasing three arguments.

test/gtest/common/mem_buffer.cc
rakhmets: Fixed.

test/gtest/common/mem_buffer.cc
rakhmets: Removed branch from CUDA_MANAGED case.

test/gtest/common/mem_buffer.h
rakhmets: Removed parameter from ctor.

test/gtest/common/mem_buffer.cc
rakhmets: > Adding `async` argument may look misleading, because it can be passed for any memory type. Maybe we can use extended enum for memory type?
> 
> ```c
> typedef enum {
>     MEMORY_TYPE_CUDA_ASYNC = UCS_MEMORY_TYPE_LAST
> } ucs_memory_type_ext_t;
> ```
> 
> And use this enum in allocate and release methods

I had the idea in mind. However, it would lead to significant code changes for the sake of a rather special use case.

brminich:
some CI errors are relevant:
```
/__w/1/s/ucx_src_89282/contrib/../test/gtest/common/mem_buffer.cc: In static member function 'static void* mem_buffer::allocate(size_t, ucs_memory_type_t, bool)':
/__w/1/s/ucx_src_89282/contrib/../test/gtest/common/mem_buffer.cc:242:23: error: 'cudaMallocAsync' was not declared in this scope
             CUDA_CALL(cudaMallocAsync(&ptr, size, 0), ": size=" << size);
                       ^
```

brminich (dismissed):

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1902,20 +1909,72 @@ class test_ucp_am_nbx_rndv_memtype : public test_ucp_am_nbx_rndv {
...
+        }
+
+        add_variant_memtypes(variants,
+                             test_ucp_am_nbx_prereg::get_test_variants);
+    }
+
+    static void add_variant_memtypes(variant_vec_t &variants,
+                                     get_variants_func_t generator)
+    {
+        ucp_test::add_variant_memtypes(variants, generator);
```
yosefe, line 1926: where do we add HOST memory?

test/gtest/common/mem_buffer.cc
```diff
@@ -238,7 +238,16 @@ void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type)
         return ptr;
 #if HAVE_CUDA
     case UCS_MEMORY_TYPE_CUDA:
-        CUDA_CALL(cudaMalloc(&ptr, size), ": size=" << size);
+#if CUDA_VERSION >= 11020
```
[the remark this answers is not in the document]
rakhmets: The compilation will fail without this directive. Because there is no `cudaMallocAsync` in older version.

test/gtest/common/mem_buffer.h
```diff
@@ -28,10 +28,12 @@ class mem_buffer {
     static bool is_mem_type_supported(ucs_memory_type_t mem_type);
 
     /* allocate buffer of a given memory type */
-    static void *allocate(size_t size, ucs_memory_type_t mem_type);
+    static void *allocate(size_t size, ucs_memory_type_t mem_type,
+                          bool async = false);
```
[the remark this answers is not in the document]
rakhmets: Sorry, not sure if I get the comment right. `is_async_supported(ucs_memory_type_t mem_type)` will return true only if memory type is CUDA. And is the method supposed to be used in `add_variant_memtypes`?

test/gtest/ucp/test_ucp_am.cc
rakhmets: `add_variant_memtypes` iterates through the vector returned by `supported_mem_types`. And host memory type is added as a first element to the vector in `supported_mem_types`.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1902,20 +1909,72 @@ class test_ucp_am_nbx_rndv_memtype : public test_ucp_am_nbx_rndv {
...
+    {
+        return get_variant_value(2) == MEMORY_TYPE_CUDA_ASYNC;
+    }
+
+    bool rx_memtype_async() const override
+    {
+        return get_variant_value(3) == MEMORY_TYPE_CUDA_ASYNC;
+    }
+
+    ucs_memory_type_t variant_value_to_mem_type(unsigned index) const
```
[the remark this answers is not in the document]
rakhmets: Fixed.

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1902,20 +1909,72 @@ class test_ucp_am_nbx_rndv_memtype : public test_ucp_am_nbx_rndv {
...
+        case UCS_MEMORY_TYPE_HOST:
+        case UCS_MEMORY_TYPE_CUDA:
+        case UCS_MEMORY_TYPE_CUDA_MANAGED:
+        case UCS_MEMORY_TYPE_ROCM:
+        case UCS_MEMORY_TYPE_ROCM_MANAGED:
+            return static_cast<ucs_memory_type_t>(variant_value);
+        case MEMORY_TYPE_CUDA_ASYNC:
+            return UCS_MEMORY_TYPE_CUDA;
+        default:
+            UCS_TEST_ABORT("invalid memory type");
```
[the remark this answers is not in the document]
rakhmets: Fixed.

test/gtest/common/mem_buffer.cc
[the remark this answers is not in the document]
yosefe: so maybe fail at runtime if user provided async==true and Cuda does not support it? Instead of just allocating memory without aysnc

test/gtest/common/mem_buffer.h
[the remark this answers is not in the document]
yosefe: Yes, this method is supposed to be used in add_variant_memtypes, to decide whether to create a test instance that uses a given memtype with async==true.
Currently we just assume async would be supported if CUDA is supported:
```c
        if (mem_buffer::is_mem_type_supported(UCS_MEMORY_TYPE_CUDA)) {
            add_variant_values(variants, generator, MEMORY_TYPE_CUDA_ASYNC,
                               "cuda-async");
        }
```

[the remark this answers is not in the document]
rakhmets: Added `is_async_supported`.

test/gtest/common/mem_buffer.cc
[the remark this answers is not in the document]
rakhmets: Added check using `is_async_supported`.

test/gtest/common/mem_buffer.cc
```diff
@@ -238,14 +244,14 @@ void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type, bool async)
         return ptr;
 #if HAVE_CUDA
     case UCS_MEMORY_TYPE_CUDA:
-#if CUDA_VERSION >= 11020
+#if CUDART_VERSION >= 11020
```
yosefe, line 247: maybe move this preprocessor check inside if (async) condition, to ensure no fallback will happen?
smth like
```c
if (async) {
#if CUDART_VERSION >= 11020
    CUDA_CALL(cudaMallocAsync(&ptr, size, 0), ": size=" << size);
    cudaStreamSynchronize(0);
#else
    UCS_TEST_ABORT(..);
#endif
}
```

test/gtest/common/mem_buffer.cc
```diff
@@ -269,6 +275,12 @@ void *mem_buffer::allocate(size_t size, ucs_memory_type_t mem_type, bool async)
 
 void mem_buffer::release(void *ptr, ucs_memory_type_t mem_type, bool async)
 {
+    if (async && !is_async_supported(mem_type)) {
+        UCS_TEST_MESSAGE << "asynchronous release for "
+                         << std::string(ucs_memory_type_names[mem_type])
+                         << " memory type is not supported";
+    }
```
yosefe, line 282: i guess this should not happen because otherwise we wouldnt be able to allocate? in this case we can just not release anything IMO..

test/gtest/common/mem_buffer.cc
```diff
@@ -535,6 +547,15 @@ std::string mem_buffer::mem_type_name(ucs_memory_type_t mem_type)
     return ucs_memory_type_names[mem_type];
 }
 
+bool mem_buffer::is_async_supported(ucs_memory_type_t mem_type)
+{
+#if CUDART_VERSION >= 11020
+    return (mem_type == UCS_MEMORY_TYPE_CUDA);
```
yosefe, line 553: extra ( )

test/gtest/common/mem_buffer.cc
rakhmets: But this is already covered by the if block at line 232-237.

test/gtest/common/mem_buffer.cc
rakhmets: Changed to `UCS_TEST_ABORT`.

test/gtest/common/mem_buffer.cc
rakhmets: Fixed.

test/gtest/common/mem_buffer.cc
yosefe: right, that block could be removed to simplify the code

rakhmets: Applied suggestion and removed extra code.

yosefe (approved):

brminich (approved):

yosefe (approved):

PR #10174, opened 2024-09-24
UCS/ASYNC: Use heap to process expired timers in batch

tvegas1:
## What
Use heap-based instead of stack-based allocation, while avoiding crashes.

## Why ?
The `ucs_alloca()` asserts for allocations above 1kB and we generally want to avoid using heap for such allocations anyways.

## How ?
Use `malloc()` with bigger max size.

### DIFF
--- a/src/ucs/async/async.c
+++ b/src/ucs/async/async.c
@@ -332,12 +332,19 @@ ucs_status_t ucs_async_dispatch_handlers(int *handler_ids, size_t count,
 ucs_status_t ucs_async_dispatch_timerq(ucs_timer_queue_t *timerq,
                                        ucs_time_t current_time)
 {
-    size_t max_timers, num_timers = 0;
+    size_t num_timers = 0;
+    ucs_status_t status;
+    size_t max_timers, size;
     int *expired_timers;
     ucs_timer_t *timer;
 
-    max_timers     = ucs_max(1, ucs_timerq_size(timerq));
-    expired_timers = ucs_alloca(max_timers * sizeof(*expired_timers));
+    max_timers = ucs_max(1, ucs_timerq_size(timerq));
+    size       = max_timers * sizeof(*expired_timers);
+
+    expired_timers = ucs_alloc_on_stack(size, "async_dispatch_timerq");
+    if (expired_timers == NULL) {
+        return UCS_ERR_NO_MEMORY;
+    }
 
     ucs_timerq_for_each_expired(timer, timerq, current_time, {
         expired_timers[num_timers++] = timer->id;
@@ -346,8 +353,10 @@ ucs_status_t ucs_async_dispatch_timerq(ucs_timer_queue_t *timerq,
         }
     })
 
-    return ucs_async_dispatch_handlers(expired_timers, num_timers,
-                                       UCS_ASYNC_EVENT_DUMMY);
+    status = ucs_async_dispatch_handlers(expired_timers, num_timers,
+                                         UCS_ASYNC_EVENT_DUMMY);
+    ucs_free_on_stack(expired_timers, size);
+    return status;
 }
 
 ucs_status_t ucs_async_context_init(ucs_async_context_t *async, ucs_async_mode_t mode)
--- a/src/ucs/sys/compiler.h
+++ b/src/ucs/sys/compiler.h
@@ -57,6 +57,26 @@
     }
 
 /**
+ * Use stack or heap allocation depending on size. Free function @ref
+ * ucs_free_on_stack must always be called.
+ */
+#define ucs_alloc_on_stack(_size, _name) \
+    ({ \
+        size_t _sz = (_size); \
+        (_sz <= UCS_ALLOCA_MAX_SIZE) ? alloca(_sz) : ucs_malloc(_sz, _name); \
+    })
+
+/**
+ * Release function to be paired with @ref ucs_alloc_on_stack.
+ */
+#define ucs_free_on_stack(_ptr, _size) \
+    ({ \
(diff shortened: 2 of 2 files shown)


tvegas1:
this PR is introducing batching processing.

src/ucs/async/async.c
```diff
@@ -332,12 +332,18 @@ ucs_status_t ucs_async_dispatch_handlers(int *handler_ids, size_t count,
...
+    size_t num_timers = 0;
+    ucs_status_t status;
+    size_t max_timers;
     int *expired_timers;
     ucs_timer_t *timer;
 
-    max_timers     = ucs_max(1, ucs_timerq_size(timerq));
-    expired_timers = ucs_alloca(max_timers * sizeof(*expired_timers));
+    max_timers     = ucs_min(1024, ucs_max(1, ucs_timerq_size(timerq)));
+    expired_timers = ucs_malloc(max_timers * sizeof(*expired_timers),
```
brminich, line 342: seems it can potentially hurt performance. Is there really the case when 1K is not enough?

tvegas1: currently `ucs_alloca()` allows max 1200 bytes, here we are capping at 1024*4 bytes = 4096 bytes so we should not hurt performance, as the limit has grown.

brminich: i meant malloc overhead

tvegas1: I see, yes seems so, also I though we are going away from stack allocation anyways. Do you think we should keep that buffer around, in some context?

yosefe: question is why we have so many timers, AFAIR the UD timer is per iface?
and if really need to support many timers, maybe allocate small size using alloca and large size using malloc

tvegas1: Yes, could not find users in UCX and UD transport is disabled, but I suspect we have external app using it, and with this PR, issue is completely gone.

tvegas1: fixed

brminich (approved):

src/ucs/async/async.c
```diff
@@ -332,12 +332,23 @@ ucs_status_t ucs_async_dispatch_handlers(int *handler_ids, size_t count,
...
+    size       = max_timers * sizeof(*expired_timers);
+
+    if (size > UCS_ALLOCA_MAX_SIZE) {
+        expired_timers = ucs_malloc(size, "async_dispatch_timerq");
+        if (expired_timers == NULL) {
+            return UCS_ERR_NO_MEMORY;
+        }
+    } else {
+        expired_timers = ucs_alloca(size);
+    }
```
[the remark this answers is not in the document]
tvegas1: fixed

src/ucs/sys/compiler.h
```diff
@@ -56,6 +56,26 @@
...
     }
 
+/**
+ * Use stack or heap allocation depending on size. Free function @ref
+ * ucs_free_on_stack must always be called.
+ */
+#define ucs_alloc_on_stack(_size, _name) \
+    ({ \
+        size_t _sz = (_size); \
+        ((_sz) <= UCS_ALLOCA_MAX_SIZE) ? alloca(_sz) : ucs_malloc(_sz, _name); \
```
yosefe, line 66: minor: remove () around _sz

src/ucs/async/async.c
```diff
@@ -332,12 +332,19 @@ ucs_status_t ucs_async_dispatch_handlers(int *handler_ids, size_t count,
...
-    size_t max_timers, num_timers = 0;
+    size_t num_timers = 0;
+    ucs_status_t status;
+    size_t max_timers, size;
     int *expired_timers;
     ucs_timer_t *timer;
 
-    max_timers     = ucs_max(1, ucs_timerq_size(timerq));
-    expired_timers = ucs_alloca(max_timers * sizeof(*expired_timers));
+    max_timers = ucs_min(1024, ucs_max(1, ucs_timerq_size(timerq)));
```
yosefe, line 341: why 1024 is a minimum?

tvegas1: This is to cap to allow batching, if we end-up having a really huge number of timers to expire. Previously we were using alloca() with 1200B limit, or 1200/4=300 timers, so we are quadrupling it, but true we could also removing it altogether, you want to remove it?

yosefe: yes, i'd prefer to keep previous behavior

tvegas1:
@yosefe, @brminich, followed up, issue will happen when creating many workers and/or with many ud interfaces in a single linux process, as they are all added on the same global object, so the fix makes sense.

Seen on nccl ucx plugin at scale and seen on EFA (>30 interfaces) running gtest `ud/test_ucp_rma_mt.put_get/0`.

src/ucs/async/async.c
```diff
@@ -332,12 +332,19 @@ ucs_status_t ucs_async_dispatch_handlers(int *handler_ids, size_t count,
...
     int *expired_timers;
     ucs_timer_t *timer;
 
-    max_timers     = ucs_max(1, ucs_timerq_size(timerq));
-    expired_timers = ucs_alloca(max_timers * sizeof(*expired_timers));
+    max_timers = ucs_max(1, ucs_timerq_size(timerq));
+    size       = max_timers * sizeof(*expired_timers);
+
+    expired_timers = ucs_alloc_on_stack(size, "async_dispatch_timerq");
+    if (expired_timers == NULL) {
```
brminich, line 345: minor
```suggestion
    if (ucs_unlikely(expired_timers == NULL)) {
```

brminich (approved):

yosefe (approved):

PR #10176, opened 2024-09-24
UCP: Use assertv in rkey size calc

brminich:
## What
Use `assertv` in `ucp_rkey_packed_size`

## Why ?
Better debugging experience

### DIFF
--- a/src/ucp/core/ucp_rkey.c
+++ b/src/ucp/core/ucp_rkey.c
@@ -65,7 +65,8 @@ size_t ucp_rkey_packed_size(ucp_context_h context, ucp_md_map_t md_map,
 
     ucs_for_each_bit(md_index, md_map) {
         tl_rkey_size = context->tl_mds[md_index].attr.rkey_packed_size;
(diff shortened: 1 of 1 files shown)


src/ucp/core/ucp_rkey.c
```diff
@@ -65,7 +65,8 @@ size_t ucp_rkey_packed_size(ucp_context_h context, ucp_md_map_t md_map,
 
     ucs_for_each_bit(md_index, md_map) {
         tl_rkey_size = context->tl_mds[md_index].attr.rkey_packed_size;
-        ucs_assert_always(tl_rkey_size <= UINT8_MAX);
+        ucs_assertv_always(tl_rkey_size <= UINT8_MAX, "md %s: tl_rkey_size=%zu",
```
tvegas1, line 68: there is the similar assert line 97 in this file, shall we also add same debugs?

tvegas1 (approved):

yosefe (approved):

PR #10177, opened 2024-09-24
UCS/DEBUG: use HAVE_ASAN instead of __SANITIZE_ADDRESS__

roiedanino:
## What
Use `#if HAVE_ASAN` when checking whether address sanitizer was enabled or not

## Why ?
`__SANITIZE_ADDRESS__` is not a project define and can affect users which try compiling their code with address sanitizer and UCX.


tvegas1 (dismissed):

test/gtest/uct/test_md.cc
```diff
@@ -1086,7 +1086,7 @@ UCS_TEST_SKIP_COND_P(test_md_fork, fork,
     ASSERT_EQ(pid, waitpid(pid, &child_status, 0));
     EXPECT_TRUE(WIFEXITED(child_status)) << ucs::exit_status_info(child_status);
 
-#ifndef __SANITIZE_ADDRESS__
+#if HAVE_ASAN
```
evgeny-leksikov, line 1089:
```suggestion
#if !HAVE_ASAN
```

tvegas1 (dismissed):

config/m4/compiler.m4
```diff
@@ -443,6 +443,11 @@ AS_IF([test "x$enable_asan" = xyes],
       [])
 
 
+
+AS_IF([test "x$enable_asan" = xyes],
+      [AC_DEFINE([HAVE_ASAN], 1, [Define if ASAN is enabled])])
```
tvegas1, line 448: maybe "Enable address sanitizer support"

tvegas1 (approved):

src/ucs/debug/debug.h
```diff
@@ -8,12 +8,7 @@
 #define UCS_DEBUG_H_
 
 #include <ucs/sys/compiler_def.h>
-
-#ifdef __SANITIZE_ADDRESS__
-#include <sanitizer/asan_interface.h>
-#include <ucs/debug/assert.h>
-#endif
-
+#include <stddef.h>
```
yosefe, line 11: why needed?

src/ucs/debug/debug.h
```diff
@@ -24,14 +19,11 @@ BEGIN_C_DECLS
  */
 void ucs_debug_disable_signal(int signum);
 
+void ucs_debug_asan_validate_address(const char *ptr_name, void *address,
+                                     size_t size);
 
-#ifdef __SANITIZE_ADDRESS__
```
yosefe, line 28: can we keep __SANITIZE_ADDRESS__ check here, to avoid extra function call overhead when not using the address sanitizer?

roiedanino: wouldn't that force us to add this check everywhere this function is being called from?

src/ucs/debug/debug.h
roiedanino: for size_t

src/ucs/debug/debug.h
yosefe: just need to restore the deteled ifdef. it means the check affects the macro definition, and code using the macro does not need to change.

src/ucs/debug/debug.c
```diff
@@ -1331,13 +1331,12 @@ void ucs_debug_disable_signals()
     ucs_recursive_spin_unlock(&ucs_kh_lock);
 }
 
-
+#ifdef __SANITIZE_ADDRESS__
```
yosefe, line 1334: after this change, if UCX was built without ASAN, and the user program is using ASAN, it would call an undefined symbol ucs_debug_asan_validate_address.
lets keep only the check itself under ifdef

src/ucs/debug/debug.c
```diff
@@ -27,6 +27,9 @@
 #ifdef HAVE_DETAILED_BACKTRACE
 #  include <bfd.h>
 #endif /* HAVE_DETAILED_BACKTRACE */
+#ifdef __SANITIZE_ADDRESS__
+#   include <sanitizer/asan_interface.h>
```
yosefe, line 31: use 2 spaces instead of 3

src/ucs/debug/debug.h
```diff
@@ -24,15 +19,17 @@ BEGIN_C_DECLS
...
 #ifdef __SANITIZE_ADDRESS__
 #define UCS_ASAN_ADDRESS_IS_VALID(_ptr, _size) \
-    ucs_assertv(!__asan_region_is_poisoned((void*)(_ptr), _size), "%s: %p", \
-                #_ptr, (void*)(_ptr))
+    ucs_debug_asan_validate_address(#_ptr, (void*)(_ptr), (_size))
 #else
 #define UCS_ASAN_ADDRESS_IS_VALID(_ptr, _size)
 #endif
 
+
```
yosefe, line 32: revert?

yosefe (approved):

ivankochin:
Seems like PR name and desription don't represent PR content anymore

PR #10178, opened 2024-09-24
UCP/MM: Disable global VA MR if error handling is in use

Artemy-Mellanox:


yosefe:

src/ucp/proto/proto_common.inl
```diff
@@ -362,6 +362,14 @@ ucp_proto_request_pack_rkey(ucp_request_t *req, ucp_md_map_t md_map,
                   md_map);
     }
 
+    if ((memh->flags & UCP_MEMH_FLAG_HAS_GVA) &&
+        (ucp_ep_config(req->send.ep)->key.err_mode ==
+         UCP_ERR_HANDLING_MODE_PEER)) {
```
evgeny-leksikov, line 367: Maybe extend and re-use `ucp_ep_config_err_mode_check_mismatch` adding message parameter?

src/ucp/proto/proto_common.inl
```diff
@@ -362,6 +362,14 @@ ucp_proto_request_pack_rkey(ucp_request_t *req, ucp_md_map_t md_map,
...
+        if (ucp_memh_disable_gva(memh, md_map) != UCS_OK) {
+            return 0;
```
evgeny-leksikov, line 369: how does proto-infrastructure handle this? I think need to abort request here.

src/ucp/proto/proto_common.inl
```diff
@@ -362,6 +362,14 @@ ucp_proto_request_pack_rkey(ucp_request_t *req, ucp_md_map_t md_map,
                   md_map);
     }
 
+    if ((memh->flags & UCP_MEMH_FLAG_HAS_GVA) &&
```
yosefe, line 365: 1. maybe do this before line 360: so if we disable gva, we still allow partial registration or registration failure, and don't fail the rkey packing completely.
2. Pls add comment before this code block to explain why we are disabling GVA here
3. Maybe extend GVA_ENABLE param to yes/no/auto - when auto means that we disable it with error handling. yes means always enable, even if may cause data corruption in error flow. default is still no, though.

src/ucp/proto/proto_common.inl
yosefe: it will send RTS/RTR without rkey, just with address

evgeny-leksikov: it seems weird to me, I would expect that we need handle error from `ucp_memh_disable_gva`

src/ucp/core/ucp_ep.inl
```diff
@@ -283,4 +283,10 @@ static UCS_F_ALWAYS_INLINE int ucp_ep_use_indirect_id(ucp_ep_h ep)
     return ep->flags & UCP_EP_FLAG_INDIRECT_ID;
 }
 
+static UCS_F_ALWAYS_INLINE int
+ucp_ep_config_err_mode_eq(ucp_ep_h ep, ucp_err_handling_mode_t err_mode)
+{
+    return ucp_ep_config(ep)->key.err_mode == err_mode;
+}
```
evgeny-leksikov, line 290: why is new function needed? we already have such one

Artemy-Mellanox: it's code common for `ucp_ep_config_err_mode_check_mismatch()` and new code in `ucp_proto_request_pack_rkey()`

src/ucp/proto/proto_common.inl
Artemy-Mellanox: AFAIU like in case of error in ucp_rkey_pack_memh() we fallback to protocol without rkey

src/ucp/core/ucp_ep.inl
evgeny-leksikov: maybe just extend old one like this?
```c
ucs_status_t
ucp_ep_config_err_mode_check_mismatch(ucp_ep_h ep,
                                      ucp_err_handling_mode_t err_mode,
                                      const char* err_message)
```

src/ucp/proto/proto_common.inl
yosefe: disable_gva should return void, if it fails memh->md_map will be partial

src/ucp/core/ucp_ep.inl
Artemy-Mellanox: we decided to use inline function in ucp_proto_request_pack_rkey(), its fast-path

evgeny-leksikov (approved):
dont forget to revert to "n" 👀

src/ucp/core/ucp_mm.c
```diff
@@ -507,14 +507,17 @@ static ucs_status_t ucp_memh_register_gva(ucp_context_h context, ucp_mem_h memh,
...
+    }
     return UCS_OK;
 }
 
 static ucs_status_t
 ucp_memh_register_internal(ucp_context_h context, ucp_mem_h memh,
                            ucp_md_map_t md_map, unsigned uct_flags,
                            const char *alloc_name, ucs_log_level_t err_level,
-                           int allow_partial_reg)
+                           int allow_partial_reg, int gva)
```
yosefe, line 520: gva -> gva_enable

src/ucp/core/ucp_mm.c
```diff
@@ -645,7 +650,19 @@ ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
...
+
+void ucp_memh_disable_gva(ucp_mem_h memh, ucp_md_map_t md_map)
+{
+    ucp_context_h context = memh->context;
+    ucs_status_t UCS_V_UNUSED status;
+
+    memh->md_map &= ~context->gva_md_map[memh->mem_type];
+    status = ucp_memh_register_internal(context, memh, md_map, 0, "disable gva",
+                                        UCS_LOG_LEVEL_DIAG, 1, 0);
+    ucs_assert(status == UCS_OK);
```
yosefe, line 664: why assert?

src/ucp/core/ucp_mm.c
```diff
@@ -645,7 +650,19 @@ ucs_status_t ucp_memh_register(ucp_context_h context, ucp_mem_h memh,
...
+    memh->flags &= ~UCP_MEMH_FLAG_HAS_GVA;
```
yosefe, line 665: maybe do this right after updating memh->md_map? for clarity

src/ucp/core/ucp_mm.c
```diff
@@ -507,14 +507,17 @@ static ucs_status_t ucp_memh_register_gva(ucp_context_h context, ucp_mem_h memh,
         memh->md_map |= UCS_BIT(md_index);
     }
 
+    if (context->config.ext.gva_enable == UCS_CONFIG_AUTO) {
```
yosefe, line 510: what if context->config.ext.gva_enable is UCS_YES? then UCP_MEMH_FLAG_HAS_GVA flag is not really represent the state of the memh, which is confusing during debug
maybe rename the flag to UCP_MEMH_FLAG_HAS_AUTO_GVA or similar?

src/ucp/proto/proto_common.inl
```diff
@@ -357,6 +357,16 @@ ucp_proto_request_pack_rkey(ucp_request_t *req, ucp_md_map_t md_map,
                 ucp_datatype_class_names[dt_iter->dt_class]);
 
     memh = dt_iter->type.contig.memh;
+
+    /* Since global VA registration doesn't support invalidation yet, and error
+     * handling is enabled on this EP, we replace GVA registrations with
+     * regular ones */
+    if ((memh->flags & UCP_MEMH_FLAG_HAS_GVA) &&
+        ucp_ep_config_err_mode_eq(req->send.ep,
+                                  UCP_ERR_HANDLING_MODE_PEER)) {
```
yosefe, line 366: for optimization:
```suggestion
    if (ucp_ep_config_err_mode_eq(req->send.ep,
                                  UCP_ERR_HANDLING_MODE_PEER) &&
        ucs_unlikely(memh->flags & UCP_MEMH_FLAG_HAS_GVA)) {
```

src/ucp/core/ucp_mm.c
```diff
@@ -507,14 +507,17 @@ static ucs_status_t ucp_memh_register_gva(ucp_context_h context, ucp_mem_h memh,
...
+        memh->flags |= UCP_MEMH_FLAG_HAS_GVA;
+    }
```
yosefe, line 512: space line after

src/ucp/proto/proto_common.inl
Artemy-Mellanox: AFAIK unlikely doesn't work for subexpressions, only for entire if condition.
do you expect checking ep configuration to be faster then memh flag? i thought vice-versa

src/ucp/core/ucp_mm.c
Artemy-Mellanox: to emphasis that function can't fail.
currently when passing `allow_partial_reg` status may only be OK, but if in future it changes we might need do something here

yosefe: `ucs_assert_always(status == UCS_OK); /* When allow_partial_reg==1 registration should not fail */`

src/ucp/proto/proto_common.inl
yosefe: > AFAIK unlikely doesn't work for subexpressions, only for entire if condition.

Do you have a doc ref for this? I could not find anything official about this..

> do you expect checking ep configuration to be faster then memh flag? i thought vice-versa

In the most performance-important case (no err handling) - we would have GVA, and the suggested order will have only one branch (since first check fails).

yosefe (approved):

PR #10182, opened 2024-09-25
TOOLS/PERF: Use mkey_pack_v2

ivankochin:
## What
Use `mkey_pack_v2` in `ucx_perftest`

## Why ?
Some transports (e.g. `cuda_ipc`) require passing address to `mkey_pack`. Fixes bug that fails these transports mkey packaging.

### DIFF
--- a/src/tools/perf/lib/libperf.c
+++ b/src/tools/perf/lib/libperf.c
@@ -19,6 +19,7 @@
 #include <ucs/sys/string.h>
 #include <ucs/type/serialize.h>
 #include <tools/perf/lib/libperf_int.h>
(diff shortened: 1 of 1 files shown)


iyastreb (approved):

brminich:
Pls add to description perftest bug which is fixed by this PR

brminich (approved):

ivankochin:
@yosefe pls review

yosefe (approved):
Regardless of making ucx_perftest work with v2, we should fix cuda_ipc so it will work also with the legacy uct_md_mkey_pack - for example by taking the first handle in the region, and failing only if there is more than one handle in the region.

PR #10183, opened 2024-09-25
UCP/CORE: Pack network address based on iface attribute flag

tvegas1:
## What
Check an added interface attribute flag bit when packing network addresses only.

## Why ?
The cuda_ipc mnnvl transport should not be part of network devices list, as it breaks configuration schemes where cuda_ipc is not listed in the device list.

## How ?
Sanity tested, but not end-to-end tested.


Akshay-Venkatesh (approved):
@tvegas1 I tested your PR and OSU benchmarks all worked as expected with different memory types.

brminich (dismissed):

src/ucp/core/ucp_worker.c
```diff
@@ -2951,6 +2951,7 @@ static ucs_status_t ucp_worker_address_pack(ucp_worker_h worker,
     unsigned flags        = ucp_worker_default_address_pack_flags(worker);
     ucp_tl_bitmap_t tl_bitmap;
     ucp_rsc_index_t tl_id;
+    uct_iface_attr_t *iface_attr;
```
yosefe, line 2954: const

yosefe:

src/ucp/core/ucp_worker.c
tvegas1: force-pushed that

yosefe (approved):

brminich:
There is a problem with this PR that we can wrongly detect ep configuration as `intra-node` even for inter-node cases.
Please update this check to avoid that
https://github.com/openucx/ucx/blob/master/src/ucp/wireup/select.c#L2329

brminich (commented):
need to also update ep config scope detection

shamisp:
> There is a problem with this PR that we can wrongly detect ep configuration as `intra-node` even for inter-node cases. Please update this check to avoid that https://github.com/openucx/ucx/blob/master/src/ucp/wireup/select.c#L2329

This is the right concern. MNNVL really becomes infer-node shared memory.

brminich (dismissed):

src/ucp/wireup/select.c
```diff
@@ -2326,8 +2326,10 @@ static ucs_status_t ucp_wireup_select_set_locality_flags(
     for (lane = 0; lane < key->num_lanes; ++lane) {
         rsc_index = key->lanes[lane].rsc_index;
         if ((rsc_index != UCP_NULL_RESOURCE) &&
-            (worker->context->tl_rscs[rsc_index].tl_rsc.dev_type ==
-             UCT_DEVICE_TYPE_SHM)) {
+            ((worker->context->tl_rscs[rsc_index].tl_rsc.dev_type ==
+             UCT_DEVICE_TYPE_SHM) &&
```
yosefe, line 2330: do we really need to check UCT_DEVICE_TYPE_SHM, maybe enough to test UCT_IFACE_FLAG_INTER_NODE?

tvegas1: removed the UCT_DEVICE_TYPE_SHM check here.

brminich (approved):

brminich:
@tvegas1 can you pls squash? The pr is small anyway

Issue #10180, opened 2024-09-24
ERROR mlx5_0: both WC and NC_DEDICATED UAR allocation types  are not supported


tonycurtis:
### Describe the bug

After building UCX @ github recently started getting errors as in subject line.  Adding `--without-devx`clears them up.

### Steps to Reproduce
- Command line

```
mpirun -n 2 ./a.out
```

- UCX version used (from github branch XX or release YY) + UCX configure flags (can be checked by `ucx_info -v`)

```
# Library version: 1.18.0
# Library path: /lustre/projects/ucx-group/opt/ucx/gcc/git/lib/libucs.so.0
# API headers version: 1.18.0
# Git branch 'master', revision 21c1594
# Configured with: --prefix=/lustre/projects/ucx-group/opt/ucx/gcc/git --with-knem=/opt/knem-1.1.4.90mlnx1 --with-xpmem=/opt/xpmem --enable-cma --without-java --without-go --enable-mt --enable-logging --enable-optimizations --with-valgrind --enable-builtin-memcpy --enable-frame-pointer
```

- **Any UCX environment variables used**

### Setup and versions
- OS version (e.g Linux distro) + CPU architecture (x86_64/aarch64/ppc64le/...)
   - `cat /etc/issue` or `cat /etc/redhat-release` + `uname -a`

```
fj-debug1[1](~/mpi) cat /etc/os-release
NAME="Rocky Linux"
VERSION="8.4 (Green Obsidian)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.4"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Rocky Linux 8.4 (Green Obsidian)"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:rocky:rocky:8.4:GA"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
ROCKY_SUPPORT_PRODUCT="Rocky Linux"
ROCKY_SUPPORT_PRODUCT_VERSION="8"
fj-debug1[1](~/mpi) uname -a
Linux login1 4.18.0-305.25.1.el8_4.aarch64 #1 SMP Tue Nov 2 21:37:36 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux
``
- For RDMA/IB/RoCE related issues:
    - Driver version:
        - `rpm -q rdma-core` or `rpm -q libibverbs`

[...]

PR #10187, opened 2024-09-26
UCT/IB/MLX5: reduce log level of both WC and NC unsupported

michal-shalev:
## What?
Reduce log level of both `WC` and `NC_DEDICATED` unsupported in `uct_ib_mlx5_devx_check_uar()`.

## Why?
Following https://github.com/openucx/ucx/issues/10180.

## How?
Reduce the log level of "both WC and NC_DEDICATED UAR allocation types are not supported" from ERROR to DIAG, and let UCX disable DevX automatically in this case.


gleon99 (approved):

gleon99:
@michal-shalev  please close the corresponding issue after this one is merged.

michal-shalev:
@gleon99 done

PR #10189, opened 2024-09-27
UCT/CUDA_IPC: Handle lack of IPC capability in unpack

Akshay-Venkatesh:
## What/Why?
[VMM ](https://developer.nvidia.com/blog/introducing-low-level-gpu-virtual-memory-management/) and [stream-ordered allocations](https://developer.nvidia.com/blog/using-cuda-stream-ordered-memory-allocator-part-2/) allocations can opt for fabric handle generation for IPC capabilities. Currently, UCX supports only allocations with fabric handles associated with it and fails in `uct_md_mem_reg` if allocations lack fabric handles. But with https://github.com/openucx/ucx/pull/10104 merged, UCX can pipeline VMM and stream-ordered allocations using staged pipelines that use library internal fabric pinned device memory buffers. This PR introduces a new key type (NON_IPC) that allows `uct_md_mem_reg` to pass when non-fabric user buffers are passed to it and forces cuda_ipc transport to fail during unpack phase. This in turn prompts UCX protov2 to attempt staged pipeline.

### DIFF
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -116,9 +116,13 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
     uct_cuda_ipc_lkey_t *key;
     ucs_status_t status;
 #if HAVE_CUDA_FABRIC
+#define UCT_CUDA_IPC_QUERY_NUM_ATTRS 2
     CUmemGenericAllocationHandle handle;
     CUmemoryPool mempool;
+    CUpointer_attribute attr_type[UCT_CUDA_IPC_QUERY_NUM_ATTRS];
+    void *attr_data[UCT_CUDA_IPC_QUERY_NUM_ATTRS];
     int legacy_capable;
+    int allowed_handle_types;
 #endif
 
     key = ucs_calloc(1, sizeof(*key), "uct_cuda_ipc_lkey_t");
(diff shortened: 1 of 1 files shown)


brminich (approved):

iyastreb (approved):

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -153,7 +162,7 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
                         CU_MEM_HANDLE_TYPE_FABRIC, 0));
         if (status != UCS_OK) {
             cuMemRelease(handle);
-            goto non_ipc;
+            goto err;
```
brminich, line 165: Wasn't that an original intent to fallback if `cuMemExportToShareableHandle` fails?

Akshay-Venkatesh: We shouldn't be hitting this case if the conditions are already met here https://github.com/openucx/ucx/pull/10189/files#diff-91e9ff28493264cecd03ffb88c50c0af9ff1be814254383dcec6ec40814d0f80R152. And if allocation doesn't have a fabric handle associated with it, then we will end up going to `non_ipc` path.

brminich: discussed offline

brminich (approved):

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -143,6 +144,14 @@ uct_cuda_ipc_mem_add_reg(void *addr, uct_cuda_ipc_memh_t *memh,
         goto legacy_path;
     }
 
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuPointerGetAttribute(
```
yosefe, line 147: 1. do we really have to check this, or cuMemExportToShareableHandle would fail anyway?
2. can we unite it with checking CU_POINTER_ATTRIBUTE_IS_LEGACY_CUDA_IPC_CAPABLE to check several attributes at once?
3. Is the goal of the PR is to move the failure from memory registration to rkey unpack? Which test would fail without this PR?

brminich: 1. It would fail, but it may also fail due to other reasons. 
3. Sending  VMM memory allocated without fabric capability will fail (registration failure will lead to MPI_Send abortion)

Akshay-Venkatesh: @yosefe 
1. Right now we're checking to avoid more checks in case memory isn't fabric capable. Export can still fail and we want to go through same handling
2. We can combine. I'll make the change. 
3. yes this is the goal of the PR.

yosefe: 1. ok

brminich (approved):

brminich:
@Akshay-Venkatesh, can you pls squash? (PR is small anyway)

Akshay-Venkatesh:
@brminich There is an unrelated error and test needs to be rerun. Can you please rerun?

brminich:
> @brminich There is an unrelated error and test needs to be rerun. Can you please rerun?

done

PR #10190, opened 2024-09-27
UCS/UCP/RCACHE: Store memh completion in request

iyastreb:
## What
The aim of this PR is to prepare infra for RNDV PPLN invalidation, which will be implemented without using rcache invalidation. Currently rcache stores completion entries within its private memory pool, and the proposed change is to move that completion entry to UCP request.

## Why ?
- This allows to use invalidation completion without rcache (will be used in the subsequent PR)
- There is no memory overhead in UCP request, because we reuse `invalidate` union struct
- Less memory usage in rcache
- Rcache does not need to bother with memory pool and acquire locks to deal with completion entries

## How ?
Request owns the completion entry within `invalidate` struct


src/ucp/rndv/rndv_rtr.c
```diff
@@ -222,11 +222,12 @@ static void ucp_proto_rndv_rtr_abort(ucp_request_t *req, ucs_status_t status)
     rreq->status = status;
     ucp_request_set_callback(req, send.cb, ucp_proto_rndv_rtr_abort_super);
 
-    if (ucp_request_memh_invalidate(req, status)) {
+    if (ucp_request_memh_check_invalidate(req)) {
         if (req->send.rndv.rkey != NULL) {
             ucp_proto_rndv_rkey_destroy(req);
         }
         ucp_proto_request_zcopy_id_reset(req);
+        ucp_request_memh_invalidate(req, status);
```
tvegas1, line 230: are we able to add `ucp_request_memh_invalidate()` right after the check line 226, or some `req` field might touched after?

src/ucp/core/ucp_request.c
```diff
@@ -395,50 +395,65 @@ static ucp_md_map_t ucp_request_get_invalidation_map(ucp_ep_h ep)
     return inv_map;
 }
 
-int ucp_request_memh_invalidate(ucp_request_t *req, ucs_status_t status)
+static UCS_F_ALWAYS_INLINE ucp_mem_h* ucp_request_get_memh(ucp_request_t *req)
```
evgeny-leksikov, line 398: it seems like no need in pointer
```suggestion
static UCS_F_ALWAYS_INLINE ucp_mem_h ucp_request_get_memh(ucp_request_t *req)
```

src/ucp/core/ucp_request.h
```diff
@@ -54,10 +54,11 @@ enum {
     UCP_REQUEST_FLAG_RKEY_INUSE            = UCS_BIT(18),
     UCP_REQUEST_FLAG_USER_HEADER_COPIED    = UCS_BIT(19),
     UCP_REQUEST_FLAG_USAGE_TRACKED         = UCS_BIT(20),
+    UCP_REQUEST_FLAG_INVALIDATED           = UCS_BIT(21),
```
evgeny-leksikov, line 57: - pls move under `UCS_ENABLE_ASSERT` section
- need better name since it's irrelevant for any request, e.g. `UCP_REQUEST_FLAG_MEMH_INVALIDATED`

iyastreb: In this PR it's used only within `ucs_assert`, but in the upcoming invalidation PR I use this flag in a couple of places to check whether request had been already invalidated. But I'll check once again if I manage to get rid those checks. So far it's not easy, because in some parts of the code after invalidation we still access `rndv` part.

This flag indicates invalidated request (not just memory handle), meaning the request which populated `invalidate` struct from the `send` union. As a consequence, we cannot safely access other union structures from invalidated request (e.g. we can't anymore safely access `rndv` part). So in fact this is the property of request itself

src/ucp/core/ucp_request.c
evgeny-leksikov: pointer is not needed as a return value of this function

src/ucp/core/ucp_request.h
evgeny-leksikov: - lets move back when and IF it's really needed
 - we dont have dedicated flags for union members, all prevoius flags named by state and/or action, IMO "invalidated request " may have to wide meaning without context

src/ucp/core/ucp_request.c
iyastreb: it's needed, please check again
We assign NULL value to the request field, which address is stored in `memh_p`

src/ucp/core/ucp_request.h
iyastreb: I think I will add code to this PR that actually uses this flag in production code path

src/ucp/core/ucp_request.c
evgeny-leksikov: ah.. found it but IMO this is a not good idea to set value using "getter", maybe reset the field directly by `req->send.state.dt_iter.type.contig.memh = NULL` at line 448?

iyastreb: As you can see this getter returns different address depending on request type. It can be:
- `&req->send.state.dt_iter.type.contig.memh`
- `&req->send.state.dt.dt.contig.memh`
- plus in the new version `&req->send.rndv.mdesc->memh`

And the idea is to get this address once and then treat it uniformly.
I could set the field directly as you proposed, but then I would need to repeat the check

evgeny-leksikov: maybe then add a helper function `ucp_request_extract_memh` like `ucp_wireup_extract_lane`?

iyastreb: Thanks for the hint. That would work if we always unconditionally replace original memh field by NULL.
But it's slightly more involved here, because I use the same function in `ucp_request_memh_check_invalidate`, that must not change the request, so it must be pure get.
But I'll think a bit more on how to make it less confusing

iyastreb: I refactored it in the latest PR with entire invalidation solution, used extract approach:
https://github.com/openucx/ucx/pull/10204/commits/03123f757fb58dccf12d48de4a7840c77c037867

evgeny-leksikov: 👍 does it mean that this PR is not relevant any more?
( just cross reference PRs https://github.com/openucx/ucx/pull/10204 )

iyastreb: I was hoping to merge this small low-risk PR in the current release, but no luck.
I will confirm with tech leads on Monday whether it makes sense to split the overall solution into several pieces

evgeny-leksikov: then maybe cherry-pick https://github.com/openucx/ucx/commit/03123f757fb58dccf12d48de4a7840c77c037867 to here?

src/ucp/core/ucp_request.h
iyastreb: done

iyastreb:
@brminich I addressed [your comment](https://github.com/openucx/ucx/pull/10204/files#r1791360335) in this PR

yosefe:
If we plan to remove completion callback from rcache, maybe just remove it when possible, and not expose internal rcache structure to be used by UCP.
BTW, we could have removed user-defined argument now, and use container_of

iyastreb:
> If we plan to remove completion callback from rcache, maybe just remove it when possible, and not expose internal rcache structure to be used by UCP. BTW, we could have removed user-defined argument now, and use container_of

@yosefe Maybe I misunderstood your comment, but
1. Regarding exposing rcache internals: I don't think we expose more than before this change
Currently rcache exposes `ucs_rcache_region_invalidate` function taking `ucs_rcache_invalidate_comp_func_t` as an argument. And the proposed change is to expose structure that caller will allocate:
```C
typedef struct {
    ucs_list_link_t                   list;
    ucs_rcache_invalidate_comp_func_t func;
    void                              *arg;
} ucs_rcache_comp_entry_t;
```
Apart from the `list` field it's basically the same stuff that we pass today, IMO it does not expose too much. But this allows us to get rid of a lock and extra memory overhead.

2. We can use `container_of` only if we assume that passed argument is always `ucp_request`, right?
Currently there is no such assumption, and I think it's better not to have it, to not introduce tight coupling between components

yosefe:
need to reconsider based on new invalidation design ("child" memh for packing rkey that could be invalidated without rela deregistration)

PR #10195, opened 2024-09-30
UCT/GDRCOPY: Detect runtime version match (API changed >=v2.4)

tvegas1:
## What
Ensure that the loaded `libgdrapi.so` is binary compatible with the version of UCX it was built against.

Issue: [4088373](https://redmine.mellanox.com/issues/4088373)

## Why ?
We see late error/crash, instead of gracefully disabling GDR transport:
```
symbol lookup error: /opt/nvidia/hpc_sdk/Linux_x86_64/24.9/comm_libs/12.6/hpcx/hpcx-
2.20/ucx/lib/ucx/libuct_cuda_gdrcopy.so.0: undefined symbol: gdr_get_info_v2
```

## How ?
Detect version currently in use.

### DIFF
--- a/src/uct/cuda/gdr_copy/gdr_copy_md.c
+++ b/src/uct/cuda/gdr_copy/gdr_copy_md.c
@@ -251,22 +251,40 @@ uct_gdr_copy_mem_dereg(uct_md_h uct_md,
     return status;
 }
 
+static int uct_gdr_is_gdr_get_info_v1(int major, int minor)
+{
+    return (major < 2) || ((major == 2) && (minor <= 3));
+}
+
 static ucs_status_t
 uct_gdr_copy_query_md_resources(uct_component_t *component,
                                 uct_md_resource_desc_t **resources_p,
                                 unsigned *num_resources_p)
 {
     gdr_t ctx;
+    int major, minor;
+
+    gdr_runtime_get_version(&major, &minor);
+    if (uct_gdr_is_gdr_get_info_v1(GDR_API_MAJOR_VERSION,
+                                   GDR_API_MINOR_VERSION) !=
+        uct_gdr_is_gdr_get_info_v1(major, minor)) {
+        ucs_warn("could not open gdr copy, built for v%d.%d but loaded v%d.%d",
+                 GDR_API_MAJOR_VERSION, GDR_API_MINOR_VERSION, major, minor);
+        goto err;
+    }
 
     ctx = gdr_open();
     if (ctx == NULL) {
         ucs_debug("could not open gdr copy. disabling gdr copy resource");
-        return uct_md_query_empty_md_resource(resources_p, num_resources_p);
+        goto err;
     }
     gdr_close(ctx);
 
     return uct_cuda_base_query_md_resources(component, resources_p,
                                             num_resources_p);
+
+err:
+    return uct_md_query_empty_md_resource(resources_p, num_resources_p);
 }
 
 static void uct_gdr_copy_md_close(uct_md_h uct_md)


tvegas1:
In theory it might be possible to support both version at a time as the gap is below. Not sure if it's a good idea (wrt alignment etc).

```diff
-struct gdr_info {
+typedef struct gdr_info_v2 {
     uint64_t va;
     uint64_t mapped_size;
     uint32_t page_size;
@@ -105,9 +112,12 @@ struct gdr_info {
     uint32_t cycles_per_ms;
     unsigned mapped:1;
     unsigned wc_mapping:1;
-};
-typedef struct gdr_info gdr_info_t;
-int gdr_get_info(gdr_t g, gdr_mh_t handle, gdr_info_t *info);
+    gdr_mapping_type_t mapping_type;
+} gdr_info_v2_t;
+typedef gdr_info_v2_t gdr_info_t;
+int gdr_get_info_v2(gdr_t g, gdr_mh_t handle, gdr_info_v2_t *info);
+
+#define gdr_get_info gdr_get_info_v2


```

brminich (approved):

src/uct/cuda/gdr_copy/gdr_copy_md.c
```diff
@@ -251,22 +251,40 @@ uct_gdr_copy_mem_dereg(uct_md_h uct_md,
...
+
 static ucs_status_t
 uct_gdr_copy_query_md_resources(uct_component_t *component,
                                 uct_md_resource_desc_t **resources_p,
                                 unsigned *num_resources_p)
 {
+    int major, minor;
     gdr_t ctx;
 
+    gdr_runtime_get_version(&major, &minor);
```
iyastreb, line 267: maybe question of taste, but IMO it will be more reusable from other functions if we encapsulate the whole block into a separate (inline?) function `uct_gdr_is_compatible` or something. Especially if in the future we want to support multiple versions

tvegas1: using separate function. not sure what reusability you mean, since we have one use-case. in the future, when we know what is needed we will anyway need to change all that.

brminich (approved):

src/uct/cuda/gdr_copy/gdr_copy_md.c
```diff
@@ -251,22 +251,50 @@ uct_gdr_copy_mem_dereg(uct_md_h uct_md,
...
+static int uct_gdr_is_gdr_compatible(void)
+{
+    int major, minor;
+
+    gdr_runtime_get_version(&major, &minor);
+    if (uct_gdr_is_gdr_get_info_v1(GDR_API_MAJOR_VERSION,
+                                   GDR_API_MINOR_VERSION) !=
+        uct_gdr_is_gdr_get_info_v1(major, minor)) {
+        ucs_warn("could not open gdr copy, built for v%d.%d but loaded v%d.%d",
+                 GDR_API_MAJOR_VERSION, GDR_API_MINOR_VERSION, major, minor);
```
yosefe, line 268: i'd move this warning to uct_gdr_copy_query_md_resources since that function prints the warning in other error cases

src/uct/cuda/gdr_copy/gdr_copy_md.c
```diff
@@ -251,22 +251,50 @@ uct_gdr_copy_mem_dereg(uct_md_h uct_md,
     return status;
 }
 
+static int uct_gdr_is_gdr_get_info_v1(int major, int minor)
+{
+    return (major < 2) || ((major == 2) && (minor <= 3));
+}
+
+static int uct_gdr_is_gdr_compatible(void)
```
yosefe, line 259: minor: remove "void"

tvegas1: removed function

yosefe (approved):

yosefe:
@tvegas1 pls squash

PR #10198, opened 2024-10-01
UCT/CUDA-IPC: Fix MNNVL reachability check

brminich:
## What
Fix MNNVL reachability check

### DIFF
--- a/config/m4/cuda.m4
+++ b/config/m4/cuda.m4
@@ -66,6 +66,13 @@ AS_IF([test "x$cuda_checked" != "xyes"],
                                     [AC_MSG_ERROR([libnvidia-ml not found. Install appropriate nvidia-driver package])])
                               cuda_happy="no"])])
 
+         # Check for nvmlDeviceGetGpuFabricInfoV
+         AC_CHECK_DECLS([nvmlDeviceGetGpuFabricInfoV],
+                        [AC_DEFINE([HAVE_NVML_FABRIC_INFO], 1, [Enable NVML GPU fabric info support])],
+                        [AC_MSG_NOTICE([nvmlDeviceGetGpuFabricInfoV function not found in libnvidia-ml. MNNVL support will be disabled.])],
+                        [[#include <nvml.h>]])
+
+
          # Check for cuda static library
          have_cuda_static="no"
          AS_IF([test "x$cuda_happy" = "xyes"],
--- a/src/uct/api/uct.h
+++ b/src/uct/api/uct.h
@@ -977,7 +977,13 @@ enum uct_ep_params_field {
     UCT_EP_PARAM_FIELD_PRIV_DATA_LENGTH           = UCS_BIT(15),
 
     /** Enables @ref uct_ep_params::local_sockaddr */
-    UCT_EP_PARAM_FIELD_LOCAL_SOCKADDR             = UCS_BIT(16)
+    UCT_EP_PARAM_FIELD_LOCAL_SOCKADDR             = UCS_BIT(16),
+
+    /** Enables @ref uct_ep_params::dev_addr_length */
+    UCT_EP_PARAM_FIELD_DEV_ADDR_LENGTH            = UCS_BIT(17),
+
+    /** Enables @ref uct_ep_params::iface_addr_length */
+    UCT_EP_PARAM_FIELD_IFACE_ADDR_LENGTH          = UCS_BIT(18)
 };
 
 
@@ -1427,7 +1433,19 @@ struct uct_ep_params {
      * @note The interface in this routine requires the
      * @ref UCT_IFACE_FLAG_CONNECT_TO_SOCKADDR capability.
      */
-    const ucs_sock_addr_t             *local_sockaddr;
+    const ucs_sock_addr_t               *local_sockaddr;
+
+    /**
+     * Device address length. If not provided, the transport will assume a
+     * default minimum length according to the address buffer contents.
+     */
+    size_t                              dev_addr_length;
+
+    /**
+     * Iface address length. If not provided, the transport will assume a
+     * default minimum length according to the address buffer contents.
+     */
+    size_t                              iface_addr_length;
 };
 
 
@@ -2339,7 +2357,7 @@ ucs_status_t uct_iface_reject(uct_iface_h iface,
  *    remote endpoint, @ref uct_ep_connect_to_ep will need to be called. Use of
  *    this mode requires @ref uct_ep_params_t::iface has the
  *    @ref UCT_IFACE_FLAG_CONNECT_TO_EP capability flag. It may be obtained by
- *    @ref uct_iface_query .
+ *    @ref uct_iface_query.
  * -# Connect to a remote interface: If @ref uct_ep_params_t::dev_addr and
  *    @ref uct_ep_params_t::iface_addr are set, this will establish an endpoint
  *    that is connected to a remote interface. This requires that
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -273,7 +273,9 @@ typedef enum {
     UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR         = UCS_BIT(1), /**< iface_addr field */
     UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING        = UCS_BIT(2), /**< info_string field */
     UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING_LENGTH = UCS_BIT(3), /**< info_string_length field */
-    UCT_IFACE_IS_REACHABLE_FIELD_SCOPE              = UCS_BIT(4) /**<  scope field */
+    UCT_IFACE_IS_REACHABLE_FIELD_SCOPE              = UCS_BIT(4), /**< scope field */
+    UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR_LENGTH = UCS_BIT(5), /**< device_addr_length field */
+    UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR_LENGTH  = UCS_BIT(6)  /**< iface_addr_length field */
 } uct_iface_is_reachable_field_mask_t;
 
 
@@ -611,6 +613,18 @@ typedef struct uct_iface_is_reachable_params {
      * Reachability scope.
      */
     uct_iface_reachability_scope_t scope;
+
+    /**
+     * Device address length. If not provided, the transport will assume a
+     * default minimum length according to the address buffer contents.
+     */
+    size_t                        device_addr_length;
+
+    /**
+     * Iface address length. If not provided, the transport will assume a
+     * default minimum length according to the address buffer contents.
+     */
+    size_t                        iface_addr_length;
 } uct_iface_is_reachable_params_t;
 
 
@@ -1058,8 +1072,6 @@ int uct_iface_is_reachable_v2(uct_iface_h iface,
  *
  * @param [in] ep           Endpoint to connect.
  * @param [in] device_addr  Remote device address.
- * @param [in] iface_addr   Remote interface address or NULL if such address is
- *                          not available.
  * @param [in] ep_addr      Remote endpoint address.
  * @param [in] params       Parameters as defined in @ref
  *                          uct_ep_connect_to_ep_params_t.
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
@@ -20,6 +20,18 @@
 #include <pthread.h>
 #include <nvml.h>
 
+
+typedef enum {
+    UCT_CUDA_IPC_DEVICE_ADDR_FLAG_MNNVL = UCS_BIT(0)
+} uct_cuda_ipc_device_addr_flags_t;
+
+
+typedef struct {
+    uint64_t     system_uuid;
+    uint8_t      flags; /* uct_cuda_ipc_device_addr_flags_t */
+} UCS_S_PACKED uct_cuda_ipc_device_addr_t;
+
+
 static ucs_config_field_t uct_cuda_ipc_iface_config_table[] = {
 
     {"", "", NULL,
@@ -63,7 +75,18 @@ static void UCS_CLASS_DELETE_FUNC_NAME(uct_cuda_ipc_iface_t)(uct_iface_t*);
 ucs_status_t uct_cuda_ipc_iface_get_device_address(uct_iface_t *tl_iface,
                                                    uct_device_addr_t *addr)
 {
-    *(uint64_t*)addr = ucs_get_system_id();
+    uct_cuda_ipc_device_addr_t *dev_addr = (uct_cuda_ipc_device_addr_t*)addr;
+    uct_cuda_ipc_iface_t *iface          = ucs_derived_of(tl_iface,
+                                                          uct_cuda_ipc_iface_t);
+    uct_cuda_ipc_md_t *md                = ucs_derived_of(iface->super.super.md,
+                                                           uct_cuda_ipc_md_t);
+
+    if (md->enable_mnnvl) {
+        dev_addr->flags = UCT_CUDA_IPC_DEVICE_ADDR_FLAG_MNNVL;
+    }
+
+    dev_addr->system_uuid = ucs_get_system_id();
+
     return UCS_OK;
 }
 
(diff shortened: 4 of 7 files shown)


src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -74,56 +101,69 @@ static ucs_status_t uct_cuda_ipc_iface_get_address(uct_iface_h tl_iface,
...
 {
-#if HAVE_CUDA_FABRIC
-    CUdevice cu_device;
-    int coherent;
-    ucs_status_t status;
-
-    status = UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&cu_device, 0));
-    if (status != UCS_OK) {
+#if HAVE_NVML_FABRIC_INFO
+    if (memcmp(&dev_addr->mnnvl_addr.cluster_uuid,
```
tvegas1, line 111: can remove the `&` for both

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -74,56 +101,69 @@ static ucs_status_t uct_cuda_ipc_iface_get_address(uct_iface_h tl_iface,
...
 
     if (!uct_iface_is_reachable_params_addrs_valid(params)) {
         return 0;
     }
 
-    if (getpid() == *(pid_t*)params->iface_addr) {
+    dev_addr_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params,
+                                   device_addr_length, DEVICE_ADDR_LENGTH,
+                                   sizeof(uint64_t));
+    dev_addr        = (const uct_cuda_ipc_device_addr_t *)UCS_PARAM_VALUE(
```
tvegas1, line 147: rest of function assumes that dev_addr is non-null, so we could read fields directly?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -74,56 +101,69 @@ static ucs_status_t uct_cuda_ipc_iface_get_address(uct_iface_h tl_iface,
...
+    if ((getpid() == *(pid_t*)params->iface_addr) && same_uuid) {
         uct_iface_fill_info_str_buf(params, "same process");
         return 0;
     }
 
-    /* Either multi-node NVLINK should be supported or iface has to be on the
-     * same node for cuda-ipc to be reachable */
-    if ((ucs_get_system_id() != *((const uint64_t*)params->device_addr)) &&
-        !uct_cuda_ipc_iface_is_mnnvl_supported(md)) {
+    if (md->enable_mnnvl && (dev_addr_len != sizeof(uint64_t))) {
```
tvegas1, line 157: maybe check `dev_addr_len >= sizeof(uct_cuda_ipc_device_addr_t)` since we access those fields

config/m4/cuda.m4
```diff
@@ -66,6 +66,13 @@ AS_IF([test "x$cuda_checked" != "xyes"],
                                     [AC_MSG_ERROR([libnvidia-ml not found. Install appropriate nvidia-driver package])])
                               cuda_happy="no"])])
 
+         # Check for nvmlDeviceGetGpuFabricInfoV
+         AC_CHECK_DECLS([nvmlDeviceGetGpuFabricInfoV],
+                        [AC_DEFINE([HAVE_NVML_FABRIC_INFO], 1, [Enable NVML GPU fabric info support])],
+                        [AC_MSG_NOTICE([nvmlDeviceGetGpuFabricInfoV function not found in libnvidia-ml. MNNVL support will be disabled.])],
```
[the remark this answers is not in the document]
brminich: without that we can't check MNNVL reachability, so I'd say yes we can't support MNNVL properly.

src/uct/api/v2/uct_v2.h
```diff
@@ -273,7 +273,8 @@ typedef enum {
     UCT_IFACE_IS_REACHABLE_FIELD_IFACE_ADDR         = UCS_BIT(1), /**< iface_addr field */
     UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING        = UCS_BIT(2), /**< info_string field */
     UCT_IFACE_IS_REACHABLE_FIELD_INFO_STRING_LENGTH = UCS_BIT(3), /**< info_string_length field */
-    UCT_IFACE_IS_REACHABLE_FIELD_SCOPE              = UCS_BIT(4) /**<  scope field */
+    UCT_IFACE_IS_REACHABLE_FIELD_SCOPE              = UCS_BIT(4), /**< scope field */
+    UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR_LENGTH = UCS_BIT(5),
```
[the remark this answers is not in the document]
brminich: we do not need for this PR. Maybe add later when there is a real need?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -63,7 +75,22 @@ static void UCS_CLASS_DELETE_FUNC_NAME(uct_cuda_ipc_iface_t)(uct_iface_t*);
...
 {
-    *(uint64_t*)addr = ucs_get_system_id();
+    uct_cuda_ipc_device_addr_t *dev_addr = (uct_cuda_ipc_device_addr_t*)addr;
+#if HAVE_NVML_FABRIC_INFO
+    uct_cuda_ipc_iface_t *iface          = ucs_derived_of(tl_iface,
+                                                          uct_cuda_ipc_iface_t);
+    uct_cuda_ipc_md_t *md                = ucs_derived_of(iface->super.super.md,
+                                                           uct_cuda_ipc_md_t);
+
+    if (md->enable_mnnvl) {
```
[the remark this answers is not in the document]
brminich: no, because we do not pack it in this case

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -245,15 +285,17 @@ static ucs_status_t uct_cuda_ipc_iface_query(uct_iface_h tl_iface,
     uct_base_iface_query(&iface->super.super, iface_attr);
 
     iface_attr->iface_addr_len          = sizeof(pid_t);
-    iface_attr->device_addr_len         = sizeof(uint64_t);
+    iface_attr->device_addr_len         = (md->enable_mnnvl) ?
```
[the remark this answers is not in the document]
brminich: > are we ok for wire-compatibility? i guess yes?

yes, wire-compat should be ok

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -407,6 +407,58 @@ uct_cuda_ipc_mem_dereg(uct_md_h md, const uct_md_mem_dereg_params_t *params)
...
+        goto err_sd;
+    }
+
+    mnnvl_supported = 1;
+
+err_sd:
+    UCT_NVML_FUNC_LOG_ERR(nvmlShutdown());
+err:
+#endif
+    if ((mnnvl_enable == UCS_YES) && !mnnvl_supported) {
```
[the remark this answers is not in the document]
brminich: no, on try we do not print errors/warnings. I can change to error for YES

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -74,56 +101,69 @@ static ucs_status_t uct_cuda_ipc_iface_get_address(uct_iface_h tl_iface,
...
+    int same_uuid;
 
     if (!uct_iface_is_reachable_params_addrs_valid(params)) {
         return 0;
     }
 
-    if (getpid() == *(pid_t*)params->iface_addr) {
+    dev_addr_len = UCS_PARAM_VALUE(UCT_IFACE_IS_REACHABLE_FIELD, params,
+                                   device_addr_length, DEVICE_ADDR_LENGTH,
+                                   sizeof(uint64_t));
```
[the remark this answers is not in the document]
brminich: where shall we use it?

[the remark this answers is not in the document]
yosefe: instead of sizeof(uint64_t)

src/uct/api/v2/uct_v2.h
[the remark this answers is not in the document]
yosefe: ok

config/m4/cuda.m4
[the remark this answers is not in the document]
yosefe: so maybe have single macro HAVE_MNNVL that depends on having all required functions including nvmlDeviceGetGpuFabricInfoV\;

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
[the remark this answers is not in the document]
brminich: ok, replaced with `sizeof(dev_addr->system_uuid)` (seems to be easier to readd than ucs_field_sizeof)

config/m4/cuda.m4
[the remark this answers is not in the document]
brminich: i think we still want to be able to allocate fabric memory if needed. We have `UCX_CUDA_COPY_ENABLE_FABRIC` for managing that

yosefe (dismissed):

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -407,6 +407,27 @@ uct_cuda_ipc_mem_dereg(uct_md_h md, const uct_md_mem_dereg_params_t *params)
     return UCS_OK;
 }
 
+#if HAVE_NVML_FABRIC_INFO
+static const char* uct_cuda_ipc_cluster_uuid_str(const uint8_t *cluster_uuid,
```
tvegas1, line 411: maybe use `ucs_str_dump_hex`, adapted if we need different `:` period

yosefe (approved):

config/m4/cuda.m4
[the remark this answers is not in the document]
tvegas1: as discussed offline, runtime options in place are sufficient and we want to avoid configure parameter inflation.

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -407,6 +408,62 @@ uct_cuda_ipc_mem_dereg(uct_md_h md, const uct_md_mem_dereg_params_t *params)
...
+    }
+
+    mnnvl_supported = 1;
+
+out_sd:
+    UCT_NVML_FUNC_LOG_ERR(nvmlShutdown());
+out:
+#endif
+    if ((mnnvl_enable == UCS_YES) && !mnnvl_supported) {
+        ucs_error("multi-node NVLINK support is requested but not supported");
```
tvegas1, line 461: minor: adapt message to show when it's build that does not support it?

tvegas1 (dismissed):

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -74,56 +101,67 @@ static ucs_status_t uct_cuda_ipc_iface_get_address(uct_iface_h tl_iface,
...
         return 0;
     }
 
-    status = UCT_CUDADRV_FUNC_LOG_ERR(
-            cuDeviceGetAttribute(&coherent,
-                                 CU_DEVICE_ATTRIBUTE_PAGEABLE_MEMORY_ACCESS_USES_HOST_PAGE_TABLES,
-                                 cu_device));
-    if (status != UCS_OK) {
+    if (dev_addr->mnnvl_addr.clique_id != md->fabric_info.cliqueId){
+        uct_iface_fill_info_str_buf(params, "clique id doesn't match");
```
Akshay-Venkatesh, line 118: @brminich The main concern with reachability check I have here is that device 0 determines reachability for the eventual device that this thread will use during transfer/progress time. Devices on the same node aren't required to be part of the same cluster or clique even though that is most likely true.

brminich: what is your suggestion?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -74,56 +101,67 @@ static ucs_status_t uct_cuda_ipc_iface_get_address(uct_iface_h tl_iface,
...
+        uct_iface_fill_info_str_buf(params, "cluster uuid doesn't match");
         return 0;
     }
 
-    status = UCT_CUDADRV_FUNC_LOG_ERR(
-            cuDeviceGetAttribute(&coherent,
-                                 CU_DEVICE_ATTRIBUTE_PAGEABLE_MEMORY_ACCESS_USES_HOST_PAGE_TABLES,
-                                 cu_device));
-    if (status != UCS_OK) {
+    if (dev_addr->mnnvl_addr.clique_id != md->fabric_info.cliqueId){
```
brminich, line 117:
```suggestion
    if (dev_addr->mnnvl_addr.clique_id != md->fabric_info.cliqueId) {
```

config/m4/cuda.m4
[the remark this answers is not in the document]
yosefe: AC_CHECK_DECLS would define a macro like HAVE_DECL__NVML_DEVICE_GET_GOU_FABRIC_INFO_V automatically, maybe just use it? instead of adding HAVE_NVML_FABRIC_INFO

src/uct/cuda/base/cuda_md.c
```diff
@@ -101,6 +101,37 @@ uct_cuda_base_query_md_resources(uct_component_t *component,
...
+        goto out;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(cuDeviceGet(&cu_device, 0));
+    if (status != UCS_OK) {
+        coherent = 0;
+        goto out;
+    }
+
+    status = UCT_CUDADRV_FUNC_LOG_ERR(
```
yosefe, line 121: why are we checking CU_DEVICE_ATTRIBUTE_PAGEABLE_MEMORY_ACCESS_USES_HOST_PAGE_TABLES for MNNVL? can you pls add doc?
Is it really needed given that we also check NVLINKS state?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -63,7 +75,18 @@ static void UCS_CLASS_DELETE_FUNC_NAME(uct_cuda_ipc_iface_t)(uct_iface_t*);
...
 {
-    *(uint64_t*)addr = ucs_get_system_id();
+    uct_cuda_ipc_device_addr_t *dev_addr = (uct_cuda_ipc_device_addr_t*)addr;
+    uct_cuda_ipc_iface_t *iface          = ucs_derived_of(tl_iface,
+                                                          uct_cuda_ipc_iface_t);
+    uct_cuda_ipc_md_t *md                = ucs_derived_of(iface->super.super.md,
+                                                           uct_cuda_ipc_md_t);
+
+    if (md->enable_mnnvl) {
+        dev_addr->flags = UCT_CUDA_IPC_DEVICE_ADDR_FLAG_MNNVL;
```
yosefe, line 85: maybe instead of extending device address with flags, we can set system_uuid to 0 (or some other constant) when have MNNVL? then no need to extend UCT API

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
yosefe: do we still need to have mnnvl_enable config param?

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
brminich: in this case we will not be able to compare system_uuids when on one node we have different GPUs with and without MNNVL support (probably unrealistic use-case)

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
yosefe: so value of YES should override the detection logic and not generate an error

src/uct/cuda/base/cuda_md.c
brminich: took it from the existing logic. 
@Akshay-Venkatesh, can you pls comment? Do we really need this check?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
brminich: what does override the detection logic mean? Error is printed if MNNVL support is requested but can't be initialized

src/uct/cuda/base/cuda_md.c
Akshay-Venkatesh: This is because MNNVL is supported only on coherent platforms and the above device attribute is a check for coherent platforms. If this is not true, then other checks can be completely skipped.

src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
brminich: another case is wire compatibility. If different processes on the same node would use different UCX libs, they will not use cuda-ipc

yosefe: I don't think supporting mixed versions is so important for MNNVL at this point

brminich: ok, but found another use-case:
we can't check whether it is the same process or not (in which case we do not have to use cuda-ipc)
If pids (iface addresses) are the same, we have to check if system ids are also the same, which is not possible with the proposed approach

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -407,6 +408,63 @@ uct_cuda_ipc_mem_dereg(uct_md_h md, const uct_md_mem_dereg_params_t *params)
...
+                                  ucs_ternary_auto_value_t mnnvl_enable)
+{
+    int mnnvl_supported = 0;
+#if HAVE_NVML_FABRIC_INFO
+    nvmlGpuFabricInfoV_t fabric_info;
+    nvmlDevice_t device;
+    ucs_status_t status;
+    char buf[64];
+
+    if (!uct_cuda_base_is_coherent() || (mnnvl_enable == UCS_NO)) {
```
yosefe, line 422: what if we didnt check uct_cuda_base_is_coherent here?

src/uct/api/uct.h
```diff
@@ -977,7 +977,13 @@ enum uct_ep_params_field {
     UCT_EP_PARAM_FIELD_PRIV_DATA_LENGTH           = UCS_BIT(15),
 
     /** Enables @ref uct_ep_params::local_sockaddr */
-    UCT_EP_PARAM_FIELD_LOCAL_SOCKADDR             = UCS_BIT(16)
+    UCT_EP_PARAM_FIELD_LOCAL_SOCKADDR             = UCS_BIT(16),
+
+    /** Enables @ref uct_ep_params::dev_addr_length */
+    UCT_EP_PARAM_FIELD_DEV_ADDR_LENGTH            = UCS_BIT(17),
```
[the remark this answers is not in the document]
brminich: I made it consistent with `UCT_EP_PARAM_FIELD_DEV_ADDR` , which is already present in this enum. 
And in  `is_reachable`  there is `UCT_IFACE_IS_REACHABLE_FIELD_DEVICE_ADDR`

So it is not possible to make `ep_create` and `is_rechable` consistent without renaming existing API

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -407,6 +408,63 @@ uct_cuda_ipc_mem_dereg(uct_md_h md, const uct_md_mem_dereg_params_t *params)
...
+                  sizeof(buf), SIZE_MAX));
+
+    if ((fabric_info.state != NVML_GPU_FABRIC_STATE_COMPLETED) ||
+        (fabric_info.status != NVML_SUCCESS)) {
+        goto out_sd;
+    }
+
+    mnnvl_supported = 1;
+
+out_sd:
```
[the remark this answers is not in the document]
brminich: does not look cleaner to me

src/uct/api/uct.h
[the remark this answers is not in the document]
yosefe: we can rename v2 api to make it consistent

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
Akshay-Venkatesh: Better to return early from this function and not go through nvml calls if platform is not coherent. MNNVL is only supported on coherent platforms.

brminich: this is what happens now

Akshay-Venkatesh: Spoke to Yossi offline. We can just do the nvml checks as these checks are performed once per MD and is off the critical path. It maybe worth profiling this function overall just to keep tabs on nvml call overheads as they are generally on the higher side.

brminich: fixed

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -412,14 +412,20 @@ static int
...
     nvmlDevice_t device;
     ucs_status_t status;
     char buf[64];
 
-    if (!uct_cuda_base_is_coherent() || (mnnvl_enable == UCS_NO)) {
+    if (mnnvl_supported != -1) {
+        goto out;
+    }
+
+    mnnvl_supported = 0;
```
yosefe, line 426: if called from multiple threads, a thread may think mnnvl is not supported while it actually is

brminich: it should be protected by the lock on UCP level. UCT is not thread-safe

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -423,27 +423,22 @@ uct_cuda_ipc_md_check_fabric_info(uct_cuda_ipc_md_t *md,
...
-
-    if (mnnvl_enable == UCS_NO) {
-        goto out;
-    }
-
-    status = UCT_NVML_FUNC(nvmlInit_v2(), UCS_LOG_LEVEL_DIAG);
-    if (status != UCS_OK) {
+    if ((mnnvl_enable == UCS_NO) ||
+        (UCT_NVML_FUNC(nvmlInit_v2(), UCS_LOG_LEVEL_DIAG) != UCS_OK)) {
+        mnnvl_supported = 0;
```
yosefe, line 428: can we use "goto out_not_supported;" here?

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -407,6 +408,67 @@ uct_cuda_ipc_mem_dereg(uct_md_h md, const uct_md_mem_dereg_params_t *params)
...
+        goto out_sd;
+    }
+
+out_not_supported:
+    mnnvl_supported = 0;
+out_sd:
+    UCT_NVML_FUNC_LOG_ERR(nvmlShutdown());
+out:
+#else
+    mnnvl_supported = 0;
```
yosefe, line 463: maybe initialize mnnvl_supported to 0 instead of -1 when HAVE_NVML_FABRIC_INFO is not defined?

yosefe (approved):

PR #10201, opened 2024-10-02
GTEST/UCP: Test workers on CPUs according to max open file count

tvegas1:
## What
Don't create worker if we cannot open enough file descriptors.

## Why ?
Prevent test failure.

## How ?
- Don't cache `RLIMIT_NOFILE` value so that we don't dump misleading error message on socket creation failure
  - Function only used in testing
- Try to grow soft limit but don't touch max limit value
- Estimate number of file descriptors and limit the test accordingly

### DIFF
--- a/src/ucs/sys/sys.c
+++ b/src/ucs/sys/sys.c
@@ -316,22 +316,29 @@ uint64_t ucs_generate_uuid(uint64_t seed)
            __sumup_host_name(5);
 }
 
-int ucs_sys_max_open_files()
+static ucs_status_t
+ucs_get_rlimit(int resource, rlim_t *rlimit_value, const char *name)
 {
-    static int file_limit = 0;
     struct rlimit rlim;
-    int ret;
 
-    if (file_limit == 0) {
-        ret = getrlimit(RLIMIT_NOFILE, &rlim);
-        if (ret == 0) {
-            file_limit = (int)rlim.rlim_cur;
-        } else {
-            file_limit = 1024;
-        }
+    if (getrlimit(resource, &rlim) != 0) {
+        ucs_debug("unable to get %s limit: %m", name);
+        return UCS_ERR_IO_ERROR;
     }
 
-    return file_limit;
+    *rlimit_value = rlim.rlim_cur;
+    return UCS_OK;
+}
+
+int ucs_sys_max_open_files()
+{
+    rlim_t value;
+
+    if (ucs_get_rlimit(RLIMIT_NOFILE, &value, "max opened file") != UCS_OK) {
+        return -1;
+    }
+
+    return (int)value;
 }
 
 ucs_status_t
@@ -1585,8 +1592,8 @@ err:
 
 ucs_status_t ucs_sys_get_effective_memlock_rlimit(size_t *rlimit_value)
 {
-    struct rlimit limit_info;
-    int res;
+    ucs_status_t status;
+    rlim_t value;
 
     /* Privileged users have no lock limit */
     if (ucs_sys_get_mlock_cap()) {
@@ -1596,15 +1603,12 @@ ucs_status_t ucs_sys_get_effective_memlock_rlimit(size_t *rlimit_value)
 
     /* Check the value of the max locked memory which is set on the system
      * (ulimit -l) */
-    res = getrlimit(RLIMIT_MEMLOCK, &limit_info);
-    if (res != 0) {
-        ucs_debug("unable to get the max locked memory limit: %m");
+    status = ucs_get_rlimit(RLIMIT_MEMLOCK, &value, "max locked memory");
+    if (status != UCS_OK) {
         return UCS_ERR_IO_ERROR;
     }
 
-    *rlimit_value = (limit_info.rlim_cur == RLIM_INFINITY) ?
-                            SIZE_MAX :
-                            limit_info.rlim_cur;
+    *rlimit_value = (value == RLIM_INFINITY) ? SIZE_MAX : value;
     return UCS_OK;
 }
 
--- a/test/gtest/ucp/test_ucp_worker.cc
+++ b/test/gtest/ucp/test_ucp_worker.cc
@@ -935,6 +935,11 @@ public:
         add_variant_with_value(variants, UCP_FEATURE_TAG, 0, "");
     }
 
+    test_worker_cpu_mask() :
+        m_rlimit(RLIMIT_NOFILE, m_fd_per_cpu * ucs_sys_get_num_cpus())
+    {
+    }
+
 protected:
     void init() override
     {
@@ -955,6 +960,9 @@ protected:
         m_cpu = cpu;
     }
 
+    ucs::rlimit_setter   m_rlimit;
+    static constexpr int m_fd_per_cpu = 60;
+
 private:
     unsigned m_cpu = 0;
 };
@@ -964,9 +972,11 @@ UCS_TEST_P(test_worker_cpu_mask, all_cpus)
     ucs_sys_cpuset_t mask;
 
     CPU_ZERO(&mask);
-
     ASSERT_GE(ucs_sys_getaffinity(&mask), 0);
-    for (auto cpu = 0; cpu < ucs_sys_get_num_cpus(); ++cpu) {
+
+    long num_cpus = ucs_min(m_rlimit.limit() / m_fd_per_cpu,
+                            ucs_sys_get_num_cpus());
+    for (auto cpu = 0; cpu < num_cpus; ++cpu) {
         if (CPU_ISSET(cpu, &mask)) {
             set_cpu(cpu);
             UCS_TEST_MESSAGE << "create entity on cpu " << cpu;
--- a/test/gtest/common/test_helpers.h
+++ b/test/gtest/common/test_helpers.h
@@ -33,6 +33,7 @@
 #include <dirent.h>
 #include <stdint.h>
 #include <ifaddrs.h>
+#include <sys/resource.h>
 
 
 #ifndef UINT16_MAX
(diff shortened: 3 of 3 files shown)


test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -936,9 +984,10 @@ class test_worker_cpu_mask : public ucp_test {
     }
 
 protected:
+    // Don't create any entities on startup
     void init() override
     {
-        // Don't create any entities on startup
+        m_rlimit.set(RLIMIT_NOFILE, m_fd_per_cpu * ucs_sys_get_num_cpus());
```
tvegas1, line 990: this cannot be pure raii in the test as we need `m_rlimit` to persist while `class test_worker_cpu_mask` is there.

iyastreb: Maybe you can still have it with RAII by using `std::unique_ptr`?

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
 
 UCP_INSTANTIATE_TEST_CASE_TLS(test_pci_bw, all, "all")
 
+class rlimit_setter {
+public:
+    rlimit_setter() : m_resource(0), m_limit(0), m_orig({})
+    {
```
iyastreb, line 936: I think we need to save original rlimit in the ctor (and only here).
Because subsequent set calls may override the stored value, and we never get back to original

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+    }
+
+    void set(int resource, rlim_t new_limit)
+    {
+        struct rlimit limit;
+
+        m_resource = resource;
+        if (getrlimit(m_resource, &m_orig) != 0) {
+            UCS_TEST_MESSAGE << "could not get rlimit";
```
iyastreb, line 945: maybe EXPECT_EQ(0)? So that caller test fails in this case

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+            m_limit = limit.rlim_cur;
+            UCS_TEST_MESSAGE << "set rlimit to " << m_limit;
+        } else {
+            UCS_TEST_MESSAGE << "could not set rlimit to " << limit.rlim_cur;
+        }
+    }
+
+    ~rlimit_setter()
+    {
+        if (m_limit != 0) {
```
iyastreb, line 963: with RAII with unique_ptr this check is redundant

test/gtest/ucp/test_ucp_worker.cc
tvegas1: there might also be multiple instantiation of rlimit_setter, released in different order

test/gtest/ucp/test_ucp_worker.cc
tvegas1: added unique_ptr, but we still need to check if we were able to getrlimit() in the first place

iyastreb (dismissed):

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+class rlimit_setter {
+public:
+    rlimit_setter(int resource, rlim_t new_limit) :
+        m_ret(-1), m_resource(resource), m_limit(1024), m_orig({})
+    {
+        struct rlimit limit;
+
+        m_ret = getrlimit(m_resource, &m_orig);
+        if (m_ret != 0) {
+            UCS_TEST_MESSAGE << "could not get rlimit";
```
yosefe, line 942: can we print the resource name or id that could not get?

src/ucs/sys/sys.c
```diff
@@ -318,20 +318,9 @@ uint64_t ucs_generate_uuid(uint64_t seed)
...
-        ret = getrlimit(RLIMIT_NOFILE, &rlim);
-        if (ret == 0) {
-            file_limit = (int)rlim.rlim_cur;
-        } else {
-            file_limit = 1024;
-        }
-    }
 
-    return file_limit;
+    return getrlimit(RLIMIT_NOFILE, &rlim) != 0 ? -1 : (int)rlim.rlim_cur;
```
yosefe, line 323: can we share commeon code with ucs_sys_get_effective_memlock_rlimit?

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+            return;
+        }
+
+        limit.rlim_cur = new_limit > m_orig.rlim_max ? m_orig.rlim_max :
```
yosefe, line 946: 1. define 'struct rlimit limit;' here and maybe init it inline
2. use std::min ?

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+                                                       new_limit;
+        limit.rlim_max = m_orig.rlim_max;
+
+        if (setrlimit(m_resource, &limit) == 0) {
+            m_limit = limit.rlim_cur;
+            UCS_TEST_MESSAGE << "set rlimit to " << m_limit;
```
yosefe, line 952: maybe remove this print?

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+        } else {
+            m_limit = m_orig.rlim_cur;
+            UCS_TEST_MESSAGE << "could not set rlimit to " << limit.rlim_cur;
```
yosefe, line 955: 1. print the resource name/id
2. maybe ucs_error?

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+        }
+    }
+
+    ~rlimit_setter()
+    {
+        if (m_ret == 0) {
+            UCS_TEST_MESSAGE << "restored rlimit";
```
yosefe, line 962: remove this print

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+    rlim_t limit() const
+    {
+        return m_limit;
+    }
+
+private:
+    int           m_ret;
+    int           m_resource;
+    rlim_t        m_limit;
+    struct rlimit m_orig;
```
yosefe, line 976: m_orig_limit

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+    }
+
+    rlim_t limit() const
+    {
+        return m_limit;
+    }
+
+private:
+    int           m_ret;
+    int           m_resource;
```
yosefe, line 974: can we make it const?

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
 
 UCP_INSTANTIATE_TEST_CASE_TLS(test_pci_bw, all, "all")
 
+class rlimit_setter {
+public:
+    rlimit_setter(int resource, rlim_t new_limit) :
+        m_ret(-1), m_resource(resource), m_limit(1024), m_orig({})
```
yosefe, line 936: does 1024 make sense for all resource types?

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -928,6 +930,52 @@ UCS_TEST_P(test_pci_bw, get_pci_bw)
...
+        }
+    }
+
+    rlim_t limit() const
+    {
+        return m_limit;
+    }
+
+private:
+    int           m_ret;
```
yosefe, line 973: maybe instead we can compare if m_limit==m_orig.rlim_cur and if equal do not need to restore during the dtor. so domn't need to keep m_ret variable.

test/gtest/ucp/test_ucp_worker.cc
```diff
@@ -936,9 +984,12 @@ class test_worker_cpu_mask : public ucp_test {
     }
 
 protected:
+    // Don't create any entities on startup
     void init() override
     {
-        // Don't create any entities on startup
+        m_rlimit = std::unique_ptr<rlimit_setter>(
```
yosefe, line 990: shall we reset it during cleanup()?
otherwise, it means it's reset during the destructor, which means it can be just a member variable.. and not need for unique_ptr

test/gtest/ucp/test_ucp_worker.cc
tvegas1: fixed

test/gtest/ucp/test_ucp_worker.cc
tvegas1: fixed

test/gtest/ucp/test_ucp_worker.cc
tvegas1: fixed

test/gtest/ucp/test_ucp_worker.cc
tvegas1: fixed

test/gtest/ucp/test_ucp_worker.cc
tvegas1: fixed

tvegas1: test might still pass even if setting fail

src/ucs/sys/sys.c
tvegas1: fixed

test/gtest/ucp/test_ucp_worker.cc
tvegas1: fixed

test/gtest/ucp/test_ucp_worker.cc
tvegas1: fixed

test/gtest/ucp/test_ucp_worker.cc
tvegas1: fixed

test/gtest/ucp/test_ucp_worker.cc
tvegas1: fixed

test/gtest/ucp/test_ucp_worker.cc
tvegas1: reverted to not using unique_ptr

yosefe (approved):

brminich (approved):

PR #10202, opened 2024-10-02
BINDINGS/GO: Add go mod support

Artemy-Mellanox:


bindings/go/src/examples/perftest/perftest.go
```diff
@@ -14,10 +14,11 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
-	. "ucx"
 	"unsafe"
-	. "cuda"
 	"runtime"
+
```
yosefe, line 19: remove space line?

bindings/go/src/examples/perftest/perftest.go
```diff
@@ -14,10 +14,11 @@ import (
...
+	. "github.com/openucx/ucx/bindings/go/src/ucx"
+	. "github.com/openucx/ucx/bindings/go/src/cuda"
```
yosefe, line 21: does it mean that when building perftest it will take ucx code from github and not local work dir?

buildlib/az-helpers.sh
```diff
@@ -291,3 +291,11 @@ git_clone_with_retry() {
     echo "Failed to clone UCX after $max_attempts attempts"
     return 1
 }
+
+setup_go_env() {
+    go env -w GO111MODULE=auto
+    cat <<EOF >> bindings/go/go.work
+use .
+replace github.com/openucx/ucx/bindings/go v0.0.0-00010101000000-000000000000 => $(pwd)/bindings/go
```
yosefe, line 299: what happens when we build perftest locally and not in CI?

Artemy-Mellanox: If building master may build as-is - it will take module from github, if need to build specific version or localy modified - need do the same, create go.work with redirection.

bindings/go/src/examples/perftest/perftest.go
Artemy-Mellanox: Yes, go import should use name matching the one in `go.mod` file, `github.com/openucx/ucx/bindings/go`.
To redirect to local work dir `go.work` and `replace` directive used.

buildlib/az-helpers.sh
yosefe: but i would expect that if i build locally everything will use the local code
why do we want to redirect to github?

Artemy-Mellanox: AFAIU in golang, when importing, you need to reference a repository, and for publicly accessible code, it should point to a public repository accordingly.

there's a somewhat similar issue with dependencies in our build system. for instance, when building the `ucx_perftest` binary, the actual binary is stored in `src/tools/perf/.libs/ucx_perftest` and expects to find the libraries under the install prefix. meanwhile, in `src/tools/perf/`, there's a local wrapper (`ucx_perftest`) that redirects to libraries that have been built but not yet installed. 

difference here is that "install prefix" in golang is github.com/ (or other public repo).

yosefe (approved):

yosefe (approved):

PR #10203, opened 2024-10-02
CONTRIB: Add checking script before push

tvegas1:
## What
Shell script to perform simple preliminary checks before pushing.

## Why ?
Avoid spending CI cycles with repeated pushing.

## How ?
Implement commit title, code format and spell checks locally. Ran `shellcheck` on it.
```
$ ./contrib/gitready.sh
$ ./contrib/gitready.sh --push origin/my_branch
```

### DIFF
--- a/buildlib/tools/git_ready.sh
+++ b/buildlib/tools/git_ready.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# See file LICENSE for terms.
+#
+
+source ./buildlib/tools/codestyle.sh
+
+ucx=://github.com/openucx/ucx
+upstream=$(git remote -v  | grep -P "[\t ].*${ucx}.*[\t ].*fetch" | cut -f 1 | head -n 1)
+base=$(git merge-base "$upstream"/master HEAD)
+
+if ! git diff-index --quiet HEAD; then
+    echo "error: tree not clean"
+    exit 1
+fi
+
+codestyle_check_commit_title "$base"..HEAD
+if [[ $? -ne 0 ]]
+then
+    echo "error: fix commit title"
+    exit 1
+fi
+
+# Code format
+module load dev/llvm-ucx || :
+git clang-format --diff "$base" HEAD | patch -p1
+module unload dev/llvm-ucx || :
+
+# Codespell
+codestyle_check_spell --write-changes || :
+
+# Create commit for pre-push fixes
+if ! git diff-index --quiet HEAD; then
+    git add -p
+    # returns 1 if cached index is not empty
+    git diff --cached --exit-code || git commit -m "$title"
+fi
+
+# Pushing
+if [ "${1-}" = "--push" ]
+then
+
+    opt="${2-}"
+    remote="${opt%%/*}"
+    branch="${opt#*/}"
+
+    if [ "$remote" = "$opt" ] || [ -z "$remote" ] || [ -z "$branch" ]
+    then
+        echo "error: specify push location with '--push <remote>/<branch_name>'"
+        exit 1
+    fi
+
+    cmd="git push $remote HEAD:refs/heads/$branch"
+    echo "$cmd"
+    echo "<enter> or <ctrl-c> to abort"
+    read -r
+    $cmd
+fi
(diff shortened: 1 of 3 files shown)


contrib/gitready.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash -eu
+
+git_commit() {
+    if ! git diff-index --quiet HEAD; then
+        git add -p
+        # returns 1 if cached index is not empty
+        git diff --cached --exit-code || git commit -m "$title"
+        git reset --hard
```
brminich, line 8: why is it needed? Don't we expect that there are no any uncommited changes at this stage?

tvegas1: while handling `git add -p` it might be that we did not stage all buffers, hence the cleaning of the remaining changes before doing the same for spellcheck.

yosefe (commented):
can we unify the manual checks with the ones run in CI, so they will not diverge over time?

tvegas1:
> can we unify the manual checks with the ones run in CI, so they will not diverge over time?

fixed, i guess before final merge and after approve/squash, it would be good that I trigger one commit title failure and one spellcheck  failure.

buildlib/tools/git_ready.sh
```diff
@@ -0,0 +1,65 @@
...
+#
+
+source ./buildlib/tools/codestyle.sh
+
+git_commit() {
+    if ! git diff-index --quiet HEAD; then
+        git add -p
+        # returns 1 if cached index is not empty
+        git diff --cached --exit-code || git commit -m "$title"
+        git reset --hard
```
brminich, line 14: why can't we just leave what remains after `git add -p`? Doing reset looks a bit dangerous to me

tvegas1: I think it is ok because we only run this script if tree is completely clean (line 22 early bail otherwise) and `git reset` does not touch untracked files. Also the we do sequence:
1- generate diff (format) + git add -p, commit
2- generate diff (spellcheck) + git add -p, commit
So between the two we need the tree to be clean of leftover changes.

buildlib/tools/git_ready.sh
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+codestyle_check_commit_title "$base"..HEAD
+if [[ $? -ne 0 ]]
+then
+    echo "error: fix commit title"
+    exit 1
+fi
+
+# Indent
```
yosefe, line 34: Code format

buildlib/tools/git_ready.sh
```diff
@@ -0,0 +1,65 @@
...
+module load dev/llvm-ucx || :
+git clang-format --diff "$base" HEAD | patch -p1
+module unload dev/llvm-ucx || :
+
+git_commit
+
+# Codespell
+codestyle_check_spell --write-changes || :
+
+git_commit
```
yosefe, line 44: why have git commit both here and in line 39?

buildlib/tools/git_ready.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Pushing
+if [ "${1-}" = "--push" ]
+then
+
+    opt="${2-}"
+    remote="${opt%%/*}"
+    branch="${opt#*/}"
+
+    if [ "$remote" = "$opt" ] || [ -z "$remote" ] || [ -z "$branch" ]
```
yosefe, line 54: if we have option to specify remote name, do we still need to define upstream url in "ucx" var?

buildlib/tools/git_ready.sh
tvegas1: fixed

buildlib/tools/git_ready.sh
tvegas1: yes because we need to push to our own repo, not the upstream we diff against.

buildlib/tools/git_ready.sh
tvegas1: merged both commits in one

buildlib/tools/git_ready.sh
yosefe: you mean the fork?

yosefe: maybe instead of defining ucx as upstream url, take it from the list of remotes?
what if the remote url is a bit different, for example using git:// or http:// ?

tvegas1: removed the leading `https`, probably did not get the suggestion to take it from the list of remotes.

yosefe (approved):

PR #10207, opened 2024-10-03
UCP/PROTO: Report MTYPE copy instead of TL transport overhead for rkey_ptr/mtype

ivankochin:
## What
Report right performance scheme for `rkey_ptr/mtype` proto.

## Why ?
`rkey_ptr/mtype` proto doesn't have transport overhead since it only acquires mapped memory rkey ptr from RTR and copies it to required buffer.


gleon99:

PR #5237, opened 2020-06-03
UCT/IB/UD: packet filter by any GID present on device

evgeny-leksikov:
## What
packet filter by any GID present on device

## Why ?
rdmacm can select different GID on addr/route resolve than UCT does it locally. In this case GRH will contain DGID according to remote EP configuration based on CM settings.

PR #10209, opened 2024-10-03
 UCT/IB/UD: Use interface address family when checking RoCE GRH

hjelmn:
Some RoCE device drivers fail to set the service level in the work completion for UD completions. This leads to silently dropping all (or almost all) UD packets on these networks. This commit updates the GRH check logic to use the GRH to detect IPv4  vs IPv6.

## What

Update the GID length calculation to use the GRH instead of the service level in the work completion.

## Why ?

We have a system using Intel iRDMA which is not setting the service level for any UD work completion. This makes UCT silently drop all UD packets. This PR changes the code for non-mlx hardware to detect the address family (IPv4 vs IPv6) based on the data in the GRH. For IPv6 the first byte will contain 0x6X where X is any value whereas this byte will be untouched in IPv4. To ensure that a mix of IPv6 and IPv4 packets does not confuse the logic all posted UD receive buffers will have the first byte set to a known value (0xff).

For mlx hardware the gid length will be calculated based on the RoCE packet type in flags_rqpn as was done before.

### DIFF
--- a/src/uct/ib/mlx5/ib_mlx5.inl
+++ b/src/uct/ib/mlx5/ib_mlx5.inl
@@ -1,5 +1,6 @@
 /**
  * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2016. ALL RIGHTS RESERVED.
+ * Copyright (c) Google, LLC, 2024. ALL RIGHTS RESERVED.
  *
  * See file LICENSE for terms.
  */
@@ -86,6 +87,35 @@ uct_ib_mlx5_cqe_is_grh_present(struct mlx5_cqe64* cqe)
 {
     return cqe->flags_rqpn & htonl(UCT_IB_MLX5_CQE_FLAG_L3_IN_DATA |
                                    UCT_IB_MLX5_CQE_FLAG_L3_IN_CQE);
+}
+
+static UCS_F_ALWAYS_INLINE size_t
+uct_ib_mlx5_cqe_roce_gid_len(struct mlx5_cqe64* cqe)
+{
+    /*
+     * Take the packet type from CQE, because:
+     * 1. According to Annex17_RoCEv2 (A17.4.5.1):
+     * For UD, the Completion Queue Entry (CQE) includes remote address
+     * information (InfiniBand Specification Vol. 1 Rev 1.2.1 Section 11.4.2.1).
+     * For RoCEv2, the remote address information comprises the source L2
+     * Address and a flag that indicates if the received frame is an IPv4,
+     * IPv6 or RoCE packet.
+     *
+     * 2. According to PRM, for responder UD/DC over RoCE sl represents RoCE
+     * packet type as:
+     * bit 3    : when set R-RoCE frame contains an UDP header otherwise not
+     * Bits[2:0]: L3_Header_Type, as defined below
+     *     - 0x0 : GRH - (RoCE v1.0)
+     *     - 0x1 : IPv6 - (RoCE v1.5/v2.0)
+     *     - 0x2 : IPv4 - (RoCE v1.5/v2.0)
+     *
+     * The service level is the most significant byte of cqe->flags_rqpn.
+     *
+     * Alternatively, this could be detected by examining the packet contents
+     * as is done for non-mlx5 transports.
+     */
+    return (cqe->flags_rqpn & htonl(UCT_IB_MLX5_RQPN_ROCE_FLAG_IPV4)) ?
+        UCS_IPV4_ADDR_LEN : UCS_IPV6_ADDR_LEN;
 }
 
 static UCS_F_ALWAYS_INLINE void*
--- a/src/uct/ib/mlx5/ud/ud_mlx5.c
+++ b/src/uct/ib/mlx5/ud/ud_mlx5.c
@@ -2,6 +2,7 @@
 * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2019. ALL RIGHTS RESERVED.
 * Copyright (C) ARM Ltd. 2017.  ALL RIGHTS RESERVED.
 * Copyright (C) Advanced Micro Devices, Inc. 2024. ALL RIGHTS RESERVED.
+* Copyright (c) Google, LLC, 2024. ALL RIGHTS RESERVED.
 *
 * See file LICENSE for terms.
 */
@@ -521,7 +522,7 @@ uct_ud_mlx5_iface_poll_rx(uct_ud_mlx5_iface_t *iface, int is_async)
 
     if (!uct_ud_iface_check_grh(&iface->super, packet,
                                 uct_ib_mlx5_cqe_is_grh_present(cqe),
-                                cqe->flags_rqpn & 0xFF)) {
+                                uct_ib_mlx5_cqe_roce_gid_len(cqe))) {
         ucs_mpool_put_inline(desc);
         goto out_polled;
     }
--- a/src/uct/ib/ud/base/ud_iface.h
+++ b/src/uct/ib/ud/base/ud_iface.h
@@ -1,5 +1,6 @@
 /**
 * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2020. ALL RIGHTS RESERVED.
+* Copyright (c) Google, LLC, 2024. ALL RIGHTS RESERVED.
 *
 * See file LICENSE for terms.
 */
@@ -395,10 +396,9 @@ static UCS_F_ALWAYS_INLINE void uct_ud_leave(uct_ud_iface_t *iface)
 
 static UCS_F_ALWAYS_INLINE int
 uct_ud_iface_check_grh(uct_ud_iface_t *iface, void *packet, int is_grh_present,
-                       uint8_t roce_pkt_type)
+                       size_t gid_len)
 {
     struct ibv_grh *grh = (struct ibv_grh *)packet;
-    size_t gid_len;
     union ibv_gid *gid;
     khiter_t khiter;
     char gid_str[128] UCS_V_UNUSED;
@@ -411,25 +411,6 @@ uct_ud_iface_check_grh(uct_ud_iface_t *iface, void *packet, int is_grh_present,
         ucs_warn("RoCE packet does not contain GRH");
         return 1;
     }
-
-    /*
-     * Take the packet type from CQE, because:
-     * 1. According to Annex17_RoCEv2 (A17.4.5.1):
-     * For UD, the Completion Queue Entry (CQE) includes remote address
-     * information (InfiniBand Specification Vol. 1 Rev 1.2.1 Section 11.4.2.1).
-     * For RoCEv2, the remote address information comprises the source L2
-     * Address and a flag that indicates if the received frame is an IPv4,
-     * IPv6 or RoCE packet.
-     * 2. According to PRM, for responder UD/DC over RoCE sl represents RoCE
-     * packet type as:
-     * bit 3    : when set R-RoCE frame contains an UDP header otherwise not
-     * Bits[2:0]: L3_Header_Type, as defined below
-     *     - 0x0 : GRH - (RoCE v1.0)
-     *     - 0x1 : IPv6 - (RoCE v1.5/v2.0)
-     *     - 0x2 : IPv4 - (RoCE v1.5/v2.0)
-     */
-    gid_len = ((roce_pkt_type & UCT_IB_CQE_SL_PKTYPE_MASK) == 0x2) ?
-              UCS_IPV4_ADDR_LEN : UCS_IPV6_ADDR_LEN;
 
     if (ucs_likely((gid_len == iface->gid_table.last_len) &&
                     uct_ud_gid_equal(&grh->dgid, &iface->gid_table.last,
--- a/src/uct/ib/ud/verbs/ud_verbs.c
+++ b/src/uct/ib/ud/verbs/ud_verbs.c
@@ -1,5 +1,6 @@
 /**
 * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2019. ALL RIGHTS RESERVED.
+* Copyright (c) Google, LLC, 2024. ALL RIGHTS RESERVED.
 *
 * See file LICENSE for terms.
 */
@@ -393,6 +394,20 @@ uct_ud_verbs_iface_poll_tx(uct_ud_verbs_iface_t *iface, int is_async)
     return 1;
 }
 
+static UCS_F_ALWAYS_INLINE size_t uct_ud_verbs_iface_get_gid_len(void *packet)
+{
+  /* The GRH will contain either an IPv4 or IPv6 header. If the former is
+   * present the header will start at offset 20 in the buffer otherwise it
+   * will start at offset 0. Since the two headers are of fixed size (20 or
+   * 40 bytes) this means we will either see 0x6? at offset 0 (IPv6) or 0x45
+   * at offset 20. The detection is a little tricky for IPv6 given that the
+   * first 20B are undefined for IPv4. To overcome this the first byte of
+   * the posted receive buffer is set to 0xff.
+   */
+  return ((((uint8_t*)packet)[0] & 0xf0) == 0x60) ? UCS_IPV6_ADDR_LEN :
+                                                    UCS_IPV4_ADDR_LEN;
+}
+
 static UCS_F_ALWAYS_INLINE unsigned
 uct_ud_verbs_iface_poll_rx(uct_ud_verbs_iface_t *iface, int is_async)
 {
@@ -413,7 +428,8 @@ uct_ud_verbs_iface_poll_rx(uct_ud_verbs_iface_t *iface, int is_async)
 
     UCT_IB_IFACE_VERBS_FOREACH_RXWQE(&iface->super.super, i, packet, wc, num_wcs) {
         if (!uct_ud_iface_check_grh(&iface->super, packet,
-                                    wc[i].wc_flags & IBV_WC_GRH, wc[i].sl)) {
+                                    wc[i].wc_flags & IBV_WC_GRH,
+                                    uct_ud_verbs_iface_get_gid_len(packet))) {
             ucs_mpool_put_inline((void*)wc[i].wr_id);
             continue;
         }
@@ -696,7 +712,7 @@ uct_ud_verbs_iface_post_recv_always(uct_ud_verbs_iface_t *iface, int max)
     struct ibv_recv_wr *bad_wr;
     uct_ib_recv_wr_t *wrs;
     unsigned count;
-    int ret;
+    int ret, i;
 
     wrs  = ucs_alloca(sizeof *wrs  * max);
 
@@ -704,6 +720,14 @@ uct_ud_verbs_iface_post_recv_always(uct_ud_verbs_iface_t *iface, int max)
                                         wrs, max);
     if (count == 0) {
         return;
+    }
+
+    /* Set the first byte in the receive buffer grh to a known value not equal to
+     * 0x6?. This should aid in the detection of IPv6 vs IPv4 because the first
+     * byte is undefined in the later and 0x6? in the former. It is unlikely
+     * this byte is touched with IPv4. */
+    for (i = 0; i < count; ++i) {
+        ((uint8_t*)wrs[i].sg.addr)[0] = 0xff;
     }
 
     ret = ibv_post_recv(iface->super.qp, &wrs[0].ibwr, &bad_wr);
--- a/src/uct/ib/mlx5/ib_mlx5.h
+++ b/src/uct/ib/mlx5/ib_mlx5.h
@@ -1,6 +1,7 @@
 /**
 * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2014. ALL RIGHTS RESERVED.
 * Copyright (C) ARM Ltd. 2016.  ALL RIGHTS RESERVED.
+* Copyright (c) Google, LLC, 2024. ALL RIGHTS RESERVED.
 *
 * See file LICENSE for terms.
 */
@@ -66,6 +67,9 @@
 #define UCT_IB_MLX5_ATOMIC_MODE_EXT      3
 #define UCT_IB_MLX5_CQE_FLAG_L3_IN_DATA  UCS_BIT(28) /* GRH/IP in the receive buffer */
 #define UCT_IB_MLX5_CQE_FLAG_L3_IN_CQE   UCS_BIT(29) /* GRH/IP in the CQE */
+/* Bits 24-26 of flags_rqpn indicate the packet type */
+#define UCT_IB_MLX5_RQPN_ROCE_FLAG_IPV6  UCS_BIT(24)
+#define UCT_IB_MLX5_RQPN_ROCE_FLAG_IPV4  UCS_BIT(25)
 #define UCT_IB_MLX5_CQE_FORMAT_MASK      0xc
 #define UCT_IB_MLX5_MINICQE_ARR_MAX_SIZE 7
 #define UCT_IB_MLX5_MP_RQ_BYTE_CNT_MASK  0x0000FFFF  /* Byte count mask for multi-packet RQs */


hjelmn:
I actually can't find anywhere that `wc.sl & 0x7` has any special value. I see the service level being set in Mellanox provides but not in the Intel provider. Basically, either the code in UCX was wrong or the Intel provider is wrong.

src/uct/ib/ud/base/ud_iface.h
```diff
@@ -427,8 +426,11 @@ uct_ud_iface_check_grh(uct_ud_iface_t *iface, void *packet, int is_grh_present,
      *     - 0x0 : GRH - (RoCE v1.0)
      *     - 0x1 : IPv6 - (RoCE v1.5/v2.0)
      *     - 0x2 : IPv4 - (RoCE v1.5/v2.0)
+     * Some RoCE drivers fail to set the sl correctly so instead of relying
+     * on the sl the check uses the current address family to detect the
+     * correct gid length.
      */
-    gid_len = ((roce_pkt_type & UCT_IB_CQE_SL_PKTYPE_MASK) == 0x2) ?
```
brminich, line 431: `UCT_IB_CQE_SL_PKTYPE_MASK` is not used anymore - can be deleted

src/uct/ib/ud/base/ud_iface.h
```diff
@@ -427,8 +426,11 @@ uct_ud_iface_check_grh(uct_ud_iface_t *iface, void *packet, int is_grh_present,
...
+    gid_len = (iface->super.gid_info.roce_info.addr_family == AF_INET) ?
```
brminich, line 433: @evgeny-leksikov, can we take `iface->gid_table.last_len` instead?

src/uct/ib/ud/base/ud_iface.h
hjelmn: Removes.

src/uct/ib/ud/base/ud_iface.h
hjelmn: Had the same question. Figured there was a reason that the code cared about what the address family was instead of using the cached size. I can change it to use that and see if it works on these systems. The address family can not change so the grh should always match and it it doesn't this will silently drop the packet. We were seeing this since most of the packets were being treated as IPv6 even though it is using IPv4. Only 1/7 were being handled correctly.

yosefe: This change actually breaks https://github.com/openucx/ucx/pull/5237 logic: it assumes the length of the received DGID is same as the length of the default GID index on the interface (say, GID[0]). Therefore the uct_ud_gid_equal() result might be incorrect when a non-default GID index is used (with different address family/length)

@hjelmn can we use a different heuristic to deduce the packet GID length, according to the GID contents for example? and do this only for cases where the RoCE driver fail to set the sl correctly?

yosefe (changes requested):

src/uct/ib/ud/base/ud_iface.h
hjelmn: From what I can tell setting the sl in this way is proprietary and only done by Mellanox. We determined this after some more research into the the service level. So, technically it is not incorrect to not set the sl (though I agree that Intel should be initializing it to something instead of not setting it at all). Let me know if that determination is incorrect.

We can probably use the grh itself to deduce the address type. For IPv4 the first 20 bytes are undefined but with IPv6 all 40 bytes are used. Let me look at the structure and see if there is something in the second 20 bytes we can hook into.

hjelmn:
@brminich Can you give insight into

https://github.com/openucx/ucx/commit/ed28845b88633e65d64fce8ec880060aa61bd59c

Which added the use of the sl (which references PRM (internal nVidia documentation apparently). What was broken when reading the IP version out of the GRH?

hjelmn:
Ok, I see. This was a really bad choice on the RoCE side. I mean, I can see why they put the IPv4 header 20B in (to make the copy of the header and packet a single copy for both IPv6 and IPv4. The problem is that the protocol version for IPv4 is in the middle of the source address of IPv6. So, we have to use great care here.

One option. Let's zero out the grh before posting a receive (if this is not already done). We could even just zero the first 20 bytes since the underfined is just not used in all likelihood. We can then use grh[0] == 0x6a  is IPv6 and grh[20] == 0x45 as IPv4 since the size of the two headers is fixed at 10 32-bit and 5 32-bit respectively. Let me go ahead and implement that now.

hjelmn:
Confirmed. At least with this hardware setting the first byte in all posted receives to 0xff shows that it is not overwritten for IPv4. Will update this PR shortly.

Example GRH:

```
0000 : ff 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 : �...............
0010 : 00 00 00 00 45 34 00 40 00 00 40 00 ff 11 00 00 : ....E4.@..@.�...
0020 : 0a 02 00 ef 0a 02 00 09                         : ...�....
```

hjelmn:
Ok, how about this? I left the mlx5 logic as is because, why not. For all other devices the first byte of the GRH is set before posting a receive. Once the packet is received we use byte 0 to detect IPv6 and byte 20 for IPv4. Tested with IRDMA.

src/uct/ib/ud/verbs/ud_verbs.c
```diff
@@ -393,6 +393,20 @@ uct_ud_verbs_iface_poll_tx(uct_ud_verbs_iface_t *iface, int is_async)
...
+{
+  /* The GRH will contain either an IPv4 or IPv6 header. If the former is
+   * present the header will start at offset 20 in the buffer otherwise it
+   * will start at offset 0. Since the two headers are of fixed size (20 or
+   * 40 bytes) this means we will either see 0x6a at offset 0 (IPv6) or 0x45
+   * at offset 20. The detection is a little tricky for IPv6 given that the
+   * first 20B are undefined for IPv4. To overcome this the first byte of
+   * the posted receive buffer is set to 0xff.
+   */
+    return (((unsigned char *)packet)[0] == 0x6a) ? UCS_IPV6_ADDR_LEN :
```
brminich, line 406: it would be safer to check just version in the first half-byte of the ipv6 header

src/uct/ib/ud/verbs/ud_verbs.c
```diff
@@ -393,6 +393,20 @@ uct_ud_verbs_iface_poll_tx(uct_ud_verbs_iface_t *iface, int is_async)
     return 1;
 }
 
+static UCS_F_ALWAYS_INLINE size_t uct_ud_iface_get_gid_len(void *packet)
```
yosefe, line 396: uct_ud_verbs_iface_get_gid_len

src/uct/ib/ud/verbs/ud_verbs.c
hjelmn: True, but two things. 1) IPv6 headers are a minimum of 40 bytes and 2) the GRH is always 40 bytes. I can make the change but the more explicit the check the less likely it will match a random byte if the provider does something unexpected.

src/uct/ib/ud/verbs/ud_verbs.c
```diff
@@ -706,6 +721,14 @@ uct_ud_verbs_iface_post_recv_always(uct_ud_verbs_iface_t *iface, int max)
         return;
     }
 
+    /* Set the first byte in the receive buffer grh to a known value not equal to
+     * 0x6a. This should aid in the detection of IPv6 vs IPv4 because the first
+     * byte is undefined in the later and 0x6a in the former. It is unlikely
+     * this byte is touched with IPv4. */
+    for (i = 0 ; i < count ; ++i) {
+      ((unsigned char *)wrs[i].sg.addr)[0] = 0xff;
```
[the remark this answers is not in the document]
hjelmn: Fixed in the next push.

src/uct/ib/ud/verbs/ud_verbs.c
hjelmn: Fixed in the next push. Coming once we have resolved one non-trivial comment.

src/uct/ib/ud/verbs/ud_verbs.c
brminich: but in ipv6 the first byte contains a part of traffic class, which can potentially vary (it's not a header len like in ipv4).

hjelmn: Ah, you are correct. I swear I was looking at the right document but it is indeed traffic class. Will fix.

src/uct/ib/mlx5/ud/ud_mlx5.c
```diff
@@ -519,9 +520,31 @@ uct_ud_mlx5_iface_poll_rx(uct_ud_mlx5_iface_t *iface, int is_async)
...
+     * bit 3    : when set R-RoCE frame contains an UDP header otherwise not
+     * Bits[2:0]: L3_Header_Type, as defined below
+     *     - 0x0 : GRH - (RoCE v1.0)
+     *     - 0x1 : IPv6 - (RoCE v1.5/v2.0)
+     *     - 0x2 : IPv4 - (RoCE v1.5/v2.0)
+     *
+     * Alternatively, this could be detected by examining the packet contents
+     * as is done for non-mlx5 transports (see
+     */
+    gid_len = (cqe->flags_rqpn & 0xff) == 0x2 ? UCS_IPV4_ADDR_LEN :
```
yosefe, line 543: add ( ) around == expression
`((cqe->flags_rqpn & 0xff) == 0x2) ? UCS_IPV4_ADDR_LEN :`

src/uct/ib/ud/base/ud_iface.h
```diff
@@ -1,5 +1,7 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
```
yosefe, line 1: pls remove

src/uct/ib/ud/base/ud_iface.h
```diff
@@ -412,24 +413,10 @@ uct_ud_iface_check_grh(uct_ud_iface_t *iface, void *packet, int is_grh_present,
...
-     * packet type as:
-     * bit 3    : when set R-RoCE frame contains an UDP header otherwise not
-     * Bits[2:0]: L3_Header_Type, as defined below
-     *     - 0x0 : GRH - (RoCE v1.0)
-     *     - 0x1 : IPv6 - (RoCE v1.5/v2.0)
-     *     - 0x2 : IPv4 - (RoCE v1.5/v2.0)
-     */
-    gid_len = ((roce_pkt_type & UCT_IB_CQE_SL_PKTYPE_MASK) == 0x2) ?
-              UCS_IPV4_ADDR_LEN : UCS_IPV6_ADDR_LEN;
+    if (ucs_unlikely(gid_len == 0)) {
```
yosefe, line 416: why needed?

src/uct/ib/ud/verbs/ud_verbs.c
```diff
@@ -1,5 +1,7 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
```
yosefe, line 1: pls remove

src/uct/ib/ud/verbs/ud_verbs.c
```diff
@@ -393,6 +395,20 @@ uct_ud_verbs_iface_poll_tx(uct_ud_verbs_iface_t *iface, int is_async)
...
+  /* The GRH will contain either an IPv4 or IPv6 header. If the former is
+   * present the header will start at offset 20 in the buffer otherwise it
+   * will start at offset 0. Since the two headers are of fixed size (20 or
+   * 40 bytes) this means we will either see 0x6? at offset 0 (IPv6) or 0x45
+   * at offset 20. The detection is a little tricky for IPv6 given that the
+   * first 20B are undefined for IPv4. To overcome this the first byte of
+   * the posted receive buffer is set to 0xff.
+   */
+  return ((((unsigned char *)packet)[0] & 0xf0) == 0x60) ? UCS_IPV6_ADDR_LEN :
+        (((unsigned char *)packet)[20] == 0x45) ? UCS_IPV4_ADDR_LEN : 0;
```
yosefe, line 409: do we really have to check also byte [20]?

src/uct/ib/ud/base/ud_iface.h
hjelmn: sure. useful for emacs to keep the file consistent with the intended style but I don't see any reason I will be editing these files again any time soon. removing in next push.

src/uct/ib/ud/base/ud_iface.h
hjelmn: A sanity check mostly intended for ud_verbs to catch an invalid format. If it is not worth checking I can remove both this and the check on byte 20 since they are related.

src/uct/ib/ud/verbs/ud_verbs.c
hjelmn: Removed in next push.

src/uct/ib/mlx5/ud/ud_mlx5.c
hjelmn: Fixed in next push.

src/uct/ib/ud/verbs/ud_verbs.c
hjelmn: Intended to catch an invalid grh. Figured it is worth warning if the packet neither starts with 0x6? or has 0x45 at offset 20.

yosefe: IMO better remove for optimization

hjelmn: Fair enough. Removing in next push.

src/uct/ib/ud/base/ud_iface.h
hjelmn: Removing.

hjelmn:
AddressSanitizer roce_proto_disable on worker 1 is failing. I may push a change to dump the contents of the GRH to see what is going on.

hjelmn:
It is detecting IPv4 packets as IPv6. This must be using the mlx5 UD transport. That still uses the CQE. Looking.

hjelmn:
Found the bug. Forgot to mask the bits down to 0x7. Since I am keeping the CQE calculation I am re-adding the #define for the mask.

src/uct/ib/mlx5/ud/ud_mlx5.c
```diff
@@ -519,9 +520,33 @@ uct_ud_mlx5_iface_poll_rx(uct_ud_mlx5_iface_t *iface, int is_async)
...
+     *     - 0x0 : GRH - (RoCE v1.0)
+     *     - 0x1 : IPv6 - (RoCE v1.5/v2.0)
+     *     - 0x2 : IPv4 - (RoCE v1.5/v2.0)
+     *
+     * The service level is the most significant byte of cqe->flags_rqpn.
+     *
+     * Alternatively, this could be detected by examining the packet contents
+     * as is done for non-mlx5 transports (see
+     */
+    gid_len = (((ntohl(cqe->flags_rqpn) >> 24) & UCT_IB_CQE_SL_PKTYPE_MASK) == 0x2) ?
```
brminich, line 545: seems this would be enough

```suggestion
    gid_len = ((cqe->flags_rqpn & UCT_IB_CQE_SL_PKTYPE_MASK) == 0x2) ?
```

hjelmn: Won't that break on big-endian systems? I was looking at the mlx5 provider which uses this to get the sl:

https://github.com/linux-rdma/rdma-core/blob/master/providers/mlx5/cq.c#L320

I see other instances of byte-swapping in ud_mlx5.c so figured I should do the same because the CQE is in big-endian format. Just want to verify before I go ahead and make the change.

hjelmn: Yes, and the byte we care about is the MSB of flags_rqpn so it is the LSB on little endian (why just masking the value without a byte-swap and bit-shift will work on LE) but on BE you will get LE MSB which is not what we want. Correct?

brminich: Right.
maybe now, when `UCT_IB_CQE_SL_PKTYPE_MASK`  is used only for mlx5, we can move it to `src/uct/ib/mlx5/ib_mlx5.h`  and define like
`UCT_IB_MLX5_CQE_SL_PKTYPE_MASK = 0x7000000` 
and do this:
` gid_len = ((cqe->flags_rqpn & htonl(UCT_IB_CQE_SL_PKTYPE_MASK)) == 0x2) ` (can be an inline function like `uct_ib_mlx5_cqe_is_grh_present`)
similarly to what `uct_ib_mlx5_cqe_is_grh_present` does using `UCT_IB_MLX5_CQE_FLAG_L3_IN_DATA` and `UCT_IB_MLX5_CQE_FLAG_L3_IN_CQE`

note flags_rqpn is:
image

brminich: Right.
maybe now, when UCT_IB_CQE_SL_PKTYPE_MASK is used only for mlx5, we can move it to src/uct/ib/mlx5/ib_mlx5.h and define like
UCT_IB_MLX5_CQE_SL_PKTYPE_MASK = 0x7000000
and do this:
gid_len = ((cqe->flags_rqpn & htonl(UCT_IB_MLX5_CQE_SL_PKTYPE_MASK )) == 0x2) (can be an inline function like uct_ib_mlx5_cqe_is_grh_present)
similarly to what uct_ib_mlx5_cqe_is_grh_present does using UCT_IB_MLX5_CQE_FLAG_L3_IN_DATA and UCT_IB_MLX5_CQE_FLAG_L3_IN_CQE

hjelmn: That won't quite work the same way because it is not a single bit. I changed it to avoid the << 24. What do you think of the new implementation. The alternative would be to treat it as a single bit and use 0x02000000 as the bit to detect IPv4. I can do that easily enough.

yosefe: The current implementation looks over complicated to me. 
IMO we should go with @brminich 's suggestion: Regardless on endianness, when we want to check if specific bits from a 32-bit value match expected value, a single & operation along with == should be enough. No need to add overheads of conversion and bit shifts.

brminich: we still want to convert a mask (constat) i guess

src/uct/ib/mlx5/ud/ud_mlx5.c
```diff
@@ -519,9 +519,31 @@ uct_ud_mlx5_iface_poll_rx(uct_ud_mlx5_iface_t *iface, int is_async)
...
+    /*
+     * Take the packet type from CQE, because:
+     * 1. According to Annex17_RoCEv2 (A17.4.5.1):
+     * For UD, the Completion Queue Entry (CQE) includes remote address
+     * information (InfiniBand Specification Vol. 1 Rev 1.2.1 Section 11.4.2.1).
+     * For RoCEv2, the remote address information comprises the source L2
+     * Address and a flag that indicates if the received frame is an IPv4,
+     * IPv6 or RoCE packet.
+     *
+     * According to PRM, for responder UD/DC over RoCE sl represents RoCE
```
brminich, line 531: pls move this comment to `uct_ib_mlx5_cqe_roce_gid_len`

src/uct/ib/mlx5/ud/ud_mlx5.c
hjelmn: If the valid values are 0-3 then this is the minimum code that is needed. If they are 0-2 (which appears to be the case) I can simplify it more. Will do that.

src/uct/ib/mlx5/ud/ud_mlx5.c
hjelmn: Done.

src/uct/ib/mlx5/ud/ud_mlx5.c
hjelmn: Ok. Now treating the roce packet types as flags with bit 24 being IPv6 and bit 25 being IPv4.

hjelmn:
Ok, should be good to go. This is clean but won't work if the packet type can be 0x3. If that is possible I can make one more tweak that doesn't require byte-swapping the flags.

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -88,6 +88,36 @@ uct_ib_mlx5_cqe_is_grh_present(struct mlx5_cqe64* cqe)
...
+    /*
+     * Take the packet type from CQE, because:
+     * 1. According to Annex17_RoCEv2 (A17.4.5.1):
+     * For UD, the Completion Queue Entry (CQE) includes remote address
+     * information (InfiniBand Specification Vol. 1 Rev 1.2.1 Section 11.4.2.1).
+     * For RoCEv2, the remote address information comprises the source L2
+     * Address and a flag that indicates if the received frame is an IPv4,
+     * IPv6 or RoCE packet.
+     *
+     * According to PRM, for responder UD/DC over RoCE sl represents RoCE
```
brminich, line 103:
```suggestion
     * 2. According to PRM, for responder UD/DC over RoCE sl represents RoCE
```

src/uct/ib/ud/base/ud_iface.h
```diff
@@ -1,5 +1,6 @@
 /**
 * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2001-2020. ALL RIGHTS RESERVED.
+* Copyright (c) Google, LLC, 2024. ALL RIGHTS RESERVED.
```
brminich, line 3: minor: you did not  include it to other files modified in this patch

brminich (commented):
:+1 besides minor comments

src/uct/ib/mlx5/ib_mlx5.inl
hjelmn: Fixed.

src/uct/ib/ud/verbs/ud_verbs.c
```diff
@@ -706,6 +722,14 @@ uct_ud_verbs_iface_post_recv_always(uct_ud_verbs_iface_t *iface, int max)
         return;
     }
 
+    /* Set the first byte in the receive buffer grh to a known value not equal to
+     * 0x6?. This should aid in the detection of IPv6 vs IPv4 because the first
+     * byte is undefined in the later and 0x6? in the former. It is unlikely
+     * this byte is touched with IPv4. */
+    for (i = 0 ; i < count ; ++i) {
```
brminich, line 729: our coding style:
```suggestion
    for (i = 0; i < count; ++i) {
```

brminich (dismissed):

src/uct/ib/ud/verbs/ud_verbs.c
hjelmn: Sure. I am running that now. Did not realize there was a clang format file. Will update this shortly.

hjelmn: Done.

hjelmn: Or not. One second.

src/uct/ib/mlx5/ib_mlx5.inl
```diff
@@ -88,6 +89,36 @@ uct_ib_mlx5_cqe_is_grh_present(struct mlx5_cqe64* cqe)
...
+     *
+     * The service level is the most significant byte of cqe->flags_rqpn.
+     *
+     * Alternatively, this could be detected by examining the packet contents
+     * as is done for non-mlx5 transports.
+     */
+    return (cqe->flags_rqpn & htonl(UCT_IB_MLX5_RQPN_ROCE_FLAG_IPV4)) ?
+        UCS_IPV4_ADDR_LEN : UCS_IPV6_ADDR_LEN;
+}
+
```
yosefe, line 120: minor: remove one space line

src/uct/ib/ud/verbs/ud_verbs.c
```diff
@@ -393,6 +394,20 @@ uct_ud_verbs_iface_poll_tx(uct_ud_verbs_iface_t *iface, int is_async)
...
+{
+  /* The GRH will contain either an IPv4 or IPv6 header. If the former is
+   * present the header will start at offset 20 in the buffer otherwise it
+   * will start at offset 0. Since the two headers are of fixed size (20 or
+   * 40 bytes) this means we will either see 0x6? at offset 0 (IPv6) or 0x45
+   * at offset 20. The detection is a little tricky for IPv6 given that the
+   * first 20B are undefined for IPv4. To overcome this the first byte of
+   * the posted receive buffer is set to 0xff.
+   */
+  return ((((unsigned char *)packet)[0] & 0xf0) == 0x60) ? UCS_IPV6_ADDR_LEN :
```
yosefe, line 407: minor: can use uint8_t instead on 'unsigned char'

src/uct/ib/ud/verbs/ud_verbs.c
```diff
@@ -706,6 +722,14 @@ uct_ud_verbs_iface_post_recv_always(uct_ud_verbs_iface_t *iface, int max)
         return;
     }
 
+    /* Set the first byte in the receive buffer grh to a known value not equal to
+     * 0x6?. This should aid in the detection of IPv6 vs IPv4 because the first
+     * byte is undefined in the later and 0x6? in the former. It is unlikely
+     * this byte is touched with IPv4. */
+    for (i = 0 ; i < count ; ++i) {
+        ((unsigned char *)wrs[i].sg.addr)[0] = 0xff;
```
yosefe, line 730: minor: can use uint8_t instead on 'unsigned char'

src/uct/ib/mlx5/ib_mlx5.inl
hjelmn: Ok, that appears to be inconsistent in this file with some having 1 line and others having 2 lines between methods. Will delete the line.

src/uct/ib/ud/verbs/ud_verbs.c
hjelmn: done.

src/uct/ib/ud/verbs/ud_verbs.c
hjelmn: done.

brminich (approved):

yosefe (approved):

PR #10210, opened 2024-10-04
PERF/UCT: increase window size for put_bw and get tests

evgeny-leksikov:
## What
increase window size for put_bw and get tests

## Why ?
to reach FWS with GGA transport (https://github.com/openucx/ucx/pull/9993)

### DIFF
--- a/src/tools/perf/perftest.c
+++ b/src/tools/perf/perftest.c
@@ -54,7 +54,10 @@ test_type_t tests[] = {
      "active message bandwidth / message rate", "overhead", 1},
 
     {"put_bw", UCX_PERF_API_UCT, UCX_PERF_CMD_PUT, UCX_PERF_TEST_TYPE_STREAM_UNI,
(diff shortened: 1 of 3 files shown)


src/tools/perf/perftest.c
```diff
@@ -39,7 +39,7 @@ test_type_t tests[] = {
      "atomic add latency", "latency", 1},
 
     {"get", UCX_PERF_API_UCT, UCX_PERF_CMD_GET, UCX_PERF_TEST_TYPE_STREAM_UNI,
-     "get latency / bandwidth / message rate", "latency", 1},
+     "get latency / bandwidth / message rate", "latency", 32},
```
yosefe, line 42: we should create 2 new tests here: get and get_bw.
since "get" will not test "latency" anymore, when using larger window.
and add the new test to CI

evgeny-leksikov:
TODO: check that new test is running in CI

yosefe (approved):

PR #10212, opened 2024-10-04
CI: Fix spellcheck on generated files

tvegas1:
## What
Don't write to inapropriate files when running `codespell --write-changes`.


brminich (approved):

.codespellrc
```diff
@@ -1,3 +1,3 @@
 [codespell]
-skip = ./test/gtest/common,./src/ucs/datastruct/sglib.h,./docs/doxygen/Architecture*,./docs/doxygen/header.tex.in,./src/ucm/ptmalloc286,./config/m4/ax_prog_doxygen.m4,./VENV/*
+skip = ./test/gtest/common,./src/ucs/datastruct/sglib.h,./docs/doxygen/Architecture*,./docs/doxygen/header.tex.in,./src/ucm/ptmalloc286,./config/m4/ax_prog_doxygen.m4,./VENV/*,./ltmain.sh,./configure,./config.guess,./config.sub,./.git,./depcomp,./libtool,./config/m4/libtool.m4
```
yosefe, line 2: can we use '\' to split long line?

tvegas1: I think no, `ConfigParser` supports it, but `codespell` is not able to sanitize its output.

```
['./test/gtest/common,\\\n./src/ucm/ptmalloc286/malloc.c']
```
and

```
 93 class GlobMatch(object):
 94     def __init__(self, pattern):
 95         if pattern:
 96             # Pattern might be a list of comma-delimited strings
 97             self.pattern_list = ','.join(pattern).split(',')
 ```

tvegas1:
https://github.com/codespell-project/codespell/issues/3561

PR #10216, opened 2024-10-09
UCP/PROTO: Remove extra remote-cpu overhead for zcopy

ivankochin:
## What
Remove extra-cpu overhead for zcopy operations

## Why ?
In terms of current performance estimation logic zcopy and bcopy are same operations in terms of transport that just performed on differrent buffers, so there shouldn't be difference at least for async (multi) case.

To fix degradations, https://github.com/openucx/ucx/pull/10217 and https://github.com/openucx/ucx/pull/10213 are also required

### DIFF
--- a/src/ucp/proto/proto_init.c
+++ b/src/ucp/proto/proto_init.c
@@ -168,11 +168,14 @@ ucp_proto_init_add_tl_perf(const ucp_proto_common_init_params_t *params,
                                                           tl_perf->bandwidth;
     }
 
-    if (/* Protocol is waiting for response */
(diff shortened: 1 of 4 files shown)


yosefe (dismissed):

ivankochin:
Performance checked, no degradations that can be considered as stopper for merge

brminich (dismissed):

ivankochin:
@brminich @yosefe I accidently pushed unrelated commit to that branch so your approves were dismissed. Could you please approve again?

brminich (approved):

yosefe (approved):

PR #10217, opened 2024-10-09
UCT/GDRCOPY: Add rcache overhead for GH systems

ivankochin:
## What
Increase gdrcopy cpu overhead to 200ns on GH. Additional 170ns is rcache lookup.

## Why ?
Current estimation doesn't include rcache lookup for gdrcopy which leads to performance degradations.

To fix degradations, #10216 and #10213 are also required


yosefe:
what is the "gdrcopy latency brnchmark"?
what is the latency we get with ucx_perftest with gdr_copy (UCT level, loopback)?

yosefe (approved):

brminich (approved):

PR #10220, opened 2024-10-10
UCT/GDRCOPY: Add rcache overhead for AMD Milan systems

ivankochin:
## What
Add rcache lookup overhead to gdrcopy perf estimation. Now it is just part of CPU overhead, but later should be moved to separate perf node.

## Why ?
To fix broad degradation for osu_latency inter-node d2d d2h on rock cluster.


brminich (dismissed):

ivankochin:
During performance testing on DGX several significant degradations were detected so it was decided to patch latency only for AMD Milan systems.

brminich (approved):

yosefe (approved):

PR #10221, opened 2024-10-10
UCT/TEST: Fixed race condition in test_sockaddr

shasson5:
## What
Fixed race condition when waiting for client disconnect event.

## Why ?
Prevent `concurrent_handle_conn_req_and_close_listener` test failure with the following error:
```
{{2024-09-25T05:13:37.8435372Z Value of: m_ep_client_data.empty()}}
{{2024-09-25T05:13:37.8436767Z   Actual: false}}
{{2024-09-25T05:13:37.8438018Z Expected: true}}
```

### DIFF
--- a/test/gtest/uct/test_sockaddr.cc
+++ b/test/gtest/uct/test_sockaddr.cc
@@ -547,7 +547,6 @@ protected:
         remote_data = connect_args->remote_data;
         status      = connect_args->status;
 
-        self->handle_client_connecting_status(status);
         if (status == UCS_OK) {
             EXPECT_TRUE(ucs_test_all_flags(remote_data->field_mask,
                                            (UCT_CM_REMOTE_DATA_FIELD_CONN_PRIV_DATA_LENGTH |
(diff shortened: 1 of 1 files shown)


test/gtest/uct/test_sockaddr.cc
```diff
@@ -565,6 +564,8 @@ class test_uct_sockaddr : public uct_test {
         } else {
             self->del_user_data(sa_user_data);
         }
+
+        self->handle_client_connecting_status(status);
```
yosefe, line 568: why moving handle_client_connecting_status to the end fixed the issue? it just updates m_state

shasson5: there is a race condition between `m_state` and `m_ep_client_data` which can be observed in line 1094:
```
    wait_for_bits(&m_state, TEST_STATE_SERVER_REJECTED |
                            TEST_STATE_CLIENT_GOT_REJECT);
```

yosefe: So the race is because handle_client_connecting_status adds TEST_STATE_CLIENT_GOT_REJECT to m_state which is atomic
pls add comment before line 568, swap order of lines 562-563 and make counters like m_client_connect_cb_cnt also atomic:
```c
    volatile int           m_client_connect_cb_cnt,
                           m_server_connect_cb_cnt;
    volatile int           m_server_disconnect_cnt, m_client_disconnect_cnt;
```

shasson5:
@yosefe please review

yosefe (approved):

PR #6745, opened 2021-05-01
UCP/RKEY/TEST: Refactor rkey pack/unpack and tests

yosefe:
# Why
Preparation for passing system topology information in rkey buffer

PR #4153, opened 2019-09-04
IFACE/STUB-AM-HANDLER: added dump of payload

hoopoepg:
- added dump of payload (first 64 bytes) into
  stub AM handler to simplify debugging

PR #10223, opened 2024-10-11
UCP/ASAN: Fixed ASAN detected buffer overflow

iyastreb:
## What
When running ucp_perftest_daemon with ASAN and trace logs, 2 buffer overflows are detected, both related to HEX dump of mkey/rkey. The regression was introduced in PR: https://github.com/openucx/ucx/pull/6745

The original hex dump was correct: https://github.com/openucx/ucx/pull/4153/files#diff-81ab501a3a76470dfbfe72e890118aa3083efe0434532272d12f2d5614047ec7R92
Then the impl was refactored to use common serialization API. And the issue is that `ucs_serialize_next_raw` already advances the pointer, so we dump the garbage value behind the key bytes

### DIFF
--- a/src/ucp/core/ucp_rkey.c
+++ b/src/ucp/core/ucp_rkey.c
@@ -174,7 +174,8 @@ UCS_PROFILE_FUNC(ssize_t, ucp_rkey_pack_memh,
         }
 
         ucs_trace("rkey %s for md[%d]=%s",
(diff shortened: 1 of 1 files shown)


brminich (approved):

brminich:
@yosefe, seems to be low risk for 1.18

tvegas1 (approved):

PR #10226, opened 2024-10-14
UCS/MPMC: Don't assume pointer validity without lock

tvegas1:
## What
Don't check for pointer validity when accessing without lock.

## Why ?
We cannot assume stable memory content outside of lock.

## How ?
Add and use specific function for unlocked accesses as it's racing with either free or alloc path.

### Original failure
```
[ RUN      ] test_mpmc.multi_threaded <> <>
[swx-rdmz-ucx-gpu-01:31308:4:35972]       debug.c:1339 Fatal: (queue)->ptail at: 0x6020002503b0 is poisoned
```

### New log
```
queue.h:73   Assertion `!__asan_region_is_poisoned((void*)((queue)->ptail), sizeof(*(queue)->ptail))' failed: (queue)->ptail: address=0x602000055ed0 size=8
```

### DIFF
--- a/src/ucs/datastruct/mpmc.c
+++ b/src/ucs/datastruct/mpmc.c
@@ -59,7 +59,7 @@ ucs_status_t ucs_mpmc_queue_pull(ucs_mpmc_queue_t *mpmc, uint64_t *value_p)
     ucs_status_t status = UCS_ERR_NO_PROGRESS;
     ucs_mpmc_elem_t *elem;
 
-    if (ucs_queue_is_empty(&mpmc->queue)) {
+    if (ucs_mpmc_queue_is_empty(mpmc)) {
         return status;
     }
 
--- a/src/ucs/datastruct/mpmc.h
+++ b/src/ucs/datastruct/mpmc.h
@@ -93,7 +93,7 @@ void ucs_mpmc_queue_remove_if(ucs_mpmc_queue_t *mpmc,
  */
 static inline int ucs_mpmc_queue_is_empty(ucs_mpmc_queue_t *mpmc)
 {
-    return ucs_queue_is_empty(&mpmc->queue);
+    return ucs_queue_is_empty_no_deref(&mpmc->queue);
 }
 
 #endif
--- a/src/ucs/datastruct/queue.h
+++ b/src/ucs/datastruct/queue.h
@@ -58,12 +58,20 @@ ucs_queue_is_tail(ucs_queue_head_t *queue, ucs_queue_elem_t *elem)
 }
 
 /**
+ * @return Whether the queue is empty without assumption on pointer validity
+ */
+static inline int ucs_queue_is_empty_no_deref(const ucs_queue_head_t *queue)
+{
+    return queue->ptail == &queue->head;
+}
+
+/**
  * @return Whether the queue is empty.
  */
 static inline int ucs_queue_is_empty(const ucs_queue_head_t *queue)
 {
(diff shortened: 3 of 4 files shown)


src/ucs/datastruct/mpmc.c
```diff
@@ -59,7 +59,7 @@ ucs_status_t ucs_mpmc_queue_pull(ucs_mpmc_queue_t *mpmc, uint64_t *value_p)
     ucs_status_t status = UCS_ERR_NO_PROGRESS;
     ucs_mpmc_elem_t *elem;
 
-    if (ucs_queue_is_empty(&mpmc->queue)) {
+    if (ucs_queue_is_empty_no_deref(&mpmc->queue)) {
```
iyastreb, line 62: I doubt this is the best way to fix it.
I see the only usage of `ucs_mpmc_queue_pull` in production code:
```C
async.c
    while (!ucs_mpmc_queue_is_empty(&async->missed)) {
        status = ucs_mpmc_queue_pull(&async->missed, &value);
```
So from there we also check the queue size without taking a mutex, which may end up with the same crash, right?
I think we better change this function:
```C
static inline int ucs_mpmc_queue_is_empty(ucs_mpmc_queue_t *mpmc)
{
    return ucs_queue_is_empty(&mpmc->queue);
}
```
And call it from ucs_mpmc_queue_pull maybe?

iyastreb: In addition to that, I'm not sure at all what's the advantage of this micro-optimisation.
It's fragile and error-prone, and also not surer reliable. Accessing shared data without a lock is typically not a good idea. Without acquiring a lock the changes made in one thread might not be visible in another one until some memory fence call occurs. IMO better do it properly with spin lock acquisition, especially if it is not in the hot path

tvegas1: please note the intent of that code (meant for busy polling as i understand it) is to check if it _could_ be not empty. doing that, there is _no_ dereference. if it _could_ be not empty, then the code checks again, under the lock this time to confirm. with contention i saw (huge) benefit with this scheme.

also the crash is purely asan checking related, as it implicitly dereferences the `->tail` pointer.

iyastreb: Right, I see. Basically what I wanted to say in the last comment is that this check is not reliable, because without acquiring a lock thread might not see the write from another thread, meaning that we may return even if queue is not empty (due to visibility across threads).
That may be totally fine to skip updates, will catch up on them later, if you say performance gain is big

And what do you think about my first comment: since ucs_mpmc_queue_is_empty is actually used to check the queue size, it should be also probably using `no_deref` version

iyastreb: ah, you already implemented that)
Maybe just use `ucs_mpmc_queue_is_empty` here?

tvegas1: fixed

iyastreb (approved):

brminich:
why is `__asan_region_is_poisoned` assertion triggered if you removed the check from `ucs_mpmc_queue_pull`?

tvegas1:
> why is `__asan_region_is_poisoned` assertion triggered if you removed the check from `ucs_mpmc_queue_pull`?

With this PR, no assertion can be seen anymore. As I also changed, the assertion line format with added `size`, I mentioned the resulting output, for reference. Please let me know if I got your question right.

brminich (approved):

yosefe (approved):

PR #10228, opened 2024-10-14
GGA: Always use exported_mkey_packed_size for mkey pack

iyastreb:
## What
When packing mkey GGA always sets `UCT_MD_MKEY_PACK_FLAG_EXPORT` which forces to create a mkey for export of size `exported_mkey_packed_size` (even if not explicitly requested by caller). This is unexpected by the caller, who allocates an array of `rkey_packed_size` to handle the rkey.

Currently it works by chance, because `rkey_packed_size > exported_mkey_packed_size`.
But it becomes an issue when we increase the `uct_ib_md_packed_mkey_t` (like we do for XGVMI purposes).
The fix is to set `rkey_packed_size = exported_mkey_packed_size` for GGA transport to match the pack behavior

### DIFF
--- a/src/uct/ib/mlx5/gga/gga_mlx5.c
+++ b/src/uct/ib/mlx5/gga/gga_mlx5.c
@@ -73,6 +73,21 @@ ucs_config_field_t uct_gga_mlx5_iface_config_table[] = {
 };
 
 static ucs_status_t
+uct_ib_mlx5_gga_md_query(uct_md_h uct_md, uct_md_attr_v2_t *md_attr)
+{
+    ucs_status_t status;
+
+    status = uct_ib_mlx5_devx_md_query(uct_md, md_attr);
+    if (status != UCS_OK) {
+        return status;
+    }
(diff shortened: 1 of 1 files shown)


evgeny-leksikov (approved):

brminich (approved):

yosefe (approved):

yosefe:
@iyastreb do we need this fix for v1.18.x?

evgeny-leksikov:
> do we need this fix for v1.18.x?

I don't think so, unless we increase `uct_ib_md_packed_mkey_t` which is part of other activity

iyastreb:
> @iyastreb do we need this fix for v1.18.x?

I second what Evgeny said. 1.18 works without fix because `rkey_packed_size > exported_mkey_packed_size`.
If for some reason we extend `uct_ib_md_packed_mkey_t` in 1.18 then we need this fix.
But since there is no such plan - I think we don't need it

PR #10230, opened 2024-10-15
UCP/PROTO: Fix RNDV_SCHEME logic

ivankochin:
## What
Change `RNDV_SCHEME` and `RNDV_THRESH` logic to work properly.

## Why ?
Currently if user set certain `RNDV_SCHEME` without setting `RNDV_THRESH` this scheme would be forced on all sizes since `cfg_thresh` for this protocol would be `0`. This change fixes that.

## How ?
With this change remote variant can report only 2 values as `cfg_thresh`: `AUTO` and `INF`. `AUTO` reported for:
- all protocols if `RNDV_SCHEME=auto`
- protocol that is set in `RNDV_SCHEME`

`INF` is reported for all protocols that are not set in `RNDV_SCHEME` if `RNDV_SCHEME` is set to some certain protocol.

## Manual testing cases:
| RNDV_THRESH| RNDV_SCHEME| RESULT|
|--------|--------|--------|
| AUTO| AUTO| All protocols are selected basing on perf comparison|
| AUTO| GET_ZCOPY| No hard thresh, but only GET_ZCOPY protocol participates in proto selection |
| AUTO| PUT_ZCOPY| No hard thresh, but only GET_ZCOPY protocol participates in proto selection |
| 1024| AUTO| Starting from 1024 RNDV is forced but which RNDV protocol would be used decided by proto selection |
| 1024| GET_ZCOPY| RNDV_GET used starting from 1024 |
| 1024| PUT_ZCOPY| RNDV_PUT used starting from 1024 |
| INF| AUTO| RNDV is not used |
| INF| GET_ZCOPY| RNDV is not used  |
| INF| PUT_ZCOPY| RNDV is not used |

### DIFF
--- a/src/ucp/rndv/proto_rndv.c
+++ b/src/ucp/rndv/proto_rndv.c
@@ -395,24 +395,14 @@ static void ucp_proto_rndv_ctrl_variant_probe(
     /* Set priority and threshold for this variant */
     cfg_thresh   = params->super.cfg_thresh;
     cfg_priority = params->super.cfg_priority;
-    if ((remote_proto->cfg_thresh != UCS_MEMUNITS_INF) &&
-        (remote_proto->cfg_thresh != UCS_MEMUNITS_AUTO)) {
-        /* Consider remote priority and threshold only if RNDV_SCHEME or
-            * BCOPY/ZCOPY thresh are set to force these settings */
-        cfg_priority = remote_proto->cfg_priority;
-        cfg_thresh   = (cfg_thresh == UCS_MEMUNITS_AUTO) ?
-                                 remote_proto->cfg_thresh :
-                                 ucs_max(cfg_thresh, remote_proto->cfg_thresh);
+    if (remote_proto->cfg_thresh != UCS_MEMUNITS_AUTO) {
+        /* If RNDV_SCHEME is set, all protocols except forced one report INF */
+        ucs_assertv(remote_proto->cfg_thresh == UCS_MEMUNITS_INF,
+                    "variant_name=%s remote_proto->cfg_thresh=%zu",
+                    variant_name, remote_proto->cfg_thresh);
+        cfg_thresh = remote_proto->cfg_thresh;
     }
 
-    /* Remote variants priorities are used to respect RNDV_SCHEME setting
-     * so they should contain value greater than CTRL message `cfg_thresh`.
-     * Equality is allowed for RTR remote variants.
-     */
-    ucs_assertv(params->super.cfg_priority <= remote_proto->cfg_priority,
-                "remote_proto=%s params->super.cfg_priority=%u "
-                "remote_proto->cfg_priority=%u", variant_name,
-                params->super.cfg_priority, remote_proto->cfg_priority);
     ucp_proto_select_add_proto(&params->super.super, cfg_thresh, cfg_priority,
                                perf, rpriv, priv_size);
 
--- a/src/ucp/rndv/proto_rndv.inl
+++ b/src/ucp/rndv/proto_rndv.inl
@@ -20,15 +20,14 @@
 static UCS_F_ALWAYS_INLINE size_t
 ucp_proto_rndv_cfg_thresh(ucp_context_h context, uint64_t rndv_modes)
 {
+    ucp_rndv_mode_t mode = context->config.ext.rndv_mode;
     ucs_assert(!(rndv_modes & UCS_BIT(UCP_RNDV_MODE_AUTO)));
 
-    if (context->config.ext.rndv_mode == UCP_RNDV_MODE_AUTO) {
-        return UCS_MEMUNITS_AUTO; /* automatic threshold */
-    } else if (rndv_modes & UCS_BIT(context->config.ext.rndv_mode)) {
-        return 0; /* enabled by default */
-    } else {
-        return UCS_MEMUNITS_INF; /* used only as last resort */
+    if ((mode == UCP_RNDV_MODE_AUTO) || (rndv_modes & UCS_BIT(mode))) {
+        return UCS_MEMUNITS_AUTO;
     }
+
+    return UCS_MEMUNITS_INF; /* used only as last resort */
 }
 
 static UCS_F_ALWAYS_INLINE ucs_status_t
--- a/test/gtest/ucp/test_ucp_am.cc
+++ b/test/gtest/ucp/test_ucp_am.cc
@@ -1627,14 +1627,18 @@ public:
                                  void *data, size_t length,
                                  const ucp_am_recv_param_t *rx_param)
     {
-        EXPECT_TRUE(rx_param->recv_attr & UCP_AM_RECV_ATTR_FLAG_RNDV);
-        EXPECT_FALSE(rx_param->recv_attr & UCP_AM_RECV_ATTR_FLAG_DATA);
-
         ucs_status_t status = test_ucp_am_nbx::am_data_handler(header,
                                                                header_length,
                                                                data, length,
                                                                rx_param);
         EXPECT_FALSE(UCS_STATUS_IS_ERR(status));
+
+        if (!m_check_recv_rndv_flags) {
+            return status;
+        }
+
+        EXPECT_TRUE(rx_param->recv_attr & UCP_AM_RECV_ATTR_FLAG_RNDV);
+        EXPECT_FALSE(rx_param->recv_attr & UCP_AM_RECV_ATTR_FLAG_DATA);
 
         return UCS_INPROGRESS;
     }
@@ -1732,8 +1736,16 @@ public:
         return cfg->rndv_frag_size[mem_type];
     }
 
+    void check_rma_support()
+    {
+        if (!sender().is_rndv_supported()) {
+            UCS_TEST_SKIP_R("RNDV is not supported");
+        }
+    }
+
 protected:
     static constexpr unsigned RNDV_THRESH = 128;
+    bool m_check_recv_rndv_flags          = true;
     ucs_status_t m_status;
     bool m_am_recv_cb_invoked;
 };
@@ -1745,11 +1757,13 @@ UCS_TEST_P(test_ucp_am_nbx_rndv, rndv_auto, "RNDV_SCHEME=auto")
 
 UCS_TEST_P(test_ucp_am_nbx_rndv, rndv_get, "RNDV_SCHEME=get_zcopy")
 {
+    check_rma_support();
     test_am_send_recv(64 * UCS_KBYTE);
 }
 
 UCS_TEST_P(test_ucp_am_nbx_rndv, rndv_put, "RNDV_SCHEME=put_zcopy")
 {
+    check_rma_support();
     test_am_send_recv(64 * UCS_KBYTE);
 }
 
@@ -2084,6 +2098,8 @@ UCS_TEST_P(test_ucp_am_nbx_rndv_ppln, host_buff_host_frag,
            "RNDV_FRAG_MEM_TYPE=host")
 {
     // Host memory should not be pipelined thru host staging buffers
+    m_check_recv_rndv_flags = false;
+
     test_ppln_send(UCS_MEMORY_TYPE_HOST, 2, 0);
 }
 
--- a/test/gtest/ucp/test_ucp_sockaddr.cc
+++ b/test/gtest/ucp/test_ucp_sockaddr.cc
@@ -899,13 +899,18 @@ protected:
         return get_variant_value() & TEST_MODIFIER_SA_DATA_V1;
     }
 
-    bool has_rndv_lanes(ucp_ep_h ep)
+    bool has_rndv_lanes(const entity &entity)
     {
+        const auto *config   = &entity.worker()->context->config.ext;
+        const auto &ep       = entity.ep();
+        uint64_t iface_flags = (config->rndv_mode == UCP_RNDV_MODE_GET_ZCOPY) ?
+                                       UCT_IFACE_FLAG_GET_ZCOPY :
+                                       UCT_IFACE_FLAG_PUT_ZCOPY;
+
         for (ucp_lane_index_t lane_idx = 0;
              lane_idx < ucp_ep_num_lanes(ep); ++lane_idx) {
             if ((lane_idx != ucp_ep_get_cm_lane(ep)) &&
-                (ucp_ep_get_iface_attr(ep, lane_idx)->cap.flags &
-                 (UCT_IFACE_FLAG_GET_ZCOPY | UCT_IFACE_FLAG_PUT_ZCOPY)) &&
+                (ucp_ep_get_iface_attr(ep, lane_idx)->cap.flags & iface_flags) &&
                 /* RNDV lanes should be selected if transport supports GET/PUT
                  * Zcopy and: */
                 (/* - either memory invalidation can be done on its MD */
@@ -938,6 +943,10 @@ protected:
         constexpr size_t length = 4 * UCS_KBYTE;
 
         listen_and_communicate(false, SEND_DIRECTION_BIDI);
+
+        if (!has_rndv_lanes(sender())) {
+            UCS_TEST_SKIP_R("no RNDV lanes");
+        }
 
         mem_buffer send_buffer(length, UCS_MEMORY_TYPE_HOST);
         send_buffer.pattern_fill(1, length);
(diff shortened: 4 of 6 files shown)


ivankochin:
I think it worth covering this logic by tests. If we merge https://github.com/openucx/ucx/pull/9989 it would allow to cover it by the following way:

```c++

UCS_TEST_P(test_ucp_proto_mock, rndv_thresh, "RNDV_THRESH=1024", "NET_DEVICES=mlx5_0:1")
{
    setup_mock("ib").set_mock_iface_attr("rc_mlx5/mlx5_0:1",
        [](uct_iface_attr_t &iface_attr) {
            iface_attr.dev_num_paths    = 1;
            iface_attr.cap.am.max_short = 208;
            iface_attr.bandwidth.shared = 10000000000;
            iface_attr.latency.c        = 0.000006;
            iface_attr.latency.m        = 0.000000001;
        });

    connect();

    ucp_proto_select_key_t key = any_key();
    key.param.op_id_flags      = UCP_OP_ID_AM_SEND;
    key.param.op_attr          = 0;

    check_ep_config(sender(), {
        {"0",          "200",    "short",                "rc_mlx5/mlx5_0:1"},
        {"201",        "1023",   "copy-in",              "rc_mlx5/mlx5_0:1"},
        {"1024",       "91894",  "rendezvous fragmented copy-in copy-out",
                                                         "rc_mlx5/mlx5_0:1"},
        {"91895",       "inf",   "rendezvous zero-copy read from remote",
                                                         "rc_mlx5/mlx5_0:1"},
    }, key);
}

UCS_TEST_P(test_ucp_proto_mock, rndv_thresh_and_scheme, "RNDV_THRESH=1024", "RNDV_SCHEME=get_zcopy", "NET_DEVICES=mlx5_0:1")
{
    setup_mock("ib").set_mock_iface_attr("rc_mlx5/mlx5_0:1",
        [](uct_iface_attr_t &iface_attr) {
            iface_attr.dev_num_paths    = 1;
            iface_attr.cap.am.max_short = 208;
            iface_attr.bandwidth.shared = 10000000000;
            iface_attr.latency.c        = 0.000006;
            iface_attr.latency.m        = 0.000000001;
        });

    connect();

    ucp_proto_select_key_t key = any_key();
    key.param.op_id_flags      = UCP_OP_ID_AM_SEND;
    key.param.op_attr          = 0;

    check_ep_config(sender(), {
        {"0",          "200",    "short",                "rc_mlx5/mlx5_0:1"},
        {"201",        "1023",   "copy-in",              "rc_mlx5/mlx5_0:1"},
        {"1024",       "inf",    "rendezvous zero-copy read from remote",
                                                         "rc_mlx5/mlx5_0:1"},
    }, key);
}

```

src/ucp/rndv/proto_rndv.inl
```diff
@@ -22,13 +22,11 @@ ucp_proto_rndv_cfg_thresh(ucp_context_h context, uint64_t rndv_modes)
...
-        return UCS_MEMUNITS_AUTO; /* automatic threshold */
-    } else if (rndv_modes & UCS_BIT(context->config.ext.rndv_mode)) {
-        return 0; /* enabled by default */
-    } else {
-        return UCS_MEMUNITS_INF; /* used only as last resort */
+    if ((context->config.ext.rndv_mode == UCP_RNDV_MODE_AUTO) ||
+        (rndv_modes & UCS_BIT(context->config.ext.rndv_mode))) {
+        return UCS_MEMUNITS_AUTO; /* used only as last resort */
     }
+    return UCS_MEMUNITS_INF;
```
rakhmets, line 29: Looks like we only need rndv mode in this function.
```C
ucp_proto_rndv_cfg_thresh(ucp_rndv_mode_t rndv_mode, uint64_t rndv_modes)
{
    ucs_assert(!(rndv_modes & UCS_BIT(UCP_RNDV_MODE_AUTO)));

    if ((rndv_mode == UCP_RNDV_MODE_AUTO) ||
        (rndv_modes & UCS_BIT(rndv_mode))) {
        return UCS_MEMUNITS_AUTO;
    }

    return UCS_MEMUNITS_INF;
```

src/ucp/rndv/proto_rndv.inl
```diff
@@ -22,13 +22,11 @@ ucp_proto_rndv_cfg_thresh(ucp_context_h context, uint64_t rndv_modes)
...
 
-    if (context->config.ext.rndv_mode == UCP_RNDV_MODE_AUTO) {
-        return UCS_MEMUNITS_AUTO; /* automatic threshold */
-    } else if (rndv_modes & UCS_BIT(context->config.ext.rndv_mode)) {
-        return 0; /* enabled by default */
-    } else {
-        return UCS_MEMUNITS_INF; /* used only as last resort */
+    if ((context->config.ext.rndv_mode == UCP_RNDV_MODE_AUTO) ||
+        (rndv_modes & UCS_BIT(context->config.ext.rndv_mode))) {
+        return UCS_MEMUNITS_AUTO; /* used only as last resort */
```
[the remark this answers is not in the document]
ivankochin: done

brminich (dismissed):

src/ucp/rndv/proto_rndv.c
```diff
@@ -395,24 +395,13 @@ static void ucp_proto_rndv_ctrl_variant_probe(
...
-     * Equality is allowed for RTR remote variants.
-     */
-    ucs_assertv(params->super.cfg_priority <= remote_proto->cfg_priority,
-                "remote_proto=%s params->super.cfg_priority=%u "
-                "remote_proto->cfg_priority=%u", variant_name,
-                params->super.cfg_priority, remote_proto->cfg_priority);
+    if (remote_proto->cfg_thresh != UCS_MEMUNITS_AUTO) {
+        /* If RNDV_SCHEME is set, all protocols except forced one report INF */
+        ucs_assertv(remote_proto->cfg_thresh == UCS_MEMUNITS_INF,
+                    "remote_proto->cfg_thresh=%zu", remote_proto->cfg_thresh);
```
ivankochin, line 401:
```suggestion
                    "variant_name=%s remote_proto->cfg_thresh=%zu",
                    variant_name, remote_proto->cfg_thresh);
```

brminich (dismissed):

ivankochin:
Drawback of the proposed logic is that if defined `RNDV_SCHEME` is unavailable, `RNDV_THRESH` is not considered at all. 

So if we set `RNDV_SCHME=put_zcopy` and `RNDV_THRESH=0` and there are no `rma_bw_lanes` protocols selection wouldn't be fallen back to `rndv/am/bcopy/zcopy` and eager still can be chosen. That can cause many changes in tests which are relates on that fallback.

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3257,7 +3274,7 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_err_sender,
 {
     size_t num_sends = ucs_max(100, 100000 / ucs::test_time_multiplier() /
                                     ucs::test_time_multiplier());
-    do_tag_rndv_killed_sender_test(4, 128, num_sends);
+    do_tag_rndv_killed_sender_test(4, 130, num_sends);
```
ivankochin, line 3277:
```suggestion
    do_tag_rndv_killed_sender_test(4, 1024, num_sends);
```

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3248,7 +3265,7 @@ UCS_TEST_P(test_ucp_sockaddr_protocols_err_sender,
 {
     size_t num_sends = ucs_max(100, 100000 / ucs::test_time_multiplier() /
                                     ucs::test_time_multiplier());
-    do_tag_rndv_killed_sender_test(1, 128, num_sends);
+    do_tag_rndv_killed_sender_test(1, 130, num_sends);
```
ivankochin, line 3268:
```suggestion
    do_tag_rndv_killed_sender_test(1, 1024, num_sends);
```

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -899,13 +899,23 @@ class test_ucp_sockaddr : public ucp_test {
         return get_variant_value() & TEST_MODIFIER_SA_DATA_V1;
     }
 
-    bool has_rndv_lanes(ucp_ep_h ep)
+    bool has_rndv_lanes(const entity &entity)
     {
+        const auto *config   = &sender().worker()->context->config.ext;
+        const auto &ep       = entity.ep();
+        uint64_t iface_flags = UCT_IFACE_FLAG_GET_ZCOPY |
+                               UCT_IFACE_FLAG_PUT_ZCOPY;
```
rakhmets, line 907: The initialization is redundant.

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -899,13 +899,23 @@ class test_ucp_sockaddr : public ucp_test {
         return get_variant_value() & TEST_MODIFIER_SA_DATA_V1;
     }
 
-    bool has_rndv_lanes(ucp_ep_h ep)
+    bool has_rndv_lanes(const entity &entity)
     {
+        const auto *config   = &sender().worker()->context->config.ext;
```
rakhmets, line 904:
```suggestion
        const auto *config   = &entity.worker()->context->config.ext;
```

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -939,6 +949,10 @@ class test_ucp_sockaddr : public ucp_test {
 
         listen_and_communicate(false, SEND_DIRECTION_BIDI);
 
+        if (!has_rndv_lanes(sender())) {
+            UCS_TEST_SKIP_R("Has no RNDV lanes");
```
rakhmets, line 953: I think `mo RNDV lanes` is enough.

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3169,6 +3182,10 @@ class test_ucp_sockaddr_protocols_err_sender
             send_recv(sender(), receiver(), send_recv_type(), false, cb_type(),
                       sender_idx);
 
+            if (!has_rndv_lanes(sender())) {
+                UCS_TEST_SKIP_R("Has no RNDV lanes");
```
rakhmets, line 3186: The same as above.

test/gtest/ucp/test_ucp_sockaddr.cc
ivankochin: Done

test/gtest/ucp/test_ucp_sockaddr.cc
ivankochin: Done

test/gtest/ucp/test_ucp_sockaddr.cc
ivankochin: Done

test/gtest/ucp/test_ucp_sockaddr.cc
ivankochin: Done

rakhmets (approved):

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1732,8 +1736,17 @@ class test_ucp_am_nbx_rndv : public test_ucp_am_nbx_prereg {
         return cfg->rndv_frag_size[mem_type];
     }
 
+    void check_rma_support()
+    {
+        if (!sender().is_rndv_supported()) {
+            UCS_TEST_MESSAGE << "RNDV is not supported";
+            m_check_recv_rndv_flags = false;
```
brminich, line 1743: why? what if this is AM-based rndv?
also if rndv is not supported this test suite should be skipped

ivankochin: Maybe message confused you. That case is created to handler AM-based RNDV case. AFAIR AM RNDV doesn't set `recv_attr` flags. 

I can change message here to "RMA is not supported" would it be better from your point of view?

brminich: what flag do you mean? AM-based RNDV is using generic flow RTS->RTR->data chunks

ivankochin: My bad, you are right. The thing here is that due to new logic if `RNDV_SCHEME=get_zcopy` and rndv/get/zcopy is unavailable eager can be selected instead of RNDV.

So we cannot rely on fact that `UCP_AM_RECV_ATTR_FLAG_RNDV` and `UCP_AM_RECV_ATTR_FLAG_DATA`  will be set in `am_data_handler`

brminich: eager should not be selected because we set RNDV_THRESH in  `test_ucp_am_nbx_rndv`
Please either remove `check_rma_support` at all, or skip the test if RMA is not supported

src/ucp/rndv/proto_rndv.inl
```diff
@@ -20,15 +20,13 @@
...
-    if (context->config.ext.rndv_mode == UCP_RNDV_MODE_AUTO) {
-        return UCS_MEMUNITS_AUTO; /* automatic threshold */
-    } else if (rndv_modes & UCS_BIT(context->config.ext.rndv_mode)) {
-        return 0; /* enabled by default */
-    } else {
-        return UCS_MEMUNITS_INF; /* used only as last resort */
+    if ((mode == UCP_RNDV_MODE_AUTO) || (rndv_modes & UCS_BIT(mode))) {
+        return UCS_MEMUNITS_AUTO;
     }
+    return UCS_MEMUNITS_INF; /* used only as last resort */
```
yosefe, line 29: minor: space line before. or use else block

test/gtest/ucp/test_ucp_am.cc
yosefe: can we check sender().is_rndv_supported() directly in the callback and avoid m_check_recv_rndv_flags variable?

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -939,6 +944,10 @@ class test_ucp_sockaddr : public ucp_test {
 
         listen_and_communicate(false, SEND_DIRECTION_BIDI);
 
+        if (!has_rndv_lanes(sender())) {
+            UCS_TEST_SKIP_R("no RNDV lanes");
+        }
```
yosefe, line 949: why need this check now - how did it pass before?

test/gtest/ucp/test_ucp_sockaddr.cc
```diff
@@ -3169,6 +3177,10 @@ class test_ucp_sockaddr_protocols_err_sender
             send_recv(sender(), receiver(), send_recv_type(), false, cb_type(),
                       sender_idx);
 
+            if (!has_rndv_lanes(sender())) {
+                UCS_TEST_SKIP_R("no RNDV lanes");
+            }
```
yosefe, line 3182: why need this check now - how did it pass before?

src/ucp/rndv/proto_rndv.inl
ivankochin: Done

test/gtest/ucp/test_ucp_am.cc
ivankochin: Together with @brminich we've decided to skip this test if RNDV not supported. @yosefe does this option work for you? `m_check_recv_rndv_flags` still needed for `test_ucp_am_nbx_rndv_ppln.host_buff_host_frag` test

test/gtest/ucp/test_ucp_sockaddr.cc
ivankochin: Before that it worked through AM-based RNDV. This patch fixes `RNDV_SCHEME` but the drawback is that when both `RNDV_SCHEME` and `RNDV_THRESH` are set and `RNDV_SCHEME` protocol is not available, `RNDV_THRESH` is not considered anymore. 

So if in that case there is no RNDV lanes, eager can be chosen.

test/gtest/ucp/test_ucp_sockaddr.cc
ivankochin: See here: https://github.com/openucx/ucx/pull/10230/files#r1846321667

test/gtest/ucp/test_ucp_sockaddr.cc
yosefe: When RNDV_THRESH is not set and RNDV_SCHEME is not available, i would expect a fallback to AM-based rndv, since RDNV_THRESH should have higher "priority".

ivankochin: So if the only setting is "RNDV_SCHEME=get_zcopy" and this protocol is not available then falling back to AM-based RNDV is preferrable over eager protocols?

brminich: i guees @yosefe meant when RNDV_THRESH is set, like it is in this case

yosefe: > i guees @yosefe meant when RNDV_THRESH is set, like it is in this case

Right, i've meant set

yosefe (approved):

yosefe:
@ivankochin can you pls squash?

PR #10231, opened 2024-10-15
UCT/GGA: fix compilation --with-mlx5-dv=no --with-devx=no

evgeny-leksikov:
## What
fix build `--with-mlx5-dv=no --with-devx=no`

## Why ?
to fix non-default build configuration

### DIFF
--- a/buildlib/tools/builds.sh
+++ b/buildlib/tools/builds.sh
@@ -461,5 +461,5 @@ do
 	[ -d "${ucx_build_dir}" ] && rm -rf ${ucx_build_dir}/*
 
 	# run the test
-	$test_name || { azure_log_error "Test failed: $test_name"; exit 1; }
+	$test_name
 done
--- a/src/uct/ib/mlx5/ib_mlx5_log.c
+++ b/src/uct/ib/mlx5/ib_mlx5_log.c
@@ -239,13 +239,18 @@ static unsigned uct_ib_mlx5_parse_dseg(void **dseg_p, unsigned op_flags,
     return ds;
 }
 
-static void* uct_ib_mlx5_dump_dma_seg(uct_ib_mlx5_dma_seg_t *dma_seg,
-                                      char *buf, size_t max)
+static void* uct_ib_mlx5_dump_dma_seg(void *seg, char *buf, size_t max)
 {
+#if HAVE_MLX5_MMO
+    uct_ib_mlx5_dma_seg_t *dma_seg = seg;
+
     snprintf(buf, max, " DMA[lkey 0x%x va 0x%"PRIx64"]",
              be32toh(dma_seg->be_opaque_lkey),
              be64toh(dma_seg->be_opaque_vaddr));
     return dma_seg + 1;
+#else
+    ucs_fatal("WQE dump: unexpected DMA segment at %p", seg);
+#endif
 }
 
 static uint64_t network_to_host(void *ptr, int size)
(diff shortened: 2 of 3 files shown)


brminich:
is it possible to add some test to CI?

src/uct/ib/mlx5/ib_mlx5_log.c
```diff
@@ -239,6 +239,7 @@ static unsigned uct_ib_mlx5_parse_dseg(void **dseg_p, unsigned op_flags,
     return ds;
 }
 
+#if HAVE_MLX5_MMO
 static void* uct_ib_mlx5_dump_dma_seg(uct_ib_mlx5_dma_seg_t *dma_seg,
```
tvegas1, line 243: need to wrap `uct_ib_mlx5_dump_dma_seg` calls with `HAVE_MLX5_MMO` too, or do we have a way to put #if/#end inside `uct_ib_mlx5_dump_dma_seg`?

evgeny-leksikov:
> is it possible to add some test to CI?

actually, `--with-devx=no` is enough to reproduce and we have such test, probably it was disabled by some reason

src/uct/ib/mlx5/ib_mlx5_log.c
evgeny-leksikov: yeah, already caught and fixed

brminich:
> > is it possible to add some test to CI?
> 
> actually, `--with-devx=no` is enough to reproduce and we have such test, probably it was disabled by some reason

can you pls check why it did not caught the problem and fix that?

evgeny-leksikov:
> > > is it possible to add some test to CI?
> > 
> > 
> > actually, `--with-devx=no` is enough to reproduce and we have such test, probably it was disabled by some reason
> 
> can you pls check why it did not caught the problem and fix that?

https://dev.azure.com/ucfconsort/ucx/_build/results?buildId=90283&view=logs&j=b225495a-a138-58be-0e0c-72778e1ff6b6&t=c6c5013e-bc26-58e0-88ac-2d26030ca3ad&l=69067 the problem was caught but not handled

evgeny-leksikov:
thanks to @tvegas1 for the fix!
let's see how it goes and how many failures we missed

evgeny-leksikov:
link to another build fix https://github.com/openucx/ucx/pull/10235

tvegas1:
> link to another build fix #10235

We now see expected CI failures. #10235 will fix them.

buildlib/tools/builds.sh
```diff
@@ -461,5 +461,5 @@ do
 	[ -d "${ucx_build_dir}" ] && rm -rf ${ucx_build_dir}/*
 
 	# run the test
-	$test_name || { azure_log_error "Test failed: $test_name"; exit 1; }
+	$test_name
```
yosefe, line 464: how did it hid the build issues?

yosefe (approved):

buildlib/tools/builds.sh
tvegas1: when using `||` or `if !` on a bash function, seems `set -e` is ignored:
```shell
#!/bin/bash -eEl

run() {
    echo options set $-
    # grep returns error $? 2 like make
    grep abc abc
    echo Should never be seen
}

echo == broken 1
run || echo FAIL 1

echo == broken 2
if ! run; then
    echo FAIL 2
fi

echo == broken 3
( run ) || echo FAIL 3

echo == broken 3
{ run; } || echo FAIL 4

echo == success
run
echo Should not be seen 2
```

```
$ ./a.sh
== broken 1
options set ehBE
grep: abc: No such file or directory
Should never be seen
== broken 2
options set ehBE
grep: abc: No such file or directory
Should never be seen
== broken 3
options set ehBE
grep: abc: No such file or directory
Should never be seen
== broken 3
options set ehBE
grep: abc: No such file or directory
Should never be seen
== success
options set ehBE
grep: abc: No such file or directory
```

PR #8584, opened 2022-10-01
UCP/CORE: Implement exported/imported memh

dmitrygx:
## What

 Implement exported/imported memh.

## Why ?

To implement support of memory handles which could be used on DPU side to perform AM/TAG/RMA/AMO operations without involving HOST side into this process.

## How ?

1. Add implementation of API introduced in https://github.com/openucx/ucx/pull/8489 and https://github.com/openucx/ucx/pull/8577.
2. Add RCACHE for exported/imported memory handles.

PR #10233, opened 2024-10-16
UCP: Fixed stack overflow in exported rkey unpack

iyastreb:
## What
Fix for the segfault described in [RM#4105410](https://redmine.mellanox.com/issues/4105410)
The root cause is that export unpack logic stores all **combinations** of remote_md-local_md in 
`tl_mkeys[UCP_MAX_MDS]` array, which is not sufficient to store all of them. WIth large number of IB devices (like on nemo01) this overrides memory on stack and leads to stack corruption, therefore we don't see the backtrace.
This is initial behaviour introduced in https://github.com/openucx/ucx/pull/8584

## How ?
- Changed `ucp_unpacked_exported_tl_mkey_t` structure to store MD map instead of an individual index, to avoid storing combinations.
- Changed impl of key export to deal with `local_md_map`
- Added assertion to avoid stack overflow in the future

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -1774,37 +1774,38 @@ ucp_memh_import_attach(ucp_context_h context, ucp_mem_h memh,
     uct_mem_h uct_memh;
 
     for (tl_mkey_index = 0; tl_mkey_index < num_tl_mkeys; ++tl_mkey_index) {
-        md_index    = tl_mkeys[tl_mkey_index].md_index;
         tl_mkey_buf = tl_mkeys[tl_mkey_index].tl_mkey_buf;
-        md_attr     = &context->tl_mds[md_index].attr;
-        ucs_assert_always(md_attr->flags & UCT_MD_FLAG_EXPORTED_MKEY);
+        ucs_for_each_bit(md_index, tl_mkeys[tl_mkey_index].local_md_map) {
+            md_attr = &context->tl_mds[md_index].attr;
+            ucs_assert_always(md_attr->flags & UCT_MD_FLAG_EXPORTED_MKEY);
 
-        if (memh->uct[md_index] != NULL) {
-            continue;
-        }
+            if (memh->uct[md_index] != NULL) {
+                continue;
+            }
 
-        attach_params.field_mask = UCT_MD_MEM_ATTACH_FIELD_FLAGS;
-        attach_params.flags      = UCT_MD_MEM_ATTACH_FLAG_HIDE_ERRORS;
+            attach_params.field_mask = UCT_MD_MEM_ATTACH_FIELD_FLAGS;
+            attach_params.flags      = UCT_MD_MEM_ATTACH_FLAG_HIDE_ERRORS;
 
-        status = uct_md_mem_attach(context->tl_mds[md_index].md, tl_mkey_buf,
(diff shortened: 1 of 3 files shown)


tvegas1 (dismissed):

brminich (approved):

evgeny-leksikov (dismissed):

yosefe:
Fix needed also for v1.18.x branch

src/ucp/core/ucp_mm.c
```diff
@@ -1774,37 +1774,38 @@ ucp_memh_import_attach(ucp_context_h context, ucp_mem_h memh,
     uct_mem_h uct_memh;
 
     for (tl_mkey_index = 0; tl_mkey_index < num_tl_mkeys; ++tl_mkey_index) {
-        md_index    = tl_mkeys[tl_mkey_index].md_index;
-        tl_mkey_buf = tl_mkeys[tl_mkey_index].tl_mkey_buf;
-        md_attr     = &context->tl_mds[md_index].attr;
-        ucs_assert_always(md_attr->flags & UCT_MD_FLAG_EXPORTED_MKEY);
+        ucs_for_each_bit(md_index, tl_mkeys[tl_mkey_index].local_md_map) {
+            tl_mkey_buf = tl_mkeys[tl_mkey_index].tl_mkey_buf;
```
yosefe, line 1778: minor: we can initialize it outside the ucs_for_each_bit loop because it depends only on tl_mkey_index

src/ucp/core/ucp_rkey.c
```diff
@@ -558,12 +557,13 @@ ucp_memh_exported_unpack(ucp_context_h context, const void *export_mkey_buffer,
...
-            tl_mkey->tl_mkey_buf = tl_mkey_buf;
-            ++unpacked->num_tl_mkeys;
-        }
+        ucs_assertv(unpacked->num_tl_mkeys < UCP_MAX_MDS, "num_tl_mkeys=%u"
+                    ", max=%u", unpacked->num_tl_mkeys, UCP_MAX_MDS);
+
+        tl_mkey               = &unpacked->tl_mkeys[unpacked->num_tl_mkeys];
+        tl_mkey->local_md_map = local_md_map;
+        tl_mkey->tl_mkey_buf  = tl_mkey_buf;
+        ++unpacked->num_tl_mkeys;
```
yosefe, line 566:
```suggestion
        ucs_assertv(unpacked->num_tl_mkeys < UCP_MAX_MDS, "num_tl_mkeys=%u"
                    " UCP_MAX_MDS=%u", unpacked->num_tl_mkeys, UCP_MAX_MDS);
        tl_mkey = &unpacked->tl_mkeys[unpacked->num_tl_mkeys++];
        ucp_memh_exported_tl_mkey_data_unpack(context, &p, &tl_mkey->tl_mkey_buf,
                                              &tl_mkey->local_md_map);
```

src/ucp/core/ucp_mm.c
iyastreb: sure

src/ucp/core/ucp_rkey.c
iyastreb: ok

iyastreb (commented):
Forced push, since changeset is small

yosefe (approved):

brminich (approved):

PR #10235, opened 2024-10-16
BUILD/CONFIG: Fix PGI compiler error: routine both inline and noinline

tvegas1:
## What
Fix PGI compiler errors found during `build_pgi` step like:
```
2024-10-10T08:54:25.0639396Z "/__w/1/s/contrib/../test/gtest/ucs/test_algorithm.cc", line 51: error #1626: routine is both "inline" and "noinline"
2024-10-10T08:54:25.0640811Z   UCS_TEST_F(test_algorithm, crc16) {
```

### DIFF
--- a/config/m4/compiler.m4
+++ b/config/m4/compiler.m4
@@ -523,6 +523,7 @@ AC_LANG_POP
 # --diag_suppress 188  - Suppress enumerated type mixed with another type
 # --diag_suppress 381  - Suppress extra ";" ignored
 # --diag_suppress 1215 - Suppress deprecated API warning for PGI18 compiler
+# --diag_suppress 1626 - Suppress routine is both "inline" and "noinline"
 # --diag_suppress 1901 - Use of a const variable in a constant expression is nonstandard in C
 # --diag_suppress 1902 - Use of a const variable in a constant expression is nonstandard in C (same as 1901)
 ADD_COMPILER_FLAGS_IF_SUPPORTED([[--display_error_number],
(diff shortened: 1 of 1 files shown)


tvegas1:
Confirmed that the failure at `==== Build with PGI compiler ====
` step is not seen anymore

yosefe (approved):

yosefe (changes requested):
maybe we need to fix the UCS_TEST_F macro that defines both inline and noninline routine?

config/m4/compiler.m4
```diff
@@ -534,6 +534,7 @@ ADD_COMPILER_FLAGS_IF_SUPPORTED([[--display_error_number],
                                  [--diag_suppress 188],
                                  [--diag_suppress 381],
                                  [--diag_suppress 1215],
+                                 [--diag_suppress 1626],
```
brminich, line 537: shall we also update the comment above?

tvegas1:
> maybe we need to fix the UCS_TEST_F macro that defines both inline and noninline routine?

I think the problem is with the compiler so I would not touch the code (which is gtest lib afaiu). Apparently it assumes that defining function as `static` inside class implies `inline` keyword, so when adding `__attribute__((noinline))` it declares a conflict.

```c++
class fails {
    static int __attribute__((noinline)) test() { return 1; }
};

class works {
    static int __attribute__((noinline)) test();
};
int __attribute__((noinline)) works::test() { return 1; }
```


For reference, from `UCS_TEST_F(gtest_common, auto_ptr)`, we get:
```c++
class gtest_common_auto_ptr_Test : public gtest_common {
...
    static ::testing::TestInfo *__attribute__((noinline)) create_test_info()
    {
        return ::testing::internal::MakeAndRegisterTestInfo(

```

config/m4/compiler.m4
tvegas1: force pushed

yosefe (approved):

PR #10236, opened 2024-10-16
UCT/GGA: rkey resolution for many MDs

evgeny-leksikov:
## What
rkey resolution for many MDs in GGA transport

## Why ?
https://github.com/openucx/ucx/pull/9993#discussion_r1791845290

## How ?
1. rkey has cached part for quick resolution on fast path
2. rkey has common part independent on MD
3. on usage on new MD, resolution mechanism allocates MD part and adds it to md->hash
4. also there is allocation of a helper structure which links MD to rkey and used in md_close to cleanup rkeys


src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -192,57 +279,122 @@ static uct_component_t uct_gga_component = {
...
+                               uct_gga_mlx5_rkey_handle_t *rkey_handle)
 {
-    uct_md_h uct_md                         = &md->super.super;
-    uct_md_mem_attach_params_t atach_params = { 0 };
-    uct_md_mkey_pack_params_t repack_params = { 0 };
+    uct_md_h uct_md                       = &md->super.super.super;
+    uct_gga_mlx5_rkey_hash_key_t hash_key = (uintptr_t)rkey_handle;
+    uct_gga_mlx5_md_rkey_handle_t* md_rkey_handle;
+    uct_gga_mlx5_rkey_handle_md_link_t *md_link;
+    uct_md_mem_attach_params_t atach_params;
```
iyastreb, line 308: typo, attach_params

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -72,20 +108,7 @@ ucs_config_field_t uct_gga_mlx5_iface_config_table[] = {
...
-    status = uct_ib_mlx5_devx_md_query(uct_md, md_attr);
-    if (status != UCS_OK) {
-        return status;
-    }
-
-    /* GGA always packs exported mkey */
-    md_attr->rkey_packed_size = md_attr->exported_mkey_packed_size;
-    return UCS_OK;
-}
+static pthread_mutex_t uct_gga_mlx5_rkeys_locker = PTHREAD_MUTEX_INITIALIZER;
```
iyastreb, line 111: maybe use "_lock" or "_mutex" to conform with the codebase
Maybe store it in the component uct_gga_component?

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -118,41 +141,105 @@ uct_gga_mlx5_rkey_unpack(uct_component_t *component, const void *rkey_buffer,
...
+    rkey_handle->packed_mkey = *mkey;
 
     *rkey_p   = (uintptr_t)rkey_handle;
     *handle_p = NULL;
     return UCS_OK;
 }
 
+static ucs_status_t
+uct_gga_mlx5_md_put_rkey(uct_gga_mlx5_md_t *md,
+                         const uct_gga_mlx5_rkey_hash_key_t key,
```
iyastreb, line 156: minor: since key is primitive `const` is redundant

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -118,41 +141,105 @@ uct_gga_mlx5_rkey_unpack(uct_component_t *component, const void *rkey_buffer,
...
+                         uct_gga_mlx5_md_rkey_handle_t *rkey_handle)
+{
+    khint_t iter;
+    ucs_kh_put_t ret;
+
+    iter = kh_put(resolved_rkeys, &md->rkey_cache, key, &ret);
+    if (ret == UCS_KH_PUT_FAILED) {
+        ucs_error("md %p: failed to add rkey to hash", md);
```
iyastreb, line 164: here and in ucs_assert I propose to print key value

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -192,57 +279,122 @@ static uct_component_t uct_gga_component = {
...
+{
+    uct_gga_mlx5_rkey_hash_key_t hash_key = (uintptr_t)rkey_handle;
+    const uct_gga_mlx5_md_rkey_handle_t* md_rkey_handle;
+
+    if (ucs_likely(rkey_handle->md == md)) {
+        ucs_assert(rkey_handle->rkey != UCT_INVALID_RKEY);
+        uct_gga_mlx5_rkey_trace(md, rkey_handle, "cached");
+        return UCS_OK;
+    } else if (rkey_handle->rkey != UCT_INVALID_RKEY) {
+        md_rkey_handle = uct_gga_mlx5_md_get_rkey(md, hash_key);
```
iyastreb, line 387: shouldn't this access be protected by mutex?

test/gtest/uct/test_p2p_rma.cc
```diff
@@ -165,3 +165,51 @@ UCS_TEST_SKIP_COND_P(test_p2p_rma_madvise, madvise,
...
+
+        e0->connect(0, *e2, 0);
+        e1->connect(0, *e2, 1);
+        flush();
+    }
+};
+
+UCS_TEST_SKIP_COND_P(uct_rma_common_receiver, reuse_rkey,
+                     !check_caps(UCT_IFACE_FLAG_PUT_ZCOPY|
+                                 UCT_IFACE_FLAG_GET_ZCOPY))
```
evgeny-leksikov, line 193: redundant requirement for GET_ZCOPY

src/uct/ib/mlx5/gga/gga_mlx5.c
evgeny-leksikov: I also thought about it, but don't found much benefit, just more code related to new type/struct

src/uct/ib/mlx5/gga/gga_mlx5.c
evgeny-leksikov: it's more for readability and clean API + `uct_gga_mlx5_rkey_hash_key_t ` might be changed in future, so this place won't require changes

iyastreb (dismissed):

evgeny-leksikov:
@gleon99 @yosefe could you please review

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -14,14 +14,50 @@
...
+} uct_gga_mlx5_md_t;
+
+/**
+ * Entry of uct_gga_mlx5_rkey_handle_t::mds to implement many-to-many
+ * relationship between rkeys and mds
+ */
+typedef struct {
+    ucs_list_link_t   list_link;
+    uct_gga_mlx5_md_t *md;
+} uct_gga_mlx5_rkey_handle_md_link_t;
```
yosefe, line 52: as discussed offline, change the structs to be global hash table and not per md, and review locking scheme when accessing from multiple threads (md)

test/gtest/uct/test_p2p_rma.cc
```diff
@@ -165,3 +165,50 @@ UCS_TEST_SKIP_COND_P(test_p2p_rma_madvise, madvise,
...
 UCT_INSTANTIATE_TEST_CASE(test_p2p_rma_madvise)
+
+class uct_rma_common_receiver : public uct_test {
+protected:
+    virtual void init() override {
+        uct_test::init();
+        EXPECT_EQ(0, m_entities.size());
+
+        const size_t rx_headroom = 0;
+        entity* e0 = create_entity(rx_headroom);
```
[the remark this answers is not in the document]
evgeny-leksikov: the default `create_entity()` requires CM support
`[     SKIP ] (cm is not supported on component gga)`

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -28,13 +33,58 @@ typedef struct {
...
+{
+    return (a.rkey_handle == b.rkey_handle) && (a.md == b.md);
+}
+
+KHASH_TYPE(resolved_rkeys,
+           uct_gga_mlx5_rkey_hash_key_t, uct_gga_mlx5_rkey_bundle_t*);
+
+KHASH_IMPL(resolved_rkeys, uct_gga_mlx5_rkey_hash_key_t,
+           uct_gga_mlx5_rkey_bundle_t*, 1,
+           uct_gga_mlx5_rkey_hash_func, uct_gga_mlx5_rkey_hash_key_equal)
```
rakhmets, line 79:
```suggestion
KHASH_INIT(resolved_rkeys, uct_gga_mlx5_rkey_hash_key_t,
           uct_gga_mlx5_rkey_bundle_t*, 1, uct_gga_mlx5_rkey_hash_func,
           uct_gga_mlx5_rkey_hash_key_equal)
```

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -28,13 +33,58 @@ typedef struct {
...
+
+static khash_t(resolved_rkeys) uct_gga_mlx5_rkey_cache;
+
+KHASH_TYPE(garbage_rkeys, uct_gga_mlx5_rkey_hash_key_t, char);
+
+KHASH_IMPL(garbage_rkeys, uct_gga_mlx5_rkey_hash_key_t, char, 0,
+           uct_gga_mlx5_rkey_hash_func, uct_gga_mlx5_rkey_hash_key_equal)
```
rakhmets, line 86:
```suggestion
KHASH_INIT(garbage_rkeys, uct_gga_mlx5_rkey_hash_key_t, char, 0,
           uct_gga_mlx5_rkey_hash_func, uct_gga_mlx5_rkey_hash_key_equal)
```

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -28,13 +33,58 @@ typedef struct {
...
+typedef struct {
+    uct_gga_mlx5_rkey_handle_t *rkey_handle;
+    uct_gga_mlx5_md_t          *md;
+} uct_gga_mlx5_rkey_hash_key_t;
+
+static UCS_F_ALWAYS_INLINE khint32_t
+uct_gga_mlx5_rkey_hash_func(uct_gga_mlx5_rkey_hash_key_t key)
+{
+    khint64_t k1 = (uintptr_t)key.rkey_handle;
+    khint64_t k2 = (uintptr_t)key.md;
```
rakhmets, line 62:
```suggestion
    khint64_t k1 = (khint64_t)key.rkey_handle;
    khint64_t k2 = (khint64_t)key.md;
```

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -268,57 +361,133 @@ static uct_component_t uct_gga_component = {
...
+    uct_gga_mlx5_rkey_bundle_t *rkey_bundle;
+    uct_md_mem_attach_params_t attach_params;
+    uct_md_mkey_pack_params_t repack_params;
     uint64_t repack_mkey;
     ucs_status_t status;
 
-    if (ucs_likely(rkey_handle->memh != NULL)) {
-        uct_gga_mlx5_rkey_trace(md, rkey_handle, "reuse");
-        return UCS_OK;
+    rkey_bundle = ucs_malloc(sizeof(*rkey_bundle), "gga_rkey_bundle");
```
rakhmets, line 398: You can avoid malloc/free if you use uct_gga_mlx5_rkey_bundle_t as hash map value instead of uct_gga_mlx5_rkey_bundle_t*.
khash implementation already has malloc/free for hash map values. So after kh_put, you can initialize the value.

src/uct/ib/mlx5/gga/gga_mlx5.c
```diff
@@ -854,16 +1034,60 @@ UCT_TL_DEFINE_ENTRY(&uct_gga_component, gga_mlx5, uct_gga_mlx5_query_tl_devices,
...
+
+    ucs_trace("md %p: closing on GGA device %s", md,
+              uct_ib_device_name(&gga_md->super.super.dev));
+
+    kh_init_inplace(garbage_rkeys, &grkeys);
+    pthread_mutex_lock(&uct_gga_mlx5_rkeys_lock);
+    kh_foreach(&uct_gga_mlx5_rkey_cache, hash_key, hash_val, {
+        if (gga_md == hash_key.md) {
+            uct_gga_mlx5_md_rkey_handle_release(hash_key.md, &hash_val);
+            kh_put(garbage_rkeys, &grkeys, hash_key, &put_ret);
```
rakhmets, line 1058: You iterate through keys of the hash map, and put keys to hash set with the same hash function. It means that you don't need hash set here. It looks like an array of hash map iterators is enough.

evgeny-leksikov: It's needed to safely cleanup the hash map by keys. Are you sure that one iterator of khash remains valid after deleting other one?

rakhmets: I see. Then it should an array of keys.

evgeny-leksikov: I used `khash` in set mode to simpify logic with (re-)allocation, we can use `ucs_array` but I dont see much benefit on cleanup path.

rakhmets: The point is to use the simplest data structure that satisfies the conditions of the problem. You do not need the features provided by the set. You need only an array. You can use dynamic array for the purpose. You also can get rid of the extra code hidden by the macro in the beginning of the file.

evgeny-leksikov: dynamic array with grow factor  2 may consume more memory for example, my point is both containers have pros and cons

rakhmets: khash set also does reallocation in frames of rehashing when the size changes. dynamic array is more lightweight акщь this perspective. I don't think it's an advantage from this point of view.

evgeny-leksikov: > khash also does reallocation in frames of rehashing when the size changes

yes but by bucket size vs miltiplying by 2 for array, so it's hard to say what's defenitely better

evgeny-leksikov: ucs_list requires additional type/field for an entry

rakhmets: You added a structure with redundant fields along with 10 extra methods to introduce khash set. This is an extra stuff which is not required for your purpose.

evgeny-leksikov: Adding list entry to the key structure means we need to store extra 8 bytes per key, adding new inherited structure means extra allocation on cleanup every key to add the key to the list.

rakhmets: If we characterize them in terms of complexity they are the same.
Agree that ucs_list (double linked list) is not needed here since you do not need to insert in the middle.
So ucs_array is preferrable solution then, a typical choice for a simple variable-length collection.

>> khash also does reallocation in frames of rehashing when the size changes

> yes but by bucket size vs miltiplying by 2 for array, so it's hard to say what's defenitely better

khash does the same. So `ucs_array` is better.

evgeny-leksikov: In current state of the PR this discussion is meaningless since we need to decide if we would like to have global hash table or separate it per MD, in second case this code is not relevant at all. So, for now I propose keep the khash just because it's already implemented and work as expected. In case of decision that we go with "global hash table", I agree that ucs_array is preferable as a local storage.

gleon99:
Hi @evgeny-leksikov - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #10239, opened 2024-10-17
CONTRIB: Fixed typo in env var for cuda_hooks gtest.

rakhmets:
## What
Fixed typo in env var for cuda_hooks gtest.

### DIFF
--- a/contrib/test_jenkins.sh
+++ b/contrib/test_jenkins.sh
@@ -878,7 +878,7 @@ run_malloc_hook_gtest() {
 
 	echo "==== Running cuda hooks with far jump, $compiler_name compiler ===="
 	$TIMEOUT env \
-		UCM_BISTRO_FORCE_FAR_JUMP=y \
+		UCX_MEM_BISTRO_FORCE_FAR_JUMP=y \
 		GTEST_FILTER='cuda_hooks.*' \
 			make -C test/gtest test
 }


iyastreb (approved):

brminich (dismissed):

contrib/test_jenkins.sh
```diff
@@ -878,7 +878,7 @@ run_malloc_hook_gtest() {
 
 	echo "==== Running cuda hooks with far jump, $compiler_name compiler ===="
 	$TIMEOUT env \
-		UCM_BISTRO_FORCE_FAR_JUMP=y \
+		UCX_BISTRO_FORCE_FAR_JUMP=y \
```
yosefe, line 881: i think it should be `UCX_MEM_BISTRO_FORCE_FAR_JUMP=y`
```
$ ./build-devel/src/tools/info/ucx_info -c|grep BIST
UCX_MEM_BISTRO_FORCE_FAR_JUMP=n
```

yosefe (changes requested):

contrib/test_jenkins.sh
rakhmets: Yes, you're right.
Force-pushed the change since it is one line commit.

brminich (approved):

PR #10240, opened 2024-10-18
UCM: Fixed race condition in parsing proc maps

iyastreb:
## What
Segfault in MT applications using CUDA memory allocated before UCX is initialised.

## Why ?
Segfault happens because UCX treats passed memory buffer as HOST memory instead of CUDA
Because `memtype_cache` does not contain all maps allocated before UCX init
Because `ucs_memtype_cache_get_global` creates multiple instances of `memtype_cache`, but concurrent init is buggy
```
ucp_datatype_iter_detect_mem_info
  ucp_memory_detect
    ucp_memory_detect_internal
      ucs_memtype_cache_lookup
        ucs_memtype_cache_get_global
          **create ucs_memtype_cache_t per each concurrent thread**
            ucm_set_event_handler
              get_existing_alloc
                ucm_cudamem_get_existing_alloc
                  ucm_parse_proc_self_maps  << parallel reads, but read is **destructive**
```

## How ?
- Fixed destructive read in ucm_parse_proc_self_maps
- Added unit test that reproduces the issue
- Micro optimisations for constant string comparison

### DIFF
--- a/src/ucm/cuda/cudamem.c
+++ b/src/ucm/cuda/cudamem.c
@@ -368,15 +368,15 @@ out:
 static int ucm_cudamem_scan_regions_cb(void *arg, void *addr, size_t length,
                                        int prot, const char *path)
 {
-    static const char *cuda_path_pattern = "/dev/nvidia";
-    ucm_event_handler_t *handler         = arg;
+    static const char cuda_path_pattern[] = "/dev/nvidia";
+    ucm_event_handler_t *handler          = arg;
     ucm_event_t event;
 
     /* we are interested in blocks which don't have any access permissions, or
      * mapped to nvidia device.
      */
     if ((prot & (PROT_READ | PROT_WRITE | PROT_EXEC)) &&
-        strncmp(path, cuda_path_pattern, strlen(cuda_path_pattern))) {
+        strncmp(path, cuda_path_pattern, sizeof(cuda_path_pattern) - 1)) {
         return 0;
     }
 
--- a/src/ucm/rocm/rocmmem.c
+++ b/src/ucm/rocm/rocmmem.c
@@ -172,12 +172,12 @@ out:
 static int ucm_rocm_scan_regions_cb(void *arg, void *addr, size_t length,
                                     int prot, const char *path)
 {
-    static const char *rocm_path_pattern = "/dev/dri";
-    ucm_event_handler_t *handler = arg;
+    static const char rocm_path_pattern[] = "/dev/dri";
+    ucm_event_handler_t *handler          = arg;
     ucm_event_t event;
 
     if ((prot & (PROT_READ | PROT_WRITE | PROT_EXEC)) &&
-        strncmp(path, rocm_path_pattern, strlen(rocm_path_pattern))) {
+        strncmp(path, rocm_path_pattern, sizeof(rocm_path_pattern) - 1)) {
         return 0;
     }
     ucm_debug("dispatching initial memtype allocation for %p..%p %s", addr,
--- a/src/ucm/util/sys.c
+++ b/src/ucm/util/sys.c
@@ -150,7 +150,7 @@ void ucm_parse_proc_self_maps(ucm_proc_maps_cb_t cb, void *arg)
 {
     static char  *buffer         = MAP_FAILED;
     static size_t buffer_size    = 32768;
-    static pthread_rwlock_t lock = PTHREAD_RWLOCK_INITIALIZER;
+    static pthread_mutex_t lock  = PTHREAD_MUTEX_INITIALIZER;
     ssize_t read_size, offset;
     unsigned long start, end;
     char prot_c[4];
@@ -168,7 +168,7 @@ void ucm_parse_proc_self_maps(ucm_proc_maps_cb_t cb, void *arg)
     }
 
     /* read /proc/self/maps fully into the buffer */
-    pthread_rwlock_wrlock(&lock);
+    pthread_mutex_lock(&lock);
 
     if (buffer == MAP_FAILED) {
         buffer = ucm_orig_mmap(NULL, buffer_size, PROT_READ|PROT_WRITE,
@@ -223,11 +223,8 @@ void ucm_parse_proc_self_maps(ucm_proc_maps_cb_t cb, void *arg)
             offset += read_size;
         }
     }
-    pthread_rwlock_unlock(&lock);
 
     close(maps_fd);
-
-    pthread_rwlock_rdlock(&lock);
 
     ptr      = buffer;
     line_num = 1;
@@ -265,7 +262,7 @@ void ucm_parse_proc_self_maps(ucm_proc_maps_cb_t cb, void *arg)
     }
 
 out:
-    pthread_rwlock_unlock(&lock);
+    pthread_mutex_unlock(&lock);
 }
 
 typedef struct {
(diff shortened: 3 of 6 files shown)


src/ucm/cuda/cudamem.c
```diff
@@ -368,15 +368,15 @@ static ucs_status_t ucm_cudamem_install(int events)
...
+    static const char cuda_path_pattern[] = "/dev/nvidia";
+    ucm_event_handler_t *handler          = arg;
     ucm_event_t event;
 
     /* we are interested in blocks which don't have any access permissions, or
      * mapped to nvidia device.
      */
     if ((prot & (PROT_READ | PROT_WRITE | PROT_EXEC)) &&
-        strncmp(path, cuda_path_pattern, strlen(cuda_path_pattern))) {
+        strncmp(path, cuda_path_pattern, sizeof(cuda_path_pattern) - 1)) {
```
tvegas1, line 379: I think this could be an issue since sizeof now incudes '\0' and we don't do prefix comparison anymore?

src/ucm/rocm/rocmmem.c
```diff
@@ -172,12 +172,12 @@ static ucs_status_t ucm_rocmmem_install(int events)
...
 {
-    static const char *rocm_path_pattern = "/dev/dri";
-    ucm_event_handler_t *handler = arg;
+    static const char rocm_path_pattern[] = "/dev/dri";
+    ucm_event_handler_t *handler          = arg;
     ucm_event_t event;
 
     if ((prot & (PROT_READ | PROT_WRITE | PROT_EXEC)) &&
-        strncmp(path, rocm_path_pattern, strlen(rocm_path_pattern))) {
+        strncmp(path, rocm_path_pattern, sizeof(rocm_path_pattern) - 1)) {
```
tvegas1, line 180: same as above

test/gtest/ucs/test_memtype_cache.cc
```diff
@@ -439,6 +439,12 @@ class test_memtype_cache_deferred_create : public test_memtype_cache {
         /* do nothing */
     }
 
+    virtual void cleanup() {
+        /* Clean the global memtype cache */
+        ucs_memtype_cache_cleanup();
+        ucs_memtype_cache_global_init();
```
tvegas1, line 445: for my understanding, why ..global_init()?

src/ucm/util/sys.c
```diff
@@ -235,14 +235,14 @@ void ucm_parse_proc_self_maps(ucm_proc_maps_cb_t cb, void *arg)
         /* address           perms offset   dev   inode   pathname
          * 00400000-0040b000 r-xp  00001a00 0a:0b 12345   /dev/mydev
          */
-        *newline = '\0';
+        length = newline - ptr;
         ret = sscanf(ptr, "%lx-%lx %4c %*x %*x:%*x %*d %n",
```
tvegas1, line 239: shall we also add \n at the end of sscanf line, and verify that ret is exactly the match?

test/gtest/ucs/test_memtype_cache.cc
```diff
@@ -439,6 +439,12 @@ class test_memtype_cache_deferred_create : public test_memtype_cache {
         /* do nothing */
     }
 
+    virtual void cleanup() {
+        /* Clean the global memtype cache */
+        ucs_memtype_cache_cleanup();
```
brminich, line 444: why need to cleanup?

src/ucm/cuda/cudamem.c
```diff
@@ -368,15 +368,15 @@ static ucs_status_t ucm_cudamem_install(int events)
 static int ucm_cudamem_scan_regions_cb(void *arg, void *addr, size_t length,
                                        int prot, const char *path)
 {
-    static const char *cuda_path_pattern = "/dev/nvidia";
-    ucm_event_handler_t *handler         = arg;
+    static const char cuda_path_pattern[] = "/dev/nvidia";
+    ucm_event_handler_t *handler          = arg;
```
brminich, line 372: why is this change needed?

brminich (commented):
can we just use single lock for the whole `ucm_parse_proc_self_maps`?

src/ucm/cuda/cudamem.c
iyastreb: This is a micro-optimisation that I've made while investigating the usage of `ucm_parse_proc_self_maps` callbacks. Basically instead of calculating the length of the string we just use compile time constant

src/ucm/cuda/cudamem.c
tvegas1: missed -1

iyastreb:
> can we just use single lock for the whole `ucm_parse_proc_self_maps`?

Yes, we discussed that, and I reflected a bit more on this.
I tend to agree that having plain mutex lock seems to be a right choice here.

Currently there are 3 use-cases of `ucm_parse_proc_self_maps`:
1. ucs_memtype_cache_get_global initialisation. There is no much sense in making it more scalable, because callbacks ends up calling `ucs_memtype_cache_update_internal` which locks internally. Also it happens only at startup, so not a big deal
2. ucm_shmdt:
```C
    if (!ucm_shm_del_entry_from_khash(shmaddr, &size)) {        // unlikely?
        size = ucm_get_shm_seg_size(shmaddr); -> ucm_parse_proc_self_maps
    }
```
3. ucs_rcache_check_overlap_one:
ucs_get_mem_prot -> ucm_parse_proc_self_maps

So I guess we have RW lock to decouple 2 and 3 workflows, which should not very frequent events.
Therefore I don't think we have a good use case for RW lock here. Locking a read-write lock is more complicated, and because it involves operating system calls, they are slightly slower than mutexes.

src/ucm/util/sys.c
iyastreb: with extended mutex scope we ain't gonna need it anyway

test/gtest/ucs/test_memtype_cache.cc
brminich: ok

brminich (approved):

yosefe:
@iyastreb 
1. Multiple instances of memtype cache: AFAIR we had a real issue of possible lock inversion deadlock. We can try to revert and see the warning. But multiple instances would be created only in case of a race, which is not likely, so IMO it's not a big issue from performance perspective.
2. `ucp_datatype_iter_detect_mem_info` still needs to detect sys_dev in case of non-host memory since we don't have UCP API to pass sys_dev
3. RW lock: i guess the destructive part is `*newline = 0` - we can avoid it probably by using a more accurate pattern of sscanf or copying each line to a local buffer on the stack

brminich:
pls ignore part 3 and use simple lock (per offline discussion)

yosefe (approved):

iyastreb:
> 1. Multiple instances of memtype cache: AFAIR we had a real issue of possible lock inversion deadlock. We can try to revert and see the warning. But multiple instances would be created only in case of a race, which is not likely, so IMO it's not a big issue from performance perspective.

From what I see - the opposite is true, it is very likely that in MT app multiple instances are created.
This happens because all threads call `ucs_memtype_cache_get_global` at startup and this indirectly calls `ucm_parse_proc_self_maps` which is a very expensive and time-consuming call. When self maps are loaded/parsed by the first thread, it is very likely that other threads enter this function as well. When I run 8 threads I typically see 8 such variables being created and then destroyed.

PR #10242, opened 2024-10-20
AZP: Improve reporting of CUDA device status during tests

Alexey-Rivkin:
## What
Trigger warning if CUDA device is unavailable on non-GPU nodes.
Trigger error if CUDA device is unavailable on GPU nodes.

## Why ?
If CUDA device was not found, UCX is configured with no CUDA:
https://github.com/openucx/ucx/blob/master/buildlib/jucx/jucx-test.yml#L39
It's expected in some cases, as we currently do not have ARM machines with GPU.

### DIFF
--- a/buildlib/az-helpers.sh
+++ b/buildlib/az-helpers.sh
@@ -207,7 +207,13 @@ try_load_cuda_env() {
 
 load_cuda_env() {
     try_load_cuda_env
-    [ "${have_cuda}" == "yes" ] || azure_log_error "Cuda device is not available"
+    if [ "${have_cuda}" != "yes" ] ; then
+        if [ "${ucx_gpu}" = "yes" ] ; then
+            azure_log_error "CUDA load failed on GPU node $(hostname -s)"
+            exit 1
+        fi
+        azure_log_warning "Cuda device is not available"
+    fi
 }
 
 check_release_build() {


michal-shalev (dismissed):
Thanks, looks good :)

yosefe:
Do we have a way to ensure that we do have GPU devices on machines that should have it, so we don't miss testing coverage?

Alexey-Rivkin:
> Do we have a way to ensure that we do have GPU devices on machines that should have it, so we don't miss testing coverage?

Unfortunately, we can't install any GPU on our ARM nodes. 
Support ticket (internal link): [RITM01330546](https://nvidia.service-now.com/esc?id=ticket&table=sc_req_item&sys_id=4cfe2fe2839592100ca7fd326daad350)

yosefe:
> > Do we have a way to ensure that we do have GPU devices on machines that should have it, so we don't miss testing coverage?
> 
> Unfortunately, we can't install any GPU on our ARM nodes. Support ticket (internal link): [RITM01330546](https://nvidia.service-now.com/esc?id=ticket&table=sc_req_item&sys_id=4cfe2fe2839592100ca7fd326daad350)

Maybe check if the agent has GPU tag?

yosefe:
> This will effectively cancel JUCX testing on ARM: https://github.com/openucx/ucx/blob/master/buildlib/pr/main.yml#L366-L369

Why?
We can check if agent has GPU label, and if yes and cuda module load fails - that's considered a test failure

Alexey-Rivkin:
> > This will effectively cancel JUCX testing on ARM: https://github.com/openucx/ucx/blob/master/buildlib/pr/main.yml#L366-L369
> 
> Why? We can check if agent has GPU label, and if yes and cuda module load fails - that's considered a test failure

Done.
Labels are readily available as env vars at runtime, so the solution is simple.

dpressle (approved):

dpressle:
@yosefe do you have more comments here or we can merge it?

buildlib/az-helpers.sh
```diff
@@ -207,7 +207,11 @@ try_load_cuda_env() {
 
 load_cuda_env() {
     try_load_cuda_env
-    [ "${have_cuda}" == "yes" ] || azure_log_error "Cuda device is not available"
+    if [ "${ucx_gpu}" = "yes" ] && [ "${have_cuda}" = "no" ]; then
+        azure_log_error "CUDA load failed on GPU node $(hostname -s)"
+        exit 1
+    fi
+    [ "${have_cuda}" = "yes" ] || azure_log_warning  "Cuda device is not available"
```
yosefe, line 214:
```suggestion
    if [ "${have_cuda}" != "yes" ] ; then
        if [ "${ucx_gpu}" = "yes" ] ; then
            azure_log_error "CUDA load failed on GPU node $(hostname -s)"
            exit 1
        fi
        azure_log_warning "Cuda device is not available"
    fi
```

yosefe (approved):

yosefe:
@Alexey-Rivkin pls squash

PR #10243, opened 2024-10-21
UCT/IB/MLX5/DC: hybrid dcs fixes - pending queue

roiedanino:
## What 
This PR fixes tx / ep scheduling for DC dcs_hybrid policy, rescheduling when the hw dci is being used by the ep.

## Why ?
Recently added `dcs_hybrid` policy failed during pending queue test in gtest.

### DIFF
--- a/src/uct/ib/mlx5/dc/dc_mlx5.c
+++ b/src/uct/ib/mlx5/dc/dc_mlx5.c
@@ -252,9 +252,10 @@ static ucs_status_t uct_dc_mlx5_iface_query(uct_iface_h tl_iface, uct_iface_attr
                                    UCT_IFACE_FLAG_PUT_ZCOPY);
     }
 
-    /* Error handling is not supported with random dci policy
+    /* Error handling is not supported with shared dci policies
      * TODO: Fix */
-    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_policy_shared(iface) ||
+        uct_dc_mlx5_iface_is_hybrid(iface)) {
         iface_attr->cap.flags &= ~(UCT_IFACE_FLAG_ERRHANDLE_PEER_FAILURE |
                                    UCT_IFACE_FLAG_ERRHANDLE_ZCOPY_BUF    |
                                    UCT_IFACE_FLAG_ERRHANDLE_REMOTE_MEM);
@@ -1179,12 +1180,12 @@ static void uct_dc_mlx5_iface_cleanup_fc_ep(uct_dc_mlx5_iface_t *iface)
     ucs_arbiter_group_cleanup(&fc_ep->arb_group);
     uct_rc_fc_cleanup(&fc_ep->fc);
 
-    if ((fc_ep->dci != UCT_DC_MLX5_EP_NO_DCI) &&
+    if ((fc_ep->dci == UCT_DC_MLX5_EP_NO_DCI) ||
         !uct_dc_mlx5_is_dci_valid(fc_dci)) {
         goto out;
     }
 
-    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
+    if (uct_dc_mlx5_is_dci_shared(iface, fc_ep->dci)) {
         txqp = &fc_dci->txqp;
         ucs_queue_for_each_safe(op, iter, &txqp->outstanding, queue) {
             if (op->handler == uct_dc_mlx5_ep_fc_pure_grant_send_completion) {
--- a/src/uct/ib/mlx5/dc/dc_mlx5.inl
+++ b/src/uct/ib/mlx5/dc/dc_mlx5.inl
@@ -41,17 +41,31 @@ uct_dc_mlx5_get_arbiter_params(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep,
 }
 
 static UCS_F_ALWAYS_INLINE void
-uct_dc_mlx5_ep_schedule(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
+uct_dc_mlx5_ep_push_pending_req(ucs_arbiter_group_t *group,
+                                uct_pending_req_t *r, int push_to_head)
 {
-    if (ep->dci == UCT_DC_MLX5_EP_NO_DCI) {
-        /* no dci:
-         * Do not grab dci here. Instead put the group on dci allocation
-         * arbiter. This way we can assure fairness between all eps waiting for
-         * dci allocation. Relevant for dcs and dcs_quota policies.
-         */
-        uct_dc_mlx5_iface_schedule_dci_alloc(iface, ep);
+    if (push_to_head) {
+        uct_pending_req_arb_group_push_head(group, r);
     } else {
-        uct_dc_mlx5_iface_dci_sched_tx(iface, ep);
+        uct_pending_req_arb_group_push(group, r);
+    }
+}
+
+static UCS_F_ALWAYS_INLINE void
+uct_dc_mlx5_ep_pending_common_shared(uct_dc_mlx5_iface_t *iface,
+                                     uct_dc_mlx5_ep_t *ep, uct_pending_req_t *r,
+                                     unsigned flags, int push_to_head,
+                                     int schedule)
+{
+    ucs_arbiter_group_t *group = uct_dc_mlx5_ep_shared_arb_group(iface, ep);
+
+    UCS_STATIC_ASSERT(sizeof(uct_dc_mlx5_pending_req_priv) <=
+                      UCT_PENDING_REQ_PRIV_LEN);
+    uct_dc_mlx5_pending_req_priv(r)->ep = ep;
+    uct_dc_mlx5_ep_push_pending_req(group, r, push_to_head);
+
+    if (schedule) {
+        ucs_arbiter_group_schedule(uct_dc_mlx5_iface_tx_waitq(iface), group);
     }
 }
 
@@ -62,26 +76,30 @@ uct_dc_mlx5_ep_pending_common(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep,
 {
     ucs_arbiter_group_t *group;
 
-    UCS_STATIC_ASSERT(sizeof(uct_dc_mlx5_pending_req_priv) <=
-                      UCT_PENDING_REQ_PRIV_LEN);
+    UCT_TL_EP_STAT_PEND(&ep->super);
 
     if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
-        uct_dc_mlx5_pending_req_priv(r)->ep = ep;
-        group = uct_dc_mlx5_ep_rand_arb_group(iface, ep);
-    } else {
-        group = &ep->arb_group;
+        uct_dc_mlx5_ep_pending_common_shared(iface, ep, r, flags, push_to_head,
+                                             schedule);
+        return;
     }
 
-    if (push_to_head) {
-        uct_pending_req_arb_group_push_head(group, r);
-    } else {
-        uct_pending_req_arb_group_push(group, r);
-    }
+    group = &ep->arb_group;
 
-    UCT_TL_EP_STAT_PEND(&ep->super);
+    uct_dc_mlx5_ep_push_pending_req(group, r, push_to_head);
+
     if (!schedule) {
         return;
     }
 
-    uct_dc_mlx5_ep_schedule(iface, ep);
+    if (ep->dci == UCT_DC_MLX5_EP_NO_DCI) {
+        /* no dci:
+         * Do not grab dci here. Instead put the group on dci allocation
+         * arbiter. This way we can assure fairness between all eps waiting for
+         * dci allocation. Relevant for dcs, dcs_quota and dcs_hybrid policies.
+         */
+        uct_dc_mlx5_iface_schedule_dci_alloc(iface, ep);
+    } else if (uct_dc_mlx5_iface_dci_has_tx_resources(iface, ep->dci)) {
+        ucs_arbiter_group_schedule(uct_dc_mlx5_iface_tx_waitq(iface), group);
+    }
 }
--- a/src/uct/ib/mlx5/dc/dc_mlx5_ep.c
+++ b/src/uct/ib/mlx5/dc/dc_mlx5_ep.c
@@ -1401,7 +1401,12 @@ uct_dc_mlx5_iface_dci_do_pending_wait(ucs_arbiter_t *arbiter,
     }
 
     ucs_assert_always(ep->dci != UCT_DC_MLX5_EP_NO_DCI);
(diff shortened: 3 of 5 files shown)


src/uct/ib/mlx5/dc/dc_mlx5_ep.c
```diff
@@ -1545,7 +1559,7 @@ uct_dc_mlx5_ep_arbiter_purge_cb(ucs_arbiter_t *arbiter, ucs_arbiter_group_t *gro
                                                     priv);
     uct_rc_pending_req_t *freq;
 
-    if (uct_dc_mlx5_iface_is_policy_shared(iface) &&
+    if ((uct_dc_mlx5_iface_is_policy_shared(iface)) &&
```
roiedanino, line 1562: remove ( )

src/uct/ib/mlx5/dc/dc_mlx5.inl
```diff
@@ -65,13 +65,12 @@ uct_dc_mlx5_ep_pending_common(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep,
     UCS_STATIC_ASSERT(sizeof(uct_dc_mlx5_pending_req_priv) <=
                       UCT_PENDING_REQ_PRIV_LEN);
 
-    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
+    if (ep->dci != UCT_DC_MLX5_EP_NO_DCI) {
```
roiedanino, line 68: revert?

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -251,7 +251,8 @@ static ucs_status_t uct_dc_mlx5_iface_query(uct_iface_h tl_iface, uct_iface_attr
 
     /* Error handling is not supported with random dci policy
      * TODO: Fix */
-    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_policy_shared(iface) ||
```
yosefe, line 254: update the comment in prev line

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -1180,7 +1181,8 @@ static void uct_dc_mlx5_iface_cleanup_fc_ep(uct_dc_mlx5_iface_t *iface)
         goto out;
     }
 
-    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_policy_shared(iface) ||
+        uct_dc_mlx5_is_hw_dci(iface, fc_ep->dci)) {
```
yosefe, line 1185: maybe use uct_dc_mlx5_is_dci_shared?

src/uct/ib/mlx5/dc/dc_mlx5.inl
yosefe, line 68: Why changed? seems the access to uct_dc_mlx5_pending_req_priv(req)->ep is still done only when uct_dc_mlx5_iface_is_policy_shared() is true, in uct_dc_mlx5_ep_arbiter_purge_cb.
Is there another usage?

src/uct/ib/mlx5/dc/dc_mlx5_ep.c
```diff
@@ -1460,7 +1460,8 @@ unsigned uct_dc_mlx5_ep_dci_release_progress(void *arg)
         dci_pool = &iface->tx.dci_pool[pool_index];
         while (dci_pool->release_stack_top >= 0) {
             dci = ucs_array_elem(&dci_pool->stack, dci_pool->release_stack_top--);
-            ucs_assert(dci < iface->tx.ndci * iface->tx.num_dci_pools);
+            ucs_assert((dci - uct_dc_mlx5_iface_is_hybrid(iface)) <
```
yosefe, line 1463: 1. ucs_assertv
2. add () around * expression

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -292,19 +304,18 @@ uct_dc_mlx5_ep_rand_arb_group(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 static UCS_F_ALWAYS_INLINE ucs_arbiter_group_t*
 uct_dc_mlx5_ep_arb_group(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 {
-    return (uct_dc_mlx5_iface_is_policy_shared(iface)) ?
-            uct_dc_mlx5_ep_rand_arb_group(iface, ep) : &ep->arb_group;
+    return uct_dc_mlx5_iface_is_policy_shared(iface) ?
+                   uct_dc_mlx5_ep_rand_arb_group(iface, ep) :
```
yosefe, line 308: maybe rename uct_dc_mlx5_ep_rand_arb_group to uct_dc_mlx5_ep_shared_arb_group ?
even though that function is not changed in this PR, seems like a better name for it

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -292,19 +304,18 @@ uct_dc_mlx5_ep_rand_arb_group(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
 
 static UCS_F_ALWAYS_INLINE void
 uct_dc_mlx5_iface_dci_sched_tx(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
 {
-    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
-        ucs_arbiter_group_schedule(uct_dc_mlx5_iface_tx_waitq(iface),
-                                   uct_dc_mlx5_ep_rand_arb_group(iface, ep));
-    } else if (uct_dc_mlx5_iface_dci_has_tx_resources(iface, ep->dci)) {
+    if (uct_dc_mlx5_iface_dci_has_tx_resources(iface, ep->dci) ||
+        uct_dc_mlx5_is_dci_shared(iface, ep->dci)) {
```
yosefe, line 316: Seems like both these functions access the dci array by ep->dci index to check something.
Maybe worth uniting the checks by getting pointer to dci object first?
BTW, similar check exists in uct_dc_mlx5_iface_dci_get

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -690,13 +691,20 @@ uct_dc_mlx5_iface_dci_detach(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
...
     ucs_assert(dci_index != UCT_DC_MLX5_EP_NO_DCI);
-    ucs_assert(iface->tx.dci_pool[uct_dc_mlx5_ep_pool_index(ep)].stack_top > 0);
 
-    if (uct_dc_mlx5_iface_dci_has_outstanding(iface, dci_index) ||
-        uct_dc_mlx5_is_hw_dci(iface, dci_index)) {
+
+    if (uct_dc_mlx5_is_hw_dci(iface, dci_index)) {
+        ep->dci    = UCT_DC_MLX5_EP_NO_DCI;
+        ep->flags &= ~UCT_DC_MLX5_EP_FLAG_TX_WAIT;
+        return 1;
```
yosefe, line 699: can we unite with lines 708-709 using goto?
we can move uct_dc_mlx5_iface_dci_schedule_release fron line 711 before lines 708-709 to achieve that

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -584,7 +583,8 @@ uct_dc_mlx5_iface_dci_put(uct_dc_mlx5_iface_t *iface, uint8_t dci_index)
 
     ucs_assert(dci_index != UCT_DC_MLX5_EP_NO_DCI);
 
-    if (uct_dc_mlx5_iface_is_policy_shared(iface)) {
+    if (uct_dc_mlx5_iface_is_policy_shared(iface) ||
+        uct_dc_mlx5_is_hw_dci(iface, dci_index)) {
```
yosefe, line 587: can we use is_dci_shared?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -165,10 +165,14 @@ UCS_TEST_P(test_dc, dcs_single) {
     ucs_status_t status;
     uct_dc_mlx5_ep_t *ep;
     uct_dc_mlx5_iface_t *iface;
+    uint16_t first_allocated_dci;
 
     m_e1->connect_to_iface(0, *m_e2);
     ep = dc_ep(m_e1, 0);
     iface = dc_iface(m_e1);
+    first_allocated_dci = uct_dc_mlx5_iface_is_hybrid(dc_iface(m_e1)) ? 1 : 0;
```
yosefe, line 173: 1. maybe use (UCT_DC_MLX5_HW_DCI_INDEX + 1) instead of 1?
2. first_allocated_dci -> first_sw_dci ?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -165,10 +165,14 @@ UCS_TEST_P(test_dc, dcs_single) {
...
+
```
yosefe, line 174: extra space line

test/gtest/uct/ib/test_dc.cc
```diff
@@ -209,8 +214,10 @@ UCS_TEST_P(test_dc, dcs_multi) {
     }
 
     /* this should fail because there are no free dci */
-    status = uct_ep_am_short(m_e1->ep(i), 0, 0, NULL, 0);
-    EXPECT_EQ(UCS_ERR_NO_RESOURCE, status);
+    if (!uct_dc_mlx5_iface_is_hybrid(iface)) {
+        status = uct_ep_am_short(m_e1->ep(i), 0, 0, NULL, 0);
+        EXPECT_EQ(UCS_ERR_NO_RESOURCE, status);
```
yosefe, line 219: maybe we can move uct_ep_am_short() before (outside) the uct_dc_mlx5_iface_is_hybrid check?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -296,6 +303,10 @@ UCS_TEST_P(test_dc, dcs_ep_flush_pending, "DC_NUM_DCI=1") {
 
     iface = dc_iface(m_e1);
 
+    if (uct_dc_mlx5_iface_is_hybrid(iface)) {
+        UCS_TEST_SKIP_R("hybrid DCI policy does not support pending flush requests while ndci = 1");
```
yosefe, line 307: 1. line seems too long
2. why it's not supported?

test/gtest/uct/ib/test_dc.cc
```diff
@@ -345,6 +356,10 @@ UCS_TEST_P(test_dc, dcs_ep_purge_pending, "DC_NUM_DCI=1") {
     m_e1->connect_to_iface(0, *m_e2);
 
     iface = dc_iface(m_e1);
+    if (uct_dc_mlx5_iface_is_hybrid(iface)) {
+        UCS_TEST_SKIP_R("hybrid DCI policy does not support pending purge requests while ndci = 1");
```
yosefe, line 360: ditto

yosefe (changes requested):

src/uct/ib/mlx5/dc/dc_mlx5.inl
roiedanino: Yes, because we want to use "shared policy" logic also when in hybrid mode

yosefe: why? where is this field used in hybrid mode?

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
roiedanino: We can't because `is_dci_shared` is not checking wether the dci is NULL (it will fail in random policy for example)

test/gtest/uct/ib/test_dc.cc
roiedanino: The test assumes that we only have a single DCI when setting DC_NUM_DCI=1 but in hybrid mode we also have the HW DCI which will prevent out-of-resource status

test/gtest/uct/ib/test_dc.cc
roiedanino: its not supported because we also have a shared DCI, and the tests expect that we will run out of resources at some point, shared policies are hanging in this test as well

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
yosefe: how can the DCI be NULL, we are releasing it after it was used?

test/gtest/uct/ib/test_dc.cc
yosefe: you mean shared policies never run out of resources?

roiedanino: Not so quickly, because shared policies are not relying on ndci (per-pool number of dcis) because they are not really using the pools, allocation methods are different

yosefe: is there another way to cause shared policy to run out of resources, like limiting QP size and/or number of DCIs/channels?

roiedanino: I tried few things, but currently MAX_BB is not affecting DC as max_bb is queried from a temporary dci created in iface open, we will have to add an environment for that or try to use the one from RC to affect it.

To try that?

yosefe:
@roiedanino pls not the merge conflict

src/uct/ib/mlx5/dc/dc_mlx5.inl
```diff
@@ -41,17 +41,29 @@ uct_dc_mlx5_get_arbiter_params(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep,
...
+
+static UCS_F_ALWAYS_INLINE void
+uct_dc_mlx5_ep_pending_common_shared(uct_dc_mlx5_iface_t *iface,
+                                     uct_dc_mlx5_ep_t *ep, uct_pending_req_t *r,
+                                     unsigned flags, int push_to_head,
+                                     int schedule)
+{
+    ucs_arbiter_group_t *group = uct_dc_mlx5_ep_shared_arb_group(iface, ep);
+
+    uct_dc_mlx5_pending_req_priv(r)->ep = ep;
```
yosefe, line 62: i'd also move the UCS_STATIC_ASSERT from uct_dc_mlx5_ep_pending_common to here, since we use uct_dc_mlx5_pending_req_priv only in this function now

test/gtest/uct/ib/test_dc.cc
yosefe: i would expect UCX_DC_MLX5_TX_QUEUE_LEN to affect the DCI QP size

yosefe (approved):

PR #10250, opened 2024-10-23
PERF: handle error status in send/recv callbacks

evgeny-leksikov:
## What
handle error status in send/recv callbacks of ucx_perftest_daemon

## Why ?
to avoid segfault on cleanup

### DIFF
--- a/src/tools/perf/perftest_daemon.c
+++ b/src/tools/perf/perftest_daemon.c
@@ -163,8 +163,11 @@ ucp_perf_daemon_send_cb(void *request, ucs_status_t status, void *user_data)
 {
     ucp_perf_daemon_context_t *ctx = user_data;
 
-    ucp_perf_daemon_send_am_eager_reply(ctx->host_ep,
-                                        UCP_PERF_DAEMON_AM_ID_SEND_CMPL);
+    if (ucs_likely(status == UCS_OK)) {
+        ucp_perf_daemon_send_am_eager_reply(ctx->host_ep,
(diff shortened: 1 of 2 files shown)


yosefe (approved):

iyastreb (commented):
This is a meaningful change, but I have already addressed it in my PR long time ago https://github.com/openucx/ucx/pull/9752

Basically I do a bit more in my fix (https://github.com/openucx/ucx/pull/9752/files#diff-d7366ea9b4351766c62598815028afa86499ee9e6155779da3d81ec27e22e1b3R200):
1. Log error/success on daemon side
2. transfer whatever status to the host
3. Handle status on the host side (don't blindly expect OK) https://github.com/openucx/ucx/pull/9752/files#diff-aa76d808cfbc2be13c3119e08470e1341bbe8f3f623f3f797780ca0c07d5db20R347

So I'm not sure if it makes sense to commit this improvement

evgeny-leksikov:
> This is a meaningful change, but I have already addressed it in my PR long time ago #9752
> ...
> So I'm not sure if it makes sense to commit this improvement

Hi @iyastreb, from quick review I think our changes handle different errors in different UCX states, in my case I have got an error on cleanup when host EP is not accessible any more. Probably I'm missing something and the error might be handled earlier in your case, I would suggest you to split up your PR to handle errors one-be-one to simplify review and speedup upstreaming process.

iyastreb (approved):

PR #10251, opened 2024-10-23
CONF/GRACE: Disable ODP for cuda-managed - v1.18.x

brminich:
## What
Disable ODP for cuda-managed by default, because currently async memory is detected as cuda-managed and it does not seem support ODP

### DIFF
--- a/config/ucx.conf
+++ b/config/ucx.conf
@@ -1,7 +1,6 @@
 [Grace Hopper]
 CPU model=Grace
-UCX_REG_NONBLOCK_MEM_TYPES=host,cuda-managed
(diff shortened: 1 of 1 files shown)


config/ucx.conf
```diff
@@ -1,7 +1,6 @@
 [Grace Hopper]
 CPU model=Grace
-UCX_REG_NONBLOCK_MEM_TYPES=host,cuda-managed
-UCX_IB_ODP_MEM_TYPES=host,cuda-managed
```
tvegas1, line 4: do we need to still keep `host` for `UCX_IB_ODP_MEM_TYPES`?

PR #10255, opened 2024-10-23
UCP: add gga_mlx5 to UCP transport aliases

evgeny-leksikov:
## What

1. add `gga_mlx5` to `ib` transport alias
2. add `gga` alias for `gga_mlx5`
3. add testing of `shm,gga`
4. fix align_mtu in `gga_mlx5` transport
5. fix test_proto_reset  test for `shm,gga` case

## Why ?
start point is `UCX_TLS=^ib` use case

## Note:
~~in dependency from https://github.com/openucx/ucx/pull/10236, early merging may cause failures of MT tests~~

### DIFF
--- a/test/gtest/ucp/test_ucp_request.cc
+++ b/test/gtest/ucp/test_ucp_request.cc
@@ -389,6 +389,7 @@ public:
         std::vector<void*> reqs;
         send_requests(1, reqs, op, sync);
         wait_reqs(op, reqs);
+        flush_workers();
 
         /* Send all messages */
         send_requests(reqs_count, reqs, op, sync);
--- a/contrib/lsan.supp
+++ b/contrib/lsan.supp
@@ -2,3 +2,5 @@ leak:libcuda
 leak:dlsym
 leak:nvmlInitWithFlags
 leak:bfd_map_over_sections
+# see https://github.com/openucx/ucx/blob/ec4407f08dbed639409d796e42bfab1ff718fda5/src/uct/ib/base/ib_md.c#L1392
+leak:ibv_alloc_pd 
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -659,7 +659,8 @@ static ucp_tl_alias_t ucp_tl_aliases[] = {
   { "mm",    { "posix", "sysv", "xpmem", NULL } }, /* for backward compatibility */
   { "sm",    { "posix", "sysv", "xpmem", "knem", "cma", NULL } },
   { "shm",   { "posix", "sysv", "xpmem", "knem", "cma", NULL } },
-  { "ib",    { "rc_verbs", "ud_verbs", "rc_mlx5", "ud_mlx5", "dc_mlx5", NULL } },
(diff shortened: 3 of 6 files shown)


brminich (dismissed):
do you think it makes sence to create `gga` alias for `gga_mlx5`?

evgeny-leksikov:
> do you think it makes sence to create `gga` alias for `gga_mlx5`?

for now `gga_mlx5` is useless as standalone transport due to lack of AM support, so it should be combined with some additional transports.

yosefe:
> > do you think it makes sence to create `gga` alias for `gga_mlx5`?
> 
> for now `gga_mlx5` is useless as standalone transport due to lack of AM support, so it should be combined with some additional transports.

Nevertheless, users could specify something like UCX_TLS=gga,rc

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -377,17 +377,23 @@ class test_proto_reset : public ucp_test {
...
             m_rbufs.push_back(std::unique_ptr<mapped_buffer>(rbuf));
             m_sbufs.push_back(std::vector<uint8_t>(msg_size));
         }
 
-        /* Send a single message to complete wireup before sending actual
+        /* Send a bunch of messages to complete wireup before sending actual
            data */
+        /* TODO: a single message is not enough to complete wireup with shm,gga
+         *       TLs set, should not we be able to restart request with wireup
+         *       message as well?
```
yosefe, line 393: why one is not enough? maybe need to flush or wait for receive?

evgeny-leksikov: AFAIR flush did not help, should not sync transfer guarantee request completion on receiver side as well? 
I don't see much difference with 1 or 5 warmup requests for this specific test, how about restart of wireup msg request? Cannot it be a real problem?

yosefe: it's not a problem to have several wireup messages, but the comment here is not clear, and why 5 is enough and not 50? and why the problem is uncovered after adding gga, is it slower to ininitialize?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -377,7 +377,7 @@ class test_proto_reset : public ucp_test {
                         unsigned reqs_count = 1000,
                         size_t msg_size = UCS_KBYTE * 70)
     {
-        unsigned warmup_req_count = 5;
+        unsigned warmup_req_count = 1;
```
yosefe, line 380: if the number if iterations is 1 now, do we still need a for loop?

evgeny-leksikov: which one? the PR does not add any new loop, only "fixes" condition for buffers allocation which is correct and type of iterators (usually compilers warn about signed/unsigned comparison).

just noticed leftover in the comment below which must be reverted, so please LMK if I you see smth else to be addressed.

yosefe (approved):

yosefe:
@evgeny-leksikov pls squash

evgeny-leksikov:
> @evgeny-leksikov pls squash

Squashed to few independent commits for different modules, LMK if you prefer single commit

yosefe (approved):

PR #10256, opened 2024-10-24
ARCH/X86: make nt_all transfer a seperate function

arun-chandran-edarath:
## What
It introduces a new function ucs_x86_nt_all_buffer_transfer() to handle the case when both source and destination are non-temporal

Other notable changes.

a) increase nt_dest_thresh
b) Smaller transfer sizes are improved for ucs_x86_nt_dst_buffer_transfer() & ucs_x86_nt_src_buffer_transfer()

## Why ?
It is an enhancement

### DIFF
--- a/src/ucs/arch/x86_64/cpu.c
+++ b/src/ucs/arch/x86_64/cpu.c
@@ -605,7 +605,7 @@ static size_t ucs_cpu_nt_bt_thresh_min(size_t user_val)
 static size_t ucs_cpu_nt_dest_thresh()
 {
     if (ucs_arch_get_cpu_vendor() == UCS_CPU_VENDOR_AMD) {
-        return ucs_cpu_get_cache_size(UCS_CPU_CACHE_L3) * 29 / 32;
+        return ucs_cpu_get_cache_size(UCS_CPU_CACHE_L3) * 9 / 8;
     } else {
         return UCS_MEMUNITS_INF;
     }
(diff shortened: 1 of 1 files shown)


arun-chandran-edarath:
This should go to the 1.18 branch also, do I need to raise another PR for that?

src/ucs/arch/x86_64/cpu.c
```diff
@@ -734,16 +734,117 @@ ucs_status_t ucs_arch_get_cache_size(size_t *cache_sizes)
...
+            _mm256_stream_si256(UCS_PTR_BYTE_OFFSET(dst, offset), y0);
+            _mm256_stream_si256(UCS_PTR_BYTE_OFFSET(dst, offset + 32), y1);
+            _mm256_stream_si256(UCS_PTR_BYTE_OFFSET(dst, offset + 64), y2);
+            _mm256_stream_si256(UCS_PTR_BYTE_OFFSET(dst, offset + 96), y3);
+            _mm256_stream_si256(UCS_PTR_BYTE_OFFSET(dst, offset + 128), y4);
+            _mm256_stream_si256(UCS_PTR_BYTE_OFFSET(dst, offset + 160), y5);
+            _mm256_stream_si256(UCS_PTR_BYTE_OFFSET(dst, offset + 192), y6);
+            _mm256_stream_si256(UCS_PTR_BYTE_OFFSET(dst, offset + 224), y7);
+
+            if ((len > 1024) && (((offset >> 8) & 3) == 0)) {
```
tvegas1, line 771: I understand that when this if() is true, we prefetch range `[src+offset, src+offset+1024[`. Also `offset >> 8` counts iterations (256 bytes at time). So with `... & 3) == 0` we enter the block every 8x256=2048 bytes, to start prefetching not for the next immediate iteration but the one after.

Should not we prefetch 2048 bytes at a time instead of 1024, or prefetch every 1024 bytes instead of 2048, (using 2 instead of 3 like `(offset>>8)&2)`)?

arun-chandran-edarath: It is designed to use prefetch with NTA only on the second half of the cache lines on a block of 1024 bytes(16 cache lines). That is the reason only 8 lines are prefetched using 'NTA'.

tvegas1: Thanks, re-reading now I see that we prefetch 512 bytes at a time (8 lines) on a 1024 bytes period.

tvegas1 (approved):

tvegas1 (approved):

yosefe (approved):

yosefe:
@arun-chandran-edarath can you pls squash?

arun-chandran-edarath:
> @arun-chandran-edarath can you pls squash?

Done. Thanks.

PR #10258, opened 2024-10-24
UCT/CUDA_IPC: Increase MAX_PEERS and MAX_STREAMS

Akshay-Venkatesh:
## Why ?
Need this ensure greater concurrency for a larger count of peers reachable by cuda_ipc.

### DIFF
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
@@ -42,7 +42,7 @@ static ucs_config_field_t uct_cuda_ipc_iface_config_table[] = {
      "Max number of event completions to pick during cuda events polling",
       ucs_offsetof(uct_cuda_ipc_iface_config_t, params.max_poll), UCS_CONFIG_TYPE_UINT},
 
-    {"MAX_STREAMS", "16",
+    {"MAX_STREAMS", UCS_PP_MAKE_STRING(UCT_CUDA_IPC_MAX_PEERS),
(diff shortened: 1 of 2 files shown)


src/uct/cuda/cuda_ipc/cuda_ipc_iface.c
```diff
@@ -30,7 +30,7 @@ static ucs_config_field_t uct_cuda_ipc_iface_config_table[] = {
      "Max number of event completions to pick during cuda events polling",
       ucs_offsetof(uct_cuda_ipc_iface_config_t, max_poll), UCS_CONFIG_TYPE_UINT},
 
-    {"MAX_STREAMS", "16",
+    {"MAX_STREAMS", "128",
```
yosefe, line 33: can we use UCS_PP_MAKE_STRING(UCT_CUDA_IPC_MAX_PEERS) here?
what happens if the user sets MAX_STREAMS config to a value larger than UCT_CUDA_IPC_MAX_PEERS?

Akshay-Venkatesh: Having larger MAX_STREAMS than UCT_CUDA_IPC_MAX_PEERS allows cuda-ipc transport to round-robin select among more number of streams. It wouldn't be a functional or performance issue in my opinion although it is unlikely to improve performance.

brminich (approved):

yosefe (approved):
What happens if a user sets the value of MAX_STREAMS to >128?
Seems it's not documented and not checked that the config value does not exceed the array size

brminich (approved):

PR #10262, opened 2024-10-28
UCT/IB/TEST: Configure available SLs in gtest

amastbaum:
## What
Configure available SLs in gtest.

## Why ?
Right now, the `test_uct_ib_sl` case in gtest assumes that all SLs map to a valid, available VL, but that’s not always the case. Some systems don’t support all VLs. This change lets us set which SLs are available, so gtest can check only the relevant ones.

## How ?
Introduce an environment variable as a bitmask to specify the SLs to be checked.

### DIFF
--- a/test/gtest/uct/ib/test_ib.cc
+++ b/test/gtest/uct/ib/test_ib.cc
@@ -445,9 +445,19 @@ protected:
 };
 
 UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
+    const char *max_avail_sl_str = getenv("GTEST_MAX_IB_SL");
+    uint8_t sl, max_avail_sl;
+
+    if (max_avail_sl_str == NULL) {
+        max_avail_sl = UCT_IB_SL_NUM;
+    } else {
+        max_avail_sl = std::min(strtoul(max_avail_sl_str, NULL, 0),
+                                static_cast<unsigned long>(UCT_IB_SL_NUM));
+    }
+
     // go over all SLs, check UCTs could be initialized on a specific SL
     // and able to send/recv traffic
-    for (uint8_t sl = 0; sl < UCT_IB_SL_NUM; ++sl)  {
+    for (sl = 0; sl < max_avail_sl; ++sl) {
         if (!has_transport("rc_verbs") && !has_transport("ud_verbs")) {
             // if AR is configured on the given SL, set AR_ENABLE to "y",
             // otherwise - to "n" in order to test that AR_ENABLE parameter


test/gtest/uct/ib/test_ib.cc
```diff
@@ -445,9 +445,19 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
+    if (avail_sls_str == NULL) {
+        avail_sls = 0xffff;
+    } else {
+        avail_sls = (unsigned short)strtoul(avail_sls_str, NULL, 16);
+    }
+
     // go over all SLs, check UCTs could be initialized on a specific SL
     // and able to send/recv traffic
-    for (uint8_t sl = 0; sl < UCT_IB_SL_NUM; ++sl)  {
+    for (sl = 0; (sl < UCT_IB_SL_NUM) && ((UCS_BIT(sl) & avail_sls) != 0); ++sl)  {
```
gleon99, line 460: Nope. The first unavail SL will kick out from the loop.

amastbaum: Right, I am just fixing it

test/gtest/uct/ib/test_ib.cc
```diff
@@ -445,9 +445,23 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
 };
 
 UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
+    const char *avail_sls_str = getenv("GTEST_AVAIL_SL");
```
gleon99, line 448: `avail_sl_mask_str`

test/gtest/uct/ib/test_ib.cc
```diff
@@ -445,9 +445,23 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
+    unsigned short avail_sls;
```
gleon99, line 449: `avail_sl_mask`

test/gtest/uct/ib/test_ib.cc
```diff
@@ -445,9 +445,23 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
+        avail_sls = 0xffff;
+    } else {
+        avail_sls = (unsigned short)strtoul(avail_sls_str, NULL, 16);
+    }
+
     // go over all SLs, check UCTs could be initialized on a specific SL
     // and able to send/recv traffic
-    for (uint8_t sl = 0; sl < UCT_IB_SL_NUM; ++sl)  {
+    for (sl = 0; sl < UCT_IB_SL_NUM; ++sl) {
+        if ((UCS_BIT(sl) & avail_sls) == 0) {
```
gleon99, line 461: Can simplify to `!` instead of `== 0`

test/gtest/uct/ib/test_ib.cc
```diff
@@ -445,9 +445,23 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
 
 UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
+    const char *avail_sl_mask_str = getenv("GTEST_AVAIL_SL");
+    unsigned short avail_sl_mask;
+    uint8_t sl;
+
+    if (avail_sl_mask_str == NULL) {
+        avail_sl_mask = 0xffff;
+    } else {
+        avail_sl_mask = (unsigned short)strtoul(avail_sl_mask_str, NULL, 16);
```
tvegas1, line 455: minor: since it would be small change, maybe add something like below as helper function:
```
int long parse_ul(const char *str, unsigned long *value) {
  int base = 10;
  if (!strncmp(str, "0x", 2)) {
    base = 16;
    str += 2;
  }
  mask = strtoul(str, &end, base);
  if (*end) {
    // err
  }
}

if (mask > UINT16_MAX
```

gleon99: Not sure we need it.
The `GTEST_AVAIL_SL` value will be passed from a script in a very specific predefined format, it has only a single user..

tvegas1: up to you, but change is small, any help in spotting typos/sanitize mask could help

yosefe: IMO instead of passing a mask from the user, pass maximal SL value

amastbaum: @yosefe so it means the user could not configure specific SLs but just use the first `n` SLs instead, right?

gleon99: It means we'll provide the max avail SL from the env, then inside the test we try up to this SL.

yosefe: yes, IMO it is simpler and good enough

test/gtest/uct/ib/test_ib.cc
```diff
@@ -445,9 +445,19 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
 };
 
 UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
+    const char *max_avail_sl_str = getenv("GTEST_MAX_AVAIL_SL");
```
yosefe, line 448: IMO "GTEST_MAX_IB_SL" is beter name

test/gtest/uct/ib/test_ib.cc
```diff
@@ -445,9 +445,19 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
+    uint8_t sl, max_avail_sl;
+
+    if (max_avail_sl_str == NULL) {
+        max_avail_sl = UCT_IB_SL_NUM;
+    } else {
+        max_avail_sl = (uint8_t)strtoul(max_avail_sl_str, NULL, 10);
+        ASSERT_TRUE(max_avail_sl <= UCT_IB_SL_NUM);
```
yosefe, line 455: ASSERT_LE

test/gtest/uct/ib/test_ib.cc
```diff
@@ -445,9 +445,19 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
 };
 
 UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
+    const char *max_avail_sl_str = getenv("GTEST_MAX_AVAIL_SL");
+    uint8_t sl, max_avail_sl;
+
+    if (max_avail_sl_str == NULL) {
+        max_avail_sl = UCT_IB_SL_NUM;
+    } else {
+        max_avail_sl = (uint8_t)strtoul(max_avail_sl_str, NULL, 10);
```
yosefe, line 454: IMO the last param to strtoul can be "0" so it would detect the base automatically

yosefe (dismissed):

gleon99:
@amastbaum please squash

test/gtest/uct/ib/test_ib.cc
```diff
@@ -452,7 +452,9 @@ UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
         max_avail_sl = UCT_IB_SL_NUM;
     } else {
         max_avail_sl = (uint8_t)strtoul(max_avail_sl_str, NULL, 0);
-        ASSERT_LE(max_avail_sl, UCT_IB_SL_NUM);
+        if (max_avail_sl > UCT_IB_SL_NUM) {
+            max_avail_sl = UCT_IB_SL_NUM;
+        }
```
amastbaum, line 457: Before squashing I had to solve a coverity issue that says we should make sure that `max_avail_sl`'s value is not out of bounds.
@gleon99 / @yosefe  Do you know why `ASSERT_LE` is not enough here?

yosefe: i guess it's because of casting to uint8_t, that is before the ASSERT_LE; i gues we can do something like
`max_avail_sl = std::min(strtoul(max_avail_sl_str, NULL, 0), UCT_IB_SL_NUM)`

gleon99 (dismissed):

gleon99:
plz sauqsh

test/gtest/uct/ib/test_ib.cc
```diff
@@ -445,9 +445,19 @@ class test_uct_ib_sl : public test_uct_ib_with_specific_port {
...
 
 UCS_TEST_P(test_uct_ib_sl, check_ib_sl_config) {
+    const char *max_avail_sl_str = getenv("GTEST_MAX_IB_SL");
+    uint8_t sl, max_avail_sl;
+
+    if (max_avail_sl_str == NULL) {
+        max_avail_sl = UCT_IB_SL_NUM;
+    } else {
+        max_avail_sl = (uint8_t)std::min(strtoul(max_avail_sl_str, NULL, 0),
+                                         (unsigned long)UCT_IB_SL_NUM);
```
yosefe, line 455:
```suggestion
        max_avail_sl = std::min(strtoul(max_avail_sl_str, NULL, 0),
                                static_cast<unsigned long>(UCT_IB_SL_NUM));
```

gleon99 (approved):

PR #10263, opened 2024-10-28
UCP/PROTO/MULTI: Allow header be greater than max_frag

iyastreb:
## What
This is a fix for [RM#4134043](https://redmine.mellanox.com/issues/4134043), which is a regression introduced in https://github.com/openucx/ucx/pull/9920

This issue happens when MAX_EAGER_LANES > 1

## Why ?
In multi-fragment protocols we set `max_frag` per lane proportionally to it's bandwidth. However there is a single AM lane that could get `max_frag` assigned less than a maximum header size. This provokes an assertion

## How ?
1. Remove an assertion checking that header size must be smaller than `max_frag`
2. If header is greater than `max_frag` - that's ok, we allow it, but we send only the header data and keep the payload empty


brminich (commented):
imo, it seems 2 separate PRs are worth creating, because you are addressing two separate issues.

iyastreb:
> imo, it seems 2 separate PRs are worth creating, because you are addressing two separate issues.

Ok, will split them

src/ucp/proto/proto_multi.inl
```diff
@@ -66,8 +66,11 @@ ucp_proto_multi_max_payload(ucp_request_t *req,
     size_t max_frag = lpriv->max_frag - hdr_size;
     size_t max_payload;
 
-    ucs_assertv(lpriv->max_frag > hdr_size, "max_frag=%zu hdr_size=%zu",
-                lpriv->max_frag, hdr_size);
+    /* If header is greater than max_frag - that's ok, we allow it, but we send
+     * only the header data and keep the payload empty */
+    if (lpriv->max_frag <= hdr_size) {
```
yosefe, line 71: maybe smth like
```
ssize_t max_frag;
...
max_frag = lpriv->max_frag - hdr_size;
if (max_frag < 0) {
     return 0;
}
```

yosefe (commented):
do we need this for v1.18.x?

src/ucp/proto/proto_multi.inl
iyastreb: ok, good idea

iyastreb:
> do we need this for v1.18.x?

@yosefe I'm not sure if we need to backport this to 1.18, because this issue happens only with MAX_EAGER_LANES > 1, which is not default config

brminich:
can you laso remove UD restrictions from
https://github.com/openucx/ucx/blob/master/test/gtest/ucp/test_ucp_am.cc#L1588
and https://github.com/openucx/ucx/blob/master/test/gtest/ucp/test_ucp_am.cc#L1599
as these tests are supposed to pass with this fix

brminich:
> > do we need this for v1.18.x?
> 
> @yosefe I'm not sure if we need to backport this to 1.18, because this issue happens only with MAX_EAGER_LANES > 1, which is not default config

imo, no need to backport

iyastreb:
> can you laso remove UD restrictions from https://github.com/openucx/ucx/blob/master/test/gtest/ucp/test_ucp_am.cc#L1588 and https://github.com/openucx/ucx/blob/master/test/gtest/ucp/test_ucp_am.cc#L1599 as these tests are supposed to pass with this fix

brminich (approved):
lgtm, if no issues are found with re-enabled tests

src/ucp/proto/proto_multi.inl
```diff
@@ -62,12 +62,19 @@ ucp_proto_multi_max_payload(ucp_request_t *req,
...
     size_t max_payload;
 
-    ucs_assertv(lpriv->max_frag > hdr_size, "max_frag=%zu hdr_size=%zu",
-                lpriv->max_frag, hdr_size);
+    /* If header is greater than max_frag - that's ok, we allow it, but we send
+     * only the header data and 1 byte of payload.
+     * 1 byte is needed to distinguish between first and middle fragments that
+     * we do by comparing req->send.state.dt_iter.offset with 0.
+     * @see ucp_worker_set_max_am_header
+     */
```
yosefe, line 74: since we decided that no need to backport this fix to v1.18.x, how hard it will be to fix according to
```
         * TODO: fix generic AM based multi-fragment protocols, so that this
         * trick is not needed.
```

src/ucp/core/ucp_worker.c
```diff
@@ -2359,17 +2359,11 @@ static void ucp_worker_set_max_am_header(ucp_worker_h worker)
...
-         * TODO: fix generic AM based multi-fragment protocols, so that this
-         * trick is not needed.
-         */
+         * first fragment header and footer. */
         if (if_attr->cap.flags & UCT_IFACE_FLAG_AM_BCOPY) {
             max_uct_fragment = ucs_max(if_attr->cap.am.max_bcopy,
-                                       max_ucp_header - 1) -
-                               max_ucp_header - 1;
+                                       max_ucp_header) -
+                               max_ucp_header;
```
ivankochin, line 2366: Minor, seems just a little more readable for me:

```c
            max_uct_fragment = ucs_max(if_attr->cap.am.max_bcopy - 
                                       max_ucp_header, 0);
```

src/ucp/dt/datatype_iter.c
```diff
@@ -265,11 +265,6 @@ size_t ucp_datatype_iter_iov_next_iov(const ucp_datatype_iter_t *dt_iter,
     ucs_assertv(length <= max_length, "length=%zu max_length=%zu", length,
                 max_length);
 
-    /* Check that if not reached the end, packed data is not empty */
-    ucs_assertv((dt_iter->offset == dt_iter->length) || (length > 0),
```
ivankochin, line 269: Is that true that all middle fragments still should always send some payload? Maybe somehow save this check for middle segments only (e.g. check offset)?

src/ucp/proto/proto_multi.inl
ivankochin: AFAIR `lpriv->max_frag` can be set to `lane_perf->max_frag` which is a transport limitation, so may there be a situation when some lanes don't really support such big header size so we cannot send even header?

test/gtest/ucp/test_ucp_am.cc
```diff
@@ -1584,9 +1584,8 @@ class test_ucp_am_nbx_dts : public test_ucp_am_nbx_reply {
 
 /* Skip tests for ud_v and ud_x because of unstable reproducible failures during
```
ivankochin, line 1585: Redundant comment?

src/ucp/core/ucp_worker.c
iyastreb: It's more readable, but not always correct. What if max_ucp_header > if_attr->cap.am.max_bcopy?
We have to be careful when deducting one `size_t` from the other, because we can easily get a "negative" result which translates into some huge number. I guess it exists in this form for exactly this reason - to always guarantee the correctness

src/ucp/dt/datatype_iter.c
iyastreb: Yes, it's true
To have this assert I would need to propagate request (or at least request flags/boolean) to this function and to all the functions calling it (`ucp_datatype_iter_next_iov` refs in 7 other files). If it's really important to have this check, we can do that.
@yosefe - since you added this assertion, do we need to keep it even at the cost of adding one more param to this function?

test/gtest/ucp/test_ucp_am.cc
iyastreb: yes, thanks

src/ucp/proto/proto_multi.inl
iyastreb: As I see in `ucp_worker_set_max_am_header` the `worker->max_am_header` is set so that header can fit in any lane. If user sends something larger than that, then we reject this input:
```C
ucp_am_send_nbx_check_header_length(ucp_worker_h worker, size_t header_length)
{
    if (ENABLE_PARAMS_CHECK && (header_length > worker->max_am_header)) {
```

src/ucp/proto/proto_multi.inl
iyastreb: done

src/ucp/dt/datatype_iter.c
ivankochin: I don't think that it is necessary to add request argument here. We can pass only information that is needed and not to this function but to `ucp_datatype_iter_next_iov`. Also maybe we can do that even without passing request at all by relaxing this assertion (E.g. check that only if offset > 0) WDYT?

iyastreb: Ok, I add relaxed check, still maybe useful

src/ucp/dt/datatype_iter.c
```diff
@@ -266,7 +266,8 @@ size_t ucp_datatype_iter_iov_next_iov(const ucp_datatype_iter_t *dt_iter,
                 max_length);
 
     /* Check that if not reached the end, packed data is not empty */
-    ucs_assertv((dt_iter->offset == dt_iter->length) || (length > 0),
+    ucs_assertv((dt_iter->offset == 0) ||
```
brminich, line 269: can you pls update the comment above and mention new check?

test/gtest/ucp/test_ucp_request.cc
```diff
@@ -452,7 +452,8 @@ class test_proto_reset : public ucp_test {
     {
         const ucp_datatype_iter_t *dt_iter = &req->send.state.dt_iter;
 
-        return !ucp_datatype_iter_is_begin(dt_iter) &&
+        return ((req->flags & UCP_REQUEST_FLAG_MIDDLE_FRAGMENT) ||
```
brminich, line 455: use `ucp_proto_multi_is_middle_fragment`?

iyastreb: Yes, looks logical, it was my initial impl, but then I removed it, because including "proto_multi.inl" into C++ translation unit requires quite some unrelated changes in that file: see my reverted commit which fixes some of them: https://github.com/openucx/ucx/pull/10263/commits/e74c74ac6d39d5bed8dd3d019fc7cb71ca82a55a

So I thought it's not worth it.
Else we can move `ucp_proto_multi_is_middle_fragment` function to `proto_multi.h`

src/ucp/dt/datatype_iter.c
iyastreb: sure

src/ucp/proto/proto_common.c
```diff
@@ -806,7 +807,8 @@ void ucp_proto_request_restart(ucp_request_t *req)
     }
 
     /* Select a protocol with resume request support */
-    if (!ucp_datatype_iter_is_begin(&req->send.state.dt_iter)) {
+    if (ucp_proto_multi_is_middle_fragment(req) ||
```
yosefe, line 810: this is active-message specific, so maybe the protocol "reset" callback should provide the indication of whether a "RESUME" flags is going to be required or not, instead of just deciding it according to ucp_datatype_iter_is_begin

src/ucp/proto/proto_multi.inl
```diff
@@ -40,6 +40,18 @@ ucp_proto_multi_request_init(ucp_request_t *req)
...
 }
 
+static UCS_F_ALWAYS_INLINE int
+ucp_proto_multi_is_middle_fragment(const ucp_request_t *req)
+{
+    return req->flags & UCP_REQUEST_FLAG_MIDDLE_FRAGMENT;
+}
+
+static UCS_F_ALWAYS_INLINE int
+ucp_proto_multi_is_first_fragment(const ucp_request_t *req)
```
yosefe, line 50: do we really need both wrappers? just for the inverse check?

src/ucp/proto/proto_multi.inl
yosefe: in this case, better not calculate `max_frag = lpriv->max_frag - hdr_size` during variable declaration, but do it after checking `(lpriv->max_frag <= hdr_size)`

src/ucp/proto/proto_common.c
iyastreb: I reduced the scope of the changes to AM eager multi, and I don't see test failures in this function

src/ucp/core/ucp_request.h
```diff
@@ -66,6 +66,14 @@ enum {
 };
 
 
+/**
+ * AM specific internal request flags
+ */
+enum ucp_request_am_impl_flags {
```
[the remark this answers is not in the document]
iyastreb: sure

src/ucp/proto/proto_am.inl
```diff
@@ -22,6 +22,14 @@
 #define UCP_STATUS_PENDING_SWITCH (UCS_ERR_LAST - 1)
 
 
+/**
+ * AM specific internal request flags
+ */
+enum ucp_request_am_impl_flags {
+    UCP_REQUEST_AM_FLAG_MIDDLE_FRAGMENT = UCS_BIT(0)
```
brminich, line 29: minor: 
```suggestion
    UCP_REQUEST_FLAG_AM_MIDDLE_FRAGMENT = UCS_BIT(0)
```

brminich (approved):

src/ucp/proto/proto_am.inl
iyastreb: ok

brminich (approved):

src/ucp/proto/proto_am.inl
```diff
@@ -22,6 +22,14 @@
 #define UCP_STATUS_PENDING_SWITCH (UCS_ERR_LAST - 1)
 
 
+/**
+ * AM specific internal request flags
+ */
+enum ucp_request_am_impl_flags {
+    UCP_REQUEST_FLAG_AM_MIDDLE_FRAGMENT = UCS_BIT(0)
```
yosefe, line 29: UCP_REQUEST_AM_FLAG_HEADER_SENT

src/ucp/core/ucp_request.h
```diff
@@ -206,7 +205,11 @@ struct ucp_request {
                             } header UCS_S_PACKED; /* packed to avoid 32-bit
                                                       padding */
                             uint16_t       am_id;
+                            /* API flags from @ref ucp_send_am_flags enum */
                             uint16_t       flags;
+                            /* Internal implementation flags from @ref
+                             * ucp_request_am_impl_flags enum */
+                            uint8_t        impl_flags;
```
yosefe, line 212: impl_flags sounds weird.. and seems it increases "am" struct size..
maybe reduce "flags" to uint8, and rename impl_flags to internal_flags?

src/ucp/am/eager_multi.c
```diff
@@ -78,6 +78,7 @@ static size_t ucp_am_eager_multi_bcopy_pack_args_first(void *dest, void *arg)
 
     ucs_assertv(req->send.state.dt_iter.offset == 0, "offset %zu",
                 req->send.state.dt_iter.offset);
+    ucs_assert(ucp_proto_multi_is_first_fragment(req));
```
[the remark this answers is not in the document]
iyastreb: ok

src/ucp/core/ucp_request.h
yosefe: or
hdr_sent: 1;

src/ucp/proto/proto_am.inl
```diff
@@ -52,14 +52,18 @@ ucp_proto_am_handle_user_header_send_status(ucp_request_t *req,
...
 ucp_proto_am_is_first_fragment(const ucp_request_t *req)
 {
-    return !(req->send.msg_proto.am.impl_flags &
-             UCP_REQUEST_FLAG_AM_MIDDLE_FRAGMENT);
+    int is_first = !(req->send.msg_proto.am.internal_flags &
+                     UCP_REQUEST_AM_FLAG_HEADER_SENT);
+    ucs_assertv(!is_first || (req->send.state.dt_iter.offset == 0),
+                "is_first=%d offset=%zu", is_first,
+                req->send.state.dt_iter.offset);
+    return is_first;
```
yosefe, line 60:
```suggestion
    if (req->send.msg_proto.am.internal_flags &
                      UCP_REQUEST_AM_FLAG_HEADER_SENT) {
        return 0;
    }
    
    ucs_assertv(req->send.state.dt_iter.offset == 0, "req=%p offset=%zu" ...);
    return 1;
}
```

src/ucp/core/ucp_request.h
iyastreb: fine for me
In this case we don't really need the ucp_request_am_internal_flags enum

yosefe (dismissed):

src/ucp/core/ucp_request.h
```diff
@@ -205,7 +205,10 @@ struct ucp_request {
                             } header UCS_S_PACKED; /* packed to avoid 32-bit
                                                       padding */
                             uint16_t       am_id;
-                            uint16_t       flags;
+                            /* API flags from @ref ucp_send_am_flags enum */
+                            uint8_t        flags: 7;
+                            /* Whether header is sent already */
+                            uint8_t        hdr_sent: 1;
```
brminich, line 211: imo, this name is more confusing than it was before. With AM API we send headers partially and user header goes to the last fragment. Also, previously we avoided using bit fields, because compilers tend to generate more instructions with them and these fields are accessed on the fast path

iyastreb: As we checked together, user header still goes into the first fragment
I agree that the bit field may require some extra cycles, that's why I didn't have it initially.
Then let's go back to one extra field `uint8_t internal_flags`.
The size of request should not grow, because `rndv` part of the union is significantly larger anyway
@yosefe ?

brminich: yes, i was wrong about user header going to the last fragment. It goes to the end of the first fragment

yosefe: if we want to avoid bitmap, we can define internal AM flag that goes into "flags" field, and does not collide with the API flags

iyastreb: ok, then I add back the internal enum, with new internal flag using bit 15:
```C
enum ucp_request_am_internal_flags {
    UCP_REQUEST_AM_FLAG_HEADER_SENT = UCS_BIT(15)
};
```
Or we could add one last element to `enum ucp_send_am_flags {`
```C
enum ucp_send_am_flags {
    UCP_AM_SEND_FLAG_REPLY       = UCS_BIT(0),
    UCP_AM_SEND_FLAG_EAGER       = UCS_BIT(1),
    UCP_AM_SEND_FLAG_RNDV        = UCS_BIT(2),
    UCP_AM_SEND_FLAG_COPY_HEADER = UCS_BIT(3),
    UCP_AM_SEND_FLAG_LAST  << And use this in the internal enum
```

src/ucp/proto/proto_am.inl
```diff
@@ -22,6 +22,15 @@
 #define UCP_STATUS_PENDING_SWITCH (UCS_ERR_LAST - 1)
 
 
+/**
+ * AM specific internal request flags
+ * These flags extend @ref ucp_send_am_flags enumeration
+ */
+enum ucp_request_am_internal_flags {
+    UCP_REQUEST_AM_FLAG_HEADER_SENT = UCS_BIT(15)
```
brminich, line 30: we have to make sure it is not packed to the AM header

src/ucp/core/ucp_request.h
iyastreb: reverted to separate `uint8_t internal_flags`

src/ucp/proto/proto_am.inl
iyastreb: reverted to separate `uint8_t internal_flags`

brminich (approved):

yosefe (approved):

PR #10264, opened 2024-10-30
UCP/PROTO/MULTI: Sort lanes by BW

iyastreb:
## What
While working on [RM#4134043](https://redmine.mellanox.com/issues/4134043), we discovered that lanes for multi-frag protocols are not sorted by bandwidth. It's desirable to sort them in order to use lanes with higher bandwidth first

## Why ?
As discussed with UCX team, in multi-frag protocols it's desirable to have lanes sorted by bandwidth in order to optimize sending. First we try to send message over lane with highest bandwidth, and therefore potentially avoid splitting a message into several parts.

## How ?
Sort lanes by BW after picking them and detecting BW

### DIFF
--- a/src/ucp/proto/proto_multi.c
+++ b/src/ucp/proto/proto_multi.c
@@ -16,9 +16,22 @@
 #include "proto_debug.h"
 #include "proto_multi.inl"
 
+#include <ucs/algorithm/qsort_r.h>
 #include <ucs/debug/assert.h>
 #include <ucs/debug/log.h>
 
+
+static int
+ucp_proto_multi_compare_tl_perf(const void *elem1, const void *elem2, void *arg)
+{
+    const ucp_lane_index_t *lane1                = elem1;
+    const ucp_lane_index_t *lane2                = elem2;
+    const ucp_proto_common_tl_perf_t *lanes_perf = arg;
+
+    /* If bandwidths are equal, prefer to maintain the original ordering */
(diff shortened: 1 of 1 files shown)


src/ucp/proto/proto_multi.c
```diff
@@ -86,6 +99,15 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
         max_bandwidth = ucs_max(max_bandwidth, lane_perf->bandwidth);
     }
 
+    /* Sort lanes by bandwidth */
+    if (num_lanes > 1) {
+        /* If type of the first lane is not the same as in the middle, then
+         * exclude this lane from sorting */
```
brminich, line 105: this comments just repeats what the code does. Can you pls descrive the reason for that? E.g. that when the first lane is different is should always be used first e.g. to preserve ordering with eager protocols for TAG and AM APIs.

src/ucp/proto/proto_multi.c
```diff
@@ -86,6 +99,15 @@ ucs_status_t ucp_proto_multi_init(const ucp_proto_multi_init_params_t *params,
...
+        lane = (params->first.lane_type == params->middle.lane_type) ? 0 : 1;
```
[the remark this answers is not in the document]
iyastreb: sure

src/ucp/proto/proto_multi.c
iyastreb: ok

yosefe (approved):

brminich (approved):

PR #10267, opened 2024-10-31
UCP/CORE: Shuffle worker->all_eps creation

michal-shalev:
## What?
This PR adds a new configuration option, SHUFFLE_WORKER_EPS, to randomize the insertion order of 
EPs within the `worker->all_eps` list. When enabled, each new EP is placed at a random position instead of sequentially at the end of the list.

## Why?
Recently, a customer raised concerns about incast behavior during SHMEM quiet operations, where many EPs communicating in a fixed order could lead to network congestion and performance degradation. They believe that randomizing the order of EPs could reduce the likelihood of incast collisions.
The customer is interested in testing a patch to address this.

## How?
A new config parameter, `SHUFFLE_WORKER_EPS`, is added to `ucp_context_config_t`. The ucp_ep_create_base function uses this flag to insert EPs randomly within `worker->all_eps` via the u`cs_list_insert_at_index helper`.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -376,6 +376,10 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "another thread, or incoming active messages, but consumes more resources.",
    ucs_offsetof(ucp_context_config_t, flush_worker_eps), UCS_CONFIG_TYPE_BOOL},
 
+  {"SHUFFLE_WORKER_EPS", "n",
+   "Enable shuffling endpoints while creating a workers",
+   ucs_offsetof(ucp_context_config_t, shuffle_worker_eps), UCS_CONFIG_TYPE_BOOL},
+
   {"FENCE_MODE", "auto",
    "Fence mode used in ucp_worker_fence routine.\n"
    " weak   - use weak fence mode.\n"
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -377,7 +377,7 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, flush_worker_eps), UCS_CONFIG_TYPE_BOOL},
 
   {"SHUFFLE_WORKER_EPS", "n",
-   "Enable shuffling endpoints while creating a workers",
+   "Shuffle worker EPs while creating them",
    ucs_offsetof(ucp_context_config_t, shuffle_worker_eps), UCS_CONFIG_TYPE_BOOL},
 
   {"FENCE_MODE", "auto",
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -131,7 +131,7 @@ typedef struct ucp_context_config {
     size_t                                 estimated_num_ppn;
     /** Enable flushing endpoints while flushing a worker */
     int                                    flush_worker_eps;
-    /** Enable shuffling endpoints while creating a worker */
+    /** Shuffle worker EPs while creating them */
     int                                    shuffle_worker_eps;
     /** Fence mode */
     ucp_fence_mode_t                       fence_mode;
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -131,6 +131,8 @@ typedef struct ucp_context_config {
     size_t                                 estimated_num_ppn;
     /** Enable flushing endpoints while flushing a worker */
     int                                    flush_worker_eps;
+    /** Enable shuffling endpoints while creating a worker */
+    int                                    shuffle_worker_eps;
     /** Fence mode */
     ucp_fence_mode_t                       fence_mode;
     /** Enable optimizations suitable for homogeneous systems */
--- a/src/ucp/core/ucp_ep.c
+++ b/src/ucp/core/ucp_ep.c
@@ -361,6 +361,7 @@ ucs_status_t ucp_ep_create_base(ucp_worker_h worker, unsigned ep_init_flags,
 {
     ucs_status_t status;
     ucp_ep_h ep;
+    int new_ep_index;
 
     ep = ucp_ep_allocate(worker, peer_name);
     if (ep == NULL) {
@@ -394,7 +395,13 @@ ucs_status_t ucp_ep_create_base(ucp_worker_h worker, unsigned ep_init_flags,
         ucp_ep_update_flags(ep, UCP_EP_FLAG_INTERNAL, 0);
         ucs_list_add_tail(&worker->internal_eps, &ep->ext->ep_list);
     } else {
-        ucs_list_add_tail(&worker->all_eps, &ep->ext->ep_list);
+        if (worker->context->config.ext.shuffle_worker_eps) {
+            new_ep_index = ucs_rand() % (worker->num_all_eps + 1);
+            ucs_list_insert_at_index(&worker->all_eps, &ep->ext->ep_list,
+                                    new_ep_index);
+        } else {
+            ucs_list_add_tail(&worker->all_eps, &ep->ext->ep_list);
+        }
         ucs_assert(ep->worker->num_all_eps < UINT_MAX);
         ++ep->worker->num_all_eps;
     }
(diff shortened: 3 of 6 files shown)


src/ucs/datastruct/list.h
```diff
@@ -98,6 +98,28 @@ static inline void ucs_list_insert_before(ucs_list_link_t *pos,
...
+static inline void ucs_list_insert_at_index(ucs_list_link_t *head,
+                                            ucs_list_link_t *new_link,
+                                            unsigned index)
+{
+    ucs_list_link_t *pos;
+    unsigned i;
+
+    pos = head;
+    for (i = 0; i < index; ++i) {
+        pos = pos->next;
```
iyastreb, line 117: maybe add `ucs_assert(pos->next != NULL)`

src/ucs/datastruct/list.h
```diff
@@ -98,6 +98,28 @@ static inline void ucs_list_insert_before(ucs_list_link_t *pos,
...
 }
 
+/**
+ * Insert an item to a list at a specific index.
+ *
+ * @param head        List head.
+ * @param new_link    Item to insert.
+ * @param index       Index at which to insert.
+ */
+static inline void ucs_list_insert_at_index(ucs_list_link_t *head,
```
iyastreb, line 108: I think it would make sense to add a few unit tests for this function covering corner cases: insert first, last, empty list etc

src/ucp/core/ucp_ep.c
```diff
@@ -361,6 +361,7 @@ ucs_status_t ucp_ep_create_base(ucp_worker_h worker, unsigned ep_init_flags,
 {
     ucs_status_t status;
     ucp_ep_h ep;
+    int new_ep_index;
```
iyastreb, line 364: unsigned?

src/uct/ib/base/ib_md.c
```diff
@@ -224,6 +224,11 @@ typedef struct {
     struct ibv_mr                 **mrs;
 } uct_ib_md_mem_reg_thread_t;
 
+typedef struct {
+    uct_ib_md_mem_reg_thread_t ctx;
+    int                        mr_count;
```
iyastreb, line 229: maybe unsigned?

src/uct/ib/base/ib_md.c
```diff
@@ -282,13 +287,22 @@ uct_ib_md_print_mem_reg_err_msg(const char *title, void *address, size_t length,
     ucs_log(level, "%s", ucs_string_buffer_cstr(&msg));
 }
 
+static void uct_ib_dereg_mrs(struct ibv_mr **mrs, int count)
```
iyastreb, line 290: always inline?

src/uct/ib/base/ib_md.c
```diff
@@ -315,13 +329,13 @@ void *uct_ib_md_mem_handle_thread_func(void *arg)
               (ctx->params != NULL) ? "reg_mr" : "dereg_mr",
               start, ctx->address, ctx->first_mr_size,
               ucs_time_to_usec(ucs_get_time() - t0));
-    return UCS_STATUS_PTR(UCS_OK);
+    return (void*)(intptr_t)mr_idx;
```
iyastreb, line 332: returning either count or status seems error-prone
Maybe thread can just update it's `mr_count` inside?
And always return status

src/ucp/core/ucp_context.c
```diff
@@ -376,6 +376,10 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "another thread, or incoming active messages, but consumes more resources.",
    ucs_offsetof(ucp_context_config_t, flush_worker_eps), UCS_CONFIG_TYPE_BOOL},
 
+  {"SHUFFLE_WORKER_EPS", "n",
+   "Enable shuffling endpoints while creating a workers",
```
evgeny-leksikov, line 380: while creating EPs actually

src/ucp/core/ucp_ep.c
evgeny-leksikov: size_t

src/ucs/datastruct/list.h
```diff
@@ -98,6 +98,28 @@ static inline void ucs_list_insert_before(ucs_list_link_t *pos,
...
+                                            ucs_list_link_t *new_link,
+                                            unsigned index)
```
evgeny-leksikov, line 110: size_t

src/uct/ib/base/ib_md.c
evgeny-leksikov: I dont know how is it omportant here, but usually we use ssize_t for such purpose, can see bcopy API

michal-shalev: this is unrelate to my PR, bad rebase, pushed a revert commit @evgeny-leksikov @iyastreb

src/ucp/core/ucp_context.c
gleon99: Maybe `Randomize order of newly created endpoints in the internal list` instead.
@yosefe wdyt?

src/ucp/core/ucp_context.h
```diff
@@ -131,6 +131,8 @@ typedef struct ucp_context_config {
     size_t                                 estimated_num_ppn;
     /** Enable flushing endpoints while flushing a worker */
     int                                    flush_worker_eps;
+    /** Enable shuffling endpoints while creating a worker */
```
gleon99, line 134: w/o `enable`, this one doesn't "enable".

src/ucp/core/ucp_ep.c
```diff
@@ -394,7 +395,13 @@ ucs_status_t ucp_ep_create_base(ucp_worker_h worker, unsigned ep_init_flags,
         ucp_ep_update_flags(ep, UCP_EP_FLAG_INTERNAL, 0);
         ucs_list_add_tail(&worker->internal_eps, &ep->ext->ep_list);
     } else {
-        ucs_list_add_tail(&worker->all_eps, &ep->ext->ep_list);
+        if (worker->context->config.ext.shuffle_worker_eps) {
+            new_ep_index = ucs_rand() % (worker->num_all_eps + 1);
+            ucs_list_insert_at_index(&worker->all_eps, &ep->ext->ep_list,
+                                    new_ep_index);
```
yosefe, line 401: this will cause O(n^2) initialization time
better solution approach could be to shuffle the list using fisher yates method every N flush operations, and only if new EPs were added since the previous shuffle. .

gleon99: Got a simpler one. Insert each time at head/tail with 50% probability. Will lead to random result, with zero cost.
wdyt?

src/uct/ib/base/ib_md.c
michal-shalev: this is unrelate to my PR, bad rebase, pushed a revert commit @iyastreb

src/uct/ib/base/ib_md.c
michal-shalev: this is unrelate to my PR, bad rebase, pushed a revert commit @iyastreb

src/ucp/core/ucp_context.c
michal-shalev: done

src/ucp/core/ucp_context.h
michal-shalev: done

src/ucp/core/ucp_ep.c
michal-shalev: Regarding the coin-flip method for insertion, I looked at the standard deviation of the positions, and it increases as the number of the element gets higher. 
Numbers that are inserted early (like 1 and 2) tend to cluster more around the middle of the list, while numbers that are inserted later (like 9 and 10) are more widely dispersed, resulting in a higher standard deviation. 
The distribution of the resulting list isn’t completely uniform, so I guess it depends on how much randomness the customer requires.
Please let me know how to proceed.
@gleon99 @yosefe 
image
image

michal-shalev:
Closing, the solution would be in ompi.
@gleon99 @yosefe @evgeny-leksikov @iyastreb

michal-shalev:
Closing, the solution would be in ompi.
@gleon99 @yosefe @evgeny-leksikov @iyastreb

PR #10270, opened 2024-10-31
UCT/IB/MLX5/DC: Change DCIs array to pointers - dynamic array grow

roiedanino:
## What
- Access DCIs by index instead of by pointer
- in poll_tx: relocate dcis array so asan would catch if memory violation is done
- Set default dcis array capacity to 1.

## Why ?
Suggesting a future proof fix for the issues caused by dynamic resizing of dcis array (use after free caused by using an old pointer after resize).

### DIFF
--- a/src/uct/ib/mlx5/dc/dc_mlx5.c
+++ b/src/uct/ib/mlx5/dc/dc_mlx5.c
@@ -135,10 +135,10 @@ ucs_config_field_t uct_dc_mlx5_iface_config_sub_table[] = {
      ucs_offsetof(uct_dc_mlx5_iface_config_t, num_dci_channels),
      UCS_CONFIG_TYPE_UINT},
 
-    {"DCIS_INITIAL_CAPACITY", "inf",
+    {"DCIS_INITIAL_CAPACITY", "auto",
      "Initial capacity of DCIs array.",
      ucs_offsetof(uct_dc_mlx5_iface_config_t, dcis_initial_capacity),
-     UCS_CONFIG_TYPE_UINT},
+     UCS_CONFIG_TYPE_ULUNITS},
 
     {NULL}
 };
@@ -278,7 +278,6 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface, int poll_flags)
     struct mlx5_cqe64 *cqe;
     uint16_t hw_ci;
     uct_dc_dci_t *dci;
-    UCT_DC_MLX5_TXQP_DECL(txqp, txwq);
 
     cqe = uct_ib_mlx5_poll_cq(&iface->super.super.super,
                               &iface->super.cq[UCT_IB_DIR_TX], poll_flags,
@@ -293,16 +292,13 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface, int poll_flags)
 
     dci_index = uct_dc_mlx5_iface_dci_find(iface, cqe);
     dci       = uct_dc_mlx5_iface_dci(iface, dci_index);
-    ucs_assert(uct_dc_mlx5_is_dci_valid(dci));
-    txqp      = &dci->txqp;
-    txwq      = &dci->txwq;
     hw_ci     = ntohs(cqe->wqe_counter);
 
-    ucs_trace_poll("dc iface %p tx_cqe: dci[%d] txqp %p hw_ci %d",
-                   iface, dci_index, txqp, hw_ci);
+    ucs_trace_poll("dc iface %p tx_cqe: dci[%d] txqp %p hw_ci %d", iface,
+                   dci_index, &dci->txqp, hw_ci);
 
-    uct_rc_mlx5_txqp_process_tx_cqe(txqp, cqe, hw_ci);
-    uct_dc_mlx5_update_tx_res(iface, txwq, txqp, hw_ci);
+    uct_rc_mlx5_txqp_process_tx_cqe(&dci->txqp, cqe, hw_ci);
+    uct_dc_mlx5_update_tx_res(iface, dci_index, hw_ci);
 
     /**
      * Note: DCI is released after handling completion callbacks,
@@ -377,8 +373,9 @@ ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
     uct_ib_iface_t *ib_iface   = &iface->super.super.super;
     uct_ib_mlx5_qp_attr_t attr = {};
     uct_ib_mlx5_md_t *md = ucs_derived_of(ib_iface->super.md, uct_ib_mlx5_md_t);
-    uct_dc_dci_t *dci    = uct_dc_mlx5_iface_dci(iface, dci_index);
+    uct_dc_dci_t *dci    = ucs_calloc(1, sizeof(uct_dc_dci_t), "dci");
     ucs_status_t status;
+
 #if HAVE_DC_DV
     uct_ib_device_t *dev               = uct_ib_iface_device(ib_iface);
     struct mlx5dv_qp_init_attr dv_attr = {};
@@ -404,7 +401,7 @@ ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
                                             &dci->txwq.super, &dci->txwq,
                                             &attr);
         if (status != UCS_OK) {
-            return status;
+            goto err_free_dci;
         }
 
         ucs_debug("created DevX DCI 0x%x, rdma_wr_disabled=%d", dci->txwq.super.qp_num,
@@ -414,12 +411,13 @@ ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
 
     if (iface->super.cq[UCT_IB_DIR_TX].type != UCT_IB_MLX5_OBJ_TYPE_VERBS) {
         ucs_error("cannot create verbs DCI with DEVX CQ");
-        return UCS_ERR_INVALID_PARAM;
+        status = UCS_ERR_INVALID_PARAM;
+        goto err_free_dci;
     }
 
     status = uct_ib_mlx5_iface_get_res_domain(ib_iface, &dci->txwq.super);
     if (status != UCS_OK) {
-        return status;
+        goto err_free_dci;
     }
 
     uct_ib_mlx5_iface_fill_attr(ib_iface, &dci->txwq.super, &attr);
@@ -439,8 +437,7 @@ ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
     }
 
     dci->txwq.super.verbs.qp = qp;
-    dci->txwq.super.qp_num = dci->txwq.super.verbs.qp->qp_num;
-
+    dci->txwq.super.qp_num   = dci->txwq.super.verbs.qp->qp_num;
 init_qp:
 #else
     uct_rc_mlx5_iface_fill_attr(&iface->super, &attr,
@@ -448,7 +445,7 @@ init_qp:
                                 &iface->super.rx.srq);
     status = uct_ib_mlx5_iface_create_qp(ib_iface, &dci->txwq.super, &attr);
     if (status != UCS_OK) {
-        return status;
+        goto err_free_dci;
     }
 #endif
     status = uct_rc_txqp_init(&dci->txqp, &iface->super.super,
@@ -457,6 +454,11 @@ init_qp:
     if (status != UCS_OK) {
         goto err_qp;
     }
+
+    ucs_assertv_always(!uct_dc_mlx5_is_dci_valid(
+                               uct_dc_mlx5_iface_dci(iface, dci_index)),
+                       "iface=%p dci_index=%d", iface, dci_index);
+    ucs_array_elem(&iface->tx.dcis, dci_index) = dci;
 
     if (connect) {
         status = uct_dc_mlx5_iface_dci_connect(iface, dci);
@@ -494,6 +496,8 @@ err_put_res_domain:
         uct_ib_mlx5_iface_put_res_domain(&dci->txwq.super);
     }
 #endif
+err_free_dci:
+    ucs_free(dci);
     return status;
 }
 
@@ -763,10 +767,11 @@ static void uct_dc_mlx5_iface_dci_pool_destroy(uct_dc_mlx5_dci_pool_t *dci_pool)
 }
 
 static void
-uct_dc_mlx5_destroy_dci(uct_dc_mlx5_iface_t *iface, uct_dc_dci_t *dci)
+uct_dc_mlx5_destroy_dci(uct_dc_mlx5_iface_t *iface, uint16_t dci_index)
 {
     uct_ib_mlx5_md_t *md = ucs_derived_of(iface->super.super.super.super.md,
                                           uct_ib_mlx5_md_t);
+    uct_dc_dci_t *dci    = uct_dc_mlx5_iface_dci(iface, dci_index);
 
     uct_rc_txqp_cleanup(&iface->super.super, &dci->txqp);
     uct_ib_mlx5_destroy_qp(md, &dci->txwq.super);
@@ -776,7 +781,8 @@ uct_dc_mlx5_destroy_dci(uct_dc_mlx5_iface_t *iface, uct_dc_dci_t *dci)
     }
     uct_ib_mlx5_qp_mmio_cleanup(&dci->txwq.super, dci->txwq.reg);
 
-    dci->txwq.super.qp_num = UCT_IB_INVALID_QPN;
+    ucs_free(dci);
+    ucs_array_elem(&iface->tx.dcis, dci_index) = NULL;
 }
 
 static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface)
@@ -792,7 +798,7 @@ static void uct_dc_mlx5_iface_dcis_destroy(uct_dc_mlx5_iface_t *iface)
             continue;
         }
 
-        uct_dc_mlx5_destroy_dci(iface, dci);
+        uct_dc_mlx5_destroy_dci(iface, dci_index);
     }
 
     for (pool_index = 0; pool_index < iface->tx.num_dci_pools; pool_index++) {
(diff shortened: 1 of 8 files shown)


src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -267,14 +267,27 @@ static void uct_dc_mlx5_iface_progress_enable(uct_iface_h tl_iface, unsigned fla
...
+static void uct_dc_mlx5_asan_relocate_dcis_array(uct_dc_mlx5_iface_t *iface)
+{
+#if __SANITIZE_ADDRESS__
+    size_t buffer_size = sizeof(uct_dc_dci_t) * ucs_array_capacity(&iface->tx.dcis);
+    size_t num_dcis = ucs_array_length(&iface->tx.dcis);
+    void *old_buffer_p;
+
+    ucs_debug_asan_relocate_array_buffer((void**)&iface->tx.dcis.buffer,
+                                         &old_buffer_p, buffer_size);
+    uct_dc_mlx5_iface_dcis_array_copy(iface->tx.dcis.buffer, old_buffer_p,
```
tvegas1, line 279: the memcpy is already done inside relocate array buffer function?

src/uct/ib/rc/base/rc_ep.h
```diff
@@ -441,31 +470,6 @@ uct_rc_txqp_completion_op(uct_rc_iface_send_op_t *op, const void *resp)
     op->handler(op, resp);
 }
 
-static UCS_F_ALWAYS_INLINE void
```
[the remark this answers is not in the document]
roiedanino: Yes because we want to access the txqp by dci index inside the loop to avoid use-after-free in case the dcis array resize during one of the completions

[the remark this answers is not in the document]
tvegas1: but we could do that while still keeping them as inline functions?

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -267,14 +267,27 @@ static void uct_dc_mlx5_iface_progress_enable(uct_iface_h tl_iface, unsigned fla
...
+                                      num_dcis);
+    ucs_array_buffer_free(old_buffer_p);
+#endif
+}
+
 static UCS_F_ALWAYS_INLINE unsigned
 uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface, int poll_flags)
```
iyastreb, line 286: Is it enough to fix only code in uct_dc_mlx5_poll_tx?
I see that `uct_dc_mlx5_iface_resize_and_fill_dcis` may be called directly/indirectly from:
- init functions, hopefully ok: uct_dc_mlx5_ep_basic_init, uct_dc_mlx5_ep_basic_init, uct_dc_mlx5_iface_init_dcis_array
- uct_dc_mlx5_iface_dci_can_alloc_or_create -> uct_dc_mlx5_iface_dci_get, which is called from many places:
1. UCT_DC_MLX5_CHECK_DCI_RES
uct_dc_mlx5_ep_fc_pure_grant_send
uct_dc_mlx5_ep_fc_hard_req_send
UCT_DC_MLX5_CHECK_RES - MANY references!
3. UCT_DC_CHECK_RES_PTR

Basically this function may be called indirectly from dozens of different flows, and it's hard to track all of them.. And each workflow caches DCI internals (txqp, txwq) into local variables, so I'm not sure we can easily guarantee that it always works reliably. Even if we fix it now, it is very error-prone and it's quite hard to find the root cause

One option to make it 100% safe is to allocate TX structures on heap:
```C
typedef struct uct_dc_dci {
    uct_rc_txqp_t                *txqp; /* DCI qp */
    uct_ib_mlx5_txwq_t           *txwq; /* DCI mlx5 wq */
```
By doing this all the other problems are solved:
1. DCI object becomes "primitive" in the sense we don't even need advanced copy logic.
2. It's safe to cache those fields from any workflow, no need to access by index
I think it has almost zero overhead in terms of performance

src/uct/ib/rc/base/rc_ep.h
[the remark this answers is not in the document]
roiedanino: Not without code duplication

src/uct/ib/mlx5/dc/dc_mlx5.c
gleon99: @iyastreb it is indeed error-prone, though your suggestion would introduce a performance penalty on fast path.

src/ucs/debug/debug.c
```diff
@@ -1341,3 +1341,21 @@ void ucs_debug_asan_validate_address(const char *ptr_name, void *address,
     }
 #endif
 }
+
+/**
+ * Relocate an array buffer to a new location, for ASAN debugging,
+ * old_arr_p has to be freed by the caller.
+ */
+void ucs_debug_asan_relocate_array_buffer(void **arr, void **old_arr_p,
```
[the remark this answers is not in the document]
roiedanino: Maybe it should all be done in dc as `uct_dc_mlx5_iface_dcis_array_copy` is already copying the array and splicing the queues

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -267,14 +270,41 @@ static void uct_dc_mlx5_iface_progress_enable(uct_iface_h tl_iface, unsigned fla
...
+#endif
+}
+
+void uct_dc_mlx5_asan_relocate_dcis_array(uct_dc_mlx5_iface_t *iface)
+{
+#ifdef __SANITIZE_ADDRESS__
+    size_t buffer_size = sizeof(uct_dc_dci_t) *
+                         ucs_array_capacity(&iface->tx.dcis);
+    size_t num_dcis    = ucs_array_length(&iface->tx.dcis);
+    void *old_buffer_p = iface->tx.dcis.buffer;
```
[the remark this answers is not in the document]
tvegas1: or could this whole malloc+copy procedure be bundled to some ucs_array_ specific macro, returning the old_buffer?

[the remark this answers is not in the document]
roiedanino: Yes we have `ucs_array_grow` but it checks whether we need to increase capacity and if not it won't allocate a new buffer

src/uct/ib/rc/base/rc_ep.h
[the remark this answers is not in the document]
tvegas1: thanks now I see it. `ucs_queue_for_each_extract` makes 3 lookups of txwq per iteration, and log could be misleading, is passing a callback / putting `for each` loop outside or having `outstanding` at constant memory position not feasible for perf?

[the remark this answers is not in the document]
tvegas1: maybe use a callback to fetch/dequeue next queue element would not have perf impact

src/uct/ib/mlx5/dc/dc_mlx5.h
```diff
@@ -61,6 +61,13 @@ struct ibv_ravh {
         (_self)->flags |= UCT_DC_MLX5_IFACE_FLAG_##_flag_name##_FULL_HANDSHAKE; \
     }
 
+#ifdef __SANITIZE_ADDRESS__
+#define UCT_DC_MLX5_ASAN_RELOCATE_DCIS_BUFFER(_iface) \
```
gleon99, line 65: Minor: the macro is `BUFFER`, while the func is `array`. Is that intentional?

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -267,14 +270,42 @@ static void uct_dc_mlx5_iface_progress_enable(uct_iface_h tl_iface, unsigned fla
...
+#endif
+}
+
+void uct_dc_mlx5_asan_relocate_dcis_array(uct_dc_mlx5_iface_t *iface)
+{
+#ifdef __SANITIZE_ADDRESS__
+    size_t buffer_size         = sizeof(uct_dc_dci_t) *
+                                 ucs_array_capacity(&iface->tx.dcis);
+    size_t num_dcis            = ucs_array_length(&iface->tx.dcis);
+    uct_dc_dci_t *old_buffer_p = ucs_array_begin(&iface->tx.dcis);
```
gleon99, line 290: Seems both `old` and `new` are `buffer`, why the 1st with `_p` and the second w/o?

src/uct/ib/mlx5/dc/dc_mlx5.c
iyastreb: @roiedanino Any comments on that?

gleon99: After reconsidering that -
@roiedanino , let's go with Ilia's suggestion.

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -377,13 +378,17 @@ ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
     uct_ib_iface_t *ib_iface   = &iface->super.super.super;
     uct_ib_mlx5_qp_attr_t attr = {};
     uct_ib_mlx5_md_t *md = ucs_derived_of(ib_iface->super.md, uct_ib_mlx5_md_t);
-    uct_dc_dci_t *dci    = uct_dc_mlx5_iface_dci(iface, dci_index);
+    uct_dc_dci_t *dci    = ucs_calloc(1, sizeof(uct_dc_dci_t), "dci");
```
yosefe, line 381: why need to calloc and not malloc?

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -377,13 +378,17 @@ ucs_status_t uct_dc_mlx5_iface_create_dci(uct_dc_mlx5_iface_t *iface,
...
     ucs_status_t status;
+
 #if HAVE_DC_DV
     uct_ib_device_t *dev               = uct_ib_iface_device(ib_iface);
     struct mlx5dv_qp_init_attr dv_attr = {};
     struct ibv_qp *qp;
 
+    ucs_assertv_always(!uct_dc_mlx5_is_dci_valid(
+                               uct_dc_mlx5_iface_dci(iface, dci_index)),
+                       "dci_index=%d iface=%p", dci_index, iface);
```
yosefe, line 391: 1. "iface=%p dci_index=%d"
2. move this right before `ucs_array_elem(&iface->tx.dcis, dci_index) = dci;`

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -763,7 +772,7 @@ static void uct_dc_mlx5_iface_dci_pool_destroy(uct_dc_mlx5_dci_pool_t *dci_pool)
 }
 
 static void
-uct_dc_mlx5_destroy_dci(uct_dc_mlx5_iface_t *iface, uct_dc_dci_t *dci)
+uct_dc_mlx5_destroy_dci(uct_dc_mlx5_iface_t *iface, uct_dc_dci_t *dci, uint16_t dci_index)
```
yosefe, line 775: do we really have to pass both dci and dci_index?

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -1597,9 +1598,20 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
...
+            self->tx.dcis_initial_capacity = self->tx.ndci *
+                                             UCT_DC_MLX5_IFACE_MAX_DCI_POOLS;
+        } else {
+            self->tx.dcis_initial_capacity = 1;
+        }
+    } else {
+        self->tx.dcis_initial_capacity =
+                ucs_min(config->dcis_initial_capacity,
+                        self->tx.ndci * UCT_DC_MLX5_IFACE_MAX_DCI_POOLS);
+    }
```
yosefe, line 1613: maybe something like
```c
if (config->dcis_initial_capacity == UCS_ULUNITS_AUTO) {
    max_capacity = uct_dc_mlx5_iface_is_dci_rand(self) ? SIZE_MAX : 1;
} else {
    max_capacity = config->dcis_initial_capacity;
}
self->tx.dcis_initial_capacity =
         ucs_min(max_capacity,
                 self->tx.ndci * UCT_DC_MLX5_IFACE_MAX_DCI_POOLS);

```

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -1750,6 +1762,10 @@ static UCS_CLASS_INIT_FUNC(uct_dc_mlx5_iface_t, uct_md_h tl_md, uct_worker_h wor
 
     uct_rc_mlx5_iface_common_prepost_recvs(&self->super);
 
+#ifdef __SANITIZE_ADDRESS__
+    self->old_dcis_buffer = NULL;
+#endif
```
yosefe, line 1767: still needed?

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -23,6 +23,9 @@
 #include <ucs/async/async.h>
 #include <ucs/debug/log.h>
 #include <string.h>
+#ifdef __SANITIZE_ADDRESS__
+#  include <sanitizer/asan_interface.h>
+#endif
```
yosefe, line 28: still needed?

src/uct/ib/mlx5/dc/dc_mlx5.h
```diff
@@ -186,7 +186,7 @@ typedef struct uct_dc_mlx5_iface_config {
     ucs_time_t                          fc_hard_req_timeout;
     uct_ud_mlx5_iface_common_config_t   mlx5_ud;
     unsigned                            num_dci_channels;
-    unsigned                            dcis_initial_capacity;
+    unsigned long                       dcis_initial_capacity;
```
yosefe, line 189: size_t

src/uct/ib/mlx5/dc/dc_mlx5.h
```diff
@@ -361,6 +361,11 @@ struct uct_dc_mlx5_iface {
     uint16_t                         flags;
 
     uct_ud_mlx5_iface_common_t       ud_common;
+
+#ifdef __SANITIZE_ADDRESS__
+    uct_dc_dci_t *                   old_dcis_buffer;
+    size_t                           old_dcis_buffer_size;
+#endif
```
yosefe, line 368: still needed?

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -292,24 +293,24 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface, int poll_flags)
     ucs_memory_cpu_load_fence();
 
     dci_index = uct_dc_mlx5_iface_dci_find(iface, cqe);
-    dci       = uct_dc_mlx5_iface_dci(iface, dci_index);
-    ucs_assert(uct_dc_mlx5_is_dci_valid(dci));
-    txqp      = &dci->txqp;
-    txwq      = &dci->txwq;
+    ucs_assert(uct_dc_mlx5_is_dci_valid(uct_dc_mlx5_iface_dci(iface, dci_index)));
```
yosefe, line 296: can we move this assertion into uct_dc_mlx5_iface_dci_find and convert to assertv?

src/uct/ib/mlx5/dc/dc_mlx5.c
```diff
@@ -292,24 +293,24 @@ uct_dc_mlx5_poll_tx(uct_dc_mlx5_iface_t *iface, int poll_flags)
...
-    ucs_trace_poll("dc iface %p tx_cqe: dci[%d] txqp %p hw_ci %d",
-                   iface, dci_index, txqp, hw_ci);
+    ucs_trace_poll("dc iface %p tx_cqe: dci[%d] txqp %p hw_ci %d", iface,
+                   dci_index, &uct_dc_mlx5_iface_dci(iface, dci_index)->txqp,
+                   hw_ci);
 
-    uct_rc_mlx5_txqp_process_tx_cqe(txqp, cqe, hw_ci);
-    uct_dc_mlx5_update_tx_res(iface, txwq, txqp, hw_ci);
+    uct_rc_mlx5_txqp_process_tx_cqe(
+            &uct_dc_mlx5_iface_dci(iface, dci_index)->txqp, cqe, hw_ci);
```
yosefe, line 304: can we restore the variable `uct_dc_dci_t *dci;` since it's used multiple times

src/uct/ib/mlx5/dc/dc_mlx5_ep.c
```diff
@@ -1743,7 +1743,7 @@ void uct_dc_mlx5_ep_handle_failure(uct_dc_mlx5_ep_t *ep,
 
     ucs_assert(!uct_dc_mlx5_iface_is_policy_shared(iface));
 
-    uct_dc_mlx5_update_tx_res(iface, txwq, txqp, pi);
+    uct_dc_mlx5_update_tx_res(iface, dci_index, pi);
```
yosefe, line 1746: can we remove txqp, txwq variables?
seems could also remove UCT_DC_MLX5_IFACE_TXQP_DCI_GET macro altogether

src/uct/ib/mlx5/dc/dc_mlx5_ep.h
```diff
@@ -346,7 +346,7 @@ uct_dc_mlx5_dci_pool_init_dci(uct_dc_mlx5_iface_t *iface, uint8_t pool_index,
                               uct_dci_index_t dci_index)
 {
     uct_dc_mlx5_dci_pool_t *pool = &iface->tx.dci_pool[pool_index];
-    uct_dc_dci_t *dci            = uct_dc_mlx5_iface_dci(iface, dci_index);
+    uct_dc_dci_t *dci;
```
yosefe, line 349: move after num_channels var

test/gtest/uct/ib/test_dc.cc
```diff
@@ -53,6 +53,15 @@ class test_dc : public test_rc {
         return ucs_derived_of(e->ep(idx), uct_dc_mlx5_ep_t);
     }
 
+    /* A utility for pending tests, shorten test time 
+       by reducing dci QP resources */
+    void set_first_dci_availability(uct_ep_h ep, int available = 8)
+    {
+        ucs_status_t status = uct_ep_am_short(ep, 0, 0, NULL, 0);
+        EXPECT_UCS_OK(status);
+        uct_dc_mlx5_iface_dci(dc_iface(m_e1), 0)->txqp.available = available;
```
iyastreb, line 62: should we assert here that `ep == uct_dc_mlx5_iface_dci(iface, ep->dci)->ep`?

src/uct/ib/mlx5/dc/dc_mlx5.c
roiedanino: To avoid garbage values issues

yosefe (dismissed):

yosefe:
Failures seem related

yosefe (approved):

Issue #186, opened 2015-03-27
Verbs code does not handle len 0 post


shamisp:
Earlier versions of OFED do not handle post zero. See #185

PR #10272, opened 2024-11-04
ARCH/X86: Add Turin support

arun-chandran-edarath:
Acked-by: Edgar Gabriel <edgar.gabriel@amd.com>

## What?
Adding support for AMD's Turin processors.

Ref: https://www.amd.com/en/products/processors/server/epyc/9005-series.html

### DIFF
--- a/config/ucx.conf
+++ b/config/ucx.conf
@@ -23,18 +23,15 @@ UCX_MM_SEND_OVERHEAD=am_short:40ns,am_bcopy:220ns
 UCX_MM_RECV_OVERHEAD=am_short:40ns,am_bcopy:220ns
 UCX_RCACHE_OVERHEAD=360ns
 
-[AMD Rome]
-CPU model=Rome
+[AMD]
+CPU vendor=AMD
+# UCX_DISTANCE_BW may affect GPU communication methods.
+# Tune it based on the actual PCIE bandwidth between the devices in your system
 UCX_DISTANCE_BW=auto,sys:5100MBs
 
 [AMD Milan]
 CPU model=Milan
-UCX_DISTANCE_BW=auto,sys:5100MBs
 # Real latencies are around 1.4 and 0.4, rest is gdrcopy rcache overhead
 # TODO: Add gdrcopy rcache overhead as separate performance graph node
 # TODO: Add rcache overhead not only for Milan and GH systems
 UCX_GDR_COPY_LAT=get:1.65e-6,put:0.65e-6
-
-[AMD Genoa]
-CPU model=Genoa
-UCX_DISTANCE_BW=auto,sys:5100MBs
--- a/src/ucs/arch/x86_64/cpu.c
+++ b/src/ucs/arch/x86_64/cpu.c
@@ -485,6 +485,13 @@ ucs_cpu_model_t ucs_arch_get_cpu_model()
                 break;
             }
             break;
+        /* AMD Zen5 */
+        case 0x1a:
+            if ((model <= 0x2f) || (model >= 0x40 && model <= 0x4f) ||
+                (model >= 0x60 && model <= 0x7f)) {
+                cpu_model = UCS_CPU_MODEL_AMD_TURIN;
+            }
+            break;
         }
     }
 
--- a/src/ucs/arch/cpu.c
+++ b/src/ucs/arch/cpu.c
@@ -2,6 +2,7 @@
 * Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2019. ALL RIGHTS RESERVED.
 * Copyright (C) Shanghai Zhaoxin Semiconductor Co., Ltd. 2020. ALL RIGHTS RESERVED.
 * Copyright (C) Tactical Computing Labs, LLC. 2022. ALL RIGHTS RESERVED.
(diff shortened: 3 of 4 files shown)


src/ucs/arch/x86_64/cpu.c
```diff
@@ -485,6 +485,16 @@ ucs_cpu_model_t ucs_arch_get_cpu_model()
                 break;
             }
             break;
+        /* AMD Zen5 */
+        case 0x1a:
+            switch (model) {
+            case 0x00 ... 0x2f:
+            case 0x40 ... 0x4f:
+            case 0x60 ... 0x7f:
```
yosefe, line 493: seems it's a non-standard extension? maybe use if condition?

arun-chandran-edarath: If I use the 'if condition' for 'model', it would make the code less structured. The 'case ranges' are a well-known extension supported by both GCC and Clang. Can we stick with it?

edgargabriel:
It would be good to have this patch btw. in 1.18.0 (if possible)

config/ucx.conf
```diff
@@ -35,3 +35,7 @@ UCX_GDR_COPY_LAT=get:1.65e-6,put:0.65e-6
 [AMD Genoa]
 CPU model=Genoa
 UCX_DISTANCE_BW=auto,sys:5100MBs
+
+[AMD Turin]
```
ivankochin, line 39: Is that possible to set `UCX_DISTANCE_BW` one time for all AMD CPU families? Something like:

```
[AMD]
CPU vendor=AMD
UCX_DISTANCE_BW=auto,sys:5100MBs
```

arun-chandran-edarath: Could you please explain more about  "UCX_DISTANCE_BW" ? Why we have to keep 'sys' at 5100MBs?[I was not sure about this, I used the same value that used in Rome, Milan & Genoa] Is this should be calculated at runtime?

ivankochin: That's a good question, that's pretty old setting that was measured long time ago on some AMD system and I am not sure whether we need to set it for Turin systems too. @yosefe WDYT? 

My previous comment was only suggesting to set it one time for all AMD system instead of patching each CPU generation separately, but probably if we don't want to set this value for this generation this change would redundant too.

arun-chandran-edarath: @yosefe could you please comment on "UCX_DISTANCE_BW" ?

yosefe: UCX_DISTANCE_BW is specifying the expected PCIe bandwidth as a function of distance as defined here: https://github.com/openucx/ucx/blob/master/src/ucs/config/global_opts.c#L78

src/ucs/arch/x86_64/cpu.c
yosefe: i'd prefer if/else

arun-chandran-edarath: I will change the style then.

config/ucx.conf
yosefe: 1. it is representing the expected PCIe badnwidth in case the devices are on different NUMA nodes
2. If not provided a default value is used. we can define the config setting as AUTO, and the code will calculate the BW if the value is auto.
3. it's between devices on different NUMA node, traversing system interconnect (such as QPI). For example GPU and RDMA NIC.

yosefe: Currently UCX cannot determine the actual ucs_global_opts.dist.sys.bandwidth value, the default is [220 MB/s](https://github.com/openucx/ucx/blob/c4acc6c3f/src/ucs/sys/topo/base/topo.c#L439) which is probably not right value for AMD.

arun-chandran-edarath: Can it be derrived reading /proc or /sys entries ? How shall we proceed further here?

yosefe: > Can it be derrived reading /proc or /sys entries ? How shall we proceed further here?

This is a question for AMD, I guess. I'm not aware of such way, but would be great to have.
I'd suggest for now to define single BW value for all AMD CPU's , as suggest by @ivankochin

arun-chandran-edarath: @yosefe @ivankochin Apologies for my ignorance, but I want to ensure I understand this correctly before proceeding. With 2 Turin-only nodes connected via IB, are there benchmark tests available to measure the impact of varying "UCX_DISTANCE_BW=auto,sys:_values_from_low_to_high_" values? Or Does this only apply to GPU-to-RDMA transfers in a GPU cluster?

yosefe: This setting may affect the threshold switching of GPU protocols, so probably benchmarks like "osu_bw D D" could uncover  the difference on some systems.

yosefe (approved):

arun-chandran-edarath:
Squashed.

yosefe (approved):

arun-chandran-edarath:
It seems some of the build is failing with below error.

"/__w/1/s/contrib/../src/ucs/arch/x86_64/cpu.c", line 490: error #186: pointless comparison of unsigned integer with zero
              if ((model >= 0x00 && model <= 0x2f) ||
                         ^
                      
   Will push a new one.

tvegas1 (approved):

PR #10275, opened 2024-11-05
UCP/RNDV: Remove RNDV_PERF_DIFF setting

ivankochin:
## What?
Get rid of control handle that allows to improve RNDV performance from the user side.

## Why?
That control allows to improve performance to change RNDV protocol selection/threshold automatic configuration. FMPOV that is not a good solution since it is not clear for customer how changing protocol performance to some value will affect protocol selection for different cases. User still would be able to use `RNVD_THRESH` for threshold manual setting.

Also after reworking of multi performance prediction this control has no effect for protov2 so probably we should completely remove it.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -191,11 +191,6 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "is zero or negative",
    ucs_offsetof(ucp_context_config_t, rndv_thresh_fallback), UCS_CONFIG_TYPE_MEMUNITS},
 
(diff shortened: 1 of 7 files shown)


yosefe:
we wanted to prefer RNDV protocol after some message size, even if perf is same as eager, to reduce unexpected buffer memory consumption on the receiver side. Is there another way to achieve it?

yosefe:
Per offline discussion with @ivankochin , we want to restore the RNDV_PERF_DIFF login in protov2

ivankochin:
It was

PR #10277, opened 2024-11-05
UCP/PROTO: Consider FAST_CMPL flag only for eager protocols

ivankochin:
## What?
Remove considering FAST_CMPL flag for RNDV protocols

## Why?
FAST_CMPL shouldn't affect RNDV protocols since they can be completed only when all operations are done.

### DIFF
--- a/src/ucp/rndv/proto_rndv.c
+++ b/src/ucp/rndv/proto_rndv.c
@@ -261,6 +261,8 @@ static ucp_proto_select_param_t ucp_proto_rndv_remote_select_param_init(
     ucp_memory_info_t mem_info;
     uint32_t op_attr_mask;
 
+    /* FAST_CMPL flag shouldn't be considered on remote side since it
+     * affects only one stage protocols (like eager) */
     op_attr_mask = ucp_proto_select_op_attr_unpack(select_param->op_attr) &
                    UCP_OP_ATTR_FLAG_MULTI_SEND;
     /* Construct select parameter for the remote protocol */
@@ -617,6 +619,18 @@ ucp_proto_rndv_ack_init(const ucp_proto_common_init_params_t *init_params,
                                     UCS_LINEAR_FUNC_ZERO, perf_p);
 }
 
+static void
+ucp_proto_rndv_report_fast_cmpl(const ucp_proto_multi_init_params_t *params)
+{
+    uint32_t op_attr_mask = ucp_proto_select_op_attr_unpack(
+            params->super.super.select_param->op_attr);
+
+    if ((op_attr_mask & UCP_OP_ATTR_FLAG_FAST_CMPL)) {
+        ucs_warn("RNDV is initialized with FAST_CMPL which is unexpected. "
+                 "Check whether receive operation was called with that flag.");
+    }
+}
+
 ucs_status_t
 ucp_proto_rndv_bulk_init(const ucp_proto_multi_init_params_t *init_params,
                          const char *op_name, const char *ack_name,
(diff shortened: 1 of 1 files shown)


src/ucp/rndv/proto_rndv.c
```diff
@@ -617,6 +619,18 @@ ucp_proto_rndv_ack_init(const ucp_proto_common_init_params_t *init_params,
...
 }
 
+static void
+ucp_proto_rndv_report_fast_cmpl(const ucp_proto_multi_init_params_t *params)
+{
+    uint32_t op_attr_mask = ucp_proto_select_op_attr_unpack(
+            params->super.super.select_param->op_attr);
+
+    if ((op_attr_mask & UCP_OP_ATTR_FLAG_FAST_CMPL)) {
+        ucs_warn("RNDV is initialized with FAST_CMPL which is unexpected. "
```
tvegas1, line 629: maybe `ucs_debug` since fast_cmpl is best effort and UCP users should (generally) not even be aware that a multi step protocol is used?

src/ucp/rndv/proto_rndv.c
```diff
@@ -630,6 +644,8 @@ ucp_proto_rndv_bulk_init(const ucp_proto_multi_init_params_t *init_params,
     const char *proto_name;
     ucs_status_t status;
 
+    ucp_proto_rndv_report_fast_cmpl(init_params);
```
yosefe, line 647: seems we remove FAST_CMPL when we handle RTR. So in which case we sould still have FAST_CMPL here?
maybe it should be an assertion?

src/ucp/rndv/proto_rndv.c
```diff
@@ -925,8 +941,11 @@ ucp_proto_rndv_handle_rtr(void *arg, void *data, size_t length, unsigned flags)
     /* RTR covers the whole send request - use the send request directly */
     ucs_assert(req->flags & UCP_REQUEST_FLAG_PROTO_INITIALIZED);
 
-    select_param = &req->send.proto_config->select_param;
-    op_attr_mask = ucp_proto_select_op_attr_unpack(select_param->op_attr);
+    select_param  = &req->send.proto_config->select_param;
+    op_attr_mask  = ucp_proto_select_op_attr_unpack(select_param->op_attr);
+    /* Remove FAST_CMPL since it should be considered only for one stage
+     * protocols (like eager) */
+    op_attr_mask &= ~UCP_OP_ATTR_FLAG_FAST_CMPL;
```
yosefe, line 948:
```suggestion
    /* Remove FAST_CMPL since it should be considered only for one stage
     * protocols (like eager) */
    op_attr_mask  = ucp_proto_select_op_attr_unpack(select_param->op_attr) &
                    ~UCP_OP_ATTR_FLAG_FAST_CMPL;
```

src/ucp/rndv/proto_rndv.c
ivankochin: AFAIU that can happen if customer set FAST_CMPL flag for receive operation. In that case RNDV-get would be initialized with FAST_CMPL flag

yosefe: so maybe make the warning more clear, or move it to tag_recv_nbx, or mask-out FAST_CMPL flag during tag_recv?

brminich (approved):

src/ucp/rndv/proto_rndv.c
ivankochin: I can move this warning to `tag_recv_nbx` (or maybe `ucp_proto_rndv_receive_start`) but then it would be extra operation on the fast path. And if we want to print a waring/debug message in that case it would cost extra `if` on fast path. Are we OK with that?

src/ucp/rndv/proto_rndv.c
ivankochin: That warning means that user passed FAST_CMPL flag to recv operation and in case of choosing RNDV it can cause undesired performance prediction change. So I would prefer to leave it as warning so users can see that and remove FAST_CMPL from recv operation.

src/ucp/rndv/proto_rndv.c
yosefe: ok, let's keep here

yosefe (approved):

PR #10278, opened 2024-11-05
UCT/DC: Increase DCI index to 16 bits

shasson5:
## What?
Increase DCI index to 16 bits

## Why?
Support large number of DCIs

### DIFF
--- a/src/uct/ib/mlx5/dc/dc_mlx5.h
+++ b/src/uct/ib/mlx5/dc/dc_mlx5.h
@@ -64,6 +64,7 @@ struct ibv_ravh {
 
 typedef struct uct_dc_mlx5_ep     uct_dc_mlx5_ep_t;
 typedef struct uct_dc_mlx5_iface  uct_dc_mlx5_iface_t;
+typedef uint16_t                  uct_dci_index_t;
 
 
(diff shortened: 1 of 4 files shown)


yosefe (approved):

roiedanino (approved):

src/uct/ib/mlx5/dc/dc_mlx5.h
```diff
@@ -64,6 +64,7 @@ struct ibv_ravh {
 
 typedef struct uct_dc_mlx5_ep     uct_dc_mlx5_ep_t;
 typedef struct uct_dc_mlx5_iface  uct_dc_mlx5_iface_t;
+typedef uint16_t                  uct_dci_index_t;
```
gleon99, line 67: `_dc_mlx5_`

roiedanino: I would prefer keeping it as short as possible, its being used in many places

shasson5: IMO would be too long and clumsy

yosefe:
@shasson5 can you pls squash?

PR #10281, opened 2024-11-06
UCT/CUDA/CUDA_IPC: Fix return value of is_peer_accessible.

rakhmets:
## What?
Fixed return value of `uct_cuda_ipc_is_peer_accessible`.
Restored old behavior, which was changed by https://github.com/openucx/ucx/pull/9978.

## Why?
Returning value other than `UCS_ERR_UNREACHABLE` causes an issue during unpacking remote key.
The issue can be reproduced by `iodemo` example application.
Client side:
```
ucp_mm.c:369  UCX  WARN  failed to dereg from md[4]=cuda_ipc: Unsupported operation
```
Server side:
```
ucp_rkey.c:899  UCX  ERROR   failed to unpack remote key from remote md[4]: Input/output error
```


iyastreb (approved):
Thanks for spotting this issue

brminich (approved):

PR #10286, opened 2024-11-07
UCT/DC: Increase DCI index to 16 bits - v1.18

shasson5:
## What?
Increase DCI index to 16 bits -> backported from master

## Why?
Support large number of DCIs

### DIFF
--- a/src/uct/ib/mlx5/dc/dc_mlx5.h
+++ b/src/uct/ib/mlx5/dc/dc_mlx5.h
@@ -64,6 +64,7 @@ struct ibv_ravh {
 
 typedef struct uct_dc_mlx5_ep     uct_dc_mlx5_ep_t;
 typedef struct uct_dc_mlx5_iface  uct_dc_mlx5_iface_t;
+typedef uint16_t                  uct_dci_index_t;
 
 
 typedef enum {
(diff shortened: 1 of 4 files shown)


yosefe (approved):

src/uct/ib/mlx5/dc/dc_mlx5.h
```diff
@@ -284,7 +285,7 @@ typedef struct {
 } uct_dc_mlx5_dci_pool_t;
 
 
-UCS_ARRAY_DECLARE_TYPE(uct_dc_dci_array_t, uint16_t, uct_dc_dci_t);
+UCS_ARRAY_DECLARE_TYPE(uct_dc_dci_array_t, uct_dci_index_t, uct_dc_dci_t);
```
roiedanino, line 288: I don't see this change in the original PR: https://github.com/openucx/ucx/pull/10278/

Maybe need to open another one and fix it

roiedanino (approved):
Just need to open another PR into master for changing `uct_dc_dci_array_t` index type (`uint16_t` -> `uct_dc_dci_index_t`)

src/uct/ib/mlx5/dc/dc_mlx5.h
shasson5: it was discovered during this PR, so indeed not in the original

PR #10291, opened 2024-11-11
BINDINGS/RUST: Initial Rust bindings

MattBBaker:
## What?
Make UCX bindings for Rust so the UCX can be used in Rust programs

## Why?
I wanted to learn Rust and decided the best way to do that is with a real project, so UCX gets Rust bindings and I wanted to share the results. This is very incomplete but I'm posting it early to see if someone with more Rust experience has suggestions. Right now you can setup and call an active message

## How?
Some of the interesting bits about Rust coming from a C/UCX programming model is that Rust functions often return a `Result` type, which has an `Ok` and `Err` type. In `lib.rs` you can find a wrapper function that will convert `ucs_status_ptr_t` returned from ucp into a `Result<Option<Handle>, ucs_status_t>`. You'll see a lot of `.unwrap()` function calls and that will evaluate the `Result` to either return the underlying `Option<Handle>` value or handle the error by aborting the programming and giving the string value of the error, something like `UCS_ERR_INVALID_PARAM`. You would think this would cause pain in the code generation, but Rust is very good at optimizing this away. It looks like Rust automatically does the same thing we often do in UCX code where there is a list of resource cleanup labels and goto's to jump to these labels if there is an error. The `Option<Handle>` offers a replacement for checking for `UCS_OK` and using the `UCS_PTR_IS_PTR()` macros. If the option is empty that is the equivalent to getting back `UCS_OK`. A non-empty option will have a `Handle` struct that represents a request handle that can be queried for completion. Rust makes very efficient assembly for this, as best as I can tell it's free? It's also a lot more readable if you compare the UCX example codes and the test in the `lib.rs` function.

Another thing worth explaining is the Builder pattern used. Rust wants to be a safe language free from undefined behavior. In UCX we use field flags in various param structs to define if a field in present in a struct or not and then UCX promises to not access those fields. However, Rust cannot see this promise since it's in the documentation and the compiler cannot read that and will be unhappy. This can be solved though using the `unsafe` keyword. So the builder pattern can be used to mark the field flags and set the parameter in the structure and only touching the struct and declaring it initialized is unsafe. In the generated code this will all compile down into a const struct. Another interesting trick is that in the request param struct builder there is an if to check if there is a mismatch with the user asking for both no immediate completion and force immediate completion. This does not translate into any assembly and will either compile down to a const or it will fail at compile time. It feels like it should be possible to go further and do things like make sure that EP param structs have one and only one connection type. 

The best way to start reviewing this code should probably start with `lib.rs`. That file is the one with the best docs in the comments. The file `context.rs` is also good. 

If you want to try it out you will need to first do the regular configure/build of the base library and then in the Rust bindings dir you can run `cargo build` to build the library and `LD_LIBRARY_PATH=$LD_LIBRARY_PATH:../../src/ucs/.libs:../../src/uct/.libs:../../src/ucm/.libs:../../src/ucp/.libs cargo test` to run the test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,7 @@ go.work
 go.work.sum
 bindings/go/pkg/
 bindings/go/go.sum
+*~
+\#*
+bindings/rust/target
+bindings/rust/Cargo.lock
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "ucx-sys"
+version = "0.1.0"
+edition = "2021"
+description = "Low level Rust bindings"
+license = "BSD"
+
+[build-dependencies]
+bindgen = "0.70.1"
+
+[dependencies.bitflags]
+version = "2.6.0"
--- a/bindings/rust/build.rs
+++ b/bindings/rust/build.rs
@@ -1,0 +1,45 @@
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    // Tell cargo to look for shared libraries in the specified directory
+    println!("cargo:rustc-link-search=../../src/ucp/.libs/");
+
+    // Tell cargo to tell rustc to link the system bzip2
+    // shared library.
+    println!("cargo:rustc-link-lib=ucp");
+
+    // The bindgen::Builder is the main entry point
+    // to bindgen, and lets you build up options for
+    // the resulting bindings.
+    let bindings = bindgen::Builder::default()
+        // Some of the UCX detailed examples in comments can confuse the
+        // bindgen parser and it will make bad code instead of comments
+        .generate_comments(false)
+        // ucs_status_t is defined as a packed enum and that will lead to
+        // badness without the flag which tells bindgen to repeat that
+        // trick with the rust enums
+        .rustified_enum(".*")
+        .clang_arg("-I../../src/ucp/api/")
+        .clang_arg("-I../../")
+        .clang_arg("-I../../src/")
+        // Annotate ucs_status_t and ucs_status_ptr_t as #[must_use]
+        .must_use_type("ucs_status_t")
+        .must_use_type("ucs_status_ptr_t")
+        // The input header we would like to generate
+        // bindings for.
+        .header("wrapper.h")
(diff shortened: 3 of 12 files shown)


gleon99:
Hi @MattBBaker 
adding Rust bindings is quite an interesting idea! Though we need to discuss whether it indeed justifies upstreaming. Some follow-up questions:
1. Can you measure the performance of the Rust version (BW, latency) by having some simple benchmark and compare it to C? The simplest one can be looping AM send/receive, sizes range from 1B to 1MB (see also OSU benchmarks for examples).
2. Can you think of any any real use-case scenario, where these bindings might be of use?
3. Please elaborate on how you handle Rust memory safety guarantees with UCX C functions which require raw buffer pointers. Seems everything runs in "unsafe" mode?

MattBBaker:
hi @gleon99 

1. I've not yet run benchmarks. I was hoping to get the RDMA put/get functions done and running a OSU a like benchmark for tags/puts/gets. I've looked at the generated assembly and it looks like most tests compile down to just setting up library function calls.
2. One thing that caught my attention was a recent talk from Google about writing a TCP/IP stack for Fuchsia in Rust and finding a total of 3 bugs in a year of use. I've done a lot of work in OpenSHMEM and remember a linter tool that reminded me of the problems that Google solved using the Rust type system for their Fuscha Netstack3 code, [OpenSHMEM analyzer](http://www.openshmem.org/site/sites/default/site_files/osa-1.0.pdf). I'm thinking that if a lot of this functionality, such as memory type analyzer, can be encoded into the Rust type system that it will make it easier to write correct applications on top of UCX. 
3. Yes, the Rust to C FFI routines always run in "unsafe" mode since the Rust compiler cannot inspect the code of the C library, in this case `libucp.so`. My plan is to make this bindings in a way that the bindings fulfill the Rust guarantees correctly. One thing I'm looking into is coding a way that the bindings can take ownership of memory in the `UCS_INPROGRESS` cases where you have to work with a `RequestHandle` and does nothing in the `UCS_OK` case where the user application still owns the memory.

gleon99:
Hi @MattBBaker - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #10292, opened 2024-11-12
UCP/PROTO: Consider RNDV_PERF_DIFF

ivankochin:
## What?
Applies `UCX_RNDV_PERF_DIFF` setting effect for protov2. 

## Why?
This control effect was removed during introducing perf-factors logic. This PR brings it back.

That how it looks like with that patch:
image

### DIFF
--- a/src/ucp/proto/proto_perf.c
+++ b/src/ucp/proto/proto_perf.c
@@ -430,13 +430,14 @@ ucs_status_t ucp_proto_perf_aggregate2(const char *name,
     return ucp_proto_perf_aggregate(name, perf_elems, 2, perf_p);
 }
 
-void ucp_proto_perf_apply_bias(ucp_proto_perf_t *perf, double bias) {
+void ucp_proto_perf_apply_bias(ucp_proto_perf_t *perf, double bias)
+{
     ucs_linear_func_t bias_func = ucs_linear_func_make(0.0, 1.0 - bias);
     ucp_proto_perf_node_t *bias_node;
     ucp_proto_perf_factor_id_t fid;
     ucp_proto_perf_segment_t *seg;
 
-    if (bias == 0) {
+    if (fabs(bias) <= UCP_PROTO_PERF_EPSILON) {
         return;
     }
 
--- a/src/ucp/proto/proto_perf.c
+++ b/src/ucp/proto/proto_perf.c
@@ -430,6 +430,27 @@ ucs_status_t ucp_proto_perf_aggregate2(const char *name,
     return ucp_proto_perf_aggregate(name, perf_elems, 2, perf_p);
 }
 
+void ucp_proto_perf_apply_bias(ucp_proto_perf_t *perf, double bias) {
+    ucs_linear_func_t bias_func = ucs_linear_func_make(0.0, 1.0 - bias);
+    ucp_proto_perf_node_t *bias_node;
+    ucp_proto_perf_factor_id_t fid;
+    ucp_proto_perf_segment_t *seg;
+
+    if (bias == 0) {
+        return;
+    }
+
+    ucp_proto_perf_segment_foreach(seg, perf) {
+        for (fid = 0; fid < UCP_PROTO_PERF_FACTOR_LAST; ++fid) {
+            seg->perf_factors[fid] = ucs_linear_func_compose(
+                    bias_func, seg->perf_factors[fid]);
+            ucp_proto_perf_node_update_factors(seg->node, seg->perf_factors);
+        }
+        bias_node = ucp_proto_perf_node_new_data("bias", "%.2f %%", bias);
+        ucp_proto_perf_node_own_child(seg->node, &bias_node);
+    }
+}
+
 /* TODO:
  * Reconsider correctness of PPLN perf estimation logic since in case of async
  * operations it seems wrong to choose the longest factor without paying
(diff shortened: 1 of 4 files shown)


ivankochin:
@yosefe @brminich currently patch improves all RNDV protocols, but maybe it worth to think about applying this change to RNDV-get/put only since there can be cases when RMA is unsupported for some reason and on big sizes there would be comparison between Eager and RNDV through AM and BW of latter would be slightly higher due to `RNDV_PERF_DIFF`. WDYT?

brminich:
> @yosefe @brminich currently patch improves all RNDV protocols, but maybe it worth to think about applying this change to RNDV-get/put only since there can be cases when RMA is unsupported for some reason and on big sizes there would be comparison between Eager and RNDV through AM and BW of latter would be slightly higher due to `RNDV_PERF_DIFF`. WDYT?

I’d leave it for all protocols for simplicity. Additionally, we have PPLN protocols, and even AM-based RNDV may be preferable to eager due to its lower memory consumption.

ivankochin:
> > @yosefe @brminich currently patch improves all RNDV protocols, but maybe it worth to think about applying this change to RNDV-get/put only since there can be cases when RMA is unsupported for some reason and on big sizes there would be comparison between Eager and RNDV through AM and BW of latter would be slightly higher due to `RNDV_PERF_DIFF`. WDYT?
> 
> I’d leave it for all protocols for simplicity. Additionally, we have PPLN protocols, and even AM-based RNDV may be preferable to eager due to its lower memory consumption.

I thought AM-based RNDV and PPLN protocols consume same amount of memory as eager, aren't they? Do AM-based protocols send message directly to user-provided buffer?

brminich:
> > > @yosefe @brminich currently patch improves all RNDV protocols, but maybe it worth to think about applying this change to RNDV-get/put only since there can be cases when RMA is unsupported for some reason and on big sizes there would be comparison between Eager and RNDV through AM and BW of latter would be slightly higher due to `RNDV_PERF_DIFF`. WDYT?
> > 
> > 
> > I’d leave it for all protocols for simplicity. Additionally, we have PPLN protocols, and even AM-based RNDV may be preferable to eager due to its lower memory consumption.
> 
> I thought AM-based RNDV and PPLN protocols consume same amount of memory as eager, aren't they? Do AM-based protocols send message directly to user-provided buffer?

In case of tag matching API eager messages are queued on the receiver until receive operation is invoked. Such eager fragments may consume significant amount of memory on the receiver. With RNDV data transfer is started only when receiver is ready

ivankochin:
> > > > @yosefe @brminich currently patch improves all RNDV protocols, but maybe it worth to think about applying this change to RNDV-get/put only since there can be cases when RMA is unsupported for some reason and on big sizes there would be comparison between Eager and RNDV through AM and BW of latter would be slightly higher due to `RNDV_PERF_DIFF`. WDYT?
> > > 
> > > 
> > > I’d leave it for all protocols for simplicity. Additionally, we have PPLN protocols, and even AM-based RNDV may be preferable to eager due to its lower memory consumption.
> > 
> > 
> > I thought AM-based RNDV and PPLN protocols consume same amount of memory as eager, aren't they? Do AM-based protocols send message directly to user-provided buffer?
> 
> In case of tag matching API eager messages are queued on the receiver until receive operation is invoked. Such eager fragments may consume significant amount of memory on the receiver. With RNDV data transfer is started only when receiver is ready

Isn't tag matching API

src/ucp/proto/proto_perf.c
```diff
@@ -430,6 +430,27 @@ ucs_status_t ucp_proto_perf_aggregate2(const char *name,
     return ucp_proto_perf_aggregate(name, perf_elems, 2, perf_p);
 }
 
+void ucp_proto_perf_apply_bias(ucp_proto_perf_t *perf, double bias) {
```
yosefe, line 433: code style

src/ucp/proto/proto_perf.c
```diff
@@ -430,6 +430,27 @@ ucs_status_t ucp_proto_perf_aggregate2(const char *name,
...
+    ucs_linear_func_t bias_func = ucs_linear_func_make(0.0, 1.0 - bias);
+    ucp_proto_perf_node_t *bias_node;
+    ucp_proto_perf_factor_id_t fid;
+    ucp_proto_perf_segment_t *seg;
+
+    if (bias == 0) {
```
yosefe, line 439: compare with some epsilon value?

src/ucp/proto/proto_perf.c
```diff
@@ -430,6 +430,27 @@ ucs_status_t ucp_proto_perf_aggregate2(const char *name,
...
+
+    ucp_proto_perf_segment_foreach(seg, perf) {
+        for (fid = 0; fid < UCP_PROTO_PERF_FACTOR_LAST; ++fid) {
+            seg->perf_factors[fid] = ucs_linear_func_compose(
+                    bias_func, seg->perf_factors[fid]);
+            ucp_proto_perf_node_update_factors(seg->node, seg->perf_factors);
+        }
+        bias_node = ucp_proto_perf_node_new_data("bias", "%.2f %%", bias);
+        ucp_proto_perf_node_own_child(seg->node, &bias_node);
+    }
```
yosefe, line 451: can we implement this using ucp_proto_perf_add_funcs?

rakhmets:
Moved to https://github.com/openucx/ucx/pull/10401.

PR #10295, opened 2024-11-13
PERF: Fix integer overflow: use size_t instead of unsigned for buffer size

iyastreb:
## What?
In perftest there were several places where `unsigned` 32b is used to represent buffer size, which is of type `size_t` in the rest of the code

## Why?
Integer overflow occurs on large message sizes > 4GB, and as a consequence perftest produces completely random results

## How?
Use `size_t` instead of `unsigned` for buffer size

### DIFF
--- a/src/tools/perf/lib/ucp_tests.cc
+++ b/src/tools/perf/lib/ucp_tests.cc
@@ -473,7 +473,7 @@ public:
     }
 
     UCS_F_ALWAYS_INLINE ucs_status_t send_daemon_req(void *buffer,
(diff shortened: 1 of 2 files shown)


brminich (approved):

yosefe (approved):
is there a way to test it in gtest/CI? could be a separate PR

iyastreb:
> is there a way to test it in gtest/CI? could be a separate PR

it should be possible to test in CI by sending a message of 4GB + 1B length
will note and address later

PR #10305, opened 2024-11-15
UCS/CONFIG/PARSER: Handle NULL entries in sparse arrays

michal-shalev:
## What?
Following https://github.com/openucx/ucx/pull/10172
Introduced `UCS_CONFIG_TYPE_FLAGS` to support sparse parsing of configuration flags.
Removed unnecessary flags from `UCX_EXTRA_OP_ATTR_FLAGS`.

## Why?
UCS_CONFIG_TYPE_BITMAP required contiguous indices, leading to core dumps when NULL values were accessed due to gaps in the array.
UCS_CONFIG_TYPE_FLAGS enables sparse arrays for flags, eliminating the need for unnecessary placeholders and simplifying the flag configuration.

## How?
Implemented `UCS_CONFIG_TYPE_FLAGS` to loop until the specified number of arguments instead of relying on NULL-terminated arrays.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -141,11 +141,8 @@ const char *ucp_object_versions[] = {
 };
 
 const char *ucp_extra_op_attr_flags_names[] = {
-    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_NO_IMM_CMPL)]    = "no_imm_cmpl",
     [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_FAST_CMPL)]      = "fast_cmpl",
-    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_FORCE_IMM_CMPL)] = "force_imm_cmpl",
-    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_MULTI_SEND)]     = "multi_send",
-    NULL
+    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_MULTI_SEND)]     = "multi_send"
 };
 
 const ucs_config_flags_args_t ucp_extra_op_attr_flags_args = {
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -148,6 +148,11 @@ const char *ucp_extra_op_attr_flags_names[] = {
     NULL
 };
 
+const ucs_config_flags_args_t ucp_extra_op_attr_flags_args = {
+    .args        = ucp_extra_op_attr_flags_names,
+    .num_of_args = ucs_static_array_size(ucp_extra_op_attr_flags_names)
+};
+
 static UCS_CONFIG_DEFINE_ARRAY(memunit_sizes, sizeof(size_t),
                                UCS_CONFIG_TYPE_MEMUNITS);
 
@@ -544,7 +549,7 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "in addition to what is set explicitly by the user. \n"
    "Possible values are: no_imm_cmpl, fast_cmpl, force_imm_cmpl, multi_send.",
    ucs_offsetof(ucp_context_config_t, extra_op_attr_flags),
-   UCS_CONFIG_TYPE_BITMAP(ucp_extra_op_attr_flags_names)},
+   UCS_CONFIG_TYPE_FLAGS((const void *)&ucp_extra_op_attr_flags_args)},
 
   {NULL}
 };
--- a/src/ucs/config/parser.c
+++ b/src/ucs/config/parser.c
@@ -406,6 +406,35 @@ static void __print_table_values(char * const *table, char *buf, size_t max)
     *buf = '[';
 }
 
+static void __print_sparse_table_values(const char * const *table, size_t num_of_args, char *buf, size_t max)
+{
+    char *ptr = buf, *end = buf + max - 1;
+    size_t i;
+
+    if (ptr < end) {
+        *ptr++ = '[';
+    }
+
+    /* Ingores NULL entries */
+    for (i = 0; i < num_of_args && ptr < end; ++i) {
+        if (table[i] == NULL) {
+            continue;
+        }
+
+        if (ptr > buf + 1 && ptr < end) {
+            *ptr++ = '|';
+        }
+
+        ptr += snprintf(ptr, end - ptr, "%s", table[i]);
+    }
+
+    if (ptr < end) {
+        *ptr++ = ']';
+    }
+
+    *ptr = '\0';
+}
+
 void ucs_config_help_enum(char *buf, size_t max, const void *arg)
 {
     __print_table_values(arg, buf, max);
@@ -473,6 +502,57 @@ void ucs_config_help_bitmap(char *buf, size_t max, const void *arg)
 {
     snprintf(buf, max, "comma-separated list of: ");
     __print_table_values(arg, buf + strlen(buf), max - strlen(buf));
+}
+
+int ucs_config_sscanf_flags(const char *buf, void *dest, const void *arg)
+{
+    char *str = ucs_strdup(buf, "config_sscanf_flags_str");
+    ucs_config_flags_args_t *args = (ucs_config_flags_args_t*)arg;
+    char *p, *saveptr;
+    int ret, i;
+
+    if (str == NULL) {
+        return 0;
+    }
+
+    ret = 1;
+    *((uint64_t*)dest) = 0;
+    p = strtok_r(str, ",", &saveptr);
+    while (p != NULL) {
+        i = ucs_string_find_in_sparse_list(p, (const char**)args->args,
+                                           args->num_of_args);
+        if (i < 0) {
+            ret = 0;
+            break;
+        }
+
+        ucs_assertv(i < (sizeof(uint64_t) * 8), "bit %d overflows for '%s'", i,
+                    p);
+        *((uint64_t*)dest) |= UCS_BIT(i);
+        p = strtok_r(NULL, ",", &saveptr);
+    }
+
+    ucs_free(str);
+    return ret;
+}
+
+int ucs_config_sprintf_flags(char *buf, size_t max, const void *src, const void *arg)
+{
+    ucs_config_flags_args_t *args = (ucs_config_flags_args_t*)arg;
+    ucs_flags_str(buf, max, *((uint64_t*)src), (const char**)args->args);
+    return 1;
+}
+
+void ucs_config_help_flags(char *buf, size_t max, const void *arg)
+{
+    ucs_config_flags_args_t *args = (ucs_config_flags_args_t*)arg;
+    int written = snprintf(buf, max, "comma-separated list of: ");
+
+    if (written < 0 || written >= max) {
+        return;
+    }
+
+    __print_sparse_table_values(args->args, args->num_of_args, buf + written, max - written);
 }
 
 int ucs_config_sscanf_bitmask(const char *buf, void *dest, const void *arg)
--- a/src/ucs/config/parser.h
+++ b/src/ucs/config/parser.h
@@ -132,6 +132,11 @@ typedef struct ucs_config_bw_spec {
 } ucs_config_bw_spec_t;
 
 
+typedef struct ucs_config_flags_args {
+    const char              **args;
+    size_t                  num_of_args;
+} ucs_config_flags_args_t;
+
 #define UCS_CONFIG_EMPTY_GLOBAL_LIST_ENTRY \
     { \
         .name        = "", \
@@ -231,6 +236,10 @@ void ucs_config_help_uint_enum(char *buf, size_t max, const void *arg);
 int ucs_config_sscanf_bitmap(const char *buf, void *dest, const void *arg);
 int ucs_config_sprintf_bitmap(char *buf, size_t max, const void *src, const void *arg);
 void ucs_config_help_bitmap(char *buf, size_t max, const void *arg);
+
+int ucs_config_sscanf_flags(const char *buf, void *dest, const void *arg);
+int ucs_config_sprintf_flags(char *buf, size_t max, const void *src, const void *arg);
+void ucs_config_help_flags(char *buf, size_t max, const void *arg);
 
 int ucs_config_sscanf_bitmask(const char *buf, void *dest, const void *arg);
 int ucs_config_sprintf_bitmask(char *buf, size_t max, const void *src, const void *arg);
@@ -382,6 +391,11 @@ void ucs_config_help_generic(char *buf, size_t max, const void *arg);
 #define UCS_CONFIG_TYPE_BITMAP(t)  {ucs_config_sscanf_bitmap,    ucs_config_sprintf_bitmap, \
                                     ucs_config_clone_uint,       ucs_config_release_nop, \
                                     ucs_config_help_bitmap,      ucs_config_doc_nop, \
+                                    t}
+
+#define UCS_CONFIG_TYPE_FLAGS(t)   {ucs_config_sscanf_flags, ucs_config_sprintf_bitmap, \
+                                    ucs_config_clone_uint,   ucs_config_release_nop, \
+                                    ucs_config_help_flags,   ucs_config_doc_nop, \
                                     t}
 
 #define UCS_CONFIG_TYPE_BITMASK    {ucs_config_sscanf_bitmask,   ucs_config_sprintf_bitmask, \
--- a/src/ucs/sys/string.c
+++ b/src/ucs/sys/string.c
@@ -423,6 +423,25 @@ ssize_t ucs_string_find_in_list(const char *str, const char **string_list,
     return -1;
 }
 
+ssize_t ucs_string_find_in_sparse_list(const char *str, const char **string_list,
+                                       size_t num_of_args)
+{
+    size_t i;
+
+    /* Ingnores NULL entries */
+    for (i = 0; i < num_of_args; ++i) {
+        if (string_list[i] == NULL) {
+            continue;
+        }
+        if (strcmp(string_list[i], str) == 0) {
+            return i;
+        }
+    }
+
+    return -1;
+}
+
+
 char* ucs_string_split(char *str, const char *delim, int count, ...)
 {
     char *p = str;
--- a/src/ucs/sys/string.h
+++ b/src/ucs/sys/string.h
@@ -324,6 +324,17 @@ const char* ucs_mask_str(uint64_t mask, ucs_string_buffer_t *strb);
 ssize_t ucs_string_find_in_list(const char *str, const char **string_list,
                                 int case_sensitive);
 
+/**
+ * Find a string in an array of strings with NULL entries.
+ *
+ * @param str          String to search for.
+ * @param string_list  Array of strings.
+ * @param num_of_args  Number of strings in the array.
+ *
+ * @return Index of the string in the array, or -1 if not found.
+ */
+ssize_t ucs_string_find_in_sparse_list(const char *str, const char **string_list,
+                                       size_t num_of_args);
 
 /**
  * Split a string to tokens. The given string is modified in-place.
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -43,14 +43,8 @@ enum {
     UCP_TL_RSC_FLAG_AUX = UCS_BIT(0)
 };
 
-#define UCP_OP_ATTR_INDEX_MASK (UCP_OP_ATTR_FLAG_NO_IMM_CMPL    | \
-                                UCP_OP_ATTR_FLAG_FORCE_IMM_CMPL | \
-                                UCP_OP_ATTR_FLAG_FAST_CMPL      | \
-                                UCP_OP_ATTR_FLAG_MULTI_SEND)
-
 #define UCP_OP_ATTR_INDEX(_op_attr_flag) \
-    (ucs_ilog2(ucp_proto_select_op_attr_pack((_op_attr_flag), \
-                                             UCP_OP_ATTR_INDEX_MASK)))
+    (ucs_ilog2(ucp_proto_select_op_attr_pack((_op_attr_flag))))
 
 
 typedef struct ucp_context_config {
--- a/src/ucp/proto/proto_select.h
+++ b/src/ucp/proto/proto_select.h
@@ -44,8 +44,9 @@
 
 
 /* Pack operation attributes from uint32_t to a uint8_t */
-#define ucp_proto_select_op_attr_pack(_op_attr, _mask) \
-    (((_op_attr) & (_mask)) / UCP_PROTO_SELECT_OP_ATTR_BASE)
+#define ucp_proto_select_op_attr_pack(_op_attr) \
+    (((_op_attr) & (UCP_PROTO_SELECT_OP_ATTR_MASK)) / \
+    UCP_PROTO_SELECT_OP_ATTR_BASE)
 
 
 typedef struct {
--- a/src/ucp/proto/proto_select.inl
+++ b/src/ucp/proto/proto_select.inl
@@ -135,7 +135,7 @@ static UCS_F_ALWAYS_INLINE void ucp_proto_select_param_init_common(
      * supported */
     select_param->op_id_flags   = op_id | op_flags;
     select_param->op_attr       = ucp_proto_select_op_attr_pack(
-        op_attr_mask, UCP_PROTO_SELECT_OP_ATTR_MASK);
+        op_attr_mask);
     select_param->dt_class      = dt_class;
     select_param->mem_type      = mem_info->type;
     select_param->sys_dev       = mem_info->sys_dev;
--- a/src/ucp/rndv/rndv_ppln.c
+++ b/src/ucp/rndv/rndv_ppln.c
@@ -84,7 +84,7 @@ ucp_proto_rndv_ppln_probe(const ucp_proto_init_params_t *init_params)
     sel_param.op_id_flags = ucp_proto_select_op_id(select_param) |
                             UCP_PROTO_SELECT_OP_FLAG_PPLN_FRAG;
     sel_param.op_attr     = ucp_proto_select_op_attr_pack(
-        UCP_OP_ATTR_FLAG_MULTI_SEND, UCP_PROTO_SELECT_OP_ATTR_MASK);
+        UCP_OP_ATTR_FLAG_MULTI_SEND);
 
     proto_select = ucp_proto_select_get(worker, init_params->ep_cfg_index,
                                         init_params->rkey_cfg_index,


src/ucs/sys/string.h
```diff
@@ -324,6 +324,18 @@ const char* ucs_mask_str(uint64_t mask, ucs_string_buffer_t *strb);
 ssize_t ucs_string_find_in_list(const char *str, const char **string_list,
                                 int case_sensitive);
 
+/**
+ * Find a string in an array of strings with NULL entries.
+ *
+ * @param str          String to search for.
+ * @param string_list  Array of strings.
+ * @param num_of_args  Number of strings in the array.
```
iyastreb, line 332: I would propose to name this arg similar to array, eg `str_array` and `str_array_size`

src/ucp/core/ucp_context.c
```diff
@@ -141,11 +141,13 @@ const char *ucp_object_versions[] = {
...
     [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_FAST_CMPL)]      = "fast_cmpl",
-    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_FORCE_IMM_CMPL)] = "force_imm_cmpl",
-    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_MULTI_SEND)]     = "multi_send",
-    NULL
+    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_MULTI_SEND)]     = "multi_send"
+};
+
+const ucs_config_flags_args_t ucp_extra_op_attr_flags_args = {
+    .args        = ucp_extra_op_attr_flags_names,
+    .num_of_args = ucs_static_array_size(ucp_extra_op_attr_flags_names)
```
iyastreb, line 150: maybe `args_size`?

src/ucp/core/ucp_context.c
```diff
@@ -544,7 +546,7 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "in addition to what is set explicitly by the user. \n"
    "Possible values are: no_imm_cmpl, fast_cmpl, force_imm_cmpl, multi_send.",
    ucs_offsetof(ucp_context_config_t, extra_op_attr_flags),
-   UCS_CONFIG_TYPE_BITMAP(ucp_extra_op_attr_flags_names)},
+   UCS_CONFIG_TYPE_FLAGS((const void *)&ucp_extra_op_attr_flags_args)},
```
iyastreb, line 549: Could you please add few unit tests for this new option (test_config.cc)

src/ucs/config/parser.c
```diff
@@ -475,6 +506,59 @@ void ucs_config_help_bitmap(char *buf, size_t max, const void *arg)
...
+    p = strtok_r(str, ",", &saveptr);
+    while (p != NULL) {
+        i = ucs_string_find_in_sparse_list(p, (const char**)args->args,
+                                           args->num_of_args);
+        if (i < 0) {
+            ret = 0;
+            break;
+        }
+
+        ucs_assertv(i < (sizeof(uint64_t) * 8), "bit %d overflows for '%s'", i,
```
iyastreb, line 531: maybe use CHAR_BIT const instead of 8

src/ucs/config/parser.c
```diff
@@ -475,6 +506,59 @@ void ucs_config_help_bitmap(char *buf, size_t max, const void *arg)
...
+    ucs_config_flags_args_t *args = (ucs_config_flags_args_t*)arg;
+    char *p, *saveptr;
+    int ret, i;
+
+    if (str == NULL) {
+        return 0;
+    }
+
+    ret = 1;
+    *((uint64_t*)dest) = 0;
```
iyastreb, line 521: I would propose to create a variable to avoid repetitive casts:
`uint64_t *flags = dest;`

src/ucs/config/parser.c
```diff
@@ -475,6 +506,59 @@ void ucs_config_help_bitmap(char *buf, size_t max, const void *arg)
...
+    p = strtok_r(str, ",", &saveptr);
```
iyastreb, line 522: consider using `ucs_string_split`

src/ucs/config/parser.c
```diff
@@ -475,6 +506,59 @@ void ucs_config_help_bitmap(char *buf, size_t max, const void *arg)
...
+    ucs_flags_str(buf, max, *((uint64_t*)src), (const char**)args->args);
+    return 1;
+}
+
+void ucs_config_help_flags(char *buf, size_t max, const void *arg)
+{
+    ucs_config_flags_args_t *args = (ucs_config_flags_args_t*)arg;
+    int written                   = snprintf(buf, max, "comma-separated list of: ");
+
+    if (written < 0 || written >= max) {
```
tvegas1, line 554: () and ()

src/ucs/config/parser.c
```diff
@@ -406,6 +406,37 @@ static void __print_table_values(char * const *table, char *buf, size_t max)
...
+{
+    char *ptr = buf, *end = buf + max - 1;
+    size_t i;
+
+    if (ptr < end) {
+        *ptr++ = '[';
+    }
+
+    /* Ingores NULL entries */
+    for (i = 0; i < num_of_args && ptr < end; ++i) {
```
tvegas1, line 421: () and (), everywhere

src/ucp/core/ucp_context.c
```diff
@@ -141,11 +141,13 @@ const char *ucp_object_versions[] = {
 };
 
 const char *ucp_extra_op_attr_flags_names[] = {
-    [UCP_OP_ATTR_INDEX(UCP_OP_ATTR_FLAG_NO_IMM_CMPL)]    = "no_imm_cmpl",
```
tvegas1, line 144: for my understanding, why removing no_imm_cmpl and force_imm_cmpl?

src/ucp/core/ucp_context.c
brminich, line 549: pls fix comment on line 547, as `no_imm_cmpl`  and `force_imm_cmpl`  are not the options anymore.

src/ucs/config/parser.h
```diff
@@ -384,6 +393,11 @@ void ucs_config_help_generic(char *buf, size_t max, const void *arg);
                                     ucs_config_help_bitmap,      ucs_config_doc_nop, \
                                     t}
 
+#define UCS_CONFIG_TYPE_FLAGS(t)   {ucs_config_sscanf_flags, ucs_config_sprintf_bitmap, \
```
brminich, line 396:
```suggestion
#define UCS_CONFIG_TYPE_FLAGS(t)   {ucs_config_sscanf_flags, ucs_config_sprintf_flags, \
```
?

tvegas1 (commented):
is it possible to fix `UCS_CONFIG_TYPE_BITMAP` for sparse, instead of introducing new `UCS_CONFIG_TYPE_FLAGS`?

src/ucs/config/parser.c
```diff
@@ -406,6 +406,37 @@ static void __print_table_values(char * const *table, char *buf, size_t max)
...
+        if (table[i] == NULL) {
+            continue;
+        }
+
+        if (ptr > buf + 1 && ptr < end) {
```
brminich, line 426:
```suggestion
        if ((ptr > (buf + 1)) && (ptr < end)) {
```

src/ucs/config/parser.c
brminich, line 554:
```suggestion
    if ((written < 0) || (written >= max)) {
```

src/ucs/sys/string.c
```diff
@@ -423,6 +423,26 @@ ssize_t ucs_string_find_in_list(const char *str, const char **string_list,
     return -1;
 }
 
+ssize_t ucs_string_find_in_sparse_list(const char *str,
+                                       const char **string_list,
+                                       size_t num_of_args)
+{
+    size_t i;
+
+    /* Ingnores NULL entries */
```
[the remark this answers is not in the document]
michal-shalev: fixed

src/ucs/sys/string.h
michal-shalev: done

src/ucp/core/ucp_context.c
michal-shalev: done

src/ucs/config/parser.c
michal-shalev: done

src/ucs/config/parser.c
michal-shalev: done

src/ucp/core/ucp_context.c
michal-shalev: it wasn't supposed to be there in the first place, it was a workaround until this PR

src/ucs/config/parser.h
michal-shalev: great catch, fixed

src/ucs/config/parser.c
michal-shalev: done

yosefe:
> is it possible to fix `UCS_CONFIG_TYPE_BITMAP` for sparse, instead of introducing new `UCS_CONFIG_TYPE_FLAGS`?

We could do it nicely by adding another "unsigned long" field to ucs_config_parser_t to hold the count, something like
```c
    ...
    void                     (*help)(char *buf, size_t max, const void *arg);
    void                     (*doc)(ucs_string_buffer_t *strb, const void *arg);
    const void               *arg_ptr;
    unsigned long            arg_ulong;
}
```

michal-shalev:
> > is it possible to fix `UCS_CONFIG_TYPE_BITMAP` for sparse, instead of introducing new `UCS_CONFIG_TYPE_FLAGS`?
> 
> We could do it nicely by adding another "unsigned long" field to ucs_config_parser_t to hold the count, something like
> 
> ```c
> typedef struct ucs_config_parser {
>     ...
>     void                     (*help)(char *buf, size_t max, const void *arg);
>     void                     (*doc)(ucs_string_buffer_t *strb, const void *arg);
>     const void               *arg_ptr;
>     unsigned long            arg_ulong;
> };
> ```
> 
> And setting arg_ulong by ucs_static_array_size

We can but it will require changes in all the parsers, which is more than the scope of the flags, LMK WDYT

src/ucs/sys/string.h
```diff
@@ -324,6 +324,18 @@ const char* ucs_mask_str(uint64_t mask, ucs_string_buffer_t *strb);
 ssize_t ucs_string_find_in_list(const char *str, const char **string_list,
                                 int case_sensitive);
 
+/**
+ * Find a string in an array of strings with NULL entries.
+ *
+ * @param str          String to search for.
+ * @param string_list  Array of strings.
```
brminich, line 331: i'd call it `str_array` to be consistent with `str_array_size`

yosefe:
> > > is it possible to fix `UCS_CONFIG_TYPE_BITMAP` for sparse, instead of introducing new `UCS_CONFIG_TYPE_FLAGS`?
> > 
> > 
> > We could do it nicely by adding another "unsigned long" field to ucs_config_parser_t to hold the count, something like
> > ```c
> > typedef struct ucs_config_parser {
> >     ...
> >     void                     (*help)(char *buf, size_t max, const void *arg);
> >     void                     (*doc)(ucs_string_buffer_t *strb, const void *arg);
> >     const void               *arg_ptr;
> >     unsigned long            arg_ulong;
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And setting arg_ulong by ucs_static_array_size
> 
> We could but it will require changes in all the parsers, which is more than the scope of the flags, LMK WDYT @yosefe

I think it's worth it , but would appreciate @gleon99 @brminich @tvegas1 opinions

tvegas1:
> > We could but it will require changes in all the parsers, which is more than the scope of the flags, LMK WDYT @yosefe
> 
> I think it's worth it , but would appreciate @gleon99 @brminich @tvegas1 opinions

I would try to have only one `UCS_CONFIG_TYPE_BITMAP` that supports holes because `_TYPE_FLAGS` seems to be more or less the same, if that's possible.

src/ucs/config/parser.h
```diff
@@ -369,20 +499,20 @@ void ucs_config_help_generic(char *buf, size_t max, const void *arg);
...
                                      "<on|off|auto>"}
 
-#define UCS_CONFIG_TYPE_ENUM(t)    {ucs_config_sscanf_enum,      ucs_config_sprintf_enum, \
-                                    ucs_config_clone_uint,       ucs_config_release_nop, \
-                                    ucs_config_help_enum,        ucs_config_doc_nop, \
-                                    t}
+#define UCS_CONFIG_TYPE_ENUM(allowed_values) {ucs_config_sscanf_enum, ucs_config_sprintf_enum, \
+                                              ucs_config_clone_uint,  ucs_config_release_nop, \
+                                              ucs_config_help_enum,   ucs_config_doc_nop, \
+                                              &UCS_CONFIG_ALLOWED_VALUES_NAME(allowed_values)}
```
iyastreb, line 505: I think you can get rid of all DECLARE and DEFINE macros by using slightly more advanced macro here, with static initializer and compound literal:
```
#define UCS_CONFIG_TYPE_ENUM(t)    {ucs_config_sscanf_enum,      ucs_config_sprintf_enum, \
                                    ucs_config_clone_uint,       ucs_config_release_nop, \
                                    ucs_config_help_enum,        ucs_config_doc_nop, \
        (const ucs_config_allowed_values_t[]) { \
            { .values=t, .count = ucs_static_array_size(t)} \
        } \
}
```

src/ucs/config/parser.h
```diff
@@ -69,7 +81,7 @@ typedef struct ucs_config_field {
 } ucs_config_field_t;
 
 
-typedef struct {
+typedef struct ucs_config_key_field {
```
iyastreb, line 84: this change seems redundant

src/ucs/config/parser.h
```diff
@@ -43,18 +43,30 @@ BEGIN_C_DECLS
...
 typedef struct ucs_config_parser {
-    int                      (*read) (const char *buf, void *dest, const void *arg);
-    int                      (*write)(char *buf, size_t max,
-                                      const void *src, const void *arg);
-    ucs_status_t             (*clone)(const void *src, void *dest, const void *arg);
-    void                     (*release)(void *ptr, const void *arg);
-    void                     (*help)(char *buf, size_t max, const void *arg);
-    void                     (*doc)(ucs_string_buffer_t *strb, const void *arg);
-    const void               *arg;
+    int          (*read)   (const struct ucs_config_parser *self,
```
iyastreb, line 46: I still struggle to understand what is exactly the reason of passing self pointer?

src/ucs/config/parser.h
```diff
@@ -43,18 +43,30 @@ BEGIN_C_DECLS
...
+                            ucs_string_buffer_t *strb);
+    const void              *arg;
 } ucs_config_parser_t;
 
+/**
+ * Passed as `arg` to the parser to handle arrays of allowed values.
+ */
+typedef struct ucs_config_allowed_values {
+    const char   **values; /**< Pointer to allowed values array */
+    unsigned int count;    /**< Number of elements in the array */
```
iyastreb, line 65:
```suggestion
    unsigned count;    /**< Number of elements in the array */
```

src/ucs/config/parser.h
```diff
@@ -43,18 +43,30 @@ BEGIN_C_DECLS
...
+                            size_t max);
+    void         (*doc)    (const struct ucs_config_parser *self,
+                            ucs_string_buffer_t *strb);
+    const void              *arg;
 } ucs_config_parser_t;
 
+/**
+ * Passed as `arg` to the parser to handle arrays of allowed values.
+ */
+typedef struct ucs_config_allowed_values {
```
iyastreb, line 63:
```suggestion
typedef struct {
```

src/ucm/api/ucm.h
```diff
@@ -66,7 +66,6 @@ typedef enum ucm_mmap_hook_mode {
     UCM_MMAP_HOOK_NONE,
     UCM_MMAP_HOOK_RELOC,
     UCM_MMAP_HOOK_BISTRO,
-    UCM_MMAP_HOOK_LAST
```
iyastreb, line 69: I'm not sure we need to remove all those LAST fields. Does it harm if we keep them?
If not, I would keep them and reduce the PR size

src/ucs/config/parser.c
```diff
@@ -339,11 +356,13 @@ int ucs_config_sprintf_on_off_auto(char *buf, size_t max,
...
 
-int ucs_config_sscanf_enum(const char *buf, void *dest, const void *arg)
+int ucs_config_sscanf_enum(const ucs_config_parser_t *self, const char *buf,
+                           void *dest)
 {
-    int i;
-
-    i = ucs_string_find_in_list(buf, (const char**)arg, 0);
+    ucs_config_allowed_values_t *allowed_values = (ucs_config_allowed_values_t*)
+                                                          self->arg;
```
iyastreb, line 363: Ok, I see how self is used, but IMO it's better to revert this change with self pointer, because we can do it with plain arg pointer. And the amount of code changes is huge for not very big benefit

src/ucs/config/parser.h
```diff
@@ -43,18 +43,30 @@ BEGIN_C_DECLS
...
+    int          (*write)  (const struct ucs_config_parser *self, char *buf,
+                            size_t max, const void *src);
+    ucs_status_t (*clone)  (const struct ucs_config_parser *self,
+                            const void *src, void *dest);
+    void         (*release)(const struct ucs_config_parser *self, void *ptr);
+    void         (*help)   (const struct ucs_config_parser *self, char *buf,
+                            size_t max);
+    void         (*doc)    (const struct ucs_config_parser *self,
+                            ucs_string_buffer_t *strb);
+    const void              *arg;
```
tvegas1, line 57: do we have a case where we need `ucs_allowed_values` along with `arg`? if not we might be able to remove `self` passing?

src/ucs/config/parser.h
```diff
@@ -177,122 +189,233 @@ extern ucs_list_link_t ucs_config_global_list;
...
+ *    [UCS_HANDLE_ERROR_FREEZE]    = "freeze",
+ *    [UCS_HANDLE_ERROR_DEBUG]     = "debug",
+ *    [UCS_HANDLE_ERROR_NONE]      = "none",
+ * };
+ * UCS_CONFIG_DEFINE_ALLOWED_VALUES(ucs_handle_error_modes);
+ */
+#define UCS_CONFIG_DEFINE_ALLOWED_VALUES(_arr) \
+    const ucs_config_allowed_values_t UCS_CONFIG_ALLOWED_VALUES_NAME(_arr) = { \
+        .values = (_arr), \
+        .count  = ucs_static_array_size(_arr) \
```
tvegas1, line 221: with bitfields, we cannot end-up with huge array (like with bit 32: 1<<32) because static array size is actually the number of bits set, right?

src/ucm/api/ucm.h
michal-shalev: I removed the LAST fields that were no longer used anywhere and were not needed anymore

src/ucs/config/parser.h
michal-shalev: done

src/ucs/config/parser.h
michal-shalev: done

src/ucs/config/parser.h
michal-shalev: done

src/ucs/sys/string.h
michal-shalev: outdated

src/ucs/config/parser.h
michal-shalev: The code is already written in an object-oriented style, with the behavior (function pointers) and state (`arg`) encapsulated in the `ucs_config_parser_t` structure. Passing the self pointer simply makes this design explicit and consistent. It clarifies that the functions operate on a specific instance of the parser, rather than treating the `arg` as an independent piece of data.

To be clear, I haven’t changed the structure or logic of the code - only the variable name and its placement in the function signature. This adjustment is meant to make the existing object-oriented approach clearer and explicit.

That said, I understand that this introduces additional changes, and I agree it's better to split this into a separate PR to keep the focus here.

src/ucs/config/parser.c
michal-shalev: The code is already written in an object-oriented style, with the behavior (function pointers) and state (`arg`) encapsulated in the `ucs_config_parser_t` structure. Passing the self pointer simply makes this design explicit and consistent. It clarifies that the functions operate on a specific instance of the parser, rather than treating the `arg` as an independent piece of data.

To be clear, I haven’t changed the structure or logic of the code - only the variable name and its placement in the function signature. This adjustment is meant to make the existing object-oriented approach clearer and explicit.

That said, I understand that this introduces additional changes, and I agree it's better to split this into a separate PR to keep the focus here.

src/ucs/config/parser.h
michal-shalev: The code is already written in an object-oriented style, with the behavior (function pointers) and state (`arg`) encapsulated in the `ucs_config_parser_t` structure. Passing the self pointer simply makes this design explicit and consistent. It clarifies that the functions operate on a specific instance of the parser, rather than treating the `arg` as an independent piece of data.

To be clear, I haven’t changed the structure or logic of the code - only the variable name and its placement in the function signature. This adjustment is meant to make the existing object-oriented approach clearer and explicit.

That said, I understand that this introduces additional changes, and I agree it's better to split this into a separate PR to keep the focus here.

src/ucs/config/parser.h
michal-shalev: The proposed approach using will not work for cases where the array is declared as `extern` in a header file and defined elsewhere, which is common in our codebase. 

For example, arrays like `ucs_memory_type_names` are declared as `extern` in the header file `memory_type.h` and defined in the source file `memory_type.c`.
When `sizeof` is used in a file where only the extern declaration is visible, the compiler treats the array as an incomplete type, resulting in the following error:
`error: invalid application of ‘sizeof’ to incomplete type ‘const char *[]’`

src/ucs/config/parser.h
michal-shalev: The size of the static array in this context is not directly tied to the full possible range of the bitfield (e.g., 1 << 32 for a 32-bit field). Instead, it depends on the indices used.

For example, in `ucp_extra_op_attr_flags_names`, the size of the array is effectively determined by the largest bit position used, rather than the total number of possible bits.

This PR ensures that bit positions corresponding to uninitialized entries (which are `NULL`) are ignored during parsing in `parser.c`, maintaining robustness even if not all bit positions are explicitly defined.

src/ucs/config/parser.c
```diff
@@ -462,20 +514,31 @@ int ucs_config_sscanf_bitmap(const char *buf, void *dest, const void *arg)
...
-int ucs_config_sprintf_bitmap(char *buf, size_t max,
-                              const void *src, const void *arg)
+int ucs_config_sprintf_bitmap(const ucs_config_parser_t *self, char *buf,
+                              size_t max, const void *src)
 {
-    ucs_flags_str(buf, max, *((uint64_t*)src), (const char**)arg);
+    ucs_config_allowed_values_t *allowed_values = (ucs_config_allowed_values_t*)
+                                                          self->arg;
+
+    ucs_flags_str(buf, max, *((uint64_t*)src),
```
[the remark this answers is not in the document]
michal-shalev: I'm checking why it fails

src/ucs/config/parser.h
iyastreb: Ok, I see
Maybe we declare these extern vars with array size?
E.g:
```
extern const char *ucm_mmap_hook_modes[UCM_MMAP_HOOK_LAST + 1];

File: src/ucs/config/parser.c (hunk 5 above)
Comment (michal-shalev) [reply src/ucs/config/parser.c]: Outdated?

File: src/ucs/config/parser.c (hunk 6 above)
Comment (michal-shalev) [reply src/ucs/config/parser.c]: Outdated?

File: src/ucs/config/parser.c (hunk 23 above)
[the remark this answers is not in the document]
Comment (michal-shalev) [reply src/ucs/config/parser.c]: I pushed a fix for it

File: bindings/go/src/ucx/ucs_constants.c [hunk 24]
```diff
@@ -4,6 +4,7 @@
  */
 #include "ucs_constants.h"
 
+#include <ucs/config/parser.h>
```
rakhmets, line 7: `types.h`?

src/ucp/core/ucp_context.c
```diff
@@ -542,7 +535,7 @@ static ucs_config_field_t ucp_context_config_table[] = {
   {"EXTRA_OP_ATTR_FLAGS", "",
    "Additional send/receive operation flags that are added for each request"
    "in addition to what is set explicitly by the user. \n"
-   "Possible values are: no_imm_cmpl, fast_cmpl, force_imm_cmpl, multi_send.",
+   "Possible values are: fast_cmpl, multi_send.",
```
rakhmets, line 538: I think the line can be remove, it is already included in the help's output.

src/ucs/config/types.h
```diff
@@ -163,4 +160,59 @@ typedef struct ucs_log_component_config {
...
+
+/**
+ * Use this in a header file when the allowed values object is defined in a
+ * different source file. If the allowed values array is defined and used
+ * in the same file, this macro is not needed.
+ * For example:
+ * UCS_CONFIG_DECLARE_ALLOWED_VALUES(ucm_log_level_names);
+ */
+#define UCS_CONFIG_DECLARE_ALLOWED_VALUES(_arr) \
+    extern const char *(_arr)[]; \
```
rakhmets, line 185: I think we can make it `const char *const`, since all related function signatures have been changed in the PR.

src/ucs/sys/string.h
```diff
@@ -241,11 +241,12 @@ const char *ucs_str_dump_hex(const void* data, size_t length, char *buf,
  * @param  max            Size of the string.
  * @param  flags          Flags to be converted.
  * @param  str_table      Conversion table - from flag value to a string.
+ * @param  str_table_size Size of the conversion table.
  *
  * @return String that holds the representation of the given flags.
  */
-const char* ucs_flags_str(char *str, size_t max,
-                          uint64_t flags, const char **str_table);
+const char* ucs_flags_str(char *buf, size_t max, uint64_t flags,
```
rakhmets, line 248: Please replace `str` by `buf` in the description.

src/uct/ib/base/ib_md.c
```diff
@@ -1325,7 +1325,7 @@ ucs_status_t uct_ib_md_open_common(uct_ib_md_t *md,
         uct_ib_check_gpudirect_driver(
                 md, "/sys/module/nv_peer_mem/version",
                 UCS_MEMORY_TYPE_CUDA);
-                
+
```
rakhmets, line 1328: Please remove the empty line.

src/ucp/core/ucp_context.c
michal-shalev: Other `UCS_CONFIG_TYPE_BITMAP` help messages in `ucp_context_config_table` show the options as well, see `reg_whole_alloc_bitmap` for example

src/ucs/config/types.h
michal-shalev: Do you mean in `parser.h`? `const char *buf`?

src/ucs/sys/string.h
michal-shalev: Good catch, thanks, done

src/uct/ib/base/ib_md.c
michal-shalev: done

bindings/go/src/ucx/ucs_constants.c
michal-shalev: Good catch, thanks, changed to `#include <ucs/config/types.h>` and compiled with `--with-go` to verify

src/ucp/core/ucp_context.c
rakhmets: I mean the line with same information is repeated twice in the output of `ucx_info -cf` is:
```
# Additional send/receive operation flags that are added for each requestin addition to what is set explicitly by the user. 
# Possible values are: fast_cmpl, multi_send.
# 
#
# syntax:    comma-separated list of: [fast_cmpl|multi_send]
```
Some other messages have the same issue.

src/ucs/config/types.h
rakhmets: I mean it is not just an array of constant strings, but a constant-size array of constant strings.

michal-shalev:
@gleon99 @iyastreb 
After a discussion with @yosefe, he requested to move forward with this code by replacing the macro-based approach with a new data structure that maps numbers to strings. Since most arrays in the config parser are based on enums, this structure should also store the size, addressing both the sparse array issue and cases where different parts of the code need to retrieve a string based on an enum offset. 
However, since this task is low priority, it will take some time before I get back to it. Moving this back to draft for now.

PR #10309, opened 2024-11-19
UCP/RNDV/RCACHE: Don't use rcache for RNDV frag mpool

iyastreb:
## What?
This is an auxiliary PR to introduce new invalidation design in UCX. The aim is to avoid using rcache for RNDV fragment memory pool

## Why?
RNDV fragment memory pool is created on demand and remains allocated during the application lifetime. Rcache is desirable for this use case, because:
1) does not bring any benefits
2) can be harmful in case of merging with other parent regions
3) redundant memory usage for parent memh

## How?
Introduce new UCP memh flag NO_RCACHE

### DIFF
--- a/src/ucp/core/ucp_mm.c
+++ b/src/ucp/core/ucp_mm.c
@@ -677,7 +677,6 @@ static void ucp_memh_init(ucp_mem_h memh, ucp_context_h context,
                           uint8_t memh_flags, unsigned uct_flags,
                           uct_alloc_method_t method, ucs_memory_type_t mem_type)
 {
-
     memh->md_map         = 0;
     memh->inv_md_map     = 0;
     memh->uct_flags      = UCP_MM_UCT_ACCESS_FLAGS(uct_flags);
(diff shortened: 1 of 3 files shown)


src/ucp/core/ucp_mm.c
```diff
@@ -1348,9 +1351,9 @@ ucp_mpool_malloc(ucp_worker_h worker, ucs_mpool_t *mp, size_t *size_p, void **ch
     ucp_mem_h memh;
     ucs_status_t status;
 
-    status = ucp_memh_alloc(worker->context, NULL,
-                            *size_p + sizeof(*chunk_hdr), UCS_MEMORY_TYPE_HOST,
-                            UCT_MD_MEM_ACCESS_RMA, ucs_mpool_name(mp), &memh);
+    status = ucp_memh_alloc(worker->context, NULL, *size_p + sizeof(*chunk_hdr),
+                            UCS_MEMORY_TYPE_HOST, UCT_MD_MEM_ACCESS_RMA, 0,
```
yosefe, line 1355: shall we add UCP_MEMH_FLAG_NO_RCACHE here as well?

src/ucp/core/ucp_mm.c
```diff
@@ -1012,7 +1013,8 @@ ucp_memh_alloc(ucp_context_h context, void *address, size_t length,
         goto err_dealloc;
     }
 
-    status = ucp_memh_init_all_uct_reg(context, memh, uct_flags, alloc_name);
+    status = ucp_memh_init_all_uct_reg(context, memh, uct_flags, ucp_flags,
```
yosefe, line 1016: instead of passing ucp_memh_create_from_mem, IMO we need to make sure UCP_MEMH_FLAG_NO_RCACHE is present in memh->flags field, from ucp_memh_init(). then ucp_memh_init_uct_reg could check  memh->flags

src/ucp/core/ucp_mm.c
iyastreb: sure, makes sense

src/ucp/core/ucp_mm.c
iyastreb: ok

brminich (approved):

yosefe (approved):

PR #10310, opened 2024-11-19
PERF: Fix final perf report discrepancy between client and server

iyastreb:
## What?
[[RM#4173667](https://redmine.mellanox.com/issues/4173667)]
There is a discrepancy between client and server final perf result, which is especially visible with pipeline protocols, but it occurs also with many others. On the client side we measure time needed to send all the messages, and with pipeline protocol we essentially measure the time needed to send message to the remote bounce buffer. With default window size of 32 we see an initial performance spike on the client side, but with larger amount of iterations the client and server results are converging to the same value due to implicit synchronisation between send/recv.
Workaround: increase iteration count (-n 1000) or decrease window size (-O 1)

## Why?
We need to provide reliable performance result on the sender side, that would be generated after all the messages are received by the server side

## How?
The fix is to send ack message from receiver to the sender at the end of processing, to confirm that all messages were received. For now we only fix the final report, while intermediate reports may still have a discrepancy.

### DIFF
--- a/src/tools/perf/lib/ucp_tests.cc
+++ b/src/tools/perf/lib/ucp_tests.cc
@@ -725,6 +725,38 @@ public:
         wait_send_window(m_max_outstanding);
     }
 
+    bool use_ack() const
+    {
+        /* TODO: daemon does not support bi-directional flow */
+        if (m_perf.params.ucp.is_daemon_mode) {
+            return false;
+        }
+
+        return (CMD == UCX_PERF_CMD_TAG) || (CMD == UCX_PERF_CMD_TAG_SYNC) ||
+               (CMD == UCX_PERF_CMD_AM);
+    }
+
+    void send_ack(void *buffer, ucp_datatype_t datatype)
+    {
+        if (!use_ack()) {
+            return;
+        }
+
+        send(m_perf.ucp.ep, buffer, 1, datatype, 0, m_perf.ucp.remote_addr,
+             m_perf.ucp.rkey, false);
+        wait_send_window(m_max_outstanding);
+    }
+
+    void recv_ack(void *buffer, ucp_datatype_t datatype)
+    {
+        if (!use_ack()) {
+            return;
+        }
+
+        recv(m_perf.ucp.worker, m_perf.ucp.ep, buffer, 1, datatype, 0);
+        wait_recv_window(m_max_outstanding);
+    }
+
     void flush()
     {
         if (m_perf.params.flags & UCX_PERF_TEST_FLAG_FLUSH_EP) {
@@ -880,6 +912,7 @@ public:
 
             wait_last_iter(recv_buffer, send_length);
             wait_recv_window(m_max_outstanding);
+            send_ack(send_buffer, send_datatype);
         } else if (my_index == 1) {
             UCX_PERF_TEST_FOREACH(&m_perf) {
                 send(ep, send_buffer, send_length, send_datatype, sn,
@@ -889,6 +922,7 @@ public:
             }
 
             send_last_iter(ep, send_buffer, send_length, remote_addr, rkey);
+            recv_ack(recv_buffer, recv_datatype);
         }
 
         flush();


src/tools/perf/lib/ucp_tests.cc
```diff
@@ -725,6 +725,39 @@ class ucp_perf_test_runner {
         wait_send_window(m_max_outstanding);
     }
 
+    bool use_ack() const
+    {
+        /* TODO: daemon does not support bi-directional flow */
+        if (m_perf.params.ucp.is_daemon_mode) {
```
tvegas1, line 731: I am assuming this boolean is always valid on both ends of the perftest in daemon mode, is that correct?

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -725,6 +725,39 @@ class ucp_perf_test_runner {
...
+    }
+
+    void send_ack(ucp_ep_h ep, void *buffer, ucp_datatype_t datatype,
+                  uint64_t remote_addr, ucp_rkey_h rkey)
+    {
+        if (!use_ack()) {
+            return;
+        }
+
+        send(ep, buffer, 1, datatype, 0, remote_addr, rkey, false);
```
tvegas1, line 746: shall we also progress this side with maybe `wait_send_window()`, in case progress would at some point happen to be needed, or the flush line 929 would be covering it?

tvegas1 (commented):
I guess that using oob barriers to synchronize both processes was not possible, maybe for timing reasons? Because otherwise calling `ucx_perf_get_time()` after `ucx_perf_*()` could have achieved similar outcome?

src/tools/perf/lib/ucp_tests.cc
iyastreb: yes, it's valid on both ends. In general the test run config is shared between sender and receiver

src/tools/perf/lib/ucp_tests.cc
iyastreb: I had `wait` initially but then removed it, because flush must be doing the job
But I can add it back, it does not harm anyway

iyastreb:
> I guess that using oob barriers to synchronize both processes was not possible, maybe for timing reasons? Because otherwise calling `ucx_perf_get_time()` after `ucx_perf_*()` could have achieved similar outcome?

Yes, this was my first proposal, to implement this sync up in 1 liner, exactly the way you said
We discussed this internally and concluded that sending such a message over OOB channel might introduce a bigger latency. Therefore the preferred approach is to send a dedicated sync up message over BW lane.

Another reason is that in the future we may want to reuse this sync up message to fix also intermediate reports. So that receiver would send this message periodically to sender, and sender would report these results. Currently there are some difficulties with implementing this change:
1) Receiver does not always accurately calculates intermediate performance. Eg in case of AM we just measure the latency of async `recv` call, not the latency of data received. For this we need to update performance stats in the payload receive callback.
2) Currently we calculate latency percentile stats both on sender and receiver, and normally we should also propagate this data from receiver to the sender.
So we need to elaborate a bit more on intermediate reports, but we can reuse this infra for that purpose

tvegas1 (dismissed):

brminich:
the error looks relevant:
```
[ RUN      ] dcx/test_ucp_perf.envelope/7 <dc_x/tag_mr_sync>
[swx-rdmz-ucx-gpu-01:49914:p-0:62475]  ucp_tests.cc:396  Assertion `m_sends_outstanding >= 0' failed
```

src/tools/perf/lib/ucp_tests.cc
iyastreb: added.
It was actually a reason of CI test failure

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -725,6 +725,40 @@ class ucp_perf_test_runner {
...
+    }
+
+    void recv_ack(ucp_worker_h worker, ucp_ep_h ep, void *buffer,
+                  ucp_datatype_t datatype)
+    {
+        if (!use_ack()) {
+            return;
+        }
+
+        flush();
```
brminich, line 757: why is flush needed?

iyastreb: Right, normally we shouldn't need it, as we wait for receiver completion anyway, so removing it

brminich (approved):

tvegas1 (approved):

src/tools/perf/lib/ucp_tests.cc
```diff
@@ -880,6 +913,7 @@ class ucp_perf_test_runner {
 
             wait_last_iter(recv_buffer, send_length);
             wait_recv_window(m_max_outstanding);
+            send_ack(ep, send_buffer, send_datatype, remote_addr, rkey);
```
yosefe, line 916: maybe send_ack/recv_ack could take ep, remote_addr, rkey from m_perf instead of passing them as params?

iyastreb: ok

yosefe (approved):

brminich (approved):

tvegas1 (approved):

PR #10311, opened 2024-11-20
UCT/GDRCOPY: rcache lookup overhead config variable

ivankochin:
## What?
Create config variable for gdrcopy lookup overhead

## Why?
Now this overhead is mixed together with latency variable

### DIFF
--- a/config/ucx.conf
+++ b/config/ucx.conf
@@ -3,10 +3,9 @@ CPU model=Grace
 UCX_REG_NONBLOCK_MEM_TYPES=host,cuda-managed
 UCX_IB_ODP_MEM_TYPES=host,cuda-managed
 UCX_IB_MLX5_DEVX_OBJECTS=
-UCX_GDR_COPY_BW=0MBs,get_dedicated:30GBs,put_dedicated:30GBs
-# Real latency is around 30ns, rest is gdrcopy rcache overhead
-# TODO: Add gdrcopy rcache overhead as separate performance graph node
-UCX_GDR_COPY_LAT=200ns
+UCX_GDR_COPY_BW=0MBs,get_dedicated:30GBs,put_dedicated:30GBsn
+UCX_GDR_COPY_LAT=30ns
+UCX_GDR_COPY_RCACHE_OVERHEAD=170ns
 UCX_DISTANCE_BW=auto,sys:16500MBs
 UCX_CUDA_COPY_ASYNC_MEM_TYPE=cuda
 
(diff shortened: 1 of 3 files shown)


src/uct/cuda/gdr_copy/gdr_copy_iface.c
```diff
@@ -46,6 +46,10 @@ static ucs_config_field_t uct_gdr_copy_iface_config_table[] = {
          ucs_offsetof(uct_gdr_copy_iface_config_t, put_latency)},
         {NULL})},
 
+    {"RCACHE_OVERHEAD", "250ns",
```
rakhmets, line 49: I think it should be in `uct_gdr_copy_md_config_table` right after `UCX_GDR_COPY_RCACHE`.
And, perhaps `md->reg_cost` should be initialized using the overhead, if rcache is enabled.

config/ucx.conf
```diff
@@ -3,10 +3,9 @@ CPU model=Grace
 UCX_REG_NONBLOCK_MEM_TYPES=host,cuda-managed
 UCX_IB_ODP_MEM_TYPES=host,cuda-managed
 UCX_IB_MLX5_DEVX_OBJECTS=
-UCX_GDR_COPY_BW=0MBs,get_dedicated:30GBs,put_dedicated:30GBs
-# Real latency is around 30ns, rest is gdrcopy rcache overhead
-# TODO: Add gdrcopy rcache overhead as separate performance graph node
-UCX_GDR_COPY_LAT=200ns
+UCX_GDR_COPY_BW=0MBs,get_dedicated:30GBs,put_dedicated:30GBsn
```
rakhmets, line 6: Typo
```suggestion
UCX_GDR_COPY_BW=0MBs,get_dedicated:30GBs,put_dedicated:30GBs
```

src/uct/cuda/gdr_copy/gdr_copy_iface.h
```diff
@@ -20,6 +20,7 @@ typedef struct uct_gdr_copy_iface {
         uct_ppn_bandwidth_t   put_bw;
         ucs_linear_func_t     get_latency;
         ucs_linear_func_t     put_latency;
+        double                rcache_ovh;
```
rakhmets, line 23: I think there is no need to shorten the word.

src/uct/cuda/gdr_copy/gdr_copy_iface.c
ivankochin: Agree with you. We discussed that with @brminich and came to a conclusion that it should be different `md_attr` field similar to `reg_cost` but `lookup_cost`. Also it seems like `ucp_proto_init_add_memreg_time` doesn't differentiate cacheable and non-cacheable MDs and adds UCP rcache overhead even for GDRcopy which is wrong. Also we need to somehow consider that overhead separately for `memtype_ep` operation if they used to do D2H and H2D copies which extends the work that need to be done significantly.

PR #10312, opened 2024-11-21
BINDINGS/GO/PERF: Implement flag.Value for UcsMemoryType

Artemy-Mellanox:
## What?
Implement `flag.Value` interface for `ucx.UcsMemoryType` and use `flag.Var` function to set `-m` argument.

## Why?
For enums, `flag.Value`/`flag.Var` is the better choice because it promotes good software design principles like type safety, reusability, and clean separation of concerns.

### DIFF
--- a/bindings/go/src/ucx/ucs_constants.go
+++ b/bindings/go/src/ucx/ucs_constants.go
@@ -5,11 +5,13 @@
 
 package ucx
 
+//#include "ucs_constants.h"
 //#include <ucp/api/ucp.h>
-//static inline const char* ucxgo_get_ucs_mem_type_name(ucs_memory_type_t idx) {
-//    return ucs_memory_type_names[idx];
-//}
 import "C"
+import (
+	"errors"
+	"unsafe"
+)
 
 type UcsThreadMode int
 
@@ -32,6 +34,16 @@ const (
 
 func (m UcsMemoryType) String() string {
 	return C.GoString(C.ucxgo_get_ucs_mem_type_name(C.ucs_memory_type_t(m)))
+}
+
+func (m *UcsMemoryType) Set (value string) error {
+	cValue := unsafe.Pointer(C.CString(value))
+	res := C.ucxgo_parse_ucs_mem_type_name(cValue)
+	if res == -1 {
+		return errors.New("memory type can be either host or cuda")
+	}
+	*m = UcsMemoryType(res)
+	return nil
 }
 
 // Checks whether context's memory type mask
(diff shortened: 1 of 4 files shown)


bindings/go/src/ucx/memory.go
```diff
@@ -67,3 +72,12 @@ func (m *UcpMemory) Close() error {
 
 	return nil
 }
+
+func (mt *UcsMemoryType) Set (value string) error {
+	switch strings.ToLower(value) {
+	case "host": *mt = UCS_MEMORY_TYPE_HOST
+	case "cuda": *mt = UCS_MEMORY_TYPE_CUDA
+	default: return errors.New("memory type can be host or cuda")
```
tvegas1, line 80: can _either_ be

tvegas1 (dismissed):

bindings/go/src/ucx/memory.go
```diff
@@ -67,3 +72,12 @@ func (m *UcpMemory) Close() error {
 
 	return nil
 }
+
+func (mt *UcsMemoryType) Set (value string) error {
```
yosefe, line 76: does it mean that ucx go bindings set memory type by string instead of value?
IMO string->Value logic should be in perftest, not in go bindings.

Artemy-Mellanox: in golang interface can't be implemented on type from other package, though may created wrapper for mem type in perftest
on the other hand value -> string logic already implemented in go bindings, https://github.com/openucx/ucx/blob/7195d0473a38e329d0a1d831f37dc5ad3b1837da/bindings/go/src/ucx/ucs_constants.go#L33  and in those languages it's convenient to implement from/to string conversion along with the object

bindings/go/src/ucx/ucs_constants.go
```diff
@@ -34,6 +36,16 @@ func (m UcsMemoryType) String() string {
...
 
+func (m *UcsMemoryType) Set (value string) error {
+	cValue := unsafe.Pointer(C.CString(value))
+	res := C.ucxgo_parse_ucs_mem_type_name(cValue)
+	if res == -1 {
+		return errors.New("memory type can be either host or cuda")
+	}
+	*m = UcsMemoryType(res)
+	return nil
+}
```
yosefe, line 47: does it mean we do string search for each send operation?

Artemy-Mellanox: no, when converting from string which we do when parsing command-line args

yosefe (approved):

PR #10319, opened 2024-11-24
UCP/TEST/TAG: Fix the limit on number of endpoint lanes

yosefe:
## Why
Fix the following CI failure:

```
$ taskset -c 2,3 make -C test/gtest test GTEST_FILTER=all_rcdc/test_ucp_tag_offload_multi.recv_from_multi/0
...
[     INFO ] Using random seed of 4577
Note: Google Test filter = all_rcdc/test_ucp_tag_offload_multi.recv_from_multi/0
[==========] Running 1 test from 1 test suite.
[----------] 1 test from all_rcdc/test_ucp_tag_offload_multi
[ RUN      ] all_rcdc/test_ucp_tag_offload_multi.recv_from_multi/0 <rc,dc>
/.autodirect/rdmzsysgwork/yosefe/ucx/test/gtest/ucp/test_ucp_tag_offload.cc:556: Failure
Expected equality of these values:
  0u
    Which is: 0
  ((&receiver().worker()->tm.offload.tag_hash)->size)
    Which is: 1
/.autodirect/rdmzsysgwork/yosefe/ucx/test/gtest/ucp/test_ucp_tag_offload.cc:562: Failure
Expected equality of these values:
  1u
    Which is: 1
  ((&receiver().worker()->tm.offload.tag_hash)->size)
    Which is: 2
[  FAILED  ] all_rcdc/test_ucp_tag_offload_multi.recv_from_multi/0, where GetParam() = rc,dc (2278 ms)
[----------] 1 test from all_rcdc/test_ucp_tag_offload_multi (2278 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (2280 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] all_rcdc/test_ucp_tag_offload_multi.recv_from_multi/0, where GetParam() = rc,dc

 1 FAILED TEST
```

## How
To make the tag offload multi test pass, we must limit the number of lanes on an endpoint to 1. Just limiting the number of eager and rendezvous lanes is not enough, since different lanes could be selected for eager and rendezvous protocols.
We add a new configuration parameter to also limit the total lanes on and endpoint.

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -230,6 +230,10 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Maximal number of devices on which a RMA operation may be executed in parallel",
    ucs_offsetof(ucp_context_config_t, max_rma_lanes), UCS_CONFIG_TYPE_UINT},
 
+  {"EP_MAX_LANES", UCS_PP_MAKE_STRING(UCP_MAX_LANES),
+   "Maximal number of lanes per endpoint",
+   ucs_offsetof(ucp_context_config_t, ep_max_lanes), UCS_CONFIG_TYPE_UINT},
+
   {"MIN_RNDV_CHUNK_SIZE", "16k",
    "Minimum chunk size to split the message sent with rendezvous protocol on\n"
    "multiple rails. Must be greater than 0.",
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -122,6 +122,8 @@ typedef struct ucp_context_config {
     unsigned                               max_rndv_lanes;
     /** RMA multi-lane support */
     unsigned                               max_rma_lanes;
+    /** Limit number of lanes for an endpoint */
+    unsigned                               ep_max_lanes;
     /** Minimum allowed chunk size when splitting rndv message over multiple
      *  lanes */
     size_t                                 min_rndv_chunk_size;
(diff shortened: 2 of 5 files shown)


roiedanino (approved):

src/ucp/core/ucp_context.c
```diff
@@ -230,6 +230,10 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Maximal number of devices on which a RMA operation may be executed in parallel",
    ucs_offsetof(ucp_context_config_t, max_rma_lanes), UCS_CONFIG_TYPE_UINT},
 
+  {"EP_MAX_LANES", UCS_PP_MAKE_STRING(UCP_MAX_LANES),
+   "Maximal number of lanes per endpoint",
+   ucs_offsetof(ucp_context_config_t, ep_max_lanes), UCS_CONFIG_TYPE_UINT},
+
```
brminich, line 236: imo, better to change just test

brminich (changes requested):
I suggest #10320

src/ucp/core/ucp_context.c
yosefe: IMO it's still a problem that we don't have a way to limit number of lanes on the endpoint, like we had in proto v1

brminich: Yes, but in proto v1 we have separate variables for eager, rndv and RMA and this patch introduces a limit for all type of lanes. Will aliases work with it (if we need extra lane for KA or wireup)?

PR #10321, opened 2024-11-25
UCP/WIREUP: Consider EP priority in iface score calculation

shasson5:
## What?
Consider EP priority in iface score calculation.

## Why?
Prioritize RC over DC for promoted EPs in selection process.

## How?
1) Add is_prioritized flag to score funcion callback.
2) Set est_num_eps to small number if EP is prioritized (in order to get better latency and thus select this EP).

### DIFF
--- a/src/ucp/core/ucp_context.c
+++ b/src/ucp/core/ucp_context.c
@@ -68,6 +68,8 @@
 #define UCP_TL_AUX_SUFFIX    "aux"
 #define UCP_TL_AUX(_tl_name) _tl_name ":" UCP_TL_AUX_SUFFIX
 
+/* Factor to multiply with in order to get infinite latency */
+#define UCP_CONTEXT_INFINITE_LAT_FACTOR 100
 
 typedef enum ucp_transports_list_search_result {
     UCP_TRANSPORTS_LIST_SEARCH_RESULT_PRIMARY      = UCS_BIT(0),
@@ -545,6 +547,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    "Possible values are: no_imm_cmpl, fast_cmpl, force_imm_cmpl, multi_send.",
    ucs_offsetof(ucp_context_config_t, extra_op_attr_flags),
    UCS_CONFIG_TYPE_BITMAP(ucp_extra_op_attr_flags_names)},
+
+  {"MAX_PRIORITY_EPS", "20",
+   "Max number of prioritized endpoints. Does not affect semantics,\n"
+   "but only transport selection criteria and resulting performance.",
+   ucs_offsetof(ucp_context_config_t, max_priority_eps),
+   UCS_CONFIG_TYPE_UINT},
 
   {NULL}
 };
@@ -2585,6 +2593,28 @@ const char* ucp_context_cm_name(ucp_context_h context, ucp_rsc_index_t cm_idx)
 {
     ucs_assert(cm_idx != UCP_NULL_RESOURCE);
     return context->tl_cmpts[context->config.cm_cmpt_idxs[cm_idx]].attr.name;
+}
+
+double ucp_tl_iface_latency_with_priority(ucp_context_h context,
+                                          const ucs_linear_func_t *latency,
+                                          int is_prioritized_ep)
+{
+    unsigned num_eps = context->config.est_num_eps;
+
+    if (ucp_context_usage_tracker_enabled(context)) {
+        if (is_prioritized_ep) {
+            /* The number of priority endpoints is limited by
+             * max_priority_eps */
+            num_eps = ucs_min(num_eps, context->config.ext.max_priority_eps);
+        } else if (latency->m > UCP_PROTO_PERF_EPSILON) {
+            /* If the transport is not scalable, assume a high number of
+             * endpoints */
+            num_eps = ucs_max(num_eps, context->config.ext.max_priority_eps *
+                                               UCP_CONTEXT_INFINITE_LAT_FACTOR);
+        }
+    }
+
+    return ucs_linear_func_apply(*latency, num_eps);
 }
 
 UCS_F_CTOR void ucp_global_init(void)
--- a/src/ucp/core/ucp_context.h
+++ b/src/ucp/core/ucp_context.h
@@ -203,6 +203,8 @@ typedef struct ucp_context_config {
     double                                 rcache_overhead;
     /** UCP extra operation attributes flags */
     uint64_t                               extra_op_attr_flags;
+    /* Upper limit to the amount of prioritized endpoints */
+    unsigned                               max_priority_eps;
 } ucp_context_config_t;
 
 
@@ -602,6 +604,10 @@ const char* ucp_feature_flags_str(unsigned feature_flags, char *str,
 void ucp_memory_detect_slowpath(ucp_context_h context, const void *address,
                                 size_t length, ucs_memory_info_t *mem_info);
 
+double ucp_tl_iface_latency_with_priority(ucp_context_h context,
+                                          const ucs_linear_func_t *latency,
+                                          int is_prioritized_ep);
+
 /**
  * Calculate a small value to overcome float imprecision
  * between two float values
@@ -646,7 +652,7 @@ int ucp_is_scalable_transport(ucp_context_h context, size_t max_num_eps)
 static UCS_F_ALWAYS_INLINE double
 ucp_tl_iface_latency(ucp_context_h context, const ucs_linear_func_t *latency)
 {
-    return ucs_linear_func_apply(*latency, context->config.est_num_eps);
+    return ucp_tl_iface_latency_with_priority(context, latency, 0);
 }
 
 static UCS_F_ALWAYS_INLINE double
--- a/src/ucp/wireup/select.c
+++ b/src/ucp/wireup/select.c
@@ -607,7 +607,7 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
             }
 
             score        = criteria->calc_score(wiface, md_attr, address, ae,
-                                                criteria->arg);
+                                                0, criteria->arg);
             priority     = iface_attr->priority + ae->iface_attr.priority;
             is_reachable = 1;
 
@@ -664,18 +664,21 @@ out:
 static inline double
 ucp_wireup_tl_iface_latency(const ucp_worker_iface_t *wiface,
                             const ucp_unpacked_address_t *unpacked_addr,
-                            const ucp_address_iface_attr_t *remote_iface_attr)
+                            const ucp_address_iface_attr_t *remote_iface_attr,
+                            int is_prioritized_ep)
 {
-    ucp_context_h context = wiface->worker->context;
+    ucp_context_h context    = wiface->worker->context;
+    ucs_linear_func_t lat_v1 = {0, wiface->attr.latency.m};
     double local_lat, lat_lossy;
 
     if (unpacked_addr->addr_version == UCP_OBJECT_VERSION_V1) {
         local_lat = ucp_wireup_iface_lat_distance_v1(wiface);
         /* Address v1 contains just latency overhead */
-        return ((local_lat + remote_iface_attr->lat_ovh) / 2) +
-               (wiface->attr.latency.m * context->config.est_num_eps);
+        lat_v1.c  = (local_lat + remote_iface_attr->lat_ovh) / 2;
+        return ucp_tl_iface_latency_with_priority(context, &lat_v1,
+                                                  is_prioritized_ep);
     } else {
-        local_lat = ucp_wireup_iface_lat_distance_v2(wiface);
+        local_lat = ucp_wireup_iface_lat_distance_v2(wiface, is_prioritized_ep);
         /* FP8 is a lossy compression method, so in order to create a symmetric
          * calculation we pack/unpack the local latency as well */
         lat_lossy = ucp_wireup_fp8_pack_unpack_latency(local_lat);
@@ -988,7 +991,7 @@ static double ucp_wireup_rma_score_func(const ucp_worker_iface_t *wiface,
                                         const uct_md_attr_v2_t *md_attr,
                                         const ucp_unpacked_address_t *unpacked_addr,
                                         const ucp_address_entry_t *remote_addr,
-                                        void *arg)
+                                        int is_prioritized_ep, void *arg)
 {
     /* best for 4k messages */
     double local_bw;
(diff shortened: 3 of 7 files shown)


src/ucp/core/ucp_context.h
```diff
@@ -643,10 +645,27 @@ int ucp_is_scalable_transport(ucp_context_h context, size_t max_num_eps)
     return (max_num_eps >= (size_t)context->config.est_num_eps);
 }
 
+static UCS_F_ALWAYS_INLINE double
+ucp_context_get_num_eps(ucp_context_h context, int is_prioritized)
```
evgeny-leksikov, line 649: name is a bit confusing, maybe `ucp_context_get_est_num_eps`

src/ucp/core/ucp_ep.h
```diff
@@ -166,8 +166,11 @@ enum {
     UCP_EP_INIT_CREATE_AM_LANE_ONLY    = UCS_BIT(8),  /**< Endpoint requires an AM lane only */
     UCP_EP_INIT_KA_FROM_EXIST_LANES    = UCS_BIT(9),  /**< Use only existing lanes to create
                                                            keepalive lane */
-    UCP_EP_INIT_ALLOW_AM_AUX_TL        = UCS_BIT(10)  /**< Endpoint allows selecting of auxiliary
+    UCP_EP_INIT_ALLOW_AM_AUX_TL        = UCS_BIT(10), /**< Endpoint allows selecting of auxiliary
                                                            transports for AM lane */
+    UCP_EP_INIT_FLAG_PRIORITIZED       = UCS_BIT(11)  /**< Endpoint is prioritized due to a
```
evgeny-leksikov, line 171: whre is this flag set?

src/ucp/wireup/select.c
```diff
@@ -664,18 +665,22 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
 static inline double
 ucp_wireup_tl_iface_latency(const ucp_worker_iface_t *wiface,
                             const ucp_unpacked_address_t *unpacked_addr,
-                            const ucp_address_iface_attr_t *remote_iface_attr)
+                            const ucp_address_iface_attr_t *remote_iface_attr,
+                            unsigned ep_init_flags)
```
evgeny-leksikov, line 669: other flags are not used here, maybe just `int prioritized`, the same for all iface related functions

shasson5: wouldn't it be to specific? if we'd want to pass more EP information in the future

src/ucp/core/ucp_ep.h
shasson5: on a separate PR

src/ucp/wireup/select.c
evgeny-leksikov: specific flag improves readability, overwise it's not clear for which case it's used, better extend usage when needed, probably never :)

src/ucp/core/ucp_ep.h
evgeny-leksikov: can share the link?

shasson5: it's not yet ready, will share when it's opened

src/ucp/wireup/select.c
```diff
@@ -606,10 +606,13 @@ static UCS_F_NOINLINE ucs_status_t ucp_wireup_select_transport(
                 continue;
             }
 
-            score        = criteria->calc_score(wiface, md_attr, address, ae,
-                                                criteria->arg);
-            priority     = iface_attr->priority + ae->iface_attr.priority;
-            is_reachable = 1;
+            is_prioritized = select_params->ep_init_flags &
+                             UCP_EP_INIT_FLAG_PRIORITIZED;
```
evgeny-leksikov, line 610: minor: the variable looks redundant to me or it's initialization should be moved before top the level loop

src/ucp/core/ucp_ep.h
evgeny-leksikov: without usage it's not clear why do we need another "est_num_eps" parameter. Also could pls point me to
> Set latency multiplier to zero if EP_PROMOTED is set (because num_eps should be ignored).

shasson5: the extra est_num_eps is for promoted EPs latency calculation (because we need to get better latency so cannot use the regular est_num_eps in case it's big).
The comment abount the multiplier is not relevant anymore, updated the description.

evgeny-leksikov: lets move this flag to next PR, the main idea in this PR is to add new parameter which affects latency calculation, the rest is uncler without usage

evgeny-leksikov (dismissed):
LGTM, the minor comment - mybe rename `is_prioritized` flag to smth like `prioritized_ep` to avoid wrong association with iface priority and please update PR title/description and the commit message on squash because current one is not relevant.

evgeny-leksikov (approved):

src/ucp/wireup/wireup.c
```diff
@@ -2075,15 +2075,16 @@ double ucp_wireup_iface_lat_distance_v1(const ucp_worker_iface_t *wiface)
         wiface->attr.latency.c;
 }
 
-double ucp_wireup_iface_lat_distance_v2(const ucp_worker_iface_t *wiface)
+double ucp_wireup_iface_lat_distance_v2(const ucp_worker_iface_t *wiface,
```
yosefe, line 2078: i would expect that wiface object will have enough information to know how many endpoints are expected to be created on it and whether it is prioritized, so passing prioritized_ep to this function should not be needed

src/ucp/core/ucp_context.c
```diff
@@ -546,6 +546,12 @@ static ucs_config_field_t ucp_context_config_table[] = {
    ucs_offsetof(ucp_context_config_t, extra_op_attr_flags),
    UCS_CONFIG_TYPE_BITMAP(ucp_extra_op_attr_flags_names)},
 
+  {"PRIO_EST_NUM_EPS", "20",
```
yosefe, line 549: why is it an "estimation" ? 
Maybe it is an upper bound on how many endpoints would be created on specific transport?

src/ucp/core/ucp_context.h
```diff
@@ -643,10 +645,27 @@ int ucp_is_scalable_transport(ucp_context_h context, size_t max_num_eps)
     return (max_num_eps >= (size_t)context->config.est_num_eps);
 }
 
+static UCS_F_ALWAYS_INLINE double
+ucp_context_get_est_num_eps(ucp_context_h context, int is_prioritized)
+{
+    return is_prioritized ? context->config.ext.prio_est_num_eps :
```
yosefe, line 651: should it be ucs_min(context->config.ext.prio_est_num_eps, context->config.est_num_eps) - in case user passes a smaller value in context->config.est_num_eps ?

brminich: then it will be set to 1 if user did not specify `ESTIMATED_NUM_EPS` (afaiu, this is the case for many non-mpi users)

shasson5: maybe we can limit the condition to scaled scenarios:
```
    return (context->config.est_num_eps > MIN_SCALED_NUM_EPS) && is_prioritized ? context->config.ext.prio_est_num_eps :
                            context->config.est_num_eps;
```

src/ucp/wireup/wireup.c
yosefe: isn't the decision on how many UCT endpoints to create of each transport depends on the transport type (i.e RC or DC)?

src/ucp/core/ucp_context.h
yosefe: if the user did not specify ESTIMATED_NUM_EPS we can end up estimating the latency for "prioritized" endpoints as higher than "regular" endpoints and IMO it's not what we want

src/ucp/wireup/wireup.c
shasson5: no, for example:
if EST_NUM_EPS=100 than:
1) RC should use NUM_EPS=100 for non prioritized EPs
2) RC should use NUM_EPS=20 for prioritized EPs

src/ucp/core/ucp_context.h
shasson5: how about:
```
static UCS_F_ALWAYS_INLINE double
ucp_context_get_est_num_eps(ucp_context_h context, int is_prioritized)
{
    return (context->config.est_num_eps > context->config.ext.prio_max_num_eps) && is_prioritized ? context->config.ext.prio_max_num_eps :
                            context->config.est_num_eps;
}
```

src/ucp/core/ucp_context.c
evgeny-leksikov: what if the specific transport is only available? it seems like we need new formula for lanes selection and probably new UCT capabilities for this.

src/ucp/core/ucp_context.h
yosefe: i think it's the same as 
```c
is_prioritized ? ucs_min(context->config.est_num_eps, 
                         context->config.ext.prio_max_num_eps) :
                         context->config.est_num_eps
```

shasson5:
@yosefe please review

src/ucp/wireup/select.c
```diff
@@ -673,9 +674,10 @@ ucp_wireup_tl_iface_latency(const ucp_worker_iface_t *wiface,
         local_lat = ucp_wireup_iface_lat_distance_v1(wiface);
         /* Address v1 contains just latency overhead */
         return ((local_lat + remote_iface_attr->lat_ovh) / 2) +
-               (wiface->attr.latency.m * context->config.est_num_eps);
+               (wiface->attr.latency.m *
+                ucp_context_get_est_num_eps(context, prioritized_ep));
```
yosefe, line 678: can we use here ucp_tl_iface_latency_with_priority?

src/ucp/core/ucp_context.h
```diff
@@ -643,10 +645,28 @@ int ucp_is_scalable_transport(ucp_context_h context, size_t max_num_eps)
...
+}
+
+static UCS_F_ALWAYS_INLINE double
+ucp_tl_iface_latency_with_priority(ucp_context_h context,
+                                   const ucs_linear_func_t *latency,
+                                   int is_prioritized)
+{
+    unsigned num_eps = ucp_context_get_est_num_eps(context, is_prioritized);
+
+    return ucs_linear_func_apply(*latency, num_eps);
```
yosefe, line 663: as we discussed offline, if context->config.est_num_eps is not set, and is_prioritized==0, and latency->m>0, we should return INF latency to prevent RC from being selected on large scale for non-prioritized endpoint

src/ucp/core/ucp_context.c
yosefe: pls adjust variable name - MAX_PRIORITY_EPS and field name to max_priority_eps

src/ucp/wireup/select.c
shasson5: do you mean call `ucp_tl_iface_latency_with_priority` and substuct lat.c from the result?

src/ucp/core/ucp_context.h
```diff
@@ -643,10 +645,45 @@ int ucp_is_scalable_transport(ucp_context_h context, size_t max_num_eps)
...
+        return context->config.est_num_eps;
+    }
+
+    if ((latency->m > 0) && !is_prioritized) {
+        return INFINITY;
+    }
+
+    return is_prioritized ? ucs_min(context->config.est_num_eps,
+                                    context->config.ext.max_priority_eps) :
+                            context->config.est_num_eps;
```
yosefe, line 669:
```suggestion
    if (is_prioritized) {
        return ucs_min(context->config.est_num_eps,
                                    context->config.ext.max_priority_eps);
    }
    
    return (latency->m > EPSILON) ? INFINITY : context->config.est_num_eps;
```

src/ucp/core/ucp_context.h
```diff
@@ -643,10 +645,45 @@ int ucp_is_scalable_transport(ucp_context_h context, size_t max_num_eps)
...
+}
+
+static UCS_F_ALWAYS_INLINE double
+ucp_tl_iface_latency_with_priority(ucp_context_h context,
+                                   const ucs_linear_func_t *latency,
+                                   int is_prioritized)
+{
+    unsigned num_eps = ucp_context_get_est_num_eps(context, latency,
```
yosefe, line 677: we could inline the code of ucp_context_get_est_num_eps function inside ucp_tl_iface_latency_with_priority. the only place where it's used separately - ucp_wireup_tl_iface_latency - could create a local linear function struct to pass to ucp_tl_iface_latency_with_priority.
this will simplify the code and avoid multiplying latency->m by INFINITY

yosefe (commented):
maybe we could use a "safer" constant that INFINITY - such as 100x the number defined by config.ext.max_priority_eps - enough to ensure the priority endpoints would haver a better latency

src/ucp/core/ucp_context.h
```diff
@@ -43,6 +43,12 @@ enum {
     UCP_TL_RSC_FLAG_AUX = UCS_BIT(0)
 };
 
+/* Threshold to consider latency as not effected by number of endpoints */
+#define UCP_CONTEXT_EPSILON_LATENCY 1e-10
```
yosefe, line 47: can we unite this constant with UCP_PROTO_PERF_EPSILON?

src/ucp/core/ucp_context.h
```diff
@@ -651,31 +657,24 @@ ucp_context_usage_tracker_enabled(ucp_context_h context)
...
+    if (!ucp_context_usage_tracker_enabled(context)) {
+        num_eps = context->config.est_num_eps;
+    } else if (is_prioritized) {
+        num_eps = ucs_min(context->config.est_num_eps,
+                          context->config.ext.max_priority_eps);
+    } else if (latency->m > UCP_CONTEXT_EPSILON_LATENCY) {
+        num_eps = context->config.ext.max_priority_eps *
+                  UCP_CONTEXT_INFINITE_LAT_FACTOR;
+    } else {
+        num_eps = context->config.est_num_eps;
```
yosefe, line 676: looks good! pls just add a single documentation line for each of the 4 cases

src/ucp/core/ucp_context.h
```diff
@@ -643,10 +648,42 @@ int ucp_is_scalable_transport(ucp_context_h context, size_t max_num_eps)
...
+                                   const ucs_linear_func_t *latency,
+                                   int is_prioritized)
+{
+    unsigned num_eps;
+
+    if (!ucp_context_usage_tracker_enabled(context)) {
+        /* No priority EPs exist, use default value */
+        num_eps = context->config.est_num_eps;
+    } else if (is_prioritized) {
+        /* Use small latency for priority EP */
```
yosefe, line 668: /* The number of priority endpoints is limited by context->config.ext.max_priority_eps */

yosefe (approved):

yosefe:
@shasson5 pls squash

PR #10324, opened 2024-11-25
EXAMPLES: Fixed checking for memory type support.

rakhmets:
## What?
Fixed checking for memory type support.
Plus minor fixes.

### DIFF
--- a/examples/ucp_client_server.c
+++ b/examples/ucp_client_server.c
@@ -99,16 +99,16 @@ static struct {
  */
 static void usage(void);
 
-void buffer_free(ucp_dt_iov_t *iov)
+static void buffer_free(ucp_dt_iov_t *iov, size_t iov_size)
 {
     size_t idx;
 
-    for (idx = 0; idx < iov_cnt; idx++) {
+    for (idx = 0; idx < iov_size; idx++) {
         mem_type_free(iov[idx].buffer);
     }
 }
 
-int buffer_malloc(ucp_dt_iov_t *iov)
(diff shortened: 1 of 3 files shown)


tvegas1 (approved):

examples/ucp_client_server.c
```diff
@@ -99,24 +99,24 @@ static struct {
  */
 static void usage(void);
 
-void buffer_free(ucp_dt_iov_t *iov)
+static void buffer_free(ucp_dt_iov_t *iov, size_t iov_size)
 {
     size_t idx;
 
-    for (idx = 0; idx < iov_cnt; idx++) {
+    for (idx = 0; idx < iov_size; idx++) {
```
iyastreb, line 106: Since `iov` is initialized with all zeroes, maybe instead of passing iov_size we can just free if element is not NULL?
```C
static void buffer_free(ucp_dt_iov_t *iov)
{
    size_t idx;

    for (idx = 0; idx < iov_cnt; ++idx) {
        if (NULL != iov[idx].buffer) {
            mem_type_free(iov[idx].buffer);
            iov[idx].buffer = NULL;
        }
    }
}
```
Or we can make check for NULL in `mem_type_malloc` itself

rakhmets: Yes, we can. However, this option looks more complicated, and adds a condition inside the loop body.

brminich (approved):

yosefe (approved):

PR #10328, opened 2024-11-26
UCP/GTEST: Workaround KSM ODP failure

Artemy-Mellanox:
## What?
Temporary disable shifted atomic address in ODP PUT tests.

## Why?
Shifted atomic address exposes an issue in the driver ([RM 4170682](https://redmine.mellanox.com/issues/4170682)).


yosefe (commented):
WDYT to disable atomic offset with ODP? so KSM would still be exercised

src/uct/ib/base/ib_md.h
```diff
@@ -333,7 +334,10 @@ uct_ib_md_is_flush_rkey_valid(uint32_t flush_rkey) {
 
 static UCS_F_ALWAYS_INLINE uint8_t uct_ib_md_get_atomic_mr_id(uct_ib_md_t *md)
 {
-    return md->flush_rkey >> 8;
+    if (md->config.enable_atomic_offset) {
+        return md->flush_rkey >> 8;
+    }
+    return 0;
```
brminich, line 340: seems it breaks remote flush functionality  `UCT_FLUSH_FLAG_REMOTE`  (e.g. see `uct_dc_mlx5_flush_remote_rkey`)

Artemy-Mellanox: in this PR it's not a problem since param is `on` by default, but in future if we make it `off` or `auto` will need to handle this

brminich: Yes, but this PR introduces new env var, tweaking which may result in a failure.
Maybe just temprorarily disable the tests instead?

yosefe: Shall we use the existing "INDIRECT_ATOMIC" env var?

yosefe (approved):

src/uct/ib/base/ib_md.h
brminich: it still breaks remote flush functionality. Can you disabled "REMOTE_FLUSH" capability when this env var is set?

yosefe:
@Artemy-Mellanox pls squash

brminich (changes requested):
it still breaks remote flush functionality. Can you disabled "REMOTE_FLUSH" capability when this env var is set?

src/uct/ib/base/ib_md.h
```diff
@@ -333,7 +333,10 @@ uct_ib_md_is_flush_rkey_valid(uint32_t flush_rkey) {
 
 static UCS_F_ALWAYS_INLINE uint8_t uct_ib_md_get_atomic_mr_id(uct_ib_md_t *md)
 {
-    return md->flush_rkey >> 8;
+    if (md->config.enable_indirect_atomic) {
```
yosefe, line 336: BTW, why do we need this change? i would expect setting IB_INDIRECT_ATOMIC=n in the test would already disable KSM atomic, w/o additional changes in IB

brminich: currently `ext.enable_indirect_atomic` is not used at all in the code (some kind of leftover).

yosefe: i see, so maybe we should make this var affect initiator side (send atomic ops to regular MR with offset 0) and not responder side, to avoid breaking remote flush

src/uct/ib/mlx5/dc/dc_mlx5_ep.c
```diff
@@ -23,6 +23,16 @@ uct_dc_mlx5_add_flush_remote(uct_dc_mlx5_ep_t *ep)
     ep->flags |= UCT_DC_MLX5_EP_FLAG_FLUSH_REMOTE;
 }
 
+static UCS_F_ALWAYS_INLINE uint16_t
+uct_dc_mlx5_atomic_offset(uct_dc_mlx5_iface_t *iface, uct_dc_mlx5_ep_t *ep)
+{
+    if (uct_ib_iface_md(&iface->super.super.super)->config.enable_indirect_atomic) {
```
yosefe, line 29: seems wrong - we can't decide to use offset 0 without also using direct (and not atomic) part of the rkey

Artemy-Mellanox: if enable_indirect_atomic is off the atomic part of the rkey won't have an offset but may still be present to enforce strict order

yosefe: can we detect it from information provided from the target side, rather than local configuration?

yosefe: check if iface/ep address can have extra flag to indicate if atomic offset is 0, or it's as defined by atomic_mr_id

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -53,17 +54,23 @@ const uct_ib_md_t &test_ib_md::ib_md() const {
...
         EXPECT_NE(nullptr, memh->atomic_dvmr);
         EXPECT_NE(UCT_IB_INVALID_MKEY, memh->atomic_rkey);
     } else {
         EXPECT_EQ(nullptr, memh->atomic_dvmr);
-        EXPECT_EQ(UCT_IB_INVALID_MKEY, memh->atomic_rkey);
+        if (is_expected_to_have_auxiliary_key) {
+            EXPECT_NE(UCT_IB_INVALID_MKEY, memh->atomic_rkey);
+        } else {
+            EXPECT_EQ(UCT_IB_INVALID_MKEY, memh->atomic_rkey);
+        }
```
yosefe, line 73: can we move this outside the 'else' block?

yosefe (approved):

brminich (approved):

PR #10331, opened 2024-11-26
UCP/TEST: Enhance EP reconfiguration tests

shasson5:
## What?
Adding parallelism to the send-recv operation

## Why?
In order to cover more usecases and test more complicated scenarios

### DIFF
--- a/test/gtest/ucp/test_ucp_ep_reconfig.cc
+++ b/test/gtest/ucp/test_ucp_ep_reconfig.cc
@@ -15,9 +15,11 @@ extern "C" {
 
 class test_ucp_ep_reconfig : public ucp_test {
 protected:
-    using address_t = std::pair<void*, ucp_unpacked_address_t>;
-    using address_p = std::unique_ptr<address_t, void (*)(address_t*)>;
-    using limits    = std::numeric_limits<unsigned>;
+    using address_t        = std::pair<void*, ucp_unpacked_address_t>;
+    using address_p        = std::unique_ptr<address_t, void (*)(address_t*)>;
+    using limits           = std::numeric_limits<unsigned>;
+    using mem_buffer_p     = std::unique_ptr<mem_buffer>;
+    using mem_buffer_vec_t = std::vector<mem_buffer_p>;
 
     class entity : public ucp_test_base::entity {
     public:
@@ -62,7 +64,12 @@ protected:
         bool                   m_exclude_ifaces;
     };
 
-    bool is_single_transport()
+    void init_buffers(mem_buffer_vec_t &sbufs, mem_buffer_vec_t &rbufs,
+                      size_t size, bool bidirectional);
+
+    void pattern_check(const mem_buffer_vec_t &rbufs) const;
+
+    bool is_single_transport() const
     {
         return GetParam().transports.size() == 1;
     }
@@ -122,29 +129,28 @@ public:
     }
 
     void send_message(const ucp_test_base::entity &e1,
-                      const ucp_test_base::entity &e2, const mem_buffer &sbuf,
-                      const mem_buffer &rbuf, size_t msg_size,
-                      std::vector<void*> &reqs)
+                      const ucp_test_base::entity &e2, const mem_buffer *sbuf,
+                      const mem_buffer *rbuf, std::vector<void*> &reqs)
     {
         const ucp_request_param_t param = {
             .op_attr_mask = UCP_OP_ATTR_FLAG_NO_IMM_CMPL
         };
 
-        void *sreq = ucp_tag_send_nbx(e1.ep(), sbuf.ptr(), msg_size, 0, &param);
-        void *sreq_sync = ucp_tag_send_sync_nbx(e1.ep(), sbuf.ptr(), msg_size,
-                                                0, &param);
+        void *sreq = ucp_tag_send_nbx(e1.ep(), sbuf->ptr(), sbuf->size(), 0,
+                                      &param);
+        void *sreq_sync = ucp_tag_send_sync_nbx(e1.ep(), sbuf->ptr(),
+                                                sbuf->size(), 0, &param);
         reqs.insert(reqs.end(), {sreq, sreq_sync});
 
         for (unsigned iter = 0; iter < 2; iter++) {
-            void *rreq = ucp_tag_recv_nbx(e2.worker(), rbuf.ptr(), msg_size, 0,
-                                          0, &param);
+            void *rreq = ucp_tag_recv_nbx(e2.worker(), rbuf->ptr(),
+                                          rbuf->size(), 0, 0, &param);
             reqs.push_back(rreq);
         }
     }
 
     void send_recv(bool bidirectional)
     {
-        static const unsigned num_iterations = 1000;
 /* TODO: remove this when 100MB asan bug is solved */
 #ifdef __SANITIZE_ADDRESS__
         static const size_t msg_sizes[] = {8, 1024, 16384, 65536};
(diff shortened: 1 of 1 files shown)


test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -288,6 +295,36 @@ test_ucp_ep_reconfig::entity::get_address(const ucp_tl_bitmap_t &tl_bitmap) cons
...
+                                        mem_buffer_vec_t &rbufs,
+                                        mem_buffer_vec_t &o_sbufs,
+                                        mem_buffer_vec_t &o_rbufs, size_t size)
+{
+    for (unsigned i = 0; i < num_iterations; ++i) {
+        sbufs.push_back(
+                mem_buffer_p(new mem_buffer(size, UCS_MEMORY_TYPE_HOST, i)));
+        rbufs.push_back(mem_buffer_p(
+                new mem_buffer(size, UCS_MEMORY_TYPE_HOST, ucs::rand())));
+        o_sbufs.push_back(
```
gleon99, line 308: Can use `bidirectional` here to avoid excessive allocation

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -134,29 +143,27 @@ class test_ucp_ep_reconfig : public ucp_test {
...
+            std::vector<void*> reqs;
 
-                std::vector<void*> reqs;
-                send_message(sender(), receiver(), sbuf, rbuf, msg_size, reqs);
-
-                if (bidirectional) {
-                    send_message(receiver(), sender(), o_sbuf, o_rbuf, msg_size,
-                                 reqs);
-                }
+            init_buffers(sbufs, rbufs, o_sbufs, o_rbufs, msg_size);
```
gleon99, line 149: Maybe init them outside the loop, only once?

shasson5: will be wrong as they are sent and verified on each iteration

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -288,6 +298,40 @@ test_ucp_ep_reconfig::entity::get_address(const ucp_tl_bitmap_t &tl_bitmap) cons
...
+                    new mem_buffer(size, UCS_MEMORY_TYPE_HOST, i)));
+            o_rbufs.push_back(mem_buffer_p(
+                    new mem_buffer(size, UCS_MEMORY_TYPE_HOST, ucs::rand())));
+        }
+    }
+}
+
+void test_ucp_ep_reconfig::pattern_check(const mem_buffer_vec_t &rbufs,
+                                         const mem_buffer_vec_t &o_rbufs,
+                                         size_t size, bool bidirectional) const
```
yosefe, line 324: do we really have to pass the size? pattern_check used to be called w/o size param, guess size was taken from the buffer

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -103,60 +113,60 @@ class test_ucp_ep_reconfig : public ucp_test {
...
-                requests_wait(reqs);
-                rbuf.pattern_check(i);
+            for (unsigned i = 0; i < num_iterations; ++i) {
+                send_message(sender(), receiver(), sbufs[i].get(),
+                             rbufs[i].get(), msg_size, reqs);
 
                 if (bidirectional) {
-                    o_rbuf.pattern_check(i);
+                    send_message(receiver(), sender(), o_sbufs[i].get(),
+                                 o_rbufs[i].get(), msg_size, reqs);
```
yosefe, line 160: maybe instead of using separate vectors for the other-way buffers, we can allocate 2x the number of buffers for bidirectional tests, and here use something like 
```c
send_message(receiver(), sender(), sbufs[i + num_iterations].get(),
             o_rbufs[i + num_iterations].get(), msg_size, reqs);
```

test/gtest/ucp/test_ucp_ep_reconfig.cc
shasson5: the buffers are initialized with max_buffer_size:
```
        init_buffers(sbufs, rbufs, o_sbufs, o_rbufs,
                     msg_sizes[ucs_static_array_size(msg_sizes)-1],
                     bidirectional);
```
so need to pass real size from outside

yosefe: we could move buffers declaration and initialization inside the `for (auto msg_size : msg_sizes) {` loop, as it used to be, then the size would be exact and not need to take the last size in the array

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -288,6 +298,28 @@ test_ucp_ep_reconfig::entity::get_address(const ucp_tl_bitmap_t &tl_bitmap) cons
     return address;
 }
 
+void test_ucp_ep_reconfig::init_buffers(mem_buffer_vec_t &sbufs,
+                                        mem_buffer_vec_t &rbufs, size_t size,
+                                        bool bidirectional)
+{
+    auto num_bufs = bidirectional ? 2 * num_iterations : num_iterations;
```
yosefe, line 305: add () around * expression

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -288,6 +298,28 @@ test_ucp_ep_reconfig::entity::get_address(const ucp_tl_bitmap_t &tl_bitmap) cons
...
+                mem_buffer_p(new mem_buffer(size, UCS_MEMORY_TYPE_HOST, i)));
+        rbufs.push_back(mem_buffer_p(
+                new mem_buffer(size, UCS_MEMORY_TYPE_HOST, ucs::rand())));
+    }
+}
+
+void test_ucp_ep_reconfig::pattern_check(const mem_buffer_vec_t &rbufs,
+                                         size_t size, bool bidirectional) const
+{
+    for (unsigned i = 0; i < rbufs.size(); ++i) {
```
yosefe, line 318: for (auto rbuf: rbufs) { .. }

test/gtest/ucp/test_ucp_ep_reconfig.cc
shasson5: ok, but notice it colides with https://github.com/openucx/ucx/pull/10331#discussion_r1865682440

test/gtest/ucp/test_ucp_ep_reconfig.cc
shasson5: this will require extra counter to replace "i"

yosefe (approved):

shasson5:
@yosefe can re-approve?

yosefe:
Failure seems relevant:
```
2024-12-12T19:48:56.4535660Z [ RUN      ] shm_gga/test_ucp_ep_reconfig.resolve_remote_id/0 <shm,gga>
2024-12-12T19:48:56.7480844Z /__w/1/s/contrib/../test/gtest/common/mem_buffer.cc:345: Failure
2024-12-12T19:48:56.7482718Z Pattern check failed at 0x567f310 offset 0: Expected: 0x0 Actual: 0x1
2024-12-12T19:48:56.7616320Z [  FAILED  ] shm_gga/test_ucp_ep_reconfig.resolve_remote_id/0, where GetParam() = shm,gga (308 ms)
```

yosefe (commented):
why needed to add a tag? tag api should already be strict order

shasson5:
tag_sync request is waiting for remote_id, so it may change order with remote side

yosefe:
> tag_sync request is waiting for remote_id, so until it is sent, the next tag_send iteration may have already happened

It sounds like a bug:
1. In MPI specification, it's is defined that tag send operations should happen in order, regardless if it's MPI_Isend (tag_send) or MPI_Issend (tag_send_sync)
2. The subsequent tag_send operation should be placed on the pending queue of the wireup_ep that is created as part of remote_id resolving, so it should not be sent before the tag_send_sync

test/gtest/ucp/test_ucp_ep_reconfig.cc
```diff
@@ -130,24 +130,22 @@ class test_ucp_ep_reconfig : public ucp_test {
...
         const ucp_request_param_t param = {
             .op_attr_mask = UCP_OP_ATTR_FLAG_NO_IMM_CMPL
         };
 
-        void *sreq      = ucp_tag_send_nbx(e1.ep(), sbuf->ptr(), sbuf->size(),
-                                           msg_index, &param);
-        void *sreq_sync = ucp_tag_send_sync_nbx(e1.ep(), sbuf->ptr(),
-                                                sbuf->size(), msg_index,
-                                                &param);
+        void *sreq = ucp_tag_send_nbx(e1.ep(), sbuf->ptr(), msg_size, 0,
```
yosefe, line 140: why need to pass msg_size and can't use sbuf->size() / rbuf->size() ?

yosefe (approved):

PR #10332, opened 2024-11-26
UCT: Derived memh

iyastreb:
## What?
This PR is a composite part of the new memory invalidation design based on the concept of memory windows.
Here we introduce changes on the UCT side.
On UCT side memory window represents a shallow copy of some existing UCT memory handle, which can be used to access the same memory region. When created, the memory window inherits the original memh access flags and state, and takes ownership of the indirect keys of the original memory handle. The lifetime of the memory window is bound to the original memh, and the original memh cannot be destroyed until all its memory windows are destroyed.

[Design doc](https://nvidia.sharepoint.com/:p:/r/sites/NBU-UCX/_layouts/15/doc2.aspx?sourcedoc=%7B76889907-A62B-49BC-B79B-82E09E4EF036%7D&file=Pipeline%20invalidation.pptx&action=edit&mobileredirect=true)

## Why?
The overall idea is to employ a lightweight invalidation: when failure happens we invalidate only the exposed indirect rkeys, but don't deregister the entire memory region, which is an expensive operation. Lightweight invalidation enables invalidation workflow for RNDV pipeline protocol.

## How?
1) Introduced a new option for `uct_md_mem_reg_params_t`: `UCT_MD_MEM_WINDOW`. Existing `mem_reg` call will create a memory window based on existing UCT memh
2) When memory window is requested, create a shallow copy of base UCT memh and take the ownership of its indirect memory keys.

### DIFF
--- a/src/uct/api/uct.h
+++ b/src/uct/api/uct.h
@@ -844,6 +844,17 @@ enum uct_md_mem_flags {
     UCT_MD_MEM_GVA                  = UCS_BIT(11),
 
     /**
+     * Register a memory window: a shallow copy of some existing UCT memory
+     * handle, which can be used to access the same memory region. When created,
+     * the memory window inherits the original memh access flags and state, and
+     * takes ownership of the indirect keys of the original memory handle. The
+     * lifetime of the memory window is bound to the original memh, and the
+     * original memh cannot be destroyed until all its memory windows are
+     * destroyed.
+     */
+    UCT_MD_MEM_WINDOW               = UCS_BIT(12),
+
+    /**
      * Enable local and remote access for all operations.
      */
     UCT_MD_MEM_ACCESS_ALL           = (UCT_MD_MEM_ACCESS_REMOTE_PUT |
--- a/src/uct/api/uct.h
+++ b/src/uct/api/uct.h
@@ -844,17 +844,6 @@ enum uct_md_mem_flags {
     UCT_MD_MEM_GVA                  = UCS_BIT(11),
 
     /**
-     * Register a memory window: a shallow copy of some existing UCT memory
-     * handle, which can be used to access the same memory region. When created,
-     * the memory window inherits the original memh access flags and state, and
-     * takes ownership of the indirect keys of the original memory handle. The
-     * lifetime of the memory window is bound to the original memh, and the
-     * original memh cannot be destroyed until all its memory windows are
-     * destroyed.
-     */
-    UCT_MD_MEM_WINDOW               = UCS_BIT(12),
-
-    /**
      * Enable local and remote access for all operations.
      */
     UCT_MD_MEM_ACCESS_ALL           = (UCT_MD_MEM_ACCESS_REMOTE_PUT |
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -225,7 +225,8 @@ typedef struct {
 typedef enum {
     UCT_MD_MEM_REG_FIELD_FLAGS         = UCS_BIT(0),
     UCT_MD_MEM_REG_FIELD_DMABUF_FD     = UCS_BIT(1),
-    UCT_MD_MEM_REG_FIELD_DMABUF_OFFSET = UCS_BIT(2)
+    UCT_MD_MEM_REG_FIELD_DMABUF_OFFSET = UCS_BIT(2),
+    UCT_MD_MEM_REG_FIELD_MEMH          = UCS_BIT(3)
 } uct_md_mem_reg_field_mask_t;
 
 
@@ -490,6 +491,13 @@ typedef struct uct_md_mem_reg_params {
      * dmabuf region, then this field must be omitted or set to 0.
      */
     size_t                       dmabuf_offset;
+
+    /**
+     * Represents a pointer to the existing memory handle.
+     * Used to create a memory window is (with flag @ref UCT_MD_MEM_WINDOW)
+     * based on this original memh.
+     */
+    uct_mem_h                    memh;
 } uct_md_mem_reg_params_t;
 
 
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -1027,7 +1027,12 @@ typedef enum {
     /**
      * Memory domain performs memory type related copy operations.
      */
-    UCT_MD_FLAG_MEMTYPE_COPY   = UCS_BIT(13)
+    UCT_MD_FLAG_MEMTYPE_COPY   = UCS_BIT(13),
+
+    /**
+     * Memory domain supports derived memory handle registration.
+     */
+    UCT_MD_FLAG_DERIVED        = UCS_BIT(14)
 } uct_md_flags_v2_t;
 
 
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -1032,7 +1032,7 @@ typedef enum {
     /**
      * Memory domain supports derived memory handle registration.
      */
-    UCT_MD_FLAG_DERIVED        = UCS_BIT(14)
+    UCT_MD_FLAG_REG_DERIVED    = UCS_BIT(14)
 } uct_md_flags_v2_t;
 
 
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -493,14 +493,14 @@ typedef struct uct_md_mem_reg_params {
     size_t                       dmabuf_offset;
 
     /**
-     * Represents a pointer to the existing memory handle.
-     * Used to register a derived memory handle: a shallow copy of existing UCT
-     * memory handle, which can be used to access the same memory region. When
-     * created, the derived memh inherits the original memh access flags and
-     * state. The lifetime of the derived memh is bound to the original memh,
-     * and the original memh cannot be destroyed until all its derived handles
-     * are destroyed. The derived memh cannot be used to register another
-     * derived memh.
+     * A pointer to an existing memory handle.
+     * Used to register a derived memh: a shallow copy of an existing UCT memh
+     * which can be used to access the same memory region. When created, the
+     * derived memh inherits the access flags and the state of the original
+     * memh. The lifetime of the derived memh is bound to the original memh.
+     * The original memh cannot be destroyed until all its derived handles have
+     * been destroyed. A derived memh cannot be used to register another derived
+     * memh.
      */
     uct_mem_h                    memh;
 } uct_md_mem_reg_params_t;
--- a/src/uct/api/v2/uct_v2.h
+++ b/src/uct/api/v2/uct_v2.h
@@ -494,8 +494,13 @@ typedef struct uct_md_mem_reg_params {
 
     /**
      * Represents a pointer to the existing memory handle.
-     * Used to create a memory window is (with flag @ref UCT_MD_MEM_WINDOW)
-     * based on this original memh.
+     * Used to register a derived memory handle: a shallow copy of existing UCT
+     * memory handle, which can be used to access the same memory region. When
+     * created, the derived memh inherits the original memh access flags and
+     * state. The lifetime of the derived memh is bound to the original memh,
+     * and the original memh cannot be destroyed until all its derived handles
+     * are destroyed. The derived memh cannot be used to register another
+     * derived memh.
      */
     uct_mem_h                    memh;
 } uct_md_mem_reg_params_t;
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -414,11 +414,11 @@ static ucs_status_t
 uct_cuda_ipc_mem_reg(uct_md_h md, void *address, size_t length,
                      const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
-    unsigned flags = UCT_MD_MEM_REG_FIELD_VALUE(params, flags, FIELD_FLAGS, 0);
+    uct_mem_h base = UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL);
     uct_cuda_ipc_memh_t *memh;
 
-    if (ENABLE_PARAMS_CHECK && (flags & UCT_MD_MEM_WINDOW)) {
-        ucs_error("CUDA IPC does not support memory windows");
+    if (ENABLE_PARAMS_CHECK && (base != NULL)) {
+        ucs_error("CUDA IPC does not support derived memory handles");
         return UCS_ERR_UNSUPPORTED;
     }
 
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -414,15 +414,11 @@ static ucs_status_t
 uct_cuda_ipc_mem_reg(uct_md_h md, void *address, size_t length,
                      const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
-    uct_mem_h base = (params != NULL) ?
-                        UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL) :
-                        NULL;
     uct_cuda_ipc_memh_t *memh;
 
-    if (ENABLE_PARAMS_CHECK && (base != NULL)) {
-        ucs_error("CUDA IPC does not support derived memory handles");
-        return UCS_ERR_UNSUPPORTED;
-    }
+    UCT_CHECK_PARAM((params == NULL) ||
+                    UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL) == NULL,
+                    "CUDA IPC does not support derived memory handles");
 
     memh = ucs_malloc(sizeof(*memh), "uct_cuda_ipc_memh_t");
     if (NULL == memh) {
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -414,7 +414,9 @@ static ucs_status_t
 uct_cuda_ipc_mem_reg(uct_md_h md, void *address, size_t length,
                      const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
-    uct_mem_h base = UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL);
+    uct_mem_h base = (params != NULL) ?
+                        UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL) :
+                        NULL;
     uct_cuda_ipc_memh_t *memh;
 
     if (ENABLE_PARAMS_CHECK && (base != NULL)) {
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -416,9 +416,7 @@ uct_cuda_ipc_mem_reg(uct_md_h md, void *address, size_t length,
 {
     uct_cuda_ipc_memh_t *memh;
 
-    UCT_CHECK_PARAM((params == NULL) ||
-                    UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL) == NULL,
-                    "CUDA IPC does not support derived memory handles");
+    UCT_CHECK_NOT_DERIVED_MEMH(params, "cuda_ipc_md");
 
     memh = ucs_malloc(sizeof(*memh), "uct_cuda_ipc_memh_t");
     if (NULL == memh) {
--- a/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
+++ b/src/uct/cuda/cuda_ipc/cuda_ipc_md.c
@@ -414,7 +414,13 @@ static ucs_status_t
 uct_cuda_ipc_mem_reg(uct_md_h md, void *address, size_t length,
                      const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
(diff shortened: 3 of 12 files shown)


src/uct/api/uct.h
```diff
@@ -843,6 +843,17 @@ enum uct_md_mem_flags {
...
+    /**
+     * Register a memory window: a shallow copy of some existing UCT memory
+     * handle, which can be used to access the same memory region. When created,
+     * the memory window inherits the original memh access flags and state, and
+     * takes ownership of the indirect keys of the original memory handle. The
+     * lifetime of the memory window is bound to the original memh, and the
+     * original memh cannot be destroyed until all its memory windows are
+     * destroyed.
+     */
+    UCT_MD_MEM_WINDOW               = UCS_BIT(12),
```
brminich, line 855: maybe call it 
```suggestion
    UCT_MD_MEM_DERIVED               = UCS_BIT(12),
```

src/uct/api/uct.h
```diff
@@ -843,6 +843,17 @@ enum uct_md_mem_flags {
      */
     UCT_MD_MEM_GVA                  = UCS_BIT(11),
 
+    /**
+     * Register a memory window: a shallow copy of some existing UCT memory
+     * handle, which can be used to access the same memory region. When created,
+     * the memory window inherits the original memh access flags and state, and
+     * takes ownership of the indirect keys of the original memory handle. The
```
brminich, line 850: indirect key is IB specific we should not mention it in UCT API

src/uct/api/v2/uct_v2.h
```diff
@@ -478,6 +479,13 @@ typedef struct uct_md_mem_reg_params {
      * dmabuf region, then this field must be omitted or set to 0.
      */
     size_t                       dmabuf_offset;
+
+    /**
+     * Represents a pointer to the existing memory handle.
+     * Used to create a memory window is (with flag @ref UCT_MD_MEM_WINDOW)
```
brminich, line 485:
```suggestion
     * Used to create a derived memory handle
```

src/uct/api/v2/uct_v2.h
```diff
@@ -478,6 +479,13 @@ typedef struct uct_md_mem_reg_params {
...
+     * based on this original memh.
+     */
+    uct_mem_h                    memh;
```
brminich, line 488: maybe we do not need both - the flag and memh?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -835,6 +855,43 @@ uct_ib_mlx5_devx_mem_reg_gva(uct_md_h uct_md, unsigned flags, uct_mem_h *memh_p)
...
+    uct_ib_mlx5_devx_mem_t *memh;
+    ucs_status_t status;
+
+    if (ENABLE_PARAMS_CHECK && (base == NULL)) {
+        ucs_error("base memory handle is not provided for memory window");
+        return UCS_ERR_INVALID_PARAM;
+    }
+
+    ucs_assertv(!(base->super.flags & UCT_IB_MEM_FLAG_MEM_WINDOW),
+                "memh=%p is already a memory window", base);
```
brminich, line 875: we  should mention this limitation in the API

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -835,6 +855,43 @@ uct_ib_mlx5_devx_mem_reg_gva(uct_md_h uct_md, unsigned flags, uct_mem_h *memh_p)
...
+                  uct_ib_mlx5_dev_name(md), ucs_status_string(status));
+        return status;
+    }
+
+    /* Move indirect keys ownership from base to memory window */
+    memh->super.flags  |= UCT_IB_MEM_FLAG_MEM_WINDOW;
+    base->atomic_dvmr   = NULL;
+    base->atomic_rkey   = UCT_IB_INVALID_MKEY;
+    base->indirect_dvmr = NULL;
+    base->indirect_rkey = UCT_IB_INVALID_MKEY;
```
brminich, line 889: why do we need to invalidate basic memh keys? I'd expect that this will break (at least performance) of some (at least atomics) operations on the base key

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -362,6 +368,41 @@ UCS_TEST_SKIP_COND_P(test_ib_md, mt_fail,
...
+    std::vector<uint8_t> mw2_rkey1 = mkey_pack(mw2, flags);
+    EXPECT_EQ(base_rkey1, mw2_rkey1);
+
+    /* Test case 3: subsequent mkey_pack calls return the same result */
+    std::vector<uint8_t> mw2_rkey2 = mkey_pack(mw2, flags);
+    EXPECT_EQ(mw2_rkey1, mw2_rkey2);
+
+    /* Test case 4: base memh can still be used to pack mkeys */
+    std::vector<uint8_t> base_rkey2 = mkey_pack(base, flags);
+    EXPECT_NE(base_rkey1, base_rkey2);
```
brminich, line 395: i do not think it is correct. base memh should not be modified once derived memh is created

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: I think it should not hinder performance, because base memh rkeys will be recreated on next mkey_pack, like is done in my test.
But you are right, strictly speaking it's NOT necessary to transfer rkeys ownership to the derived memh. It will work even without this transfer. I will remove this ownership transfer logic

The reason why I added this transfer is the following:
Currently it's UCT layer that decides whether to create rkeys (atomic + indirect), and conditions for them are different:
- **atomic** rkey is created if user requested it (with `UCT_MD_MEM_ACCESS_REMOTE_ATOMIC`), and hardware supports it. So it might be created no matter if invalidation is needed
- **indirect** rkey is created only when invalidation is needed
Now from UCP layer we assume that we will create a memory window when invalidation if needed from EP config. By this time **atomic** rkeys on base memh might be already created and I thought it's a good idea to move it to the memory window, to avoid re-creating this key per MW.

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -362,6 +368,41 @@ UCS_TEST_SKIP_COND_P(test_ib_md, mt_fail,
...
+    std::vector<uint8_t> buffer(1024);
+    uct_mem_h base;
+    EXPECT_UCS_OK(reg_mem(UCT_MD_MEM_ACCESS_RMA, buffer.data(), buffer.size(),
+                          &base));
+
+    /* Test case 1: creating MW from memh before mkey_pack */
+    uct_mem_h mw1 = reg_mem_window(base);
+
+    /* Test case 2: creating MW from memh after mkey_pack */
+    std::vector<uint8_t> base_rkey1 = mkey_pack(base, flags);
```
tvegas1, line 384: do we need invalidation on first mkey pack of base?

tvegas1: do we need test with packing without invalidation first?

src/uct/api/uct.h
iyastreb: ok

src/uct/api/uct.h
iyastreb: Done

src/uct/api/v2/uct_v2.h
iyastreb: ok

src/uct/api/v2/uct_v2.h
iyastreb: Right, flag can be removed

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: I removed ownership transfer for now, let's consider this later

test/gtest/uct/ib/test_ib_md.cc
iyastreb: ok, fixed this

test/gtest/uct/ib/test_ib_md.cc
iyastreb: In terms of UCT invalidation would mean that we destroy derived memh. Currently I don't have tests with invalidation, actually it's a good point to add them.
Or I misunderstood your comment

src/uct/ib/base/ib_md.c
```diff
@@ -584,19 +584,35 @@ ucs_status_t uct_ib_mem_advise(uct_md_h uct_md, uct_mem_h memh, void *addr,
...
                           1 /* UCT_IB_MR_DEFAULT */;
     uct_ib_mem_t *memh;
 
-    memh = ucs_calloc(1, memh_base_size + (mr_size * num_mrs), "ib_memh");
+    *memh_size_p = memh_base_size + (mr_size * num_mrs);
+    memh         = ucs_calloc(1, *memh_size_p, "ib_memh");
     if (memh == NULL) {
         ucs_error("%s: failed to allocated memh struct",
                   uct_ib_device_name(&md->dev));
+        return NULL;
```
rakhmets, line 601:
```suggestion
```

test/gtest/uct/ib/test_ib_md.cc
```diff
@@ -42,6 +42,15 @@ class test_ib_md : public test_md
     void test_mkey_pack_mt_internal(unsigned access_mask, bool invalidate);
     void test_smkey_reg_atomic(void);
 
+    uct_mem_h reg_derived_mem(uct_mem_h base) const {
```
rakhmets, line 45: const uct_mem_h

iyastreb: ok

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -372,9 +372,17 @@ static ucs_status_t
 uct_cuda_ipc_mem_reg(uct_md_h md, void *address, size_t length,
                      const uct_md_mem_reg_params_t *params, uct_mem_h *memh_p)
 {
+    uct_mem_h base = (params != NULL) ?
```
rakhmets, line 375: `params` is not checked to be non-NULL in other places. I think we can remove this check here as well.

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -372,9 +372,17 @@ static ucs_status_t
...
+                        UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL) :
```
rakhmets, line 376: `UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL)` is needed only if `ENABLE_PARAMS_CHECK`.
So, you can move in under `if (ENABLE_PARAMS_CHECK)` block.
Moreover,  you can use the following macro:
```C
UCT_CHECK_PARAM(UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL) == NULL,
                "CUDA IPC does not support derived memory handles");
```

rakhmets (dismissed):

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: Yes, initially I didn't have this check, added it to fix NULL pointer crash in CI

rakhmets: Do you remember the failed test? Looks like an issue in caller function.

iyastreb: I don't remember it, but I just see my commit named "Fix NPE" that I've made to address that failure: https://github.com/openucx/ucx/pull/10332/commits/53a4b08c9d6ec65787ba669ca8685cbb130f9ec5

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -375,6 +375,10 @@ uct_cuda_ipc_mem_reg(uct_md_h md, void *address, size_t length,
     uct_cuda_ipc_memh_t *memh;
     CUdevice cu_device;
 
+    UCT_CHECK_PARAM((params == NULL) ||
```
yosefe, line 378: IMO no need to check (params == NULL)

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
```diff
@@ -375,6 +375,10 @@ uct_cuda_ipc_mem_reg(uct_md_h md, void *address, size_t length,
...
+                    UCT_MD_MEM_REG_FIELD_VALUE(params, memh, FIELD_MEMH, NULL) == NULL,
+                    "CUDA IPC does not support derived memory handles");
```
yosefe, line 380: should we make this a common macro and use in other memory domains that dont support deriver memh?

src/uct/ib/base/ib_md.c
```diff
@@ -584,19 +584,34 @@ ucs_status_t uct_ib_mem_advise(uct_md_h uct_md, uct_mem_h memh, void *addr,
     return UCS_OK;
 }
 
-ucs_status_t uct_ib_memh_alloc(uct_ib_md_t *md, size_t length,
-                               unsigned mem_flags, size_t memh_base_size,
-                               size_t mr_size, uct_ib_mem_t **memh_p)
+static uct_ib_mem_t *
+uct_ib_memh_alloc_internal(uct_ib_md_t *md, size_t memh_base_size,
```
yosefe, line 588: can we make better names, instead of uct_ib_memh_alloc_internal+uct_ib_memh_alloc - uct_ib_memh_alloc+uct_ib_memh_new/init

src/uct/ib/base/ib_md.c
```diff
@@ -626,6 +641,23 @@ ucs_status_t uct_ib_memh_alloc(uct_ib_md_t *md, size_t length,
...
+{
+    uct_ib_mem_t *memh;
+    size_t memh_size;
+
+    memh = uct_ib_memh_alloc_internal(md, memh_base_size, mr_size, &memh_size);
+    if (memh == NULL) {
+        return UCS_ERR_NO_MEMORY;
+    }
+
+    memcpy(memh, src, memh_size);
```
yosefe, line 656: 1. clone seems wrong - if the derived memh is shallow, why need to fully copy the original memh?
2. seems weird that we use calloc() and then override everything with memcpy

src/uct/api/v2/uct_v2.h
```diff
@@ -969,7 +982,12 @@ typedef enum {
      * packed key by @ref uct_md_mkey_pack_v2 with
      * @ref UCT_MD_MKEY_PACK_FLAG_INVALIDATE_AMO flag.
      */
-    UCT_MD_FLAG_INVALIDATE_AMO = UCS_BIT(12)
+    UCT_MD_FLAG_INVALIDATE_AMO = UCS_BIT(12),
+
+    /**
+     * Memory domain supports derived memory handle registration.
+     */
+    UCT_MD_FLAG_DERIVED        = UCS_BIT(13)
```
yosefe, line 990: UCT_MD_FLAG_REG_DERIVED

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -778,6 +778,26 @@ uct_ib_mlx5_devx_memh_alloc(uct_ib_mlx5_md_t *md, size_t length,
     return UCS_OK;
 }
 
+static ucs_status_t
+uct_ib_mlx5_devx_memh_clone(uct_ib_mlx5_md_t *md,
+                            const uct_ib_mlx5_devx_mem_t *src,
+                            uct_ib_mlx5_devx_mem_t **memh_p)
+{
+    size_t mr_size = src->super.flags & UCT_IB_MEM_IMPORTED ?
```
yosefe, line 786: ( )

src/uct/ib/base/ib_md.c
iyastreb: Answering if the opposite order:
2. Right, here malloc would be enough but I'm reusing existing function `uct_ib_memh_alloc_internal` that does `calloc` and calculates the size, just to reuse the common code. I think the overhead is minimal.
We could split size calculation into a separate function etc, but I think it's not worth the effort

1. Ok, maybe it's a terminology issue here
My understanding is:
**Deep copy** creates an independent instance of an object, that can be used apart from the original.
**Shallow copy** creates an "alias" that depends on master copy and cannot be used separately.
The latter is what's implemented in this PR, whatever we name it.

Derived memh is a shallow copy, because it makes a shallow copy of MRs - the most important part of original memh. And original object remains the only owner of the MRs state.
Of course there could be different implementations of derived memh. For example, we can imagine a shallow copy looking like that:
```
struct {
   uct_ib_mlx5_devx_mem_t *base;
   // derived specific fields
} uct_derived_memh;
```
Looks ok, right? Still you need to allocate this object.
What are the issues with this approach:
- There are places where we assume that `memh` has always the `uct_ib_mem_t` base: uct_rc_mlx5_txqp_tag_inline_post: ((uct_ib_mem_t*)iov->memh)
So we must add `uct_ib_mem_t super` field to our copy
- The copy must handle a set of rkeys independently from the original memh, meaning that the following fields also needs to be copied:
```
    struct mlx5dv_devx_obj      *atomic_dvmr;
    struct mlx5dv_devx_obj      *indirect_dvmr;
    uint32_t                    atomic_rkey;
    uint32_t                    indirect_rkey;
```
This is just to show you that we need to duplicate the significant part of the original memh anyway.
- Then in each and every place we need to check whether passed memh is derived or original (because their layout is different), adding a lot of boilerplate code and CPU overhead..
- Then we need to modify quite some existing functions for key generation, because they cache keys in the original object.

To overcome all these issues I make a shallow copy of the original memh (== they have the same memory layout), and use it interchangeably in all the places. This allows me to minimize the amount of ifs in the code, basically we check for derived handle only during init and cleanup. This helps to avoid any refactoring in the existing functions.

yosefe: I think we nee to brainstorm the approaches. I don't like having structs where only some fields used, maybe need a deeper refactoring in the IB memh structure.

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
iyastreb: ok

src/uct/ib/base/ib_md.c
iyastreb: `uct_ib_memh_alloc` is an existing function, so we don't want to change its name
I just extracted some common part from `uct_ib_memh_alloc` into _internal so that it can be reused by `uct_ib_memh_clone`.

Maybe we name it `uct_ib_memh_alloc_common`?

src/uct/api/v2/uct_v2.h
```diff
@@ -480,6 +481,18 @@ typedef struct uct_md_mem_reg_params {
      * dmabuf region, then this field must be omitted or set to 0.
      */
     size_t                       dmabuf_offset;
+
+    /**
+     * A pointer to an existing memory handle.
+     * Used to register a derived memh: a shallow copy of an existing UCT memh
+     * which can be used to access the same memory region. When created, the
+     * derived memh inherits the access flags and the state of the original
```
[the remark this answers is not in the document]
iyastreb: I was not aware of this use case.
Do you propose to update this documentation? Or extend derived memh API so that we can create it with alternate access rights?

src/uct/api/v2/uct_v2.h
iyastreb: done

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: Same comment I got from Raul)
I added this check intentionally, because some unit test was triggering this call with NULL param.
Ok, let me remove it, so we see exactly that test

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: Yes, we can
I think we can use this new macro only in those MDs which advertise INVALIDATION support, but do not properly handle it. These are namely: cuda_ipc_md and cma_md

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: Ok, found and fixed this NULL pointer error in gtest: https://github.com/openucx/ucx/pull/10332/commits/b193dea998256ee80614f9820e55c70b35158a23

src/uct/cuda/cuda_ipc/cuda_ipc_md.c
iyastreb: Ok, found and fixed this NULL pointer error in gtest: https://github.com/openucx/ucx/pull/10332/commits/b193dea998256ee80614f9820e55c70b35158a23

test/gtest/uct/cuda/test_cuda_ipc_md.cc
```diff
@@ -21,7 +21,8 @@ class test_cuda_ipc_md : public test_md {
     {
         uct_mem_h memh;
         uct_cuda_ipc_rkey_t rkey;
-        EXPECT_UCS_OK(md->ops->mem_reg(md, (void *)ptr, size, NULL, &memh));
+        uct_md_mem_reg_params_t reg_params = {};
+        ASSERT_UCS_OK(md->ops->mem_reg(md, (void *)ptr, size, &reg_params, &memh));
```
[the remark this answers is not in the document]
iyastreb: makes sense, changed in the whole file

src/uct/api/v2/uct_v2.h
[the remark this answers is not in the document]
iyastreb: we could probably reuse: https://github.com/openucx/ucx/pull/10341

gleon99:
Hi @iyastreb - we're cleaning up the UCX PR backlog and plan to close this PR in 7 days. Please let us know if you'd like to keep it open.

PR #10336, opened 2024-11-27
GTEST/UCP: Disabled checking the values in test_ucp_perf.envelope.

rakhmets:
## What?
Disabled checking the values in test_ucp_perf.envelope. It doesn't make much sense to check the performance during the CI. Since multiple instances may run on the same host, and there may be slowdowns.

Reverted https://github.com/openucx/ucx/commit/4ec14557474d025e80fd7c9979ba5ae4b191b11a.

### DIFF
--- a/test/gtest/ucp/test_ucp_perf.cc
+++ b/test/gtest/ucp/test_ucp_perf.cc
@@ -16,8 +16,6 @@ extern "C" {
 }
 
 #define MB   pow(1024.0, -2)
-#define UCT_PERF_TEST_MULTIPLIER  5
-#define UCT_ARM_PERF_TEST_MULTIPLIER  15
 
 class test_ucp_perf : public ucp_test, public test_perf {
 public:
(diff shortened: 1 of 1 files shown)


test/gtest/ucp/test_ucp_perf.cc
```diff
@@ -351,19 +347,11 @@ UCS_TEST_SKIP_COND_P(test_ucp_perf, envelope, has_transport("self"))
...
-        test.min /= UCT_PERF_TEST_MULTIPLIER;
-    }
-    test.iters = ucs_min(test.iters, max_iter);
-
+    test.iters         = ucs_min(test.iters, max_iter);
     test.send_mem_type = UCS_MEMORY_TYPE_HOST;
     test.recv_mem_type = UCS_MEMORY_TYPE_HOST;
 
-    run_test(test, 0, check_perf, "", "");
+    run_test(test, 0, false, "", "");
```
brminich, line 354: can we remove `check_perf` from `run_test` now?

rakhmets: It is possible, if we disable checking performance in other tests.

yosefe (approved):

brminich (approved):

Issue #10085, opened 2024-08-23
Address Registration Error in CUDA Aware MPICH 4.2.2 + UCX 1.17.0 Application


cl3to:
I'm running an application on a cluster that uses CUDA Aware MPICH (v4.2.2) and UCX (v1.17.0). My application consists of two binaries, a server and a client, so I use the MPMD mode of `mpirun` to execute it: `mpirun -np 1 server : -np 1 client`. The problem is that when I try to run the application, either intra-node or inter-node, I get the following error and the application hangs:

```
[1724426336.207066] [c066:48733:0]           ib_md.c:293  UCX  ERROR ibv_reg_mr(address=0x55bace2c02a0, length=49792, access=0xf) failed: Bad address
[1724426336.207083] [c066:48733:0]          ucp_mm.c:70   UCX  ERROR failed to register address 0x55bace2c02a0 (host) length 49792 on md[8]=mlx5_1: Input/output error (md supports: host|cuda)
```

After some research, I found that setting the environment variable `UCX_RCACHE_ENABLE=n` allows my application to run without errors. However, the application’s runtime performance is not as expected. Profiling the application revealed that most of the time is spent on data transfer between the nodes.

When running the OSU 7.4 benchmark, I observed that the bandwidth between nodes using InfiniBand is approximately 5.75 times slower when I set the variable `UCX_RCACHE_ENABLE=n`.

```
export UCX_RCACHE_ENABLE=y
mpirun -ppn 1 -np 2 osu_bw -m 100000000:1000000000 or
mpirun -ppn 1 -np 1 osu_bw -m 100000000:1000000000 : -np 1 osu_bw -m 100000000:1000000000

# OSU MPI Bandwidth Test v7.4
# Datatype: MPI_CHAR.
# Size      Bandwidth (MB/s)
100000000           23073.19
200000000           23037.84
400000000           23001.02
800000000           22948.29
```

```
export UCX_RCACHE_ENABLE=n
mpirun -ppn 1 -np 2 osu_bw -m 100000000:1000000000 or
mpirun -ppn 1 -np 1 osu_bw -m 100000000:1000000000 : -np 1 osu_bw -m 100000000:1000000000

# OSU MPI Bandwidth Test v7.4
# Datatype: MPI_CHAR.
# Size      Bandwidth (MB/s)
100000000            4083.05
200000000            4078.84
400000000            4076.13
800000000            4076.14
```

Any suggestions on why the application might be failing to register addresses?

### Setup and versions

[...]

PR #10341, opened 2024-12-01
UCT/IB/MD: retry memory registration with reduced access flags on failure

amastbaum:
## What?
Enable memory registration for read-only addresses.

## Why?
A customer reported an issue when attempting to send read-only memory using RDMA. Details can be found in the related discussion: [https://github.com/openucx/ucx/issues/10085#issuecomment-2377368540](url).


src/ucp/rndv/proto_rndv.inl
```diff
@@ -49,7 +49,8 @@ ucp_proto_rndv_rts_request_init(ucp_request_t *req)
     status = ucp_datatype_iter_mem_reg(ep->worker->context,
                                        &req->send.state.dt_iter,
                                        rpriv->md_map,
-                                       UCT_MD_MEM_ACCESS_RMA |
+                                       UCT_MD_MEM_ACCESS_REMOTE_GET |
+                                       UCT_MD_MEM_ACCESS_LOCAL_READ |
```
gleon99, line 53: Needed?

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -703,11 +703,13 @@ uct_ib_mlx5_devx_reg_mr(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh,
...
-    uint64_t access_flags = uct_ib_memh_access_flags(&memh->super,
-                                                     md->super.relaxed_order) &
-                            access_mask;
     unsigned flags        = UCT_MD_MEM_REG_FIELD_VALUE(params, flags,
                                                        FIELD_FLAGS, 0);
+    uint64_t access_flags = uct_ib_memh_access_flags(&memh->super,
+                                                     md->super.relaxed_order,
+                                                     1, flags) &
+                            access_mask;
+    int attempt_cnt       = 0;
```
gleon99, line 712: Avoid abbreviations (`cnt`)

src/uct/ib/base/ib_md.c
```diff
@@ -626,9 +627,39 @@ ucs_status_t uct_ib_memh_alloc(uct_ib_md_t *md, size_t length,
     return UCS_OK;
 }
 
-uint64_t uct_ib_memh_access_flags(uct_ib_mem_t *memh, int relaxed_order)
+uint64_t uct_flags_to_ibv_mem_access_flags(uint64_t uct_flags)
```
gleon99, line 630: `uct_ib_..`

src/uct/ib/base/ib_md.c
```diff
@@ -460,7 +460,7 @@ uct_ib_md_handle_mr_list_mt(uct_ib_md_t *md, void *address, size_t length,
 ucs_status_t uct_ib_reg_mr(uct_ib_md_t *md, void *address, size_t length,
                            const uct_md_mem_reg_params_t *params,
                            uint64_t access_flags, struct ibv_dm *dm,
-                           struct ibv_mr **mr_p)
+                           struct ibv_mr **mr_p, int first_attempt)
```
gleon99, line 463: `first_attempt` -> `ignore_errors` or smth similar.
This func doesn't care about "attempts" and their number, the meaning of this flag here is only to "force" ignore errors.

src/uct/ib/base/ib_md.c
```diff
@@ -626,9 +627,39 @@ ucs_status_t uct_ib_memh_alloc(uct_ib_md_t *md, size_t length,
...
+
+    if (uct_flags & UCT_MD_MEM_ACCESS_REMOTE_ATOMIC) {
+        ibv_flags |= IBV_ACCESS_REMOTE_ATOMIC;
+    }
+
+    return ibv_flags;
+}
+
+uint64_t uct_ib_memh_access_flags(uct_ib_mem_t *memh, int relaxed_order,
+                                  int first_attempt, uint64_t uct_flags)
```
gleon99, line 654: Here as well, I'd rename `first_attempt` to its real meaning in this context.
Maybe smth like `use_uct_flags`?

src/uct/ib/base/ib_md.c
```diff
@@ -645,26 +676,33 @@ ucs_status_t uct_ib_verbs_mem_reg(uct_md_h uct_md, void *address, size_t length,
...
 
-    memh         = ucs_derived_of(ib_memh, uct_ib_verbs_mem_t);
-    access_flags = uct_ib_memh_access_flags(&memh->super, md->relaxed_order);
+    memh = ucs_derived_of(ib_memh, uct_ib_verbs_mem_t);
+
+    do {
+        access_flags = uct_ib_memh_access_flags(&memh->super, md->relaxed_order,
+                                                attempt_cnt == 0, uct_flags);
+
+        status = uct_ib_reg_mr(md, address, length, params, access_flags, NULL,
```
gleon99, line 701: Align `=`

src/uct/ib/mlx5/dv/ib_mlx5dv_md.c
```diff
@@ -731,8 +733,16 @@ uct_ib_mlx5_devx_reg_mr(uct_ib_mlx5_md_t *md, uct_ib_mlx5_devx_mem_t *memh,
...
     }
 
-    status = uct_ib_reg_mr(&md->super, address, length, params, access_flags,
-                           NULL, &memh->mrs[mr_type].super.ib);
+    do {
+        status = uct_ib_reg_mr(&md->super, address, length, params, access_flags,
+                               NULL, &memh->mrs[mr_type].super.ib,
+                               attempt_cnt == 0);
+        attempt_cnt++;
+        access_flags = uct_ib_memh_access_flags(&memh->super,
```
gleon99, line 741: First `reg_mr`, then `access_flags` - is that intentional?
This line always has `attempt_cnt` >= 1...

amastbaum: Yes, it's intentional, though it might be written better..
`access_flags` is first calculated at the top of the function because it's used for another function, and only then it is used in the `do while` loop

src/ucp/rndv/proto_rndv.inl
amastbaum: It's just the most basic option so I thought I can keep mention it as well, but on the other hand it doesn't matter as local read is enabled by default in ibv_reg_mr (there is no such flag in verbs.h because it's always enabled)

src/uct/ib/base/ib_md.c
```diff
@@ -626,9 +627,35 @@ ucs_status_t uct_ib_memh_alloc(uct_ib_md_t *md, size_t length,
     return UCS_OK;
 }
 
-uint64_t uct_ib_memh_access_flags(uct_ib_mem_t *memh, int relaxed_order)
+uint64_t uct_ib_flags_to_ibv_mem_access_flags(uint64_t uct_flags)
```
yosefe, line 630: static

src/uct/ib/base/ib_md.c
```diff
@@ -309,7 +309,7 @@ void *uct_ib_md_mem_handle_thread_func(void *arg)
     while (ctx->length > 0) {
         if (ctx->params != NULL) {
             status = uct_ib_reg_mr(ctx->md, ctx->address, length, ctx->params,
-                                   ctx->access_flags, NULL, &ctx->mrs[mr_idx]);
+                                   ctx->access_flags, NULL, &ctx->mrs[mr_idx], 1);
```
yosefe, line 312: why hide_errors is true here?

src/uct/ib/base/ib_md.c
```diff
@@ -645,26 +672,33 @@ ucs_status_t uct_ib_verbs_mem_reg(uct_md_h uct_md, void *address, size_t length,
...
-    access_flags = uct_ib_memh_access_flags(&memh->super, md->relaxed_order);
+    memh = ucs_derived_of(ib_memh, uct_ib_verbs_mem_t);
+
+    do {
+        access_flags = uct_ib_memh_access_flags(&memh->super, md->relaxed_order,
+                                                attempt_count > 0, uct_flags);
+
+        status       = uct_ib_reg_mr(md, address, length, params, access_flags,
+                                     NULL, &mr_default, attempt_count == 0);
+        attempt_count++;
```
yosefe, line 699: This retry should happen in UCP layer, otherwise IMO it can cause issues with UCP rcache region merges

src/uct/ib/base/ib_md.c
amastbaum: @yosefe Oh, I actually wanted to put here a value that doesn't change the previous behavior and put `1` by mistake. I'll also change it in the other places.

src/uct/ib/base/ib_md.c
amastbaum: @yosefe Where are the rcache region merges triggered from?
Do you think we should still try to register with full permissions and only then with specific ones?

src/ucp/core/ucp_mm.c
```diff
@@ -462,7 +462,7 @@ static void ucp_memh_cleanup(ucp_context_h context, ucp_mem_h memh)
 }
 
 static ucs_status_t ucp_memh_register_gva(ucp_context_h context, ucp_mem_h memh,
-                                          ucp_md_map_t md_map)
+                                          ucp_md_map_t md_map, unsigned uct_flags)
```
Artemy-Mellanox, line 465: I understand it's temp, but, just in case, it would be good to replace gva_mr with some structure and keep registrations for different access flags there.

yosefe:
@Artemy-Mellanox @iyastreb can you pls review?

src/tools/perf/lib/libperf_memory.c
```diff
@@ -53,11 +53,16 @@ static ucs_status_t ucp_perf_mem_alloc(const ucx_perf_context_t *perf,
...
                          UCP_MEM_MAP_PARAM_FIELD_LENGTH |
                          UCP_MEM_MAP_PARAM_FIELD_FLAGS |
-                         UCP_MEM_MAP_PARAM_FIELD_MEMORY_TYPE;
+                         UCP_MEM_MAP_PARAM_FIELD_MEMORY_TYPE |
+                         UCP_MEM_MAP_PARAM_FIELD_PROT;
     params.address     = NULL;
     params.memory_type = mem_type;
     params.length      = length;
     params.flags       = UCP_MEM_MAP_ALLOCATE;
+    params.prot        = UCP_MEM_MAP_PROT_LOCAL_READ |
```
iyastreb, line 62: Can it be the default behavior: if we don't pass any protection, then we assume everything is allowed?
In this case we don't need to pass the same params everywhere

test/gtest/ucp/test_ucp_mem_type.cc
```diff
@@ -143,9 +143,14 @@ UCS_TEST_P(test_ucp_cuda, sparse_regions) {
 
         ucp_mem_map_params_t params;
         params.field_mask = UCP_MEM_MAP_PARAM_FIELD_ADDRESS |
-                            UCP_MEM_MAP_PARAM_FIELD_LENGTH;
+                            UCP_MEM_MAP_PARAM_FIELD_LENGTH |
+                            UCP_MEM_MAP_PARAM_FIELD_PROT;
         params.address    = ptr[i];
         params.length     = size;
+        params.prot       = UCP_MEM_MAP_PROT_LOCAL_READ |
```
iyastreb, line 150: maybe create an alias for this combination of flags?
something like:
```
#define UCP_MEM_MAP_PROT_ALL \
UCP_MEM_MAP_PROT_LOCAL_READ |
                            UCP_MEM_MAP_PROT_LOCAL_WRITE |
                            UCP_MEM_MAP_PROT_REMOTE_READ |
                            UCP_MEM_MAP_PROT_REMOTE_WRITE;

File: src/ucp/core/ucp_context.h [hunk 14]
```diff
@@ -276,6 +276,16 @@ typedef struct ucp_tl_cmpt {
 } ucp_tl_cmpt_t;
 
 
+/**
+ * Global VA MR type
+ */
+typedef enum ucp_gva_mr_type {
```
iyastreb, line 282: cosmetic:
```suggestion
typedef enum {
```

src/ucp/core/ucp_mm.c
```diff
@@ -326,6 +344,15 @@ ucp_mem_map_params2uct_flags(const ucp_context_h context,
     return flags;
 }
 
+static ucp_gva_mr_type_t get_gva_type(unsigned uct_flags)
```
iyastreb, line 347: minor: maybe inline or UCS_F_ALWAYS_INLINE

iyastreb (dismissed):

src/ucp/core/ucp_context.h
amastbaum: I see that the structs above are defined the same way so I wanted to remain consistent

src/ucp/core/ucp_mm.c
amastbaum: I made it a macro instead

iyastreb (dismissed):

gleon99:
@yosefe  do you want to review this one?

filipesmg:
@gleon99 Thank you for the feedback! It's important also for the discussions on our side. Good luck on the further changes, I hope it doesn't take too long. 🍀

PR #10342, opened 2024-12-02
UCT/GDR_COPY: Optimize memory registration in fast-path

Artemy-Mellanox:


src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,108 @@
...
+
+static inline void ucs_rwlock_read_lock(ucs_rwlock_t *lock) {
+    int x;
+
+    while (1) {
+        while (lock->l & UCS_RWLOCK_MASK) {
+            sched_yield();
+        }
+
+        x = __sync_fetch_and_add(&lock->l, UCS_RWLOCK_READ);
```
iyastreb, line 53: For my curiosity: why we are using deprecated atomic functions, while there are modern family of `__atomic_*` functions `__atomic_fetch_add`?
Those `__sync` functions always make a full barrier, and therefore come with significant overhead when only acquire or release semantics is sufficient

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,108 @@
...
+
+static inline void ucs_rwlock_write_unlock(ucs_rwlock_t *lock) {
+    __sync_fetch_and_sub(&lock->l, UCS_RWLOCK_WRITE);
+}
+
+
+static inline void ucs_rwlock_init(ucs_rwlock_t *lock) {
+    lock->l = 0;
+}
+
```
iyastreb, line 107: I believe it must come with a set of unit tests.
And maybe multithread (like 32 threads) litmus test that would run for a while with some simple logic (like incrementing few counters by reader, acquring these counters by readers etc)

src/ucs/type/rwlock.h
iyastreb: From GCC documentation: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fsync-Builtins.html
`They [family of __sync function] should not be used for new code which should use the ‘__atomic’ builtins instead.`

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,108 @@
...
+static inline void ucs_rwlock_read_unlock(ucs_rwlock_t *lock) {
+    __sync_fetch_and_sub(&lock->l, UCS_RWLOCK_READ);
+}
+
+
+static inline void ucs_rwlock_write_lock(ucs_rwlock_t *lock) {
+    int x;
+
+    while (1) {
+        x = lock->l;
```
iyastreb, line 72: maybe reuse `ucs_rwlock_write_trylock` here?

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,108 @@
...
+static inline int ucs_rwlock_write_trylock(ucs_rwlock_t *lock) {
+    int x;
+
+    x = lock->l;
+    if ((x < UCS_RWLOCK_WRITE) &&
+        (__sync_val_compare_and_swap(&lock->l, x, UCS_RWLOCK_WRITE) == x)) {
+        return 0;
+    }
+
+    return -EBUSY;
```
iyastreb, line 95: maybe it's more logical to return 1 when lock is acquired, 0 otherwise?

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,108 @@
...
+ * WAIT: writer pending --/
+ */
+
+#define UCS_RWLOCK_WAIT  0x1 /* Writer is waiting */
+#define UCS_RWLOCK_WRITE 0x2 /* Writer has the lock */
+#define UCS_RWLOCK_MASK  (UCS_RWLOCK_WAIT | UCS_RWLOCK_WRITE)
+                             /* Writer is waiting or has lock */
+#define UCS_RWLOCK_READ  0x4 /* Reader increment */
+
+
```
iyastreb, line 36: Maybe also add `#define UCS_RWLOCK_STATIC_INITIALIZER {0}`

src/ucs/type/rwlock.h
brminich: maybe consider using/extending `atomic.h`?

src/ucp/core/ucp_mm.c
```diff
@@ -1187,6 +1187,44 @@ ucs_status_t ucp_mem_unmap(ucp_context_h context, ucp_mem_h memh)
     return UCS_OK;
 }
 
+static ucs_status_t
```
yosefe, line 1190: If we always use one md for mtype, maybe we can avoid ucp_memh as well, just use uct objects?

src/ucs/Makefile.am
```diff
@@ -75,6 +75,7 @@ nobase_dist_libucs_la_HEADERS = \
 	type/param.h \
 	type/init_once.h \
 	type/spinlock.h \
+	type/rwlock.h \
```
yosefe, line 78: can we move "introducing fast rwlock to ucs" to separate PR, including gtest for functionality and performance measurement (print lock overhead using UCS_TEST_MESSAGE + write in the PR desc how fast is new lock compared to glibc)

src/ucs/memory/rcache.inl
```diff
@@ -80,6 +81,17 @@ ucs_rcache_lookup_unsafe(ucs_rcache_t *rcache, void *address, size_t length,
     return region;
 }
 
+static UCS_F_ALWAYS_INLINE ucs_rcache_region_t *
+ucs_rcache_lookup(ucs_rcache_t *rcache, void *address, size_t length,
```
yosefe, line 85: can we integrate this optimizations into ucs_rcache_get?

Artemy-Mellanox:
Split into https://github.com/openucx/ucx/pull/10734, https://github.com/openucx/ucx/pull/10355, https://github.com/openucx/ucx/pull/10362

PR #10127, opened 2024-09-05
STACK: Change PATH_MAX stack allocation to heap

michal-shalev:
## What?
This PR changes all stack allocations of char arrays sized `PATH_MAX` to heap allocations.

## Why?
Customers with limited stack sizes have recently experienced stack overflow issues when running UCX. Moving these allocations to the heap prevents such overflows.

## How?
Heap allocations replace stack allocations for char arrays of size `PATH_MAX`.

PR #10308, opened 2024-11-19
CONTRIB: Automate AUTHORS file update

tvegas1:
## What?
Automate AUTHORS updates according to documentation.

## Why?
Seems better to try to have common approach.

## How?
```shell
$ ./contrib/authors_update.sh upstream/v1.17.x..upstream/v1.18.x
..
   Artemy Kovalyov <artemyko@nvidia.com>
++ Arun Chandran <Arun.Chandran@amd.com>
   Edgar Gabriel <Edgar.Gabriel@amd.com>
   Evgeny Leksikov <evgenylek@mellanox.com>
..
```

PR #10345, opened 2024-12-02
VFS: Ignore EEXIST errno

michal-shalev:
## What?
Ignore EEXIST errno in `vfs_mount()` and `ucs_vfs_sock_mkdir()`.

## Why?
Bug introduced in https://github.com/openucx/ucx/pull/10127 where instead of ignoring the EEXIST that returns from mkdir, `vfs_mount()` and `ucs_vfs_sock_mkdir()` return an error in case the directory already exists, which should be ignored.

### DIFF
--- a/src/tools/vfs/vfs_main.c
+++ b/src/tools/vfs/vfs_main.c
@@ -191,10 +191,15 @@ int vfs_mount(int pid)
     }
 
     ret = mkdir(mountpoint, S_IRWXU);
-    if ((ret < 0) && (errno != EEXIST)) {
-        ret = -errno;
-        vfs_error("failed to create directory '%s': %m", mountpoint);
-        goto out_free_mountpoint;
+    if (ret < 0) {
+        if (errno == EEXIST) {
+            /* Directory already exists */
+            ret = 0;
+        } else {
+            ret = -errno;
+            vfs_error("failed to create directory '%s': %m", mountpoint);
+            goto out_free_mountpoint;
+        }
     }
 
     /* Mount a new FUSE filesystem in the mount point directory */
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,6 +57,7 @@ Manjunath Gorentla Venkata <manjugv@gmail.com>
 Marek Schimara <Marek.Schimara@bull.net>
 Mark Allen <markalle@us.ibm.com>
 Matthew Baker <bakermb@ornl.gov>
+Michal Shalev <mshalev@nvidia.com>
 Mike Dubman <miked@mellanox.com>
 Mikhail Brinskii <mikhailb@nvidia.com>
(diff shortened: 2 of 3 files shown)


src/tools/vfs/vfs_main.c
```diff
@@ -191,10 +191,15 @@ int vfs_mount(int pid)
     }
 
     ret = mkdir(mountpoint, S_IRWXU);
-    if ((ret < 0) && (errno != EEXIST)) {
-        ret = -errno;
-        vfs_error("failed to create directory '%s': %m", mountpoint);
-        goto out_free_mountpoint;
+    if (ret < 0) {
```
iyastreb, line 194: I don't mind the change, but in this place there is no issue even without this fix, because `ret` gets assigned a new value on line 207. But up to you

iyastreb (approved):

src/tools/vfs/vfs_main.c
michal-shalev: If we ignore `EEXIST` than IMO ret should be 0 because it reflects the status of this code section

buildlib/pr/codestyle.yml
```diff
@@ -84,7 +84,13 @@ jobs:
...
           echo "Looking for missing AUTHORS on commit range ${range}"
           ./contrib/authors_update.sh "$range"
-          git diff --exit-code
+
+          if ! git diff --quiet AUTHORS; then
+            echo "The AUTHORS file has been updated. Committing changes..."
+            git add AUTHORS
+            git commit -m "AUTHORS: Update AUTHORS file"
+            git push
+          fi
```
yosefe, line 93: seems unrelated fix?

contrib/authors_update.sh
```diff
@@ -26,7 +26,8 @@ tmp=$(mktemp --tmpdir=./)
 grep @ AUTHORS >"$tmp"
 
 echo "Names:"
-
+new_name=""
```
gleon99, line 29: Unrelated?

buildlib/pr/codestyle.yml
michal-shalev: image
CI was failing because I wasn't in the AUTHORS file, added myself
See https://github.com/openucx/ucx/pull/10308#issuecomment-2511630243

contrib/authors_update.sh
michal-shalev: image
CI was failing because I wasn't in the AUTHORS file, added myself
See https://github.com/openucx/ucx/pull/10308#issuecomment-2511630243

gleon99:
@michal-shalev please squash

iyastreb (approved):

yosefe (approved):

PR #10355, opened 2024-12-05
UCS: Introduce lightweight rwlock

Artemy-Mellanox:
[ RUN      ] test_rwlock.perf <> <>
[     INFO ] builtin read: 145.328/0 Mops/s
[     INFO ] builtin read cont2: 4.30855/0 Mops/s
[     INFO ] builtin read cont: 0.55125/0 Mops/s
[     INFO ] builtin write: 0/160.971 Mops/s
[     INFO ] builtin write cont2: 0/3.98984 Mops/s
[     INFO ] builtin write cont: 0/0.89214 Mops/s
[     INFO ] builtin mix1: 2.18771/2.13176 Mops/s
[     INFO ] builtin mix2: 3.6686/2.72748 Mops/s
[     INFO ] builtin mix: 0.42619/0.04763 Mops/s
[       OK ] test_rwlock.perf (924 ms)
[ RUN      ] test_rwlock.pthread <> <>
[     INFO ] pthread read: 89.2717/0 Mops/s
[     INFO ] pthread read cont2: 3.52858/0 Mops/s
[     INFO ] pthread read cont: 3.80326/0 Mops/s
[     INFO ] pthread write: 0/83.4628 Mops/s
[     INFO ] pthread write cont2: 0/4.27843 Mops/s
[     INFO ] pthread write cont: 0/7.51501 Mops/s
[     INFO ] pthread mix1: 0.12964/0.11595 Mops/s
[     INFO ] pthread mix2: 6.27234/0.10625 Mops/s
[     INFO ] pthread mix: 4.1059/1e-05 Mops/s
[       OK ] test_rwlock.pthread (921 ms)


test/gtest/ucs/test_type.cc
```diff
@@ -138,6 +142,145 @@ UCS_TEST_F(test_type, pack_float) {
...
+                         << std::to_string(writers) << " writers per 256 ";
+    }
+
+    void measure(const std::function<void()> &r,
+                 const std::function<void()> &w, const std::string &name)
+    {
+        int m = std::thread::hardware_concurrency();
+        std::vector<int> threads = {1, 2, 4, m};
+        std::vector<int> writers_per_256 = {1, 25, 128, 250};
+
```
yosefe, line 190: 1. i think we also want to measure overhead of read lock+unlock regardless of concurrency since it is the reason we added lightweight rwlock
2. can you pls post example output in the PR description?

test/gtest/ucs/test_type.cc
```diff
@@ -138,6 +142,145 @@ UCS_TEST_F(test_type, pack_float) {
...
+        ucs_rwlock_write_unlock(&lock);
+    });
+    sleep();
+    EXPECT_FALSE(write_taken); /* write lock should wait for read lock release */
+
+    ucs_rwlock_read_unlock(&lock);
+    sleep();
+    EXPECT_FALSE(write_taken); /* first read lock still holding lock */
+
+    int read_taken = 0;
```
yosefe, line 220: bool

test/gtest/ucs/test_type.cc
```diff
@@ -138,6 +142,145 @@ UCS_TEST_F(test_type, pack_float) {
...
+
+UCS_TEST_F(test_rwlock, lock) {
+    ucs_rwlock_t lock = UCS_RWLOCK_STATIC_INITIALIZER;
+
+    ucs_rwlock_read_lock(&lock);
+    EXPECT_EQ(-EBUSY, ucs_rwlock_write_trylock(&lock));
+
+    ucs_rwlock_read_lock(&lock); /* second read lock should pass */
+
+    int write_taken = 0;
```
yosefe, line 207: bool

test/gtest/ucs/test_type.cc
```diff
@@ -138,6 +142,145 @@ UCS_TEST_F(test_type, pack_float) {
...
 }
 
+class test_rwlock : public ucs::test {
+protected:
+    void sleep()
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    }
+
+    void measure_one(int num, int writers, const std::function<void()> &r,
```
iyastreb, line 152: maybe introduce a typedef with `using`:
`using run_func_t = std::function<void()>;`

iyastreb: I also propose more explicit names:
`r` -> `reader`
`w` -> `writer`
`num` -> `thread_count`

test/gtest/ucs/test_type.cc
```diff
@@ -138,6 +142,145 @@ UCS_TEST_F(test_type, pack_float) {
...
+                     const std::function<void()> &w, const std::string &name)
+    {
+        std::vector<std::thread> tt;
```
iyastreb, line 155: minor: maybe name it threads

test/gtest/ucs/test_type.cc
```diff
@@ -138,6 +142,145 @@ UCS_TEST_F(test_type, pack_float) {
...
+                         << std::to_string(num) << " threads "
+                         << std::to_string(writers) << " writers per 256 ";
+    }
+
+    void measure(const std::function<void()> &r,
+                 const std::function<void()> &w, const std::string &name)
+    {
+        int m = std::thread::hardware_concurrency();
+        std::vector<int> threads = {1, 2, 4, m};
+        std::vector<int> writers_per_256 = {1, 25, 128, 250};
```
iyastreb, line 189: maybe use percent to set the number of writers? It will be easier to understand
```C
std::vector<int> writers_percent = {1, 25, 50, 75, 90};
```

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+static inline void ucs_rwlock_read_lock(ucs_rwlock_t *lock)
+{
+    int x;
+
+    while (1) {
+        while (lock->l & UCS_RWLOCK_MASK) {
+            ucs_cpu_relax();
+        }
+
+        x = __atomic_fetch_add(&lock->l, UCS_RWLOCK_READ, __ATOMIC_ACQUIRE);
```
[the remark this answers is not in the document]
iyastreb: maybe the we replace deprecated `__sync*` fuctions from `atomic.h` with new `__atomic` variants?

test/gtest/ucs/test_type.cc
```diff
@@ -138,6 +142,145 @@ UCS_TEST_F(test_type, pack_float) {
...
+
+    ucs_rwlock_write_unlock(&lock);
+    sleep();
+    EXPECT_TRUE(read_taken); /* read lock should be taken */
+    r2.join();
+}
+
+UCS_TEST_F(test_rwlock, perf) {
+    ucs_rwlock_t lock = UCS_RWLOCK_STATIC_INITIALIZER;
+    measure(
```
iyastreb, line 256: This is a good performance test, but it must change some state to guarantee the lock correctness. This is the whole purpose of litmus test. For instance, these functions may perform some simple math calculations and we check the invariant:
```C
    // Invariant: counter2 is 2 times bigger than counter1
    int counter1 = 1;
    int counter2 = 2;
    measure(
            [&]() {
                ucs_rwlock_read_lock(&lock);
                UCS_ASSERT_EQ(counter1 * 2, counter2);
                ucs_rwlock_read_unlock(&lock);
            },
            [&]() {
                ucs_rwlock_write_lock(&lock);
                counter1 += counter1;
                counter2 += counter2;
                if (counter2 > 100000) {
                    counter1 = 1;
                    counter2 = 2;
                }
                ucs_rwlock_write_unlock(&lock);
            },
```

test/gtest/ucs/test_type.cc
```diff
@@ -138,6 +142,145 @@ UCS_TEST_F(test_type, pack_float) {
...
+                ucs_rwlock_write_lock(&lock);
+                ucs_rwlock_write_unlock(&lock);
+            },
+            "builtin");
+}
+
+UCS_TEST_F(test_rwlock, pthread) {
+    pthread_rwlock_t plock;
+    pthread_rwlock_init(&plock, NULL);
+    measure(
```
iyastreb, line 271: same here, update/verify some state

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        __atomic_fetch_sub(&lock->l, UCS_RWLOCK_READ, __ATOMIC_RELAXED);
+    }
+}
+
+
+static inline void ucs_rwlock_read_unlock(ucs_rwlock_t *lock)
+{
+    __atomic_fetch_sub(&lock->l, UCS_RWLOCK_READ, __ATOMIC_RELAXED);
```
iyastreb, line 67: I'm not sure that relaxed memory ordering is enough here (btw litmus test should fail in this case)
I think we need `__ATOMIC_RELEASE` to ensures memory operations before releasing the lock are completed.

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+    volatile int l;
+} ucs_rwlock_t;
+
+
+static inline void ucs_rwlock_read_lock(ucs_rwlock_t *lock)
+{
+    int x;
+
+    while (1) {
+        while (lock->l & UCS_RWLOCK_MASK) {
```
iyastreb, line 51: I'm not sure why we read atomic without mem order guarantees, normally it should be something
```C
__atomic_load_n(&lock->l, __ATOMIC_ACQUIRE)
```
Same in other read cases

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+#define UCS_RWLOCK_READ  0x4 /* Reader increment */
+
+#define UCS_RWLOCK_STATIC_INITIALIZER {0}
+
+
+/**
+ * Read-write lock.
+ */
+typedef struct {
+    volatile int l;
```
iyastreb, line 42: If we do all the loads with `__atomic_load_n` then volatile shouldn't be needed

src/ucs/type/rwlock.h
Artemy-Mellanox: why? AFAIU relaxed mem order means - no mem order guarantees

src/ucs/type/rwlock.h
Artemy-Mellanox: btw litmus test doesn't detect it

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+{
+    int x;
+
+    while (1) {
+        x = lock->l;
+        if ((x < UCS_RWLOCK_WRITE) &&
+            (__atomic_compare_exchange_n(&lock->l, &x, UCS_RWLOCK_WRITE, 0,
+                                         __ATOMIC_ACQUIRE, __ATOMIC_RELAXED))) {
+            return;
+        }
```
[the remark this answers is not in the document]
Artemy-Mellanox: they are different. here we set the write flag and remove the wait. in trylock we leave the wait alone to avoid write starvation. here we use the value of  `x` that we get and trylock doesn't return it.

test/gtest/ucs/test_type.cc
Artemy-Mellanox: I'm not sure this is a good idea. This test doesn't guarantee lock correctness, it will very easily give a false negative result.

test/gtest/ucs/test_type.cc
Artemy-Mellanox: 1. you mean write percent 0?
2. didn't i do that?

test/gtest/ucs/test_type.cc
iyastreb: Well, testing correctness of MT algorithms is hard topic. You're right about false negatives results, that's ok. That's actually a nature of litmus tests: when they pass, it does not guarantee that your algorithm is 100% correct, but when they fail - it's obviously broken. Personally I catch tons of MT issues with the help of litmus tests.
If you propose some other way of testing correctness - let's discuss that. What I'm proposing is well established industry practise, and I'm sure it's better to have it than no testing at all.

Moreover, the example that I provided is just scratching the surface. We should also consider adding tests with nested locks, try-locks, etc

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+        return 0;
+    }
+
+    return -EBUSY;
+}
+
+
+static inline void ucs_rwlock_write_unlock(ucs_rwlock_t *lock)
+{
+    __atomic_fetch_sub(&lock->l, UCS_RWLOCK_WRITE, __ATOMIC_RELAXED);
```
tvegas1, line 111: i think it is usually `__ATOMIC_RELEASE` to be used on all release paths to contain and be sure that what happened under lock is visible, any reason for not doing so in this PR?

test/gtest/ucs/test_type.cc
```diff
@@ -138,6 +142,145 @@ UCS_TEST_F(test_type, pack_float) {
...
+
+        for (auto t : threads) {
+            for (auto writers : writers_per_256) {
+                measure_one(t, writers, r, w, name);
+            }
+        }
+    }
+};
+
+UCS_TEST_F(test_rwlock, lock) {
```
tvegas1, line 199: Do we have a way to run this/some tests with maximal optimizations (maybe inline compiler pragma, ..) ? I suspect we end-up running it without optimisations which might affect correctness-related tests?

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+#define UCS_RWLOCK_WAIT  0x1 /* Writer is waiting */
+#define UCS_RWLOCK_WRITE 0x2 /* Writer has the lock */
+#define UCS_RWLOCK_MASK  (UCS_RWLOCK_WAIT | UCS_RWLOCK_WRITE)
+#define UCS_RWLOCK_READ  0x4 /* Reader increment */
+
+#define UCS_RWLOCK_STATIC_INITIALIZER {0}
+
+
+/**
+ * Read-write lock.
```
tvegas1, line 39: spinlock

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+*/
+
+#ifndef UCS_RWLOCK_H
+#define UCS_RWLOCK_H
+
+#include <ucs/arch/cpu.h>
+#include <errno.h>
+
+/**
+ * The ucs_rwlock_t type.
```
tvegas1, line 14: suggestion: `ucs_rw_spinlock_t` and for all apis?

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+                                         __ATOMIC_ACQUIRE, __ATOMIC_RELAXED))) {
+            return;
+        }
+
+        if (!(x & UCS_RWLOCK_WAIT)) {
+            __atomic_fetch_or(&lock->l, UCS_RWLOCK_WAIT, __ATOMIC_RELAXED);
+        }
+
+        while (lock->l > UCS_RWLOCK_WAIT) {
+            ucs_cpu_relax();
```
tvegas1, line 88: would we benefit from either using `sched_yeld()`, or X times `ucs_cpu_relax()` before retrying to read `lock->l`?

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+        if (!(x & UCS_RWLOCK_MASK)) {
+            return;
+        }
+
+        __atomic_fetch_sub(&lock->l, UCS_RWLOCK_READ, __ATOMIC_RELAXED);
+    }
+}
+
+
+static inline void ucs_rwlock_read_unlock(ucs_rwlock_t *lock)
```
tvegas1, line 65: add assertions/checks to detect underflow using returned value, only on debug builds (and also overflow on the other path)

src/ucs/type/rwlock.h
tvegas1, line 111: detect underflow on debug builds?

src/ucs/type/rwlock.h
tvegas1, line 42: use `unsigned int` to have defined behavior on overflow/underflow (and detect it)?

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,120 @@
...
+{
+    __atomic_fetch_sub(&lock->l, UCS_RWLOCK_READ, __ATOMIC_RELAXED);
+}
+
+
+static inline void ucs_rwlock_write_lock(ucs_rwlock_t *lock)
```
tvegas1, line 71: maybe annotate for coverity with something like `/* coverity[lock] */`, to help it with sanity checks?

tvegas1: same for lock paths

src/ucs/type/rwlock.h
Artemy-Mellanox: we don't expect overflow by design so signed int is more suitable i this case IMO

src/ucs/arch/x86_64/cpu.h
```diff
@@ -135,13 +135,12 @@ ucs_memcpy_nontemporal(void *dst, const void *src, size_t len)
     ucs_x86_memcpy_sse_movntdqa(dst, src, len);
 }
 
-#ifdef __SSE2__
 static UCS_F_ALWAYS_INLINE void ucs_cpu_relax()
 {
+#ifdef __SSE2__
```
yosefe, line 140: minor: do something else if __SSE2__ not defined, like "":::"memory" ?

src/ucs/arch/atomic.h
```diff
@@ -138,4 +139,59 @@ UCS_DEFINE_ATOMIC_BOOL_CSWAP(16, w);
...
+{
+    if (flags & UCS_ATOMIC_FENCE_LOCK) {
+        return __ATOMIC_ACQUIRE;
+    }
+
+    if (flags & UCS_ATOMIC_FENCE_UNLOCK) {
+        return __ATOMIC_RELEASE;
+    }
+
+    return __ATOMIC_RELAXED;
```
yosefe, line 158: why are UCS_ATOMIC_xx defined as flags if there is no meaning to combine more than one of them?
also, how does it relate to UCS_DEFINE_ATOMIC_xx macros? should we remove/extend them?

src/ucs/arch/atomic.h
```diff
@@ -138,4 +139,59 @@ UCS_DEFINE_ATOMIC_BOOL_CSWAP(16, w);
...
+
+
+static UCS_F_ALWAYS_INLINE void ucs_atomic_or(int *ptr, int val, unsigned flags)
+{
+    __atomic_fetch_or(ptr, val, ucs_atomic_memorder(flags));
+}
+
+
+static UCS_F_ALWAYS_INLINE int
+ucs_atomic_cswap(int *ptr, int old_val, int new_val, unsigned flags)
```
yosefe, line 189: should we define it as macros to allow any type (not just int)?

src/ucs/arch/atomic.h
```diff
@@ -138,4 +139,59 @@ UCS_DEFINE_ATOMIC_BOOL_CSWAP(16, w);
...
+{
+    return __atomic_compare_exchange_n(ptr, &old_val, new_val,
+                                       flags & UCS_ATOMIC_WEAK,
+                                       ucs_atomic_memorder(flags),
```
yosefe, line 193: why in case of success we want weak mem order?

src/ucs/type/rwlock.h
```diff
@@ -27,94 +30,108 @@
...
 
 
 /**
- * Read-write lock.
+ * Reader-writer spin lock.
  */
 typedef struct {
-    volatile int l;
-} ucs_rwlock_t;
+    int state;
```
yosefe, line 45: maybe uint32_t ? since it's flags value and not numeric value

src/ucs/type/rwlock.h
```diff
@@ -27,94 +30,108 @@
...
 {
     int x;
 
-    x = lock->l;
+    x = ucs_atomic_get(&lock->state, 0);
     if ((x < UCS_RWLOCK_WRITE) &&
-        (__atomic_compare_exchange_n(&lock->l, &x, x + UCS_RWLOCK_WRITE, 1,
-                                     __ATOMIC_ACQUIRE, __ATOMIC_RELAXED))) {
-        return 0;
+        ucs_atomic_cswap(&lock->state, x, x + UCS_RWLOCK_WRITE,
```
yosefe, line 109: maybe use 
"!(x & UCS_RWLOCK_WRITE)" instead of "x < UCS_RWLOCK_WRITE"
"x | UCS_RWLOCK_WRITE" instead of "x + UCS_RWLOCK_WRITE" 
since this is bit value

src/ucs/type/rwlock.h
```diff
@@ -27,94 +30,108 @@
...
+static UCS_F_ALWAYS_INLINE void ucs_rw_spinlock_init(ucs_rw_spinlock_t *lock)
 {
-    lock->l = 0;
+    lock->state = 0;
+}
+
+
+static UCS_F_ALWAYS_INLINE void ucs_rw_spinlock_cleanup(ucs_rw_spinlock_t *lock)
+{
+    ucs_assert(lock->state == 0);
```
yosefe, line 134: assertv

test/gtest/ucs/test_type.cc
```diff
@@ -144,89 +145,91 @@ UCS_TEST_F(test_type, pack_float) {
...
-    ucs_rwlock_read_lock(&lock);
-    EXPECT_EQ(-EBUSY, ucs_rwlock_write_trylock(&lock));
+    ucs_rw_spinlock_read_lock(&lock);
+    EXPECT_FALSE(ucs_rw_spinlock_write_trylock(&lock));
 
-    ucs_rwlock_read_lock(&lock); /* second read lock should pass */
+    ucs_rw_spinlock_read_lock(&lock); /* second read lock should pass */
 
-    int write_taken = 0;
+    bool write_taken = 0;
```
yosefe, line 210: use true/false for bool instead of 1/0

test/gtest/ucs/test_type.cc
yosefe: 1. yes ( isee you added it)
2. yes, pls update with percent 0

src/ucs/arch/atomic.h
```diff
@@ -138,4 +139,59 @@ UCS_DEFINE_ATOMIC_BOOL_CSWAP(16, w);
 UCS_DEFINE_ATOMIC_BOOL_CSWAP(32, l);
 UCS_DEFINE_ATOMIC_BOOL_CSWAP(64, q);
 
+
+#define UCS_ATOMIC_WEAK         UCS_BIT(0)
```
iyastreb, line 143: If we want to encapsulate these values, then maybe it should be an enum?
And also we miss few other values:
__ATOMIC_CONSUME  - probably we don't need this
Data dependency only for both barrier and synchronization with another thread.
__ATOMIC_ACQ_REL - I'm not sure about this one
Full barrier in both directions and synchronizes with acquire loads and release stores in another thread.
__ATOMIC_SEQ_CST  - the strongest mode, we definitely need it
Full barrier in both directions and synchronizes with acquire loads and release stores in all threads.

src/ucs/arch/atomic.h
```diff
@@ -138,4 +139,59 @@ UCS_DEFINE_ATOMIC_BOOL_CSWAP(16, w);
...
+#define UCS_ATOMIC_FENCE_LOCK   UCS_BIT(1)
+#define UCS_ATOMIC_FENCE_UNLOCK UCS_BIT(2)
+
+
+static UCS_F_ALWAYS_INLINE int ucs_atomic_memorder(unsigned flags)
```
tvegas1, line 148: maybe we can simply use the actual flag (`__ATOMIC_ACQUIRE` etc) directly

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,137 @@
...
+
+        ucs_atomic_sub(&lock->state, UCS_RWLOCK_READ, 0);
+    }
+}
+
+
+static UCS_F_ALWAYS_INLINE void
+ucs_rw_spinlock_read_unlock(ucs_rw_spinlock_t *lock)
+{
+    ucs_assert(lock->state >= UCS_RWLOCK_READ);
```
tvegas1, line 73: assertv with details?

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,137 @@
+/*
+* Copyright (c) NVIDIA CORPORATION & AFFILIATES, 2024. ALL RIGHTS RESERVED.
```
tvegas1, line 2: copyrights would need 2025

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,137 @@
...
+#include <ucs/debug/assert.h>
+#include <ucs/sys/compiler_def.h>
+#include <errno.h>
+
+/**
+ * The ucs_rw_spinlock_t type.
+ *
+ * Readers increment the counter by UCS_RWLOCK_READ (4)
+ * Writers set the UCS_RWLOCK_WRITE bit when lock is held
+ * and set the UCS_RWLOCK_WAIT bit while waiting.
```
tvegas1, line 21: maybe detail something in the line of "the wait bit is meant for all subsequent read lock to let any write lock go first"

src/ucs/type/rwlock.h
tvegas1: maybe keep wait bit included in the check too

src/ucs/type/rwlock.h
```diff
@@ -131,7 +134,7 @@ static UCS_F_ALWAYS_INLINE void ucs_rw_spinlock_init(ucs_rw_spinlock_t *lock)
 
 static UCS_F_ALWAYS_INLINE void ucs_rw_spinlock_cleanup(ucs_rw_spinlock_t *lock)
 {
-    ucs_assert(lock->state == 0);
+    ucs_assertv(lock->state == 0, "lock no released state:%u", lock->state);
```
tvegas1, line 137: no -> not, maybe add `lock=%p`

tvegas1 (approved):
looks good, is the test code run with optim too?

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,149 @@
...
+                (_lock), (_lock)->state, \
+                (_lock)->state & UCS_RWLOCK_WAIT ? " WAIT" : "", \
+                (_lock)->state & UCS_RWLOCK_WRITE ? " WRITE" : "")
+
+
+/**
+ * Reader-writer spin lock.
+ */
+typedef struct {
+    uint32_t state;
```
yosefe, line 52: volatile?

src/ucs/type/rwlock.h
```diff
@@ -0,0 +1,149 @@
...
+        goto wait;
+    }
+
+    for (;;) {
+        if (__atomic_compare_exchange_n(&lock->state, &x, UCS_RWLOCK_WRITE, 0,
+                                        __ATOMIC_ACQUIRE, __ATOMIC_RELAXED)) {
+            return;
+        }
+
+wait:
```
yosefe, line 101: seems weird to goto to middle of loop, is there a better way to write this?
also, can we implement write_lock using write_trylock?

src/ucs/type/rwlock.h
Artemy-Mellanox: @iyastreb suggested to remove it for better performance and use __atomic_load_n to ensure data is not cached

src/ucs/type/rwlock.h
Artemy-Mellanox: this way we avoid redundant memory access in the loop https://github.com/openucx/ucx/pull/10355/commits/605883044e1de2c63fadf57101d099a5b101ceea

src/ucs/type/rwlock.h
yosefe: so we shall comment this field can be accessed only with atomic operations?

src/ucs/type/rwlock.h
yosefe: ok

src/ucs/type/rwlock.h
Artemy-Mellanox: it's internal state, shouldn't be accessed externally

yosefe (approved):

